PR #1054, opened 2019-07-30
Use -i/-m instead of -a/-e with GHDL

corna:
Currently, if multiple libraries or files depend on each other, they
must be analyzed in the correct order.
With -i/-m GHDL is able to automatically detect the correct build order.


corna:
Doc [here](https://ghdl.readthedocs.io/en/latest/using/InvokingGHDL.html#import-i).

cmarqu:
Is this maybe a fairly new feature of GHDL though, one that we cannot expect our users to have?

themperek:
@cmarqu It seems to be ok for v0.36.

corna:
It was present at least in [v0.32](https://github.com/ghdl/ghdl/blob/ghdl-0.32/doc/ghdl.texi#L1055-L1065) (2014).

themperek:
Thank you @corna !

PR #1056, opened 2019-08-01
Fixes #1055. Py_Finalize called before simulator end

### LINKED ISSUE #1055: Py_Finalize() is never called
`Py_Finalize()` runs a number of very important cleanup routines, like flushing `stderr` and `stdout` (how I figured this out) and running `__del__` on living objects in Python 3.

ktbarrett:


cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -119,6 +119,7 @@ void gpi_embed_end(void)
 
 {
     embed_sim_event(SIM_FAIL, "Simulator shutdown prematurely");
+    embed_sim_end();
```
eric-wieser: Based on the line above, this only seems to cover failed shutdowns - how are successful shutdowns handled?

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -384,3 +384,10 @@ void embed_sim_event(gpi_event_t level, const char *msg)
 
     FEXIT
 }
+
+void embed_sim_end(void)
+{
+    FENTER
+
+    Py_Finalize();
```
eric-wieser: personally I think this FENTER / FEXIT stuff is garbage, but if we're going to use it, you should add an FEXIT here too

ktbarrett: I was under the impression that the gpi_log used python's logger. If the interpreter is finalized we can't log anything thereafter.

cocotb/share/lib/gpi/GpiCommon.cpp
ktbarrett: Every interface calls `gpi_embed_end` as a part of their shutdown callback. Perhaps the above line shouldn't be there.

ktbarrett:
There is a failure in the regressions on the `test_exit_code` testcase where the code segfaults in `Py_Finalize()`. Along with the Eric's comment about how this is handled in successful shutdown I think this is above my understanding of the C internals, which are unfortunately undocumented. I will take this issue to someone who understand the C code a bit better.

Issue #838, opened 2019-03-10
Should all simulators write their logfile to $SIM_BUILD/sim.log?


cmarqu:
Right now, ``Makefile.aldec``, ``Makefile.ius`` and ``Makefile.questa`` send the simulator logfile to ``$SIM_BUILD/sim.log`` - the rest appear not to. 
Should we make this consistent?

cmarqu:
Duplicated by #1746.

PR #1063, opened 2019-08-08
Questa: start simulation from PWD, not PWD/sim_build

TC01:
The Icarus and Cadence (ius) simulator targets run the simulation relative to ```${PWD}```; i.e. the location of the Makefile. The Questa one, though, seems to run everything out of the ```${PWD}/sim_build``` directory.

This commit changes the questa Makefile to run from ```${PWD}```, while keeping the "work" library and the run script in sim_build, for better compatibility with the other simulators.

In particular, I just tried to port a project from Cadence to Questa and found that, because my Makefile didn't explicitly refer to ```${PWD}``` but instead just provided source paths relative to the current working directory, the simulation would not start. Hence this change.

Hopefully this does not break anything: it would be great if people with more mature Questa projects could test!

I should say: it's possible that this isn't desired behavior. In which case, maybe the other simulators should be changed instead for consistency? (This all also may be immaterial if the makefile approach to cocotb is replaced with the purer-Python cocotb-test approach, but assuming we keep some kind of compatibility layer with the Makefiles, it would probably still be good to fix this).

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.questa
+++ b/cocotb/share/makefiles/simulators/Makefile.questa
@@ -89,10 +89,10 @@ $(SIM_BUILD)/runsim.do : $(VHDL_SOURCES) $(VERILOG_SOURCES) $(CUSTOM_SIM_DEPS) $
 	@echo "onerror {" >> $@
 	@echo "	quit -f -code 1" >> $@
 	@echo "}" >> $@
-	@echo "if [file exists $(RTL_LIBRARY)] {vdel -lib $(RTL_LIBRARY) -all}" >> $@
-	@echo "vlib $(RTL_LIBRARY)" >> $@
+	@echo "if [file exists $(SIM_BUILD)/$(RTL_LIBRARY)] {vdel -lib $(SIM_BUILD)/$(RTL_LIBRARY) -all}" >> $@
+	@echo "vlib $(SIM_BUILD)/$(RTL_LIBRARY)" >> $@
 	@echo "vmap -c" >> $@
-	@echo "vmap $(RTL_LIBRARY) $(RTL_LIBRARY)" >> $@
+	@echo "vmap $(RTL_LIBRARY) $(SIM_BUILD)/$(RTL_LIBRARY)" >> $@
 ifneq ($(VHDL_SOURCES),)
 	@echo "vcom -work $(RTL_LIBRARY) $(VCOM_ARGS) $(VHDL_SOURCES)" >> $@
 endif
@@ -102,7 +102,7 @@ endif
 ifdef SCRIPT_FILE
 	@echo "do $(SCRIPT_FILE)" >> $@
 endif
-	@echo "vsim $(VSIM_ARGS) $(EXTRA_ARGS) $(TOPLEVEL)" >> $@
+	@echo "vsim $(VSIM_ARGS) $(EXTRA_ARGS) $(SIM_BUILD)/$(TOPLEVEL)" >> $@
 ifeq ($(WAVES),1)
 	@echo "log -recursive /*" >> $@
 endif
@@ -146,9 +146,9 @@ endif
 $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/runsim.do $(COCOTB_LIBS) $(INT_LIBS)
 	-@rm -f $(COCOTB_RESULTS_FILE)
 
-	set -o pipefail; cd $(SIM_BUILD) && $(LIB_LOAD) MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) COCOTB_SIM=1 \
+	set -o pipefail; $(LIB_LOAD) MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) COCOTB_SIM=1 \
 	GPI_EXTRA=$(GPI_EXTRA) TOPLEVEL_LANG=$(TOPLEVEL_LANG) PYTHONPATH=$(LIB_DIR):$(PWD):$(NEW_PYTHONPATH) \
-	$(CMD) $(PLUSARGS) -do runsim.do 2>&1 | tee sim.log
+	$(CMD) $(PLUSARGS) -do $(SIM_BUILD)/runsim.do 2>&1 | tee $(SIM_BUILD)/sim.log
 
 	# check that the file was actually created, since we can't set an exit code from cocotb
 	test -f $(COCOTB_RESULTS_FILE)
--- a/documentation/source/building.rst
+++ b/documentation/source/building.rst
@@ -58,10 +58,12 @@ and
 .. make:var:: VERILOG_SOURCES
 
       A list of the Verilog source files to include.
+      Paths can be absolute or relative; if relative, they are interpreted as relative to the Makefile's location.
 
 .. make:var:: VHDL_SOURCES
 
       A list of the VHDL source files to include.
+      Paths can be absolute or relative; if relative, they are interpreted as relative to the Makefile's location.
 
 .. make:var:: VHDL_SOURCES_<lib>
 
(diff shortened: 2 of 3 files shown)


TC01:
For what it's worth, here is a [very simple example project](https://github.com/TC01/questa-cocotb-paths-example) where simulation successfully starts with Cadence and Icarus but fails with Questa for this issue.

themperek:
Hard to say what should be a desirable way. VCS and aldec makefile seem also different. 

In cocotb-test I decided to run all simulators form sim_build. 

One alternative (maybe in any case) would be to make an absolute path for sources in Makefile (maybe here: https://github.com/cocotb/cocotb/blob/master/cocotb/share/makefiles/Makefile.sim) before giving this to the simulator (I do this in cocotb-test).  In makefile can call python to do this or `$(abspath ..)` for every source file.

In any case, if we change should change everywhere the same way. In any case, touching those Makefiles is always "controversial".

TC01:
> One alternative (maybe in any case) would be to make an absolute path for sources in Makefile (maybe here: https://github.com/cocotb/cocotb/blob/master/cocotb/share/makefiles/Makefile.sim) before giving this to the simulator (I do this in cocotb-test).

I like this idea, it seems much more elegant than just moving the location the simulator starts from around. (I briefly wondered if this could be done by passing an argument to vsim/vlog but didn't think to try changing it in Makefile.sim).

For the makefile approach, perhaps we could have some new parameter (```$SRC_DIR``` or ```$RTL_SRC_DIR``` or similar) that defaults to the current working directory but could be changed at the users discretion?

I can try to put that together and submit a new PR.

Separately, I'm still wondering whether or not it would be good to be consistent about where the simulators start from, but as you say it'd be better to fix that across all the simulators rather than one at a time. (I only have access to questa, ius, and icarus at the moment, which is why I only looked at those).

themperek:
@TC01 in cocotb-test I did different/simple compatibility/adoption (https://github.com/themperek/cocotb-test/blob/master/cocotb_test/Makefile.inc) so we do not have to maintain  2 systems. 

We need @imphil to help on this.

imphil:
Thanks for this PR and your quest for consistency @TC01! I'd say the current behavior of how  relative paths in VERILOG_SOURCES/VHDL_SOURCES is somewhat "undefined behavor"; all our examples and our documentation use absolute paths.

To push this forward,
1. We should define the desired behavior: if we encounter a relative path in VERILOG_SOURCES or VHDL_SOURCES, what is it relative to?
2. We need to document this behavior.
3. We should make sure all simulators behave the same.

My thinking on this:
1. The only sane way for relative paths to behave is relative to the file they are mentioned in. That means relative to the Makefile, which is the behavior we currently have for icarus and ius.
2. Can you add a commit updating the documentation in `documentation/source/building.rst` regarding these two variables?
3. Your approach is one option to ensure that relative source paths are interpreted relative to the Makefile. Another would be to transform all paths to absolute ones. I'm fine with both approaches.

cocotb/share/makefiles/simulators/Makefile.questa
```diff
@@ -43,7 +43,7 @@ else
     export MODELSIM_BIN_DIR
 endif
 
-RTL_LIBRARY ?= work
+RTL_LIBRARY ?= sim_build/work
```
imphil: RTL_LIBRARY is expected to be a library name, not a path (that's what we use for other simulators as well). So this needs to stay "work".

cocotb/share/makefiles/simulators/Makefile.questa
```diff
@@ -146,9 +146,9 @@ endif
 # Depending on the version of modelsim the interfaces that it supports can change
 # Append or remove values to INT_LIBS depenending on your license
 results.xml: $(SIM_BUILD)/runsim.do $(COCOTB_LIBS) $(INT_LIBS)
-	set -o pipefail; cd $(SIM_BUILD) && $(LIB_LOAD) MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) COCOTB_SIM=1 \
+	set -o pipefail; $(LIB_LOAD) MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) COCOTB_SIM=1 \
 	GPI_EXTRA=$(GPI_EXTRA) TOPLEVEL_LANG=$(TOPLEVEL_LANG) PYTHONPATH=$(LIB_DIR):$(PWD):$(NEW_PYTHONPATH) \
-	$(CMD) $(PLUSARGS) -do runsim.do 2>&1 | tee sim.log
+	$(CMD) $(PLUSARGS) -do ${SIM_BUILD}/runsim.do 2>&1 | tee sim.log
```
imphil: What's the issue with starting questa from PWD and having all libraries and temporary files there as well, i.e. get rid of sim_build entirely. That's what we do for other simulators, so it's the expected behavior (it's not beautiful, but at least consistent).

documentation/source/building.rst
```diff
@@ -34,10 +34,10 @@ Make Variables
       Selects which simulator Makefile to use.  Attempts to include a simulator specific makefile from :file:`cocotb/makefiles/makefile.$(SIM)`
 
     ``VERILOG_SOURCES``
-      A list of the Verilog source files to include.
+      A list of the Verilog source files to include. Paths can be absolute or relative; if relative, they are interpreted as relative to the Makefile's location.
```
imphil: Nit: we typically have one sentence per line.

cocotb/share/makefiles/simulators/Makefile.questa
TC01: Well, the current behavior is that questa is started from sim_build and loads a script that lives there (runsim.do).

This commit changes things so the generated script is still in sim_build but questa starts from PWD. I thought that was more consistent with the other simulators-- Cadence, at least, puts a log file and the compiled libraries (the "INCA_libs" directory) in sim_build.

But we could put everything in PWD if that makes more sense.

cocotb/share/makefiles/simulators/Makefile.questa
TC01: Ah, okay. I think that is fine; it'll just require rewriting runsim.do to use ```${SIM_BUILD}/${RTL_LIBRARY}``` everywhere (instead of just ```${RTL_LIBRARY}```) I guess.

Is that an okay solution? The library gets passed to questa via the ```-work``` flag, which seems to be quite happy taking a path (hence why this commit works).

(Of course, if we got rid of sim_build here then that wouldn't be necessary, as discussed in the other comment).

documentation/source/building.rst
TC01: Will fix and rebase!

cocotb/share/makefiles/simulators/Makefile.questa
themperek: BTW: one can do without do file but with -do argument like here: https://github.com/themperek/cocotb-test/blob/54602a6a954a46d9d6583e897a7d814b8d408022/cocotb_test/simulator.py#L298 but not sure if this can be done with Makefile.

cocotb/share/makefiles/simulators/Makefile.questa
TC01: Sorry for the delay here. I've pushed a new commit making this change; RTL_LIBRARY is still "work" and references to it in the makefile use SIM_BUILD/RTL_LIBRARY rather than just RTL_LIBRARY.

TC01:
I also just noticed that these changes moved sim.log from ```$PWD/sim_build/sim.log``` to ```$PWD/sim.log```. I've just pushed another commit fixing that.

Let me know how I should proceed here: if this looks okay as is or if we want to take a different approach in making the simulators consistent.

cmarqu:
Related: #838

TC01:
PIng @imphil?

imphil:
@TC01 sorry for the significant delay. I've rebased this PR on top of master, and will merge it once CI passes.

PR #1067, opened 2019-08-15
Add spell checking for docs.

cmarqu:
New target ``make spelling`` in ``documentation/``.
This sets the language to ``en_GB`` - while we gradually have become more "americanized" in spelling, the number of reported errors was still lower when using the GB English dictionary compared to US English.
The actual fixing of errors will be done in a separate PR. Once *that* is finished with zero remaining errors we could think of running this check in CI.


imphil:
Please go for en_US right from the start, it's the de-facto standard for technical documentation and international projects these days.

documentation/source/c_symbols.txt
```diff
@@ -0,0 +1,63 @@
+FliCbHdl
```
eric-wieser: Why is this file needed? If these words are used in cross-references, aren't they excluded from the spellcheck?

cmarqu: There is an option to skip such MixedCase identifiers but I explicitly disabled it because I wanted "GitHub", "SystemVerilog" etc. to be checked.
The identifiers in this file are generated by breathe for the C++ Library Reference section but apparently not marked up so that spellchecking would skip them.

documentation/source/conf.py
```diff
@@ -305,4 +307,21 @@
...
 breathe_show_define_initializer = True
 
-autoclass_content = "both"
+# -- Extra setup for spelling check --------------------------------------------
+
+# Spelling check needs an additional module that is not installed by default.
+# Add it only if spelling check is requested so docs can be generated without it.
+
+if 'spelling' in sys.argv:
+    extensions.append("sphinxcontrib.spelling")
```
imphil: I guess you can unconditionally enable it; it's part of the requirements.txt and should be available to everyone building the documentation.

documentation/source/spelling_wordlist.txt
```diff
@@ -0,0 +1,87 @@
...
+Lite
+Microsemi
+MinGW
+ModelSim
+Msys
+NVC
+Questa
+Quickstart
+README
+Roadmap
```
imphil: That's not a normal word? Or is it just the capitalization?

documentation/source/conf.py
eric-wieser: I suspect it fails right now, hence it not being enabled by default.

documentation/source/c_symbols.txt
eric-wieser: I imagine that in future we could either
* Create a custom spelling filter that reads the breathe data-files to produce this list at runtime
* Change breathe to not output a format that does this

For now, what you have is a tidy workaround - shame you can't put a comment in a wordlist file.

documentation/source/spelling_wordlist.txt
cmarqu: It's going to differ depending on what spell checker backend enchant is configured to use, and what dictionary it has available, but on my fairly vintage system, it's indeed considered a typo (with "roadkill" being one of the suggestions :)

documentation/source/conf.py
cmarqu: I had copied that part of the setup from Django-CMS, without knowing exactly why they have it like that, but it seemed safe.

PR #1069, opened 2019-08-15
Use PyErr_SetString() instead of fprintf() before a return NULL

### LINKED ISSUE #1068: Use `PyErr_SetString("...")`before a `return NULL`
Note that if this line is reached, python will throw a `SystemError` shortly after anyway. This needs to be `PyErr_SetString("...")`, as does almost every printf that appears before a `return NULL` - python functions are only allowed to return `NULL` if they set an exception

_Originally posted by @eric-wieser in https://github.com/cocotb/cocotb/pull/1042_

cmarqu:
Closes #1068.


imphil:
Thanks!

PR #1070, opened 2019-08-16
Add target 'regression' to the 'all tests' Makefile.

cmarqu:
I'm gradually working through all tests in my local, non-Travis setting and was missing the ``regression`` target there.


imphil:
LGTM, thanks!

PR #1071, opened 2019-08-16
Fix extension for files that are SystemVerilog.

cmarqu:
These files use ``typedef`` and are thus SystemVerilog really, so fix their extension. This makes them automatically compile in Cadence simulators without further switches.


themperek:
Thank you @cmarqu and @eric-wieser.

PR #1072, opened 2019-08-16
Add support for the RTL_LIBRARY make variable.

cmarqu:
This was missing from Cadence simulator makefiles, and the ``viterbi_decoder_axi4s`` design uses it.


cocotb/share/makefiles/simulators/Makefile.ius
```diff
@@ -62,6 +62,12 @@ else
     EXTRA_ARGS +=
 endif
 
+ifdef RTL_LIBRARY
```
themperek: something like this would not be shorter?
```
RTL_LIBRARY ?= $(TOPLEVEL)
MAKE_LIB = -makelib $(RTL_LIBRARY )
```

cmarqu: Oh sure, thanks for pointing this out.

imphil:
I've opened https://github.com/cocotb/cocotb/issues/1088 to get this variable documented.

PR #1074, opened 2019-08-18
Add XCELIUM to existing IUS define checks.

cmarqu:
For some reason I missed to also check for the ``XCELIUM`` define when I introduced ``Makefile.xcelium``, fix that.

### DIFF
--- a/cocotb/share/include/vhpi_user.h
+++ b/cocotb/share/include/vhpi_user.h
@@ -169,7 +169,7 @@ extern "C" {
 #ifndef IUS
         size_t bufSize;  /* the size in bytes of the value buffer; this is set by the user */
 #else
-        int32_t bufSize;  /* IUS defines this as 32-bits, even when running in 64-bit mode */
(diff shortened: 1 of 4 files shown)


themperek:
Is not better to just keep one in the define (across the code).

cmarqu:
I thought about this but didn't like it so much - it then has some magic going on behind the scenes, instead of the current straightforward ``simname.upper() == DEFINE`` relation.
Also, I would expect that in the future with ghdl and Verilator getting incrementally more VPI/VHPI support, we may need to have more such switches based on both simulator names and also their versions, so the slightly more complex ``#if defined`` will not stand out much.

Of course, I'm open to change this if normalizing to a single define is the mainstream opinion.

eric-wieser:
The normalization could happen in the c preprocessor. Is there a name that covers both of these?

cmarqu:
``CADENCE`` would cover both.

themperek:
Fot Qesta we keep MODELSIM do do not see what we cannot do same here.

cmarqu:
Alright, I now used the method of the existing ModelSim/Questa normalization.

PR #1078, opened 2019-08-21
Removed SIM_ARGS from test_iteration_vhdl

cogsy23:
for #1073.

The total warning look like this:
```
../../src/ieee/numeric_std-body.v93:2098:7:@0ms:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
../../src/ieee/numeric_std-body.v93:2098:7:@0ms:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
../../src/ieee/numeric_std-body.v93:2098:7:@0ms:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
../../src/ieee/numeric_std-body.v93:2098:7:@0ms:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
../../src/ieee/numeric_std-body.v93:1710:7:@0ms:(assertion warning): NUMERIC_STD."=": metavalue detected, returning FALSE
../../src/ieee/numeric_std-body.v93:1710:7:@0ms:(assertion warning): NUMERIC_STD."=": metavalue detected, returning FALSE
```
I'm happy to just remove SIM_ARGS if it's all the same.  It's not required and I can't see other examples of sim specific args.

PR #1082, opened 2019-08-23
Disabled warnings at time 0 for test_iteration_vhdl

cogsy23:
Follow up for #1078.

I can only test the GHDL path, just guessing with IUS/Xcelium :P


themperek:
Thank you.

PR #1086, opened 2019-08-24
BUG: Fix inverted condition on the log filter

eric-wieser:
Fixes a regression introduced in e3e1c06f14978cb563370ceeee906c5278574f55 (which unfortunately made it into the 1.2 release)

---

Also handles non-boolean return values correctly.

### DIFF
--- a/cocotb/share/lib/gpi_log/gpi_logging.c
+++ b/cocotb/share/lib/gpi_log/gpi_logging.c
@@ -159,14 +159,19 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
         PyGILState_Release(gstate);
         return;
     }
-
-    if (filter_ret == Py_True) {
-        Py_DECREF(filter_ret);
+    int is_enabled = PyObject_IsTrue(filter_ret);
+    Py_DECREF(filter_ret);
+    if (is_enabled < 0) {
+        /* A python exception occured while converting `filter_ret` to bool */
+        PyErr_Print();
         PyGILState_Release(gstate);
         return;
     }
 
-    Py_DECREF(filter_ret);
+    if (!is_enabled) {
+        PyGILState_Release(gstate);
+        return;
+    }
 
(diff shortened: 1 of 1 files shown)


eric-wieser:
Failure is here:

https://travis-ci.org/cocotb/cocotb/jobs/576079403#L3908

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -159,14 +159,18 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
         PyGILState_Release(gstate);
         return;
     }
-
-    if (filter_ret == Py_True) {
```
eric-wieser: Changing to `!=` would also have worked here, but this formulation makes it work if `isEnabledFor` returns `1` instead of `True.

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -159,14 +159,18 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
...
+        PyErr_Print();
         PyGILState_Release(gstate);
         return;
     }
 
-    Py_DECREF(filter_ret);
+    if (!is_enabled) {
+        PyGILState_Release(gstate);
+        return;
+    }
```
imphil: That code is slightly confusing; I didn't expect `PyObject_IsTrue()` to return -1 in my initial read of the code, and then it didn't make much sense. How about making this a bit clearer by
- checking for `is_enabled == -1` for the error condition, and add a comment like "unable to determine if |filter_ret| is true or false"
- checking for `is_enabled == 0` for the false case.

imphil: Thanks for adding a comment. People reading this file are not typically reading cpython source code, so a comment can help reduce confusion here.

imphil:
Thanks @eric-wieser!

Issue #804, opened 2019-02-06
Replace license header with SPDX identifier


imphil:
Currently we have a large comment block at every source file containing the 3-clause BSD license header. I'm proposing to strip this down to a reference to the LICENSE file and a SPDX license identifier (28 -> 3 lines).

Currently we have:
```
###############################################################################
# Copyright (c) 2013 Potential Ventures Ltd
# Copyright (c) 2013 SolarFlare Communications Inc
# All rights reserved.
#
# Redistribution and use in source and binary forms, with or without
# modification, are permitted provided that the following conditions are met:
#     * Redistributions of source code must retain the above copyright
#       notice, this list of conditions and the following disclaimer.
#     * Redistributions in binary form must reproduce the above copyright
#       notice, this list of conditions and the following disclaimer in the
#       documentation and/or other materials provided with the distribution.
#     * Neither the name of Potential Ventures Ltd,
#       SolarFlare Communications Inc nor the
#       names of its contributors may be used to endorse or promote products
#       derived from this software without specific prior written permission.
#
# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
# DISCLAIMED. IN NO EVENT SHALL POTENTIAL VENTURES LTD BE LIABLE FOR ANY
# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
###############################################################################
```

I'm proposing to change this to:
```
# Copyright cocotb contributors
[...]

PR #1093, opened 2019-08-28
Move backports to their own file, to more clearly separate them from utilities

eric-wieser:
By starting this new module name with a `_`, we indicate that users are not supposed to be touching these pieces, and they will be removed without notice (when cocotb increases its minimum python version).

### DIFF
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -28,6 +28,7 @@ Backports and compatibility shims for newer python features.
 These are for internal use - users should use a third party library like `six`
 if they want to use these shims in their own code
 """
+import abc
 import sys
 
 # This is six.integer_types
@@ -131,3 +132,8 @@ class nullcontext(object):
 
     def __exit__(self, *excinfo):
         pass
+
+
+# https://stackoverflow.com/a/38668373
+# Backport of abc.ABC, compatible with Python 2 and 3
+abc_ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -1,0 +1,133 @@
+# Copyright (c) cocotb contributors
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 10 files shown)


cocotb/_compat.py
```diff
@@ -0,0 +1,109 @@
+"""
```
imphil: Please include the standard license header.

Maybe name this file "pycompat"? We might introduce compatibility code in the future which maintains compat with older cocotb versions.

cocotb/_compat.py
```diff
@@ -0,0 +1,109 @@
...
+# this is six.raise_from
+if sys.version_info[:2] == (3, 2):
+    exec_("""def raise_from(value, from_value):
+    try:
+        if from_value is None:
+            raise value
+        raise value from from_value
+    finally:
+        value = None
+    """)
```
imphil: We don't support Python 3.2 any more

imphil: Sure, that's fine.

In general, our policy has been "all officially supported Python versions". I thought we had documented that somewhere more prominently, but I didn't find it except for the latest release notes "Python 2.6 is officially not supported any more. cocotb supports Python 2.7 and Python 3.5+." (https://cocotb.readthedocs.io/en/latest/release_notes.html)

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -46,6 +46,7 @@
 from cocotb.utils import get_sim_time
 
 from cocotb.binary import BinaryValue
+from cocotb import _compat
```
imphil: that should be _py_compat as well (also below)

cocotb/_py_compat.py
```diff
@@ -0,0 +1,109 @@
+"""
```
imphil: Please add a license header.

imphil: Since this is a new file, please make it "Copyright cocotb contributors" in the first line (no year needed). I hope we can get these headers cleaned up one day (#804).

eric-wieser: Updated

PR #1094, opened 2019-08-28
Align makefiles for examples/ with those for tests/.

cmarqu:
List all examples in higher-level makefile; add target ``regression`` to it.
Add Travis fold markers in case we want to run examples in CI at some point.
Add option for VHDL93 for Cadence simulators in the single example's makefiles.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -27,10 +27,13 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
-
 EXAMPLES := adder/tests \
-	    endian_swapper/tests \
             axi_lite_slave/tests \
+            dff/tests \
+            endian_swapper/tests \
+            mean/tests \
+            mixed_language/tests \
+            ping_tun_tap/tests
 
 .PHONY: $(EXAMPLES)
 
@@ -38,8 +41,17 @@ EXAMPLES := adder/tests \
 all: $(EXAMPLES)
 
 $(EXAMPLES):
+ifeq ($(TRAVIS),true)
+	@echo "travis_fold:start:$@"
+endif
 	@cd $@ && $(MAKE)
+ifeq ($(TRAVIS),true)
+	@echo "travis_fold:end:$@"
+endif
 
 .PHONY: clean
 clean:
 	$(foreach TEST, $(EXAMPLES), $(MAKE) -C $(TEST) clean;)
+
+regression:
+	$(foreach TEST, $(EXAMPLES), $(MAKE) -C $(TEST) regression;)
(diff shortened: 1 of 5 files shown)


examples/Makefile
```diff
@@ -27,19 +27,31 @@
...
 all: $(EXAMPLES)
 
 $(EXAMPLES):
+ifeq ($(TRAVIS),true)
+	@echo "travis_fold:start:$@"
+endif
 	@cd $@ && $(MAKE)
+ifeq ($(TRAVIS),true)
+	@echo "travis_fold:end:$@"
+endif
```
imphil: Please don't add these markers in our general Makefile; we don't want to expose our CI specialties there. (The markers don't add much value anyways from my PoV.)

examples/Makefile
```diff
@@ -27,19 +27,31 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
-
 EXAMPLES := adder/tests \
-	    endian_swapper/tests \
             axi_lite_slave/tests \
+            dff/tests \
```
imphil: These tests fail in Travis with

` ValueError: Unable to accurately represent 10(us) with the simulator precision of 1e0`

examples/Makefile
```diff
@@ -27,19 +27,31 @@
...
+            endian_swapper/tests \
+            mean/tests \
+            mixed_language/tests \
+            ping_tun_tap/tests
```
imphil: This last test requires root permissions to run (and that's why it's failing currently in Travis). Requiring root to run "make test" isn't reasonable. Please remove this test from the list. You can optionally run the test by itself using sudo in Travis.

examples/dff/tests/Makefile
```diff
@@ -15,13 +15,17 @@ else
...
-TOPLEVEL=dff
-MODULE=$(TOPLEVEL)_cocotb
+TOPLEVEL = dff
+MODULE := $(TOPLEVEL)_cocotb
 
 CUSTOM_SIM_DEPS=$(CWD)/Makefile
 
 ifeq ($(SIM),questa)
-SIM_ARGS=-t 1ps
+    SIM_ARGS=-t 1ps
```
imphil: The same problem is visible in Icarus (see above). Can't we add a "timescale" option to the file directly?

cmarqu: Done in #1132.

examples/dff/tests/Makefile
```diff
@@ -15,13 +15,17 @@ else
...
+endif
+
+ifeq ($(SIM),$(filter $(SIM),ius xcelium))
```
cmarqu: Theses moved to #1134.

examples/Makefile
cmarqu: Fixed in #1132.

examples/Makefile
cmarqu: Ok, they are not part of the PRs any longer.

examples/Makefile
cmarqu: Cleaned up in #1135.

cmarqu:
Closing this, most parts moved to other PRs.

PR #1096, opened 2019-08-28
Do spellchecking

cmarqu:


cocotb/drivers/__init__.py
```diff
@@ -207,7 +207,7 @@ def _send_thread(self):
 
 
 class BusDriver(Driver):
-    """Wrapper around common functionality for busses which have:
+    """Wrapper around common functionality for buses which have:
```
imphil: busses is valid in AE, but fine either way.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -203,7 +203,7 @@ static PyObject *log_msg(PyObject *self, PyObject *args)
 }
 
 
-// Register a callback for read only state of sim
+// Register a callback for ReadOnly state of sim
```
imphil: ReadOnly() is the name of our trigger. The state of the simulation is actually read-only (or readonly).

cmarqu:
Note that this is still a draft PR currently.
I actually *am* using the amend option and I push with force, but for some reason it's not enough...

cmarqu:
Superseded by https://github.com/cocotb/cocotb/pull/1124.

PR #1097, opened 2019-08-28
Add wavedrom module to library reference.

cmarqu:


documentation/source/library_reference.rst
```diff
@@ -269,3 +269,14 @@ XGMII
     :members:
     :member-order: bysource
     :show-inheritance:
+
+Miscellaneous
+=============
+
+WaveDrom
```
imphil: Please call this already "A signal tracer for WaveDrom". The section title otherwise sounds like we provide wavedrom here, which we don't.

cocotb/wavedrom.py
```diff
@@ -124,19 +115,17 @@ def get(self, add_clock=True):
...
-
-    Arguments are an arbitrary number of signals or busses to trace.
+    """Context manager to enable tracing of signals.
 
-    We also require a clock to sample on, passed in as a keyword argument
+    Arguments are an arbitrary number of signals or buses to trace.
+    We also require a clock to sample on, passed in as a keyword argument.
 
-    Usage:
+    Usage::
```
imphil: Is that change intended?

documentation/source/library_reference.rst
cmarqu: Yes, it is intended, with this the following code will be shown as code properly.

cmarqu: Argh, replied to the wrong comment. I will change the heading of course.

documentation/source/library_reference.rst
```diff
@@ -269,3 +269,14 @@ XGMII
...
+--------
+
+.. automodule:: cocotb.wavedrom
+    :members:
+    :member-order: bysource
+    :synopsis: A signal tracer for WaveDrom.
```
eric-wieser: How does this piece render?

cmarqu: Like this: https://cocotb-cmarqu.readthedocs.io/en/add-wavedrom-to-lib-ref/library_reference.html#miscellaneous

eric-wieser: So, this line does not render at all?

cmarqu: Ah, the synopsis line you mean... This is needed and apparently only used for the Module Index: https://cocotb-cmarqu.readthedocs.io/en/add-wavedrom-to-lib-ref/py-modindex.html

documentation/source/library_reference.rst
cmarqu: Oh, I had forgotten about that... I'll be without access to a computer until the end of the next week unfortunately.

Issue #1064, opened 2019-08-13
Riviera-PRO Windows: Cannot load the "libvpi.dll" library  /dependency issues


mikep996:
There is a problem (title) to get working cocotb if you follow the official instructions:
  https://cocotb.readthedocs.io/en/latest/quickstart.html#installing-cocotb
32-bit tools on windows10 64 bit environment.
The  libvpi.dll exist but is corrupted.

I was able to run cocotb it if the libvpi/vhpi.dll has been compiled by gcc version provided with Riviera-PRO.

Is any matter from which the python27.dll is used?
Should I also define (change order) on PATH variable?
thanks

imphil:
These instructions are meant only for Modelsim 32 bit. Is Riviera and your compiler really a 32 bit application?

mikep996:
Yes, I am using Riviera-PRO 2019.4 32-bit version:
  vsim -ver
  Aldec, Inc. Riviera-PRO version 2019.04.134.7332 built for Windows on May 10, 2019.
Looks the official cocotb instructions are misleading. 

(BTW. If you try to follow instructions and use 'make' command without any arguments you get the info about the lack of Icarus simulator.)

The cocotb quick start provide link to mingw 6.3.0.
Riviera-PRO use own gcc 4.8.4 (Modelsim;  4.2.1)
After manually compiling all dlls. with gcc4.8.4 from Riviera-PRO I was able to execute the endian_swapper example.
This is creepy workaround only. I am looking something more automated.

thanks a lot

cmarqu:
@mikep996 There is a actually an open pull request to move this part of the installation instructions into https://github.com/cocotb/cocotb/wiki/Tier-2-Setup-Instructions
Feel free to mention your workaround in a subsection there, maybe someone with Windows experience can then expand/automate it more.

mikep996:
done.

Issue #498, opened 2016-11-17
Windows 10 64-bit + Cocotb instruction


4ong:
It works well with
- http://msys2.github.io/
- MSYS + internal make  ( >pacman -S make )
- Windows GCC 5.1.0 64 bit from here - http://tdm-gcc.tdragon.net/
- Anaconda Python3 64-bit ( after >conda install mingw libpython )

And small fix in Makefile.pylib.Msys (Python3 print syntax)
https://gist.github.com/OrionBytes/fa4b46f956c5dc4637fd45f2e8c768a4

I'm not sure is it optimal solution or not - but it seems working with QuestaSim

OrionBytes:
And it works with Intel Python - here is recipe for Anaconda
https://software.intel.com/en-us/articles/using-intel-distribution-for-python-with-anaconda

Visual Studio supports both Python(PTVS) and GCC (plugins).
In addition it's possible to use Performance Analysis from Intel in case of intensive verification flow
https://software.intel.com/en-us/videos/performance-analysis-of-python-applications-with-intel-vtune-amplifier

LinuxFanatic777:
Thank you, interesting info!
I plan to use cocotb also on Windows 64 bit, because developers are using Windows while I prefer to run cocotb on Linux. Glad to see that Windows OS is also supported.

LinuxFanatic777:
I trying to reproduce your success. Makefile.pylib.Msys fix was helpful. I'm using Windows 7 64-bit and Modelsim Starter Edition which is 32-bit. Due to 32 bit of modelsim I have to use msys2 32-bit + Anaconda 3 32-bit. It builds libcocotbutils.dll and then I see the following error:

```
make results.xml
make[1]: Entering directory '/f/tests/test_cocotb'
echo "if [file exists work] {vdel -lib work -all}" > sim_build/runsim.do
echo "vlib work" >> sim_build/runsim.do
echo "vmap -c" >> sim_build/runsim.do
echo "vmap work work" >> sim_build/runsim.do
echo "vlog -work work +define+COCOTB_SIM -timescale 1ns/100ps -mfcu +acc=rmb -sv  /f/tests/test_cocotb/top_cocotb.v" >> sim_build/runsim.do
echo "vsim -onfinish stop -pli libvpi.dll work.top_cocotb" >> sim_build/runsim.do
echo "add wave -recursive /*" >> sim_build/runsim.do
make -C /c/work/cocotb-master/lib/gpi_log SIM=modelsim
[...]

Issue #436, opened 2016-04-12
not working on Windows


LinuxFanatic777:
No problems when running on Linux, but now I have to use Windows Server 2012 R2 Datacenter 64-bit.
I'm using Anaconda 32-bit (Python), ModelSim Altera Started Edition, cocotb from git of version 1.0.
Tried both MinGW 32-bit from Qt 5.6.0 installation and separate MinGW from SourceForge: the result is same, please see log:

```
python : Python 2.7.11 :: Anaconda 4.0.0 (32-bit)
At C:\Users\user\work\src\tests\testbench__project\run.ps1:5 char:1
+ python -V
+ ~~~~~~~~~
    + CategoryInfo          : NotSpecified: (Python 2.7.11 :: Anaconda 4.0.0 (32-bit):String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError

mingw32-make : process_begin: CreateProcess(NULL, uname, ...) failed.
At C:\Users\user\work\src\tests\testbench__project\run.ps1:12 char:1
+ mingw32-make SIM=modelsim
+ ~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (process_begin: ...e, ...) failed.:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError

process_begin: CreateProcess(NULL, uname, ...) failed.

...
**here is many many times of "CreateProcess(NULL, uname, ...) failed"**
...

process_begin: CreateProcess(NULL, uname, ...) failed.
C:/Users/user/work/w/cocotb/makefiles/Makefile.pylib.:35: C:/Users/user/work/w/cocotb/makefiles/Makefile.pyli
b.: Too many open files

process_begin: CreateProcess(NULL, uname -m, ...) failed.
process_begin: CreateProcess(NULL, uname -m, ...) failed.
process_begin: CreateProcess(NULL, uname, ...) failed.
process_begin: CreateProcess(NULL, uname, ...) failed.
process_begin: CreateProcess(NULL, uname, ...) failed.
process_begin: CreateProcess(NULL, uname, ...) failed.
C:\Users\user\work\src\tests\testbench__project/../../../w/cocotb/makefiles/Makefile.inc:86: *** "Unsup
ported os " .  Stop.

```

How to use cocotb on Windows???
**Makefile.inc:86: **\* "Unsupported os " .  Stop.** - why?

leftink:
@LinuxFanatic777 
[...]

Issue #428, opened 2016-04-01
Cocotb on Windows


Paebbels:
Hello,

We are planning to integrate Cocotb into our [open-source IP-core library PoC](https://github.com/VLSI-EDA/PoC). Currently, we are supporting classic VHDL testbenches for various simulators and platforms. Cocotb states that it also supports Windows tool chains, but I encountered several problems....

First off all, MinGW32 can not be used because, Windows, QuestaSim and Python are 64-bit installations. Even compiling in 64-bit mode results in linking processes to i686 dlls from MinGW32...

So I installed MinGW64/MSys2 and added several packages. I modified some of Cocotb's makefiles and got it to compile all required dlls. I think I can provide a PR for it soon.

Finally I git stuck at this dll loading error in QuestaSim:

```
Paebbels@ite190 MINGW64 /d/git/PoC/lib/cocotb/examples/dff/tests
$ make SIM=questa TOPLEVEL_LANG=verilog
make results.xml
make[1]: Verzeichnis „/d/git/PoC/lib/cocotb/examples/dff/tests“ wird betreten
cd sim_build && LD_LIBRARY_PATH=/d/git/PoC/lib/cocotb/build/libs/x86_64::/c/msys64/usr/lib/python2.7/config:/c/msys64/usr/lib/python2.7/config:/c/msys64/usr/lib/python2.7/config:/c/msys64/usr/lib/python2.7/config && PATH=/usr/bin:/mingw64/bin:/usr/local/bin:/usr/bin:/bin:/c/Windows/System32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0/:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/c/Mentor/QuestaSim64/10.4c/:/c/Mentor/QuestaSim64/10.4c/win64 MODULE=dff_cocotb TESTCASE= TOPLEVEL=dff \
GPI_EXTRA= TOPLEVEL_LANG=verilog PYTHONPATH=/d/git/PoC/lib/cocotb/build/libs/x86_64:/d/git/PoC/lib/cocotb:/d/git/PoC/lib/cocotb/examples/dff/tests: \
vsim -c -do runsim.do 2>&1 | tee sim.log
Reading C:/Mentor/QuestaSim64/10.4c/tcl/vsim/pref.tcl

#10.4c

# do runsim.do
[...]

PR #1099, opened 2019-08-29
Better Windows support

themperek:
Resolves: #1064 #428 #498 #436 #354 

## Usage:
- (For Aldec) Do not install C++ compiler together with Riviera or remove/rename `mingw` directory from the install directory
- Install [Miniconda](https://conda.io/en/latest/miniconda.html)  (according to your simulation version 32/64bit)
- Start `Anaconda Prompt` 
-  Install msys `conda install -c msys2 m2-base m2-make libpython m2w64-toolchain` 
- Set proper `%path%` to the binary of the simulator

For msys2 install [msys2](https://www.msys2.org). Install `pacman -S --needed base-devel mingw-w64-i686-python3-pip` or `pacman -S --needed mingw-w64-x86_64-toolchain mingw-w64-x86_64-python3-pip`

## Test matrix on Windows 10 (64):

|            | icarus 10.1,64 | questa 10.7,64 | questa 10.7,32 | Rivera 2019.04,64 | Rivera 2019.04,32 | Modelsim 10.1b,32 |
|------------|----------------|----------------|----------------|-------------------|-------------------|-------------------|
| conda(3.7) | ok             | ok             | ok             | ok               | ok                | ok                |
| conda(2.7) | ok             | ok             | ok             | ok               | ok                | ok                |
| msys(3.7)   | ok             | ok             | ok             | ok               | ok                | ok                |

If this is acceptable will update documentation.

PS. Thanks to [Aldec](https://www.aldec.com/) for providing the license.

### DIFF
--- a/cocotb/share/makefiles/Makefile.pylib.Msys
+++ b/cocotb/share/makefiles/Makefile.pylib.Msys
@@ -29,17 +29,9 @@
 
 # All common python related rules
 
-# if not explicitly set, auto-detect python dir from system path
-ifeq ($(PYTHON_DIR),)
-	PYTHON_DIR ?= $(shell dirname $(shell which python))
-endif
+PYTHON_BIN?=python.exe
 
-# make sure python dir was set properly
-ifeq ($(PYTHON_DIR),)
-	$(error "Path to Python directory must be included in system path or defined in PYTHON_DIR environment variable")
-endif
-
-PYTHON_BIN?=$(PYTHON_DIR)/python.exe
+PYTHON_DIR := $(shell dirname $(shell which $(PYTHON_BIN)))
 
 # We might work with other Python versions
 LOCAL_PYTHON_VERSION?=$(shell $(PYTHON_BIN) -c 'import distutils.sysconfig; print( distutils.sysconfig.get_python_version() )')
@@ -54,5 +46,31 @@ empty :=
 PYTHON_VERSION=$(subst $(period),$(empty),$(LOCAL_PYTHON_VERSION))
 PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'import distutils.sysconfig; print( distutils.sysconfig.get_python_inc() )')
 
-PYLIBS = -lPython$(PYTHON_VERSION)
-PYTHON_DYN_LIB = Python$(PYTHON_VERSION).dll
+# Convert `libpython.so` to `python`
+SYSCONFIG_LDLIBRARY = $(shell $(PYTHON_BIN) -c 'from distutils import sysconfig; print( sysconfig.get_config_var("LDLIBRARY") )')
+ifeq ($(SYSCONFIG_LDLIBRARY),None)
+    PYLIBS := -lPython$(PYTHON_VERSION)
+else
+    PYLIBS:=$(shell $(PYTHON_BIN) -c 'from distutils import sysconfig; import os; print("-l"+os.path.splitext(sysconfig.get_config_var("LDLIBRARY"))[0][3:])')
+endif
+
+# Some simulators accept only paths in form of "c:/../.."
+ifneq ($(VERILOG_SOURCES),)
+    VERILOG_SOURCES := $(shell cygpath -m $(VERILOG_SOURCES))
+endif
+ifneq ($(VHDL_SOURCES),)
+    VHDL_SOURCES := $(shell cygpath -m $(VHDL_SOURCES))
+endif
+
+PWD ?= $(shell pwd)
+
+# sys.path in form "/c/../:/c/.."
+PYTHON_SYS_PATH := $(shell $(PYTHON_BIN) -c "import sys, os; print(':'.join(['/'+dir.replace('\\\ ',' ').replace(os.sep,'/').replace(':','') for dir in sys.path]))")
+# escape "() " with "\"
+PYTHONPATH_ESCAPE := $(shell echo "$(PYTHON_SYS_PATH)" | sed 's/(/\\(/g' | sed 's/)/\\)/g' | sed 's/ /\\ /g')
+PYTHONPATH := $(PYTHONPATH_ESCAPE)
+
+PYTHONHOME:=$(shell $(PYTHON_BIN) -c 'from distutils import sysconfig; print( sysconfig.get_config_var("prefix") )')
+# escape "() " with "\"
+PATH_ESCAPE := $(shell echo "$(PATH)" | sed 's/(/\\(/g' | sed 's/)/\\)/g' | sed 's/ /\\ /g')
+LIB_LOAD := PATH=$(LIB_DIR):$(PATH_ESCAPE) PYTHONHOME=$(PYTHONHOME)
--- a/cocotb/share/makefiles/simulators/Makefile.aldec
+++ b/cocotb/share/makefiles/simulators/Makefile.aldec
@@ -45,7 +45,7 @@ endif
 ifeq (, $(CMD))
     $(error "Unable to locate command >$(CMD_BIN)<")
 else
-	ALDEC_BIN_DIR := $(shell dirname $(CMD))
+    ALDEC_BIN_DIR := $(shell dirname $(CMD))
     export ALDEC_BIN_DIR
 endif
 
@@ -144,36 +144,14 @@ ifeq ($(COVERAGE),1)
 endif
 endif
 
-ifeq ($(OS),Msys)
-
-# Windows allows the situation where the libstc++ used at link time as
-# specified by -L can be different to the one that is used at runtime which
-# comes from the first libstdc++ that it finds in the path. As such
-# we use the mingw lib used at build time and put this at the start of the path
-# before running
-
-MINGW_BIN_DIR = $(shell dirname $(shell which gcc))
-
-EXTRA_LIBS := -laldecpli
-EXTRA_LIBDIRS := -L$(ALDEC_BIN_DIR)
-OLD_PATH := $(shell echo "$(PATH)" | sed 's/(/\\(/g' | sed 's/)/\\)/g' | sed 's/ /\\ /g')
-LIB_LOAD = PATH=$(MINGW_BIN_DIR):$(OLD_PATH):$(LIB_DIR)
-NEW_PYTHONPATH := $(shell echo "$(PYTHONPATH)" | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):\//\/\1\//g' -e 's/;/:/g')
-else
 EXTRA_LIBS = -laldecpli
-ifeq ($(ARCH),i686)
-EXTRA_LIBDIRS = -L$(ALDEC_BIN_DIR)/Linux
-else
-EXTRA_LIBDIRS = -L$(ALDEC_BIN_DIR)/Linux64
-endif
-LIB_LOAD := LD_LIBRARY_PATH=$(LIB_DIR):$(LD_LIBRARY_PATH)
-NEW_PYTHONPATH := $(PYTHONPATH)
-endif
+EXTRA_LIBDIRS := -L$(ALDEC_BIN_DIR)
+
 
 # Note it's the redirection of the output rather than the 'do' command
 # that turns on batch mode (i.e. exit on completion/error)
 $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/runsim.tcl $(COCOTB_LIBS) $(COCOTB_VHPI_LIB) $(COCOTB_VPI_LIB) $(CUSTOM_SIM_DEPS)
-	set -o pipefail; cd $(SIM_BUILD) && PYTHONPATH=$(LIB_DIR):$(PWD):$(NEW_PYTHONPATH) GPI_EXTRA=$(GPI_EXTRA) TOPLEVEL_LANG=$(TOPLEVEL_LANG) \
+	set -o pipefail; cd $(SIM_BUILD) && PYTHONPATH=$(LIB_DIR):$(PWD):$(PYTHONPATH) GPI_EXTRA=$(GPI_EXTRA) TOPLEVEL_LANG=$(TOPLEVEL_LANG) \
 	$(LIB_LOAD) MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) COCOTB_SIM=1 $(CMD) -do runsim.tcl | tee sim.log
 
 clean::
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -34,6 +34,7 @@
 #include "embed.h"
 #include "../compat/python3_compat.h"
 #include "locale.h"
+#include <unistd.h>
 
 #if defined(_WIN32)
 #include <windows.h>
--- a/cocotb/share/makefiles/Makefile.pylib.Darwin
+++ b/cocotb/share/makefiles/Makefile.pylib.Darwin
@@ -53,3 +53,5 @@ PYLIBS:=$(shell $(PYTHON_BIN) -c 'from __future__ import print_function; from di
 
 PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'from __future__ import print_function; import distutils.sysconfig; print(distutils.sysconfig.get_python_inc())')
 PYTHON_DYN_LIB:=$(shell $(PYTHON_BIN) -c 'from __future__ import print_function; from distutils import sysconfig; print(sysconfig.get_config_var("LDLIBRARY"))')
+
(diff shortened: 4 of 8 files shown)


cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -54,5 +54,30 @@ empty :=
...
-PYTHON_DYN_LIB = Python$(PYTHON_VERSION).dll
+SYSCONFIG_LDLIBRARY = $(shell $(PYTHON_BIN) -c 'from distutils import sysconfig; print( sysconfig.get_config_var("LDLIBRARY") )')
+
+ifeq ($(SYSCONFIG_LDLIBRARY),None)
+    PYLIBS := -lPython$(PYTHON_VERSION)
+else
+    PYLIBS:=$(shell $(PYTHON_BIN) -c 'from distutils import sysconfig; import os; print("-l"+os.path.splitext(sysconfig.get_config_var("LDLIBRARY"))[0][3:])')
+endif
+
+# Some simulator accept only paths in form of "c:/../.."
```
cmarqu: Very minor: "Some simulator**s**"

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -54,5 +54,30 @@ empty :=
...
+ifneq ($(VHDL_SOURCES),)
+    VHDL_SOURCES := $(shell cygpath -m $(VHDL_SOURCES))
+endif
+
+PWD ?= $(shell pwd)
+
+# sys.path in form "/c/../:/c/.."
+PYTHON_SYS_PATH := $(shell $(PYTHON_BIN) -c "import sys, os; print(':'.join(['/'+dir.replace('\\\ ',' ').replace(os.sep,'/').replace(':','') for dir in sys.path]))")
+PYTHONPATH_ESCAPE := $(shell echo "$(PYTHON_SYS_PATH)" | sed 's/(/\\(/g' | sed 's/)/\\)/g' | sed 's/ /\\ /g')
+PYTHONPATH := $(PYTHONPATH_ESCAPE)
```
eric-wieser: Can't you do this escaping within the python? I don't understand what's going on here

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -54,5 +54,30 @@ empty :=
...
+    VERILOG_SOURCES := $(shell cygpath -m $(VERILOG_SOURCES))
+endif
+ifneq ($(VHDL_SOURCES),)
+    VHDL_SOURCES := $(shell cygpath -m $(VHDL_SOURCES))
+endif
+
+PWD ?= $(shell pwd)
+
+# sys.path in form "/c/../:/c/.."
+PYTHON_SYS_PATH := $(shell $(PYTHON_BIN) -c "import sys, os; print(':'.join(['/'+dir.replace('\\\ ',' ').replace(os.sep,'/').replace(':','') for dir in sys.path]))")
```
eric-wieser: ':' is better spelt `os.sep`

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: Yes. Also prefer python ;-). It was there before like this. Atlully I am not sure if not add some command to `cooctb-config` to return all of this (would make this less ugly)?

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -54,5 +54,30 @@ empty :=
...
+PYTHONPATH_ESCAPE := $(shell echo "$(PYTHON_SYS_PATH)" | sed 's/(/\\(/g' | sed 's/)/\\)/g' | sed 's/ /\\ /g')
+PYTHONPATH := $(PYTHONPATH_ESCAPE)
+
+PYTHONHOME:=$(shell $(PYTHON_BIN) -c 'from distutils import sysconfig; print( sysconfig.get_config_var("prefix") )')
+PATH_ESCAPE := $(shell echo "$(PATH)" | sed 's/(/\\(/g' | sed 's/)/\\)/g' | sed 's/ /\\ /g')
+LIB_LOAD := PATH=$(LIB_DIR):$(PYTHONPATH_ESCAPE):$(PATH_ESCAPE) PYTHONHOME=$(PYTHONHOME)
```
eric-wieser: Why are we putting `sys.path` in `PATH`? The python include path and the system binary path are very different things, aren't they.

themperek: This is a good question but it does not work other way.

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: I tried to do this in line with python but it started to look even more ugly. I have added a comment. If you have some good idea let me know.

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: This actully does not work. `os.pathsep` is `;` on windows and for whatever reason it does not like it.

eric-wieser: Huh, I wasnt partying attention here, : is correct.

cocotb/share/makefiles/Makefile.pylib.Msys
eric-wieser: This whole operation still seems wrong to me, why are we copying the python import path to the system binary path?

eric-wieser: > It was there before like this

I completely missed this when reviewing this code. It might be worth adding to the commit message that much of this code was just moved

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -29,18 +29,18 @@
...
+    PYTHON_DIR ?= $(shell dirname $(shell which $(PYTHON_BIN)))
 endif
 
 # make sure python dir was set properly
 ifeq ($(PYTHON_DIR),)
-	$(error "Path to Python directory must be included in system path or defined in PYTHON_DIR environment variable")
+    $(error "Path to Python directory must be included in system path or defined in PYTHON_DIR environment variable")
 endif
 
-PYTHON_BIN?=$(PYTHON_DIR)/python.exe
```
eric-wieser: This doesn't look right any more - if someone passes `PYTHON_DIR` explicitly, they end up with `PYTHON_DIR` and `PYTHON_BIN` pointing to different installations.

Is `PYTHON_DIR` used at all after this change?

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -54,5 +54,31 @@ empty :=
...
 PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'import distutils.sysconfig; print( distutils.sysconfig.get_python_inc() )')
 
-PYLIBS = -lPython$(PYTHON_VERSION)
-PYTHON_DYN_LIB = Python$(PYTHON_VERSION).dll
+SYSCONFIG_LDLIBRARY = $(shell $(PYTHON_BIN) -c 'from distutils import sysconfig; print( sysconfig.get_config_var("LDLIBRARY") )')
+
+ifeq ($(SYSCONFIG_LDLIBRARY),None)
+    PYLIBS := -lPython$(PYTHON_VERSION)
+else
+    PYLIBS:=$(shell $(PYTHON_BIN) -c 'from distutils import sysconfig; import os; print("-l"+os.path.splitext(sysconfig.get_config_var("LDLIBRARY"))[0][3:])')
```
eric-wieser: Could do with a very brief comment, ``Convert `libpython.so` to `python` ``

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: Are you sure you looking on the right side? I think I corrected this? And it does not seems to used anywhere. Also was hesitating to just `PYTHON_DIR`.

cocotb/share/makefiles/Makefile.pylib.Msys
eric-wieser: I think it should be sufficient to use `pipes.quote` / `shlex.quote`. So:
```define PYTHON_PATH_SOURCE=
import sys
import os
try:
    from shlex import quote
except ImportError:
    from pipes import quote
print(quote(':'.join([
    '/'+dir.replace(os.sep,'/').replace(':','')
    for dir in sys.path
])))
endef

# note single quotes
PYTHONPATH_ESCAPE=$(shell python -c '$(PYTHON_PATH_SOURCE)')

File: cocotb/share/makefiles/Makefile.pylib.Msys (hunk 5 above)
Comment (eric-wieser) [reply cocotb/share/makefiles/Makefile.pylib.Msys]: Sorry - to be clear, my comment is that removing this seems unsafe, since it now does the wrong thing if the user does pass `PYTHON_DIR`.

Since `PYTHON_DIR` is used in only one place (line 49), I'd be inclined to remove the ability for the user to set the variable.

File: cocotb/share/makefiles/Makefile.pylib.Msys [hunk 7]
```diff
@@ -29,18 +29,18 @@
...
 ifeq ($(PYTHON_DIR),)
-	PYTHON_DIR ?= $(shell dirname $(shell which python))
+    PYTHON_DIR ?= $(shell dirname $(shell which $(PYTHON_BIN)))
 endif
 
 # make sure python dir was set properly
 ifeq ($(PYTHON_DIR),)
-	$(error "Path to Python directory must be included in system path or defined in PYTHON_DIR environment variable")
+    $(error "Path to Python directory must be included in system path or defined in PYTHON_DIR environment variable")
 endif
```
eric-wieser: I think this `if` can be removed entirely, and the `ifeq` above should set `PYTHON_DIR` unconditionally.

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: Yes. OK

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: Are we sure we want such script to be embedded in the `Makefile`?

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -29,17 +29,9 @@
...
-endif
-
-# make sure python dir was set properly
-ifeq ($(PYTHON_DIR),)
-	$(error "Path to Python directory must be included in system path or defined in PYTHON_DIR environment variable")
-endif
+PYTHON_BIN?=python.exe
 
-PYTHON_BIN?=$(PYTHON_DIR)/python.exe
+PYTHON_DIR ?= $(shell dirname $(shell which $(PYTHON_BIN)))
```
eric-wieser: I think you want `:=` here

cocotb/share/makefiles/Makefile.pylib.Msys
eric-wieser: I think it's fine - the script is very simple, and embedding it on multiple lines is better than wrapping it all onto one line.

cocotb/share/makefiles/simulators/Makefile.aldec
```diff
@@ -144,36 +144,14 @@ ifeq ($(COVERAGE),1)
 endif
 endif
 
-ifeq ($(OS),Msys)
-
-# Windows allows the situation where the libstc++ used at link time as
-# specified by -L can be different to the one that is used at runtime which
-# comes from the first libstdc++ that it finds in the path. As such
-# we use the mingw lib used at build time and put this at the start of the path
-# before running
```
eric-wieser: What happened to this comment? Was it wrong?

themperek: On all configurations, I tested it was not needed. I also would consider this a user issues with wrong `PATH` settings. In `conda` and `msys2` prompts the path to `gcc` is set at the beginning of `PATH` and there is no need to manipulate this. I do not know how this was in mingw. I would guess the same issue will be on any other platform.

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: OK.
What you think about adding `cocotb-config --pythonpath` and `cocotb-config --path` ?

cmarqu: This has the advantage that you can have a verbose/debug mode that explains what environment it detected etc.

themperek: Any conclusion about what would be best:

1. leave as is/was (`sed 's/(/\\(/g' | sed 's/)/\\)/g' | sed 's/ /\\ /g')`)
2. add a few line python script to Makefile
3. add `cocotb-config --pythonpath` and `cocotb-config --path`

@imphil Maybe you could pick.

imphil:
Thanks @themperek for starting this. Before we go into the details, I have a couple questions/assumptions to confirm. (It helps to start with documentation for PRs like this to clearly outline the scope of the proposed change.)

To summarize my understanding of our recommended/supported Windows setup:

- Windows 10 (and Windows 7?) x64 only
- 32 and 64 bit simulators and compilers, needs to be consistent. 64 bit recommended, 32 bit only for free version of Modelsim?
- MinGW only (no MSVC)
- Python 3 only
- MinGW builds of Python only. Official Windows Python builds are not supported (they us MSVC as documented at https://wiki.python.org/moin/WindowsCompilers)

Questions:
- Do I understand correctly that you can either use anaconda, or msys2 + their respective package managers?
- Are we aiming for msys2 or mingw? I.e., do we require the POSIX-compat layer of msys2 (most visible: Unix-style path names), or are we OK with only mingw as compiler, but native Windows binaries and paths (C:/...)? That's then also visible in sys.platform, for example.

Remarks:
- Need to list tested simulators under Windows.
- There is some experimental code on setting up all of this on Windows is in https://github.com/cocotb/cocotb/pull/726
- We should help users setting up a supported version of Icarus Verilog in the same way we use it in CI.
  Should we use the one from anaconda (https://anaconda.org/conda-forge/iverilog), or the "official build" from http://bleyer.org/icarus/
- Whatever we recommend, we should do exactly the same in CI. I'll do a PR to enable Azure Pipelines shortly, and we can build on top of that for Windows CI.
- We need to document the constraints/requirements we have on Windows, e.g. regarding compiler selection. 
- Python 2.7 is on its way out. I see no point in documenting and adding CI for a Python version which will be deprecated in a couple months.

themperek:
@imphil 

>Windows 10 (and Windows 7?) x64 only

I can only test on Windows10 (can try to find 7 somewhere) , x64 only


>32 and 64 bit simulators and compilers, needs to be consistent. 64 bit recommended, 32 bit only for free version of Modelsim?

Yes

>MinGW only (no MSVC)

Yes

>Python 3 only

We can do 2.7. It does not make troubles yet. Consider that EDA is "bit" behind normally maybe we should still keep it.

>MinGW builds of Python only. Official Windows Python builds are not supported (they us MSVC as documented at https://wiki.python.org/moin/WindowsCompilers)

It seems to work with msys2 and anaconda (cannot find what is anaconda using). Did not check with official one (with Anaconda is much easier).

>Do I understand correctly that you can either use anaconda, or msys2 + their respective package managers?

Yes

>Are we aiming for msys2 or mingw?

If I would have to decide I would go for Anaconda only.

>Need to list tested simulators under Windows.
>There is some experimental code on setting up all of this on Windows is in #726
>We should help users setting up a supported version of Icarus Verilog in the same way we use it in CI.
>We need to document the constraints/requirements we have on Windows, e.g. regarding compiler selection.

Will do with documentation update. Would prefer separate PR but can do here?

>Should we use the one from anaconda (https://anaconda.org/conda-forge/iverilog), or the "official build" from http://bleyer.org/icarus/

Good point only tried "official build" but for beginner, conda is probably easier.

>Whatever we recommend, we should do exactly the same in CI. I'll do a PR to enable Azure Pipelines shortly, and we can build on top of that for Windows CI.

Yes. Can take care of this.

themperek:
This PR does not fundamentally change anything just simplifies individual simulator Makefiles and cleanup.

imphil:
@themperek this is most likely not necessary any more with the setuptools-based compilation now almost in place?

themperek:
@imphil Some are needed, unfortunately. I was thinking of maybe better use `cocotb-config` to set `PYTHONPATH`, `PATH` and `PYTHONHOME` in a form `cocotb-config --python-path`. In this case the makefiles are simpler and we can do all this strig manipulation in python (that is easier to read and manage).

eric-wieser:
We shouldn't need any of the `PYTHON*` variables once that goes in though, should we?

themperek:
@eric-wieser I think it is needed but I will investigate this.

themperek:
Closed in favour of #1462.

PR #899, opened 2019-04-08
Keep traceback for unhandled exceptions in Python 2

garmin-mjames:
In Python 2, we don't get implicit exception chaining ala the
__context__ attribute of exceptions from Python 3.

When exceptions are raised in coroutines, if the exception is not
handled as it propagates up through the coroutine stack, keep
the traceback information as we go so that the eventual
call to raise_error() can print out the full traceback to where
the exception was raised.

Also added printing of traceback when a forked coroutine raises an
exception and nothing was waiting on a Join.

PR #1100, opened 2019-08-31
Preserve tracebacks correctly in python 2

eric-wieser:
Closes gh-899 using a slightly lower-impact approach - with thanks to @garmin-mjames for drawing attention to the solution


tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -896,17 +898,53 @@ def return_it(x):
...
+        .*
+      File ".*test_cocotb\.py", line \d+, in raise_inner
+        raise ValueError\('It is soon now'\)
+    ValueError: It is soon now""").strip()
+
+
+    if not re.match(expected, tb_text):
+        raise TestFailure(
+            (
+                "Traceback didn't match - got:\n"
```
eric-wieser: Pretty sure this will fail in CI - will push a new rev once I have results to go off

PR #1104, opened 2019-09-09
Add GUI=1 support to Makefile.vcs

THofstee:
I see the behavior exists in the ius/xcelium Makefiles already, so this PR adds the same functionality to vcs.

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.vcs
+++ b/cocotb/share/makefiles/simulators/Makefile.vcs
@@ -57,8 +57,6 @@ endif
 
 ifeq ($(GUI),1)
     EXTRA_ARGS += -gui
-else
-    EXTRA_ARGS +=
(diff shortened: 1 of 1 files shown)


cocotb/share/makefiles/simulators/Makefile.vcs
```diff
@@ -55,6 +55,12 @@ ifeq ($(ARCH),x86_64)
 EXTRA_ARGS += -full64
 endif
 
+ifeq ($(GUI),1)
+    EXTRA_ARGS += -gui
+else
+    EXTRA_ARGS +=
+endif
```
imphil: You don't need the else part, please remove it.

THofstee: Done.

Issue #657, opened 2018-09-01
ability to force/freeze signal


s3bs:
rather than just being able to deposit signals, it is helpful to force signals inside a design. 
Would be great if cocotb would support that.
Maybe as an signalname attribute.
`dut.signalname._force = 1`

lukedarnell:
@s3bs we have a change that implements force, freeze, deposit and release.

The syntax is:
` dut.stream_in_ready <= Force(1)`
` dut.stream_in_ready <= Release()`
`dut.stream_in_ready <= Deposit(BinaryValue('x'))`
`dut.stream_in_ready <= Freeze()`

@chiggs @stuarthodgson Does that syntax make sense?

lukedarnell:
Ok, I will do a pull request with the changes.

cmarqu:
Duplicate of #472.

cmarqu:
Force/release would allow us to build a tool for what is called "testbench qualification" or "mutation coverage" - inject an error into the design, run a regression test and check that at least one of the tests finds the injected error. If the error is not detected, the testbench is too sloppy.

We could start by injecting stuck-at faults at the DUT ports (I have a pull request coming up for providing ``_is_port`` and ``_port_direction`` attributes for discovered objects).

This could be a good project for something like Google's Summer of Code.

ashmanskas:
If I understand correctly, force/release would also make it possible (by overriding connected inputs) to embed a unit test of a single module into a testbench whose main job is to test the larger design, which I would find quite appealing. 
If I understand correctly, unit-testing a non-top-level module now requires a separate Makefile, a separate invocation of cocotb (hence a separate scoreboard printout), and for all practical purposes a separate directory in which to run it.

tonywmnix:
Is there a pull request somewhere for this? This would be helpful to me..

Somewhat related, is there a way to issue tcl commands to the ucli prompt in VCS? Something like:

`cocotb.send_tcl_command("force dut.rst_n 1'b0")`

imphil:
> Is there a pull request somewhere for this? This would be helpful to me..

no, but please feel free to do one yourself!

> Somewhat related, is there a way to issue tcl commands to the ucli prompt in VCS? Something like:

No, it's not. cocotb interacts only through VPI (and similar interfaces) with the simulator; so you'd need to find the equivalent VPI function to force a signal.

lukedarnell:
@imphil @tonywmnix we do have changes with this functionality implemented as documented earlier in this issue.
[...]

PR #1107, opened 2019-09-14
Add vpi Net type support

themperek:
One can distinguish between `NET` and `REG `when using vpi. 

Used for error injection.

### DIFF
--- a/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -77,6 +77,8 @@ gpi_objtype_t to_gpi_objtype(int32_t vpitype)
     switch (vpitype) {
         case vpiNet:
         case vpiNetBit:
+            return GPI_NET;
+
         case vpiReg:
         case vpiRegBit:
         case vpiMemoryWord:
@@ -308,7 +310,7 @@ GpiObjHdl* VpiImpl::native_check_create(int32_t index, GpiObjHdl *parent)
         writable.push_back('\0');
 
         new_hdl = vpi_handle_by_name(&writable[0], NULL);
-    } else if (obj_type == GPI_REGISTER || obj_type == GPI_ARRAY || obj_type == GPI_STRING) {
+    } else if (obj_type == GPI_REGISTER || obj_type == GPI_NET || obj_type == GPI_ARRAY || obj_type == GPI_STRING) {
         new_hdl = vpi_handle_by_index(vpi_hdl, index);
 
         /* vpi_handle_by_index() doesn't work for all simulators when dealing with a two-dimensional array.
@@ -379,7 +381,7 @@ GpiObjHdl* VpiImpl::native_check_create(int32_t index, GpiObjHdl *parent)
             }
         }
     } else {
-        LOG_ERROR("VPI: Parent of type %s must be of type GPI_GENARRAY, GPI_REGISTER, GPI_ARRAY, or GPI_STRING to have an index.", parent->get_type_str());
+        LOG_ERROR("VPI: Parent of type %s must be of type GPI_GENARRAY, GPI_REGISTER, GPI_NET, GPI_ARRAY, or GPI_STRING to have an index.", parent->get_type_str());
         return NULL;
     }
 
--- a/tests/test_cases/test_discovery/test_discovery.py
+++ b/tests/test_cases/test_discovery/test_discovery.py
@@ -443,3 +443,33 @@ def custom_type(dut):
     if expected_top != count:
         raise TestFailure("Expected %d found %d for cosLut" % (expected_top, count))
 
+@cocotb.test(skip=cocotb.LANGUAGE in ["vhdl"])
+def type_check_verilog(dut):
+    """
+    Test if types are recognized
+    """
+
+    tlog = logging.getLogger("cocotb.test")
+
+    yield Timer(1)
+
+    test_handles = [
+        (dut.stream_in_ready, "GPI_REGISTER"),
+        (dut.register_array, "GPI_ARRAY"),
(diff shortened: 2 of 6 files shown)


imphil:
Can you elaborate a bit more on what this change enables, and what simulations you tested it with? Is there a way to add a test for this feature?

themperek:
>Can you elaborate a bit more on what this change enables

I want to be able to recognize if something is declared as `wire` or `reg` in simple Verilog or gate-level this allows to distinguish synchronous elements form combinatorial. This allows proper random error testing for [SEEs](https://en.wikipedia.org/wiki/Single-event_upset).  Full functionality will be possible when [force/freeze/release](https://github.com/cocotb/cocotb/issues/657#issuecomment-531094789) will be added. 

>what simulations you tested it with

icarus, ius, questa

>Is there a way to add a test for this feature?

I have added a simple test.

eric-wieser:
> I want to be able to recognize if something is declared as wire or reg in simple Verilog or gate-level this allows to distinguish synchronous elements from combinatorial.

But that's not even what `reg` means, is it?

This code is combinatorial, but uses a `reg`:

```verilog
reg a_or_b;
always @(*) begin
   a_or_b = a || b;
end
```

From what I've seen so far, there is no useful distinction between `wire` and `reg`. Perhaps it becomes relevant for force /freeze/release.

themperek:
@eric-wieser For that reason we use preprocessed verilog where all combinatorial logic uses only `wire` / `assign` (yosys can do this) and gate level netlist.

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -76,6 +76,8 @@ gpi_objtype_t to_gpi_objtype(int32_t vpitype)
     switch (vpitype) {
         case vpiNet:
         case vpiNetBit:
+        	return GPI_NET;
```
imphil: nit: Fix indentation.

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -443,3 +443,28 @@ def _discover(obj):
...
+        (dut.NUM_OF_MODULES, "GPI_PARAMETER"),
+        (dut.temp, "GPI_REGISTER"),
+        (dut.and_output, "GPI_NET"),
+        (dut.stream_in_data, "GPI_NET"),
+    ]
+
+    for handle in test_handles:
+        tlog.info("Handle %s" %  (handle[0]._fullname,))
+        if handle[0]._type != handle[1]:
+                 raise TestFailure("Expected %s found %s for %s" %
```
imphil: Nit: fix indentation

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -443,3 +443,28 @@ def _discover(obj):
...
+    yield Timer(10)
+
+    test_handles = [
+        (dut.stream_in_ready, "GPI_REGISTER"),
+        (dut.register_array, "GPI_ARRAY"),
+        (dut.NUM_OF_MODULES, "GPI_PARAMETER"),
+        (dut.temp, "GPI_REGISTER"),
+        (dut.and_output, "GPI_NET"),
+        (dut.stream_in_data, "GPI_NET"),
+    ]
```
imphil: Can we also have a  test with `logic`?

themperek: I have added. 

`logic` should default to `vpiNet` but for vcs it is becoming `GPI_MODULE` so the test fails (all others seem happy).

imphil: You modified an input port to explicitly specify `logic` as net type. In SV, the default is already `logic` unless overwritten with the default_nettype compiler directive. 

The story is different for variables/nets within the design. Can you add a test for such an element as well (logic, wire and reg)?

If VCS behaves differently, can you exclude the test from VCS (see test_cocotb.py for examples on how to do that). Ideally you can note the version number of VCS that fails to allow us to revisit this exception in the future more easily.

cmarqu:
Is there a way to special-case VCS in the VPI-to-GPI layer so that the VCS specifics are abstracted away in GPI?

imphil:
It's hard for cocotb to know what VCS meant when it returns vpiModule; so while we could think of some workarounds, I don't think it's necessary in this case. And we should certainly not do it before we have reported to Synopsys.

tests/test_cases/test_discovery/test_discovery.py
themperek: If I do this:
```systemverilog
logic logic_a, logic_b, logic_c;
assign logic_a = stream_in_valid;
always@* logic_b = stream_in_valid;
always@(posedge clk) logic_c <= stream_in_valid;
```
In icarus `logic_a` is `vpiNet` and `logic_b`/`logic_c` `vpiReg` in all other simulators it will be all `vpiReg` .

Any opinion on how to handle this?

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -443,3 +443,30 @@ def _discover(obj):
...
 
+@cocotb.test(skip=cocotb.LANGUAGE in ["vhdl"])
+def type_check_verilog(dut):
+    """
+    Test if types are recognized
+    """
+
+    tlog = logging.getLogger("cocotb.test")
+
+    yield Timer(10)
```
imphil: `Timer(1)` should be sufficient and reduce test duration.

tests/test_cases/test_discovery/test_discovery.py
imphil: Please special-case icarus in the code, but test for the existing behavior for both icarus and other simulators. And file an icarus upstream bug and reference the bug number here.

imphil:
Ah, I missed the fact the newsfragment is already there.

PR #1109, opened 2019-09-14
Remove all remaining callers of `raise_error`.

eric-wieser:
This seemed to exist only because tracebacks were not otherwise being printed.

Now that they are, there is little reason to keep this function around.  Summarizing the changes:
* Hooks behave exactly as before, silencing all exceptions and logging tracebacks. Now that the `__call__` method does not need to catch any exceptions, it can be removed entirely in favor of using the method from the base class.
* Failing triggers throw `TriggerException` rather than `TestError`. Since the exception already existed, it seemed appropriate to throw.

### DIFF
--- a/cocotb/result.py
+++ b/cocotb/result.py
@@ -36,6 +36,10 @@ from io import StringIO, BytesIO
 def raise_error(obj, msg):
     """Creates a :exc:`TestError` exception and raises it after printing a traceback.
 
+    Note that since cocotb will print a traceback for you when an exception is
+    caught, you are unlikely to need this - use ``raise TestError(msg)``
+    instead.
+
     Args:
         obj: Object with a log method.
         msg (str): The log message.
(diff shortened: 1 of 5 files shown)


cocotb/result.py
```diff
@@ -36,6 +36,10 @@
 def raise_error(obj, msg):
     """Creates a :exc:`TestError` exception and raises it after printing a traceback.
 
+    Note that since cocotb will print a traceback for you when an exception is
+    caught, you are unlikely to need this - use ``raise TestError(msg)``
+    instead.
+
```
imphil: Can we add a `.. deprecated:: 1.3` here (http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated), and possibly throw a DeprecationWarning`?

Same applies for the `create_error` function below.

eric-wieser: Sure, I suppose there's no reason not to do that. Updated with tests.

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -725,15 +751,8 @@ def test_binary_value_compat(dut):
...
     # Test for the DeprecationWarning when using |bits|
-    with warnings.catch_warnings(record=True) as w:
-        warnings.simplefilter("error")
-
-        try:
-            vec = BinaryValue(value=0, bits=16)
-        except DeprecationWarning:
-            pass
-        else:
-            TestFailure("Expected DeprecationWarning when using bits instead of n_bits.")
```
eric-wieser: `raise` was missing here!

Issue #1102, opened 2019-09-08
VCS Timescale Mismatch?


hofstee:
I have a design that has a clock with a period of 10 (in default cocotb units), and a simulation that runs for 8508 cycles. What I see in the report for VCS is the following:

```
*********************************************************************
** TEST         PASS/FAIL  SIM TIME(NS)  REAL TIME(S)  RATIO(NS/S) **
*********************************************************************
** tb.test_app    PASS     85081000000000.00        [redacted]     **
*********************************************************************
```

Well, that's not right, there are 9 extra 0s. 1ns is 1e-9s, so that's a bit suspicious. What does Incisive say?

```
*********************************************************************
** TEST         PASS/FAIL  SIM TIME(NS)  REAL TIME(S)  RATIO(NS/S) **
*********************************************************************
** tb.test_app    PASS        85081.00        [redacted]           **
*********************************************************************
```

Ok, that looks correct.

What's going on here? Did I misconfigure something or is there something wrong with how cocotb is setting up the VCS testbench?

In any case, it's questionable that the same testbench is not consistent across simulators.

THofstee:
I should add that if I query for the `get_sim_time()` it returns 85081 in both cases.

cmarqu:
Instead of trying to find out and document what the different simulators have as a default timestep, maybe we should just output a warning telling the users to always include the ``units`` argument?

THofstee:
That's a reasonable suggestion, but that raises another issue of setting `'ns'` segfaults VCS with the default makefile.

```
Send raised exception:
Traceback (most recent call last):
  File "cocotb/decorators.py", line 259, in _advance
    return outcome.send(self._coro)
  File "cocotb/outcomes.py", line 37, in send
    return gen.send(self.value)
  File "tb.py", line 182, in test_app
    cocotb.fork(Clock(dut.clk, CLK_PERIOD, 'ns').start())
  File "cocotb/clock.py", line 76, in __init__
    self.period = get_sim_steps(period, units)
  File "cocotb/utils.py", line 122, in get_sim_steps
    time, units, _LOG_SIM_PRECISION))
ValueError: Unable to accurately represent 10(ns) with the simulator precision of 1e0

```

THofstee:
[...]

PR #1113, opened 2019-09-16
Timescale

hofstee:
Addressing #1102 

This adds two new makefile variables:
- `COCOTB_HDL_TIMESTEP`, which defaults to 1ns
- `COCOTB_HDL_TIMEPRECISION`, which defaults to 1ps

I've generally tried to set it before EXTRA_ARGS in places so that if someone specified the timescale in a simulator specific fashion there, I would hope that a tool takes the last one specified.


cmarqu:
In IUS and Xcelium, I'm pretty sure that the simulator would complain about more than one ``-timescale`` option and error out, but this should rarely happen and the user can then just use these new variables.

THofstee:
I've tested it with VCS so far and it seems to behave properly. If you specify the timescale in EXTRA_ARGS then VCS takes that one, which was the behavior I was aiming for at least.

Also fixed issues where I was using COCOTB_HDL_TIMESCALE instead of COCOTB_HDL_TIMESTEP in a few simulator targets. Incisive (and presumably Xcelium) also seem to recognize that timescale was set with the new variables. I tried seeing if the behavior was as expected but couldn't because of server-related issues. I was also going to test what happens with EXTRA_ARGS, but beyond getting a warning about multiple timescale declarations (as you expected) I wasn't able to see which it uses for similar reasons.

I don't have any way of testing questa, our install doesn't work on the machines I use.

documentation/source/building.rst
```diff
@@ -51,6 +51,12 @@ Make Variables
     ``EXTRA_ARGS``
       Passed to both the compile and execute phases of simulators with two rules, or passed to the single compile and run command for simulators which don't have a distinct compilation stage.
 
+    ``COCOTB_HDL_TIMESTEP``
+      The default timestep that should be assumed for simulation when not specified by modules in the design. If this isn't specified then it is assumed to be 1ns.
+
+    ``COCOTB_HDL_TIMEPRECISION``
+      The default precision that should be assumed for simulation when not specified by modules in the design. If this isn't specified then it is assumed to be 1ps.
```
cmarqu: I would say "The default time precision" so that it can't be mistaken for numerical precision.

THofstee:
Tested with Xcelium and Incisive. I've added a check in their respective Makefiles to check if the user has specified `-timescale` and prompts them to use `COCOTB_HDL_TIMESTEP` and `COCOTB_HDL_TIMEPRECISION` instead, as Xcelium and Incisive error out if multiple timescales are specified on the command line.

Hopefully by adding this more descriptive error message, someone that gets bitten by this will have a simpler time figuring out why their tools are complaining.

cmarqu:
> As an aside: I'm not really sure why the Travis builds are failing.

The last row with ``SIM=ghdl`` is failing all the time (and allowed to do so), and the Mac OS fails are a new thing that is affecting other PRs too, so nothing to do with your changes.

documentation/source/building.rst
```diff
@@ -51,6 +51,12 @@ Make Variables
     ``EXTRA_ARGS``
       Passed to both the compile and execute phases of simulators with two rules, or passed to the single compile and run command for simulators which don't have a distinct compilation stage.
 
+    ``COCOTB_HDL_TIMESTEP``
+      The default timestep that should be assumed for simulation when not specified by modules in the design. If this isn't specified then it is assumed to be 1ns.
```
cmarqu: We should probably detail this a bit further:
```suggestion
      The default timestep that should be assumed for simulation when not specified by modules in the design. If this isn't specified then it is assumed to be ``1ns``. Allowed units are ``s``, ``ms``, ``us``, ``ns``, ``ps``, ``fs``.
```

documentation/source/building.rst
```diff
@@ -51,6 +51,12 @@ Make Variables
...
+
+    ``COCOTB_HDL_TIMEPRECISION``
+      The default time precision that should be assumed for simulation when not specified by modules in the design. If this isn't specified then it is assumed to be 1ps.
```
cmarqu:
```suggestion
      The default time precision that should be assumed for simulation when not specified by modules in the design. If this isn't specified then it is assumed to be ``1ps``. Allowed units are ``s``, ``ms``, ``us``, ``ns``, ``ps``, ``fs``.
```

hofstee:
I've added the suggested changes.

imphil:
Thanks for doing this work @hofstee! LGTM with one minor naming nit: Can you rename timestep to timeunit, to be in line with the SystemVerilog spec? Beyond that I think we can take this as it is.

image

documentation/source/building.rst
```diff
@@ -51,6 +51,12 @@ Make Variables
     ``EXTRA_ARGS``
       Passed to both the compile and execute phases of simulators with two rules, or passed to the single compile and run command for simulators which don't have a distinct compilation stage.
 
+    ``COCOTB_HDL_TIMESTEP``
+      The default timestep that should be assumed for simulation when not specified by modules in the design. If this isn't specified then it is assumed to be ``1ns``. Allowed units are ``s``, ``ms``, ``us``, ``ns``, ``ps``, ``fs``.
+
+    ``COCOTB_HDL_TIMEPRECISION``
```
eric-wieser: I think this needs rebasing, these should be using `..make:var::` or similar

hofstee:
@imphil updated to latest master and renamed TIMESTEP to TIMEUNIT. I also changed the docs to say "time unit" instead of "timestep" but I wonder if there's a less ambiguous way of describing it.

documentation/source/building.rst
```diff
@@ -62,6 +62,14 @@ Make Variables
 
       Passed to both the compile and execute phases of simulators with two rules, or passed to the single compile and run command for simulators which don't have a distinct compilation stage.
 
+.. make:var:: COCOTB_HDL_TIMEUNIT
+
+      The default time unit that should be assumed for simulation when not specified by modules in the design. If this isn't specified then it is assumed to be ``1ns``. Allowed units are ``s``, ``ms``, ``us``, ``ns``, ``ps``, ``fs``.
+
+.. make:var:: COCOTB_HDL_TIMEPRECISION
+
+      The default time precision that should be assumed for simulation when not specified by modules in the design. If this isn't specified then it is assumed to be ``1ps``. Allowed units are ``s``, ``ms``, ``us``, ``ns``, ``ps``, ``fs``.
```
imphil: There's one more restriction: allowed values (orders of magnitude) are only 1, 10, or 100, for both timeunit and timeprecision. 

You could add an example like `timescale COCOTB_HDL_TIMEUNIT/COCOTB_HDL_TIMEPRECISION this should make it rather obvious what is meant to a HDL engineer, even if they have not heard the naming as it is used in the Verilog spec.

imphil:
I've updated and rebased this PR and pushed it to master. Thanks for your contribution!

imphil:
599b21ba453aa5a047b24dec80f2c475523105cc for reference

PR #1114, opened 2019-09-18
Avoid rounding errors in simulation times when using rational numbers

eric-wieser:
Previously the rounding errors inherent in floats containing negative powers of 10 would cause exact rational numbers to be incorrectly rejected.

Both tests in this patch would fail before this change.

This prevents rounding in `Timer(...)`, and overflow in `get_sim_time` if the units are smaller than the simulation timestep. Precision is still lost in `get_sim_time(unit='s')` - fixing this would require choosing `Fraction` or `Decimal` as a return type, which would break existing code.

### DIFF
--- a/cocotb/utils.py
+++ b/cocotb/utils.py
@@ -1,4 +1,4 @@
-from __future__ import print_function
+from __future__ import print_function, division
 
 # Copyright (c) 2013 Potential Ventures Ltd
 # Copyright (c) 2013 SolarFlare Communications Inc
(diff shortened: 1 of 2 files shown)


cocotb/utils.py
```diff
@@ -107,16 +117,17 @@ def get_sim_steps(time, units=None):
     """
     result = time
     if units is not None:
-        result = result * (10.0**(_get_log_time_scale(units) - _LOG_SIM_PRECISION))
+        result = _ldexp10(result, _get_log_time_scale(units) - _LOG_SIM_PRECISION)
 
-    err = int(result) - math.ceil(result)
+    result_rounded = math.floor(result)
```
eric-wieser: Previously this would allow all negative numbers, but require exact matches for positive ones. That didn't look deliberate.

Issue #936, opened 2019-05-11
TestError and expect_error


ktbarrett:
The `cocotb.test` decorator can take an `expect_error` argument. Currently this is a boolean deciding that the given test passed if it got a `TestError`, and failing otherwise. There are many reasons why a test would fail, and usually only one that is actually expected. However, `TestError` muddles all errors, making `expect_error` work if *any* error is caught. Most good regression systems allow you to expect an error and specify what type of exception is anticipated, failing otherwise.

I'd like to see this
```python
class MyException(Exception): pass

@cocotb.coroutine
def throws_my_exception():
    raise MyException

@cocotb.coroutine
def throws_other_exception():
    raise NotImplementedError

@cocotb.test(expect_error=MyException)
def passing_test(dut):
    yield throws_my_exception()

@cocotb.test(expect_error=MyException)
def failing_test(dut):
    yield throws_other_exception()
```

This would require re-writing a good bit of the scheduler and regression manager to not use `TestError` and properly propagate exceptions. I can't see this being too big of a break, user code doesn't/shouldn't be catching or throwing `TestError`.

garmin-mjames:
I don't think it's safe to assume that non-core code wont use `TestError`.
But you could keep backwards compatibility if `expect_error=True` expects a `TestError` object.

ktbarrett:
@garmin-mjames 
You are correct about that. TestError can be used to short-circuit test failure in forked coroutines, a behavior not available in any other way. TestError isn't bad, but how it is being used in the regression manager/scheduler is an issue.

eric-wieser:
> This would require re-writing a good bit of the scheduler and regression manager to not use TestError and properly propagate exceptions. 

I think this might be done? Your proposed change looks very straightforward to me now.

Issue #922, opened 2019-05-06
Improve exception handling in forked coroutines


imphil:
(Spin-off from #893.)

In the long run, i think we need a better approach to handling exceptions in forked coroutines. Right now the following doesn't work as intended:

```py
task = cocotb.fork(a_coroutine_that_fails_soon())
yield longer_than_soon()
try:
    yield task
except Exception as e:
    pass  # we never get as far as here right now, but I think we ought to
else:
    assert False
```

ktbarrett:
In the above example, the failure happens, and the outcome is stored. What happens if the forked coro is never yielded on? Where does the exception go?

eric-wieser:
> In the above example, the failure happens, and the outcome is stored.

That's what _should_ happen, but it's not what _does_ happen

ktbarrett:
I don't agree. `fork`ed coroutines throwing exceptions should causing `TestError`. Many uses of `fork` are meant to start up 'threads' (yes, not great terminology but it mirrors `fork`) that run forever and are never meant to be yielded upon. This is used in many monitors and drivers. If exceptions are captured and the test never yields on the drive/monitor 'threads' (if they are meant to run forever they won't be yielded), those exceptions are never seen. I can see the how the behavior is useful, but I think it should exist in a new interface.

ktbarrett:
One potential way to support both behaviors is to make `scheduler.add` work as listed in the OP, and make `fork` wrap that with a function that yields on the coroutine and throws a `TestError` if an exception is caught. This will maintain the old behavior of `fork` while allowing the new behavior in a new interface that simply aliases `scheduler.add` as `fork` does now.

Also,there would have to be a special case for `TestComplete`s otherwise the test would continue longer than desired.

eric-wieser:
> If exceptions are captured and the test never yields on the drive/monitor 'threads' (if they are meant to run forever they won't be yielded), those exceptions are never seen.

That doesn't have to be true. The test cleanup could collect exceptions from all the remaining "threads".

PR #1116, opened 2019-09-19
Users can specify an expected error type on tests

ktbarrett:
Addresses issue from #936.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2013 Potential Ventures Ltd
 # Copyright (c) 2013 SolarFlare Communications Inc
 # All rights reserved.
-# 
+#
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions are met:
 #     * Redistributions of source code must retain the above copyright
@@ -13,7 +13,7 @@
 #       SolarFlare Communications Inc nor the
 #       names of its contributors may be used to endorse or promote products
 #       derived from this software without specific prior written permission.
-# 
+#
 # THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
 # ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
@@ -437,10 +437,24 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
             value representing simulation timeout (not implemented).
         expect_fail (bool, optional):
             Don't mark the result as a failure if the test fails.
-        expect_error (bool, optional):
-            Don't mark the result as an error if an error is raised.
-            This is for cocotb internal regression use 
-            when a simulator error is expected.
+        expect_error (bool or exception type or tuple of exception types, optional):
+            If ``True``, consider this test passing if it raises *any* :class:`Exception`, and failing if it does not.
+            If given an exception type or tuple of exception types, catching *only* a listed exception type is considered passing.
+            This is primarily for cocotb internal regression use for when a simulator error is expected.
+
+            Users are encouraged to use the following idiom instead::
+
+                @cocotb.test()
+                def my_test(dut):
+                    try:
+                        yield thing_that_should_fail()
+                    except ExceptionIExpect:
+                        pass
+                    else:
+                        assert False, "Exception did not occur"
+
+            .. versionchanged:: 1.3
+                Specific exception types can be expected
         skip (bool, optional):
             Don't execute this test as part of the regression.
         stage (int, optional)
@@ -452,6 +466,10 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
 
         self.timeout = timeout
         self.expect_fail = expect_fail
+        if expect_error is True:
+            expect_error = (Exception,)
+        elif expect_error is False:
+            expect_error = ()
         self.expect_error = expect_error
         self.skip = skip
         self.stage = stage
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -305,7 +305,7 @@ class RegressionManager(object):
             result_pass = False
 
         elif isinstance(result, SimFailure):
-            if test.expect_error:
+            if isinstance(result, test.expect_error):
                 self.log.info("Test errored as expected: " + _result_was())
             else:
                 self.log.error("Test error has lead to simulator shutting us "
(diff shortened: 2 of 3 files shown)


eric-wieser:
One possible argument against doing this - `expect_error` is documented as for internal cocotb testing:
```
        expect_error (bool, optional):
            Don't mark the result as an error if an error is raised.
            This is for cocotb internal regression use 
            when a simulator error is expected.
```

In some ways, users can already write such a test today, except for the complexities of gh-922:
```python
@cocotb.test()
def test_it(dut):
    try:
        # stuff which might fail
    except TheExceptionIExpect:
        pass
    else:
        assert False
```

cocotb/regression.py
```diff
@@ -316,7 +316,12 @@ def _result_was():
                 return
 
         elif test.expect_error:
-            self.log.info("Test errored as expected: " + _result_was())
+            if test.expect_error is True or isinstance(result, test.expect_error):
```
eric-wieser: Could simplify this logic perhaps with a patch like:
```diff
--- C:\Users\wiese\Repos\cocotb\cocotb\decorators.py
+++ C:\Users\wiese\Repos\cocotb\cocotb\decorators.py
@@ -437,7 +437,7 @@
             value representing simulation timeout (not implemented).
         expect_fail (bool, optional):
             Don't mark the result as a failure if the test fails.
-        expect_error (bool, optional):
+        expect_error (bool, Type[Exception], optional):
             Don't mark the result as an error if an error is raised.
             This is for cocotb internal regression use 
             when a simulator error is expected.
@@ -450,6 +450,10 @@
                  skip=False, stage=None):
         super(test, self).__init__(f)
 
+        if expect_error is True:
+            expect_error = (Exception,)
+        elif expect_error is False:
+            expect_error = ()
         self.timeout = timeout
         self.expect_fail = expect_fail
         self.expect_error = expect_error

```

This saves you the `is True` comparison at the call site, and lets you use `isinstance(result, test.expect_error)` unconditionally

ktbarrett: Good idea, will capture the docstring changes as well.

ktbarrett:
@eric-wieser Even if this feature is intended for internal testing only, it will make internal tests better by ensuring we are getting the error we intend to invoke.

eric-wieser:
> Even if this feature is intended for internal testing only,

Agreed - that comment was really just about the PR title / commit message conveying a questionable motive with the word "user"

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1235,5 +1235,20 @@ def test_assertion_is_failure(dut):
...
+    pass
+
+@cocotb.test(expect_error=MyException)
+def test_expect_particular_exception(dut):
+    yield Timer(1)
+    raise MyException()
+
+
+@cocotb.test(expect_error=(MyException, AssertionError))
+def test_expect_exception_list(dut):
```
eric-wieser: Adding AssertionError here is a bit weird - I don't think that expect_error can handle that today?

ktbarrett: Hmmm, not sure. I just added something to ensure a tuple works, I'll change to `ValueError`.

cocotb/decorators.py
```diff
@@ -437,9 +437,9 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
             value representing simulation timeout (not implemented).
         expect_fail (bool, optional):
             Don't mark the result as a failure if the test fails.
-        expect_error (bool, optional):
+        expect_error (Union[bool, Exception], optional):
             Don't mark the result as an error if an error is raised.
-            This is for cocotb internal regression use 
+            This is for cocotb internal regression use
             when a simulator error is expected.
```
eric-wieser: Was just about to merge - but I think this could do with a sentence explaining what passing an exception types or tuple of exceptions means.

Also, strictly this does not take `Union[bool, Exception]` - it takes `Union[bool, Type[Exception], Tuple[Type[Exception, ], ...]]` Perhaps this is better in english after all, as:
```
expect_error (bool or exception type or tuple of exception types, optional):
    If true, consider this test passing if it raises an exception, and failing if it does not.
    If an exception type or tuple of exception types, this gives the specific type of exception that is considered passing.
    This is primarily for cocotb internal regression use when a simulator error is expected.

    For use elsewhere, consider::

        @cocotb.test()
        def my_test(dut):
            try:
                thing_that_should_fail()
            except ExceptionIExpect:
                pass
            else:
                assert False, "Exception did not occur"
```

ktbarrett: Is there is some particular reason this functionality should be reserved for internal use only? I'm think we should just leave the last part off.

eric-wieser: I suppose not. What I'd like to encourage is that `expect_fail` is used either for:

* Tests which ought to pass, but are known to fail and haven't been fixed yet (like `@pytest.mark.xfail`
* Tests internal to cocotb that test scheduler and regression manager errors which can't otherwise be caught (such as your timeout-level stuff)

What I don't think this should be used for is testing that user APIs throw exceptions - you can write much better tests using the try / except formulation (ie checking error messages, multiple exceptions checked in a single test, etc.

For that reason, I'd like to try and include an example like the above.

cocotb/decorators.py
```diff
@@ -437,10 +437,21 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
...
-        expect_error (bool, optional):
-            Don't mark the result as an error if an error is raised.
-            This is for cocotb internal regression use 
-            when a simulator error is expected.
+        expect_error (bool or exception type or tuple of exception types, optional):
+            If ``True``, consider this test passing if it raises `any` :class:`Exception`, and failing if it does not.
+            If given an exception type or tuple of exception types, catching `only` a listed exception type is considered passing.
+            This is primarily for cocotb internal regression use for when a simulator error is expected.
+
+            User are encouraged to use the following idiom instead::
```
```suggestion
            Users are encouraged to use the following idiom instead::
```

cocotb/decorators.py
```diff
@@ -437,10 +437,21 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
...
+
+                @cocotb.test()
+                def my_test(dut):
+                    try:
+                        yield thing_that_should_fail()
+                    except ExceptionIExpect:
+                        pass
+                    else:
+                        assert False, "Exception did not occur"
```
```suggestion
                        assert False, "Exception did not occur"

```

cocotb/decorators.py
```diff
@@ -437,10 +437,21 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
             value representing simulation timeout (not implemented).
         expect_fail (bool, optional):
             Don't mark the result as a failure if the test fails.
-        expect_error (bool, optional):
-            Don't mark the result as an error if an error is raised.
-            This is for cocotb internal regression use 
-            when a simulator error is expected.
+        expect_error (bool or exception type or tuple of exception types, optional):
+            If ``True``, consider this test passing if it raises `any` :class:`Exception`, and failing if it does not.
```
```suggestion
            If ``True``, consider this test passing if it raises *any* :class:`Exception`, and failing if it does not.
```

cocotb/decorators.py
```diff
@@ -437,10 +437,21 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
...
+            If given an exception type or tuple of exception types, catching `only` a listed exception type is considered passing.
```
```suggestion
            If given an exception type or tuple of exception types, catching *only* a listed exception type is considered passing.
```

Backticks are for references.

PR #1118, opened 2019-09-19
Don't register VPI systemtf functions for Verilator

wallento:
Hi,

this does not necessarily need to be merged before the Verilator part is done, but I want to put it to discussion. Surprisingly, the most contentious point is that Verilator does not have vpi_register_systf. Adding it would be a massive task, for the fact that it is just for display-style printfs. So, I am wondering if people would accept this patch. The VERILATOR define is already generated by cocotb itself and the actual Verilator infrastructure would then use DPI and overload display, info, ...

Looking forward to your feedback.

Thanks,
Stefan

### DIFF
--- a/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -592,6 +592,7 @@ static void register_final_callback(void)
     sim_finish_cb->arm_callback();
 }
 
+#ifndef VERILATOR
 // Called at compile time to validate the arguments to the system functions
 // we redefine (info, warning, error, fatal).
 //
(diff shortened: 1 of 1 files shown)


wallento:
The discussion and alternative solution can be found here: https://www.veripool.org/issues/1470-Verilator-VPI-systemtf

Both are very trivial..

themperek:
Maybe a comment in code would be useful for the future.

eric-wieser:
Instead of using an ifdef here  - should we just register the functions unconditionally, but print a warning if doing so returns an error?

wallento:
Thanks, it was actually very easy with the standard methods.

PR #872, opened 2019-03-18
Issue #7: Add test timeout

leftink:
This should be the last thing preventing #7 from being able to be closed.

PR #1119, opened 2019-09-20
Implements general-purpose timeout function and test timeout on decorator

ktbarrett:
Picked up where #872 left off. I sure hope my grammar is right. Addresses #7.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -409,8 +409,14 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
     Used as ``@cocotb.test(...)``.
 
     Args:
-        timeout (int, optional):
-            value representing simulation timeout (not implemented).
+        timeout_time (int, optional):
+            Value representing simulation timeout.
+
+            .. versionadded:: 1.3
+        timeout_unit (str, optional):
+            Unit of timeout value, see :class:`~cocotb.triggers.Timer` for more info.
+
+            .. versionadded:: 1.3
         expect_fail (bool, optional):
             Don't mark the result as a failure if the test fails.
         expect_error (bool or exception type or tuple of exception types, optional):
@@ -436,11 +442,27 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
         stage (int, optional)
             Order tests logically into stages, where multiple tests can share a stage.
     """
-    def __init__(self, f, timeout=None, expect_fail=False, expect_error=False,
+    def __init__(self, f, timeout_time=None, timeout_unit=None,
+                 expect_fail=False, expect_error=False,
                  skip=False, stage=None):
+
+        if timeout_time is not None:
+            co = coroutine(f)
+            @functools.wraps(f)
+            def f(*args, **kwargs):
+                running_co = co(*args, **kwargs)
+                try:
+                    res = yield cocotb.triggers.with_timeout(running_co, self.timeout_time, self.timeout_unit)
+                except cocotb.result.SimTimeoutError:
+                    running_co.kill()
+                    raise
+                else:
+                    raise ReturnValue(res)
+
         super(test, self).__init__(f)
 
-        self.timeout = timeout
+        self.timeout_time = timeout_time
+        self.timeout_unit = timeout_unit
         self.expect_fail = expect_fail
         if expect_error is True:
             expect_error = (Exception,)
--- a/cocotb/result.py
+++ b/cocotb/result.py
@@ -29,8 +29,8 @@
 import traceback
 import sys
 import warnings
-# from StringIO import StringIO
 from io import StringIO, BytesIO
+from cocotb import _py_compat
 
 """Exceptions and functions for simulation result handling."""
 
@@ -128,4 +128,9 @@ class TestSuccess(TestComplete):
 
 class SimFailure(TestComplete):
     """Exception showing that the simulator exited unsuccessfully."""
+    pass
+
+
+class SimTimeoutError(_py_compat.TimeoutError):
+    """Exception for when a timeout, in terms of simulation time, occurs"""
     pass
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -729,3 +729,40 @@ class ClockCycles(Waitable):
         for _ in range(self.num_cycles):
             yield trigger
         raise ReturnValue(self)
+
+
+@decorators.coroutine
+def with_timeout(triggers, timeout_time, timeout_unit=None):
+    """
+    Waits on triggers, throws an exception if it waits longer than the given time
+
+    Usage:
+
+    .. code-block:: python
+
+        yield with_timeout(coro, 100, 'ns')
+        yield with_timeout([coro, event.wait(), ...], 100, 'ns')
+
+    Args:
+        triggers (cocotb_yield_expr):
+            Any expression that could be right of a :keyword:`yield`
+            (or :keyword:`await` in Python 3) expression in cocotb.
+        timeout_time (int):
+            Time duration
+        timeout_unit (str, optional):
+            Units of duration, accepts any values that :class:`~cocotb.triggers.Timer` does
+
+    Returns:
+        First trigger that completed if timeout did not occur.
+
+    Raises:
+        :exc:`SimTimeoutError`: If timeout occurs
+
+    .. versionadded:: 1.3
+    """
+    timeout_timer = cocotb.triggers.Timer(timeout_time, timeout_unit)
+    res = yield [timeout_timer, triggers]
+    if res is timeout_timer:
+        raise cocotb.result.SimTimeoutError
+    else:
+        raise ReturnValue(res)
--- a/documentation/source/triggers.rst
+++ b/documentation/source/triggers.rst
@@ -75,3 +75,5 @@ These are used to synchronize coroutines with each other.
 .. autoclass:: cocotb.triggers.Lock
     :members:
     :member-order: bysource
+
+.. autofunction:: cocotb.triggers.with_timeout
--- a/tests/test_cases/test_cocotb/test_cocotb.py
+++ b/tests/test_cases/test_cocotb/test_cocotb.py
@@ -1254,6 +1254,7 @@ def test_assertion_is_failure(dut):
 class MyException(Exception):
     pass
 
+
 @cocotb.test(expect_error=MyException)
 def test_expect_particular_exception(dut):
     yield Timer(1)
@@ -1264,6 +1265,28 @@ def test_expect_particular_exception(dut):
 def test_expect_exception_list(dut):
     yield Timer(1)
     raise MyException()
+
+
+@cocotb.coroutine
+def example():
+    yield Timer(10, 'ns')
+    raise ReturnValue(1)
+
+
+@cocotb.test()
+def test_timeout_func_fail(dut):
+    try:
+        yield cocotb.triggers.with_timeout(example(), timeout_time=1, timeout_unit='ns')
+    except cocotb.result.SimTimeoutError:
+        pass
+    else:
+        assert False, "Expected a Timeout"
+
+
+@cocotb.test()
+def test_timeout_func_pass(dut):
+    res = yield cocotb.triggers.with_timeout(example(), timeout_time=100, timeout_unit='ns')
+    assert res == 1
 
 
 @cocotb.test()
--- a/tests/test_cases/test_cocotb/test_cocotb.py
+++ b/tests/test_cases/test_cocotb/test_cocotb.py
@@ -1289,6 +1289,29 @@ def test_timeout_func_pass(dut):
     assert res == 1
 
 
+@cocotb.test(expect_error=cocotb.result.SimTimeoutError, timeout_time=1, timeout_unit='ns')
+def test_timeout_testdec_fail(dut):
+    yield Timer(10, 'ns')
+
+
+@cocotb.test(timeout_time=100, timeout_unit='ns')
+def test_timeout_testdec_pass(dut):
+    yield Timer(10, 'ns')
+
+
+@cocotb.test(timeout_time=10, timeout_unit='ns')
+def test_timeout_testdec_simultaneous(dut):
+    try:
+        yield cocotb.triggers.with_timeout(Timer(1, 'ns'), timeout_time=1, timeout_unit='ns')
+    except cocotb.result.SimTimeoutError:
+        pass
+    else:
+        assert False, "Expected a Timeout"
+    # Whether this test fails or passes depends on the behavior of the
+    # scheduler, simulator, and the implementation of the timeout function.
+    # CAUTION: THIS MAY CHANGE
+
+
 @cocotb.test()
 def test_bad_attr(dut):
     yield cocotb.triggers.NullTrigger()
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -137,3 +137,11 @@ class nullcontext(object):
 # https://stackoverflow.com/a/38668373
 # Backport of abc.ABC, compatible with Python 2 and 3
 abc_ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
+
+
+# source TimeoutError introduced in Python 3.3 to be used by timeout functions
+if sys.version_info < (3, 3):
+    class TimeoutError(OSError):
+        pass
+else:
+    TimeoutError = TimeoutError


eric-wieser:
I wonder whether this makes sense as more general `cocotb.utils.timeout` decorator that can be applied to any coroutine?

eric-wieser:
Implementation looks nice and clean though - just a question of whether this belongs.

ktbarrett:
I thought it was a good idea, so I tried with `_decorator_helper` first. And that in combination with `update_wrapper` was causing unexpected behavior, so I just just went with the straightforward approach. 

This PR still addresses #7, I'll just change the title to reflect change in direction.

cocotb/decorators.py
```diff
@@ -460,3 +457,25 @@ def __init__(self, f, timeout=None, expect_fail=False, expect_error=False,
...
+@public
+def timeout(timeout_time, timeout_unit=None):
+
+    def decorator(f):
+
+        def wrapper(*args, **kwargs):
+            timeout_timer = cocotb.triggers.Timer(timeout_time, timeout_unit)
+            res = yield [timeout_timer, coroutine(f)(*args, **kwargs)]
+            if res is timeout_timer:
+                raise TimeoutError
```
eric-wieser: I don't think `TimeoutError` exists in python 2, sadly

ktbarrett: Ahh Python 2, definitely the farthest thing from my mind at 1 AM. So, should a new/separate exception be defined? Or should a definition be conditionally sourced?

```python
if sys.version_info < (3, 3):
    @public
    class TimeoutError(OSError):
        pass
```

cmarqu:
Nice. Please add some documentation too (with ``.. versionadded:: 1.3``). The docs should mention that this is in simulation time (as Philipp requested in the older PR's discussion).

ktbarrett:
Well I'm actually not so sure I like the whole timeout decorator idea on second glance. First, it means that the coroutine has a fixed timeout on every call. So we would still need a standalone timeout function for when it isn't fixed. Second, implementation was slightly problematic, as described before using `update_wrapper` on a coroutine does not work and I'm not sure why.

So I think I'll go back to the original plan of putting timeout on the test decorator and adding a general timeout function for all other cases.
```python
def timeout(timeout_time, timeout_unit=None):

    @cocotb.coroutine
    def wrapper(*triggers):
        timeout_timer = cocotb.triggers.Timer(timeout_time, timeout_unit)
        res = yield [timeout_timer] + triggers
        if res is timeout_timer:
            raise cocotb.results.SimTimeoutError
        else:
            return res

    return wrapper
```

cocotb/decorators.py
eric-wieser: > Or should a definition be conditionally sourced?

I'd probably compromise on declaring a custom exception type, but derive it from the builtin `TimeoutError` if it exists

eric-wieser:
> not so sure I like the whole timeout decorator idea on second glance. First, it means that the coroutine has a fixed timeout on every call. 

I think I agree with that, good point.

Perhaps as a compromise then, add a `cocotb.utils.with_timeout` function, used as:
```
try:
    yield cocotb.utils.with_timeout(lambda: my_coro(args), 10, units='sec')
except cocotb.utils.TimeoutError:
    pass
```

And then use this internally within the `test` decorator.

themperek:
BTW: if we ever move to pytest one can use this https://pypi.org/project/pytest-timeout/

cmarqu:
@themperek Though this would be a wall-clock timeout, not a simulation time timeout. (Still useful.)

cocotb/triggers.py
```diff
@@ -734,3 +734,35 @@ def _wait(self):
...
+
+    Results are the first trigger that completed, or a SimTimeoutError if timeout
+    occured.
+
+    Usage:
+        yield with_timeout(coro, 100, 'ns')
+        yield with_timeout([coro, event.wait(), ...], 100, 'ns')
+
+    Args:
+        triggers : yield_expr
```
cmarqu: We use the Google style which looks like ``arg1 (type): Description of arg1`` (see https://www.sphinx-doc.org/en/2.0/usage/extensions/napoleon.html#google-vs-numpy)

cocotb/triggers.py
```diff
@@ -734,3 +734,35 @@ def _wait(self):
...
+            Any expression that could be right of a yield statement in cocotb
+        timeout_time : int
+            Time duration
+        timeout_unit : Optional[str]
+            Units of duration, accepts any values that cocotb.triggers.Timer does
```
```suggestion
            Units of duration, accepts any values that :class:`Timer` does
```
will symlink this nicely.

cocotb/triggers.py
```diff
@@ -734,3 +734,35 @@ def _wait(self):
...
             yield trigger
         raise ReturnValue(self)
+
+
+@decorators.coroutine
+def with_timeout(triggers, timeout_time, timeout_unit=None):
+    """
+    Waits on triggers, throws an exception if it waits longer than the given time
+
+    Results are the first trigger that completed, or a SimTimeoutError if timeout
```
```suggestion
    Results are the first trigger that completed, or a :exc:`SimTimeoutError` if timeout
```

cocotb/decorators.py
```diff
@@ -434,23 +434,42 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
 
     Args:
         timeout (int, optional):
```
garmin-mjames: timeout_time is used for the actual argument name.

cocotb/decorators.py
```diff
@@ -433,24 +433,43 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
...
+            def timeout_wrapper(*args, **kwargs):
+                running_co = co(*args, **kwargs)
+                try:
+                    yield cocotb.triggers.with_timeout(running_co, self.timeout_time, self.timeout_unit)
+                except cocotb.result.SimTimeoutError:
+                    running_co.kill()
+                    raise
+
+            functools.update_wrapper(timeout_wrapper, f)
+            f = timeout_wrapper
```
eric-wieser: nit: shorter as
```
@functools.wraps(f)
def f(...):
    ...
# no need for update_wrapper
```

cocotb/result.py
```diff
@@ -129,3 +129,13 @@ class TestSuccess(TestComplete):
 class SimFailure(TestComplete):
     """Exception showing that simulator exited unsuccessfully."""
     pass
+
+
+if sys.version_info < (3,3):
+    class TimeoutError(OSError):
+        pass
```
eric-wieser: This belongs in https://github.com/cocotb/cocotb/blob/master/cocotb/_py_compat.py, I think

cocotb/triggers.py
```diff
@@ -735,3 +735,44 @@ def _wait(self):
...
+
+    Raises:
+        :exc:`SimTimeoutError`: If timeout occurs
+    """
+
+    if not isinstance(triggers, list):
+        triggers = [triggers]
+
+    timeout_timer = cocotb.triggers.Timer(timeout_time, timeout_unit)
+    res = yield [timeout_timer,] + triggers
```
eric-wieser: I'm curious if you can drop the `if` above and just use:
```
res = yield cocotb.triggers.First(timeout_trigger, triggers)
```

cocotb/triggers.py
```diff
@@ -735,3 +735,44 @@ def _wait(self):
...
+    if res is timeout_timer:
+        raise cocotb.result.SimTimeoutError
+    else:
+        res
```
```suggestion
        raise ReturnValue(res)
```

documentation/source/triggers.rst
```diff
@@ -75,3 +75,6 @@ These are used to synchronize coroutines with each other.
 .. autoclass:: cocotb.triggers.Lock
     :members:
     :member-order: bysource
+
+.. autoclass:: cocotb.triggers.with_timeout
+.. versionadded:: 1.3
```
eric-wieser: I would recommend you put this in the docstring itself

cocotb/decorators.py
```diff
@@ -433,24 +433,43 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
...
+        if timeout_time is not None:
+            co = coroutine(f)
+
+            def timeout_wrapper(*args, **kwargs):
+                running_co = co(*args, **kwargs)
+                try:
+                    yield cocotb.triggers.with_timeout(running_co, self.timeout_time, self.timeout_unit)
+                except cocotb.result.SimTimeoutError:
+                    running_co.kill()
+                    raise
```
eric-wieser: Doesn't really matter, but I'd be inclined to propagate the return value here, by adding `ret = yield ...` and `raise ReturnValue(ret)` - just in case we ever use it, or someone abuses it for weird lifetime semantics

cocotb/triggers.py
```diff
@@ -735,3 +735,44 @@ def _wait(self):
...
+    Usage:
+
+    .. code-block:: python
+
+        yield with_timeout(coro, 100, 'ns')
+        yield with_timeout([coro, event.wait(), ...], 100, 'ns')
+
+    Args:
+        triggers (cocotb_yield_expr):
+            Any expression that could be right of a yield statement in cocotb
```
eric-wieser: It's a shame we don't really have a concise way to describe _"a thing accepted by `_trigger_from_any`"_. No action needed here.

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1251,5 +1251,31 @@ def test_assertion_is_failure(dut):
...
+
+
+@cocotb.test(expect_error=True, timeout_time=1, timeout_unit='ns')
+def test_timeout_testdec_fail(dut):
+    yield Timer(10, 'ns')
+
+
+@cocotb.test(expect_error=False, timeout_time=100, timeout_unit='ns')
+def test_timeout_testdec_pass(dut):
+    yield Timer(10, 'ns')
```
eric-wieser: Can you add a test where the two timeouts are identical, and make a judgement call on whether you expect it to fail?

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1251,5 +1251,31 @@ def test_assertion_is_failure(dut):
...
+        yield Timer(10, 'ns')
+    yield cocotb.triggers.with_timeout(fail(), timeout_time=1, timeout_unit='ns')
+
+
+@cocotb.test(expect_error=False)
+def test_timeout_func_pass(dut):
+    @cocotb.coroutine
+    def success():
+        yield Timer(10, 'ns')
+    yield cocotb.triggers.with_timeout(success(), timeout_time=100, timeout_unit='ns')
```
eric-wieser: Please add a return value to `success() here, to test the bug I identified in https://github.com/cocotb/cocotb/pull/1119/files#r346068338

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1251,5 +1251,31 @@ def test_assertion_is_failure(dut):
...
 
+@cocotb.test(expect_error=True)
+def test_timeout_func_fail(dut):
+    @cocotb.coroutine
+    def fail():
+        yield Timer(10, 'ns')
+    yield cocotb.triggers.with_timeout(fail(), timeout_time=1, timeout_unit='ns')
+
+
+@cocotb.test(expect_error=False)
```
eric-wieser:
```suggestion
@cocotb.test()
```

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1251,5 +1251,31 @@ def test_assertion_is_failure(dut):
...
+        yield Timer(10, 'ns')
+    yield cocotb.triggers.with_timeout(success(), timeout_time=100, timeout_unit='ns')
+
+
+@cocotb.test(expect_error=True, timeout_time=1, timeout_unit='ns')
+def test_timeout_testdec_fail(dut):
+    yield Timer(10, 'ns')
+
+
+@cocotb.test(expect_error=False, timeout_time=100, timeout_unit='ns')
```
eric-wieser:
```suggestion
@cocotb.test(timeout_time=100, timeout_unit='ns')
```

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1251,5 +1251,31 @@ def test_assertion_is_failure(dut):
     assert False
 
 
+@cocotb.test(expect_error=True)
+def test_timeout_func_fail(dut):
+    @cocotb.coroutine
+    def fail():
+        yield Timer(10, 'ns')
+    yield cocotb.triggers.with_timeout(fail(), timeout_time=1, timeout_unit='ns')
```
eric-wieser: I'd prefer to write this as:
```
try:
    yield cocotb.triggers.with_timeout(fail(), timeout_time=1, timeout_unit='ns')
except cocotb.triggers.SimTimeoutError:
    pass
else:
    assert False, "Expected a timeout"
```
Because that proves that execution continues after a non-test-level timeout

cocotb/triggers.py
ktbarrett: `timeout_timer` is a trigger and `triggers` is either a single trigger or a list of triggers. `First` expects an arglist of triggers. That would work if you splatted `triggers`, but you would have to condition that on it being a `list` first. You can't really get around normalizing it.

You could force `timeout_time` and `timeout_unit` to be keyword-only and then take an arglist, but end up adding a check to `timeout_time` to force the user to give a value. That might be cleaner though.

tests/test_cases/test_cocotb/test_cocotb.py
ktbarrett: I had considered this. This depends upon the scheduler's behavior doesn't it? If you yield a list, and two triggers fire simultaneously, the scheduler returns the first trigger in the list. If the scheduler does not guarantee what I think it does, and my tests are happy accidents, or if you expect the scheduler's behavior to change, the behavior of this function in that case will also change. I'd probably consider this case UB.

cocotb/decorators.py
ktbarrett: That does not work. After the call to `coroutine`, `f` is wrapped by the `co` coroutine object, Wrapping an object multiple times is apparently very sketch and doesn't work well. I had issues with this in the original implementation as stated earlier in the thread. This way works though...

ktbarrett:
So I rebased this on #1116. Once that PR is in, this can be merged.

cocotb/decorators.py
```diff
@@ -433,25 +433,61 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
     Used as ``@cocotb.test(...)``.
 
     Args:
-        timeout (int, optional):
-            value representing simulation timeout (not implemented).
+        timeout_time (int, optional):
+            value representing simulation timeout.
+        timeout_unit (str, optional):
+            unit of timeout value, see :class:`~cocotb.triggers.Timer` for more info.
```
eric-wieser: ``..versionadded` would be good on both of these

documentation/source/triggers.rst
```diff
@@ -75,3 +75,5 @@ These are used to synchronize coroutines with each other.
 .. autoclass:: cocotb.triggers.Lock
     :members:
     :member-order: bysource
+
+.. autoclass:: cocotb.triggers.with_timeout
```
eric-wieser: This should be `.. autofunction` I think

tests/test_cases/test_cocotb/test_cocotb.py
eric-wieser: > This depends upon the scheduler's behavior doesn't it?

Yes it does.

I'd still like to have a test for it though, so that we know if the behavior changed - I think tests for UB are still sensible, as long as they're labelled to indicate that it's ok to change the test if it fails.

cocotb/decorators.py
eric-wieser: > Wrapping an object multiple times is apparently very sketch and doesn't work well.

Is there some failure traceback you can quote here?

ktbarrett: I must be seeing things... nvm.

tests/test_cases/test_cocotb/test_cocotb.py
ktbarrett: test was added

cocotb/triggers.py
```diff
@@ -735,3 +735,46 @@ def _wait(self):
...
+        :exc:`SimTimeoutError`: If timeout occurs
+
+    .. versionadded:: 1.3
+    """
+
+    if not isinstance(triggers, list):
+        triggers = [triggers]
+
+    timeout_timer = cocotb.triggers.Timer(timeout_time, timeout_unit)
+    res = yield [timeout_timer] + triggers
```
eric-wieser: I know I said this before, but I think this will still work if you change this to the following:
```python
timeout_timer = cocotb.triggers.Timer(timeout_time, timeout_unit)
res = yield [timeout_timer, triggers]
```
Nested sequences are unpacked automatically, as is proven in this test:
https://github.com/cocotb/cocotb/blob/8db47936dace61ff3b9700f26c2f57a2b527ce13/tests/test_cases/test_cocotb/test_cocotb.py#L1171-L1178

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1251,5 +1251,61 @@ def test_assertion_is_failure(dut):
...
+def test_timeout_testdec_pass(dut):
+    yield Timer(10, 'ns')
+
+
+@cocotb.test(expect_error=True, timeout_time=10, timeout_unit='ns')
+def test_timeout_testdec_simultaneous(dut):
+    yield Timer(10, 'ns')
+    # Whether this test fails or passes depends on the behavior of the
+    # scheduler and the implementation of the timeout function.
+    # CAUTION: THIS MAY CHANGE
```
eric-wieser: Again, I'd prefer you test this with a `try`/`except`:
```python
    try:
        yield cocotb.triggers.with_timeout(Timer(1, 'ns'), timeout_time=1, timeout_unit='ns')
    except cocotb.result.SimTimeoutError:
        pass
    else:
        assert False, "Expected a Timeout"
    # Whether this test fails or passes depends on the behavior of the
    # scheduler and the implementation of the timeout function.
    # CAUTION: THIS MAY CHANGE
```

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1266,5 +1267,50 @@ def test_expect_exception_list(dut):
...
+        assert False, "Expected a Timeout"
+
+
+@cocotb.test()
+def test_timeout_func_pass(dut):
+    res = yield cocotb.triggers.with_timeout(example(), timeout_time=100, timeout_unit='ns')
+    assert res == 1
+
+
+@cocotb.test(expect_error=True, timeout_time=1, timeout_unit='ns')
```
eric-wieser:
```suggestion
@cocotb.test(expect_error=cocotb.result.SimTimeoutError, timeout_time=1, timeout_unit='ns')
```

cocotb/triggers.py
```diff
@@ -729,3 +729,42 @@ def _wait(self):
...
         raise ReturnValue(self)
+
+
+@decorators.coroutine
+def with_timeout(triggers, timeout_time, timeout_unit=None):
+    """
+    Waits on triggers, throws an exception if it waits longer than the given time
+
+    Can be used with any expression that would otherwise appear to the right of a
+    `yield` (or `await` in Python 3) statement.
```
```suggestion
    :keyword:`yield` (or :keyword:`await` in Python 3) expression.
```

eric-wieser: Actually, let's remove this whole sentence and put it under `triggers` below - saying the same thing twice isn't really that helpful.

cocotb/triggers.py
```diff
@@ -729,3 +729,42 @@ def _wait(self):
...
+    Usage:
+
+    .. code-block:: python
+
+        yield with_timeout(coro, 100, 'ns')
+        yield with_timeout([coro, event.wait(), ...], 100, 'ns')
+
+    Args:
+        triggers (cocotb_yield_expr):
+            Any expression that could be right of a yield statement in cocotb
```
```suggestion
            Any expression that could be right of a :keyword:`yield`
            (or :keyword:`await` in Python 3) expression in cocotb.
```

cocotb/triggers.py
```diff
@@ -729,3 +729,42 @@ def _wait(self):
...
+        timeout_time (int):
+            Time duration
+        timeout_unit (str, optional):
+            Units of duration, accepts any values that :class:`cocotb.triggers.Timer` does
```
eric-wieser: Could use a `~` here to make this render as `Timer` rather than `cocotb.triggers.Timer`, while still linking correctly:

```suggestion
            Units of duration, accepts any values that :class:`~cocotb.triggers.Timer` does
```

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1266,5 +1267,50 @@ def test_expect_exception_list(dut):
...
+@cocotb.test(timeout_time=10, timeout_unit='ns')
+def test_timeout_testdec_simultaneous(dut):
+    try:
+        yield cocotb.triggers.with_timeout(Timer(1, 'ns'), timeout_time=1, timeout_unit='ns')
+    except cocotb.result.SimTimeoutError:
+        pass
+    else:
+        assert False, "Expected a Timeout"
+    # Whether this test fails or passes depends on the behavior of the
+    # scheduler and the implementation of the timeout function.
```
eric-wieser: Not only the scheduler, but the simulator too. You're right to claim this, we draw attention to it in the `First` docs here: https://external-builds.readthedocs.io/html/cocotb/1119/triggers.html#cocotb.triggers.First

(no action needed)

eric-wieser:
Will leave this a few days to give @imphil a chance to look, otherwise will merge soon

ktbarrett:
@eric-wieser I think soon has come and passed.

eric-wieser:
Whoops, thanks for the ping

PR #1120, opened 2019-09-21
Tidy test_external

eric-wieser:
See individual commit messages

PR #1123, opened 2019-09-24
Fix Python 3.8 deprecation/syntax warnings

TC01:
A number of places in the tests trigger the following new warning in Python 3.8:

> SyntaxWarning: "is not" with a literal. Did you mean "!="?

See here for discussion of this change: https://bugs.python.org/issue34850

This commit resolves this by changing all instances of "is not" with a literal to use "!=".

There is also an additional compatibility-related change; the combine_results.py script uses a deprecated "getiterator" method in the xml.etree.elementtree package. This method has been deprecated in favor of the "iter" method since Python 2.7, so it should be safe to switch.

(These problems were all discovered while I was working on [packaging cocotb in Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=1747574)).

### DIFF
--- a/bin/combine_results.py
+++ b/bin/combine_results.py
@@ -52,7 +52,7 @@ def main():
     for fname in find_all("results.xml", args.directory):
         if args.debug : print("Reading file %s" % fname)
         tree = ET.parse(fname)
-        for ts in tree.getiterator("testsuite"):
+        for ts in tree.iter("testsuite"):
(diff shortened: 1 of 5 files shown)


eric-wieser:
Some overlap with gh-1120, but should be an easy rebase whichever goes in first

imphil:
LGTM, thanks!

PR #1124, opened 2019-09-24
Clean up spellchecker findings.

cmarqu:


documentation/source/coroutines.rst
```diff
@@ -172,7 +172,7 @@ of :keyword:`yield`. Provided they are decorated with ``@cocotb.coroutine``,
 ``async def`` functions using :keyword:`await` and regular functions using
 :keyword:`yield` can be used interchangeable - the appropriate keyword to use
 is determined by which type of function it appears in, not by the
-sub-coroutinue being called.
```
eric-wieser: I can't help myself with this typo...

documentation/source/library_reference.rst
```diff
@@ -2,6 +2,21 @@
...
+    BinaryRepresentation
+    binaryRepresentation
+    sim
+    SimHandle
+    Todo
+    un
+    arg
+    idx
+    Args
+    Enum
```
eric-wieser: Are all of these needed? Looks like some of them should be marked up instead. Or are they section titles?

documentation/source/coroutines.rst
cmarqu: Seems you have all the best words! ;)

documentation/source/library_reference.rst
cmarqu: I now removed them and mysteriously spell checking still is fine, while it wasn't before... no idea why.

PR #1125, opened 2019-09-24
Fix markup problem.

### LINKED ISSUE #1106: [docs] Fix API docs in Bus()
The arguments of the cocotb.bus.Bus() class don't get rendered properly at the moment. See https://cocotb.readthedocs.io/en/latest/library_reference.html#cocotb.bus.Bus

cmarqu:
Closes #1106.


cmarqu:
The effect is apparently that the class docstring and the ``__init__`` docstring are combined by Sphinx, and the "Args:" is not indented as would be the case in all other docstrings.

PR #1127, opened 2019-09-26
Use units argument for Timer and Clock in examples.

### LINKED ISSUE #1117: Examples should use units argument
As best practice examples, we should always use the units argument when specifying times.

cmarqu:
Closes #1117. The tests running in CI were not touched as they are not user-facing.
It looks like sometimes the assumption was that the simulator timescale would be in picoseconds, sometimes in nanoseconds. Normalization was done to nanoseconds mostly.

### DIFF
--- a/documentation/source/coroutines.rst
+++ b/documentation/source/coroutines.rst
@@ -14,8 +14,8 @@ when it occurs.  For example:
 
     @cocotb.coroutine
     def wait_10ns():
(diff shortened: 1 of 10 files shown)


cmarqu:
Depends on #1132.

cmarqu:
Trigger Travis again now that #1132 is in.

PR #796, opened 2019-02-04
Add port support and testbench template generator.

cmarqu:
Fixes #695.

One thing to be noted is that this needs ``SINGLETON_HANDLES`` in ``lib/gpi/Makefile`` _not set_ to actually discover ports.
I do not know the background behind this switch, but removing it didn't seem to have a harmful effect for me.

PR #1128, opened 2019-09-26
Add support for _is_port, _port_direction attrs.

### LINKED ISSUE #695: Support _is_port and _port_direction attributes; implement testbench generator
When iterating over the design, we can already differentiate the found objects in a number of ways, but it's not possible to get information about whether an object is an entity or module port, and what its direction is.

I have a branch where I have added this support (for VPI, VHPI and FLI), the attribute names are ``_is_port`` (a boolean), ``_port_direction`` (a GPI enum), and ``_port_direction_string`` (a string). Any feedback on these names?

Going further, I have added a ``create_testbench_template.py`` in ``bin/`` which writes out another Python file (a testbench template) that is working on ports that have been found by looping over the DUT toplevel (it tries to identify clock, reset, supply and ground ports based on name matching and drives them accordingly).
The testbench template also takes the rest of the found input signals and exemplary drives a 0 to them, clocks the DUT, reads and prints the output port values, and write a WaveDrom file.
This should make it much easier to start with a testbench for a design with many ports.

Please take a look at https://github.com/cmarqu/cocotb/tree/portsupport
Not working yet: accessing real value and structural ports, but not many real-life DUT will have those.

cmarqu:
Fixes #695. Supersedes #796.

One thing to be noted is that this needs ``SINGLETON_HANDLES`` in ``lib/gpi/Makefile`` not set to actually discover ports.
I do not know the background behind this switch, but removing it didn't seem to have a harmful effect for me.


cmarqu:
Missing: documentation.

cocotb/share/lib/gpi/gpi_priv.h
```diff
@@ -125,6 +131,9 @@ class GpiObjHdl : public GpiHdl {
     virtual const char* get_type_str(void);
     gpi_objtype_t get_type(void) { return m_type; };
     bool get_const(void) { return m_const; };
+    bool get_is_port(void) { return m_is_port; };
+    virtual const char* get_port_direction_str(void);
```
eric-wieser: Why is this virtual?

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -798,6 +798,72 @@ static PyObject *get_const(PyObject *self, PyObject *args)
...
 }
 
+static PyObject *get_is_port(PyObject *self, PyObject *args)
+{
+    int result;
+    gpi_sim_hdl hdl;
+    PyObject *pyresult;
+
+    PyGILState_STATE gstate;
+    gstate = TAKE_GIL();
```
eric-wieser: There's no need to take the gil here, these functions are only callable from within python, not from the gpi code. See `get_const` for an example.

cocotb/share/lib/gpi/gpi_priv.h
cmarqu: Thanks for the review, I now made this a helper function.
Your enum suggestion (I like enums! :) ) should probably be a new issue so that adding a dependency can be discussed.

cocotb/share/lib/simulator/simulatormodule.c
cmarqu: Thanks, removed now.

cmarqu:
Not sure what's up with LGTM here.

cmarqu:
Trying to trigger LGTM again.

cmarqu:
@eric-wieser In https://github.com/cocotb/cocotb/pull/796#discussion_r257115940 you said that whether something is a port shouldn't be a boolean flag but a sub-class - however currently, depending on the port's type, we already get ``IntegerObject``, ``StringObject``, ``RealObject`` etc., and whether such an object is also a port is somewhat orthogonal.

cmarqu:
For ``SINGLETON_HANDLES``, see also #1500.

ktbarrett:
After #1540 is in, I'll add in support for object cleanup, then we are safe to remove `SINGLETON_HANDLES`, although I'm not entirely sure how necessary it was in the first place.

cocotb/handle.py
```diff
@@ -78,6 +78,8 @@ def __init__(self, handle, path):
 
         self._name = simulator.get_name_string(self._handle)
         self._type = simulator.get_type_string(self._handle)
+        self._is_port = simulator.get_is_port(self._handle)
+        self._port_direction_string = simulator.get_port_direction_string(self._handle)
```
eric-wieser, line 82: Can we make this an enum now that we've dropped Python 2? Could either do that within this file, as
```
from enum import Enum

class PortDirection(Enum):
    UNHANDLED = 0
    INPUT = 1
    OUTPUT = 2
    INOUT = 3
    MIXEDIO = 4  # Verilog only
    NO_DIRECTION = 5  # Verilog only
    BUFFER = 6  # VHDL only
    LINKAGE = 7  # VHDL only
    INTERNAL = 8  # FLI (Mentor) only, meaning "is not a port"
```
```python
port_dir = simulator.get_port_direction_string(self._handle)
self._port_direction = None if port_dir == -1 else PortDirection(port_dir)
```

ktbarrett:
@cmarqu Could you explain how `SINGLETON_HANDLES` is getting in the way of this change working?

cmarqu:
> @cmarqu Could you explain how `SINGLETON_HANDLES` is getting in the way of this change working?

I can't say how, only that it did prevent some (or even all, I don't remember) ports from being marked.

ktbarrett:
@cmarqu could you rebase this? I have a nasty feeling that replacing/removing the singleton handle cache is going to be *a lot* of work. Instead, I'd like to investigate why this isn't working with the cache in place and perhaps work around that.

cmarqu:
I have tried to rebase, but I didn't manage to get the code working as before. 
The current state is here, I'm not sure if I can find the time to rebase *that* before my four weeks of travel: https://github.com/cmarqu/cocotb/tree/portsupport-only2

ktbarrett:
So running [with](https://github.com/cocotb/cocotb/files/4929144/with_cache.log) and [without](https://github.com/cocotb/cocotb/files/4929146/without_cache.log) the cache using the original branch with Icarus.

When the `dut` object, which is a module, is iterated over, `vpi_scan` returns iterators over each type of subobject in the module: registers, then memory, then parameters, then ports, etc. The same handle can be returned multiple times during this iteration, for example `and_output` is returned both during `vpiReg` iteration and `vpiPort` iteration. The check whether a handle is a `vpiPort` during the `vpiReg` iteration returns false. If this occurs *before* the `vpiPort` iteration, the handle is cached as not being a port. That's why caching doesn't work in this case.

I tried unconditionally getting `vpiDirection` on signals, but Icarus doesn't like that. There might be some other way to implement this, but I'll look at it tomorrow.

ktbarrett:
So unfortunately, there is no way given a vpiHandle to immediately check if it a port or not. Instead, what you must do is iterate over the parent and see if the same object is returned during `vpiPort` iteration. There is no concept of accessing the parent in the GPI (a major design flaw IMO). So instead we could get the parent via the VPI (`vpiParent`) then do the above procedure.

garmin-mjames:
@ktbarrett You can get the parent module with `vpi_handle(vpiModule, net_or_var_hdl)`, then iterate with `vpiPort`.
We could also check the lower connection via `vpi_handle(vpiLowConn, new_hdl)` but I'm not sure if it'll return the same handle or not.

ktbarrett:
@garmin-mjames 
> You can get the parent module with `vpi_handle(vpiModule, net_or_var_hdl)`, then iterate with `vpiPort`.

I tried that in #1997. I would have to check for object equality with `vpi_compare_objects` which isn't implemented in any of the FOSS simulators. There is a bug in Mentor and Aldec simulators where iterating over the same type on the same object more than once causes the iterator return nothing the second time.

garmin-mjames:
@ktbarrett everything is broken 😞

themperek:
@cmarqu Any plans to work on this? Can I help?

ktbarrett:
@themperek @cmarqu I think the best way to support this is to introduce another iterator on `GPI_MODULE` objects over the ports. That is similar to how the interfaces do it. I would maybe add this to `gpi_iterator_sel_t` enum and support it in each implementation. I don't think being able to interrogate every object for portness is a reasonable solution. Maybe there should be a specialization of `HierarchyObject` made for `MODULE`s? Finally, I don't think we need to keep adding features in `_underscore_names`.

cmarqu:
@themperek I was able to hack this in initially by looking at how other attributes are queried, but anything more advanced than that is beyond my capabilities.

PR #1129, opened 2019-09-26
Add testbench template creator script.

cmarqu:
Depends on #1128.


cmarqu:
Open question: where should this live? Is bin/ okay? Should it become a separate module/package?

PR #1130, opened 2019-09-28
Implements SimTime type for working with durations of simulation time

ktbarrett:
`SimTime` was modeled after the `datetime.timedelta` in the Python standard library. The class supports similar functions for doing arithmetic on `timedelta`s. The constructor looks just like `Timer` and the to-be-merged `with_timeout` function.

Changing the existing timing-related functions in cocotb, adding documentation, and adding copious tests will be done after positive initial review.

### DIFF
--- a/cocotb/utils.py
+++ b/cocotb/utils.py
@@ -551,7 +551,7 @@ class SimTime():
 
     def __init__(self, time, unit=None):
         if isinstance(time, SimTime):
-            tme = time.steps
+            time = get_time_from_sim_steps(time.steps, unit)
         self.steps = get_sim_steps(time, unit)
         self.unit = unit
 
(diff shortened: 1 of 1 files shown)


cocotb/utils.py
```diff
@@ -545,3 +545,123 @@ def remove_traceback_frames(tb_or_exc, frame_names):
...
         return tb
+
+
+class SimTime():
+
+    def __init__(self, time, unit=None):
+        if isinstance(time, SimTime):
+            tme = time.steps
+        self.steps = get_sim_steps(time, unit)
+        self.unit = unit
```
eric-wieser: I'm not sure this property makes sense - when you get here, the time is now in "sim units"

cocotb/utils.py
```diff
@@ -545,3 +545,123 @@ def remove_traceback_frames(tb_or_exc, frame_names):
...
+
+    def __sub__(self, other):
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        result = SimTime(self.steps - other.steps)
+        result.unit = self._gcd_time_unit(self.unit, other.unit)
+        return result
+
+    def __mul__(self, other):
+        if not isinstance(other, (int, float)):
```
eric-wieser: This precludes more exotic numbers like `fraction.Fraction`, `decimal.Decimal`, or `np.uint64`

cocotb/utils.py
```diff
@@ -545,3 +545,123 @@ def remove_traceback_frames(tb_or_exc, frame_names):
...
+    def __rmul__(self, other):
+        if not isinstance(other, (int, float)):
+            return NotImplemented
+        result = SimTime(self.steps * other)
+        result.unit = self.unit
+        return result
+
+    def __floordiv__(self, other):
+        if not isinstance(other, (int, float)):
+            return NotImplemented
```
eric-wieser: This should probably have an `int = SimTime // Simtime` behavior too.

cocotb/utils.py
ktbarrett: I guess it depends upon your point of view. Do operations on `SimTime` objects that underflow the simulator's precision even make sense? The current implementation throws if you try to create a duration less than the simulator's precision. If we go to a floating point implementation we have to worry about how to round to the simulator's precision. With this implementation we get floordiv and remainder so we don't lose any data. It's a design decision, is the underlying type abstract or concrete?

eric-wieser: I think we're talking past each other here. My complaint is that the `units` attribute doesn't actually tell me anything about the object. Changing it does not have any effect on the length of time it represents.

I also think line 554 is suspicious - to me `SimTime(SimTime(1, 'sec'), 'sec')` is nonsense, but for you it creates something that depends on the simulator resolution.

ktbarrett: The whole reason I have the units member is for serialization. I originally didn't store it, got to `__str__` and started scratching my head as to what a good representation was without units. But I'm not against just removing `__str__` and all references to units. Alternatively, you could think of units as a runtime type parameter.

eric-wieser: > started scratching my head as to what a good representation was without units

I might be tempted to just pick a unit value that results in a magnitude between 0 and 1000 (ie, engineering notation). You could add an overload of __format__ to let the user choose the unit for printing.

cocotb/utils.py
```diff
@@ -619,5 +619,11 @@ def __neg__(self):
     def __abs__(self):
         return SimTime(abs(self.steps))
 
+    def __str__(self):
+        return repr(self)
```
eric-wieser: This is redundant - if you omit it, python generates exactly this code for you.

cocotb/utils.py
```diff
@@ -619,5 +619,11 @@ def __neg__(self):
...
+
     def __repr__(self):
-        return "{}({})".format(self.__class__.__name__, self.steps)
+        for unit in ('fs', 'ps', 'ns', 'us', 'ms', 'sec'):
+            time = get_time_from_sim_steps(self.steps, unit)
+            if 0 <= time and time < 1000:
+                return "{}({}, '{}')".format(self.__class__.__name__, time, unit)
```
eric-wieser: Nit:`"'{}'".format(s)` is in general better spelt `"{!r}".format(s)` (r for repr), which handles embedded quotes and escape sequences for you. Works either way here though.

Issue #1122, opened 2019-09-23
Compilation warnings  [-Wmaybe-uninitialized]


themperek:
```
    /cocotb/cocotb/share/lib/fli/FliImpl.cpp: In member function ‘virtual GpiIterator::Status FliIterator::next_handle(std::string&, GpiObjHdl**, void**)’:
    /cocotb/cocotb/share/lib/fli/FliImpl.cpp:799:31: warning: ‘c_name’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```
```
    /cocotb/cocotb/share/lib/vhpi/VhpiImpl.h: In member function ‘virtual void VhpiImpl::get_sim_time(uint32_t*, uint32_t*)’:
    /cocotb/cocotb/share/lib/vhpi/VhpiImpl.h:51:9: warning: ‘loglevel’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```

Issue #1087, opened 2019-08-24
Can cocotb use C++11?


eric-wieser:
Might be good to have a minimum supported C++ standard version stated in our docs, just like we have a minimum supported python version.

imphil:
More useful than language features is typically specifying the minimum supported compiler, *and* having a demonstrated need for a given feature from a language version.

Probably the most common Linux distribution in the EDA industry is RHEL, next to SLES and Ubuntu.

- RHEL7: GCC 4.8 (https://access.redhat.com/solutions/19458)
- RHEL6: GCC 4.4 (https://access.redhat.com/solutions/19458)

- Ubuntu 16.04: GCC 5.3.1
- Ubuntu 18.04: GCC 7.4.0

- SLES 12 SP3: GCC 4.8
(Documentation for SLES is as usual harder to find than for other distributions.)

For C++11, many things are already available in GCC 4.4, some are not ([docs](https://gcc.gnu.org/projects/cxx-status.html)). Would you be ok with specifying GCC 4.4 as minimum version for now, or do you have a specific feature you'd like to use which isn't available there?

TC01:
Note that RHEL6 ships with python 2.6, which is too old to support cocotb anyway. To get python 2.7, you need to either build it yourself or enable a SCL. (I'm currently using cocotb on a RHEL6 server and have done exactly that).

So there is a potential argument to make for moving to gcc 4.8 instead of 4.4, if necessary: it already doesn't work out of the box due to the system Python version.

themperek:
@TC01 You can use cond and install locally (https://conda.io/en/latest/)

TC01:
Sure-- I would count that as "build it yourself" though. You're not getting the system Python.

It's also possible to get a newer gcc on RHEL6 (again, either by enabling a SCL or setting it up yourself).

PR #1137, opened 2019-10-03
Clean makefiles, fix C/C++ warnings, support Clang

### LINKED ISSUE #1136: Makefile issues, support for building cocotb with Clang
* GCC is called out specifically by the makefiles to build cocotb. There really isn't a good reason for this, Clang is capable of compiling the code.
   * I don't know much about mingw, is it possible we could switch to calling out `cc` and `c++`?
    *  A quick check shows that Clang's `-Wall` includes `-Woverloaded-virtual` so the build fails.
* The name of GCC_ARGS and GXX_ARGS should change.
* The makefiles should respect user supplied CFLAGS and CXXFLAGS. 
* The makefiles don't actually uniformly enable warnings for all platforms, and the number of warnings enabled is rather weak.
* There are still lurking warnings that should be addressed.
 * I have some opinions about `-Werror`, but I'll keep them to myself.

ktbarrett:
Fixes #1136.
* Makefiles use `cc` and `c++` for compilation, so Clang or any other
    compiler can be used to build cocotb.
* Makefiles respect user CFLAGS, CXXFLAGS, LDFLAGS, and LDLIBS
* Increased set of warnings
* Warnings applied for all archs
* Fixed warnings in C/C++ code
* renamed GCC_ARGS and GXX_ARGS to not allude to GCC

### DIFF
--- a/cocotb/share/include/gpi.h
+++ b/cocotb/share/include/gpi.h
@@ -221,7 +221,7 @@ void *gpi_get_callback_data(gpi_sim_hdl gpi_hdl);
 
 // Print out what implementations are registered. Python needs to be loaded for this,
 // Returns the number of libs
-int gpi_print_registered_impl(void);
+size_t gpi_print_registered_impl(void);
 
 #define GPI_RET(_code) \
     if (_code == 1) \
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -153,8 +153,10 @@ void embed_init_python(void)
        such that they can attach */
     const char *pause = getenv("COCOTB_ATTACH");
     if (pause) {
-        long sleep_time = strtol(pause, NULL, 10);
-        if (errno == ERANGE && (sleep_time == LONG_MAX || sleep_time == LONG_MIN)) {
+        unsigned long sleep_time = strtoul(pause, NULL, 10);
+        /* This should check for out-of-range parses which returns ULONG_MAX and sets errno,
+           as well as correct parses that would be sliced by the narrowing cast */
+        if (errno == ERANGE || sleep_time >= UINT_MAX) {
             LOG_ERROR("COCOTB_ATTACH only needs to be set to ~30 seconds");
             goto out;
         }
@@ -165,7 +167,7 @@ void embed_init_python(void)
         }
 
         LOG_ERROR("Waiting for %lu seconds - attach to PID %d with your debugger\n", sleep_time, getpid());
-        sleep(sleep_time);
+        sleep((unsigned int)sleep_time);
     }
 out:
     FEXIT;
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -30,6 +30,7 @@
 // Embed Python into the simulator using GPI
 
 #include <Python.h>
+#include <unistd.h>
 #include <cocotb_utils.h>
 #include "embed.h"
 #include "../compat/python3_compat.h"
--- a/cocotb/share/lib/fli/FliImpl.cpp
+++ b/cocotb/share/lib/fli/FliImpl.cpp
@@ -752,8 +752,8 @@ GpiIterator::Status FliIterator::next_handle(std::string &name, GpiObjHdl **hdl,
     }
 
     char *c_name;
-    PLI_INT32 accType = 0;
-    PLI_INT32 accFullType = 0;
+    PLI_INT32 accType;
+    PLI_INT32 accFullType;
     switch (*one2many) {
         case FliIterator::OTM_CONSTANTS:
         case FliIterator::OTM_VARIABLE_SUB_ELEMENTS:
@@ -776,6 +776,9 @@ GpiIterator::Status FliIterator::next_handle(std::string &name, GpiObjHdl **hdl,
             accFullType = acc_fetch_fulltype(obj);
             break;
         default:
+            c_name = NULL;
+            accType = 0;
+            accFullType = 0;
             LOG_WARN("Unhandled OneToMany Type (%d)", *one2many);
     }
 
--- a/cocotb/share/lib/gpi/GpiCbHdl.cpp
+++ b/cocotb/share/lib/gpi/GpiCbHdl.cpp
@@ -81,7 +81,7 @@ const std::string & GpiObjHdl::get_name()
 /* Genertic base clss implementations */
 char *GpiHdl::gpi_copy_name(const char *name)
 {
-    int len;
+    size_t len;
     char *result;
     const char null[] = "NULL";
 
--- a/cocotb/share/lib/gpi/GpiCbHdl.cpp
+++ b/cocotb/share/lib/gpi/GpiCbHdl.cpp
@@ -108,12 +108,6 @@ bool GpiHdl::is_this_impl(GpiImplInterface *impl)
     return impl == this->m_impl;
 }
 
-int GpiHdl::initialise(std::string &name)
-{
-    LOG_WARN("Generic initialize, doubt you should have called this");
-    return 0;
-}
-
 int GpiObjHdl::initialise(std::string &name, std::string &fq_name)
 {
     m_name = name;
--- a/cocotb/share/lib/gpi/GpiCommon.cpp
+++ b/cocotb/share/lib/gpi/GpiCommon.cpp
@@ -94,7 +94,7 @@ static GpiHandleStore unique_handles;
 #endif
 
 
-int gpi_print_registered_impl()
+size_t gpi_print_registered_impl()
 {
     vector<GpiImplInterface*>::iterator iter;
     for (iter = registered_impls.begin();
--- a/cocotb/share/lib/gpi/GpiCommon.cpp
+++ b/cocotb/share/lib/gpi/GpiCommon.cpp
@@ -351,8 +351,6 @@ gpi_sim_hdl gpi_get_handle_by_name(gpi_sim_hdl parent, const char *name)
 
 gpi_sim_hdl gpi_get_handle_by_index(gpi_sim_hdl parent, int32_t index)
 {
-    vector<GpiImplInterface*>::iterator iter;
-
     GpiObjHdl *hdl         = NULL;
     GpiObjHdl *base        = sim_to_hdl<GpiObjHdl*>(parent);
     GpiImplInterface *intf = base->m_impl;
--- a/cocotb/share/lib/gpi/gpi_priv.h
+++ b/cocotb/share/lib/gpi/gpi_priv.h
@@ -34,6 +34,7 @@
 #include <string>
 #include <vector>
 #include <map>
+#include <cocotb_utils.h>  // COCOTB_UNUSED
 
 typedef enum gpi_cb_state {
     GPI_FREE = 0,
@@ -229,10 +230,10 @@ protected:
 /* We would then have */
 class GpiClockHdl {
 public:
-    GpiClockHdl(GpiObjHdl *clk) { }
-    GpiClockHdl(const char *clk) { }
+    GpiClockHdl(GpiObjHdl *clk) { COCOTB_UNUSED(clk); }
+    GpiClockHdl(const char *clk) { COCOTB_UNUSED(clk); }
     ~GpiClockHdl() { }
-    int start_clock(const int period_ps) { return 0; } ; /* Do things with the GpiSignalObjHdl */
+    int start_clock(const int period_ps) { COCOTB_UNUSED(period_ps); return 0; } ; /* Do things with the GpiSignalObjHdl */
     int stop_clock() { return 0; }
 };
 
@@ -251,6 +252,7 @@ public:
     virtual ~GpiIterator() { }
 
     virtual Status next_handle(std::string &name, GpiObjHdl **hdl, void **raw_hdl) {
+        COCOTB_UNUSED(raw_hdl);
         name = "";
         *hdl = NULL;
         return GpiIterator::END;
--- a/cocotb/share/lib/gpi/gpi_priv.h
+++ b/cocotb/share/lib/gpi/gpi_priv.h
@@ -65,7 +65,6 @@ public:
     GpiHdl(GpiImplInterface *impl) : m_impl(impl), m_obj_hdl(NULL) { }
     GpiHdl(GpiImplInterface *impl, void *hdl) : m_impl(impl), m_obj_hdl(hdl) { }
     virtual ~GpiHdl() { }
-    virtual int initialise(std::string &name);                   // Post constructor init
 
 
     template<typename T> T get_handle() const {
--- a/cocotb/share/lib/gpi/gpi_priv.h
+++ b/cocotb/share/lib/gpi/gpi_priv.h
@@ -343,12 +343,12 @@ void gpi_embed_end();
 void gpi_embed_event(gpi_event_t level, const char *msg);
 void gpi_load_extra_libs();
 
-typedef const void (*layer_entry_func)();
+typedef void (*layer_entry_func)();
 
 /* Use this macro in an implementation layer to define an entry point */
 #define GPI_ENTRY_POINT(NAME, func) \
     extern "C" { \
-        const void NAME##_entry_point()  \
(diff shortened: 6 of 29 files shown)


cmarqu:
Does this also fix #1122?

ktbarrett:
It can, I haven't yet seen those warnings with my compilers (GCC 9.2.1 and Clang 8.0.1), but I see why they are thrown and can fix that.

ktbarrett:
I think I will also respect the `CC` and `CXX` user variables, and they will default to `cc` and `c++`.

cmarqu:
This will also help https://github.com/cocotb/cocotb/pull/796#issuecomment-462010801, good.

ktbarrett:
Sorry, don't have python2 on my machine (nor will I).

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -142,7 +141,7 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
         return;
     }
 
-    if (level < local_level)
+    if (level < (long)local_level)
```
imphil: The better type for both `level` and `local_level` is `enum gpi_log_levels`. `local_level` is already of that type, so please change the `level` parameter to be of that type as well. (Most callers of this function come through the `LOG_*` functions in `include/gpi_logging.h` which already uses the enum constants.)

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -386,7 +391,16 @@ static PyObject *register_timed_callback(PyObject *self, PyObject *args)
 
     // Extract the time
     PyObject *pTime = PyTuple_GetItem(args, 0);
-    time_ps = PyLong_AsLongLong(pTime);
+    if (PyLong_Check(pTime)) {
+        time_ps = PyLong_AsUnsignedLongLong(pTime);
```
imphil: The call `PyLong_AsUnsignedLongLong()` needs error handling, according to https://docs.python.org/3/c-api/long.html#c.PyLong_AsUnsignedLongLong.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -386,7 +391,16 @@ static PyObject *register_timed_callback(PyObject *self, PyObject *args)
...
+#if PY_MAJOR_VERSION < 3
+    } else if (PyInt_Check(pTime)) {
+        time_ps = PyInt_AsUnsignedLongLongMask(pTime);
```
imphil: As above: please add error handling

cocotb/share/makefiles/Makefile.inc
```diff
@@ -72,27 +72,31 @@ export INCLUDES := -I$(COCOTB_SHARE_DIR)/include -I$(PYTHON_INCLUDEDIR)
...
 ifeq ($(ARCH),i686)
-GXX_ARGS    := -m32
+CXX_ARGS    += -m32
 endif
 
 ifeq ($(OS),Darwin)
-GXX_ARGS	+= -g -DDEBUG -fpic
-GCC_ARGS	:= $(GXX_ARGS)
+CXX_ARGS	+= -g -DDEBUG -fpic
+C_ARGS	    += $(CXX_ARGS)
```
imphil: nit: indent

cocotb/share/makefiles/Makefile.inc
```diff
@@ -72,27 +72,31 @@ export INCLUDES := -I$(COCOTB_SHARE_DIR)/include -I$(PYTHON_INCLUDEDIR)
...
 LIB_EXT		:= dll
 PY_EXT		:= pyd
 else ifeq ($(OS),Linux)
-GXX_ARGS	+= -Werror -Wcast-qual -Wcast-align -Wwrite-strings \
-			-Wall -Wno-unused-parameter \
-			-fno-common -g -DDEBUG -fpic
-GCC_ARGS	= $(GXX_ARGS) -Wstrict-prototypes -Waggregate-return
+CXX_ARGS	+=  -fno-common -g -DDEBUG -fpic
+C_ARGS	    += $(CXX_ARGS)
```
imphil: For all of this file: please double-check the indentation

cocotb/share/makefiles/Makefile.rules
```diff
@@ -50,14 +50,18 @@ SIM_DEFINE = IUS
 endif
 
 # Add Simulator Define for compilation
-GCC_ARGS+= -D$(SIM_DEFINE)
-GXX_ARGS+= -D$(SIM_DEFINE)
+C_ARGS+= -D$(SIM_DEFINE)
+CXX_ARGS+= -D$(SIM_DEFINE)
```
imphil: nit: indentation, and please add space before `+=`

cocotb/share/lib/simulator/simulatormodule.c
eric-wieser: Why can't the same code path be used for both of these cases?

cocotb/share/makefiles/Makefile.rules
eric-wieser: meta-nit: this is alignment, not indentation

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -165,7 +166,7 @@ void embed_init_python(void)
         }
 
         LOG_ERROR("Waiting for %lu seconds - attach to PID %d with your debugger\n", sleep_time, getpid());
-        sleep(sleep_time);
+        sleep((unsigned)sleep_time);
```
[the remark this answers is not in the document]
ktbarrett: This is a narrowing cast on most archs. Clang and GCC complain with -Wconversion on Linux.

ktbarrett:
@imphil I think I've addressed most of your concerns, could you re-review?

cocotb/share/include/gpi.h
```diff
@@ -85,7 +85,7 @@ EXTERN_C_START
 typedef enum gpi_event_e {
     SIM_INFO = 0,
     SIM_TEST_FAIL = 1,
-    SIM_FAIL = 2,
+    SIM_FAIL = 2
 } gpi_event_t;
```
eric-wieser: Any reason for this change? Trailing commas are perfectly legal in C99 and C++11. Perhaps we should close out on #1087 decisively first

cocotb/share/lib/embed/gpi_embed.c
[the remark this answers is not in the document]
eric-wieser: Agreed with @ktbarrett, this is correct - although I might be tempted to spell it 
```suggestion
        sleep((unsigned int)sleep_time);
```

cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -751,7 +751,7 @@ GpiIterator::Status FliIterator::next_handle(std::string &name, GpiObjHdl **hdl,
         return GpiIterator::END;
     }
 
-    char *c_name;
+    char *c_name = NULL;
```
eric-wieser: Would it be clearer to put this in the `default` below instead?

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -90,7 +90,7 @@ int gpi_print_registered_impl()
     {
         LOG_INFO("%s registered", (*iter)->get_name_c());
     }
-    return registered_impls.size();
+    return (int)registered_impls.size();
```
eric-wieser: Can we instead change the return type to `size_t`?

cocotb/share/lib/gpi/GpiCbHdl.cpp
```diff
@@ -108,12 +108,6 @@ bool GpiHdl::is_this_impl(GpiImplInterface *impl)
     return impl == this->m_impl;
 }
 
-int GpiHdl::initialise(std::string &name)
-{
-    LOG_WARN("Generic initialize, doubt you should have called this");
-    return 0;
-}
-
```
eric-wieser: Is this function also removed from the corresponding header?

cocotb/share/lib/gpi/gpi_priv.h
```diff
@@ -65,7 +65,6 @@ class GpiHdl {
     GpiHdl(GpiImplInterface *impl) : m_impl(impl), m_obj_hdl(NULL) { }
     GpiHdl(GpiImplInterface *impl, void *hdl) : m_impl(impl), m_obj_hdl(hdl) { }
     virtual ~GpiHdl() { }
-    virtual int initialise(std::string &name);                   // Post constructor init
```
eric-wieser: No derived classes define this?

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -208,6 +209,7 @@ static PyObject *log_msg(PyObject *self, PyObject *args)
 // Remaining arguments are keyword arguments to be passed to the callback
 static PyObject *register_readonly_callback(PyObject *self, PyObject *args)
 {
+    (void)self;
```
eric-wieser: I think we might want a cleaner solution to this problem, perhaps
```
#define COCOTB_UNUSED(x) __attribute__((unused)) x
```
which you can then use inside the function signature. Worth looking at how other projects solve this problem cleanly.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -384,9 +389,26 @@ static PyObject *register_timed_callback(PyObject *self, PyObject *args)
...
+    {   // Extract the time
+        PyObject *pTime = PyTuple_GetItem(args, 0);
+        long long pTime_as_longlong = -1;
+        if (PyLong_Check(pTime)) {
+            // overflow case returns -1
+            pTime_as_longlong = PyLong_AsLongLong(pTime);
+#if PY_MAJOR_VERSION < 3
+        } else if (PyInt_Check(pTime)) {
+            pTime_as_longlong = (long long)PyInt_AsLong(pTime);
+#endif
```
eric-wieser: Why can't you use `PyLong_AsLongLong` unconditionally here?

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -384,9 +389,26 @@ static PyObject *register_timed_callback(PyObject *self, PyObject *args)
...
+        }
+        /* check negative covers negative values parsed the above functions, the overflow case with
+           PyLong_AsLongLong, and the case where the given value is neither a Long or an Int */
+        if (pTime_as_longlong < 0) {
+            PyErr_SetString(PyExc_TypeError, "Time value must be a positive integer!\n");
+            return NULL;
```
eric-wieser: I would add:
```
if (pTime_as_longlong == -1 && PyError_Occured()) {
    return NULL;
}
else if (pTime_as_longlong < 0) {
    PyErr_SetString(PyExc_ValueError, "Time value must be a positive integer!\n");
}
```

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -776,14 +815,15 @@ static PyObject *get_type(PyObject *self, PyObject *args)
         return NULL;
     }
 
-    result = gpi_get_object_type((gpi_sim_hdl)hdl);
+    result = (int)gpi_get_object_type((gpi_sim_hdl)hdl);
```
eric-wieser: I think this would be clearer if you declared `result` to be the correct type, and added the cast on the line below.

cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -327,7 +327,7 @@ int VpiSignalObjHdl::set_signal_value(long value)
 {
     s_vpi_value value_s;
 
-    value_s.value.integer = value;
+    value_s.value.integer = (int)value;
```
eric-wieser: This looks like a bug to me - although one that was already here. Worth at least calling out with a TODO and a github issue, if you don't want to think about fixing it now.

cocotb/share/makefiles/Makefile.inc
```diff
@@ -69,41 +69,45 @@ export LIB_DIR=$(BUILD_DIR)/libs/$(ARCH)
...
-GXX_ARGS    := -m32
+CXX_ARGS   += -m32
 endif
 
 ifeq ($(OS),Darwin)
-GXX_ARGS	+= -g -DDEBUG -fpic
-GCC_ARGS	:= $(GXX_ARGS)
+CXX_ARGS   := -g -DDEBUG -fpic
+C_ARGS     := $(CXX_ARGS)
 else ifeq ($(OS),Msys)
```
eric-wieser: confused why you use `+=` above then overwrite with `:=` here

cocotb/triggers.py
```diff
@@ -173,7 +173,7 @@ def __init__(self, time_ps, units=None):
     def prime(self, callback):
         """Register for a timed callback"""
         if self.cbhdl == 0:
-            self.cbhdl = simulator.register_timed_callback(self.sim_steps,
+            self.cbhdl = simulator.register_timed_callback(int(self.sim_steps),
```
eric-wieser: Why is this necessary?

cocotb/share/lib/vpi/VpiCbHdl.cpp
imphil: The correct solution would be making set_signal_value take an int instead of a long. I've opened https://github.com/cocotb/cocotb/issues/1202 to take care of that.

cocotb/share/lib/fli/FliImpl.cpp
ktbarrett: Uninitialized variables are code smells IMO. The code base is littered with them, but that's a problem for another day.

cocotb/share/lib/simulator/simulatormodule.c
ktbarrett: If all the sources were C++ you could just leave off the parameter name. But since they are not, we could use the void cast or the unused attribute. But there [doesn't seem to be any difference](https://godbolt.org/z/y8TauJ), and the attribute yields longer function specs that have to be duplicated at declaration and definition, and is an extension.

cocotb/share/lib/fli/FliImpl.cpp
eric-wieser: They're a feature once checked by a compiler - if we were to add a case to the switch below that forgets to set c_name, then either:
* This line exists, the compiler does not find our mistake
* This line does not exist, the compiler warns that c_name is used initialized

The second case seems like an obvious win to me.

cocotb/share/lib/gpi/GpiCommon.cpp
eric-wieser: CI fails because you changed the return type but forgot to remove this cast.

cocotb/share/lib/simulator/simulatormodule.c
eric-wieser: There's obviously no difference in compiler output, my argument stems from readability. To a newcomer, `UNUSED` is easier to understand than `(void)`, especially if you out an explanatory comment above the macro definition.

> function specs that have to be duplicated at declaration and definition

Are you sure this is true?

> and is an extension.

Hence the suggestion to wrap it in a macro. Searching online finds many much better macros, the one I suggested was just an example.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -394,9 +399,15 @@ static PyObject *register_timed_callback(PyObject *self, PyObject *args)
         return NULL;
     }
 
-    // Extract the time
-    PyObject *pTime = PyTuple_GetItem(args, 0);
-    time_ps = PyLong_AsLongLong(pTime);
+    {   // Extract the time
+        PyObject *pTime = PyTuple_GetItem(args, 0);
+        long long pTime_as_longlong = PyLong_AsLongLong(pTime);
+        if (pTime_as_longlong < 0) {
```
eric-wieser: Please use the logic from https://github.com/cocotb/cocotb/pull/1137#discussion_r348406932, this doesn't handle errors correctly.

cocotb/share/include/gpi.h
ktbarrett: the commit was removed

ktbarrett:
@imphil Could you have another look at this?

Issue #1008, opened 2019-07-09
Package for protocols 


themperek:
At some point, we decided to separately package protocols like Amba/Avalon/opb.

There is still a question should every protocol have a separate repository of it would be part of main cocotb repository (in the dedicated folder)?

cmarqu:
VUnit had the same question to answer recently. Their discussions about this (e.g. licenses) could be useful to us too. Here is roughly the start of it: https://gitter.im/VUnit/vunit?at=5cf3ce9b6fc5846bab4ca190

themperek:
Some arguments for the separated repository:
- can assign different maintainers
- can be more involved (more dependency on external packages like cocotb-coverage, numpy, etc..)
- can do separate CI

    Should be under placed  https://github.com/cocotb if possible. Should have listed all in documentation/READMY?

Against:
 - more complex maintenance

Anything more?

wallento:
I think now that everything is even more Pythonic we should have them as separate modules and repositories. I find the sphinx infrastructure quite efficient, where a few extensions reside in the main tree, while the majority just shares the namespace. I tried to replicate this in my cocomod-uart and cocomod-fifointerface, where cocomod is a kind of unelegant working title ;)

mciepluc:
Given the fact, that protocol-specific work may, at the end of the day, expand to standalone complex "Verification IP", and also you may imagine the situation that multiple versions of such VIPs may exist, it is definitely better to keep them in a separate packages. 
I have a dream that one day commercial versions of cocotb-based VIPs will also be available :)

wallento:
But they should be uniquely named, right? Going along with a default namespace seems like a good in my opinion and it works quite well in Sphinx. It differentiates the packages in the Python world and streamline the usage and distribution a bit.

Martoni:
Is there a «guideline» on how to write a library protocol for cocotb ?
If not, it could be a good start for developing a library «network» for Cocotb no ?

themperek:
I do not know of any guidelines. 

Do you consider some particular network protocols?

Martoni:
Not really network in fact but i'm using a SPI bus protocol for my actual project.

cmarqu:
FWIW, I know @sjalloq has worked on an SPI (master) too.

themperek:
If there is interest in SPI (and willingness to work on) maybe we do separate issue? One way this can be approached is to look at what exists in UVM and "make it better"?

Martoni:
[...]

PR #1141, opened 2019-10-09
documentation proposal for cocotb extension modules

Martoni:
This is a guideline proposal to explain how to write an extension protocol module.
Related to [this](https://github.com/cocotb/cocotb/issues/1008) discussion.

I think that a howto is necessary to encourage people to publish their modules.

This PR is mainly to start the discussion.

### DIFF
--- a/documentation/source/protocol_extension.rst
+++ b/documentation/source/protocol_extension.rst
@@ -1,0 +1,90 @@
+How to write a protocol extension
+=================================
+
+Cocotb take care of the core communication between the simulator and testbench.
+To use some known busses like wishbone, avalon, spi, i2c,... An extension is required.
+
+This is a guideline that will explain how to write a python module protocol
+extension for cocotb.
+
+Naming convention
+-----------------
+
+A cocotb extension module must begin by `cocotb_` and terminate with protocol
+name.
+
+For exemple, with spi the module extension is `cocotb_spi`.
+
+Bus driver
+----------
+
+A Bus driver is a python class that factorize common behavior and check for
+presence of some signals.
+
+A Bus driver is an active class that be driven in the main testbench. Example is
+given with avalon bus in `cocotb/drivers/avalon.py`.
+
+With AvalonMaster class we can generate stimulis for reading and writing on bus.
+
+The main class is named `BusDriver` and is defined in `cocotb/drivers/__init__.py`
+
+TODO: give an example.
+
+Bus Monitor
+-----------
+
+A Bus monitor is mainly a passive class that will monitor the bus.
+
+TODO: explain monitor
+
+Python module
+-------------
+
+The module file hierarchy should be as following ::
+
+  ├── cocotb_spi/
+  │   └── __init__.py
+  ├── README.md
+  └── setup.py
+
+With python module source code in directory cocotb_spi and module configuration
+in setup.py
+
+setup.py is a python configuration file that be used for packaging. An example
+is given above ::
+
+  try:
+    from setuptools import setup
+  except ImportError:
+    from distutils.core import setup
+
+  setup(name='cocotb_spi',
+      version='0.1',
+      py_modules=['cocotb_spi'],
+      )
+
+With this config file, python module can be installed with ::
+
+  $ python setup.py install
+
+or for development::
+
+  $ python setup.py develop
+
+
+Then module can be imported with import keyword in your testbench ::
+
+  import cocotb_spi
+
+Documentation
+-------------
+
+How to write documentation for the module correctly.
+
+Distribution
+------------
+
+Cocotb official package is hosted on github :
+https://github.com/cocotb
+
+Ask maintainer to host your module repository in it.
--- a/documentation/source/protocol_extension.rst
+++ b/documentation/source/protocol_extension.rst
@@ -37,6 +37,23 @@ A Bus monitor is mainly a passive class that will monitor the bus.
 
 TODO: explain monitor
 
+Handling Errors
+---------------
+
+It may not be clear when to raise an exception and when to use ``log.error()``.
+
+Use ``raise`` if the caller called your function in an invalid way,
+and it doesn't make sense to continue.
+
+Use ``log.error()`` if the hardware itself is misbehaving,
+and throwing an error immediately would leave it an invalid state.
+
+Even if you do eventually throw an exception (if you weren't able to do it immediately),
+you should also ``log.error()`` so that the simulation time of when things went wrong
+is recorded.
+
+TL;DR: ``log.error()`` is for humans only, raise is for both humans and code.
+
 Python module
 -------------
 
--- a/documentation/source/protocol_extension.rst
+++ b/documentation/source/protocol_extension.rst
@@ -10,10 +10,10 @@ extension for cocotb.
 Naming convention
 -----------------
 
-A cocotb extension module must begin by `cocotb_` and terminate with protocol
+A cocotb extension module must begin by `cocotbext.` and terminate with protocol
 name.
 
-For exemple, with spi the module extension is `cocotb_spi`.
+For exemple, with spi the module extension is `cocotbext.spi`.
 
 Bus driver
 ----------
@@ -33,9 +33,10 @@ TODO: give an example.
 Bus Monitor
 -----------
 
-A Bus monitor is mainly a passive class that will monitor the bus.
+A Bus monitor is mainly a passive class that will monitor the bus. A bus monitor
+will record some values on the bus and emit errors if wrong state occur.
 
-TODO: explain monitor
+TODO: give an example.
 
 Handling Errors
 ---------------
@@ -59,8 +60,9 @@ Python module
 
 The module file hierarchy should be as following ::
 
-  ├── cocotb_spi/
-  │   └── __init__.py
+  ├── cocotbext/
+  │   └── spi/
+  │       └── __init__.py
   ├── README.md
   └── setup.py
 
@@ -72,10 +74,16 @@ is given above ::
 
   from setuptools import setup
 
-  setup(name='cocotb_spi',
+  setup(name='cocotbext.spi',
+        use_scm_version={
+            "relative_to": __file__,
+            "write_to": "cocotbext/spi/version.py",
+        },
         version='0.1',
-        py_modules=['cocotb_spi'],
         install_requires=['cocotb'],
+        setup_requires=[
+            'setuptools_scm',
+        ],
         classifiers=[
           "Programming Language :: Python :: 3",
           "License :: OSI Approved :: MIT License",
@@ -84,7 +92,7 @@ is given above ::
 
 With this config file, python module can be installed with::
 
-  $ python -m pip install cocotb_spi
+  $ python -m pip install cocotbext-spi
 
 or for development (in package directory)::
 
@@ -92,7 +100,7 @@ or for development (in package directory)::
 
 Then module can be imported with import keyword in your testbench ::
 
-  import cocotb_spi
+  import cocotbext.spi
 
 Documentation
 -------------
--- a/documentation/source/protocol_extension.rst
+++ b/documentation/source/protocol_extension.rst
@@ -53,24 +53,25 @@ in setup.py
 setup.py is a python configuration file that be used for packaging. An example
 is given above ::
 
-  try:
-    from setuptools import setup
-  except ImportError:
-    from distutils.core import setup
+  from setuptools import setup
 
   setup(name='cocotb_spi',
-      version='0.1',
-      py_modules=['cocotb_spi'],
-      )
+        version='0.1',
+        py_modules=['cocotb_spi'],
+        install_requires=['cocotb'],
+        classifiers=[
+          "Programming Language :: Python :: 3",
+          "License :: OSI Approved :: MIT License",
+          "Operating System :: OS Independent",
+          "Topic :: Scientific/Engineering :: Electronic Design Automation (EDA)"])
 
-With this config file, python module can be installed with ::
+With this config file, python module can be installed with::
 
-  $ python setup.py install
+  $ python -m pip install cocotb_spi
 
-or for development::
+or for development (in package directory)::
 
-  $ python setup.py develop
-
+  $ python -m pip install -e .
 
 Then module can be imported with import keyword in your testbench ::
 
--- a/documentation/source/index.rst
+++ b/documentation/source/index.rst
@@ -20,6 +20,7 @@ Contents:
    examples
    troubleshooting
    simulator_support
+   protocol_extension
    roadmap
    release_notes
 


documentation/source/protocol_extension.rst
```diff
@@ -0,0 +1,90 @@
...
+
+setup.py is a python configuration file that be used for packaging. An example
+is given above ::
+
+  try:
+    from setuptools import setup
+  except ImportError:
+    from distutils.core import setup
+
+  setup(name='cocotb_spi',
```
themperek: Should the package be called `cocotb-spi`? We have few in such convention.

documentation/source/protocol_extension.rst
```diff
@@ -0,0 +1,90 @@
...
+      version='0.1',
+      py_modules=['cocotb_spi'],
+      )
+
+With this config file, python module can be installed with ::
+
+  $ python setup.py install
```
themperek: Another option is `pip install [-e] ` that would be consistent with cocotb documentation.

themperek:
In general, it sound like a good start for me.

documentation/source/protocol_extension.rst
```diff
@@ -0,0 +1,90 @@
...
+-----------
+
+A Bus monitor is mainly a passive class that will monitor the bus.
+
+TODO: explain monitor
+
+Python module
+-------------
+
+The module file hierarchy should be as following ::
```
eric-wieser: I think we should link to the pypa distribution docs here, rather than repeating them

documentation/source/protocol_extension.rst
```diff
@@ -0,0 +1,90 @@
...
+With python module source code in directory cocotb_spi and module configuration
+in setup.py
+
+setup.py is a python configuration file that be used for packaging. An example
+is given above ::
+
+  try:
+    from setuptools import setup
+  except ImportError:
+    from distutils.core import setup
```
eric-wieser: Does cocotb do this? If not, we may as well assume setuptools

documentation/source/protocol_extension.rst
```diff
@@ -0,0 +1,90 @@
...
+  $ python setup.py install
+
+or for development::
+
+  $ python setup.py develop
+
+
+Then module can be imported with import keyword in your testbench ::
+
+  import cocotb_spi
```
eric-wieser: An alternative would be `cocotb.ext.spi`

documentation/source/protocol_extension.rst
Martoni: Yes, '-' is better I think. I like the eric proposition with `cocotb.ext.spi` but is it possible with an external package ?

documentation/source/protocol_extension.rst
Martoni: I discover the pip option -e, yes it better.

documentation/source/protocol_extension.rst
Martoni: As I understand setuptools, the package `distutils.core` is for old python version (python2) and new python version use from `setuptools`.

documentation/source/protocol_extension.rst
Martoni: If it's possible with an external package it could be better yes

documentation/source/protocol_extension.rst
cmarqu: Make sure the ``setup.py`` contains lines like these two: https://github.com/wallento/cocomod-uart/blob/64f7af8922ea67efda4a1a3c185ce2ecaa45e413/setup.py#L18 and https://github.com/wallento/cocomod-uart/blob/64f7af8922ea67efda4a1a3c185ce2ecaa45e413/setup.py#L26.

documentation/source/protocol_extension.rst
eric-wieser: PEP420 makes it easy to make `cocotb.ext` a namespace package, which allows multiple distributions (things with a `setup.py`) to exist within the same package - all that is required is that third-party extensions do _not_ include a `cocotb/__init__.py` or a `cocotb/ext/__init__.py`.

documentation/source/protocol_extension.rst
Martoni: I discover the module cocomod-uart. Should not it be referenced in github [cocotb organisation ](https://github.com/cocotb)?

documentation/source/protocol_extension.rst
Martoni: have you got a template example for that @eric-wieser ? I'm trying to do that without success [here]( https://github.com/Martoni/cocotb_ext_spi.git).

Martoni:
I improved the document according to discussion. I'm trying it with this package.
I can't manage to setup @eric-wieser proposal : cocotb.ext.spi

documentation/source/protocol_extension.rst
eric-wieser: This is an entirely different question - there are two things to decide here:

* The module name, as in `import module_name`. This is what goes in `py_modules`, and the name of the python file
* The distribution name, as in `pip install distribution-name`. This is what goes here in setup.py

@themperek is arguing that we don't need to choose these two names to be the same. While this is true, users might find it easier to realize that `import cocotb_spi` means they need to run `pip install cocotb_spi` instead of `pip install cocotb-spi`, so it might not be worth making the names different.

I'll see if I can get anywhere with an example project using `cocotb.ext.spi`.

documentation/source/protocol_extension.rst
eric-wieser: It's looking like that might be a bad idea after all. I'll try to collect some example of other projects choosing to do this here:

* `flask.ext` was removed: https://github.com/pallets/flask/issues/1135#issuecomment-61860862

documentation/source/protocol_extension.rst
cmarqu: > I discover the module cocomod-uart. Should not it be referenced in github [cocotb organisation ](https://github.com/cocotb)?

It's already kinda tied in by this new "Used by:" button that GitHub shows in the top row that leads to https://github.com/cocotb/cocotb/network/dependents

documentation/source/protocol_extension.rst
Martoni: Ok, glad to hear that. It was really painful to do (without success !)

documentation/source/protocol_extension.rst
Martoni: @wallento use `cocomod` as prefix for his [protocol package](https://github.com/wallento/cocomod-uart). Is there other package that using this `cocomod` name or is it only name from @wallento packages ?

documentation/source/protocol_extension.rst
eric-wieser: A much simpler option would be `cocotb_ext.spi`, but I'm not sure that's an improvement over `cocotb_spi`

Martoni:
I wrote a minimal module for SPI test with some documentation explaination [here](https://github.com/Martoni/cocomod-spi).
I followed the wallento module [cocomod-uart](https://github.com/wallento/cocomod-uart). It seemed the most correspond to the desired extension model.
All comments and contribution are welcome :)

documentation/source/protocol_extension.rst
wallento: And yes, it should use hypen ('-') for the package name

wallento: Ah, sorry, in sphinx its sphinxcontrib, sphinx.ext is in the source tree too

eric-wieser:
IMO the module prefix should contain `cocotb`, which makes it very clear that its for interop with cocotb, unlike `cocomod` which is less clear.

Martoni:
@eric-wieser 
>   IMO the module prefix should contain cocotb, 

Then for uart example : 
- `cocotb-uart` ?
- `cocotb-ext-uart` ?
- `cocotb-mod-uart` ?

Or with dot '.' : 
- `cocotb.uart` ?
- `cocotb.ext.uart` ?
- `cocotb.mod.uart` ?

But it's more a problem with dot to package it correctly through Python2/Python3 no ?

wallento:
The package name can be the first. The second is how the modules are organized and "cocotb.*" doesn't work as it would require the main cocotb package to not use the toplevel `__init__.py`. I don't want to rule it out, but it extremely inconvenient and most probably breaks things.

@eric-wieser understood, how about cocotbmod.* cocotbext.*? The idea of cocomod was that its not the testbench (cocotb) but the modules for the testbench (cocomod). Any other suggestions?

eric-wieser:
Removing `cocotb/__init__.py` sounds like a non-starter to me - I'm happy to outright rule it out.

Happy with either, although would consider an underscore, as `cocotb_mod.mymod`, `cocotb_ext.mymod`, or perhaps to align with Sphinx, `cocotb_contrib.mymod`.

The latter has the advantage of being a coverall that spans both testbench modules (eg protocol drivers), extensions (eg wavedrom-like things), and miscellaneous utilities (eg constrained randomizers)

wallento:
I think `cocotbext` is the best, no underscore. Let's maybe give it a first try for a couple of modules and we can change it later anyhow.

Fatsie:
I don't like the *contrib proposal. As hinted during ORConf I am working on Wishbone support for cocotb. I don't mind become maintainer of that but I would still like it to be under the cocotb umbrella and have the repo hosted under https://github.com/cocotb.

eric-wieser:
@Fatsie: To be clear, are you expressing opposition to the package name being `cocotb_contrib`, or to creating a separate github org like https://github.com/sphinx-contrib

Fatsie:
I prefer `ext` or `mod` and don't like `contrib` in the name.
I propose to host the wishbone repo under https://github.com/cocotb/cocotbext-wishbone

wallento:
okay, so lets go with this naming. I have prepared the infrastructure with the old naming here https://github.com/wallento/cocomod-wishbone based on @Fatsie's repository. We should prepare the infrastructure there and start playing with it.

Fatsie:
Stefan Wallentowitz schreef op do 14-11-2019 om 06:49 [-0800]:
> okay, so lets go with this naming. I have prepared the infrastructure with the old naming here https://github.com/wallento/cocomod-wishbone based on @Fatsie's repository. We should prepare the infrastructure there and start playing with it.

One of the annoying aspects of having separate repos for each module is that each repo will need to setup their own unit test framework and CI/CD...

greets,
Staf.

eric-wieser:
@Fatsie: We could copy `sphinx-contrib` here and create a cookie-cutter project, like https://github.com/sphinx-contrib/cookiecutter

They've pinned it on their org page too, https://github.com/sphinx-contrib, to make it easy to find.

imphil:
So let's try to wrap this up.


**Related work:**
- Flask ([Docs](https://flask.palletsprojects.com/en/1.1.x/extensiondev/)): flask_* as package name, flask-* or *-flask as distribution name.
- Sphinx [doesn't document](https://www.sphinx-doc.org/en/master/extdev/index.html) the naming of external extensions, but de-facto most extensions use `sphinxcontrib.*` as package name, and `sphinxcontrib-*` as distribution ("pip") name. [For in-tree extensions](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-extensions) Sphinx uses the `sphinx.ext.*`  namespace. The namespacing [is implemented as pkg_resource-style namespace packages](https://github.com/sphinx-contrib/cookiecutter/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/sphinxcontrib/__init__.py#L12) (now deprecated).

**Package name: Use namespaces?**
  - Sphinx uses them and it seems to work for them.
  - Flask did use them, and moved back to non-namespaced packages (https://github.com/pallets/flask/pull/1484).

**Package name: Use subnamespaces? `cocotbXXX.*` or `cocotb.XXX.*`?**
Subnamespaces (`cocotb.XXX.*`) require modifications/removal of the `cocotb/__init__.py` file, so that's not really an option.
  
**Implementing namespaced packages**
All options are outlined in https://packaging.python.org/guides/packaging-namespace-packages/.

  - Native namespace packages, as described in PEP420, are not supported in Python 2.7. They are compatible with pkgutil-style namespaced packages (below), so this scheme can be used for Python3-only packages.
  - ["pkgutil-style namespace packages"](https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages) require an idential `__init__.py` everywhere, but should work in all supported configurations.
  - [pkg_resources-style namespace packages](https://packaging.python.org/guides/packaging-namespace-packages/#pkg-resources-style-namespace-packages) are deprecated.

**Hosting/Discoverability**
- No matter where extensions are hosted, they are discoverable in GitHub https://github.com/cocotb/cocotb/network/dependents and similar sites (e.g. https://libraries.io/pypi/cocotb/dependents)
- Personal opinion: It's not important if an extension is hosted in the cocotb GitHub organization; we can do that for some extensions if extension authors prefer that, but it's not a requirement or even recommended.

**Naming**
Let's come back to that later once we have answers to the namespace or not question.


So let's tackle this step-by-step to get consensus. **First: Should we use namespaces for package names, or not? That is, do you prefer and/or have good reasons for:**

```python3
# with namespaces (sphinx-style)
import cocotbXXX.myextension

# or
# without namespaces (flask-style)
import cocotbXXX_myextension
```

documentation/source/protocol_extension.rst
ktbarrett: @eric [you can't extend from the cocotb namespace](https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages). We have actually tried this recently.

wallento:
I prefer namespaces as it is a clean nice solution. Did you see any information about what exactly didn't work with flask? Re implementation I don't mind about 2.7, but pkg-util style is a good thing and it has the advantage that it explicitly codes the information.

So, namespace +1

Fatsie:
I also prefer use of namespace.

Martoni:
Namespace seems to be clear. But if no namespace easier it's ok for me also.

documentation/source/protocol_extension.rst
```diff
@@ -0,0 +1,91 @@
...
+
+TODO: give an example.
+
+Bus Monitor
+-----------
+
+A Bus monitor is mainly a passive class that will monitor the bus.
+
+TODO: explain monitor
+
```
cmarqu: I feel here is a good place to add the following information based on https://github.com/cocotb/cocotb/pull/1190#discussion_r347848900

```
Handling Errors
---------------

It may not be clear when to raise an exception and when to use ``log.error()``.

Use ``raise`` if the caller called your function in an invalid way,
and it doesn't make sense to continue.

Use ``log.error()`` if the hardware itself is misbehaving,
and throwing an error immediately would leave it an invalid state.

Even if you do eventually throw an exception (if you weren't able to do it immediately),
you should also ``log.error()`` so that the simulation time of when things went wrong
is recorded.

TL;DR: ``log.error()`` is for humans only, raise is for both humans and code.
```

imphil:
Thanks for the vote. So namespaces it is: `cocotbXXX.myextension`. 

Now step 2: what should XXX be?

Options we've heard so far:
- `cocomod.*`
- `cocotbmod.*`
- `cocotbext.*`
- `cocotbcontrib.*`

Please let me know your preferences!

cmarqu:
I have a slight preference for ``cocotb_ext``.

imphil:
So on the namespace names, we have the following votes so far:

- cocotb_ext.*: @themperek, @cmarqu 
- cocotbext.*: @wallento, @imphil

imphil:
Moving on to the final naming part, package names:

If we go with cocotb_ext as namespace, we could have:

- cocotb-ext-YOURPACKAGE
- cocotb_ext-YOURPACKAGE

If we go with cocotbext as namespace, we would have:
- cocotbext-YOURPACKAGE


Any opinions on that? This should be the last vote before we are done with naming.

Fatsie:
I prefer cocotbext.*

eric-wieser:
> If we go with cocotb_ext as namespace, we could have

`cocotb_ext.YOURPACKAGE` is a totally legal name too - see https://pypi.org/project/backports.functools-lru-cache/

imphil:
@eric-wieser what do you mean by "github package name"? The name of the repository, including the question of where the code lives? That would be a question for another day, taking one step at a time here.

eric-wieser:
@imphil: exactly. Edited my comment, feel free to edit it more as you see fit to focus discussion

Martoni:
I prefer cocotbext.* too

ktbarrett:
I like either `cocotb_ext` or `cocotbext`, preference is for the former.

documentation/source/protocol_extension.rst
Martoni: Added

Martoni:
I updated PR and my spi module for following naming convention:
package name:

>  cocotbext-spi

import command :  
```
import cocotbext.spi
```
The cocotbext-spi module can be [seen here](https://github.com/Martoni/cocotbext-spi).

imphil:
Thanks for the discussion everyone, and thanks for starting this @Martoni! 

The cocotb maintainers have discussed this today and we have finalized the decision on naming:

- "cocotbext" for the namespace
- "cocotbext-" to prefix the distribution (package) name.

This PR already has the right naming, so that part is good to go.

When I looked through the text I saw a couple grammar/wording nits I'd like to clean up, it's probably easier to do that right away and not comment on each small thing individually. I'll do that tonight and update this PR with my changes.

imphil:
I did reword the contents in this PR and opened https://github.com/cocotb/cocotb/pull/1344 with the updated wording. I thought it would be clearer to do that in a separate PR and not overwrite the changes done in this PR. @Martoni can you pleae have a look there?

PR #1142, opened 2019-10-10
Force event.set() to append the transaction

sjalloq:
- when calling Monitor::_recv(self,trxn), the trxn was not pushed into
   self._event when the Event.set method was called.


imphil:
LGTM. This is actually new behavior, the current documentation doesn't imply that the transaction is set as data. But the behavior makes sense, and I've opened #1146 to get it documented.

PR #1145, opened 2019-10-12
Add python_requires to setup().

### LINKED ISSUE #1005: Packaging: Add python_requires to manifest
Define our Python version requirements in our package manifest, as described here: https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

cmarqu:
Closes #1005.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,7 @@ setup(
     author='Chris Higgs, Stuart Hodgson',
     author_email='cocotb@potentialventures.com',
     install_requires=[],
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
     packages=find_packages(),
     include_package_data=True,
     package_data={'cocotb': package_files('cocotb/share')},


setup.py
```diff
@@ -55,6 +55,7 @@ def package_files(directory):
     author='Chris Higgs, Stuart Hodgson',
     author_email='cocotb@potentialventures.com',
     install_requires=[],
+    python_requires='=2.7, >=3.5',
```
imphil: =2.7 should be ==2.7, see https://www.python.org/dev/peps/pep-0440/#examples

cmarqu:
Ugh, sorry for the many commits... Not even the simplest changes should be done on a cell phone.

imphil:
What's the problem with `==2.7, >=3.5`?

cmarqu:
> What's the problem with `==2.7, >=3.5`?

It did not work (https://travis-ci.com/cmarqu/cocotb/builds/131656099). For instance, ``==2.7`` means 2.7 and 2.7.0, but not 2.7.15.

imphil:
Use `~=2.7` then, at least thats what the docs suggest. And please rebase into to a single commit.

cmarqu:
``~=2.7, ...`` did not work (see https://travis-ci.com/cmarqu/cocotb/builds/131711726), so I reverted back to ``>=``. 
The specifier format feels intuitive at first since the comma means **and**.

imphil:
Ah OK, the comma behavior is slightly unexpected. Thanks for trying!

PR #1146, opened 2019-10-12
Improve API docs of Monitor class

imphil:
Small fixes to make the apidocs more accurate and render properly.

### DIFF
--- a/cocotb/monitors/__init__.py
+++ b/cocotb/monitors/__init__.py
@@ -36,15 +36,16 @@ the transactions.
 from collections import deque
 
 import cocotb
-from cocotb.decorators import coroutine
-from cocotb.triggers import Event, Timer
 from cocotb.bus import Bus
+from cocotb.decorators import coroutine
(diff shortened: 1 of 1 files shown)


cocotb/monitors/__init__.py
```diff
@@ -68,17 +69,17 @@ class Monitor(object):
         callback (callable): Callback to be called with each recovered transaction
             as the argument. If the callback isn't used, received transactions will
             be placed on a queue and the event used to notify any consumers.
-        event (event): Object that supports a ``set`` method that will be called when
+        event (cocotb.triggers.Event): (Subclass of) Event that will be called when
```
eric-wieser: Not sure (subclass of) is interesting here - it's implied pretty much everywhere, and it's unlikely that someone would subclass event anyway, I think.

imphil: Good point, I updated the text accordingly.

PR #1149, opened 2019-10-14
Use the C++11 override keyword on virtual functions, for clarity.

eric-wieser:
By doing this, it means that a typo in a virtual function signature results in a compilation error, rather than silently not overriding the intended function.

Also:
* removes a bunch of empty destructors which are the compiler-generated default.
* configures the compiler to require the use of the override keyword

Note that I was only able to test the `Vpi*` changes locally, the `Vhpi*` and `Fli*` changes have not been verified by a compiler.

### DIFF
--- a/cocotb/share/lib/fli/FliImpl.h
+++ b/cocotb/share/lib/fli/FliImpl.h
@@ -51,10 +51,8 @@ public:
     FliProcessCbHdl(GpiImplInterface *impl) : GpiCbHdl(impl),
                                               m_proc_hdl(NULL),
                                               m_sensitised(false) { }
-    virtual ~FliProcessCbHdl() { }
 
-    virtual int arm_callback() = 0;
-    virtual int cleanup_callback();
+    int cleanup_callback() override;
 
 protected:
     mtiProcessIdT       m_proc_hdl;
@@ -69,9 +67,8 @@ public:
                    FliSignalObjHdl *sig_hdl,
                    unsigned int edge);
 
-    virtual ~FliSignalCbHdl() { }
-    int arm_callback();
-    int cleanup_callback() {
+    int arm_callback() override;
+    int cleanup_callback() override {
         return FliProcessCbHdl::cleanup_callback();
     }
 
@@ -86,9 +83,8 @@ public:
     FliSimPhaseCbHdl(GpiImplInterface *impl, mtiProcessPriorityT priority) : GpiCbHdl(impl),
                                                                              FliProcessCbHdl(impl),
                                                                              m_priority(priority) { }
-    virtual ~FliSimPhaseCbHdl() { }
 
-    int arm_callback();
+    int arm_callback() override;
 
 protected:
     mtiProcessPriorityT         m_priority;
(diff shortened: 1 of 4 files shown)


cocotb/share/makefiles/Makefile.inc
```diff
@@ -92,11 +92,13 @@ else ifeq ($(OS),Linux)
 GXX_ARGS	+= -Werror -Wcast-qual -Wcast-align -Wwrite-strings \
 			-Wall -Wno-unused-parameter \
 			-fno-common -g -DDEBUG -fpic
-GCC_ARGS	= $(GXX_ARGS) -Wstrict-prototypes -Waggregate-return
+GCC_ARGS	:= $(GXX_ARGS) -Wstrict-prototypes -Waggregate-return
```
eric-wieser: This is by itself a bug - note all the other cases use `:=` to perform the copy. If desirable, I can move this to its own PR.

ktbarrett:
This commit bumps requirements up on the user to have a C++11 compiler and compatible simulator, which should be pull request in itself.

cocotb/share/makefiles/Makefile.inc
eric-wieser: Done in #1162

eric-wieser:
@ktbarrett: I think that was handled by #1108?

ktbarrett:
If you are going to call out the C++ standard, it's a good idea to call out the C standard as well. GCC 4.8.1 does not fully support C11, so the C standard should be `gnu99`. The GNU extensions are needed to enable POSIX standards required by Python, and are supported by Clang as well.

themperek:
Default on RHEL 7 (gcc 4.8.5) one gets :
```
g++ -Werror -Wcast-qual -Wcast-align -Wwrite-strings -Wall -Wno-unused-parameter -fno-common -g -DDEBUG -fpic -std=c++11 -Wsuggest-override -DVPI_CHECKING -DLIB_EXT=so -DSINGLETON_HANDLES -DMODELSIM -c -I/cocotb/cocotb/share/include -I/home/themperek/miniconda/include/python3.7m  -o /cocotb/examples/dff/tests/build/obj/x86_64/GpiCbHdl.o GpiCbHdl.cpp
g++: error: unrecognized command line option ‘-Wsuggest-override’
```

If used with `devtoolset-7` (gcc 7.3.1) it works for VPI and FLI but for VHPI gets:
```
g++ -Werror -Wcast-qual -Wcast-align -Wwrite-strings -Wall -Wno-unused-parameter -fno-common -g -DDEBUG -fpic -std=c++11 -Wsuggest-override -DVHPI_CHECKING -DIUS -c -I/cocotb/cocotb/share/include -I/miniconda/include/python3.7m  -o /cocotb/examples/dff/tests/build/obj/x86_64/VhpiImpl.o VhpiImpl.cpp
In file included from VhpiImpl.cpp:28:0:
VhpiImpl.h:102:13: error: ‘virtual int VhpiValueCbHdl::cleanup_callback()’ can be marked override [-Werror=suggest-override]
         int cleanup_callback() {
             ^~~~~~~~~~~~~~~~
VhpiImpl.h:115:13: error: ‘virtual int VhpiTimedCbHdl::cleanup_callback()’ can be marked override [-Werror=suggest-override]
         int cleanup_callback();
             ^~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

cocotb/share/makefiles/Makefile.inc
```diff
@@ -97,6 +97,8 @@ else
 $(error "Unsupported os " $(OS))
 endif
 
+GXX_ARGS += -std=c++11 -Wsuggest-override
```
imphil: GCC 4.8.5 is supported by cocotb and doesn't support `-Wsuggest-override`, so this needs to be removed. (https://github.com/cocotb/cocotb/pull/1149#issuecomment-560113536)

eric-wieser: We need this flag in place for useful CI, but agreed that we should support older GCC - I'll try to find a better place for it.

eric-wieser: Tried to make this conditional on the compiler version

eric-wieser:
@themperek, could you test both with and without `devtoolset-7`, using VHPI?

themperek:
It looks to compile fine now. 

PS. I have some issues to run on my setup but not related to this. Will try to investigate.

eric-wieser:
Since #1326 is merged, it doesn't really matter if this causes warnings on FLI and VHPI - especially since those interfaces no longer build without warnings anyway.

I think this is good to merge, @imphil? At any rate, it enforces stricter correctness for Icarus

cocotb/share/makefiles/Makefile.inc
```diff
@@ -76,6 +76,14 @@ BASE_WARNS := -Wall -Wextra -Wconversion -Wcast-qual -Wwrite-strings
 CC_WARNS   := $(BASE_WARNS) -Wstrict-prototypes -Waggregate-return
 CXX_WARNS  := $(BASE_WARNS) -Wnon-virtual-dtor -Woverloaded-virtual
 
+# version-specific flags
+# easiest to compare versions with python
+version_less_than_equal = $(shell $(PYTHON_BIN) -c "import sys; from distutils.version import LooseVersion; print(LooseVersion(sys.argv[1]) <= LooseVersion(sys.argv[2]))" $(1) $(2))
+GXX_VERSION = $(shell g++ -dumpversion)
+ifeq ($(call version_less_than_equal,4.9,$(GXX_VERSION)),True)
+CXX_WARNS += -Wsuggest-override
```
imphil: That won't fly, unfortunately. People can use clang and other compilers. I see two ways out:
- Remove this check, and only test in CI by modifying CXX_ARGS (preferred).
- Test for the feature instead of checking the version.

eric-wieser: I was under the impression that we had g++ hard-coded anyway, but perhaps we don't want to entrench ourselves further there.

imphil: We had, but the recent work by @ktbarrett removed that explicit dependency.

eric-wieser:
> I'm ready to take only the code change and sort out the compiler flags at a later point if you prefer

Let's do that, I'm getting tired of fixing conflicts in the C++ code. Updated without the makefile piece.

imphil:
Thanks for your work on this!

eric-wieser:
CI failure is #1367, let's put this in

PR #1152, opened 2019-10-16
Add Python 3.8 for Linux to Travis.

cmarqu:


eric-wieser:
Perhaps we should remove the windows 3.7 testing, since we were only testing the latest version on windows before.

themperek:
At the moment on windows it will use latest version that is distributed with conda so adding 3.8 will not make any difference. We can change this but maybe with sperate PR.

.travis.yml
```diff
@@ -82,6 +90,15 @@ jobs:
       script:
         - docker run cocotb powershell 'python setup.py install ; make test'
 
+    - stage: SimTests
+      python: 3.8
```
imphil: That's not having any effect, if you want to use a different Python version, you need to do so in the Dockerfile.windows. 

I'd suggest removing the Windows part for now, Linux Python 3.8 testing is already an important step forward.

cmarqu: Okay, Linux only it is now.

PR #1153, opened 2019-10-17
Use command(1) rather than which(1) in Makefiles

emilazy:
which(1) is not part of the POSIX standard, and is not present on
several systems and build environments out of the box (e.g. the
Nix sandbox). It's not actually part of GNU coreutils as many assume;
GNU has its own version (http://www.gnu.org/software/which/) but I
believe Debian might ship its own implementation.

`type -p` is a commonly-suggested alternative, but its output is
unfortunately not stable across systems. `command -v` is fully
portable across all POSIX systems.

We need to use $(shell :; command ...) because make tries to exec
programs directly rather than use the shell where possible, but the
logic for it is broken on old versions (omitting command(1) among other
shell builtins). This breaks things when command(1) is available only
as a shell builtin and not as an external command.

References:
* https://twitter.com/whitequark/status/1168670168173559808
* https://twitter.com/dev_console/status/1168672291271692288
* https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html
* https://git.savannah.gnu.org/cgit/make.git/tree/job.c?h=4.2&id=ec6198012261e11b72bd5a95bd9e5c4d895c5f22#n2433

### DIFF
--- a/cocotb/share/makefiles/Makefile.pylib.Msys
+++ b/cocotb/share/makefiles/Makefile.pylib.Msys
@@ -31,7 +31,7 @@
 
 # if not explicitly set, auto-detect python dir from system path
 ifeq ($(PYTHON_DIR),)
-	PYTHON_DIR ?= $(shell dirname $(shell which python))
+	PYTHON_DIR ?= $(shell dirname $(shell :; command -v python))
 endif
 
 # make sure python dir was set properly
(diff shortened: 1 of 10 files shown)


imphil:
Fixed in 114be2ed788ef3a0df7e40ed96d4af3f3955f5aa

PR #1164, opened 2019-10-22
Cleanup Python refcounts and call Py_Finalize

### LINKED ISSUE #1055: Py_Finalize() is never called

garmin-mjames:
While working on getting cocotb to work with simulator restart, this cleanup was part of getting Py_Finalize() to not segfault.

Clean up embed_sim_init() to properly account for Python object references:
Add comment when new reference is created and properly decrement references when no longer needed.
Add comment when a function steals a reference.
Use exported Py_IncRef/Py_DecRef functions. These are exports of XINCREF/XDECREF and handle NULL pointers.

Cleanup GPI and Python on shutdown:
Clean up GpiHandleStore.
Clean up Python object reference counts and reset pointers to NULL.
Call Py_Finalize on shutdown from both successful and failed runs.
Release VHPI handles when handle objects are destroyed.

Fixes #1055.

I'm not sure if this is going to work with other simulators (tested with Aldec).

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -775,9 +775,13 @@ class Scheduler(object):
         """Directly call into the regression manager and end test
            once we return the sim will close us so no cleanup is needed.
         """
-        self.log.debug("Issue sim closedown result to regression object")
-        self._test.abort(exc)
-        cocotb.regression_manager.handle_result(self._test)
+        # If there is an error during cocotb initialization, self._test may not
+        # have been set yet. Don't cause another Python exception here.
+
+        if self._test:
+            self.log.debug("Issue sim closedown result to regression object")
+            self._test.abort(exc)
+            cocotb.regression_manager.handle_result(self._test)
 
     def cleanup(self):
         """Clear up all our state.
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -228,12 +228,12 @@ int embed_sim_init(gpi_sim_info_t *info)
         }
     }
 
-    PyObject *cocotb_module, *cocotb_init, *cocotb_args, *cocotb_retval;
+    PyObject *cocotb_module, *cocotb_init, *cocotb_retval;
     PyObject *simlog_obj, *simlog_func;
-    PyObject *argv_list, *argc, *arg_dict, *arg_value;
+    PyObject *argv_list;
 
     cocotb_module = NULL;
-    arg_dict = NULL;
+    simlog_obj = NULL;
 
     // Ensure that the current thread is ready to call the Python C API
     PyGILState_STATE gstate = PyGILState_Ensure();
@@ -243,124 +243,163 @@ int embed_sim_init(gpi_sim_info_t *info)
         goto cleanup;
 
     // Obtain the loggpi logger object
-    simlog_obj = PyObject_GetAttrString(cocotb_module, "loggpi");
+    simlog_obj = PyObject_GetAttrString(cocotb_module, "loggpi");       // New reference
 
     if (simlog_obj == NULL) {
         PyErr_Print();
         LOG_ERROR("Failed to get simlog object from loggpi\n");
+        goto cleanup;
     }
 
-    simlog_func = PyObject_GetAttrString(simlog_obj, "_logFromC");
+    // Obtain the function to use when logging from C code
+    simlog_func = PyObject_GetAttrString(simlog_obj, "_logFromC");      // New reference
     if (simlog_func == NULL) {
         PyErr_Print();
         LOG_ERROR("Failed to get the _logFromC method");
         goto cleanup;
     }
-
     if (!PyCallable_Check(simlog_func)) {
-        PyErr_Print();
         LOG_ERROR("_logFromC is not callable");
+        Py_DECREF(simlog_func);
         goto cleanup;
     }
 
-    set_log_handler(simlog_func);
+    set_log_handler(simlog_func);                                       // Note: This function steals a reference to simlog_func.
 
-    Py_DECREF(simlog_func);
-
-    simlog_func = PyObject_GetAttrString(simlog_obj, "isEnabledFor");
+    // Obtain the function to check whether to call log function
+    simlog_func = PyObject_GetAttrString(simlog_obj, "isEnabledFor");   // New reference
     if (simlog_func == NULL) {
         PyErr_Print();
         LOG_ERROR("Failed to get the isEnabledFor method");
         goto cleanup;
     }
-
     if (!PyCallable_Check(simlog_func)) {
-        PyErr_Print();
         LOG_ERROR("isEnabledFor is not callable");
+        Py_DECREF(simlog_func);
         goto cleanup;
     }
 
-    set_log_filter(simlog_func);
+    set_log_filter(simlog_func);                                        // Note: This function steals a reference to simlog_func.
 
-    argv_list = PyList_New(0);
+    // Build argv for cocotb module
+    argv_list = PyList_New(info->argc);                                 // New reference
+    if (argv_list == NULL) {
+        PyErr_Print();
+        LOG_ERROR("Unable to create argv list");
+        goto cleanup;
+    }
     for (i = 0; i < info->argc; i++) {
-        arg_value = PyString_FromString(info->argv[i]);
-        PyList_Append(argv_list, arg_value);
+        PyObject *argv_item = PyString_FromString(info->argv[i]);       // New reference
+        if (argv_item == NULL) {
+            PyErr_Print();
+            LOG_ERROR("Unable to create Python string from argv[%d] = \"%s\"", i, info->argv[i]);
+            Py_DECREF(argv_list);
+            goto cleanup;
+        }
+        PyList_SET_ITEM(argv_list, i, argv_item);                       // Note: This function steals the reference to argv_item
     }
 
-    arg_dict = PyModule_GetDict(cocotb_module);
-    PyDict_SetItemString(arg_dict, "argv", argv_list);
-
-    argc = PyInt_FromLong(info->argc);
-    PyDict_SetItemString(arg_dict, "argc", argc);
-
-    if (!PyCallable_Check(simlog_func)) {
+    // Add argv list to cocotb module
+    if (-1 == PyModule_AddObject(cocotb_module, "argv", argv_list)) {   // Note: This function steals the reference to argv_list if successful
         PyErr_Print();
-        LOG_ERROR("_printRecord is not callable");
+        LOG_ERROR("Unable to set argv");
+        Py_DECREF(argv_list);
+        goto cleanup;
+    }
+
+    // Add argc to cocotb module
+    if (-1 == PyModule_AddIntConstant(cocotb_module, "argc", info->argc)) {
+        PyErr_Print();
+        LOG_ERROR("Unable to set argc");
         goto cleanup;
     }
 
     LOG_INFO("Running on %s version %s", info->product, info->version);
     LOG_INFO("Python interpreter initialized and cocotb loaded!");
 
-    // Now that logging has been set up ok we initialize the testbench
-    if (-1 == PyObject_SetAttrString(cocotb_module, "SIM_NAME", PyString_FromString(info->product))) {
+    // Now that logging has been set up ok, we initialize the testbench
+    if (-1 == PyModule_AddStringConstant(cocotb_module, "SIM_NAME", info->product)) {
         PyErr_Print();
         LOG_ERROR("Unable to set SIM_NAME");
         goto cleanup;
     }
 
-    if (-1 == PyObject_SetAttrString(cocotb_module, "SIM_VERSION", PyString_FromString(info->version))) {
-            PyErr_Print();
-            LOG_ERROR("Unable to set SIM_VERSION");
-            goto cleanup;
+    if (-1 == PyModule_AddStringConstant(cocotb_module, "SIM_VERSION", info->version)) {
+        PyErr_Print();
+        LOG_ERROR("Unable to set SIM_VERSION");
+        goto cleanup;
     }
 
     // Set language in use as an attribute to cocotb module, or None if not provided
     const char *lang = getenv("TOPLEVEL_LANG");
-    PyObject* PyLang;
-    if (lang)
-        PyLang = PyString_FromString(lang);
-    else
+    PyObject *PyLang;
+    if (lang) {
+        PyLang = PyString_FromString(lang);                             // New reference
+    } else {
+        Py_INCREF(Py_None);
         PyLang = Py_None;
-
-    if (-1 == PyObject_SetAttrString(cocotb_module, "LANGUAGE", PyLang)) {
-        LOG_ERROR("Unable to set LANGUAGE");
-        goto cleanup;
     }
-
-    // Hold onto a reference to our _fail_test function
-    pEventFn = PyObject_GetAttrString(cocotb_module, "_sim_event");
-
-    if (!PyCallable_Check(pEventFn)) {
+    if (PyLang == NULL) {
         PyErr_Print();
+        LOG_ERROR("Unable to create Python object for cocotb.LANGUAGE");
+        goto cleanup;
+    }
+    if (-1 == PyObject_SetAttrString(cocotb_module, "LANGUAGE", PyLang)) {
+        PyErr_Print();
+        LOG_ERROR("Unable to set LANGUAGE");
+        Py_DECREF(PyLang);
+        goto cleanup;
+    }
+    Py_DECREF(PyLang);
+
+    pEventFn = PyObject_GetAttrString(cocotb_module, "_sim_event");     // New reference
+    if (pEventFn == NULL) {
+        PyErr_Print();
+        LOG_ERROR("Failed to get the _sim_event method");
+        goto cleanup;
+    }
+    if (!PyCallable_Check(pEventFn)) {
         LOG_ERROR("cocotb._sim_event is not callable");
-        goto cleanup;
-    }
-    Py_INCREF(pEventFn);
-
-    cocotb_init = PyObject_GetAttrString(cocotb_module, "_initialise_testbench");         // New reference
-
-    if (cocotb_init == NULL || !PyCallable_Check(cocotb_init)) {
-        if (PyErr_Occurred())
-            PyErr_Print();
-        LOG_ERROR("Cannot find function \"%s\"\n", "_initialise_testbench");
+        Py_DECREF(pEventFn);
+        pEventFn = NULL;
         goto cleanup;
     }
 
-    cocotb_args = PyTuple_New(1);
-    if (dut == NULL)
-        PyTuple_SetItem(cocotb_args, 0, Py_BuildValue(""));        // Note: This function “steals” a reference to o.
-    else
-        PyTuple_SetItem(cocotb_args, 0, PyString_FromString(dut));        // Note: This function “steals” a reference to o.
-    cocotb_retval = PyObject_CallObject(cocotb_init, cocotb_args);
+    cocotb_init = PyObject_GetAttrString(cocotb_module, "_initialise_testbench");   // New reference
+    if (cocotb_init == NULL) {
+        PyErr_Print();
+        LOG_ERROR("Failed to get the _initialise_testbench method");
+        goto cleanup;
+    }
+    if (!PyCallable_Check(cocotb_init)) {
+        LOG_ERROR("cocotb._initialise_testbench is not callable");
+        Py_DECREF(cocotb_init);
+        goto cleanup;
+    }
+
+    PyObject *dut_arg;
+    if (dut == NULL) {
+        Py_INCREF(Py_None);
+        dut_arg = Py_None;
+    } else {
+        dut_arg = PyString_FromString(dut);                             // New reference
+    }
+    if (dut_arg == NULL) {
+        PyErr_Print();
+        LOG_ERROR("Unable to create Python object for dut argument of _initialise_testbench");
+        goto cleanup;
+    }
+
+    cocotb_retval = PyObject_CallFunctionObjArgs(cocotb_init, dut_arg, NULL);
+    Py_DECREF(dut_arg);
+    Py_DECREF(cocotb_init);
 
     if (cocotb_retval != NULL) {
         LOG_DEBUG("_initialise_testbench successful");
         Py_DECREF(cocotb_retval);
     } else {
         PyErr_Print();
-        LOG_ERROR("cocotb initialization failed - exiting\n");
+        LOG_ERROR("cocotb initialization failed - exiting");
         goto cleanup;
     }
 
@@ -370,12 +409,9 @@ int embed_sim_init(gpi_sim_info_t *info)
 cleanup:
     ret = -1;
 ok:
-    if (cocotb_module) {
-        Py_DECREF(cocotb_module);
-    }
-    if (arg_dict) {
-        Py_DECREF(arg_dict);
-    }
+    Py_XDECREF(cocotb_module);
+    Py_XDECREF(simlog_obj);
+
     PyGILState_Release(gstate);
     to_simulator();
 
@@ -392,20 +428,22 @@ void embed_sim_event(gpi_event_t level, const char *msg)
         to_python();
         gstate = PyGILState_Ensure();
 
-        PyObject *fArgs = PyTuple_New(2);
-        PyTuple_SetItem(fArgs, 0, PyInt_FromLong(level));
+        if (msg == NULL) {
+            msg = "No message provided";
+        }
 
-        if (msg != NULL)
-            PyTuple_SetItem(fArgs, 1, PyString_FromString(msg));
-        else
-            PyTuple_SetItem(fArgs, 1, PyString_FromString("No message provided"));
-        PyObject *pValue = PyObject_CallObject(pEventFn, fArgs);
-        if (!pValue) {
+#if PY_MAJOR_VERSION >= 3
+        PyObject* pValue = PyObject_CallFunction(pEventFn, "ls", level, msg);
+#else
+        // Workaround for bpo-9369, fixed in 3.4
+        static char format[] = "ls";
+        PyObject* pValue = PyObject_CallFunction(pEventFn, format, level, msg);
+#endif
+        if (pValue == NULL) {
             PyErr_Print();
             LOG_ERROR("Passing event to upper layer failed");
         }
-
-        Py_DECREF(fArgs);
+        Py_XDECREF(pValue);
         PyGILState_Release(gstate);
         to_simulator();
     }
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -172,6 +172,36 @@ out:
 }
 
 /**
+ * @name    Simulator cleanup
+ * @brief   Called by the simulator on shutdown.
+ * @ingroup python_c_api
+ *
+ * GILState before calling: Not held
+ *
+ * GILState after calling: Not held
+ *
+ * Makes one call to PyGILState_Ensure and one call to Py_Finalize.
+ *
+ * Cleans up reference counts for Python objects and calls Py_Finalize function.
+ */
+void embed_sim_cleanup(void)
+{
+    // If initialization fails, this may be called twice:
+    // Before the initial callback returns and in the final callback.
+    // So we check if Python is still initialized before doing cleanup.
+    if (Py_IsInitialized()) {
+        to_python();
+        PyGILState_Ensure();    // Don't save state as we are calling Py_Finalize
+        Py_DecRef(pEventFn);
+        pEventFn = NULL;
+        clear_log_handler();
+        clear_log_filter();
+        Py_Finalize();
+        to_simulator();
+    }
+}
+
+/**
  * @name    Initialization
  * @brief   Called by the simulator on initialization. Load cocotb Python module
  * @ingroup python_c_api
--- a/cocotb/share/lib/gpi/GpiCommon.cpp
+++ b/cocotb/share/lib/gpi/GpiCommon.cpp
@@ -65,6 +65,17 @@ public:
         return handle_map.size();
     }
 
+    void clear() {
+        std::map<std::string, GpiObjHdl*>::iterator it;
+
+        // Delete the object handles before clearing the map
+        for (it = handle_map.begin(); it != handle_map.end(); it++)
+        {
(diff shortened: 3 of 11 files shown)


cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -243,121 +268,133 @@ int embed_sim_init(gpi_sim_info_t *info)
...
         PyErr_Print();
         LOG_ERROR("Failed to get the _logFromC method");
         goto cleanup;
     }
 
-    if (!PyCallable_Check(simlog_func)) {
+    if (!PyCallable_Check(log_func)) {
         PyErr_Print();
         LOG_ERROR("_logFromC is not callable");
+        Py_DecRef(log_func);
```
eric-wieser: What's the rationale for using the non-macro versions? I've only ever seen the macro versions used  in other codebases.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -243,121 +268,133 @@ int embed_sim_init(gpi_sim_info_t *info)
...
-    set_log_filter(simlog_func);
+    set_log_filter(filter_func);    // Note: This function steals a reference to filter_func.
 
-    argv_list = PyList_New(0);
+    // Build argv for cocotb module
+    argv_list = PyList_New(info->argc);                                     // New reference
     for (i = 0; i < info->argc; i++) {
-        arg_value = PyString_FromString(info->argv[i]);
-        PyList_Append(argv_list, arg_value);
+        PyList_SetItem(argv_list, i, PyString_FromString(info->argv[i]));   // Note: This function steals the reference to created PyString
```
eric-wieser: Strictly you ought to check that PyString_FromString doesn't return NULL here.

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: From the documentation:
"The following functions are for runtime dynamic embedding of Python: Py_IncRef(PyObject *o), Py_DecRef(PyObject *o). They are simply exported function versions of Py_XINCREF() and Py_XDECREF(), respectively."

I'm not strongly attached. I could use the macros and be explicit about Py_DECREF vs Py_XDECREF (implied use under ok label).

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: If it fails for an string in argv, should we put in an empty string, None, or fail init?

eric-wieser: Fail init. If it does fail, you're not going to be able to recover.

garmin-mjames: Added more error checking for argv creation.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -171,6 +171,31 @@ void embed_init_python(void)
...
+ * GILState after calling: Not held
+ *
+ * Makes one call to PyGILState_Ensure and one call to Py_Finalize.
+ *
+ * Cleans up reference counts for Python objects and calls Py_Finalize function.
+ */
+void embed_sim_cleanup(void)
+{
+    to_python();
+    PyGILState_Ensure();    // Don't save state as we are calling Py_Finalize
```
garmin-mjames: Unused local variable was causing errors on linux builds, so I removed it. We shouldn't need it since `Py_Finalize()` should clean up the interpreter.

cocotb/share/lib/embed/gpi_embed.c
eric-wieser: > for runtime dynamic embedding of Python

My understanding is that this is for when you are trying to pull `Py_IncRef` from a dll, and do not have any header files available. Given that's not our use case, I think we should use the macro versions.

On this particular line, you can use `Py_DECREF`, since you are sure that it is not null.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -243,124 +268,154 @@ int embed_sim_init(gpi_sim_info_t *info)
...
     // Set language in use as an attribute to cocotb module, or None if not provided
     const char *lang = getenv("TOPLEVEL_LANG");
     PyObject* PyLang;
     if (lang)
-        PyLang = PyString_FromString(lang);
-    else
-        PyLang = Py_None;
+        PyLang = PyString_FromString(lang);         // New reference
+    else {
+        PyLang = Py_BuildValue("");                 // New reference to None object
```
eric-wieser: This is a very strange spelling - I'd be inclined to use the faster
```C
Py_INCREF(Py_None);
PyLang = Py_None;
```

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -116,16 +129,22 @@ void gpi_embed_init(gpi_sim_info_t *info)
...
 }
 
 void gpi_sim_end()
 {
     registered_impls[0]->sim_end();
 }
 
+void gpi_cleanup(void)
+{
+    CLEAR_STORE();
```
eric-wieser: What's the point in this?

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -183,6 +186,12 @@ int handle_gpi_callback(void *user_data)
 
 err:
     to_simulator();
+
+    if (sim_ending)
+        // This is the last callback of a successful run,
+        // so call the cleanup function as we'll never return
+        // to Python
+        gpi_cleanup();
```
eric-wieser: nit: braces please

cocotb/share/lib/gpi/GpiCommon.cpp
garmin-mjames: If SINGLETON_HANDLES is defined, we store sim handles and re-use them.
This will delete those handle objects, and in the case of VHPI, release the handles back to the simulator (see VhpiCbHdl.cpp changes).

eric-wieser: Not paying attention, sorry - feel free to resolve this comment chain.

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: Changed to macro versions.
I used Py_XDECREF in a few places where appropriate.

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: Added error checking in more places.

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: Changed.

cocotb/share/lib/simulator/simulatormodule.c
garmin-mjames: Added braces.

garmin-mjames:
I was seeing unhandled exception errors in `test_exit_error` as @ktbarrett mentioned in #1056 in Aldec, and the Travis build logs showed iverilog segfaulting.

I tracked it down to embed_sim_cleanup() getting called a second time during the final callback, so I added a check with `Py_IsInitialized()`.

cocotb/share/lib/simulator/simulatormodule.c
garmin-mjames: Added.

cocotb/scheduler.py
```diff
@@ -775,9 +775,13 @@ def finish_scheduler(self, exc):
         """Directly call into the regression manager and end test
            once we return the sim will close us so no cleanup is needed.
         """
-        self.log.debug("Issue sim closedown result to regression object")
-        self._test.abort(exc)
-        cocotb.regression_manager.handle_result(self._test)
+        # If there is an error during cocotb initialization, self._test may not
+        # have been set yet. Don't cause another Python exception here.
```
eric-wieser: Might be worth adding a `self.log.warn` if this gets reached without `self._test` being set

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -228,139 +258,202 @@ int embed_sim_init(gpi_sim_info_t *info)
...
         PyErr_Print();
         LOG_ERROR("Failed to get the _logFromC method");
         goto cleanup;
     }
-
-    if (!PyCallable_Check(simlog_func)) {
-        PyErr_Print();
+    if (!PyCallable_Check(pLogFn)) {
+        if (PyErr_Occurred())
+            PyErr_Print();
```
eric-wieser: This is unreachable - from the docs,

> This function always succeeds.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -228,139 +258,202 @@ int embed_sim_init(gpi_sim_info_t *info)
...
         PyErr_Print();
         LOG_ERROR("Failed to get the isEnabledFor method");
         goto cleanup;
     }
-
-    if (!PyCallable_Check(simlog_func)) {
-        PyErr_Print();
+    if (!PyCallable_Check(pFilterFn)) {
+        if (PyErr_Occurred())
+            PyErr_Print();
```
eric-wieser: As above

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -228,139 +258,202 @@ int embed_sim_init(gpi_sim_info_t *info)
...
+    pArgvList = PyList_New(info->argc);                                 // New reference
+    if (pArgvList == NULL) {
+        PyErr_Print();
+        LOG_ERROR("Unable to create argv list");
+        goto cleanup;
+    }
     for (i = 0; i < info->argc; i++) {
-        arg_value = PyString_FromString(info->argv[i]);
-        PyList_Append(argv_list, arg_value);
+        pArgvItem = PyString_FromString(info->argv[i]);                 // New reference
```
eric-wieser:
```suggestion
        PyObject *pArgvItem = PyString_FromString(info->argv[i]);                 // New reference
```
and remove the definition above

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -228,139 +258,202 @@ int embed_sim_init(gpi_sim_info_t *info)
...
-    if (!PyCallable_Check(pEventFn)) {
+    pEventFn = PyObject_GetAttrString(pCocotbModule, "_sim_event");                 // New reference
+    if (pEventFn == NULL) {
         PyErr_Print();
+        LOG_ERROR("Failed to get the _sim_event method");
+        goto cleanup;
+    }
+    if (!PyCallable_Check(pEventFn)) {
+        if (PyErr_Occurred())
+            PyErr_Print();
```
eric-wieser: Unreachable

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -228,139 +258,202 @@ int embed_sim_init(gpi_sim_info_t *info)
...
-    if (cocotb_init == NULL || !PyCallable_Check(cocotb_init)) {
+    pCocotbInit = PyObject_GetAttrString(pCocotbModule, "_initialise_testbench");   // New reference
+    if (pCocotbInit == NULL) {
+        PyErr_Print();
+        LOG_ERROR("Failed to get the _initialise_testbench method");
+        goto cleanup;
+    }
+    if (!PyCallable_Check(pCocotbInit)) {
         if (PyErr_Occurred())
             PyErr_Print();
```
eric-wieser: Unreachable

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -32,20 +32,30 @@
...
 void set_log_filter(void *filter)
 {
-    pLogFilter = (PyObject *)filter;
-    Py_INCREF(pLogFilter);
+    pLogFilter = (PyObject *)filter;        // Note: This function steals a reference to filter.
+}
+
+void clear_log_filter(void)
+{
+    Py_DecRef(pLogFilter);
```
eric-wieser: nit: `Py_XDECREF`

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -149,50 +159,82 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
...
-        PyGILState_Release(gstate);
-        return;
-    }
-    int is_enabled = PyObject_IsTrue(filter_ret);
-    Py_DECREF(filter_ret);
+    PyObject *pLevel = PyLong_FromLong(level);              // New reference
+    if (pLevel == NULL)
+        goto error;
+    if (0 != PyTuple_SetItem(pCheckArgs, 0, pLevel))        // Note: This function steals a reference to pLevel
+        goto error;
```
eric-wieser: This is overkill - you could use `PyTuple_SETITEM` here instead, which always succeeds (and below)

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: Ok, then the `PyErr_Print()` should also be removed, right? From the docs:

> void PyErr_PrintEx(int set_sys_last_vars)
> ...
>    Call this function only when the error indicator is set. Otherwise it will cause a fatal error!

cocotb/share/lib/gpi_log/gpi_logging.c
garmin-mjames: Should we check the `_From` function returns or assume everything will work as before?

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -149,50 +159,82 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
...
 
-    PyObject *call_args = PyTuple_New(5);
-    PyTuple_SetItem(call_args, 0, PyLong_FromLong(level));           // Note: This function steals a reference.
-    PyTuple_SetItem(call_args, 1, PyUnicode_FromString(pathname));   // Note: This function steals a reference.
-    PyTuple_SetItem(call_args, 2, PyLong_FromLong(lineno));          // Note: This function steals a reference.
-    PyTuple_SetItem(call_args, 3, PyUnicode_FromString(log_buff));   // Note: This function steals a reference.
-    PyTuple_SetItem(call_args, 4, PyUnicode_FromString(funcname));
-
-    PyObject *handler_ret = PyObject_CallObject(pLogHandler, call_args);
-    Py_DECREF(call_args);
```
eric-wieser: If you're going to change them, you can simplify these lines to
```C
PyObject *handler_ret = PyObject_CallFunction(pLogHandler, "lslss", level, pathname, lineno, log_buff, funcname);
```
which handles all the ref-counting for you.

cocotb/share/lib/gpi_log/gpi_logging.c
eric-wieser: The `_From` functions _can_ fail - they can throw `MemoryError` - so your change is an improvement there, albeit an irrelevant one. See my comment below though - using `PyObject_CallFunction` or `PyBuildValue` makes this type of tuple construction far less painful.

cocotb/share/lib/embed/gpi_embed.c
eric-wieser: Yes - my comment refers to these two lines together.

cocotb/share/lib/gpi_log/gpi_logging.c
eric-wieser: Crap, that doesn't mean the right thing on python 2 :(. I think what you have is the only option

garmin-mjames: Are you sure? `PyObject_CallFunction()` looks like it fits this case in the Python 2.7 docs.

cocotb/scheduler.py
garmin-mjames: The only place `finish_scheduler()` is called is from `_sim_event()`, which logs an error just before calling.

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: Removed all occurences. Resolving other comments.

cocotb/share/lib/gpi_log/gpi_logging.c
garmin-mjames: Missed this one. Changed.

cocotb/share/lib/gpi_log/gpi_logging.c
garmin-mjames: It works locally. The 2.7 build on linux is failing because of a const warning elevated to error.
Passing in a string literal to the `PyObject_CallFuction()` format argument causes a const mismatch...
Python 3.4+ changed to a const char * so it builds fine.

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: Changed.

cocotb/share/lib/gpi_log/gpi_logging.c
garmin-mjames: Added `#ifdef`s for Python 2/3 to get it building without warnings.

garmin-mjames: Now it has `gpi_logging.c:166:65: error: cast discards 'const' qualifier from pointer target type [-Werror=cast-qual]`
I thought this might happen. Is there an easy way to make gcc happy here for 2.7?

eric-wieser: Aide from compilation, the problem is that "s" will construct a `str` object in Python 2, but the current code constructs `unicode` objects.

eric-wieser: Using `const_cast<char *>("...")` may solve the warning.

eric-wieser: Perhaps worth abandoning this rabbit home for now.

ktbarrett:
```python
char fmt[] = "lslss";
PyObject *handler_ret = PyObject_CallFunction(pLogHandler, fmt, level, pathname, lineno, log_buff, funcname);
```
Works.

garmin-mjames: Ok, I split the difference. Still creating `unicode` objects but calling with `PyObject_CallFunctionObjArgs()` so we don't have to build the tuple first.

I went with @ktbarrett's method, but another option would be adding `-Wno-discarded-qualifiers` to the Makefile.

eric-wieser: I'd thought this was a C++ file. Evidently not!

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -146,53 +156,65 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
...
+
+    PyObject *pFunctionName = PyUnicode_FromString(funcname);   // New reference
+    if (pFunctionName == NULL)
+        goto error;
+
+    Py_INCREF(pLevel);
+
+    // Log function args are level, filename, lineno, msg, function
+    PyObject *pLogReturn = PyObject_CallFunctionObjArgs(pLogHandler, pLevel, pFileName, pLineNo, pMessage, pFunctionName, NULL);
+    Py_DECREF(pLevel);
```
eric-wieser: This line seems a little pointless, since you decref all the variables down below anyway.

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -146,53 +156,65 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
...
+
+    PyObject *pMessage = PyUnicode_FromString(log_buff);        // New reference
+    if (pMessage == NULL)
+        goto error;
+
+    PyObject *pFunctionName = PyUnicode_FromString(funcname);   // New reference
+    if (pFunctionName == NULL)
+        goto error;
+
+    Py_INCREF(pLevel);
```
eric-wieser: Similarly, this is not needed

cocotb/share/lib/gpi_log/gpi_logging.c
garmin-mjames: You're correct. I was still thinking of how I did it before, where the tuple argument to pLogFilter stole the reference. Removed the extra INCREF/DECREF pair.

garmin-mjames:
Fixed gpi_log error handling by properly initializing `PyObject` pointers to `NULL`.

cocotb/share/lib/gpi_log/gpi_logging.c
ktbarrett: `const_cast` is never a good idea. I have no idea what Python does with the builder string, but if it tries to modify it in a function and you give it a `const_cast` string literal (string literals are stored in read-only section of the object), you'll end up with a crash. Likely, Python isn't modifying it and just forgot the qualifier, hooray.

ktbarrett: @eric-wieser Is there any reason why all of the sources aren't C++ files?

eric-wieser: @ktbarrett:

> Likely, Python isn't modifying it and just forgot the qualifier, hooray.

If you can prove this to be the case by reading the source code, then `const_cast` should be perfectly fine. I believe this is exactly what happened - it comes from a time when compilers were less strict about pointer types.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -228,139 +258,177 @@ int embed_sim_init(gpi_sim_info_t *info)
...
     }
 
-    cocotb_args = PyTuple_New(1);
-    if (dut == NULL)
-        PyTuple_SetItem(cocotb_args, 0, Py_BuildValue(""));        // Note: This function “steals” a reference to o.
-    else
-        PyTuple_SetItem(cocotb_args, 0, PyString_FromString(dut));        // Note: This function “steals” a reference to o.
-    cocotb_retval = PyObject_CallObject(cocotb_init, cocotb_args);
+    pInitReturn = PyObject_CallFunctionObjArgs(pCocotbInit, pDutObject, NULL);
+    Py_DECREF(pDutObject);
```
eric-wieser: Missing `Py_DECREF(pCocotbInit)` I think

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -228,139 +258,177 @@ int embed_sim_init(gpi_sim_info_t *info)
         }
     }
 
-    PyObject *cocotb_module, *cocotb_init, *cocotb_args, *cocotb_retval;
-    PyObject *simlog_obj, *simlog_func;
-    PyObject *argv_list, *argc, *arg_dict, *arg_value;
+    PyObject *pCocotbModule, *pCocotbInit, *pInitReturn;
+    PyObject *pSimLogObject, *pLogFn, *pFilterFn;
+    PyObject *pArgvList;
```
eric-wieser: This `pCamelCase` naming convention from my experience is not all that common in cocotb vs what was here before. Are you use `p` to mean `pointer` or `python`? Note that you would never use a `PyObject` by itself, so there's little value in marking it as a pointer.

I'd be inclined to keep the diff small, and leave these names in the same style as they were before

garmin-mjames: That's fine. I was trying to bring it in-line with the naming scheme of other variables, such as pEventFn, pLogHandler, pLogFilter.
It's probably better to do that on a different commit/PR specifically for that.

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: Added. Thanks for the thorough review of these changes!

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -243,124 +273,163 @@ int embed_sim_init(gpi_sim_info_t *info)
...
+    argv_list = PyList_New(info->argc);                                 // New reference
+    if (argv_list == NULL) {
+        PyErr_Print();
+        LOG_ERROR("Unable to create argv list");
+        goto cleanup;
+    }
     for (i = 0; i < info->argc; i++) {
-        arg_value = PyString_FromString(info->argv[i]);
-        PyList_Append(argv_list, arg_value);
+        PyObject *pArgvItem = PyString_FromString(info->argv[i]);       // New reference
```
eric-wieser: nit: `arg_item` would be mode consistent with the other locals

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -392,20 +458,17 @@ void embed_sim_event(gpi_event_t level, const char *msg)
...
-            PyTuple_SetItem(fArgs, 1, PyString_FromString(msg));
-        else
-            PyTuple_SetItem(fArgs, 1, PyString_FromString("No message provided"));
-        PyObject *pValue = PyObject_CallObject(pEventFn, fArgs);
-        if (!pValue) {
+#if PY_MAJOR_VERSION >= 3
+        PyObject* pValue = PyObject_CallFunction(pEventFn, "ls", level, (msg != NULL) ? msg : "No message provided");
+#else
+        static char format[] = "ls";
+        PyObject* pValue = PyObject_CallFunction(pEventFn, format, level, (msg != NULL) ? msg : "No message provided");
```
eric-wieser: Probably worth a comment here, maybe just `// workaround for bpo-9369, fixed in 3.4`

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -392,20 +458,17 @@ void embed_sim_event(gpi_event_t level, const char *msg)
...
-        PyTuple_SetItem(fArgs, 0, PyInt_FromLong(level));
-
-        if (msg != NULL)
-            PyTuple_SetItem(fArgs, 1, PyString_FromString(msg));
-        else
-            PyTuple_SetItem(fArgs, 1, PyString_FromString("No message provided"));
-        PyObject *pValue = PyObject_CallObject(pEventFn, fArgs);
-        if (!pValue) {
+#if PY_MAJOR_VERSION >= 3
+        PyObject* pValue = PyObject_CallFunction(pEventFn, "ls", level, (msg != NULL) ? msg : "No message provided");
```
eric-wieser: nit: could eliminate these ternaries with
```
if (msg == NULL) {
    msg = "No message provided";
}
```
above

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -145,54 +155,65 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
...
         return;
 
-    // Ignore truncation
-    // calling args is level, filename, lineno, msg, function
-    //
     PyGILState_STATE gstate = PyGILState_Ensure();
 
-    PyObject *check_args = PyTuple_New(1);
-    PyTuple_SetItem(check_args, 0, PyLong_FromLong(level));
+    PyObject *filename_arg = NULL, *lineno_arg = NULL, *msg_arg = NULL, *function_arg = NULL;
```
eric-wieser: Probably worth a comment like `declared here in order to be initialized before any goto statements`

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -145,54 +155,65 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
...
-    PyObject *handler_ret = PyObject_CallObject(pLogHandler, call_args);
-    Py_DECREF(call_args);
-    if (handler_ret == NULL){
-        PyErr_Print();
-        PyGILState_Release(gstate);
-        return;
-    }
+    filename_arg = PyUnicode_FromString(pathname);      // New reference
+    if (filename_arg == NULL)
+        goto error;
```
eric-wieser: nit: up to @imphil. but braces are probably a good idea here.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -43,6 +57,10 @@ VhpiSignalObjHdl::~VhpiSignalObjHdl()
 
     if (m_binvalue.value.str)
         free(m_binvalue.value.str);
+
+    LOG_DEBUG("Releasing VhpiSignalObjHdl handle at 0x%X\n", get_handle<vhpiHandleT>());
```
eric-wieser: Does this do printf formatting? If so, you need `zX`.

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: Changed to `argv_item`.

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: Added comment.

cocotb/share/lib/embed/gpi_embed.c
garmin-mjames: Changed.

cocotb/share/lib/gpi_log/gpi_logging.c
garmin-mjames: Added comment.

cocotb/share/lib/gpi_log/gpi_logging.c
garmin-mjames: Added braces to the if statements.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
garmin-mjames: Changed to `%p` and cast handle to `(void *)`.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -30,6 +30,20 @@
 
 extern "C" void handle_vhpi_callback(const vhpiCbDataT *cb_data);
 
+VhpiArrayObjHdl::~VhpiArrayObjHdl()
+{
+    LOG_DEBUG("Releasing VhpiArrayObjHdl handle at %p\n", (void *)get_handle<vhpiHandleT>());
```
eric-wieser: nit: I think `vhpiHandleT` might already be `void*`. If it's not, then `static_cast<void*>(...)` would be the C++ spelling.

Fine as is though - this is only debug code anyway, and you fixed the bug already.

garmin-mjames: From `vhpi_user.h`:

`typedef uint32_t *vhpiHandleT;`

imphil:
Thanks for this change @garmin-mjames, and thanks for the review @eric-wieser!

PR #1166, opened 2019-10-24
Windows: Print error messages in DLL functions.

garmin-mjames:
When `utils_dyn_open()` or `utils_dyn_sym()` fail, get the system error message and print it.

### DIFF
--- a/cocotb/share/lib/utils/cocotb_utils.c
+++ b/cocotb/share/lib/utils/cocotb_utils.c
@@ -66,7 +66,17 @@ void* utils_dyn_open(const char* lib_name)
     SetErrorMode(0);
     ret = LoadLibrary(lib_name);
     if (!ret) {
-        printf("Unable to open lib %s\n", lib_name);
+        printf("Unable to open lib %s", lib_name);
+        LPSTR msg_ptr;
+        if (FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM |
+                           FORMAT_MESSAGE_ALLOCATE_BUFFER, NULL,
+                           GetLastError(),
+                           MAKELANGID(LANG_NEUTRAL, SUBLANG_SYS_DEFAULT),
+                           (LPSTR)&msg_ptr, 255, NULL)) {
+            printf(": %s", msg_ptr);
+            LocalFree(msg_ptr);
+        } else
+            printf("\n");
     }
 #else
     /* Clear status */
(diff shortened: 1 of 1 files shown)


cocotb/share/lib/utils/cocotb_utils.c
```diff
@@ -66,7 +66,19 @@ void* utils_dyn_open(const char* lib_name)
...
+        FormatMessage
+        (
+            FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_ALLOCATE_BUFFER,
+            NULL,
+            GetLastError(),
+            MAKELANGID(LANG_NEUTRAL, SUBLANG_SYS_DEFAULT),
+            (LPTSTR)&msg_ptr,
+            255,
+            NULL
+        );
```
imphil: Please format this function call more like our existing function calls, so 

```
FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM | 
              FORMAT_MESSAGE_ALLOCATE_BUFFER, NULL, 
              GetLastError(), ...);
```

garmin-mjames: I condensed the formatting and added check of return value. Also provided more detail in git commit message.

cocotb/share/lib/utils/cocotb_utils.c
```diff
@@ -66,7 +66,17 @@ void* utils_dyn_open(const char* lib_name)
...
     ret = LoadLibrary(lib_name);
     if (!ret) {
-        printf("Unable to open lib %s\n", lib_name);
+        printf("Unable to open lib %s", lib_name);
+        LPSTR msg_ptr;
+        if (FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM |
+                           FORMAT_MESSAGE_ALLOCATE_BUFFER, NULL,
+                           GetLastError(),
+                           MAKELANGID(LANG_NEUTRAL, SUBLANG_SYS_DEFAULT),
+                           (LPSTR)&msg_ptr, 255, NULL)) {
```
eric-wieser: This looks very wrong to me - casting a `T*` to a `T` is nonsense.

eric-wieser: Huh, guess the windows API is just bizarre: https://stackoverflow.com/a/24214707/102441

garmin-mjames: I took it directly from the docs: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-formatmessagea

imphil:
Thanks! I added missing curly braces around the else block and pushed it.

PR #1167, opened 2019-10-24
Improve debug output, fix typos.

cmarqu:


cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -146,13 +146,13 @@ static void gpi_load_libs(std::vector<std::string> to_load)
...
         if (!lib_handle) {
-            printf("Error loading lib %s\n", now_loading);
+            LOG_ERROR("Error loading shared library %s\n", now_loading);
             exit(1);
         }
         std::string sym = (*iter) + "_entry_point";
         void *entry_point = utils_dyn_sym(lib_handle, sym.c_str());
         if (!entry_point) {
-            printf("Unable to find entry point for %s\n", now_loading);
+            LOG_ERROR("Unable to find entry point %s for shared library %s\n", sym.c_str(), now_loading);
```
imphil: The LOG_* functions come from libgpilog, which are (as far as I can see) not safe to use in the library loading routine. Please revert back to printf().

cmarqu: Okay. While I **have** seen the message printed with VCS, it may have been by luck then.

cmarqu: Changed back. I prefixed the string with "cocotb: " since it wasn't clear where the message came from when I last saw it.

Issue #1103, opened 2019-09-08
Support VHDL with VCS


cmarqu:
Right now, we are using a two-step setup for Synopsys VCS which according to http://www.vlsiip.com/vcs/ is limited to Verilog only. VCS(-MX) is capable of simulating VHDL too with a three-step process using ``vhdlan``.

cmarqu:
I'll work on this soon.

cmarqu:
I can successfully compile and elaborate ``test_iteration_vhdl``with https://github.com/cmarqu/cocotb/tree/support-vhdl-in-vcs and also bring up the simulator, but it turns out that the ``vhpi_user.h`` that comes with Synopsys VCS-MX is very different from what is in the appendix of the VHDL standard - enums instead of single constants, and struct elements of different types.
(https://github.com/jamiekang/embedded-sw-dev-tools/blob/master/binsim/vhpi_user.h seems to be an older version of it.)

hashhsah:
@cmarqu below is an incomplete attempt at mixed verilog/vhdl simulation, with verilog as top-level module. Simulation runs fine, but I didn't try to access anything within the module written in VHDL.

```
ifeq ($(ARCH),x86_64)
EXTRA_ARGS += -full64
endif
VLOGAN_ARGS += -sverilog +define+COCOTB_SIM=1

$(SIM_BUILD)/pli.tab :
	[ -d $(SIM_BUILD) ] || mkdir $(SIM_BUILD)
	echo "acc+=rw,wn:*" > $@

$(SIM_BUILD)/flist_vlog.f : $(VERILOG_SOURCES)
	[ -d $(SIM_BUILD) ] || mkdir $(SIM_BUILD)
	rm -f $@ ; touch  $@
	for f in $(VERILOG_SOURCES); do echo "$$f" >> $@; done

$(SIM_BUILD)/flist_vhdl.f : $(VHDL_SOURCES)
	[ -d $(SIM_BUILD) ] || mkdir $(SIM_BUILD)
	rm -f $@ ; touch  $@
	for f in $(VHDL_SOURCES); do echo "$$f" >> $@; done

# Compilation phase
$(SIM_BUILD)/_vmake.vlogan: $(SIM_BUILD)/flist_vlog.f
	cd $(SIM_BUILD) && \
    source $(SNPS_ENV) && \
	if [ -s flist_vlog.f ]; then \
[...]

Issue #1062, opened 2019-08-08
Integrate cocotb-test into cocotb


themperek:
I would like to know if there is consensus to merge [cocotb-tests](https://github.com/themperek/cocotb-test) in current form into cocotb?

In case you would like to see this in should I prepare one big PR or split somehow into smaller pieces?

``cocotb-test`` provides standard python unit testing capabilities 
- compiles the libraries during setup/installation phase using setuptools (separate directory for every simulator) 
- uses python to start simulators (remove the need for Makefiles), includes Makefile compatibility mode
- provides simple pytest plugin to merge results
- allow the look and feel of Python unit testing
- allow easy customization of simulation flow
- easy installation (especially on Windows)
- can create wheel package (no need to compile libraries by user)
- allow to use [pytest-xdist](https://pypi.org/project/pytest-xdist/) / [pytest-parallel](https://github.com/browsertron/pytest-parallel)

Somehow releted: #549 #437 #428 #354 #1057 #991 #960 #807 #482 #595 #711 #696 #667 #508 #528 #498 #488 ...

Talk from WOSDV 2019: [cocotb-test_WOSDV2019.pdf](https://github.com/cocotb/cocotb/files/3908378/cocotb-test_WOSDV2019.pdf)

TC01:
I haven't gotten the chance to try cocotb-test yet, but one question:

> compiles the libraries during setup/installation phase using setuptools (separate directory for every simulator)

Does this mean that I can't install the libraries for a simulator I don't have installed on my system?

I think it might be nice to still support the current way of doing things, where the libraries are built at run-time rather than at installation-time. Maybe this falls under Makefile compatibility mode?

I was mainly thinking here about possibly making packages for Linux distributions (for instance, I've been planning to make a package for Fedora/RHEL now that cocotb 1.2 is out). A distribution won't be able to ship libraries compiled against nonfree simulators, but it could ship the source code for all the simulator libraries and let the user build each one on demand (as things work at the moment).

Or I guess as another example... if the libraries are built at installation time, what happens if I need to upgrade the simulator version? (Or if there are two versions of a simulator installed side-by-side on a shared machine and users are able to switch between them). Do I need to reinstall cocotb to recompile the libraries?

I agree that having the option to install prebuilt libraries is nice though. So it might be nice to support both, if possible...

themperek:
[...]

PR #1168, opened 2019-10-24
Rename libvhpi to libcocotbvhpi.

cmarqu:
This is necessary to prevent a naming collision with a ``libvhpi.so`` that is shipped with Synopsys VCS-MX.
Related to #1103.

### DIFF
--- a/cocotb/share/lib/Makefile
+++ b/cocotb/share/lib/Makefile
@@ -48,7 +48,7 @@ $(LIB_DIR)/libcocotb.$(LIB_EXT): $(COCOTB_SHARE_DIR)/lib/embed/gpi_embed.c | $(L
 $(LIB_DIR)/libvpi.$(LIB_EXT): $(COCOTB_SHARE_DIR)/lib/vpi/VpiImpl.cpp $(COCOTB_SHARE_DIR)/lib/vpi/VpiCbHdl.cpp | $(LIB_DIR)
 	make -C $(COCOTB_SHARE_DIR)/lib/vpi EXTRA_LIBS=$(EXTRA_LIBS) EXTRA_LIBDIRS=$(EXTRA_LIBDIRS) SIM=$(SIM)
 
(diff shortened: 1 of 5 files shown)


ktbarrett:
Why not `libcocotbvcs`? Since we have to compile the GPI using simulator-specific defines, `libcocotbvhpi` is misleading, as it might not work with another simulator with VHPI support. This is probably universal to the other objects as well.

cmarqu:
Good point. I'll open a ticket to solicit comments about the extent of changes in that area.

imphil:
@cmarqu The change itself LGTM, but please amend the commit message to describe why this change was done. If we have a look at the git history in the future we want to know what motivated a change.

cmarqu:
@imphil Oh, I would not have expected that response after the good comment @ktbarrett made. I know Chris and Stu always tried to have no tool-specifics in the code, but it looks like we already crossed that line without a way to go back - with the necessary changes to support VCS' VHPI implementation even less. What do you think? More discussion needed, or do we do it step by step with this PR as the first one?

ktbarrett:
Tool specific code is only an issue if it affects user code or user-facing code. Simulator-specific workarounds at the simulator interface level are unavoidable, but as long as the GPI abstracts them away, there is no problem. `libvhpi.so` and all the other simulator-to-GPI abstraction objects should be associativity named to the simulator they support. Then we do not have to rebuild an object every time we change simulators that use the same interface.

Personally, I'd prefer `libgpi.{SIM}.so` seeing as these objects are a specific implementation of the GPI.

themperek:
When we are at this. Following https://github.com/cocotb/cocotb/issues/1062 I was proposing to compile everything at setup using setuptools (separate for every simulator). I can prepare this for cocotb it there is a will to move in this direction. It is implemented here: https://github.com/themperek/cocotb-test/blob/master/cocotb_test/build_libs.py Can change naming conversion.

imphil:
- This change got the green light as it is a minimal solution to a real problem, there's no pressing need to go further at this point.
- The discussion of "one library per simulator" vs "one library per interface" is much beyond the scope of this PR and should be discussed on the mailing list or a separate issue, not here.

cmarqu:
Commit message is amended to state the reason for this change, and #1181 is opened for further discussion.

PR #1171, opened 2019-10-28
Fix race condition when external threads make function calls

dmiller-jumptrading:
I observed frequent nondeterminism in some tests when using an external function that makes callbacks (a la https://github.com/cocotb/cocotb/blob/master/tests/test_cases/test_external/test_external.py#L139-L144). What's going on is that the main thread sits in `thread_wait` until `thread_suspend` is called, but once it is, the main thread can immediately continue the test and read/modify `_pending_coros`, which causes a race with the external thread making the call. This fix removed nondeterminism in my test across several trials.

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -580,19 +580,29 @@ class Scheduler(object):
 
         @cocotb.coroutine
         def wrapper():
+            # This function runs in the scheduler thread
             try:
                 _outcome = outcomes.Value((yield coro))
             except BaseException as e:
                 _outcome = outcomes.Error(e)
             event.outcome = _outcome
+            # Notify the current (scheduler) thread that we are about to wake
+            # up the background (`@external`) thread, making sure to do so
+            # before the background thread gets a chance to go back to sleep by
+            # calling thread_suspend.
+            # We need to do this here in the scheduler thread so that no more
+            # coroutines run until the background thread goes back to sleep.
+            t.thread_resume()
             event.set()
 
         event = threading.Event()
-        t.thread_suspend()
         self._pending_coros.append(wrapper())
-        # This blocks the calling external thread until the coroutine finishes
+        # The scheduler thread blocks in `thread_wait`, and is woken when we
+        # call `thread_suspend` - so we need to make sure the coroutine is
+        # queued before that.
+        t.thread_suspend()
+        # This blocks the calling `@external` thread until the coroutine finishes
         event.wait()
-        t.thread_resume()
         return event.outcome.get()
 
     def run_in_executor(self, func, *args, **kwargs):
@@ -752,20 +762,20 @@ class Scheduler(object):
         # chaining
         if coro_completed:
             self.unschedule(coroutine)
-            return
 
         # Don't handle the result if we're shutting down
         if self._terminate:
             return
 
-        try:
-            result = self._trigger_from_any(result)
-        except TypeError as exc:
-            # restart this coroutine with an exception object telling it that
-            # it wasn't allowed to yield that
-            result = NullTrigger(outcome=outcomes.Error(exc))
+        if not coro_completed:
+            try:
+                result = self._trigger_from_any(result)
+            except TypeError as exc:
+                # restart this coroutine with an exception object telling it that
+                # it wasn't allowed to yield that
+                result = NullTrigger(outcome=outcomes.Error(exc))
 
-        self._coroutine_yielded(coroutine, result)
+            self._coroutine_yielded(coroutine, result)
 
         # We do not return from here until pending threads have completed, but only
         # from the main thread, this seems like it could be problematic in cases
--- a/tests/test_cases/test_external/test_external.py
+++ b/tests/test_cases/test_external/test_external.py
@@ -104,6 +104,29 @@ def test_time_in_external(dut):
     if time != time_now:
         raise TestFailure("Time has elapsed over external call")
 
+@cocotb.function
+def wait_cycles(dut, n):
+    for _ in range(n):
+        yield RisingEdge(dut.clk)
+
+def wait_cycles_wrapper(dut, n):
+    return wait_cycles(dut, n)
+
+@cocotb.test()
+def test_time_in_external_yield(dut):
+    """Test that an external function calling back into a cocotb function
+    takes the expected amount of time"""
+    clk_gen = cocotb.fork(Clock(dut.clk, 100).start())
+    yield Timer(10, 'ns')
+    for n in range(5):
+        for i in range(20):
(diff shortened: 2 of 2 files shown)


cocotb/scheduler.py
```diff
@@ -578,8 +578,8 @@ def queue_function(self, coroutine):
         # each entry always has a unique thread.
         t, = matching_threads
 
-        t.thread_suspend()
         self._pending_coros.append(coroutine)
+        t.thread_suspend()
```
eric-wieser: Doesn't this now have the opposite problem where if `coroutine` ends too quickly, `t` is resumed before it is suspended?

dmiller-jumptrading: The coroutine is run in the main thread, which is stuck [here](https://github.com/cocotb/cocotb/blob/master/cocotb/scheduler.py#L757-L760) until `thread_suspend` is called. The function wrapper [waits for the coroutine to have run](https://github.com/cocotb/cocotb/blob/master/cocotb/decorators.py#L350) before calling `thread_resume`, so resume can't happen before suspend.

eric-wieser: That argument seems reasonable. Can you add a comment like
```python
# scheduler thread is blocked here and resumes when thread_suspend is called, so make sure everything
# is queued first
```

eric-wieser: Also, a test for the case above would be handy

dmiller-jumptrading:
The test I wrote uncovered a couple other issues around the returning of externally called functions, which I've fixed in the updated commit:
 - When a function returns, the calling thread may continue nondeterministically ordered with other operations. I fixed this by adding another thread event so that the coroutine cannot complete until the calling thread has called `thread_resume`.
 - Sometimes, when a function returns, sim time may elapse before the calling thread continues. Although this is not an issue with determinism, it does seem to be a correctness issue, so I now have `schedule` check for pending threads even when `coro_completed` is true.

eric-wieser:
With these changes, can `test_ext_exit_error` have the `expect_fail` removed?

cocotb/decorators.py
```diff
@@ -336,20 +336,23 @@ def log(self):
...
 
         event = threading.Event()
-        waiter = cocotb.scheduler.queue_function(execute_function(self, event))
+        back_event = threading.Event()
+        waiter = cocotb.scheduler.queue_function(execute_function(self, event, back_event))
         # This blocks the calling external thread until the coroutine finishes
         event.wait()
         waiter.thread_resume()
+        back_event.set()
         return event.outcome.get()
```
eric-wieser: A comment explaining the purpose of `back_event` would be good here. It seem the goal is to ensure that `thread_resume` is called before the `coroutine` completes. Can you explain _why_ that matters in the comment?

dmiller-jumptrading:
Added comment. It looks like `test_ext_exit_error` still fails though, I think that that is a separate issue.

eric-wieser:
> I think that that is a separate issue.

I had my hopes... Oh well. Let's not try and expand the scope of this any further than it already has

eric-wieser:
I did some refactoring in gh-1173 of this code to try and understand what's going on - synchronizing multiple types of event across multiple files was making this hard to understand. I think the rationale here would become clearer after rebasing on that, and I'm happy to take ownership of that rebase.

dmiller-jumptrading:
That refactoring does make things clearer. I can do the rebase in this PR, or you can take it over--whichever is easier for you.

eric-wieser:
Alright, ready for the rebase - probably best if you make it, but I'll pick it up if you don't make it in the next week or so.

dmiller-jumptrading:
Rebased, with minor adjustments

dmiller-jumptrading:
The CI failure on Windows looks spurious...
```
Step 8/11 : RUN conda install --yes -c msys2 m2-base m2-make m2w64-toolchain libpython
 ---> Running in b731de74080f
Collecting package metadata (current_repodata.json): ...working... failed
CondaHTTPError: HTTP 504 GATEWAY_TIMEOUT for url <https://conda.anaconda.org/msys2/win-64/current_repodata.json>
```

cocotb/scheduler.py
```diff
@@ -585,14 +585,19 @@ def wrapper():
             except BaseException as e:
                 _outcome = outcomes.Error(e)
             event.outcome = _outcome
+            # We need to make sure that thread_resume has been called for the
+            # calling thread so that the scheduler will return the outcome to
+            # the calling thread before advancing the sim
```
eric-wieser: I think this would be clearer with a remark like:

> Ensure we notify the current (scheduler) thread that we are about to wake up the background thread,
> making sure to do so before the background thread gets a chance to notify that it is waking again
> via thread_suspend

I'll let you decide the right compromise between your comment and that suggestion

cocotb/scheduler.py
```diff
@@ -752,20 +757,20 @@ def schedule(self, coroutine, trigger=None):
...
+        if not coro_completed:
+            try:
+                result = self._trigger_from_any(result)
+            except TypeError as exc:
+                # restart this coroutine with an exception object telling it that
+                # it wasn't allowed to yield that
+                result = NullTrigger(outcome=outcomes.Error(exc))
 
-        self._coroutine_yielded(coroutine, result)
+            self._coroutine_yielded(coroutine, result)
```
eric-wieser: In my opinion this new conditional is a bit messy, but so is all the handling of external / function - so fine as is for now. Cleaning this up is on my todo list, and it will be a lot easier with this race condition fixed.

eric-wieser:
Ping on updating the comment and commit messages

cocotb/scheduler.py
```diff
@@ -752,20 +757,20 @@ def schedule(self, coroutine, trigger=None):
...
         if self._terminate:
             return
 
-        try:
-            result = self._trigger_from_any(result)
-        except TypeError as exc:
-            # restart this coroutine with an exception object telling it that
-            # it wasn't allowed to yield that
-            result = NullTrigger(outcome=outcomes.Error(exc))
+        if not coro_completed:
```
eric-wieser: @dmiller-jumptrading, I don't really understand the motivation for this change. Happy to put it in, but I can't think of a case where it makes a difference.

cocotb/scheduler.py
eric-wieser: I merged our comments and force-pushed.

cocotb/scheduler.py
eric-wieser: Worked it out myself at least, and your tests fail without it.

eric-wieser:
Updated the commit message and comments myself, code and tests unchanged. Let's put this in when it passes.

PR #1174, opened 2019-10-30
Make Eric maintainer

imphil:
Thanks Eric for accepting the invitation to become cocotb maintainer!

PR #1173, opened 2019-10-30
Push external/function code out of decorators.py and into the scheduler

eric-wieser:
Inspired by #1171 - this pushes all the code from `function.__call__` and `external.__call__` into `scheduler.run_in_executor` and `scheduler.queue_function`.

Neither `scheduler.run_in_executor` nor `scheduler.queue_function` were remotely safe to use directly anyway, and were heavily entangled with the design of the decorators.

After this change, the decorators become nothing but syntactic sugar.

PR #1179, opened 2019-10-30
Add scheduler module to library reference.

### LINKED ISSUE #1178: Add Scheduler to Library Reference
The scheduler module is not at all listed in the Library Reference.

cmarqu:
Closes #1178.

### DIFF
--- a/documentation/source/library_reference.rst
+++ b/documentation/source/library_reference.rst
@@ -67,7 +67,6 @@ classes used within ``cocotb``.
     :members:
     :member-order: bysource
 
-
 Testbench Structure
 ===================
 
@@ -282,3 +281,24 @@ Signal Tracer for WaveDrom
     :members:
     :member-order: bysource
     :synopsis: A signal tracer for WaveDrom.
+
+
+Developer-focused
+=================
+
+The Scheduler
+-------------
+
+.. note::
+    The scheduler object should generally not be interacted with directly - 
+    the only part of it that a user will need is encapsulated in :func:`~cocotb.fork`, 
+    everything else works behind the scenes.
+
+.. currentmodule:: cocotb.scheduler
+
+.. autodata:: cocotb.scheduler
+
+.. autoclass:: Scheduler
+    :members:
+    :member-order: bysource
+
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -85,6 +85,8 @@ if "COCOTB_SIM" in os.environ:
 
 
 scheduler = Scheduler()
+"""The global scheduler instance."""
+
 regression_manager = None
 
 plusargs = {}


eric-wieser:
I think it's worth clarifying that none of these methods should be used directly in test code - perhaps in a ..note::.

There's an awful lot of fragile methods that probably ought to be private.

documentation/source/library_reference.rst
```diff
@@ -66,6 +66,15 @@ classes used within ``cocotb``.
     :members:
     :member-order: bysource
 
+The Scheduler
+-------------
+
+.. currentmodule:: cocotb.scheduler
+
+.. autoclass:: Scheduler
```
eric-wieser: Maybe:
```
.. note::
    The scheduler object should generally not be interacted with directly - the only part of it that you'll need is encapsulated in :func:`cocotb.fork`, everything else works behind the scenes.

Comment (eric-wieser) [reply documentation/source/library_reference.rst]: Also perhaps add:
```
.. py:variable:: scheduler
    The global scheduler instance
```
Worth looking up what the correct Sphinx syntax is for that.

Comment (cmarqu) [reply documentation/source/library_reference.rst]: Added this with a docstring on ``scheduler`` and ``.. autodata:: cocotb.scheduler``.

**Comment by eric-wieser** [comment]
close / reopen to rebuild with the fork change merged

File: documentation/source/library_reference.rst [hunk 2]
```diff
@@ -282,3 +281,24 @@ Signal Tracer for WaveDrom
     :members:
     :member-order: bysource
     :synopsis: A signal tracer for WaveDrom.
+
+
+Developer-focused
+=================
```
eric-wieser: For reviewing: https://external-builds.readthedocs.io/html/cocotb/1179/library_reference.html#developer-focused

cmarqu: I don't immediately see why ``fork()`` is still not linked... :/

eric-wieser: It is linked?

cmarqu: Indeed it is linked, maybe some caching glitch.

cmarqu:
> A little wierd how the docstring for `add` now appears in two places, but this is still clearly an improvement, and I'm not sure there's much that can be done about it anyway.

Maybe adding "``fork()``/``scheduler.add()`` are aliases" makes that clear enough?

eric-wieser:
@cmarqu: It's tempting to either (in another PR):

1. change fork to:
    ```
    def fork(coro):
        """
        long user facing docstring here
        """
        return scheduler.add(coro)
    ```
2. change `Scheduler.add` to `Scheduler._fork`, so that it doesn't appear in the docs for `Scheduler` at all, and is instead just accessible only through `cocotb.fork`
3. Do the above but without the underscore, which at least aligns the names
4. Eliminate `Scheduler.add` entirely, and move its contents to `cocotb.fork`. It's not clear to me why the scheduler itself uses `self.add` rather than `self.schedule`. It's something I want to look into, but i'm waiting on #1173 which is waiting on a review from @imphil before I investigate further.

eric-wieser:
I'd merge this, but I'm waiting for the merge of #1174 before merging anything non-trivial

eric-wieser:
Nevermind, this is definitely under the realm of "trivial"

eric-wieser:
Thanks @cmarqu!

PR #1180, opened 2019-11-02
Allow dropping into interactive pdb shell when hitting exception.  Ad…

benreynwar:
…d COCOTB_PDB_ON_EXCEPTION environment variable.

I've been using this with cocotb-test, and has worked well for me.
I haven't checked that it works with the Makefile flow.

This makes it much easier to debug coroutines since when the the test fails you're dropped directly into the interactive pdb shell where you can better investigate why exactly the test failed.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -33,6 +33,12 @@ from itertools import product
 import sys
 import os
 import traceback
+import pdb
+
+if "COCOTB_PDB_ON_EXCEPTION" in os.environ:
+    _pdb_on_exception = True
+else:
+    _pdb_on_exception = False
 
 if "COCOTB_SIM" in os.environ:
     import simulator
@@ -325,6 +331,12 @@ class RegressionManager(object):
 
         else:
             self.log.error("Test Failed: " + _result_was(), exc_info=exc_info)
+            if _pdb_on_exception:
+                if sys.version_info >= (3, 5):
+                    traceback = exc_info.__traceback__
+                else:
+                    traceback = exc_info[2]
+                pdb.post_mortem(traceback)
             self._add_failure(result)
             result_pass = False
 
--- a/documentation/source/building.rst
+++ b/documentation/source/building.rst
@@ -149,6 +149,10 @@ Environment Variables
     If defined, log lines displayed in the terminal will be shorter. It will print only
     time, message type (``INFO``, ``WARNING``, ``ERROR``, ...) and the log message itself.
 
+.. envvar:: COCOTB_PDB_ON_EXCEPTION
+
+   If defined, cocotb will drop into the Python debugger (:mod:`pdb`) if a test fails with an exception.
+
 .. envvar:: MODULE
 
     The name of the module(s) to search for test functions.  Multiple modules can be specified using a comma-separated list.


cocotb/scheduler.py
```diff
@@ -505,6 +511,8 @@ def unschedule(self, coro):
                 outcome = outcomes.Error(e).without_frames(['unschedule', 'get'])
                 self._test._force_outcome(outcome)
             except Exception as e:
+                if _pdb_on_exception:
+                    pdb.post_mortem(e.__traceback__)
```
eric-wieser: This won't work on Python 2

benreynwar: I modified it so it logs a warning if the exception doesn't have a '__traceback__' attribute.

cocotb/scheduler.py
```diff
@@ -512,7 +512,10 @@ def unschedule(self, coro):
                 self._test._force_outcome(outcome)
             except Exception as e:
                 if _pdb_on_exception:
-                    pdb.post_mortem(e.__traceback__)
+                    if not hasattr(e, '__traceback__'):
```
ktbarrett: The [documentation](https://docs.python.org/2/library/pdb.html#pdb.post_mortem) seems to indicate that you can call `post_mortem` without an argument to use the currently handled exception. And that would work on Python 2 and Python 3.

benreynwar: Nice!  That's much better.

I've also noticed that deals with the Exceptions here doesn't work when the exception was raised in the cocotb.test function or in any function that the test is directly awaiting.  It only works for explicitly forked coroutines so I also need to fix that.

ktbarrett: Yes, the main test will never be `unschedule`d if it fails, the exception will trickle back up to the regression manager. You'll probably want to move this code in `RegressionManager.handle_result`.

eric-wieser: @ktbarrett is right in both of their comments :)

cocotb/regression.py
```diff
@@ -319,6 +325,12 @@ def _result_was():
             self.log.info("Test errored as expected: " + _result_was())
 
         else:
+            if _pdb_on_exception:
+                if sys.version_info >= (3, 5):
+                    traceback = exc_info.__traceback__
+                else:
+                    traceback = exc_info[2]
+                pdb.post_mortem(traceback)
             self.log.error("Test Failed: " + _result_was(), exc_info=exc_info)
```
eric-wieser: I think it would be helpful for this line to appear before the post-mortem, to remind the user exactly which test they're looking at.

cocotb/regression.py
```diff
@@ -33,6 +33,12 @@
 import sys
 import os
 import traceback
+import pdb
+
+if "COCOTB_PDB_ON_EXCEPTION" in os.environ:
+    _pdb_on_exception = True
+else:
+    _pdb_on_exception = False
```
eric-wieser: This environment variable needs to be documented alongside the other ones in https://github.com/cocotb/cocotb/blob/master/documentation/source/building.rst

documentation/source/building.rst
```diff
@@ -107,6 +107,10 @@ Environment Variables
     If defined, log lines displayed in terminal will be shorter. It will print only
     time, message type (``INFO``, ``WARNING``, ``ERROR``) and log message.
 
+.. envvar:: COCOTB_PDB_ON_EXCEPTION
+
+   If defined, when a python exception is hit, cocotb will drop into the python debugger.
```
eric-wieser:
```suggestion
   If defined, cocotb will drop into the python debugger if a test fails with an exception.
```
As written it suggest that all exceptions will cause pdb, but its really just the ones that result in test failures

benreynwar:
Thanks for helping me get it in shape, and sorry for the slow response time.

cocotb/regression.py
```diff
@@ -320,6 +326,12 @@ def _result_was():
 
         else:
             self.log.error("Test Failed: " + _result_was(), exc_info=exc_info)
+            if _pdb_on_exception:
+                if sys.version_info >= (3, 5):
```
imphil: How that we dropped Python 2 support this condition can be removed (or you leave it in for a follow-up cleanup, as you like)

documentation/source/building.rst
```diff
@@ -107,6 +107,10 @@ Environment Variables
     If defined, log lines displayed in terminal will be shorter. It will print only
     time, message type (``INFO``, ``WARNING``, ``ERROR``) and log message.
 
+.. envvar:: COCOTB_PDB_ON_EXCEPTION
+
+   If defined, cocotb will drop into the python debugger if a test fails with an exception.
```
imphil: nit: python -> Python

imphil:
Please rebase this into a single commit with a descriptive commit message before we can merge it.

cocotb/regression.py
eric-wieser: Let's do this in a follow-up, it will be easier to review all at once I think

documentation/source/building.rst
eric-wieser:
```suggestion
   If defined, cocotb will drop into the Python debugger (:mod:`pdb`) if a test fails with an exception.
```

eric-wieser: @imphil, does it let you hit the "commit suggestion" button above? It doesn't let me...

imphil: no, he most likely disabled "maintainers can update this PR"

eric-wieser: Seems odd github would show me this then:

image

I've opened an issue with github support about this ambiguity

benreynwar: Weird.  I dont't think I disabled that option, and I'm not seeing the box to enable it where it should be.

eric-wieser:
Thanks! I took the liberty of tweaking your commit message - generally you want the first line to be a little shorter than you had it.

Issue #1182, opened 2019-11-04
The sysconfig.get_config_var("LDLIBRARY") returns wrong value, a static Python library


tymonx:
Issue:
```
# Loading sv_std.std
# Loading work.endian_swapper_sv(fast)
# Loading cocotb/examples/endian_swapper/tests/build/libs/x86_64/libvpi.so
# Unable to open lib libpython3.7m.a (libpython3.7m.a: invalid ELF header)
#      -.--ns ERROR    cocotb.gpi                                  gpi_embed.c:139  in embed_init_python               Failed to find Python shared library
#
#      -.--ns INFO     cocotb.gpi                                  gpi_embed.c:102  in set_program_name_in_venv        Using Python virtual environment interpreter at cocotb_test/env/bin/python
#      -.--ns INFO     cocotb.gpi                                GpiCommon.cpp:91   in gpi_print_registered_impl       VPI registered
#      -.--ns ERROR    cocotb.gpi                                  gpi_embed.c:196  in get_module_ref                  Failed to load Python module "cocotb"
#
# End time: 10:09:08 on Nov 04,2019, Elapsed time: 0:00:07
# Errors: 0, Warnings: 0
```

Root cause in the `cocotb/share/makefiles/Makefile.pylib.Linux` file:
```
PYTHON_DYN_LIB:=$(shell $(PYTHON_BIN) -c 'from __future__ import print_function; from distutils import sysconfig; print(sysconfig.get_config_var("LDLIBRARY"))')
```

We have both `libpython3.7m.a` and `libpython3.7m.so` libraries in the same location but the `sysconfig.get_config_var("LDLIBRARY")` call returns the `libpython3.7m.a`

Quick workaround in the `cocotb/share/makefiles/Makefile.pylib.Linux` file that doesn't effect the `.so` extension (`.replace(".a", ".so")`):
```
PYTHON_DYN_LIB:=$(shell $(PYTHON_BIN) -c 'from __future__ import print_function; from distutils import sysconfig; print(sysconfig.get_config_var("LDLIBRARY").replace(".a",".so"))')
```

tymonx:
It seems that this behavior is expected when the Python was configured and built with a disabled shared library support during the `./configure --disable-shared` step. In that case the `LDLIBRARY` variable can point to a Python static library and still you can have a Python shared library under a Python library path. Relying only on the `LDLIBRARY` variable value is not good enough.

https://github.com/python/cpython/blob/master/configure#L5870

```
[...]

PR #1183, opened 2019-11-04
Added support for mixed Python static/shared libraries

tymonx:
Some Python installations can return a static Python library (libpython*.a)
even if a shared Python library (libpython*.so) is also present in a Python
installation library path. The Python replace function is harmless for a Python
dynamic installation

Closes https://github.com/cocotb/cocotb/issues/1182

### DIFF
--- a/cocotb/share/makefiles/Makefile.pylib.Darwin
+++ b/cocotb/share/makefiles/Makefile.pylib.Darwin
@@ -52,4 +52,7 @@ endif
 PYLIBS:=$(shell $(PYTHON_BIN) -c 'from __future__ import print_function; from distutils import sysconfig; import os; print("-l"+os.path.splitext(sysconfig.get_config_var("LIBRARY"))[0][3:])')
 
 PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'from __future__ import print_function; import distutils.sysconfig; print(distutils.sysconfig.get_python_inc())')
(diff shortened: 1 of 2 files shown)


eric-wieser:
This feels like a hack, but seems safe, and at worst will convert a broken install into a slightly-more-confusingly broken install.

Possibly relevant discussion about trying to eliminate parts of this hack: https://mail.python.org/archives/list/python-ideas@python.org/thread/QTCPOM5YBOKCWWNPDP7Z4QL2K6OWGSHL/

tymonx:
@eric-wieser I agree with you, it looks like a hack at the first glance. But for more detailed checks I would move this one-line Python command line solution to a separate Python script and do more proper validation. 

In environment where I am currently working, all Python versions (2.x, 3.x) are configured statically (during a Python `./configure` step) and return the **LDLIBRARY** variable as a `libpython*.a` value but they have also a dynamic Python library in the same location.

imphil:
Ugh. So the install is actually two builds of Python installed on top of each other? I'm surprised it only breaks that little. Let's take this hack for now, but please understand that this is neither a supported configuration, nor something we will put any effort into to keep it working.

Please work with your admins to fix the Python installation.

tymonx:
@imphil Thanks! We already raised an internal issue ticket about it. But they are very persistent for any changes. Lucky we are starting using the Kubernetes + Docker that solves many issues but these old-fashioned farms with a static/dynamic Python mix have a quite rich pool of powerful machines :expressionless: And it would be a waste not to use them.

PR #1184, opened 2019-11-06
Doc improvements

cmarqu:


cocotb/handle.py
```diff
@@ -146,7 +146,14 @@ def __getattr__(self, name):
             return object.__getattr__(self, name)
 
 class RegionObject(SimHandleBase):
-    """Region objects don't have values, they are effectively scopes or namespaces."""
+    """A region object.
```
eric-wieser: Perhaps
```suggestion
    """A region object, such as a scope or namespace.
```
since the first line is often used prominently in docstrings, and as it there's nothing conveyed.

cocotb/handle.py
```diff
@@ -398,17 +410,17 @@ def _getvalue(self):
...
 
     def _setcachedvalue(self, value):
-        raise TypeError("Not permissible to set values on object %s type %s" % (self._name, type(self)))
+        raise TypeError("Not permissible to set values on object %s of type %s" % (self._name, type(self)))
 
     def __le__(self, value):
-        """Overload the less than or equal to operator to
-            provide an hdl-like shortcut
+        """Overload the less-than-or-equal-to operator to
+            provide an HDL-like shortcut
```
eric-wieser: Indent is very weird here. Docstring ought to have a one-line summary, followed by two blank lines.

cocotb/handle.py
```diff
@@ -438,10 +450,19 @@ def __hash__(self):
...
+    won't change within a simulation.
 
-    We can also cache the value since it is elaboration time fixed and won't
-    change within a simulation.
+    Args:
+        handle (int): The GPI handle to the simulator object.
+        path (str): Path to this handle, ``None`` if root.
+        handle_type: The type of the handle 
+            (``simulator.INTEGER``, ``simulator.ENUM``, 
+            ``simulator.REAL``, ``simulator.STRING``).
```
eric-wieser: Aside: It's really tempting to take a dependency on `enum34` for python2, so we can clean up a bunch of these into real enum types.

cocotb/regression.py
```diff
@@ -458,7 +458,19 @@ def _my_test(dut):
...
+    """Factory to automatically generate tests.
+
+    Args:
+        test_function: The function that executes a test.
+            Must take *dut* as the first argument.
+        *args: Remaining arguments are passed directly to the test function.
+            Note that these arguments are not varied. An argument that
+            varies with each test must be a keyword argument to the
+            test function.
+        *kwargs: Remaining keyword arguments are passed directly to the test function.
```
eric-wieser:
```suggestion
        **kwargs: Remaining keyword arguments are passed directly to the test function.
```

It's possible for this to render correctly with sphinx, that you need to use `\*\*` and use an `r"""` string

documentation/source/library_reference.rst
```diff
@@ -30,6 +30,9 @@ Writing and Generating tests
     :members:
     :member-order: bysource
 
+.. autofunction:: cocotb.decorators.RunningCoroutine.log
+
+.. autofunction:: cocotb.decorators.RunningCoroutine.join
```
eric-wieser: I think these would make more sense next to `fork`- you're very unlikely to get a coroutine handle unless you use that.

cocotb/regression.py
cmarqu: Good catch, fixed.
Sphinx appears to handle the unescaped ``**kwargs`` just fine though, have you seen otherwise?

cocotb/handle.py
cmarqu: Done.

cocotb/handle.py
cmarqu: Done.

cocotb/handle.py
cmarqu: In a new issue, yes :)

documentation/source/library_reference.rst
cmarqu: Done (and added also ``kill()``, what do you think about that?)

documentation/source/library_reference.rst
```diff
@@ -30,6 +30,7 @@ Writing and Generating tests
     :members:
     :member-order: bysource
 
+.. autofunction:: cocotb.decorators.RunningCoroutine.log
```
eric-wieser: This isn't a function, it's a property.

I'm not even sure it's worth documenting - I can't think of any scenario where you both actually have access to this object, and want to use it. It seems to only be for internal use.

documentation/source/library_reference.rst
eric-wieser: Good idea, `kill` definitely belongs there.

I was suggesting you move `log` as well, but I'm no longer convinced it even provides any value - see the other comment.

documentation/source/library_reference.rst
cmarqu: Removed.

eric-wieser:
This is waiting on gh-1179 to be merged, right?

cmarqu:
Indeed, thanks for merging that.

cocotb/handle.py
```diff
@@ -438,10 +451,19 @@ def __hash__(self):
         return SimHandleBase.__hash__(self)
 
 class ConstantObject(NonHierarchyObject):
-    """Constant objects have a value that can be read, but not set.
+    """A constant object.
+
+    Constant objects have a value that can be read, but not set.
```
eric-wieser: Perhaps
```
An object which has a value that can be read but not set
```
which then both achieves your goal of being a singular noun, but also provides more information than the class name

cocotb/regression.py
```diff
@@ -493,20 +506,6 @@ class TestFactory(object):
...
-
-            *args: Remaining arguments are passed directly to the test function.
-                   Note that these arguments are not varied. An argument that
-                   varies with each test must be a keyword argument to the
-                   test function.
-            *kwargs: Remaining keyword arguments are passed directly to the test function.
-                   Note that these arguments are not varied. An argument that
-                   varies with each test must be a keyword argument to the
-                   test function.
-        """
```
eric-wieser: Does this render incorrectly if left here?

cocotb/handle.py
cmarqu: Thanks, done.

cocotb/regression.py
cmarqu: No, but it relied on the non-intuitive initial line break immediately after the triple opening quote .

eric-wieser:
Any reason this is still marked as draft?

PR #1185, opened 2019-11-11
Extract version for docs from 'version' file.

### LINKED ISSUE #1077: Documentation version number is stale
Currently conf.py hard-coded the version number to 1.1, but we're currently on 1.2. It would be good if it could extract it in conf.py from elsewhere, so that it never goes out of sync again.

cmarqu:
Closes #1077.

The simplistic splitting on "dot" may break if we start to use weird version tags or put git SHAs in, but supporting that now would be premature optimization.

### DIFF
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -68,11 +68,15 @@ copyright = u'2014-{0}, PotentialVentures'.format(datetime.datetime.now().year)
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
-#
+
+version_raw = open(os.path.join(os.path.dirname(__file__), "..", "..", "version")).read()[8:].strip()
+version_elems = version_raw.split('.')
+
 # The short X.Y version.
-version = '1.1'
+version = '.'.join([version_elems[0], version_elems[1]])
+
 # The full version, including alpha/beta/rc tags.
-release = '1.1'
+release = version_raw
 
 autoclass_content = "both"
 


eric-wieser:
What are your thoughts on solving this like I did in https://github.com/pygae/clifford/pull/172 + https://github.com/pygae/clifford/pull/178?

documentation/source/conf.py
```diff
@@ -68,11 +68,15 @@
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
-#
+
+version_raw = open(os.path.join("..", "..", "version")).read()[8:].strip()
```
eric-wieser:
```suggestion
version_raw = open(os.path.join(os.path.dirname(__file__), "..", "..", "version")).read()[8:].strip()
```
Would still prefer the solution I propose in the comment above

cmarqu: Done that for now.

cmarqu:
> What are your thoughts on solving this like I did in [pygae/clifford#172](https://github.com/pygae/clifford/pull/172) + [pygae/clifford#178](https://github.com/pygae/clifford/pull/178) (it took two tries, I forgot to commit the first time)

This is certainly cleaner than reading the file ``version`` in two places (``setup.py`` and ``conf.py``), but I'd want to hear what @imphil thinks.

eric-wieser:
There's a third place - we write it into the `VERSION` environment variable as `VERSION=$(shell $(PYTHON_BIN) -c 'import pkg_resources; print(pkg_resources.get_distribution("cocotb").version)')`, then have python look at that environment variable again.

Which seems daft, if we can just store the

PR #1186, opened 2019-11-14
Aliasing of bus signals

Fatsie:
I use this feature in my wishbone code to map standard conformant dat_i/dat_o to datwr/datrd of cocotb Wishbone bus. I have also seen other namings like dat_miso/dat_mosi and that should all be able to be handled by this patch.

### DIFF
--- a/cocotb/bus.py
+++ b/cocotb/bus.py
@@ -54,7 +54,8 @@ class Bus(object):
     TODO:
         Support for ``struct``/``record`` ports where signals are member names.
     """
-    def __init__(self, entity, name, signals, optional_signals=[], bus_separator="_", array_idx=None):
+    def __init__(self, entity, name, signals, optional_signals=[], bus_separator="_",
+                 array_idx=None, aliases=None):
         """
         Args:
             entity (SimHandle): :any:`SimHandle` instance to the entity containing the bus.
@@ -72,12 +73,35 @@ class Bus(object):
             bus_separator (str, optional): Character(s) to use as separator between bus
                 name and signal name. Defaults to '_'.
             array_idx (int or None, optional): Optional index when signal is an array.
+            aliases (dict, optional): gives aliases for bus names so one can adopt to
+                different signal naming conventions. The index is the name of the signal
+                for the bus and the value is the name in the RTL code.
+                For example for Wishbone slave {"datrw": "dat_i", "datrd": "dat_o"}
         """
         self._entity = entity
         self._name = name
         self._signals = {}
 
-        for attr_name, sig_name in _build_sig_attr_dict(signals).items():
+        _signals = _build_sig_attr_dict(signals)
+        _optional_signals = _build_sig_attr_dict(optional_signals)
+
+        if aliases is not None:
+            if not isinstance(aliases, dict):
+                raise ValueError("aliases parameter has to be a dict")
+
+            for attr_name, port_name in aliases.items():
+                if attr_name in signals:
+                    _signals[attr_name] = port_name
+                elif attr_name in optional_signals:
+                    _optional_signals[attr_name] = port_name
+                else:
+                    raise ValueError(
+                        "'{}' not signal of bus {}".format(
+                            attr_name, self.__class__.__name__
+                        )
+                    )
+
+        for attr_name, sig_name in _signals.items():
             if name:
                 signame = name + bus_separator + sig_name
             else:
@@ -88,7 +112,7 @@ class Bus(object):
             self._add_signal(attr_name, signame)
 
         # Also support a set of optional signals that don't have to be present
-        for attr_name, sig_name in _build_sig_attr_dict(optional_signals).items():
+        for attr_name, sig_name in _optional_signals.items():
             if name:
                 signame = name + bus_separator + sig_name
             else:
--- a/cocotb/drivers/__init__.py
+++ b/cocotb/drivers/__init__.py
@@ -37,7 +37,6 @@ from cocotb.triggers import (Event, RisingEdge, ReadOnly, NextTimeStep,
                              Edge)
 from cocotb.bus import Bus
 from cocotb.log import SimLog
-from cocotb.utils import reject_remaining_kwargs
 
 
 class BitDriver(object):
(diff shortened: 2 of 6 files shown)


imphil:
How is that functionality different from the `signals` parameter?

Fatsie:
``BusDriver`` and ``BusMonitor`` and thus also it's child classes like ``WishboneMaster`` have the signals in the ``_signals`` attribute; e.g.

```python
class Wishbone(BusDriver):
    _signals = ["cyc", "stb", "we", "sel", "adr", "datwr", "datrd", "ack"]
    _optional_signals = ["err", "stall", "rty"]
```

AFAICS you are not supposed to override these definitions anymore and if you do you will override everything. If your code now uses dat_o/dat_i you can do with my patch:
```python
    wb = Wishbone(dut, "WB", clk, aliases={"datwr": "dat_o", "datrd": "dat_i"})
```

eric-wieser:
It sounds like the fix is needed in `BusDriver` and `BusMonitor` which isn't exposing the API of `Bus` very well, rather than adding a side-channel to `Bus`/

eric-wieser:
That is to say, perhaps `BusDriver` should do something like
```
signals = kwargs.pop("signals", {})
expected_signals = dict(self._signals) if isinstance(self._signals, dict) else {n: n for n in self._signals}
expected_optional_signals = dict(self._optional_signals) if isinstance(self._optional_signals, dict) else {n: n for n in self._optional_signals}
for n, alias in signals:
    if n in expected_signals:
        expected_signals[n] = alias 
    elif n in expected_optional_signals:
        expected_optional_signals[n] = alias
    else:
        raise ValueError("Passed a signal name that is not part of this bus")
```

Fatsie:
I don't see this as a side channel but a convenience feature to easily adopt to RTL code not conforming to expected/standardized naming.

Originally I had this actually implemented in ``BusDriver`` but moved it to ``Bus`` once I saw I actually needed the exact same code in ``BusMonitor``. When working on Wishbone support in general I did not like the fact that current ``BusDriver``/``BusMonitor`` setup actually forces to duplicate code between ``WishboneDriver`` and ``WishboneMonitor``.

Actually with the move out from the actual drivers out of cocotb I would advocate to move ``BusDriver``/``BusMonitor`` into `bus.py`

eric-wieser:
> in general I did not like the fact that current BusDriver/BusMonitor setup actually forces to duplicate code between WishboneDriver and WishboneMonitor.

Can you give an example of such code duplication?

eric-wieser:
> but a convenience feature to easily adopt to RTL code not conforming to expected/standardized naming.

That's exactly my point - `Bus` does not know about expected/standardized naming, all it knows is "here is a mapping between signal names and python names, please use them".

It's the driver / monitor classes that introduce the idea of standardized naming, so the code to allow alias belongs there.

Fatsie:
> Can you give an example of such code duplication?

The definition of _signals and _optional_signals for a start and now this aliasing feature.

eric-wieser:
The aliasing feature would at least live in the `cocotb` base class though, right? (once we get this PR into a state where everyone agrees)

I see your point about signals being duplicated - I assume you're referring to something like:
```python
class WishboneDriver(BusDriver):
    _signals = [
        'my', 'signals'
    ]  # way more here, enough to be annoying
class WishboneMonitor(BusMonitor):
    _signals = ['my', 'signals']
```
Remember you can always write
```python
class WishboneDriver(BusDriver):
    _signals = [
        'my', 'signals'
    ]  # way more here, enough to be annoying
class WishboneMonitor(BusMonitor):
    _signals = WishboneDriver._signals 
```

Fatsie:
> Remember you can always write
> 
> ```python
> class WishboneDriver(BusDriver):
>     _signals = [
>         'my', 'signals'
>     ]  # way more here, enough to be annoying
> class WishboneMonitor(BusMonitor):
>     _signals = WishboneDriver._signals 
> ```

But in the wishbone branch in cocotb that I took as start for my work, ``WishboneDriver`` lives in `drivers/wishbone.py` and ``WishboneMonitor`` in `monitors/wishbone.py`. Actually I would like to do:
```python
class Wishbone(Bus):
    _signals = [...]
    _optional_signals = [...]

...

class WishboneMaster(Wishbone, BusDriver):

...

class WishboneSlave(Wishbone, BusDriver):

...

class WishboneMonitor(Wishbone, BusMonitor):

...
```

And this would then live in one file.

Fatsie:
> That's exactly my point - `Bus` does not know about expected/standardized naming, all it knows is "here is a mapping between signal names and python names, please use them".
> 
> It's the driver / monitor classes that introduce the idea of standardized naming, so the code to allow aliases belongs there.

What is the rationale to have ``Bus`` have signals as parameter to ``__init__()`` and have ``BusDriver``/``BusMonitor`` have it as _signal class attribute ?

eric-wieser:
I don't agree with making `issubclass(WishboneMaster, WishboneBus)` true, that sounds like a misuse of inheritance.

What makes sense to me is adding a new class (better name suggestions welcome)
```python
class ConcreteBus(Bus):
    """ A bus with the set of allowed signal names baked into the type """
    # override in subclass
    _optional_signals = {}
    _signals = {}

    def __init__(entity, name, *, signals={}, **kwargs):
        signals = _build_sig_attr_dict(self._signals).copy()
        optional_signals = _build_sig_attr_dict(self._optional_signals).copy()
        for n, alias in signals,item():
            if n in signals:
                signals[n] = alias
            elif n in optional_signals:
                optional_signals[n] = alias
            else:
                raise ValueError("Passed a signal name, {!r}, that is not part of this bus".format(n))

        super().__init__(entity, name, signals, optional_signals, **kwargs)
```
which will be used within `BusDriver` as:
```
class BusDriver(Driver):
    def __init__(self, entity, name, clock, **kwargs):
        if self._bus_type is None:
            # TODO: emit a warning here telling the user to use the new API
            class _AnonymousBus(ConcreteBus):
                _optional_signals = self._optional_signals
                _signals = self._signals
            self._bus_type = _AnonymousBus

        self.bus = self._bus_type(entity, name, **kwargs)
```
Allowing your code to become
```python
class WishboneBus(ConcreteBus):
    _signals = ...


class WishBoneDriver(BusDriver):
    _bus_type = WishboneBus
```

Fatsie:
Agree with proposal, will rework patch.
What about ``TypedBus`` ?

eric-wieser:
`TypedBus` sounds like a better name to me

Fatsie:
> ```python
> class ConcreteBus(Bus):
>     """ A bus with the set of allowed signal names baked into the type """
>     # override in subclass
>     _optional_signals = {}
>     _signals = {}
> ```

``_signal``/``_optional_signals`` seems good candidates to be an  ``abstractproperty`` but implementing this in a way compatible with 2.7 and 3 seems a PITA.

eric-wieser:
@Fatsie: If you haven't already found it, take a look at the py3 compat file. That at least gives you the tools to use `abc.ABC` in python 2, and to attach metaclasses.

Fatsie:
> If you haven't already found it, take a look at the py3 compat file. That at least gives you the tools to use abc.ABC in python 2, and to attach metaclasses.

I am using six now, which seems to be commonly used in other code for python2/3 compatibility.

BTW, as I am reworking the patch fully as agreed and it is not about aliasing anymore, I will close this PR and refer to it in new PR for reworked patch.

PR #1187, opened 2019-11-14
Use more pythonic dict initialization.

Fatsie:
Small pythonic optimization; please let me know if such patches are not wanted.
I can suppress the urge to perform such changes if need be.

### DIFF
--- a/cocotb/bus.py
+++ b/cocotb/bus.py
@@ -36,10 +36,7 @@ def _build_sig_attr_dict(signals):
     if isinstance(signals, dict):
         return signals
     else:
-        sig_to_attr = {}
-        for sig in signals:
-            sig_to_attr[sig] = sig
-        return sig_to_attr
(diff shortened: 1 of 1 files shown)


cocotb/bus.py
```diff
@@ -36,10 +36,7 @@ def _build_sig_attr_dict(signals):
     if isinstance(signals, dict):
         return signals
     else:
-        sig_to_attr = {}
-        for sig in signals:
-            sig_to_attr[sig] = sig
-        return sig_to_attr
+        return dict([(sig, sig) for sig in signals])
```
eric-wieser: Why not go one further?
```suggestion
        return {sig: sig for sig in signals}
```

Fatsie: Always happy to learn new tricks.
Changed as suggested.

eric-wieser:
Thanks!

PR #1188, opened 2019-11-14
Made configuration data global in cocotb.config

Fatsie:
This way the data can be accessed inside a python script without
needing to call `cocotb-config`.

I am considerig to use cocotb for simulating nmigen code from a nmigen support class. With this change I could find out inside this class the path of the cocotb makefiles without needing to fire up a shell that calls `cocotb-config`.

### DIFF
--- a/cocotb/config.py
+++ b/cocotb/config.py
@@ -28,11 +28,28 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
+"""
+Module for querying the cocotb configuration
+
+This module provides information in module global variables and through a
+``main()`` function that is used in the cocotb-config script.
+
+Global variables:
+    share_dir: str, path where the cocotb data is stored
+    makefiles_dir: str, path where the cocotb makefiles are installed
+"""
 import os
 import sys
 import cocotb
 import argparse
 import pkg_resources
+
+
+__all__ = ["share_dir", "makefiles_dir"]
+
+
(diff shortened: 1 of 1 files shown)


eric-wieser:
If you're going to make these public, can you add a docstring (just below the license)?

Something like
```
"""
Can be run at the command line to do ...

.. data:: share_dir
    the package-share of cocotb
.. data:: prefix_dir
    ...
"""

File: cocotb/config.py [hunk 1]
```diff
@@ -28,13 +28,33 @@
...
 ###############################################################################
 
+"""Module for querying the cocotb configuration
+
+This module provides information in module global variables and through a
+``main()`` function that is used in the cocotb-config script.
+
+Global variables:
+    version: str, the cocotb version
+    prefix_dir: str, path where cocotb module is installed
```
eric-wieser: I'm not actually sure we want to expose either of these here. The version should just be at `cocotb.__version__`, and the `prefix_dir` doesn't strike me as useful - the user can always just use `cocotb.__file__` themselves

cocotb/config.py
```diff
@@ -28,13 +28,30 @@
...
+    makefiles_dir: str, path where the cocotb makefiles are installed
+"""
 import os
 import sys
 import cocotb
 import argparse
 import pkg_resources
 
 
+__all__ = ["share_dir" "makefiles_dir"]
```
eric-wieser:
```suggestion
__all__ = ["share_dir", "makefiles_dir"]
```

Fatsie: Done

Issue #767, opened 2019-01-13
Dropping Python 2 support


eric-wieser:
Upstream support for Python 2.7 [ends on 2020-01-01](https://pythonclock.org/).

I know it's a long way off right now, and the cocotb release cycle is in its infancy - but at some point it would be good for cocotb to decide if and/or when to make its final Python 2.7 release, as [many other projects are already doing](https://python3statement.org/).

Creating this issue now to act as a catalyst for this discussion - it doesn't need to be acted on any time soon.

Matthewar:
Potentially somewhat related to #610.
Also, while Python 2.7 is reaching end of its upstream support, are enterprise solutions (such as RHEL7) [promising support until the end of their lives](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/7.5_release_notes/chap-red_hat_enterprise_linux-7.5_release_notes-deprecated_functionality)?
That doesn't mean that we can't start utilising Python3 features such as for the scheduler (so long, as discussed in that issue, old testbench syntax still works) but perhaps a branch would remain with Python2 backported fixes?

stuarthodgson:
There seem to be a lot of outstanding issues and features people want. Dropping python 2 seems quite low down. It's not as if 3 does not work, so why the push?   Even when 2 maintenance ends it's not like world will stop using it.

eric-wieser:
I'd propose we make 1.3 the last minor release to support Python 2.

It's clear to me that 1.2 (already released) cannot be the last release for Python 2. There are many fixes for Python 2 in master, and it would be foolish to never release them - even if the release does come after the Python 2.7 EOL date. I don't think there's much need to discuss this here.

The more controversial claim I'm making here, and the one I'd like responses to, is that we should make cocotb 1.4.x support only Python >= 3.5.

ktbarrett:
Perhaps we can please all parties by instead of going to cocotb 1.4.x and Python 3 after this release, move to cocotb 2.0; allowing a cocotb 1.x branch to stay for long-term Python 2 support. Those that are stuck or unwilling to upgrade to Python 3 will still be allowed a place to contribute back bugfixes for Python 2, or even backport features from the 2.x branch if they are particularly stubborn. I don't think we need to put as much resources into maintaining the 1.x branch as developing the 2.x branch, but there should be at least one maintainer who is willing to look at 1.x issues.

The [Catch](https://github.com/catchorg/Catch2) project does this.

hugovk:
[...]

PR #1190, opened 2019-11-18
Bus cleanup patch

Fatsie:
This is a clean-up patch to prepare for the TypedBus patch as discussed in #1186.
Some extra comments next to commit messages:
* I did delete AD9361 driver as it seemed to be wrongly subclassing ``BusDriver`` and I don't think this code is maintained. Alternatively I could have it subclass ``Driver`` but I have no way of checking that patch.
* I removed ``AvalonSTPktsWithChannel`` and added ``report_channel`` parameter to ``AvalanSTPkts.__init__()`` which I think should allow for the same functionality. I need input if that is not the case.
  It does mean that code using ``AvalonSTPktsWithChannel`` has to be adapted.

### DIFF
--- a/cocotb/drivers/__init__.py
+++ b/cocotb/drivers/__init__.py
@@ -37,7 +37,6 @@ from cocotb.triggers import (Event, RisingEdge, ReadOnly, NextTimeStep,
                              Edge)
 from cocotb.bus import Bus
 from cocotb.log import SimLog
-from cocotb.utils import reject_remaining_kwargs
 
 
 class BitDriver(object):
@@ -215,32 +214,30 @@ class BusDriver(Driver):
         * a name
         * an entity
 
-        Args:
-            entity (SimHandle): A handle to the simulator entity.
-            name (str or None): Name of this bus. ``None`` for a nameless bus, e.g.
-                bus-signals in an interface or a ``modport``.
-                (untested on ``struct``/``record``, but could work here as well).
-            clock (SimHandle): A handle to the clock associated with this bus.
-            array_idx (int or None, optional): Optional index when signal is an array.
-            bus_separator (str, optional): Character(s) to use as separator between bus
-                name and signal name. Defaults to '_'.
+    Args:
+        entity (SimHandle): A handle to the simulator entity.
+        name (str or None): Name of this bus. ``None`` for a nameless bus, e.g.
+            bus-signals in an interface or a ``modport``.
+            (untested on ``struct``/``record``, but could work here as well).
+        clock (SimHandle): A handle to the clock associated with this bus.
+        **kwargs (dict): Keyword arguments forwarded to :class:`cocotb.Bus`,
+            see docs for that class for more information.
 
     """
     
     _optional_signals = []
 
     def __init__(self, entity, name, clock, **kwargs):
-        # emulate keyword-only arguments in python 2
-        index = kwargs.pop("array_idx", None)
-        bus_separator = kwargs.pop("bus_separator", "_")
-        reject_remaining_kwargs('__init__', kwargs)
+        index = kwargs.get("array_idx", None)
 
         self.log = SimLog("cocotb.%s.%s" % (entity._name, name))
         Driver.__init__(self)
         self.entity = entity
         self.clock = clock
-        self.bus = Bus(self.entity, name, self._signals,
-                       self._optional_signals, bus_separator=bus_separator, array_idx=index)
+        self.bus = Bus(
+            self.entity, name, self._signals, optional_signals=self._optional_signals,
+            **kwargs
+        )
 
         # Give this instance a unique name
         self.name = name if index is None else "%s_%d" % (name, index)
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -51,8 +51,8 @@ class AXI4LiteMaster(BusDriver):
                 "ARVALID", "ARADDR", "ARREADY",        # Read address channel
                 "RVALID", "RREADY", "RRESP", "RDATA"]  # Read data channel
 
-    def __init__(self, entity, name, clock):
-        BusDriver.__init__(self, entity, name, clock)
+    def __init__(self, entity, name, clock, **kwargs):
+        BusDriver.__init__(self, entity, name, clock, **kwargs)
 
         # Drive some sensible defaults (setimmediatevalue to avoid x asserts)
         self.bus.AWVALID.setimmediatevalue(0)
@@ -237,9 +237,9 @@ class AXI4Slave(BusDriver):
     ]
 
     def __init__(self, entity, name, clock, memory, callback=None, event=None,
-                 big_endian=False):
+                 big_endian=False, **kwargs):
 
-        BusDriver.__init__(self, entity, name, clock)
+        BusDriver.__init__(self, entity, name, clock, **kwargs)
         self.clock = clock
 
         self.big_endian = big_endian
--- a/cocotb/drivers/avalon.py
+++ b/cocotb/drivers/avalon.py
@@ -245,8 +245,8 @@ class AvalonMemory(BusDriver):
             }
 
     def __init__(self, entity, name, clock, readlatency_min=1,
-                 readlatency_max=1, memory=None, avl_properties={}):
-        BusDriver.__init__(self, entity, name, clock)
+                 readlatency_max=1, memory=None, avl_properties={}, **kwargs):
+        BusDriver.__init__(self, entity, name, clock, **kwargs)
 
         if avl_properties != {}:
             for key, value in self._avalon_properties.items():
--- a/cocotb/drivers/avalon.py
+++ b/cocotb/drivers/avalon.py
@@ -512,9 +512,9 @@ class AvalonST(ValidatedBusDriver):
 
     _default_config = {"firstSymbolInHighOrderBits" : True}
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, entity, name, clock, **kwargs):
         config = kwargs.pop('config', {})
-        ValidatedBusDriver.__init__(self, *args, **kwargs)
+        ValidatedBusDriver.__init__(self, entity, name, clock, **kwargs)
 
         self.config = AvalonST._default_config.copy()
 
@@ -605,9 +605,9 @@ class AvalonSTPkts(ValidatedBusDriver):
         "readyLatency"                  : 0
     }
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, entity, name, clock, **kwargs):
         config = kwargs.pop('config', {})
-        ValidatedBusDriver.__init__(self, *args, **kwargs)
+        ValidatedBusDriver.__init__(self, entity, name, clock, **kwargs)
 
         self.config = AvalonSTPkts._default_config.copy()
 
--- a/cocotb/drivers/opb.py
+++ b/cocotb/drivers/opb.py
@@ -47,8 +47,8 @@ class OPBMaster(BusDriver):
     _optional_signals = ["seqAddr"]
     _max_cycles = 16
 
-    def __init__(self, entity, name, clock):
-        BusDriver.__init__(self, entity, name, clock)
+    def __init__(self, entity, name, clock, **kwargs):
+        BusDriver.__init__(self, entity, name, clock, **kwargs)
         self.bus.select.setimmediatevalue(0)
         self.log.debug("OPBMaster created")
         self.busy_event = Event("%s_busy" % name)
--- a/cocotb/monitors/avalon.py
+++ b/cocotb/monitors/avalon.py
@@ -55,9 +55,9 @@ class AvalonST(BusMonitor):
 
     _default_config = {"firstSymbolInHighOrderBits": True}
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, entity, name, clock, **kwargs):
         config = kwargs.pop('config', {})
-        BusMonitor.__init__(self, *args, **kwargs)
+        BusMonitor.__init__(self, entity, name, clock, **kwargs)
 
         self.config = self._default_config.copy()
 
@@ -109,10 +109,10 @@ class AvalonSTPkts(BusMonitor):
         "invalidTimeout"                : 0,
     }
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, entity, name, clock, **kwargs):
         config = kwargs.pop('config', {})
         report_channel = kwargs.pop('report_channel', False)
-        BusMonitor.__init__(self, *args, **kwargs)
+        BusMonitor.__init__(self, entity, name , clock, **kwargs)
 
         self.config = self._default_config.copy()
         self.report_channel = report_channel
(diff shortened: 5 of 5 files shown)


cocotb/drivers/ad9361.py
```diff
@@ -1,214 +0,0 @@
...
-
-import cocotb
-from cocotb.triggers import Timer, RisingEdge, Event
-from cocotb.drivers import BusDriver
-from cocotb.binary import BinaryValue, BinaryRepresentation
-
-from collections import deque
-
-
-class AD9361(BusDriver):
```
eric-wieser: Instead of deleting this class as part of this Pr, I'd just remove the base class, and open an issue about removing it.

cocotb/drivers/__init__.py
```diff
@@ -221,26 +220,22 @@ class BusDriver(Driver):
...
     
     _optional_signals = []
 
     def __init__(self, entity, name, clock, **kwargs):
-        # emulate keyword-only arguments in python 2
-        index = kwargs.pop("array_idx", None)
-        bus_separator = kwargs.pop("bus_separator", "_")
-        reject_remaining_kwargs('__init__', kwargs)
+        index = kwargs.get("array_idx")
+        kwargs["optional_signals"] = self._optional_signals + kwargs.get("optional_signals", [])
```
eric-wieser: It's not clear to me why this makes sense - what is the use case of passing extra optional signals here?

cocotb/drivers/avalon.py
```diff
@@ -512,9 +513,8 @@ class AvalonST(ValidatedBusDriver):
 
     _default_config = {"firstSymbolInHighOrderBits" : True}
 
-    def __init__(self, *args, **kwargs):
-        config = kwargs.pop('config', {})
-        ValidatedBusDriver.__init__(self, *args, **kwargs)
+    def __init__(self, entity, name, clock, *, config={}, **kwargs):
```
eric-wieser: This is python3-only syntax, and we haven't yet decided what to do for #767. At any rate, we should probably aim to do at least one more python 2 release.

cocotb/drivers/opb.py
```diff
@@ -47,8 +47,8 @@ class OPBMaster(BusDriver):
     _optional_signals = ["seqAddr"]
     _max_cycles = 16
 
-    def __init__(self, entity, name, clock):
-        BusDriver.__init__(self, entity, name, clock)
+    def __init__(self, entity, name, clock, **kwargs):
+        BusDriver.__init__(self, entity, name, clock, **kwargs)
```
eric-wieser: All these `BusDriver.__init__` changes are great, thanks

cocotb/monitors/__init__.py
```diff
@@ -174,14 +174,16 @@ class BusMonitor(Monitor):
...
+                 **kwargs):
         self.log = SimLog("cocotb.%s.%s" % (entity._name, name))
         self.entity = entity
         self.name = name
         self.clock = clock
         self.bus = Bus(self.entity, self.name, self._signals,
                        optional_signals=self._optional_signals,
-                       bus_separator=bus_separator, array_idx=array_idx)
+                       bus_separator=bus_separator, array_idx=array_idx,
+                       **kwargs)
```
eric-wieser: I'm not sold on this - what kwargs are you expecting?

cocotb/drivers/ad9361.py
Fatsie: I'll leave the AS9361 in as is then and make separate issue for removing it.
The code will only cause problems when actually trying to create an object of that type and also when removing the base class.

cocotb/monitors/__init__.py
Fatsie: In follow-up patch introducing ``TypedBus`` the Bus class will not be fixed and different call signatures need to be supported. But I can do that in the follow-up patch then. I will also plan to rename kwargs to bus_kwargs then. I can either move this patch to follow-up patch or do the following:
```python
class BusMonitor(Monitor):
    """Wrapper providing common functionality for monitoring buses."""
    _signals = []
    _optional_signals = []

    def __init__(self, entity, name, clock, reset=None, reset_n=None,
                 callback=None, event=None, **bus_kwargs):
        self.log = SimLog("cocotb.%s.%s" % (entity._name, name))
        self.entity = entity
        self.name = name
        self.clock = clock
        self.bus = Bus(self.entity, self.name, self._signals,
                       optional_signals=self._optional_signals,
                       **bus_kwargs)
```

cocotb/drivers/avalon.py
Fatsie: So the question is then what is preferred:
```python
    def __init__(self, entity, name, clock, config={}, **kwargs):
```
or
```python
    def __init__(self, entity, name, clock, **kwargs):
        config = kwargs.pop('config', {})
```
I like the former because then the parameter is shown when doing help in ipython etc. but then people could wrongly use positional passing of config.

eric-wieser: I would go for the latter - we can fix the docs as soon as we drop python 2, but we can't change the API once we release it.

cocotb/monitors/__init__.py
eric-wieser: Let's move this to the follow-up, I think it will be easier to understand in context.

cocotb/drivers/ad9361.py
eric-wieser: Does this even use a single method from `BusDriver`? If not, removing the base class ought to be perfectly safe

cocotb/monitors/__init__.py
Fatsie: I'll move it to follow-up patch, it's more natural there.

cocotb/drivers/avalon.py
Fatsie: Will keep use of pop for the moment; can be reconsidered later.

Fatsie:
* Fully reworked patch based on Erik's comments.
* Some extra enforcements of passing by name using ``kwargs.pop()``

cocotb/drivers/amba.py
```diff
@@ -236,10 +236,11 @@ class AXI4Slave(BusDriver):
...
     ]
 
-    def __init__(self, entity, name, clock, memory, callback=None, event=None,
-                 big_endian=False):
-
-        BusDriver.__init__(self, entity, name, clock)
+    def __init__(self, entity, name, clock, memory, **kwargs):
+        callback = kwargs.pop('callback', None)
+        event = kwargs.pop('event', None)
+        big_endian = kwargs.pop('big_endian', False)
```
eric-wieser: Making these keyword-only when they used not to be is a breaking change, sadly. I don't think we should risk breaking existing users for such minor cleanup.

Fatsie: I'll move it to follow-up patch.

Fatsie:
Only change is removing patch enforcing passing default parameters by name.
Moved to follow-up patch.

cocotb/drivers/__init__.py
```diff
@@ -215,32 +214,29 @@ class BusDriver(Driver):
...
-            bus_separator (str, optional): Character(s) to use as separator between bus
-                name and signal name. Defaults to '_'.
+    Args:
+        entity (SimHandle): A handle to the simulator entity.
+        name (str or None): Name of this bus. ``None`` for a nameless bus, e.g.
+            bus-signals in an interface or a ``modport``.
+            (untested on ``struct``/``record``, but could work here as well).
+        clock (SimHandle): A handle to the clock associated with this bus.
+        **kwargs (dict): kwargs is passed on to __init__ of the created .bus field.
+            See ``cocotb.Bus.__init__()`` for more information.
```
eric-wieser: Perhaps just ``Keyword arguments forwarded to :class:`cocotb.Bus`, see the docs for that class for more information.``

cocotb/monitors/avalon.py
```diff
@@ -100,15 +100,20 @@ class AvalonSTPkts(BusMonitor):
...
 
         self.config = self._default_config.copy()
+        self.report_channel = report_channel
 
         # Set default config maxChannel to max value on channel bus
         if hasattr(self.bus, 'channel'):
             self.config['maxChannel'] = (2 ** len(self.bus.channel)) -1
+        else:
+            if report_channel:
+                self.log.error("Channel reporting asked on bus without channel signal")
```
eric-wieser: This should be `raise ValueError(...)`

cocotb/monitors/avalon.py
```diff
@@ -222,20 +230,3 @@ def valid():
...
-    def __init__(self, *args, **kwargs):
-        AvalonSTPkts.__init__(self, *args, **kwargs)
-
-    def _recv(self, pkt):
-        """Force use of channel in recv function.
-
-        Args:
-            pkt: (string) Monitored data.
-        """
-        AvalonSTPkts._recv(self, {"data": pkt, "channel": self.channel})
```
eric-wieser: Again, removing this class will break any existing code that uses it.

It looks like you could keep old code working by removing everything but `__init__`, and changing it to pass `record_channel=True` to `__init__`? Perhaps:
```
def __init__(*args, **kwargs):
    warnings.warn(
        "AvalonSTPktsWithChannel is deprecated, use AvalonSTPkts(..., record_channel=True)",
    AvalonSTPkts.__init__(self, *args, record_channel=True, **kwargs)

File: cocotb/monitors/avalon.py [hunk 10]
```diff
@@ -100,15 +100,20 @@ class AvalonSTPkts(BusMonitor):
         "invalidTimeout"                : 0,
     }
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, entity, name, clock, **kwargs):
         config = kwargs.pop('config', {})
-        BusMonitor.__init__(self, *args, **kwargs)
+        report_channel = kwargs.pop('report_channel', False)
```
eric-wieser: Please document this new argument

cocotb/monitors/avalon.py
Fatsie: OK, that was original version of patch but changed it to use self.log.error. Any guidance on when the raise and when to use self.log.error ?

eric-wieser: Use `raise` if the caller called your function in an invalid way, and it doesn't make sense to continue.

Use `self.log.error` if the bus itself is misbehaving, and throwing an error immediately would leave the bus in an invalid state. Even if you do eventually throw an exception, if you weren't able to do it immediately, you should also `log.error` so that the sim time of when things went wrong is recorded.

TL;DR: `log.error` is for humans only, `raise` is for both humans and code.

cocotb/monitors/avalon.py
Fatsie: Can I use [deprecated](https://pypi.org/project/Deprecated/) ? I really think we should get rid of this old legacy somewhere in the future.

eric-wieser: I'd prefer not to in this patch. Cocotb currently has no dependencies, we should have that discussion independently.

Perhaps open an issue suggesting we replace all instances of `DeprecationWarning` with that third party module, and keep doing it by hand for now.

Agreed that it's nice to remove legacy, but if it's user facing we should always deprecate first.

Fatsie: warn & DeprecationWarning is fine. Was just used to @deprecated.

cocotb/drivers/__init__.py
Fatsie: OK

Fatsie:
Handled Erik's comments

cocotb/monitors/avalon.py
```diff
@@ -224,18 +241,15 @@ def valid():
...
+    """
 
-        Args:
-            pkt: (string) Monitored data.
-        """
-        AvalonSTPkts._recv(self, {"data": pkt, "channel": self.channel})
+    def __init__(self, entity, name, clock, **kwargs):
+        warnings.warn(
+            "Use of AvalonSTPktsWithChannel is deprecated\n"
+            "\tUse AvanlonSTPkts(..., report_channel=True, ...)",
```
eric-wieser:
```suggestion
            "\tUse AvalonSTPkts(..., report_channel=True, ...)",
```

cocotb/monitors/avalon.py
```diff
@@ -224,18 +241,15 @@ def valid():
...
+            DeprecationWarning
```
eric-wieser:
```suggestion
            DeprecationWarning, stacklevel=2
```

cocotb/monitors/avalon.py
cmarqu: > TL;DR: `log.error` is for humans only, `raise` is for both humans and code.

This text should be part of a "How to write cocotb code", maybe in #1141.

Fatsie:
Squashed corrections.

cocotb/drivers/__init__.py
```diff
@@ -215,32 +214,29 @@ class BusDriver(Driver):
...
+        index = kwargs.get("array_idx", None)
 
         self.log = SimLog("cocotb.%s.%s" % (entity._name, name))
         Driver.__init__(self)
         self.entity = entity
         self.clock = clock
         self.bus = Bus(self.entity, name, self._signals,
-                       self._optional_signals, bus_separator=bus_separator, array_idx=index)
+                       optional_signals=self._optional_signals, **kwargs
+        )
```
eric-wieser: nit: this is some weird line wrapping. I'd use either:
```
x = foo(arg1,
        arg2)
```
or
```
x = foo(
    arg1,
    arg2
)

Comment (eric-wieser) [reply cocotb/drivers/__init__.py]: In fact, this is exactly what `flake8` is complaining about in CI:
```
./cocotb/drivers/__init__.py:239:9: E124 closing bracket does not match visual indentation
```

Comment (Fatsie) [reply cocotb/drivers/__init__.py]: Pushed change.

**Comment by Fatsie** [comment]
See #1194; DeprecationWarning not working for me.

**Comment by Fatsie** [comment]
Pushed following additional changes:
```diff
diff --git a/cocotb/monitors/avalon.py b/cocotb/monitors/avalon.py
index 1fa3645..405a43c 100644
--- a/cocotb/monitors/avalon.py
+++ b/cocotb/monitors/avalon.py
@@ -139,9 +139,6 @@ class AvalonSTPkts(BusMonitor):
         self.config["useEmpty"] = (num_data_symbols > 1)
 
         if hasattr(self.bus, 'channel'):
-            if "channel" in self._optional_signals:
-                self.log.warning("Channel is not fully implemented in this monitor. Recommend use of AvalonSTPktsWithChannel.")
-
             if len(self.bus.channel) > 128:
                 raise AttributeError("AvalonST interface specification defines channel width as 1-128. "
                                      "%d channel width is %d" %
```

cocotb/monitors/avalon.py
```diff
@@ -210,7 +221,10 @@ def valid():
                     self.log.info("Received a packet of %d bytes", len(pkt))
                     self.log.debug(hexdump(str((pkt))))
                     self.channel = channel
-                    self._recv(pkt)
+                    if channel is not None and self.report_channel:
```
eric-wieser:
```suggestion
                    if self.report_channel:
```
`channel` can never be `None` here anyway

imphil:
Thanks for this work @Fatsie! I leave the content review to @eric-wieser, but please fix the commit messages to follow the template:

```
summary line (up to 50 characters)

longer description, wrap at 72 characters
```

Read more about that in https://github.com/cocotb/cocotb/blob/master/CONTRIBUTING.md#patch-requirements

eric-wieser:
Commit message suggestions:

* add to the first one a comment explaining that the motivation is to keep the list of optional and required signals fixed
* Change the second one to something like
```
Use more explicit signatures the avalon driver

Replace `*args` in signatures with explicit argument names for clarity
```

Fatsie:
Rebased; handled requested changes.

> summary line (up to 50 characters)
> 
> longer description, wrap at 72 characters

OK with wrapping at 72 character but I do feel 50 characters is reducing usefulness of summary. 10 more would be much better.
I complied though.

Fatsie:
Looking at recent merges more than 50% seem to violate the 50 characters summary rule.

imphil:
Thanks for your update. This PR is not the best place to discuss agreed-on policy, so thanks for opening a new issue.

In your commit messages here, you can save characters by not trying to format commit message in rst; they are plaintext.

PR #157, opened 2014-09-09
For chris

SnookEE:
Adds AXI4Slave model into amba generator.
Adds Signed representation to binary.py. This includes signed-magnitude and two's-complement. 
Adds AD9361 generator.

PR #1191, opened 2019-11-18
Remove AD9361 driver.

Fatsie:
This driver will be broken by an upcoming patch so I propose to remove it unless a maintainer with right hardware is found.


eric-wieser:
> This driver will be broken by an upcoming patch

I'd dispute that - it can be fixed by simply removing `(BusDriver)`, which is a no-op anyway

Fatsie:
> Removal LGTM, but please also remove it from documentation/source/library_reference.rst.

Done

> I'd dispute that - it can be fixed by simply removing (BusDriver), which is a no-op anyway

Feel free to submit alternative patch. I'm not going to spend more time on AD9361.

eric-wieser:
> Feel free to submit alternative patch. I'm not going to spend more time on AD9361.

Done so in https://github.com/cocotb/cocotb/pull/1199

> If there's real need for that we can always bring that driver back in an extension module, or the probably single user can add it to their project.

I added @chiggs since they apparently requested this module in the first place in #157.

If they do add it to their project, they'll probably need the fix from #1199 if @Fatsie is going to make the changes I think they are to `BusDriver`. For that reason, it _might_ make sense to merge 1199 then merge this.

imphil:
I'll go ahead and merge this. Reasoning: We do not have tests for this driver, its not parts of the core cocotb functionality, and it's a rather odd device anyways. To make it easier to restore the functionality at a later time as part of an extension, I'll merge #1199 first. Thanks for preparing that PR @eric-wieser!

imphil:
Rebased and merged as e086d4b8803ce2e017885df4f1c22e51599b29fb

PR #1192, opened 2019-11-19
Fix Avalon Memory with byteenable using Python 3

bergerar:
Python 3 always returns a float for a normal division /. As range() does not accept a float value, the Avalon Memory can not be used bit bitenable using Python 3.

By using the floor division operator // the code works in Python 2 and 3.

Also see https://riptutorial.com/python/example/2797/integer-division


eric-wieser:
Obviously correct, thanks. Would be good to have tests for this code, but no need to do that in this PR.

PR #1193, opened 2019-11-19
Add missing "-top" parameter to VCS

ni140:
- add missing "-top" parameter to VCS
- replace spaces to tabs

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.vcs
+++ b/cocotb/share/makefiles/simulators/Makefile.vcs
@@ -38,17 +38,17 @@ else
 CMD_BIN := vcs
 
 ifdef VCS_BIN_DIR
-    CMD := $(shell which $(VCS_BIN_DIR)/$(CMD_BIN) 2>/dev/null)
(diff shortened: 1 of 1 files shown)


tymonx:
This is needed for projects that can have multiple compiled top modules. One common example is when project use Quartus simulation primitives. The Quartus lnsim file has dozens of unused modules 😬

ni140:
> Thanks!
> 
> For the next time:
> 
> * Please don't mix cleanups with actual changes. Whenever you are tempted to do a bulleted list in a commit message, you should do multiple commits instead. (They can be in one PR of course.)
> * In the commit message, describe why a change was done. It's obvious from the code change that you add the `-top` parameter, but why is this actually necessary? Again, in this case, it's reasonably obvious, but that's not always the case.

Sorry, first time doing PR on Github.

imphil:
No worries, your contribution is very much appreciated. Thanks for helping out!

eric-wieser:
> Please don't mix cleanups with actual changes.

It seems this cleanup was actually itself a bug: gh-1241

PR #1196, opened 2019-11-19
Store the version-number more simply

eric-wieser:
This:
* eliminates the `VERSION` environment variable
* adds `cocotb.__version__`
* changes `setup.py` and `conf.py` to read this version number

Fixes gh-1077, closes gh-1185


documentation/source/conf.py
```diff
@@ -69,10 +71,10 @@
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
-# The short X.Y version.
-version = '1.1'
 # The full version, including alpha/beta/rc tags.
-release = '1.1'
+release = cocotb.__version__
+# The short X.Y version.
+version = '.'.join(release.split('.')[:2])
```
imphil: Handwritten version parsing always makes me nervous. In this case, we get slightly inconsistent behavior for "1.2rc1" vs "1.2.0rc1". Looking through the code, it seems we could get rid of the version vs release distinction and always use the full "release"?

eric-wieser: > Handwritten version parsing always makes me nervous.

That's fair, but the cost of getting this wrong is a worst a weird string somewhere in the generated docs. Also, I think it would be rather foolish of us to change the number of components in our version.

> Looking through the code,

Which code? Sphinx itself?

---

I could switch this to:
```
from distutils.version import LooseVersion  # note: builtin
version = '.'.join(LooseVersion(release).version[:2]))
```

> and always use the full "release"

I'm ok.0 with this if you are.0 :)

cocotb/__init__.py
```diff
@@ -123,12 +125,8 @@ def _initialise_testbench(root_name):
     if exec_path is None:
         exec_path = 'Unknown'
 
-    version = os.getenv('VERSION')
-    if version is None:
-        log.info("Unable to determine Cocotb version from %s" % exec_path)
-    else:
-        log.info("Running tests with Cocotb v%s from %s" %
-                 (version, exec_path))
+    log.info("Running tests with Cocotb v%s from %s" %
```
cmarqu: We (well, I, through my documentation PRs) normalized the casing to an all-lowercase "cocotb",  this is one of the few remaining outliers - mind changing it here?

Issue #1194, opened 2019-11-19
warning with DeprecationWarning does not give output.


Fatsie:
Using ``DeprecationWarning`` seems to suppress the warning.
I have the following test:
```python
import cocotb
import warnings
from cocotb.triggers import Timer


@cocotb.test()
def deprecation_warning(dut):
    warnings.warn(
        "With DeprecationWarning",
        DeprecationWarning, stacklevel=2
    )
    warnings.warn(
        "Without DeprecationWarning",
        stacklevel=2
    )
    yield Timer(1)
```
Given the following output:
```console
[verhaegs@localhost test_deprecation]$ make
make results.xml
make[1]: Map '/home/verhaegs/eda/code/cocotb/tests/test_cases/test_deprecation' wordt binnengegaan
PYTHONPATH=/home/verhaegs/eda/code/cocotb/tests/test_cases/test_deprecation/build/libs/x86_64:/home/verhaegs/eda/code/cocotb/tests/test_cases/test_deprecation:/home/verhaegs/eda/code/cocotb: LD_LIBRARY_PATH=/home/verhaegs/eda/code/cocotb/tests/test_cases/test_deprecation/build/libs/x86_64:/home/verhaegs/software/lib:/home/verhaegs/software/lib64:/home/verhaegs/software/lib:/home/verhaegs/software/lib64::/usr/lib64:/usr/lib64:/usr/lib64:/usr/lib64 MODULE=test_deprecation \
        TESTCASE= TOPLEVEL=sample_module TOPLEVEL_LANG=verilog COCOTB_SIM=1 \
        /home/verhaegs/software/bin/vvp -M /home/verhaegs/eda/code/cocotb/tests/test_cases/test_deprecation/build/libs/x86_64 -m gpivpi sim_build/sim.vvp   
     -.--ns INFO     cocotb.gpi                                  gpi_embed.c:78   in set_program_name_in_venv        Did not detect Python virtual environment. Using system-wide Python interpreter
     -.--ns INFO     cocotb.gpi                                GpiCommon.cpp:91   in gpi_print_registered_impl       VPI registered
     0.00ns INFO     cocotb.gpi                                  gpi_embed.c:303  in embed_sim_init                  Running on Icarus Verilog version 11.0 (devel)
     0.00ns INFO     cocotb.gpi                                  gpi_embed.c:304  in embed_sim_init                  Python interpreter initialized and cocotb loaded!
[...]

PR #1201, opened 2019-11-20
Suggest what to use for deprecated attribute.

cmarqu:


cocotb/handle.py
```diff
@@ -139,7 +139,7 @@ def __setattr__(self, name, value):
     def __getattr__(self, name):
         if name in self._compat_mapping:
             if name not in _deprecation_warned:
-                warnings.warn("Use of %s attribute is deprecated" % name)
+                warnings.warn("Use of attribute '%s' is deprecated, use '%s' instead" % (name, self._compat_mapping[name]))
                 _deprecation_warned[name] = True
```
eric-wieser: Better to use `%r` instead of `'%s'`, it adds quotes for you

cmarqu: Done, thanks.

cocotb/handle.py
```diff
@@ -130,7 +130,7 @@ def __str__(self):
     def __setattr__(self, name, value):
         if name in self._compat_mapping:
             if name not in _deprecation_warned:
-                warnings.warn("Use of %s attribute is deprecated" % name)
+                warnings.warn("Use of attribute %r is deprecated, use %r instead" % (name, self._compat_mapping[name]))
                 _deprecation_warned[name] = True
```
eric-wieser: Out of scope for this patch, but this should really be
```
warnings.warn("...", DeprecationWarning, stacklevel=2)
```
(leaving python to deal with the "already-warned" state)

We'll want to fix #1194 before we make that change.

cmarqu:
We may even want to remove the "compatibility mapping" altogether, those attributes have been deprecated for quite a while now.

PR #1203, opened 2019-11-21
Fix error when ending test with assert in forked coorutine

ktbarrett:
A slight oversight when ending tests with `assert` was added. Currently, this follows the next exception path handle path which prints out an error.


eric-wieser:
In the long run I think we want asserts to take the same path as other exceptions, but for now this does improve the log message, so sure, why not.

PR #1204, opened 2019-11-21
Remove GPITrigger.__del__()

Fatsie:
Trigger.__del__() calls overloaded unprime() so duplicating
GPITrigger.unprime() functionality in GPITrigger.__del__()
is not necessary.


eric-wieser:
Nice cleanup, thanks

PR #1205, opened 2019-11-22
Doc improvements, fixes

cmarqu:
The fixed duplicated misrendering can be seen in e.g. 
https://external-builds.readthedocs.io/html/cocotb/1184/library_reference.html#cocotb.handle.RegionObject

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -495,7 +495,7 @@ class Join(_py_compat.with_metaclass(_ParameterizedSingletonAndABC, PythonTrigge
 
         @cocotb.coroutine()
         def coro_inner():
-            yield Timer(1)
+            yield Timer(1, units='ns')
             raise ReturnValue("Hello world")
 
         task = cocotb.fork(coro_inner())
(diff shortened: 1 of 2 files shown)


cocotb/handle.py
```diff
@@ -66,11 +70,6 @@ class SimHandleBase(object):
         }
 
     def __init__(self, handle, path):
-        """
-        Args:
-            handle (int): The GPI handle to the simulator object.
-            path (str): Path to this handle, ``None`` if root.
-        """
```
eric-wieser: Not a huge fan of moving docstrings away from the signature they describe, but maybe there's no way to make it render otherwise.

cmarqu: Maybe there is a way to use this docstring inheritance to reduce duplication, but for now I wanted to have the end result to not look bad.

eric-wieser: Yeah, the issue here is that the subclass repeats the args - removing it there would work too - I'll give that a try in a separate PR.

eric-wieser: Done in #1229.

cmarqu: Rebased to only include the rest of the fixes in this PR.

PR #1206, opened 2019-11-22
Allow (B|R)VALID at the same time as (AW|W|AR)READY

leastrobino:
- Write transaction: change the order of RisingEdge() and ReadOnly() to read BVALID in the same clock cycle as the last of AWREADY and WREADY
- Read transaction: change the order of RisingEdge() and ReadOnly() to read RVALID in the same clock cycle as ARREADY
- Use lock on read transaction

### DIFF
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -68,9 +68,7 @@ class AXI4LiteMaster(BusDriver):
 
     @cocotb.coroutine
     def _send_write_address(self, address, delay=0):
-        """
-        Send the write address, with optional delay (in clocks)
-        """
+        """Send the write address, with optional delay (in clocks)."""
         yield self.write_address_busy.acquire()
         for cycle in range(delay):
             yield RisingEdge(self.clock)
@@ -89,7 +87,7 @@ class AXI4LiteMaster(BusDriver):
 
     @cocotb.coroutine
     def _send_write_data(self, data, delay=0, byte_enable=0xF):
-        """Send the write address, with optional delay (in clocks)."""
+        """Send the write data, with optional delay (in clocks) and byte enable."""
         yield self.write_data_busy.acquire()
         for cycle in range(delay):
             yield RisingEdge(self.clock)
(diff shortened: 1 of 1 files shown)


imphil:
Thanks for your patch! 

I'm not an AXI expert, but nobody voiced objection and the commit looks fine, so there's no reason to keep this open any longer.

cocotb/drivers/amba.py
```diff
@@ -146,11 +144,11 @@ def write(self, address, value, byte_enable=0xf, address_latency=0,
 
         # Wait for the response
         while True:
-            yield ReadOnly()
             if self.bus.BVALID.value and self.bus.BREADY.value:
                 result = self.bus.BRESP.value
                 break
             yield RisingEdge(self.clock)
+            yield ReadOnly()
```
eric-wieser: This change doesn't look right to me - we shouldn't be looking at signals until the `ReadOnly` phase

eric-wieser:
@leastrobino, can you elaborate on the reasoning behind the first two bullet points? It's my understanding that yielding `ReadOnly` while already in a readonly phase does not advance time, so I'm not sure why adding or removing it would affect which clock cycle a read happens in.

What simulator were you using when you had this trouble? It's also possible that the behavior of these triggers varies by simulator.

leastrobino:
@eric-wieser, I proposed those changes because some IPs I have in my design behave like that, BVALID or RVALID is set at the same time as the other channels READY signals.

Although accepted by ARM processors and/or Xilinx AXI Interconnect, it seems this behavior is slightly off-standard according to AMBA AXI and ACE Protocol Specification, A3.3.1:
- Read transaction dependencies:
  The slave must wait for both ARVALID and ARREADY to be asserted before it asserts RVALID to indicate that valid data is available.
- AXI4 and AXI5 write transaction dependencies:
  The slave must wait for AWVALID, AWREADY, WVALID, and WREADY to be asserted before asserting BVALID.

After having read more thoroughly the standard, I propose to revert these changes and keep only the changes on read transaction lock (pull request #1283).

eric-wieser:
Thanks for the analysis - I think I understand your intent better now, but I don't think this patch correctly implements it. We'll get this reverted before the 1.3 release.

PR #1207, opened 2019-11-22
Monitor: transaction log helper

Fatsie:
A Monitor method that can be used to easily log a transaction.

### DIFF
--- a/cocotb/monitors/__init__.py
+++ b/cocotb/monitors/__init__.py
@@ -113,15 +113,6 @@ class Monitor(object):
                        callback.__name__)
         self._callbacks.append(callback)
 
-    def log_transaction(self, transaction):
-        """Helper to easily log a received transaction.
-
-        Typical usage: ``moni_obj.add_callback(moni_obj.log_transaction)``
-        
-        .. versionadded:: 1.3
-        """
-        self.log.info("Transaction received: {!r}".format(transaction))
-
     @coroutine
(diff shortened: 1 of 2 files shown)


cocotb/monitors/__init__.py
```diff
@@ -113,6 +113,13 @@ def add_callback(self, callback):
                        callback.__name__)
         self._callbacks.append(callback)
 
+    def log_transaction(self, transaction):
+        """Helper to easily log a received transaction
```
cmarqu:
```suggestion
        """Helper to easily log a received transaction.
```

cocotb/monitors/__init__.py
```diff
@@ -113,6 +113,13 @@ def add_callback(self, callback):
...
+
+        Typical usage: :any:`moni_obj.add_callback(moni_obj.log_transaction)`
```
cmarqu:
```suggestion
        Typical usage: ``moni_obj.add_callback(moni_obj.log_transaction)``
```
(since you don't want a hyperlink)

Fatsie:
Applied and squashed suggested changes.

eric-wieser:
I'm not convinced this helper is helpful enough to justify bring part of the monitor API, vs the user just doing:
```
@m.add_callback
def _(t):
    m.log.info("got {}".format(t))
```

Fatsie:
Yeah, added another possiblity in the patch as example using a lambda function.

PR #1209, opened 2019-11-23
Convenience triggers

Fatsie:
Avoid construction/destruction of Triggers without parameters by providing objects for the classes.

Some possible discussion:
* The second patch is need to avoid having exceptions during exit of a script. It seems that classes could have been set to ``None`` before the objects ``__del__()`` method is called. This leads to following output:
```console
[verhaegs@localhost test_avalon_stream]$ make
make results.xml
make[1]: Map '/home/verhaegs/eda/code/cocotb/tests/test_cases/test_avalon_stream' wordt binnengegaan
PYTHONPATH=/home/verhaegs/eda/code/cocotb/tests/test_cases/test_avalon_stream/build/libs/x86_64:/home/verhaegs/eda/code/cocotb/tests/test_cases/test_avalon_stream:/home/verhaegs/eda/code/cocotb: LD_LIBRARY_PATH=/home/verhaegs/eda/code/cocotb/tests/test_cases/test_avalon_stream/build/libs/x86_64:/home/verhaegs/software/lib:/home/verhaegs/software/lib64:/home/verhaegs/software/lib:/home/verhaegs/software/lib64::/usr/lib64:/usr/lib64:/usr/lib64:/usr/lib64 MODULE=test_avalon_stream \
        TESTCASE= TOPLEVEL=avalon_streaming TOPLEVEL_LANG=verilog COCOTB_SIM=1 \
        /home/verhaegs/software/bin/vvp -M /home/verhaegs/eda/code/cocotb/tests/test_cases/test_avalon_stream/build/libs/x86_64 -m gpivpi sim_build/sim.vvp   
     -.--ns INFO     cocotb.gpi                                  gpi_embed.c:78   in set_program_name_in_venv        Did not detect Python virtual environment. Using system-wide Python interpreter
     -.--ns INFO     cocotb.gpi                                GpiCommon.cpp:104  in gpi_print_registered_impl       VPI registered
     0.00ns INFO     cocotb.gpi                                  gpi_embed.c:347  in embed_sim_init                  Running on Icarus Verilog version 11.0 (devel)
     0.00ns INFO     cocotb.gpi                                  gpi_embed.c:348  in embed_sim_init                  Python interpreter initialized and cocotb loaded!
     0.00ns INFO     cocotb                                      __init__.py:131  in _initialise_testbench           Running tests with cocotb v1.2.0 from /home/verhaegs/eda/code/cocotb
     0.00ns INFO     cocotb                                      __init__.py:148  in _initialise_testbench           Seeding Python random module with 1574516959
     0.00ns INFO     cocotb.regression                         regression.py:190  in initialise                      Found test test_avalon_stream.test_avalon_stream
     0.00ns INFO     cocotb.regression                         regression.py:348  in execute                         Running test 1/1: test_avalon_stream
     0.00ns INFO     ..test_avalon_stream.0x7f5423e66990       decorators.py:253  in _advance                        Starting test: "test_avalon_stream"
                                                                                                                     Description: Test stream of avalon data
     0.00ns INFO     cocotb.scoreboard.avalon_streaming        scoreboard.py:216  in add_interface                   Created with reorder_depth 0
VCD info: dumpfile waveform.vcd opened for output.
VCD warning: $dumpvars: Unsupported argument type (vpiPackage)
681000000000.00ns INFO     cocotb.regression                         regression.py:288  in handle_result                   Test Passed: test_avalon_stream
681000000000.00ns INFO     cocotb.regression                         regression.py:213  in tear_down                       Passed 1 tests (0 skipped)
681000000000.00ns INFO     cocotb.regression                         regression.py:400  in _log_test_summary               ***********************************************************************************************
                                                                                                                           ** TEST                                   PASS/FAIL  SIM TIME(NS)  REAL TIME(S)  RATIO(NS/S) **
                                                                                                                           ***********************************************************************************************
                                                                                                                           ** test_avalon_stream.test_avalon_stream    PASS     681000000000.00          0.11   6359732066279.84  **
                                                                                                                           ***********************************************************************************************
                                                                                                                           
681000000000.00ns INFO     cocotb.regression                         regression.py:417  in _log_sim_summary                *************************************************************************************
                                                                                                                           **                                 ERRORS : 0                                      **
                                                                                                                           *************************************************************************************
                                                                                                                           **                               SIM TIME : 681000000000.00 NS                     **
                                                                                                                           **                              REAL TIME : 0.19 S                                 **
                                                                                                                           **                        SIM / REAL TIME : 3494026234056.32 NS/S                  **
                                                                                                                           *************************************************************************************
                                                                                                                           
681000000000.00ns INFO     cocotb.regression                         regression.py:222  in tear_down                       Shutting down...
Exception AttributeError: "'NoneType' object has no attribute 'unprime'" in <bound method ReadWrite.__del__ of <cocotb.triggers.ReadWrite object at 0x7f54241cda00>> ignored
Exception AttributeError: "'NoneType' object has no attribute 'unprime'" in <bound method ReadOnly.__del__ of <cocotb.triggers.ReadOnly object at 0x7f54241cd9b0>> ignored
Exception AttributeError: "'NoneType' object has no attribute 'unprime'" in <bound method NextTimeStep.__del__ of <cocotb.triggers.NextTimeStep object at 0x7f54241cda50>> ignored
make[1]: Map '/home/verhaegs/eda/code/cocotb/tests/test_cases/test_avalon_stream' wordt verlaten
```
* I provided a third patch to provide a possible back-wards compatibiliy problem for user code that would use Scheduler._read_only. Is user code actually supposed to use attributes starting with a `_`.

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -152,10 +152,14 @@ class GPITrigger(Trigger):
 
     def unprime(self):
         """Disable a primed trigger, can be re-primed."""
-        if self.cbhdl != 0:
+        # Due to race condition in object global object we rely on may have been
+        # deleted. This will happen only right before exit so just skip over the
+        # action.
+        if self.cbhdl != 0 and simulator is not None:
             simulator.deregister_callback(self.cbhdl)
         self.cbhdl = 0
-        Trigger.unprime(self)
+        if Trigger is not None:
+            Trigger.unprime(self)
 
 
 class Timer(GPITrigger):
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -729,3 +729,10 @@ class ClockCycles(Waitable):
         for _ in range(self.num_cycles):
             yield trigger
         raise ReturnValue(self)
+
+
+# Global objects to allow using Triggers that don't have parameters to be used without
+# constructing/construction for each use.
+readonly = ReadOnly()
+readwrite = ReadWrite()
+nexttimestep = NextTimeStep()
--- a/cocotb/drivers/__init__.py
+++ b/cocotb/drivers/__init__.py
@@ -33,7 +33,7 @@ from collections import deque
 
 import cocotb
 from cocotb.decorators import coroutine
-from cocotb.triggers import (Event, RisingEdge, ReadOnly, NextTimeStep,
+from cocotb.triggers import (Event, RisingEdge, readonly, nexttimestep,
                              Edge)
 from cocotb.bus import Bus
 from cocotb.log import SimLog
@@ -262,11 +262,11 @@ class BusDriver(Driver):
         to move to :class:`~cocotb.triggers.NextTimeStep` before
         registering more callbacks can occur.
         """
-        yield ReadOnly()
+        yield readonly
         while signal.value.integer != 1:
             yield RisingEdge(signal)
-            yield ReadOnly()
-        yield NextTimeStep()
+            yield readonly
+        yield nexttimestep
 
     @coroutine
     def _wait_for_nsignal(self, signal):
@@ -276,11 +276,11 @@ class BusDriver(Driver):
         to move to :class:`~cocotb.triggers.NextTimeStep` before
         registering more callbacks can occur.
         """
-        yield ReadOnly()
+        yield readonly
         while signal.value.integer != 0:
             yield Edge(signal)
-            yield ReadOnly()
-        yield NextTimeStep()
+            yield readonly
+        yield nexttimestep
 
     def __str__(self):
         """Provide the name of the bus"""
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -27,7 +27,7 @@
 """Drivers for Advanced Microcontroller Bus Architecture."""
 
 import cocotb
-from cocotb.triggers import RisingEdge, ReadOnly, Lock
+from cocotb.triggers import RisingEdge, readonly, Lock
 from cocotb.drivers import BusDriver
 from cocotb.result import ReturnValue
 from cocotb.binary import BinaryValue
@@ -79,7 +79,7 @@ class AXI4LiteMaster(BusDriver):
         self.bus.AWVALID <= 1
 
         while True:
-            yield ReadOnly()
+            yield readonly
             if self.bus.AWREADY.value:
                 break
             yield RisingEdge(self.clock)
@@ -99,7 +99,7 @@ class AXI4LiteMaster(BusDriver):
         self.bus.WSTRB <= byte_enable
 
         while True:
-            yield ReadOnly()
+            yield readonly
             if self.bus.WREADY.value:
                 break
             yield RisingEdge(self.clock)
@@ -146,7 +146,7 @@ class AXI4LiteMaster(BusDriver):
 
         # Wait for the response
         while True:
-            yield ReadOnly()
+            yield readonly
             if self.bus.BVALID.value and self.bus.BREADY.value:
                 result = self.bus.BRESP.value
                 break
@@ -182,7 +182,7 @@ class AXI4LiteMaster(BusDriver):
         self.bus.ARVALID <= 1
 
         while True:
-            yield ReadOnly()
+            yield readonly
             if self.bus.ARREADY.value:
                 break
             yield RisingEdge(self.clock)
@@ -191,7 +191,7 @@ class AXI4LiteMaster(BusDriver):
         self.bus.ARVALID <= 0
 
         while True:
-            yield ReadOnly()
+            yield readonly
             if self.bus.RVALID.value and self.bus.RREADY.value:
                 data = self.bus.RDATA.value
                 result = self.bus.RRESP.value
@@ -268,13 +268,13 @@ class AXI4Slave(BusDriver):
         while True:
             while True:
                 self.bus.WREADY <= 0
-                yield ReadOnly()
+                yield readonly
                 if self.bus.AWVALID.value:
                     self.bus.WREADY <= 1
                     break
                 yield clock_re
 
-            yield ReadOnly()
+            yield readonly
             _awaddr = int(self.bus.AWADDR)
             _awlen = int(self.bus.AWLEN)
             _awsize = int(self.bus.AWSIZE)
@@ -316,12 +316,12 @@ class AXI4Slave(BusDriver):
 
         while True:
             while True:
-                yield ReadOnly()
+                yield readonly
                 if self.bus.ARVALID.value:
                     break
                 yield clock_re
 
-            yield ReadOnly()
+            yield readonly
             _araddr = int(self.bus.ARADDR)
             _arlen = int(self.bus.ARLEN)
             _arsize = int(self.bus.ARSIZE)
@@ -348,7 +348,7 @@ class AXI4Slave(BusDriver):
 
             while True:
                 self.bus.RVALID <= 1
-                yield ReadOnly()
+                yield readonly
                 if self.bus.RREADY.value:
                     _burst_diff = burst_length - burst_count
                     _st = _araddr + (_burst_diff * bytes_in_beat)
--- a/cocotb/drivers/avalon.py
+++ b/cocotb/drivers/avalon.py
@@ -36,7 +36,7 @@ import random
 
 import cocotb
 from cocotb.decorators import coroutine
-from cocotb.triggers import RisingEdge, FallingEdge, ReadOnly, NextTimeStep, Event
+from cocotb.triggers import RisingEdge, FallingEdge, readonly, nexttimestep, Event
 from cocotb.drivers import BusDriver, ValidatedBusDriver
 from cocotb.utils import hexdump
 from cocotb.binary import BinaryValue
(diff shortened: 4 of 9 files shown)


eric-wieser:
I don't really think this is necessary - `ReadOnly()`  already returns an existing object rather than creating a new one.

cocotb/triggers.py
```diff
@@ -152,10 +152,14 @@ def __init__(self):
...
-        if self.cbhdl != 0:
+        # Due to race condition in object global object we rely on may have been
+        # deleted. This will happen only right before exit so just skip over the
+        # action.
+        if self.cbhdl != 0 and simulator is not None:
             simulator.deregister_callback(self.cbhdl)
         self.cbhdl = 0
-        Trigger.unprime(self)
+        if Trigger is not None:
+            Trigger.unprime(self)
```
eric-wieser: Can this really ever happen?

eric-wieser: (it being none)

eric-wieser:
> Is user code actually supposed to use attributes starting with a _.

No, it is not.

cocotb/triggers.py
Fatsie: Yes, would not have implemented this check; the exceptions at the end of the output given in my first comment are caused by Trigger being None.
Took some time to understand myself how this could happen. It seems that when python cleans up a module at exit he may set class globals to None (here Trigger) before calling __del__ on objects in that module.

Fatsie:
> I don't really think this is necessary - ReadOnly() already returns an existing object rather than creating a new one.

Checked and confirmed; RisingEdge(dut.clk) will also return same object when given same signal. Will make alternative patch with removing the premature optimizations from the Avalon code.

cmarqu:
If this is indeed the same in terms of speed/byte code, then we can clean up more such places - I'm sure I have seen them also outside of Avalon.

eric-wieser:
Without profiling it's difficult to say if the optimizations are premature

Fatsie:
Did a little benchmarking on my old Core2 Quad Q8200.
I used the following test with the sample test design:
```python
import time

import cocotb
from cocotb.triggers import Timer, ReadOnly, RisingEdge, readonly


@cocotb.coroutine
def loop1(clk, n):
    for _ in range(n):
        yield ReadOnly()
        yield RisingEdge(clk)

@cocotb.coroutine
def loop2(clk, n):
    ro = ReadOnly()
    edge = RisingEdge(clk)
    for _ in range(n):
        yield ro
        yield edge

@cocotb.coroutine
def loop3(event, n):
    for _ in range(n):
        yield readonly
        yield event

@cocotb.test()
def test_bed(dut):
    event = RisingEdge(dut.clk)

    cocotb.fork(Clock(dut.clk, 10, 'ns').start())

    for n in (100000):
        print("{} iterations".format(n))
        
        start = time.time()
        yield loop1(dut.clk, n)
        end = time.time()
        print("\tloop1: {:.2f}".format(end - start))

        start = time.time()
        yield loop2(dut.clk, n)
        end = time.time()
        print("\tloop2: {:.2f}".format(end - start))

        start = time.time()
        yield loop3(event, n)
        end = time.time()
        print("\tloop3: {:.2f}".format(end - start))
```
Used a few simulators I had installed.

| lang | python | sim | loop1 | loop2 | loop3 |
|-|-|-|-|-|-|
| Verilog | 2.7.5 | icarus | 50.20 | 48.96 | 48.96 |
| Verilog | 3.7.4 | icarus | 34.45 | 33.20 | 33.13 |
| VHDL | 2.7.5 | ghdl | 51.30 | 50.10 | 50.26 |
| VHDL | 3.7.4 | ghdl | 35.11 | 33.66 | 33.80 |
| Verilog | 3.7.1 (32 bit) | modelsim (32 bit) | 100.06 | 97.58 | 97.31 |

The modelsim from ISE 17.1 WebPack does not seem to work with VHDL as top level language. Don't know if it is supposed to.
When looking at the disassembly of the loops it can be seen that loop2 and loop3 avoid a function call in the loop.
There is some advantage in speed and I don't consider this patch ugly so I would propose to continue with it.

eric-wieser:
It's possible we can get a similar improvement by speeding up `type(ReadOnly).__call__`. Right now it performs a dictionary lookup, but it could return a variable directly.

I'd rather not introduce two ways to spell the same thing unless the speedup is significant in real cases.

Your benchmark is unfair, because you're throwing in RisingEdge improvements (which aren't part of this patch) with ReadOnly improvements. Can you repeat it focusing on only the improvement that would be offered by this patch?

Fatsie:
Update test case with some possible improvements added:
```python
import time

import cocotb
from cocotb.triggers import Timer, ReadOnly, RisingEdge, readonly, Waitable

_ro = ReadOnly()

class ReadOnlyTest(Waitable):
    @cocotb.coroutine
    def _wait(self):
        yield _ro

def ReadOnlyTest2():
    return _ro

@cocotb.coroutine
def loop1(event, n):
    for _ in range(n):
        yield ReadOnly()
        yield event

@cocotb.coroutine
def loop2(event, n):
    ro = ReadOnly()
    for _ in range(n):
        yield ro
        yield event

@cocotb.coroutine
def loop3(event, n):
    for _ in range(n):
        yield readonly
        yield event

@cocotb.coroutine
def loop4(event, n):
    for _ in range(n):
        yield ReadOnlyTest()
        yield event

@cocotb.coroutine
def loop5(event, n):
    for _ in range(n):
        yield ReadOnlyTest2()
        yield event

@cocotb.test()
def test_bed(dut):
    event = RisingEdge(dut.clk)

    cocotb.fork(Clock(dut.clk, 10, 'ns').start())

    for n in (100000,):
        print("{} iterations".format(n))
        
        yield loop1(event, 100)
        start = time.time()
        yield loop1(event, n)
        end = time.time()
        print("\tloop1: {:.2f}".format(end - start))

        yield loop2(event, 100)
        start = time.time()
        yield loop2(event, n)
        end = time.time()
        print("\tloop2: {:.2f}".format(end - start))

        yield loop3(event, 100)
        start = time.time()
        yield loop3(event, n)
        end = time.time()
        print("\tloop3: {:.2f}".format(end - start))

        yield loop4(event, 100)
        start = time.time()
        yield loop4(event, n)
        end = time.time()
        print("\tloop4: {:.2f}".format(end - start))

        yield loop5(event, 100)
        start = time.time()
        yield loop5(event, n)
        end = time.time()
        print("\tloop5: {:.2f}".format(end - start))
```
Results for icarus only:
| lang | python | sim | loop1 | loop2 | loop3 | loop4 | loop5 |
|-|-|-|-|-|-|-|-|
| Verilog | 2.7.5 | icarus | 48.55 | 48.11 | 48.10 | 62.79 | 48.14 |
| Verilog | 3.7.4 | icarus | 33.00 | 32.41 | 32.40 | 42.31 | 32.42 |

It seems indeed possible to use something like ReadOnlyTest2. The Waitable case seems to be much slower and may also  be candidate for optimization but I leave that to other people.
Personally I want to remove the optimizations from the monitors/drivers so either we live with the performance hit or we do optimization like ReadOnlyTest2.

eric-wieser:
Note that the `ReadOnlyFunction` approach is what we used to have, but it failed a test like `isinstance(ReadOnly(), ReadOnly)`.

Changing `type(ReadOnly).__call__` would probably resolve this. The relevant code is:

https://github.com/cocotb/cocotb/blob/17ce4e51da43b6d2d1b5834d629374ed5535a8d8/cocotb/utils.py#L413-L443

Introducing a `Singleton` metaclass for non-parametrized singletons would give a small performance boost - which would be very similar, but replace the `WeakValueDictionary` with a `weakref.ref`, and would be without the `__singleton_key__`

It's worth noting the best improvement here is only 1% - which IMO doesn't fall in the category of optimizations worth introducing a new way to spell something. To quote the zen of python

eric-wieser:
That new metaclass would look like:
```
class Singleton(type):
    """A metaclass that allows class construction to reuse an existing instance.

    We use this so that :class:`ReadOnly() <cocotb.triggers.RisingEdge>` always
    returns the same instance, rather than creating new copies.

    An optimized version of :class:`ParametrizedSingleton`, to avoid a lookup
    of no parameters - which gives a 1% speedup.
    """

    def __init__(cls, *args, **kwargs):
        # Weak such that if the instance is no longer referenced, it can be
        # collected.
        cls.__instance_ref = lambda: None

    def __call__(cls, *args, **kwargs):
        self = cls.__instance_ref()
        if self is None:
            # construct the object as normal
            self = super(ParametrizedSingleton, cls).__call__(*args, **kwargs)
            cls.__instance_ref = weakref.ref(self)
        return self
```

You could probably drop the weakref if it helps performance.

cmarqu:
> ```python
> We use this so that :class:`ReadOnly() <cocotb.triggers.RisingEdge>`
> ```

Inconsistency ``ReadOnly`` vs. ``RisingEdge``.

Fatsie:
weakref does not seem to work as it will delete the object anyway and return None when accessing the ref. I'm not sure why i does not happen for ``ParametrizedSingleton`` and ``WeakValueDictionary``.
Will use without weakref.

Fatsie:
> I'm not sure why it does not happen for ParametrizedSingleton and WeakValueDictionary.

I do think I get it. The key will also be removed of the dict so the object will be regenerated without error.

eric-wieser:
The object won't be deleted as long as something holds onto it (eg, the global simulator instance's cache of the variables).

Fatsie:
> The object won't be deleted as long as something holds onto it (eg, the global simulator instance's cache of the variables).

Seems python can be quite aggressive with garbage collecting objects, see #1211

Fatsie:
Closing this patch after new PRs
* #1216: Singleton proposal
* #1215: Clean up and related issue #1211 on performance of cached ``RisingEdge(clk)`` etc.

Issue #1212, opened 2019-11-26
Why is the default logger overridden?


ktbarrett:
https://github.com/cocotb/cocotb/blob/ba0fb097efbd2939ad6a961ee8d02793104ef8f7/cocotb/__init__.py#L60

How about instead of forcing users to use the cocotb logger through `logging`, we just extend `cocotb.log` to act more like the logging module and get users to use that instead?

eric-wieser:
This is already a lot simpler than it used to be.

If you look at the class, it doesn't really do very much. The most important thing is the _logFromC method.

I suppose that could be a free function, there's not a particularly strong argument for it to be a method.

ktbarrett:
That's not really what I'm asking. I'm asking why that quoted line is there. Is there any reason to set the logger class in `logging`? I see in a lot of files there are lines like

```python
import logging
testlog = logging.getLogger('cocotb.test')
```

Why override `logging` when you could instead just set up `cocotb.log` to act just like the `logging` module, so that you could do

```python
import cocotb.log
cocotb.log.getLogger('test')
```

Why replace the world, instead of coexist? Sometimes people want to use the `logging` module for things that aren't a simulation status/debugging logfile.

eric-wieser:
> Is there any reason to set the logger class in logging?

Here's why we do it today:

* To set a default logging formatter / handler
* To provide a hook to allow our C code to call into the python logging stuff
* To add some random color attributes

The first two of these would be possible without setting the logging class. The third isn't, but arguably isn't very useful.

stuarthodgson:
The last point is quite useful for me at least.

eric-wieser:
@stuarthodgson: How would you feel about deprecating `log.colour` in favor of just calling `cocotb.utils.want_color_output()` directly?

eric-wieser:
@ktbarrett, back to your original question:

> How about instead of forcing users to use the cocotb logger

Are you complaining that users can't now specify their own handler class?  There's nothing to stop a user doing:
```
@logging.setLoggerClass
class MyLog(logging.getLoggerClass()):
   pass  # this will work both with and without cocotb
```

Or are you complaining about the fact we set a default handler / formatter for all loggers?

ktbarrett:
> Or are you complaining about the fact we set a default handler / formatter for all loggers?

[...]

PR #1213, opened 2019-11-26
Use a free function for `_log_from_c`

eric-wieser:
As raised in gh-1212, there is no need to use subclassing here, and doing so removes flexibility from the user.

After this change, `libgpi` logging will work independently of whatever logging class is configured via the builtin `logging.setLoggerClass`. More importantly, it removes one more reason for `SimBaseLog` to exist - the ultimate goal is to remove that class entirely, since it's behavior is hostile to users wanting to use the logging module normally.

This also fixes a bug where the name argument of `gpilog` is ignored.

### DIFF
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -259,37 +259,32 @@ int embed_sim_init(gpi_sim_info_t *info)
     }
 
     PyObject *cocotb_module, *cocotb_init, *cocotb_retval;
-    PyObject *simlog_obj, *simlog_func;
+    PyObject *cocotb_log_module = NULL;
+    PyObject *simlog_func;
     PyObject *argv_list;
 
     cocotb_module = NULL;
-    simlog_obj = NULL;
 
     // Ensure that the current thread is ready to call the Python C API
     PyGILState_STATE gstate = PyGILState_Ensure();
     to_python();
 
-    if (get_module_ref(COCOTB_MODULE, &cocotb_module))
+    if (get_module_ref("cocotb", &cocotb_module))
         goto cleanup;
 
-    // Obtain the loggpi logger object
-    simlog_obj = PyObject_GetAttrString(cocotb_module, "loggpi");       // New reference
-
-    if (simlog_obj == NULL) {
-        PyErr_Print();
-        LOG_ERROR("Failed to get simlog object from loggpi\n");
+    if (get_module_ref("cocotb.log", &cocotb_log_module)) {
         goto cleanup;
     }
 
     // Obtain the function to use when logging from C code
-    simlog_func = PyObject_GetAttrString(simlog_obj, "_logFromC");      // New reference
+    simlog_func = PyObject_GetAttrString(cocotb_log_module, "_log_from_c");      // New reference
     if (simlog_func == NULL) {
         PyErr_Print();
-        LOG_ERROR("Failed to get the _logFromC method");
+        LOG_ERROR("Failed to get the _log_from_c function");
         goto cleanup;
     }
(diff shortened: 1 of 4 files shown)


eric-wieser:
I'll improve the commit message when I merge, if that works for you

eric-wieser:
Sorry @imphil, ended up making some more substantial changes after having thought about this some more. Wasn't really a fan of putting these functions in `__init__`

ktbarrett:
I preferred the previous implementation. When we add #1225, we will need the whole interface required by C to be defined in the entry module.

eric-wieser:
@ktbarrett: perhaps I'm missing something, but I don't think there is a good reason to allow the user to configure gpilog to do anything other than hook up to the python logging infrastructure, so that needn't be party of a custom entry point.

eric-wieser:
I could put them in `cocotb._gpilog` if that would make more sense to you.

ktbarrett:
When I say custom entry point I mean a modular interface. You could wipe away the cocotb python module and put something totally different in place, and it will work. So saying you have a custom entry point, but will always use `cocotb.log` for the gpilog is floating in a weird halfway between a proper modular layered design and the monolith we have now.

And that's why I want those functions at known location in the entry module. Not doing so prevents people from taking advantage of a (future) modular design and offers us nothing.

eric-wieser:
I assume that all you object to in the second revision of this patch is the file which defines these functions.

I would argue then that until we have a clear design for this modularity (perhaps a wiki page), we should stick to keeping all logging-related python code in a single file. If the functions are private, as they are here, we can move them later without concern for compatibility.

Making these functions easy for a future maintainer to find is more important than optimizing for a design we don't have yet.

eric-wieser:
Or perhaps a simpler argument: these functions have not moved file - moving file is out of scope for this patch, let's move them in a subsequent one with suitable motivation in the commit message.

ktbarrett:
I have no problem it being tomorrow's problem. Just seeing if it was worth the effort today.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -259,52 +259,47 @@ int embed_sim_init(gpi_sim_info_t *info)
...
 
     cocotb_module = NULL;
-    simlog_obj = NULL;
 
     // Ensure that the current thread is ready to call the Python C API
     PyGILState_STATE gstate = PyGILState_Ensure();
     to_python();
 
-    if (get_module_ref(COCOTB_MODULE, &cocotb_module))
+    if (get_module_ref("cocotb", &cocotb_module))
```
imphil: What's the motivation for this change? For consistency with `cocotb.log` below? If so, please remove the definition of COCOTB_MODULE.

cocotb/share/lib/gpi_log/gpi_logging.c
```diff
@@ -207,7 +212,7 @@ void gpi_log(const char *name, long level, const char *pathname, const char *fun
     }
 
     // Log function args are level, filename, lineno, msg, function
```
imphil: Please adjust this comment for the added name argument.

cocotb/share/lib/embed/gpi_embed.c
eric-wieser: Seemed like a fairly useless constant given its defined as a shorter string than its name. I'll come back and remove the env bar.

cocotb/share/lib/gpi_log/gpi_logging.c
eric-wieser: Will do

eric-wieser:
Comments addressed, will merge when CI passes

PR #1214, opened 2019-11-27
Add StableCondition and StableValue trigger.

Fatsie:
These triggers allow to wait until a certain condition is met or a signal has a given value when in ReadOnly state. The event after which to do the check is configurable.

Some remarks:
* I'm open for naming suggestions. Alternatives I considered ConditionReadOnly/ValueReadOnly or ConditionInReadOnly/ValueInReadOnly.
* I deprecated BusDriver._wait_for_(n)signal() as I don't see a reason why these are methods. They don't use the object.
* For testing I used the Avalon tests. The amba changes have been done blindly. Would be good if a user could test this, or even better submit a test.

### DIFF
--- a/cocotb/drivers/__init__.py
+++ b/cocotb/drivers/__init__.py
@@ -30,11 +30,11 @@
 """Set of common driver base classes."""
 
 from collections import deque
+import warnings
 
 import cocotb
 from cocotb.decorators import coroutine
-from cocotb.triggers import (Event, RisingEdge, ReadOnly, NextTimeStep,
-                             Edge)
+from cocotb.triggers import Event, RisingEdge, NextTimeStep, StableValue
 from cocotb.bus import Bus
 from cocotb.log import SimLog
 
@@ -262,10 +262,12 @@ class BusDriver(Driver):
         to move to :class:`~cocotb.triggers.NextTimeStep` before
         registering more callbacks can occur.
         """
-        yield ReadOnly()
-        while signal.value.integer != 1:
-            yield RisingEdge(signal)
-            yield ReadOnly()
+        warnings.warn(
+            "Use of BusDriver._wait_for_signal() is deprecated\n"
+            "\tPlease, use cocotb.trigggers.StableValue instead",
+            DeprecationWarning, stacklevel=2
+        )
+        yield StableValue(signal, 1)
         yield NextTimeStep()
 
     @coroutine
@@ -276,10 +278,12 @@ class BusDriver(Driver):
         to move to :class:`~cocotb.triggers.NextTimeStep` before
         registering more callbacks can occur.
         """
-        yield ReadOnly()
-        while signal.value.integer != 0:
-            yield Edge(signal)
-            yield ReadOnly()
+        warnings.warn(
+            "Use of BusDriver._wait_for_signal() is deprecated\n"
+            "\tPlease, use cocotb.triggers.StableValue instead",
+            DeprecationWarning, stacklevel=2
+        )
+        yield StableValue(signal, 0)
         yield NextTimeStep()
 
     def __str__(self):
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -27,7 +27,7 @@
 """Drivers for Advanced Microcontroller Bus Architecture."""
 
 import cocotb
-from cocotb.triggers import RisingEdge, ReadOnly, Lock
+from cocotb.triggers import RisingEdge, ReadOnly, Lock, StableCondition, StableValue
 from cocotb.drivers import BusDriver
 from cocotb.result import ReturnValue
 from cocotb.binary import BinaryValue
@@ -78,12 +78,8 @@ class AXI4LiteMaster(BusDriver):
         self.bus.AWADDR <= address
         self.bus.AWVALID <= 1
 
-        while True:
-            yield ReadOnly()
-            if self.bus.AWREADY.value:
-                break
-            yield RisingEdge(self.clock)
-        yield RisingEdge(self.clock)
+        yield StableValue(self.bus.AWREADY, 1, event=RisingEdge(self.clock),
+                          extra_event=True)
         self.bus.AWVALID <= 0
         self.write_address_busy.release()
 
@@ -98,12 +94,8 @@ class AXI4LiteMaster(BusDriver):
         self.bus.WVALID <= 1
         self.bus.WSTRB <= byte_enable
 
-        while True:
-            yield ReadOnly()
-            if self.bus.WREADY.value:
-                break
-            yield RisingEdge(self.clock)
-        yield RisingEdge(self.clock)
+        yield StableValue(self.bus.WREADY, 1, event=RisingEdge(self.clock),
+                          extra_event=True)
         self.bus.WVALID <= 0
         self.write_data_busy.release()
 
@@ -145,19 +137,14 @@ class AXI4LiteMaster(BusDriver):
             yield c_data.join()
 
         # Wait for the response
-        while True:
-            yield ReadOnly()
-            if self.bus.BVALID.value and self.bus.BREADY.value:
-                result = self.bus.BRESP.value
-                break
-            yield RisingEdge(self.clock)
-
-        yield RisingEdge(self.clock)
-
-        if int(result):
+        yield StableCondition(
+            lambda: self.bus.BVALID.value and self.bus.BREADY.value,
+            event=RisingEdge(self.clock), extra_event=True
+        )
+        result = self.bus.BRESP.value.integer
+        if result:
             raise AXIProtocolError("Write to address 0x%08x failed with BRESP: %d"
-                               % (address, int(result)))
-
+                                   % (address, result))
         raise ReturnValue(result)
 
     @cocotb.coroutine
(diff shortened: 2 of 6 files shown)


cocotb/triggers.py
```diff
@@ -729,3 +729,74 @@ def _wait(self):
...
+        while True:
+            yield ReadOnly()
+            if self.condition():
+                break
+            yield self.event
+        if self.extra_event:
+            yield self.event
+
+
+class StableValue(Waitable):
```
```suggestion
class StableValue(StableCondition):
```

cocotb/triggers.py
```diff
@@ -729,3 +729,74 @@ def _wait(self):
...
+    Wait until a given condition is True in ReadOnly state.
+
+    The trigger will first do ``yield ReadOnly()`` and then do the check.
+
+    Args:
+        condition (callable): Output of condition called without any parameters
+            is checked to be equivalent of logic True for condition to be met.
+        event (Trigger): event after which to check the condition; defaults to
+            nexttimestep.
+        extra_event (bool): wether to wait for an extra event after condition
```
```suggestion
        extra_event (bool): whether to wait for an extra event after condition
```

cocotb/drivers/__init__.py
```diff
@@ -276,10 +278,12 @@ def _wait_for_nsignal(self, signal):
         to move to :class:`~cocotb.triggers.NextTimeStep` before
         registering more callbacks can occur.
         """
-        yield ReadOnly()
-        while signal.value.integer != 0:
-            yield Edge(signal)
-            yield ReadOnly()
+        warnings.warn(
+            "Use of BusDriver._wait_for_signal() is deprecated\n"
+            "\tPlease, use cocotb.trigggers.StableValue instead",
```
eric-wieser:
```suggestion
            "\tPlease, use cocotb.triggers.StableValue instead",
```

cocotb/drivers/__init__.py
```diff
@@ -262,10 +262,12 @@ def _wait_for_signal(self, signal):
...
-        yield ReadOnly()
-        while signal.value.integer != 1:
-            yield RisingEdge(signal)
-            yield ReadOnly()
+        warnings.warn(
+            "Use of BusDriver._wait_for_signal() is deprecated\n"
+            "\tPlease, use cocotb.trigggers.StableValue instead",
+            DeprecationWarning, stacklevel=2
+        )
+        yield StableValue(signal, 1)
```
eric-wieser: The meaning changed on this line from `RisingEdge` to `Edge` - I don't think we can change things like that.

Perhaps the default should be `RisingEdge`?

cocotb/triggers.py
```diff
@@ -729,3 +729,74 @@ def _wait(self):
...
+            right value; defaults to ``False``.
+
+    .. versionadded: 1.3
+    """
+    def __init__(self, signal, value, **kwargs):
+        if value not in (0, 1):
+            raise ValueError("value has to be 0 or 1")
+        event = kwargs.pop('event', Edge(signal))
+        extra_event = kwargs.pop('extra_event', False)
+        reject_remaining_kwargs('StableCondition.__init__', kwargs)
```
eric-wieser: It might make more sense to just forward `**kwargs` here

cocotb/drivers/__init__.py
Fatsie: It has the same functionality. Original code used the fact that the value is 0 so waits for ``RisingEdge`` but if it would have waited for ``Edge`` it would have the same effect.
``RisingEdge`` does not work if you do ``StableValue(sig, 0)``.

Fatsie:
> not sure how I feel about this as a whole

At least I want to get rid of all these ``while True: ... break`` constructions spread all over the code. Am open for better suggestions in doing that.

cocotb/triggers.py
Fatsie: Will do, but will cause more confusing error messages when people are making this mistake.

cocotb/triggers.py
```diff
@@ -729,3 +729,74 @@ def _wait(self):
...
+    """
+    Wait until a given condition is True in ReadOnly state.
+
+    The trigger will first do ``yield ReadOnly()`` and then do the check.
+
+    Args:
+        condition (callable): Output of condition called without any parameters
+            is checked to be equivalent of logic True for condition to be met.
+        event (Trigger): event after which to check the condition; defaults to
+            nexttimestep.
```
```suggestion
            :class:`NextTimeStep`.
```

cocotb/triggers.py
```diff
@@ -729,3 +729,74 @@ def _wait(self):
...
+    Wait until a signal has a given value in ReadOnly state.
+
+    The trigger will first do ``yield ReadOnly()`` and then do the comparison.
+
+    Args:
+        signal (SimHandle): the signal to monitor.
+        event (Trigger): event after which next is done; default is waiting for
+            a change of the given signal; defaults to ``Edge(signal)``.
+        extra_event (bool): wether to wait for an extra event after value has the
+            right value; defaults to ``False``.
```
eric-wieser: Missing `value` argument.

For `event` and `extra_event`, perhaps it's best to just refer to the docstring of StableCondition?

cocotb/triggers.py
```diff
@@ -729,3 +729,74 @@ def _wait(self):
...
+class StableValue(StableCondition):
+    """
+    Wait until a signal has a given value in ReadOnly state.
+
+    The trigger will first do ``yield ReadOnly()`` and then do the comparison.
+
+    Args:
+        signal (SimHandle): the signal to monitor.
+        event (Trigger): event after which next is done; default is waiting for
+            a change of the given signal; defaults to ``Edge(signal)``.
```
eric-wieser: Perhaps add - "Note that this is equivalent to passing RisingEdge when value == 1, and FallingEdge when value == 0

cocotb/triggers.py
Fatsie: Yeah, cut and paste error.

cocotb/triggers.py
Fatsie: For ``StableValue`` the default for ``event`` changes from ``NextTimeStep`` to ``Edge``; that's why I could not pass kwargs on as is.

Fatsie: Will change StableValue docstring to:
```python
class StableValue(StableCondition):
    """
    Wait until a signal has a given value in ReadOnly state.

    The trigger will first do ``yield ReadOnly()`` and then do the comparison.

    Args:
        signal (SimHandle): the signal to monitor.
        value (int, 0 or 1): value to wait for.
        **kwargs: keyword arguments passed on to :class:`StableCondition`
            The default for ``event`` is changed to ``Edge(signal)``. This is
            equivalent to ``RisingEdge(signal)`` when waiting for 1 and
            ``FallingEdge(signal)`` when waiting for 0.

    .. versionadded: 1.3
    """
```

cocotb/triggers.py
Fatsie: Brought back popping of ``event`` as it changes the default value. See discussion below.

Fatsie:
Squashed all proposed changes. Handle linting errors.

cocotb/triggers.py
```diff
@@ -729,3 +729,73 @@ def _wait(self):
...
+class StableValue(StableCondition):
+    """
+    Wait until a signal has a given value in ReadOnly state.
+
+    The trigger will first do ``yield ReadOnly()`` and then do the comparison.
+
+    Args:
+        signal (SimHandle): the signal to monitor.
+        value (int, 0 or 1): value to wait for.
+        **kwargs: keyword arguments passed on to :class:`StableCondition`
```
```suggestion
        **kwargs: keyword arguments passed on to :class:`StableCondition`.
```

cocotb/drivers/amba.py
```diff
@@ -78,12 +78,8 @@ def _send_write_address(self, address, delay=0):
...
         self.bus.AWVALID <= 1
 
-        while True:
-            yield ReadOnly()
-            if self.bus.AWREADY.value:
-                break
-            yield RisingEdge(self.clock)
-        yield RisingEdge(self.clock)
+        yield StableValue(self.bus.AWREADY, 1, event=RisingEdge(self.clock),
+                          extra_event=True)
```
eric-wieser: For the sake of argument, do other reviewers think this is clearer than the following?
```python
yield StableCondition(lambda: self.bus.AWREADY == 1, event=RisingEdge(self.clock),
                      extra_event=True)
```

eric-wieser:
A possible alternate spelling:
```
yield RisingEdge(clk).until(lambda: sig == 1)
```
Which would mean creating a `TriggerMixins` class, and using it as the base class for `Trigger` and `_Waitable`, implementing something like:
```python
class TriggerMixins:
    def until(self, cond, **kwargs):
         return StableCondition(cond, event=self, **kwargs)
```

cocotb/drivers/amba.py
Fatsie: Reason I kept ``StableValue`` is that seems to be the common case.

Fatsie:
I like the ``.until()`` syntax and if we make it a coroutine we don't need the StableCondition/StableValue classes anymore:
```python
class TriggerHelpers:
    @cocotb.coroutine
    def until(self, cond, readonly=True):
        while not cond():
            yield self
            if readonly:
                yield ReadOnly()

**Comment by eric-wieser** [comment]
> if we make it a coroutine we don't need the classes any more

It's worth remembering what the point of the classes is:

* It lets a trigger be created and re-used:
    ```
    t = RisingEdge(clk).until(lambda: sig == 1)
    await t
    await t  # if this were a coroutine it would be a no-op, as the coroutine already finished
    ```
* It lets the identity be checked when used in `First`:
    ```
    t = RisingEdge(clk).until(lambda: sig == 1)
    which = await First(t, Timer(50))
    if which is t:
        pass  # wouldn't get here if `t` were a coroutine. For a waitable, this is the return value of `_wait`
    ```

**Comment by Fatsie** [comment]
If we keep ``StableValue``/``StableCondition`` I will leave the ``.until()`` syntax to follow-up patch. I got sidetracked from extending Wishbone support for my use case and really need to get back on track.

**Comment by jahagirdar** [comment]
The most common usescases  of this design pattern for me is, 
1. Waiting for the reset cycle to complete.
 The external reset cycle might be over, but I might have an internal reset sequence which can run for thousands of cycles.
Depending on the situation I will be waiting for either the signal to have a Non-X value, or a Non-Z value or a Non-0 value.

2. Waiting for transaction to start
Here I typically use the pattern
```if signal.value.integer==0:
       yield RisingEdge(signal) 
yield FallingEdge(clk)
```

ktbarrett:
I don't think we need this. And the use of `ReadOnly` is troublesome. I'm not sure if there is ever a good reason to use it. I vote to close.

Issue #1211, opened 2019-11-26
ParametrizedSingleton not so effective


Fatsie:
I modified code for ParametrizedSingleton  a little bit to give a message when a KeyError happens:
```python
class ParametrizedSingleton(type)
...
    def __call__(cls, *args, **kwargs):
        key = cls.__singleton_key__(*args, **kwargs)
        try:
            return cls.__instances[key]
        except KeyError:
            # construct the object as normal
            print("KeyError in {!r}: key={!r}".format(cls, key))
            self = super(ParametrizedSingleton, cls).__call__(*args, **kwargs)
            cls.__instances[key] = self
            return self
```
I made following test bench:
```python
import cocotb
from cocotb.triggers import ReadOnly, RisingEdge
from cocotb.clock import Clock


@cocotb.coroutine
def loop1(clk, n):
    for _ in range(n):
        yield RisingEdge(clk)

@cocotb.coroutine
def loop2(clk, n):
    for _ in range(n):
        yield ReadOnly()
        yield RisingEdge(clk)

@cocotb.test()
def test_bed3(dut):

    cocotb.fork(Clock(dut.clk, 10, 'ns').start())

    print("loop1")

    yield loop1(dut.clk, 10)

    print("loop2")

    yield loop2(dut.clk, 10)
```
It gives the following output:
```console
make results.xml
make[1]: Map '/home/verhaegs/eda/code/cocotb/tests/test_cases/test_bed' wordt binnengegaan
PYTHONPATH=/home/verhaegs/eda/code/cocotb/tests/test_cases/test_bed/build/libs/x86_64:/home/verhaegs/eda/code/cocotb/tests/test_cases/test_bed:/home/verhaegs/eda/code/cocotb: LD_LIBRARY_PATH=/home/verhaegs/eda/code/cocotb/tests/test_cases/test_bed/build/libs/x86_64:/home/verhaegs/software/lib:/home/verhaegs/software/lib64:/home/verhaegs/software/lib:/home/verhaegs/software/lib64::/usr/lib64:/usr/lib64:/usr/lib64:/usr/lib64 MODULE=test_bed3 \
        TESTCASE= TOPLEVEL=sample_module TOPLEVEL_LANG=verilog COCOTB_SIM=1 \
        /home/verhaegs/software/bin/vvp -M /home/verhaegs/eda/code/cocotb/tests/test_cases/test_bed/build/libs/x86_64 -m gpivpi sim_build/sim.vvp   
[...]

PR #1215, opened 2019-11-27
Clean up trigger caching.

Fatsie:
This follow perfomance discussion in #1209 and #1211 that came out of it.

The main question is if code should do optimization of trigger usage by caching them a before loop. If yes this PR can be closed without merging. Otherwise likely some optimization may still be needed before merging this PR. For the latter I don't have currently the bandwidth to work on it.

### DIFF
--- a/cocotb/drivers/__init__.py
+++ b/cocotb/drivers/__init__.py
@@ -76,17 +76,15 @@ class BitDriver(object):
         if generator is not None:
             self._generator = generator
 
-        edge = RisingEdge(self._clk)
-
         # Actual thread
         while True:
             on, off = next(self._generator)
             self._signal <= 1
             for _ in range(on):
-                yield edge
+                yield RisingEdge(self._clk)
             self._signal <= 0
             for _ in range(off):
-                yield edge
+                yield RisingEdge(self._clk)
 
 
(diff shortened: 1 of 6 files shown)


eric-wieser:
> The main question is if code should do optimization of trigger usage by caching them a before loop.

It's worth noting that this might not only be a performance optimization, but a _readability_ one. The existing code is easier to check that the same signal is being used for every clock edge, as there a fewer words.

Fatsie:
> It's worth noting that this might not only be a performance optimization, but a _readability_ one.

I agree that there may be other reasons than performance for assigning these events to a variable. Although in this case were often the event is only used once in the loop. I guess this is also a matter of taste; I would only do it if you have more than one clock and/or more than one signal.

What I do want to achieve with this patch to get rid of the believe that events have to be cached for performance reasons. Existing code is often used as starting point rather than docs so it should not be reason for creating false believes.

imphil:
I do think this change improves readability. However: do you have performance numbers for this?

ktbarrett:
@eric-wieser I don't really see any reason this needs to be merged. If you agree, please close.

ktbarrett:
This change no longer applies to this repo.

PR #1216, opened 2019-11-27
Singleton

Fatsie:
Implementation Singleton as proposed during discussion in #1209; a change I did was that I now create the object already during class init.

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -158,6 +158,12 @@ class external_waiter(object):
 
         return self.state
 
+# Singleton events, cached to avoid need for calling __init__() method
+# when using the event.
+_next_time_step = NextTimeStep()
+_read_write = ReadWrite()
+_read_only = ReadOnly()
+
 class Scheduler(object):
     """The main scheduler.
 
@@ -210,12 +216,6 @@ class Scheduler(object):
     _MODE_WRITE    = 3  # noqa
     _MODE_TERM     = 4  # noqa
 
-    # Singleton events, cached to avoid need for calling __init__() method
-    # when using the event.
-    _next_time_step = NextTimeStep()
-    _read_write = ReadWrite()
-    _read_only = ReadOnly()
-
     _timer1 = Timer(1)
 
     def __init__(self):
@@ -259,9 +259,9 @@ class Scheduler(object):
         while True:
             yield self._writes_pending.wait()
             if self._mode != Scheduler._MODE_NORMAL:
-                yield Scheduler._next_time_step
+                yield _next_time_step
 
-            yield Scheduler._read_write
+            yield _read_write
 
             while self._writes:
                 handle, value = self._writes.popitem()
@@ -383,7 +383,7 @@ class Scheduler(object):
                                    str(trigger))
                 return
 
-            if trigger is Scheduler._read_only:
+            if trigger is _read_only:
                 self._mode = Scheduler._MODE_READONLY
             # Only GPI triggers affect the simulator scheduling mode
             elif isinstance(trigger, GPITrigger):
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -259,9 +259,9 @@ class Scheduler(object):
         while True:
             yield self._writes_pending.wait()
             if self._mode != Scheduler._MODE_NORMAL:
-                yield self._next_time_step
+                yield Scheduler._next_time_step
 
-            yield self._read_write
+            yield Scheduler._read_write
 
             while self._writes:
                 handle, value = self._writes.popitem()
@@ -383,7 +383,7 @@ class Scheduler(object):
                                    str(trigger))
                 return
 
-            if trigger is self._read_only:
+            if trigger is Scheduler._read_only:
                 self._mode = Scheduler._MODE_READONLY
             # Only GPI triggers affect the simulator scheduling mode
             elif isinstance(trigger, GPITrigger):
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -210,10 +210,12 @@ class Scheduler(object):
     _MODE_WRITE    = 3  # noqa
     _MODE_TERM     = 4  # noqa
 
-    # Singleton events, recycled to avoid spurious object creation
+    # Singleton events, cached to avoid need for calling __init__() method
+    # when using the event.
     _next_time_step = NextTimeStep()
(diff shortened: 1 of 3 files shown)


cocotb/scheduler.py
```diff
@@ -210,10 +210,9 @@ class Scheduler(object):
     _MODE_WRITE    = 3  # noqa
     _MODE_TERM     = 4  # noqa
 
-    # Singleton events, recycled to avoid spurious object creation
-    _next_time_step = NextTimeStep()
-    _read_write = ReadWrite()
-    _read_only = ReadOnly()
```
eric-wieser: I'd be inclined to keep these. Within the scheduler itself, every bit of optimization counts much more than it does in user or driver code.

cocotb/utils.py
```diff
@@ -408,6 +408,22 @@ def highlight(string, colour=ANSI.COLOR_HILITE_HEXDIFF_DEFAULT):
...
+    We use this so that :class:`ReadOnly() <cocotb.triggers.ReadOnly>` always returns
+    the same instance, rather than creating new copies. This is an optimized version
+    of :class:`ParametrizedSingleton`, to avoid a lookup parameters - which gives a 1-2%
+    speedup. The object will also be created at class init so it will be available
+    all the time.
+    """
+
+    def __init__(cls, *args, **kwargs):
+        # Generate an object during class init and return that later on when called.
+        cls.__instance_ref = super(Singleton, cls).__call__()
```
eric-wieser:
```suggestion
        super(Singleton, cls).__init__(name, bases, dict)
        cls.__instance_ref = super(Singleton, cls).__call__()
```

cocotb/utils.py
```diff
@@ -408,6 +408,22 @@ def highlight(string, colour=ANSI.COLOR_HILITE_HEXDIFF_DEFAULT):
...
+    """A metaclass that allows class construction to reuse an existing instance.
+
+    We use this so that :class:`ReadOnly() <cocotb.triggers.ReadOnly>` always returns
+    the same instance, rather than creating new copies. This is an optimized version
+    of :class:`ParametrizedSingleton`, to avoid a lookup parameters - which gives a 1-2%
+    speedup. The object will also be created at class init so it will be available
+    all the time.
+    """
+
+    def __init__(cls, *args, **kwargs):
```
eric-wieser:
```suggestion
    def __init__(cls, name, bases, dict):
```

eric-wieser:
@andresdemski, mind running your profiler before, after the first commit, and after the second commit?

cocotb/scheduler.py
eric-wieser: Let's see what profiling says the effect is before reverting this commit - thanks for keeping it separate

andresdemski:
Hi eric!
I've run the performance test and generated the profile.

Here are the results:

# Performance

```
7f72c5c134e1ee798b2b1d3eb5c995f9bd03c0ae
SIMTIME:270162.40
SIM/REALTIME:20135.83
1e242ebaeeab0ba8665c59957a81bdc8a9b825d3
SIMTIME:270162.40
SIM/REALTIME:20612.09
17ce4e51da43b6d2d1b5834d629374ed5535a8d8
SIMTIME:270162.40
SIM/REALTIME:20333.09
```
**Results** : Almost the same SIM/REAL TIME

# Profiles

## 1e242eba
image

## 17c34e5
image

## 7f72c5c13
image


The graph looks like different but i don't know if it is expected.
[sigleton_profiles.zip](https://github.com/cocotb/cocotb/files/3902960/sigleton_profiles.zip)

eric-wieser:
Looks like the first commit is a1% speedup, and the second is a worse slowdown.

Fatsie:
If we are looking at performance in such a detail, maybe we can go all the way. I added some more possible optimizations.
@andresdemski, could you also look at profile after 8fefbf4 and after 3d1e47b ?

Fatsie:
Changed to not touch ``_timer1``. I do think that field could use some documentation from a guy who is familiar with it's intricacies.
@andresdemski , I am interested to know performance impact now after 5faf2d4 and 4ee9a14.

andresdemski:
```
 url https://github.com/Fatsie/cocotb.git
 commit: 4ee9a14482b1449afd150ea27fd3ab417b67c73b
 path: /tmp/fatsie
 results: 
    simtime: 270162.40 ns
    realtime: 13.06s
    performance: 20689.18 ns/s


 url https://github.com/Fatsie/cocotb.git
 commit: 5faf2d4b59ff1ffad26a8e8c3c4a077a49ef6bb5
 path: /tmp/fatsie
 results: 
    simtime: 270162.40 ns
    realtime: 12.75s
    performance: 21182.44 ns/s


 url https://github.com/Fatsie/cocotb.git
 commit: 773c846d5f2e0993df1da6f0f2d0f426d60c4204
 path: /tmp/fatsie
 results: 
    simtime: 270162.40 ns
    realtime: 12.91s
    performance: 20927.63 ns/s


 url https://github.com/Fatsie/cocotb.git
 commit: afe35ecbca18a887a1e766e26d166d54bdde8bc0
 path: /tmp/fatsie
 results: 
    simtime: 270162.40 ns
    realtime: 13.94s
    performance: 19377.56 ns/s
```

eric-wieser:
@andresdemski: Looks like the benchmarking technique isn't precise enough to be useful here - the impact of the addition of the comment was larger than any other change!

andresdemski:
Yep, i suppose i should run it many times and do the mean of the results. I'm really busy now, maybe in a few weeks i could do it.

Fatsie:
@andresdemski What is principle for the performace measurement ?
Is it done by instrumenting the function calls in python or is it based on statistic sampling. In the former case one has to be careful with interpreting the values as the instrumentation may impact measured performance.

cmarqu:
FWIW, Andres' code is here: https://github.com/andresdemski/cocotb-performance

imphil:
@Fatsie @eric-wieser Is this purely a performance optimization? Which then, according to the measurements, doesn't change runtime in a measurable way? Should we close this as an optimization which doesn't seem to work out, or is there more to it?

Fatsie:
I leave the judging to core devs if this is appropriate for joining. Feel free to close this ticket; it does not provide extra functionality.
In general I get the idea that it is not currenlty clear where one wants to go with optimizations in cocotb development.

ktbarrett:
Closing as out of date and looks like it's not that valuable.

Issue #1195, opened 2019-11-19
Document supported C and C++ version


imphil:
From https://github.com/cocotb/cocotb/pull/1149#issuecomment-545019530

> If you are going to call out the C++ standard, it's a good idea to call out the C standard as well. GCC 4.8.1 does not fully support C11, so the C standard should be gnu99. The GNU extensions are needed to enable POSIX standards required by Python, and are supported by Clang as well.

Documentation should be added to https://cocotb.readthedocs.io/en/latest/quickstart.html

We currently specify the minimum supported GCC version (intentionally), since that's the compiler we need to be compatible (see #1087 for details).

While I'd like to keep specifying the minimum compiler version (since that's what matters for compat), we should add the C and C++ versions to make it obvious what can be used in software:

- For C++, it's C++11 ("GCC 4.8.1 was the first feature-complete implementation of the 2011 C++ standard, previously known as C++0x.", https://gcc.gnu.org/projects/cxx-status.html)
- For C, it's gnu99 (1)

(1) Full C11 might never happen if we want to stay compatible with MSVC, but that's a very different story for another day.

ktbarrett:
#1137 adds `-std=gnu99` to the makefiles and mentions that [Clang 3.3+](https://clang.llvm.org/cxx_status.html) is supported.

eric-wieser:
@imphil, probably worth doing this for the release - I don't remember whether the requirements have changed since 1.3.

ktbarrett:
This is probably only necessary in the case of C++. On Linux LD_LIBRARY_PATH has to point to a version of libstdc++ that support C++11. I've run into this problem before when using really old version of Questa, because it prepends it's libraries to the LD_LIBRARY_PATH. Recently we had #1933 who was running an old OS with a screwed up environment. And that will be an issue once (if?) we start distributing binary packages and people try running cocotb on old OSes (like RHEL6). It would no longer fail to install, but it wouldn't run :)

PR #1217, opened 2019-11-30
Provide a Task-Based BFM Interface

mballance:
This PR implements a task-based BFM interface between Cocotb and the HDL simulation environment. This enables connecting Python method calls to HDL task calls and vice versa. More details are provided in the task_bfms.rst documentation file, and a test is provided in test_bfms_verilog.

https://external-builds.readthedocs.io/html/cocotb/1217/task_bfms.html


cocotb/bfmgen.py
```diff
@@ -0,0 +1,398 @@
...
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL POTENTIAL VENTURES LTD BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
```
cmarqu: I saw @imphil mention in the Verilator PR that new code should get the SPDX license header:
```
```

cocotb/bfmgen.py
```diff
@@ -0,0 +1,398 @@
...
+    
+    
+    return t.safe_substitute(param_m)
+
+def bfm_generate_vl(args):
+    inst = BfmMgr.inst()
+    
+    out = open(args.o, "w")
+    out.write("//***************************************************************************\n")
+    out.write("//* BFMs file for CocoTB. \n")
```
cmarqu: Nit: I have tried to normalize the spelling of "cocotb" to be all lowercase in the last few months.
(I think we are still inconsistent when it's at the beginning of a sentence though.)

cocotb/bfmgen.py
```diff
@@ -0,0 +1,398 @@
...
+    out_c.write("}\n")
+    out_c.write("#endif\n")
+    
+    out_sv.close() 
+    out_c.close()
+
+def bfm_generate(args):
+    '''
+    Generates BFM files required for simulation
+    '''
```
cmarqu: Nit: Write this as """Generate BFM files required for simulation.""" (https://www.python.org/dev/peps/pep-0257/#id16)

cmarqu:
Awesome, thanks for this PR!
The documentation in rendered for can be looked at in https://external-builds.readthedocs.io/html/cocotb/1217/task_bfms.html

cocotb/bfms.py
```diff
@@ -0,0 +1,236 @@
...
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL POTENTIAL VENTURES LTD BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
```
cmarqu: Another case for the SPDX header.

cocotb/bfms.py
```diff
@@ -0,0 +1,236 @@
...
+        '''
+        import simulator
+        n_bfms = simulator.bfm_get_count()
+        for i in range(n_bfms):
+            info = simulator.bfm_get_info(i)
+            typename = info[0]
+            instname = info[1]
+            clsname = info[2]
+            if clsname.find('.') == -1:
+                raise Exception("Incorrectly-formatted BFM class name \"" + clsname + "\"")
```
cmarqu: I recently learned from @eric-wieser that this can be written like
```suggestion
                raise Exception("Incorrectly-formatted BFM class name {!r}".format(clsname)
```
and you don't have to care about the quote escaping anymore: https://docs.python.org/3/library/string.html#formatspec

cocotb/bfmgen.py
mballance: Yeah, there will be a few of those. Glad to hear that the direction is towards a more-standard copyright header... @imphil, please confirm that the header @cmarqu shows above is the correct one and I'll happily update the copyright headers on the new files.

cocotb/regression.py
```diff
@@ -124,8 +124,10 @@ def initialise(self):
         self._dut = cocotb.handle.SimHandle(handle) if handle else None
 
         if self._dut is None:
-            raise AttributeError("Can not find Root Handle (%s)" %
```
cmarqu: I guess this is going to raise a few eyebrows without an explanation.

cocotb/share/include/cocotb_bfm_api.h
```diff
@@ -0,0 +1,142 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Callback function type used to notify
+// implementation when a message is available
+typedef void (*cocotb_bfm_notify_f)(void *);
+
+int cocotb_bfm_register(
+		const char				*type_name,
```
cmarqu: Nit: I haven't seen tabs used elsewhere AFAIR.

cocotb/regression.py
mballance: Ahh... Perhaps I should remove this from the task-based BFM PR. I added this to allow cocotb to run with mainline Verilator (4.022). Mainline Verilator doesn't support full VPI, but still works quite nicely with task-based BFMs as long as cocotb doesn't error out due to not being able to access signals.

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,192 @@
...
+            def write_req(self, d):
+                pass
+    
+            @cocotb.bfm_export()
+            def write_ack(self):
+                self.ack_ev.set()
+
+Python methods that will result in task calls in the HDL are 
+decorated with the @cocotb.bfm_import decorator, while 
+Python methods that will be called from the HDL are decoarated
```
cmarqu:
```suggestion
Python methods that will be called from the HDL are decorated
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,192 @@
+###############
+Task-Based BFMs
+###############
+
```
cmarqu:
```suggestion
.. versionadded:: 1.3
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,192 @@
...
+Overview
+========
+
+Cocotb supports a task-based bus functional model (BFM) interface between 
+Python, Verilog, and SystemVerilog (VHDL is TBD). 
+
+BFM Implementation (Python)
+===========================
+
+The Python aspect of a BFM is captured as a Python class with the @cocotb.bfm
```
cmarqu: This will both mark up and link the decorator name:
```suggestion
The Python aspect of a BFM is captured as a Python class with the :class:`cocotb.bfm`
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,192 @@
...
+Using BFMs from HDL
+===================
+The HDL portion of the testbench must instantiate BFMs where needed.
+These instances will register their existence with Cocotb when simulation
+starts. 
+
+
+Using BFMs from Python
+======================
+Available BFM instances are registered with the ``cocotb.BfmMgr`` class.
```
cmarqu:
```suggestion
Available BFM instances are registered with the :class:`cocotb.BfmMgr` class. 
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,192 @@
...
+The HDL portion of the testbench must instantiate BFMs where needed.
+These instances will register their existence with Cocotb when simulation
+starts. 
+
+
+Using BFMs from Python
+======================
+Available BFM instances are registered with the ``cocotb.BfmMgr`` class. 
+Static methods provide access to the list of available BFMs, and the
+``find_bfm`` method accepts a regular expression to find a BFM based
```
cmarqu:
```suggestion
:meth:`~cocotb.BfmMgr.find_bfm` method accepts a regular expression to find a BFM based
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,192 @@
...
+
+.. code-block:: python3
+    @cocotb.coroutine
+    def run_c(self):
+        out_bfm = BfmMgr.find_bfm(".*u_bfm")
+        
+        for i in range(1,101):
+            yield out_bfm.write_c(i)
+
+The code snippet above shows typical use within a test. The ``find_bfm``
```
cmarqu:
```suggestion
The code snippet above shows typical use within a test. The :meth:`~cocotb.BfmMgr.find_bfm`
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,192 @@
...
+method is used to find a BFM with the expected instance path. Then,
+methods on the BFM object are called to send data via the BFM.
+
+Cocotb Makefile Interface
+=========================
+If you are using the Cocotb Makefiles, simply append the BFM packages
+used by your testbench to the COCOTB_BFM_MODULES variable
```
cmarqu:
```suggestion
used by your testbench to the :make:var:`COCOTB_BFM_MODULES` variable
```
This variable should probably also be added to the ``building.rst`` file.

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,192 @@
...
+files along with the rest of your testbench.
+
+Manually Generating BFM Interface Files
+=======================================
+The interface code that allows Cocotb to call HDL tasks, and to enable
+HDL to call Python methods is auto-generated. This ensures that the 
+HDL interface is always up-to-date with the Python definition of the
+BFM API.
+
+The ``cocotb-bfmgen`` script generates the appropriate BFM interface
```
cmarqu: Nit: semantic markup is nice (I like to use https://sphinx-tutorial.readthedocs.io/cheatsheet/ to find what's there):
```suggestion
The :command:`cocotb-bfmgen` script generates the appropriate BFM interface
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,192 @@
...
+files based on the BFMs required for a given testbench.
+
+The ``cocotb-bfmgen`` script accepts the following options:
+- -m <module> -- Specifies a Python module to load. Typically, this will
```
cmarqu: FWIW, there is https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-program but no need to go overboard now.

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,192 @@
...
+=======================================
+The interface code that allows Cocotb to call HDL tasks, and to enable
+HDL to call Python methods is auto-generated. This ensures that the 
+HDL interface is always up-to-date with the Python definition of the
+BFM API.
+
+The ``cocotb-bfmgen`` script generates the appropriate BFM interface
+files based on the BFMs required for a given testbench.
+
+The ``cocotb-bfmgen`` script accepts the following options:
```
cmarqu: This doesn't render as a list, I guess you need an empty line after the colon:

tests/test_cases/test_bfms_verilog/rv_bfms/rv_data_monitor_bfm.py
```diff
@@ -0,0 +1,29 @@
+'''
```
cmarqu: I guess these should rather have the SPDX header, if at all.

cocotb/bfmgen.py
```diff
@@ -0,0 +1,398 @@
...
+    
+    args = parser.parse_args()
+ 
+    # Ensure the BfmMgr is created
+    BfmMgr.inst()
+    
+    if hasattr(args, 'm') and args.m is not None:
+        bfm_load_modules(args.m)
+        
+    args.func(args)
```
cmarqu: If I call ``cocotb-bfmgen`` without any arguments or options, Python says about this line:
``AttributeError: 'Namespace' object has no attribute 'func'``

cocotb/bfmgen.py
```diff
@@ -0,0 +1,398 @@
...
+    
+    return t.safe_substitute(param_m)
+
+def bfm_generate_vl(args):
+    inst = BfmMgr.inst()
+    
+    out = open(args.o, "w")
+    out.write("//***************************************************************************\n")
+    out.write("//* BFMs file for CocoTB. \n")
+    out.write("//* Note: This file is generated. Do Not Edit\n")
```
cmarqu: Maybe mention ``cocotb-bfmgen`` in this line?

cocotb/bfmgen.py
cmarqu: See also https://github.com/cocotb/cocotb/pull/943#discussion_r329165473

cocotb/bfmgen.py
mballance: Perfect, update on the way

cocotb/regression.py
mballance: Added an explanatory comment, and updated the message to better-describe the impact of not having access to the design root.

cocotb/bfmgen.py
mballance: Makes sense, done!

tests/test_cases/test_bfms_verilog/rv_bfms/rv_data_monitor_bfm.py
mballance: Yes, just took a pass through and updated all new files and added the SPDX header

cocotb/bfmgen.py
mballance: Perfect, just took a pass through and updated all new files with the SPDX header

cocotb/bfmgen.py
mballance: Good catch! Just pushed a fix

cocotb/share/include/cocotb_bfm_api.h
mballance: Good point. I've converted tabs to spaces in the C/C++ files

documentation/source/task_bfms.rst
mballance: Okay, got this one taken care of. I also linked the Task BFM docs into the documentation tree

cocotb/bfms.py
```diff
@@ -0,0 +1,208 @@
...
+    global import_info_l
+    global export_info_l
+    
+    type_info = BfmTypeInfo(T, hdl, import_info_l.copy(), export_info_l.copy())
+    BfmMgr.inst().add_type_info(T, type_info)
+    import_info_l = []
+    export_info_l = []
+    
+def register_bfm_import_info(info):
+    global import_info_l
```
eric-wieser:
```suggestion
```
This line doesn't do anything. Global is for when you assign to the name.

documentation/source/task_bfms.rst
cmarqu: Ah, it's still not hyperlinked because the whole module should be getting a section in ``library_reference.rst``.
(Also, you are referencing the same decorator a few lines down, please add the same markup there as well.)

cocotb/share/lib/gpi/GpiBfm.cpp
```diff
@@ -0,0 +1,87 @@
...
+    m_active_inbound_msg = 0;
+}
+
+GpiBfm::~GpiBfm() {
+    if (m_active_msg) {
+        delete m_active_msg;
+        m_active_msg = 0;
+    }
+    if (m_active_inbound_msg) {
+    	delete m_active_inbound_msg;
```
eric-wieser:
```suggestion
    	  delete m_active_inbound_msg;
```

cocotb/bfmgen.py
```diff
@@ -0,0 +1,371 @@
...
+        filename_c = os.path.splitext(filename_c)[0]
+    filename_c += ".c"
+    
+    out_sv = open(args.o, "w")
+    out_sv.write("//***************************************************************************\n")
+    out_sv.write("//* BFMs file for CocoTB. \n")
+    out_sv.write("//* Note: This file is generated. Do Not Edit\n")
+    out_sv.write("//***************************************************************************\n")
+
+    out_c = open(filename_c, "w")
```
eric-wieser: Better to use
````
with open(filename_c, "w") as out_c:
    out_c.write(...)
    # don't call out_c.close() any more
```

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+        out.write("//***************************************************************************\n")
+        out.write("//* BFMs file for cocotb. \n")
+        out.write("//* Note: This file is generated by cocotb-bfmgen. Do Not Edit\n")
+        out.write("//***************************************************************************\n")
+
+        for t in inst.bfm_type_info_m.keys():
+            info = inst.bfm_type_info_m[t]
+        
+            if cocotb.bfm_vlog not in info.hdl.keys():
+                raise Exception("BFM \"" + t.__name__ + "\" does not support Verilog")
```
eric-wieser: nit:
```suggestion
                raise Exception("BFM {!r} does not support Verilog".format(t.__name__))
```

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+            if j+1 < len(exp.signature):
+                bfm_export_tasks += ", "
+        bfm_export_tasks += ");\n"
+        bfm_export_tasks += "    begin\n"
+        bfm_export_tasks += "        cocotb_bfm_begin_msg(bfm_id, " + str(i) + ");\n"
+        for p in exp.signature:
+            if p.ptype.s:
+                bfm_export_tasks += "        cocotb_bfm_add_si_param(bfm_id, " + p.pname + ");\n"
+            else:
+                bfm_export_tasks += "        cocotb_bfm_add_ui_param(bfm_id, " + p.pname + ");\n"
```
eric-wieser: nit: tempting to write these using `.format`, such as
```suggestion
                bfm_export_tasks += "        cocotb_bfm_add_ui_param(bfm_id, {});\n".format(p.pname)
```

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+        
+            if cocotb.bfm_vlog not in info.hdl.keys():
+                raise Exception("BFM \"" + t.__name__ + "\" does not support Verilog")
+        
+            with open(info.hdl[cocotb.bfm_vlog], "r") as template_f:
+                template = template_f.read()
+        
+            out.write(process_template_vl(template, info))
+        
+def process_template_sv(template, bfm_name, info):
```
eric-wieser: Please add a docstring here that explains the arguments and return value

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+        
+    t = Template(template)
+    
+    bfm_import_calls = ""
+    for i in range(len(info.import_info)):
+        imp = info.import_info[i]
```
eric-wieser:
```suggestion
    for i, imp in enumerate(info.import_info):
```

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+from cocotb.bfms import BfmMgr
+
+
+def bfm_load_modules(module_l): 
+    for m in module_l:
+        try:
+            importlib.import_module(m)
+        except Exception as e:
+            print("Error: failed to load module \"" + str(m) + "\": " + str(e))
+            raise e
```
eric-wieser: This doesn't seem like a good idea, the caller should log this exception anyway.

If you really want to keep it, can you change it to use a logger instead?

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+
+def process_template_vl(template, info):
+    t = Template(template)
+    
+    bfm_import_calls = ""
+    for i in range(len(info.import_info)):
```
eric-wieser:
```suggestion
    for i, imp in enumerate(info.import_info):
```
and remove the line below

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+        imp = info.import_info[i]
+        bfm_import_calls += "              " + str(i) + ": begin\n"
+        bfm_import_calls += "                  " + imp.T.__name__  + "(\n"
+        for pi in range(len(imp.signature)):
+            p = imp.signature[pi]
```
eric-wieser: `enumerate` here too

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
+###############################################################################
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+###############################################################################
```
eric-wieser: Please add a docstring describing this module below this line. Even a brief description is better than nothing.

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+            
+            if pi+1 < len(imp.signature):
+                bfm_import_calls += ","
+            bfm_import_calls += "\n"
+        bfm_import_calls += "                      );\n"
+        bfm_import_calls += "              end\n"
+        
+    bfm_export_tasks = ""
+    for i in range(len(info.export_info)):
+        exp = info.export_info[i]
```
eric-wieser: `enumerate`

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+            
+        bfm_export_tasks += "        $cocotb_bfm_end_msg(bfm_id);\n"
+        bfm_export_tasks += "    end\n"
+        bfm_export_tasks += "    endtask\n"
+        
+    
+    impl_param_m = {
+        "bfm_classname" : info.T.__module__ + "." + info.T.__qualname__,
+        "bfm_import_calls" : bfm_import_calls,
+        "bfm_export_tasks" : bfm_export_tasks
```
eric-wieser: nit: PEP8 wants no whitespace before the `:`:
```suggestion
        "bfm_export_tasks": bfm_export_tasks
```

In practice, I like to write this type of thing using `dict` so that the names are unquoted:
```
impl_param_m = dict(
    bfm_classname=info.T.__module__ + "." + info.T.__qualname__,
    bfm_import_calls=bfm_import_calls,
    bfm_export_tasks=bfm_export_tasks
)
```

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+def bfm_generate_vl(args):
+    inst = BfmMgr.inst()
+    
+    with open(args.o, "w") as out:
+        out.write("//***************************************************************************\n")
+        out.write("//* BFMs file for cocotb. \n")
+        out.write("//* Note: This file is generated by cocotb-bfmgen. Do Not Edit\n")
+        out.write("//***************************************************************************\n")
+
+        for t in inst.bfm_type_info_m.keys():
```
eric-wieser:
```suggestion
        for t, info in inst.bfm_type_info_m.items():
```

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+int ${bfm_name}_process_msg() { }
+
+static void ${bfm_name}_notify_cb(void *user_data) {
+    svSetScope(user_data);
+    ${bfm_name}_process_msg();
+}
+
+int ${bfm_name}_register(const char *inst_name) {
+    return cocotb_bfm_register(
+        \"XXXX\",
```
eric-wieser: What are the backslashes here for?

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+    Generate BFM files required for simulation
+    '''
+    
+    if args.o is None:
+        if args.language == "vlog":
+            args.o = "cocotb_bfms.v"
+        elif args.language == "sv":
+            args.o = "cocotb_bfms.sv"
+        elif args.language == "vhdl":
+            args.o = "cocotb_bfms.vhdl"
```
eric-wieser:
```suggestion
            args.o = "cocotb_bfms.vhdl"
        else:
            raise NotImplementedError
```
or similar

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+# def bfm_list(args):
+#     inst = BfmMgr.inst()
+#     
+#     # TODO: implement list command
+#     
+#     print("inst=" + str(inst))
+# 
+#     print("Number of keys: " + str(len(inst.bfm_type_info_m.keys())))
+#     for t in inst.bfm_type_info_m.keys():
+#         print("BFM: \"" + str(t) + "\"")
```
eric-wieser: Just a reminder that we should either finish or remove this code before merging

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
...
+            raise Exception("Attempting to register a global method as a BFM method")
+        
+        args = fi.co_varnames[1:fi.co_argcount]
+        if len(signature) != len(args):
+            raise Exception("Wrong number of parameter-type elements: expect " + str(len(args)) + " but received " + str(len(signature)))
+        
+        
+        for i in range(len(args)):
+            a = args[i]
+            t = signature[i]
```
eric-wieser:
```suggestion
        for a, t in zip(args, signature):
```

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
...
+                import simulator
+                if isinstance(t, bfm_param_int_t):
+                    if t.s:
+                        self.type_info.append(simulator.BFM_SI_PARAM)
+                    else:
+                        self.type_info.append(simulator.BFM_UI_PARAM)
+            except Exception:
+                # When we're not running in simulation, don't 
+                # worry about being able to access constants from simulation
+                self.type_info.append(None)
```
eric-wieser:
```suggestion
            try:
                import simulator
            except Exception:
                # When we're not running in simulation, don't 
                # worry about being able to access constants from simulation
                self.type_info.append(None)
            else:
                if isinstance(t, bfm_param_int_t):
                    if t.s:
                        self.type_info.append(simulator.BFM_SI_PARAM)
                    else:
                        self.type_info.append(simulator.BFM_UI_PARAM)
```

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class BfmTypeInfo():
```
eric-wieser: Because we still support python 2 (#767), all of your classes need to be changed to use `(object)`, which is only implied in python 3.

```suggestion
class BfmTypeInfo(object):
```

Either that, or we need to put this PR off until after #767.

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
...
+        self.inst_name = inst_name
+        self.type_info = type_info
+        
+    def call_method(self, method_id, params):
+        self.type_info.export_info[method_id].T(
+            self.bfm, *params)
+
+class BfmMgr():
+    
+    m_inst = None
```
eric-wieser: This is C-style naming - in python the convention is `_inst`, or `__inst` if you're really paranoid.

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
...
+        bfm = None
+       
+        path_pattern_re = re.compile(path_pattern)
+        
+        for b in inst.bfm_l:
+            if path_pattern_re.match(b.bfm_info.inst_name):
+                bfm = b
+                break
+        
+        return bfm
```
eric-wieser: nit: Arguably more readable as:
```python
matches = (
    b
    for b in inst.bfm_l
    if path_pattern_re.match(b.bfm_info.inst_name
)
return next(matches, None)
```

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
...
+        import simulator
+        n_bfms = simulator.bfm_get_count()
+        for i in range(n_bfms):
+            info = simulator.bfm_get_info(i)
+            instname = info[1]
+            clsname = info[2]
+            if clsname.find('.') == -1:
+                raise Exception("Incorrectly-formatted BFM class name {!r}".format(clsname))
+            pkgname = clsname[:clsname.rfind('.')]
+            clsleaf = clsname[clsname.rfind('.')+1:]
```
eric-wieser: Simpler as:
```python
try:
    pkgname, clsleaf = clsname.rsplit('.', 1)
except ValueError:
    raise Exception("Incorrectly-formatted BFM class name {!r}".format(clsname))
```

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
...
+          
+            try: 
+                pkg = importlib.import_module(pkgname)
+            except Exception as e:
```
eric-wieser: Unused variable e, 
```suggestion
            except Exception:
```

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
+###############################################################################
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+###############################################################################
+
+import os
+import importlib
```
eric-wieser: This is new in python 3.1 - are you waiting for #767 to be resolved?

cocotb/decorators.py
```diff
@@ -454,3 +454,95 @@ def __init__(self, f, timeout=None, expect_fail=False, expect_error=False,
...
+        
+        return ret
+
+# Constants for use in specifying BFM API signatures
+bfm_int8_t = bfm_param_int_t(32, True)
+bfm_uint8_t = bfm_param_int_t(32, False)
+bfm_int16_t = bfm_param_int_t(32, True)
+bfm_uint16_t = bfm_param_int_t(32, False)
+bfm_int32_t = bfm_param_int_t(32, True)
+bfm_uint32_t = bfm_param_int_t(32, False)
```
eric-wieser: Copy-paste error, presumably the first 4 should be 8 and 16 bit?

cocotb/regression.py
eric-wieser: This should at least log.

I'm thinking we should only allow this to continue after checking the simulator environment variable for whether this is verilator.

cocotb/share/include/cocotb_bfm_api.h
```diff
@@ -0,0 +1,118 @@
...
+        uint32_t            bfm_id,
+        uint64_t            p);
+
+void cocotb_bfm_add_si_param(
+        uint32_t            bfm_id,
+        int64_t                p);
+
+void cocotb_bfm_add_str_param(
+        uint32_t            bfm_id,
+        const char            *p);
```
eric-wieser: nit: whitespace is all over the place with the column alignment, I'd be inclined to just replace all these spaces with a single space, and perhaps fold them onto a single line.

Same throughout this file.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -134,6 +134,11 @@ void embed_init_python(void)
     if (gtstate)
         return;
 
+    // Before loading Python, ensure that COCOTB_SIM is set
+    if (!getenv("COCOTB_SIM") || !getenv("COCOTB_SIM")[0]) {
+    	setenv("COCOTB_SIM", "1", 1);
```
eric-wieser:
```suggestion
    	  setenv("COCOTB_SIM", "1", 1);
```

cocotb/share/lib/gpi/GpiBfm.h
```diff
@@ -0,0 +1,74 @@
...
+    std::string                      m_clsname;
+    cocotb_bfm_notify_f              m_notify_f;
+    void                             *m_notify_data;
+    std::vector<GpiBfmMsg *>         m_msg_queue;
+    GpiBfmMsg                        *m_active_msg;
+    // Message ready to be sent to
+    GpiBfmMsg                        *m_active_inbound_msg;
+
+    static bfm_recv_msg_f            m_recv_msg_f;
+    static std::vector<GpiBfm *>     m_bfm_l;
```
eric-wieser: Please add a doxygen comment (`/**< comment */`) after each of these fields

cocotb/share/lib/gpi/GpiBfmMsg.h
```diff
@@ -0,0 +1,75 @@
...
+protected:
+
+private:
+    uint32_t                                m_id;
+    cocotb_bfm_msg_param_t                  *m_param_l;
+    uint32_t                                m_param_l_idx;
+    uint32_t                                m_param_l_max;
+
+    std::vector<std::string>                m_str_l;
+    uint32_t                                m_idx;
```
eric-wieser: Comments here too please!

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -956,11 +959,152 @@ static void add_module_constants(PyObject* simulator)
...
+
+    if (!PyArg_ParseTuple(args, "i", &id)) {
+        return NULL;
+    }
+
+	return PyTuple_Pack(3,
+			Py_BuildValue("s", cocotb_bfm_typename(id)),
+			Py_BuildValue("s", cocotb_bfm_instname(id)),
+			Py_BuildValue("s", cocotb_bfm_clsname(id))
+			);
```
eric-wieser: Safer as:
```
return Py_BuildValue("sss",
    cocotb_bfm_typename(id),
    cocotb_bfm_instname(id)),
    cocotb_bfm_clsname(id)
);
```
which handles errors correctly.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -956,11 +959,152 @@ static void add_module_constants(PyObject* simulator)
...
+	PyObject *param_l, *type_l;
+	cocotb_bfm_msg_param_t *paramv = 0;
+	int paramc = 0;
+	uint32_t i;
+
+	(void)param_l;
+	(void)type_l;
+
+	if (!PyArg_ParseTuple(args, "iiOO", &bfm_id, &msg_id, &param_l, &type_l)) {
+		return 0;
```
eric-wieser:
```suggestion
		return NULL;
```

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+        bfm_export_tasks += "    endtask\n"
+        
+    
+    impl_param_m = {
+        "bfm_classname" : info.T.__module__ + "." + info.T.__qualname__,
+        "bfm_import_calls" : bfm_import_calls,
+        "bfm_export_tasks" : bfm_export_tasks
+        }
+    
+    cocotb_bfm_api_impl = '''
```
cmarqu: Nit: use ``"""`` instead of ``'''`` (or run https://pypi.org/project/black/)

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+    import "DPI-C" context function void cocotb_bfm_add_si_param(int bfm_id, longint v);
+    import "DPI-C" context function void cocotb_bfm_add_ui_param(int bfm_id, longint unsigned v);
+    import "DPI-C" context function void cocotb_bfm_end_msg(int bfm_id);
+    
+    task automatic ${bfm_name}_process_msg();
+        int msg_id = cocotb_bfm_claim_msg(bfm_id);
+        case (msg_id)
+${bfm_import_calls}
+        default: begin
+            $display("Error: BFM %m received unsupported message with id %0d", msg_id);
```
cmarqu: I like to print the simulation time in such cases, but since you ``$finish`` anyway, it's going to be implicitly done for most if not all simulators.

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+   
+    filename_c = args.o
+    if filename_c.find('.') != -1:
+        filename_c = os.path.splitext(filename_c)[0]
+    filename_c += ".c"
+    
+    with open(args.o, "w") as out_sv:
+        with open(filename_c, "w") as out_c:
+            out_sv.write("//***************************************************************************\n")
+            out_sv.write("//* BFMs file for CocoTB. \n")
```
cmarqu: Lowercase ``cocotb``.

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+            out_sv.write("//* Note: This file is generated. Do Not Edit\n")
+            out_sv.write("//***************************************************************************\n")
+
+            out_c.write("//***************************************************************************\n")
+            out_c.write("//* BFMs DPI interface file for CocoTB. \n")
```
cmarqu: Lowercase ``cocotb``.

cocotb/bfmgen.py
```diff
@@ -0,0 +1,364 @@
...
+def main():
+    parser = argparse.ArgumentParser(prog="cocotb-bfmgen")
+    
+    subparser = parser.add_subparsers()
+    subparser.required = True
+    subparser.dest = 'command'
+    generate_cmd = subparser.add_parser("generate")
+    generate_cmd.set_defaults(func=bfm_generate)
+    generate_cmd.add_argument("-m", action='append')
+    generate_cmd.add_argument("-language", default="vlog")
```
cmarqu:
```suggestion
    generate_cmd.add_argument("--language", default="vlog")
```
(maybe support ``-l`` in addition)

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
...
+    @staticmethod
+    def call(
+            bfm_id, 
+            method_id,
+            params):
+        inst = BfmMgr.inst()
+        bfm = inst.bfm_l[bfm_id]
+         
+        if not hasattr(bfm, "bfm_info"):
+            raise Exception("BFM object does not contain 'bfm_info' field")
```
cmarqu: Maybe ``AttributeError`` would be better?

cocotb/share/include/cocotb_bfm_api.h
```diff
@@ -0,0 +1,118 @@
...
+// Callback function type used to notify
+// implementation when a message is available
+typedef void (*cocotb_bfm_notify_f)(void *);
+
+int cocotb_bfm_register(
+        const char                *type_name,
+        const char                *inst_name,
+        const char                *cls_name,
+        cocotb_bfm_notify_f        notify_f,
+        void                    *notify_data);
```
cmarqu: Nit: some small misalignment (from the tab cleanup?)

cocotb/share/lib/gpi/cocotb_bfm_api.cpp
```diff
@@ -0,0 +1,130 @@
...
+}
+
+void cocotb_bfm_add_si_param(uint32_t bfm_id, int64_t pval) {
+    GpiBfm *bfm = GpiBfm::get_bfms().at(bfm_id);
+    GpiBfmMsg *msg = bfm->active_inbound_msg();
+
+    if (msg) {
+        msg->add_param_si(pval);
+    } else {
+        fprintf(stdout, "Error: attempting to add an signed parameter to a NULL message\n");
```
cmarqu:
```suggestion
        fprintf(stdout, "Error: attempting to add a signed parameter to a NULL message\n");
```

cocotb/share/lib/vpi/vpi_bfm_api.cpp
```diff
@@ -0,0 +1,344 @@
...
+    // cocotb_bfm_register
+    tf_data.type = vpiSysFunc;
+    tf_data.tfname = "$cocotb_bfm_register";
+    tf_data.calltf = &cocotb_bfm_register_tf;
+    tf_data.compiletf = 0;
+    tf_data.sizetf = 0;
+    tf_data.user_data = 0;
+    vpi_register_systf(&tf_data);
+
+    // cocotb_bfm_claim_message
```
cmarqu:
```suggestion
    // cocotb_bfm_claim_msg
```

cocotb/share/lib/vpi/vpi_bfm_api.cpp
```diff
@@ -0,0 +1,344 @@
...
+    // cocotb_bfm_begin_msg
+    tf_data.type = vpiSysTask;
+    tf_data.tfname = "$cocotb_bfm_begin_msg";
+    tf_data.calltf = &cocotb_bfm_begin_msg_tf;
+    tf_data.compiletf = 0;
+    tf_data.sizetf = 0;
+    tf_data.user_data = 0;
+    vpi_register_systf(&tf_data);
+
+    // cocotb_bfm_add_ui_param
```
cmarqu:
```suggestion
    // cocotb_bfm_add_param_ui
```

cocotb/share/lib/vpi/vpi_bfm_api.cpp
```diff
@@ -0,0 +1,344 @@
...
+    tf_data.type = vpiSysTask;
+    tf_data.tfname = "$cocotb_bfm_add_param_ui";
+    tf_data.calltf = &cocotb_bfm_add_param_ui_tf;
+    tf_data.compiletf = 0;
+    tf_data.sizetf = 0;
+    tf_data.user_data = 0;
+    vpi_register_systf(&tf_data);
+
+    // cocotb_bfm_add_si_param
```
cmarqu:
```suggestion
    // cocotb_bfm_add_param_si
```

cocotb/share/lib/vpi/vpi_bfm_api.cpp
```diff
@@ -0,0 +1,344 @@
...
+    tf_data.type = vpiSysTask;
+    tf_data.tfname = "$cocotb_bfm_add_param_si";
+    tf_data.calltf = &cocotb_bfm_add_param_si_tf;
+    tf_data.compiletf = 0;
+    tf_data.sizetf = 0;
+    tf_data.user_data = 0;
+    vpi_register_systf(&tf_data);
+
+    // cocotb_bfm_add_str_param
```
cmarqu:
```suggestion
    // cocotb_bfm_add_param_str
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,196 @@
...
+            @cocotb.bfm_import(cocotb.bfm_uint64_t)
+            def write_req(self, d):
+                pass
+    
+            @cocotb.bfm_export()
+            def write_ack(self):
+                self.ack_ev.set()
+
+Python methods that will result in task calls in the HDL are 
+decorated with the @cocotb.bfm_import decorator, while
```
cmarqu:
```suggestion
decorated with the :class:`cocotb.bfm_import` decorator, while 
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,196 @@
...
+                pass
+    
+            @cocotb.bfm_export()
+            def write_ack(self):
+                self.ack_ev.set()
+
+Python methods that will result in task calls in the HDL are 
+decorated with the @cocotb.bfm_import decorator, while 
+Python methods that will be called from the HDL are decorated
+with the @cocotb.bfm_export decorator.
```
cmarqu:
```suggestion
with the :class:`cocotb.bfm_export` decorator. 
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,196 @@
...
+BFM API.
+
+The :command:`cocotb-bfmgen` script generates the appropriate BFM interface
+files based on the BFMs required for a given testbench.
+
+The ``cocotb-bfmgen`` script accepts the following options:
+
+- -m <module> -- Specifies a Python module to load. Typically, this will
+  be a BFM package.
+- -language <target> -- Specifies the target testbench language. ``vlog`` and ``sv``
```
cmarqu:
```suggestion
- --language <target> -- Specifies the target testbench language. ``vlog`` and ``sv`` 
```
(if you do that change in the argparse setup)

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,196 @@
...
+files based on the BFMs required for a given testbench.
+
+The ``cocotb-bfmgen`` script accepts the following options:
+
+- -m <module> -- Specifies a Python module to load. Typically, this will
+  be a BFM package.
+- -language <target> -- Specifies the target testbench language. ``vlog`` and ``sv`` 
+  are currently accepted.
+- -o <file> -- Specifies the output file. By default, the name will 
+  be cocotb_bfms.v.
```
cmarqu:
```suggestion
  be ``cocotb_bfms.v``.
```

tests/designs/rv_bfm/rv_bfm_top.sv
```diff
@@ -0,0 +1,101 @@
+//-----------------------------------------------------------------------------
```
cmarqu: Your SPDX replacer missed this :)

cocotb/regression.py
mballance: Hmm... that's a good idea. Let me have a look to see how feasible it would be to get sim-info at this point

tests/designs/rv_bfm/rv_bfm_top.sv
mballance: Good catch!

documentation/source/task_bfms.rst
mballance: Good point. I've added short and long options and updated the docs.

cocotb/bfmgen.py
mballance: Done. I've added both a short (-l) and long (--language) version of the option

mballance:
@eric-wieser, I do agree about #767 since there are a couple of Python3 items in the code (eg __qualname__). Hopefully that's not too much of an obstacle, given that Python2 is EOL at the end of 2019. I'm not as clear on the impact of #1195 and the timeline for getting that resolved. Any insights there?

cocotb/bfmgen.py
mballance: Done

cocotb/share/lib/simulator/simulatormodule.c
mballance: Appreciate the tip! Done.

cocotb/share/lib/gpi/GpiBfm.h
mballance: Done. I've added comments on both the Bfm and BfmMsg classes

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,196 @@
...
+BFM API.
+
+The :command:`cocotb-bfmgen` script generates the appropriate BFM interface
+files based on the BFMs required for a given testbench.
+
+The ``cocotb-bfmgen`` script accepts the following options:
+
+- -m <module> -- Specifies a Python module to load. Typically, this will
+  be a BFM package.
+- -l,--language <target> -- Specifies the target testbench language. ``vlog`` and ``sv``
```
cmarqu:
```suggestion
- ``-l``, ``--language <target>`` -- Specifies the target testbench language. ``vlog`` and ``sv`` 
```
otherwise Sphinx will render the ``--`` as an em dash (https://external-builds.readthedocs.io/html/cocotb/1217/task_bfms.html#manually-generating-bfm-interface-files)
The other options should get the same treatment if only for consistency.

cocotb/share/lib/gpi/GpiBfm.h
cmarqu: Ironic how the private members are now better documented than the public functions :)
https://external-builds.readthedocs.io/html/cocotb/1217/generated/file/GpiBfm_8h.html
I wish all our code was documented like this!

cocotb/regression.py
mballance: I've changed this to check the SIM_NAME (VPI 'product') and report the current note if the simulator is Verilator, and retain the previous exception if not.

cocotb/bfms.py
eric-wieser: No longer relevant, 2.7 is no longer supported in master

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
...
+            clsname = info[1]
+            if clsname.find('.') == -1:
+                raise Exception("Incorrectly-formatted BFM class name {!r}".format(clsname))
+            pkgname = clsname[:clsname.rfind('.')]
+            clsleaf = clsname[clsname.rfind('.')+1:]
+          
+            try: 
+                pkg = importlib.import_module(pkgname)
+            except Exception as e:
+                raise Exception("Failed to import BFM package \"" + pkgname + "\"")
```
eric-wieser:
```suggestion
                raise Exception("Failed to import BFM package {!r}".format(pkgname))
```
will add the quotes for you. Same throughout

cocotb/bfmgen.py
mballance: Hmm... The caller is the bfmgen command-line script. I suppose we could just let the exception propagate back up, but I was concerned that the user wouldn't get appropriate information about which module had failed to load. Happy to change this if you feel the default Python exception trace will be better.

cocotb/bfmgen.py
```diff
@@ -0,0 +1,370 @@
...
+    parser = argparse.ArgumentParser(prog="cocotb-bfmgen")
+    
+    subparser = parser.add_subparsers()
+    subparser.required = True
+    subparser.dest = 'command'
+    generate_cmd = subparser.add_parser("generate")
+    generate_cmd.set_defaults(func=bfm_generate)
+    generate_cmd.add_argument("-m", action='append')
+    generate_cmd.add_argument("-l", "--language", default="vlog")
+    generate_cmd.add_argument("-o", default=None)
```
eric-wieser: Can you move this to a `def get_parser()` that returns `parser`? If so, you can generate documentation from the `ArgumentParser` object by using `sphinx-argparse`. I'll try to put a PR together with an example of that...

cocotb/bfmgen.py
mballance: Done!

cocotb/bfmgen.py
mballance: Cleaned up indirectly

cocotb/bfmgen.py
mballance: This looks to be out-of-date. The suggested change is the same as the current code.

cocotb/bfmgen.py
eric-wieser: (Done in #1304, although that will need to be merged first)

cocotb/bfms.py
mballance: It might make things simpler. What's the timeline for resolving #767? I'd hate to leave this PR in limbo, but if #767 is close to being resolved I can wait.

cocotb/bfmgen.py
eric-wieser: Doesn't the exception information tell them which one failed anyway?

cocotb/bfmgen.py
mballance: Done.

cocotb/bfmgen.py
mballance: Moved to using dict()

cocotb/bfms.py
eric-wieser: It's already been resolved, so this is no longer an issue :)

cocotb/bfmgen.py
mballance: Decided to omit this for now.

eric-wieser:
CI is failing because some files have trailing spaces.

cocotb/bfmgen.py
mballance: Done

cocotb/share/include/cocotb_bfm_api.h
mballance: I've moved to single-line prototypes where feasible, and cleaned up the post-tab-replacement indent issues.

cocotb/share/include/cocotb_bfm_api.h
mballance: Done

cocotb/bfms.py
mballance: Nice!

cocotb/bfms.py
mballance: Like it!

cocotb/bfms.py
mballance: Done

cocotb/bfms.py
mballance: Didn't realize you could have an 'else' clause on an try/except block. Like it.

cocotb/bfms.py
mballance: Done

cocotb/bfms.py
mballance: Makes sense. Done

cocotb/decorators.py
mballance: Good catch. Corrected

cocotb/bfmgen.py
mballance: Okay, makes sense. I've added the Python-side changes. Looks like #1304 adds some boilerplate changes, as well as a doc changes that depends on those boilerplate changes.

eric-wieser: You can add the rst file change in this PR today, it will just be ignored until the boilerplate is merged too.

cocotb/bfmgen.py
mballance: It looks like this:
```
  File "/usr/lib64/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/project/fun/cocotb/cocotb-task-bfms/cocotb/bfmgen.py", line 354, in <module>
    main()
  File "/project/fun/cocotb/cocotb-task-bfms/cocotb/bfmgen.py", line 348, in main
    bfm_load_modules(args.m)
  File "/project/fun/cocotb/cocotb-task-bfms/cocotb/bfmgen.py", line 27, in bfm_load_modules
    importlib.import_module(m)
  File "/usr/lib64/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'foobar'
```
So, yes, the missing module is there at the bottom. I'm concerned that it's unclear whether this is a user error (ie provoked by user input) or a system error (ie provoked by a bad installation). I'm find leaving it this way pending feedback if that's what seems to work best.

mballance:
Hi @eric-wieser and @cmarqu, I just updated this branch with the latest changes in master, and I believe we've resolved the open issues (of course, let me know if that's not the case). Any advice on the next steps in order to get this merged into cocotb master?

Thanks,
Matthew

cocotb/bfms.py
```diff
@@ -0,0 +1,206 @@
...
+        self.bfm_type_info_m[T] = type_info
+
+    @staticmethod
+    def get_bfms():
+        return BfmMgr.inst().bfm_l
+
+    @staticmethod
+    def find_bfm(path_pattern):
+        inst = BfmMgr.inst()
+        bfm = None
```
eric-wieser: I don't think these should be static methods - can you just make them regular methods, replace `inst` with `self`, and have the caller use `BgmMgr.inst().find_bfm(...)` instead?

cocotb/decorators.py
```diff
@@ -490,3 +490,95 @@ def f(*args, **kwargs):
...
+
+# Constants for use in specifying BFM API signatures
+bfm_int8_t = bfm_param_int_t(8, True)
+bfm_uint8_t = bfm_param_int_t(8, False)
+bfm_int16_t = bfm_param_int_t(16, True)
+bfm_uint16_t = bfm_param_int_t(16, False)
+bfm_int32_t = bfm_param_int_t(32, True)
+bfm_uint32_t = bfm_param_int_t(32, False)
+bfm_int64_t = bfm_param_int_t(64, True)
+bfm_uint64_t = bfm_param_int_t(64, False)
```
eric-wieser: Would it make sense to use the builtin `ctypes` constants (`ctypes.c_uint8` etc) here?

documentation/source/library_reference.rst
```diff
@@ -300,6 +300,13 @@ The Scheduler
     :members:
     :member-order: bysource
 
+The ``cocotb-bfmgen`` script
+----------------------------
+
+.. argparse::
+     :module: cocotb.bfmgen
+     :func: get_parser
+     :prog: cocotb-bfmgen
```
eric-wieser: This probably makes sense inside your main `rst` document

cocotb/decorators.py
mballance: I considered doing that. BFM task parameters support a subset of the types supported by ctypes, so I decided to declare dedicated types for the BFM task parameters.
If you feel it's cleaner, we could have the user specify ctypes types and document which ones are supported (and, of course, throw an exception if an unsupported type is used). Let me know if you'd prefer to go this direction.

eric-wieser: Probably fine as is, just wanted to check you were aware of ctypes.

cocotb/bfmgen.py
```diff
@@ -0,0 +1,346 @@
...
+    parser = argparse.ArgumentParser(prog="cocotb-bfmgen")
+
+    subparser = parser.add_subparsers()
+    subparser.required = True
+    subparser.dest = 'command'
+    generate_cmd = subparser.add_parser("generate")
+    generate_cmd.set_defaults(func=bfm_generate)
+    generate_cmd.add_argument("-m", action='append')
+    generate_cmd.add_argument("-l", "--language", default="vlog")
+    generate_cmd.add_argument("-o", default=None)
```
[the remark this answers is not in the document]
eric-wieser: You can just lift the text from your rst file here :)

cocotb/decorators.py
```diff
@@ -490,3 +490,95 @@ def f(*args, **kwargs):
...
+        self.s = s
+
+    def sv_type(self):
+        if self.w in bfm_param_int_t.sv_type_m.keys():
+            ret = bfm_param_int_t.sv_type_m[self.w]
+            if not self.s:
+                ret += " unsigned"
+            return ret
+        else:
+            raise Exception("parameter-width \"" + str(self.w) + "\" not supported by SystemVerilog")
```
```suggestion
            raise ValueError("parameter-width {} not supported by SystemVerilog".format(self.w))
```

documentation/source/task_bfms.rst
```diff
@@ -0,0 +1,196 @@
...
+The Python aspect of a BFM is captured as a Python class with the :class:`cocotb.bfm`
+decorator. The Python class provides both the user-facing and implementation
+API. In addition to identifying the class as a BFM class, the @cocotb.bfm
+decorator associates HDL template files with the BFM class.
+
+.. code-block:: python3
+
+        @cocotb.bfm(hdl={
+            cocotb.bfm_vlog : cocotb.bfm_hdl_path(__file__, "hdl/rv_data_out_bfm.v"),
+            cocotb.bfm_sv   : cocotb.bfm_hdl_path(__file__, "hdl/rv_data_out_bfm.v")
```
eric-wieser: Are these two filenames deliberately the same?

cocotb/bfmgen.py
```diff
@@ -0,0 +1,346 @@
...
+            bfm_export_tasks += ";\n"
+        bfm_export_tasks += "    begin\n"
+        bfm_export_tasks += "        $cocotb_bfm_begin_msg(bfm_id, " + str(i) + ");\n"
+        for p in exp.signature:
+            if p.ptype.s:
+                bfm_export_tasks += "        $cocotb_bfm_add_param_si(bfm_id, " + p.pname + ");\n"
+            else:
+                bfm_export_tasks += "        $cocotb_bfm_add_param_ui(bfm_id, " + p.pname + ");\n"
+
+        bfm_export_tasks += "        $cocotb_bfm_end_msg(bfm_id);\n"
```
eric-wieser: It's tempting to make this a single call, `$cocotb_bfm_send_msg(bfm_id, i, p1, p2, p3, ...)`, as:

```
        bfm_export_tasks += "        $cocotb_bfm_begin_msg(bfm_id, {}{});\n".format(
            i,
            "".join(", {}".format(p.pname)) for p in exp.signature)
        )
```
and have `$cocotb_bfm_begin_msg` work out the signed vs unsigned from the value of `i`.

cmarqu:
> Any advice on the next steps in order to get this merged into cocotb master?

This could be one of those cases where we would like to offer the feature as an integral part of cocotb but mark it as "experimental"/"beta"/"preview" (or some better word) to express that API breaking changes due to user feedback can happen. 
Such PRs are bound to occur more often now that we have nailed the basics, so maybe we need a process. Worth a new issue to discuss?
(@imphil, @eric-wieser, @themperek)

garmin-mjames:
Wouldn't this make more sense to be a cocotb extension?

eric-wieser:
I think since this touches the vpi interface, at least part of it will probably end up within cocotb.

garmin-mjames:
I think you could build separate libraries linked against whichever `cocotb` libs are necessary, and add a separate extension to get access from Python. Then use `GPI_EXTRA` to call `register_bfm_tf()` if that needs to be done at startup.

mballance:
All,
    I appreciate the feedback, and discussion about the best way to get a new feature like this into cocotb proper. As @eric-wieser points out, this feature is a bit different than some others in that it touches the simulator (VPI/DPI/VHPI/etc) integration, and benefits from being integrated with the existing C/C++ simulator-integration codebase.
    I'd like to understand a bit more about the suggestion to make this an extension. Is this a path that has been successfully used in the past to trial new features that eventually are incorporated into the cocotb core? If so, I'd be happy to learn what I can from the process of trialing and adding that feature. 

Best Regards,
Matthew

ktbarrett:
It would be new territory making this an external module, but it might be useful to make that interface *now* considering this change is pretty large, only works with one interface (VPI), and not everyone will use it, which is the perfect justification for an extension. Though I think this should probably be an "official" extension that lives in the cocotb org so we can continue to maintain it.

This change adds a bunch of new methods to the C/C++ codebase, but they seem orthogonal to it, meaning it could be moved into a separate set of sources built independently of cocotb. The only real feature we need from the existing code base to support this is the ability to late load extensions to run during `vlog_startup_rountines`.

Instead of directly registering the function as is done here, we could provide an extension interface, late load any required extensions specified via environment variable, and call into entry functions on those extensions in an extension loader routine we register in the startup function tables. An instead of modifying the simulator module, those functions would be moved into another Python extension module.

I'll look into this a bit more and am willing to help get this working.

themperek:
I am very in favour of extension (if possible). Other way is very hard (if not impossible) to maintain in the current situation.  Should try to keep cocotb core as small as possible.

mballance:
@ktbarrett. I've been looking at the extension-module route as well. I don't see any technical reason that this specific functionality couldn't be supported as an extension module that is mostly-independent of cocotb. With an extension-module approach (whether tightly or loosely bound to cocotb), I think we would be asking users to take action in order to enable support for task-based BFMs. As an independent extension, we would ask them to load an addition VPI/FLI/VHPI library. As a cocotb extension, I suppose we would ask them to load an additional cocotb extension library (ie provide a hook for cocotb to initialize the extension). 
Perhaps targeting an independent extension module is the right route for development of functionality that is likely to be under active development for some time. In other words, develop "significant" new functionality as an independent extension module. Once it has mileage on it, look at incorporating it into cocotb proper. Does this sound like a path that could be recommended going forward?

ktbarrett:
Do we have to register the system functions during startup? Or can we load them later?

If the extension module is a built as a Python package that is installed to the system we could simply have the system tasks register when `import`ing the extension module. No need for special extension support at all. And this would have to be the case for all the interfaces for when we support them with the GPI BFM interface as well.

mballance:
Yes, system functions need to be registered during startup. Simulators complain (and fail) if system functions can't be located during elab ;-).

So, as an independent extension, the PLI library would need to be given to the simulator so the simulator could call the 'register_tf' method, and loaded as a Python extension. It could later be integrated with cocotb by having the VPI/GPI initialization call the BFM library's 'register_tf' method. 

Okay, so I guess that's the recommendation?

eric-wieser:
One thing that would be really cool is if cocotb core provided a way to register python functions as `$sys_funcs` in verilog.

> Simulators complain (and fail) if system functions can't be located during elab 

During elaboration, what's actually happened so far? Have we had a chance to load the python interpreter already?

garmin-mjames:
Libs in `GPI_EXTRA` are loaded at startup, although there are restrictions on the expected name of the lib and entry function.

imphil:
Thanks @mballance for your work on this. I looked through the docs and had an initial look at the implementation. A couple comments:

On the general approach:

- I'd love to have this functionality as part of core cocotb. 
  - It logically extends what cocotb does to a higher level and fits in nicely.
  - It gives us a sustainable path forward to better support tools like Verilator which have a rather slow VPI interface (by design), but a very efficient DPI interface.
  - Personally, I write a fair amount of DPI C code for various bus interfaces, doing so in cocotb would be fantastic.
- You currently support "Verilog" and "SystemVerilog" (DPI). What's the point of the "Verilog" variant? It's 2020, after all, are there still simulators without SystemVerilog DPI support?
- Please try to rebase this PR into a commit structure that makes it easier to review. That would be one commit per self-contained change, and a commit message with it describing roughly what the commit does (and why). Please reach out if you need help with the right git commands to get that done.


I'll probably have more comments on API and structure once we've sorted these initial things out to make it easier to review.

mballance:
Thank you for the feedback, @imphil. A couple of comments in response:
- There does seem to be some disagreement about whether this should be a core-cocotb feature or an extension. While that discussion is ongoing, I've created a standalone extension (https://github.com/pybfms/pybfms). Regardless of what the short-term looks like, my long-term preference reference would be to have this a part of cocotb proper. 
- My belief is that all commercial tools support DPI at this point. On the open-source front, however, Icarus Verilog does not. Icarus seems sufficiently popular that I didn't want to preclude Icarus users from getting performance benefits from task-based BFMs.
- I could use some pointers on how to break this down into multiple PRs. Do you have something in mind like:
  - Core VPI/DPI API implementation (basically all the C/C++ stuff) and the Python interface
  - BFM shell generation command

ktbarrett:
@mballance I am working on a number of things, including adding support for GPI extensions in #1293. I will document it thoroughly soon, but to summarize:

1.  Create a library of desired name (example will be "foo"): `libfoo.so`
2.  Library is required to have a function at a known location: `foo_entry_point`. This function will use function added to the GPI to register extension init functions and other required callbacks.
  a. `gpi_register_extension`: takes a function that looks like a main: `int func(int argc, char** argv)`
  b. `gpi_register_sim_event_callback`: takes a function to call when a sim event occurs (like premature shutdown)
  c. `gpi_register_sim_end_callback`: takes a function to call when the simulator is shutting down so you can perform final cleanup.
3.   Add `"foo"` to the `GPI_EXTRA` envvar when you run cocotb, ensuring the library is in the `LD_LIBRARY_PATH`
4. cocotb will load the extension for you

And in your case, you will need a separate Python extension library, much like the `simulator` module. Is this sufficient to implement your extension?

Alex-Mann:
I'm wondering if/what the status is on this task and there is anything I can do to help out to enable this functionality, even if under a beta/preview/prerelease tag. Please let me know.

mballance:
Hi @Alex-Mann,
    The consensus that eventually emerged was that this would be best-implemented as an extension to cocotb, since that allows core cocotb to stay focused. This functionality is now developed as 'PyBFMs' here:
https://github.com/pybfms/pybfms/
    If you're interested in either using the PyBFMs or contributing to its development, I'd be happy to have your involvement.

Best Regards,
Matthew

ktbarrett:
@mballance AFAIK this is done in an external library now. Could you post that here FFR?

Alex-Mann:
The MR showing the example is here: https://github.com/fvutils/pyhdl-if/pull/5

Issue #1220, opened 2019-12-03
Issue use handle.drivers() in Python 3.7+


ktbarrett:
This uses a `next`-like function, but it's defined in C. See it catches `GeneratorExit`.

https://github.com/cocotb/cocotb/blob/17ce4e51da43b6d2d1b5834d629374ed5535a8d8/cocotb/handle.py#L572-L580

The C however throws `StopIteration`. This will bubble up instead of stop the generator in Python 3.7+.

https://github.com/cocotb/cocotb/blob/17ce4e51da43b6d2d1b5834d629374ed5535a8d8/cocotb/share/lib/simulator/simulatormodule.c#L528-L556

I will make a test for this soon.

eric-wieser:
You're right, that looks like nonsense to me.

Introduced in https://github.com/cocotb/cocotb/commit/1719e9875168e1ecb76d0e5e59710576c2b123c7

eric-wieser:
Fixed in #1221

PR #1221, opened 2019-12-03
Handle StopIteration in iteration methods in handles

ktbarrett:
As mentioned in #1220 , a potentially problematic commit was reverted, and methods that wrap `simulator.next` catch the `StopIteration` it throws to end iteration.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -31,6 +31,11 @@
 
 import ctypes
 import warnings
+import sys
+if sys.version_info.major < 3:
+    import collections as collections_abc
+else:
+    import collections.abc as collections_abc
 
 import os
 
@@ -548,6 +553,19 @@ class NonHierarchyIndexableObject(NonHierarchyObject):
                 left = left + 1
 
 
+class _SimIterator(collections_abc.Iterator):
+    """Iterator over simulator objects. For internal use only."""
+
+    def __init__(self, handle, mode):
+        self._iter = simulator.iterate(handle, mode)
+
+    def __next__(self):
+        return simulator.next(self._iter)
+
+    if sys.version_info.major < 3:
+        next = __next__
+
+
 class NonConstantObject(NonHierarchyIndexableObject):
     # FIXME: what is the difference to ModifiableObject? Explain in docstring.
 
@@ -562,17 +580,11 @@ class NonConstantObject(NonHierarchyIndexableObject):
 
     def drivers(self):
         """An iterator for gathering all drivers for a signal."""
-        iterator = simulator.iterate(self._handle, simulator.DRIVERS)
-        while True:
-            # Path is left as the default None since handles are not derived from the hierarchy
-            yield SimHandle(simulator.next(iterator))
+        return _SimIterator(self._handle, simulator.DRIVERS)
 
     def loads(self):
         """An iterator for gathering all loads on a signal."""
-        iterator = simulator.iterate(self._handle, simulator.LOADS)
-        while True:
-            # Path is left as the default None since handles are not derived from the hierarchy
-            yield SimHandle(simulator.next(iterator))
+        return _SimIterator(self._handle, simulator.LOADS)
 
 
 class ModifiableObject(NonConstantObject):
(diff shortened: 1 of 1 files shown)


cocotb/handle.py
```diff
@@ -526,16 +526,14 @@ def __getitem__(self, index):
         return self._sub_handles[index]
 
     def __iter__(self):
-        if self._range is None:
-            raise StopIteration
```
eric-wieser: Not: tempting to leave this if at the start to minimize the diff

cocotb/handle.py
```diff
@@ -565,14 +563,20 @@ def drivers(self):
...
     def loads(self):
         """An iterator for gathering all loads on a signal."""
         iterator = simulator.iterate(self._handle, simulator.LOADS)
         while True:
             # Path is left as the default None since handles are not derived from the hierarchy
-            yield SimHandle(simulator.next(iterator))
+            try:
+                yield SimHandle(simulator.next(iterator))
+            except StopIteration:
+                return
```
eric-wieser: Strictly this would be better as:
```
try:
    v = simulator.next(iterator)
except StopIteration:
    return
yield SimHandle(v)
```
as this doesn't swallow unexpected exceptions from `SimHandle` or `yield`. Same throughout.

A possibly neater option would be
```python
class SimIterator(collections.abc.Iterator):
    def __init__(self, handle, mode):
        self._iterator = simulator.iterate(handle, mode)
    def __next__(self):
        return simulator.next(self._iterator)
    if sys.version_info.major < 3:
        next = __next__
```

making this function simply `return SimIterator(self._handle, simulator.LOADS)`

ktbarrett: The class is one more thing to document that's only being used in 2 places. But you are definitely right about the first bit.

ktbarrett:
Will add a test for both `drivers` and `loads` to `test_cocotb`. Internal testing proves the solution works.

cocotb/handle.py
eric-wieser: It wouldn't need documenting - it's a private type just like the built-in "dict_iterator" and "list_iterator".

eric-wieser: Could call it `_SimIterator` to emphasie that

ktbarrett:
Icarus does not like `drivers()` or `loads()`. It's probably unimplemented. There is a test for `drivers()` in test_iteration_mixedlang which did not fail after the previous update when running Questa.

cocotb/handle.py
```diff
@@ -529,33 +528,40 @@ def __getitem__(self, index):
...
-                    result = self[i]
-                    yield result
-                except IndexError:
-                    continue
-        except GeneratorExit:
-            pass
+        if self._range is None:
+            return
 
+        self._log.debug("Iterating with range [%d:%d]" % (self._range[0], self._range[1]))
```
eric-wieser: This was actually more correct before - it saves formatting the string unless the log is actually printed

cocotb/handle.py
```diff
@@ -31,6 +31,8 @@
 
 import ctypes
 import warnings
+import sys
+import collections.abc
```
```suggestion
if sys.version_info.major < 3:
    import collections as collections_abc
else:
    import collections.abc as collections_abc
```

cocotb/handle.py
```diff
@@ -529,33 +528,40 @@ def __getitem__(self, index):
...
+            while left >= right:
+                yield left
+                left = left - 1
+        else:
+            while left <= right:
+                yield left
+                left = left + 1
+
+
+class _SimIterator(collections.abc.Iterator):
```
```suggestion
class _SimIterator(collections_abc.Iterator):
```

eric-wieser:
That's a shame that we can't test it in CI. Thanks for confirming, anyway.

Ci failure is a python 2 thing - `collections.abc.Iterator` moved between versions.

Either fix the imports as I suggest above, or add `def __iter__(self): return self`

cocotb/handle.py
ktbarrett: Yeah, just a byproduct of the revert I didn't see.

PR #1223, opened 2019-12-04
object.__getattr__ does not exist

ktbarrett:
If accessing a non-existent attribute you get an error about `object` not having an attribute `__getattr__` instead of the attribute you actually tried to incorrectly use. Using `__getattribute__` instead should work.


eric-wieser:
Can you add a test? Something like:
```
try:
    obj.illegal_attr
except AttributeError as e:
    assert 'illegal_attr' in str(e)
else:
    assert False, "Expected an attribute error"
```

PR #1224, opened 2019-12-04
PR: Added Logging to File.

jahagirdar:
This PR adds file logging with log file rotation.

Most of the times while debugging I prefer opening a log file in my editor instead of searching the (truncated) console log.
The immediate need for this changes was:

I am planning to run a week long testcase (1000*64000 packets) and If it fails at anypoint I need the detailed log at the point of failure which means every driver and monitor printing debug information. This has a potential to fillup the diskspace hence adding file logging + log rotation.

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -34,6 +34,8 @@ import sys
 import logging
 import warnings
 
+from logging.handlers import RotatingFileHandler
+
 from cocotb.utils import get_sim_time, want_color_output
 
 import cocotb.ANSI as ANSI
@@ -67,6 +69,15 @@ class SimBaseLog(logging.getLoggerClass()):
 
         self.propagate = False
         self.addHandler(hdlr)
+
+        want_filelogger=os.getenv("COCOTB_FILE_LOGGER")
+        if want_filelogger is not None:
+            logfile=os.path.join(os.getenv("RESULT_PATH"),"results.log")
+            file_handler = RotatingFileHandler(
+                        logfile, maxBytes=(1048576*5), backupCount=4
+                        )
+            file_handler.setFormatter(SimLogFormatter())
+            self.addHandler(file_handler)
 
     def _logFromC(self, level, filename, lineno, msg, function):
         """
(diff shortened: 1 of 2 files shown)


eric-wieser:
For your own use, you can enable this without having to modify cocob, simply by adding this somewhere in your code:
```python
@logging.setLoggerClass
class MyLog(logging.getLoggerClass()):
    def __init__(self, name):
        super().__init__(name)
        logfile = os.path.join(os.getenv("RESULT_PATH"), "results.log")
        file_handler = RotatingFileHandler(
            logfile, maxBytes=(1048576*5), backupCount=4)
        file_handler.setFormatter(cocotb.log.SimLogFormatter())
        self.addHandler(file_handler)
```

jahagirdar:
I am already using a `configparser` based configuration file in my test environment which I am using to configure the environment behavior at the project, testsuite, and testcase level.

It can be generalized and loaded via COCOTB_CONFIG_FILE to to control the various Env and Make Options that we are now using...

Doing this will have a wide impact as almost every module which uses the configuration variables will be touched. I will open a new issue and check for feedback before starting the effort.

I will wait for #1212 status to change to make further changes to this PR.

eric-wieser:
Once #1266 goes in, I think the way to handle this is:
```python
root_logger = logging.getLogger()

# undo the setup cocotb did
for handler in root_logger.handlers:
    root_logger.remove_handler(handler)

file_handler = RotatingFileHandler(
    logfile, maxBytes=(1048576*5), backupCount=4)
file_handler.setFormatter(cocotb.log.SimLogFormatter())
root_logger.addHandler(file_handler)
```
at which point this seems like something that doesn't need to be in cocotb itself.

imphil:
Thanks for this PR @jahagirdar. Taking your use case into consideration, we were able to redesign the logging subsystem so that your use case is now supported without modifying cocotb itself. I've opened  #1400 to add documentation for this use case, so that other users trying to do the same can learn from that example.

Again, thanks for improving cocotb, and please feel free to re-open this PR or open a new one if your use case isn't fully supported yet.

cmarqu:
@jahagirdar This is a PR for the docs now: #1732

PR #1229, opened 2019-12-04
Ensure handle docstrings are inherited

eric-wieser:
By putting the argument docstrings in `__init__`, we ensure they are inherited automatically by subclasses with either:

* Do not define `__init__`
* Define `__init__` with no docstring

By clearing up the inconsistency, this also fixes a rendering bug in the sphinx docs where some argument lists appeared twice:

> image

This also adds some blank lines to appease flake8 a little more.

cc @cmarqu, alternative to parts of #1205

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -67,6 +67,8 @@ class SimHandleBase(object):
 
     def __init__(self, handle, path):
         """
+        .. Constructor. This RST comment works around sphinx-doc/sphinx#6885
+
         Args:
             handle (int): The GPI handle to the simulator object.
             path (str): Path to this handle, ``None`` if root.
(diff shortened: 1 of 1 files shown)


eric-wieser:
This doesn't work unfortunately, due to https://github.com/sphinx-doc/sphinx/pull/6885

cocotb/handle.py
```diff
@@ -72,6 +72,8 @@ class SimHandleBase(object):
 
     def __init__(self, handle, path):
         """
+        .. Constructor. This RST comment works around sphinx-doc/sphinx#6885
+
```
eric-wieser: Rendering fixed by adding this line. Sphinx otherwise strips spaces for inherited (but not direct) docstrings, which breaks autodoc

By adding a comment here, the stripped spaces don't matter.

PR #1230, opened 2019-12-05
Allow users to override the cocotb logger

ktbarrett:
To ensure logging customization occurs before base loggers are made, logging customization must occur right after cocotb module entry. This provides the infrastructure to allow the user to do logger customization, while preserving current functionality.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -39,7 +39,12 @@ import time
 
 import cocotb.handle
 from cocotb.scheduler import Scheduler
-from cocotb.log import SimBaseLog, SimLog
+try:
+    import cocotb_custom.log
+except ImportError:
+    from cocotb.log import SimLog
+else:
+    SimLog = cocotb_custom.log.SimLog
 from cocotb.regression import RegressionManager
 
 
@@ -53,8 +58,6 @@ from cocotb.decorators import test, coroutine, hook, function, external  # noqa:
 
 from ._version import __version__
 
-logging.basicConfig()
-logging.setLoggerClass(SimBaseLog)
 log = SimLog('cocotb')
 level = os.getenv("COCOTB_LOG_LEVEL", "INFO")
 try:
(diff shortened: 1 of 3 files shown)


cocotb/__init__.py
```diff
@@ -53,20 +56,21 @@
 
 from ._version import __version__
 
+logging.basicConfig()
```
ktbarrett: Moving this outside allows me to use the cocotb logger in an interactive python session.

cocotb/utils.py
```diff
@@ -75,6 +75,13 @@ def get_sim_time(units=None):
     return result
 
 
+if "COCOTB_SIM" not in os.environ:
+
+    @functools.wraps(get_sim_time)
+    def get_sim_time(units=None):
+        return 0
```
ktbarrett: This is required for the cocotb logger to be used interactively.

cocotb/__init__.py
```diff
@@ -39,7 +39,10 @@
 
 import cocotb.handle
 from cocotb.scheduler import Scheduler
-from cocotb.log import SimBaseLog, SimLog
+try:
+    from cocotb_custom.log import SimBaseLog, SimLog
+except ModuleNotFoundError:
+    from cocotb.log import SimBaseLog, SimLog
```
ktbarrett: Tries loading user code from a known location. Requires no more environment variables. User required to add the correct path to their PYTHONPATH. The `cocotb_custom` namespace can be reserved for future use in the project for user customization points. This parallels the consensus of `cocotb_ext` or `cocotbext` for the naming of the cocotb extension module namespace.

ktbarrett:
Test using this feature will come tomorrow.

eric-wieser:
To state my preference now: I'm not a massive fan of this approach. Thanks for writing it up though, will be useful to compare to other approaches.

cocotb/__init__.py
eric-wieser: This sounds like something that should probably be disabled by default for interactive sessions - perhaps `cocotb.log.setup()` or something would enable it, which would also achieve the goal of moving stuff out of `__init__.py`. That might be good as a standalone cleanup PR.

PR #1232, opened 2019-12-08
Show inheritance for classes in handle.py

cmarqu:
Visualizing the class hierarchy in ``handle.py`` is quite useful as can be seen in https://external-builds.readthedocs.io/html/cocotb/1232/library_reference.html#simulation-object-handles

### DIFF
--- a/cocotb/bus.py
+++ b/cocotb/bus.py
@@ -30,7 +30,7 @@
 """Common bus related functionality.
 A bus is simply defined as a collection of signals.
 """
-from cocotb.handle import AssignmentResult
+from cocotb.handle import _AssignmentResult
 
 def _build_sig_attr_dict(signals):
     if isinstance(signals, dict):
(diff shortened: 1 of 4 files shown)


eric-wieser:
`AssignmentResult` looks somewhat out of place there

cmarqu:
Yeah... There is apparently no way to selectively exclude a class from the diagram, and the alternative is to list the leaf classes which then becomes a maintenance burden.
Shouldn't it be ``_AssignmentResult`` anyway? Then it maybe wouldn't show up by default.

eric-wieser:
I was just thinking that - let's add the leading underscore.

cmarqu:
> I was just thinking that - let's add the leading underscore.

Done; the graph looks good now.

eric-wieser:
Fantastic, thanks!

A little annoying that there's a white background on the SVG image. It's tempting to try and fix that upstream...

eric-wieser:
PR at https://github.com/sphinx-doc/sphinx/pull/6910

PR #1237, opened 2019-12-13
Add logical to phyiscal name mapping for buses.

grebe:
A bus interface will often be defined in terms of some logical names, e.g. in AMBA `TVALID`, `TDATA`, etc. However, a design may use different names for these fields, eg `valid`, `data`, etc. This commit adds the ability to pass in a dictionary to a bus interface that describes how to map these logical names to physical names.

A test using the avalon streaming interface is added to test the functionality.

### DIFF
--- a/cocotb/bus.py
+++ b/cocotb/bus.py
@@ -74,6 +74,8 @@ class Bus(object):
             bus_separator (str, optional): Character(s) to use as separator between bus
                 name and signal name. Defaults to '_'.
             array_idx (int or None, optional): Optional index when signal is an array.
+            name_map (dict or None, optional): Optional dictionary mapping
+                logical bus port names to physical port names.
         """
         self._entity = entity
         self._name = name
--- a/cocotb/bus.py
+++ b/cocotb/bus.py
@@ -32,11 +32,15 @@ A bus is simply defined as a collection of signals.
 """
 from cocotb.handle import AssignmentResult
 
-def _build_sig_attr_dict(signals):
+
+def _build_sig_attr_dict(signals, name_map=None):
+    if name_map is None:
+        name_map = {}
     if isinstance(signals, dict):
         return signals
     else:
-        return {sig: sig for sig in signals}
+        return {sig: (name_map[sig] if sig in name_map else sig)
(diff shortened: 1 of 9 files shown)


eric-wieser:
I think this might overlap a lot with some of the changes that I discussed with @Fatsie, it would be worth getting their input here

cocotb/bus.py
```diff
@@ -32,11 +32,15 @@
...
-def _build_sig_attr_dict(signals):
+
+def _build_sig_attr_dict(signals, name_map=None):
+    if name_map is None:
+        name_map = {}
     if isinstance(signals, dict):
         return signals
     else:
-        return {sig: sig for sig in signals}
+        return {sig: (name_map[sig] if sig in name_map else sig)
```
eric-wieser:
```suggestion
        return {sig: name_map.get(sig, sig)
```
is equivalent and faster

cocotb/bus.py
```diff
@@ -73,8 +78,11 @@ def __init__(self, entity, name, signals, optional_signals=[], bus_separator="_"
         self._entity = entity
         self._name = name
         self._signals = {}
+        self._name_map = kwargs.pop('name_map', None)
```
cmarqu: Could you add ``name_map`` to the docstring?

Fatsie:
I did same proposal in #1186 and after discussion it was decided that no extra parameter should be needed but that the `signals` parameter should be used; e.g. `BusDriver.__init__()` and `BusMonitor.__init__()` should combine `signals` parameter to put the aliases inside the `__class__._signals` and `__class__._optional_signals`. In your Avalon example you should then be to replace `name_map=...` with `signals=...`.
I have some development work locally that also implements non-backwards compatible change with `TypedBus` but had to reduce development effort on this as I need to currently focus on other things.

grebe:
Ah, bummer I wasn't able to find that PR before. Thanks for pointing me at it.

I think I follow the rationale, and it seems to me that the code would end up being pretty similar to what we both implemented in these PRs. I have some cycles to work on implementing it this way, let me know if it'd be well received.

PR #1239, opened 2019-12-15
Format Python code with black (and have "compliance" checked in CI)

cmarqu:
This is meant to show what black (https://github.com/psf/black) actually changes when run on our codebase. 
**I do not intend to open this PR as is. If we decide to do such a step, it should be at a strategic point in time to minimize impact on people's work.**


cmarqu:
Lesson learned so far: a black version a few weeks old (19.3b0) produces files that an up-to-date version (19.10b0) would format differently, so the check in CI fails.

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -953,10 +1028,11 @@ def _check_traceback(running_coro, exc_type, pattern):
 @cocotb.test()
 def test_exceptions_direct(dut):
     """ Test exception propagation via a direct yield statement """
+
     @cocotb.coroutine
     def raise_inner():
         yield Timer(10)
-        raise ValueError('It is soon now')
+        raise ValueError("It is soon now")
```
cmarqu: This change is not reflected in the expected traceback string later in this (and the two other such) test(s).
Easy to change, but to be aware of.

eric-wieser:
I'm not convinced that `black` is providing much value here over flake8 - there are a lot of places that use column alignment in cocotb for readability, and in my opinion applying black makes them less readable.

If you're after a more strict style, another approach you could take is to trim down these ignored warnings:

https://github.com/cocotb/cocotb/blob/5262444be60a4c1e7d443df5d2c18a9a77871a95/setup.cfg#L10-L36

cmarqu:
A stricter style is not my intention, but the "autofmt" nature.
I saw people write how they miss support for auto-formatting in other languages once they have gotten used to it (by using Go I think it was), and I imagine the same would happen to me.

imphil:
I did use black and yapf in the past, and none of them provide stable enough output to enforce in CI. (The same is true for many other "opinionated formatters", such as clang-format.) In my experience, the only viable solution is encouraging users to format the diff of their commit, and optionally checking that diff against flake8 or a similar rule-based tool.

PR #1240, opened 2019-12-16
Verilator: Fix trace code

wallento:
Fix trace code after adding unique_ptr instead of simple pointers.


wallento:
After giving it some consideration, I would prefer to stay on the heap for now. Generally it should be the same to allocate this large object on stack or heap, but I want to evaluate and discuss it before changing it.

PR #1241, opened 2019-12-16
Replace tabs with spaces in non-targets

cmarqu:
This fixes the "*** recipe commences before first target. Stop." error
that we would get otherwise (e.g. when the vcs command is not in PATH).

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.vcs
+++ b/cocotb/share/makefiles/simulators/Makefile.vcs
@@ -30,7 +30,7 @@
 ifneq ($(VHDL_SOURCES),)
 
 $(COCOTB_RESULTS_FILE):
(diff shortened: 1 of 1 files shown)


eric-wieser:
Looks like this snuck in with #1193

eric-wieser:
CI failure is persistent:

https://travis-ci.org/cocotb/cocotb/jobs/625948590?utm_medium=notification&utm_source=github_status

cmarqu:
The Travis fails is likely unrelated though, https://travis-ci.org/cocotb/cocotb/jobs/626343327?utm_medium=notification&utm_source=github_status is the same.

imphil:
Ci failure doesn't look related (it should not even touch the VCS makefiles), so merging this to unbreak VCS.

PR #1242, opened 2019-12-17
Fix plusargs testcase

cmarqu:


eric-wieser:
Will trust @cmarqu and @themperek on this one

PR #1243, opened 2019-12-17
Add info about plusargs to doc.

cmarqu:
I never had a need to use plusargs myself, so please review my explanation.

### DIFF
--- a/documentation/source/building.rst
+++ b/documentation/source/building.rst
@@ -62,6 +62,17 @@ Make Variables
 
       Passed to both the compile and execute phases of simulators with two rules, or passed to the single compile and run command for simulators which don't have a distinct compilation stage.
 
+.. make:var:: PLUSARGS
+
+      "Plusargs" are options that are starting with a plus (``+``) sign. 
+      They are passed to the simulator and are also available within cocotb as :data:`cocotb.plusargs`.
+      In the simulator, they can be read by the Verilog/SystemVerilog system functions 
+      ``$test$plusargs`` and ``$value$plusargs``.
+
+      The special plusargs ``+ntb_random_seed`` and ``+seed``, if present, are evaluated
+      to set the random seed value if :envvar:`RANDOM_SEED` is not set.
+      If both ``+ntb_random_seed`` and ``+seed`` are set, ``+ntb_random_seed`` is used.
+
 .. make:var:: COCOTB_HDL_TIMEUNIT
 
       The default time unit that should be assumed for simulation when not specified by modules in the design.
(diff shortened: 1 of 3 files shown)


documentation/source/building.rst
```diff
@@ -62,6 +62,17 @@ Make Variables
 
       Passed to both the compile and execute phases of simulators with two rules, or passed to the single compile and run command for simulators which don't have a distinct compilation stage.
 
+.. make:var:: PLUSARGS
+
+      "Plusargs" are options that are starting with a plus (``+``) sign. 
+      They are passed to the simulator and are also available within cocotb as :attr:`cocotb.plusargs`.
```
eric-wieser: Perhaps should be the following, but maybe it works either way
```suggestion
      They are passed to the simulator and are also available within cocotb as :data:`cocotb.plusargs`.
```

documentation/source/building.rst
```diff
@@ -62,6 +62,17 @@ Make Variables
...
+      In the simulator, they can be read by the (System)Verilog system functions 
+      ``$test$plusargs`` and ``$value$plusargs``.
+
+      The special plusargs ``+ntb_random_seed`` and ``+seed``, if present, are evaluated
```
eric-wieser: Is this simulator specific, or handled by cocotb internally?

imphil: It's handled by cocotb in `__init__.py`. IIRC, `ntb_random_seed` is actually a VCS thing, Mentor and Cadence use `svseed` instead. 

I'd be happy to deprecate this special handling, it seems rather hard to discover and doesn't work universally. But let's discuss that in a separate issue, documenting the current behavior is the right thing to do for this PR.

documentation/source/building.rst
```diff
@@ -62,6 +62,17 @@ Make Variables
 
       Passed to both the compile and execute phases of simulators with two rules, or passed to the single compile and run command for simulators which don't have a distinct compilation stage.
 
+.. make:var:: PLUSARGS
+
+      "Plusargs" are options that are starting with a plus (``+``) sign. 
+      They are passed to the simulator and are also available within cocotb as :attr:`cocotb.plusargs`.
+      In the simulator, they can be read by the (System)Verilog system functions
```
imphil: You can skip the parentheses, the language is called SystemVerilog since ~ 10 years now.

documentation/source/building.rst
cmarqu: Done.

documentation/source/building.rst
cmarqu: I wanted to express that this is present in both plain old Verilog and SystemVerilog. I have rewritten this a bit now.

documentation/source/building.rst
cmarqu: Opened #1252 for that.

eric-wieser:
Let's put this in, comments were addressed.

PR #1244, opened 2019-12-19
Passing Questa arguments consistently

### LINKED ISSUE #711: PLUSARGS handling with QuestaSim
Hi,

I noted that ```PLUSARGS``` is not passed to the Python correctly under Questasim (and maybe other simulators as well). It seems to work as expected under Icarus. Considering the test [test_plusargs](https://github.com/potentialventures/cocotb/blob/master/tests/test_cases/test_plusargs/plusargs.py) the results is OK in Icarus:
```
/usr/local/bin/vvp -M /home/mci/cocotb-pv/build/libs/x86_64 -m gpivpi sim_build/sim.vvp   +foo=bar +test1 +test2 +options=fubar
     -.--ns INFO     cocotb.gpi                                  gpi_embed.c:111  in embed_init_python               Did not detect Python virtual environment. Using system-wide Python interpreter.
     -.--ns INFO     cocotb.gpi                                GpiCommon.cpp:91   in gpi_print_registered_impl       VPI registered
    0.00ns INFO     cocotb.gpi                                  gpi_embed.c:278  in embed_sim_init                  Running on Icarus Verilog version 0.10.0 (devel)
    0.00ns INFO     cocotb.gpi                                  gpi_embed.c:279  in embed_sim_init                  Python interpreter initialised and cocotb loaded!
    0.00ns INFO     cocotb                                      __init__.py:118  in _initialise_testbench           Running tests with Cocotb v1.0.1 from /home/mci/cocotb-pv
    0.00ns INFO     cocotb                                      __init__.py:134  in _initialise_testbench           Seeding Python random module with 1544089045
    0.00ns INFO     cocotb.regression                         regression.py:187  in _initialise                     Found test plusargs.plusargs_test
    0.00ns INFO     cocotb.regression                         regression.py:319  in execute                         Running test 1/1: plusargs_test
    0.00ns INFO     ..tine.plusargs_test.0x7f43a6bccf60       decorators.py:211  in send                            Starting test: "plusargs_test"
                                                                                                                    Description: Demonstrates plusarg access from Python test
SIM: Plusargs test
SIM: Plusarg foo has value bar
SIM: Test running
COCOTB: foo bar
COCOTB: test1 True
COCOTB: test2 True
COCOTB: options fubar
20001000000000.00ns INFO     cocotb.regression                         regression.py:264  in handle_result                   Test Passed: plusargs_test
```

The same test run under Questa:

```
/opt/tools/ams/amsv_2017_3_1/amsv/modeltech/bin/vsim -c -64 -do runsim.do 2>&1 +foo=bar +test1 +test2 +options=fubar | tee sim.log
...
Reading pref.tcl
# Loading sv_std.std
# Loading work.tb_top
# Loading /home/mci/cocotb-pv/build/libs/x86_64/libvpi.so
#      -.--ns INFO     cocotb.gpi                                  gpi_embed.c:111  in embed_init_python               Did not detect Python virtual environment. Using system-wide Python interpreter.
#      -.--ns INFO     cocotb.gpi                                GpiCommon.cpp:91   in gpi_print_registered_impl       VPI registered
#     0.00ns INFO     cocotb.gpi                                  gpi_embed.c:278  in embed_sim_init                  Running on ModelSim for Questa-64 version 10.6c_1 2017.09
#     0.00ns INFO     cocotb.gpi                                  gpi_embed.c:279  in embed_sim_init                  Python interpreter initialised and cocotb loaded!
#     0.00ns INFO     cocotb                                      __init__.py:118  in _initialise_testbench           Running tests with Cocotb v1.0.1 from /home/mci/cocotb-pv
#     0.00ns INFO     cocotb                                      __init__.py:134  in _initialise_testbench           Seeding Python random module with 1544090493
#     0.00ns INFO     cocotb.regression                         regression.py:187  in _initialise                     Found test plusargs.plusargs_test
#     0.00ns INFO     cocotb.regression                         regression.py:319  in execute                         Running test 1/1: plusargs_test
#     0.00ns INFO     ..tine.plusargs_test.0x7fffe9ffe470       decorators.py:211  in send                            Starting test: "plusargs_test"
#                                                                                                                     Description: Demonstrates plusarg access from Python test
# SIM: Plusargs test
# SIM: Plusarg foo has value 
# SIM: Test running
#  2000.10ns INFO     cocotb.regression                         regression.py:264  in handle_result                   Test Passed: plusargs_test

```

Interestingly, arguments are passed correctly under Questa by ```EXTRA_ARGS``` variable. 
Any clues what may be going on?

### LINKED ISSUE #1150: QuestaSim: -noautoldlibpath does not work
In the Questa Makefile, `vsim` is called twice recursively - once to execute the `runsim.do` script, where `SIM_ARGS` are not applied and once in the `runsim.do` script with the `SIM_ARGS` arguments. The `-noautoldlibpath` argument prevents the `vsim` launch script from setting the `LD_LIBRARY_PATH` environment variable - but if this is done only when the nested `vsim` call is executed, it's too late - `LD_LIBRARY_PATH` is already contaminated. This can be a problem because the Mentor-distributed GCC distributions are really old and some dependencies in your CocoTB code may require newer C/C++ libraries. In my case, this was a problem when I tried to use PyTorch.

I have worked around this by adding a `PRE_SIM_ARGS` variable which is applied at the first call to `vsim` already, but maybe there is a nicer way to do it...

ktbarrett:
Fixes #1150. Tested locally, the plusargs test now passes and using `-noautoldlibpath` should be specified in `SIM_ARGS` and will be passed to both calls to vsim.

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.questa
+++ b/cocotb/share/makefiles/simulators/Makefile.questa
@@ -148,7 +148,7 @@ $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/runsim.do $(COCOTB_LIBS) $(INT_LIBS)
 
 	set -o pipefail; $(LIB_LOAD) MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) COCOTB_SIM=1 \
 	GPI_EXTRA=$(GPI_EXTRA) TOPLEVEL_LANG=$(TOPLEVEL_LANG) PYTHONPATH=$(LIB_DIR):$(PWD):$(NEW_PYTHONPATH) \
-	$(CMD) $(PLUSARGS) -do $(SIM_BUILD)/runsim.do 2>&1 | tee $(SIM_BUILD)/sim.log
+	$(CMD) $(RUN_ARGS) $(PLUSARGS) -do $(SIM_BUILD)/runsim.do 2>&1 | tee $(SIM_BUILD)/sim.log
 
 	# check that the file was actually created, since we can't set an exit code from cocotb
 	test -f $(COCOTB_RESULTS_FILE)
--- a/documentation/source/building.rst
+++ b/documentation/source/building.rst
@@ -77,6 +77,12 @@ and
 
       Any arguments or flags to pass to the execution of the compiled simulation.
 
+.. make:var:: RUN_ARGS
+
+      Any argument to be passed to the "first" invocation of a simulator that runs via a TCL script.
+      One motivating usage is to pass `-noautoldlibpath` to Questa to prevent it from loading the out-of-date libraries it ships with.
+      Used by Aldec Riviera-PRO and Mentor Graphics Questa simulator.
+
 .. make:var:: EXTRA_ARGS
 
       Passed to both the compile and execute phases of simulators with two rules, or passed to the single compile and run command for simulators which don't have a distinct compilation stage.
--- a/cocotb/share/makefiles/simulators/Makefile.aldec
+++ b/cocotb/share/makefiles/simulators/Makefile.aldec
@@ -176,7 +176,7 @@ $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/runsim.tcl $(COCOTB_LIBS) $(COCOTB_VHPI_LIB
 	-@rm -f $(COCOTB_RESULTS_FILE)
 
 	set -o pipefail; cd $(SIM_BUILD) && PYTHONPATH=$(LIB_DIR):$(PWD):$(NEW_PYTHONPATH) GPI_EXTRA=$(GPI_EXTRA) TOPLEVEL_LANG=$(TOPLEVEL_LANG) \
-	$(LIB_LOAD) MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) COCOTB_SIM=1 $(CMD) -do runsim.tcl | tee sim.log
+	$(LIB_LOAD) MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) COCOTB_SIM=1 $(CMD) $(RUN_ARGS) -do runsim.tcl | tee sim.log
 
 	# check that the file was actually created
 	test -f $(COCOTB_RESULTS_FILE)
--- a/documentation/source/newsfragments/1244.feature.rst
+++ b/documentation/source/newsfragments/1244.feature.rst
@@ -1,0 +1,1 @@
+Simulators run through a TCL script (Aldec Riviera Pro and Mentor simulators) now support a new :make:var:`RUN_ARGS` Makefile variable, which is passed to the first invocation of the tool during runtime.


cocotb/share/makefiles/simulators/Makefile.questa
```diff
@@ -146,7 +146,7 @@ endif
 $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/runsim.do $(COCOTB_LIBS) $(INT_LIBS)
 	set -o pipefail; cd $(SIM_BUILD) && $(LIB_LOAD) MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) COCOTB_SIM=1 \
 	GPI_EXTRA=$(GPI_EXTRA) TOPLEVEL_LANG=$(TOPLEVEL_LANG) PYTHONPATH=$(LIB_DIR):$(PWD):$(NEW_PYTHONPATH) \
-	$(CMD) $(PLUSARGS) -do runsim.do 2>&1 | tee sim.log
+	$(CMD) $(VSIM_ARGS) $(PLUSARGS) -do runsim.do 2>&1 | tee sim.log
```
ktbarrett: Is `PLUSARGS` necessary here?

themperek: As I know not needed here.

mciepluc:
Please note that ```VSIM_ARGS``` refers specifically to the simulation phase. The ```PLUSARGS``` was introduced in order to provide an opportunity to pass some arguments for the command call, which is a different thing. I don't think that ```VSIM_ARGS``` is going to work here, please read the https://github.com/cocotb/cocotb/issues/711#issuecomment-445711892

ktbarrett:
@mciepluc Just a typo, I meant passing `-noautoldlibpath` via `SIM_ARGS` so that it goes to both vsim calls. Unless you are saying you are using `SIM_ARGS` for options that would break if they got passed to the first `vsim` call?

themperek:
There is actually a way to avoid setting `LD_LIBRARY_PATH` in general by passing `-Wl,-rpath,$ORIGIN` to the linker and passing full path library to the simulator. I used it here: https://github.com/themperek/cocotb-test/blob/1dc351d971254d10811b589028322c35f67ac9cd/cocotb_test/build_libs.py#L133 and seem to work. I will try to make some PR for this but not sure how this will be accepted.

ktbarrett:
@themperek I think that's missing the point. Doesn't that just link only the cocotb libraries against the specified libraries? Or does Questa load foreign dependencies first? The idea is that we want Questa to load with a newer libstdc++ than it ships with so that we can run models and support libraries that utilize modern C++ standards. I've had issues loading C++11 python extensions in cocotb when running Questa 10.5c because it loads it's own (ancient and incompatible) libstdc++ first.

themperek:
@ktbarrett Not sure what exactly you need but there is a way to remove setting `LD_LIBRARY_PATH` from cocotb so the user has more freedom do whatever is needed.

ktbarrett:
@themperek What cocotb does is just fine. This is more of a tool issue. 

If we can't pass `-noautoldlibpath` to both instances of vsim, it will *prepend* it's own libraries to LD_LIBRARY_PATH. At that point we cannot load other libraries (like MATLAB) that aren't compatible with Questa's libraries. We can't always control the compilation of these other libraries, so if the LD_LIBRARY_PATH is modified by Questa, it will crash when loading those libraries (CXXABI version error). There is no alternative solution, we need to be able to turn off this behavior in Questa.

eric-wieser:
I think I'd like @mciepluc to sign off on this, since I don't have the background to review and it seems they do.

cocotb/share/makefiles/simulators/Makefile.questa
```diff
@@ -146,7 +146,7 @@ endif
 $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/runsim.do $(COCOTB_LIBS) $(INT_LIBS)
 	set -o pipefail; cd $(SIM_BUILD) && $(LIB_LOAD) MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) COCOTB_SIM=1 \
 	GPI_EXTRA=$(GPI_EXTRA) TOPLEVEL_LANG=$(TOPLEVEL_LANG) PYTHONPATH=$(LIB_DIR):$(PWD):$(NEW_PYTHONPATH) \
-	$(CMD) $(PLUSARGS) -do runsim.do 2>&1 | tee sim.log
+	$(CMD) $(VSIM_ARGS) -do runsim.do 2>&1 | tee sim.log
```
ktbarrett:
```suggestion
	$(CMD) $(RUN_ARGS) -do runsim.do 2>&1 | tee sim.log
```

This is an alternative if SIM_ARGS is not right for here. If not SIM_ARGS, not of the other variables make sense.

themperek:
Where we are here we could clean this according to [documentation](https://github.com/cocotb/cocotb/blob/master/documentation/source/building.rst#make-variables) and use `SIM_ARGS`, `COMPILE_ARGS` and `EXTRA_ARGS`. 

I think it is ok to add  `RUN_ARGS` or `CMD_ARGS` for the Questa case (same will be the case for Aldec).

ktbarrett:
@themperek Is there someone who can test Riviera? I'd rather not make changes blindly.

themperek:
@ktbarrett I guess Riviera we will have to do another time.

cocotb/share/makefiles/simulators/Makefile.questa
mciepluc: This is going to be OK, please update docs as well.

ktbarrett:
Moved PLUSARGS makefile issue to #1256 seeing as it's not objectionable.

@mciepluc 
@da-gazzi
This good for you guys? An alternative would be to have a single variable that goes to both invocations, but this is more flexible, even if it is less clean.

da-gazzi:
> Moved PLUSARGS makefile issue to #1256 seeing as it's not objectionable.
> 
> @mciepluc
> @da-gazzi
> This good for you guys? An alternative would be to have a single variable that goes to both invocations, but this is more flexible, even if it is less clean.

Very nice, it's actually exactly the same as what I did ;) Thanks!

documentation/source/building.rst
```diff
@@ -67,6 +67,10 @@ and
 
       Any arguments or flags to pass to the execution of the compiled simulation.
 
+.. make:var:: RUN_ARGS
+
+      Any argument to be passed to the "first" invocation of a simulator that runs via a TCL script. Used by Aldec Riviera-PRO and Mentor Graphics Questa simulator.
```
imphil: - Riviera isn't currently using this variable. Either remove it from the docs, or add it to the Riviera Makefile.
- Please use one sentence per line, i.e. insert a newline before `Used`.
- Can you please document an example of when this might be useful, as you did in the commit message?

imphil:
Rebased and added newsfragment. Merging once CI is green.

PR #1248, opened 2019-12-20
Add AXI4Master driver

corna:
This is based on the existing AXI4Lite driver, which has been extended
to use the AXI4 signals.
The existing AXI4LiteMaster has been modified to wrap the AXI4Master
methods, providing so a backwards-compatible class.

### DIFF
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -188,31 +188,42 @@ class AXI4Master(BusDriver):
             self.bus.AWVALID <= 0
 
     async def _send_write_data(
-        self, data: Sequence[int], delay: int,
-        byte_enable: Sequence[Optional[int]], sync: bool
+        self, address, data: Sequence[int], burst: AXIBurst, size: int,
+        delay: int, byte_enable: Sequence[Optional[int]], sync: bool
     ) -> None:
         """Send the write data, with optional delay (in clocks)."""
+
+        # Helper function for narrow bursts
+        def mask_and_shift(value: int, block_size: int, block_num: int) -> int:
+            return (value & (2**block_size - 1)) << (block_num * block_size)
+
         async with self.write_data_busy:
             if sync:
                 await RisingEdge(self.clock)
 
+            wdata_bytes = len(self.bus.WDATA) // 8
             byte_enable_iterator = iter(byte_enable)
+            narrow_block = (address % wdata_bytes) // size
 
             for i, word in enumerate(data):
                 await ClockCycles(self.clock, delay)
 
-                self.bus.WDATA <= word
                 self.bus.WVALID <= 1
+                self.bus.WDATA <= mask_and_shift(word, size * 8, narrow_block)
 
                 try:
                     current_byte_enable = next(byte_enable_iterator)
-                    self.bus.WSTRB <= (2**self.bus.WSTRB.value.n_bits - 1
-                                       if current_byte_enable is None
-                                       else current_byte_enable)
+                    strobe = 2**self.bus.WSTRB.value.n_bits - 1 \
+                        if current_byte_enable is None else current_byte_enable
                 except StopIteration:
-                    # Do not update WSTRB if we have reached the end of the
+                    # Do not update strobe if we have reached the end of the
                     # iterator
                     pass
+
+                self.bus.WSTRB <= mask_and_shift(strobe, size, narrow_block)
+
+                if burst is not AXIBurst.FIXED:
+                    narrow_block = (narrow_block + 1) % (wdata_bytes // size)
 
                 if hasattr(self.bus, "WLAST"):
                     if i == len(data) - 1:
@@ -277,8 +288,8 @@ class AXI4Master(BusDriver):
         write_address = self._send_write_address(address, len(value), burst,
                                                  size, address_latency, sync)
 
-        write_data = self._send_write_data(value, data_latency, byte_enable,
-                                           sync)
+        write_data = self._send_write_data(address, value, burst, size,
+                                           data_latency, byte_enable, sync)
 
         await Combine(cocotb.fork(write_address), cocotb.fork(write_data))
 
@@ -334,6 +345,13 @@ class AXI4Master(BusDriver):
                 not match the requested one.
         """
 
+        # Helper function for narrow bursts
+        def shift_and_mask(binvalue: BinaryValue, bytes_num: int,
+                           byte_shift: int) -> BinaryValue:
+            start = byte_shift * 8
+            end = (bytes_num + byte_shift) * 8
+            return binvalue[len(binvalue) - end:len(binvalue) - start - 1]
+
         if size is None:
             size = len(self.bus.RDATA) // 8
         else:
@@ -341,6 +359,9 @@ class AXI4Master(BusDriver):
 
         AXI4Master._check_length(length, burst)
         AXI4Master._check_4kB_boundary_crossing(address, burst, size, length)
+
+        rdata_bytes = len(self.bus.RDATA) // 8
+        byte_offset = address % rdata_bytes
 
         async with self.read_address_busy:
             if sync:
@@ -375,8 +396,22 @@ class AXI4Master(BusDriver):
                 while True:
                     await ReadOnly()
                     if self.bus.RVALID.value and self.bus.RREADY.value:
-                        data.append(self.bus.RDATA.value)
+                        # Shift and mask to correctly handle narrow bursts
+                        new_val = shift_and_mask(self.bus.RDATA.value,
+                                                 size, byte_offset)
+
+                        # Push the value into a new BinaryValue, identical to
+                        # the previous one but with a lower number of bits
+                        data.append(BinaryValue(
+                            value=new_val.binstr, n_bits=size * 8,
+                            bigEndian=new_val.big_endian,
+                            binaryRepresentation=new_val.binaryRepresentation))
+
                         rresp.append(AXIxRESP(self.bus.RRESP.value.integer))
+
+                        if burst is not AXIBurst.FIXED:
+                            byte_offset = (byte_offset + size) % rdata_bytes
+
                         break
                     await RisingEdge(self.clock)
 
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -188,7 +188,10 @@ class AXI4Master(BusDriver):
             self.bus.WVALID <= 1
 
             try:
-                self.bus.WSTRB <= next(byte_enable_iterator)
+                current_byte_enable = next(byte_enable_iterator)
+                self.bus.WSTRB <= (2**self.bus.WSTRB.value.n_bits - 1
+                                   if current_byte_enable is None
+                                   else current_byte_enable)
             except StopIteration:
                 # Do not update WSTRB if we have reached the end of the iter
                 pass
@@ -211,7 +214,7 @@ class AXI4Master(BusDriver):
 
     @cocotb.coroutine
     async def write(self, address, value, *, size=None, burst=AXIBurst.INCR,
-                    byte_enable=0xf, address_latency=0, data_latency=0,
+                    byte_enable=None, address_latency=0, data_latency=0,
                     sync=True):
         """Write a value to an address.
 
@@ -222,8 +225,8 @@ class AXI4Master(BusDriver):
                 the data bus).
             burst: The burst type, either ``FIXED``, ``INCR`` or ``WRAP``.
                 Defaults to ``INCR``.
-            byte_enable: Which bytes in value to actually write. Default is to
-                write all bytes.
+            byte_enable: Which bytes in value to actually write. Defaults to
+                None (write all bytes).
             address_latency: Delay before setting the address (in clock
                 cycles). Default is no delay.
             data_latency: Delay before setting the data value (in clock
@@ -401,7 +404,7 @@ class AXI4LiteMaster(AXI4Master):
 
     @cocotb.coroutine
     async def write(
-        self, address: int, value: int, byte_enable: int = 0xf,
+        self, address: int, value: int, byte_enable: Optional[int] = None,
         address_latency: int = 0, data_latency: int = 0, sync: bool = True
     ) -> BinaryValue:
         """Write a value to an address.
@@ -409,8 +412,8 @@ class AXI4LiteMaster(AXI4Master):
         Args:
             address: The address to write to.
             value: The data value to write.
-            byte_enable: Which bytes in value to actually write. Default is to
-                write all bytes.
+            byte_enable: Which bytes in value to actually write. Defaults to
+                None (write all bytes).
             address_latency: Delay before setting the address (in clock
                 cycles). Default is no delay.
             data_latency: Delay before setting the data value (in clock
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -37,7 +37,9 @@ from cocotb.triggers import ClockCycles, Combine, Lock, ReadOnly, RisingEdge
 
 
 class AXIProtocolError(Exception):
-    pass
+    def __init__(self,  message, xresp):
+        super().__init__(message)
+        self.xresp = xresp
 
 
 class AXIReadBurstLengthMismatch(Exception):
@@ -287,7 +289,7 @@ class AXI4Master(BusDriver):
 
                 raise AXIProtocolError(
                     err_msg.format(address, len(value), burst.name,
-                                   result.value, result.name))
+                                   result.value, result.name), result)
 
     @cocotb.coroutine
     async def read(self, address, length=1, *, size=None, burst=AXIBurst.INCR,
@@ -389,7 +391,7 @@ class AXI4Master(BusDriver):
                             address, beat_number + 1, length, burst,
                             beat_result.value, beat_result.name)
 
-                        raise AXIProtocolError(err_msg)
+                        raise AXIProtocolError(err_msg, beat_result)
 
                 return data
 
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -29,6 +29,7 @@
 import array
 import collections.abc
 import enum
+import itertools
 from typing import Any, List, Optional, Sequence, Tuple, Union
 
 import cocotb
@@ -197,36 +198,60 @@ class AXI4Master(BusDriver):
         def mask_and_shift(value: int, block_size: int, block_num: int) -> int:
             return (value & (2**block_size - 1)) << (block_num * block_size)
 
+        # [0x33221100, 0x77665544] --> [0x221100XX, 0x66554433]
+        def unalign_data(
+            data: Sequence[int], size_bits: int, shift: int
+        ) -> List[int]:
+            padded_data = (0,) + tuple(value for value in data)
+            low_mask = 2**(size_bits - shift) - 1
+            high_mask = (2**shift - 1) << (size_bits - shift)
+
+            return [(padded_data[i] & high_mask) >> (size_bits - shift) |
+                    (padded_data[i + 1] & low_mask) << shift
+                    for i in range(len(data))]
+
+        strobes = []
+        byte_enable_iterator = iter(byte_enable)
+        try:
+            for i in range(len(data)):
+                current_byte_enable = next(byte_enable_iterator)
+                strobes.append(2**size - 1 if current_byte_enable is None
+                               else current_byte_enable)
+        except StopIteration:
+            # Fill the remaining strobes with the last one if we have reached
+            # the end of the iterator
+            strobes += [strobes[-1]] * (len(data) - i)
+
+        # Unalign the words and strobes (if unaligned and not FIXED)
+        if address % size != 0:
+            shift = (address % size) * 8
+            if burst is AXIBurst.FIXED:
+                data = [(word << shift) & (2**(size * 8) - 1) for word in data]
+                strobes = \
+                    [(strb << shift // 8) & (2**size - 1) for strb in strobes]
+            else:
+                data = unalign_data(data, size * 8, shift)
+                strobes = unalign_data(strobes, size, address % size)
+
         async with self.write_data_busy:
             if sync:
                 await RisingEdge(self.clock)
 
             wdata_bytes = len(self.bus.WDATA) // 8
-            byte_enable_iterator = iter(byte_enable)
             narrow_block = (address % wdata_bytes) // size
 
-            for i, word in enumerate(data):
+            for beat_num, (word, strobe) in enumerate(zip(data, strobes)):
                 await ClockCycles(self.clock, delay)
 
                 self.bus.WVALID <= 1
                 self.bus.WDATA <= mask_and_shift(word, size * 8, narrow_block)
-
-                try:
-                    current_byte_enable = next(byte_enable_iterator)
-                    strobe = 2**self.bus.WSTRB.value.n_bits - 1 \
-                        if current_byte_enable is None else current_byte_enable
-                except StopIteration:
-                    # Do not update strobe if we have reached the end of the
-                    # iterator
-                    pass
-
                 self.bus.WSTRB <= mask_and_shift(strobe, size, narrow_block)
 
                 if burst is not AXIBurst.FIXED:
                     narrow_block = (narrow_block + 1) % (wdata_bytes // size)
 
                 if hasattr(self.bus, "WLAST"):
-                    if i == len(data) - 1:
+                    if beat_num == len(data) - 1:
                         self.bus.WLAST <= 1
                     else:
                         self.bus.WLAST <= 0
@@ -236,7 +261,7 @@ class AXI4Master(BusDriver):
                     if self.bus.WREADY.value:
                         break
 
-                if i == len(data) - 1:
+                if beat_num == len(data) - 1:
                     self.bus.WVALID <= 0
 
     @cocotb.coroutine
@@ -247,6 +272,11 @@ class AXI4Master(BusDriver):
         address_latency: int = 0, data_latency: int = 0, sync: bool = True
     ) -> None:
         """Write a value to an address.
+
+        With unaligned writes (when ``address`` is not a multiple of ``size``),
+        only the low ``size - address % size`` Bytes are written for:
+         * the last element of ``value`` for INCR or WRAP bursts, or
+         * every element of ``value`` for FIXED bursts.
 
         Args:
             address: The address to write to.
@@ -322,6 +352,13 @@ class AXI4Master(BusDriver):
     ) -> Union[List[BinaryValue], List[Tuple[BinaryValue, AXIxRESP]]]:
         """Read from an address.
 
+        With unaligned reads (when ``address`` is not a multiple of ``size``)
+        with INCR or WRAP bursts, the last element of the returned read data
+        will be only the low-order ``size - address % size`` Bytes of the last
+        word.
+        With unaligned reads with FIXED bursts, every element of the returned
+        read data will be only the low-order ``size - address % size`` Bytes.
+
         Args:
             address: The address to read from.
             length: Number of words to transfer. Defaults to 1.
@@ -352,6 +389,17 @@ class AXI4Master(BusDriver):
             end = (bytes_num + byte_shift) * 8
             return binvalue[len(binvalue) - end:len(binvalue) - start - 1]
 
+        # [0x221100XX, 0x66554433] --> [0x33221100, 0x665544]
+        def realign_data(
+            data: Sequence[BinaryValue], size_bits: int, shift: int
+        ) -> List[BinaryValue]:
+            binstr_join = "".join([word.binstr[::-1] for word in data])
+            binstr_join = binstr_join[shift:]
+            data_binstr = [binstr_join[i * size_bits:(i + 1) * size_bits][::-1]
+                           for i in range(len(data))]
+            return [BinaryValue(value=binstr, n_bits=len(binstr))
+                    for binstr in data_binstr]
+
         if size is None:
             size = len(self.bus.RDATA) // 8
         else:
@@ -361,7 +409,7 @@ class AXI4Master(BusDriver):
         AXI4Master._check_4kB_boundary_crossing(address, burst, size, length)
 
         rdata_bytes = len(self.bus.RDATA) // 8
-        byte_offset = address % rdata_bytes
+        byte_offset = (address % rdata_bytes) // size * size
 
         async with self.read_address_busy:
             if sync:
@@ -392,21 +440,15 @@ class AXI4Master(BusDriver):
             data = []
             rresp = []
 
-            while True:
+            for beat_num in itertools.count():
                 while True:
                     await ReadOnly()
                     if self.bus.RVALID.value and self.bus.RREADY.value:
                         # Shift and mask to correctly handle narrow bursts
-                        new_val = shift_and_mask(self.bus.RDATA.value,
-                                                 size, byte_offset)
+                        beat_value = shift_and_mask(self.bus.RDATA.value,
+                                                    size, byte_offset)
 
-                        # Push the value into a new BinaryValue, identical to
-                        # the previous one but with a lower number of bits
-                        data.append(BinaryValue(
-                            value=new_val.binstr, n_bits=size * 8,
-                            bigEndian=new_val.big_endian,
-                            binaryRepresentation=new_val.binaryRepresentation))
-
+                        data.append(beat_value)
                         rresp.append(AXIxRESP(self.bus.RRESP.value.integer))
 
                         if burst is not AXIBurst.FIXED:
@@ -428,6 +470,14 @@ class AXI4Master(BusDriver):
                     "words, received {})"
                     .format("more" if len(data) > length else "less",
                             length, len(data)))
+
+            # Re-align the words
+            if address % size != 0:
+                shift = (address % size) * 8
+                if burst is AXIBurst.FIXED:
+                    data = [word[0:size * 8 - shift - 1] for word in data]
+                else:
+                    data = realign_data(data, size * 8, shift)
 
             if return_rresp:
                 return list(zip(data, rresp))
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -29,21 +29,13 @@
 import array
 import collections.abc
 import enum
+from typing import Any, List, Optional, Sequence, Tuple, Union
 
 import cocotb
 from cocotb.binary import BinaryValue
 from cocotb.drivers import BusDriver
+from cocotb.handle import SimHandleBase
 from cocotb.triggers import ClockCycles, Combine, Lock, ReadOnly, RisingEdge
-
-
-class AXIProtocolError(Exception):
-    def __init__(self,  message, xresp):
-        super().__init__(message)
-        self.xresp = xresp
-
-
-class AXIReadBurstLengthMismatch(Exception):
-    pass
 
 
 class AXIBurst(enum.IntEnum):
@@ -57,6 +49,16 @@ class AXIxRESP(enum.IntEnum):
     EXOKAY = 0b01
     SLVERR = 0b10
     DECERR = 0b11
+
+
+class AXIProtocolError(Exception):
+    def __init__(self,  message: str, xresp: AXIxRESP):
+        super().__init__(message)
+        self.xresp = xresp
+
+
+class AXIReadBurstLengthMismatch(Exception):
+    pass
 
 
 class AXI4Master(BusDriver):
@@ -76,7 +78,8 @@ class AXI4Master(BusDriver):
                          "WLAST",
                          "ARREGION", "ARLOCK", "ARCACHE", "ARPROT", "ARQOS"]
 
-    def __init__(self, entity, name, clock, **kwargs):
+    def __init__(self, entity: SimHandleBase, name: str, clock: SimHandleBase,
+                 **kwargs: Any):
         BusDriver.__init__(self, entity, name, clock, **kwargs)
 
         # Drive some sensible defaults (setimmediatevalue to avoid x asserts)
@@ -111,7 +114,7 @@ class AXI4Master(BusDriver):
         self.write_response_busy = Lock(name + "_bbusy")
 
     @staticmethod
-    def _check_length(length, burst):
+    def _check_length(length: int, burst: AXIBurst) -> None:
         """Check that the provided burst length is valid."""
         if length <= 0:
             raise ValueError("Burst length must be a positive integer")
@@ -128,7 +131,7 @@ class AXI4Master(BusDriver):
                 raise ValueError("Maximum burst length for FIXED bursts is 16")
 
     @staticmethod
-    def _check_size(size, data_bus_width):
+    def _check_size(size: int, data_bus_width: int) -> None:
         """Check that the provided transfer size is valid."""
         if size > data_bus_width:
             raise ValueError("Beat size ({} B) is larger than the bus width "
@@ -137,7 +140,8 @@ class AXI4Master(BusDriver):
             raise ValueError("Beat size must be a positive power of 2")
 
     @staticmethod
-    def _check_4kB_boundary_crossing(address, burst, size, length):
+    def _check_4kB_boundary_crossing(address: int, burst: AXIBurst, size: int,
+                                     length: int) -> None:
         """Check that the provided burst does not cross a 4kB boundary."""
         if burst is AXIBurst.INCR:
             last_address = address + size * (length - 1)
@@ -149,8 +153,10 @@ class AXI4Master(BusDriver):
                     .format(address, last_address,
                             (address & ~0xfff) + 0x1000))
 
-    async def _send_write_address(self, address, length, burst, size, delay,
-                                  sync):
+    async def _send_write_address(
+        self, address: int, length: int, burst: AXIBurst, size: int,
+        delay: int, sync: bool
+    ) -> None:
         """Send the write address, with optional delay (in clocks)"""
         async with self.write_address_busy:
             if sync:
@@ -181,7 +187,10 @@ class AXI4Master(BusDriver):
             await RisingEdge(self.clock)
             self.bus.AWVALID <= 0
 
-    async def _send_write_data(self, data, delay, byte_enable, sync):
+    async def _send_write_data(
+        self, data: Sequence[int], delay: int,
+        byte_enable: Sequence[Optional[int]], sync: bool
+    ) -> None:
         """Send the write data, with optional delay (in clocks)."""
         async with self.write_data_busy:
             if sync:
@@ -220,9 +229,12 @@ class AXI4Master(BusDriver):
                     self.bus.WVALID <= 0
 
     @cocotb.coroutine
-    async def write(self, address, value, *, size=None, burst=AXIBurst.INCR,
-                    byte_enable=None, address_latency=0, data_latency=0,
-                    sync=True):
+    async def write(
+        self, address: int, value: Union[int, Sequence[int]], *,
+        size: Optional[int] = None, burst: AXIBurst = AXIBurst.INCR,
+        byte_enable: Union[Optional[int], Sequence[Optional[int]]] = None,
+        address_latency: int = 0, data_latency: int = 0, sync: bool = True
+    ) -> None:
         """Write a value to an address.
 
         Args:
@@ -292,8 +304,11 @@ class AXI4Master(BusDriver):
                                    result.value, result.name), result)
 
     @cocotb.coroutine
-    async def read(self, address, length=1, *, size=None, burst=AXIBurst.INCR,
-                   return_rresp=False, sync=True):
+    async def read(
+        self, address: int, length: int = 1, *,
+        size: Optional[int] = None, burst: AXIBurst = AXIBurst.INCR,
+        return_rresp: bool = False, sync: bool = True
+    ) -> Union[List[BinaryValue], List[Tuple[BinaryValue, AXIxRESP]]]:
         """Read from an address.
 
         Args:
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -385,6 +385,8 @@ class AXI4Master(BusDriver):
 
                 await RisingEdge(self.clock)
 
+            await RisingEdge(self.clock)
+
             if len(data) != length:
                 raise AXIReadBurstLengthMismatch(
                     "AXI4 slave returned {} data than expected (requested {} "
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -101,10 +101,12 @@ class AXI4Master(BusDriver):
             except AttributeError:
                 pass
 
-        # Mutex for each channel that we master to prevent contention
-        self.write_address_busy = Lock("%s_wabusy" % name)
-        self.read_address_busy = Lock("%s_rabusy" % name)
-        self.write_data_busy = Lock("%s_wbusy" % name)
+        # Mutex for each channel to prevent contention
+        self.write_address_busy = Lock(name + "_awbusy")
+        self.read_address_busy = Lock(name + "_arbusy")
+        self.write_data_busy = Lock(name + "_wbusy")
+        self.read_data_busy = Lock(name + "_rbusy")
+        self.write_response_busy = Lock(name + "_bbusy")
 
     @staticmethod
     def _check_length(length, burst):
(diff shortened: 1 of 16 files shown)


corna:
Rebased.
Let me know if it is OK.

corna:
Quite OT: I developed the drivers and a monitor for AXI4-Stream, another protocol part of the AMBA family. It is currently published as [extension](https://github.com/corna/cocotbext-axi4stream): should I merge it in cocotb or keep it as separate extension?

corna:
Rebased.
Any feedback on this?

eric-wieser:
Is there any chance you could write some tests for this?

One of the big problems we have with bus drivers in cocotb is they never have tests, so they get broken easily.

We've been thinking about moving buses to live in their own repository - so I can't guarantee we'll merge this here, but I'm sure it will end up being useful to someone!

cocotb/drivers/amba.py
```diff
@@ -26,31 +26,43 @@
...
+         "ARVALID", "ARADDR", "ARREADY", "ARID", "ARLEN", "ARSIZE", "ARBURST",
+         "RVALID", "RREADY", "RRESP", "RDATA", "RID", "RLAST"]
+
+    _optional_signals = ["AWREGION", "AWLOCK", "AWCACHE", "AWPROT", "AWQOS",
+                         "WLAST",
+                         "ARREGION", "ARLOCK", "ARCACHE", "ARPROT", "ARQOS"]
 
-    def __init__(self, entity, name, clock, **kwargs):
+    _xRESP_msg = ("OKAY", "EXOKAY", "SLVERR", "DECERR")
+    _BURST_values = {"FIXED": 0b00, "INCR": 0b01, "WRAP": 0b10}
```
eric-wieser, line 62: I'd recommend using for this
```python
class Axi4Burst(enum.IntEnum):
    FIXED = 0b00
    INCR = 0b01
    WRAP = 0b10
```

cocotb/drivers/amba.py
```diff
@@ -129,21 +239,32 @@ def write(
...
+            TestError: If any of the input parameters is invalid.
             AXIProtocolError: If write response from AXI is not ``OKAY``.
         """
+
+        try:
+            iter(value)
+        except TypeError:
+            value = (value,)    # If value is not an iterable, make it
+
+        AXI4Master._check_burst(burst)
```
eric-wieser, line 251: This would become `burst = AXI4Burst(burst)`

corna:
OK, I will write a test for it.

cocotb/drivers/amba.py
```diff
@@ -60,23 +72,88 @@ def __init__(self, entity, name, clock, **kwargs):
...
+                                "16")
+
+    @staticmethod
+    def _check_size(size: int, data_bus_width: int) -> None:
+        """Check that the provided transfer size is valid."""
+        if size > data_bus_width:
+            raise TestError("Transfer size ({} B) is larger than the bus "
+                            "width ({} B)".format(size, data_bus_width))
+        elif size <= 0 or size & (size - 1) != 0:
+            raise TestError("Transfer size must be a positive power of 2")
```
eric-wieser, line 131: These should all be `ValueError`

cocotb/drivers/amba.py
```diff
@@ -26,31 +26,43 @@
...
+    _optional_signals = ["AWREGION", "AWLOCK", "AWCACHE", "AWPROT", "AWQOS",
+                         "WLAST",
+                         "ARREGION", "ARLOCK", "ARCACHE", "ARPROT", "ARQOS"]
 
-    def __init__(self, entity, name, clock, **kwargs):
+    _xRESP_msg = ("OKAY", "EXOKAY", "SLVERR", "DECERR")
+    _BURST_values = {"FIXED": 0b00, "INCR": 0b01, "WRAP": 0b10}
+
+    def __init__(self, entity: SimHandleBase, name: str, clock: SimHandleBase,
+                 **kwargs: Any):
```
eric-wieser, line 65: If you're willing to split this PR in two, we'd be more than happy to merge the type annotations by themselves.

cocotb/drivers/amba.py
cmarqu: (Also in the docstrings)

cocotb/drivers/amba.py
corna: Should `burst` (the one passed by the user) be an enum or a str?

eric-wieser: I'd lean towards enum

corna:
It should now contain all the required changes and a complete test.
The HDL files (AXI4 RAM, interconnect and registers) have been taken from [here](https://github.com/alexforencich/verilog-axi/),(MIT licensed).

corna:
Also, I don't know if I should use `yield` or `async`/`await` in the test, let me know.

eric-wieser:
At this point all new code should use `async`/`await`. Thanks for writing the tests, will review when I have time.

eric-wieser:
> The HDL files (AXI4 RAM, interconnect and registers) have been taken from here, (

It's tempting to use a `git submodule` for that, perhaps after forking the repo to the cocotb org just in case it disappears. Thoughts @imphil?

tests/designs/axi4_ram/Makefile
```diff
@@ -0,0 +1,58 @@
...
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL POTENTIAL VENTURES LTD BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
```
eric-wieser, line 28: We switched to a shorter license header:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause
```

cocotb/drivers/amba.py
```diff
@@ -26,29 +26,54 @@
...
-                "WVALID", "WREADY", "WDATA", "WSTRB",  # Write data channel
-                "BVALID", "BREADY", "BRESP",           # Write response channel
-                "ARVALID", "ARADDR", "ARREADY",        # Read address channel
-                "RVALID", "RREADY", "RRESP", "RDATA"]  # Read data channel
+    _signals = \
+        ["AWVALID", "AWADDR", "AWREADY", "AWID", "AWLEN", "AWSIZE", "AWBURST",
+         "WVALID", "WREADY", "WDATA", "WSTRB",
+         "BVALID", "BREADY", "BRESP", "BID",
+         "ARVALID", "ARADDR", "ARREADY", "ARID", "ARLEN", "ARSIZE", "ARBURST",
+         "RVALID", "RREADY", "RRESP", "RDATA", "RID", "RLAST"]
```
eric-wieser, line 72: As a general style comment, I would recommend avoiding `\` wherever possible. Here, you can do that as
```suggestion
    _signals = [
        "AWVALID", "AWADDR", "AWREADY", "AWID", "AWLEN", "AWSIZE", "AWBURST",
        "WVALID", "WREADY", "WDATA", "WSTRB",
        "BVALID", "BREADY", "BRESP", "BID",
        "ARVALID", "ARADDR", "ARREADY", "ARID", "ARLEN", "ARSIZE", "ARBURST",
        "RVALID", "RREADY", "RRESP", "RDATA", "RID", "RLAST"]
```

corna:
Done
I have to keep the `coroutine` decorator until #1846 is merged (or things like [this](https://github.com/cocotb/cocotb/pull/1248/commits/fd30ce3c8bd3fa1a57dd8a87010f50eb265b26f9#diff-9c2cea336a95c439e7060a39371adcb6R278) fail)

cocotb/drivers/amba.py
```diff
@@ -60,67 +85,164 @@ def __init__(self, entity, name, clock, **kwargs):
...
+            # Wait until acknowledged
+            while True:
+                await ReadOnly()
+                if self.bus.AWREADY.value:
+                    break
+                await RisingEdge(self.clock)
+            await RisingEdge(self.clock)
+            self.bus.AWVALID <= 0
+
+    @cocotb.coroutine
```
eric-wieser, line 184: No need for this any more
```suggestion
```

cocotb/drivers/amba.py
```diff
@@ -129,86 +251,229 @@ def write(
...
+            Raises:
+                ValueError: If any of the input parameters is invalid.
+                AXIProtocolError: If write response from AXI is not ``OKAY``.
+        """
+
+        try:
+            iter(value)
+            raise ValueError("AXI4Lite does not support burst transfers")
+        except TypeError:
+            await super(AXI4LiteMaster, self).write(
```
eric-wieser, line 450: Python 3 allows the shorter:
```suggestion
            await super().write(
```

cocotb/drivers/amba.py
```diff
@@ -60,67 +85,164 @@ def __init__(self, entity, name, clock, **kwargs):
...
+
+    @staticmethod
+    def _check_4kB_boundary_crossing(address, burst, size, length):
+        """Check that the provided burst does not cross a 4kB boundary."""
+        if burst is AXIBurst.INCR:
+            last_address = address + size * (length - 1)
+            if address & ~0xfff != last_address & ~0xfff:
+                raise ValueError(
+                    "INCR burst with start address {:#x} and last address "
+                    "{:#x} crosses the 4kB boundary {:#x}, which is forbidden"
```
eric-wieser, line 146:
```suggestion
                    "{:#x} crosses the 4kB boundary {:#x}, which is forbidden "
```
Missing space

cocotb/drivers/amba.py
```diff
@@ -60,67 +85,164 @@ def __init__(self, entity, name, clock, **kwargs):
...
+                    self.bus.WVALID <= 0
 
     @cocotb.coroutine
-    def write(
-        self, address: int, value: int, byte_enable: int = 0xf,
-        address_latency: int = 0, data_latency: int = 0, sync: bool = True
-    ) -> BinaryValue:
+    async def write(self, address, value, size=None, burst=AXIBurst.INCR,
+                    byte_enable=None, address_latency=0, data_latency=0,
+                    sync=True):
```
eric-wieser, line 230: Changing the argument order is a breaking change. Can you add the new arguments at the end?

cocotb/drivers/amba.py
corna: See my comment above

cocotb/drivers/amba.py
```diff
@@ -281,19 +275,18 @@ def write(self, address, value, size=None, burst=AXIBurst.INCR,
         write_data = self._send_write_data(value, data_latency, byte_enable,
                                            sync)
 
-        yield Combine(write_address, write_data)
+        await Combine(write_address, write_data)
```
eric-wieser, line 278: You can change this to
```suggestion
        await Combine(cocotb.fork(write_address), cocotb.fork(write_data))
```
which means you can drop the `coroutine`

corna: Isn't it going to break existing user code? E.g., in my test I have sometimes:
```
result = await First(axim.write(...), ClockCycles(...))
```
to check for a timeout in the write, however dropping `coroutine` on `write()` before #1846 is going to break it.

eric-wieser: I hadn't considered that, good point. That probably means I need to be more careful in #2022...

eric-wieser: For internal coroutines I'd still encourage you to remove the `@coroutine` in the way I suggest above, but I think you're right that the public API needs to remain decorated for now.

cocotb/drivers/amba.py
corna: The arguments of AXI4LiteMaster.write() did not change, the extra arguments are available only on the new AXI4Master.write() method.

eric-wieser: Ah, apologies, I missed that.

cocotb/drivers/amba.py
```diff
@@ -129,86 +251,229 @@ def write(
...
+                raise AXIProtocolError(
+                    err_msg.format(address, len(value), burst.name,
+                                   result.value, result.name), result)
+
+            return result
 
     @cocotb.coroutine
-    def read(self, address: int, sync: bool = True) -> BinaryValue:
+    async def read(self, address, length=1, size=None, burst=AXIBurst.INCR,
+                   return_rresp=False, sync=True):
```
eric-wieser, line 305: I'd recommend making these keyword only. I think `length` is just about obvious enough that `bus.read(0xFF0000, 4)` is worth allowing as a shorthand for `bus.read(0xFF0000, length=4)`, but for all the other arguments explicit names seem wise.
```suggestion
    async def read(self, address, length=1, *, size=None, burst=AXIBurst.INCR,
                   return_rresp=False, sync=True):
```

eric-wieser: Or actually, while you're here, could add type annotations:
```suggestion
    async def read(self, address: int, length: int = 1, *
                   size: Optional[int] = None,
                   burst: AXIBurst = AXIBurst.INCR,
                   return_rresp: bool = False,
                   sync: bool=True):
```

cocotb/drivers/amba.py
eric-wieser, line 230: Type annotations and keyword-only arguments would be good here too, something like

```suggestion
    async def write(self, address: int, value: Any, *,
                    size: Optional[int] = None,
                    burst: AXIBurst = AXIBurst.INCR,
                    byte_enable: Any=None,  # not sure what type this is
                    address_latency: int = 0,
                    data_latency: int = 0,
                    sync: bool = True) -> AXIxRESP:
```

corna: You previously asked for type annotations in a separate PR: should I add them in this one or in a new one?

eric-wieser: Ah, apologies, I forget what I said before. What I think I'd like to see, if possible, is a split of (in order):

* Convert the old code to use `async` / `await`. I already started on this in #2022, but skipped this file to avoid conflicts. I'd be fine with including it if you're happy to resolve those (hopefully trivial) conflicts later.
* Convert the old code to have type annotations.
* Everything else in this PR.

cocotb/drivers/amba.py
```diff
@@ -129,86 +251,229 @@ def write(
...
+            byte_enable: Which bytes in value to actually write. Defaults to
+                None (write all bytes).
+            address_latency: Delay before setting the address (in clock
+                cycles). Default is no delay.
+            data_latency: Delay before setting the data value (in clock
+                cycles). Default is no delay.
+            sync: Wait for rising edge on clock initially. Defaults to True.
+
+            Returns:
+                The write response value.
```
corna, line 439: I'd like some suggestions here.

Currently, AXI4LiteMaster.write() returns the write response value. However, this does not make sense, as any write response other than `OKAY` just makes it raise an `AXIProtocolError`. In other words, while the docstring promises to return the write response value, in reality it either throws an exception or it returns always the same value.

In my opinion, this method should not return anything.
Is it possibile to deprecate/remove the return value? Or are we forced to add `return AXIxRESP.OKAY` to maintain backwards compatibility?

cocotb/drivers/amba.py
corna: * Feel free to do it in #2022, I'll rebase this PR on it.
 * The old code already has type annotations, at least for public methods. I'll include the type annotations for AXI4Master in this PR.

corna:
@eric-wieser This PR should be complete if you do not have any other request, let me know

cocotb/drivers/amba.py
cmarqu: I guess it would be safest to continue returning the response; the overhead is negligible.

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -0,0 +1,370 @@
...
+                "Read {:#x} at beat {}/{} with starting address {:#x}, but "
+                "was expecting {:#x} ({:#x} with {:#x} as strobe and {:#x} as "
+                "previous value)"
+                .format(read_values[i].integer, i + 1, burst_length, address,
+                        expected_values[i], write_values[i], strobes[i],
+                        previous_values[i].integer))
+
+
+async def test_unmapped(dut, driver):
+    """Chech whether a read/write over an unmapped address returns a DECERR"""
```
cmarqu, line 134:
```suggestion
    """Check whether a read/write over an unmapped address returns a DECERR"""
```

cmarqu:
@corna Could you add the ``AXI4Master`` class to ``documentation/source/library_reference.rst`` so that it shows up in https://cocotb--1248.org.readthedocs.build/en/1248/library_reference.html#implemented-testbench-structures?
This makes it easier to spot possible docstring rendering problems.

All in all, thanks very much for this highest-quality contribution!
I'll add some small comments still and approve then.

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,401 @@ def __init__(self, entity, name, clock, **kwargs):
...
+                the data bus).
+            burst: The burst type, either ``FIXED``, ``INCR`` or ``WRAP``.
+                Defaults to ``INCR``.
+            return_rresp: Return the list of RRESP values, instead of raising a
+                AXIProtocolError in case of not OKAY. Defaults to False.
+            sync: Wait for rising edge on clock initially. Defaults to True.
 
         Returns:
-            The read data value.
+            The read data values or, if **return_rresp** is True, a list of
```
cmarqu, line 326:
```suggestion
            The read data values or, if *return_rresp* is True, a list of
```
Single asterisks (i.e. italics) is convention.

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,401 @@ def __init__(self, entity, name, clock, **kwargs):
...
+            pairs each containing the data and RRESP values.
 
         Raises:
-            AXIProtocolError: If read response from AXI is not ``OKAY``.
+            ValueError: If any of the input parameters is invalid.
+            AXIProtocolError: If read response from AXI is not ``OKAY`` and
+                **return_rresp** is False
```
cmarqu, line 332:
```suggestion
                *return_rresp* is False
```

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -0,0 +1,370 @@
...
+            "Write with length={} has been performed by the driver, but "
+            "burst operations are not allowed on AXI4-Lite".format(length))
+
+    except ValueError:
+        pass
+
+
+@cocotb.test()
+async def test_read_length_mismatch(dut):
+    """That a mismatch in the read data length is correctly identified"""
```
cmarqu, line 318:
```suggestion
    """Test that a mismatch in the read data length is correctly identified"""
```

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -0,0 +1,370 @@
...
+    for length_delta in (-1, 1):
+        try:
+            # Override the driver's AWLEN value, forcing a wrong one
+            await RisingEdge(dut.clk)
+            getattr(dut, AXI_PREFIX + '_AWLEN') <= burst_length + length_delta
+            await axim.read(address, burst_length)
+
+            raise TestFailure("Mismatch between ARLEN value ({}) and number "
+                              "of read words ({}), but the driver did not "
+                              "raise and exception"
```
cmarqu, line 344:
```suggestion
                              "raise an exception"
```

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -0,0 +1,370 @@
...
+
+    except ValueError:
+        pass
+
+
+@cocotb.test()
+async def test_read_length_mismatch(dut):
+    """That a mismatch in the read data length is correctly identified"""
+
+    class AXI4Master_umanaged_arlen(AXI4Master):
```
cmarqu, line 320:
```suggestion
    class AXI4Master_unmanaged_arlen(AXI4Master):
```

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -0,0 +1,370 @@
...
+        _signals = [signal for signal in AXI4Master._signals
+                    if signal != "AWLEN"]
+
+    axim = AXI4Master_umanaged_arlen(dut, AXI_PREFIX, dut.clk)
```
cmarqu, line 324:
```suggestion
    axim = AXI4Master_unmanaged_arlen(dut, AXI_PREFIX, dut.clk)
```

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -0,0 +1,370 @@
...
+
+CLK_PERIOD = (10, "ns")
+AXI_PREFIX = "S_AXI"
+
+
+def get_parameters(dut):
+    address_width = dut.ADDR_WIDTH.value.integer // 8
+    data_width = dut.DATA_WIDTH.value.integer // 8
+    ram_first = dut.RAM_BASE_ADDRESS.value.integer
+    ram_last = ram_first + 2**dut.RAM_WIDTH.value.integer
```
cmarqu, line 28: From this alone, I thought you had an off-by-one error, but since you use the value in ``randrange`` (https://docs.python.org/3/library/random.html#random.randrange), the ``ram_last`` value itself is not included.
It might be clearer to use the terminology of ``(rand)range``and call these values ``ram_addr_start`` and ``ram_addr_stop``.

tests/designs/axi4_ram/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+else
+
+TOPLEVEL := top
+
+PWD=$(shell pwd)
+
+COCOTB?=$(PWD)/../../..
+
+VERILOG_SOURCES = \
```
cmarqu, line 21: Personally, I like to append multiple lines using ``VERILOG_SOURCES += ...`` repeatedly because it enables adding comments in between, but that's a matter of preference.

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,401 @@ def __init__(self, entity, name, clock, **kwargs):
...
+            Returns:
+                The write response value.
+
+            Raises:
+                ValueError: If any of the input parameters is invalid.
+                AXIProtocolError: If write response from AXI is not ``OKAY``.
+        """
+
+        if isinstance(value, collections.abc.Sequence):
+            raise ValueError("AXI4Lite does not support burst transfers")
```
cmarqu, line 455:
```suggestion
            raise ValueError("AXI4-Lite does not support burst transfers")
```

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,401 @@ def __init__(self, entity, name, clock, **kwargs):
...
         Args:
             address: The address to read from.
-            sync: Wait for rising edge on clock initially.
-                Defaults to True.
+            length: Number of words to transfer. Defaults to 1.
+            size: The size (in bytes) of each beat. Defaults to None (width of
+                the data bus).
+            burst: The burst type, either ``FIXED``, ``INCR`` or ``WRAP``.
+                Defaults to ``INCR``.
+            return_rresp: Return the list of RRESP values, instead of raising a
```
cmarqu, line 321:
```suggestion
            return_rresp: Return the list of RRESP values, instead of raising an
```

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,401 @@ def __init__(self, entity, name, clock, **kwargs):
...
-
-    async def _send_write_data(self, data, delay=0, byte_enable=0xF):
-        """Send the write address, with optional delay (in clocks)."""
-        await self.write_data_busy.acquire()
-        for cycle in range(delay):
+        # Set the default value (0) for the unsupported signals, which
+        # translate to:
+        #  * Transaction IDs to 0
+        #  * Region identifier to 0
+        #  * Normal (non exclusive) access
```
cmarqu, line 96:
```suggestion
        #  * Normal (non-exclusive) access
```

corna:
All done.
I have also changed a couple of `randrange` -> `randint`.

Thanks to you and all the cocotb contributors!

ktbarrett (commented):
More reviews incoming. @andrewandrepowell You know more about this AXI stuff than me.

cocotb/drivers/amba.py
```diff
@@ -26,31 +26,60 @@
...
+    _signals = [
+        "AWVALID", "AWADDR", "AWREADY", "AWID", "AWLEN", "AWSIZE", "AWBURST",
+        "WVALID", "WREADY", "WDATA", "WSTRB",
+        "BVALID", "BREADY", "BRESP", "BID",
+        "ARVALID", "ARADDR", "ARREADY", "ARID", "ARLEN", "ARSIZE", "ARBURST",
+        "RVALID", "RREADY", "RRESP", "RDATA", "RID", "RLAST"]
+
+    _optional_signals = ["AWREGION", "AWLOCK", "AWCACHE", "AWPROT", "AWQOS",
+                         "WLAST",
+                         "ARREGION", "ARLOCK", "ARCACHE", "ARPROT", "ARQOS"]
```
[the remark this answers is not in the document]
corna: They are grouped by channel (AW, W and AR).

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,401 @@ def __init__(self, entity, name, clock, **kwargs):
...
+                raise ValueError(
+                    "INCR burst with start address {:#x} and last address "
+                    "{:#x} crosses the 4kB boundary {:#x}, which is forbidden "
+                    "in a single burst"
+                    .format(address, last_address,
+                            (address & ~0xfff) + 0x1000))
+
+    async def _send_write_address(
+        self, address: int, length: int, burst: AXIBurst, size: int,
+        delay: int, sync: bool
```
ktbarrett, line 158: I understand what `sync` is *doing*, but I'm not sure what the point is. When would you not want to sync?

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,401 @@ def __init__(self, entity, name, clock, **kwargs):
...
-        await RisingEdge(self.clock)
-        self.bus.ARVALID <= 0
+        async with self.read_address_busy:
+            if sync:
+                await RisingEdge(self.clock)
+
+            self.bus.ARADDR <= address
+            self.bus.ARVALID <= 1
+
+            if hasattr(self.bus, "AWLEN"):
```
ktbarrett, line 352:
```suggestion
            if hasattr(self.bus, "ARLEN"):
```

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,401 @@ def __init__(self, entity, name, clock, **kwargs):
...
+                self.bus.ARLEN <= length - 1
+
+            if hasattr(self.bus, "ARSIZE"):
+                self.bus.ARSIZE <= size.bit_length() - 1
+
+            if hasattr(self.bus, "AWBURST"):
```
ktbarrett, line 358:
```suggestion
            if hasattr(self.bus, "ARBURST"):
```

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,401 @@ def __init__(self, entity, name, clock, **kwargs):
...
+        #  * Normal (non-exclusive) access
+        #  * Device non-bufferable access
+        #  * Unprivileged, secure data access
+        #  * No QoS
+        unsupported_signals = [
+            "AWID", "AWREGION", "AWLOCK", "AWCACHE", "AWPROT", "AWQOS",
+            "ARID", "ARREGION", "ARLOCK", "ARCACHE", "ARPROT", "ARQOS"]
+        for signal in unsupported_signals:
+            try:
+                getattr(self.bus, signal).setimmediatevalue(0)
```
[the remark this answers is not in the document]
corna: I don't think so: these signals simply specifies additional behaviors which are not currently supported by the driver, but the defaults are the one dictated by the specification, which are in general the most restrictive. No slave should misbehave due to these defaults.

cocotb/drivers/amba.py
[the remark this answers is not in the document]
ktbarrett: No I mean this variable is right-aligned and the previous variable starts the list on next line and is single indented.

cocotb/drivers/amba.py
corna: No idea, `sync` was already present in the AXI4-Lite driver so I kept it. Maybe if you are already on a clock rising edge from a previous operation and you don't want to wait another one.

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,401 @@ def __init__(self, entity, name, clock, **kwargs):
...
+        write_address = self._send_write_address(address, len(value), burst,
+                                                 size, address_latency, sync)
 
-        return result
+        write_data = self._send_write_data(value, data_latency, byte_enable,
+                                           sync)
+
+        await Combine(cocotb.fork(write_address), cocotb.fork(write_data))
+
+        async with self.write_response_busy:
```
ktbarrett, line 285: Can write responses can come back immediately? If you waiting until after the address and data are written out to start checking the response channel may lead to the response being missed?

ktbarrett (commented):
It's apparently requiring me to have text here now :/

cocotb/drivers/amba.py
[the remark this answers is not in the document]
ktbarrett: It's a small nit. Right-aligned code triggers me... Consider this not actionable.

cocotb/drivers/amba.py
corna: No, that's forbidden by the standard.

A3.3.1: "The slave must wait for AWVALID, AWREADY, WVALID, and WREADY to be asserted before asserting BVALID. The slave must also wait for WLAST to be asserted before asserting BVALID because the write response, BRESP must be signaled only after the last data transfer of a write transaction"

cmarqu:
Not sure if this is the appropriate place to put it... https://osvvm.org/archives/1688 writes:

> Hence, to support and verify the full capability of an AXI Interface, a verification component needs to support independent (aka concurrent or multi-threaded) execution of the Write Address, Write Data, Write Response, Read Address, and Read Data interfaces.  At the same time, it also needs to correlate aspects of the same transaction (such as a write).  Sometimes the verification component will need to block until a transaction completes and sometimes not.  Sometimes the write address will need to be dispatched independently from the write data (to introduce a skew between them) and then optionally queue the next set of transactions without waiting.
> An interesting model indeed.  Unfortunately, most open source verification components fall short and do not support independent execution of the AXI interfaces. 

Is this master there as well?

ktbarrett:
@cmarqu Nice find! What they are describing is how we do our AXI agents. We have the full gamut: AXIFull reader, writer, and observer. We build on top of 5 instance of AXIStream agents for each of the channels, which are operating independently, We support blocking read and write operations by having the AXIFull agent maintain a queue of outstanding request as Futures. It works very well.

I think via use of Locks @corna's implementation can achieve the same: multiple write/read operations can be issued and forked all at once and the Lock will order their service correctly. That's probably worth a test.

corna:
> I think via use of Locks @corna's implementation can achieve the same: multiple write/read operations can be issued and forked all at once and the Lock will order their service correctly while allowing full bandwidth. That's probably worth a test.

It should work indeed. It had a bug, fixed in 3127962.

Now it works as expected, as confirmed by the test:

axi_concurrent

One current missing feature is the support for out-of-order operations (aka, AWID, BID, ARID, RID). For now, all operations issued have to be executed in order by the slave (but most of the slaves always execute the operation in-order anyways).

ktbarrett:
@corna Looking at your diagram it looks like it's only achieving 50% bandwidth. AWREADY stays high after the first write request goes out, but the driver must lower AWVALID for a clock cycle before it goes back high. AXI allows you send messages back to back, and in fact this is the only way to achieve 100% bandwidth utilization with AXI Lite.

I don't think this needs to be remedied in this PR, but it is something to consider going forward.

garmin-mjames:
@ktbarrett Is that the fault of `sync=True`? I checked the coverage earlier and `sync=False` is not covered.

ktbarrett:
The logic in the `aw_write` holds the lock while it writes the request out *and* waits for the next cycle to set the valid low. It *could* work by first using `sync=True` and on following usages using `sync=False`, but that's a hack IMO and requires manual users interaction which is less than ideal.

What we do is have the interface running in a separate Task that will lower the valid the cycle after a transaction occurs if no new transaction comes in on the next clock. The blocking `write` coroutine returns as soon as the transaction occurs. So if you call `write` immediately after the previous `write` returns it will synchronize on the next clock, keep the valid high and put the new data on the bus. 

That behavior we have encapsulated in our AXI Stream agents so we only have to worry about protocol details in the AXI Full agent.

cocotb/drivers/amba.py
```diff
@@ -60,153 +89,403 @@ def __init__(self, entity, name, clock, **kwargs):
...
+                self.bus.AWSIZE <= size.bit_length() - 1
+
+            # Wait until acknowledged
+            while True:
+                await ReadOnly()
+                if self.bus.AWREADY.value:
+                    break
+                await RisingEdge(self.clock)
             await RisingEdge(self.clock)
+            self.bus.AWVALID <= 0
```
garmin-mjames, line 188:
```suggestion
            while True:
                await ReadOnly()
                if self.bus.AWREADY.value:
                    break
            await RisingEdge(self.clock)
            self.bus.AWVALID <= 0
```
I don't think you need the `RisingEdge` in the loop. The coroutine may run on more timesteps but you should see the *AWREADY* value at the end of the last timestep before the edge of the clock.

Alternatively, couldn't you just do this:
```py3
            while True:
                yield RisingEdge(self.clock)
                if self.bus.AWREADY.value:
                    break
            self.bus.AWVALID <= 0
```
Either would allow back to back writes assuming you don't use sync.

garmin-mjames:
@ktbarrett The problem is that you have to wait until after the `RisingEdge` for the transaction to finish, so any use of `sync=True` forces at best every other clock, because we've already missed the edge trigger when the next transaction starts.

ktbarrett:
> The problem is that you have to wait until after the RisingEdge for the transaction to finish.
Why's that?

garmin-mjames:
The handshake (both VALID/READY asserted at rising clock edge) is when the "transfer occurs".
The purpose of `sync` is to wait until the rising edge of a clock before asserting VALID, instead of asserting it immediately when `write` is called.
Returning before the handshake clock edge just so that `sync=True` works in back-to-back writes seems odd.

ktbarrett:
> The handshake (both VALID/READY asserted at rising clock edge) is when the "transfer occurs"

I think that's a little academic. By convention we know we aren't going to modify the signal until the next clock edge. To parallel HDL: the process will have completed shortly after the rising edge of the clock.

> instead of asserting it immediately when write is called.

That's a bit odd to me, and we have had issues with this at my place of work. There is the possibility of a signal changing during the middle of a clock edge on the first write in a burst due to `Timer`s, `NextTimeStep`s, waiting on value changes of combo signals, etc, etc. There are a lot of ways to get unsynchronized.

The way around that is always remembering to synchronize before doing the first write. People usually don't do that and don't *know* to do that. They expect the reuse component to handle it for them and not violate their assumptions. Perhaps that's just lazy, but keeping track of where you are in time isn't trivial, people tend to not do it, and the tools don't help you.

garmin-mjames:
You have some valid points.
What happens if someone else gets the Lock and does a non-sync write?

ktbarrett:
@garmin-mjames I think in this case it will just cause a non-synchronized transition on those signals. As long as the HDL being stimulated doesn't do anything "bad" like react combinationally to inputs there should be no problem. There is also the possibility that cocotb code won't be executed in the order you think it will.

We've definitely had a lot of arguments on what's the "best" or "most correct" way of implementing cocotb interface code for reuse. Sometime these differences cause issues when used together 😩. I don't claim to know everything, I'm still figuring it out.

garmin-mjames:
@ktbarrett I think the data from the first transaction will be overwritten by the second before the first transaction finishes, which is unexpected behavior. If the `Lock` mechanisms don't handle bus contention properly they aren't working.

imphil:
Thanks for all your discussion so far. Even though this PR might not be perfect, it's still a clear step forward, and we should get it merged as soon as all major issues are sorted out. Let's file follow-up issues for all remaining issues.

@ktbarrett @garmin-mjames can you have a look which things *must* get resolved before merging, and then do a final review round to get this merged?

garmin-mjames:
I would like to see the tests get full coverage of the new driver. It's almost there but there are a few missing cases according to codecov:

- Valid `AXIBurst.WRAP`
- Valid `AXIBurst.FIXED`
- Read and Write with valid `size` parameter
- Read and Write with `sync=False`

As far as supporting back-to-back transactions, that may require additional work and specifying all cases and the expected behavior.

corna:
It is complete now (only some minor lines are missing).

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -131,6 +131,87 @@ async def test_burst(dut, return_rresp):
...
+    await setup_dut(dut)
+
+    base = randrange(ram_start, ram_stop, 4096)
+    offset = randrange(0, 4096 - (burst_length - 1) * data_width, data_width)
+    address = base + offset
+
+    for burst in (AXIBurst.FIXED, AXIBurst.WRAP):
+
+        write_values = \
+            [randrange(0, 2**(data_width * 8)) for i in range(burst_length)]
```
garmin-mjames, line 153:
```suggestion
        write_values = [
            randrange(0, 2**(data_width * 8))
            for _ in range(burst_length)
        ]
```

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -131,6 +131,87 @@ async def test_burst(dut, return_rresp):
...
+
+    if data_width == 1:
+        raise TestFailure("This test cannot be executed with data_width == 1")
+
+    await setup_dut(dut)
+
+    burst_length = (randrange(2, 256) // 2) * 2
+    address = ram_start
+    write_values = \
+        [randrange(0, 2**(data_width * 8 // 2)) for i in range(burst_length)]
```
garmin-mjames, line 194:
```suggestion
    write_values = [
        randrange(0, 2**(data_width * 8 // 2)) 
        for _ in range(burst_length)
    ]
```

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -131,6 +131,87 @@ async def test_burst(dut, return_rresp):
...
+
+    await axim.write(address, write_values, size=data_width // 2)
+
+    # This slave does not support unaligned transfers or size different than
+    # the data bus width and simply drops the address' least significant bits.
+    # So, for each pair of writes, the second one overwrites the first one.
+    expected_values = \
+        [write_values[(i // 2) * 2 + 1] for i in range(len(write_values))]
```
garmin-mjames, line 202:
```suggestion
    expected_values = [
        write_values[(i // 2) * 2 + 1]
        for i in range(len(write_values))
    ]
```

garmin-mjames (commented):
Thanks for adding test cases!

ktbarrett (approved):
Thanks for adding more tests! The coverage is now near 100% for your additions/changes. I don't see the need to nitpick code style too much.

alexforencich:
priority_encoder.v should probably be updated to the current non-recursive one.  And the bus name matching really needs to be case-insensitive, at least for the non-user-specified part.  All caps is for constants only!

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -0,0 +1,451 @@
...
+
+    burst_length = randrange(2, 256)
+    base = randrange(ram_start, ram_stop, 4096)
+    offset = randrange(0, 4096 - (burst_length - 1) * data_width, data_width)
+    address = base + offset
+    write_values = \
+        [randrange(0, 2**(data_width * 8)) for i in range(burst_length)]
+
+    # Make strobe one item less than data, to test also that driver behavior
+    # This RAM implementation does not honor WSTRB=0
```
alexforencich, line 98: This should not be the case, WSTRB=0 should be handled correctly by axi_ram.v

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -0,0 +1,451 @@
...
+    burst_length = (randrange(2, 256) // 2) * 2
+    address = ram_start
+    write_values = \
+        [randrange(0, 2**(data_width * 8 // 2)) for i in range(burst_length)]
+
+    await axim.write(address, write_values, size=data_width // 2)
+
+    # This slave does not support unaligned transfers or size different than
+    # the data bus width and simply drops the address' least significant bits.
+    # So, for each pair of writes, the second one overwrites the first one.
```
alexforencich, line 200: AXI does not support unaligned transfers.  The lower-order bits are only used for narrow bursts.

tests/test_cases/test_axi4/test_axi4.py
```diff
@@ -0,0 +1,451 @@
...
+single_beat_with_latency.add_option('address_latency', (0, 5))
+single_beat_with_latency.add_option('data_latency', (1, 10))
+single_beat_with_latency.generate_tests(postfix="_latency")
+
+incr_burst = TestFactory(test_incr_burst)
+incr_burst.add_option('return_rresp', (True, False))
+incr_burst.generate_tests()
+
+# AXI4Master cannot be tested, as axi_interconnect.v acknowledges only one
+# WDATA for unmapped addresses in some cases
```
alexforencich, line 444: If this is a bug in the interconnect, then it should be tracked down and fixed.  If you can provide more info, that would be great!

corna: You can reproduce the bug with:

```
@cocotb.test()
async def test_bug(dut):
    axim = AXI4Master(dut, AXI_PREFIX, dut.clk)
    ram_stop = get_parameters(dut)[3]

    await setup_dut(dut)
    await First(ClockCycles(dut.clk, 32), axim.write(ram_stop, [0x00000000] * 4))
```

This is the result:
bug

Let me know if you need more details.

tests/test_cases/test_axi4/test_axi4.py
corna: I had this issue where all the WSTRB were handled correctly, except for all WSTRB=0 which was treated as all ones. However I can not reproduce it anymore, so I do not know if it was a mistake in my test code or not.

As it does not appear here, I will use the full WSTRB range in this test.

corna:
> priority_encoder.v should probably be updated to the current non-recursive one.

Done

> And the bus name matching really needs to be case-insensitive, at least for the non-user-specified part. All caps is for constants only!

Honestly I'm not sure how to proceed here: the AXI4 specification uses all-uppercase names, which I usually reserve for constants as well. If we decide to make the bus name matching case-insensitive, this should be optional and should be implemented in the parent class.

tests/test_cases/test_axi4/test_axi4.py
corna: You are right, bad wording. And I did not handle correctly narrow bursts anyways, so this has to be fixed. Thanks.

alexforencich:
Agreed, see #2137 / #2138 for case-insensitive naming

tests/test_cases/test_axi4/test_axi4.py
alexforencich: Yep, that's a bug alright.  In STATE_WRITE_DROP, please try ANDing in wlast like so:

    if (current_s_axi_wready && current_s_axi_wvalid && current_s_axi_wlast) begin

and let me know if that fixes the problem.

corna: Yes, it fixes it:
bug

alexforencich: Excellent; that fix is now in the upstream repo.

ktbarrett:
@alexforencich If you are happy with the changes, could you add an approval review? Are you and @corna interested in getting #2138 in before okaying this PR? Or is that something that can come in after this PR is accepted?

alexforencich:
Honestly, it doesn't really matter to me as I'm most likely just going to use my own AXI code here: https://github.com/alexforencich/cocotbext-axi which supports 100% throughput, ID signals, reordering, multiple in-flight operations, etc.  The default names are all lowercase in mine, so I'm not blocked on #2138.  I also dropped using BusDriver and BusMonitor so I'm not blocked on #2151, not to mention I don't really like the interface provided by BusDriver and BusMonitor and I wasn't using it anyway.  Although I did get rid of some mixins I was using, so I might have been blocked on #2151 anyway.

corna:
We can create another PR after #2138, I'd like to see this PR merged without further edits.

ktbarrett:
@corna It doesn't pass regressions.

alexforencich:
See figures A3-13, A3-14, and A3-15 in the AXI spec for more details on how unaligned operations are handled.  

For the user-facing API, there is a question of what the best format should be.  Personally, I think raw bytes is the way to go, divorced from any notion of the bus width - i.e. use a bytearray, and pack/unpack into transfers of the appropriate width within the model.  Right now, you're passing through the individual transfers, but you're modifying them in inconsistent ways, which is just going to be a real pain to deal with.

corna:
> See figures A3-13, A3-14, and A3-15 in the AXI spec for more details on how unaligned operations are handled.

Indeed, that was what I followed.

I don't particularly like returning a single bytearray, as, in most of the cases, the user just wants a list of words, returning a bytearray forces him to manually slice it. Unaligned read/writes are more like a special case.

I see three possible choices for the user API:
 * ignore the fact that we are performing an unaligned transaction and simply return/write the whole data bus: easy, but may lead the user to violate the standard (like performing an unaligned write and provide a all-one WSTRB) - (`[0x11223344, 0x55667788, 0x99aabbcc]`)
 * return the value of the data bus for each beat but mask and shift the invalid bits (current solution): prevents any protocol violation, but is not very intuitive for the user (for example, the first word is shorter than the others) - (`[0x112233, 0x55667788, 0x99aabbcc]`)
 * fully reconstruct the words: great user interface, but now the words are not mapped 1:1 to the number of beats (the user requires N words and we have to perform N+1 beats) and the FIXED bursts are problematic - (`[0x88112233, 0xcc556677, ...]`)

Opinions?

alexforencich:
I don't like returning word-size transfers as in many cases I am going to be reading and writing blocks of binary data, and working with words means I have to manually pack and unpack into words, and this needs to be done differently based on the alignment and word size of the interface.  Might as well push that up into the driver to make things simpler.  However, providing read_dwords/write_dwords etc. wrappers might be useful, where the wrapper will convert a list of ints into a bytearray or vise-versa internally.   

On the flip side, I can see the argument for accessing the signals directly in case you want to do something really weird, or test things like early-aborted writes (where WSTRB of all 0 is sent) or other corner cases.  It seems like there are maybe at least two modes where you might want to use one of these models, one where you basically want to do memcpy or similar and just write data to various registers or memory regions, and another where you want to exercise all sorts of AXI protocol corner-cases.  

- Providing an all-1 WSTRB with an unaligned operation is not a violation of the standard, at least as far as I am aware.  The low-order bits only affect narrow bursts and are otherwise ignored.  Even with a narrow burst, providing all-1 or all-0 WSTRB is fine, you just might overwrite data from a previous cycle.  

- This is a really bad idea.  If you're returning words, either shift all of them so you get [0x88112233, 0xcc556677, 0x99aabb] or don't do any shifting at all.  Masking is fine, but isn't necessary - extra write data should be ignored by the destination, and extra read data should be ignored by the application.  Bit shifting just creates confusion.  

- This isn't a particularly great idea, unless it's divorced from the actual bus width - i.e. it's implemented as a wrapper around a byte-based interface.  

I think "proper" handling of fixed bursts is not a problem, if you're doing anything other than an INCR burst, things need to be done with the necessary care.  For instance, an unaligned FIXED burst doesn't really make sense - the low order address bits get discarded, so it isn't really even possible to do an unaligned FIXED burst.

corna:
> * Providing an all-1 WSTRB with an unaligned operation is not a violation of the standard, at least as far as I am aware. The low-order bits only affect narrow bursts and are otherwise ignored. Even with a narrow burst, providing all-1 or all-0 WSTRB is fine, you just might overwrite data from a previous cycle.

If I interpret the spec correctly, this is forbidden (A3.4.3 - Unaligned transfers):
```
The information on the low-order address lines must be consistent with the information on the byte lane strobes.
```
However I agree: as the slaves typically ignore the low order bits in non-narrow bursts, this should not generate any issue.

> * This is a really bad idea. If you're returning words, either shift all of them so you get [0x88112233, 0xcc556677, 0x99aabb] or don't do any shifting at all. 

This is a good suggestion, I think I will go this way. Thanks!

> I think "proper" handling of fixed bursts is not a problem, if you're doing anything other than an INCR burst, things need to be done with the necessary care. For instance, an unaligned FIXED burst doesn't really make sense - the low order address bits get discarded, so it isn't really even possible to do an unaligned FIXED burst.

The specification describes it, I will just copy it (i.e., return the high `size - address % size` B for each word on reads and mask the low bytes on writes)

Thanks for your suggestions, I will update the PR soon.

alexforencich:
TBH, I don't like shifting and returning words either.  I think there are only two reasonable options: 1. deal in full-width words, and don't manipulate those words at all - no masking, no shifting - or 2. deal in bytes, pack those into words transparently based on the bus width and address alignment, and possibly provide wrappers around read() and write() that can accept/return words of arbitrary size.

ktbarrett:
@corna @alexforencich We would like to have this PR merged soon. After it is merged we are planning on moving all bus and testbenching code into another repo in the organization (cocotb-bus). The 1.5 release will occur the first of the new year and we would like to do that move before the release. To give us ample time to do the move and all the other ancillary tasks, we would like this PR in a final working order for merging by the beginning of December.

This doesn't mean it has to be 100% perfect by then, the cocotb-bus repo may have a different release schedule, and more iteration can be done in the other repo before an official release. As long as the code is in working order in 2 weeks we will merge it. If you think it's good now, or won't have time before then, just say the word.

corna:
Should be complete now, I do not have anything to add at this phase and I am not aware of any bug or major missing feature.
For me it is ready, let me know.

Again, note that d46fa17 breaks the previous behavior for unaligned read and writes, which however are now much more easy for the user.

PR #1249, opened 2019-12-20
Added Waitable to the library reference.

Fatsie:


eric-wieser:
Sure, why not - I was leaving this out because I wasn't sure whether to make it public or private, but I assume people are depending on it now anyway, and we can always deprecate it if we decide its not needed.

PR #1250, opened 2019-12-20
Conditional waiting on events.

Fatsie:
This is alternative pull request to #1214 implementing some proposed changes by @eric-wieser to syntax and structure. This is curently implemented on top of #1249.

@imphil, @themperek would really love to have some feedback if this is considered appropriate for cocotb.

### DIFF
--- a/cocotb/drivers/__init__.py
+++ b/cocotb/drivers/__init__.py
@@ -30,11 +30,11 @@
 """Set of common driver base classes."""
 
 from collections import deque
+import warnings
 
 import cocotb
 from cocotb.decorators import coroutine
-from cocotb.triggers import (Event, RisingEdge, ReadOnly, NextTimeStep,
-                             Edge)
+from cocotb.triggers import Event, RisingEdge, NextTimeStep, Edge
 from cocotb.bus import Bus
 from cocotb.log import SimLog
 
@@ -262,10 +262,12 @@ class BusDriver(Driver):
         to move to :class:`~cocotb.triggers.NextTimeStep` before
         registering more callbacks can occur.
         """
-        yield ReadOnly()
-        while signal.value.integer != 1:
-            yield RisingEdge(signal)
-            yield ReadOnly()
+        warnings.warn(
+            "Use of BusDriver._wait_for_signal() is deprecated\n"
+            "\tPlease, use cocotb.triggers.WaitCondition or event.until() instead",
+            DeprecationWarning, stacklevel=2
+        )
+        yield RisingEdge(signal).until(lambda: signal.value.integer == 1)
         yield NextTimeStep()
 
     @coroutine
@@ -276,10 +278,12 @@ class BusDriver(Driver):
         to move to :class:`~cocotb.triggers.NextTimeStep` before
         registering more callbacks can occur.
         """
-        yield ReadOnly()
-        while signal.value.integer != 0:
-            yield Edge(signal)
-            yield ReadOnly()
+        warnings.warn(
+            "Use of BusDriver._wait_for_signal() is deprecated\n"
+            "\tPlease, use cocotb.triggers.WaitCondition or event.until() instead",
+            DeprecationWarning, stacklevel=2
+        )
+        yield Edge(signal).until(lambda: signal.value.integer == 0)
         yield NextTimeStep()
 
     def __str__(self):
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -76,12 +76,7 @@ class AXI4LiteMaster(BusDriver):
         self.bus.AWADDR <= address
         self.bus.AWVALID <= 1
 
-        while True:
-            yield ReadOnly()
-            if self.bus.AWREADY.value:
-                break
-            yield RisingEdge(self.clock)
-        yield RisingEdge(self.clock)
+        yield RisingEdge(self.clock).until_after(lambda: self.bus.AWREADY.value)
         self.bus.AWVALID <= 0
         self.write_address_busy.release()
 
@@ -96,12 +91,7 @@ class AXI4LiteMaster(BusDriver):
         self.bus.WVALID <= 1
         self.bus.WSTRB <= byte_enable
 
-        while True:
-            yield ReadOnly()
-            if self.bus.WREADY.value:
-                break
-            yield RisingEdge(self.clock)
-        yield RisingEdge(self.clock)
+        yield RisingEdge(self.clock).until_after(lambda: self.bus.WREADY.value)
         self.bus.WVALID <= 0
         self.write_data_busy.release()
 
@@ -143,12 +133,10 @@ class AXI4LiteMaster(BusDriver):
             yield c_data.join()
 
         # Wait for the response
-        while True:
-            if self.bus.BVALID.value and self.bus.BREADY.value:
-                result = self.bus.BRESP.value
-                break
-            yield RisingEdge(self.clock)
-            yield ReadOnly()
+        yield RisingEdge(self.clock).until(
+            lambda: self.bus.BVALID.value and self.bus.BREADY.value,
+        )
+        result = self.bus.BRESP.value
 
         yield RisingEdge(self.clock)
 
@@ -166,12 +154,7 @@ class AXI4LiteMaster(BusDriver):
         self.bus.ARADDR <= address
         self.bus.ARVALID <= 1
 
-        while True:
-            yield ReadOnly()
-            if self.bus.ARREADY.value:
-                break
-            yield RisingEdge(self.clock)
-        yield RisingEdge(self.clock)
+        yield RisingEdge(self.clock).until_after(lambda: self.bus.ARREADY.value)
         self.bus.ARVALID <= 0
         self.read_address_busy.release()
 
@@ -196,19 +179,17 @@ class AXI4LiteMaster(BusDriver):
         yield self._send_read_address(address)
 
         # Wait for the response
-        while True:
-            if self.bus.RVALID.value and self.bus.RREADY.value:
-                data = self.bus.RDATA.value
-                result = self.bus.RRESP.value
-                break
-            yield RisingEdge(self.clock)
-            yield ReadOnly()
+        yield RisingEdge(self.clock).until(
+            lambda: self.bus.RVALID.value and self.bus.RREADY.value
+        )
+        data = self.bus.RDATA.value
+        result = self.bus.RRESP.value.integer
 
         yield RisingEdge(self.clock)
 
-        if int(result):
+        if result:
             raise AXIProtocolError("Read address 0x%08x failed with RRESP: %d" %
-                               (address, int(result)))
+                                   (address, result))
 
         raise ReturnValue(data)
 
@@ -274,13 +255,9 @@ class AXI4Slave(BusDriver):
         clock_re = RisingEdge(self.clock)
 
         while True:
-            while True:
-                self.bus.WREADY <= 0
-                yield ReadOnly()
-                if self.bus.AWVALID.value:
-                    self.bus.WREADY <= 1
-                    break
-                yield clock_re
+            self.bus.WREADY <= 0
+            yield clock_re.until(lambda: self.bus.AWVALID.value)
+            self.bus.WREADY <= 1
 
             yield ReadOnly()
             _awaddr = int(self.bus.AWADDR)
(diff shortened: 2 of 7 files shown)


Fatsie:
Handled linting and fix ``TriggerHelper`` methods.

Fatsie:
Fix AXI4Lite test.

themperek:
@Fatsie Sorry but I do not feel competent to judge those changes.

cocotb/drivers/__init__.py
```diff
@@ -262,10 +262,12 @@ def _wait_for_signal(self, signal):
         to move to :class:`~cocotb.triggers.NextTimeStep` before
         registering more callbacks can occur.
         """
-        yield ReadOnly()
-        while signal.value.integer != 1:
-            yield RisingEdge(signal)
-            yield ReadOnly()
+        warnings.warn(
+            "Use of BusDriver._wait_for_signal() is deprecated\n"
+            "\tPlease, use cocotb.trigggers.WaitCondition or event.aslongas() instead",
```
cmarqu:
```suggestion
            "\tPlease, use cocotb.triggers.WaitCondition or event.aslongas() instead",
```

cocotb/triggers.py
```diff
@@ -52,7 +52,36 @@
 class TriggerException(Exception):
     pass
 
-class Trigger(_py_compat.with_metaclass(abc.ABCMeta)):
+
+class TriggerHelper(object):
+    """A base class with commen methods for event type classes"""
```
```suggestion
    """A base class with common methods for event type classes."""
```

cocotb/triggers.py
```diff
@@ -52,7 +52,36 @@
...
+
+    @decorators.coroutine
+    def aslongas(self, condition, readonly=True):
```
cmarqu: ``while`` is typically used as the counterpart to ``until`` - any specific reason as to why are not using that?

cocotb/triggers.py
```diff
@@ -731,6 +765,58 @@ def _wait(self):
...
+    """
+    Wait until a given condition is True state.
+
+    Args:
+        condition (callable): Output of condition called without any parameters
+            is checked to be equivalent of logic True for condition to be met.
+        event (Trigger): event after which to check the condition; defaults to
+            :class:`NextTimeStep`.
+        readonly: whether to do check in readonly state; defaults to ``True``.
+        before (bool): wether to do the event once before doing first check;
```
```suggestion
        before (bool, optional): whether to do the event once before doing first check;
```

cocotb/triggers.py
```diff
@@ -731,6 +765,58 @@ def _wait(self):
...
 
 
+class WaitCondition(Waitable):
+    """
+    Wait until a given condition is True state.
+
+    Args:
+        condition (callable): Output of condition called without any parameters
+            is checked to be equivalent of logic True for condition to be met.
+        event (Trigger): event after which to check the condition; defaults to
```
```suggestion
        event (Trigger, optional): event after which to check the condition; defaults to
```

cocotb/triggers.py
```diff
@@ -731,6 +765,58 @@ def _wait(self):
...
+            :class:`NextTimeStep`.
+        readonly: whether to do check in readonly state; defaults to ``True``.
```
```suggestion
        readonly (bool, optional): whether to do check in ReadOnly state; defaults to ``True``.
```

cocotb/triggers.py
```diff
@@ -731,6 +765,58 @@ def _wait(self):
...
+        before (bool): wether to do the event once before doing first check;
+            defaults to ``False``.
+        after (bool): whether to wait for an extra event after condition
```
```suggestion
        after (bool, optional): whether to wait for an extra event after condition
```

cocotb/triggers.py
eric-wieser: +1, but needs to be `while_` to avoid conflicts. Underscore suffix is standard to avoid keyword clashes

cocotb/triggers.py
```diff
@@ -52,7 +52,36 @@
...
+
+        No extra ReadOnly will be done after the extra yield."""
+        yield WaitCondition(lambda: not condition(), event=self, readonly=readonly, before=False, after=True)
+
+    @decorators.coroutine
+    def until(self, condition, readonly=True):
+        """yield itself until condition is ``True``"""
+        yield WaitCondition(condition, event=self, readonly=readonly, before=True, after=False)
+
+    @decorators.coroutine
```
eric-wieser: You can remove this decorator, and replace the `yield` with `return` below, same for the others

cocotb/triggers.py
cmarqu: Unfortunately, the rest of cocotb uses a leading underscore...
Regardless of prefix or suffix, the "until" should maybe get the same treatment for consistency.

Fatsie: To me prefix underscore means things not to be used by user code so I will put it as suffix.

eric-wieser: Not sure what you're referring to @cmarqu: `_name` is for private fields, `name_` is for public fields which clash with keywords.

eric-wieser: I'm not sure we need `aslongas`/ `while` at all - all the use cases I can see about use `not` anyway, so would be clearer spelt with `until`.

Fatsie: The difference is that while does not wait for the event if condition is True and until does first yield the event and then check the condition; e.g. latter is modeled after `do ... until`.

cmarqu: > Not sure what you're referring to @cmarqu: `_name` is for private fields, `name_` is for public fields which clash with keywords.

I mean these: https://github.com/cocotb/cocotb/blob/a62759901c30db72b592da6c559dbe71efa5b518/cocotb/handle.py#L67-L70

Fatsie: I used `not` to keep with old code to interpret the `.value` as `boolean`; could also have used `== 0`. Actually did the former on first internal iteration.

eric-wieser: @cmarqu, that's a subtly different case, and _does_ match the built-in `namedtuple` behavior - so I don't think matters.

cocotb/triggers.py
```diff
@@ -127,6 +156,11 @@ def __await__(self):
         """))
 
 
+class Trigger(TriggerBase, TriggerHelper):
+    def __init__(self):
+        TriggerBase.__init__(self)
```
eric-wieser: No need to define `__init__` here

cocotb/triggers.py
eric-wieser: > I used `not` to keep with old code to interpret the `.value` as `boolean`; could also have used `== 0`. Actually did the former on first internal iteration.

Don't think there any need to make the lambda return a boolean. It's used eventually in a boolean context anyway.

Fatsie: New iteration of patch will not use ``while`` anymore but methods ``until``, ``until_after``, ``do_until`` and ``do_until_after``

cocotb/triggers.py
Fatsie: Actually I removed ``TriggerBase`` again and now use:
```python
class Trigger(_py_compat.abc_ABC, TriggerHelper)
```

cocotb/triggers.py
Fatsie: Done

Fatsie:
New iteration will all remarks handled plus some more docs of class and methods.

cocotb/triggers.py
```diff
@@ -731,6 +809,70 @@ def _wait(self):
...
+
+        .. code-block:: python
+
+            yield WaitCondition(lambda: bus.ack.integer, Edge(bus.ack))
+
+    .. versionadded: 1.3
+    """
+    def __init__(self, condition, **kwargs):
+        self.condition = condition
+        self.event = kwargs.pop('event', NextTimeStep())
```
eric-wieser: I think it's worth a type check here that `event` is either a `Trigger` or `Waitable`, this will behave poorly if a coroutine is passed.

cocotb/triggers.py
```diff
@@ -731,6 +809,70 @@ def _wait(self):
...
+        self.readonly = kwargs.pop('readonly', True)
+        self.before = kwargs.pop('before', False)
+        self.after = kwargs.pop('after', False)
+        reject_remaining_kwargs('WaitCondition.__init__', kwargs)
+
+
```
eric-wieser: A `__repr__` would go a long way here, perhaps:
```
def __repr__(self):
    return 'WaitCondition({s.condition!r}, event={s.event!r}, readonly={s.readonly!r}, before={s.before!r}, after={s.after!r})'.format(s=self)
```

imphil:
I'm not really opposed, but given the rather large design space for such changes, I'd like to find a way to add such changes without having to keep the forever. Maybe we can mark such features as experimental, possibly even with a opt-in mechanism, and if we are happy with it after a release or two remove the experimental marker.

cocotb/triggers.py
Fatsie: OK

cocotb/triggers.py
Fatsie: Will add it.

Fatsie:
Changes:
* marked as experimental. I use ``.. versionaddedexperimental::`` for this; let me know if there is better way. Making it optional would be somewhat involved.
* As suggested, added ``__repr__`` for ``WaitCondition`` and check if passed event is ``Trigger`` or ``Waitable`` object.

cmarqu:
Sphinx allows new directives/roles like ``versionaddedexperimental`` to be added with ``add_object_type`` (example at https://github.com/sphinx-doc/sphinx/blob/a4cce3e5613b4c721c19426fc120bbeeb830c3b2/doc/conf.py#L140-L142), but apparently then cocotb couldn't be used as a target for intersphinx: https://github.com/sphinx-doc/sphinx/pull/5598
This may not be a blocker though.

eric-wieser:
I think `..note` would be enough here for now, let's not open a Sphinx can of worms.

Will think about this more later.

Fatsie:
Put now in a note that WaitCondition and TriggerHelper are experimental. Should I add ``.. versionadded`` also ?

cocotb/drivers/avalon.py
```diff
@@ -149,9 +149,12 @@ def read(self, address, sync=True):
...
 
-        # Wait for waitrequest to be low
-        if hasattr(self.bus, "waitrequest"):
-            yield self._wait_for_nsignal(self.bus.waitrequest)
+        # Wait for waitrequest to be low if the bus has that signal
+        try:
+            handle = self.bus.waitrequest
+            yield FallingEdge(handle).until(lambda: not handle.value.integer)
+        except AttributeError:
+            pass
```
eric-wieser: Better as:
```python
        # Wait for waitrequest to be low if the bus has that signal
        try:
            handle = self.bus.waitrequest
        except AttributeError:
            pass
        else:
            yield FallingEdge(handle).until(lambda: not handle.value.integer)
```

cocotb/drivers/amba.py
```diff
@@ -351,22 +320,16 @@ def _read_data(self):
...
+            while burst_count > 0:
+                yield clock_re.do_until(lambda: self.bus.RREADY.value)
+                _burst_diff = burst_length - burst_count
+                _st = _araddr + (_burst_diff * bytes_in_beat)
+                _end = _araddr + ((_burst_diff + 1) * bytes_in_beat)
+                word.buff = self._memory[_st:_end].tostring()
+                self.bus.RDATA <= word
+                if burst_count == 1:
+                    self.bus.RLAST <= 1
                 burst_count -= 1
```
eric-wieser: This doesn't look right any more - `_burst_diff` is now always 1, whereas before it was not.

eric-wieser:
Here's another thought about alternate approaches - today we have something like:
```
            while True:
                yield ReadOnly()
                if int(self.bus.readdatavalid):
                    break	
                yield RisingEdge(self.clock)
```
What if we added a new `Read` trigger, which would let this be rewritten as:
```
            while not (yield Read(self.bus.readdatavalid)):
                yield RisingEdge(self.clock)
```
and implemented as:
```python
class Read(Waitable):
    def __init__(self, *signals):
        self.signals = signals
    @cocotb.coroutine
    def _wait(self):
        yield ReadOnly()
        vals = tuple(sig.value for sig in self.signals)
        if len(vals) == 1:
            vals, = vals  # unpack single-signals reads
        return vals
```

Fatsie:
> What if we added a new `Read` trigger, which would let this be rewritten as:
> 
> ```
>             while not (yield Read(self.bus.readdatavalid)):
>                 yield RisingEdge(self.clock)
> ```

I do think the the ``.until()`` syntax has more potential for optimization. If special condition types are introduced it should be able to do the full wait in the simulator without round trips to python. The new syntax does not allow that.

cocotb/drivers/amba.py
Fatsie: I don't see why ``_burst_diff`` is always 1; ``burst_count`` is decreased for every iteration of the loop so ``burst_diff`` increases.

cocotb/drivers/avalon.py
Fatsie: OK

Fatsie:
Use ``else:`` now for statements that should not be trapped by exception handler.

ktbarrett:
I'm not against a "wait until condition" convenience triggers, but I'm against it being depended upon #1249.

PR #1251, opened 2019-12-21
Add TypedBus.

grebe:
Currently, drivers and monitors list the signals that they expect to see. TypedBus moves this list of signals into the bus definition, e.g. AXI4Bus or AvalonMMBus. These objects can be shared between drivers and monitors so they share the same bus declaration.

Furthermore, this PR adds a mechanism to set the physical name of a logical port. The main use case for this arises from the fact that a standard may call a port one name (e.g. TVALID, TDATA, etc. in AXI) but the RTL may call the port a different name (e.g. valid, data, etc.). TypedBus provides a simple way to set the physical name for a logical port, for example:
```
  AvalonSTDriver(dut, "in", dut.clk, valid="A", ready="B", data="C")
```
will look for "in_A" for the valid signal, "in_B" for the ready signal, and "in_C" for the data signal.

This is based on discussion in #1186 and supercedes #1237.


cocotb/amba.py
```diff
@@ -0,0 +1,65 @@
...
+
+class AXI4LiteBus(TypedBus):
+    _signals = ["AWVALID", "AWADDR", "AWREADY",        # Write address channel
+                "WVALID", "WREADY", "WDATA", "WSTRB",  # Write data channel
+                "BVALID", "BREADY", "BRESP",           # Write response channel
+                "ARVALID", "ARADDR", "ARREADY",        # Read address channel
+                "RVALID", "RREADY", "RRESP", "RDATA"]  # Read data channel
+
+    def __init__(self, entity, name, **kwargs):
+        TypedBus.__init__(self, entity, name, **kwargs)
```
eric-wieser: These lines can be deleted, they do nothing - same throughout

grebe:
I think this is mostly what was suggested in the mentioned issue. The main question I'm uncertain about at this point is the mechanism by which names are overridden, e.g. init(..., `valid="OTHER_VALID_NAME")`. I'm a little worried that you might run into a situation where a parameter name clashes with a signal name. Do you see this as a valid objection? Is there a better way? The most direct way I can think of is passing in a name mapping dict, but then you need to decide what to do when there's a signal in the map that doesn't exist in the signal list.

ktbarrett:
The `Bus` sources are being moved to [cocotb-bus](github.com/cocotb/cocotb-bus). Please reopen this PR there if you wish. An easy way to do that is:

```bash
git checkout my-pr-branch
git fetch cocotb-bus
git merge cocotb-bus/master
```

PR #1255, opened 2019-12-22
Support non-decorated async coroutine in fork

### LINKED ISSUE #1246: Support forking non-decorated async coroutines
Support the following example of forking an non-decorated async coroutine.

```python
async def example():
    for i in range(10):
        await cocotb.triggers.Timer(10, 'ns')

cocotb.fork(example())
```

ktbarrett:
Resolves #1246. `scheduler.add` accepts non-decorated async coroutines.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -214,11 +214,6 @@ class RunningCoroutine(object):
 
     __bool__ = __nonzero__
 
-    def sort_name(self):
-        if self.stage is None:
-            return "%s.%s" % (self.module, self.funcname)
-        else:
-            return "%s.%d.%s" % (self.module, self.stage, self.funcname)
 
 class RunningTest(RunningCoroutine):
     """Add some useful Test functionality to a RunningCoroutine."""
@@ -280,6 +275,12 @@ class RunningTest(RunningCoroutine):
         aborting.
         """
         return self._force_outcome(outcomes.Error(exc))
+
+    def sort_name(self):
+        if self.stage is None:
+            return "%s.%s" % (self.module, self.funcname)
+        else:
+            return "%s.%d.%s" % (self.module, self.stage, self.funcname)
 
 
 class coroutine(object):
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -76,22 +76,22 @@ class CoroutineComplete(Exception):
         Exception.__init__(self, text)
 
 
-class RunningCoroutine(object):
-    """Per instance wrapper around an function to turn it into a coroutine.
+class RunningTask(object):
+    """Per instance wrapper around a running generator.
 
     Provides the following:
 
-        coro.join() creates a Trigger that will fire when this coroutine
+        task.join() creates a Trigger that will fire when this coroutine
         completes.
 
-        coro.kill() will destroy a coroutine instance (and cause any Join
+        task.kill() will destroy a coroutine instance (and cause any Join
         triggers to fire.
     """
-    def __init__(self, inst, parent):
+    def __init__(self, inst):
         if hasattr(inst, "__name__"):
             self.__name__ = "%s" % inst.__name__
 
-        if sys.version_info[:2] >= (3, 5) and inspect.iscoroutine(inst):
+        if inspect.iscoroutine(inst):
             self._natively_awaitable = True
             self._coro = inst.__await__()
         else:
@@ -99,10 +99,6 @@ class RunningCoroutine(object):
             self._coro = inst
         self._started = False
         self._callbacks = []
-        self._parent = parent
-        self.__doc__ = parent._func.__doc__
-        self.module = parent._func.__module__
-        self.funcname = parent._func.__name__
         self._outcome = None
 
         if not hasattr(self._coro, "send"):
@@ -215,6 +211,20 @@ class RunningCoroutine(object):
     __bool__ = __nonzero__
 
 
+class RunningCoroutine(RunningTask):
+    """
+    The result of calling a :any:`cocotb.coroutine` decorated coroutine.
+
+    All this class does is provide some extra attributes.
+    """
+    def __init__(self, inst, parent):
+        RunningTask.__init__(self, inst)
+        self._parent = parent
+        self.__doc__ = parent._func.__doc__
+        self.module = parent._func.__module__
+        self.funcname = parent._func.__name__
+
+
 class RunningTest(RunningCoroutine):
     """Add some useful Test functionality to a RunningCoroutine."""
 
@@ -263,7 +273,7 @@ class RunningTest(RunningCoroutine):
         self._outcome = outcome
         cocotb.scheduler.unschedule(self)
 
-    # like RunningCoroutine.kill(), but with a way to inject a failure
+    # like RunningTask.kill(), but with a way to inject a failure
     def abort(self, exc):
         """Force this test to end early, without executing any cleanup.
 
@@ -288,7 +298,7 @@ class coroutine(object):
 
     ``log`` methods will log to ``cocotb.coroutine.name``.
 
-    :meth:`~cocotb.decorators.RunningCoroutine.join` method returns an event which will fire when the coroutine exits.
+    :meth:`~cocotb.decorators.RunningTask.join` method returns an event which will fire when the coroutine exits.
 
     Used as ``@cocotb.coroutine``.
     """
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -88,24 +88,20 @@ class RunningTask(object):
         triggers to fire.
     """
     def __init__(self, inst):
-        if hasattr(inst, "__name__"):
-            self.__name__ = "%s" % inst.__name__
 
         if inspect.iscoroutine(inst):
             self._natively_awaitable = True
             self._coro = inst.__await__()
-        else:
+        elif inspect.isgenerator(inst):
             self._natively_awaitable = False
             self._coro = inst
+        else:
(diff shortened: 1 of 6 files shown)


eric-wieser:
While we're here, we could enable `yield async_func()` too

eric-wieser:
I think we need to keep around the old attributes on existing coroutines for compatibility, likely with a deprecation warning.

cocotb/decorators.py
```diff
@@ -87,38 +94,38 @@ class RunningCoroutine(object):
...
+
+    def __init__(self, inst, parent=None):
 
         if sys.version_info[:2] >= (3, 5) and inspect.iscoroutine(inst):
             self._natively_awaitable = True
             self._coro = inst.__await__()
-        else:
+            self._funcname = inst.__name__
+            self._module = modulename_from_filename(inst.cr_code.co_filename)
+            self.__doc__ = inst.__doc__
```
eric-wieser: To be clear here: I don't think we need to keep these properties when parent=None, only for existing users where parent is not none (and where extracting them is trivial)

cocotb/decorators.py
```diff
@@ -214,11 +221,23 @@ def __await__(self):
 
     __bool__ = __nonzero__
 
+    @property
+    def funcname(self):
+        warnings.warn("This was for internal use only and will be removed",
```
```suggestion
        warnings.warn("`RunningCoroutine.funcname` was for internal use only and will be removed",
```
Same for the others.

cocotb/decorators.py
```diff
@@ -234,14 +253,16 @@ def handle(self, record):
     def __init__(self, inst, parent):
         self.error_messages = []
         RunningCoroutine.__init__(self, inst, parent)
-        self.log = SimLog("cocotb.test.%s" % self.__name__, id(self))
+        self._log_name = "cocotb.test.%s" % inst.__name__
```
eric-wieser: A comment saying this overrides the base class would be helpful.

cocotb/decorators.py
```diff
@@ -360,6 +392,9 @@ def __get__(self, obj, type=None):
             and standalone functions"""
         return self.__class__(self._func.__get__(obj, type))
 
+    def __str__(self):
+        return str(self._func.__name__)
```
eric-wieser: If we need this at all, can you instead use
```
def __repr__(self):
    return "external({!r})".format(self._func)
```

cocotb/decorators.py
ktbarrett: And what will the behavior for `funcname` and `module` properties be? `AttributeError`? I'm not against it, just figured it would be considered a less optimal solution.

cocotb/decorators.py
ktbarrett: Honestly, the `__str__` methods are mostly useless. All of these classes and the regression manager have a massive and inconsistent public interface. If I can't break the public interface, I was hoping for consistency at a minimum. I'd prefer to remove it.

cocotb/decorators.py
```diff
@@ -214,11 +215,23 @@ def __await__(self):
...
+        return self._module
+
     def sort_name(self):
-        if self.stage is None:
-            return "%s.%s" % (self.module, self.funcname)
-        else:
-            return "%s.%d.%s" % (self.module, self.stage, self.funcname)
+        warnings.warn("`RunningCoroutine.sort_name` was for internal use only and will be removed",
+                      DeprecationWarning, stacklevel=2)
+        _ = self.stage  # will always throw as it previously did
```
eric-wieser: I think this one is safe to just remove, as you had before

cocotb/decorators.py
```diff
@@ -87,38 +90,36 @@ class RunningCoroutine(object):
...
-        if hasattr(inst, "__name__"):
-            self.__name__ = "%s" % inst.__name__
+
+    def __init__(self, inst, parent=None):
 
         if sys.version_info[:2] >= (3, 5) and inspect.iscoroutine(inst):
             self._natively_awaitable = True
             self._coro = inst.__await__()
-        else:
+            self.__name__ = inst.__name__
```
eric-wieser: You should continue to populate things as before if `parent` is not none.

eric-wieser:
I'm tempted to go further here, and emit deprecation warnings when `@cocotb.coroutine` is used on an async function.

eric-wieser:
Please also update the docs you can see here: https://cocotb.readthedocs.io/en/latest/coroutines.html#async-functions, which come from `coroutines.rst`.

cocotb/decorators.py
eric-wieser: Put an alternative approach in #1259, let me know what you think, and I'll close that PR if you argue against it or integrate the approach into this PR.

cocotb/decorators.py
```diff
@@ -281,13 +283,19 @@ def abort(self, exc):
         """
         return self._force_outcome(outcomes.Error(exc))
 
+    def sort_name(self):
+        if self.stage is None:
+            return "%s.%s" % (self.module, self.funcname)
+        else:
```
imphil: style: avoid else after return.

documentation/source/coroutines.rst
```diff
@@ -167,6 +167,13 @@ of :keyword:`yield`. Provided they are decorated with ``@cocotb.coroutine``,
 is determined by which type of function it appears in, not by the
 sub-coroutine being called.
 
+.. note::
+    The :any:`cocotb.coroutine` decorator is no longer necessary for ``async def`` functions. Such functions can be
+    used un-decorated wherever decorated coroutines are accepted, including :keyword:`yield` statements and
```
imphil:
```suggestion
    used without the ``@cocotb.coroutine`` decorator wherever decorated coroutines are accepted, including :keyword:`yield` statements and
```

imphil:
I didn't really follow the whole discussion here and in #1259, so I'd like a positive review from @eric-wieser before merging.

cocotb/decorators.py
```diff
@@ -214,11 +207,20 @@ def __await__(self):
...
-    def sort_name(self):
-        if self.stage is None:
-            return "%s.%s" % (self.module, self.funcname)
-        else:
-            return "%s.%d.%s" % (self.module, self.stage, self.funcname)
+
+class RunningCoroutine(RunningTask):
+    r"""
+    A compatibility wrapper for running :class:`coroutine`\ s.
+
```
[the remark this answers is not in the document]
ktbarrett: I don't know what the original docstring but, but this type is the result of calling a `cocotb.coroutine` coroutine. The class isn't useful, it just exists to support the legacy interface.

cocotb/decorators.py
eric-wieser: I'd disagree with this in this particular instance, since the if and else clauses are pretty symmetric. Not a big deal either way.

documentation/source/coroutines.rst
```diff
@@ -167,6 +167,13 @@ of :keyword:`yield`. Provided they are decorated with ``@cocotb.coroutine``,
 is determined by which type of function it appears in, not by the
 sub-coroutine being called.
 
+.. note::
```
eric-wieser:
```suggestion
.. versionchanged:: 1.4
```

documentation/source/coroutines.rst
```diff
@@ -167,6 +167,13 @@ of :keyword:`yield`. Provided they are decorated with ``@cocotb.coroutine``,
...
+    The :any:`cocotb.coroutine` decorator is no longer necessary for ``async def`` functions. Such functions can be
+    used un-decorated wherever decorated coroutines are accepted, including :keyword:`yield` statements and
+    :any:`cocotb.fork`.
+
+    .. versionadded:: 1.3
```
eric-wieser:
```suggestion
```

to make up for it being at the top of this section.

cocotb/decorators.py
```diff
@@ -214,11 +207,20 @@ def __await__(self):
...
+    All this class does is provide some extra attributes
```
```suggestion
    All this class does is provide some extra attributes.
```

tests/test_cases/test_cocotb/test_cocotb_35.py
```diff
@@ -116,3 +116,29 @@ def test_annotated_async_from_coro(dut):
...
+    ran = False
+
+    async def example():
+        nonlocal ran
+        await cocotb.triggers.Timer(1, 'ns')
+        ran = True
+
+    yield cocotb.fork(example()).join()
+    assert ran
+
```
eric-wieser:
```suggestion


```

cocotb/decorators.py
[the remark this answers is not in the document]
eric-wieser: To elaborate on the rationale - there are a bunch of attributes of `RunningCoroutine` which:
* aren't really useful in the first place
* are impossible to construct from `async def` functions

In order to keep existing code working, the compromise is to only produce these attributes when using `@coroutine`s, and not worry about them when we're not.

In future, we may want to deprecate these attributes - but we may also just want to deprecate `@coroutine` entirely.

cocotb/decorators.py
ktbarrett: I have to agree with Eric. In this case it's a statement that could be written as a ternary, but would otherwise be too long. Having separate branches that end with the last statement being similar encodes the same meaning. However, I know many linters (e.g. pylint) disagree.

cocotb/decorators.py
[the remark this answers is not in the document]
ktbarrett: I did reword this a bit to make it more clear about how it relates to `cocotb.coroutine`.

imphil:
Thanks!

Issue #1245, opened 2019-12-19
ReadWrite after ReadOnly


ktbarrett:
As discovered in #834, the test `test_readwrite_in_readonly` is expected to fail for "ModelSim" simulators, but when using Questa 10.5c and 10.7c the test passes. The test checks if the following fails:

```python
@cocotb.coroutine
def do_test_readwrite_in_readonly(dut):
    yield RisingEdge(dut.clk)
    yield ReadOnly()
    yield ReadWrite()
```

Why is this expected behavior? Why do the other simulators listed fail? Does Questa <10.5 or ModelSim actually fail here?

eric-wieser:
Relevant lines:

https://github.com/cocotb/cocotb/blob/6d4ce2593bb720f01a9fd6b1287bc99ab9d1957f/tests/test_cases/test_cocotb/test_cocotb.py#L348-L354

https://github.com/cocotb/cocotb/blob/6d4ce2593bb720f01a9fd6b1287bc99ab9d1957f/tests/test_cases/test_cocotb/test_cocotb.py#L375-L390

eric-wieser:
I'm not sure I understand why this is marked both as expected failure and expect error.

ktbarrett:
@eric-wieser I'm guessing it's because it causes some simulators to fail hard (`SimFailure`) and others to not re-enter (the coro will not continue to run `exited = True`) and will fail because of the time via `TestFailure`. The `expect_fail` case should also be predicated on the simulator, but all simulators apparently failed the test, so `True` was an easy catch-all.

eric-wieser:
I assume the expectation here is that the user must wait for `NextTimeStep` before waiting for `ReadWrite` again?

While obviously we should get the tests to pass, we should perhaps either make this fail on quests artificially, or we should make the other simulators work.

ktbarrett:
I'm in favor the former, we could also get all the failing simulators to error instead. Silently not working is bad.

ktbarrett:
GHDL and Verilator now also fail on this test.

Issue #834, opened 2019-03-06
Mutliple Test Failures for Questa and Riviera


leftink:
# Description
While checking that other fixes I am working on didn't break anything, I noticed several test failures when running regressions.  Both QuestaSIM and Riviera-Pro have failures.  See below for commands and result summaries.  I have attached the logs of the failures.

This issue may need to be split into multiple issues.

# Environment
* Linux 64-bit (Redhad 6)
* Anaconda 4.0.0 (Python 3.6)
* QuestaSIM (10.5c)
* Riviera-Pro (2018.02)
* COCOTB (8cb1e9e)

#  QuestaSIM Default
`make SIM=questa SIM_ARGS=-novopt`
## QuestaSIM Default Results
```
Failure in testsuite: 'all' classname: 'test_cocotb_array' testcase: 'test_in_2d_arr_unpacked' with parameters 'all'
Failure in testsuite: 'all' classname: 'test_cocotb_array' testcase: 'test_in_arr_packed_unpacked' with parameters 'all'
Failure in testsuite: 'all' classname: 'test_cocotb_array' testcase: 'test_in_vect_packed_packed_unpacked' with parameters 'all'
Ran a total of 1 TestSuites and 167 TestCases
```
Results: [questa_default_results.zip](https://github.com/potentialventures/cocotb/files/2938176/questa_results_1.zip)

#  QuestaSIM VHDL
`make SIM=questa SIM_ARGS=-novopt TOPLEVEL_LANG=vhdl`
## QuestaSIM VHDL Results
```
make[3]: *** [results.xml] Error 211
make[3]: Leaving directory `/scratch/z7023396/git_wa/rms-cdl-reuse/cocotb/examples/axi_lite_slave/tests'
make[2]: *** [sim] Error 2
make[2]: Target `all' not remade because of errors.
make[2]: Leaving directory `/scratch/z7023396/git_wa/rms-cdl-reuse/cocotb/examples/axi_lite_slave/tests'
make[1]: *** [axi_lite_slave/tests] Error 2
make[1]: Target `all' not remade because of errors.
make[1]: Leaving directory `/scratch/z7023396/git_wa/rms-cdl-reuse/cocotb/examples'
make: *** [do_tests] Error 2

--- FORCED TO MANUALLY RUN COMBINE_RESULTS ---
./bin/combine_results
Failure in testsuite: 'all' classname: 'issue_768' testcase: 'do_test' with parameters 'all'
[...]

PR #1256, opened 2019-12-22
Addresses multiple issues with Questa regression

ktbarrett:
Fixing some issues from #834. 

* test_plusargs fix 
* test_closedown test bug fix for Questa
* issue_957 test bug fix for all VHDL sims
* test_read_write test bug fix for all VHDL sims
* test_readwrite_in_readonly test bug fix for Questa (#1245)

The other issues will be fixed in their own PRs.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -625,7 +625,7 @@ class ModifiableObject(NonConstantObject):
             self._log.critical("Unsupported type for value assignment: %s (%s)", type(value), repr(value))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
-        simulator.set_signal_val_str(self._handle, value.binstr)
+        simulator.set_signal_val_binstr(self._handle, value.binstr)
 
     @NonConstantObject.value.getter
     def value(self):
--- a/cocotb/share/lib/fli/FliCbHdl.cpp
+++ b/cocotb/share/lib/fli/FliCbHdl.cpp
@@ -26,6 +26,10 @@
 ******************************************************************************/
 
 #include "FliImpl.h"
+#include "mti.h"
+#undef DLLEXPORT  // redefinition in tcl.h causes warning
+#include "tcl.h"
+#include <limits>
 
 /**
  * @name    cleanup callback
@@ -140,6 +144,55 @@ int FliStartupCbHdl::arm_callback()
     return 0;
 }
 
+
+static void get_args(gpi_sim_info_t *info)
+{
+    /*  Necessary to implement PLUSARGS
+        There is no function available on the FLI to obtain argc+argv directly from the
+        simulator. To work around this we use the TCL interpreter that ships with Questa,
+        some TCL commands, and the TCL variable `argv` to obtain the simulator argc/argv.
+    */
+    info->argc = 0;
+    info->argv = NULL;
+
+    Tcl_Interp* interp = reinterpret_cast<Tcl_Interp*>(mti_Interp());
+
+    // get argc
+    if (mti_Cmd("llength $argv") != TCL_OK) {
+        return;
+    }
+    unsigned long argc_u = strtoul(interp->result, NULL, 10);
+    Tcl_ResetResult(interp);
+    if (argc_u > std::numeric_limits<int>::max()) {
+        return;
+    }
+    int argc = static_cast<int>(argc_u);
+    if (argc_u == 0) {
+        return;
+    }
+
+    // allocate storage for argv
+    char** argv = new char*[argc];
+
+    // get each argv and copy into internal storage
+    for (int i = 0; i < argc; i++) {
+        std::string cmd = "lindex $argv " + std::to_string(i);
+        if (mti_Cmd(cmd.c_str()) != TCL_OK) {
+            return;
+        }
+        argv[i] = new char[strlen(interp->result)+1];
+        strcpy(argv[i], interp->result);
+        Tcl_ResetResult(interp);
+    }
+
+    // set only if all operations succeed
+    info->argc = argc;
+    info->argv = argv;
+
+    // argv storage leaks here
+}
+
+
 int FliStartupCbHdl::run_callback()
 {
     gpi_sim_info_t sim_info;
@@ -168,8 +221,7 @@ int FliStartupCbHdl::run_callback()
 
 
     // copy in sim_info.product
-    sim_info.argc = 0;
-    sim_info.argv = NULL;
+    get_args(&sim_info);
     sim_info.product = &product[0];
     sim_info.version = &version[0];
 
--- a/cocotb/share/lib/fli/FliObjHdl.cpp
+++ b/cocotb/share/lib/fli/FliObjHdl.cpp
@@ -147,7 +147,14 @@ int FliValueObjHdl::set_signal_value(long value)
     return -1;
 }
 
-int FliValueObjHdl::set_signal_value(std::string &value)
+int FliValueObjHdl::set_signal_value_binstr(std::string &value)
+{
+    COCOTB_UNUSED(value);
+    LOG_ERROR("Setting signal/variable value via string not supported for %s of type %d", m_fullname.c_str(), m_type);
+    return -1;
+}
+
+int FliValueObjHdl::set_signal_value_str(std::string &value)
 {
     COCOTB_UNUSED(value);
     LOG_ERROR("Setting signal/variable value via string not supported for %s of type %d", m_fullname.c_str(), m_type);
@@ -334,7 +341,7 @@ int FliLogicObjHdl::set_signal_value(const long value)
     return 0;
 }
 
-int FliLogicObjHdl::set_signal_value(std::string &value)
+int FliLogicObjHdl::set_signal_value_binstr(std::string &value)
 {
     if (m_fli_type == MTI_TYPE_ENUM) {
         mtiInt32T enumVal = m_enum_map[value.c_str()[0]];
@@ -506,7 +513,7 @@ const char* FliStringObjHdl::get_signal_value_str()
     return m_val_buff;
 }
 
-int FliStringObjHdl::set_signal_value(std::string &value)
+int FliStringObjHdl::set_signal_value_str(std::string &value)
 {
     strncpy(m_mti_buff, value.c_str(), static_cast<size_t>(m_num_elems));
 
--- a/cocotb/share/lib/simulator/simulatormodule.c
+++ b/cocotb/share/lib/simulator/simulatormodule.c
@@ -648,8 +648,7 @@ static PyObject *get_signal_val_long(PyObject *self, PyObject *args)
     return retval;
 }
 
-
-static PyObject *set_signal_val_str(PyObject *self, PyObject *args)
+static PyObject *set_signal_val_binstr(PyObject *self, PyObject *args)
 {
     COCOTB_UNUSED(self);
     gpi_sim_hdl hdl;
@@ -660,7 +659,24 @@ static PyObject *set_signal_val_str(PyObject *self, PyObject *args)
         return NULL;
     }
 
-    gpi_set_signal_value_str(hdl, binstr);
+    gpi_set_signal_value_binstr(hdl, binstr);
+    res = Py_BuildValue("s", "OK!");
+
+    return res;
+}
+
+static PyObject *set_signal_val_str(PyObject *self, PyObject *args)
+{
+    COCOTB_UNUSED(self);
+    gpi_sim_hdl hdl;
+    const char *str;
+    PyObject *res;
+
+    if (!PyArg_ParseTuple(args, "O&s", gpi_sim_hdl_converter, &hdl, &str)) {
+        return NULL;
+    }
+
+    gpi_set_signal_value_str(hdl, str);
     res = Py_BuildValue("s", "OK!");
 
     return res;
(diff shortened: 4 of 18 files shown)


tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -372,10 +372,9 @@ def do_test_afterdelay_in_readonly(dut, delay):
     exited = True
 
 
-@cocotb.test(expect_error=True,
+@cocotb.test(expect_error=(not cocotb.SIM_NAME.lower().startswith(("modelsim",))),
```
ktbarrett:
```suggestion
@cocotb.test(expect_error=(cocotb.result.SimFailure if cocotb.SIM_NAME.lower().startswith(("modelsim",)) else False),
```

tests/test_cases/test_closedown/test_closedown.py
```diff
@@ -34,7 +34,8 @@
 """
 
 import cocotb
-from cocotb.triggers import Timer, Join, RisingEdge
+from cocotb.triggers import Timer, RisingEdge
+from cocotb.result import SimFailure
```
ktbarrett: Remove.

ktbarrett:
After testing...

### Verilog
```
Failure in testsuite: 'all' classname: 'test_cocotb_array' testcase: 'test_in_2d_arr_unpacked' with parameters 'all'
Failure in testsuite: 'all' classname: 'test_cocotb_array' testcase: 'test_in_arr_packed_unpacked' with parameters 'all'
Failure in testsuite: 'all' classname: 'test_cocotb_array' testcase: 'test_in_vect_packed_packed_unpacked' with parameters 'all'
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'access_var_string_verilog' with parameters 'all'
Ran a total of 1 TestSuites and 150 TestCases
```

### VHDL
```
Failure in testsuite: 'all' classname: 'test_cocotb_array' testcase: 'test_in_2d_arr_unpacked' with parameters 'all'
Failure in testsuite: 'all' classname: 'test_cocotb_array' testcase: 'test_in_arr_packed_unpacked' with parameters 'all'
Failure in testsuite: 'all' classname: 'test_cocotb_array' testcase: 'test_in_vect_packed_packed_unpacked' with parameters 'all'
Failure in testsuite: 'all' classname: 'test_cocotb' testcase: 'test_readwrite_in_readonly' with parameters 'all'
Failure in testsuite: 'all' classname: 'plusargs' testcase: 'plusargs_test' with parameters 'all'
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'access_internal_register_array' with parameters 'all'
Ran a total of 1 TestSuites and 161 TestCases
```

Seems like kind of a crapshoot. `plusargs` still fails with the FLI, behavior of `test_readwrite_in_readonly` is inconsistent between VPI and FLI. 

I missed `access_internal_register_array`, but it's a buggy test, there is no object `register_array` in the module. This will be skipped.

tests/test_cases/test_cocotb/test_cocotb.py
ktbarrett: Perhaps this needs more thought. Failing is doing the wrong thing, should we adjust all the simulators to fail in this case?

tests/test_cases/issue_957/issue_957.py
```diff
@@ -5,7 +5,7 @@
 def wait_edge(dut):
     # this trigger never fires
     yield First(
-        RisingEdge(dut.and_output)
+        RisingEdge(dut.stream_out_ready)
```
cmarqu: While this fixes the error "``and_output`` not found in design" maybe we should align the VHDL with the Verilog version of the design. There is at least one other such problem in the tests (for a different signal).

tests/test_cases/test_array/test_array.py
```diff
@@ -175,7 +175,7 @@ def test_read_write(dut):
         _check_logic(tlog, dut.sig_t6[0][2][7], 0)
 
     if cocotb.LANGUAGE in ["vhdl"]:
-        _check_str(tlog, dut.port_str_out, "Testing")
+        _check_str(tlog, dut.port_str_out, "TEsting")
```
cmarqu: I discovered this independently, PR #1325.

cocotb/share/lib/fli/FliObjHdl.cpp
```diff
@@ -146,7 +146,13 @@ int FliValueObjHdl::set_signal_value(long value)
     return -1;
 }
 
-int FliValueObjHdl::set_signal_value(std::string &value)
+int FliValueObjHdl::set_signal_value_binstr(std::string &value)
+{
```
eric-wieser:
```suggestion
{
    COCOTB_UNUSED(value);
```

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -654,7 +653,24 @@ static PyObject *set_signal_val_str(PyObject *self, PyObject *args)
...
+}
+
+static PyObject *set_signal_val_str(PyObject *self, PyObject *args)
+{
+    COCOTB_UNUSED(self);
+    gpi_sim_hdl hdl;
+    const char *str;
+    PyObject *res;
+
+    if (!PyArg_ParseTuple(args, "O&s", gpi_sim_hdl_converter, &hdl, &str)) {
```
eric-wieser: This null-terminates strings - are you sure you want that? Otherwise, you want `y#` here I think.

cocotb/share/lib/simulator/simulatormodule.h
```diff
@@ -98,7 +99,8 @@ static PyMethodDef SimulatorMethods[] = {
     {"get_signal_val_binstr", get_signal_val_binstr, METH_VARARGS, "Get the value of a signal as a binary string"},
     {"get_signal_val_real", get_signal_val_real, METH_VARARGS, "Get the value of a signal as a double precision float"},
     {"set_signal_val_long", set_signal_val_long, METH_VARARGS, "Set the value of a signal using a long"},
-    {"set_signal_val_str", set_signal_val_str, METH_VARARGS, "Set the value of a signal using a binary string"},
+    {"set_signal_val_str", set_signal_val_str, METH_VARARGS, "Set the value of a signal using an ASCII string"},
+    {"set_signal_val_binstr", set_signal_val_binstr, METH_VARARGS, "Set the value of a signal using a binary string"},
```
eric-wieser:
```suggestion
    {"set_signal_val_binstr", set_signal_val_binstr, METH_VARARGS, "Set the value of a signal using a string with a character per bit"},
```

cocotb/share/lib/simulator/simulatormodule.h
```diff
@@ -98,7 +99,8 @@ static PyMethodDef SimulatorMethods[] = {
     {"get_signal_val_binstr", get_signal_val_binstr, METH_VARARGS, "Get the value of a signal as a binary string"},
     {"get_signal_val_real", get_signal_val_real, METH_VARARGS, "Get the value of a signal as a double precision float"},
     {"set_signal_val_long", set_signal_val_long, METH_VARARGS, "Set the value of a signal using a long"},
-    {"set_signal_val_str", set_signal_val_str, METH_VARARGS, "Set the value of a signal using a binary string"},
+    {"set_signal_val_str", set_signal_val_str, METH_VARARGS, "Set the value of a signal using an ASCII string"},
```
eric-wieser:
```suggestion
    {"set_signal_val_str", set_signal_val_str, METH_VARARGS, "Set the value of a signal using an 8-bit string"},
```

cocotb/share/lib/simulator/simulatormodule.c
eric-wieser: Edit: Looks like you do want embedded nulls - but `y` would be more appropriate, since HDL doesn't know about unicode.

ktbarrett:
Currently the failures in Questa:

* plusargs (only with FLI)
  * this is currently unimplemented, though I see a way forward. 
* multi-dimensional arrays
  * need more development of leftink/cocotb/issue_813
* readwrite_in_readonly (only with FLI)
  * inconsistent behavior in this case, passes time with FLI and fails the simulator (as expected) with VPI

cocotb/share/lib/simulator/simulatormodule.c
ktbarrett: I made this a separate commit for you to ok.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -666,7 +666,7 @@ static PyObject *set_signal_val_str(PyObject *self, PyObject *args)
     const char *str;
     PyObject *res;
 
-    if (!PyArg_ParseTuple(args, "O&s", gpi_sim_hdl_converter, &hdl, &str)) {
+    if (!PyArg_ParseTuple(args, "O&y", gpi_sim_hdl_converter, &hdl, &str)) {
```
ktbarrett: This is wrong, should use `s` instead of `y` here.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -642,19 +642,35 @@ static PyObject *get_signal_val_long(PyObject *self, PyObject *args)
...
 }
 
+static PyObject *set_signal_val_binstr(PyObject *self, PyObject *args)
+{
+    COCOTB_UNUSED(self);
+    gpi_sim_hdl hdl;
+    const char *binstr;
+    PyObject *res;
+
+    if (!PyArg_ParseTuple(args, "O&y", gpi_sim_hdl_converter, &hdl, &binstr)) {
```
eric-wieser: Strictly this should be accepting unicode strings, but it's probably easier to `encode` from the python side than it is to encode them here.

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -26,6 +26,9 @@
 ******************************************************************************/
 
 #include "FliImpl.h"
+#include "mti.h"
+#include "tcl.h"
```
ktbarrett: tcl.h redefines `DLLEXPORT` which causes a warning, `undef` it first.

cocotb/handle.py
```diff
@@ -625,7 +625,7 @@ def setimmediatevalue(self, value):
             self._log.critical("Unsupported type for value assignment: %s (%s)", type(value), repr(value))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
-        simulator.set_signal_val_str(self._handle, value.binstr)
+        simulator.set_signal_val_binstr(self._handle, value.binstr.encode())
```
ktbarrett: There is a call to `simulator.set_signal_val_str` below that also needs the string to bytes conversion.

eric-wieser: I'd argue that for `set_signal_val_str` by design, and that the user should perform the conversion. As far as I can tell, systemverilog has no concept of encoding, and just copies across raw bytes.

ktbarrett: It's more of a convenience thing. People expect sending `"TEST"` off to just work. We could add a check in here to try `value.encode('ascii')` if given a `str` object, pass through bytes, or complain otherwise.

eric-wieser: I understand where you're coming from, but python 3 aimed to draw the line at that type of behavior.

eric-wieser: And passing `b"test"` isn't much harder

ktbarrett: It will break existing correct sims. Also, we need to reconsider what gets return by `get_signal_value_str` and `get_signal_value_binstr`. This is more of a change than a fix, and all the tests that I intend to fix will still work despite this change. So, I will remove this commit leaving it parsed with `s` and open a separate PR.

eric-wieser: I don't necessarily agree that such sims are _correct_, but if they're already executing I'm with you that we don't want to change the behavior in this PR.

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -140,6 +144,53 @@ int FliStartupCbHdl::arm_callback()
...
+static void get_args(gpi_sim_info_t *info)
+{
+    info->argc = 0;
+    info->argv = NULL;
+
+    Tcl_Interp* interp = reinterpret_cast<Tcl_Interp*>(mti_Interp());
+    LOG_DEBUG("Starting to parse argc and argv");
+
+    // get argc
+    if (mti_Cmd("llength $argv") != TCL_OK)
```
imphil: Please always use `{`, even for single-line statements.

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -140,6 +144,53 @@ int FliStartupCbHdl::arm_callback()
     return 0;
 }
 
+
+static void get_args(gpi_sim_info_t *info)
```
imphil: Please add documentation to this function. It's a sufficiently obscure way to pass plusargs to a simulator to warrant a couple lines of docs.

tests/test_cases/issue_957/issue_957.py
cmarqu: Done, #1396.

ktbarrett:
@imphil I also removed unnecessary `LOG_DEBUG` statements from the function.

cocotb/share/lib/vhpi/VhpiImpl.h
```diff
@@ -189,7 +189,8 @@ class VhpiSignalObjHdl : public GpiSignalObjHdl {
     using GpiSignalObjHdl::set_signal_value;
     int set_signal_value(long value) override;
     int set_signal_value(double value) override;
-    int set_signal_value(std::string &value) override;
+    int set_signal_value_binst(std::string &value) override;
```
eric-wieser: Typo here, cc @themperek

PR #1259, opened 2019-12-23
Alternative to 1255

eric-wieser:
@ktbarrett: Created this because it was easier that writing a comment on your PR.

This is #1255 without the deprecations. Last two commits are @ktbarrett's.

Instead of deprecating some of the members of `RunningCoroutine`, this extracts a base class `RunningTask` which does not contain the offending members.

That way, new code can construct the base class for simplicity, and old code will continue to get the attributes it expects.

Instead of deprecating the attributes of `RunningCoroutine`, I think we should just (slowly) push users away from using `cocotb.coroutine` at all, which will naturally phase out all uses of `RunningCoroutine`.

The name `RunningTask` is inspired by `asyncio.Task`.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -76,18 +76,18 @@ class CoroutineComplete(Exception):
         Exception.__init__(self, text)
 
 
-class RunningCoroutine(object):
-    """Per instance wrapper around an function to turn it into a coroutine.
+class RunningTask(object):
+    """Per instance wrapper around a running generator.
 
     Provides the following:
 
-        coro.join() creates a Trigger that will fire when this coroutine
+        task.join() creates a Trigger that will fire when this coroutine
         completes.
 
-        coro.kill() will destroy a coroutine instance (and cause any Join
+        task.kill() will destroy a coroutine instance (and cause any Join
         triggers to fire.
     """
-    def __init__(self, inst, parent):
+    def __init__(self, inst):
         if hasattr(inst, "__name__"):
             self.__name__ = "%s" % inst.__name__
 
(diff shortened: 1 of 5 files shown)


cocotb/decorators.py
```diff
@@ -76,18 +76,18 @@ def __init__(self, text=""):
...
         completes.
 
-        coro.kill() will destroy a coroutine instance (and cause any Join
+        task.kill() will destroy a coroutine instance (and cause any Join
         triggers to fire.
     """
-    def __init__(self, inst, parent):
+    def __init__(self, inst):
         if hasattr(inst, "__name__"):
             self.__name__ = "%s" % inst.__name__
```
ktbarrett: This does not need to be conditioned, both `cocotb.decorators.coroutine` and async coroutines have `__name__`s. `__str__` will fail if this isn't set. And I would move this after the type check to get better error messages in error cases.

cocotb/decorators.py
```diff
@@ -99,10 +99,6 @@ def __init__(self, inst, parent):
             self._coro = inst
         self._started = False
         self._callbacks = []
-        self._parent = parent
-        self.__doc__ = parent._func.__doc__
-        self.module = parent._func.__module__
-        self.funcname = parent._func.__name__
         self._outcome = None
 
         if not hasattr(self._coro, "send"):
```
ktbarrett: Like in my PR this should be moved to earlier in the method and check `inspect.isgenerator` instead of checking for only `send` for non-async-coroutine cases.

ktbarrett:
Feel free to cherry-pick the documentation change from my PR if you like it.

ktbarrett:
@eric-wieser I took your commits, addied in some documentation, and fixed a few bugs in my PR #1255.

eric-wieser:
Sorry for not keeping up with this. Thanks for pulling in the idea into your PR.

PR #1260, opened 2019-12-23
Stop using the deprecated `colour` attribute.

eric-wieser:
This prevents a `DeprecationWarning` that was introduced by #1231.

Found while trying out #1258


ktbarrett:
Isn't this in #1258?

eric-wieser:
Yes, it now is.

eric-wieser:
Merged as part of the linked pr

PR #1262, opened 2019-12-24
Do not ignore failures during test initialization

### LINKED ISSUE #1253: Tests that fail before the scheduler are not handled properly
For example:
https://github.com/cocotb/cocotb/blob/92a032a4349b2fe9e2f9bff5d1060ef44cd4cd72/tests/test_cases/test_cocotb/test_cocotb.py#L85-L88

This is considered "N/A" in the regression results banner and passes whether `expect_fail` or `expect_error` are set to `True` or `False`. A traceback is printed, but it's not apparent there was any failure if you are running a number of tests, like in `test_cocotb`.

The regression manager nor the `cocotb.test` decorator make any attempt to handle potential issues.
https://github.com/cocotb/cocotb/blob/92a032a4349b2fe9e2f9bff5d1060ef44cd4cd72/cocotb/regression.py#L157
https://github.com/cocotb/cocotb/blob/92a032a4349b2fe9e2f9bff5d1060ef44cd4cd72/cocotb/decorators.py#L477-L478

I put some time into refactoring `initialize` and `handle_result` in the regression manager, but felt like I more or less had to copy-paste to victory, which I didn't like. I think a larger refactor is in order.

ktbarrett:
Fixes #1253. PEP8 fixes and comments.

There are two places in the RM that initialize coroutines, each had slightly different code, so this was factored out into a method `init_test`.

`init_test` can fail, and so can a test at any point in it's run (previous handled by `handle_result`), so scoring a test as a pass or a fail depending upon the result and expected behavior is factored out into a method `score_test`.

`score_test` can print out tracebacks on certain failures, and non-pertinent traceback frames are stripped. However because there are two entry points to `score_test`, the exception handling functionality had to be factored out into a static method `exception_info`.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -108,6 +108,9 @@ class RegressionManager(object):
         self.skipped = 0
         self.failures = 0
 
+        # Setup XUnit
+        ###################
+
         results_filename = os.getenv('COCOTB_RESULTS_FILE', "results.xml")
         suite_name = os.getenv('RESULT_TESTSUITE', "all")
         package_name = os.getenv('RESULT_TESTPACKAGE', "all")
@@ -118,12 +121,18 @@ class RegressionManager(object):
                                  package=package_name)
 
         if (self._seed is not None):
-            self.xunit.add_property(name="random_seed", value=("%d"%self._seed))
+            self.xunit.add_property(name="random_seed", value=("%d" % self._seed))
+
+        # Setup Coverage
+        ####################
 
         if coverage is not None:
             self.log.info("Enabling coverage collection of Python code")
             self._cov = coverage.coverage(branch=True, omit=["*cocotb*"])
             self._cov.start()
+
+        # Setup DUT object
+        #######################
 
         handle = simulator.get_root_handle(self._root_name)
 
@@ -133,7 +142,9 @@ class RegressionManager(object):
             raise AttributeError("Can not find Root Handle (%s)" %
                                  self._root_name)
 
-        # Auto discovery
+        # Test Discovery
+        ####################
+
         for module_name in self._modules:
             try:
                 self.log.debug("Python Path: " + ",".join(sys.path))
@@ -148,7 +159,7 @@ class RegressionManager(object):
 
             if self._functions:
 
-                # Specific functions specified, don't auto discover
+                # Specific functions specified, don't auto-discover
                 for test in self._functions.rsplit(','):
                     try:
                         _test = getattr(module, test)
@@ -158,35 +169,18 @@ class RegressionManager(object):
                         _py_compat.raise_from(err, None)  # discard nested traceback
 
                     if not hasattr(_test, "im_test"):
-                        self.log.error("Requested %s from module %s isn't a cocotb.test decorated coroutine", test, module_name)
+                        self.log.error("Requested %s from module %s isn't a cocotb.test decorated coroutine",
+                                       test, module_name)
                         raise ImportError("Failed to find requested test %s" % test)
-                    self._queue.append(_test(self._dut))
-                    self.ntests += 1
+                    self._init_test(_test)
+
+                # only look in first module for all functions and don't complain if all functions are not found
                 break
 
+            # auto-discover
             for thing in vars(module).values():
                 if hasattr(thing, "im_test"):
-                    try:
-                        test = thing(self._dut)
-                        skip = test.skip
-                    except Exception:
-                        skip = True
-                        self.log.warning("Failed to initialize test %s" %
-                                         thing.name, exc_info=True)
-
-                    if skip:
-                        self.log.info("Skipping test %s" % thing.name)
-                        self.xunit.add_testcase(name=thing.name,
-                                                classname=module_name,
-                                                time="0.0",
-                                                sim_time_ns="0.0",
-                                                ratio_time="0.0")
-                        self.xunit.add_skipped()
-                        self.skipped += 1
-                        self._store_test_result(module_name, thing.name, None, 0.0, 0.0, 0.0)
-                    else:
-                        self._queue.append(test)
-                        self.ntests += 1
+                    self._init_test(thing)
 
         self._queue.sort(key=lambda test: test.sort_name())
 
@@ -195,8 +189,11 @@ class RegressionManager(object):
                           (valid_tests.module,
                            valid_tests.funcname))
 
+        # Process Hooks
+        ###################
+
         for module_name in self._hooks:
-            self.log.info("Loading hook from module '"+module_name+"'")
+            self.log.info("Loading hook from module '" + module_name + "'")
             module = _my_import(module_name)
 
             for thing in vars(module).values():
@@ -208,25 +205,29 @@ class RegressionManager(object):
                     else:
                         cocotb.scheduler.add(test)
 
-
     def tear_down(self):
-        """It's the end of the world as we know it"""
+
+        # Write out final log messages
         if self.failures:
             self.log.error("Failed %d out of %d tests (%d skipped)" %
                            (self.failures, self.count - 1, self.skipped))
         else:
             self.log.info("Passed %d tests (%d skipped)" %
                           (self.count - 1, self.skipped))
+        if len(self.test_results) > 0:
+            self._log_test_summary()
+        self._log_sim_summary()
+        self.log.info("Shutting down...")
+
+        # Generate output reports
+        self.xunit.write()
         if self._cov:
             self._cov.stop()
             self.log.info("Writing coverage data")
             self._cov.save()
             self._cov.html_report()
-        if len(self.test_results) > 0:
-            self._log_test_summary()
-        self._log_sim_summary()
-        self.log.info("Shutting down...")
-        self.xunit.write()
+
+        # Setup simulator finalization
         simulator.stop_simulator()
 
     def next_test(self):
@@ -235,25 +236,19 @@ class RegressionManager(object):
             return None
         return self._queue.pop(0)
 
-    def _add_failure(self, result):
-        self.xunit.add_failure(stdout=repr(str(result)),
-                               stderr="\n".join(self._running_test.error_messages),
-                               message="Test failed with random_seed={}".format(self._seed))
-        self.failures += 1
-
     def handle_result(self, test):
         """Handle a test completing.
 
-        Dump result to XML and schedule the next test (if any).
+        Dump result to XML and schedule the next test (if any). Entered by the scheduler.
 
         Args:
             test: The test that completed
         """
         assert test is self._running_test
 
-        real_time   = time.time() - test.start_time
+        real_time = time.time() - test.start_time
         sim_time_ns = get_sim_time('ns') - test.start_sim_time
-        ratio_time  = self._safe_divide(sim_time_ns, real_time)
+        ratio_time = self._safe_divide(sim_time_ns, real_time)
 
         self.xunit.add_testcase(name=test.funcname,
                                 classname=test.module,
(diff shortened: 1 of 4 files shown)


cocotb/regression.py
```diff
@@ -152,35 +163,18 @@ def initialise(self):
...
-                                                time="0.0",
-                                                sim_time_ns="0.0",
-                                                ratio_time="0.0")
-                        self.xunit.add_skipped()
-                        self.skipped += 1
-                        self._store_test_result(module_name, thing.name, None, 0.0, 0.0, 0.0)
-                    else:
-                        self._queue.append(test)
-                        self.ntests += 1
+                    self.init_test(thing)
```
eric-wieser: Let's make this `_init_test` to make it super clear that it's only to be used in this file.

cocotb/regression.py
```diff
@@ -189,38 +183,45 @@ def initialise(self):
...
+            self.log.info("Loading hook from module '" + module_name + "'")
             module = _my_import(module_name)
 
             for thing in vars(module).values():
                 if hasattr(thing, "im_hook"):
                     try:
                         test = thing(self._dut)
                     except Exception:
-                        self.log.warning("Failed to initialize hook %s" % thing.name, exc_info=True)
+                        self.log.error("Failed to initialize hook %s" % thing.name, exc_info=True)
```
eric-wieser: I'd leave this as is. Hooks aren't really useful, and an `error` here would suggest propagation in a way that doesn't happen - a warning as was here before was more indicative of the fact that things continue regardless.

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -178,11 +178,6 @@ def clock_yield(generator):
     test_flag = True
 
 
-@cocotb.test(expect_fail=True)
-def test_duplicate_yield(dut):
-    """A trigger can not be yielded on twice"""
```
eric-wieser: Note for myself: would be good to check when this was introduced in order to guess what was intended to be tested here.

eric-wieser:
Will come back to reviewing this in a few days

tests/test_cases/test_cocotb/test_cocotb.py
ktbarrett: Looks like it was just commited unfinished 13128df6bcb6062b12b660ad67b50b2e36df8680.

ktbarrett: I think it means yielding on a list of triggers where the triggers are not unique.

eric-wieser: Ah, that could be it. Would be interesting to see what happens in that case, but obviously out of scope here.

cocotb/regression.py
```diff
@@ -229,63 +230,112 @@ def next_test(self):
...
+            self.failures += 1
+
+        # Fail if required
+        if sim_failed:
+            self.tear_down()
+
+        self.execute()
+
+    @staticmethod
+    def exception_info(e, frame_names):
```
eric-wieser: Instead of taking e and frame_names, could we pass in an outcome here? It would make the interface a little simpler, and remove the restriction that this method can only be called in an except block.

ktbarrett: True. I removed the function and changed `score_test` to take an outcome. I noticed a potential improvement on `Outcome` would be to add a `__bool__` method for checking whether the outcome is an `Error` or a `Value`. Seeing as `Outcome` is basically an option monad, its natural for people to use it that way.

ktbarrett: I removed this function and changed `score_test` to take an outcome.

cocotb/regression.py
```diff
@@ -229,63 +230,111 @@ def next_test(self):
...
+                                        sim_time_ns="0.0",
+                                        ratio_time="0.0")
+                self.xunit.add_skipped()
+                self.skipped += 1
+                self._store_test_result(test.module, test_func.name, None, 0.0, 0.0, 0.0)
+            else:
+                self._queue.append(test)
+                self.ntests += 1
+
+    def score_test(self, test, outcome):
```
eric-wieser: Again, this should be private I think:
```suggestion
    def _score_test(self, test, outcome):
        """ score the provided outcome into a ``(did_pass, did_stop_simulation)`` pair """
```
Some vague docstring would be handy, if only to encourage more docs here if things change.

cocotb/regression.py
```diff
@@ -229,63 +230,111 @@ def next_test(self):
...
+        sim_failed = False
 
-        # check what exception the test threw
         try:
-            test._outcome.get()
+            outcome.get()
         except Exception as e:
             if sys.version_info >= (3, 5):
-                result = remove_traceback_frames(e, ['handle_result', 'get'])
+                result = remove_traceback_frames(e, ['score_test', 'get'])
```
eric-wieser: Don't forget to change these with the function name
```suggestion
                result = remove_traceback_frames(e, ['_score_test', 'get'])
```

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -82,7 +82,7 @@ def assert_raises(exc_type):
 # Tests relating to providing meaningful errors if we forget to use the
 # yield keyword correctly to turn a function into a coroutine
 
-@cocotb.test(expect_fail=True)
+@cocotb.test(expect_error=True)
```
eric-wieser: Mind using an explicit exception type here now that you added support for it?

cocotb/regression.py
```diff
@@ -229,63 +230,111 @@ def next_test(self):
...
+            self.xunit.add_failure()
+            self.failures += 1
+
+        # Fail if required
+        if sim_failed:
+            self.tear_down()
+
+        self.execute()
+
+    def _init_test(self, test_func):
```
eric-wieser: Please add a brief docstring

cocotb/regression.py
```diff
@@ -229,63 +230,111 @@ def next_test(self):
...
+                                    ratio_time="0.0")
+            result_pass, sim_failed = self.score_test(test_func, test_init_outcome.without_frames(['capture']))
+            # Save results
+            self._store_test_result(test_func.__module__, test_func.__name__, result_pass, 0.0, 0.0, 0.0)
+            if not result_pass:
+                self.xunit.add_failure()
+                self.failures += 1
+            # Fail if required
+            if sim_failed:
+                self.tear_down()
```
eric-wieser: This won't actually stop subsequent tests being run, will it? I think that's the intent of `tear_down`.

ktbarrett: `sim_failed` is only true if `SimFailure` was the result and `tear_down` causes cocotb to stop. I'm not sure how possible that is for initialization, but it's a potential result so it's covered.

eric-wieser: Not sure I got my point across very well there.

As written, if test1 causes the simulator to die (which as you correctly remark is unlikely in the first place), then the regression manager will still try to start all the remaining tests - stopping the simulator isn't going to immediately stop the regression manager setup code.

Perhaps since this is so unlikely anyway, you should just raise `RuntimeError` here after calling `tear_down`, since if the simulator really did die in setup something has gone very wrong.

cocotb/regression.py
```diff
@@ -229,63 +230,119 @@ def next_test(self):
...
+                                        ratio_time="0.0")
+                self.xunit.add_skipped()
+                self.skipped += 1
+                self._store_test_result(test.module, test_func.name, None, 0.0, 0.0, 0.0)
+            else:
+                self._queue.append(test)
+                self.ntests += 1
+
+    def _score_test(self, test, outcome):
+        """ Given a test and the test's outcome, determine if the test met expectations """
```
eric-wieser: Please include the fact that this function is responsible for logging the result too.

cocotb/regression.py
ktbarrett: `tear_down` calls `simulator.stop_simulator` which calls `gpi_sim_end`, so the regression manager code stops there immediately. `tear_down` is basically an exit function, it can't be run more than once.

ktbarrett: I added an exception here anyway with a message indicating that it should not be possible.

cocotb/regression.py
```diff
@@ -229,63 +230,122 @@ def next_test(self):
...
+
+        # Save results
+        self._store_test_result(test.__module__, test.__name__, result_pass, sim_time_ns, real_time, ratio_time)
+        if not result_pass:
+            self.xunit.add_failure()
+            self.failures += 1
+
+        # Fail if required
+        if sim_failed:
+            self.tear_down()
```
eric-wieser:
```suggestion
            self.tear_down()
            return
```
which matches what was here before

cocotb/regression.py
```diff
@@ -229,63 +230,122 @@ def next_test(self):
...
+
+        self.execute()
+
+    def _init_test(self, test_func):
+        """
+        Initializes a test.
+
+        Records outcome if the initialization fails.
+        Records skip if the test is skipped.
+        Saves the initialized test if it sucessfully initializes.
```
eric-wieser:
```suggestion
        Saves the initialized test if it successfully initializes.
```

cocotb/regression.py
ktbarrett: Throwing is more appropriate here, just as it was below, `tear_down` is an exit function. The throw should probably be moved into `tear_down` to right after the call to `stop_simulator`. And I will add a docstring to `tear_down` to indicate that it does not return.

eric-wieser: That sounds reasonable, thanks

cocotb/regression.py
```diff
@@ -229,63 +230,123 @@ def next_test(self):
...
+            result_pass, sim_failed = self._score_test(test_func, test_init_outcome.without_frames(['capture']))
+            # Save results
+            self._store_test_result(test_func.__module__, test_func.__name__, result_pass, 0.0, 0.0, 0.0)
+            if not result_pass:
+                self.xunit.add_failure()
+                self.failures += 1
+            # Fail if required
+            if sim_failed:
+                self.tear_down()
+                raise SimFailure
```
eric-wieser: This line is now trivially unreachable, right?

cocotb/regression.py
```diff
@@ -310,27 +369,21 @@ def _result_was():
             else:
                 self.log.error("Test error has lead to simulator shutting us "
                                "down", exc_info=exc_info)
-                self._add_failure(result)
-                self._store_test_result(test.module, test.funcname, False, sim_time_ns, real_time, ratio_time)
-                self.tear_down()
-                return
+                result_pass = False
+            sim_failed = True
```
eric-wieser:
```suggestion
            # whether we expected it or not, the simulation has failed unrecoverably
            sim_failed = True
```

tests/test_cases/issue_1279/issue_1279.py
```diff
@@ -0,0 +1,13 @@
...
+
+@cocotb.test(expect_error=cocotb.result.SimFailure)
+async def test_sim_failure_a(dut):
+    # invoke a deadlock
+    await cocotb.triggers.RisingEdge(dut.clk)
+
+
+@cocotb.test()
+async def test_sim_failure_b(dut):
+    raise cocotb.result.TestFailure("This test should never run")
```
eric-wieser: Unless your plan is to wait until after we drop python 2, you can't use `async` here yet.

tests/test_cases/issue_1279/issue_1279.py
```diff
@@ -0,0 +1,13 @@
+import cocotb
+
+
```
eric-wieser:
```suggestion
# note: this file relies on the fact the simulator runs tests sorted by name
```

tests/test_cases/issue_1279/issue_1279.py
```diff
@@ -0,0 +1,13 @@
+import cocotb
```
eric-wieser:
```suggestion
"""
Test that once a SimFailure occurs, no further tests are run
"""
import cocotb
```

cocotb/regression.py
```diff
@@ -310,27 +369,21 @@ def _result_was():
...
 
         elif test.expect_error:
             if isinstance(result, test.expect_error):
                 self.log.info("Test errored as expected: " + _result_was())
             else:
-                self.log.info("Test errored with unexpected type: " + _result_was())
-                self._add_failure(result)
+                self.log.error("Test errored with unexpected type", exc_info=exc_info)
```
eric-wieser: Any reason for dropping `_result_was()` here?

cocotb/regression.py
ktbarrett: Apparently it is. This line will throw the exception to be caught by `gpi_embed`. I'll add a message.

cocotb/regression.py
ktbarrett: I think it's kind of redundant if you are printing the stacktrace, but I will add it back in, it might be used.

tests/test_cases/issue_1279/issue_1279.py
ktbarrett: I instead used the test decorator's `stage`.

cocotb/regression.py
eric-wieser: So `tear_down` does not stop the python interpreter after all?

imphil:
It's a pretty large change at this point, let's not take this for 1.3.

ktbarrett:
@eric-wieser If this is being held back. I wonder if this could/should be brought in on this PR as well. https://github.com/ktbarrett/cocotb/pull/3

eric-wieser:
Needs a rebase - conflicts are with the addition of `pdb` on failure - up to you whether you think that belongs in `score_test` or elsewhere.

ktbarrett:
@eric-wieser rebased and ready.

ktbarrett:
Had to move the recent changes from #1268. They are now in the conceptually same spot right before `_store_test_result` in `_handle_result` which is only entered if the test completed and did not fail during initialization.

cocotb/regression.py
```diff
@@ -316,36 +378,25 @@ def _result_was():
...
+                self.log.error("Test errored with unexpected type: " + _result_was(), exc_info=exc_info)
                 result_pass = False
 
         else:
             self.log.error("Test Failed: " + _result_was(), exc_info=exc_info)
-            if _pdb_on_exception:
-                if sys.version_info >= (3, 5):
-                    traceback = exc_info.__traceback__
-                else:
-                    traceback = exc_info[2]
```
eric-wieser: Please don't remove compatibility code until we've gotten around to updating `setup.py` to not allow python 2.7

eric-wieser: Done in #1337

ktbarrett: Calling `pdb.post_mortem` without args works on both Python 2 and 3, we discussed this in the original PR but it was not addressed. But I guess it doesn't matter now.

eric-wieser: I don't think that's true - there's no exception in flight here, we're not in an `except` block.

ktbarrett: fixed

tests/test_cases/issue_1279/issue_1279.py
```diff
@@ -0,0 +1,16 @@
+"""
+Test that once a SimFailure occurs, no further tests are run
+"""
+import cocotb
+
+
+@cocotb.test(expect_error=cocotb.result.SimFailure, stage=1)
+def test_sim_failure_a(dut):
+    # invoke a deadlock
```
eric-wieser:
```suggestion
    # invoke a deadlock, as nothing is driving this clock
```

ktbarrett:
@eric-wieser better commit messages

eric-wieser:
Fantastic, thanks!

PR #1264, opened 2019-12-26
Fix with_timeout trigger

ktbarrett:
One cannot yield a list of lists to the scheduler as originally thought.

>TypeError: All triggers must be instances of Trigger! Got: list

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -163,7 +163,7 @@ class Timer(GPITrigger):
     def __init__(self, time_ps, units=None):
         """
         Args:
-           time_ps (numbers.Real or decimal.Decimal): The time value. 
+           time_ps (numbers.Real or decimal.Decimal): The time value.
                Note that despite the name this is not actually in picoseconds
                but depends on the *units* argument.
            units (str or None, optional): One of
@@ -174,24 +174,24 @@ class Timer(GPITrigger):
         Examples:
 
             >>> yield Timer(100, units='ps')
-    
+
             The time can also be a ``float``:
-    
+
             >>> yield Timer(100e-9, units='sec')
-    
+
             which is particularly convenient when working with frequencies:
-    
+
             >>> freq = 10e6  # 10 MHz
             >>> yield Timer(1 / freq, units='sec')
-    
+
             Other builtin exact numeric types can be used too:
-    
+
             >>> from fractions import Fraction
             >>> yield Timer(Fraction(1, 10), units='ns')
-    
+
             >>> from decimal import Decimal
             >>> yield Timer(Decimal('100e-9'), units='sec')
-    
+
             These are most useful when using computed durations while
             avoiding floating point inaccuracies.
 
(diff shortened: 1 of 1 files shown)


eric-wieser:
What are your thoughts on the claim that this is by design, and the user must user `First` explicitly, argued from the perspective that a user of `await` can't use raw lists either?

ktbarrett:
I dislike the interface as it is. I'd prefer...

```python
yield with_timeout(..., timeout_time=10, timeout_units='ns')  # keyword only
```

or

```python
yield with_timeout(10, 'ns')(...)
```

I had preferred both of these to what was implemented when I was implementing it, I'm not sure how I got to what we have. Luckily, it's not been in the wild for long so we can probably change it.

eric-wieser:
+1 for keyword-only

ktbarrett:
@eric-wieser This is ready for review.

cocotb/triggers.py
```diff
@@ -732,24 +732,23 @@ def _wait(self):
...
-            Any expression that could be right of a :keyword:`yield`
-            (or :keyword:`await` in Python 3) expression in cocotb.
-        timeout_time (int):
+        *triggers (cocotb_waitable):
+            Triggers to wait on
+        **timeout_time (int):
             Time duration
-        timeout_unit (str, optional):
+        **timeout_unit (str):
             Units of duration, accepts any values that :class:`~cocotb.triggers.Timer` does
```
eric-wieser: Don't change the docs here

cocotb/triggers.py
```diff
@@ -732,24 +732,23 @@ def _wait(self):
 
 
 @decorators.coroutine
-def with_timeout(triggers, timeout_time, timeout_unit=None):
+def with_timeout(*triggers, **timeout_spec):
```
```suggestion
def with_timeout(*triggers, **kwargs):
```

cocotb/triggers.py
```diff
@@ -760,9 +759,12 @@ def with_timeout(triggers, timeout_time, timeout_unit=None):
 
     .. versionadded:: 1.3
     """
-    timeout_timer = cocotb.triggers.Timer(timeout_time, timeout_unit)
-    res = yield [timeout_timer, triggers]
-    if res is timeout_timer:
+    timeout_time = timeout_spec.get('timeout_time')
+    timeout_unit = timeout_spec.get('timeout_unit')
```
```suggestion
    timeout_unit = kwargs.pop('timeout_unit')
    reject_remaining_kwargs(kwargs)
```

cocotb/triggers.py
```diff
@@ -760,9 +759,12 @@ def with_timeout(triggers, timeout_time, timeout_unit=None):
 
     .. versionadded:: 1.3
     """
-    timeout_timer = cocotb.triggers.Timer(timeout_time, timeout_unit)
-    res = yield [timeout_timer, triggers]
-    if res is timeout_timer:
+    timeout_time = timeout_spec.get('timeout_time')
```
```suggestion
    timeout_time = kwargs.pop('timeout_time')
```

cocotb/triggers.py
cmarqu: I also touched this in #1276.

ktbarrett:
If that's the case I could just change the documentation to not hint that you can pass a list, it will take only one trigger always. No Python would change at that point.

eric-wieser:
That change would be fine by me too, with or without also making `timeout_unit` and `timeout_time` keyword-only

PR #1266, opened 2019-12-27
Stop setting `logger.propagate=True` on every logger

### LINKED ISSUE #1212: Why is the default logger overridden?

eric-wieser:
There are two ways to configure logger handlers in python:

1. Attach a handler to the root logger, and use the fact that records logged to child loggers will automatically inherit this handler by default.
2. Attach a handler to every logger, and disable propagation for every single one.

Up to this commit, we use approach 2, by subclassing `logging.Logger` to use this behavior by default.
The problem with this approach is it makes it very difficult for a user to set their own logging format, as their nested loggers will not propagate as they are used to.

This commit changes us to use the approach 1.
To most cocotb application which do not attempt to reconfigure the logger, this will have no effect.
Applications which configure their own loggers will find that nested loggers now work as they expected.

This also moves all of the logging initialization out of `___init__.py` to `log.py`, to keep everything together.

Another side effect of this change is that `logging.basicConfig` can now be used to remove the cocotb formatting and handlers.

Closes #1212, with the answer being "for backwards compatibility only"

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -39,8 +39,8 @@ import time
 import warnings
 
 import cocotb.handle
+import cocotb.log
 from cocotb.scheduler import Scheduler
-from cocotb.log import SimBaseLog, SimLog
 from cocotb.regression import RegressionManager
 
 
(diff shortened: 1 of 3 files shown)


eric-wieser:
cc @ktbarrett

cocotb/log.py
```diff
@@ -51,20 +51,48 @@
...
+
+    logging.setLoggerClass(SimBaseLog)  # For backwards compatibility
+    logging.basicConfig(handlers=[hdlr])
+
+    # apply level settings for cocotb
+    log = logging.getLogger('cocotb')
+    level = os.getenv("COCOTB_LOG_LEVEL", "INFO")
+    try:
+        _default_log = getattr(logging, level)
+    except AttributeError as e:
```
ktbarrett: unused variable

eric-wieser:
Test failure is an existing bug which is now exposed - `RunningTest.error_messages` previously did not work. After this change, it working causes a test to fail.

cocotb/log.py
eric-wieser: Thanks, fixed

ktbarrett:
Is it even a good idea to continue to support `error_messages`? It never worked, obviously no one has ever needed it. It looks like it saves the entire log output (`log_level=DEBUG`) to the result file if the test fails? That's a good way to explode memory usage and file size. I have log files in excess of 1GB and that + overhead will now have to be stored in memory until the test ends?

eric-wieser:
That's a good enough question that I'll split this PR so that I don't have to answer it to fix gh-1212.

cocotb/__init__.py
```diff
@@ -56,20 +56,9 @@
...
-    log.setLevel(_default_log)
-    loggpi = SimLog('cocotb.gpi')
-    # Notify GPI of log level
-    simulator.log_level(_default_log)
+    cocotb.log.default_config()
+
+    log = logging.getLogger(__name__)
 
     # If stdout/stderr are not TTYs, Python may not have opened them with line
     # buffering. In that case, try to reopen them with line buffering
```
eric-wieser: After thinking some more, these lines should now be _above_ the logging config.

Before, what happened was:

* The `cocotb` log object was configured with stdout
* `stdout` was fixed
* The `cocotb.scheduler` and all other nested log objects were configured

As currently written, the behavior is

* All log objects are configured with stdout
* `stdout` is fixed

This probably ought to be
* `stdout` is fixed
* All log objects are configured with stdout

cocotb/decorators.py
```diff
@@ -229,7 +229,13 @@ def __init__(self, fn):
...
 
         def handle(self, record):
-            self.fn(self.format(record))
+            # For historical reasons, only logs sent directly to the `cocotb`
+            # logger are recorded - logs to `cocotb.scheduler` for instance
+            # are not recorded. Changing this behavior may have significant
+            # memory usage implications, so should not be done without some
+            # thought.
+            if record.name == 'cocotb':
+                self.fn(self.format(record))
```
eric-wieser: Moved to #1268

cocotb/__init__.py
```diff
@@ -56,36 +56,42 @@
...
+    _stderr_ok = _reopen_stream_with_buffering('stderr')
+
+    # don't set the logging up until we've attempted to fix the standard IO,
+    # otherwise it will end up connected to the unfixed IO.
+    cocotb.log.default_config()
+    log = logging.getLogger(__name__)
+
+    # we can't log these things until the logging is set up!
+    if _stderr_ok is True:
+        log.debug("Reopened stderr with line buffering")
```
imphil: These log messages are nice, but don't convey much useful information. (What are you doing with that info?) It would simplify the code not to log them.

cocotb/log.py
```diff
@@ -51,20 +51,49 @@
...
+    level = os.getenv("COCOTB_LOG_LEVEL", "INFO")
+    try:
+        _default_log = getattr(logging, level)
+    except AttributeError:
+        log.error("Unable to set logging level to %r" % level)
+        _default_log = logging.INFO
+    log.setLevel(_default_log)
+
+    # Notify GPI of log level, which it uses as an optimization to avoid
+    # calling into python.
```
imphil: nit: python -> Python

cocotb/__init__.py
```diff
@@ -56,36 +56,42 @@
...
-        if not sys.stdout.isatty():
-            sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 1)
-            log.debug("Reopened stdout with line buffering")
-        if not sys.stderr.isatty():
-            sys.stderr = os.fdopen(sys.stderr.fileno(), 'w', 1)
-            log.debug("Reopened stderr with line buffering")
-    except Exception as e:
-        log.warning("Failed to ensure that stdout/stderr are line buffered: %s", e)
+    _stdout_ok = _reopen_stream_with_buffering('stdout')
+    _stderr_ok = _reopen_stream_with_buffering('stderr')
```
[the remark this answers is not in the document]
eric-wieser: `stdout_buf_result` seems easiest to me

cocotb/__init__.py
eric-wieser: They're marked `debug` because they're probably not very useful :). It seems like for someone debugging why their exception tracebacks are being swallowed, this might provide a little bit of useful info.

PR #1268, opened 2019-12-28
Fix the broken log record accumulation

eric-wieser:
Before this commit, the assertion in this test would fail, and `str_counter` would be equal to the total number of tests.
The bug was that the log handlers would be installed at startup before any test was run, and never removed - so every test logged to the captured logs of every other tests.

Now the regression manager installs the handler before starting the test, and removes it when the test is complete.
As a result, log messages are recorded exactly twice - once in stdout, and once in the results.xml file.

Tested by injecting a failure in a test, and looking at results.xml manually to see that all the logs appear.

---

Split from #1266.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -243,21 +243,17 @@ class RunningTest(RunningCoroutine):
         self.skip = parent.skip
         self.stage = parent.stage
 
-        self.handler = RunningTest.ErrorLogHandler(self._handle_error_message)
-        cocotb.log.addHandler(self.handler)
+        # make sure not to create a circular reference here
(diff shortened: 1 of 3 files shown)


ktbarrett:
I don't see the value in saving every message to the results file. Simulators support redirection of standard output to a file. Storing all those messages in memory for the duration of the test will be a performance regression. And as indicated, no one has used the feature because it's been broken for a while and no issues have been opened against it.

eric-wieser:
The feature isn't broken, it just stored 60x more copies before this patch compared to after. This doesn't change the scope of which messages are logged

ktbarrett:
> it just stored 60x more copies before this patch compared to after
Oh jeez, okay, I see what was going on now. I guess my position is that we should remove the feature.

PR #1271, opened 2019-12-31
Make deprecation warning clearer.

cmarqu:


cocotb/result.py
```diff
@@ -73,14 +73,14 @@ def create_error(obj, msg):
...
-        Use ``TestError(msg)`` directly instead of this function.
+        Use ``raise TestError(msg)`` directly instead of this function.
 
     Args:
         obj: Object with a log method.
         msg (str): The log message.
     """
     warnings.warn(
-        "``create_error`` is deprecated - use ``TestError(msg)`` (or any other "
+        "``create_error`` is deprecated - use ``raise TestError(msg)`` (or any other "
```
eric-wieser: This change isn't correct

cmarqu: Oh okay, it looked like an inconsistency.

cocotb/result.py
```diff
@@ -73,14 +73,14 @@ def create_error(obj, msg):
     simply to avoid too many levels of nested `try/except` blocks.
 
     .. deprecated:: 1.3
-        Use ``TestError(msg)`` directly instead of this function.
+        Use ``raise TestError(msg)`` directly instead of this function.
```
eric-wieser: This change _is_ correct.

cocotb/result.py
eric-wieser: See above- the other change is correct

cmarqu:
Good now?

cocotb/result.py
eric-wieser: Whoops, I somehow got this backwards. Sorry! The either one is correct n

eric-wieser:
Sorry, I think I completely misread this patch. Both the docstring and message for raise_error should include the word raise. Create_error should not change.

cmarqu:
I think it might be correct now - I just added the word "directly", no new "raise".
If not correct, I can work on it again in the decade where Python 2 is deprecated. Cheers to that!

eric-wieser:
Not sure this is really that much clearer, but it's no worse, so why not

PR #1274, opened 2020-01-01
Added --trace-structs to Makefile.verilator.

nndurj:
By default, structs are dumped as single dimensional packed vectors in
the VCD file. Adding this will make verilator dump structs as
hierarchial objects with struct fields as children.

### Code

```systemverilog
module t();

struct packed {
    logic [15:0] src;
    logic [15:0] dst;
} hdr;

assign hdr.src = 20;
assign hdr.dst = 40;

endmodule
```

### Before
image

### After
image


themperek:
@wallento Is this ok?

wallento:
yes, lgtm, thanks @nndurj

eric-wieser:
Thanks for the review @wallento

Issue #1115, opened 2019-09-18
Timer value argument is misleading 


cmarqu:
With the existence of the ``units`` argument, the value in ``Timer`` should not be ``time_ps`` (https://cocotb.readthedocs.io/en/latest/triggers.html?highlight=Timer#cocotb.triggers.Timer). 

How about ``value``?

ktbarrett:
Regarding this and #1117. 

I don't even think the default units are picoseconds, but the simulator's timescale, which can change via flags to the simulator. Not using `units` is probably a bug/bug waiting to happen. Maybe a `DeprecationWarning` is appropriate.

cmarqu:
Maybe we should even take the step and create ``SimTime`` as @imphil suggested in #872.

ktbarrett:
I'd like something like `std::chrono::duration` for expressing sim time, I've run into the need for such a type on a few occassions.

ktbarrett:
@garmin-mjames What is the request for 1.5 here? The rename, or the simulation time physical type I described?

garmin-mjames:
A rename was my goal for 1.5.
Supporting a physical `time` type in Python would help #1233 as well.

cmarqu:
https://stackoverflow.com/questions/49802412/how-to-implement-deprecation-in-python-with-argument-alias could be useful here

ktbarrett:
The first part of the first solution would not be good for Sphinx, but the second part would work fine. I would introduce two special objects that would be helpful for showing the intended use.

```python
DEPRECATED = object()
REQUIRED = object()

def function(new_name=REQUIRED, *, old_name=DEPRECATED):
    # ... logic from answer 1 part 2 just replacing `None`, `None` might be a possible argument anyway, so we need a better sentinel
```

PR #1276, opened 2020-01-01
Improve documentation; fix typos.

cmarqu:
Document arguments of ``Timer`` since now also e.g. ``timeout_unit`` of ``cocotb.test`` refers to it for the acceptable units.
Related: #1115, #1130

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -161,11 +161,48 @@ class GPITrigger(Trigger):
 class Timer(GPITrigger):
     """Fires after the specified simulation time period has elapsed."""
     def __init__(self, time_ps, units=None):
+        """
+        Args:
+           time_ps (numbers.Real or decimal.Decimal): The time value. 
+               Note that despite the name this is not actually in picoseconds
+               but depends on the *units* argument.
+           units (str or None, optional): One of
+               ``None``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
+               When no *units* is given (``None``) the timestep is determined by
+               the simulator.
+
+        Examples:
+
+            >>> yield Timer(100, units='ps')
+    
+            The time can also be a ``float``:
+    
+            >>> yield Timer(100e-9, units='sec')
+    
+            which is particularly convenient when working with frequencies:
+    
+            >>> freq = 10e6  # 10 MHz
+            >>> yield Timer(1 / freq, units='sec')
+    
+            Other builtin exact numeric types can be used too:
+    
+            >>> from fractions import Fraction
+            >>> yield Timer(Fraction(1, 10), units='ns')
+    
+            >>> from decimal import Decimal
+            >>> yield Timer(Decimal('100e-9'), units='sec')
+    
+            These are most useful when using computed durations while
+            avoiding floating point inaccuracies.
+
+        See Also:
+            :func:`~cocotb.utils.get_sim_steps`
+        """
         GPITrigger.__init__(self)
         self.sim_steps = get_sim_steps(time_ps, units)
 
     def prime(self, callback):
-        """Register for a timed callback"""
+        """Register for a timed callback."""
         if self.cbhdl == 0:
             self.cbhdl = simulator.register_timed_callback(self.sim_steps,
                                                            callback, self)
@@ -418,7 +455,7 @@ class Lock(object):
         self._pending_unprimed = []
         self._pending_primed = []
         self.name = name
-        self.locked = False  #: True if the lock is held
+        self.locked = False  #: ``True`` if the lock is held.
 
     def _prime_trigger(self, trigger, callback):
         self._pending_unprimed.remove(trigger)
@@ -694,7 +731,7 @@ class First(_AggregateWaitable):
             w.kill()
 
         # These lines are the way they are to make tracebacks readable:
-        #  - The comment helps the user nderstand why they are seeing the
+        #  - The comment helps the user understand why they are seeing the
         #    traceback, even if it is obvious top cocotb maintainers.
         #  - Raising ReturnValue on a separate line avoids confusion about what
         #    is actually raising the error, because seeing
@@ -708,13 +745,14 @@ class First(_AggregateWaitable):
 
 
 class ClockCycles(Waitable):
-    """
-    Fires after *num_cycles* transitions of *signal* from ``0`` to ``1``.
-    """
+    """Fires after *num_cycles* transitions of *signal* from ``0`` to ``1``."""
     def __init__(self, signal, num_cycles, rising=True):
         """
-        :param rising: If true, the default, count rising edges. Otherwise,
-            count falling edges
+        Args:
+            signal: The signal to monitor.
+            num_cycles (int): The number of cycles to count.
+            rising (bool, optional): If ``True``, the default, count rising edges.
+                Otherwise, count falling edges.
         """
         self.signal = signal
         self.num_cycles = num_cycles
@@ -734,7 +772,7 @@ class ClockCycles(Waitable):
 @decorators.coroutine
 def with_timeout(triggers, timeout_time, timeout_unit=None):
     """
-    Waits on triggers, throws an exception if it waits longer than the given time
+    Waits on triggers, throws an exception if it waits longer than the given time.
 
     Usage:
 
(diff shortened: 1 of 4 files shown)


cocotb/triggers.py
```diff
@@ -161,11 +161,36 @@ def unprime(self):
 class Timer(GPITrigger):
     """Fires after the specified simulation time period has elapsed."""
     def __init__(self, time_ps, units=None):
+        """
+        .. Constructor. This RST comment works around sphinx-doc/sphinx#6885
```
eric-wieser: Don't think this is needed, it's only for when the docstring is inherited, and this one isn't

cocotb/triggers.py
```diff
@@ -161,11 +161,36 @@ def unprime(self):
...
+
+        Args:
+           time_ps: The time value. This can also be a 
+               :class:`fractions.Fraction` or a :class:`decimal.Decimal` instance.
```
eric-wieser: I'd move this sentence to the end

cocotb/triggers.py
```diff
@@ -708,13 +733,15 @@ def on_done(ret):
...
 
 class ClockCycles(Waitable):
-    """
-    Fires after *num_cycles* transitions of *signal* from ``0`` to ``1``.
-    """
     def __init__(self, signal, num_cycles, rising=True):
         """
-        :param rising: If true, the default, count rising edges. Otherwise,
-            count falling edges
+        Fires after *num_cycles* transitions of *signal* from ``0`` to ``1``.
```
eric-wieser: I'd put this back in the class docstring

cocotb/triggers.py
```diff
@@ -161,11 +161,36 @@ def unprime(self):
...
+
+        Args:
+           time_ps: The time value. This can also be a 
+               :class:`fractions.Fraction` or a :class:`decimal.Decimal` instance.
+               Note that despite the name this is not actually in picoseconds
+               but depends on the *units* argument.
+           units (str, optional): One of
+               ``None``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
+               When no *units* is given (``None``) the timestep is determined by
+               the simulator.
```
eric-wieser: I'm tempted to just refer to the docs for `get_sim_steps` here, rather than listing all the values.

cocotb/triggers.py
cmarqu: Oh, right.

cocotb/triggers.py
cmarqu: Done.

cocotb/triggers.py
cmarqu: I have since re-arranged that part a bit, please take another look.

cocotb/triggers.py
cmarqu: This would give us two indirections from the docs of the timeout feature, and would also need more cleanup because we already duplicate this info in other places. I'd say dupli/triplicate for now and not open another can of worms at this time.
The real fix would be ``SimTime`` from #1130 which unfortunately @ktbarrett has closed.

eric-wieser: Can we compromise and do both? So add
```
See Also:
```

cmarqu: Sure, done.

cocotb/triggers.py
```diff
@@ -748,15 +775,15 @@ def with_timeout(triggers, timeout_time, timeout_unit=None):
             Any expression that could be right of a :keyword:`yield`
             (or :keyword:`await` in Python 3) expression in cocotb.
         timeout_time (int):
```
cmarqu: Is this actually a ``numbers.Number`` too? At least floats are handled here, right? Same goes for ``Timer``s ``time_ps`` argument.

eric-wieser: Turns out I documented the other one wrongly - strictly this is `numbers.Real or decimal.Decimal`, since complex numbers are forbidden, and `Decimal` isn't a number at all!

cocotb/triggers.py
```diff
@@ -161,11 +161,37 @@ def unprime(self):
...
+
+        .. code-block:: python
+    
+            yield Timer(100, units='us')
+            # alternate spelling 1:
+            from fractions import Fraction
+            yield Timer(Fraction(1, 10000), units='sec')
+            # alternate spelling 2:
+            from decimal import Decimal
+            yield Timer(Decimal('1e-9'), units='sec')
```
eric-wieser: Examples:

            >>> yield Timer(100, units='ps')

            The time can also be a ``float``:

            >>> yield Timer(100e-9, units='sec')

            which is particularly convenient when working with frequencies:

            >>> freq = 10e6  # 10 MHz
            >>> yield Timer(1 / freq, units='sec')

        	Other builtin exact numeric types can be used too:

            >>> from fractions import Fraction
            >>> yield Timer(Fraction(1, 10), units='ns')

            >>> from decimal import Decimal
            >>> yield Timer(Decimal('100e-9'), units='sec')

            These are most useful when using computed durations while
            avoiding floating point inaccuracies.

cmarqu: I put this in.

cocotb/triggers.py
cmarqu: I now changed this and ``get_sim_steps`` to ``numbers.Real or decimal.Decimal`` but left ``Timer(time_ps)`` without type description since the examples are doing a good job for that.

cocotb/triggers.py
```diff
@@ -161,11 +161,50 @@ def unprime(self):
...
+               a :class:`decimal.Decimal` instance.
+               Note that despite the name this is not actually in picoseconds
+               but depends on the *units* argument.
+           units (str or None, optional): One of
+               ``None``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
+               When no *units* is given (``None``) the timestep is determined by
+               the simulator.
+
+               See Also:
+                   :func:`~cocotb.utils.get_sim_steps`
```
eric-wieser: This should be unindented to the same level as `Args`

cocotb/triggers.py
```diff
@@ -161,11 +161,50 @@ def unprime(self):
...
+
+     Examples:
```
eric-wieser: This should be indented to the same level as `Args`

cocotb/triggers.py
```diff
@@ -161,11 +161,50 @@ def unprime(self):
...
+        The time can also be a ``float``:
+
+        >>> yield Timer(100e-9, units='sec')
+
+        which is particularly convenient when working with frequencies:
+
+        >>> freq = 10e6  # 10 MHz
+        >>> yield Timer(1 / freq, units='sec')
+
+	Other builtin exact numeric types can be used too:
```
eric-wieser: This should be indented to the same level as the `>>>`

cocotb/triggers.py
```diff
@@ -161,11 +161,50 @@ def unprime(self):
...
+               ``None``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
+               When no *units* is given (``None``) the timestep is determined by
+               the simulator.
+
+               See Also:
+                   :func:`~cocotb.utils.get_sim_steps`
+
+     Examples:
+
+        >>> yield Timer(100, units='ps')
```
eric-wieser: This and everything below it should be indented under `Examples`

cocotb/triggers.py
cmarqu: Done.

cocotb/triggers.py
cmarqu: Tabs vs spaces, don't you love it...

cocotb/triggers.py
```diff
@@ -161,11 +161,50 @@ def unprime(self):
...
+               a :class:`decimal.Decimal` instance.
+               Note that despite the name this is not actually in picoseconds
+               but depends on the *units* argument.
+           units (str or None, optional): One of
+               ``None``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
+               When no *units* is given (``None``) the timestep is determined by
+               the simulator.
+
+        See Also:
+            :func:`~cocotb.utils.get_sim_steps`
```
eric-wieser: Nit: this maybe works better below the examples, not above.

Maybe we should just add more `See also` sections in a later doc improvement round, and fix this then.

cocotb/triggers.py
```diff
@@ -161,11 +161,50 @@ def unprime(self):
 class Timer(GPITrigger):
     """Fires after the specified simulation time period has elapsed."""
     def __init__(self, time_ps, units=None):
+        """
+        Args:
+           time_ps: The time value.
```
```suggestion
           time_ps (numbers.Real or decimal.Decimal): The time value. 
```

cocotb/triggers.py
```diff
@@ -161,11 +161,50 @@ def unprime(self):
 class Timer(GPITrigger):
     """Fires after the specified simulation time period has elapsed."""
     def __init__(self, time_ps, units=None):
+        """
+        Args:
+           time_ps: The time value. 
+               This can also be a :class:`fractions.Fraction` or
+               a :class:`decimal.Decimal` instance.
```
eric-wieser: Maybe just remove these lines, the example covers it just fine.

eric-wieser:
Little point waiting for CI on a doc change, let's put this in as is

PR #1280, opened 2020-01-03
Differentiate set_signal_value for logic string from ASCII string

### LINKED ISSUE #1254: Munging of setting binary and ASCII strings in GPI/VPI
As discovered in #834 the `access_var_string` test fails in Questa when using the VPI. A solution is to change the format in the following snippet from `vpiBinStrVal` to `vpiStringVal`.

https://github.com/cocotb/cocotb/blob/92a032a4349b2fe9e2f9bff5d1060ef44cd4cd72/cocotb/share/lib/vpi/VpiCbHdl.cpp#L346-L357

Conceptually this is changing the format from a binary string to an ASCII string. This makes the test pass on Questa because we are writing actual strings on the failing test. However, this fix causes other tests to fail on Icarus.

Fundamentally, setting ASCII strings and binary string are different operation on different types, which is why there are distinct VPI format types. But the GPI only has one function overload for setting signal values to strings. For the VPI it makes the assumption that `vpiBinStrVal` is correct, and this works for Icarus by "happenstance". Really the whole thing is just wrong, there should be two functions: one for handling binary strings, and the other for ASCII strings. This requires modifying the GPI and all the interfaces.

ktbarrett:
Fixes #1254.

Previously there were two methods for getting string values from the simulator `get_signal_value_str` and `get_signal_value_binstr`, but there was only one method for setting strings `set_signal_value_str`. This was mostly enforced as binary string, however the VHPI handled both cases. Icarus worked with only this one method setting both binary strings and ASCII strings, but Questa does not. Questa expects strings to be set with `vpiStringVal` instead of `vpiBinStrVal` as discovered by #834.

`set_signal_value_str` has been split into two functions to mirror `get_signal_value_(bin)str`. Testing with Questa and other simulator should be done before acceptance.

### DIFF
--- a/cocotb/share/include/gpi.h
+++ b/cocotb/share/include/gpi.h
@@ -197,7 +197,8 @@ int gpi_is_indexable(gpi_sim_hdl gpi_hdl);
 // Functions for setting the properties of a handle
 void gpi_set_signal_value_real(gpi_sim_hdl gpi_hdl, double value);
 void gpi_set_signal_value_long(gpi_sim_hdl gpi_hdl, long value);
-void gpi_set_signal_value_str(gpi_sim_hdl gpi_hdl, const char *str);    // String of binary char(s) [1, 0, x, z]
+void gpi_set_signal_value_binstr(gpi_sim_hdl gpi_hdl, const char *str); // String of binary char(s) [1, 0, x, z]
+void gpi_set_signal_value_str(gpi_sim_hdl gpi_hdl, const char *str);    // String of ASCII char(s)
 
 typedef enum gpi_edge {
     GPI_RISING = 1,
--- a/cocotb/share/lib/fli/FliObjHdl.cpp
+++ b/cocotb/share/lib/fli/FliObjHdl.cpp
@@ -145,7 +145,13 @@ int FliValueObjHdl::set_signal_value(const long value)
     return -1;
 }
 
-int FliValueObjHdl::set_signal_value(std::string &value)
+int FliValueObjHdl::set_signal_value_binstr(std::string &value)
+{
+    LOG_ERROR("Setting signal/variable value via string not supported for %s of type %d", m_fullname.c_str(), m_type);
+    return -1;
+}
+
+int FliValueObjHdl::set_signal_value_str(std::string &value)
 {
     LOG_ERROR("Setting signal/variable value via string not supported for %s of type %d", m_fullname.c_str(), m_type);
     return -1;
@@ -330,7 +336,7 @@ int FliLogicObjHdl::set_signal_value(const long value)
     return 0;
 }
 
-int FliLogicObjHdl::set_signal_value(std::string &value)
+int FliLogicObjHdl::set_signal_value_binstr(std::string &value)
 {
     if (m_fli_type == MTI_TYPE_ENUM) {
         mtiInt32T enumVal = m_enum_map[value.c_str()[0]];
@@ -501,7 +507,7 @@ const char* FliStringObjHdl::get_signal_value_str()
     return m_val_buff;
 }
 
-int FliStringObjHdl::set_signal_value(std::string &value)
+int FliStringObjHdl::set_signal_value_str(std::string &value)
 {
     strncpy(m_mti_buff, value.c_str(), m_num_elems);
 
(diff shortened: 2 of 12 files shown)


cocotb/share/lib/fli/FliObjHdl.cpp
```diff
@@ -341,7 +347,7 @@ int FliLogicObjHdl::set_signal_value(std::string &value)
             mti_SetSignalValue(get_handle<mtiSignalIdT>(), enumVal);
         }
     } else {
-
```
eric-wieser: Mind doing these whitespace cleanups in a standalone PR? They make reviewing a little awkward

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -668,6 +668,11 @@ int VhpiSignalObjHdl::set_signal_value(std::string &value)
     return 0;
 }
 
+int VhpiSignalObjHdl::set_signal_value_binstr(std::string &value)
+{
+    return set_signal_value_str(value);
```
eric-wieser: A comment here explaining that VHPI does not distinguish these cases would be useful

cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -356,6 +356,19 @@ int VpiSignalObjHdl::set_signal_value(std::string &value)
...
 }
 
+int VpiSignalObjHdl::set_signal_value_str(std::string &value)
+{
+    s_vpi_value value_s;
+
+    std::vector<char> writable(value.begin(), value.end());
+    writable.push_back('\0');
+
+    value_s.value.str = &writable[0];
```
eric-wieser: nit: `value.c_str()`, which includes the trailing 0. Fine to leave to a follow-up PR.

eric-wieser:
Might be good for me to get #1149 in first

cocotb/share/lib/fli/FliObjHdl.cpp
ktbarrett: There is a radio box in the top right corner when reviewing to turn off whitespace diffs. You can add `?w=1` to the URL if you can't find it. My editor auto-trims trailing whitespace, I don't do them manually. I will do a whitespace cleanup PR in future, but we really should run whitespace checks on PRs, I have no clue how they get in there anyways, are people still using WordPerfect?

cocotb/share/lib/vpi/VpiCbHdl.cpp
ktbarrett: I was going to do that for both functions, but I get poor reception to changing non-pertinent code, and I am not making the implementations inconsistent. I will change both here though.

ktbarrett: Actually, there is a `const`ness issue with `c_str`. I am leaving it as it is.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
ktbarrett: I changed the implementation so that `binstr` only works with `enum`s and `logic`s and `str` only works with `string`s.

cocotb/share/lib/vpi/VpiCbHdl.cpp
eric-wieser: Good catch

cocotb/share/include/gpi.h
```diff
@@ -197,7 +197,8 @@ int gpi_is_indexable(gpi_sim_hdl gpi_hdl);
 // Functions for setting the properties of a handle
 void gpi_set_signal_value_real(gpi_sim_hdl gpi_hdl, double value);
 void gpi_set_signal_value_long(gpi_sim_hdl gpi_hdl, long value);
-void gpi_set_signal_value_str(gpi_sim_hdl gpi_hdl, const char *str);    // String of binary char(s) [1, 0, x, z]
+void gpi_set_signal_value_binstr(gpi_sim_hdl gpi_hdl, const char *str); // String of binary char(s) [1, 0, x, z]
+void gpi_set_signal_value_str(gpi_sim_hdl gpi_hdl, const char *str);    // String of ASCII char(s)
```
eric-wieser: Can you use `/**  String of ASCII char(s) */` instead, which I think will be picked up by doxygen?

ktbarrett: That wouldn't be picked up correctly, it must come the line before the declaration and there must be an empty line between declarations. Also, it would be the only thing in the file that has documentation. I think that should be a different commit.

eric-wieser: Fair

ktbarrett:
merged with #1256.

PR #1281, opened 2020-01-03
Remove trailing whitespace

ktbarrett:
Remove trailing whitespace in all files and adds trailing whitespace checks to Travis.


eric-wieser:
@imphil, ok with this in principle? (Obviously not for 1.3)

If so, I'll go through the whitespace-sensitive files and check that nothing is relying on trailing whitespace.

.travis.yml
```diff
@@ -14,6 +14,10 @@ services:
 
 jobs:
   include:
+    - stage: PythonLint
+      os: linux
+      script: "[[ -z \"$(git ls-files | xargs egrep -l '\s+$')\" ]]"
```
ktbarrett: Still not too sure about this, I tried running this without the whitespace changes to see if it failed, but I guess it saw there were no changes and didn't run Travis.

eric-wieser: I think travis might have failed to parse the yaml

eric-wieser: This run is from an older version: https://travis-ci.org/cocotb/cocotb/jobs/632467451#L183

eric-wieser: I think the backslash needs escaping for yaml:
```suggestion
      script: "[[ -z \"$(git ls-files | xargs egrep -l '\\s+$')\" ]]"
```

ktbarrett: Yeah, that's all it needed. Travis failed with master sources and passed with fixed sources.

imphil:
Yes, that's a great change and enforcing this in CI is the right move. Thanks!

eric-wieser:
Thanks @ktbarrett. This is probably going to create conflicts in some open PRs, but hopefully they'll be easy to resolve

PR #941, opened 2019-05-16
Add support for vpiInterconnectNet, new file vpi_user_ext.h.

cmarqu:
I came across this ``vpiInterconnectNet`` definition with a Cadence AMS simulation. I haven't really exercised this in any way.
The newly introduced file ``vpi_user_ext.h`` could be used in a revised #641 as well.

PR #641, opened 2018-08-04
Add support for structured nets (IUS Real Number  Models).

nevsan:
This allows CocoTB to observe structured nets, like the `EEnets` that are used in IUS AMS/Real Number Model flows. You can't drive individual components (even in SystemVerilog), so the solution for driving `EEnet` types would be similar to the following:
```verilog
import EE_pkg::*;

module dut(inout EEnet out);
    real out_v, out_r;

    // EEnet = '{V, I, R};
    assign out = '{out_v, 0, out_r};
endmodule
```
You can then drive `out_v` and `out_r` from CocoTB. This change will allow you to also observe values via `dut.out.V.value`, for example.

I also made a small change to allow `float` convertible types (like `int`) to be assigned to `real` values.

PR #1282, opened 2020-01-04
Improve support for IUS Real Number Models

nevsan:
This is a rebase of #641 onto #941. It adds a little bit of extra feature support for `EEnet` types.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -664,8 +664,11 @@ class RealObject(ModifiableObject):
             TypeError: If target has an unsupported type for
                 real value assignment.
         """
-        if not isinstance(value, float):
-            self._log.critical("Unsupported type for real value assignment: %s (%s)", type(value), repr(value))
+        try:
+            value = float(value)
+        except ValueError:
+            self._log.critical("Unsupported type for real value assignment: %s (%s)" %
+                               (type(value), repr(value)))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
         simulator.set_signal_val_real(self._handle, value)
--- a/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -82,6 +82,7 @@ gpi_objtype_t to_gpi_objtype(int32_t vpitype)
         case vpiMemoryWord:
             return GPI_REGISTER;
 
+        case vpiRealNet:
         case vpiRealVar:
             return GPI_REAL;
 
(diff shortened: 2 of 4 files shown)


cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -155,9 +156,6 @@ GpiObjHdl* VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
         case vpiIntegerNet:
         case vpiRealVar:
         case vpiRealNet:
-        case vpiStringVar:
```
cmarqu: Are these removed intentionally?

nevsan: A casualty of the rebase. Fixed.

cocotb/share/include/vpi_user.h
```diff
@@ -853,6 +853,8 @@ typedef struct t_cb_data
 #define cbNBASynch               30
 #define cbAtEndOfSimTime         31
 
+/**************************** Verilog AMS ****************************/
+#define vpiRealNet             526
```
cmarqu: Actually, I had added this constant already to ``vpi_user_ext.h``, so this part of the PR is not necessary.

cmarqu:
Thanks very much. All the review comments from #641 have been addressed, and the new ``test_reals.assign_int`` works in Xcelium, Questa and VCS.

cocotb/handle.py
```diff
@@ -664,8 +664,11 @@ def setimmediatevalue(self, value):
             TypeError: If target has an unsupported type for
                 real value assignment.
         """
-        if not isinstance(value, float):
-            self._log.critical("Unsupported type for real value assignment: %s (%s)", type(value), repr(value))
+        try:
+            value = float(value)
```
eric-wieser: This is probably a bad idea, it allows nonsense like `.value = "1.5"`

cmarqu: That already slipped through the previous review round...

nevsan: One could put in a check like `isinstance(value, numbers.Number)` to catch all possible valid numeric values (including NumPy numeric types). However, these checks do start to get expensive if users put them in speed critical loops (and assigning values seems like an area where we should be careful about performance). A quick benchmark using `timeit` to compare some variations of this chunk of code.

1. This PR (`try: ... except:`): **220ns**
2. Original`isinstance(value, float)`: **250ns**
3. `isinstance(value, numbers.Number)`: **1,330ns**
4. `isinstance(value, float) or isinstance(value, int)`: **540ns**

I think letting some users do things like `.value = "1.5"`, although cringeworthy, is probably worth the 2-6x speed improvement we get from this as written. Some might even consider it a feature 😉.

eric-wieser: How does `isinstance(value, (float, int, numbers.Number))` compare?

nevsan: That one takes:
1. **515ns** if `value` is `float`.
2. **585ns** if `value` is `int`.
3. **1,580ns** if `value` is `np.uint8`.
4. **1,930ns** if `value` is `str`.

eric-wieser: Another option that should be both fast and correct is to use `value.__float__()`, with the same `try`  / `except` as you have now.

nevsan: That one is faster than the others at about **410ns**. I'm actually a bit surprised that it's not closer to the `float(value)` solution.

PR #1283, opened 2020-01-06
Use lock on read transaction

leastrobino:
New PR after changes mentioned in #1206


eric-wieser:
I don't think github is playing well here - to merge this, we need to first revert #1206 - otherwise github will just do a no-op commit

eric-wieser:
Oh crap, that didn't work at all - was trying to get github to recompute the diff

PR #1007, opened 2019-07-09
AXI4 Slave Driver - More complete implementation

cogsy23:
Hi All,

This is a mess I'm just hoping to get some feedback on (and yes I know it's not on master).  I've just got past my proof of concept stage and though I'd get the maintainers thoughts before I put too much more work in.

I'm currently designing an AXI master for my day job and need something to test it against.  It's fairly high performance and so requires advanced parts of the AXI standard such as bursts and threads.  The existing AXI slave driver is very simplistic, so this is my attempt to match more of the AXI standard.

I'm intending to implement both read and write transactions while including the (lack of) ordering between address and data channels, as well as bursts and threads.

PR #1285, opened 2020-01-06
Use lock on read transaction

eric-wieser:
Reopened from #1283, which I closed by accident, sorry @leastrobino

### DIFF
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -68,9 +68,7 @@ class AXI4LiteMaster(BusDriver):
 
     @cocotb.coroutine
     def _send_write_address(self, address, delay=0):
-        """
-        Send the write address, with optional delay (in clocks)
-        """
+        """Send the write address, with optional delay (in clocks)."""
         yield self.write_address_busy.acquire()
         for cycle in range(delay):
             yield RisingEdge(self.clock)
@@ -89,7 +87,7 @@ class AXI4LiteMaster(BusDriver):
 
     @cocotb.coroutine
     def _send_write_data(self, data, delay=0, byte_enable=0xF):
-        """Send the write address, with optional delay (in clocks)."""
+        """Send the write data, with optional delay (in clocks) and byte enable."""
         yield self.write_data_busy.acquire()
         for cycle in range(delay):
             yield RisingEdge(self.clock)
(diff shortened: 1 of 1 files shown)


cocotb/drivers/amba.py
```diff
@@ -198,6 +204,8 @@ def read(self, address, sync=True):
                 break
             yield RisingEdge(self.clock)
 
+        yield RisingEdge(self.clock)
```
eric-wieser: This doesn't look like it belongs in this patch

leastrobino: @eric-wieser, I copied the content of the `write` function into the `read` function for the sake of uniformity. Now both write and read functions end one rising edge after the slave has asserted `BVALID` or `RVALID`, which is the effective end of transaction.

eric-wieser: That may be, but I've seen a lot of code that depends on this edge _not_ being there.

At any rate, this is not part of the change to use a lock - its best to put unrelated changes like this line in a separate PR, so that they can be reviewed independently.

leastrobino: Okay, I removed that line.

Does it make sense to uniformize both functions? If a lot of code depends on this edge not being there, should I do it the other way - remove that edge in the write function as well?

eric-wieser: Well, it's likely that code depends on the edge not being there in the other function too - best to leave it for now, and maybe make a brand new `cocotb_ext.axilite` module in the somewhat distant future with all these things fixed.

eric-wieser:
Can you add a reply here summarizing what the purpose of this lock is?

leastrobino:
This implements the read transaction lock - which was defined in `__init__` but not used yet - in the same way it is done for the write transaction.

Situation in which it is needed:
```python
axi = AXI4LiteMaster(dut, 's_axi', clk)
addr = 0x3210
axi_read_1 = cocotb.fork(axi.read(addr))
yield Timer(delay,'ns') # if delay == 0, second transaction overrides the first one completely;
                        # if delay != 0, it interrupts the previous one (changes values while transaction is ongoing)
axi_read_2 = cocotb.fork(axi.read(addr+4))
```

Without lock, forking two or more AXI transactions fails, the last transaction overriding the previous ones even if they are still ongoing.
With lock, the next transaction waits for the end of the previous one before starting.

cmarqu:
@leastrobino Are you aware of #1248 and #1007 and maybe have feedback, as you apparently are an active AXI user? (If the lock is not implemented there, we should know and add it.)

ktbarrett:
@leastrobino If you are interested in getting this in, rebase this change and reopen it in [cocotb-bus](https://github.com/cocotb-bus).

leastrobino:
@ktbarrett Thanks, I have to admit this one slipped my mind. But it is now obsolete as it has been properly implemented in the new AXI4Master, on which AXI4LiteMaster is now based.

PR #1020, opened 2019-07-15
Fix for BinaryValue type

Matthewar:
A number of issues have noted an problem with the `BinaryValue` type, which seems to centre around the `big_endian` member variable. It is treated as both byte-order and bit-order in different places in the code, so I have attempted to clarify this in comments/function docstrings.

- When the binary value type was first added, there was no concept of endian-ness ([commit](https://github.com/Matthewar/cocotb/commit/55023b9644aaf0574579edc07e30a1166f166550#diff-77cf4987173ad37eab355c44a547ffefR56))
- The `bigEndian` argument was added to the constructor/`big_endian`member added to the class to convert to/from a bytestring IE it represents byte-ordering ([commit](https://github.com/Matthewar/cocotb/commit/761ead050745328d317250a6c20898aaa78db8cd#diff-77cf4987173ad37eab355c44a547ffefR106))
- When signed representation was added to the class, `big_endian` was used for bit ordering ([commit](https://github.com/Matthewar/cocotb/commit/030a25bdeae4d02db12becf461f33c5d4217c67f))

Big-endian doesn't seem to have been used or tested thoroughly, at least for when it represents bit ordering, which makes sense as hardware designs typically don't use a big endian bit order, and the languages don't really have the concept of bit-endianness in the same way:

- In 1364-2005 (the verilog specification) section 4.3.1: "The most significant bit specified by the _msb_ constant expression is the left-hand value in the range, and the least signficant bit specified by the _lsb_ constant expression is the right-hand value in the range".
    - Implicit extension/truncation will occur on the left side, even if an ascending range (e.g. `logic [0:3]`) is declared.
    - Implicit integer conversion will always result in the vector being treated as having a little-endian bit order
- In the 1076-2002 (VHDL revision) section 16.8.5.1 which defines the `numeric_std`package, `signed` and `unsigned` types clearly define: "The type UNSIGNED represents an unsigned binary integer with the most significant bit on the left, while the type SIGNED represents a two's-complement binary integer with the most significant bit on the left."
    - Explicit extension/truncation can occur on whichever side (left/right) you want, so for big-endian bit order signals (e.g. `signal example : std_logic_vector(0 to 9);`) you can extend/truncate on the right side if you'd prefer
    - The `RESIZE` function acts like Verilog implicit truncation/extension and always operations on the left side
    - `TO_INTEGER` function takes a little-endian bit-order vector

To solve this:
- Returned `big_endian` member to represent the byte order of the internal string
[...]

Issue #608, opened 2018-04-02
wrong BinaryValue assign/initialization with 3 bits


Fatsie:
Following code:
```#!python
    v = BinaryValue(3,3)
    print(v.integer)
```
gives 6 as output
This is under python 3.6

wstucker:
@Fatsie  Your example used the default kwarg of bigEndian=True. I think that setting bigEndian=False would probably give you the expected behavior.

```
be = BinaryValue(3,3)
print("Big Endian", be.integer)
le = BinaryValue(3,3, bigEndian=False)
print("Little Endian", le.integer)
```

Yields:

```
('Big Endian', 6)
('Little Endian', 3)
```

Fatsie:
I can't see this reply online so I reply by email.
OK, but I think this still indicates inconsistency with how bigEndian
value is treated during initialization and conversion back to integer.
greets,Staf.
William Stucker schreef op do 05-04-2018 om 23:15 [+0000]:

Fatsie:
Did some more testing with results given below.

init int | init length | init bigEndian | .binstr | .integer
-------- | ----------- | ------------------ | -------- | ---------
3 | 3 | True | 110 | 6
4 | 3 | True | 100 | 4
3 | 3 | False | 011 | 3
4 | 3 | False | 100 | 4
5 | 5 | True | 10100 | 20
2 | 5 | True | 10000 | 16
5 | 5 | False | 00101 | 5
2 | 5 | False | 00010 | 2

So to me the bug seems to be that conversion from and to integer is wrong for bigEndian=True

eric-wieser:
Seems I re-reported this as #636

hashhsah:
I hit this bug, too. Looking forward to its fix being merged.

In addition, I wonder `if big_endian` could be a class variable so I can set endianess globally in a project. This way, I would still have the option of setting the opposite endianess to a few `BinaryValue` object.

juliusbaxter:
I'd second exposing the BinaryValue default endianess as a global switch which can be set.

I'm not sure how Pythonic that is, or whether this would result in compatibility issues, for example if one module sets it one way, and the next sets it back, would that break functionality in the former?

aubindetrez:
I started working on a fix for this binary.BinaryValue issue regarding BigEndian.

Each BinaryRepresentation needs to be tested/fixed:
- [x] UNSIGNED
- [ ] SIGNED_MAGNITUDE
- [ ] TWOS_COMPLEMENT

[...]

PR #1286, opened 2020-01-07
Improve BinaryValue performance 

garmin-mjames:
Improving performance related to `BinaryValue` objects.

Relies on #1502 providing a performance regression test. The `matrix_multiplier` test makes heavy use of the `BinaryValue` class.

CI results
----------
Python | master (4eaf429c67739886fab45afd647b4eeb37436694) | PR (d43dcdf359a8ce41d50d51c46fddae470b4baa89) | improvement
------- | --------------- | ----- | --------------
3.8 | [825 NS/S](https://github.com/cocotb/cocotb/runs/1564866751?check_suite_focus=true#step:22:5077) | [1982 NS/S]() | +140%

Outdated results
-----------------

Running `test_matrix_multiplier` locally:

sim | 5b71032 average | 71ff728 average | improvement
--- | ---------- | -------- | ------------- 
icarus | 1583 NS/S | 2413 NS/S | +52% !
aldec (vhdl) | 1204 NS/S | 2046 NS/S | +70% !!

These numbers are outsized due to the `BinaryValue`-focused test, but I think other tests will see at least a small performance improvement.

@andresdemski could you run your test against this?

### DIFF
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -29,27 +29,61 @@
 
 import os
 import random
+import re
 import warnings
 
+
+_RESOLVE_TO_0      = "-lL"
+_RESOLVE_TO_1      = "hH"
+_RESOLVE_TO_CHOICE = "xXzZuUwW"
 resolve_x_to = os.getenv('COCOTB_RESOLVE_X', "VALUE_ERROR")
 
 
-def resolve(string):
-    for char in BinaryValue._resolve_to_0:
-        string = string.replace(char, "0")
-    for char in BinaryValue._resolve_to_1:
-        string = string.replace(char, "1")
-    for char in BinaryValue._resolve_to_error:
-        if resolve_x_to == "VALUE_ERROR" and char in string:
-            raise ValueError("Unable to resolve to binary >%s<" % string)
+class _ResolveTable(dict):
+    """Translation table class for resolving binary strings.
+
+    For use with :func:`str.translate()`, which indexes into table with Unicode ordinals.
+    """
+    def __init__(self):
+        self.update({ord("0"): ord("0"), ord("1"): ord("1")})
+        self.update({ord(k): ord("0") for k in _RESOLVE_TO_0})
+        self.update({ord(k): ord("1") for k in _RESOLVE_TO_1})
+
+        # Do not resolve if resolve_x_to is not set to one of the supported values
+        def no_resolve(key):
+            return key
+        self.resolve_x = no_resolve
+
+        if resolve_x_to == "VALUE_ERROR":
+            def resolve_error(key):
+                raise ValueError("Unresolvable bit in binary string: '{}'".format(chr(key)))
+            self.resolve_x = resolve_error
         elif resolve_x_to == "ZEROS":
-            string = string.replace(char, "0")
+            self.update({ord(k): ord("0") for k in _RESOLVE_TO_CHOICE})
         elif resolve_x_to == "ONES":
-            string = string.replace(char, "1")
+            self.update({ord(k): ord("1") for k in _RESOLVE_TO_CHOICE})
         elif resolve_x_to == "RANDOM":
-            bits = "{0:b}".format(random.getrandbits(1))
-            string = string.replace(char, bits)
-    return string
+            def resolve_random(key):
+                # convert to correct Unicode ordinal:
+                # ord('0') = 48
+                # ord('1') = 49
+                return random.getrandbits(1) + 48
+            self.resolve_x = resolve_random
+
+        self._resolve_to_choice = {ord(c) for c in _RESOLVE_TO_CHOICE}
+
+    def __missing__(self, key):
+        if key in self._resolve_to_choice:
+            return self.resolve_x(key)
+        else:
+            return key
+
+
+_resolve_table = _ResolveTable()
+
+
+def resolve(string):
+    return string.translate(_resolve_table)
 
 
 def _clog2(val):
@@ -89,10 +123,7 @@ class BinaryValue:
     b'*'
 
     """
-    _resolve_to_0     = "-lL"  # noqa
-    _resolve_to_1     = "hH"  # noqa
-    _resolve_to_error = "xXzZuUwW"  # Resolve to a ValueError() since these usually mean something is wrong
-    _permitted_chars  = _resolve_to_0 +_resolve_to_1 + _resolve_to_error + "01"  # noqa
+    _permitted_chars  = _RESOLVE_TO_0 +_RESOLVE_TO_1 + _RESOLVE_TO_CHOICE + "01"  # noqa
 
     def __init__(self, value=None, n_bits=None, bigEndian=True,
                  binaryRepresentation=BinaryRepresentation.UNSIGNED,
@@ -197,10 +228,10 @@ class BinaryValue:
         return binstr
 
     def _convert_from_unsigned(self, x):
-        return int(resolve(x), 2)
+        return int(x.translate(_resolve_table), 2)
 
     def _convert_from_signed_mag(self, x):
-        rv = int(resolve(self._str[1:]), 2)
+        rv = int(self._str[1:].translate(_resolve_table), 2)
         if self._str[0] == '1':
             rv = rv * -1
         return rv
@@ -213,7 +244,7 @@ class BinaryValue:
             if x[0] == '1':
                 rv = rv * -1
         else:
-            rv = int(resolve(x), 2)
+            rv = int(x.translate(_resolve_table), 2)
         return rv
 
     def _invert(self, x):
@@ -312,7 +343,7 @@ class BinaryValue:
     @property
     def signed_integer(self):
         """The signed integer representation of the underlying vector."""
-        ival = int(resolve(self._str), 2)
+        ival = int(self._str.translate(_resolve_table), 2)
         bits = len(self._str)
         signbit = (1 << (bits - 1))
         if (ival & signbit) == 0:
@@ -337,7 +368,7 @@ class BinaryValue:
         This is similar to the SystemVerilog Assertion ``$isunknown`` system function
         or the VHDL function ``is_x`` (with an inverted meaning).
         """
-        return not any(char in self._str for char in BinaryValue._resolve_to_error)
+        return not any(char in self._str for char in _RESOLVE_TO_CHOICE)
 
     @property
     def buff(self) -> bytes:
@@ -351,7 +382,7 @@ class BinaryValue:
             Note that for older versions used with Python 2 these types were
             indistinguishable.
         """
-        bits = resolve(self._str)
+        bits = self._str.translate(_resolve_table)
 
         if len(bits) % 8:
             bits = "0" * (8 - len(bits) % 8) + bits
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -233,8 +233,7 @@ class BinaryValue:
             binstr = x[1:]
             binstr = self._invert(binstr)
             rv = int(binstr, 2) + 1
-            if x[0] == '1':
-                rv = rv * -1
+            rv = rv * -1
         else:
             rv = int(x.translate(_resolve_table), 2)
         return rv
@@ -251,16 +250,10 @@ class BinaryValue:
         BinaryRepresentation.TWOS_COMPLEMENT  : _convert_from_twos_comp,
     }
 
+    _invert_table = str.maketrans({'0': '1', '1': '0'})
+
     def _invert(self, x):
-        inverted = ''
-        for bit in x:
-            if bit == '0':
-                inverted = inverted + '1'
-            elif bit == '1':
-                inverted = inverted + '0'
-            else:
-                inverted = inverted + bit
-        return inverted
+        return x.translate(self._invert_table)
 
     def _adjust_unsigned(self, x):
         if self._n_bits is None:
@@ -404,12 +397,9 @@ class BinaryValue:
 
     @buff.setter
     def buff(self, val: bytes):
-        self._str = ""
-        for char in val:
-            if self.big_endian:
-                self._str += "{0:08b}".format(char)
-            else:
-                self._str = "{0:08b}".format(char) + self._str
+        if not self.big_endian:
+            val = reversed(val)
+        self._str = ''.join([format(char, "08b") for char in val])
         self._adjust()
 
     def _adjust(self):
@@ -436,12 +426,14 @@ class BinaryValue:
         """ The binary representation stored as a string of ``0``, ``1``, and possibly ``x``, ``z``, and other states. """
         return self._str
 
+    _non_permitted_regex = re.compile("[^{}]".format(_permitted_chars))
+
     @binstr.setter
     def binstr(self, string):
-        for char in string:
-            if char not in BinaryValue._permitted_chars:
-                raise ValueError("Attempting to assign character %s to a %s" %
-                                 (char, type(self).__qualname__))
+        match = self._non_permitted_regex.search(string)
+        if match:
+            raise ValueError("Attempting to assign character %s to a %s" %
+                             (match.group(), self.__class__.__name__))
         self._str = string
         self._adjust()
 
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -448,6 +448,9 @@ class BinaryValue:
     get_binstr = binstr.fget
     set_binstr = binstr.fset
 
+    def _set_trusted_binstr(self, string):
+        self._str = string
+
     @property
     def n_bits(self):
         """The number of bits of the binary value."""
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -157,17 +157,9 @@ class BinaryValue:
 
         self._n_bits = n_bits
 
-        self._convert_to = {
-            BinaryRepresentation.UNSIGNED         : self._convert_to_unsigned   ,
-            BinaryRepresentation.SIGNED_MAGNITUDE : self._convert_to_signed_mag ,
-            BinaryRepresentation.TWOS_COMPLEMENT  : self._convert_to_twos_comp  ,
-        }
+        self._convert_to = self._convert_to_map[self.binaryRepresentation].__get__(self, self.__class__)
 
-        self._convert_from = {
-            BinaryRepresentation.UNSIGNED         : self._convert_from_unsigned   ,
-            BinaryRepresentation.SIGNED_MAGNITUDE : self._convert_from_signed_mag ,
-            BinaryRepresentation.TWOS_COMPLEMENT  : self._convert_from_twos_comp  ,
-        }
+        self._convert_from = self._convert_from_map[self.binaryRepresentation].__get__(self, self.__class__)
 
         if value is not None:
             self.assign(value)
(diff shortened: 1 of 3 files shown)


cocotb/handle.py
```diff
@@ -629,7 +657,10 @@ def setimmediatevalue(self, value):
 
     def _getvalue(self):
         binstr = simulator.get_signal_val_binstr(self._handle)
-        result = BinaryValue(binstr, len(binstr))
+        # Skip BinaryValue.assign() as we know we are using a binstr
+        result = BinaryValue(n_bits=len(binstr))
+        # Skip the permitted characters check as we trust the simulator
+        result.set_trusted_binstr(binstr)
```
garmin-mjames: This assumes that we can always trust the string the simulator returns from `get_signal_val_binstr()` only contains permitted characters.
The speed-up was quite significant, so I hope that's a valid assumption.

cmarqu:
Could you add a short section for ``matrix_multiplier`` to https://github.com/cocotb/cocotb/blob/master/documentation/source/examples.rst?
I made an effort to at least mention all examples in the docs a while ago.

cocotb/binary.py
```diff
@@ -32,27 +32,59 @@
 
 import os
 import random
+import re
 import warnings
 
+
+RESOLVE_TO_0     = "-lL"
+RESOLVE_TO_1     = "hH"
+RESOLVE_TO_ERROR = "xXzZuUwW"  # Resolve to a ValueError() since these usually mean something is wrong
```
eric-wieser: What's the motivation for moving these to global public constants?

cocotb/handle.py
```diff
@@ -442,12 +476,6 @@ def __ne__(self, other):
             return SimHandleBase.__ne__(self, other)
         return self.value != other
 
-    # We want to maintain compatibility with python 2.5 so we can't use @property with a setter
-    value = property(fget=lambda self: self._getvalue(),
-                     fset=lambda self, v: self._setcachedvalue(v),
-                     fdel=None,
-                     doc="A reference to the value")
```
eric-wieser: Your metaclass above feels like crazy overkill - One simple option would be:
```python
@property
def value(self):
    return self._getvalue()
@value.setter
def value(self, v):
    self._setcachedvalue(v)
```

If you're trying to optimize this further, then there are still tidier ways to do it - I'll put up an alternate patch for this path

cocotb/binary.py
```diff
@@ -284,11 +320,11 @@ def _adjust_twos_comp(self, x):
 
     def get_value(self):
         """Return the integer representation of the underlying vector."""
-        return self._convert_from[self.binaryRepresentation](self._str)
```
eric-wieser: This change means that users can no longer change `.binaryRepresentation` after the object is created. I don't know whether that was ever supported.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    for i in range(num_samples):
+        yield [[getrandbits(DATA_WIDTH) for col in range(A_COLUMNS_B_ROWS)] for row in range(A_ROWS)]
+
+
+def gen_b(num_samples=NUM_SAMPLES):
+    """Generate random matrix data for B
+    """
+    for i in range(num_samples):
+        yield [[getrandbits(DATA_WIDTH) for col in range(B_COLUMNS)] for row in range(A_COLUMNS_B_ROWS)]
```
eric-wieser: Can you extract a helper here:
```python
def gen_random_matrix(rows, cols):
    return [[getrandbits(DATA_WIDTH) for col in range(cols)] for row in range(rows)]
```

---

It's then _maybe_ tempting to use it as
```python
factory.add_option('a_data', [
    lambda: (gen_random_matrix(A_ROWS, A_COLUMNS_B_ROWS) for i in range(NUM_SAMPLES))
])
factory.add_option('b_data', [
    lambda: (gen_random_matrix(A_COLUMNS_B_ROWS, B_COLUMNS) for i in range(NUM_SAMPLES))
])
```

cocotb/binary.py
garmin-mjames: I moved them so that they were defined before both `Resolver` and `BinaryValue` classes.
`BinaryValue` calls `resolver.resolve()` so it was a chicken/egg problem.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,185 @@
...
+
+            if self.bus.i_valid.value:
+                if self.multi_dimension:
+                    # Store BinaryValue objects
+                    in_values = (self.bus.A.value, self.bus.B.value)
+                else:
+                    a_matrix = [[self.bus.A[(i * A_COLUMNS_B_ROWS) + j].value for j in range(A_COLUMNS_B_ROWS)] for i in range(A_ROWS)]
+                    b_matrix = [[self.bus.B[(i * B_COLUMNS) + j].value for j in range(B_COLUMNS)] for i in range(A_COLUMNS_B_ROWS)]
+
+                    in_values = (a_matrix, b_matrix)
```
eric-wieser: Thsi becomes far more readable with a helper function:
```python
def _unpack_matrix(self, signal, rows, cols):
    if self.multi_dimension:
        return [[signal[(i * cols) + j].value for j in range(cols)] for i in range(row)]
    else:
        return signal.value
```
used as:
```python
    if self.bus.o_valid.value and in_values:
        self._recv((in_values[0], in_values[1], self._unpack_matrix(self.bus.C, A_ROWS, B_COLUMNS)))
        in_values = None

    if self.bus.i_valid.value:
        # Store BinaryValue objects
        in_values = (
            self._unpack_matrix(self.bus.A, A_ROWS, A_COLUMNS_B_ROWS),
            self._unpack_matrix(self.bus.B, A_COLUMNS_B_ROWS, B_COLUMNS),
        )
```

cocotb/binary.py
eric-wieser: It's not a chicken and egg problem though - `resolver.resolve` can use `BinaryValue._resolve_to_1` without any issue.

If you think they're clearer at the top of the file, then please rename them to start with an underscore (ie, to be private), and remove the `BinaryValue` attributes completely.

cocotb/binary.py
```diff
@@ -32,27 +32,59 @@
...
+                return self.resolve_x()
+            except AttributeError:
+                return match.group(0)
+
+    def resolve(self, string):
+        """Pass-through function to :meth:`re.RegexObject.sub()`
+        """
+        return self._regex.sub(self, string)
 
+resolver = Resolver()
```
eric-wieser: Ah, now I see what you mean about the chicken and egg problem. You could put this on the last line of the file and it would work. `_RESOLVE_TO_ERROR` is fine by me though.

cocotb/binary.py
```diff
@@ -32,27 +32,59 @@
...
+            self.resolve_x = resolve_x
         elif resolve_x_to == "ZEROS":
-            string = string.replace(char, "0")
+            self.update({k: "0" for k in RESOLVE_TO_ERROR})
         elif resolve_x_to == "ONES":
-            string = string.replace(char, "1")
+            self.update({k: "1" for k in RESOLVE_TO_ERROR})
         elif resolve_x_to == "RANDOM":
-            bits = "{0:b}".format(random.getrandbits(1))
-            string = string.replace(char, bits)
```
eric-wieser: Woa, this `RANDOM` thing is nonsense - it resolves the Xs to either all ones or all zeros!

cocotb/binary.py
```diff
@@ -32,27 +32,59 @@
...
-        string = string.replace(char, "1")
-    for char in BinaryValue._resolve_to_error:
-        if resolve_x_to == "VALUE_ERROR" and char in string:
-            raise ValueError("Unable to resolve to binary >%s<" % string)
+class Resolver(dict):
+    """One-pass resolver for binary strings.
+
+    Sub-class :class:`dict` and use the key-value mapping for bit resolution.
+
+    Utilizes :meth:`re.sub()` to resolve all characters in one pass.
```
eric-wieser: You might want to look at using `string.maketrans` which will likely be even faster

examples/matrix_multiplier/tests/test_matrix_multiplier.py
garmin-mjames: Agreed, thanks.

cocotb/binary.py
garmin-mjames: I noticed that, and I think resolving each bit is better.

cocotb/binary.py
```diff
@@ -32,27 +32,59 @@
...
+    def __call__(self, match):
+        """Called by :meth:`re.sub()` on each match of a bit to be resolved.
+        """
+        try:
+            return self[match.group(0)]
+        except KeyError:
+            try:
+                return self.resolve_x()
+            except AttributeError:
+                return match.group(0)
```
eric-wieser: Likely faster would be:
```
    def __missing__(self, key):
        try:
             return self.resolve_x()
        except AttributeError:
            return key
    def __call__(self, match):
        return self[match.group(0)]
```

garmin-mjames: I also tried:
```
self.resolve = partial(self._regex.sub, self)
```
And saw a very small speedup, but it was lost in the variability of the ratio from run-to-run.

eric-wieser: I would recommend timing this type of thing locally in ipython, rather than as part of a full test suite.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
garmin-mjames: The only downside to the lambdas is we lose the docstrings:
```
     0.00ns INFO     cocotb.test._my_test.0x7feabcca0860       decorators.py:253  in _advance                        Starting test: "run_test_array_001"
                                                                                                                     Description: Automatically generated test

                                                                                                                       a_data: <lambda> (No docstring supplied)
                                                                                                                       b_data: <lambda> (No docstring supplied)
```

eric-wieser: Good point - them maybe use the helper function, but use it from within `def gen_a` and `def gen_b`

cocotb/binary.py
ktbarrett: It was. Changing `binaryRepresentation` after creation would result in a different result for `get_value` and all the `conversion` functions from that point forward. That being said, it was probably never a good idea, and it isn't documented.

cocotb/handle.py
eric-wieser: See gh-1287 for a less complex solution

cocotb/binary.py
garmin-mjames: After investigating using `str.translate()`, it's much faster but it uses the translation value for all matching characters, so all `x` or `Z` become the same number.
The effect is the current behavior, where `xxxx` becomes `1111` or `0000`.

I can cleanup the `re.sub()` implementation if we really want per-bit random values.

eric-wieser: Since this patch is about performance, we should err on the side of keeping the current behavior.

In a future pass we could do a second pass to randomize X's.

cocotb/handle.py
garmin-mjames: Removed the commit in favor of your PR.

garmin-mjames:
@cmarqu Added documentation of example.
@ktbarrett Fixed a few of the performance issues you pointed out.

Generally, the BinaryValue class has some serious problems with its internal representation and also big vs little endian. See issue #608 and PR #1020.

I stuck with performance improvements while trying to match current behavior.

cocotb/binary.py
```diff
@@ -32,28 +32,61 @@
...
-    for char in BinaryValue._resolve_to_1:
-        string = string.replace(char, "1")
-    for char in BinaryValue._resolve_to_error:
-        if resolve_x_to == "VALUE_ERROR" and char in string:
-            raise ValueError("Unable to resolve to binary >%s<" % string)
+
+class ResolveTable(dict):
+    """Translation table class for resolving binary strings.
+
+    For use with :func:`str.translate()`, which indexes into table with Unicode ordinals.
```
garmin-mjames: I think this is only Python 3 compatible as-is (assumes Unicode).

ktbarrett: With #1289 already filed, it seems Eric is eager to rid us of Python 2. So I wouldn't worry about it.

cocotb/binary.py
```diff
@@ -32,28 +32,61 @@
...
+    """
+    def __init__(self):
+        self.update({ord("0"): ord("0"), ord("1"): ord("1")})
+        self.update({ord(k): ord("0") for k in _RESOLVE_TO_0})
+        self.update({ord(k): ord("1") for k in _RESOLVE_TO_1})
```
eric-wieser: No need for `ord` I think

cocotb/binary.py
```diff
@@ -32,28 +32,61 @@
...
+                # ord('0') = 48
+                # ord('1') = 49
+                return random.getrandbits(1) + 48
+            self.resolve_x = resolve_random
+
+        self._resolve_to_choice = {ord(c) for c in _RESOLVE_TO_CHOICE}
+
+    def __missing__(self, key):
+        if key in self._resolve_to_choice:
+            return self.resolve_x(key)
```
eric-wieser: This still won't actually do what you're expecting it to, `str.translate` caches the calls to `__missing__`.

cocotb/binary.py
garmin-mjames: `str.translate()` indexes into table with an ordinal, but you can return ordinals or strings.
This is faster than converting back to a string: 
```
return random.getrandbits(1) + 48
```
so I return ordinals for everything.
Also `str.maketrans()` produces a `dict` with ordinal->ordinal mapping.

cocotb/binary.py
garmin-mjames: All instances of each character to be translated will be the same, either `0` or `1`, which is the same as current behavior. I noted this in the commit message.

cocotb/binary.py
ktbarrett: #1291

garmin-mjames:
@eric-wieser As implemented, this keeps the current `resolve` behavior (see #1487).
Should I go back to a `re.sub` approach to get true per-bit resolving? It's slightly slower than `str.translate` but much faster than the current approach.

eric-wieser:
Can you put the new example in a standalone PR first?

garmin-mjames:
We should probably resolve #1487 in this PR as well.

ktbarrett:
@garmin-mjames The maintainers decided to deprecate the BinaryValue class and replace it (#2059). We are still deciding on how to go about the whole deprecation process of BinaryValue; and will come to that decision before reviewing/merging this PR so we aren't wasting our or your effort on something that may be immediately dropped. I hope you understand, and please feel free to comment on #2059!

ktbarrett:
@garmin-mjames Are you still interested in getting this in for 1.5? Seems like all the prereq PRs are merged.

garmin-mjames:
Yes. I'll rebase and update this in the coming week.

cocotb/handle.py
```diff
@@ -718,7 +718,10 @@ def _check_for_set_action(self, value):
     @NonConstantObject.value.getter
     def value(self) -> BinaryValue:
         binstr = self._handle.get_signal_val_binstr()
-        result = BinaryValue(binstr, len(binstr))
+        # Skip BinaryValue.assign() as we know we are using a binstr
+        result = BinaryValue(n_bits=len(binstr))
```
themperek, line 722: Any idea if we create BinaryValue once and return a copy (https://docs.python.org/3.5/library/copy.html) ? Not to `__init__`?

garmin-mjames: Would that save any time? This already bypasses the slow path of `assign() -> binstr()`, which is most of the time taken up by `BinaryValue.__init__()` when you pass in a value.

It might be slightly faster to change `set_trusted_binstr()` to a classmethod and skip the deprecation checks in `__init__()`.

themperek: You are probably be right, it is `assign`.

cocotb/binary.py
```diff
@@ -29,27 +29,61 @@
...
 
-def resolve(string):
-    for char in BinaryValue._resolve_to_0:
-        string = string.replace(char, "0")
-    for char in BinaryValue._resolve_to_1:
-        string = string.replace(char, "1")
-    for char in BinaryValue._resolve_to_error:
-        if resolve_x_to == "VALUE_ERROR" and char in string:
-            raise ValueError("Unable to resolve to binary >%s<" % string)
+class ResolveTable(dict):
```
ktbarrett, line 42:
```suggestion
class _ResolveTable(dict):
```

cocotb/binary.py
```diff
@@ -29,27 +29,61 @@
...
+        self._resolve_to_choice = {ord(c) for c in _RESOLVE_TO_CHOICE}
+
+    def __missing__(self, key):
+        if key in self._resolve_to_choice:
+            return self.resolve_x(key)
+        else:
+            return key
+
+
+_resolve_table = ResolveTable()
```
ktbarrett, line 82:
```suggestion
_resolve_table = _ResolveTable()
```

cocotb/binary.py
```diff
@@ -401,18 +426,24 @@ def binstr(self):
...
+        if match:
+            raise ValueError("Attempting to assign character %s to a %s" %
+                             (match.group(), self.__class__.__name__))
         self._str = string
         self._adjust()
 
     get_binstr = binstr.fget
     set_binstr = binstr.fset
 
+    def set_trusted_binstr(self, string):
```
ktbarrett, line 443:
```suggestion
    def _set_trusted_binstr(self, string):
        """set value of a BinaryValue using a str without doing sanity checking"""
```

cocotb/binary.py
```diff
@@ -401,18 +426,24 @@ def binstr(self):
...
+        self._str = string
+        self._adjust()
```
ktbarrett, line 445: If we trust the string do we need this?
```suggestion
```

cocotb/handle.py
ktbarrett:
```suggestion
        # Skip the permitted characters check as we trust the simulator
        result._set_trusted_binstr(binstr)
```

cocotb/binary.py
garmin-mjames: good point.

cocotb/binary.py
```diff
@@ -29,27 +29,61 @@
 
 import os
 import random
+import re
 import warnings
 
+
+_RESOLVE_TO_0      = "-lL"
```
ktbarrett, line 36: I know this was just moved, but resolving `-` to `0` is wrong. "Don't care" is a synonym for `X`. Something to note in the future.

ktbarrett (approved):

cocotb/binary.py
garmin-mjames: #1487 with your comment is still open.

garmin-mjames:
Does this warrant a newsfragment? Behavior is identical, just faster.

ktbarrett:
@garmin-mjames I don't think so. I think it's g2g.

garmin-mjames:
Just updated commit messages.

Issue #1289, opened 2020-01-09
Python 2 Cleanup tracking issue


eric-wieser:
A list of cleanups and improvements we can make when dropping python 2.

Relates to #767. Maintainers, please add items to this list / edit the list to link PRs.

### Easy changes:

If you're a new contributor, these are a great place to start.

* [x] Adjust (#1337) https://github.com/cocotb/cocotb/blob/9caf77cb6ceae739858ecd4a3760631470aac77c/setup.py#L59
* [x] Change minimum supported version in documentation (gh-1341)
* [x] Remove 2.7 version from Travis CI setup
* [x] Remove reference from `tox.ini`
* [x] Remove function ``isclose`` from ``tests/test_cases/test_cocotb/test_cocotb.py`` (#1489)
* [x] Combine `cocotb_test` and `cocotb_test_35` testbenches (#1489)
* [x] Remove `from __future__ import` line from Python files (but not `Makefile.pylib.*`) (gh-1579)
* [x] Remove `(object)` from `class MyClass(object):` (gh-1579)
* [x] use `type(self)` instead of `self.__class__`. (gh-1579)
* [x] Use `__class__.__qualname__` instead of `__class__.__name__`
* [x] Remove all `__nonzero__` functions which alias `__bool__` (gh-1579)
* [x] Remove 2x "# NOTE: isclose is a Python 3.5+ feature" from ``documentation/source/coroutines.rst`` (gh-1576)
* [x] Update/check installation instructions in documentation and READM
* [x] Remove the conditional compilation here: https://github.com/cocotb/cocotb/blob/ad1e718eab6a107b56c47032f53f282f7d2dbae0/examples/endian_swapper/cosim/io.c#L116. 
* Marginal:
  * [x] make ``u""`` strings normal strings (e.g. in ``documentation/source/conf.py``) (gh-1579)
  * [ ] Move to format string language instead of using old modulus string formatting.

### More involved

* [x] Delete https://github.com/cocotb/cocotb/blob/master/cocotb/share/lib/simulator/simulatormodule_python2.c, and merge https://github.com/cocotb/cocotb/blob/master/cocotb/share/lib/simulator/simulatormodule_python3.c with https://github.com/cocotb/cocotb/blob/master/cocotb/share/lib/simulator/simulatormodule.c (#1342)
* [x] Removes internal uses of `cocotb./_py_compat` (#1434)
[...]

PR #1291, opened 2020-01-10
Remove Python 2 from Travis

ktbarrett:
Had to move some yaml refs over so the diff isn't as clean as you would expect. #1289.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,21 +26,12 @@ jobs:
 
     - stage: SimTests
       os: linux
-      python: 2.7
+      python: 3.5
       before_install: &docker_prep
         - docker build -t cocotb .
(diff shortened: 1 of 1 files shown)


eric-wieser:
Let's wait for signoff from @imphil on dropping support for 2.7 in master immediately

ktbarrett:
@themperek Wouldn't you want them to run at the same time? I can agree that `PythonLint` is no longer the right name for them though.

ktbarrett:
@eric-wieser If you are worried about disabling the Python2 tests I can add them back in and set them to allow failure.

eric-wieser:
We agreed in the call today to drop python 2.7 from master - so lets put this in

Issue #1225, opened 2019-12-04
Allow a custom entry point


eric-wieser:
Currently, the entry point from C to python is hard-coded here:

https://github.com/cocotb/cocotb/blob/17ce4e51da43b6d2d1b5834d629374ed5535a8d8/cocotb/share/lib/embed/gpi_embed.c#L398

Which is

https://github.com/cocotb/cocotb/blob/ba0fb097efbd2939ad6a961ee8d02793104ef8f7/cocotb/__init__.py#L108

From about here, the things it does become opinionated:

https://github.com/cocotb/cocotb/blob/ba0fb097efbd2939ad6a961ee8d02793104ef8f7/cocotb/__init__.py#L154-L172

It would be great if a user could set an env var or plusarg to launch into a different entry point

ktbarrett:
Plusargs don't work with Questa. Don't know why anyone didn't think of that when they added the feature. I'm not a fan of environment variables when we could use a Python module (or modules).

This would require some modification of the scheduler and regression manager. They both reach into eachother by way of global scheduler and regression manager variables. Custom code might not use either.
https://github.com/cocotb/cocotb/blob/17ce4e51da43b6d2d1b5834d629374ed5535a8d8/cocotb/__init__.py#L89-L92

eric-wieser:
>  I'm not a fan of environment variables when we could use a Python module (or modules).

There are two approaches to this problem I've seen in the wild:

* `conftest.py` (pytest), `conf.py` (Sphinx) - this isn't really an entry point, but is an example of settings being loaded from a python file. Note that I'm pretty sure this isn't imported the usual way, but is read from the file system and then exec'd, which gives more flexible rules about where it can live.
* a `--entry-point=my.python_module:my_func` syntax, which is how other runtime environments (eg the python WSGI web server hosts, `cherrypy`, `gunicorn`, etc) let you choose an entry point. The dot and colon syntax is a `setuptools` protocol, so seems sensible to reuse.

My preference is towards the latter, because a the world with arbitrary entry points (eg IPython), cocotb feels more like a runtime environment than a configurable test suite.

ktbarrett:
>I think @ktbarrett's comments above are in favor of something like the former.

Correct. I don't see a reason that normal imports can't be used. After looking at your examples, I like pytest's approach of calling well defined hooks, but instead of hooks overrides. The overrides can call the default implementation if they just want to hook behavior.

[...]

PR #1293, opened 2020-01-12
Allow users to load into any Python code from GPI

ktbarrett:
Adds `PYTHON_ENTRY` environment variable for users to enter into a custom environment rather than loading into the `cocotb` Python module as described in #1225. Also tried to shrink the interface between `gpi_embed` and the `cocotb` module by pushing much of what `embed_sim_init` into Python, and communicating the rest by passing by argument to the entry point.

This is best reviewed commit by commit.

### Interface
 - `_sim_event` in entry module
 - `_log_from_c` in entry module
 - `_filter_from_c` in entry module
 - entry function in entry module
  - takes `argc`, `argv`, and simulator details

### TODO
 - [x] document `COCOTB_ENTRY`
 - [x] document custom entry point's interface requirements
 - [x] more thorough `COCOTB_ENTRY` spec checking
 - [x] cleanup `gpi_embed` to reduce interface requirements on custom module
 - [x] Add example

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -56,6 +56,7 @@ from ._version import __version__
 
 # GPI logging instance
 if "COCOTB_SIM" in os.environ:
+    import simulator
 
     def _reopen_stream_with_buffering(stream_name):
         try:
@@ -134,9 +135,14 @@ def _initialise_testbench(argv_):
     """
     _rlock.acquire()
 
-    global argc, argv
+    global argc, argv, SIM_NAME, SIM_VERSION
     argv = argv_
     argc = len(argv)
+    SIM_NAME = simulator.product()
+    SIM_VERSION = simulator.version()
+
+    cocotb.log.info("Running on {} version {}".format(SIM_NAME, SIM_VERSION))
+    cocotb.log.info("Python interpreter initialized and cocotb loaded!")
 
     root_name = os.getenv("TOPLEVEL", None)
     if root_name is not None:
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -112,6 +112,8 @@ fork = scheduler.add
 # FIXME is this really required?
 _rlock = threading.RLock()
 
+LANGUAGE = os.getenv("TOPLEVEL_LANG")
+
 
 def mem_debug(port):
     import cocotb.memdebug
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -120,7 +120,7 @@ def mem_debug(port):
     cocotb.memdebug.start(port)
 
 
-def _initialise_testbench(root_name):
+def _initialise_testbench():
     """Initialize testbench.
 
     This function is called after the simulator has elaborated all
@@ -133,6 +133,13 @@ def _initialise_testbench(root_name):
     comma-separated list of modules to be executed before the first test.
     """
     _rlock.acquire()
+
+    root_name = os.getenv("TOPLEVEL", None)
+    if root_name is not None:
+        if root_name == "":
+            root_name = None
+        elif '.' in root_name:
+            root_name = root_name.split(".", 1)[1]
 
     memcheck_port = os.getenv('MEMCHECK')
     if memcheck_port is not None:
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -120,7 +120,7 @@ def mem_debug(port):
     cocotb.memdebug.start(port)
 
 
-def _initialise_testbench():
+def _initialise_testbench(argv_):
     """Initialize testbench.
 
     This function is called after the simulator has elaborated all
@@ -133,6 +133,10 @@ def _initialise_testbench():
     comma-separated list of modules to be executed before the first test.
     """
     _rlock.acquire()
+
+    global argc, argv
+    argv = argv_
+    argc = len(argv)
 
     root_name = os.getenv("TOPLEVEL", None)
     if root_name is not None:
--- a/cocotb/_gpi_embed.py
+++ b/cocotb/_gpi_embed.py
@@ -1,0 +1,22 @@
+import logging
+
+
+def _filter_from_c(logger_name, level):
+    return logging.getLogger(logger_name).isEnabledFor(level)
+
+
+def _log_from_c(logger_name, level, filename, lineno, msg, function_name):
+    """Log from the C world, allowing to insert C stack information."""
+    logger = logging.getLogger(logger_name)
+    if logger.isEnabledFor(level):
+        record = logger.makeRecord(
+            logger.name,
+            level,
+            filename,
+            lineno,
+            msg,
+            None,
+            None,
+            function_name
+        )
+        logger.handle(record)
--- a/cocotb/_gpi_embed.py
+++ b/cocotb/_gpi_embed.py
@@ -1,4 +1,7 @@
 import logging
+import os
+import importlib
+from functools import reduce
 
 
 def _filter_from_c(logger_name, level):
@@ -20,3 +23,19 @@ def _log_from_c(logger_name, level, filename, lineno, msg, function_name):
             function_name
         )
         logger.handle(record)
+
+
+def _load_entry():
+    """Gather entry point information by parsing :envar:`COCOTB_ENTRY_POINT`."""
+    entry_point_str = os.environ.get("COCOTB_ENTRY", "cocotb:_initialise_testbench")
+    try:
+        if ":" not in entry_point_str:
+            raise ValueError("Invalid COCOTB_ENTRY, missing entry function (no colon).")
+        entry_module_str, entry_func_str = entry_point_str.split(":", 1)
+        entry_module = importlib.import_module(entry_module_str)
+        entry_func = reduce(getattr, entry_func_str.split('.'), entry_module)
+    except Exception as e:
+        raise RuntimeError("Failure to parse COCOTB_ENTRY ('{}')".format(entry_point_str)) from e
+    entry_sim_event = getattr(entry_module, "_sim_event")
+    assert callable(entry_sim_event)  # won't be called immediately, so we check now for time's sake
+    return (entry_module, entry_func, entry_sim_event)
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -232,7 +232,7 @@ int embed_sim_init(int argc, char const* const* argv)
         return ret;
 
     PyObject *cocotb_module, *cocotb_init, *cocotb_retval;
-    PyObject *cocotb_log_module = NULL;
+    PyObject *cocotb_gpi_module = NULL;
     PyObject *simlog_func;
     PyObject *argv_list;
 
@@ -245,12 +245,12 @@ int embed_sim_init(int argc, char const* const* argv)
     if (get_module_ref("cocotb", &cocotb_module))
         goto cleanup;
 
-    if (get_module_ref("cocotb.log", &cocotb_log_module)) {
+    if (get_module_ref("cocotb._gpi_embed", &cocotb_gpi_module)) {
         goto cleanup;
     }
 
     // Obtain the function to use when logging from C code
-    simlog_func = PyObject_GetAttrString(cocotb_log_module, "_log_from_c");      // New reference
+    simlog_func = PyObject_GetAttrString(cocotb_gpi_module, "_log_from_c");      // New reference
     if (simlog_func == NULL) {
         PyErr_Print();
         LOG_ERROR("Failed to get the _log_from_c function");
@@ -265,7 +265,7 @@ int embed_sim_init(int argc, char const* const* argv)
     set_log_handler(simlog_func);                                       // Note: This function steals a reference to simlog_func.
 
     // Obtain the function to check whether to call log function
-    simlog_func = PyObject_GetAttrString(cocotb_log_module, "_filter_from_c");   // New reference
+    simlog_func = PyObject_GetAttrString(cocotb_gpi_module, "_filter_from_c");   // New reference
     if (simlog_func == NULL) {
         PyErr_Print();
         LOG_ERROR("Failed to get the _filter_from_c method");
@@ -344,7 +344,7 @@ cleanup:
     ret = -1;
 ok:
     Py_XDECREF(cocotb_module);
-    Py_XDECREF(cocotb_log_module);
+    Py_XDECREF(cocotb_gpi_module);
 
     PyGILState_Release(gstate);
     to_simulator();
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -242,8 +242,9 @@ int embed_sim_init(int argc, char const* const* argv)
     PyGILState_STATE gstate = PyGILState_Ensure();
     to_python();
 
-    if (get_module_ref("cocotb", &cocotb_module))
+    if (get_module_ref("cocotb", &cocotb_module)) {
         goto cleanup;
+    }
 
     if (get_module_ref("cocotb._gpi_embed", &cocotb_gpi_module)) {
         goto cleanup;
@@ -256,11 +257,6 @@ int embed_sim_init(int argc, char const* const* argv)
         LOG_ERROR("Failed to get the _log_from_c function");
         goto cleanup;
     }
-    if (!PyCallable_Check(simlog_func)) {
-        LOG_ERROR("_log_from_c is not callable");
-        Py_DECREF(simlog_func);
-        goto cleanup;
-    }
 
     set_log_handler(simlog_func);                                       // Note: This function steals a reference to simlog_func.
 
@@ -269,11 +265,6 @@ int embed_sim_init(int argc, char const* const* argv)
     if (simlog_func == NULL) {
         PyErr_Print();
         LOG_ERROR("Failed to get the _filter_from_c method");
-        goto cleanup;
-    }
-    if (!PyCallable_Check(simlog_func)) {
-        LOG_ERROR("_filter_from_c is not callable");
-        Py_DECREF(simlog_func);
         goto cleanup;
     }
 
@@ -285,22 +276,11 @@ int embed_sim_init(int argc, char const* const* argv)
         LOG_ERROR("Failed to get the _sim_event method");
         goto cleanup;
     }
-    if (!PyCallable_Check(pEventFn)) {
-        LOG_ERROR("cocotb._sim_event is not callable");
-        Py_DECREF(pEventFn);
-        pEventFn = NULL;
-        goto cleanup;
-    }
 
     cocotb_init = PyObject_GetAttrString(cocotb_module, "_initialise_testbench");   // New reference
     if (cocotb_init == NULL) {
         PyErr_Print();
         LOG_ERROR("Failed to get the _initialise_testbench method");
-        goto cleanup;
-    }
-    if (!PyCallable_Check(cocotb_init)) {
-        LOG_ERROR("cocotb._initialise_testbench is not callable");
-        Py_DECREF(cocotb_init);
         goto cleanup;
     }
 
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -242,10 +242,6 @@ int embed_sim_init(int argc, char const* const* argv)
     PyGILState_STATE gstate = PyGILState_Ensure();
     to_python();
 
-    if (get_module_ref("cocotb", &cocotb_module)) {
-        goto cleanup;
-    }
-
     if (get_module_ref("cocotb._gpi_embed", &cocotb_gpi_module)) {
         goto cleanup;
     }
@@ -270,19 +266,22 @@ int embed_sim_init(int argc, char const* const* argv)
 
     set_log_filter(simlog_func);                                        // Note: This function steals a reference to simlog_func.
 
-    pEventFn = PyObject_GetAttrString(cocotb_module, "_sim_event");     // New reference
-    if (pEventFn == NULL) {
+    PyObject* entry_tuple = PyObject_CallMethod(cocotb_gpi_module, "_load_entry", NULL);
+    if (entry_tuple == NULL) {
         PyErr_Print();
-        LOG_ERROR("Failed to get the _sim_event method");
+        LOG_ERROR("Unable to load entry point");
         goto cleanup;
     }
-
-    cocotb_init = PyObject_GetAttrString(cocotb_module, "_initialise_testbench");   // New reference
-    if (cocotb_init == NULL) {
+    if (!PyArg_ParseTuple(entry_tuple, "OOO", &cocotb_module, &cocotb_init, &pEventFn)) {
         PyErr_Print();
-        LOG_ERROR("Failed to get the _initialise_testbench method");
+        LOG_ERROR("Bad output from cocotb._gpi_embed._load_entry: should return (entry module, entry function, sim event callback)");
+        Py_DECREF(entry_tuple);
         goto cleanup;
     }
+    Py_INCREF(cocotb_module);
+    Py_INCREF(cocotb_init);
+    Py_INCREF(pEventFn);
+    Py_DECREF(entry_tuple);
 
     // Build argv for cocotb module
     argv_list = PyList_New(argc);                                       // New reference
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -219,7 +219,8 @@ int get_module_ref(const char *modname, PyObject **mod)
     return 0;
 }
 
-int embed_sim_init(gpi_sim_info_t *info)
+
+int embed_sim_init(int argc, char const* const* argv)
 {
     FENTER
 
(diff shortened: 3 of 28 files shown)


ktbarrett:
I definitely want some notes on documentation. @cmarqu I know I need to probably improve the docstrings on the linked functions.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -388,7 +339,36 @@ int embed_sim_init(gpi_sim_info_t *info)
...
+            Py_DECREF(argv_list);
+            goto cleanup;
+        }
+        PyList_SET_ITEM(argv_list, i, argv_item);                       // Note: This function steals the reference to argv_item
+    }
+
+    LOG_INFO("Running on %s version %s", info->product, info->version);
+
+    char spec[5];
+    strcpy(spec, "iOss");
```
garmin-mjames: Donn't need this `bpo-9369` work-around for Python 3.

documentation/source/python_entry_point.rst
```diff
@@ -0,0 +1,28 @@
+Python Entry Point
+==================
+
+By default, cocotb starts the :class:`RegressionManager` as the testing framework for cocotb. However, there are many reason why a user might want to supplant the whole or part of the existing Python framework with an alternative. To this end, we have defined :envvar:`PYTHON_ENTRY`, to allow the user to enter into an alternative testing framework.
```
garmin-mjames:
```suggestion
By default, cocotb starts the :class:`RegressionManager` as the testing framework for cocotb. However, there are many reasons why a user might want to supplant the whole or part of the existing Python framework with an alternative. To this end, we have defined :envvar:`PYTHON_ENTRY`, to allow the user to enter into an alternative testing framework.
```

cmarqu:
We have prefixed all newly introduced environment variables with ``COCOTB_``, so I would suggest the same here.
Also, for markup languages like ReST where single line breaks do not matter in rendered output, I would put in line breaks at least after every period so that later diffs are small. (https://rhodesmill.org/brandon/2012/one-sentence-per-line/)

documentation/source/python_entry_point.rst
cmarqu:
```suggestion
By default, cocotb starts the :class:`~cocotb.regression.RegressionManager` as the testing framework for cocotb. 
However, there are many reasons why a user might want to supplant the whole or part of the existing Python framework with an alternative.
To this end, we have defined :envvar:`COCOTB_PYTHON_ENTRY`, to allow the user to enter into an alternative testing framework.
```

documentation/source/python_entry_point.rst
```diff
@@ -0,0 +1,28 @@
...
+
+PyGPI expects the following functions to be implemented in the entry module. The names are required.
+
+.. autofunction:: _sim_event
+
+.. autofunction:: _log_from_c
+
+.. autofunction:: _filter_from_c
+
+Finally, there is the entry function itself. ``_initialise_testbench`` is being used as an example, the entry function is not required to be named the same. A single entry module can even contain multiple entry functions.
```
cmarqu: Maybe it's just me, but could someone read this last sentence as "A single-entry module can even contain multiple entry functions." and be confused for a second?
Would "One entry module can even contain multiple entry functions." circumvent that and be still correct?

documentation/source/python_entry_point.rst
```diff
@@ -0,0 +1,28 @@
...
+==================
+
+By default, cocotb starts the :class:`RegressionManager` as the testing framework for cocotb. However, there are many reason why a user might want to supplant the whole or part of the existing Python framework with an alternative. To this end, we have defined :envvar:`PYTHON_ENTRY`, to allow the user to enter into an alternative testing framework.
+
+.. versionadded:: 2.0
+
+How It Works
+------------
+
+After GPI implementations (VPI, VHPI, FLI, etc.) are registered, the Python implementation to the GPI (henceforth, PyGPI) starts a Python interpreter, loads the ``cocotb`` module, gets references to necessary helper functions from that module, and then calls into the ``_initialise_testbench`` function in that module to start the regression manager. The :envvar:`PYTHON_ENTRY` variable allows the user to specify a module and function other than ``cocotb._initialise_testbench`` that contains the alternative testing framework.
```
cmarqu: We have slowly americanized the spelling also of functions etc. as we touched them, so maybe better ``_initialize_testbench``.

ktbarrett: This is a name of an existing function. Should the function be renamed in concert?

documentation/source/python_entry_point.rst
garmin-mjames: I agree it could be ambiguous. What about "**The** entry module can even contain multiple entry functions."

ktbarrett:
@cmarqu Seeing as this has nothing to do with the Python module `cocotb`, I'm not sure if starting it with `COCOTB_` is a good idea. I think we should probably parse out who is responsible for each environment variable a little better: build (like `SIM`), PyGPI (like `PYTHON_ENTRY`), RegressionManager and the start up stuff in the `cocotb` module (like `TESTCASE`), other modules (like `COCOTB_ANSI_OUTPUT`), and make-based simulator run arguments (`SIM_ARGS` and friends).

cmarqu:
I would still see the environment variable as belonging to the cocotb ecosystem. When I would encounter ``PYTHON_ENTRY`` without context (e.g. when it has been exported by some project setup tool - not an inconceivable use case IMO) I would assume it somehow belongs to Python itself.

documentation/source/python_entry_point.rst
cmarqu: Indeed, that sounds better.

documentation/source/python_entry_point.rst
cmarqu: Existing functions should probably stay the same, I didn't realize that.

documentation/source/python_entry_point.rst
```diff
@@ -0,0 +1,41 @@
+Python Entry Point
+==================
+
+By default, cocotb starts the :class:`~cocotb.regression.RegressionManager` as the testing framework for cocotb.
+However, there are many reasons why a user might want to replace the whole or a part of the existing Python framework with an alternative.
+To support this, the environment variable :envvar:`PYTHON_ENTRY` was created.
```
cmarqu:
```suggestion
To support this, the environment variable :envvar:`PYTHON_ENTRY` can be used.
```
The "was created" makes sense now in this PR, but in the future, the variable is just there, and referring to the creation may sound... quirky? (Not a native speaker, so my feeling may be wrong.)

documentation/source/python_entry_point.rst
```diff
@@ -0,0 +1,41 @@
+Python Entry Point
+==================
+
+By default, cocotb starts the :class:`~cocotb.regression.RegressionManager` as the testing framework for cocotb.
```
cmarqu: This isn't hyperlinked because ``RegressionManager`` is not in the Library Reference chapter - would it make sense to add it?

ktbarrett: I don't know if any of the regression manager was meant to be touched by the user and documenting it makes changing it even harder. Should I remove the ref instead?

cocotb/__init__.py
```diff
@@ -126,6 +128,19 @@ def _initialise_testbench(root_name):
     """
     _rlock.acquire()
 
+    global argc, argv, SIM_NAME, SIM_VERSION
+    argc = argc_
+    argv = argv_
```
eric-wieser: What are your thoughts about setting `sys.argv` with these as part of `gpi_embed`?

ktbarrett: That might be useful if we were trying to make the entry module seem more like a script where we don't specify an entry function, but I'm not sure if that is best. Right now I'm seeing the entry function as more of a `main` function. We could (and probably should) move the simulator info to getters in the simulator module.

eric-wieser: > We could (and probably should) move the simulator info to getters in the simulator module and we can get rid of argc form the call.

Makes sense to me.

documentation/source/python_entry_point.rst
```diff
@@ -0,0 +1,41 @@
...
+.. currentmodule:: cocotb
+
+PyGPI expects the following functions to be implemented in the entry module. The names are required.
+
+.. autofunction:: _sim_event
+
+.. autofunction:: _log_from_c
+
+.. autofunction:: _filter_from_c
+
```
eric-wieser: I know I've already expressed this elsewhere, for but reviewers who haven't seen my comments - I don't think we want to do this. These functions are designed specifically to connect to the Python `logging` infrastructure, and to so me it doesn't make a whole lot of sense to allow users to intercept them. If users want to configure logging, then they can do it through the builtin `logging` module.

The only reason they're defined in python at all is it makes things a little easier to write them. Idealistically, I'd consider them part of the same unit as `gpi_embed.c`. I wouldn't be opposed to moving them to `cocotb/_gpi_embed.py` to indicate this.

ktbarrett: This feature was never really intended to allow users to override the behavior, just allow them to run without invoking the cocotb Python module. Whether they are in `cocotb.log` or `cocotb._gpi_embed`, `cocotb.__init__` is still run. 

You are right though, after the refactoring these functions are not tied to cocotb's logger in any way and probably could be moved into C. Writing C API code sucks, is it possible we could define the functions in Python in a C string and compile them into Python function objects from `gpi_embed`?

eric-wieser: > just allow them to run without invoking the cocotb Python module

One option would be to create a `_simulator_gpi_embed.py` file, and use that.

> `cocotb.__init__` is still run.

perhaps the issue then is that `cocotb.__init__` being run is stateful. Presumably if it ran but did nothing other than define functions, you would have no issues with it?

ktbarrett: Presumably if it ran but did nothing other than define functions, you would have no issues with it?
It's better, and what I want to eventually do, but what's C's should be C's in this case. This can be pushed to another day. I'll revert the commit and continue pointing to `cocotb.log`.

eric-wieser: > but what's C's should be C's in this case

What I'm modelling this after is `numpy.core._internal`, which contains random python functions that are needed by the C code, and are far too much effort to write in C. I don't think we should be afraid of pushing details to python as long as those details remain private.

> I'll revert the commit and continue pointing to cocotb.log. 

Since this is a draft, feel free to keep iterating on the rest before doing that.

ktbarrett: I would like to refactor the module, but I don't want it too balloon too much. Large PRs tend to not get pulled in.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -229,53 +229,62 @@ int get_module_ref(const char *modname, PyObject **mod)
     return 0;
 }
 
+int parse_entry_point_str (
+    char *entry_point_str,
+    const char **entry_module_name,
+    const char **entry_func_name)
```
eric-wieser: Looking at [setuptools](https://github.com/pypa/setuptools/blob/682b6511ac67e021b542e74ce30e13fe52bc2da9/pkg_resources/__init__.py#L2484), the format there is `some.module:some.attr` - ie, the dots can appear after the colon too

ktbarrett: I am going to move this into Python. There I can much more easily express checking and handle issues, while more likely being correct. I'm with you on being able to add utility functions to Python. I will go through the __init__ refactor as a separate PR.

ktbarrett:
C changes are probably causing the failures. Still want to clean up `__init__.py` to work like what was done in #1332. Unfortunately, I can just cherry-pick the cleanup commit.

cocotb/_gpi_embed.py
```diff
@@ -0,0 +1,41 @@
...
+
+def _load_entry():
+    """
+    Parses entry point specified by :envar:`COCOTB_ENTRY_POINT`, gathers entry point information
+    """
+    entry_point_str = os.environ.get("COCOTB_ENTRY", "cocotb:_initialise_testbench")
+    entry_module_str, entry_func_str = entry_point_str.split(":", 1)
+    entry_module = importlib.import_module(entry_module_str)
+    entry_func = reduce(getattr, entry_func_str.split('.'), entry_module)
+    entry_sim_event = getattr(entry_module, "_sim_event")
```
eric-wieser: I wonder whether we could lift the restriction on the name of this function - would it be reasonable to have the entry point call `simulator.get_sim_event_handler(...)`?

cocotb/__init__.py
```diff
@@ -112,13 +113,15 @@ def _reopen_stream_with_buffering(stream_name):
...
+LANGUAGE = os.getenv("TOPLEVEL_LANG")
+
 
 def mem_debug(port):
     import cocotb.memdebug
     cocotb.memdebug.start(port)
 
 
-def _initialise_testbench(root_name):
+def _initialise_testbench(argv_):
```
eric-wieser: Opened https://github.com/cocotb/cocotb/issues/1428 to discuss just putting this in `sys.argv`, which I think we touched on before.

cocotb/__init__.py
eric-wieser: Opened https://github.com/cocotb/cocotb/issues/1428 to discuss just putting this in `sys.argv`

cocotb/_gpi_embed.py
ktbarrett: No, it wouldn't be too late, `_sim_event` is only callable by the simulator after initialization, so if a callback is registered in the entry function, all is well.

documentation/source/python_entry_point.rst
ktbarrett: I have decided to move as much processing from `gpi_embed.c` into a `_gpi_embed.py`

cocotb/_gpi_embed.py
ktbarrett: Well, there is an issue here, GPI expects to call `embed_sim_event` in the case of a sim event, which in turn calls the saved Python function `pEvent`. The simulator module does not have access to set anything in `gpi_embed`, it only works with the GPI interface.

We would have to change this interface so that GPI has a `register_sim_event_callback` function. `gpi_embed` would use the function to register `embed_sim_event` instead the function being at a known location.

ktbarrett: Continuing discussion in #1446.

ktbarrett:
All currently failing because of makefile issues. I can't seem to get them to work? Adding `GPI_EXTRA ?= cocotb` in `Makefile.inc` seems like it should work, but it doesn't.

ktbarrett:
I talked to @leftink about this change and his opinion was that extensions should be separate from `GPI_EXTRA`. `GPI_EXTRA` was originally intended for loading additional implementations only.

So I will refactor this to add `GPI_USERS`, and at that point we won't need `gpi_register_extension`. And of course I will document `GPI_EXTRA` anyways.

cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -434,6 +434,41 @@ void FliImpl::get_sim_precision(int32_t *precision)
...
+        char *c_info       = mti_GetProductVersion();      // Returned pointer must not be freed
+        std::string info   = c_info;
+        std::string search = " Version ";
+        std::size_t found  = info.find(search);
+
+        std::string product_str = c_info;
+        std::string version_str = c_info;
+
+        if (found != std::string::npos) {
+            product_str = info.substr(0,found);
```
cmarqu:
```suggestion
            product_str = info.substr(0, found);
```

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -122,15 +125,39 @@ int gpi_register_impl(GpiImplInterface *func_tbl)
...
+            break;
+        }
+    }
+    extension_entry_functions.clear();
 }
 
 void gpi_embed_end()
 {
-    embed_sim_event(SIM_FAIL, "Simulator shutdown prematurely");
+    gpi_embed_event(SIM_FAIL, "Simulator shutdown prematurely");
```
cmarqu: Our chance to change this to
```suggestion
    gpi_embed_event(SIM_FAIL, "Simulator shut down prematurely");
```
? (Feels a bit stupid to mention these tiny things, but our code is in such a good shape these days...)

documentation/source/building.rst
```diff
@@ -201,6 +205,16 @@ Environment Variables
 
     .. versionadded:: 1.3
 
+.. envvar:: COCOTB_ENTRY
+
+        Python framework entry point from the GPI.
```
cmarqu:
```suggestion
        The Python framework entry point from the GPI.
```

documentation/source/building.rst
```diff
@@ -201,6 +205,16 @@ Environment Variables
...
+        Entry point is specified as the module to load and function in that module to run at initialization.
```
cmarqu:
```suggestion
        The entry point is specified as the module to load and the function in that module to run at initialization.
```

documentation/source/building.rst
```diff
@@ -201,6 +205,16 @@ Environment Variables
...
+        Entry point is specified as ``entry.module.name:entry_function_name``.
```
cmarqu:
```suggestion
        Its format is ``entry.module.name:entry_function_name``.
```

documentation/source/python_entry_point.rst
```diff
@@ -0,0 +1,37 @@
+Custom Entry Point
+==================
+
+By default, cocotb starts the regression manager as the testing framework for cocotb.
+However, there are many reasons why a user might want to replace the whole or a part of the existing Python framework with an alternative.
+To support this, the environment variable :envvar:`COCOTB_ENTRY` can be used.
+This allow the user to specify the location of an alternative testing framework to run at initialization.
```
cmarqu:
```suggestion
This allows the user to specify the location of an alternative testing framework to run at initialization.
```

cocotb/_gpi_embed.py
```diff
@@ -0,0 +1,41 @@
...
+            None,
+            function_name
+        )
+        logger.handle(record)
+
+
+def _load_entry():
+    """
+    Parses entry point specified by :envar:`COCOTB_ENTRY_POINT`, gathers entry point information
+    """
```
cmarqu:
```suggestion
    """Gather entry point information by parsing :envar:`COCOTB_ENTRY_POINT`."""
```

cocotb/_gpi_embed.py
```diff
@@ -0,0 +1,41 @@
...
+
+def _filter_from_c(logger_name, level):
+    return logging.getLogger(logger_name).isEnabledFor(level)
+
+
+def _log_from_c(logger_name, level, filename, lineno, msg, function_name):
+    """
+    This is for use from the C world, and allows us to insert C stack
+    information.
+    """
```
cmarqu:
```suggestion
    """Log from the C world, allowing to insert C stack information."""
```
(I know this is from original code, but now is a chance to clean this up.)

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -219,9 +222,16 @@ int get_module_ref(const char *modname, PyObject **mod)
     return 0;
 }
 
-int embed_sim_init(gpi_sim_info_t *info)
+
+static void embed_sim_event(void*, gpi_event_t, const char*);
+static void cocotb_log_v(void* userdata, const char *name, enum gpi_log_levels level, const char *pathname, const char *funcname, long lineno, const char *msg, va_list argp);
```
cmarqu: Add a line break in the argument list?

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -230,214 +240,115 @@ int embed_sim_init(gpi_sim_info_t *info)
...
         goto cleanup;
     }
-    if (!PyCallable_Check(simlog_func)) {
-        LOG_ERROR("_log_from_c is not callable");
-        Py_DECREF(simlog_func);
-        goto cleanup;
-    }
 
-    set_log_handler(simlog_func);                                       // Note: This function steals a reference to simlog_func.
+    pLogHandler = simlog_func;                                          // Note: This function steals a reference to simlog_func.
```
cmarqu: The comment does not apply fully anymore, does it?

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -230,214 +240,115 @@ int embed_sim_init(gpi_sim_info_t *info)
...
     if (simlog_func == NULL) {
         PyErr_Print();
         LOG_ERROR("Failed to get the _filter_from_c method");
         goto cleanup;
     }
-    if (!PyCallable_Check(simlog_func)) {
-        LOG_ERROR("_filter_from_c is not callable");
-        Py_DECREF(simlog_func);
+
+    pLogFilter = simlog_func;                                           // Note: This function steals a reference to simlog_func.
```
cmarqu: The comment does not apply fully anymore, does it?

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -230,214 +240,115 @@ int embed_sim_init(gpi_sim_info_t *info)
...
+
+    gpi_set_log_handler(cocotb_log_v, NULL);
+
+    entry_tuple = PyObject_CallMethod(cocotb_gpi_module, "_load_entry", NULL);
+    if (entry_tuple == NULL) {
+        PyErr_Print();
+        LOG_ERROR("Unable to load entry point");
```
cmarqu: Printing the entry point name here could aid debugging.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -230,214 +240,115 @@ int embed_sim_init(gpi_sim_info_t *info)
...
+        goto cleanup;
+    }
+    if (!PyArg_ParseTuple(entry_tuple, "OOO", &cocotb_module, &cocotb_init, &pEventFn)) {
+        PyErr_Print();
+        LOG_ERROR("Unable to parse output of _load_entry");
```
cmarqu: Maybe add some more context for debugging?

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -122,15 +132,48 @@ int gpi_register_impl(GpiImplInterface *func_tbl)
...
+}
+
+static std::pair<std::string, std::string> parse_library_string(const char* begin, const char* end);
+static std::vector<std::pair<std::string, std::string>> parse_library_list_string(const char* lib_list);
+template <typename Func>
+static bool load_libraries(const std::vector<std::pair<std::string, std::string>>& libraries, Func&& load);
+
+void gpi_embed_init(int argc, char const* const* argv)
+{
+    // get GPI_USERS string
```
cmarqu: This new envvar should be documented, and I would call it ``COCOTB_GPI_USERS``.

ktbarrett: GPI is a self-sufficient entity that lives in this repo. "cocotb" should be scrubbed from GPI code. I'm going to add sections to the environment variables tab to disambiguate the owner of a particular environment variables when I add documentation.

ktbarrett:
Updated TODO.

cocotb/share/lib/embed/gpi_embed.c
ktbarrett: I added a try-catch block in Python to add the COCOTB_ENTRY string to the exception that gets reported here.

cocotb/share/lib/embed/gpi_embed.c
ktbarrett: Add exact function name and expected output.

documentation/source/building.rst
ktbarrett: I English real good.

cocotb/share/lib/gpi/GpiCommon.cpp
cmarqu: Okay, GPI should then get its own documentation chapter (or "part") at some point (I'm  working on some general doc reorganization anyway). If GPI even moves to its own repo, we are prepared.

cocotb/share/lib/embed/gpi_embed.c
ktbarrett: I broke every arg out onto its own line. I'm not a fan of multiple args on each of multiple lines.

ktbarrett:
I am probably going to break this PR into 2 or more. There are two major features being introduced: GPI extensions (`GPI_USERS`) and custom entry point from Python GPI interface ('COCOTB_ENTRY`). And on top of that there is a lot of cleanup, some of that is "necessary" to reduce interface requirements on use of those two new custom entry points.

ktbarrett:
The plan is there will be follow up PRs that: 1. add the GPI extension interface, 2. merge GPI logging into the GPI, 3. add new GPI interfaces to the simulator module and use them to break dependence on `_sim_event`, `_log_from_c`, and `_filter_from_c` being at a known location, they instead will be registered in the Python. 4. Cleanup `__init__.py` so we don't run any behavior when we load `_gpi_embed.py`.

cocotb/share/lib/gpi/gpi_priv.h
```diff
@@ -325,7 +327,7 @@ class GpiImplInterface {
 /* Called from implementation layers back up the stack */
 int gpi_register_impl(GpiImplInterface *func_tbl);
 
-void gpi_embed_init(gpi_sim_info_t *info);
+void gpi_embed_init(int argc, char const* const*argv);
```
cmarqu:
```suggestion
void gpi_embed_init(int argc, char const* const* argv);
```

documentation/source/building.rst
```diff
@@ -207,6 +207,16 @@ Environment Variables
     You can usually write out such a file from the simulator's GUI.
     This is currently supported for the Mentor Questa, Mentor ModelSim and Aldec Riviera simulators.
 
+.. envvar:: COCOTB_ENTRY
+
+        The Python framework entry point from the GPI.
```
cmarqu: Could you also add a line like ``COCOTB_ENTRY      Python entry point (default: cocotb._initialize_testbench)`` to https://github.com/cocotb/cocotb/blob/master/cocotb/share/makefiles/Makefile.sim#L76 ?

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -192,42 +192,16 @@ static std::vector<std::string> get_argv()
...
-    std::vector<char*> argv_cstr;
-    for (const auto& arg : argv) {
+    std::vector<std::string> argv_storage = get_argv();
+    std::vector<const char*> argv_cstr;
+    for (const auto& arg : argv_storage) {
         argv_cstr.push_back(const_cast<char*>(arg.c_str()));
     }
-    sim_info.argc = static_cast<int>(argv.size());
-    sim_info.argv = argv_cstr.data();
+    int argc = argv.size();
```
ktbarrett:
```suggestion
    int argc = argv_cstr.size();
```
Won't compile on Questa.

documentation/source/building.rst
ktbarrett: Done.

ktbarrett:
I'm not comfortable with this implementation anymore. I think what I had originally was the better solution. I will come back to this.

PR #1298, opened 2020-01-15
Increase version number to 1.4.0 pre-release

imphil:
The next (planned) version of cocotb will be 1.4.0. Increase the in-tree
version number to reflect that.

The naming is based on the rules outlined in
https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning.

### DIFF
--- a/cocotb/_version.py
+++ b/cocotb/_version.py
@@ -5,4 +5,4 @@
 # 1) we don't load dependencies by storing it in __init__.py
 # 2) we can import it in setup.py for the same reason
 # 3) we can import it into your module
-__version__ = '1.3.0'
+__version__ = '1.4.0.dev0'


imphil:
Note instead of using .dev1, we could use .a1. I'm not sold on either option and a quick look around didn't find a definitive answer in other packages. Let me know what you think.

In the future, we could think about using setuptools_scm to extract the version number from the git metadata, but as discussed in the maintainers meeting today, let's tackle that as a separate step in the future.

cocotb/_version.py
```diff
@@ -5,4 +5,4 @@
 # 1) we don't load dependencies by storing it in __init__.py
 # 2) we can import it in setup.py for the same reason
 # 3) we can import it into your module
-__version__ = '1.3.0'
+__version__ = '1.4.0.dev1'
```
eric-wieser:
```suggestion
__version__ = '1.4.0.dev0'
```
maybe?

imphil: Yes, let's go for "dev0". After reading https://packaging.python.org/guides/distributing-packages-using-setuptools/#standards-compliance-for-interoperability I got the impression that numbers start with 1, but https://www.python.org/dev/peps/pep-0440/ is quite clear that 0 is allowed.

Issue #1296, opened 2020-01-15
Swig is missing from CI


eric-wieser:
From [this CI run](https://travis-ci.org/cocotb/cocotb/jobs/635275148#L4128-L4139):
```
make[2]: Entering directory '/home/travis/build/cocotb/cocotb/examples/endian_swapper/tests'
/usr/bin/make results.xml
make[3]: Entering directory '/home/travis/build/cocotb/cocotb/examples/endian_swapper/tests'
mkdir -p sim_build
cd ../cosim && make
make[4]: Entering directory '/home/travis/build/cocotb/cocotb/examples/endian_swapper/cosim'
gcc  -fPIC -pthread -fno-strict-aliasing -O2 -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 \
-fexceptions --param=ssp-buffer-size=4 -D_GNU_SOURCE \
-fwrapv -DNDEBUG -I/opt/python/3.8.0/include/python3.8 -c io.c -o io_module.o
gcc -pthread -shared  io_module.o -L/opt/python/3.8.0/lib  -o io_module.so
swig -python -outcurrentdir ../hal/endian_swapper_hal.h
make[4]: swig: Command not found
```

themperek:
Need to add swig here https://github.com/cocotb/cocotb/blob/master/Dockerfile#L9

eric-wieser:
@themperek: Mind making a PR?

eric-wieser:
Or shall I do it in 1314?

themperek:
5min

eric-wieser:
Done in #1316

PR #1301, opened 2020-01-16
Install swig on linux ci

themperek:
Addressing #1296

As far as I can see to make it do anything one needs to change `MODULE` to `test_endian_swapper_hal` here: https://github.com/cocotb/cocotb/blob/ccb8a8311271a1f333553897e58b392d1655487c/examples/endian_swapper/tests/Makefile#L55

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ jobs:
     - stage: SimTests
       python: 3.7
       before_install: &travis_linx_prep
-        - sudo apt-get install gperf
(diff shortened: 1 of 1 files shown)


themperek:
@eric-wieser Mind that those libraries are not used in ci test anywhere because   `test_endian_swapper_hal` is not enabled.

eric-wieser:
Indeed, but that can be fixed in a follow-up

PR #1292, opened 2020-01-10
Replace removed Py_InitModule by PyModule_Create in HAL example

mariushegele:
`Py_InitModule` does not exist in python 3.

PR #1302, opened 2020-01-17
Add test_endian_swapper_hal tests to CI

eric-wieser:
Follows up on #1296, ~~relates to~~ replaces the accidentally deleted #1292

### DIFF
--- a/examples/endian_swapper/tests/Makefile
+++ b/examples/endian_swapper/tests/Makefile
@@ -54,18 +54,26 @@ else
     $(error "A valid value (verilog or vhdl) was not provided for TOPLEVEL_LANG=$(TOPLEVEL_LANG)")
 endif
 
-MODULE=test_endian_swapper
+MODULE := test_endian_swapper
 
-CUSTOM_COMPILE_DEPS = hal
-
-LD_LIBRARY_PATH := $(PWD)/../cosim:$(LD_LIBRARY_PATH)
+ifeq ($(OS),Linux)
+    # TODO: work out how to build the extension module on windows
+    MODULE := $(MODULE),test_endian_swapper_hal
+    CUSTOM_COMPILE_DEPS = hal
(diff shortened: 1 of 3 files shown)


eric-wieser:
Fails: https://travis-ci.org/cocotb/cocotb/jobs/638407465#L4531

themperek:
@eric-wieser But why it does fail the entire test?

eric-wieser:
This patch needs to be merged with the changes in #1292. Without that change, we get a linker error. The function isn't just deprecated, it doesn't exist any more.

themperek:
At the moment our makefile based system will not fail the test if it is not started. This is rather bad.

eric-wieser:
Agreed, but let's fix one thing at a time

themperek:
Agree

eric-wieser:
Test seems to be running: https://travis-ci.org/cocotb/cocotb/jobs/639932812#L5018

examples/endian_swapper/tests/Makefile
```diff
@@ -54,7 +54,7 @@ else
     $(error "A valid value (verilog or vhdl) was not provided for TOPLEVEL_LANG=$(TOPLEVEL_LANG)")
 endif
 
-MODULE=test_endian_swapper
+MODULE=test_endian_swapper,test_endian_swapper_hal
```
themperek:
```suggestion
ifeq ($(OS),Linux)
    MODULE=test_endian_swapper,test_endian_swapper_hal
else
    MODULE=test_endian_swapper
endif
```

eric-wieser: Did something more elaborate

eric-wieser:
Running fine on linux still: https://travis-ci.org/cocotb/cocotb/jobs/639958471#L4962

Let's check it no longer fails on windows / Mac

eric-wieser:
Force push was just a manual squash, no diff since the approved revision

PR #955, opened 2019-06-18
Use setuptools to compile the simulator extension module

eric-wieser:
cc @themperek. This might be a dead end, but it shows a solution that might help simplify makefiles

PR #1303, opened 2020-01-17
Add towncrier to maintain release notes

cmarqu:
Towncrier (https://towncrier.readthedocs.io/en/actual-freaking-docs/) allows user-defined categories which we may want to use later, but this PR just uses the defaults for now.

https://external-builds.readthedocs.io/html/cocotb/1303/release_notes.html

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,7 @@ All changes which should go into the main codebase of cocotb must follow this se
 - All code must pass existing tests.
   New functionality must be accompanied by tests, and bug fixes should add tests to increase the test coverage and prevent regressions.
 - If code changes or enhances documented behavior the documentation should be updated.
+- If a change is user-visible, a newsfragment should be added to `documentation/source/newsfragments`.
 - All pull requests must be accepted by at least one maintainer, with no maintainer strongly objecting.
   Reviews must be performed by a person other than the primary author of the code.
 - All commits should follow established best practices when creating a commit message:
--- a/documentation/requirements.txt
+++ b/documentation/requirements.txt
@@ -8,3 +8,4 @@ sphinxcontrib-spelling
 pyenchant
 sphinx-issues
 sphinx-argparse
+towncrier
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -334,6 +334,15 @@ spelling_ignore_pypi_package_names = False
 spelling_ignore_wiki_words = False
 spelling_show_suggestions = True
 
-# -- Setup for inheritance_diagram directive which uses graphviz ---------------
+# -- Extra setup for inheritance_diagram directive which uses graphviz ---------
 
 graphviz_output_format = 'svg'
+
+# -- Extra setup for towncrier -------------------------------------------------
+# see also https://towncrier.readthedocs.io/en/actual-freaking-docs/
+
+in_progress_notes = subprocess.check_output(['towncrier', '--draft'],
+                                            cwd='../..',
+                                            universal_newlines=True)
+with open('generated/master-notes.rst', 'w') as f:
+    f.write(in_progress_notes)
--- a/documentation/source/newsfragments/1255.feature.rst
+++ b/documentation/source/newsfragments/1255.feature.rst
@@ -1,0 +1,13 @@
+Cocotb now supports the following example of forking a *non-decorated* :ref:`async coroutine <async_functions>`.
+
+.. code-block:: python3
+
+   async def example():
+       for i in range(10):
+           await cocotb.triggers.Timer(10, "ns")
+
+   cocotb.fork(example())
+
+..
+   towncrier will append the issue number taken from the file name here:
+Issue
--- a/documentation/source/newsfragments/README.rst
+++ b/documentation/source/newsfragments/README.rst
@@ -1,0 +1,42 @@
+:orphan:
+
+Release Notes
+=============
+
+We are using `towncrier <https://pypi.org/project/towncrier/>`_ to handle
+our release notes, and this directory contains the input for it -
+"news fragments" which are short files that contain a small
+**ReST**-formatted text that will be added to the next version's
+Release Notes page.
+
+Each file should be named like ``<ISSUE_OR_PR>.<TYPE>.rst``,
+where ``<ISSUE_OR_PR>`` is an issue or a pull request number -
+whatever is most useful to link to,
+and ``<TYPE>`` is one of:
+
+* ``feature``: New user-facing feature.
+* ``bugfix``: A bug fix.
+* ``doc``: Documentation improvement.
+* ``removal``: Deprecation or removal of public API or behavior.
+* ``change``: A change in public API or behavior.
+
+In that file, make sure to use full sentences with correct case and punctuation,
+and do not use a bullet point at the beginning of the file.
+Use Sphinx references (see https://sphinx-tutorial.readthedocs.io/cheatsheet/)
+if you refer to added classes, methods etc.
+
+An example file could consist of the content between the marks:
+
+>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+Summary of a new feature.
+
+This is a second paragraph.
+<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+Note that the last paragraph should be a normal sentence and not e.g. code,
+because the issue number is appended there.
+
+Towncrier automatically assembles a list of unreleased changes when building Sphinx,
+meaning your notes will be visible in the documentation immediately after merging.
+When performing a release, ``towncrier`` should be run independently (in cocotb's root directory),
+which will delete all the merged newsfragments, and create new commits.
--- a/documentation/source/release_notes.rst
+++ b/documentation/source/release_notes.rst
@@ -4,6 +4,10 @@ Release Notes
 
 All releases are available from the `GitHub Releases Page <https://github.com/cocotb/cocotb/releases>`_.
 
+.. include:: generated/master-notes.rst
+
+.. towncrier release notes start
+
 cocotb 1.3.0
 ============
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,32 @@
+[tool.towncrier]
+    package = "cocotb"
+    directory = "documentation/source/newsfragments"
+    filename = "documentation/source/release_notes.rst"
+    issue_format = ":pr:`{issue}`"
+    underlines = '=-^"'  # the full list is '#*=-^"' but the two highest levels are not handled by towncrier
+    all_bullets = false
+
+    [[tool.towncrier.type]]
+        directory = "feature"
+        name = "Features"
+        showcontent = true
+
+    [[tool.towncrier.type]]
+        directory = "bugfix"
+        name = "Bugfixes"
+        showcontent = true
+
+    [[tool.towncrier.type]]
+        directory = "doc"
+        name = "Improved Documentation"
+        showcontent = true
+
+    [[tool.towncrier.type]]
+        directory = "removal"
+        name = "Deprecations and Removals"
+        showcontent = true
+
+    [[tool.towncrier.type]]
+        directory = "change"
+        name = "Changes"
+        showcontent = true


cmarqu:
Should this PR add a line to ``CONTRIBUTING.md`` to tell people to add a newsfragment for PRs?

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,46 @@
...
+    * ``c_api``: Changes in the Numpy C-API exported functions
+    * ``new_feature``: New user facing features like ``kwargs``.
+    * ``improvement``: Performance and edge-case changes
+    * ``change``: Other changes
+    * ``highlight``: Adds a highlight bullet point to use as a possibly highlight
+      of the release.
+
+You can install ``towncrier`` and run ``towncrier --draft``
+if you want to get a preview of how your change will look in the final release
+notes.
```
eric-wieser: I think we need to add a brief section somewhere as instructions for maintainers about how to assemble a release.

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,46 @@
+:orphan:
+
+Release Notes
+=============
+
+We are using `towncrier <https://pypi.org/project/towncrier/>`_ to handle
+our relase notes, and this directory  contains the input for it -
```
eric-wieser:
```suggestion
our release notes, and this directory contains the input for it -
```

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,46 @@
...
+"news fragments" which are short files that contain a small
+**ReST**-formatted text that will be added to the next version's
+Release Notes page.
+
+Make sure to use full sentences with correct case and punctuation,
+and use Sphinx references
+(see https://sphinx-tutorial.readthedocs.io/cheatsheet/).
```
eric-wieser: Do we want:
```
TheFeature
----------
Summary of new feature
```
or

```
* Summary of new feature
```
or 
```
Summary of new feature
```

as the format in each fragment? We should give an example of whatever we pick in this file (I don't care which)

documentation/source/newsfragments/README.rst
cmarqu: Once this PR is accepted I will add that information to https://github.com/cocotb/cocotb/wiki/Release-Process

documentation/requirements.txt
```diff
@@ -7,3 +7,4 @@ sphinxcontrib-makedomain
 sphinxcontrib-spelling
 pyenchant
 sphinx-issues
+towncrier
```
cmarqu: Actually, only the person doing a release needs to have towncrier, but it probably doesn't hurt listing it here anyway?

documentation/source/newsfragments/README.rst
cmarqu: Added (the last one since the bullet point is added by towncrier in our configuration).

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,50 @@
...
+    * ``highlight``: Adds a highlight bullet point to use as a possibly highlight
+      of the release.
+
+In that file, make sure to use full sentences with correct case and punctuation,
+and do not use a bullet point at the beginning of the file.
+Use Sphinx references
+(see https://sphinx-tutorial.readthedocs.io/cheatsheet/) if you refer
+to added classes, methods etc.
+
+An example line could be ``Summary of new feature.``.
```
eric-wieser: Are you describing the first line of the file, or its entire contents? Are multiple paragraphs supported?

documentation/source/newsfragments/1286.doc.rst
```diff
@@ -0,0 +1 @@
+New example ``matrix_multiplier``.
```
eric-wieser: #1286 isn't merged, so let's not do this yet. Can you instead add one for #1255, which is merged and release-note worthy?

CONTRIBUTING.md
```diff
@@ -61,6 +61,7 @@ All changes which should go into the main codebase of cocotb must follow this se
 - All code must pass existing tests.
   New functionality must be accompanied by tests, and bug fixes should add tests to increase the test coverage and prevent regressions.
 - If code changes or enhances documented behavior the documentation should be updated.
+- If a change is user-visible, a newsfragment should be added.
```
eric-wieser:
```suggestion
- If a change is user-visible, a newsfragment should be added to `documentation/source/newsfragments`.
```

documentation/requirements.txt
eric-wieser: Ideally we'd run towncrier as part of building the docs to get "in development" release notes, but we can do that in a future PR.

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[tool.towncrier]
```
eric-wieser: There's a reasonable chance the presence of this file will break #955. @themperek, could you test with an empty `pyproject.toml` file on your branch and see if things are ok still?

themperek: It seems to not cause any issues.

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,50 @@
...
+Release Notes page.
+
+Each file should be named like ``<PULL REQUEST>.<TYPE>.rst``, where
+``<PULL REQUEST>`` is a pull request number, and ``<TYPE>`` is one of:
+
+* ``feature``: New user-facing feature.
+* ``bugfix``: A bug fix.
+* ``doc``: Documentation improvement.
+* ``removal``: Deprecation or removal of public API.
+* ``misc``: A ticket has been closed, but it is not of interest to users.
```
imphil: If something is not of interest to a user, then it doesn't need to appear in the release notes.

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,50 @@
...
+**ReST**-formatted text that will be added to the next version's
+Release Notes page.
+
+Each file should be named like ``<PULL REQUEST>.<TYPE>.rst``, where
+``<PULL REQUEST>`` is a pull request number, and ``<TYPE>`` is one of:
+
+* ``feature``: New user-facing feature.
+* ``bugfix``: A bug fix.
+* ``doc``: Documentation improvement.
+* ``removal``: Deprecation or removal of public API.
```
imphil: API -> behavior

(API is a common term for libraries, but cocotb has more than that, e.g. the Makefiles)

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,50 @@
...
+    * ``future``: Changes existing code to emit a FutureWarning.
+    * ``expired``: Removal of a deprecated part of the API.
+    * ``compatibility``: A change which requires users to change code and is not
+      backwards compatible. (Not to be used for removal of deprecated features.)
+    * ``c_api``: Changes in the Numpy C-API exported functions
+    * ``new_feature``: New user facing features like ``kwargs``.
+    * ``improvement``: Performance and edge-case changes
+    * ``change``: Other changes
+    * ``highlight``: Adds a highlight bullet point to use as a possibly highlight
+      of the release.
```
imphil: What's the purpose of this section? Can it be removed?

documentation/source/newsfragments/README.rst
imphil: Can you give a full example of such a file?

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,50 @@
...
+An example line could be ``Summary of new feature.``.
+
+You can install ``towncrier`` and run ``towncrier --draft``
+if you want to get a preview of how your change will look in the final release
+notes.
+
+.. note::
+
+    This README was adapted from the numpy changelog readme which in turn
+    took it from pytest under the terms of the MIT licence.
```
imphil: You can remove that paragraph, the text is trivial enough.

pyproject.toml
```diff
@@ -0,0 +1,7 @@
...
+    package = "cocotb"
+    directory = "documentation/source/newsfragments"
+    filename = "documentation/source/release_notes.rst"
+    issue_format = ":pr:`{issue}`"
+    underlines = '=-^"'  # the full ist is '#*=-^"' but the two highest levels are not handled by towncrier
```
imphil: ist -> list

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[tool.towncrier]
+    package = "cocotb"
+    directory = "documentation/source/newsfragments"
+    filename = "documentation/source/release_notes.rst"
+    issue_format = ":pr:`{issue}`"
```
imphil: Is this for issues and PRs? (It doesn't matter much in GitHub, but would still be good to have the right semantics.)

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,50 @@
+:orphan:
```
imphil: We tried to move documentation to one sentence per line style. How about doing that for this file as well?

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,50 @@
...
+=============
+
+We are using `towncrier <https://pypi.org/project/towncrier/>`_ to handle
+our release notes, and this directory contains the input for it -
+"news fragments" which are short files that contain a small
+**ReST**-formatted text that will be added to the next version's
+Release Notes page.
+
+Each file should be named like ``<PULL REQUEST>.<TYPE>.rst``, where
+``<PULL REQUEST>`` is a pull request number, and ``<TYPE>`` is one of:
```
imphil: In many cases, the issue is the start and end of a discussion, and the PR only implements it. Of course, that's not always true, and not every PR has an associated issue.

In case where the issue is the main center of a discussion (as it should be), then the issue number is the one to use here. Can you reword that somehow to allow both issue and PR numbers?

documentation/source/newsfragments/README.rst
imphil: Why doesn't towncrier always run when creating the docs? This way we can also see if the fragments were added correctly by previewing the RTD page that's generated for each PR.

documentation/source/newsfragments/README.rst
eric-wieser:
```suggestion
* ``removal``: Deprecation or removal of public API or behavior.
```

documentation/source/newsfragments/README.rst
eric-wieser: It's just a comment, which is there for people who come back to this and think about adding new categories and need inspiration. We could replace it with a link to the numpy one to make this shorter.

documentation/source/newsfragments/README.rst
eric-wieser: What matters most is that the number is unique, but agreed that sometimes the issue is more valuable.

It's usually trivial to get from the PR to the issue though through github anyway, so I'd be inclined to leave this advice as is.

documentation/source/newsfragments/README.rst
cmarqu: Right, it's meant for this discussion mainly - are we okay with the categories that towncrier supports by default, or do we want different ones? For example, the deprecation vs. removal differentiation.

documentation/requirements.txt
cmarqu: Will do that in this PR.

documentation/source/newsfragments/README.rst
cmarqu: I'm adding that.

documentation/source/release_notes.rst
```diff
@@ -4,6 +4,972 @@ Release Notes
 
 All releases are available from the `GitHub Releases Page <https://github.com/cocotb/cocotb/releases>`_.
 
+.. towncrier release notes start
+
+Cocotb 1.4.0.dev0 (2020-02-07)
```
cmarqu: Of course, this doesn't belong in this PR, but my VPN is down so I have to fix it later.

cmarqu:
I need to think more about what the work flow with this will be, and investigate how other projects do it.

documentation/source/newsfragments/1255.feature.rst
```diff
@@ -0,0 +1,11 @@
...
+
+.. code-block:: python3
+
+   async def example():
+       for i in range(10):
+           await cocotb.triggers.Timer(10, "ns")
+
+   cocotb.fork(example())
+
+Issue
```
eric-wieser:
```suggestion
```
CI fails due to the trailing space on line. If this formatting is deliberate, it seems a little weird

cmarqu: It's indeed deliberate, because towncrier will append ``(#1255)`` there. Without an extra line with at least a single non-whitespace character, it will just append it to the last code line. :(
towncrier has lots of little warts like that and at the same time is not very actively maintained, so I'm getting doubtful of it in general.
On the other hand, CI should maybe not care about ReST files.

documentation/source/newsfragments/README.rst
cmarqu: I would not follow that rule strictly in cases where it still leads to longish lines, but punctuation is in general a good candidate for line breaks.
In the case of this specific file, the ReST is not rendered but read in raw form, so should be pleasant to read in raw as well.

pyproject.toml
cmarqu: In the README, I now leave it to the author to pick issue or PR number, but we have to artificially pick one way here. It will end up as a working link always.

documentation/source/newsfragments/1255.feature.rst
eric-wieser: I assume that without the trailing space, it also does the wrong thing?

eric-wieser: Another option would be:
```
Issue (:pr:`1255`)

.. ignore the towncrier issue text
..
```

cmarqu: > I assume that without the trailing space, it also does the wrong thing?

It looks like it indeed adds the space itself.

cmarqu:
Running towncrier from ``documentation/source/conf.py`` cannot stay as is because that modifies ``release_notes.rst`` and removes the newsfragments. This is okay when done as part of the RTD processing, but not when someone is building the Sphinx doc locally (see e.g. https://github.com/ethereum/lahja/pull/153).
We would need to introduce a ``readthedocs.yml`` to be able to run towncrier only there, where a dirtied environment does not matter because it gets thrown away.

Also, all projects I have seen using towncrier have elaborate CI for the docs already and plug in some more for towncrier (https://github.com/ethereum/web3.py/pull/1431#discussion_r317047958). But we have lots of warnings because of breathe (converting Doxygen for the C/C++ files), so that's not a good starting point.

towncrier somehow doesn't feel right to me, with its magic that is removing files as it runs.
What if we used those newsfragments, sorted them into subdirectories by release number, and then used https://nedbatchelder.com/blog/201904/cog_30.html to refresh the ``release_notes.rst`` in place? (This idea, written down after midnight, may have other problems.)

documentation/source/conf.py
```diff
@@ -334,6 +334,11 @@
...
 
-# -- Setup for inheritance_diagram directive which uses graphviz ---------------
+# -- Extra setup for inheritance_diagram directive which uses graphviz ---------
 
 graphviz_output_format = 'svg'
+
+# -- Extra setup for towncrier -------------------------------------------------
+# see also https://towncrier.readthedocs.io/en/actual-freaking-docs/
+
+subprocess.call(['towncrier', '--yes'], cwd='../..')
```
eric-wieser: Does `--draft` help at all?

eric-wieser: Yeah, it sounds like you want:
```
in_progress_notes = subprocess.check_output(['towncrier', '--draft'])
with open('source/generated/master-notes.rst', 'w') as f:
    f.write(in_progress_notes )
```
and then add a `.. include source/generated/master-notes.rst` in the release notes.

cmarqu: Ok, that looks good.
Let me try to write down then what to do at release time assuming this change (feel free to edit this text directly):

> * ``cd documentation/source/``
> * insert the contents of ``generated/master-notes.rst`` into ``release_notes.rst`` below the existing ``.. include``
> * move the heading (with the version number this release is for) from above the ``.. include`` to below it
> * add a new heading with the next version number above ``.. include``
> * run ``git rm newsfragments/[0-9]+.*.rst``

Anything I forgot?

eric-wieser: I don't think you need any of that. Isn't the step at release time just run `towncrier` by itself?

eric-wieser: > move the heading (with the version number this release is for) from above the .. include to below it

This is solved by just having `conf.py` output the heading as part of `master-notes.rst`

cmarqu: Oh right, we just leave the special comment for towncrier in, next to the include.

eric-wieser: Bingo - the special comment goes below the include, and is exercised only during release

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Each file should be named like ``<ISSUE_OR_PR>.<TYPE>.rst``,
+where ``<ISSUE_OR_PR>`` is an issue or a pull request number -
+whatever is most useful to link to,
+and ``<TYPE>`` is one of:
+
+* ``feature``: New user-facing feature.
+* ``bugfix``: A bug fix.
+* ``doc``: Documentation improvement.
+* ``removal``: Deprecation or removal of public API or behavior.
```
eric-wieser: I think we might want a category for behavior changes, such as #1375 which isn't really a feature, but is a user-visible change that might be worth at least mentioning

cmarqu: I just added that locally and will push once my network problems are over.

cmarqu: ``change`` is now in.

documentation/source/newsfragments/README.rst
```diff
@@ -0,0 +1,39 @@
...
+>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+Summary of a new feature.
+
+This is a second paragraph.
+<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+Note that the last paragraph should be a normal sentence and not e.g. code,
+because the issue number is appended there.
+
+Towncrier runs automatically when you run any of the Sphinx output build commands.
```
eric-wieser: I thought I commented on this but maybe I forgot to submit the comment.

```suggestion
Towncrier automatically assembles a list of unreleased changes when building sphinx.

When performing a release, it should be run independently, which will delete all the merged newsfragments, and create new commits.
```
or something.

cmarqu: Added, thanks.

PR #1304, opened 2020-01-18
Add documentation for the `cocotb-config` script.

eric-wieser:
https://external-builds.readthedocs.io/html/cocotb/1304/library_reference.html#the-cocotb-config-script

image


eric-wieser:
As a follow-up, would be good to add docs to the root argparse object

Issue #1181, opened 2019-11-03
Library naming and structuring discussion


cmarqu:
Pull Request #1168 renames ``libvhpi`` to ``libcocotbvhpi`` because Synopsys VCS-MX already ships a ``libvhpi``.
The PR spawned a discussion about library names in general; this issue collects the comments made there so that they don't get lost.

Technical points made so far were by @ktbarrett who wrote initially:
> Why not ``libcocotbvcs``? Since we have to compile the GPI using simulator-specific defines, ``libcocotbvhpi`` is misleading, as it might not work with another simulator with VHPI support. This is probably universal to the other objects as well.

and  later:

> Tool specific code is only an issue if it affects user code or user-facing code. Simulator-specific workarounds at the simulator interface level are unavoidable, but as long as the GPI abstracts them away, there is no problem. ``libvhpi.so`` and all the other simulator-to-GPI abstraction objects should be associativity named to the simulator they support. Then we do not have to rebuild an object every time we change simulators that use the same interface.

> Personally, I'd prefer ``libgpi.{SIM}.so`` seeing as these objects are a specific implementation of the GPI.

@themperek then wrote:
> When we are at this. Following #1062 I was proposing to compile everything at setup using setuptools (separate for every simulator). I can prepare this for cocotb it there is a will to move in this direction. It is implemented here: https://github.com/themperek/cocotb-test/blob/master/cocotb_test/build_libs.py Can change naming conversion.

- [ ] Use default neme of `libvpi` also for `Icarus`.

ktbarrett:
I want to add another point before further discussion. A particular simulator can have multiple interfaces. For example, Questa supports the FLI, VPI, and will soon support VHPI.

ktbarrett:
I like @themperek's idea of building all supported gpi implementations at setup and installing them to the venv. We could add a script to the distribution that would return information about the build and installation for those that don't use the makefiles for their projects.

eric-wieser:
> A particular simulator can have multiple interfaces.

Perhaps then `libgpi_{SIM}_{which_pi}.so`?

I don't see a lot of precedence for using `.` within library names, so would be inclined not to

ktbarrett:
[...]

PR #1306, opened 2020-01-19
Build libs on setup using setuptools

themperek:
All libraries (vpi/vhpi/fli) are built on setup using [setuptools](https://setuptools.readthedocs.io/en/latest/setuptools.html) into separate directories per simulator. No Makefiles are used for compilation. Everything is defined in single `_build_libs.py` file. For Windows, a .def file has been added that allows compilation without simulator binaries available. Expedition is Questa/Modelsim (all platforms) which needs `mti.h` file.

- fewer Makefiles (only used for the simulator start)
- gives the possibility to distribute a binary package ([bdist_wheel](https://www.python.org/dev/peps/pep-0427/))
- allow easier implementation of alternative execution to makefiles (ex. [edalize](https://github.com/olofk/edalize))
- faster startup
- easier multi-os support

The changes are in `_build_libs.py` and `setup.py` and the rest is Makefile remove and cleanup. 

The changes should be invisible to the user (faster startup time since the libraries are compiled once).

NVC is not supported since it does not seem to work in any case https://github.com/nickg/nvc/issues/64

To get any information on compilation one needs to `pip install -v` see.: https://github.com/pypa/pip/issues/2933 https://github.com/pypa/pip/issues/6634


cocotb/build_libs.py
```diff
@@ -0,0 +1,476 @@
...
+    except OSError as e:
+        if e.errno == errno.EEXIST:
+            os.remove(link_name)
+            os.rename(target, link_name)
+        else:
+            raise e
+
+
+def _build_lib(lib, dist, build_dir):
+    """Use setuptools to build `lib` into `build_dir`."""
```
cmarqu:
```suggestion
    """Use setuptools to build *lib* into *build_dir*."""
```

setup.py
```diff
@@ -30,7 +30,36 @@
...
+    def run(self):
+        install.run(self)
+
+        lib_dir = path.join(self.install_lib, "cocotb", "libs")
+        if not path.exists(lib_dir):
+            makedirs(lib_dir)
+
+        build(build_dir=lib_dir)
+
+class PostDevelopCommand(develop):
```
cmarqu:
```suggestion
class PostInstallCommandDev(develop):
```

setup.py
```diff
@@ -45,8 +74,25 @@ def package_files(directory):
 # this sets the __version__ variable
 exec(read_file(path.join('cocotb', '_version.py')))
 
+# force platform specyfic wheel  (root_is_pure)
```
cmarqu:
```suggestion
# force platform specific wheel  (root_is_pure)
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,476 @@
...
+            sim_define="VCS",
+        )
+
+    #
+    # Aldec
+    #
+    vsimsa_path = find_executable("vsimsa")
+    if vsimsa_path is None:
+        logger.warning(
+            "Riviera executable not found. No VPI/VHPI interface will not be avaliable."
```
cmarqu:
```suggestion
            "Riviera executable not found. No VPI/VHPI interface will not be available."
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,476 @@
...
+            share_lib_dir=share_lib_dir,
+            dist=dist,
+            sim_define="MODELSIM",
+            extra_lib=modelsim_extra_lib,
+            extra_lib_dir=modelsim_extra_lib_path,
+        )
+
+    if vsim_path is None:
+        logger.warning(
+            "Modelsim/Questa executable (vopt) executable not found. FLI interface will not be avaliable."
```
cmarqu:
```suggestion
            "Modelsim/Questa executable (vopt) executable not found. FLI interface will not be available."
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,476 @@
...
+    logger.warning("Compiling interface libraries for Icarus Verilog ...")
+    icarus_build_dir = os.path.join(build_dir, "icarus")
+    icarus_compile = True
+    icarus_extra_lib = []
+    icarus_extra_lib_path = []
+    if os.name == "nt":
+        iverilog_path = find_executable("iverilog")
+        if iverilog_path is None:
+            logger.warning(
+                "Icarus Verilog executable not found. VPI interface will not be avaliable."
```
cmarqu:
```suggestion
                "Icarus Verilog executable not found. VPI interface will not be available."
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,476 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2013 Potential Ventures Ltd
```
cmarqu: Use SPDX here since it is a new file?

cocotb/build_libs.py
```diff
@@ -0,0 +1,476 @@
+#!/usr/bin/env python
```
eric-wieser:
```suggestion
```
This looks like it is a module to import, not a script - so it should not have a shebang

setup.py
```diff
@@ -30,7 +30,36 @@
 
 from setuptools import setup
 from setuptools import find_packages
-from os import path, walk
+from setuptools.command.install import install
+from setuptools.command.develop import develop
+from os import path, walk, makedirs
+import cocotb
```
eric-wieser: Importing `cocotb` during `setup.py` is probably a bad idea.

cocotb/build_libs.py
```diff
@@ -0,0 +1,476 @@
...
+def build_vhpi_lib(
+    build_dir,
+    include_dir,
+    share_lib_dir,
+    dist,
+    sim_define,
+    extra_lib=[],
+    extra_lib_dir=[],
+):
+    libvhpi = Extension(
```
cmarqu: This has been renamed to ``libcocotbvhpi`` (#1168).

themperek:
How do you see this? Should I clean up/delete the Makefiles? @eric-wieser @cmarqu @imphil

cocotb/build_libs.py
```diff
@@ -0,0 +1,492 @@
...
+            os.path.join(icarus_build_dir, icarus_vpi_lib_name),
+            os.path.join(icarus_build_dir, "gpivpi.vpl"),
+        )
+
+    #
+    #  Modelsim/Questa
+    #
+    logger.warning("Compiling interface libraries for Modelsim/Questa ...")
+    vsim_path = find_executable("vopt")
+    modelsim_build_dir = os.path.join(build_dir, "modelsi")
```
cmarqu:
```suggestion
    modelsim_build_dir = os.path.join(build_dir, "modelsim")
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,492 @@
...
+            share_lib_dir=share_lib_dir,
+            dist=dist,
+            sim_define="MODELSIM",
+            extra_lib=modelsim_extra_lib,
+            extra_lib_dir=modelsim_extra_lib_path,
+        )
+
+    if vsim_path is None:
+        logger.warning(
+            "Modelsim/Questa executable (vopt) executable not found. FLI interface will be available."
```
cmarqu:
```suggestion
            "Modelsim/Questa executable (vopt) executable not found. No FLI interface will be available."
```
?

cocotb/build_libs.py
```diff
@@ -0,0 +1,492 @@
...
+    vsim_path = find_executable("vopt")
+    modelsim_build_dir = os.path.join(build_dir, "modelsi")
+    modelsim_compile = True
+    modelsim_extra_lib = []
+    modelsim_extra_lib_path = []
+
+    if os.name == "nt":
+        if vsim_path is None:
+            logger.warning(
+                "Modelsim/Questa executable (vopt) not found. VPI interface will be available."
```
cmarqu:
```suggestion
                "Modelsim/Questa executable (vopt) not found. No VPI interface will be available."
```
?

cocotb/build_libs.py
```diff
@@ -0,0 +1,471 @@
...
+    logger.warning("Compiling interface libraries for Icarus Verilog ...")
+    icarus_build_dir = os.path.join(build_dir, "icarus")
+    icarus_compile = True
+    icarus_extra_lib = []
+    icarus_extra_lib_path = []
+    if os.name == "nt":
+        iverilog_path = find_executable("iverilog")
+        if iverilog_path is None:
+            logger.warning(
+                "Icarus Verilog executable not found. VPI interface will be available."
```
cmarqu:
```suggestion
                "Icarus Verilog executable not found. No VPI interface will be available."
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,471 @@
...
+                os.path.join(share_lib_dir, "fli", "FliObjHdl.cpp"),
+            ],
+            extra_link_args=["-Wl,-rpath,$ORIGIN"],
+        )
+
+        try:
+            _build_lib(libfli, dist, modelsim_build_dir)
+        except:  # noqa: E722
+            logger.warning(
+                "Building FLI intercae for Modelsim faild!"
```
cmarqu:
```suggestion
                "Building FLI interface for Modelsim failed!"
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,471 @@
...
+            )  # some Modelsim version doesn not include FLI?
```
cmarqu:
```suggestion
            )  # some Modelsim version does not include FLI?
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,471 @@
...
+
+def _extra_link_args(lib_name):
+    if sys.platform == "darwin":
+        return ["-Wl,-install_name,@loader_path/%s.so" % lib_name]
+    else:
+        return []
+
+
+def build_common_libs(build_dir, include_dir, share_lib_dir, dist):
+
```
cmarqu: This being a non-private function should have a docstring (also the ones below).
Funny that the private ones _do_ have docstrings :)

cocotb/build_libs.py
```diff
@@ -0,0 +1,471 @@
...
+    if sys.platform == "darwin":
+        ld_library = sysconfig.get_config_var("LIBRARY")
+    else:
+        ld_library = sysconfig.get_config_var("LDLIBRARY")
+
+    if ld_library:
```
cmarqu:
```suggestion
    if ld_library is not None:
```
(https://docs.python.org/3/library/sysconfig.html#sysconfig.get_config_var, returns ``None`` if not found)

cocotb/build_libs.py
```diff
@@ -0,0 +1,471 @@
...
+        libraries=[python_lib_link, "pthread", "m", "cocotbutils"],
+        library_dirs=gpilog_library_dirs,
+        sources=[os.path.join(share_lib_dir, "gpi_log", "gpi_logging.c")],
+        extra_link_args=_extra_link_args("libgpilog"),
+    )
+
+    _build_lib(libgpilog, dist, build_dir)
+
+    #
+    #  libgpilog
```
garmin-mjames:
```suggestion
    #  libcocotb
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,471 @@
...
+        library_dirs=[build_dir],
+        libraries=["gpilog", "cocotbutils"],
+        sources=[os.path.join(share_lib_dir, "embed", "gpi_embed.c")],
+        extra_link_args=_extra_link_args("libcocotb"),
+    )
+
+    _build_lib(libcocotb, dist, build_dir)
+
+    #
+    #  libgpilog
```
garmin-mjames:
```suggestion
    #  libgpi
```

setup.py
```diff
@@ -45,8 +72,25 @@ def package_files(directory):
 # this sets the __version__ variable
 exec(read_file(path.join('cocotb', '_version.py')))
 
+# force platform specific wheel  (root_is_pure)
```
cmarqu:
```suggestion
# force platform-specific wheel (root_is_pure=False means "not pure Python", i.e. has C libs)
```

cocotb/share/makefiles/Makefile.sim
```diff
@@ -56,7 +56,7 @@ endif
 # recompilation of the simulation if cocotb is updated.
 CUSTOM_SIM_DEPS += $(shell find $(COCOTB_PY_DIR)/cocotb/ -name "*.py")
 
-include $(COCOTB_SHARE_DIR)/lib/Makefile
+#include $(COCOTB_SHARE_DIR)/lib/Makefile
```
cmarqu: So actually, this whole file could be removed?: https://github.com/cocotb/cocotb/blob/master/cocotb/share/lib/Makefile

themperek: yes and many more

cocotb/build_libs.py
```diff
@@ -0,0 +1,471 @@
...
+
+    return _build_lib(libcocotbvhpi, dist, build_dir)
+
+
+def build(build_dir):
+
+    logger = logging.getLogger(__name__)
+
+    distutils.log.set_verbosity(0)  # Disable logging compilation commands
+    # distutils.log.set_verbosity(distutils.log.DEBUG) # Set DEBUG level
```
garmin-mjames: Remove commented-out line.
Alternatively, use DEBUG for develop mode.

themperek:
If anyone has an idea why `LGTM analysis: C/C++ ` please let me know.

garmin-mjames:
Any thoughts on the `cocotb-build-libs` command or adding a command to `cocotb-config` to rebuild the libs after install? This would be useful for cocotb development.

From: https://github.com/cocotb/cocotb/issues/1062#issuecomment-519586473

themperek:
@garmin-mjames `cocotb-build-libs` sound good. Will do in separate PR.

cocotb/build_libs.py
```diff
@@ -0,0 +1,473 @@
...
+
+    if (
+        os.name == "nt"
+    ):  # On Windows there is an issue with symlink and rename? !Workaround!
+        shutil.copy2(target, link_name)
+        return
+
+    try:
+        os.rename(target, link_name)
+    except OSError as e:
```
eric-wieser:
```suggestion
    except FileExistsError:
```

eric-wieser:
Is there a reason we can't hook into the normal python extension module infrastructure, as in 

https://github.com/cocotb/cocotb/pull/955/files#diff-2eeaed663bd0d25b7e608891384b7298R79-R86

cocotb/build_libs.py
```diff
@@ -0,0 +1,473 @@
...
+
+
+def _extra_link_args(lib_name):
+    if sys.platform == "darwin":
+        return ["-Wl,-install_name,@loader_path/%s.so" % lib_name]
+    else:
+        return []
+
+
+def _build_common_libs(build_dir, include_dir, share_lib_dir, dist):
```
eric-wieser: Could this function be refactored into:

* A function that returns a list of `Extension` objects
* A for loop that calls `_build_lib` on each one

cocotb/build_libs.py
```diff
@@ -0,0 +1,473 @@
...
+                "Icarus Verilog executable not found. No VPI interface will be available."
+            )
+            icarus_compile = False
+        else:
+            icarus_path = os.path.dirname(os.path.dirname(iverilog_path))
+            icarus_extra_lib = ["vpi"]
+            icarus_extra_lib_path = [os.path.join(icarus_path, "lib")]
+
+    if icarus_compile:
+        _build_common_libs(icarus_build_dir, include_dir, share_lib_dir, dist)
```
eric-wieser: Why do we rebuild the common libraries anew for each simulator? It might be worth a todo about this.

themperek:
>Is there a reason we can't hook into the normal python extension module infrastructure, as in

There ware a few reasons but some of them went away and maybe I know how to work around others too. Will try.

eric-wieser:
> Will try.

Perhaps make my suggested refactor to return a list of `Extension` objects first - that should make the final setuptools step easy to do later if it turns out not working.

themperek:
@eric-wieser Life would get easier if we name `libvpi` -> `libvpi_simulator` ex. `libvpi_ius`.  I think for `vhpi` would need custom entry per simulator here ?
https://github.com/cocotb/cocotb/blob/9caf77cb6ceae739858ecd4a3760631470aac77c/cocotb/share/lib/vhpi/VhpiImpl.cpp#L966

eric-wieser:
Let's try to avoid doing that as part of this PR, but agreed it might help.

setup.py
```diff
@@ -58,8 +70,8 @@ def package_files(directory):
     install_requires=[],
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
     packages=find_packages(),
-    include_package_data=True,
     package_data={'cocotb': package_files('cocotb/share')},
+    ext_modules=get_ext(),
```
eric-wieser: Nice job!

cocotb/build_libs.py
themperek: see: https://github.com/cocotb/cocotb/pull/1306/files#diff-67c744c69df11e3f27fa958b42c1d390R84-R86

```python
# We need to build common libraries for each simulator to avoid issues with loading
# libraries from different directories in multi os configuration without setting
# globally LD_LIBRARY_PATH or similar.
```

cocotb/build_libs.py
themperek: Something like this is done now.

setup.py
themperek: It looks simple ... at the end

cocotb/build_libs.py
```diff
@@ -0,0 +1,401 @@
...
+logger = logging.getLogger(__name__)
+
+
+class build_ext(_build_ext):
+
+    # Needed for Windows to not assume python module (generate interface in def file)
+    def get_export_symbols(self, ext):
+        return None
+
+    # For proper cocotb library nameing, based on https://github.com/cython/cython/issues/1740
```
eric-wieser: [This comment](https://github.com/cython/cython/issues/1740#issuecomment-411835514) makes me worry a little.

Perhaps just add something here like:
```
# TODO: for the simulator python library, we should be leaving this suffix in place
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,401 @@
...
+    def get_ext_filename(self, ext_name):
```
eric-wieser:
```suggestion
    def get_ext_filename(self, ext_name):
        """ Like the base class method, but removes the `.cpython-36m-x86_64-linux-gnu.` part before the extension
```

cocotb/build_libs.py
```diff
@@ -0,0 +1,401 @@
...
+    # Add extra library_dirs path
+    def finalize_options(self):
+        _build_ext.finalize_options(self)
+
+        for ext in self.extensions:
+            ext.library_dirs.append(
+                os.path.join(self.build_lib, os.path.dirname(ext._full_name))
+            )
+
+    # copy libs into proper direcotry in develop
```
eric-wieser:
```suggestion
    # copy libs into proper directory in develop
```

setup.py
```diff
@@ -28,9 +28,12 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
+import logging
 from setuptools import setup
 from setuptools import find_packages
 from os import path, walk
+from cocotb.build_libs import get_ext ,build_ext
```
eric-wieser:
```suggestion
from cocotb.build_libs import get_ext, build_ext
```

setup.py
```diff
@@ -45,8 +48,17 @@ def package_files(directory):
 # this sets the __version__ variable
 exec(read_file(path.join('cocotb', '_version.py')))
 
+# store log form build_libs and display at the end in verbose mode
```
cmarqu:
```suggestion
# store log from build_libs and display at the end in verbose mode
```

themperek:
CI sees to fail now due to https://github.com/cocotb/cocotb/pull/1326 https://travis-ci.org/cocotb/cocotb/jobs/642255747#L788-L801 

@ktbarrett Fix should probably go into https://github.com/cocotb/cocotb/pull/1335 if not there yet.

cocotb/build_libs.py
themperek: We could filter this out but this is how it at the moment.

ktbarrett:
@themperek I added the fix into my PR.

How did removing `-Werror` cause your issue? Doesn't your script source compile options from `sysconfig?. It should only have `-Werror` if the system Python was compiled with it.

themperek:
@ktbarrett `-Werror` propagates same way as in makefiles when extensions are compiled.

themperek:
on hold until #1335

ktbarrett:
@themperek So on OSX the sysconfig picks up `CFLAGS` and/or `CXXFLAGS`, while on other systems it doesn't? 

After #1330 comes in, I had planned on dropping `-Werror` from the SimTests stages of Travis, and only using it on the BuildLibsTests, which should also prevent this.

themperek:
@ktbarrett OSX uses clang as a compiler. I guess it just makes different warnings than GCC.

ktbarrett:
@themperek For sure it does, but I don't see `-Werror` in the other builds at all, perhaps I'm blind.

themperek:
@ktbarrett They are compiled during `pip install` (pip by default hides all the output, on osx it crashes so ..) you need to do `pip install -v`

ktbarrett:
Ahh, okay. I am blind.

eric-wieser:
@themperek: #1335 is in, let's see if you can get this to pass again

themperek:
@ktbarrett some more: https://travis-ci.org/cocotb/cocotb/jobs/643078251#L546-L572

```  cocotb/share/lib/vhpi/VhpiImpl.cpp:691:37: error: implicit conversion changes signedness: 'int' to 'unsigned int' [-Werror,-Wsign-conversion]
                              idx += (scale * (left - raw_idx));
                                      ^~~~~ ~
  cocotb/share/lib/vhpi/VhpiImpl.cpp:691:53: error: implicit conversion changes signedness: 'int' to 'unsigned int' [-Werror,-Wsign-conversion]
                              idx += (scale * (left - raw_idx));
                                                    ~ ^~~~~~~
  cocotb/share/lib/vhpi/VhpiImpl.cpp:694:37: error: implicit conversion changes signedness: 'int' to 'unsigned int' [-Werror,-Wsign-conversion]
                              idx += (scale * (raw_idx - left));
                                      ^~~~~ ~
  cocotb/share/lib/vhpi/VhpiImpl.cpp:694:46: error: implicit conversion changes signedness: 'int' to 'unsigned int' [-Werror,-Wsign-conversion]
                              idx += (scale * (raw_idx - left));
                                               ^~~~~~~ ~
  cocotb/share/lib/vhpi/VhpiImpl.cpp:697:39: error: implicit conversion changes signedness: 'unsigned int' to 'int' [-Werror,-Wsign-conversion]
                          scale = scale * len;
                                ~ ~~~~~~^~~~~
  cocotb/share/lib/vhpi/VhpiImpl.cpp:697:33: error: implicit conversion changes signedness: 'int' to 'unsigned int' [-Werror,-Wsign-conversion]
                          scale = scale * len;
                                  ^~~~~ ~
  cocotb/share/lib/vhpi/VhpiImpl.cpp:715:28: error: implicit conversion changes signedness: 'int' to 'uint32_t' (aka 'unsigned int') [-Werror,-Wsign-conversion]
                  idx = left - index;
                      ~ ~~~~~^~~~~~~
  cocotb/share/lib/vhpi/VhpiImpl.cpp:717:29: error: implicit conversion changes signedness: 'int' to 'uint32_t' (aka 'unsigned int') [-Werror,-Wsign-conversion]
                  idx = index - left;
                      ~ ~~~~~~^~~~~~
  cocotb/share/lib/vhpi/VhpiImpl.cpp:722:72: error: implicit conversion changes signedness: 'uint32_t' (aka 'unsigned int') to 'int32_t' (aka 'int') [-Werror,-Wsign-conversion]
              new_hdl = vhpi_handle_by_index(vhpiIndexedNames, vhpi_hdl, idx);
                        ~~~~~~~~~~~~~~~~~~~~                             ^~~
```

@eric-wieser Not sure if disable this or wait for a patch.

ktbarrett:
@themperek  I will open up another patch and fix all the libraries with `-Wsign-conversion` enabled. However, going forward if this script is going to source it's compile options from places we can't control (not that it's a bad thing), it shouldn't be run with `-Werror`. If this is the way libraries will be compiled from here on out, `-Werror` should be removed from the environment in `tox.ini` as a part of this PR.

To attempt to track and enforce C code warning compliance we should bring in #1330. However, this means we can't remove the existing makefiles.

themperek:
I do not see why it cannot stay like it is. Tox is used only for testing so it looks like the correct place.

ktbarrett:
Warnings are important, and we should enforce their compliance, but that needs to be decoupled from behavioral checks.

The only way to run `-Werror`, while pulling in random warning options from sysconfig, indefinitely, is to make the C code warning free. That's a monumental task, almost every line of C I've seen has some issue. If we don't, every time Python or the compiler updates on our CI, there is the potential for the regressions to stop passing. And that shouldn't get in the way of progress.

ktbarrett:
@themperek Looks like fixing `-Wsign-conversion` is going to be pretty painful. The interface header `vhpi_user.h` defines all of the internal types as unsigned types, which is just ***wrong*** and I'll have to explicitly cast to the correct type at every call site, unless changing `vhpi_user.h` is possible.  Either way, please continue on, this isn't life or death stuff.

eric-wieser:
> Looks like fixing -Wsign-conversion is going to be painful

Let's make a PR to turn that warning back off then - once CI is building these libs, we can attempt to reinstate it.

themperek:
In this PR I have disabled warnings: https://github.com/cocotb/cocotb/pull/1306/files#diff-67c744c69df11e3f27fa958b42c1d390R95
I think it may be better to remove "-Werror" for now and first, fix all the warnings. I can configure tox to display all the workings in CI.

setup.py
```diff
@@ -28,9 +28,12 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
+import logging
 from setuptools import setup
 from setuptools import find_packages
 from os import path, walk
+from cocotb.build_libs import get_ext, build_ext
```
eric-wieser: I don't think this self-import is particularly safe - it will likely fail if installed from anywhere but `.`.

Numpy does something similar with `numpy.distutils`, but wraps the work in:
```
    src_path = os.path.dirname(os.path.abspath(__file__))
    old_path = os.getcwd()
    os.chdir(src_path)
    sys.path.insert(0, src_path)
    try:
        setup(...)
    finally:
        del sys.path[0]
        os.chdir(old_path)
```

themperek: Seems to work for me as is (tired different directory and form page).

eric-wieser: So `cd /tmp; pip uninstall cocotb; pip install full/path/to/cocotb` works?

eric-wieser: I think this might break in virtualenvs, I remember running into this problem with another package.

themperek: >So cd /tmp; pip uninstall cocotb; pip install full/path/to/cocotb works?
Yes

This also `pip install -v http://github.com/themperek/cocotb/archive/build_on_setup.zip`

eric-wieser: Ok, I now remember why this is a bad idea - it prevents cocotb ever taking external dependencies, because you import cocotb before you try and install its dependencies

eric-wieser: Do you envisage users ever importing `build_libs`? If not, then I'd propose we move this stuff to `setup_build_libs.py` or similar, and import it from there (and not install it with cocotb).

themperek: It seems some requested to have a separate command like "cocotb-build-libs": https://github.com/cocotb/cocotb/issues/1062#issuecomment-519586473

themperek: Can we keep this "not ugly" and fix in case someone spots an issue?

themperek: As far as I can read this import should be fine if there are no extra dependencies.

ktbarrett:
I'd prefer to see this as a build script that can be used to build and install specific libraries for a specific simulator on demand (`cocotb-build --sim ghdl --prefix=/usr/local --libs cocotb vhpi` or something), and then a thin wrapper that runs during setup that autodetects installed simulators and calls this script.

The base script would be useful for building wheels and warning compliance so there is value added to less direct implementation.

eric-wieser:
> The base script would be useful for building wheels

Aren't you describing `setup.py`?

setup.py
eric-wieser: How about:
* rename to `cocotb._build_libs`, to stop anyone depending on this directly (yet)
* Adding a comment here saying something like
  ```
  # note: cocotb is not installed properly yet, but we can import it anyway
  # because it's in the currently directory. We'll need to change this if we add to
  # `install_requires`.
  ```

cocotb/build_libs.py
```diff
@@ -0,0 +1,420 @@
...
+            ext_name = "dll"
+        else:
+            ext_name = "so"
+
+        filename_short = os.path.join(head, tail_split[0] + "." + ext_name)
+
+        # icarus requires vpl extension
+        if filename.find("icarus") >= 0:
+            filename_short = filename_short.replace("libvpi.so", "gpivpi.vpl")
+            filename_short = filename_short.replace("libvpi.dll", "gpivpi.vpl")
```
eric-wieser: Why the rename from `lib` to `gpi`? Would be nice to indicate the reason in the comment too.

cocotb/build_libs.py
```diff
@@ -0,0 +1,420 @@
...
+            )
+            if ext._needs_stub:
+                self.write_stub(package_dir or os.curdir, ext, True)
+
+
+def _extra_link_args(lib_name):
+    if sys.platform == "darwin":
+        return ["-Wl,-install_name,@loader_path/%s.so" % lib_name]
+    else:
+        return []
```
eric-wieser: If you know why this ia needed, a comment would be great

cocotb/build_libs.py
```diff
@@ -0,0 +1,420 @@
...
+
+
+_base_warns = ["-Wall", "-Wextra", "-Wconversion", "-Wcast-qual", "-Wwrite-strings"]
+_cc_warns = _base_warns + ["-Wstrict-prototypes", "-Waggregate-return"]
+_ccx_warns = _base_warns + ["-Wnon-virtual-dtor", "-Woverloaded-virtual"]
+
+# TODO Enable warnings when fixed
+_extra_cc_compile_args = ["-std=gnu99"]  # + _cc_warns
+_extra_cxx_compile_args = ["-std=c++11"]  # + _ccx_warns
```
eric-wieser: Is this now fixed? If not, then can you comment out the specific warnings that are failing, so that CI can enforce we don't regress on the ones that are fixed.

cocotb/build_libs.py
```diff
@@ -0,0 +1,420 @@
...
+
+# We need to build common libraries for each simulator to avoid issues with loading
+# libraries from different directories in multi os configuration without setting
+# globally LD_LIBRARY_PATH or similar.
```
eric-wieser: Could you elaborate on this further?

Am I right in thinking that the issue is that all the libraries live in different directories? Would the problem go away if they didn't?

Not asking for any code changes here, but a thorough explanation would be valuable in this comment.

cocotb/build_libs.py
```diff
@@ -0,0 +1,420 @@
...
+    else:
+        python_version = sysconfig.get_python_version().replace(".", "")
+        python_lib_link = "python" + python_version
+
+    if os.name == "nt":
+        ext_name = "dll"
+        python_lib = python_lib_link + "." + ext_name
+    else:
+        ext_name = "so"
+        python_lib = "lib" + python_lib_link + "." + ext_name
```
eric-wieser: Can you extract this to a helper function
```python
def _get_python_lib():
    """ Get the library for embedded the python interpreter """
    ...
    return python_lib
```

cocotb/build_libs.py
themperek: >Am I right in thinking that the issue is that all the libraries live in different directories?

Yes. As of my understanding and playing a bit with this: In Linux/Windows one can do `-rpath,$ORIGIN` to point where to load dependencies but on osx one can only do  `-install_name,@loader_path/%s.so`  on the depended library once. Since there are multiple dependencies I did not find a way to make it work if not all are in one directory. One could set `LD_LIBRARY_PATH/DYLD_LIBRARY_PATH/PATH` but this is not good practice and can cause some side effects.

>Would the problem go away if they didn't (and were uniquely named)?

Yes.

cocotb/build_libs.py
themperek: This is how it is now in the makefiles. I propose to make new PR when we use unique names for all simulators.

ktbarrett:
I see that the script is building most of the libraries unconditionally. Is there anyway to not?

* We can't download things from pypi
* We use Questa mostly
* Each of our projects can be running different versions of any and every tool, including different version of cocotb so a global install is not possible
* project-local virtualenvs will get blown away from time to time and we get penalized by recompiling for a bunch of simulators we don't have

Our currently solution, when this PR comes in, it to just remove most of the build script script in our fork. But maintaining forks is bothersome. I'd prefer it if there was some way to specify only building against certain simulators.

themperek:
>project-local virtualenvs will get blown away from time to time and we get penalized by recompiling for a bunch of simulators we don't have

Sorry. Do not understand this?

themperek:
@ktbarrett You cannot do `python setup.py build --build-lib my_dir`  somehow?

ktbarrett:
Currently, our automation software uses the makefiles to build the libraries and install them to a project-local cache. Project caches get stale or buggy for some reason or another, so people just `rm -rf` them and have the software rebuild a new cache, which requires rebuilding cocotb. We just don't want to pay for what we aren't using, every time we do that.

It's not a huge deal, it only takes about a minutes to build on my machine, but I have 40-50 other people who will complain.

ktbarrett:
@themperek  no it's not custom directory thing, its that it's building for Icarus, GHDL, VCS, Xcelium, etc. when installing locally, though I don't use any of them and never will. Our situation with our local installs leads to compiling them all more often that your would otherwise, so it's a time waste thing, although not a big one. If it's too much work I'll pick it up after this is in.

eric-wieser:
It's possible we can use the setuptools "extras" feature to handle that - so `pip install path-to-cocotb[Icarus,questa]` or something.

themperek:
> although not a big one. If it's too much work I'll pick it up after this is in. 

If we use unique names for libraries it will take just few seconds more (really should not make a difference). 

>It's possible we can use the setuptools "extras" feature to handle that - so pip install path-to-cocotb[Icarus,questa] or something.

Yes, there should be a way to add extra commands.

eric-wieser:
I'd probably lean towards leaving that for a follow-up

cocotb/_build_libs.py
```diff
@@ -0,0 +1,438 @@
...
+    #  libgpilog
+    #
+    gpilog_library_dirs = []
+    if sys.platform == "darwin":
+        gpilog_library_dirs = [sysconfig.get_config_var("LIBDIR")]
+
+    libgpilog = Extension(
+        os.path.join("cocotb", "libs", sim_define.lower(), "libgpilog"),
+        include_dirs=[include_dir],
+        libraries=[_get_python_lib_link(), "pthread", "m", "cocotbutils"],
```
eric-wieser: Is this needed? I thought extension modules were automatically linked against python
```suggestion
        libraries=["pthread", "m", "cocotbutils"],
```

cocotb/setup.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python
```
eric-wieser: Duplicate file, it looks like half your changes are in this one, and half in the other setup.py. Can you merge them into the other one?

cocotb/_build_libs.py
themperek: One gets
```
gpivpi:`/cocotb/libs/icarus/gpivpi.vpl' failed to open using dlopen() because:
 /cocotb/libs/icarus/libcocotb.so: undefined symbol: _Py_NoneStruct.
```

themperek: sometimes: https://travis-ci.org/themperek/cocotb/builds/643871989 (very interesting)

themperek: Ha: `Changed in version 3.8: On Unix, C extensions are no longer linked to libpython except on Android and Cygwin.` here: https://docs.python.org/3/distutils/apiref.html#distutils.core.Extension

eric-wieser: It sounds like `_get_python_lib_link` should be in `libcocotb`, since that's the one that actually embeds python?

themperek: Can be just followed the Makefiles https://github.com/cocotb/cocotb/blob/master/cocotb/share/lib/gpi_log/Makefile

Will check.

eric-wieser: The makefile has `_get_python_lib_link` for `cocotb` too, I think.

themperek: Does this have to be done like this (is this needed)?https://github.com/cocotb/cocotb/blob/master/cocotb/share/lib/embed/gpi_embed.c#L138

It actully looks as not need if we link correctly?  (I am definitely not an expert here)

eric-wieser: Not an expert either, but that sound quite plausible. Let's not remove that in this PR though.

eric-wieser: What was the verdict here?

cocotb/_build_libs.py
```diff
@@ -0,0 +1,438 @@
...
+        """
+
+        filename = _build_ext.get_ext_filename(self, ext_name)
+
+        head, tail = os.path.split(filename)
+        tail_split = tail.split(".")
+
+        filename_short = os.path.join(head, tail_split[0] + "." + _get_lib_ext_name())
+
+        # icarus requires vpl extension, gpivpi is default in Makfiles
```
eric-wieser:
```suggestion
        # icarus requires vpl extension, gpivpi is default in Makefiles
```

cocotb/_build_libs.py
```diff
@@ -0,0 +1,438 @@
...
+        if filename.find("icarus") >= 0:
```
eric-wieser:
```suggestion
        if "icarus" in filename >= 0:
```

cocotb/_build_libs.py
```diff
@@ -0,0 +1,438 @@
...
+            os.makedirs(os.path.dirname(dest_filename), exist_ok=True)
+
+            copy_file(
+                src_filename, dest_filename, verbose=self.verbose, dry_run=self.dry_run
+            )
+            if ext._needs_stub:
+                self.write_stub(package_dir or os.curdir, ext, True)
+
+
+# Allow to load dependencies from direcotry where vpi/vhpi/fli library is located on osx
```
eric-wieser:
```suggestion
# Allow to load dependencies from the directory where vpi/vhpi/fli library is located on osx
```

cocotb/_build_libs.py
themperek: Do like it is in makefiles (`_get_python_lib_link()` to `libcocotb` and `libgpilog` for now).

eric-wieser: Alright, that seems fair for now.

On that note thought - the makefiles don't seem to add `libm` and `libpthread` to `gpilog` - is the difference intentional?

themperek: Actully is used in `simulator` and `libgpilog` will correct.

cocotb/_build_libs.py
```diff
@@ -0,0 +1,438 @@
...
+
+    return python_lib
+
+# TODO Add "-Wconversion" to _base_warns when fixed
+_base_warns = ["-Wall", "-Wextra", "-Wcast-qual", "-Wwrite-strings"]
+_cc_warns = _base_warns + ["-Wstrict-prototypes", "-Waggregate-return"]
+_ccx_warns = _base_warns + ["-Wnon-virtual-dtor", "-Woverloaded-virtual"]
+
+_extra_cc_compile_args = ["-std=gnu99"] + _cc_warns
+_extra_cxx_compile_args = ["-std=c++11"] + _ccx_warns
```
eric-wieser: Thanks! cc @ktbarrett, just so you know we've backed that warning out for now. We can reinstate it as part of a PR in which CI passes.

cocotb/_build_libs.py
eric-wieser: My question here about `m` and `pthread` still applies - I don't see them in the old makefile - can you point me to what I'm missing?

eric-wieser:
I suspect @imphil is going to want a longer commit message for this, since it's a fairly substantial (but great) change.

cocotb/_build_libs.py
eric-wieser: Searching in the repo for `pthread` finds no interesting results: https://github.com/cocotb/cocotb/search?q=pthread&unscoped_q=pthread

themperek: sorry missed this will check

eric-wieser:
Did you try building a wheel file, `python setup.py bdist_wheel`, and then installing it with `pip install path/to/cocotb.whl`? Not a blocker, but would be good to know if it works.

themperek:
>Did you try building a wheel file, python setup.py bdist_wheel, and then installing it with pip install path/to/cocotb.whl? Not a blocker, but would be good to know if it works.

Yes. Seems to work.

themperek:
Wheel package to test: https://cernbox.cern.ch/index.php/s/wSb1ohGPdPkU7kg/download

cocotb/_build_libs.py
eric-wieser: Success, I assume?

themperek: Yes. You have a good eye ;-)

cocotb/share/makefiles/Makefile.sim
eric-wieser: Can we stick a to-do comment at the top of this makefile that it isn't used any more?

themperek: All makefiles in `cocotb/share/lib` can/should be removed.

eric-wieser:
From https://github.com/cocotb/cocotb/pull/1306#discussion_r373129624

> All makefiles in `cocotb/share/lib` can/should be removed.

I think we should do one of the following:

* Declare them as defunct, to prevent anyone trying to make changes to them. Create an issue about removing them, and stick a one-line comment at the top of each saying something like
  ```
  # cocotb is now built using setup.py, and this makefile is no longer used
  # TODO[gh-XXXX]: Remove these files
  ```
* Keep them for users not ready to migrate. We should add a CI test to verify they still work.
* Remove the makefiles entirely right now, if they not longer are usable at all.

themperek:
Using Makefiles to compile libraries as it is will not work at the moment.

eric-wieser:
If the makefiles no longer work, then leaving them around is only going to lead to bug reports saying they don't work. Let's pick option 1 or option 3.

Makefile
```diff
@@ -29,7 +29,7 @@
 
 all: test
 
-include cocotb/share/makefiles/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.inc
```
eric-wieser: This is unrelated cleanup, right?

themperek: It is needed to get tests to work inside tox virtual environment.

ktbarrett:
If the makefiles no longer work, how do you recommend building the libraries without installing the Python package? Is there some pip magic I'm unaware of?

eric-wieser:
@ktbarrett: can you give a motivation behind wanting to do that? As an example, the only way to compile the numpy c extension is also to build the python components too.

Perhaps though you're looking for `python3 setup.py build_ext`?

cocotb/_build_libs.py
```diff
@@ -0,0 +1,439 @@
...
+from distutils.spawn import find_executable
+from setuptools.command.build_ext import build_ext as _build_ext
+from distutils.file_util import copy_file
+
+
+logger = logging.getLogger(__name__)
+
+
+def _get_lib_ext_name():
+    """ Get the name of default library file extension on given os """
```
cmarqu:
```suggestion
    """ Get name of default library file extension on given OS. """
```

themperek:
> If the makefiles no longer work, how do you recommend building the libraries without installing the Python package? Is there some pip magic I'm unaware of?

Not sure why one would like this but as @eric-wieser wrote `python setup.py build_ext --build-lib .` should do this.

cocotb/_build_libs.py
```diff
@@ -0,0 +1,439 @@
...
+    # copy libs into proper directory in develop
+    def copy_extensions_to_source(self):
+        build_py = self.get_finalized_command("build_py")
+        for ext in self.extensions:
+            fullname = self.get_ext_fullname(ext.name)
+            filename = self.get_ext_filename(fullname)
+            modpath = fullname.split(".")
+            package = ".".join(modpath[:-1])
+            package_dir = build_py.get_package_dir(package)
+            dest_filename = os.path.join(package_dir, filename)
```
eric-wieser:
```suggestion
            # unlike the method from `setuptools`, we do not call `os.path.basename` here
            dest_filename = os.path.join(package_dir, filename)
```

cocotb/share/makefiles/Makefile.pylib.Darwin
```diff
@@ -35,24 +35,3 @@ ifneq ($(COCOTB_PYTHON_DEBUG),)
...
-	PYTHON_DYNLIBDIR:=$(subst 64,,$(PYTHON_DYNLIBDIR))
-endif
-
-PYLIBS:=$(shell $(PYTHON_BIN) -c 'from __future__ import print_function; from distutils import sysconfig; import os; print("-l"+os.path.splitext(sysconfig.get_config_var("LIBRARY"))[0][3:])')
-
-PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'from __future__ import print_function; import distutils.sysconfig; print(distutils.sysconfig.get_python_inc())')
-
-# Some Python installations can return a static Python library (libpython*.a) even if a shared Python library (libpython*.so) is
-# also present in a Python installation library path. The Python replace function is harmless for a Python dynamic installation
-PYTHON_DYN_LIB:=$(shell $(PYTHON_BIN) -c 'from __future__ import print_function; from distutils import sysconfig; print(sysconfig.get_config_var("LDLIBRARY").replace(".a", ".so"))')
```
eric-wieser: Is this still handled anywhere? Can we test this somehow?

themperek: I have removed all of this.

eric-wieser: Does this mean that static python libraries (#1182) are broken again?

eric-wieser:
Needs a rebase

cocotb/share/makefiles/Makefile.pylib.Darwin
themperek: Should not. It works slightly different now (hopefully better) and is independent what extension will be returned. 

```python
ld_library = sysconfig.get_config_var("LDLIBRARY")
python_lib_link = os.path.splitext(ld_library)[0][3:]
python_lib = "lib" + python_lib_link + "." + _get_lib_ext_name()
```

PS. I guess we can avoid this all together (as discussed above). Removing `PYTHON_SO_LIB`.

eric-wieser:
Endian_swapper [has an error](https://travis-ci.org/cocotb/cocotb/jobs/645419312#L4558),

> io.c:36:20: fatal error: Python.h: No such file or directory

eric-wieser:
Building `endian_swapper` using a local `setup.py` might also be sensible, and would also be a good example too.

themperek:
> Building `endian_swapper` using a local `setup.py` might also be sensible, and would also be a good example too.

Yes. Will follow up with PR.

cocotb/share/makefiles/Makefile.pylib
```diff
@@ -27,22 +27,9 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
-# All common pyhon related rules
+# All common python related rules
```
eric-wieser:
```suggestion
# All common python related rules, kept only for downstream C extensions modules
# which copy the pattern of the endian_swapper HAL example. In future these modules
# should use `setuptools` to build their HAL interface.
```

cocotb/share/makefiles/Makefile.sim
```diff
@@ -56,7 +53,7 @@ endif
...
```
eric-wieser:
```suggestion
```

setup.py
```diff
@@ -76,3 +93,5 @@ def package_files(directory):
         "Topic :: Scientific/Engineering :: Electronic Design Automation (EDA)",
     ],
 )
+
+print(log_stream.getvalue())
```
eric-wieser: Perhaps
```suggestion
sys.stderr.write(log_stream.getvalue())
```

themperek: Is perhaps nicer but is this is not an error?

eric-wieser: Matters only if someone is piping the output of `pip`, which probably doesn't matter anyway

eric-wieser: Also avoids an extra trailing newline

ktbarrett:
@eric-wieser The way we install cocotb is peculiar, we don't install it as a package. We build the libraries and store them in a project-local cache and source cocotb's Python using PYTHONPATH. It mirrors how we utilize other not-necessarily-Python tools/libraries in our project. We will eventually start leveraging virtual environments, but there is a good amount of work to get there and I need a stopgap. 

`python setup.py build_exit --build-dir` seems the way to go, thanks.

eric-wieser:
@ktbarrett: are you looking for `pip install --target=your_cache_directory .` perhaps?

ktbarrett:
@eric-wieser We don't want to affect the user's global environment *in any way*. That command would install artifacts locally, but register the package with the global package manager, no? Our project environments are canned, and are loaded automatically when running certain commands and unloaded after exit.

eric-wieser:
> That command would install artifacts locally, but register the package with the global package manager,

My impression was that it did nothing global, but I didn't actually check.

themperek:
@ktbarrett Some more options http://docs.python.org/install/index.html#custom-installation

eric-wieser:
Needs a (trivial) rebase

cocotb/_build_libs.py
```diff
@@ -0,0 +1,440 @@
...
+                    os.path.join(share_lib_dir, "fli", "FliObjHdl.cpp"),
+                ],
+                extra_link_args=["-Wl,-rpath,$ORIGIN"],
+            )
+
+            ext.append(fli_ext)
+
+        else:
+            logger.warning(
+                "Building FLI interface for Modelsim failed! No mti.h available."
```
cmarqu:
```suggestion
                "Cannot build FLI interface for Modelsim/Questa - no mti.h available."
```
is closer to what is actually happening.

cocotb/_build_libs.py
```diff
@@ -0,0 +1,440 @@
...
+        ext.append(modelsim_vpi_ext)
+
+    if vsim_path is None:
+        logger.warning(
+            "Modelsim/Questa executable (vdbg) executable not found. No FLI interface will be available."
+        )
+    else:
+        modelsim_dir = os.path.dirname(os.path.dirname(vsim_path))
+        modelsim_include_dir = os.path.join(modelsim_dir, "include")
+        if os.path.isfile(os.path.join(modelsim_include_dir, "mti.h")):
```
eric-wieser:
```suggestion
        mti_path = os.path.join(modelsim_include_dir, "mti.h")
        if os.path.isfile(mti_path):
```

cocotb/_build_libs.py
eric-wieser: Or perhaps more clearly, which would have probably helped you debug:
```suggestion
                "Building FLI interface for Modelsim failed! "
                "The mti.h header for '{}' was not found at '{}' "
                .format(vsim_path, mti_path)
```

tox.ini
```diff
@@ -18,3 +18,7 @@ commands =
 
 whitelist_externals =
     make
+
+[testenv:pyw3]
```
eric-wieser: Perhaps
```suggestion
[testenv:py3_mingw]
```
would be clearer

.travis.yml
```diff
@@ -83,7 +83,7 @@ jobs:
         - powershell "$CONDA_PATH; conda install --yes -c msys2 m2-base m2-make m2w64-toolchain libpython"
         - powershell "$CONDA_PATH; pip install tox"
       script:
-        - powershell "$CONDA_PATH; tox -e py3"
+        - powershell "$CONDA_PATH; tox -e pyw3"
```
eric-wieser: Can you add a comment here explaining why mingw is needed?

cocotb/_build_libs.py
```diff
@@ -0,0 +1,447 @@
...
+    def get_ext_filename(self, ext_name):
+        """
+        Like the base class method, but removes the ``.cpython-36m-x86_64-linux-gnu.`` part before the extension.
+
+        Also replaces ``.pyd`` with ``.dll`` on windows.
+        """
+
+        filename = _build_ext.get_ext_filename(self, ext_name)
+
+        #for the simulator python extension library, leaving suffix in place
```
eric-wieser: Flake8 wants
```suggestion
        # for the simulator python extension library, leaving suffix in place
```

.travis.yml
themperek: I actully do not know if it is strictly needed. I did not try VS (doubt will work out of the box). At the moment failed because of `-Werror`

eric-wieser: Do you have a link to the failure?

themperek: https://travis-ci.org/themperek/cocotb/jobs/646754559#L454

eric-wieser: Perhaps we should just set all these warning flags via TOX, so we can pick them per-compiler?

eric-wieser:
> Perhaps we should just set all these warning flags via TOX, so we can pick them per-compiler?

Elaborating on that thought - the makefiles before this PR were mingw only, as they assumed gcc compiler flags.

However, using setuptools means that windows users will by default end up using mingw. Therefore, we should either:
* Add some documentation telling users they need to install with a special `compiler` option as you do here
* Remove all the gcc-specific flags from `_build_libs.py`, and just set them for CI only (via the `CXXFLAGS` env var, I think). This ought to make msvc work too!

Ideally we'd do both, but the latter can wait for a follow-up PR.

themperek:
>Ideally we'd do both, but the latter can wait for a follow-up PR.

In general windows documentation is outdated and I volunteer to update this if this PR is accepted.

cocotb/_build_libs.py
```diff
@@ -0,0 +1,447 @@
...
+    """ Get the library for embedded the python interpreter """
+
+    if os.name == "nt":
+        python_lib = _get_python_lib_link() + "." + _get_lib_ext_name()
+    else:
+        python_lib = "lib" + _get_python_lib_link() + "." + _get_lib_ext_name()
+
+    return python_lib
+
+# TODO Add "-Wconversion" to _base_warns when fixed
```
eric-wieser:
```suggestion
# TODO: make this work for compilres like MSVC which have different flag syntaxes
# TODO: Add "-Wconversion" to _base_warns when fixed
```

.travis.yml
```diff
@@ -83,7 +83,7 @@ jobs:
         - powershell "$CONDA_PATH; conda install --yes -c msys2 m2-base m2-make m2w64-toolchain libpython"
         - powershell "$CONDA_PATH; pip install tox"
       script:
-        - powershell "$CONDA_PATH; tox -e py3"
+        - powershell "$CONDA_PATH; tox -e py3_mingw"
```
eric-wieser:
```suggestion
        # the default compiler on windows, msvc, does not understand our compiler flags
        - powershell "$CONDA_PATH; tox -e py3_mingw"
```

themperek: There is more to this than compiler flags https://travis-ci.org/themperek/cocotb/builds/646876123
If anything something "msvc is not supported" maybe.

eric-wieser: Sounds good to me. Can you reference `gh-1372` in that comment?

eric-wieser:
```suggestion
        # TODO[gh-1372]: the default compiler on windows, msvc, is not supported
        - powershell "$CONDA_PATH; tox -e py3_mingw"
```

garmin-mjames:
I don't see any mechanism to do debug builds.

cmarqu:
> I don't see any mechanism to do debug builds.

``CFLAGS="-g" pip install -v -f -e .`` would do it?

ktbarrett:
Would a future extension to this that allows users to only build for requested simulators/interface be accepted?

I can imagine it would catch options in `setup.py` before the call to `setup`, filter out the carefully selected (as to not interfere) options, use those options to configure `get_ext`, and assign the remaining args back into `sys.argv`.

An alternative and cleaner approach would add those options to `build_ext` and getting rid of `get_ext` altogether, but that would be a major change to the script.

eric-wieser:
Another (trivial) conflict I'm afraid.

cocotb/_build_libs.py
```diff
@@ -126,6 +127,24 @@ def _get_python_lib():
 
     return python_lib
 
+
+# On Windows we generate import libraries that contains the code required to
+# load the DLL (.a) based on module definition files (.def)
+def _gen_import_libs(def_dir):
+
+    if os.name == "nt":
```
eric-wieser: I think this should be done by compiler name, perhaps - this seems like a mingw-specific thing. I'll look at numpy to try and work out how to do that.

cocotb/_build_libs.py
```diff
@@ -261,76 +280,58 @@ def get_ext():
...
-        icarus_vpi_ext = _get_vpi_lib_ext(
-            include_dir=include_dir,
-            share_lib_dir=share_lib_dir,
-            sim_define="ICARUS",
-            extra_lib=icarus_extra_lib,
-            extra_lib_dir=icarus_extra_lib_path,
-        )
-
-        ext.append(icarus_vpi_ext)
+        icarus_extra_lib = ["icarus"]
```
eric-wieser: I'm confused, aren't you still wanting to load `libvpi` and not `libicarus` at runtime?

themperek: No. Never needed on posix.

cocotb/_build_libs.py
themperek: At the moment it will not work with anything else. I think if we add MSVS than one can think about this.

eric-wieser: Fair.

This should still go in the `build_ext` subclass though, otherwise it will run even if I'm not trying to build anything.

cocotb/_build_libs.py
```diff
@@ -0,0 +1,446 @@
...
+
+    cfg_vars = distutils.sysconfig.get_config_vars()
+
+    if sys.platform == "darwin":
+        cfg_vars["LDSHARED"] = cfg_vars["LDSHARED"].replace("-bundle", "-dynamiclib")
+
+    share_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "share"))
+    share_lib_dir = os.path.relpath(os.path.join(share_dir, "lib"))
+    share_def_dir = os.path.abspath(os.path.join(share_dir, "def"))
+    include_dir = os.path.relpath(os.path.join(share_dir, "include"))
```
eric-wieser: If you move these lines out of the function, then you'll be able to use `share_def_dir` in the `build_ext` subclass.

cocotb/_build_libs.py
themperek: The dll name is in the def file.

eric-wieser: Ah, that's what I missed - thanks!

themperek:
Anything more that should be still done?

cocotb/share/def/aldec.def
```diff
@@ -0,0 +1,61 @@
+LIBRARY "aldecpli.dll"
+EXPORTS
```
eric-wieser: I think this directory could do with a readme.

cocotb/share/def/README.md
```diff
@@ -0,0 +1,3 @@
+
+This directory contains module-definition files (.def). On Windows are used to generate import libraries (.a).
```
eric-wieser:
```suggestion
This directory contains module-definition files (`.def`).
On Windows these are used to generate import libraries (`.a`, but not regular static libraries), which means we can compile our libraries without having the simulator binaries available.
They are not needed on other platforms

More details on `.def` files can be found at <some msdn link we mentioned earlier>.
```

themperek:
@imphil It should be better now. The only required file during compilation is `mti.h`. I have updated commit messages as much as I can. Is this OK? 
I will work out potential issues with smaller PRs that will follow.

eric-wieser:
> File an issue to explore run-time dynamic linking to get rid of the simulator-specific *.def files.

I don't think we want to do this. Run-time linking is far more work than managing a `.def` file.

We could look at generating the def files from lists of vpi and vhpi API functions, rather than having repetition between simulators

cocotb/_build_libs.py
```diff
@@ -0,0 +1,447 @@
...
+        # icarus requires vpl extension, gpivpi is default in Makefiles
+        if "icarus" in filename:
+            filename_short = filename_short.replace("libvpi.so", "gpivpi.vpl")
+            filename_short = filename_short.replace("libvpi.dll", "gpivpi.vpl")
+
+        return filename_short
+
+    # Add extra library_dirs path
+    def finalize_options(self):
+        _build_ext.finalize_options(self)
```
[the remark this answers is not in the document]
eric-wieser: No `self` if you use super.

cocotb/_build_libs.py
```diff
@@ -0,0 +1,447 @@
...
+
+    return python_lib
+
+# TODO Add "-Wconversion" to _base_warns when fixed
+_base_warns = ["-Wall", "-Wextra", "-Wcast-qual", "-Wwrite-strings"]
+_cc_warns = _base_warns + ["-Wstrict-prototypes", "-Waggregate-return"]
+_ccx_warns = _base_warns + ["-Wnon-virtual-dtor", "-Woverloaded-virtual"]
+
+_extra_cc_compile_args = ["-std=gnu99"] + _cc_warns
+_extra_cxx_compile_args = ["-std=c++11"] + _ccx_warns
```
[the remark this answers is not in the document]
eric-wieser: I think setuptools handles that for us?

cocotb/_build_libs.py
```diff
@@ -0,0 +1,447 @@
...
+    # Needed for Windows to not assume python module (generate interface in def file)
+    def get_export_symbols(self, ext):
+        return None
+
+    # For proper cocotb library nameing, based on https://github.com/cython/cython/issues/1740
+    def get_ext_filename(self, ext_name):
+        """
+        Like the base class method, but removes the ``.cpython-36m-x86_64-linux-gnu.`` part before the extension.
+
+        Also replaces ``.pyd`` with ``.dll`` on windows.
```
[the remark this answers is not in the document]
eric-wieser: `pyd` is only used on Windows in the first place.

eric-wieser:
> * As a first step, please file an issue to get rid of all compile-time dependencies in the common libraries.

I think this is a non-starter, because the `vpi.h` header varies by simulator (see the ifdefs there).

I think the only progress to be made here is the prefixing you suggest.

eric-wieser:
> * Instead, use `build_clib` and/or a custom command in setup.py.

I tried this in #955, but don't remember whether there was a problem. Worth re-exploring as a follow-up, anyway.

themperek:
>Building normal libraries through the setuptools mechanisms for Python extension modules is highly confusing (as if the library situation in cocotb wasn't already confusing enough) and leads to ugly code, e.g. rewriting the file name in build_ext. Instead, use build_clib and/or a custom command in setup.py. As usual in Python packaging, it's mostly undocumented and a mess, but I've seen examples around that look reasonable.

https://github.com/cocotb/cocotb/issues/1387

themperek:
>Building libraries multiple times. As Eric suggested, we instead should build libraries once. Ideally, all of them go into a single directory, and simulator-specific variants are named something like libvpi_xxx.so. The goal should be to use a single library for all simulators as long as possible, and only use specific variants if runtime conditionals are impossible. As a first step, please file an issue to get rid of all compile-time dependencies in the common libraries.

I have tied already but it looks quite involved (many files will change) but I will try. It would make this PR even more difficult if one does it now. 

Related to : https://github.com/cocotb/cocotb/issues/1181

themperek:
>ModelSim: Does the current code work for 32 bit ModelSim on 64 bit Windows? (Which is common if the free ModelSim version is used, e.g. the one distributed by Altera/Intel.) (I'm happy to not support this configuration, but we should be clear about it.)

It does not work for me but the reason is Makefiles itself (see #1099)  and not this PR.  
With this PR + #1099 it works fine for me both 32 and 64bit version.

themperek:
>This PR makes it impossible to run cocotb out of the installation directory without installing it, a (deprecated) mode which we still supported up to now. This change is OK, but needs a release note entry.

As I see understand this is not the case actually. Our tests still use run "out of the installation directory". With one adjustment (https://github.com/cocotb/cocotb/pull/1306/files#diff-b67911656ef5d18c4ae36cb6741b7965R33). If one does "pip install -e` this should not be needed.

cocotb/_build_libs.py
```diff
@@ -0,0 +1,447 @@
...
+
+        head, tail = os.path.split(filename)
+        tail_split = tail.split(".")
+
+        filename_short = os.path.join(head, tail_split[0] + "." + _get_lib_ext_name())
+
+        # icarus requires vpl extension, gpivpi is default in Makefiles
+        if "icarus" in filename:
+            filename_short = filename_short.replace("libvpi.so", "gpivpi.vpl")
+            filename_short = filename_short.replace("libvpi.dll", "gpivpi.vpl")
```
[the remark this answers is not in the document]
themperek: Part of #1181

cocotb/_build_libs.py
```diff
@@ -0,0 +1,447 @@
...
+    if os.name == "posix":
+        logger.info("Compiling libraries for VCS")
+        ext += _get_common_lib_ext(include_dir, share_lib_dir, sim_define="VCS")
+        vcs_vpi_ext = _get_vpi_lib_ext(
+            include_dir=include_dir, share_lib_dir=share_lib_dir, sim_define="VCS"
+        )
+        ext.append(vcs_vpi_ext)
+
+    #
+    # Aldec
```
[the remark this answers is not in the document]
themperek: This follow `SIM` naming convention which is not consistent.

imphil:
> As I understand this is not the case actually. Our tests still use run "out of the installation directory". With one adjustment (#1306 (files)). If one does "pip install -e` this should not be needed.

The old method of using cocotb relied on the COCOTB environment variable and including a Makefile, and didn't involve any kind of "pip install" at all. This mode isn't supported any more. (Which, again, is fine. But needs to be documented.)

cocotb/_build_libs.py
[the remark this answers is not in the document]
imphil: Please keep issues small and actionable. This is a simple consistency change which we can do without blocking it on a wider discussion. I filed https://github.com/cocotb/cocotb/issues/1389.

imphil:
Filed #1390 and #1392 to track open work items. GitHub PRs and issues are awful in threaded discussions, please file new issues and keep the discussion there for orthogonal topics.

themperek:
I have field https://github.com/cocotb/cocotb/issues/1393. I can implement any but need to decide the way.

themperek:
Can I ignore `LGTM analysis: C/C++` and marge. There is no C++ code touched here and this seems to fail randomly here.

eric-wieser:
Lgtm analysis is failing because it can no longer build - I think we ought to fix that before merging, or at least prove that it's transient

themperek:
@imphil This Lgtm pops up back and forth with this PR (even if nothing changed). I have no idea why.

themperek:
So not it is something different. It looks like LGTM for C++ analysis is trying to install cocotb with python2. This will fail since we use `super()`. 

It looks like there is a: https://discuss.lgtm.com/t/c-c-extraction-fails-with-python-2-7-error/2366/3

themperek:
@eric-wieser This looks like yours? Why now?

```
  In file included from cocotb/share/lib/vhpi/VhpiImpl.cpp:28:0:
    cocotb/share/lib/vhpi/VhpiImpl.h:192:9: error: ‘int VhpiSignalObjHdl::set_signal_value_binst(std::__cxx11::string&)’ marked ‘override’, but does not override
         int set_signal_value_binst(std::string &value) override;
             ^
    cocotb/share/lib/vhpi/VhpiImpl.cpp: In member function ‘GpiObjHdl* VhpiImpl::create_gpi_obj_from_handle(vhpiHandleT, std::__cxx11::string&, std::__cxx11::string&)’:
    cocotb/share/lib/vhpi/VhpiImpl.cpp:404:86: error: invalid new-expression of abstract class type ‘VhpiSignalObjHdl’
                 new_obj = new VhpiSignalObjHdl(this, new_hdl, gpi_type, is_const(new_hdl));
                                                                                          ^
    In file included from cocotb/share/lib/vhpi/VhpiImpl.cpp:28:0:
    cocotb/share/lib/vhpi/VhpiImpl.h:173:7: note:   because the following virtual functions are pure within ‘VhpiSignalObjHdl’:
     class VhpiSignalObjHdl : public GpiSignalObjHdl {
           ^
    In file included from cocotb/share/lib/vhpi/VhpiImpl.h:31:0,
                     from cocotb/share/lib/vhpi/VhpiImpl.cpp:28:
    cocotb/share/lib/vhpi/../gpi/gpi_priv.h:181:17: note: 	virtual int GpiSignalObjHdl::set_signal_value_binstr(std::__cxx11::string&)
         virtual int set_signal_value_binstr(std::string &value) = 0;
                     ^
```

eric-wieser:
`set_signal_value_binst(` is missing an r...

themperek:
@imphil @eric-wieser It should be ready now. Please marge if this is OK.

imphil:
Thanks a lot @themperek for pushing this through. Cocotb is better because of your dedication to such changes. I hope we were also able to learn a bit how to most effectively get large changes like developed and reviewed, so that follow-ups will be easier.

Again, thanks!

eric-wieser:
Not something to worry about in this PR, but we need to be careful that making the setup.py python3-only doesn't prevent python2 users getting a suitable error message if they try to install.

themperek:
@eric-wieser @cmarqu @imphil @garmin-mjames Thank you all for help. This was a bit harder than usually ;-). 

One idea for future that could help for changes like this is to make a dedicated branch where we can contribute directly. Not should what to do with commit history later on.

themperek:
> Not something to worry about in this PR, but we need to be careful that making the setup.py python3-only doesn't prevent python2 users getting a suitable error message if they try to install.

Yes. The problems seems to be `super()` here https://github.com/cocotb/cocotb/blob/master/cocotb/_build_libs.py#L39. This is imported before `setup()` checks python version? Maybe should add in `_build_libs.py` something like?

```python
if sys.version_info[0] < 3:
    raise Exception("cocotb package requies Python 3")
```

eric-wieser:
That check would need to go in setup.py, because otherwise it might already be too late (since cocotb is imported before _build_libs)

PR #1310, opened 2020-01-21
Pass PYTHON_BIN to lib builds

mariushegele:
A custom configured Python binary path is respected in the `Makefile.pylib.$(OS)` files, but not passed along to the library builds, therefore lost and set to the default `python`:

`PYTHON_BIN?=python`


themperek:
I maybe not aware of something but why one would need `PYTHON_BIN` when building libraries? As I can see it is not used there.

mariushegele:
As far as I see, `cocotb` requires Python to have been compiled with the `--enable-shared` flag. I am working in an environment where I can't modify `usr/local/` so I wanted to provide the locally compiled Python using `PYTHON_BIN`, but I just realized that this could also be done using a virtual environment.

themperek:
I can only recommend using conda: https://conda.io/en/latest/miniconda.html

PR #1312, opened 2020-01-21
Use -v93 switch for Cadence simulators

cmarqu:
Both ``sample_module`` and ``array_module`` use VHDL-93 features.


tests/designs/array_module/Makefile
```diff
@@ -42,6 +42,9 @@ COCOTB?=$(WPWD)/../../..
 ifeq ($(TOPLEVEL_LANG),verilog)
     VERILOG_SOURCES = $(COCOTB)/tests/designs/array_module/array_module.sv
 else ifeq ($(TOPLEVEL_LANG),vhdl)
+    ifeq ($(SIM),$(filter $(SIM),ius xcelium))
+       SIM_ARGS += -v93
```
eric-wieser: nit: indent seems off here (and below)
```suggestion
        SIM_ARGS += -v93
```

cmarqu: Right, fixed now. I copied it from a makefile where the local convention was three spaces.

Issue #1278, opened 2020-01-02
adder_module test doesn't run properly


eric-wieser:
From a random CI run:
```
/c/iverilog/bin/iverilog -o sim_build/sim.vvp -D COCOTB_SIM=1 -s adder -g2012   C:/src/examples/adder/tests/../hdl/adder.v
PYTHONPATH=/c/src/build/libs/x86_64:/c/src/examples/adder/tests:/:/C/c/srcC/src/examples/adder/model:/C/src/examples/adder/tests:/C/miniconda3/python37.zip:/C/miniconda3/DLLs:/C/miniconda3/lib:/C/miniconda3:/C/miniconda3/lib/site-packages:/C/miniconda3/lib/site-packages/cocotb-1.2.0-py3.7.egg:/C/miniconda3/lib/site-packages/win32:/C/miniconda3/lib/site-packages/win32/lib:/C/miniconda3/lib/site-packages/Pythonwin PATH=/c/Windows/system32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0:/c/Windows/System32/OpenSSH:/c/Users/ContainerAdministrator/AppData/Local/Microsoft/WindowsApps:/c/miniconda3:/mingw-w64/bin:/usr/bin:/bin:/c/miniconda3/Scripts:/c/miniconda3/bin:/c/miniconda3/condabin:/c/iverilog/bin:/c/iverilog/gtkwave/bin:/c/iverilog/bin:/c/src/build/libs/x86_64 MODULE=test_adder \
        TESTCASE= TOPLEVEL=adder TOPLEVEL_LANG=verilog COCOTB_SIM=1 \
        /c/iverilog/bin/vvp -M /c/src/build/libs/x86_64 -m gpivpi sim_build/sim.vvp   
     -.--ns INFO     cocotb.gpi                                  gpi_embed.c:78   in set_program_name_in_venv        Did not detect Python virtual environment. Using system-wide Python interpreter
     -.--ns INFO     cocotb.gpi                                GpiCommon.cpp:104  in gpi_print_registered_impl       VPI registered
     0.00ns INFO     cocotb.gpi                                  gpi_embed.c:340  in embed_sim_init                  Running on Icarus Verilog version 10.1 (stable)
     0.00ns INFO     cocotb.gpi                                  gpi_embed.c:341  in embed_sim_init                  Python interpreter initialized and cocotb loaded!
[...]

PR #1314, opened 2020-01-22
Fail if the `results.xml` file was not created

eric-wieser:
Should address https://github.com/cocotb/cocotb/issues/1278#issuecomment-570257200


themperek:
Should one remove exisitng `results.xml` before starting simulator?

eric-wieser:
Hmm, I suppose strictly yes, it should.

eric-wieser:
Ok, here's one failure:

https://travis-ci.org/cocotb/cocotb/jobs/640371381#L2785

themperek:
One can maybe do this one line here: https://github.com/cocotb/cocotb/blob/599b21ba453aa5a047b24dec80f2c475523105cc/cocotb/share/makefiles/Makefile.sim#L79 (not individually for the simulator)

BTW: the `rm` does not work would need to set COCOTB_RESULTS_FILE as an absolute path somehow https://github.com/cocotb/cocotb/blob/599b21ba453aa5a047b24dec80f2c475523105cc/cocotb/share/makefiles/Makefile.sim#L41

eric-wieser:
> One can maybe do this one line here

I thought about that, but it would still mean that `make results.xml` would claim to succeed even when it doesn't, and that other targets that depend on `$(COCOTB_RESULTS_FILE)` will be lied to.

eric-wieser:
I'm not sure I understand the relative paths problem - everything seems to work just fine as is

themperek:
>I'm not sure I understand the relative paths problem - everything seems to work just fine as is

This `rm` does nothing https://github.com/cocotb/cocotb/blob/599b21ba453aa5a047b24dec80f2c475523105cc/cocotb/share/makefiles/Makefile.sim#L41

Since the results.xml is relative to simulator `pwd` (ex. sim_build) and not makefile `pwd`

eric-wieser:
> and not makefile pwd

the makefile is included, so the makefile pwd in `.sim` is the same as the pwd of the test makefile, for instance `cocotb/tests/test_cases/test_exit_error`.

The `test` commands are working just fine, so I can't see why the `rm`  command wouldn't.

eric-wieser:
Remaining failures:

* https://travis-ci.org/cocotb/cocotb/jobs/640394194#L4566 (#1296)
* https://travis-ci.org/cocotb/cocotb/jobs/640394199#L4112 (#1278)

themperek:
Maybe I got things wrong but why this is not enough (not in the individual Makefiles)?

https://github.com/cocotb/cocotb/blob/599b21ba453aa5a047b24dec80f2c475523105cc/cocotb/share/makefiles/Makefile.sim#L79

eric-wieser:
Because that line only runs if i run `make sim`. If I run `make results.xml`, then that line is completely skipped.

themperek:
>Because that rm only runs if i run make sim. If I run make results.xml, then that line is completely skipped.

Yes. Thank you!

eric-wieser:
Just #1278 left to solve, then we can put this in

eric-wieser:
Thanks for helping out here @themperek!

PR #1317, opened 2020-01-22
Fix adder example on windows-ci #1278

themperek:


examples/adder/tests/Makefile
```diff
@@ -30,19 +30,12 @@
...
-endif
+PYTHONPATH := $(PWD)/../model:$(PYTHONPATH)
 
 ifeq ($(TOPLEVEL_LANG),verilog)
-    VERILOG_SOURCES = $(WPWD)/../hdl/adder.v
+    VERILOG_SOURCES = $(PWD)/../hdl/adder.v
 else ifeq ($(TOPLEVEL_LANG),vhdl)
-    VHDL_SOURCES = $(WPWD)/../hdl/adder.vhdl
+    VHDL_SOURCES = $(PWD)/../hdl/adder.vhdl
 else
```
eric-wieser: Based on another Makefile

https://github.com/cocotb/cocotb/blob/3c7616a2267b58321c04e75f9e16e836b33e7cf6/examples/mixed_language/tests/Makefile#L16-L32

these lines should not be changing - only the `PYTHONPATH` should

themperek: I prefer less than more if it does work. Actully Windows is an issue I do not know maybe in mingw one needed this but for msys2 it works OK with less code. Tried to start to clean this in #1099 ...

Should I revert?

eric-wieser: I think for this PR, we should try and keep these makefiles consistent. Let's just fix the PYTHONPATH in this one, and then in a follow-up you can try removing `WPWD` everywhere and see what breaks.

themperek: If https://github.com/cocotb/cocotb/pull/1306 would get it can clean up quite a bit.

examples/adder/tests/Makefile
```diff
@@ -29,11 +29,9 @@
 
 TOPLEVEL_LANG ?= verilog
 
-PWD=$(shell pwd)
-
 ifeq ($(OS),Msys)
 WPWD=$(shell sh -c 'pwd -W')
-PYTHONPATH := $(WPWD)/../model;$(PYTHONPATH)
+PYTHONPATH := $(WPWD)/../model:$(PYTHONPATH)
```
eric-wieser: I suspect this one _does_ need to be PWD

examples/adder/tests/Makefile
```diff
@@ -33,10 +33,10 @@ PWD=$(shell pwd)
 
 ifeq ($(OS),Msys)
 WPWD=$(shell sh -c 'pwd -W')
-PYTHONPATH := $(WPWD)/../model;$(PYTHONPATH)
+PYTHONPATH := $(PWD)/../model;$(PYTHONPATH)
```
eric-wieser: You also need the change from `;` to `:` here (eg, pull `PYTHONPATH` out of the `if`)

themperek: Actully I did look here: https://github.com/cocotb/cocotb/blob/master/examples/endian_swapper/tests/Makefile#L38

eric-wieser:
Looks good! https://travis-ci.org/cocotb/cocotb/jobs/640508224#L4076

PR #1319, opened 2020-01-22
Add target 'make help'

### LINKED ISSUE #1318: Add "make help"
Our Makefiles should have a target ``help`` that shows short documentation about supported targets and variables.

cmarqu:
Closes #1318.

### DIFF
--- a/cocotb/share/makefiles/Makefile.sim
+++ b/cocotb/share/makefiles/Makefile.sim
@@ -30,6 +30,7 @@
 # This is a simple wrapper Makefile to pull in the appropriate Makefile for
 # the desired simulator and set up common paths for the sims to use to build with
 
+.PHONY: all
 all: sim
 
 # NOTE: make sure to keep "helpmsg" aligned with documentation/source/building.rst
--- a/cocotb/share/makefiles/Makefile.sim
+++ b/cocotb/share/makefiles/Makefile.sim
@@ -32,6 +32,68 @@
 
 all: sim
 
+# NOTE: make sure to keep "helpmsg" aligned with documentation/source/building.rst
+# Also keep it at 80 chars.
+define helpmsg =
+Targets
+-------
+sim                       Unconditionally re-run the simulator (default)
+regression                Run simulator when dependencies have changes
+clean                     Remove build and simulation artefacts
+help                      This help text
+
+Make Variables
+--------------
+GUI                       Set this to 1 to enable the GUI mode in the simulator
+SIM                       Selects which simulator Makefile to use
+WAVES                     Enable wave traces dump for Riviera-PRO and Questa
+VERILOG_SOURCES           A list of the Verilog source files to include
+VHDL_SOURCES              A list of the VHDL source files to include
+VHDL_SOURCES_<lib>        VHDL source files to include in *lib* (GHDL only)
+COMPILE_ARGS              Arguments to pass to compile stage of simulation
+SIM_ARGS                  Arguments to pass to execution of compiled simulation
+EXTRA_ARGS                Arguments for compile and execute phases
+PLUSARGS                  Plusargs to pass to the simulator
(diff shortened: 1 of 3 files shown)


ktbarrett:
There is a `make help` in the root level `Makefile`. Should that be updated?

themperek:
@ktbarrett I would rather be in favour to move top level makefile to tests folder.

cmarqu:
> There is a `make help` in the root level `Makefile`. Should that be updated?

I hadn't noticed that, but it is a target that only helps for working with/testing cocotb itself (but indeed can use an update).
The help target of this PR is shown when you are developing your own testbench.

cocotb/share/makefiles/Makefile.sim
```diff
@@ -30,8 +30,71 @@
 # This is a simple wrapper Makefile to pull in the appropriate Makefile for
 # the desired simulator and set up common paths for the sims to use to build with
 
+.PHONY: all
 all: sim
 
+# NOTE: make sure to keep "helpmsg" aligned with documentation/source/building.rst
+# Also keep it at 80 chars.
```
eric-wieser: Keeping this aligned sounds like a lot of effort - can we just bundle this RST file with the install, and read it in here?

cmarqu: "Aligned" does not mean "duplicate", the text in ``helpmsg`` is shortened to be the essence of the description for people who want to refresh their memory and may leave out details.
I do not expect this to change often anymore, after all, we want to get rid of Makefiles at some point. Once we have them replaced with Python scripts, we can use ``sphinx-argparse`` to get the argument descriptions into the documentation.

cmarqu:
> There is a `make help` in the root level `Makefile`. Should that be updated?

This has been done in the meantime.

cocotb/share/makefiles/Makefile.sim
```diff
@@ -30,8 +30,71 @@
 # This is a simple wrapper Makefile to pull in the appropriate Makefile for
 # the desired simulator and set up common paths for the sims to use to build with
 
+.PHONY: all
```
imphil: That's not really related to this PR, it should be a separate commit at least.

cocotb/share/makefiles/Makefile.sim
```diff
@@ -30,8 +30,71 @@
...
+COCOTB_ATTACH             Pause time value in seconds before the simulator start
+COCOTB_ENABLE_PROFILING   Performance analysis of the Python portion of cocotb
+COCOTB_HOOKS              Comma-separated module list to be executed before test
+COCOTB_LOG_LEVEL          Default logging level (default INFO)
+COCOTB_RESOLVE_X          How to resolve X, Z, U, W on integer conversion
+COCOTB_SCHEDULER_DEBUG    Enable additional output of coroutine scheduler
+COVERAGE                  Report Python coverage (also HDL for some simulators)
+MEMCHECK                  HTTP port to use for debugging Python memory usage
+
+    For details, see https://cocotb.readthedocs.io/en/latest/building.html
```
imphil: If you want to be fully correct use a versioned link here (e.g. https://cocotb.readthedocs.io/en/v1.3.0/). But that might be too much work for little gain.

cocotb/share/makefiles/Makefile.sim
cmarqu: Done.

cocotb/share/makefiles/Makefile.sim
cmarqu: I may do this in a future PR. If the output of ``cocotb-config -v`` contains "dev", point to "latest", otherwise derive from the returned version.

cmarqu:
> If you really want command-line accessible documentation, we can think about building man pages from our docs using Sphinx.

I like that idea and will open a new issue for it. It doesn't solve my "problem" of wanting immediately useful reminders of what a certain variable was called again though.

PR #1320, opened 2020-01-23
Avoid recursing if get_sim_time logs

eric-wieser:
If `get_sim_time` tries to log, we enter an infinite recursion.

Ideally, we would cut this short immediately - but as an interim measure, let's at least recover gracefully when we cause a python RecursionError.

Found in #1313, but tangential to the original question.


imphil:
@eric-wieser do you want us to take this? Looks like a reasonably clean safety net to me to take it.

eric-wieser:
I have a better plan for this PR, will rebase imminently.

Issue #1323, opened 2020-01-24
Errors compiling VHPI and FLI code


cmarqu:
With #1137 in, I'm getting warnings-made-errors when compiling a VHDL design (shown here with gcc 9.2.0).
I have turned them back to be warnings to be able to show them all below:
```
[...]
g++ -fno-common -g -DDEBUG -fpic -std=c++11 -DVHPI_CHECKING -DIUS -Wall -Wextra -Wconversion -Wcast-qual -Wwrite-strings -Wnon-virtual-dtor -Woverloaded-virtual  -c -I/mypath/cocotb/cocotb/share/include -I/mypath/2019.10-python3.7/Linux-x86_64/include/python3.7m  -o /mypath/cocotb/tests/test_cases/test_external/build/obj/x86_64/VhpiImpl.o VhpiImpl.cpp
In file included from VhpiImpl.cpp:28:
VhpiImpl.h:210:5: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
  210 |     const vhpiEnumT chr2vhpi(const char value);
      |     ^~~~~
VhpiImpl.h:202:17: warning: ‘virtual int VhpiSignalObjHdl::set_signal_value(double)’ was hidden [-Woverloaded-virtual]
  202 |     virtual int set_signal_value(const double value);
      |                 ^~~~~~~~~~~~~~~~
VhpiImpl.h:228:9: warning:   by ‘VhpiLogicSignalObjHdl::set_signal_value’ [-Woverloaded-virtual]
  228 |     int set_signal_value(std::string &value);
      |         ^~~~~~~~~~~~~~~~
VhpiImpl.cpp: In member function ‘virtual void VhpiImpl::get_sim_precision(int32_t*)’:
VhpiImpl.cpp:100:27: warning: conversion from ‘uint64_t’ {aka ‘long unsigned int’} to ‘double’ may change value [-Wconversion]
  100 |     double base = 1e-15 * femtoseconds;
      |                           ^~~~~~~~~~~~
[...]

PR #1326, opened 2020-01-25
Remove -Werror

ktbarrett:
Addresses some issues in #1323. Removed the `-Werror` flag from the build scripts. This flag should only every be passed by CI system for compliance.

### DIFF
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -72,7 +72,7 @@ export INCLUDES := -I$(COCOTB_SHARE_DIR)/include -I$(PYTHON_INCLUDEDIR)
 LIB_EXT    := so
 PY_EXT     := so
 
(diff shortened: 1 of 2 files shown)


themperek:
I think tox will not forward any variable if not specified in `tox.ini` see: 
https://github.com/cocotb/cocotb/blob/master/tox.ini#L5

Actually `tox.ini` may babe a better place for those variables.

cmarqu:
I defer to @themperek on the CI setup.

ktbarrett:
Added the flags to tox instead.

eric-wieser:
Can we prove these are making it to a compilation? (Eg by checking a ci log)

cmarqu:
@eric-wieser  https://travis-ci.org/cocotb/cocotb/jobs/642035037#L619 is an example.

eric-wieser:
Thanks @ktbarrett. Regarding warnings, #1149 is somewhat related.

PR #1329, opened 2020-01-25
Building the C libraries in regression for warning compliance

ktbarrett:
Addresses some issues in #1323. Working towards #1330.

Currently I have updated the makefiles to make building just the C libraries easier. If you want to run `COCOTB_FLI_LIB` you will have to specify `MODELSIM_BIN_DIR`, though calculating this should probably be moved to `Makefile.inc`.

I am wondering how to integrate this into the regression tests. Ideally compilation of every library with every SIM define using both the most recent version of both GCC and Clang should be checked. After that we could remove the `-Werror` defines in the Python tests and only enable them in the compilation checks. This is a good number of tests (10 * 2) and might get bigger. Should this be a Travis matrix? Can you use the matrix and the specific jobs simultaneously? Should it instead be a Python or makefile?

EDIT:
It cannot be a Travis matrix, we would need to expand on two `env`s which is not currently supported.


themperek:
@ktbarrett How do you see this https://github.com/cocotb/cocotb/pull/1306 in this context. It will build all libraries that it can every time. We could add a clang build (not sure if on mac is not default already)

ktbarrett:
@themperek I just looked at your PR, I've been busy. It won't support how I envisioned I could use it, but there was mention of a `cocotb-build-libs` command, which would be perfect.

ktbarrett:
I am going to wait for #1306 and a refactor that makes it a standalone script before I go about adding clean compilation compliance to Travis.

ktbarrett:
Added another tox config file just for building the C and C++ libraries. It is a separate file because I could not get matrix expansion to work in the `tag` field in `[testenv:tag]` to segregate the two types of tests. Unfortunately, we can't test the compilation of the FLI automatically.

ktbarrett:
This is built on top of #1326.

eric-wieser:
Needs to rebase, since this repository does not allow merge commits in your previous PRs.

eine:
> Should this be a Travis matrix? Can you use the matrix and the specific jobs simultaneously?

GitHub Actions seems to be a much better fit for such a large set of tests. In ghdl/docker we moved from Travis to GHA precisely because you can have multiple workflows (equivalent to `.travis.yml`) in a single repository.

ktbarrett:
This now builds on top of #1306 instead of makefile changes. On-hold until that PR is in.

eric-wieser:
That PR is in, so this now needs a rebase.

themperek:
Not sure if this is needed anymore. We check all of this now with `tox/_build_libs`. I think we cannot do anything with `fli` without `mti.h`.

PR #1332, opened 2020-01-25
Draft: Support a "tail-less" cocotb

mballance:
It's often useful to test aspects of GUI applications without the graphical portion. We call this "headless" testing. There are times that it is useful to unit-test library code against the cocotb API without running an HDL simulator. I'm terming this "tail-less" testing, since there is no real-simulator back-end.

This draft PR shows the concept of providing this capability. I found this quite helpful when starting development of uvm-py, and believe it will continue to be useful while contributing to uvm-python. I would like to get feedback on whether this is something cocotb will accept before completing documentation, etc.

Please see https://github.com/mballance/cocotb/blob/tail-less/tests/test_cases/test_sim_standalone/test_sim_standalone.py for an example of using cocotb in standalone mode.

The key changes are as follows:
- Move initialization when running under a simulator into the _initialise_testbench method
- Provide an initialize_exec_context method that is called when using cocotb standalone
- Provide a SimulatorBase class that implements the API provided by the 'simulator' module
- Provide a SimulatorStandalone class that implements a basic timewheel for timed callbacks
- Change files that import the 'simulator' module to reference the global 'simulator' variable in the cocotb package


cmarqu:
Nice, that sounds like an implementation of what I had in mind in #1080.

cocotb/env_info.py
```diff
@@ -0,0 +1,25 @@
...
+
+class EnvInfo():
+    """Collects information about the environment in which cocotb is running"""
+    
+    def __init__(self):
+        self.sim_name    = "unknown"
+        self.sim_version = "unknown"
+        self.argv = []
+    
+    def set_sim_name(self, name):
```
cmarqu: These could become ``@property`` and ``@<attr>.setter``: https://stackoverflow.com/questions/6618002/using-property-versus-getters-and-setters

eric-wieser: This seems like Java leaking into python to me - all these getters and setters aren't useful, let's just omit them.

ktbarrett:
I started to review this, but I stopped because I did not understand the premise of this change at all. What is the point exactly? How you've implemented it makes smoketesting actual testbenches (that utilize triggers and simulator value accesses) not possible. Is the point simply to use the scheduler without a simulator? If that's the case we should improve the scheduler quality to the point where it is standalone rather than work around the poor quality by adding a bunch of new code.

mballance:
> I started to review this, but I stopped because I did not understand the premise of this change at all. What is the point exactly? How you've implemented it makes smoketesting actual testbenches (that utilize triggers and simulator value accesses) not possible. Is the point simply to use the scheduler without a simulator? That would be really useful. But, if that's the case we should improve the scheduler quality to the point where it is standalone rather than work around the poor quality by adding a bunch of new code.

Hi Kaleb,
    My objective with this change is to be able to unit testbench library code written against the cocotb API purely within Python, and without running in the context of a simulator. For my purposes, the testbench code needs to interact via Event, Lock, and Timer, and does not need to interact with signals. My unit tests do need to be able to specify the command-line arguments and, ideally, the simulator name and version. 
    So, I think the answer is a resounding yes to your question "Is the point simply to use the scheduler without a simulator?".
    If the primary concern is about the amount of code I've added, I'll be happy to take another pass at this to focus more on exposing the scheduler as something that can be used standalone. I gave it some thought today, and believe I see a path to doing that in a more focused way. 

Best Regards,
Matthew

cocotb/__init__.py
```diff
@@ -54,11 +53,92 @@
...
+
+log = None
+
+regression_manager = None
+
+plusargs = {}
+"""A dictionary of "plusargs" handed to the simulation."""
+
+# To save typing provide an alias to scheduler.add
+fork = None
```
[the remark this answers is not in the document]
mballance: Like this! I've added it to __init__.py

cocotb/__init__.py
```diff
@@ -54,11 +53,92 @@
...
+
+loggpi = None
+
+# FIXME is this really required?
+_rlock = threading.RLock()
```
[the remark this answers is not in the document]
mballance: Good deal, made the move

cocotb/scheduler.py
```diff
@@ -325,7 +326,10 @@ def _test_completed(self, trigger=None):
                 self.log.debug("Issue test result to regression object")
 
             # this may scheduler another test
-            cocotb.regression_manager.handle_result(test)
+            # Note: the regression_manager will be None when cocotb is 
+            # running in 'standalone' (no-simulator) mode
+            if cocotb.regression_manager is not None:
+                cocotb.regression_manager.handle_result(test)
```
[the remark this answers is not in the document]
mballance: I'm happy to go with the 'lighter' refactor first. I've added an option __init__ parameter to the Scheduler this is the handle_result callback. I definitely agree that removing the global refs is a good move!

mballance:
Appreciate the feedback, Kaleb! I'm good with this approach of allowing standalone use of the cocotb components for unit testing, while not necessarily making that a first-class cocotb feature. 

-Matthew

ktbarrett:
I think the last thing before making this ready for review is a bit of documentation. What's a `simulator stub` object and what's required of it, what pieces of the `cocotb` module do you need to fill out, what's the intended use case, advantages, and limitations, etc...

mballance:
Of course, Kaleb. Now that we've got an approach worked out, adding docs makes sense. I do have a couple of questions there, and figure it might just be better to ask ;-).
- Is there an existing example of documenting an API that doesn't exist in Python? The simulator API is implemented in C, so I can't just add doc comments to document the available/required methods for the simulator stub.
- Is there a recommended placed to put 'use case' documentation? The use case of using the cocotb elements for unit testing is different enough from the standard cocotb use model that I assume the docs for that should be kept separate from docs for the standard use model.

Thanks,
Matthew

cmarqu:
> * Is there an existing example of documenting an API that doesn't exist in Python? The simulator API is implemented in C, so I can't just add doc comments to document the available/required methods for the simulator stub.

You can use Doxygen annotations such as https://github.com/cocotb/cocotb/blob/d56e4c855bb1c9ed0f2620add9e8d6c86f66f0fb/cocotb/share/lib/fli/FliImpl.cpp#L243-L247
which are showing up like this (via https://breathe.readthedocs.io/): https://cocotb.readthedocs.io/en/latest/generated/class/classFliImpl.html

> * Is there a recommended placed to put 'use case' documentation? The use case of using the cocotb elements for unit testing is different enough from the standard cocotb use model that I assume the docs for that should be kept separate from docs for the standard use model.

You could just start with a new ``*.rst`` file (or multiple ones if needed) and we can then plug them into the existing documentation later.

cocotb/utils.py
```diff
@@ -74,6 +69,13 @@ def get_sim_time(units=None):
 
     return result
 
+def get_precision():
+    global _LOG_SIM_PRECISION
+
+    if _LOG_SIM_PRECISION is None:
+        _LOG_SIM_PRECISION = cocotb.simulator.get_precision()  # request once and cache
+    return _LOG_SIM_PRECISION
```
imphil: This change seems a generally useful improvement, how about doing a separate PR for this right away to keep this patch stack small?

mballance:
Appreciate the input, @imphil. I'll certainly work on adding documentation. Actually, having a simulator base class would simplify the documentation task.
I would propose to create a simulator base class (eg SimulatorBase) that defines all APIs currently provided by the simuatormodule, with implementations that throw a NotImplementedError exception. Doing so will allow users to implement useful subsets of the API (eg just the time-based event APIs) while being notified if unimplemented APIs are called.

Looking at the PR-breakdown process, I would propose the following. Please let me know if this makes sense:
- One PR to replace _LOG_SIM_PRECISION in 'utils' with a call to get_precision
- One PR for the changes that move initialization  inside the _initialise_testbench method. This will impact __init__.py, utils.py, and triggers.py due the change in the way the simulator module is referenced
- One PR for the changes to the scheduler to accept a 'handle_result' callable instead of directly calling the regression manager.
- One PR adding an "abstract" SimulatorBase class and documentation for the simulator API.

I'd much appreciate it if you can point me to an example of using git rebase to do this.

Thanks,
Matthew

eric-wieser:
Isn't this proposed "simulator interface" just the gpi abstraction we already have, but at the python level?

To that end, I can see two other paths forward:

* A `libgpipython`, which provides a `GpiImpl` that just calls python callbacks directly, rather than actually interfacing to a simulator
* Using pybind11 to directly expose the `GpiImpl` object to python, rather than the somewhat different wrapper used by the `simulator` module.

mballance:
Hi @eric-wieser, the cocotb `simulator` module (share/lib/simulator/simulatormodule.c) defines Python methods that are implemented via GPI. The proposal here is to allow Python implementations of the API provided by the `simulator` extension module.

I think there are several paths that could be taken here -- though they're generally also larger projects.
- I think we could largely do away with the C++ GPI and interface directly to simulators using ctypes. We would still need a tiny stub in C to accept initialization calls from the simulators, but I believe everything else could be done directly in Python. 
- We likely could implement the `simulator` module directly on top of the GPI API. In that case, I'm not entirely sure the `simulator` API would go away, but would be Python implemented.

I think we could definitely do more ambitious projects in the future. My hope is that taking a small step now to make the cocotb library more modular will give us short-term benefits, and possibly enable these larger changes in the future.

eric-wieser:
cc @themperek, who mentioned a possible application using sockets for a remote python runtime.

mballance:
> cc @themperek, who mentioned a possible application using sockets for a remote python runtime.

That's an application I hadn't considered, but seems quite reasonable. Another application (separate from the primary one described above) is allowing execution-environment integrations (eg SystemC) to initially be developed as extensions to cocotb.

themperek:
There is also an idea to use pybind11 on GPI and make a simulator module this way. Can make a multiple once depending on application.

ktbarrett:
I think this implementation could be simplified even further. I am working on changes to gpi_embed, the simulator module and the code in `__init__.py` to reduce the "talking" between them. Once that's in, running with an alternative "tail" to cocotb is as straight-forward as providing an interface-compatible implementation of the `simulator` module and running cocotb as a script.

That brings up concerns about #1540 potentially complicating that.

eric-wieser:
@ktbarrett: #1540 itself, or the conversations about the interface between them? As written, #1540 makes it no harder to mock out the simulator than it was before.

mballance:
> There is also an idea to use pybind11 on GPI and make a simulator module this way. Can make a multiple once depending on the application. (If understand right).

That's an interesting idea to reduce the amount of required Pyhton-specific native integration code.

mballance:
> I think this implementation could be simplified even further. I am working on changes to gpi_embed, the simulator module and the code in `__init__.py` to reduce the "talking" between them. Once that's in, running with an alternative "tail" to cocotb is as straight-forward as providing an interface-compatible implementation of the `simulator` module and running cocotb as a script.
> 
> That brings up concerns about #1540 potentially complicating that.

@ktbarrett, I'm sure this proposal can be simplified further (most can). What is your thinking on how to select the 'simulator' module that provides the back-end implementation? Hopefully we can come up with something more flexible than a fixed name.

ktbarrett:
@mballance I think the fixed name is what we should do. cocotb internally calls out `simulator.thing` in a lot of places. Supporting alternative names adds a layer of indirection and doesn't seem (to me) to offer that much, while reducing perf.

It would run like so:
 * make a Python module named `simulator` that implements some or all of the interfaces on the module.
 * add a `__main__.py` to cocotb
 * run cocotb: `MODULE=... OTHER_VARS=... PYTHONPATH=/path/to/simulator/module/ python cocotb`

eric-wieser:
I should point out that as of recently (#1425), we import `cocotb.simulator` not `simulator`.

mballance:
> @mballance I think the fixed name is what we should do. cocotb internally calls out `simulator.thing` in a lot of places. Supporting alternative names adds a layer of indirection and doesn't seem (to me) to offer that much, while reducing perf.
> 
> It would run like so:
> 
> * make a Python module named `simulator` that implements some or all of the interfaces on the module.
> * add a `__main__.py` to cocotb
> * run cocotb: `MODULE=... OTHER_VARS=... PYTHONPATH=/path/to/simulator/module/ python cocotb`

Hmm... Needing to play the order of PYTHONPATH always concerns me a bit...

mballance:
Sorry, accidentally closed instead of commenting.

mballance:
> I should point out that as of recently (#1425), we import `cocotb.simulator` not `simulator`.

@eric-wieser, do you know of any data about the relative performance of referencing a method in a module vs a method in a class handle? I didn't see any after a cursory check...

eric-wieser:
@mballance: Sure, in the former case you need to perform a type check that the first argument is what you expected, but in the latter case it's guaranteed to have the right type so no check is needed.

ktbarrett:
> I should point out that as of recently (#1425), we import `cocotb.simulator` not `simulator`.

That means we *will* have to add additional infrastructure to cocotb to support it loading a mocked out simulator module. More code, more problems.

eric-wieser:
Whoops, misclick due to github glitching out

eric-wieser:
> > I should point out that as of recently (#1425), we import `cocotb.simulator` not `simulator`.
> 
> That means we _will_ have to add additional infrastructure to cocotb to support it loading a mocked out simulator module. More code, more problems.

The move was not a particularly principled one, but was done primarily because:
* `simulator` is essentially private and undocumented API
* `simulator` is a very generic name, and not a good one to consume from the global module namespace, especially if we're not even going to advertise that we did it.

There's no reason we can't move it again.

themperek:
>run cocotb: MODULE=... OTHER_VARS=... PYTHONPATH=/path/to/simulator/module/ python cocotb

>That means we will have to add additional infrastructure to cocotb to support it loading a mocked out simulator module. More code, more problems.

I think we can (worst case) manipulate the path/name from the python side. It is typically much easier and less error-prone than any `PYTHONPATH` etc. (especially in the multi-os environment)

jahagirdar:
I am trying to evolve a Spec to RTL Python modeling framework. And I believe this PR might be the missing link between my untimed models and timed models.

I am describing my usecase below in the hope that it will result in a expanding the PR to cover it:

1. From a client I get say, a 100 line python code(Golden Model) and  test data file. 
2. I take the code and start breaking it down into lowlevel component's of bus arbiters, cache controllers, lookup tables, Memory controller, DDR access logic etc. I check consistency of my implementation interface with golden model
3. Till now everything is untimed. I have a dozen pytest based unit test.
4. The next stage is to play around with different algorithms for caching, sorting etc... This means introducing a concept of time
5. For a loosely timed model, I would prefer using the scheduler from cocotb and use the Timer/RisingEdge method. to simulate the approximate time taken by a function.
6. At this stage we have the algo selection and hardware partition ready and RTL Development can start.
7. In parallel the loosely timed model transforms into a bit accurate, cycle accurate model for RTL Verification...

This PR seems to enables me to use cocotb at Step 5 without drastically changing my Python code base or rewriting my tests. I am at step 5 currently so I will try to pull this PR, play around with it and provide my feedback in a couple of days.

tests/test_cases/test_sim_standalone/test_sim_standalone.py
```diff
@@ -0,0 +1,194 @@
...
+
+from unittest.case import TestCase
+import cocotb
+import logging
+from cocotb.log import SimLog
+from cocotb.triggers import Event, Timer
+from cocotb.scheduler import Scheduler
+from cocotb.utils import get_sim_time
+
+class TestSimStandalone(TestCase):
```
jahagirdar, line 15: I am assuming  `TestSimStandalone` class is a common class that is provided by this PR and the cocotb user will only be instantiating it. Am I correct? Or does every user need to rewrite this class as per his usecase?

eric-wieser:
> 4\. This means introducing a concept of time

In the past I've done this with something like:
```
def wait(ns):
    if in_cocotb:
        @cocotb.function
        async def _wait():
            await Timer(ns)
        _wait()
    else:
        time.sleep(ns * 1e9)
```
This means the same code can be used for python-only simulation, HDL simulation, and hardware drivers.

tests/test_cases/test_sim_standalone/test_sim_standalone.py
```diff
@@ -0,0 +1,194 @@
...
+
+        cocotb.simulator = sim
+        cocotb.scheduler = Scheduler()
+        cocotb.fork = cocotb.scheduler.add
+        cocotb.log = SimLog("cocotb")
+        cocotb.argv = []
+        cocotb.process_plusargs()
+        cocotb.log.setLevel(logging.INFO)
+
+        cocotb.scheduler.add_test(test_main(mb))
```
jahagirdar, line 180: I would assume that these set of lines are also hidden from the end user and he only needs to call something like.
if standalone_sim:
  set_simulator(STANDALONE)
else:
  set_simulator(DUT) # Default case?

tests/test_cases/test_sim_standalone/test_sim_standalone.py
mballance: TestSimStandalone is just the unit test class. Each use-case would be expected to provide an implementation of the APIs in the class SimulatorStub below. The API implemented by the class SimulatorStub (eg register_timed_callback) is currently implemented by the C/C++ Python module that integrates with the simulator. 
My understanding of the consensus was to document the APIs implemented by that module (eg register_timed_callback), but not provide a class implementing those APIs as part of cocotb.

eric-wieser: > My understanding of the consensus was to document the APIs implemented by that module (eg register_timed_callback), but not provide a class implementing those APIs as part of cocotb.

I'd perhaps relax this to "not provide a class implementing those APIs _except as controlling an attached simulator_". I'm not opposed to swapping out the way we use the `simulator` module for a `Simulator` class of some kind.

ktbarrett: Some detail on that in https://github.com/cocotb/cocotb/pull/1987#issuecomment-659945859.

PR #1333, opened 2020-01-25
Document new make variables

### LINKED ISSUE #1322: Document Makefile variables passed to C/C++ compiler
https://github.com/cocotb/cocotb/commit/7e48440155e1e589620e8465889c229db582c14f renamed and improved some variables we pass to compilers. We need to document those, and add a release notes entry for renamed items.

ktbarrett:
Adds documentation for user-facing make variables. Fixes #1322.

### DIFF
--- a/cocotb/share/makefiles/Makefile.rules
+++ b/cocotb/share/makefiles/Makefile.rules
@@ -58,10 +58,10 @@ CC  ?= cc
 CXX ?= c++
 
 $(LIB_OBJ_DIR)/%.o: %.c
-	$(CC) $(C_ARGS) $(CC_WARNS) $(CFLAGS) -c $(INCLUDES) -o $@ $<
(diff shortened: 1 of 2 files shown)


cmarqu:
https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html says 
> ``LOADLIBES`` is a deprecated (but still supported) alternative to ``LDLIBS``.
so maybe we don't need this one? It's unlikely a user has this set in their environment and expects us to honor it.

ktbarrett:
Thanks for the catch. I had never heard of `LOADLIBES`, but I saw it on the list and decided to add it. I guess never having heard of it should have been a sign.

cmarqu:
I was actually googling for it because I thought it was a typo :)

cmarqu:
Wielding my new power for the first time, thanks!

PR #1334, opened 2020-01-26
Add 'active' field to RunningCoroutine class

mballance:
The SystemVerilog 'process' class, which represents threads, provides a 'self()' function that returns the currently-running process. I've added a class field named 'active' to the RunningCoroutine class to provide the equivalent for cocotb.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -86,7 +86,12 @@ class RunningCoroutine(object):
 
         coro.kill() will destroy a coroutine instance (and cause any Join
         triggers to fire.
+        
+        RunningCoroutine.active -- currently-running coroutine
     """
+    
+    active = None # RunningCoroutine
+
     def __init__(self, inst, parent):
         if hasattr(inst, "__name__"):
             self.__name__ = "%s" % inst.__name__
@@ -149,6 +154,9 @@ class RunningCoroutine(object):
             CoroutineComplete: If the coroutine returns or throws an error, self._outcome is set, and
            :exc:`CoroutineComplete` is thrown.
         """
+
+        RunningCoroutine.active = self
+        
         try:
             self._started = True
             return outcome.send(self._coro)
--- a/documentation/source/coroutines.rst
+++ b/documentation/source/coroutines.rst
@@ -150,7 +150,33 @@ they'd naturally end.
         # NOTE: isclose is a Python 3.5+ feature
         if not isclose(edge_time_ns, start_time_ns + 4.0):
             raise TestFailure("Expected a period of 4 ns")
+            
+The currently-running coroutine can always be accessed via the ``active`` 
+member of the RunningCoroutine class.
 
+.. code-block:: python3
+
+    @cocotb.test()
+    def test_count_edge_cycles(dut, period=1, clocks=6):
+        print("Note: the active coroutine is: %s" % str(RunningCoroutine.active))
+        cocotb.fork(Clock(dut.clk, period, units='ns').start())
+        yield RisingEdge(dut.clk)
+
+        timer = Timer(period + 10)
+        task = cocotb.fork(count_edges_cycles(dut.clk, clocks))
+        count = 0
+        expect = clocks - 1
+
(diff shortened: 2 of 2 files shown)


eric-wieser:
I'm not a huge fan of this idea - I'd rather move away from `RunningCoroutine` objects existing at all. What's the application of having this object available?

mballance:
UVM uses the active thread/coroutine for a couple of things. The one I'm currently looking at is that it collects thread/coroutine handles such that threads can be waited for or shut-down at the appropriate time. The location where the active thread is stored is central and separated from the location where the thread is started.

Is there a different and better location/manner in which the currently-running coroutine could be obtained?

eric-wieser:
Can you give a concrete use-case as pseudo-code, beyond printing the function name which you can already get through `inspect`?

mballance:
> Can you give a concrete use-case as pseudo-code, beyond printing the function name which you can already get through `inspect`?

Of course, Eric. I assume we're specifically talking about providing access to the active coroutine, and not about having coroutine handles at all. 
- An implementation of thread-specific data: 
    This approach is used somewhat commonly with SystemVerilog and other threading APIs that don't support thread-specific data directly. Take, for example, a randomization library that wishes to ensure random stability within each thread. This means, essentially, that each thread needs its own random state that changes based on randomizations within the thread and is not impacted by randomizations in other threads. The randomization library could would use the active-coroutine handle as a hash key to access the coroutine-specific randomization state. 
    As I mentioned, supporting thread-specific data would be an alternative approach to providing access to the active thread in this case.

-  Enabling a library to monitor and control threads
    Say we have a library that our testbench code (all running within coroutines) can call to announce the beginning and completion of operations. This library provides a global timeout on operation duration, at which point it will kill any coroutines whose operations haven't completed yet. The code within the coroutines has the most-complete knowledge of when operations begin and complete, so it's in the best position to call this library. However, the code in the coroutines can't tell the library what thread its behavior corresponds to (the code that forked the coroutine does, but the coroutine itself does not). 


Hope this helps a bit.      

Best Regards,
Matthew

eric-wieser:
> An implementation of thread-specific data:
This approach is used somewhat commonly with SystemVerilog and other threading APIs that don't support thread-specific data directly.

You need to be careful here, in that what you're considering a thread is not equivalent to a cocotb coroutine. For instance, something like `yield subroutine()` is not "starting a new thread" in any meaningful sense - only `fork` does that (and the functions that secretly call it like `triggers.First`...).

If the goal is to have task-specific state, then the [`contextvars`](https://docs.python.org/3/library/contextvars.html) module is probably the best way to acheive that.
Unfortunately, getting that integrated into the scheduler is probably a fair bit of work, and probably not something we want to attempt until we can further simplify what's already there.

> Take, for example, a randomization library that wishes to ensure random stability within each thread. This means, essentially, that each thread needs its own random state that changes based on randomizations within the thread and is not impacted by randomizations in other threads. The randomization library could would use the active-coroutine handle as a hash key to access the coroutine-specific randomization state.

Most random number APIs have decided that implicit RNG state is a bad thing - for this type of application, it's best to just pass around `np.random.Generator` objects explicitly, so you know exactly which coroutines touch which state.

> Enabling a library to monitor and control threads
Say we have a library that our testbench code (all running within coroutines) can call to announce the beginning and completion of operations. This library provides a global timeout on operation duration, at which point it will kill any coroutines whose operations haven't completed yet.

This is the example I don't really follow, and need some psuedo-code to understand.

Perhaps the `asyncio` cancellation model is relevant here - the approach would be something like
```python
async def my_routine():
    try:
        await Timer(1000)
    except CancelledError:
        # the manager cancelled me!

async def my_main_test(dut):
    task = MyWackyManager().spawn(my_routine(), timeout=100)
    await task
```

mballance:
Hi Eric,
    Thanks for your responses. I'm fine dropping this request for now, and re-raising it if and when a really (really) compelling usecase arises that we're not able to address via other means. 

Best Regards,
Matthew

Issue #1330, opened 2020-01-25
Enforcing compilation of libraries without warnings


ktbarrett:
#1137 added more warnings to the compilation process in an attempt to increase the C code quality and locate potential bugs (which it did and they were fixed in the common code and VPI). It also added the ability for the user to pass standardized compilation flags to the build process. Now that we are detecting more warnings, it's time we fix them, and once they are fixed, we should maintain the code quality with compliance checks.

We should add to travis compilation tests of all the libraries (save FLI which requires non-free headers) for every simulator with both GCC and Clang. The current makefiles were not made with this in mind so some work is necessary to add support.

eric-wieser:
Is there anything left to be done here? I'm not sure if we build with clang yet, but we do build everything but FLI in CI now.

Please close if you think there's nothing left to do.

themperek:
On osx clang is default as far as I am aware.

ktbarrett:
Do we build with `-Werror`? When I run tox locally, the build spits out warnings about C++ not supporting some warning flag.

themperek:
>Do we build with -Werror?

Yes, we do. It sees this is a different kind of warning.

themperek:
@ktbarrett Can we close this?

PR #1335, opened 2020-01-26
Fix VHPI warnings

ktbarrett:
Addresses part of #1323 by fixing code that warns in the VHPI interface. Built on top of #1329. Removes the expected fails in that PR.

### DIFF
--- a/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
+++ b/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
@@ -27,6 +27,11 @@
 
 #include <assert.h>
 #include "VhpiImpl.h"
+#include <limits>
+
+namespace {
+    using bufSize_type = decltype(vhpiValueT::bufSize);
+}
 
 extern "C" void handle_vhpi_callback(const vhpiCbDataT *cb_data);
 
@@ -268,7 +273,7 @@ int VhpiSignalObjHdl::initialise(std::string &name, std::string &fq_name) {
         case vhpiStrVal: {
             m_indexable = true;
             m_num_elems = vhpi_get(vhpiSizeP, handle);
-            m_value.bufSize = (m_num_elems)*sizeof(vhpiCharT) + 1;
+            m_value.bufSize = (m_num_elems)*static_cast<bufSize_type>(sizeof(vhpiCharT) + 1);
             m_value.value.str = (vhpiCharT *)malloc(m_value.bufSize);
             m_value.numElems = m_num_elems;
             if (!m_value.value.str) {
@@ -290,7 +295,7 @@ int VhpiSignalObjHdl::initialise(std::string &name, std::string &fq_name) {
     }
 
     if (m_num_elems) {
-        m_binvalue.bufSize = m_num_elems*sizeof(vhpiCharT) + 1;
+        m_binvalue.bufSize = m_num_elems*static_cast<bufSize_type>(sizeof(vhpiCharT)) + 1;
         m_binvalue.value.str = (vhpiCharT *)calloc(m_binvalue.bufSize, sizeof(vhpiCharT));
 
         if (!m_binvalue.value.str) {
@@ -338,7 +343,7 @@ int VhpiLogicSignalObjHdl::initialise(std::string &name, std::string &fq_name) {
     if (vhpi_get(vhpiKindP, query_hdl) == vhpiArrayTypeDeclK) {
         m_indexable = true;
         m_value.format = vhpiLogicVecVal;
-        m_value.bufSize = m_num_elems*sizeof(vhpiEnumT);
+        m_value.bufSize = m_num_elems*static_cast<bufSize_type>(sizeof(vhpiEnumT));
         m_value.value.enumvs = (vhpiEnumT *)malloc(m_value.bufSize + 1);
         if (!m_value.value.enumvs) {
             LOG_CRITICAL("Unable to alloc mem for write buffer: ABORTING");
@@ -350,7 +355,7 @@ int VhpiLogicSignalObjHdl::initialise(std::string &name, std::string &fq_name) {
     }
 
     if (m_num_elems) {
-        m_binvalue.bufSize = m_num_elems*sizeof(vhpiCharT) + 1;
+        m_binvalue.bufSize = m_num_elems*static_cast<bufSize_type>(sizeof(vhpiCharT) + 1);
         m_binvalue.value.str = (vhpiCharT *)calloc(m_binvalue.bufSize, sizeof(vhpiCharT));
 
         if (!m_binvalue.value.str) {
@@ -442,7 +447,7 @@ error:
 }
 
 // Value related functions
-const vhpiEnumT VhpiSignalObjHdl::chr2vhpi(const char value)
+vhpiEnumT VhpiSignalObjHdl::chr2vhpi(const char value)
 {
     switch (value) {
         case '0':
@@ -550,7 +555,7 @@ int VhpiSignalObjHdl::set_signal_value(long value)
         case vhpiLogicVecVal: {
             int i;
             for (i=0; i<m_num_elems; i++)
-                m_value.value.enumvs[m_num_elems-i-1] = value&(1L<<i);
+                m_value.value.enumvs[m_num_elems-i-1] = value&(1L<<i) ? vhpi1 : vhpi0;
 
             // Since we may not get the numElems correctly from the sim and have to infer it
             // we also need to set it here as well each time.
@@ -561,17 +566,32 @@ int VhpiSignalObjHdl::set_signal_value(long value)
 
         case vhpiLogicVal:
         case vhpiEnumVal: {
-            m_value.value.enumv = value;
+            using EnumLimits = std::numeric_limits<vhpiEnumT>;
+            if ((value > EnumLimits::max()) || (value < EnumLimits::min())) {
+                LOG_ERROR("Data loss detected");
+                return -1;
+            }
+            m_value.value.enumv = static_cast<vhpiEnumT>(value);
             break;
         }
 
         case vhpiIntVal: {
-            m_value.value.intg = value;
+            using IntLimits = std::numeric_limits<vhpiIntT>;
+            if ((value > IntLimits::max()) || (value < IntLimits::min())) {
+                LOG_ERROR("Data loss detected");
+                return -1;
+            }
+            m_value.value.intg = static_cast<vhpiIntT>(value);
             break;
         }
 
         case vhpiCharVal: {
-            m_value.value.ch = value;
+            using CharLimits = std::numeric_limits<vhpiCharT>;
+            if ((value > CharLimits::max()) || (value < CharLimits::min())) {
+                LOG_ERROR("Data loss detected");
+                return -1;
+            }
+            m_value.value.ch = static_cast<vhpiCharT>(value);
             break;
         }
 
--- a/cocotb/share/lib/vhpi/VhpiImpl.cpp
+++ b/cocotb/share/lib/vhpi/VhpiImpl.cpp
@@ -92,13 +92,22 @@ void VhpiImpl::get_sim_time(uint32_t *high, uint32_t *low)
     *low = vhpi_time_s.low;
 }
 
+static int32_t log10int(uint64_t v)
+{
+    int32_t i = -1;
+    do {
+        v /= 10;
+        i += 1;
+    } while (v);
+    return i;
+}
+
 void VhpiImpl::get_sim_precision(int32_t *precision)
 {
     /* The value returned is in number of femtoseconds */
     vhpiPhysT prec = vhpi_get_phys(vhpiResolutionLimitP, NULL);
     uint64_t femtoseconds = ((uint64_t)prec.high << 32) | prec.low;
-    double base = 1e-15 * femtoseconds;
-    *precision = (int32_t)log10(base);
+    *precision = log10int(femtoseconds) - 15;
 }
 
 // Determine whether a VHPI object type is a constant or not
--- a/cocotb/share/lib/vhpi/VhpiImpl.h
+++ b/cocotb/share/lib/vhpi/VhpiImpl.h
@@ -109,9 +109,6 @@ public:
     }
 private:
     std::string initial_value;
-    bool rising;
-    bool falling;
-    VhpiSignalObjHdl *signal;
 };
 
 class VhpiTimedCbHdl : public VhpiCbHdl {
@@ -197,9 +194,9 @@ public:
     virtual double get_signal_value_real();
     virtual long get_signal_value_long();
 
-
-    virtual int set_signal_value(const long value);
-    virtual int set_signal_value(const double value);
+    using GpiSignalObjHdl::set_signal_value;
+    virtual int set_signal_value(long value);
+    virtual int set_signal_value(double value);
     virtual int set_signal_value(std::string &value);
 
     /* Value change callback accessor */
@@ -207,7 +204,7 @@ public:
     virtual int initialise(std::string &name, std::string &fq_name);
 
 protected:
-    const vhpiEnumT chr2vhpi(const char value);
+    vhpiEnumT chr2vhpi(char value);
     vhpiValueT m_value;
     vhpiValueT m_binvalue;
     VhpiValueCbHdl m_rising_cb;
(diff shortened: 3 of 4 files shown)


cocotb/share/lib/vhpi/VhpiImpl.h
```diff
@@ -224,6 +224,7 @@ class VhpiLogicSignalObjHdl : public VhpiSignalObjHdl {
 
     virtual ~VhpiLogicSignalObjHdl() { }
 
+    using GpiSignalObjHdl::set_signal_value;
```
ktbarrett: This will cause `set_signal_value(double)` to segfault. Previously, this caused an implicit conversion of a `double` argument` to a `long` and called that overload. I have a Issue/PR coming that will explain and fix the issues with the VHPI and FLI related to overloaded virtual functions.

cmarqu:
> Built on top of #1330.

On top of #1329 actually.

cocotb/share/lib/vhpi/VhpiImpl.cpp
```diff
@@ -97,7 +97,7 @@ void VhpiImpl::get_sim_precision(int32_t *precision)
     /* The value returned is in number of femtoseconds */
     vhpiPhysT prec = vhpi_get_phys(vhpiResolutionLimitP, NULL);
     uint64_t femtoseconds = ((uint64_t)prec.high << 32) | prec.low;
-    double base = 1e-15 * femtoseconds;
+    double base = 1e-15 * static_cast<double>(femtoseconds);
     *precision = (int32_t)log10(base);
```
eric-wieser: The compiler is rightly warning here - can we add a TODO, perhaps:
```
/* TODO: compute log10(base) as an integer, without losing precision by converting to double 
 * (or quantify whether this matters) */
```
Of course, if all the simulators do it this way anyway, then ignore this comment

cocotb/share/lib/vhpi/VhpiImpl.h
```diff
@@ -207,7 +207,7 @@ class VhpiSignalObjHdl : public GpiSignalObjHdl {
     virtual int initialise(std::string &name, std::string &fq_name);
 
 protected:
-    const vhpiEnumT chr2vhpi(const char value);
+    vhpiEnumT chr2vhpi(const char value);
```
eric-wieser:
```suggestion
    vhpiEnumT chr2vhpi(char value);
```
The other const is meaningless too

cocotb/share/lib/vhpi/VhpiImpl.h
```diff
@@ -197,7 +197,7 @@ class VhpiSignalObjHdl : public GpiSignalObjHdl {
     virtual double get_signal_value_real();
     virtual long get_signal_value_long();
 
-
+    using GpiSignalObjHdl::set_signal_value;
     virtual int set_signal_value(const long value);
     virtual int set_signal_value(const double value);
```
eric-wieser:
```suggestion
    virtual int set_signal_value(double value);
```

cocotb/share/lib/vhpi/VhpiImpl.h
```diff
@@ -197,7 +197,7 @@ class VhpiSignalObjHdl : public GpiSignalObjHdl {
     virtual double get_signal_value_real();
     virtual long get_signal_value_long();
 
-
+    using GpiSignalObjHdl::set_signal_value;
     virtual int set_signal_value(const long value);
```
eric-wieser:
```suggestion
    virtual int set_signal_value(long value);
```

cocotb/share/lib/vhpi/VhpiImpl.h
eric-wieser: Can you elaborate on why this causes a segfault? Is the issue that `GpiSignalObjHdl::set_signal_value` is pure-virtual? If so, isn't this a compiler error?

> Previously, this caused an implicit conversion of a double argument to a long and called that overload.

I don't agree with that - from what I can tell, overloads are not used to populate the v-table, so this doesn't actually matter for the way we use these objects.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -548,9 +549,13 @@ int VhpiSignalObjHdl::set_signal_value(long value)
     switch (m_value.format) {
         case vhpiEnumVecVal:
         case vhpiLogicVecVal: {
+            if ((value > INT_MAX) || (value < INT_MIN)) {
```
eric-wieser:
```suggestion
            using EnumLimits = std::numeric_limits<vhpiEnumT>;
            if (value > EnumLimits::max() || value < EnumLimits::min()) {
```

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -548,9 +549,13 @@ int VhpiSignalObjHdl::set_signal_value(long value)
...
+                LOG_ERROR("Data loss detected");
```
eric-wieser: Let's just `return -1` here, like we do below.

cocotb/share/lib/vhpi/VhpiImpl.cpp
ktbarrett: a flooring `log10` on an integer is easy enough to implement. All the other interfaces just ask the simulator "what's your precision"? But apparently VHPI deprecated that feature (???).

eric-wieser: Seems like a weird feature to me anyway, why restrict precision to powers of 20?

cocotb/share/lib/vhpi/VhpiImpl.h
ktbarrett: You are right, it's only an issue if we call it as a derived object.

eric-wieser:
I'm not hugely comfortable with the extensive makefile changes here, but appreciate the C++ changes. Once my review comments are addressed, I think I'll probably want to see a self-contained PR that just tidies the C++ code. We can merge that immediately to unblock local testing, and then think a little more about how to handle makefiles in this PR.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -548,9 +549,13 @@ int VhpiSignalObjHdl::set_signal_value(long value)
...
+            }
             int i;
+            vhpiEnumT value_as_enum = static_cast<vhpiEnumT>(value);
             for (i=0; i<m_num_elems; i++)
-                m_value.value.enumvs[m_num_elems-i-1] = value&(1L<<i);
+                m_value.value.enumvs[m_num_elems-i-1] = value_as_enum & vhpiEnumT(1 << i);
```
ktbarrett: Ias this ever correct? The more I look at it, the less I think it was.

ktbarrett:
I rebased this PR's commits onto master and off #1329. I think #1329 or similar should go in though.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
eric-wieser: What is `decltype(m_value.value.enumvs)` here?

ktbarrett: `vhpiEnumT*` aka `unit32_t*`. The issues I see is how the simulator interprets the values being fed to it. Can I assume the simulator does a 0/!0 check, and assumes you only ever mean bit? There are more than two logic values. I think this code sections also covers the case of vectors of user-defined enums that aren't guaranteed to be binary, so you end up feeding it garbage.

eric-wieser: You're right that this should at least be casting to `bool` before performing the assignment

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -548,9 +549,15 @@ int VhpiSignalObjHdl::set_signal_value(long value)
     switch (m_value.format) {
         case vhpiEnumVecVal:
         case vhpiLogicVecVal: {
+            using EnumLimits = std::numeric_limits<vhpiEnumT>;
+            if ((value > EnumLimits::max()) || (value < EnumLimits::min())) {
+                LOG_ERROR("Data loss detected");
+                return -1;
+            }
             int i;
+            vhpiEnumT value_as_enum = static_cast<vhpiEnumT>(value);
```
eric-wieser: This cast is probably wrong - since each bit is being  treated as a separate enum value, it's not correct to assume that all the bits fit in one.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -548,9 +549,15 @@ int VhpiSignalObjHdl::set_signal_value(long value)
...
             for (i=0; i<m_num_elems; i++)
-                m_value.value.enumvs[m_num_elems-i-1] = value&(1L<<i);
+                m_value.value.enumvs[m_num_elems-i-1] = value_as_enum & (vhpiEnumT(1) << i);
```
eric-wieser:
```suggestion
                m_value.value.enumvs[m_num_elems-i-1] = bool(value & (1L << i));
```

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -268,7 +269,7 @@ int VhpiSignalObjHdl::initialise(std::string &name, std::string &fq_name) {
         case vhpiStrVal: {
             m_indexable = true;
             m_num_elems = vhpi_get(vhpiSizeP, handle);
-            m_value.bufSize = (m_num_elems)*sizeof(vhpiCharT) + 1;
+            m_value.bufSize = (m_num_elems)*static_cast<int>(sizeof(vhpiCharT)) + 1;
```
eric-wieser: This cast is correct only on `IUS` - on other simulators, `bufSize` is `size_t`. Perhaps use the somewhat clumsy:
```suggestion
            m_value.bufSize = static_cast<decltype(m_value.bufSize)>(
                m_num_elems * sizeof(vhpiCharT) + 1
            );
```

ktbarrett: I would prefer to use that everywhere in my changes. Are we definitely not averse to usage of `decltype` and `auto`?

ktbarrett:
I'm still not an fan of the class hierarchy in the VHPI. `VhpiLogicSignalObjHdl` shouldn't derive from `VhpiSignalObjHdl`. Both should derive from a common ancestor with `set_signal_value` and `get_signal_value` being "not implemented" error functions. This means when we cast a `VhpiLogicSignalObjHdl` back to `GpiSignalObjHdl` we don't call `VhpiSignalObjHdll`'s very broad definition of that function.

I'm not even sure *why* there is a separate class for logic signals when the base class handles all cases through introspection.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
eric-wieser: Well, we need #1149 to go in first, to add `std=c++11` (or another reduced patch which does so).

ktbarrett: #1137 Already did that. We just don't have coding standards for newer C++ functionality.

eric-wieser: There are very few ways to write correct code here:
* Use `decltype`
* Use `#ifdef`, and repeat the logic in the `vhpi_user.h` header

I don't think standards are necessary here. Let's use `decltype` because it's the only tool appropriate for the job (and add a comment saying that the type of `bufSize` varies by simulator).

cocotb/share/lib/vhpi/VhpiImpl.cpp
```diff
@@ -92,13 +92,23 @@ void VhpiImpl::get_sim_time(uint32_t *high, uint32_t *low)
     *low = vhpi_time_s.low;
 }
 
+template <typename T>
+int log10int(T v)
```
eric-wieser:
```suggestion
static int log10int(T v)
```

Also, let's not template unnecessarily here. We call this in exactly one place, so let's just use `uint64_t`.

If we were templating this, it ought to be the stricter:
```C++
template<typename T, typename=std::enable_if<std::is_integral<T>::value>::type>
```

cocotb/share/lib/vhpi/VhpiImpl.cpp
```diff
@@ -92,13 +92,23 @@ void VhpiImpl::get_sim_time(uint32_t *high, uint32_t *low)
...
+}
+
 void VhpiImpl::get_sim_precision(int32_t *precision)
 {
     /* The value returned is in number of femtoseconds */
     vhpiPhysT prec = vhpi_get_phys(vhpiResolutionLimitP, NULL);
     uint64_t femtoseconds = ((uint64_t)prec.high << 32) | prec.low;
-    double base = 1e-15 * femtoseconds;
-    *precision = (int32_t)log10(base);
+    *precision = log10int(base) - 15;
```
eric-wieser: Where did `base` go?

cocotb/share/lib/vhpi/VhpiImpl.cpp
```diff
@@ -92,13 +92,23 @@ void VhpiImpl::get_sim_time(uint32_t *high, uint32_t *low)
...
+template <typename T>
+int log10int(T v)
+{
+    int i = -1;
+    do {
+        v /= 10;
+        i += 1;
+    } while (v);
+    return i;
+}
```
eric-wieser: Proof that this is a) correct, and b) fixes a bug:

https://godbolt.org/z/jcgrzN

cocotb/share/lib/vhpi/VhpiImpl.cpp
eric-wieser:
```suggestion
    *precision = log10int(femptoseconds) - 15;
```

ktbarrett: Whoops... Seems I'll need to rebase #1329 on top of this to test this interface.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
ktbarrett: For this and the related comments in this section I copied the implementation from `VhpiLogicSignalObjHdl::set_signal_value(long)` because it makes sense for this type of operation on this type of handle. I read through #333 and I'm still not sure what the original intent of this code was.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -550,28 +553,40 @@ int VhpiSignalObjHdl::set_signal_value(long value)
         case vhpiLogicVecVal: {
             int i;
             for (i=0; i<m_num_elems; i++)
-                m_value.value.enumvs[m_num_elems-i-1] = value&(1L<<i);
-
-            // Since we may not get the numElems correctly from the sim and have to infer it
-            // we also need to set it here as well each time.
```
eric-wieser: Any reason for removing this comment? Still looks relevant to me.

eric-wieser: (it describes the assignment to .numElems)

ktbarrett: I just copy-pasted the other implementation over it. I'll put it back in.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -268,7 +269,9 @@ int VhpiSignalObjHdl::initialise(std::string &name, std::string &fq_name) {
         case vhpiStrVal: {
             m_indexable = true;
             m_num_elems = vhpi_get(vhpiSizeP, handle);
-            m_value.bufSize = (m_num_elems)*sizeof(vhpiCharT) + 1;
+            m_value.bufSize = (m_num_elems)*static_cast<decltype(m_value.bufSize)>(
```
eric-wieser: We should probably introduce a typedef for this decltype, and use it for all of the size of casts below. Will come back to this with a suitable `using bufsize_t = decltype(...)` later.

eric-wieser:
Let's put this in - none of the casts are contentious, and I tested the new precision logic.

Even if this doesn't yet build without warnings in VHPI, this removes most of them

eric-wieser:
@ktbarrett, for next time a more explanatory commit message would be valuable.

PR #1338, opened 2020-01-26
Simplify traceback propagation

eric-wieser:
This commit removes all the workarounds that were necessary to propagate tracebacks on Python 2.7 and 3.4 - now that our minimum supported version is Python 3.5.

This reverts most of gh-1100, along with some other exc_info-related cleanup.

xref #1289


eric-wieser:
Rebase was fairly straightforward, and tests still pass - so let's put this in

PR #1339, opened 2020-01-26
Use syntax for keyword-only arguments

eric-wieser:
This avoids the need for popping kwargs from dictionaries.
It also leads to better documentation both in sphinx and from the `help` builtin.

This removes the last caller of `reject_remaining_kwargs`, although we leave the function around for now in case downstream users are using it.

### DIFF
--- a/cocotb/drivers/__init__.py
+++ b/cocotb/drivers/__init__.py
@@ -299,8 +299,7 @@ class ValidatedBusDriver(BusDriver):
             ``(valid, invalid)`` cycles to insert.
     """
 
-    def __init__(self, entity, name, clock, **kwargs):
-        valid_generator = kwargs.pop("valid_generator", None)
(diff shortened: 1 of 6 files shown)


cmarqu:
> This removes the last caller of `reject_remaining_kwargs`, although we leave the function around for now in case downstream users are using it.

Maybe add a comment there to that effect - some tool may report this function as being unused and in turn someone may remove it?

eric-wieser:
> some tool may report this function as being unused and in turn someone may remove it?

Such a tool would be broken - function names that don't start with `_` are usually considered public API.

ktbarrett:
Though you are leaving the function around just in case, is there any way to remove it from the documentation? And/or issue a `DeprecationWarning`?

cmarqu:
Removing the function from documentation should be possible with the ``exclude-members`` option in Sphinx.

eric-wieser:
Added a second commit to deprecate the function.

PR #1341, opened 2020-01-27
Remove references to Python 2.7 from docs

ktbarrett:
Addresses some points from #1289.

### DIFF
--- a/documentation/source/quickstart.rst
+++ b/documentation/source/quickstart.rst
@@ -10,11 +10,13 @@ Pre-requisites
 
 Cocotb has the following requirements:
 
-* Python 2.7, Python 3.5+ (recommended)
+* Python 3.5+
 * Python-dev packages
 * GCC 4.8.1+ or Clang 3.3+ and associated development packages
 * GNU Make
 * A Verilog or VHDL simulator, depending on your RTL source code
+
+.. versionchanged:: 1.4 Dropped Python 2 support
 
 Installation via PIP
 --------------------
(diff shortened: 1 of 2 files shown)


documentation/source/quickstart.rst
```diff
@@ -23,12 +23,6 @@ Installation via PIP
 
 Cocotb can be installed by running
 
-.. code-block:: bash
-
-    pip3 install cocotb
```
cmarqu: I'm not so sure about this one - the fact that we don't support Python 2 anymore does not mean that the user cannot have both versions installed still and needs to get the Python-3-pip via ``pip3``.

documentation/source/quickstart.rst
```diff
@@ -10,7 +10,7 @@ Pre-requisites
 
 Cocotb has the following requirements:
 
-* Python 2.7, Python 3.5+ (recommended)
+* Python 3.5+
```
imphil: Add a comment here like "(Use cocotb < 1.4 if you need Python 2.7 support.)

eric-wieser: Is there a permalink to the old docs we could point to as well?

documentation/source/quickstart.rst
ktbarrett: Perhaps I should remove the other leg then to remove any confusion?

documentation/source/quickstart.rst
imphil: yes: https://cocotb.readthedocs.io/en/v1.3.0/

ktbarrett: It's in the bottom left corner.

PR #1342, opened 2020-01-27
Remove Python 2 version conditions from C code

ktbarrett:
Addresses some points in #1289.


eric-wieser:
Conflicts with #1366, so let's put this in first. Rerunning CI just to be sure, since there were some CI improvements recently.

PR #1344, opened 2020-01-27
Document extension conventions

imphil:
This is a follow-up to #1141.

When fiddling with the wording in ##1141 I realized there are some changes I'd like to make, and I thought the easiest way of doing so is in a new PR. Thanks @Martoni for a great starting point, and for sticking with us through all of that discussion!

This PR keeps the basic content the same as agreed on in #1141, but modifies the structure slightly.
- It now documents cocotb extensions in general, not only bus extensions.
- The error handling bits go into a separate document, since they felt rather out of place in the extension documentation.

Please have a look and let me know what you think!

### DIFF
--- a/documentation/source/extensions.rst
+++ b/documentation/source/extensions.rst
@@ -1,0 +1,110 @@
+Writing cocotb extensions
+=========================
+
+This guide explains how to write cocotb extensions, with a focus on the conventions that should be followed.
+
+Cocotb gives its users a framework to build Python testbenches for hardware designs.
+But sometimes the functionality provided by cocotb is too low-level.
+One common example are bus drivers and monitors:
+instead of creating a bus adapter from scratch for each new project, wouldn't it be nice to share this component, and build on top it?
+In the verification world, such extensions are often called "verification IP" (VIP).
+
+In cocotb, such functionality can be packaged and distributed as extensions.
+Technically, cocotb extensions are normal Python packages, and all standard Python packaging and distribution techniques can be used.
+Additionally, the cocotb community has agreed on a set of conventions to make extensions easier to use and to discover.
+
+Naming conventions
+------------------
+
+Cocotb extensions are normal Python modules which follow these naming conventions.
+
+Assuming an extension named ``EXTNAME`` (all lower-case),
+
+- the package is in the ``cocotbext.EXTNAME`` namespace, and
+- the distribution (package) name is prefixed with ``cocotbext-EXTNAME``.
+
+Example:
+An SPI bus extension might be packaged as ``cocotbext-spi``, and its functionality would live in the ``cocotbext.spi`` namespace.
+The module can then be installed with ``pip3 install cocotbext-spi``, and used with ``import cocotbext.spi``.
+
+Types of cocotb extensions
+--------------------------
+
+For some types of cocotb extensions we have developed conventions which go beyond naming.
+These conventions help to achieve a consistent behavior across extensions of the same type.
+
+Bus extensions
+~~~~~~~~~~~~~~
+
+A cocotb extension which interacts with a bus or an interface (such as SPI or AXI) should build on top of a common set of classes to provide a uniform interface for its users.
+Typically, a bus extension provides three pieces of functionality:
+an abstraction of the bus itself, a bus driver, and a bus monitor.
+
+A bus driver is the "active" part, it drives the signals that make up the bus to request reads or writes from the bus.
+A bus monitor is the "passive" part, it observes signal changes on the bus and assigns meaning to them.
+Monitors can also check the bus behavior against a standard to ensure no invalid states are being observed.
+
+The signals which make up the bus should be grouped in a class inheriting from :any:`cocotb.bus.Bus`.
+Bus drivers should inherit from the :any:`cocotb.drivers.BusDriver` class.
+Bus monitors should inherit from the :any:`cocotb.monitors.BusMonitor` class.
+
+Packaging extensions
+--------------------
+
+To package a cocotb extension as Python package follow the :ref:`naming conventions <Naming conventions>`, and the `normal Python packaging rules <https://packaging.python.org/tutorials/packaging-projects/>`_.
+Extensions namespaced packages, implemented using the `native namespacing <https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages>`_ approach discussed in :pep:`420`.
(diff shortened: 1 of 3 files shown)


documentation/source/extensions.rst
```diff
@@ -0,0 +1,104 @@
...
+  from setuptools import setup
+
+  setup(name='cocotbext-EXTNAME',
+        version='0.1',
+        install_requires = ['cocotb'],
+        classifiers = [
+          "Programming Language :: Python :: 3",
+          "License :: OSI Approved :: MIT License",
+          "Operating System :: OS Independent",
+          "Topic :: Scientific/Engineering :: Electronic Design Automation (EDA)"])
```
eric-wieser: This is missing the rather crucial `PEP420PackageFinder.find_packages` call necessary to discover `cocotbext/EXTNAME/__init__.py`. I don't remember the exact spelling.

documentation/source/extensions.rst
```diff
@@ -0,0 +1,104 @@
...
+
+.. code-block: command
+
+  $ python3 -m pip install cocotbext-EXTNAME
+
+To use the functionality in the extension module, import it into your testbench.
+
+.. code-block:: python3
+
+  import cocotbext.EXTNAME
```
eric-wieser: Perhaps worth mentioning also `from cocotbext import EXTNAME` and `from cocotb.EXTNAME import MyVerificationClass`

documentation/source/extensions.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Cocotb extensions are normal Python modules which follow these naming conventions.
+
+Assuming an extension named ``EXTNAME`` (all lower-case),
+
+- the package is in the ``cocotbext.EXTNAME`` namespace, and
+- the distribution (package) name is prefixed with ``cocotbext-EXTNAME``.
+
+Example:
+A SPI bus extension might be packaged as ``cocotbext-spi``, and its functionality would live in the ``cocotbext.spi`` namespace.
```
cmarqu:
```suggestion
An SPI bus extension might be packaged as ``cocotbext-spi``, and its functionality would live in the ``cocotbext.spi`` namespace.
```

documentation/source/extensions.rst
```diff
@@ -0,0 +1,104 @@
...
+The signals which make up the bus should be grouped in a class inheriting from :any:`cocotb.bus.Bus`.
+Bus drivers should inherit from the :any:`cocotb.drivers.BusDriver` class.
+Bus monitors should inherit from the :any:`cocotb.monitors.BusMonitor` class.
+
+Packaging extensions
+--------------------
+
+To package a cocotb extension as Python package follow the `Naming conventions`_, and the `normal Python packaging rules <https://packaging.python.org/tutorials/packaging-projects/>`_.
+
+The module file hierarchy should be as following (replace ``EXTNAME`` with the name of the extension, e.g. ``spi``).
```
cmarqu:
```suggestion
The module file hierarchy should be as follows (replace ``EXTNAME`` with the name of the extension, e.g. ``spi``).
```
?

documentation/source/extensions.rst
```diff
@@ -0,0 +1,104 @@
...
+        install_requires = ['cocotb'],
+        classifiers = [
+          "Programming Language :: Python :: 3",
+          "License :: OSI Approved :: MIT License",
+          "Operating System :: OS Independent",
+          "Topic :: Scientific/Engineering :: Electronic Design Automation (EDA)"])
+
+With this file structure in place the cocotb extension can be installed through ``pip`` in development mode ::
+
+  $ python3 -m pip install -e .
```
cmarqu: Is there a reason not to write this as ``pip3 install -e .``?

eric-wieser: It's a nice way to be sure your pip matches your python, for users with lots of disjoint installations.

cmarqu: Makes sense, thanks for the clarification.

Martoni:
That is good for me. Thanks for the review and rewriting work.

documentation/source/extensions.rst
imphil: I've fixed that by explicitly stating that we're going for "native namespaced packaes", and added `find_namespace_packages()` as documented at https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages

imphil:
Updated with review comments, PTAL.

documentation/source/extensions.rst
```diff
@@ -0,0 +1,111 @@
...
+  # file structure of the cocotbext-EXTNAME repository
+  ├── cocotbext/
+  │   │   # No __init__.py here.
+  │   └── EXTNAME/
+  │       └── __init__.py
+  ├── README.md
+  └── setup.py
+
+The Python source code should go into the ``EXTNAME`` directory, next to the ``__init__.py`` file.
+All packaging metadata goes into ``setup.py``.
```
cmarqu: Nit: Using ``:file:`...` `` here would make it a bit more semantic.

documentation/source/extensions.rst
```diff
@@ -0,0 +1,111 @@
...
+  from setuptools import setup, find_namespace_packages
+
+  setup(name = 'cocotbext-EXTNAME',
+        version = '0.1',
+        packages = find_namespace_packages(include=['cocotbext.*']),
+        install_requires = ['cocotb'],
+        python_requires = '>=3.5',
+        classifiers = [
+          "Programming Language :: Python :: 3",
+          "License :: OSI Approved :: MIT License",
```
cmarqu: Do we want to express a requirement or (strong) recommendation that the extension be under the MIT License?

imphil: Not really. We'll most likely have requirements and recommendations for code we host in the cocotb org, and for things we endorse, but that's not yet discussed and shouldn't be in this document. I'll remove this classifier.

imphil:
Thanks @cmarqu, updated.

documentation/source/extensions.rst
eric-wieser: I'd lean towards leaving this as is, but add a comment saying `# or another license of your choice`. Otherwise we risk a bunch of extensions that people forget to license that no one can actually use.

imphil: The license would also need to go into the `license` field and a LICENSE File (or similar) is also needed. I opened https://github.com/cocotb/cocotb/issues/1362 to gather thoughts on recommended licenses (this discussion is probably larger than this small comment sub-thread), and I'd go for adding this information in a follow-up as soon as we have consensus in the issue.

PR #1346, opened 2020-01-28
Use InternalError rather than assertion

benreynwar:
…le outside scheduler.


cocotb/decorators.py
```diff
@@ -262,7 +262,8 @@ def _force_outcome(self, outcome):
         should be inlined into `abort` below, and the call in `unschedule`
         replaced with `abort(outcome.error)`.
         """
-        assert self._outcome is None
+        if self._outcome is not None:
+            raise cocotb.InternalError("Outcome already has a value, but is being set again.")
```
eric-wieser: Instead perhaps just add `from cocotb.scheduler import InternalError` at the top of the file?

cocotb/__init__.py
```diff
@@ -92,6 +92,9 @@
     if not sys.warnoptions:
         warnings.simplefilter("default")
 
+InternalError = scheduler.InternalError
+"""InternalError should be accessible outside scheduler."""
```
eric-wieser: Not sure this is necessary. No one is going to able to catch this anyway, so there's little point making it easier to get to

benreynwar: cocotb.scheduler is the instance of Scheduler defined in __init__.py.
Perhaps it makes more sense to rename scheduler.py to something else so we don't have this name collision?

eric-wieser: Not when you use the import keyword it isn't ;)

benreynwar: Didn't think of that :).

benreynwar: If we do `from cocotb.scheduler import InternalError' we get:
```ImportError: cannot import name 'InternalError' from partially initialized module 'cocotb.scheduler' (most likely due to a circular import) (/home/ben/Code/cocotb/cocotb/scheduler.py)```

Doing `from cocotb import scheduler` and then using `scheduler.InternalError` works fine.  But then we have `cocotb.scheduler` and `scheduler` both present in the decorators.py file, with one referring to the scheduler object, and the other referring to the module which is a bit smelly.

File: cocotb/decorators.py [hunk 3]
```diff
@@ -40,6 +40,7 @@
 from cocotb.utils import get_sim_time, lazy_property
 from cocotb import outcomes
 from cocotb import _py_compat
+from cocotb import scheduler
```
eric-wieser: Perhaps:
```suggestion
# renamed to avoid confusion with the scheduler instance `cocotb.scheduler`
from cocotb import scheduler as scheduler_module
```

cocotb/decorators.py
```diff
@@ -262,7 +263,8 @@ def _force_outcome(self, outcome):
         should be inlined into `abort` below, and the call in `unschedule`
         replaced with `abort(outcome.error)`.
         """
-        assert self._outcome is None
+        if self._outcome is not None:
+            raise scheduler.InternalError("Outcome already has a value, but is being set again.")
```
eric-wieser: Or alternatively:
```suggestion
            # imported here to avoid circular imports
            from cocotb.scheduler import InternalError
            raise InternalError("Outcome already has a value, but is being set again.")
```

cocotb/decorators.py
eric-wieser:
```suggestion
```

Issue #1348, opened 2020-01-29
'kill' doesn't work on a coroutine when the coroutine has a ReadOnly trigger


benreynwar:
`counter` is a counter that logs a counter every clock cycle.
`test_intermed` forks a `counter` and then kills it after 3 clock cycles.
`test_top` awaits two `test_intermed` one after the other.

If everything is working properly the outputs from the two counters should not overlap, however when a `trigger.ReadOnly()` is present in the `counter` the `kill` does not work properly and the output from the first counter overlaps with the second.

```python
@cocotb.coroutine
async def clock(dut):
    while True:
        dut.clk <= 0
        await triggers.Timer(1, units='ns')
        dut.clk <= 1
        await triggers.Timer(1, units='ns')

@cocotb.coroutine
async def counter(dut, name):
    for clock_index in range(6):
        await triggers.RisingEdge(dut.clk)
        await triggers.ReadOnly()
        logger.warning('%s: %d', name, clock_index)

@cocotb.coroutine
async def test_intermed(dut, name):
    coro = cocotb.fork(counter(dut, name))
    for i in range(3):
        await triggers.RisingEdge(dut.clk)
    coro.kill()

@cocotb.test()
async def test_top(dut):
    cocotb.fork(clock(dut))
    await test_intermed(dut, 'a')
    await test_intermed(dut, 'b')
```

```
     1.00ns WARNING  debug2                                        debug2.py:25   in coroutine1                      a: 0               
     3.00ns WARNING  debug2                                        debug2.py:25   in coroutine1                      a: 1               
     5.00ns WARNING  debug2                                        debug2.py:25   in coroutine1                      a: 2               
     7.00ns WARNING  debug2                                        debug2.py:25   in coroutine1                      b: 0               
     7.00ns WARNING  debug2                                        debug2.py:25   in coroutine1                      a: 3              
     9.00ns WARNING  debug2                                        debug2.py:25   in coroutine1                      b: 1              
     9.00ns WARNING  debug2                                        debug2.py:25   in coroutine1                      a: 4              
```

benreynwar:
I think what's going on is that in the main `for coro in scheduling:` loop in the `_event_loop`, we don't consider that these coroutines might kill each other while we're iterating through them.

ktbarrett:
You diagnosis is correct. This only occurs if both coros are waiting on the same trigger and the coro that kills is run before the one that is killed.

eric-wieser:
[...]

PR #1349, opened 2020-01-29
WIP: Raise KilledError in killed coroutines.

benreynwar:
…es have been killed in scheduler.

Currently the kill method of a coroutine can fail if a coroutine tries to kill another coroutine when the scheduler is looping over an iterable containing both coroutines.  Here I take the simple approach of adding a tag to the RunningCoroutine instance to mark it as killed.  The scheduler can then check this tag. This addresses issue #1348.

Presumably this approach problems for coroutines that are not instances of RunningCoroutine.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -166,15 +166,9 @@ class RunningTask(object):
 
     def kill(self):
         """Kill a coroutine."""
-        if self._outcome is not None:
-            # already finished, nothing to kill
-            return
-
         if _debug:
             self.log.debug("kill() called on coroutine")
-        # todo: probably better to throw an exception for anyone waiting on the coroutine
-        self._outcome = outcomes.Value(None)
-        cocotb.scheduler.unschedule(self)
+        cocotb.scheduler.kill(self)
 
     def join(self):
         """Return a trigger that will fire when the wrapped coroutine exits."""
--- a/cocotb/outcomes.py
+++ b/cocotb/outcomes.py
@@ -44,6 +44,14 @@ class Value(Outcome):
         return "Value({!r})".format(self.value)
 
 
+class KilledError(Exception):
+    """
+    This exception is thrown in a coroutine which has been explicity killed and
+    can then bubble up into coroutines that were awaiting that coroutine.
+    """
+    pass
+
+
 class _ErrorBase(Outcome):
     def __init__(self, error):
         self.error = error
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -480,12 +480,10 @@ class Scheduler(object):
                 self.log.debug("All coroutines scheduled, handing control back"
                                " to simulator")
 
-
     def kill(self, coro):
         self._pending_kills.append(coro)
 
-
-    def unschedule(self, coro):
+    def unschedule(self, coro, cleaning_up=False):
         """Unschedule a coroutine.  Unprime any pending triggers"""
 
         # Unprime the trigger this coroutine is waiting on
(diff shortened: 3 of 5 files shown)


ktbarrett:
```diff
@@ -442,6 +442,8 @@ class Scheduler(object):
                 trigger.unprime()
 
                 for coro in scheduling:
+                    if coro._outcome:
+                        continue
                     if _debug:
                         self.log.debug("Scheduling coroutine %s" % (coro.__name__))
                     self.schedule(coro, trigger=trigger)
```

Same effect, but does not introduce a new variable.

ktbarrett:
This should also include a self-checking test which should be simpler than the one from the issue.

benreynwar:
I'm instead taking the approach that `KilledError` should be raised in coroutines that are directly kliled.  This mirrors how it's done in asyncio with `CancelledError`.

Currently I'm trying to work out how I can fork a coroutine and somehow express that if it gets killed then I don't want that to bubble up and kill the test.

benreynwar:
It looks like I need to spend a bit more time on this.  The kill method is being used internally in ways that are not compatible with how I've changed things.  I expect there is a nice solution, but I'm not there yet.

eric-wieser:
I would recommend you put this on hold until we separate `RegressionManager` and `Scheduler` a little more.

benreynwar:
Sounds good to me.

ktbarrett:
@benreynwar Killing a `fork`ed coroutine *shouldn't* kill the test.  `fork`ing a coroutine should mean it runs as a parent-less coroutine, much like a test. If throwing `KilledError` in a `fork`ed coroutine is killing the test, there is yet another bug in the scheduler.

ktbarrett:
Instead of using a custom error, should we use `asyncio.CancelledError`? I have been playing around with writing code that works with both `asyncio` and cocotb. Making the two implementations raise the same error will make it easier to write code that works with both.

ktbarrett:
Actually, I think this is as simple as doing the outcome check and calling `self.close()` in `RunningTask.kill()`.

ktbarrett:
Actually, both are wrong. 

The fact we can kill any arbitrary `cocotb.coroutine` is kind of weird. I think the ideal semantics are are throwing `asyncio.CancelledError` in killed forked coroutines when they are `cancel`ed and not supporting the operation on non-forked coros. However, we don't currently have a facility to differentiate the two like asyncio does with `Task` and regular coroutines. Perhaps we should deprecate `cocotb.coroutine` for external use, only using it internally for `fork`. Then we can throw `CancelledError` on `kill()`, it might even be a good idea to change the name to `cancel()`.

eric-wieser:
> Perhaps we should deprecate cocotb.coroutine for external use, only using it internally for fork.

I've been thinking about this too. Perhaps:

* Deprecate the use of the decorator on async functions (as all it does is slow things down). This is unlikely to cause much trouble if we do it soon.
* Deprecate the use of yield functions, which removes the remaining use case of the decorator. This deprecation would probably need to last multiple releases.

ktbarrett:
@benreynwar While we are interested in fixing this bug and also adding support for `CanceledError` when coroutines are killed, we are investigating reimplementing some or all of the scheduler to accomplish that and other goals. It seems you ran into problems and stopped working on this PR. I'll close in one week if there is no response.

ktbarrett:
No activity, no response. We are interested in getting this behavior, but I think that will come with our rewrite of the scheduler.

benreynwar:
Sorry for not noticing this go by.    I agree that it makes sense to wait for the scheduler rewrite.

benreynwar:
Is it worth me coming back to this issue?  I got a workaround solution that I use to get this functionality in my own tests so it's not pressing for me, but it'd obviously be nicer for this functionality to be nicely integrated into cocotb.  Or is it already supported and I just need to start using it?

marlonjames:
We are planning on a similar mechanism for 2.0, re-using `asyncio.CancelledError`.

PR #1358, opened 2020-01-30
Ignore errors of sub-makes; add .PHONY; fix help text

cmarqu:


Makefile
```diff
@@ -27,31 +27,37 @@
...
-	$(MAKE) -C tests clean
+	-$(MAKE) -C examples clean
+	-$(MAKE) -C tests clean
 
+.PHONY: do_tests
 do_tests:
-	$(MAKE) -k -C tests
-	$(MAKE) -k -C examples
+	-$(MAKE) -k -C tests
+	-$(MAKE) -k -C examples
```
eric-wieser: These two strike me as errors we should not be ignoring

Makefile
```diff
@@ -27,31 +27,37 @@
...
+.PHONY: clean
 clean:
 	-@rm -rf $(BUILD_DIR)
 	-@find . -name "obj" | xargs rm -rf
 	-@find . -name "*.pyc" | xargs rm -rf
 	-@find . -name "*results.xml" | xargs rm -rf
-	$(MAKE) -C examples clean
-	$(MAKE) -C tests clean
+	-$(MAKE) -C examples clean
+	-$(MAKE) -C tests clean
```
eric-wieser: While safe, this doesn't seem particularly useful - don't the submake targets ignore all their errors anyway?

Makefile
ktbarrett: The issue is that if there were errors (any test failures) it does not run `combine_results.py` and print test results.

Makefile
ktbarrett: Not all apparently, I just checked. `make clean` fails halfway through because `target 'clean' not found`.

Makefile
ktbarrett: The issue is that if there were errors (any test failures) it does not run combine_results.py and print test results.

eric-wieser: This will undo the progress I made in #1314, to ensure failing runs propagate

eric-wieser: Only tests that fail to produce a results.xml at all should cause this command to fail.

cmarqu: > Only tests that fail to produce a results.xml at all should cause this command to fail.

But e.g. a ``SimFailure`` would do that, right? And apparently we have such tests/examples at least with Cadence simulators.
What we need in this target (or, to be precise, in the ``test`` target) is a way to produce as many results.xml as we can and always run the ``combine_results.py``. If you think ``do_tests`` should exit at the point where a results.xml could not be created, then we could revert this change and copy the two make lines into ``test``.

And maybe we should even extend #1314 by another command that writes a results.xml with content equivalent "Uncaught Failure" in case the new ``test -f`` returns false.

eric-wieser: What benefit does running `combine_results.py` provide?

Could what you ask for be achieved with something like 
```
$(MAKE) the_failing_target; ret=$?; combine_results; exit $ret
```

Makefile
eric-wieser: We should add a clean target to that test then

Makefile
eric-wieser: Which is pretty much exactly the top answer on stack overflow for what is pretty much exactly this problem: https://stackoverflow.com/a/3634360/102441

cmarqu: This is the accepted answer with 2 upvotes while the solution with prepending ``-`` is not accepted but has 4 upvotes ;)

cmarqu: > What benefit does running `combine_results.py` provide?

I want to take the combined_results.xml and see which of the tests and examples have failed.

> Could what you ask for be achieved with something like

Sure... or actually, not. If the first of the two make lines in the ``do_tests`` target fails, the second one will not run.

eric-wieser: > Sure... or actually, not. If the first of the two make lines in the do_tests target fails, the second one will not run.

I believe you can solve that with two separate double colon rules

cmarqu: I'd say that this is a case where a small bit of duplication leads to simpler (to understand) code.

eric-wieser: What I'm proposing is:
```
.PHONY: do_tests
do_tests::
	$(MAKE) -k -C tests
do_tests::
	$(MAKE) -k -C examples
```

cmarqu: Okay, I'll rework the PR with that code and some comments, hopefully tomorrow.
I agree that the target name ``test`` warrants some effort to return a useful exit code; if it was ``produce_teststatusfile`` not so much.

Makefile
ktbarrett: After looking through the makefiles, only `test_array` and `test_multidimensional_array` have issues and only when using Icarus. The ifdefs are missing `clean` targets. Should be an easy fix.

cmarqu: I have done this now.

Makefile
cmarqu: This is now in.
In working on the text for the ``help`` target I realized that the ``jenkins`` target was probably meant to give the non-abort behavior we have now (but didn't actually do that unless I'm wrong).
The targets ``test`` and ``jenkins`` are now almost similar, not sure if we want to remove one of them...

eric-wieser:
CI failure was #1369, close/reopen to rerun against that fix

cmarqu:
Sure, please do!

Makefile
```diff
@@ -39,26 +41,36 @@ clean:
...
+.PHONY: jenkins
 jenkins: do_tests
 	./bin/combine_results.py --suppress_rc --testsuites_name=cocotb_regression
 
 # By default want the exit code to indicate the test results
-test: do_tests
-	./bin/combine_results.py
+.PHONY: test
+test:
+	$(MAKE) do_tests; ret=$$?; ./bin/combine_results.py; exit $$ret
```
eric-wieser:
```suggestion
	$(MAKE) do_tests; ret=$$?; ./bin/combine_results.py && exit $$ret
```

Issue #1076, opened 2019-08-19
"Unable set up RisingEdge(c) Trigger" with Cadence IUS/Xcelium and VHDL


cmarqu:
Current status: See https://github.com/cocotb/cocotb/issues/1076#issuecomment-1351266296 for a summary of the problem.

Cadence tracks the implementation of the missing functionality as Jira ticket AVSREQ-183427.

---

When I run ``examples/dff/tests`` with 
```
make SIM=ius TOPLEVEL_LANG=vhdl COCOTB_LOG_LEVEL=DEBUG COCOTB_SCHEDULER_DEBUG=1 all
```
I am getting the following output:
```
     0.00ns DEBUG    cocotb.regression                         regression.py:555  in generate_tests                  Adding generated test "run_test_001" to module "dff_cocotb"
     0.00ns INFO     cocotb.regression                         regression.py:187  in initialise                      Found test dff_cocotb.run_test_001
     0.00ns INFO     cocotb.regression                         regression.py:321  in execute                         Running test 1/1: run_test_001
     0.00ns DEBUG    cocotb.scheduler                           scheduler.py:618  in add                             Adding new coroutine run_test_001
     0.00ns DEBUG    cocotb.scheduler                           scheduler.py:695  in schedule                        Scheduling coroutine run_test_001 with outcome Value(None)
     0.00ns INFO     ...test.run_test_001.0x7f4ae5046780       decorators.py:253  in _advance                        Starting test: "run_test_001"
                                                                                                                     Description: Automatically generated test
                                                                                                                     
                                                                                                                     
     0.00ns DEBUG    ...test.run_test_001.0x7f4ae5046780       decorators.py:258  in _advance                        Sending Value(None)
     0.00ns DEBUG    cocotb.scheduler                           scheduler.py:701  in schedule                        Coroutine run_test_001 yielded run_test (mode 1)
     0.00ns DEBUG    cocotb.scheduler                           scheduler.py:645  in _trigger_from_unstarted_coro    Scheduling nested coroutine: run_test
     0.00ns DEBUG    cocotb.scheduler                           scheduler.py:618  in add                             Adding new coroutine run_test
     0.00ns DEBUG    cocotb.scheduler                           scheduler.py:695  in schedule                        Scheduling coroutine run_test with outcome Value(None)
     0.00ns DEBUG    cocotb.c                                      handle.py:84   in __init__                        Created
[...]

PR #1359, opened 2020-01-30
Set some expect_error for Cadence simulators

cmarqu:


eric-wieser:
> otherwise I would leave it failing.

Normally the thing to do here is open an issue about fixing it, and then mark the failure as expected anyway. You need automated tests to pass in order to detect regressions, and marking issues as expected (as long as they make it to the bugtracker) is usually an accepted way to achieve that.

cmarqu:
> You need automated tests to pass in order to detect regressions, and marking issues as expected (as long as they make it to the bugtracker) is usually an accepted way to achieve that.

Right, this is about establishing this baseline against which regression tests can be run.

cmarqu:
> Normally the thing to do here is open an issue about fixing it, and then mark the failure as expected anyway.

The ticket for *these* markings is #1076. I'll add a comment to that effect.
... Actually, maybe we should have a ``reason`` argument so that we can use it later, like printing to stdout or the results.xml etc.

tests/test_cases/issue_120/issue_120.py
```diff
@@ -23,7 +23,7 @@ def monitor(dut):
         raise TestFailure("stream_in_valid should be high on the 5th cycle")
 
 
-@cocotb.test()
+@cocotb.test(expect_error=cocotb.triggers.TriggerException if cocotb.SIM_NAME.startswith(("xmsim", "ncsim")) and cocotb.LANGUAGE in ["vhdl"] else False)
```
eric-wieser: Do you know exactly which line this occurs on? It might be better to catch it in place.

eric-wieser:
> Actually, maybe we should have a `reason` argument so that we can use it later,

Agreed, I think that would be useful here.

tests/test_cases/issue_120/issue_120.py
cmarqu: It's in https://github.com/cocotb/cocotb/blob/05d49c0dbe481712443ec7b7517a4fb9686f5943/tests/test_cases/issue_120/issue_120.py#L20

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -274,7 +274,7 @@ def access_string_vhdl(dut):
 # TODO: add tests for Verilog "string_input_port" and "STRING_LOCALPARAM" (see issue #802)
 
 @cocotb.test(skip=cocotb.LANGUAGE in ["vhdl"],
-             expect_error=cocotb.SIM_NAME.lower().startswith("icarus"))
+             expect_error=cocotb.SIM_NAME.lower().startswith("icarus") or cocotb.SIM_NAME.startswith(("xmsim", "ncsim")))
```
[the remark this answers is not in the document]
cmarqu: This has already been done by another PR.

cmarqu:
I now added a comment with (shortened) error message and ticket number.

PR #1361, opened 2020-02-01
Fixes warning related to -Wsign-conversion

ktbarrett:
As discussed in #1306, the Apple build adds in `-Wsign-conversion` and this causes things to fail the CI because it sets `-Werror`. This adds that warning to the list of warnings enabled by default. Then it fixes those warnings in the VHPI and FLI interfaces. It also fixes inconsistency with the type of the `edge` variable in several places that was causing a sign conversion warning.

### DIFF
--- a/cocotb/share/lib/fli/FliCbHdl.cpp
+++ b/cocotb/share/lib/fli/FliCbHdl.cpp
@@ -125,9 +125,9 @@ int FliSimPhaseCbHdl::arm_callback()
 
 FliSignalCbHdl::FliSignalCbHdl(GpiImplInterface *impl,
                                FliSignalObjHdl *sig_hdl,
-                               unsigned int edge) : GpiCbHdl(impl),
-                                                    FliProcessCbHdl(impl),
-                                                    GpiValueCbHdl(impl, sig_hdl, edge)
+                               int edge) : GpiCbHdl(impl),
+                                           FliProcessCbHdl(impl),
+                                           GpiValueCbHdl(impl, sig_hdl, edge)
 {
     m_sig_hdl = m_signal->get_handle<mtiSignalIdT>();
 }
--- a/cocotb/share/lib/fli/FliCbHdl.cpp
+++ b/cocotb/share/lib/fli/FliCbHdl.cpp
@@ -56,7 +56,7 @@ FliTimedCbHdl::FliTimedCbHdl(GpiImplInterface *impl,
 int FliTimedCbHdl::arm_callback()
 {
     #if defined(__LP64__) || defined(_WIN64)
-        mti_ScheduleWakeup64(m_proc_hdl, m_time_ps);
+        mti_ScheduleWakeup64(m_proc_hdl, static_cast<mtiTime64T>(m_time_ps));
     #else
         mtiTime64T m_time_union_ps;
         MTI_TIME64_ASGN(m_time_union_ps, (mtiInt32T)((m_time_ps) >> 32), (mtiUInt32T)(m_time_ps));
--- a/cocotb/share/lib/fli/FliImpl.cpp
+++ b/cocotb/share/lib/fli/FliImpl.cpp
@@ -425,8 +425,8 @@ const char *FliImpl::reason_to_string(int reason)
  */
 void FliImpl::get_sim_time(uint32_t *high, uint32_t *low)
 {
-    *high = mti_NowUpper();
-    *low = mti_Now();
+    *high = static_cast<uint32_t>(mti_NowUpper());  // these functions return a int32_t for some reason
+    *low = static_cast<uint32_t>(mti_Now());
 }
 
 void FliImpl::get_sim_precision(int32_t *precision)
--- a/cocotb/share/lib/fli/FliImpl.h
+++ b/cocotb/share/lib/fli/FliImpl.h
@@ -65,7 +65,7 @@ class FliSignalCbHdl : public FliProcessCbHdl, public GpiValueCbHdl {
 public:
     FliSignalCbHdl(GpiImplInterface *impl,
                    FliSignalObjHdl *sig_hdl,
-                   unsigned int edge);
+                   int edge);
 
     int arm_callback() override;
     int cleanup_callback() override {
@@ -200,7 +200,7 @@ public:
                         m_either_cb(impl, this, GPI_FALLING | GPI_RISING) { }
 
 
-    GpiCbHdl *value_change_cb(unsigned int edge) override;
+    GpiCbHdl *value_change_cb(int edge) override;
     int initialise(std::string &name, std::string &fq_name) override;
 
     bool is_var() { return m_is_var; }
--- a/cocotb/share/lib/fli/FliImpl.h
+++ b/cocotb/share/lib/fli/FliImpl.h
@@ -231,7 +231,7 @@ public:
 
     ~FliValueObjHdl() override {
         if (m_val_buff != NULL)
-            free(m_val_buff);
+            delete [] m_val_buff;
         if (m_sub_hdls != NULL)
             mti_VsimFree(m_sub_hdls);
     }
@@ -315,7 +315,7 @@ public:
 
     ~FliLogicObjHdl() override {
         if (m_mti_buff != NULL)
-            free(m_mti_buff);
+            delete [] m_mti_buff;
     }
 
     const char* get_signal_value_binstr() override;
@@ -373,7 +373,7 @@ public:
 
     ~FliRealObjHdl() override {
         if (m_mti_buff != NULL)
-            free(m_mti_buff);
+            delete m_mti_buff;
     }
 
     double get_signal_value_real() override;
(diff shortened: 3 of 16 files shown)


cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -343,8 +348,10 @@ int VhpiLogicSignalObjHdl::initialise(std::string &name, std::string &fq_name) {
     if (vhpi_get(vhpiKindP, query_hdl) == vhpiArrayTypeDeclK) {
         m_indexable = true;
         m_value.format = vhpiLogicVecVal;
-        m_value.bufSize = m_num_elems*static_cast<bufSize_type>(sizeof(vhpiEnumT));
-        m_value.value.enumvs = (vhpiEnumT *)malloc(m_value.bufSize + 1);
+        int bufSize = m_num_elems * static_cast<int>(sizeof(vhpiEnumT));
+        m_value.bufSize = static_cast<bufSize_type>(bufSize);
+        size_t amt = (static_cast<size_t>(m_value.bufSize) + 1) * sizeof(vhpiEnumT);
```
ktbarrett: This looks fishy...

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -56,7 +56,7 @@ FliTimedCbHdl::FliTimedCbHdl(GpiImplInterface *impl,
 int FliTimedCbHdl::arm_callback()
 {
     #if defined(__LP64__) || defined(_WIN64)
-        mti_ScheduleWakeup64(m_proc_hdl, m_time_ps);
+        mti_ScheduleWakeup64(m_proc_hdl, static_cast<mtiInt32T>(m_time_ps));
     #else
```
eric-wieser: This seems very odd - the documentation for `mti_ScheduleWakeup64` claims this is a `mtiTime64T`, not a `mtiInt32T`:

image

Do you have a copy of the header file that says otherwise?

If so, we probably at least want a comment here.

cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -425,8 +425,8 @@ const char *FliImpl::reason_to_string(int reason)
  */
 void FliImpl::get_sim_time(uint32_t *high, uint32_t *low)
 {
-    *high = mti_NowUpper();
-    *low = mti_Now();
+    *high = static_cast<uint32_t>(mti_NowUpper());
```
eric-wieser:
```suggestion
    // mti functions return int32_t for some reason
    *high = static_cast<uint32_t>(mti_NowUpper());
```

cocotb/share/lib/fli/FliObjHdl.cpp
```diff
@@ -264,7 +265,8 @@ int FliLogicObjHdl::initialise(std::string &name, std::string &fq_name)
         m_enum_map[m_value_enum[i][1]] = i;  // enum is of the format 'U' or '0', etc.
     }
 
-    m_val_buff = (char*)malloc(m_num_elems+1);
+    size_t amt = static_cast<size_t>(m_num_elems) + 1;
```
eric-wieser: Can you change the type of `m_num_elements` to `size_t` instead?

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -99,8 +100,8 @@ bool get_range(vhpiHandleT hdl, vhpiIntT dim, int *left, int *right) {
                 if (curr_idx == dim) {
 
                     vhpi_release_handle(it);
-                    vhpiIntT l_rng = vhpi_get(vhpiLeftBoundP, constraint);
-                    vhpiIntT r_rng = vhpi_get(vhpiRightBoundP, constraint);
+                    int l_rng = static_cast<int>(vhpi_get(vhpiLeftBoundP, constraint));
+                    int r_rng = static_cast<int>(vhpi_get(vhpiRightBoundP, constraint));
 #ifdef IUS
                     if (l_rng != UNCONSTRAINED && r_rng != UNCONSTRAINED) {
```
eric-wieser: I don't think you should be casting before comparing to `UNCONSTRAINED`.

Can you put the cast on line 111 and 112 instead?

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -272,9 +273,11 @@ int VhpiSignalObjHdl::initialise(std::string &name, std::string &fq_name) {
 
         case vhpiStrVal: {
             m_indexable = true;
-            m_num_elems = vhpi_get(vhpiSizeP, handle);
-            m_value.bufSize = (m_num_elems)*static_cast<bufSize_type>(sizeof(vhpiCharT) + 1);
-            m_value.value.str = (vhpiCharT *)malloc(m_value.bufSize);
+            m_num_elems = static_cast<int>(vhpi_get(vhpiSizeP, handle));
+            int bufSize = m_num_elems * static_cast<int>(sizeof(vhpiCharT)) + 1;
```
eric-wieser:
```suggestion
            size_t bufSize = m_num_elems * sizeof(vhpiCharT) + 1u;
```

cocotb/share/lib/fli/FliCbHdl.cpp
ktbarrett: It's actually `mtiDelayT` which is `mtiInt32T`.

cocotb/share/lib/fli/FliObjHdl.cpp
ktbarrett: `m_num_elems` is usually `1` or the length of a vector whose bounds are 32-bit ints, so `int` (or better yet `int32_t`) makes the most sense. It would also create a bunch of warnings in other places, so nothing to gain from that.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
ktbarrett: God rid of the `+ 1` in the call to `malloc`, it is unnecessary.

cocotb/share/lib/fli/FliCbHdl.cpp
ktbarrett: Actually I'll need to check this, I changed it because there was a warning, but it doesn't make much sense for the "64" version of the function.

cocotb/share/lib/fli/FliObjHdl.cpp
```diff
@@ -473,13 +476,15 @@ int FliStringObjHdl::initialise(std::string &name, std::string &fq_name)
...
+    size_t amt    = static_cast<size_t>(m_num_elems);
+    m_mti_buff    = (char*)malloc(amt);
     if (!m_mti_buff) {
         LOG_CRITICAL("Unable to alloc mem for value object mti read buffer: ABORTING");
         return -1;
     }
 
-    m_val_buff = (char*)malloc(m_num_elems+1);
+    amt           = static_cast<size_t>(m_num_elems) + 1;
+    m_val_buff    = (char*)malloc(amt);
```
eric-wieser: Frankly we should just be using `m_val_buff = new char[amt];` here and everywhere else. I suppose that's out of scope though.

eric-wieser:
Apologies, created some conflicts with #1149. They should be straightforward to resolve.

cocotb/share/lib/fli/FliObjHdl.cpp
ktbarrett: Wait, they're C++ files? Couldn't tell...

If it makes warnings go away, I'll do it in this PR.

cocotb/share/lib/fli/FliCbHdl.cpp
ktbarrett: `mtiTime64T` from sources. Will fix later.

cocotb/share/lib/fli/FliImpl.h
```diff
@@ -403,7 +403,7 @@ class FliStringObjHdl : public FliValueObjHdl {
 
     ~FliStringObjHdl() override {
         if (m_mti_buff != NULL)
-            free(m_mti_buff);
+            delete m_mti_buff;
```
eric-wieser: This is incorrect - `new[]` must be paired with `delete[]` not `delete`

cocotb/share/lib/fli/FliImpl.h
```diff
@@ -373,7 +373,7 @@ class FliRealObjHdl : public FliValueObjHdl {
 
     ~FliRealObjHdl() override {
         if (m_mti_buff != NULL)
-            free(m_mti_buff);
+            delete m_mti_buff;
```
eric-wieser: An even nicer option would be to change the type of `m_mti_buff` from `T*` to `std::unique_ptr<T[]>`, which will handle this delete for you.

Fine to leave this for another patch, but you either need to fix these to `delete[]`, or just remove them entirely.

ktbarrett: I think I might, it's been so long since I've used `new` and `delete` raw that I apparently forgot how to use them correctly, might as well take that out of the equation.

Issue #1234, opened 2019-12-10
argv gets mangled with cadence xcelium


wimiko:
I see that ARGV gets mangled when using excelium. cocotb crashes, complaining about invalid utf8 characters.
The curious thing is: it only happens to the argument after the -f switch. I have modified `VpiStartupCbHdl::run_callback()` in VpiCbHdl.cpp to print info.argv and when i have 

`xrun  -v93 -f xrun_files_cmd -timescale 1ns/1ps -licqueue ... `

out comes:
`
xrun
-v93
-f
���
-timescale
1ns/1ps
-licqueue
...
`
If I move the -f option around, the filename stays mangled. If I have multiple -f options, i get multiple garbled filenames. All other options go through fine.

eric-wieser:
It might be instructive to print the raw byte values of the garbled argument - perhaps they are being replaced with file numbers

wimiko:
So, if I print the characters:

xrun  -v93 -f xrun_files_cmd -timescale 1ns/1ps 

xrun
78 72 75 6e 
-v93
2d 76 39 33 
-f
2d 66 
+M
10 2b 4d 05 
-timescale
2d 74 69 6d 65 73 63 61 6c 65 

The numbers change between runs. 
If I have multiple -f switches, with different files, it looks like this:

xrun
78 72 75 6e 
-v93
2d 76 39 33 
-f
2d 66 
���
ffffffd0 ffffffa0 ffffffea 03 
-f
2d 66 
���
ffffff90 ffffffe8 ffffffec 03 
-input
2d 69 6e 70 75 74 
xrun_commands.tcl
78 72 75 6e 5f 63 6f 6d 6d 61 6e 64 73 2e 74 63 6c 

Interestingly, the argument to the -input switch, which is also a filename comes trough ok.
Since the -f switch provides a file with more arguments, my guess is that the value is a reference to another structure that contains the arguments in the file, or something like that.

Does cocotb actually need these arguments for something? Maybe we can just let it fail in silence?

eric-wieser:
> Maybe we can just let it fail in silence?

We certainly could, but it would be cool to understand what the value we're getting instead is.

garmin-mjames:
I think I found answer to this.

From the 2001 standard:
[...]

PR #1364, opened 2020-02-03
Do not crash if argv cannot be decoded

eric-wieser:
Note that we still crash if memory allocations fail while decoding, but that seems like a sensible choice.

If argv contains byte sequences that cannot be decoded as UTF8, we now just emit a warning, and pass the raw bytes into `cocotb.argv`.

Workaround for #1234, with the compromise being that `cocotb.argv` contains a mixture of `bytes` and `str`.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -210,6 +210,10 @@ def process_plusargs():
     plusargs = {}
 
     for option in cocotb.argv:
+        # gh-1234: Sometimes we get non-strings in argv, just skip them
+        if not isinstance(option, str):
+            continue
+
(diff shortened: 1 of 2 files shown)


eric-wieser:
The gist of this patch is now:
```
try:
    argv[i] = raw.decode('utf8')
except:
    argv[i] = raw
```
I'm thinking it might be a better approach to use:
```
argv[i] = raw.decode('utf8', 'surrogateescape')
```
from which the original bytes can still be recovered using `raw.encode('utf8', 'surrogateescape')`

eric-wieser:
Closed by #1366

PR #1366, opened 2020-02-03
Use PEP383 to embed non-decodable bytes in argv

### LINKED ISSUE #1363: Intermittent simulator crash from UnicodeDecodeError
I am seeing an intermittent error at simulator start up when using `xcellium`

```
xrun(64): 18.03-s006: (c) Copyright 1995-2018 Cadence Design Systems, Inc.
Loading snapshot worklib.aes_gcm_top:sv .................... Done
xmsim: *W,DSEM2009: This SystemVerilog design is simulated as per IEEE 1800-2009 SystemVerilog simulation semantics. Use -disable_sem2009 option for turning off SV 2009 simulation semantics.
     -.--ns INFO     cocotb.gpi                                  gpi_embed.c:78   in set_program_name_in_venv        Did not detect Python virtual environment. Using system-wide Python interpreter
     -.--ns INFO     cocotb.gpi                                GpiCommon.cpp:104  in gpi_print_registered_impl       VPI registered
xcelium> source /cad/adi/apps/cadence/xlm/linux/18.03-s06/tools.lnx86/inca/files/xmsimrc
xcelium> run
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 0-1: invalid continuation byte
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 47-48: invalid continuation byte
     0.00ns ERROR    cocotb.gpi                                gpi_logging.c:224  in gpi_log                         Error calling Python logging function from C
```

Attaching the simulator log if it is any help.

eric-wieser:
Users who need the original bytes can use `.encode('utf8', 'surrogateescape')` to get them back.
Without this change, passing non-decodable bytes would just cause cocotb to fail to startup.

This behavior is consistent with how `sys.argv` is handled on UNIX platforms by python itself.

Alternative to gh-1364, and probably the better choice.

Fixes #1363

---
Workaround for #1234, using the same approach as is used for `sys.argv` in normal python.

### DIFF
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -307,7 +307,7 @@ int embed_sim_init(gpi_sim_info_t *info)
         PyObject *argv_item = PyUnicode_DecodeLocale(info->argv[i], "surrogateescape");  // New reference
         if (argv_item == NULL) {
             PyErr_Print();
-            LOG_ERROR("Unable to decode argv item");
+            LOG_ERROR("Unable to convert command line argument %d to Unicode string.", i);
             Py_DECREF(argv_list);
             goto cleanup;
         }
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -302,10 +302,20 @@ int embed_sim_init(gpi_sim_info_t *info)
         goto cleanup;
     }
     for (i = 0; i < info->argc; i++) {
-        PyObject *argv_item = PyUnicode_FromString(info->argv[i]);      // New reference
+        PyObject *argv_item_raw = PyBytes_FromString(info->argv[i]);       // New reference
+        if (argv_item_raw == NULL) {
+            PyErr_Print();
+            LOG_ERROR("Unable to create argv item");
+            Py_DECREF(argv_list);
+            goto cleanup;
+        }
+        // Decode, embedding non-decodable bytes using PEP-383. This can only
+        // fail with MemoryError or similar.
+        PyObject *argv_item = PyUnicode_FromEncodedObject(argv_item_raw, NULL, "surrogateescape");
+        Py_DECREF(argv_item_raw);
         if (argv_item == NULL) {
             PyErr_Print();
-            LOG_ERROR("Unable to create Python string from argv[%d] = \"%s\"", i, info->argv[i]);
+            LOG_ERROR("Unable to decode argv item");
             Py_DECREF(argv_list);
             goto cleanup;
         }
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -302,17 +302,9 @@ int embed_sim_init(gpi_sim_info_t *info)
         goto cleanup;
     }
     for (i = 0; i < info->argc; i++) {
-        PyObject *argv_item_raw = PyBytes_FromString(info->argv[i]);       // New reference
-        if (argv_item_raw == NULL) {
-            PyErr_Print();
-            LOG_ERROR("Unable to create argv item");
-            Py_DECREF(argv_list);
-            goto cleanup;
-        }
         // Decode, embedding non-decodable bytes using PEP-383. This can only
         // fail with MemoryError or similar.
-        PyObject *argv_item = PyUnicode_FromEncodedObject(argv_item_raw, NULL, "surrogateescape");
-        Py_DECREF(argv_item_raw);
+        PyObject *argv_item = PyUnicode_DecodeLocale(info->argv[i], "surrogateescape");  // New reference
         if (argv_item == NULL) {
             PyErr_Print();
             LOG_ERROR("Unable to decode argv item");


eric-wieser:
CI failure is unrelated (gh-1367)

eric-wieser:
Note that this issue didn't happen in 2.7, hence the Py3k tag

eric-wieser:
Confirmed working here: https://github.com/cocotb/cocotb/issues/1363#issuecomment-583585270.

Printing `cocotb.argv` gives nonsense as in https://github.com/cocotb/cocotb/issues/1363#issuecomment-583634605, but it definitely does not crash, and ought to have exactly the same effect as printing via `fprintf` in C.

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -302,10 +302,20 @@ int embed_sim_init(gpi_sim_info_t *info)
         goto cleanup;
     }
     for (i = 0; i < info->argc; i++) {
-        PyObject *argv_item = PyUnicode_FromString(info->argv[i]);      // New reference
+        PyObject *argv_item_raw = PyBytes_FromString(info->argv[i]);       // New reference
+        if (argv_item_raw == NULL) {
+            PyErr_Print();
+            LOG_ERROR("Unable to create argv item");
```
imphil:
```suggestion
            LOG_ERROR("Unable to create Python bytes object from command line argument %d", i);
```

cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -302,10 +302,20 @@ int embed_sim_init(gpi_sim_info_t *info)
...
+            goto cleanup;
+        }
+        // Decode, embedding non-decodable bytes using PEP-383. This can only
+        // fail with MemoryError or similar.
+        PyObject *argv_item = PyUnicode_FromEncodedObject(argv_item_raw, NULL, "surrogateescape");
+        Py_DECREF(argv_item_raw);
         if (argv_item == NULL) {
             PyErr_Print();
-            LOG_ERROR("Unable to create Python string from argv[%d] = \"%s\"", i, info->argv[i]);
+            LOG_ERROR("Unable to decode argv item");
```
imphil:
```suggestion
            LOG_ERROR("Unable to convert command line argument %d to Unicode string.", i);
```

eric-wieser:
> * The code above decodes all arguments to ASCII.

Before this PR: arguments are decoded _from_ the default encoding (which I think is UTF8 for python always), non-decodable bytes cause an error

After this PR: as above, but non-decodable bytes are embedded as python-specific surrogate escapes

> * The "hidden" byte information is only available if the str -> byte conversion is called with surrogateescape again.

Correct - however, there are parts of python that do this by default, such as perhaps file and standard io writes

> Users on POSIX operating systems expect arguments to be in their current locale, not in [UTF8]

This is a valid complaint, but was already an issue before this patch so is possibly declarable as out of scope.
I suppose I could look into using `Py_DecodeLocale` instead. Can you give me a reference for the fact that locale applies to command line arguments?

> To get the round-trip right, we consistently need to use `argv[i].encode(errors="surrogateescape")` in our code.

Our code never needs to perform this encoding though - we work in terms of unicode codepoints with no associated encodings. The only user who realistically needs to do that is someone who is passing raw bytestrings in argv and wants to recover them within cocotb.

cocotb/share/lib/embed/gpi_embed.c
eric-wieser: Could do, but there's little value here. The only way this error is reached is if we just printed out `MemoryError` due to having no space left to allocate the object.

Happy to apply that change if you think it's useful though.

cocotb/share/lib/embed/gpi_embed.c
eric-wieser: Same comment as above

imphil:
>>    Users on POSIX operating systems expect arguments to be in their current locale, not in [UTF8]

> This is a valid complaint, but was already an issue before this patch so is possibly declarable as out of scope.
I suppose I could look into using PyUnicode_DecodeLocale or possibly PyUnicode_DecodeFSDefaultAndSize instead. Can you give me a reference for the fact that locale applies to command line arguments?

Before, we always assumed UTF-8, and failed otherwise. Assuming UTF-8 isn't universally correct, but mostly in practice, since all modern Linux distributions default to a UTF-8 locale. With this PR, we force ASCII for all users of argv strings which don't know about the hidden byte information. So there's a behavior change that we need to be aware of.

As to your point what users expect: There's no universal rule how command line arguments should or must be interpreted. But it's quite natural to expect a program to interpret characters the way you typed them into your terminal. So it makes sense to use the locale as an hint how to interpret characters, which is what locale-dependent decoding will do.

As a data point, that's also what Python does for `sys.argv` (https://docs.python.org/3/library/sys.html#sys.argv), which would be `PyUnicode_DecodeFSDefaultAndSize()` probably.

eric-wieser:
> With this PR, we force ASCII for all users of argv strings

I don't understand where you're concluding ASCII from.

eric-wieser:
> As a data point, that's also what Python does for `sys.argv`

Pushed a second commit that imitates what python does in these lines:

* https://github.com/python/cpython/blob/5305cc9dbfe8a5a0ab666511f3ba7f026c8983f8/Python/fileutils.c#L595-L610
* https://github.com/python/cpython/blob/5305cc9dbfe8a5a0ab666511f3ba7f026c8983f8/Python/frozenmain.c#L64-L73

imphil:
Re ASCII vs UTF-8: The code was `PyObject *argv_item = PyUnicode_FromEncodedObject(argv_item_raw, NULL, "surrogateescape");`, where NULL is defined in https://docs.python.org/3/c-api/unicode.html#builtincodecs to "Setting encoding to NULL causes the default encoding to be used which is ASCII.".  I might be missing something, though.

eric-wieser:
I think those docs are wrong, they don't match the source - good find!

https://github.com/python/cpython/blob/58ac700fb09497df14d4492b6f820109490b2b88/Objects/unicodeobject.c#L3358-L3378

cocotb/share/lib/embed/gpi_embed.c
imphil: I'd prefer a good error message, even if it's very rare.

eric-wieser:
CI failure looks unrelated, and seems to be another windows CI issue.

Pushed an update with your error message.

Issue #1367, opened 2020-02-04
Travis windows CI is broken


eric-wieser:
Diagnosed here: https://travis-ci.community/t/docker-windows-containers-on-windows-the-operating-system-of-the-container-does-not-match-the-operating-system-of-the-host/427/10

I don't know if we can do anything about this, other than perhaps disable the CI for now.

PR #1369, opened 2020-02-04
fix windows travis build

themperek:
Related to #1367

- remove docker -> it is faster (12m->6m)
- use latest Icarus build
- use tox


.travis.yml
```diff
@@ -73,9 +73,17 @@ jobs:
...
       language: shell
       before_install:
-        - docker build -t cocotb -f Dockerfile.windows .
+        - powershell "Invoke-WebRequest -outfile miniconda3.exe https://repo.continuum.io/miniconda/Miniconda3-latest-Windows-x86_64.exe"
+        - powershell "Start-Process .\miniconda3.exe -ArgumentList '/S /D=C:\miniconda3' -Wait"
+
+        - powershell "Invoke-WebRequest -Uri http://bleyer.org/icarus/iverilog-v11-20190809-x64_setup.exe -OutFile iverilog-x64_setup.exe"
+        - powershell "Start-Process .\iverilog-x64_setup -ArgumentList '/VERYSILENT' -Wait"
+
+        - export CONDA_PATH="\$env:Path = ';C:\miniconda3;C:\miniconda3\Library\mingw-w64\bin;C:\miniconda3\Library\usr\bin;C:\miniconda3\Library\bin;C:\miniconda3\Scripts;C:\miniconda3\bin;C:\miniconda3\condabin;C:\iverilog\bin;' + \$env:Path"
```
eric-wieser: Can you set this env var directly, so that you can avoid some of this quoting?
```suggestion
        - export PATH="/C/miniconda3:/C/miniconda3/Library/mingw-w64/bin:/C/miniconda3/Library/usr/bin:/C/miniconda3/Library/bin:/C/miniconda3/Scripts:/C/miniconda3/bin:/C/miniconda3/condabin:/C/iverilog/bin:$PATH'
```

.travis.yml
```diff
@@ -73,9 +73,17 @@ jobs:
...
+        - powershell "$CONDA_PATH; conda install --yes -c msys2 m2-base m2-make m2w64-toolchain libpython"
+        - powershell "$CONDA_PATH; pip install tox"
```
eric-wieser: Can probably skip powershell for these bits with the above suggestion

themperek: Actually, for whatever reason this does not work (i tried ~10 versions) see: https://travis-ci.org/themperek/cocotb/jobs/645871415

themperek: Same for the above comment.

eric-wieser: You have a typo:
```
-export "PATH=$PATH:/C/miniconda3:/C/miniconda3/Library/mingw-w64/bin:/C/miniconda3/Library/usr/bin/C/miniconda3/Library/bin:/C/miniconda3/Scripts:/C/miniconda3/bin:/C/miniconda3/condabin"
+export "PATH=$PATH:/C/miniconda3:/C/miniconda3/Library/mingw-w64/bin:/C/miniconda3/Library/usr/bin:/C/miniconda3/Library/bin:/C/miniconda3/Scripts:/C/miniconda3/bin:/C/miniconda3/condabin"

.travis.yml
themperek: Actually, for whatever reason, this does not work (i tried ~10 versions) see: https://travis-ci.org/themperek/cocotb/jobs/645871415

.travis.yml
themperek: see https://travis-ci.org/themperek/cocotb/jobs/645871415

eric-wieser: Looks like it works with that fixed: https://travis-ci.org/themperek/cocotb/jobs/646043715

themperek: Something stuck actually (all this makes me cry)

eric-wieser: Well, something else breaks. Looks like something very weird is happening, perhaps worth adding `cat /c/Users/travis/.travis/functions` to see what's up

themperek: It is here: https://travis-ci.org/themperek/cocotb/jobs/646053844

themperek: I have no idea.

eric-wieser: This is the failing line, I'll see if I can work out why. Would be nice to report this to travis if we can work it out 
https://travis-ci.org/themperek/cocotb/jobs/646053844#L604

eric-wieser: Issue is that we added a different `date` to the path, which messes with Travis.

themperek: https://travis-ci.org/themperek/cocotb/jobs/646084512

eric-wieser: Nice, that found a bug in the makefile.

themperek: Not sure if a bug. It works in other cases.

eric-wieser: Patch incoming - it's a bug in all of our makefiles, we're just never using the broken bits.

eric-wieser: #1370

themperek:
It may have something to do with this bash shell installed on travis. May work with `which`.

PR #1370, opened 2020-02-04
Fix incorrect use of tabs in makefiles

eric-wieser:
These tabs cause the variable assignments to become part of the preceding recipe, which is not what was intended here.


eric-wieser:
CI failure is #1367, which probably wants this fix anyway.

PR #1375, opened 2020-02-08
Improve the __repr__ of all trigger objects.

eric-wieser:
Overloading only `__str__` meant that `print([trig1, trig2])` did not run our overloads at all.
Overloading only `__repr__` defines `__str__ = __repr__` implicitly.
Convention in python is to either:

* Make `eval(repr(x)) == x`
* Make `repr(x) == "<...>"` if the above is not possible

This patch respects that convention.

Since these are primarily for debugging, the object id is kept in the repr - it's been invaluable for spotting scheduler bugs.

Tested by running `test_cocotb` with debug enabled, so that every trigger is logged.

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -49,6 +49,17 @@ from cocotb import _py_compat
 import cocotb
 
 
+def _pointer_str(obj):
+    """
+    Get the memory address of *obj* as used in :meth:`object.__repr__`.
+
+    This is equivalent to ``sprintf("%p", id(obj))``, but python does not
+    support ``%p``.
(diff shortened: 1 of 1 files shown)


cocotb/triggers.py
```diff
@@ -49,6 +49,17 @@
 import cocotb
 
 
+def _pointer_str(obj):
+    """
+    Get the memory address of obj as used in object.__repr__.
```
```suggestion
    Get the memory address of *obj* as used in ``object.__repr__``.
```

eric-wieser: Done

PR #1378, opened 2020-02-10
Add timescale to Aldec `alog`

hofstee:
Adds timescale argument to `alog` following comments in #1102.

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.aldec
+++ b/cocotb/share/makefiles/simulators/Makefile.aldec
@@ -53,7 +53,7 @@ ifeq ($(GUI),1)
     CMD += -nosplash
 endif
 
-ALOG_ARGS += -timescale $(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION) 
+ALOG_ARGS += -timescale $(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION)
 
 # below allows for maintaining legacy syntax as well as enables using cross-simulator vars COMPILE_ARGS/SIM_ARGS
 ALOG_ARGS += $(COMPILE_ARGS)
--- a/cocotb/share/makefiles/simulators/Makefile.aldec
+++ b/cocotb/share/makefiles/simulators/Makefile.aldec
@@ -54,7 +54,7 @@ ifeq ($(GUI),1)
 endif
 
 # below allows for maintaining legacy syntax as well as enables using cross-simulator vars COMPILE_ARGS/SIM_ARGS
-ALOG_ARGS += $(COMPILE_ARGS)
+ALOG_ARGS += -timescale $(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION) $(COMPILE_ARGS)
 ACOM_ARGS += $(COMPILE_ARGS)
 ASIM_ARGS += $(SIM_ARGS)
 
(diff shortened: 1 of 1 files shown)


hofstee:
From what I can tell, `acom` (the VHDL compiler) doesn't support the argument. So this only adds it to the `ALOG_ARGS`.

cmarqu:
Is there anyone with Aldec Riviera who could test this?
(Speak up if you could but are not sure how to.)

cocotb/share/makefiles/simulators/Makefile.aldec
```diff
@@ -54,7 +54,7 @@ ifeq ($(GUI),1)
 endif
 
 # below allows for maintaining legacy syntax as well as enables using cross-simulator vars COMPILE_ARGS/SIM_ARGS
-ALOG_ARGS += $(COMPILE_ARGS)
+ALOG_ARGS += -timescale $(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION) $(COMPILE_ARGS)
```
cmarqu: Actually, since the comment above applies only to ``$(COMPILE_ARGS)`` and not ``-timescale``, could you make this two separate lines? This will append rather than prepend the timescale option, but I guess this doesn't matter.
Or, to be completely clear, move the timescale line above the comment?
```suggestion
ALOG_ARGS += $(COMPILE_ARGS)
ALOG_ARGS += -timescale $(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION)
```

binupr: I can check this as we have Aldec Riviera Pro. But we use VHDL as design language so may not be able to test the Verilog implementations. Also please guide me how I can test it? Download the latest dev clone and run/check for what etc.?

hofstee: @cmarqu reasoning for prepending was that if anyone already has `timescale` in their `COMPILE_ARGS` then by putting the cocotb timescale first it wouldn't break existing code.

I can still split it into separate lines. Do you have a preference for if it should be appended/prepended?

cmarqu: @hofstee Your reasoning makes sense, so let's do it like
```
ALOG_ARGS += -timescale $(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION)
# below allows for maintaining legacy syntax as well as enables using cross-simulator vars COMPILE_ARGS/SIM_ARGS
ALOG_ARGS += $(COMPILE_ARGS)
```

cmarqu: > I can check this as we have Aldec Riviera Pro. But we use VHDL as design language so may not be able to test the Verilog implementations. Also please guide me how I can test it? 

Does your license allow you to run Verilog code too, or is it VHDL-only? 
If the former, you can get the code of this pull request by doing
```
git fetch upstream pull/1378/head:pr1378
git checkout pr1378
```
(you'd need to ``git clone`` cocotb initially unless you already have it).
``cocotb/share/makefiles/simulators/Makefile.aldec`` should then show the ``-timescale`` line.
In that checkout, there are many testcases for designs in both Verilog and VHDL under ``tests/test_cases/``, I would guess running e.g.
```
cd tests/test_cases/test_cocotb
make SIM=aldec
```
would be fine for our purpose (it will automatically use default values for ``COCOTB_HDL_TIMESCALE`` and ``COCOTB_HDL_TIMEPRECISION``). Thanks in advance!

hofstee:
I'm just going to copy @binupr's reply here so it doesn't get lost in the resolved review block.

> I can check this as we have Aldec Riviera Pro. But we use VHDL as design language so may not be able to test the Verilog implementations. Also please guide me how I can test it? Download the latest dev clone and run/check for what etc.?

themperek:
@cmarqu I have tested and seems to be OK.

PR #1379, opened 2020-02-10
Fix UnboundLocalError if no tests run 

eric-wieser:
https://travis-ci.org/cocotb/cocotb/jobs/648364866#L2247

This was introduced in 6b6c623.

Also fixes an incorrect test count that counts only the number of tests in the last suite, not the number in all suites.

### DIFF
--- a/bin/combine_results.py
+++ b/bin/combine_results.py
@@ -68,8 +68,12 @@ def main():
 
     if args.debug : ET.dump(result)
 
-    for testsuite_count, testsuite in enumerate(result.iter('testsuite'),1):
-        for testcase_count, testcase in enumerate(testsuite.iter('testcase'),1):
+    testsuite_count = 0
+    testcase_count = 0
+    for testsuite in result.iter('testsuite'):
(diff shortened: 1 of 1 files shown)


eric-wieser:
This works correctly:

https://travis-ci.org/cocotb/cocotb/jobs/648367746#L2242

themperek:
Why CI fails?

eric-wieser:
Not sure, but it's failing with and without this change.

Windows CI is broken again on master, this just makes the error message the bottom a little clearer.

themperek:
It looks like virtualenv has changed.

Can  try to add `- powershell "$CONDA_PATH; conda install virtualenv"` after/before:
https://github.com/cocotb/cocotb/blob/f018c9c645c34cd24ccaf79c05a3a4d8d1b4695e/.travis.yml#L83


Maybe on windows, we need to give up `tox` (it was the case in the past) and just `make test`?

eric-wieser:
Can we merge this first and try that in a separate PR?

PR #1380, opened 2020-02-10
Change tests to run in definition-order

ktbarrett:
Previously, tests ran according to their names in lexical order. Now they run in order of their definition in a test module top-to-bottom. When specifying multiple test modules they are evaluated in the order they are listed.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -249,6 +249,7 @@ class RunningTest(RunningCoroutine):
         self.expect_error = parent.expect_error
         self.skip = parent.skip
         self.stage = parent.stage
+        self._id = parent._id
 
         # make sure not to create a circular reference here
         self.handler = RunningTest.ErrorLogHandler(self.error_messages.append)
@@ -456,12 +457,19 @@ class test(_py_compat.with_metaclass(_decorator_helper, coroutine)):
         stage (int, optional)
             Order tests logically into stages, where multiple tests can share a stage.
     """
+
+    _id_count = 0  # used by the RegressionManager to sort tests in definition order
+
     def __init__(self, f, timeout_time=None, timeout_unit=None,
                  expect_fail=False, expect_error=False,
                  skip=False, stage=None):
 
+        self._id = self._id_count
+        type(self)._id_count += 1
+
         if timeout_time is not None:
             co = coroutine(f)
+
             @functools.wraps(f)
             def f(*args, **kwargs):
                 running_co = co(*args, **kwargs)
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -182,7 +182,7 @@ class RegressionManager(object):
                 if hasattr(thing, "im_test"):
                     self._init_test(thing)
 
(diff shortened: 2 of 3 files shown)


cocotb/decorators.py
```diff
@@ -456,12 +457,18 @@ def my_test(dut):
...
             Order tests logically into stages, where multiple tests can share a stage.
     """
+    _id_count = 0
+
     def __init__(self, f, timeout_time=None, timeout_unit=None,
                  expect_fail=False, expect_error=False,
                  skip=False, stage=None):
 
+        self._id = self._id_count
+        self._id_count += 1
```
eric-wieser: I think this does the wrong thing, and you want
```suggestion
        type(self)._id_count += 1
```

eric-wieser:
For what it's worth, pytest also does things in definition order, but does so by line number rather than decorator evaluation order. The difference isn't very important, but is measurable with something like:
```
def test_1():
    yield Timer(1)

@cocotb.test()
def test_2():
    yield Timer(2)

test_1 = cocotb.test()(test_1)
```
Your PR runs `test_2` before `test_1`, pytest has no notion of test decorators so would not.

ktbarrett:
Line numbers have some issues: you can decorate the same function multiple times or decorate things that don't have line numbers like objects, or something even weirder if you wanted too... With pytest, most of the motivation for this usage is supported with custom decorators (like `pytest.mark.parameterize`) that handle that for you, allowing them to remain in control of the ordering.

cocotb/decorators.py
```diff
@@ -456,12 +457,18 @@ def my_test(dut):
         stage (int, optional)
             Order tests logically into stages, where multiple tests can share a stage.
     """
+    _id_count = 0
+
     def __init__(self, f, timeout_time=None, timeout_unit=None,
                  expect_fail=False, expect_error=False,
                  skip=False, stage=None):
 
+        self._id = self._id_count
```
```suggestion
        self._id = self._id_count  # used by the RegressionManager to sort tests in definition order
```

ktbarrett:
@eric-wieser Issues addressed.

imphil:
LGTM for the change itself. I've opened https://github.com/cocotb/cocotb/issues/1395 to get this documented.

imphil:
@ktbarrett Thanks for this change. To help us manage the release notes, we now suggest people add a newsfragment to changes like this. Find a discussion of how to do that at https://github.com/cocotb/cocotb/blob/master/documentation/source/newsfragments/README.rst. For this PR, let's move on by doing that in a follow-up together with the documentation of the feature itself.

ajeshp:
Looks like after this change `_log_test_summary` at the end of the test is not showing the names of the test as well as it used to before the change.
For one, instead of showing `MODULE.<test_name>` it is showing `cocotb.decorators.<test_name>` 
More importantly, all the TestFactory tests show up as `cocotb.decorators._my_test` instead of `MODULE.test_factory_func_name_test_number`
(By `MODULE`, I mean the `MODULE` name defined in Makefile)

Issue #1374, opened 2020-02-07
Handling of string values on GPI handles


ktbarrett:
As discussed in #1256. 

In the `simulator` module, strings are parsed using the `"s"` format, which encodes the string using UTF-8 (which is default). This means the user cannot choose a representation for strings. Another issue is that users can't send certain sequences like those containing embedded NULs.

In the aforementioned issue, it was recommended this be changed to `"y"`. But if these GPI "set" functions require `bytes`, "get" should result in `bytes`. This is a change in behavior that will break existing simulations. Even with this change, it only offers some advantage because the GPI implementation cannot handle embedded NULs.

We can handle embedded NULs from the Python C API by using the parsing format `"y#"` which results in a `char *` and a `int` length. We could pass around both values, *or* `std::string` could handle this, as it is not a NUL-terminated string. But we can't use `std::string` in the simulator module or GPI because they are C files, not C++. There really isn't a reason for any source to be C and not C++.

I propose the following set of changes:
* Make ALL source files C++ files
* Update the GPI to use `std::string` to handle embedded NULs
* Update all the interfaces to handle embedded NULs
* Move parsing format of "get" and "set" value functions in the `simulator` module to `"y#"`
* Update `BinaryValue` to use `bytes` as it's representation.

eric-wieser:
> Update BinaryValue to use bytes as it's representation.

I don't think there's a particularly good reason to do this. Given the only allowable values are ascii (`01xz` etc), and this is text not bytes, the current `str` is just fine.

The problem comes with true `string` types, where the encoding is unspecified, and so we should refuse the temptation to guess an encoding.

> Update all the interfaces to handle embedded NULs

Do VPI and VHPI actually support these? Is the systemverilog string type allowed to contain these?

eric-wieser:
> But we can't use std::string in the simulator module or GPI because they are C files, not C++

However, we can easily pass through the pointer and length returned by `"#y"` to `gpi_set_...`, and then construct the `std::string` before calling into the C++ish API.

ktbarrett:
Just gauging what you thought was wrong, how far you wanted to go with it, and documenting current limitations.

eric-wieser:
As an update, I think item 4 on your list is done, and item 5 is rejected as by design. That leaves:

* Make ALL source files C++ files (aka, remove the C gpi interface and use only the C++ one)
* Update the GPI to use std::string to handle embedded NULs
[...]

PR #1381, opened 2020-02-10
Change string type of handles to bytes

ktbarrett:
The simulator interfaces only deal in bytes, however the current implementation will take any string object including unicode sequences, which it will encode using UTF-8 (default) before writing it to the
simulator interface. Instead, string encoding/decoding should be lifted into userspace to offer more control.

`simulator.set_signal_value_str` and `simulator.get_signal_value` now expect and return a bytes object. `StringObject` handles now expect and return `bytes`.

Handles some issues in #1374.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -802,15 +802,28 @@ class StringObject(ModifiableObject):
         object, e.g. net, signal or variable.
 
         Args:
-            value (str): The value to drive onto the simulator object.
+            value (bytes): The value to drive onto the simulator object.
 
         Raises:
             TypeError: If target has an unsupported type for
                  string value assignment.
+
+        .. versionchanged:: 1.4
+            Takes :class:`bytes` instead of :class:`str`.
+            Users are now expected to choose an encoding when using these objects.
+            As a convenience, when assigning :class:`str` values, ASCII encoding will be used as a safe default.
+
         """
         value, set_action = self._check_for_set_action(value)
 
-        if not isinstance(value, str):
+        if isinstance(value, str):
+            warnings.warn(
+                "Handles on string objects will soon not accept `str` objects. "
+                "Please use a bytes object by encoding the string as you see fit. "
+                "`str.encode('ascii')` is typically sufficient.", DeprecationWarning, stacklevel=2)
+            value = value.encode('ascii')  # may throw UnicodeEncodeError
+
+        if not isinstance(value, bytes):
             self._log.critical("Unsupported type for string value assignment: %s (%s)", type(value), repr(value))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
--- a/documentation/source/newsfragments/1381.removal.rst
+++ b/documentation/source/newsfragments/1381.removal.rst
@@ -1,0 +1,1 @@
+The value of :class:`cocotb.handle.StringObject`\ s is now of type :class:`bytes`, instead of  :class:`str` with an implied ASCII encoding scheme.
--- a/tests/test_cases/test_cocotb/test_cocotb.py
+++ b/tests/test_cases/test_cocotb/test_cocotb.py
@@ -1436,3 +1436,13 @@ async def test_except_lock(dut):
         pass
     async with lock:
         pass
+
+
+# strings are not supported on Icarus
+@cocotb.test(skip=cocotb.SIM_NAME.lower().startswith("icarus"))
+async def test_string_handle_takes_bytes(dut):
+    dut.string_input_port.value = b"bytes"
+    await cocotb.triggers.Timer(10, 'ns')
+    val = dut.string_input_port.value
+    assert isinstance(val, bytes)
+    assert val == b"bytes"
--- a/tests/test_cases/test_cocotb/test_deprecated.py
+++ b/tests/test_cases/test_cocotb/test_deprecated.py
@@ -1,19 +1,38 @@
 import cocotb
 import warnings
+from contextlib import contextmanager
+
+
+@contextmanager
+def assert_deprecated():
+    warns = []
+    try:
+        with warnings.catch_warnings(record=True) as warns:
+            # Cause all warnings to always be triggered.
+            warnings.simplefilter("always")
+            yield warns  # note: not a cocotb yield, but a contextlib one!
+    finally:
+        assert len(warns) == 1
+        assert issubclass(warns[0].category, DeprecationWarning)
 
 
 @cocotb.test()
 async def test_returnvalue_deprecated(dut):
+
     @cocotb.coroutine
     def get_value():
         yield cocotb.triggers.Timer(1, units='ns')
         raise cocotb.result.ReturnValue(42)
 
-    with warnings.catch_warnings(record=True) as w:
-        # Cause all warnings to always be triggered.
-        warnings.simplefilter("always")
+    with assert_deprecated() as warns:
         val = await get_value()
     assert val == 42
-    assert len(w) == 1
-    assert issubclass(w[-1].category, DeprecationWarning)
-    assert "return statement instead" in str(w[-1].message)
+    assert "return statement instead" in str(warns[0].message)
+
+
+# strings are not supported on Icarus
+@cocotb.test(skip=cocotb.SIM_NAME.lower().startswith("icarus"))
+async def test_unicode_handle_assignment_deprecated(dut):
+    with assert_deprecated() as warns:
+        dut.string_input_port <= "Bad idea"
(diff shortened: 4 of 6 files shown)


cocotb/handle.py
```diff
@@ -752,7 +754,10 @@ def setimmediatevalue(self, value):
             TypeError: If target has an unsupported type for
                  string value assignment.
         """
-        if not isinstance(value, str):
+        if isinstance(value, str):
```
ktbarrett: This is just for convenience. It was likely this was the original intent, and will produce values safe for the simulator to consume or will throw an exception.

ktbarrett: Perhaps add a `DeprecationWarning` on this leg.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -648,15 +648,14 @@ static PyObject *get_signal_val_long(PyObject *self, PyObject *args)
     return retval;
 }
 
-
 static PyObject *set_signal_val_str(PyObject *self, PyObject *args)
```
eric-wieser: Is this used for both `binstr` and `str`? I remember a change to separate those two paths. Is it currently still in an open PR>

ktbarrett: Yeah that was with the Questa fixes. #1256. We talked about this in a comment there.

ktbarrett: I could break that commit out into a separate PR, I'd like to expand it to just not use overloading on `set_signal_value`, which obligates the need for the `using` statements to prevent hiding warnings.

eric-wieser: I think we want the change adding `gpi_set_signal_value_binstr` that's in that PR before this PR - either by finding time to review that PR in its entirety, or cherry-picking the single commit in question into a standalone PR.

ktbarrett: You have reviewed that PR. Opening this PR was the one outstanding issue. It probably deserves a few testers since it's FLI.

eric-wieser:
Needs a rebase now.

In my opinion, `get_signal_value_binstr` and `get_signal_value_str` should remain unchanged.

ktbarrett:
@eric-wieser added newsfragment.

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1316,4 +1316,14 @@ def test_bad_attr(dut):
...
 
 
+@cocotb.test(skip=cocotb.SIM_NAME.lower().startswith("icarus"))
+async def test_string_handle_takes_bytes(dut):
+    test_string = "bytes"
+    dut.string_input_port.value = test_string.encode("ascii")
+    await cocotb.triggers.Timer(10, 'ns')
+    val = dut.string_input_port.value
+    assert isinstance(val, bytes)
+    assert val.decode("ascii") == test_string
```
eric-wieser: Perhaps simpler to avoid an encoding dance here:
```suggestion
    dut.string_input_port.value = b"bytes"
    await cocotb.triggers.Timer(10, 'ns')
    val = dut.string_input_port.value
    assert isinstance(val, bytes)
    assert val == b"bytes"
```

tests/test_cases/test_cocotb/test_deprecated.py
```diff
@@ -0,0 +1,23 @@
...
+        assert issubclass(warn.category, DeprecationWarning)
+        return res, warn
+
+
+@cocotb.test(skip=cocotb.SIM_NAME.lower().startswith("icarus"))
+async def test_unicode_handle_assignment_deprecated(dut):
+    async def write_str():
+        dut.string_input_port <= "Bad idea"
+
+    result, warn = await catch_deprecation(write_str())
```
eric-wieser: A context manager would be easier here
```suggestion
from contextlib import contextmanager
@contextmanager
def assert_deprecated():
    warn = []
    try:
        with warnings.catch_warnings(record=True) as warn:
            # Cause all warnings to always be triggered.
            warnings.simplefilter("always")
            yield warn  # note: not a cocotb yield, but a contextlib one!
    finally:
        assert len(warn) == 1
        warn = warn[0]
        assert issubclass(warn.category, DeprecationWarning)

@cocotb.test(skip=cocotb.SIM_NAME.lower().startswith("icarus"))
async def test_unicode_handle_assignment_deprecated(dut):
    with assert_deprecated() as w:
        dut.string_input_port <= "Bad idea"
    assert "bytes" in w[0].message
```

ktbarrett: Since we can't test this with Icarus, I will change my other PR to include this so we have a reasonable good idea it works.

eric-wieser: No need to change that PR, but once I've merged it you can update the test from that PR to use this function too.

cocotb/handle.py
```diff
@@ -782,7 +782,14 @@ def setimmediatevalue(self, value):
         """
```
eric-wieser: This docstring should get a `.. deprecated:: 1.4.0` entry, and `str` needs to change to `bytes` above

documentation/source/newsfragments/1381.removal.rst
```diff
@@ -0,0 +1 @@
+String hierarchy objects now expect and resolve to :class:`bytes` objects, rather than taking :class:`str` objects and assuming an encoding scheme. Users are now expected to choose an encoding when using such objects. As a convience, and to prevent breakage of currently functioning tests, when assinging :class:`str` values to a string heirarchy object, the ASCII encoding scheme will be used as a safe default. This convenience will be removed after 1.4.
```
eric-wieser:
```suggestion
The value of :class:`cocotb.handle.StringObject`\ s is now of type :class:`bytes`, instead of  :class:`str` objects with an implied utf8 encoding scheme. Users are now expected to choose an encoding when using such objects. As a convience, and to prevent breakage of currently functioning tests, when assigning :class:`str` values, the ASCII encoding scheme will be used as a safe default.
```

tests/test_cases/test_cocotb/test_deprecated.py
```diff
@@ -1,19 +1,37 @@
...
+    with assert_deprecated() as warns:
         val = await get_value()
     assert val == 42
-    assert len(w) == 1
-    assert issubclass(w[-1].category, DeprecationWarning)
-    assert "return statement instead" in str(w[-1].message)
+    assert "return statement instead" in str(warns[0].message)
+
+
+@cocotb.test(skip=cocotb.SIM_NAME.lower().startswith("icarus"))
```
eric-wieser:
```suggestion
# strings are not supported in icarus
@cocotb.test(skip=cocotb.SIM_NAME.lower().startswith("icarus"))
```

ktbarrett: Done.

eric-wieser:
Let's get @cmarqu's approval since they added the support for strings in the first place.

cmarqu:
The string tests are not really running in CI, so I'm going to try locally soon.

cocotb/handle.py
```diff
@@ -802,15 +802,28 @@ def setimmediatevalue(self, value):
...
+            value (bytes): The value to drive onto the simulator object.
 
         Raises:
             TypeError: If target has an unsupported type for
                  string value assignment.
+
+        .. versionchanged:: 1.4
+            Takes :class:`bytes` instead of :class:`str`.
+            Users are now expected to choose an encoding when using these objects.
+            As a convenience, when assinging :class:`str` values, ASCII encoding will be used as a safe default.
```
cmarqu, line 814:
```suggestion
            As a convenience, when assigning :class:`str` values, ASCII encoding will be used as a safe default.
```

cmarqu:
At least the comparisons of and better also the assignments to ``string_const`` in
https://github.com/cocotb/cocotb/blob/ce09c28de28153e2e20fd8e350c8fda03a886f73/tests/test_cases/test_discovery/test_discovery.py#L279
and ``string_var`` in https://github.com/cocotb/cocotb/blob/ce09c28de28153e2e20fd8e350c8fda03a886f73/tests/test_cases/test_discovery/test_discovery.py#L301
need to be changed to ``b"string"``.

(While looking at this, I have done some unrelated improvements to this file which I'll put into another PR, so please do not change other things.)

ktbarrett:
@cmarqu Addressed test_discovery.

eric-wieser:
#1557 should fix CI

eric-wieser:
Thanks @ktbarrett!

PR #1382, opened 2020-02-11
Fail other tests on SimFailure

ktbarrett:
If a test gets a `SimFailure` or otherwise goes through `RegressionManager.tear_down`, iterate through remaining tests and score them as if the outcome was `SimFailure`, ensuring these tests get added as failures to the banner and result.xml. Previously, these tests were ignored.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -206,6 +206,21 @@ class RegressionManager(object):
                         cocotb.scheduler.add(test)
 
     def tear_down(self):
+        # fail remaining tests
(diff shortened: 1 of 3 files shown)


ktbarrett:
@eric-wieser the failure was just from the Windows CI issue.

cmarqu:
Could you provide a newsfragment in the ``bugfix`` category? (See https://github.com/cocotb/cocotb/blob/master/documentation/source/newsfragments/README.rst)

PR #1385, opened 2020-02-14
Add a pull request template for GitHub

cmarqu:
When merged, this is acting as a reminder when opening a pull request.
It started out as several paragraphs with explanatory links etc.,
but now I think succinct is better.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,16 @@
+<!--
+
+Thanks for improving cocotb! Here are some points to make this as smooth as possible.
+Not all of them may be applicable.
+
+Most important: please explain *why* you are proposing this change.
+
+* Make sure you have read https://github.com/cocotb/cocotb/blob/master/CONTRIBUTING.md
+* Extend or add a test under `tests/test_cases/`.
+* Add documentation under `documentation/source/`,
+  docstrings in Python code, or Doxygen markup in C/C++ code.
+  Use ``versionadded``/``versionchanged``/``deprecated``.
+* Add a newsfragment - see `documentation/source/newsfragments/README.rst`.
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+
+* Make sure you have read `CONTRIBUTING.md`.
+* Extend or add a test under `tests/test_cases/`.
+* Add documentation under `documentation/source/`,
+  docstrings in Python code, or Doxygen markup in C/C++ code.
+  Use ``versionadded``/``versionchanged``/``deprecated``.
+* Add a newsfragment - see `documentation/source/newsfragments/README.rst`.
+* Use `closes #XXXX` to auto-close the issue that this PR fixes (if such).
+
+*Please remove this template before submitting.*
```
eric-wieser: If you surround the template in `<!-- -->`, then there is no need to ask for it to be removed.

cmarqu: Done.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
+<!--
+
+Thanks for improving cocotb! Here are some points to make this as smooth as possible.
+Not all of them may be applicable.
+
+Most important: please explain *why* you are proposing this change.
+
+* Make sure you have read `CONTRIBUTING.md`.
```
imphil: Can we add a link here? A normal http://...

eric-wieser: Note that the PR form already has such a link, here:
image

imphil: Yes, but the whole point of the PR template is to be very explicit about the requirements for people who might not know how an open source project or GitHub works.

PR #1394, opened 2020-02-16
use pip3 to install tox on travis and Dockerfile

themperek:
Since on setup, we will use python3 syntax (#1306) need to install tox with `pip3`.

Python 2 is end of support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ jobs:
         - docker ps -a
       script:
         - echo 'Running cocotb tests with Python 3.5 ...' && echo -en 'travis_fold:start:cocotb_py35'
-        - docker exec -it -e TRAVIS=true cocotb bash -lc 'pip install tox; cd /src; tox -e py35'
+        - docker exec -it -e TRAVIS=true cocotb bash -lc 'pip3 install tox; cd /src; tox -e py35'
         - echo 'travis_fold:end:cocotb_py35'
 
     - stage: SimTests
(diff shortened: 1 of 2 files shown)


eric-wieser:
How does this work with the Travis python version? I'm surprised pip doesn't already refer to the correct version.

themperek:
On travis/ubuntu was ok. But not in docker and osx see: https://travis-ci.org/cocotb/cocotb/builds/651121802

eric-wieser:
I'm worried that this change causes Travis/Ubuntu to use an incorrect version. Can we fix just the osx / docker ones?

themperek:
`pip3` is the recommended way by our documentation (https://cocotb.readthedocs.io/en/latest/quickstart.html#installation-via-pip)

>Can we fix just the osx / docker ones?

For Ubuntu yes. For osx this is not a docker. There is probably some way to set pip to default to python os osx but ...

themperek:
Something like may work on osx :`python3 -m pip install tox` But to me does not look better see: https://stackoverflow.com/questions/50052050/how-to-set-default-version-of-python-to-3-x-vs-2-7-on-mac-osx

eric-wieser:
To be clear, I'm suggesting we make this change, but remove the parts that aren't relevant to one of docker and osx.

themperek:
@eric-wieser Now?

themperek:
@eric-wieser It is great. Thank you.

PR #1401, opened 2020-02-17
Add documentation for test ordering change

### LINKED ISSUE #1395: Document run order of tests
With #1380 merged, tests are now defined to run in the order they are listed in the source file. Document this fact somewhere in the docs, and the change in the release notes.

ktbarrett:
Per https://github.com/cocotb/cocotb/pull/1380#issuecomment-586724801 adding a newsfragment and documentation of how tests are ordered. Fixes #1395.

Not entirely sure where to stick the documentation besides on the decorator itself.

### DIFF
--- a/documentation/source/newsfragments/1380.change.rst
+++ b/documentation/source/newsfragments/1380.change.rst
@@ -1,0 +1,1 @@
+Tests are now evaluated in order of their appearance in the :envvar:`MODULE` environment variable, their stage, and the order of invocation of the :class:`cocotb.test` decorator within a module.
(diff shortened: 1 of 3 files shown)


documentation/source/newsfragments/1380.change.rst
```diff
@@ -0,0 +1 @@
+Tests are now evaluated in order of their appearance in the :envvar:`MODULE` environment variable, their stage, and the order of invocation of the :class:`cocotb.test` decorator within a module.
```
eric-wieser: I think we should mention what used to happen too

imphil: Good idea! This PR is merged, but feel free to open a follow-up PR or issue.

PR #1403, opened 2020-02-17
Expose force/release to cocotb

thasti:
I finally got around to looking at #657. I have mostly re-applied the patch prepared by @lukedarnell to current master, fixed small issues with it and gave it a test.

As it stands, a testbench running on a VPI simulator can do:
```
from cocotb.handle import Force, Release, Deposit
...
dut.q <= 0 # default action: deposit
dut.q <= Deposit(0) # does the same as above
dut.q <= Force(1) # forces the net to 1
dut.q <= Release() # releases the net
```

What's still to be done:
  - add documentation
  - implement the mechanisms also for VHPI/FLI
  - add a proper unit test/example

I'd already like to hear whether there are some thoughts on how it was implemented to figure out if any significant changes would be required in your opinion.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -577,6 +577,35 @@ class NonConstantObject(NonHierarchyIndexableObject):
         """An iterator for gathering all loads on a signal."""
         return _SimIterator(self._handle, simulator.LOADS)
 
+class _SetAction:
+    """Base class representing the type of action used while write-accessing a handle."""
+    pass
+
+class _SetValueAction(_SetAction):
+    __slots__ = ("value",)
+    """Base class representing the type of action used while write-accessing a handle with a value."""
+    def __init__(self, value):
+        self.value = value
+
+class Deposit(_SetValueAction):
+    """Action used for placing a value into a given handle."""
+    def _as_gpi_args_for(self, hdl):
+        return self.value, 0  # GPI_DEPOSIT
+
+class Force(_SetValueAction):
+    """Action used to force a handle to a given value until a release is applied."""
+    def _as_gpi_args_for(self, hdl):
+        return self.value, 1  # GPI_FORCE
+
+class Freeze(_SetAction):
+    """Action used to make a handle keep its current value until a release is used."""
+    def _as_gpi_args_for(self, hdl):
+        return hdl.value, 1  # GPI_FORCE
+
+class Release(_SetAction):
+    """Action used to stop the effects of a previously applied force/freeze action."""
+    def _as_gpi_args_for(self, hdl):
+        return 0, 2  # GPI_RELEASE
 
 class ModifiableObject(NonConstantObject):
     """Base class for simulator objects whose values can be modified."""
@@ -598,10 +627,11 @@ class ModifiableObject(NonConstantObject):
             TypeError: If target is not wide enough or has an unsupported type
                  for value assignment.
         """
-        if isinstance(value, _py_compat.integer_types) and value < 0x7fffffff and len(self) <= 32:
-            simulator.set_signal_val_long(self._handle, value)
-            return
+        value, set_action = self._check_for_set_action(value)
 
+        if isinstance(value, _py_compat.integer_types) and value < 0x7fffffff and len(self) <= 32:
+            simulator.set_signal_val_long(self._handle, set_action, value)
+            return
         if isinstance(value, ctypes.Structure):
             value = BinaryValue(value=cocotb.utils.pack(value), n_bits=len(self))
         elif isinstance(value, _py_compat.integer_types):
@@ -625,7 +655,12 @@ class ModifiableObject(NonConstantObject):
             self._log.critical("Unsupported type for value assignment: %s (%s)", type(value), repr(value))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
-        simulator.set_signal_val_binstr(self._handle, value.binstr)
+        simulator.set_signal_val_binstr(self._handle, set_action, value.binstr)
+
+    def _check_for_set_action(self, value):
+        if not isinstance(value, _SetAction):
+            return value, 0  # GPI_DEPOSIT
+        return value._as_gpi_args_for(self)
 
     @NonConstantObject.value.getter
     def value(self):
@@ -666,6 +701,8 @@ class RealObject(ModifiableObject):
             TypeError: If target has an unsupported type for
                 real value assignment.
         """
+        value, set_action = self._check_for_set_action(value)
+
         try:
             value = float(value)
         except ValueError:
@@ -673,7 +710,7 @@ class RealObject(ModifiableObject):
                                (type(value), repr(value)))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
-        simulator.set_signal_val_real(self._handle, value)
+        simulator.set_signal_val_real(self._handle, set_action, value)
 
     @ModifiableObject.value.getter
     def value(self):
@@ -699,13 +736,15 @@ class EnumObject(ModifiableObject):
             TypeError: If target has an unsupported type for
                  integer value assignment.
         """
+        value, set_action = self._check_for_set_action(value)
+
         if isinstance(value, BinaryValue):
             value = int(value)
         elif not isinstance(value, _py_compat.integer_types):
             self._log.critical("Unsupported type for integer value assignment: %s (%s)", type(value), repr(value))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
-        simulator.set_signal_val_long(self._handle, value)
+        simulator.set_signal_val_long(self._handle, set_action, value)
 
     @ModifiableObject.value.getter
     def value(self):
@@ -728,13 +767,15 @@ class IntegerObject(ModifiableObject):
             TypeError: If target has an unsupported type for
                  integer value assignment.
         """
+        value, set_action = self._check_for_set_action(value)
+
         if isinstance(value, BinaryValue):
             value = int(value)
         elif not isinstance(value, _py_compat.integer_types):
             self._log.critical("Unsupported type for integer value assignment: %s (%s)", type(value), repr(value))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
-        simulator.set_signal_val_long(self._handle, value)
+        simulator.set_signal_val_long(self._handle, set_action, value)
 
     @ModifiableObject.value.getter
     def value(self):
@@ -757,11 +798,13 @@ class StringObject(ModifiableObject):
             TypeError: If target has an unsupported type for
                  string value assignment.
         """
+        value, set_action = self._check_for_set_action(value)
+
         if not isinstance(value, str):
             self._log.critical("Unsupported type for string value assignment: %s (%s)", type(value), repr(value))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
-        simulator.set_signal_val_str(self._handle, value)
+        simulator.set_signal_val_str(self._handle, set_action, value)
 
     @ModifiableObject.value.getter
     def value(self):
--- a/cocotb/share/lib/fli/FliObjHdl.cpp
+++ b/cocotb/share/lib/fli/FliObjHdl.cpp
@@ -140,30 +140,34 @@ long FliValueObjHdl::get_signal_value_long()
     return -1;
 }
 
-int FliValueObjHdl::set_signal_value(long value)
+int FliValueObjHdl::set_signal_value(long value, gpi_set_action_t action)
 {
     COCOTB_UNUSED(value);
+    COCOTB_UNUSED(action);
     LOG_ERROR("Setting signal/variable value via long not supported for %s of type %d", m_fullname.c_str(), m_type);
     return -1;
 }
 
-int FliValueObjHdl::set_signal_value_binstr(std::string &value)
+int FliValueObjHdl::set_signal_value_binstr(std::string &value, gpi_set_action_t action)
 {
     COCOTB_UNUSED(value);
+    COCOTB_UNUSED(action);
     LOG_ERROR("Setting signal/variable value via string not supported for %s of type %d", m_fullname.c_str(), m_type);
     return -1;
 }
 
-int FliValueObjHdl::set_signal_value_str(std::string &value)
+int FliValueObjHdl::set_signal_value_str(std::string &value, gpi_set_action_t action)
 {
     COCOTB_UNUSED(value);
+    COCOTB_UNUSED(action);
     LOG_ERROR("Setting signal/variable value via string not supported for %s of type %d", m_fullname.c_str(), m_type);
     return -1;
 }
 
-int FliValueObjHdl::set_signal_value(double value)
+int FliValueObjHdl::set_signal_value(double value, gpi_set_action_t action)
 {
     COCOTB_UNUSED(value);
+    COCOTB_UNUSED(action);
     LOG_ERROR("Setting signal/variable value via double not supported for %s of type %d", m_fullname.c_str(), m_type);
     return -1;
 }
@@ -221,8 +225,13 @@ long FliEnumObjHdl::get_signal_value_long()
     }
 }
 
-int FliEnumObjHdl::set_signal_value(const long value)
+int FliEnumObjHdl::set_signal_value(const long value, const gpi_set_action_t action)
 {
+    if (action != GPI_DEPOSIT) {
+        LOG_CRITICAL("Force or release action not supported for FLI.");
+        return -1;
+    }
+
     if (value > m_num_enum || value < 0) {
         LOG_ERROR("Attempted to set a enum with range [0,%d] with invalid value %d!\n", m_num_enum, value);
         return -1;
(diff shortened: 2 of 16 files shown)


imphil:
This PR has a large number of conflicts. Can you try to rebase on top of the current master?

thasti:
Indeed, I was a bit behind current master. Should be better now :)

cocotb/handle.py
```diff
@@ -625,7 +637,19 @@ def setimmediatevalue(self, value):
             self._log.critical("Unsupported type for value assignment: %s (%s)", type(value), repr(value))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
-        simulator.set_signal_val_binstr(self._handle, value.binstr)
+        simulator.set_signal_val_binstr(self._handle, set_action, value.binstr)
+
+    def check_for_set_action(self, value):
+        if isinstance(value,Freeze):
```
imphil: nit: space before comma here and in other places in this patch.

```suggestion
        if isinstance(value, Freeze):
```

cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -367,26 +367,45 @@ int VpiSignalObjHdl::set_signal_value_str(std::string &value)
...
+    switch (action) {
+    case GPI_FORCE:
+      vpi_put_flag = vpiForceFlag;
+      break;
+    case GPI_RELEASE:
+      // Best to pass its current value to the sim when releasing
+      vpi_get_value(GpiObjHdl::get_handle<vpiHandle>(),&value_s);
+      vpi_put_flag = vpiReleaseFlag;
+      break;
+    default:
```
imphil: Can you use the explicit GPI_DEPOSIT flag here, and put an `assert(0)` into the default block to catch us potential bugs when extending the enum?

cocotb/handle.py
```diff
@@ -625,7 +637,19 @@ def setimmediatevalue(self, value):
             self._log.critical("Unsupported type for value assignment: %s (%s)", type(value), repr(value))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
-        simulator.set_signal_val_binstr(self._handle, value.binstr)
+        simulator.set_signal_val_binstr(self._handle, set_action, value.binstr)
+
+    def check_for_set_action(self, value):
```
imphil: @eric-wieser Two questions: do you see a more elegant way to do that, and have you seen the signal setting code in the hot path of profiles? (I.e. do we need to have a closer look at performance before merging.)

cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -367,26 +368,47 @@ int VpiSignalObjHdl::set_signal_value_str(std::string &value)
...
+          // Use Inertial delay to schedule an event, thus behaving like a verilog testbench
+          vpi_put_flag = vpiInertialDelay;
+      }
+      break;
+    case GPI_FORCE:
+      vpi_put_flag = vpiForceFlag;
+      break;
+    case GPI_RELEASE:
+      // Best to pass its current value to the sim when releasing
+      vpi_get_value(GpiObjHdl::get_handle<vpiHandle>(),&value_s);
```
cmarqu: Nit:
```suggestion
      vpi_get_value(GpiObjHdl::get_handle<vpiHandle>(), &value_s);
```

tests/test_cases/test_force_release/Makefile
```diff
@@ -0,0 +1,30 @@
+###############################################################################
```
cmarqu: This being a new file, you can use the following SPDX identifier instead of the long license text:
```
```

cocotb/handle.py
eric-wieser: Using `isinstance` checks like this is generally slower than using method dispatching, I think.

So perhaps the thing to do is:
```python
if not isinstance(value, SetAction):
    value = Deposit(value)
return value._as_gpi_args_for(self)
```
I haven't done any profiling for a while, so don't know what the hot path is

cocotb/handle.py
```diff
@@ -577,6 +577,26 @@ def loads(self):
         """An iterator for gathering all loads on a signal."""
         return _SimIterator(self._handle, simulator.LOADS)
 
+class SetAction:
+    """Base class representing the type of action used while setting a handle value."""
```
eric-wieser: Marginal performance increase, maybe:
```suggestion
    """Base class representing the type of action used while setting a handle value."""
    __slots__ = ("value",)
```

cocotb/handle.py
```diff
@@ -577,6 +577,26 @@ def loads(self):
...
+
+class Deposit(SetAction):
+    """Action used for placing a value into a given handle."""
+    _action = 0
+
+class Force(SetAction):
+    """Action used to force a handle to a given value until a release is applied."""
+    _action = 1
+
+class Release(SetAction):
```
eric-wieser: I don't think `Release(0xFF)` should be legal if you're going to ignore the value. Can you perhaps instead have:
```python
class _SetAction: pass  # no .value member

class _SetValueAction(_SetAction):
    # what you currently have as _SetAction

class Release(_SetAction): ...

class Force(_SetValueAction): ...
```
Note also that I added a leading `_` for `SetAction` - I don't think we want users using it directly.

cocotb/share/lib/fli/FliObjHdl.cpp
```diff
@@ -221,8 +225,13 @@ long FliEnumObjHdl::get_signal_value_long()
     }
 }
 
-int FliEnumObjHdl::set_signal_value(const long value)
+int FliEnumObjHdl::set_signal_value(const long value, const gpi_set_action_t action)
 {
+    if (action != GPI_DEPOSIT) {
+        LOG_CRITICAL("Force or release action not supported for FLI.");
+        return -1;
+    }
```
eric-wieser: Nice, thanks - with this in place, I don't think we need to wait for FLI support before merging

cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -367,26 +368,47 @@ int VpiSignalObjHdl::set_signal_value_str(std::string &value)
...
 
     vpi_time_s.type = vpiSimTime;
     vpi_time_s.high = 0;
     vpi_time_s.low  = 0;
 
-    if (vpiStringVar == vpi_get(vpiType, GpiObjHdl::get_handle<vpiHandle>())) {
-        // assigning to a vpiStringVar only seems to work with vpiNoDelay
+    switch (action) {
+    case GPI_DEPOSIT:
+      if (vpiStringVar == vpi_get(vpiType, GpiObjHdl::get_handle<vpiHandle>())) {
```
eric-wieser: Nit: this block should be indented 2 more spaces, our indents are 4 and not 2 spaces

cocotb/share/lib/vpi/VpiImpl.h
```diff
@@ -173,17 +173,17 @@ class VpiSignalObjHdl : public GpiSignalObjHdl {
...
     long get_signal_value_long() override;
 
-    int set_signal_value(const long value) override;
-    int set_signal_value(const double value) override;
-    int set_signal_value_binstr(std::string &value) override;
-    int set_signal_value_str(std::string &value) override;
+    int set_signal_value(const long value, const gpi_set_action_t action) override;
+    int set_signal_value(const double value, const gpi_set_action_t action) override;
+    int set_signal_value_binstr(std::string &value, const gpi_set_action_t action) override;
+    int set_signal_value_str(std::string &value, const gpi_set_action_t action) override;
```
eric-wieser: nit: Drop the `const` here, it's meaningless

documentation/source/quickstart.rst
```diff
@@ -250,6 +250,24 @@ writes are not applied immediately, but delayed until the next write cycle.
 Use ``sig.setimmediatevalue(new_val)`` to set a new value immediately
 (see :meth:`~cocotb.handle.ModifiableObject.setimmediatevalue`).
 
+In addition to regular value assignments (deposits), signals can be forced
+to a predetermined value or frozen at their current value. To achieve this,
+the various actions derived from :class:`~cocotb.handle.SetAction` can be used.
```
eric-wieser: You'll need to add `autoclass` directives to the `library_reference.rst` in order for this link to work. I suggest you instead use something like:
```
.. _set-actions-section

Set actions   # this is a lousy title, please use a better one!
===========

.. currentmodule:: cocotb.handle

.. autoclass:: Deposit
```
and then make this line
```suggestion
the various actions described in :ref:`set-actions-section` can be used.
```

documentation/source/quickstart.rst
```diff
@@ -250,6 +250,24 @@ writes are not applied immediately, but delayed until the next write cycle.
...
+
+.. code-block:: python3
+
+    # Deposit action
+    dut.my_signal <= 12
+    dut.my_signal <= Deposit(12) # equivalent syntax
```
eric-wieser: nit: PEP8 wants
```suggestion
    dut.my_signal <= Deposit(12)  # equivalent syntax
```
same throughout

documentation/source/newsfragments/1403.feature.rst
```diff
@@ -0,0 +1,15 @@
+Cocotb now supports force/release/freeze actions on simulator handles, exposing functionality similar to the respective verilog assignments to cocotb.
```
cmarqu:
```suggestion
Cocotb now supports deposit/force/release/freeze actions on simulator handles, exposing functionality similar to the respective Verilog/VHDL assignments.
```
VHDL-2008 also supports (part of) this: https://www.doulos.com/knowhow/vhdl_designers_guide/vhdl_2008/vhdl_200x_small/#force

documentation/source/quickstart.rst
```diff
@@ -250,6 +250,24 @@ writes are not applied immediately, but delayed until the next write cycle.
...
+
+    # Force action
+    dut.my_signal <= Force(12) # my_signal stays 12 until released
+
+    # Release action
+    dut.my_signal <= Release() # Reverts any force/release assignments
```
cmarqu: Should this be
```suggestion
    dut.my_signal <= Release() # Reverts any force/freeze assignments
```
?

documentation/source/newsfragments/1403.feature.rst
thasti: Sounds better indeed, applied!

tests/test_cases/test_force_release/test_force_release.py
```diff
@@ -0,0 +1,34 @@
...
+from cocotb.handle import Force, Release
+
+
+@cocotb.test(expect_fail=cocotb.SIM_NAME in ["GHDL"])
+def test_force_release(dut):
+    """
+    Test force and release on simulation handles
+    """
+    log = logging.getLogger("cocotb.test")
+    yield Timer(10)
```
cmarqu:
```suggestion
    yield Timer(10, "ns")
```
(and a few times more below). We try to establish this as best practice, we just haven't gotten around to changing the existing tests yet.

documentation/source/library_reference.rst
```diff
@@ -166,6 +166,21 @@ Simulation Object Handles
     :show-inheritance:
     :synopsis: Classes for simulation objects.
 
+.. _set-actions-section
```
eric-wieser: Could change this to `assignment-methods-section` to match the heading. Fine to leave as is if you think matching the python class name is more important.

tests/test_cases/test_force_release/test_force_release.py
cmarqu: Actually, you may even want to use https://cocotb.readthedocs.io/en/latest/coroutines.html#async-functions here to be state-of-the-art :)

cocotb/handle.py
```diff
@@ -577,6 +577,31 @@ def loads(self):
...
+
+class _SetValueAction(_SetAction):
+    __slots__ = ("value",)
+    """Base class representing the type of action used while write-accessing a handle with a value."""
+    def __init__(self, value):
+        self.value = value
+
+class Deposit(_SetValueAction):
+    """Action used for placing a value into a given handle."""
+    _action = 0
```
eric-wieser: As suggested below in https://github.com/cocotb/cocotb/pull/1403/files#r380357823:
```suggestion
    def _as_gpi_args_for(self, hdl):
        return self.value, 0
```

cocotb/handle.py
```diff
@@ -577,6 +577,31 @@ def loads(self):
...
+    """Action used to force a handle to a given value until a release is applied."""
+    _action = 1
+
+class Release(_SetAction):
+    """Action used to stop the effects of a previously applied force/freeze action."""
+    _action = 2
+
+class Freeze(_SetAction):
+    """Action used to make a handle keep its current value until a release is used."""
+    _action = 3
```
```suggestion
    def _as_gpi_args_for(self, hdl):
        return hdl.value, 3
```

cocotb/handle.py
```diff
@@ -625,7 +655,12 @@ def setimmediatevalue(self, value):
             self._log.critical("Unsupported type for value assignment: %s (%s)", type(value), repr(value))
             raise TypeError("Unable to set simulator value with type %s" % (type(value)))
 
-        simulator.set_signal_val_binstr(self._handle, value.binstr)
+        simulator.set_signal_val_binstr(self._handle, set_action, value.binstr)
+
+    def check_for_set_action(self, value):
```
```suggestion
    def _check_for_set_action(self, value):
```
so as not to be public API. Not a huge fan of this name, but I don't have a better one to hand.

cocotb/handle.py
```diff
@@ -625,7 +655,12 @@ def setimmediatevalue(self, value):
...
+        if not isinstance(value, _SetAction):
+            value = Deposit(value)
```
eric-wieser: It _might_ be worth making this `return value, 0  # GPI_DEPOSIT`. It would be faster, and is possibly the most likely case. However, it is less readable.

Another possibly-faster spelling would be:
```
try:
    as_gpi_args_for = value._as_gpi_args_for
except AttributeError:
    return value, 0  # GPI_DEPOSIT
else:
    return as_gpi_args_for(self)
```

Difficult to know without profiling.

thasti: The first one makes sense to me, at it is (for most testbenches) the most common occurrence. Also it's still pretty readable, given that the return values are clearly named in the callee and the scope is pretty limited. The try/except one tends to look a bit messy, though.

tests/test_cases/test_force_release/test_force_release.py
cmarqu: @thasti Did you see my second comment here? (Maybe I shouldn't have appended it.)

thasti: Yes, was seen. Applied now as well!

thasti:
Any chance someone can test the VHPI implementation? I do not have access to a simulator implementing VHPI at the moment.

cocotb/handle.py
thasti: I feel the same, but don't have a better name either. Sorry!

cocotb/handle.py
eric-wieser: The idea was that `try / except` may be faster than calling `isinstance`. I don't have much to back that up though.

eric-wieser:
@themperek can probably test that

cocotb/handle.py
thasti: I built a quick out-of-context benchmark to check. It appears the try-except version fares much worse whenever an exception needs to be handled (i.e. for the most common code path in this situation).

https://gist.github.com/thasti/a2c4e553b39a9d914b34766eb42fb779

The less common path (a SetAction instance is passed) is pretty much the same cost for both tests, but the more common one is faster in the "isinstance" implementation:

```
Try/Except - Except: 0.385
Try/Except - Else:   0.208
isinstance - No:     0.125
isinstance - Yes:    0.194
```

Also (not in the gist): Directly returning the tuple instead of constructing yet another object saves considerable overhead for such small data structures, so the current version appears to be the best in performance evaluated so far.

thasti:
```
Try/Except - Except: 0.407
Try/Except - Else:   0.207
isinstance - No:     0.238
isinstance - Yes:    0.201
getattr - None:      0.133
getattr - not None:  0.232
```

Quite a bit faster for the most common path, while not slower for the less common one. Changed in the PR.

thasti: Actually I bodged this last benchmark by not reverting the modification to test2 that creates a new instance. So, when run as in the gist, the results are:

```
Try/Except - Except: 0.383
Try/Except - Else:   0.201
isinstance - No:     0.101
isinstance - Yes:    0.194
getattr - None:      0.125
getattr - not None:  0.221
```

And therefore the isinstance-based one is consistently the fastest. Reverted the code changes to use this implementation.

thasti:
Many thanks everyone for the constructive review yesterday. Today I was able to check the patch also compiles in a Python 3.6.8 environment and the testcase correctly works for Cadence Xcelium and Incisive. On my other machine (Python 3.8) I can test against iverilog (works fine) and ghdl (runs, but force/release is not implemented).

Let me know in case there's something else you'd like to see happening for this PR.

ktbarrett:
>"Force or release action not supported for FLI."

This isn't true, `mti_ForceSignal` has the ability to force, deposit, freeze, (and drive!) a signal, and `mti_ReleaseSignal` can release a signal. They are just not what is currently used to write to signal handles. The drive method is particularly interesting and is probably worth another `SetValueAction`, even if only Questa will be able to do it right now.

thasti:
I'm aware of that. Perhaps 'not implemented' is the wording I should have used instead. In case I can establish access to Questa before this is merged I may try also adding an FLI implementation.

ktbarrett:
Ah, if you don't have access to Questa, and have no way to test it, it's probably best to leave it out of this PR. I might take a look at adding a more thorough FLI implementation after this PR is in, it's definitely something we want. Thank you for the work.

Issue #1404, opened 2020-02-17
Passing plusargs to FLI: Find code that works with different TCL versions


themperek:
There is an error on RHEL7 (gcc version 4.8.5 20150623 (Red Hat 4.8.5-36) (GCC)

Compilation error in ‘Tcl_Interp’ has no member named ‘result’ 

```
 cocotb/share/lib/fli/FliCbHdl.cpp:164:44: error: ‘Tcl_Interp’ has no member named ‘result’
       unsigned long argc_u = strtoul(interp->result, NULL, 10);
                                              ^
  cocotb/share/lib/fli/FliCbHdl.cpp:183:43: error: ‘Tcl_Interp’ has no member named ‘result’
           argv[i] = new char[strlen(interp->result)+1];
                                             ^
  cocotb/share/lib/fli/FliCbHdl.cpp:184:33: error: ‘Tcl_Interp’ has no member named ‘result’
           strcpy(argv[i], interp->result);
                                   ^
```

Seems to be related to https://github.com/cocotb/cocotb/pull/1256  @ktbarrett

imphil:
Is cocotb picking up the right tcl library and header file from Questa, or is it using the system-provided one?

themperek:
It looks as if it is ok for Questa 10.x but 2019.x fails.

imphil:
Can you try reverting https://github.com/cocotb/cocotb/commit/b283b48368a3a7f507110c6699db3a6c7ce01677 and see if that fixes compilation? (If so, I'll back out this change until we find a solution that universally works.)

themperek:
Checked https://github.com/cocotb/cocotb/commit/b283b48368a3a7f507110c6699db3a6c7ce01677 seems to be a problem.

themperek:
The version of tcl changed from 8.5 to 8.6

imphil:
OK, revert in #1407. @ktbarrett can you look into finding a solution which works for both simulators? Ideally, one that doesn't involve conditional compilation (see #1392).

PR #1407, opened 2020-02-17
Revert "Implemented PLUSARGS for FLI"

imphil:
This reverts commit b283b48368a3a7f507110c6699db3a6c7ce01677.

The Questa-shipped TCL interface changed from TCL 8.5 to 8.6 in newer
versions, which is incompatible with this change.

See #1404

### DIFF
--- a/cocotb/share/lib/fli/FliCbHdl.cpp
+++ b/cocotb/share/lib/fli/FliCbHdl.cpp
@@ -26,10 +26,6 @@
 ******************************************************************************/
 
 #include "FliImpl.h"
-#include "mti.h"
-#undef DLLEXPORT  // redefinition in tcl.h causes warning
-#include "tcl.h"
-#include <limits>
 
 /**
  * @name    cleanup callback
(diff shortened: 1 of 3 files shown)


imphil:
@themperek can you check if that fixes the problem for you?

eric-wieser:
@ktbarrett, can we add an ifdef to fix this?

@imphil, can you be clear which version of tcl this code _did_ work with?

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -144,55 +140,6 @@ int FliStartupCbHdl::arm_callback()
...
-    info->argc = 0;
-    info->argv = NULL;
-
-    Tcl_Interp* interp = reinterpret_cast<Tcl_Interp*>(mti_Interp());
-
-    // get argc
-    if (mti_Cmd("llength $argv") != TCL_OK) {
-        return;
-    }
-    unsigned long argc_u = strtoul(interp->result, NULL, 10);
```
eric-wieser: I think `Tcl_GetStringResult(interp)` would fix this in 8.6, but may break 8.5

imphil:
This is purely a mechanical revert to get code back to compile. 

I have neither written the original code, nor done the testing. Any improvements must be done in later PRs and discussed in #1404.

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -144,55 +140,6 @@ int FliStartupCbHdl::arm_callback()
...
-    }
-    int argc = static_cast<int>(argc_u);
-    if (argc_u == 0) {
-        return;
-    }
-
-    // allocate storage for argv
-    char** argv = new char*[argc];
-
-    // get each argv and copy into internal storage
```
eric-wieser: Tempting to use `Tcl_ListObjGetElements` and `Tcl_GetString` here instead.

PR #1409, opened 2020-02-17
add python version check in setup.py

themperek:
This avoids non-obvious error with python2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,10 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
+import sys
+if sys.version_info[:2] < (3, 5):
+    raise RuntimeError("Python version >= 3.5 required.")
+
 import logging
 from setuptools import setup
 from setuptools import find_packages


setup.py
```diff
@@ -28,6 +28,10 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
+import sys
+if (sys.version_info < (3, 5)):
```
eric-wieser:
```suggestion
if sys.version_info < (3, 5):
```
no need for extra parens

setup.py
```diff
@@ -28,6 +28,10 @@
...
+    raise Exception('cocotb requires Python version >= 3.5')
```
eric-wieser: Could follow `numpy`'s lead here: https://github.com/numpy/numpy/blob/master/setup.py#L29-L30
```suggestion
    raise RuntimeError('cocotb requires Python version >= 3.5')
```

setup.py
eric-wieser: I'd be inclined to keep a comment like this somewhere, but we can add that as a follow-up

PR #1410, opened 2020-02-17
Log non-string messages with colored log formatter

### LINKED ISSUE #1408: Logging non-string messages leads to AttributeError
Test code:

```
dut._log.info(dut.empty.value)
# dut.empty is a signal in the DUT, .value gets its BinaryValue object
```

This worked fine in older versions of cocotb, haven't attempted to find exact regression range yet. Now I get this trace:

```
  File "/usr/lib64/python3.7/logging/__init__.py", line 1034, in emit
    msg = self.format(record)
  File "/usr/lib64/python3.7/logging/__init__.py", line 880, in format
    return fmt.format(record)
  File "/home/philipp/.local/lib/python3.7/site-packages/cocotb/log.py", line 201, in format
    msg = '\n'.join([SimColourLogFormatter.loglevel2colour[record.levelno] % line for line in msg.split('\n')])
AttributeError: 'BinaryValue' object has no attribute 'split'
Call stack:
  File "/home/philipp/.local/lib/python3.7/site-packages/cocotb/scheduler.py", line 355, in react
    self._event_loop(trigger)
  File "/home/philipp/.local/lib/python3.7/site-packages/cocotb/scheduler.py", line 448, in _event_loop
    self.schedule(coro, trigger=trigger)
  File "/home/philipp/.local/lib/python3.7/site-packages/cocotb/scheduler.py", line 758, in schedule
    result = coroutine._advance(send_outcome)
  File "/home/philipp/.local/lib/python3.7/site-packages/cocotb/decorators.py", line 264, in _advance
    return super(RunningTest, self)._advance(outcome)
  File "/home/philipp/.local/lib/python3.7/site-packages/cocotb/decorators.py", line 146, in _advance
    return outcome.send(self._coro)
  File "/home/philipp/.local/lib/python3.7/site-packages/cocotb/outcomes.py", line 38, in send
    return gen.send(self.value)
  File "path/to/test_fifo.py", line 45, in test_fifo_manual
    dut._log.info(dut.empty.value)
```

I'll try to come up with a test and a fix soon-ish.

imphil:
The colored log formatter expects messages to be strings (e.g. by
calling `.split()` on them. Ensure that the messages are actually
strings.

Fixes #1408

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -167,12 +167,8 @@ class SimLogFormatter(logging.Formatter):
 
     def format(self, record):
         """Prettify the log output, annotate with simulation time"""
-        if record.args:
-            msg = record.msg % record.args
-        else:
-            msg = record.msg
 
-        msg = str(msg)
+        msg = record.getMessage()
         level = record.levelname.ljust(_LEVEL_CHARS)
 
         return self._format(level, record, msg)
@@ -192,10 +188,7 @@ class SimColourLogFormatter(SimLogFormatter):
     def format(self, record):
         """Prettify the log output, annotate with simulation time"""
 
-        if record.args:
-            msg = record.msg % record.args
-        else:
-            msg = record.msg
+        msg = record.getMessage()
 
         # Need to colour each line in case coloring is applied in the message
         msg = '\n'.join([SimColourLogFormatter.loglevel2colour[record.levelno] % line for line in msg.split('\n')])
--- a/documentation/source/newsfragments/1410.bug.rst
+++ b/documentation/source/newsfragments/1410.bug.rst
@@ -1,0 +1,1 @@
+Logging of non-string messages with colored log output is now working.


imphil:
Testing for the exception we see in #1408 is actually tricky, as the exception isn't thrown when writing the log, but out of band.

@eric-wieser do you have a good idea how to test that? Or do we need to write back the log buffer somehow? I'm tempted to leave this fix as-is without a test if we find it too annoying to test for it ...

cocotb/log.py
```diff
@@ -192,6 +192,9 @@ class SimColourLogFormatter(SimLogFormatter):
     def format(self, record):
         """Prettify the log output, annotate with simulation time"""
 
+        # To be able to format them, messages must be strings.
+        record.msg = str(record.msg)
+
         if record.args:
             msg = record.msg % record.args
         else:
```
eric-wieser: Correct fix I think is to replace all these lines with `msg = record.getMessage()`, as is done [by python in the default logger](https://github.com/python/cpython/blob/4c1b6a6f4fc46add0097efb3026cf3f0c89f88a2/Lib/logging/__init__.py#L651-L664).

eric-wieser: Same fix is needed on line 173

imphil: Good find, updated.

eric-wieser:
Change looks good - worth a test perhaps?

eric-wieser:
Probably worth a bugfix newsfragment too

imphil:
I added a newsfragment. 

For the test, I didn't find a reasonable way to test. Directly checking for the exception doesn't seem to do the trick, as it is thrown out of band. Any ideas? I can think of a couple involved ways, but that seems overkill.

documentation/source/newsfragments/1410.bug.rst
```diff
@@ -0,0 +1 @@
+Logging of non-string messages with colored log output is now working.
```
eric-wieser: Isn't this true of uncolored output too?

imphil: Yes. But that was already the case before.

eric-wieser: Ah yep, I see now.

PR #1411, opened 2020-02-17
Store the simulator time in the log record, rather than retrieving it when printing

eric-wieser:
Without this change, it was possible that records had incorrect timestamps - if someone attached a handlers that put the records in a list, and formatted them only at shutdown, they'd find that all their log messages had the same timestamp.

This also fixes a crash if `get_sim_time` tries to log. (#1320)

https://external-builds.readthedocs.io/html/cocotb/1411/library_reference.html#cocotb.log.SimTimeContextFilter

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -34,7 +34,9 @@ import sys
 import logging
 import warnings
 
-from cocotb.utils import get_sim_time, want_color_output
+from cocotb.utils import (
+    get_sim_time, get_time_from_sim_steps, want_color_output
+)
 
 import cocotb.ANSI as ANSI
 
@@ -56,7 +58,9 @@ def default_config():
 
     This hooks up the logger to write to stdout, using either
     :class:`SimColourLogFormatter` or :class:`SimLogFormatter` depending
-    on whether colored output is requested.
+    on whether colored output is requested. It also adds a
+    :class:`SimTimeContextFilter` filter so that
+    :attr:`~logging.LogRecord.created_sim_time` is available to the formatter.
 
     The logging level for cocotb logs is set based on the
     :envvar:`COCOTB_LOG_LEVEL` environment variable, which defaults to ``INFO``.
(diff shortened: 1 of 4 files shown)


cocotb/log.py
```diff
@@ -122,8 +125,35 @@ def SimLog(name, ident=None):
...
+            # attach a simulator time to this message.
+            record.created_sim_time = None
+        return True
+
+
 class SimLogFormatter(logging.Formatter):
-    """Log formatter to provide consistent log message handling."""
+    """Log formatter to provide consistent log message handling.
+
+    This will only add simualtor timestamps if the logger object has a
```
imphil:
```suggestion
    This will only add simulator timestamps if the logger object has a
```

documentation/source/newsfragments/1411.feature.rst
```diff
@@ -0,0 +1,4 @@
+Custom logging handlers can now access the simulator time using
+``LogRecord.created_sim_time``, provided the
+:class:`~cocotb.log.SimTimeContextFilter` filter is not removed from the logger
+instance.
```
imphil: That describes the internal change well. How about also describing the user-visible outcome of this, i.e. time stamps are now matching the time when the log entry was created, and not when it was printed.

eric-wieser: I don't think this is actually user visible unless the user is doing something very unusual.

cocotb/log.py
```diff
@@ -122,8 +127,38 @@ def SimLog(name, ident=None):
...
+    This uses the approach describes at
+    https://docs.python.org/3/howto/logging-cookbook.html#using-filters-to-impart-contextual-information
+
+    This adds a ``.created_sim_time`` attribute to the :class:`~logging.LogRecord`
+    instances:
+
+    .. attribute:: logging.LogRecord.created_sim_time
+        The result of :func:`get_sim_time` at the point the log was created
+        (in simulator units). The formatter is responsible for converting this
+        to something like nanoseconds via :func:`get_time_from_sim_steps`.
```
eric-wieser: I'm not super hopeful that this will actually work...

eric-wieser:
Let's do https://github.com/cocotb/cocotb/pull/1412 first, it's difficult to document this when none of the existing things it relates to are documented yet

eric-wieser:
Docs are causing me some trouble, will merge once I get the links working

eric-wieser:
Link to the cookbook now works:

image

PR #1412, opened 2020-02-17
Add Sphinx docs for the non-deprecated logging functions and classes

eric-wieser:
https://external-builds.readthedocs.io/html/cocotb/1412/library_reference.html#logging

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -125,6 +125,12 @@ def SimLog(name, ident=None):
 class SimLogFormatter(logging.Formatter):
     """Log formatter to provide consistent log message handling."""
 
+    # Removes the arguments from the base class. Docstring needed to make
+    # sphinx happy.
+    def __init__(self):
+        """ Takes no arguments. """
+        super().__init__()
+
     # Justify and truncate
     @staticmethod
     def ljust(string, chars):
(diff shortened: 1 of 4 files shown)


cocotb/log.py
```diff
@@ -125,6 +125,12 @@ def SimLog(name, ident=None):
 class SimLogFormatter(logging.Formatter):
     """Log formatter to provide consistent log message handling."""
 
+    # Removes the arguments from the base class. Docstring needed to make
+    # sphinx happy.
+    def __init__(self):
+        """ Takes no arguments. """
+        super().__init__()
```
eric-wieser: @imphil, if you make the change I suggested in #1410, we could add the `style` argument here.

eric-wieser: Nevermind, ignore that remark, I'm incorrect.

eric-wieser:
@cmarqu: In particular, would appreciate suggestions about which heading this should fall under. Ok with you moving it later as part of a wider restructure too.

cmarqu:
@eric-wieser The heading is fine IMO. What would be beneficial in addition is a cross-reference from https://external-builds.readthedocs.io/html/cocotb/1412/testbench_tools.html#logging

eric-wieser:
@cmarqu: Updated

eric-wieser:
And doc CI build is now done.

PR #1413, opened 2020-02-17
remove unused COCOTB_*_LIB* variables

themperek:
Before #1306 `COCOTB_LIBS`, `COCOTB_VPI_LIB`, `COCOTB_FLI_LIB`, `COCOTB_VHPI_LIB`
ware used to pointing to compiled libraries and used as dependency on simulation (**they are empty and not used anymore**). 
Use `CUSTOM_SIM_DEPS` to add simulation dependency on libraries. 

cleanup after #1306


eric-wieser:
Is it worth adding a phony `cocotb_installed` target to the root makefile, and using that in place of these?

The implementation could be something simple that just checks if it's installed, and prints an error if it's not.

cocotb/share/makefiles/simulators/Makefile.icarus
```diff
@@ -85,7 +85,7 @@ $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/sim.vvp $(CUSTOM_SIM_DEPS) $(COCOTB_LIBS) $
 	# check that the file was actually created, since we can't set an exit code from cocotb
 	test -f $(COCOTB_RESULTS_FILE)
 
-debug: $(SIM_BUILD)/sim.vvp $(CUSTOM_SIM_DEPS) $(COCOTB_LIBS)
+debug: $(SIM_BUILD)/sim.vvp $(CUSTOM_SIM_DEPS)
 	-@rm -f $(COCOTB_RESULTS_FILE)
```
eric-wieser: This one is missing the vpi_lib, which was somehow not copied from the target above

eric-wieser:
Can you include in the commit message:
* What this variable contains / used to contain
* Why it is no longer needed

Otherwise looks good.

themperek:
> Is it worth adding a phony `cocotb_installed` target to the root makefile, and using that in place of these?

For this, I may need some help.


Not sure if need at this point we need to assume cocotb is not installed but we should check if `fli` is there and give appropriate error.

eric-wieser:
To be clear, no need to add that phony target as part of this PR.

eric-wieser:
> COCOTB_LIBS variable was pointing to cocotb compiled libraries

Which ones specifically? Did it point to the GPI libraries that are already present in the other variables?

themperek:
Here: https://github.com/cocotb/cocotb/blob/2d11eeac80e97f2f356662b029228690dfc990d5/cocotb/share/lib/Makefile#L66-L70

eric-wieser:
Removing these seems possibly unwise, as it will cause `make results.xml` to no longer re-run a simulation if I reinstall cocotb.

themperek:
>Removing these seems possibly unwise, as it will cause make results.xml to no longer re-run a simulation if I reinstall cocotb.

This is already removed (od commit) I can only add them.

themperek:
Can add  ?

`CUSTOM_SIM_DEPS += $(shell find $(LIBDIR)/ -name "*.so")`

eric-wieser:
That's not a bad idea. It might lead to a few unnecessary rebuilds, but that doesn't matter nearly as much as stale builds.

themperek:
Added `CUSTOM_SIM_DEPS += $(shell find $(LIB_DIR)/ -name "*")`

themperek:
> Added `CUSTOM_SIM_DEPS += $(shell find $(LIB_DIR)/ -name "*")`

I am not sure if this should not be in `CUSTOM_SIM_DEPS` or got back to `COCOTB_LIBS` or remove.

cocotb/share/makefiles/simulators/Makefile.ghdl
```diff
@@ -64,7 +64,7 @@ analyse: $(VHDL_SOURCES) $(SIM_BUILD)
 	$(CMD) -i $(GHDL_ARGS) $(COMPILE_ARGS) --work=$(RTL_LIBRARY) $(VHDL_SOURCES) && \
 	$(CMD) -m $(GHDL_ARGS) $(COMPILE_ARGS) --work=$(RTL_LIBRARY) $(TOPLEVEL)
 
-$(COCOTB_RESULTS_FILE): analyse  $(COCOTB_LIBS) $(COCOTB_VPI_LIB)
+$(COCOTB_RESULTS_FILE): analyse
```
eric-wieser:
```suggestion
$(COCOTB_RESULTS_FILE): analyse $(CUSTOM_SIM_DEPS)
```
?

cocotb/share/makefiles/simulators/Makefile.icarus
themperek: ?

eric-wieser: No longer relevant after your latest changes.

eric-wieser:
Needs a rebase.

themperek:
>Needs a rebase.

Done

cocotb/share/makefiles/Makefile.sim
```diff
@@ -127,6 +127,9 @@ endif
 # recompilation of the simulation if cocotb is updated.
 CUSTOM_SIM_DEPS += $(shell find $(COCOTB_PY_DIR)/cocotb/ -name "*.py")
 
+# This triggers a recompilation of the simulation if cocotb library is updated.
+CUSTOM_SIM_DEPS += $(shell find $(LIB_DIR)/ -name "*")
```
eric-wieser, line 131: Could include the simulator module here too, but not really a big deal.

PR #1414, opened 2020-02-17
remove unused EXTRA_LIBS and EXTRA_LIBDIRS vars

themperek:
cleanup after #1306

Linking is handled now in `_build_libs.py` during setup using def files. No need to specify this in simulator makefiles.


eric-wieser:
Can you include in the commit message _why_ these can be removed (I think the answer is that this is handled by the windows-specific parts of build_libs, and the def files).

eric-wieser:
Tweaked the commit message a little more, but otherwise looks good, thanks.

Can you start your commits with a capital letter in future?

themperek:
>Tweaked the commit message a little more, but otherwise looks good, thanks.

🥇 

>Can you start your commits with a capital letter in future?

Yes.

PR #1415, opened 2020-02-17
remove cocotb python files from CUSTOM_SIM_DEPS

themperek:
There is no reason for simulation compilation to depend on cocotb python files


themperek:
Yes. But this is not the user would do (modify cocotb files). As a developer one should know what is doing?

eric-wieser:
The user falls into one of two cases:

* They ran `pip install -e <some_dir>`. They have every intent of modifying the files (else they would not use `-e`), and this PR breaks them. I am this type of user.
* They installed via `pip install <some_dir>` or from PyPI. This PR has no effect on them.

Given this PR at best does nothing, and and at worst breaks people developing patches for cocotb, I don't think it's a good idea.

eric-wieser:
Thanks for starting the discussion anyway - it's worth revisiting why things exist occasionally.

PR #1416, opened 2020-02-17
remove PY_SO_LIB and do not try to load python dynamically

themperek:
Since python libraries are always linked during compilation there is no need to try to load it dynamically. Remove unused code.

### DIFF
--- a/cocotb/share/lib/embed/gpi_embed.c
+++ b/cocotb/share/lib/embed/gpi_embed.c
@@ -112,20 +112,9 @@ void embed_init_python(void)
 {
     FENTER;
 
-#ifndef PYTHON_SO_LIB
-#error "Python version needs passing in with -DPYTHON_SO_VERSION=libpython<ver>.so"
-#else
-#define PY_SO_LIB xstr(PYTHON_SO_LIB)
-#endif
-
     // Don't initialize Python if already running
     if (gtstate)
         return;
-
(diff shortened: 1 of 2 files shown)


cocotb/share/lib/embed/gpi_embed.c
```diff
@@ -112,21 +112,10 @@ void embed_init_python(void)
...
-#error "Python version needs passing in with -DPYTHON_SO_VERSION=libpython<ver>.so"
-#else
-#define PY_SO_LIB xstr(PYTHON_SO_LIB)
-#endif
-
     // Don't initialize Python if already running
     if (gtstate)
         return;
 
-    void * lib_handle = utils_dyn_open(PY_SO_LIB);
```
eric-wieser: Does `utils_dyn_open` have any callers remaining?

eric-wieser: Ah yes, it does - the gpi libraries.

themperek: As I see it is used to load libraries specified in `GPI_EXTRA`.

eric-wieser:
Pleased to see this just worked!

themperek:
How to revert this. For whatever reason, this is needed for commercial simulators. I am sorry did not check.

PR #1420, opened 2020-02-18
Ensure that the SimTimeContextFilter is actually applied correctly 

### LINKED ISSUE #1419: Simulator timestamps are broken in logs
Looks like gh-1411 didn't work properly. I'll try and find the fix shortly, but if we want to do a release before I do we should revert it.

eric-wieser:
Fixes #1419.

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -139,9 +139,9 @@ class SimTimeContextFilter(logging.Filter):
     """
 
     # needed to make our docs render well
-    def __init__(self, *args, **kwargs):
-        """ See :class:`logging.Filter` for argument descriptions """
-        super().__init__(*args, **kwargs)
+    def __init__(self):
+        """ Takes no arguments """
+        super().__init__()
 
     def filter(self, record):
         try:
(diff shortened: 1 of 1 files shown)


eric-wieser:
Proof that this now works: https://travis-ci.org/cocotb/cocotb/jobs/652193890#L1731

cocotb/log.py
```diff
@@ -72,17 +72,16 @@ def default_config():
     """
     # construct an appropriate handler
     hdlr = logging.StreamHandler(sys.stdout)
+    hdlr.addFilter(SimTimeContextFilter())
```
imphil: There are some tests in test_cocotb which do something similar, should we align them for consistency, or don't we care? (I assume the latter.)

eric-wieser: I don't understand the question

imphil: I now see why -- it can't make sense :-) I had modified the logging test code locally for my last logging changes and had introduced similar code. But none of that is actually in cocotb, so you can't see what I had in mind :smile:

PR #1423, opened 2020-02-18
Add more .gitignore entries

cmarqu:


themperek:
Can you also add `.tox` folder and there are some files in the `endian_swapper` example generated (try run `tox`).

.gitignore
```diff
@@ -64,3 +67,8 @@ waves.shm/
 
 # Interface libraries built on setup
 cocotb/libs
+pip-wheel-metadata/
+
+# Mentor Modelsim/Questa
```
themperek: `modelsim.ini and transcript are also in examples maybe block globally

themperek: also `*.wlf`

themperek: also some `*.json` files

cmarqu: > `modelsim.ini` and `transcript` are also in examples maybe block globally

I think a user can add local ``modelsim.ini`` files to define libraries etc., so ignoring them globally may be bad.

cmarqu: > also some `*.json` files

Globally, this may be too much. Is this specific to wavedrom maybe?

cmarqu:
> Can you also add `.tox` folder and there are some files in the `endian_swapper` example generated (try run `tox`).

Added .tox, and the ``endian_swapper`` ignores in its own file in that directory.

.gitignore
```diff
@@ -64,3 +68,11 @@ waves.shm/
...
+tests/test_cases/*/modelsim.ini
+tests/test_cases/*/transcript
+examples/test_cases/*/modelsim.ini
```
themperek: `examples/*/tests/modelsim.ini` ?

cmarqu: Good catch, fixed.

examples/endian_swapper/.gitignore
```diff
@@ -0,0 +1,2 @@
+cosim/endian_swapper_hal_wrap.c
+cosim/hal.py
```
themperek:
```suggestion
cosim/hal.py
tests/wavedrom.json
```

cmarqu: Added.

PR #1424, opened 2020-02-19
Implemented PLUSARGS for FLI

### LINKED ISSUE #1404: Passing plusargs to FLI: Find code that works with different TCL versions

ktbarrett:
Fixed version of b283b48368a3a7f507110c6699db3a6c7ce01677. Uses `Tcl_GetStringResult` instead of accessing the internal `result` field, this works on TCL 8.5 and 8.6. Fixes #1404.

Finally got 2019.1 setup and tested this PR against it as well as 10.5c and 10.7c.

I tried using `Tcl_GetVar2Ex` to grab a reference to `argv`, then do processing on it, but there is something weird about how Questa maintains its TCL interpreter because it returns a message that the variable isn't defined. `mti_Cmd` seems to work.

### DIFF
--- a/cocotb/share/lib/fli/FliCbHdl.cpp
+++ b/cocotb/share/lib/fli/FliCbHdl.cpp
@@ -26,6 +26,9 @@
 ******************************************************************************/
 
 #include "FliImpl.h"
+#include "mti.h"
+#include "tcl.h"
+#include <limits>
 
 /**
  * @name    cleanup callback
@@ -140,6 +143,53 @@ int FliStartupCbHdl::arm_callback()
     return 0;
 }
 
+
+static std::vector<std::string> get_argv()
+{
+    /*  Necessary to implement PLUSARGS
+        There is no function available on the FLI to obtain argc+argv directly from the
+        simulator. To work around this we use the TCL interpreter that ships with Questa,
+        some TCL commands, and the TCL variable `argv` to obtain the simulator argc+argv.
+    */
+    std::vector<std::string> argv;
+
+    // obtain a reference to TCL interpreter
+    Tcl_Interp* interp = reinterpret_cast<Tcl_Interp*>(mti_Interp());
+
+    // get argv TCL variable
+    if (mti_Cmd("return -level 0 $argv") != TCL_OK) {
+        const char* errmsg = Tcl_GetStringResult(interp);
+        LOG_WARN("Failed to get reference to argv: %s", errmsg);
+        Tcl_ResetResult(interp);
+        return argv;
+    }
+    Tcl_Obj* result = Tcl_GetObjResult(interp);
+    Tcl_IncrRefCount(result);
+    Tcl_ResetResult(interp);
+
+    // split TCL list into length and element array
+    int argc;
+    Tcl_Obj** tcl_argv;
+    if (Tcl_ListObjGetElements(interp, result, &argc, &tcl_argv) != TCL_OK) {
+        const char* errmsg = Tcl_GetStringResult(interp);
+        LOG_WARN("Failed to get argv elements: %s", errmsg);
+        Tcl_DecrRefCount(result);
+        Tcl_ResetResult(interp);
+        return argv;
+    }
+    Tcl_ResetResult(interp);
+
+    // get each argv arg and copy into internal storage
+    for (int i = 0; i < argc; i++) {
+        const char* arg = Tcl_GetString(tcl_argv[i]);
+        argv.push_back(arg);
+    }
+    Tcl_DecrRefCount(result);
+
+    return argv;
+}
+
+
 int FliStartupCbHdl::run_callback()
 {
     gpi_sim_info_t sim_info;
@@ -166,12 +216,16 @@ int FliStartupCbHdl::run_callback()
     product.push_back('\0');
     version.push_back('\0');
 
-
-    // copy in sim_info.product
-    sim_info.argc = 0;
-    sim_info.argv = NULL;
     sim_info.product = &product[0];
     sim_info.version = &version[0];
+
+    std::vector<std::string> argv = get_argv();
+    std::vector<char*> argv_cstr;
+    for (const auto& arg : argv) {
+        argv_cstr.push_back(const_cast<char*>(arg.c_str()));
+    }
+    sim_info.argc = static_cast<int>(argv.size());
+    sim_info.argv = argv_cstr.data();
 
     gpi_embed_init(&sim_info);
 
--- a/documentation/source/newsfragments/1424.feature.rst
+++ b/documentation/source/newsfragments/1424.feature.rst
@@ -1,0 +1,3 @@
+Questa now supports :envvar:`PLUSARGS`.
+This requires that ``tcl.h`` be present on the system.
+This is likely included in your installation of Questa, otherwise, specify ``CFLAGS=-I/path/to/tcl/includedir``.
--- a/cocotb/share/include/gpi.h
+++ b/cocotb/share/include/gpi.h
@@ -59,12 +59,6 @@ we have to create a process with the signal on the sensitivity list to imitate a
 
 #include <gpi_logging.h>
 
-#if defined(__MINGW32__) || defined (__CYGWIN32__)
-#  define DLLEXPORT __declspec(dllexport)
-#else
-#  define DLLEXPORT
-#endif
-
 #ifdef __cplusplus
 # define EXTERN_C_START extern "C" {
 # define EXTERN_C_END }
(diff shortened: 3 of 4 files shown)


eric-wieser:
Can you use `return -level 0 $argv` in place of `GetVar2Ex`, and then extract items from the output of `GetObjResulr`?

ktbarrett:
Don't merge just yet, need to update newsfragment to right PR and add some deallocation.

@eric-wieser That seems like it would work, I'll try it.

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -140,6 +144,57 @@ int FliStartupCbHdl::arm_callback()
...
+    if (argc_u > std::numeric_limits<int>::max()) {
+        return;
+    }
+    int argc = static_cast<int>(argc_u);
+    if (argc_u == 0) {
+        return;
+    }
+
+    // allocate storage for argv
+    char** argv = new char*[argc];
```
eric-wieser: To avoid the appearance leaking, can you declare this a `static vector<std::string>`, and unpack the pointers into the return value?

The value shouldn't change at runtime anyway, so `static` seems appropriate.

ktbarrett: It doesn't even need to be `static`, it's no longer used after the call to `gpi_embed_init`, I just wasn't cleaning up afterwards.

eric-wieser: Right, but the design of the GPI doesn't even allow you to clean up, does it? Other implementations return non-owned pointers.

ktbarrett: Well it travels up from`gpi_embed_init` to `embed_sim_init`, it gets copied into a Python list and  returns back down the stack. No one holds a reference to it elsewhere. So you could clean up after the call to `gpi_embed_init` in all of the interfaces. You'll see that VHPI cleans up.

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -140,6 +144,53 @@ int FliStartupCbHdl::arm_callback()
...
+    }
+    Tcl_Obj* result = Tcl_GetObjResult(interp);
+    Tcl_IncrRefCount(result);
+    Tcl_ResetResult(interp);
+
+    // split TCL list into length and element array
+    int argc;
+    Tcl_Obj** tcl_argv;
+    if ((Tcl_ListObjGetElements(interp, result, &argc, &tcl_argv) != TCL_OK) || (argc == 0)) {
+        const char* errmsg = Tcl_GetStringResult(interp);
```
eric-wieser: What value will this hold if `argc == 0`?

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -166,13 +217,17 @@ int FliStartupCbHdl::run_callback()
...
 
-
-    // copy in sim_info.product
-    sim_info.argc = 0;
-    sim_info.argv = NULL;
     sim_info.product = &product[0];
     sim_info.version = &version[0];
 
+    std::vector<std::string> argv = get_argv();
+    std::vector<char*> argv_cstr;
```
eric-wieser: I assume we can't make this `char const*`?

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -140,6 +144,53 @@ int FliStartupCbHdl::arm_callback()
...
+    int argc;
+    Tcl_Obj** tcl_argv;
+    if ((Tcl_ListObjGetElements(interp, result, &argc, &tcl_argv) != TCL_OK) || (argc == 0)) {
+        const char* errmsg = Tcl_GetStringResult(interp);
+        LOG_WARN("Failed to get argv elements: %s", errmsg);
+        Tcl_DecrRefCount(result);
+        Tcl_ResetResult(interp);
+        return argv;
+    }
+    Tcl_ResetResult(interp);
```
eric-wieser: Any idea what the result contains here?

cocotb/share/lib/fli/FliCbHdl.cpp
eric-wieser: I think it would be best to split this `if` into two.

documentation/source/newsfragments/1424.feature.rst
```diff
@@ -0,0 +1 @@
+Questa now supports :envvar:`PLUSARGS`. This requires that ``tcl.h`` be present on the system. This is likely included in your installation of Questa. Otherwise, specify ``CXXFLAGS=-I/path/to/tcl/includedir``.
```
eric-wieser: Does this `CXXFLAGS` thing still work with the setup.py changes?

themperek: Yes. All should work.

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -26,6 +26,10 @@
 ******************************************************************************/
 
 #include "FliImpl.h"
+#include "mti.h"
+#undef DLLEXPORT  // redefinition in tcl.h causes warning
```
eric-wieser: Can you clarify where the original definition came from too?

documentation/source/newsfragments/1424.feature.rst
themperek: This works so should change `CFLAGS="-I/path_to_mti_h" pip install -v -e .`

cocotb/share/lib/fli/FliCbHdl.cpp
ktbarrett: sim_info.argv is `char**`, adding const there would require a `const_cast`. But I suppose we should make `sim_info.argv` `const` instead.

cocotb/share/lib/fli/FliCbHdl.cpp
ktbarrett: Ideally nothing if it succeeds, but I wanted to get the interpreter result back into a clean state before leaving as a matter of cleanliness.

documentation/source/newsfragments/1424.feature.rst
ktbarrett: The file that includes tcl.h is a C++ file, it's curious that `CFLAGS` works, or is this due to setuptools not supporting `CXXFLAGS` and only using `CFLAGS`?

themperek: setuptools seems to support CFLAGS and not CXXFLAG

cocotb/share/lib/fli/FliCbHdl.cpp
ktbarrett: I think we can remove the second check. `argc == 0` and the function returns `TCL_OK` only when the list is empty, `tcl_argv == NULL`. However, the way this function is written means we will never dereference that `NULL` pointer.

documentation/source/newsfragments/1424.feature.rst
ktbarrett: Alright, will change.

cocotb/share/lib/fli/FliCbHdl.cpp
ktbarrett: `gpi.h`, but it isn't used anywhere in the project, so we should probably remove it.

cocotb/share/lib/fli/FliCbHdl.cpp
eric-wieser: Let's leave it as is, this is probably motivated by C's bad handling of indirected const.

cocotb/share/lib/fli/FliCbHdl.cpp
ktbarrett: Removed it in a separate commit.

PR #1426, opened 2020-02-19
Do not return the string "OK!" to indicate function success

eric-wieser:
Doing this is pointless, because nothing looks at this string, and these functions can't fail anyway.
If we did want to indicate failure we'd do so with an exception.

This might marginally speed things up, as it stops us creating and destroying strings over and over again.


eric-wieser:
`simulator.log_msg` is probably worth keeping around as a mechanism to test `libgpilog`

PR #1427, opened 2020-02-19
Make the gpi C types stricter

eric-wieser:
This means it is no longer legal to accidentally pass an arbitrary void pointer to gpi functions.
This turned segfaults into compilations errors for me, while working on some other changes.

This also eliminates the need for a lot of internal casting, as the types being cast from and to are already identical.

Making this change exposes some dubious uses of the type system internally - for now, these are plastered over with a TODO comment and a reinterpret_cast.

---

Some discussion about the merits of a similar approach here: https://softwareengineering.stackexchange.com/q/294761/205

### DIFF
--- a/cocotb/share/include/gpi.h
+++ b/cocotb/share/include/gpi.h
@@ -85,8 +85,10 @@ we have to create a process with the signal on the sensitivity list to imitate a
      * incomplete types, as this avoids the need for any casting in GpiCommon.cpp.
      */
     class GpiObjHdl;
+    class GpiCbHdl;
     class GpiIterator;
     typedef GpiObjHdl *gpi_sim_hdl;
+    typedef GpiCbHdl *gpi_cb_hdl;
     typedef GpiIterator *gpi_iterator_hdl;
 #else
     /* In C, we declare some incomplete struct types that we never complete.
@@ -94,8 +96,10 @@ we have to create a process with the signal on the sensitivity list to imitate a
      * names.
      */
     struct GpiObjHdl;
+    struct GpiCbHdl;
     struct GpiIterator;
     typedef struct GpiObjHdl *gpi_sim_hdl;
+    typedef struct GpiCbHdl *gpi_cb_hdl;
     typedef struct GpiIterator *gpi_iterator_hdl;
 #endif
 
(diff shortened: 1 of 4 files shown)


cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -618,22 +605,24 @@ gpi_sim_hdl gpi_register_readwrite_callback(int (*gpi_function)(const void *),
...
     delete(clock);
 }
 
 void gpi_deregister_callback(gpi_sim_hdl hdl)
 {
-    GpiCbHdl *cb_hdl = sim_to_hdl<GpiCbHdl*>(hdl);
+    // TODO: Why are we preteding this is one type when it is another?
+    GpiCbHdl *cb_hdl = reinterpret_cast<GpiCbHdl*>(hdl);
     cb_hdl->m_impl->deregister_callback(cb_hdl);
 }
```
eric-wieser: Probably best to just introduce a `gpi_cb_hdl` and `gpi_clock_hdl` C typedef for this in a follow-up.

cocotb/share/include/gpi.h
```diff
@@ -97,11 +107,13 @@ typedef struct gpi_sim_info_s
...
 
+
+
 // Define a type for our simulation handle.
-typedef void * gpi_sim_hdl;
+typedef struct GpiObjHdl *gpi_sim_hdl;
 
 // Define a handle type for iterators
-typedef void * gpi_iterator_hdl;
+typedef struct GpiIterator *gpi_iterator_hdl;
```
eric-wieser: By doing this, it now becomes a compilation error to pass one in place of the other.

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -618,22 +605,24 @@ gpi_sim_hdl gpi_register_readwrite_callback(int (*gpi_function)(const void *),
...
+    // TODO: Why are we pretending this is one type when it is another?
+    GpiClockHdl *clock = reinterpret_cast<GpiClockHdl*>(clk_object);
     clock->stop_clock();
     delete(clock);
 }
 
 void gpi_deregister_callback(gpi_sim_hdl hdl)
 {
-    GpiCbHdl *cb_hdl = sim_to_hdl<GpiCbHdl*>(hdl);
+    // TODO: Why are we preteding this is one type when it is another?
```
cmarqu:
```suggestion
    // TODO: Why are we pretending this is one type when it is another?
```

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -618,22 +605,24 @@ gpi_sim_hdl gpi_register_readwrite_callback(int (*gpi_function)(const void *),
...
+    GpiObjHdl *clk_hdl = clk_signal;
     GpiClockHdl *clock = new GpiClockHdl(clk_hdl);
     clock->start_clock(period);
     return (gpi_sim_hdl)clock;
 }
 
 void gpi_stop_clock(gpi_sim_hdl clk_object)
 {
-    GpiClockHdl *clock = sim_to_hdl<GpiClockHdl*>(clk_object);
+    // TODO: Why are we pretending this is one type when it is another?
```
eric-wieser: xref gh-1429, let's just delete this code.

cocotb/share/lib/gpi/GpiCommon.cpp
eric-wieser: Actually, I'll update this PR to include that change once #1429 is in

eric-wieser:
@ktbarrett: Are you sure it needs a rebase? Github seems happy as is.

ktbarrett:
I was just mentioning I needed to rebase to test it due to an FLI regression that has since been fixed. But no conflicts.

cocotb/share/include/gpi.h
```diff
@@ -72,7 +72,32 @@ we have to create a process with the signal on the sensitivity list to imitate a
 # define __attribute__(x)
 #endif
 
-
+/*
+ * Declare the handle types.
+ *
+ * We want these handles to be opaque pointers, since their layout is not
+ * exposed to C. We do this by using incomplete types. The assumption being
+ * made here is that `sizeof(some_cpp_class*) == sizeof(some_c_struct*)`, which
```
imphil: You could actually assert for that.

cocotb/share/include/gpi.h
```diff
@@ -72,7 +72,32 @@ we have to create a process with the signal on the sensitivity list to imitate a
...
+ * is true on all reasonable platforms.
+ */
+#ifdef __cplusplus
+/* In C++, we use forward-declarations of the types in gpi_priv.h as our
```
imphil: Can you indent the if/else bodies here to make the conditionals easier to see?

cocotb/share/include/gpi.h
eric-wieser: I couldn't, because what I need to assert is that the compiler behaves the same way in C and C++ mode. I could compromize and hard-code a size in, and assert it at both ends.

cocotb/share/include/gpi.h
```diff
@@ -72,7 +72,36 @@ we have to create a process with the signal on the sensitivity list to imitate a
...
+ * The names of these are irrelevant, but for simplicity they match the C++
+ * names.
+ */
+struct GpiObjHdl;
+struct GpiCbHdl;
+struct GpiIterator;
+typedef struct GpiObjHdl *gpi_sim_hdl;
+typedef struct GpiCbHdl *gpi_cb_hdl;
+typedef struct GpiIterator *gpi_iterator_hdl;
+#endif
```
eric-wieser: Like this, I assume?

```suggestion
#ifdef __cplusplus
    /* In C++, we use forward-declarations of the types in gpi_priv.h as our
     * incomplete types, as this avoids the need for any casting in GpiCommon.cpp.
     */
    class GpiObjHdl;
    class GpiCbHdl;
    class GpiIterator;
    typedef GpiObjHdl *gpi_sim_hdl;
    typedef GpiCbHdl *gpi_cb_hdl;
    typedef GpiIterator *gpi_iterator_hdl;
#else
    /* In C, we declare some incomplete struct types that we never complete.
     * The names of these are irrelevant, but for simplicity they match the C++
     * names.
     */
    struct GpiObjHdl;
    struct GpiCbHdl;
    struct GpiIterator;
    typedef struct GpiObjHdl *gpi_sim_hdl;
    typedef struct GpiCbHdl *gpi_cb_hdl;
    typedef struct GpiIterator *gpi_iterator_hdl;
#endif
```

Will change locally so I can squash the commits correctly.

cocotb/share/include/gpi.h
imphil: ... or only add the assertion in C++ mode. But don't worry, I don't expect to stay this code around forever anyways.

cocotb/share/include/gpi.h
imphil: :+1:

PR #1434, opened 2020-02-21
Remove uses of `_py_compat` which are no longer needed

eric-wieser:
xref #1289

### DIFF
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -28,89 +28,7 @@ Backports and compatibility shims for newer python features.
 These are for internal use - users should use a third party library like `six`
 if they want to use these shims in their own code
 """
-import abc
 import sys
-
-# This is six.integer_types
-if sys.version_info.major >= 3:
-    integer_types = (int,)
-else:
-    integer_types = (int, long)  # noqa
-
-
-# This is essentially six.exec_
-if sys.version_info.major == 3:
-    # this has to not be a syntax error in py2
-    import builtins
-    exec_ = getattr(builtins, 'exec')
-else:
-    # this has to not be a syntax error in py3
-    def exec_(_code_, _globs_=None, _locs_=None):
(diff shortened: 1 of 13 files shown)


cocotb/_py_compat.py
```diff
@@ -133,15 +51,11 @@ def __enter__(self):
...
-        pass
+# On python 3.7 onwards, `dict` is guaranteed to preserve insertion order.
+# Since `OrderedDict` is a little slower that `dict`, we prefer the latter
+# when possible.
+if sys.version_info[:2] >= (3, 7):
+    insertion_ordered_dict = dict
 else:
-    TimeoutError = TimeoutError
+    import collections
+    insertion_ordered_dict = collections.OrderedDict
```
imphil: Can you double-check if OrderedDict isn't making that optimization itself already? (I know you copied this code around, but let's not copy things around which are outdated.)

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1322,5 +1316,4 @@ def test_bad_attr(dut):
         assert False, "Expected AttributeError"
 
 
-if sys.version_info[:2] >= (3, 5):
-    from test_cocotb_35 import *
+from test_cocotb_35 import *
```
imphil: There's no need to keep these test files separate any more I guess? Can we combine them into one?

cocotb/_py_compat.py
eric-wieser: It's not, because `OrderedDict` has extra features to change the order that `dict` does not have. We don't need those features here.

tests/test_cases/test_cocotb/test_cocotb.py
eric-wieser: Yep, that's a separate task in the tracking issue - so I wasn't intending to do it in this PR.

PR #1430, opened 2020-02-20
Define __STDC_FORMAT_MACROS for old libc versions

cmarqu:
The pre-C++11 glibc in e.g. Redhat 6 needs this symbol for the
cinttypes include.
Fixes #1422.

PR #1436, opened 2020-02-21
Selectively include <cinttypes> or "inttypes.h"

### LINKED ISSUE #1422: Compile error in VhpiCbHdl.cpp: expected ‘)’ before ‘PRIu32’
I'm getting the following error:
```
  In file included from cocotb/share/include/gpi.h:60,
                   from cocotb/share/lib/vhpi/../gpi/gpi_priv.h:32,
                   from cocotb/share/lib/vhpi/VhpiImpl.h:31,
                   from cocotb/share/lib/vhpi/VhpiCbHdl.cpp:29:
  cocotb/share/lib/vhpi/VhpiCbHdl.cpp: In member function ‘virtual GpiIterator::Status VhpiIterator::next_handle(std::string&, GpiObjHdl**, void**)’:
  cocotb/share/lib/vhpi/VhpiCbHdl.cpp:1150:69: error: expected ‘)’ before ‘PRIu32’
   1150 |         LOG_DEBUG("Unable to get the name for this object of type " PRIu32, type);
        |                                                                     ^~~~~~
  cocotb/share/include/gpi_logging.h:52:98: note: in definition of macro ‘LOG_DEBUG’
     52 | #define LOG_DEBUG(...)     gpi_log("cocotb.gpi", GPIDebug,         __FILE__, __func__, __LINE__, __VA_ARGS__);
        |                                                                                                  ^~~~~~~~~~~
  cocotb/share/include/gpi_logging.h:52:35: note: to match this ‘(’
     52 | #define LOG_DEBUG(...)     gpi_log("cocotb.gpi", GPIDebug,         __FILE__, __func__, __LINE__, __VA_ARGS__);
        |                                   ^
  cocotb/share/lib/vhpi/VhpiCbHdl.cpp:1150:9: note: in expansion of macro ‘LOG_DEBUG’
   1150 |         LOG_DEBUG("Unable to get the name for this object of type " PRIu32, type);
        |         ^~~~~~~~~
```
This came in via #1361 (https://github.com/cocotb/cocotb/pull/1361/files#diff-8baf84ef545a81134ec388a5b271c5b6R1131).

cmarqu:
This prevents a compilation error on RHEL6.
Fixes #1422.

### DIFF
--- a/cocotb/share/include/vpi_user.h
+++ b/cocotb/share/include/vpi_user.h
@@ -37,7 +37,15 @@ typedef signed __int8 int8_t;
 #elif defined(__MINGW32__)
 #include <stdint.h>
 #elif defined(__linux)
+/* If included first, the <inttypes.h> header causes the <cinttypes> header to
+ * behave incorrectly on RHEL6, failing to define the formatting macros.
+ * We need to ensure other cpp files which include both this file and
+ * <cinttypes> get the macros they expect. See [gh-1422] */
+#ifdef __cplusplus
+#include <cinttypes>
+#else
 #include <inttypes.h>
+#endif
 #else
 #include <sys/types.h>
 #endif


cmarqu:
Alternative to #1430.

eric-wieser:
I assume this also fixes hour problem locally.

Which approach would you prefer to use?

cmarqu:
Yes, it fixes my problem.
I don't have a strong opinion which approach to use, maybe this one is a bit preferable because other projects may reuse our ``vpi_user.h`` but are unlikely to use the ``_build_libs.py``.

eric-wieser:
Another option would be to just `#define __STDC_FORMAT_MACROS` in this file.

cocotb/share/include/vpi_user.h
```diff
@@ -37,7 +37,12 @@ typedef signed __int8 int8_t;
 #elif defined(__MINGW32__)
 #include <stdint.h>
 #elif defined(__linux)
+/* the inttypes.h header is contagiously broken on RHEL6, see [gh-1422] */
```
eric-wieser:
```suggestion
/* If included first, the <inttypes.h> header causes the <cinttypes> header to
 * behave incorrectly on RHEL6, failing to define the formatting macros. We need to ensure other cpp
 * files include which inlucde both this file and <cinttypes> get the macros they expect. See [gh-1422] */
```

cmarqu:
@imphil Do you want to be the arbiter on the method being used?

cocotb/share/include/vpi_user.h
cmarqu: Added.

imphil:
Let's take https://github.com/cocotb/cocotb/pull/1430 instead. Thanks for your patch @cmarqu!

PR #1440, opened 2020-02-22
Remove the use of `ReturnValue` internally

eric-wieser:
Now that all supported python versions are newer than 3.3, `return` works natively in coroutines
This may give a marginal performance improvement.

For now, we leave behind the `ReturnValue`s in our tests.

xref #1289


eric-wieser:
It's one of the items in the tracking issue linked above, yes

Issue #1131, opened 2019-09-28
Conda Environment "Did not detect Python"


dan-fritchman:
It seems that cocotb is tripped up by finding the Python interpreter in a Conda environment. 

To reproduce: 
* Create a fresh conda env
* Activate said conda env
* Follow the [Quickstart](https://github.com/cocotb/cocotb#quickstart) instructions 

Commands:

```
conda create --name cocotb-py37-env python=3.7
conda activate cocotb-py37-env
git clone https://github.com/cocotb/cocotb
pip install ./cocotb
cd cocotb/examples/endian_swapper/tests
make
```

Generates these errors:

```
     -.--ns INFO     cocotb.gpi                                  gpi_embed.c:78   in set_program_name_in_venv        Did not detect Python virtual environment. Using system-wide Python interpreter
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
Fatal Python error: initfsencoding: unable to load the file system codec
ModuleNotFoundError: No module named 'encodings'
```

themperek:
@dan-fritchman This should not happen and I cannot reproduce this. Can you try with the latest version?

ktbarrett:
This should be resolved now, if it is not, please re-open.

PR #1441, opened 2020-02-22
Add --python-bin to cocotb-config

themperek:
I am working on better multi-os support and makefile cleanup and trying with a series of small PRs. 

This will become very handy to simplify Makefiles and set PYTHONHOME variable that will make much nicer windows support and fix issues like #1131

Based on work by Philipp Wagner <imphil@philipp-wagner.com> in #990

### DIFF
--- a/cocotb/config.py
+++ b/cocotb/config.py
@@ -38,17 +38,16 @@ Global variables:
     share_dir: str, path where the cocotb data is stored
     makefiles_dir: str, path where the cocotb makefiles are installed
 """
+import argparse
 import os
 import sys
 import cocotb
-import argparse
-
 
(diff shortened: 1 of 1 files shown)


cocotb/config.py
```diff
@@ -61,26 +60,50 @@ def __call__(self, parser, namespace, values, option_string=None):
         parser.exit()
 
 
-def get_parser():
```
eric-wieser: Please add back `get_parser`, it's needed to make the docs work.

themperek: :+1:

eric-wieser:
Let's just put this in, now has come and gone

PR #1442, opened 2020-02-22
Convert all existing C files to C++ files

ktbarrett:
This will allow us to use the extra features that C++ offers through the entire code base.

Moving the compilation standard from gnu99 to c++11 was mostly painless, there were a few cases of gotos jumping over variable initialization.

### DIFF
--- a/cocotb/share/lib/embed/gpi_embed.cpp
+++ b/cocotb/share/lib/embed/gpi_embed.cpp
@@ -108,7 +108,7 @@ static void set_program_name_in_venv(void)
  * Stores the thread state for cocotb in static variable gtstate
  */
 
-void embed_init_python(void)
+extern "C" void embed_init_python(void)
 {
     FENTER;
 
@@ -174,7 +174,7 @@ out:
  *
  * Cleans up reference counts for Python objects and calls Py_Finalize function.
  */
-void embed_sim_cleanup(void)
+extern "C" void embed_sim_cleanup(void)
 {
     // If initialization fails, this may be called twice:
     // Before the initial callback returns and in the final callback.
@@ -219,7 +219,7 @@ int get_module_ref(const char *modname, PyObject **mod)
     return 0;
 }
 
-int embed_sim_init(gpi_sim_info_t *info)
+extern "C" int embed_sim_init(gpi_sim_info_t *info)
 {
     FENTER
 
@@ -239,7 +239,7 @@ int embed_sim_init(gpi_sim_info_t *info)
             dut = NULL;
         } else {
             // Skip any library component of the toplevel
-            char *dot = strchr(dut, '.');
+            const char *dot = strchr(dut, '.');
             if (dot != NULL) {
                 dut += (dot - dut + 1);
             }
@@ -346,26 +346,28 @@ int embed_sim_init(gpi_sim_info_t *info)
     }
 
     // Set language in use as an attribute to cocotb module, or None if not provided
-    const char *lang = getenv("TOPLEVEL_LANG");
-    PyObject *PyLang;
-    if (lang) {
-        PyLang = PyUnicode_FromString(lang);                            // New reference
-    } else {
-        Py_INCREF(Py_None);
-        PyLang = Py_None;
-    }
-    if (PyLang == NULL) {
-        PyErr_Print();
-        LOG_ERROR("Unable to create Python object for cocotb.LANGUAGE");
-        goto cleanup;
-    }
-    if (-1 == PyObject_SetAttrString(cocotb_module, "LANGUAGE", PyLang)) {
-        PyErr_Print();
-        LOG_ERROR("Unable to set LANGUAGE");
+    {
+        const char *lang = getenv("TOPLEVEL_LANG");
+        PyObject *PyLang;
+        if (lang) {
+            PyLang = PyUnicode_FromString(lang);                            // New reference
+        } else {
+            Py_INCREF(Py_None);
+            PyLang = Py_None;
+        }
+        if (PyLang == NULL) {
+            PyErr_Print();
+            LOG_ERROR("Unable to create Python object for cocotb.LANGUAGE");
+            goto cleanup;
+        }
+        if (-1 == PyObject_SetAttrString(cocotb_module, "LANGUAGE", PyLang)) {
+            PyErr_Print();
+            LOG_ERROR("Unable to set LANGUAGE");
+            Py_DECREF(PyLang);
+            goto cleanup;
+        }
         Py_DECREF(PyLang);
-        goto cleanup;
     }
-    Py_DECREF(PyLang);
 
     pEventFn = PyObject_GetAttrString(cocotb_module, "_sim_event");     // New reference
     if (pEventFn == NULL) {
@@ -433,7 +435,7 @@ ok:
     return ret;
 }
 
-void embed_sim_event(gpi_event_t level, const char *msg)
+extern "C" void embed_sim_event(gpi_event_t level, const char *msg)
 {
     FENTER
     /* Indicate to the upper layer a sim event occurred */
--- a/cocotb/share/lib/gpi_log/gpi_logging.cpp
+++ b/cocotb/share/lib/gpi_log/gpi_logging.cpp
@@ -35,29 +35,29 @@ static PyObject *pLogHandler = NULL;
 static PyObject *pLogFilter = NULL;
 static enum gpi_log_levels local_level = GPIInfo;
 
-void set_log_handler(void *handler)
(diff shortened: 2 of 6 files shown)


ktbarrett:
Changed the way I fixed the jump over initialization errors. Smaller diffs.

ktbarrett:
Had to fix the Windows and Mac builds.

cocotb/share/lib/gpi_log/gpi_logging.cpp
```diff
@@ -185,12 +185,14 @@ static void gpi_log_v(const char *name, enum gpi_log_levels level, const char *p
         goto error;
     }
 
-    PyObject *filter_ret = PyObject_CallFunctionObjArgs(pLogFilter, logger_name_arg, level_arg, NULL);
+    PyObject *filter_ret;
+    filter_ret = PyObject_CallFunctionObjArgs(pLogFilter, logger_name_arg, level_arg, NULL);
```
ktbarrett: This might look a little weird, but it makes sense. You cannot jump over a variable initialization, but you can jump over variable declarations *if* they have trivial default constructors. Value initialization calls a non-default constructor, so it falls into the "ill-formed" category. The language could probably support what we are trying to do by checking for a trivial default constructor and trivial assignment, but it doesn't...

ktbarrett:
@eric-wieser I'm guessing this is waiting on #1427?

eric-wieser:
Ideally, yes

eric-wieser:
That PR is now in, would appreciate if you could rebase

eric-wieser:
I'm a little wary of this change, primarily because:
* I don't know the motivation for using C here in the first place.
* It's unusual for me to see python extension modules written in C++ (except via pybind11)

I'm not sure either of these are real concerns though, and it certainly would be convenient to work in just one language.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -173,10 +174,10 @@ int handle_gpi_callback(void *user_data)
         sim_ending = 1;
         ret = 0;
         goto out;
-    }
 
     // Free up our mess
     Py_DECREF(pValue);
+    }
```
eric-wieser: This looks like a bug

ktbarrett:
@eric-wieser I've been making some changes to the GPI and could really use C++ features in many places. I can't say why anyone would prefer to make extensions in C or why it was done using C in this project besides obliviousness, or maybe lack of familiarity with C++. I still think that the GPI interface should remain a C interface.

It's funny you mention pybind11, is there any hope in switching to it for our binding purposes? It also supports interpreter embedding. gpi_embed and the simulator module could be replaced. It would not end up being a runtime requirement. I'm rather familiar with it, but don't have any experience using it with distutils.

eric-wieser:
I would certainly be keen on trying pybind11 at some point, I think it would make #1332 possible without proliferating the number of different interfaces, and would make #1540 easier too.
I'd just like to caution you about trying to change too much at once, as it means reviewer time will be spread even more thinly across your many open PRs. Top of my list of your PRs right now is #1523 - I'd appreciate if you could check the test I linked to from there looks ok, so we can wrap that one up.

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -346,7 +346,8 @@ int embed_sim_init(gpi_sim_info_t *info)
     }
 
     // Set language in use as an attribute to cocotb module, or None if not provided
-    const char *lang = getenv("TOPLEVEL_LANG");
+    const char *lang;
+    lang = getenv("TOPLEVEL_LANG");
     PyObject *PyLang;
```
eric-wieser: If you wrapped everything from line 349 to line 369 in a set of braces, then this change wouldn't be needed, and it would nicely show that these variables are gone before reaching the cleanup.

cocotb/share/lib/gpi_log/gpi_logging.cpp
```diff
@@ -185,12 +185,14 @@ static void gpi_log_v(const char *name, enum gpi_log_levels level, const char *p
...
     if (filter_ret == NULL) {
         goto error;
     }
 
-    int is_enabled = PyObject_IsTrue(filter_ret);
+    int is_enabled;
+    is_enabled = PyObject_IsTrue(filter_ret);
     Py_DECREF(filter_ret);
```
eric-wieser: Wrapping these lines in braces would allow you to keep `filter_ret` on a single line.

You'd need to pull `is_enabled` out of the braces, but I think it would overall add clarity

cocotb/share/lib/simulator/simulatormodule.h
```diff
@@ -130,4 +130,6 @@ static PyMethodDef SimulatorMethods[] = {
     {NULL, NULL, 0, NULL}        /* Sentinel */
 };
 
+PyMODINIT_FUNC PyInit_simulator(void);
+
```
eric-wieser: Why did we get away with not having this before? Does this need to be `extern "C"`?

eric-wieser: Looks like python adds the extern C for us.

ktbarrett: That was a unintended commit missed after a lot of rebasing. It's been removed. This whole header file is pointless...

cocotb/share/lib/embed/gpi_embed.cpp
ktbarrett: I had originally done this, but changed it to prefer a smaller diff, but I tend to think getting local as possible with variable declarations is a good idea.

eric-wieser: The diff is still small if you tell github to ignore whitespace

cocotb/share/lib/gpi_log/gpi_logging.cpp
```diff
@@ -229,12 +232,13 @@ static void gpi_log_v(const char *name, enum gpi_log_levels level, const char *p
         goto error;
     }
 
-    // Log function args are logger_name, level, filename, lineno, msg, function
-    PyObject *handler_ret = PyObject_CallFunctionObjArgs(pLogHandler, logger_name_arg, level_arg, filename_arg, lineno_arg, msg_arg, function_arg, NULL);
-    if (handler_ret == NULL) {
-        goto error;
+    {   // Log function args are logger_name, level, filename, lineno, msg, function
+        PyObject *handler_ret = PyObject_CallFunctionObjArgs(pLogHandler, logger_name_arg, level_arg, filename_arg, lineno_arg, msg_arg, function_arg, NULL);
```
eric-wieser: Style nit:
```suggestion
    {
        // Log function args are logger_name, level, filename, lineno, msg, function
        PyObject *handler_ret = PyObject_CallFunctionObjArgs(pLogHandler, logger_name_arg, level_arg, filename_arg, lineno_arg, msg_arg, function_arg, NULL);
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -169,31 +169,32 @@ int handle_gpi_callback(void *user_data)
...
-    // calls will go back to Python which is now in an unknown state
-    if (pValue == NULL)
-    {
-        fprintf(stderr, "ERROR: called callback function returned NULL\n");
-        if (PyErr_Occurred()) {
-            fprintf(stderr, "Failed to execute callback due to Python exception\n");
-            PyErr_Print();
-        } else {
-            fprintf(stderr, "Failed to execute callback\n");
+    { // Call the callback
```
eric-wieser: style nit:
```suggestion
    // Call the callback
    {
```

cocotb/share/lib/gpi_log/gpi_logging.cpp
```diff
@@ -185,20 +185,23 @@ static void gpi_log_v(const char *name, enum gpi_log_levels level, const char *p
...
-    int is_enabled = PyObject_IsTrue(filter_ret);
-    Py_DECREF(filter_ret);
-    if (is_enabled < 0) {
-        /* A python exception occured while converting `filter_ret` to bool */
-        goto error;
-    }
+        int is_enabled = PyObject_IsTrue(filter_ret);
+        Py_DECREF(filter_ret);
+        if (is_enabled < 0) {
+            /* A python exception occured while converting `filter_ret` to bool */
```
cmarqu:
```suggestion
            // A Python exception occured while converting `filter_ret` to bool
```

eric-wieser: Let's do that in a follow-up, it'll just make the diff harder to read.

PR #1444, opened 2020-02-23
Fix missing function arguments in FLI after #1403

themperek:
We, unfortunately, we cannot build FLI in CI and `gpi_set_action_t action` was missing in #1403 .

### DIFF
--- a/cocotb/share/lib/fli/FliImpl.h
+++ b/cocotb/share/lib/fli/FliImpl.h
@@ -353,7 +353,7 @@ public:
     long get_signal_value_long() override;
 
     using FliValueObjHdl::set_signal_value;
-    int set_signal_value(long value) override;
+    int set_signal_value(long value, gpi_set_action_t action) override;
 
     int initialise(std::string &name, std::string &fq_name) override;
 };


themperek:
> Nice catch. I wonder if we can compile the header alone at CI time?

Without `mti.h` may be hard?

BTW: I see quite a bit of traffic from Mentor, maybe they would be willing to provide `mti.h` with some nice license. We could also copy `mti.h` in CI only from here:  https://github.com/grubbymits/esdg-opencl/blob/master/install-dir/include/insizzle/mti.h but this may not be ...

eric-wieser:
> BTW: I see quite a bit of traffic from Mentor, maybe they would be willing to provide `mti.h` with some nice license.

That would be really great - even if the license doesn't let us distribute it via PyPI, having it somewhere special just for ci would be super valuable.

ktbarrett:
@themperek I had thought about this as well. If you want, I can add my company's name and a few senior engineer's cotontact info to the document. We spend WAY to much on them to ignore us.

themperek:
Maybe someone can fill a support request and ask for providing `mti.h` with some reasonable license?

ktbarrett:
I'm mostly clueless when it comes to licensing. Is there a license they might find reasonable?

themperek:
I am also quite clueless  but at the moment it looks like nogo to me, see:

https://github.com/grubbymits/esdg-opencl/blob/master/install-dir/include/insizzle/mti.h

eric-wieser:
It;'s possible that the [copyright on headers is unenforceable](https://softwareengineering.stackexchange.com/q/216471/205)

ktbarrett:
That does change things. But I'd still rather get Mentor's blessing than have to beg the FSF for legal protection.

PR #1445, opened 2020-02-23
Better back compatibility and use cocotb-config

themperek:
Prior #1304 it was possible to use "previous file structure" without installation now it is not possible. 

This PR:
 - add a warning message about deprecated file structure
 - add error message if cocotb is not installed 
 - use cocotb-config instead of MAKEFILE_LIST so "previous file structure" still works better (independent of makefile structure)  and for (future) better multi-os support

### DIFF
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -29,20 +29,12 @@
 
 # Common makefile included by everything
 
-# Directory containing the cocotb Python module
-ifeq ($(COCOTB_PY_DIR),)
-export COCOTB_PY_DIR:=$(realpath $(dir $(lastword $(MAKEFILE_LIST)))/../../..)
-endif
+# Directory containing the cocotb Python module (realpath for Windows compatibility)
+COCOTB_PY_DIR := $(realpath $(shell cocotb-config --prefix))
 
 # Directory containing all support files required to build cocotb-based
 # simulations: Makefile fragments, and the simulator libraries.
-ifeq ($(COCOTB_SHARE_DIR),)
-export COCOTB_SHARE_DIR:=$(realpath $(dir $(lastword $(MAKEFILE_LIST)))/..)
-endif
-
-ifeq ($(COCOTB_PY_DIR),$(COCOTB_SHARE_DIR))
-$(warning 'Deprecated cocotb file structure detected. Please use "include $$(shell cocotb-config --makefile)/Makefile.inc" in Makefiles')
-endif
+COCOTB_SHARE_DIR := $(COCOTB_PY_DIR)/cocotb/share
 
 ifeq ($(USER_DIR),)
 export USER_DIR:=$(realpath $(dir $(firstword $(MAKEFILE_LIST))))
--- a/documentation/source/newsfragments/1445.change.rst
+++ b/documentation/source/newsfragments/1445.change.rst
@@ -1,0 +1,1 @@
+Cocotb must now be :ref:`installed <installation_via_pip>` before it can be used.
--- a/examples/endian_swapper/cosim/Makefile
+++ b/examples/endian_swapper/cosim/Makefile
@@ -1,5 +1,5 @@
-include $(COCOTB_SHARE_DIR)/makefiles/Makefile.pylib
-include $(COCOTB_SHARE_DIR)/makefiles/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.pylib
 
 PYTHON_LIBDIR ?= /usr/lib64
 SWIG ?= swig
--- a/makefiles/Makefile.inc
+++ b/makefiles/Makefile.inc
@@ -27,10 +27,14 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ##############################################################################
 
-# Backwards-compatibility wrapper for people using
-# "include $COCOTB/makefiles/Makefile.inc" in their Makefile.
-COCOTB_INSTALL_METHOD = srctree
-export COCOTB_INSTALL_METHOD
+COCOTB_CONFIG_CMD := $(shell :; command -v cocotb-config)
+ifeq ($(COCOTB_CONFIG_CMD),)
+    $(error 'Cocotb is not installed (cocotb-config not found). \
+             Please refer to https://cocotb.readthedocs.io/en/latest/quickstart.html#installing-cocotb \
+             for installation instructions.')
+endif
 
-_NEW_SHARE_DIR := $(realpath $(dir $(lastword $(MAKEFILE_LIST))))/../cocotb/share
-include $(_NEW_SHARE_DIR)/makefiles/Makefile.inc
+$(error 'Deprecated cocotb file structure detected. \
+         Use "include $$(shell cocotb-config --makefile)/Makefile.inc" instead. \
+         Please refer to https://cocotb.readthedocs.io/en/latest/quickstart.html#creating-a-makefile \
+         for instructions.')
--- a/makefiles/Makefile.sim
+++ b/makefiles/Makefile.sim
@@ -27,13 +27,14 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ##############################################################################
 
-# Backwards-compatibility wrapper for people using
-# "include $COCOTB/makefiles/Makefile.sim" in their Makefile.
+COCOTB_CONFIG_CMD := $(shell :; command -v cocotb-config)
+ifeq ($(COCOTB_CONFIG_CMD),)
+    $(error 'Cocotb is not installed (cocotb-config not found). \
+             Please refer to https://cocotb.readthedocs.io/en/latest/quickstart.html#installing-cocotb \
+             for installation instructions.')
+endif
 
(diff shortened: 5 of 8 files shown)


makefiles/Makefile.inc
```diff
@@ -29,6 +29,11 @@
 
 # Backwards-compatibility wrapper for people using
 # "include $COCOTB/makefiles/Makefile.inc" in their Makefile.
+
+$(warning 'Deprecated cocotb file structure detected. Please install cocotb \
+(see documentation) and set "COCOTB:=$$(shell cocotb-config --share)" \
```
eric-wieser: Should this say:

```suggestion
(see documentation) and set "COCOTB_SHARE_DIR :=$$(shell cocotb-config --share)" \
```
Or is this advice now just wrong and can be removed?

themperek: As I know in the past we used `COCOTB` everyware (still used in tests). I think advice should stay and we may remove those Makefiles at some point.

examples/endian_swapper/cosim/Makefile
```diff
@@ -1,5 +1,5 @@
-include $(COCOTB_SHARE_DIR)/makefiles/Makefile.pylib
-include $(COCOTB_SHARE_DIR)/makefiles/Makefile.inc
+include $(shell cocotb-config --makefile)/Makefile.pylib
+include $(shell cocotb-config --makefile)/Makefile.inc
```
ktbarrett:
```suggestion
include $(shell cocotb-config --makefiles)/Makefile.inc
```

makefiles/Makefile.inc
eric-wieser: Can you give an example of where we use / used to use the `$COCOTB` environment variable?

themperek: https://github.com/cocotb/cocotb/blob/master/tests/designs/sample_module/Makefile#L54

See v1.1 docs: https://cocotb.readthedocs.io/en/v1.1/quickstart.html#creating-a-makefile

examples/endian_swapper/cosim/Makefile
```diff
@@ -1,5 +1,5 @@
 include $(shell cocotb-config --makefile)/Makefile.pylib
```
eric-wieser:
```suggestion
include $(shell cocotb-config --makefiles)/Makefile.pylib
```

makefiles/Makefile.inc
eric-wieser: Thanks!

makefiles/Makefile.inc
```diff
@@ -29,6 +29,11 @@
...
+or "include $$(shell cocotb-config --makefile)/Makefile.sim" in Makefiles')
+
 COCOTB_INSTALL_METHOD = srctree
 export COCOTB_INSTALL_METHOD
```
eric-wieser: Is this used anywhere?

cocotb/share/makefiles/Makefile.inc
```diff
@@ -29,20 +29,18 @@
...
 # Common makefile included by everything
 
-# Directory containing the cocotb Python module
-ifeq ($(COCOTB_PY_DIR),)
-export COCOTB_PY_DIR:=$(realpath $(dir $(lastword $(MAKEFILE_LIST)))/../../..)
+# Make sure cocotb is installed
+COCOTB_CONFIG_CMD := $(shell :; command -v cocotb-config)
+ifeq ($(COCOTB_CONFIG_CMD),)
+    $(error "Cocotb not installed (cocotb-config not found). Please install cocotb (see documentation for details).")
 endif
```
eric-wieser: I think this belongs in the other `Makefile.inc`. If the only way users are including this file is via `cocotb-config --makefiles`, then the fact this was included at all is an indication that it's already installed.

makefiles/Makefile.inc
themperek: I can not see this anywhere. I think added by @imphil

cocotb/share/makefiles/Makefile.inc
themperek: In this case, this https://github.com/cocotb/cocotb/pull/1445/files#diff-9b99a3d072edd3bd56cffb557c7d942fR33 would need to be an error and sing other Makefile.inc would be not possible (ok for me).

makefiles/Makefile.inc
```diff
@@ -29,6 +29,11 @@
 
 # Backwards-compatibility wrapper for people using
 # "include $COCOTB/makefiles/Makefile.inc" in their Makefile.
+
+$(warning 'Deprecated cocotb file structure detected. Please install cocotb \
+(see documentation) and set "COCOTB:=$$(shell cocotb-config --share)" \
+or "include $$(shell cocotb-config --makefile)/Makefile.sim" in Makefiles')
```
eric-wieser:
```suggestion
# Make sure cocotb is installed
COCOTB_CONFIG_CMD := $(shell :; command -v cocotb-config)
ifeq ($(COCOTB_CONFIG_CMD),)
    $(error "Cocotb not installed (cocotb-config not found). Please install cocotb (see documentation for details).")
endif

$(warning 'Deprecated cocotb file structure detected. Using "include $$(shell cocotb-config --makefile)/Makefile.sim" instead')
include $(shell cocotb-config --makefile)/Makefile.sim
```

cocotb/share/makefiles/Makefile.inc
eric-wieser: It's already not possible to use the other makefile. One of three cases happens as you have this patch:

1. Cocotb is not installed. Whether the user uses old or new `Makefile.inc`, they're going to hit the error
2. Cocotb is a regular install. The old `Makefile.inc` includes the new `Makefile.inc` from the source tree, which then goes on to use submakefiles from the installation
3. Cocotb is a development install. The above happens, but installation and source tree are equivalent.

Case 2 is just confusion waiting to happen. If old `Makefile.inc` always redirects to the _installed_ new `Makefile.inc`, case 2 becomes much more sensible.

makefiles/Makefile.inc
cmarqu: Missing the "is" in the suggestion: ``$(error "Cocotb is not installed``

eric-wieser:
This needs a release note (`.change`, probably) saying that cocotb must now be installed via `pip` or `setup.py` before use.

documentation/source/newsfragments/1445.change.rst
```diff
@@ -0,0 +1 @@
+Cocotb must now be :ref:`installed <installation-via-pip>` before it can be use.
```
eric-wieser:
```suggestion
Cocotb must now be :ref:`installed <installation-via-pip>` before it can be used.
```

themperek: Do you know why this `ref` does not work?

eric-wieser: Can you point me to the rst you're trying to link to?

cmarqu:
The angle brackets are too much.

documentation/source/newsfragments/1445.change.rst
themperek: I got this sorry ...

eric-wieser:
I think the error message should be tailored to cover:
* cocotb is installed, but the user is including the wrong makefile
* cocotb is installed, but cocotb-config is not on the path
* cocotb is not installed

themperek:
> I think the error message should be tailored to cover:
> 
> * cocotb is installed, but the user is including the wrong makefile
> * cocotb is installed, but cocotb-config is not on the path
> * cocotb is not installed

I do not see a way to distinguish between 2 & 3  but will try to do something.

eric-wieser:
I don't think you need to distinguish 2 and 3 provided you emit a message that covers both cases.

themperek:
@eric-wieser @imphil Better now?

makefiles/Makefile.sim
```diff
@@ -27,13 +27,14 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ##############################################################################
 
-# Backwards-compatibility wrapper for people using
-# "include $COCOTB/makefiles/Makefile.sim" in their Makefile.
+COCOTB_CONFIG_CMD := $(shell :; command -v cocotb-config)
+ifeq ($(COCOTB_CONFIG_CMD),)
+    $(error 'Cocotb is not installed (cocotb-config not found). \
+             Please refer to https://cocotb.readthedocs.io/en/latest/quickstart.html#installing-cocotb \
```
eric-wieser, line 33: Does this link mention adding it to the path?

themperek: No. I originally had this error in  `cocotb/share/makefiles/Makefile.inc`. Maybe it should go there also? 

The problem may be for people who use '--user' (that is not in our documentation).

themperek: It would be ok to make an issue for me and add NOTE in the documentation?

imphil:
Yes, please add it to the docs (or file an issue for it), it's too involved to put all of it into an error message. No need to do it as part of this PR. This PR can be merged.

makefiles/Makefile.sim
eric-wieser: I've also had this problem on systems where python itself is not on the path, typically Windows if the "add to path" box in the installer is not checked.

My own system is set up this way. Python 3.5 is on the path, but cocotb is installed into 3.8 which is not.

All I'd suggest is replacing "not installed" with "not installed (or installed but not on the PATH)

themperek: You are right. I will do a patch immediately. Sorry.

Issue #1443, opened 2020-02-22
Passing simulator options (flags) to Icarus Verilog


mballance:
Icarus Verilog makes a strong distinction between simulator options and simulation arguments. The first come before specification of the .vvp file, the later after. The cocotb Makefiles allow adding simulation arguments via the SIM_ARGS, EXTRA_ARGS, and PLUSARGS variables. However, they do not allow adding options to vvp itself.
Would having a SIM_FLAGS variable make sense? I'm envisioning something like the following for the VVP command in the simulation rule:

```
	PYTHONPATH=$(LIB_DIR):$(PWD):$(NEW_PYTHONPATH) $(LIB_LOAD) MODULE=$(MODULE) \
        TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) TOPLEVEL_LANG=$(TOPLEVEL_LANG) COCOTB_SIM=1 \
        $(ICARUS_BIN_DIR)/vvp -M $(LIB_DIR) -m gpivpi $(SIM_FLAGS) $(SIM_BUILD)/sim.vvp $(SIM_ARGS) $(EXTRA_ARGS) $(PLUSARGS)
```

I've added $(SIM_FLAGS) reference directly after the gpivpi VPI module is loaded. 

Is there another way to do this? If not, would it make sense to have this capability?

themperek:
To me, it looks like `$(SIM_ARGS) $(EXTRA_ARGS)` should moved before `$(SIM_BUILD)/sim.vvp`  rather than adding new variables. What is the use of fo `$(SIM_ARGS)` at the current position?

mballance:
`$(SIM_ARGS)` and `$(EXTRA_ARGS)` allow non-plusarg arguments to be passed to VVP. Not entirely sure if this is useful, since it's not generally easy to get access to non-plusargs in simulation.

I think a fine solution would be to move `$(SIM_ARGS)` and `$(EXTRA_ARGS)` ahead of the .vvp file as you suggest. This would be consistent with cocotb's support for other simulators.

Should I create a pull request with this change?

themperek:
I also think this makes much more sense and how this now looks like a mistake.

It was added here: https://github.com/cocotb/cocotb/commit/bd579dc4f62810441bf2c6e9470df4e5549d4feb

PR #1447, opened 2020-02-23
Rearrange order of SIM_ARGS, EXTRA_ARGS to match VVP restrictions

mballance:
VVP requires simulator options and arguments (those that control behavior of the VVP simulator) to be passed before the .vvp file. Based on the documentation, SIM_ARGS are:
```
Any arguments or flags to pass to the execution of the compiled simulation.  Only applies to simulators with a separate compilation stage (currently Icarus and VCS).
```

In the current order, simulator options (eg -M/-m to load VPI libraries) cannot be controlled via the cocotb Makefiles. 

The change in this PR, as discussed with @themperek in #1443, is to pass `$(SIM_ARGS)` and `$(EXTRA_ARGS)` before the .vvp file is specified.

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.icarus
+++ b/cocotb/share/makefiles/simulators/Makefile.icarus
@@ -78,7 +78,7 @@ $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/sim.vvp $(CUSTOM_SIM_DEPS) $(COCOTB_LIBS) $
 
 	PYTHONPATH=$(LIB_DIR):$(PWD):$(NEW_PYTHONPATH) $(LIB_LOAD) MODULE=$(MODULE) \
         TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) TOPLEVEL_LANG=$(TOPLEVEL_LANG) COCOTB_SIM=1 \
-        $(ICARUS_BIN_DIR)/vvp -M $(LIB_DIR) -m libcocotbvpi $(SIM_BUILD)/sim.vvp $(SIM_ARGS) $(EXTRA_ARGS) $(PLUSARGS)
+        $(ICARUS_BIN_DIR)/vvp -M $(LIB_DIR) -m libcocotbvpi $(SIM_ARGS) $(EXTRA_ARGS) $(SIM_BUILD)/sim.vvp $(PLUSARGS)
 
 	# check that the file was actually created, since we can't set an exit code from cocotb
 	test -f $(COCOTB_RESULTS_FILE)
@@ -88,7 +88,7 @@ debug: $(SIM_BUILD)/sim.vvp $(CUSTOM_SIM_DEPS) $(COCOTB_LIBS)
 
 	PYTHONPATH=$(LIB_DIR):$(PWD):$(NEW_PYTHONPATH) $(LIB_LOAD) MODULE=$(MODULE) \
         TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) TOPLEVEL_LANG=$(TOPLEVEL_LANG) COCOTB_SIM=1 \
-        gdb --args $(ICARUS_BIN_DIR)/vvp -M $(LIB_DIR) -m libcocotbvpi $(SIM_BUILD)/sim.vvp $(SIM_ARGS) $(EXTRA_ARGS) $(PLUSARGS)
+        gdb --args $(ICARUS_BIN_DIR)/vvp -M $(LIB_DIR) -m libcocotbvpi $(SIM_ARGS) $(EXTRA_ARGS) $(SIM_BUILD)/sim.vvp $(PLUSARGS)
 
 	# check that the file was actually created, since we can't set an exit code from cocotb
 	test -f $(COCOTB_RESULTS_FILE)


ktbarrett:
Does this change the contents of `cocotb.argv`?

mballance:
> Does this change the contents of `cocotb.argv`?

The order of arguments will change, but all the arguments that were present before will still be there with this change. 

`cocotb.argv` is set based on the content of  `argv` returned by `vpi_get_vlog_info`.  The arguments returned by `vpi_get_vlog_info` are the entire simulator command line, including tool-specific options.

themperek:
Some "interesting" error here: https://travis-ci.org/cocotb/cocotb/jobs/654110146#L2112-L2133

themperek:
@eric-wieser CI error related to #1440 ?

mballance:
> @eric-wieser CI error related to #1440 ?

Several of the tests (I'm looking at test_plusargs) look like they're trying to add -v93 to SIM_ARGS if the simulator is NCSIM (ius, xcelium):

```
ifeq ($(SIM),$(filter $(SIM),ius xcelium))
    SIM_ARGS += -v93
endif
```

The core ius and Makefiles also adds this to SIM_ARGS:
```
GPI_EXTRA:=
    GPI_ARGS = -loadvpi $(LIB_DIR)/libvpi
    EXTRA_ARGS += -v93
    HDL_SOURCES = $(VERILOG_SOURCES)
    ROOT_LEVEL = $(TOPLEVEL)
ifneq ($(VHDL_SOURCES),)
    HDL_SOURCES += $(VHDL_SOURCES)
    GPI_EXTRA = cocotbvhpi
endif
    GPI_EXTRA = cocotbvhpi
    EXTRA_ARGS += -v93
    EXTRA_ARGS += -top $(TOPLEVEL)
    RTL_LIBRARY ?= $(TOPLEVEL)
    MAKE_LIB = -makelib $(RTL_LIBRARY)
    HDL_SOURCES = $(VHDL_SOURCES)
ifneq ($(VERILOG_SOURCES),)
    HDL_SOURCES += $(VERILOG_SOURCES)
endif
else
endif
```

I see two options:
- debug the conditional in the test-specific Makefiles to understand why -v93 is being passed to Icarus
- Rely on the IUS/Xcellium Makefiles to add -v93 to SIM_ARGS

ktbarrett:
Not specifying `SIM` used to default to `icarus`. Now it doesn't, and that conditional evaluates to true because `""` == `""`.

cmarqu:
> The core ius Makefile also adds this to SIM_ARGS:

> * Rely on the IUS/Xcellium Makefiles to add -v93 to SIM_ARGS

When I introduced the ``Makefile.xcelium``, I removed the ``-v93`` from the default ``SIM_ARGS`` because this would preclude having other conflicting options used in user Makefiles. ``Makefile,ius`` is gradually becoming obsolete because IUS/Incisive stopped being developed by Cadence in 2015, so I left the options untouched there so that there would not be a backward compatibility problem.
In short, adding ``-v93`` by default would be a step backward.

mballance:
> > The core ius Makefile also adds this to SIM_ARGS:
> 
> > * Rely on the IUS/Xcellium Makefiles to add -v93 to SIM_ARGS
> 
> When I introduced the `Makefile.xcelium`, I removed the `-v93` from the default `SIM_ARGS` because this would preclude having other conflicting options used in user Makefiles. `Makefile,ius` is gradually becoming obsolete because IUS/Incisive stopped being developed by Cadence in 2015, so I left the options untouched there so that there would not be a backward compatibility problem.
> In short, adding `-v93` by default would be a step backward.

Thanks for the input, @cmarqu. How about if change the conditional in the test Makefiles? We could have separate checks for ius and xcelium. In other words:

```
ifeq ($(SIM),"ius")
SIM_ARGS += -v93
endif
ifeq ($(SIM),"xcelium")
SIM_ARGS += -v93
endif
```
Not a creative solution, but should be functional in the presence of SIM not being set. Thoughts or other suggestions?

cmarqu:
> How about if change the conditional in the test Makefiles? We could have separate checks for ius and xcelium.

In that case, we could even leave ``ius`` out since the core Makefile already sets the ``-93``(we need to remember not to remove it), but maybe we could have an extra check for "is ``SIM`` empty or just whitespace" with 
```
ifeq ($(strip $(SIM)),)
# SIM not given
else ifeq ($(SIM),$(filter $(SIM),ius xcelium))
    SIM_ARGS += -v93
endif
```
(untested). Ah, I now read that the ``else ifeq`` is not supported in make 3.80, but maybe we don't need to care for this old version? This make stuff is really tiring...

mballance:
@cmarqu, good point on just adding a layer of nesting to avoid getting tripped up by an empty `$(SIM)`. I'll update the PR with this change

themperek:
Could you split this to 2 diffrent PRs. It looks not really related to me.

tests/designs/vhdl_configurations/Makefile
```diff
@@ -34,8 +34,10 @@ else
 endif
 VHDL_SOURCES += $(COCOTB)/tests/designs/vhdl_configurations/configurations.vhd
 
-ifeq ($(SIM),$(filter $(SIM),ius xcelium))
-    SIM_ARGS += -v93
+ifneq (,$(SIM))
```
eric-wieser: I don't get this - is `SIM` ever blank?

tests/designs/viterbi_decoder_axi4s/Makefile
```diff
@@ -65,8 +65,10 @@ VHDL_SOURCES = $(SRC_BASE)/packages/pkg_helper.vhd \
 	       $(SRC_BASE)/src/recursion.vhd \
 	       $(SRC_BASE)/src/dec_viterbi.vhd
 
-ifeq ($(SIM),$(filter $(SIM),ius xcelium))
-    SIM_ARGS += -v93
+ifneq (,$(SIM))
+    ifeq ($(SIM),$(filter $(SIM),ius xcelium))
+        SIM_ARGS += -v93
+    endif
 endif
```
eric-wieser: Simpler as:
```
ifneq ($(filter $(SIM),ius xcelium),)
    SIM_ARGS += -v93
endif
```

mballance:
> Could you split this to 2 different PRs? They look not really related to me.

I considered this fix related, since it was required in order for #1447 to pass tests. If splitting into two is the shortest path to getting both in, I can certainly do that. Please confirm.

tests/designs/vhdl_configurations/Makefile
mballance: According to @ktbarrett, SIM is now blank if not specified by the user. It used to be defaulted to "icarus". Now, blank is equivalent to "icarus".

tests/designs/viterbi_decoder_axi4s/Makefile
mballance: That's fair, @eric-wieser. Let's see where we're actually putting this fix, and I'll make the changes.

themperek:
Yes. I think it is better to split.

SIM is not empty https://github.com/cocotb/cocotb/blob/master/cocotb/share/makefiles/Makefile.inc#L68

tests/designs/viterbi_decoder_axi4s/Makefile
mballance: Focused this PR on Icarus argument passing

themperek:
@mballance Why is CI failing now?

mballance:
@themperek, it's because somehow the if conditionals are triggering and adding `-v93` to SIM_ARGS when running Icarus. Before #1447, this wouldn't trigger a failure because SIM_ARGS were to the right of the .vvp file. Icarus considers any arguments to the right of the .vvp file to be uninterpreted. Now that SIM_ARGS are on the left, Icarus is complaining that it doesn't accept `-v93`... Here's the output from Travis:

```
  TESTCASE= TOPLEVEL=tb_top TOPLEVEL_LANG=verilog COCOTB_SIM=1 \
        /usr/local/bin/vvp -M /Users/travis/build/cocotb/cocotb/.tox/py37/lib/python3.7/site-packages/cocotb/libs/icarus -m gpivpi -v93  sim_build/sim.vvp +foo=bar +test1 +test2 +options=fubar
/usr/local/bin/vvp: illegal option -- 9
/usr/local/bin/vvp: illegal option -- 3
make[4]: *** [results.xml] Error 2
make[3]: *** [sim] Error 2
make[3]: Target `all' not remade because of errors.
make[2]: *** [test_plusargs] Error 2
```

themperek:
OK. So let's try to fix one at a time. Sorry for all this. If you could make be PR to fix this first.

ktbarrett:
@themperek I've noticed this when I was working on my custom entry point PR that those default assignments in Makefile.inc aren't working. Not sure why. I added `GPI_EXTRA ?= cocotb` directly above `SIM ?= icarus` and it didn't take. The only reason Matt is seeing those issues is because it's not working. So that needs some thought, but that's a separate issue. I think the current makefiles are masking an actual issue.

themperek:
@ktbarrett No I think  I know what happened with #1306 I had to move `SIM ?= icarus` to `Makefile.inc`. In general, I would like to remove `Makefile.inc` and move everything to `Makefile.sim` but probably in a few steps. Maybe should make a PR and move https://github.com/cocotb/cocotb/blob/master/cocotb/share/makefiles/Makefile.inc#L67-L88 to `Makefile.sim` first.

themperek:
Maybe we can replace with this?
```Makefile
ifneq (,$(findstring $(SIM),ius xcelium))
    SIM_ARGS += -v93
endif
```
@cmarqu

eric-wieser:
That's what I suggest in https://github.com/cocotb/cocotb/pull/1447#discussion_r383220980, so I'm on board with that.

ktbarrett:
Something's up with the Travis reporter, the tests all pass.

Issue #1451, opened 2020-02-24
VHPI compilation fails with Conda GCC


mballance:
I'm seeing a compilation failure on the VHPI library when using Conda and top-of-trunk cocotb. This is due to Conda using a fairly recent version of GCC (7.3.0) that only defines the standard format macros (eg PRIu32) for C code (not C++) or if a special macro is defined.
Definition of the macros is protected as follows in inttypes.h:

```
/* "C++ implementations should define these macros only when
 *  __STDC_FORMAT_MACROS is defined before <inttypes.h> is included."
 */
#if (! defined(__cplusplus)) || defined(__STDC_FORMAT_MACROS)
```

I have a PR that defines _STDC_FORMAT_MACROS in VhpiImpl.h, and this works for me.

ktbarrett:
Copy of #1422.

eric-wieser:
I think this is fixed, please open a new issue if you disagree.

PR #1452, opened 2020-02-24
Conda fixup -- Ensure format macros (PRIu32) are defined

mballance:
This PR ensures the standard format macros (eg PRIu32) are defined when C++ code is compiled with a newer GCC (eg 7.3.0 from Conda).

This PR addresses #1451.


mballance:
Oops... Overlapping PRs...

PR #1453, opened 2020-02-24
Fixup compilation with Conda GCC

mballance:
This PR ensures the standard format macros (eg PRIu32) are defined when C++ code is compiled with a newer GCC (eg 7.3.0 from Conda).

This PR addresses #1451.


ktbarrett:
#1430 and #1436 address this issue as well. However, the chosen implementation is up in the air. This being the third and final possible solution :^)

imphil:
Let's take https://github.com/cocotb/cocotb/pull/1430 instead. Thanks for your patch @mballance!

PR #1455, opened 2020-02-24
Do not add -v93 to SIM_ARGS when SIM not set

themperek:
In many places `-v93` is add `SIM_ARGS` if `SIM` is not set. This causes errors if one corrects the order of  `SIM_ARGS` for Icarus #1447. 

This PR corrects this situation. Based on @eric-wieser suggestion https://github.com/cocotb/cocotb/pull/1447#discussion_r383220980

### DIFF
--- a/tests/test_cases/test_iteration_vhdl/Makefile
+++ b/tests/test_cases/test_iteration_vhdl/Makefile
@@ -28,10 +28,10 @@
 include ../../designs/viterbi_decoder_axi4s/Makefile
 MODULE = test_iteration
 
-ifeq ($(SIM),$(filter $(SIM), ius xcelium))
+ifneq ($(filter $(SIM),ius xcelium),)
     SIM_ARGS += -notimezeroasrtmsg
 endif
 
-ifeq ($(SIM),$(filter $(SIM), ghdl))
+ifeq ($(SIM),ghdl)
     SIM_ARGS += --ieee-asserts=disable-at-0
 endif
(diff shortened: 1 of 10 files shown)


tests/test_cases/test_iteration_vhdl/Makefile
```diff
@@ -28,10 +28,10 @@
 include ../../designs/viterbi_decoder_axi4s/Makefile
 MODULE = test_iteration
 
-ifeq ($(SIM),$(filter $(SIM), ius xcelium))
+ifneq ($(filter $(SIM),ius xcelium),)
     SIM_ARGS += -notimezeroasrtmsg
 endif
 
-ifeq ($(SIM),$(filter $(SIM), ghdl))
+ifeq ($(SIM),ghdl)
```
eric-wieser: For consistency can we use the same spelling as above here?

themperek: I see there are more places like this. Should I change everywhere?

eric-wieser: I suppose we already use this format elsewhere, so probably not important.

themperek: I the meantime I corrected every else (also for Questa). Is this OK for you?

eric-wieser: Either is fine

cmarqu:
Thanks for the PR, I haven't had much time recently.

PR #1464, opened 2020-02-29
Make heading markup match Python convention

cmarqu:
Reference: https://devguide.python.org/documenting/#sections

We don't use the ``#`` mark for "Part" yet so that we can use it later when we introduce the
"Tutorials", "Howtos", "Reference", "Explanation" parts as per https://www.divio.com/blog/documentation/.


cmarqu:
This doesn't have any user-visible change.

PR #1465, opened 2020-02-29
Fix Sphinx warnings

cmarqu:
* Always include sphinxcontrib.spelling
* Add files not directly included to exclude_patterns

### DIFF
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -95,7 +95,14 @@ autoclass_content = "both"
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = []
+exclude_patterns = [
+    # these are compiled into a single file at build-time,
+    # so there is no need to build them separately:
+    "newsfragments/*.rst",
+    # unused outputs from breathe:
+    "generated/namespacelist.rst",
+    "generated/namespace/*.rst",
+    ]
 
 # The reST default role (used for this markup: `text`) to use for all documents.
 #default_role = None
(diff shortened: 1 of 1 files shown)


documentation/source/conf.py
```diff
@@ -95,7 +96,11 @@
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = []
+exclude_patterns = [
+    "newsfragments/*.rst",
+    "generated/namespacelist.rst",
+    "generated/namespace/*.rst",
```
eric-wieser: If you know where these come from, can you add a comment?

documentation/source/conf.py
```diff
@@ -95,7 +96,11 @@
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = []
+exclude_patterns = [
+    "newsfragments/*.rst",
```
eric-wieser:
```suggestion
    # these are compiled into a single file at build-time, so there's no need to build them separately
    "newsfragments/*.rst",
```

documentation/source/conf.py
eric-wieser: I think this is correct?
```suggestion
    # unused outputs from breathe
    "generated/namespacelist.rst",
    "generated/namespace/*.rst",
```

documentation/source/conf.py
cmarqu: Added.

documentation/source/conf.py
cmarqu: It is, thanks.

PR #1468, opened 2020-03-01
Update Travis builds

ktbarrett:
We should have the following matrix of tests in CI:

Python version
 - 3.5
 - 3.6
 - 3.7
 - 3.8

OS
 - Ubuntu 16.04
 - Ubuntu 18.04
 - CentOS 6
 - CentOS 7
 - Windows 10 + mingw-w64
 - OSX?

Simulators
 - Icarus
 - GHDL

Additionally we should have targets for warnings compliance with additional warning flags enabled building for all supported compilers: GCC, Clang. And linting tests: Python, C++, whitespace, maybe more...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,26 +27,29 @@ jobs:
       script: flake8
 
     - stage: SimTests
-      os: linux
       python: 3.5
-      before_install: &docker_prep
-        - docker build -t cocotb .
-        - docker images -a
-        - docker run -d --name cocotb cocotb tail -f /dev/null
-        - docker ps -a
-      script:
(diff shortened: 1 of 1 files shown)


themperek:
I would keep testing a clean ubuntu environment or even add more (centos 6/7).

ktbarrett:
>add more (centos 6/7)

This seems like a good idea. What OSes are being used for cocotb? I use Debian Testing at home and most people are running RHEL 6 or 7 at work. There have been a few occasions where people were forced to work on Windows 10 and they use mingw64/msys2.

themperek:
The problems seem to be new virtualenv.
Solutions:
 - do not use tox on docker 
 - pip3 install tox==3.5.0 (or some newer that work)

PR #1469, opened 2020-03-01
Hack for issue with importlib-resources

ktbarrett:
There is an upstream issue with importlib-resources that is causing tox to fail to create and load an environment. A bug report has been filed upstream, but until it is fixed this should keep tests passing.


eric-wieser:
Can you reference the upstream issue in the commit message and perhaps also a comment?

eric-wieser:
Can you point me to an example of a failing cocotb test? Otherwise looks good.

ktbarrett:
@eric-wieser https://travis-ci.org/cocotb/cocotb/jobs/656966871#L920

eric-wieser:
Thanks!

PR #1474, opened 2020-03-03
Add newsfragment for gh-1343

cmarqu:


documentation/source/newsfragments/1343.bugfix.rst
```diff
@@ -0,0 +1 @@
+The Makefiles for the Aldec Riviera and Cadence IUS simulators have been fixed to use the correct name of the VHPI library (``libcocotbvhpi``). The bug prevented VHDL designs from being simulated.
```
eric-wieser:
```suggestion
The Makefiles for the Aldec Riviera and Cadence IUS simulators have been fixed to use the correct name of the VHPI library (``libcocotbvhpi``).
The bug prevented VHDL designs from being simulated, and was a regression in 1.3.0.
```

eric-wieser:
Does this need adjusting now that the change was part of the 1.3.1 release?

cmarqu:
I think we don't need this anymore at all actually?

PR #1475, opened 2020-03-03
Fix newsfragment categories (bug -> bugfix)

cmarqu:


cmarqu:
These newsfragments are now rendering: https://cocotb--1475.org.readthedocs.build/en/1475/release_notes.html#bugfixes

PR #1489, opened 2020-03-11
Clean Python 2 from tests

ktbarrett:
Merged test_cocotb_35.py into test_cocotb.py. Moved to Python 3 `math.isclose` in test_cocotb. Removed `ReturnValue` from all examples and tests.

xref #1289


tests/test_cases/test_external/test_external.py
```diff
@@ -265,7 +265,7 @@ def test_function_returns_exception(dut):
     @cocotb.function
     def func():
         # avoid using `return` syntax here since that requires Python >= 3.3
```
eric-wieser:
```suggestion
```

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1316,4 +1309,160 @@ def test_bad_attr(dut):
         assert False, "Expected AttributeError"
 
 
-from test_cocotb_35 import *
```
eric-wieser: I'm almost wondering if this is worth keeping separate, but renaming to `test_async_await` - this file is already pretty long.

ktbarrett: Splitting the file up is a good idea, but it should probably be done on a feature basis, not syntax. Also, we probably should be using `MODULE` instead of importing tests like this. But this is out of scope for this PR.

eric-wieser: Generally good points. The only danger of merging now is that splitting again later makes the history marginally hard to follow. Your commits are nicely separated anyway though, so it shouldn't be too hard to follow even if we do split that file up again later.

cocotb/result.py
```diff
@@ -86,7 +86,12 @@ def create_error(obj, msg):
 
 class ReturnValue(Exception):
     """Helper exception needed for Python versions prior to 3.3."""
+
     def __init__(self, retval):
+        warnings.warn(
+            "``ReturnValue`` is deprecated, use a normal return statement instead.",
+            DeprecationWarning,
+            stacklevel=2)
```
eric-wieser: Can you add a test somewhere of this warning?

cocotb/result.py
```diff
@@ -86,7 +86,12 @@ def create_error(obj, msg):
 
 class ReturnValue(Exception):
     """Helper exception needed for Python versions prior to 3.3."""
```
eric-wieser:
```suggestion
    """Helper exception needed for Python versions prior to 3.3.

    .. deprecated:: 1.4.0
        Use :keyword:`return` instead, which works in all supported versions of python
    """
```

cocotb/result.py
eric-wieser: Something like:
```python
@cocotb.test()
async def test_returnvalue_deprecated(dut):
	@cocotb.coroutine
	def get_value():
	    yield Timer(1, units='ns')
	    raise ReturnValue(42)

	with warnings.catch_warnings(record=True) as w:
	    # Cause all warnings to always be triggered.
	    val = await get_value()
	    assert val == 42
	    assert len(w) == 1
	    assert issubclass(w[-1].category, DeprecationWarning)
	    assert "deprecated" in str(w[-1].message)
```

ktbarrett: Added this to a new file specifically for testing deprecated features.

tests/designs/sample_module/Makefile
```diff
@@ -51,5 +51,5 @@ else
     $(error "A valid value (verilog or vhdl) was not provided for TOPLEVEL_LANG=$(TOPLEVEL_LANG)")
 endif
 
-include $(COCOTB)/makefiles/Makefile.inc
-include $(COCOTB)/makefiles/Makefile.sim
+include $(shell cocotb-config --makefiles)/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.sim
```
eric-wieser: This change seems sort of out of place, but I suppose if CI passes it's fine.

tests/test_cases/test_cocotb/test_deprecated.py
```diff
@@ -0,0 +1,19 @@
...
+            raise cocotb.result.ReturnValue(42)
+
+        with warnings.catch_warnings(record=True) as w:
+            # Cause all warnings to always be triggered.
+            warnings.simplefilter("always")
+            val = await get_value()
+            assert val == 42
+            assert len(w) == 1
+            assert issubclass(w[-1].category, DeprecationWarning)
+            assert "deprecated" in str(w[-1].message)
```
eric-wieser:
```suggestion
async def test_returnvalue_deprecated(dut):
    @cocotb.coroutine
    def get_value():
        yield cocotb.triggers.Timer(1, units='ns')
        raise cocotb.result.ReturnValue(42)

    with warnings.catch_warnings(record=True) as w:
        # Cause all warnings to always be triggered.
        warnings.simplefilter("always")
        val = await get_value()
        assert val == 42
        assert len(w) == 1
        assert issubclass(w[-1].category, DeprecationWarning)
        assert "deprecated" in str(w[-1].message)
```

eric-wieser:
```suggestion
            assert "use return instead" in str(w[-1].message)
```

tests/designs/sample_module/Makefile
ktbarrett: Yeah, that's an oopsie commit on me, will remove.

documentation/source/newsfragments/1489.removal.rst
```diff
@@ -0,0 +1 @@
+:class:`ReturnValue` is now deprecated. It will be removed in the release following 1.4. Use a :keyword:`return` statement instead; this works in all supported versions of Python.
```
eric-wieser:
```suggestion
:class:`ReturnValue` is now deprecated. Use a :keyword:`return` statement instead; this works in all supported versions of Python.
```

Let's not make promises about when things will be removed. The fact it is deprecated indicates that we will remove it without warning at some point.

tests/test_cases/test_cocotb/test_deprecated.py
```diff
@@ -0,0 +1,19 @@
...
+        raise cocotb.result.ReturnValue(42)
+
+    with warnings.catch_warnings(record=True) as w:
+        # Cause all warnings to always be triggered.
+        warnings.simplefilter("always")
+        val = await get_value()
+        assert val == 42
+        assert len(w) == 1
+        assert issubclass(w[-1].category, DeprecationWarning)
+        assert "use return instead" in str(w[-1].message)
```
eric-wieser: Unindenting these ensures you don't silence warnings you weren't expecting.
```suggestion
    assert val == 42
    assert len(w) == 1
    assert issubclass(w[-1].category, DeprecationWarning)
    assert "use return instead" in str(w[-1].message)
```

tests/test_cases/test_cocotb/test_deprecated.py
```diff
@@ -0,0 +1,19 @@
...
+        raise cocotb.result.ReturnValue(42)
+
+    with warnings.catch_warnings(record=True) as w:
+        # Cause all warnings to always be triggered.
+        warnings.simplefilter("always")
+        val = await get_value()
+    assert val == 42
+    assert len(w) == 1
+    assert issubclass(w[-1].category, DeprecationWarning)
+    assert "use return instead" in str(w[-1].message)
```
eric-wieser: Any reason for the `str`?
```suggestion
    assert "use return instead" in w[-1].message
```

ktbarrett: I was just following the [example](https://docs.python.org/3.8/library/warnings.html#testing-warnings). Will remove.

eric-wieser: Huh, interesting. Fine either way then, I guess.

ktbarrett: Turns out it was there for a very good reason, `.message` isn't a string...

eric-wieser: Whoops, `.message` is the `Warning` instance.

PR #1493, opened 2020-03-11
Simplify `cocotb.generators` to use builtins where possible

eric-wieser:
~Also fix `cocotb.generators.byte` to produce bytes and not unicode in Python 3.
Without this change, the generators in `cocotb.generators.packet` produces UTF-8 encoded latin1 (so never actually emits 0xFF, for instance!).~~

---

~~Alternatively...~~ In a future PR should we just remove this submodule?
* It's not documented
* It's somewhat duplicated by the standard library
* ~~`generator.byte` doesn't produce bytes at all on python 3~~
* `generators.packet` has dependencies on `scapy`.

### DIFF
--- a/cocotb/generators/__init__.py
+++ b/cocotb/generators/__init__.py
@@ -44,7 +44,10 @@ def repeat(obj, nrepeat=None):
     Kwargs:
         nrepeat (int): The number of times to repeatedly yield obj
     """
-    return itertools.repeat(obj, times=nrepeat)
+    if nrepeat is None:
+        return itertools.repeat(obj)
+    else:
+        return itertools.repeat(obj, times=nrepeat)
 
 
 @public
--- a/cocotb/generators/__init__.py
+++ b/cocotb/generators/__init__.py
@@ -30,6 +30,7 @@
 """
 import math
 import random
+import itertools
 from cocotb.decorators import public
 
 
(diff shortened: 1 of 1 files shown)


ktbarrett:
What was the intended use case?

eric-wieser:
Moved the interesting part of this to #1514, as these two pieces need to happen at the same time.

cocotb/generators/__init__.py
```diff
@@ -43,12 +44,7 @@ def repeat(obj, nrepeat=None):
     Kwargs:
         nrepeat (int): The number of times to repeatedly yield obj
     """
-    if nrepeat is None:
-        while True:
-            yield obj
-    else:
-        for i in range(nrepeat):
-            yield obj
+    return itertools.repeat(obj, times=nrepeat)
```
eric-wieser, line 47:
```suggestion
    if times is None:
        return itertools.repeat(obj)
    else:
        return itertools.repeat(obj, times=nrepeat)
```

ktbarrett:
> Alternatively... should we just remove this submodule?

This change is trivial enough now I say we bring it in, we can save that for 2.0.

eric-wieser:
I'll probably follow up with a PR to deprecate these two functions.

Issue #1495, opened 2020-03-11
Deprecate simulator-specific make variables


cmarqu:
The maintainers decided to deprecate some make variables that have not been implemented for all simulators. Reasons are the general maintenance burden, and that we often cannot test the effect of these variables in CI.
If any of these variables are used, we will print a warning with a reference to the documentation where we will explain how to achieve the same effect in a different way.

The initial list:
- [ ] Deprecate ``LICENSE_QUEUE``
- [ ] Deprecate ``COCOTB_NVC_TRACE``
- [ ] Deprecate ``VERILATOR_SIM_DEBUG``
- [ ] Deprecate ``VERILATOR_TRACE``

To be decided:
- [ ] Deprecate ``WAVES``
- [ ] Deprecate ``SCRIPT_FILE``

themperek:
Also all `V*_ARG` and `A*_LOG`?

alexforencich:
We should keep WAVES, but consider renaming it and implement it for all simulators.  Same with basic support for module parameters, perhaps at the top-level only.  Anything more simulator-specific than that can then be done manually, if necessary.

imphil:
Removing 1.9 milestone: We're done with changes for 1.9.0, anything else needs to go in a future release.

PR #1496, opened 2020-03-11
Deprecate COCOTB_NVC_TRACE, with doc reference

cmarqu:
Introduce a new ``Makefile.deprecations`` to collect make variable
deprecation warnings in a central place.

More deprecations to come via #1495, but this PR creates the "infrastructure" these will base on,  so it should go in first.

### DIFF
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -140,3 +140,10 @@ GHDL
 
 Support is preliminary.
 Noteworthy is that despite GHDL being a VHDL simulator, it implements the VPI interface.
+
+.. _sim-nvc:
+
+NVC
+===
+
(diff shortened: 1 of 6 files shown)


documentation/source/simulator_support.rst
```diff
@@ -140,3 +140,10 @@ GHDL
 
 Support is preliminary.
 Noteworthy is that despite GHDL being a VHDL simulator, it implements the VPI interface.
+
+.. _sim-nvc:
+
+NVC
+===
+
+To enable display of VHPI traces, add ``TRACE=--vhpi-trace`` to the make command.
```
themperek: Add to $SIM_ARGS  ?

cmarqu: Oh, right :)

themperek:
I would be for b) and an error/warning seems good enough but ...

Issue #719, opened 2018-12-12
Switch from property() to @property syntax


imphil:
Since Python 2.6 the @property/@x.setter syntax is supported, and gives a cleaner syntax than `property()`. Convert existing uses of `property()` to the newer syntax. 

This requires figuring out what to do with the existing setters and getter functions, which are often public (i.e. set_x() and get_x()). If existing code depends on these, we probably should keep compat wrappers and issue a deprecation warning.

eric-wieser:
> we probably should keep compat wrappers

An easy trick for this is
```python
@property
def my_prop(self):
    ...

@my_prop.setter
def my_prop(self, value):
    ...

# aliases for compatibility
set_my_prop = my_prop.fset
get_my_prop = my_prop.fset
```

eric-wieser:
Ah, not addressed by gh-1497: Deprecating these. I'll open a new issue.

PR #1497, opened 2020-03-13
Use new style property decorators.

andrewandrepowell:
Hello,

This change is to resolve issue #719.

### DIFF
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -281,13 +281,6 @@ class BinaryValue(object):
             rv = x
         return rv
 
-    def get_value(self):
-        """Return the integer representation of the underlying vector."""
-        return self._convert_from[self.binaryRepresentation](self._str)
-
-    def set_value(self, integer):
-        self._str = self._convert_to[self.binaryRepresentation](integer)
-
     @property
     def is_resolvable(self):
         """Does the value contain any ``X``'s?  Inquiring minds want to know."""
@@ -297,20 +290,23 @@ class BinaryValue(object):
     def value(self):
         "Integer access to the value"
         warnings.warn("deprecated", DeprecationWarning)
-        return self.get_value()
+        return self.integer
 
     @value.setter
     def value(self, val):
-        self.set_value(val)
+        self.integer = val
 
     @property
     def integer(self):
         "The integer representation of the underlying vector."
-        return self.get_value()
+        return self._convert_from[self.binaryRepresentation](self._str)
 
     @integer.setter
     def integer(self, val):
-        self.set_value(val)
+        self._str = self._convert_to[self.binaryRepresentation](val)
+
+    get_value = integer.fget
+    set_value = integer.fset
 
     @property
     def signed_integer(self):
@@ -327,9 +323,30 @@ class BinaryValue(object):
 
     @signed_integer.setter
     def signed_integer(self, val):
-        self.set_value(val)
+        self.integer = val
 
-    def get_buff(self):
+    def get_hex_buff(self):
+        bstr = self.buff
+        hstr = '%0*X' % ((len(bstr) + 3) // 4, int(bstr, 2))
+        return hstr
+
+    def _adjust(self):
+        """Pad/truncate the bit string to the correct length."""
+        if self._n_bits is None:
+            return
+        l = len(self._str)
+        if l < self._n_bits:
+            if self.big_endian:
+                self._str = self._str + "0" * (self._n_bits - l)
+            else:
+                self._str = "0" * (self._n_bits - l) + self._str
+        elif l > self._n_bits:
+            print("WARNING: truncating value to match requested number of bits "
+                  "(%d -> %d)" % (l, self._n_bits))
+            self._str = self._str[l - self._n_bits:]
+
+    @property
+    def buff(self):
         """Attribute :attr:`buff` represents the value as a binary string buffer.
 
         >>> "0100000100101111".buff == "\x41\x2F"
@@ -351,51 +368,26 @@ class BinaryValue(object):
                 buff = chr(val) + buff
         return buff
 
-    def get_hex_buff(self):
-        bstr = self.get_buff()
-        hstr = '%0*X' % ((len(bstr) + 3) // 4, int(bstr, 2))
-        return hstr
-
-    def set_buff(self, buff):
+    @buff.setter
+    def buff(self, val):
         self._str = ""
-        for char in buff:
+        for char in val:
             if self.big_endian:
                 self._str += "{0:08b}".format(ord(char))
             else:
                 self._str = "{0:08b}".format(ord(char)) + self._str
         self._adjust()
 
-    def _adjust(self):
-        """Pad/truncate the bit string to the correct length."""
-        if self._n_bits is None:
-            return
-        l = len(self._str)
-        if l < self._n_bits:
-            if self.big_endian:
-                self._str = self._str + "0" * (self._n_bits - l)
-            else:
-                self._str = "0" * (self._n_bits - l) + self._str
-        elif l > self._n_bits:
-            print("WARNING: truncating value to match requested number of bits "
-                  "(%d -> %d)" % (l, self._n_bits))
-            self._str = self._str[l - self._n_bits:]
+    get_buff = buff.fget
+    set_buff = buff.fset
 
     @property
-    def buff(self):
-        "Access to the value as a buffer."
-        return self.get_buff()
-
-    @buff.setter
-    def buff(self, val):
-        "Access to the value as a buffer."
-        self.set_buff(val)
-
-    def get_binstr(self):
-        """Attribute :attr:`binstr` is the binary representation stored as
-        a string of ``1`` and ``0``."""
+    def binstr(self):
+        "Access to the binary string."
         return self._str
 
-    def set_binstr(self, string):
+    @binstr.setter
+    def binstr(self, string):
         for char in string:
             if char not in BinaryValue._permitted_chars:
                 raise ValueError("Attempting to assign character %s to a %s" %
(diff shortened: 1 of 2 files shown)


cocotb/binary.py
```diff
@@ -376,16 +406,25 @@ def set_binstr(self, string):
...
         """The number of bits of the binary value."""
         return self._n_bits
-
-    n_bits = property(_get_n_bits, None, None,
-                      "Access to the number of bits of the binary value.")
+        
+    @property
+    def n_bits(self):
+        "Access to the number of bits of the binary value."
+        return self._get_n_bits()
```
eric-wieser: Can you just inline this method, since it's private?

eric-wieser:
This is changing <=2.5 style to >=2.6 style, not 2.7 to 3.x!

cocotb/binary.py
andrewandrepowell: Done

andrewandrepowell:
Made the requested changes.

cocotb/binary.py
```diff
@@ -303,12 +303,35 @@ def is_resolvable(self):
...
 
-    value = property(get_value, set_value, None,
-                     "Integer access to the value. **deprecated**")
-    integer = property(get_value, set_value, None,
-                       "The integer representation of the underlying vector.")
-    signed_integer = property(get_value_signed, set_value, None,
-                              "The signed integer representation of the underlying vector.")
+    @property
+    def value(self):
+        "Integer access to the value. **deprecated**"
```
ktbarrett: I think we should take this opportunity to deprecate this interface by adding a warning and removing the message from the docstring.

cocotb/binary.py
```diff
@@ -303,12 +303,35 @@ def is_resolvable(self):
...
+        self.set_value(val)
+
+    @property
+    def integer(self):
+        "The integer representation of the underlying vector."
+        return self.get_value()
+
+    @integer.setter
+    def integer(self, val):
+        "The integer representation of the underlying vector."
```
ktbarrett: You can leave off the docstrings on all of the setters, only the getter's docstring is used.

cocotb/binary.py
```diff
@@ -303,12 +303,35 @@ def is_resolvable(self):
...
+        self.set_value(val)
+
+    @property
+    def signed_integer(self):
+        "The signed integer representation of the underlying vector."
+        return self.get_value_signed()
```
eric-wieser: I think we should flip the definitions here, from
```
    @property
    def signed_integer(self):
        "The signed integer representation of the underlying vector."
        return self.get_value_signed()
    def get_value_signed(self):
         ...
```
to
```
    @property
    def signed_integer(self):
        "The signed integer representation of the underlying vector."
        ...

    get_value_signed = signed_integer.fget
```
or similar

andrewandrepowell: Completed this specific request. The other cases are the `set_value` and `get_value` methods. Should the same change be applied to those as well, and I'm assuming their definitions will be used for the integer property.

eric-wieser: Yes please!

ktbarrett:
The warning is throwing, but only on Python 3.8. I figured somewhere the filter was changed to `"error"`, but I don't see that. Sounds like you are for some digging.

eric-wieser:
Yep, only 3.8 let's you configure warnings as errors via an environment var, so 3.8 is the only CI we have that turned on for.

One of our tests or examples is using this deprecated property.

Let's not worry about the deprecation in this PR - mixing deprecations and cleanups is usually a bad idea.

cocotb/binary.py
```diff
@@ -295,22 +319,34 @@ def get_value_signed(self):
...
-    integer = property(get_value, set_value, None,
-                       "The integer representation of the underlying vector.")
-    signed_integer = property(get_value_signed, set_value, None,
-                              "The signed integer representation of the underlying vector.")
+    def get_hex_buff(self):
+        bstr = self.buff
+        hstr = '%0*X' % ((len(bstr) + 3) // 4, int(bstr, 2))
+        return hstr
+
+    def _adjust(self):
```
eric-wieser: Can you try to keep the code in the same order as before to keep the diff small? This function for instance did not change, but you've moved so much code from above it to below it that git thinks it did.

cocotb/binary.py
```diff
@@ -295,22 +319,34 @@ def get_value_signed(self):
         else:
             return -1 * (1 + (int(~ival) & (signbit - 1)))
 
-    def set_value(self, integer):
-        self._str = self._convert_to[self.binaryRepresentation](integer)
+    get_value_signed = signed_integer.fget
 
-    @property
-    def is_resolvable(self):
```
eric-wieser: Similarly, it would be good to put this function back where it was.

cocotb/binary.py
```diff
@@ -332,14 +352,10 @@ def get_buff(self):
                 buff = chr(val) + buff
         return buff
 
-    def get_hex_buff(self):
-        bstr = self.get_buff()
-        hstr = '%0*X' % ((len(bstr) + 3) // 4, int(bstr, 2))
-        return hstr
-
```
eric-wieser: Can you move this function back too?

cocotb/binary.py
eric-wieser: Tracked by #1503, marking as resolved here.

cocotb/binary.py
eric-wieser: (the change from `self.get_buff()` to `self.buff` is still a good one)

cocotb/binary.py
```diff
@@ -361,35 +377,34 @@ def _adjust(self):
...
-                    "Access to the value as a buffer.")
+    get_buff = buff.fget
+    set_buff = buff.fset
 
-    def get_binstr(self):
-        """Attribute :attr:`binstr` is the binary representation stored as
-        a string of ``1`` and ``0``."""
+    @property
+    def binstr(self):
+        "Access to the binary string."
```
eric-wieser:
```suggestion
        """ The binary representation stored as a string of ``0``, ``1``, and possibly ``x``, ``z``, and other states. """
```

cocotb/binary.py
```diff
@@ -281,12 +281,30 @@ def _adjust_twos_comp(self, x):
...
+    @integer.setter
+    def integer(self, val):
+        self._str = self._convert_to[self.binaryRepresentation](val)
+
+    get_value = integer.fget
+    set_value = integer.fset
+
+    @property
+    def signed_integer(self):
+        "The signed integer representation of the underlying vector."
```
eric-wieser:
```suggestion
        """The signed integer representation of the underlying vector."""
```
While what you have works, it's not conventional to write docstrings with a single set of quotes.

cocotb/binary.py
```diff
@@ -281,12 +281,30 @@ def _adjust_twos_comp(self, x):
...
         return self._convert_from[self.binaryRepresentation](self._str)
 
-    def get_value_signed(self):
-        """Return the signed integer representation of the underlying vector."""
+    @integer.setter
+    def integer(self, val):
+        self._str = self._convert_to[self.binaryRepresentation](val)
+
+    get_value = integer.fget
+    set_value = integer.fset
```
eric-wieser: Would like it to be a little clearer that this is tied to the `value` property, and will be deprecated alongside it if we ever do that.
```suggestion

    get_value = value.fget
    set_value = value.fset
```

cocotb/binary.py
andrewandrepowell: Oops. Missed that one!

cocotb/binary.py
```diff
@@ -295,22 +313,19 @@ def get_value_signed(self):
...
 
-    value = property(get_value, set_value, None,
-                     "Integer access to the value. **deprecated**")
-    integer = property(get_value, set_value, None,
-                       "The integer representation of the underlying vector.")
-    signed_integer = property(get_value_signed, set_value, None,
-                              "The signed integer representation of the underlying vector.")
+    @signed_integer.setter
+    def signed_integer(self, val):
+        self.integer = val
```
eric-wieser: Might make sense to put this on line 315 so that the getter and setter are adjacent.

cocotb/binary.py
```diff
@@ -281,12 +281,30 @@ def _adjust_twos_comp(self, x):
...
-        """Return the integer representation of the underlying vector."""
+    @property
+    def value(self):
+        """Integer access to the value. **deprecated**"""
+        return self.integer
+
+    @value.setter
+    def value(self, val):
+        self.integer = val
+
```
eric-wieser: Looking at the docs, this makes this appear before `integer`: https://cocotb--1497.org.readthedocs.build/en/1497/library_reference.html?highlight=binaryvalue#cocotb.binary.BinaryValue.value

Maybe move these lines down to line 301?

cocotb/binary.py
```diff
@@ -281,12 +281,30 @@ def _adjust_twos_comp(self, x):
...
         return self._convert_from[self.binaryRepresentation](self._str)
 
-    def get_value_signed(self):
-        """Return the signed integer representation of the underlying vector."""
+    @integer.setter
+    def integer(self, val):
+        self._str = self._convert_to[self.binaryRepresentation](val)
+
+    get_value = value.fget
+    set_value = value.fset
```
ktbarrett: The nittiest of nitpicks, this should probably be moved up to directly below the `value` property definitions.

eric-wieser: Better to do the reverse and move the property down: https://github.com/cocotb/cocotb/pull/1497#discussion_r395069321

eric-wieser:
If you're feeling generous, can you add `:exclude-members: get_value, get_buff, get_binstr, get_value_signed` to `BinaryValue` [here](https://github.com/cocotb/cocotb/blob/master/documentation/source/library_reference.rst#interacting-with-the-simulator)?

eric-wieser:
Went ahead and made those tweaks myself, will merge once CI passes. Thanks again @andrewandrepowell!

eric-wieser:
Docs look good now: https://cocotb--1497.org.readthedocs.build/en/1497/library_reference.html#cocotb.binary.BinaryValue

PR #1502, opened 2020-03-17
Add matrix_multiplier to examples.

garmin-mjames:
Add matrix multiplication example and test. The test is heavily
weighted towards use of the BinaryValue class. By fixing the random
seed by default, it also provides a deterministic test for performance
regression.

This example and test are designed to be a testbench for various
performance enhancements, especially involving BinaryValue.

Split from gh-1286

### DIFF
--- a/documentation/source/examples.rst
+++ b/documentation/source/examples.rst
@@ -35,6 +35,47 @@ to start and stop generation of input data.
 A :class:`.TestFactory` is used to generate the random tests.
 
 
+.. _matrix_multiplier:
+
+Matrix Multiplier
+=================
+
+The directory :file:`cocotb/examples/matrix_multiplier`
+contains a module for multiplying two matrices together,
+implemented in both **VHDL** and **SystemVerilog**.
+
+The module takes two matrices ``a_i`` and ``b_i`` as inputs
+and provides the resulting matrix ``c_o`` as an output.
+On each rising clock edge,
+``c_o`` is calculated and output.
+When input ``valid_i`` is high
+and ``c_o`` is calculated,
+``valid_o`` goes high to signal a valid output value.
+
+The testbench defines ``MatrixInMonitor`` and ``MatrixOutMonitor``
+(both sub-classes of :class:`.BusMonitor`)
+and a test case ``test_multiply``.
+
+``MatrixInMonitor`` watches for valid input matrices,
+then does the multiplication in Python
+and stores the result as the expected output matrix.
+
+``MatrixOutMonitor`` watches for valid output matrices
+and compares the result to the expected value.
+
+The testbench makes use of :class:`.TestFactory`
+and random data generators
+to test many sets of matrices,
+and :class:`.Scoreboard` to compare expected and actual results.
+
+The number of data bits for each entry in the matrices,
+as well as the row and column counts for each matrix,
+are configurable in the Makefile.
+
+.. note::
+    The example module uses one-dimensional arrays in the port definition to represent the matrices.
+
+
 Mean
 ====
 
--- a/documentation/source/newsfragments/1502.doc.rst
+++ b/documentation/source/newsfragments/1502.doc.rst
@@ -1,0 +1,1 @@
+New example :ref:`matrix_multiplier`.
--- a/examples/matrix_multiplier/hdl/matrix_multiplier.sv
+++ b/examples/matrix_multiplier/hdl/matrix_multiplier.sv
@@ -1,0 +1,78 @@
+// This file is public domain, it can be freely copied without restrictions.
+// SPDX-License-Identifier: CC0-1.0
+
+// Matrix Multiplier DUT
+`timescale 1ns/1ps
+
+module matrix_multiplier #(
+  parameter int DATA_WIDTH = 8,
+  parameter int A_ROWS = 8,
+  parameter int B_COLUMNS = 5,
+  parameter int A_COLUMNS_B_ROWS = 4,
+  parameter int C_DATA_WIDTH = (2 * DATA_WIDTH) + $clog2(A_COLUMNS_B_ROWS)
+) (
+  input                           clk_i,
+  input                           reset_i,
+  input                           valid_i,
+  output logic                    valid_o,
+  input        [DATA_WIDTH-1:0]   a_i[A_ROWS * A_COLUMNS_B_ROWS],
+  input        [DATA_WIDTH-1:0]   b_i[A_COLUMNS_B_ROWS * B_COLUMNS],
+  output logic [C_DATA_WIDTH-1:0] c_o[A_ROWS * B_COLUMNS]
+);
+
+  logic [C_DATA_WIDTH-1:0] c_calc[A_ROWS * B_COLUMNS];
+
+  always @(*) begin
+    logic [C_DATA_WIDTH-1:0] c_element;
+    for (int i = 0; i < A_ROWS; i = i + 1) begin
+      for (int j = 0; j < B_COLUMNS; j = j + 1) begin
+        c_element = 0;
+        for (int k = 0; k < A_COLUMNS_B_ROWS; k = k + 1) begin
+          c_element = c_element + (a_i[(i * A_COLUMNS_B_ROWS) + k] * b_i[(k * B_COLUMNS) + j]);
+        end
+        c_calc[(i * B_COLUMNS) + j] = c_element;
+      end
+    end
+  end
+
+  always @(posedge clk_i) begin
+    if (reset_i) begin
+      valid_o <= 1'b0;
+
+      for (int idx = 0; idx < (A_ROWS * B_COLUMNS); idx++) begin
+        c_o[idx] <= '0;
+      end
+    end else begin
+      valid_o <= valid_i;
+
+      for (int idx = 0; idx < (A_ROWS * B_COLUMNS); idx++) begin
+        c_o[idx] <= c_calc[idx];
+      end
+    end
+  end
+
+  // Dump waves
+`ifndef VERILATOR
+  initial begin
+    integer idx;
+    $dumpfile("dump.vcd");
+    $dumpvars(1, matrix_multiplier);
+  `ifdef __ICARUS__
+      for (idx = 0; idx < (A_ROWS * A_COLUMNS_B_ROWS); idx++) begin
+        $dumpvars(1, a_i[idx]);
+      end
+      for (idx = 0; idx < (A_COLUMNS_B_ROWS * B_COLUMNS); idx++) begin
+        $dumpvars(1, b_i[idx]);
+      end
+      for (idx = 0; idx < (A_ROWS * B_COLUMNS); idx++) begin
+        $dumpvars(1, c_o[idx]);
+      end
+  `else
+    $dumpvars(1, a_i);
+    $dumpvars(1, b_i);
+    $dumpvars(1, c_o);
+  `endif
+  end
+`endif
+
+endmodule
--- a/examples/matrix_multiplier/hdl/matrix_multiplier.vhd
+++ b/examples/matrix_multiplier/hdl/matrix_multiplier.vhd
@@ -1,0 +1,75 @@
+-- This file is public domain, it can be freely copied without restrictions.
+-- SPDX-License-Identifier: CC0-1.0
+
+-- Matrix Multiplier DUT
+library ieee ;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+use work.matrix_multiplier_pkg.all;
+use work.mean_pkg.clog2;
+
+entity matrix_multiplier is
+  generic (
+    DATA_WIDTH : positive := 8;
+    A_ROWS : positive := 8;
+    B_COLUMNS : positive := 5;
+    A_COLUMNS_B_ROWS : positive := 4
+  );
+  port (
+    clk_i   : in    std_logic;
+    reset_i : in    std_logic;
+    valid_i : in    std_logic;
+    valid_o : out   std_logic;
+    a_i     : in    flat_matrix_type(0 to (A_ROWS * A_COLUMNS_B_ROWS) - 1)(DATA_WIDTH - 1 downto 0);
+    b_i     : in    flat_matrix_type(0 to (A_COLUMNS_B_ROWS * B_COLUMNS) - 1)(DATA_WIDTH - 1 downto 0);
+    c_o     : out   flat_matrix_type(0 to (A_ROWS * B_COLUMNS) - 1)((2 * DATA_WIDTH) + clog2(A_COLUMNS_B_ROWS) - 1 downto 0)
+  );
+end entity matrix_multiplier;
+
+architecture rtl of matrix_multiplier is
+
+  signal c_calc : flat_matrix_type(c_o'RANGE)(c_o(0)'RANGE);
+
+begin
+
+  multiply : process (all) is
+
+    variable c_var : flat_matrix_type(c_o'RANGE)(c_o(0)'RANGE);
+
+  begin
+
+    c_var := (others => (others => '0'));
+
+    C_ROWS : for i in 0 to A_ROWS-1 loop
+      C_COLUMNS : for j in 0 to B_COLUMNS-1 loop
+        DOT_PRODUCT : for k in 0 to A_COLUMNS_B_ROWS-1 loop
+          c_var((i * B_COLUMNS) + j) := c_var((i * B_COLUMNS) + j) + (a_i((i * A_COLUMNS_B_ROWS) + k) * b_i((k * B_COLUMNS) + j));
+        end loop;
+      end loop;
+    end loop;
+
+    c_calc <= c_var;
+
+  end process multiply;
+
+  proc_reg : process (clk_i) is
+  begin
+
+    if (rising_edge(clk_i)) then
+      if (reset_i) then
+        valid_o <= '0';
+        c_o <= (others => (others => '0'));
+      else
+        valid_o <= valid_i;
+
+        if (valid_i) then
+          c_o <= c_calc;
+        else
+          c_o <= (others => (others => 'X'));
+        end if;
+      end if;
+    end if;
+
+  end process proc_reg;
+
+end architecture rtl;
--- a/examples/matrix_multiplier/tests/Makefile
+++ b/examples/matrix_multiplier/tests/Makefile
@@ -1,0 +1,82 @@
+# This file is public domain, it can be freely copied without restrictions.
+# SPDX-License-Identifier: CC0-1.0
+
+TOPLEVEL_LANG ?= verilog
+SIM ?= icarus
+
+PWD=$(shell pwd)
+
+# Matrix parameters
+DATA_WIDTH ?= 32
+A_ROWS ?= 10
+B_COLUMNS ?= 4
+A_COLUMNS_B_ROWS ?= 6
+
+ifeq ($(TOPLEVEL_LANG),verilog)
+    VERILOG_SOURCES = $(PWD)/../hdl/matrix_multiplier.sv
+
+    # Set module parameters
+    ifeq ($(SIM),icarus)
+        COMPILE_ARGS += -Pmatrix_multiplier.DATA_WIDTH=$(DATA_WIDTH) -Pmatrix_multiplier.A_ROWS=$(A_ROWS) -Pmatrix_multiplier.B_COLUMNS=$(B_COLUMNS) -Pmatrix_multiplier.A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifneq ($(filter $(SIM),questa modelsim riviera activehdl),)
+        SIM_ARGS += -gDATA_WIDTH=$(DATA_WIDTH) -gA_ROWS=$(A_ROWS) -gB_COLUMNS=$(B_COLUMNS) -gA_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifeq ($(SIM),vcs)
+        COMPILE_ARGS += -pvalue+matrix_multiplier/DATA_WIDTH=$(DATA_WIDTH) -pvalue+matrix_multiplier/A_ROWS=$(A_ROWS) -pvalue+matrix_multiplier/B_COLUMNS=$(B_COLUMNS) -pvalue+matrix_multiplier/A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifeq ($(SIM),verilator)
+        COMPILE_ARGS += -GDATA_WIDTH=$(DATA_WIDTH) -GA_ROWS=$(A_ROWS) -GB_COLUMNS=$(B_COLUMNS) -GA_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifneq ($(filter $(SIM),ius xcelium),)
+        EXTRA_ARGS += -defparam "matrix_multiplier.DATA_WIDTH=$(DATA_WIDTH)" -defparam "matrix_multiplier.A_ROWS=$(A_ROWS)" -defparam "matrix_multiplier.B_COLUMNS=$(B_COLUMNS)" -defparam "matrix_multiplier.A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)"
+    endif
+
+    ifneq ($(filter $(SIM),riviera activehdl),)
+        COMPILE_ARGS += -sv2k12
+    endif
+
+else ifeq ($(TOPLEVEL_LANG),vhdl)
+    VHDL_SOURCES = $(PWD)/../../mean/hdl/mean_pkg.vhd $(PWD)/../hdl/matrix_multiplier_pkg.vhd $(PWD)/../hdl/matrix_multiplier.vhd
+
+    ifneq ($(filter $(SIM),ghdl questa modelsim riviera activehdl),)
+        # ghdl, questa, and aldec all use SIM_ARGS with '-g' for setting generics
+        SIM_ARGS += -gDATA_WIDTH=$(DATA_WIDTH) -gA_ROWS=$(A_ROWS) -gB_COLUMNS=$(B_COLUMNS) -gA_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifneq ($(filter $(SIM),ius xcelium),)
+        SIM_ARGS += -generic "matrix_multiplier:DATA_WIDTH=>$(DATA_WIDTH)" -generic "matrix_multiplier:A_ROWS=>$(A_ROWS)" -generic "matrix_multiplier:B_COLUMNS=>$(B_COLUMNS)" -generic "matrix_multiplier:A_COLUMNS_B_ROWS=>$(A_COLUMNS_B_ROWS)"
+    endif
+
+    ifeq ($(SIM),ghdl)
+        EXTRA_ARGS += --std=08
+        SIM_ARGS += --wave=wave.ghw
+    else ifneq ($(filter $(SIM),questa modelsim riviera activehdl),)
+        COMPILE_ARGS += -2008
+    endif
+else
+    $(error "A valid value (verilog or vhdl) was not provided for TOPLEVEL_LANG=$(TOPLEVEL_LANG)")
+endif
+
+# Fix the seed to ensure deterministic tests
+export RANDOM_SEED := 123456789
+
+# Export generics to test
+# Can't use plusargs since they aren't supported by ghdl
+export DATA_WIDTH
+export A_ROWS
+export B_COLUMNS
+export A_COLUMNS_B_ROWS
+
+TOPLEVEL    := matrix_multiplier
+MODULE      := test_matrix_multiplier
+
+include $(shell cocotb-config --makefiles)/Makefile.sim
+
+
+# Profiling
+
+DOT_BINARY ?= dot
+
+test_profile.pstat: sim
+
+callgraph.svg: test_profile.pstat
+	$(shell cocotb-config --python-bin) -m gprof2dot -f pstats ./$< | $(DOT_BINARY) -Tsvg -o $@
+
+.PHONY: profile
+profile:
+	COCOTB_ENABLE_PROFILING=1 $(MAKE) callgraph.svg
--- a/examples/matrix_multiplier/tests/test_matrix_multiplier.py
+++ b/examples/matrix_multiplier/tests/test_matrix_multiplier.py
@@ -1,0 +1,148 @@
+# This file is public domain, it can be freely copied without restrictions.
+# SPDX-License-Identifier: CC0-1.0
+
+import math
+import os
+from random import getrandbits
+
+import cocotb
+from cocotb.binary import BinaryValue
+from cocotb.clock import Clock
+from cocotb.monitors import BusMonitor
+from cocotb.regression import TestFactory
+from cocotb.scoreboard import Scoreboard
+from cocotb.triggers import RisingEdge, ReadOnly
+
+NUM_SAMPLES = int(os.environ.get('NUM_SAMPLES', 3000))
+DATA_WIDTH = int(os.environ['DATA_WIDTH'])
+A_ROWS = int(os.environ['A_ROWS'])
+B_COLUMNS = int(os.environ['B_COLUMNS'])
+A_COLUMNS_B_ROWS = int(os.environ['A_COLUMNS_B_ROWS'])
+
+
+class MatrixMonitor(BusMonitor):
+    """Base class for monitoring inputs/outputs of Matrix Multiplier."""
+    def __init__(self, dut, callback=None, event=None):
+        super().__init__(dut, "", dut.clk_i, callback=callback, event=event)
+
+
+class MatrixInMonitor(MatrixMonitor):
+    """Monitor inputs to Matrix Multiplier module.
+
+    Generate expected results for each multiplication operation.
+    """
+    _signals = {"A": "a_i", "B": "b_i", "valid": "valid_i"}
+
+    async def _monitor_recv(self):
+        while True:
+            await RisingEdge(self.clock)
+            await ReadOnly()
+
+            if self.bus.valid.value:
+                a_matrix = self.bus.A.value
+                b_matrix = self.bus.B.value
+
+                # Calculate the expected result of C
+                c_expected = [
+                    BinaryValue(
+                        sum(
+                            [
+                                a_matrix[(i * A_COLUMNS_B_ROWS) + n] * b_matrix[(n * B_COLUMNS) + j]
+                                for n in range(A_COLUMNS_B_ROWS)
+                            ]
+                        ),
+                        n_bits=(DATA_WIDTH * 2) + math.ceil(math.log2(A_COLUMNS_B_ROWS)),
+                        bigEndian=False
+                    )
+                    for i in range(A_ROWS)
+                    for j in range(B_COLUMNS)
+                ]
+
+                self._recv(c_expected)
+
+
+class MatrixOutMonitor(MatrixMonitor):
+    """Monitor outputs from Matrix Multiplier module.
+
+    Capture result matrix for each multiplication operation.
+    """
+    _signals = {"C": "c_o", "valid": "valid_o"}
+
+    async def _monitor_recv(self):
+        while True:
+            await RisingEdge(self.clock)
+            await ReadOnly()
+
+            if self.bus.valid.value:
+                c_actual = self.bus.C.value
+                self._recv(c_actual)
+
+
+async def test_multiply(dut, a_data, b_data):
+    """Test multiplication of many matrices."""
+
+    cocotb.fork(Clock(dut.clk_i, 10, units='ns').start())
+
+    # Configure Scoreboard to compare module results to expected
+    expected_output = []
+
+    in_monitor = MatrixInMonitor(dut, callback=expected_output.append)
+    out_monitor = MatrixOutMonitor(dut)
+
+    scoreboard = Scoreboard(dut)
+    scoreboard.add_interface(out_monitor, expected_output)
+
+    # Initial values
+    dut.valid_i <= 0
+    dut.a_i <= create_a(lambda x: 0)
+    dut.b_i <= create_b(lambda x: 0)
+
+    # Reset DUT
+    dut.reset_i <= 1
+    for _ in range(3):
+        await RisingEdge(dut.clk_i)
+    dut.reset_i <= 0
+
+    # Do multiplication operations
+    for A, B in zip(a_data(), b_data()):
+        await RisingEdge(dut.clk_i)
+        dut.a_i <= A
+        dut.b_i <= B
+        dut.valid_i <= 1
+
+        await RisingEdge(dut.clk_i)
+        dut.valid_i <= 0
+
+    await RisingEdge(dut.clk_i)
+
+    raise scoreboard.result
+
+
+def create_matrix(func, rows, cols):
+    return [func(DATA_WIDTH) for row in range(rows) for col in range(cols)]
+
+
+def create_a(func):
+    return create_matrix(func, A_ROWS, A_COLUMNS_B_ROWS)
+
+
+def create_b(func):
+    return create_matrix(func, A_COLUMNS_B_ROWS, B_COLUMNS)
+
+
+def gen_a(num_samples=NUM_SAMPLES, func=getrandbits):
+    """Generate random matrix data for A"""
+    for _ in range(num_samples):
+        yield create_a(func)
+
+
+def gen_b(num_samples=NUM_SAMPLES, func=getrandbits):
+    """Generate random matrix data for B"""
+    for _ in range(num_samples):
+        yield create_b(func)
+
+
+factory = TestFactory(test_multiply)
+factory.add_option('a_data', [gen_a])
+factory.add_option('b_data', [gen_b])
+factory.generate_tests()
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -32,6 +32,7 @@ EXAMPLES := adder/tests \
             dff/tests \
             simple_dff \
             endian_swapper/tests \
+            matrix_multiplier/tests \
             mean/tests \
             mixed_language/tests
 # requires root permissions: ping_tun_tap/tests
--- a/examples/matrix_multiplier/hdl/matrix_multiplier_pkg.vhd
+++ b/examples/matrix_multiplier/hdl/matrix_multiplier_pkg.vhd
@@ -1,0 +1,13 @@
+-- This file is public domain, it can be freely copied without restrictions.
+-- SPDX-License-Identifier: CC0-1.0
+
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+
+package matrix_multiplier_pkg is
+
+    -- VHDL-2008 required for unconstrained array types
+    type flat_matrix_type is array (integer range <>) of unsigned;
+
+end package;


documentation/source/examples.rst
```diff
@@ -35,6 +35,23 @@ to start and stop generation of input data.
...
+=================
+
+The directory :file:`cocotb/examples/matrix_multiplier` contains a module for multiplying two matrices together, implemented in both VHDL and Verilog.
+
+The module takes two matrices ``A`` and ``B`` as inputs and provides the resulting matrix ``C`` as an output. When input ``i_valid`` is high on a rising clock edge, ``C`` is calculated and output, and ``o_valid`` goes high.
+
+The testbench defines a ``MatMultMonitor`` (a sub-class of :class:`.BusMonitor`) and a test case, which is provided in two versions:
+
+* ``run_test`` - For simulators that support multi-dimensional arrays
+* ``run_test_array`` -  For simulators that concatenate multi-dimensional arrays into a single long array
```
cmarqu: I would find the names ``ndim_array`` and ``1dim_array`` (also in the design etc.) more intuitive.

examples/matrix_multiplier/hdl/matrix_multiplier.v
```diff
@@ -0,0 +1,57 @@
...
+reg [(DATA_WIDTH*A_COLUMNS_B_ROWS)-1:0] C[0:A_ROWS-1][0:B_COLUMNS-1];
+reg                                     o_valid;
+
+reg [(DATA_WIDTH*A_COLUMNS_B_ROWS)-1:0] C_RESULT;
+
+integer i;
+integer j;
+integer k;
+
+    always @(posedge clk) begin
```
cmarqu: FWIW, not using a dedicated reset feels strange to me as an ASIC guy, but I can get used to it. :)

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,181 @@
...
+        yield gen_random_matrix(A_ROWS, A_COLUMNS_B_ROWS)
+
+def gen_b(num_samples=NUM_SAMPLES):
+    """Generate random matrix data for B
+    """
+    for i in range(num_samples):
+        yield gen_random_matrix(A_COLUMNS_B_ROWS, B_COLUMNS)
+
+
+if cocotb.SIM_NAME.lower().startswith(("icarus", "ghdl")):
```
cmarqu: If this is a general limitation, we should add this to the Simulator Support section in the docs.

examples/matrix_multiplier/tests/Makefile
```diff
@@ -0,0 +1,79 @@
...
+# Matrix parameters
+DATA_WIDTH ?= 32
+A_ROWS ?= 10
+B_COLUMNS ?= 4
+A_COLUMNS_B_ROWS ?= 6
+
+ifeq ($(TOPLEVEL_LANG),verilog)
+    VERILOG_SOURCES = $(WPWD)/../hdl/matrix_multiplier.v
+
+    # Set module parameters
```
cmarqu: For Cadence simulators, the option is ``-gpg`` for both parameters and generics, but I have to look up the exact formatting of the arguments.

eric-wieser:
There were some unaddressed comments in #1286 that would be good to apply here, I think.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,181 @@
+import os
+from random import getrandbits
+
```
eric-wieser: See https://github.com/cocotb/cocotb/pull/1286#discussion_r364016538 for some changes suggested to this file

eric-wieser: Nevermind, you already did this. I think I never wore the comment I'm thinking of.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,181 @@
...
+    # Do multiplication
+    for A, B in zip(a_data(), b_data()):
+        yield RisingEdge(dut.clk)
+        for i, row in enumerate(A):
+            for j, value in enumerate(row):
+                dut.A[(i*A_COLUMNS_B_ROWS)+j] = value
+        for i, row in enumerate(B):
+            for j, value in enumerate(row):
+                dut.B[(i*B_COLUMNS)+j] = value
```
eric-wieser: Can you make these three lines a helper function, `assign_matrix(dut.B, value)`?

I think it would be clearer if you can combine the overall tests for the two sets of simulator support, and just swap out the behavior of some helper functions instead.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,181 @@
...
+    expected_output = []
+
+    def calc_c(transaction):
+        a_array, b_array, c_array = transaction
+
+        c_expected = [
+            BinaryValue(
+                sum(
+                    [
+                        a_array[(i*A_COLUMNS_B_ROWS)+n] * b_array[(n*B_COLUMNS)+j]
```
eric-wieser: This should be
```suggestion
                        a_array[I][n] * b_array[n][j]
```
Due to the use of unpack_matrix

examples/matrix_multiplier/hdl/matrix_multiplier.vhdl
```diff
@@ -0,0 +1,55 @@
...
+
+    signal clk : std_logic := '0';
+    signal i_valid : std_logic := '0';
+    signal A : a_type := (others => (others => (others => '0')));
+    signal B : b_type := (others => (others => (others => '0')));
+    signal C : c_type := (others => (others => (others => '0')));
+    signal o_valid : std_logic := '0';
+begin
+
+    process(clk, i_valid)
```
elgorwi: Only 'clk' should be in the sensitivity list, 'i_valid' should be removed

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,181 @@
...
+NUM_SAMPLES = int(os.environ.get('NUM_SAMPLES', 3000))
+DATA_WIDTH = int(os.environ['DATA_WIDTH'])
+A_ROWS = int(os.environ['A_ROWS'])
+B_COLUMNS = int(os.environ['B_COLUMNS'])
+A_COLUMNS_B_ROWS = int(os.environ['A_COLUMNS_B_ROWS'])
+
+
+class MatMultMonitor(BusMonitor):
+
+    _signals = ["A", "B", "i_valid", "C", "o_valid"]
```
ktbarrett: It seems like you could just instance a data/valid monitor at the input and an output, set the software model as the callback on the input monitor, and have that model populate the expected data list. But I also don't know anything about the included base classes. They seem hard to use.

examples/matrix_multiplier/hdl/matrix_multiplier.v
eric-wieser: But it's fine for an FPGA, right? I don't think cocotb needs to guarantee that examples are synthesizable - we're a simulation tool, not a synthesis one.

On the other hand, if it doesn't add much complexity there's little harm in making this synthesizable either.

garmin-mjames:
Thoughts on synchronous vs asynchronous resets?

ktbarrett:
@garmin-mjames Synchronous IMO, but it doesn't really matter.

cmarqu:
Asynchronous assert, synchronous deassert, active-low is ASIC style. But I agree with Kaleb, it doesn't really matter here.

examples/matrix_multiplier/hdl/matrix_multiplier.vhdl
```diff
@@ -0,0 +1,55 @@
...
+use ieee.numeric_std.all;
+
+entity matrix_multiplier is
+generic(
+    DATA_WIDTH : positive := 8;
+    A_ROWS : positive := 8;
+    B_COLUMNS : positive := 5;
+    A_COLUMNS_B_ROWS : positive := 4
+    );
+end matrix_multiplier;
```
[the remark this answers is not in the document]
elgorwi: That would be nice, however arrays on ports would require SystemVerilog. Not sure if there is enough SystemVerilog support in Icarus.
In VHDL we would need to use 2008, or define the types in a package, but that would not allow overriding the generics.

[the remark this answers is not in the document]
ktbarrett: Can't say much about Verilog. But you're right, we would need to define a 2 dim unconstrained matrix type in a pack which would require 2008. Given how poor simulator support for 2008 is (shame on all of you!), this isn't a good idea.

examples/matrix_multiplier/hdl/matrix_multiplier.v
ktbarrett: It's bad practice for FPGAs, the design can't be reset without a power cycle. Implementation is as simple as adding a reset condition in the process and setting `o_valid` to 0.

garmin-mjames:
Needs #1507 first.

examples/matrix_multiplier/tests/Makefile
garmin-mjames: I found and added options for `ius`, but can't test them.

examples/matrix_multiplier/tests/Makefile
```diff
@@ -0,0 +1,85 @@
...
+    ifeq ($(SIM),icarus)
+        COMPILE_ARGS += -Pmatrix_multiplier.DATA_WIDTH=$(DATA_WIDTH) -Pmatrix_multiplier.A_ROWS=$(A_ROWS) -Pmatrix_multiplier.B_COLUMNS=$(B_COLUMNS) -Pmatrix_multiplier.A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifeq ($(SIM),$(filter $(SIM),questa aldec))
+        SIM_ARGS += -gDATA_WIDTH=$(DATA_WIDTH) -gA_ROWS=$(A_ROWS) -gB_COLUMNS=$(B_COLUMNS) -gA_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifeq ($(SIM),vcs)
+        COMPILE_ARGS += -pvalue+matrix_multiplier/DATA_WIDTH=$(DATA_WIDTH) -pvalue+matrix_multiplier/A_ROWS=$(A_ROWS) -pvalue+matrix_multiplier/B_COLUMNS=$(B_COLUMNS) -pvalue+matrix_multiplier/A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifeq ($(SIM),ius)
```
cmarqu: You can change this line (and the one for VHDL) to
```
ifneq ($(filter $(SIM),ius xcelium),)
```
because the options will be the same for both simulators.
BTW, we recently arrived at the ``ifneq`` style to handle ``SIM``being empty, I guess it would be good to change the lines for questa and aldec accordingly, even if only for general consistency.

examples/matrix_multiplier/tests/Makefile
```diff
@@ -0,0 +1,85 @@
...
+ifeq ($(SIM),$(filter $(SIM),GHDL aldec))
+    PSTAT_DIR = ./sim_build/
+else
+    PSTAT_DIR = ./
+endif
+
+test_profile.pstat: sim
+
+callgraph.svg: test_profile.pstat
+	python -m gprof2dot -f pstats $(PSTAT_DIR)$< | $(DOT_BINARY) -Tsvg -o $@
```
eric-wieser: This is likely to invoke the wrong python version

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,159 @@
...
+
+NUM_SAMPLES = int(os.environ.get('NUM_SAMPLES', 3000))
+DATA_WIDTH = int(os.environ['DATA_WIDTH'])
+A_ROWS = int(os.environ['A_ROWS'])
+B_COLUMNS = int(os.environ['B_COLUMNS'])
+A_COLUMNS_B_ROWS = int(os.environ['A_COLUMNS_B_ROWS'])
+
+MULTI_DIMENSIONAL_ARRAYS = not cocotb.SIM_NAME.lower().startswith(("icarus", "ghdl"))
+
+class MatrixUtils(object):
```
eric-wieser:
```suggestion
class MatrixUtils:
```

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,159 @@
...
+        else:
+            return [[signal[(i * cols) + j].value for j in range(cols)] for i in range(rows)]
+
+    def set_matrix_value(self, signal, matrix_value):
+        if self.multi_dimension:
+            signal <= matrix_value
+        else:
+            for i, row in enumerate(matrix_value):
+                for j, value in enumerate(row):
+                    signal[(i*len(row))+j] <= value
```
eric-wieser: Perhaps simpler as:
```suggestion
MULTI_DIMENSIONAL_ARRAYS = not cocotb.SIM_NAME.lower().startswith(("icarus", "ghdl"))

if MULTI_DIMENSIONAL_ARRAYS:
    def get_matrix_value(self, signal, rows, cols):
        return signal.value

    def set_matrix_value(self, signal, matrix_value):
        signal <= matrix_value
else:
    def get_matrix_value(self, signal, rows, cols):
        return [[signal[(i * cols) + j].value for j in range(cols)] for i in range(rows)]

    def set_matrix_value(self, signal, matrix_value):
        for i, row in enumerate(matrix_value):
            for j, value in enumerate(row):
                signal[(i*len(row))+j] <= value
```

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,159 @@
...
+
+class MatrixMonitor(BusMonitor):
+
+    def __init__(self, dut, utils, callback=None, event=None):
+        self.utils = utils
+        self.clock = dut.clk
```
eric-wieser:
```suggestion
```
`BusMonitor.__init_` does this for you

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,159 @@
...
+
+            if self.bus.valid.value:
+                a_matrix = self.utils.get_matrix_value(self.bus.A, A_ROWS, A_COLUMNS_B_ROWS)
+                b_matrix = self.utils.get_matrix_value(self.bus.B, A_COLUMNS_B_ROWS, B_COLUMNS)
+
+                # Calculate the expected result of C
+                c_expected = [
+                    [
+                        BinaryValue(
+                            value = sum(
```
eric-wieser: PEP8 says
```suggestion
                            value=sum(
```

garmin-mjames:
Rebased and addressed comments.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,163 @@
...
+                signal[(i * cols) + j].value
+                for j in range(cols)
+            ]
+            for i in range(rows)
+        ]
+
+    def set_matrix_value(signal, matrix_value):
+        for i, row in enumerate(matrix_value):
+            for j, value in enumerate(row):
+                signal[(i * len(row)) + j] <= value
```
eric-wieser, line 40: One more idea, not sure if it's worthwhile:
```suggestion
class UnpackedMatrixSignal:
    def __init__(self, signal, rows, cols):
        self.signal = signal
        self.rows = rows
        self.cols = cols

    @property
    def value(self):
        return [
            [
                self.signal[(i * self.cols) + j].value
                for j in range(self.cols)
            ]
            for i in range(self.rows)
        ]

    @value.setter
    def value(signal, matrix_value):
        for i, row in enumerate(matrix_value):
            for j, value in enumerate(row):
                self.signal[(i * len(row)) + j].value = value

if MULTI_DIMENSIONAL_ARRAYS:
    matrix_signal = lambda x, rows, cols: x
else:
    matrix_signal = UnpackedMatrixSignal
```

garmin-mjames:
@imphil 
Originally I started with ports, but there are some considerations that pushed me towards internal signals, which were also discussed previously in this PR.

- Requires SystemVerilog for 2D array
    - I wasn't able to get `iverilog` to compile it, but I have little experience with Verilog/SystemVerilog
- Requires VHDL 2008 and a package

I'm not against doing it that way but the consensus favored internal signals for broader compatibility.

I used the prefix `i_`/`o_` to make use of `BusMonitor` easier. I could use a `dict` mapping logical names to signal names rather than a "named" bus, but this was simpler.
As for the mixed case, I opted for the conventional matrix names **`AB = C`** for matrix multiplication.

I haven't used SystemVerilog, but doing some web searching, array support looks much better than plain Verilog. I can rework the module.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
garmin-mjames: I've used wrapper classes around signals before, but there might be something simpler.

Remove `set_matrix_value` and just generate 1D array since we recover values from the signals anyway on `i_valid` assertion to calculate expected result.
```python3
if MULTI_DIMENSIONAL_ARRAYS:
    def gen_matrix(func, rows, cols):
        return [[func(DATA_WIDTH) for col in range(cols)] for row in range(rows)]
else:
    def gen_matrix(func, rows, cols):
        return [func(DATA_WIDTH) for col in range(cols) for row in range(rows)]
```
Not sure about `get_matrix value`. It might be nice to use the `value` getter on the top matrix handle, then "unflatten" to 2D in Python.

imphil:
No worries, I'll follow up with a proposed implementation later today. Having flattened arrays as a port is actually quite common, would that be OK to still show the benchmarking properties you're looking for in this example?

garmin-mjames:
We could go the route of flattening the arrays. It would simplify the test, as we wouldn't distinguish between simulators that support VPI/VHPI access to multi-dimensional arrays and those that don't. It shouldn't affect the benchmarking, as that's focused on heavy use of `BinaryValue` objects for the values in the matrices.

I think the design files are clearer and easier to understand using 2D arrays, but this example module is fairly trivial.

gretzteam:
This may or may not be useful for this...

2D array in module ports forces the use of a wrapper when running gate level simulation so I try to avoid them. I prefer to flatten the ports, and use the _new_ streaming operator to go back to arrays within the module. It is fully synthesizeable.
This also allowed me to keep using verilator (which doesn't make 2D array accessible through cocotb yet). Unfortunately, Icarus doesn't seem to support the streaming operator yet, so it only moves the problem around...

For example

```
module test(
input [79:0] din //10 x 8-bit 
);

logic [9:0]][7:0] din_array;
assign din_array = {>>{din}};
```

You can go back and forth with different 'streaming' style. More details here:
https://www.amiq.com/consulting/2017/06/23/how-to-unpack-data-using-the-systemverilog-streaming-operators/

https://www.amiq.com/consulting/2017/05/29/how-to-pack-data-using-systemverilog-streaming-operators/

elgorwi:
> This may or may not be useful for this...
> 
> 2D array in module ports forces the use of a wrapper when running gate level simulation so I try to avoid them. I prefer to flatten the ports, and use the _new_ streaming operator to go back to arrays within the module. It is fully synthesizeable.
> This also allowed me to keep using verilator (which doesn't make 2D array accessible through cocotb yet). Unfortunately, Icarus doesn't seem to support the streaming operator yet, so it only moves the problem around...
> 
> For example
> 
> ```
> module test(
> input [79:0] din //10 x 8-bit 
> );
> 
> logic [9:0]][7:0] din_array;
> assign din_array = {>>{din}};
> ```
> 
> You can go back and forth with different 'streaming' style. More details here:
> https://www.amiq.com/consulting/2017/06/23/how-to-unpack-data-using-the-systemverilog-streaming-operators/
> 
> https://www.amiq.com/consulting/2017/05/29/how-to-pack-data-using-systemverilog-streaming-operators/

I believe the streaming operator is unnecessary here as the array in your example is unpacked.

```
module test(
input [79:0] din //10 x 8-bit 
);
 
logic [9:0] [7:0] din_array;
assign din_array = din;
```

gretzteam:
> I believe the streaming operator is unnecessary here as the array in your example is packed.
> 
> ```
> module test(
> input [79:0] din //10 x 8-bit 
> );
>  
> logic [9:0] [7:0] din_array;
> assign din_array = din;
> ```

I didn't know you could do that! 
The streaming works with unpacked array as well but this is quite nice!
Thanks

garmin-mjames:
@imphil Would you prefer packed arrays as top-level ports?

examples/matrix_multiplier/hdl/matrix_multiplier.vhdl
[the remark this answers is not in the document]
elgorwi: > @imphil Would you prefer packed arrays as top-level ports?

@garmin-mjames Just updated the mean example with a SystemVerilog version. I found that access to unpacked array ports is in fact supported in icarus 10.3. This was using a 1-d unpacked array. However, I could not get packed array ports working under icarus. Maybe worthwhile investigating if 2-d unpacked arrays ports are supported in icarus 10.3?
If not, then a wrapper file without ports as you have already done here, but rather instantiating a conventional matrix_multiplier module with array ports will make it clear that this is a workaround for a simulator limitation. Simulators that do support array ports properly can then use the conventional matrix_multiplier module (with array ports) instead of the wrapper module.
Then for GHDL, we can use a similar approach.

examples/matrix_multiplier/tests/Makefile
```diff
@@ -0,0 +1,85 @@
+TOPLEVEL_LANG ?= verilog
+SIM ?= icarus
+
+PWD=$(shell pwd)
+
+ifeq ($(OS),Msys)
```
elgorwi, line 6: There have been some changes to how paths are handled, so this needs to be removed.

examples/matrix_multiplier/tests/Makefile
```diff
@@ -0,0 +1,85 @@
...
+    VHDL_SOURCES = $(WPWD)/../../mean/hdl/mean_pkg.vhd $(WPWD)/../hdl/matrix_multiplier.vhd
+
+    ifneq ($(filter $(SIM),GHDL questa aldec),)
+        # GHDL, questa, and aldec all use SIM_ARGS with '-g' for setting generics
+        SIM_ARGS += -gDATA_WIDTH=$(DATA_WIDTH) -gA_ROWS=$(A_ROWS) -gB_COLUMNS=$(B_COLUMNS) -gA_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifneq ($(filter $(SIM),ius xcelium),)
+        SIM_ARGS += -generic "matrix_multiplier:DATA_WIDTH=>$(DATA_WIDTH)" -generic "matrix_multiplier:A_ROWS=>$(A_ROWS)" -generic "matrix_multiplier:B_COLUMNS=>$(B_COLUMNS)" -generic "matrix_multiplier:A_COLUMNS_B_ROWS=>$(A_COLUMNS_B_ROWS)"
+    endif
+
+    ifeq ($(SIM),GHDL)
```
elgorwi, line 42: I've done some experiments with arrays using GHDL. However, cocotb could not discover the array at all. 
Does this example run correctly with GHDL? Which version did you use?

examples/matrix_multiplier/hdl/matrix_multiplier.v
```diff
@@ -0,0 +1,58 @@
...
+        end else begin
+            o_valid <= 0;
+        end
+    end
+
+    // Dump waves
+    initial begin
+        $dumpfile("dump.vcd");
+        $dumpvars(1, matrix_multiplier);
+    end
```
elgorwi, line 56: When running with Icarus, the .vcd file will not contain the arrays. The solution is to loop over all the array elements and call $dumpvars. Have a look at cocotb/examples/mean/rtl/mean,sv for an example how to dump arrays.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,163 @@
...
+from cocotb.monitors import BusMonitor
+from cocotb.regression import TestFactory
+from cocotb.scoreboard import Scoreboard
+from cocotb.triggers import RisingEdge, ReadOnly
+
+NUM_SAMPLES = int(os.environ.get('NUM_SAMPLES', 3000))
+DATA_WIDTH = int(os.environ['DATA_WIDTH'])
+A_ROWS = int(os.environ['A_ROWS'])
+B_COLUMNS = int(os.environ['B_COLUMNS'])
+A_COLUMNS_B_ROWS = int(os.environ['A_COLUMNS_B_ROWS'])
```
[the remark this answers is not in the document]
garmin-mjames: I don't know how I would do that with the `TestFactory` setup.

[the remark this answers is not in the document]
ktbarrett: Well I guess this will have to do until #1827.

examples/matrix_multiplier/hdl/matrix_multiplier.v
```diff
@@ -0,0 +1,58 @@
...
+                    end
+                    o_C[i][j] <= C_RESULT;
+                end
+            end
+
+            o_valid <= 1;
+        end else begin
+            o_valid <= 0;
+        end
+    end
```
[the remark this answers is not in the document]
imphil: It's correct, but uncommon and generally discouraged to mix blocking and non-blocking assignments.

examples/matrix_multiplier/tests/Makefile
garmin-mjames: I wanted it to at least compile correctly in GHDL. Arrays do not work yet, but see https://github.com/ghdl/ghdl/pull/1249.

garmin-mjames:
@imphil I opted for ports and used your style guide in #1831 as a starting point, then made `iverilog` work.

documentation/source/newsfragments/1502.doc.rst
```diff
@@ -0,0 +1 @@
+New example ``matrix_multiplier``.
```
ktbarrett, line 1: Could you add a file link?

examples/matrix_multiplier/hdl/matrix_multiplier.sv
```diff
@@ -0,0 +1,74 @@
...
+    if (reset_i) begin
+      valid_o <= 1'b0;
+
+      for (idx = 0; idx < (A_ROWS * B_COLUMNS); idx++) begin
+        c_o[idx] <= '0;
+      end
+    end else begin
+      valid_o <= valid_i;
+
+      if (valid_i) begin
```
ktbarrett, line 51: This condition shouldn't be here.

examples/matrix_multiplier/hdl/matrix_multiplier.vhd
```diff
@@ -0,0 +1,75 @@
...
+  begin
+
+    if (rising_edge(clk_i)) then
+      if (reset_i) then
+        valid_o <= '0';
+        c_o <= (others => (others => '0'));
+      else
+        valid_o <= valid_i;
+
+        if (valid_i) then
```
ktbarrett, line 67: Or here.

examples/matrix_multiplier/hdl/matrix_multiplier.sv
ktbarrett: At least in FPGA designs (or ASIC based on my educated guesses :^) we avoid conditioning the datapath:
1. it's unnecessary, `valid` marks the data as valid or not, so the value output here doesn't matter
2. it increases resource utilization, every bit is now muxed
3. it decreases performance, extra logic means timing issues with fast clocks

eric-wieser: It might be reasonable to do the current conditioning with an `else begin c_o <= 'bx; end`. A synthesis too worth it's salt should cancel out those X's to what you suggest, but it catches more bugs in simulation.

examples/matrix_multiplier/hdl/matrix_multiplier.sv
```diff
@@ -0,0 +1,74 @@
...
+  output logic                    valid_o,
+  input        [DATA_WIDTH-1:0]   a_i[A_ROWS * A_COLUMNS_B_ROWS],
+  input        [DATA_WIDTH-1:0]   b_i[A_COLUMNS_B_ROWS * B_COLUMNS],
+  output logic [C_DATA_WIDTH-1:0] c_o[A_ROWS * B_COLUMNS]
+);
+
+  logic [C_DATA_WIDTH-1:0] c_element;
+  logic [C_DATA_WIDTH-1:0] c_calc[A_ROWS * B_COLUMNS];
+
+  integer i, j, k, idx;
```
eric-wieser, line 27: Can you see declare these within the always blocks below?

examples/matrix_multiplier/tests/Makefile
```diff
@@ -0,0 +1,80 @@
...
+A_ROWS ?= 10
+B_COLUMNS ?= 4
+A_COLUMNS_B_ROWS ?= 6
+
+ifeq ($(TOPLEVEL_LANG),verilog)
+    VERILOG_SOURCES = $(PWD)/../hdl/matrix_multiplier.sv
+
+    # Set module parameters
+    ifeq ($(SIM),icarus)
+        COMPILE_ARGS += -Pmatrix_multiplier.DATA_WIDTH=$(DATA_WIDTH) -Pmatrix_multiplier.A_ROWS=$(A_ROWS) -Pmatrix_multiplier.B_COLUMNS=$(B_COLUMNS) -Pmatrix_multiplier.A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
```
eric-wieser, line 20: This is looking like something we might want cocotb to provide a script to abstract over.

eric-wieser: (in a different PR in future)

elgorwi: There used to be a ```GENERICS``` parameter in the makefiles for that purpose. Unfortunately that was removed a while ago.

examples/matrix_multiplier/hdl/matrix_multiplier.v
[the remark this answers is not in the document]
ktbarrett: Is it possible to declare the accumulator variable in the loop? Maybe he could if he used a generate loop for the outer loop.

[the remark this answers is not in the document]
eric-wieser: > Is it possible to declare the accumulator variable in the loop? Maybe he could if he used a generate loop for the outer loop.

Yes, I think something like this would be possible:

```verilog
  always @(*) begin
    for (i = 0; i < A_ROWS; i = i + 1) begin
      for (j = 0; j < B_COLUMNS; j = j + 1) begin : dot_product
        reg c_element = 0;
        for (k = 0; k < A_COLUMNS_B_ROWS; k = k + 1) begin
          c_element = c_element + (a_i[(i * A_COLUMNS_B_ROWS) + k] * b_i[(k * B_COLUMNS) + j]);
        end
        c_calc[(i * B_COLUMNS) + j] = c_element;
      end
    end
  end
```

garmin-mjames:
The force push didn't propogate changes here, so I was going to try closing and re-opening.

Now I can't reopen due to:
> The matrix_mult branch was force-pushed or recreated.

examples/matrix_multiplier/hdl/matrix_multiplier.vhd
```diff
@@ -0,0 +1,75 @@
...
+  );
+end entity matrix_multiplier;
+
+architecture rtl of matrix_multiplier is
+
+  signal c_calc : flat_matrix_type(c_o'RANGE)(c_o(0)'RANGE);
+
+begin
+
+  multiply : process (a_i, b_i) is
```
ktbarrett, line 35:
```suggestion
  multiply : process (all) is
```
VHDL 2008 feature, I'd test it in few simulators first. I know Questa supports it.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,143 @@
...
+
+    cocotb.fork(Clock(dut.clk_i, 5, units='ns').start())
+
+    # Configure Scoreboard to compare module results to expected
+    expected_output = []
+
+    def add_expected(transaction):
+        expected_output.append(transaction)
+
+    in_monitor = MatrixInMonitor(dut, callback=add_expected)
```
ktbarrett, line 92: A slight simplification.
```suggestion
    in_monitor = MatrixInMonitor(dut, callback=expected_output.append)
```

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,143 @@
...
+    out_monitor = MatrixOutMonitor(dut)
+
+    scoreboard = Scoreboard(dut)
+    scoreboard.add_interface(out_monitor, expected_output)
+
+    # Initial values
+    dut.valid_i <= 0
+    dut.a_i <= next(gen_a(1, lambda x: 0))
+    dut.b_i <= next(gen_b(1, lambda x: 0))
```
ktbarrett, line 101: This is not particularly readable IMO. See the following comments for the full suggestion.
```suggestion
    dut.a_i <= gen_a(lambda x: 0)
    dut.b_i <= gen_b(lambda x: 0)
```

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,143 @@
...
+
+def gen_b(num_samples=NUM_SAMPLES, func=getrandbits):
+    """Generate random matrix data for B"""
+    for i in range(num_samples):
+        yield gen_matrix(func, A_COLUMNS_B_ROWS, B_COLUMNS)
+
+
+factory = TestFactory(test_multiply)
+factory.add_option('a_data', [gen_a])
+factory.add_option('b_data', [gen_b])
```
ktbarrett, line 142:
```suggestion
factory.add_option('a_data', [lambda: (gen_a() for _ in range(NUM_SAMPLES)])
factory.add_option('b_data', [lambda: (gen_b() for _ in range(NUM_SAMPLES)])
```

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def gen_matrix(func, rows, cols):
+    return [func(DATA_WIDTH) for row in range(rows) for col in range(cols)]
+
+
+def gen_a(num_samples=NUM_SAMPLES, func=getrandbits):
+    """Generate random matrix data for A"""
+    for i in range(num_samples):
+        yield gen_matrix(func, A_ROWS, A_COLUMNS_B_ROWS)
```
ktbarrett, line 131:
```suggestion
    return gen_matrix(func, A_ROWS, A_COLUMNS_B_ROWS)
```

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def gen_b(num_samples=NUM_SAMPLES, func=getrandbits):
+    """Generate random matrix data for B"""
+    for i in range(num_samples):
+        yield gen_matrix(func, A_COLUMNS_B_ROWS, B_COLUMNS)
```
ktbarrett, line 137:
```suggestion
    return gen_matrix(func, A_COLUMNS_B_ROWS, B_COLUMNS)
```

examples/matrix_multiplier/hdl/matrix_multiplier.vhd
garmin-mjames: Compiles in GHDL, and works with Aldec simulators as well.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
garmin-mjames: The problem is lambdas mess up the auto-generated test description:
```
Description: Automatically generated test

    a_data: <lambda> (No docstring supplied)
    b_data: <lambda> (No docstring supplied)
```

examples/matrix_multiplier/tests/test_matrix_multiplier.py
garmin-mjames: Added `create_a` and `create_b` to improve readability.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
ktbarrett: Fun. Oh well.

examples/matrix_multiplier/hdl/matrix_multiplier.sv
```diff
@@ -0,0 +1,72 @@
+// This file is public domain, it can be freely copied without restrictions.
+// SPDX-License-Identifier: CC0-1.0
+
+// Matrix Multiplier DUT
+`default_nettype none
```
elgorwi, line 5: This should either be removed or alternatively the nettype of all input ports should be explicitly defined as 'wire'

examples/matrix_multiplier/hdl/matrix_multiplier.sv
```diff
@@ -0,0 +1,72 @@
...
+        for (int k = 0; k < A_COLUMNS_B_ROWS; k = k + 1) begin
+          c_element = c_element + (a_i[(i * A_COLUMNS_B_ROWS) + k] * b_i[(k * B_COLUMNS) + j]);
+        end
+        c_calc[(i * B_COLUMNS) + j] = c_element;
+      end
+    end
+  end
+
+  always @(posedge clk_i) begin
+    integer idx;
```
elgorwi, line 40: This can be declared inside the loop statement similar to what was done on lines 28-29.

examples/matrix_multiplier/hdl/matrix_multiplier.sv
```diff
@@ -0,0 +1,72 @@
...
+      end
+    end
+  end
+
+  // Dump waves
+  initial begin
+    integer idx;
+    $dumpfile("dump.vcd");
+    $dumpvars(1, matrix_multiplier);
+    for (idx = 0; idx < (A_ROWS * A_COLUMNS_B_ROWS); idx++) begin
```
elgorwi, line 61: I recommend adding an ifdef specific for Icarus as Modelsim emits warnings when $dumpvars is called in this manner. This is done already in the 'mean' example.

examples/matrix_multiplier/tests/Makefile
```diff
@@ -0,0 +1,80 @@
...
+    else ifneq ($(filter $(SIM),questa modelsim riviera activehdl),)
+        SIM_ARGS += -gDATA_WIDTH=$(DATA_WIDTH) -gA_ROWS=$(A_ROWS) -gB_COLUMNS=$(B_COLUMNS) -gA_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifeq ($(SIM),vcs)
+        COMPILE_ARGS += -pvalue+matrix_multiplier/DATA_WIDTH=$(DATA_WIDTH) -pvalue+matrix_multiplier/A_ROWS=$(A_ROWS) -pvalue+matrix_multiplier/B_COLUMNS=$(B_COLUMNS) -pvalue+matrix_multiplier/A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifneq ($(filter $(SIM),ius xcelium),)
+        EXTRA_ARGS += -defparam "matrix_multiplier.DATA_WIDTH=$(DATA_WIDTH)" -defparam "matrix_multiplier.A_ROWS=$(A_ROWS)" -defparam "matrix_multiplier.B_COLUMNS=$(B_COLUMNS)" -defparam "matrix_multiplier.A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)"
+    endif
+
```
elgorwi, line 28: Could you add Verilator as well?

ktbarrett: Would Verilator support need additional work on the HDL?

elgorwi: Don't think any changes would be required, but best to try and see what happens.

examples/matrix_multiplier/tests/test_matrix_multiplier.py
```diff
@@ -0,0 +1,148 @@
...
+
+            if self.bus.valid.value:
+                c_actual = self.bus.C.value
+                self._recv(c_actual)
+
+
+async def test_multiply(dut, a_data, b_data):
+    """Test multiplication of many matrices."""
+
+    cocotb.fork(Clock(dut.clk_i, 5, units='ns').start())
```
ktbarrett, line 84:
```suggestion
    cocotb.fork(Clock(dut.clk_i, 10, units='ns').start())
```

examples/matrix_multiplier/tests/Makefile
```diff
@@ -0,0 +1,82 @@
...
+    VHDL_SOURCES = $(PWD)/../../mean/hdl/mean_pkg.vhd $(PWD)/../hdl/matrix_multiplier_pkg.vhd $(PWD)/../hdl/matrix_multiplier.vhd
+
+    ifneq ($(filter $(SIM),GHDL questa modelsim riviera activehdl),)
+        # GHDL, questa, and aldec all use SIM_ARGS with '-g' for setting generics
+        SIM_ARGS += -gDATA_WIDTH=$(DATA_WIDTH) -gA_ROWS=$(A_ROWS) -gB_COLUMNS=$(B_COLUMNS) -gA_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifneq ($(filter $(SIM),ius xcelium),)
+        SIM_ARGS += -generic "matrix_multiplier:DATA_WIDTH=>$(DATA_WIDTH)" -generic "matrix_multiplier:A_ROWS=>$(A_ROWS)" -generic "matrix_multiplier:B_COLUMNS=>$(B_COLUMNS)" -generic "matrix_multiplier:A_COLUMNS_B_ROWS=>$(A_COLUMNS_B_ROWS)"
+    endif
+
+    ifeq ($(SIM),GHDL)
```
garmin-mjames, line 45: Should **GHDL** be lower-case?

ktbarrett: [The docs](https://docs.cocotb.org/en/stable/simulator_support.html#ghdl) say lower case.

garmin-mjames:
@elgorwi Could you test the changes with Verilator?

ktbarrett:
@garmin-mjames It fails. Generics are not being passed properly.

garmin-mjames:
I couldn't tell from the Verilator documentation whether I need the top-level module name in the generic parameters. The examples in the docs don't have it, so I didn't put it in.

examples/matrix_multiplier/tests/Makefile
```diff
@@ -0,0 +1,82 @@
...
+
+    # Set module parameters
+    ifeq ($(SIM),icarus)
+        COMPILE_ARGS += -Pmatrix_multiplier.DATA_WIDTH=$(DATA_WIDTH) -Pmatrix_multiplier.A_ROWS=$(A_ROWS) -Pmatrix_multiplier.B_COLUMNS=$(B_COLUMNS) -Pmatrix_multiplier.A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifneq ($(filter $(SIM),questa modelsim riviera activehdl),)
+        SIM_ARGS += -gDATA_WIDTH=$(DATA_WIDTH) -gA_ROWS=$(A_ROWS) -gB_COLUMNS=$(B_COLUMNS) -gA_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifeq ($(SIM),vcs)
+        COMPILE_ARGS += -pvalue+matrix_multiplier/DATA_WIDTH=$(DATA_WIDTH) -pvalue+matrix_multiplier/A_ROWS=$(A_ROWS) -pvalue+matrix_multiplier/B_COLUMNS=$(B_COLUMNS) -pvalue+matrix_multiplier/A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
+    else ifeq ($(SIM),verilator)
+        COMPILE_ARGS += -GDATA_WIDTH=$(DATA_WIDTH) -GA_ROWS=$(A_ROWS) -GB_COLUMNS=$(B_COLUMNS) -GA_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
```
garmin-mjames, line 26: This might work:
```suggestion
    else ifneq ($(filter $(SIM),vcs verilator),)
        COMPILE_ARGS += -pvalue+matrix_multiplier/DATA_WIDTH=$(DATA_WIDTH) -pvalue+matrix_multiplier/A_ROWS=$(A_ROWS) -pvalue+matrix_multiplier/B_COLUMNS=$(B_COLUMNS) -pvalue+matrix_multiplier/A_COLUMNS_B_ROWS=$(A_COLUMNS_B_ROWS)
```

ktbarrett:
Nvm me. I was a little behind, before you added support to the makefile for verilator. It runs in Verilator. We should work on adding verilator to the CI.

garmin-mjames:
Pinging @imphil.

imphil:
Close/reopen to get a fresh private CI run (which was failing before). If that one passes LGTM. Thanks a lot @garmin-mjames for your persistence on this one, it looks like a really nice example now -- and sorry for being so slow to actually getting back to you. You've at least earned a drink of your choice should we meet at some time :smile:

imphil (approved):

PR #1507, opened 2020-03-19
Allow setting array values with lists.

garmin-mjames:
This change allows setting the value of a NonHierarchyIndexableObject
handle with a Python list. The values of the list indexes will be
assigned to the sub-handles in range order from left to right. Supports
multi-dimensional arrays.

Fix value getter for NonHierarchyIndexableObject. Now supports arrays
that don't start at index 0, and gives value in range order left to right.

Added a test for this because `test_array` is skipped by iverilog.

When adding support for arrays that don't start at index 0, I opted to go with mapping index `0` of the python list to `range'left` and the last index to `range'right`. I'm not sure if this is better than using `range'low` to `range'high`, but matches how we assign `binstr` values.

Also I'm not sure if this would go under `bugfix` or `feature`.

- [x] Document **left-to-right** assignment in NonHierarchyIndexableObject.value docstring
    - [sphinx preview here](https://cocotb--1507.org.readthedocs.build/en/1507/library_reference.html#cocotb.handle.NonHierarchyIndexableObject)
- [x] Release note regarding change of behavior, bugfix, additional feature

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -237,19 +237,8 @@ class HierarchyObject(RegionObject):
             return SimHandleBase.__setattr__(self, name, value)
         if self.__hasattr__(name) is not None:
             sub = self.__getattr__(name)
-            if type(sub) is NonHierarchyIndexableObject:
-                if type(value) is not list:
-                    raise AttributeError("Attempting to set %s which is a NonHierarchyIndexableObject to something other than a list?" % (name))
-
-                if len(sub) != len(value):
-                    raise IndexError("Attempting to set %s with list length %d but target has length %d" % (
-                        name, len(value), len(sub)))
-                for idx in range(len(value)):
-                    sub[idx] = value[idx]
-                return
-            else:
-                sub.value = value
-                return
+            sub.value = value
+            return
         if name in self._compat_mapping:
             return SimHandleBase.__setattr__(self, name, value)
         raise AttributeError("Attempt to access %s which isn't present in %s" %(
@@ -482,15 +471,7 @@ class NonHierarchyIndexableObject(NonHierarchyObject):
 
     def __setitem__(self, index, value):
         """Provide transparent assignment to indexed array handles."""
-        if type(value) is list:
-            if len(value) != len(self.__getitem__(index)):
-                raise IndexError("Assigning list of length %d to object %s of length %d" % (
-                    len(value), self.__getitem__(index)._fullname, len(self.__getitem__(index))))
-            self._log.info("Setting item %s to %s", self.__getitem__(index)._fullname, value)
-            for idx in range(len(value)):
-                self.__getitem__(index).__setitem__(idx, value[idx])
-        else:
-            self.__getitem__(index).value = value
+        self[index].value = value
 
     def __getitem__(self, index):
         if isinstance(index, slice):
@@ -530,11 +511,52 @@ class NonHierarchyIndexableObject(NonHierarchyObject):
 
     @NonHierarchyObject.value.getter
     def value(self):
-        # need to iterate over the sub-object
-        result = []
-        for x in range(len(self)):
-            result.append(self[x].value)
-        return result
+        """A list of each value within this simulation object.
+
+        Getting and setting the current value of an array is done
+        by iterating through sub-handles in left-to-right order.
+
+        Given an HDL array ``arr``:
+
+        +--------------+---------------------+--------------------------------------------------------------+
+        | Verilog      | VHDL                | ``arr.value`` is equivalent to                               |
+        +==============+=====================+==============================================================+
+        | ``arr[4:7]`` | ``arr(4 to 7)``     | ``[arr[4].value, arr[5].value, arr[6].value, arr[7].value]`` |
+        +--------------+---------------------+--------------------------------------------------------------+
+        | ``arr[7:4]`` | ``arr(7 downto 4)`` | ``[arr[7].value, arr[6].value, arr[5].value, arr[4].value]`` |
+        +--------------+---------------------+--------------------------------------------------------------+
+
+        When setting this property as in ``arr.value = ...``, the same index equivalence as noted in the table holds.
+
+        .. note::
+            When setting this property, the values will be cached as explained in :attr:`ModifiableObject.value`.
+
+        .. warning::
+            Assigning a value to a sub-handle:
+
+            - **Wrong**: ``dut.some_array.value[0] = 1`` (gets value as a list then updates index 0)
+            - **Correct**: ``dut.some_array[0].value = 1``
+        """
+        # Don't use self.__iter__, because it has an unwanted `except IndexError`
+        return [
+            self[i].value
+            for i in self._range_iter(self._range[0], self._range[1])
+        ]
+
+    @value.setter
+    def value(self, value):
+        """Assign value from a list of same length to an array in left-to-right order.
+        Index 0 of the list maps to the left-most index in the array.
+
+        See the docstring for :attr:`value` above.
+        """
+        if type(value) is not list:
+            raise TypeError("Assigning non-list value to object %s of type %s" % (self._name, type(self)))
+        if len(value) != len(self):
+            raise ValueError("Assigning list of length %d to object %s of length %d" % (
+                len(value), self._name, len(self)))
+        for val_idx, self_idx in enumerate(self._range_iter(self._range[0], self._range[1])):
+            self[self_idx].value = value[val_idx]
 
 
 class _SimIterator(collections.abc.Iterator):
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -379,7 +379,6 @@ class NonHierarchyObject(SimHandleBase):
 
     @property
     def value(self):
-        "A reference to the value"
         raise TypeError("Not permissible to get values of object %s of type %s" % (self._name, type(self)))
 
     def setimmediatevalue(self, value):
@@ -422,8 +421,8 @@ class NonHierarchyObject(SimHandleBase):
 class ConstantObject(NonHierarchyObject):
     """An object which has a value that can be read, but not set.
 
-    We can also cache the value since it is fixed at elaboration time and
-    won't change within a simulation.
+    The value is cached in the class since it is fixed at elaboration
+    time and won't change within a simulation.
     """
     def __init__(self, handle, path, handle_type):
         """
@@ -457,6 +456,7 @@ class ConstantObject(NonHierarchyObject):
 
     @NonHierarchyObject.value.getter
     def value(self):
+        """The value of this simulation object."""
         return self._value
 
     def __str__(self):
@@ -668,17 +668,23 @@ class ModifiableObject(NonConstantObject):
 
     @NonConstantObject.value.getter
     def value(self):
+        """The value of this simulation object.
+
+        .. note::
+            When setting this property, the value is stored by the :class:`~cocotb.scheduler.Scheduler`
+            and all stored values are written at the same time at the end of the current simulator time step.
+
+            Use :meth:`setimmediatevalue` to set the value immediately.
+        """
         binstr = simulator.get_signal_val_binstr(self._handle)
         result = BinaryValue(binstr, len(binstr))
         return result
 
     @value.setter
     def value(self, value):
-        """Intercept the store of a value and hold in cache.
+        """Assign value to this simulation object.
 
-        This operation is to enable all of the scheduled callbacks to complete
-        with the same read data and for the writes to occur on the next
-        sim time.
+        See the docstring for :attr:`value` above.
         """
         cocotb.scheduler.save_write(self, value)
 
--- a/documentation/source/newsfragments/1507.bugfix.rst
+++ b/documentation/source/newsfragments/1507.bugfix.rst
@@ -1,0 +1,1 @@
+Getting and setting the value of an :class:`~cocotb.handle.NonHierarchyIndexableObject` now iterates through the correct range of the simulation object, so arrays that do not start/end at index 0 are supported.
--- a/documentation/source/newsfragments/1507.bugfix.rst
+++ b/documentation/source/newsfragments/1507.bugfix.rst
@@ -1,1 +1,1 @@
-Getting and setting the value of an :class:`~cocotb.handle.NonHierarchyIndexableObject` now iterates through the correct range of the simulation object, so arrays that do not start/end at index 0 are supported.
+Getting and setting the value of a :class:`~cocotb.handle.NonHierarchyIndexableObject` now iterates through the correct range of the simulation object, so arrays that do not start/end at index 0 are supported.
--- a/documentation/source/newsfragments/1507.change.rst
+++ b/documentation/source/newsfragments/1507.change.rst
@@ -1,0 +1,2 @@
+:attr:`cocotb.handle.NonHierarchyIndexableObject.value` is now a list in left-to-right range order of the underlying simulation object.
+Previously the list was always ordered low-to-high.
--- a/documentation/source/newsfragments/1507.feature.rst
+++ b/documentation/source/newsfragments/1507.feature.rst
@@ -1,0 +1,13 @@
+New methods for setting the value of an :class:`~cocotb.handle.NonHierarchyIndexableObject` (HDL arrays). (:pr:`1507`)
+
+.. code-block:: python3
+
+    # Now supported
+    dut.some_array <= [0xAA, 0xBB, 0xCC]
+    dut.some_array.value = [0xAA, 0xBB, 0xCC]
+
+    # For simulators that support n-dimensional arrays
+    dut.some_2d_array <= [[0xAA, 0xBB], [0xCC, 0xDD]]
+    dut.some_2d_array.value = [[0xAA, 0xBB], [0xCC, 0xDD]]
+
+.. consume the towncrier issue number on this line.
--- a/documentation/source/newsfragments/1507.feature.rst
+++ b/documentation/source/newsfragments/1507.feature.rst
@@ -1,4 +1,4 @@
-New methods for setting the value of an :class:`~cocotb.handle.NonHierarchyIndexableObject` (HDL arrays). (:pr:`1507`)
+New methods for setting the value of a :class:`~cocotb.handle.NonHierarchyIndexableObject` (HDL arrays). (:pr:`1507`)
 
 .. code-block:: python3
 
--- a/tests/test_cases/test_array_simple/test_array_simple.py
+++ b/tests/test_cases/test_array_simple/test_array_simple.py
@@ -1,0 +1,197 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+"""Test getting and setting values of arrays"""
+import contextlib
+import logging
+
+import cocotb
+from cocotb.binary import BinaryValue
+from cocotb.clock import Clock
+from cocotb.result import TestFailure
+from cocotb.triggers import Timer
+
+tlog = logging.getLogger("cocotb.test")
+
+def _check_value(tlog, hdl, expected):
+    if hdl.value != expected:
+        raise TestFailure("{2!r}: Expected >{0}< but got >{1}<".format(expected, hdl.value, hdl))
+    else:
+        tlog.info("   Found {0!r} ({1}) with value={2}".format(hdl, hdl._type, hdl.value))
+
+
+@cocotb.test()
+async def test_1dim_array_handles(dut):
+    """Test getting and setting array values using the handle of the full array."""
+
+    cocotb.fork(Clock(dut.clk, 1000, 'ns').start())
+
+    # Set values with '<=' operator
+    dut.array_7_downto_4 <= [0xF0, 0xE0, 0xD0, 0xC0]
+    dut.array_4_to_7     <= [0xB0, 0xA0, 0x90, 0x80]
+    dut.array_3_downto_0 <= [0x70, 0x60, 0x50, 0x40]
+    dut.array_0_to_3     <= [0x30, 0x20, 0x10, 0x00]
+
+    await Timer(1000, 'ns')
+
+    _check_value(tlog, dut.array_7_downto_4, [0xF0, 0xE0, 0xD0, 0xC0])
+    _check_value(tlog, dut.array_4_to_7    , [0xB0, 0xA0, 0x90, 0x80])
+    _check_value(tlog, dut.array_3_downto_0, [0x70, 0x60, 0x50, 0x40])
+    _check_value(tlog, dut.array_0_to_3    , [0x30, 0x20, 0x10, 0x00])
+
+    # Set values through HierarchyObject.__setattr__ method on 'dut' handle
+    dut.array_7_downto_4 = [0x00, 0x11, 0x22, 0x33]
+    dut.array_4_to_7     = [0x44, 0x55, 0x66, 0x77]
+    dut.array_3_downto_0 = [0x88, 0x99, 0xAA, 0xBB]
+    dut.array_0_to_3     = [0xCC, 0xDD, 0xEE, 0xFF]
+
+    await Timer(1000, 'ns')
+
+    _check_value(tlog, dut.array_7_downto_4, [0x00, 0x11, 0x22, 0x33])
+    _check_value(tlog, dut.array_4_to_7    , [0x44, 0x55, 0x66, 0x77])
+    _check_value(tlog, dut.array_3_downto_0, [0x88, 0x99, 0xAA, 0xBB])
+    _check_value(tlog, dut.array_0_to_3    , [0xCC, 0xDD, 0xEE, 0xFF])
+
+
+@cocotb.test(skip=cocotb.SIM_NAME.lower().startswith(("icarus", "ghdl")))
+async def test_ndim_array_handles(dut):
+    """Test getting and setting multi-dimensional array values using the handle of the full array."""
+
+    cocotb.fork(Clock(dut.clk, 1000, 'ns').start())
+
+    # Set values with '<=' operator
+    dut.array_2d <= [
+        [0xF0, 0xE0, 0xD0, 0xC0],
+        [0xB0, 0xA0, 0x90, 0x80]
+    ]
+
+    await Timer(1000, 'ns')
+
+    _check_value(tlog, dut.array_2d, [[0xF0, 0xE0, 0xD0, 0xC0], [0xB0, 0xA0, 0x90, 0x80]])
+
+    # Set values through HierarchyObject.__setattr__ method on 'dut' handle
+    dut.array_2d = [
+        [0x00, 0x11, 0x22, 0x33],
+        [0x44, 0x55, 0x66, 0x77]
+    ]
+
+    await Timer(1000, 'ns')
+
+    _check_value(tlog, dut.array_2d, [[0x00, 0x11, 0x22, 0x33], [0x44, 0x55, 0x66, 0x77]])
+
+
+@cocotb.test()
+async def test_1dim_array_indexes(dut):
+    """Test getting and setting values of array indexes."""
+
+    cocotb.fork(Clock(dut.clk, 1000, 'ns').start())
+
+    dut.array_7_downto_4 <= [0xF0, 0xE0, 0xD0, 0xC0]
+    dut.array_4_to_7     <= [0xB0, 0xA0, 0x90, 0x80]
+    dut.array_3_downto_0 <= [0x70, 0x60, 0x50, 0x40]
+    dut.array_0_to_3     <= [0x30, 0x20, 0x10, 0x00]
+
+    await Timer(1000, 'ns')
+
+    # Check indices
+    _check_value(tlog, dut.array_7_downto_4[7], 0xF0)
+    _check_value(tlog, dut.array_7_downto_4[4], 0xC0)
+    _check_value(tlog, dut.array_4_to_7[4]    , 0xB0)
+    _check_value(tlog, dut.array_4_to_7[7]    , 0x80)
+    _check_value(tlog, dut.array_3_downto_0[3], 0x70)
+    _check_value(tlog, dut.array_3_downto_0[0], 0x40)
+    _check_value(tlog, dut.array_0_to_3[0]    , 0x30)
+    _check_value(tlog, dut.array_0_to_3[3]    , 0x00)
+    _check_value(tlog, dut.array_0_to_3[1]    , 0x20)
+
+    # Get sub-handles through NonHierarchyIndexableObject.__getitem__
+    dut.array_7_downto_4[7] <= 0xDE
+    dut.array_4_to_7[4]     <= 0xFC
+    dut.array_3_downto_0[0] <= 0xAB
+    dut.array_0_to_3[1]     <= 0x7A
+    dut.array_0_to_3[3]     <= 0x42
+
+    await Timer(1000, 'ns')
+
+    _check_value(tlog, dut.array_7_downto_4[7], 0xDE)
+    _check_value(tlog, dut.array_4_to_7[4]    , 0xFC)
+    _check_value(tlog, dut.array_3_downto_0[0], 0xAB)
+    _check_value(tlog, dut.array_0_to_3[1]    , 0x7A)
+    _check_value(tlog, dut.array_0_to_3[3]    , 0x42)
+
+    # Set sub-handle values through NonHierarchyIndexableObject.__setitem__
+    dut.array_7_downto_4[7] = 0x77
+    dut.array_4_to_7[4]     = 0x44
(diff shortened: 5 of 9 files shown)


tests/test_cases/test_array_simple/test_array_simple.py
```diff
@@ -0,0 +1,79 @@
...
+    """Test getting and setting array values using the handle of the full array.
+    """
+
+    # Set values with '<=' operator
+    dut.array_7_downto_4 <= [0xF0, 0xE0, 0xD0, 0xC0]
+    dut.array_4_to_7     <= [0xB0, 0xA0, 0x90, 0x80]
+    dut.array_3_downto_0 <= [0x70, 0x60, 0x50, 0x40]
+    dut.array_0_to_3     <= [0x30, 0x20, 0x10, 0x00]
+
+    yield Timer(1000)
```
cmarqu:
```suggestion
    yield Timer(1000, 'ns')
```
Let's slowly add the units in the tests too.

cocotb/handle.py
```diff
@@ -532,10 +521,21 @@ def _range_iter(self, left, right):
     def value(self):
         # need to iterate over the sub-object
         result = []
-        for x in range(len(self)):
-            result.append(self[x].value)
+        for sub_handle in self:
+            result.append(sub_handle.value)
         return result
```
eric-wieser: Perhaps while you're here, just:
```suggestion
return [sub_handle.value for sub_handle in self]
```

cocotb/handle.py
```diff
@@ -532,10 +521,21 @@ def _range_iter(self, left, right):
...
 
+    @value.setter
+    def value(self, value):
+        if type(value) is list:
+            if len(value) != len(self):
+                raise IndexError("Assigning list of length %d to object %s of length %d" % (
+                    len(value), self._name, len(self)))
+            for val_idx, self_idx in enumerate(self._range_iter(self._range[0], self._range[1])):
+                self.__setitem__(self_idx, value[val_idx])
```
```suggestion
                self[self_idx] = value[val_idx]
```

tests/test_cases/test_array_simple/test_array_simple.py
```diff
@@ -0,0 +1,79 @@
...
+    _check_value(tlog, dut.array_7_downto_4[7], 0x00)
+    _check_value(tlog, dut.array_7_downto_4[4], 0x33)
+    _check_value(tlog, dut.array_4_to_7[4]    , 0x44)
+    _check_value(tlog, dut.array_4_to_7[7]    , 0x77)
+    _check_value(tlog, dut.array_3_downto_0[3], 0x88)
+    _check_value(tlog, dut.array_3_downto_0[0], 0xBB)
+    _check_value(tlog, dut.array_0_to_3[0]    , 0xCC)
+    _check_value(tlog, dut.array_0_to_3[3]    , 0xFF)
+
+    _check_value(tlog, dut.array_0_to_3[1]    , 0xDD)
```
eric-wieser: I don't like that this relies on the previous test having run - tests should be independent, because they can be run separately via environment variables. Can you copy-paste lines 24 - 28 to this test too?

eric-wieser:
> When adding support for arrays that don't start at index 0, I opted to go with mapping index `0` of the python list to `range'left` and the last index to `range'right`. I'm not sure if this is better than using `range'low` to `range'high`, but matches how we assign `binstr` values.

Can you elaborate a little more on what the different between these two choices is, preferably with an example?

garmin-mjames:
The question to answer is, should index `0` of the Python list map to the **left-most** index in the array, or to the **lowest** index.
With `downto`, the `left` of the range is the highest index, while with `to`, it's the lowest,

Given:
module.v
```
reg [7:0]  array_downto[3:0]; // left: 3, right: 0
reg [7:0]  array_to[0:3]; // left: 0, right: 3
```
test.py
```python
dut.array_downto = [0xAA, 0xBB, 0xCC, 0xDD]
dut.array_to     = [0xAA, 0xBB, 0xCC, 0xDD]
```
Assigning range left to right:
Py index | 0 | 1 | 2 | 3
----------- | - | - | - | -
Py value | 0xAA | 0xBB | 0xCC | 0xDD
downto value| 0xAA | 0xBB | 0xCC | 0xDD
downto range | 3 | 2 | 1 | 0
to value| 0xAA | 0xBB | 0xCC | 0xDD
to range | 0 | 1 | 2 | 3

Assigning range low to high (previous behavior):
Py index | 0 | 1 | 2 | 3
----------- | - | - | - | -
Py value | 0xAA | 0xBB | 0xCC | 0xDD
downto value| 0xDD | 0xCC | 0xBB | 0xAA
downto range | 3 | 2 | 1 | 0
to value| 0xAA | 0xBB | 0xCC | 0xDD
to range | 0 | 1 | 2 | 3

garmin-mjames:
I think we should use **left-to-right** assignment, as that matches iteration over the object, and the corrected `value` getter that relies on iteration.

eric-wieser:
Is there a way to spell `dut.array_to     = [0xAA, 0xBB, 0xCC, 0xDD]` within (system)verilog or VHDL?

garmin-mjames:
VHDL: `array_to <= (x"AA", x"BB", x"CC", x"DD");` matches **left-to-right**.

From the 2008 Ref Manual 9.3.3.3:
`the leftmost element of the value matches the left bound of the range`

You can also map to indexes explicitly:
`array_to <= (3 => x"AA", 2 => x"BB", 1 => x"CC", 0 => x"DD");`

eric-wieser:
How about in SystemVerilog?

garmin-mjames:
Found this in Verilog 2005 Ref Manual:

> An element can be assigned a value in a single assignment, but complete or partial array dimensions cannot. Nor can complete or partial array dimensions be used to provide a value to an expression. To assign a value to an element of an array, an index for every dimension shall be specified.

SystemVerilog 🤷‍♂

eric-wieser:
Systemverilog has `array_to <= '{1, 2, 3}`, which matches in the same way.

I don't particularly like the left-to-right matching, but it seems it's the choice consistent with the hdl design.

Can you make sure to put this choice in the docstring for `NonHierarchyIndexableObject.value`, if it's not already somewhere else?

cocotb/handle.py
```diff
@@ -530,11 +519,19 @@ def _range_iter(self, left, right):
...
+    @value.setter
+    def value(self, value):
+        if type(value) is list:
+            if len(value) != len(self):
+                raise IndexError("Assigning list of length %d to object %s of length %d" % (
+                    len(value), self._name, len(self)))
+            for val_idx, self_idx in enumerate(self._range_iter(self._range[0], self._range[1])):
+                self[self_idx] = value[val_idx]
+        else:
+            raise TypeError("Assigning non-list value to object %s of type %s" % (self._name, type(self)))
```
eric-wieser: Indentation is shorter as:

```suggestion
        if type(value) is not list:
            raise TypeError("Assigning non-list value to object %s of type %s" % (self._name, type(self)))
        if len(value) != len(self):
            raise IndexError("Assigning list of length %d to object %s of length %d" % (
                len(value), self._name, len(self)))
        for val_idx, self_idx in enumerate(self._range_iter(self._range[0], self._range[1])):
            self[self_idx] = value[val_idx]
```

cocotb/handle.py
```diff
@@ -530,11 +519,19 @@ def _range_iter(self, left, right):
 
     @NonHierarchyObject.value.getter
     def value(self):
-        # need to iterate over the sub-object
-        result = []
-        for x in range(len(self)):
-            result.append(self[x].value)
-        return result
+        # need to iterate over the sub-objects
+        return [sub_handle.value for sub_handle in self]
```
eric-wieser: I'm a little worried that actually you do need your original code here, because `__iter__` has a rather scary `except IndexError` that probably isn't what you want.

eric-wieser:
I assume this was motivated by #1502.

Over in that PR, you had two branches - one which did iterative assignment like you do here, and one that did a single assignment all at once.

It sounds then like on some simulators this assignment was already working - if that's the case, which handle type was being used in those simulators?

garmin-mjames:
Originally I had:
```python
dut.A = A
dut.B = B
```
Which goes through `HierarchyObject.__setattr__`:
https://github.com/cocotb/cocotb/blob/8248bd336bdb0a641334f17a18176833428ae6fd/cocotb/handle.py#L240-L249

...which happens to work for arrays that start/end at `0`, and in the correct order if your array is ascending.

garmin-mjames:
When I switched it to `signal <= value` when refactoring to a helper function as you suggested, found this issue. I'll push changes to #1502 after this is resolved.

cocotb/handle.py
eric-wieser: Namely:
```suggestion
        return [
            self[i]
            for i in self._range_iter(self._range[0], self._range[1])
        ]
```

cocotb/handle.py
```diff
@@ -530,11 +519,69 @@ def _range_iter(self, left, right):
 
     @NonHierarchyObject.value.getter
     def value(self):
-        # need to iterate over the sub-object
-        result = []
-        for x in range(len(self)):
-            result.append(self[x].value)
-        return result
+        """Get value of array by iterating through sub-handles in left-to-right order.
```
eric-wieser: This docstring should be phrased for the getter and setter combined. Replacing 'Get' with 'The', and 'the returned' with 'this' should resolve that.

eric-wieser:
Possibly worth a note in the docs, applies both before and after this patch: `dut.some_array.value[0] = 1` will silently do nothing. I don't think we can do much about this other than a `.. warning::` telling users to not do that.

cocotb/handle.py
```diff
@@ -530,11 +519,69 @@ def _range_iter(self, left, right):
...
+            Verilog [4:7]       => [4, 5, 6, 7]
+            VHDL    (4 to 7)    => [4, 5, 6, 7]
+
+        Arrays with descending ranges:
+            Verilog [7:4]           => [7, 6, 5, 4]
+            VHDL    (7 downto 4)    => [7, 6, 5, 4]
+
+        Given an HDL array, the returned list contains the values of each sub-handle
+        in the order indicated in the table:
+        +----------------+-----------------------+------------------+
```
eric-wieser: Perhaps this needs a blank line before?

cocotb/handle.py
```diff
@@ -530,11 +519,69 @@ def _range_iter(self, left, right):
...
+        | ``array[4:7]`` | ``array(4 to 7)``     | ``array[4]`` gets ``value[0]`` |
+        |                |                       | ...                            |
+        |                |                       | ``array[7]`` gets ``value[3]`` |
+        +----------------+-----------------------+--------------------------------+
+        | ``array[7:4]`` | ``array(7 downto 4)`` | ``array[7]`` gets ``value[0]`` |
+        |                |                       | ...                            |
+        |                |                       | ``array[7]`` gets ``value[3]`` |
+        +----------------+-----------------------+--------------------------------+
+
+        """
```
eric-wieser: This docstring is not accessible unless you try really hard. You'll need to merge this with the getter docstring. I didn't notice it when I wrote my comment above

cocotb/handle.py
```diff
@@ -530,11 +519,69 @@ def _range_iter(self, left, right):
...
+        Arrays with descending ranges:
+            Verilog [7:4]           => [7, 6, 5, 4]
+            VHDL    (7 downto 4)    => [7, 6, 5, 4]
+
+        Given an HDL array, the returned list contains the values of each sub-handle
+        in the order indicated in the table:
+        +----------------+-----------------------+------------------+
+        | Verilog        | VHDL                  | Python list      |
+        +================+=======================+==================+
+        | ``array[4:7]`` | ``array(4 to 7)``     | ``[4, 5, 6, 7]`` |
```
eric-wieser: Would this perhaps be clearer as:
```suggestion
        | ``arr[4:7]`` | ``arr(4 to 7)``     | ``[arr[4], arr[5], arr[6], arr[7]]`` |
```

garmin-mjames:
Figured out how to render docs locally 👍

garmin-mjames:
Should we add suggested code for assigning lowest index to highest, or is that obvious?
```python
for idx, val in enumerate(value_list):
    dut.some_array[idx].value = val
```

cmarqu:
Showing the code would be a nice gesture IMO.

eric-wieser:
It's clearly not obvious, because what you suggest there is wrong!

I suspect we'd want to introduce a helper function to iterate over the indices in an appropriate order. Let's try not to go too deep down this rabbit-hole you're following - I'd be happy to leave it to another PR some other time, especially since neither VHDL or systemverilog has an obvious way to do it themselves.

garmin-mjames:
It's not a general solution, but works for arrays starting at 0.
I'll leave it for now.

cocotb/handle.py
```diff
@@ -530,11 +519,67 @@ def _range_iter(self, left, right):
 
     @NonHierarchyObject.value.getter
     def value(self):
-        # need to iterate over the sub-object
-        result = []
-        for x in range(len(self)):
-            result.append(self[x].value)
-        return result
+        """Reference to the value of the simulation object.
```
```suggestion
        """A list of each value within this simulation object.
```
`reference` has implications of the write through semantics that don't exist that you warn about below

garmin-mjames: How should we word it for the non-array classes.
Also I'm confused why `ModifiableObject` is a descendant of `NonHierarchyIndexableObject`.

cocotb/handle.py
```diff
@@ -530,11 +519,67 @@ def _range_iter(self, left, right):
...
+        | Verilog      | VHDL                | Array values                   |
+        +==============+=====================+================================+
+        | ``arr[4:7]`` | ``arr(4 to 7)``     | - ``arr[4]`` gets ``value[0]`` |
+        |              |                     | - ...                          |
+        |              |                     | - ``arr[7]`` gets ``value[3]`` |
+        +--------------+---------------------+--------------------------------+
+        | ``arr[7:4]`` | ``arr(7 downto 4)`` | - ``arr[7]`` gets ``value[0]`` |
+        |              |                     | - ...                          |
+        |              |                     | - ``arr[7]`` gets ``value[3]`` |
+        +--------------+---------------------+--------------------------------+
```
eric-wieser: I'd argue we can combine this table with the one above - the table above already contains the information, namely that
```python
val = arr.value
# means
val = [arr[4].value, arr[5].value, arr[6].value, arr[7].value]
```
and
```python
arr.value = val
# means
[arr[4].value, arr[5].value, arr[6].value, arr[7].value] = val
```

Perhaps the table becomes
```
        +--------------+---------------------+--------------------------------------------------------------+
        | Verilog      | VHDL                | ``arr.value`` is equivalent to                               |
        +==============+=====================+==============================================================+
        | ``arr[4:6]`` | ``arr(4 to 7)``     | ``[arr[4].value, arr[5].value, arr[6].value, arr[7].value]`` |
        +--------------+---------------------+--------------------------------------------------------------+
        | ``arr[7:4]`` | ``arr(7 downto 4)`` | ``[arr[7].value, arr[6].value, arr[5].value, arr[4].value]`` |
        +--------------+---------------------+--------------------------------------------------------------+

        Note that this property can be set, and the same equivalence holds in ``arr.value = ...``.
```

Not entirely sure how to word that last sentence.

garmin-mjames: That's pretty slick. Thanks for the feedback on this, good documentation can be hard.

cocotb/handle.py
eric-wieser: > How should we word it for the non-array classes.

Not sure. I assume you're referring to

https://github.com/cocotb/cocotb/blob/8248bd336bdb0a641334f17a18176833428ae6fd/cocotb/handle.py#L391-L394

> Also I'm confused why `ModifiableObject` is a descendant of `NonHierarchyIndexableObject`.

I'm afraid I can't help you there

tests/test_cases/test_array_simple/test_array_simple.py
```diff
@@ -0,0 +1,86 @@
...
+    _check_value(tlog, dut.array_7_downto_4, [0xF0, 0xE0, 0xD0, 0xC0])
+    _check_value(tlog, dut.array_4_to_7    , [0xB0, 0xA0, 0x90, 0x80])
+    _check_value(tlog, dut.array_3_downto_0, [0x70, 0x60, 0x50, 0x40])
+    _check_value(tlog, dut.array_0_to_3    , [0x30, 0x20, 0x10, 0x00])
+
+    # Set values through HierarchyObject.__setattr__ method on 'dut' handle
+    dut.array_7_downto_4 = [0x00, 0x11, 0x22, 0x33]
+    dut.array_4_to_7     = [0x44, 0x55, 0x66, 0x77]
+    dut.array_3_downto_0 = [0x88, 0x99, 0xAA, 0xBB]
+    dut.array_0_to_3     = [0xCC, 0xDD, 0xEE, 0xFF]
```
eric-wieser: What about `setimmediatevalue`?

cocotb/handle.py
```diff
@@ -530,11 +519,67 @@ def _range_iter(self, left, right):
...
+            for i in self._range_iter(self._range[0], self._range[1])
+        ]
+
+    @value.setter
+    def value(self, value):
+        """Assign value from a list of same length to an array in left-to-right order.
+        Index 0 of the list maps to the left-most index in the array.
+
+        See getter docstring above.
+        """
```
eric-wieser: This needs:
```suggestion
        """
        if type(self) is not NonHierarchyIndexableObject:
            # subclasses do not get this behavior
            return NonHierarchyObject.value.fset(value)
```

eric-wieser: Hmm. Actually, I think the fix is to implement `setimmediatevalue` instead of `value.setter`. That's automatically replaced by the derived classes.

cocotb/handle.py
```diff
@@ -530,11 +519,67 @@ def _range_iter(self, left, right):
...
+        if type(value) is not list:
+            raise TypeError("Assigning non-list value to object %s of type %s" % (self._name, type(self)))
+        if len(value) != len(self):
+            raise IndexError("Assigning list of length %d to object %s of length %d" % (
+                len(value), self._name, len(self)))
+        for val_idx, self_idx in enumerate(self._range_iter(self._range[0], self._range[1])):
+            self[self_idx] = value[val_idx]
```
eric-wieser: I'm thinking this should maybe be
```suggestion
    @value.setter
    def setimmediatevalue(self, value):
        """Assign value from a list of same length to an array in left-to-right order.
        Index 0 of the list maps to the left-most index in the array.

        See the docstring for :attr:`value` above
        """
        if type(value) is not list:
            raise TypeError("Assigning non-list value to object %s of type %s" % (self._name, type(self)))
        if len(value) != len(self):
            raise IndexError("Assigning list of length %d to object %s of length %d" % (
                len(value), self._name, len(self)))
        for val_idx, self_idx in enumerate(self._range_iter(self._range[0], self._range[1])):
            self[self_idx].setimmediatevalue(value[val_idx])
```

cocotb/handle.py
eric-wieser: It's a lot easier when the contributor writes such detailed explanations of the intent in the PR comments, as you've done above :)

cocotb/handle.py
garmin-mjames: This would break the `handle <= value` semantics of using the cached values.
`setimmediatevalue` only works for directly modifiable objects atm.

cocotb/handle.py
garmin-mjames: Changing the base class of `NonConstantObject` to `NonHierarchyObject` might help things here. I'm not sure of all of the implications though.

cocotb/handle.py
eric-wieser: Spotted a typo and updated.

You're right, for `setimmediatevalue` to work, this would need to move:
https://github.com/cocotb/cocotb/blob/8248bd336bdb0a641334f17a18176833428ae6fd/cocotb/handle.py#L653-L661

Since it doesn't implement `setimmediatevalue` at all today, I'm fine with putting it off till a later patch.

cocotb/handle.py
garmin-mjames: `_range` is set to `None` if it's not an indexable object, which disables `__iter__` and causes `__getitem__` to throw an `IndexError`.

eric-wieser:
I think none of my above comments apply any more, if we limit the scope to not touch `setimmediatevalue`.

cocotb/handle.py
garmin-mjames: I could implement both.

cocotb/handle.py
eric-wieser: My suggested change here is irrelevant - this doesn't actually matter because all it affects is the behavior of `NonConstantObject.value = ...`, and I already made a similar change to the getter in a previous commit.

No action is needed here. I think the implications are broad enough that I don't want to have to think about it as part of this PR!

cocotb/handle.py
eric-wieser: Let's leave that to a follow-up, and get this in as is.

garmin-mjames:
Still need a newsfragment. It's a change that fixes a bug at the same time, and also a new feature. Hmm.

eric-wieser:
You can add multiple news fragments of different types in the same PR. If that helps

cocotb/handle.py
```diff
@@ -486,7 +475,7 @@ def __setitem__(self, index, value):
             if len(value) != len(self.__getitem__(index)):
                 raise IndexError("Assigning list of length %d to object %s of length %d" % (
                     len(value), self.__getitem__(index)._fullname, len(self.__getitem__(index))))
-            self._log.info("Setting item %s to %s", self.__getitem__(index)._fullname, value)
+            self._log.debug("Setting item %s to %s", self.__getitem__(index)._fullname, value)
             for idx in range(len(value)):
                 self.__getitem__(index).__setitem__(idx, value[idx])
         else:
```
garmin-mjames: Can these `__getitem__` calls be changed to `[]` syntax?

cocotb/handle.py
```diff
@@ -530,11 +519,52 @@ def _range_iter(self, left, right):
...
+        +==============+=====================+==============================================================+
+        | ``arr[4:6]`` | ``arr(4 to 7)``     | ``[arr[4].value, arr[5].value, arr[6].value, arr[7].value]`` |
+        +--------------+---------------------+--------------------------------------------------------------+
+        | ``arr[7:4]`` | ``arr(7 downto 4)`` | ``[arr[7].value, arr[6].value, arr[5].value, arr[4].value]`` |
+        +--------------+---------------------+--------------------------------------------------------------+
+
+        When setting this property as in ``arr.value = ...``, the same index equivalence as noted in the table holds.
+
+        .. note::
+            When setting this property, the values will be cached as explained in :meth:`ModifiableObject.value`.
```
```suggestion
            When setting this property, the values will be cached as explained in :attr:`ModifiableObject.value`.
```
Otherwise this shows parens.

cocotb/handle.py
eric-wieser: Yes. Better yet, this entire `if` statement can now be replaced with it's else clause, I think.

eric-wieser: If you want to go down this road, please add a test for 2d arrays.

garmin-mjames: I'm glad I checked, because...
```python
for idx in range(len(value)):
```
...is wrong.

cocotb/handle.py
```diff
@@ -530,11 +519,52 @@ def _range_iter(self, left, right):
...
+
+        See the docstring for :attr:`value` above.
+        """
+        if type(value) is not list:
+            raise TypeError("Assigning non-list value to object %s of type %s" % (self._name, type(self)))
+        if len(value) != len(self):
+            raise IndexError("Assigning list of length %d to object %s of length %d" % (
+                len(value), self._name, len(self)))
+        for val_idx, self_idx in enumerate(self._range_iter(self._range[0], self._range[1])):
+            self[self_idx] = value[val_idx]
```
garmin-mjames: Should this be:
```suggestion
            self[self_idx].value = value[val_idx]
```
Otherwise it goes through `__setitem__`, which resolves to the same thing once I fix it per other comment.

eric-wieser: It doesn't make much difference. My reading of your comment if that you're worried about recursion here, but you needn't be.

Having said that, using `.value` will avoid some indirection and be slightly faster.

documentation/source/newsfragments/1507.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Getting and setting the value of an :attr:`cocotb.handle.NonHierarchyIndexableObject` now iterates through the correct range of the simulation object, so arrays that do not start/end at index 0 are supported.
+Setting value using ``dut.some_array = [...]`` and getting value using ``dut.some_array.value`` work correctly.
```
eric-wieser: Not sure this sentence is needed here.

documentation/source/newsfragments/1507.change.rst
```diff
@@ -0,0 +1,5 @@
+Assigning a list to :attr:`cocotb.handle.NonHierarchyIndexableObject` is now done by iterating through sub-handles in left-to-right order.
+Previously iteration used ``range(len(value))``.
+
+Getting :attr:`cocotb.handle.NonHierarchyIndexableObject.value` now returns a list in left-to-right range order.
```
eric-wieser:
```suggestion
:attr:`cocotb.handle.NonHierarchyIndexableObject.value` is now a list in left-to-right range order.
```
And then the 3 lines above aren't needed

documentation/source/newsfragments/1507.feature.rst
```diff
@@ -0,0 +1,9 @@
+New methods for setting the value of an :attr:`cocotb.handle.NonHierarchyIndexableObject` (HDL arrays). (:pr:`1507`)
```
eric-wieser:
```suggestion
New methods for setting the value of an :attr:`~cocotb.handle.NonHierarchyIndexableObject` (HDL arrays). (:pr:`1507`)
```

cocotb/handle.py
eric-wieser: Well spotted!

garmin-mjames:
Added tests for 2D arrays. I'm not sure if there are other simulators that should skip these tests.

cocotb/handle.py
```diff
@@ -530,11 +511,52 @@ def _range_iter(self, left, right):
...
+    def value(self, value):
+        """Assign value from a list of same length to an array in left-to-right order.
+        Index 0 of the list maps to the left-most index in the array.
+
+        See the docstring for :attr:`value` above.
+        """
+        if type(value) is not list:
+            raise TypeError("Assigning non-list value to object %s of type %s" % (self._name, type(self)))
+        if len(value) != len(self):
+            raise IndexError("Assigning list of length %d to object %s of length %d" % (
```
eric-wieser: Frankly this should probably be `ValueError`
```suggestion
            raise ValueError("Assigning list of length %d to object %s of length %d" % (
```

eric-wieser: Could you add a test for this exception?

garmin-mjames: Sure.

cocotb/handle.py
```diff
@@ -530,11 +511,52 @@ def _range_iter(self, left, right):
...
+
+        Getting and setting the current value of an array is done
+        by iterating through sub-handles in left-to-right order.
+
+        Given an HDL array ``arr``:
+
+        +--------------+---------------------+--------------------------------------------------------------+
+        | Verilog      | VHDL                | ``arr.value`` is equivalent to                               |
+        +==============+=====================+==============================================================+
+        | ``arr[4:6]`` | ``arr(4 to 7)``     | ``[arr[4].value, arr[5].value, arr[6].value, arr[7].value]`` |
```
cmarqu:
```suggestion
        | ``arr[4:7]`` | ``arr(4 to 7)``     | ``[arr[4].value, arr[5].value, arr[6].value, arr[7].value]`` |
```

cocotb/handle.py
```diff
@@ -530,11 +511,52 @@ def _range_iter(self, left, right):
...
+
+        When setting this property as in ``arr.value = ...``, the same index equivalence as noted in the table holds.
+
+        .. note::
+            When setting this property, the values will be cached as explained in :attr:`ModifiableObject.value`.
+
+        .. warning::
+            Assigning a value to a sub-handle:
+
+            - **Wrong**: ``dut.some_array.value[0] = 1`` (Gets value as a list then updates index 0)
```
cmarqu:
```suggestion
            - **Wrong**: ``dut.some_array.value[0] = 1`` (gets value as a list then updates index 0)
```
Somehow, the capitalization looks strange to me... but feel free to disregard.

tests/test_cases/test_array_simple/test_array_simple.py
```diff
@@ -0,0 +1,174 @@
+"""
+Test getting and setting values of arrays
+"""
```
cmarqu:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause

"""Test getting and setting values of arrays"""
```

cocotb/handle.py
garmin-mjames: good catch

documentation/source/newsfragments/1507.bugfix.rst
```diff
@@ -0,0 +1 @@
+Getting and setting the value of an :class:`~cocotb.handle.NonHierarchyIndexableObject` now iterates through the correct range of the simulation object, so arrays that do not start/end at index 0 are supported.
```
cmarqu:
```suggestion
Getting and setting the value of a :class:`~cocotb.handle.NonHierarchyIndexableObject` now iterates through the correct range of the simulation object, so arrays that do not start/end at index 0 are supported.
```

documentation/source/newsfragments/1507.feature.rst
```diff
@@ -0,0 +1,13 @@
+New methods for setting the value of an :class:`~cocotb.handle.NonHierarchyIndexableObject` (HDL arrays). (:pr:`1507`)
```
cmarqu:
```suggestion
New methods for setting the value of a :class:`~cocotb.handle.NonHierarchyIndexableObject` (HDL arrays). (:pr:`1507`)
```

tests/test_cases/test_array_simple/test_array_simple.py
```diff
@@ -0,0 +1,197 @@
...
+from cocotb.binary import BinaryValue
+from cocotb.clock import Clock
+from cocotb.result import TestFailure
+from cocotb.triggers import Timer
+
+tlog = logging.getLogger("cocotb.test")
+
+def _check_value(tlog, hdl, expected):
+    if hdl.value != expected:
+        raise TestFailure("{2!r}: Expected >{0}< but got >{1}<".format(expected, hdl.value, hdl))
```
cmarqu: Nit: The 2-0-1 order is unnecessary here, or is there a deeper meaning that I'm missing?
This is also in ``test_array.py``.

tests/test_cases/test_array_simple/test_array_simple.py
```diff
@@ -0,0 +1,197 @@
...
+    cocotb.fork(Clock(dut.clk, 1000, 'ns').start())
+
+    dut.array_7_downto_4 <= [0xF0, 0xE0, 0xD0, 0xC0]
+    dut.array_4_to_7     <= [0xB0, 0xA0, 0x90, 0x80]
+    dut.array_3_downto_0 <= [0x70, 0x60, 0x50, 0x40]
+    dut.array_0_to_3     <= [0x30, 0x20, 0x10, 0x00]
+
+    await Timer(1000, 'ns')
+
+    # Check indices
```
cmarqu:
```suggestion
    # Check indexes
```
?

tests/test_cases/test_array_simple/test_array_simple.py
```diff
@@ -0,0 +1,197 @@
...
+    cocotb.fork(Clock(dut.clk, 1000, 'ns').start())
+
+    dut.array_2d <= [
+        [0xF0, 0xE0, 0xD0, 0xC0],
+        [0xB0, 0xA0, 0x90, 0x80]
+    ]
+
+    await Timer(1000, 'ns')
+
+    # Check indices
```
cmarqu:
```suggestion
    # Check indexes
```

tests/test_cases/test_array_simple/test_array_simple.py
garmin-mjames: I just copied the from the other `_check_*` functions for consistency.

tests/test_cases/test_array_simple/test_array_simple.py
garmin-mjames: I've always seen **indices** in a mathematics context relating to arrays or matrices.

cmarqu: Makes sense.

eric-wieser:
I'll merge this in a couple days if there are no more comments.

eric-wieser:
Thanks for following the rabbit-hole @garmin-mjames and fixing the sharp edges here rather than working around them in your other PRs.

PR #1510, opened 2020-03-19
Remove HierarchyObject.__hasattr__

eric-wieser:
This was introduced in 7ce68f1 as a hack because `__getattr__` contained log statements.
This is not a real magic method, and now that we do not log in `__getattr__` there is no need to have a "peek" accessor.

(note: this PR has had most of its contents moved to other PRs)

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -225,6 +225,28 @@ class RegionObject(SimHandleBase):
 class HierarchyObject(RegionObject):
     """Hierarchy objects are namespace/scope objects."""
 
+    def __get_sub_handle_by_name(self, name):
+        try:
+            return self._sub_handles[name]
+        except KeyError:
+            pass
+
+        # Cache to avoid a call to the simulator if we already know the name is
+        # invalid. Unclear if we care, but we had this before.
+        if name in self._invalid_sub_handles:
+            return None
+
+        new_handle = simulator.get_handle_by_name(self._handle, name)
+
+        if not new_handle:
+            self._invalid_sub_handles.add(name)
+            return None
+
+        sub_handle = SimHandle(new_handle, self._child_path(name))
+        self._sub_handles[name] = sub_handle
+        return sub_handle
+
+
     def __setattr__(self, name, value):
         """Provide transparent access to signals via the hierarchy.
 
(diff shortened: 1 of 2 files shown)


cocotb/handle.py
```diff
@@ -225,6 +225,28 @@ def __dir__(self):
 class HierarchyObject(RegionObject):
     """Hierarchy objects are namespace/scope objects."""
 
+    def __get_sub_handle_by_name(self, name):
+        try:
+            return self._sub_handles[name]
```
ktbarrett, line 230: This cache has the ability to store every subhandle, so there is the potential for this to take up at least 2x (if not 3-4x because of overhead from Python and PLI handles) the memory of the entire design. I have projects that I simulate with cocotb that have >1e6 accessible objects, keeping all of them the entire lifetime of the program is a bad idea. Could/should we instead move to `lru_cache`? When we finally tie PLI handle lifetimes to Python lifetimes, this will work nicely.

cocotb/handle.py
```diff
@@ -233,60 +255,34 @@ def __setattr__(self, name, value):
...
 
-        new_handle = simulator.get_handle_by_name(self._handle, name)
-        if new_handle:
-            self._sub_handles[name] = SimHandle(new_handle, self._child_path(name))
-        else:
-            self._invalid_sub_handles.add(name)
-        return new_handle
+        if name in self._compat_mapping:
+            return SimHandleBase.__getattr__(self, name)
+        raise AttributeError("%s contains no object named %s" % (self._name, name))
```
ktbarrett, line 285:
```suggestion
            return SimHandleBase.__getattr__(self, name)

        raise AttributeError("%s contains no object named %s" % (self._name, name))
```

cocotb/handle.py
ktbarrett: Actually, this isn't a great idea. If the design has deep narrow hierarchies, an LRU cache per parent node won't perform well.

cocotb/handle.py
eric-wieser: I assume you also want this in setattr?

ktbarrett: Yeah.

eric-wieser: Added

eric-wieser:
Added the blank lines @ktbarrett requested, they make sense to me too.

Will merge once CI passes.

eric-wieser:
Doc failure is #1603, I'm beginning to think that once RTD creates the environment badly, it locks it for the lifetime of a PR.

cocotb/handle.py
```diff
@@ -233,60 +255,36 @@ def __setattr__(self, name, value):
...
         """Query the simulator for a object with the specified name
         and cache the result to build a tree of objects.
         """
-        try:
-            return self._sub_handles[name]
-        except KeyError:
-            pass
-
-        if name.startswith("_"):
-            return SimHandleBase.__getattr__(self, name)
```
eric-wieser, line 257: Whoops, removing these was a regression I think, xref #1634

PR #1514, opened 2020-03-19
Make `BinaryValue.buff` bytes not str.

eric-wieser:
This makes the behavior consistent with python 2, where the two types were equivalent.

`str` is not appropriate for storing binary data.

Also adjusts the docstring to pass.

probably relates to #1381

Milestoned as 1.4 since this is part of #1289

### DIFF
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -84,8 +84,8 @@ class BinaryValue:
     >>> vec.integer = 42
     >>> print(vec.binstr)
     101010
-    >>> print(repr(vec.buff))
-    '*'
+    >>> print(vec.buff)
+    b'*'
 
     """
     _resolve_to_0     = "-lL"  # noqa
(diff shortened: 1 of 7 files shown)


cocotb/binary.py
```diff
@@ -325,36 +330,41 @@ def is_resolvable(self):
...
             bits = bits[8:]
             val = int(byte, 2)
             if self.big_endian:
-                buff += chr(val)
+                buff += [val]
             else:
-                buff = chr(val) + buff
-        return buff
+                buff = [val] + buff
+        return bytes(buff)
```
ktbarrett: Alternatively,

```python
n_bytes = ceil(len(bits) / 8)
val = int(bits, 2)
return val.to_bytes(n_bytes, byteorder=('big' if self.big_endian else 'little'))
```

cocotb/binary.py
```diff
@@ -325,36 +330,41 @@ def is_resolvable(self):
...
-    def buff(self, val):
+    def buff(self, val: bytes):
         self._str = ""
         for char in val:
             if self.big_endian:
-                self._str += "{0:08b}".format(ord(char))
+                self._str += "{0:08b}".format(char)
             else:
-                self._str = "{0:08b}".format(ord(char)) + self._str
+                self._str = "{0:08b}".format(char) + self._str
```
ktbarrett: Alternatively,

```python
val = int.from_bytes(val, byteorder=('big' if self.big_endian else 'little'))
self._str = format(val, "b")
```

cocotb/binary.py
eric-wieser: I'd probably want to leave that to a follow-up. I'm not 100% convinced that's the same, and would need to write some tests to be sure.

I also wonder if there's a way to avoid using the intermediate arbitrary-precision integer.

eric-wieser: Also, a neat trick you might enjoy that doesn't need floating point numbers - `n_bytes = -(-bits // 8)`

eric-wieser:
This has opened up a can of worms, the extent of which is described in a subsection of #1289.

cocotb/drivers/amba.py
```diff
@@ -352,7 +352,7 @@ def _read_data(self):
                     _burst_diff = burst_length - burst_count
                     _st = _araddr + (_burst_diff * bytes_in_beat)
                     _end = _araddr + ((_burst_diff + 1) * bytes_in_beat)
-                    word.buff = self._memory[_st:_end].tostring()
+                    word.buff = self._memory[_st:_end].tobytes()
```
eric-wieser: This change is #1546, so we should merge that PR first.

eric-wieser:
Release notes had a rendering issue, updated

eric-wieser:
Transient CI issue in RTD, closing and reopening to try and get a more favorable cache

eric-wieser:
CI failure is #1603, and there's nothing we can do about it.

cocotb/binary.py
eric-wieser: I don't disagree, I just don't want to mix behavior changes and cleanups in the same commit, because it makes them hard to tell apart. The simplification is nice, but I'd want to write some tests to verify the behavior is unchanged.

imphil: Great, thanks!

eric-wieser:
Comments hopefully addressed, will merge when CI passes.

cocotb/binary.py
eric-wieser: Well, maybe I do disagree - it seems a bit silly to convert from bytes to bitstring via an arbitrary-precision integer. Will write the test anyway.

PR #1515, opened 2020-03-20
Fix test name in regression summary

### LINKED ISSUE #1499: Test names are incorrect in summary
From https://github.com/cocotb/cocotb/pull/1380#issuecomment-599155133. Introduced in #1262. It's due to all the unnecessary monkeying the decorators and regression manager do to compute a name. The PR changed how the module name is passed to `_store_test_result`. It uses `__module__`, but on `RunningTest`s that value is `cocotb.decorator` instead of the module name of the underlying function.

ktbarrett:
Fix for regression introduced by 1abdd1d12f049415743f6fcc6b617228d2f85bcb which caused the incorrect test name to be displayed.

Fixes #1499. The refactor discussed will occur in a separate PR.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -216,7 +216,7 @@ class RegressionManager(object):
                                     sim_time_ns=repr(0),
                                     ratio_time=repr(0))
             result_pass, _ = self._score_test(test, cocotb.outcomes.Error(SimFailure()))
-            self._store_test_result(test.__module__, test.__name__, result_pass, 0, 0, 0)
+            self._store_test_result(test.module, test.funcname, result_pass, 0, 0, 0)
             if not result_pass:
                 self.xunit.add_failure()
                 self.failures += 1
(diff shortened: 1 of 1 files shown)


eric-wieser:
Something still looks wrong here:

https://travis-ci.org/github/cocotb/cocotb/jobs/664656576#L997

ktbarrett:
Keeping track of what's a `RunningTest` and what's a `test` is apparently hard for me.

Speaking of the refactor, is there a particular reason why tests are initialized all up front, then scheduled later? Why shouldn't they be initialized and scheduled in the same step?

eric-wieser:
No, I have no idea why they are all initialized up front. I think changing them to not be would be fine, and might even simplify things.

eric-wieser:
Let's just put this in, the Mac test is not going to care about a non-makefile change, and it'll free up some Mac capacity for someone else,

PR #1516, opened 2020-03-20
Better error messages when accidentally using yield in coroutines

### LINKED ISSUE #1512: Improve error reporting for `yield` inside async function
Those that are in the process of converting `cocotb.coroutine` functions using `yield` to async functions may run into problems where they miss a `yield`, which turns their coroutine into an async generator. If an async generator makes it to the scheduler it will result in a less than helpful error message. Instead we should check for async generators where we expect coroutines and throw a `TypeError` with a more helpful message, much like we do with old style coroutines missing `yield` statements.

ktbarrett:
Fixes #1512. There are a few points where we expect coroutines, so each place got a warning.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -85,12 +85,18 @@ class RunningTask(object):
         task.kill() will destroy a coroutine instance (and cause any Join
         triggers to fire.
     """
+
     def __init__(self, inst):
 
         if inspect.iscoroutine(inst):
             self._natively_awaitable = True
         elif inspect.isgenerator(inst):
             self._natively_awaitable = False
+        elif sys.version_info >= (3, 6) and inspect.isasyncgen(inst):
+            raise TypeError(
+                "{} is an async generator, not a coroutine. "
+                "You likely used the yield keyword instead of await.".format(
+                    inst.__qualname__))
         else:
             raise TypeError(
                 "%s isn't a valid coroutine! Did you forget to use the yield keyword?" % inst)
--- a/tests/test_cases/test_cocotb/test_async_generators.py
+++ b/tests/test_cases/test_cocotb/test_async_generators.py
@@ -1,0 +1,42 @@
+import cocotb
+
+
+async def whoops_async_generator():
+    # the user should have used `await` here, but they wrote `yield` by accident.
+    yield cocotb.triggers.Timer(1)
+
+
+@cocotb.test()
+def test_yielding_accidental_async_generator(dut):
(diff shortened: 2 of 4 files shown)


tests/test_cases/test_cocotb/test_async_generators.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+async def whoops_async_generator():
+    yield cocotb.triggers.Timer(1)
+
+
+@cocotb.test()
+def test_yielding_accidental_async_generator(dut):
+    yield cocotb.triggers.NullTrigger()
+
```
eric-wieser: Is this needed?

tests/test_cases/test_cocotb/test_async_generators.py
```diff
@@ -0,0 +1,44 @@
+import cocotb
+
+
+async def whoops_async_generator():
+    yield cocotb.triggers.Timer(1)
+
+
+@cocotb.test()
+def test_yielding_accidental_async_generator(dut):
```
eric-wieser:
```suggestion
@cocotb.test()
def test_yielding_accidental_async_generator(dut):
    # this test deliberately does not use `async def`, as we are testing the behavior of `yield`
```

tests/test_cases/test_cocotb/test_async_generators.py
```diff
@@ -0,0 +1,44 @@
+import cocotb
+
+
+async def whoops_async_generator():
+    yield cocotb.triggers.Timer(1)
```
eric-wieser:
```suggestion
    # the user should have used `await` here, but they wrote `yield` by accident.
    yield cocotb.triggers.Timer(1)
```

cocotb/decorators.py
```diff
@@ -85,12 +85,18 @@ class RunningTask(object):
         task.kill() will destroy a coroutine instance (and cause any Join
         triggers to fire.
     """
+
```
eric-wieser: Unless this is deliberate:
```suggestion
```

ktbarrett: It was deliberate. I know it is errant, but my IDE is highlighting the line (and a bunch of others) and it's starting to bother me.

tests/test_cases/test_cocotb/test_async_generators.py
ktbarrett: Nope.

eric-wieser:
Squashed the commits since the test doesn't pass without the change anyway.

PR #1517, opened 2020-03-20
 Updated tests and examples to use installed makefiles, not the ones pointed to by $COCOTB

ktbarrett:
Tests previously depended upon the `COCOTB` environment variable to obtain a reference to the standard makefile includes. This was deprecated recently (link); tests should instead use `cocotb-config --makefiles` to obtain a reference to the makefiles in the package installation directory.

The included regression tests were previously not updated, they required a compatibility layer that the regression scripts sourced. This meant that the existing regressions could not be run independently unless the user manually set the `COCOTB` variable. This PR fixes that issue by updating all the designs to use the new recommended method.

This changed slowed down the regressions notably (20s on master, 50s after change). So a new variable that caches the result of the shell out is used to avoid repeated shelling. After the fix, regressions are now faster (15s) due to the same change being made to the examples.

### DIFF
--- a/tests/designs/array_module/Makefile
+++ b/tests/designs/array_module/Makefile
@@ -54,5 +54,5 @@ else
     $(error "A valid value (verilog or vhdl) was not provided for TOPLEVEL_LANG=$(TOPLEVEL_LANG)")
 endif
 
-include $(COCOTB)/makefiles/Makefile.inc
-include $(COCOTB)/makefiles/Makefile.sim
+include $(shell cocotb-config --makefiles)/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.sim
--- a/tests/designs/avalon_streaming_module/Makefile
+++ b/tests/designs/avalon_streaming_module/Makefile
@@ -20,7 +20,7 @@ COCOTB?=$(WPWD)/../../..
 
 VERILOG_SOURCES = $(COCOTB)/tests/designs/avalon_streaming_module/avalon_streaming.sv
 
-include $(COCOTB)/makefiles/Makefile.inc
-include $(COCOTB)/makefiles/Makefile.sim
+include $(shell cocotb-config --makefiles)/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.sim
 
 endif
--- a/tests/designs/avalon_module/Makefile
+++ b/tests/designs/avalon_module/Makefile
@@ -49,7 +49,7 @@ COCOTB?=$(WPWD)/../../..
 
 VERILOG_SOURCES = $(COCOTB)/tests/designs/avalon_module/burst_read_master.v
 
-include $(COCOTB)/makefiles/Makefile.inc
-include $(COCOTB)/makefiles/Makefile.sim
+include $(shell cocotb-config --makefiles)/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.sim
 
 endif
--- a/tests/designs/basic_hierarchy_module/Makefile
+++ b/tests/designs/basic_hierarchy_module/Makefile
@@ -24,7 +24,7 @@ COCOTB?=$(WPWD)/../../..
 
 VERILOG_SOURCES = $(COCOTB)/tests/designs/basic_hierarchy_module/basic_hierarchy_module.v
 
-include $(COCOTB)/makefiles/Makefile.inc
-include $(COCOTB)/makefiles/Makefile.sim
+include $(shell cocotb-config --makefiles)/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.sim
 
 endif
--- a/tests/designs/close_module/Makefile
+++ b/tests/designs/close_module/Makefile
@@ -50,8 +50,8 @@ COCOTB?=$(WPWD)/../../..
 VERILOG_SOURCES = $(COCOTB)/tests/designs/close_module/close_module.v
 
 ifneq ($(SIM),vcs)
-include $(COCOTB)/makefiles/Makefile.inc
-include $(COCOTB)/makefiles/Makefile.sim
+include $(shell cocotb-config --makefiles)/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.sim
 else
 all:
 	@echo "Skipping test as system call overrides seqfault VCS"
--- a/tests/designs/multi_dimension_array/Makefile
+++ b/tests/designs/multi_dimension_array/Makefile
@@ -50,7 +50,7 @@ COCOTB?=$(WPWD)/../../..
 VERILOG_SOURCES = $(COCOTB)/tests/designs/multi_dimension_array/cocotb_array_pkg.sv \
                   $(COCOTB)/tests/designs/multi_dimension_array/cocotb_array.sv
 
-include $(COCOTB)/makefiles/Makefile.inc
-include $(COCOTB)/makefiles/Makefile.sim
+include $(shell cocotb-config --makefiles)/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.sim
 
 endif
--- a/tests/designs/plusargs_module/Makefile
+++ b/tests/designs/plusargs_module/Makefile
@@ -51,5 +51,5 @@ ifneq ($(filter $(SIM),ius xcelium),)
     SIM_ARGS += -v93
 endif
 
-include $(COCOTB)/makefiles/Makefile.inc
-include $(COCOTB)/makefiles/Makefile.sim
+include $(shell cocotb-config --makefiles)/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.sim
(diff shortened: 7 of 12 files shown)


ktbarrett:
Seems it fails because of the rule to always check for result.xml. That rule should be thought out some more. `test_exit_error` won't ever generate a result.xml, and a custom entry point might not either.

eric-wieser:
> This changed slowed down the regressions notably (20s on master, 50s after change)

This is because the config script is at `cocotb.config`, so to run it we first have to import the entirety of `cocotb`. Moving the script to `_cocotb/config`, and replacing `cocotb/config.py` with `from _cocotb.config import *` would likely help a lot.

themperek:
Why we need `COCOTB_INSTALL_MAKEFILES` at all?

ktbarrett:
@eric-wieser Ahhh, that would explain it, I thought it was a bit much for shell outs.

@themperek It is the cached result of `cocotb-config --makefiles` so we don't have to call `$(shell cocotb-config --makefiles)` in every test, only once at the top of the regression script. But as Eric as pointed out, the slow down was not what I thought it was and we can work around the slow down a different way.

eric-wieser:
The crucial part of that trick is to never put anything in `_cocotb/__init__.py` (other than perhaps a docstring)

eric-wieser:
Hmm, actually that's still a bit tricky - if you can't use `cocotb.__file__`, you'd need to mess around with `importlib` to find the file path without actually importing it.

themperek:
I would rather leave COCOTB variable than in this case than COCOTB_INSTALL_MAKEFILES (just my personal opinion).

themperek:
```
#time for i in {1..100}; do  python -c "print('a')"; done
real	0m2.074s
user	0m1.668s
sys	0m0.412s
```

```
#time for i in {1..100}; do cocotb-config --makefiles; done
real	0m14.332s
user	0m12.395s
sys	0m1.870s
````

**Comment by ktbarrett** [comment]
@eric-wieser I think you are on the right train of thought though, it would be nice if we had a python directory other than `cocotb` to put things not associated with the runtime in.

**Comment by eric-wieser** [comment]
I think `_cocotb` is fine as such a directory, there's precedent in `pytest` doing the same.

**Comment by themperek** [comment]
If I remove everything from `__init__.py` no difference.

**Comment by eric-wieser** [comment]
That seems very unlikely to me. Are you sure the __init__.py you modified is the one that is installed?

Good idea to test it that way first though.

**Comment by themperek** [comment]
>That seems very unlikely to me. Are you sure the init.py you modified is the one that is installed?

Seems like. If I do some syntax error in this file it crashes.

**Comment by eric-wieser** [comment]
Did you pip install with `-e`? Apparently `console_scripts` are _much_ slower if you do that vs installing without.

**Comment by themperek** [comment]
```
#time for i in {1..100}; do python cocotb/config.py --makefiles; done
real    0m4.769s
user    0m3.911s
sys     0m0.789s
```

Same but empty `__init__.py`:
```
real    0m2.744s
user    0m2.241s
sys     0m0.479s
```

**Comment by eric-wieser** [comment]
Wait, I thought you said an empty `__init__.py` was the same speed, not really fast?

**Comment by themperek** [comment]
with `pip install cocotb` (all default):
```
# time for i in {1..100}; do cocotb-config --makefiles; done
real    0m4.821s
user    0m4.080s
sys     0m0.748s
```

**Comment by eric-wieser** [comment]
Can you combine your comments into a single comment with all the cases side-by-side, for ease of comparison?

**Comment by themperek** [comment]
I looks to me the difference is not significant enough to do workarounds. Our tests are not realistic. The difference is a factor of ~2 which is ~0.02s/test.

**Comment by themperek** [comment]
> Can you combine your comments into a single comment with all the cases side-by-side, for ease of comparison?

OK. Above.


>Wait, I thought you said an empty __init__.py was the same speed, not really fast?

14.332s vs 0m13.399s Kind of same for me. Sorry.

**Comment by ktbarrett** [comment]
So what I'm getting out of this is that being installed in editable mode was the reason that the slowdown was so dramatic?

Just tested doing a full `pip install .`, change commit vs speedup commit of the full regression time.

Master: 17s
After change: 25.5s
After speedup: 16s

That being said, a lot of us develop in editable mode, do we care enough to address the problem? I think either way that any performance change can come in on a different commit.

**Comment by ktbarrett** [comment]
Final note: there is still a `COCOTB` variable in many of the makefiles, but they are computed in the files instead of it being a global variable, and they are only used to reference repo objects, not objects which are now installed.

**Comment by themperek** [comment]
You should be able to remove this line now: https://github.com/cocotb/cocotb/blob/master/Makefile#L33

**Comment by themperek** [comment]
>You should be able to remove this line now: >https://github.com/cocotb/cocotb/blob/master/Makefile#L33

Maybe not (sorry) but this may be better (from:https://github.com/cocotb/cocotb/pull/1445)
```
COCOTB:=$$(shell cocotb-config --share)
```

**Comment by eric-wieser** [comment]
I'm confused - is something still using the `COCOTB` var?

@themperek, I assume that extra dollar is accidental.

File: Makefile [hunk 1]
```diff
@@ -30,8 +30,6 @@
 .PHONY: all
 all: test
 
-include $(shell cocotb-config --makefiles)/Makefile.inc
-
```
eric-wieser: Let's put this line back and merge it as it was.

themperek: I will look at this at some point.

tests/designs/array_module/Makefile
```diff
@@ -54,5 +54,5 @@ else
     $(error "A valid value (verilog or vhdl) was not provided for TOPLEVEL_LANG=$(TOPLEVEL_LANG)")
 endif
 
-include $(COCOTB)/makefiles/Makefile.inc
-include $(COCOTB)/makefiles/Makefile.sim
+include $(shell cocotb-config --makefiles)/Makefile.inc
+include $(shell cocotb-config --makefiles)/Makefile.sim
```
eric-wieser: Could halve the number of calls and use:
```suggestion
include $(shell cocotb-config --makefiles)/Makefile.sim
include $(COCOTB_SHARE_DIR)/makefiles/Makefile.sim
```
@themperek, thoughts?

tests/designs/array_module/Makefile
```diff
@@ -54,5 +54,5 @@ else
     $(error "A valid value (verilog or vhdl) was not provided for TOPLEVEL_LANG=$(TOPLEVEL_LANG)")
 endif
 
-include $(COCOTB)/makefiles/Makefile.inc
```
eric-wieser: We can now delete the `COCOTB` definition on line 40, right?

tests/designs/avalon_streaming_module/Makefile
```diff
@@ -20,7 +20,7 @@ COCOTB?=$(WPWD)/../../..
 
 VERILOG_SOURCES = $(COCOTB)/tests/designs/avalon_streaming_module/avalon_streaming.sv
```
eric-wieser: If the goal is to eliminate uses of $(COCOTB), should this be:
```suggestion
VERILOG_SOURCES = $(WPWD)/avalon_streaming.sv
```

tests/designs/array_module/Makefile
themperek: Yes but better to remove `include $(shell cocotb-config --makefiles)/Makefile.inc` than half solutions?

tests/designs/avalon_streaming_module/Makefile
ktbarrett: See https://github.com/cocotb/cocotb/pull/1517#issuecomment-601751008.

eric-wieser: If these are no longer global, shouldn't we change `COCOTB?=` to `COCOTB:=` in those files?
Allowing users to run the makefile from one install to run the tests in a different install (by setting COCOTB) doesn't make any sense.

eric-wieser: Ah, I see with your new commit message it's pretty clear that this change was out of scope, I was still reading the PR title

ktbarrett: >If these are no longer global, shouldn't we change COCOTB?= to COCOTB:= in those files?
Allowing users to run the makefile from one install to run the tests in a different install (by setting COCOTB) doesn't make any sense.
That's a good follow on commit. I can put that in this PR if you want.

eric-wieser: Nah, I think fine to leave it till another PR. I'd still prefer we eliminate it entirely, as the current structure makes renaming the folder really hard. `../../../../path/to/current/folder` is just a fragile way to spell `.`.

Makefile
ktbarrett: It seems that `LD_LIBRARY_PATH` is not correct if we don't have the include here. It works on my machine because I have cocotb installed in editable mode, so there is a build dir.

Fails, pointing locally: https://travis-ci.org/github/cocotb/cocotb/jobs/664893296#L628
Passes, points into install dir: https://travis-ci.org/github/cocotb/cocotb/jobs/664959814#L628

themperek: I will try to look at this. I hope LD_LIBRARY_PATH will go away soon.

tests/designs/array_module/Makefile
eric-wieser: Right, I'd forgotten the goal eventually is to remove the first file entirely.

PR #1518, opened 2020-03-20
Fix doctests in cocotb.binary, and run them as part of the regression test

eric-wieser:

### DIFF
--- a/tests/test_cases/test_cocotb/test_doctests.py
+++ b/tests/test_cases/test_cocotb/test_doctests.py
@@ -1,0 +1,18 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+import doctest
+
+import cocotb
+
+
+@cocotb.test()
+async def test_utils(dut):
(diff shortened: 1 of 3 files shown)


ktbarrett:
This might be overboard, but is there a way to iterate over all the doctests and build each test into a cocotb test?

eric-wieser:
That's definitely possible using the advanced doctest API, but wasn't something I wanted to dive into.

While it would be cool, it would ultimately be a waste of time if we added support for running `unittest.TestSuite` objects, for which doctests already provide hooks.

tests/test_cases/test_cocotb/test_doctests.py
```diff
@@ -0,0 +1,16 @@
+import doctest
```
cmarqu: This should get the SPDX header.

eric-wieser: Can you remind me what that is?

cmarqu: Sorry, yes, this one :)

eric-wieser:
Seems that travis decided not to run on this...

PR #1519, opened 2020-03-20
Deprecate hexdump and hexdiffs on str arguments

eric-wieser:
This function should accept bytes, not str.
Indeed, the upstream `scapy.utils.hexdump` this is derived from already behaves in this way.

Adds some annotations while we're here.

It would make sense to merge #1518 first, since those tests will verify this change.

### DIFF
--- a/cocotb/utils.py
+++ b/cocotb/utils.py
@@ -202,42 +202,59 @@ def unpack(ctypes_obj, string, bytes=None):
 import cocotb.ANSI as ANSI
 
 
-def _sane_color(x):
+# A note on the use of latin1 in the deprecations below:
+# Latin1 is the only encoding `e` that satisfies
+# `all(chr(x).encode(e) == bytes([x]) for x in range(255))`
+# Our use of `ord` and `chr` throughout other bits of code make this the most
+# compatible choice of encoding. Under this convention, old code can be upgraded
+# by changing "binary" strings from `"\x12\x34"` to `b"\x12\x34"`.
+
+
+def _sane_color(x: bytes) -> str:
     r = ""
-    for i in x:
-        j = ord(i)
+    for j in x:
         if (j < 32) or (j >= 127):
             r += "."
         else:
-            r += i
+            r += chr(j)
     return r
 
 
-def hexdump(x):
+def hexdump(x: bytes) -> str:
     """Hexdump a buffer.
 
     Args:
-        x: Object that supports conversion via the ``str`` built-in.
+        x: Object that supports conversion to :class:`bytes`.
 
     Returns:
         A string containing the hexdump.
 
+    .. deprecated:: 1.4
+        Passing a :class:`str` to this function is deprecated, as it
+        is not an appropriate type for binary data. Doing so anyway
+        will encode the string to latin1.
+
     Example:
-        >>> print(hexdump('this somewhat long string'))
+        >>> print(hexdump(b'this somewhat long string'))
         0000   74 68 69 73 20 73 6F 6D 65 77 68 61 74 20 6C 6F   this somewhat lo
         0010   6E 67 20 73 74 72 69 6E 67                        ng string
         <BLANKLINE>
     """
     # adapted from scapy.utils.hexdump
     rs = ""
-    x = str(x)
+    if isinstance(x, str):
+        warnings.warn(
+            "Passing a string to hexdump is deprecated, pass bytes instead",
+            DeprecationWarning, stacklevel=2)
+        x = x.encode('latin1')
+    x = b"%b" % x
     l = len(x)
     i = 0
     while i < l:
         rs += "%04x   " % i
         for j in range(16):
             if i + j < l:
-                rs += "%02X " % ord(x[i + j])
+                rs += "%02X " % x[i + j]
             else:
                 rs += "   "
             if j % 16 == 7:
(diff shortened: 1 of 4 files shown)


cocotb/utils.py
```diff
@@ -202,97 +202,127 @@ def unpack(ctypes_obj, string, bytes=None):
 import cocotb.ANSI as ANSI
 
 
-def _sane_color(x):
+# A note on the use of latin1 in the deprecations below:
+# Latin1 is the only encoding `e` that satisfies
+# `all(chr(x).encode(e) == bytes([x]) for x in range(255))`
+# Our use of `ord` and `chr` throughout other bits of code make this the most
+# compatible choice of encoding. Under this convention, old code can be upgraded
+# by changing "binary" strings from `"\x12\x34"` to `b"\x12\x34"`.
```
cmarqu: Maybe that "upgrade information" would be nice to have in the deprecation warnings.

eric-wieser: This upgrade information isn't particularly applicable to callers of `hexdump`, who will not notice the difference in behavior. Users will be dumping data they got from their simulation, not constants from their python.

cocotb/utils.py
```diff
@@ -202,97 +202,127 @@ def unpack(ctypes_obj, string, bytes=None):
...
         for j in range(16):
             if i + j < l:
-                rs += "%02X " % ord(x[i + j])
+                rs += "%02X " % x[i + j]
             else:
                 rs += "   "
             if j % 16 == 7:
                 rs += ""
         rs += "  "
+        # latin1 is fine, because we conver
```
cmarqu:
```suggestion
        # latin1 is fine, because we convert
```

cocotb/utils.py
```diff
@@ -202,97 +202,127 @@ def unpack(ctypes_obj, string, bytes=None):
...
-def _sane_color(x):
+# A note on the use of latin1 in the deprecations below:
+# Latin1 is the only encoding `e` that satisfies
+# `all(chr(x).encode(e) == bytes([x]) for x in range(255))`
+# Our use of `ord` and `chr` throughout other bits of code make this the most
+# compatible choice of encoding. Under this convention, old code can be upgraded
+# by changing "binary" strings from `"\x12\x34"` to `b"\x12\x34"`.
+
+
+def _sane_color(x: bytes) -> str:
```
cmarqu: We are actually implementing ``sane``, not ``sane_color`` of scapy.utils (https://github.com/secdev/scapy/blob/master/scapy/utils.py#L161), so this is a good opportunity to rename that function IMO.

cocotb/utils.py
```diff
@@ -202,97 +202,127 @@ def unpack(ctypes_obj, string, bytes=None):
...
         rs += "  "
+        # latin1 is fine, because we conver
         rs += _sane_color(x[i:i + 16]) + "\n"
         i += 16
     return rs
 
 
-def hexdiffs(x, y):
-    """Return a diff string showing differences between two binary strings.
+def hexdiffs(x: bytes, y: bytes) -> str:
```
cmarqu: Wondering why this is called ``hexdiffs`` and not ``hexdiff``(as in scapy.utils), and whether we should rename this to ``hexdiff`` and make ``hexdiffs`` an alias (that can be deprecated)?

cmarqu: Not in this PR maybe.

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1466,3 +1466,25 @@ def test_undecorated_coroutine_yield(dut):
     global last_ordered_test
     val, last_ordered_test = last_ordered_test, 1
     assert val == 2
+
+
+@cocotb.test()
+async def test_hexdump_int_illegal(dut):
+    # this used to be legal, but are deliberately no longer
```
cmarqu:
```suggestion
    # this used to be legal, but is deliberately no longer
```

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1466,3 +1466,25 @@ def test_undecorated_coroutine_yield(dut):
...
+        cocotb.utils.hexdump(1)
+    except TypeError:
+        pass
+    else:
+        raise AssertionError("Expected a type error")
+
+
+@cocotb.test()
+async def test_hexdiffs_int_illegal(dut):
+    # this used to be legal, but are deliberately no longer
```
cmarqu:
```suggestion
    # this used to be legal, but is deliberately no longer
```

cocotb/utils.py
eric-wieser: Not in this PR, I agree

cocotb/utils.py
eric-wieser: Duplication removed

eric-wieser:
Comments addressed, thanks

tests/pytest/test_utils.py
```diff
@@ -0,0 +1,38 @@
...
+import pytest
+
+import cocotb.utils
+
+
+class TestHexDump:
+    def test_int_illegal(dut):
+        # this used to be legal, but deliberately is no longer
+        with pytest.raises(TypeError):
+            cocotb.utils.hexdump(1)
```
eric-wieser: Thanks to my delay in responding to suggestions, the tests can now be done in pytest

PR #1520, opened 2020-03-20
Remove unnecessary __await__ call when creating async-based coroutines

ktbarrett:
The call to `__await__()` on coroutines when we created `RunningTask`s offer nothing and prevents introspection.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -89,13 +89,12 @@ class RunningTask(object):
 
         if inspect.iscoroutine(inst):
             self._natively_awaitable = True
-            self._coro = inst.__await__()
         elif inspect.isgenerator(inst):
             self._natively_awaitable = False
-            self._coro = inst
         else:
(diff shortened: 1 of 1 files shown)


cocotb/decorators.py
```diff
@@ -89,13 +89,12 @@ def __init__(self, inst):
 
         if inspect.iscoroutine(inst):
             self._natively_awaitable = True
-            self._coro = inst.__await__()
         elif inspect.isgenerator(inst):
             self._natively_awaitable = False
-            self._coro = inst
```
eric-wieser: In principle we could do the reverse here and make this line
```suggestion
            self._coro = types._GeneratorWrapper(inst)
```
but we're not really supposed to rely on `_GeneratorWrapper`.

Either way, I don't think we need to worry about it.

eric-wieser:
CI passing is convincing enough for me, thanks!

PR #1521, opened 2020-03-20
Make `RunningTask._coro` always a coroutine

eric-wieser:
This makes things marginally simpler, and hopefully also makes introspection a little easier

Follow-up to gh-1520, cc @ktbarrett 

Hopefully CI will be able to tell us if this was the right thing to do.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -30,8 +30,9 @@ import sys
 import time
 import logging
 import functools
-import inspect
+import collections.abc
 import os
+import types
 
 import cocotb
 from cocotb.log import SimLog
@@ -75,7 +76,7 @@ class CoroutineComplete(Exception):
 
 
 class RunningTask(object):
-    """Per instance wrapper around a running generator.
+    """Per instance wrapper around a running coroutine.
 
     Provides the following:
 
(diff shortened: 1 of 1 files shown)


eric-wieser:
Something weird is going on, https://stackoverflow.com/questions/60777119/why-does-types-coroutine-not-produce-coroutines

cocotb/decorators.py
```diff
@@ -85,13 +86,8 @@ class RunningTask(object):
...
     """
-    def __init__(self, inst):
-
-        if inspect.iscoroutine(inst):
-            self._natively_awaitable = True
-        elif inspect.isgenerator(inst):
-            self._natively_awaitable = False
-        else:
+    def __init__(self, inst: collections.abc.Coroutine):
+        if not isinstance(inst, collections.abc.Coroutine):
```
garmin-mjames:
```suggestion
        if not inspect.isawaitable(inst):
```
https://docs.python.org/3.5/library/inspect.html#inspect.isawaitable

eric-wieser: That's not strict enough, because we need the object to have `send`, `throw`, and `close` methods.

eric-wieser:
I'm going to close this, I don't think `types.coroutine` does what I wanted it to do.

cocotb/decorators.py
```diff
@@ -291,6 +287,7 @@ class coroutine(object):
     """
 
     def __init__(self, func):
+        func = types.coroutine(func)
```
garmin-mjames: Does this affect stack tracing at all?
From `types.coroutine` docs:
> If gen_func is a generator function, it will be modified in-place.
> 
> If gen_func is not a generator function, it will be wrapped. If it returns an instance of collections.abc.Generator, the instance will be wrapped in an awaitable proxy object. All other types of objects will be returned as is.

eric-wieser: As best I can tell, the only benefit to this line is that:
* `yield from some_async_function` would work inside cocotb coroutines
  * No one is using `yield from` in cocotb today anyway, so this provides little value. I suppose what it does do is allow users to convert to async functions from the bottom of the stack out and gain a tiny bit of performance over using `yield some_async_function`, but I don't think confusing users with multiple spellings is worth it
* `await some_coro._func(...)` would work inside async functions.
  * This attribute is private.

PR #1523, opened 2020-03-21
Updated core libraries to use async/await

ktbarrett:
Updated core components (things at the root directory of cocotb) to use `async`/`await` syntax over `@cocotb.coroutine` and `yield`. Docstrings and a few comments associated with those functions were also updated to remove traces of the old syntax.

Will update everything in the Python library reference in the "core components" to not reference `coroutine`/`yield`. Updating the docs to remove the old syntax will be a separate commit.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -434,9 +434,9 @@ class test(coroutine, metaclass=_decorator_helper):
             Users are encouraged to use the following idiom instead::
 
                 @cocotb.test()
-                def my_test(dut):
+                async def my_test(dut):
                     try:
-                        yield thing_that_should_fail()
+                        await thing_that_should_fail()
                     except ExceptionIExpect:
                         pass
                     else:
@@ -463,10 +463,11 @@ class test(coroutine, metaclass=_decorator_helper):
             co = coroutine(f)
 
             @functools.wraps(f)
-            def f(*args, **kwargs):
+            async def f(*args, **kwargs):
                 running_co = co(*args, **kwargs)
+
                 try:
-                    res = yield cocotb.triggers.with_timeout(running_co, self.timeout_time, self.timeout_unit)
+                    res = await cocotb.triggers.with_timeout(running_co, self.timeout_time, self.timeout_unit)
                 except cocotb.result.SimTimeoutError:
                     running_co.kill()
                     raise
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -525,8 +525,8 @@ def _create_test(function, name, documentation, mod, *args, **kwargs):
     Returns:
         Decorated test function
     """
-    def _my_test(dut):
-        yield function(dut, *args, **kwargs)
+    async def _my_test(dut):
+        await function(dut, *args, **kwargs)
 
     _my_test.__name__ = name
     _my_test.__doc__ = documentation
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -246,15 +246,14 @@ class Scheduler(object):
         self._write_coro_inst = None
         self._writes_pending = Event()
 
-    @cocotb.decorators.coroutine
-    def _do_writes(self):
+    async def _do_writes(self):
         """ An internal coroutine that performs pending writes """
         while True:
-            yield self._writes_pending.wait()
+            await self._writes_pending.wait()
             if self._mode != Scheduler._MODE_NORMAL:
-                yield self._next_time_step
+                await self._next_time_step
 
-            yield self._read_write
+            await self._read_write
 
             while self._writes:
                 handle, value = self._writes.popitem()
@@ -511,8 +510,7 @@ class Scheduler(object):
         # TODO: we should be able to better keep track of when this needs to
         # be scheduled
         if self._write_coro_inst is None:
-            self._write_coro_inst = self._do_writes()
-            self.schedule(self._write_coro_inst)
+            self._write_coro_inst = self.add(self._do_writes())
 
         self._writes[handle] = value
         self._writes_pending.set()
@@ -573,11 +571,10 @@ class Scheduler(object):
         # each entry always has a unique thread.
         t, = matching_threads
 
-        @cocotb.coroutine
-        def wrapper():
+        async def wrapper():
             # This function runs in the scheduler thread
             try:
-                _outcome = outcomes.Value((yield coro))
+                _outcome = outcomes.Value(await coro)
             except BaseException as e:
                 _outcome = outcomes.Error(e)
             event.outcome = _outcome
@@ -591,7 +588,7 @@ class Scheduler(object):
             event.set()
 
         event = threading.Event()
-        self._pending_coros.append(wrapper())
+        self._pending_coros.append(cocotb.decorators.RunningTask(wrapper()))
         # The scheduler thread blocks in `thread_wait`, and is woken when we
         # call `thread_suspend` - so we need to make sure the coroutine is
         # queued before that.
@@ -602,11 +599,11 @@ class Scheduler(object):
 
     def run_in_executor(self, func, *args, **kwargs):
         """Run the coroutine in a separate execution thread
-        and return a yieldable object for the caller.
+        and return an awaitable object for the caller.
         """
         # Create a thread
         # Create a trigger that is called as a result of the thread finishing
-        # Create an Event object that the caller can yield on
+        # Create an Event object that the caller can await on
         # Event object set when the thread finishes execution, this blocks the
         #   calling coroutine (but not the thread) until the external completes
 
@@ -616,8 +613,7 @@ class Scheduler(object):
                 self.log.debug("Execution of external routine done %s" % threading.current_thread())
             _waiter.thread_done()
 
-        @cocotb.coroutine
-        def wrapper():
+        async def wrapper():
             waiter = external_waiter()
             thread = threading.Thread(group=None, target=execute_external,
                                       name=func.__name__ + "_thread",
@@ -626,7 +622,7 @@ class Scheduler(object):
             waiter.thread = thread
             self._pending_threads.append(waiter)
 
-            yield waiter.event.wait()
+            await waiter.event.wait()
 
             return waiter.result  # raises if there was an exception
 
@@ -697,7 +693,7 @@ class Scheduler(object):
         return result.join()
 
     def _trigger_from_waitable(self, result: cocotb.triggers.Waitable) -> Trigger:
-        return self._trigger_from_unstarted_coro(result._wait())
+        return self._trigger_from_unstarted_coro(cocotb.decorators.RunningTask(result._wait()))
 
     def _trigger_from_list(self, result: list) -> Trigger:
         return self._trigger_from_waitable(cocotb.triggers.First(*result))
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -25,10 +25,11 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-"""A collections of triggers which a testbench can yield."""
+"""A collections of triggers which a testbench can await."""
 
 import os
 import abc
+from collections.abc import Awaitable
 
 if "COCOTB_SIM" in os.environ:
     import simulator
@@ -40,7 +41,6 @@ from cocotb.utils import (
     get_sim_steps, get_time_from_sim_steps, ParametrizedSingleton,
     lazy_property, remove_traceback_frames,
 )
-from cocotb import decorators
 from cocotb import outcomes
 import cocotb
 
@@ -59,7 +59,7 @@ def _pointer_str(obj):
 class TriggerException(Exception):
     pass
 
-class Trigger(metaclass=abc.ABCMeta):
+class Trigger(Awaitable):
     """Base class to derive from."""
 
     # __dict__ is needed here for the `.log` lazy_property below to work.
@@ -115,7 +115,7 @@ class Trigger(metaclass=abc.ABCMeta):
 
     @property
     def _outcome(self):
-        """The result that `yield this_trigger` produces in a coroutine.
+        """The result that `await this_trigger` produces in a coroutine.
 
         The default is to produce the trigger itself, which is done for
         ease of use with :class:`~cocotb.triggers.First`.
@@ -173,24 +173,24 @@ class Timer(GPITrigger):
 
         Examples:
 
-            >>> yield Timer(100, units='ps')
+            >>> await Timer(100, units='ps')
 
             The time can also be a ``float``:
 
-            >>> yield Timer(100e-9, units='sec')
+            >>> await Timer(100e-9, units='sec')
 
             which is particularly convenient when working with frequencies:
 
             >>> freq = 10e6  # 10 MHz
-            >>> yield Timer(1 / freq, units='sec')
+            >>> await Timer(1 / freq, units='sec')
 
             Other builtin exact numeric types can be used too:
 
             >>> from fractions import Fraction
-            >>> yield Timer(Fraction(1, 10), units='ns')
+            >>> await Timer(Fraction(1, 10), units='ns')
 
             >>> from decimal import Decimal
-            >>> yield Timer(Decimal('100e-9'), units='sec')
+            >>> await Timer(Decimal('100e-9'), units='sec')
 
             These are most useful when using computed durations while
             avoiding floating point inaccuracies.
@@ -376,7 +376,7 @@ class _Event(PythonTrigger):
 class Event(object):
     """Event to permit synchronization between two coroutines.
 
-    Yielding :meth:`wait()` from one coroutine will block the coroutine until
+    Awaiting :meth:`wait()` from one coroutine will block the coroutine until
     :meth:`set()` is called somewhere else.
     """
 
@@ -458,7 +458,7 @@ class Lock(object):
 
     This should be used as::
 
-        yield lock.acquire()
+        await lock.acquire()
         try:
             # do some stuff
         finally:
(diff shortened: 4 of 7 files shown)


cocotb/decorators.py
```diff
@@ -460,18 +461,12 @@ def __init__(self, f, timeout_time=None, timeout_unit=None,
...
+            async def f(*args, **kwargs):
                 running_co = co(*args, **kwargs)
-                try:
-                    res = yield cocotb.triggers.with_timeout(running_co, self.timeout_time, self.timeout_unit)
-                except cocotb.result.SimTimeoutError:
-                    running_co.kill()
-                    raise
-                else:
-                    return res
+                return await cocotb.triggers.with_timeout(running_co, self.timeout_time, self.timeout_unit)
```
ktbarrett: This change was possible because `running_co` never touches the scheduler until it is `fork`ed by `First._wait`, which cleans up after the first trigger fires. This change was necessary to get the above `types.coroutine` line to work.

cocotb/triggers.py
```diff
@@ -656,12 +641,11 @@ def __init__(self, *triggers):
         self.triggers = tuple(triggers)
 
         # Do some basic type-checking up front, rather than waiting until we
-        # yield them.
-        allowed_types = (Trigger, Waitable, decorators.RunningTask)
+        # await them.
         for trigger in self.triggers:
-            if not isinstance(trigger, allowed_types):
+            if not inspect.isawaitable(trigger):
```
ktbarrett: This simplification allows the use of `types.coroutine` and other awaitable objects. We don't need to be so strict about types anymore because we bypass the scheduler.

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1015,7 +1015,8 @@ def raise_soon():
       File ".*test_cocotb\.py", line \d+, in raise_soon
         yield cocotb\.triggers\.First\(raise_inner\(\)\)
       File ".*triggers\.py", line \d+, in _wait
-        return \(yield first_trigger\)[^\n]*
+        return \(await first_trigger\)[^\n]*
+    .*
```
ktbarrett: I feel like there is a better way, but I don't know it.

cocotb/scheduler.py
```diff
@@ -573,11 +571,10 @@ def queue_function(self, coro):
         # each entry always has a unique thread.
         t, = matching_threads
 
-        @cocotb.coroutine
-        def wrapper():
+        async def wrapper():
             # This function runs in the scheduler thread
             try:
-                _outcome = outcomes.Value((yield coro))
+                _outcome = outcomes.Value((await coro))
```
eric-wieser:
```suggestion
                _outcome = outcomes.Value(await coro)
```
Await has a more convenient place in the grammar than yield, which means the parens aren't requires.

cocotb/triggers.py
```diff
@@ -25,10 +25,11 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-"""A collections of triggers which a testbench can yield."""
+"""A collections of triggers which a testbench can be awaited."""
```
```suggestion
"""A collections of triggers which a testbench can await."""
```

cocotb/triggers.py
```diff
@@ -631,16 +617,15 @@ class Waitable(object):
...
-    @decorators.coroutine
-    def _wait(self):
+
+    async def _wait(self):
         """
-        Should be implemented by the sub-class. Called by `yield self` to
+        Should be implemented by the sub-class. Called by `await self` to
         convert the waitable object into a coroutine.
 
         ReturnValue can be used here.
```
eric-wieser: This line can go too.

cocotb/triggers.py
```diff
@@ -765,7 +742,7 @@ def on_done(ret):
         #  - Using `NullTrigger` here instead of `result = completed[0].get()`
         #    means we avoid inserting an `outcome.get` frame in the traceback
         first_trigger = NullTrigger(outcome=completed[0])
-        return (yield first_trigger)  # the first of multiple triggers that fired
+        return (await first_trigger)  # the first of multiple triggers that fired
```
```suggestion
        return await first_trigger  # the first of multiple triggers that fired
```

tests/test_cases/test_cocotb/test_cocotb.py
eric-wieser: What does this traceback look like now?

`[\s\S]*` would also work here, and avoid the need for "dot all"

cocotb/triggers.py
```diff
@@ -831,9 +805,9 @@ def with_timeout(trigger, timeout_time, timeout_unit=None):
...
-    if res is timeout_timer:
-        raise cocotb.result.SimTimeoutError
-    else:
-        return res
+
+    async def timeout():
+        await Timer(timeout_time, timeout_unit)
+        raise cocotb.result.SimTimeoutError()
+
+    return await First(timeout(), trigger)
```
eric-wieser: Any reason for the rewrite here?

cocotb/triggers.py
```diff
@@ -673,13 +657,12 @@ def __repr__(self):
...
-@decorators.coroutine
-def _wait_callback(trigger, callback):
+async def _wait_callback(trigger, callback):
     """
-    Wait for a trigger, and call `callback` with the outcome of the yield.
+    Wait for a trigger, and call `callback` with the outcome of the await.
     """
     try:
-        ret = outcomes.Value((yield trigger))
+        ret = outcomes.Value((await trigger))
```
```suggestion
        ret = outcomes.Value(await trigger)
```

cocotb/decorators.py
eric-wieser: I'm a bit confused by the fact that you no longer kill the coroutine here.

cocotb/triggers.py
```diff
@@ -751,7 +728,7 @@ def on_done(ret):
             waiters.append(cocotb.fork(_wait_callback(t, on_done)))
 
         # wait for a waiter to complete
-        yield e.wait()
+        await e.wait()
 
         # kill all the other waiters
         # TODO: Should this kill the coroutines behind any Join triggers?
```
eric-wieser: Has this behavior changed?

cocotb/triggers.py
eric-wieser: Or more accurately, is no longer correct

eric-wieser:
It would be interesting to instrument the scheduler and see how much the number of `Join` triggers reduces with this patch.

cocotb/triggers.py
ktbarrett: Do you mean the TODO line, the "kill all the waiters" line, or the change?

tests/test_cases/test_cocotb/test_cocotb.py
ktbarrett: There are potentially many lines there chaining through `__await__`s there due to `First` using `await` on `raise_inner()`.
```
  File "/home/kaleb/dev/cocotb/tests/test_cases/test_cocotb/test_cocotb.py", line 922, in _check_traceback
    yield running_coro
  File "/home/kaleb/dev/cocotb/tests/test_cases/test_cocotb/test_cocotb.py", line 1007, in raise_soon
    yield cocotb.triggers.First(raise_inner())
  File "/home/kaleb/dev/cocotb/cocotb/triggers.py", line 745, in _wait
    return await first_trigger  # the first of multiple triggers that fired
  File "/home/kaleb/dev/cocotb/cocotb/triggers.py", line 128, in __await__
    return (yield self)
  File "/home/kaleb/dev/cocotb/cocotb/decorators.py", line 204, in __await__
    return (yield self)
  File "/home/kaleb/dev/cocotb/tests/test_cases/test_cocotb/test_cocotb.py", line 1002, in raise_inner
    raise ValueError('It is soon now')
ValueError: It is soon now
```

cocotb/triggers.py
ktbarrett: In the previous implementation, we used `yield []` which returns the trigger than fires. Now we have to use `First` which returns the result of the first trigger that fires. This makes it hard to distinguish which trigger fired. The implementation was going to have to change, so I went with an implementation that tests new functionality.

cocotb/decorators.py
ktbarrett: It's no longer necessary thanks to the improved semantics offered by the `with_timeout` rewrite. `running_co` doesn't get scheduled, and doesn't need to be killed.

ktbarrett:
Travis and these Mac builds lately...

cocotb/triggers.py
eric-wieser: `yield [a, b]` is identical to `yield First(a, b)`, which is almost identical to `await First(a, b)`.

Both return the _result_ of the first trigger that fires, not the trigger itself.

However, it happens that the result of a `Timer` _is_ the trigger itself.

cocotb/decorators.py
eric-wieser: Here's the real answer - you changed the semantics of `First` in this patch to kill non-first coroutines. I don't think we want that change.

cocotb/triggers.py
```diff
@@ -673,13 +654,12 @@ def __repr__(self):
...
-@decorators.coroutine
-def _wait_callback(trigger, callback):
+async def _wait_callback(trigger, callback):
     """
-    Wait for a trigger, and call `callback` with the outcome of the yield.
+    Wait for a trigger, and call `callback` with the outcome of the await.
     """
     try:
-        ret = outcomes.Value((yield trigger))
+        ret = outcomes.Value(await trigger)
```
eric-wieser: This is a behavior change, and means that `fork(_wait_callback()).kill()` now throws an exception inside the `trigger` itself if `trigger` is an async coroutine and not a `RunningTask`.

I'd prefer we not change that behavior in this patch. Leaving this function untouched would restore the old behavior, and we can come back to it after more though.

ktbarrett: This change is fine, it's the change to `AggregateWaitable` that allows users to pass any Awaitable.

cocotb/triggers.py
ktbarrett: Returned to original implementation.

cocotb/triggers.py
ktbarrett: This changes the behavior of `First` and `Combine`, this was reverted.

cocotb/triggers.py
ktbarrett: Change was reverted.

eric-wieser: Unfortunately this still changes the semantics of cancellation inside `Waitable` objects.

The thing to realize is that `yield trigger` translates to `await scheduler._trigger_from_any(trigger)`.

If you used that here, then everything would work as before, and you'd still be able to pass async coroutines directly to `First`.

I'm ok with us calling a private function here as long as we add a todo comment about moving it somewhere more sensible.

tests/test_cases/test_cocotb/test_cocotb.py
eric-wieser: I'd prefer you list those explicitly, because it's ugliness in the traceback we may want to try to remove in future.
```suggestion
      File ".*triggers.py", line \d+, in __await__
        return \(yield self\)
      File ".*decorators.py", line \d+, in __await__
        return \(yield self\)
```

cocotb/triggers.py
```diff
@@ -631,16 +615,13 @@ class Waitable(object):
     supporting Python versions (>=3.3).
```
garmin-mjames: You can remove this comment in the docstring since cocotb supports 3.5+ now.

cocotb/triggers.py
ktbarrett: Ahh, forgot about Waitables. Advise what you think the proper solution is.

eric-wieser: I recommend you:

1. Put back your simplification in `AggregateWaitable`
2. Pick between:
   * Not converting this routine to use `async`, leave it using `cocotb.coroutine`
   * Using `trigger = scheduler._trigger_from_any(trigger)` above line +662.

I'd lean towards you just picking the second option in 2, since that makes the diff smaller.

cocotb/triggers.py
eric-wieser: If we're ok with scope creep, we can also add `collections.abc.Awaitable` as a base class.

cocotb/triggers.py
ktbarrett: So if you pass an async coroutine, it would schedule it and it would be on the user to know to kill it after the fact? Seems like we should leave the simplification off until we commit to changing the semantics as that seems antagonistic.

cocotb/triggers.py
ktbarrett: Do we really need Waitable anymore? They could easily be rewritten as coroutines and we could remove that special case from the scheduler.

eric-wieser: The purpose of `Waitable` is to be reusable coroutines - you add them arguments just once, and to can then await them repeatedly.

Essentially, the idea is they behave like triggers, but we get to implement them using coroutines.

cocotb/triggers.py
eric-wieser: I believe that already happens today if you pass an async function.

eric-wieser: I'll add some tests for the quirks of the current behavior in a new PR.

ktbarrett: You currently can't pass it an async coroutine at all. There is a type guard that matches on `RunningTask`, `Waitable`, and `Trigger`.

cocotb/triggers.py
ktbarrett: Added in `Awaitable` base classes to `Trigger` and `Waitable`.

cocotb/triggers.py
eric-wieser: Good catch. However, it's still what happens today if you pass a `cocotb.coroutine` in.

ktbarrett: You can kill those however, if you pass in an un-decorated async coroutine how do kill it after `First` fires?

eric-wieser: Ah, I see. I was thinking about situations like:
```python
await First(some_coroutine(), some_trigger)
```
where the caller has already thrown away their hope of killing the coroutine. What you're describing I think is
```python
some_task = some_coroutine()
await First(some_task, some_trigger)
# user can call some_task.kill()
```

Perhaps it is best to play it safe and continue to forbid passing async coroutines to `First` until a later patch, as you've done.

cocotb/triggers.py
```diff
@@ -721,10 +701,6 @@ class First(_AggregateWaitable):
 
     Returns the result of the trigger that fired.
 
-    As a shorthand, ``t = yield [a, b]`` can be used instead of
-    ``t = yield First(a, b)``. Note that this shorthand is not available when
-    using :keyword:`await`.
```
eric-wieser: I think we need to keep this to help explain to existing users that `First` is just another name for the `yield [...]` syntax they're used to.

However, we could move it down to the bottom of the docstring, and remove the word "shorthand" to indicate that it isn't worth switching to. Perhaps:
```
.. note::
    When using the old-style :keyword:`yield`-based coroutines, ``t = yield [a, b]`` was another spelling
    of ``t = yield First(a, b)``. This spelling is no longer available when using :keyword:`await`-based
    coroutines.

File: cocotb/triggers.py [hunk 14]
```diff
@@ -623,24 +608,20 @@ def __repr__(self):
         return "{}({!r})".format(type(self).__name__, self._coroutine)
 
 
-class Waitable(object):
+class Waitable(Awaitable):
```
eric-wieser: Just a note - this doesn't actually have any effect other than making `issubclass(Waitable, Awaitable)` faster. It's good documentation though, so I'm not opposed.

cocotb/triggers.py
```diff
@@ -623,24 +608,20 @@ def __repr__(self):
...
     """
     Compatibility layer that emulates `collections.abc.Awaitable`.
```
```suggestion
    Base class for trigger-like objects implemented using coroutines.
```
There's not really any compatibility going on here at all.

We may eventually want to remove this in place of using `Awaitable` directly, but there lie the dragons of distinguishing async coroutines from triggers / running tasks, which for now we should leave sleeping.

cocotb/triggers.py
```diff
@@ -673,13 +654,13 @@ def __repr__(self):
...
-def _wait_callback(trigger, callback):
+async def _wait_callback(trigger, callback):
     """
-    Wait for a trigger, and call `callback` with the outcome of the yield.
+    Wait for a trigger, and call `callback` with the outcome of the await.
     """
+    trigger = cocotb.scheduler._trigger_from_any(trigger)
     try:
-        ret = outcomes.Value((yield trigger))
+        ret = outcomes.Value(await trigger)
```
eric-wieser: Regarding my earlier remark about the behavior change here on `Waitable` objects - I think since we only really have three `Waitable` types in the wild, and none of them leave much behind to clean up, this is probably ok.

cocotb/triggers.py
ktbarrett: I'd like to use the first case, but I feel like `First` should do the scheduling *and* the killing in the case of passing an undecorated async coroutine, and if you needed to have it continue to run after `First`, you would wrap it in a `fork`, or perhaps something like `asyncio.shield`?

cocotb/regression.py
```diff
@@ -525,8 +526,8 @@ def _create_test(function, name, documentation, mod, *args, **kwargs):
     Returns:
         Decorated test function
     """
-    def _my_test(dut):
-        yield function(dut, *args, **kwargs)
+    async def _my_test(dut):
+        await types.coroutine(function)(dut, *args, **kwargs)
```
eric-wieser: This seems pointless due to line 588, where we already know that `function` is a `cocotb.coroutine` and will therefore produce `RunningTask` instances which are already awaitable.

ktbarrett: It might even do the wrong thing, will change.

eric-wieser:
Following up on https://github.com/cocotb/cocotb/pull/1523#discussion_r396025730

> I'll add some tests for the quirks of the current behavior in a new PR.

Done in #1528. Let's try and get that in first, so we can run that test against this PR.

ktbarrett:
@eric-wieser I rebased this PR on top of yours and it passes. Do you want me to put that commit in with this PR?

PR #1526, opened 2020-03-21
Add fork-like interface that propagates exceptions

ktbarrett:
## Issue

`fork` is the primary mechanism to run coroutines concurrently in cocotb, however, it has a major flaw. When `fork`ing a coroutine, if a `Join` trigger is not created or the forked coroutine is not `await`ed nor `yield`ed (which creates a `Join`) *before* the coroutine finishes with an exception, the test is killed.

xref #922.

For example, the following causes a test end rather than throwing `NameError` when the forked coroutine is awaited on.

```python
async def name_error():
    await does_not_exist()
    #  ... other stuff

task = cocotb.fork(name_error())
await Timer(100)
await task  # should throw NameError here, but the test dies before it gets here
```

But this works.

```python
async def name_error():
    await Timer(100)
    await does_not_exist()
    #  ... other stuff

task = cocotb.fork(name_error())
await Timer(10)
await task  # throws NameError here because we start waiting *before* name_error throws
```

While the above behavior is bad and unexpected (and depending upon the code, non-deterministic!), the test-killing behavior is leveraged in many tests and is useful. For instance, monitors use this to end tests when the monitoring "thread" fails, which might otherwise allow the test to run in an invalid state.

## Proposal
* change the scheduler to not kill tests in these cases and always hold exceptions until awaited on
* add a new interface that uses the newer cleaner semantics
* re-implement fork to obtain the legacy behavior using the new interface

## Alternative Proposals

2. Just change fork to never kill the test
   * users will have to update tests
   * the behavior is deterministic
   * should add warnings for un-joined coroutines much like asyncio does with un-awaited tasks
3. Add new interface and leave fork, but make fork *always* kill the test
   * users will have to update tests
   * the behavior is deterministic
   * still useful for many cases

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -106,8 +106,88 @@ regression_manager = None
 plusargs = {}
 """A dictionary of "plusargs" handed to the simulation."""
 
-# To save typing provide an alias to scheduler.add
-fork = scheduler.add
+
+def run_task(coro):
+    """
+    Run a coroutine concurrently.
+
+    The result can be awaited at any point to block until a result is available.
+
+    .. versionadded:: 1.4
+    """
+    return scheduler.add(coro)
+
+
+def fire_and_forget(coro):
+    """
+    Run a coroutine concurrently that is never meant to be awaited.
+
+    This will end the test if an exception occurs.
+
+    .. versionadded:: 1.4
+    """
+    co = cocotb.decorators.make_task(coro)
+
+    class FireAndForgetCoroutine(cocotb.decorators.RunningTask):
+
+        def __init__(self, co):
+
+            async def watcher():
+                return await co
+
+            super().__init__(watcher())
+
+        def _advance(self, outcome):
+            try:
+                return super()._advance(outcome)
+            except cocotb.decorators.CoroutineComplete:
+                if isinstance(self._outcome, cocotb.outcomes.Error):
+                    cocotb.utils.remove_traceback_frames(self._outcome.error, ['watcher', '__await__'])
+                    self.log.warning("Ending test; fire-and-forget coroutine resulted in an exception.")
+                    scheduler.finish_test(self._outcome.error)
+                else:
+                    self.log.info("Fire-and-forget coroutine resulted in {}.".format(self._outcome))
+                raise
+
+    return run_task(FireAndForgetCoroutine(co))
+
+
+def fork(coro):
+    """
+    Runs a coroutine concurrently. Will end the current test *if* the coroutine results in an exception *before* being
+    joined.
+
+    Preserved for legacy, use :func:`~cocotb.run_task` or :func:`~cocotb.fire_and_forget` instead.
+    """
+    co = cocotb.decorators.make_task(coro)
+
+    class ForkedCoroutine(cocotb.decorators.RunningTask):
+
+        def __init__(self, co):
+
+            async def watcher():
+                return await co
+
+            super().__init__(watcher())
+
+        def _advance(self, outcome):
+            try:
+                return super()._advance(outcome)
+            except cocotb.decorators.CoroutineComplete:
+                if isinstance(self._outcome, cocotb.outcomes.Error):
+                    cocotb.utils.remove_traceback_frames(self._outcome.error, ['watcher', '__await__'])
+                    if cocotb.triggers.Join(self) not in scheduler._trigger2coros:
+                        self.log.warning(
+                            "If you intend to run a task that should end the test if it encounters an error,"
+                            " prefer cocotb.fire_and_forget")
+                        scheduler.finish_test(self._outcome.error)
+                    else:
+                        self.log.warning(
+                            "If you intend to run a task and retrieve the outcome later, perfer cocotb.run_task")
+                raise
+
+    return run_task(ForkedCoroutine(co))
+
 
 # FIXME is this really required?
 _rlock = threading.RLock()
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -490,3 +490,39 @@ class test(coroutine, metaclass=_decorator_helper):
 
     def __call__(self, *args, **kwargs):
         return RunningTest(self._func(*args, **kwargs), self)
+
+
+def make_task(coro):
+    """
+    Turn the given object into a task.
+
+    Perform type checking, convert compatible types into :class:`~cocotb.decorators.RunningTask`,
+    and report common gotchas.
+
+    .. versionadded:: 1.4
+    """
+
+    if isinstance(coro, RunningTask):
+        return coro
+
+    if inspect.iscoroutine(coro):
+        return RunningTask(coro)
+
+    if isinstance(coro, coroutine):
+        raise TypeError(
+            "Attempt to convert a cocotb.coroutine object ({}) into a task.\n"
+            "Did you forget to call the cocotb.coroutine-decorated function?"
+            .format(coro)
+        )
+
+    if sys.version_info >= (3, 6) and inspect.isasyncgen(coro):
+        raise TypeError(
+            "{} is an async generator, not a coroutine. "
+            "You likely used the yield keyword instead of await.".format(
+                coro.__qualname__))
+
+    raise TypeError(
+        "Attempt to convert an incompatible object of type {} into a task: {!r}\n"
+        "Did you forget to use the @cocotb.coroutine decorator, or async/await?"
+        .format(type(coro), coro)
+    )
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -638,30 +638,7 @@ class Scheduler(object):
         Just a wrapper around self.schedule which provides some debug and
         useful error messages in the event of common gotchas.
         """
-        if isinstance(coroutine, cocotb.decorators.coroutine):
-            raise TypeError(
-                "Attempt to schedule a coroutine that hasn't started: {}.\n"
-                "Did you forget to add parentheses to the @cocotb.test() "
-                "decorator?"
-                .format(coroutine)
-            )
-
-        if inspect.iscoroutine(coroutine):
-            return self.add(cocotb.decorators.RunningTask(coroutine))
-
-        elif sys.version_info >= (3, 6) and inspect.isasyncgen(coroutine):
-            raise TypeError(
-                "{} is an async generator, not a coroutine. "
-                "You likely used the yield keyword instead of await.".format(
-                    coroutine.__qualname__))
-
-        elif not isinstance(coroutine, cocotb.decorators.RunningTask):
-            raise TypeError(
-                "Attempt to add a object of type {} to the scheduler, which "
-                "isn't a coroutine: {!r}\n"
-                "Did you forget to use the @cocotb.coroutine decorator?"
-                .format(type(coroutine), coroutine)
-            )
+        coroutine = cocotb.decorators.make_task(coroutine)
 
         if _debug:
             self.log.debug("Adding new coroutine %s" % coroutine.__name__)
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -61,8 +61,6 @@ import cocotb.decorators
 from cocotb.triggers import (Trigger, GPITrigger, Timer, ReadOnly,
                              NextTimeStep, ReadWrite, Event, Join, NullTrigger)
 from cocotb.log import SimLog
-from cocotb.result import TestComplete
-from cocotb.utils import remove_traceback_frames
 from cocotb import _py_compat
 
 
@@ -461,7 +459,6 @@ class Scheduler(object):
                 self.log.debug("All coroutines scheduled, handing control back"
                                " to simulator")
 
-
     def unschedule(self, coro):
         """Unschedule a coroutine.  Unprime any pending triggers"""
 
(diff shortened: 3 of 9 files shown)


cocotb/__init__.py
```diff
@@ -106,8 +107,36 @@ def _reopen_stream_with_buffering(stream_name):
 plusargs = {}
 """A dictionary of "plusargs" handed to the simulation."""
 
-# To save typing provide an alias to scheduler.add
-fork = scheduler.add
+
+def current_test() -> cocotb.decorators.RunningTest:
+    """Returns a reference to the currently running test"""
+    global scheduler
```
eric-wieser:
```suggestion
```

`global x` is for changing the meaning of `x = ...`. You don't assign to `scheduler` here, so `global` isn't doing anything other than making me look really hard for a `scheduler = ...` line that doesn't exist. Same below.

cocotb/__init__.py
```diff
@@ -106,8 +107,36 @@ def _reopen_stream_with_buffering(stream_name):
 plusargs = {}
 """A dictionary of "plusargs" handed to the simulation."""
 
-# To save typing provide an alias to scheduler.add
-fork = scheduler.add
+
+def current_test() -> cocotb.decorators.RunningTest:
```
eric-wieser: My hunch is that we don't want to make this public until we have a good use case.

cocotb/__init__.py
```diff
@@ -106,8 +107,36 @@ def _reopen_stream_with_buffering(stream_name):
...
+    async def watcher():
+        try:
+            await co
+        except Exception as e:
+            if cocotb.triggers.Join(co) not in scheduler._trigger2coros:
+                current_test().abort(e)
+
+    create_task(watcher())
+
+    return co
```
eric-wieser: At first glance, this looks quite tidy - thanks!

cocotb/__init__.py
```diff
@@ -106,8 +107,36 @@ def _reopen_stream_with_buffering(stream_name):
...
+    global scheduler
+
+    co = create_task(coro)
+
+    async def watcher():
+        try:
+            await co
+        except Exception as e:
+            if cocotb.triggers.Join(co) not in scheduler._trigger2coros:
+                current_test().abort(e)
```
eric-wieser: Does it make sense to warn the user that just because it worked for them once doesn't mean it will always work, and to inform them they may be looking for `create_task`?

Message lousy, but hopefully enough to give you the idea.
```suggestion
            if cocotb.triggers.Join(co) not in scheduler._trigger2coros:
                current_test().abort(e)
            else:
                warnings.warn("If you care about the result of fork, you might be after create_task instead")
```

cocotb/__init__.py
ktbarrett: But, it is also incorrect. You might have missed my commit message.

https://github.com/cocotb/cocotb/blob/d1a305da30fa0ffdd0174358657fa94ff778bcb6/cocotb/scheduler.py#L401-L404

I can't check for the `Join` trigger because the first thing it does before servicing all the waiting coroutines is pop the trigger. I'm not sure an easy way around this... I feel like instead of popping the trigger to prevent a potential infinite recursion, the entire `triggers2coros` object should be immutably copied and cleared for the next "wave" of schedulings.

Any help would be appreciated.

cocotb/__init__.py
ktbarrett: The only use case I can think of is aborting the test, assuming we eventually move to test just being a `RunningTask`. Will remove.

cocotb/__init__.py
ktbarrett: Newest commit "fixes" this issue. But there are still yet more issues which I guess hasn't been seen before? When there are multiple coroutines (B and C) waiting on coroutine A, if coroutine A ends with an exception, it will rethrow in one coroutine (B) **and add B's stacktrace to the stacktrace object**. So when coroutine C handles the exception, the stacktrace will include B's stack...

ktbarrett: Well I used `remove_traceback_frame` and we are good, but this is still an outstanding issue in general.

cocotb/__init__.py
```diff
@@ -106,8 +107,38 @@ def _reopen_stream_with_buffering(stream_name):
...
+    co = create_task(coro)
+
+    async def watcher():
+        try:
+            await co
+        except Exception as e:
+            if len(scheduler._scheduling) == 1:  # if we are the only thing waiting on the join
+                scheduler.finish_test(e)
+            else:
+                cocotb.utils.remove_traceback_frames(e, ['watcher', '__await__'])
```
eric-wieser: It's pretty annoying that you have to do this. Can you construct an equivalent case in `asyncio` or `trio`?

I'm remembering now that `trio.outcome`s had an  odd feature that prevented them being "sent" to more than one place. Perhaps this was the motivation.

ktbarrett: These are supposed to be asymmetric coroutines, being able to have multiple joins on the same coroutine is not kosher.

I'll work around this another way.

ktbarrett: I need a way to obtain a reference to a `Join` on the watcher coroutine to check within the watcher if there is a join on it. I think that means I'll have to make subclass of `RunningTask` specifically for `fork`.

```python
def fork(coro: Union[Coroutine, cocotb.decorators.RunningTask]) -> cocotb.decorators.RunningTask:
    """
    Runs a coroutine concurrently. Will end the current test if the coroutine results in an exception *before* being awaited.

    Preserved for legacy.
    """
    co = create_task(coro)

    async def watcher():
        try:
            return await co
        except Exception as e:
            cocotb.utils.remove_traceback_frames(e, ['watcher', '__await__'])
            if cocotb.triggers.Join(self) in scheduler._trigger2coros:
                warnings.warn(
                    "If you want to run a coroutine concurrently and want it to kill the test if there is an error, "
                    "prefer cocotb.fire_and_forget().")
                scheduler.finish_test(e)
            raise

    return create_task(watcher())
```

ktbarrett: We have no way to react to a `kill` here either... If the watcher is `kill`ed the coroutine it watches should be as well, so that's even more of a case for a a `RunningTask` subclass. We need to get #1349 in... That's a pretty critical feature IMO.

cocotb/__init__.py
ktbarrett: The 3.8 regression won't like that. Should we put a log there instead?

cocotb/__init__.py
ktbarrett: This has been addressed.

garmin-mjames:
Maybe something like:
```python3
def run(coro, end_on_exc=False):
    if end_on_exc:
        # fire and forget
    else:
        # run
```

cmarqu:
> Should it be in the docstrings, or in the "Coroutines" section of the documentation? `fork` seems to rely on that section rather than the docstring.

Better don't take the doc style for ``fork`` as an example for anything, we just needed to have something to link to a while ago.

cocotb/__init__.py
```diff
@@ -106,8 +106,67 @@ def _reopen_stream_with_buffering(stream_name):
 plusargs = {}
 """A dictionary of "plusargs" handed to the simulation."""
 
-# To save typing provide an alias to scheduler.add
-fork = scheduler.add
+
+def run_task(coro):
+    """
+    Runs a coroutine concurrently. Can be awaited at any point to block until a result is available.
```
cmarqu:
```suggestion
    Run a coroutine concurrently.
    
    This can be awaited at any point to block until a result is available.
```

cocotb/__init__.py
```diff
@@ -106,8 +106,67 @@ def _reopen_stream_with_buffering(stream_name):
...
+
+    .. versionadded:: 1.4
+    """
+    return scheduler.add(coro)
+
+
+def fire_and_forget(coro):
+    """
+    Runs a coroutine concurrently that is never meant to be awaited. Will end the test if an exception occurs.
```
cmarqu:
```suggestion
    Run a coroutine concurrently that is never meant to be awaited. 
    
    This will end the test if an exception occurs.
```

cocotb/__init__.py
```diff
@@ -106,8 +106,67 @@ def _reopen_stream_with_buffering(stream_name):
...
+    .. versionadded:: 1.4
+    """
+    co = cocotb.decorators.make_task(coro)
+
+    async def watcher():
+        try:
+            res = await co
+        except Exception as e:
+            cocotb.utils.remove_traceback_frames(e, ['watcher', '__await__'])
+            log.warning("Ending test, fire and forget coroutine {} resulted in an exception.".format(coro))
```
cmarqu:
```suggestion
            log.warning("Ending test; fire-and-forget coroutine {} resulted in an exception.".format(coro))
```
(insert "not a native speaker" disclaimer, you decide)

cocotb/__init__.py
```diff
@@ -106,8 +106,67 @@ def _reopen_stream_with_buffering(stream_name):
...
+            scheduler.finish_test(e)
+        else:
+            log.info("Fire and forget coroutine {} resulted in {}.".format(coro, res))
+
+    return run_task(watcher())
+
+
+def fork(coro):
+    """
+    Runs a coroutine concurrently. Will end the current test *if* the coroutine results in an exception *before* being
```
cmarqu:
```suggestion
    Run a coroutine concurrently. 
    
    This will end the current test *if* the coroutine results in an exception *before* being
```

cocotb/decorators.py
```diff
@@ -490,3 +490,41 @@ def f(*args, **kwargs):
...
     def __call__(self, *args, **kwargs):
         return RunningTest(self._func(*args, **kwargs), self)
+
+
+def make_task(coro):
+    """
+    Turns the given object into a task.
+
+    Performs type checking, converts compatible types into :class:`~cocotb.decorators.RunningTask`,
+    and reports common gotchas.
```
cmarqu:
```suggestion
    Turn the given object into a task.

    Perform type checking, convert compatible types into :class:`~cocotb.decorators.RunningTask`,
    and report common gotchas.
```

cocotb/decorators.py
```diff
@@ -490,3 +490,41 @@ def f(*args, **kwargs):
...
+        )
+
+    if sys.version_info >= (3, 6) and inspect.isasyncgen(coro):
+        raise TypeError(
+            "{} is an async generator, not a coroutine. "
+            "You likely used the yield keyword instead of await.".format(
+                coro.__qualname__))
+
+    raise TypeError(
+        "Attempt to add a object of type {} to the scheduler, which "
```
cmarqu:
```suggestion
        "Attempt to add an object of type {} to the scheduler, which "
```

ktbarrett: These strings need to be changed outright, we are not scheduling here.

ktbarrett: Could you have another look at them?

garmin-mjames:
Currently when calling `fork`, the new forked coroutine gets to run immediately, and the originating coroutine waits until `fork` returns before continuing. It might be useful to provide a way to queue the forked coroutine to run after the current one yields back to the scheduler.

ktbarrett:
@garmin-mjames So you are saying to change the behavior of `fork` and friends to not run the forked coroutine immediately, but only after yielding control to the scheduler?

garmin-mjames:
@ktbarrett That would be my preferred behavior, but it's a change from what happens now.

Ideally tests should use proper synchronization triggers such as `Lock` and `Event`, but changing this may break tests that relied on old behavior... we might want to expose the choice: `run_now` and `run_soon`?

cocotb/__init__.py
cmarqu: Was this an accidental "resolve"?

cocotb/__init__.py
```diff
@@ -106,8 +106,88 @@ def _reopen_stream_with_buffering(stream_name):
...
+                if isinstance(self._outcome, cocotb.outcomes.Error):
+                    cocotb.utils.remove_traceback_frames(self._outcome.error, ['watcher', '__await__'])
+                    if cocotb.triggers.Join(self) not in scheduler._trigger2coros:
+                        self.log.warning(
+                            "If you intend to run a task that should end the test if it encounters an error,"
+                            " prefer cocotb.fire_and_forget")
+                        scheduler.finish_test(self._outcome.error)
+                    else:
+                        self.log.warning(
+                            "If you intend to run a task and retrieve the outcome later, perfer cocotb.run_task")
```
cmarqu:
```suggestion
                            "If you intend to run a task and retrieve the outcome later, prefer cocotb.run_task")
```

cocotb/decorators.py
cmarqu: Done.

ktbarrett:
@garmin-mjames We do have `scheduler.queue`, so that wouldn't be too difficult. But I think we should introduce that in some other PR through another interface. If `fork` doesn't run immediately and the user never yields to the scheduler, it may never run, which may be surprising or difficult to reason about. So I think `run_soon` should come with syntax that forces scheduler re-entry. Trio solves this issue using `nursery`, we might want something similar? I guess it depends on what the use cases of `run_soon` are.

garmin-mjames:
`asyncio` has `call_later()` and `call_at()` for callbacks. We could do `run_after(coro, time, units)`, `run_at(coro, time, units)`, `run_on_yield`... lots of options. Scope creep here, but I think it's worth considering the behavior.

ktbarrett:
One potential concern is that if a coroutine that is `run` that is actually never awaited there is no warning or consequence.

ktbarrett:
There is also the question of whether `fire_and_forget` should be a part of this PR or shouldn't exist at all. `run` is the interface that propagates exceptions.

ktbarrett:
Well change is bad, so I'm closing this.

PR #1528, opened 2020-03-22
Add a test to prove that coroutines keep running when passed to First

eric-wieser:
This spun out of a discussion with @ktbarrett  reviewing #1523.


eric-wieser:
@ktbarrett, this look reasonable to you?

eric-wieser:
Merged already as part of #1523

PR #1532, opened 2020-03-23
Add support for using Lock objects in async with

### LINKED ISSUE #1031: Add support for using Lock objects with async with
It would be great to add support for spelling
```python
await lock.acquire()
try:
    # do some stuff
finally:
    lock.release()
```
as
```python
async with lock:
    # do some stuff
```
This would be an easy first issue for someone contributing to cocotb, and would involve implementing `__aenter__` and `__exit__`, and writing a simple test.

andrewandrepowell:
Hello (again),

This change is to resolve #1031.


tests/test_cases/issue_1031/Makefile
```diff
@@ -0,0 +1,31 @@
+###############################################################################
+# Copyright (c) 2020 Potential Ventures Ltd
```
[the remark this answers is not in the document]
cmarqu: For new files, we now use
```
```

tests/test_cases/issue_1031/issue_1031.py
```diff
@@ -0,0 +1,24 @@
+from cocotb import test, fork
```
cmarqu:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause

from cocotb import test, fork
```

tests/test_cases/issue_1031/Makefile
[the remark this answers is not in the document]
cmarqu: (We should probably add this to the PR template...)

tests/test_cases/issue_1031/Makefile
```diff
@@ -0,0 +1,31 @@
...
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
+
+
+include ../../designs/sample_module/Makefile
+
+MODULE = issue_1031
```
eric-wieser: I think it would be better to add this to an existing test like `test_cocotb`. Tests named after their issue number are useful for weird bugs, but not new features.

tests/test_cases/issue_1031/issue_1031.py
```diff
@@ -0,0 +1,24 @@
...
+    resource = [0]
+    lock = Lock()
+
+    fork(co(resource, lock))
+    async with lock:
+        for i in range(4):
+            resource[0] += 1
+            await Timer(10, "ns")
+    assert resource[0]==4
+    async with lock:
```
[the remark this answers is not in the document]
andrewandrepowell: I am not super familiar with the scheduler, but I had assumed co-routines that are blocking on an acquired lock would have priority over co-routines that attempt to acquire the same lock in the future. Is this not case?

Since I am trying to test the async context management stuff, I added an `await Timer(10, "ns")` instead to guarantee the forked co-routine acquires its lock prior to the test in this instance.

tests/test_cases/issue_1031/issue_1031.py
```diff
@@ -0,0 +1,24 @@
...
+async def test_trigger_lock(dut):
+
+    resource = [0]
+    lock = Lock()
+
+    fork(co(resource, lock))
+    async with lock:
+        for i in range(4):
+            resource[0] += 1
+            await Timer(10, "ns")
```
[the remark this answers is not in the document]
andrewandrepowell: My idea was to write a simple test that verified the context management was working by checking to see if a resource was its expected value after those points. Can you write an example of how this would work if the context management was handled from within the for loops?

tests/test_cases/issue_1031/issue_1031.py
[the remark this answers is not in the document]
eric-wieser: Yeah, I agree with @andrewandrepowell here, the thing being tested for here is supposed to be guaranteed.

tests/test_cases/issue_1031/issue_1031.py
```diff
@@ -0,0 +1,24 @@
+from cocotb import test, fork
+from cocotb.triggers import Lock, Timer
+
+@test()
+async def test_trigger_lock(dut):
+
+    resource = [0]
```
eric-wieser:
```suggestion
    resource = 0
```

tests/test_cases/issue_1031/issue_1031.py
```diff
@@ -0,0 +1,24 @@
...
+    lock = Lock()
+
+    fork(co(resource, lock))
+    async with lock:
+        for i in range(4):
+            resource[0] += 1
```
eric-wieser:
```suggestion
            resource += 1
```

tests/test_cases/issue_1031/issue_1031.py
```diff
@@ -0,0 +1,24 @@
...
+    assert resource[0]==4
+    async with lock:
+        assert resource[0]==8
+
+async def co(resource, lock):
+    await Timer(10, "ns")
+    async with lock:
+        for i in range(4):
+            resource[0] += 1
+            await Timer(10, "ns")
```
eric-wieser:
```suggestio
```

tests/test_cases/issue_1031/issue_1031.py
```diff
@@ -0,0 +1,24 @@
+from cocotb import test, fork
+from cocotb.triggers import Lock, Timer
+
+@test()
+async def test_trigger_lock(dut):
+
+    resource = [0]
+    lock = Lock()
+
```
eric-wieser: Would be easier to follow if you defined this within the test
```suggestion

    async def co():
        nonlocal resource
        await Timer(10, "ns")
        async with lock:
            for i in range(4):
                resource += 1
            await Timer(10, "ns")
```

tests/test_cases/issue_1031/issue_1031.py
```diff
@@ -0,0 +1,24 @@
...
+    fork(co(resource, lock))
+    async with lock:
+        for i in range(4):
+            resource[0] += 1
+            await Timer(10, "ns")
+    assert resource[0]==4
+    async with lock:
+        assert resource[0]==8
```
eric-wieser:
```suggestion
        assert resource == 8
```

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1466,3 +1466,47 @@ def test_undecorated_coroutine_yield(dut):
...
+@cocotb.test(timeout_time=100, timeout_unit="ns")
+async def test_except_lock(dut):
+    """
+    Checks to see if exceptions cause the lock to be
+    released.
+    """
+    lock = Lock()
+    try:
+        async with lock:
+            assert False
```
eric-wieser: Best not to use assertions for this,
```suggestion
            raise RuntimeError
```

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1466,3 +1466,47 @@ def test_undecorated_coroutine_yield(dut):
...
+    except AssertionError:
+        pass
+    async with lock:
+        assert True
```
eric-wieser:
```suggestion
        pass
```

tests/test_cases/issue_1031/issue_1031.py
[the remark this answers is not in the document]
ktbarrett: @eric-wieser It is pretty atypical of a lock to make an guarantees about how it is scheduled. That is not documented. With line 24 now added however, it is valid.

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1475,25 +1475,27 @@ def test_undecorated_coroutine_yield(dut):
...
     amount, which is easily predictable if the context management is working.
     """
-    resource = [0]
+    resource = 0
     lock = Lock()
 
-    cocotb.fork(co(resource, lock))
+    async def co():
+        nonlocal resource
+        nonlocal lock
```
eric-wieser:
```suggestion
```
`nonlocal` is only needed for names on the left of `=`, `+=`  etc

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1475,25 +1475,27 @@ def test_undecorated_coroutine_yield(dut):
...
-        assert resource[0]==8
-
-async def co(resource, lock):
+    assert resource==4
     await Timer(10, "ns")
     async with lock:
-        for i in range(4):
-            resource[0] += 1
-            await Timer(10, "ns")
+        assert resource==8
```
eric-wieser: nit: PEP8 says

```suggestion
        assert resource == 8
```

tests/test_cases/issue_1031/issue_1031.py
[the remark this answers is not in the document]
ktbarrett: This is actually fine. You have lock acquisition, you are forcing acquisition order so you aren't depending upon scheduling behavior, and you have statements that cause scheduler re-entry while retaining the lock. Consider this resolved.

cocotb/triggers.py
```diff
@@ -456,13 +456,19 @@ def __repr__(self):
...
         try:
             # do some stuff
         finally:
             lock.release()
+
+    Alternatively, the lock can be used
+    in an asynchronous context::
+
+        async with lock:
+            # do some stuff
```
```suggestion
    .. versionchanged:: 1.4
        The lock can be used as an asynchronous context manager in a :keyword:`with`,
        statement, provided Python 3.6 and onwards are used::

            async with lock:
                # do some stuff
```

eric-wieser:
Just the `newsfragment` to go after that.

cocotb/triggers.py
```diff
@@ -456,13 +456,20 @@ def __repr__(self):
...
             # do some stuff
         finally:
             lock.release()
+
+    The lock can be used as an asynchronous
+    context manager in a :keyword:`with`,
+    statement::
+
+        async with lock:
+            # do some stuff
```
eric-wieser: Ir's important to get the `versionchanged` here and a mention of python 3.6

```suggestion
    .. versionchanged:: 1.4
        The lock can be used as an asynchronous context manager in a
        :keyword:`with`, statement, provided Python 3.6 or newer is used::

            async with lock:
                # do some stuff
```
Also rewrapped the lines, since this is unusually narrow.

andrewandrepowell: Okay. I should point out asynchronous context management was introduced in Python3.5 https://docs.python.org/3/whatsnew/3.5.html.

eric-wieser: Ah, that's a very good point. Then no need to mention the python version.

Still need `versionchanged` though.

cocotb/triggers.py
```diff
@@ -456,13 +456,21 @@ def __repr__(self):
...
+        await lock.acquire()
         try:
             # do some stuff
         finally:
             lock.release()
+
+    The lock can be used as an asynchronous
+    context manager in a :keyword:`with`,
+    statement, provided Python 3.5 or newer is
+    used::
```
ktbarrett: is :keyword:`async with` supported? Also we can drop the version statement. Python 3.5 is already the lowest supported version.

eric-wieser: Seemingly yes. Good idea

```suggestion
    The lock can be used as an asynchronous context manager in an
    :keyword:`async with` statement::
```

cocotb/triggers.py
```diff
@@ -456,13 +456,19 @@ def __repr__(self):
...
         try:
             # do some stuff
         finally:
             lock.release()
+
+    The lock can be used as an asynchronous context manager in an
+    :keyword:`async with` statement::
+
+        async with lock:
+            # do some stuff
```
```suggestion
    .. versionchanged:: 1.4
        The lock can be used as an asynchronous context manager in an
        :keyword:`async with` statement::

            async with lock:
                # do some stuff
```

eric-wieser:
Made the final doc fixups myself

cocotb/triggers.py
```diff
@@ -456,13 +456,20 @@ def __repr__(self):
...
             # do some stuff
         finally:
             lock.release()
+
+    .. versionchanged:: 1.4
+        The lock can be used as an asynchronous context manager in an
+        :keyword:`async with` statement::
+
+            async with lock:
+                # do some stuff
```
eric-wieser: Not rendering correctly: https://cocotb--1532.org.readthedocs.build/en/1532/triggers.html#cocotb.triggers.Lock

Perhaps

```suggestion

        The lock can be used as an asynchronous context manager in an
        :keyword:`async with` statement::

            async with lock:
                # do some stuff
```

documentation/source/newsfragments/1031.feature.rst
```diff
@@ -0,0 +1 @@
+Implemented asynchronous context management with :class:`~cocotb.triggers.Lock`.
```
eric-wieser:
```suggestion
:class:`~cocotb.triggers.Lock` can now be used in :keyword:`async with` statements.
```
newsfargments should describe what's new, not what was done.

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -1466,3 +1466,48 @@ def test_undecorated_coroutine_yield(dut):
...
+            for i in range(4):
+                resource += 1
+                await Timer(10, "ns")
+
+    cocotb.fork(co())
+    async with lock:
+        for i in range(4):
+            resource += 1
+            await Timer(10, "ns")
+    assert resource == 4
```
eric-wieser: How can you guarantee this is `4` here? Couldn't the other coroutine have grabbed the lock and incremented already?

eric-wieser: Right now this test is assuming that releasing a lock is not a yield point. I don't know if it's necessary to guarantee that, so I'd prefer not to assume it in the test.

ktbarrett: That question opens a can of worms... Ideally, no. We do not yield control back to the scheduler between the final await in the with block (still holding the lock) and the next await, which is after this line. **And the `release()` method does not cause and immediate context switch or re-enter the scheduler**.

Lines in bold I'm uncertain about. Lines in bold *should* hold true. I hope this thing doesn't do that...

eric-wieser: You're correct that we don't re-enter the scheduler today, but it's not clear to me that this is a promise we need to make.

I'd be more comfortable if we change this test to the following, simply because then the test would still pass if we swapped from coroutines to threads.
```suggestion
        assert resource == 4
```

andrewandrepowell: > How can you guarantee this is `4` here? Couldn't the other coroutine have grabbed the lock and incremented already?

Hmmm. So, I was making an assumption that releasing the lock doesn't trigger the scheduler. I changed the test so that it should be guaranteed to work.

ktbarrett: I think it's a promise we should make. It makes concurrency *much* easier to reason about when you know that context switches only happen when you see syntax for it. Since you've mentioned Trio, I actually looked into it the other day, and they make a BIG DEAL about that. They call them "checkpoints".

eric-wieser: Right, but the whole point in `async with` is it has the potential for _two_ checkpoints, one at the start and one at the end. If you want a one-sided checkpoint, you can use `with (await lock)` instead of `async with lock`.

eric-wieser: From their docs:

> Partial exception for async context managers: Both the entry and exit of an async with block are defined as async functions; but for a particular type of async context manager, it’s often the case that only one of them is able to block, which means only that one will act as a checkpoint. This is documented on a case-by-case basis.

So basically, entrypoints are a good idea, but you can't see them from syntax alone.

eric-wieser:
No further comments, let's put this in. Thanks for the persistence on the tests!

PR #1534, opened 2020-03-23
Add pytest to CI

garmin-mjames:
Adds a PythonTests stage to Travis CI that runs `pytest`. For use in unit testing Python-only code.

I've used a local test for `BinaryValue` as the first example test.

I've never messed with Travis settings so I have no idea if this initial commit will work.

Tasks:
- [x] Make Travis work


eric-wieser:
I think I'd prefer to not try and fix BinaryValue in the same patch as adding the test stuff to travis.

Can you start by just testing the bits that don't seem broken, and putting a PR up for that. The tests don't need to be long, the important thing is having an obvious place for everyone else to add more tests later.

garmin-mjames:
I pared down the test to just the parts of `BinaryValue` that work and removed the broken edge cases.

garmin-mjames:
Is Python 3.5 the best target for `pytest`?

themperek:
Maybe adding a makefile in pytest directory that calls pytest would be better. It would run within existing regressions. Maybe here: https://github.com/cocotb/cocotb/blob/master/tests/Makefile ?

eric-wieser:
If you're feeling generous, `test_cocotb.test_binary_value_compat` and `test_cocotb.test_binary_value` could be moved into this file and converted to pytest. Fine to leave to a follow-up

eric-wieser:
> Is Python 3.5 the best target for `pytest`?

Ideally we'd test both 3.5 and 3.8 (oldest and newest supported version)

eric-wieser:
> Maybe adding a makefile in pytest directory that calls pytest would be better. It would run within existing regressions. Maybe here: https://github.com/cocotb/cocotb/blob/master/tests/Makefile ?

If we do this, we'd probably want to make it a pre-requisite of the simulation tests - little point wasting power running simulations if our code can't pass unit tests, especially since the unit tests give better log messages.

themperek:
>If we do this, we'd probably want to make it a pre-requisite of the simulation tests - little point wasting power running simulations if our code can't pass unit tests, especially since the unit tests give better log messages.

In this case, adding `py.test tests/pytest/` before 
https://github.com/cocotb/cocotb/blob/master/tox.ini#L17 should do this.

eric-wieser:
That sounds like a good idea to me.

tests/pytest/test_binary_value.py
```diff
@@ -0,0 +1,244 @@
...
+    with pytest.deprecated_call():
+        vec = BinaryValue(value=0, bits=16)
+    assert vec.n_bits == 16
+
+    vec = BinaryValue(0, 16)
+    assert vec.n_bits == 16
+
+    with pytest.raises(TypeError):
+        vec = BinaryValue(value=0, bits=16, n_bits=17)
+        pytest.fail("Expected TypeError when using bits and n_bits at the same time")
```
eric-wieser:
```suggestion
        BinaryValue(value=0, bits=16, n_bits=17)
```

tests/pytest/test_binary_value.py
```diff
@@ -0,0 +1,244 @@
...
+        bin3 = bin1[32]
+
+    with pytest.raises(IndexError):
+        bin4 = bin1[1:2]
+
+    with pytest.raises(IndexError):
+        bin5 = bin1[-1:4]
+
+    with pytest.raises(IndexError):
+        bin6 = bin1[2:-2]
```
eric-wieser: If it raises an exception there's no point naming the result, is there?
```suggestion
    with pytest.raises(IndexError):
        bin1[32]

    with pytest.raises(IndexError):
        bin1[1:2]

    with pytest.raises(IndexError):
        bin1[-1:4]

    with pytest.raises(IndexError):
        bin1[2:-2]
```

pytest.ini
```diff
@@ -0,0 +1,6 @@
+[pytest]
```
themperek: We could add this to `tox.ini` and `setup.cfg` to not add even more files see https://docs.pytest.org/en/latest/customize.html#finding-the-rootdir

eric-wieser: Would probably favor `setup.cfg`. Note that this line then becomes `[tool:pytest]`

eric-wieser:
Travis seems to be having trouble with the webhook, we'll need to manually find the travis run.

themperek:
here: https://travis-ci.org/github/cocotb/cocotb/pull_requests

garmin-mjames:
It seems to have kicked off a new build: https://travis-ci.org/github/cocotb/cocotb/builds/666485746?utm_medium=notification&utm_source=github_status

eric-wieser:
https://travis-ci.org/github/cocotb/cocotb/jobs/666485749#L922

PR #1535, opened 2020-03-24
Remove unused entrypoint.vhdl file

themperek:
It looks like `entrypoint.vhdl` is not used anywhere and is a leftover from early FLI development.

@ktbarrett Can you confirm?


ktbarrett:
@themperek it works.

PR #1536, opened 2020-03-24
Remove flake8 ignores on blank lines

eric-wieser:
This should give better error messages than our whitespace linter, but won't touch all files.

xref #1281, cc @ktbarrett


cmarqu:
I'm totally okay with not checking e.g. ReST files.

eric-wieser:
They're already checked by a `grep` elsewhere anyway.

PR #1538, opened 2020-03-25
Enable SV assertion in Makefile.verilator

nndurj:
SV assertions are disabled by default. Adding `--assert` to make it consistent with commercial simulators


themperek:
@nndurj We will actually remove all settings that are not essential https://github.com/cocotb/cocotb/issues/1495 and can be made by the user. 

You can add it in your makefile by adding `COMPILE_ARGS=--assert`

nndurj:
@themperek  makes sense

PR #1540, opened 2020-03-27
Use extension types for handles instead of integers

eric-wieser:
This means that:
* Passing handles back and forth should be faster, as there is no need to convert back and forth between big integers and pointers
* The simulator module becomes more type safe, and does not allow passing objects of one type to functions expecting a different type
* It is no longer possible to pass arbitrary integers in that don't correspond to actual handles
* Instead of returning `0`, failure is indicated by returning `None`. This results in a `TypeError` rather than a segfault if used again unchecked.

There is one behavior change here, in that `simulator.next(simulator.iterate(not_iterable))` no longer raises StopIteration.
If you are seeing this in the commit history, then this change didn't manifest in any regression tests.

---

This follows on from #1427.

### DIFF
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -643,13 +643,6 @@ static PyObject *next(PyObject *self, PyObject *args)
         return NULL;
     }
 
-    // TODO:eric-wieser: It's valid for iterate to return a NULL handle, to make the Python
-    // intuitive we simply raise StopIteration on the first iteration
-    // if (!hdl) {
-    //     PyErr_SetNone(PyExc_StopIteration);
-    //     return NULL;
-    // }
-
     result = gpi_next(hdl_obj->hdl);
 
     // Raise StopIteration when we're done
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -41,10 +41,122 @@ static int sim_ending = 0;
 
 #include "simulatormodule.h"
 #include <cocotb_utils.h>     // COCOTB_UNUSED
+#include <type_traits>
+
 
 struct module_state {
     PyObject *error;
 };
+
+
+/* define the extension types as templates */
+namespace {
+    template<typename gpi_hdl>
+    struct gpi_hdl_Object {
+        PyObject_HEAD
+        gpi_hdl hdl;
+
+        // The python type object, in a place that is easy to retrieve in templates
+        static PyTypeObject py_type;
+    };
+
+    /** __repr__ shows the memory address of the internal handle */
+    template<typename gpi_hdl>
+    static PyObject *gpi_hdl_repr(gpi_hdl_Object<gpi_hdl> *self) {
+        auto *type = Py_TYPE(self);
+        return PyUnicode_FromFormat("<%s at %p>", type->tp_name, self->hdl);
+    }
+
+    /** __hash__ returns the pointer itself */
+    template<typename gpi_hdl>
+    static Py_hash_t gpi_hdl_hash(gpi_hdl_Object<gpi_hdl> *self) {
+        auto ret = reinterpret_cast<Py_hash_t>(self->hdl);
+        // hash must never return -1
+        if (ret == (Py_hash_t)-1) {
+            ret = (Py_hash_t)-2;
+        }
+        return ret;
+    }
+
+    /**
+     * Create a new python handle object from a pointer, returning None if the
+     * pointer is NULL.
+     */
+    template<typename gpi_hdl>
+    static PyObject *gpi_hdl_New(gpi_hdl hdl) {
+        if (hdl == NULL) {
+            Py_RETURN_NONE;
+        }
+        auto *obj = PyObject_New(gpi_hdl_Object<gpi_hdl>, &gpi_hdl_Object<gpi_hdl>::py_type);
+        if (obj == NULL) {
+            return NULL;
+        }
+        obj->hdl = hdl;
+        return (PyObject *)obj;
+    }
+
+    /** Comparison checks if the types match, and then compares pointers */
+    template<typename gpi_hdl>
+    static PyObject *gpi_hdl_richcompare(PyObject *self, PyObject *other, int op) {
+        if (
+            Py_TYPE(self) != &gpi_hdl_Object<gpi_hdl>::py_type ||
+            Py_TYPE(other) != &gpi_hdl_Object<gpi_hdl>::py_type
+        ) {
+            Py_RETURN_NOTIMPLEMENTED;
+        }
+
+        auto self_hdl_obj = reinterpret_cast<gpi_hdl_Object<gpi_hdl>*>(self);
+        auto other_hdl_obj = reinterpret_cast<gpi_hdl_Object<gpi_hdl>*>(other);
+
+        switch (op) {
+            case Py_EQ: return PyBool_FromLong(self_hdl_obj->hdl == other_hdl_obj->hdl);
+            case Py_NE: return PyBool_FromLong(self_hdl_obj->hdl != other_hdl_obj->hdl);
+            default: Py_RETURN_NOTIMPLEMENTED;
+        }
+    }
+
+    // Initialize the python type slots
+    template<typename gpi_hdl>
+    PyTypeObject fill_common_slots() {
+        PyTypeObject type;
+        type.ob_base = {PyObject_HEAD_INIT(NULL) 0};
+        type.tp_basicsize = sizeof(gpi_hdl_Object<gpi_hdl>);
+        type.tp_repr = (reprfunc)gpi_hdl_repr<gpi_hdl>;
+        type.tp_hash = (hashfunc)gpi_hdl_hash<gpi_hdl>;
+        type.tp_flags = Py_TPFLAGS_DEFAULT;
+        type.tp_richcompare = gpi_hdl_richcompare<gpi_hdl>;
+        return type;
+    }
+
+    template<>
+    PyTypeObject gpi_hdl_Object<gpi_sim_hdl>::py_type = []() -> PyTypeObject {
+        auto type = fill_common_slots<gpi_sim_hdl>();
+        type.tp_name = "gpi_sim_hdl";
+        return type;
+    }();
+
+    template<>
+    PyTypeObject gpi_hdl_Object<gpi_iterator_hdl>::py_type = []() -> PyTypeObject {
+        auto type = fill_common_slots<gpi_iterator_hdl>();
+        type.tp_name = "gpi_iterator_hdl";
+        return type;
+    }();
+
+    template<>
+    PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
+        auto type = fill_common_slots<gpi_cb_hdl>();
+        type.tp_name = "gpi_cb_hdl";
+        return type;
+    }();
+
+    // get the python type corresponding to the static type of a PyObject pointer
+    template<typename gpi_hdl_Object_ptr>
+    PyTypeObject* python_type_for() {
+        return &std::remove_pointer<gpi_hdl_Object_ptr>::type::py_type;
+    }
+}
+
+
 
 #define GETSTATE(m) ((struct module_state*)PyModule_GetState(m))
 
@@ -69,53 +181,6 @@ struct sim_time {
 };
 
 static struct sim_time cache_time;
-
-// Converter function for turning a Python long into a sim handle, such that it
-// can be used by PyArg_ParseTuple format O&.
-static int gpi_sim_hdl_converter(PyObject *o, gpi_sim_hdl *data)
-{
-    void *p = PyLong_AsVoidPtr(o);
-    if ((p == NULL) && PyErr_Occurred()) {
-        return 0;
-    }
-    if (p == NULL) {
-        PyErr_SetString(PyExc_ValueError, "handle cannot be 0");
-        return 0;
-    }
-    *data = (gpi_sim_hdl)p;
-    return 1;
-}
-
-// Same as above, for a callback handle.
-static int gpi_cb_hdl_converter(PyObject *o, gpi_cb_hdl *data)
-{
-    void *p = PyLong_AsVoidPtr(o);
-    if ((p == NULL) && PyErr_Occurred()) {
-        return 0;
-    }
-    if (p == NULL) {
-        PyErr_SetString(PyExc_ValueError, "handle cannot be 0");
-        return 0;
-    }
-    *data = (gpi_cb_hdl)p;
-    return 1;
-}
-
-
-// Same as above, for an iterator handle.
-static int gpi_iterator_hdl_converter(PyObject *o, gpi_iterator_hdl *data)
-{
-    void *p = PyLong_AsVoidPtr(o);
-    if ((p == NULL) && PyErr_Occurred()) {
-        return 0;
-    }
-    if (p == NULL) {
-        PyErr_SetString(PyExc_ValueError, "handle cannot be 0");
-        return 0;
-    }
-    *data = (gpi_iterator_hdl)p;
-    return 1;
-}
 
 /**
  * @name    Callback Handling
@@ -287,7 +352,7 @@ static PyObject *register_readonly_callback(PyObject *self, PyObject *args)
 
     hdl = gpi_register_readonly_callback((gpi_function_t)handle_gpi_callback, callback_data_p);
 
-    PyObject *rv = PyLong_FromVoidPtr(hdl);
+    PyObject *rv = gpi_hdl_New(hdl);
     FEXIT
 
     return rv;
@@ -340,7 +405,7 @@ static PyObject *register_rwsynch_callback(PyObject *self, PyObject *args)
 
     hdl = gpi_register_readwrite_callback((gpi_function_t)handle_gpi_callback, callback_data_p);
 
-    PyObject *rv = PyLong_FromVoidPtr(hdl);
+    PyObject *rv = gpi_hdl_New(hdl);
     FEXIT
 
     return rv;
@@ -393,7 +458,7 @@ static PyObject *register_nextstep_callback(PyObject *self, PyObject *args)
 
     hdl = gpi_register_nexttime_callback((gpi_function_t)handle_gpi_callback, callback_data_p);
 
-    PyObject *rv = PyLong_FromVoidPtr(hdl);
+    PyObject *rv = gpi_hdl_New(hdl);
     FEXIT
 
     return rv;
@@ -465,7 +530,7 @@ static PyObject *register_timed_callback(PyObject *self, PyObject *args)
     hdl = gpi_register_timed_callback((gpi_function_t)handle_gpi_callback, callback_data_p, time_ps);
 
     // Check success
-    PyObject *rv = PyLong_FromVoidPtr(hdl);
+    PyObject *rv = gpi_hdl_New(hdl);
     FEXIT
 
     return rv;
@@ -496,10 +561,12 @@ static PyObject *register_value_change_callback(PyObject *self, PyObject *args) 
         return NULL;
     }
 
-    PyObject *pSihHdl = PyTuple_GetItem(args, 0);
-    if (!gpi_sim_hdl_converter(pSihHdl, &sig_hdl)) {
+    PyObject *pSigHdl = PyTuple_GetItem(args, 0);
+    if (Py_TYPE(pSigHdl) != &gpi_hdl_Object<gpi_sim_hdl>::py_type) {
+        PyErr_SetString(PyExc_TypeError, "First argument must be a gpi_sim_hdl");
         return NULL;
     }
+    sig_hdl = ((gpi_hdl_Object<gpi_sim_hdl> *)pSigHdl)->hdl;
 
     // Extract the callback function
     function = PyTuple_GetItem(args, 1);
@@ -538,7 +605,7 @@ static PyObject *register_value_change_callback(PyObject *self, PyObject *args) 
                                              edge);
 
     // Check success
-    PyObject *rv = PyLong_FromVoidPtr(hdl);
+    PyObject *rv = gpi_hdl_New(hdl);
     FEXIT
 
     return rv;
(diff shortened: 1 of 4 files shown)


cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -46,6 +46,161 @@ struct module_state {
...
+    if (hdl == NULL) {
+        Py_RETURN_NONE;
+    }
+    obj = PyObject_New(gpi_sim_hdl_Object, &gpi_sim_hdl_Type);
+    if (obj == NULL) {
+        return NULL;
+    }
+    obj->hdl = hdl;
+    return (PyObject *)obj;
+}
```
eric-wieser: Note that #1442 would allow this chunk of code to become a template used for all three types, rather than it needing to be repeated 2 more times below

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -495,9 +603,11 @@ static PyObject *register_value_change_callback(PyObject *self, PyObject *args)
     }
 
     PyObject *pSihHdl = PyTuple_GetItem(args, 0);
```
cmarqu:
```suggestion
    PyObject *pSigHdl = PyTuple_GetItem(args, 0);
```
(and usage below obviously too)

eric-wieser: Good catch.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -46,6 +46,161 @@ struct module_state {
...
+static PyObject *gpi_sim_hdl_repr(gpi_sim_hdl_Object *self) {
+    PyTypeObject *type = Py_TYPE(self);
+    return PyUnicode_FromFormat("<%s at %p>",
+                                  type->tp_name, self->hdl);
+}
+static Py_hash_t gpi_sim_hdl_hash(gpi_sim_hdl_Object *self) {
+    Py_hash_t ret = (Py_hash_t)self->hdl;
+    if (ret == (Py_hash_t)-1) {
+        ret = (Py_hash_t)-2;
+    }
```
ktbarrett: This seems unnecessary, `gpi_sim_hdl`s alignment is not 1, the pointer will never be equal to -1.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -46,6 +46,161 @@ struct module_state {
...
+    return ret;
+}
+
+static PyObject *gpi_sim_hdl_richcompare(PyObject *self, PyObject *other, int op) {
+    gpi_sim_hdl_Object *self_hdl_obj;
+    gpi_sim_hdl_Object *other_hdl_obj;
+
+    if (
+            Py_TYPE(self) != &gpi_sim_hdl_Type ||
+            Py_TYPE(other) != &gpi_sim_hdl_Type) {
```
ktbarrett: The [documentation](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_richcompare) states 

>The first parameter is guaranteed to be an instance of the type that is defined by PyTypeObject.

So you can remove the first condition in this if, and even change the type on `self` in the parameter list.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -46,6 +46,161 @@ struct module_state {
...
+    .tp_name = "simulator.gpi_sim_hdl",
+    .tp_basicsize = sizeof(gpi_sim_hdl_Object),
+    .tp_flags = Py_TPFLAGS_DEFAULT,
+    .tp_repr = (reprfunc)gpi_sim_hdl_repr,
+    .tp_hash = (hashfunc)gpi_sim_hdl_hash,
+    .tp_richcompare = gpi_sim_hdl_richcompare,
+};
+
+static PyObject *gpi_sim_hdl_New(gpi_sim_hdl hdl) {
+    gpi_sim_hdl_Object *obj;
```
ktbarrett: Nit, you can move this declaration to the assignment below.

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -943,16 +1053,16 @@ static PyObject *get_num_elems(PyObject *self, PyObject *args)
...
+    if (!PyArg_ParseTuple(args, "O!", &gpi_sim_hdl_Type, &hdl_obj)) {
         return NULL;
     }
 
-    int indexable = gpi_is_indexable((gpi_sim_hdl)hdl);
-    int rng_left  = gpi_get_range_left((gpi_sim_hdl)hdl);
-    int rng_right = gpi_get_range_right((gpi_sim_hdl)hdl);
+    int indexable = gpi_is_indexable(hdl_obj->hdl);
+    int rng_left  = gpi_get_range_left(hdl_obj->hdl);
+    int rng_right = gpi_get_range_right(hdl_obj->hdl);
```
ktbarrett: Yikes... What does this do if the object is not indexable?

cocotb/share/lib/simulator/simulatormodule.c
```diff
@@ -566,30 +676,30 @@ static PyObject *iterate(PyObject *self, PyObject *args)
...
+    // TODO:eric-wieser: It's valid for iterate to return a NULL handle, to make the Python
     // intuitive we simply raise StopIteration on the first iteration
-    if (!hdl) {
-        PyErr_SetNone(PyExc_StopIteration);
-        return NULL;
-    }
+    // if (!hdl) {
+    //     PyErr_SetNone(PyExc_StopIteration);
+    //     return NULL;
+    // }
```
ktbarrett: So,
1. `iterate` is not defined on the object and `gpi_iterate` returns `NULL`
2. `gpi_iterator_hdl_New` returns `None` if the iterator is `NULL`,
3. `_SimIterator` sets `._iter` to `None` 
4. On the first iteration, `next` gets `None`
5. `PyArg_ParseTuple` above results in `TypeError`

Seems like you could get rid of this, no `gpi_iterator_hdl_Type` will have a `NULL` handle. But maybe you should add a check to `iterate` to check the outcome of `gpi_iterate` instead?

cocotb/share/lib/simulator/simulatormodule.c
eric-wieser: As written here this guards against subclasses.

cocotb/share/lib/simulator/simulatormodule.c
eric-wieser: I wanted to play it safe here. I don't think the standard makes any guarantee about pointer bitpatterns, but python requires that -1 is never returns.

eric-wieser:
Updated with templates. Unfortunately the move from C to C++ means I have to drop the designated initializers, so the type slot definitions become a lot longer.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -41,11 +41,127 @@ static int sim_ending = 0;
...
+    }
+    return ret;
+}
+
+template<typename gpi_hdl>
+static PyObject *gpi_hdl_New(gpi_hdl hdl) {
+    if (hdl == NULL) {
+        Py_RETURN_NONE;
+    }
+    gpi_hdl_Object<gpi_hdl> *obj = PyObject_New(gpi_hdl_Object<gpi_hdl>, &gpi_hdl_Object<gpi_hdl>::py_type);
```
ktbarrett: Does `PyObject_New`'s implementation allow you to use `auto` here?

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -41,11 +41,127 @@ static int sim_ending = 0;
...
+    gpi_hdl_Object<gpi_hdl> *other_hdl_obj;
+
+    if (
+            Py_TYPE(self) != &gpi_hdl_Object<gpi_hdl>::py_type ||
+            Py_TYPE(other) != &gpi_hdl_Object<gpi_hdl>::py_type) {
+        Py_RETURN_NOTIMPLEMENTED;
+    }
+
+    self_hdl_obj = (gpi_hdl_Object<gpi_hdl> *)self;
+    other_hdl_obj = (gpi_hdl_Object<gpi_hdl> *)other;
```
ktbarrett: These can be `auto` and not declared above as well. Take this as a general comment.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -548,18 +619,18 @@ static PyObject *register_value_change_callback(PyObject *self, PyObject *args)
...
 {
     COCOTB_UNUSED(self);
-    gpi_sim_hdl hdl;
+    gpi_hdl_Object<gpi_sim_hdl> *hdl_obj;
     int type;
     gpi_iterator_hdl result;
     PyObject *res;
 
-    if (!PyArg_ParseTuple(args, "O&i", gpi_sim_hdl_converter, &hdl, &type)) {
+    if (!PyArg_ParseTuple(args, "O!i", &std::remove_pointer<decltype(hdl_obj)>::type::py_type, &hdl_obj, &type)) {
```
ktbarrett: Yuck, C++ does have a tendency to get ugly. In this case type reflection isn't necessary. My rule of thumb is use it when you are writing generic code or when it is more concise/readable.
```suggestion
    if (!PyArg_ParseTuple(args, "O!i", &gpi_hdl_Object<gpi_sim_hdl>::py_type, &hdl_obj, &type)) {
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -41,11 +41,127 @@ static int sim_ending = 0;
...
+    /* tp_getset            */ 0,
+    /* tp_base              */ 0,
+    /* tp_dict              */ 0,
+    /* tp_descr_get         */ 0,
+    /* tp_descr_set         */ 0,
+    /* tp_dictoffset        */ 0,
+    /* tp_init              */ 0,
+    /* tp_alloc             */ 0,
+    /* tp_new               */ 0,
+};
```
ktbarrett: You might be interested in [this](https://github.com/googleprojectzero/functionsimsearch/blob/ec5d9e1224ff915b3dc9e7af19e21e110c25239c/pybindings/pybindings.cpp#L346-L358) as a way around this.

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: Yes!

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: I'd argue it is necessary, because a typo here would not be detected at compile-time.

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: That's a nice trick, thanks

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: In favor of this comment generally, but I'd appreciate if you could point out where else I've forgotten to apply it.

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: I've made it shorter anyway

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: Adopted. Needed slight modification not to fail due to uninitialized field warnings.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -41,11 +41,115 @@ static int sim_ending = 0;
...
+    template<typename gpi_hdl>
+    static PyObject *gpi_hdl_richcompare(PyObject *self, PyObject *other, int op) {
+        if (
+                Py_TYPE(self) != &gpi_hdl_Object<gpi_hdl>::py_type ||
+                Py_TYPE(other) != &gpi_hdl_Object<gpi_hdl>::py_type) {
+            Py_RETURN_NOTIMPLEMENTED;
+        }
+
+        auto self_hdl_obj = (gpi_hdl_Object<gpi_hdl> *)self;
+        auto other_hdl_obj = (gpi_hdl_Object<gpi_hdl> *)other;
```
ktbarrett, line 101:
```suggestion
        auto self_hdl_obj = reinterpret_cast<gpi_hdl_Object<gpi_hdl> *>(self);
        auto other_hdl_obj = reinterpret_cast<gpi_hdl_Object<gpi_hdl> *>(other);
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -41,11 +41,115 @@ static int sim_ending = 0;
...
+        PyObject_HEAD
+        gpi_hdl hdl;
+
+        // The python type object, in a place that is easy to retrieve in templates
+        static PyTypeObject py_type;
+    };
+
+    template<typename gpi_hdl>
+    static PyObject *gpi_hdl_repr(gpi_hdl_Object<gpi_hdl> *self) {
+        PyTypeObject *type = Py_TYPE(self);
```
ktbarrett, line 65:
```suggestion
        auto type = Py_TYPE(self);
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -41,11 +41,115 @@ static int sim_ending = 0;
...
+    }
+
+    // Initialize the python type slots
+    template<typename gpi_hdl>
+    PyTypeObject fill_common_slots() {
+        PyTypeObject type;
+        type.ob_base = {PyObject_HEAD_INIT(NULL) 0};
+        type.tp_basicsize = sizeof(gpi_hdl_Object<gpi_hdl>);
+        type.tp_repr = (reprfunc)gpi_hdl_repr<gpi_hdl>;
+        type.tp_hash = (hashfunc)gpi_hdl_hash<gpi_hdl>;
```
ktbarrett, line 117: It would be nice if you didn't have to cast this. Kind of makes using `gpi_hdl_Object` in the function pointless.

eric-wieser: I don't agree, it makes the function body simpler. At any rate, it's convention to do so in python.

ktbarrett: The C++ standard (and probably the C standard as well), state that it's UB. Not that I expect it to do the wrong thing, but I tend to shy away from it if possible.

eric-wieser: In that case, the python runtime is already full of the same UB. I think matching the python docs is more valuable than avoiding UB we already rely on elsewhere.

eric-wieser:
@ktbarret, regarding https://github.com/cocotb/cocotb/pull/1540#discussion_r400644004

> So,
> 
> 1. `iterate` is not defined on the object and `gpi_iterate` returns `NULL`
> 2. `gpi_iterator_hdl_New` returns `None` if the iterator is `NULL`,
> 3. `_SimIterator` sets `._iter` to `None`
> 4. On the first iteration, `next` gets `None`
> 5. `PyArg_ParseTuple` above results in `TypeError`
> 
> Seems like you could get rid of this, no `gpi_iterator_hdl_Type` will have a `NULL` handle. But maybe you should add a check to `iterate` to check the outcome of `gpi_iterate` instead?

Are you suggesting I simply remove the commented out code, or asking for more?

ktbarrett:
@eric-wieser I'm not a fan of leaving commenting out code around. I'm also not against this PR including a small fix for this case. Which would include removing the commented out code and adding a check to the result of `gpi_iterate`, or whatever you think is best.

eric-wieser:
So something like this?
```diff
--- C:\Users\wiese\Repos\cocotb\cocotb\share\lib\simulator\simulatormodule.cpp	Wed Apr  1 14:09:25 2020
+++ C:\Users\wiese\Repos\cocotb\cocotb\share\lib\simulator\simulatormodule.cpp	Wed Apr  1 14:13:04 2020
@@ -625,6 +625,11 @@
 
     result = gpi_iterate(hdl_obj->hdl, (gpi_iterator_sel_t)type);
 
+    if (result == NULL) {
+        PyErr_SetString(PyExc_ValueError, "Unable to construct iterator");
+        return NULL;
+    }
+
     res = gpi_hdl_New(result);
 
     return res;
```

ktbarrett:
@eric-wieser That's appropriate. This way there is no way to construct an invalid `gpi_sim_iterator` object, so if `next` gets a `gpi_sim_iterator`, it won't fail. Think RAII.

eric-wieser:
> This way there is no way to construct an invalid gpi_sim_iterator object,

I'm a bit confused by that. Are you referring to the C type `gpi_iterator_hdl`, or the python type `simulator.gpi_iterator_hdl`? The former is unavoidably constructed as an invalid `NULL`, while the later is already not constructed in the patch as written

ktbarrett:
No, currently the patch would construct and invalid C object into `None` rather than throw an error. I see now that the Python is dependent upon that behavior and the change would get much larger. Ehhh, maybe we should hold this off for now...

eric-wieser:
cc @themperek, in the long run this could be replaced with pybind11

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -568,30 +634,30 @@ static PyObject *iterate(PyObject *self, PyObject *args)
...
     PyObject *res;
 
-    if (!PyArg_ParseTuple(args, "O&", gpi_iterator_hdl_converter, &hdl)) {
+    if (!PyArg_ParseTuple(args, "O!", python_type_for<decltype(hdl_obj)>(), &hdl_obj)) {
         return NULL;
     }
 
-    // It's valid for iterate to return a NULL handle, to make the Python
+    // TODO:eric-wieser: It's valid for iterate to return a NULL handle, to make the Python
     // intuitive we simply raise StopIteration on the first iteration
```
imphil, line 646: Do we with the code commented out below? Please file a follow-up issue and reference it here if needed, it's easier to track than TODO comments.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -41,11 +41,122 @@ static int sim_ending = 0;
...
+        }
+        auto *obj = PyObject_New(gpi_hdl_Object<gpi_hdl>, &gpi_hdl_Object<gpi_hdl>::py_type);
+        if (obj == NULL) {
+            return NULL;
+        }
+        obj->hdl = hdl;
+        return (PyObject *)obj;
+    }
+
+    /** Comparison checks the types match, and then compares pointers */
```
imphil, line 98:
```suggestion
    /** Comparison checks if the types match, and then compares pointers */
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -41,11 +41,122 @@ static int sim_ending = 0;
...
+    template<typename gpi_hdl>
+    static PyObject *gpi_hdl_richcompare(PyObject *self, PyObject *other, int op) {
+        if (
+                Py_TYPE(self) != &gpi_hdl_Object<gpi_hdl>::py_type ||
```
imphil, line 102: Looks like clang-format messed up here.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -41,11 +41,122 @@ static int sim_ending = 0;
...
 
 struct module_state {
     PyObject *error;
 };
 
+
+/* define the extension types as templates */
+namespace {
+    template<typename gpi_hdl>
+    struct gpi_hdl_Object {
```
imphil, line 55: I'm not a fan of these mixed underscore_camelCase names. We probably have to die one death, but internal consistency would go for gpi_hdl_object and gpi_hdl_new.

eric-wieser: `_Object` and `_New` is to invoke similarities to `PyTupleObject` and `PyTuple_New`. I could go for `GpiHdlObject` instead if you'd prefer, but then that bears less resemblance to the types from `gpi.h`.

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: Unlikely given I'm not using it! I think some other open source project I contributed too wanted two indent levels for wrapped lines, to distinguish them from nested blocks.

I don't feel strongly, let me know if you want me to unindent this one level.

eric-wieser: Do you prefer the following?
```C
        if (
            Py_TYPE(self) != &gpi_hdl_Object<gpi_hdl>::py_type ||
            Py_TYPE(other) != &gpi_hdl_Object<gpi_hdl>::py_type
        ) {
            Py_RETURN_NOTIMPLEMENTED;
        }
```

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: Will fix, I've worked out the root issue, and it's just documented badly.

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: Went ahead and used that approach, since it was quicker than awaiting a response.

imphil: I'd go for 
```
        if (Py_TYPE(self) != &gpi_hdl_Object<gpi_hdl>::py_type ||
            Py_TYPE(other) != &gpi_hdl_Object<gpi_hdl>::py_type
        ) {
            Py_RETURN_NOTIMPLEMENTED;
        }
```

but either way is fine.

cocotb/share/lib/simulator/simulatormodule.cpp
imphil: That's what I mean dying one death: `gpi_hdl` isn't named the way you'd expect Python objects to be named (PythonObject), which makes the `_New` suffix look strange. 

In our C/C++ code base, we go for lower_snake_case, so gpi_hdl_object and gpi_hdl_new would fit that scheme. GpiHdlObject isn't consistent with the rest of our C/C++ code base, so that's not really an option IMO. 

Up to you: either keep the current "compromise solution" with "gpi_hdl_Object", or go for "gpi_hdl_object" to fully match our style, giving up the similarity to the Python code base.

eric-wieser: > GpiHdlObject isn't consistent with the rest of our C/C++ code base

Not sure what you mean, in `GpiCommon.cpp` we already have `GpiObjHdl`.

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: Not a fan of that because it's quite hard to distinguish at a glance from the nonsensical but legal:
```C++
        if (Py_TYPE(self) != &gpi_hdl_Object<gpi_hdl>::py_type) {
            Py_TYPE(other) != &gpi_hdl_Object<gpi_hdl>::py_type;
            Py_RETURN_NOTIMPLEMENTED;
        }
```

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: We just have this C interface that renames all our `PascalCaseClasses::with_snake_methods` to `gpi_class_snake_method`, so have to juggle multiple naming conventions.

eric-wieser: `PyGpiHdlObject` might also be sensible, although I'm not quite sure if we're supposed to use the `Py` prefix.

PR #1543, opened 2020-03-29
Do not ignore fail in hal target

### LINKED ISSUE #1539: Quickstart fails on Centos 7.7
Following the Quckstart instructions in README.md failed for me. I'm using CentOS 7.7 with Python 3.6.8. I think swig needs to be added to the list of prerequisites for Linux. It also doesn't seem to work with swig 2.0.10 - maybe this is expected?

I would supply a patch to the docs, but I'm uncertain exactly what the dependency should be so that it works on all systems. I only have my CentOS box and I'm new to swig.

Without swig installed the tests fail with:

```
     0.00ns CRITICAL cocotb.regression                         regression.py:153  in initialise                      Failed to import module test_endian_swapper_hal: No module named 'hal'
```

It took me longer than it should have to notice that further up it said that it could not find swig:

```
swig -python -outcurrentdir ../hal/endian_swapper_hal.h
make[2]: swig: Command not found
make[2]: *** [endian_swapper_hal_wrap.c] Error 127
make[2]: Leaving directory `/home/user/git/cocotb/examples/endian_swapper/cosim'
make[1]: [hal] Error 2 (ignored)
```

With swig 2.0.10 installed it does better, but still fails:

```
30755.03ns ERROR    cocotb.regression                         regression.py:396  in _score_test                     Test Failed: initial_hal_test (result was SystemError)
                                                                                                                     StopIteration
                                                                                                                     
                                                                                                                     The above exception was the direct cause of the following exception:
                                                                                                                     
                                                                                                                     SystemError: <built-in function delete_endian_swapper_state_t> returned a result with an error set
```

With package swig3 (3.0.12) installed all is well.

cmarqu:
This is to "fail early" and prevents a more obscure fail later in the run.
Fixes #1539.

### DIFF
--- a/examples/endian_swapper/tests/Makefile
+++ b/examples/endian_swapper/tests/Makefile
@@ -69,7 +69,7 @@ include $(shell cocotb-config --makefiles)/Makefile.sim
 ifeq ($(OS),Linux)
 .PHONY: hal
 hal:
-	-cd ../cosim && make
+	cd ../cosim && make
 
 clean::
 	-cd ../cosim && make clean


eric-wieser:
It would probably be worth looking for other misuses of `-` in the makefiles.

cmarqu:
I now grepped (``ack``ed actually) for ``\t-`` and ``\t\+-`` and ``\t@-`` and didn't find more suspicious cases.
Where it's used is for ``rm``ing things. The only place where it likely isn't needed is
``-if [ -x $(PRODUCT) ]; then chmod -x $(PRODUCT); fi"`` in ``tests/test_cases/*/sim_build/csrc/Makefile``

eric-wieser:
I think the - on line 75 of this file is also wrong, but not in nearly as notable a way.

PR #1545, opened 2020-03-29
Fix the XGMII monitor to work correctly on raw bytes

eric-wieser:
This was already fixed for the driver in 11356124bbe5bd8b928ac1cf2b63d7dc12ecc6e2 (part of #1011).
Without this change, the calls to `hexdump` give a DeprecationWarning.

The change to the driver files is just for consistency, and should not affect behavior.

Not tested unfortunately, but hopefully more correct.

### DIFF
--- a/cocotb/monitors/xgmii.py
+++ b/cocotb/monitors/xgmii.py
@@ -40,11 +40,11 @@ from cocotb.utils import hexdump
 from cocotb.monitors import Monitor
 from cocotb.triggers import RisingEdge
 
-_XGMII_IDLE      = "\x07"  # noqa
-_XGMII_START     = "\xFB"  # noqa
-_XGMII_TERMINATE = "\xFD"  # noqa
+_XGMII_IDLE      = 0x07  # noqa
+_XGMII_START     = 0xFB  # noqa
+_XGMII_TERMINATE = 0xFD  # noqa
 
-_PREAMBLE_SFD = "\x55\x55\x55\x55\x55\x55\xD5"
+_PREAMBLE_SFD = b"\x55\x55\x55\x55\x55\x55\xD5"
 
 
 class XGMII(Monitor):
@@ -53,6 +53,10 @@ class XGMII(Monitor):
     Assumes a single vector, either 4 or 8 bytes plus control bit for each byte.
 
     If interleaved is ``True`` then the control bits are adjacent to the bytes.
+
+    .. versionchanged:: 1.4.0
+        This now emits packets of type :class:`bytes` rather than :class:`str`,
+        which matches the behavior of :class:`cocotb.drivers.xgmii.XGMII`.
     """
 
     def __init__(self, signal, clock, interleaved=True, callback=None,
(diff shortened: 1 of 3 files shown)


cocotb/monitors/xgmii.py
```diff
@@ -160,7 +164,7 @@ def _monitor_recv(self):
                 if preamble_sfd != _PREAMBLE_SFD:
                     self.log.error("Got a frame with unknown preamble/SFD")
                     self.log.error(hexdump(preamble_sfd))
-                    self._pkt = ""
+                    self._pkt = b""
                     continue
 
                 expected_crc = struct.pack("<I",
```
eric-wieser: The line below this one raised `TypeError` on python3 before this patch.

documentation/source/newsfragments/1545.bugfix.rst
```diff
@@ -0,0 +1 @@
+The :class:`~cocotb.monitors.xgmii.XGMII` monitor no longer crashes on Python 3, and now assembles packets as :class:`bytes` not :class:`str`. The :class:`~cocotb.drivers.xgmii.XGMII` driver has expected :class:`bytes` since cocotb 1.2.0.
```
cmarqu:
```suggestion
The :class:`~cocotb.monitors.xgmii.XGMII` monitor no longer crashes on Python 3, and now assembles packets as :class:`bytes`, not :class:`str`. The :class:`~cocotb.drivers.xgmii.XGMII` driver has expected :class:`bytes` since cocotb 1.2.0.
```
?

cocotb/monitors/xgmii.py
```diff
@@ -53,6 +53,10 @@ class XGMII(Monitor):
     Assumes a single vector, either 4 or 8 bytes plus control bit for each byte.
 
     If interleaved is ``True`` then the control bits are adjacent to the bytes.
+
+    .. versionchanged:: 1.4.0
```
cmarqu: Spelling out the ".0" hasn't been done so far IIRC, but may be good practice actually...

eric-wieser: Could go either way. The reality is we shouldn't be changing things in a way that needs documenting in bugfix releases anyway.

Perhaps worth a sweep through all of these for consistency before release.

documentation/source/newsfragments/1545.bugfix.rst
eric-wieser: You're right, but I'm not a fan of the chained commas there. Rephrasing to...
```suggestion
The :class:`~cocotb.monitors.xgmii.XGMII` monitor no longer crashes on Python 3, and now assembles packets as :class:`bytes` instead of :class:`str`. The :class:`~cocotb.drivers.xgmii.XGMII` driver has expected :class:`bytes` since cocotb 1.2.0.
```

LeChuck42:
`self._pkt` is initialized to `b""` (immutable), but we are using `.append()` to add payload data to it. This won't work. Maybe instead use a `bytearray()`?

eric-wieser:
Nice catch @LeChuck42, mind making a patch?

Did you find this by inspection, or by running into the bug?

LeChuck42:
Originally I tried running the monitor on 1.3.1 with Python 3.6, noting the issues with `zlib.crc32`. I then found this patch and tried it, finding the other problem.

As I'm still on 1.3.1 for the rest, I also had to patch the calls to `hexdump`, but this is already addressed in #1519

eric-wieser:
If you find the time, it would be very helpful if you could write a simple test for XGMII.

As is, we have no tests for this code, which is why it's still broken.

PR #1425, opened 2020-02-19
Compile common libraries to single `libs` directory 

themperek:
- common libraries are compiled once to `/cocotb/libs` directory which results in a faster compilation, less space and easier use 
- `vpi/vhpi/fli` are compiled with `_simulator_name` sufix (`.` has special (directory separation) meaning in setuptools) in same directory `/cocotb/libs`. 

- `simulator` module is compiled once and located in `/cocotb` folder. This allows using `form cocotb import simulator`  
-> `simulator` module can be added to documentation (need to install/compile) -> future PR.
-> can remove `COCOTB_SIM` -> future PR

- removed from makefiles `LD_LIBRARY_PATH` everywhere -> less potential collision with user environment settings

Resolves https://github.com/cocotb/cocotb/issues/1181

Checked with **ALL** supported (Linux) simulators.

PR #1551, opened 2020-03-31
Add CUSTOM_*_DEPS as dependencies where missing

### LINKED ISSUE #1549: Support CUSTOM_COMPILE_DEPS and CUSTOM_SIM_DEPS for ghdl
We are missing support for ``CUSTOM_COMPILE_DEPS`` and ``CUSTOM_SIM_DEPS`` in ``Makefile.ghdl``.
See https://github.com/cocotb/cocotb/issues/1539#issuecomment-605706516.

cmarqu:
Add support for ``CUSTOM_COMPILE_DEPS`` and ``CUSTOM_SIM_DEPS`` in ``Makefile.ghdl``, ``Makefile.nvc``.
Add support for ``CUSTOM_COMPILE_DEPS`` in ``Makefile.questa`` and ``Makefile.aldec``.

Closes #1549.


eric-wieser:
Will this conflict with #1425? If so, I'd rather get that one in first.

cmarqu:
I don't think it would conflict, or in case it does, it would require reworking of the concept as such for *all* simulators.

(I only now realized that ``CUSTOM_SIM_DEPS`` is used to be sensitive to changes in the Python files, so that means that without this PR, ghdl and nvc *aren't* sensitive.)

themperek:
A bit related to https://github.com/cocotb/cocotb/pull/1413

themperek:
There is some inconsistency in how all this is done for aldec and questa.  Maybe you can make it consistent where it is applied?

cmarqu:
> There is some inconsistency in how all this is done for aldec and questa. Maybe you can make it consistent where it is applied?

I'd prefer not to move these dependencies to a different target, and I can't test it for Aldec.

PR #1553, opened 2020-04-01
Guard against user setting vars we append to

### LINKED ISSUE #1552: Document that users should use EXTRA_ARGS=... make instead of make EXTRA_ARGS=...
When we tell users to use ``SIM_ARGS`` or ``EXTRA_ARGS`` etc. to add their options and expect that this works for them on the command line like make ``SIM_ARGS=-whatever``, we cannot append to these variables in our ``Makefile.<simulator>``, because only the variable values given on the command line would be set (https://stackoverflow.com/a/32696474).

This will be relevant for the documentation part of #1495.

Affected: 
* ``EXTRA_ARGS``  in ``Makefile.xcelium``/``Makefile.ius``, ``Makefile.vcs``
* ``COMPILE_ARGS`` in ``Makefile.verilator``
* probably more

cmarqu:
Do not directly append to ``COMPILE_ARGS``, ``SIM_ARGS``, ``EXTRA_ARGS`` etc., because if a user sets these variables on the make command line, only the user's variable values will be seen.
Instead, append to local variables.

Fixes #1552

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.icarus
+++ b/cocotb/share/makefiles/simulators/Makefile.icarus
@@ -54,11 +54,20 @@ endif
 
 BUILD_VPI=1
 
-COMPILE_ARGS += -g2012 # Default to latest SystemVerilog standard
+# Do not directly append to COMPILE_ARGS, SIM_ARGS, EXTRA_ARGS etc.,
+# because if a user sets these variables on the make command line,
+# only the user's variable values will be seen.
+# Instead, append to local variables.
+_COMPILE_ARGS += $(COMPILE_ARGS)
+_SIM_ARGS     += $(SIM_ARGS)
+_EXTRA_ARGS   += $(EXTRA_ARGS)
+_PLUSARGS     += $(PLUSARGS)
+
+_COMPILE_ARGS += -g2012 # Default to latest SystemVerilog standard
 
 # Compilation phase
 $(SIM_BUILD)/sim.vvp: $(SIM_BUILD) $(VERILOG_SOURCES) $(CUSTOM_COMPILE_DEPS)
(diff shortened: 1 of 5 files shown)


cocotb/share/makefiles/simulators/Makefile.icarus
```diff
@@ -54,11 +54,20 @@ endif
...
+# Do not directly append to COMPILE_ARGS, SIM_ARGS, EXTRA_ARGS etc.,
+# because if a user sets these variables on the make command line,
+# only the user's variable values will be seen.
+# Instead, append to local variables.
+_COMPILE_ARGS += $(COMPILE_ARGS)
+_SIM_ARGS     += $(SIM_ARGS)
+_EXTRA_ARGS   += $(EXTRA_ARGS)
+_PLUSARGS     += $(PLUSARGS)
+
+_COMPILE_ARGS += -g2012 # Default to latest SystemVerilog standard
```
eric-wieser, line 66: This is exactly the type of case where we _do_ want the user to be able to override, as otherwise they have no way to replace this.

cmarqu: I'd say that *this* option should not be in this simulator Makefile at all, like all other non-general assumptions.
But I now realized that the approach of this PR does not hold up well at all in the case where you would try to set this option in the Makefile of the testcase, which you'd do with ``COMPILE_ARGS``, and if you then try to append to it via ``make COMPILE_ARGS=-foo``, you are back to square one. So indeed the only solution is to educate users about setting this as an environment variable.

cmarqu:
This PR does not hold up well at all in the case where you would try to set an option in the Makefile of the testcase, which you'd do with ``COMPILE_ARGS``, and if you then try to append to it via ``make COMPILE_ARGS=-foo``, you are back to square one.
So the only solution is to educate users about setting this as an environment variable.
Closing this.

PR #1554, opened 2020-04-01
Improve test_discovery

cmarqu:
Improve some expect_error; add better debug output for fails.

### DIFF
--- a/tests/test_cases/test_discovery/test_discovery.py
+++ b/tests/test_cases/test_discovery/test_discovery.py
@@ -163,15 +163,15 @@ def access_integer(dut):
     yield Timer(10)
     test_int = dut.stream_in_int
     if not isinstance(test_int, IntegerObject):
-        raise TestFailure("dut.stream_in_int is not an integer")
+        raise TestFailure("dut.stream_in_int is not an integer but {} instead".format(type(test_int)))
 
     try:
         bit = test_int[3]
     except IndexError as e:
         tlog.info("Access to bit is an error as expected")
-        bitFail = True
+        bitfail = True
 
-    if not bitFail:
+    if not bitfail:
         raise TestFailure("Access into an integer should be invalid")
 
     length = len(test_int)
@@ -205,7 +205,7 @@ def access_string_vhdl(dut):
     """Access to a string, both constant and signal."""
     tlog = logging.getLogger("cocotb.test")
     yield Timer(10)
-    constant_string = dut.isample_module1.EXAMPLE_STRING;
+    constant_string = dut.isample_module1.EXAMPLE_STRING
     tlog.info("%r is %s" % (constant_string, str(constant_string)))
     if not isinstance(constant_string, ConstantObject):
         raise TestFailure("EXAMPLE_STRING was not constant")
@@ -265,7 +265,7 @@ def access_string_vhdl(dut):
 
     test_string = test_string.upper()
 
-    result = str(variable_string);
+    result = str(variable_string)
     tlog.info("After setting bytes of string value is %s" % result)
     if variable_string != test_string:
         raise TestFailure("%r %s != '%s'" % (variable_string, result, test_string))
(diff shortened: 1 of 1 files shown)


tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -274,48 +274,45 @@ def access_string_vhdl(dut):
...
     tlog.info("Modifying const string")
-    string_const <= "MODIFIED"
-    yield Timer(10)
-    string_const = dut.STRING_CONST;
-    if string_const != "TESTING_CONST":
-        raise TestFailure("STRING_CONST was not still \'TESTING_CONST\'")
+    string_const <= b"MODIFIED"
+    yield Timer(10, 'ns')
+    if string_const != b"TESTING_CONST":
+        raise TestFailure("STRING_CONST was not still b\'TESTING_CONST\' after modification but {} instead".format(string_const))
```
eric-wieser, line 294: This duplicates #1381 I assume, which should be merged first.

cmarqu: Yes, I'll keep this as a draft until it is in.

eric-wieser:
Probably would be good to rebase this now that that PR is in. CI failure is transient, so a rerun after the rebase should fix it.

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -154,7 +154,7 @@ def access_single_bit_erroneous(dut):
     dut.stream_in_data[bit] <= 1
     yield Timer(10)
 
-@cocotb.test(expect_error=cocotb.SIM_NAME.lower().startswith(("icarus")),
+@cocotb.test(expect_error=cocotb.SIM_NAME.lower().startswith(("icarus", "chronologic simulation vcs")),
```
eric-wieser, line 157: SIM_NAME can be multiple words?

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -274,48 +274,45 @@ def access_string_vhdl(dut):
...
 @cocotb.test(skip=cocotb.LANGUAGE in ["vhdl"] or cocotb.SIM_NAME.lower().startswith("icarus"),
-             expect_fail=cocotb.SIM_NAME.lower().startswith("modelsim"),
-             expect_error=cocotb.SIM_NAME.startswith(("xmsim", "ncsim")))
+             expect_error=cocotb.result.TestFailure if cocotb.SIM_NAME.lower().startswith(("xmsim", "ncsim", "modelsim", "chronologic simulation vcs")) else False)
 def access_const_string_verilog(dut):
     """Access to a const Verilog string."""
     tlog = logging.getLogger("cocotb.test")
-
-    yield Timer(10)
     string_const = dut.STRING_CONST;
```
eric-wieser, line 281:
```suggestion
    string_const = dut.STRING_CONST
```

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -274,48 +274,45 @@ def access_string_vhdl(dut):
...
+
+    yield Timer(10, 'ns')
     tlog.info("%r is %s" % (string_const, str(string_const)))
     if not isinstance(string_const, StringObject):
         raise TestFailure("STRING_CONST was not StringObject")
-    if string_const != "TESTING_CONST":
-        raise TestFailure("STRING_CONST was not == \'TESTING_CONST\'")
+    if string_const != b"TESTING_CONST":
+        raise TestFailure("Initial value of STRING_CONST was not == b\'TESTING_CONST\' but {} instead".format(string_const))
```
eric-wieser, line 288: Or perhaps:
```suggestion
    expected = b"TESTING_CONST"
    if string_const != expected:
        raise TestFailure("Initial value of STRING_CONST was not == {} but {} instead".format(expected, string_const))
```

tests/test_cases/test_discovery/test_discovery.py
cmarqu: Apparently yes.

tests/test_cases/test_discovery/test_discovery.py
cmarqu: I started this change but then discovered that there are quite a number of such duplications (e.g. also for object types) in the test so that I'll skip this for now.
If ``expected`` would be used a third time I would not hesitate to make this kind of change, but the additionally required mental backtracking for the reader is diminishing the positive keeping-in-sync effect here IMO.

tests/test_cases/test_discovery/test_discovery.py
cmarqu: Done.

tests/test_cases/test_discovery/test_discovery.py
eric-wieser: Fair enough. This is where the pytest assertion rewriters really provide value.

PR #1555, opened 2020-04-01
Advertise user survey

imphil:
We are having our user survey. Advertise it as widely as we can to make
users aware of it. The project relies on getting enough answers to steer
development efforts into the right directions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,8 @@
+**Your input is needed!** Please help out by taking 10 minutes to fill out this year’s cocotb user survey. This survey gives the development community important feedback to steer the future of cocotb into the right direction for your use case.
+[**Take the cocotb user survey now**](https://docs.google.com/forms/d/e/1FAIpQLSfD36PldzszbuZjysss3AMvxkf6XCtSbDTVh9qVNNYDaHTZ_w/viewform).
+
+---
+
 **cocotb** is a coroutine based cosimulation library for writing VHDL and Verilog testbenches in Python.
 
 [![Documentation Status](https://readthedocs.org/projects/cocotb/badge/?version=latest)](http://cocotb.readthedocs.org/en/latest/)
--- a/documentation/source/_templates/layout.html
+++ b/documentation/source/_templates/layout.html
@@ -1,0 +1,14 @@
+{% extends "!layout.html" %}
+
+{% block document %}
+
+<div class="admonition info">
+  <p>The cocotb user survey 2020 is awaiting your answers.
+  Please take 10 minutes to <strong>help cocotb developers understand your use of cocotb better, and influence the direction of the project.</strong></p>
+  <div class="wy-text-right">
+    <a href="https://docs.google.com/forms/d/e/1FAIpQLSfD36PldzszbuZjysss3AMvxkf6XCtSbDTVh9qVNNYDaHTZ_w/viewform" class="btn btn-success">Take the survey now!</a>
+  </div>
+</div>
+
+{{ super() }}
+{% endblock %}


documentation/source/_templates/layout.html
```diff
@@ -0,0 +1,12 @@
+{% extends "!layout.html" %}
+
+{% block content %}
```
eric-wieser, line 3: Tempting to use `document` instead of `content` so this ends up below the breadcrumb

README.md
```diff
@@ -1,3 +1,8 @@
+**Your input is needed!** Please help out by taking 10 minutes to fill out this year’s cocotb user survey. This survey gives the development community important feedback to steer future of cocotb into the right direction for your use case.
```
eric-wieser, line 1:
```suggestion
**Your input is needed!** Please help out by taking 10 minutes to fill out this year’s cocotb user survey. This survey gives the development community important feedback to steer the future of cocotb into the right direction for your use case.
```

documentation/source/_templates/layout.html
```diff
@@ -0,0 +1,12 @@
...
+
+<div style="background-color: rgb(255, 248, 225); color: #444; padding: 1em; border: 1px solid rgb(255, 111, 0); margin-top: 5px; margin-bottom: 5px;">
+  The <a href="https://docs.google.com/forms/d/e/1FAIpQLSfD36PldzszbuZjysss3AMvxkf6XCtSbDTVh9qVNNYDaHTZ_w/viewform">cocotb user survey 2020</a> is awaiting your answers.
+  Please take 10 minutes to <b>help cocotb developers understand your use of cocotb better, and influence the direction of the project.</b>
+  <b><a href="https://docs.google.com/forms/d/e/1FAIpQLSfD36PldzszbuZjysss3AMvxkf6XCtSbDTVh9qVNNYDaHTZ_w/viewform">Click here to take the survey now!</a></b>
+</div>
```
eric-wieser, line 9: After putting this in `content` as I suggest above, you can piggy back on RST styles
```suggestion
<div class="admonition primary" style="display: table">
  <p>The cocotb user survey 2020 is awaiting your answers.
  Please take 10 minutes to help cocotb developers understand your use of cocotb better, and influence the direction of the project.</p>
  <a href="https://docs.google.com/forms/d/e/1FAIpQLSfD36PldzszbuZjysss3AMvxkf6XCtSbDTVh9qVNNYDaHTZ_w/viewform" class="btn btn-primary float-right">Take the survey now!</a>
</div>
```

eric-wieser: image

cmarqu:
Do we want to run this until we have a new version of the survey (next year), or close it after a few weeks? Maybe we should tell the users whether they have a deadline or not.

documentation/source/_templates/layout.html
eric-wieser: (feel free to add back the bold as needed)

eric-wieser: Or:
```suggestion
<div class="admonition info" style="display: table">
  <p class="admonition-title">The cocotb user survey 2020</p>
<p>Please take 10 minutes to help cocotb developers understand your use of cocotb better, and influence the direction of the project.</p>

  <a href="https://docs.google.com/forms/d/e/1FAIpQLSfD36PldzszbuZjysss3AMvxkf6XCtSbDTVh9qVNNYDaHTZ_w/viewform" class="btn btn-info float-right">Take the survey now!</a>
</div>
```
image

imphil:
Thanks for your ideas! I've updated the PR to use blue instead of yellow and place the banner under the breadcrumb nav. The button is green to at least slightly attention-grabbing, I don't want it to look too much like normal docs content.

imphil:
Ah, @cmarqu re duration: My plan is to leave it online for ~4 weeks or so and then close the survey. I didn't put an explicit end date on it to give us a bit more flexibility to react depending on how participation turns out to be.

eric-wieser:
Looks like we have a reasonable number of responses already

ktbarrett:
The Python 3.5 regression fails because of the non-ASCII apostrophe use in the README.

imphil:
Good catch, thanks! (I'm pretty sure I checked and thought the Python 3.5 failure was present before, which is actually not the case. Thanks for your investigation.) Opened #1556 to fix.

PR #1556, opened 2020-04-02
Remove non-ASCII character from README.md

imphil:
It breaks CI on Python 3.5.


eric-wieser:
Note that if you _want_ a straight quote here, you can still merge this - but we should merge #1557 anyway, and if we do, can keep our curly quote here if we want it.

imphil:
Thanks, you beat me to it! Let's merge #1557 first and then still merge this one. Non-ASCII characters in the README will certainly cause other trouble down the road, there's no good reason for them in there.

eric-wieser:
> there's no good reason for them in there

I'd perhaps argue there's no reason for them _not_ to be in there, python is almost completely utf8 at this point. Still, I'm not going to argue about a few pixels on a `'` :).

Merge or close as you wish.

imphil:
@eric-wieser can you approve this PR please?

PR #1557, opened 2020-04-03
Ensure source files are read as utf8, rather than a system-specific encoding.

eric-wieser:
We authored the files being read here, so we know their encoding in advance.
This fixes a CI failure in Python 3.5.

Also switch to use a `with` statement to ensure the file is closed a little sooner.

Fixes the underlying issue worked around in gh-1556.


imphil:
Thanks!

PR #1558, opened 2020-04-03
More helpful Python version check in setup.py

imphil:
In setup.py we check if Python is sufficiently recent. For users running
the installation from Python 2 this can be a blocker to install (and
subsequently use) cocotb, we therefore need to invest into making it
easy for users to diagnose and fix the problem. In many cases Python 3
is available on a user's system, but not used, commonly due to the use
of 'pip' instead of 'pip3'. Explicitly add a hint in this regard and
point to the documentation for more details.

To make the error message even more friendly, remove the backtrace
generated as part of `RuntimeError()` and use normal stderr instead. To
be able to support older Python 2 versions we don't use `print` but
`sys.stderr.write()`.

On Ubuntu 18.04/2.7.17:

Before:

```
philipp@philipp-VirtualBox:~/src/cocotb$ pip install .
Processing /home/philipp/src/cocotb
    Complete output from command python setup.py egg_info:
      File "/tmp/pip-kZcdoF-build/setup.py", line 33, in <module>
        raise RuntimeError("Python version >= 3.5 required.")
    RuntimeError: Python version >= 3.5 required.

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-kZcdoF-build/
```

after:

```
philipp@philipp-VirtualBox:~/src/cocotb$ pip install .
Processing /home/philipp/src/cocotb
    Complete output from command python setup.py egg_info:
    This version of cocotb requires at least Python 3.5 to run.
    If you have Python 3 installed on your machine try
    using 'pip3' instead of 'pip' to install cocotb.
    For more information please refer to the documentation at
    https://cocotb.readthedocs.io.

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-bd_QP1-build/
```


setup.py
```diff
@@ -30,7 +30,20 @@
 
 import sys
 if sys.version_info[:2] < (3, 5):
-    raise RuntimeError("Python version >= 3.5 required.")
+    msg = ["This version of cocotb requires at least Python 3.5 to run."]
```
themperek, line 33: One could display what is the version installed.

setup.py
```diff
@@ -30,7 +30,20 @@
...
+    if sys.version_info[0] == 2:
+        msg += [
+            "If you have Python 3 installed on your machine try ",
+            "using 'pip3' instead of 'pip' to install cocotb."
+        ]
```
eric-wieser, line 38: Perhaps best here to recommend `python3 -m pip`, if `pip` was pointing at the wrong thing

setup.py
```diff
@@ -30,7 +30,20 @@
...
+    msg += [
+        "For more information please refer to the documentation at ",
+        "https://cocotb.readthedocs.io."
+    ]
+
+    # The print() function is not available in older Python versions.
+    sys.stderr.write('\n'.join(msg))
+    sys.exit(1)
```
eric-wieser, line 46:
```suggestion
    raise SystemExit('\n'.join(msg))
```

imphil:
Addressed all review comments.

Note: I didn't use `sys.version` as it gives a rather long string, e.g. 


```
Python 2.7.17 (default, Nov  7 2019, 10:07:09)
    [GCC 7.4.0].
```

PR #1560, opened 2020-04-04
Better error message when missing COCOTB_RESULTS_FILE and cleaner log

themperek:
Better error message when missing `COCOTB_RESULTS_FILE` and cleaner log

### DIFF
--- a/cocotb/share/makefiles/Makefile.sim
+++ b/cocotb/share/makefiles/Makefile.sim
@@ -131,6 +131,11 @@ PWD := $(shell pwd)
 
 include $(COCOTB_SHARE_DIR)/makefiles/simulators/Makefile.$(SIM_LOWERCASE)
 
+# Check that the COCOTB_RESULTS_FILE was created, since we can't set an exit code from cocotb.
+define check_for_results_file
+    @test -f $(COCOTB_RESULTS_FILE) || (echo "ERROR: $(COCOTB_RESULTS_FILE) was not written by the simulation!" >&2 && exit 1)
+endef
+
 # Append directory containing the cocotb Python package to PYTHONPATH
 # XXX: This is only needed if cocotb is not installed into the default
 # Python search path.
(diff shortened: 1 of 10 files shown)


cocotb/share/makefiles/Makefile.sim
```diff
@@ -131,6 +131,10 @@ PWD := $(shell pwd)
 
 include $(COCOTB_SHARE_DIR)/makefiles/simulators/Makefile.$(SIM_LOWERCASE)
 
+define check_for_results_file
+    @test -f $(COCOTB_RESULTS_FILE) || (echo "ERROR: $(COCOTB_RESULTS_FILE) does not exist!" >&2 && exit 1)
```
eric-wieser, line 135:
```suggestion
    @test -f $(COCOTB_RESULTS_FILE) || (echo "ERROR: $(COCOTB_RESULTS_FILE) was not written by the simulation!" >&2 && exit 1)
```
Otherwise the user might think it was supposed to be there already.

cocotb/share/makefiles/simulators/Makefile.ghdl
```diff
@@ -72,8 +72,7 @@ $(COCOTB_RESULTS_FILE): analyse  $(COCOTB_LIBS) $(COCOTB_VPI_LIB)
         TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) TOPLEVEL_LANG=$(TOPLEVEL_LANG) COCOTB_SIM=1 \
 	$(CMD) -r $(GHDL_ARGS) --work=$(RTL_LIBRARY) $(TOPLEVEL) --vpi=$(LIB_DIR)/libcocotbvpi_ghdl.$(LIB_EXT) $(SIM_ARGS) $(PLUSARGS)
 
-	# check that the file was actually created, since we can't set an exit code from cocotb
```
eric-wieser, line 75: Can we copy the last part of this comment to the function definition?

Issue #494, opened 2016-10-21
Integrate pytest


javValverde:
First of all, thanks for this amazing tool. After using it, I cannot think of any other way in which I could write testbenches.

I think it is a great idea if cocotb uses pytest as the testing framework. It supports different kinds of setup/teardowns, parameterization, can generate junit results etc. As far as I know, it is also capable of running tests in unittest and nose formats.

I could contribute in this feature, but I don't really know how to start.

themperek:
https://github.com/potentialventures/cocotb/issues/280

eric-wieser:
Turns out you can already do this right now (!):
```python
import cocotb
import pytest
import decorator
import concurrent.futures
import Queue

class CocotbExecutor:
    def __init__(self):
        self.queue = Queue.Queue()

    def submit(self, f, *args, **kwargs):
        fut = concurrent.futures.Future()
        @cocotb.coroutine
        def wrapper():
            try:
                ret = yield f(*args, **kwargs)
            except Exception as e:
                fut.set_exception(e)
            else:
                fut.set_result(ret)
            yield Timer(1)
        self.queue.put(wrapper())
        return fut

ex = CocotbExecutor()


@decorator.decorator
def coroutine_test(f, *args, **kwargs):
    return ex.submit(cocotb.coroutine(f), *args, **kwargs).result()

@cocotb.test()
def the_pytest_hack(dut):
    tpe = concurrent.futures.ThreadPoolExecutor(1)
    the_tests = tpe.submit(pytest.main, ['--pyargs', __name__, '-v', '-s'])
    while True:
        try:
            coro = ex.queue.get(timeout=0.5)
        except Queue.Empty:
            if the_tests.done():
                break
        else:
            yield coro
    assert the_tests.result() == 0


@pytest.fixture(scope='module')
def dut():
    return cocotb.regression._dut
```

Once you have that in place, you can write tests (in that file) the pytest way:

```
@coroutine_test
def test_foo(dut)
    yield RisingEdge(dut.clk)
    assert True

@coroutine_test
def test_bar(dut)
    yield RisingEdge(dut.clk)
    assert False
```

You'll get a pytest session embedded within your cocotb one

ktbarrett:
xref #667, #1062.

ktbarrett:
@eric-wieser I think we can replace the regression manager with pytest using a somewhat similar approach. This depends on #791.

[...]

Issue #437, opened 2016-04-21
Replace makefiles through Python


Paebbels:
Hello, 

we from PoC have integrated Cocotb with QuestaSim back-end as another simulator into our list of supported tool chains. But there are some issues.

Using make files is a bit counterproductive when being platform independent. Moreover, these files are very hard to maintain. We gathered several workaround.

While reading and understanding these files, we figured out that more then 50% of the make file code is sequential code using Bash-like code just in make file format and using target rules as sub-programs.

Wouldn't it be better to put all this code into a clever Python run? Controlling simulators from Python is much easier. gcc runs only once for a hand full of files.

Maybe we can share some Python code to do these build jobs.

stuarthodgson:
@Paebbels 

When you mention work arounds where needed can you share these? they could be of use to others.

From our point of view we do not need to maintain the Makefiles as they just work for us. Which I accept may not be the case for everyone.

The problem with any build system is that it always has detractors, nothing is perfect with any of them particularity custom solutions that may appear better at first glance.

Please do share the python code that you have developed but i'm not convinced yet there is a need to make what would be a pretty sweeping change, I am open to being convinced though.

Thanks

Stu

Paebbels:
Hello Stuart,

I'll test my workarounds and improvements. If it works out, I'll create a pull-request for your repository.

You can find our Python code in the [`py/`](https://github.com/VLSI-EDA/PoC/tree/release/py) sub-directory. It's not perfect nor fully decoupled, but it's a major step in the direction of OOP compared to the initial version of our Python infrastructure behind our IP cores.

Major parts:
  - base classes for simulators, compilers, configurations, tool-chains...
  - one simulator package per supported simulator
      - Active-HDL
      - [Cocotb (with QuestaSim backend)](https://github.com/VLSI-EDA/PoC/blob/release/py/Simulator/CocotbSimulator.py)
      - GHDL (with mcode, gcc or llvm backend)
      - [QuestaSim](https://github.com/VLSI-EDA/PoC/blob/release/py/Simulator/QuestaSimulator.py)
      - ISE Simulator
      - Vivado Simulator
  - one compiler package per supported (synthesis) compiler
      - Altera Quartus II
      - Lattice Diamond LSE
      - Xilinx CoreGen
      - Xilinx XST
  - one ToolChain package for each tool-chain
      - Aldec_ActiveHDL
      - Altera_Quartus
      - GHDL_GTKWave
      - Lattice_Diamond
[...]

PR #1562, opened 2020-04-04
Python-based runtime

themperek:
Python-based runtime based on https://github.com/themperek/cocotb-test/blob/master/cocotb_test/simulator.py see also: [cocotb-test slides](https://github.com/cocotb/cocotb/files/3908378/cocotb-test_WOSDV2019.pdf)

This is just something we can start.

**Some issues/questions:**
- How should we handle `results.xml` (multiple tests in the same director/parallel runs?, handed by pytest plugin?)
- How to handle logging (can we forward everything to the top?)
- ... many more

Example usage (for example using pytest):
```python
from cocotb.run import run

@pytest.mark.skipif(os.getenv("SIM") == "ghdl", reason="Verilog not suported")
@pytest.mark.parametrize("module_name", ["test_deprecated", "test_doctests"])
def test_cocotb(module_name):
    run(
        verilog_sources=["sample_module.sv"],
        toplevel="sample_module",
        module=module_name,
    )
```

### DIFF
--- a/cocotb/run.py
+++ b/cocotb/run.py
@@ -1,0 +1,929 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+import subprocess
+import os
+import sys
+import tempfile
+import re
+import cocotb
+import logging
+import shutil
+from xml.etree import cElementTree as ET
+import threading
+import signal
+import cocotb._vendor.find_libpython as find_libpython
+import cocotb.config
+
+from distutils.spawn import find_executable
+from distutils.sysconfig import get_config_var
+
+_magic_re = re.compile(r"([\\{}])")
+_space_re = re.compile(r"([\s])", re.ASCII)
+
+
+def as_tcl_value(value):
+    # add '\' before special characters and spaces
+    value = _magic_re.sub(r"\\\1", value)
+    value = value.replace("\n", r"\n")
+    value = _space_re.sub(r"\\\1", value)
+    if value[0] == '"':
+        value = "\\" + value
+
+    return value
+
+
+class Simulator(object):
+    def __init__(
+        self,
+        toplevel,
+        module,
+        work_dir=None,
+        python_search=None,
+        toplevel_lang="verilog",
+        verilog_sources=None,
+        vhdl_sources=None,
+        includes=None,
+        defines=None,
+        parameters=None,
+        compile_args=None,
+        sim_args=None,
+        extra_args=None,
+        plus_args=None,
+        force_compile=False,
+        testcase=None,
+        sim_build="sim_build",
+        seed=None,
+        extra_env=None,
+        compile_only=False,
+        waves=None,
+        gui=False
+    ):
+
+        self.logger = logging.getLogger()
+
+        self.sim_dir = os.path.abspath(sim_build)
+        os.makedirs(self.sim_dir, exist_ok=True)
+
+        self.lib_dir = os.path.join(os.path.dirname(cocotb.__file__), "libs")
+
+        self.lib_ext = "so"
+        if os.name == "nt":
+            self.lib_ext = "dll"
+
+        if isinstance(module, str):
+            self.module = module
+        else:
+            self.module = ",".join(module)
+
+        self.work_dir = self.sim_dir
+
+        if work_dir is not None:
+            absworkdir = os.path.abspath(work_dir)
+            if os.path.isdir(absworkdir):
+                self.work_dir = absworkdir
+
+        if python_search is None:
+            python_search = []
+
+        self.python_search = python_search
+
+        self.toplevel = toplevel
+        self.toplevel_lang = toplevel_lang
+
+        if verilog_sources is None:
+            verilog_sources = []
+
+        self.verilog_sources = self.get_abs_paths(verilog_sources)
+
+        if vhdl_sources is None:
+            vhdl_sources = []
+
+        self.vhdl_sources = self.get_abs_paths(vhdl_sources)
+
+        if includes is None:
+            includes = []
+
+        self.includes = self.get_abs_paths(includes)
+
+        if defines is None:
+            defines = []
+
+        self.defines = defines
+
+        if parameters is None:
+            parameters = {}
+
+        self.parameters = parameters
+
+        if compile_args is None:
+            compile_args = []
+
+        if extra_args is None:
+            extra_args = []
+
+        self.compile_args = compile_args + extra_args
+
+        if sim_args is None:
+            sim_args = []
+
+        self.sim_args = sim_args + extra_args
+
+        if plus_args is None:
+            plus_args = []
+
+        self.plus_args = plus_args
+        self.force_compile = force_compile
+        self.compile_only = compile_only
+
+        # by copy since we modify
+        self.env = dict(extra_env) if extra_env is not None else {}
+
+        if testcase is not None:
+            self.env["TESTCASE"] = testcase
+
+        if seed is not None:
+            self.env["RANDOM_SEED"] = str(seed)
+
+        if waves is None:
+            self.waves = bool(int(os.getenv("WAVES", 0)))
+        else:
+            self.waves = bool(waves)
+
+        self.gui = gui
+
+        # Catch SIGINT and SIGTERM
+        self.old_sigint_h = signal.getsignal(signal.SIGINT)
+        self.old_sigterm_h = signal.getsignal(signal.SIGTERM)
+
+        # works only if main thread
+        if threading.current_thread() is threading.main_thread():
+            signal.signal(signal.SIGINT, self.exit_gracefully)
+            signal.signal(signal.SIGTERM, self.exit_gracefully)
+
+        self.process = None
+
+    def set_env(self):
+
+        for e in os.environ:
+            self.env[e] = os.environ[e]
+
+        self.env["LIBPYTHON_LOC"] = find_libpython.find_libpython()
+
+        self.env["PATH"] += os.pathsep + self.lib_dir
+
+        self.env["PYTHONPATH"] = os.pathsep.join(sys.path)
+        for path in self.python_search:
+            self.env["PYTHONPATH"] += os.pathsep + path
+
+        self.env["PYTHONHOME"] = get_config_var("prefix")
+
+        self.env["TOPLEVEL"] = self.toplevel
+        self.env["MODULE"] = self.module
+
+        if not os.path.exists(self.sim_dir):
+            os.makedirs(self.sim_dir)
+
+    def build_command(self):
+        raise NotImplementedError()
+
+    def run(self):
+
+        sys.tracebacklimit = 0  # remove not needed traceback from assert
+
+        # TODO: This probably need to be unique for multiple/parallel runs
+        results_xml_file = os.getenv("COCOTB_RESULTS_FILE",os.path.join(self.sim_dir ,"results.xml"))
+
+        cmds = self.build_command()
+        self.set_env()
+        self.execute(cmds)
+
+        if not self.compile_only:
+            results_file_exist = os.path.isfile(results_xml_file)
+            assert results_file_exist, "Simulation terminated abnormally. Results file not found."
+
+            tree = ET.parse(results_xml_file)
+            for ts in tree.iter("testsuite"):
+                for tc in ts.iter("testcase"):
+                    for failure in tc.iter("failure"):
+                        assert False, '{} class="{}" test="{}" error={}'.format(
+                            failure.get("message"), tc.get("classname"), tc.get("name"), failure.get("stdout"),
+                        )
+
+        print("Results file: %s" % results_xml_file)
+        return results_xml_file
+
+    def get_include_commands(self, includes):
+        raise NotImplementedError()
+
+    def get_define_commands(self, defines):
+        raise NotImplementedError()
+
+    def get_parameter_commands(self, parameters):
+        raise NotImplementedError()
+
+    def get_abs_paths(self, paths):
+        paths_abs = []
+        for path in paths:
+            if os.path.isabs(path):
+                paths_abs.append(os.path.abspath(path))
+            else:
+                paths_abs.append(os.path.abspath(os.path.join(os.getcwd(), path)))
+
+        return paths_abs
+
+    def execute(self, cmds):
+        self.set_env()
+        for cmd in cmds:
+            self.logger.info("Running command: " + " ".join(cmd))
+
+            # TODO: create at thread to handle stderr and log as error?
+            # TODO: log forwarding
+
+            process = subprocess.run(cmd, cwd=self.work_dir, env=self.env)
+
+            if process.returncode != 0:
+                raise RuntimeError(
+                    "Process '%s' termindated with error %d"
+                    % (process.args[0], process.returncode)
+                )
+
+    def outdated(self, output, dependencies):
+
+        if not os.path.isfile(output):
+            return True
+
+        output_mtime = os.path.getmtime(output)
+
+        dep_mtime = 0
+        for file in dependencies:
+            mtime = os.path.getmtime(file)
+            if mtime > dep_mtime:
+                dep_mtime = mtime
+
+        if dep_mtime > output_mtime:
+            return True
+
+        return False
+
+    def exit_gracefully(self, signum, frame):
+        pid = None
+        if self.process is not None:
+            pid = self.process.pid
+            self.process.stdout.flush()
+            self.process.kill()
+            self.process.wait()
+        # Restore previous handlers
+        signal.signal(signal.SIGINT, self.old_sigint_h)
+        signal.signal(signal.SIGTERM, self.old_sigterm_h)
+        assert False, "Exiting pid: {} with signum: {}".format(str(pid), str(signum))
+
+
+class Icarus(Simulator):
+    def __init__(self, *argv, **kwargs):
+        super(Icarus, self).__init__(*argv, **kwargs)
+
+        if self.vhdl_sources:
+            raise ValueError("This simulator does not support VHDL")
+
+        self.sim_file = os.path.join(self.sim_dir, self.toplevel + ".vvp")
+
+    def get_include_commands(self, includes):
+        include_cmd = []
+        for dir in includes:
+            include_cmd.append("-I")
+            include_cmd.append(dir)
+
+        return include_cmd
+
+    def get_define_commands(self, defines):
+        defines_cmd = []
+        for define in defines:
+            defines_cmd.append("-D")
+            defines_cmd.append(define)
+
+        return defines_cmd
+
+    def get_parameter_commands(self, parameters):
+        parameters_cmd = []
+        for name, value in parameters.items():
+            parameters_cmd.append("-P")
+            parameters_cmd.append(self.toplevel + "." + name + "=" + str(value))
+
+        return parameters_cmd
+
+    def compile_command(self):
+
+        cmd_compile = (
+            ["iverilog", "-o", self.sim_file, "-D", "COCOTB_SIM=1", "-s", self.toplevel, "-g2012"]
+            + self.get_define_commands(self.defines)
+            + self.get_include_commands(self.includes)
+            + self.get_parameter_commands(self.parameters)
+            + self.compile_args
+            + self.verilog_sources
+        )
+
+        return cmd_compile
+
+    def run_command(self):
+        return (
+            ["vvp", "-M", self.lib_dir, "-m", cocotb.config.lib_name("vpi", "icarus")]
+            + self.sim_args
+            + [self.sim_file]
+            + self.plus_args
+        )
+
+    def build_command(self):
+        if self.waves:
+            dump_mod_name = "iverilog_dump"
+            dump_file_name = self.toplevel+".fst"
+            dump_mod_file_name = os.path.join(self.sim_dir, dump_mod_name+".v")
+
+            if not os.path.exists(dump_mod_file_name):
+                with open(dump_mod_file_name, 'w') as f:
+                    f.write("module iverilog_dump();\n")
+                    f.write("initial begin\n")
+                    f.write("    $dumpfile(\"%s\");\n" % dump_file_name)
+                    f.write("    $dumpvars(0, %s);\n" % self.toplevel)
+                    f.write("end\n")
+                    f.write("endmodule\n")
+
+            self.verilog_sources.append(dump_mod_file_name)
+            self.compile_args.extend(["-s", dump_mod_name])
+            self.plus_args.append("-fst")
+
+        cmd = []
+        if self.outdated(self.sim_file, self.verilog_sources) or self.force_compile:
+            cmd.append(self.compile_command())
+        else:
+            self.logger.warning("Skipping compilation:" + self.sim_file)
+
+        # TODO: check dependency?
+        if not self.compile_only:
+            cmd.append(self.run_command())
+
+        return cmd
+
+
+class Questa(Simulator):
+    def get_include_commands(self, includes):
+        include_cmd = []
+        for dir in includes:
+            include_cmd.append("+incdir+" + as_tcl_value(dir))
+
+        return include_cmd
+
+    def get_define_commands(self, defines):
+        defines_cmd = []
+        for define in defines:
+            defines_cmd.append("+define+" + as_tcl_value(define))
+
+        return defines_cmd
+
+    def get_parameter_commands(self, parameters):
+        parameters_cmd = []
+        for name, value in parameters.items():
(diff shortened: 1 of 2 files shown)


cocotb/run.py
```diff
@@ -0,0 +1,382 @@
...
+            results_xml_file = fo.name
+            fo.close()
+            self.env["COCOTB_RESULTS_FILE"] = results_xml_file
+        else:
+            results_xml_file = os.getenv("COCOTB_RESULTS_FILE")
+
+        cmds = self.build_command()
+        self.execute(cmds)
+
+        sys.tracebacklimit = 0  # remove not needed traceback from assert
```
eric-wieser, line 172: Better to use the pytest `traceback_hide` thing here, which I forget the spelling of

imphil:
Bevor we spend effort on the implementation details of this PR, let's first have a discussion about the general approach. I'll try to look at the docs and follow up with questions.

themperek:
>Bevor we spend effort on the implementation details of this PR, let's first have a discussion about the general approach. I'll try to look at the docs and follow up with questions.

Yes. It just some possible initial implementation just to start the discussion.

eric-wieser:
Found something interesting in the history - we actually used to have this:

https://github.com/cocotb/cocotb/commit/55023b9644aaf0574579edc07e30a1166f166550#diff-d5065fb45574b798030c936d155e6152R42

ktbarrett:
Yes, I think that the way we run cocotb should change, and I think we should develop on master as an experimental feature. I also think if we are going to change implementations, we should do a complete re-design rather than just translate what we have into Python. 

It's also a good opportunity to decide whether including tool automation in cocotb is even a good idea. I understand we need something to run our own tests, but I'm not sure if we should be evangelizing what we come up with as "the solution". While *more* features is nice (I'd like grouping, parameterization, isolation, etc), I don't think I want it in this repo. In fact I think there could be *less* in this repo.

eric-wieser:
What's the relation between this PR and the cocotb-test repo? Is this just a total merge of the two repos? If not, which parts are ommitted?

themperek:
> What's the relation between this PR and the cocotb-test repo? Is this just a total merge of the two repos? If not, which parts are ommitted?

cocotb-test is doing 2 things (compiling on setup) this is now in cocotb (and much better ;-) and this PR run control.

I think this is a bit improved version. Other simulators can be basically just copy/adjust.

I also removed logging, for now, that is ugly and should be done right (already have locally forwarding with serializing/deserialization). Can add now but maybe better wait (one thing at a time).

If you have some ideas/suggestion I am welcome to try.

themperek:
@ktbarrett 

>It's also a good opportunity to decide whether including tool automation in cocotb is even a good idea. I understand we need something to run our own tests, but I'm not sure if we should be evangelizing what we come up with as "the solution". While more features is nice (I'd like grouping, parameterization, isolation, etc), I don't think I want it in this repo. In fact I think there could be less in this repo.

I agree. One way I can see is to provide a very simple python-based launcher that can be easily extended/reused supporting very basic functionality all the rest "`rouping, parameterization, isolation, etc..`" should come from outside. Other ideas welcome.

eric-wieser:
It might be worth comparing this to `distutils` (which contains c compiler invocation tools) and `numpy.distutils` (which contains Fortran compiler invocation tools). Although actually running the simulator doesn't really fit into that model, the building vhdl and verilog bits do.

themperek:
We also should probably support edalize/vunit etc. and not try to replace those.

ktbarrett:
So it seems like we do want to reinvent the wheel when it comes to tools. The company I work for has something like edalize at work, so if we can build cocotb around support for that, our software shouldn't have much trouble using it as well.

I guess the other question is regressions. Something like pytest-parallel/xdist could be used for test isolation and parallelization, but that seems like it would best fit a role *wrapping* around the tool abstraction framework. I'm not entirely sure how to support grouping tests into a single simulator run using pytest.

No matter what the implementation is, we could move test discovery/selection out of cocotb. cocotb only *needs* to support running tests given as a list of names. Something like: `TESTS="test_mod_A.test1;test_mod_B:test5;..." and whatever tool can generate that list.

themperek:
>Something like: TESTS="test_mod_A.test1;test_mod_B:test5;..." and whatever tool can generate that list.

If I understand you, this is kind of is maybe supported even now with `MODULE` and `TESTCASE`?  See the example in the description. You can decide if you want to run in parallel or sequentially.

ktbarrett:
No `MODULE` and `TESTCASE` doesn't support that. You can only call out one test in `TESTCASE` and it only looks in the first module in the `MODULE` list. It could be better.

eric-wieser:
I'd rather we nice towards a custom entry point rather than add more complex env var parsing.

ktbarrett:
@eric-wieser I would actually remove `MODULE` and put everything in a single variable that does module and qualified name; a lot like how we would specify a custom entry point, just multiple tests. How does a custom entry point help us here? Or are you suggesting cocotb would only ever run a single test as the "custom entry"?

themperek:
Digression: Fundamentally we have 2 items:
- compile the sources
- run the test

Sometimes we may have multiple compilations (example changing generics) for the same test but most of the time multiple tests for the same compilation. At the moment we can run multiple tests within one simulation or multiple simulations. I ask myself how to present this to the user in the nicest way?

ktbarrett:
@themperek Doesn't pytest have the concept of grouping tests in a class? Such a group could represent a single simulator run.

themperek:
> @themperek Doesn't pytest have the concept of grouping tests in a class? Such a group could represent a single simulator run.

Not even pytest. Standard python unittest (https://docs.python.org/3/library/unittest.html)

garmin-mjames:
If I was going to use `pytest`, I'd want the actual cocotb tests to be integrated with `pytest`, so that pass/fail would be recorded correctly. Ideally, the cocotb tests themselves could use `pytest` idioms, and the simulator details would be hidden in fixtures and threads and whatever else in the background.
That way there wouldn't be this 2 step process of `pytest` tests that exist only to call cocotb tests.

themperek:
@garmin-mjames So we have to define design/simulation somehow. 
At the moment I would rather think running tests with cocotb itself should be rather simple and framework agnostic. For a specific one can do an extension. But if someone would have a good proposal...

ksigurdsson:
> If I was going to use `pytest`, I'd want the actual cocotb tests to be integrated with `pytest`, so that pass/fail would be recorded correctly. Ideally, the cocotb tests themselves could use `pytest` idioms, and the simulator details would be hidden in fixtures and threads and whatever else in the background.
> That way there wouldn't be this 2 step process of `pytest` tests that exist only to call cocotb tests.

I'm in agreement with @garmin-mjames. To be able to write tests once using pytest and fixture them with either the simulator/cocotb or FPGA or ASIC hardware would be a game changer for us in terms of efficiency.

(As an aside, this is my first post but I've already used cocotb to verify multiple blocks that have gone into an ASIL-D automotive ASIC. Love it. Keep up the good work everyone!)

themperek:
@ktbarrett The question is if such a way like here (that works and you can try): https://github.com/cocotb/cocotb/blob/6df59b8c4c1b5210e3756ccb070c5e9e2eae4ef4/tests/pytest/test_cocotb.py is fine or there are better ideas?

themperek:
@ksigurdsson The question is if such a way like here (that works and you can try): https://github.com/cocotb/cocotb/blob/6df59b8c4c1b5210e3756ccb070c5e9e2eae4ef4/tests/pytest/test_cocotb.py is fine or there are better ideas?

themperek:
I could imagine also something like this (but this is potentially built on top) what is proposed in this PR:
```python3
from cocotb import TestCase  # derived from unittest.TestCase

class TestStringMethods(TestCase):
    
    @classmethod
    def setUpClass(cls):
        run(
            verilog_sources=["sample_module.sv"],
            toplevel="sample_module"
        )

    def test_one(self):
        pass

    def test_two(self):
        # ...
        self.assertEqual("two","two")

if __name__ == '__main__':
    unittest.main()
```

eric-wieser:
I think we're conflating discussions here - we have #494 for building pytest integration. I think we should keep discussion here to "what would a python replacement for the makefile look like".

themperek:
> should keep discussion here to "what would a python replacement for the makefile look like".

Yes. And cocotb should probably be agnostic if someone uses pytest or anything else.

ktbarrett:
I vote that we describe the interface for running cocotb (which really isn't that much), and suggest an existing framework like edalize or vunit for user's regressions. We should document frameworks that are 'officially' supported (by us or the framework). We can even continue to use the makefiles internally if we choose, we just need to stop preaching them as "the solution". In my opinion us providing scripts to compile and run simulators has always been out of scope.

What the makefiles do now:
1. compile sources
2. setup the simulators environment
3. invoke the simulator
4. check/gather results (part of the regression system)

My thoughts:
1. shouldn't be our problem, frameworks like vunit do it better, don't reinvent the wheel
2. almost nothing here is cocotb specific, this can be done in the framework, or we can provide a script for the few that we do *need*
3. easy, all cocotb usually needs is passing an option to load one of the cocotb interface libraries
4. we don't want to over specify the regression system, build support for this into the framework

The regression framework we use at work uses the makefiles only to build cocotb (we haven't yet moved to the package installation approach since it's a huge bother and we don't have time or money for that right now). I honestly think us adding anything else will make integration into other frameworks more difficult, they already know how to call the simulator, we just need to provide a few extra bits to get cocotb to work.

themperek:
So a more general question.

Do we want python base flow in cocotb or we will only provide makefiles?

Options for runtime:
1. makefiles only, anything else by extension
2. makefiles and python
3. python only (for 2.0)

gretzteam:
It is now easy for people to include cocotb in their custom flow. 
I would thus vote for having an all python, simple and lightweight way to run the cocotb tests.
Pytest or vunit integration should be easy. And being all python, it would opens the door for clean design of new creative way of doing this stuff without requiring workaround of integrating makefile into modern flow.

ktbarrett:
@themperek. I think 4 is ideal, it will serve as a set of examples for users, but then we have to decide whether we want to add one of those frameworks (and which one) as a testing dependency.

ktbarrett:
Is anyone familiar enough with the existing regression frameworks to comment on which would be 'best' for us?

cmarqu:
It looks to me like edalize is intended exactly for our use case and is slim; and there is the recent PR https://github.com/olofk/edalize/pull/138 where @imphil and @themperek have commented.

themperek:
I personly think verification is software and would rather like to use standard software tools (nose, pytest, ...) than dedicated hardware development. I would like to run 100s test in parallel to collect coverage. This is not a problem though and can be done with extension.

ktbarrett:
@themperek I think what we are looking for is 2 things: a way to abstract away the specifics of calling a simulator, and a regression system. We can leverage an outside tool like edalize to fill the first goal, and potentially use pytest, nose, or something else for the regression system. It's easier to play with alternatives if we adopt a layered approach.

gretzteam:
One of the main point of cocotb is to leverage Python as a productive language. Using Makefiles to 'drive' the entire thing is a liability and slows down progress. As it is, cocotb out of the box does 90% of what a user need for developing and testing RTL modules. This is a very important feature, regardless if it is slightly outside the core scope of cocotb. 

Pushing this problem onto another flow like edalize, that probably tries to do too much, will also be a liability and cause maintenance/ownership issues (this is an edalize problem! this is a cocotb problem!)... In fact, all the work going on in the Makefile to support multiple simulators and different settings is an important part of cocotb, again, even if not core.

In fact, the only advantage Makefile have is that they are already there, and they work. So instead of asking the philosophical question of 'what should cocotb really be, what is the most atomic way to design it', I would vote for a simple Makefile rewrite using Python code like proposed in this PR. The philosophical discussions can keep happening, but at least, meanwhile, everyone is working with a modern platform...and new interesting ideas will come out of this that nobody can predict...

ktbarrett:
@gretzteam 

> One of the main point of cocotb is to leverage Python as a productive language. Using Makefiles to 'drive' the entire thing is a liability and slows down progress. As it is, cocotb out of the box does 90% of what a user need for developing and testing RTL modules. This is a very important feature, regardless if it is slightly outside the core scope of cocotb.

This can be solved by metapackages.

> Pushing this problem onto another flow like edalize, that probably tries to do too much, will also be a liability and cause maintenance/ownership issues (this is an edalize problem! this is a cocotb problem!)...

That is a possibility... But not one I think it prohibitive. If it becomes quite serious, we can always fork the project. Is there any reason, if we were to be a super project that does everything, that something like edalize still wouldn't be our solution? We would just copy a project that fits the role into our own, no? And if that is the ideal solution, why not let it be a separate entity so we don't have to worry about it for now?

> In fact, the only advantage Makefile have is that they are already there, and they work.

True-ish. I'm not against leaving them there. I know people use them. But make isn't available out of the box on Windows. Makefiles are a pretty crummy automation tool (a large section of the company I work for uses make for this, it's a convoluted mess that isn't portable). And I think 'work' is subjective, we do have a lot of problems with them...

What we have might be fine for small projects, but it isn't viable for large, complex projects, like the ones I work on daily. Regardless of how it's implemented, it will be insufficient until its parity with an existing project automation tool. What we can afford in maintainer time is better spent on other things.

If there is some simulator abstraction we could make that would allow us to support both basic projects ***and*** edalize without any extra work, we should investigate that, but I don't see it...

ktbarrett:
If you are interested in keeping the makefiles, or rewriting them in Python, if we adopt an approach where we don't overspecify how we run cocotb, we can still support the makefiles as a supported framework.

gretzteam:
Maybe I wasn't clear: I **_really_** want to get rid of the Makefiles - but at the same would like cocotb to keep working out of the box. I feel like the PR as presented here is the right tradeoff.

ktbarrett:
@gretzteam A simple port to Python wouldn't really solve any of the problems with the makefiles besides not running on Windows without cygwin or msys2. If that's important, sure we can change them, but it seems like a waste of effort, we could just keep the makefiles for legacy. Or did I still miss the point?

gretzteam:
I think it would solve more than just running on windows. For example, there were at least a few people asking ''how can i set VERILOG_FILES using some file_list.f that I already have lying around?'' The solution is some ninja one-liner bash/shell command. I'm sure there are many more of those little annoyance that would just go away. 
It would open the door to expansion to pytest, custom scripts to run simulation in a loop by varying parameters etc...

themperek:
@imphil

m-kru:
`cocotb` is a bit like simulator, that needs another simulators to run. This is very peculiar. The main goal of the `edalize` is

> Edalize is a Python Library for interacting with EDA tools. It can create project files for supported tools and run them in batch or GUI mode (where supported).

I think we all agree that `cocotb` is an EDA tool. It is an EDA tool needing another EDA tool to do some useful job. The question arises, should `cocotb` support `edalize` or should `edalize` support cocotb, or maybe both? Lets analyze 3 scenarios.

1. `cocotb` supports `edalize`
```
cocotb --calls--> edalize --calls--> simulator needed by cocotb
```
**Pros:** `cocotb` utilizes the knowledge of `edalize` on how to run different EDA tools. Less code to maintain. **Cons:** This approach has one big drawback. I think most people using `edalize` also uses `FuseSoc`. If `edalize` does not support `cocotb`, but cocotb supports `edalize`, there is no way to define cocotb  `FuseSoc` target. I am day to day `FuseSoc` user and such scenario does not ease my life.

2. `edalize` supports `cocotb`
```
edalize --calls--> cocotb --calls--> simulator needed by cocotb
```
In this scenarios cocotb is responsible for preparing and calling simulator, just like it is right now. **Pros:** No need to change internals of `cocotb`, backward compatibility, `cocotb` just needs to expose some API to become callable from Python runtime. `FuseSoc` users can define cocotb target.
**Cons:** No reuse of `edalize` knowledge on how to run different EDA tools. `cocotb` handles EDA tools in its own way.

3. `cocotb` supports `edalize`, `edalize` supports `cocotb`. This one is the most flexible. You can have this:
```
cocotb --calls--> edalize --calls--> simulator needed by cocotb
```
But you can also have this:
```
edalize --calls--> cocotb --calls--> edalize --calls--> simulator needed by cocotb
```
**Pros:** Supports all uses cases, `cocotb` utilizes the knowledge of `edalize` on how to run different EDA tools. Less code to maintain. **Cons:** I am not sure if you would be able to maintain backward compatibility from user perspective. It probably requires more work than option 1 or 2.

ktbarrett:
@m-kru 

>cocotb is a bit like simulator, that needs another simulators to run. 
>I think we all agree that cocotb is an EDA tool.

I don't agree with this. cocotb is a library that embeds a Python interpreter into a simulator, provide types and routines for interacting with the simulator, and a scheduler implementation. If anything cocotb should be considered a scripting plugin for a simulator.

Considering that, the only flow that makes sense to me is:

1. regression manager identifies tests through discovery or user specification
2. regression manager groups test according to user preference into runs
3. regression manager calls the runner
4. the runner compiles the sources and calls the simulator (with the bits necessary to load cocotb)
5. cocotb tests run and save the result
6. regression manager collects all the partial results from each run and combines them

With VUnit, it acts as both the regression manager and the test runner. With edalize, it acts only as the runner because it supports building and running simulations, but we would still need a regression manager. @themperek suggests pytest, nose, or something from the open source Python ecosystem. That's sort of how cocotb-test works, from what I recall.

So cocotb doesn't "call" anything, it is simply a plugin enabled during the call to the simulator which implies that support would need to be built into the runner (edalize or vunit) to support loading that plugin when invoking the simulator.

In response to your suggestions:

1. Is more or less what I'm suggesting, but cocotb would not sit outside of edalize, unless it also provides regression manager code.
2. edalize wrapping the makefiles? That kinda defeats the purpose of using edalize, like you said.
3. This seems like case 1, but the second example is unnecessarily complex. What's the point?

As for your worry about the use of fusesoc, it seems that fusesoc uses edalize for it's backend. So adding support for cocotb into edalize should implicitly create support in fusesoc. It's probably a little more complicated than that, but hopefully not prohibitively so.

m-kru:
@ktbarrett 

> I don't agree with this. cocotb is a library that embeds a Python interpreter into a simulator, provides types and routines for interacting with the simulator, and a scheduler implementation. If anything cocotb should be considered a scripting plugin for a simulator.

But the truth is, that cocotb does more. It also prepares and runs simulator and tries to manage your  tests. To me it looks more like an EDA, not pure library/plugin.

I think in this discussion we have some misunderstanding. Could someone provide clear definitions for
1. Python-based runtime
2. Regression manager
3. Runner

ktbarrett:
@m-kru 

>But the truth is, that cocotb does more. It also prepares and runs simulator and tries to manage your tests.

That's the part we want to replace with an outside tool because the current solution isn't great. None of that is vital for cocotb's function.

As I have mentioned before, at my company we don't use any of makefiles to run cocotb tests. We call the simulator with the appropriate arguments, set the appropriate environment variables, and viola. Our tools handle defining a project and dependencies, describing sources, describing a runtime environment, loading various EDA tools, building the project from sources, and running one of many supported testing frameworks, including cocotb, or any other supported flow, like synthesis. Anything that cocotb could provide will simply be in the way of that, and the same goes for all other frameworks. Making your framework depend on the makefiles to compile your sources and set up your environment steps on the features those more complete frameworks have to offer. For example, #437 concluded that wrapping the makefiles was also a pain. I am not interested in *wrapping* the makefiles, but creating a well-defined interface so alternatives can *supplant* it.

1. Python-based runtime
A tool or set of tools written in Python that would allow you to run cocotb tests. I'm not going to specify all the requirements because it could be basic, like our makefiles, or more feature rich, like fusesoc. Python implementation is important so we can support Windows without mingw.

3. Runner
A tool that supports the necessary features to get to the point of running a cocotb test. At a minimum this requires compiling sources and calling a simulator. Ideally, this will provide a consistent interface to the caller so they don't have to worry about differences between simulators. But it could be a lot more than that, for example edalize also support synthesis and implementation, but we don't really care about that for cocotb.

2. Regression manager
It must determine what tests to run and then call the runner to actually run them. It can call multiple runners, perhaps even in parallel, to implement test grouping, isolation, parallelization, etc. After all the runners finish, it should aggregate results for each run.

ktbarrett:
It took almost nothing to run cocotb from VUnit. https://github.com/ktbarrett/vunit-cocotb

We could embed knowledge into `cocotb-config` on what the simulator arguments are for each simulator/toplevel lang and that would be sufficient. The rest of the support would have to be done by the regression provider.

gretzteam:
In this context I guess Vunit is similar to what pytest would offer? 
I have no experience with either, could someone lay out the pros /cons? 
Regardless I'm sure it makes sense for cocotb to easily be used with both...

ktbarrett:
VUnit does it's testing in HDL and provides VHDL and SystemVerilog libraries for testbenching. We wouldn't use those libraries, but run cocotb in place of it. We would leverage it's compilation system and perhaps it's regression system.

Their compilation system:
* supports multiple simulators
* handles compilation order automatically
* allows for SV or VHDL at the top level (our makefiles don't)
* the projects are described as a Python script
* it's implemented totally in Python and supports Windows

Their regression is custom-built to work with their testbenching libraries, so additional work remains there. But it:
* supports multiple testbench tops
* supports multiple simulator runs per invocation
* joins the results of each simulator run into a summary
* can output summed results in XUnit format like cocotb does

So it fills all the major requirements, but there are some issues with it:
* It doesn't support Icarus (Icarus doesn't have enough SV support for it to use their testbenching libraries, so support was never added)
* Work needs to be done to tie into their test reporting format

Once we add a custom Python entry point, I think we could build a compat layer for VUnit support. That should probably be an extension.

The proof-of-concept is there, but there is a lot of work to get to the point for proper support.

gretzteam:
Thanks for the info. Vunit looks well made. Anybody has experience with Pytest and could compare what is proposed in the current PR + Pytest with a Vunit based solution?

gchqdev1729:
I had another go at cocotb support in Edalize. It's just a draft right now because there are some features missing and it's a bit different to other Edalize backends so will need some discussion. 

Nevertheless, it works and I think it fits with the direction of travel in this thread.

I based some of it off @ktbarrett 's sample VUnit code above.

https://github.com/olofk/edalize/pull/162

nredd:
@themperek @ktbarrett @eric-wieser @garmin-mjames I sort of skimmed the above conversation, I think this might be interesting to y'all:

I've written a pytest plugin for cocotb which my team has been successfully using to 'combine' cocotb test modules with their corresponding Makefile. My main goal was to create a low-overhead path to properly integrate our existing cocotb testbenches into our pytest setup without sacrificing pytest's nifty modularity/reporting/filtering logic, etc. I also know how frustrating it can be to rely on/maintain Makefiles :slightly_smiling_face: 

As is, users set their cocotb build flags as module-level attributes (i.e. you define `TOPLEVEL` and `VERILOG_SOURCES` constants in your cocotb modules), and the pytest plugin takes care of the rest:
   - generates a unique Makefile for each **unique** cocotb test/testbench item detected in a cocotb test module; this currently assumes anything marked with `@cocotb.test` should get a corresponding pytest item, and I should mention that we use this along with `cocotb.regression.TestFactory` instances without issue
   - calls each test item's Makefile using `subprocess.check_output` to invoke builds; all generated files are structured in such a way that parallel building is (more-or-less) seamless, and I can confirm that `pytest-xdist` works beautifully with the plugin. My main hesitation with parallel builds at the moment is that we're getting random hangs when running certain test items with Verilator (my gut feeling is that it is related to  [this issue](https://github.com/cocotb/cocotb/issues/2008)), which highlights some improvements to make in the process management logic when running tests. 
   - Parses each items produced `results.xml` to check for simulator failures, run time, and total sim time
   - Marks each cocotb test item with `TB` pytest marker which pytest can then use to properly filter out the legitimate cocotb test items. There is quite a bit going on with static-binding of test instances when e.g. using `TestFactory` instances to generate regression testbenches and, by default, pytest collects all object instances which it deems to be fair game--so, currently, the user needs to manually filter out collected items which aren't marked with `TB`. This is another area to iterate on.

There is obviously a lot going on under the hood here so if people are interested, I can plan on open-sourcing this code--probably as an extension, unless people feel otherwise. Personally, I don't have the bandwidth to guarantee (what I consider to be) adequate future support/maintenance for this code, so I would very much like to open it up to community if the interest is there :slightly_smiling_face: If so, feel free to suggest features that I can look into in the meantime. We only run Verilator on my team, so there is sure to be questions about generalization/support.

themperek:
@nredd Have you seen https://github.com/themperek/cocotb-test/ ?

themperek:
Can close and open new PR if it is better.

ktbarrett:
@themperek Could you create a new PR? I think this PR became more of a discussion and less of a review and we can leave it as it a discussion.

To summarize the discussion. We are tired of waiting for the mythical "integrated" solution and want to bring in some kind of Python-based runner now. More featureful, integrated solutions can come later. cocotb-test will be the basis of the new Python runner.

themperek:
Moved to https://github.com/cocotb/cocotb/pull/2634

PR #1564, opened 2020-04-04
Add documentation note for missing cocotb-config

### LINKED ISSUE #1561: Add a note what to do in case missing cocotb-config after installation
In case `cocotb-config : command not found...` suggest adding `PATH`.

With pip 20.0.2 there is a big warning when using `pip install --user`:
```
Installing collected packages: cocotb
  WARNING: The script cocotb-config is installed in '/cocotb/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
Successfully installed cocotb-1.4.0.dev0
``` 

From https://github.com/cocotb/cocotb/pull/1445#issuecomment-609051032

themperek:
Resolves #1561

### DIFF
--- a/documentation/source/quickstart.rst
+++ b/documentation/source/quickstart.rst
@@ -41,6 +41,11 @@ To install the development version of cocotb:
     git clone https://github.com/cocotb/cocotb
     pip3 install -e ./cocotb
 
+.. note::
+
+    After installation, you should be able to execute ``cocotb-config``.
+    If it is not found, you need to append its location to the ``PATH`` environment variable.
+    This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented :ref:`here <python:inst-alt-install-user>`.
 
 Native Linux Installation
 -------------------------


documentation/source/quickstart.rst
```diff
@@ -41,6 +41,11 @@ To install the development version of cocotb:
     git clone https://github.com/cocotb/cocotb
     pip3 install -e ./cocotb
 
+.. note::
+
+    After installation, you should be able to execute ``cocotb-config``.
+    If it is not found, you need to append to the ``PATH`` environment variable
+    so that it contains an entry for your Python installation.
```
cmarqu, line 48:
```suggestion
    If it is not found, you need to append its location to the ``PATH`` environment variable.
    This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented `here <https://docs.python.org/3/library/site.html#site.USER_BASE>`_.
```

documentation/source/quickstart.rst
```diff
@@ -41,6 +41,11 @@ To install the development version of cocotb:
     git clone https://github.com/cocotb/cocotb
     pip3 install -e ./cocotb
 
+.. note::
+
+    After installation, you should be able to execute ``cocotb-config``.
+    If it is not found, you need to append its location to the ``PATH`` environment variable.
+    This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented `here <https://docs.python.org/3/library/site.html#site.USER_BASE>`_.
```
eric-wieser, line 48:
```suggestion
    This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented `here <python:site.USER_BASE>`_.
```
Or something, will generate an intersphinx link.

eric-wieser:
Ci failure was caused by me tweaking settings, and gh-1569

documentation/source/quickstart.rst
```diff
@@ -41,6 +41,11 @@ To install the development version of cocotb:
     git clone https://github.com/cocotb/cocotb
     pip3 install -e ./cocotb
 
+.. note::
+
+    After installation, you should be able to execute ``cocotb-config``.
+    If it is not found, you need to append its location to the ``PATH`` environment variable.
+    This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented `here <python:site.USER_BASE>`_.
```
eric-wieser, line 48: Perhaps...
```suggestion
    This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented :data:`here <python:site.USER_BASE>`_.
```

documentation/source/quickstart.rst
```diff
@@ -41,6 +41,11 @@ To install the development version of cocotb:
     git clone https://github.com/cocotb/cocotb
     pip3 install -e ./cocotb
 
+.. note::
+
+    After installation, you should be able to execute ``cocotb-config``.
+    If it is not found, you need to append its location to the ``PATH`` environment variable.
+    This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented :ref:`here <python:inst-alt-install-user>`_.
```
eric-wieser, line 48: Hmm, still not right. Either

```suggestion
    This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented :ref:`here <python:inst-alt-install-user>`.
```
or perhaps
```suggestion
    This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented `here <python:inst-alt-install-user>`_.
```
Maybe we should stop guessing and read the intersphinx docs.

themperek: Should work now (works locally for me)

cmarqu: Works: https://cocotb--1564.org.readthedocs.build/en/1564/quickstart.html#installation-via-pip

PR #1462, opened 2020-02-27
Better Windows Support

themperek:
- Set `PYTHONHOME` in order to properly populate `sys.path` in embedded python interpreter in `Makefile.pylib.Msys`. (No need to set all pythonpath in the command line , just additional).
- Set `PATH` to include a path to cocotb compiled libraries in `Makefile.pylib.Msys`.
- Set appropriate path format for sources and paths.
- Remove unnecessary logic from example Makefiles.
- Reorder necessary variables and cleanup.

The important changes are in `cocotb/share/makefiles/Makefile.inc` and 
 `cocotb/share/makefiles/Makefile.pylib.Msys` the rest is just a cleanup (mostly remove a lot of obsolete code).

Resolves: #1064 #428 #498 #436 #354 

## Usage:
- (For Aldec) You may have to **not install** C++ compiler together with Riviera or remove/rename `mingw` directory in the install directory.
- Install [Miniconda](https://conda.io/en/latest/miniconda.html)  (according to your simulation version 32/64bit)
- Start `Anaconda Prompt` 
-  Install msys `conda install -c msys2 m2-base m2-make libpython m2w64-toolchain` 
- Set proper `%path%` to the binary of the simulator

For msys2 install [msys2](https://www.msys2.org). Install `pacman -S --needed base-devel mingw-w64-i686-python3-pip` or `pacman -S --needed mingw-w64-x86_64-toolchain mingw-w64-x86_64-python3-pip`. Use mingw32/64.exe.

## Test matrix on Windows 10 (64):

|            | icarus 10.1,64 | questa,64 | questa,32 | Rivera 2019.04,64 | Rivera 2019.04,32 | Modelsim 10.1b,32 |
|------------|----------------|----------------|----------------|-------------------|-------------------|-------------------|
| conda(3.7) | ok             | ok             | ok             | ok*              | ok*               | ok                |
| msys(3.7)   | ok             | ok             | ok             | ok               | ok                | ok                |

If this is acceptable will update documentation.
[*] Simulation is running ok but on exit `segmentation fault`.

PS. Thanks to [Aldec](https://www.aldec.com/) for providing the license.

PR #1565, opened 2020-04-04
Cleanup and fix tcl sources paths on Windows

themperek:
This PR introduced sources paths formatting on Windows using `cygpath` in `Makefile.pylib.Msys`

- It fixes sources paths format in tcl scripts (questa and aldec) 
- Simplifies the multi-OS support in Makefiles (not required special Windows path construction)
- Cleanup test and examples from unused code

It is a subset of #1462. Hope this makes it easier to review.

### DIFF
--- a/cocotb/share/makefiles/Makefile.pylib.Msys
+++ b/cocotb/share/makefiles/Makefile.pylib.Msys
@@ -48,7 +48,7 @@ endif
 
 PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'import distutils.sysconfig; print( distutils.sysconfig.get_python_inc() )')
 
-# For tcl scripts purpose source files paths in form of C:/User/..
+# For Tcl script purposes, generate source file paths in the form of C:/User/...
 ifneq ($(VERILOG_SOURCES),)
     VERILOG_SOURCES := $(shell cygpath -m $(VERILOG_SOURCES))
 endif
--- a/cocotb/share/makefiles/Makefile.pylib.Msys
+++ b/cocotb/share/makefiles/Makefile.pylib.Msys
@@ -47,3 +47,11 @@ ifeq ($(PYTHON_LIBDIR),None)
 endif
 
 PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'import distutils.sysconfig; print( distutils.sysconfig.get_python_inc() )')
+
+# For tcl scripts purpose source files paths in form of C:/User/..
+ifneq ($(VERILOG_SOURCES),)
+    VERILOG_SOURCES := $(shell cygpath -m $(VERILOG_SOURCES))
+endif
+ifneq ($(VHDL_SOURCES),)
+    VHDL_SOURCES := $(shell cygpath -m $(VHDL_SOURCES))
+endif
--- a/examples/endian_swapper/tests/Makefile
+++ b/examples/endian_swapper/tests/Makefile
@@ -30,22 +30,20 @@
 # Default to verilog
 TOPLEVEL_LANG ?= verilog
 
-WPWD=$(shell sh -c 'pwd -W')
 PWD=$(shell pwd)
 
 ifeq ($(OS),Msys)
 WPWD=$(shell sh -c 'pwd -W')
-PYTHONPATH := $(PWD)/../cosim;$(PYTHONPATH)
+PYTHONPATH := $(WPWD)/../cosim;$(PYTHONPATH)
 else
-WPWD=$(shell pwd)
 PYTHONPATH := $(PWD)/../cosim:$(PYTHONPATH)
 endif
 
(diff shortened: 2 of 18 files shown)


cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -47,3 +47,11 @@ ifeq ($(PYTHON_LIBDIR),None)
 endif
 
 PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'import distutils.sysconfig; print( distutils.sysconfig.get_python_inc() )')
+
+# For tcl scripts purpose source files paths in form of C:/User/..
```
cmarqu, line 51:
```suggestion
# For Tcl script purposes, generate source file paths in the form of C:/User/...
```

examples/endian_swapper/tests/Makefile
```diff
@@ -30,22 +30,20 @@
 # Default to verilog
 TOPLEVEL_LANG ?= verilog
 
-WPWD=$(shell sh -c 'pwd -W')
 PWD=$(shell pwd)
 
 ifeq ($(OS),Msys)
 WPWD=$(shell sh -c 'pwd -W')
-PYTHONPATH := $(PWD)/../cosim;$(PYTHONPATH)
+PYTHONPATH := $(WPWD)/../cosim;$(PYTHONPATH)
```
eric-wieser, line 37: Are you sure this windows special case is actually necessary?

My understanding was that msys automatically translated colon separated paths.

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -47,3 +47,11 @@ ifeq ($(PYTHON_LIBDIR),None)
 endif
 
 PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'import distutils.sysconfig; print( distutils.sysconfig.get_python_inc() )')
+
+# For Tcl script purposes, generate source file paths in the form of C:/User/...
```
eric-wieser, line 51: If this is for tcl scripts, should we perform this transformation at the point we emit tcl code?

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -47,3 +47,11 @@ ifeq ($(PYTHON_LIBDIR),None)
...
+ifneq ($(VERILOG_SOURCES),)
+    VERILOG_SOURCES := $(shell cygpath -m $(VERILOG_SOURCES))
```
eric-wieser, line 53: I worry this will handle spaces poorly. Maybe that's already a lost cause.

examples/endian_swapper/tests/Makefile
themperek: Not sure so did not touch but can check.

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: It is 2 files (quest and aldec). Where it should go? It should help with any input path construction.

eric-wieser: Perhaps define a `to_tcl_path` function using `define` based on Windows vs posix, and call that function from both makefiles?

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: If there are spaces in the path it does not work in any case (with or without this change).

themperek: Possibly for different reasons.

examples/endian_swapper/tests/Makefile
themperek: I have tried to make this work on Windows but this is not so easy (multiple quite involved issues). So whatever we do here does not make any difference.

themperek: I can try to come back to this and fix this example with pybind11 or setup tools at some point.

themperek:
The alternative is to have WPWD everywhere which is fine so I close this.

cocotb/share/makefiles/Makefile.pylib.Msys
eric-wieser: Does this idea work?

cocotb/share/makefiles/Makefile.pylib.Msys
eric-wieser: I think this is fine to ignore

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: It should work. Can try. Byt all this extra WPWD in user makefiles is also working if applied consistently.  Mether of taste.

themperek:
It can be reopened later. It is not critical for now.

Issue #1569, opened 2020-04-05
Installation does not work on readthedocs


eric-wieser:
If I check this checkbox:

![image](https://user-images.githubusercontent.com/425260/78471617-75fbf600-772a-11ea-9ce3-c0fb076e7b3a.png)

Then the build fails with:
```
building 'cocotb/libs/libgpilog' extension
creating build/temp.linux-x86_64-3.7/cocotb/share/lib/gpi_log
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -Icocotb/share/include -I/home/docs/.pyenv/versions/3.7.3/include/python3.7m -c cocotb/share/lib/gpi_log/gpi_logging.cpp -o build/temp.linux-x86_64-3.7/cocotb/share/lib/gpi_log/gpi_logging.o -std=c++11 -Wall -Wextra -Wcast-qual -Wwrite-strings -Wconversion -Wnon-virtual-dtor -Woverloaded-virtual -D__STDC_FORMAT_MACROS
g++ -pthread -shared -L/home/docs/.pyenv/versions/3.7.3/lib -L/home/docs/.pyenv/versions/3.7.3/lib build/temp.linux-x86_64-3.7/cocotb/share/lib/gpi_log/gpi_logging.o -Lbuild/lib.linux-x86_64-3.7/cocotb/libs -lpython3.7m -lcocotbutils -o build/lib.linux-x86_64-3.7/cocotb/libs/libgpilog.so
/usr/bin/ld: /home/docs/.pyenv/versions/3.7.3/lib/libpython3.7m.a(ceval.o): relocation R_X86_64_PC32 against symbol `_PyRuntime' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
error: command 'g++' failed with exit status 1
```

I'm a little worried this represents a regression WRT #1183. @tymonx, would you be able to test that master still works for your use-case?

---

I've turned this change back off for now, but it's required for #1568

themperek:
I think we are following #1183. Is this python compiled with "--enable-shared"?

eric-wieser:
I think you're right that it isn't compiled with `--enable-shared`. Wasn't the purpose of #1183 to make it work in that case though?

themperek:
I am not sure but as I understand it was to fix if you have both.

eric-wieser:
[...]

PR #1572, opened 2020-04-05
Add a readthedocs configuration file

eric-wieser:
As recommended by the readthedocs build logs.
This means if we rename these files in a PR, we don't have to globally change settings across all branches.
It also makes it easier to test solutions to gh-1569.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,13 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+version: 2
+
+python:
+  version: 3.7
+  install:
+    - requirements: documentation/requirements.txt
+
+sphinx:
+  configuration: documentation/source/conf.py


.readthedocs.yml
```diff
@@ -0,0 +1,13 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+version: 2
+
+python:
+  version: 3.7
```
eric-wieser, line 8: This is what readthedocs is currently using. We can upgrade to 3.8 at some point, but the goal here was just to record the existing settings in the repo

.readthedocs.yml
```diff
@@ -0,0 +1,13 @@
...
+  install:
+    - requirements: documentation/requirements.txt
```
eric-wieser, line 10: gh-1569 can be reproduced by adding the following here:
```yaml
     - method: pip
       path: .
```

PR #1573, opened 2020-04-05
Build docs against installed cocotb

eric-wieser:
This seems more common in packages I've looked at before.

CI almost certainly fails due to #1569

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,9 +5,11 @@
 version: 2
 
 python:
-  version: 3.7
+  version: 3.8
   install:
     - requirements: documentation/requirements.txt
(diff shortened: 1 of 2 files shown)


eric-wieser:
Restarting CI now that `PYTHON_CONFIGURE_OPTS="--enable-shared"` is in the (invisible) RTD config

themperek:
@eric-wieser I can set this up but need to switch to `conda` if you like.

eric-wieser:
Worth a try, go for it.

Kinda a shame we have to do that, but I don't think https://github.com/readthedocs/readthedocs-docker-images/pull/130 will get merged any time soon.

PR #1574, opened 2020-04-05
Set PYTHON_BIN using cocotb-config

themperek:
Mentioned here: https://github.com/cocotb/cocotb/pull/1462#discussion_r403546411

Use `cocotb-config` to resolve python executable that cocotb was installed with.
Remove `COCOTB_PYTHON_DEBUG` (undocumented) still `PYTHON_BIN` can be overwritten if needed.

It will help in many cases like #1539 and for multi-os support.

### DIFF
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -54,6 +54,9 @@ OS=Msys
 endif
 export OS
 
+# this ensures we use the same python as the one cocotb was installed into
+PYTHON_BIN ?= $(shell cocotb-config --python-bin)
+
 include $(COCOTB_SHARE_DIR)/makefiles/Makefile.deprecations
 include $(COCOTB_SHARE_DIR)/makefiles/Makefile.pylib
 
--- a/cocotb/share/makefiles/Makefile.pylib.Msys
+++ b/cocotb/share/makefiles/Makefile.pylib.Msys
@@ -29,21 +29,9 @@
 
 # All common python related rules
 
-# if not explicitly set, auto-detect python dir from system path
-ifeq ($(PYTHON_DIR),)
-    PYTHON_DIR ?= $(shell dirname $(shell :; command -v python))
-endif
-
-# make sure python dir was set properly
-ifeq ($(PYTHON_DIR),)
-    $(error "Path to Python directory must be included in system path or defined in PYTHON_DIR environment variable")
-endif
-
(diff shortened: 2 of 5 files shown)


eric-wieser:
Msys doesn't use python_bin?

themperek:
> Msys doesn't use python_bin?

Mistake. Fixed.

cocotb/share/makefiles/Makefile.inc
```diff
@@ -54,6 +54,8 @@ OS=Msys
 endif
 export OS
 
+PYTHON_BIN ?= $(shell cocotb-config --python-bin)
```
eric-wieser, line 57:
```suggestion
# this ensures we use the same python as the one cocotb was installed into
PYTHON_BIN ?= $(shell cocotb-config --python-bin)
```

eric-wieser:
Maybe worth a release note?

eric-wieser:
Something like _"there's no longer any need to set the `PYTHON_BIN` makefile variable, the python executable automatically matches the one cocotb was installed into"_.

eric-wieser:
You might need to rebase

cocotb/share/makefiles/Makefile.inc
eric-wieser: This got lost

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -29,17 +29,7 @@
...
-    PYTHON_DIR ?= $(shell dirname $(shell :; command -v python))
-endif
-
-# make sure python dir was set properly
-ifeq ($(PYTHON_DIR),)
-    $(error "Path to Python directory must be included in system path or defined in PYTHON_DIR environment variable")
-endif
-
-PYTHON_BIN?=$(PYTHON_DIR)/python.exe
+PYTHON_DIR = $(shell dirname $(PYTHON_BIN))
```
eric-wieser, line 32: Can you inline this into line 36, as
```make
ifeq ($(PYTHON_LIBDIR),None)
    PYTHON_LIBDIR=$(shell dirname $(PYTHON_BIN))/libs
endif
```

Nothing else uses this variable now

documentation/source/newsfragments/1574.change.rst
```diff
@@ -0,0 +1 @@
+There's no longer any need to set the ``PYTHON_BIN`` makefile variable, the python executable automatically matches the one cocotb was installed into.
```
imphil, line 1: python -> Python

PR #1576, opened 2020-04-05
Update docs for async/await syntax

ktbarrett:
This updates the documentation on coroutines and triggers in the documentation to use `async`/`await`. What was previously the `Async Syntax` section now describes the old generator-based approach and mentions it's compatibility with new-style coroutines.

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -724,7 +724,7 @@ class First(_AggregateWaitable):
             t_ret = await First(t1, t2)
 
     .. note::
-        When using the old-style :keyword:`yield`-based coroutines, ``t = yield [a, b]`` was another spelling of
+        In the old-style :ref:`generator-based coroutines <yield-syntax>`, ``t = yield [a, b]`` was another spelling of
         ``t = yield First(a, b)``. This spelling is no longer available when using :keyword:`await`-based
         coroutines.
     """
--- a/documentation/source/coroutines.rst
+++ b/documentation/source/coroutines.rst
@@ -5,80 +5,50 @@ Coroutines
 **********
 
 Testbenches built using cocotb use coroutines. While the coroutine is executing
-the simulation is paused. The coroutine uses the :keyword:`yield` keyword to
-pass control of execution back to the simulator and simulation time can advance
-again.
+the simulation is paused. The coroutine uses the :keyword:`await` keyword to
+block on another coroutine's execution or pass control of execution back to the
+simulator, allowing simulation time to advance.
 
-Typically coroutines :keyword:`yield` a :any:`Trigger` object which
+Typically coroutines :keyword:`await` a :class:`~cocotb.triggers.Trigger` object which
 indicates to the simulator some event which will cause the coroutine to be woken
 when it occurs.  For example:
 
 .. code-block:: python3
 
-    @cocotb.coroutine
-    def wait_10ns():
+    async def wait_10ns():
         cocotb.log.info("About to wait for 10 ns")
-        yield Timer(10, units='ns')
+        await Timer(10, units='ns')
         cocotb.log.info("Simulation time has advanced by 10 ns")
 
-Coroutines may also yield other coroutines:
+Coroutines may also :keyword:`await` on other coroutines:
 
 .. code-block:: python3
 
-    @cocotb.coroutine
-    def wait_100ns():
+    async def wait_100ns():
         for i in range(10):
-            yield wait_10ns()
+            await wait_10ns()
 
-Coroutines can return a value, so that they can be used by other coroutines.
+Coroutines can :keyword:`return` a value, so that they can be used by other coroutines.
 
 .. code-block:: python3
 
-    @cocotb.coroutine
-    def get_signal(clk, signal):
-        yield RisingEdge(clk)
+    async def get_signal(clk, signal):
+        await RisingEdge(clk)
         return signal.value
 
-    @cocotb.coroutine
-    def check_signal_changes(dut):
-        first = yield get_signal(dut.clk, dut.signal)
-        second = yield get_signal(dut.clk, dut.signal)
+    async def check_signal_changes(dut):
+        first = await get_signal(dut.clk, dut.signal)
+        second = await get_signal(dut.clk, dut.signal)
         if first == second:
             raise TestFailure("Signal did not change")
 
 
-Coroutines may also yield a list of triggers and coroutines to indicate that
-execution should resume if *any* of them fires:
-
-.. code-block:: python3
-
-    @cocotb.coroutine
-    def packet_with_timeout(monitor, timeout):
-        """Wait for a packet but time out if nothing arrives"""
-        yield [Timer(timeout, units='ns'), RisingEdge(dut.ready)]
-
-
-The trigger that caused execution to resume is passed back to the coroutine,
-allowing them to distinguish which trigger fired:
-
-.. code-block:: python3
-
-    @cocotb.coroutine
-    def packet_with_timeout(monitor, timeout):
-        """Wait for a packet but time out if nothing arrives"""
-        tout_trigger = Timer(timeout, units='ns')
-        result = yield [tout_trigger, RisingEdge(dut.ready)]
-        if result is tout_trigger:
-            raise TestFailure("Timed out waiting for packet")
-
-
-Coroutines can be forked for parallel operation within a function of that code and
-the forked code.
+Coroutines can be used with :func:`~cocotb.fork` for concurrent operation.
 
 .. code-block:: python3
 
     @cocotb.test()
-    def test_act_during_reset(dut):
+    async def test_act_during_reset(dut):
         """While reset is active, toggle signals"""
         tb = uart_tb(dut)
         # "Clock" is a built in class for toggling a clock signal
@@ -87,28 +57,28 @@ the forked code.
         # part of the user-generated "uart_tb" class
         cocotb.fork(tb.reset_dut(dut.rstn, 20))
 
-        yield Timer(10, units='ns')
+        await Timer(10, units='ns')
         print("Reset is still active: %d" % dut.rstn)
-        yield Timer(15, units='ns')
-        print("Reset has gone inactive: %d" % dut.rstn)
+        await Timer(15, units='ns')
+        await("Reset has gone inactive: %d" % dut.rstn)
 
 
-Coroutines can be joined to end parallel operation within a function.
+Forked coroutines can be used in an :keyword:`await` statement to block until the forked coroutine finishes.
 
 .. code-block:: python3
 
     @cocotb.test()
-    def test_count_edge_cycles(dut, period=1, clocks=6):
-        cocotb.fork(Clock(dut.clk, period, units='ns').start())
-        yield RisingEdge(dut.clk)
+    async def test_count_edge_cycles(dut, period_ns=1, clocks=6):
+        cocotb.fork(Clock(dut.clk, period_ns, units='ns').start())
+        await RisingEdge(dut.clk)
 
-        timer = Timer(period + 10)
+        timer = Timer(period_ns + 10, 'ns')
         task = cocotb.fork(count_edges_cycles(dut.clk, clocks))
         count = 0
         expect = clocks - 1
 
         while True:
-            result = yield [timer, task.join()]
+            result = await First(timer, task)
             if count > expect:
                 raise TestFailure("Task didn't complete in expected time")
             if result is timer:
@@ -117,71 +87,42 @@ Coroutines can be joined to end parallel operation within a function.
             else:
                 break
 
-Coroutines can be killed before they complete, forcing their completion before
+Forked coroutines can be killed before they complete, forcing their completion before
 they'd naturally end.
 
 .. code-block:: python3
 
     @cocotb.test()
-    def test_different_clocks(dut):
+    async def test_different_clocks(dut):
         clk_1mhz   = Clock(dut.clk, 1.0, units='us')
         clk_250mhz = Clock(dut.clk, 4.0, units='ns')
 
         clk_gen = cocotb.fork(clk_1mhz.start())
         start_time_ns = get_sim_time(units='ns')
-        yield Timer(1, units='ns')
-        yield RisingEdge(dut.clk)
+        await Timer(1, units='ns')
+        await RisingEdge(dut.clk)
         edge_time_ns = get_sim_time(units='ns')
-        # NOTE: isclose is a Python 3.5+ feature
         if not isclose(edge_time_ns, start_time_ns + 1000.0):
             raise TestFailure("Expected a period of 1 us")
 
-        clk_gen.kill()
+        clk_gen.kill()  # kill clock coroutine here
 
         clk_gen = cocotb.fork(clk_250mhz.start())
         start_time_ns = get_sim_time(units='ns')
-        yield Timer(1, units='ns')
-        yield RisingEdge(dut.clk)
+        await Timer(1, units='ns')
+        await RisingEdge(dut.clk)
         edge_time_ns = get_sim_time(units='ns')
-        # NOTE: isclose is a Python 3.5+ feature
         if not isclose(edge_time_ns, start_time_ns + 4.0):
             raise TestFailure("Expected a period of 4 ns")
 
-.. _async_functions:
 
-Async functions
-===============
-
-Python 3.5 introduces :keyword:`async` functions, which provide an alternative
-syntax. For example:
-
-.. code-block:: python3
-
-    @cocotb.coroutine
-    async def wait_10ns():
-        cocotb.log.info("About to wait for 10 ns")
-        await Timer(10, units='ns')
-        cocotb.log.info("Simulation time has advanced by 10 ns")
-
-To wait on a trigger or a nested coroutine, these use :keyword:`await` instead
-of :keyword:`yield`. Provided they are decorated with ``@cocotb.coroutine``,
-``async def`` functions using :keyword:`await` and regular functions using
-:keyword:`yield` can be used interchangeable - the appropriate keyword to use
-is determined by which type of function it appears in, not by the
-sub-coroutine being called.
-
-.. versionadded:: 1.4
+.. versionchanged:: 1.4
     The :any:`cocotb.coroutine` decorator is no longer necessary for ``async def`` coroutines.
     ``async def`` coroutines can be used, without the ``@cocotb.coroutine`` decorator, wherever decorated coroutines are accepted,
     including :keyword:`yield` statements and :any:`cocotb.fork`.
 
-.. note::
-    It is not legal to ``await`` a list of triggers as can be done in
-    ``yield``-based coroutine with ``yield [trig1, trig2]``. Use
-    ``await First(trig1, trig2)`` instead.
-
 Async generators
-----------------
+================
 
 In Python 3.6, a ``yield`` statement within an ``async`` function has a new
 meaning (rather than being a ``SyntaxError``) which matches the typical meaning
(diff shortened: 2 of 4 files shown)


ktbarrett:
Addresses some points in #1289 
* Remove 2x "# NOTE: isclose is a Python 3.5+ feature" from documentation/source/coroutines.rst
* Change coroutines to use async def, to teach good practice in docs

documentation/source/coroutines.rst
```diff
@@ -202,3 +137,50 @@ type of generator function that can be iterated with ``async for``:
 
 More details on this type of generator can be found in :pep:`525`.
 
+
+Generator-based coroutines
+==========================
+
+Previous to Python 3.5 and the introduction of :keyword:`async` and :keyword:`await`, coroutines were implemented as wrappers around generators.
```
eric-wieser, line 144:
```suggestion
Prior to Python 3.5, and the introduction of :keyword:`async` and :keyword:`await`, coroutines were implemented as wrappers around generators.
```

documentation/source/coroutines.rst
```diff
@@ -202,3 +137,50 @@ type of generator function that can be iterated with ``async for``:
...
+        while True:
+            yield Timer(half_period, half_period_units)
+            signal <= ~signal
+
+    @cocotb.test()
+    async def test(dut):
+        cocotb.fork(clock(dut.clk, 5, 'ns'))
+        await RisingEdge(dut.clk)
+
+    .. note:: This style is no longer recommended and support may someday be removed.
```
eric-wieser, line 186:
```suggestion
.. note:: This style is no longer recommended and support may someday be removed.
```

documentation/source/coroutines.rst
```diff
@@ -117,71 +87,36 @@ Coroutines can be joined to end parallel operation within a function.
...
-of :keyword:`yield`. Provided they are decorated with ``@cocotb.coroutine``,
-``async def`` functions using :keyword:`await` and regular functions using
-:keyword:`yield` can be used interchangeable - the appropriate keyword to use
-is determined by which type of function it appears in, not by the
-sub-coroutine being called.
-
-.. versionadded:: 1.4
-    The :any:`cocotb.coroutine` decorator is no longer necessary for ``async def`` coroutines.
-    ``async def`` coroutines can be used, without the ``@cocotb.coroutine`` decorator, wherever decorated coroutines are accepted,
-    including :keyword:`yield` statements and :any:`cocotb.fork`.
```
eric-wieser, line 176: This section should be kept, but probably moved down to the section below

documentation/source/coroutines.rst
```diff
@@ -117,71 +87,36 @@ Coroutines can be joined to end parallel operation within a function.
...
-
-.. note::
-    It is not legal to ``await`` a list of triggers as can be done in
-    ``yield``-based coroutine with ``yield [trig1, trig2]``. Use
-    ``await First(trig1, trig2)`` instead.
```
eric-wieser, line 181: I think we should keep this note around somewhere too.

documentation/source/coroutines.rst
```diff
@@ -202,3 +137,50 @@ type of generator function that can be iterated with ``async for``:
 
 More details on this type of generator can be found in :pep:`525`.
 
+
+Generator-based coroutines
```
eric-wieser, line 141: Likely will want a label on this, maybe

```suggestion
.. yield-syntax:

Generator-based coroutines
```

Or whatever the syntax for label markers is.

ktbarrett: Because it took me a while to figure it out, I'll mention it's:

`.. _ref_name:`

`:ref:'ref_name'`

Note the required underscore and lack of it when using it out.

eric-wieser: Pretty sure the ref should be with backticks, not quotes.

ktbarrett: yeah, I just haven't figured out how to next backticks.

documentation/source/coroutines.rst
ktbarrett: It's in the documentation for `First`.

documentation/source/coroutines.rst
ktbarrett: I moved it to the bottom of the coroutine section since it applies to the `async` coroutines used in the examples.

documentation/source/coroutines.rst
```diff
@@ -5,80 +5,50 @@ Coroutines
...
-    @cocotb.coroutine
-    def wait_10ns():
+    async def wait_10ns():
         cocotb.log.info("About to wait for 10 ns")
-        yield Timer(10, units='ns')
+        await Timer(10, units='ns')
         cocotb.log.info("Simulation time has advanced by 10 ns")
 
-Coroutines may also yield other coroutines:
+Coroutines may also wait other coroutines:
```
eric-wieser, line 23: Either
```suggestion
Coroutines may also wait for other coroutines:
```
```suggestion
Coroutines may also wait upon other coroutines:
```
```suggestion
Coroutines may also wait on other coroutines:
```
```suggestion
Coroutines may also await other coroutines:
```

documentation/source/coroutines.rst
```diff
@@ -117,71 +87,42 @@ Coroutines can be joined to end parallel operation within a function.
...
-        cocotb.log.info("Simulation time has advanced by 10 ns")
-
-To wait on a trigger or a nested coroutine, these use :keyword:`await` instead
-of :keyword:`yield`. Provided they are decorated with ``@cocotb.coroutine``,
-``async def`` functions using :keyword:`await` and regular functions using
-:keyword:`yield` can be used interchangeable - the appropriate keyword to use
-is determined by which type of function it appears in, not by the
-sub-coroutine being called.
 
 .. versionadded:: 1.4
```
eric-wieser, line 119: Maybe out of scope, but thinking this should be
```suggestion
.. versionchanged:: 1.4
```

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,52 @@ type of generator function that can be iterated with ``async for``:
 
 More details on this type of generator can be found in :pep:`525`.
 
+
+.. _yield_syntax:
```
eric-wieser, line 147:
```suggestion
.. _yield-syntax:
```
rst seems to prefer hyphenated targets

eric-wieser: Can you put a link to here in the `First` docs?

ktbarrett:
whoops...

documentation/source/coroutines.rst
ktbarrett: Is a note at the bottom sufficient? I'm not sure where else would be a natural place to add that.

eric-wieser: Just add a link by the current "When using the old-style yield-based coroutines" note

documentation/source/coroutines.rst
cmarqu: I like the version using await, otherwise someone could find it ambiguous why we don't use the keyword when it fits perfectly.

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,53 @@ type of generator function that can be iterated with ``async for``:
...
+    def wait_100ns():
+        for _ in range(10):
+            yield wait_10ns()
+
+You may see existing code that uses this syntax for coroutines, but do not worry, it is compatible with async coroutines.
+Any place that will accept async coroutines, will also accept generator-based coroutines;
+including :func:`~cocotb.fork`.
+Likewise, any object that can be used in an :keyword:`await` statement can also be used in a :keyword:`yield` statement while in a generator-based coroutine;
+including triggers like :class:`~cocotb.triggers.Timer`.
+You may also see syntax like ``yield [triggerA, tiggerB, ...]``, this is syntax sugar for :class:`~cocotb.triggers.First`.
```
eric-wieser, line 173:
```suggestion
You may also see syntax like ``yield [trigger_a, trigger_b, ...]``, which is syntactic sugar for :class:`~cocotb.triggers.First`.
```

documentation/source/quickstart.rst
```diff
@@ -302,41 +302,41 @@ We can also cast the signal handle directly to an integer:
...
+        # This timer will complete before the timer in the concurrently executing "reset_thread"
+        await Timer(250, units='ns')
         dut._log.debug("During reset (reset_n = %s)" % reset_n.value)
 
         # Wait for the other thread to complete
-        yield reset_thread.join()
+        await reset_thread
         dut._log.debug("After reset")
+
+See :ref:`coroutines` for more examples of whay can be done with coroutines.
```
eric-wieser, line 342:
```suggestion
See :ref:`coroutines` for more examples of what can be done with coroutines.
```

documentation/source/coroutines.rst
eric-wieser: Would still like to see a backlink from the First docs to this section.

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,53 @@ type of generator function that can be iterated with ``async for``:
...
+        while True:
+            yield Timer(half_period, half_period_units)
+            signal <= ~signal
+
+    @cocotb.test()
+    async def test(dut):
+        cocotb.fork(clock(dut.clk, 5, 'ns'))
+        await RisingEdge(dut.clk)
+
+.. note:: This style is no longer recommended and support may someday be removed.
```
eric-wieser, line 195: I think this note would be best at the start of this section.

documentation/source/coroutines.rst
```diff
@@ -87,28 +57,28 @@ the forked code.
...
+        await Timer(10, units='ns')
         print("Reset is still active: %d" % dut.rstn)
-        yield Timer(15, units='ns')
-        print("Reset has gone inactive: %d" % dut.rstn)
+        await Timer(15, units='ns')
+        await("Reset has gone inactive: %d" % dut.rstn)
 
 
-Coroutines can be joined to end parallel operation within a function.
+Forked coroutines can used in an :keyword:`await` statement to block until the coroutine finishes.
```
cmarqu, line 66:
```suggestion
Forked coroutines can be used in an :keyword:`await` statement to block until the coroutine finishes.
```

documentation/source/quickstart.rst
```diff
@@ -302,41 +302,41 @@ We can also cast the signal handle directly to an integer:
 
 
 
-Parallel and sequential execution
+Blocking and concurrent execution
```
cmarqu, line 305: Users searching for "parallel" or "sequential" should still be able to find this section, can we offer these terms in parentheses maybe?

documentation/source/quickstart.rst
```diff
@@ -302,41 +302,41 @@ We can also cast the signal handle directly to an integer:
...
+Wrapping the call in :func:`~cocotb.fork` runs the coroutine concurrently, allowing the current coroutine to continue executing.
+At any time you can :keyword:`await` the result of the forked coroutine, which will block until the forked coroutine finishes.
 
 The following example shows these in action:
 
 .. code-block:: python3
 
-    @cocotb.coroutine
-    def reset_dut(reset_n, duration):
+    async def reset_dut(reset_n, duration):
```
cmarqu, line 317: Here is our chance to rename ``duration``to ``duration_ns``. :)

documentation/source/quickstart.rst
```diff
@@ -302,41 +302,41 @@ We can also cast the signal handle directly to an integer:
...
 
     @cocotb.test()
-    def parallel_example(dut):
+    async def parallel_example(dut):
         reset_n = dut.reset
 
-        # This will call reset_dut sequentially
         # Execution will block until reset_dut has completed
-        yield reset_dut(reset_n, 500)
+        await reset_dut(reset_n, 500)
```
cmarqu, line 328:
```suggestion
        await reset_dut(reset_n, duration=500)
```
(or ``duration_ns``).

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,53 @@ type of generator function that can be iterated with ``async for``:
...
+    @cocotb.coroutine
+    def clock(signal, half_period, half_period_units):
+        signal <= 0
+        while True:
+            yield Timer(half_period, half_period_units)
+            signal <= ~signal
+
+    @cocotb.test()
+    async def test(dut):
+        cocotb.fork(clock(dut.clk, 5, 'ns'))
```
cmarqu, line 192:
```suggestion
        cocotb.fork(clock(dut.clk, 5, units='ns'))
```
for consistency

documentation/source/quickstart.rst
ktbarrett: Is there someway to make a search for those terms drop them at this section without displaying the terms? "Parallel" is not correct.

cmarqu: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-only
seems to suggest that ``.. only:: never`` (assuming ``never`` being undefined and thus ``False``) could work.

ktbarrett: While that successfully generated nothing, it does not turn up in a search.

cmarqu: Ah, too bad.

While something could probably be done with https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_object_type, it's not worth the effort.

ktbarrett: I'll change the heading back for now.

ktbarrett:
@eric-wieser @cmarqu Thanks for the reviews!

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,53 @@ type of generator function that can be iterated with ``async for``:
...
+    def wait_100ns():
+        for _ in range(10):
+            yield wait_10ns()
+
+You may see existing code that uses this syntax for coroutines, but do not worry, it is compatible with async coroutines.
+Any place that will accept async coroutines, will also accept generator-based coroutines;
+including :func:`~cocotb.fork`.
+Likewise, any object that can be used in an :keyword:`await` statement can also be used in a :keyword:`yield` statement while in a generator-based coroutine;
+including triggers like :class:`~cocotb.triggers.Timer`.
+You may also see syntax like ``yield [trigger_a, tigger_b, ...]``, which is syntactic sugar for :class:`~cocotb.triggers.First`.
```
eric-wieser, line 175: I think this sentence should be below the example, since it seems to me the example is associated with the previous sentence.

cocotb/triggers.py
```diff
@@ -726,7 +726,7 @@ class First(_AggregateWaitable):
             t_ret = await First(t1, t2)
 
     .. note::
-        When using the old-style :keyword:`yield`-based coroutines, ``t = yield [a, b]`` was another spelling of
+        When using :ref:`generator-based coroutines <yield-syntax>`, ``t = yield [a, b]`` was another spelling of
```
eric-wieser, line 729:
```suggestion
        In the old-style :ref:`generator-based coroutines <yield-syntax>`, ``t = yield [a, b]`` was another spelling of
```
Part tense doesn't read to well in the second half without some time marker

documentation/source/coroutines.rst
```diff
@@ -87,28 +57,28 @@ the forked code.
...
+        await RisingEdge(dut.clk)
 
         timer = Timer(period + 10)
         task = cocotb.fork(count_edges_cycles(dut.clk, clocks))
         count = 0
         expect = clocks - 1
 
         while True:
-            result = yield [timer, task.join()]
+            result = await cocotb.triggers.First(timer, task)
```
eric-wieser, line 81: Perhaps we should import the short name like we do for the other triggers, as
```suggestion
            result = await First(timer, task)
```

documentation/source/coroutines.rst
ktbarrett: I also added a small example that is the example from `First`'s docstring, but using `yield []`.

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,60 @@ type of generator function that can be iterated with ``async for``:
...
+        while True:
+            yield Timer(half_period, half_period_units)
+            signal <= ~signal
+
+    @cocotb.test()
+    async def test(dut):
+        cocotb.fork(clock(dut.clk, 5, units='ns'))
+        await RisingEdge(dut.clk)
+
+You may also see syntax like ``yield [trigger_a, tigger_b, ...]``, which is syntactic sugar for :class:`~cocotb.triggers.First`.
```
eric-wieser, line 196:
```suggestion
You may also see syntax like ``yield [trigger_a, trigger_b, ...]``, which is syntactic sugar for :class:`~cocotb.triggers.First`.
```

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,60 @@ type of generator function that can be iterated with ``async for``:
...
+        yield Timer(10, units='ns')
+        cocotb.log.info("Simulation time has advanced by 10 ns")
+
+    @cocotb.coroutine
+    def wait_100ns():
+        for _ in range(10):
+            yield wait_10ns()
+
+You may see existing code that uses this syntax for coroutines, but do not worry, it is compatible with async coroutines.
+Any place that will accept async coroutines, will also accept generator-based coroutines;
```
cmarqu, line 171:
```suggestion
Any place that will accept async coroutines will also accept generator-based coroutines;
```
?

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,60 @@ type of generator function that can be iterated with ``async for``:
...
+        cocotb.fork(clock(dut.clk, 5, units='ns'))
+        await RisingEdge(dut.clk)
+
+You may also see syntax like ``yield [trigger_a, tigger_b, ...]``, which is syntactic sugar for :class:`~cocotb.triggers.First`.
+
+.. code-block:: python3
+
+    t1 = Timer(10, units='ps')
+    t2 = Timer(10, units='ps')
+    t_ret = yield[t1, t2]
```
eric-wieser, line 202:
```suggestion
    t_ret = yield [t1, t2]
```

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,60 @@ type of generator function that can be iterated with ``async for``:
...
+.. code-block:: python3
+
+    @cocotb.coroutine
+    def clock(signal, half_period, half_period_units):
+        signal <= 0
+        while True:
+            yield Timer(half_period, half_period_units)
+            signal <= ~signal
+
+    @cocotb.test()
```
eric-wieser, line 191: Perhaps worth mentioning that `test` objects can be generators too without the coroutine decorator.

documentation/source/coroutines.rst
ktbarrett: That would be correct. Imagine replacing the phrase "Any place that will accept async coroutines" with "They"; the comma would be out-of-place.

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,62 @@ type of generator function that can be iterated with ``async for``:
...
+Likewise, any object that can be used in an :keyword:`await` statement can also be used in a :keyword:`yield` statement while in a generator-based coroutine;
+including triggers like :class:`~cocotb.triggers.Timer`.
+Tests can also be generator-based coroutines.
+Tests are not required to be decorated with :class:`~cocotb.decorators.coroutine` as the :class:`~cocotb.decorators.test` decorator will handle this case automatically.
+
+.. code-block:: python3
+
+    async def wait_100ns():
+        for _ in range(10):
+            await wait_10ns()
```
eric-wieser, line 182:
```suggestion
            # generator-based coroutines can still be awaited
            await wait_10ns()
```
Also, I think this example belongs after line 272.

ktbarrett: I'm not sure what line you meant, but I think you mean I could break up the previous paragraph and move the examples after every statement?

eric-wieser: That's basically what I'm suggesting, yes - make it as clear as possible what each example is demonstrating.

eric-wieser: Or at least, do it for these two blocks which are already separate.

ktbarrett: Done.

documentation/source/coroutines.rst
```diff
@@ -153,52 +153,80 @@ Generator-based coroutines
...
-    async def wait_100ns():
-        for _ in range(10):
-            await wait_10ns()
+    @cocotb.coroutine
+    def start_clock(clk):
+        # generator-based coroutines can still be forked
+        cocotb.fork(clock(clk, 5, units='ns'))
+        yield RisingEdge(clk)
+
+Async coroutines can yielded in generator-based coroutines.
```
eric-wieser, line 182:
```suggestion
Async coroutines can be yielded in generator-based coroutines.
```

eric-wieser:
For reference: https://cocotb--1576.org.readthedocs.build/en/1576/coroutines.html

documentation/source/coroutines.rst
```diff
@@ -5,80 +5,50 @@ Coroutines
...
-        """Wait for a packet but time out if nothing arrives"""
-        tout_trigger = Timer(timeout, units='ns')
-        result = yield [tout_trigger, RisingEdge(dut.ready)]
-        if result is tout_trigger:
-            raise TestFailure("Timed out waiting for packet")
-
-
-Coroutines can be forked for parallel operation within a function of that code and
-the forked code.
+Coroutines can used with :func:`~cocotb.fork` for concurrent operation.
```
cmarqu, line 46: can be used

documentation/source/coroutines.rst
```diff
@@ -87,28 +57,28 @@ the forked code.
...
 
 
-Coroutines can be joined to end parallel operation within a function.
+Forked coroutines can be used in an :keyword:`await` statement to block until the forked coroutine finishes.
 
 .. code-block:: python3
 
     @cocotb.test()
-    def test_count_edge_cycles(dut, period=1, clocks=6):
+    async def test_count_edge_cycles(dut, period=1, clocks=6):
```
cmarqu, line 71: We could switch to period_ns while we are at it.

documentation/source/coroutines.rst
```diff
@@ -87,28 +57,28 @@ the forked code.
...
         cocotb.fork(Clock(dut.clk, period, units='ns').start())
-        yield RisingEdge(dut.clk)
+        await RisingEdge(dut.clk)
 
         timer = Timer(period + 10)
```
cmarqu, line 75: Use the units argument

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,90 @@ type of generator function that can be iterated with ``async for``:
...
+            count = await monitor()
+            assert count == (prev_count + 1)
+            prev_count = count
+
+You may also see syntax like ``yield [trigger_a, trigger_b, ...]``, which is syntactic sugar for :class:`~cocotb.triggers.First`.
+
+.. code-block:: python3
+
+    @cocotb.coroutine
+    def run_for(coro, time, units):
```
cmarqu, line 214: You are not actually using time in the function body

documentation/source/coroutines.rst
```diff
@@ -202,3 +143,90 @@ type of generator function that can be iterated with ``async for``:
...
+Coroutine functions would be decorated with :class:`~cocotb.decorators.coroutine` and would use :keyword:`yield` to block on other coroutines or triggers.
+You may see existing code that uses this syntax for coroutines, but do not worry, it is compatible with async coroutines.
+
+Any object that can be used in an :keyword:`await` statement can also be used in a :keyword:`yield` statement while in a generator-based coroutine;
+including triggers like :class:`~cocotb.triggers.Timer`.
+
+.. code-block:: python3
+
+    @cocotb.coroutine
+    def clock(signal, half_period, half_period_units):
```
cmarqu, line 164: Maybe call this eg myclock to make it clear even to a quick reader that this isn't our Clock class when used later

eric-wieser:
Rtd failure is real, but not the fault of this PR. #1603 tracks it, and our docs will not build at all until we fix it.

PR #1578, opened 2020-04-05
Update modelsim.def to fix compilation errors

themperek:
Some function definitions ware missing from https://github.com/cocotb/cocotb/pull/1424 that was casing FLI compilation error on Windows.


.gitignore
```diff
@@ -2,6 +2,7 @@
 
 # C extensions
 *.so
+cocotb/share/def/*.a
```
eric-wieser, line 5: I think this would be clearer inside the def directory

PR #1579, opened 2020-04-05
Cleanup of several Python 2 legacy code styles

ktbarrett:
Solves parts of #1289, mostly the low hanging fruit. I'd like to get most of the Python 2 cleanup in before we release 1.4.

 * Remove `from __future__ import` line from Python files (but not Makefile.pylib.*)
 * Remove `(object)` from `class MyClass(object)`
 * use `type(self)` instead of `self.__class__`
 * Remove all `__nonzero__` functions which alias `__bool__`
 * make `u""` strings normal strings (e.g. in documentation/source/conf.py)

I also started making the `__name__` => `__qualname__` changes, but they will require a more deft touch due to the schizophrenic usage of `__name__` in `decorators.py` and `regression.py`.

Holding off review until CI starts passing.

### DIFF
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -32,7 +32,7 @@ import sys
 
 
 # backport of Python 3.7's contextlib.nullcontext
-class nullcontext(object):
+class nullcontext:
     """Context manager that does no additional processing.
     Used as a stand-in for a normal context manager, when a particular
     block of code is only sometimes used with a normal context manager:
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -383,7 +383,7 @@ class BinaryValue:
         for char in string:
             if char not in BinaryValue._permitted_chars:
                 raise ValueError("Attempting to assign character %s to a %s" %
-                                 (char, self.__class__.__name__))
+                                 (char, type(self).__name__))
         self._str = string
         self._adjust()
 
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -61,13 +61,13 @@ def _clog2(val):
         exp += 1
 
 
-class BinaryRepresentation():  # noqa
+class BinaryRepresentation:  # noqa
     UNSIGNED         = 0  #: Unsigned format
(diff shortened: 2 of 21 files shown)


eric-wieser:
> I'd like to get most of the Python 2 cleanup in before we release 1.4.

Agreed, although I was hoping that fruit would bait in new contributors, rather than consuming the time of experienced ones! Still, better to have it than not.

cocotb/_py_compat.py
```diff
@@ -32,7 +32,7 @@
 
 
 # backport of Python 3.7's contextlib.nullcontext
-class nullcontext(object):
+class nullcontext():
```
eric-wieser, line 35:
```suggestion
class nullcontext:
```
Is the style I'm more used to. Not a big deal

ktbarrett: I'm indifferent, but there are existing classes that use empty parens, so that's what I went with.

ktbarrett:
@eric-wieser I tried getting a colleague to help out, but he didn't seem all that interested. I'm working according to a schedule. I don't want to tackle async-ifying the tests until I split them up. I don't want to do `__qualname__` until I do the refactor on the scheduler/regression manager, which is blocked by my `fork` PR. I'd rather not have two PR that touch the scheduler open at once, if the fork change is decidedly not going to make 1.4 than I'll let it rot and work on the refactor. Adding type annotations are going to need some research before I tackle it.

As for the `type()` change, I see one SO answer that say it's percentages faster, but it's a matter of consistency (there are occurrences of both in the code base) and preference (functions are better for numerous and, in this case, petty, reasons).

cocotb/_py_compat.py
eric-wieser: Python docs seem to strongly prefer `class C:` over `class C():`, so I'd appreciate if you could change this.

ktbarrett: Everywhere, or just on the classes I've touched?

eric-wieser: Up to you

ktbarrett: Scope crept.

PR #1583, opened 2020-04-06
Add a missing `static` on a private function

eric-wieser:


eric-wieser:
RTD needs it's cache clearerd, that failure is #1577

imphil:
RTD cache is cleared, master looks good again.

eric-wieser:
My memory is that we don't have control over the cache for the PR builds, unfortunately.

PR #1586, opened 2020-04-06
Use sets instead of dicts with meaningless values

eric-wieser:
A dict where all the values are true is just a set with extra overhead.

Split from #1510 for ease of review


eric-wieser:
ReadTheDocs failure is a caching issue beyond our control with their PR builder, which has not taken the changes from #1577.

PR #1588, opened 2020-04-06
Improve windows support in Makefiles

themperek:
Minimal PR to make Widows work with anaconda and msys2 (based on #1462).  
Something like #1565 would also help and keep user code cleaner.

- Set `PYTHONHOME` in order to properly populate `sys.path` in an embedded python interpreter.
- Set `PATH` to include a path to cocotb compiled libraries.
- Set an appropriate path format.
- Cleanup simulator Makefiles (more can be done still).

Tested with Icarus, Aldec Riviera, Questa, Modelsim (32)

Resolves: #1064 #428 #498 #436 #354


cocotb/share/makefiles/Makefile.sim
```diff
@@ -136,11 +136,7 @@ define check_for_results_file
     @test -f $(COCOTB_RESULTS_FILE) || (echo "ERROR: $(COCOTB_RESULTS_FILE) was not written by the simulation!" >&2 && exit 1)
 endef
 
-# Append directory containing the cocotb Python package to PYTHONPATH
-# XXX: This is only needed if cocotb is not installed into the default
-# Python search path.
-PYTHONPATH := $(COCOTB_PY_DIR):$(PYTHONPATH)
-export PYTHONPATH
+PYTHONPATH := $(PYTHONPATH):$(PWD)
```
eric-wieser, line 139: Any idea why we need `PWD`?

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -35,3 +35,13 @@ ifeq ($(PYTHON_LIBDIR),None)
...
+
+# On Windows there is no support for @rpath/LD_LIBRARY_PATH so we need to set PATH to find cocotb libraries
+# https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order?redirectedfrom=MSDN#standard-search-order-for-desktop-applications
+export PATH := $(LIB_DIR):$(PATH)
+
+# Set PYTHONHOME to properly populate sys.path in embedded python interpreter
+export PYTHONHOME := $(shell $(PYTHON_BIN) -c "from distutils.sysconfig import get_config_var; print(get_config_var('prefix'))")
+
+# For tcl scripts the path to dll has to be in form of C:/User/..
+export LIB_DIR := $(shell sh -c 'cd $(LIB_DIR) && pwd -W')
```
eric-wieser, line 47: I don't think these belong in this file - `pylib` is documented as:
```
# All common python related rules, kept only for downstream C extensions modules
# which copy the pattern of the endian_swapper HAL example. In future these modules
# should use `setuptools` to build their HAL interface
```

cocotb/share/makefiles/simulators/Makefile.questa
```diff
@@ -114,25 +114,6 @@ else
...
-MINGW_BIN_DIR = $(shell dirname $(shell :; command -v gcc))
-
-OLD_PATH := $(shell echo "$(PATH)" | sed 's/(/\\(/g' | sed 's/)/\\)/g' | sed 's/ /\\ /g')
-
-LIB_LOAD := PATH=$(MINGW_BIN_DIR):$(OLD_PATH):$(LIB_DIR)
-NEW_PYTHONPATH := $(shell echo "$(PYTHONPATH)" | sed -e 's/\\/\//g' -e 's/\([a-zA-Z]\):\//\/\1\//g' -e 's/;/:/g')
-else
-LIB_LOAD :=
-NEW_PYTHONPATH := $(PYTHONPATH)
-endif
```
eric-wieser, line 134: How do we know we no longer need this?

examples/endian_swapper/tests/Makefile
```diff
@@ -30,17 +30,16 @@
...
 PWD=$(shell pwd)
 
 ifeq ($(OS),Msys)
-WPWD=$(shell sh -c 'pwd -W')
-PYTHONPATH := $(PWD)/../cosim;$(PYTHONPATH)
+    WPWD=$(shell sh -c 'pwd -W')
 else
-WPWD=$(shell pwd)
-PYTHONPATH := $(PWD)/../cosim:$(PYTHONPATH)
+    WPWD=$(shell pwd)
```
eric-wieser, line 38:
```suggestion
    WPWD=$(PWD)
```

examples/endian_swapper/tests/Makefile
```diff
@@ -30,17 +30,16 @@
 # Default to verilog
 TOPLEVEL_LANG ?= verilog
 
-WPWD=$(shell sh -c 'pwd -W')
 PWD=$(shell pwd)
 
 ifeq ($(OS),Msys)
-WPWD=$(shell sh -c 'pwd -W')
-PYTHONPATH := $(PWD)/../cosim;$(PYTHONPATH)
+    WPWD=$(shell sh -c 'pwd -W')
```
eric-wieser, line 36:
```suggestion
    WPWD=$(shell cygpath -m "$(PWD)")
```

cocotb/share/makefiles/simulators/Makefile.questa
themperek: Do you ask about `MINGW_BIN_DIR` ? The only argument I have is that I spend quite a bit on time on this tested with multiple configurations and never seen this would be needed. 
I also do not think we need to fix here some strange user environment (that is very easy on Windows).  Our compilation method has changed.

cocotb/share/makefiles/Makefile.pylib.Msys
eric-wieser:
```suggestion

# On Windows there is no support for @rpath/LD_LIBRARY_PATH so we need to set PATH to find cocotb libraries
# https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order?redirectedfrom=MSDN#standard-search-order-for-desktop-applications
export PATH := $(LIB_DIR):$(PATH)

# Set PYTHONHOME to properly populate sys.path in embedded python interpreter
export PYTHONHOME := $(shell $(PYTHON_BIN) -c "from distutils.sysconfig import get_config_var; print(get_config_var('prefix'))")

# For tcl scripts the path to dll has to be in form of C:/User/..
export LIB_DIR := $(shell cygpath -m "$(LIB_DIR)")
```

cocotb/share/makefiles/Makefile.inc
```diff
@@ -55,7 +55,10 @@ endif
 export OS
 
 # this ensures we use the same python as the one cocotb was installed into
-PYTHON_BIN ?= $(shell cocotb-config --python-bin)
+# realpath for Windows compatibility
```
eric-wieser, line 58: What exactly is the effect of `realpath` on windows?

themperek: `cocotb-config --python-bin` will return `C:\User\` and `realpath` will do `/c/User` same above in the file (would need to escape `\` in cocotb-config maybe).

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: I was thinking to avoid `cygpath` (no good reason) as long as possible but any is OK for me.

cocotb/share/makefiles/Makefile.sim
themperek: It is assumed that in user makefiles. If not would need to add this in every makefile or add to `sys.path` example in `__init__.py`

cocotb/share/makefiles/Makefile.inc
eric-wieser: Ah, so this is `cygpath -u`, to convert windows path to unix paths.

```suggestion
# realpath for to convert windows path to unix paths, like cygpath -u
```

cocotb/share/makefiles/Makefile.pylib.Msys
eric-wieser: I'd argue `cygpath` is more explicit here, as well as behaving predictably when paths do not exist.

cocotb/share/makefiles/Makefile.sim
eric-wieser: What puzzles me is how when I run `python` normally, `sys.path` automatically contains `.`.

Is this not the default for embedded python?

cocotb/share/makefiles/simulators/Makefile.questa
themperek: If you ask about `NEW_PYTHONPATH` not  need all this if `PYTHONHOME` is set.

cocotb/share/makefiles/Makefile.sim
themperek: This is strange but this is mention in multiple places (I can google for you). Founded it by chance.

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -44,4 +44,4 @@ export PATH := $(LIB_DIR):$(PATH)
 export PYTHONHOME := $(shell $(PYTHON_BIN) -c "from distutils.sysconfig import get_config_var; print(get_config_var('prefix'))")
 
 # For tcl scripts the path to dll has to be in form of C:/User/..
-export LIB_DIR := $(shell sh -c 'cd $(LIB_DIR) && pwd -W')
+export LIB_DIR := $(shell cygpath -m "$(LIB_DIR)")
```
eric-wieser, line 47: Perhaps this should be single quotes, I don't know enough about makefiles.

cocotb/share/makefiles/Makefile.inc
```diff
@@ -55,7 +55,10 @@ endif
 export OS
 
 # this ensures we use the same python as the one cocotb was installed into
-PYTHON_BIN ?= $(shell cocotb-config --python-bin)
+# realpath for to convert windows paths to unix paths, like cygpath -u
```
eric-wieser, line 58:
```suggestion
# realpath to convert windows paths to unix paths, like cygpath -u
```

cocotb/share/makefiles/Makefile.sim
themperek: Seems like. Could add to `sys.path` just before: https://github.com/cocotb/cocotb/blob/master/cocotb/regression.py#L151

cocotb/share/makefiles/simulators/Makefile.questa
eric-wieser: `MINGW_BIN_DIR` was my concern, I'm not worried about whatever garbage is going on in PYTHON_PATH here.

-

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: Line 44 (above) uses `"` so should be safe.

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -35,3 +35,13 @@ ifeq ($(PYTHON_LIBDIR),None)
...
+
+# On Windows there is no support for @rpath/LD_LIBRARY_PATH so we need to set PATH to find cocotb libraries
+# https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order?redirectedfrom=MSDN#standard-search-order-for-desktop-applications
+export PATH := $(LIB_DIR):$(PATH)
+
+# Set PYTHONHOME to properly populate sys.path in embedded python interpreter
+export PYTHONHOME := $(shell $(PYTHON_BIN) -c 'from distutils.sysconfig import get_config_var; print(get_config_var("prefix"))')
+
+# For tcl scripts the path to dll has to be in form of C:/User/..
+export LIB_DIR := $(shell cygpath -m '$(LIB_DIR)')
```
eric-wieser, line 47: I don't think these belong in this file - `pylib` is documented as:
```
# All common python related rules, kept only for downstream C extensions modules
# which copy the pattern of the endian_swapper HAL example. In future these modules
# should use `setuptools` to build their HAL interface
```

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -35,3 +35,13 @@ ifeq ($(PYTHON_LIBDIR),None)
 endif
 
 PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'import distutils.sysconfig; print( distutils.sysconfig.get_python_inc() )')
+
+# On Windows there is no support for @rpath/LD_LIBRARY_PATH so we need to set PATH to find cocotb libraries
+# https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order?redirectedfrom=MSDN#standard-search-order-for-desktop-applications
+export PATH := $(LIB_DIR):$(PATH)
+
+# Set PYTHONHOME to properly populate sys.path in embedded python interpreter
+export PYTHONHOME := $(shell $(PYTHON_BIN) -c 'from distutils.sysconfig import get_config_var; print(get_config_var("prefix"))')
```
eric-wieser, line 44: Shouldn't we do this on all platforms?

Also, the docs seem to suggest that PYTHONHOME is set automatically based on the python interpreter used. Can you confirm it fails without it?

cocotb/share/makefiles/Makefile.sim
eric-wieser: Let's put off this converation for now, this isn't actually a change made by this patch.

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: >Shouldn't we do this on all platforms?

I think would not harm and possibly help.

>Can you confirm it fails without it?

On Windows yes. It will fail to `import site` as the first error.

eric-wieser: From [the python docs](https://docs.python.org/3.9/using/windows.html#finding-modules):

> For those who want to bundle Python into their application or distribution, the following advice will prevent conflicts with other installations:
> 
> * Include a ._pth file alongside your executable containing the directories to include. This will ignore paths listed in the registry and environment variables, and also ignore site unless import site is listed.
> 
> * If you are loading python3.dll or python37.dll in your own executable, explicitly call Py_SetPath() or (at least) Py_SetProgramName() before Py_Initialize().
> 
> * Clear and/or overwrite PYTHONPATH and set PYTHONHOME before launching python.exe from your application.
> 
> * If you cannot use the previous suggestions (for example, you are a distribution that allows people to run python.exe directly), ensure that the landmark file (Lib\os.py) exists in your install directory. (Note that it will not be detected inside a ZIP file, but a correctly named ZIP file will be detected instead.)

So I think this is ok for now, although calling `Py_SetProgramName` may make more sense in future.

themperek:
I forgot to include the correction for msys2. Now is in.

cocotb/share/makefiles/Makefile.pylib.Msys
eric-wieser: > I think would not harm and possibly help.

I think we should do this then.

cocotb/_build_libs.py
```diff
@@ -77,6 +77,9 @@ def get_ext_filename(self, ext_name):
         head, tail = os.path.split(filename)
         tail_split = tail.split(".")
 
+        # mingw on msys2 uses `-` as seperator
+        tail_split = tail_split[0].split("-")
```
eric-wieser, line 81: Should we split on `-cpython` or `.cpython` instead?

cocotb/share/makefiles/Makefile.pylib.Msys
```diff
@@ -35,3 +35,13 @@ ifeq ($(PYTHON_LIBDIR),None)
 endif
 
 PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'import distutils.sysconfig; print( distutils.sysconfig.get_python_inc() )')
+
+# On Windows there is no support for @rpath/LD_LIBRARY_PATH so we need to set PATH to find cocotb libraries
+# https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order?redirectedfrom=MSDN#standard-search-order-for-desktop-applications
+export PATH := $(LIB_DIR):$(PATH)
```
eric-wieser, line 41: Can we just put this inside the `ifeq ($(OS),Msys)` in `Makefile.inc`? It looks unrelated to python, so doesn't seem to belong in this file.

cocotb/share/makefiles/Makefile.pylib.Msys
eric-wieser: You did this, then force-pushed a change that removed it.

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: BTW: We can remove `pylib`s files in general also `Makefile.inc`

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: Should be ok?

cocotb/share/makefiles/Makefile.pylib.Msys
themperek: Same with `LIB_DIR` so ...

eric-wieser: We can't, because `.pylib` is for users who copied the endian example, as I remarked  in https://github.com/cocotb/cocotb/pull/1588#discussion_r404159930.

For that reason, we should not be adding anything new to it, unless it's needed for those users.

cocotb/share/makefiles/Makefile.inc
```diff
@@ -75,10 +81,17 @@ ifeq ($(SIM_DEFINE),XCELIUM)
...
 LIB_EXT    := so
 endif
+
+ifeq ($(OS),Msys)
+    # On Windows there is no support for @rpath/LD_LIBRARY_PATH so we need to set PATH to find cocotb libraries
+    # https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order?redirectedfrom=MSDN#standard-search-order-for-desktop-applications
+    export PATH := $(LIB_DIR):$(PATH)
+
+    # For tcl scripts the path to dll has to be in form of C:/User/..
+    export LIB_DIR := $(shell cygpath -m '$(LIB_DIR)')
```
eric-wieser, line 96:
```suggestion
    LIB_DIR := $(shell cygpath -m '$(LIB_DIR)')
```
This isn't exported on other platforms, is that intentional?

cocotb/share/makefiles/Makefile.inc
```diff
@@ -55,7 +55,13 @@ endif
...
 
 # this ensures we use the same python as the one cocotb was installed into
-PYTHON_BIN ?= $(shell cocotb-config --python-bin)
+# realpath to convert windows paths to unix paths, like cygpath -u
+PYTHON_BIN ?= $(realpath $(shell cocotb-config --python-bin))
+
+# Set PYTHONHOME to properly populate sys.path in embedded python interpreter
+export PYTHONHOME := $(shell $(PYTHON_BIN) -c 'from distutils.sysconfig import get_config_var; print(get_config_var("prefix"))')
+
+LIB_DIR=$(COCOTB_PY_DIR)/cocotb/libs
```
eric-wieser, line 64: Is there a reason this happens before the other includes?

cocotb/share/makefiles/Makefile.inc
```diff
@@ -75,10 +81,17 @@ ifeq ($(SIM_DEFINE),XCELIUM)
 SIM_DEFINE = IUS
 endif
 
-export LIB_DIR=$(COCOTB_PY_DIR)/cocotb/libs
```
eric-wieser, line 78: Or phrased another way, did this line need to move?

cocotb/share/makefiles/Makefile.inc
themperek: Lagecy

cocotb/share/makefiles/Makefile.inc
themperek: Legacy

cocotb/share/makefiles/Makefile.sim
eric-wieser, line 139: This is backward, `PWD` is always inserted first on the path by a regular python interpreter
```suggestion
PYTHONPATH := $(PWD):$(PYTHONPATH)
```

themperek: This will fail because of `:` in case `PYTHONPATH` is empty (heh) so need to guard this with `ifdef` or add to `sys.path` inside cocotb/embedded interpreter.

elgorwi: Shouldn’t the path separator be ‘;’ rather than ‘ :’ on windows?

eric-wieser: Hmm, good catch. I think the msys path-conversion stuff doesn't kick in unless the colon has a character after it.

themperek: No. msys understand `:`.

eric-wieser: (edited my comment above)

themperek: @eric-wieser 
```
['C:\\Users\\Tomek\\git\\cocotb\\examples\\adder\\model',
 'C:\\Users\\Tomek\\git\\cocotb\\examples\\adder\\tests',
 'C:\\Users\\Tomek\\Miniconda3\\python37.zip',
 'C:\\Users\\Tomek\\Miniconda3\\DLLs',
 'C:\\Users\\Tomek\\Miniconda3\\lib',
 'C:\\Users\\Tomek\\iverilog\\bin',
 'C:\\Users\\Tomek\\Miniconda3',
 'C:\\Users\\Tomek\\Miniconda3\\lib\\site-packages',
 'c:\\users\\tomek\\git\\cocotb-coverage',
 'c:\\users\\tomek\\git\\cocotb-test',
 'c:\\users\\tomek\\git\\cocotb',
 'C:\\Users\\Tomek\\Miniconda3\\lib\\site-packages\\win32',
 'C:\\Users\\Tomek\\Miniconda3\\lib\\site-packages\\win32\\lib',
 'C:\\Users\\Tomek\\Miniconda3\\lib\\site-packages\\Pythonwin']
```

eric-wieser: Discussed out of band - this issue is minor, and @themperek have a solution to eliminate `PYTHONPATH` entirely. Let's go ahead and merge.

PR #1590, opened 2020-04-06
Mention proper way of setting make vars

### LINKED ISSUE #1552: Document that users should use EXTRA_ARGS=... make instead of make EXTRA_ARGS=...

cmarqu:
Document that users should use the syntax ``EXTRA_ARGS=... make`` instead of ``make  EXTRA_ARGS=...`` because some Makefiles append to these variables internally, and the second syntax would break them.

Closes #1552

### DIFF
--- a/documentation/source/building.rst
+++ b/documentation/source/building.rst
@@ -1,3 +1,5 @@
+.. _building:
+
 ***************************************
 Build options and Environment Variables
 ***************************************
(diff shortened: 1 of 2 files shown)


cmarqu:
I repeated them because this is a consultation section - you look up what you want to know but don't necessarily read any of the surrounding paragraphs.

cmarqu:
Ping @eric-wieser

documentation/source/building.rst
```diff
@@ -1,3 +1,5 @@
+.. _building:
```
eric-wieser, line 1: This shouldn't be needed, you can already reference this as ``:doc:`building` ``

cmarqu: This is for resilience against file renamings likely to happen in the future.

PR #1592, opened 2020-04-07
Address issues with unexpected MODULE values

### LINKED ISSUE #1591: Problems with MODULE
If the `MODULE` env variable is not set cocotb fails hard by raising (an `ImportError` of all things...) in the startup routine. It seems to me it should just run and exit with no failure because there are no tests to run.

`MODULE` might also accidentally contain trailing or leading separators, this isn't handled, and tests show it fails in an ugly way. `HOOKS`, whose usefulness is debatable, suffers the same problem.

ktbarrett:
Fixes #1591. Also addresses issue with `HOOKS`.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -164,12 +164,11 @@ def _initialise_testbench(root_name):
     test_str = os.getenv('TESTCASE')
     hooks_str = os.getenv('COCOTB_HOOKS', '')
 
-    if not module_str:
-        raise ImportError("Environment variables defining the module(s) to " +
-                          "execute not defined.  MODULE=\"%s\"" % (module_str))
+    if module_str is None:
+        raise ValueError("Environment variable MODULE, which defines the module(s) to execute, is not defined.")
 
-    modules = module_str.split(',')
-    hooks = hooks_str.split(',') if hooks_str else []
+    modules = [s.strip() for s in module_str.split(',') if s.strip()]
+    hooks = [s.strip() for s in hooks_str.split(',') if s.strip()]
 
     global regression_manager
 
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -144,6 +144,7 @@ class RegressionManager:
         # Test Discovery
         ####################
 
+        have_tests = False
         for module_name in self._modules:
             try:
                 self.log.debug("Python Path: " + ",".join(sys.path))
@@ -180,6 +181,10 @@ class RegressionManager:
             for thing in vars(module).values():
                 if hasattr(thing, "im_test"):
                     self._init_test(thing)
+                    have_tests = True
+
+        if not have_tests:
+            self.log.warning("No tests were discovered")
 
         self._queue.sort(key=lambda test: (test.stage, test._id))
 
--- a/tests/test_cases/test_module_var_empty/Makefile
+++ b/tests/test_cases/test_module_var_empty/Makefile
@@ -1,0 +1,5 @@
+include ../../designs/sample_module/Makefile
+
+# test MODULE is empty
+
+MODULE := " "
--- a/tests/test_cases/test_module_without_tests/Makefile
+++ b/tests/test_cases/test_module_without_tests/Makefile
@@ -1,0 +1,5 @@
+include ../../designs/sample_module/Makefile
+
+# test MODULE is set, but there are no tests across any of the modules
+
+MODULE=test_nothing
--- a/tests/test_cases/test_module_var_messy/Makefile
+++ b/tests/test_cases/test_module_var_messy/Makefile
@@ -1,0 +1,5 @@
+include ../../designs/sample_module/Makefile
+
+# test MODULE contains leading and trailing separators
+
+MODULE=" , test_nothing  ,"
--- a/tests/test_cases/test_module_var_messy/test_nothing.py
+++ b/tests/test_cases/test_module_var_messy/test_nothing.py
@@ -1,0 +1,1 @@
+""" This test module is purposefully empty """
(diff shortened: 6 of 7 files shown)


ktbarrett:
Added support for trailing and leading whitespace in elements.

imphil:
Code looks good modulo the empty MODULE error message change. Let's discuss that in https://github.com/cocotb/cocotb/issues/1591 and then return to this PR.

eric-wieser:
I put some comments on  #1591.

eric-wieser:
A concern orthogonal to the discussion there - I'm not convinced naming tests by their issue numbers are useful, unless the corner case is hard to describe concisely. More useful test names would be:

* `test_module_var_unset`
* `test_module_var_empty`
* `test_module_without_tests`

ktbarrett:
Addressed comments here and in the linked issue.

If module is not set, we throw an error and exist uncleanly (without result.xml).
If module is empty (or no tests were found in any of the modules), we warn that no tests were run and exit cleanly (with result.xml).

Because of https://github.com/cocotb/cocotb/issues/1591#issuecomment-610414495, I removed the test for unset `MODULE` because the makefiles make testing that impossible right now.

ktbarrett:
@imphil We can start returning non-0 exit codes once we move to something like #1562. The regression system will have control start to finish and the simulator will sit in the middle.

imphil:
Yeah, that's a whole different topic.

cocotb/__init__.py
```diff
@@ -164,12 +164,12 @@ def _initialise_testbench(root_name):
     test_str = os.getenv('TESTCASE')
     hooks_str = os.getenv('COCOTB_HOOKS', '')
 
-    if not module_str:
+    if module_str is None:
         raise ImportError("Environment variables defining the module(s) to " +
                           "execute not defined.  MODULE=\"%s\"" % (module_str))
```
eric-wieser, line 169: This is dumb, if the module string was not defined then there's no poitn printing it, especially if we're going to imply it was set to the string `None`.
```suggestion
        raise ImportError("Environment variables MODULE which defines the module(s) to 
                          "execute is not defined.")
```

cocotb/regression.py
```diff
@@ -180,6 +181,10 @@ def initialise(self):
             for thing in vars(module).values():
                 if hasattr(thing, "im_test"):
                     self._init_test(thing)
+                    have_tests = True
+
+        if not have_tests:
```
eric-wieser, line 186: Could you use `if self._queue` here? Or does that not work?

tests/test_cases/test_module_var_empty/Makefile
```diff
@@ -0,0 +1,6 @@
+include ../../designs/sample_module/Makefile
+
+# test MODULE is empty
+
+MODULE := ""
+
```
eric-wieser, line 6: nit:
```suggestion
```

tests/test_cases/test_module_without_tests/Makefile
```diff
@@ -0,0 +1,5 @@
+include ../../designs/sample_module/Makefile
+
+MODULE=test_nothing
+
+# test MODULE is set, but there are no tests across any of the modules
```
eric-wieser, line 5: nit, to match the others:
```suggestion
# test MODULE is set, but there are no tests across any of the modules

MODULE=test_nothing
```

cocotb/regression.py
ktbarrett: Thing are only added to the queue if they initialize properly and are not skipped. What we want to see is if we find anything marked as a `test`.

cocotb/__init__.py
ktbarrett: I feel like `ImportError` is also improper, will change to `ValueError`.

cocotb/regression.py
eric-wieser: I'd perhaps argue we want a warning for either case, but I'm happy to call that out of scope.

cocotb/__init__.py
eric-wieser: Hadn't even noticed that, good idea.

cocotb/__init__.py
```diff
@@ -164,12 +164,11 @@ def _initialise_testbench(root_name):
     test_str = os.getenv('TESTCASE')
     hooks_str = os.getenv('COCOTB_HOOKS', '')
 
-    if not module_str:
-        raise ImportError("Environment variables defining the module(s) to " +
-                          "execute not defined.  MODULE=\"%s\"" % (module_str))
+    if module_str is None:
+        raise ValueError("Environment variable MODULE, which defines the module(s) to execute, is not defined.")
```
eric-wieser, line 168: Line's a little long, but whatever

ktbarrett:
So this is weird.

https://travis-ci.org/github/cocotb/cocotb/jobs/672124226#L4055-L4083

`MODULE` is set to the empty string, but the "MODULE not set" error is raised, indicating that `getenv` returned `None` instead of the empty string. Does `getenv` or the environment work differently on Windows? I don't know much of anything about Windows.

eric-wieser:
Sounds like you ran into https://github.com/msys2/MINGW-packages/issues/1599

eric-wieser:
CI failure is #1603

eric-wieser:
Does this justify a newsfragment? In that whitespace is now stripped in the `MODULE` variable, meaning it can be wrapped across multiple lines?

ktbarrett:
@eric-wieser I'll do some additional cleanup on other env var parsing in a later PR and I'll make a note then saying they are all whitespace and empty element tolerant.

PR #1594, opened 2020-04-07
Split test_cocotb into multiple modules

ktbarrett:
Split `test_cocotb.py` from a 1400 loc file into 15 individual test files under ~200 loc each. I did no touchups, simply moved existing code, PEP8 violations and all, to new files. I moved a few tests into issue test dirs. I hopped into a REPL and did some processing on the logs to ensure I didn't miss any tests except those I moved out. I'm open for suggestions on alternative groupings or naming.

~~This is built on top of #1592, so it's on hold until that's in. Hopefully this will be a motivating example for that change.~~


tests/test_cases/issue_1223/issue_1223.py
```diff
@@ -0,0 +1,12 @@
+import cocotb
```
eric-wieser, line 1: I think this would make more sense in test_handle. Py

tests/test_cases/issue_637/issue_637.py
```diff
@@ -0,0 +1,18 @@
+import cocotb
+from cocotb.triggers import Timer, NullTrigger
+
+
+@cocotb.test()
+def test_stack_overflow(dut):
```
eric-wieser, line 6: Can we put this in `test_scheduler`?

tests/test_cases/issue_843/issue_843.py
```diff
@@ -0,0 +1,40 @@
+import cocotb
+from cocotb.triggers import Event, Timer
+
+
+@cocotb.test()
+def test_yield_list_stale(dut):
+    """ Test that a trigger yielded as part of a list can't cause a spurious wakeup """
+    # gh-843
```
eric-wieser, line 8: This looks like it would fit well in `test_concurrency_primitives.py`

tests/test_cases/issue_637/issue_637.py
eric-wieser: Or perhaps `NullTrigger` can be thought of a concurrency primitive, since I think it lets another coroutine be scheduled?

ktbarrett: It was made in reaction to a scheduler bug, so `test_scheduler` makes sense.

tests/test_cases/issue_843/issue_843.py
ktbarrett: It seems you favor not moving tests tied to issues into their own dirs. Do you think that other issue dirs should be absorbed into test_cocotb?

eric-wieser: I don't think you should increase work for yourself and do that as part of this PR, but in general, yes.

It's a good idea to _reference_ issues in the tests, but naming after the issue makes it impossible to tell when writing a new test whether a similar one already exists - especially now that you've spent a lot of effort categorizing the `test_cocotb` tests!

Only for an exceptionally contrived case would I recommend using an issue number.

imphil:
General comment for all new files, including Makefiles: please add a license header in the form


```
```

ktbarrett:
@eric-wieser All tests kept in test_cocotb.

@imphil Added header to all new files.

tests/test_cases/test_cocotb/test_edge_triggers.py
```diff
@@ -0,0 +1,244 @@
...
+    Test that the result of trigger expression have a predictable type
+    """
+    assert isinstance(RisingEdge(dut.clk), RisingEdge)
+    assert isinstance(FallingEdge(dut.clk), FallingEdge)
+    assert isinstance(Edge(dut.clk), Edge)
+    assert isinstance(NextTimeStep(), NextTimeStep)
+    assert isinstance(ReadOnly(), ReadOnly)
+    assert isinstance(ReadWrite(), ReadWrite)
+
+    yield Timer(1)
```
eric-wieser, line 208: This is maybe a little out of place since it also touches ReadOnly etc, but is better than before, so I don't think is worth worrying about

tests/test_cases/test_cocotb/test_cocotb.py
```diff
@@ -27,125 +27,17 @@
...
-import warnings
-from fractions import Fraction
-from decimal import Decimal
-from math import isclose
-
 """
 A set of tests that demonstrate cocotb functionality
 
 Also used as regression test of cocotb capabilities
 """
```
eric-wieser, line 34:
```suggestion
"""
Helper functions used by multiple tests in this directory
"""
```

May as well replace the license header with a short one to

tests/test_cases/test_cocotb/Makefile
```diff
@@ -29,8 +29,23 @@
 
 include ../../designs/sample_module/Makefile
 
-MODULE := test_cocotb,test_deprecated,test_doctests
+MODULE := "\
+	test_cocotb,\
```
eric-wieser, line 33: This doesn't contain any tests any more, right?
```suggestion
```

tests/test_cases/test_cocotb/Makefile
```diff
@@ -29,8 +29,23 @@
...
+	test_tests,\
+	test_generator_coroutines,\
+	test_timing_triggers,\
+	test_scheduler,\
+	test_clock,\
+	test_edge_triggers,\
+	test_async_coroutines,\
+	test_handle,\
+	test_logging,\
+	"
```
imphil, line 47: Maybe as a follow-up we should think of enabling space-separated test lists being passed to MODULE to make all of this less ugly.

tests/test_cases/test_cocotb/test_generator_coroutines.py
```diff
@@ -0,0 +1,210 @@
...
+    yield Timer(10000)
+
+
+@cocotb.coroutine
+def syntax_error():
+    yield Timer(100)
+    fail
+
+
+#@cocotb.test(expect_error=True)
```
imphil, line 100: Can that be removed?

tests/test_cases/test_cocotb/test_edge_triggers.py
ktbarrett: I had thought to split it up, but I was trying to not introduce any new tests/change existing tests in this PR. I'm fine with scope creep on this though, will change.

eric-wieser: This one is probably more coherent as a single test than it is split up, but I don't care too much.

tests/test_cases/test_cocotb/Makefile
eric-wieser: Would it really make much difference?

I think using a `define` block would eliminate the need for the backslashes

ktbarrett: I tried that, but how the makefile uses the variable means it isn't interpreted correctly and tries to run the string as a command.

eric-wieser: Using `export MODULE` should fix that. Quotes at the use sie might too

PR #1595, opened 2020-04-07
Remove PYTHONPATH from the simulator invocation

themperek:
- Simplify the makefile and remove `PYTHONPATH` from the simulator invocation
- Append "" to sys.path when python is embedded that is normally included.
- Resolve minor issue with` PYTHONPATH` order  `PYTHONPATH := $(PYTHONPATH):$(PWD)`
- It is on the user to append and export `PYTHONPATH` if needed


examples/adder/tests/Makefile
```diff
@@ -37,7 +37,7 @@ else
 WPWD=$(shell pwd)
 endif
 
-PYTHONPATH := $(PWD)/../model:$(PYTHONPATH)
+export PYTHONPATH := $(PWD)/../model:$(PYTHONPATH)
```
eric-wieser, line 40: It's very tempting to remove this in a future patch and just use `import model.something` in the test.

themperek: Agree. Much nicer.  In the test, one can append to `sys.path` also rather than this. I can do in this PR?

eric-wieser: CI has already passed here, may as well merge this one first.

eric-wieser:
Some typos in the commit message, will fix when I merge

eric-wieser:
Hmm, on second thoughts this should have had a release note, since any user makefiles that set `PYTHONPATH` without `export` will now fail to work.

themperek:
@eric-wieser Ths is causing an **error** on all simulator that start in `SIM_BUILD` folder `cd $(SIM_BUILD)`

It was discussed here: https://github.com/cocotb/cocotb/pull/1063 and somehow decided.

Should revert this or make a patch to align all simulators?
On the bad side, it will produce all empty files in test directory @imphil .

eric-wieser:
It sounds like this means that `os.getcwd()` is set differently on different simulators, which is bad because it breaks python code attempting to `open("some_data_file")`.

Can you point to which simulators do that?

themperek:
https://github.com/cocotb/cocotb/search?q=%22cd+%24%28SIM_BUILD%29%22&unscoped_q=%22cd+%24%28SIM_BUILD%29%22

We should rather align all of them. Since this is also issued for where `results.xml` is located. This is not related to this issues caused by this.

eric-wieser:
So to summarize, we already fixed this issue in questa, but the other files are lagging behind?

That seems like a clear argument for changing those simulators to match.

themperek:
Ok, will do.

PR #1599, opened 2020-04-07
Align ghdl simulators with others based on #1063

themperek:
It is failing at the moment because of the wrong locations of `results.xml` and #1595


themperek:
Looks safe here: https://travis-ci.org/github/cocotb/cocotb/jobs/672143792 compared to before https://travis-ci.org/github/cocotb/cocotb/jobs/672149789

PR #1606, opened 2020-04-08
Fix libraries runpath on newer Debian

themperek:
Due to changes in Debian (default `--enable-new-dtagshas`), `rpath` has to be set on dependent libraries.
See: https://stackoverflow.com/questions/52018092/how-to-set-rpath-and-runpath-with-gcc-ld

### DIFF
--- a/cocotb/_build_libs.py
+++ b/cocotb/_build_libs.py
@@ -144,10 +144,12 @@ def _extra_link_args(lib_name=None, rpath=None):
     On Linux use`rpath`
     """
 
+    args = []
     if sys.platform == "darwin" and lib_name is not None:
-        return ["-Wl,-install_name,@rpath/%s.so" % lib_name]
-    elif sys.platform == "linux" and rpath is not None:
-        return ["-Wl,-rpath,%s" % rpath]
+        args += ["-Wl,-install_name,@rpath/%s.so" % lib_name]
+    if rpath is not None:
+        args += ["-Wl,-rpath,%s" % rpath]
+    return args
     else:
         return []
 
--- a/cocotb/_build_libs.py
+++ b/cocotb/_build_libs.py
@@ -137,14 +137,17 @@ class build_ext(_build_ext):
                 )
 
 
-def _extra_link_args(lib_name):
+def _extra_link_args(lib_name=None, rpath=None):
     """
-    Add linker `install_name` argument on osx to load dependencies from the directory
-    where vpi/vhpi/fli library is located
+    Add linker argument to load dependencies from the directory where vpi/vhpi/fli library is located
+    On osx use`install_name`
+    On Linux use`rpath`
     """
 
-    if sys.platform == "darwin":
+    if sys.platform == "darwin" and lib_name is not None:
         return ["-Wl,-install_name,@rpath/%s.so" % lib_name]
+    elif sys.platform == "linux" and rpath is not None:
+        return ["-Wl,-rpath,%s" % rpath]
     else:
         return []
 
@@ -201,7 +204,7 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
         os.path.join("cocotb", "libs", "libcocotbutils"),
         include_dirs=[include_dir],
         sources=[os.path.join(share_lib_dir, "utils", "cocotb_utils.cpp")],
-        extra_link_args=_extra_link_args("libcocotbutils"),
+        extra_link_args=_extra_link_args(lib_name="libcocotbutils", rpath="$ORIGIN"),
         extra_compile_args=_extra_cxx_compile_args,
     )
 
@@ -218,7 +221,7 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
         libraries=["cocotbutils"],
         library_dirs=python_lib_dirs,
         sources=[os.path.join(share_lib_dir, "gpi_log", "gpi_logging.cpp")],
-        extra_link_args=_extra_link_args("libgpilog"),
+        extra_link_args=_extra_link_args(lib_name="libgpilog", rpath="$ORIGIN"),
         extra_compile_args=_extra_cxx_compile_args,
     )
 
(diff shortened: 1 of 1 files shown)


eric-wieser:
I'm not sure I understand why `RUNPATH` isn't working for our use-case. Are you able to work that out?

eric-wieser:
Supposedly:

> The new-style DT_RUNPATH is not extended recursively when resolving transitive dependencies, as was the case with RPATH. That is, if we dlopen() libA.so, which in turn marks libB.so as DT_NEEDED, which in turn requires a libC.so — then the resolution for libC.so will visit only the RUNPATH given in libB.so, but not the RUNPATH given in libA.so (to the contrary, the old RPATH used to visit all these locations).
This behaviour was chosen deliberately, in compliance with the ELF spec, as can be seen in this glibc bug #13945 and the developer comment by Roland McGrath from 2002 mentioned therein.

themperek:
If I set `-rpath,$ORIGIN` on all depend libraries it seems to work with RUNPATH. I will try to update today.

eric-wieser:
That might be the more appropriate fix.

themperek:
It seems to work no test on Ubuntu and Windows WLS1.

cocotb/_build_libs.py
```diff
@@ -248,7 +251,7 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
             os.path.join(share_lib_dir, "gpi", "GpiCbHdl.cpp"),
             os.path.join(share_lib_dir, "gpi", "GpiCommon.cpp"),
         ],
-        extra_link_args=_extra_link_args("libgpi"),
+        extra_link_args=_extra_link_args(lib_name="libgpi", rpath="$ORIGIN"),
```
eric-wieser, line 254: Any idea why this needs `lib_name` but the ones below do not?

cocotb/_build_libs.py
```diff
@@ -137,14 +137,17 @@ def _gen_import_libs(self, def_dir):
...
+    Add linker argument to load dependencies from the directory where vpi/vhpi/fli library is located
+    On osx use`install_name`
+    On Linux use`rpath`
     """
 
-    if sys.platform == "darwin":
+    if sys.platform == "darwin" and lib_name is not None:
         return ["-Wl,-install_name,@rpath/%s.so" % lib_name]
+    elif sys.platform == "linux" and rpath is not None:
+        return ["-Wl,-rpath,%s" % rpath]
```
eric-wieser, line 150: Is it safe to do this on darwin too, as
```suggestion
    args = []
    if sys.platform == "darwin" and lib_name is not None:
        args += ["-Wl,-install_name,@rpath/%s.so" % lib_name]
    if rpath is not None:
        args += ["-Wl,-rpath,%s" % rpath]
    return args
```
not that it really

cocotb/_build_libs.py
themperek: On osx it seems to work the way this needs to be set on dependencies only. Not really hudge expert here.

cocotb/_build_libs.py
```diff
@@ -137,14 +137,19 @@ def _gen_import_libs(self, def_dir):
...
-    if sys.platform == "darwin":
-        return ["-Wl,-install_name,@rpath/%s.so" % lib_name]
+    args = []
+    if sys.platform == "darwin" and lib_name is not None:
+        args += ["-Wl,-install_name,@rpath/%s.so" % lib_name]
+    if rpath is not None:
+        args += ["-Wl,-rpath,%s" % rpath]
+    return args
     else:
         return []
```
eric-wieser, line 154:
```suggestion
```

cocotb/_build_libs.py
eric-wieser: Are we sure is still needs to be set now that we set the rpath on these libraries?

themperek: I can set. It should not harm to set it.

themperek: Should I do ?

eric-wieser: I'm suggesting the reverse. What happens if you _don't_ set it?

themperek: It looks it does not work https://travis-ci.org/github/themperek/cocotb/jobs/672471998 on osx.

eric-wieser: Thanks for trying

cocotb/_build_libs.py
```diff
@@ -137,16 +137,19 @@ def _gen_import_libs(self, def_dir):
...
 
 
-def _extra_link_args(lib_name):
+def _extra_link_args(lib_name=None, rpath=None):
     """
-    Add linker `install_name` argument on osx to load dependencies from the directory
-    where vpi/vhpi/fli library is located
+    Add linker argument to load dependencies from the directory where vpi/vhpi/fli library is located
+    On osx use`install_name`
+    On Linux use`rpath`
```
eric-wieser, line 144:
```suggestion
    Use `rpath` on all platforms
```

cocotb/_build_libs.py
```diff
@@ -137,16 +137,19 @@ def _gen_import_libs(self, def_dir):
                 )
 
 
-def _extra_link_args(lib_name):
+def _extra_link_args(lib_name=None, rpath=None):
     """
-    Add linker `install_name` argument on osx to load dependencies from the directory
-    where vpi/vhpi/fli library is located
+    Add linker argument to load dependencies from the directory where vpi/vhpi/fli library is located
+    On osx use`install_name`
```
eric-wieser, line 143:
```suggestion
    On osx use `install_name`.
```

cocotb/_build_libs.py
themperek: You are sure? On Windows there is rpath.

eric-wieser: Well, this matches the implementation, and what we were doing before.

themperek:
What is wrong with the message?

eric-wieser:
Nevermind, misread it

Issue #1603, opened 2020-04-07
Breathe needs pinning in CI


eric-wieser:
https://readthedocs.org/projects/cocotb/builds/10787038/

Since we have Sphinx pinned to <3.0.0, we need to pin breathe too:
```
 Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/cocotb/envs/1592/lib/python3.7/site-packages/sphinx/cmd/build.py", line 275, in build_main
    args.tags, args.verbosity, args.jobs, args.keep_going)
  File "/home/docs/checkouts/readthedocs.org/user_builds/cocotb/envs/1592/lib/python3.7/site-packages/sphinx/application.py", line 245, in __init__
    self.setup_extension(extension)
  File "/home/docs/checkouts/readthedocs.org/user_builds/cocotb/envs/1592/lib/python3.7/site-packages/sphinx/application.py", line 397, in setup_extension
    self.registry.load_extension(self, extname)
  File "/home/docs/checkouts/readthedocs.org/user_builds/cocotb/envs/1592/lib/python3.7/site-packages/sphinx/registry.py", line 465, in load_extension
    metadata = setup(app)
  File "/home/docs/checkouts/readthedocs.org/user_builds/cocotb/envs/1592/lib/python3.7/site-packages/breathe/__init__.py", line 13, in setup
    from . import directives
  File "/home/docs/checkouts/readthedocs.org/user_builds/cocotb/envs/1592/lib/python3.7/site-packages/breathe/directives.py", line 4, in <module>
    from .renderer import DoxygenToRstRendererFactory
  File "/home/docs/checkouts/readthedocs.org/user_builds/cocotb/envs/1592/lib/python3.7/site-packages/breathe/renderer/__init__.py", line 5, in <module>
    from . import sphinxrenderer
  File "/home/docs/checkouts/readthedocs.org/user_builds/cocotb/envs/1592/lib/python3.7/site-packages/breathe/renderer/sphinxrenderer.py", line 74, in <module>
    class DomainDirectiveFactory(object):
  File "/home/docs/checkouts/readthedocs.org/user_builds/cocotb/envs/1592/lib/python3.7/site-packages/breathe/renderer/sphinxrenderer.py", line 90, in DomainDirectiveFactory
    'define': (c.CMacroObject, 'macro'),
AttributeError: module 'sphinx.domains.c' has no attribute 'CMacroObject'

Exception occurred:
[...]

PR #1607, opened 2020-04-08
adding verilator waveform capability in documentation

Martoni:
Explain how to generate dump.vcd waveform with verilator simulator.


eric-wieser:
Rtd build failure is #1603

themperek:
It is not that verilator support $dumpfile and $dumpvars now?

Martoni:
> It is not that verilator support $dumpfile and $dumpvars now?
I don't know. I just tested it yesterday and to get a waveform in vcd format, I just have to add this extra args.

I changed syntax for dump.vcd file name. I think that was the cause of error.

themperek:
If ventilator support standard Verilog syntax to do this now I am not sure we should do anything n cocotb about this.

Martoni:
I tested $dumpvar with the code [given here](https://docs.cocotb.org/en/latest/simulator_support.html?highlight=dumpvar#waveforms). But it didn't work because of delay #1 that is not supported by verilator. If I remove delay (#1), it works !

The EXTRA_ARGS option is useful anyway because it's not necessary to modify verilog for dumping VCD traces.

No ?

ktbarrett:
@Martoni It is less portable though. Is there a difference in what gets logged between `--trace` and `$dumpvars`?

cmarqu:
This kind of change is actually a part of what we want to do in #1495

documentation/source/simulator_support.rst
```diff
@@ -94,6 +94,19 @@ To enable HDL code coverage, add Verilator's coverage option(s) to the :make:var
 
 This will result in coverage data being written to ``coverage.dat``.
 
+Waveforms
```
cmarqu, line 97:
```suggestion
.. _sim-verilator-waveforms:

Waveforms
```

documentation/source/simulator_support.rst
```diff
@@ -94,6 +94,19 @@ To enable HDL code coverage, add Verilator's coverage option(s) to the :make:var
...
+---------
+
+To get waveforms in VCD format, add Verilator's trace option(s) to the
+:make:var:`EXTRA_ARGS` make variable, for example:
+
+  .. code-block:: make
+
+    EXTRA_ARGS += --trace
+
+A VCD file named ``dump.vcd`` will be generated in current directory.
```
cmarqu, line 107:
```suggestion
A VCD file named ``dump.vcd`` will be generated in the current directory.
```

documentation/source/simulator_support.rst
```diff
@@ -94,6 +94,19 @@ To enable HDL code coverage, add Verilator's coverage option(s) to the :make:var
...
 
+Waveforms
+---------
+
+To get waveforms in VCD format, add Verilator's trace option(s) to the
+:make:var:`EXTRA_ARGS` make variable, for example:
+
+  .. code-block:: make
+
+    EXTRA_ARGS += --trace
```
cmarqu, line 105:
```suggestion
    EXTRA_ARGS += --trace --trace-structs
```

documentation/source/simulator_support.rst
```diff
@@ -94,6 +94,19 @@ To enable HDL code coverage, add Verilator's coverage option(s) to the :make:var
 
 This will result in coverage data being written to ``coverage.dat``.
 
+Waveforms
+---------
+
+To get waveforms in VCD format, add Verilator's trace option(s) to the
+:make:var:`EXTRA_ARGS` make variable, for example:
```
cmarqu, line 101:
```suggestion
:make:var:`EXTRA_ARGS` make variable, for example in a Makefile:
```

documentation/source/simulator_support.rst
```diff
@@ -94,6 +94,19 @@ To enable HDL code coverage, add Verilator's coverage option(s) to the :make:var
...
+
+  .. code-block:: make
+
+    EXTRA_ARGS += --trace
+
```
cmarqu, line 106:
```suggestion

To set the same options on the command line, use ``EXTRA_ARGS=--trace --trace-structs make ...``.
```
(that's a bit of #1590)

documentation/source/simulator_support.rst
Martoni: Why ""--trace-structs", to show that we can add several option in EXTRA_ARGS ?

eric-wieser: Yeah, and to make users aware of the typical arguments they care about for tracing.

Note also that on line 109 below you already list both arguments, so it would be consistent to do so here too.

cmarqu: Also, that's what ``VERILATOR_TRACE`` did (which we are replacing with this documentation section).

eric-wieser:
Needs a rebase to fix the doc builds.

Martoni:
> Hi @Martoni, could do the small changes I suggested?

Should be done now.

PR #1612, opened 2020-04-08
Deprecate VERILATOR_TRACE with doc reference

cmarqu:
The documentation itself is handled by #1607.
Refs #1495.

### DIFF
--- a/cocotb/share/makefiles/Makefile.deprecations
+++ b/cocotb/share/makefiles/Makefile.deprecations
@@ -3,5 +3,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 ifdef COCOTB_NVC_TRACE
-   $(warning COCOTB_NVC_TRACE is deprecated, see the "Simulator Support" section in the documentation.)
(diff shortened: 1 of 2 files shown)


cocotb/share/makefiles/Makefile.deprecations
```diff
@@ -5,3 +5,7 @@
 ifdef COCOTB_NVC_TRACE
    $(warning COCOTB_NVC_TRACE is deprecated, see the "Simulator Support" section in the documentation.)
 endif
+
+ifdef VERILATOR_TRACE
+   $(warning VERILATOR_TRACE is deprecated, see the "Simulator Support" section in the documentation.)
```
imphil, line 10: Do we also use three-space indents in other makefiles? Seems unusual.

eric-wieser: No, I think this one is unique in that regard.

cmarqu: Oops, fixed.

PR #1615, opened 2020-04-09
Add tests for getting and setting `BinaryValue.buff`

eric-wieser:

### DIFF
--- a/tests/pytest/test_binary_value.py
+++ b/tests/pytest/test_binary_value.py
@@ -256,7 +256,7 @@ def test_buff_big_endian():
     assert v.buff == orig_bytes
     assert v.binstr == orig_str
 
-    # extra bits are stripped
+    # extra bits are stripped because they don't fit into the 12 bits
     v.buff = b'\xF6\xC9'
     assert v.buff == orig_bytes
     assert v.binstr == orig_str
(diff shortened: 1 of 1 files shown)


tests/pytest/test_binary_value.py
```diff
@@ -242,3 +242,39 @@ def test_backwards_compatibility():
...
+
+    v = BinaryValue(value=orig_str, n_bits=12, bigEndian=True)
+    assert v.buff == orig_bytes
+
+    # should be unchanged
+    v.buff = orig_bytes
+    assert v.buff == orig_bytes
+    assert v.binstr == orig_str
+
+    # extra bits are stripped
```
cmarqu, line 259:
```suggestion
    # extra bits are stripped because they don't fit into the 12 bits
```

tests/pytest/test_binary_value.py
```diff
@@ -242,3 +242,39 @@ def test_backwards_compatibility():
...
+
+    v = BinaryValue(value=orig_str, n_bits=12, bigEndian=False)
+    assert v.buff == orig_bytes
+
+    # should be unchanged
+    v.buff = orig_bytes
+    assert v.buff == orig_bytes
+    assert v.binstr == orig_str
+
+    # extra bits are stripped
```
cmarqu, line 277:
```suggestion
# extra bits are stripped because they don't fit into the 12 bits
```

Issue #1585, opened 2020-04-06
Remove support for cross-compilation in the endian example


eric-wieser:
We have some weird code here that seemingly tries to support building 32-bit extension modules with 64-bit python, if the user overrides the `ARCH` variable:

https://github.com/cocotb/cocotb/blob/688d391376eda47446d8f949d9e62f8ca93d51b0/cocotb/share/makefiles/Makefile.pylib#L34-L35

https://github.com/cocotb/cocotb/blob/688d391376eda47446d8f949d9e62f8ca93d51b0/cocotb/share/makefiles/Makefile.pylib.Linux#L34-L38

https://github.com/cocotb/cocotb/blob/688d391376eda47446d8f949d9e62f8ca93d51b0/cocotb/share/makefiles/Makefile.pylib.Darwin#L34-L38

I don't know why we'd ever want to support this. If the user needs a 32-bit extension module, then they should be using `cocotb-config` installed inside a 32-bit python. Or better yet, they should be looking at the python docs to build it,  not our docs.

If we removed it, we could merge all the `Makefile.pylib` files into a single file.

It seems that maybe this is a remnant of when this override was in place:

https://github.com/cocotb/cocotb/blob/688d391376eda47446d8f949d9e62f8ca93d51b0/cocotb/share/makefiles/simulators/Makefile.modelsim#L31

themperek:
I am for it. In any case, it may not be so easy and not sure that those lines will help (rather more confusion). 

It can be moved to the endian example worst case.

eric-wieser:
> It can be moved to the endian example worst case.

Or perhaps also in the best case. After removing these workaround, there are only two interesting lines left in the makefile:

https://github.com/cocotb/cocotb/blob/688d391376eda47446d8f949d9e62f8ca93d51b0/cocotb/share/makefiles/Makefile.pylib.Linux#L32

https://github.com/cocotb/cocotb/blob/688d391376eda47446d8f949d9e62f8ca93d51b0/cocotb/share/makefiles/Makefile.pylib.Linux#L40

These could easily be moved to within the endian example makefile

eric-wieser:
After gh-1619, this is already not supported - we may as well delete the makefiles now.

PR #1619, opened 2020-04-09
Remove dependency on Makfiles in endian_swapper

themperek:
Gives the ability to simplify Makefiles (remove pylib Makefiles).

Remove dependency on cocotb Makefiles in endian_swapper (test_endian_swapper_hal) example.
Cleanup.

### DIFF
--- a/examples/endian_swapper/cosim/Makefile
+++ b/examples/endian_swapper/cosim/Makefile
@@ -1,30 +1,34 @@
-include $(shell cocotb-config --makefiles)/Makefile.inc
-include $(shell cocotb-config --makefiles)/Makefile.pylib
+PYTHON_BIN ?= $(realpath $(shell cocotb-config --python-bin))
 
-PYTHON_LIBDIR ?= /usr/lib64
-SWIG ?= swig
-CC=gcc
-ifeq ($(ARCH),i686)
-	CC += -m32
+PYTHON_LIBDIR := $(shell $(PYTHON_BIN) -c 'from distutils import sysconfig; print( sysconfig.get_config_var("LIBDIR") )')
+ifeq ($(PYTHON_LIBDIR),None)
+    PYTHON_LIBDIR := $(shell dirname $(PYTHON_BIN))/libs
 endif
 
+PYTHON_INCLUDEDIR := $(shell $(PYTHON_BIN) -c 'import distutils.sysconfig; print( distutils.sysconfig.get_python_inc() )')
+
+SWIG ?= swig
+CC ?= gcc
+
+OS := $(shell uname)
+
 ifeq ($(OS),Darwin)
-PYTHON_LD_FLAGS += -undefined dynamic_lookup
+    CFLAGS += -undefined dynamic_lookup
 endif
 
 .PHONY: all
 all: io_module.so _hal.so
 
 io_module.o: io.c io_module.h io.h
-	$(CC) $(GCC_FLAGS) -fPIC -pthread -fno-strict-aliasing -O2 -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 \
+	$(CC) $(CFLAGS) -fPIC -pthread -fno-strict-aliasing -O2 -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 \
 	-fexceptions --param=ssp-buffer-size=4 -D_GNU_SOURCE \
 	-fwrapv -DNDEBUG -I$(PYTHON_INCLUDEDIR) -c $< -o $@
 
 io_module.so: io_module.o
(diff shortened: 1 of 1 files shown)


examples/endian_swapper/cosim/Makefile
```diff
@@ -1,30 +1,32 @@
...
 
-PYTHON_LIBDIR ?= /usr/lib64
-SWIG ?= swig
-CC=gcc
-ifeq ($(ARCH),i686)
-	CC += -m32
+PYTHON_LIBDIR := $(shell $(PYTHON_BIN) -c 'from distutils import sysconfig; print( sysconfig.get_config_var("LIBDIR") )')
+ifeq ($(PYTHON_LIBDIR),None)
+    PYTHON_LIBDIR := $(shell dirname $(PYTHON_BIN))/libs
 endif
```
imphil, line 6: I think we have similar code in at least two places now. Can we add a --python-libdir and --python-includedir option to cocotb-config instead?

examples/endian_swapper/cosim/Makefile
```diff
@@ -1,30 +1,32 @@
...
 
 .PHONY: all
 all: io_module.so _hal.so
 
 io_module.o: io.c io_module.h io.h
-	$(CC) $(GCC_FLAGS) -fPIC -pthread -fno-strict-aliasing -O2 -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 \
+	$(CC) $(CFLAGS) -fPIC -pthread -fno-strict-aliasing -O2 -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 \
 	-fexceptions --param=ssp-buffer-size=4 -D_GNU_SOURCE \
 	-fwrapv -DNDEBUG -I$(PYTHON_INCLUDEDIR) -c $< -o $@
```
imphil, line 23: While you're at it: can you remove the NDEBUG (ideally in a separate commit)? It seems a dangerous choice for a piece of demo software.

examples/endian_swapper/cosim/Makefile
```diff
@@ -1,30 +1,32 @@
...
 
 io_module.so: io_module.o
-	$(CC) -pthread -shared  $< -L$(PYTHON_LIBDIR) $(PYTHON_LD_FLAGS) -o $@
+	$(CC) $(CFLAGS) -pthread -shared  $< -L$(PYTHON_LIBDIR)  -o $@
 
 _hal.so: ../hal/endian_swapper_hal.c endian_swapper_hal_wrap.c io_module.so
-	$(CC) -g -ldl -shared -fPIC -I$(shell pwd) $(PYTHON_LD_FLAGS) -I$(PYTHON_INCLUDEDIR) -I../hal ../hal/endian_swapper_hal.c endian_swapper_hal_wrap.c io_module.so -o $@
+	$(CC) $(CFLAGS) -g -ldl -shared -fPIC -I$(shell pwd) -I$(PYTHON_INCLUDEDIR) -I../hal ../hal/endian_swapper_hal.c endian_swapper_hal_wrap.c io_module.so -o $@
 	if [ $(OS) = Darwin ]; then \
```
imphil, line 30: That needs replacing as well, $(OS) isn't available any more.

examples/endian_swapper/cosim/Makefile
themperek: Only here in the makefiles (pylib will be removed). We will have only (some of it) in `_build_libs.py` as I can see.

themperek: But this is can be done. Possible useful and much easier to control from Python any fixes.

eric-wieser: I don't think there's much value in doing this. A better use of time would be for this example to show how to use setuptools to build an extension, and have it invoke `$(PYTHON_BIN) setup.py build_ext` or similar

themperek: Long run I would like to do what @eric-wieser suggests.

imphil: Don't worry if it's used only once.

examples/endian_swapper/cosim/Makefile
themperek: OK. Can do separate.

themperek: https://github.com/cocotb/cocotb/pull/1621

eric-wieser: Again, this might be easier to just let setuptools handle

eric-wieser:
Note that gh-1585 has implicitly been resolved by doing this, as the code it refers to is now unused.

PR #1620, opened 2020-04-09
Add dff_simple example

themperek:
A simple example that works correctly for most simulators.

### DIFF
--- a/examples/simple_dff/Makefile
+++ b/examples/simple_dff/Makefile
@@ -1,0 +1,15 @@
+# This file is public domain, it can be freely copied without restrictions.
+# SPDX-License-Identifier: CC0-1.0
+
+TOPLEVEL_LANG ?= verilog
+
+ifeq ($(TOPLEVEL_LANG),verilog)
+  VERILOG_SOURCES = $(shell pwd)/dff.sv
+else ifeq ($(TOPLEVEL_LANG),vhdl)
+  VHDL_SOURCES = $(shell pwd)/dff.vhdl
+endif
+
+MODULE = test_dff
+TOPLEVEL = dff
+
+include $(shell cocotb-config --makefiles)/Makefile.sim
--- a/examples/simple_dff/dff.sv
+++ b/examples/simple_dff/dff.sv
@@ -1,0 +1,15 @@
+// This file is public domain, it can be freely copied without restrictions.
+// SPDX-License-Identifier: CC0-1.0
+
+`timescale 1us/1us
+
+module dff (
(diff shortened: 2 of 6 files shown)


cmarqu:
Should these new files get the SPDX header?

themperek:
I think we do not do this fo examples? @imphil

themperek:
@elgorwi Thank you. I have not seen VHDL for a very very long time.

examples/simple_dff/dff.sv
```diff
@@ -0,0 +1,13 @@
+
+`timescale 1us/1us
+
+module dff (
+    input logic clk, d,
+    output logic q
```
imphil, line 6: Let's go for a 2-space indent in Verilog to align with the VHDL example.

examples/simple_dff/dff.vhdl
```diff
@@ -0,0 +1,19 @@
...
+port(
+  clk: in std_logic;
+  d: in std_logic;
+  q: out std_logic);
+end dff;
+
+architecture behavioral of dff is
+begin
+  process(clk)
+  begin
```
imphil, line 14: In Verilog we would format this `process (clk) begin` (i.e. begin not in a new line, and a space before parentheses).

imphil:
Can you add a CC0 (public domain) header to this example (and going forward, to all examples).

```python3
# This file is public domain, it can be freely copied without restrictions.
# SPDX-License-Identifier: CC0-1.0
```

This allows everybody to do whatever they want with those examples.

examples/simple_dff/Makefile
```diff
@@ -0,0 +1,16 @@
...
+  VERILOG_SOURCES = $(shell pwd)/dff.sv
+else ifeq ($(TOPLEVEL_LANG),vhdl)
+  VHDL_SOURCES = $(shell pwd)/dff.vhdl
+endif
+
+MODULE = test_dff
+TOPLEVEL = dff
+
+COCOTB_HDL_TIMEPRECISION = 1us
+COCOTB_HDL_TIMEUNIT = 1us
```
imphil, line 14: Was that needed for something other than Verilator? So we can remove that now for simplicity?

themperek: Yes. Verialotr supports `timescale` now so can be removed.

examples/simple_dff/Makefile
```diff
@@ -0,0 +1,16 @@
+
```
imphil, line 1: You can remove this empty line.

PR #1622, opened 2020-04-10
Update README example so verilator runs fast

themperek:
Add `COCOTB_HDL_TIMEPRECISION` and `COCOTB_HDL_TIMEPRECISION` in the Makefile and set to 1us.


themperek:
Not really needed anymore https://github.com/verilator/verilator/pull/2253

PR #1300, opened 2020-01-15
Initialize the PWD variable only in sim makefile

sfavazza-duagon:
Explicitly initialize the PWD variable in all simulation Makefiles.

Fixes #1299

PR #1623, opened 2020-04-10
Remove unused and undocumented Makefile variables.

themperek:


cocotb/share/makefiles/Makefile.sim
```diff
@@ -130,8 +130,6 @@ CUSTOM_SIM_DEPS += $(shell find $(COCOTB_PY_DIR)/cocotb/ -name "*.py")
 # This triggers a recompilation of the simulation if cocotb library is updated.
 CUSTOM_SIM_DEPS += $(shell find $(LIB_DIR)/ -name "*")
 
-PWD := $(shell pwd)
```
eric-wieser, line 133: Are we sure all the other makefiles define this? It was added recently to fix a big where we forgot to define it I think

eric-wieser: #1300

themperek: It was used to set `PYTHONPATH` now is here: https://github.com/cocotb/cocotb/blob/master/cocotb/__init__.py#L63

eric-wieser: So to confirm, there are no makefiles left that use but do not define this variable?

themperek: Yes. To what I can see. CI should fail also.

eric-wieser: Here's one that seems undefined: https://github.com/cocotb/cocotb/blob/master/tests/designs/avalon_module/Makefile#L42

themperek: It is in CI so how does this work?

eric-wieser: I think that's the only one

themperek: HA: https://unix.stackexchange.com/questions/174990/what-is-pwd-vs-current-working-directory

themperek: And we do `SHELL := bash` also here: https://github.com/cocotb/cocotb/blob/master/cocotb/share/makefiles/Makefile.sim#L34

themperek: Should I remove all this `$(shell pwd)` and just use `$(PWD)` ? Separate PR?

eric-wieser: I don't think we should do that, apparently the variable is empty within sudo

eric-wieser: Let's worry about that later, anyway

eric-wieser: But I think we should add the one missing call to that file, for consistency

themperek: BTW: https://pubs.opengroup.org/onlinepubs/009604599/utilities/xcu_chap02.html 
Maybe should even avoid `PWD:=$(shell pwd)`. Yes. For another time I guess.

eric-wieser: `$(CURDIR)` is possibly a better choice, for a follow-up PR.

themperek: >But I think we should add the one missing call to that file, for consistency

Should be done.

tests/designs/avalon_module/Makefile
```diff
@@ -39,7 +39,7 @@ else
 
 TOPLEVEL := burst_read_master
 
-COCOTB?=$(PWD)/../../..
+COCOTB?=$(shell pwd)/../../..
```
eric-wieser, line 42: Would prefer we stay consistent with all the other files

```suggestion
PWD=$(shell pwd)

COCOTB?=$(PWD)/../../..
```

That way we can do further cleanup with a find and replace

PR #1629, opened 2020-04-10
Rearrange Makefiles

themperek:
### Changes:
- Move all "logic" to `Makefile.inc`
- `Makefile.sim` serves only to include appropriate simulator Makefile depending on `SIM` environmental variable
- Include `Makefile.inc` in every simulator Makefile

### Consequences:
- Users need to include **ONLY** : `include $(shell cocotb-config --makefiles)/Makefile.sim`
- Users can create/copy any of simulator Makefiles, modify as needed and include instead of `Makefile.sim`. This way can easily use custom/advanced configuration/options and **cocotb does not need to support them**.
- Backward compatibility is kept.

Update of examples, tests and documentation will follow.

### DIFF
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -29,6 +29,19 @@
 
 # Common makefile included by everything
 
+ifndef COCOTB_MAKEFILE_INC_INCLUDED # Protect against multiple includes
+COCOTB_MAKEFILE_INC_INCLUDED = 1
+
+# Default sim rule will force a re-run of the simulation (though the cocotb library
+# and RTL compilation phases are still evaluated by makefile dependencies)
+.PHONY: sim
+sim:
+	-@rm -f $(COCOTB_RESULTS_FILE)
+	$(MAKE) $(COCOTB_RESULTS_FILE)
+
+# Make sure to use bash for the pipefail option used in many simulator Makefiles
+SHELL := bash
+
 # Directory containing the cocotb Python module (realpath for Windows compatibility)
 COCOTB_PY_DIR := $(realpath $(shell cocotb-config --prefix))
 
@@ -38,9 +51,9 @@ COCOTB_SHARE_DIR := $(COCOTB_PY_DIR)/cocotb/share
 
 OS=$(shell uname)
 ifneq (, $(findstring MINGW, $(OS)))
-OS=Msys
+    OS := Msys
 else ifneq (, $(findstring MSYS, $(OS)))
-OS=Msys
+    OS := Msys
 endif
 export OS
 
@@ -50,27 +63,12 @@ PYTHON_BIN ?= $(realpath $(shell cocotb-config --python-bin))
 
 include $(COCOTB_SHARE_DIR)/makefiles/Makefile.deprecations
 
-# Default to Icarus if no simulator is defined
-SIM ?= icarus
-
-SIM_DEFINE := $(shell echo $(SIM) | tr a-z A-Z)
-
-# Use a common define for Questa and Modelsim and cvc
-ifeq ($(SIM_DEFINE),$(filter $(SIM_DEFINE),QUESTA CVC))
-SIM_DEFINE = MODELSIM
-endif
-
-# Use a common define for Xcelium and IUS
-ifeq ($(SIM_DEFINE),XCELIUM)
-SIM_DEFINE = IUS
-endif
-
 LIB_DIR=$(COCOTB_PY_DIR)/cocotb/libs
 
 ifeq ($(OS),Msys)
-LIB_EXT    := dll
+    LIB_EXT := dll
 else
-LIB_EXT    := so
+    LIB_EXT := so
 endif
 
 PYTHON_ARCH := $(shell $(PYTHON_BIN) -c 'from platform import architecture; print(architecture()[0])')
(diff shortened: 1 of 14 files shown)


cocotb/share/makefiles/Makefile.inc
```diff
@@ -29,6 +29,15 @@
 
 # Common makefile included by everything
 
+ifndef COCOTB_MAKEFILE_IN_INCLUDED # Protect against multiple includes
+COCOTB_MAKEFILE_IN_INCLUDED = 1
```
eric-wieser, line 33:
```suggestion
ifndef COCOTB_MAKEFILE_INC_INCLUDED # Protect against multiple includes
COCOTB_MAKEFILE_INC_INCLUDED = 1
```

cocotb/share/makefiles/Makefile.inc
```diff
@@ -29,6 +29,15 @@
...
 # Common makefile included by everything
 
+ifndef COCOTB_MAKEFILE_INC_INCLUDED # Protect against multiple includes
+COCOTB_MAKEFILE_INC_INCLUDED = 1
+
+# Make sure to use bash for the pipefail option used in many simulator Makefiles
+SHELL := bash
+
+.PHONY: all
+all: sim
```
eric-wieser, line 39: I worry this is going to irritate users who want to define their own `all` target

cocotb/share/makefiles/Makefile.inc
```diff
@@ -91,3 +85,38 @@ ifeq ($(OS),Msys)
...
+regression: $(COCOTB_RESULTS_FILE)
+
+# Default sim rule will force a re-run of the simulation (though the cocotb library
+# and RTL compilation phases are still evaluated by makefile dependencies)
+.PHONY: sim
+sim:
+	-@rm -f $(COCOTB_RESULTS_FILE)
+	$(MAKE) $(COCOTB_RESULTS_FILE)
+
+endif
```
eric-wieser, line 122:
```suggestion
else
    $(warning "Including both Makefile.inc and Makefile.sim is deprecated, as the latter now includes the former")
endif
```

cocotb/share/makefiles/Makefile.inc
themperek: Can move those 2 to simulator Makefiles or Makefile.sim (than need to be added by user)

eric-wieser: I think you should:

* move the `sim:` target to the top, so that `make` by itself runs the simulation
* Move the `all` target to `Makefile.sim`

themperek: Hope it is right now. Seems that everything still works.

cocotb/share/makefiles/Makefile.sim
```diff
@@ -27,11 +27,7 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
-# This is a simple wrapper Makefile to pull in the appropriate Makefile for
-# the desired simulator and set up common paths for the sims to use to build with
-
-# Make sure to use bash for the pipefail option used in many simulator Makefiles
-SHELL := bash
+# This file  includes appropriate makefile depending on SIM variable
```
cmarqu, line 30:
```suggestion
# This file includes the appropriate makefile depending on the SIM variable
```

cocotb/share/makefiles/Makefile.sim
```diff
@@ -94,59 +90,44 @@ endef
...
-endif
-$(info $(helpmsg))
-# is there a cleaner way to exit here?
-$(error "Stopping after printing help")
+    _VERSION := $(shell cocotb-config -v)
+    # for non-intuitive use of ifneq see https://www.gnu.org/software/make/manual/make.html#Testing-Flags
+    ifneq (,$(findstring dev,$(_VERSION)))
+        DOCLINK := https://cocotb.readthedocs.io/en/latest/building.html
+    else
+        DOCLINK := https://cocotb.readthedocs.io/en/v$(_VERSION)/building.html
```
cmarqu, line 98:
```suggestion
        DOCLINK := https://docs.cocotb.org/en/latest/building.html
    else
        DOCLINK := https://docs.cocotb.org/en/v$(_VERSION)/building.html
```

documentation/source/newsfragments/1629.feature.rst
```diff
@@ -0,0 +1,2 @@
+Only ``Makefile.sim`` is required to be included in user Makefile (``Makefile.inc`` is not needed).
+Add possibility to create custom simulator control by copying/creating simulator Makefiles and including instead of ``Makefile.sim``.
```
imphil, line 2: I'm not sure if we really want to advertise/support this configuration, it could make it harder for us to change the build system in the future. I'd tend towards leaving this as a "hidden feature".

documentation/source/newsfragments/1629.feature.rst
```diff
@@ -0,0 +1,2 @@
+Only ``Makefile.sim`` is required to be included in user Makefile (``Makefile.inc`` is not needed).
```
imphil, line 1:
```suggestion
Including ``Makefile.inc`` from user makefiles is now a no-op and deprecated. Lines like  ``include $(shell cocotb-config --makefiles)/Makefile.inc`` can be removed from user makefiles without loss in functionality.
```

Can you put this into the removals category of news fragments, not in the features one?

cocotb/share/makefiles/Makefile.sim
imphil, line 30:
```suggestion
# This file includes appropriate an makefile depending on the SIM variable.
```

cocotb/share/makefiles/Makefile.inc
```diff
@@ -91,3 +88,33 @@ ifeq ($(OS),Msys)
...
+$(SIM_BUILD):
+	mkdir -p $@
+
+# Regression rule uses Make dependencies to determine whether to run the simulation
+.PHONY: regression
+regression: $(COCOTB_RESULTS_FILE)
+
+else
+    $(warning "Including both Makefile.inc and Makefile.sim is deprecated, as the latter now includes the former")
+endif
```
imphil, line 120:
```suggestion
endif # COCOTB_MAKEFILE_INC_INCLUDED
```

cocotb/share/makefiles/Makefile.inc
```diff
@@ -91,3 +88,33 @@ ifeq ($(OS),Msys)
...
+
+$(SIM_BUILD):
+	mkdir -p $@
+
+# Regression rule uses Make dependencies to determine whether to run the simulation
+.PHONY: regression
+regression: $(COCOTB_RESULTS_FILE)
+
+else
+    $(warning "Including both Makefile.inc and Makefile.sim is deprecated, as the latter now includes the former")
```
imphil, line 119:
```suggestion
    $(warning "Including Makefile.inc from a user makefile is a no-op and deprecated. Remove the Makefile.inc inclusion from your makefile, and only leave the Makefile.sim include.")
```

I'd be happier if we can tighten the condition under which this deprecation message is shown to exactly the case where Makefile.sim and Makefile.inc are both included (e.g. by using a special variable COCOTB_MAKEFILE_SIM_INCLUDED or so), and not use the fact that Makefile.inc is included twice as a proxy.

documentation/source/newsfragments/1629.feature.rst
themperek: Just mind that the only Makefile left in practice is `Makefile.inc` and with some luck will not change much.

imphil: We are having so much work with slowly deprecating things currently around this subsystem, adding more API-like surface to it seems to me like a risk without much benefit.

cocotb/share/makefiles/Makefile.inc
themperek: I think to achieve this we can just move this warning to `Makefile.sim` (check for `COCOTB_MAKEFILE_INC_INCLUDED`). Should I do?   Should this be maybe `info`?

PR #1630, opened 2020-04-11
Improve timeout docs and link with_timeout in test decorator

### LINKED ISSUE #1617: Getting rid of None in an option of TestFactory
Hello everyone,

I did not find this in the documentation, so maybe you can push me in the right direction:
I want to use TestFactory to supply a list of discrete values to create different tests. To achieve this, I create a function annotated with @cocotb.test **run_test(dut, my_option=None)** which has the general structure of my test, depending on the value of my_option.
After defining this function, I do the following:
```
factory = TestFactory(run_test)
factory.add_option("my_option", list_of_values)
factory.generate_tests()
```
And it almost works as expected. It creates a test for each of my entries. However, the first test that gets created provides the value None to the parameter my_option. Therefore, I have one more test than intended which obviously fails, as I cannot do anything sensible with None in my test. I assume that this is the result of cocotb running all functions annotated with @cocotb.test if I don't supply the TESTCASE environment variable. Using cocotb.coroutine as an annotation resolves my problem, but I highly appreciate the timeout functionality of the cocotb.test annotation. Is there a clean way how to ensure that the default implementation of my function is not used as a test, without the need to hardcode the test names in TESTCASE?

Cheers and thanks in advance!

ktbarrett:
From https://github.com/cocotb/cocotb/issues/1617#issuecomment-612061921. Closes #1617.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -413,12 +413,17 @@ class test(coroutine, metaclass=_decorator_helper):
     Used as ``@cocotb.test(...)``.
 
     Args:
-        timeout_time (int, optional):
-            Value representing simulation timeout.
+        timeout_time (numbers.Real or decimal.Decimal, optional):
+            Simulation time duration before timeout occurs.
 
             .. versionadded:: 1.3
-        timeout_unit (str, optional):
-            Unit of timeout value, see :class:`~cocotb.triggers.Timer` for more info.
+
+            .. note::
+                Test timeout is intended for protection against deadlock.
+                Users should use :class:`~cocotb.triggers.with_timeout` if they require a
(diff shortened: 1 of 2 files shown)


cocotb/decorators.py
```diff
@@ -413,12 +413,17 @@ class test(coroutine, metaclass=_decorator_helper):
     Used as ``@cocotb.test(...)``.
 
     Args:
-        timeout_time (int, optional):
-            Value representing simulation timeout.
+        timeout_time (numbers.Real or decimal.Decimal):
```
cmarqu, line 416:
```suggestion
        timeout_time (numbers.Real or decimal.Decimal, optional):
```
?

PR #1631, opened 2020-04-11
Fix trigger documentation error

ktbarrett:
https://docs.cocotb.org/en/latest/triggers.html#triggers first example is busted.


documentation/source/triggers.rst
```diff
@@ -5,7 +5,7 @@ Triggers
 Triggers are used to indicate when the cocotb scheduler should resume coroutine execution.
 To use a trigger, a coroutine should :keyword:`await` it.
 This will cause execution of the current coroutine to pause.
-When the trigger fires, execution of the paused coroutine will resume::
+When the trigger fires, execution of the paused coroutine will resume
```
eric-wieser, line 8:
```suggestion
When the trigger fires, execution of the paused coroutine will resume:
```

eric-wieser:
Thanks for finding that

PR #1632, opened 2020-04-11
Remove unused Makefile.pylib files.

### LINKED ISSUE #1585: Remove support for cross-compilation in the endian example

themperek:
After #1306 and #1619 those files are not used anymore. 

Resolves #1585


documentation/source/newsfragments/1632.change.rst
```diff
@@ -0,0 +1,2 @@
+``Makefile.plib`` has been removed.
```
eric-wieser, line 1:
```suggestion
``Makefile.pylib`` has been removed. 
```

documentation/source/newsfragments/1632.change.rst
```diff
@@ -0,0 +1,2 @@
+``Makefile.pylib`` has been removed. 
+In case used ``PYTHON_LIBDIR`` and ``PYTHON_INCLUDEDIR`` have to be set by the user (see endian_swapper example).
```
themperek, line 2:
```suggestion
In case used ``PYTHON_LIBDIR`` or ``PYTHON_INCLUDEDIR`` have to be set by the user (see endian_swapper example).
```

documentation/source/newsfragments/1632.change.rst
```diff
@@ -0,0 +1,2 @@
+``Makefile.pylib`` has been removed. 
+In case used ``PYTHON_LIBDIR`` or ``PYTHON_INCLUDEDIR`` have to be set by the user (see endian_swapper example).
```
cmarqu, line 2:
```suggestion
In case ``PYTHON_LIBDIR`` or ``PYTHON_INCLUDEDIR`` have been used, they now have to be set by the user (see the ``endian_swapper`` example).
```
(if that's what you wanted to say?)

documentation/source/newsfragments/1632.removal.rst
```diff
@@ -0,0 +1,2 @@
+``Makefile.pylib`` has been removed.
+In case ``PYTHON_LIBDIR`` or ``PYTHON_INCLUDEDIR`` have been used, they now have to be set by the user (see the ``endian_swapper`` example).
```
eric-wieser, line 2:
```suggestion
``Makefile.pylib``, which provided helpers for building C extension modules for Python, has been removed.
Users of the ``PYTHON_LIBDIR`` and ``PYTHON_INCLUDEDIR`` variables will now have to compute these values themselves.
See the ``endian_swapper`` example for how to do this.
```

eric-wieser:
Looks good to me, assuming we're fine with breaking the (hopefully very small number of) users who relied on these variables. I'll let @imphil make the final call on that, but I think it's ok given the release note.

Issue #1315, opened 2020-01-22
Signals name starting with underscore


JaewonHur:
When module has variable whose name starts with underscore, such as
```verilog
  reg _myname;
```
cocotb can not access it. 

It seems 'handle.py' : 277
```python
  if name.startswith("_"):
      return SimHandleBase.__getattr__(self, name)
```
need modification.

eric-wieser:
I think this is by design - underscore names are reserved for cocotb internals.

Perhaps we should (or do) expose these as `dut.signals['_myname']` or `dut._.myname`

JaewonHur:
I see.

Thank you

eric-wieser:
Reopening, we should at least document this

cmarqu:
It works when you do it like this:
```
    # need to fill in _sub_handles
    _ = dir(dut)
    dut._sub_handles["_underscore_name"] <= 1
    yield Timer(1, 'ns')
    assert dut._sub_handles["_underscore_name"].value == 1
```

cmarqu:
Tried with Questa and Xcelium where it works, but unfortunately, Icarus doesn't see the signal.

eric-wieser:
Sounds related to #1610, where I found a similar problem with icarus

PR #1610, opened 2020-04-08
Fix handling of escaped identifiers containing dots

eric-wieser:
Needs tests.

Fixes #1548, probably.

PR #1634, opened 2020-04-11
Add test for accessing _underscore_name

cmarqu:
Refs #1315

### DIFF
--- a/tests/designs/sample_module/sample_module.sv
+++ b/tests/designs/sample_module/sample_module.sv
@@ -135,4 +135,6 @@ assign logic_a = stream_in_valid;
 always@* logic_b = stream_in_valid;
 always@(posedge clk) logic_c <= stream_in_valid;
 
+reg _underscore_name;
+
 endmodule
--- a/tests/designs/sample_module/sample_module.sv
+++ b/tests/designs/sample_module/sample_module.sv
@@ -136,5 +136,11 @@ always@* logic_b = stream_in_valid;
 always@(posedge clk) logic_c <= stream_in_valid;
 
 reg _underscore_name;
+`ifdef __ICARUS__
+    // By default, a variable must be used in some way in order
+    // to be visible to VPI in Icarus Verilog.
+    // See https://github.com/steveicarus/iverilog/issues/322
+    assign _underscore_name = 0;
+`endif
 
 endmodule
--- a/tests/test_cases/test_cocotb/test_handle.py
+++ b/tests/test_cases/test_cocotb/test_handle.py
@@ -131,7 +131,7 @@ async def test_access_underscore_name(dut):
     """Test accessing HDL name starting with an underscore"""
     # direct access does not work because we consider such names cocotb-internal
     with assert_raises(AttributeError):
-        dut._underscore_name.value
+        dut._underscore_name
 
     # indirect access works
     dut._id("_underscore_name", extended=False) <= 0
--- a/tests/test_cases/test_cocotb/test_handle.py
+++ b/tests/test_cases/test_cocotb/test_handle.py
@@ -123,3 +123,23 @@ def test_real_assign_int(dut):
     log.info("Read back value %d" % got)
     if got != float(val):
         raise TestFailure("Values didn't match!")
+
+
+# identifiers starting with `_` are illegal in VHDL
+@cocotb.test(skip=cocotb.LANGUAGE in ["vhdl"])
+async def test_access_underscore_name(dut):
+    """Test accessing HDL name starting with an underscore"""
+    # direct access does not work because we consider such names cocotb-internal
+    with assert_raises(AttributeError):
+        dut._underscore_name.value
+
+    # indirect access works
+    dut._id("_underscore_name", extended=False) <= 0
+    await Timer(1, 'ns')
+    assert dut._id("_underscore_name", extended=False).value == 0
+    dut._id("_underscore_name", extended=False) <= 1
+    await Timer(1, 'ns')
+    assert dut._id("_underscore_name", extended=False).value == 1
+    dut._id("_underscore_name", extended=False) <= 0
+    await Timer(1, 'ns')
+    assert dut._id("_underscore_name", extended=False).value == 0
--- a/tests/test_cases/test_discovery/test_discovery.py
+++ b/tests/test_cases/test_discovery/test_discovery.py
@@ -33,6 +33,7 @@ from cocotb.triggers import Timer
 from cocotb.result import TestError, TestFailure
 from cocotb.handle import IntegerObject, ConstantObject, HierarchyObject, StringObject
 
+
 @cocotb.test()
 def recursive_discover(dut):
     """Discover absolutely everything in the DUT"""
--- a/tests/test_cases/test_discovery/test_discovery.py
+++ b/tests/test_cases/test_discovery/test_discovery.py
@@ -33,7 +33,6 @@ from cocotb.triggers import Timer
 from cocotb.result import TestError, TestFailure
 from cocotb.handle import IntegerObject, ConstantObject, HierarchyObject, StringObject
 
-
 @cocotb.test()
 def recursive_discover(dut):
     """Discover absolutely everything in the DUT"""
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -275,6 +275,8 @@ class HierarchyObject(RegionObject):
         """Query the simulator for a object with the specified name
         and cache the result to build a tree of objects.
         """
+        if name.startswith("_"):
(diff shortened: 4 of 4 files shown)


tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -471,3 +471,19 @@ def type_check_verilog(dut):
...
         if handle[0]._type != handle[1]:
             raise TestFailure("Expected %s found %s for %s" % (handle[1], handle[0]._type, handle[0]._fullname))
+
+
+@cocotb.test()
+def test_access_underscore_name(dut):
+    """Test accessing HDL name starting with an underscore"""
+    # need to fill in _sub_handles
+    _ = dir(dut)
+    dut._sub_handles["_underscore_name"] <= 0
```
eric-wieser, line 481: No need for dir with:
```suggestion
    dut._id("_underscore_name", extended=False) <= 0
```

cmarqu: Done, thanks. Note to self - add documentation for these members.

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -471,3 +471,17 @@ def type_check_verilog(dut):
         tlog.info("Handle %s" %  (handle[0]._fullname,))
         if handle[0]._type != handle[1]:
             raise TestFailure("Expected %s found %s for %s" % (handle[1], handle[0]._type, handle[0]._fullname))
+
+
+@cocotb.test()
```
eric-wieser, line 476: This should be marked verilog-only - underscores are illegal at the start of VHDL identifiers

cmarqu: True, I had forgotten to add that.
But since it's valid using extended identifiers, I maybe should wait until your #1610 is in and add this test near it?

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -471,3 +471,17 @@ def type_check_verilog(dut):
         tlog.info("Handle %s" %  (handle[0]._fullname,))
         if handle[0]._type != handle[1]:
             raise TestFailure("Expected %s found %s for %s" % (handle[1], handle[0]._type, handle[0]._fullname))
+
+
+@cocotb.test(skip=cocotb.LANGUAGE in ["vhdl"])
```
eric-wieser, line 476:
```suggestion
# identifiers starting with `_` are illegal in vhdl
@cocotb.test(skip=cocotb.LANGUAGE in ["vhdl"])
```

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -471,3 +471,17 @@ def type_check_verilog(dut):
...
+    """Test accessing HDL name starting with an underscore"""
+    dut._id("_underscore_name", extended=False) <= 0
+    yield Timer(1, 'ns')
+    assert dut._id("_underscore_name", extended=False).value == 0
+    dut._id("_underscore_name", extended=False) <= 1
+    yield Timer(1, 'ns')
+    assert dut._id("_underscore_name", extended=False).value == 1
+    dut._id("_underscore_name", extended=False) <= 0
+    yield Timer(1, 'ns')
+    assert dut._id("_underscore_name", extended=False).value == 0
```
eric-wieser, line 487: What does `dut._underscore_name` give here? I'd argue it should either fail or emit a deprecation warning

tests/test_cases/test_discovery/test_discovery.py
cmarqu: Done.

tests/test_cases/test_discovery/test_discovery.py
cmarqu: For some reason, it just works for me now, but I'm sure it failed before.

eric-wieser: Perhaps I accidentally changed it in #1510.

eric-wieser: Yep, I did - ok to add those lines back in this PR, and add the test for the expected `AttributeError`?

cmarqu: Okay for me, but it may take me a few days until I can work on this again. Feel free to adopt.

cmarqu: Like this?

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -470,3 +470,34 @@ def type_check_verilog(dut):
         tlog.info("Handle %s" %  (handle[0]._fullname,))
         if handle[0]._type != handle[1]:
             raise TestFailure("Expected %s found %s for %s" % (handle[1], handle[0]._type, handle[0]._fullname))
+
+
+# identifiers starting with `_` are illegal in VHDL
+@cocotb.test(skip=cocotb.LANGUAGE in ["vhdl"],
+             expect_error=AttributeError)
```
eric-wieser, line 477: Better to do a single try/except within the other test.

eric-wieser: Or rather, use https://github.com/cocotb/cocotb/blob/7db4c187e2c303d31f8226782b2afece6432eade/tests/test_cases/test_cocotb/common.py#L48

cmarqu: Done, did you mean like that?

eric-wieser: Yes!

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -64,3 +64,26 @@ async def test_delayed_assignment_still_errors(dut):
...
+# identifiers starting with `_` are illegal in VHDL
+@cocotb.test(skip=cocotb.LANGUAGE in ["vhdl"])
+async def test_access_underscore_name(dut):
+    """Test accessing HDL name starting with an underscore"""
+    # direct access does not work because we consider such names cocotb-internal
+    with assert_raises(AttributeError):
+        dut._underscore_name <= 0
+    await cocotb.triggers.Timer(1, 'ns')
+    with assert_raises(AttributeError):
+        assert dut._underscore_name.value == 0
```
eric-wieser, line 78: These test the same thing. I think we should simplify this to simply:

```suggestion
    with assert_raises(AttributeError):
        dut._underscore_name
```

I don't think we can test `__setattr__`, because `dut._underscore_name = 0` would just silently set a private attribute.

cmarqu: Done.

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -33,7 +33,6 @@
 from cocotb.result import TestError, TestFailure
 from cocotb.handle import IntegerObject, ConstantObject, HierarchyObject, StringObject
 
-
```
eric-wieser, line 36:
```suggestion
```

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -33,7 +33,6 @@
 from cocotb.result import TestError, TestFailure
 from cocotb.handle import IntegerObject, ConstantObject, HierarchyObject, StringObject
```
eric-wieser, line 35:
```suggestion


```

eric-wieser:
Looks like you removed a newline in an unrelated file which makes tests fail.

eric-wieser:
CI now fails, seems unable to find the signal. Something from the rebase perhaps? A new Icarus version?

Probably best to test locally to work out what's up.

cmarqu:
Yeah, trying with GitPod.

cmarqu:
This seems to be an Icarus problem; Verilator has no problem with it (after fixing #1777), as does Xcelium.
I opened https://github.com/steveicarus/iverilog/issues/322

cmarqu:
Added some "usage" of the reg (conditional on Icarus) which makes CI happy again.

tests/designs/sample_module/sample_module.sv
```diff
@@ -135,4 +135,12 @@ assign logic_a = stream_in_valid;
 always@* logic_b = stream_in_valid;
 always@(posedge clk) logic_c <= stream_in_valid;
 
+reg _underscore_name;
+`ifdef __ICARUS__
+    // By default, a variable must be used in some way in order
+    // to be visible to VPI in Icarus Verilog.
+    // See https://github.com/steveicarus/iverilog/issues/322
+    assign _underscore_name = 0;
+`endif
```
eric-wieser, line 144: Tempting to do this unconditionally, but not something that needs to hold up the release.

PR #1635, opened 2020-04-11
Ensure simulators are launched in 32-bit mode with 32-bit python

### LINKED ISSUE #1600: Set/use $(ARCH) based python binary version not system in the Makefiles
See: https://stackoverflow.com/questions/2208828/detect-64bit-os-windows-in-python

themperek:
Change `ARCH` to `PYTHON_ARCH` (32bit/64bit) and use Python binary version to decide about simulator options.

"If you've installed cocotb into a 32-bit python, then you'll only be able to run it with 32-bit simulators."

Closes #1600

### DIFF
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -36,9 +36,6 @@ COCOTB_PY_DIR := $(realpath $(shell cocotb-config --prefix))
 # simulations: Makefile fragments, and the simulator libraries.
 COCOTB_SHARE_DIR := $(COCOTB_PY_DIR)/cocotb/share
 
-ARCH?=$(shell uname -m)
-export ARCH
-
 OS=$(shell uname)
 ifneq (, $(findstring MINGW, $(OS)))
 OS=Msys
@@ -75,6 +72,11 @@ ifeq ($(OS),Msys)
 LIB_EXT    := dll
 else
 LIB_EXT    := so
+endif
+
+PYTHON_ARCH := $(shell $(PYTHON_BIN) -c 'from platform import architecture; print(architecture()[0])')
+ifeq ($(filter $(PYTHON_ARCH),64bit 32bit),)
+    $(error Unknown Python architecture: $(PYTHON_ARCH))
 endif
 
 # Set PYTHONHOME to properly populate sys.path in embedded python interpreter
--- a/cocotb/share/makefiles/simulators/Makefile.ius
+++ b/cocotb/share/makefiles/simulators/Makefile.ius
@@ -49,8 +49,8 @@ EXTRA_ARGS += $(COMPILE_ARGS)
 EXTRA_ARGS += $(SIM_ARGS)
 EXTRA_ARGS += -licqueue
 
-ifneq ($(ARCH),i686)
-EXTRA_ARGS += -64
+ifeq ($(PYTHON_ARCH),64bit)
+    EXTRA_ARGS += -64
 endif
 
 EXTRA_ARGS += -nclibdirpath $(SIM_BUILD)
--- a/cocotb/share/makefiles/simulators/Makefile.questa
+++ b/cocotb/share/makefiles/simulators/Makefile.questa
@@ -114,8 +114,8 @@ else
 	@echo "quit" >> $@
 endif
 
-ifneq ($(ARCH),i686)
-CMD += -64
+ifeq ($(PYTHON_ARCH),64bit)
+    CMD += -64
 endif
 
 $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/runsim.do
--- a/cocotb/share/makefiles/simulators/Makefile.vcs
+++ b/cocotb/share/makefiles/simulators/Makefile.vcs
@@ -51,7 +51,7 @@ else
      export VCS_BIN_DIR
 endif
 
-ifeq ($(ARCH),x86_64)
+ifeq ($(PYTHON_ARCH),64bit)
     EXTRA_ARGS += -full64
 endif
 
--- a/cocotb/share/makefiles/simulators/Makefile.xcelium
+++ b/cocotb/share/makefiles/simulators/Makefile.xcelium
@@ -49,7 +49,7 @@ EXTRA_ARGS += $(COMPILE_ARGS)
 EXTRA_ARGS += $(SIM_ARGS)
 EXTRA_ARGS += -licqueue
 
-ifneq ($(ARCH),i686)
+ifeq ($(PYTHON_ARCH),64bit)
     EXTRA_ARGS += -64
 endif
 


cocotb/_build_libs.py
```diff
@@ -149,6 +150,11 @@ def _extra_link_args(lib_name=None, rpath=None):
         args += ["-Wl,-install_name,@rpath/%s.so" % lib_name]
     if rpath is not None:
         args += ["-Wl,-rpath,%s" % rpath]
+
+    # allow 3bit cross compilation
```
eric-wieser, line 154:
```suggestion
    # allow 32bit cross compilation
```

cocotb/_build_libs.py
```diff
@@ -149,6 +150,11 @@ def _extra_link_args(lib_name=None, rpath=None):
...
+    if platform.architecture()[0] == "32bit":
+        args += ["-m32"]
```
eric-wieser, line 156: Doesn't python handle this for us? It wouldn't make any sense for it not to.

cocotb/share/makefiles/simulators/Makefile.questa
```diff
@@ -114,8 +114,8 @@ else
 	@echo "quit" >> $@
 endif
 
-ifneq ($(ARCH),i686)
-CMD += -64
+ifeq ($(PYTHON_ARCH),64bit)
+    CMD += -64
```
eric-wieser, line 118: Why == 64 here and != 64 there?

cocotb/_build_libs.py
```diff
@@ -188,6 +194,10 @@ def _get_python_lib():
 # Make PRI* format macros available with C++11 compiler but older libc, e.g. on RHEL6.
 _extra_cxx_compile_args += ["-D__STDC_FORMAT_MACROS"]
 
+# enable 32bit cross compilation
+if platform.architecture()[0] == "32bit":
+    _extra_cxx_compile_args += ["-m32"]
```
eric-wieser, line 199: Again, this doesn't make sense to me

cocotb/share/makefiles/simulators/Makefile.questa
themperek: No good reason. How you prefer?

cocotb/_build_libs.py
themperek: Unfortunately not. Python assumes gcc is also 32bit. I have checked.

cocotb/_build_libs.py
themperek: I think python assumes gcc is 32bit (than you do not need -m32).

eric-wieser: I'd appreciate it if you could find a reference for that in the python bugtracker.

Also, I think " support systems with 32-bit python but 64-bit gcc" would be more helpful in the comment

themperek: Only what I could find:
https://stackoverflow.com/questions/10478802/compiling-32-bit-python-modules-e-g-numpy-on-64-bit-linux

This can be different for different platforms potentially. Anaconda 32bit does not set "-m32" by default.

stuarthodgson:
I'm not entirely convinced on this. Why use python to determine if the build needs to be 32b? It's the simulator in use that loads the libs and boots they processes. The arch of python required is a result of this not the driver.

themperek:
@stuarthodgson Maybe I do not understand something but 
- If we want to use 32bit simulator we need to use 32bit python 
- If we use 32bit python we need to use 32bit simulator/version (determine this not based on the system but based on python binary version)?

cocotb/_build_libs.py
themperek: Can always use `CFLAGS="-m32" pip install ..` but I thought this can be automatic.

eric-wieser: > Anaconda 32bit does not set "-m32" by default.

Doesn't anaconda 32-bit come bundled with a 32-bit GCC?

eric-wieser:
@stuarthodgson: If you've installed cocotb into a 32-bit python, then you'll only be able to run it with 32-bit simulators. The part of this patch that makes sense to me is the bit that starts the simulator binary with the appropriate argument to match this python.

The other half seems redundant, but is also safe - you should only ever compile extension modules for the same bitness of python you intend to use them with, and in practice you only _build_ extension modules with the same python you intend to install into.

To support both 32-bit and 64-bit simulators, you need two python installations and two cocotb installations.

cocotb/_build_libs.py
themperek: Not by default (for me). Not sure how this looks if one compiles own python. 

Should better remove this?

eric-wieser: Yeah, I think I'd prefer it for now. Can put up as a separate patch, but I'd like a bugs.python.org reference before we put in a workaround - even if that means filing an issue ourselves.

cocotb/share/makefiles/simulators/Makefile.vcs
```diff
@@ -51,7 +51,7 @@ else
      export VCS_BIN_DIR
 endif
 
-ifeq ($(ARCH),x86_64)
+ifeq ($(PYTHON_ARCH),64bit)
```
eric-wieser, line 54:
```suggestion
ifneq ($(PYTHON_ARCH),32bit)
```
for consistency

cocotb/_build_libs.py
themperek: I have cross-compiled python on centos 7 and -m32 is added automatically in this case, so it looks like this is conda. Our instructions recommend conda+gcc so should be fine: https://github.com/cocotb/cocotb/wiki/Tier-2-Setup-Instructions#32-bit-python

Sorry for the confusion.

cocotb/share/makefiles/simulators/Makefile.ius
```diff
@@ -49,8 +49,8 @@ EXTRA_ARGS += $(COMPILE_ARGS)
 EXTRA_ARGS += $(SIM_ARGS)
 EXTRA_ARGS += -licqueue
 
-ifneq ($(ARCH),i686)
-EXTRA_ARGS += -64
+ifneq ($(PYTHON_ARCH),32bit)
```
imphil, line 52: Can we please change that to a positive match on 64bit (ifeq)? It's rather confusing to read it this way. (Also applies for all other uses below.)

cocotb/share/makefiles/Makefile.inc
```diff
@@ -36,8 +36,7 @@ COCOTB_PY_DIR := $(realpath $(shell cocotb-config --prefix))
 # simulations: Makefile fragments, and the simulator libraries.
 COCOTB_SHARE_DIR := $(COCOTB_PY_DIR)/cocotb/share
 
-ARCH?=$(shell uname -m)
-export ARCH
+PYTHON_ARCH := $(shell $(PYTHON_BIN) -c 'from platform import architecture; print(architecture()[0])')
```
imphil, line 39: Just as a safety net for the assumptions in all other code, please check for 32bit and 64bit and bail out if it's anything else.

cocotb/share/makefiles/simulators/Makefile.ius
themperek: You should agree with @eric-wieser what you like more ;-) (see above)

themperek: The idea as for now would be to default to 64 bit in any case that is not 32.

imphil: What would those cases be? That's why I said "assert" PYTHON_ARCH being either 32bit or 64bit. The code currently is easy to misread, as the if**n**def is easily missed.

Issue #1638, opened 2020-04-11
Remove deprecated "_compat_mapping"


cmarqu:
Use of ``_compat_mapping``:
https://github.com/cocotb/cocotb/blob/90023310e8f6b6c9124c725266733578de3665f8/cocotb/handle.py#L58-L66

has been giving deprecation warnings for a long time (https://github.com/cocotb/cocotb/blame/ba0fb097efbd2939ad6a961ee8d02793104ef8f7/cocotb/handle.py#L131), so  we can remove it.

eric-wieser:
Remember that deprecation warnings have been visible by default for much less time

cmarqu:
Indeed, I'm setting the 2.0 milestone for this. 

Related: #1642

marlonjames:
Removed in #3455

Issue #1504, opened 2020-03-17
Refactor of handle objects


ktbarrett:
Handles (in handle.py) support dot syntax for traversal of the HDL object hierarchy, but they also contain a number of attributes for using the value. These namespaces are known to collide and there currently is no mechanism to work around this issue. And to some degree it is unavoidable, Python objects have a number of builtin attributes (usually starting and ending with 2 underscores).

To solve this I recommend we split the traversal and analysis of HDL objects into two. Object traversal objects will only be able to be used to traverse the hierarchy. We can then obtain a `HDLValue` object which will contain all the other expected and useful behavior. One would obtain this value with an attribute that will not collide with valid identifiers in Verilog, VHDL or Python: `use` is a keyword in both languages and not a keyword in Python.

```python
type(dut.input)  # -> child of SimHandleBase
int(dut.input)  # -> Error
type(dut.input.use)  # -> HDLValue
int(dut.input.use)  # -> 0
```

Alternatively, since we should already mention Python builtin attributes names (`__{name}__`) will not work we could put such a function in a similar name (e.g. `__value__`), and provide a free function (or constructor) for access.

```python
type(HDLValue(dut.input))  # -> HDLValue
int(HDLValue(dut.input))  # -> 0
```

eric-wieser:
Can you give some concrete examples of name conflicts, and how each example would be spelt under each of your proposed new spellings?

eric-wieser:
(in particular, is it only `.value` that is problematic?)

ktbarrett:
I have run into `.value` before, it's the biggest offender. In that case the solution was to rename the object :/ This may not be easy depending upon the user. Formal verification processes are extremely bureaucratic, turn around time on a rename might be over a week. But *every* attribute on those classes is a valid Verilog identifier (VHDL identifiers can't start with `_`) and has the potential to collide.

* `setimmediatevalue`, `drivers`, `loads`, and `value`: valid Verilog and VHDL identifiers
* `_thing`: while they are "private", they are still valid Verilog identitifers
* `__operator__`: still valid Verilog identifiers, but unfortunately unavoidable

I just made this issue to highlight an implementation flaw that might one day be worked and poll for potential solution.

ktbarrett:
[...]

PR #1639, opened 2020-04-11
Document "private" members in cocotb.handle

### LINKED ISSUE #1636: Add documentation for _sub_handles, _id, _full_name ...
Add documentation for ``_sub_handles``, ``_id``, ``_full_name`` etc. since some of them are also useful for users.

cmarqu:
These members are starting with an underscore so as to not collide with
regular HDL names, though most are not really private in the usual meaning
but are useful for user code.

Some type hints are added to aid documentation.

Closes #1636

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -67,18 +67,42 @@ class SimHandleBase:
             path (str): Path to this handle, ``None`` if root.
         """
         self._handle = handle
-        self._len = None
-        self._sub_handles = {}  # Dictionary of children
-        self._invalid_sub_handles = set()  # Set of invalid queries
+        self._len = None  # type: int
+        """The "length" (the number of elements) of the underlying object. For vectors this is the number of bits."""
+        self._sub_handles = {}  # type: dict
+        """Dictionary of this handle's children."""
+        self._invalid_sub_handles = set()  # type: set
+        """Python :class:`set` of invalid queries, for caching purposes."""
+        self._name = self._handle.get_name_string()  # type: str
+        """The name of an object.
 
-        self._name = self._handle.get_name_string()
-        self._type = self._handle.get_type_string()
-        self._fullname = self._name + "(%s)" % self._type
-        self._path = self._name if path is None else path
+        :meta public:
+        """
+        self._type = self._handle.get_type_string()  # type: str
+        """The type of an object as a string.
+
+        :meta public:
+        """
+        self._fullname = self._name + "(%s)" % self._type  # type: str
+        """The name of an object with its type appended in parentheses."""
+        self._path = self._name if path is None else path  # type: str
+        """The path to this handle, or its name if this is the root handle.
+
+        :meta public:
+        """
         self._log = SimLog("cocotb.%s" % self._name)
+        """The logging object."""
         self._log.debug("Created")
-        self._def_name = self._handle.get_definition_name()
-        self._def_file = self._handle.get_definition_file()
+        self._def_name = self._handle.get_definition_name()  # type: str
+        """The name of a GPI object's definition.
+
+        :meta public:
+        """
+        self._def_file = self._handle.get_definition_file()  # type: str
+        """The file that sources the object's definition.
+
+        :meta public:
+        """
 
     def get_definition_name(self):
         return self._def_name
@@ -90,7 +114,7 @@ class SimHandleBase:
         return hash(self._handle)
 
     def __len__(self):
-        """Returns the 'length' of the underlying object.
+        """Return the "length" (the number of elements) of the underlying object.
 
         For vectors this is the number of bits.
         """
@@ -155,7 +179,7 @@ class RegionObject(SimHandleBase):
 
     def __init__(self, handle, path):
         SimHandleBase.__init__(self, handle, path)
-        self._discovered = False
+        self._discovered = False  # True if this object has already been discovered
 
     def __iter__(self):
         """Iterate over all known objects in this layer of hierarchy."""
@@ -176,8 +200,8 @@ class RegionObject(SimHandleBase):
                 yield handle
 
     def _discover_all(self):
-        """When iterating or performing tab completion, we run through ahead of
-        time and discover all possible children, populating the ``_sub_handles``
+        """When iterating or performing IPython tab completion, we run through ahead of
+        time and discover all possible children, populating the :any:`_sub_handles`
         mapping. Hierarchy can't change after elaboration so we only have to
         do this once.
         """
@@ -202,12 +226,12 @@ class RegionObject(SimHandleBase):
 
         self._discovered = True
 
-    def _child_path(self, name):
-        """Returns a string of the path of the child :any:`SimHandle` for a given *name*."""
+    def _child_path(self, name) -> str:
+        """Return a string of the path of the child :any:`SimHandle` for a given *name*."""
         return self._path + "." + name
 
     def _sub_handle_key(self, name):
-        """Translates the handle name to a key to use in ``_sub_handles`` dictionary."""
+        """Translate the handle name to a key to use in :any:`_sub_handles` dictionary."""
         return name.split(".")[-1]
 
     def __dir__(self):
@@ -266,7 +290,7 @@ class HierarchyObject(RegionObject):
         raise AttributeError("%s contains no object named %s" % (self._name, name))
 
     def __getattr__(self, name):
-        """Query the simulator for a object with the specified name
+        """Query the simulator for an object with the specified name
         and cache the result to build a tree of objects.
         """
         if name.startswith("_"):
@@ -281,10 +305,14 @@ class HierarchyObject(RegionObject):
 
         raise AttributeError("%s contains no object named %s" % (self._name, name))
 
-    def _id(self, name, extended=True):
-        """Query the simulator for a object with the specified name,
-        including extended identifiers,
+    def _id(self, name, extended: bool = True):
+        """Query the simulator for an object with the specified *name*,
         and cache the result to build a tree of objects.
+
+        If *extended* is ``True``, run the query only for VHDL extended identifiers.
+        For Verilog, only ``extended=False`` is supported.
+
+        :meta public:
         """
         if extended:
             name = "\\"+name+"\\"
@@ -300,7 +328,7 @@ class HierarchyArrayObject(RegionObject):
     """Hierarchy Arrays are containers of Hierarchy Objects."""
 
     def _sub_handle_key(self, name):
-        """Translates the handle name to a key to use in ``_sub_handles`` dictionary."""
+        """Translate the handle name to a key to use in :any:`_sub_handles` dictionary."""
         # This is slightly hacky, but we need to extract the index from the name
         #
         # FLI and VHPI(IUS):  _name(X) where X is the index
@@ -319,7 +347,7 @@ class HierarchyArrayObject(RegionObject):
             raise ValueError("Unable to match an index pattern: {}".format(name))
 
     def __len__(self):
-        """Returns the 'length' of the generate block."""
+        """Return the "length" of the generate block."""
         if self._len is None:
             if not self._discovered:
                 self._discover_all()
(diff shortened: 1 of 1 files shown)


cocotb/handle.py
```diff
@@ -74,18 +74,17 @@ def __init__(self, handle, path):
...
-        self._invalid_sub_handles = set()  # Set of invalid queries
-
-        self._name = simulator.get_name_string(self._handle)
-        self._type = simulator.get_type_string(self._handle)
-        self._fullname = self._name + "(%s)" % self._type
-        self._path = self._name if path is None else path
-        self._log = SimLog("cocotb.%s" % self._name)
+        self._len : int = None  #: The "length" (the number of elements) of the underlying object. For vectors this is the number of bits.
+        self._sub_handles : dict = {}  #: Dictionary of this handle's children.
+        self._invalid_sub_handles : set = set()  #: Python :keyword:`set` of invalid queries, for caching purposes.
```
eric-wieser, line 79: These two members strike me as actually private, since they're just a cache.

cocotb/handle.py
```diff
@@ -161,7 +160,7 @@ class RegionObject(SimHandleBase):
     """
     def __init__(self, handle, path):
         SimHandleBase.__init__(self, handle, path)
-        self._discovered = False
+        self._discovered : bool = False  #: ``True`` if this object has already been discovered.
```
eric-wieser, line 163: This syntax isn't legal in python 3.5, hence the CI failure.

ktbarrett:
I do not get why we are doing it this way. This concerns #1634, #1638, #1636, #1315, #1504 and a couple other issues probably.

Instead of levying requirements on user code and trying to tip-toe around the namespace collision problem, just segregate the namespaces. It wouldn't be much work to implement `__getindex__`/`__setindex__` to always do sub handle lookup. `__getattr__`/`__setattr__` could simply call those methods as an "alternative" syntax, you would just have to document that dot syntax does not always work and to prefer index syntax.

1. Nothing would have to be a special private-but-not-really attribute
2. No more requirements on user code
3. Users will be able to access literally any symbol, including those that contains dots

cmarqu:
> Not that I think these PRs aren't worthwhile, but they don't actually solve the problem, just side-step it.

True. But we will have to support the current style until the next major release at least, even if we had a new style in the code today.

cocotb/handle.py
cmarqu: We do have other docstrings (``_discover_all()``) pointing to ``_sub_handles``, so ... I'm undecided.
I agree that ``_invalid_sub_handles`` isn't something a user is going to be interested in.

cocotb/handle.py
cmarqu: Fixed.

documentation/source/library_reference.rst
```diff
@@ -195,6 +195,8 @@ Simulation Object Handles
 
 .. automodule:: cocotb.handle
     :members:
+    :private-members:
+    :exclude-members: _sub_handles, _invalid_sub_handles, _AssignmentResult, _SimIterator, _SetAction, _SetValueAction
```
eric-wieser, line 199: I'd prefer we write an exhaustive list of the members which _are_ public, rather than accidentally showing a private one

cocotb/handle.py
```diff
@@ -286,10 +296,11 @@ def __getattr__(self, name):
...
         raise AttributeError("%s contains no object named %s" % (self._name, name))
 
-    def _id(self, name, extended=True):
-        """Query the simulator for a object with the specified name,
-        including extended identifiers,
+    def _id(self, name, extended : bool = True):
+        """Query the simulator for an object with the specified *name*,
         and cache the result to build a tree of objects.
+
+        If *extended* is ``True``, run the query only for extended identifiers.
```
eric-wieser, line 303: After doing a bit more reading, `extended` is exclusively a VHDL feature, in verilog there is no such concept, only an escape sequence which makes `\foo<space>` and `foo` equivalent.

documentation/source/library_reference.rst
eric-wieser: Opened https://github.com/sphinx-doc/sphinx/pull/7473 to make this a little less painful.

cmarqu: Thanks, I ran out of time fighting with ``autoattribute`` and ``autodata`` without success.

cocotb/handle.py
```diff
@@ -74,18 +74,27 @@ def __init__(self, handle, path):
...
-        self._name = simulator.get_name_string(self._handle)
-        self._type = simulator.get_type_string(self._handle)
-        self._fullname = self._name + "(%s)" % self._type
-        self._path = self._name if path is None else path
+        self._len = None  # type: int
+        """The "length" (the number of elements) of the underlying object. For vectors this is the number of bits."""
+        self._sub_handles = {}  # type: dict
+        """Dictionary of this handle's children."""
+        self._invalid_sub_handles = set()  # type: set
+        """Python :keyword:`set` of invalid queries, for caching purposes."""
```
eric-wieser, line 82:
```suggestion
        """Python :class:`set` of invalid queries, for caching purposes."""
```

cocotb/handle.py
```diff
@@ -161,7 +170,8 @@ class RegionObject(SimHandleBase):
     """
     def __init__(self, handle, path):
         SimHandleBase.__init__(self, handle, path)
-        self._discovered = False
+        self._discovered = False  # type: bool
+        """``True`` if this object has already been discovered."""
```
eric-wieser, line 174: This strikes me as a private member

cocotb/handle.py
```diff
@@ -286,10 +296,11 @@ def __getattr__(self, name):
 
         raise AttributeError("%s contains no object named %s" % (self._name, name))
 
-    def _id(self, name, extended=True):
-        """Query the simulator for a object with the specified name,
-        including extended identifiers,
+    def _id(self, name, extended : bool = True):
```
eric-wieser, line 299:
```suggestion
    def _id(self, name, extended: bool = True):
```

documentation/source/library_reference.rst
eric-wieser: That PR was merged - do we want to consider using an unreleased Sphinx from master?

cmarqu: I wouldn't mind, if the pinning is not painful?

cocotb/handle.py
cmarqu: Converted to a simple comment.

cmarqu:
Works nicely: https://cocotb--1639.org.readthedocs.build/en/1639/library_reference.html#cocotb.handle.HierarchyObject._id
This PR now needs another review for what's deemed to be sufficiently public.

cocotb/handle.py
```diff
@@ -67,18 +67,48 @@ def __init__(self, handle, path):
...
-        self._len = None
-        self._sub_handles = {}  # Dictionary of children
-        self._invalid_sub_handles = set()  # Set of invalid queries
-
-        self._name = self._handle.get_name_string()
-        self._type = self._handle.get_type_string()
-        self._fullname = self._name + "(%s)" % self._type
-        self._path = self._name if path is None else path
+        self._len = None  # type: int
+        """The "length" (the number of elements) of the underlying object. For vectors this is the number of bits.
```
eric-wieser, line 71: This is private, it's a cache for `__len__`

cocotb/handle.py
```diff
@@ -67,18 +67,48 @@ def __init__(self, handle, path):
...
+        """
+        self._type = self._handle.get_type_string()  # type: str
+        """The type of an object as a string.
+
+        :meta public:
+        """
+        self._fullname = self._name + "(%s)" % self._type  # type: str
+        """The name of an object with its type appended in parentheses.
+
+        :meta public:
```
eric-wieser, line 92: This strikes me as kinda useless, and something we should look to remove (and therefore leave private)

cocotb/handle.py
```diff
@@ -67,18 +67,48 @@ def __init__(self, handle, path):
...
+        """
         self._log = SimLog("cocotb.%s" % self._name)
+        """The logging object."""
         self._log.debug("Created")
-        self._def_name = self._handle.get_definition_name()
-        self._def_file = self._handle.get_definition_file()
+        self._def_name = self._handle.get_definition_name()  # type: str
+        """The name of a GPI object's definition.
```
eric-wieser, line 103: It would be great to contrast this with `_name` if you can.

cocotb/handle.py
```diff
@@ -202,12 +232,15 @@ def _discover_all(self):
 
         self._discovered = True
 
-    def _child_path(self, name):
-        """Returns a string of the path of the child :any:`SimHandle` for a given *name*."""
+    def _child_path(self, name) -> str:
+        """Return a string of the path of the child :any:`SimHandle` for a given *name*.
+
+        :meta public:
```
eric-wieser, line 238: This one strikes me as private, it's not useful to the user who can just use `obj.some_child.path`

cocotb/handle.py
```diff
@@ -340,7 +376,10 @@ def __getitem__(self, index):
         return self._sub_handles[index]
 
     def _child_path(self, name):
-        """Returns a string of the path of the child :any:`SimHandle` for a given name."""
+        """Return a string of the path of the child :any:`SimHandle` for a given name.
+
+        :meta public:
```
eric-wieser, line 381: As above

cocotb/handle.py
cmarqu: I added ``_def_name`` and ``_def_file`` to ``test_discovery`` to see what it actually contains, and I found that only for ``HierarchyArrayObject`` the string is not empty: 
```
[...]
     0.00ns INFO     cocotb.sample_module                  test_discovery.py:44   in _discover                       Found name=stream_in_string def_name= def_file= type=(<class 'cocotb.handle.HierarchyObject'>)
     0.00ns INFO     cocotb.sample_module                  test_discovery.py:44   in _discover                       Found name=genblk1 def_name=sample_module def_file=/home/colin/tools/cocotb/tests/designs/sample_module/sample_module.sv type=(<class 'cocotb.handle.HierarchyArrayObject'>)
     0.00ns INFO     cocotb.sample_module                  test_discovery.py:44   in _discover                       Found name=genblk1[0] def_name= def_file= type=(<class 'cocotb.handle.HierarchyObject'>)
[...]
```
I'll open this as another issue in case it's a regression.
This may not be important enough to delay the release, @imphil? I would not make these public then?

cmarqu: Ooops, I now realize I don't have the latest code installed, let me try with that.

eric-wieser: I don't really mind much about this property, it was clearly intended to be public, even if it doesn't work very well.

cocotb/handle.py
```diff
@@ -281,10 +314,13 @@ def __getattr__(self, name):
...
         raise AttributeError("%s contains no object named %s" % (self._name, name))
 
-    def _id(self, name, extended=True):
-        """Query the simulator for a object with the specified name,
-        including extended identifiers,
+    def _id(self, name, extended: bool = True):
+        """Query the simulator for an object with the specified *name*,
         and cache the result to build a tree of objects.
+
+        If *extended* is ``True``, run the query only for VHDL extended identifiers.
```
eric-wieser, line 321:
```suggestion
        If *extended* is ``True``, run the query only for VHDL extended identifiers.
        For Verilog, only `extended=False` is supported.
```

cocotb/handle.py
cmarqu: Ok, I'll leave this public, but I'm not sure how to better describe what ``_def_name`` is given the small sample. Minor issue.

imphil:
Thanks everyone, let's get this in!

cocotb/handle.py
```diff
@@ -67,18 +67,42 @@ def __init__(self, handle, path):
...
+        """The type of an object as a string.
+
+        :meta public:
+        """
+        self._fullname = self._name + "(%s)" % self._type  # type: str
+        """The name of an object with its type appended in parentheses."""
+        self._path = self._name if path is None else path  # type: str
+        """The path to this handle, or its name if this is the root handle.
+
+        :meta public:
```
eric-wieser, line 91: I'm not convinced these markers have worked on these attributes...

cmarqu: Indeed, https://github.com/sphinx-doc/sphinx/pull/7473 seems to only work for functions.

PR #1641, opened 2020-04-12
Unpin Sphinx and breathe

eric-wieser:
The upstream bug in sphinxcontrib-domaintools has been fixed and released in 0.3, so we should be able to unpin these again.


eric-wieser:
Ping @cmarqu, in case you were waiting for a Sphinx 3.0 feature.

PR #1644, opened 2020-04-12
GHDL support on Windows

themperek:
Tests fail but everything starts correctly (32bit version 0.37) most likely due to GHDL itself.

```
     4.00ns ERROR    cocotb.regression                         regression.py:401  in _score_test                     Test Failed: adder_basic_test (result was TypeError)
                                                                                                                       File "C:\Users\Tomek\git\cocotb\examples\adder\tests\test_adder.py", line 21, in adder_basic_test
                                                                                                                         if int(dut.X) != adder_model(A, B):
                                                                                                                       File "c:\users\tomek\git\cocotb\cocotb\handle.py", line 691, in __int__
                                                                                                                         return int(self.value)
                                                                                                                       File "c:\users\tomek\git\cocotb\cocotb\handle.py", line 679, in value
                                                                                                                         result = BinaryValue(binstr, len(binstr))
                                                                                                                     TypeError: object of type 'NoneType' has no len()
     4.00ns INFO     cocotb.regression                         regression.py:418  in execute                         Running test 2/2: adder_randomised_test
     4.00ns INFO     ..t.adder_randomised_test.0x3b50d60       decorators.py:254  in _advance                        Starting test: "adder_randomised_test"
                                                                                                                     Description: Test for adding 2 random numbers multiple times
```


cocotb/share/makefiles/simulators/Makefile.ghdl
```diff
@@ -54,6 +54,11 @@ RTL_LIBRARY ?= work
 GHDL_ARGS ?=
 GHDL_ARGS += $(EXTRA_ARGS)
 
+# On Windows add GHDL lib folder to PATH to find libraries
+ifeq ($(OS),Msys)
+	export PATH := $(GHDL_BIN_DIR)/../lib:$(PATH)
```
cmarqu, line 59: This is a tab ... but shouldn't be, right?

themperek: heh

imphil:
There's no need to support the 32 bit version as long as a 64 bit version is available.

themperek:
> There's no need to support the 32 bit version as long as a 64 bit version is available.

64bit build they provide does not start for me in standalone even ;-(

Issue #1604, opened 2020-04-08
Move the concept of tests out of the scheduler


ktbarrett:
Spun off from #1499.

`RunningTest` duplicates all the information available on `test`. This duplication is only necessary because of way that tests are used by the regression manager. The test-specific bits in the scheduler follow from the same poor design choice in the regression manager.

The regression manager currently:
 * gathers tests
 * during initialization starts all the tests (calls the coroutines)
 * if the test initialization fails, it uses the `test` object's info for judgement
 * adds non-skipped `RunningTest` objects to a queue
 * runs each test by scheduling the coroutine
 * because the scheduler determines test end, it must call back into the regression manager (currently via a global reference)
 * pulls info from the `RunningTest` for test judgement, because that's all it has at this point

I think it should go something like this:
 * gathers tests and adds to queue
 * for each test schedule a regression-manager-bound coroutine that will handle test result
   * Don't need global access because bound method
   * have access to `test` object, so we don't need `RunningTest` duplication
 * initialize and deinitialize a regression-manager-local scheduler for each test

This also brings up the case for deprecating `RunningCoroutine`. It's just a `RunningTask` with a few extra interfaces: `__doc__` (silly), `_parent` (perhaps we should be doing this for `RunningTask`?), other two are redundant if we have parent.

This will decrease code size, complexity, reduce number of special objects, shrink the public interface, make scheduler and regression manager first-class objects, and improve post-test scheduler cleanup.

ktbarrett:
One more "feature" I'd like to take into account in the refactor is black-boxing test discovery, so we can experiment with a more pytest-like test discovery in the future.

ktbarrett:
I have started to test some of the changes I've made and decided to hold off tearing the scheduler down between tests for now, since I think that will require a lot of changes to the triggers to make them handle cleanup more gracefully. Also will hold off on deprecating `RunningCoroutine` and `RunningTest` for now. They can both be follow up PRs.

Right now the changes

ktbarrett:
So I found a way forward with the decoupling. It involves removing all references to tests from the scheduler and injecting behavior that decides when to end the test that is defined on the regression manager.

marlonjames:
The approach I started was to add callback functions to the `Task` that are called when it finishes. The regression manager would create the test `Task`, add the callback for handling the result, and then add it to the scheduler. All the scheduler needs to know is to call the callbacks.

ktbarrett:
Dupe of #791. xref #4280.

PR #1647, opened 2020-04-13
Some cleanup to Regression Manager

ktbarrett:
Addresses some issues from #1604 relating to the `RegressionManager`.

* Encapsulates all test/hook discovery into a single method
* Defers test initialization
* Saves `cocotb.test` function object for each test
* Uses the saved `cocotb.test` object to obtain test information, making `RunningTest` mostly useless
* Implements test timing, moving it out of `RunningTest`
* adds type annotations
* some shifting of responsibility/cleanup

This sets the stage for deprecation of `RunningTest` and the RM/scheduler refactor to remove all knowledge of tests from the scheduler.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -178,7 +178,6 @@ def _initialise_testbench(root_name):
     global regression_manager
 
     regression_manager = RegressionManager(root_name, modules, tests=test_str, seed=RANDOM_SEED, hooks=hooks)
-    regression_manager.initialise()
     regression_manager.execute()
 
     _rlock.release()
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -176,7 +176,7 @@ def _initialise_testbench(root_name):
 
     # start Regression Manager
     global regression_manager
-    regression_manager = RegressionManager(dut)
+    regression_manager = RegressionManager.from_discovery(dut)
     regression_manager.execute()
 
     _rlock.release()
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -165,19 +165,8 @@ def _initialise_testbench(root_name):
         log.info("Seeding Python random module with supplied seed %d" % (RANDOM_SEED))
     random.seed(RANDOM_SEED)
 
-    module_str = os.getenv('MODULE')
-    test_str = os.getenv('TESTCASE')
-    hooks_str = os.getenv('COCOTB_HOOKS', '')
-
-    if module_str is None:
-        raise ValueError("Environment variable MODULE, which defines the module(s) to execute, is not defined.")
-
-    modules = [s.strip() for s in module_str.split(',') if s.strip()]
-    hooks = [s.strip() for s in hooks_str.split(',') if s.strip()]
-
     global regression_manager
-
-    regression_manager = RegressionManager(root_name, modules, tests=test_str, seed=RANDOM_SEED, hooks=hooks)
+    regression_manager = RegressionManager(root_name)
     regression_manager.execute()
 
     _rlock.release()
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -165,8 +165,18 @@ def _initialise_testbench(root_name):
         log.info("Seeding Python random module with supplied seed %d" % (RANDOM_SEED))
     random.seed(RANDOM_SEED)
 
+    # Setup DUT object
+    from cocotb import simulator
+
+    handle = simulator.get_root_handle(root_name)
+    if not handle:
+        raise RuntimeError("Can not find root handle ({})".format(root_name))
+
+    dut = cocotb.handle.SimHandle(handle)
+
+    # start Regression Manager
     global regression_manager
-    regression_manager = RegressionManager(root_name)
+    regression_manager = RegressionManager(dut)
     regression_manager.execute()
 
     _rlock.release()
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -65,6 +65,7 @@ from cocotb.utils import get_sim_time, remove_traceback_frames, want_color_outpu
 from cocotb.xunit_reporter import XUnitReporter
 from cocotb.decorators import test as Test, hook as Hook, RunningTask
 from cocotb.outcomes import Outcome
+from cocotb.handle import SimHandle
 
 
 def _my_import(name: str) -> Any:
@@ -78,13 +79,12 @@ def _my_import(name: str) -> Any:
 class RegressionManager:
     """Encapsulates all regression capability into a single place"""
 
-    def __init__(self, root_name: str):
+    def __init__(self, dut: SimHandle):
         """
         Args:
-            root_name (str): The name of the root handle.
+            dut (SimHandle): The root handle to pass into test functions.
         """
-        self._root_name = root_name
-        self._dut = None
+        self._dut = dut
         self._running_test = None
         self._cov = None
         self.log = SimLog("cocotb.regression")
@@ -114,17 +114,6 @@ class RegressionManager:
             self.log.info("Enabling coverage collection of Python code")
             self._cov = coverage.coverage(branch=True, omit=["*cocotb*"])
             self._cov.start()
-
-        # Setup DUT object
-        #######################
-
-        handle = simulator.get_root_handle(self._root_name)
-
-        self._dut = cocotb.handle.SimHandle(handle) if handle else None
-
-        if self._dut is None:
-            raise AttributeError("Can not find Root Handle (%s)" %
-                                 self._root_name)
 
         # Test Discovery
         ####################
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -75,28 +75,17 @@ def _my_import(name):
 class RegressionManager:
     """Encapsulates all regression capability into a single place"""
 
-    def __init__(self, root_name, modules, tests=None, seed=None, hooks=[]):
+    def __init__(self, root_name):
         """
         Args:
             root_name (str): The name of the root handle.
-            modules (list): A list of Python module names to run.
-            tests (list, optional): A list of tests to run.
-                Defaults to ``None``, meaning all discovered tests will be run.
-            seed (int,  optional): The seed for the random number generator to use.
-                Defaults to ``None``.
-            hooks (list, optional): A list of hook modules to import.
-                Defaults to the empty list.
         """
         self._queue = []
         self._root_name = root_name
         self._dut = None
-        self._modules = modules
-        self._functions = tests
         self._running_test = None
         self._cov = None
         self.log = SimLog("cocotb.regression")
-        self._seed = seed
-        self._hooks = hooks
         self.start_time = time.time()
         self.test_results = []
         self.ntests = 0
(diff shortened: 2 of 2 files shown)


cocotb/regression.py
```diff
@@ -130,37 +117,49 @@ def initialise(self):
...
-                                 self._root_name)
+        # Process Hooks
+        ###################
+        for hook in self.hook_discovery():
+            self.log.info("Found hook {}.{}".format(hook.__module__, hook.__qualname__))
+            self._init_hook(hook)
 
-        # Test Discovery
-        ####################
+    def test_discovery(self) -> Iterable[Test]:
```
eric-wieser, line 137:
```suggestion
    def discover_tests() -> Iterable[Test]:
```
or perhaps
```suggestion
    def iter_discovered_tests(self) -> Iterable[Test]:
```

cocotb/regression.py
```diff
@@ -172,56 +171,50 @@ def initialise(self):
...
-            self.log.warning("No tests were discovered")
-
-        self._queue.sort(key=lambda test: (test.stage, test._id))
+                    yield thing
 
-        for valid_tests in self._queue:
-            self.log.info("Found test %s.%s" %
-                          (valid_tests.module,
-                           valid_tests.funcname))
+    def hook_discovery(self) -> Iterable[Hook]:
```
eric-wieser, line 184: Same conversation as above applies to this name

cocotb/regression.py
```diff
@@ -406,9 +400,16 @@ def _result_was():
 
         return result_pass, sim_failed
 
-    def execute(self):
-        self._running_test = cocotb.regression_manager.next_test()
-        if self._running_test:
+    def execute(self) -> None:
+        while True:
```
eric-wieser, line 404: This doesn't look like it should be looped, unless `add_test` has suddenly become blocking

cocotb/regression.py
```diff
@@ -406,9 +400,16 @@ def _result_was():
...
+            self._test = self.next_test()
+            if not self._test:
+                return self.tear_down()
+
+            self._running_test = self._init_test(self._test)
+            if not self._running_test:
+                continue
+
```
eric-wieser, line 412: Can you unindent everything below this line to be out of the loop?

cocotb/regression.py
eric-wieser: Nevermind, I now understand: see my below comment.

cocotb/regression.py
```diff
@@ -406,9 +400,16 @@ def _result_was():
 
         return result_pass, sim_failed
 
-    def execute(self):
-        self._running_test = cocotb.regression_manager.next_test()
-        if self._running_test:
+    def execute(self) -> None:
+        while True:
+            self._test = self.next_test()
+            if not self._test:
```
eric-wieser, line 406:
```suggestion
            if self._test is None:
```
is perhaps a little clearer

cocotb/regression.py
ktbarrett: This style looks a bit odd, but it will make sense in following commits and I didn't want to diff it back again. The unconditioned break below will be conditioned on whether the test finished after the call to `scheduler.add`. In the case of tests which finish immediately like:

```python
@cocotb.test()
async def returns_immediately(dut):
    return 1
```

eric-wieser: Can you perhaps then instead break out a helper `_execute_one(self) -> bool`, so that this instead becomes:

```
while True:
    self._test = self.next_test()
    returned_immediately = self._execute_one()
    if not returned_immediately:
        break

File: cocotb/regression.py [hunk 6]
```diff
@@ -75,31 +79,16 @@ def _my_import(name):
...
         Args:
-            root_name (str): The name of the root handle.
-            modules (list): A list of Python module names to run.
-            tests (list, optional): A list of tests to run.
-                Defaults to ``None``, meaning all discovered tests will be run.
-            seed (int,  optional): The seed for the random number generator to use.
-                Defaults to ``None``.
-            hooks (list, optional): A list of hook modules to import.
-                Defaults to the empty list.
+            dut (SimHandle): The name of the root handle.
```
eric-wieser, line 85:
```suggestion
            dut (SimHandle): The root handle to pass into test functions
```

ktbarrett:
Removing the test from the scheduler and moving all the bits of the execution to the RM is going to be one big nondecomposable commit, so I'd like it in it's own PR. Looking at doing additional cleanups in this PR. 
* centralize test result reporting
* try to avoid communication via member variables
* PEP8 cleanup

ktbarrett:
Had to fix a bug in the test numbering with `Running test A/B:". Also added highlighting to skipped tests.

ktbarrett:
Some of the test reporting has changed:

* in the test summary, skipped tests are listed in the order in which they would otherwise appear
* when "running" a skipped test it now highlights the line and adds the test number, much like it does when running the test
* now all tests are counted in the number of total tests and not just tests that aren't skipped
* test functions are not actually started if they are skipped which prevents spurious "coroutine never awaited" warnings

One thing that changed that might not be seen as an improvement is that the ratio time field of skipped tests is now `NaN` and not `0`.

ktbarrett:
Fixed one more small bug relating to an off by one in the number of tests run in the regression summary.

cocotb/__init__.py
```diff
@@ -165,20 +165,18 @@ def _initialise_testbench(root_name):
...
-    test_str = os.getenv('TESTCASE')
-    hooks_str = os.getenv('COCOTB_HOOKS', '')
+    # Setup DUT object
+    from cocotb import simulator
 
-    if module_str is None:
-        raise ValueError("Environment variable MODULE, which defines the module(s) to execute, is not defined.")
+    handle = simulator.get_root_handle(root_name)
+    if not handle:
+        raise AttributeError("Can not find Root Handle (%s)" % root_name)
```
eric-wieser, line 173: Probably would just go with `RuntimeError`, there's not really any attribute access happening here.

cocotb/regression.py
```diff
@@ -75,35 +79,18 @@ def _my_import(name):
 class RegressionManager:
     """Encapsulates all regression capability into a single place"""
 
-    def __init__(self, root_name, modules, tests=None, seed=None, hooks=[]):
+    def __init__(self, dut: SimHandle):
```
eric-wieser, line 82: It's tempting to split this in two, as:
```
def __init__(self, dut: SimHandle, tests: List[Test], hooks: List[Hook]):
   ...

@classmethod
def from_env_vars(cls, dut: SimHandle):
    # possibly just inline these methods
    tests = cls.discover_tests()
    hooks = cls.discover_hooks()
    return cls(dut, tests=tests, hooks=hooks)
```

Fine as a follow-up PR. One complexity is what logger to use before the manager is instantiated.

cocotb/regression.py
```diff
@@ -130,37 +115,51 @@ def initialise(self):
...
                 self.log.info("Traceback: ")
                 self.log.info(traceback.format_exc())
                 raise
 
-            if self._functions:
+            if test_str:
 
                 # Specific functions specified, don't auto-discover
-                for test in self._functions.rsplit(','):
+                for test in test_str.rsplit(','):
```
eric-wieser, line 162: Can you rename `test` here to `test_name`, and `_test` to `test`?

cocotb/regression.py
ktbarrett: It almost seems like instead of moving the discovery code into the RM, we should move it out and not have the named constructor. At that point we would use the root logger.

cocotb/regression.py
```diff
@@ -75,35 +79,18 @@ def _my_import(name):
...
-                Defaults to the empty list.
+            dut (SimHandle): The root handle to pass into test functions.
         """
-        self._queue = []
-        self._root_name = root_name
-        self._dut = None
-        self._modules = modules
-        self._functions = tests
+        self._dut = dut
         self._running_test = None
```
garmin-mjames, line 88: Having both `_test` and `_running_test` members is confusing. Maybe:
```suggestion
        self._test_task = None
```

cocotb/regression.py
eric-wieser: Or just use `SimLog("cocotb.regression")` in both places anyway, since that's a global logger instance.

cocotb/regression.py
eric-wieser: Seems reasonable.

cocotb/regression.py
eric-wieser: Looks like you didn't actually commit this change

cocotb/regression.py
```diff
@@ -261,84 +249,49 @@ def handle_result(self, test: RunningTask) -> None:
...
-            # Save results
-            self._store_test_result(test_func.__module__, test_func.__qualname__, result_pass, 0.0, 0.0, 0.0)
-            if not result_pass:
-                self.xunit.add_failure()
-                self.failures += 1
-            # Fail if required
-            if sim_failed:
-                self.tear_down()
-                raise SimFailure("Test initialization caused a simulator failure. Shutting down.")
+            self.log.error("Failed to initialize test %s" % test.__qualname__, exc_info=True)
```
eric-wieser, line 289: I think this needs to be
```suggestion
            self.log.error("Failed to initialize test %s" % test.__qualname__, exc_info=test_init_outcome.error)
```
since `exc_info` pulls from the currently thrown exception. I suspect this was already broken before your change.

ktbarrett:
Great I can't reopen this PR. GH has been spotty and didn't pick up on my pushed changes. I thought closing and reopening would work, it didn't...

cocotb/regression.py
```diff
@@ -130,111 +117,135 @@ def initialise(self):
...
+        return cls(dut, tests, hooks)
+
+    @staticmethod
+    def _discover_tests() -> Iterable[Test]:
+        """
+        Discovers tests in files automatically.
+
+        See :envvar:`MODULE` and :envvar:`TESTCASE` for details on how tests are discovered.
+        """
+        log = SimLog("cocotb.regression")
```
eric-wieser, line 160: Any reason not to put this at the top of the file at this point, perhaps as `_logger` to keep it private?

ktbarrett: I think this can be done to a few files so I'll do a sweep as a PR. Is there a better way than `_logger = logging.getLogger("cocotb.{}".format(__NAME__))`? Maybe some way that can detect cocotb as the package and return a package-relative name?

eric-wieser: `__name__` is already `"cocotb.regression"` here

ktbarrett: Could you link me to documentation to explain how it gets that name? I can't seem to find anything.

eric-wieser: Some discussion here: https://stackoverflow.com/q/15883526/102441

ktbarrett: I added the logger as requested using `__name__`.

cocotb/regression.py
```diff
@@ -56,54 +67,35 @@
...
-        self._queue = []
-        self._root_name = root_name
-        self._dut = None
-        self._modules = modules
-        self._functions = tests
-        self._running_test = None
+        self._dut = dut
+        self._test_task = None
         self._cov = None
         self.log = SimLog("cocotb.regression")
```
eric-wieser, line 95:
```suggestion
        self.log = _logger
```

eric-wieser:
Will rebase-merge once CI passes (much though I would prefer to regular merge, it's disabled on this repo ...)

PR #1649, opened 2020-04-13
Add support for launching IPython

### LINKED ISSUE #1081: Add support for integration with IPython's "autoawait" feature
At the moment, it is not possible to run cocotb coroutines from an IPython shell. 
IPython is able to run asynchronous code (see https://ipython.readthedocs.io/en/stable/interactive/autoawait.html), but support for this has to be added to the scheduler.

eric-wieser:
Closes #1081

Also need to decide where this should live.
* Should it be an extension, `cocotb_ext.ipython`?
   * How about  living in the extension namespace, but distributed as part of cocotb itself?
   * If not an extension, how should we handle the dependency on `IPython`?
* Should I split this into `ipython_runner` and `ipython_support`, where the former is used in a `MODULE=cocotb.ipython_runner` type example, while the latter is used for providing general tools like the custom prompt options I added here?

You can test this today by pasting the contents of this file into a standalone module.

An example session with the `dff` example:
```
eric@eric-thinkpa:~/win-home/Repos/cocotb/examples/dff/tests$ make MODULE=cocotb.ipython_support
make results.xml
make[1]: Entering directory '/mnt/c/Users/wiese/Repos/cocotb/examples/dff/tests'
MODULE=cocotb.ipython_support \
        TESTCASE= TOPLEVEL=dff TOPLEVEL_LANG=verilog COCOTB_SIM=1 \
        /usr/bin/vvp -M /mnt/c/Users/wiese/Repos/cocotb/cocotb/libs -m libcocotbvpi_icarus   sim_build/sim.vvp
     -.--ns INFO     cocotb.gpi                         ..mbed/gpi_embed.cpp:71   in set_program_name_in_venv        Did not detect Python virtual environment. Using system-wide Python interpreter
     -.--ns INFO     cocotb.gpi                         ../gpi/GpiCommon.cpp:106  in gpi_print_registered_impl       VPI registered
     0.00ns INFO     cocotb.gpi                                gpi_embed.cpp:332  in embed_sim_init                  Running on Icarus Verilog version 10.1 (stable)
     0.00ns INFO     cocotb.gpi                                gpi_embed.cpp:333  in embed_sim_init                  Python interpreter initialized and cocotb loaded!
     0.00ns INFO     cocotb                                      __init__.py:145  in _initialise_testbench           Running tests with cocotb v1.4.0.dev0 from /mnt/c/Users/wiese/Repos/cocotb/cocotb
     0.00ns INFO     cocotb                                      __init__.py:162  in _initialise_testbench           Seeding Python random module with 1586771832
     0.00ns INFO     cocotb.regression                         regression.py:194  in initialise                      Found test cocotb.ipython_support.run_ipython
     0.00ns INFO     cocotb.regression                         regression.py:422  in execute                         Running test 1/1: run_ipython
     0.00ns INFO     ..b.test.run_ipython.0x7f865bb9b6a0       decorators.py:255  in _advance                        Starting test: "run_ipython"
                                                                                                                     Description: Launch an interactive Python shell.
Python 3.5.2 (default, Nov 23 2017, 16:37:01)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.9.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from cocotb.clock import Clock

In [2]: cocotb.fork(Clock(dut.c, 10, 'us').start(start_high=False))
Out[2]: <cocotb.decorators.RunningTask at 0x7f86584e4e48>

In [3]: await cocotb.triggers.RisingEdge(dut.c)
Out[3]: RisingEdge(ModifiableObject(dff.c))

sim time: 5000000
In [4]: await cocotb.triggers.RisingEdge(dut.c)
Out[4]: RisingEdge(ModifiableObject(dff.c))

sim time: 15000000
In [5]:   
```

### DIFF
--- a/cocotb/ipython_support.py
+++ b/cocotb/ipython_support.py
@@ -1,0 +1,89 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+import IPython
+from IPython.terminal.ipapp import load_default_config
+from IPython.terminal.prompts import Prompts, Token
+
+import cocotb
+
+
+class SimTimePrompt(Prompts):
+    """ custom prompt that shows the sim time after a trigger fires """
+    _show_time = 1
+
+    def in_prompt_tokens(self, cli=None):
+        tokens = super().in_prompt_tokens()
+        if self._show_time == self.shell.execution_count:
+            tokens = [
+                (Token.Comment, "sim time: {}".format(cocotb.utils.get_sim_time())),
+                (Token.Text, "\n"),
+            ] + tokens
+        return tokens
+
+
+def _runner(shell, x):
+    """ Handler for async functions """
+    ret = cocotb.scheduler.queue_function(x)
+    shell.prompts._show_time = shell.execution_count
+    return ret
+
+
+async def embed(user_ns: dict = {}):
+    """
+    Start an ipython shell in the current coroutine.
+
+    Unlike using :func:`IPython.embed` directly, the :keyword:`await` keyword
+    can be used directly from the shell to wait for triggers.
+    The :keyword:`yield` keyword from the legacy :ref:`yield-syntax` is not supported.
+
+    This coroutine will complete only when the user exits the interactive session.
+
+    Args:
+        user_ns:
+            The variables to have made available in the shell.
+            Passing ``locals()`` is often a good idea.
+            ``cocotb`` will automatically be included.
+
+    Notes:
+
+        If your simulator does not provide an appropriate ``stdin``, you may
+        find you cannot type in the resulting shell. Using simulators in batch
+        or non-gui mode may resolve this. This feature is experimental, and
+        not all simulators are supported.
+    """
+    # ensure cocotb is in the namespace, for convenience
+    default_ns = dict(cocotb=cocotb)
+    default_ns.update(user_ns)
+
+    # build the config to enable `await`
+    c = load_default_config()
+    c.TerminalInteractiveShell.loop_runner = lambda x: _runner(shell, x)
+    c.TerminalInteractiveShell.autoawait = True
+
+    # create a shell with access to the dut, and cocotb pre-imported
+    shell = IPython.terminal.embed.InteractiveShellEmbed(
+        user_ns=default_ns,
+        config=c,
+    )
+
+    # add our custom prompts
+    shell.prompts = SimTimePrompt(shell)
+
+    # start the shell in a background thread
+    @cocotb.external
+    def run_shell():
+        shell()
+    await run_shell()
+
+
+@cocotb.test()
+async def run_ipython(dut):
+    """ A test that launches an interactive Python shell.
+
+    Do not call this directly - use this as ``make MODULE=cocotb.ipython_support``.
+
+    Within the shell, a global ``dut`` variable pointing to the design will be present.
+    """
+    await cocotb.triggers.Timer(0)  # workaround for gh-637
+    await embed(user_ns=dict(dut=dut))
--- a/cocotb/ipython_support.py
+++ b/cocotb/ipython_support.py
@@ -49,7 +49,7 @@ async def embed(user_ns: dict = {}):
 
         If your simulator does not provide an appropriate ``stdin``, you may
         find you cannot type in the resulting shell. Using simulators in batch
-        or non-gui mode may resolve this. This feature is experimental, and
+        or non-GUI mode may resolve this. This feature is experimental, and
         not all simulators are supported.
     """
     # ensure cocotb is in the namespace, for convenience
--- a/documentation/requirements.txt
+++ b/documentation/requirements.txt
@@ -17,3 +17,4 @@ pyenchant
 sphinx-issues
 sphinx-argparse
 towncrier
+IPython
--- a/documentation/source/newsfragments/1649.feature.rst
+++ b/documentation/source/newsfragments/1649.feature.rst
@@ -1,0 +1,1 @@
+Support for using ``await`` inside an embedded IPython terminal, using :mod:`cocotb.ipython_support`.
--- a/documentation/source/troubleshooting.rst
+++ b/documentation/source/troubleshooting.rst
@@ -64,3 +64,19 @@ accessible via a TCP socket:
    .. code:: shell
 
       telnet 127.0.0.1 4000
+
+
+Embedding an IPython shell
+==========================
+
+.. module:: cocotb.ipython_support
+
+.. versionadded:: 1.4
+
+A prebuilt test is included to easily launch an IPython shell in an existing design.
+
+.. autofunction:: run_ipython
+
+To embed a shell within an existing test, where it can inspect local variables, the :func:`embed` function can be used.
+
+.. autofunction:: embed
--- a/examples/dff/tests/Makefile
+++ b/examples/dff/tests/Makefile
@@ -32,6 +32,3 @@ endif
 
 include $(shell cocotb-config --makefiles)/Makefile.inc
 include $(shell cocotb-config --makefiles)/Makefile.sim
-
-# list all required Python files here
-sim: $(MODULE).py


cmarqu:
Awesome! Is Jupyter Notebook support within reach as well?

eric-wieser:
No, I had some issues attempting to use `embed_kernel` instead, as it seems that the use of `await` within the tornado webserver conflicted with our use of await. It's likely that we can copy whatever approach trio is using to get around this, but it was more effort than I wanted to spend.

eric-wieser:
Oh, also `jupyter notebook` doesn't support connecting to an existing kernel anyway: https://github.com/ipython/ipython/issues/4066

ktbarrett:
Excellent! That is surprisingly simple too. 

I'll need to try it out later to get a better feel for it, but initial questions:
* Does it hang the GUI? (I see `external` so I'm guessing it doesn't)
* Is there anything that cocotb supports that won't work interactively?

> Also need to decide where this should live. Should it be an extension, `cocotb_ext.ipython`?

I don't think it needs to live in an extension, it's fairly small, and if the `RegressionManager` deserves to live internally, this does too. 

Finally, there are tests in the regression for starting IPython that are currently skipped. Should they be changed or removed at this time?

eric-wieser:
> Is there anything that cocotb supports that won't work interactively?

I'd wager you might get into trouble using `function` and `external` interactively

> Does it hang the GUI? (I see external so I'm guessing it doesn't)

No idea, please report back!

> Finally, there are tests in the regression for starting IPython that are currently skipped. Should they be changed or removed at this time?

Good point, I should come back and remove that test. I can't think of a useful way to actually "test" the ipython integration, but let me know if you think of one.

cmarqu:
> > Also need to decide where this should live. Should it be an extension, `cocotb_ext.ipython`?
> 
> I don't think it needs to live in an extension

I agree.

ktbarrett:
I don't think this works with the Questa makefiles w/ `GUI=1`, I don't see how to run commands.

eric-wieser:
> I don't think this works with the Questa makefiles w/ `GUI=1`, I don't see how to run commands =(.

That doesn't really surprise me, I've had similar experiences trying to use `pdb` before. The cause is that modelsim hasn't really connected stdin to anything useful. This is where getting `IPython.embed_kernel` working would be handy, but I didn't want to try to solve to many problems at once

> Can't currently test in batch mode, due to license issues.

Are these transient issues, or should I seek others to test?

ktbarrett:
> Are these transient issues, or should I seek others to test?
Site issues, nothing wrong with cocotb.

eric-wieser:
Sorry, what I meant by that is: Will you be able to test modelsim batch mode in the near future, or are your site issues likely to persist for a while?

ktbarrett:
I will try to test tonight. Are you trying to gather which simulators are supported for documentation?

gretzteam:
This is cool! Two questions:

1. I tried with both verilator (vcd) and xcelium (shm), and can't get waveform outputs in 'real time' while in ipython. Verilator does create the vcd but it doesn't advance, and xcelium never create the shm - even though it's there on a normal sim.

2. Would it be possible to 'drop' into a working IPython shell as part of a normal test? I currently debug by using putting an embed() in my python test code. This allowed to probe around and debut the testbench, but none of it was connected to the simulation.

eric-wieser:
@ktbarrett

> I will try to test tonight. Are you trying to gather which simulators are supported for documentation?

I wasn't, but that sounds like it's a good idea, so now I will be

eric-wieser:
@gretzteam 

> I tried with both verilator (vcd) and xcelium (shm), and can't get waveform outputs in 'real time' while in ipython.

I'm not sure what you mean by that. Can you show me what you were trying to do which didn't work?

> Would it be possible to 'drop' into a working IPython shell as part of a normal test?

Yes, that's definitely possible. I'll push an update with a change to enable that.

eric-wieser:
@gretzteam: `cocotb.ipython_support.embed(user_ns=locals())` probably ticks that box.

gretzteam:
@eric-wieser 

> I tried with both verilator (vcd) and xcelium (shm), and can't get waveform outputs in 'real time' while in ipython.

While entering commands in the IPython shell, we are making the simulation advance. It would be nice that the waveform output show that in real time. 
Each simulator have their own way to output waveforms so this might be simulator dependent. Here I have both verilator (producing a vcd) and xcelium (producing a shm) working. However, neither those files were getting updated as I made the simulation advance in IPython.

Essentially, if I have a vcd of the simulation open, I would like to be able to type in IPython:
>dut.din <= 1

reload the vcd in gtkwave, and see the signal set high...

eric-wieser:
You need to run `await Timer(1)` or something for that write to take effect. `<=` is a delayed write that does not run until the next `ReadWrite` cycle.

ktbarrett:
@gretzteam That's really up to the simulator. We can't make those guarantees.

gretzteam:
Sure you need a delay, this was just an example. 
I was using the await clock when I tried.
I can see that some simulators do not write out every single event as they happen, so it's hard to guarantee especially for a few events..

gretzteam:
Small update: this works with xcelium when outputting a vcd. You need to advance the simulation by some 'amount' before it dumps into the vcd, but you can then reload and see the waveform.
It doesn't work with shm dump since the file is 'locked' until I exit the IPython.

csquaredphd:
Thanks, this is great!

Using `make MODULE=cocotb.ipython_support`, I was able to test this using `/examples/adder`.  I used Icarus and was able to see the VCD updated during the sim (after adding a `$dumpflush()`).

I also tried adding an interactive regression test, but I think I'm missing something.  I added this to `test_adder.py`:
```python
@cocotb.test()
def adder_interactive_test(dut):
    yield Timer(2, units='ns')
    cocotb.ipython_support.embed(user_ns=locals())
```
And got this error:
```
(cocotb_3.7) cc284@tmpo-crc-l01:/c/Projects/sim_tools/cocotb/examples/adder/tests$ make
make results.xml
make[1]: Entering directory '/c/Projects/sim_tools/cocotb/examples/adder/tests'
MODULE=test_adder \
        TESTCASE= TOPLEVEL=adder TOPLEVEL_LANG=verilog COCOTB_SIM=1 \
        /usr/bin/vvp -M /c/Projects/sim_tools/cocotb/cocotb/libs -m libcocotbvpi_icarus   sim_build/sim.vvp
     -.--ns INFO     cocotb.gpi                         ..mbed/gpi_embed.cpp:94   in set_program_name_in_venv        Using Python virtual environment interpreter at /mnt/c/Projects/pyenv/cocotb_3.7/bin/python
     -.--ns INFO     cocotb.gpi                         ../gpi/GpiCommon.cpp:106  in gpi_print_registered_impl       VPI registered
     0.00ns INFO     cocotb.gpi                                gpi_embed.cpp:332  in embed_sim_init                  Running on Icarus Verilog version 10.1 (stable)
     0.00ns INFO     cocotb.gpi                                gpi_embed.cpp:333  in embed_sim_init                  Python interpreter initialized and cocotb loaded!
     0.00ns INFO     cocotb                                      __init__.py:145  in _initialise_testbench           Running tests with cocotb v1.4.0.dev0 from /c/Projects/sim_tools/cocotb/cocotb
     0.00ns INFO     cocotb                                      __init__.py:162  in _initialise_testbench           Seeding Python random module with 1587082030
     0.00ns INFO     cocotb.regression                         regression.py:194  in initialise                      Found test test_adder.adder_basic_test
     0.00ns INFO     cocotb.regression                         regression.py:194  in initialise                      Found test test_adder.adder_interactive_test
     0.00ns INFO     cocotb.regression                         regression.py:194  in initialise                      Found test test_adder.adder_randomised_test
     0.00ns INFO     cocotb.regression                         regression.py:422  in execute                         Running test 1/3: adder_basic_test
     0.00ns INFO     ..t.adder_basic_test.0x7f32ee2c3b00       decorators.py:255  in _advance                        Starting test: "adder_basic_test"
                                                                                                                     Description: Test for 5 + 10
VCD info: dumpfile dump.vcd opened for output.
     4.00ns INFO     cocotb.adder                              test_adder.py:25   in adder_basic_test                Ok!
     4.00ns INFO     cocotb.regression                         regression.py:366  in _score_test                     Test Passed: adder_basic_test
     4.00ns INFO     cocotb.regression                         regression.py:422  in execute                         Running test 2/3: adder_interactive_test
     4.00ns INFO     ..r_interactive_test.0x7f32ee2c3be0       decorators.py:255  in _advance                        Starting test: "adder_interactive_test"
                                                                                                                     Description: None
     6.00ns ERROR    cocotb.regression                         regression.py:401  in _score_test                     Test Failed: adder_interactive_test (result was AttributeError)
                                                                                                                       File "/c/Projects/sim_tools/cocotb/examples/adder/tests/test_adder.py", line 30, in adder_interactive_test
                                                                                                                         cocotb.ipython_support.embed(user_ns=locals())
                                                                                                                     AttributeError: module 'cocotb' has no attribute 'ipython_support'
```

ktbarrett:
@csquaredphd Did you `import cocotb.ipython_support`? Unlike the rest of the core libraries this module isn't imported into the base package.

csquaredphd:
> import cocotb.ipython_support

That fixed the error, but now it passes without ever dropping into IPython:
```
     4.00ns INFO     cocotb.regression                         regression.py:422  in execute                         Running test 2/3: adder_interactive_test
     4.00ns INFO     ..r_interactive_test.0x7f50235f8588       decorators.py:255  in _advance                        Starting test: "adder_interactive_test"
                                                                                                                     Description: None
/c/Projects/sim_tools/cocotb/examples/adder/tests/test_adder.py:31: RuntimeWarning: coroutine 'embed' was never awaited
  cocotb.ipython_support.embed(user_ns=locals())
     6.00ns INFO     cocotb.regression                         regression.py:366  in _score_test                     Test Passed: adder_interactive_test
```

ktbarrett:
@csquaredphd See how the [builtin test](https://github.com/cocotb/cocotb/pull/1649/files#diff-d0495473378e7b953d5ba25f8c4c8b7dR71-R76) uses it. `embed` is an async function, you'll need to wait on it.

Try:
```python
@cocotb.test()
def adder_interactive_test(dut):
    yield Timer(2, units='ns')
    yield cocotb.ipython_support.embed(user_ns=locals())
```

or using the more modern `await`/`async` syntax:
```python
@cocotb.test()
async def adder_interactive_test(dut):
    await Timer(2, units='ns')
    await cocotb.ipython_support.embed(user_ns=locals())
```

ktbarrett:
@eric-wieser Looks like it doesn't work in Questa, even in batch mode. This is likely due to how the makefiles are setup where there is a nested invocation.

eric-wieser:
> This is likely due to how the makefiles are setup where there is a nested invocation.

Can you elaborate on this thought? Are you talking about the way `make sim` reinvokes make? This works fine for me on icarus and WSL.

elgorwi:
@ktbarrett could it be because questa is using a tcl prompt rather than a normal command prompt?

eric-wieser:
Updated with docs at https://cocotb--1649.org.readthedocs.build/en/1649/troubleshooting.html#embedding-an-ipython-shell

cmarqu:
Do you see this mainly as a troubleshooting tool? I may have put it as a new section under "How-to Guides".
Granted, having it under "Troubleshooting" doesn't make the sidebar longer which is a plus if we get more topics....

cocotb/ipython_support.py
```diff
@@ -0,0 +1,89 @@
...
+
+    This coroutine will complete only when the user exits the interactive session.
+
+    Args:
+        user_ns:
+            The variables to have made available in the shell.
+            Passing ``locals()`` is often a good idea.
+            ``cocotb`` will automatically be included.
+
+    Notes:
```
cmarqu, line 48: Use a Sphinx ``.. note::``?

cocotb/ipython_support.py
```diff
@@ -0,0 +1,89 @@
...
+
+        If your simulator does not provide an appropriate ``stdin``, you may
+        find you cannot type in the resulting shell. Using simulators in batch
+        or non-gui mode may resolve this. This feature is experimental, and
```
cmarqu, line 52:
```suggestion
        or non-GUI mode may resolve this. This feature is experimental, and
```

cocotb/ipython_support.py
eric-wieser: I think this is part of the `sphinx.napoleon` syntax, but I suspect we don't use it elsewhere. Happy to go either way.

cmarqu:
Re the docstring part about not being able to type when in GUI mode:
I have seen cases in the past where you could not type into the simulator GUI's "Console" window (that's what Cadence calls it), but it worked when you typed into the shell where you started the simulator from. 
In case this works here too it's probably not useful for regular IPython usage, but may be enough to gracefully exit IPython.

eric-wieser:
> Do you see this mainly as a troubleshooting tool? I may have put it as a new section under "How-to Guides".
> Granted, having it under "Troubleshooting" doesn't make the sidebar longer which is a plus if we get more topics....

It seemed quite closely related to pdb, and suffers from the same problems regarding simulator stdin. Perhaps more of a debugging tool for the dut rather than the test?

ktbarrett:
@cmarqu I like the idea of using this for interactively testing units during their development, before building up a proper testbench which can take time. I might also use this to probe the design for bugs, and once I find something that breaks I can save it to a file to be added to the actual regressions. It's also a good way for beginners to "step" through their code to see how it works.

Of course, I haven't tried any of this yet.

cmarqu:
@ktbarrett Yes, I also imagine bringing up a mixed-signal simulation in steps, where you let supply ramp up a bit, then drop into IPython where you look at signals in the analog part and their effect on the digital part, the ramp up some more until the interfaces start working, try some communication etc.
I have wasted CPU hours or even days in the past with simulations where I started communication too early to have an effect.

I can imagine that we need some cocotb-specific information on how to best work with IPython, ideally accessible from within the simulation environment.
@eric-wieser, is it possible to append some paragraphs of help to IPythons ``?`` magic command? Also, can we automatically save whatever the user typed into the IPython shell and keep it for use after the simulation? A new simulation run should maybe not even overwrite that file but rather append or write timestamped files.
Maybe we can even preload the IPython history with the history from the previous run?

eric-wieser:
> Maybe we can even preload the IPython history with the history from the previous run?

I believe it all gets saved in the global ipython history.

eric-wieser:
> @eric-wieser, is it possible to append some paragraphs of help to IPythons `?` magic command?

I'm not sure what you mean by this. Are you suggesting we mention `?` in the docs, or are you trying to hook into `?` itself?

eric-wieser:
> Also, can we automatically save whatever the user typed into the IPython shell and keep it for use after the simulation?

They can type `%save` to do so, I don't think we need to make it automatic. Perhaps we should draw attention to it in the docs though.

cmarqu:
> I'm not sure what you mean by this. Are you suggesting we mention `?` in the docs, or are you trying to hook into `?` itself?

The latter - ``?`` prints some generic help, while it may be useful to also mention cocotb specifics. (Not yet sure what they would be.)

eric-wieser:
> while it may be useful to also mention cocotb specifics. (Not yet sure what they would be.)

Ok with leaving that to a follow-up, I assume? Replacing `?` would probably mean replacing `InteractiveShell.show_usage`, which wouldn't be too hard - but we probably want to leave all the Ipython text there too.

cmarqu:
> > while it may be useful to also mention cocotb specifics. (Not yet sure what they would be.)
> 
> Ok with leaving that to a follow-up, I assume?

Yes. We should see how people actually use this feature and what problems they have (if any).

themperek:
Is there any reason not to merge this?

cmarqu:
> It's likely that we can copy whatever approach trio is using to get around this, but it was more effort than I wanted to spend.

I saw this recently, not sure if it's helpful: https://gist.github.com/mikenerone/786ce75cf8d906ae4ad1e0b57933c23f

PR #1651, opened 2020-04-13
Use VHDL as top language for GHDL on travis and update version to 0.37

themperek:


imphil:
Can you:
- Split this into two commits for the TOPLEVEL_LANG changes, and another one for the update.
- Keep the summary line of the commit message around 50 characters (see https://github.com/cocotb/cocotb/blob/master/CONTRIBUTING.md#patch-requirements)
- Add a short description what the TOPLEVEL_LANG change does, and why it is necessary.

PR #1652, opened 2020-04-14
Make `gpi_iterator_hdl` objects natively iterable

eric-wieser:
This saves some python wrapping and some argument parsing

Follow up to #1540.

This removes `cocotb.simulator.next(iterator_hdl)`, which is now just spelt `next(iterator_hdl)`.


imphil:
(Merge once CI passes.)

PR #1653, opened 2020-04-14
Move simulator module methods to instance methods

eric-wieser:
Another follow-up to #1540 and #1652

This changes instances of `simulator.some_function(some_handle, args)` to `some_handle.some_function(args)`.

This removes a lot of type-checking from our code, making it significantly shorter.

cc @ktbarrett, who expressed some concerns about doing this, I think regarding #1332 (and therefore also @mballance).

Edit: Both are in approval of this patch, so looks like this is good to go.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -77,14 +77,14 @@ class SimHandleBase:
         self._sub_handles = {}  # Dictionary of children
         self._invalid_sub_handles = set()  # Set of invalid queries
 
-        self._name = simulator.get_name_string(self._handle)
-        self._type = simulator.get_type_string(self._handle)
+        self._name = self._handle.get_name_string()
+        self._type = self._handle.get_type_string()
         self._fullname = self._name + "(%s)" % self._type
         self._path = self._name if path is None else path
         self._log = SimLog("cocotb.%s" % self._name)
         self._log.debug("Created")
-        self._def_name = simulator.get_definition_name(self._handle)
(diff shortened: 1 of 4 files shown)


mballance:
I really like this approach, and believe it will simplify implementation of #1332, since the core simulator API will be smaller.

eric-wieser:
I'm glad to hear that, since it was your PR which motivated this change in the first place.

ktbarrett:
I like this a lot. I can see that it won't increase the complexity of a mock out, so I have no concerns. One general comment is renaming all the methods, we don't need "get" and "set" in all the names.

I'd like it if we didn't have to name the types in `set_signal_value_long`, but there are a bunch of issues with that. pybind11 would make that easier as well.

eric-wieser:
> One general comment is renaming all the methods, we don't need "get" and "set" in all the names.

Whether or not I agree with that, I was trying to minimize the impact of this patch, and renames would have thwarted my find and replaces (especially since I had to write this multiple times, it wasn't readily rebaseable onto the C++ version of #1540 that got merged).

Perhaps best to leave renames to a follow up?

> I'd like it if we didn't have to name the types in `set_signal_value_long` etc. However, there are a bunch of issues following that.

All that would really achieve is moving the type-dispatching from the python code where it's easy to write to the C code where it's hard to write. In practice the dispatching is complex anyway, as python integers end up needing to be passed to `binstr` if they're too long.

PR #1658, opened 2020-04-16
Tidy python bindings in the endian_swapper example

eric-wieser:
See individual commit messages.

It's possible but unlikely that this solves @ktbarrett's bug report from gitter, https://gitter.im/cocotb/Lobby?at=5e977cf563e7b73a5fda24af

### DIFF
--- a/examples/endian_swapper/cosim/io.c
+++ b/examples/endian_swapper/cosim/io.c
@@ -113,7 +113,6 @@ int IOWR(unsigned int base, unsigned int address, unsigned int value)
     return 0;
 }
 
-#if PY_MAJOR_VERSION >= 3
 static struct PyModuleDef io_module =
 {
     PyModuleDef_HEAD_INIT,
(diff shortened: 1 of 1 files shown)


examples/endian_swapper/cosim/io.c
```diff
@@ -36,84 +36,78 @@
...
-    if (PyErr_Occurred())
-        PyErr_Print();
-
     Py_DECREF(rv);
-    Py_DECREF(call_args);
-
     return 0;
 }
 
-#if PY_MAJOR_VERSION >= 3
```
eric-wieser, line 116: xref #1289

themperek:
Same error for me whit swig 2.0.10

themperek:
Maybe we should run this test only in CI for now since documentation point to this as the first things to do and it is has many issues not related to what is important.

eric-wieser:
Agreed, the big issue with this example right now is that it's the one we point users to first.

imphil:
To keep this PR focused I moved the discussion beyond that to https://github.com/cocotb/cocotb/issues/1660.

Issue #1657, opened 2020-04-16
Add type checking to SimHandle non-blocking writes


ktbarrett:
When using `setimeddiatevalue` to set a signal's value, it does type checking to ensure it can pass a valid value to the C layer. However, the non-blocking write methods `value` and the syntax `<=` depend upon the type checking in `setimmediatevalue`. If these non-blocking writes are scheduled, the type checking it not done immediately. The exception will throw in the internal writer coroutine instead of at the use site.

Given this code:

```python
@cocotb.test()
async def test_breaks(dut):
    dut.stream_in_data <= "wrong type"
    await cocotb.triggers.Timer(10)
```

I get this stacktrace:
```
Traceback (most recent call last):
  File "/home/kaleb/dev/cocotb/cocotb/scheduler.py", line 260, in _do_writes
    handle.setimmediatevalue(value)
  File "/home/kaleb/dev/cocotb/cocotb/handle.py", line 648, in setimmediatevalue
    .format(type(value), value))
TypeError: Unsupported type for value assignment: <class 'str'> ('wrong type')
```

Move type/value checking for a given `SimHandle` subtype into a central function and use it in all write methods.

eric-wieser:
My fear is it's worse than that, and all future writes fail because the writing thread crashed

ktbarrett:
In the writer coroutine I'm guessing we should catch, log, and squash?

eric-wieser:
I have a better plan, which I started prototyping but don't have time today to finish. The idea would be to queue the fully-processed simulator calls from `handle.py`, with no further checks or computations. I'll push it to a branch for you to take a look.

PR #1661, opened 2020-04-16
Ensure errors happen immediately when using delayed assignment

eric-wieser:
Fixes gh-1657.

This possibly results in the docs being a bit less useful, as now every handle type has a `.value` property with exactly the same documentation. The documentation would be a lot more useful with https://github.com/sphinx-doc/sphinx/issues/7383, but alas that's probably a way off.

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -229,8 +229,8 @@ class Scheduler:
         # Our main state
         self._mode = Scheduler._MODE_NORMAL
 
-        # A dictionary of pending writes
-        self._writes = _py_compat.insertion_ordered_dict()
+        # A list of pending (write_func, args)
+        self._write_calls = []
 
         self._pending_coros = []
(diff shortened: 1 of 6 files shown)


cocotb/scheduler.py
```diff
@@ -503,7 +503,8 @@ def unschedule(self, coro):
                 e = remove_traceback_frames(e, ['unschedule', 'get'])
                 self._test.abort(e)
 
-    def save_write(self, handle, value):
+    def _schedule_write(self, handle, write_func, *args):
+        """ Queue `write_func` to be called on the next ReadWrite trigger. """
```
cmarqu, line 507:
```suggestion
        """ Queue *write_func* to be called on the next ReadWrite trigger. """
```

eric-wieser:
@ktbarrett, mind looking over this too, since it sounds like you thought about it as well?

PR #1663, opened 2020-04-16
Show how to set SIM to start each simulator

cmarqu:


documentation/source/simulator_support.rst
```diff
@@ -115,25 +140,44 @@ Aldec Riviera-PRO
...
 .. _sim-modelsim:
 
 Mentor ModelSim
 ===============
 
+In order to use this simulator, set :make:var:`SIM` to ``modelsim``:
+
+.. code-block:: bash
+
+    make SIM=modelsim
```
themperek, line 179: I am not sure if we should not do `Mentor Questa/Modelsim` since this is one and the same at this point?

cmarqu: Maybe... But not in this PR :)

PR #1665, opened 2020-04-17
More small cleanups to the simulator extension module

eric-wieser:
Once again, the less code the better.

### DIFF
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -270,6 +270,24 @@ static PyObject *log_msg(PyObject *self, PyObject *args)
 }
 
 
+static p_callback_data make_callback_data(PyObject *func, PyObject *args, PyObject *kwargs)
+{
+    p_callback_data data = (p_callback_data)malloc(sizeof(s_callback_data));
+    if (data == NULL) {
+        PyErr_NoMemory();
+        return NULL;
+    }
+
+    data->_saved_thread_state = PyThreadState_Get();
+    data->id_value = COCOTB_ACTIVE_ID;
+    data->function = func;
+    data->args = args;
+    data->kwargs = kwargs;
+
+    return data;
+}
+
+
 // Register a callback for read-only state of sim
 // First argument is the function to call
 // Remaining arguments are keyword arguments to be passed to the callback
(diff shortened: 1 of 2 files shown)


cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -270,6 +270,24 @@ static PyObject *log_msg(PyObject *self, PyObject *args)
 }
 
 
+static p_callback_data make_callback_data(PyObject *func, PyObject *args, PyObject *kwargs)
+{
+    p_callback_data data = (p_callback_data)malloc(sizeof(s_callback_data));
+    if (data == NULL) {
+        PyErr_NoMemory();
+        return NULL;
```
imphil, line 278:
```suggestion
        return PyErr_NoMemory();
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -270,6 +270,24 @@ static PyObject *log_msg(PyObject *self, PyObject *args)
 }
 
 
+static p_callback_data make_callback_data(PyObject *func, PyObject *args, PyObject *kwargs)
```
imphil, line 273:
```suggestion
static p_callback_data p_callback_data_new(PyObject *func, PyObject *args, PyObject *kwargs)
```

to indicate it's an allocating function for the p_callback_data type.

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: Doesn't work, because `decltype(PyErr_NoMemory())` is `PyObject *` which is an incompatible pointer type.

cocotb/share/lib/simulator/simulatormodule.cpp
eric-wieser: Should that be `s_callback_data_new`, since `p_callback_data` is `s_callback_data*`?

imphil: Would make sense, yes. That being said, the naming of this struct and the two associated typedefs look quite unusual to me. I'd go for a normal (yet not fully clean naming-wise) `callback_data_t` or (my preferred option) simply using `struct callback_data`.

imphil: (The struct naming is obviously not part of this PR.)

eric-wieser: I've gone ahead and added an extra commit to rename `s_callback_data` to `callback_data`. This is C++ code now, so `struct callback_data` and `callback_data` are synonymous anyway.

PR #1666, opened 2020-04-17
Remove dead code

eric-wieser:
The C++ standard mandates that `new T ` can never return NULL.
If an allocation fails, it must throw `std::bad_alloc`.

The only way `new` can return NULL is in the case `new (std::nothrow) T`, which we do not use.


eric-wieser:
Looks like these are here because we originally used `malloc` which _could_ return `NULL` (7ddcfc39a1a5613193fdbfa4c37ead0d31e74e32), but then switched to `new` (450640997a04e454fec2b6e671f2f3f35c13780d, cc @ktbarrett).

PR #1672, opened 2020-04-19
Fix include path in modelsim makefile

themperek:
Reported on gitter by @ajeshp

Leftover from #1629

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.modelsim
+++ b/cocotb/share/makefiles/simulators/Makefile.modelsim
@@ -31,5 +31,4 @@
 #ARCH:=i686
 
 # Everything else is identical to Quest
-include $(COCOTB_SHARE_DIR)/makefiles/simulators/Makefile.questa
-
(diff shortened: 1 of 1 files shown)


cocotb/share/makefiles/simulators/Makefile.modelsim
```diff
@@ -31,5 +31,4 @@
 #ARCH:=i686
 
 # Everything else is identical to Quest
-include $(COCOTB_SHARE_DIR)/makefiles/simulators/Makefile.questa
-
+include $(COCOTB_MAKEFILES_DIR)/simulators/Makefile.questa
```
eric-wieser, line 34: Where is this variable defined?

themperek: In `Makefile.sim` can also use `cocotb-config`

eric-wieser: I think invoking `cocotb-config`, or using a relative path, would be clearer.

PR #1674, opened 2020-04-19
Cleanup of modelsim and questa makefiles

themperek:
Fix intension.
Remove outdated comments.

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.questa
+++ b/cocotb/share/makefiles/simulators/Makefile.questa
@@ -40,9 +40,6 @@ endif
 
 ifeq (, $(CMD))
     $(error "Unable to locate command >$(CMD_BIN)<")
-else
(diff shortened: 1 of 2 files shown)


cocotb/share/makefiles/simulators/Makefile.questa
```diff
@@ -42,18 +42,18 @@ ifeq (, $(CMD))
     $(error "Unable to locate command >$(CMD_BIN)<")
 else
     MODELSIM_BIN_DIR := $(shell dirname $(CMD))
```
eric-wieser, line 44: Is this line used?

themperek: Not used. Was used in the past. -> Removed

eric-wieser:
Commit message has a typo, `Fix intension` -> `Fix indentation`

PR #1637, opened 2020-04-11
Update installation instructions

themperek:
Related to #1289 

Help needed!

PR #1679, opened 2020-04-19
Split installation instructions out of quickstart

cmarqu:
Refs: https://github.com/cocotb/cocotb/pull/1637#issuecomment-612684306 and https://github.com/cocotb/cocotb/pull/1637#issuecomment-612738470

To be done: (part of) https://github.com/cocotb/cocotb/pull/1637#issuecomment-613958024


documentation/source/install.rst
```diff
@@ -0,0 +1,114 @@
...
+
+.. seealso::
+
+   For more installation options, please see `our Wiki <https://github.com/cocotb/cocotb/wiki/Tier-2-Setup-Instructions>`_.
+
+
+Windows
+-------
+
+`Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for Windows.
```
themperek, line 57: Maybe `conda` should be first (possibly easiest way, also does not need any root access) for all OSes.

documentation/source/install_devel.rst
```diff
@@ -0,0 +1,17 @@
...
+
+**********************************
+Installing the Development Version
+**********************************
+
+To install the development version of cocotb, run:
+
+.. code-block:: bash
+
+    pip install https://github.com/cocotb/cocotb/archive/master.zip
```
eric-wieser, line 11: I'd probably recommend the git install, `pip install git+https://github.com/cocotb/cocotb.git` or similar.

eric-wieser: Although if users do not have git, the zip install is a nice backup option

documentation/source/quickstart.rst
```diff
@@ -1,123 +1,16 @@
+.. _quickstart:
```
eric-wieser, line 1: Should be no need for this

cmarqu: Strictly speaking yes, but I wanted these anchors to be more resilient against file renamings, something that may happen as we restructure more.

documentation/source/install_devel.rst
cmarqu: Let's go for the safe option and point to the git capabilities of pip under a Key Topics section.

documentation/source/install.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Download and install `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_ from https://conda.io/.
+From a Command Prompt or the Windows PowerShell, use the following line to install a compiler (GCC or Clang) and GNU Make:
+
+.. code-block:: bash
+
+    conda install -c msys2 m2-base m2-make m2w64-toolchain libpython
+
```
themperek, line 41:
```suggestion

.. note:: 

    In case errors related to ``Microsoft Visual Studio`` you need to set MinGW as the default compiler details can be found `here <https://stackoverflow.com/questions/3297254/how-to-use-mingws-gcc-compiler-when-installing-python-package-using-pip>`_.

```

documentation/source/install_devel.rst
```diff
@@ -4,12 +4,20 @@
 Installing the Development Version
 **********************************
 
-To install the development version of cocotb, run:
+To install the development version of cocotb on Windows, run the following:
+
+.. code-block:: bash
+
+    pip install --global-option build_ext --global-option --compiler=mingw32 https://github.com/cocotb/cocotb/archive/master.zip
```
themperek, line 11: Same for `pip install cocotb` ;-(

cmarqu: Restructured the release installation part and added it there too.

documentation/source/install.rst
```diff
@@ -0,0 +1,23 @@
+.. _install:
+
+************
+Installation
+************
+
+Using cocotb requires :ref:`installation of pre-requisites<install-pre-requisites>` and
```
ktbarrett, line 7:
```suggestion
Using cocotb requires :ref:`installation of prerequisites<install-prerequisites>` and
```

documentation/source/install_prerequisites_common.rst
```diff
@@ -0,0 +1,14 @@
+Installation of Pre-requisites
```
ktbarrett, line 1:
```suggestion
Installation of Prerequisites
```

documentation/source/quickstart.rst
```diff
@@ -1,123 +1,16 @@
...
-    After installation, you should be able to execute ``cocotb-config``.
-    If it is not found, you need to append its location to the ``PATH`` environment variable.
-    This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented :ref:`here <python:inst-alt-install-user>`.
-
 
 Running your first Example
 ==========================
 
-Assuming you have installed the prerequisites as above,
+Assuming you have the :ref:`pre-requisites<install-pre-requisites>`
```
ktbarrett, line 11:
```suggestion
Assuming you have the :ref:`prerequisites<install-prerequisites>`
```

documentation/source/install_devel.rst
```diff
@@ -0,0 +1,26 @@
...
+    pip install https://github.com/cocotb/cocotb/archive/master.zip
+
+The instructions in :ref:`installation-via-pip` also apply.
+
+For more installation options, please see `our Wiki <https://github.com/cocotb/cocotb/wiki/Tier-2-Setup-Instructions>`_.
+
+
+FIXME: add instructions for editable installation
+
+FIXME: add "You may also need to use the ``--no-use-pep517`` option to ``pip``." if...
```
ktbarrett, line 26: I added a section to the installation documentation on the Wiki for "restricted" systems that mentions `--no-use-pep517`.

documentation/source/install.rst
cmarqu: Unfortunately, since I an ``.. include::``ing the file where this anchor used to live several times for the different OSes, we cannot really link to it anymore, so I removed it now.
I changed the spelling as suggested.

documentation/source/install_devel.rst
cmarqu: Great, thanks.

documentation/source/install_cocotb_windows.rst
```diff
@@ -0,0 +1,18 @@
+Installation of cocotb
+======================
+
+.. versionadded:: 1.2
+
+The **latest release** of cocotb can be installed by running
+
+.. code-block:: bash
+
+    pip install --global-option build_ext --global-option --compiler=mingw32 cocotb
```
garmin-mjames, line 10: I don't think that the `--global-option`s are passed to `setup.py` unless you include `--no-use-pep517`.

eric-wieser:
Perhaps we should use https://github.com/djungelorm/sphinx-tabs for the different platforms

documentation/source/install_cocotb_windows.rst
cmarqu: @themperek Do you agree? I don't have Windows...

themperek: It works here: https://github.com/cocotb/cocotb/blob/master/tox.ini#L25 used on Travis  (does not work/wants use VS) in other case. So it seems it does the job.

cmarqu:
Using sphinx-tabs really makes things look much nicer. It doesn't support ```.. include::`` properly (there are rendering errors) so there is some duplicated text, but this is outweighed by the improved UI.

documentation/source/install.rst
```diff
@@ -0,0 +1,125 @@
...
+* GCC 4.8.1+ or Clang 3.3+ and associated development packages
+* GNU Make
+* A Verilog or VHDL simulator, depending on your RTL source code
+
+.. versionchanged:: 1.4 Dropped Python 2 support
+
+.. note:: In order to use a 32-bit simulator you need to use a 32-bit version of Python.
+
+The installation instructions vary depending on your operating system.
+Please choose:
```
garmin-mjames, line 30:
```suggestion
The installation instructions vary depending on your operating system:
```

documentation/source/py-modindex.rst
```diff
@@ -0,0 +1,7 @@
+..
+   This file is a placeholder and will be replaced by Sphinx
+   See https://stackoverflow.com/a/42310803
```
eric-wieser, line 3:
```suggestion
   This is a hack to put the module index in the toctree 
   See https://stackoverflow.com/a/42310803
```

documentation/source/install.rst
```diff
@@ -0,0 +1,124 @@
+.. _install:
```
eric-wieser, line 1: Nit: would marginally prefer `installation.rst`. I'll see what other projects do for this first.

cmarqu: Google says ``"source/installation.rst"`` wins with 2.1k vs. 1.6k hits.
But we also have ``install_devel.rst``, should this then be called ``installation_devel.rst`` or ``installation_development.rst``?

(While we are at it, Sphinx' default for the directory would be ``docs/source/``.)

.gitpod.Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM gitpod/workspace-full-vnc
+
```
eric-wieser, line 2: Rebase seems to have gone wrong

cmarqu: Oh. Fixed.

documentation/source/install.rst
eric-wieser: Not a big deal - although it is tempting to rename the folder to "docs". Thanks for collecting the data!

documentation/source/install.rst
```diff
@@ -0,0 +1,221 @@
...
+
+          .. container:: header
+
+              **Show/Hide Example Installation Output**
+
+          The following shows an example of the output of the above installation step.
+
+          .. code-block::
+
+             <FIXME: add installation output here>
```
eric-wieser, line 96: Not sure this is all that useful

documentation/source/_templates/page.html
```diff
@@ -0,0 +1,14 @@
+{% extends "!page.html" %}
+
+{% block footer %}
+ <script type="text/javascript">
+    $(document).ready(function() {
+        $(".toggle > *").hide();
+        $(".toggle .header").show();
+        $(".toggle .header").click(function() {
```
eric-wieser, line 8: Perhaps better would be https://pypi.org/project/sphinxcontrib-details-directive/

documentation/source/install.rst
cmarqu: This has been suggested by @imphil in https://github.com/cocotb/cocotb/pull/1637#issuecomment-613958024 and I like the idea - we don't need to show the full output if it's long, but in order to get a feel for what a successful step looks like, it's nice and reassuring for newbies.

documentation/source/install.rst
```diff
@@ -0,0 +1,221 @@
...
+
+   .. group-tab:: Linux - Debian
+
+      In a terminal, run
+
+      .. code-block:: bash
+
+          sudo apt-get install make gcc g++ python3 python3-dev python3-pip
+          sudo apt-get install swig  # optional, needed for one of the examples
```
ktbarrett, line 63:
```suggestion
```

documentation/source/install.rst
```diff
@@ -0,0 +1,221 @@
...
+             <FIXME: add installation output here>
+
+   .. group-tab:: Linux - Red Hat
+
+      In a terminal, run
+
+      .. code-block:: bash
+
+          sudo yum install make gcc gcc-c++ libstdc++-devel python3 python3-devel python3-pip
+          sudo yum install swig  # optional, needed for one of the examples
```
ktbarrett, line 84:
```suggestion
```

themperek:
It is not that this should get merged and then a correction can be done. Would it be not easier?

documentation/source/install.rst
eric-wieser: I suppose it's not harmful. In the interest of getting this merged soon though, I'd be tempted to merge without this, and add it in a follow-up.

cmarqu: Ok, removed.

documentation/source/install_cocotb_windows.rst
cmarqu: Ok, leaving this out here; the wiki has it.

documentation/source/install.rst
```diff
@@ -0,0 +1,123 @@
...
+The installation instructions vary depending on your operating system:
+
+.. tabs::
+
+   .. group-tab:: Windows
+
+      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for Windows.
+
+      Download and install `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_ from https://conda.io/.
+      From a Command Prompt or the Windows PowerShell, use the following line to install a compiler (GCC or Clang) and GNU Make:
```
themperek, line 38:
```suggestion
      From an Anaconda Prompt use the following line to install a compiler (GCC or Clang) and GNU Make:
```

cmarqu: Good point.

PR #1684, opened 2020-04-21
Prevent killed coroutines from running in corner case

ktbarrett:
From #1348. There is a corner case in the scheduling logic: if two coroutines are waiting on the same trigger and one kills the other, and if the killer is scheduled *before* the victim, the victim will still be scheduled.

This is distinct from #1349, this fixes the issue in the scheduling logic, while that PR attempts to add a feature to the scheduler.

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -434,6 +434,9 @@ class Scheduler:
                 trigger.unprime()
 
                 for coro in scheduling:
+                    if coro._outcome is not None:
+                        # coroutine was unscheduled by another coroutine waiting on the same trigger
+                        continue
                     if _debug:
                         self.log.debug("Scheduling coroutine %s" % (coro.__qualname__))
                     self.schedule(coro, trigger=trigger)
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -435,7 +435,7 @@ class Scheduler:
 
                 for coro in scheduling:
                     if coro._outcome is not None:
-                        # coroutine was unscheduled by another coroutine waiting on the same trigger
+                        # coroutine was killed by another coroutine waiting on the same trigger
                         continue
                     if _debug:
                         self.log.debug("Scheduling coroutine %s" % (coro.__qualname__))
--- a/tests/test_cases/test_cocotb/test_scheduler.py
+++ b/tests/test_cases/test_cocotb/test_scheduler.py
@@ -10,7 +10,7 @@ Test for scheduler and coroutine behavior
 """
 
 import cocotb
-from cocotb.triggers import Join, Timer, RisingEdge, Trigger, NullTrigger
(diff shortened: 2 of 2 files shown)


cocotb/scheduler.py
```diff
@@ -434,6 +434,9 @@ def _event_loop(self, trigger):
                 trigger.unprime()
 
                 for coro in scheduling:
+                    if coro._outcome:
```
eric-wieser, line 437:
```suggestion
                    if coro._outcome is not None:
```

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -178,3 +178,26 @@ def null_coroutine():
...
+async def test_kill_coroutine_waiting_on_the_same_trigger(dut):
+    # gh-1348
+    # NOTE: this test depends on scheduling priority, but the fix for gh-1348
+    # should prevent that from mattering
+    dut.clk.setimmediatevalue(0)
+
+    async def victim():
+        await Timer(1)  # prevent scheduling of RisingEdge before killer
+        await RisingEdge(dut.clk)
+        raise RuntimeError
```
eric-wieser, line 193: Possibly clearer as
```
nonlocal victim_resumed
victim_resumed = True
```

and then doing `assert not victim_resumed` below.
This means we don't have to rely on the exception-in-forked-coroutine semantics to catch the failure.

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -178,3 +178,30 @@ def null_coroutine():
         yield null_coroutine()
 
     yield Timer(100)
+
+
+@cocotb.test()
+async def test_kill_coroutine_waiting_on_the_same_trigger(dut):
+    # gh-1348
+    # NOTE: this test depends on scheduling priority, but the fix for gh-1348
+    # should prevent that from mattering
```
eric-wieser, line 187:
```suggestion
    # NOTE: this test depends on scheduling priority, assuming that the first
    # task to wait on a trigger is woken first. The fix for gh-1348 should
    # prevent that from mattering.
```

cocotb/scheduler.py
```diff
@@ -434,6 +434,9 @@ def _event_loop(self, trigger):
                 trigger.unprime()
 
                 for coro in scheduling:
+                    if coro._outcome is None:
```
eric-wieser, line 437:
```suggestion
                    if coro._outcome is not None:
```

cocotb/scheduler.py
```diff
@@ -434,6 +434,9 @@ def _event_loop(self, trigger):
                 trigger.unprime()
 
                 for coro in scheduling:
+                    if coro._outcome is not None:
+                        # coroutine was unscheduled by another coroutine waiting on the same trigger
```
eric-wieser, line 438:
```suggestion
                        # coroutine was killed by another coroutine waiting on the same trigger
```

PR #1685, opened 2020-04-21
Mention that this is the complete simulator list

cmarqu:


documentation/source/simulator_support.rst
```diff
@@ -4,7 +4,8 @@
 Simulator Support
 *****************
 
-This page documents specifics, limitations, workarounds etc. in the various simulators.
+This page lists all simulators that cocotb provides support for
+and documents specifics, limitations, workarounds etc.
```
themperek, line 9: We could add what support means. 
Maybe better "Cocotb can be used with the following simulators ..." 
Support is a big word in this case.

cmarqu: Changed, though the page title said "Support" since a very long time.

themperek: Long run we could add something that we try but there are differences and etc... Tell your simulator vendor to contribute! ;-)

PR #1692, opened 2020-04-22
Introduce SIM=riviera, keep SIM=aldec as deprecated alias

### LINKED ISSUE #1691: Rename Makefile.aldec to Makefile.riviera
Since we now also have support for Aldec Active-HDL (#1614), we should rename ``Makefile.aldec`` to ``Makefile.riviera``.
``Makefile.aldec`` should be kept around as an alias, but usage of ``SIM=aldec`` should be deprecated.

cmarqu:
Closes #1691


themperek:
Just suggestions: 
- To keep the history bit better should maybe do `git mv` on this makefile.
- One can potentially do this change form `aldec` to `riviera` in  `Makefile.sim` instead of keeping the makefile: https://github.com/cocotb/cocotb/blob/master/cocotb/share/makefiles/Makefile.sim#L115-L118

eric-wieser:
`git mv` is no different to manually moving the file. Neither informs git that the file was moved.

If git doesn't detect a move, that's because the file changed too much

eric-wieser:
Making one commit to move the file, and a second to add the shim, would solve the history issue

cmarqu:
Me trying to fix history now with git will end in me breaking things and having to start over. The Makefile shim has prior art in Modelsim/Questa, so it's internally consistent :)

eric-wieser:
Split the commit since it made it easier to rebase

eric-wieser:
@themperek, I'm a bit confused by the last two commits - are you suggesting that we don't want `Makefile.aldec` after all?

themperek:
> @themperek, I'm a bit confused by the last two commits - are you suggesting that we don't want `Makefile.aldec` after all?

That is my proposal. To be decided.

themperek:
@eric-wieser Do you think this is a bad idea?

eric-wieser:
It breaks any users who were including `Makefile.aldec` directly.

eric-wieser:
> Agreed with @cmarqu.

Not sure what you mean by that. @cmarqu  argues the files should be kept:

> The Makefile shim has prior art in Modelsim/Questa, so it's internally consistent :)

themperek:
> Not sure what you mean by that. @cmarqu argues the files should be kept:

Just that I have talked about those changes with @cmarqu.

>It breaks any users who were including Makefile.aldec directly. Maybe none of those exist.

Up to know this would quite cumbersome to do so (very unlikely).

cocotb/share/makefiles/Makefile.sim
```diff
@@ -120,6 +120,11 @@ endif
 # Maintain backwards compatibility by supporting upper and lower case SIM variable
 SIM_LOWERCASE := $(shell echo $(SIM) | tr A-Z a-z)
 
+# Use riviera instead of aldec (depricated) 
+ifeq ($(SIM_LOWERCASE),aldec)    
+    SIM_LOWERCASE := riviera
+endif
```
eric-wieser, line 126: Can we put this in `Makefile.deprecations`?

cocotb/share/makefiles/Makefile.deprecations
```diff
@@ -5,3 +5,7 @@
 ifdef VERILATOR_TRACE
     $(warning VERILATOR_TRACE is deprecated, see the "Simulator Support" section in the documentation.)
 endif
+
+ifeq ($(SIM),aldec)
```
eric-wieser, line 9:
```suggestion
ifeq ($(SIM_LOWERCASE),aldec)
```

cocotb/share/makefiles/Makefile.sim
themperek: Yes! I was considering this and it seems a better place.

eric-wieser: Note that the first commit in this patch already did exactly that :)

cocotb/share/makefiles/Makefile.deprecations
```diff
@@ -5,3 +5,7 @@
 ifdef VERILATOR_TRACE
     $(warning VERILATOR_TRACE is deprecated, see the "Simulator Support" section in the documentation.)
 endif
+
+ifeq ($(SIM_LOWERCASE),aldec)
+    $(warning Using SIM=aldec is deprecated, please use SIM=riviera instead.)
+endif
```
themperek, line 11:
```suggestion

ifeq ($(SIM_LOWERCASE),aldec)
    $(warning Using SIM=aldec is deprecated, please use SIM=riviera instead.)
    SIM_LOWERCASE := riviera
endif
```

cocotb/share/makefiles/Makefile.sim
themperek, line 126:
```suggestion
```

cocotb/share/makefiles/Makefile.sim
```diff
@@ -120,6 +120,11 @@ endif
...
+
```
themperek, line 127:
```suggestion
# Use riviera instead of aldec (depricated) 
ifeq ($(SIM_LOWERCASE),aldec)    
    SIM_LOWERCASE := riviera
endif

```

themperek, line 127:
```suggestion
```

PR #1694, opened 2020-04-22
Remove NVC without deprecation phase

### LINKED ISSUE #1693: Remove NVC support
The maintainers decided to remove the NVC "support" (it didn't ever work really) without a deprecation phase.

cmarqu:
Closes #1693


themperek:
newsfragment ?

documentation/source/newsfragments/1693.removal.rst
```diff
@@ -0,0 +1 @@
+Makefile and documentation for the NVC simulator which hasn't been working since some time have been removed.
```
themperek, line 1: Not really necessary but ...
```suggestion
Makefile and documentation for the NVC simulator which does not work (and never was) have been removed.
```

documentation/source/newsfragments/1495.removal.rst
```diff
@@ -1,5 +1,4 @@
-The makefile variables
-:make:var:`COCOTB_NVC_TRACE`, :make:var:`VERILATOR_TRACE`
-that were not supported for all simulators have been deprecated.
-Using them prints a deprecation warning and points to the documentation section
+The makefile variable :make:var:`VERILATOR_TRACE`
+that was not supported for all simulators has been deprecated.
+Using it prints a deprecation warning and points to the documentation section
```
ktbarrett, line 3: This is still pretty wordy.

```suggestion
:make:var:`VERILATOR_TRACE` has been deprecated. Use :make:var:`SIM_ARGS` to pass the `--trace` option instead.
```
Might have gotten that second sentence wrong.

We are going to have one chunky 1.4 changelog if we don't keep these more brief. The PR is linked, users can look at them if they want more detail.

documentation/source/newsfragments/1693.removal.rst
ktbarrett, line 1:
```suggestion
Support for the NVC simulator has been removed.
```

cmarqu: Done in this spirit.

documentation/source/newsfragments/1495.removal.rst
cmarqu: This newsfragment will over time collect several deprecations (#1495 is a collection ticket), so I believe the original wording will be small compared to the number of deprecated variables it will mention.

documentation/source/newsfragments/1693.removal.rst
cmarqu: I initially had it like that, but this sounds as if there ever was "support" to speak of in the first place, which apparently there wasn't. "Non-support has been  removed" would be correct :)

PR #1697, opened 2020-04-22
De-duplicate doc index entries

cmarqu:


documentation/source/library_reference.rst
```diff
@@ -139,6 +139,7 @@ Clock
 -----
 
 .. autoclass:: cocotb.clock.Clock
+    :noindex:
```
eric-wieser, line 142: Where's the other reference? A comment might be good here

cmarqu: Done.

documentation/source/library_reference.rst
```diff
@@ -141,6 +141,9 @@ Clock
 .. autoclass:: cocotb.clock.Clock
     :members:
     :member-order: bysource
+    :noindex:
+..
+   Primary occurrence in this document under "Interacting with the Simulator", so don't index this
```
eric-wieser, line 146: Might make more sense to remove one of these, documenting the same class twice on the same page seems confusing.

cmarqu: Done.

PR #1698, opened 2020-04-22
Remove explicit function signatures

cmarqu:
~~These Sphinx versions didn't show signatures for coroutines, so they had to be added by automethod.~~
Sphinx didn't show signatures for coroutines so they had to be added via automethod before we were using async functions.


eric-wieser:
Are you sure this isn't just an artifact of the move to async def? Either way, thanks for the cleanup.

cmarqu:
Oh indeed, that's much more likely the reason. Amended, thanks.

Issue #1541, opened 2020-03-27
Consider fork-like interface that doesn't schedule new coroutine immediately


garmin-mjames:
Currently when calling `fork`, the new forked coroutine gets to run immediately, and the originating coroutine waits until `fork` returns before continuing. It might be useful to provide a way to queue the forked coroutine to run after the current one yields back to the scheduler.

_Originally posted by @garmin-mjames in https://github.com/cocotb/cocotb/pull/1526#issuecomment-605092873_

eric-wieser:
That's how `asyncio.ensure_future` behaves, right?

garmin-mjames:
Yes I believe so. From a StackOverflow example:
```python3
task = asyncio.create_task(coro)
await asyncio.sleep(0) # return control to loop so task can start
```

ktbarrett:
What's the use case for this?

As mentioned in #1526, this can be implemented by adding coroutines to the scheduler using `scheduler.queue`. Which seems trivial until you realize there are other problems.

We don't really have a way to force a reschedule. `NullTrigger` fires immediately, not allowing any other coroutines to run. `Timer(0)` unnecessarily creates a GPI call and leaves behavior up to the simulator (Icarus fails if `await Timer(0)` follows an `await ReadOnly()`). So we need a way to force a reschedule.

I also brought up in my comment that because you must yield control to the scheduler for the coroutines to start, and you may forget to do this, it's possible for the test to end before the coroutine runs; resulting in unexpected or invalid behavior. We need a way to force the scheduler to run after a call to `run_soon`, but in a way the user has control over. Trio accomplishes this by using `nursery`. This would allow us to implement `run_soon` as a delayed call to `fork`, which is run when the context exits. Perhaps there are other ways of doing this.

ktbarrett:
I'll amend this "What's the use case for supporting both behaviors?" I wouldn't be against changing the behavior of `fork`, but that would probably have to wait for a major version change.

ktbarrett:
After diagnosing the root cause of #1688 and #1347, I am even more on board with the idea of `fork` not causing an immediate reschedule. Scheduler re-entry seems like a bad idea on it's surface, and obviously leads to strange bugs. So if/when we make a rewrite, we should design the scheduler around avoiding that.

garmin-mjames:
It also puts the originating Task in a weird state, where it's not pending a trigger but it's also not the running task.
[...]

Issue #637, opened 2018-08-01
coroutines that return before their first yield cause the simulator to shutdown


eric-wieser:
Repro:
```python
@cocotb.test()
def test_func_empty(dut):
    """ Test that a function can complete before the first yield """
    @cocotb.coroutine
    def func_empty():
        print("This line runs")
        return
        yield  # needed to make this a coroutine
    yield func_empty()
    print("This line is never reached")
```

eric-wieser:
A less contrived example is:

```python
@cocotb.test()
def test_func_empty(dut):
    """ Test that a function can complete before the first yield """
    @cocotb.coroutine
    def wait_if_even(x):
        print("This line runs")
        if x % 2 == 0:
            yield Timer(x)
    yield wait_if_even(2)
    print("This line also runs")
    yield wait_if_even(3)
    print("This line is never reached")
```

eric-wieser:
Note that this was auto-closed, but by a commit that said "mostly fix".

garmin-mjames:
@eric-wieser looks like #1705 does fix the rest of this.
```python3
@cocotb.test()
def test_immediate_test(dut):
    """
    Test that tests can return immediately
    """
    return
    yield
```
Passes 👍

eric-wieser:
This issue doesn't track the problem very well any more. If you looks through the tests, there are a few that still reference 637, and fail without the line the comment describes. I think they concern `function` and `external`.

eric-wieser:
One relevant test is this one:

https://github.com/cocotb/cocotb/blob/5ab94f03c179c839899badb741e7113c30a34923/tests/test_cases/test_external/test_external.py#L207-L222

ktbarrett:
@eric-wieser What's the error here? Declaring the external function before starting the clock?

eric-wieser:
The failure mode is starting an `external` at the beginning of a test, such that the first wakeup is the external finishing, and not some other trigger.

garmin-mjames:
If I remove all of the `clk_gen = cocotb.fork(Clock(dut.clk, 100).start())` lines and run the `test_external` module on #1705, everything passes.

eric-wieser:
Milestoning as 1.4 so we remember to remove all those lines and test before releasing.

garmin-mjames:
`test_ext_exit_error` passes without error when I don't skip it, so we get a FAIL on that. It needs reworking to get the desired result.

eric-wieser:
[...]

PR #1705, opened 2020-04-25
Start tests with react

garmin-mjames:
Currently tests are started with `add`. This causes problems with how
certain Triggers behave, namely `Event` and `NullTrigger`.

They behave differently at the beginning of a test than after the test
awaits a `GPITrigger`. Using `react` to start the test makes the
triggers behave the same during the entirety of the test.

Trigger behavior is different when in a `react` event loop versus a straight `schedule` called through `add`.

Found while discussing scheduler details in #1541. Thanks to @ktbarrett for `NullTrigger` test.

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -520,8 +520,8 @@ class Scheduler:
         self._write_calls.append((write_func, args))
         self._writes_pending.set()
 
-    def _coroutine_yielded(self, coro, trigger):
-        """Prime the trigger and update our internal mappings."""
+    def _resume_coro_upon(self, coro, trigger):
+        """Schedule `coro` to be resumed when `trigger` fires."""
         self._coro2trigger[coro] = trigger
 
         trigger_coros = self._trigger2coros.setdefault(trigger, [])
@@ -548,12 +548,10 @@ class Scheduler:
                 self._trigger2coros.pop(trigger)
 
                 # replace it with a new trigger that throws back the exception
-                error_trigger = NullTrigger(outcome=outcomes.Error(e))
-                self._coro2trigger[coro] = error_trigger
-                self._trigger2coros[error_trigger] = [coro]
-
-                # wake up the coroutines
-                error_trigger.prime(self.react)
+                self._resume_coro_upon(
+                    coro,
+                    NullTrigger(name="Trigger.prime() Error", outcome=outcomes.Error(e))
+                )
 
     def queue(self, coroutine):
         """Queue a coroutine for execution"""
@@ -675,8 +673,12 @@ class Scheduler:
         """Called by the regression manager to queue the next test"""
         if self._test is not None:
             raise InternalError("Test was added while another was in progress")
+
         self._test = test_coro
-        return self.add(test_coro)
+        self._resume_coro_upon(
+            test_coro,
+            NullTrigger(name="Start {!s}".format(test_coro), outcome=outcomes.Value(None))
+        )
 
     # This collection of functions parses a trigger out of the object
     # that was yielded by a coroutine, converting `list` -> `Waitable`,
@@ -784,7 +786,7 @@ class Scheduler:
                 # it wasn't allowed to yield that
                 result = NullTrigger(outcome=outcomes.Error(exc))
 
-            self._coroutine_yielded(coroutine, result)
+            self._resume_coro_upon(coroutine, result)
 
         # We do not return from here until pending threads have completed, but only
         # from the main thread, this seems like it could be problematic in cases
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -811,6 +811,7 @@ class Scheduler:
 
     def finish_test(self, exc):
         self._test.abort(exc)
+        self._check_termination()
 
     def finish_scheduler(self, exc):
         """Directly call into the regression manager and end test
--- a/documentation/source/newsfragments/1705.bugfix.rst
+++ b/documentation/source/newsfragments/1705.bugfix.rst
@@ -1,0 +1,1 @@
+Scheduling behavior is now consistent before and after the first :keyword:`await` of a :class:`~cocotb.triggers.GPITrigger`.
--- a/tests/test_cases/test_closedown/test_closedown.py
+++ b/tests/test_cases/test_closedown/test_closedown.py
@@ -27,12 +27,6 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-"""
-A set of tests that demonstrate cocotb functionality
-
-Also used a regression test of cocotb capabilities
-"""
-
 import cocotb
 from cocotb.triggers import Timer, RisingEdge
 from cocotb.clock import Clock
@@ -46,23 +40,31 @@ def test_read(dut):
         test_count += 1
 
 
-@cocotb.coroutine
-def run_external(dut):
-    yield cocotb.external(test_read)(dut)
+async def run_external(dut):
+    await cocotb.external(test_read)(dut)
 
 
-@cocotb.coroutine
-def clock_mon(dut):
-    yield RisingEdge(dut.clk)
+async def clock_mon(dut):
+    await RisingEdge(dut.clk)
 
 
 @cocotb.test(expect_fail=True,
-             expect_error=cocotb.SIM_NAME.lower().startswith(("modelsim",  # $fail_test() fails hard on Questa
+             expect_error=cocotb.SIM_NAME.lower().startswith(("modelsim",  # $fatal() fails hard on Questa
                                                               )))
-def test_failure_from_system_task(dut):
-    """Allow the dut to call $fail_test() from verilog"""
-    clock = Clock(dut.clk, 100)
-    clock.start()
+async def test_failure_from_system_task(dut):
+    """
+    Allow the dut to call system tasks from verilog.
+    $fatal() will fail the test, and scheduler will cleanup forked coroutines.
+    """
+    cocotb.fork(Clock(dut.clk, 100, units='ns').start())
     cocotb.fork(clock_mon(dut))
     cocotb.fork(run_external(dut))
-    yield Timer(10000000)
+    await Timer(10000000, units='ns')
+
+
+@cocotb.test()
+async def test_after_system_task_fail(dut):
+    """
+    Test to run after failed test.
+    """
+    await Timer(1, units='ns')
--- a/tests/test_cases/test_cocotb/test_scheduler.py
+++ b/tests/test_cases/test_cocotb/test_scheduler.py
@@ -10,7 +10,7 @@ Test for scheduler and coroutine behavior
 """
 
 import cocotb
-from cocotb.triggers import Join, Timer, RisingEdge, Trigger, NullTrigger
+from cocotb.triggers import Join, Timer, RisingEdge, Trigger, NullTrigger, Combine, Event
 from cocotb.result import TestFailure
 from cocotb.clock import Clock
 from common import clock_gen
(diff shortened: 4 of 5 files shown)


cocotb/scheduler.py
```diff
@@ -547,12 +547,18 @@ def _coroutine_yielded(self, coro, trigger):
...
+
+    def _trigger_task_with_outcome(self, task, trigger_name=None, outcome=None):
+        """Trigger task with new outcome"""
+        new_trigger = NullTrigger(name=trigger_name, outcome=outcome)
+
+        self._coro2trigger[task] = new_trigger
+        self._trigger2coros[new_trigger] = [task]
+
+        # wake up the Task
+        self.react(new_trigger)
```
eric-wieser, line 560: I'd prefer we write this in terms of `prime`, else we end up being woken by an unprimed trigger.

garmin-mjames:
There's some issue with `test_closedown`.

ktbarrett:
I'm not near my computer to try, but couldn't you just use `queue` and `react`, so you dont have to add another method?

ktbarrett:
@garmin-mjames I've tracked down the issue in `test_closedown` to the implementation of `_finished_scheduler`. I'm not sure how it even worked beforehand, that deserves some investigation. Just add `cocotb.regression_manager.handle_result(self._test)` after the test abortion and it will pass.

ktbarrett:
Perhaps this is a good time to address the issue in `test_closedown`. This point of it is to have the TB close the simulator and stop cocotb gracefully, but it contains a bunch of additional "stuff" in the test that has somehow been obscuring and implementation bug in `_finish_test`. Perhaps this test should be simplified to just enter and then wait for a long time.

eric-wieser:
I'd probably err on the side of duplicating the test and making the copy simpler, just in case the current form contains an additional corner case we forget about.

cocotb/scheduler.py
```diff
@@ -547,12 +547,18 @@ def _coroutine_yielded(self, coro, trigger):
...
+
+    def _trigger_task_with_outcome(self, task, trigger_name=None, outcome=None):
+        """Trigger task with new outcome"""
+        new_trigger = NullTrigger(name=trigger_name, outcome=outcome)
+
+        self._coro2trigger[task] = new_trigger
+        self._trigger2coros[new_trigger] = [task]
+
+        # wake up the Task
+        new_trigger.prime(self.react)
```
eric-wieser, line 560: Why not just:
```suggestion
        self._coroutine_yielded(task, new_trigger)
```
optimization doesn't seem particularly worthwhile here.

eric-wieser: Perhaps `_coroutine_yielded(coro, trigger)` should be renamed something like `_resume_coro_upon(coro, trigger)`.

ktbarrett: If he does that he could just inline the method as well.

ktbarrett:
I think the other elements there are to ensure that forked coroutines and externals are killed correctly when the test is ended via this method. If we only had one entry point for test end it would be redundant.

garmin-mjames:
Why isn't `Clock.start()` being `fork`ed in `test_closedown`?

ktbarrett:
@garmin-mjames A leftover from when clocks were implemented in C and didn't need to be forked.

garmin-mjames:
This in `_sim_event` doesn't appear to be the case, at least with icarus:
https://github.com/cocotb/cocotb/blob/beb42627bdc6d6d0271b51f2903a885706eeb5cd/cocotb/__init__.py#L198-L199

tests/test_cases/test_closedown/test_closedown.py
```diff
@@ -57,12 +51,20 @@ def clock_mon(dut):
...
-    clock = Clock(dut.clk, 100)
-    clock.start()
+    """Allow the dut to call system tasks from verilog.
+    $fatal() will fail the test, and scheduler will cleanup forked coroutines.
+    """
+    cocotb.fork(Clock(dut.clk, 100).start())
     cocotb.fork(clock_mon(dut))
     cocotb.fork(run_external(dut))
-    yield Timer(10000000)
+    yield Timer(10000000, units='ns')
```
ktbarrett, line 63: Why don't you make this `async`/`await` now. It's on our Python 2 cleanup todo list.

tests/test_cases/test_closedown/test_closedown.py
```diff
@@ -57,12 +51,20 @@ def clock_mon(dut):
...
+
+
+@cocotb.test()
+async def test_after_system_task_fail(dut):
+    """Test to run after failed test.
+    """
```
ktbarrett, line 69:
```suggestion
    """
    Test to run after failed test.
    """
```

documentation/source/newsfragments/1705.bugfix.rst
```diff
@@ -0,0 +1 @@
+Scheduling behavior in the beginning of a test, before the first ``await`` of a :class:`~cocotb.triggers.GPITrigger`, now matches the rest of the test.
```
ktbarrett, line 1:
```suggestion
Scheduling behavior is now consistent before and after the first ``await`` of a :class:`~cocotb.triggers.GPITrigger`.
```
?

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -206,3 +206,59 @@ async def killer():
...
+    Test that Event.set() doesn't cause an immediate reschedule.
+
+    The coroutine waiting with Event.wait() will be woken after
+    the current coroutine awaits a trigger.
+    """
+
+    e = Event()
+    event_set = False
+
+    @cocotb.coroutine
```
ktbarrett, line 249: Don't need this anymore.

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -206,3 +206,59 @@ async def killer():
...
+async def test_event_set_schedule(dut):
+    """
+    Test that Event.set() doesn't cause an immediate reschedule.
+
+    The coroutine waiting with Event.wait() will be woken after
+    the current coroutine awaits a trigger.
+    """
+
+    e = Event()
+    event_set = False
```
ktbarrett, line 247: `Event` has a `.fired` attribute, use that instead.

ktbarrett:
Good catch using `check_termination`, there is a bunch of cleanup that needs to be done between tests. Is there a way to move this to use `react` as well? I haven't looked at it enough. Ideally we will only have one external entry into the scheduler, so there is only one place where we need to check for test completion. That can be done in another PR if you feel this is good for now.

tests/test_cases/test_cocotb/test_scheduler.py
ktbarrett: It should probably also have a `is_set()` method to mirror asyncio.

documentation/source/newsfragments/1705.bugfix.rst
garmin-mjames: Should we add bugfix info about `SIM_TEST_FAIL` too?

garmin-mjames:
There are several paths between the Scheduler and RegressionManager, and having `add` break into the scheduler complicates it further. I think after `add`/`fork` is changed and further decoupling happens things will shake out and we can simplify further.

tests/test_cases/test_cocotb/test_scheduler.py
garmin-mjames: I actually want to check if the other coro/task wakes up. I'll fix the naming.

garmin-mjames:
I don't think that we can use `react` without changing how aborting a test works. Currently it just unschedules the test, so the scheduler has do deal with the cleanup afterwords.

There have been discussions in the past about raising an exception in the test that can be used for cleanup etc. That would match what the scheduler does for a Trigger.prime() failure or if the test awaits an invalid object. Then cleanup after the test exits (and raise another Exception if it tries to ignore the abort and await another trigger?).

documentation/source/newsfragments/1705.bugfix.rst
ktbarrett: That's a good idea, I'm not sure how to have multiple newsfrags per PR in the same category though.

ktbarrett:
@garmin-mjames If we add a `_check_termination` or some similar "check if running test is complete" method to `react`, simply `abort`ing the current test will work. But let's leave that for another PR.

documentation/source/newsfragments/1705.bugfix.rst
garmin-mjames: I could break the `finish_test` changes into another PR and get that in first, or we can just use the combined newsfragment here.

ktbarrett: [This](https://cocotb--1705.org.readthedocs.build/en/1705/release_notes.html#bugfixes) is what is currently looks like, we don't have 2 bullet points. I'll that up to @eric-wieser.

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -206,3 +206,60 @@ async def killer():
...
+        await event.wait()
+        nonlocal waiter_scheduled
+        waiter_scheduled = True
+
+    cocotb.fork(waiter(e))
+
+    e.set()
+
+    # waiter() shouldn't run until after test awaits a trigger
+    assert waiter_scheduled == False
```
ktbarrett, line 261:
```suggestion
    assert waiter_scheduled is False
```

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -206,3 +206,60 @@ async def killer():
...
+
+    await NullTrigger()
+
+    assert waiter_scheduled == True
```
ktbarrett, line 265:
```suggestion
    assert waiter_scheduled is True
```

cocotb/scheduler.py
```diff
@@ -519,7 +519,7 @@ def _schedule_write(self, handle, write_func, *args):
         self._write_calls.append((write_func, args))
         self._writes_pending.set()
 
-    def _coroutine_yielded(self, coro, trigger):
+    def _resume_coro_upon(self, coro, trigger):
         """Prime the trigger and update our internal mappings."""
```
eric-wieser, line 523:
```suggestion
        """Schedule `coro` to be resumed when `trigger` fires. """
```

documentation/source/newsfragments/1705.bugfix.rst
eric-wieser: I think you can create a second fragment, `1705.bugfix.1.rst`

eric-wieser:
I don't fully understand the `check_termination()` change, and haven't followed all the discussion here. If you haven't already posted it, can you post the logs of `test_closedown` failing before that change?

garmin-mjames:
Log of `test_closedown` using `react` to start test:
[sim.log](https://github.com/cocotb/cocotb/files/4553331/sim.log)
Log of `test_closedown` after applying the cleanup commit but not the `_check_termination` call:
[cleanup_sim.log](https://github.com/cocotb/cocotb/files/4553332/cleanup_sim.log)

What's happening:

1. `$fatal()` calls up to `finish_test`, which calls `abort` on the test
2. `abort` calls `unschedule` which calls `cleanup`
3. The result isn't handled by `RegressionManager`, so it returns to the simulator
4. Simulator has no callbacks queued up due to `cleanup`, so it ends simulation and calls `finish_scheduler`

`_check_termination` catches the fact that the test was aborted and continues the test running loop back into the `RegressionManager`, allowing further tests to run.

eric-wieser:
Release notes seem to work: https://cocotb--1705.org.readthedocs.build/en/1705/release_notes.html#bugfixes

Will come back and review those logs  in the next day or so.

eric-wieser:
I wonder if this fixes what's left with #637

garmin-mjames:
I wonder how many `yield Timer(1)`s at the beginning of tests are because of this

eric-wieser:
That, is the right :)

documentation/source/newsfragments/1705.bugfix.1.rst
```diff
@@ -0,0 +1 @@
+Scheduler now correctly cleans up when a test is failed using the ``$fatal()`` System Task in Verilog.
```
cmarqu, line 1:
```suggestion
The scheduler now correctly cleans up when a test is failed using the ``$fatal()`` System Task in Verilog.
```

eric-wieser: Is this actually a bug that was present in 1.3? Or just a bug that was introduced by the commit before it?

eric-wieser: The latter should not be in a release note.

garmin-mjames: @eric-wieser I guess that's true, it worked before. I'll remove the newsfragment.

garmin-mjames:
Rebased on master.

eric-wieser:
Thanks for tracking down what I could not!

Would appreciate if you could follow up with a PR to remove all the gh-637 hacks (should be able to search for `gh-637` or `#637`)

PR #1706, opened 2020-04-25
Use conda to build documentation

themperek:
Conda is built with `--enable-shared` by default.
Closes https://github.com/cocotb/cocotb/issues/1569
In favor of https://github.com/cocotb/cocotb/pull/1573

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,10 +4,14 @@
 
 version: 2
 
+# Use conda because default Python is not compiled with --enable-shared
+conda:
+  environment: documentation/conda.yml
+
 python:
-  version: 3.7
   install:
-    - requirements: documentation/requirements.txt
+    - method: pip
+      path: .
 
 sphinx:
   configuration: documentation/source/conf.py
--- a/documentation/conda.yml
+++ b/documentation/conda.yml
@@ -4,11 +4,6 @@ name: cocotb-docs3
 dependencies:
   - python=3.8
   - pip
-  # ReadTheDoc dependencies
-  - mock
-  - pillow
-  - sphinx
-  - sphinx_rtd_theme
   # Packages installed from PyPI
   - pip:
       - -r requirements.txt
--- a/documentation/conda.yml
+++ b/documentation/conda.yml
@@ -1,0 +1,14 @@
+# Python environment and dependencies to build the documentation
+
+name: cocotb-docs3
+dependencies:
+  - python=3.8
+  - pip
+  # ReadTheDoc dependencies
+  - mock
+  - pillow
+  - sphinx
+  - sphinx_rtd_theme
+  # Packages installed from PyPI
+  - pip:
+      - -r requirements.txt
--- a/documentation/conda.yml
+++ b/documentation/conda.yml
@@ -1,6 +1,6 @@
 # Python environment and dependencies to build the documentation
 
-name: cocotb-docs3
+name: cocotb-docs
 dependencies:
   - python=3.8
   - pip
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -12,11 +12,6 @@ import os
 import subprocess
 import sys
 
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
-# documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('../..'))
-
 # Add in-tree extensions to path
 sys.path.insert(0, os.path.abspath('../sphinxext'))
 


documentation/conda.yml
```diff
@@ -0,0 +1,21 @@
...
+  - python=3.8
+  - pip
+# ReadTheDoc dependencies
+  - mock
+  - pillow
+  - sphinx
+  - sphinx_rtd_theme
+# Packages installed from PyPI
+  - pip:
+    - breathe
```
eric-wieser, line 14: Can you leave the requirements.txt alone, and use
```
    - -r requirements.txt
``
here? (https://stackoverflow.com/a/59056234/102441)

File: .readthedocs.yml [hunk 2]
```diff
@@ -4,10 +4,13 @@
 
 version: 2
 
+conda:
+  environment: documentation/conda.yml
```
eric-wieser, line 8: Please add a comment indicating _why_ we use conda in this file, mentioning `--enable-shared`

documentation/requirements.txt
```diff
@@ -1,8 +1,9 @@
-# Python dependencies to build the documentation
+# Python (PyPI) dependencies to build the documentation
 
-Sphinx ~= 3.0
+# Packages installed from conda
+# Sphinx ~= 3.0
```
eric-wieser, line 4: What happens if you leave these here, and remove them from the conda config?

themperek:
Than you get this https://github.com/readthedocs/readthedocs.org/issues/6870

eric-wieser:
How about if you do both? My goal here is to keep the requirements working, so non-conda users can still build the docs locally.

documentation/conda.yml
```diff
@@ -0,0 +1,14 @@
+# Python environment and dependencies to build the documentation
+
+name: cocotb-docs
+dependencies:
+  - python=3.8
+  - pip
+# ReadTheDoc dependencies
+  - mock
+  - pillow
+  - sphinx
```
eric-wieser, line 10: Can we pin this to `~= 3` like we do for pip?

eric-wieser, line 10: Can we pin this to `~= 3` like we do for pip?

documentation/conda.yml
```diff
@@ -0,0 +1,14 @@
+# Python environment and dependencies to build the documentation
+
+name: cocotb-docs
+dependencies:
+  - python=3.8
+  - pip
+# ReadTheDoc dependencies
```
eric-wieser, line 7:
```suggestion
  # ReadTheDoc dependencies
```

documentation/conda.yml
```diff
@@ -0,0 +1,14 @@
...
+  - mock
+  - pillow
+  - sphinx
+  - sphinx_rtd_theme
+# Packages installed from PyPI
```
eric-wieser, line 12:
```suggestion
  # Other dependencies managed by pip
```

documentation/conda.yml
themperek: With conda this syntax does not work.

eric-wieser:
```suggestion
  - sphinx >=3.0,<4
```

.readthedocs.yml
```diff
@@ -4,10 +4,14 @@
 
 version: 2
 
+# Use cond a because default Python is not compiled with --enable-shared
```
eric-wieser, line 7:
```suggestion
# Use conda because default Python is not compiled with --enable-shared
```

documentation/conda.yml
themperek: This also does not work.

documentation/requirements.txt
themperek: Then you get this readthedocs/readthedocs.org#6870

documentation/conda.yml
themperek: But it looks pip will take over the requirements in any case. Conda sphinx version is 2.4.4 and if I use this conda.yml I am getting 3.0.2 from pypi (same for ReadTheDoc)

eric-wieser: And yet it fails if you omit it entirely?

documentation/requirements.txt
```diff
@@ -1,8 +1,8 @@
 # Python dependencies to build the documentation
 
 Sphinx ~= 3.0
-breathe
 sphinx-rtd-theme
+breathe
```
eric-wieser, line 5: Nit: can we leave this file alone? I assume this order makes no difference

documentation/conda.yml
themperek: >And yet it fails if you omit it entirely?
Yes, It looks like some dependency issue.

eric-wieser:
https://github.com/readthedocs/readthedocs.org/issues/3829#issuecomment-515765278 seems relevant.

The failure mode looks like some bad interaction between conda and pip that leads to a Sphinx installation that is a mix of two versions.

themperek:
Looks like that is all I can do.

eric-wieser:
Let's ask rtd to turn that feature flag on so that we can at least pin sphinx

themperek:
@eric-wieser Is this really so important?

eric-wieser:
Yes, we were broken very recently by a Sphinx update, and the only workaround was go pin it. With this patch, we wouldn't even be able to do that, and our docs would be unfixable.

themperek:
>With this patch, we wouldn't even be able to do that, and our docs would be unfixable.
This does not look to be the case. You can pin version in requirements.txt and it is taken.

eric-wieser:
Unfortunately you get a weird mix of two installations, which causes the issue that you originally linked to if the versions don't match (in a comment you seem to have deleted).

themperek:
The comment is here: https://github.com/cocotb/cocotb/pull/1706#discussion_r415113425

themperek:
I am just not sure if some special flags set by RTD is much better.

documentation/conda.yml
themperek: Unfortunately looks like conda in the yml file supports only "=" so yu can pin particular version only.

eric-wieser: Claims to be supported here: https://stackoverflow.com/a/54849994/102441

eric-wieser: What's the failure look like with the syntax I suggested above?

themperek: So this is maybe that the latest version in conda is 2.4.4 and it fails. I could have missed this.

themperek:
```
ResolvePackageNotFound:
  - sphinx[version='>=3.0,<4']
```

eric-wieser: Was that all the error there was?

eric-wieser: You're right though, it does seem conda is on an older version

themperek: If one skip sphinx one gets this: https://readthedocs.org/projects/cocotb/builds/10910139/

It may be due to the order.  They do this after the user environment:
`conda install --yes --quiet --name 1706 mock pillow sphinx sphinx_rtd_theme`
Why not before?

themperek: In any case ...

eric-wieser: > Why not before?

That's what the issue I link to in https://github.com/cocotb/cocotb/pull/1706#issuecomment-619428557 is all about. The flag is the fix for that problem, but they don't want to make it the default in  case it breaks people.

themperek: Yes, but appending will make no difference practically? Just will not see this in our file.

themperek:
So if I am not mistaken what will happen is with `CONDA_APPEND_CORE_REQUIREMENTS` will append to our `conda.yml` those 4 packages `mock pillow sphinx sphinx_rtd_theme`. It is the same if we add them without the flag. I can ask for the flag any is fine for me.

themperek:
I do not have RTD admin right to cocotb probably someone who has should ask for the flag.

eric-wieser:
> It is the same if we add them without the flag.

The key difference is that with the flag, `conda` is _not_ invoked a second time, which it currently is. On that second invocation, our pinned versions are currently ignored.

themperek:
>On that second invocation, our pinned versions are currently ignored.

Not sure how this works but on the second invocation it does nothing at the moment:
```
# All requested packages already installed.
```
form https://readthedocs.org/projects/cocotb/builds/10909922/

eric-wieser:
> on the second invocation it does nothing at the moment:

Right, that's because they're not pinned. If they were pinned, the second invocation would replace the pinned versions with the latest versions.

eric-wieser:
Enabled the flag, let's see what happens

documentation/conda.yml
```diff
@@ -0,0 +1,14 @@
...
+
+name: cocotb-docs3
+dependencies:
+  - python=3.8
+  - pip
+  # ReadTheDoc dependencies
+  - mock
+  - pillow
+  - sphinx
+  - sphinx_rtd_theme
```
themperek, line 11:
```suggestion
  # ReadTheDoc dependencies
  - mock
  - pillow
  - sphinx
  - sphinx_rtd_theme
```

themperek: @eric-wieser Should we try this?

eric-wieser: Am I blind, or is your suggestion a no-op?

themperek: Now better?

documentation/conda.yml
```diff
@@ -0,0 +1,9 @@
+# Python environment and dependencies to build the documentation
+
+name: cocotb-docs3
```
themperek, line 3:
```suggestion
name: cocotb-docs
```

eric-wieser:
Can you add this change too?

https://github.com/cocotb/cocotb/pull/1573/files#diff-a37c48e6a7da66ba8fe8772d938370da

Without it, we end up with a mix of installed and non-installed cocotb being used.

themperek:
@eric-wieser Thank you and sorry for the troubles. This was much worse than expected ... 😭

Issue #1566, opened 2020-04-04
Remove COCOTB_SIM environmental variable


themperek:
Since https://github.com/cocotb/cocotb/pull/1425 it should not be needed

ktbarrett:
I am fine with removing this as an environment variable, however I'd like some way for a user to know if they operating in a valid cocotb context or not. If `simulator` can be loaded unconditionally, perhaps we can put a function there to disambiguate (`running` property)? Or maybe the function that I will soon add that gets the product and version strings could return "nosim" if no simulator is running?

The reason I ask is because I am trying to write a compatibility layer for concurrency primitives so we can run our models for cocotb tests in python-only contexts using asyncio. I need to know what context the library is running in to know whether to source cocotb or asyncio implementations.

eric-wieser:
As an interim solution, could we remove this today, and change `_initialize_testbench` to do the following?
```
IN_SIMULATOR = false

def _initialize_testbench(...):
    nonlocal IN_SIMULATOR
    IN_SIMULATOR = true
```

themperek:
I am for any solution.

Maybe add `get_sim_name()` to siumlator module? Retun `None` if none?

themperek:
So if one does:

``python3
SIM_NAME = None

def _initialize_testbench(...):
..
```

Should be able to check for `cocotb.SIM_NAME` if I am not mistaken?

eric-wieser:
My suggestion above was an attempt to continue avoiding importing the simulator module, without the need for an env var.

We can't solve that problem from the simulator until we solve #1568 (or re-introduced `SPHINX_BUILD` as a variable)

themperek:
@eric-wieser We cannot import `cocotb.simulator` now? I just cannot call some functions but able to import.

eric-wieser:
That only works on systems with `--enabled-shared`. Ideally we'd lift that restriction by reversing the dependency between `libgpi` and `libcocotb`.

ktbarrett:
@eric-wieser Funnily enough, I have that change hanging out as well. I was hoping to take a swing at refactoring the scheduler before that, but I can move up those changes first so they can be done over the next few weeks.

themperek:
@eric-wieser 
>That only works on systems with --enabled-shared. Ideally we'd lift that restriction by reversing the >dependency between libgpi and libcocotb.

Any case without `--enabled-shared` setup will not work and something will fail since we need to link with python?

eric-wieser:
[...]

PR #1713, opened 2020-04-26
Remove COCOTB_SIM from makefiles

themperek:
Set `COCOTB_SIM` in GPI

Addressing #1566 

This is a temporary suggestion.
@ktbarrett should have a better one.


cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -208,6 +208,10 @@ void gpi_load_extra_libs()
     /* Lets look at what other libs we were asked to load too */
     char *lib_env = getenv("GPI_EXTRA");
 
+    /* Allow recognize if we are in simulation */
+    char cooctb_sim_env[] = "COCOTB_SIM=1";
+    putenv(cooctb_sim_env);
```
eric-wieser, line 213: `putenv` looks like a bad idea: https://stackoverflow.com/a/5876818/102441.

Can you use setenv instead?

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -208,6 +208,10 @@ void gpi_load_extra_libs()
     /* Lets look at what other libs we were asked to load too */
     char *lib_env = getenv("GPI_EXTRA");
 
+    /* Allow recognize if we are in simulation */
```
eric-wieser, line 211:
```suggestion
    /* inform python that we are in simulation */
```

eric-wieser:
This looks like a good start, definitely a win getting it out of the makefile. Just the comment above to address.

cocotb/share/lib/gpi/GpiCommon.cpp
themperek: There seems to be an issue on mingw (https://lists.gnu.org/archive/html/octave-maintainers/2012-09/msg00245.html) . I have `error: 'setenv' was not declared in this scope`. (I have aded `#include <stdlib.h>`)

Mabe this way (before import)?:
```c
PyRun_SimpleString("import os;os.environ['COCOTB_SIM']='1'");
```

themperek: More: https://stackoverflow.com/questions/17258029/c-setenv-undefined-identifier-in-visual-studio

eric-wieser: Can you use the `_s` variant?

themperek: I can but `_putenv_s`  does not seem to exist on Linux (https://travis-ci.org/github/themperek/cocotb/jobs/679792338#L1069) . Can do `ifdef` if this is better?

cmarqu, line 213:
```suggestion
    char cocotb_sim_env[] = "COCOTB_SIM=1";
    putenv(cocotb_sim_env);
```

themperek: So I see 3 options:
1. `putenv` in this our should be safe? but ...
2. `_putenv_s` on Windows and `setenv` on rest if `ifdef`
3. `PyRun_SimpleString("import os;os.environ['COCOTB_SIM']='1'");`

?

eric-wieser: 1 is ok only if you make the `char[]` static, else you risk a use after free.
2 sounds possibly safest.
3 isn't an option because we haven't embedded python yet.

eric-wieser: 1 is ok only if you make the `char[]` static, else you risk a use after free.
2 sounds possibly safest.
3 isn't an option because we haven't embedded python yet.

themperek: ad.3 We can do the same before (seems o work): https://github.com/cocotb/cocotb/blob/a797fcb086f032eeb12b05172ad94482a54b729f/cocotb/share/lib/embed/gpi_embed.cpp#L259-L261

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -208,6 +208,10 @@ void gpi_load_extra_libs()
     /* Lets look at what other libs we were asked to load too */
     char *lib_env = getenv("GPI_EXTRA");
 
+    /* inform python that we are in simulation */
```
eric-wieser, line 211:
```suggestion
    /* inform python that we are in simulatio
     * TODO[gh-1566]: Eliminate the need for this completely. */
```

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -208,6 +208,10 @@ void gpi_load_extra_libs()
...
+    char cocotb_sim_env[] = "COCOTB_SIM=1";
+    putenv(cocotb_sim_env);
```
eric-wieser, line 213:
```suggestion
    static char cocotb_sim_env[] = "COCOTB_SIM=1";
    putenv(cocotb_sim_env);  // putenv requires the array lifetime to be as long as the program lifetime, hence the static

eric-wieser:
CI was cancelled, just restarted it. I think it makes sense to put this in, even if @ktbarrett has a better solution - it moves hacks from 12 files to 1 file, which will make the follow-up patch easier to review.

Issue #1387, opened 2020-02-16
Use `build_clib` instead of `build_ext` for non python extension libraries 


themperek:
From https://github.com/cocotb/cocotb/pull/1306#pullrequestreview-357334578

>Building normal libraries through the setuptools mechanisms for Python extension modules is highly confusing (as if the library situation in cocotb wasn't already confusing enough) and leads to ugly code, e.g. rewriting the file name in `build_ext`. Instead, use `build_clib` and/or a custom command in `setup.py`. As usual in Python packaging, it's mostly undocumented and a mess, but I've seen examples around that look reasonable.

themperek:
See also  #955

themperek:
@eric-wieser This seems to be not so easy taking account for build_clib state.  Can we close?

eric-wieser:
I found a stackoverflow post a couple days ago that might provide a path forward for this. I think we should leave it open.

eric-wieser:
xref #1739, which uses https://github.com/mdavidsaver/setuptools_dso instead

PR #1714, opened 2020-04-26
Fixes for libs on MacOS

mtikekar:
- $ORIGIN changed to @loader_path
- libpython's extension is .dylib, no .so
- add install_names to vpi and vhpi extensions
- add libbpython's path to rpath

### DIFF
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -282,8 +282,9 @@ def _get_vpi_lib_ext(
 def _get_vhpi_lib_ext(
     include_dir, share_lib_dir, sim_define, extra_lib=[], extra_lib_dir=[]
 ):
+    lib_name = "libcocotbvhpi_" + sim_define.lower()
     libcocotbvhpi = Extension(
-        os.path.join("cocotb", "libs", "libcocotbvhpi_" + sim_define.lower()),
+        os.path.join("cocotb", "libs", lib_name),
         include_dirs=[include_dir],
         define_macros=[("VHPI_CHECKING", 1)] + [(sim_define, "")],
         libraries=["gpi", "gpilog", "stdc++"] + extra_lib,
@@ -292,7 +293,7 @@ def _get_vhpi_lib_ext(
             os.path.join(share_lib_dir, "vhpi", "VhpiImpl.cpp"),
             os.path.join(share_lib_dir, "vhpi", "VhpiCbHdl.cpp"),
         ],
-        extra_link_args=_extra_link_args(rpath="$ORIGIN"),
+        extra_link_args=_extra_link_args(lib_name=lib_name, rpath="$ORIGIN"),
         extra_compile_args=_extra_cxx_compile_args,
     )
 
@@ -374,7 +375,7 @@ def get_ext():
                     os.path.join(share_lib_dir, "fli", "FliCbHdl.cpp"),
                     os.path.join(share_lib_dir, "fli", "FliObjHdl.cpp"),
                 ],
-                extra_link_args=_extra_link_args(rpath="$ORIGIN"),
+                extra_link_args=_extra_link_args(lib_name="libcocotbfli_modelsim", rpath="$ORIGIN"),
                 extra_compile_args=_extra_cxx_compile_args,
             )
 
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -362,8 +362,9 @@ def get_ext():
         modelsim_include_dir = os.path.join(modelsim_dir, "include")
         mti_path = os.path.join(modelsim_include_dir, "mti.h")
         if os.path.isfile(mti_path):
+            lib_name = "libcocotbfli_modelsim"
             fli_ext = Extension(
-                os.path.join("cocotb", "libs", "libcocotbfli_modelsim"),
+                os.path.join("cocotb", "libs", lib_name),
                 include_dirs=[include_dir, modelsim_include_dir],
                 libraries=["gpi", "gpilog", "stdc++"] + modelsim_extra_lib,
                 library_dirs=modelsim_extra_lib_path,
@@ -372,7 +373,7 @@ def get_ext():
                     os.path.join(share_lib_dir, "fli", "FliCbHdl.cpp"),
                     os.path.join(share_lib_dir, "fli", "FliObjHdl.cpp"),
                 ],
-                extra_link_args=_extra_link_args(lib_name="libcocotbfli_modelsim", rpaths=["$ORIGIN"]),
+                extra_link_args=_extra_link_args(lib_name=lib_name, rpaths=["$ORIGIN"]),
                 extra_compile_args=_extra_cxx_compile_args,
             )
 
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -38,22 +38,6 @@ class build_ext(_build_ext):
 
         super().run()
 
-    # On osx need extra extra rpath with install_name_tool to find libraries
-    if sys.platform == "darwin":
-        def build_extension(self, ext):
-            super().build_extension(ext)
-
-            subprocess.run([
-                "install_name_tool",
-                "-add_rpath",
-                "@loader_path",
-                "-add_rpath",
-                "@loader_path/libs",
-                "-add_rpath",
-                sysconfig.get_config_var("LIBDIR"),
-                os.path.join(self.build_lib, ext._file_name),
-            ], check=True)
-
     # Needed for Windows to not assume python module (generate interface in def file)
     def get_export_symbols(self, ext):
         return None
@@ -146,6 +130,8 @@ def _extra_link_args(lib_name=None, rpath=None):
     if sys.platform == "darwin" and lib_name is not None:
         args += ["-Wl,-install_name,@rpath/%s.so" % lib_name]
     if rpath is not None:
+        if sys.platform == "darwin":
+            rpath = rpath.replace("$ORIGIN", "@loader_path")
         args += ["-Wl,-rpath,%s" % rpath]
     return args
 
@@ -226,6 +212,9 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
     #
     #  libcocotb
     #
+    link_args = _extra_link_args(lib_name="libcocotb", rpath="$ORIGIN")
+    for lib_dir in python_lib_dirs:
+        link_args += _extra_link_args(rpath=lib_dir)
     libcocotb = Extension(
         os.path.join("cocotb", "libs", "libcocotb"),
         define_macros=[("PYTHON_SO_LIB", _get_python_lib())],
@@ -233,7 +222,7 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
         libraries=[_get_python_lib_link(), "gpilog", "cocotbutils"],
         library_dirs=python_lib_dirs,
         sources=[os.path.join(share_lib_dir, "embed", "gpi_embed.cpp")],
-        extra_link_args=_extra_link_args(lib_name="libcocotb", rpath="$ORIGIN"),
+        extra_link_args=link_args,
         extra_compile_args=_extra_cxx_compile_args,
     )
 
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -47,12 +47,10 @@ class build_ext(_build_ext):
                 "install_name_tool",
                 "-add_rpath",
                 "@loader_path",
-                os.path.join(self.build_lib, ext._file_name),
-            ], check=True)
-            subprocess.run([
-                "install_name_tool",
                 "-add_rpath",
                 "@loader_path/libs",
+                "-add_rpath",
+                sysconfig.get_config_var("LIBDIR"),
                 os.path.join(self.build_lib, ext._file_name),
             ], check=True)
 
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -119,7 +119,7 @@ class build_ext(_build_ext):
                 )
 
 
-def _extra_link_args(lib_name=None, rpath=None):
+def _extra_link_args(lib_name=None, rpaths=[]):
     """
     Add linker argument to load dependencies from the directory where vpi/vhpi/fli library is located
     On osx use `install_name`.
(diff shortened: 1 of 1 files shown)


themperek:
What error/issue does this solve?

mtikekar:
I could not use cocotb at all without all these fixes. The first error is when running `make SIM=icarus` or `make SIM=verilator` -

```
dyld: Library not loaded: @rpath/libpython3.7m.dylib
  Referenced from: .../lib/python3.7/site-packages/cocotb/libs/libcocotb.so
  Reason: image not found
```

This is fixed by adding `sys.prefix + "/lib"` to rpath.

The next error was in cocotb/share/lib/embed/gpi_embed.cpp: "Failed to find python shared library" which was due to `-DPYTHON_SO_LIB=libpython3.7m.so` instead of `libpython3.7m.dylib`.

The third error was that the simulator was able to link to libcocotbvpi_verilator.so but could not run because the install_name was not set on the .so file. I'm not sure if other simulators need it, but it shouldn't hurt.

The change from `$ORIGIN` to `@loader_path` is just cleanup. `$ORIGIN` is not understood on MacOS.

I don't think these changes should affect Windows or Linux, but I will be checking on Linux soon.

cocotb/_build_libs.py
```diff
@@ -174,6 +170,8 @@ def _get_python_lib():
 
     if os.name == "nt":
         python_lib = _get_python_lib_link() + "." + _get_lib_ext_name()
+    elif sys.platform == "darwin":
+        python_lib = "lib" + _get_python_lib_link() + ".dylib"
```
cmarqu, line 174: You could just extend ``_get_lib_ext_name()`` instead, right?

themperek:
@mtikekar So there are few issues here. This would be sperate PRs or minimum sperate commits with message that include information what is the issue this commit solves.

themperek:
Aby idea why Travis does not run for this PR?

cocotb/_build_libs.py
mtikekar: Yes, I could do that. But I found a few places where the default extension for libraries was assumed to be .so (Linux, Mac) or .dll (Windows). I wasn't sure I caught them all.

mtikekar:
I can do separate commits on the same PR.

mtikekar:
Travis says "abuse detected" for this and my earlier PR. Some discussion [here](https://travis-ci.community/t/some-pr-are-not-analyzed-with-message-abuse-detected/1191/7) and [here](https://travis-ci.community/t/abuse-detected-for-my-prs-and-some-commits/6124/3).

cocotb/_build_libs.py
```diff
@@ -46,13 +46,7 @@ def build_extension(self, ext):
...
                 "install_name_tool",
                 "-add_rpath",
-                "@loader_path",
-                os.path.join(self.build_lib, ext._file_name),
-            ], check=True)
-            subprocess.run([
-                "install_name_tool",
-                "-add_rpath",
-                "@loader_path/libs",
+                sys.prefix + "/lib",
```
themperek, line 49: Should not be better?
```suggestion
                sysconfig.get_config_var("LIBDIR"),
```

themperek:
Interesting is the fact that it seems to be possible to do `-Wl,-install_name,@rpath/abc.so` and  `-Wl,-rpath,@loader_path` same time. I have tried different combinations to make it work but not this one (I do not use and do not have Mac).

mtikekar:
On Mac, the `install_name` of a library is used to resolve the library at load time. For example, if libA has install_name of `/path/to/libA.dylib` and libB is linked with `-lA`, libB will list `/path/to/libA.dylib` as its dependency (unlike Linux which will list `libA`).

The concept of `rpath` is the same on Linux and Mac, except that instead of `$ORIGIN`, you have to use `@load_path` or `@executable_path`as values for `rpath`.

If we set `-Wl,-rpath` correctly, we can remove the `install_name_tool` invocation.

cocotb/_build_libs.py
```diff
@@ -148,6 +142,8 @@ def _extra_link_args(lib_name=None, rpath=None):
     if sys.platform == "darwin" and lib_name is not None:
         args += ["-Wl,-install_name,@rpath/%s.so" % lib_name]
```
themperek, line 143: So based on https://github.com/cocotb/cocotb/pull/1714#issuecomment-619625219 we can remove those also?

cocotb/_build_libs.py
themperek: Possibly we can add this only on `libcocotb` and remove this here?

cocotb/_build_libs.py
mtikekar: We need to keep it. The default `install_name` is the path where the library got written by the compiler which will be some temporary build directory that pip uses.

cocotb/_build_libs.py
mtikekar: Yes, I am familiar with `sysconfig`.

mtikekar:
Reading through the build_libs file, I realized that a lot of the code is for building libraries that are not Python extensions. I was wondering if this will help: https://github.com/mdavidsaver/setuptools_dso

themperek:
Another possibility is #1387 but any solution needs some time to debug typically

eric-wieser:
@mtikekar, that looks very useful, thanks. I'd be keen to avoid maintaining our own code to build shared libraries. If you have time to attempt to put a patch together to use that, it would be appreciated!

mtikekar:
@themperek, I have pushed the smallest changes that get things working on my Mac setup. Hopefully, these can be merged without issues. 

In the first, I added libpython's path to rpath for libcocotb to find it. On Travis, you're using the Homebrew python whose libpython can be found in the hardcoded location set in its install_name (something like `/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/Python`). I am using conda python whose install name is `@rpath/libpython3.7.dylib`, so setting an rpath is necessary.

In the second commit, I added install_name to the vpi libraries so that verilator can work. Other simulators load the vpi library at runtime but verilator links it into the simulation executable. Without the right install_name, the executable looks for `build/lib.macosx-10.9-x86_64-3.7/cocotb/libs/libcocotbvpi_verilator.so` which it cannot find.

Finally, I cleaned up rpath at link-time so that `install_name_tool` is not needed.

cocotb_build_libs.py
```diff
@@ -284,7 +272,7 @@ def _get_vpi_lib_ext(
             os.path.join(share_lib_dir, "vpi", "VpiImpl.cpp"),
             os.path.join(share_lib_dir, "vpi", "VpiCbHdl.cpp"),
         ],
-        extra_link_args=_extra_link_args(rpath="$ORIGIN"),
+        extra_link_args=_extra_link_args(lib_name=lib_name, rpath="$ORIGIN"),
```
themperek, line 275: Should we do the same for VHPI and FLI? Even if there is not simulator that supports this on Mac?

mtikekar: Sure, I can make them all consistent.

elgorwi:
> For me it looks like a good change:
> 
>     * fixes issues on different osx setups
> 
>     * less hacky build_ext
> 
> 
> Let's wait for a few days. It would be good if @eric-wieser had a look too.
> 
> Should cleanup/squash commits.

Just tried this on my mac and installing now works without issues. 
 @eric-wieser @themperek Would be great to get this merged

mtikekar:
Do you need this squashed to a single commit?

eric-wieser:
We can squash to a single commit through the github UI. If the commits are reasonably well segregated, then we tend to rebase-merge.

cocotb_build_libs.py
```diff
@@ -47,12 +47,10 @@ def build_extension(self, ext):
...
                 "-add_rpath",
                 "@loader_path",
-                os.path.join(self.build_lib, ext._file_name),
-            ], check=True)
-            subprocess.run([
-                "install_name_tool",
                 "-add_rpath",
                 "@loader_path/libs",
+                "-add_rpath",
+                sysconfig.get_config_var("LIBDIR"),
```
eric-wieser: I assume you combined these into a single invocation because it was tidier, not because it was necessary?

Would be great if this commit message could mention that.

eric-wieser: Nevermind, I see this gets removed in a later commit anyway - guess we should squash this PR when we merge.

cocotb_build_libs.py
```diff
@@ -228,14 +212,17 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
     #
     #  libcocotb
     #
+    link_args = _extra_link_args(lib_name="libcocotb", rpath="$ORIGIN")
+    for lib_dir in python_lib_dirs:
+        link_args += _extra_link_args(rpath=lib_dir)
```
eric-wieser, line 217: Perhaps instead the `rpath` argument to `_extra_link_args` should be a list, making this:

```suggestion
    link_args = _extra_link_args(lib_name="libcocotb", rpath=["$ORIGIN"] + python_lib_dirs)
    for lib_dir in python_lib_dirs:
        link_args += _extra_link_args(rpath=lib_dir)
```

cocotb_build_libs.py
```diff
@@ -386,7 +375,7 @@ def get_ext():
                     os.path.join(share_lib_dir, "fli", "FliCbHdl.cpp"),
                     os.path.join(share_lib_dir, "fli", "FliObjHdl.cpp"),
                 ],
-                extra_link_args=_extra_link_args(rpath="$ORIGIN"),
+                extra_link_args=_extra_link_args(lib_name="libcocotbfli_modelsim", rpath="$ORIGIN"),
```
eric-wieser, line 378: Can we create a `lib_name` variable for consistency here too?

cocotb_build_libs.py
mtikekar: Good idea. Changed `rpath` to `rpaths` list.

cocotb_build_libs.py
mtikekar: Done

mtikekar:
Did the suggested cleanup.

I think some more changes are needed to make a wheel out of this to put on pypi. The problem is with how `simulator` and `libcocotb` find `libpython`. Right now, on Linux and Mac, they're given a `rpath` to the location of `libpython` that was used in the build environment.

For `simulator`, I was looking at how `numpy` does it. And their extension libraries don't even link to `libpython` (on both Linux and Mac). On Mac, I think it works because of an `-undefined dynamic_lookup` link option which allows symbols from `libpython` to be found at load time. And at load time, the `python` interpreter has already loaded `libpython` before it loads the extensions, so the extensions get their symbols that way. On Linux, something similar might be at work. 

This is just my guess, by the way. It will need some more investigation. I could be completely wrong. In any case, the above mechanism will not work for `libcocotb` as that is not a python extension but a library that embeds python.

cocotb_build_libs.py
```diff
@@ -137,7 +119,7 @@ def _gen_import_libs(self, def_dir):
                 )
 
 
-def _extra_link_args(lib_name=None, rpath=None):
+def _extra_link_args(lib_name=None, rpaths=[]):
```
themperek, line 122:
```suggestion
def _extra_link_args(lib_name=None, rpaths=None):
```
I guess we should avoid `[]` here see: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

mtikekar:
I'm aware of that - it only matters if you mutate the value in the function
which we are not doing. But, to be safe, we can make it an empty tuple.

>
>
>

cocotb_build_libs.py
themperek: You are right doe not really apply here.

cocotb_build_libs.py
```diff
@@ -235,7 +219,7 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
         libraries=[_get_python_lib_link(), "gpilog", "cocotbutils"],
         library_dirs=python_lib_dirs,
         sources=[os.path.join(share_lib_dir, "embed", "gpi_embed.cpp")],
-        extra_link_args=_extra_link_args(lib_name="libcocotb", rpath="$ORIGIN"),
+        extra_link_args=_extra_link_args(lib_name="libcocotb", rpaths=["$ORIGIN"] + python_lib_dirs),
```
themperek, line 222: So we have to add `python_lib_dirs` here because you get an error 
```
dyld: Library not loaded: @rpath/libpython3.7m.dylib
  Referenced from: .../lib/python3.7/site-packages/cocotb/libs/libcocotb.so
  Reason: image not found
```
Why does it work when using `brew`?

mtikekar: brew python is always installed in a fixed global location. So the `install_name` of `libpython` contains the full path and any library that links to `libpython` will look for it in the full path. So, rpath is not needed. Compare the output of `otool -L libcocotb.so` (Mac-equivalent of `ldd`) on the master branch for brew vs conda pythons:

For brew python:
```
libcocotb.so:
        @rpath/libcocotb.so (compatibility version 0.0.0, current version 0.0.0)
        /usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/Python (compatibility version 3.7.0, current version 3.7.0)
        @rpath/libgpilog.so (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libcocotbutils.so (compatibility version 0.0.0, current version 0.0.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
```

For conda python:
```
libcocotb.so:
        @rpath/libcocotb.so (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libpython3.7m.dylib (compatibility version 3.7.0, current version 3.7.0)
        @rpath/libgpilog.so (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libcocotbutils.so (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
```

Hence, `rpath` needs `python_lib_dirs` for conda python.

themperek: Thank you for your reply.
It seems adds up complication if we ever would want to distribute binary package (as you commented before) but this is one of many issues ...

themperek:
Thank you @mtikekar for your contribution.

Issue #1716, opened 2020-04-26
Move utility scripts out of the cocotb package


ktbarrett:
We should introduce another namespace to put python libraries/scripts that don't depend on the cocotb runtime. For example, the build script.

ktbarrett:
In the case of the `build_ext` it probably should be merged into setup.py, as it is done in the [main python distribution](https://github.com/python/cpython/blob/master/setup.py).

eric-wieser:
But note that that setup.py imports `distutils`, so its still the case that not all the code is in one place. In my opinion the split aids clarity for now, so we should keep it as long as we do so in a way that doesn't cause other issues.

ktbarrett:
I tried just moving build libs into the top level, but the import fails when using tox. Perhaps there is some option to get tox to recognize it as a build dependency? CPython puts their `build_ext` overload in setup.py; it's a big file, I feel if they knew how to split it up, they would.

eric-wieser:
Did you remember to add it to `MANIFEST.in`?

PR #1568, opened 2020-04-05
Unconditionally import the simulator module

eric-wieser:
Part of #1566. I expect this to fail initially, hence the draft status.

PR #1717, opened 2020-04-26
Unconditional import of cocotb and removal of COCOTB_SIM

### LINKED ISSUE #1566: Remove COCOTB_SIM environmental variable

ktbarrett:
Closes #1566. This removes `COCOTB_SIM` from the cocotb Python package. It was typically used to condition import of the simulator module. There was a number of accesses on the simulator module that caused segfaults if it wasn't conditioned. Now,

- the simulator module is imported unconditionally (usage still causes a segfault). Supplants #1568.
- all code that accessed the simulator module that ran at cocotb module's load has been moved so it runs later, in code paths only used for actual simulations
- because of that unconditional import when loading the package, the build script had to be moved out of the package. See #1716.

#1713 is a sister PR to remove the variable from the makefiles.

The merge commit was done so we can retain the history from `_build_libs.py` in `setup.py`. I know it looks weird. Perhaps that can be squashed? EDIT: no it cannot.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -47,11 +47,6 @@ from cocotb.regression import RegressionManager
 # Things we want in the cocotb namespace
 from cocotb.decorators import test, coroutine, hook, function, external  # noqa: F401
 
-# Singleton scheduler instance
-# NB this cheekily ensures a singleton since we're replacing the reference
-# so that cocotb.scheduler gives you the singleton instance and not the
-# scheduler package
-
 from ._version import __version__
 
 # GPI logging instance
@@ -102,6 +97,12 @@ if "COCOTB_SIM" in os.environ:
     # warning settings in their test module.
     if not sys.warnoptions:
         warnings.simplefilter("default")
+
+
+# Singleton scheduler instance
+# NB this cheekily ensures a singleton since we're replacing the reference
+# so that cocotb.scheduler gives you the singleton instance and not the
+# scheduler package
 
 scheduler = Scheduler()
 """The global scheduler instance."""
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -112,8 +112,10 @@ regression_manager = None
 plusargs = {}
 """A dictionary of "plusargs" handed to the simulation."""
 
-# To save typing provide an alias to scheduler.add
-fork = scheduler.add
+
+def fork(coro):
+    """ Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on it's use. """
+    return scheduler.add(coro)
 
 # FIXME is this really required?
 _rlock = threading.RLock()
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -49,13 +49,9 @@ from cocotb.decorators import test, coroutine, hook, function, external  # noqa:
 
 from ._version import __version__
 
-# GPI logging instance
-if "COCOTB_SIM" in os.environ:
 
-    # sys.path normally includes "" (the current directory), but does not appear to when python is embedded.
-    # Add it back because users expect to be able to import files in their test directory.
-    # TODO: move this to gpi_embed.cpp
-    sys.path.insert(0, "")
+def _setup_logging():
+    global log
 
     def _reopen_stream_with_buffering(stream_name):
         try:
@@ -92,12 +88,6 @@ if "COCOTB_SIM" in os.environ:
 
     del _stderr_buffer_result, _stdout_buffer_result
 
-    # From https://www.python.org/dev/peps/pep-0565/#recommended-filter-settings-for-test-runners
-    # If the user doesn't want to see these, they can always change the global
-    # warning settings in their test module.
-    if not sys.warnoptions:
-        warnings.simplefilter("default")
-
 
 # Singleton scheduler instance
 # NB this cheekily ensures a singleton since we're replacing the reference
@@ -116,6 +106,7 @@ plusargs = {}
 def fork(coro):
     """ Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on it's use. """
     return scheduler.add(coro)
+
 
 # FIXME is this really required?
 _rlock = threading.RLock()
@@ -153,6 +144,19 @@ def _initialise_testbench(argv_):
         elif '.' in root_name:
             # Skip any library component of the toplevel
             root_name = root_name.split(".", 1)[1]
+
+    # sys.path normally includes "" (the current directory), but does not appear to when python is embedded.
+    # Add it back because users expect to be able to import files in their test directory.
+    # TODO: move this to gpi_embed.cpp
+    sys.path.insert(0, "")
+
+    _setup_logging()
+
+    # From https://www.python.org/dev/peps/pep-0565/#recommended-filter-settings-for-test-runners
+    # If the user doesn't want to see these, they can always change the global
+    # warning settings in their test module.
+    if not sys.warnoptions:
+        warnings.simplefilter("default")
 
     from cocotb import simulator
 
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -38,6 +38,7 @@ import random
 import time
 import warnings
 
+import cocotb._os_compat  # must appear first, before the first import of cocotb.simulator
 import cocotb.handle
 import cocotb.log
 from cocotb.scheduler import Scheduler
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -104,7 +104,7 @@ if "COCOTB_SIM" in os.environ:
 # so that cocotb.scheduler gives you the singleton instance and not the
 # scheduler package
 
-scheduler = Scheduler()
+scheduler = None
 """The global scheduler instance."""
 
 regression_manager = None
@@ -170,6 +170,9 @@ def _initialise_testbench(argv_):
     # Create the base handle type
 
     process_plusargs()
+
+    global scheduler
+    scheduler = Scheduler()
 
     # Seed the Python random number generator to make this repeatable
     global RANDOM_SEED
--- a/cocotb/_os_compat.py
+++ b/cocotb/_os_compat.py
@@ -1,0 +1,21 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+"""
+OS-specific hacks
+"""
+import os
+import sys
+
+# This is necessary for Windows, which does not support RPATH, causing it
+# to not be able to locate the simulator module, which is imported
+# unconditionally.
+
+extra_dll_dir = os.path.join(os.path.dirname(__file__), 'libs')
+
+if sys.platform == 'win32' and os.path.isdir(extra_dll_dir):
+    if sys.version_info >= (3, 8):
+        os.add_dll_directory(extra_dll_dir)
+    else:
+        os.environ.setdefault('PATH', '')
+        os.environ['PATH'] += os.pathsep + extra_dll_dir
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -27,15 +27,10 @@
 
 """A collections of triggers which a testbench can await."""
 
-import os
 import abc
 from collections.abc import Awaitable
 
-if "COCOTB_SIM" in os.environ:
-    from cocotb import simulator
-else:
-    simulator = None
-
+from cocotb import simulator
 from cocotb.log import SimLog
 from cocotb.utils import (
     get_sim_steps, get_time_from_sim_steps, ParametrizedSingleton,
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -33,7 +33,7 @@ from collections.abc import Awaitable
 from cocotb import simulator
 from cocotb.log import SimLog
 from cocotb.utils import (
-    get_sim_steps, get_time_from_sim_steps, ParametrizedSingleton,
+    get_sim_steps, ParametrizedSingleton,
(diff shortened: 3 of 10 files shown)


ktbarrett:
The issue with the documentation comes down to cocotb needing to be built before documentation can be built for it. Perhaps this can be done in the refactor in #1706?

ktbarrett:
Will split this into two commits and do the merge of build libs into setup.py first.

eric-wieser:
> do the merge of build libs into setup.py first.

My approach here was simply to rename `cocotb/_build_libs` to `cocotb_build_libs.py`. Another option would be to copy the numpy approach, which allows `numpy.distutils` to be used before numpy is installed, via conditional execution of `numpy/__init__.py`

ktbarrett:
@eric-wieser I tried just moving build libs into the top level, but the import fails when using tox. Perhaps there is some option to get tox to recognize it as a build dependency? CPython puts their `build_ext` overload in the same file; it's a big file, I feel if they knew how to split it up, they would.

cocotb/__init__.py
```diff
@@ -135,8 +136,11 @@ def _initialise_testbench(root_name):
     The environment variable :envvar:`COCOTB_HOOKS`, if present, contains a
     comma-separated list of modules to be executed before the first test.
     """
+    _rlock = threading.RLock()
```
eric-wieser, line 139: This defeats the point of the rlock, which looks to have been to avoid `_initialise_testbench` being called twice concurrently. Locks can't do anything if only one function has them in scope.

cocotb/utils.py
```diff
@@ -35,12 +35,7 @@
 import functools
 import warnings
 
-if "COCOTB_SIM" in os.environ:
-    from cocotb import simulator
-    _LOG_SIM_PRECISION = simulator.get_precision()  # request once and cache
-else:
-    simulator = None
-    _LOG_SIM_PRECISION = -15
```
eric-wieser, line 43: Seems a shame we can't cache this - do we have a feel for how expensive `simulator.get_precision()` is on some common simulators? Going from `Python -> C -> GPI -> VPI` seems like a fair bit of overhead, but perhaps its not.

One option would be to use my IPython branch and use `%timeit simulator.get_precision()`

cocotb/__init__.py
ktbarrett: When would that ever happen? Perhaps we should just remove it.

cocotb/utils.py
ktbarrett: I'll test the timing, but we could just cache the value in the simulator module, it cannot change during the run.

ktbarrett: Or we could cache it in a local function.

cocotb/__init__.py
ktbarrett: I'll move that change to a separate PR.

cocotb/utils.py
ktbarrett: timeit reports `simulator.get_precision()` takes 112ns on my laptop, and the cached local function impl takes 78ns.

eric-wieser: Caching it in the `simulator.get_precision` might be sensible too

ktbarrett: Just tested that at 103ns.

cocotb/utils.py
```diff
@@ -35,12 +35,12 @@
...
+from cocotb import simulator
+
+
+@functools.lru_cache(maxsize=1)
+def _get_simulator_precision():
+    return simulator.get_precision()
```
eric-wieser, line 43: One final option:
```suggestion
def _get_simulator_precision():
    # cache and replace this function
    _get_simulator_precision.__code__ = (lambda: simulator.get_precision()).__code__
    return _get_simulator_precision()
```

Possibly a micro-optimization, but I'd expect it to be significantly faster.

ktbarrett: That's 56ns, I'll put that it.

ktbarrett:
Adding one more commit to revert the hack to set `COCOTB_SIM` from #1713.

ktbarrett:
@eric-wieser Any idea what is causing the windows build to fail? Your PR had the same issue.

themperek:
This is something overlooked 😭  `simulator` module is trying to load dependencies but since Windows does not support any `rpath` cannot find it. 

One way would be on Windows to add this:
```python
os.environ["PATH"] += os.pathsep + os.path.join(os.path.dirname(__file__), "libs")
```
at the beggining of `__init__.py` we could also remove from Makefiles then:
https://github.com/cocotb/cocotb/blob/934ec942d66f9663eb62b85fcc138d3356b9cfcb/cocotb/share/makefiles/Makefile.inc#L82-L86

eric-wieser:
Numpy uses something a bit cleverer, `os.add_dll_directory`:
```python
            import os
            import sys

            extra_dll_dir = os.path.join(os.path.dirname(__file__), '.libs')

            if sys.platform == 'win32' and os.path.isdir(extra_dll_dir):
                if sys.version_info >= (3, 8):
                    os.add_dll_directory(extra_dll_dir)
                else:
                    os.environ.setdefault('PATH', '')
                    os.environ['PATH'] += os.pathsep + extra_dll_dir
```

eric-wieser:
I think doing that and removing from the makefiles would be a good idea - perhaps in a standalone PR, so that it's not blocked on the conda doc changes.

cocotb/utils.py
eric-wieser: That's a sizeable improvement, great!

eric-wieser:
Removed on-hold, the doc build should pass after a rebase.

ktbarrett:
How do other projects handle that? Moving that to `__init__.py` seems like just as much of a hack. I'm sure numpy has a similar problem, it supports windows and builds private extensions.

themperek:
Removing `cocotb/libs` from PATH in Makefiles will probably not work since cocotbvpi/vhpi/fli have dependencies too and are loaded before python.

ktbarrett:
@eric-wieser Do you think the documentation build failure is due to a stale build cache?

eric-wieser:
I have no idea what the build failure is caused by - perhaps a new towncrier release broke us?

eric-wieser:
Is this print in the logs?

https://github.com/twisted/towncrier/blob/bca0525b9b35bbc4017d6681c2aaa385cd70766a/src/towncrier/_project.py#L27

themperek:
Maybe missign `package_dir` in  `pyproject.toml`? See: https://pypi.org/project/towncrier/

eric-wieser:
I think towncrier is trying to import cocotb from the source directory, not the installed directory.

Perhaps we need to do an editable install in order to make everything happy.

themperek:
> Perhaps we need to do an editable install in order to make everything happy.

It seems if one set  `package_dir` in `pyproject.toml` to anything it works problem is:
https://github.com/twisted/towncrier/blob/bca0525b9b35bbc4017d6681c2aaa385cd70766a/src/towncrier/_project.py#L22

eric-wieser:
Set it to `/tmp/` perhaps?

themperek:
So nicer (maybe) solution seems to be adding `--version 1.4.0.dev0 --name cocotb` here:

https://github.com/cocotb/cocotb/blob/c2a6285432fee5090b2a4c1b890544be094451f5/documentation/source/conf.py#L339-L341

It does not import in this case anything.

eric-wieser:
Agreed, that does sound like the best solution, assuming we sub in `cocotb.__version__` which we already have imported anyway.

cocotb/_os_compat.py
```diff
@@ -0,0 +1,11 @@
+import os
+import sys
+
+extra_dll_dir = os.path.join(os.path.dirname(__file__), '.libs')
```
themperek, line 4:
```suggestion
extra_dll_dir = os.path.join(os.path.dirname(__file__), 'libs')
```

ktbarrett:
I'm guessing it tries to get documentation on something that ends up calling the simulator module, that's what the segfault is. I'll look it that later.

ktbarrett:
I'm not sure about that last commit's message; I don't really know what it's doing. `towncrier --help` was no help.

themperek:
> I'm not sure about that last commit's message; I don't really know what it's doing. `towncrier --help` was no help.

Avoid importing cocotb by adding a version and module name explicitly in the command line

ktbarrett:
@themperek But why is that still necessary? We are installing cocotb and have made it so that we can import it without issue.

themperek:
> But why is that still necessary? We are installing cocotb and have made it so that we can import it without issue.

Because `towncrier` manipulates `sys.path` and tries to import cocotb from source directory.

ktbarrett:
Could we instead change the CWD?

themperek:
> Could we instead change the CWD?

Maybe There are a few ways,

documentation/source/conf.py
```diff
@@ -336,7 +336,7 @@
 # -- Extra setup for towncrier -------------------------------------------------
 # see also https://towncrier.readthedocs.io/en/actual-freaking-docs/
 
-in_progress_notes = subprocess.check_output(['towncrier', '--draft'],
+in_progress_notes = subprocess.check_output(['towncrier', '--draft', '--name', 'cocotb', '--version', release],
```
eric-wieser, line 339:
```suggestion
# we pass the name and version directly, to avoid towncrier failing to import the non-installed version
in_progress_notes = subprocess.check_output(['towncrier', '--draft', '--name', 'cocotb', '--version', release],
```

documentation/Makefile
```diff
@@ -21,7 +21,7 @@ help: .venv
 	python3 -m venv .venv
 	.venv/bin/pip -q install --upgrade pip
 	.venv/bin/pip -q install -r requirements.txt
-	.venv/bin/pip -q install ..
+	.venv/bin/pip -q install -e ..
```
eric-wieser, line 24: Note that this isn't used in CI. Is it needed locally either?

cocotb/triggers.py
```diff
@@ -199,7 +194,12 @@ def __init__(self, time_ps, units=None):
             :func:`~cocotb.utils.get_sim_steps`
         """
         GPITrigger.__init__(self)
-        self.sim_steps = get_sim_steps(time_ps, units)
+        self._time_ps = time_ps
+        self._units = units
+
+    @lazy_property
+    def sim_steps(self):
+        return get_sim_steps(self._time_ps, self._units)
```
eric-wieser, line 202:
```suggestion
        # lazy so that we don't call into the simulator until we need to
        return get_sim_steps(self._time_ps, self._units)
```

documentation/Makefile
ktbarrett: I added it for use locally. Otherwise you would have to rebuild the virtualenv every time you rebuilt the documentation locally. Locally its building documentation hy importing the cocotb installation in the virtualenv. That's why I was questioning the CWD of towncrier.

documentation/source/conf.py
ktbarrett: It seems to me a better solution would be to make towncrier use the installed version of cocotb rather than the source version. I guess that can come in a later PR if we can figure it out.

themperek: It may be that towncrier does not need cocotb it just look at news fragments that are not part of cooctb, It just needs version and name. I am not an expert.

cocotb/__init__.py
```diff
@@ -38,6 +38,7 @@
 import time
 import warnings
 
+import cocotb._os_compat  # must appear first
```
eric-wieser, line 41:
```suggestion
import cocotb._os_compat  # must appear first, before anything imports cocotb.simulator
```

cocotb/__init__.py
```diff
@@ -103,16 +100,24 @@ def _reopen_stream_with_buffering(stream_name):
     if not sys.warnoptions:
         warnings.simplefilter("default")
```
eric-wieser, line 101: This doesn't really belong in a function called `setup_logging`. Perhaps it should move to the regression manager.

cocotb/__init__.py
```diff
@@ -47,15 +48,11 @@
...
 from ._version import __version__
 
-# GPI logging instance
-if "COCOTB_SIM" in os.environ:
+
+def _setup_logging():
+    global log
 
     # sys.path normally includes "" (the current directory), but does not appear to when python is embedded.
     # Add it back because users expect to be able to import files in their test directory.
```
eric-wieser, line 58: This probably also belongs in `regression.py`

eric-wieser: If you don't want to handle this now, then perhaps just rename the function to `_setup_logging_warnings_and_path`

cocotb/__init__.py
ktbarrett: It doesn't really have much to do with regression, but I will move it into `_initialize_testbench`

cocotb/_os_compat.py
```diff
@@ -0,0 +1,11 @@
+import os
```
imphil, line 1: This file needs a license header.

cocotb/_os_compat.py
```diff
@@ -0,0 +1,11 @@
...
+import sys
+
+extra_dll_dir = os.path.join(os.path.dirname(__file__), 'libs')
+
+if sys.platform == 'win32' and os.path.isdir(extra_dll_dir):
```
imphil, line 6: Can you add the information from the commit message as comment to this file?

eric-wieser:
Conflicts unfortunately

eric-wieser:
It would be interesting to see if Sphinx can now autodoc the simulator module.

cmarqu:
> It would be interesting to see if Sphinx can now autodoc the simulator module.

It can indeed: #1825

PR #1719, opened 2020-04-27
Move build_libs.py out of cocotb package

### LINKED ISSUE #1716: Move utility scripts out of the cocotb package

ktbarrett:
Prevents the running of the cocotb package when importing. Need for unconditioned import of the simulator module. Closes #1716.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,10 +53,9 @@ from setuptools import find_packages
 from os import path, walk
 from io import StringIO
 
-# note: cocotb is not installed properly yet, but we can import it anyway
-# because it's in the current directory. We'll need to change this if we
-# add `install_requires` to the `setup()` call.
-from cocotb._build_libs import get_ext, build_ext
+# Note: cocotb is not installed properly yet and is missing dependencies and binaries
+# We can still import other files next to setup.py, as long as they're in MANIFEST.in
+from cocotb_build_libs import get_ext, build_ext
 
 def read_file(fname):
     with open(path.join(path.dirname(__file__), fname), encoding='utf8') as f:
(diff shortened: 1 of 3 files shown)


build_libs.py
```diff
@@ -16,7 +16,7 @@
 
 
 logger = logging.getLogger(__name__)
-cocotb_share_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "share"))
+cocotb_share_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "cocotb", "share"))
```
eric-wieser, line 19: We might want to consider moving this source code out of `cocotb` at some point too, or at least removing it from `package_data` in `setup.py` - today we install the source, which doesn't really make any sense now that installation builds binaries.

setup.py
```diff
@@ -56,7 +56,7 @@
 # note: cocotb is not installed properly yet, but we can import it anyway
 # because it's in the current directory. We'll need to change this if we
 # add `install_requires` to the `setup()` call.
```
eric-wieser, line 58:
```suggestion
# Note: cocotb is not installed properly yet and is missing dependencies and binaries
# We can still import other files next to setup.py, as long as they're in MANIFEST.in
```

setup.py
```diff
@@ -56,7 +56,7 @@
...
-from cocotb._build_libs import get_ext, build_ext
+from build_libs import get_ext, build_ext
```
eric-wieser, line 59: Possibly worth giving it an obvious prefix, either
```suggestion
from cocotb_build_libs import get_ext, build_ext
```
or
```suggestion
from setup_build_libs import get_ext, build_ext
```

build_libs.py
ktbarrett: I had thought about it too, but it's not pertinent for the changes I want to make immediately.

ktbarrett: Actually, it's pretty trivial, why not.

ktbarrett: Nvm, not trivial.

PR #1723, opened 2020-04-28
Add is_set method to Event

ktbarrett:
Mirrors [`asyncio.Event`](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event.is_set). While `.fired` does exist, and is public, it is not documented.

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -420,6 +420,10 @@ class Event:
         :meth:`~cocotb.triggers.Event.set` is called again."""
         self.fired = False
 
+    def is_set(self):
+        """ Return true if event has been set """
+        return self.fired
+
     def __repr__(self):
         if self.name is None:
             fmt = "<{0} at {2}>"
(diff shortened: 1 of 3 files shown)


ktbarrett:
@eric-wieser Did I go overboard or underboard on the type annotations? I annotated everything but the special methods that have known output types, like `__repr__`. I put it through mypy and got it to not scream about anything.

eric-wieser:
> @eric-wieser Did I go overboard or underboard on the type annotations?

If you're going to go that overboard, I'd prefer you do so in a standalone PR next time.

eric-wieser:
Can you add a simple test too?

cocotb/triggers.py
```diff
@@ -420,6 +420,10 @@ def clear(self):
         :meth:`~cocotb.triggers.Event.set` is called again."""
         self.fired = False
 
+    def is_set(self):
```
eric-wieser, line 423:
```suggestion
    def is_set(self) -> bool:
```
while you're here, maybe

eric-wieser:
Unless you deliberately want to save all the typing to be done at once, it's tempting to add the annotation while we're here. Feel free to commit through the github UI, I'll squash merge anyway.

PR #1732, opened 2020-04-29
Add howto section "Rotating Log Files"

### LINKED ISSUE #1400: Document how to get a rotating logger
In #1224 we had the request to support a "rotating" logger in cocotb. 
Now that #1266 has gone in there is a reasonably comfortable way to do that without modifying cocotb itself.

```python
root_logger = logging.getLogger()

# undo the setup cocotb did
for handler in root_logger.handlers:
    root_logger.remove_handler(handler)
    handler.close()

# do whatever configuration you want instead
file_handler = RotatingFileHandler(logfile, maxBytes=(1048576*5), backupCount=4)
file_handler.setFormatter(cocotb.log.SimLogFormatter())
root_logger.addHandler(file_handler)
```
at which point this seems like something that doesn't need to be in cocotb itself.

_Originally posted by @eric-wieser in https://github.com/cocotb/cocotb/pull/1224#issuecomment-581368001_

This issue is to document this as a useful snippet somewhere in the docs, as the general problem seems to be something other users might stumble across.

cmarqu:
Closes #1400.

https://cocotb--1732.org.readthedocs.build/en/1732/rotating_logger.html

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -67,8 +67,10 @@ def default_config():
 
     If desired, this logging configuration can be overwritten by calling
     ``logging.basicConfig(..., force=True)`` (in Python 3.8 onwards), or by
-    manually resetting the root logger instance, for which examples can be
-    found online.
+    manually resetting the root logger instance.
+    An example of this can be found in the section on :ref:`rotating-logger`.
+
+    .. versionadded:: 1.4
     """
     # construct an appropriate handler
     hdlr = logging.StreamHandler(sys.stdout)
@@ -135,11 +137,13 @@ class SimTimeContextFilter(logging.Filter):
     This uses the approach described in the :ref:`Python logging cookbook <python:filters-contextual>`.
 
     This adds the :attr:`~logging.LogRecord.created_sim_time` attribute.
+
+    .. versionadded:: 1.4
     """
 
     # needed to make our docs render well
     def __init__(self):
-        """ Takes no arguments """
+        """"""
         super().__init__()
 
     def filter(self, record):
--- a/documentation/source/rotating_logger.rst
+++ b/documentation/source/rotating_logger.rst
@@ -1,0 +1,27 @@
+.. _rotating-logger:
+
+******************
+Rotating Log Files
+******************
+
+The following is an example of how to support rotation of log files.
+It will keep the newest 3 files which are at most 5 MiB in size.
+
+See also :ref:`logging-reference-section`,
+and the Python documentation for :class:`logging.RotatingFileHandler`.
+
+.. code-block:: python3
+
+    from logging.handlers import RotatingFileHandler
+
+    root_logger = logging.getLogger()
+
+    # undo the setup cocotb did
+    for handler in root_logger.handlers:
+        root_logger.remove_handler(handler)
+        handler.close()
+
+    # do whatever configuration you want instead
+    file_handler = RotatingFileHandler(logfile, maxBytes=(5 * 1024 * 1024), backupCount=2)
+    file_handler.setFormatter(cocotb.log.SimLogFormatter())
+    root_logger.addHandler(file_handler)
--- a/documentation/source/rotating_logger.rst
+++ b/documentation/source/rotating_logger.rst
@@ -8,7 +8,7 @@ The following is an example of how to support rotation of log files.
 It will keep the newest 3 files which are at most 5 MiB in size.
 
 See also :ref:`logging-reference-section`,
-and the Python documentation for :class:`logging.RotatingFileHandler`.
+and the Python documentation for :class:`logging.handlers.RotatingFileHandler`.
 
 .. code-block:: python3
 
--- a/documentation/source/index.rst
+++ b/documentation/source/index.rst
@@ -120,6 +120,7 @@ A test can spawn multiple coroutines, allowing for independent flows of executio
    coroutines
    triggers
    testbench_tools
+   rotating_logger
 
 .. todo::
    - Add WaveDrom, IPython sections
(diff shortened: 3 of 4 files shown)


cmarqu:
Should this have a ``.. versionadded:: 1.4`` to make clear this didn't work before?

eric-wieser:
Can we either merge with or link to this section of the docs: https://docs.cocotb.org/en/latest/library_reference.html#logging

eric-wieser:
I think the `.. versionadded:: 1.4` belongs in the new entries in the logging section I link to above (`default_config` and `SimTimeContextFilter`) - then the rotating logger example can just reference those functions.

cmarqu:
Done - though should the ``.. versionadded:: 1.4`` better go into the docstrings maybe?
I just added a generic "see also" reference since the code example doesn't use any of these added things directly, did you have something more specific in mind?

eric-wieser:
> though should the `.. versionadded:: 1.4` better go into the docstrings maybe?

yes, that's what I meant!

cmarqu:
Hmm, https://cocotb--1732.org.readthedocs.build/en/1732/library_reference.html#cocotb.log.SimTimeContextFilter doesn't look so great, but the "Takes no arguments" comes from ``__init__``, and it feels bad to add the ``versionadded`` there...

cocotb/log.py
```diff
@@ -135,6 +137,8 @@ class SimTimeContextFilter(logging.Filter):
     This uses the approach described in the :ref:`Python logging cookbook <python:filters-contextual>`.
 
     This adds the :attr:`~logging.LogRecord.created_sim_time` attribute.
+
+    .. versionadded:: 1.4
     """
 
     # needed to make our docs render well
```
eric-wieser, line 144: Could maybe make the docstring empty on the line below.

cocotb/log.py
```diff
@@ -69,6 +69,8 @@ def default_config():
     ``logging.basicConfig(..., force=True)`` (in Python 3.8 onwards), or by
     manually resetting the root logger instance, for which examples can be
     found online.
```
eric-wieser, line 71:
```suggestion
    found online.
    
    An example of this can be found in the section on :ref:`rotating-logger`.
```

cocotb/log.py
cmarqu: This works indeed, will push later.

cocotb/log.py
cmarqu: Done (with the online examples part removed).

cmarqu:
> Not super happy with where this has ended up in the docs

I think the location is in spirit of the Daniele Procida/Divio structure (https://documentation.divio.com/) that I'm trying to follow.

documentation/source/rotating_logger.rst
```diff
@@ -0,0 +1,26 @@
+.. _rotating-logger:
+
+***********************
+Rotating Disk Log Files
+***********************
```
eric-wieser, line 5:
```suggestion
*****************
Log File Rotation
*****************
```
I'm not sure "disk" is a useful word here.

documentation/source/rotating_logger.rst
```diff
@@ -0,0 +1,26 @@
...
+
+The following is an example of how to support rotation of disk log files.
```
eric-wieser, line 7:
```suggestion
The following is an example of how to support rotation of log files.
```

documentation/source/rotating_logger.rst
```diff
@@ -0,0 +1,26 @@
...
+It will keep the newest 3 files which are at a maximum 5 MiB large.
```
eric-wieser, line 8:
```suggestion
It will keep the newest 3 files which are at a maximum of 5 MiB in size.
```

documentation/source/rotating_logger.rst
```diff
@@ -0,0 +1,26 @@
...
+
+See also :ref:`logging-reference-section`.
```
eric-wieser, line 10:
```suggestion
See also :ref:`logging-reference-section`, and the python docs for :class:`logging.RotatingFileHandler`.
```

documentation/source/rotating_logger.rst
cmarqu: I need to lose the "at" in the sentence too, right?

eric-wieser: Perhaps 'at most 5 MiB in size"

documentation/source/rotating_logger.rst
cmarqu: Done.

documentation/source/rotating_logger.rst
cmarqu: Done.

documentation/source/rotating_logger.rst
cmarqu: Done.

documentation/source/rotating_logger.rst
cmarqu: Done.

documentation/source/rotating_logger.rst
```diff
@@ -0,0 +1,27 @@
...
+
+******************
+Rotating Log Files
+******************
+
+The following is an example of how to support rotation of log files.
+It will keep the newest 3 files which are at most 5 MiB in size.
+
+See also :ref:`logging-reference-section`,
+and the Python documentation for :class:`logging.RotatingFileHandler`.
```
eric-wieser, line 11:
```suggestion
and the Python documentation for :class:`logging.handlers.RotatingFileHandler`.
```

eric-wieser:
That follow-up commit fixed the python cross-reference, squashed together in the merge.

PR #1735, opened 2020-04-29
Change how cocotb is built for documentation

ktbarrett:
Encountered an error while building documentation locally on master. It seems that the cocotb package doesn't make it into the virtualenv that the documentation build uses. I get the following error.

```
~/dev/cocotb/documentation (master) $ make html
make .venv
make[1]: Entering directory '/home/kaleb/dev/cocotb/documentation'
make[1]: '.venv' is up to date.
make[1]: Leaving directory '/home/kaleb/dev/cocotb/documentation'
source .venv/bin/activate; .venv/bin/sphinx-build -M html "source" "build"  
/bin/sh: 1: source: not found
Running Sphinx v3.0.3

Configuration error:
There is a programmable error in your configuration file:

  File "/home/kaleb/dev/cocotb/documentation/.venv/lib/python3.7/site-packages/sphinx/config.py", line 319, in eval_config_file
    execfile_(filename, namespace)
  File "/home/kaleb/dev/cocotb/documentation/.venv/lib/python3.7/site-packages/sphinx/util/pycompat.py", line 81, in execfile_
    exec(code, _globals)
  File "/home/kaleb/dev/cocotb/documentation/source/conf.py", line 18, in <module>
    import cocotb
ModuleNotFoundError: No module named 'cocotb'

make: *** [Makefile:43: html] Error 2
```

I added cocotb to the list of packages to install in the Makefile. Should this instead go into the `requirements.txt`?

### DIFF
--- a/documentation/Makefile
+++ b/documentation/Makefile
@@ -21,6 +21,7 @@ help: .venv
 	python3 -m venv .venv
 	.venv/bin/pip -q install --upgrade pip
 	.venv/bin/pip -q install -r requirements.txt
+	.venv/bin/pip -q install ..
 
 .PHONY: clean
 clean: .venv Makefile


cmarqu:
Does this have to do with https://github.com/cocotb/cocotb/pull/1706/commits/022bf0b86b66e629502b7e4751982659cf6a7d98 ?

ktbarrett:
Potentially. But once #1717 is in, that won't work, we have to have the simulator module built or the importing cocotb will fail.

cmarqu:
Adding ``cocotb`` to the ``requirements.txt`` sounds correct to me.

.readthedocs.yml
```diff
@@ -8,10 +8,5 @@ version: 2
 conda:
   environment: documentation/conda.yml
 
-python:
-  install:
-    - method: pip
-      path: .
```
eric-wieser, line 14: By doing this, we lose the ability to see seperate logs in rtd for our build vs that of dependencies.

I preferred the first iteration of this patch which touched only the makefile.

documentation/requirements.txt
```diff
@@ -10,3 +10,4 @@ pyenchant
 sphinx-issues
 sphinx-argparse
 towncrier
+..
```
eric-wieser, line 13: This is probably harmful for users who already have installed cocotb with `-e`

ktbarrett: That will install cocotb into the virtualenv created for documentation, it won't collide.

eric-wieser: Not if I consume this file directly - you're assuming I came through the makefiles, which I did not.

PR #1737, opened 2020-04-30
Move product and version information to simulatormodule

ktbarrett:
Previously, this information was gathered during startup, and passed to Python via gpi_embed. This change reduces the responsibility of gpi_embed and allows either C code or Python code to determine the product information.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -137,6 +137,14 @@ def _initialise_testbench(root_name):
     """
     _rlock.acquire()
 
+    from cocotb import simulator
+
+    global SIM_NAME, SIM_VERSION
+    SIM_NAME = simulator.get_simulator_product()
+    SIM_VERSION = simulator.get_simulator_version()
+
+    cocotb.log.info("Running on {} version {}".format(SIM_NAME, SIM_VERSION))
+
     memcheck_port = os.getenv('MEMCHECK')
     if memcheck_port is not None:
         mem_debug(int(memcheck_port))
--- a/cocotb/share/include/gpi.h
+++ b/cocotb/share/include/gpi.h
@@ -111,15 +111,6 @@ typedef enum gpi_event_e {
     SIM_FAIL = 2,
 } gpi_event_t;
 
-typedef struct gpi_sim_info_s
-{
-    int32_t   argc;
-    char      **argv;
-    char      *product;
-    char      *version;
-    int32_t   *reserved[4];
-} gpi_sim_info_t;
-
 // Functions for controlling/querying the simulation state
 
 // Stop the simulator
@@ -132,6 +123,19 @@ void gpi_cleanup(void);
 void gpi_get_sim_time(uint32_t *high, uint32_t *low);
 void gpi_get_sim_precision(int32_t *precision);
 
+/**
+ * Returns a string with the running simulator product information
+ *
+ * @return simulator product string
+ */
+const char *gpi_get_simulator_product(void);
+
+/**
+ * Returns a string with the running simulator version
+ *
+ * @return simulator version string
+ */
+const char *gpi_get_simulator_version(void);
 
 // Functions for extracting a gpi_sim_hdl to an object
 // Returns a handle to the root simulation object,
--- a/cocotb/share/lib/fli/FliCbHdl.cpp
+++ b/cocotb/share/lib/fli/FliCbHdl.cpp
@@ -192,42 +192,16 @@ static std::vector<std::string> get_argv()
 
 int FliStartupCbHdl::run_callback()
 {
-    gpi_sim_info_t sim_info;
 
-    char *c_info       = mti_GetProductVersion();      // Returned pointer must not be freed
-    std::string info   = c_info;
-    std::string search = " Version ";
-    std::size_t found  = info.find(search);
-
-    std::string product_str = c_info;
-    std::string version_str = c_info;
-
-    if (found != std::string::npos) {
-        product_str = info.substr(0,found);
-        version_str = info.substr(found+search.length());
-
-        LOG_DEBUG("Found Version string at %d", found);
-        LOG_DEBUG("   product: %s", product_str.c_str());
-        LOG_DEBUG("   version: %s", version_str.c_str());
+    std::vector<std::string> const argv_storage = get_argv();
+    std::vector<const char*> argv_cstr;
+    for (const auto& arg : argv_storage) {
+        argv_cstr.push_back(arg.c_str());
     }
+    int argc = static_cast<int>(argv_storage.size());
+    const char** argv = argv_cstr.data();
 
-    std::vector<char> product(product_str.begin(), product_str.end());
-    std::vector<char> version(version_str.begin(), version_str.end());
-    product.push_back('\0');
-    version.push_back('\0');
-
-    sim_info.product = &product[0];
-    sim_info.version = &version[0];
-
-    std::vector<std::string> argv = get_argv();
-    std::vector<char*> argv_cstr;
-    for (const auto& arg : argv) {
-        argv_cstr.push_back(const_cast<char*>(arg.c_str()));
-    }
-    sim_info.argc = static_cast<int>(argv.size());
-    sim_info.argv = argv_cstr.data();
-
-    gpi_embed_init(&sim_info);
+    gpi_embed_init(argc, argv);
 
     return 0;
 }
--- a/cocotb/share/lib/fli/FliImpl.cpp
+++ b/cocotb/share/lib/fli/FliImpl.cpp
@@ -434,6 +434,30 @@ void FliImpl::get_sim_precision(int32_t *precision)
     *precision = mti_GetResolutionLimit();
 }
 
+const char *FliImpl::get_simulator_product()
+{
+    if (m_product.empty() && m_version.empty()) {
+        const std::string info = mti_GetProductVersion(); // Returned pointer must not be freed, does not fail
+        const std::string search = " Version ";
+        const std::size_t found = info.find(search);
+
+        if (found != std::string::npos) {
+            m_product = info.substr(0, found);
+            m_version = info.substr(found + search.length());
+        } else {
+            m_product = info;
+            m_version = "UNKNOWN";
+        }
+    }
+    return m_product.c_str();
+}
+
+const char *FliImpl::get_simulator_version()
+{
+    get_simulator_product();
+    return m_version.c_str();
+}
+
 /**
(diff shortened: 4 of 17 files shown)


cocotb/__init__.py
```diff
@@ -137,6 +137,15 @@ def _initialise_testbench(root_name):
...
     _rlock.acquire()
 
+    from cocotb import simulator
+
+    global SIM_NAME, SIM_VERSION
+    SIM_NAME = simulator.get_simulator_product()
+    SIM_VERSION = simulator.get_simulator_version()
+
+    cocotb.log.info("Running on {} version {}".format(SIM_NAME, SIM_VERSION))
+    cocotb.log.info("Python interpreter initialized and cocotb loaded!")
```
ktbarrett, line 147: This was moved to preserve the order of the log statements, but it probably shouldn't be here.

cocotb/share/include/gpi.h
```diff
@@ -132,6 +123,8 @@ void gpi_cleanup(void);
 void gpi_get_sim_time(uint32_t *high, uint32_t *low);
 void gpi_get_sim_precision(int32_t *precision);
 
+const char *gpi_get_simulator_product(void);
+const char *gpi_get_simulator_version(void);
```
ktbarrett, line 127: Add documentation.

cocotb/share/lib/vhpi/VhpiImpl.h
```diff
@@ -241,12 +241,16 @@ class VhpiImpl : public GpiImplInterface {
     VhpiImpl(const std::string& name) : GpiImplInterface(name),
                                         m_read_write(this),
                                         m_next_phase(this),
-                                        m_read_only(this) { }
+                                        m_read_only(this),
+                                        m_product(NULL),
+                                        m_version(NULL) { }
```
ktbarrett, line 246: These are redundant, remove them.

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -72,6 +72,26 @@ void VpiImpl::get_sim_precision(int32_t *precision)
     *precision = vpi_get(vpiTimePrecision, NULL);
 }
 
+const char *VpiImpl::get_simulator_product()
+{
+    if (m_product.empty()) {
+        s_vpi_vlog_info info;
+        vpi_get_vlog_info(&info);
+        m_product = info.product;
```
ktbarrett, line 80: The VHPI implementation releases the handle, do we need to do that here? The original code did not.

cocotb/share/lib/vpi/VpiImpl.h
```diff
@@ -266,6 +268,8 @@ class VpiImpl : public GpiImplInterface {
     VpiReadwriteCbHdl m_read_write;
     VpiNextPhaseCbHdl m_next_phase;
     VpiReadOnlyCbHdl m_read_only;
+    std::string m_product;
+    std::string m_version;
```
garmin-mjames, line 272: Can these not go in `GpiImplInterface` as `protected`?

ktbarrett: Originally the implementations were different types but I later normalized them to strings, that more appropriate now.

cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -192,42 +192,16 @@ static std::vector<std::string> get_argv()
...
-    version.push_back('\0');
-
-    sim_info.product = &product[0];
-    sim_info.version = &version[0];
-
-    std::vector<std::string> argv = get_argv();
-    std::vector<char*> argv_cstr;
-    for (const auto& arg : argv) {
-        argv_cstr.push_back(const_cast<char*>(arg.c_str()));
+    std::vector<std::string> argv_storage = get_argv();
```
eric-wieser, line 196:
```suggestion
    std::vector<std::string> const argv_storage = get_argv();
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -761,6 +761,24 @@ static PyObject *get_precision(PyObject *self, PyObject *args)
     return PyLong_FromLong(precision);
 }
 
+static PyObject *get_simulator_product(PyObject *m, PyObject *args)
+{
+    COCOTB_UNUSED(m);
+    COCOTB_UNUSED(args);
+    const char *product_string = gpi_get_simulator_product();
+    PyObject *result = Py_BuildValue("s", product_string);
+    return result;
```
eric-wieser, line 770:
```suggestion
    return PyUnicode_FromString(product_string);
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -761,6 +761,24 @@ static PyObject *get_precision(PyObject *self, PyObject *args)
...
+}
+
+static PyObject *get_simulator_version(PyObject *m, PyObject *args)
+{
+    COCOTB_UNUSED(m);
+    COCOTB_UNUSED(args);
+    const char *version_string = gpi_get_simulator_version();
+    PyObject *result = Py_BuildValue("s", version_string);
+    return result;
```
eric-wieser, line 779: Don't think extra lines offer any clarity here:
```suggestion
    return PyUnicode_FromString(gpi_get_simulator_version());
```

cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -434,6 +434,38 @@ void FliImpl::get_sim_precision(int32_t *precision)
...
+            product_str = info.substr(0, found);
+            version_str = info.substr(found+search.length());
+
+            LOG_DEBUG("Found Version string at %d", found);
+            LOG_DEBUG("   product: %s", product_str.c_str());
+            LOG_DEBUG("   version: %s", version_str.c_str());
+        }
+
+        m_product.assign(product_str.begin(), product_str.end());
+        m_version.assign(version_str.begin(), version_str.end());
```
eric-wieser, line 458: This code would be clearer without the intermediate `product_str` variable and using `operator=` instead of `.assign`, but I can see this is just moved code, so no need to tidy it in this PR.

ktbarrett: Tidied up anyways. Will test first.

ktbarrett: It builds and prints the correct string when running tests.

cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -434,6 +434,31 @@ void FliImpl::get_sim_precision(int32_t *precision)
...
+        const char *c_info       = mti_GetProductVersion();      // Returned pointer must not be freed
+        const std::string info   = c_info;
+        const std::string search = " Version ";
+        const std::size_t found  = info.find(search);
+
+        if (found != std::string::npos) {
+            m_product = info.substr(0, found);
+            m_version = info.substr(found+search.length());
+        } else {
+            m_product = "UNKNOWN";
```
eric-wieser, line 449:
```suggestion
            m_product = info;
```
Better to assume we have the product string than drop it on the ground

cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -434,6 +434,31 @@ void FliImpl::get_sim_precision(int32_t *precision)
     *precision = mti_GetResolutionLimit();
 }
 
+const char *FliImpl::get_simulator_product()
+{
+    if (m_product.empty() && m_version.empty()) {
+        const char *c_info       = mti_GetProductVersion();      // Returned pointer must not be freed
+        const std::string info   = c_info;
+        const std::string search = " Version ";
+        const std::size_t found  = info.find(search);
```
eric-wieser, line 443: nit: could make this shorter
```suggestion
        const std::string info   = mti_GetProductVersion();
        const std::size_t found  = info.find(" Version ");
```

cocotb/share/lib/vhpi/VhpiImpl.cpp
```diff
@@ -110,6 +110,30 @@ void VhpiImpl::get_sim_precision(int32_t *precision)
...
+}
+
+const char *VhpiImpl::get_simulator_version()
+{
+    if (m_version.empty()) {
+        vhpiHandleT tool = vhpi_handle(vhpiTool, NULL);
+        if (tool) {
+            m_version = vhpi_get_str(vhpiToolVersionP, tool);
+            vhpi_release_handle(tool);
+        }
```
eric-wieser, line 132: For consistency with the other impls:
```suggestion
        }
        else {
            m_version = "UNKNOWN";
        }
```

cocotb/share/lib/fli/FliImpl.cpp
ktbarrett: `search` is used later, and I'd rather not hardcode the length, but I did fold `c_info` away.

ktbarrett:
FLI still builds, runs, and reports product and version correctly.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -761,6 +761,20 @@ static PyObject *get_precision(PyObject *self, PyObject *args)
     return PyLong_FromLong(precision);
 }
 
+static PyObject *get_simulator_product(PyObject *m, PyObject *args)
+{
+    COCOTB_UNUSED(m);
+    COCOTB_UNUSED(args);
+    return Py_BuildValue("s", gpi_get_simulator_product());
```
eric-wieser, line 768: Any reason you're not using the `PyUnicode_FromString` I suggested before? This indirection makes it harder to work out whether this is unicode or bytes IMO.

```suggestion
    return PyUnicode_FromString(gpi_get_simulator_product());
```

It also makes it more consistent with the rest of the file.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -761,6 +761,20 @@ static PyObject *get_precision(PyObject *self, PyObject *args)
...
+}
+
+static PyObject *get_simulator_version(PyObject *m, PyObject *args)
+{
+    COCOTB_UNUSED(m);
+    COCOTB_UNUSED(args);
+    return Py_BuildValue("s", gpi_get_simulator_version());
```
eric-wieser, line 775:
```suggestion
    return PyUnicode_FromString(gpi_get_simulator_version());
```

eric-wieser:
@imphil, want to give this a quick skim before I merge?

imphil:
Argh, should have been squashed first. Too late now.

PR #966, opened 2019-06-27
Use scikit-build to build the so files and lib files

eric-wieser:
Putting up here mainly to demonstrate an option.

Not a complete solution, but this correctly builds the python extension module and the non-proprietary shared libraries as part of `setup.py`, which means they're also suited for packaging as a wheel.

I've been having some issues with this on windows still (https://github.com/scikit-build/scikit-build/issues/435), suggesting that the upstream project is perhaps not mature enough yet.

PR #1739, opened 2020-04-30
Use setuptools_dso to build libraries (WIP)

mtikekar:
This was my shot at using [setuptools_dso](https://github.com/mdavidsaver/setuptools_dso) to build the C++ libraries based on discussion in this [issue](https://github.com/cocotb/cocotb/pull/1714#issuecomment-621254257). There are many changes, but I've tried to break them into the smallest possible pieces per commit.

The biggest change is that setuptools_dso is a build-time requirement (`setup_requires` vs `install_requires` in `setup.py`). I learned that this is a major pain-point in pip although `pyproject.toml` aims to fix that. Right now, `pip install .` works, but for develop-mode, you need to use `python setup.py build_dso -i develop`.

I have only done preliminary tests with Xcelium and Icarus on Linux, and Icarus and Verilator on Mac. Would appreciate some help there.


eric-wieser:
Not sure why Travis did not run

cocotb_build_libs.py
```diff
@@ -294,17 +206,17 @@ def _get_vpi_lib_ext(
 def _get_vhpi_lib_ext(
     include_dir, share_lib_dir, sim_define, extra_lib=[], extra_lib_dir=[]
 ):
-    libcocotbvhpi = Extension(
-        os.path.join("cocotb", "libs", "libcocotbvhpi_" + sim_define.lower()),
+    libcocotbvhpi = DSO(
+        dso_join("cocotb", "libs", "cocotbvhpi_" + sim_define.lower()),
```
eric-wieser, line 210: Presumably these no longer end up in subdirectories? If not, where are they installed to?

themperek:
Travis: https://travis-ci.org/github/themperek/cocotb/builds/681357094 (You should write to support or something to fix Travis for you, other way will be difficult)

BTW: This would be the first external dependency in cocotb.

cocotb_build_libs.py
mtikekar: They do end up in the same path. setuptools_dso converts from `cocotb.libs.name` to `cocotb/libs/libname.so`, `cocotb/libs/name.dll`, `cocotb/libs/libname.dylib`

eric-wieser: Mind adding a comment to that effect in `dso_join`?

eric-wieser:
Was vaguely hopeful that would awake travis

ktbarrett:
@eric-wieser That's a little too brief, could you explain a bit better?

>Yes, it works on windows too,

Make or CMake wouldn't work on Windows?

> and integrates with distutils.

That seems like circular logic?

mtikekar:
I've emailed Travis support. Let's see. I tried manually triggering a build on my fork but that gave an "Abuse detected" as well.

eric-wieser:
> Make or CMake wouldn't work on Windows?

I suppose I'm thinking further down the line to MSVC support, which ideally we would have. In that situation, most windows users will not have make or cmake installed anywhere -

CMake was something I considered with #966.

> That seems like circular logic?

Well, the main logic is that `distutils` is the best tool for the `simulator` module, so we need it anyway. Once we have that constraint, the fewer extra install steps a user needs, the easier it would be to install. From a technical level, a build dependency is zero-cost, because pip handles it for users.

Of course, you're right that a dependency presents a bureaucratic obstacle.

mtikekar:
> Well, the main logic is that `distutils` is the best tool for the `simulator` module, so we need it anyway

It is possible to build `simulator` with make: http://notes.secretsauce.net/notes/2017/11/14_python-extension-modules-without-setuptools-or-distutils.html

I would have also preferred Make or CMake. On Windows, there is an official anaconda [package](https://anaconda.org/anaconda/cmake) for CMake.

themperek:
Does this solve/improve any problem we have or is it just nice to have? We may consider some threshold for adding a dependency. 

PS. As far as I can see we may consider a dependency on pybin11 at some point.

imphil:
Thanks for your contribution @mtikekar!

Closing as stale. We have a working build solution for now. It's not overly beautiful, but does the trick. This PR will be around in case we hit fresh limitations and need to re-think our approach.

PR #1740, opened 2020-04-30
Fix mingw32 debug build

marlonjames:
In order to build extensions with debug symbols using the `mingw32`
compiler, the `--debug` option for `setuptools` is required (otherwise
the debug symbols are always stripped).

(See https://github.com/python/cpython/blob/3.5/Lib/distutils/cygwinccompiler.py#L234-L241)

Additionally, for non-MSVC Windows compilers, the Python DLL is always
added to the end of the list of libraries for an `Extension`, as it is
needed in certain instances where it is not needed on other platforms
(i.e., `libgpilog`).

When using the `--debug` option, the compiler also assumes a debug
version of Python, which on Windows has a "_d" suffix in the DLL name
(i.e., python37_d.DLL). When linking, `distutils` tries to link against
this debug DLL.

(See https://github.com/python/cpython/blob/3.5/Lib/distutils/command/build_ext.py#L705-L715)

To build cocotb with debug symbols and link to a non-debug Python,
the "_d" suffix must be removed.

`--no-use-pep517` is also required in order for `--global-option --debug` to be passed to `setup.py`.

Used as:
`pip install --no-use-pep517 --global-option build_ext --global-option --compiler=mingw32 --global-option --debug --global-option --nondebug-python cocotb`

---

I tested locally in Conda with Python 3.7. I confirmed no `-s` was passed to gcc when linking.

### Non-debug libcocotb.dll:
size: 22528
```
$ objdump -s libcocotb.dll | rg section
Contents of section .text:
Contents of section .data:
Contents of section .rdata:
Contents of section .eh_fram:
Contents of section .edata:
Contents of section .idata:
Contents of section .CRT:
Contents of section .tls:
Contents of section .reloc:
```

### Debug libcocotb.dll:
size: 116833
```
$ objdump -s libcocotb.dll | rg section
Contents of section .text:
Contents of section .data:
Contents of section .rdata:
Contents of section .eh_frame:
Contents of section .edata:
Contents of section .idata:
Contents of section .CRT:
Contents of section .tls:
Contents of section .reloc:
Contents of section .debug_aranges:
Contents of section .debug_info:
Contents of section .debug_abbrev:
Contents of section .debug_line:
Contents of section .debug_str:
Contents of section .debug_loc:
Contents of section .debug_ranges:
```

---

- [ ] Document on [wiki](https://github.com/cocotb/cocotb/wiki/Tier-2-Setup-Instructions)

### DIFF
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -31,6 +31,17 @@ def _get_lib_ext_name():
 
 
 class build_ext(_build_ext):
+
+    user_options = _build_ext.user_options + [
+        ('nondebug-python', None, 'link to non-debug Python on Windows (used with --debug)'),
+    ]
+
+    boolean_options = _build_ext.boolean_options + ['nondebug-python']
+
+    def initialize_options(self):
+        _build_ext.initialize_options(self)
+        self.nondebug_python = None
+
     def run(self):
 
         def_dir = os.path.join(cocotb_share_dir, "def")
@@ -69,6 +80,23 @@ class build_ext(_build_ext):
         filename_short = filename_short.replace("libcocotbvpi_icarus.dll", "libcocotbvpi_icarus.vpl")
 
         return filename_short
+
+    def get_libraries(self, ext):
+        """
+        Call base-class and remove any ``_d`` suffix from Python DLL on Windows.
+        """
+        libs = _build_ext.get_libraries(self, ext)
+
+        if os.name == "nt" and self.nondebug_python:
+            # pythonXX_d should be at the end for non-MSVC debug builds,
+            # but check the whole list in case this changes.
+            for idx, lib in enumerate(libs):
+                if lib.startswith("python"):
+                    if lib.endswith("_d"):
+                        libs[idx] = lib[:-2]
+                        break
+
+        return libs
 
     def finalize_options(self):
         """ Like the base class method,but add extra library_dirs path. """


garmin-mjames:
This conflicts with #1739.
Looking at `setuptools_dso`, it doesn't support debug builds through `--debug` option to `setup.py`.
I didn't look close enough to see if there's a different way to get debug builds on Windows.

themperek:
@garmin-mjames This does not work https://www.idiap.ch/software/bob/docs/bob/bob.extension/v2.5.0/pip.html#developing-a-c-package ?

garmin-mjames:
I tried using CFLAGS and CXXFLAGS, but https://github.com/python/cpython/blob/3.5/Lib/distutils/cygwinccompiler.py#L234-L241 always adds `-s` to the link step with strips the symbols, even if `-g` is also provided.

garmin-mjames:
Note that this change breaks building against an actual debug build of Python on Windows.

`libcocotb` explicitly links against the Python DLL with `_get_python_lib_link()`, and I don't know enough to tell whether doing that breaks it already.

themperek:
I am for it but would like to know @eric-wieser opinion.

I would suggest to add a line of documentation so this knowledge does not get lost.

garmin-mjames:
I was going to put it on the wiki page https://github.com/cocotb/cocotb/wiki/Tier-2-Setup-Instructions, unless we want to add a documentation section on debugging / developing.

tpambor:
Can this be put behind an option? It does not seem favorable to me to break the way a user would normally expect to create a debug build.

Should be easy to add a flag like this 
```
user_options = _build_ext.user_options + [
    ('debug-only-cocotb', None, 'link cocotb with non-debug build on Windows'),
]

boolean_options = _build_ext.boolean_options + ['debug-only-cocotb']

def initialize_options(self):
    self.debug_only_cocotb = False
    _build_ext.initialize_options(self)
```
and then check `self.debug_only_cocotb` in `get_libraries`

garmin-mjames:
@tpambor I went down that path originally, but we only support Conda on Windows.
That looks like that may be changing soon, though.

garmin-mjames:
Anything else needed for this? I can add documentation to the https://docs.cocotb.org/en/latest/install_devel.html page, or put it somewhere on the wiki.

eric-wieser:
Is there any reason we would want `--debug-only-cocotb` vs just having a debug build of both?

garmin-mjames:
@eric-wieser It takes some extra work and manual steps to get a debug Python build into Conda.

It's nice to be able to debug the GPI layer in a mingw32 environment without that extra effort. So far I haven't needed to step into the Python interpreter.

Because cocotb is using `setuptools` to build the non-extension libraries, this is needed to keep the debug info from being stripped.

eric-wieser:
This is starting to sound like a bug in `distutils`. Can we report it upstream? I'm not saying we need to wait for a fix upstream, but it sounds like something we should check the intended behavior of.

For comparison, most platforms essentially use `sys.abiflags` to decide whether to link debug python

https://github.com/python/cpython/blob/55a6a16a46239a71b635584e532feb8b17ae7fdf/Lib/distutils/command/build_ext.py#L747-L755

Ie, if you want a build _against_ debug python, you use `python_d setup.py install` instead of `python setup.py install`.

garmin-mjames:
@eric-wieser The only reason the `--debug` flag needs to be used is because the mingw compiler strips debug otherwise: https://github.com/python/cpython/blob/3.5/Lib/distutils/cygwinccompiler.py#L234-L241.

It's been that way since at least Python 2.7

eric-wieser:
Sorry, to be clear I now think it's a bug that `--debug` causes `python35_d` to be linked. On linux, that flag is ignored, and the dll is chosen based on whether the python used to build it is in debug mode.

imphil:
@marlonjames Is this PR still relevant? Probably a bit, but maybe not worth the hassle?

marlonjames:
We can close this and just accept the limitation.

imphil:
Let's close it and hope that mingw32 just dies out. Otherwise we can come back to this PR :-)

PR #1741, opened 2020-04-30
Use raw string for docstring with escape codes

cmarqu:


eric-wieser:
Where's the escape code?

cmarqu:
Far down, there is a ``\e``.

Issue #913, opened 2019-05-04
Assigning Python ints to signals >32 bits


ktbarrett:
If I assign a signed value to a signal larger than 32 bits wide, cocotb attempts to convert the int into an unsigned little-endian BinaryValue before assigning the value ([code](https://github.com/potentialventures/cocotb/blob/master/cocotb/handle.py#L592)). When it receives my negative value, the `_convert_to_unsigned` method fails and gives me a totally useless stacktrace. I know to look for this now, but wasted my time until I figured this out; and lucky for me I could, others may not be so lucky.

This convenience conversion makes a poor assumption rather than informing the user it doesn't have enough information to continue. This is bad coding practice. Instead, it might be better to give a warning or throw an error stating that you cannot assign Python ints to signals greater than 32 bit wide, but instead it will only take a BinaryValue for such signals.

```python
width = len(dut.data)   # dut.data begin a 36 bit width sfixed
for i in range(n):
    yield RisingEdge(dut.clk)
    dut.data <= randint(-(2**(width-1)), (2**(width-1))-1)
```

elgorwi:
Fixed by #2316

Issue #268, opened 2015-07-22
32-bit signals are driven by binary string and not the integer value


leftink:
In handle.py, the function setimmediatevalue() will check if the value is an integer and then check if the integer is less than 0x7fffffff and then it will drive the simulator quicker if it is.  It seems like any 32-bit value should be able to be driven in the same manner.  I realize that that if the value uses the most significant bit, then the C-function, will treat that value as a negative integer, but from the code I have looked at, there wouldn't be any impact to the way the signal is driven.

@chiggs Is there a reason that any 32-bit and below values don't use the faster interface, instead of 31-bit and below?  It seems like the C-Interface would support this.

chiggs:
It likely stems from the VPI definition of a value, which by the specification is an `int32_t`.

It would be difficult to prove that all simulators treated this as `uint32_t`, however VHPI uses `uint32_t` and FLI arguably doesn't matter - we should even default to `uint64_t` before switching to strings.

It would be nice if in Python, we determined the threshold for switching to string-based slow path when the signal object is discovered, based on some constant we could read from the interface.

stuarthodgson:
We can do that with a capabilities method that python can call. Would be extensible and www can have some sensible defaults that are inplementation dependent.

imphil:
The VPI definition of int32_t comes most likely from the fact that int/integer in SystemVerilog are defined as signed 32 bit type. So the `IntegerObject` should at least support the negative range as well.

However, as far as I could see, the `unsigned` qualifier is not reflected in VPI; both `logic int unsigned something` and `logic int something` are most likely mapped to `vpiIntegerNet`. This means we can safely cast uint32 to int32 and set it through `simulator.set_signal_val_long()`.

Needs more research, but could be a very nice performance win for cocotb.

elgorwi:
Fixed by #2316

PR #1742, opened 2020-05-01
Closes #1202: set IntegerObject using 32-bit set_signal_value_int().

### LINKED ISSUE #1202: IntegerObject: Document and ensure its 32 bit
`IntegerObject` is used to set int/integer nets as defined in SystemVerilog, which are 32 bit (signed) values. See https://github.com/cocotb/cocotb/issues/268 for a discussion on signed-ness.

- `IntegerObject` doesn't check the size, which could lead to loss of data.
- We need to document that `IntegerObject` is 32 bit.
- Our backend functionality, starting with `simulator.set_signal_val_long`, uses long instead of int as data type and in the function names. That should be cleaned up to say int.

elgorwi:
- Change gpi_set_signal_value_long(long) to gpi_set_signal_value_int(int32_t)
- Set IntegerObject using set_signal_value_int().  
- Check if value for IntegerObject is in 32-bit range and raise OverflowError for out of range values
- Add tests for the supported range and out of range
- Fixes a compilation warning on 32-bit platforms.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -31,6 +31,7 @@
 
 import ctypes
 import warnings
+import enum
 
 import cocotb
 from cocotb import simulator
@@ -40,6 +41,11 @@ from cocotb.result import TestError
 
 # Only issue a warning for each deprecated attribute access
 _deprecation_warned = set()
+
+
+class _Limits(enum.IntEnum):
+    NONE         = 0
+    SIGNED_32BIT = 1  # -2**(Nbits - 1) <= value <= 2**(Nbits - 1)  where Nbits = 32
 
 
 class SimHandleBase:
@@ -651,7 +657,7 @@ class ModifiableObject(NonConstantObject):
         value, set_action = self._check_for_set_action(value)
 
         if isinstance(value, int) and value < 0x7fffffff and len(self) <= 32:
-            call_sim(self, self._handle.set_signal_val_long, set_action, value)
+            call_sim(self, self._handle.set_signal_val_int, set_action, value, int(_Limits.NONE))
             return
         if isinstance(value, ctypes.Structure):
             value = BinaryValue(value=cocotb.utils.pack(value), n_bits=len(self))
@@ -755,7 +761,7 @@ class EnumObject(ModifiableObject):
                 "Unsupported type for enum value assignment: {} ({!r})"
                 .format(type(value), value))
 
-        call_sim(self, self._handle.set_signal_val_long, set_action, value)
+        call_sim(self, self._handle.set_signal_val_int, set_action, value, int(_Limits.NONE))
 
     @ModifiableObject.value.getter
     def value(self) -> int:
@@ -790,7 +796,7 @@ class IntegerObject(ModifiableObject):
                 "Unsupported type for integer value assignment: {} ({!r})"
                 .format(type(value), value))
 
-        call_sim(self, self._handle.set_signal_val_int, set_action, value)
+        call_sim(self, self._handle.set_signal_val_int, set_action, value, int(_Limits.SIGNED_32BIT))
 
     @ModifiableObject.value.getter
     def value(self) -> int:
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -777,6 +777,9 @@ class IntegerObject(ModifiableObject):
         Raises:
             TypeError: If target has an unsupported type for
                  integer value assignment.
+
+            OverflowError: If value is out of range for assignment
+                 of 32-bit IntegerObject.
         """
         value, set_action = self._check_for_set_action(value)
 
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -787,7 +787,7 @@ class IntegerObject(ModifiableObject):
                 "Unsupported type for integer value assignment: {} ({!r})"
                 .format(type(value), value))
 
-        call_sim(self, self._handle.set_signal_val_long, set_action, value)
+        call_sim(self, self._handle.set_signal_val_int, set_action, value)
 
     @ModifiableObject.value.getter
     def value(self) -> int:
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -617,6 +617,19 @@ static PyObject *set_signal_val_real(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject
     Py_RETURN_NONE;
 }
 
+static PyObject *set_signal_val_int(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject *args)
+{
+    int value;
+    gpi_set_action_t action;
+
+    if (!PyArg_ParseTuple(args, "ii:set_signal_val_int", &action, &value)) {
+        return NULL;
+    }
+
+    gpi_set_signal_value_int(self->hdl, static_cast<int32_t>(value), action);
+    Py_RETURN_NONE;
+}
+
 static PyObject *set_signal_val_long(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject *args)
 {
     long long value;
@@ -955,6 +968,13 @@ static PyMethodDef gpi_sim_hdl_methods[] = {
             "set_signal_val_long($self, action, value, /)\n"
             "--\n\n"
             "Set the value of a signal using a long"
+        )
+    },
+    {"set_signal_val_int",
+        (PyCFunction)set_signal_val_int, METH_VARARGS, PyDoc_STR(
+            "set_signal_val_int($self, action, value, /)\n"
+            "--\n\n"
+            "Set the value of a signal using an int"
         )
     },
     {"set_signal_val_str",
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -156,6 +156,11 @@ struct sim_time {
 
 static struct sim_time cache_time;
 
+enum class limits_e : int {
+    NONE = 0,
+    SIGNED_32BIT = 1
+};
+
 /**
  * @name    Callback Handling
  * @brief   Handle a callback coming from GPI
@@ -622,29 +627,19 @@ static PyObject *set_signal_val_int(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject 
 {
     long long value;
     gpi_set_action_t action;
+    limits_e limits;
 
-    if (!PyArg_ParseTuple(args, "iL:set_signal_val_int", &action, &value)) {
+    if (!PyArg_ParseTuple(args, "iLi:set_signal_val_int", &action, &value, &limits)) {
         return NULL;
     }
 
-    if (value < std::numeric_limits<int32_t>::min() ||
-        value > std::numeric_limits<int32_t>::max()) {
-        const char *m_name = gpi_get_signal_name_str(self->hdl);
-        PyErr_Format(PyExc_OverflowError, "Value (%lld) out of range for assignment of 32-bit integer signal (%s)", value, m_name);
-        return NULL;
-    }
-
-    gpi_set_signal_value_int(self->hdl, static_cast<int32_t>(value), action);
-    Py_RETURN_NONE;
-}
-
-static PyObject *set_signal_val_long(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject *args)
-{
-    long long value;
-    gpi_set_action_t action;
-
-    if (!PyArg_ParseTuple(args, "iL:set_signal_val_long", &action, &value)) {
-        return NULL;
+    if (limits == limits_e::SIGNED_32BIT) {
+        if (value < std::numeric_limits<int32_t>::min() ||
+            value > std::numeric_limits<int32_t>::max()) {
+            const char *m_name = gpi_get_signal_name_str(self->hdl);
+            PyErr_Format(PyExc_OverflowError, "Value (%lld) out of range for assignment of 32-bit signed integer signal (%s)", value, m_name);
+            return NULL;
+        }
     }
 
     gpi_set_signal_value_int(self->hdl, static_cast<int32_t>(value), action);
@@ -971,16 +966,9 @@ static PyMethodDef gpi_sim_hdl_methods[] = {
             "Get the value of a signal as a double precision float"
         )
     },
-    {"set_signal_val_long",
-        (PyCFunction)set_signal_val_long, METH_VARARGS, PyDoc_STR(
-            "set_signal_val_long($self, action, value, /)\n"
-            "--\n\n"
-            "Set the value of a signal using a long"
-        )
-    },
     {"set_signal_val_int",
         (PyCFunction)set_signal_val_int, METH_VARARGS, PyDoc_STR(
-            "set_signal_val_int($self, action, value, /)\n"
+            "set_signal_val_int($self, action, value, limits, /)\n"
             "--\n\n"
             "Set the value of a signal using an int"
         )
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -42,6 +42,7 @@ static int sim_ending = 0;
 #include "simulatormodule.h"
 #include <cocotb_utils.h>     // COCOTB_UNUSED
 #include <type_traits>
+#include <limits>
 
 
 /* define the extension types as templates */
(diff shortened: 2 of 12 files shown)


cocotb/handle.py
```diff
@@ -642,7 +642,7 @@ def _set_value(self, value, call_sim):
         value, set_action = self._check_for_set_action(value)
 
         if isinstance(value, int) and value < 0x7fffffff and len(self) <= 32:
```
ktbarrett, line 644: This line was never correct and is even less correct now.
```suggestion
        if isinstance(value, int) and len(self) <= 32:
            if value < min_value(len(self)) or max_value(len(self)) < value:
                # throw or truncate. What would BinaryValue do in this case?
```

elgorwi:
> I thought the consensus was that we should be using 32 bit ints? Your PR also states 32 bit ints. But you've changed everything to `int64_t`?

I've changed the ```set_signal_value_long()``` value type from 'long' to 'int64_t', to get rid of a compilation warning on 32-bit platform. This should ensure similar behaviour on 32-bit vs 64-bit platforms. In addition, I think we want to use int64_t for value to allow setting both int32_t and uint32_t ranges from a python int (See #268). Changing ```gpi_set_signal_value_long()``` to use either int32_t or uint32_t seems a bit more involved and something I would prefer to do in a follow-up if needed. 
I've introduced ```set_signal_value_int()```, which uses an int32_t to set IntegerObject. It does use ```gpi_set_signal_value_long()``` under the hood.

cocotb/handle.py
elgorwi: Yes, this is something I was planning to tackle next (together with #913). 
Don't agree with the 64-bit comment as we are limited by the (32-bit) integer data width of ```vpi_put_value()```.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -491,7 +491,7 @@ int VhpiLogicSignalObjHdl::set_signal_value(long value, gpi_set_action_t action)
         case vhpiLogicVecVal: {
             int i;
             for (i=0; i<m_num_elems; i++)
-                m_value.value.enumvs[m_num_elems-i-1] = value&(1L<<i) ? vhpi1 : vhpi0;
+                m_value.value.enumvs[m_num_elems-i-1] = value&(1LL<<i) ? vhpi1 : vhpi0;
```
eric-wieser, line 494: Strictly speaking this should be
```suggestion
                m_value.value.enumvs[m_num_elems-i-1] = value&(UINT64_C(1)<<i) ? vhpi1 : vhpi0;
```
But it doesn't matter anyway.

cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -325,11 +325,11 @@ long VpiSignalObjHdl::get_signal_value_long()
 }
 
 // Value related functions
-int VpiSignalObjHdl::set_signal_value(long value, gpi_set_action_t action)
+int VpiSignalObjHdl::set_signal_value(int64_t value, gpi_set_action_t action)
 {
     s_vpi_value value_s;
 
-    value_s.value.integer = (int)value;
+    value_s.value.integer = static_cast<PLI_INT32>(value);
```
eric-wieser, line 332: A little worried about silent truncation here.

elgorwi: Actually, this should not be a problem as we are only expecting 32-bit values here.
I'm working on adding some proper value range checking in C++, which will ensure that we never overflow silently. However, I think this is more related to #268 rather than #1202, so unless you prefer another commit in this PR, I will address this in a separate PR together with the change to handle.py as suggested by @k tbarrett in the comment above.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -627,12 +627,25 @@ static PyObject *set_signal_val_real(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject
...
+}
+
 static PyObject *set_signal_val_long(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject *args)
 {
-    long value;
+    int64_t value;
     gpi_set_action_t action;
 
-    if (!PyArg_ParseTuple(args, "il:set_signal_val_long", &action, &value)) {
+    if (!PyArg_ParseTuple(args, "iL:set_signal_val_long", &action, &value)) {
```
eric-wieser, line 648: Just to be safe:
```suggestion
    static_assert(sizeof(value) == sizeof(long long), "L is not the right format character on this platform");
    if (!PyArg_ParseTuple(args, "iL:set_signal_val_long", &action, &value)) {
```

ktbarrett:
@elgorwi I think you are getting caught about the fact that `vhpiIntT` is unsigned. The 2008 standard states:

> 5.2.3.2
The only predefined integer type is the type INTEGER. The range of INTEGER is implementation
dependent, but it is guaranteed to include the range –2147483647 to +2147483647. It is defined with an ascending range.

The definitions of `NATURAL` and `POSITIVE` are *not* unsigned 32 bit integers.

> subtype NATURAL is INTEGER range 0 to INTEGER'HIGH;
> subtype POSITIVE is INTEGER range 1 to INTEGER'HIGH;

The VHDL 2008 standard states for `vhpiIntT`:

> 22.2.3
A value of type vhpiIntT or vhpiLongIntT represents a value of a VHDL integer type. A value of
type vhpiIntT shall be represented with at least 32 bits, and a value of type vhpiLongIntT shall be
represented with at least 64 bits. The value represented by a given value of either type is the given value, interpreted as a signed twos-complement binary number.

The older standard is under-specified and simply states that INTEGER is implementation defined, and that `vhpiIntT` should be able to represent all possible VHDL integer values, but the definitions of `NATURAL` and `POSITIVE` are the same.

This all points to the realization that `vhpiIntT` *shouldn't* have been unsigned. I think `gpi_set_signal_value_int` should use 32 bit signed integers. You should just `reinterpret_cast` to the appropriate signedness on VHPI. We should test this works in all simulators that support VHPI, but it is sound. That also means you don't have to worry about handling overflow in the implementation.

elgorwi:
Thanks for investigating vhpi @ktbarret. I don’t have access to a vhpi simulator, so I only looked at vhpi_user.h
Did a bit of googling and found another vhpi_user.h:
https://opensource.ieee.org/vasg/Packages/blob/a03b96d46e0fa28df73b7e259861a5ff596c23a5/code/vhpi_user.h
```
typedef int32_t vhpiIntT;
typedef int64_t vhpiLongIntT;
```
These typedefs seem to make a lot more sense and confirm what you have found. 
Perhaps our vhpi_user.h should be updated to the correct types?

themperek:
One more vhpi_user.h: https://github.com/jamiekang/embedded-sw-dev-tools/blob/master/binsim/vhpi_user.h

ktbarrett:
Is there some reason that vhpi_user.h isn't something that is provided by the standards group? I'm sure there is one that is "correct".

garmin-mjames:
Here's the official version from VHDL-2019:
https://opensource.ieee.org/vasg/Packages/-/blob/release/code/vhpi_user.h

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
elgorwi: Thanks, I didn't know about those macros

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -627,16 +627,29 @@ static PyObject *set_signal_val_real(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject
...
+    long long value;
     gpi_set_action_t action;
 
-    if (!PyArg_ParseTuple(args, "il:set_signal_val_long", &action, &value)) {
+    if (!PyArg_ParseTuple(args, "iL:set_signal_val_long", &action, &value)) {
         return NULL;
     }
 
-    gpi_set_signal_value_long(self->hdl, value, action);
+    gpi_set_signal_value_int(self->hdl, static_cast<int32_t>(value), action);
```
eric-wieser, line 652:
```suggestion
    if (value < std::numeric_limits<int32_t>::min()) {
        PyErr_SetString(PyExc_OverflowError, "value is less than minimum int32");
        return NULL;
    }
    else if (value > std::numeric_limits<int32_t>::max()) {
        PyErr_SetString(PyExc_OverflowError, "value is greater than maximum int32");
        return NULL;
    }
    gpi_set_signal_value_int(self->hdl, static_cast<int32_t>(value), action);
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -897,6 +910,9 @@ static PyMethodDef gpi_sim_hdl_methods[] = {
     {"get_signal_val_real",
         (PyCFunction)get_signal_val_real, METH_NOARGS,
         "Get the value of a signal as a double precision float"},
+    {"set_signal_val_int",
+        (PyCFunction)set_signal_val_int, METH_VARARGS,
+        "Set the value of a signal using an int"},
```
eric-wieser, line 915: Is this needed any more? Can we rename the other one to `int32`?

cocotb/handle.py
```diff
@@ -47,6 +47,8 @@
 # Only issue a warning for each deprecated attribute access
 _deprecation_warned = set()
 
+_INT_MAX = 2**31-1
+_INT_MIN = -2**31
```
eric-wieser, line 51:
```suggestion
_INT_RANGE = range(-2**31, 2**31)
```

cocotb/handle.py
```diff
@@ -778,7 +783,12 @@ def _set_value(self, value, call_sim):
                 "Unsupported type for integer value assignment: {} ({!r})"
                 .format(type(value), value))
 
-        call_sim(self._handle.set_signal_val_long, set_action, value)
+        if value < _INT_MIN or value > _INT_MAX:
```
eric-wieser, line 786: Tempting to just handle this in C, where we actually require this to be true.

elgorwi:
> 
> 
> > I thought the consensus was that we should be using 32 bit ints? Your PR also states 32 bit ints. But you've changed everything to `int64_t`?
> 
> I've changed the `set_signal_value_long()` value type from 'long' to 'int64_t', to get rid of a compilation warning on 32-bit platform. This should ensure similar behaviour on 32-bit vs 64-bit platforms. In addition, I think we want to use int64_t for value to allow setting both int32_t and uint32_t ranges from a python int (See #268). Changing `gpi_set_signal_value_long()` to use either int32_t or uint32_t seems a bit more involved and something I would prefer to do in a follow-up if needed.
> I've introduced `set_signal_value_int()`, which uses an int32_t to set IntegerObject. It does use `gpi_set_signal_value_long()` under the hood.

I've changed to ```gpi_set_signal_value_int(int32_t)```.

cocotb/share/lib/simulator/simulatormodule.cpp
elgorwi: Nice! I was just wondering how to do this exactly from C++.

cocotb/handle.py
elgorwi: Will do

cocotb/share/lib/simulator/simulatormodule.cpp
elgorwi: @eric-wieser Actually, I think we want to keep both functions around to do different range checks. For #268 we want to support both int32_t range and uint32_t range from set_signal_value_long. I was thinking about renaming set_signal_value_long to set_signal_value_int_vec to distinguish it better from set_signal_value_int, but I'm open to suggestions.

eric-wieser: Are you actually after `set_signal_val_uint32` and `set_signal_val_int32`?

elgorwi: ```set_signal_val_int32``` for setting integerObjects and ```set_signal_val_uint32_int32``` for setting bitvectors up to 32 bit.

tests/test_cases/issue_134/test_reals.py
```diff
@@ -1,43 +0,0 @@
-import logging
```
eric-wieser, line 1: It would be good if these tests could be moved in a standalone PR, rather than being muddled with the rest of this patch. Would you mind filing a separate PR for that first?

elgorwi: Ok, no problem

cocotb/share/lib/simulator/simulatormodule.cpp
elgorwi: @eric-wieser Changed everything to use ```set_signal_val_int()``` and introduced a 'limits' parameter to set the desired limits of the range check. Updated PR #1744 to also us a 'limits' parameter.
Is there anything else that needs to be done in this PR?

elgorwi:
> So your plan now is to have a single function for handling setting signals with Python `int`s and pass another argument to decide whether to impose 32 bit signed boundaries or treat it as arbitrary length bit vector? I like that approach better, but I still think they are better off as 2 separate functions.

@ktbarrett The plan is to use ```set_signal_val_int()``` for setting values using Python ints to signals under 32-bit. So something like:
- IntegerObject -> limits=SIGNED_32BIT (PR #1742)
- ModifiableObject -> limits=VECTOR_NBIT (PR #1744)
- EnumObject -> limits=UNSIGNED_32BIT (TODO)

For signals larger than 32-bit we currently use ```BinaryValue```, but the performance is not great. For that reason I propose to convert the Python ```int``` to ```bytes```, using Python builtin ```int.to_bytes()``` which is faster than ```BinaryValue```. Then introduce another function ```set_signal_val_bytes()``` (PR #1766). In addition, ```int.to_bytes()``` also provides a convenient solution to transfer values larger than 64-bit from Python to C++.

ktbarrett:
The way I see it, you only need two functions: `set_signal_value_int` and `set_signal_value_long`. I'm borrowing the meanings of `int` and `long` from Python 2. 

`IntegerObject` only would use `set_signal_value_int` to set values and only check for incoming Python `int`s to be within the bounds of a signed 32 bit integer.
`EnumObject` would also only use `set_signal_value_int`. I'm not entirely sure what each language's standard guarantees (or what each simulator actually does), but there are no reasonable enumerated types with over 2 billion enumerations, but less than 4 billion enumerations, so worrying about the full range of an unsigned 32 bit value is silly.
`ModifiableObject` would use the other function which takes a Python `int` and check them against the signal bounds. Overflow would truncate, like it currently does. If the signal is less than 32 bits, it would use `gpi_set_signal_value_long` to set the signal as an attempt at improving performance. If the signal is over 32 bits it would encode the value as a binary string and use `gpi_set_signal_value_binstr`. There is no need to turn the value into `bytes` and create a special function to handle that, we can handle Python `int`s directly in C.

elgorwi:
> The way I see it, you only need two functions: `set_signal_value_int` and `set_signal_value_long`. I'm borrowing the meanings of `int` and `long` from Python 2.
> 
> *`IntegerObject` only would use `set_signal_value_int` to set values and only check for incoming Python `int`s to be within the bounds of a signed 32 bit integer.
> 
> *`EnumObject` would also only use `set_signal_value_int`. I'm not entirely sure what each language's standard guarantees (or what each simulator actually does), but there are no reasonable enumerated types with over 2 billion enumerations, but less than 4 billion enumerations, so worrying about the full range of an unsigned 32 bit value is silly.
> 
> *`ModifiableObject` would use the other function which takes a Python `int` and check them against the signal bounds. Overflow would truncate, like it currently does. If the signal is less than 32 bits, it would use `gpi_set_signal_value_long` to set the signal as an attempt at improving performance. If the signal is over 32 bits it would encode the value as a binary string and use `gpi_set_signal_value_binstr`. There is no need to turn the value into `bytes` and create a special function to handle that, we can handle Python `int`s directly in C.

Just trying to clear up some confusion here. Looking at the earlier comments above, I don't think it makes sense to keep ```set_signal_val_long()``` around, as ```gpi_set_signal_value_long(long)``` will be replaced by ```gpi_set_signal_value_int(int32_t)```. (the GPI backend is limited to 32-bit anyway, because of ```vpi_put_value(PLI_INT32)```)
So with the proposed solution in this PR, we can set any values up to 32-bit using ```set_signal_val_int()``` and do some specific range-checking depending on the object type in handle.py, by providing the appropriate limits parameter. (range checks are done in C++ for improved performance as recommended in earlier comments above).

Would be great to get this PR merged first, so that I can rebase and update the other open related PR's (#1792 and #1744) to avoid further confusion.   

When #1792 and #1744 will have been merged, we could achieve some further performance improvements setting values > 32-bit if desired. This would be using ```set_signal_val_bytes()``` (#1766), which needs the python int converted into a bytes object first, then transfers it to c++ as an array of char, converts it to a binstring and finally calls ```gpi_set_signal_value_binstr```. I've also been working on ```gpi_set_signal_value_hexstr``` as an alternative, but still need to benchmark everything in more detail, to show the exact performance improvement compared to the current solution. 

Regarding enums, from what I've seen at first glance, these should be restricted to positive values. With the proposed solution in this PR, that can be easily added by extending the _Limits class in handle.py and implementing the corresponding unsigned 32-bit range check in simulatormodule.cpp.

cocotb/handle.py
```diff
@@ -48,6 +48,11 @@
 _deprecation_warned = set()
 
 
+class _Limits:
```
eric-wieser, line 51:
```suggestion
class _Limits(enum.Enum):
```

eric-wieser:
I need to read this through a few times to get the full picture.

cocotb/handle.py
elgorwi: OK, done

elgorwi:
> 
> 
> Here's the official version from VHDL-2019:
> https://opensource.ieee.org/vasg/Packages/-/blob/release/code/vhpi_user.h
> 
> Edit: oops, same as above.
> 
> @ktbarrett It's Annex B in the 2008 standard.

This will be tracked under #1759

elgorwi:
@imphil can you please review. This closes an issue that you reported: #1202

Issue #1736, opened 2020-04-29
Error messages that don't hint at cocotb


cmarqu:
The strings in
https://github.com/cocotb/cocotb/blob/master/cocotb/share/lib/utils/cocotb_utils.cpp
do not say that they are coming from cocotb, but they should so that they can be differentiated from simulator messages.
Also, can e.g. "We are calling up again" get a bit more context?

ktbarrett:
I feel like we should make those functions macro functions and use the `LOG_*` functions instead of directly printing to stdut/stderr. I have that change hanging out in a branch.

cmarqu:
Fixed by #1743.

PR #1743, opened 2020-05-01
Use GPI log in cocotb_utils

ktbarrett:
`to_python` and `to_simulator` were updated to use the `LOG_*` macros. This required we change them to macro functions so we get a useful line number in the log message. Updated the dynamic loading helpers to use the `LOG_*` macros as well.

xref #1736.

### DIFF
--- a/cocotb/share/include/cocotb_utils.h
+++ b/cocotb/share/include/cocotb_utils.h
@@ -30,6 +30,8 @@
 #ifndef COCOTB_UTILS_H_
 #define COCOTB_UTILS_H_
 
+#include <gpi_logging.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -42,8 +44,26 @@ extern void* utils_dyn_sym(void *handle, const char* sym_name);
 
 extern int is_python_context;
 
-void to_python(void);
-void to_simulator(void);
+// to_python and to_simulator are implemented as macros instead of functions so
+// that the logs reference the user's lineno and filename
+
+#define to_python() do { \
+    if (is_python_context) { \
+        LOG_ERROR("FATAL: We are calling up again"); \
+        exit(1); \
+    } \
+    ++is_python_context; \
+    LOG_DEBUG("Returning to Python"); \
+} while (0)
+
+#define to_simulator() do { \
+    if (!is_python_context) { \
+        LOG_ERROR("FATAL: We have returned twice from python\n"); \
+        exit(1); \
+    } \
+    --is_python_context; \
+    LOG_DEBUG("Returning to simulator"); \
+} while (0)
 
 #define COCOTB_UNUSED(x) ((void)x)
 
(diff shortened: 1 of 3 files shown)


cmarqu:
Sorry, I am not knowledgeable enough to review this.

ktbarrett:
If linking `gpilog` into `cocotb_utils` is a problem, this change can wait until I bring in my change to the GPI logging interface. The change puts Python logging code in the simulator module and adds an implementation agnostic logger registry interface to the GPI.

cocotb_build_libs.py
```diff
@@ -218,7 +219,6 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
     libgpilog = Extension(
         os.path.join("cocotb", "libs", "libgpilog"),
         include_dirs=[include_dir],
-        libraries=["cocotbutils"],
```
eric-wieser, line 221: Can you explain why this was here before? Would it have worked fine without it?

cocotb_build_libs.py
```diff
@@ -268,7 +268,7 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
         extra_link_args=_extra_link_args(rpath="$ORIGIN/libs"),
     )
 
-    return [libcocotbutils, libgpilog, libcocotb, libgpi, libsim]
```
eric-wieser, line 271: If this order matters, a comment saying why it matters would be nice.

cocotb_build_libs.py
ktbarrett: Yeah, there is no actual dependency there. Not sure why it was there to begin with. Likely we just missed it when we moved to distutils, and it was there before.

cocotb_build_libs.py
ktbarrett: cocotbutils depends on gpilog, so gpilog must be built first. Libraries are built in order.

cocotb/share/include/cocotb_utils.h
```diff
@@ -42,8 +44,23 @@ extern void* utils_dyn_sym(void *handle, const char* sym_name);
...
+} while (0)
+
+#define to_simulator() do { \
+    if (!is_python_context) { \
+        LOG_ERROR("FATAL: We have returned twice from python\n"); \
+        exit(1); \
+    } \
+    --is_python_context; \
+    LOG_DEBUG("Returning to simulator"); \
+} while (0)
```
eric-wieser, line 63: I assume these are macros so that the filename and line number is more useful in the log message? If so, can you add a comment explaining that?

eric-wieser:
@jhelie, any idea why the LGTM checkout is failing, without an error message? (pinging you because I think I remember you being helpful with LGTM issues in numpy/numpy in the past, let me know if I should ask elsewhere for support)

cocotb/share/include/cocotb_utils.h
```diff
@@ -42,8 +44,26 @@ extern void* utils_dyn_sym(void *handle, const char* sym_name);
 
 extern int is_python_context;
 
-void to_python(void);
-void to_simulator(void);
+// to_python and to_simulator are implemented as macros functions so that the
+// logs reference the user's lineno and filename
```
eric-wieser, line 48:
```suggestion
// to_python and to_simulator are implemented as macros instead of functions so
// that the logs reference the user's lineno and filename
```

PR #1744, opened 2020-05-01
Closes #268: allow using full 32-bit range of set_signal_val_long()

### LINKED ISSUE #268: 32-bit signals are driven by binary string and not the integer value

elgorwi:
- For signals up to 32-bit wide, set_signal_val_long() can be used to set the full signed ```(-2**31 to 2**31-1)``` and unsigned ```(0 - 2**32-1)``` data range using a python integer value.
- Added a check for max width of 32 bits and a signal width dependent overflow check to set_signal_val_long().
- Added some tests.
- This should improve performance of setting values > 2**31-1 on 32-bit signals

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -48,6 +48,12 @@ from cocotb.result import TestError
 _deprecation_warned = set()
 
 
+class _Limits:
+    NONE         = 0
+    SIGNED_32BIT = 1  # -2**(Nbits - 1) <= value <= 2**(Nbits - 1)  where Nbits = 32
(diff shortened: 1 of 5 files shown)


elgorwi:
This PR closes #268 , it has been tested with modelsim. @imphil can you review this?

PR #1747, opened 2020-05-01
Add info to CONTRIBUTING.md that examples should get the CC0/PD SPDX header

cmarqu:
See also https://github.com/cocotb/cocotb/pull/1620#issuecomment-622070911


imphil:
Thanks!

PR #1750, opened 2020-05-02
Check for Python dynamic library on setup

### LINKED ISSUE #1618: Check for `--enable-shared`
Should check (in `setup.py`?) if Python was compiled with `--enable-shared` and error if this is not the case.

It looks like on Linux (ubuntu, centos, conda) this is consistent and can be checked with `python -c "import sysconfig; print(sysconfig.get_config_vars('LDSHARED'))"`. Mind that `Py_ENABLE_SHARED ` is not consistant and does not work with conda https://github.com/conda/conda-build/issues/2738

On Windows is not consistent (VS compiler?). I would expect on Windows nobody builds its own python binary.

Do not know on osx.

themperek:
Cocotb requires to link Python dynamic library.

On Windows, the Python library is always a dynamic (an import library).
Python must be compiled with the --enable-shared flag on Linux
or the --enable-framework flag on Mac.

Closes #1618

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,9 @@ from distutils import sysconfig
 if sys.platform == "posix":
     config_args = sysconfig.get_config_vars().get("CONFIG_ARGS", "")
     if "--enable-framework" not in config_args or "--enable-shared" not in config_args:
-        raise SystemExit("ERROR: You must use a Python compiled with --enable-shared or --enable-framework.\nPython executable location is %s" % sys.executable)
+        raise SystemExit(
+            "ERROR: You must use a Python compiled with --enable-shared or --enable-framework.\n"
+            "Python executable location is {}".format(sys.executable))
 
 import logging
 from setuptools import setup
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,18 @@ if sys.version_info[:2] < (3, 5):
 
     raise SystemExit("\n".join(msg))
 
+
+# Cocotb requires linking to a dynamic Python library.
+# On Windows, the Python library is always dynamic (an "Import Library").
+# Python must be compiled with the `--enable-shared` flag on Linux and macOS
+# and the `--enable-framework` flag on macOS.
+from distutils import sysconfig
+
+if sys.platform == "posix":
+    config_args = sysconfig.get_config_vars().get("CONFIG_ARGS", "")
+    if "--enable-framework" not in config_args or "--enable-shared" not in config_args:
+        raise SystemExit("ERROR: You must use a Python compiled with --enable-shared or --enable-framework.\nPython executable location is %s" % sys.executable)
+
 import logging
 from setuptools import setup
 from setuptools import find_packages


setup.py
```diff
@@ -47,6 +47,20 @@
 
     raise SystemExit("\n".join(msg))
 
+
+# Cocotb requires to link Python dynamic library.
+# On Windows, the Python library is always a dynamic (an import library).
+# Python must be compiled with the --enable-shared flag on Linux
+# or the --enable-framework flag on Mac.
```
cmarqu, line 54:
```suggestion
# Cocotb requires linking to a dynamic Python library.
# On Windows, the Python library is always dynamic (an "Import Library").
# Python must be compiled with the `--enable-shared` flag on Linux
# and the `--enable-framework` flag on macOS.
```

setup.py
```diff
@@ -47,6 +47,20 @@
...
+from distutils import sysconfig
+
+if sys.platform == "darwin" and "--enable-framework" not in sysconfig.get_config_vars().get("CONFIG_ARGS", ""):
+    raise SystemExit("ERROR: You must use a Python compiled with --enable-framework")
+
+if sys.platform == "linux" and "--enable-shared" not in sysconfig.get_config_vars().get("CONFIG_ARGS", ""):
+    raise SystemExit("ERROR: You must use a Python compiled with --enable-shared")
+
```
cmarqu, line 62: I wonder if we should just dump out as much information about the environment as possible to help the user in case there are multiple installations?
https://docs.python.org/3/library/sysconfig.html#using-sysconfig-as-a-script shows how to do that. The paths are the most important thing, and it says there it uses ``get_path()``, but according to the description, it looks like ``get_paths()`` is actually the function to use? (I cannot try it right now.)

eric-wieser: At least including `sys.executable` in the message would be a good idea.

I'm also a little worried this might have a false positive. Can we add an override env var to continue with compilation anyway,`COCOTB_OVERRIDE_SHARED_CHECK`?

setup.py
```diff
@@ -47,6 +47,20 @@
...
+# Cocotb requires linking to a dynamic Python library.
+# On Windows, the Python library is always dynamic (an "Import Library").
+# Python must be compiled with the `--enable-shared` flag on Linux
+# and the `--enable-framework` flag on macOS.
+from distutils import sysconfig
+
+if sys.platform == "darwin" and "--enable-framework" not in sysconfig.get_config_vars().get("CONFIG_ARGS", ""):
+    raise SystemExit("ERROR: You must use a Python compiled with --enable-framework")
+
+if sys.platform == "linux" and "--enable-shared" not in sysconfig.get_config_vars().get("CONFIG_ARGS", ""):
```
eric-wieser, line 60:
```suggestion
elif sys.platform == "linux" and "--enable-shared" not in sysconfig.get_config_vars().get("CONFIG_ARGS", ""):
```

setup.py
themperek: >Can we add an override env var to continue with compilation anyway,COCOTB_OVERRIDE_SHARED_CHECK?

Are you sure we should try to workaround issue that we do know if exist?

eric-wieser: I suppose we can already do a bugfix release with that change of it did become a problem - so you're right, probably not a good idea after all

themperek:
I am not sure about this message formatting.

mtikekar:
Conda python on MacOS is compiled with `--enable-shared` and not `--enable-framework`. As you're doing this to make sure that the dynamic library exists, it might be best to directly check for that file.

eric-wieser:
> Conda python on MacOS is compiled with `--enable-shared` and not `--enable-framework`. As you're doing this to make sure that the dynamic library exists, it might be best to directly check for that file.

That sounds like a winning idea to me, assuming we already compute the full patch to the library to pass to `dlopen`.

setup.py
```diff
@@ -47,6 +47,18 @@
...
+# Cocotb requires linking to a dynamic Python library.
+# On Windows, the Python library is always dynamic (an "Import Library").
+# Python must be compiled with the `--enable-shared` flag on Linux and macOS
+# and the `--enable-framework` flag on macOS.
+from distutils import sysconfig
+
+if sys.platform == "posix":
+    config_args = sysconfig.get_config_vars().get("CONFIG_ARGS", "")
+    if "--enable-framework" not in config_args or "--enable-shared" not in config_args:
+        raise SystemExit("ERROR: You must use a Python compiled with --enable-shared or --enable-framework.\nPython executable location is %s" % sys.executable)
```
eric-wieser, line 60:
```suggestion
        raise SystemExit(
            "ERROR: You must use a Python compiled with --enable-shared or --enable-framework.\n"
            "Python executable location is {}".format(sys.executable))
```

ktbarrett:
Now that we have `find_libpython`, I think this can be modified to use that in setup.py.

PR #1751, opened 2020-05-02
Enable flake8 for examples and tests

cmarqu:
Do not exclude the directories ``examples`` and ``tests`` from the flake8 config in ``setup.cfg`` anymore;
fix flake8 reports in those directories.

There is one point where I wasn't entirely sure whether that's the best fix; it is marked with
``# FIXME: can we be more specific?``, please review.

### DIFF
--- a/tests/test_cases/test_avalon_stream/test_avalon_stream.py
+++ b/tests/test_cases/test_avalon_stream/test_avalon_stream.py
@@ -1,10 +1,8 @@
 #!/usr/bin/env python
 """Test to demonstrate functionality of the avalon basic streaming interface"""
 
-import logging
 import random
 import struct
-import sys
 
 import cocotb
 from cocotb.drivers import BitDriver
@@ -55,7 +53,7 @@ def test_avalon_stream(dut):
     tb.backpressure.start(wave())
 
     for _ in range(20):
-        data = random.randint(0,(2^7)-1)
+        data = random.randint(0, (2**7)-1)
         yield tb.send_data(data)
         yield tb.clkedge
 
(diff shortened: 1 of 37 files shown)


tests/test_cases/test_cocotb/test_generator_coroutines.py
```diff
@@ -50,7 +50,7 @@ def test_function_not_decorated(dut):
     except TypeError as exc:
         assert "yielded" in str(exc)
         assert "scheduler can't handle" in str(exc)
-    except:
+    except Exception:  # FIXME: can we be more specific?
```
eric-wieser, line 53: I think this is actually a bug - translating from one execution to another here isn't really useful.

What makes this a bug is the fact it's not failing if no exception is thrown, which would be spelt:
```suggestion
    else:
```

tests/test_cases/test_avalon_stream/test_avalon_stream.py
```diff
@@ -55,7 +53,7 @@ def test_avalon_stream(dut):
     tb.backpressure.start(wave())
 
     for _ in range(20):
-        data = random.randint(0,(2^7)-1)
+        data = random.randint(0,(2 ^ 7)-1)
```
eric-wieser, line 56: This is meant to be an exponent, right?

```suggestion
        data = random.randint(0, (2**7)-1)
```

cmarqu: Oh, yes, likely; fixed.

tests/test_cases/test_cocotb/test_generator_coroutines.py
cmarqu: Done.

eric-wieser:
Could you avoid amending when updating this PR, to make it easier to incrementally review?

cmarqu:
Oh, yes, sure.

Issue #1550, opened 2020-03-30
Normalize and improve type checking and support where coroutines are expected


ktbarrett:
There are many places that currently expect coroutines:

- `First`, `Combine`, `with_timeout`
- `yield` statements in old-style coroutines
- `fork`

All these places should be able to accept async def `coroutine`s, or `RunningTask`s, and much like `fork` do type checking. `First`, `Combine` and `with_timeout` don't currently accept `coroutine`s, and everything in the list has slightly different type checking. This can all be done in a central place by using a function that does the type checking and turns coroutine-like types into `RunningTask`s.

Additionally, but slightly differently, `RunningTask`'s constructor expects `coroutine` or generators and should trap common errors as well.

I think this is as simple as moving all the types checking to `RunningTask` and implementing a function that tries to convert the value if it is not already a `RunningTask`.

eric-wieser:
Learnings from #1752 are that we need a special error message for when `inspect.isasyncgenfunction(coro)`, because users are likely to forget to convert a yield to an await

ktbarrett:
My intention is to move all type checking into `cocotb.decorators.coroutine` for coroutine functions and `cocotb.decorators.RunningTask` for coroutine objects. I will also add in #1593 to the constructors. Then I will introduce.

```python
def make_coroutine(thing) -> cocotb.decorators.coroutine:
    if isinstance(thing, cocotb.decorators.coroutine):
        return thing
    else:
        return cocotb.decorators.coroutine(thing)

def make_task(thing) -> cocotb.decorators.RunningTask:
    if isinstance(thing, cocotb.decorators.RunningTask):
        return thing
    else:
        return cocotb.decorators.RunningTask(thing)

def iscoroutinefunction(thing) -> bool:
    return not throws(make_coroutine, thing)

def iscoroutine(thing) -> bool:
    return not throws(make_task, thing)

def isyeildable(thing) -> bool:  # name?
    """ Schedulable thing that can go in a yield/await statement in cocotb """
    return iscoroutine(thing) or isinstance(thing, (cocotb.triggers.Trigger, cocotb.triggers.Waitable))
```

These can be used all over: `scheduler.add`'s checks can be replaced with `make_task`, an `iscoroutine` check can be done in #1752. There are a lot of places in the codebase that do very similar type checking but do it to varying degrees of effectiveness. We shoudl locate and replace them with these functions.

eric-wieser:
[...]

PR #1752, opened 2020-05-03
Make examples use async instead of cocotb.coroutine

m-kru:

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -641,10 +641,14 @@ class TestFactory:
     """
 
     def __init__(self, test_function, *args, **kwargs):
-        if not isinstance(test_function, cocotb.coroutine):
+        if sys.version_info > (3, 6) and inspect.isasyncgenfunction(test_function):
+            raise TypeError("Expected a coroutine function, but got the async generator '{}'. "
+                            "Did you forget to convert a `yield` to an `await`?"
+                            .format(test_function.__qualname__))
+        if not (isinstance(test_function, cocotb.coroutine) or inspect.iscoroutinefunction(test_function)):
             raise TypeError("TestFactory requires a cocotb coroutine")
         self.test_function = test_function
-        self.name = self.test_function._func.__qualname__
+        self.name = self.test_function.__qualname__
 
         self.args = args
         self.kwargs_constant = kwargs
--- a/examples/dff/tests/dff_cocotb.py
+++ b/examples/dff/tests/dff_cocotb.py
@@ -29,7 +29,6 @@ import random
 
 import cocotb
 from cocotb.clock import Clock
-from cocotb.decorators import coroutine
 from cocotb.triggers import RisingEdge
 from cocotb.monitors import Monitor
 from cocotb.drivers import BitDriver
@@ -55,13 +54,12 @@ class BitMonitor(Monitor):
         self.clk = clk
         Monitor.__init__(self, callback, event)
 
-    @coroutine
-    def _monitor_recv(self):
+    async def _monitor_recv(self):
         clkedge = RisingEdge(self.clk)
 
         while True:
             # Capture signal at rising edge of clock
-            yield clkedge
+            await clkedge
             vec = self.signal.value
             self._recv(vec)
 
@@ -131,8 +129,7 @@ class DFF_TB(object):
         self.stopped = True
 
 
-@cocotb.coroutine
-def run_test(dut):
+async def run_test(dut):
     """Setup testbench and run a test."""
 
     cocotb.fork(Clock(dut.c, 10, 'us').start(start_high=False))
@@ -144,12 +141,12 @@ def run_test(dut):
     # Apply random input data by input_gen via BitDriver for 100 clock cycles.
     tb.start()
     for _ in range(100):
-        yield clkedge
+        await clkedge
 
     # Stop generation of input data. One more clock cycle is needed to capture
     # the resulting output of the DUT.
     tb.stop()
-    yield clkedge
+    await clkedge
 
     # Print result of scoreboard.
     raise tb.scoreboard.result


eric-wieser:
Error is:
```
Failed to import module dff_cocotb: 'yield' inside async function (dff_cocotb.py, line 144)
```
You need to convert `yield` to `await` inside the function.

m-kru:
@eric-wieser Where do you see this message, and why I do not get this message when I try to run locally? Do you know? I have reinstalled cocotb to include changes.

eric-wieser:
Python 3.5 CI is the one that gives that.

This raises that we need a better error for `inspect.iscoroutinegenerator`

cocotb/regression.py
```diff
@@ -641,7 +641,7 @@ class TestFactory:
     """
 
     def __init__(self, test_function, *args, **kwargs):
-        if not isinstance(test_function, cocotb.coroutine):
+        if not (isinstance(test_function, cocotb.coroutine) or inspect.iscoroutinefunction(test_function)):
```
eric-wieser, line 644:
```suggestion
        if inspect.isasyncgenfunction(test_function):
            raise TypeError("some message about mixing yield and await")
        if not (isinstance(test_function, cocotb.coroutine) or inspect.iscoroutinefunction(test_function)):
```

m-kru:
@eric-wieser There is new error this time.

cocotb/regression.py
```diff
@@ -641,7 +641,9 @@ class TestFactory:
     """
 
     def __init__(self, test_function, *args, **kwargs):
-        if not isinstance(test_function, cocotb.coroutine):
+        if inspect.isasyncgenfunction(test_function):
+            raise TypeError(f"Found yield in async function '{test_function.__name__}'")
```
eric-wieser, line 645: We still support python 3.5, so you can't use f-strings yet.

m-kru: Ok, thanks the E999 error message from flake8 was not helpful. Nevertheless, it is not the issues I have mentioned. Let me fix this first.

m-kru:
Ok, I meant this one:
```
AttributeError: 'function' object has no attribute '_func'
````

File: cocotb/regression.py [hunk 3]
```diff
@@ -641,7 +641,9 @@ class TestFactory:
     """
 
     def __init__(self, test_function, *args, **kwargs):
-        if not isinstance(test_function, cocotb.coroutine):
+        if inspect.isasyncgenfunction(test_function):
+            raise TypeError("Found yield in async function '%s'" % test_function.__name__)
```
eric-wieser, line 645: Apologies, was on mobile earlier so had difficultly suggesting a message
```suggestion
        if sys.version_info > (3, 6) and inspect.isasyncgenfunction(test_function):
            raise TypeError(
                "Expected a coroutine function, but got the async generator '{}'. "
                "Did you forget to convert a `yield` to an `await`?"
                .format(test_function.__qualname__))
            )
```
Also needs a version check since `isasyncgenfunction` is new in 3.6.

cocotb/regression.py
```diff
@@ -641,7 +641,9 @@ class TestFactory:
...
+        if not (isinstance(test_function, cocotb.coroutine) or inspect.iscoroutinefunction(test_function)):
             raise TypeError("TestFactory requires a cocotb coroutine")
         self.test_function = test_function
         self.name = self.test_function._func.__qualname__
```
eric-wieser, line 649: I think the best fix would be:
```suggestion
        self.name = self.test_function.__qualname__
```
And add the `__qualname__` attribute to `coroutime` in the constructor.

m-kru:
@eric-wieser One more question. There is following class:
```python
class BitMonitor(Monitor):
    """Observe a single-bit input or output of the DUT."""
    def __init__(self, name, signal, clk, callback=None, event=None):
        self.name = name
        self.signal = signal
        self.clk = clk
        Monitor.__init__(self, callback, event)

    @coroutine
    def _monitor_recv(self):
        clkedge = RisingEdge(self.clk)

        while True:
            # Capture signal at rising edge of clock
            yield clkedge
            vec = self.signal.value
            self._recv(vec)
```

Should `_monitor_recv` also be transformed to `async`?

eric-wieser:
Yes, it should!

ktbarrett:
@m-kru I think all of the drivers and monitors need to be updated to async, unless it's blocking you, skip it.

ktbarrett:
Your need to do type checking on coroutines is yet anther argument for #1550. I'll make a PR for that today.

m-kru:
What about `@cocotb.test()` decorator?

```python
@cocotb.test()
def mean_basic_test(dut):
    """Test n*5/n = 5"""
    yield value_test(dut, 5)
```

eric-wieser:
That decorator needs to stay, but you can still make the function async.

m-kru:
@eric-wieser You `can` or rather `must`/`should`?

eric-wieser:
Someone should, but it doesn't need to be you, and it doesn't need to be this PR.

examples/mean/tests/Makefile
```diff
@@ -1,27 +1,23 @@
```
eric-wieser, line 1: Please don't change this file in this PR. Feel free to file another PR with these changes, but this PR should be limited to python / documentation changes.

themperek:
@m-kru can you remove last commit "Adjust mean example to use async functions." then we can probably marge this as is. There is another PR for this #1779

m-kru:
@themperek I could, but I wanted to end this PR (adjust all examples), however it turned out that I do not have simulator capable of running all examples.

themperek:
@m-kru Maybe remove the last commit we try to merge it you can open another PR.

m-kru:
@themperek done.

themperek:
@m-kru Should not make 2 commits out or this one for async execution one other for dff example change? @eric-wieser

m-kru:
Well, in my humble opinion we are becoming too pedantic, but I am not the one managing the project.

eric-wieser:
@themperek: Agreed, I squashed the commits myself.

PR #1754, opened 2020-05-03
Remove F403 from flake8 ignore list

cmarqu:
Fix occurrence of this error.
Refs #1547.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,6 @@ ignore =
     E402  # module level import not at top of file
     E501  # line too long (... > 79 characters)
     E741  # ambiguous variable name 'l'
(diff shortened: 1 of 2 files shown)


imphil:
F403 -> E403

setup.cfg
```diff
@@ -27,7 +27,6 @@ ignore =
     E402  # module level import not at top of file
     E501  # line too long (... > 79 characters)
     E741  # ambiguous variable name 'l'
-    F403  # 'from ... import *' used; unable to detect
```
eric-wieser, line 30: @imphil: this is an F

imphil:
Good point, I got carried away by the other one which was an E indeed :smile: Thanks for catching that.

PR #1755, opened 2020-05-03
Remove F741 from flake8 ignore list

cmarqu:
Fix occurrence of this error.
Refs #1547.

### DIFF
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -229,40 +229,40 @@ class BinaryValue:
     def _adjust_unsigned(self, x):
         if self._n_bits is None:
             return x
-        l = len(x)
-        if l <= self._n_bits:
+        _len = len(x)
+        if _len <= self._n_bits:
             if self.big_endian:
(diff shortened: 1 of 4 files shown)


eric-wieser:
Not sure if this one is worth the churn

imphil:
I'm OK with renaming, but a local variable with an underscore (`_len`) looks inconsistent compared to the other code. (I personally would probably continue to ignore this warning, `l` is a perfectly good variable name in my mind; but fine either way.)

imphil:
... and please fix the commit message to say `E741` instead of `F741`.

eric-wieser:
As an aside, the convention for avoiding name clashes with builtins is `name_` not `_name`. But I'd either pick a better name, or just leave this as `l`.

cmarqu:
I personally wouldn't use a single-character identifier, and especially in ``utils.py`` usage of ``l`` is not restricted to within just 3 or so lines, but given that this is not our original code anyway, let's leave it at that.

PR #1760, opened 2020-05-04
Fix broken reporting of available SIM settings

cmarqu:
This duplicates two lines from ``Makefile.sim`` which is not included here.

The original error was (when  running from cocotb's root directory):
```
$ make help
[...]
Default simulator is Icarus. To use another set environment variable SIM as below
/bin/sh: -c: line 1: syntax error near unexpected token `echo'
/bin/sh: -c: line 1: `echo $X | sed 's/^[^.]*./export SIM=/';\'
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,9 @@ jenkins: do_tests
 test:
 	$(MAKE) do_tests; ret=$$?; ./bin/combine_results.py && exit $$ret
 
+COCOTB_MAKEFILES_DIR = $(realpath $(shell cocotb-config --makefiles))
+AVAILABLE_SIMULATORS = $(patsubst .%,%,$(suffix $(wildcard $(COCOTB_MAKEFILES_DIR)/simulators/Makefile.*)))
+
 .PHONY: help
 help:
 	@echo ""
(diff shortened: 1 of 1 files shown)


cmarqu:
Hmm, need to think more... in this case, we probably do not want to use the installed cocotb but the one in the relative path, as it was before?

cmarqu:
Since we require installing cocotb now always, this change is okay I think.

Makefile
```diff
@@ -67,7 +70,8 @@ help:
 	@echo "clean     - remove build directory and all simulation artefacts"
 	@echo ""
 	@echo "Default simulator is Icarus. To use another set environment variable SIM as below"
-	@for X in $(shell ls cocotb/share/makefiles/simulators/); do \
-		echo $$X | sed 's/^[^.]*./export SIM=/';\
+	@echo "Available simulators: $(AVAILABLE_SIMULATORS)"
+	@for X in ${AVAILABLE_SIMULATORS}; do \
```
eric-wieser, line 74: Shouldn't this be
```suggestion
	@for X in $(AVAILABLE_SIMULATORS); do \
```

Makefile
```diff
@@ -67,7 +70,8 @@ help:
 	@echo "clean     - remove build directory and all simulation artefacts"
 	@echo ""
 	@echo "Default simulator is Icarus. To use another set environment variable SIM as below"
-	@for X in $(shell ls cocotb/share/makefiles/simulators/); do \
-		echo $$X | sed 's/^[^.]*./export SIM=/';\
+	@echo "Available simulators: $(AVAILABLE_SIMULATORS)"
```
eric-wieser, line 73:
```suggestion
	@echo "Available simulators:"
```

cmarqu: Ooops.

Makefile
cmarqu: Both work, but for consistency this is indeed nicer.

cmarqu:
While we are at it, let's make this a bit nicer by sorting the output and using full sentences. Please give it another look:
```
> make help

Available simulators:
export SIM=activehdl
export SIM=aldec
export SIM=cvc
export SIM=ghdl
export SIM=icarus
export SIM=ius
export SIM=modelsim
export SIM=questa
export SIM=vcs
export SIM=verilator
export SIM=xcelium

```

PR #1764, opened 2020-05-04
Move variable help text out of Makefile.sim

cmarqu:
This allows to get the short help summary for available variables for Makefiles (as before) but also other flows we may want to support in the future.

I am  (mis?)using ``cocotb-config`` here which is not optimal, but an extra binary seems overkill... In the past, a manpage has been suggested, but IMO this is not as discoverable and flexible.
Any other options or opinions?

### DIFF
--- a/cocotb/config.py
+++ b/cocotb/config.py
@@ -41,6 +41,7 @@ Global variables:
 import argparse
 import os
 import sys
+import textwrap
 import cocotb
 
 __all__ = ["share_dir", "makefiles_dir"]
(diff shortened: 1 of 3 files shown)


themperek:
Maybe makefile variables should be separate?

cmarqu:
Yes, maybe they should stay in ``Makefile.sim``... I'll do that.

cocotb/config.py
```diff
@@ -50,6 +51,49 @@
...
 
+def help_vars_text():
+    if "dev" in cocotb.__version__:
+        doclink = "https://docs.cocotb.org/en/latest/building.html"
+    else:
+        doclink = "https://docs.cocotb.org/en/v{}/building.html".format(cocotb.__version__)
+
+    # NOTE: make sure to keep "helpmsg" aligned with documentation/source/building.rst
+    # Also keep it at 80 chars.
+    helpmsg = textwrap.dedent("""
```
eric-wieser, line 62:
```suggestion
    helpmsg = textwrap.dedent("""\
```
to remove the leading newline

cocotb/config.py
```diff
@@ -50,6 +51,49 @@
...
+    ---
+    GPI_EXTRA                 Extra libraries to load at runtime (comma-separated)
+
+    Scheduler
+    ---------
+    COCOTB_SCHEDULER_DEBUG    Enable additional output of coroutine scheduler
+
+    For details, see {}
+    """).format(doclink)
+    return helpmsg[:-1]
```
eric-wieser, line 94:
```suggestion
    For details, see {}""").format(doclink)
    return helpmsg
```

cocotb/config.py
cmarqu: Done, thx.

Issue #134, opened 2014-06-25
Improve exchange of values between sim<->python


chiggs:
The code in the `binary` module needs a refresh - it's currently a bottleneck in small intensive simulations and doesn't provide all of the features it should.  As per issue #119, we need to minimise the number of methods on the `Handle` class to avoid naming clashes with real signals.  We must therefore prefer built-in conversions where possible (e.g. `int(dut.data_out)`, `float(dut.data_out)`, `str(dut.data_out)`.
#### Where possible move access directly into the `Handle` class

At the moment everything goes through a `BinaryValue`, however it would be far more convenient (and faster) to retrieve values directly from the handle.

Suggest we only use `BinaryValue` where _special_ access is required, defined as non-native Python types, unresolved representations etc.
#### Maintain ability to read/write unresolved values

The simplest way to achieve this is to use strings, for example `0010x01z` is a valid value to drive / read back.  However this is typically _not_ the fast-path assignment mechanism.  Access in the simulator is using strings is slow and this is a rarely used path so shouldn't be the general mechanism (as current design).
#### Provide a way to specify a format for a handle

Typically signals have some meaning beyond a bunch of bits and generally speaking this meaning is fixed.  We should provide some convenience mechanisms for describing the signal format such that Cocotb can do the appropriate conversions (and bounds checking) on access.

```
>>> print len(dut.data_in)
4
>>> dut.data_in = -4
>>> dut.data_in = -28
ValueError("-28 outside the range represented by a 4-bit signed integer")
```

We should support the following formats:
- Unsigned integer
- Signed integer
- Fixed point
- Floating point (specify sign, exponent, mantissa)

We also probably require a way to set the endianness of the vector.

The actual mechanism for specifying the format is up for debate:
##### 1. Methods / attributes on `Handle` class

`dut.data_out.format = Float`
`dut.data_out.set_format(Float)`

This is not preferred as we introduce a possible naming clash with traversing the RTL hierarchy (see issue #119)
##### 2. Utility functions

Could provide a set of utility functions in a different namespace to avoid issue above

```
>>> cocotb.set_format(dut.data_in, Float)
>>> cocotb.set_signed(dut.data_in)
>>> cocotb.set_unsigned(dut.data_out)
```
##### 3. Use a built-in on the `Handle` class

For example, using the `__call__` method:

```
dut.data_in(Float)
dut.data_out(BigEndian)
```
#### Improve the speed of access

[...]

PR #1765, opened 2020-05-04
Move issue_134 tests

elgorwi:
- Move issue_134 tests to test_handle.
- Add test_real_ prefix to assign_double and assign_int test names for consistency.


eric-wieser:
xref #134, for reviewers

PR #1766, opened 2020-05-05
Issue #913: fix setting ints to signals wider than 32-bit

elgorwi:
- Fix bug for setting signed ints to signals wider than 32-bit. (handle negative values correctly, Issue #913)
- Improve performance by using to_bytes rather than BinaryValue.
- Add overflow checks.
- Add tests for 32-bit and 39bit signals.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -656,13 +656,17 @@ class ModifiableObject(NonConstantObject):
         """
         value, set_action = self._check_for_set_action(value)
 
-        if isinstance(value, int) and value < 0x7fffffff and len(self) <= 32:
-            call_sim(self._handle.set_signal_val_long, set_action, value)
-            return
+        if isinstance(value, int):
+            if len(self) <= 32 and -(2**31) <= value <= (2**31)-1:
+                call_sim(self._handle.set_signal_val_long, set_action, value)
+                return
+            else:
+                value = value.to_bytes((len(self)+9)//8, byteorder='big', signed=True)
+                call_sim(self._handle.set_signal_val_bytes, set_action, value)
+                return
+
         if isinstance(value, ctypes.Structure):
             value = BinaryValue(value=cocotb.utils.pack(value), n_bits=len(self))
-        elif isinstance(value, int):
-            value = BinaryValue(value=value, n_bits=len(self), bigEndian=False)
         elif isinstance(value, dict):
             # We're given a dictionary with a list of values and a bit size...
             num = 0
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -42,6 +42,8 @@ static int sim_ending = 0;
 #include "simulatormodule.h"
 #include <cocotb_utils.h>     // COCOTB_UNUSED
 #include <type_traits>
+#include <array>
+#include <string>
 
 
 /* define the extension types as templates */
@@ -588,6 +590,60 @@ static PyObject *get_signal_val_long(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject
     return PyLong_FromLong(result);
 }
 
+
+static PyObject *set_signal_val_bytes(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject *args)
+{
+    gpi_set_action_t action;
+    const char *bytes;
+    Py_ssize_t len;
+
+    if (!PyArg_ParseTuple(args, "iy#:set_signal_val_bytes", &action, &bytes, &len)) {
+        return NULL;
+    }
+
+    const std::array<std::string, 16> BINSTR_LUT = {
+        {"0000", "0001", "0010", "0011", "0100", "0101", "0110", "0111",
+         "1000", "1001", "1010", "1011", "1100", "1101", "1110", "1111"}
+    };
+
+    const std::string HEXSTR_LUT = {"0123456789abcdef"};
+
+    size_t m_elems{0};
+    uint8_t m_idx_hi{0};
+    uint8_t m_idx_lo{0};
+    std::string m_binstr{""};
+    std::string m_hexstr{""};
+
+    for (auto i = 0; i < len; ++i)
+    {
+        m_idx_hi = (bytes[i] >> 4) & UINT8_C(0x0F);
+        m_idx_lo = bytes[i] & UINT8_C(0x0F);
+
+        m_binstr += BINSTR_LUT[m_idx_hi];
+        m_binstr += BINSTR_LUT[m_idx_lo];
+        m_hexstr += HEXSTR_LUT[m_idx_hi];
+        m_hexstr += HEXSTR_LUT[m_idx_lo];
+    }
+
+    m_elems = static_cast<size_t>(gpi_get_num_elems(self->hdl));
+
+    if (m_binstr.length() > m_elems) {
+        std::string m_msb = m_binstr.substr(0, m_binstr.length() - m_elems);
+        bool m_ovfl = m_msb.find_first_not_of(m_msb[0]) != std::string::npos; // are MSBs not all the same?
+
+        if (m_ovfl) {
+            const char *m_name = gpi_get_signal_name_str(self->hdl);
+            PyErr_Format(PyExc_OverflowError, "Value (0x%s) out of range for assignment of %d-bit signal (%s)", m_hexstr.c_str(), m_elems, m_name);
+            return NULL;
+        } else {
+            m_binstr = m_binstr.substr(m_binstr.length() - m_elems, std::string::npos); // safely drop MSBs
+        }
+    }
+
+    gpi_set_signal_value_binstr(self->hdl, m_binstr.c_str(), action);
+    Py_RETURN_NONE;
+}
+
 static PyObject *set_signal_val_binstr(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject *args)
 {
     const char *binstr;
@@ -903,6 +959,9 @@ static PyMethodDef gpi_sim_hdl_methods[] = {
     {"set_signal_val_str",
         (PyCFunction)set_signal_val_str, METH_VARARGS,
         "Set the value of a signal using an NUL-terminated 8-bit string"},
+    {"set_signal_val_bytes",
+        (PyCFunction)set_signal_val_bytes, METH_VARARGS,
+        "Set the value of a signal using a bytes object"},
     {"set_signal_val_binstr",
         (PyCFunction)set_signal_val_binstr, METH_VARARGS,
         "Set the value of a signal using a string with a character per bit"},
--- a/tests/test_cases/test_cocotb/test_handle.py
+++ b/tests/test_cases/test_cocotb/test_handle.py
@@ -48,7 +48,7 @@ def test_bad_attr(dut):
 @cocotb.test(skip=cocotb.SIM_NAME.lower().startswith("icarus"))
 async def test_string_handle_takes_bytes(dut):
     dut.stream_in_string.value = b"bytes"
-    await cocotb.triggers.Timer(10, 'ns')
+    await Timer(10, 'ns')
     val = dut.stream_in_string.value
     assert isinstance(val, bytes)
     assert val == b"bytes"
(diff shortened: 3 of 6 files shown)


cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -588,6 +590,54 @@ static PyObject *get_signal_val_long(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject
...
+        if (m_ovfl) {
+            const char *m_name = gpi_get_signal_name_str(self->hdl);
+            PyErr_Format(PyExc_OverflowError, "Value (0x%s) out of range for assignment of %d-bit signal (%s)", m_hexstr.c_str(), m_elems, m_name);
+            return NULL;
+        } else {
+            m_binstr = m_binstr.substr(m_binstr.length() - m_elems, std::string::npos); // safely drop MSBs
+        }
+    }
+
+    gpi_set_signal_value_binstr(self->hdl, m_binstr.c_str(), action);
```
elgorwi, line 637: We could investigate if adding gpi_set_signal_value_hexstr(), using vpiHexStrVal/vhpiHexStrVal would improve performance.

cocotb/handle.py
```diff
@@ -656,13 +656,16 @@ def _set_value(self, value, call_sim):
         """
         value, set_action = self._check_for_set_action(value)
 
-        if isinstance(value, int) and value < 0x7fffffff and len(self) <= 32:
-            call_sim(self._handle.set_signal_val_long, set_action, value)
+        if isinstance(value, int):
+            if value >= -0x80000000 and value <= 0x7fffffff and len(self) <= 32:
```
ktbarrett:
```suggestion
            if len(self) <= 32 and -(2**31) <= value <= (2**31)-1:
```

cocotb/handle.py
```diff
@@ -656,13 +656,16 @@ def _set_value(self, value, call_sim):
...
+                call_sim(self._handle.set_signal_val_long, set_action, value)
+            else:
+                value = value.to_bytes((len(self)+7)//8, byteorder='big', signed=value<0)
```
ktbarrett, line 663: This line is problematic. Defaulting to big endian is probably not a good idea, most systems are little endian nowadays. We could use `sys.byteorder`, but that makes an assumption that the user knows the byte order of their system. We might end up in a situations where an unexpected value is written.

cocotb/handle.py
```diff
@@ -656,13 +656,16 @@ def _set_value(self, value, call_sim):
...
+                call_sim(self._handle.set_signal_val_bytes, set_action, value)
             return
+
         if isinstance(value, ctypes.Structure):
             value = BinaryValue(value=cocotb.utils.pack(value), n_bits=len(self))
-        elif isinstance(value, int):
-            value = BinaryValue(value=value, n_bits=len(self), bigEndian=False)
```
ktbarrett, line 665: Should we get rid of this, or change the `binaryRepresentation` to two's complement? It works the same for unsigned numbers, but supports negative numbers in the representation used by all (fight me C standards committee) modern processor archs.

tests/test_cases/issue_142/vsim_stacktrace.vstf
```diff
@@ -0,0 +1,11 @@
+# Current time Tue May 05 12:42:13 2020
```
ktbarrett, line 1: Whoops, checked in some garbage.

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -65,3 +65,36 @@ async def test_delayed_assignment_still_errors(dut):
...
+    U_MAX = 2**N-1
+    S_MIN = -2**(N-1)
+    S_MAX = 2**(N-1)-1
+
+    for value in [0, 1, -1, 4, -4, U_MIN, U_MAX, S_MIN, S_MAX]:
+        dut.stream_in_data_36bit = value
+        await cocotb.triggers.Timer(10, 'ns')
+        got = int(dut.stream_in_data_36bit)
+        if value < 0:
+            got = (got & (2**(N-1)-1)) - (got & (2**(N-1)))
```
ktbarrett, line 86:
```suggestion
        got = dut.stream_in_data_36bit.value.signed_integer
```

cocotb/handle.py
ktbarrett: Nvm, we probably should get rid of it. But I don't think Eric will agree.

cocotb/handle.py
elgorwi: Nice, didn’t know the combined comparison syntax was supported

cocotb/handle.py
elgorwi: Actually, I think we want to hardcode this to big endian, as that is how we want to present it to the simulator. From the docs:
```The byteorder argument determines the byte order used to represent the integer. If byteorder is "big", the most significant byte is at the beginning of the byte array. If byteorder is "little", the most significant byte is at the end of the byte array. To request the native byte order of the host system, use sys.byteorder as the byte order value.```

File: cocotb/handle.py (hunk 4 above)
Comment (elgorwi) [reply cocotb/handle.py:665]: This case should be covered by the else clause on line 662.
Essentially we could achieve the same by using the ```binaryRepresentation``` for negative values and default for positive values

File: tests/test_cases/issue_142/vsim_stacktrace.vstf (hunk 5 above)
Comment (elgorwi) [reply tests/test_cases/issue_142/vsim_stacktrace.vstf:1]: I’ll clean this up

File: tests/test_cases/test_cocotb/test_handle.py (hunk 6 above)
Comment (elgorwi) [reply tests/test_cases/test_cocotb/test_handle.py:86]: This is not the same as we test both signed and unsigned values. It is a matter of choice if we interpret the bits we read back as signed or unsigned. (Eg. -1 signed is the same as max value unsigned)

File: cocotb/handle.py (hunk 3 above)
Comment (ktbarrett) [reply cocotb/handle.py]: Endianness isn't a matter of "correct", it's a matter of preference, and the reality of the hardware being used.

File: tests/test_cases/test_cocotb/test_handle.py [hunk 7]
```diff
@@ -123,3 +123,93 @@ def test_real_assign_int(dut):
...
+    """
+    Test signed integer access to 32-bit vector
+    """
+    N = len(dut.stream_in_data_dword)
+    S_MIN = -2**(N-1)
+    S_MAX = 2**(N-1)-1
+
+    for value in [0, 1, -1, 4, -4, S_MIN, S_MAX]:
+        dut.stream_in_data_dword = value
+        await cocotb.triggers.Timer(10, 'ns')
```
cmarqu, line 139:
```suggestion
        await Timer(10, 'ns')
```
and below; unless you had a specific reason (which then isn't obvious).

cocotb/handle.py
elgorwi: I think there is some confusion here with big/little endianness vs ascending/descending vector range. 
```set_signal_value_long()``` is used for setting vectors < 32-bit and assumes that the vector we're setting has a "descending" range. I wanted to ensure that the behaviour of setting vectors > 32-bit using ```set_signal_value_bytes()``` is consistent with how it is for setting vectors < 32-bit.

Here's a code snippet that illustrates that we want to use big (network) endian to ensure the bit pattern remains as intended left to right and avoid any extra work with reordering of the bytes in the C++ code. 

```python
In [1]: N=0x0123456789abcdef

In [2]: N.to_bytes(10, byteorder="big", signed=False).hex()
Out[2]: '00000123456789abcdef'

In [3]: N.to_bytes(10, byteorder="little", signed=False).hex()
Out[3]: 'efcdab89674523010000'

In [4]: N.to_bytes(10, byteorder="big", signed=False).hex()[::-1]
Out[4]: 'fedcba98765432100000'

In [5]: N.to_bytes(10, byteorder="little", signed=False).hex()[::-1]
Out[5]: '00001032547698badcfe'
```

cocotb/handle.py
elgorwi: I think there is some confusion here with big/little endianness vs ascending/descending vector range. 
```set_signal_value_long()``` is used for setting vectors < 32-bit and assumes that the vector we're setting has a "descending" range. I wanted to ensure consistency with this behaviour for setting vectors > 32-bit using ```set_signal_value_bytes()``` 
Here's a code snippet that illustrates that we want to use big (network) endian to ensure the bit pattern remains as intended left to right and avoid any extra work with reordering of the bytes in the C++ code.
```python
In [1]: N=0x0123456789abcdef

In [2]: N.to_bytes(10, byteorder="big", signed=False).hex()
Out[2]: '00000123456789abcdef'

In [3]: N.to_bytes(10, byteorder="little", signed=False).hex()
Out[3]: 'efcdab89674523010000'

In [4]: N.to_bytes(10, byteorder="big", signed=False).hex()[::-1]
Out[4]: 'fedcba98765432100000'

In [5]: N.to_bytes(10, byteorder="little", signed=False).hex()[::-1]
Out[5]: '00001032547698badcfe'
```

tests/test_cases/test_cocotb/test_handle.py
elgorwi: Thanks for the suggestion, still learning about await.

ktbarrett:
My general opinion is that there is no reason to introduce another function to fix #913, `int`s being applied to signals over 32 bits should go through BinaryValue. That set of operations you have is simply doing the same thing that BinaryValue is doing, but doing it in C. If you think there is performance benefit to doing it this way, open another PR and for that change and post benchmarks.

The fundamental issue of #913 is that `_set_value` was written assuming you will never feed it a negative number, which is just silly. We are just trying to fix the logic to cover all possible paths:

* value is `int` (we aren't worried about the other paths right now)
* signal is less than / equal to / greater than 32 bits in size
* value is negative / positive

Our requirements/constraints are:

* can't represent values greater than the limit of `set_signal_value_long` which is a signed 32 bit integer
* if the value is positive, we should assume the signal is unsigned and support up to values that are (2 ** len(signal)) - 1
* if the value is negative, we should assume the signal is signed with two's complement (since that's what `int` is in Python) and support values down to -(2 ** (len(signal) - 1))
* if the value is larger than the bounds supported, we must go through BinaryValue to preserve the truncation behavior

Accordingly, the logic should read something like this:
```python
if isinstance(value, int):
    min_value = -(2 ** (len(self) - 1))  # minimum negative value for two's complement of this size
    max_value = (2 ** min(len(self), 31) - 1  # maximum positive value up to 31 bit signals since we can should support 'unsigned' signals for positive values
    if len(self) <= 32 and min_value <= value <= max_value:
        return call_sim(self._handle.set_signal_val_long, set_action, value)
    elif value < 0:
        value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.TWOS_COMPLEMENT)
    else:
        value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.UNSIGNED)
```

elgorwi:
> My general opinion is that there is no reason to introduce another function to fix #913, `int`s being applied to signals over 32 bits should go through BinaryValue. That set of operations you have is simply doing the same thing that BinaryValue is doing, but doing it in C. If you think there is performance benefit to doing it this way, open another PR for that change and post benchmarks.
> 
> The fundamental issue of #913 is that `_set_value` was written assuming you will never feed it a negative number, which is just silly. We are just trying to fix the logic to cover all possible paths:
> 
>     * value is `int` (we aren't worried about the other paths right now)
> 
>     * signal is less than / equal to / greater than 32 bits in size
> 
>     * value is negative / positive
> 
> 
> Our requirements/constraints are:
> 
>     * can't represent values greater than the limit of `set_signal_value_long` which is a signed 32 bit integer
> 
>     * if the value is positive, we should assume the signal is unsigned and support up to values that are (2 ** len(signal)) - 1
> 
>     * if the value is negative, we should assume the signal is signed with two's complement (since that's what `int` is in Python) and support values down to -(2 ** (len(signal) - 1))
> 
>     * if the value is larger than the bounds supported, we must go through BinaryValue to preserve the truncation behavior
> 
> 
> Accordingly, the logic should read something like this:
> 
> ```python
> if isinstance(value, int):
>     min_value = -(2 ** (len(self) - 1))  # minimum negative value for two's complement of this size
>     max_value = (2 ** min(len(self), 31) - 1  # maximum positive value up to 31 bit signals since we can should support 'unsigned' signals for positive values
>     if len(self) <= 32 and min_value <= value <= max_value:
>         return call_sim(self._handle.set_signal_val_long, set_action, value)
>     elif value < 0:
>         value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.TWOS_COMPLEMENT)
>     else:
>         value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.UNSIGNED)
> ```
> 
> It might make sense to make the calculations for `min_value` and `max_value` to be cached function calls to improve performance because there are few potential signal sizes.

This PR is part of a larger story (#1742 , #1744 ). The feedback has been to move range checks to C++ (https://github.com/cocotb/cocotb/pull/1742#discussion_r419086506). 
This PR uses the same approach, but I can understand if this is too big of a leap. 
Let me prepare a new PR that fixes #1766 in Python only and then we can choose how to proceed.

ktbarrett:
@elgorwi This PR and #1744 seem to be related, they relate to setting the value of a logic vector using a Python `int`. you can probably merge them into one PR (#1744 has no comments).

Instead of deciding between using `set_signal_value_int` and some other function in Python, you could do *all* of this work in C? `set_signal_value_int` can remain for setting integer signals, but we can have another C function that takes a Python `int` to set logic vectors, and at that point we can decide in C what the best way to perform the operation is. Whether that be `gpi_set_signal_value_long` or something else.

Also as a general note, `BinaryValue` does not throw if the value exceeds the bounds, it logs a warning and truncates the value.

elgorwi:
> 
> 
> @elgorwi This PR and #1744 seem to be related, they relate to setting the value of a logic vector using a Python `int`. you can probably merge them into one PR (#1744 has no comments).
> 
> Instead of deciding between using `set_signal_value_int` and some other function in Python, you could do _all_ of this work in C? `set_signal_value_int` can remain for setting integer signals, but we can have another C function that takes a Python `int` to set logic vectors, and at that point we can decide in C what the best way to perform the operation is. Whether that be `gpi_set_signal_value_long` or something else.
> 
> Also as a general note, `BinaryValue` does not throw if the value exceeds the bounds, it logs a warning and truncates the value.

@ktbarrett , just opened a new PR with a Python only solution and a comprehensive set of tests (#1792 ). I've indeed noticed that BinaryValue will not throw an OverflowError, so I've included a range check in #1792 .
I'm happy to proceed with #1792  first to fix the bug described in issue #913 and when merged introduce further improvements incrementally by changing the other open PRs.
#1744 is indeed related but let's take 1 step at a time.

PR #1767, opened 2020-05-05
Centralize type checking/coercion

### LINKED ISSUE #1550: Normalize and improve type checking and support where coroutines are expected

ktbarrett:
Partially fixes #1550.

Introduces new functions for type checking values to be a cocotb coroutine function, coroutine object (`RunningTask`), or 'yieldable' object. Introduces functions for coercing values into either a coroutine function or a coroutine object. Moves all type checking to for coroutine functions to the constructor of `RunningTask`. Seeks to replace all similar logic in the core library and user code with a central definition that covers all cases. For instance, the logic in `scheduer.add` (AKA `fork`) was replaced with `make_task`.

Originally I had type checking for `coroutine` in the constructor, but it is currently expected that invalid `coroutine` decorated objects not fail because it would cause the test module to fail to load. It is expected to catch it when it is constructed into a task.

Will look for other places to use these functions, I've seen similar code in several places.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -316,27 +316,6 @@ class coroutine:
     """
 
     def __init__(self, func):
-
-        if inspect.isgeneratorfunction(func) or inspect.iscoroutinefunction(func):
-            # check and exit early in the correct case
-            pass
-
-        elif sys.version_info >= (3, 6) and inspect.isasyncgenfunction(func):
-            raise TypeError(
-                "'{}' is an async generator, not a coroutine function.\n"
-                "Did you forget to update your `yield`s to `await`s?".format(
-                    func.__qualname__))
-
-        elif inspect.isfunction(func):
-            raise TypeError(
-                "'{}' is not a generator function.\n"
-                "Did you forget to use the `yield` keyword?".format(
-                    func.__qualname__))
-
-        else:
-            raise TypeError(
-                "'{}' is not a valid cocotb coroutine function.".format(func))
-
         self._func = func
         functools.update_wrapper(self, func)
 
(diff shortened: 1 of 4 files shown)


cocotb/decorators.py
```diff
@@ -497,3 +515,68 @@ async def f(*args, **kwargs):
...
+    """
+    try:
+        make_task(thing)
+    except TypeError:
+        return False
+    else:
+        return True
+
+
+def isyeildable(thing: Any) -> bool:  # name?
```
garmin-mjames, line 578:
```suggestion
def isyieldable(thing: Any) -> bool:
```

cocotb/decorators.py
```diff
@@ -91,16 +92,37 @@ def __init__(self, inst):
...
+                "You likely used the `yield` keyword instead of `await`.\n"
+                "And you also forgot to call the function."
+                .format(inst.__qualname__))
+
         else:
             raise TypeError(
-                "%s isn't a valid coroutine! Did you forget to use the yield keyword?" % inst)
+                "'{}' isn't a valid cocotb coroutine.\n"
+                "Did you forget to decorate your function with `@cocotb.coroutine`?\n"
+                "And also forget a `yield` statement?"
```
eric-wieser, line 123: Perhaps worth just mentioning `async` / `await` here, instead of pointing users to `@cocotb.coroutine`.

ktbarrett:
@eric-wieser What are your thoughts about catching invalid coroutine decorated functions at decoration time? See the first commit and what I had to do in the second commit because we currently assume that can you wrap an invalid coroutine and it won't raise an exception. It's better to be warned early IMO, but it's a change in behavior.

ktbarrett:
@garmin-mjames You're right there is a word salad going on here. "Task" is more appropriate, but I think it's a technicality that will be lost on users. "Yeah, we call them 'coroutines', but there is a difference between the result of calling an `async` function and a `cocotb.coroutine`-decorated generator function; except in just about every way you can use it".

So perhaps you are correct with your renames of the functions, we need a functions that state "can be used as a task" or maybe "compatible with a task interface", or "is task substitutable object".

```python
def is_like_task(thing: Any) -> bool: pass
def is_like_coroutine_function(thing: Any) -> bool: pass
def is_like_trigger(thing: Any) -> bool: pass
```

I have a feeling a scheduler-rewrite is in order for 2.0 and `cocotb.coroutine` will be tossed with it as well. I wouldn't hurt your head trying to rationalize what we have. Perhaps we should move discussion of what we would like to see in a scheduler to a wiki page or shared google doc instead of piece-mealing the discussion across multiple issues and PRs?

cocotb/scheduler.py
```diff
@@ -725,18 +699,7 @@ def _trigger_from_any(self, result) -> Trigger:
...
-                "You likely used the yield keyword instead of await.".format(
-                    result.__qualname__))
-
-        raise TypeError(
-            "Coroutine yielded an object of type {}, which the scheduler can't "
-            "handle: {!r}\n"
-            "Did you forget to decorate with @cocotb.coroutine?"
-            .format(type(result), result)
-        )
+        return self._trigger_from_unstarted_coro(cocotb.decorators.make_task(result))
```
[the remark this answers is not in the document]
ktbarrett: I am, but you're right, it's a bad idea here because it will construct naked generators into tasks, same with `scheduler.add`. Perhaps this whole PR is junk...

ktbarrett:
Turns out there aren't as many use cases as I thought; and it is advantageous to treat each use case differently.

PR #1769, opened 2020-05-05
Make mean example runnable with GHDL simulator (WIP)

m-kru:
I am trying to get mean example work with GHDL simulator. However, when I run the test, I get
```
     0.00ns ERROR    cocotb.regression                         regression.py:386  in _score_test                     Test Failed: mean_randomised_test (result was AttributeError)
                                                                                                                       File "/home/mkru/workspace/cocotb/examples/mean/tests/test_mean.py", line 108, in mean_randomised_test
                                                                                                                         dut.i_data[i] = 0
                                                                                                                       File "/home/mkru/.local/lib/python3.8/site-packages/cocotb-1.4.0.dev0-py3.8-linux-x86_64.egg/cocotb/handle.py", line 286, in __getattr__
                                                                                                                     AttributeError: mean contains no object named i_data
```

I do not know what is wrong, as `mean` entity has `i_data` port.

### DIFF
--- a/examples/mean/hdl/mean.vhd
+++ b/examples/mean/hdl/mean.vhd
@@ -9,7 +9,8 @@ use work.mean_pkg.all;
 
 entity mean is
 generic (
-  BUS_WIDTH : natural := 2);
+  DATA_WIDTH : natural := DATA_WIDTH;
+  BUS_WIDTH  : natural := 2);
 port (
   clk      : in  std_logic;
   rst      : in  std_logic;
--- a/examples/mean/tests/Makefile
+++ b/examples/mean/tests/Makefile
@@ -1,15 +1,14 @@
 
 TOPLEVEL_LANG ?= verilog
 
-ifneq ($(TOPLEVEL_LANG),verilog)
-
-all:
-	@echo "Skipping example due to TOPLEVEL_LANG=$(TOPLEVEL_LANG) not being verilog"
-clean::
-
+ifeq ($(TOPLEVEL_LANG),verilog)
+  VERILOG_SOURCES = $(PWD)/../hdl/mean_sv.sv
+  TOPLEVEL := mean_sv
+else ifeq ($(TOPLEVEL_LANG),vhdl)
+  TOPLEVEL := mean
 else
-
-TOPLEVEL := mean_sv
+  $(error "A valid value (verilog or vhdl) was not provided for TOPLEVEL_LANG=$(TOPLEVEL_LANG)")
+endif
 
 PWD=$(shell pwd)
 
@@ -20,8 +19,6 @@ ifeq ($(SIM),questa)
 COMPILE_ARGS = -mixedsvvh
 endif
 
-VERILOG_SOURCES = $(PWD)/../hdl/mean_sv.sv
-
 MODULE := test_mean
 
 ifneq ($(filter $(SIM),ius xcelium),)
@@ -29,5 +26,3 @@ ifneq ($(filter $(SIM),ius xcelium),)
 endif
 
 include $(shell cocotb-config --makefiles)/Makefile.sim
-
-endif
--- a/examples/mean/tests/test_mean.py
+++ b/examples/mean/tests/test_mean.py
@@ -29,7 +29,7 @@ class StreamBusMonitor(BusMonitor):
 
 
 @cocotb.coroutine
-def clock_gen(signal, period=10):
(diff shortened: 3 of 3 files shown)


elgorwi:
> I am trying to get mean example work with GHDL simulator. However, when I run the test, I get
> 
> ```
>      0.00ns ERROR    cocotb.regression                         regression.py:386  in _score_test                     Test Failed: mean_randomised_test (result was AttributeError)
>                                                                                                                      Traceback (most recent call last):
>                                                                                                                        File "/home/mkru/workspace/cocotb/examples/mean/tests/test_mean.py", line 108, in mean_randomised_test
>                                                                                                                          dut.i_data[i] = 0
>                                                                                                                        File "/home/mkru/.local/lib/python3.8/site-packages/cocotb-1.4.0.dev0-py3.8-linux-x86_64.egg/cocotb/handle.py", line 286, in __getattr__
>                                                                                                                          raise AttributeError("%s contains no object named %s" % (self._name, name))
>                                                                                                                      AttributeError: mean contains no object named i_data
> ```
> 
> I do not know what is wrong, as `mean` entity has `i_data` port.

This is probably due to GHDL not supporting array access through VPI: https://github.com/ghdl/ghdl/issues/237

m-kru:
Sigh, in that case it makes cocotb + GHDL combination unusable in serious project.

elgorwi:
> 
> 
> Sigh, in that case it makes cocotb + GHDL combination unusable in serious project.

Actually, this might be possible after all with a wrapper file. Have a look at #1502 ```matrix_multiplier.vhd``` for some inspiration. 
I recommend trying to create a ```mean_tb.vhd``` having only generics, but no input ports similar to ```matrix_multiplier.vhd```. Then instantiate the existing ```mean.vhd``` and wire it up to the signals (with the same name as the i/o-ports of ```mean.vhd```) defined in the wrapper file. Then in the makefile change the TOPLEVEL to point to the wrapper file.
Also have a look at  #1778, which is doing the same as this PR for SystemVerilog and Icarus. 
Would be great to get this example running on free simulators.

umarcor:
FTR, using VHPIDIRECT instead of VPI, it is currently possible to share a pointer/access to a constrained array between VHDL and C. In https://ghdl.github.io/ghdl-cosim/vhpidirect/examples/arrays.html#matrices, you will find some examples with matrices of integers, matrices of reals/doubles and/or an RGB24 frame buffer. Clicking on the header of each section will take you to the sources of the example.

Both VHPIDIRECT and VPI should be compatible. Hence, you can use cocotb and VPI for signal toggling at runtime, but use VHPIDIRECT for heavy data transfers. For this same reason, I think that cocotb could reuse existing VHDL VCs/BFMs when targeting GHDL. But this is an issue for another day...

eric-wieser:
@umarcor, is vhpidirect a ghdl-exclusive feature? If not, please file a new issue about supporting it. If it is, then it might be a good candidate for a cocotb extension.

umarcor:
> @umarcor, is vhpidirect a ghdl-exclusive feature? If not, please file a new issue about supporting it. If it is, then it might be a good candidate for a cocotb extension.

Yes, it is ghdl-exclusive for now.

It is similar to SystemVerilog's DPI. It has been discussed in the VASG. It didn't make it for VHDL 2019, but it could for the next revision: http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/DpiProposal. I would expect the standarized interface to be "based on VHPIDIRECT", or GHDL to be updated to match the standard. Although I am not familiar with FLI, I believe that it provides a similar interface too.

Hence, I don't think it makes sense to add support in cocotb yet. However, it can be a complement. Exploring how to use it as an extension might be interesting. Comparing it with FLI might be interesting too.

In one of the examples in ghdl-cosim (ref above), there is a VUnit `run.py` script. There, you can see that the only required modification to use VHPIDIRECT in this case is to add `-Wl,main.c` to elaboration: https://github.com/ghdl/ghdl-cosim/blob/master/vhpidirect/arrays/matrices/vunit_axis_vcs/run.py#L13. No need to build any shared library, nor to include any additional header. There are ~20 examples in https://ghdl.github.io/ghdl-cosim/vhpidirect/examples/index.html so you can have a look.

imphil:
Closing this PR as stale.

I'll note, however, that https://github.com/ghdl/ghdl/issues/237 is resolved for a while now, so this could be picked up again if somebody is interested.

PR #1780, opened 2020-05-08
Use last scheduled write to set handle value

garmin-mjames:
Before #1661, scheduled writes were stored in a dictionary which was
keyed by the signal handle. This meant that additional writes scheduled
to the same handle overwrote the previous one.

~~Now that scheduled writes are stored in a callback list, the order of
the callbacks is important. Callbacks should be called in the order that
they were scheduled so that the last scheduled write for a signal handle
wins.~~

Partially revert #1661 and use dictionary again.

Reported by @hofstee in the Gitter lobby.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -391,13 +391,11 @@ class NonHierarchyObject(SimHandleBase):
 
     @value.setter
     def value(self, value):
-        def _call_on_readwrite(f, *args):
-            cocotb.scheduler._schedule_write(self, f, *args)
-        self._set_value(value, _call_on_readwrite)
+        self._set_value(value, cocotb.scheduler._schedule_write)
 
     def setimmediatevalue(self, value):
         """ Assign a value to this simulation object immediately. """
-        def _call_now(f, *args):
+        def _call_now(handle, f, *args):
             f(*args)
         self._set_value(value, _call_now)
 
@@ -407,7 +405,7 @@ class NonHierarchyObject(SimHandleBase):
         This is used to implement both the setter for :attr:`value`, and the
         :meth:`setimmediatevalue` method.
 
-        ``call_sim(f, *args)`` should be used to schedule simulator writes,
+        ``call_sim(handle, f, *args)`` should be used to schedule simulator writes,
         rather than performing them directly as ``f(*args)``.
         """
         raise TypeError("Not permissible to set values on object %s of type %s" % (self._name, type(self)))
@@ -657,7 +655,7 @@ class ModifiableObject(NonConstantObject):
         value, set_action = self._check_for_set_action(value)
 
         if isinstance(value, int) and value < 0x7fffffff and len(self) <= 32:
-            call_sim(self._handle.set_signal_val_long, set_action, value)
+            call_sim(self, self._handle.set_signal_val_long, set_action, value)
             return
         if isinstance(value, ctypes.Structure):
             value = BinaryValue(value=cocotb.utils.pack(value), n_bits=len(self))
@@ -681,7 +679,7 @@ class ModifiableObject(NonConstantObject):
                 "Unsupported type for value assignment: {} ({!r})"
                 .format(type(value), value))
 
-        call_sim(self._handle.set_signal_val_binstr, set_action, value.binstr)
+        call_sim(self, self._handle.set_signal_val_binstr, set_action, value.binstr)
 
     def _check_for_set_action(self, value):
         if not isinstance(value, _SetAction):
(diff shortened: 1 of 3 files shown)


eric-wieser:
Hmm, do you think restoring the dictionary behavior would be more correct? I hasn't thought of repeated writes when I made that change.

hofstee:
I'll just provide the same example here as I did in the gitter. If you have something like the [amba driver](https://github.com/cocotb/cocotb/blob/master/cocotb/drivers/amba.py) and you want to say, pipeline a bunch of writes, one way of doing this is to use `cocotb.fork` and make one fork call per write. Then you can `Combine` or individually join each of the writes.

With how the driver is set up, you'll see this behavior of repeated writes exactly in that case. The mutex is released, the next available write acquires it, and overrides the signals to be driven on the bus.

Personally, that was something I found to be really nice in cocotb - I could swap from a sequential to a pipelined testbench just by forking more coroutines off. I think if the guarantee was removed you would probably need to have some odd `await Timer(0)` calls in there or something. I'm not exactly sure which cocotb events (if any at present) match the Verilog scheduling zone you would need to hit here.

eric-wieser:
To be clear, my question is about what:
```
dut.signal <= 0
dut.signal <= 1
```
should do. In master, it does
```
vpi_set_value_somehow(dut.signal, 1)
vpi_set_value_somehow(dut.signal, 0)
```
which is obviously wrong. With this patch, it does
```
vpi_set_value_somehow(dut.signal, 0)
vpi_set_value_somehow(dut.signal, 1)
```
In the most recent release, it did
```
vpi_set_value_somehow(dut.signal, 1)
```

Which of these last two behaviors do we want?

eric-wieser:
(no release note needed since this is a regression since the last release)

garmin-mjames:
I think the behavior from last release is better. Since we're scheduling writes for a later time, it makes more sense to update the value to be written rather than queue up lots of writes.

garmin-mjames:
I don't think it needs to be an ordered dictionary like it was before, if we go that route.

eric-wieser:
For the sake of determinism I'd rather it were ordered.

cocotb/scheduler.py
```diff
@@ -233,8 +233,8 @@ def __init__(self):
         # Our main state
         self._mode = Scheduler._MODE_NORMAL
 
-        # A list of pending (write_func, args)
-        self._write_calls = []
+        # A dictionary of pending (write_func, args)
```
eric-wieser, line 236: Let's expand this comment so someone doesn't try to make the same simplification as me.

```suggestion
        # A dictionary of pending (write_func, args), one per handle. Only the final write in a timestep
        # propagates to the simulator, all the rest are discarded in python.
```

eric-wieser: (or any similar wording, mine isn't particularly good).

garmin-mjames:
Hmm this isn't as simple as I'd hoped.

For `NonHierarchyIndexableObject`, this sets the dictionary key to the parent handle. We'll have to pass `self` to `call_sim` inside `_set_value`.

https://github.com/cocotb/cocotb/blob/master/cocotb/handle.py#L395

eric-wieser:
Can you elaborate on what you mean by that?

We should probably add a test that
```python
dut.my_array <= [1, 2, 3, 4]
dut,my_array[0] <= 10
await Timer(1)
```
means
```python
dut.my_array <= [10, 2, 3, 4]
```

eric-wieser:
Ah, now I understand the issue. Agreed that `self` needs to be passed in everywhere `call_sim` is called.

eric-wieser:
Although in principle you could extract the raw simulator handle with `func.__self__`.

cocotb/handle.py
```diff
@@ -391,13 +391,13 @@ def value(self):
 
     @value.setter
     def value(self, value):
-        def _call_on_readwrite(f, *args):
-            cocotb.scheduler._schedule_write(self, f, *args)
+        def _call_on_readwrite(handle, f, *args):
+            cocotb.scheduler._schedule_write(handle, f, *args)
         self._set_value(value, _call_on_readwrite)
```
eric-wieser, line 396: Possibly:
```suggestion
        self._set_value(value, cocotb.scheduler._schedule_write)
```

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -263,3 +263,40 @@ async def waiter(event):
...
+
+    @cocotb.coroutine   # TODO: Remove once Combine accepts bare coroutines
+    async def first(dut, event):
+        await Timer(1)
+        dut._log.info("scheduling stream_in_data <= 1")
+        dut.stream_in_data <= 1
+        event.set()
+
+    @cocotb.coroutine   # TODO: Remove once Combine accepts bare coroutines
+    async def second(dut, event):
```
eric-wieser, line 284: Nit: could omit the arguments here and just use the closure

PR #1781, opened 2020-05-09
Avoid calling gpi_load_extra_libs reentrantly

ktbarrett:
Currently libraries specified in GPI_EXTRA are loaded using `gpi_load_extra_libs` which loads extra libs and calls their `register_embed` functions, which calls `gpi_load_extra_libs`... It is re-entrant, hence the funky static `is_loading` variable to prevent loading all the libraries again and calling `gpi_embed_init` more than once. This change makes the loading behavior more normal.

### DIFF
--- a/cocotb/share/lib/embed/gpi_embed.cpp
+++ b/cocotb/share/lib/embed/gpi_embed.cpp
@@ -33,6 +33,7 @@
 #include <cocotb_utils.h>
 #include "embed.h"
 #include "locale.h"
+#include <cassert>
 
 #if defined(_WIN32)
 #include <windows.h>
@@ -119,9 +120,7 @@ extern "C" void embed_init_python(void)
 #define PY_SO_LIB xstr(PYTHON_SO_LIB)
 #endif
 
-    // Don't initialize Python if already running
-    if (gtstate)
-        return;
+    assert(!gtstate);  // this function should not be called twice
 
     void * lib_handle = utils_dyn_open(PY_SO_LIB);
     if (!lib_handle) {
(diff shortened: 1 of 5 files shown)


garmin-mjames:
You could remove this check in `embed_init_python` too, as it's only called from `gpi_load_extra_libs`:
https://github.com/cocotb/cocotb/blob/322e5ec992f541ef92402aa75eeb3174b19f63c6/cocotb/share/lib/embed/gpi_embed.cpp#L122-L123

garmin-mjames:
This may be a good time to rename this function, as it is also responsible for initializing the embedded Python interpreter.

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -118,10 +118,6 @@ extern "C" void embed_init_python(void)
 #define PY_SO_LIB xstr(PYTHON_SO_LIB)
 #endif
 
-    // Don't initialize Python if already running
-    if (gtstate)
-        return;
```
eric-wieser, line 123: Perhaps safest would be `assert(!gtstate, "embed_init_python should not be called twice")`

eric-wieser:
Generally looks good, just one comment.

"Avoid calling gpi_load_extra_libs reentrantly" might be a better commit message.

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -119,9 +120,7 @@ extern "C" void embed_init_python(void)
...
+    assert(!gstate);  // this function should not be called twice
```
eric-wieser, line 123:
```suggestion
    assert(!gtstate);  // this function should not be called twice
```

PR #1690, opened 2020-04-22
Fix scheduler bug

ktbarrett:
Fixes #1688. See the issue for a diagnosis of the problem.

PR #1782, opened 2020-05-09
Scheduler Refactor

### LINKED ISSUE #791: Remove knowledge of tests from the scheduler
Right now, the scheduler has a lot of special cases for scheduling tests in different and exciting ways.

Now that #640 is merged, this isn't really necessary - we can bootstrap the scheduler to support tests with something like:

```python
@cocotb.coroutine
def run_all_tests():
    for test in tests:
        # most of the special casing right now deals with the fact that exceptions could not
        # previously be caught
        log_test_begin(test)
        try:
            yield test()
            raise TestSuccess
        except TestSuccess:
            log_test_pass(test)
        except Exception:
            log_test_failure(test)

        # this is probably the only special-casing within `scheduler` that can't be avoided.
        scheduler.kill_all_forked_coroutines()
        scheduler.unprime_all_triggers()

        yield Timer(1)
```

This would also open the doors to hooking in other test runners, like pytest (#494)

### LINKED ISSUE #1347: Two coroutines that fail on the same trigger are not handled gracefully by the scheduler.
```python
@cocotb.coroutine
async def coroutine1(timer):
    await timer
    assert False

@cocotb.coroutine
async def coroutine2(timer):
    await timer
    assert False

@cocotb.test()
async def test_top(dut):
    timer = triggers.Timer(100, units='ns')                                                                                            
    cocotb.fork(coroutine1(timer))                                                                                                     
    cocotb.fork(coroutine2(timer))
    await timer
```
This will fail when `_force_outcome` is called by the scheduler for the second time.  The scheduler needs to keep track of whether any coroutines that have been scheduled by this trigger have already failed.

### LINKED ISSUE #1675: Test abortion methods don't work
`RunningTest` objects have `abort` methods, the scheduler has the `finish_test` method. I have run several tests and found that neither actual ends the test with the correct outcome. `finish_scheduler` does work, but doesn't clean up before scheduling the next test, causing the next test to fail.

Obviously no one is using these methods or we would have complaints. As a part of the scheduler refactor I vote we remove all of these methods and put a test abortion method in the regression manager (which will be the only things holding a reference to the current test after the refactor).

### LINKED ISSUE #1688: Incomplete cleanup in scheduler leading to bugs
The test that caused the error was added in ktbarrett/cocotb@d589303f9ae8cca17789b21b597e4bec66cf80bf. The failure can be observed by running the failing test and at least one other test after it.

```
cd tests/test_cases/test_cocotb
make MODULE=test_concurrency_primitives COCOTB_SCHEDULER_DEBUG=1 TESTCASE="test_exceptions_forked_late,test_exceptions_first"
``` 
[sim.log](https://github.com/cocotb/cocotb/files/4516438/sim.log)

You can see from the log that the coroutine `test_exceptions_forked_late.<locals>.await_later` that was scheduled in the first test was not cleaned up and fires in the second test and causes it to fail.

### LINKED ISSUE #1803: SimFailure causes test results to be printed multiple times
There's a loop in `RegressionManager` on tests with `SimFailure` result that goes:

`_record_result` -> `tear_down` -> `_record_result` -> ...

And prints out the results multiple times.

See: https://travis-ci.org/github/cocotb/cocotb/jobs/686173872#L704

Found while reviewing #1793.

ktbarrett:
Moves the concept of tests out of the scheduler, closes #791. Fixes #1688 and fixes #1347 by implementing what was talked about in https://github.com/cocotb/cocotb/pull/1690#issuecomment-620932033. Fixes #1803 by not allowing `tear_down` to recurse. Fixes #1675 by implementing the methods correctly.

From the commit message
> This moves the concept of the test out of the scheduler and into the
regression manager. This requires the regression manager to hook all
trigger reactions, so that it can check for test end, do cleanup,
score the test, and enter the regression manager's execution loop
all without re-entry.
>
> To accomplish cleanup, all cleanup operations were moved to a central
place ('cleanup') rather than being smeared across that and
'_check_termination'.
> 
> Test start up requires us to enter the scheduler after react to ensure
consistent scheduling behavior, and also catch immediate test end via
the sole 'check_termination' run by the regression manager's react hook.
Additionally, test following the first are started after a wait of
'Timer(1)' to ensure they start in normal operating mode.

Follow on clean up commits and commits that add tests from #1690 are expected. While the change is mostly split up, this change is not really decomposable, it all works in tandem.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -258,18 +258,20 @@ def _sim_event(level, message):
     from cocotb.result import TestFailure, SimFailure
 
     if level is SIM_TEST_FAIL:
-        scheduler.log.error("Failing test at simulator request")
-        scheduler.finish_test(TestFailure("Failure from external source: %s" %
-                              message))
+        if cocotb.regression_manager is not None:
+            # the regression manager may not have been created (_initialize_testbench failed)
+            msg = "Failing test at request of external source: " + message
+            cocotb.log.error(msg)
+            scheduler.finish_test(TestFailure(msg))
     elif level is SIM_FAIL:
-        # We simply return here as the simulator will exit
-        # so no cleanup is needed
-        msg = ("Failing test at simulator request before test run completion: "
-               "%s" % message)
-        scheduler.log.error(msg)
-        scheduler.finish_scheduler(SimFailure(msg))
+        if cocotb.regression_manager is not None:
+            # the regression manager may not have been created (_initialize_testbench failed)
+            msg = ("Failing remaining tests at request of external source "
+                   "before test run completion: ") + message
+            cocotb.log.error(msg)
+            scheduler.finish_scheduler(SimFailure(msg))
     else:
-        scheduler.log.error("Unsupported sim event")
+        cocotb.log.error("Unsupported sim event")
 
     return True
 
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -262,14 +262,14 @@ def _sim_event(level, message):
             # the regression manager may not have been created (_initialize_testbench failed)
             msg = "Failing test at request of external source: " + message
             cocotb.log.error(msg)
-            scheduler.finish_test(TestFailure(msg))
+            regression_manager._finish_test(TestFailure(msg))
     elif level is SIM_FAIL:
         if cocotb.regression_manager is not None:
             # the regression manager may not have been created (_initialize_testbench failed)
             msg = ("Failing remaining tests at request of external source "
                    "before test run completion: ") + message
             cocotb.log.error(msg)
-            scheduler.finish_scheduler(SimFailure(msg))
+            regression_manager._finish_test(SimFailure(msg))
     else:
         cocotb.log.error("Unsupported sim event")
 
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -95,7 +95,6 @@ class RegressionManager:
         self.count = 0
         self.skipped = 0
         self.failures = 0
-        self._tearing_down = False
 
         # Setup XUnit
         ###################
@@ -232,12 +231,6 @@ class RegressionManager:
             return cocotb.scheduler.add(test)
 
     def tear_down(self) -> None:
-        # prevent re-entering the tear down procedure
-        if not self._tearing_down:
-            self._tearing_down = True
-        else:
-            return
-
         # fail remaining tests
         while True:
             test = self.next_test()
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -478,6 +478,16 @@ class RegressionManager:
 
         cocotb.scheduler.schedule(self._test_task)
 
+    def _finish_test(self, exc: Exception) -> None:
+        """
+        Finishes the test with the given exception as the outcome. Not to be called by the user.
+        """
+        try:
+            self._test_task.abort(exc)
+        except _EscapeHatch:
+            pass
+        self._check_termination()
+
     def _log_test_summary(self) -> None:
 
         if self.failures:
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -45,6 +45,7 @@ from cocotb.xunit_reporter import XUnitReporter
 from cocotb.decorators import test as Test, hook as Hook, RunningTask
 from cocotb.outcomes import Outcome, Error
 from cocotb.handle import SimHandle
+from cocotb.triggers import Trigger, Timer
 
 from cocotb import simulator
 
@@ -95,6 +96,8 @@ class RegressionManager:
         self.count = 0
         self.skipped = 0
         self.failures = 0
+        self._timer1 = Timer(1)
+        self._sim_failed = False
 
         # Setup XUnit
         ###################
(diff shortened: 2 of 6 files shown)


cocotb/__init__.py
```diff
@@ -191,18 +191,18 @@ def _sim_event(level, message):
     from cocotb.result import TestFailure, SimFailure
 
     if level is SIM_TEST_FAIL:
-        scheduler.log.error("Failing test at simulator request")
-        scheduler.finish_test(TestFailure("Failure from external source: %s" %
-                              message))
+        cocotb.log.error("Failing test at simulator request")
+        if cocotb.regression_manager is not None:
```
ktbarrett, line 195: This can occur if the regression_manager hasn't finished being created when the test fails (`_initialize_testbench` exists early). This won't be a problem i teh future after I add the ability to register this callback handler late (last operation in init).

cocotb/regression.py
```diff
@@ -457,7 +448,40 @@ def _start_test(self) -> None:
...
         self._test_start_sim_time = get_sim_time('ns')
-        cocotb.scheduler.add_test(self._test_task)
+
+        cocotb.scheduler._mode = Scheduler._MODE_NORMAL
+        cocotb.scheduler._resume_coro_upon(
+            self._test_task,
+            NullTrigger(name="Start {!s}".format(self._test_task), outcome=Value(None))
+        )
+
+    def tear_down(self) -> None:
```
ktbarrett, line 458: Just moved it where I thought it should be to flow with reading.

ktbarrett:
I can't replicate the seg faults locally. I think the issue is that the refactor has changed when and where the shutdown is being initiated. The shutdown behavior is very complex and I ran into constant issues with it (that I never resolved) when I was working on the GPI. It probably needs to be worked on before I do anything else.

cocotb/__init__.py
ktbarrett: Added comment here.

ktbarrett:
So I realize now this likely does not work. As long as `NullTrigger` reacts immediately by calling the provided callback when primed, `react` *will* be re-entrant, which kind of defeats the purpose of this PR, which was to reduce that while removing the concept of the test from the scheduler. It has been a learning experience for me. And more enforces the idea an interface breaking rewrite is necessary .

ktbarrett:
I love how it works perfectly on my machine, but segfaults after completely all the tests successfully in CI.

ktbarrett:
Valgrind is showing some issue with clean up.
[sim.log](https://github.com/cocotb/cocotb/files/4619830/sim.log)

cocotb/scheduler.py
```diff
@@ -720,7 +720,10 @@ def schedule(self, coroutine, trigger=None):
             self.add(self._pending_coros.pop(0))
 
     def finish_test(self, exc):
```
garmin-mjames, line 722: This can be moved to `RegressionManager` and `finish_scheduler` can be removed as redundant.

garmin-mjames: I would also add a docstring explaining it is designed to be called outside of a scheduled test/task.

ktbarrett: I would like to move them, these functions don't really belong here anymore. @eric-wieser is that permissible? Will add a docstring and make them private methods as well.

cocotb/regression.py
```diff
@@ -425,21 +436,22 @@ def _record_result(
...
     def execute(self) -> None:
+        if self._sim_failed:
+            return self.tear_down()
         while True:
             self._test = self.next_test()
             if self._test is None:
                 return self.tear_down()
 
             self._test_task = self._init_test(self._test)
             if self._test_task:
```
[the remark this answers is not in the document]
ktbarrett: yeah, good point

ktbarrett:
Considering the given error, I will postpone this until we have made handle_vpi_callback and friends memory safe.

ktbarrett:
Added commit to temporarily prevent clearing of the SINGLETON_HANDLES cache which prevents the use-after-free (by not freeing the handles). Freeing the handles isn't necessary as the OS will reclaim the memory once the simulator ends anyway. I will re-introduce the cleanup as a part of #1793, always invoking the free in the shutdown callback will prevent use-after-free..

cocotb/scheduler.py
ktbarrett: So I did the change.

ktbarrett:
@eric-wieser @imphil @garmin-mjames What would be needed to accept this change? I am willing to write up a design document seeing as I want to document the scheduler's behavior for internal documentation anyways.

eric-wieser:
I need to take some time to review this, but my initial feeling is that this design is backwards:
* What you have is a regression manager than spawns multiple schedulers for each tests.
* What I'd envisaged is a scheduler which spawns the regression manager.

Since we don't tear down and restart the simulator state after each test, it seems odd to tear down the `Scheduler` object which aims to track that same state. If we actually _could_ restart the simulator after each regression test, then I'd be in favor of this the approach in this PR.

But again, this is just a hunch of which way round is better, I've forgotten my original rationale.

ktbarrett:
@eric-wieser After your response, me and @andrewandrepowell talked about it for a few minutes. He agreed with you and explained it well. And I can see how it would play into the "big picture" as I see it.

It's a shame because the RegressionManager/Scheduler are already basically partitioned this way, they just needed some extra work. Flipping it would require a lot more work.

eric-wieser:
> And I can see going for this structure would play into the "big picture" as I see it. It's the right way forward.

It might be a good idea to record this discussion in a wiki page or issue somewhere

ktbarrett:
Since we are heading in another direction I'll close this.

Issue #1390, opened 2020-02-16
Windows: Evaluate runtime dynamic linking for VPI/VHPI/FLI libraries


imphil:
To communicate with the simulator, cocotb uses the VPI/VHPI/FLI interface. This interface consists of functions which are implemented by the simulator and called by cocotb's libvpi/libvhpi/libfli. 

On POSIX operating systems, these libraries contain unresolved references to functions like `vpi_*`, which are then resolved by the loader at runtime.  

On Windows, unresolved references in DLLs are not possible. Instead, two options of linking are available: load-time dynamic or implicit linking, or run-time dynamic or explicit linking. Load-time dynamic linking requires a static stub library at compile time called "import libary", which in our case must be provided by the simulator vendor. 

Using the import library therefore requires the simulator to be present at cocotb installation time, a behavior we want to avoid. As a workaround introduced in #1306 , cocotb provides `*.def` files, which can be used to create import libraries. 

The goal of this issue is to explore run-time dynamic or explicit linking, which however requires Windows-specific code. (See https://docs.microsoft.com/en-us/windows/win32/dlls/run-time-dynamic-linking and https://docs.microsoft.com/en-us/cpp/build/linking-an-executable-to-a-dll?view=vs-2019#how-to-link-explicitly-to-a-dll.) 

Potential benefits:
- No dependency on simulators at compile time (potentially not even the DLL name).
- Enablement for a unified libvpi/libvhpi/libfli for all simulators on Windows.

Potential drawback:
- Too much engineering effort, or too fragile.

A potential alternative solution could be a static library stub produced from a `*.def` file which contains all common exported symbols. It's not clear if it is possible to do that without also including the name of the simulator DLL. (And then we're back to multiple libraries.)

ZiaCheemaGit:
still relevant?

ktbarrett:
Still relevant. We maybe have issues with this if we ever start supporting system functions on Windows since the VPI for Icarus needs to link against two different binaries, one for the compiler for it to understand the system function specs, and the other for the compiled simulation binary.

Its a nice to have to that's not immediately relevant, but if someone wanted to work on it would be welcome. It's a good bit of work though.

Issue #1759, opened 2020-05-04
Update vhpi_user.h to follow standard's implementation


ktbarrett:
From https://github.com/cocotb/cocotb/pull/1742#issuecomment-622599691

The standards committee provides an implementation of vhpi_user.h, we should be using their definition if possible. I'm not sure from where the original definition was sourced, or why it was changed along the way. Perhaps these definitions are necessary to make certain simulators work? I think we should investigate.

eric-wieser:
Github has the history a little fragmented, but you can see most of it here: https://github.com/cocotb/cocotb/commits/9ba81fe3f7a129b4888a519df8f9645d811a90a7/include/vhpi_user.h

themperek:
VCS usees this one: https://github.com/jamiekang/embedded-sw-dev-tools/blob/master/binsim/vhpi_user.h (if we ever would like to do VHPI in VCS). So we may finish with few versions,

eric-wieser:
Perhaps we should go the other way and not ship one at all, instead building against the one we find

ktbarrett:
In an ideal world yes we would build against every simulator's vhpi_user.h, but that means we have to write an implementation of VHPI that works with all support's simulator's header impl and then we have to add support for finding that header file for each supported simulator to the build script. That's a lot of work...

I have had the idea floated to me by @leftink that it might actually be easier to have simulator-specific libraries rather than interface-specific libraries. That way we can contain in one object all supported interfaces of a simulator and handle all the weird quirks of that simulator (like not implementing certain callbacks, using special vhpi_users.h, etc.). It might also make supporting new simulators easier since we don't have to try to write an implementation that works for *everyone* and end up breaking someone else in the proces.

PR #1783, opened 2020-05-09
Implementation of runtime dynamic linking for VPI/VHPI on Windows

tpambor:
This PR implements runtime dynamic linking for VPI libraries on Windows.

Potential benefits have already been outlined in #1390. 

`vpi_*/vhpi_*` functions are resolved only once on their first call and then execution is forwarded to the respective implementation. `vpi_*/vhpi_*` wrappers are implemented thread-safe.

Currently, the compile-time dependency on the DLL name still exists to follow closely how symbols are resolved using the old import library approach. The dependency on the DLL name could be dropped by searching in all modules of the current process for `vpi_*/vhpi_*` functions. This works as long as only one module in the process exposes `vpi_*/vhpi_*` as otherwise it would be ambigious which `vpi_*/vhpi_*` function to use. Therefore, vpi libraries are changed to not expose the internal `vpi_*/vhpi_*` wrappers. Searching all modules is left for a follow-up PR to keep this PR smaller.

The `endian_swapper` has been run 10 times to evaluate the performance effect of this PR.
```
Without this PR: SIM / REAL TIME 25227 +- 795 NS/S
With this PR:    SIM / REAL TIME 25156 +- 968 NS/S
```
No significant performance drawback could be observed.


cocotb/share/def/modelsim.def
```diff
@@ -1,115 +0,0 @@
...
-Tcl_ResetResult
-Tcl_GetObjResult
-Tcl_ResetResult
-Tcl_ListObjGetElements
-Tcl_GetStringResult
-TclFreeObj
-Tcl_ResetResult
-Tcl_ResetResult
-Tcl_GetString
-TclFreeObj
```
eric-wieser, line 115: What about all these non-vpi bits?

tpambor: These are loaded from the import library shipped with modelsim `mtipli.lib`. We already dependent on the header files shipped with modelsim so this does not introduce any additional dependency

themperek: I will try to chek but have filling that it has to be done for fli and vhpi (tcl?) also to work without def files. 

We also remove simple def file and add more c code? Is there more advantage than this?

eric-wieser: Agreed, it's not clear to me what the benefit of this is.

tpambor: I have tested it, VPI and FLI for modelsim compile fine. If VHPI is also converted to runtime dynamic linking, all .def files could be dropped and the path to compile with MSVC would be unblocked. Actually, if uncommenting the aldec vhpi library and applying a bunch of hacks (adapting the compiler flags for MSVC, adding the __declspec(dllexport) where necesarry), cocotb can be installed with `pip install .` using the official binary version of python for windows.

Moreover, if the dependency on the DLL name is dropped, the `libcocotbvpi_*` could be easily merged into a single `libcocotbvpi` library.

tpambor: Additionally, icarus master, starting with [`a59b183b`](https://github.com/steveicarus/iverilog/commit/a59b183bd1890eff63cc59d38084dde15de281ff#diff-9bb3b161b30ef5d95427f8ed62cb9e88R685) is no longer working on windows because instead of exporting `vpi_*` in `vvp.exe`, now a function table is passed. As a result, the current `libcocotbvpi_icarus.vpl` is unable to load because it no longer can import the symbols. Based on this PR, `libcocotbvpi_icarus.vpl` could be easily adapted to support both versions of iverilog.

eric-wieser: > Additionally, icarus master, starting with [`a59b183b`](https://github.com/steveicarus/iverilog/commit/a59b183bd1890eff63cc59d38084dde15de281ff#diff-9bb3b161b30ef5d95427f8ed62cb9e88R685) is no longer working on windows because instead of exporting `vpi_*` in `vvp.exe`, now a function table is passed. As a result, the current `libcocotbvpi_icarus.vpl` is unable to load because it no longer can import the symbols. Based on this PR, `libcocotbvpi_icarus.vpl` could be easily adapted to support both versions of iverilog.

Could you create a new issue to explain and track this?

cocotb_build_libs.py
```diff
@@ -363,24 +368,37 @@ def get_ext():
...
-                library_dirs=modelsim_extra_lib_path,
-                sources=[
-                    os.path.join(share_lib_dir, "fli", "FliImpl.cpp"),
-                    os.path.join(share_lib_dir, "fli", "FliCbHdl.cpp"),
-                    os.path.join(share_lib_dir, "fli", "FliObjHdl.cpp"),
-                ],
-                extra_link_args=_extra_link_args(lib_name=lib_name, rpaths=["$ORIGIN"]),
-                extra_compile_args=_extra_cxx_compile_args,
-            )
+            if _can_load_library(modelsim_libmtipli_path):
```
themperek, line 378: Not sure if `_can_load_library` is good idea. If I am not mistaken (I can easily be) we do not need to load libraries here (on POSIX this would mean `libmtipli.so` is available to load which may not be the case at this point on Windows `modelsim_lib_path` is on `PATH`). If `mti.h` is available we can probably assume "dll/so" also. Alternative just check `os.path.isfile`?

cocotb/share/def/modelsim.def
themperek: >Additionally, icarus master, starting with a59b183b is no longer working on windows

This is a good reason to change.

>Moreover, if the dependency on the DLL name is dropped, the libcocotbvpi_* could be easily merged into a single libcocotbvpi library.

This is may hard to do for other/compatibility reasons, unfortunately.

themperek: >I have tested it, VPI and FLI for modelsim compile fine. If VHPI is also converted to runtime dynamic linking, all .def files could be dropped and the path to compile with MSVC would be unblocked. Actually, if uncommenting the aldec vhpi library and applying a bunch of hacks (adapting the compiler flags for MSVC, adding the __declspec(dllexport) where necesarry), cocotb can be installed with pip install . using the official binary version of python for windows.

Do you know if such dll does actually work/loads without issues?

eric-wieser: Are def files not compatible with msvc? I thought they were.

themperek:
Some very basic crosscheck on windows shows that it is working fine. We will do more if we are ready to merge.

cocotb/share/lib/vpi/VpiTrampoline.cpp
```diff
@@ -0,0 +1,251 @@
...
+* ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+* WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+* DISCLAIMED. IN NO EVENT SHALL POTENTIAL VENTURES LTD BE LIABLE FOR ANY
+* DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+* (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+* LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+* ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************************************************************/
```
cmarqu, line 26: This should be
```
```
for new files (see ``CONTRIBUTING.md``).

imphil:
@tpambor Thanks a lot for this PR, what a great initial cocotb contribution! I'll have a closer look in the coming days, but I certainly appreciate the work to get rid of the tool-specific `def` files, and removing the dependency on the tool being present at installation time.

cocotb_build_libs.py
tpambor: This is not a about checking whether the file is present but rather if it is compatible. If 32-bit modelsim is installed and a 64-bit python environment is used `pip install` results in a linker error because the 32-bit `libmtipli` cannot be linked with the 64-bit `libcocotbfli_modelsim`. Rather than stoping with a linker error I prefer to throw a warning and skip `libcocotbfli_modelsim`. `libmtipli.so` is loaded from the folder which also contains `vdbg`. An absolute path is used and therefore it should not be necessary to have different `LD_LIBRARY_PATH`. It works for me on Windows and Linux.

themperek: OK now I get it. In this case probbaly a line of comment would be good.

cocotb/share/def/modelsim.def
tpambor: > Are def files not compatible with msvc? I thought they were.

It would be possible. `dlltool` is not part of the msvc toolchain and therefore would need to be substituted. Probably, it would also be necessary to locate the msvc toolchain because typically it is not on `PATH`. IMHO the extra maintenance burden of this way would object it. Whereas this PR reduces dependencies instead of introducing new ones. VPI/VHPI APIs have not really changed in years and therefore I do not expect much maintenance effort in having a wrapper for runtime loading.

tpambor:
I am going to rebase this PR on #1796 to clean it up a little more

tpambor:
Changes:
- Rebased on #1796
- Squashed commetics, cleanup and SPDX license header changes
- Added comment with explanation for `_can_load_library`
- No longer touches `vpi_user.h`, `vhpi_user.h` to keep the version from the standard

cocotb/share/lib/vpi/VpiTrampoline.cpp
```diff
@@ -0,0 +1,228 @@
...
+            static HMODULE hModule = GetModuleHandle("aldecpli.dll");
+#elif defined(GHDL)
+            static HMODULE hModule = GetModuleHandle("libghdlvpi.dll");
+#elif defined(ICARUS)
+            static HMODULE hModule = GetModuleHandle("vvp.exe");
+#elif defined(MODELSIM)
+            static HMODULE hModule = GetModuleHandle("mtipli.dll");
+#else
+#error No target module defined for trampoline
+#endif
```
eric-wieser, line 158: Is this any better than having the names in the `.def` files?

cocotb/share/def/modelsim.def
eric-wieser: Right, the MSVC equivalent is `lib /def:modelsim.def /out:modelsim.lib /machine:x86`

cocotb/share/lib/vpi/VpiTrampoline.cpp
tpambor: This is here for compatibility to not break any of the existing tools. The symbols are resolved here in the same way as the OS does. In the future the module names can be removed by searching for `vpi_*` in the whole process space but this works only if there is only one module exposing `vpi_*` as otherwise it would be ambiguous which to select. As a consequence this requires testing in all tools and is therefore left for a future PR.

themperek:
Since Icarus 11 is released it would be important to put this in.

@eric-wieser Could you have a look?

cocotb/share/lib/vpi/VpiTrampoline.cpp
```diff
@@ -0,0 +1,228 @@
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "VpiImpl.h"
+#include <cinttypes>
+#include <Windows.h>
+
+/* Define function pointers for routines */
```
imphil, line 9: Can you expand this comment a bit explaining what this file actually does? It's a rather niche setup we have here, and I'm sure many people will be confused why we have such an "interesting" machinery in here. (Or add more documentation to the `VpiTrampoline` class below.)

cocotb/share/lib/vpi/VpiTrampoline.cpp
imphil: Can you add this piece of information to the function as comment please?

cocotb_build_libs.py
```diff
@@ -495,29 +508,45 @@ def get_ext():
...
-                define_macros=[("COCOTBFLI_EXPORTS", "")],
-                include_dirs=[include_dir, modelsim_include_dir],
-                libraries=["gpi", "gpilog", "stdc++", "mtipli"],
-                library_dirs=modelsim_lib_path,
-                sources=fli_sources,
-                extra_link_args=_extra_link_args(lib_name=lib_name, rpaths=["$ORIGIN"]),
-                extra_compile_args=_extra_cxx_compile_args,
-            )
+            # Check if the library is compatible with our current architecture
+            # e.g. 32-bit libmtipli and 32-bit python runtime
```
imphil, line 495: python -> Python

cocotb_build_libs.py
```diff
@@ -495,29 +508,45 @@ def get_ext():
...
+            # To check can load the library into our current python runtime
+            # if it works we are sure that it is compatible
```
imphil, line 497: Small wording nit: "To check if the FLI library can be loaded, we simply load it. This catches cases where there is a mismatch between the library architecture and the Python architecture (i.e. a 32 bit FLI library used on 64 bit Python)."

cocotb_build_libs.py
```diff
@@ -495,29 +508,45 @@ def get_ext():
...
+                    extra_compile_args=_extra_cxx_compile_args,
+                )
 
-            ext.append(fli_ext)
+                ext.append(fli_ext)
+            else:
+                logger.warning(
+                    "Cannot build FLI interface for Modelsim/Questa - "
+                    "the mtipli library for '{}' at '{}' cannot be loaded.".format(
+                        vsim_path, modelsim_libmtipli_path
```
imphil, line 523: Is there more information in what is actually going wrong in the OSError that you catch in _can_load_library?

imphil:
So here we go ... Thanks @tpambor for your PR and the massive amount of work you put into this PR. As expected, the solution isn't a piece of beauty, but it looks like it could do the trick. It's a shame I didn't have time to look into this when you first opened this PR, but better late than never!

Overall, this PR looks good and I'm ready to merge it without major changes.

In addition to the comments inline, some high-level points:
* The C/C++ code has some really long lines and other formatting issues where it doesn't match our typical style. Instead of manually formatting them right now, I'd propose to get this PR in and then do a full clang-format run on all of our C/C++ code. I filed https://github.com/cocotb/cocotb/issues/2189 to keep track of that.
* The duplication of function declarations/definitions in the trampolines is unfortunate. We should look into macro magic (including a file multiple times with right helper macros set), or a generator script. This can be done in a follow-up, though.
* Can you please rename the VpiTrampoline and VhpiTrampoline files to have Windows in their name, e.g. VpiTrampolineWindows?

cocotb/share/lib/vhpi/VhpiImpl.h
```diff
@@ -35,9 +35,32 @@
 #define COCOTBVHPI_EXPORT COCOTB_IMPORT
 #endif
 
-#include "../gpi/gpi_priv.h"
+// On Windows vhpi_* are provided by VhpiTrampoline, therefore do not use dllimport
+#if !defined(__linux__) && !defined(__APPLE__)
+#define PLI_DLLISPEC
+
+// Workaround broken WIN32 detection in vhpi_user.h
```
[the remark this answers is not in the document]
garmin-mjames: xref #1759

PR #1788, opened 2020-05-11
vpiFinish and vhpiFinish require an argument

tpambor:
IEEE.1364-2005 says "Causes the $finish built-in Verilog system task to be executed upon return of the application routine. This operation shall be passed one additional integer argument, which is the same as the diagnostic message level argument passed to $finish".

Add the required integer argument. `1` is used as it is the default argument which is used when `$finish` is called without an argument in verilog. Same applies to vhpi.


imphil:
The change itself looks good, thanks! Would you mind adding either a comment or (preferred) constants to give the `1` value a more descriptive name?

For others reading along, that's the SV2017 spec table of allowed values:

image

themperek:
It seems to hang VCS.

cocotb/share/include/vhpi_user.h
```diff
@@ -1122,6 +1122,14 @@ typedef enum
...
 } vhpiSimControlT;
 
+/* arguments for vhpiStop or vhpiFinish */
+typedef enum
+{
+  vhpiDiagNone          = 0, /* prints nothing */
+  vhpiDiagTimeLoc       = 1, /* prints simulation time and location */
+  vhpiDiagTimeLocCPUMem = 2  /* prints simulation time, location and statistics about CPU and memory usage */
+} vhpiDiagT;
+
```
eric-wieser, line 1132: If these are symbols we invented that are not part of the standard vhpi header, then I do not think we should put them here. If they came from some non-standard extension of that header, can you link to its source?

imphil: Yes, these headers should stay a copy of the ones in the standard. We have `vpi_user_ext.h` for VPI, could you add the VPI part there, and create a similar header for VHPI?

garmin-mjames:
The VHDL 2008 standard:

> If command is vhpiFinish, after control returns to the tool from the callback function from
which the vhpi_control function was invoked, the tool enters the termination phase (see 20.10).
Additional arguments to vhpi_control beyond the command argument may be interpreted by
the tool in an implementation-defined manner. The tool shall provide an implementation-defined
default action if no additional arguments are provided.

So additional arguments are simulator specific and there's default behavior when there are no extra arguments. I would vote to leave off additional arguments for `vhpi_control` to keep the broadest support.

imphil:
@tpambor since this seems to cause (yet unknown) trouble with VCS and VHPI doesn't seem to need it either: what was the motivation for this change? Correctness with regards to the VPI spec, or a problem with a real simulator out there?

tpambor:
There hasn't be a problem with a real simulator. I did see the problem more from a code point of view. `vpi_control` is a variadic function. For C variadic functions the callee cannot know how many arguments get passed, this always has to determined by some other function argument or a specification, in this case the VPI spec. So if we do not pass an extra argument the callee does not know and nevertheless reads an yet undefined value from either a CPU register or the stack. 

@themperek Maybe you can test passing `0`, `2` and see if it doesn't hang and we could stick with a different argument.

themperek:
So it is a problem on the master if I am not mistaken.

themperek:
So. Sorry for confusion the problem was #1812. Now everything works.

What is also interesting is that it looks VCS with this PR now with status 0 (was always 4 and I guess nobody knew why?). So `-` here is not needed: https://github.com/cocotb/cocotb/blob/56586c4c3fd878f9ada6fc170b9a4e46dfee15a9/cocotb/share/makefiles/simulators/Makefile.vcs#L84
 :+1:

cocotb/share/include/vhpi_user.h
tpambor: Done

PR #1789, opened 2020-05-11
GPI embed cleanup

ktbarrett:
Moves `TOPLEVEL`, `LANGUAGE`, `argc`, and `argv` cocotb module attributes from C into Python. There wasn't much of a justification for having this stuff in C to begin with. Also cleans up the `embed_sim_init`'s ref-counting and removes the currently pointless `FENTER` and `FEXIT` macros from the C codebase.

Following PRs will almost move the responsibility of registering `_sim_event` as the callback for simulator event, and the registering of the cocotb logger as the logger for the GPI, from C into Python. The only thing in gpi_embed should be interpreter initialization, loading the entry module and entry function (which will be made customizable), calling the entry function with the argument list, and interpreter finalization.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -125,7 +125,7 @@ def mem_debug(port):
     cocotb.memdebug.start(port)
 
 
-def _initialise_testbench(root_name):
+def _initialise_testbench():
     """Initialize testbench.
 
     This function is called after the simulator has elaborated all
@@ -138,6 +138,14 @@ def _initialise_testbench(root_name):
     comma-separated list of modules to be executed before the first test.
     """
     _rlock.acquire()
+
+    root_name = os.getenv("TOPLEVEL")
+    if root_name is not None:
+        if root_name == "":
+            root_name = None
+        elif '.' in root_name:
+            # Skip any library component of the toplevel
+            root_name = root_name.split(".", 1)[1]
 
     from cocotb import simulator
 
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -117,6 +117,8 @@ fork = scheduler.add
 # FIXME is this really required?
 _rlock = threading.RLock()
 
+LANGUAGE = os.getenv("TOPLEVEL_LANG")
+
 
 def mem_debug(port):
     import cocotb.memdebug
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -125,7 +125,7 @@ def mem_debug(port):
     cocotb.memdebug.start(port)
 
 
-def _initialise_testbench():
+def _initialise_testbench(argv_):
     """Initialize testbench.
 
     This function is called after the simulator has elaborated all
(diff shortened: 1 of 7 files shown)


cocotb/__init__.py
```diff
@@ -139,6 +139,13 @@ def _initialise_testbench(root_name):
     """
     _rlock.acquire()
 
+    root_name = os.getenv("TOPLEVEL", None)
```
garmin-mjames, line 146: `os.getenv` defaults to `None`.

cocotb/__init__.py
```diff
@@ -139,6 +139,13 @@ def _initialise_testbench(root_name):
...
+    if root_name is not None:
+        if root_name == "":
+            root_name = None
+        elif '.' in root_name:
+            root_name = root_name.split(".", 1)[1]
```
garmin-mjames, line 151: An explanatory comment would be nice here.

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -258,11 +258,6 @@ extern "C" int embed_sim_init(int argc, char const * const * argv)
         LOG_ERROR("Failed to get the _log_from_c function");
         goto cleanup;
     }
-    if (!PyCallable_Check(simlog_func)) {
-        LOG_ERROR("_log_from_c is not callable");
-        Py_DECREF(simlog_func);
-        goto cleanup;
-    }
```
eric-wieser, line 281: Just a note - this is safe anyway, as if it's not callable we'll get a `TypeError` later, which isn't any different really from the function itself failing.

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -299,20 +299,15 @@ extern "C" int embed_sim_init(int argc, char const * const * argv)
...
-            Py_DECREF(argv_list);
             goto cleanup;
         }
         PyList_SET_ITEM(argv_list, i, argv_item);                       // Note: This function steals the reference to argv_item
     }
 
-
     cocotb_retval = PyObject_CallFunctionObjArgs(cocotb_init, argv_list, NULL);
-    Py_DECREF(argv_list);
-    Py_DECREF(cocotb_init);
```
eric-wieser: I'm not really a fan of this change (read: this commit as a whole) - I think it's more instructive to see things cleaned up as soon as they're no longer needed

ktbarrett: That's a great sentiment, but that approach is more verbose, less likely to be correct, and more fragile, considering how the function is implemented. If we had a considerable amount of wrapper code, I might say we should use `std::unique_ptr` and introduce new scopes for scope-guard-style code.

eric-wieser: Would you mind popping that commit off so I can merge this, then I'll look at it again in a standalone PR?

ktbarrett:
Travis status isn't updating?

cocotb/share/lib/embed/gpi_embed.cpp
eric-wieser: Thanks!

PR #1790, opened 2020-05-11
Remove E126 and E127 from flake8 ignore list

cmarqu:
Fix occurrences of E127; E126 did not occur.
Refs #1547.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -706,14 +706,14 @@ class TestFactory:
                     doc += "\t%s: %s\n" % (optname, repr(optvalue))
 
             self.log.debug("Adding generated test \"%s\" to module \"%s\"" %
-                             (name, mod.__name__))
+                           (name, mod.__name__))
             kwargs = {}
             kwargs.update(self.kwargs_constant)
             kwargs.update(testoptions)
             if hasattr(mod, name):
                 self.log.error("Overwriting %s in module %s. "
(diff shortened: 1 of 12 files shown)


cocotb/regression.py
```diff
@@ -706,14 +706,14 @@ def generate_tests(self, prefix="", postfix=""):
...
             kwargs.update(self.kwargs_constant)
             kwargs.update(testoptions)
             if hasattr(mod, name):
                 self.log.error("Overwriting %s in module %s. "
-                                 "This causes a previously defined testcase "
-                                 "not to be run. Consider setting/changing "
-                                 "name_postfix" % (name, mod))
+                               "This causes a previously defined testcase "
+                               "not to be run. Consider setting/changing "
+                               "name_postfix" % (name, mod))
```
ktbarrett, line 717: It's a matter of preference, but I really dislike this style. It forces large diffs and pushes content to the right of the screen.
```suggestion
                self.log.error(
                    "Overwriting %s in module %s. This causes a previously defined testcase "
                    "not to be run. Consider setting/changing name_postfix" % (name, mod))
```

eric-wieser: Better to not use `%` at all here, it's built into the logger.

cmarqu: While I'm fine with both comments, let's do this in another PR - there are many more such chances to clean up, just in this file alone.

PR #1792, opened 2020-05-12
Closes #913: fix setting ints to signals wider than 32-bit (Python only solution)

### LINKED ISSUE #913: Assigning Python ints to signals >32 bits

elgorwi:
- Fix handling of negative numbers for signal widths wider than 32-bit (issue #913)
- Add value range check
- Add testcases
- Add documentation about signed and unsigned values and supported range

Alternative solution to #1766 as discussed here: https://github.com/cocotb/cocotb/pull/1766#issuecomment-627008793

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -34,7 +34,7 @@ import warnings
 
 import cocotb
 from cocotb import simulator
-from cocotb.binary import BinaryValue
+from cocotb.binary import BinaryValue, BinaryRepresentation
 from cocotb.log import SimLog
 from cocotb.result import TestError
 
@@ -645,18 +645,32 @@ class ModifiableObject(NonConstantObject):
                 The value to drive onto the simulator object.
 
         Raises:
-            TypeError: If target is not wide enough or has an unsupported type
-                 for value assignment.
+            TypeError: If target has an unsupported type for value assignment.
+
+            OverflowError: If int value is out of the range that can be represented
+                by the target. -2**(Nbits - 1) <= value <= 2**Nbits - 1
         """
         value, set_action = self._check_for_set_action(value)
 
-        if isinstance(value, int) and value < 0x7fffffff and len(self) <= 32:
-            call_sim(self, self._handle.set_signal_val_long, set_action, value)
-            return
+        if isinstance(value, int):
+            min_value = -2 ** (len(self) - 1)  # minimum negative value for 'signed' number
+            max_value = 2 ** len(self) - 1     # maximum positive value for 'unsigned' number
+
+            if min_value <= value <= max_value:
+                if len(self) <= 32 and value < 0x7fffffff:  # set_signal_val_long() does not support full 32-bit 'unsigned' range
+                    call_sim(self, self._handle.set_signal_val_long, set_action, value)
+                    return
+                elif value < 0:
+                    value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.TWOS_COMPLEMENT)
+                else:
+                    value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.UNSIGNED)
+            else:
+                raise OverflowError(
+                    "Int value ({!r}) out of range for assignment of {!r}-bit signal ({!r})"
+                    .format(value, len(self), self._name))
+
         if isinstance(value, ctypes.Structure):
             value = BinaryValue(value=cocotb.utils.pack(value), n_bits=len(self))
-        elif isinstance(value, int):
-            value = BinaryValue(value=value, n_bits=len(self), bigEndian=False)
         elif isinstance(value, dict):
             # We're given a dictionary with a list of values and a bit size...
             num = 0
--- a/documentation/source/quickstart.rst
+++ b/documentation/source/quickstart.rst
@@ -152,6 +152,42 @@ writes are not applied immediately, but delayed until the next write cycle.
 Use ``sig.setimmediatevalue(new_val)`` to set a new value immediately
 (see :meth:`~cocotb.handle.NonHierarchyObject.setimmediatevalue`).
 
+Signed and unsigned values
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Both signed and unsigned values can be assigned to signals using a Python int.
+Cocotb makes no assumptions regarding the signedness of the signal. It only
+considers the width of the signal, so it will allow values in the range from
+the minimum negative value for a signed number up to the maximum positive
+value for an unsigned number: ``-2**(Nbits - 1) <= value <= 2**Nbits - 1``
+Note: assigning out-of-range values will raise an :exc:`OverflowError`.
+
+A :class:`BinaryValue` object can be used instead of a Python int to assign a
+value to signals with more fine-grained control (e.g. signed values only).
+
+.. code-block:: verilog
+
+    module my_module (
+        input   logic       clk,
+        input   logic       rst,
+        input   logic [2:0] data_in,
+        output  logic [2:0] data_out
+        );
+
+.. code-block:: python3
+
+    # assignment of negative value
+    dut.data_in <= -4
+
+    # assignment of positive value
+    dut.data_in <= 7
+
+    # assignment of out-of-range values
+    dut.data_in <= 8   # raises OverflowError
+    dut.data_in <= -5  # raises OverflowError
+
+Forcing and freezing signals
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 In addition to regular value assignments (deposits), signals can be forced
 to a predetermined value or frozen at their current value. To achieve this,
 the various actions described in :ref:`assignment-methods` can be used.
--- a/tests/test_cases/test_cocotb/test_handle.py
+++ b/tests/test_cases/test_cocotb/test_handle.py
@@ -70,6 +70,147 @@ async def test_delayed_assignment_still_errors(dut):
         dut.stream_in_int <= []
 
 
+async def int_values_test(signal, values):
+    """Test integer access to a signal."""
+
+    log = logging.getLogger("cocotb.test")
+    for val in values:
+        signal <= val
+        await Timer(10, 'ns')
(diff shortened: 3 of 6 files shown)


tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -70,6 +71,178 @@ async def test_delayed_assignment_still_errors(dut):
         dut.stream_in_int <= []
 
 
+async def int_values_test(signal, values):
+    """
+    Test integer access to a signal
+    """
+    log = logging.getLogger("cocotb.test")
+    for val in values:
+        signal <= val
```
elgorwi, line 80: Interestingly, if this line is changed to ```signal = val```, running a test will result in following error:
```
#     20.00ns ERROR    Test Failed: test_int_8bit (result was AttributeError)
#
#                      Traceback (most recent call last):
#
#                        File "C:/cocotb/tests/test_cases/test_cocotb/test_handle.py", line 117, in test_int_8bit
#
#                          await int_values_test(dut.stream_in_data, values)
#
#                        File "C:/cocotb/tests/test_cases/test_cocotb/test_handle.py", line 84, in int_values_test
#
#                          got = signal.value.signed_integer
#
#                      AttributeError: 'int' object has no attribute 'value'
```
Is this expected?

ktbarrett: Yes that's expected. That is just binding the variable `signal` to the value of `val`, not setting the value of the signal.

cocotb/handle.py
```diff
@@ -651,14 +651,30 @@ def _set_value(self, value, call_sim):
...
+            OverflowError: If int value is out of the range that can be represented
+                by the target. -2**(Nbits - 1) <= value <= 2**Nbits - 1
         """
         value, set_action = self._check_for_set_action(value)
 
-        if isinstance(value, int) and value < 0x7fffffff and len(self) <= 32:
-            call_sim(self._handle.set_signal_val_long, set_action, value)
-            return
+        if isinstance(value, int):
+            min_value = -2 ** (len(self) - 1)  # minimum negative value for 'signed' number
```
ktbarrett, line 662:
```suggestion
            min_value = -(2 ** (len(self) - 1))  # minimum negative value for 'signed' number
```
negation is higher precedence than power, `-2 ** 3 = -8` in Python.

eric-wieser: That's not an interesting example, perhaps you meant `-2 ** 2 = -4`, which I think shows the opposite of what you claim

ktbarrett: Yep, not entirely sure what I was thinking. I feel like I got bit by something related to that before and am remembering wrong.

tests/test_cases/test_cocotb/test_handle.py
elgorwi: Ok, thanks for confirming

cocotb/handle.py
```diff
@@ -651,14 +651,30 @@ def _set_value(self, value, call_sim):
                 The value to drive onto the simulator object.
 
         Raises:
-            TypeError: If target is not wide enough or has an unsupported type
-                 for value assignment.
+            TypeError: If target has an unsupported type for value assignment.
+
+            OverflowError: If int value is out of the range that can be represented
+                by the target. -2**(Nbits - 1) <= value <= 2**Nbits - 1
```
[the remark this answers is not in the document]
elgorwi: No, there isn’t, at least not when working with simple logic vectors, rather than signed/unsigned types in vhdl or using signed in verilog.
I think we should allow setting the widest int range possible given the number of bits. If the user needs finer grained control, BinaryValue should be used instead of int.

cocotb/handle.py
elgorwi: Pretty sure the code works as intended

cocotb/handle.py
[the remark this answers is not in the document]
eric-wieser: So vpi cannot detect the "signed" keyword on a signal declaration?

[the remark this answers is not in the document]
elgorwi: I would not be surprised if vpi can detect this, but the point is that the RTL may be using ```std_logic_vector(7 downto 0)``` rather than ```signed(7 downto 0)``` or ```unsigned(7 downto 0)```, so we can’t infer if the vector is assuming signed or unsigned values.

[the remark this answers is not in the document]
elgorwi: Just added some documentation to clarify the behaviour of assigning signals using ints to the user.

documentation/source/quickstart.rst
```diff
@@ -258,6 +258,42 @@ writes are not applied immediately, but delayed until the next write cycle.
 Use ``sig.setimmediatevalue(new_val)`` to set a new value immediately
 (see :meth:`~cocotb.handle.ModifiableObject.setimmediatevalue`).
 
+Signed and unsigned values
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Both signed and unsigned values can be assigned to signals using a Python int.
+Cocotb makes no assumptions regarding signed/unsigned of the signal. It only
+considers the width of the signal, so it will allow values in the range from
+the minumum negative value for a signed number up to the maximum positive
```
cmarqu, line 267:
```suggestion
the minimum negative value for a signed number up to the maximum positive
```

documentation/source/quickstart.rst
```diff
@@ -258,6 +258,42 @@ writes are not applied immediately, but delayed until the next write cycle.
 Use ``sig.setimmediatevalue(new_val)`` to set a new value immediately
 (see :meth:`~cocotb.handle.ModifiableObject.setimmediatevalue`).
 
+Signed and unsigned values
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Both signed and unsigned values can be assigned to signals using a Python int.
+Cocotb makes no assumptions regarding signed/unsigned of the signal. It only
```
cmarqu, line 265:
```suggestion
Cocotb makes no assumptions regarding the signedness of the signal. It only
```
?

documentation/source/quickstart.rst
```diff
@@ -258,6 +258,42 @@ writes are not applied immediately, but delayed until the next write cycle.
...
+considers the width of the signal, so it will allow values in the range from
+the minumum negative value for a signed number up to the maximum positive
+value for an unsigned number. ``-2**(Nbits - 1) <= value <= 2**Nbits - 1``
+Note: assigning out-of-range values will raise an ``OverflowError``.
```
cmarqu, line 269:
```suggestion
value for an unsigned number: ``-2**(Nbits - 1) <= value <= 2**Nbits - 1``
Note: assigning out-of-range values will raise an :exc:`OverflowError`.
```

documentation/source/quickstart.rst
```diff
@@ -258,6 +258,42 @@ writes are not applied immediately, but delayed until the next write cycle.
...
+
+A :class:`BinaryValue` object can be used instead of a Python int to assign a
+value to signals with more fine-grained control (eg. signed values only).
```
cmarqu, line 272:
```suggestion
value to signals with more fine-grained control (e.g. signed values only).
```

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -70,6 +70,180 @@ async def test_delayed_assignment_still_errors(dut):
         dut.stream_in_int <= []
 
 
+async def int_values_test(signal, values):
+    """
+    Test integer access to a signal
+    """
```
cmarqu, line 76: Nit
```suggestion
    """Test integer access to a signal."""
```
(from PEP 257)

documentation/source/quickstart.rst
elgorwi: OK, done

documentation/source/quickstart.rst
elgorwi: OK, done

documentation/source/quickstart.rst
elgorwi: OK, done

documentation/source/quickstart.rst
elgorwi: OK, done

tests/test_cases/test_cocotb/test_handle.py
elgorwi: OK, done.

cocotb/handle.py
```diff
@@ -645,18 +645,32 @@ def _set_value(self, value, call_sim):
...
+                    call_sim(self, self._handle.set_signal_val_long, set_action, value)
+                    return
+                elif value < 0:
+                    value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.TWOS_COMPLEMENT)
+                else:
+                    value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.UNSIGNED)
+            else:
+                raise OverflowError(
+                    "Int value ({!r}) out of range for assignment of {!r}-bit signal ({!r})"
+                    .format(value, len(self), self._name))
```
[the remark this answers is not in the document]
elgorwi: That is correct. And when setting vectors < 32 bit, there was even silent truncation.

elgorwi:
@imphil can you review?

Issue #19, opened 2013-06-25
Consolidate examples


chiggs:
The examples also double as a regression to ensure we don't break the APIs. We probably want only 2 examples in the repo:

1 - A set of separate tests that exercise all of the available features
- Communication between coroutines using events
- Fork/Join simultaneous execution
- All the various trigger types
- Combining triggers
- Test timeout
- Expected test failure etc
- Use of Driver, Monitor, Scoreboard class etc

2 - An example of a verified block with a suite of tests

Example 1 ("features") would also work as a tutorial / introduction to cocotb.

ktbarrett:
I think we have decided against only having 2 examples in the repo. We have several examples and a slew of unit tests. I don't think a single example could act as a good unit test (or a good example) either.

PR #1793, opened 2020-05-12
Cleanup shutdown behavior in GPI

ktbarrett:
Shutdown cleanup was simplified to depend on the shutdown callback handler. This effectively side-steps the use-after-free issue seen in #1782.

It also unconditionally sends cocotb a sim event during shut down. This will be important once multiple GPI users are supported. They all need to be notified of a shut down.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -285,8 +285,8 @@ def _sim_event(level, message):
             cocotb.log.error(msg)
 
     elif level is SIM_FAIL:
-        if regression_manager is None:  # pragma: no cover
-            # Sim failure/end before regression manager started. Should never happen.
+        if regression_manager is None:
+            # Sim failure/end before regression manager started. Occurs when initialization fails.
             cocotb.log.error("Regression ended before Regression Manager started: " + message)
 
         elif not regression_manager.finished:
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -269,18 +269,40 @@ def _sim_event(level, message):
     from cocotb.result import TestFailure, SimFailure
 
     if level is SIM_TEST_FAIL:
-        scheduler.log.error("Failing test at simulator request")
-        scheduler.finish_test(TestFailure("Failure from external source: %s" %
-                              message))
+        if regression_manager is None:  # pragma: no cover
+            # Test failure before regression manager started. Should never happen.
+            cocotb.log.error("Test failed before Regression Manager started: " + message)
+
+        elif not regression_manager.finished:
+            # Test failure during a test, fail it
+            msg = "Test failure from external source: " + message
+            cocotb.log.error(msg)
+            scheduler.finish_test(TestFailure(msg))
+
+        else:  # pragma: no cover
+            # Test failure after the regression manager finishes. Should never happen.
+            msg = "Received test failure, but all tests have finished: " + message
+            cocotb.log.error(msg)
+
     elif level is SIM_FAIL:
-        # We simply return here as the simulator will exit
-        # so no cleanup is needed
-        msg = ("Failing test at simulator request before test run completion: "
-               "%s" % message)
+        if regression_manager is None:  # pragma: no cover
+            # Sim failure/end before regression manager started. Should never happen.
+            cocotb.log.error("Regression ended before Regression Manager started: " + message)
+
+        elif not regression_manager.finished:
+            # Sim failure/end during test. Fail the current test and all the rest
+            msg = "Failing test at simulator request before test run completion: " + message
+            cocotb.log.error(msg)
+            scheduler.finish_scheduler(SimFailure(msg))
+
+        else:
+            # Sim failure/end after the regression manager finishes. Do nothing.
+            cocotb.log.info(message)
+
+    else:  # pragma: no cover
+        # should never happen
+        msg = "Unsupported sim event: level={}, msg={}".format(level, message)
         scheduler.log.error(msg)
-        scheduler.finish_scheduler(SimFailure(msg))
-    else:
-        scheduler.log.error("Unsupported sim event")
 
     return True
 
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -251,21 +251,21 @@ class RegressionManager:
         self._log_test_summary()
         self._log_sim_summary()
 
-        # Generate output reports
+        # Write user coverage
         self.xunit.write()
         if self._cov:
             self._cov.stop()
             self.log.info("Writing coverage data")
             self._cov.save()
             self._cov.html_report()
-        if cocotb._library_coverage is not None:
-            # TODO: move this once we have normal shutdown behavior to _sim_event
-            cocotb._library_coverage.stop()
-            cocotb._library_coverage.save()
 
         # Setup simulator finalization
         self._finished = True
         simulator.stop_simulator("Shutting down...")
+
+        if cocotb._library_coverage is not None:
+            cocotb._library_coverage.stop()
+            cocotb._library_coverage.save()  # pragma: no cover
 
     @property
     def finished(self) -> bool:
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -235,8 +235,7 @@ class RegressionManager:
         else:
             return cocotb.scheduler.add(test)
 
-    def tear_down(self) -> None:
-        # fail remaining tests
+    def _fail_remaining_tests(self) -> None:
         while True:
             test = self.next_test()
             if test is None:
@@ -247,6 +246,7 @@ class RegressionManager:
                 wall_time_s=0,
                 sim_time_ns=0)
 
+    def tear_down(self) -> None:
         # Write out final log messages
         self._log_test_summary()
         self._log_sim_summary()
@@ -300,11 +300,14 @@ class RegressionManager:
         # stop capturing log output
         cocotb.log.removeHandler(test.handler)
 
-        self._record_result(
+        sim_failed = self._record_result(
             test=self._test,
             outcome=self._test_task._outcome,
             wall_time_s=real_time,
             sim_time_ns=sim_time_ns)
+
+        if sim_failed:
+            self._fail_remaining_tests()
 
         self.execute()
 
@@ -335,7 +338,9 @@ class RegressionManager:
         if isinstance(test_init_outcome, cocotb.outcomes.Error):
             self.log.error("Failed to initialize test %s" % test.__qualname__,
                            exc_info=test_init_outcome.error)
-            self._record_result(test, test_init_outcome, 0, 0)
+            sim_failed = self._record_result(test, test_init_outcome, 0, 0)
+            if sim_failed:
+                self._fail_remaining_tests()
             return
 
         test = test_init_outcome.get()
@@ -415,7 +420,7 @@ class RegressionManager:
         outcome: Optional[Outcome],
         wall_time_s: float,
         sim_time_ns: float
-    ) -> None:
+    ) -> bool:
 
         ratio_time = self._safe_divide(sim_time_ns, wall_time_s)
 
@@ -443,9 +448,7 @@ class RegressionManager:
             'real': wall_time_s,
             'ratio': ratio_time})
 
-        if sim_failed:
-            self.tear_down()
-            return
+        return sim_failed
 
     def execute(self) -> None:
         while True:
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -95,7 +95,6 @@ class RegressionManager:
         self.count = 0
         self.skipped = 0
         self.failures = 0
-        self._tearing_down = False
 
         # Setup XUnit
         ###################
@@ -236,12 +235,6 @@ class RegressionManager:
             return cocotb.scheduler.add(test)
 
     def tear_down(self) -> None:
-        # prevent re-entering the tear down procedure
-        if not self._tearing_down:
-            self._tearing_down = True
-        else:
-            return
-
         # fail remaining tests
         while True:
             test = self.next_test()
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -95,6 +95,7 @@ class RegressionManager:
         self.count = 0
         self.skipped = 0
         self.failures = 0
+        self._finished = False
 
         # Setup XUnit
         ###################
@@ -249,7 +250,6 @@ class RegressionManager:
         # Write out final log messages
         self._log_test_summary()
         self._log_sim_summary()
-        self.log.info("Shutting down...")
 
         # Generate output reports
         self.xunit.write()
@@ -264,7 +264,18 @@ class RegressionManager:
             cocotb._library_coverage.save()
 
         # Setup simulator finalization
-        simulator.stop_simulator()
+        self._finished = True
+        simulator.stop_simulator("Shutting down...")
+
+    @property
+    def finished(self) -> bool:
+        """
+        Returns ``True`` once the regression manager has finished.
+
+        Finished implies all tests have been run or skipped and test results
+        and coverage information written out to their respective files.
+        """
+        return self._finished
 
     def next_test(self) -> Optional[Test]:
         """Get the next test to run"""
--- a/cocotb/share/lib/fli/FliImpl.cpp
+++ b/cocotb/share/lib/fli/FliImpl.cpp
@@ -42,13 +42,10 @@ static FliImpl         *fli_table;
 
 void FliImpl::sim_end()
 {
-    if (GPI_DELETE != sim_finish_cb->get_call_state()) {
-        sim_finish_cb->set_call_state(GPI_DELETE);
-        if (mti_NowUpper() == 0 && mti_Now() == 0 && mti_Delta() == 0) {
-            mti_Quit();
-        } else {
-            mti_Break();
-        }
+    if (mti_NowUpper() == 0 && mti_Now() == 0 && mti_Delta() == 0) {
+        mti_Quit();
+    } else {
+        mti_Break();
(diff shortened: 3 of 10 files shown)


cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -547,18 +547,9 @@ int VpiImpl::deregister_callback(GpiCbHdl *gpi_hdl)
...
-// If the Python world wants things to shut down then unregister
-// the callback for end of sim
 void VpiImpl::sim_end()
 {
-    /* Some sims do not seem to be able to deregister the end of sim callback
-     * so we need to make sure we have tracked this and not call the handler
-     */
-    if (GPI_DELETE != sim_finish_cb->get_call_state()) {
-        sim_finish_cb->set_call_state(GPI_DELETE);
-        vpi_control(vpiFinish);
```
eric-wieser, line 559: Was this not needed at all?

ktbarrett: The first bit is deregistering the shutdown callback. Seemingly to prevent it from running because it would try to run through shutdown behavior again once the simulator actually shuts down. That should certainly be tossed, that's part of the normalization.

`vpiFinish` is important, I just though that through this morning, I will add that back in.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -219,7 +217,6 @@ int handle_gpi_callback(void *user_data)
             PyErr_Print();
 
             gpi_sim_end();
```
garmin-mjames, line 219: This now causes `FATAL: We are calling up again` error because we aren't in simulator context when `gpi_embed_event` is called.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -803,7 +793,6 @@ static PyObject *stop_simulator(PyObject *self, PyObject *args)
     COCOTB_UNUSED(self);
     COCOTB_UNUSED(args);
     gpi_sim_end();
```
garmin-mjames, line 795: This now also causes the `FATAL: We are calling up again` error.
We aren't using it after this PR, should we remove it?

garmin-mjames:
Is the shutdown callback working and tested for all of our supported simulators?

cocotb/share/lib/simulator/simulatormodule.cpp
ktbarrett: I had debated whether I should add the line there or not, but we signal to the users we are shutting down in the shutdown callback, so they can handle shutdown then, we don't need to tell them early.

ktbarrett:
We should fix the multiple prints in another PR. I'll look into that tonight. `tear_down` shouldn't be called in `_score_test`, only in the top-level execution loop of the regression manager. But considering what we have in the RM/Scheduler right now, I guess we can prevent that by checking against this PRs `finished` attribute in `tear_down` to prevent it happening several times.

I haven't checked all simulators yet, just Icarus in CI.

cocotb/share/lib/simulator/simulatormodule.cpp
cmarqu: Side question: is this ``FATAL: We are calling up again`` printing some more context now since https://github.com/cocotb/cocotb/pull/1743/files is in, or did I close #1736 prematurely?

garmin-mjames:
It's working in Active-HDL, although with a caveat.
The shutdown callback looks like it is called from a different thread, so `cleanup` raises an Exception...

cocotb/share/lib/simulator/simulatormodule.cpp
garmin-mjames: @cmarqu You get the location where the call happened:
```
     0.00ns ERROR    cocotb.gpi                                gpi_embed.cpp:432  in embed_sim_event                 FATAL: We are calling up again
```

garmin-mjames: I think it could be worded better, though.

garmin-mjames: @ktbarrett The comment here in `handle_gpi_callback` suggests that we don't want to return to Python as there was an Exception and we don't know the state of things. With `embed_sim_event` now being called from `gpi_sim_end`, we ARE returning into Python to notify the Scheduler/RM. We get a separate error if the event call into Python fails, so maybe that's enough in the case where Python is really screwed.

ktbarrett:
That is surprising and obnoxious. Do we really need the exception there? Does `cleanup` need to be in the thread that started the regression manager?

garmin-mjames:
@ktbarrett Sorry, looks like that's only for `SimFailure` runs, but still annoying.

ktbarrett:
@garmin-mjames 
>looks like that's only for SimFailure runs

Do you mean it only fails when the sim fails due to returning without pending triggers? That means it should fail on master as well?

garmin-mjames:
@ktbarrett I mean that the `Exception` in `cleanup` is only raised when there is a `SIM_FAIL` event before the `RegressionManager` is finalized. so normal succesful tests don't cause the `Exception` because `_sim_event` doesn't call `finish_scheduler`.

garmin-mjames:
The Exception in `cleanup`:
https://github.com/cocotb/cocotb/blob/cf22bfdc4a8caa0c71273890a30394bcf71a3f75/cocotb/scheduler.py#L845-L846
Looking at the commit history and #19, it seems to be centered around the `@external` stuff.

garmin-mjames:
I confirmed Active-HDL calls the shutdown callback from a different thread in the same process. It results in a Dummy thread object from `threading.current_thread()`.

@eric-wieser Do you have insight into this error checking?

imphil:
OK, the Aldec (Riviera/ActiveHDL) issue is now tracked in #1859 and was already around before this PR. We don't need to block this PR on fixing it first.

cocotb/share/lib/simulator/simulatormodule.cpp
ktbarrett: The call to `gpi_embed_event`. no longer occurs since I've refactored sim_end.

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -127,26 +127,22 @@ int gpi_register_impl(GpiImplInterface *func_tbl)
...
 void gpi_embed_end()
 {
-    embed_sim_event(SIM_FAIL, "Simulator shutdown prematurely");
-    gpi_cleanup();
+    gpi_embed_event(SIM_FAIL, "Simulator shutdown prematurely");
+    CLEAR_STORE();
+    embed_sim_cleanup();
 }
 
 void gpi_sim_end()
```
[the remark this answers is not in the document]
ktbarrett: I'll do that as a follow on. https://github.com/ktbarrett/cocotb/tree/sim-end-message.

cocotb/share/lib/simulator/simulatormodule.cpp
ktbarrett: This function is still useful. Users should be able to request shut down whenever, we can't always assume the simulator can determine when cocotb is done. For example, if you have an HDL clock, there will always be pending triggers and never shut down.

garmin-mjames:
With #1863 in, the `gpi_embed_end` calls added after `LOG_CRITICAL` need to be changed in this PR.

ktbarrett:
@garmin-mjames In what way?

garmin-mjames:
@ktbarrett Shouldn't they be changed to `gpi_sim_end` to send an event back up to Python and allow possible cleanup, and use the shutdown callback for `gpi_embed_end`?

cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -42,13 +42,10 @@ static FliImpl         *fli_table;
 
 void FliImpl::sim_end()
 {
-    if (GPI_DELETE != sim_finish_cb->get_call_state()) {
```
eric-wieser, line 45: Whose job is it to deregister the callback now?

eric-wieser: > This is necessary if there are multiple GPI users. If we deregister the shutdown callback, not every GPI user would be notified of a shutdown

Can you elaborate on this?

ktbarrett: Previously, when `sim_end` was called, it deregistered the shutdown callback. I assume this was done to prevent shutdown behavior from running *twice* because it currently does shutdown behavior immediately (after the stack unravels). This isn't a problem if we put all shutdown behavior in the shutdown callback and don't deregister it.

The reason this is necessary for multiple GPI users is in the case of one user calling for shutdown, it must notify all the other users of shutdown using `gpi_sim_event`. If we call this when `sim_end` is called, we get re-entrance. Placing this in the shutdown callback means it's always called from a simulator context.

ktbarrett:
@garmin-mjames Could I get a final review on this?

garmin-mjames (commented):
#1803 is back in `issue_1279` test:
https://github.com/cocotb/cocotb/pull/1793/checks?check_run_id=1278725108#step:22:159

ktbarrett:
Fixed #1803 again with a better solution (IMO).

cocotb/regression.py
```diff
@@ -256,22 +250,32 @@ def tear_down(self) -> None:
...
         self.xunit.write()
         if self._cov:
             self._cov.stop()
             self.log.info("Writing coverage data")
             self._cov.save()
             self._cov.html_report()
+
+        # Setup simulator finalization
+        self._finished = True
+        simulator.stop_simulator("Shutting down...")
```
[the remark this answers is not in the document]
ktbarrett: Fixed.

cocotb/__init__.py
```diff
@@ -269,18 +269,39 @@ def _sim_event(level, message):
...
+
+        elif not regression_manager.finished:
+            # Test failure during a test, fail it
+            msg = "Test failure from external source: " + message
+            cocotb.log.error(msg)
+            scheduler.finish_test(TestFailure(msg))
+
+        else:  # pragma: no cover
+            # Test failure after the regression manager finishes. Should never happen.
+            cocotb.log.info(message)
```
ktbarrett, line 284: Maybe make this an error since it shouldn't ever happen.

cocotb/__init__.py
```diff
@@ -269,18 +269,40 @@ def _sim_event(level, message):
...
+
     elif level is SIM_FAIL:
-        # We simply return here as the simulator will exit
-        # so no cleanup is needed
-        msg = ("Failing test at simulator request before test run completion: "
-               "%s" % message)
+        if regression_manager is None:  # pragma: no cover
+            # Sim failure/end before regression manager started. Should never happen.
+            cocotb.log.error("Regression ended before Regression Manager started: " + message)
+
```
garmin-mjames, line 291: This will happen any time `embed_sim_init` fails or `_initialise_testbench` raises an Exception before `regression_manager` is initialized. An easy way to test this is to set `TOPLEVEL` to a bad value.

cocotb/regression.py
```diff
@@ -439,9 +448,7 @@ def _record_result(
             'real': wall_time_s,
             'ratio': ratio_time})
 
-        if sim_failed:
-            self.tear_down()
-            return
+        return sim_failed
```
garmin-mjames, line 451: Update return type hint.

garmin-mjames (changes requested):

cocotb/regression.py
```diff
@@ -253,25 +246,36 @@ def tear_down(self) -> None:
...
         if cocotb._library_coverage is not None:
-            # TODO: move this once we have normal shutdown behavior to _sim_event
             cocotb._library_coverage.stop()
-            cocotb._library_coverage.save()
+            cocotb._library_coverage.save()  # pragma: no cover
 
-        # Setup simulator finalization
-        simulator.stop_simulator()
+    @property
+    def finished(self):
```
ktbarrett, line 271:
```suggestion
    def finished(self) -> bool:
```

cocotb/regression.py
```diff
@@ -235,14 +235,7 @@ def _init_hook(self, hook: Hook) -> Optional[RunningTask]:
...
 
-    def tear_down(self) -> None:
-        # prevent re-entering the tear down procedure
-        if not self._tearing_down:
-            self._tearing_down = True
-        else:
-            return
-
-        # fail remaining tests
+    def _fail_remaining_tests(self):
```
ktbarrett, line 238:
```suggestion
    def _fail_remaining_tests(self) -> None:
```

ktbarrett (commented):
`_record_result` should have the type hint updated to return `None`.

cocotb/__init__.py
ktbarrett: Icarus will fail to find the top level and error before cocotb is loaded.

ktbarrett: Added a test for this in the most recent commit.

garmin-mjames: `TOPLEVEL` discovery is in `_initialise_testbench`:

```
     -.--ns INFO     cocotb.gpi                         ..mbed/gpi_embed.cpp:97   in set_program_name_in_venv        Using Python virtual environment interpreter at .../env/bin/python
     -.--ns INFO     cocotb.gpi                         ../gpi/GpiCommon.cpp:106  in gpi_print_registered_impl       VPI registered
     -.--ns INFO     cocotb.gpi                         ..mbed/gpi_embed.cpp:244  in embed_sim_init                  Python interpreter initialized and cocotb loaded!
     0.00ns INFO     cocotb                                      __init__.py:212  in _initialise_testbench           Running on Icarus Verilog version 12.0 (devel)
     0.00ns INFO     cocotb                                      __init__.py:218  in _initialise_testbench           Running tests with cocotb v1.5.0.dev0 from .../env/lib/python3.8/site-packages/cocotb
     0.00ns INFO     cocotb                                      __init__.py:239  in _initialise_testbench           Seeding Python random module with 1603472827
     0.00ns ERROR    cocotb.gpi                                  VpiImpl.cpp:495  in get_root_handle                 VPI: Couldn't find root handle garbage
     0.00ns ERROR    cocotb.gpi                                  VpiImpl.cpp:501  in get_root_handle                 VPI: Toplevel instances: garbage != $unit...
     0.00ns ERROR    cocotb.gpi                                  VpiImpl.cpp:501  in get_root_handle                 VPI: Toplevel instances: garbage != dff...
     0.00ns ERROR    cocotb.gpi                                GpiCommon.cpp:288  in gpi_get_root_handle             No root handle found
  File ".../env/lib/python3.8/site-packages/cocotb/__init__.py", line 250, in _initialise_testbench
    raise RuntimeError("Can not find root handle ({})".format(root_name))
RuntimeError: Can not find root handle (garbage)
     0.00ns ERROR    cocotb.gpi                                gpi_embed.cpp:314  in embed_sim_init                  cocotb initialization failed - exiting
     0.00ns ERROR    cocotb                                      __init__.py:290  in _sim_event                      Regression ended before Regression Manager started: Failed to initialize GPI user.
```

ktbarrett: Clean that directory and run it again and you'll see what I mean.

garmin-mjames: Remove the pragma and update the comment?

garmin-mjames:
New test needs a make override like `test_exit_error`.

tests/test_cases/test_module_var_bad/Makefile
```diff
@@ -0,0 +1,13 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+# should cause a failure in cocotb._initialize_testbench which will trigger a path in
+# cocotb._sim_event that isn't otherwise seen
+MODULE := bad
+
+PHONY: override_for_this_test
```
garmin-mjames, line 9:
```suggestion
.PHONY: override_for_this_test
```

garmin-mjames (dismissed):
Also tested with Active-HDL 11.1 ✔️

ktbarrett:
Testing with Questa 2020.1 I'm getting the following after the test finishes. I'll run it down later
```
# ** Note: $finish    : UNKNOWN(-1)
#      0.00ns ERROR    cocotb.gpi                                gpi_embed.cpp:338  in embed_sim_event                 FATAL: We are calling up again
```

ktbarrett:
@garmin-mjames What was the "calling up again" test there to protect against originally? Is re-entering Python actually dangerous? Or is this a distant side-effect of preventing scheduler re-entry?

garmin-mjames:
The context checks were added in https://github.com/cocotb/cocotb/commit/acbc6e131f67ff85cc9427a82b7ec0054c6564b5.
On the same issue as the `Exception("Cleanup() called outside of the main thread") `.
I haven't used the `@external`/`@function` stuff at all, but maybe it was for protecting against multiple threads entering python from the simulator?

garmin-mjames:
That is pretty bad... I wonder if it we could queue up the callback data in `handle_vpi_callback` and unwind the `GpiCbHdl::run_callback` first before handling it. I don't know if the simulator will try to kill things after we return from the shutdown callback before we can do that though.

garmin-mjames:
Does this put the kibosh on this method?

> NOTE—In response to a call to `vhpi_control` with the argument `vhpiFinish`, the tool does not perform any
`vhpiCbEndOfSimulation` callbacks.

Aldec simulators still call the callbacks.

ktbarrett:
@garmin-mjames I think so... I guess there was a reason it's set up so weird. It might be worth a refactor, but I don't think this strategy will work.

PR #1795, opened 2020-05-12
Don't include unistd.h on windows

tpambor:
Only `getpid()` is used on windows from `unistd.h`. As it is not always present use `GetCurrentProcessId()` and include it only for non-windows platforms.

### DIFF
--- a/cocotb/share/lib/embed/gpi_embed.cpp
+++ b/cocotb/share/lib/embed/gpi_embed.cpp
@@ -30,7 +30,6 @@
 // Embed Python into the simulator using GPI
 
 #include <Python.h>
-#include <unistd.h>
 #include <cocotb_utils.h>
 #include "embed.h"
 #include "locale.h"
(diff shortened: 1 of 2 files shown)


eric-wieser:
To be clear, the issue is with msvc, not windows, right?

tpambor:
I think it is the other way around. `unistd.h` is only present on POSIX systems, `mingw` ships a `unistd.h` for compatibility reasons. I am not 100% sure but I guess other compilers available on Windows like Intel C++ also don't have `unistd.h`.

ktbarrett:
@tpambor the ifdef should be "if Windows and not POSIX" so mingw compilations continue to use `getpid` from `unistd.h`.

tpambor:
> @tpambor the ifdef should be "if Windows and not POSIX" so mingw compilations continue to use `getpid` from `unistd.h`.

`getpid()` just calls `GetCurrentProcessId()` [https://github.com/mirror/mingw-w64/blob/master/mingw-w64-crt/misc/_getpid.c#L11](https://github.com/mirror/mingw-w64/blob/master/mingw-w64-crt/misc/_getpid.c#L11)

IMHO I don't see any benefit in complicating the ifdef logic here, we always have `windows.h` on windows and already use it for `Sleep`

ktbarrett:
Alright, I was worried they were doing something that we weren't handling.

PR #1796, opened 2020-05-12
Add visibility attributes for functions/classes/variables to be exported

tpambor:
`*_EXPORT` macro is used to either mark a function, class or variable as exported with `__declspec(dllexport)` (on Windows) or `__attribute__ ((visibility ("default")))` for gcc and clang, if the matching `*_EXPORTS` macro is set or as imported with `__declspec(dllimport)`.

The naming convention for the macros used here is based on the library name `LIBNAME_EXPORT` and `LIBNAME_EXPORTS` which is defined during compilation of `LIBNAME`.

MSVC requires symbols to be exported to be marked with `__declspec(dllexport)` by specifying `-Wl,--exclude-all-symbols` we align the behavior between GCC and MSVC. This largely narrows the gap for future MSVC support. For non-windows platform the default visibility is changed to hidden by specifying `-fvisibility=hidden -fvisibility-inlines-hidden`. Therefore visible symbols in the shared libraries are the same across all (supported) platform compiler combinations.

Advantages are "very substantially improves load times", "lets the optimiser produce better code", "reduces the size of your DSO by 5-20%", "Much lower chance of symbol collision". For more details see [https://gcc.gnu.org/wiki/Visibility](https://gcc.gnu.org/wiki/Visibility).

On Linux code size is reduced by ~40% and performance increased by a few percent depending on the tests run.

Avoiding symbol collisions is important if a wrapper for runtime resolving is added. This helps slim down #1783 which only included exporting functions for `libcocotbvpi_*` and `libcocotbvhpi_*`.

### DIFF
--- a/cocotb/share/include/embed.h
+++ b/cocotb/share/include/embed.h
@@ -30,16 +30,23 @@
 #ifndef COCOTB_EMBED_H_
 #define COCOTB_EMBED_H_
 
+#include <exports.h>
+#ifdef COCOTB_EMBED_EXPORTS
+#define COCOTB_EMBED_EXPORT COCOTB_EXPORT
+#else
+#define COCOTB_EMBED_EXPORT COCOTB_IMPORT
+#endif
+
 #include <gpi.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-extern void embed_init_python(void);
-extern void embed_sim_cleanup(void);
-extern int embed_sim_init(int argc, char const* const* argv);
-extern void embed_sim_event(gpi_event_t level, const char *msg);
+extern COCOTB_EMBED_EXPORT void embed_init_python(void);
+extern COCOTB_EMBED_EXPORT void embed_sim_cleanup(void);
+extern COCOTB_EMBED_EXPORT int embed_sim_init(int argc, char const* const* argv);
+extern COCOTB_EMBED_EXPORT void embed_sim_event(gpi_event_t level, const char *msg);
 
 #ifdef __cplusplus
 }
--- a/cocotb/share/include/exports.h
+++ b/cocotb/share/include/exports.h
@@ -1,0 +1,19 @@
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef COCOTB_EXPORTS_H_
+#define COCOTB_EXPORTS_H_
+
+// Make cocotb work correctly when the default visibility is changed to hidden
+// Changing the default visibility to hidden has the advantage of significantly reducing the code size,
+// load times as well as letting the optimizer produce better code
+#if defined(__linux__) || defined(__APPLE__)
+#define COCOTB_EXPORT __attribute__ ((visibility ("default")))
+#define COCOTB_IMPORT
+#else
+#define COCOTB_EXPORT __declspec(dllexport)
+#define COCOTB_IMPORT __declspec(dllimport)
+#endif
+
+#endif
--- a/cocotb/share/lib/vhpi/VhpiImpl.h
+++ b/cocotb/share/lib/vhpi/VhpiImpl.h
@@ -28,6 +28,13 @@
 #ifndef COCOTB_VHPI_IMPL_H_
 #define COCOTB_VHPI_IMPL_H_
 
+#include <exports.h>
+#ifdef COCOTBVHPI_EXPORTS
+#define COCOTBVHPI_EXPORT COCOTB_EXPORT
+#else
+#define COCOTBVHPI_EXPORT COCOTB_IMPORT
+#endif
+
 #include "../gpi/gpi_priv.h"
 #include <vhpi_user.h>
 #include <vhpi_user_ext.h>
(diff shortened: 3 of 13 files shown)


themperek:
What is the reason to use `*_EXPORTS` defines if we always set them? (Sorry if this is a naive question I am not very familiar with meters.)

tpambor:
> What is the reason to use `*_EXPORTS` defines if we always set them? (Sorry if this is a naive question I am not very familiar with meters.)

They are not always defined. They are only defined when the respective library is compiled. e.g. `COCOTBUTILS_EXPORTS` is defined when `libcocotbutils` is compiled. As a consequence `utils_dyn_open` is then marked `__declspec(dllexport)` but when `libgpi` is compiled `COCOTBUTILS_EXPORTS` is undefined and therefore `COCOTBUTILS_EXPORT` is `__declspec(dllimport)` and when `cocotb_utils.h` is imported during compilation of `libgpi` `utils_dyn_open` then is marked `__declspec(dllimport)`.

cocotb/share/include/embed.h
```diff
@@ -30,6 +30,20 @@
...
+#define COCOTB_EXPORT __declspec(dllexport)
+#endif
+#else
+#if defined(__linux__) || defined(__APPLE__)
+#define COCOTB_EXPORT
+#else
+#define COCOTB_EXPORT __declspec(dllimport)
+#endif
+#endif
+
```
eric-wieser, line 46: Can we define this in a single header and use it across all our libraries?

ktbarrett: If I am remember correctly, these macros once existed in cocotb_util.h, but were removed because they weren't used.

tpambor: They distinct from each other. They are needed for every library that is build. There are only as many as we have libraries. For an explanation of their function, see my comment [above](https://github.com/cocotb/cocotb/pull/1796#issuecomment-627261359).

tpambor:
This is now also enabled for non-windows platforms by specifying the gcc visibility attributes and changing default visibility to hidden.

On Linux code size is reduced by ~40% and performance by a few percent depending on the tests run.

eric-wieser:
Can we split this in two?
```
#if defined(__linux__) || defined(__APPLE__)
#define COCOTB_EXPORT __attribute__ ((visibility ("default")))
#define COCOTB_IMPORT
#else
#define COCOTB_EXPORT __declspec(dllexport)
#define COCOTB_IMPORT __declspec(dllimport)
#endif
```
in some shared file, and then define `COCOTBUTILS_API` as one of those two macros based on COCOTBUTILS_EXPORTS?

tpambor:
> Can we split this in two?
> 
> ```
> #if defined(__linux__) || defined(__APPLE__)
> #define COCOTB_EXPORT __attribute__ ((visibility ("default")))
> #define COCOTB_IMPORT
> #else
> #define COCOTB_EXPORT __declspec(dllexport)
> #define COCOTB_IMPORT __declspec(dllimport)
> #endif
> ```
> 
> in some shared file, and then define `COCOTBUTILS_API` as one of those two macros based on COCOTBUTILS_EXPORTS?

Good idea, done!

cocotb/share/lib/vhpi/VhpiImpl.h
```diff
@@ -28,6 +28,16 @@
 #ifndef COCOTB_VHPI_IMPL_H_
 #define COCOTB_VHPI_IMPL_H_
 
+#ifdef COCOTBVHPI_EXPORTS
+#if defined(__linux__) || defined(__APPLE__)
+#define COCOTBVHPI_EXPORT __attribute__ ((visibility ("default")))
+#else
+#define COCOTBVHPI_EXPORT __declspec(dllexport)
+#endif
```
eric-wieser, line 36: Missed this one?

tpambor: Slipped through, thanks for the catch

cocotb_build_libs.py
```diff
@@ -167,7 +172,9 @@ def _get_python_lib():
 # TODO [gh-1372]: make this work for MSVC which has a different flag syntax
 _base_warns = ["-Wall", "-Wextra", "-Wcast-qual", "-Wwrite-strings", "-Wconversion"]
 _ccx_warns = _base_warns + ["-Wnon-virtual-dtor", "-Woverloaded-virtual"]
-_extra_cxx_compile_args = ["-std=c++11"] + _ccx_warns
+_extra_cxx_compile_args = ["-std=c++11", "-fvisibility=hidden", "-fvisibility-inlines-hidden"] + _ccx_warns
+if os.name != "nt":
+    _extra_cxx_compile_args += ["-flto"]
```
eric-wieser, line 177: Why is this both here and on line 140?

cocotb/share/include/exports.h
```diff
@@ -0,0 +1,16 @@
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef COCOTB_EXPORTS_H_
+#define COCOTB_EXPORTS_H_
+
+#if defined(__linux__) || defined(__APPLE__)
```
eric-wieser, line 8: It would be good to include a comment above this line saying:

* Why we care about visibility / dll settings.
* How the macros in this file are intended to be used.

cocotb_build_libs.py
tpambor: On line 140 are linker arguments and on line 177 are compiler arguments `-flto` needs to be passed to linker and compiler.

tpambor:
Rebased on current master

PR #1797, opened 2020-05-12
Remove unused includes in embed.h

tpambor:
These includes are not used and can be cleaned up. VPI interfaces are abstracted in `libcocotbvpi_*` and should not be included here.

### DIFF
--- a/cocotb/share/include/embed.h
+++ b/cocotb/share/include/embed.h
@@ -31,7 +31,6 @@
 #define COCOTB_EMBED_H_
 
 #include <gpi.h>
-#include <gpi_logging.h>
 
 #ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


eric-wieser:
Not quite true @ktbarrett, we need the definition of `gpi_event_t`

cocotb/share/include/embed.h
```diff
@@ -32,8 +32,6 @@
 
 #include <gpi.h>
 #include <gpi_logging.h>
```
eric-wieser, line 34:
```suggestion
```
This one can probably go too

eric-wieser:
Travis status pushes seem stuck, but the link says the build passed.

PR #1798, opened 2020-05-12
Add support for MSVC

### LINKED ISSUE #1372: Add support for MSVC
#1306 adds support for using setuptools, which knows how to invoke MSVC. However, initial experimentation revealed some follow-up work:

* Compiler flags need fixing (they are currently `-Werror` instead of `/W4` for instance)
* `.lib` files seem missing (https://travis-ci.org/themperek/cocotb/builds/646876123)
  * Can't use `dlltool`, need `lib /def:modelsim.def /out:modelsim.lib /machine:x86` etc

Opening this issue to track future work.

tpambor:
This PR adds support for compiling with MSVC.

Depends on:
- [x] #1795 Don't include unistd.h on windows
- [x] #1797 Remove unused includes in embed.h
- [x] #1796 Add __declspec(dllexport) for functions/classes/variables to be exported

TODO:
- [x] Add CI
- [x] Clarify what warning level is desired on MSVC (at the moment the default `/W3` is used), W4 per recommendation on https://docs.microsoft.com/de-de/cpp/build/reference/compiler-option-warning-level
- [x] #2334 Portable compiler/linker flags
- [x] Make Makefiles work without lib prefix

Closes #1372, #2334, #2341

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -238,6 +238,9 @@ jobs:
       run: python -m pip install  --global-option build_ext --global-option --compiler=mingw32 -e .
     - name: Install Python testing dependencies (Windows)
       run: pip install coverage xunitparser pytest pytest-cov
+    - name: SxS trace test (Windows)
+      if: startsWith(matrix.os, 'windows')
+      run: .\tests\sxs.ps1
     - name: Test (Windows)
       if: startsWith(matrix.os, 'windows')
       id: windowstesting
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -137,6 +137,16 @@ jobs:
             lang: verilog
             python-version: 3.8
             os: windows-latest
+            toolchain: mingw
+            extra_name: "mingw | "
+
+          - sim: icarus             # use msvc instead of mingw
+            sim-version: master
+            lang: verilog
+            python-version: 3.8
+            os: windows-latest
+            toolchain: msvc
+            extra_name: "msvc | "
 
             # Other
 
@@ -230,14 +240,22 @@ jobs:
         sudo make install
 
       # Windows Testing
-    - name: Install cocotb build dependencies (Windows)
-      if: startsWith(matrix.os, 'windows')
-      run: conda install --yes -c msys2 m2-base m2-make m2w64-toolchain libpython
-    - name: Install cocotb (Windows)
-      if: startsWith(matrix.os, 'windows')
-      run: python -m pip install  --global-option build_ext --global-option --compiler=mingw32 -e .
     - name: Install Python testing dependencies (Windows)
       run: pip install coverage xunitparser pytest pytest-cov
+    - name: Install cocotb build dependencies (Windows, mingw)
+      if: startsWith(matrix.os, 'windows') && matrix.toolchain == 'mingw'
+      run: conda install --yes -c msys2 m2-base m2-make m2w64-toolchain libpython
+    - name: Install cocotb (Windows, mingw)
+      if: startsWith(matrix.os, 'windows') && matrix.toolchain == 'mingw'
+      run: python -m pip install  --global-option build_ext --global-option --compiler=mingw32 -v -e .
+      continue-on-error: ${{matrix.may_fail || false}}
+    - name: Install cocotb (Windows, msvc)
+      if: startsWith(matrix.os, 'windows') && matrix.toolchain == 'msvc'
+      run: python -m pip install -v -e .
+      continue-on-error: ${{matrix.may_fail || false}}
+    - name: Install cocotb runtime dependencies (Windows, msvc)
+      if: startsWith(matrix.os, 'windows') && matrix.toolchain == 'msvc'
+      run: conda install --yes -c msys2 m2-base m2-make
     - name: SxS trace test (Windows)
       if: startsWith(matrix.os, 'windows')
       run: .\tests\sxs.ps1
--- a/cocotb/share/lib/gpi/GpiCommon.cpp
+++ b/cocotb/share/lib/gpi/GpiCommon.cpp
@@ -159,6 +159,11 @@ void gpi_embed_event(gpi_event_t level, const char *msg)
 static void gpi_load_libs(std::vector<std::string> to_load)
 {
 #define DOT_LIB_EXT "." xstr(LIB_EXT)
+#ifdef LIB_PREFIX
+#define LIB_PREFIX_STR xstr(LIB_PREFIX)
+#else
+#define LIB_PREFIX_STR ""
+#endif
     std::vector<std::string>::iterator iter;
 
     for (iter = to_load.begin();
@@ -182,7 +187,7 @@ static void gpi_load_libs(std::vector<std::string> to_load)
             func_name = std::string(it+1, arg.end());
         }
 
-        std::string full_name = "lib" + lib_name + DOT_LIB_EXT;
+        std::string full_name = LIB_PREFIX_STR + lib_name + DOT_LIB_EXT;
         void *lib_handle = utils_dyn_open(full_name.c_str());
         if (!lib_handle) {
             printf("cocotb: Error loading shared library %s\n", full_name.c_str());
--- a/cocotb/share/lib/gpi/GpiCommon.cpp
+++ b/cocotb/share/lib/gpi/GpiCommon.cpp
@@ -158,12 +158,6 @@ void gpi_embed_event(gpi_event_t level, const char *msg)
 
 static void gpi_load_libs(std::vector<std::string> to_load)
 {
-#define DOT_LIB_EXT "." xstr(LIB_EXT)
-#ifdef LIB_PREFIX
-#define LIB_PREFIX_STR xstr(LIB_PREFIX)
-#else
-#define LIB_PREFIX_STR ""
-#endif
     std::vector<std::string>::iterator iter;
 
     for (iter = to_load.begin();
@@ -173,24 +167,19 @@ static void gpi_load_libs(std::vector<std::string> to_load)
 
         std::string arg = *iter;
 
-        std::string lib_name;
-        std::string func_name;
-        auto it = std::find(arg.begin(), arg.end(), ':');
-        bool const has_colon = it == arg.end();
-        if (has_colon) {
-            // no colon in the string, default
-            lib_name = arg;
-            func_name = arg + "_entry_point";
-        }
-        else {
-            lib_name = std::string(arg.begin(), it);
-            func_name = std::string(it+1, arg.end());
+        auto const idx = arg.rfind(':');  // find from right since path could contain colons (Windows)
+        if (idx == std::string::npos) {
+            // no colon in the string
+            printf("cocotb: Error parsing GPI_EXTRA %s\n", arg.c_str());
+            exit(1);
         }
 
-        std::string full_name = LIB_PREFIX_STR + lib_name + DOT_LIB_EXT;
-        void *lib_handle = utils_dyn_open(full_name.c_str());
+        std::string const lib_name = arg.substr(0, idx);
+        std::string const func_name = arg.substr(idx+1, std::string::npos);
+
+        void *lib_handle = utils_dyn_open(lib_name.c_str());
         if (!lib_handle) {
-            printf("cocotb: Error loading shared library %s\n", full_name.c_str());
+            printf("cocotb: Error loading shared library %s\n", lib_name.c_str());
             exit(1);
         }
 
@@ -198,7 +187,7 @@ static void gpi_load_libs(std::vector<std::string> to_load)
         if (!entry_point) {
             char const* fmt = "cocotb: Unable to find entry point %s for shared library %s\n%s";
             char const* msg = "        Perhaps you meant to use `,` instead of `:` to separate library names, as this changed in cocotb 1.4?\n";
-            printf(fmt, func_name.c_str(), full_name.c_str(), has_colon ? msg : "");
+            printf(fmt, func_name.c_str(), lib_name.c_str(), msg);
             exit(1);
         }
 
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -64,9 +64,13 @@ PYTHON_BIN ?= $(realpath $(shell cocotb-config --python-bin))
 include $(COCOTB_SHARE_DIR)/makefiles/Makefile.deprecations
 
 LIB_DIR=$(COCOTB_PY_DIR)/cocotb/libs
+LIB_PREFIX := lib
 
 ifeq ($(OS),Msys)
     LIB_EXT := dll
+    ifneq (,$(wildcard $(LIB_DIR)/cocotb.dll))
+        LIB_PREFIX :=
+    endif
 else
     LIB_EXT := so
 endif
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -151,10 +151,6 @@ def _get_lib_ext_name():
 
 class build_ext(_build_ext):
     def run(self):
-
-        def_dir = os.path.join(cocotb_share_dir, "def")
-        self._gen_import_libs(def_dir)
-
         if os.name == "nt":
             create_sxs_appconfig(self.get_ext_fullpath(os.path.join("cocotb", "simulator")))
 
@@ -168,6 +164,16 @@ class build_ext(_build_ext):
                 create_rc_file(name, filename, libraries)
 
         super().run()
+
+    def build_extensions(self):
+        if os.name == "nt":
+            def_dir = os.path.join(cocotb_share_dir, "def")
+            self._gen_import_libs(def_dir)
+
+            for e in self.extensions:
+                e.library_dirs += [def_dir]
+
+        super().build_extensions()
 
     def build_extension(self, ext):
         """Build each extension in its own temp directory to make gcov happy.
@@ -247,17 +253,16 @@ class build_ext(_build_ext):
         load the DLL (.a) based on module definition files (.def)
         """
 
-        if os.name == "nt":
-            for sim in ["icarus", "modelsim", "aldec", "ghdl"]:
-                subprocess.run(
-                    [
-                        "dlltool",
-                        "-d",
-                        os.path.join(def_dir, sim + ".def"),
-                        "-l",
-                        os.path.join(def_dir, "lib" + sim + ".a"),
-                    ]
-                )
+        for sim in ["icarus", "modelsim", "aldec", "ghdl"]:
+            subprocess.run(
+                [
+                    "dlltool",
+                    "-d",
+                    os.path.join(def_dir, sim + ".def"),
+                    "-l",
+                    os.path.join(def_dir, "lib" + sim + ".a"),
+                ]
+            )
 
 
 def _extra_link_args(lib_name=None, rpaths=[]):
@@ -535,7 +540,6 @@ def get_ext():
 
     share_lib_dir = os.path.relpath(os.path.join(cocotb_share_dir, "lib"))
     include_dir = os.path.relpath(os.path.join(cocotb_share_dir, "include"))
-    share_def_dir = os.path.relpath(os.path.join(cocotb_share_dir, "def"))
 
     ext = []
 
@@ -547,18 +551,15 @@ def get_ext():
     #  Icarus Verilog
     #
     icarus_extra_lib = []
-    icarus_extra_lib_path = []
     logger.info("Compiling libraries for Icarus Verilog")
     if os.name == "nt":
         icarus_extra_lib = ["icarus"]
-        icarus_extra_lib_path = [share_def_dir]
 
     icarus_vpi_ext = _get_vpi_lib_ext(
         include_dir=include_dir,
         share_lib_dir=share_lib_dir,
         sim_define="ICARUS",
         extra_lib=icarus_extra_lib,
-        extra_lib_dir=icarus_extra_lib_path,
     )
     ext.append(icarus_vpi_ext)
 
@@ -566,18 +567,15 @@ def get_ext():
     #  Modelsim/Questa
     #
     modelsim_extra_lib = []
-    modelsim_extra_lib_path = []
     logger.info("Compiling libraries for Modelsim/Questa")
     if os.name == "nt":
         modelsim_extra_lib = ["modelsim"]
-        modelsim_extra_lib_path = [share_def_dir]
 
     modelsim_vpi_ext = _get_vpi_lib_ext(
         include_dir=include_dir,
         share_lib_dir=share_lib_dir,
         sim_define="MODELSIM",
         extra_lib=modelsim_extra_lib,
-        extra_lib_dir=modelsim_extra_lib_path,
     )
     ext.append(modelsim_vpi_ext)
 
@@ -604,7 +602,6 @@ def get_ext():
                 define_macros=[("COCOTBFLI_EXPORTS", "")] + _extra_defines,
                 include_dirs=[include_dir, modelsim_include_dir],
                 libraries=["gpi", "gpilog"] + modelsim_extra_lib,
-                library_dirs=modelsim_extra_lib_path,
                 sources=fli_sources,
                 extra_link_args=_extra_link_args(lib_name=lib_name, rpaths=["$ORIGIN"]),
                 extra_compile_args=_extra_cxx_compile_args,
@@ -624,18 +621,15 @@ def get_ext():
     # GHDL
     #
     ghdl_extra_lib = []
-    ghdl_extra_lib_path = []
     logger.info("Compiling libraries for GHDL")
     if os.name == "nt":
         ghdl_extra_lib = ["ghdl"]
-        ghdl_extra_lib_path = [share_def_dir]
 
     ghdl_vpi_ext = _get_vpi_lib_ext(
         include_dir=include_dir,
         share_lib_dir=share_lib_dir,
         sim_define="GHDL",
         extra_lib=ghdl_extra_lib,
-        extra_lib_dir=ghdl_extra_lib_path,
     )
     ext.append(ghdl_vpi_ext)
 
@@ -668,18 +662,15 @@ def get_ext():
     # Aldec Riviera Pro
     #
     aldec_extra_lib = []
-    aldec_extra_lib_path = []
     logger.info("Compiling libraries for Riviera")
     if os.name == "nt":
         aldec_extra_lib = ["aldec"]
-        aldec_extra_lib_path = [share_def_dir]
 
     aldec_vpi_ext = _get_vpi_lib_ext(
         include_dir=include_dir,
         share_lib_dir=share_lib_dir,
         sim_define="ALDEC",
         extra_lib=aldec_extra_lib,
-        extra_lib_dir=aldec_extra_lib_path,
     )
     ext.append(aldec_vpi_ext)
 
@@ -688,7 +679,6 @@ def get_ext():
         share_lib_dir=share_lib_dir,
         sim_define="ALDEC",
         extra_lib=aldec_extra_lib,
-        extra_lib_dir=aldec_extra_lib_path,
     )
     ext.append(aldec_vhpi_ext)
 
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -20,6 +20,15 @@ from typing import List
 logger = logging.getLogger(__name__)
 cocotb_share_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "cocotb", "share"))
 
+_base_warns = ["-Wall", "-Wextra", "-Wcast-qual", "-Wwrite-strings", "-Wconversion"]
+_ccx_warns = _base_warns + ["-Wnon-virtual-dtor", "-Woverloaded-virtual"]
+_extra_cxx_compile_args = ["-std=c++11", "-fvisibility=hidden", "-fvisibility-inlines-hidden"] + _ccx_warns
+if os.name != "nt":
+    _extra_cxx_compile_args += ["-flto"]
+
+# Make PRI* format macros available with C++11 compiler but older libc, e.g. on RHEL6.
+_extra_defines = [("__STDC_FORMAT_MACROS", "")]
+
 
 def create_sxs_assembly_manifest(name: str, filename: str, libraries: List[str], dependency_only=False) -> str:
     """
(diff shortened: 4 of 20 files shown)


themperek:
If you change `py3_mingw` to `py3` CI should pick MSVC compiler.
https://github.com/cocotb/cocotb/blob/c7da01ce2d09aec7f76da9f1b802a3bd679942fc/.travis.yml#L88

Not sure if the right versions are available: https://docs.travis-ci.com/user/reference/windows/#pre-installed-chocolatey-packages

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -122,14 +122,6 @@ namespace {
...
         return type;
     }
-
-    // these will be initialized later, once the members are all defined
-    template<>
-    PyTypeObject gpi_hdl_Object<gpi_sim_hdl>::py_type;
-    template<>
-    PyTypeObject gpi_hdl_Object<gpi_iterator_hdl>::py_type;
-    template<>
-    PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type;
```
eric-wieser, line 132: I remember needing to forward-declare these in order to reference them below. Perhaps the references were removed in a later commit.

Does msvc reject these?

cocotb/share/lib/vpi/VpiTrampoline.cpp
```diff
@@ -0,0 +1,228 @@
...
+        static PLI_INT32  vpi_mcd_flush       (PLI_UINT32 mcd)                                                                     { static fptr_vpi_mcd_flush             f = reinterpret_cast<fptr_vpi_mcd_flush>            (resolve_function("vpi_mcd_flush"            )); return f(mcd);                                }
+        static PLI_INT32  vpi_control         (PLI_INT32 operation, ...)                                                           { va_list ap; va_start(ap, operation); PLI_INT32 ret = vpi_vcontrol(operation, ap); va_end(ap); return ret; }
+        static vpiHandle  vpi_handle_by_multi_index (vpiHandle obj, PLI_INT32 num_index, PLI_INT32 *index_array)                   { static fptr_vpi_handle_by_multi_index f = reinterpret_cast<fptr_vpi_handle_by_multi_index>(resolve_function("vpi_handle_by_multi_index")); return f(obj, num_index, index_array);        }
+
+        static PLI_INT32 vpi_vcontrol         (PLI_INT32 operation, va_list ap)
```
eric-wieser, line 121: You can avoid this special case with
```suggestion
        template<typename...Args>
        static PLI_INT32 vpi_vcontrol         (PLI_INT32 operation, Args... args)
```
Then use `f(operation, args...)` later.

cocotb/share/lib/simulator/simulatormodule.cpp
tpambor:
```
  cocotb\share\lib\simulator\simulatormodule.cpp(964): error C2086: 'PyTypeObject `anonymous namespace'::gpi_hdl_Object<GpiObjHdl *>::py_type': redefinition
  cocotb\share\lib\simulator\simulatormodule.cpp(128): note: see declaration of 'py_type'
```
Without forward-decleration MSVC is happy but clang complains:
```
  cocotb/share/lib/simulator/simulatormodule.cpp:956:43: error: explicit specialization of 'py_type' after instantiation
  PyTypeObject gpi_hdl_Object<gpi_sim_hdl>::py_type = []() -> PyTypeObject {
                                            ^
  cocotb/share/lib/simulator/simulatormodule.cpp:485:59: note: implicit instantiation first required here
      if (Py_TYPE(pSigHdl) != &gpi_hdl_Object<gpi_sim_hdl>::py_type) {
```
GCC is fine either way

eric-wieser: Any idea which compiler is following the standard here?

tpambor: I think the relevant part from spec is §14.7.3 [temp.expl.spec]/p6 (emphasis added):
> If a template, a member template or a member of a class template is explicitly specialized then **that specialization shall be declared before the first use of that specialization that would cause an implicit instantiation to take place, in every translation unit in which such a use occurs**; no diagnostic is required.

I updated the code to move the template before its first use and now it works fine on all compilers, see [`078dc8b`](https://github.com/cocotb/cocotb/pull/1798/commits/078dc8b3cccae2fce9e8feaa1d862443a500c6fe).

eric-wieser: I asked [on StackOverflow here](https://stackoverflow.com/q/61770519/102441). To me this looks like an MSVC bug, the existing code already contains the declaration.

am11: @eric-wieser, I read your comment on community [bug report](https://developercommunity.visualstudio.com/content/problem/89837/vc2017-a-declaration-of-variable-template-is-alway.html) (which was mysteriously closed by some sort of a canned response..). However, [similar report](https://developercommunity.visualstudio.com/content/problem/319447/explicit-specialization-of-static-data-member-inco.html) was made where it was explained that it has been fixed in VS2019 16.5; but since it is a breaking change, it will require `/permissive-` to compile in standard conformance mode.

library-free repro:

```c++
// specialization-declaration.cpp
template<typename X> class A { };

template<typename Y> class B
{
public:
    static A<Y> j;
};

template<> A<int> B<int>::j;
A<int> y = B<int>::j;
template<> A<int> B<int>::j = { };
```
with:
```cmd
cl /c .\specialization-declaration.cpp
```
gives `C2086` errors, while in std conformance mode:

```cmd
cl /c /permissive- .\specialization-declaration.cpp
```
compiles fine:
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.26.28806 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

specialization-declaration.cpp
```
(Visual Studio 2019, 16.6.3)

tpambor: @am11 Thank you, adding `/permissive-` fixed the compile error.

ktbarrett:
@tpambor Is #1783 *necessary* to enable MSVC support? It's a rather big change that some maintainers have issue with, and we will discuss it soon. It would be nice to just bring in the last 2 commits so we have MSVC support now, if at all possible.

I'd like these large changes in early in the release window so we have a few months to detect issues.

tpambor:
@ktbarrett This could probably go in without #1783.

This PR is not yet ready to be merged. I have split out in #2019, #2020 the parts that are ready. These are cleanups not directly related with adding support for MSVC. Once they are merged I will give this PR a respin without #1783.

ktbarrett:
@tpambor Is there any news on the updates? We are hoping to release 1.5 soon.

tpambor:
At the moment this is blocked on #2334.

ktbarrett:
Does your last commit in this PR not solve the problem?

tpambor:
I did a clean respin without #1783. It already compiles fine and should not regress the other targets. Tests are currently still failing, which is probably due to problems with the manifest.

ktbarrett (commented):
Looks pretty solid.

cocotb_build_libs.py
```diff
@@ -20,6 +20,17 @@
 logger = logging.getLogger(__name__)
 cocotb_share_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "cocotb", "share"))
 
+_base_warns = ["-Wall", "-Wextra", "-Wcast-qual", "-Wwrite-strings", "-Wconversion"]
+_ccx_warns = _base_warns + ["-Wnon-virtual-dtor", "-Woverloaded-virtual"]
+_extra_cxx_compile_args = ["-std=c++11", "-fvisibility=hidden", "-fvisibility-inlines-hidden"] + _ccx_warns
+if os.name != "nt":
+    _extra_cxx_compile_args += ["-flto"]
+
+_extra_cxx_compile_args_msvc = ["/W4", "/permissive-"]
```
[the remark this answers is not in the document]
tpambor: C++11 should be the default.

> Note that there is no plan to add a C++11 switch.
https://devblogs.microsoft.com/cppblog/standards-version-switches-in-the-compiler/

cocotb_build_libs.py
```diff
@@ -343,7 +384,7 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
     ]
     if os.name == "nt":
         libcocotbutils_sources += ["libcocotbutils.rc"]
-    libcocotbutils_libraries = ["gpilog"]
+    libcocotbutils_libraries = _prefix_libraries(["gpilog"])
```
eric-wieser, line 387: Why are we prefixing the libraries here? [this answer](https://stackoverflow.com/a/9796089/102441) seems to suggest the the library names should be unprefixed.

tpambor: If you pass `-lcocotb` to mingw it looks for `libcocotb.dll` while msvc looks for `cocotb.dll`. As we probably still want the library to be called `libcocotb.dll` we prefix it with lib if we compile with msvc.

cocotb_build_libs.py
```diff
@@ -335,15 +354,11 @@ def _get_python_lib():
...
-_ccx_warns = _base_warns + ["-Wnon-virtual-dtor", "-Woverloaded-virtual"]
-_extra_cxx_compile_args = ["-std=c++11", "-fvisibility=hidden", "-fvisibility-inlines-hidden"] + _ccx_warns
-if os.name != "nt":
-    _extra_cxx_compile_args += ["-flto"]
-
-# Make PRI* format macros available with C++11 compiler but older libc, e.g. on RHEL6.
-_extra_defines = [("__STDC_FORMAT_MACROS", "")]
+def _prefix_libraries(libs):
+    if os.name == "nt":
+        return ["###WINLIBPREFIX###" + l for l in libs]
```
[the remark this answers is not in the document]
tpambor: See https://github.com/cocotb/cocotb/pull/1798#discussion_r555233126_. I will add comment.

cocotb_build_libs.py
eric-wieser: I don't really agree with that. Under msvc, "libpython" is really called `python.dll`. I think we should try to stick with the platform defaults unless there is a good reason not to.

Presumably the issue is that the makefiles assume this prefix?

tpambor:
> Where does the `lib` prefix get added?

It comes from the extension name `libcocotb` etc.

eric-wieser:
> > Where does the `lib` prefix get added?
> 
> It comes from the extension name `libcocotb` etc.

I feel like we should not be putting `lib` in the extension name - so instead of removing the `lib` prefix on MSVC, we should be adding it on non-MSVC; the reason we need it is because we are abusing `Extension` to build shared libraries.

themperek:
I see quite a persistence here 👍🏻 .
Good luck.

themperek:
I think there was discussion software (cannot find it now). 
Do we really want to have a different name (no lib) just for msvc? 
Is this required?

cocotb_build_libs.py
themperek: In this case name of vpi/fli/vhpi library will depend on the compiler one uses? So depending on compiler mings/msvc different name has to be used?

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -182,7 +187,7 @@ static void gpi_load_libs(std::vector<std::string> to_load)
             func_name = std::string(it+1, arg.end());
         }
 
-        std::string full_name = "lib" + lib_name + DOT_LIB_EXT;
+        std::string full_name = LIB_PREFIX_STR + lib_name + DOT_LIB_EXT;
```
ktbarrett, line 190: I was never a fan of this. I'm not sure why the user shouldn't just pass the full name. This prevents the ability to pass relative or absolute paths. I'll split that out into another issue.

cocotb_build_libs.py
themperek: Is not better to have a bit more "ugly" compilation and keep those names the same?

ktbarrett:
@tpambor I realized that may change requires a lot of the makefile changes you have already done. So [here](https://github.com/ktbarrett/cocotb/tree/gpi-extra-libs) are my suggested changes to the `gpi_load_libs` code along with newsfragment and documentation change. Might save you a little work.

cocotb/share/makefiles/Makefile.inc
```diff
@@ -64,9 +64,13 @@ PYTHON_BIN ?= $(realpath $(shell cocotb-config --python-bin))
 include $(COCOTB_SHARE_DIR)/makefiles/Makefile.deprecations
 
 LIB_DIR=$(COCOTB_PY_DIR)/cocotb/libs
+LIB_PREFIX := lib
 
 ifeq ($(OS),Msys)
     LIB_EXT := dll
+    ifneq (,$(wildcard $(LIB_DIR)/cocotb.dll))
```
themperek, line 71: I do not have a very strong opinion here but is a good idea to have the name of vpi/fli/vhpi different only for one simulator.

tpambor:
All tests are passing now 🚀

cocotb/share/makefiles/Makefile.inc
tpambor: That's not for a simulator here, mingw/msys uses the unix style `lib` prefix while microsoft convention on windows is without. At the moment this is mostly a change internal to cocotb but if `libgpi` or other components might be distributed separately, it may be a slight advantage to follow the native way.

themperek: Will it complicate potential extensions?

themperek: >That's not for a simulator here, mingw/msys uses the unix style lib prefix while microsoft convention on windows is without.

That is the point so now it is just for windows and msvc different. 

Is this really an advantage?

tpambor:
> @tpambor I realized that my changes required a lot of the makefile changes you have already done. So [here](https://github.com/ktbarrett/cocotb/tree/gpi-extra-libs) are my suggested changes to the `gpi_load_libs` code along with newsfragment and documentation change. Might save you a little work.

I have merged your changes.

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -168,32 +167,27 @@ static void gpi_load_libs(std::vector<std::string> to_load)
...
-        if (has_colon) {
-            // no colon in the string, default
-            lib_name = arg;
-            func_name = arg + "_entry_point";
-        }
-        else {
-            lib_name = std::string(arg.begin(), it);
-            func_name = std::string(it+1, arg.end());
+        auto const it = std::rfind(arg.begin(), arg.end(), ':');  // find from right since path could contain colons (Windows)
+        if (it == arg.end()) {
```
ktbarrett, line 171:
```suggestion
        auto const idx = arg.rfind(':');  // find from right since path could contain colons (Windows)
        if (idx == std::string::npos) {
```

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -168,32 +167,27 @@ static void gpi_load_libs(std::vector<std::string> to_load)
...
+            // no colon in the string
+            printf("cocotb: Error parsing GPI_EXTRA %s\n", arg.c_str());
+            exit(1);
         }
 
-        std::string full_name = "lib" + lib_name + DOT_LIB_EXT;
-        void *lib_handle = utils_dyn_open(full_name.c_str());
+        std::string const lib_name {arg.begin(), it};
+        std::string const func_name {it+1, arg.end()};
```
ktbarrett, line 178:
```suggestion
        std::string const lib_name = arg.substr(0, idx);
        std::string const func_name = arg.substr(idx+1, std::string::npos);
```

ktbarrett (commented):
Sorry, switched to `rfind` at the last minute without checking it. These changes work on your branch locally.

cocotb_build_libs.py
```diff
@@ -150,13 +161,32 @@ def _get_lib_ext_name():
...
+    def _uses_msvc(self):
+        if self.compiler == "msvc":
+            return True
+        if self.compiler is None:
+            return get_default_compiler() == "msvc"
+        else:
+            return "msvc" == getattr(self.compiler, "compiler_type", None)
+
     def run(self):
+        create_sxs_appconfig(self.get_ext_fullpath(os.path.join("cocotb", "simulator")))
```
ktbarrett, line 173: I noticed it was creating a file on Linux.
```suggestion
        if os.name == "nt":
            create_sxs_appconfig(self.get_ext_fullpath(os.path.join("cocotb", "simulator")))
```

cocotb/share/makefiles/Makefile.inc
ktbarrett: Depending on the system that we use to load extensions, it might not be a problem. Since we are supporting full library names and paths in `GPI_EXTRA`, the file names in the list could be absolute paths to files on the system. These names would probably be populated with globing, so there is no issue of conditional computation of the library name.

I think our current use of `GPI_EXTRA` for loading "other" libraries for mixed language support will go the way of the dinosaur; and most of the library name computing added in this PR will be removed.

themperek: Maybe can add this to `cocotb-config` so custom flows is a bit easier to build?
Something like `cocotb-config --vpi-lib icarus`?

cocotb/share/lib/gpi/GpiCommon.cpp
tpambor: Private CI still seems unhappy. I guess `GPI_EXTRA` needs to be changed in the Makefiles as well?

themperek: Log [az.log](https://github.com/cocotb/cocotb/files/5814367/az.log)

ktbarrett: Yeah my change requires full names. So .so/.dll extensions are necessary. The idea is whatever the user puts in GPI_EXTRA is fed directly to utils_dyn_open.

cocotb/share/makefiles/Makefile.inc
ktbarrett: I think so. `cocotb-config` should be a one stop shop for custom flows. Maybe we can do that as a follow on.

ktbarrett (commented):
Can you also add a newsfragment for 1798 stating there is no MSVC support? So close...

.github/workflows/regression-tests.yml
```diff
@@ -230,14 +240,25 @@ jobs:
...
+    - name: Install cocotb (Windows, msvc)
+      if: startsWith(matrix.os, 'windows') && matrix.toolchain == 'msvc'
+      run: python -m pip install -e .
+      continue-on-error: ${{matrix.may_fail || false}}
+    - name: Install cocotb runtime dependencies (Windows, msvc)
+      if: startsWith(matrix.os, 'windows') && matrix.toolchain == 'msvc'
+      run: conda install --yes -c msys2 m2-base m2-make
+    - name: SxS trace test (Windows)
+      if: startsWith(matrix.os, 'windows')
+      run: .\tests\sxs.ps1
```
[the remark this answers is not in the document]
tpambor: At least it helps track down problems with library loading on windows more quickly. Mostly it is helpful for PRs like #2315 which touch the dependencies between libraries. The normal error message doesn't provide a lot of details and with the additional output here you could get a hint what is wrong even if you don't have a windows setup.

ktbarrett:
One last thing is to update the [installation docs](https://cocotb--1798.org.readthedocs.build/en/1798/install.html#installation) and the [development version installation docs](https://cocotb--1798.org.readthedocs.build/en/1798/install_devel.html#installing-the-development-version). For the typical end user who doesn't care which compiler is used, they don't need to specify the compiler and it should automatically pick the right one? 

Then I might link to the installation instructions in the newsfragment:
```rst
See :ref:`install` for more details. 
```

ktbarrett (approved):

ktbarrett:
Going to get one more approval since this is a large change, I will give it one week.

ktbarrett:
I will also test it locally and with Questa, just to make sure nothing got messed up on the systems we don't test in CI.

ktbarrett:
Status of testing:

* It builds and runs on RHEL7 using Questa.
* It builds and runs on Windows using GCC in an Anaconda environment.
* It does not build in an MSYS2 environment using GCC (#2378), but I think we can skip that issue for now.
* It builds on Windows using MSVC in an Anaconda environment ___but does not run___.

I think your MSVC test is a red herring. Run `pip install cocotb` with the `-v` flag and see what compiler is being used. GCC is being installed previously to compile Icarus, GCC remains in the anaconda environment (see the log output when installing the runtime dependencies note they are already installed), GCC is picked up as the compiler and everything works.

tpambor:
> * It does not build in an MSYS2 environment using GCC (#2378), but I think we can skip that issue for now.

Works for me locally with `pip install .`, note I am not using `pip -e` might be related to that.

> * It builds on Windows using MSVC in an Anaconda environment _**but does not run**_.

Works for me locally with
```
conda install -c msys2 m2-base m2-make
pip install .
make test
```
on a clean install of miniconda.

> I think your MSVC test is a red herring. Run `pip install cocotb` with the `-v` flag and see what compiler is being used. GCC is being installed previously to compile Icarus, GCC remains in the anaconda environment (see the log output when installing the runtime dependencies note they are already installed), GCC is picked up as the compiler and everything works.

Should be working, I will add `-v` again, this is also useful because I think LGTM only inspects the warnings for gcc on linux and therefore we also would have an impression for msvc and mingw gcc.

ktbarrett:
Odd... it is using MSVC without any arguments despite GCC being installed.

I have an existing MSVC install. I did a fresh miniconda install. If I `pip install cocotb` it uses MSVC and builds successfully. As soon as I `conda install -c msys2 m2-base m2-make` and run `pip install .` it errors leaving me a traceback through `distutils/.../cygwincompiler.py`. So it seems to be trying to use GCC though I haven't installed it.

On your MSYS2 installation, when compiling, is it creating libraries with the `.so` extension or `.dll`? I get `.so`s which confuses `ld` when trying to link. I made this environment fresh, since I didn't have MSYS2 installed before yesterday.

themperek:
>On your MSYS2 installation, when compiling, is it creating libraries with the .so extension or .dll? I get .sos which confuses ld when trying to link. I made this environment fresh since I didn't have MSYS2 installed before yesterday.

Straight MSYS2 never worked I believe. Need to use mingw64/32 https://github.com/cocotb/cocotb/wiki/Tier-2-Setup-Instructions#windows-installation-using-msys2.

tpambor:
> Odd... it is using MSVC without any arguments despite GCC being installed.
> 
> I have an existing MSVC install. I did a fresh miniconda install. If I `pip install .` it uses MSVC and builds successfully. As soon as I `conda install -c msys2 m2-base m2-make` and run `pip install .` it errors leaving me a traceback through `distutils/.../cygwincompiler.py`. So it seems to be trying to use GCC though I haven't installed it.

For me it sticks to MSVC. I think normally the default is the compiler that was also used to compile python because I guess extension compiled with gcc would not work for python compiled with MSVC and viceversa.

```
(base) C:\>make
make: *** No targets specified and no makefile found.  Stop.      
(base) C:\>python
Python 3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32
```


> On your MSYS2 installation, when compiling, is it creating libraries with the `.so` extension or `.dll`? I get `.so`s which confuses `ld` when trying to link. I made this environment fresh, since I didn't have MSYS2 installed before yesterday.

```
    Directory: C:\msys64\mingw64\lib\python3.8\site-packages\cocotb

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       22.01.2021     13:20                drivers
d-----       22.01.2021     13:20                generators
d-----       22.01.2021     13:20                libs
d-----       22.01.2021     13:20                monitors
d-----       22.01.2021     13:20                share
d-----       22.01.2021     13:20                vendor
d-----       22.01.2021     13:20                __pycache__
-a----       22.01.2021     13:20           3665 ANSI.py
-a----       22.01.2021     13:20          25333 binary.py
-a----       22.01.2021     13:20           8942 bus.py
-a----       22.01.2021     13:20           6468 clock.py
-a----       22.01.2021     13:20           6529 config.py
-a----       22.01.2021     13:20          21455 decorators.py
-a----       22.01.2021     13:20          36769 handle.py
-a----       22.01.2021     13:20           2926 ipython_support.py
-a----       22.01.2021     13:20          10365 log.py
-a----       22.01.2021     13:20           1913 memdebug.py
-a----       22.01.2021     13:20           1377 outcomes.py
-a----       22.01.2021     13:20          30200 regression.py
-a----       22.01.2021     13:20           5476 result.py
-a----       22.01.2021     13:20          35604 scheduler.py
-a----       22.01.2021     13:20          11336 scoreboard.py
-a----       22.01.2021     13:20          58880 simulator-cpython-38.dll
-a----       22.01.2021     13:20            263 simulator-cpython-38.dll.2.config
-a----       22.01.2021     13:20          30300 triggers.py
-a----       22.01.2021     13:20          21515 utils.py
-a----       22.01.2021     13:20           7036 wavedrom.py
-a----       22.01.2021     13:20           3630 xunit_reporter.py
-a----       22.01.2021     13:20           2752 _py_compat.py
-a----       22.01.2021     13:20           2928 _sim_versions.py
-a----       22.01.2021     13:20            308 _version.py
-a----       22.01.2021     13:20          11045 __init__.py

    Directory: C:\msys64\mingw64\lib\python3.8\site-packages\cocotb\libs

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----       22.01.2021     13:20          19968 libcocotb.dll
-a----       22.01.2021     13:20          14848 libcocotbutils.dll
-a----       22.01.2021     13:20         101888 libcocotbvhpi_aldec.dll
-a----       22.01.2021     13:20          92160 libcocotbvpi_aldec.dll
-a----       22.01.2021     13:20          92160 libcocotbvpi_ghdl.dll
-a----       22.01.2021     13:20          92160 libcocotbvpi_icarus.vpl
-a----       22.01.2021     13:20          92160 libcocotbvpi_modelsim.dll
-a----       22.01.2021     13:20          15360 libembed.dll
-a----       22.01.2021     13:20          65536 libgpi.dll
-a----       22.01.2021     13:20          40448 libgpilog.dll
-a----       22.01.2021     13:20          40960 libpygpilog.dll
```

ktbarrett:
@themperek Those instructions did not work for me. They are also out of date: you can no longer install from master.zip; and it mentions `pip3` and `python3`, but MSYS2 and MinGW32/64 have renamed `python` and `pip` for Python 3.

<details><summary>Output</summary>

```
kaleb@DESKTOP-2CHBMVI MINGW64 ~/cocotb
$ pip install .
Processing c:/msys64/home/kaleb/cocotb
    ERROR: Command errored out with exit status 1:
     command: C:/msys64/mingw64/bin/python.exe -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:/Users/kaleb/AppData/Local/Temp/pip-req-build-0uja_t1j/setup.py'"'"'; __file__='"'"'C:/Users/kaleb/AppData/Local/Temp/pip-req-build-0uja_t1j/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base C:/Users/kaleb/AppData/Local/Temp/pip-pip-egg-info-3y2f0v_u
         cwd: C:/Users/kaleb/AppData/Local/Temp/pip-req-build-0uja_t1j/
    Complete output (34 lines):
    Modelsim/Questa executable (vdbg) executable not found. No FLI interface will be available.
      File "C:/Users/kaleb/AppData/Local/Temp/pip-req-build-0uja_t1j/setup.py", line 84, in <module>
        setup(
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/__init__.py", line 153, in setup
        return distutils.core.setup(**attrs)
      File "C:/msys64/mingw64/lib/python3.8/distutils/core.py", line 108, in setup
        _setup_distribution = dist = klass(attrs)
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/dist.py", line 430, in __init__
        _Distribution.__init__(self, {
      File "C:/msys64/mingw64/lib/python3.8/distutils/dist.py", line 292, in __init__
        self.finalize_options()
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/dist.py", line 706, in finalize_options
        ep(self)
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/dist.py", line 713, in _finalize_setup_keywords
        ep.load()(self, ep.name, value)
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools_scm/integration.py", line 26, in version_keyword
        dist.metadata.version = _get_version(config)
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools_scm/__init__.py", line 173, in _get_version
        parsed_version = _do_parse(config)
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools_scm/__init__.py", line 127, in _do_parse
        version = _version_from_entrypoints(config) or _version_from_entrypoints(
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools_scm/__init__.py", line 65, in _version_from_entrypoints
        version = _call_entrypoint_fn(root, config, ep.load())
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools_scm/__init__.py", line 46, in _call_entrypoint_fn
        return fn(root, config=config)
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools_scm/git.py", line 97, in parse
        wd = GitWorkdir.from_potential_worktree(config.absolute_root)
      File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools_scm/git.py", line 33, in from_potential_worktree
        if not samefile(real_wd, wd):
      File "C:/msys64/mingw64/lib/python3.8/genericpath.py", line 100, in samefile
        s1 = os.stat(f1)
    FileNotFoundError: [WinError 3] The system cannot find the path specified: '/c/Users/kaleb/AppData/Local/Temp/pip-req-build-0uja_t1j'
    ----------------------------------------
WARNING: Discarding file:///C://msys64/home/kaleb/cocotb. Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

</details>

@tpambor Apparently if I have the libpython conda package installed it tries to use GCC, but will use MSVC otherwise. Of course, if I don't have libpython installed it doesn't run correctly. I doesn't run correct even if I do install it after the fact. So the question is why does CI and your machine work when mine doesn't? Do you have an MSVC compiled libpythonXX.dll somewhere else on your system that it is picking up?

```
/c/iverilog/bin/iverilog -o sim_build/sim.vvp -D COCOTB_SIM=1 -s sample_module -f sim_build/cmds.f -g2012   /c/Users/kaleb/cocotb/tests/test_cases/issue_120/../../../tests/designs/sample_module/sample_module.sv
C:/iverilog/bin/iverilog.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory
```

marlonjames:
Tested, builds and runs:

Active-HDL 32-bit - Conda Python 3.8.5 (default, Sep  4 2020, 00:03:40) [MSC v.1916 32 bit (Intel)] :: Anaconda, Inc. on win32 - cocotb regression tests ✅ 
Active-HDL 32-bit - System Python 3.8.7 (tags/v3.8.7:6503f05, Dec 21 2020, 17:43:54) [MSC v.1928 32 bit (Intel)] on win32 - custom Tcl flow ✅

ktbarrett:
@themperek Seems my issue is https://github.com/pypa/setuptools_scm/issues/415. Though that is not an issue this PR can address.

@tpambor I guess it was some issue with my Icarus. Even old versions work fine now. I think this PR is good to go.

documentation/source/newsfragments/1798.feature.rst
```diff
@@ -0,0 +1,3 @@
+Support for building with Microsoft Visual C++ has been added.
+The official CPython distribution for Windows is now supported in addition to the mingw version.
```
ktbarrett, line 2: The issue with setuptools_scm means the mingw environments are not currently supported. That might be fixed upstream, or perhaps installing from sdists where setuptools_scm has already run would work? We don't currently have any instructions for how to use the official distribution for Python on Windows. I think we should remove this line.

ktbarrett:
That is the last change I see necessary before merging. If you want to clean up the commits, now is the time. Alternatively, I can just squash them.

ktbarrett (approved):

ktbarrett:
@tpambor Thanks for all the work, this is great! It was a long process, but I'm glad to finally see this come in. 

Waiting until Wednesday morning to see if there are any last minute comments, then I will rebase.

ktbarrett:
Making sure the documentation build isn't an issue (it looks like it just blew up on their end).

ktbarrett:
@tpambor Thanks again for the help! I don't think many of us here are very Windows savvy, so your contributions are greatly appreciated.

PR #1800, opened 2020-05-12
#1799: Fix avalon monitor hexdump debug log

### LINKED ISSUE #1799: AvalonSTPkts hexdump needs a minor fix now that the transaction is of type bytes rather than string
When DEBUG logging level is enabled, the `hexdump` call in `_monitor_recv` function of AvalonSTPkts calls hexdump after converting the `pkt` in to `str` which is no longer needed

ajeshp:
Closes #1799


cmarqu:
Thanks for your first pull request in cocotb!

PR #1051, opened 2019-07-25
Analog Mixed Signal Example

cmarqu:
See https://cocotb--1051.org.readthedocs.build/en/1051/examples.html#mixed-signal-analog-digital for documentation.

PR #1804, opened 2020-05-12
Ignore png files in CI whitespace check

cmarqu:
I have a PR (#1051) that includes PNG files, and our whitespace check in CI trips over them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ jobs:
     - stage: Lint
       name: whitespace
       os: linux
-      script: "[[ -z \"$(git ls-files | xargs egrep -l '\\s+$')\" ]]"
+      script: "! git --no-pager grep --color --line-number --full-name --extended-regexp -I '\\s+$'"
 
     - stage: Lint
       name: flake8


eric-wieser:
It would be neat to ask git which files to ignore, as in this question: https://stackoverflow.com/q/29694540/102441

.travis.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - stage: Lint
       name: whitespace
       os: linux
-      script: "[[ -z \"$(git ls-files | xargs egrep -l '\\s+$')\" ]]"
+      script: "[[ -z \"$(git ls-files | grep -v '.png$' | xargs egrep -l '\\s+$')\" ]]"
```
eric-wieser, line 20:
```suggestion
      script: "git --no-pager grep --color -line-number --full-name -I '\\s+$'"
```
@ktbarrett, thoughts?

cmarqu: When trying this locally, I need to escape the ``+`` too - what works for me is ``'\s\+$'``.

eric-wieser: Let's use `-E` instead

cmarqu: Done.

PR #1805, opened 2020-05-12
Debug: Improve RunningTask.__repr__

garmin-mjames:
Add `RunningTask.__repr__` that shows current status and unique task id.

This is the base for a follow-on PR to overhaul scheduler debug.

Each `RunningTask` has it's own "coroutine stack", and currently that's not easily accessible, so it can be difficult during debugging to figure out the state and call location of a task. Generator coroutines can `yield from` other generators and coroutines can `await` other coroutines, both of which will bypass the scheduler.

Depending on the status of the task, the pending trigger object or outcome will also be displayed.

Example `repr()` outputs from `test_task_repr`:

- `<Task 2 created coro=generator_coro_outer()>`
- `<Task 2 pending coro=generator_coro_inner() trigger=<Timer of 1000.00ps at 0x0F0B2A58>>`
- `<Task 2 finished coro=generator_coro_outer() outcome=Error(ValueError('inner'))>`
- `<Task 3 adding coro=coroutine_outer()>`
- `<Task 3 running coro=coroutine_outer()>`
- ~~`<Task 3 pending coro=coroutine_inner() trigger=<<Event for Combine(<Task 4 pending coro=coroutine_wait() trigger=<Timer of 1000.00ps at 0x0F418210>>, <Task 5 pending coro=coroutine_wait() trigger=<Timer of 1000.00ps at 0x0F0B2AA8>>) at 0x0F0A038B>.wait() at 0x0F0B29B8>>`~~
- `<Task 3 pending coro=coroutine_inner() trigger=Combine(Join(<Task 4>), Join(<Task 5>)>`

Tests are displayed `<Test test_name ...>` rather than with the task id.

The current scheduler debug output is less useful after this PR, as `__name__` and `__qualname__` are repurposed, so I changed it to get them from the coroutine wrapped by `RunningTask`.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -107,6 +107,7 @@ class RunningTask:
         self._started = False
         self._callbacks = []
         self._outcome = None
+        self._trigger = None
 
     @lazy_property
     def log(self):
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -35,7 +35,7 @@ import os
 import cocotb
 from cocotb.log import SimLog
 from cocotb.result import ReturnValue
-from cocotb.utils import get_sim_time, lazy_property, remove_traceback_frames
+from cocotb.utils import get_sim_time, lazy_property, remove_traceback_frames, extract_coro_stack
 from cocotb import outcomes
 
 # Sadly the Python standard logging module is very slow so it's better not to
@@ -87,6 +87,8 @@ class RunningTask:
         triggers to fire.
     """
 
+    _id_count = 0  # used by the scheduler for debug
+
     def __init__(self, inst):
 
         if inspect.iscoroutine(inst):
@@ -102,12 +104,15 @@ class RunningTask:
             raise TypeError(
                 "%s isn't a valid coroutine! Did you forget to use the yield keyword?" % inst)
         self._coro = inst
-        self.__name__ = inst.__name__
-        self.__qualname__ = inst.__qualname__
         self._started = False
         self._callbacks = []
         self._outcome = None
         self._trigger = None
+
+        self._task_id = self._id_count
+        RunningTask._id_count += 1
+        self.__name__ = "Task %d" % self._task_id
+        self.__qualname__ = self.__name__
 
     @lazy_property
     def log(self):
@@ -129,7 +134,48 @@ class RunningTask:
         return self
 
     def __str__(self):
-        return str(self.__qualname__)
+        return "<{}>".format(self.__name__)
+
+    def _get_coro_stack(self):
+        """Get the coroutine callstack of this Task."""
+        coro_stack = extract_coro_stack(self._coro)
+
+        # Remove Trigger.__await__() from the stack, as it's not really useful
+        if self._natively_awaitable and len(coro_stack):
+            if coro_stack[-1].name == '__await__':
+                coro_stack.pop()
+
+        return coro_stack
+
+    def __repr__(self):
+        coro_stack = self._get_coro_stack()
+
+        if cocotb.scheduler._current_task is self:
+            fmt = "<{name} running coro={coro}()>"
+        elif self._finished:
+            fmt = "<{name} finished coro={coro}() outcome={outcome}>"
+        elif self._trigger is not None:
+            fmt = "<{name} pending coro={coro}() trigger={trigger}>"
+        elif not self._started:
+            fmt = "<{name} created coro={coro}()>"
+        else:
+            fmt = "<{name} adding coro={coro}()>"
+
+        try:
+            coro_name = coro_stack[-1].name
+        # coro_stack may be empty if:
+        # - exhausted generator
+        # - finished coroutine
+        except IndexError:
+            coro_name = self._coro.__name__
+
+        repr_string = fmt.format(
+            name=self.__name__,
+            coro=coro_name,
+            trigger=self._trigger,
+            outcome=self._outcome
+        )
+        return repr_string
 
     def _advance(self, outcome):
         """Advance to the next yield in this coroutine.
@@ -238,7 +284,7 @@ class RunningTest(RunningCoroutine):
     def __init__(self, inst, parent):
         self.error_messages = []
         RunningCoroutine.__init__(self, inst, parent)
-        self.log = SimLog("cocotb.test.%s" % self.__qualname__, id(self))
+        self.log = SimLog("cocotb.test.%s" % inst.__qualname__, id(self))
         self.started = False
         self.start_time = 0
         self.start_sim_time = 0
@@ -246,10 +292,14 @@ class RunningTest(RunningCoroutine):
         self.expect_error = parent.expect_error
         self.skip = parent.skip
         self.stage = parent.stage
-        self._id = parent._id
+        self.__name__ = "Test %s" % inst.__name__
+        self.__qualname__ = "Test %s" % inst.__qualname__
 
         # make sure not to create a circular reference here
         self.handler = RunningTest.ErrorLogHandler(self.error_messages.append)
+
+    def __str__(self):
+        return "<{}>".format(self.__name__)
 
     def _advance(self, outcome):
         if not self.started:
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -227,9 +227,6 @@ class Scheduler:
         # indexed by trigger
         self._trigger2coros = _py_compat.insertion_ordered_dict()
 
-        # A dictionary mapping coroutines to the trigger they are waiting for
-        self._coro2trigger = _py_compat.insertion_ordered_dict()
-
         # Our main state
         self._mode = Scheduler._MODE_NORMAL
 
@@ -285,7 +282,6 @@ class Scheduler:
 
             self._timer1.prime(self._test_completed)
             self._trigger2coros = _py_compat.insertion_ordered_dict()
-            self._coro2trigger = _py_compat.insertion_ordered_dict()
             self._terminate = False
             self._write_calls = _py_compat.insertion_ordered_dict()
             self._writes_pending.clear()
@@ -470,12 +466,9 @@ class Scheduler:
         """Unschedule a coroutine.  Unprime any pending triggers"""
 
         # Unprime the trigger this coroutine is waiting on
-        try:
-            trigger = self._coro2trigger.pop(coro)
-        except KeyError:
-            # coroutine probably finished
-            pass
-        else:
+        trigger = coro._trigger
+        if trigger is not None:
+            coro._trigger = None
             if coro in self._trigger2coros.setdefault(trigger, []):
                 self._trigger2coros[trigger].remove(coro)
             if not self._trigger2coros[trigger]:
@@ -523,7 +516,7 @@ class Scheduler:
 
     def _resume_coro_upon(self, coro, trigger):
         """Schedule `coro` to be resumed when `trigger` fires."""
-        self._coro2trigger[coro] = trigger
+        coro._trigger = trigger
 
         trigger_coros = self._trigger2coros.setdefault(trigger, [])
         if coro is self._write_coro_inst:
@@ -758,6 +751,7 @@ class Scheduler:
 
         coro_completed = False
         try:
+            coroutine._trigger = None
             result = coroutine._advance(send_outcome)
             if _debug:
                 self.log.debug("Coroutine %s yielded %s (mode %d)" %
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -426,7 +426,7 @@ class Scheduler:
                     continue
 
                 if _debug:
-                    debugstr = "\n\t".join([coro.__qualname__ for coro in scheduling])
+                    debugstr = "\n\t".join([coro._coro.__qualname__ for coro in scheduling])
                     if len(scheduling):
                         debugstr = "\n\t" + debugstr
                     self.log.debug("%d pending coroutines for event %s%s" %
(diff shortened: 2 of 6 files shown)


cocotb/utils.py
```diff
@@ -578,3 +579,36 @@ def remove_traceback_frames(tb_or_exc, frame_names):
...
+                f = getattr(coro, 'gi_frame')
+                coro = coro.gi_yieldfrom
+            except AttributeError:
+                f = None
+                coro = None
+        if f is not None:
+            yield (f, f.f_lineno)
+
+
+def extract_coro_stack(coro, limit=None):
```
garmin-mjames, line 601: This is based on `traceback.extract_tb`, using the `StackSummary.extract` classmethod for ease of use.
I think it might be possible to get the `__qualname__` for each coroutine in the stack and display that.

cocotb/decorators.py
```diff
@@ -128,7 +134,50 @@ def __iter__(self):
         return self
 
     def __str__(self):
-        return str(self.__qualname__)
+        return "<{}>".format(self.__name__)
+
+    def _get_coro_stack(self):
+        """
+        Get the coroutine callstack of this Task.
+        """
```
cmarqu, line 142:
```suggestion
        """Get the coroutine callstack of this Task."""
```
https://www.python.org/dev/peps/pep-0257/#one-line-docstrings (because I don't have to say anything substantial :)

cocotb/scheduler.py
```diff
@@ -81,6 +83,15 @@ def __exit__(self, *excinfo):
         _profile.disable()
 
 
+@contextmanager
+def task_context(scheduler, task):
+    """Context manager for currently running task."""
```
cmarqu, line 88:
```suggestion
    """Context manager for the currently running task."""
```

cocotb/utils.py
```diff
@@ -578,3 +579,36 @@ def remove_traceback_frames(tb_or_exc, frame_names):
             assert tb.tb_frame.f_code.co_name == frame_name
             tb = tb.tb_next
         return tb
+
+
+def walk_coro_stack(coro):
+    """Walk down the coroutine stack. Supports coroutines and generators."""
```
cmarqu, line 585:
```suggestion
    """Walk down the coroutine stack, starting at *coro*.
    
    Supports coroutines and generators."""
```
?

cocotb/utils.py
```diff
@@ -578,3 +579,36 @@ def remove_traceback_frames(tb_or_exc, frame_names):
...
+            yield (f, f.f_lineno)
+
+
+def extract_coro_stack(coro, limit=None):
+    """
+    Return a :class:`traceback.StackSummary` object representing a list of
+    pre-processed entries from traceback.
+
+    This is useful for alternate formatting of stack traces.  If
+    'limit' is omitted or None, all entries are extracted.  A
```
cmarqu, line 607:
```suggestion
    *limit* is omitted or None, all entries are extracted.  A
```

cocotb/utils.py
```diff
@@ -578,3 +579,36 @@ def remove_traceback_frames(tb_or_exc, frame_names):
...
+                coro = coro.gi_yieldfrom
+            except AttributeError:
+                f = None
+                coro = None
+        if f is not None:
+            yield (f, f.f_lineno)
+
+
+def extract_coro_stack(coro, limit=None):
+    """
```
cmarqu, line 602: There should be one-line summary on the same line as the opening """.

cocotb/utils.py
```diff
@@ -578,3 +579,36 @@ def remove_traceback_frames(tb_or_exc, frame_names):
...
+    Return a :class:`traceback.StackSummary` object representing a list of
+    pre-processed entries from traceback.
+
+    This is useful for alternate formatting of stack traces.  If
+    'limit' is omitted or None, all entries are extracted.  A
+    pre-processed stack trace entry is a FrameSummary object
+    containing attributes filename, lineno, name, and line
```
cmarqu, line 609:
```suggestion
    pre-processed stack trace entry is a :class:`traceback.FrameSummary` object
    containing the attributes ``filename``, ``lineno``, ``name``, and ``line``
```

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -263,3 +264,81 @@ async def waiter(event):
     await NullTrigger()
 
     assert waiter_scheduled is True
+
+
+@cocotb.test()
+async def test_task_repr(dut):
+    """
+    Test RunningTask.__repr__
+    """
```
cmarqu, line 273:
```suggestion
    """Test RunningTask.__repr__"""
```
(Our tests are not usually that polished, but why not start :)

cocotb/decorators.py
```diff
@@ -128,7 +134,50 @@ def __iter__(self):
...
+        if cocotb.scheduler._current_task is self:
+            fmt = "<{name} running coro={coro}()>"
+        elif self._finished:
+            fmt = "<{name} finished coro={coro}() outcome={outcome}>"
+        elif self._trigger is not None:
+            fmt = "<{name} pending coro={coro}() trigger={trigger}>"
+        elif not self._started:
+            fmt = "<{name} created coro={coro}()>"
+        else:
+            fmt = "<{name} adding coro={coro}()>"
```
garmin-mjames, line 164: While the coroutine is running, we can't get the full stack, so for `running` and `adding` states the initial coroutine name is printed.

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -300,3 +301,78 @@ async def second():
     await ReadOnly()
 
     assert dut.array_7_downto_4.value == [10, 2, 3, 4]
+
+
+async def test_task_repr(dut):
```
ktbarrett, line 306:
```suggestion
@cocotb.test()
async def test_task_repr(dut):
```

cocotb/triggers.py
```diff
@@ -696,7 +696,7 @@ class Combine(_AggregateWaitable):
 
     async def _wait(self):
         waiters = []
-        e = Event()
+        e = Event(self)
```
ktbarrett, line 699: I don't think the abuse is a big issue, this object cannot directly leak to the user, but I think this transition could be smoother so we don't see the `Event` at all, it is an implementation detail. I think that would require a special subclass of `Event`. Sounds like a follow up PR.

tests/test_cases/test_cocotb/test_scheduler.py
garmin-mjames: whoops, missed that in the rebase.

garmin-mjames:
@ktbarrett 
> I can imagine each stack level could be a newline and indentation, and trees in the execution due to `Combine` or `First` would be on new lines without indentation.

Where are you imagining this would be printed?

ktbarrett:
That's up to the user, no? Presumably the log. Maybe I'm not understanding what you're saying.

garmin-mjames:
The stack returned from `_get_coro_stack` is only the "task-local" Python coroutine stack outside of the scheduler. Anything involving triggers has to involve the `scheduler`, as that's cross-task. This PR gets us the actual coroutine that has awaited a `Trigger`, not just the name of the coroutine that the task was started with.

I have another follow-on PR for the scheduler debug output that shows the different tasks and what triggers they are pending on when cocotb returns to the simulator.

ktbarrett:
I was just saying I would prefer better formatting.

cocotb/decorators.py
```diff
@@ -237,19 +284,24 @@ def handle(self, record):
...
         self.start_time = 0
         self.start_sim_time = 0
         self.expect_fail = parent.expect_fail
         self.expect_error = parent.expect_error
         self.skip = parent.skip
         self.stage = parent.stage
-        self._id = parent._id
+        self._test_id = parent._id
+        self.__name__ = "Test %s" % inst.__name__
+        self.__qualname__ = "Test %s" % inst.__qualname__
```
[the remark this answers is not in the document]
garmin-mjames: You could overwrite them in `RegressionManager._init_test` before starting the test.

cocotb/scheduler.py
```diff
@@ -81,6 +83,15 @@ def __exit__(self, *excinfo):
         _profile.disable()
 
 
+@contextmanager
+def task_context(scheduler, task):
```
ktbarrett, line 87: Is there a particular reason this is a free function and not a method on `Scheduler`?

cocotb/decorators.py
```diff
@@ -238,19 +284,24 @@ def handle(self, record):
...
+        self.log = SimLog("cocotb.test.%s" % inst.__qualname__, id(self))
         self.started = False
         self.start_time = 0
         self.start_sim_time = 0
         self.expect_fail = parent.expect_fail
         self.expect_error = parent.expect_error
         self.skip = parent.skip
         self.stage = parent.stage
-        self._id = parent._id
+        self._test_id = parent._id
```
ktbarrett, line 295: What's the reason for this change? I don't see it anywhere else.

garmin-mjames: It's an artifact from when all tests were initialized during test discovery, leftover after rebase.

I wanted to sort based on the id's of the `test` object not the id of the created `RunningTask`, so each `RunningTest` would have an `_id` and a `_test_id`, and `RegressionManager` would sort via the latter. I'll remove it as it's no longer needed.

cocotb/scheduler.py
garmin-mjames: I originally tried a custom context manager like `profiling_context` above, but couldn't get it to work correctly, then I found `@contextmanager`. I haven't tried as a method on `Scheduler`.

eric-wieser: `@contextmanager` works just fine on methods

cocotb/scheduler.py
```diff
@@ -81,6 +83,15 @@ def __exit__(self, *excinfo):
...
+    """Context manager for the currently running task."""
+    old_task = scheduler._current_task
+    scheduler._current_task = task
+    yield
+    scheduler._current_task = old_task
```
eric-wieser, line 92:
```suggestion
    scheduler._current_task = task
    try:
        yield
    finally:
        scheduler._current_task = old_task
```

cocotb/scheduler.py
```diff
@@ -246,6 +254,9 @@ def __init__(self):
         self._test = None
         self._main_thread = threading.current_thread()
 
+        self._current_task = None
+        self._task_context = partial(task_context, self)
```
eric-wieser, line 258:
```suggestion
```
once you move `task_context` to a private method

garmin-mjames:
I'm not really satisfied with the implementation of `_InternalEvent`. It should only be used by cocotb internals, but it doesn't properly enforce the single-coroutine waiting because a coroutine could `await e` instead of `await e.wait()` and it would still work.

ktbarrett:
@garmin-mjames You are worried about the inherited implementation of `__await__` in `PythonTrigger` not doing the single-user enforcement check? You could implement `__await__` as a proxy to `.wait()`?

garmin-mjames:
@ktbarrett I'll look at that, thanks.

cocotb/decorators.py
```diff
@@ -87,6 +87,8 @@ class RunningTask:
         triggers to fire.
     """
 
+    _id_count = 0  # used by the scheduler for debug
```
ktbarrett, line 90: What are the merits of this over `id`?

cocotb/triggers.py
```diff
@@ -408,7 +408,7 @@ def wait(self):
         """
         if self.fired:
             return NullTrigger(name="{}.wait()".format(str(self)))
-        return _Event(self)
+        return _EventWaiter(self)
```
ktbarrett, line 411: This probably doesn't need to come in.

cocotb/decorators.py
garmin-mjames: I think the main one is this from the docs:
> Two objects with non-overlapping lifetimes may have the same `id()` value.
But aesthetically I prefer a smaller integer and I think it's easier to match up visually when looking at log outputs.

ktbarrett:
@garmin-mjames One last thing. Could you add a test/example of `Combine` to the test_task_repr?

garmin-mjames:
> 
> 
> @garmin-mjames One last thing. Could you add a test/example of `Combine` to the test_task_repr?

There's a `Combine` in `test_task_repr`. Did you mean `First`?

ktbarrett:
@garmin-mjames nvm me, maybe I typo'd the Ctrl+f. All my reviews seem to end the same way. Losing my mind...

garmin-mjames:
@ktbarrett All good. I added `First` for good measure.

garmin-mjames:
Added another case to `test_task_repr` to get full coverage of `_get_coro_stack`.

garmin-mjames:
What about this case:

Both `<Task 1>` and `<Task 2>` wait on the same `Combine(Timer(1), Join(<Task 3>)` object. They will have identical `repr` outputs for the two separate triggers.

`Event` and `Lock` have the address of each trigger in the `repr` output.

ktbarrett:
> Both `<Task 1>` and `<Task 2>` wait on the same `Combine(Timer(1), Join(<Task 3>)` object.

I know our system allows you to do this, but I don't think it's ever a good idea. There are a lot of issues when it comes to exceptions if you do that, same thing if you await on the same task in more than one task. Future-like objects should only have one awaiting task, well defined programs form a tree.

ktbarrett:
@garmin-mjames I'm happy, are you?

ktbarrett:
Well, if you want to change something we still have several months before the release.

garmin-mjames:
@ktbarrett Yeah, I'm happy with where this is at now, unless you think `RunningTask.__repr__` needs a docstring.

PR #1806, opened 2020-05-13
Fix segfault due to uninitialized PyTypeObject

tpambor:
On importing `cocotb.simulator` the program may segfault in `PyType_Ready` due to accesses to invalid addresses.

Fix the segfault by initiliazing unused fields of `PyTypeObject` to `0`.


tpambor:
CI seems to have some hiccups. Should be green.

eric-wieser:
CI is still running

eric-wieser:
I'm surprised we weren't caught out by this before. Perhaps return value optimization intersected in some weird way with static variable 0-initialization and we got away with it. Either way, this clearly wasn't guaranteed to work by the standard.

PR #1807, opened 2020-05-13
Make Kaleb maintainer

imphil:
Kaleb has consistently shown great work in various parts of cocotb, and
we're happy that he has agreed to become co-maintainer of cocotb.


imphil:
CC @ktbarrett

cmarqu:
\o/

PR #1529, opened 2020-03-22
Update README.md with a simple example

themperek:
Adding a very simple and easy to understand example to `README.md` that would better explain the idea behind cocotb. The result can be seen here: https://github.com/themperek/cocotb/tree/update_readmy

Contributions welcome.

Closes #1448

PR #1811, opened 2020-05-13
Give more context when truncating a `BinaryValue`

cmarqu:
Print original and truncated value with a reworded message.
Add more tests for the different `BinaryRepresentation`s.

This was triggered by @ktbarrett's comment https://github.com/cocotb/cocotb/pull/1529#discussion_r404097212 and implements two thirds of the suggestions there.

### DIFF
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -237,12 +237,12 @@ class BinaryValue:
             else:
                 rv = '0' * (self._n_bits - l) + x
         elif l > self._n_bits:
-            print("WARNING: truncating value to match requested number of bits "
-                  "(%d -> %d)" % (l, self._n_bits))
             if self.big_endian:
                 rv = x[l - self._n_bits:]
             else:
                 rv = x[:l - self._n_bits]
+            warnings.warn("{}-bit value requested, truncating value {!r} ({} bits) to {!r}".format(
+                self._n_bits, x, l, rv), category=RuntimeWarning, stacklevel=3)
         return rv
 
     def _adjust_signed_mag(self, x):
@@ -258,12 +258,12 @@ class BinaryValue:
                 rv = '0' * (self._n_bits - 1 - l) + x[1:]
                 rv = x[0] + rv
         elif l > self._n_bits:
-            print("WARNING: truncating value to match requested number of bits "
-                  "(%d -> %d)" % (l, self._n_bits))
             if self.big_endian:
                 rv = x[l - self._n_bits:]
             else:
                 rv = x[:-(l - self._n_bits)]
+            warnings.warn("{}-bit value requested, truncating value {!r} ({} bits) to {!r}".format(
+                self._n_bits, x, l, rv), category=RuntimeWarning, stacklevel=3)
         else:
             rv = x
         return rv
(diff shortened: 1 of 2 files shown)


cocotb/binary.py
```diff
@@ -237,12 +237,12 @@ def _adjust_unsigned(self, x):
...
                 rv = '0' * (self._n_bits - l) + x
         elif l > self._n_bits:
-            print("WARNING: truncating value to match requested number of bits "
-                  "(%d -> %d)" % (l, self._n_bits))
             if self.big_endian:
                 rv = x[l - self._n_bits:]
             else:
                 rv = x[:l - self._n_bits]
+            print("WARNING: %d-bit value requested, truncating value %s (%d bits) to %s"
+                  % (self._n_bits, x, l, rv))
```
eric-wieser, line 245: Let's switch to using `warnings.warn` while we're here.

ktbarrett: Switching to warnings will cause failures in CI when adding tests for this feature. Or at least has in the past.

cmarqu: @ktbarrett I have covered some in ``tests/pytest/test_binary_value.py``, let's see how many more there are.

cmarqu: Hey, passes.

tests/pytest/test_binary_value.py
```diff
@@ -111,25 +111,39 @@ def test_init_unsigned_negative_value():
...
-    assert bin2.binstr == "110000101100"
-    assert bin2.integer == -980
-
-    bin3 = BinaryValue(value="1111110000101100", n_bits=11, bigEndian=False,
-                       binaryRepresentation=BinaryRepresentation.SIGNED_MAGNITUDE)
-    assert bin3._str == "10000101100"
-    assert bin3.binstr == "10000101100"
-    assert bin3.integer == -44
+def test_init_not_enough_bits():
+    with pytest.warns(RuntimeWarning, match=r"\d+-bit value requested, truncating value"):
```
eric-wieser, line 115: You should have one of these around _each_ line you expect to warn

cocotb/binary.py
```diff
@@ -237,12 +237,12 @@ def _adjust_unsigned(self, x):
...
                 rv = '0' * (self._n_bits - l) + x
         elif l > self._n_bits:
-            print("WARNING: truncating value to match requested number of bits "
-                  "(%d -> %d)" % (l, self._n_bits))
             if self.big_endian:
                 rv = x[l - self._n_bits:]
             else:
                 rv = x[:l - self._n_bits]
+            warnings.warn("{}-bit value requested, truncating value {} ({} bits) to {}".format(
+                self._n_bits, x, l, rv), category=RuntimeWarning)
```
eric-wieser, line 245:
```suggestion
                self._n_bits, x, l, rv), category=RuntimeWarning, stacklevel=n)
```
Where `n > 2` is some integer that makes the warning message show an interesting line to the user.

cmarqu: Done.

tests/pytest/test_binary_value.py
cmarqu: Done.

PR #1816, opened 2020-05-14
Remove duplicate always block and fix alignment

elgorwi:


elgorwi:
> The PR does what it says it does. There is other cleanup possible if you want to expand the PR's scope.

These things just caught my eye. Feel free to suggest any more cleanup.

PR #1822, opened 2020-05-14
[Do not merge, RFC] Delay creation of compiler and linker flags to build time, part 1

tpambor:
Delay creation of compiler and linker flags to build time during which we know which compiler will be invoked

Part 1 does not remove anything. It just checks if the flags before and after restructuring are identical
Part 2 can then remove `extra_link_args`, `extra_compile_args` from the extension definitions

Advantages:
- More than one compiler can be supported per platform
- Cross compilation is possible
- Possible to probe compiler flags

### DIFF
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -19,6 +19,12 @@ logger = logging.getLogger(__name__)
 cocotb_share_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "cocotb", "share"))
 
 
+# TODO [gh-1372]: make this work for MSVC which has a different flag syntax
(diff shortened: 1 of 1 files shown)


cocotb_build_libs.py
```diff
@@ -38,6 +44,34 @@ def run(self):
...
+    def build_extensions(self):
+        for ext in self.extensions:
+            refactor = {"extra_link_args":[], "extra_compile_args":[]}
+
+            refactor["extra_compile_args"] += _extra_cxx_compile_args
+
+            rpaths = ext.library_dirs
+
+            ext.library_dirs.append(os.path.join(self.build_lib, "cocotb", "libs"))
+            if os.path.split(ext.name)[-1] == "simulator":
```
eric-wieser, line 56: Not a huge fan of having this knowledge here. Can we just add $ORIGIN to library_dirs?

PR #1824, opened 2020-05-14
Force doctest for `cocotb.utils` to run uncolored

cmarqu:
When color output had been requested via environment variables,
`test_utils` would fail because of the ANSI codes being put into
the output of `hexdiffs()`.
Thus, force this test to run in an uncolored environment.

### DIFF
--- a/tests/test_cases/test_cocotb/test_doctests.py
+++ b/tests/test_cases/test_cocotb/test_doctests.py
@@ -2,12 +2,15 @@
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
 import doctest
+import os
 
 import cocotb
 
 
 @cocotb.test()
 async def test_utils(dut):
+    # prevent failure in case colored output is requested from the environment
+    os.environ['COCOTB_ANSI_OUTPUT'] = "0"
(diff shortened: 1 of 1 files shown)


cmarqu:
An example of a fail:
```
Failed example:
    print(hexdiffs(b'a', b'b'))
Expected:
    0000      61                                               a
         0000 62                                               b
    <BLANKLINE>
Got:
    [33m0000[49m[39m      [31m61[49m[39m                                               [46m[30ma[49m[39m
         [33m0000[49m[39m [31m62[49m[39m                                               [46m[30mb[49m[39m
    <BLANKLINE>
```

tests/test_cases/test_cocotb/test_doctests.py
```diff
@@ -2,12 +2,14 @@
...
 # SPDX-License-Identifier: BSD-3-Clause
 import doctest
+import os
 
 import cocotb
 
 
 @cocotb.test()
 async def test_utils(dut):
+    os.environ['COCOTB_ANSI_OUTPUT'] = "0"
```
eric-wieser, line 12: A brief comment about why this is needed would be nice

cmarqu: Done.

PR #1825, opened 2020-05-14
Add cocotb.simulator to Python library reference

cmarqu:


ktbarrett:
It's a good idea to add that `stop_simulator` should not be called by the user because it does not currently exit cleanly.

Is there any way to get documentation on the `gpi_hdl_Object`s?

eric-wieser:
> Is there any way to get documentation on the `gpi_hdl_Object`s?
Yes, we need to add those types as module attributes so that python and Sphinx can see them.

documentation/source/library_reference.rst
```diff
@@ -346,6 +346,14 @@ The Scheduler
     :member-order: bysource
 
 
+Simulator
+---------
+
```
eric-wieser, line 351: Perhaps worth a note like I had in https://github.com/cocotb/cocotb/pull/1568/files#diff-7d491a84366bbf6a395ee614a5de3f21R365?

cmarqu: Done.

eric-wieser: You don't appear to have pushed

cmarqu:
> > Is there any way to get documentation on the `gpi_hdl_Object`s?
> 
> Yes, we need to add those types as module attributes so that python and Sphinx can see them.

With https://docs.python.org/3/c-api/structures.html#c.PyMemberDef ?

cmarqu:
> It's a good idea to add that `stop_simulator` should not be called by the user because it does not currently exit cleanly.

This only appears in the "Developer-focused" section so a normal user should stay away from it anyway - will a developer need this info (and just this) in the docs?

eric-wieser:
> > > Is there any way to get documentation on the `gpi_hdl_Object`s?
> > 
> > 
> > Yes, we need to add those types as module attributes so that python and Sphinx can see them.
> 
> With https://docs.python.org/3/c-api/structures.html#c.PyMemberDef ?

I'll address this in a follow-up.

eric-wieser:
> > It's a good idea to add that `stop_simulator` should not be called by the user because it does not currently exit cleanly.
> 
> This only appears in the "Developer-focused" section so a normal user should stay away from it anyway - will a developer need this info (and just this) in the docs?

I think this can be a follow-up to improve the docstrings.

documentation/source/library_reference.rst
cmarqu: I was waiting to see if I could add more, but will push now.

documentation/source/library_reference.rst
```diff
@@ -346,6 +346,19 @@ The Scheduler
     :member-order: bysource
 
 
+The ``cocotb.simulator`` module
+-------------------------------
+
+This module is the interface to the GPI interface.
```
eric-wieser, line 352: Does the `i` in gpi mean interface? Perhaps my original wording was bad, and this should be:
```suggestion
This module is a python wrapper to libgpi.
```

eric-wieser: @ktbarrett, any opinions?

ktbarrett: Looks like a bit of RAS syndrome.

cmarqu: Done.

documentation/source/library_reference.rst
```diff
@@ -346,6 +346,19 @@ The Scheduler
...
+-------------------------------
+
+This module is a Python wrapper to libgpi.
+It should not be considered public API, but is documented here for developers
+of cocotb.
+
+.. automodule:: cocotb.simulator
+    :members:
+    :undoc-members:
+    :member-order: bysource
```
eric-wieser, line 359: Note that this does not actually work, but it might do after https://github.com/sphinx-doc/sphinx/pull/7676 is merged

Issue #1821, opened 2020-05-14
Style guide for HDL examples


imphil:
It would be nice to give our HDL coding examples a uniform style. Ideally, we can find two style guides which are similar between VHDL and Verilog.

For Verilog I'd go with https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md (mostly because I'm quite familiar with it, and it is the most complete Verilog style guide I know with linter and formatter support)

For VHDL I found https://github.com/jeremiah-c-leary/vhdl-style-guide quite useful, but I'm not writing VHDL day-to-day so I'm open to suggestions.

gretzteam:
For Verilog, where is the formatter (automatic?)

ktbarrett:
We really should be using `default_nettype none` as well. I don't see that mentioned in the style guide, but agree with everything else it says.

imphil:
@gretzteam one formatter compatible with that style guide is at https://github.com/google/verible.

ktbarrett:
Should this stick to code style only, or consider behavioral style, like prefer synchronous resets, no latches, examples (not test deisgns) should be synthesizable, etc?

As for the VHDL, I'd like a written out style guide like what was provided for Verilog, a linter is nice but doesn't suggest "best practices". Also I really disagree with some of the decisions the posted VHDL linter makes (I think you can turn them off though).

It's probably not necessary to run a linter in CI, just when changing HDL source files.

Issue #1730, opened 2020-04-29
Convert examples to use `async def` syntax


eric-wieser:
Currently, the examples use [the old `yield` syntax.](https://docs.cocotb.org/en/latest/coroutines.html#generator-based-coroutines)

We should change them to use the now-recommended [`async def` / `await` syntax](https://docs.cocotb.org/en/latest/coroutines.html#async-functions).

Split from #1289.

These are:

* [x] `adder` (#1826)
* [x] `axi_lite_slave` (#1941)
* [x] `dff` (#1752)
* [x] `simple_dff` (always in the new style)
* [x] `endian_swapper` (#1941)
* [x] `mean` (#1902)
* [x] `mixed_language` (#1941)
* [x] ~~`ping_tun_tap`~~ (#2234)

ktbarrett:
#2234 would remove the last unconverted example.

PR #1826, opened 2020-05-15
Cleanup adder

ktbarrett:
As a part of #1730 and #1821, update the adder example as a trial for what others think should be done. The makefile looked okay, so it wasn't changed.

Are we still interested in updating the examples before the release?


examples/adder/hdl/adder.v
```diff
@@ -1,18 +1,16 @@
...
-    parameter   DATA_WIDTH = 4
+  parameter integer DATA_WIDTH = 4
 ) (
-    input      [DATA_WIDTH-1:0] A,
-    input      [DATA_WIDTH-1:0] B,
-    output reg [DATA_WIDTH:0] X
-    );
+  input  logic unsigned [DATA_WIDTH-1:0] A,
+  input  logic unsigned [DATA_WIDTH-1:0] B,
+  output logic unsigned [DATA_WIDTH:0]   X
```
ktbarrett, line 10: Icarus supports the `logic` type, so we should use it. I assume that all supported Verilog simulators support `logic` as well.

examples/adder/hdl/adder.v
```diff
@@ -1,18 +1,16 @@
...
+);
 
-  always @(A or B) begin
-    X = A + B;
-  end
+  assign X = A + B;
```
ktbarrett, line 13: Prefer `assign` statements over single line `always` statements.
*Don't* use sensitivity list in combinational always statements.

examples/adder/hdl/adder.v
```diff
@@ -1,18 +1,16 @@
 // Adder DUT
-
+`default_nettype none
```
ktbarrett, line 2: We should use this in all Verilog files to ensure correctness.

examples/adder/hdl/adder.vhdl
```diff
@@ -1,24 +1,23 @@
 -- Adder DUT
-library ieee ;
-use ieee.std_logic_1164.all;
-use ieee.numeric_std.all;
+library ieee;
+  use ieee.std_logic_1164.all;
+  use ieee.numeric_std.all;
```
ktbarrett, line 4: Style suggested by VSG.

examples/adder/hdl/adder.vhdl
```diff
@@ -1,24 +1,23 @@
...
-port(
-    A : in  unsigned(DATA_WIDTH-1 downto 0);
-    B : in  unsigned(DATA_WIDTH-1 downto 0);
-    X : out unsigned(DATA_WIDTH downto 0)
-    );
-end adder;
+  port (
+    A : in    unsigned(DATA_WIDTH-1 downto 0);
+    B : in    unsigned(DATA_WIDTH-1 downto 0);
+    X : out   unsigned(DATA_WIDTH downto 0));
```
ktbarrett, line 12: Style suggested by VSG, besides the lack of space around the operator in the bounds definition. The Verilog style guide linked in #1821 explicitly excuses the rule for bounds in Verilog, for sake of consistency we should do the same in VHDL. Also, I happen to prefer this.

examples/adder/tests/test_adder.py
```diff
@@ -1,47 +1,38 @@
...
 
     for i in range(10):
+
         A = random.randint(0, 15)
         B = random.randint(0, 15)
 
-        dut.A = A
-        dut.B = B
+        dut.A <= A
+        dut.B <= B
```
ktbarrett, line 33: Prefer this syntax. I've seen at least twice this week people thinking the following should work based on `__setattr__` working. We shouldn't confuse people. Alternatively, we could add support for overloading the assignment operator, which I've mentioned before.

```python
alias = dut.signal
alias = 4
# why isn't dut.signal 4?
```

examples/adder/tests/test_adder.py
```diff
@@ -1,47 +1,38 @@
...
 
-    yield Timer(2, units='ns')
+    await Timer(2, units='ns')
 
-    if int(dut.X) != adder_model(A, B):
-        raise TestFailure(
-            "Adder result is incorrect: %s != 15" % str(dut.X))
-    else:  # these last two lines are not strictly necessary
-        dut._log.info("Ok!")
+    assert int(dut.X) == adder_model(A, B), "Adder result is incorrect: {} != 15".format(dut.X)
```
ktbarrett, line 20: Prefer `assert` over raising a `TestFailure` and use test end to mark success rather than raising `TestSuccess`.

examples/adder/hdl/adder.v
elgorwi: This can be useful to avoid bugs with undeclared signals, but in that case all module inputs need to have the nettype explicitly defined:
```input wire logic unsigned [DATA_WIDTH-1 : 0 ]  A```
Or the ‘logic’ could be implicit:
```input wire unsigned [DATA_WIDTH-1 : 0] A ```

https://stackoverflow.com/questions/52173730/systemverilog-changing-port-type-from-wire-to-logic-gives-error-when-usingn-defa

Comment (elgorwi) [reply examples/adder/hdl/adder.v:2]: @ktbarrett This won't work on all tools unless you define the nettype for all input port explicitly.

File: examples/adder/hdl/adder.v (hunk 1 above)
Comment (elgorwi) [reply examples/adder/hdl/adder.v:10]: The inputs should have the nettype defined as per my previous comment.

Comment (eric-wieser) [reply examples/adder/hdl/adder.v:10]: @ktbarrett, do you agree?

Comment (ktbarrett) [reply examples/adder/hdl/adder.v:10]: Yes, I read through the SO question and did some additional research and the change should be done. Icarus and Questa don't care, but apparently some simulators do print warnings (Vivado errors, but we don't support Vivado so who cares). I think the problem comes down to the interpretation of `default_nettype` and if that applies to input ports. I love Verilog.

File: examples/adder/hdl/adder.v [hunk 8]
```diff
@@ -1,18 +1,16 @@
 // Adder DUT
```
ktbarrett, line 1: Should this file have it's extension changed to `sv`?

examples/adder/hdl/adder.sv
```diff
@@ -0,0 +1,23 @@
+// Adder DUT
+`default_nettype none
+`timescale 1ns/1ps
+
+module adder #(
+  parameter integer DATA_WIDTH = 4
+) (
+  input  wire logic unsigned [DATA_WIDTH-1:0] A,
+  input  wire logic unsigned [DATA_WIDTH-1:0] B,
+  output      logic unsigned [DATA_WIDTH:0]   X
```
eric-wieser, line 10:
```suggestion
  output reg  logic unsigned [DATA_WIDTH:0]   X
```
Net type needed here too

ktbarrett:
The VHDL should be good. @elgorwi Any more comments?

elgorwi:
> Never mind, I think I'm wrong about `reg`.
> 
> I didn't check the VHDL, but assume you did.

Here's a good post about nettypes and data types in systemVerilog:
https://www.verilogpro.com/verilog-reg-verilog-wire-systemverilog-logic/

elgorwi:
> The VHDL should be good. @elgorwi Any more comments?

@ktbarrett In the port declaration of adder.sv, ```unsigned``` specifier is not strictly necessary as by default ```logic``` is already unsigned in systemVerilog. But feel free to leave as is.

examples/adder/hdl/adder.sv
```diff
@@ -0,0 +1,23 @@
...
+);
+
+  assign X = A + B;
+
+`ifndef VERILATOR // traced differently
```
themperek, line 15: Verialtor should support `dumpfile`/`dumpvars` now.

examples/adder/hdl/adder.sv
```diff
@@ -0,0 +1,21 @@
+// Adder DUT
```
imphil, line 1: This file needs a CC0 license header.

examples/adder/hdl/adder.sv
```diff
@@ -0,0 +1,21 @@
...
+`default_nettype none
+`timescale 1ns/1ps
+
+module adder #(
+  parameter integer DATA_WIDTH = 4
+) (
+  input  wire logic unsigned [DATA_WIDTH-1:0] A,
```
imphil, line 8:
```suggestion
  input  logic [DATA_WIDTH-1:0] A,
```

is sufficient.

examples/adder/hdl/adder.sv
```diff
@@ -0,0 +1,21 @@
+// Adder DUT
+`default_nettype none
```
imphil, line 2: Fine with that, but not a fan: every half-decent linter (including Verilator) will catch bugs prevented with this line.

examples/adder/hdl/adder.sv
```diff
@@ -0,0 +1,21 @@
...
+) (
+  input  wire logic unsigned [DATA_WIDTH-1:0] A,
+  input  wire logic unsigned [DATA_WIDTH-1:0] B,
+  output      logic unsigned [DATA_WIDTH:0]   X
+);
+
+  assign X = A + B;
+
+// Dump waves
+initial begin
```
imphil, line 16: Indent that block also with two spaces.

examples/adder/model/adder_model.py
```diff
@@ -1,3 +1,5 @@
-def adder_model(a, b):
+
```
imphil, line 1: Add a CC0 license header.

examples/adder/hdl/adder.sv
eric-wieser: I've been bitten by this line being absent in the past - either our tools let it slip silently, or the other engineers on the project had configured all warnings to be ignored. So I'd be in favor of keeping it the way @ktbarrett has it.

ktbarrett: The Verilog style guide in #1821 doesn't mention it.

I'm fine either way, I haven't touched Verilog seriously since college so I have no strong opinions on style. I simply followed the style guide and added that since it's something I remember using in college.

ktbarrett: Left off `default_nettype`.

PR #1835, opened 2020-05-16
Prevent printing summary multiple times

### LINKED ISSUE #1803: SimFailure causes test results to be printed multiple times

ktbarrett:
When a sim fails, it sets the outcome of the remaining tests to `SimFailure`, which causes the `tear_down` procedure to re-enter and print the summary multiple times. We add a bit of logic to prevent `tear_down` from being run more than once.

This is a temporary fix, a more permanent solution is pending in #1782.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -95,6 +95,7 @@ class RegressionManager:
         self.count = 0
         self.skipped = 0
         self.failures = 0
+        self._tearing_down = False
 
         # Setup XUnit
         ###################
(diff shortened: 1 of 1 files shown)


ktbarrett:
Before: https://travis-ci.org/github/cocotb/cocotb/jobs/687458891#L708
After: https://travis-ci.org/github/cocotb/cocotb/jobs/687688492#L712

cocotb/regression.py
```diff
@@ -229,6 +230,11 @@ def _init_hook(self, hook: Hook) -> Optional[RunningTask]:
             return cocotb.scheduler.add(test)
 
     def tear_down(self) -> None:
+        # prevent re-entering the tear down procedure
+        if not self._tearing_down:
+            self._tearing_down = True
+        else:
+            return
```
eric-wieser, line 237: super-nit:
```suggestion
            return

```

PR #1836, opened 2020-05-16
Support rpath-like "<probing privatePath='' />" on Windows

### LINKED ISSUE #1829: ImportError: from cocotb import simulator
Just pulled the latest from master and build is broken on windows mingw 32-bit.
All is working fine using version b08bf0ce74144886c17e6f824438890d08503a71
Could be related to this change:
https://github.com/cocotb/cocotb/commit/3a0bf9460515ae0c2269890b75ddbeaa80085cff

Here's a log when running the adder example: 
``` make SIM=modelsim
  File "C:\msys32\mingw32\bin\cocotb-config-script.py", line 11, in <module>
    load_entry_point('cocotb', 'console_scripts', 'cocotb-config')()
  File "C:/msys32/mingw32/lib/python3.8/site-packages/pkg_resources/__init__.py", line 490, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "C:/msys32/mingw32/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2859, in load_entry_point
    return ep.load()
  File "C:/msys32/mingw32/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2450, in load
    return self.resolve()
  File "C:/msys32/mingw32/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2456, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "C:/work/cocotb/cocotb/__init__.py", line 42, in <module>
    import cocotb.handle
  File "C:/work/cocotb/cocotb/handle.py", line 36, in <module>
    from cocotb import simulator
ImportError: DLL load failed while importing simulator: The specified module could not be found.
Makefile:47: /Makefile.sim: No such file or directory
make: *** No rule to make target '/Makefile.sim'.  Stop.
```
Any suggestions?

tpambor:
This PR implements support for the [probing element](https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/file-schema/runtime/probing-element) which can be used to specify a rpath-like search path for a assembly which for us equals to a shared library (.dll).

A `Manifest` is embedded into the shared libraries on Windows. On load, these libraries will then be handled via the [SxS](https://en.wikipedia.org/wiki/Side-by-side_assembly) loader. The loader reads the `*.2.config` configuration file. In the configuration file, using the probing element the relative search path is specified.

With this PR it is no longer required to have the `libs` folder in `PATH` or added via `AddDllDirectory`.

Closes #1829


themperek:
@tpambor Just to let you know. Thank you for all the work. I think it all should get in (other PRs too) but we will wait until 1.5 (1.4 release that should be very soon).

themperek:
This also can be removed? 

https://github.com/cocotb/cocotb/blob/924f35a3b7d39543118b7bfaed77dd4808e6612b/cocotb/share/makefiles/Makefile.inc#L82-L86

cocotb_build_libs.py
```diff
@@ -19,6 +19,56 @@
...
+        '\n'
+        'ISOLATIONAWARE_MANIFEST_RESOURCE_ID RT_MANIFEST\n'
+        'BEGIN\n'
+        '    "<?xml version=""1.0"" encoding=""UTF-8"" standalone=""yes""?>"\n'
+        '    "<assembly xmlns=""urn:schemas-microsoft-com:asm.v1"" manifestVersion=""1.0"">"\n'
+        '        "<assemblyIdentity name=""%s"" version=""1.0.0.0"" type=""win32"" processorArchitecture=""%s"" />"\n'
+        '        "<file name=""%s"" />"\n'
+        '%s'
+        '    "</assembly>"\n'
+        'END\n' % (name, archtitecture, filename, "".join(dependencies)))
```
eric-wieser, line 47: Likely more readable with textwrap.dedent.
```suggestion
    rc_body = textwrap.dedent('''
        #include <WinUser.h>

        LANGUAGE 0x00, 0x00

        ISOLATIONAWARE_MANIFEST_RESOURCE_ID RT_MANIFEST
        BEGIN
            "<?xml version=""1.0"" encoding=""UTF-8"" standalone=""yes""?>"
            "<assembly xmlns=""urn:schemas-microsoft-com:asm.v1"" manifestVersion=""1.0"">"
                "<assemblyIdentity name=""%s"" version=""1.0.0.0"" type=""win32"" processorArchitecture=""%s"" />"
                "<file name=""%s"" />"
        %s
            "</assembly>"
        END''') % (name, archtitecture, filename, "".join(dependencies))
```

cocotb_build_libs.py
```diff
@@ -19,6 +19,56 @@
 cocotb_share_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "cocotb", "share"))
 
 
+def create_rc_file(name, filename, libraries):
+    archtitecture = "amd64" if sys.maxsize > 2**32 else "x86"
```
eric-wieser, line 23:
```suggestion
    architecture = "amd64" if sys.maxsize > 2**32 else "x86"
```

cocotb_build_libs.py
```diff
@@ -19,6 +19,56 @@
...
+        '    <windows>\n'
+        '        <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">\n'
+        '            <probing privatePath="libs" />\n'
+        '        </assemblyBinding>\n'
+        '    </windows>\n'
+        '</configuration>\n')
+
+    dirpath = os.path.dirname(filename)
+    if dirpath and not os.path.exists(dirpath):
+        os.makedirs(dirpath)
```
eric-wieser, line 66:
```suggestion
    os.makedirs(dirpath, exist_ok=True)
```

cocotb_build_libs.py
```diff
@@ -19,6 +19,56 @@
...
+
+    with open(filename + ".2.config", "w", encoding='ascii') as f:
```
eric-wieser, line 68: Is it possible for a user non-ascii path component to end up in here?

cocotb_build_libs.py
```diff
@@ -19,6 +19,56 @@
 cocotb_share_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "cocotb", "share"))
 
 
+def create_rc_file(name, filename, libraries):
```
eric-wieser, line 22: I think this needs a comment linking to the docs on what exactly an rc file is.

tpambor:
> This also can be removed?
> 
> https://github.com/cocotb/cocotb/blob/924f35a3b7d39543118b7bfaed77dd4808e6612b/cocotb/share/makefiles/Makefile.inc#L82-L86

This is no longer necessary, I removed it.

cocotb_build_libs.py
tpambor: I have changed it to textwrap.dedent(...)

cocotb_build_libs.py
tpambor: It should not be possible because the paths are only relative and therefore only contain `libs` but I changed it to UTF-8 as it doesn't hurt either.

cocotb_build_libs.py
tpambor: I have added comments and links to the documentation provided by Microsoft.

cocotb_build_libs.py
```diff
@@ -36,6 +127,20 @@ def run(self):
...
+            create_rc_file("libcocotb", "libcocotb.dll", ["libcocotbutils", "libgpilog"])
+            create_rc_file("libgpi", "libgpi.dll", ["libcocotbutils", "libgpilog", "libcocotb"])
+            create_rc_file("simulator", _build_ext.get_ext_filename(self, "simulator"), ["libcocotbutils", "libgpilog", "libgpi"])
+            create_rc_file("libcocotbvpi_icarus", "libcocotbvpi_icarus.vpl", ["libgpi", "libgpilog"])
+            create_rc_file("libcocotbvpi_modelsim", "libcocotbvpi_modelsim.dll", ["libgpi", "libgpilog"])
+            create_rc_file("libcocotbvpi_ghdl", "libcocotbvpi_ghdl.dll", ["libgpi", "libgpilog"])
+            create_rc_file("libcocotbvpi_aldec", "libcocotbvpi_aldec.dll", ["libgpi", "libgpilog"])
+            create_rc_file("libcocotbvhpi_aldec", "libcocotbvhpi_aldec.dll", ["libgpi", "libgpilog"])
+            create_rc_file("libcocotbfli_modelsim", "libcocotbfli_modelsim.dll", ["libgpi", "libgpilog"])
+            create_sxs_appconfig(self.get_ext_fullpath(os.path.join("cocotb", "simulator")))
```
eric-wieser, line 142: Can you construct this by walking over the list of extension modules, rather than repeating the dependencies here? I'm ok if we still have to special case `simulator` somehow.

cocotb_build_libs.py
```diff
@@ -19,6 +20,96 @@
...
+    """
+    Create side-by-side (sxs) assembly manifest
+    It contains dependencies to other assemblies (in our case the assemblies are equal to the other libraries)
+    For more details see:
+     - https://docs.microsoft.com/en-us/windows/win32/sbscs/assembly-manifests
+     - https://docs.microsoft.com/en-us/windows/win32/sbscs/using-side-by-side-assemblies
+
+    `name` is the name of the assembly for which the manifest is generated, e.g. `libcocotbutils`
+    `filename` is the filename of the library, e.g. `libcocotbutils.dll`
+    `libraries` is a list of names of dependent manifests, e.g. `["libgpilog"]`
```
eric-wieser, line 33: In case we ever build sphinx docs from this:

```suggestion
    Create side-by-side (sxs) assembly manifest
    
    It contains dependencies to other assemblies (in our case the assemblies are equal to the other libraries).
    For more details see:
     - https://docs.microsoft.com/en-us/windows/win32/sbscs/assembly-manifests
     - https://docs.microsoft.com/en-us/windows/win32/sbscs/using-side-by-side-assemblies

    Args:
        name: The name of the assembly for which the manifest is generated, e.g. ``libcocotbutils``.
        filename: The filename of the library, e.g. ``libcocotbutils.dll``.
        libraries: A list of names of dependent manifests, e.g. ``["libgpilog"]``.
```

cocotb_build_libs.py
```diff
@@ -19,6 +20,96 @@
 cocotb_share_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "cocotb", "share"))
 
 
+def create_sxs_assembly_manifest(name, filename, libraries):
```
eric-wieser, line 23:
```suggestion
def create_sxs_assembly_manifest(name: str, filename: str, libraries: List[str]) -> s:
```

eric-wieser: (with `from typing import List` at the top)

cocotb_build_libs.py
```diff
@@ -19,6 +20,96 @@
...
+        ''') % (name, architecture, filename, textwrap.indent("".join(dependencies), '    ').strip())
+
+    return manifest_body
+
+
+def create_sxs_appconfig(filename):
+    """
+    Create side-by-side (sxs) application configuration file
+    The application configuration specifies additional search paths for manifests
+    For more details see: https://docs.microsoft.com/en-us/windows/win32/sbscs/application-configuration-files
```
eric-wieser, line 64:
```suggestion
    Create side-by-side (sxs) application configuration file.

    The application configuration specifies additional search paths for manifests.
    For more details see: https://docs.microsoft.com/en-us/windows/win32/sbscs/application-configuration-files
```

cocotb_build_libs.py
```diff
@@ -19,6 +20,96 @@
...
+    os.makedirs(dirpath, exist_ok=True)
+
+    with open(filename + ".2.config", "w", encoding='utf-8') as f:
+        f.write(config_body)
+
+
+def create_rc_file(name, filename, libraries):
+    """
+    Creates windows resource definition script to embed the side-by-side assembly manifest into the libraries
+    For more details see: https://docs.microsoft.com/en-us/windows/win32/menurc/about-resource-files
```
eric-wieser, line 88:
```suggestion
    Creates windows resource definition script to embed the side-by-side assembly manifest into the libraries.

    For more details see: https://docs.microsoft.com/en-us/windows/win32/menurc/about-resource-files
```

cocotb_build_libs.py
```diff
@@ -36,6 +132,18 @@ def run(self):
...
+        if os.name == "nt":
+            create_sxs_appconfig(self.get_ext_fullpath(os.path.join("cocotb", "simulator")))
+
+            ext_names = {os.path.split(ext.name)[-1] for ext in self.extensions}
+            for ext in self.extensions:
+                fullname = self.get_ext_fullname(ext.name)
+                filename = self.get_ext_filename(fullname)
+                name = os.path.split(fullname)[-1]
+                filename = os.path.split(filename)[-1]
+                libraries = list({"lib" + lib for lib in ext.libraries}.intersection(ext_names))
```
eric-wieser, line 144: Nit:
```suggestion
                libraries = {"lib" + lib for lib in ext.libraries}.intersection(ext_names)
```

ktbarrett:
@eric-wieser You happy with this, can it be merged?

PR #1837, opened 2020-05-16
Expose type objects in the `cocotb.simulator` namespace

eric-wieser:
This means that python code can now use `isinstance(obj, cocotb.simulator.gpi_sim_hdl)`.
Perhaps more usefully, it also means these types will appear in the docs.

This also fixes an incorrect use of `tp_name`, which should have contained the fully-qualified name.

https://cocotb--1837.org.readthedocs.build/en/1837/library_reference.html#cocotb.simulator.gpi_cb_hdl

### DIFF
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -849,6 +849,35 @@ static int add_module_constants(PyObject* simulator)
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


ktbarrett:
oh well

eric-wieser:
Still a good idea. The module is missing a docstring in a similar way.

eric-wieser:
Wanted to get this one merged so I could rebase #1838

PR #1840, opened 2020-05-16
gpi: Move the problem of checking if an interface is available to the caller

eric-wieser:
Instead of a bunch of gpi functions reaching for `impls[0]`, this makes it the caller's job to pass in an impl.
This means if no impl is available, we no longer segfault - it becomes the caller's problem, as they simply do not have the handle they need to call our functions.

This does not make most of the changes that are needed to the python extension module, hence its draft status. CI will fail.

### DIFF
--- a/cocotb/share/include/gpi.h
+++ b/cocotb/share/include/gpi.h
@@ -87,9 +87,11 @@ we have to create a process with the signal on the sensitivity list to imitate a
     class GpiObjHdl;
     class GpiCbHdl;
     class GpiIterator;
+    class GpiImplInterface;
     typedef GpiObjHdl *gpi_sim_hdl;
     typedef GpiCbHdl *gpi_cb_hdl;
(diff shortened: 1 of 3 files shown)


eric-wieser:
> For your typical GPI user, they will not be started until they are loaded by the GPI

Those users could be passed a `gpi_interface*` as an argument to the gpi startup function.

ktbarrett:
That's a good idea and would work as a stand in for a boolean flag for us to check if the simulator was GPI loaded. It could come in with argc/argv in `embed_sim_init`, then we would need to transfer ownership to the simulator module (unless they could reasonably become one, not sure if that would work).

ktbarrett:
@eric-wieser Since #1843 is in, does this need to be open?

eric-wieser:
I think the idea is still worthwhile, but as you point out the justification is no longer driven by segfaults or urgent. I think this would help with mocking out the GPI though.

Issue #1839, opened 2020-05-16
Abort (Core dumped) when running unittests with uvm-python


tpoikela:
Running `make test` (or any unittest file) for https://github.com/tpoikela/uvm-python results in the following error:

```
/usr/include/c++/9/bits/stl_vector.h:1042: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = GpiImplInterface*; _
Alloc = std::allocator<GpiImplInterface*>; std::vector<_Tp, _Alloc>::reference = GpiImplInterface*&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
```

I bisected the issue to commit 3a0bf9460515ae0c2269890b75ddbeaa80085cff. It seems some imports have been changed to unconditional ones. More specifically, changing cocotb/utils.py back to conditional import seems to fix this issue. Any ideas on the source of error?

Background: uvm-python unit tests are run without simulator, but files under test are still importing stuff from cocotb.

To reproduce the issue:

```
mkdir temp
cd temp
python -m env .venv
source .venv/bin/activate
git clone https://github.com/cocotb/cocotb.git
cd cocotb
pip install -e .
cd ..
git clone https://github.com/tpoikela/uvm-python.git
cd uvm-python
ln -s src/uvm ./uvm
pip install cocotb_coverage pip install regex
python -m unittest unit/test_sv.py  # Or make test-unit
```

eric-wieser:
Right, I think the key difference is that `cocotb.utils._LOG_SIM_PRECISION` used to default to `1e-15` when not in a simulator, but now it's replacement assumes that anyone requestion something in simsteps must be connected to a simulator.

The crash looks like asking for item 0 from an empty list of simulators.

eric-wieser:
Can you enable the PYTHONFAULTHANDER mechanism to get a python stack trace?

tpoikela:
```
Current thread 0x00007fbec61676c0 (most recent call first):
  File "/home/tpoikela/temp/cocotb/cocotb/utils.py", line 43 in _get_simulator_precision
  File "/home/tpoikela/temp/cocotb/cocotb/utils.py", line 100 in get_time_from_sim_steps
  File "/home/tpoikela/temp/cocotb/cocotb/utils.py", line 73 in get_sim_time
[...]

PR #1843, opened 2020-05-17
Prevent segfaults when using simulatormodule without a simulator

ktbarrett:
A more permanent fix to #1839. When accessing the simulator module without a simulator present, all methods will throw an exception to prevent a possible segfault. `get_sim_precision` is an exception, previous to #1717, this defaulted to `-15`, this behavior has been restored.

@tpoikela

### DIFF
--- a/cocotb/share/include/gpi.h
+++ b/cocotb/share/include/gpi.h
@@ -113,6 +113,13 @@ typedef enum gpi_event_e {
 
 // Functions for controlling/querying the simulation state
 
+/**
+ * Returns 1 if there is a registered GPI implementation, 0 otherwise.
+ *
+ * Useful for checking if a simulator is running.
+ */
+bool gpi_has_registered_impl(void);
+
 // Stop the simulator
 void gpi_sim_end(void);
 
--- a/cocotb/share/lib/gpi/GpiCommon.cpp
+++ b/cocotb/share/lib/gpi/GpiCommon.cpp
@@ -123,6 +123,11 @@ int gpi_register_impl(GpiImplInterface *func_tbl)
     return 0;
 }
 
+bool gpi_has_registered_impl()
+{
+    return registered_impls.size() > 0;
+}
+
 void gpi_embed_init(int argc, char const* const* argv)
 {
     if (embed_sim_init(argc, argv))
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -749,6 +749,14 @@ static PyObject *get_type_string(gpi_hdl_Object<gpi_sim_hdl> *self, PyObject *ar
 }
 
 
+static PyObject *is_running(PyObject *self, PyObject *args)
+{
+    COCOTB_UNUSED(self);
+    COCOTB_UNUSED(args);
+
+    return PyBool_FromLong(gpi_has_registered_impl());
+}
+
 // Returns a high, low, tuple of simulator time
 // Note we can never log from this function since the logging mechanism calls this to annotate
 // log messages with the current simulation time
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -255,6 +255,7 @@ err:
 static PyObject *log_msg(PyObject *self, PyObject *args)
 {
     COCOTB_UNUSED(self);
+
     const char *name;
     const char *path;
     const char *msg;
@@ -295,6 +296,11 @@ static PyObject *register_readonly_callback(PyObject *self, PyObject *args)
 {
     COCOTB_UNUSED(self);
 
+    if (!gpi_has_registered_impl()) {
+        PyErr_SetString(PyExc_RuntimeError, "No simulator available!");
+        return NULL;
+    }
+
     Py_ssize_t numargs = PyTuple_Size(args);
 
     if (numargs < 1) {
@@ -332,6 +338,11 @@ static PyObject *register_readonly_callback(PyObject *self, PyObject *args)
 static PyObject *register_rwsynch_callback(PyObject *self, PyObject *args)
 {
     COCOTB_UNUSED(self);
+
+    if (!gpi_has_registered_impl()) {
+        PyErr_SetString(PyExc_RuntimeError, "No simulator available!");
+        return NULL;
+    }
 
     Py_ssize_t numargs = PyTuple_Size(args);
 
@@ -371,6 +382,11 @@ static PyObject *register_rwsynch_callback(PyObject *self, PyObject *args)
 static PyObject *register_nextstep_callback(PyObject *self, PyObject *args)
 {
     COCOTB_UNUSED(self);
+
+    if (!gpi_has_registered_impl()) {
+        PyErr_SetString(PyExc_RuntimeError, "No simulator available!");
+        return NULL;
+    }
 
     Py_ssize_t numargs = PyTuple_Size(args);
 
@@ -414,6 +430,11 @@ static PyObject *register_nextstep_callback(PyObject *self, PyObject *args)
 static PyObject *register_timed_callback(PyObject *self, PyObject *args)
 {
     COCOTB_UNUSED(self);
+
+    if (!gpi_has_registered_impl()) {
+        PyErr_SetString(PyExc_RuntimeError, "No simulator available!");
+        return NULL;
+    }
 
     Py_ssize_t numargs = PyTuple_Size(args);
 
@@ -472,6 +493,11 @@ static PyObject *register_timed_callback(PyObject *self, PyObject *args)
 static PyObject *register_value_change_callback(PyObject *self, PyObject *args) //, PyObject *keywds)
 {
     COCOTB_UNUSED(self);
+
+    if (!gpi_has_registered_impl()) {
+        PyErr_SetString(PyExc_RuntimeError, "No simulator available!");
+        return NULL;
+    }
 
     Py_ssize_t numargs = PyTuple_Size(args);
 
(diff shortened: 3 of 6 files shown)


documentation/source/newsfragments/1843.change.rst
```diff
@@ -0,0 +1 @@
+Simulator Python module now throws an exception if used when no simulator is present. Making the use of cocotb without a simulator present safe.
```
cmarqu, line 1:
```suggestion
The :mod:`cocotb.simulator` Python module now throws an exception if used when no simulator is present, making the use of cocotb without a simulator present safe.
```
(Cannot test right now if this will result in the correct link.)

cocotb/share/include/cocotb_utils.h
```diff
@@ -43,6 +43,7 @@ extern void* utils_dyn_open(const char* lib_name);
 extern void* utils_dyn_sym(void *handle, const char* sym_name);
 
 extern int is_python_context;
+extern int simulator_is_active;
```
eric-wieser, line 46:
```suggestion
extern bool simulator_is_active;
```

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -227,6 +227,8 @@ extern "C" int embed_sim_init(int argc, char const * const * argv)
     if (pEventFn)
         return ret;
 
+    simulator_is_active = 1;
```
eric-wieser, line 230: Can you convince me this is the right place?

ktbarrett: The line place in the function, or the whole approach? If just the line placement in the function, it's kind of arbitrary, as long as it's set before we load into the Python entry point.

eric-wieser: The line in the function - ideally this would go right next to the code that initializes `registered_impls`

ktbarrett: That's not in the function? You mean this should be moved into the GPI, which is a different approach.

The GPI should only ever be loaded by the simulator, so we can assume once we are in the GPI, there is a valid simulator. Because we want to support loading the simulator module outside of the typical flow, the responsibility should rest with the simulator module to prevent itself from using the GPI when it knows it would be invalid.

Moving the flag into the GPI and adding a function to the GPI interface to return that flag means the user must check the function before using any interface. Robust design dictates we shouldn't trust the user and now we have to add error code handling to the entry points of the GPI.

In short, my approach is "No touchy!". I have tested using all the methods and types available on the simulator module from the terminal and ensured they do not segfault. If you think that warrants a test, I'd be happy to add one. I will also add a `simulator.is_running` function. No matter the approach that would be useful.

eric-wieser: > The GPI should only ever be loaded by the simulator, so we can assume once we are in the GPI, there is a valid simulator.

There are two different bits of the GPI - the C wrapper, which can be loaded by anyone, and the C++ code, which only gets instantiated when the simulator is entered.

I suppose what I'm advocating in the short term is a `gpi_has_registered_impl()` function implemented as `return registered_impls.size() == 0`. We already have `gpi_print_registered_impl()`, so it seems like a reasonable extension

eric-wieser: Actually, the `gpi_print_registered_impl` function already exposes the information you need via its return value, exposing that in the C extension API would do the trick.

ktbarrett: > the C wrapper, which can be loaded by anyone

I think that's the crux of the issue. My assumption is that most users know you can't load any library at any time, start using it, and expect it to not blow up; which allows us not do any of the work in the GPI. You seem to suggesting making the GPI interface tolerant against invalid usage in general, which is fine, it's just more work. #1844 seems related to that.

eric-wieser: I'm suggesting as a short term, we just add a method to the gpi that simply answers the question "will this blow up". We already have one at `gpi_print_registered_impl`, it just prints useless log messages we do not want.

With that method around, you could replace `if (simulator_is_active)` with `if(gpi_has_registered_impl())`. The function call cost (which ultimately just returns a data member of a global anyway) is insignificant compared to the python boundary cost.

ktbarrett: That's fine. I'll update this PR.

eric-wieser: Thanks!

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -744,6 +786,11 @@ static PyObject *get_precision(PyObject *self, PyObject *args)
 {
     COCOTB_UNUSED(self);
     COCOTB_UNUSED(args);
+
+    if (!simulator_is_active) {
+        return PyLong_FromLong(-15);  // preserves old behavior
```
eric-wieser, line 791: It's probably worth emitting a warning here that no simulator is attached:
```suggestion
        if (PyErr_WarnEx(PyExc_RuntimeWarning, "something about the simulator not actually being attached, assuming whatever precision this is", 1) < 0) {
            return NULL;
        }
        return PyLong_FromLong(-15);  // preserves old behavior
```

cocotb/share/include/gpi.h
```diff
@@ -113,6 +113,13 @@ typedef enum gpi_event_e {
 
 // Functions for controlling/querying the simulation state
 
+/**
+ * Returns 1 if there is a registered GPI implementation.
```
ktbarrett, line 117:
```suggestion
 * Returns 1 if there is a registered GPI implementation, 0 otherwise.
```

documentation/source/newsfragments/1843.change.rst
```diff
@@ -0,0 +1 @@
+The :mod:`cocotb.simulator` Python module now throws an exception if used when no simulator is present, making the use of cocotb without a simulator present safe.
```
eric-wieser, line 1:
```suggestion
The :mod:`cocotb.simulator` Python module now throws an :exc`RuntimeError` if used when no simulator is present, making it safe to use cocotb without a simulator present.
```

cocotb/share/include/gpi.h
```diff
@@ -113,6 +113,13 @@ typedef enum gpi_event_e {
 
 // Functions for controlling/querying the simulation state
 
+/**
+ * Returns 1 if there is a registered GPI implementation, 0 otherwise.
+ *
+ * Useful for checking if a simulator is running.
+ */
+int gpi_has_registered_impl(void);
```
eric-wieser, line 121: Any reason not to use bool?
```suggestion
bool gpi_has_registered_impl(void);
```

cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -123,6 +123,11 @@ int gpi_register_impl(GpiImplInterface *func_tbl)
     return 0;
 }
 
+int gpi_has_registered_impl()
```
eric-wieser, line 126:
```suggestion
bool gpi_has_registered_impl()
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -255,6 +255,12 @@ int handle_gpi_callback(void *user_data)
 static PyObject *log_msg(PyObject *self, PyObject *args)
 {
     COCOTB_UNUSED(self);
+
+    if (!gpi_has_registered_impl()) {
+        PyErr_SetString(PyExc_RuntimeError, "No simulator available!");
+        return NULL;
+    }
+
```
eric-wieser, line 263:
```suggestion
```
This one is safe without a simulator, right?

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -810,6 +879,12 @@ static PyObject *deregister(gpi_hdl_Object<gpi_cb_hdl> *self, PyObject *args)
 static PyObject *log_level(PyObject *self, PyObject *args)
 {
     COCOTB_UNUSED(self);
+
+    if (!gpi_has_registered_impl()) {
+        PyErr_SetString(PyExc_RuntimeError, "No simulator available!");
+        return NULL;
+    }
+
```
eric-wieser, line 887: Again, I don't think `gpilog` needs a simulator.
```suggestion
```

cocotb/share/lib/simulator/simulatormodule.h
```diff
@@ -115,7 +115,12 @@ static PyMethodDef SimulatorMethods[] = {
         "--\n\n"
         "Set the log level for GPI"
     )},
-
+    {"is_running", is_running, METH_NOARGS, PyDoc_STR(
+        "is_running()\n"
+        "--\n\n"
+        "is_running() -> bool\n"
+        "Returns ``True`` if the caller is running within a simulator."
```
eric-wieser, line 122: Needs a newsfragment and maybe a `..versionadded`

ktbarrett: Done. Sorry for the late add. Forgot I wanted this in to address the needs of #1839.

documentation/source/newsfragments/1843.feature.rst
```diff
@@ -0,0 +1 @@
+The :func:`cocotb.simulator.is_running` method was added so a user of cocotb could determine if they are running within a simulator.
```
eric-wieser, line 1:
```suggestion
The :func:`cocotb.simulator.is_running` function was added so a user of cocotb could determine if they are running within a simulator.
```

eric-wieser:
Thanks!

PR #1845, opened 2020-05-18
Fix #1519 newsfragment to correctly display str

garmin-mjames:
`str` is not displayed correctly on master:

image

https://docs.cocotb.org/en/latest/release_notes.html


garmin-mjames:
Fixed: https://cocotb--1845.org.readthedocs.build/en/1845/release_notes.html#deprecations-and-removals

PR #1846, opened 2020-05-19
Support coroutines in First and Combine

marlonjames:
Add support for Python coroutines to `First` and `Combine` triggers.

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -29,6 +29,7 @@
 
 import abc
 from collections.abc import Awaitable
+import inspect
 
 from cocotb import simulator
 from cocotb.log import SimLog
@@ -655,17 +656,23 @@ class _AggregateWaitable(Waitable):
     __slots__ = ('triggers',)
 
     def __init__(self, *triggers):
-        self.triggers = tuple(triggers)
+        usable_triggers = []
 
         # Do some basic type-checking up front, rather than waiting until we
         # await them.
         allowed_types = (Trigger, Waitable, cocotb.decorators.RunningTask)
-        for trigger in self.triggers:
-            if not isinstance(trigger, allowed_types):
+        for trigger in triggers:
+            if inspect.iscoroutine(trigger):
+                usable_triggers.append(cocotb.decorators.RunningTask(trigger))
+            elif isinstance(trigger, allowed_types):
+                usable_triggers.append(trigger)
+            else:
                 raise TypeError(
-                    "All triggers must be instances of Trigger! Got: {}"
+                    "All triggers must be awaitable! Got: {}"
                     .format(type(trigger).__qualname__)
                 )
+
+        self.triggers = tuple(usable_triggers)
 
     def __repr__(self):
         # no _pointer_str here, since this is not a trigger, so identity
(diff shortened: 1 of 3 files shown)


ktbarrett:
Some previous discussion https://github.com/cocotb/cocotb/pull/1523#discussion_r396007756.

cocotb/triggers.py
```diff
@@ -655,18 +656,24 @@ class _AggregateWaitable(Waitable):
...
         # await them.
         allowed_types = (Trigger, Waitable, cocotb.decorators.RunningTask)
-        for trigger in self.triggers:
-            if not isinstance(trigger, allowed_types):
+        for trigger in triggers:
+            if inspect.iscoroutine(trigger):
+                usable_triggers.append(cocotb.decorators.RunningTask(trigger))
+            elif isinstance(trigger, allowed_types):
+                usable_triggers.append(trigger)
+            else:
```
ktbarrett, line 669: Do we want to explicitly add coroutine objects to the list of acceptable types or change the predicate to `inspect.isawaitable` since the waiter coroutine just `await`s on each object?

garmin-mjames: I thought about that, but `isawaitable` supports undecorated generators, which we don't support anywhere else that I know of.

garmin-mjames:
@ktbarrett referring to the previous discussion, we could track the `RunningTask`s that are created inside the trigger and kill those at the end of `First`. I think it's better to clean up those tasks and not leave them running.

cocotb/triggers.py
ktbarrett: > but isawaitable supports undecorated generators

You mean `asyncio.coroutine` or `types.coroutine` decorated generators? Regular generators are not awaitable. However, they would likely work (if not so rigidly type-checked by the scheduler) since they are coroutine-like, but I guess `isawaitable` is not the right way to do that.

cocotb/triggers.py
```diff
@@ -815,7 +822,7 @@ async def with_timeout(trigger, timeout_time, timeout_unit=None):
         await with_timeout(First(coro, event.wait()), 100, 'ns')
 
     Args:
-        trigger (:class:`~cocotb.triggers.Trigger` or :class:`~cocotb.triggers.Waitable` or :class:`~cocotb.decorators.RunningTask`):
+        trigger (:class:`~cocotb.triggers.Trigger` or :class:`~cocotb.triggers.Waitable` or :class:`~cocotb.decorators.RunningTask` or :class:`types.CoroutineType`):
```
ktbarrett, line 825:
```suggestion
        trigger (:class:`~cocotb.triggers.Trigger`, :class:`~cocotb.triggers.Waitable`, :class:`~cocotb.decorators.RunningTask`, or :class:`types.CoroutineType`):
```

garmin-mjames: Is there a way to reference `coroutine` in https://docs.python.org/3/glossary.html#term-coroutine? I don't know if `:term:` would link it or not.

ktbarrett:
@garmin-mjames That's awfully odd, and a function that behaves differently depending on the current state of the arguments means building abstractions around it will be frustrating. 

What I think really needs to be done is to create the concept of shielding a task from cancellation, then we can change `First`'s default behavior to always kill everything in cleanup. Those tasks that are protected will continue to run. This means that `Combine` would have to shield all of it's arguments in the call to `First`. This resembles `asyncio.shield`.

cocotb/triggers.py
ktbarrett: A hyperlink. `:term:` seems to require a glossary be present, which is a good idea, but out of scope for this PR.

cmarqu: ``:term:`coroutine` `` or ``:term:`python:coroutine` `` should work even without a cocotb glossary (which we may want to have at some point, so the 2nd, more explicit, syntax is better), the magic is done with intersphinx. If you try this locally, you may have to run ``make clean`` in ``documentation/`` first.

garmin-mjames:
Changed to on-hold until some of the Task mechanics are decided/improved.

ktbarrett:
With `select` and `combine` in I think this can close.

PR #1849, opened 2020-05-20
drivers: Move type information from docstrings to annotations

eric-wieser:
Only the functions which had typesin their docstrings are touched here.
Some instances of `str` were corrected to `bytes`, since those were my fault in a previous PR.
Everything else is left as in the docstring.


ktbarrett:
I'll have some time to look at this over the weekend.

PR #1853, opened 2020-05-22
Don't install unneeded objects with python package

ktbarrett:
We remove `include_package_data` option and instead depend on detecting the cocotb package and specifying the makefiles as cocotb package data.

Previously, this installed everything under cocotb/, but now it only installs the python files and the makefiles.

I don't know much about packaging, so I might be totally off base here.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,8 +97,13 @@ setup(
     install_requires=[],
     python_requires='>=3.5',
     packages=find_packages(),
-    include_package_data=True,
-    package_data={'cocotb': package_files('cocotb/share')},
+    package_data={
+        'cocotb': (
+            package_files('cocotb/share/makefiles') +   # noqa: W504
+            package_files('cocotb/share/include') +     # noqa: W504
+            package_files('cocotb/share/def')
+        )
+    },
     ext_modules=get_ext(),
     entry_points={
         'console_scripts': [


ktbarrett:
tox seems to need MANIFEST.in, but only for build requirements: cocotb_build_libs, and the sources.

eric-wieser:
`MANIFEST.in` is for indicating which files go into the distribution tarball, which should include source, license, etc.

ktbarrett:
But do we need `include_package_data=True` for distribution?

eric-wieser:
Quite possibly, yes. If you leave that in, does it put any files in `site-packages/cocotb` that are unwanted?

ktbarrett:
With the current changes `python setup.py sdist` includes sources mentioned in MANIFEST in the tarball. Installing the package using `pip install .` or creating a wheel and installing from that yield nothing in cocotb/share besides the makefiles. But perhaps we should include the headers...

eric-wieser:
Can you tell the difference between `include_package_data=True` and `include_package_data=False` in terms of:
* What ends up in the tarball?
* What ends up installed?

ktbarrett:
`find $(cocotb-config --prefix)/cocotb > installed_files` was run with values of `install_package_data` for both `True` and `False`.

diff false true:
```diff
56a57,85
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/simulator
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/simulator/simulatormodule.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/simulator/simulatormodule.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/embed
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/embed/gpi_embed.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/gpi_log
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/gpi_log/gpi_logging.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/utils
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/utils/cocotb_utils.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/gpi
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/gpi/GpiCbHdl.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/gpi/gpi_priv.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/gpi/GpiCommon.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/vpi
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/vpi/VpiImpl.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/vpi/VpiCbHdl.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/vpi/VpiImpl.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/vhpi
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/vhpi/VhpiImpl.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/vhpi/VhpiImpl.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/fli
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/fli/FliImpl.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/fli/FliObjHdl.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/fli/FliImpl.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/fli/FliCbHdl.cpp
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/verilator
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/lib/verilator/verilator.cpp
72a102,116
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/include
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/include/vhpi_user.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/include/sv_vpi_user.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/include/vpi_user_ext.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/include/cocotb_utils.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/include/vhpi_user_ext.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/include/gpi_logging.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/include/embed.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/include/vpi_user.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/include/gpi.h
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/def
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/def/icarus.def
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/def/aldec.def
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/def/ghdl.def
> /home/kaleb/anaconda3/envs/py3.8/lib/python3.8/site-packages/cocotb/share/def/modelsim.def
```

There is no difference in the tarball.

eric-wieser:
I'm not used to reading non-git (unified) diffs. The effect here is to add all those files to the install?

ktbarrett:
@eric-wieser Read the diff as "add these files to > (install_package_data=False) to make them equal". Meaning with `install_package_data=False` those files are gone. 

```
pip install --no-use-pep517 -v .
```
[installation.master.log](https://github.com/cocotb/cocotb/files/4702076/installation.master.log)
[installation.branch.log](https://github.com/cocotb/cocotb/files/4702077/installation.branch.log)

You can see we aren't adding the sources to the wheel or installing them anymore. But only MANIFEST.in is used to determine what goes in the tarball, which hasn't changed in this PR.

eric-wieser:
The one thing that concerns me here - you've mentioned repeatedly that you're interested in using `libgpi` without the python bits. For this to work, we need the the headers inside the wheel.

Are you actually linking against `libgpi` today? If so, I feel like it would be more convenient for you if we continued to install the `.h` files as part of the wheel.

ktbarrett:
Yeah, that was my one concern with this change, but I'm not doing it today. We will partially revert some of this later on, but I'd rather use the package_data field to explicitly specify what we want rather than depending on the MANIFEST which includes everything.

eric-wieser:
With this change, it seems a bit weird that we provide `cocotb-config --share` at all, given it no longer contains anything useful other than `cocotb-config --makefiles`. Perhaps we should remove the former, so that users who are somehow looking for the header get a clearer error than "file not found".


Thoughts, @imphil?

ktbarrett:
Hmmm, rather than making that decision now, could we just add the include and def directory to the package_data?

eric-wieser:
Yes, that sounds like a good compromise

setup.py
```diff
@@ -97,8 +97,11 @@ def package_files(directory):
     install_requires=[],
     python_requires='>=3.5',
     packages=find_packages(),
-    include_package_data=True,
-    package_data={'cocotb': package_files('cocotb/share')},
+    package_data={'cocotb':
+        package_files('cocotb/share/makefiles')
+        + package_files('cocotb/share/include')  # noqa: W503
+        + package_files('cocotb/share/def')      # noqa: W503
+    },
```
ktbarrett, line 104: Alternatively:
```suggestion
    package_data={'cocotb': [file for files in (
        package_files('cocotb/share/makefiles'),
        package_files('cocotb/share/include'),
        package_files('cocotb/share/def'),
    ) for file in files]},
```

ktbarrett: Took this when I fixed for flake8.

eric-wieser: Just put the + at the end of the line?

setup.py
```diff
@@ -97,8 +97,13 @@ def package_files(directory):
...
     python_requires='>=3.5',
     packages=find_packages(),
-    include_package_data=True,
-    package_data={'cocotb': package_files('cocotb/share')},
+    package_data={
+        'cocotb': [file for files in (
+            package_files('cocotb/share/makefiles'),
+            package_files('cocotb/share/include'),
+            package_files('cocotb/share/def'),
+        ) for file in files]
```
eric-wieser, line 105: Would be clearer as list concatenation I think

setup.py
```diff
@@ -97,8 +97,13 @@ def package_files(directory):
...
     python_requires='>=3.5',
     packages=find_packages(),
-    include_package_data=True,
-    package_data={'cocotb': package_files('cocotb/share')},
+    package_data={
+        'cocotb': [
+            *package_files('cocotb/share/makefiles'),
+            *package_files('cocotb/share/include'),
+            *package_files('cocotb/share/def'),
+        ]
```
eric-wieser, line 105: This is still pretty weird compared to
```suggestion
        'cocotb': (
            package_files('cocotb/share/makefiles') +
            package_files('cocotb/share/include') +
            package_files('cocotb/share/def')
        )
```

ktbarrett: The first allows you to specify particular files because it is already a list. But I'll change it

PR #1860, opened 2020-05-23
Tox: Whitelist common license file variables

imphil:
Proprietary EDA tools need a license which is typically provided through
FlexLM. The correct license file is found through an environment
variable which we need to pass through to the tool.


imphil:
Just to state the obvious: I know there are more variables, let's add them as needed in subsequent PRs.

Issue #1857, opened 2020-05-23
[Riviera-Pro] Simulator crash in test_vhdl_access.port_not_hierarchy


imphil:
Riviera-PRO 2019.10 on linux64 crashes in the test_vhdl_access.port_not_hierarchy test. Could be related to #1854.

```
$ make -C ~/src/cocotb/tests/test_cases/test_vhdl_access SIM=riviera COCOTB_REDUCED_LOG_FMT=1 clean results.xml

# COUT:      0.00ns INFO     Running test 1/3: check_enum_object
# COUT:      0.00ns INFO     Starting test: "check_enum_object"
# COUT:                      Description: 
# COUT:                          Enumerations currently behave as normal signals
# COUT:                      
# COUT:                          TODO: Implement an EnumObject class and detect valid string mappings
# COUT:                          
# COUT:      0.10ns INFO     Test Passed: check_enum_object
# COUT:      0.10ns INFO     Running test 2/3: check_objects
# COUT:      0.10ns INFO     Starting test: "check_objects"
# COUT:                      Description: 
# COUT:                          Check the types of objects that are returned
# COUT:                          
# COUT:      0.20ns INFO     inst_axi4s_buffer(GPI_MODULE) is HierarchyObject
# COUT:      0.20ns INFO     gen_branch_distance__0(GPI_MODULE) is HierarchyObject
# COUT:      0.20ns INFO     inst_branch_distance(GPI_MODULE) is HierarchyObject
# COUT:      0.20ns INFO     inbranch_tdata_low(GPI_REGISTER) is ModifiableObject
# COUT:      0.20ns INFO     inbranch_tdata_low(0)(GPI_REGISTER) is ModifiableObject
# COUT:      0.20ns INFO     aclk(GPI_REGISTER) is ModifiableObject
# COUT:      0.20ns INFO     s_axis_input_tdata(GPI_REGISTER) is ModifiableObject
# COUT:      0.20ns INFO     current_active(GPI_INTEGER) is IntegerObject
# COUT:      0.20ns INFO     DATA_WIDTH(GPI_INTEGER) is ConstantObject
# COUT:      0.20ns INFO     inst_ram_ctrl(GPI_MODULE) is HierarchyObject
# COUT:      0.20ns INFO     Test Passed: check_objects
# COUT:      0.20ns INFO     Running test 3/3: port_not_hierarchy
# COUT:      0.20ns INFO     Starting test: "port_not_hierarchy"
# COUT:                      Description: 
[...]

PR #1861, opened 2020-05-23
Deal with failing tests in Riviera-PRO

imphil:
Riviera-PRO 2019.10 fails some tests in various ways, see the individual commits for details.
This PR doesn't fix any of the issues, only disables tests in different ways. Individual bugs have been filed to debug issues where possible.

### DIFF
--- a/tests/test_cases/test_vhdl_access/test_vhdl_access.py
+++ b/tests/test_cases/test_vhdl_access/test_vhdl_access.py
@@ -95,7 +95,9 @@ def check_objects(dut):
         raise TestFailure("%d Failures during the test" % fails)
 
 
-@cocotb.test()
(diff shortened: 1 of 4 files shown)


tests/test_cases/test_vhdl_access/test_vhdl_access.py
```diff
@@ -95,7 +95,9 @@ def check_instance(obj, objtype):
         raise TestFailure("%d Failures during the test" % fails)
 
 
-@cocotb.test()
+# This test crashes Riviera-PRO 2019.10 (at least); skip to avoid hanging the
+# tests. See https://github.com/cocotb/cocotb/issues/1857 for details.
```
eric-wieser, line 99: Would typically abbreviate this to gh-1857, but not a big deal

Issue #1844, opened 2020-05-18
Handling errors in the GPI


ktbarrett:
The GPI has two customers, the simulator and the user (cocotb). It is necessary to provide information to the end user (you and me) and the other customer when important events occur while one customer is calling into the GPI.

When the simulator is calling into the GPI we should use the `LOG_*` macros and use `embed_sim_init` to feed critical information to cocotb. This is already done. 

However, when the user (cocotb) is calling into the GPI (e.g. `gpi_set_signal_value_int`) the most appropriate way should be to use `LOG_*` for warnings and below, while errors should feed back to the user like an exception (and would become an exception in the simulator module). I think the best way forward is to change these functions to return an optional string, `NULL` represents no error, and not `NULL` is a string with the error information.

For example,
https://github.com/cocotb/cocotb/blob/b54acce7e1879c49d18dd10ea4499b24a4a3b653/cocotb/share/lib/fli/FliObjHdl.cpp#L228-L247
Would become:
```c++
char const * FliEnumObjHdl::set_signal_value(const long value, const gpi_set_action_t action)
{
    if (action != GPI_DEPOSIT) {
       return "Force or release action not supported for FLI.";
    }

    if (value > m_num_enum || value < 0) {
       return format_str("Attempted to set a enum with range [0,%d] with invalid value %d!\n", m_num_enum, value).c_str();
    }

    if (m_is_var) {
        mti_SetVarValue(get_handle<mtiVariableIdT>(), value);
    } else {
        mti_SetSignalValue(get_handle<mtiSignalIdT>(), value);
    }

    return NULL;
}
```

eric-wieser:
Code of the form `return std::string(....).c_str()` results in a use-after-free, so that's a no go.

I'd recommend we make the C++ API throw exceptions, and then deal with converting to error codes at the C boundary.

ktbarrett:
At a minimum we should address all instances of `LOG_CRITICAL` used in functions called by cocotb.

ktbarrett:
From https://github.com/cocotb/cocotb/pull/1863#discussion_r429747349.

[...]

Issue #1859, opened 2020-05-23
Riviera: Test issue_1279 fails with  "Cleanup() called outside of the main thread"


imphil:
Riviera: Test issue_1279 fails with  "Cleanup() called outside of the main thread"

Riviera-PRO 2019.10

```
philipp@philipp-VirtualBox:~/src/cocotb$ make SIM=riviera COCOTB_REDUCED_LOG_FMT=1 -C tests/test_cases/issue_1279/ clean all
...
# COUT:      0.00ns INFO     Running on Riviera-PRO version 2019.10.138.7537
# COUT:      0.00ns INFO     Running tests with cocotb v1.4.0.dev0 from /home/philipp/.local/lib/python3.6/site-packages/cocotb
# COUT:      0.00ns INFO     Seeding Python random module with 1590270879
# COUT:      0.00ns INFO     Found test issue_1279.test_sim_failure_a
# COUT:      0.00ns INFO     Found test issue_1279.test_sim_failure_b
# COUT:      0.00ns INFO     Running test 1/2: test_sim_failure_a
# COUT:      0.00ns INFO     Starting test: "test_sim_failure_a"
# COUT:                      Description: None
# KERNEL: Simulation has finished. There are no more test vectors to simulate.
endsim
# COUT:      0.00ns ERROR    Failing test at simulator request before test run completion: Simulator shutdown prematurely
# COUT: Traceback (most recent call last):
# COUT:   File "/home/philipp/.local/lib/python3.6/site-packages/cocotb/__init__.py", line 236, in _sim_event
# COUT:     scheduler.finish_scheduler(SimFailure(msg))
# COUT:   File "/home/philipp/.local/lib/python3.6/site-packages/cocotb/scheduler.py", line 826, in finish_scheduler
# COUT:     self._test.abort(exc)
# COUT:   File "/home/philipp/.local/lib/python3.6/site-packages/cocotb/decorators.py", line 281, in abort
# COUT:     cocotb.scheduler.unschedule(self)
# COUT:   File "/home/philipp/.local/lib/python3.6/site-packages/cocotb/scheduler.py", line 493, in unschedule
# COUT:     self.cleanup()
# COUT:   File "/home/philipp/.local/lib/python3.6/site-packages/cocotb/scheduler.py", line 846, in cleanup
# COUT:     raise Exception("Cleanup() called outside of the main thread")
# COUT: Exception: Cleanup() called outside of the main thread
[...]

PR #1863, opened 2020-05-24
cleanup usage of LOG_CRITICAL in GPI and implementations

### LINKED ISSUE #1476: test_force_release regression on Questa with FLI
Introduced #1403.

```
 $ make SIM=questa COMPILE_ARGS="+acc" SIM_ARGS="-t ps" TOPLEVEL_LANG=vhdl
make results.xml
make[1]: Entering directory `.../cocotb/tests/test_cases/test_force_release'
mkdir -p sim_build
set -o pipefail; LD_LIBRARY_PATH=.../lib/python3.7/site-packages/cocotb/libs/modelsim: MODULE=test_force_release TESTCASE= TOPLEVEL="work.sample_module" COCOTB_SIM=1 \
GPI_EXTRA= TOPLEVEL_LANG=vhdl PYTHONPATH=.../lib/python3.7/site-packages/cocotb/libs/modelsim:.../cocotb/tests/test_cases/test_force_release:.../lib/python3.7/site-packages:... \
.../bin/vsim -c -64  -do sim_build/runsim.do  2>&1 | tee sim_build/sim.log
Reading pref.tcl

# 2019.1

# do sim_build/runsim.do
# QuestaSim-64 vmap 2019.1 Lib Mapping Utility 2019.01 Jan  1 2019
# vmap -c 
# Copying .../linux_x86_64/../modelsim.ini to modelsim.ini
# QuestaSim-64 vmap 2019.1 Lib Mapping Utility 2019.01 Jan  1 2019
# vmap work sim_build/work 
# Modifying modelsim.ini
# QuestaSim-64 vcom 2019.1 Compiler 2019.01 Jan  1 2019
# Start time: 16:46:50 on Mar 03,2020
# vcom -work work "+acc" .../cocotb/tests/test_cases/test_force_release/../../../tests/designs/sample_module/sample_module_pack.vhdl .../cocotb/tests/test_cases/test_force_release/../../../tests/designs/sample_module/sample_module_1.vhdl .../cocotb/tests/test_cases/test_force_release/../../../tests/designs/sample_module/sample_module.vhdl 
# -- Loading package STANDARD
# -- Loading package TEXTIO
# -- Loading package std_logic_1164
# -- Loading package NUMERIC_STD
# -- Compiling package sample_module_pack
# -- Compiling package body sample_module_pack
# -- Loading package sample_module_pack
# -- Compiling entity sample_module_1
# -- Compiling architecture impl of sample_module_1
# -- Loading package sample_module_pack
# -- Compiling entity sample_module
# -- Compiling architecture impl of sample_module
# End time: 16:46:50 on Mar 03,2020, Elapsed time: 0:00:00
# Errors: 0, Warnings: 0
# vsim -t ps -onfinish exit -foreign "cocotb_init .../lib/python3.7/site-packages/cocotb/libs/modelsim/libfli.so" sim_build/work.sample_module 
# Start time: 16:46:50 on Mar 03,2020
# ** Note: (vsim-3812) Design is being optimized...
# //  Questa Sim-64
# //  Version 2019.1 linux_x86_64 Jan  1 2019
# //
# //  Copyright 1991-2019 Mentor Graphics Corporation
# //  All Rights Reserved.
# //
# //  QuestaSim and its associated documentation contain trade
# //  secrets and commercial or financial information that are the property of
# //  Mentor Graphics Corporation and are privileged, confidential,
# //  and exempt from disclosure under the Freedom of Information Act,
# //  5 U.S.C. Section 552. Furthermore, this information
# //  is prohibited from disclosure under the Trade Secrets Act,
# //  18 U.S.C. Section 1905.
# //
# Loading std.standard
# Loading std.textio(body)
# Loading ieee.std_logic_1164(body)
# Loading ieee.numeric_std(body)
# Loading work.sample_module_pack(body)
# Loading work.sample_module(impl)#1
# Loading work.sample_module_1(impl)#1
# Loading .../lib/python3.7/site-packages/cocotb/libs/modelsim/libfli.so
#      -.--ns INFO     cocotb.gpi                         ..ib/fli/FliImpl.cpp:1077 in cocotb_init                     cocotb_init called
# 
#      -.--ns INFO     cocotb.gpi                         ../embed/gpi_embed.c:71   in set_program_name_in_venv        Did not detect Python virtual environment. Using system-wide Python interpreter
#      -.--ns INFO     cocotb.gpi                         ../gpi/GpiCommon.cpp:104  in gpi_print_registered_impl       FLI registered
#      0.00ns INFO     cocotb                                      __init__.py:144  in _initialise_testbench           Running on ModelSim for Questa-64 version 2019.1 2019.01
#      0.00ns INFO     cocotb                                      __init__.py:145  in _initialise_testbench           Python interpreter initialized and cocotb loaded!
#      0.00ns INFO     cocotb                                      __init__.py:163  in _initialise_testbench           Running tests with cocotb v1.4.0.dev0 from .../lib/python3.7/site-packages
#      0.00ns INFO     cocotb                                      __init__.py:180  in _initialise_testbench           Seeding Python random module with 1583275611
#      0.00ns INFO     cocotb.regression                         regression.py:189  in initialise                      Found test test_force_release.test_force_release
#      0.00ns INFO     cocotb.regression                         regression.py:417  in execute                         Running test 1/1: test_force_release
#      0.00ns INFO     ..test_force_release.0x7ffff1543ef0       decorators.py:253  in _advance                        Starting test: "test_force_release"
#                                                                                                                      Description: 
#                                                                                                                          Test force and release on simulation handles
#                                                                                                                          
FATAL: We are calling up again
U/I: Unexpected EOF on RPC channel sock2dc2490

** Fatal: Trouble with Simulation Kernel.
#     10.00ns CRITICAL cocotb.gpi                                FliObjHdl.cpp:328  in set_signal_value                Force or release action not supported for FLI.
make[1]: *** [results.xml] Error 218
make[1]: Leaving directory `.../cocotb/tests/test_cases/test_force_release'
make: *** [sim] Error 2
```

ktbarrett:
xref #1844. Should fix #1476.

`LOG_CRITICAL` calls `exit`. On Questa 2019 and 2020, `exit` calls the registered shutdown callback immediately. This causes GPI re-entrance from the simulator. The re-entrant call shuts down cocotb and finalizes the interpreter, but the original caller of `exit` might return through now invalid Python code. `LOG_CRITICAL` is *never* a good idea in user-called code. To remedy this, all calls to `LOG_CRITICAL` in user-called code were replaced with `LOG_ERROR` and a error code returned.

For simulator-called function that use `LOG_CRITICAL`, it should be safe. However, some simulator do not call the shutdown callback when `exit` is called, so they don't exit cleanly. Instead we should replace all instances of `LOG_CRITICAL` with `LOG_ERROR` and then call `gpi_embed_end` to cleanup cocotb gracefully.

The one outlier is the `LOG_CRITICAL`s in `cleanup_callback`s. This can be called by either the user (`deregister_callback`) or the simulator (in `handle_(vpi/vhpi/fli)_callback`). This could be solved by replacing `LOG_CRITICAL` with `LOG_ERROR` and returning an error code. If it returns to simulator control, it would *then* call `gpi_embed_end`. However, the current usage of `clenaup_callback` stomps on the return code and just returns 0, so this will require some work.

### DIFF
--- a/cocotb/share/include/gpi_logging.h
+++ b/cocotb/share/include/gpi_logging.h
@@ -53,10 +53,7 @@ enum gpi_log_levels {
 #define LOG_INFO(...)      gpi_log("cocotb.gpi", GPIInfo,          __FILE__, __func__, __LINE__, __VA_ARGS__);
 #define LOG_WARN(...)      gpi_log("cocotb.gpi", GPIWarning,       __FILE__, __func__, __LINE__, __VA_ARGS__);
 #define LOG_ERROR(...)     gpi_log("cocotb.gpi", GPIError,         __FILE__, __func__, __LINE__, __VA_ARGS__);
-#define LOG_CRITICAL(...)  do { \
-    gpi_log("cocotb.gpi", GPICritical,      __FILE__, __func__, __LINE__, __VA_ARGS__); \
-    exit(1); \
-} while (0)
+#define LOG_CRITICAL(...)  gpi_log("cocotb.gpi", GPICritical,      __FILE__, __func__, __LINE__, __VA_ARGS__);
 
 void set_log_handler(void *handler);
 void clear_log_handler(void);
--- a/cocotb/share/lib/fli/FliImpl.cpp
+++ b/cocotb/share/lib/fli/FliImpl.cpp
@@ -1016,6 +1016,8 @@ void handle_fli_callback(void *data)
 
     if (!cb_hdl) {
         LOG_CRITICAL("FLI: Callback data corrupted: ABORTING");
+        gpi_embed_end();
+        return;
     }
 
     gpi_cb_state_e old_state = cb_hdl->get_call_state();
--- a/cocotb/share/lib/vpi/VpiCbHdl.cpp
+++ b/cocotb/share/lib/vpi/VpiCbHdl.cpp
@@ -164,7 +164,8 @@ int VpiArrayObjHdl::initialise(std::string &name, std::string &fq_name) {
         }
 
         if (rangeHdl == NULL) {
-            LOG_CRITICAL("Unable to get range for indexable object");
+            LOG_ERROR("Unable to get range for indexable object");
+            return -1;
         } else {
             vpi_free_object(iter); // Need to free iterator since exited early
 
@@ -185,7 +186,8 @@ int VpiArrayObjHdl::initialise(std::string &name, std::string &fq_name) {
         check_vpi_error();
         m_range_right = val.value.integer;
     } else {
-        LOG_CRITICAL("Unable to get range for indexable object");
+        LOG_ERROR("Unable to get range for indexable object");
+        return -1;
     }
 
     /* vpiSize will return a size that is incorrect for multi-dimensional arrays so use the range
@@ -260,7 +262,8 @@ int VpiSignalObjHdl::initialise(std::string &name, std::string &fq_name) {
                         check_vpi_error();
                         m_range_right = val.value.integer;
                     } else {
-                        LOG_CRITICAL("Unable to get range for indexable object");
+                        LOG_ERROR("Unable to get range for indexable object");
+                        return -1;
                     }
                 }
                 else {
--- a/cocotb/share/lib/vpi/VpiCbHdl.cpp
+++ b/cocotb/share/lib/vpi/VpiCbHdl.cpp
@@ -96,11 +96,13 @@ int VpiCbHdl::cleanup_callback()
 
     if (m_state == GPI_PRIMED) {
         if (!m_obj_hdl) {
-            LOG_CRITICAL("VPI: passed a NULL pointer : ABORTING");
+            LOG_ERROR("VPI: passed a NULL pointer");
+            return -1;
         }
 
         if (!(vpi_remove_cb(get_handle<vpiHandle>()))) {
-            LOG_CRITICAL("VPI: unable to remove callback : ABORTING");
+            LOG_ERROR("VPI: unable to remove callback");
+            return -1;
         }
 
         check_vpi_error();
@@ -108,7 +110,8 @@ int VpiCbHdl::cleanup_callback()
 #ifndef MODELSIM
         /* This is disabled for now, causes a small leak going to put back in */
         if (!(vpi_free_object(get_handle<vpiHandle>()))) {
-            LOG_CRITICAL("VPI: unable to free handle : ABORTING");
+            LOG_ERROR("VPI: unable to free handle");
+            return -1;
         }
 #endif
     }
@@ -461,7 +464,8 @@ int VpiValueCbHdl::cleanup_callback()
     /* This is a recurring callback so just remove when
      * not wanted */
     if (!(vpi_remove_cb(get_handle<vpiHandle>()))) {
-        LOG_CRITICAL("VPI: unbale to remove callback : ABORTING");
+        LOG_ERROR("VPI: unable to remove callback");
+        return -1;
     }
 
     m_obj_hdl = NULL;
--- a/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -572,6 +572,8 @@ int32_t handle_vpi_callback(p_cb_data cb_data)
 
     if (!cb_hdl) {
         LOG_CRITICAL("VPI: Callback data corrupted: ABORTING");
+        gpi_embed_end();
+        return -1;
     }
 
     gpi_cb_state_e old_state = cb_hdl->get_call_state();
--- a/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -543,8 +543,7 @@ GpiCbHdl *VpiImpl::register_nexttime_callback()
 
 int VpiImpl::deregister_callback(GpiCbHdl *gpi_hdl)
 {
-    gpi_hdl->cleanup_callback();
-    return 0;
+    return gpi_hdl->cleanup_callback();
 }
 
 // If the Python world wants things to shut down then unregister
@@ -587,13 +586,15 @@ int32_t handle_vpi_callback(p_cb_data cb_data)
 
         /* We have re-primed in the handler */
         if (new_state != GPI_PRIMED)
-            if (cb_hdl->cleanup_callback())
+            if (cb_hdl->cleanup_callback()) {
                 delete cb_hdl;
+            }
 
     } else {
         /* Issue #188: This is a work around for a modelsim */
-        if (cb_hdl->cleanup_callback())
+        if (cb_hdl->cleanup_callback()) {
             delete cb_hdl;
+        }
     }
 
     return rv;
--- a/cocotb/share/lib/fli/FliObjHdl.cpp
+++ b/cocotb/share/lib/fli/FliObjHdl.cpp
@@ -83,7 +83,7 @@ int FliObjHdl::initialise(std::string &name, std::string &fq_name)
             m_num_elems = 1;
             break;
         default:
-            LOG_CRITICAL("Invalid object type for FliObjHdl. (%s (%s))", name.c_str(), get_type_str());
+            LOG_ERROR("Invalid object type for FliObjHdl. (%s (%s))", name.c_str(), get_type_str());
             return -1;
     }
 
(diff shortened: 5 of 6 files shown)


ktbarrett:
I don't know how far this PR should go. It depends on when the release will occur and how much risk is acceptable in bringing this in right before release. I'd like to get #1476 fixed before the release, since it is the only thing standing in the way of clean runs in 2019 and 2020. Otherwise, seg faults occur when using `Force` and `Release`.

ktbarrett:
As it stands Questa doesn't seg fault on test_force_release anymore.

cocotb/share/include/gpi_logging.h
```diff
@@ -53,10 +53,6 @@ enum gpi_log_levels {
 #define LOG_INFO(...)      gpi_log("cocotb.gpi", GPIInfo,          __FILE__, __func__, __LINE__, __VA_ARGS__);
 #define LOG_WARN(...)      gpi_log("cocotb.gpi", GPIWarning,       __FILE__, __func__, __LINE__, __VA_ARGS__);
 #define LOG_ERROR(...)     gpi_log("cocotb.gpi", GPIError,         __FILE__, __func__, __LINE__, __VA_ARGS__);
-#define LOG_CRITICAL(...)  do { \
-    gpi_log("cocotb.gpi", GPICritical,      __FILE__, __func__, __LINE__, __VA_ARGS__); \
-    exit(1); \
```
eric-wieser, line 58: Are we sure we don't just want to remove this line?

ktbarrett: No. Critical-level usually do terminate the program. It's not a good idea, but it is a widespread one. I'd rather just not have it there to reduce confusion.

* DEBUG -> verbose information useful to the developer about the program while it runs
* INFO -> information useful to the end user about the program while it runs
* WARN -> when you encounter a situation that may result in undefined or surprising behavior, or an error that you could immediately handle
* return code/exception -> when you encounter an error you can handle
* ERROR -> when you encounter an error you cannot handle
* CRITICAL -> ERROR, but contains some default way to end the program

CRITICAL assumes there is a simple parameter-less function that could be called to terminate the program (otherwise is hard to implement). In many implementations this ignores clean shutdown which is usually not a good idea (like Go). It makes ERROR moot since if you are using ERROR you have encountered an error you cannot handle, you probably should be trying to shut down.

eric-wieser: I'd argue that since:
* This is really just an interface to `logging.log` in python
* `logging.critical` does not terminate the program

That it is fine to use `critical` here.

> Critical-level logs usually do terminate the program.

I'd phrase this the other way - critical-level logs usually _immediately precede_ termination of the program. I wouldn't expect the logger to handle the termination itself.

cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -1015,7 +1015,9 @@ void handle_fli_callback(void *data)
     FliProcessCbHdl *cb_hdl = (FliProcessCbHdl*)data;
 
     if (!cb_hdl) {
-        LOG_CRITICAL("FLI: Callback data corrupted: ABORTING");
+        LOG_ERROR("FLI: Callback data corrupted: ABORTING");
+        gpi_embed_end();
```
eric-wieser, line 1019: Since this one really is a program termination, I think `critical` would be very appropriate here.

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -571,7 +570,9 @@ int32_t handle_vpi_callback(p_cb_data cb_data)
     VpiCbHdl *cb_hdl = (VpiCbHdl*)cb_data->user_data;
 
     if (!cb_hdl) {
-        LOG_CRITICAL("VPI: Callback data corrupted: ABORTING");
+        LOG_ERROR("VPI: Callback data corrupted: ABORTING");
+        gpi_embed_end();
```
eric-wieser, line 574: This one too also looks like a `critical` to me

cocotb/share/include/gpi_logging.h
ktbarrett: In several logging frameworks critical logs do terminate the program, I mentioned Go (there are others, just don't remember off the top of my head). And you are assuming that a Python logger will be the only implementation, the GPI should not tie itself to a particular language. In fact I have a change on a branch that implements a more black-boxed approach to the GPI logger and moves all the Python bits into the simulator module. I'm trying to move the GPI into a state where it could be used standalone, service multiple user, or be wrapped in another language.

You mention in certain places logs should be critical because we immediately shutdown, but I think you are missing the point. All `LOG_ERROR`s should be immediately followed by a shutdown. Where `LOG_ERROR` is used currently in user code (that doesn't immediately shut down), that is temporary until we decide on how to turn them into exceptions or error codes. There should never be a `LOG_ERROR` that isn't immediately followed by a shut down because `LOG_ERROR` means an error was detect that can't be handled; the program should not continue. In that case ERROR and CRITICAL fill them same role and one is redundant.

If we keep `LOG_CRITICAL`, `LOG_ERROR` will (should) disappear from the code. I don't really care what we call it, but we only need one.

ktbarrett:
If you are still not convinced, could you at least agree on the first commit so I can close #1476?

cocotb/share/include/gpi_logging.h
eric-wieser: > Where LOG_ERROR is used currently in user code (that doesn't immediately shut down), that is temporary until we decide on how to turn them into exceptions or error codes. 

But we use `LOG_ERROR` from the python side to mean "a test failed". So if you insist on your model, then it sounds like `LOG_ERROR` should map to `log.critical` in python?

eric-wieser:
> could you at least agree on the first commit (`LOG_CRITICAL` is definitely wrong there)

Yes, the first commit looks fine to me.

cocotb/share/include/gpi_logging.h
ktbarrett: `LOG_CRITICAL` would probably be the one to keep since it better parallels the intent. It might be a better idea to use `_sim_event` rather than log in error cases much like an exception. I think I need to think this through a bit more before confronting the issue.

I'll just follow your recommendation and make sure we are using `LOG_CRITICAL` in the proper places and do clean shutdown in those cases.

cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -96,19 +96,22 @@ int VpiCbHdl::cleanup_callback()
...
+            return -1;
         }
 
         check_vpi_error();
     } else {
 #ifndef MODELSIM
         /* This is disabled for now, causes a small leak going to put back in */
         if (!(vpi_free_object(get_handle<vpiHandle>()))) {
-            LOG_CRITICAL("VPI: unable to free handle : ABORTING");
+            LOG_ERROR("VPI: unable to free handle : ABORTING");
```
eric-wieser, line 113: These "ABORTING" messages are no longer accurate, are they?

cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -458,7 +464,7 @@ int VpiValueCbHdl::cleanup_callback()
     /* This is a recurring callback so just remove when
      * not wanted */
     if (!(vpi_remove_cb(get_handle<vpiHandle>()))) {
-        LOG_CRITICAL("VPI: unbale to remove callback : ABORTING");
+        LOG_CRITICAL("VPI: unbale to remove callback");
```
eric-wieser, line 467: Should this one either:
* end the sim
* be `ERROR` instead?

ktbarrett: Just missed that, it should be ERROR. Or at least that's what I had it as previously.

cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -458,7 +464,7 @@ int VpiValueCbHdl::cleanup_callback()
     /* This is a recurring callback so just remove when
      * not wanted */
     if (!(vpi_remove_cb(get_handle<vpiHandle>()))) {
-        LOG_CRITICAL("VPI: unbale to remove callback : ABORTING");
+        LOG_ERROR("VPI: unbale to remove callback");
     }
```
eric-wieser, line 468: Should this code path return `-1` too?

eric-wieser:
CI failure is #1859 , this should be safe to merge.

PR #1868, opened 2020-05-26
Improve Active-HDL makefile

garmin-mjames:
Cleanup/bugfix for Active-HDL:

1. Overwrite `runsim.do` instead of appending when it already exists
~~2. Always re-create `runsim.do` to account for variable changes~~

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.activehdl
+++ b/cocotb/share/makefiles/simulators/Makefile.activehdl
@@ -46,7 +46,7 @@ endif
 
 # Create a DO script (Tcl-like but not fully compatible) based on the list of $(VERILOG_SOURCES)
 $(SIM_BUILD)/runsim.do : $(VERILOG_SOURCES) $(VHDL_SOURCES) $(SIM_BUILD)
-	@echo "alib $(RTL_LIBRARY)" >> $@
+	@echo "alib $(RTL_LIBRARY)" > $@
 	@echo "set worklib $(RTL_LIBRARY)" >> $@
 ifneq ($(VHDL_SOURCES),)
 	@echo "acom $(ACOM_ARGS) $(call to_tcl_path,$(VHDL_SOURCES))" >> $@


cocotb/share/makefiles/simulators/Makefile.activehdl
```diff
@@ -44,9 +44,11 @@ else
    $(error "A valid value (verilog or vhdl) was not provided for TOPLEVEL_LANG=$(TOPLEVEL_LANG)")
 endif
 
+# Always create the DO script
+FORCE_RUNSIM:
 # Create a DO script (Tcl-like but not fully compatible) based on the list of $(VERILOG_SOURCES)
-$(SIM_BUILD)/runsim.do : $(VERILOG_SOURCES) $(VHDL_SOURCES) $(SIM_BUILD)
-	@echo "alib $(RTL_LIBRARY)" >> $@
+$(SIM_BUILD)/runsim.do : $(VERILOG_SOURCES) $(VHDL_SOURCES) $(SIM_BUILD) FORCE_RUNSIM
```
themperek, line 50: This would be then different than riviera and Questa/Modelsim. 

What is interesting this is effectively the case. I am not some expert here and maybe wrong but it looks dependency on `$(SIM_BUILD)` is causing `runsim.do`to be always re-created effectively (check with Moelsim, `sim_build` timestamp is updated). Not sure if this is intended? Maybe it should be done explicitly than? @eric-wieser

themperek:
Maybe fix 1 here and make new PR for 2?

cocotb/share/makefiles/simulators/Makefile.activehdl
garmin-mjames: Locally with Conda on Windows, `sim_build` is not re-created, and `runsim.do` is not updated. I have to delete it first if I want for instance PLUSARGS to change between runs.

themperek: In any case maybe `.PHONY: $(SIM_BUILD)/runsim.do` be better? Any negative consequences @eric-wieser ?

gnu.org/software/make/manual/html_node/Phony-Targets.html

garmin-mjames: Force target makes more sense to me, and is supported on versions of `make` that don't support `.PHONY`
https://www.gnu.org/software/make/manual/html_node/Force-Targets.html

themperek: We just have plenty of ".PHONY" other places.

garmin-mjames:
Split the Force Target commit into #1873

Issue #1494, opened 2020-03-11
[Active-HDL] VHPI callback issue in some versions prevents cocotb from working


marlonjames:
There is a bug in several versions of Aldec Active-HDL that prevent cocotb from working.

Simplified, the issue is as follows:
The VHPI callback used for cocotb's `ReadWrite` trigger is not called on delta cycles of a time step.

So this test will not work:
```python3
@cocotb.test()
async def test_signal_set(dut):
    cocotb.fork(Clock(dut.clk, 10, units='ns').start())
    await RisingEdge(dut.clk)
    dut.sig1 <= 1
    
    await ReadOnly()
    assert dut.sig1.value == 1
```
The first `ReadWrite` callback that sets the clock value will work. The `RisingEdge` callback also works, and is triggered on the next delta cycle. However, the `ReadWrite` callback where `dut.sig1` will be set is never called.

### Versions affected: ###
- Aldec Active-HDL 10.4a
- Aldec Active-HDL 10.5a
### Versions maybe affected (not explicitly tested): ###
- Aldec Active-HDL 10.4
- Aldec Active-HDL 10.5

I submitted a support ticket (**SPT78258**), and it was fixed in Active-HDL 11.1.
From 11.1 Release Notes:
```
Problems Corrected in Version 11.1
An issue with the vhpiCbRepEndOfProcesses callback from the VHPI interface was resolved. (SPT78258)
```

### Followup ###
- [x] Add warning to Active-HDL section in [Simulator Support](https://cocotb.readthedocs.io/en/latest/simulator_support.html).

cmarqu:
We could even do a check for these versions in cocotb - IIRC @ktbarrett wanted to expose getting simulator name and version from Python.

garmin-mjames:
The check is easy in Tcl:
```tcl
proc getActiveVersion {} {
    set versionList [split $version { .}]
    if {[lindex $versionList 0] == "build"} {
        set versionList [lrange $versionList 1 end]
    }
    return [join [lrange $versionList 0 1] "."]
}
```
Supports both Active-HDL GUI and VSimSA. VSimSA puts "build" at the beginning:
VSimSA:
```
$puts $version
build 10.1.3088.5434
```
Active-HDL GUI:
```
# puts $version
# 10.1.3088.5434
```

ktbarrett:
[...]

PR #1869, opened 2020-05-26
Document non-working versions of Active-HDL

garmin-mjames:
Document issue #1494.


cmarqu:
Maybe wording like "cocotb is known to not work with specific versions" is a bit more friendly, or does this sound too much like it's cocotb's fault?

imphil:
I'd generally be more specific and not only link to an issue, but good enough for now.

PR #1873, opened 2020-05-27
Active-HDL: Always create new runsim.do

garmin-mjames:
If a user changes make variable values from the command line
(i.e., PLUSARGS), the change is not reflected in the run if runsim.do
already exists.

This change causes runsim.do to be generated every time.

I have this problem on Windows in a Conda environment.

Force targets are more widely supported than `.PHONY`, but either can accomplish the same result.
https://www.gnu.org/software/make/manual/html_node/Force-Targets.html


themperek:
If we are here maybe we do it a bit better?
- compile target `$(SIM_BUILD)/runsim.do` creates (part) of `runsim.do` file and compiles design (only if  sources/dependencies change)
- run target `$(COCOTB_RESULTS_FILE)` that creates `runsim.do` always and runs the simulation

I just do not know how Acive-HDL is smart enough not co compile every time? If it is smart, all this not needed.

garmin-mjames:
The `-incr` argument to both `acom` and `alog` does incremental compilation, and won't re-compile if the contents of the file(s) did not change.

themperek:
Fine with me but @eric-wieser may know something more/unintended.

themperek:
So considering #1876 that makes all makefiles consistent.

As far as I can see options here are:
-  Add `.PHONY: sim_build/runsim.do` in user Makefile
-  Add `$(CUSTOM_COMPILE_DEPS)` dependency to $(SIM_BUILD)/runsim.do (Should be there in 1rst place) and use this to add add `FORCE_RUNSIM` user Makefile
-  Run (split) Active-HDL once to compile and the second time to run something like?

```
$(SIM_BUILD)/compile.do : $(VERILOG_SOURCES) $(VHDL_SOURCES) $(CUSTOM_COMPILE_DEPS) | $(SIM_BUILD)
    @echo "alib $(RTL_LIBRARY)"  > $@
    ....
    $(CMD) -do $(SIM_BUILD)/compile.do | tee $(SIM_BUILD)/compile.log

$(COCOTB_RESULTS_FILE): $(SIM_BUILD)/compile.do
	-@rm -f $(COCOTB_RESULTS_FILE)

	@echo "aim ..."  > : $(SIM_BUILD)/runsim.do

	set -o pipefail; MODULE=$(MODULE) TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) \
	GPI_EXTRA=$(GPI_EXTRA) TOPLEVEL_LANG=$(TOPLEVEL_LANG) \
	$(CMD) $(RUN_ARGS) -do $(SIM_BUILD)/runsim.do $(PLUSARGS) 2>&1 | tee $(SIM_BUILD)/sim.log

	$(call check_for_results_file)

```

imphil:
You've hit a problem in make: arguments passed to Make are not considered a dependency by default. We have many cases where this is problematic, the most annoying example (I find) is a change of `SIM`, which breaks in interesting ways. 

There are solutions out there which are reasonably widely used, e.g. https://cr.yp.to/redo/honest-script.html.  

I'd prefer a solution which doesn't only target Active-HDL, but eliminates the problem for all the other simulators at the same time, by properly tracking the necessary make variables as dependency.

Even though it would be nice, I don't think we need an efficient solution, but we should aim for a correct solution. So running `make clean` whenever an argument changes would be "brutal", but sufficient IMO.

PR #1881, opened 2020-05-28
XGMII Monitor: using bytearray instead of bytes

### LINKED ISSUE #1879: XGMII monitor crashes with AttributeError
`self._pkt` is initialized to `b""` (immutable), but we are using `.append()` to add payload data to it. This won't work. Maybe instead use a `bytearray()`?

_Originally posted by @LeChuck42 in https://github.com/cocotb/cocotb/pull/1545#issuecomment-635394899_

These lines exemplify the issue:

https://github.com/cocotb/cocotb/blob/924f35a3b7d39543118b7bfaed77dd4808e6612b/cocotb/monitors/xgmii.py#L107-L121

LeChuck42:
Fixes #1879.


eric-wieser:
Were you able to test this locally? Was it enough to fix the problem, or does another problem just appear?

LeChuck42:
I tested this against 1.3.1. It requires additional patches for the calls to `hexdump` since 1.3.1 still uses str there. #1519 already addresses this issue, so it *should* be ok. :)

I'll try to get a recent snapshot to test this with.

LeChuck42:
That was easier than imagined. Works without issues using the current master snapshot:
```
# COUT:  42100.40ns DEBUG    cocotb.xgmii_tx                                xgmii.py:151  in _monitor_recv                   Received:
# COUT:                                                                                                                      0000   55 55 55 55 55 55 D5 FF FF FF FF FF FF 6A D7 26   UUUUUU.......j.&
# COUT:                                                                                                                      0010   D9 79 B1 08 06 00 01 08 00 06 04 00 01 6A D7 26   .y...........j.&
# COUT:                                                                                                                      0020   D9 79 B1 C0 A8 0A 0A 00 00 00 00 00 00 C0 A8 0A   .y..............
# COUT:                                                                                                                      0030   01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
# COUT:                                                                                                                      0040   00 00 00 6C DD 33 42                              ...l.3B
# COUT:
# COUT: ###[ Ethernet ]###
# COUT:   dst       = ff:ff:ff:ff:ff:ff
# COUT:   src       = 6a:d7:26:d9:79:b1
# COUT:   type      = ARP
# COUT: ###[ ARP ]###
# COUT:      hwtype    = 0x1
# COUT:      ptype     = IPv4
# COUT:      hwlen     = 6
# COUT:      plen      = 4
# COUT:      op        = who-has
# COUT:      hwsrc     = 6a:d7:26:d9:79:b1
# COUT:      psrc      = 192.168.10.10
# COUT:      hwdst     = 00:00:00:00:00:00
# COUT:      pdst      = 192.168.10.1
# COUT: ###[ Padding ]###
# COUT:         load      = '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
# COUT:

```

PR #1882, opened 2020-05-28
Don't release handles of VHPI pseudo-region objects.

### LINKED ISSUE #1854: [Riviera-PRO] test_iteration_vhdl fails with "Not able to map type ((null)) 0 to object"
Riviera-Pro 2019.10 crashes when running the `test_iteration_vhdl` test.

```
# COUT:      0.00ns INFO     Starting test: "recursive_discovery"
# COUT:                      Description:
# COUT:                          Recursively discover every single object in the design
# COUT:
# COUT:      0.10ns WARNING  VHPI: Implementation does not know how to iterate over vhpiProcessStmtK(1082)
# COUT:      0.10ns WARNING  VHPI: Implementation does not know how to iterate over vhpiProcessStmtK(1082)
# COUT:      0.10ns ERROR    Not able to map type ((null)) 0 to object
# COUT:      0.10ns WARNING  Unable to create gen_generic_sp_ram via any registered implementation
# COUT: Stack pointer: 0x7f2e714a4b30
# COUT: Local variable address: 0x7f2e714a4f4f
# KERNEL: Fatal Error: System exception occurred: SIGSEGV. Cannot continue.
```

[Full log](https://github.com/cocotb/cocotb/files/4671974/log.txt)

I didn't debug this further yet.

garmin-mjames:
Fixes #1854 and #1857.

I introduced this bug in #1164. 
Looks like this is a regression in 1.3 for Riviera.

All I can say is mea culpa. I should have caught this with Active-HDL testing locally.

---

The handles for pseudo-regions are the handle of the parent
region object containing the `for generate` statement.

The problem occurs when iterating over the parent region's objects:
each of the `vhpiForGenerateK` objects attempts to create the
pseudo-region, relying on the `SINGLETON_HANDLES` functionality to
ensure only one exists. The `GpiHandleStore` deletes the duplicate
pseudo-region object and so the shared underlying handle is released
when it shouldn't be. When iterating continues, the parent region's
object handle is used, but it has been released.

The fix is to not release the handle when deleting the pseudo-region
object.


garmin-mjames:
@imphil I'm guessing the `recursive_discovery` test failure is now a `pass_total` mismatch rather than the simulator crashing.

ktbarrett:
Is the number of total objects

themperek:
Isues:
```
2020-05-29T03:53:32.0060434Z # COUT:      0.00ns INFO     cocotb.regression                         regression.py:457  in _start_test                     Running test 1/5: recursive_discovery
2020-05-29T03:53:32.0062470Z # COUT:      0.00ns INFO     ..ecursive_discovery.0x7f8962ad29b0       decorators.py:256  in _advance                        Starting test: "recursive_discovery"
2020-05-29T03:53:32.0065077Z # COUT:                                                                                                                      Description: 
2020-05-29T03:53:32.0067075Z # COUT:                                                                                                                          Recursively discover every single object in the design
2020-05-29T03:53:32.0068854Z # COUT:                                                                                                                          
2020-05-29T03:53:34.7942962Z # COUT:      0.10ns INFO     cocotb.test                           test_iteration.py:57   in recursive_discovery             Found a total of 27359 things
2020-05-29T03:53:44.6643777Z # COUT:      0.10ns ERROR    cocotb.regression                         regression.py:390  in _score_test                     Test Failed: recursive_discovery (result was TestFailure)
2020-05-29T03:53:44.6648247Z # COUT:                                                                                                                      Traceback (most recent call last):
2020-05-29T03:53:44.6651675Z # COUT:                                                                                                                        File "/home/az_devops_agent/agent/_work/1/s/tests/test_cases/test_iteration_vhdl/test_iteration.py", line 59, in recursive_discovery
2020-05-29T03:53:44.6656235Z # COUT:                                                                                                                          raise TestFailure("Expected %d objects but found %d" % (pass_total, total))
2020-05-29T03:53:44.6661071Z # COUT:                                                                                                                      cocotb.result.TestFailure: Expected 34569 objects but found 27359
```

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -45,9 +45,12 @@ VhpiArrayObjHdl::~VhpiArrayObjHdl()
 
 VhpiObjHdl::~VhpiObjHdl()
 {
-    LOG_DEBUG("Releasing VhpiObjHdl handle at %p\n", (void *)get_handle<vhpiHandleT>());
-    if (vhpi_release_handle(get_handle<vhpiHandleT>()))
-        check_vhpi_error();
+    /* Don't release handles for pseudo-regions, as they use the handle of the containing region */
```
eric-wieser, line 48: Is this vhpi behavior or or own behavior?

ktbarrett:
That recursive test that checks for the total number of elements has never set well with me... Those totals are too high to count out, so it was probably just gotten from observation and justified. In that case those numbers are really only valid for a particular simulator of a particular version. Perhaps we need to change it reflect that.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
garmin-mjames: The pseudo-regions are an invention of cocotb for convenience.
I added a wiki page with explanation rather than detailing it here: https://github.com/cocotb/cocotb/wiki/cocotb-Internals

tests/test_cases/test_iteration_vhdl/test_iteration.py
```diff
@@ -30,9 +30,7 @@
 from cocotb.result import TestFailure
 
 
-# This test crashes Riviera-PRO 2019.10 (at least); skip to avoid hanging the
-# tests. See issue #1854 for details.
-@cocotb.test(skip=cocotb.SIM_NAME.lower().startswith("riviera") and cocotb.SIM_VERSION.startswith("2019.10"))
+@cocotb.test()
```
ktbarrett, line 33: Take my comment as an action. The result of this test should be justified as to why they are different than the other number in this test for Riviera; then that number should be added to the test. As long as the number "makes sense", the test is really there just to prevent regression. Since we have a version of Riviera in our CI it important we actually run this test.

Perhaps we need a function to determine what the expected count is for the given simulator and version, which can double as a skip condition?

garmin-mjames: I moved the expected count to a function and added the justifications for Active-HDL differences.

I don't have access to Modelsim so I can't provide a justification for why it has a higher object count.

cocotb/__init__.py
```diff
@@ -183,8 +183,8 @@ def _initialise_testbench(argv_):
     from cocotb import simulator
 
     global SIM_NAME, SIM_VERSION
-    SIM_NAME = simulator.get_simulator_product()
-    SIM_VERSION = simulator.get_simulator_version()
+    SIM_NAME = simulator.get_simulator_product().strip()
+    SIM_VERSION = simulator.get_simulator_version().strip()
```
ktbarrett, line 187: Good catch.

tests/test_cases/test_iteration_vhdl/test_iteration.py
ktbarrett: I see you decided to keep the default expected count. I think we should be skipping the test unless we are sure we have a valid count. The function could return `0` if there is no confirmed count and you could use that in the test decorator.

```suggestion
@cocotb.test(skip=(total_object_count() == 0))
```

What do you think?

garmin-mjames:
@ktbarrett took your suggestion. I'm a little skeptical of the older Riviera numbers, though.

garmin-mjames:
Does this need a newsfragment?

ktbarrett:
@garmin-mjames I am too. I'll check them later today, but this is a good point to move forward.

ktbarrett:
If we bring it in before 1.4 no, if it slips, yes.

garmin-mjames:
@ktbarrett the bug was introduced before 1.3, so it it's present in 1.3.0 and 1.3.1

ktbarrett:
Ahh before 1.3? I guess I read that sentence wrong. Definitely will need a newsfragment.

documentation/source/newsfragments/1882.bugfix.rst
```diff
@@ -0,0 +1 @@
+Iterating over ``for generate`` statements using VHPI has been fixed. This bug caused some simulators to crash, and was a regression in 1.3.0
```
ktbarrett, line 1:
```suggestion
Iterating over ``for generate`` statements using VHPI has been fixed. This bug caused some simulators to crash, and was a regression in 1.3.0.
```

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
eric-wieser, line 48:
```suggestion
    /* Don't release handles for pseudo-regions, as they share the handle with the containing region */
```

PR #1885, opened 2020-05-30
Fix typo in scoreboard error message

regis-c:
Scoreboard use to report message as :
cocotb.result.TestFailure: Received transaction differed from expectedtransaction
with as missing space.
Just correcting adding a space.

First pull request, very basic fix, think good way to test the contribution flow.


eric-wieser:
Well spotted, this type of mistake sneaks in all the time.

themperek:
Thank you @regis-c for contribution.

PR #1890, opened 2020-05-31
Deprecate implicit conversion of handle values to strings.

### LINKED ISSUE #1865: Constant string handle __str__ returns garbage value
If a constant is a string, it gets the value with `get_signal_value_str` which returns a `bytes` object. The `__str__` implementation simply casts the underlying value using `str()` and returns that. Casting a bytes object to a string results in a garbage value: `"b'handle value'"`. This needs to be better thought out. Perhaps constant objects need to be split out into different type classes like `ModifiableObjects` are.

eric-wieser:
Note that most visibly, this will make code like the following emit warnings:
```python
log.info("The value of X is %s", dut.X)
```
unless users change it to
```python
log.info("The value of X is %s", dut.X.value)
```
Users who don't make this change will eventually see `the value of X is dut.X` in their logs.

This is a candidate fix for #1865. If we don't want to break user code doing the above, then perhaps we should close #1865 as "working as intended".

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -479,7 +479,12 @@ class ConstantObject(NonHierarchyObject):
         return self._value
 
     def __str__(self):
-        return str(self.value)
+        if isinstance(self.value, bytes):
+            StringObject._emit_str_warning(self)
+            return self.value.decode('ascii')
+        else:
+            ModifiableObject._emit_str_warning(self)
+            return str(self.value)
 
 
 class NonHierarchyIndexableObject(NonHierarchyObject):
@@ -691,7 +696,15 @@ class ModifiableObject(NonConstantObject):
     def __int__(self):
         return int(self.value)
 
+    def _emit_str_warning(self):
+        warnings.warn(
+            "`str({t})` is deprecated, and in future will return `{t}._path`. "
+            "To get a string representation of the value, use `str({t}.value)`."
+            .format(t=type(self).__qualname__),
+            FutureWarning, stacklevel=3)
+
     def __str__(self):
+        self._emit_str_warning()
         return str(self.value)
 
 
@@ -835,6 +848,17 @@ class StringObject(ModifiableObject):
     @ModifiableObject.value.getter
     def value(self) -> bytes:
         return self._handle.get_signal_val_str()
+
+    def _emit_str_warning(self):
+        warnings.warn(
+            "`str({t})` is deprecated, and in future will return `{t}._path`. "
+            "To access the `bytes` value of this handle, use `{t}.value`."
+            .format(t=type(self).__qualname__),
+            FutureWarning, stacklevel=3)
+
+    def __str__(self):
+        self._emit_str_warning()
+        return self.value.decode('ascii')
 
 
 _handle2obj = {}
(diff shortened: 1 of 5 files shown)


tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -208,7 +208,7 @@ def access_string_vhdl(dut):
     tlog = logging.getLogger("cocotb.test")
```
eric-wieser, line 208: I'm quite doubtful that this tests actually works, either before or after. In particular, line 260 looks like nonsense to me. Could someone test if for me in a vhdl simulator?

ktbarrett:
Actually how about for this release instead of changing behavior we just do what it did before while throwing the deprecation warning? When we take values, if the user gives a string, it defaults to using ASCII encoding and warning. Why couldn't we do the same thing when we give values?

This is from `StringObject`.
https://github.com/cocotb/cocotb/blob/43421561b8f856d3a11dc958eb748521810bbaed/cocotb/handle.py#L821-L826

eric-wieser:
That's reasonable

cocotb/handle.py
```diff
@@ -479,7 +479,16 @@ def value(self):
...
+        if isinstance(self.value, bytes):
+            # this would be nonsensical for bytes, so skip the deprecation
+            return super().__str__()
+        else:
+
+            warnings.warn(
+                "`str({t})` is deprecated, and in future will return `{t}._path`. "
+                "To get a string representation of the value, use `str({t}.value)`",
+                FutureWarning, stacklevel=2)
+            return str(self.value)
```
ktbarrett, line 491:
```suggestion
        warnings.warn(
            "`str({t})` is deprecated, and in future will return `{t}._path`. "
            "To get a string representation of the value, use `str({t}.value)`",
            FutureWarning, stacklevel=2)
            
        if isinstance(self.value, bytes):
            # default to ASCII as is done in StringObject
            return self.value.decode('ascii') 
        else:
            return str(self.value)
```

Per my previous suggestion.

cocotb/handle.py
```diff
@@ -836,6 +849,11 @@ def _set_value(self, value, call_sim):
     def value(self) -> bytes:
         return self._handle.get_signal_val_str()
 
+    def __str__(self):
+        # Skip the deprecated overload in ModifiableObject, which is nonsense
+        # for bytes.
+        return SimHandleBase.__str__(self)
```
ktbarrett, line 855:
```suggestion
        # default to ASCII representation as is done when setting values
        return self.value.decode('ascii')
```

eric-wieser: This should have the same deprecation warning as its parent class then.

eric-wieser: Atually, this is made redundant by your implementation above, right?

cocotb/handle.py
```diff
@@ -836,6 +849,11 @@ def _set_value(self, value, call_sim):
...
+
```
eric-wieser, line 856:
```suggestion
```

cocotb/handle.py
eric-wieser: Ah I see, this was needed after all.

tests/test_cases/test_cocotb/test_deprecated.py
```diff
@@ -43,6 +44,29 @@ async def test_unicode_handle_assignment_deprecated(dut):
     assert "bytes" in str(warns[0].message)
 
 
+@cocotb.test()
```
eric-wieser, line 47: I forgot this line in the previous revision!

PR #1891, opened 2020-05-31
Including the top-level Makefiles is outdated

imphil:
Including these makefiles doesn't work any more -- this functionality
isn't deprecated, it's outdated. Users need to take action. Point them
to the documentation to get that fixed on their side, which often
also involves installing cocotb (instead of running it directly from the
source tree).

This also removes the fragile check for cocotb-config, which was used as
a proxy for "is cocotb installed". The intention is to gently push such
users towards the documentation, since they really need to read up on
the updated installation instructions and cannot simply move along by
going from error message to error message.

### DIFF
--- a/makefiles/Makefile.inc
+++ b/makefiles/Makefile.inc
@@ -2,13 +2,7 @@
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
 
-COCOTB_CONFIG_CMD := $(shell :; command -v cocotb-config)
-ifeq ($(COCOTB_CONFIG_CMD),)
-    $(error 'cocotb-config not found! Cocotb is not installed (or is installed but not on the PATH). \
-             Please refer to https://cocotb.readthedocs.io/en/latest/quickstart.html#installing-cocotb \
-             for installation instructions.')
-endif
-
-$(error 'Deprecated cocotb file structure detected. \
-         Please refer to https://cocotb.readthedocs.io/en/latest/quickstart.html#creating-a-makefile \
-         for instructions.')
+$(error Outdated use of cocotb detected. \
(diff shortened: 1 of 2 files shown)


makefiles/Makefile.inc
```diff
@@ -2,13 +2,7 @@
...
-             for installation instructions.')
-endif
-
-$(error 'Deprecated cocotb file structure detected. \
-         Please refer to https://cocotb.readthedocs.io/en/latest/quickstart.html#creating-a-makefile \
-         for instructions.')
+$(error Outdated use of cocotb detected. \
+        Please install cocotb, and modify your makefile to \
+        "include $$(shell cocotb-config --makefile)/Makefile.sim" instead. \
+        Please refer to the documentation at https://docs.cocotb.org.)
```
eric-wieser, line 8: Are you sure we don't want to point specifically to the installation page?

eric-wieser:
```suggestion
        Please refer to the documentation at https://docs.cocotb.org/en/latest/install.html )
```

eric-wieser:
> The intention is to gently push such
users towards the documentation, since they really need to read up on
the updated installation instructions and cannot simply move along by
going from error message to error message.

I don't really agree with this. Pointing users to the documentation is rather heavy-handed when they might just need a reminder to set their PATH variable - especially if they're in the practice of setting it manually per-shell session.

makefiles/Makefile.inc
imphil: Because users need to look at two pages: the install page for how to install cocotb, and the quickstart page for how to fix their Makefile.

imphil:
Users including this Makefile are using the installation instructions from before cocotb was a package (<1.3). They typically need to do more than fixing their PATH variable -- they typically need to (a) install cocotb, and (b) adjust all their Makefiles.

PR #1895, opened 2020-05-31
Use logger.warning() instead of logger.warn()

imphil:
`.warn()` is deprecated and leads to deprecation warnings, see
https://docs.python.org/3/library/logging.html#logging.Logger.warning.


imphil:
That was quick, thanks @eric-wieser!

PR #1896, opened 2020-05-31
Collect coverage for code paths containing cocotb

imphil:
The Python coverage collection (enabled with `COVERAGE=1`)
excludes all paths containing `cocotb` in any form. For example, users
put their code in in `/home/joe/myproject/cocotb-tests`, no coverage
would be collected for their code.

Unfortunately, restricting coverage collection to "user code" doesn't
seem to be as easy as one would hope. This commit goes for a similar
workaround as before, just hopefully slightly more useful: installed
code is identified by looking for the `site-packages` path component,
and then excluded.

I'm not convinced that this exclusion is worth it, maybe we should
either find a correct solution, or don't exclude cocotb code (which is
"ugly", but at least correct).

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -115,7 +115,9 @@ class RegressionManager:
 
         if coverage is not None:
             self.log.info("Enabling coverage collection of Python code")
-            self._cov = coverage.coverage(branch=True, omit=["*cocotb*"])
+            # Exclude cocotb itself from coverage collection.
+            cocotb_package_dir = os.path.dirname(__file__)
+            self._cov = coverage.coverage(branch=True, omit=["{}/*".format(cocotb_package_dir)])
             self._cov.start()
 
         # Test Discovery


cocotb/regression.py
```diff
@@ -115,7 +115,11 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test], hooks: Iterable[Hook])
 
         if coverage is not None:
             self.log.info("Enabling coverage collection of Python code")
-            self._cov = coverage.coverage(branch=True, omit=["*cocotb*"])
+            # TODO: We want to exclude cocotb code form coverage collection and
+            # only collect coverage for user code. Using omit as shown below is
+            # a hackish workaround to exclude installed packages, but it's
+            # nowhere near fool-proof.
+            self._cov = coverage.coverage(branch=True, omit=["*site-packages*"])
```
eric-wieser, line 122: Can we be more specific to not ignore installed user packages?
```suggestion
            self._cov = coverage.coverage(branch=True, omit=["*site-packages/cocotb/*"])
```
I don't know if this needs `\` instead of `/` on windows.

imphil: Done. (I'm not sure which approach would be more useful in the end, but only stripping out cocotb is certainly what was intended before, so keeping that behavior makes sense.)

Forward slashes are also used on Windows, see https://github.com/nedbat/coveragepy/blob/225dd7cd63336b90c6e0b55fbe800a9241b3695b/coverage/files.py#L297.

cocotb/regression.py
```diff
@@ -115,7 +115,11 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test], hooks: Iterable[Hook])
 
         if coverage is not None:
             self.log.info("Enabling coverage collection of Python code")
-            self._cov = coverage.coverage(branch=True, omit=["*cocotb*"])
+            # TODO: We want to exclude cocotb code form coverage collection and
```
eric-wieser, line 118:
```suggestion
            # TODO: We want to exclude cocotb code from coverage collection and
```

eric-wieser:
Commit message seems backwards, shouldn't it be "do not collect ...", or perhaps "... cocotb which are not cocotb itself"?

imphil:
> Commit message seems backwards, shouldn't it be "do not collect ...", or perhaps "... cocotb which are not cocotb itself"?

Not really, since that's the bug that is fixed: collect coverage for user testbench code which happens to live in a directory with `cocotb` in the name. I reworded it to be less easy to get confused by.

cocotb/regression.py
```diff
@@ -115,7 +115,11 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test], hooks: Iterable[Hook])
 
         if coverage is not None:
             self.log.info("Enabling coverage collection of Python code")
-            self._cov = coverage.coverage(branch=True, omit=["*cocotb*"])
+            # TODO: We want to exclude cocotb code from coverage collection and
+            # only collect coverage for user code. Using omit as shown below is
+            # a hackish workaround to exclude the installed cocotb package, but
+            # it is nowhere near fool-proof.
+            self._cov = coverage.coverage(branch=True, omit=["*site-packages/cocotb/*"])
```
ktbarrett, line 122: This assumes the user installed cocotb typically. This won't work for my current setup.
```suggestion
            cocotb_package_dir = os.path.dirname(__file__)
            self._cov = coverage.coverage(branch=True, omit=["{}/*".format(cocotb_package_dir)])
```

imphil: Good idea! Done that. (Forward slashes are OK on Windows here, see https://github.com/cocotb/cocotb/pull/1896#discussion_r433155085)

eric-wieser:
> I guess we would have to worry about skipped tests affecting that measurement and transitioning to asserts over the `if blah: raise TestError` paradigm I see throughout.

Not really sure what you mean by that, and how it would affect coverage. I don't think we care so much about coverage of the tests themselves anyway, it's the coverage of cocotb itself that's much more interesting.

ktbarrett:
@eric-wieser Wouldn't that mean we would have to turn the condition here on it's head? Perhaps we should change `COVERAGE` to be a string of omits? If it is "1" do the current behavior.

PR #1898, opened 2020-06-02
Provide better user experience for setting log levels

### LINKED ISSUE #1888: Provide friendlier handling of invalid COCOTB_LOG_LEVEL
When running with `COCOTB_LOG_LEVEL=debug` I get the shown stack trace.

We should be able to improve the user experience a bit by:
- Converting values to ALLCAPS, i.e. debug becomes DEBUG
- Provide a more helpful error message with allowed options instead of a stacktrace.

```
        sim_build/fifo 
     -.--ns INFO     cocotb.gpi                         ..mbed/gpi_embed.cpp:245  in embed_sim_init                  Python interpreter initialized and cocotb loaded!
  File "/home/philipp/.local/lib/python3.8/site-packages/cocotb/__init__.py", line 175, in _initialise_testbench
    _setup_logging()
  File "/home/philipp/.local/lib/python3.8/site-packages/cocotb/__init__.py", line 76, in _setup_logging
    cocotb.log.default_config()
  File "/home/philipp/.local/lib/python3.8/site-packages/cocotb/log.py", line 95, in default_config
    log.setLevel(_default_log)
  File "/usr/lib64/python3.8/logging/__init__.py", line 1409, in setLevel
    self.level = _checkLevel(level)
  File "/usr/lib64/python3.8/logging/__init__.py", line 197, in _checkLevel
    raise TypeError("Level not an integer or a valid string: %r" % level)
TypeError: Level not an integer or a valid string: <function debug at 0x7f8653508430>
```

imphil:
The log level for the "cocotb" logger can be set with the environment
variable COCOTB_LOG_LEVEL. This variable is documented to accept only
uppercase log level names, and that works.

However, the checking code was slightly fragile. If the `logging` module
happened to have an attribute named like a value of COCOTB_LOG_LEVEL, the
error checking wouldn't catch it and instead a TypeError would be
thrown. See #1888 for a full stacktrace. Hitting this behavior is quite
easy by passing in `debug` instead of `DEBUG`, for example.

This patch improves this behavior by:
- Always converting log level names passed in through the environment
  variable to upper case. That's a convenience function which is
  intentionally not documented to encourage our users to do the right
  thing (TM). It's tested, however.
- Strengthening the error checking by explicitly checking for valid
  values. (Unfortunately, the `logging` class is missing an essential
  piece of functionality here to avoid code duplication.)
- Making the error message more actionable by mentioning (a) where the
  invalid value is coming from, (b) showing which value is used instead,
  and (c) giving other valid options to select from.
- Adding a test.

Fixes #1888

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -52,6 +52,9 @@ _FILENAME_CHARS = 20  # noqa
 _LINENO_CHARS   = 4  # noqa
 _FUNCNAME_CHARS = 31  # noqa
 
+# Default log level if not overwritten by the user.
+_COCOTB_LOG_LEVEL_DEFAULT = "INFO"
+
 
 def default_config():
     """ Apply the default cocotb log formatting to the root logger.
@@ -86,18 +89,25 @@ def default_config():
 
     # apply level settings for cocotb
     log = logging.getLogger('cocotb')
-    level = os.getenv("COCOTB_LOG_LEVEL", "INFO")
+
     try:
-        _default_log = getattr(logging, level)
-    except AttributeError:
-        log.error("Unable to set logging level to %r" % level)
-        _default_log = logging.INFO
-    log.setLevel(_default_log)
+        # All log levels are upper case, convert the user input for convenience.
+        level = os.environ["COCOTB_LOG_LEVEL"].upper()
+    except KeyError:
+        level = _COCOTB_LOG_LEVEL_DEFAULT
+
+    try:
+        log.setLevel(level)
+    except ValueError:
+        valid_levels = ('CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG')
+        raise ValueError("Invalid log level %r passed through the "
+                         "COCOTB_LOG_LEVEL environment variable. Valid log "
+                         "levels: %s" % (level, ', '.join(valid_levels)))
 
     # Notify GPI of log level, which it uses as an optimization to avoid
     # calling into Python.
     from cocotb import simulator
-    simulator.log_level(_default_log)
+    simulator.log_level(log.getEffectiveLevel())
 
 
 class SimBaseLog(logging.getLoggerClass()):
--- a/tests/test_cases/test_cocotb/test_logging.py
+++ b/tests/test_cases/test_cocotb/test_logging.py
@@ -4,9 +4,14 @@
 """
 Tests for the cocotb logger
 """
+
 import cocotb
+import cocotb.log
 from cocotb.triggers import Timer
+
+from common import assert_raises
 import logging
+import os
 
 
 class StrCallCounter(object):
@@ -39,3 +44,39 @@ def test_logging_with_args(dut):
     dut._log.warning("Testing multiple line\nmessage")
 
     yield Timer(100)  # Make it do something with time
+
+
+@cocotb.test()
+def test_logging_default_config(dut):
+    # The cocotb.log module is shadowed by an instance of
+    # cocotb.log.SimBaseLog()
+    from cocotb.log import default_config as log_default_config
+
+    cocotb_log = logging.getLogger('cocotb')
+
+    # Save pre-test configuration
+    log_level_prev = cocotb_log.level
+    os_environ_prev = os.environ.copy()
+
+    try:
+        # Set a valid log level
+        os.environ['COCOTB_LOG_LEVEL'] = 'DEBUG'
+        log_default_config()
+        assert cocotb_log.level == logging.DEBUG, cocotb_log.level
+
+        # Try to set log level to an invalid log level
+        os.environ['COCOTB_LOG_LEVEL'] = 'INVALID_LOG_LEVEL'
+        with assert_raises(ValueError):
+            log_default_config()
+
+        # Try to set log level to a valid log level with wrong capitalization
+        os.environ['COCOTB_LOG_LEVEL'] = 'error'
+        log_default_config()
+        assert cocotb_log.level == logging.ERROR, cocotb_log.level
+
+    finally:
+        # Restore pre-test configuration
+        os.environ = os_environ_prev
+        cocotb_log.level = log_level_prev
+
+    yield Timer(1)
--- a/documentation/source/newsfragments/1843.change.rst
+++ b/documentation/source/newsfragments/1843.change.rst
@@ -1,1 +1,1 @@
-Certain methods on the :mod:`cocotb.simulator` Python module now throw a :exc`RuntimeError` when no simulator is present, making it safe to use :mod:`cocotb` without a simulator present.
+Certain methods on the :mod:`cocotb.simulator` Python module now throw a :exc:`RuntimeError` when no simulator is present, making it safe to use :mod:`cocotb` without a simulator present.
--- a/documentation/source/newsfragments/1898.change.rst
+++ b/documentation/source/newsfragments/1898.change.rst
@@ -1,0 +1,2 @@
+Invalid values of the environment variable :envvar:`COCOTB_LOG_LEVEL` are no longer ignored.
+They now raise an exception with instructions how to fix the problem.


cocotb/log.py
```diff
@@ -86,12 +89,24 @@ def default_config():
...
+    # prevent getting any property other than the logging constants.
+    valid_levels = ('CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG')
+    if level not in valid_levels:
+        log.error("Invalid log level %r passed through the COCOTB_LOG_LEVEL "
+                  "environment variable, using %r instead. Valid log levels: %s"
+                  % (level, logging.getLevelName(_DEFAULT_LOG_LEVEL),
+                     ', '.join(valid_levels)))
+        _default_log = _DEFAULT_LOG_LEVEL
+    else:
         _default_log = getattr(logging, level)
```
eric-wieser, line 109: No point doing the work if the env var isn't there:

```suggestion
    try:
        level = os.environ["COCOTB_LOG_LEVEL"]
    except KeyError:
        _default_log = _DEFAULT_LOG_LEVEL
    else:
        # All log levels are upper case, convert the user input for convenience.
        level = level.upper()

        # Python does not provide a supported and public function to get the
        # numeric log level from a string: logging.getLevelName() could be used,
        # but is apparently unloved for this task. logging._checkLevel() is
        # private. This leaves us with getattr() as used below, but we need to
        # prevent getting any property other than the logging constants.
        valid_levels = ('CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG')
        if level not in valid_levels:
            log.error("Invalid log level %r passed through the COCOTB_LOG_LEVEL "
                      "environment variable, using %r instead. Valid log levels: %s",
                      level, logging.getLevelName(_DEFAULT_LOG_LEVEL), ', '.join(valid_levels)))
            _default_log = _DEFAULT_LOG_LEVEL
        else:
            _default_log = getattr(logging, level)
```

tests/test_cases/test_cocotb/test_logging.py
```diff
@@ -39,3 +43,37 @@ def test_logging_with_args(dut):
...
+    cocotb_log.level == logging.INFO, cocotb_log.level
+
+    # Try to set log level to a valid log level with wrong capitalization
+    os.environ['COCOTB_LOG_LEVEL'] = 'error'
+    log_default_config()
+    assert cocotb_log.level == logging.ERROR, cocotb_log.level
+
+    # Restore pre-test configuration
+    os.environ = os_environ_prev
+    cocotb_log.level = log_level_prev
```
eric-wieser, line 77:
```suggestion
    # Save pre-test configuration
    log_level_prev = cocotb_log.level
    os_environ_prev = os.environ.copy()
    try:

        # Set a valid log level
        os.environ['COCOTB_LOG_LEVEL'] = 'DEBUG'
        log_default_config()
        assert cocotb_log.level == logging.DEBUG, cocotb_log.level

        # Try to set log level to an invalid log level (should be fixed to INFO)
        os.environ['COCOTB_LOG_LEVEL'] = 'INVALID_LOG_LEVEL'
        log_default_config()
        cocotb_log.level == logging.INFO, cocotb_log.level

        # Try to set log level to a valid log level with wrong capitalization
        os.environ['COCOTB_LOG_LEVEL'] = 'error'
        log_default_config()
        assert cocotb_log.level == logging.ERROR, cocotb_log.level
    finally:
        # Restore pre-test configuration
        os.environ = os_environ_prev
        cocotb_log.level = log_level_prev
```

cocotb/log.py
```diff
@@ -86,12 +89,24 @@ def default_config():
...
-    try:
+
+    # All log levels are upper case, convert the user input for convenience.
+    level = os.getenv("COCOTB_LOG_LEVEL", logging.getLevelName(_DEFAULT_LOG_LEVEL)).upper()
+
+    # Python does not provide a supported and public function to get the
+    # numeric log level from a string: logging.getLevelName() could be used,
+    # but is apparently unloved for this task. logging._checkLevel() is
+    # private. This leaves us with getattr() as used below, but we need to
+    # prevent getting any property other than the logging constants.
```
eric-wieser, line 100: Can't we just use `log.setLevel(string_from_user)`, and let that handle the error?

eric-wieser: Namely, I think all we need is 
```python
try:
    level = os.environ['COCOTB_LOG_LEVEL'].upper()
except KeyError:
    level = logging.INFO
log.setLevel(level)  # accepts strings as well as constants
```

imphil:
Good point, `logging.setLevel()` actually does the trick (gotta love the Python documentation, which mentions the data type for that function only as a change note in the end ...). I've used that now. Slightly different implementation approach than what you suggested, but should keep the properties you're looking for (avoid checking code on our side in the "everything works" case), while also keeping the improved error messages for the user.

cocotb/log.py
```diff
@@ -86,18 +89,23 @@ def default_config():
...
-        _default_log = getattr(logging, level)
-    except AttributeError:
-        log.error("Unable to set logging level to %r" % level)
-        _default_log = logging.INFO
-    log.setLevel(_default_log)
+        log.setLevel(level)
+    except ValueError as e:
+        valid_levels = ('CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG')
+        log.error("Invalid log level %r passed through the COCOTB_LOG_LEVEL "
+                  "environment variable, using %r instead. Valid log levels: %s"
```
eric-wieser, line 101: This isn't applying the default any more.

cocotb/log.py
```diff
@@ -86,18 +89,23 @@ def default_config():
 
     # apply level settings for cocotb
     log = logging.getLogger('cocotb')
-    level = os.getenv("COCOTB_LOG_LEVEL", "INFO")
+
+    # All log levels are upper case, convert the user input for convenience.
+    level = os.getenv("COCOTB_LOG_LEVEL", logging.getLevelName(_DEFAULT_LOG_LEVEL)).upper()
```
eric-wieser, line 94: nit: Still seems a bit pointless to translate `_DEFAULT_LOG_LEVEL` to string and back, vs the only slightly longer:
```suggestion
    try:
        level = os.environ["COCOTB_LOG_LEVEL"].upper()
    except KeyError:
        level = _DEFAULT_LOG_LEVEL
```

cocotb/log.py
```diff
@@ -86,18 +89,23 @@ def default_config():
...
-        log.error("Unable to set logging level to %r" % level)
-        _default_log = logging.INFO
-    log.setLevel(_default_log)
+        log.setLevel(level)
+    except ValueError as e:
+        valid_levels = ('CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG')
+        log.error("Invalid log level %r passed through the COCOTB_LOG_LEVEL "
+                  "environment variable, using %r instead. Valid log levels: %s"
+                  % (level, logging.getLevelName(_DEFAULT_LOG_LEVEL),
+                     ', '.join(valid_levels)))
```
eric-wieser, line 103:
```suggestion
                  "environment variable, using %r instead. Valid log levels: %s",
                  level, logging.getLevelName(_DEFAULT_LOG_LEVEL),
                  ', '.join(valid_levels)))
```
No need to use `%` here, the logging module does it for you

cocotb/log.py
```diff
@@ -86,18 +89,23 @@ def default_config():
...
 
     # Notify GPI of log level, which it uses as an optimization to avoid
     # calling into Python.
     from cocotb import simulator
-    simulator.log_level(_default_log)
+    simulator.log_level(log.getEffectiveLevel())
```
eric-wieser, line 108: Sensible, thanks

cocotb/log.py
eric-wieser: Are we sure that continuing is actually a reasonable thing to do? It's more likely the user wants to tweak their environment variable and retry than it is that they want to look at their logs and work out why the `DEBG` log they thought they enable isn't showing up.

imphil:
Updated to
- Avoid str/int conversions of the log level.
- Bail out with a ValueError if the log level is invalid, as suggested by @eric-wieser. This is a change to previous behavior, but in line with documented behavior and I think it makes sense.

cocotb/log.py
imphil: I did go for slightly different code as I didn't want to have `level` be either a string or an integer depending on the try/except branch we're in.

cocotb/log.py
imphil: Good suggestion, done that in the latest push.

eric-wieser:
> * This is a change to previous behavior

Is it? The original issue gave anywhere from AttributeError to TypeError. Didn't it?

imphil:
The intended behavior (as far as I see) was to "warn" and use INFO instead, as seen here: 

https://github.com/cocotb/cocotb/blob/43421561b8f856d3a11dc958eb748521810bbaed/cocotb/log.py#L92-L94

The TypeError was only present if an invalid value happened to be an attribute of `logging`, e.g. `debug`.

cocotb/log.py
eric-wieser: Hmm, I'd say that's a feature because otherwise you convert back and forth between an int and are. The signature of setLevel is `Union[str, int]`, so keeping a local of that type seems totally reasonable.

eric-wieser:
Should probably add a newsfragment one way or another

tests/test_cases/test_cocotb/test_logging.py
```diff
@@ -39,3 +43,43 @@ def test_logging_with_args(dut):
...
+        os.environ['COCOTB_LOG_LEVEL'] = 'error'
+        log_default_config()
+        assert cocotb_log.level == logging.ERROR, cocotb_log.level
+
+    finally:
+        # Restore pre-test configuration
+        os.environ = os_environ_prev
+        cocotb_log.level = log_level_prev
+
+    yield Timer(1)
```
cmarqu, line 85:
```suggestion
    yield Timer(1, 'ns')
```
and maybe even use ``await``?

imphil: Generator-based coroutines are the norm in this file, I'd keep things consistent for the time being (until we convert the file as a whole). The same goes for a Timer without any unit; it's only doing something, the actual trigger doesn't matter at all.

tests/test_cases/test_cocotb/test_logging.py
```diff
@@ -39,3 +43,43 @@ def test_logging_with_args(dut):
...
+        assert cocotb_log.level == logging.DEBUG, cocotb_log.level
+
+        # Try to set log level to an invalid log level
+        os.environ['COCOTB_LOG_LEVEL'] = 'INVALID_LOG_LEVEL'
+        try:
+            log_default_config()
+        except ValueError:
+            pass
+        else:
+            assert 0, "Expected ValueError wasn't raised."
```
eric-wieser, line 73:
```suggestion
        with assert_raises(ValueError):
            log_default_config()
```

tests/test_cases/test_cocotb/test_logging.py
```diff
@@ -4,9 +4,13 @@
 """
 Tests for the cocotb logger
 """
+
 import cocotb
+import cocotb.log
 from cocotb.triggers import Timer
+
 import logging
+import os
```
eric-wieser, line 14:
```suggestion

from common import assert_raises
```

eric-wieser:
Ping on the comment above about using `assert_raises` - do you want me to just apply it through the github UI?

imphil:
Updated with your suggestions @eric-wieser, thanks!

Issue #1851, opened 2020-05-21
Verilator issues


elgorwi:
I’ve stumbled upon some issues while trying to get the mean example running with verilator. 
One of the issues is also described in #1313 and happens when accessing arrays.
Here’s a patch that I’ve been working on to make the array writes work for verilator:
https://github.com/cocotb/cocotb/compare/master...elgorwi:mean_verilator

It looks like there could be multiple problems:

- I’ve noticed that the Verilator VPI error ```info.code``` can be null. Not sure if this could cause issues with the logging mechanism. 
https://github.com/elgorwi/cocotb/blob/017ba6d4d58d6b3399ffe28f958571664afca906/cocotb/share/lib/vpi/VpiImpl.h#L90

- The recursionError mentioned in #1313 appears when verilator raises a VPI error. Commenting out the ```check_vpi_error()``` call in ```VpiImpl::get_sim_time()``` gets rid of the recursionError. 
https://github.com/elgorwi/cocotb/blob/017ba6d4d58d6b3399ffe28f958571664afca906/cocotb/share/lib/vpi/VpiImpl.cpp#L65
Perhaps the VPI error does not get reset by Verilator after calling vpi_get_time()? Still need to look into this in more detail.  

- Verilator exposes the ‘i_data’ array as VPI type ```vpiMemory```, where Icarus uses ```vpiNetArray```. In addition, there seems to be a problem with following: https://github.com/elgorwi/cocotb/blob/017ba6d4d58d6b3399ffe28f958571664afca906/cocotb/share/lib/vpi/VpiCbHdl.cpp#L163 Not sure why we need this exactly, but it is also mentioned in #1834, where it causes a segfault with Active-HDL. I’ve added a special case for vpiMemory as a workaround, but I don’t think this is a permanent solution.

- There are still some other issues with accessing parameters/localparams, but this is due to Verilator not exposing them. The ‘mean_discovery_test’ fails for that reason. 
Another issue is that moving the discovery test up before the other tests in test_mean.py and running everything will cause subsequent tests to fail. So something is not getting cleaned up between tests.

eric-wieser:
> I’ve noticed that the Verilator VPI error info.code can be null. Not sure if this could cause issues with the logging mechanism.

Isn't that addressed here?

https://github.com/cocotb/cocotb/blob/d5d3f78e8cd39886719ec59c0d28f76a3e88bf63/cocotb/share/lib/vpi/VpiImpl.h#L47-L48

[...]

PR #1901, opened 2020-06-03
Mean Verilator

elgorwi:
- Fix some logging issues that surfaced in #1851 . 
- Add a waiver to mean example for Verilator linting warnings.
- Add /*verilator public_flat_rd*/ to mean example parameter declarations to make them visible in Verilator.

### DIFF
--- a/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -241,7 +241,7 @@ GpiObjHdl* VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
 
     new_obj->initialise(name, fq_name);
 
-    LOG_DEBUG("VPI: Created object with type was %s(%d)",
+    LOG_DEBUG("VPI: Created GPI object from type %s(%d)",
               vpi_get_str(vpiType, new_hdl), type);
 
     return new_obj;
--- a/cocotb/share/lib/vpi/VpiImpl.h
+++ b/cocotb/share/lib/vpi/VpiImpl.h
@@ -65,9 +65,8 @@ static inline int __check_vpi_error(const char *file, const char *func, long lin
             loglevel = GPIWarning;
     }
 
-    gpi_log("cocotb.gpi", loglevel, file, func, line,
-            "VPI Error %s\nPROD %s\nCODE %s\nFILE %s",
-            info.message, info.product, info.code, info.file);
+    gpi_log("cocotb.gpi", loglevel, file, func, line, "VPI error");
+    gpi_log("cocotb.gpi", loglevel, info.file, info.product, info.line, info.message);
 
 #endif
     return level;
--- a/examples/mean/hdl/mean.sv
+++ b/examples/mean/hdl/mean.sv
@@ -7,10 +7,16 @@
 
 `timescale 1ns/1ps
 
+`ifdef VERILATOR  // make parameter readable from VPI
+  `define VL_RD /*verilator public_flat_rd*/
+`else
+  `define VL_RD
+`endif
+
 
 module mean #(
-  parameter int BUS_WIDTH = 4,
-  parameter int DATA_WIDTH = 6
+  parameter int BUS_WIDTH  `VL_RD = 4,
+  parameter int DATA_WIDTH `VL_RD = 6
 ) (
   input  logic                  clk,
   input  logic                  rst,
@@ -60,13 +66,17 @@ module mean #(
   assign o_data = s_sum >> $clog2(BUS_WIDTH);
 
 
+`ifndef VERILATOR
   initial begin
     int idx;
     $dumpfile("dump.vcd");
     $dumpvars(0, mean);
-    for (idx = 0; idx < BUS_WIDTH; idx++)
+  `ifdef __ICARUS__
+    for (idx = 0; idx < BUS_WIDTH; idx++) begin
         $dumpvars(0, i_data[idx]);
-    #1;
+    end
+  `endif
   end
+`endif
 
 endmodule
--- a/cocotb/share/lib/vpi/VpiCbHdl.cpp
+++ b/cocotb/share/lib/vpi/VpiCbHdl.cpp
@@ -713,10 +713,10 @@ VpiIterator::VpiIterator(GpiImplInterface *impl, GpiObjHdl *hdl) : GpiIterator(i
         return;
     }
 
-    LOG_DEBUG("Created iterator working from type %d %s (%s)",
-              *one2many,
+    LOG_DEBUG("Created iterator working from '%s' with type %s(%d)",
               vpi_get_str(vpiFullName, vpi_hdl),
-              vpi_get_str(vpiType, vpi_hdl));
+              vpi_get_str(vpiType, vpi_hdl),
+              type);
 
     m_iterator = iterator;
 }
--- a/examples/mean/tests/Makefile
+++ b/examples/mean/tests/Makefile
@@ -21,4 +21,8 @@ ifneq ($(filter $(SIM),ius xcelium),)
     SIM_ARGS += -v93
 endif
 
+ifeq ($(SIM),verilator)
+    EXTRA_ARGS += $(PWD)/verilator_waiver.vlt
+endif
+
 include $(shell cocotb-config --makefiles)/Makefile.sim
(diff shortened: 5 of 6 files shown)


examples/mean/hdl/mean.sv
```diff
@@ -7,10 +7,14 @@
 
 `timescale 1ns/1ps
 
+`ifdef VERILATOR
+  `define VL_RD /*verilator public_flat_rd*/
```
cmarqu, line 11: I guess this could use a comment for the user.

cocotb/share/lib/vpi/VpiImpl.h
```diff
@@ -66,8 +66,8 @@ static inline int __check_vpi_error(const char *file, const char *func, long lin
     }
 
     gpi_log("cocotb.gpi", loglevel, file, func, line,
-            "VPI Error %s\nPROD %s\nCODE %s\nFILE %s",
-            info.message, info.product, info.code, info.file);
+            "VPI Error %s\nPROD %s\nCODE %s\nFILE %s:%d",
```
cmarqu, line 69: When I saw such messages in the past, I had assumed they were produced by the simulator by the way they looked.
Is that strange format some kind of standard, or could we make it nicer?

elgorwi: This patch adds the line number to the VPI error info from the simulator. This was very helpful to track down exactly where the error was happening in Verilator while debugging VPI issues. 
The error message is printing the relevant members of the VPI error info struct:
https://github.com/cocotb/cocotb/blob/97e28803308d09d2bd885e4279d5036ef9861d42/cocotb/share/include/vpi_user.h#L767
The VPI error message above is nested in the GPI error message, which will show file, function and line number of where in the GPI layer an error occurred.  
Do you have any suggestions to improve this?

examples/mean/hdl/mean.sv
elgorwi: Done

cocotb/share/lib/vpi/VpiImpl.h
cmarqu: I understand what it says, it's just that the four letter all-uppercase "identifiers" separated by newlines look like they were formatted for machine readability instead of being for humans.
PROD we don't need at all IMO, the newlines can go, and the rest should read like a sentence. (I'm typing this on a phone, I can try to be more specific later).

eric-wieser: You could perhaps use two separate log lines, and pass a custom file/func/line into the second one

elgorwi: OK, I've updated to separate 2 log lines as suggested. 
Here's an example error log message:
```
     0.00ns ERROR    cocotb.gpi                                 VpiCbHdl.cpp:172  in initialise                      VPI error
     0.00ns ERROR    cocotb.gpi                            verilated_vpi.cpp:1615 in Verilator                       vpi_get_value: Unsupported vpiHandle (0x0)
```

cmarqu: Thanks, that looks much nicer in my opinion.

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -241,7 +241,7 @@ GpiObjHdl* VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
 
     new_obj->initialise(name, fq_name);
 
-    LOG_DEBUG("VPI: Created object with type was %s(%d)",
+    LOG_DEBUG("VPI: Created GPI object from type %s(%d)",
```
ktbarrett, line 244:
```suggestion
    LOG_DEBUG("VPI: Created GPI object of type %s(%d)",
```
?

cocotb/share/lib/vpi/VpiImpl.h
```diff
@@ -65,9 +65,8 @@ static inline int __check_vpi_error(const char *file, const char *func, long lin
             loglevel = GPIWarning;
     }
 
-    gpi_log("cocotb.gpi", loglevel, file, func, line,
-            "VPI Error %s\nPROD %s\nCODE %s\nFILE %s",
-            info.message, info.product, info.code, info.file);
+    gpi_log("cocotb.gpi", loglevel, file, func, line, "VPI error");
+    gpi_log("cocotb.gpi", loglevel, info.file, info.product, info.line, info.message);
```
ktbarrett, line 69:
```suggestion
    gpi_log("cocotb.gpi.vpi", loglevel, info.file, info.product, info.line, info.message);
```

We don't put the interface in the logger name anywhere in the C code, but there is no reason we shouldn't.

eric-wieser: What does `info.product` contain here?

eric-wieser: Did you deliberately suggest removing the first log here @ktbarrett? Without it, our line number does not escape.

cocotb/share/lib/vpi/VpiImpl.cpp
elgorwi: I don't think that would be correct as the ```type %s(%d)``` will print the VPI types, not the GPI types.

cocotb/share/lib/vpi/VpiImpl.h
elgorwi: info.product returns 'verilator', which makes it clear that the error comes from verilator. 
Looking at the example log above, I think we want to keep both log lines to give all the context.

ktbarrett: > Without it, our line number does not escape.

Ahh, good point. Nvm.

cocotb/share/lib/vpi/VpiImpl.cpp
ktbarrett: I'm not sure that's wrong, GPI is an abstraction around VPI and friends so making a GPI object of a concrete VPI type makes sense, at least to me. I'm not that worried if you want to leave it, it's a debug log, it's not really meant for an average end user.

elgorwi:
Verilator support for VPI parameters has been merged, so this example can now run correctly on  Verilator version 4.037 devel.

ktbarrett:
> Verilator support for VPI parameters has been merged, so this example can now run correctly on Verilator version 4.037 devel.

What does that mean for this PR? Is 5ba601e9b76b634ad585f63245c744ceee01e6b4 now redundant?

elgorwi:
> 
> 
> > Verilator support for VPI parameters has been merged, so this example can now run correctly on Verilator version 4.037 devel.
> 
> What does that mean for this PR? Is [5ba601e](https://github.com/cocotb/cocotb/commit/5ba601e9b76b634ad585f63245c744ceee01e6b4) now redundant?
> 
> EDIT: I guess that commit should come in anyways as long as it doesn't break newer versions of verilator. If that is the case could you add to the commit message that the fix is only necessary for supporting versions of verilator < 4.037? Just to leave breadcrumbs for our future selves.

Verilator VPI access to parameters: verilator/verilator#2370 has been merged. This means that this PR should be merged as there is no longer any blockers on the Verilator side, ie. the example will run correctly without errors on Verilator version 4.037. 
https://github.com/cocotb/cocotb/commit/5ba601e9b76b634ad585f63245c744ceee01e6b4 were some patches I used for debugging verilator issues a while ago, but everything has now been fixed on the Verilator side. I've updated https://github.com/cocotb/cocotb/commit/5ba601e9b76b634ad585f63245c744ceee01e6b4 and it is now actually pointing to this PR.

ktbarrett:
Ahh I made some faulty assumptions. So Verilator still does not support unfettered access to parameters, you must use the pragmas. Support for those pragmas wasn't added to Verilator until recently. I was thinking the pragmas were the hack to support old Verilators.

So after this PR comes in, Verilator >=v4.037 will work, but Verilator < v4.037 won't work no matter what, correct?

elgorwi:
> Ahh I made some faulty assumptions. So Verilator still does not support unfettered access to parameters, you must use the pragmas. Support for those pragmas wasn't added to Verilator until recently. I was thinking the pragmas were the hack to support old Verilators.
> 
> So after this PR comes in, Verilator >=v4.037 will work, but Verilator < v4.037 won't work no matter what, correct?

Yes, correct

eric-wieser:
> Example code isn't part of the release.

I don't agree with this claim - example code _is_ part of the release, because it's what we tell people to use in the documentation tied to the release.

ktbarrett:
The whole reason I tagged this as 1.4 was because #1941 was tagged as 1.4, so I figured we took on a stretch goal to convert the examples to async syntax for 1.4.

> Example code isn't part of the release.
All of the examples are in the sdist.

Issue #1904, opened 2020-06-04
XGMII Driver initialization incomplete


LeChuck42:
The XGMII Driver initialization is incomplete resulting in a corrupted word on the first transfer.

The `_XGMII_Bus` helper class sets it's value to XGMII IDLE during the `__init__`:
https://github.com/cocotb/cocotb/blob/97e28803308d09d2bd885e4279d5036ef9861d42/cocotb/drivers/xgmii.py#L61-L83

But this value is never read during initialization of `XGMII(Driver)` ...
https://github.com/cocotb/cocotb/blob/97e28803308d09d2bd885e4279d5036ef9861d42/cocotb/drivers/xgmii.py#L122-L140

... and as reading the value clears the internal `_integer` buffer of the helper class ...
https://github.com/cocotb/cocotb/blob/97e28803308d09d2bd885e4279d5036ef9861d42/cocotb/drivers/xgmii.py#L104-L113
 
... the first word of the first transaction (actually the PREAMBLE and START byte) is OR'ed with the IDLE value corrupting the transmission.

One way of fixing would be to simply read the value of the bus during `__init__` of `XGMII(Driver)` and thus clearing the buffer.
```
self.signal <= self.bus.value
```

Alternatively (and maybe cleaner) we could remove the default value of the helper class and use the `idle` helper of `XGMII(Driver)`.

I'll submit a pull for the latter, as I feel it's cleaner. Thoughts?

ktbarrett:
Do you have a design and tests cocotb could have to test this interface going forward?

ktbarrett:
This was resolved by #1905, right?

ecg173-bi:
Hi!, 
I think that I have this error, but I am new at this, and I am not sure about the solution. Could someone give me some more indications? 
My xgmii file is as indicated before in lines 61-83, 104-113 and 122-140 and the error is the following one: 


------ Error: ------- 
# COUT:      0.00ns ERROR    cocotb.regression                         regression.py:400  in _score_test                     Test Failed: write_address_0 (result was TypeError)
# COUT:                                                                                                                      Traceback (most recent call last):
# COUT:                                                                                                                        File "/home/.../test_gmii_rx.py", line 35, in write_address_0
# COUT:                                                                                                                          axim = XGMII(dut, dut.gmii_clk)
# COUT:                                                                                                                        File "/home/user/.local/lib/python3.7/site-packages/cocotb/drivers/xgmii.py", line 136, in __init__
[...]

PR #1905, opened 2020-06-04
XGMII Driver: Remove init value in helper class and use idle() function instead

LeChuck42:
Fix candidate for #1904.
It also clears up the use of the helper class a bit, as the values are now only set by the main `XGMII(Driver)` which already has a function to set the IDLE value.

### DIFF
--- a/cocotb/drivers/xgmii.py
+++ b/cocotb/drivers/xgmii.py
@@ -78,10 +78,6 @@ class _XGMIIBus:
         self._interleaved = interleaved
         self._nbytes = nbytes
 
-        # Default to idle
-        for i in range(nbytes):
-            self[i] = (0x07, True)
-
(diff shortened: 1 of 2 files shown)


eric-wieser:
I know it's a lot more work, but is there any chance you could write a test-bench for this?

LeChuck42:
I'll look into it. Whats the situation on the extensions by the way? Will these modules get moved at some point?

ktbarrett:
@LeChuck42 Yes we want to move them out. But before we do we need a couple things: a maintainer and a testbench. Right now all of the external interfaces are "fake" useful, as you have found out. They are untested and I don't think any of the maintainers on cocotb use them or know how they work, so they are effectively unmaintained. I don't really want to accept any changes that aren't obvious bugfixes without a testbench, somebody else might.

PR #1906, opened 2020-06-04
Use std::initializer_list in place of GpiIteratorMapping

eric-wieser:
The helper functions this type provided are subsumed by C++11 syntax in the standard library.

### DIFF
--- a/cocotb/share/lib/fli/FliImpl.cpp
+++ b/cocotb/share/lib/fli/FliImpl.cpp
@@ -576,63 +576,60 @@ GpiIterator *FliImpl::iterate_handle(GpiObjHdl *obj_hdl, gpi_iterator_sel_t type
     return new_iter;
 }
 
-void fli_mappings(GpiIteratorMapping<int, FliIterator::OneToMany> &map)
-{
-    FliIterator::OneToMany region_options[] = {
-        FliIterator::OTM_CONSTANTS,
-        FliIterator::OTM_SIGNALS,
-        FliIterator::OTM_REGIONS,
-        FliIterator::OTM_END,
-    };
-    map.add_to_options(accArchitecture, &region_options[0]);
-    map.add_to_options(accEntityVitalLevel0, &region_options[0]);
-    map.add_to_options(accArchVitalLevel0, &region_options[0]);
-    map.add_to_options(accArchVitalLevel1, &region_options[0]);
-    map.add_to_options(accBlock, &region_options[0]);
-    map.add_to_options(accCompInst, &region_options[0]);
-    map.add_to_options(accDirectInst, &region_options[0]);
-    map.add_to_options(accinlinedBlock, &region_options[0]);
-    map.add_to_options(accinlinedinnerBlock, &region_options[0]);
-    map.add_to_options(accGenerate, &region_options[0]);
-    map.add_to_options(accIfGenerate, &region_options[0]);
+decltype(FliIterator::iterate_over) FliIterator::iterate_over = []{
+    std::initializer_list<FliIterator::OneToMany> region_options;
+    std::initializer_list<FliIterator::OneToMany> signal_options;
+    std::initializer_list<FliIterator::OneToMany> variable_options;
+
+    return decltype(FliIterator::iterate_over) {
+        {accArchitecture, region_options = {
+            FliIterator::OTM_CONSTANTS,
+            FliIterator::OTM_SIGNALS,
+            FliIterator::OTM_REGIONS,
+        }},
+        {accEntityVitalLevel0, region_options},
+        {accArchVitalLevel0, region_options},
+        {accArchVitalLevel1, region_options},
+        {accBlock, region_options},
+        {accCompInst, region_options},
+        {accDirectInst, region_options},
+        {accinlinedBlock, region_options},
+        {accinlinedinnerBlock, region_options},
+        {accGenerate, region_options},
+        {accIfGenerate, region_options},
 #ifdef accElsifGenerate
-    map.add_to_options(accElsifGenerate, &region_options[0]);
+        {accElsifGenerate, region_options},
 #endif
 #ifdef accElseGenerate
-    map.add_to_options(accElseGenerate, &region_options[0]);
+        {accElseGenerate, region_options},
 #endif
 #ifdef accCaseGenerate
-    map.add_to_options(accCaseGenerate, &region_options[0]);
+        {accCaseGenerate, region_options},
 #endif
 #ifdef accCaseOTHERSGenerate
-    map.add_to_options(accCaseOTHERSGenerate, &region_options[0]);
+        {accCaseOTHERSGenerate, region_options},
 #endif
-    map.add_to_options(accForGenerate, &region_options[0]);
-    map.add_to_options(accConfiguration, &region_options[0]);
+        {accForGenerate, region_options},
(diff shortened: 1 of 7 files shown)


cocotb/share/lib/fli/FliImpl.h
```diff
@@ -434,7 +434,7 @@ class FliIterator : public GpiIterator {
     void populate_handle_list(OneToMany childType);
 
 private:
-    static GpiIteratorMapping<int, OneToMany> iterate_over;      /* Possible mappings */
+    static std::map<int, std::vector<OneToMany>> iterate_over;    /* Possible mappings */
```
ktbarrett, line 437:
```suggestion
    static std::map<int, std::vector<OneToMany>> iterate_over;   /* Possible mappings */
```
Alignment nit.

cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -645,7 +645,10 @@ FliIterator::FliIterator(GpiImplInterface *impl, GpiObjHdl *hdl) : GpiIterator(i
 
     LOG_DEBUG("fli_iterator::Create iterator for %s of type %d:%s", m_parent->get_fullname().c_str(), type, acc_fetch_type_str(type));
 
-    if (NULL == (selected = iterate_over.get_options(type))) {
+    try {
+        selected = &iterate_over.at(type);
+    }
+    catch (std::out_of_range const&) {
```
ktbarrett, line 651: I don't really think there is a reason to use exceptions here.
```suggestion
    if (iterate_over.end() == (selected = iterate_over.find(type))) {
```

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -1000,7 +983,10 @@ VhpiIterator::VhpiIterator(GpiImplInterface *impl, GpiObjHdl *hdl) : GpiIterator
     vhpiHandleT vhpi_hdl = m_parent->get_handle<vhpiHandleT>();
 
     vhpiClassKindT type = (vhpiClassKindT)vhpi_get(vhpiKindP, vhpi_hdl);
-    if (NULL == (selected = iterate_over.get_options(type))) {
+    try {
+        selected = &iterate_over.at(type);
+    }
+    catch (std::out_of_range const&) {
```
ktbarrett, line 989:
```suggestion
    if (iterate_over.end() == (selected = iterate_over.find(type))) {
```

eric-wieser:
Not sure where the travis build has gotten to...

cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -671,7 +648,10 @@ VpiIterator::VpiIterator(GpiImplInterface *impl, GpiObjHdl *hdl) : GpiIterator(i
     vpiHandle vpi_hdl = m_parent->get_handle<vpiHandle>();
 
     int type = vpi_get(vpiType, vpi_hdl);
-    if (NULL == (selected = iterate_over.get_options(type))) {
+    try {
+        selected = &iterate_over.at(type);
+    }
+    catch (std::out_of_range const&) {
```
ktbarrett, line 654:
```suggestion
    if (iterate_over.end() == (selected = iterate_over.find(type))) {
```

eric-wieser: Doesn't work because `find` returns a `pair`. I can write the more verbose version, but my feeling was that in the long run this constructor should probably throw anyway if the type doesn't exist, rather than allowing an invalid object to be constructed.

eric-wieser: The alternative is probably to add back a helper like:
```C++
template<typename K, typename V>
V* get_or_null(std::map<K, V>& m, K const& k) {
    auto it = m.find(k);
    return it == m.end() ? nullptr : &it->second;
}
```
which is still a lot lighter weight than the whole class.

ktbarrett: >`find` returns a `pair`.

What? https://en.cppreference.com/w/cpp/container/map/find. Or are you talking about std::multimap::equal_range which does return a pair?

And you are correct that we should throw eventually, but I feel that perhaps we can throw our own more descriptive exception than a standard exception.

eric-wieser: Sorry, I meant "returns an `iterator<pair<K, V>>`". The takeaway is the same though, `selected` needs to be a `K*`, like in my `get_or_null` function.

> And you are correct that we should throw eventually, but I feel that perhaps we can throw our own more descriptive exception than a standard exception.

Indeed, which is why for now I'm discarding the exception rather than re-reraising it.

ktbarrett: I guess it really should never trigger, so whatever. It's easy enough to refactor in the future if necessary.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -916,81 +916,64 @@ VhpiNextPhaseCbHdl::VhpiNextPhaseCbHdl(GpiImplInterface *impl) : GpiCbHdl(impl),
...
+        {vhpiCompInstStmtK, root_options},
+
+        {vhpiGenericDeclK, sig_options = {
+            vhpiIndexedNames,
+            vhpiSelectedNames,
+        }},
+        {vhpiSigDeclK, sig_options},
+        {vhpiSelectedNameK, sig_options},
+        {vhpiIndexedNameK, sig_options},
+        {vhpiPortDeclK, sig_options},
```
eric-wieser, line 947: This assign and reuse trick feels a little sneaky, but it's a nice way to keep the initialization and use together.

cocotb/share/lib/fli/FliImpl.h
```diff
@@ -418,7 +418,6 @@ class FliTimerCache {
 class FliIterator : public GpiIterator {
 public:
     enum OneToMany {
-        OTM_END = 0,
         OTM_CONSTANTS,  // include Generics
```
eric-wieser, line 421: Not sure if `OTM_CONSTANTS` now needs to be `1`, or if this is a true enum whose values are meaningless.

cocotb/share/lib/vpi/VpiCbHdl.cpp
ktbarrett: A good [validation](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e2-throw-an-exception-to-signal-that-a-function-cant-perform-its-assigned-task) of my heartache with this. I'm still willing to bring this in. Its a very Pythonic solution, but this is C++.

eric-wieser:
Thanks for testing the FLI. Updated with `;`s and `();`.

cocotb/share/lib/fli/FliImpl.h
ktbarrett: It doesn't cause any issue in the regression and looking at the code I don't see why it would be an issue.

ktbarrett:
@eric-wieser Could you rebase this so we can see this run with the new CI tests?

eric-wieser:
That will have the same effect, only rtd uses the unmerged head.

PR #1908, opened 2020-06-05
Add +acc to default VHDL compile options for Questa

ktbarrett:
This is required for regressions to pass and is already default for Verilog compilation.


themperek:
I am probably missing something but why we need those `ifndef V..` ? For example in `Makefile.riviera` they are not present.

ktbarrett:
I'm guessing the idea is if you use `COMPILE_ARGS` and `SIM_ARGS` those options are added onto reasonable defaults, but if you specify `VLOG_ARGS`, `VCOM_ARGS` and `VSIM_ARGS` you can override defaults. I was simply following what it does for `VLOG_ARGS` directly above.

themperek:
I guess it is OK but would not be better to clean this by occasion use `COMPILE_ARGS` and `SIM_ARGS` etc?

ktbarrett:
Are you suggesting unconditionally adding that option to `COMPILE_ARGS` or something else?

```make
COMPILE_ARGS += +acc
```

themperek:
> Are you suggesting unconditionally adding that option to `COMPILE_ARGS` or something else?
> 
> ```makefile
> COMPILE_ARGS += +acc
> ```

Yes. I think it would be more consistent. @cmarqu What you think?

ktbarrett:
@themperek the change was made.

PR #1909, opened 2020-06-05
Initial library / CI coverage support

ktbarrett:
Initial addition of cocotb library coverage support for regressions.

I started the collection in `__init__.py`, therefor there no collection in the root of that file.Ideally, we would start coverage before we imported cocotb the first time. This would probably be best living near the code that will also do custom entry point loading.

Total coverage was obtained with the following commands:
```bash
make LIBRARY_COVERAGE=1
find . -name ".library_coverage" | xargs coverage combine
coverage report  # or coverage html
```

I attached the [current coverage file for Icarus](https://github.com/cocotb/cocotb/files/4734146/coverage.zip) for your reading pleasure. Total coverage isn't great, at 63%.

### TODO
- [x] Figure out how to combine the results between CI runs?
- [ ] Move coverage start and stop farther out past the first import and last use of the cocotb module
- [x] Add badge to README

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,6 +95,8 @@ jobs:
         - export CONDA_PATH="\$env:Path = ';C:\miniconda3;C:\miniconda3\Library\mingw-w64\bin;C:\miniconda3\Library\usr\bin;C:\miniconda3\Library\bin;C:\miniconda3\Scripts;C:\miniconda3\bin;C:\miniconda3\condabin;C:\iverilog\bin;' + \$env:Path"
         - powershell "$CONDA_PATH; conda install --yes -c msys2 m2-base m2-make m2w64-toolchain libpython"
         - powershell "$CONDA_PATH; conda install --yes virtualenv"
+        # needed to make the builtin sqlite3 module (used by coverage) work - see gh-1909
+        - powershell '$CONDA_PATH; cp C:\miniconda3\Library\bin\sqlite3.dll C:\miniconda3\DLLs\'
         - powershell "$CONDA_PATH; pip install tox"
       script:
         # TODO[gh-1372]: the default compiler on windows, msvc, is not supported
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -126,6 +126,9 @@ The value passed to the Python default random number generator.
 See :envvar:`RANDOM_SEED` for details on how the value is computed.
 """
 
+_library_coverage = None
+""" used for cocotb library coverage """
+
 
 def fork(coro):
     """ Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on its use. """
@@ -154,6 +157,16 @@ def _initialise_testbench(argv_):
     comma-separated list of modules to be executed before the first test.
     """
     _rlock.acquire()
+
+    if "COCOTB_LIBRARY_COVERAGE" in os.environ:
+        import coverage
+
+        global _library_coverage
+        _library_coverage = coverage.coverage(
+            data_file=".coverage.cocotb",
+            branch=True,
+            include=["{}/*".format(os.path.dirname(__file__))])
+        _library_coverage.start()
 
     global argc, argv
     argv = argv_
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -261,6 +261,10 @@ class RegressionManager:
             self.log.info("Writing coverage data")
             self._cov.save()
             self._cov.html_report()
+        if cocotb._library_coverage is not None:
+            # TODO: move this once we have normal shutdown behavior to _sim_event
+            cocotb._library_coverage.stop()
+            cocotb._library_coverage.save()
 
         # Setup simulator finalization
         simulator.stop_simulator()
--- a/tox.ini
+++ b/tox.ini
@@ -5,6 +5,7 @@ envlist = py3
 setenv =
     CFLAGS = -Werror
     CXXFLAGS = -Werror
+    COCOTB_LIBRARY_COVERAGE = 1
 passenv =
     SIM
     TOPLEVEL_LANG
@@ -21,9 +22,11 @@ deps =
 commands =
     pytest
     make test
+    bash -c 'find . -type f -name "\.coverage\.cocotb" | xargs coverage combine --append'
 
 whitelist_externals =
     make
+    bash
 
 # needed for coverage to work
 usedevelop=True


cocotb/__init__.py
```diff
@@ -31,6 +31,20 @@
 See https://docs.cocotb.org for full documentation
 """
 import os
+
+_library_coverage = None
+if "LIBRARY_COVERAGE" in os.environ:
```
cmarqu, line 36: ``COCOTB_LIBRARY_COVERAGE``?

eric-wieser:
> Figure out how to combine the results between CI runs?

codecov will handle this if we emit our coverage data in the right place. We should add something like

https://github.com/codecov/example-python/blob/master/.travis.yml#L18-L20

cocotb/__init__.py
ktbarrett: Will do.

ktbarrett:
I've never done something like this before, so excuse me while I muddle through.

cocotb/__init__.py
```diff
@@ -31,6 +31,20 @@
...
+
+_library_coverage = None
+if "COCOTB_LIBRARY_COVERAGE" in os.environ:
+    try:
+        import coverage
+    except ImportError:
+        raise
+    else:
+        _library_coverage = coverage.coverage(
+            data_file=".library_coverage",
```
eric-wieser, line 43: This might not match the default patterns codecov looks for.

ktbarrett: Does codecov search the tree for `.coverage` files? I was planning on using `coverage combine`.

Looks like `codecov` supports the `-f` option, so the name doesn't matter and we can skip merging.

.travis.yml
```diff
@@ -45,12 +45,14 @@ jobs:
...
         - cd iverilog && autoconf && ./configure && make -j2 && sudo make install && cd ..
-        - pip install tox
+        - pip install tox codecov
         - pyenv global system $TRAVIS_PYTHON_VERSION
       script:
         - echo 'Running cocotb tests with Python 3.7 ...' && echo -en 'travis_fold:start:cocotb_py37'
         - tox -e py37
         - echo 'travis_fold:end:cocotb_py37'
+      after_script:
+        - find . -name ".library_coverage" | xargs codecov -f
```
eric-wieser, line 55:
```suggestion
        - find . -name ".library_coverage" | xargs codecov -f
        # Push the results back to codecov
        after_success:
          - codecov
```

eric-wieser:
I've found Travis doesn't reload the config after force pushes

eric-wieser:
https://codecov.io/github/cocotb/cocotb/commit/9923f02861a3c5d5ade1aeadbe5886a20b7cdc75

ktbarrett:
@eric-wieser Is that why the Mac build is failing?

eric-wieser:
```
Users\travis\build\cocotb\cocotb\.tox\py3_mingw\lib\site-packages\cocotb\__init__.py", line 38, in <module>
    import coverage
  File "C:\Users\travis\build\cocotb\cocotb\.tox\py3_mingw\lib\site-packages\coverage\__init__.py", line 15, in <module>
    from coverage.control import Coverage, process_startup
  File "C:\Users\travis\build\cocotb\cocotb\.tox\py3_mingw\lib\site-packages\coverage\control.py", line 20, in <module>
    from coverage.data import CoverageData, combine_parallel_data
  File "C:\Users\travis\build\cocotb\cocotb\.tox\py3_mingw\lib\site-packages\coverage\data.py", line 17, in <module>
    from coverage.sqldata import CoverageData
  File "C:\Users\travis\build\cocotb\cocotb\.tox\py3_mingw\lib\site-packages\coverage\sqldata.py", line 15, in <module>
    import sqlite3
  File "c:\miniconda3\Lib\sqlite3\__init__.py", line 23, in <module>
    from sqlite3.dbapi2 import *
  File "c:\miniconda3\Lib\sqlite3\dbapi2.py", line 27, in <module>
    from _sqlite3 import *
ImportError: DLL load failed: The specified module could not be found.
```

ktbarrett:
That problem seems out of our control. Should I go back to passing the `COCOTB_LIBRARY_COVERAGE` in and not do so for Windows?

eric-wieser:
This slooks like something we can fix - we seem to be mixing two different python installations, based on the paths in that traceback.

eric-wieser:
Or perhaps our path is the issue: https://github.com/jupyter/notebook/issues/4332#issuecomment-459168997

ktbarrett:
@eric-wieser I don't think it's our path. The path talked about in that issue is already in our path.
https://github.com/cocotb/cocotb/blob/49c71718d0a6696052a9a80afaa50abace84e591/.travis.yml#L82

eric-wieser:
Hmm, you're right

.travis.yml
```diff
@@ -52,6 +52,7 @@ jobs:
         - tox -e py37
         - echo 'travis_fold:end:cocotb_py37'
       after_success: &travis_linux_post
+        - find . -name ".library_coverage" | xargs coverage combine .coverage
```
ktbarrett, line 55: This overwrites the output of the pytest run. pytest should deposit it's results to `.library_coverage`.

ktbarrett:
Well I give on the Windows issue. I am willing to disable coverage collection on Windows and move forward.

Also, I will definitely be skipping checkbox 2 this PR. That code will go well next to custom entry point code.

tox.ini
```diff
@@ -20,9 +21,11 @@ deps =
 commands =
     pytest
     make test
+    bash -c 'find . -type f -name "\.coverage\.cocotb" | xargs coverage combine --append'
```
eric-wieser, line 24: I don't think this is even necessary for codecov in CI, but maybe it's helpful locally.

cocotb/__init__.py
```diff
@@ -155,6 +158,19 @@ def _initialise_testbench(argv_):
     """
     _rlock.acquire()
 
+    global _library_coverage
+    if "COCOTB_LIBRARY_COVERAGE" in os.environ:
+        try:
+            import coverage
+        except ImportError:
+            raise
```
eric-wieser, line 166: This try except is a no-op, you may as well just leave the import to raise.

ktbarrett:
Dear god... So after some investigation, I found that I can fix the Windows issue by moving sqlite3.dll from `miniconda_root\Library\bin` to `miniconda_root\DLLs`. I'm not sure who's issue that is, but it certainly isn't ours. I think it only manifests when running an embedded Python interpreter. Perhaps Conda's Python is setting things when entered as an application that doesn't occur when being loaded as an embedded interpreter? I'm just making wild guesses though.

.travis.yml
```diff
@@ -95,6 +95,7 @@ jobs:
         - export CONDA_PATH="\$env:Path = ';C:\miniconda3;C:\miniconda3\Library\mingw-w64\bin;C:\miniconda3\Library\usr\bin;C:\miniconda3\Library\bin;C:\miniconda3\Scripts;C:\miniconda3\bin;C:\miniconda3\condabin;C:\iverilog\bin;' + \$env:Path"
         - powershell "$CONDA_PATH; conda install --yes -c msys2 m2-base m2-make m2w64-toolchain libpython"
         - powershell "$CONDA_PATH; conda install --yes virtualenv"
+        - powershell '$CONDA_PATH; cp C:\miniconda3\Library\bin\sqlite3.dll C:\miniconda3\DLLs\'
```
eric-wieser, line 98: I think this needs a todo comment pointing at this PR. I'm happy to commit the hack in the interest of having coverage, but lets leave a breadcrumb in case someone works out how to remove it.

```suggestion
        # needed to make the builtin sqlite3 module (used by coverage) work - see gh-1909
        - powershell '$CONDA_PATH; cp C:\miniconda3\Library\bin\sqlite3.dll C:\miniconda3\DLLs\'
```

ktbarrett: Because we run with coverage on in tox, this change will effectively prevent any Windows user from running regressions. I'll move it into tox.ini (along with the message) first.

cocotb/regression.py
```diff
@@ -261,6 +261,10 @@ def tear_down(self) -> None:
             self.log.info("Writing coverage data")
             self._cov.save()
             self._cov.html_report()
+        if cocotb._library_coverage is not None:
+            # move this once we have normal shutdown behavior to _sim_event
```
eric-wieser, line 265: nit:
```suggestion
            # TODO: move this once we have normal shutdown behavior to _sim_event
```
otherwise it looks like a description of what this line is doing, not of what needs to be done

eric-wieser:
Feel free to merge once this is passing

PR #1912, opened 2020-06-06
Collect coverage information for the pytest tests.

eric-wieser:
This does not attempt to collect coverage for the simulator tests, as gh-1909 handles that.

@ktbarrett, since your PR is running into some weird python environment issues, maybe we should merge this first?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,12 +45,14 @@ jobs:
         - sudo apt-get -y install gperf swig
         - if [[ ! -e "./iverilog/README.txt" ]]; then rm -rf iverilog; git clone https://github.com/steveicarus/iverilog.git --depth=1 --branch v10_3; fi
         - cd iverilog && autoconf && ./configure && make -j2 && sudo make install && cd ..
-        - pip install --upgrade tox virtualenv
+        - pip install --upgrade tox virtualenv codecov
         - pyenv global system $TRAVIS_PYTHON_VERSION
       script:
         - echo 'Running cocotb tests with Python 3.7 ...' && echo -en 'travis_fold:start:cocotb_py37'
         - tox -e py37
         - echo 'travis_fold:end:cocotb_py37'
+      after_success: &travis_linux_post
+        - codecov
 
     - stage: SimTests
       python: 3.8
@@ -59,6 +61,7 @@ jobs:
         - echo 'Running cocotb tests with Python 3.8 ...' && echo -en 'travis_fold:start:cocotb_py38'
         - tox -e py38
         - echo 'travis_fold:end:cocotb_py38'
+      after_success: *travis_linux_post
 
     - stage: SimTests
       python: 3.6.7
@@ -67,6 +70,7 @@ jobs:
         - echo 'Running cocotb tests with Python 3.6.7 ...' && echo -en 'travis_fold:start:cocotb_py36'
         - tox -e py36
         - echo 'travis_fold:end:cocotb_py36'
+      after_success: *travis_linux_post
 
     - stage: SimTests
       python: 3.7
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -23,6 +24,9 @@ commands =
 whitelist_externals =
     make
 
+# needed for coverage to work
+usedevelop=True
```
ktbarrett, line 28: Could you explain this a bit better to me?

eric-wieser:
@imphil, for this to integrate well we need to install the codecov app into the org, which I do not have permission to do. You should have received a request.

tox.ini
eric-wieser: `--cov=cocotb` matches against `./cocotb`. If we are not using an `-e` install (which this forces), then no coverage is collected for those files.

This seems to be how I've seen the problem handled elsewhere.

README.md
```diff
@@ -4,6 +4,7 @@
 [![Build Status](https://travis-ci.org/cocotb/cocotb.svg?branch=master)](https://travis-ci.org/cocotb/cocotb)
 [![PyPI](https://img.shields.io/pypi/dm/cocotb.svg?label=PyPI%20downloads)](https://pypi.org/project/cocotb/)
 [![Gitpod Ready-to-Code](https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/cocotb/cocotb)
+[![codecov](https://codecov.io/gh/cocotb/cocotb/branch/master/graph/badge.svg)](https://codecov.io/gh/cocotb/cocotb)
```
cmarqu, line 7: I would move this up to e.g. after "Build Status".

eric-wieser: Let's leave that until our number is higher than 30%!

.travis.yml
```diff
@@ -45,12 +45,14 @@ jobs:
         - sudo apt-get -y install gperf swig
         - if [[ ! -e "./iverilog/README.txt" ]]; then rm -rf iverilog; git clone https://github.com/steveicarus/iverilog.git --depth=1 --branch v10_3; fi
         - cd iverilog && autoconf && ./configure && make -j2 && sudo make install && cd ..
-        - pip install tox
+        - pip install tox codecov virtualenv --upgrade
```
imphil, line 48: Nit: Can you put the `--upgrade` flag right after `pip install`?

eric-wieser: Done

eric-wieser:
> Please document steps how to create coverage locally (if you wish to in a follow-up)

We do not even have documentation on how to run the `pytest` tests, so I'd prefer to leave this to a follow-up. The simple answer is "we use `tox`".

Do we even have a document for on-boarding contributors that we should update in future?

imphil:
We're lacking a lot of documentation in that space, but shouldn't use it as an excuse for not dumping a couple sentences somewhere, no matter how unpolished or unframed they might be.

Private CI is actually failing with `../../designs/sample_module/Makefile:50: /Makefile.sim: No such file or directory`, so probably `cocotb-config` isn't found by the Makefiles.

eric-wieser:
> so probably `cocotb-config` isn't found by the Makefiles.

What's happening is that `cocotb-config` is throwing a `DeprecationWarning`, I think. You may need to make a change like #1915 to the private CI.

eric-wieser:
Rebased, and added a very brief section to the contribution guide on how to run tests locally.

eric-wieser:
Private CI still needs a `pip install --update virtualenv`to be run manually on the machines. @imphil, are you fine to merge without that, and handle it later, or do you wait just in case?

imphil:
Private CI is fixed now, feel free to merge at your convenience.

PR #1917, opened 2020-06-09
Do not issue deprecation warnings in cocotb-config

imphil:
cocotb-config is a command line tool, the output of which is consumed by 
scripts (mostly makefiles in our case). Having any kind of output in these
tools which isn't controlled by us breaks the users of
`cocotb-config` and must be avoided.

This issue was triggered by virtualenv outputting a DeprecationWarning; it
is loaded by default in our tox testing environments.

A deprecation warning shouldn't break all users of cocotb.


imphil:
Reverting #1915 for now to test if that change actually works in CI, since I haven't been able to reproduce it locally.

eric-wieser:
> A deprecation warning shouldn't break all users of cocotb.

It doesn't. The only reason you're seeing this is because we configured CI to crash on all warnings, so that we knew about them.

cocotb/config.py
```diff
@@ -38,6 +38,15 @@
...
 """
+
+# The output (on STDOUT) of cocotb-config is typically assigned to variables in
+# our makefiles. All additional output breaks this use case and must be avoided
+# if possible. DeprecationWarnings by third-party libraries are not something we
+# can control; not showing those avoids breaking cocotb-config for our users as
+# they make changes to the list of installed Python packages.
+import warnings
+warnings.filterwarnings("ignore", category=DeprecationWarning)
+
```
eric-wieser, line 49: I find it hard to believe this makes any difference. The error in private CI only starts appearing when we change to doing `pip install -e`.

The warning is being emitted by the virtualenv code that launches `cocotb.config`, so is beyond our control.

eric-wieser: I'll see if I can show you exactly what that wrapper code looks like...

eric-wieser: Nope can't find it. But if you look at the tracebacks that I mentioned before, not one line referenced any of our files.

imphil:
Ah, with `use_develop` I can actually reproduce the problem.

Thanks for your hint to https://github.com/cocotb/cocotb/blob/dd88c00c9fa82d94d498c0c89b9176edb8990d67/tests/Makefile#L29

And yes, that's not going to work. 

Looking at the current `PYTHONWARNINGS` we already starting going down the route of disabling DeprecationWarning's for virtualenv-related issues, so let's see if extending that idea does the trick. 

I'm not sure which fix will be better in the end (require an updated virtualenv, or disable the warning) or if we should take both as a "defense in depth", let's see if we can actually make it work first.

imphil:
Test 1 successful: I can reproduce the error with the temporary commits in this PR.

imphil:
Test 2 successful as well: The code in here actually "fixes" the problem.

imphil:
OK @eric-wieser the patch in this PR now actually works. I'm not sure if is the right thing to do, however. 

Since we already have code that apparently deals with virtualenv DeprecationWarnings, it feels right to add the current source of that warning to it as well.

I also think we should keep #1915, and do the same for our Private CI runner.

WDYT?

tests/Makefile
```diff
@@ -25,8 +25,9 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
-# distutils warnings are caused by an old version of virtualenv in travis
-export PYTHONWARNINGS = error,ignore::DeprecationWarning:distutils
+# Do not fail on DeprecationWarning caused by virtualenv, which might come from
+# the distutils or site modules.
+export PYTHONWARNINGS = error,ignore::DeprecationWarning:distutils,ignore::DeprecationWarning:site
```
eric-wieser, line 30: To be clear: I'm fine either with this change or with removing the existing filter.

PR #1918, opened 2020-06-09
Put Python development dependencies into a separate file

imphil:
Provide a single file with all "development dependencies", and use this file in Travis.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,0 +1,8 @@
+# Python packages which are required/recommended to develop cocotb
+
+tox
+
+# virtualenv is used by tox; versions below 16.1.0 cause a DeprecationWarning
+# to be shown for all code which uses virtualenv, which is the case for all code
+# we run through tox. (https://github.com/pypa/virtualenv/pull/1064)
+virtualenv >= 16.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,8 @@ jobs:
         - sudo apt-get -y install gperf swig
         - if [[ ! -e "./iverilog/README.txt" ]]; then rm -rf iverilog; git clone https://github.com/steveicarus/iverilog.git --depth=1 --branch v10_3; fi
         - cd iverilog && autoconf && ./configure && make -j2 && sudo make install && cd ..
-        - pip install --upgrade tox virtualenv codecov
+        - pip install -r dev-requirements.txt
+        - pip install codecov
         - pyenv global system $TRAVIS_PYTHON_VERSION
       script:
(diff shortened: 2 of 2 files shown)


dev-requirements.txt
```diff
@@ -0,0 +1,11 @@
...
+
+tox
+
+# virtualenv is used by tox; versions below 16.1.0 cause a DeprecationWarning
+# to be shown for all code which uses virtualenv, which is the case for all code
+# we run through tox. (https://github.com/pypa/virtualenv/pull/1064)
+virtualenv >= 16.1
+
+# Requirements to build the documentation
+-r documentation/requirements.txt
```
imphil, line 11: I'm not sure about this one. I typically build the documentation with the provided Makefile, which installs the dependencies into a venv; installing them here again doesn't help.

At the same time, I know that others don't use the Makefile -- @eric-wieser @cmarqu @themperek how do you build documentation locally?

cmarqu: I do use the makefile.

eric-wieser: This will just result in dependencies being installed twice in CI - once in the virtualenv, and once outside the virtualenv.

eric-wieser:
If we're going to do this at all, I recommend the file be called `tox-requirements.txt`, and have a comment indicating that this is for the requirements of `tox` itself, not the requirements that are installed _by_ tox.

dev-requirements.txt
eric-wieser: Since we already use tox to manage our other dependencies, it might make sense to retire the makefile and use tox for sphinx too: https://tox.readthedocs.io/en/latest/example/documentation.html#sphinx

ktbarrett:
What are "dev requirements"? There are build requirements, installation requirements, runtime requirements, testing requirements, optional requirements for each of those (like cherrypy and dowser for memory debugging), etc. Is dev requirements just the union of them? Is there any way to include in this file that you need doxygen for documentation as well?

Either way I'm fine with the way this is right now because it automates and centralizes those dependencies; but this doesn't obviate the need for documentation.

dev-requirements.txt
```diff
@@ -0,0 +1,8 @@
+# Python packages which are required/recommended to develop cocotb
+
+tox
+
+# virtualenv is used by tox; versions below 16.1.0 cause a DeprecationWarning
+# to be shown for all code which uses virtualenv, which is the case for all code
+# we run through tox. (https://github.com/pypa/virtualenv/pull/1064)
+virtualenv >= 16.1
```
eric-wieser, line 8: This is now managed by `tox.ini`: https://github.com/cocotb/cocotb/blob/master/tox.ini#L5-L8

imphil:
We moved more of this functionality into tox, no more need for this PR.

PR #1924, opened 2020-06-10
Document TOPLEVEL_LANG

### LINKED ISSUE #1847: TOPLEVEL_LANG is not documented
I can imagine this frustrates a lot of people.

ktbarrett:
Fixes #1847. Document `TOPLEVEL_LANG` in the makefile variable reference section of the documentation.

### DIFF
--- a/documentation/source/building.rst
+++ b/documentation/source/building.rst
@@ -221,6 +221,14 @@ Makefile-based Test Scripts
       Set this to 1 to enable wave traces dump for the Aldec Riviera-PRO and Mentor Graphics Questa simulators.
       To get wave traces in Icarus Verilog see :ref:`sim-icarus-waveforms`.
 
+.. make:var:: TOPLEVEL_LANG
+
+    Used to inform the makefile scripts which HDL language the top-level design element is written in.
+    Currently it supports the values ``verilog`` for Verilog or SystemVerilog tops, and ``vhdl`` for VHDL tops.
+    This is used by simulators that support more than one interface (VPI, VHPI, or FLI) to select the appropriate interface to start cocotb.
+
+    The variable is also made available to cocotb tests conveniently as :data:`cocotb.LANGUAGE`.
+
 .. make:var:: VERILOG_SOURCES
 
       A list of the Verilog source files to include.


documentation/source/building.rst
```diff
@@ -301,6 +301,16 @@ Makefile-based Test Scripts
     You can usually write out such a file from the simulator's GUI.
     This is currently supported for the Mentor Questa, Mentor ModelSim and Aldec Riviera simulators.
 
+.. make:var:: TOPLEVEL_LANG
+
+    Used to inform the makefile scripts what HDL language of the toplevel design element is written in.
```
eric-wieser, line 306:
```suggestion
    Used to inform the makefile scripts which HDL language the top-level design element is written in.
```

documentation/source/building.rst
```diff
@@ -301,6 +301,16 @@ Makefile-based Test Scripts
...
+    Currently it supports the values ``verilog`` for Verilog or SystemVerilog tops, and ``vhdl`` for VHDL tops.
+    This is used by simulators that support more than one interface (VPI, VHPI, or FLI) to select the appropriate interface to start cocotb.
+    It is also commonly used to deter execution of a cocotb test suite if the tests or DUT do not support being run with the selected :make:var:`TOPLEVEL_LANG`.
```
eric-wieser, line 309: This doesn't seem like user information - is anything other than cocotb itself going to have two copies of their code, one vhdl and one verilog?

documentation/source/building.rst
```diff
@@ -301,6 +301,16 @@ Makefile-based Test Scripts
...
+
+    The variable is also made available to cocotb tests conveniently at :data:`cocotb.LANGUAGE`.
```
cmarqu, line 311: s/at/as/?

documentation/source/building.rst
ktbarrett: Potentially? Really the first few sentences are the pertinent information and this just follows. I'm not sure where else to document how to run the regression with different langs. 

Speaking of which, I just realized we don't have TOPLEVEL_LANG exempted in tox.

eric-wieser: That probably should go in the contribution guide, for how to run cocotb's own unit tests. I think only the first three sentences are useful here.

I suppose you could put the remaining sentences in an `.. info` box, with a comment that this is something that cocotb does in its own test-suite, not something that users actually ought to do.

ktbarrett: I'll add a section to the CONTRIBUTING.md about it as well as `SIM` being supported by `tox` once the `TOPLEVEL_LANG` exemption is added to tox.ini.

documentation/source/building.rst
```diff
@@ -301,6 +301,15 @@ Makefile-based Test Scripts
...
 
+.. make:var:: TOPLEVEL_LANG
+
+    Used to inform the makefile scripts which HDL language the top-level design element is written in.
+    Currently it supports the values ``verilog`` for Verilog or SystemVerilog tops, and ``vhdl`` for VHDL tops.
+    This is used by simulators that support more than one interface (VPI, VHPI, or FLI) to select the appropriate interface to start cocotb.
+
+    The variable is also made available to cocotb tests conveniently as :data:`cocotb.LANGUAGE`.
+    This can be used to prevent individual tests from running with the selected :make:var:`TOPLEVEL_LANG`.
+
```
eric-wieser, line 312: Should this be right next to `TOP_LEVEL`? This isn't optional, like all the other variables in this section.

eric-wieser: Or perhaps vice versa...

documentation/source/building.rst
```diff
@@ -301,6 +301,15 @@ Makefile-based Test Scripts
...
     This is currently supported for the Mentor Questa, Mentor ModelSim and Aldec Riviera simulators.
 
+.. make:var:: TOPLEVEL_LANG
+
+    Used to inform the makefile scripts which HDL language the top-level design element is written in.
+    Currently it supports the values ``verilog`` for Verilog or SystemVerilog tops, and ``vhdl`` for VHDL tops.
+    This is used by simulators that support more than one interface (VPI, VHPI, or FLI) to select the appropriate interface to start cocotb.
+
+    The variable is also made available to cocotb tests conveniently as :data:`cocotb.LANGUAGE`.
+    This can be used to prevent individual tests from running with the selected :make:var:`TOPLEVEL_LANG`.
```
eric-wieser, line 311: Perhaps move this sentence to the docs for `cocotb.LANGUAGE`, maybe as:
```
.. info::
    cocotb's own regression tests sometimes run the same tests against both a
    Verilog and VHDL implementation of the same module, where this variable
    can be used to skip tests that only make sense in one of the two languages.
```

documentation/source/building.rst
ktbarrett: Moved to right before `VERILOG_SOURCES`.

documentation/source/building.rst
ktbarrett: I removed the sentence, but I'm hesitant to document the use of `LANGUAGE` because I'm not sure if we should be doing that *at all*. Really tests should run agnostic to the implementation language, that's one of the main points of the GPI. If this is necessary we should be segregating the tests into different suite where in the makefiles we decide whether to skip the whole suite or not.

PR #1925, opened 2020-06-11
Add TOPLEVEL_LANG to tox

ktbarrett:
Add `TOPLEVEL_LANG` passthrough to tox and add it to GHDL CI run where it is currently missing (that might explain some of the failures).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,7 +105,7 @@ jobs:
 
     - stage: SimTests
       python: 3.7
-      env: SIM=ghdl
+      env: SIM=ghdl TOPLEVEL_LANG=vhdl
       before_install:
         - sudo apt-get install gnat
         - git clone https://github.com/ghdl/ghdl.git --depth=1 --branch v0.36
@@ -132,4 +132,4 @@ jobs:
   allow_failures:
     - stage: SimTests
       python: 3.7
-      env: SIM=ghdl
+      env: SIM=ghdl TOPLEVEL_LANG=vhdl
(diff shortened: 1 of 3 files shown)


eric-wieser:
Sounds like an explosion of configuration options to me. I don't think we need separate tox environments unless the installation process is different between them.

ktbarrett:
I was under the impression that tox supported automagically defining such targets by combining fragments of testenvs? Either way, such a discussion shouldn't block this coming in. Is the documentation I added good enough for now?

ktbarrett:
Isn't the GHDL test exempt from failing? Not sure why it's reporting the Travis run as failed.

.travis.yml
```diff
@@ -105,7 +105,7 @@ jobs:
 
     - stage: SimTests
       python: 3.7
-      env: SIM=ghdl
+      env: SIM=ghdl TOPLEVEL_LANG=vhdl
```
eric-wieser, line 108: This changes the test ID, so now it's not marked xfail any more

ktbarrett: Whoops. Will fix.

eric-wieser:
Feel free to merge once you sort the GHDL CI.

Would you mind checking if this actually helps the GHDL tests at all?

ktbarrett:
Previously it the GHDL CI run skipped most of the tests since `TOPLEVEL_LANG` was wrong, so now they are actually running. Still *lots* of issues.

PR #1928, opened 2020-06-12
Add Python 3.9 to Linux Travis CI

### LINKED ISSUE #1911: Start testing in Python 3.9
Python 3.9.0b1 is release already, and soon there'll be a full release. We should start running CI against it, to catch things like #1910

ktbarrett:
Adds Python 3.9 (currently dev) to Linux Travis CI. Makes a single fix to support Python 3.9. Closes #1911.

Justification[https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads] for removing the call to `PyEval_InitThreads()`.

### DIFF
--- a/cocotb/share/lib/embed/gpi_embed.cpp
+++ b/cocotb/share/lib/embed/gpi_embed.cpp
@@ -131,7 +131,6 @@ extern "C" void embed_init_python(void)
     set_program_name_in_venv();
     Py_Initialize();                    /* Initialize the interpreter */
     PySys_SetArgvEx(1, argv, 0);
-    PyEval_InitThreads();               /* Create (and acquire) the interpreter lock */
 
     /* Swap out and return current thread state and release the GIL */
     gtstate = PyEval_SaveThread();
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,15 @@ jobs:
       after_success: *travis_linux_post
 
     - stage: SimTests
+      python: 3.9-dev
+      before_install: *travis_linx_prep
+      script:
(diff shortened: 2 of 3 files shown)


cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -131,7 +131,6 @@ extern "C" void embed_init_python(void)
     set_program_name_in_venv();
     Py_Initialize();                    /* Initialize the interpreter */
     PySys_SetArgvEx(1, argv, 0);
-    PyEval_InitThreads();               /* Create (and acquire) the interpreter lock */
```
imphil, line 134: Even though this is unnecessary since Python 3.7+, we still need it for 3.5 and 3.6, if I'm reading the docs at https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads correctly. Can you gate it with an ifdef?

eric-wieser: Or we could just leave it in:

> Changed in version 3.9: The function now does nothing.

eric-wieser:
Does it make sense to add Python 3.9 to the supported versions in `setup.py` as part of this PR?

cocotb/share/lib/embed/gpi_embed.cpp
ktbarrett: I'd re-read the second sentence in the documentation carefully. And we cannot leave it in, it is marked as deprecated and `-Werror` causes the CI run to fail.

imphil: Ah yes. What a confusing piece of documentation.

PR #1929, opened 2020-06-12
README changes

### LINKED ISSUE #1914: README issues
Installation instructions in the README don't work for Windows, we should add a blurb and point to the documentation for Windows users.

The blurb about how to install the development version doesn't lead to the docs on installing the development version.

ktbarrett:
Fixes #1914. Also updates links to *latest* documentation to stable. Unfortunately, I couldn't do this for the Windows installation instruction because the link will have to change after the release. After the release this could be fixed. I'm not sure if updating the doc link in the documentation badge is a good idea or not, there are other links to the documentation on the page...

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,17 @@ After installing these dependencies, the latest stable version of cocotb can be 
 pip install cocotb
 ```
 
-For more details, including how to install a development version of cocotb, see [the documentation](https://docs.cocotb.org/en/latest/install.html).
+**!!! Windows Users !!!** See [here](https://docs.cocotb.org/en/stable/install.html) for installation instructions.
+
+For more details on installation, including prerequisites, see [the documentation](https://docs.cocotb.org/en/stable/install.html).
+
+For detail on how to install the *development* version of cocotb, see [the lastest documentation](https://docs.cocotb.org/en/latest/install_devel.html#install-devel).
 
 ## Usage
 
 As a first trivial introduction to cocotb, the following example "tests" a flip-flop.
 
-First, we need a hardware design which we can test. For this example, create a file `dff.sv` with SystemVerilog code for a simple [D flip-flop](https://en.wikipedia.org/wiki/Flip-flop_(electronics)#D_flip-flop). You could also use any other language a [cocotb-supported simulator](https://docs.cocotb.org/en/latest/simulator_support.html) understands, e.g. VHDL.
+First, we need a hardware design which we can test. For this example, create a file `dff.sv` with SystemVerilog code for a simple [D flip-flop](https://en.wikipedia.org/wiki/Flip-flop_(electronics)#D_flip-flop). You could also use any other language a [cocotb-supported simulator](https://docs.cocotb.org/en/stable/simulator_support.html) understands, e.g. VHDL.
 
 ```systemverilog
 // dff.sv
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 **cocotb** is a coroutine based cosimulation library for writing VHDL and Verilog testbenches in Python.
 
-[![Documentation Status](https://readthedocs.org/projects/cocotb/badge/?version=latest)](https://docs.cocotb.org/en/latest/)
+[![Documentation Status](https://readthedocs.org/projects/cocotb/badge/?version=stable)](https://docs.cocotb.org/en/stable/)
```
imphil, line 3: I'd argue this badge should continue to track latest, to be aligned with the build status badge.

README.md
```diff
@@ -26,13 +26,15 @@ After installing these dependencies, the latest stable version of cocotb can be
 pip install cocotb
 ```
 
-For more details, including how to install a development version of cocotb, see [the documentation](https://docs.cocotb.org/en/latest/install.html).
+:warning: **Windows Users** See [here](https://docs.cocotb.org/en/latest/install.html) for installation instructions.
```
eric-wieser, line 29: `:warning:` will not render on PyPI.

On that note - should we be linking to `stable` here?

README.md
```diff
@@ -26,13 +26,15 @@ After installing these dependencies, the latest stable version of cocotb can be
...
+
+For more details, including how to install a development version of cocotb, see [the documentation](https://docs.cocotb.org/en/latest/install_devel.html#install-devel).
```
eric-wieser, line 31:
```suggestion
For more details, including how to install a development version of cocotb, see [the documentation](https://docs.cocotb.org/en/stable/install_devel.html#install-devel).
```
?

README.md
ktbarrett: We could, but as soon as 1.4 releases, the link would break. I'll put in the unicode character.

README.md
ktbarrett: That link doesn't work. It's something else that will have to be updated after the release.

eric-wieser: We need to update it _before_ the release, because the readme ends up snapshotted on PyPI.

ktbarrett: You're right, it's okay if it's broken for a little bit, right?

eric-wieser: Much better than being broken for a whole release cycle.

ktbarrett: Actually I'm not sure if this should change. If we want to offer users the latest version, the instructions for how to install the latest dev version may not be correct if we reference stable.

The above line for Windows users should change though and will be broken until 1.4 is released.

eric-wieser: Unfortunately, this sentence is both about:

* More details (stable)
* Development version (latest)

Perhaps we should have
* This link to stable
* The development section of stable link to latest.

ktbarrett: I split the sentence in two. One reference stable for general "more info" and the other for documentation on installing the development version, which will deposit them in the right context to browse the rest of the development documentation.

README.md
ktbarrett: Perhaps it was just me, but that warning symbol overlaps the next character on my browser on my home computer, so I just went for the old `!!!`.

ktbarrett:
Please squash once ready.

Issue #778, opened 2019-01-22
Example mixed_language is incomplete


cmarqu:
The example ``mixed_language`` is not complete in that the SV and VHDL toplevel do not mirror each other (the VHDL toplevel is incomplete), and the testbench doesn't really check all that much - it should e.g. also check that the two chained ``endian_swappers`` produce output that is equal to the input.

ZiaCheemaGit:
@cmarqu @ktbarrett is there any reason this issue is still not closed even after merged PR #2912 and if still mixed language requires anything not covered in that PR I am looking to work on it.

ktbarrett:
Good eye. Closed by #2912.

PR #1931, opened 2020-06-13
Skip mixed_language example if TOPLEVEL_LANG is not Verilog.

themperek:
Skip mixed_language example if toplevel_lang  is not Verilog. It fails VHDL regression.


cmarqu:
See also #778

PR #1935, opened 2020-06-18
add share/lib/verilator/verilator.cpp to package

mtikekar:
The file was removed in #1853 but it is actually needed. Without it, you get an error like:

```
make[1]: *** No rule to make target '.../lib/python3.7/site-packages/cocotb/share/lib/verilator/verilator.cpp', needed by 'verilator.o'.  Stop.
```

We probably should add a verilator test to avoid a regression.


ktbarrett:
You are absolutely right about adding a verilator test. We also should really get the GHDL test working. Is verilator's support good enough to pass at least most of the regression?

PR #1936, opened 2020-06-19
C++11 docs

### LINKED ISSUE #1195: Document supported C and C++ version

ktbarrett:
Closes #1195. Specifically calls out C++11 compiler and adds documentation to the troubleshooting section on debugging issues related to loading the wrong libstdc++ (which is somewhat common).

### DIFF
--- a/documentation/source/troubleshooting.rst
+++ b/documentation/source/troubleshooting.rst
@@ -101,3 +101,46 @@ and having them ignored is likely to lead to strange errors.
 As a side note,
 when you need to *clear* a Makefile variable from the command line,
 use the syntax ``make EXTRA_ARGS=``.
+
+``GLIBCXX_3.4.XX`` not found
+============================
+
+This error can occur on Linux, and will raise ``ImportError: /some/libstdc++.so.6: version `GLIBCXX_3.4.XX' not found``.
+This occurs because an older non-C++11 version of libstdc++ is being loaded by the simulator or cocotb.
+It is usually an issue with your environment, but sometimes can occur when using very old version of certain simulators.
+
+Check your environment
+----------------------
+
+To see if your environment is the issue, look at the value of the ``LD_LIBRARY_PATH`` environment variable.
+Ensure the first path in the colon-deliminated list is the path to the libstdc++ that shipped with the compiler you used to build cocotb.
+
+.. code:: shell
+
+    echo $LD_LIBRARY_PATH
+
+This variable might be empty, in which case the loader looks in the system's libraries.
+If the library you built cocotb with is not first, prepend that path to the list.
+
+.. code:: shell
+
+    export LD_LIBRARY_PATH=/path/to/newer/libraries/:$LD_LIBRARY_PATH
+
+Check your simulator
+--------------------
+
+Sometimes, simulators modify the ``LD_LIBRARY_PATH`` so they point to the libraries that are shipped with instead of the system libraries.
+If you are running an old simulator, the packaged libraries may include a pre-C++11 libstdc++.
+To see if your simulator is modifying the ``LD_LIBRARY_PATH``, open the simulator up to an internal console and obtain the environment variable.
+
+For example, with Questa one could open Questa to a TCL console and run the ``env`` command to list the current environment.
+The ``LD_LIBRARY`` path should appear in the list.
+
+If the simulator does modify the ``LD_LIBRARY_PATH``, refer to the simulator documenation on how to prevent or work around this issue.
+
+For example, Questa ships with GCC.
+Sometimes that version of GCC is old enough to not support C++11 (<4.8).
+When you install cocotb, ``pip`` uses the system (or some other) compiler that supports C++11.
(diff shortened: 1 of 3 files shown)


documentation/source/troubleshooting.rst
```diff
@@ -101,3 +101,46 @@ and having them ignored is likely to lead to strange errors.
...
+
+To see if your environment is the issue, look at the value of the ``LD_LIBRARY_PATH`` environment variable.
+Ensure the first path in the colon-deliminated list is the path to the libstdc++ that shipped with the compiler you used to build cocotb.
+
+.. code:: shell
+
+    echo $LD_LIBRARY_PATH
+
+This variable might be empty, in which case the loader looks in the system's libraries.
+If the simulator you built cocotb with is not first, prepend that path to the list.
```
eric-wieser, line 123: Did you mean simulator?

documentation/source/troubleshooting.rst
```diff
@@ -101,3 +101,46 @@ and having them ignored is likely to lead to strange errors.
 As a side note,
 when you need to *clear* a Makefile variable from the command line,
 use the syntax ``make EXTRA_ARGS=``.
+
+``GLIBCXX_3.4.XX`` not found
+============================
```
eric-wieser, line 106: This heading doesn't seem to fit with the other headings in this file. Perhaps
```suggestion
Diagnosing ``ImportError``\ s of cocotb
=======================================
```
and then some subheadings for each message.

documentation/source/troubleshooting.rst
```diff
@@ -101,3 +101,46 @@ and having them ignored is likely to lead to strange errors.
...
+
+This error can occur on Linux, and will raise ``ImportError: /some/libstdc++.so.6: version `GLIBCXX_3.4.XX' not found``.
+This occurs because an older non-C++11 version of libstdc++ is being loaded by the simulator or cocotb.
+It is usually an issue with your environment, but sometimes can occur when using very old version of certain simulators.
```
cmarqu, line 110:
```suggestion
It is usually an issue with your environment, but sometimes can occur when using a very old version of certain simulators.
```

documentation/source/troubleshooting.rst
```diff
@@ -101,3 +101,46 @@ and having them ignored is likely to lead to strange errors.
...
+    export LD_LIBRARY_PATH=/path/to/newer/libraries/:$LD_LIBRARY_PATH
+
+Check your simulator
+--------------------
+
+Sometimes, simulators modify the ``LD_LIBRARY_PATH`` so they point to the libraries that are shipped with instead of the system libraries.
+If you are running an old simulator, the packaged libraries may include a pre-C++11 libstdc++.
+To see if your simulator is modifying the ``LD_LIBRARY_PATH``, open the simulator up to an internal console and obtain the environment variable.
+
+For example, with Questa one could open Questa to a TCL console and run the ``env`` command to list the current environment.
```
cmarqu, line 136:
```suggestion
For example, with Mentor Questa and Cadence Xcelium, one could open a Tcl console and run the ``env`` command to list the current environment.
```

documentation/source/troubleshooting.rst
```diff
@@ -101,3 +101,46 @@ and having them ignored is likely to lead to strange errors.
...
+The ``LD_LIBRARY`` path should appear in the list.
+
+If the simulator does modify the ``LD_LIBRARY_PATH``, refer to the simulator documenation on how to prevent or work around this issue.
```
cmarqu, line 139:
```suggestion
If the simulator does modify the ``LD_LIBRARY_PATH``, refer to the simulator documentation on how to prevent or work around this issue.
```

documentation/source/troubleshooting.rst
ktbarrett: That's too general and makes an assumption there is more than one type of that error and they are somehow related. We can group once we have more individual cases.

documentation/source/troubleshooting.rst
ktbarrett: I decided to look up whether it should be spelled TCL or Tcl since that stood out to me. Even though it started life as an acronym, the author, core documentation, and most people spell is Tcl. How fitting.

PR #1941, opened 2020-06-23
Convert most examples to async/await

eric-wieser:
Most of the rest of #1730 

Since this was easy, I think it's valuable to have these examples in `async` form for people learning from them in the next release.

### DIFF
--- a/examples/axi_lite_slave/tests/test_axi_lite_slave.py
+++ b/examples/axi_lite_slave/tests/test_axi_lite_slave.py
@@ -19,7 +19,7 @@ def setup_dut(dut):
 
 # Write to address 0 and verify that value got through
 @cocotb.test()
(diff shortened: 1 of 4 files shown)


imphil:
Wow, a 40 percent increase in coverage? Something must be very wrong with our collection mechanism.

eric-wieser:
> Wow, a 40 percent increase in coverage? Something must be very wrong with our collection mechanism.

I think that's just the `/head` vs `/merge` behavior of CI tools striking again.

ktbarrett:
Adder updates are in #1826, and this leaves the `ping_tun_tap` example as it was.

eric-wieser:
`ping_tun_tap` isn't tested by CI, so I figured I'd play it safe.

PR #1944, opened 2020-06-25
Backport #1882 to 1.3 branch

### LINKED ISSUE #1939: Backport VHDL pseudo-region handle fix to 1.3
Backport #1882 to 1.3 branch to fix the regression caused by #1164.

ktbarrett:
Closes #1939. Backports just the fix from PR #1882 to the 1.3 branch and moves the release notes to the correct place.


imphil:
The backport itself is fine, thanks. Before releasing 1.3.2 we should attempt to get CI at least running, i.e. disable the flake8 check.

eric-wieser:
Looks like the flake8 check fails because of an update to flake8 since 1.3. We should pin flake8 in our tox config to prevent this biting future maintenance branches.

PR #1946, opened 2020-06-25
Fix flake8 in 1.3 branch

ktbarrett:
Flake8 3.7.9 was used in the 1.3.1 release and checks passed, with flake8 3.8 a behavior changed that cause the checks to fail. Instead of updating the source in a released version, we pin flake8 to the old version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ jobs:
     - stage: PythonLint
       os: linux
       python: 3.5
-      install: pip install flake8
+      install: pip install "flake8>=3.7,<3.8"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - stage: PythonLint
       os: linux
       python: 3.5
-      install: pip install flake8
+      install: pip install "flake8>=3.7,<3.8"
```
eric-wieser, line 20: It might nice if we could make this a tox job on master, to keep any version-pinning in one place.

Issue #1942, opened 2020-06-23
Release cocotb 1.4.0


imphil:
Milestone: https://github.com/cocotb/cocotb/milestone/11

Release process checklist:
- [x] Close all issues blocking the milestone
- [x] Prepare release announcement email
- [x] Tag a release
- [x] Test release build on PyPi
- [x] Send out announcement to mailing list

cmarqu:
As a reminder, you once created https://github.com/cocotb/cocotb/wiki/Release-Process :)

ktbarrett:
Task list for myself related to this
- [ ] Test with Questa 10.5a (earliest version that runs on RHEL7)
- [ ] Test with Questa 10.7c
- [ ] Test with Questa 2019.03 (has had issues recently)
- [ ] Test with Questa 2020.02 (has had issues recently)
- [ ] Compare GHDL results between 1.3 and 1.4rc1 for additional regressions
- [ ] Check tarball
  - packaged examples run
  - essential non-runnable files exist

ktbarrett:
It does look like GHDL has regressed since 1.3.1. I'll look into it a bit more later.

elgorwi:
> Task list for myself related to this
> 
>     * [x]  Test with Questa 10.5a (earliest version that runs on RHEL7)
> 
>     * [x]  Test with Questa 10.7c
> 
>     * [x]  Test with Questa 2019.2 (has had issues recently)
> 
>     * [x]  Test with Questa 2020.1 (has had issues recently)
> 
>     * [x]  Compare GHDL results between 1.3 and 1.4rc1 for additional regressions
> 
>     * [x]  Check tarball
>       
>       * packaged examples run
>       * essential non-runnable files exist
> 
> 
> ## Results
> 
> Looks like GHDL did regress between 1.3.1 and 1.4. I will look into it more later.
> Running the packaged examples works with Icarus, Questa, GHDL (almost, the mean example is broken), Verilator (almost, dff example hangs, mean example won't compile).

Running the DFF example with Verilator is very slow (48s on my system). Changing the clock frequency from us to ns or alternatively changing the time precision to ns in dff.v will speed this up considerably.    
Running the mean example with Verilator is only matter of merging  #1901.
Support for arrays is required (https://github.com/ghdl/ghdl/pull/1249) to run the mean example with GHDL.

eric-wieser:
I note that the PyPI build of 1.4.0 does not contain a wheel. I assume this is a deliberate choice?

[...]

PR #1949, opened 2020-06-30
Support vpiConstant, which is used by GHDL

ktbarrett:
From https://github.com/cocotb/cocotb/issues/1942#issuecomment-651453987. For some reason GHDL is returning the type of constants as `vpiConstant` sometimes and `vpiParameter` other times. `vpiConstant` is synonymous with `vpiParameter` so this is not a problem.

### DIFF
--- a/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -127,6 +127,7 @@ gpi_objtype_t to_gpi_objtype(int32_t vpitype)
         case vpiIntegerNet:
             return GPI_INTEGER;
 
+        case vpiConstant:
         case vpiParameter:
             return GPI_PARAMETER;
 
@@ -186,6 +187,7 @@ GpiObjHdl* VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
         case vpiInterconnectNet:
             new_obj = new VpiSignalObjHdl(this, new_hdl, to_gpi_objtype(type), false);
             break;
+        case vpiConstant:
         case vpiParameter:
             new_obj = new VpiSignalObjHdl(this, new_hdl, to_gpi_objtype(type), true);
             break;


garmin-mjames:
I think GHDL is returning the correct types. It's just that GHDL uses `VPI` to access the VHDL design.
In https://github.com/cocotb/cocotb/blob/master/examples/mean/hdl/mean.vhd, `DATA_WIDTH` is a `constant` while `BUS_WIDTH` is a `generic`.

ktbarrett:
That's a reasonable explanation, but I'm not sure if that intended. Perhaps we can get @tgingold to weigh in?

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -127,6 +127,7 @@ gpi_objtype_t to_gpi_objtype(int32_t vpitype)
         case vpiIntegerNet:
             return GPI_INTEGER;
 
+        case vpiConstant:
```
garmin-mjames, line 130: We should probably figure out which type of constant it is and return that type here. In any case, the handle will be a `ConstantObject`, but calling `get_type` on the handle will be wrong.

garmin-mjames:
I think it makes sense to map entity generics to `vpiParameter`, as it has the same functionality.

cocotb/share/lib/vpi/VpiImpl.cpp
garmin-mjames: `vpiParameter` also supports `vpiConstType`, so this would be a good opportunity to deprecate `GPI_PARAMETER` and use the appropriate types for both cases.

ktbarrett: Yeah, GPI_PARAMETER is not a great solution and I'd like to remove it, but I was considering this PR may come in with 1.4 since it's a fairly innocuous change. That could be another PR if you are up to it.

tgingold:
Yes, I have mapped a constant declaration to vpiConstant and a generic declaration to vpiParameter as they are naturally equivalent.

garmin-mjames:
This might also partially fix `test_discovery.access_const_string_verilog` for Aldec simulators. At least in Active-HDL, `STRING_CONST` is a `vpiConstant`.
The test would still fail, but not raise an `AttributeError`.

cocotb/share/lib/vpi/VpiImpl.cpp
ktbarrett: This would probably be a part of a bigger refactor to handle where we deprecate `ConstantObject` and add a constant check to the instances.

ktbarrett:
Going forward with (https://github.com/cocotb/cocotb/pull/1949#discussion_r447371701) since this isn't making 1.4.

PR #1957, opened 2020-07-08
Documentation: Fix error in coroutines doc

imphil:


imphil:
Just stumbled across this one when writing the release notes, seems like an oversight.

PR #1961, opened 2020-07-09
fix vhdl error in adder example

themperek:
Fix syntax error.


ktbarrett:
Hmmm, I guess that accidentally snuck in when I was making my final changes for #1826. Sorry about that...

PR #1968, opened 2020-07-10
Add Glossary section

### LINKED ISSUE #1966: Add glossary definitions for coroutine terms
There are a few cases in the documentation and a few more in docstrings (and a lot of people I talk to) where the terms regarding coroutines are a bit muddled together. It would be nice to add a glossary to the documentation and explicitly define all the terms regarding coroutines.

* **coroutine function**: the definition of a function that when called returns a coroutine object. Implemented using async functions.
* **coroutine**: the result of calling an coroutine function. Coroutines are not run immediately, you must either: `await` on them which blocks the awaiting coroutine until it is finished; or fork the coroutine, turning it into a task, which runs concurrently.
* **task**: the result of forking a coroutine, it represents a concurrently running coroutine.

The old syntax uses `coroutine`s for coroutine functions, and calling them resulted in a `RunningCoroutine`, which confuses the matter.

cmarqu:
Closes #1966.

### DIFF
--- a/documentation/source/glossary.rst
+++ b/documentation/source/glossary.rst
@@ -1,0 +1,64 @@
+.. _glossary:
+
+Glossary
+========
+
+.. glossary::
+
+   coroutine function
+      The definition of a function that, when called, returns a coroutine object.
+      Implemented using :keyword:`async` functions.
+      See also the :term:`Python glossary <python:coroutine function>`.
+   coroutine
+      The result of calling a :term:`coroutine function`.
+      Coroutines are not run immediately, you must either
+      :keyword:`await` on them which blocks the awaiting coroutine until it is finished;
+      or fork the coroutine, turning it into a :term:`task`, which runs concurrently.
+      See also the :term:`Python glossary <python:coroutine>`.
+
+   DUT
+      Design under Test
+
+   DUV
+      Design under Verification
+
+   FLI
+      Foreign Language Interface. Mentor Graphics' equivalent to :term:`VHPI`
+
+   GPI
+      General Procedural Interface, cocotb's abstraction over :term:`VPI`, :term:`VHPI`, and :term:`FLI`.
+
+   HDL
+      Hardware Description Language
+
+   MDV
+      Metric-driven Verification
+
+   RTL
+      Register Transfer Level
+
+   task
+      The result of forking a :term:`coroutine`. A task represents a concurrently running coroutine.
+
+   UVM
+      Universal Verification Methodology
+
+   VHPI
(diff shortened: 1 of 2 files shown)


cmarqu:
@ktbarrett Could you suggest something for "async function"?

documentation/source/glossary.rst
```diff
@@ -0,0 +1,66 @@
...
+========
+
+.. glossary::
+
+   async function
+      FIXME
+
+   coroutine function
+      The definition of a function that, when called, returns a coroutine object.
+      Implemented using :term:`async functions<async function>`.
```
ktbarrett, line 13: I would link back to Python's documentation on async syntax. Here is Python's glossary on [coroutines and coroutine functions](https://docs.python.org/3/glossary.html#term-coroutine) and [async def functions](https://docs.python.org/3/reference/compound_stmts.html#async-def).

documentation/source/glossary.rst
```diff
@@ -0,0 +1,66 @@
...
+
+   coroutine
+      The result of calling a :term:`coroutine function`.
+      Coroutines are not run immediately, you must either:
+      await on them which blocks the awaiting coroutine until it is finished;
```
ktbarrett, line 18: I seem to remember being able to crossref syntax, I may be wrong...

cmarqu:
Looking good?: https://cocotb--1968.org.readthedocs.build/en/1968/glossary.html

documentation/source/glossary.rst
```diff
@@ -0,0 +1,64 @@
+.. _glossary:
+
+Glossary
+========
+
+.. glossary::
+
+   coroutine function
+      The definition of a function that, when called, returns a coroutine object.
+      Implemented using :ref:`async functions<python:async def>`.
```
ktbarrett, line 10:
```suggestion
      Implemented using :keyword:`async` functions.
      See also the :term:`Python glossary <python:coroutine function>`.
```

documentation/source/glossary.rst
```diff
@@ -0,0 +1,64 @@
...
+
+   coroutine
+      The result of calling a :term:`coroutine function`.
+      Coroutines are not run immediately, you must either
+      :keyword:`await` on them which blocks the awaiting coroutine until it is finished;
+      or fork the coroutine, turning it into a :term:`task`, which runs concurrently.
+      See also the :term:`Python coroutine glossary <python:coroutine>`.
```
ktbarrett, line 17:
```suggestion
      See also the :term:`Python glossary <python:coroutine>`.
```

PR #1970, opened 2020-07-10
Fix flake8 F841 reports for core code

cmarqu:
Ignore this ("local variable ... is assigned to but never used")
for tests and examples where names are useful for didactic purposes.

Refs #1547.

### DIFF
--- a/cocotb/drivers/amba.py
+++ b/cocotb/drivers/amba.py
@@ -292,6 +292,7 @@ class AXI4Slave(BusDriver):
                     "AWLEN   %d\n" % _awlen +
                     "AWSIZE  %d\n" % _awsize +
                     "AWBURST %d\n" % _awburst +
+                    "AWPROT %d\n" % _awprot +
                     "BURST_LENGTH %d\n" % burst_length +
                     "Bytes in beat %d\n" % bytes_in_beat)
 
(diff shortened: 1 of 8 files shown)


cocotb/drivers/amba.py
```diff
@@ -341,6 +342,7 @@ def _read_data(self):
                     "ARLEN   %d\n" % _arlen +
                     "ARSIZE  %d\n" % _arsize +
                     "ARBURST %d\n" % _arburst +
+                    "ARPROT %d\n" % _arprot +
```
ktbarrett, line 345: I would go the other way with this and remove the `_arprot` variable. And the same goes for `_awprot`. It's not particularly useful information, and isn't widely used, which is probably why it wasn't printed in the first place.

setup.cfg
```diff
@@ -4,6 +4,7 @@ exclude =
     documentation
     makefiles
     venv
+    .tox
```
ktbarrett, line 7: Thank you, I've noticed flake8 takes forever and fails a lot because of this.

cocotb/drivers/amba.py
cmarqu: Actually, my current project does/will use it (unfortunately, it's not verified using cocotb). Not sure if I would need to see it printed.

PR #1975, opened 2020-07-12
Fix warning introduced by recent change

### LINKED ISSUE #1974: Compiler error
It seems that recently GCC has added a new warning, or started catching warnings, etc. The clever hack from #1906 now throws warnings. Tested with GCC 9.3 on Debian Unstable.

```
    gcc -pthread -B /home/kaleb/miniconda3/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -Werror -fPIC -DVHPI_CHECKING=1 -DIUS= -Icocotb/share/include -I/home/kaleb/dev/cocotb/.tox/py3/include -I/home/kaleb/miniconda3/include/python3.7m -c cocotb/share/lib/vhpi/VhpiCbHdl.cpp -o build/temp.linux-x86_64-3.7/cocotb/share/lib/vhpi/VhpiCbHdl.o -std=c++11 -Wall -Wextra -Wcast-qual -Wwrite-strings -Wconversion -Wnon-virtual-dtor -Woverloaded-virtual -D__STDC_FORMAT_MACROS
    cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++
    cocotb/share/lib/vhpi/VhpiCbHdl.cpp: In lambda function:
    cocotb/share/lib/vhpi/VhpiCbHdl.cpp:940:9: error: assignment from temporary initializer_list does not extend the lifetime of the underlying array [-Werror=init-list-lifetime]
      940 |         }},
          |         ^
    cocotb/share/lib/vhpi/VhpiCbHdl.cpp:946:9: error: assignment from temporary initializer_list does not extend the lifetime of the underlying array [-Werror=init-list-lifetime]
      946 |         }},
          |         ^
    cocotb/share/lib/vhpi/VhpiCbHdl.cpp:957:9: error: assignment from temporary initializer_list does not extend the lifetime of the underlying array [-Werror=init-list-lifetime]
      957 |         }},
          |         ^
    cocotb/share/lib/vhpi/VhpiCbHdl.cpp:969:9: error: assignment from temporary initializer_list does not extend the lifetime of the underlying array [-Werror=init-list-lifetime]
      969 |         }},
          |         ^
    cc1plus: all warnings being treated as errors
```

ktbarrett:
Fixes #1974. The clever hack used to improve readability in #1906 invokes UB relating to the lifetime of the initializer_list. This causes GCC to throw a warning. Which fails due to `-Werror`.

### DIFF
--- a/cocotb/share/lib/fli/FliImpl.cpp
+++ b/cocotb/share/lib/fli/FliImpl.cpp
@@ -577,16 +577,20 @@ GpiIterator *FliImpl::iterate_handle(GpiObjHdl *obj_hdl, gpi_iterator_sel_t type
 }
 
 decltype(FliIterator::iterate_over) FliIterator::iterate_over = []{
-    std::initializer_list<FliIterator::OneToMany> region_options;
-    std::initializer_list<FliIterator::OneToMany> signal_options;
-    std::initializer_list<FliIterator::OneToMany> variable_options;
+    std::initializer_list<FliIterator::OneToMany> region_options = {
+        FliIterator::OTM_CONSTANTS,
+        FliIterator::OTM_SIGNALS,
+        FliIterator::OTM_REGIONS,
+    };
+    std::initializer_list<FliIterator::OneToMany> signal_options = {
+        FliIterator::OTM_SIGNAL_SUB_ELEMENTS,
+    };
+    std::initializer_list<FliIterator::OneToMany> variable_options = {
+        FliIterator::OTM_VARIABLE_SUB_ELEMENTS,
+    };
 
     return decltype(FliIterator::iterate_over) {
-        {accArchitecture, region_options = {
-            FliIterator::OTM_CONSTANTS,
-            FliIterator::OTM_SIGNALS,
-            FliIterator::OTM_REGIONS,
-        }},
+        {accArchitecture, region_options},
         {accEntityVitalLevel0, region_options},
         {accArchVitalLevel0, region_options},
         {accArchVitalLevel1, region_options},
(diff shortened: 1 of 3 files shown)


eric-wieser:
I'd be keen to get a language-lawyer ruling from stackoverflow on whether the original code was defined by the standard, and if so if the gcc warning is in error.

ktbarrett:
Even if GCC is erroneously complaining, without this change I can't run regressions on my local machine. And this will continue to be an issue for users with the particular versions of GCC that are affected. And the change is simply aesthetic.

eric-wieser:
I'm not saying we shouldn't put this change in - I'd just like the commit message to either say "the previous code invoked undefined behavior" or "the previous code ran into a bug in GCC x.y"

ktbarrett:
Made a [stackoverflow question](https://stackoverflow.com/questions/62877374/gcc-throws-init-list-lifetime-warning-on-potentially-valid-code) about it.

eric-wieser:
I'd perhaps encourage you to rewrite the question to just use the minimal godbolt example, which is a great demo of the issue.

ktbarrett:
The issue is example `i4` in [this](http://eel.is/c++draft/dcl.init.list#6) section of the C++ standard.

```c++
struct A {
  std::initializer_list<int> i4;
  A() : i4{ 1, 2, 3 } {}            // ill-formed, would create a dangling reference
};
```

> For i4, the initializer_­list object is initialized in the constructor's ctor-initializer as if by binding a temporary array to a reference member, so the program is ill-formed.

So in fact, the usage is UB and is correctly detected as errant by GCC.

ktbarrett:
@eric-wieser Can we get this merged? I still can't run regressions with tox locally.

eric-wieser:
I'd prefer to remove the "invokes undefined behavior" part of the commit message - it remains my opinion that the code does not. Maybe put the gcc warning message and a link to the stackoverflow post in the commit message?

eric-wieser:
I think the UB claim is equivalent to claiming this code invokes UB:
```
{
    initializer_list<int> a;
    {
        initializer_list<int> b = {1, 2};
        a = b;
    }
    // a has a dangling pointer here, but is it actually UB if we never try to use it?
}
```

ktbarrett:
@eric-wieser Changed the commit message to not mention UB, just that GCC doesn't like it and points it back to this PR.

eric-wieser:
Although it's always a good idea to write commit messages on the assumption that GitHub may no longer exist for the user reading them - hence the inclination to include the actually warning message.

ktbarrett:
I can't say in this case I grasp the details well enough to write a good commit message.

PR #1976, opened 2020-07-12
Deprecate GPI_PARAMETER and remove it from VPI

ktbarrett:
GPI_PARAMETER represents a parameter or a constant in VHDL, but it
doesn't say anything about what *kind* of constant value it is. VHPI and
FLI already don't support GPI_PARAMETER because of this issue, but the
VPI does. This commit implements the small amount of features necessary
to map parameters to the correct GPI type. Constness is handled in the
VpiSignalObj class. It also deprecates the GPI_PARAMETER constant in
both C and Python.

Following https://github.com/cocotb/cocotb/pull/1949#discussion_r447365438.

### DIFF
--- a/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -94,7 +94,7 @@ const char *VpiImpl::get_simulator_version()
     return m_version.c_str();
 }
 
-gpi_objtype_t to_gpi_objtype(int32_t vpitype)
+static gpi_objtype_t to_gpi_objtype(int32_t vpitype)
 {
     switch (vpitype) {
         case vpiNet:
@@ -127,9 +127,6 @@ gpi_objtype_t to_gpi_objtype(int32_t vpitype)
         case vpiIntegerNet:
             return GPI_INTEGER;
 
-        case vpiParameter:
-            return GPI_PARAMETER;
-
         case vpiStructVar:
         case vpiStructNet:
         case vpiUnionVar:
@@ -153,6 +150,27 @@ gpi_objtype_t to_gpi_objtype(int32_t vpitype)
 
         default:
             LOG_DEBUG("Unable to map VPI type %d onto GPI type", vpitype);
+            return GPI_UNKNOWN;
+    }
+}
+
+static gpi_objtype_t const_type_to_gpi_objtype(int32_t const_type)
+{
+    switch (const_type)
+    {
+        case vpiDecConst:
+        case vpiBinaryConst:
+        case vpiOctConst:
+        case vpiHexConst:
+        case vpiIntConst:
+            return GPI_INTEGER;
+        case vpiRealConst:
+            return GPI_REAL;
+        case vpiStringConst:
+            return GPI_STRING;
+        //case vpiTimeConst:  // Not implemented
+        default:
+            LOG_DEBUG("Unable to map vpiConst type %d onto GPI type", const_type);
             return GPI_UNKNOWN;
     }
 }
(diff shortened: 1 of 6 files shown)


cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -157,6 +154,27 @@ gpi_objtype_t to_gpi_objtype(int32_t vpitype)
     }
 }
 
+static gpi_objtype_t param_type_to_gpi_objtype(int32_t param_type)
```
garmin-mjames, line 157:
```suggestion
static gpi_objtype_t const_type_to_gpi_objtype(int32_t param_type)
```

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -157,6 +154,27 @@ gpi_objtype_t to_gpi_objtype(int32_t vpitype)
...
+        case vpiHexConst:
+        case vpiIntConst:
+            return GPI_INTEGER;
+        case vpiRealConst:
+            return GPI_REAL;
+        case vpiStringConst:
+            return GPI_STRING;
+        //case vpiTimeConst:
+        default:
+            LOG_DEBUG("Unable to map VPI type %d onto GPI type", param_type);
```
garmin-mjames, line 173:
```suggestion
            LOG_DEBUG("Unable to map vpiConst type %d onto GPI type", param_type);
```

garmin-mjames:
I think that `test_discovery.access_const_string_verilog` can be run by Riviera now. It'll still fail due to `ConstantObject` rather than `StringObject`, though.
https://github.com/cocotb/cocotb/blob/597327e8742bb2dd7fb7d67d410ab24e6469338f/tests/test_cases/test_discovery/test_discovery.py#L277-L279

ktbarrett:
@garmin-mjames Yeah that test isn't correct because of the `ConstantObject`/`StringObject` confusion. I might work on blowing `ConstantObject` away next, then I'll worry about the test.

garmin-mjames:
Your commit message says:

> GPI_PARAMETER represents a parameter or a constant in VHDL. ...

Shouldn't that be Verilog, not VHDL?

ktbarrett:
Better commit message.

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -461,7 +461,7 @@ def type_check_verilog(dut):
     test_handles = [
         (dut.stream_in_ready, "GPI_REGISTER"),
         (dut.register_array, "GPI_ARRAY"),
-        (dut.NUM_OF_MODULES, "GPI_PARAMETER"),
+        (dut.NUM_OF_MODULES, "GPI_INTEGER"),
```
elgorwi, line 464: Are there any tests for the other cases such as GPI_REAL, GPI_STRING?
Which simulators have been tested with this change?

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -157,6 +154,27 @@ gpi_objtype_t to_gpi_objtype(int32_t vpitype)
     }
 }
 
+static gpi_objtype_t const_type_to_gpi_objtype(int32_t const_type)
```
elgorwi, line 157: Is something similar needed for VHPI/FLI?

ktbarrett: Nope, they already don't use GPI_PARAMETER.

tests/test_cases/test_discovery/test_discovery.py
ktbarrett: Icarus, GHDL, and Active-HDL. I suppose I can test Questa real quick. And I don't think there are other tests, they are not exhaustive, I'll add some.

ktbarrett: Added constant integers, strings, and reals to this test. It should be tested using Icarus, Verilator, and Active-HDL.

garmin-mjames:
Note to self:
- [ ] Test Active-HDL

imphil (approved):
LGTM, thanks!

PR #1979, opened 2020-07-12
Add link to issues for the respective simulator

cmarqu:


umarcor:
@cmarqu, you can defin helpers to reduce verbosity and make it easier to read. See https://github.com/ghdl/ghdl-cosim/blob/master/doc/conf.py#L137-L145.

PR #1982, opened 2020-07-14
Detection of TOPLEVEL_LANG

### LINKED ISSUE #1893: Guess or better document the need to use TOPLEVEL_LANG
TOPLEVEL_LANG is a requirement for using some, but not all simulators. To make our "recommended" Makefile portable across simulators, we should *either*

- (easy option) Add it to our recommended makefile at https://docs.cocotb.org/en/latest/quickstart.html#creating-a-makefile
- or (better option) Guess TOPLEVEL_LANG if possible. A reasonably safe guess would be: if only VERILOG_SOURCES is set, use verilog. If only VHDL_SOURCES is set, use vhdl. Otherwise, do nothing.

ktbarrett:
Adds support for the automatic detection of an appropriate `TOPLEVEL_LANG` value based on the value of `VERILOG_SOURCES` and `VHDL_SOURCES`. This PR also adds detection of invalid `TOPLEVEL_LANG` values in simulators that only support one language like Verilator, Icarus, and GHDL. Closes #1893.

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.ghdl
+++ b/cocotb/share/makefiles/simulators/Makefile.ghdl
@@ -26,12 +26,14 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
+TOPLEVEL_LANG ?= vhdl
+
 include $(shell cocotb-config --makefiles)/Makefile.inc
 
-ifneq ($(VERILOG_SOURCES),)
+ifneq ($(or $(filter-out $(TOPLEVEL_LANG),vhdl),$(VERILOG_SOURCES)),)
 
 $(COCOTB_RESULTS_FILE):
-	@echo "Skipping simulation as Verilog is not supported on simulator=$(SIM)"
+	@echo "Skipping simulation as only VHDL is supported on simulator=$(SIM)"
 clean::
 
 else
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -111,6 +111,17 @@ $(SIM_BUILD):
 .PHONY: regression
 regression: $(COCOTB_RESULTS_FILE)
 
+# Attempt to detect TOPLEVEL_LANG based on available sources if not set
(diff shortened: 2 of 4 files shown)


cocotb/share/makefiles/simulators/Makefile.ghdl
```diff
@@ -26,15 +26,12 @@
...
-ifneq ($(VERILOG_SOURCES),)
+include $(shell cocotb-config --makefiles)/Makefile.inc
 
-$(COCOTB_RESULTS_FILE):
-	@echo "Skipping simulation as Verilog is not supported on simulator=$(SIM)"
-clean::
-
-else
+ifeq ($(VERILOG_SOURCES),)
+ifeq ($(TOPLEVEL_LANG),vhdl)
```
themperek, line 34:
```suggestion
ifneq ($(and $(VHDL_SOURCES),$(filter $(TOPLEVEL_LANG),vhdl)),)
```
Something like this? Does not look great but avoids repeating code.

ktbarrett: Do we even care about the value of sources? GHDL, Verilator, and Icarus will just ignore the value of `*_SOURCES` for the language it doesn't support.

ktbarrett: I don't even think we need the `*_SOURCES` check.

cocotb/share/makefiles/simulators/Makefile.ghdl
```diff
@@ -26,12 +26,14 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
+TOPLEVEL_LANG ?= vhdl
+
 include $(shell cocotb-config --makefiles)/Makefile.inc
 
-ifneq ($(VERILOG_SOURCES),)
+ifeq ($(filter $(TOPLEVEL_LANG),vhdl),)
```
themperek, line 33: You can skip filter.

ktbarrett:
@themperek Turns out we did need the `*_SOURCES` check to prevent issues with mixed language tests.

themperek (approved):

Issue #1943, opened 2020-06-23
Hard-coded RPATH in libraries


mtikekar:
When I run `pip install cocotb==1.4.0rc1` in one conda environment, it builds and installs a wheel and saves the wheel in a cache. Now, when I run the same command in a new conda environment, it reuses the wheel. The problem is that the `.so` files have a hard-coded RPATH to the first environment, so when that env is deleted, cocotb breaks in the other env.

From what I can tell, conda solves this problem for its packages: https://docs.conda.io/projects/conda-build/en/latest/resources/make-relocatable.html

themperek:
What OS is this? What is the error?

mtikekar:
This was on Linux. The error message was something like libpython3.7m.so not found followed by the simulator (xcelium) crashing.

eric-wieser:
Can you give an example of the hard-coded rpath?

themperek:
Is this?
https://github.com/cocotb/cocotb/blob/c2acc9f4608660271f7035a178eb0cc9273ce065/cocotb_build_libs.py#L217

Can we maybe use `PYTHONHOME` to load this?
Block making the wheel package?

mtikekar:
> Can you give an example of the hard-coded rpath?

The libcocotb.so file has rpath that looks something like: `/home/mtikekar/.conda/envs/envname/lib:$ORIGIN`.

> Can we maybe use `PYTHONHOME` to load this?

I built a conda package for cocotb to see what they do. I ran:

```
conda skeleton pypi cocotb
conda build cocotb
```

It basically runs pip and then goes through each executable and replaces any hard-coded rpaths with relative paths using `$ORIGIN`. For libcocotb.so in the conda package, rpath is `$ORIGIN/../../../..:$ORIGIN`.

eric-wieser:
@themperek's point about `PYTHON_SO_LIB` seems salient too - our binary contains a string constant that presumably has a bad path as well.

themperek:
So I have this:
```
building 'cocotb/libs/libcocotb' extension
    creating build/temp.linux-x86_64-3.7/cocotb/share/lib/embed
[...]

PR #1417, opened 2020-02-17
Revert "remove PY_SO_LIB and do not try to load python dynamically"

themperek:
Reverts cocotb/cocotb#1416

It is needed for commercial simulators.

PR #1985, opened 2020-07-14
Do not link with libpython

tpambor:


ktbarrett:
Unless we add the path to libpython to the library search path before loading libpython3.X.so this won't work (there are multiple ways to accomplish this). It works on Travis Linux with Python 3.8 because the Python is installed to the system, so libpython is in the system library search path. If you used conda, this wouldn't work. There is a treatment of the issues in #1943.

tpambor:
This was a little test related to #1943.

In [`gpi_embed.cpp`](https://github.com/cocotb/cocotb/blob/master/cocotb/share/lib/embed/gpi_embed.cpp#L125) libpython is loaded dynamically (`utils_dyn_open(PY_SO_LIB)`) but as it is linked explicitly actually the system will already try to load it when `libcocotb.so` is loaded. Therefore, I think `utils_dyn_open(PY_SO_LIB)` does not really have an effect in loading `libpython`.

In case of #1943, the error on load of `libcocotb.so` is:
`libpython3.7m.so.1.0: cannot open shared object file: No such file or directory.`

If `libpython` is not explicitly linked this error will go away because runtime loading is used but likely a new error will appear:
`..s/cocotb_utils.cpp:69   in utils_dyn_open Unable to open lib libpython3.7m.so: dlopen(libpython3.7m.so, 9): image not found`
[See here in CI](https://travis-ci.org/github/cocotb/cocotb/jobs/707984795#L1724)

#1943 could be closed by doing something like `utils_dyn_open(find_libpython())` or maybe by passing the path via an environment variable. This way `LD_LIBRARY_PATH` and `PATH` could stay untouched. [JuliaPy](https://github.com/JuliaPy/PyCall.jl/blob/master/src/pyinit.jl#L161) is solving it more or less this way.

ktbarrett:
> Therefore, I think `utils_dyn_open(PY_SO_LIB)` does not really have an effect in loading `libpython`.

This is correct, I've seen logs where it failed to load PY_SO_LIB, but cocotb still runs. Doing both load-time and runtime dynamic loading of libpython is unnecessary.

>  [JuliaPy](https://github.com/JuliaPy/PyCall.jl/blob/master/src/pyinit.jl#L161) is solving it more or less this way.

:eyes:

ktbarrett:
[Here](https://gist.github.com/tkf/d980eee120611604c0b9b5fef5b8dae6) is the original source for the `find_libpython` script. It's MIT licensed.

themperek:
Some observations:
1. If one does not link dynamically (second time) some simulators fail (especially commercial) in a very intriguing way see: https://github.com/cocotb/cocotb/pull/1417#issuecomment-587170566

2. To load dynamically `PY_SO_LIB` has to be right and on mac was always some issues. Since I do not have mac is very hard to debug for me.

Some kind for `find_python` function could be a solution. This can also try to resolve #1943.

ktbarrett:
@themperek After looking the issue you were seeing I feel like this is an issue with *how* the loader is loading libpython when it's loaded as a dependency compared to how we are loading with the dlopen call (does the loader not load globally by default?). It is definitely loading the same library; there is no way the library search path is changing between the calls. I need to find some documentation and run some tests.

@tpambor Since I see you are using `find_libpython`, I should mention that I found it useful, so took the liberty of turning it into a standalone package last night and it's now available on [Pypi](https://pypi.org/project/find-libpython/). If it ends up being a solution, we can decide whether to copy it or have it as a dependency.

themperek:
BTW. 
I do not know if this is a good idea/practice but can we call `cocotb-config` from `c` code? This would simplified makefiles and custom flows. @eric-wieser

ktbarrett:
Developing for Windows is a confusing nightmare to me, I'm still learning. Am I correct in understanding that the Windows issues would be resolved with an import library? It would ensure those missing functions are defined as stubs that would runtime load the implementations, right? Is there an import library available for libpython?

tpambor:
> Developing for Windows is a confusing nightmare to me, I'm still learning. Am I correct in understanding that the Windows issues would be resolved with an import library? It would ensure those missing functions are defined as stubs that would runtime load the implementations, right? Is there an import library available for libpython?
> 
> EDIT: turns out what the libpython anaconda package provides _is_ an import library. So maybe on Windows we should be linking against libpython.

An import library is used when the library/program is linked to the DLL implicitly at link time. Here a delay-import library is required which is used for linking to the DLL at run time. It can be created with `dlltool --def libpython3.8.def --output-delaylib libpython3.8.dll.a`. The .def file itself can be created with `gendef libpython3.8.dll > libpython3.8.def`

tpambor:
I think this can be closed with #2268 merged.

PR #1987, opened 2020-07-16
Provide <SimName>Version classes to compare simulator versions

### LINKED ISSUE #1951: Function for comparing simulator version strings?
Icarus Verilog master, reporting ``SIM_VERSION`` "11.0 (devel)", now supports bit selects (see https://github.com/steveicarus/iverilog/issues/323).
Thus, we need a way to selectively expect an error in our tests based on the version.

While I could compare with either the "11.0" string part, or the "10." string of the released version (and assume nobody uses a version before 10), or treat the number part as a float etc., this gets all ugly for the general case.
Since we have to support more such older/newer comparisons in the future (e.g. for upcoming GHDL improvements), does it make sense to provide a version comparison function that would gradually be extended to handle more simulators as needed?

cmarqu:
Currently, these are directly using `distutils.version.LooseVersion`.
Intended use case is for setting `expect_errors` for testcases.

Closes #1951

### DIFF
--- a/cocotb/_sim_versions.py
+++ b/cocotb/_sim_versions.py
@@ -6,7 +6,9 @@
 Classes to compare simulation versions.
 
 These are for cocotb-internal use only.
-Warning: these classes silently allow comparing versions of different simulators.
+
+.. warning::
+    These classes silently allow comparing versions of different simulators.
 """
 
 from distutils.version import LooseVersion
--- a/cocotb/_sim_versions.py
+++ b/cocotb/_sim_versions.py
@@ -1,0 +1,117 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""
+Classes to compare simulation versions.
+
+These are for cocotb-internal use only.
+Warning: these classes silently allow comparing versions of different simulators.
+"""
+
+from distutils.version import LooseVersion
+
+
+class ActivehdlVersion(LooseVersion):
+    """Version numbering class for Aldec Active-HDL.
+
+    NOTE: unsupported versions exist, e.g.
+    ActivehdlVersion("10.5a.12.6914") > ActivehdlVersion("10.5.216.6767")
+    """
+    pass
+
+
+class CvcVersion(LooseVersion):
+    """Version numbering class for Tachyon DA CVC.
+
+    Example:
+        >>> CvcVersion("OSS_CVC_7.00b-x86_64-rhel6x of 07/07/14 (Linux-elf)") > CvcVersion("OSS_CVC_7.00a-x86_64-rhel6x of 07/07/14 (Linux-elf)")
+        True
+    """
+    pass
+
+
+class GhdlVersion(LooseVersion):
+    """Version numbering class for GHDL."""
+    pass
+
+
+class IcarusVersion(LooseVersion):
+    """Version numbering class for Icarus Verilog.
+
+    Example:
+        >>> IcarusVersion("11.0 (devel)") > IcarusVersion("10.3 (stable)")
+        True
+        >>> IcarusVersion("10.3 (stable)") <= IcarusVersion("10.3 (stable)")
+        True
+    """
+    pass
+
+
+class ModelsimVersion(LooseVersion):
+    """Version numbering class for Mentor ModelSim."""
+    pass
+
+
+class QuestaVersion(LooseVersion):
+    """Version numbering class for Mentor Questa.
+
+    Example:
+        >>> QuestaVersion("10.7c 2018.08") > QuestaVersion("10.7b 2018.06")
+        True
+        >>> QuestaVersion("2020.1 2020.01") > QuestaVersion("10.7c 2018.08")
+        True
+    """
+    pass
+
+
+class RivieraVersion(LooseVersion):
+    """Version numbering class for Aldec Riviera-PRO.
+
+    Example:
+        >>> RivieraVersion("2019.10.138.7537") == RivieraVersion("2019.10.138.7537")
+        True
+    """
+    pass
+
+
+class VcsVersion(LooseVersion):
+    """Version numbering class for Synopsys VCS.
+
+    Example:
+        >>> VcsVersion("Q-2020.03-1_Full64") > VcsVersion("K-2015.09_Full64")
+        True
+    """
+    pass
+
+
+class VerilatorVersion(LooseVersion):
+    """Version numbering class for Verilator.
+
+    Example:
+        >>> VerilatorVersion("4.032 2020-04-04") > VerilatorVersion("4.031 devel")
+        True
+    """
+    pass
+
+
+class XceliumVersion(LooseVersion):
+    """Version numbering class for Cadence Xcelium.
+
+    Example:
+        >>> XceliumVersion("20.06-g183") > XceliumVersion("20.03-s002")
+        True
+        >>> XceliumVersion("20.07-e501") > XceliumVersion("20.06-g183")
+        True
+    """
+    pass
+
+
+class IusVersion(XceliumVersion):  # inherit everything from Xcelium
+    """Version numbering class for Cadence IUS.
+
+    Example:
+        >>> IusVersion("15.20-s050") > IusVersion("15.20-s049")
+        True
+    """
+    pass
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -95,7 +95,6 @@ jobs:
             lang: verilog
             python-version: 3.8
             os: ubuntu-20.04
-            may_fail: true
 
             # Test GHDL on Ubuntu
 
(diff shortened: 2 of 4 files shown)


cocotb/utils.py
```diff
@@ -581,3 +565,127 @@ def remove_traceback_frames(tb_or_exc, frame_names):
...
+        >>> IusVersion("15.20-s050") > IusVersion("15.20-s049")
+        True
+    """
+    pass
+
+
+if __name__ == "__main__":
+    import random
+    import doctest
+    doctest.testmod()
```
eric-wieser, line 678: I think doctests are run by pytest now?

eric-wieser:
I'd be tempted to put these in their own submodule, and not lump them into utils.

cocotb/utils.py
cmarqu: Hmm, are they? We would need the ``--doctest-modules`` option in ``tox.ini`` for that, or is it done differently?

cmarqu:
> I'd be tempted to put these in their own submodule, and not lump them into utils.

What would be a good name for it? ``testutils``, ``testsupport``, ...?

cocotb/utils.py
cmarqu: Via ``tests/test_cases/test_cocotb/test_doctests.py``, I see.

cmarqu:
For this to pass in Travis, the run with ``iverilog-v11-20190809-x64_setup.exe`` would have to be removed/replaced because it reports as "Icarus Verilog version 11.0 (devel)" but is too old to have https://github.com/steveicarus/iverilog/issues/323.

cocotb/utils.py
eric-wieser: Now that we have pytest it would make sense to remove that file, and do it the way you suggest :)

eric-wieser: (in another PR)

eric-wieser:
`cocotb.simulator.version.IcarusVersion`? We could move `cocotb.simulator` to `cocotb.simulator._gpi`, and reimport the members in `cocotb/simulator/__init__.py`. Thoughts, @ktbarrett?

cocotb/utils.py
cmarqu: Opened an issue for it, #1988.

ktbarrett:
@eric-wieser what were you thinking about how that would play with a user overrided simulatormodule (#1332)?

eric-wieser:
Overriding a module isn't really a done thing in python. More common would be either:
* Mock out methods in the module (usnig the builtin `mock`), which actually becomes easier if `cocotb.simulator` is a pure-python module that happens to re-export the extension methods
* create a cocotb.simulator.Simulator` class for which an instance is used by default, and have users provide a substitute instance for that instead.

ktbarrett:
@eric-wieser I like the idea of co-locating all the simulator stuff in `cocotb.simulator`, I'm just cautious so we don't code ourselves into a corner. It might be nice at the `cocotb.simulator` level to have all information about the *current* simulator available. So I like putting all the different versions in `cocotb.simulator.versions`, then in `cocotb.simulator` we can import the correct `Version` at `cocotb.simulator.version` after parsing the simulator name.

Not all of this has to be in this PR though, just enough to get this stuff in.

eric-wieser:
> we can import the correct `Version` at `cocotb.simulator.version` after parsing the simulator name.

This isn't a good idea though, we don't want `Version('some_icarus version')` to resolve to `QuestaVersion('some_icarus version')`.

I don't think users will ever need to know the version constructor for their current simulator.

On that note, we probably need to decide what `IcarusVersion < QuestaVersion` should do. Raising an exception is probably wise, but this patch silently proceeds anyway.

ktbarrett:
I think you misunderstood me, `cocotb.simulator.version` would be `<SimName>Version(<SimVersion>)`, there wouldn't be an immediate way for a user to break it. It could be used in the test exemption or by users for switching behaviors based on simulator version (probably not that common), but that's out of scope for this PR anyways.

>On that note, we probably need to decide what IcarusVersion < QuestaVersion should do. Raising an exception is probably wise, but this patch silently proceeds anyway.

This would probably come up doing the test exemption, I'm not sure what's right here. Throwing is obvious, but maybe not useful.

garmin-mjames:
Related to this, it would be nice to canonicalize the simulator name in Python, and not just use the raw `cocotb.SIM_NAME`.
It may be as simple as exporting **SIM** from the makefiles and capturing that in `cocotb.simulator.name` or similar.

Edit: that wouldn't cover non-Makefile flows though...

ktbarrett:
@garmin-mjames you could just use `os.getenv('SIM')`?

ktbarrett:
So the `SimVersion` classes exist. We are just trying to decide where to put them now? I agree that I don't think this should go in `cocotb.utils`. I think the suggestion @eric-wieser made is good, we maybe just need to mail down exactly what that looks like. @cmarqu What do you think? Do you want to expand the scope of this PR, shelve it until the changes are made in another PR, or think it's fine as is?

cmarqu:
I don't have a strong opinion, I lean toward shelving it for now. OTOH, with the Icarus 0.11 release, it would be useful.
Let's talk about it in the call tomorrow.

ktbarrett (commented):
I don't see a need to write up examples in the docstrings. You should prepend a `_` to all the names to show they are private internal use classes. Then no one will be surprised if/when we move them before 2.0 releases.

cmarqu:
> I don't see a need to write up examples in the docstrings.

These double as doctests (and are actually executed IIRC).

cocotb/utils.py
```diff
@@ -630,3 +614,134 @@ def extract_coro_stack(coro, limit=None):
...
+    b = a
+    for error in range(random.randint(2, 9)):
+        offset = random.randint(0, len(a))
+        b = b[:offset] + chr(random.randint(0, 255)) + b[offset+1:]
+
+    diff = hexdiffs(a.encode('latin1'), b.encode('latin1'))
+    print(diff)
+
+    space = '\n' + (" " * 20)
+    print(space.join(diff.split('\n')))
```
eric-wieser, line 747: ???

eric-wieser: Oh, I see this just moved from above. I think we can kill this with prejudice, or at least migrate it to a test.

cocotb/utils.py
```diff
@@ -630,3 +614,134 @@ def extract_coro_stack(coro, limit=None):
...
+
+class _CvcVersion(LooseVersion):
+    """
+    Version numbering class for Tachyon DA CVC.
+
+    Example:
+        >>> _CvcVersion("OSS_CVC_7.00b-x86_64-rhel6x of 07/07/14 (Linux-elf)") > _CvcVersion("OSS_CVC_7.00a-x86_64-rhel6x of 07/07/14 (Linux-elf)")
+        True
+    """
+    pass
```
eric-wieser, line 637: If we're declaring these not  (yet) for public consumption, then perhaps simpler to put them in a new file, `cocotb._sim_versions.ActivehdlVersion` - that is, make the module `_`-prefixed instead of the classes. That also leaves a good place to stick a docstring warning that these are not for public use.

eric-wieser (commented):
It's worth noting that these classes with silently allow `_GhdlVersion("1") > _IcarusVersion("1")`. That's probably OK for now, but something we'd want to fix before making them public.

cocotb/utils.py
ktbarrett: Please do.

cmarqu: Let's do that  in a different PR.

cocotb/utils.py
cmarqu: Done.

ktbarrett (commented):
`_sim_versions.py` isn't covered with a doctest now that it isn't in `utils.py`. You should add the path to `pytest.ini` now that #2237 is in. And you will need to rebase now that #2228 is in. Home stretch...

cmarqu:
> `_sim_versions.py` isn't covered with a doctest now that it isn't in `utils.py`. You should add the path to `pytest.ini` now that #2237 is in. And you will need to rebase now that #2228 is in. Home stretch...

Done.

ktbarrett:
This fixes the Icarus (master) regression on Mac, Linux, and Windows :tada:. Can you remove the `may_fail` from the GHA?

cmarqu:
> Can you remove the `may_fail` from the GHA?

Done.

cocotb/_sim_versions.py
```diff
@@ -0,0 +1,117 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""
+Classes to compare simulation versions.
+
+These are for cocotb-internal use only.
+Warning: these classes silently allow comparing versions of different simulators.
```
eric-wieser, line 9: I know it's not in sphinx, but:
```suggestion

.. warning::
    these classes silently allow comparing versions of different simulators.
```

ktbarrett (approved):

PR #1991, opened 2020-07-19
Merge simulatormodule.h into simulatormodule.cpp

### LINKED ISSUE #1973: Merge simulatormodule files
simulatormodule.h and simulatormodule.cpp can be merged into 1 file. simulatormodule.h is not a usable header file and the separation is redundant.

ktbarrett:
simulatormodule.h isn't a usable header file and the separation isn't necessary. Fixes #1973.


ktbarrett:
@eric-wieser Just copy-pasted. Moved one section directly before it's first usage, if that's not ideal we can refactor later.

PR #1997, opened 2020-07-21
Add support to GPI to determine if a handle is a port

ktbarrett:
Supercedes #1128. Implements the same features a slightly different way that 1. works, 2. works with the use of `SINGLETON_HANDLES`.

### Differences
The previous implementation set a port direction on the constructor, which in combination with the `SINGLETON_HANDLES` cache, caused handles to not be recorded as ports. This is handled by adding a method to determine portness after creation, rather make that decision during construction. This will prevent the original issue and allow lazy evaluation on a likely uncommonly used property.

Additionally, *how* the detection of portness is done is suspect, running `vpi_get(vpiType, handle)` can return multiple types given the same object depending on the type of handle to that object there is. And there is no way to test an object for portness in the VPI. Instead, the safest and most consistent way to check this is:
1. obtain the module using `vpi_get(vpiModule, handle)`.
2. `vpi_iterate(vpiPort, module)` to check all ports.
3. see if any ports in the module have the same name as you.

### Alternatives

One can get ports associated with objects using `vpi_get(vpiPorts, object)`. But the LRM doesn't say what ports are returned... Are they ports connected to the object? Are they all ports in a module? What does that say about the object's name? It's not very clear.

### WIP

- [ ] Check if individual bits get counted as ports
- [ ] Check if struct members get counted as ports
- [ ] Implement VPI
    - [X] ~~I'm not sure how well `vpiFullName` works in context of mixedlanguage but it's probably necessary to check if we have the same object. For instance you can have two separate object with the same name that have the same module, if they are in different scopes.~~ Use `vpi_compare_objects`. Funnily, this is not support by *any* of the FOSS simulators.
    - [ ] If all the implementations in VPI truly are the same, move implementation to function
    - [ ] running `vpi_iterate(vpiPort, module)` iff the module handle was obtained from a port via `vpi_get(vpiModule, port)` returns NULL on both Riviera and Questa. I'm not entirely sure why...
- [x] Implement reasonable default implementation of is_port in GpiObjHdl
- [X] Implement FLI
- [ ] Implement VHPI
- [ ] Add caching for performance reasons?

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -104,6 +104,24 @@ class SimHandleBase:
         :meta public:
         """
 
+    @property
+    def _is_port(self):
+        """
+        Whether the GPI object is a port or not.
+
+        :meta public:
+        """
+        return self._handle.is_port()
+
+    @property
+    def _port_direction(self):
+        """
+        The directionality of the port, if the object is a port, otherwise ``GPI_UNDEFINED``.
+
+        :meta public:
+        """
+        return self._handle.get_port_direction()
+
     def get_definition_name(self):
         return self._def_name
 
--- a/cocotb/share/include/gpi.h
+++ b/cocotb/share/include/gpi.h
@@ -228,6 +228,29 @@ int gpi_is_constant(gpi_sim_hdl gpi_hdl);
 // Determine whether an object is indexable
 int gpi_is_indexable(gpi_sim_hdl gpi_hdl);
 
+/** Determine whether an object is a port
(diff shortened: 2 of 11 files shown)


ktbarrett:
I think I will probably abandon this approach. Simulator support for many VPI commands is too poor. Note, that the FLI works perfectly.

Instead, this should be accomplished by supporting the notion of modules in the GPI object hierarchy, and the concept of parents and a hierarchical relationship of GPI objects. That should be more than enough information for more than one possible implementation.

PR #1998, opened 2020-07-22
Bump release notes via towncrier

ktbarrett:
Bumping the release notes by calling `towncrier` in root dir.

This was missed during the [release process](https://github.com/cocotb/cocotb/wiki/Release-Process).
You can see the issue [here](https://docs.cocotb.org/en/latest/release_notes.html). All of 1.4 changes are being listed as 1.5-dev changes.


eric-wieser:
Did any post-1.4 notes end up merged in here too?

imphil:
Can you please give a bit more background/motivation for this change?

eric-wieser:
Whenever you do a release, the idea is to run towncrier to lock in the final changelog.

eric-wieser:
The purpose of the separate files is just to avoid merge conflicts during development, it's not a substitute for the final release notes.

cmarqu:
This is Item 3 in https://github.com/cocotb/cocotb/wiki/Release-Process
Maybe this should move to Sphinx.

eric-wieser:
By "this", I assume you mean that document you link to.

ktbarrett:
@eric-wieser No changes brought in since the release have had release notes. Might be worthe checking to see if they did.

@imphil You can see the issue [here](https://docs.cocotb.org/en/latest/release_notes.html). The 1.4 release notes were never finalized and are being marked as 1.5.0-dev changes.

PR #1999, opened 2020-07-22
Bump release notes via towncrier for 1.4 release branch

ktbarrett:
Backport of #1998 to `stable/1.4` branch.


.azp-private.yml
```diff
@@ -22,6 +22,7 @@ resources:
     type: github
     endpoint: cocotb
     name: cocotb/cocotb-private-ci
+    ref: refs/heads/stable/1.4
```
eric-wieser, line 25: What's the purpose of this?

imphil: To use the CI job description corresponding to the stable branch, reducing potential breakage as we change private CI to match things on master. I didn't create the branch, but I like the idea.

eric-wieser: Ah, so this refers to a branch within `cocotb/cocotb-private-ci`. I'm surprised this works if the branch doesn't exist.

ktbarrett: Yeah I created the branch and made this commit after I was seeing CI not passing because of the VHDL tests which were added after the release.

Issue #1994, opened 2020-07-20
Support multiple GPI users


ktbarrett:
Currently, GPI depends on the embed library. The embed library is basically a required single-user of the GPI. The net affect of this choice is that multiple GPI users cannot coexist. There may also be issues with loading dependent libraries. This should help with the implementation of cocotb extensions that use or extend the GPI (like #1217).

Instead, I propose a way to have GPI users loaded dynamically by the GPI and offer them interfaces to register for callback on important events (like startup, shutdown, and other events). We already have a way to load additional libraries with the GPI that supports arbitrarily named entry points: `GPI_EXTRA`. We should probably use this for GPI users as well. Additionally, I would introduce the following interfaces on the GPI.

```c++
/**
 * Type of an entry function loaded by GPI_EXTRA.
 */
typedef void (gpi_extra_entry_func)(void);

/**
 * Type of a simulator startup event callback.
 * @param[in] userdata  State information passed to every call of the callback
 * @param[in] argc      Number of arguments in argv
 * @param[in] argv      List of argument strings of length argc
 * @returns             0 for successful initialization, non-0 for unsuccessful initialization
 */
typedef int (gpi_sim_startup_callback)(void *userdata, int argc, char const *const *argv);

/**
 * Type of a simulator shutdown event callback.
 * @param[in] userdata  State information passed to every call of the callback
 */
typedef void (gpi_sim_shutdown_callback)(void *userdata);

/**
 * Type of a simulator event callback.
 * This callback acts as a catch-all for all non-startup/shutdown events.
 * @param[in] userdata  State information passed to every call of the callback
 * @param[in] event     The type of event that occured
 * @param[in] msg       A message string giving detail to the event
 */
typedef void (gpi_sim_event_callback)(void *userdata, gpi_event_t event, char const *msg);

/**
 * Register a callback to be called during startup after elaboration.
 * @param[in] callback   The callback function to be called when startup event occurs
 * @param[in] userdata  State information to be passed to the callback whenever it is called
 */
void gpi_register_sim_startup_callback(gpi_sim_startup_callback *callback, void *userdata);

/**
 * Register a callback to be called during shutdown.
 * @param[in] callback   The callback function to be called when shutdown event occurs
 * @param[in] userdata  State information to be passed to the callback whenever it is called
 */
void gpi_register_sim_shutdown_callback(gpi_sim_shutdown_callback *callback, void *userdata);

/**
 * Register a callback to be called when a runtime simulator event occurs.
[...]

PR #493, opened 2016-10-20
Issue #10:  Logging Enhancements

leftink:
# Summary of Enhancements
- User has full control of logging configuration, i.e. handlers, formatters, levels, etc.
- Added a new level DEEP_DEBUG to allow two different levels of debug messages
- Like the python logging facility, added module level log functions, i.e. critical(...), info(...), etc
- Update SimLog to ensure consistent log naming and easily add keyword arguments into the format record
- Added a ColumnFormatter with several features that is now the base class of the default log formatters.  This formatter will allow users the ability to define their own columns for the logs if they want to.
- Added test_logging to demonstrate the features of the logging facility
# Logging Configuration

The user can now specify a logging configuration file to be loaded at start-up by setting the environment variable **COCOTB_LOG_FMT**.
The user has access to the logging configuration through `cocotb.log.configure(cfg)`, where cfg is one of the formats listed below.
## Configuration Formats
- INI format (see python logging fileConfig format)
  - cfg (str):  Name of file containing configuration
  - cfg (obj): File-Like object, i.e. StringIO containing the configuration
- DICT format (see python logging dictConfig format)
  - cfg (dict): A dictionary containing the configuration
  - cfg (str):  Name of file containing configuration encoded with JSON
  - cfg (obj): File-Like object, i.e. StringIO containing configuration encoded with JSON
  - If PyYAML available:
    - cfg (str):  Name of file containing configuration encoded with YAML
    - cfg (obj): File-Like object, i.e. StringIO containing configuration encoded with YAML

**Note**: dictConfig() was not available until Python 2.7 so the full power of using a dictionary configuration will not work, but I did implement the "incremental" configuration so even older versions of python can benefit from using a DICT configuration for setting logger and handler levels.
# Logging Levels

All existing log levels have been brought into the cocotb.log namespace so users should not need to import the python logging module at all.

Added a new Log Level **DEEP_DEBUG** and registered with the python logging module.  I have always found two levels of debugging to be extremely useful.  An example of where this could be used today would be to overhaul the debugging in the scheduler.  Currently there is another Environment variable to enable debugging in the scheduler, but that could be removed now and then current logging updated to have some basic deubug and then use the **DEEP_DEBUG** level for the information that will print a lot.  With the other changes, this will allow the users to have fine control when to enable/disable this extra logging.

[...]

Issue #130, opened 2014-05-13
System-call implementations do not handle format values


chiggs:
We've overloaded `$error` etc to allow us to provide a consistent logging mechanism.

Unfortunately this implementation doesn't handle format values appropriately.

tsengr0916:
Hi Stuart,
Any updates on this issue? This is actually a show stopper for our sim environment. We are unable to upgrade the Modelsim version because of this bug. And the new Modelsim version has the features that we need.

Thanks,
Richard

pgfarley:
It looks to me like this is unfortunately going to be a more involved fix than handling the arguments over to vasprintf . 

Issues:
- verilog $display type functions support verilog specific format characters like %m (hierarchical name) and and %T (simulation time). 
- As part of iterating over the vpiArgument we'll need to do the same Verilog to C type conversion  of the arguments that a simulator would have to do.
- It's not obvious to me how we can get the runtime iterated arguments into va_list to hand off for formatting.

So I think we'll need to implement the string interpolation/formatting ourselves similar to what a simulator would do if it were handling the call to $info, $error, etc.
 
I can work on this but, as it's a non trivial amount of new code, I wanted to run it by the list to see if anyone had a better idea.

eric-wieser:
We might have to port the extremely heavyweight implementation of `get_display` from `iverilog`:

https://github.com/steveicarus/iverilog/blob/9d0d1938dc66b715b1ac6693913a9e79d3cb401c/vpi/sys_display.c#L960-L1137

@pgfarley I think your last point is a non-issue, it can be solved with code like:
```C
// this doesn't use C varargs, but VPI iteration
char *str = format_using_verilog_semantics(args_from_vpi);

// no formatting or varargs needed here
gpi_log(..., "%s", str);
```

PR #2000, opened 2020-07-22
Remove Python dependency from gpilog library

### LINKED ISSUE #1993: Remove Python from gpilog
The Python logger implementation for the GPI library is located *in* the gpilog library. This means that users of the GPI that aren't going to use Python still must link against libpython (this dependency on libpython isn't listed in cocotb_build_libs.py, but it is a reality). Instead, to support non-Python GPI users we should change the gpilog interface to one where the implementation is registered instead of prescribed.

Desired interface for gpilog:
```c
/**
 * Named log levels.
 * Log levels can be any value, but only these values will be named when using the native logger.
 */
enum gpi_log_levels {
    GPIDebug = 10,
    GPIInfo = 20,
    GPIWarning = 30,
    GPIError = 40,
    GPICritical = 50
};

/**
 * Convenience functions for logging, see #gpi_log for more detail.
 */
#define LOG_DEBUG(...)     gpi_log("cocotb.gpi", GPIDebug,         __FILE__, __func__, __LINE__, __VA_ARGS__);
#define LOG_INFO(...)      gpi_log("cocotb.gpi", GPIInfo,          __FILE__, __func__, __LINE__, __VA_ARGS__);
#define LOG_WARN(...)      gpi_log("cocotb.gpi", GPIWarning,       __FILE__, __func__, __LINE__, __VA_ARGS__);
#define LOG_ERROR(...)     gpi_log("cocotb.gpi", GPIError,         __FILE__, __func__, __LINE__, __VA_ARGS__);
#define LOG_CRITICAL(...)  gpi_log("cocotb.gpi", GPICritical,      __FILE__, __func__, __LINE__, __VA_ARGS__); 

/**
 * A GPI logging handler function.
 * See #gpi_log for other params.
 *
 * @param[in] userdata  State data used by the logging handler
 * @returns boolean representing success (0) or failure (non-0). If there is a failure, `gpi_log` will fall back on the native logger.
 */
typedef int (gpi_log_handler_type)(
    void *userdata,
    const char *name,
    int level,
    const char *pathname,
    const char *funcname,
    long lineno,
    const char *msg,
    va_list args);

/**
 * Log a message.
 * If a user log handler object is set, uses the registered log handler, otherwise uses the native logger.
 * @param[in] name      Name of the logger
 * @param[in] level     Level at which to log a message at
 * @param[in] pathname  Name of the file where the call site is located
 * @param[in] funcname  Name of the function where the call site is located
 * @param[in] lineno    Line number of the call site
 * @param[in] msg       The message to log, uses C-style format specifier
 * @param[in] ...       Additional arguments; formatted at inserted in message according to format specifier in msg argument
 */
void gpi_log(const char *name, int level, const char *pathname, const char *funcname, long lineno, const char *msg, ...);

/**
 * Retrieve current custom log handler.
 * @param[out] handler  Custom log handler registered previously, or NULL if no handler was set
 * @param[out] userdata Custom log handler userdata registered previously, or NULL if no handler was set
 */
void gpi_get_log_handler(gpi_log_handler_type **handler, void **userdata);

/**
 * Set custom log handler.
 * @param[in] handler   Handler function to call when the GPI logs a message
 * @param[in] userdata  State data to pass to the handler function when logging a message
 */
void gpi_set_log_handler(gpi_log_handler_type *handler, void *userdata);

/**
 * Clear current custom log handler and use native logger
 */
void gpi_clear_log_handler(void);

/**
 * Builtin logger implementation, used as a fallback when no custom log handler is set.
 * See #gpi_log for details on parameters.
 */
void gpi_native_logger_log(
    const char *name,
    int level,
    const char *pathname,
    const char *funcname,
    long lineno,
    const char *msg,
    ...);

/**
 * Set minimum logging level of the builtin logger implementation.
 * If a logging request occurs where the logging level is lower than the level set by this function, it is not logged.
 * *Only* affects the native logger.
 * @param[in] level     Logging level
 * @return              Previous logging level
 */
int gpi_native_logger_set_level(int level);
```

gpilog will contain a default implementation, the "native" logger, which will just use stdio. This native logger will be able to log messages and have it's log level set. Users may register their own loggers using `gpi_set_log_handler` which takes a handler function and an opaque userdata pointer. This should be sufficient for gpi_embed to register a Python logger implementation after it is loaded and remove the Python dependency in gpilog. The only thing the user may do with the custom log handler is clear it, to continue using the native logger, or retrieve it. The custom logger will not respect the native logger's log level.

I had thought about merging gpilog into GPI, but this is not possible because cocotb_utils depends upon gpilog. If gpi and gpilog were merged, cocotb_utils would then depend upon gpi. However, gpi depends upon cocotb_utils, so we encounter a loop. Everything could be merged into gpi, but I feel leaving it parsed out for now is fine.

ktbarrett:
Closes #1993. This change moves the Python logger into the gpi_embed library and refactors the gpilog interface to handle *registering* the Python logger once it's ready to use. As explained in the aforementioned issue, this removes the dependency on Python from the gpilog library. Now, nothing in the actual GPI depends upon Python. This will allow other wrappers in potentially other languages to use the GPI.

I followed @eric-wieser's advice to maintain the current behavior of requiring custom log handlers to call the native logger manually if they fail to log themselves. This is more flexible, allowing the user to try other fallbacks if possible, and allows the user to print an additional diagnostic.

The best way to read these changes is to recognize that:
* gpi_logging.h was rewritten for clarity and a lot more documentation
* The Python logger and the two private static variables used by the logger were moved to gpi_embed
* The calls that set up and tear down the Python logger in gpi_embed have been updated
* gpi_embed now calls `gpi_set_log_handler` to register that Python logger implementation with the GPI
* All the support functions in gpi_logging were re-written
* The native logger implementation is still the same, save one bugfix regarding the level at which to log messages

### DIFF
--- a/cocotb/share/include/gpi_logging.h
+++ b/cocotb/share/include/gpi_logging.h
@@ -37,39 +37,183 @@
 #define GPILOG_EXPORT COCOTB_IMPORT
 #endif
 
+#include <cstdarg>
+
 #ifdef __cplusplus
-# define EXTERN_C_START extern "C" {
-# define EXTERN_C_END }
-#else
-# define EXTERN_C_START
-# define EXTERN_C_END
+extern "C" {
 #endif
 
-EXTERN_C_START
-
+/** Named logging level
+ *
+ *  Only logs at one of these log level will print a level name.
+ */
 enum gpi_log_levels {
-    GPIDebug = 10,
-    GPIInfo = 20,
-    GPIWarning = 30,
-    GPIError = 40,
-    GPICritical = 50
+    GPIDebug    = 10,   ///< Prints `DEBUG`    . Verbose information, useful for debugging
+    GPIInfo     = 20,   ///< Prints `INFO`     . Information about major events in the current program
+    GPIWarning  = 30,   ///< Prints `WARN`     . Encountered a recoverable bug, or information about surprising behavior
+    GPIError    = 40,   ///< Prints `ERROR`    . An unrecoverable error
+    GPICritical = 50    ///< Prints `CRITICAL` . An unrecoverable error, to be followed by immediate simulator shutdown
 };
 
+/** Logs a message at DEBUG log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
+#define LOG_DEBUG(...) \
+    gpi_log("cocotb.gpi", GPIDebug,         __FILE__, __func__, __LINE__, __VA_ARGS__);
 
-#define LOG_DEBUG(...)     gpi_log("cocotb.gpi", GPIDebug,         __FILE__, __func__, __LINE__, __VA_ARGS__);
-#define LOG_INFO(...)      gpi_log("cocotb.gpi", GPIInfo,          __FILE__, __func__, __LINE__, __VA_ARGS__);
-#define LOG_WARN(...)      gpi_log("cocotb.gpi", GPIWarning,       __FILE__, __func__, __LINE__, __VA_ARGS__);
-#define LOG_ERROR(...)     gpi_log("cocotb.gpi", GPIError,         __FILE__, __func__, __LINE__, __VA_ARGS__);
-#define LOG_CRITICAL(...)  gpi_log("cocotb.gpi", GPICritical,      __FILE__, __func__, __LINE__, __VA_ARGS__);
+/** Logs a message at INFO log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
+#define LOG_INFO(...) \
+    gpi_log("cocotb.gpi", GPIInfo,          __FILE__, __func__, __LINE__, __VA_ARGS__);
 
-GPILOG_EXPORT void set_log_handler(void *handler);
-GPILOG_EXPORT void clear_log_handler(void);
-GPILOG_EXPORT void set_log_filter(void *filter);
-GPILOG_EXPORT void clear_log_filter(void);
-GPILOG_EXPORT void set_log_level(enum gpi_log_levels new_level);
+/** Logs a message at WARN log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
+#define LOG_WARN(...) \
+    gpi_log("cocotb.gpi", GPIWarning,       __FILE__, __func__, __LINE__, __VA_ARGS__);
 
-GPILOG_EXPORT void gpi_log(const char *name, enum gpi_log_levels level, const char *pathname, const char *funcname, long lineno, const char *msg, ...);
+/** Logs a message at ERROR log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
+#define LOG_ERROR(...) \
+    gpi_log("cocotb.gpi", GPIError,         __FILE__, __func__, __LINE__, __VA_ARGS__);
 
-EXTERN_C_END
+/** Logs a message at CRITICAL log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
+#define LOG_CRITICAL(...) \
+    gpi_log("cocotb.gpi", GPICritical,      __FILE__, __func__, __LINE__, __VA_ARGS__);
+
+/** Logs a message at the given log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+    @param level The gpi_log_levels at this to log the message
+ */
+#define LOG_(level, ...) \
+    gpi_log("cocotb.gpi", level,      __FILE__, __func__, __LINE__, __VA_ARGS__);
+
+/** Type of a log handler function.
+    @param userdata  private implementation data registered with this function
+    @param name      Name of the logger
+    @param level     Level at which to log a message at
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-sprintf-style format specifier
+    @param args      Additional arguments; formatted and inserted in message according to format specifier in msg argument
+ */
+typedef void (gpi_log_handler_type)(
+    void *userdata,
+    const char *name,
+    int level,
+    const char *pathname,
+    const char *funcname,
+    long lineno,
+    const char *msg,
+    va_list args);
+
+/** Log a message using the currently registered log handler.
+    User is expected to populate all arguments to this function.
+    @param name      Name of the logger
+    @param level     Level at which to log a message at
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-sprintf-style format specifier
+    @param ...       Additional arguments; formatted and inserted in message according to format specifier in msg argument
+ */
+GPILOG_EXPORT void gpi_log(
+    const char *name,
+    int level,
+    const char *pathname,
+    const char *funcname,
+    long lineno,
+    const char *msg,
+    ...);
+
+/** Log a message using the currently registered log handler.
+    User is expected to populate all arguments to this function.
+    @param name      Name of the logger
+    @param level     Level at which to log a message at
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-sprintf-style format specifier
+    @param args      Additional arguments; formatted and inserted in message according to format specifier in msg argument
+ */
+GPILOG_EXPORT void gpi_vlog(
+    const char *name,
+    int level,
+    const char *pathname,
+    const char *funcname,
+    long lineno,
+    const char *msg,
+    va_list args);
+
+/** Retrieve the current log handler.
+    @param handler  Location to return current log handler. If no custom logger is register this will be `NULL`.
+    @param userdata Location to return log handler userdata
+ */
+GPILOG_EXPORT void gpi_get_log_handler(gpi_log_handler_type **handler, void **userdata);
+
+/** Set custom log handler
+    @param handler   Handler function to call when the GPI logs a message
+    @param userdata  Data to pass to the handler function when logging a message
+ */
+GPILOG_EXPORT void gpi_set_log_handler(gpi_log_handler_type *handler, void *userdata);
+
+/** Clear the current custom log handler and use native logger
+ */
+GPILOG_EXPORT void gpi_clear_log_handler(void);
+
+/** Log a message using the native log handler.
+    User is expected to populate all arguments to this function.
+    @param name      Name of the logger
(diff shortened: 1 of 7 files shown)


cocotb/share/include/gpi_logging.h
```diff
@@ -30,39 +30,183 @@
...
+    User is expected to populate all arguments to this function.
+    @param name      Name of the logger
+    @param level     Level at which to log a message at
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-style format specifier
+    @param args      Additional arguments; formatted at inserted in message according to format specifier in msg argument
+ */
+void gpi_log_v(
```
eric-wieser, line 137:
```suggestion
void gpi_vlog(
```
is a more common naming convention

cocotb/share/lib/gpi_log/gpi_logging.cpp
```diff
@@ -135,132 +197,10 @@ static void gpi_log_native_v(const char *name, enum gpi_log_levels level, const
...
-    /* We first check that the log level means this will be printed
-     * before going to the expense of formatting the variable arguments
-     */
-    if (!pLogHandler) {
-        gpi_log_native_v(name, level, pathname, funcname, lineno, msg, argp);
-        return;
-    }
-
-    if (level < local_level)
-        return;
```
eric-wieser, line 171: I don't see this in the new code - I think this was a pretty important optimization to avoid too many python/c boundaries

cmarqu:
@martinda You may be interested in this.

eric-wieser:
> this removes the dependency on Python from the gpilog library.

I have one reservation about whether this is useful in the long run - to solve #130, it's likely we'll want to call into a python implementation to let it handle formatting verilog messages (since the code will be easier to write than C++).

At any rate, it doesn't seem to do much harm to attempt to separate them for now.

cocotb/share/include/gpi_logging.h
```diff
@@ -30,39 +30,183 @@
...
+void gpi_clear_log_handler(void);
+
+/** Log a message using the native log handler.
+    User is expected to populate all arguments to this function.
+    @param name      Name of the logger
+    @param level     Level at which to log a message at
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-style format specifier
```
eric-wieser, line 169:
```suggestion
    @param msg       The message to log, uses C-sprintf-style format specifier
```

cocotb/share/lib/gpi_log/gpi_logging.cpp
ktbarrett: I tested just to see what affect this actually has.

#### This branch
total regression wall time: 2m2s
endian swapper example ratio: 6882.48 NS/S

#### Master
rotal regression wall time: 1m55s
endian swapper example ratio: 7716.36 NS/S

So it looks like 10% hit across the board, so it is significant. I'll add this back in. I'm wondering if there is a way to get the Python logger to automatically update the local level if the level is changed in Python.

eric-wieser: >  I'm wondering if there is a way to get the Python logger to automatically update the local level if the level is changed in Python.

#493 explored doing this, along with a large pile of other stuff. We'd have to go back to providing a non-empty logger subclass, which hooked into `setLevel` calls.

ktbarrett: The most recent commit added an implementation of this.

cocotb/share/include/gpi_logging.h
```diff
@@ -30,39 +30,183 @@
...
+
+/** Type of a log handler function.
+    @param userdata  private implementation data registered with this function
+    @param name      Name of the logger
+    @param level     Level at which to log a message at
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-sprintf-style format specifier
+    @param ...       Additional arguments; formatted at inserted in message according to format specifier in msg argument
```
garmin-mjames, line 96:
```suggestion
    @param args      Additional arguments; formatted and inserted in message according to format specifier in msg argument
```

"formatted and inserted" in the other doc comments below as well.

cocotb/share/include/gpi_logging.h
```diff
@@ -30,39 +30,183 @@
...
+void gpi_get_log_handler(gpi_log_handler_type **handler, void **userdata);
 
-EXTERN_C_END
+/** Set custom log handler
+    @param handler   Handler function to call when the GPI logs a message
+    @param userdata  Data to pass to the handler function when logging a message
+ */
+void gpi_set_log_handler(gpi_log_handler_type *handler, void *userdata);
+
+/** Clears the current custom log handler and use native logger
```
garmin-mjames, line 158:
```suggestion
/** Clear the current custom log handler and use native logger
```

cocotb/share/include/gpi_logging.h
```diff
@@ -113,7 +113,7 @@ typedef void (gpi_log_handler_type)(
     @param funcname  Name of the function where the call site is located
     @param lineno    Line number of the call site
     @param msg       The message to log, uses C-sprintf-style format specifier
-    @param ...       Additional arguments; formatted at inserted in message according to format specifier in msg argument
+    @param args      Additional arguments; formatted and inserted in message according to format specifier in msg argument
```
garmin-mjames, line 116: Should the non-"vlog" functions keep `...` instead of `args`?

ktbarrett: Yep.

cocotb/share/include/gpi_logging.h
```diff
@@ -30,39 +30,183 @@
...
-# define EXTERN_C_START
-# define EXTERN_C_END
+extern "C" {
 #endif
 
-EXTERN_C_START
-
+/** Named logging level
+ *
+ *  Only logs at one of these log level will print a level name.
```
eric-wieser, line 41: This isn't true:
* The GPI log implementer could choose to print other level names
* In python, `logging.addLevelName` can be used to associate new level names

I'd probably instead say something like
```suggestion
 *  These logging levels deliberately align with the builtin constants in the python logging module.
 *  Implementers of `gpi_log_handler_type` should emit human readable level names for these value, but may support other values
```

cocotb/share/include/gpi_logging.h
```diff
@@ -30,39 +30,183 @@
...
-    GPIDebug = 10,
-    GPIInfo = 20,
-    GPIWarning = 30,
-    GPIError = 40,
-    GPICritical = 50
+    GPIDebug    = 10,   ///< Prints `DEBUG`    . Verbose information, useful for debugging
+    GPIInfo     = 20,   ///< Prints `INFO`     . Information about major events in the current program
+    GPIWarning  = 30,   ///< Prints `WARN`     . Encountered a recoverable bug, or information about surprising behavior
+    GPIError    = 40,   ///< Prints `ERROR`    . An unrecoverable error
+    GPICritical = 50    ///< Prints `CRITICAL` . An unrecoverable error, to be followed by immediate simulator shutdown
```
eric-wieser, line 48: I'd recommend dropping the `Prints ...`, the logger can print anything it likes.

cocotb/share/include/gpi_logging.h
```diff
@@ -30,39 +30,183 @@
...
+ */
+#define LOG_CRITICAL(...) \
+    gpi_log("cocotb.gpi", GPICritical,      __FILE__, __func__, __LINE__, __VA_ARGS__);
+
+/** Logs a message at the given log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+    @param level The gpi_log_levels at this to log the message
+ */
+#define LOG_(level, ...) \
+    gpi_log("cocotb.gpi", level,      __FILE__, __func__, __LINE__, __VA_ARGS__);
```
eric-wieser, line 86: If you're adding this, can you implement all the other macros in terms of it?

cocotb_build_libs.py
```diff
@@ -245,7 +245,7 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
     libsim = Extension(
         os.path.join("cocotb", "simulator"),
         include_dirs=[include_dir],
-        libraries=["cocotbutils", "gpilog", "gpi"],
```
eric-wieser, line 248: This makes me a little uneasy - if I'm importing `cocotb.simulator` from a python terminal, I'd rather not pull in the shared library that tries to embed a python interpreter.

ktbarrett: This isn't an issue when importing cocotb from a Python interpreter, libpython is already loaded at that point (or if it's static, is already defined).

This is due to the fact that there is no shared library between the simulator module and the embedding library, the Python API to update the log level has to be in the extension module, and the Python logger implementation is located in the embed library.

If this is an issue, I can factor the Python logger out into another library. I'm also wondering if we could merge the extension module and the embedding library into one library to avoid the need for a 3rd utility library.

ktbarrett: @eric-wieser https://drive.google.com/file/d/1fFskjmbM695NIwschUPsGzr554n6AtZ1/view?usp=sharing

ktbarrett:
This is actually quite a bother. I'm punting this to some time in the future. Just like every major change...

ktbarrett:
Factored out the Python GPI logger into a separate library rather than sticking it in the embed library. Both simulatormodule and gpi_embed talk to this object.

cocotb_build_libs.py
ktbarrett: Went in another direction.

ktbarrett:
Since this is purely internal, I don't think I need to make a newsfragment or add additional documentation. Should I go ahead and merge the `gpi_logging` into `gpi`?

garmin-mjames (commented):
I think `libgpilog` can be merged into `libgpi`, but it should probably wait until after #1994 is resolved.

I unresolved some of @eric-wieser's old comments. It looked to me like they either hadn't been addressed or maybe the changes were lost in the refactor of this PR. If that's inaccurate go ahead and re-resolve them.

ktbarrett:
@garmin-mjames #2268 builds off of this and the two PRs are now out of sync since changes were not made in the appropriate PR :( I'm closing this PR, since apparently the other one is digestible enough. I will consider your comments here in the other one.

cocotb/share/include/gpi_logging.h
ktbarrett: The native logger only prints those words. While a custom logger (like a Python logger being registered and then messed with an in unsavory way) might do something different, we are documenting default behavior. I will add "by default" phrase.

PR #842, opened 2019-03-11
Add osx and more python versions to travis (linux)

themperek:
- Added python 3.6 and 3.7 Linux on travis.
- Add osx testing (2.7 and 3.6)

PR #2001, opened 2020-07-23
Fix Travis

ktbarrett:
The error is seen [here](https://travis-ci.org/github/cocotb/cocotb/jobs/710954784). I'm guessing this bit of code I'm removing is design to ensure we are using the correct version of Python. I'm not sure this is 100% necessary, but maybe it is. If it is we should use some other way to enforce this. Just seeing if this fixes the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,6 @@ jobs:
         - if [[ ! -e "./iverilog/README.txt" ]]; then rm -rf iverilog; git clone https://github.com/steveicarus/iverilog.git --depth=1 --branch v10_3; fi
         - cd iverilog && autoconf && ./configure && make -j2 && sudo make install && cd ..
         - pip install --upgrade tox virtualenv codecov
-        - pyenv global system $TRAVIS_PYTHON_VERSION
       script:
         - echo 'Running cocotb tests with Python 3.7 ...' && echo -en 'travis_fold:start:cocotb_py37'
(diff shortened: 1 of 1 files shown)


eric-wieser:
This was added in #842. I imagine there's a tox-travis thing of some kind to deal with this if it remains a problem.

ktbarrett:
@eric-wieser 
> I imagine there's a tox-travis thing of some kind to deal with this if it remains a problem.

[Hmmmmm...](https://github.com/tox-dev/tox-travis)

eric-wieser:
Just checked, looks fine

ktbarrett:
@eric-wieser Tox fails if it can't load the correct Python.

```
(cocotb) ~/dev/cocotb/tests/test_cases/test_discovery (port-support) 
> python --version
Python 3.8.3

(cocotb) ~/dev/cocotb/tests/test_cases/test_discovery (port-support) 
> tox -e py37
py37 create: /home/kaleb/dev/cocotb/.tox/py37
ERROR: InterpreterNotFound: python3.7
__________________________________________________________________________ summary ___________________________________________________________________________
ERROR:  py37: InterpreterNotFound: python3.7
```

We are calling out the python version in every tox call. 

In the future we could use [tox-travis](https://github.com/tox-dev/tox-travis) to automate that, but I didn't think it immediately necessary.

PR #2003, opened 2020-07-24
Fix build with gcc 10

tpambor:
Compiling with gcc 10 fails:
```
    cocotb/share/lib/vpi/VpiCbHdl.cpp: In constructor 'VpiIterator::VpiIterator(GpiImplInterface*, GpiObjHdl*)':
    cocotb/share/lib/vpi/VpiCbHdl.cpp:661:17: error: 'out_of_range' in namespace 'std' does not name a type
      661 |     catch (std::out_of_range const&) {
          |                 ^~~~~~~~~~~~
    cocotb/share/lib/vpi/VpiCbHdl.cpp:661:29: error: expected ')' before 'const'
      661 |     catch (std::out_of_range const&) {
          |           ~                 ^~~~~~
          |                             )
    cocotb/share/lib/vpi/VpiCbHdl.cpp:661:30: error: expected '{' before 'const'
      661 |     catch (std::out_of_range const&) {
          |                              ^~~~~
    cocotb/share/lib/vpi/VpiCbHdl.cpp:661:36: error: expected unqualified-id before ')' token
      661 |     catch (std::out_of_range const&) {
          |                                    ^
```
Fix by including `#include <stdexcept>`

### DIFF
--- a/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
+++ b/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
@@ -29,6 +29,7 @@
 #include "VhpiImpl.h"
 #include <limits>     // numeric_limits
 #include <cinttypes>  // fixed-size int types and format strings
+#include <stdexcept>
 
 namespace {
     using bufSize_type = decltype(vhpiValueT::bufSize);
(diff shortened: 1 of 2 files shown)


tpambor:
Probably this header was pulled in by one of the the other includes. Could be caused by this change [`beb0086`](https://github.com/gcc-mirror/gcc/commit/beb0086f592563ddd2b14444d4e2495b17b1d4bf).

PR #2006, opened 2020-07-24
Added checks for detecting when coroutine functions are passed into functions that expect coroutine objects.

### LINKED ISSUE #1593: Catch unstarted async coroutines where coroutines can be accepted
In `fork` we catch and provide an error message if the user passes a `cocotb.coroutine` object they did not call, we should do the same for `async` coroutines that weren't called. This is as simple as `inspect.iscoroutinefunction()` (I checked, it is available in 3.5).

I'll add this in with #1550.

andrewandrepowell:
Hello,

This PR closes #1593.

In all places where a coroutine object is expected, I added a check that determines whether or not the argument is a coroutine function, and if it is, a `TypeError` exception is raised with message indicating a coroutine object should be used. The code of concern is the following.

- `cocotb.scheduler.Scheduler.add()`
- `cocotb.decorators.RunningTask()`

The tests themselves are written such that they intentionally pass coroutine functions directly into the calls to trigger the exception. The exceptions are caught and asserted to ensure the correct exception is thrown.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -93,6 +93,11 @@ class RunningTask:
             self._natively_awaitable = True
         elif inspect.isgenerator(inst):
             self._natively_awaitable = False
+        elif inspect.iscoroutinefunction(inst):
+            raise TypeError(
+                "Coroutine function {} should be called prior to being "
+                "scheduled."
+                .format(inst))
         elif sys.version_info >= (3, 6) and inspect.isasyncgen(inst):
             raise TypeError(
                 "{} is an async generator, not a coroutine. "
--- a/documentation/source/newsfragments/1593.feature.rst
+++ b/documentation/source/newsfragments/1593.feature.rst
@@ -1,0 +1,1 @@
+:meth:`cocotb.fork` will now raise a descriptive :class:`TypeError` if a coroutine function is passed into them.
--- a/tests/test_cases/test_cocotb/common.py
+++ b/tests/test_cases/test_cocotb/common.py
@@ -45,10 +45,13 @@ def _check_traceback(running_coro, exc_type, pattern):
 
 
 @contextmanager
-def assert_raises(exc_type):
+def assert_raises(exc_type, pattern=None):
     try:
         yield
-    except exc_type:
+    except exc_type as e:
+        if pattern:
+            assert re.match(pattern, str(e)), \
+                "Correct exception type caught, but message did not match pattern"
         pass
     else:
         raise AssertionError("{} was not raised".format(exc_type.__name__))
--- a/tests/test_cases/test_cocotb/test_async_coroutines.py
+++ b/tests/test_cases/test_cocotb/test_async_coroutines.py
@@ -8,6 +8,7 @@ Test function and substitutablility of async coroutines
 import cocotb
 from cocotb.triggers import Timer
 from cocotb.outcomes import Value, Error
+from common import assert_raises
 
 
 class produce:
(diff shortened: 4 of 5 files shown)


tests/test_cases/test_cocotb/test_async_coroutines.py
```diff
@@ -135,3 +135,45 @@ async def example():
...
+@cocotb.test()
+async def test_task_coroutine_function_exception(dut):
+    async def coro(dut):
+        pass
+
+    try:
+        cocotb.decorators.RunningTask(coro)
+        assert False
+    except TypeError as e:
+        cocotb.log.info("error: {}".format(e))
```
ktbarrett, line 177: The log messages you have here are not so useful for unit tests. Perhaps knock them down to debug.

cocotb/regression.py
```diff
@@ -651,6 +651,12 @@ class TestFactory:
     """
 
     def __init__(self, test_function, *args, **kwargs):
+        if inspect.iscoroutinefunction(test_function):
+            raise TypeError(
+                "Coroutine function {} should be called prior to being "
+                "scheduled."
+                .format(test_function))
+
```
ktbarrett, line 659: We want coroutine functions here. Remove this entire section.

tests/test_cases/test_cocotb/test_async_coroutines.py
```diff
@@ -135,3 +135,31 @@ async def example():
...
     yield example()
     assert ran
+
+
+@cocotb.test()
+async def test_fork_coroutine_function_exception(dut):
+    async def coro():
+        pass
+
+    try:
```
eric-wieser, line 145: Can you use`with assert_raises` here instead, from `common.py`?

ktbarrett:
Does this deserve a newsfragment?

tests/test_cases/test_cocotb/test_async_coroutines.py
andrewandrepowell: Part the check requires converting the `TypeError` exception into string that I can compare against. I could wrap the `try-except` block with `assert_raises` if you still need me to include it.

ktbarrett: We could slightly change the definition of `assert_raises` to support this.

```python
@contextmanager
def assert_raises(exc_type, pattern=None):
    try:
        yield
    except exc_type:
        if pattern:
            assert re.match(pattern, str(exc_type)), "Correct exception type caught, but message did not match pattern"
    else:
        raise AssertionError("{} was not raised".format(exc_type.__name__))
```

cocotb/decorators.py
```diff
@@ -93,6 +93,11 @@ def __init__(self, inst):
             self._natively_awaitable = True
         elif inspect.isgenerator(inst):
             self._natively_awaitable = False
+        elif inspect.iscoroutinefunction(inst):
+            raise TypeError(
+                "Coroutine function {} should be called prior to being "
+                "scheduled."
+                .format(inst))
```
ktbarrett, line 100: I'm thinking this isn't necessary either. This is never called directly and all the uses type check the arguments before calling the constructor. I second check that.

tests/test_cases/test_cocotb/test_async_coroutines.py
eric-wieser: I think you mean `except exc_type as e`, `str(e)`, but otherwise that seems like a good idea to me.

cocotb/decorators.py
eric-wieser: Perhaps it's not, but it's consistent with how we already error on `isasyncgen` below, so I'm inclined to leave worrying about moving it to a later PR.

ktbarrett: Well ideally we would co-locate all the coroutine type checking stuff so we wouldn't have to keep repeating ourselves (inconsistently I'll add). I tried this in #1767, but I wasn't happy enough with the implementation to finish it out.

ktbarrett:
@andrewandrepowell You need to fix flake8 issues first.

documentation/source/newsfragments/1593.feature.rst
```diff
@@ -0,0 +1 @@
+:meth:`cocotb.scheduler.Scheduler.add` and :meth:`cocotb.decorators.RunningTask.__init__` will now raise a descriptive :class:`TypeError` if a coroutine function is passed into them.
```
ktbarrett, line 1: I'd not be too explicit, users aren't that aware of the internals. 
```suggestion
:meth:`cocotb.fork` will now raise a descriptive :class:`TypeError` if a coroutine function is passed into them.
```

tests/test_cases/test_cocotb/common.py
```diff
@@ -45,10 +45,13 @@ def _check_traceback(running_coro, exc_type, pattern):
 
 
 @contextmanager
-def assert_raises(exc_type):
+def assert_raises(exc_type, pattern=None):
```
cmarqu, line 48: I suppose at some point we may move (some of) this module out of test_cocotb to be more widely accessible, and then a docstring would be nice.

PR #2009, opened 2020-07-28
Verilator regression

### LINKED ISSUE #1996: Add Verilator to CI suite
We have Icarus Verilog already, and we support Verilator as a simulator.

It's free software. Is there any reason we couldn't add it?

ktbarrett:
Adds Verilator to Travis CI regression. Added just enough exceptions to the regressions to allow them to finish, there are still several failures. Closes #1996.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -155,6 +155,18 @@ jobs:
       script:
         - tox -e py37
 
+    - stage: SimTests
+      python: 3.7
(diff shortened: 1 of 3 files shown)


cmarqu:
In addition to Marlon's comment, I would also clone into a directory ``verilator_devel`` for some isolation that may be wanted later.

eric-wieser (approved):
Looks reasonable

PR #2011, opened 2020-08-03
Cleanup documentation for Trigger __init__ parameters.

garmin-mjames:
The parameters generated by `autodoc`'s `.. autoclass::` directive  don't match the source.

https://cocotb--2011.org.readthedocs.build/en/2011/triggers.html#simulator-triggers

https://docs.cocotb.org/en/latest/triggers.html#simulator-triggers

### DIFF
--- a/documentation/source/triggers.rst
+++ b/documentation/source/triggers.rst
@@ -22,11 +22,11 @@ Simulator Triggers
 Signals
 -------
 
-.. autoclass:: cocotb.triggers.Edge
+.. autoclass:: cocotb.triggers.Edge(signal)
 
-.. autoclass:: cocotb.triggers.RisingEdge
+.. autoclass:: cocotb.triggers.RisingEdge(signal)
 
-.. autoclass:: cocotb.triggers.FallingEdge
+.. autoclass:: cocotb.triggers.FallingEdge(signal)
 
 .. autoclass:: cocotb.triggers.ClockCycles
 
@@ -36,11 +36,11 @@ Timing
 
 .. autoclass:: cocotb.triggers.Timer
 
-.. autoclass:: cocotb.triggers.ReadOnly
+.. autoclass:: cocotb.triggers.ReadOnly()
 
-.. autoclass:: cocotb.triggers.ReadWrite
+.. autoclass:: cocotb.triggers.ReadWrite()
 
-.. autoclass:: cocotb.triggers.NextTimeStep
+.. autoclass:: cocotb.triggers.NextTimeStep()
 
 
 Python Triggers
@@ -50,7 +50,7 @@ Python Triggers
 
 .. autoclass:: cocotb.triggers.First
 
-.. autoclass:: cocotb.triggers.Join
+.. autoclass:: cocotb.triggers.Join(coroutine)
     :members: retval
 
 
--- a/cocotb/utils.py
+++ b/cocotb/utils.py
@@ -28,6 +28,7 @@
 """Collection of handy functions."""
 
 import ctypes
+import inspect
 import math
 import os
 import sys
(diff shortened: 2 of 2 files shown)


garmin-mjames:
New docs: https://cocotb--2011.org.readthedocs.build/en/2011/triggers.html

eric-wieser:
It would be better if we could fix this somehow - presumably `help(Edge)` still gives the useless help text?

garmin-mjames:
```
>>> help(Edge)
Help on class Edge in module cocotb.triggers:

class Edge(_EdgeBase)
 |  Edge(*args, **kwargs)
 |
 |  Fires on any value change of *signal*.
 |
 |  Method resolution order:
 |      Edge
 |      _EdgeBase
 |      GPITrigger
 |      Trigger
 |      collections.abc.Awaitable
 |      builtins.object
 |
 |  Data and other attributes defined here:
 |
 |  __abstractmethods__ = frozenset()
 |
 |  ----------------------------------------------------------------------
 |  Methods inherited from _EdgeBase:
 |
 |  __init__(self, signal)
 |      Initialize self.  See help(type(self)) for accurate signature.
 |
 |  __repr__(self)
 |      Return repr(self).
 |
 |  prime(self, callback)
 |      Register notification of a value change via a callback
 |
 |  ----------------------------------------------------------------------
 |  Class methods inherited from _EdgeBase:
 |
 |  __singleton_key__(signal) from cocotb.triggers._ParameterizedSingletonAndABC
 |
 |  ----------------------------------------------------------------------
 |  Data descriptors inherited from _EdgeBase:
 |
 |  signal
 |
 |  ----------------------------------------------------------------------
 |  Methods inherited from GPITrigger:
 |
 |  unprime(self)
 |      Disable a primed trigger, can be re-primed.
 |
 |  ----------------------------------------------------------------------
 |  Data descriptors inherited from GPITrigger:
 |
 |  cbhdl
 |
--More--
```

eric-wieser:
Implementing `__signature__` as a property in `ParametrizedSingleton` would likely fix both. I think you can do something like
```
@property
def __signature__(cls):
    return inspect.signature(cls.__singleton_key__)
```
although you may find you have to remove an argument from the signature before returning it.

garmin-mjames:
@eric-wieser Thanks! Better to fix it at the root cause.

eric-wieser:
Does that at least fix `help`? I think there might be an open sphinx issue about that not working for `sphinx`...

garmin-mjames:
It fixed `help` locally:
```
>>> help(triggers.Edge)
Help on class Edge in module cocotb.triggers:

class Edge(_EdgeBase)
 |  Edge(signal)
 |
 |  Fires on any value change of *signal*.
 |
 |  Method resolution order:
 |      Edge
 |      _EdgeBase
 |      GPITrigger
 |      Trigger
 |      collections.abc.Awaitable
 |      builtins.object
 |
 |  Data and other attributes defined here:
 |
 |  __abstractmethods__ = frozenset()
 |
 |  ----------------------------------------------------------------------
 |  Methods inherited from _EdgeBase:
 |
 |  __init__(self, signal)
 |      Initialize self.  See help(type(self)) for accurate signature.
 |
...
```

garmin-mjames:
@eric-wieser https://github.com/sphinx-doc/sphinx/issues/7613 ?

documentation/source/triggers.rst
```diff
@@ -36,11 +36,11 @@ Timing
...
 .. autoclass:: cocotb.triggers.Timer
 
-.. autoclass:: cocotb.triggers.ReadOnly
+.. autoclass:: cocotb.triggers.ReadOnly()
 
-.. autoclass:: cocotb.triggers.ReadWrite
+.. autoclass:: cocotb.triggers.ReadWrite()
 
-.. autoclass:: cocotb.triggers.NextTimeStep
+.. autoclass:: cocotb.triggers.NextTimeStep()
```
ktbarrett, line 43: I don't like that Sphinx leaves off the parameter list parens if there are no parameters, but `help()` does that as well, so I guess it is the way it is.

eric-wieser: Me neither: I'd encourage you to file a patch against sphinx to fix it.

PR #2019, opened 2020-08-14
Do not link explicitly with libstdc++

tpambor:
There are various implementations of the C++ standard library, most prominently `libstdc++`, `libc++`, [MS STL](https://github.com/microsoft/STL).


eric-wieser:
I wonder why this was added in the first place

PR #2020, opened 2020-08-15
Cleanup of cocotb_build_libs.py

tpambor:
Cleanup of `cocotb_build_libs.py` and fix for runtime loading of `libpython` on mac os


cocotb_build_libs.py
```diff
@@ -270,9 +270,17 @@ def _get_python_lib():
...
-        python_lib = "lib" + _get_python_lib_link() + "." + _get_lib_ext_name()
+        python_lib = "lib" + _get_python_lib_link() + "."
+        if sys.platform == "darwin":
+            python_lib = os.path.join(sysconfig.get_config_var("LIBDIR"), python_lib)
+            if os.path.exists(python_lib + "dylib"):
+                python_lib += "dylib"
+            else:
+                python_lib += "so"
+        else:
+            python_lib += "so"
```
ktbarrett, line 283: So on Linux and Windows, `PY_SO_LIB` is only the library name and we are depending upon the loader mechanisms to find libpython. However, on Mac OS we are giving the whole path to the library? Why can't we rely on the loader's search path on Mac OS? It probably deserves a comment in the source as well.

A follow on question is, why don't we provide a full path on Windows and Linux as well? I think eventually we will take #1739 in, and libpython will have to be runtime loaded. A hardcoded path to the library compiled into the source will work to consistently load libpython and avoid issue with loader search behavior. This isn't actionable for this PR (seeing as the PR is about cleanup), unless you want it to be.

eric-wieser: Nit: I think this is easier to follow as
```suggestion
        python_lib = "lib" + _get_python_lib_link()
        if sys.platform == "darwin":
            python_lib = os.path.join(sysconfig.get_config_var("LIBDIR"), python_lib)
            if os.path.exists(python_lib + ".dylib"):
                python_lib += ".dylib"
            else:
                python_lib += ".so"
        else:
            python_lib += ".so"
```

tpambor: Here is an overview of `libpython` loading as far as I understood:
- On Linux, Python <3.8: `libpython*.so` is loaded on startup, it is expected to be found in `LD_LIBRARY_PATH`. Probably could be changed to runtime loading easily:
```
def get_libraries(self, ext):
        return ext.libraries
```
- On Linux, Python >=3.8: `libpython*.so` is loaded at runtime, it is expected to be found in `LD_LIBRARY_PATH`
- On Windows, `python*.dll` is loaded on startup, it is expected to be found in `PATH`. Runtime loading on Windows can only be easily implemented for functions, all acceses to symbols exported in the python headers with `PyAPI_DATA` ([example](https://github.com/python/cpython/blob/master/Include/pyerrors.h#L73)) result in segfaults. To make it work `GetProcAddress(...)` would have to be called after loading the library for all symbols exported with `PyAPI_DATA`.
- On macOS, pre this PR, `libpython*.so` is loaded on startup, the path to the folder containing `libpython` is given via rpath.
- On macOS, post this PR, `libpython*.dylib` or `libpython*.so` is loaded at runtime, full path is passed to `PY_SO_LIB` because `dlopen()` does not use the rpath [1].
- On macOS, possible alternative, `libpython*.dylib` or `libpython*.so` is loaded at runtime, it is expected to be found in `LD_LIBRARY_PATH`. We require the user to adapt his `LD_LIBRARY_PATH` or we set it in the `Makefile`

[1] [https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dlopen.3.html](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dlopen.3.html)

ktbarrett: > Runtime loading on Windows can only be easily implemented for functions, all acceses to symbols exported in the python headers with PyAPI_DATA (example) result in segfaults. To make it work GetProcAddress(...) would have to be called after loading the library for all symbols exported with PyAPI_DATA.

I was afraid of this... That *really* throws a wrench in my plans. If runtime loading libpython on Windows is a no-go, I don't see the need to support runtime loading of libpython at all. We should figure out why #1417 didn't work. We have not historically depended upon LD_LIBRARY_PATH for good reason. Commercial simulators tend to monkey with it, perhaps even adding their own Python installation to the search path. It's brittle, which is why we have avoided it.

That being said, none of that prevents this PR from coming in.

ktbarrett: So after a meeting we decided a good direction is to stick with conda-only binary redistribution on Windows using whatever conda offers in that respect. We would like to avoid `LD_LIBRARY_PATH` in general, so we will use runtime-load libpython via absolute path on Linux and MacOS. Compiling it into `PY_SO_LIB` is an effective solution until we actually get around to distributing binary wheels.

I have no problems with this PR. Thanks @tpambor for working on this stuff; it's not fun, fairly difficult, but important.

PR #2022, opened 2020-08-19
Replace yield with await in most of the monitors and drivers

eric-wieser:
This should give a small performance boost to a handful of drivers.

I've avoided the `amba` AXI drivers because that would create unhelpful conflicts with #1248.

### DIFF
--- a/cocotb/drivers/__init__.py
+++ b/cocotb/drivers/__init__.py
@@ -72,8 +72,7 @@ class BitDriver:
         """Stop generating data."""
         self._cr.kill()
 
-    @cocotb.coroutine
-    def _cr_twiddler(self, generator=None):
+    async def _cr_twiddler(self, generator=None):
         if generator is None and self._generator is None:
             raise Exception("No generator provided!")
         if generator is not None:
@@ -86,10 +85,10 @@ class BitDriver:
             on, off = next(self._generator)
             self._signal <= 1
             for _ in range(on):
-                yield edge
+                await edge
             self._signal <= 0
             for _ in range(off):
-                yield edge
+                await edge
 
 
 class Driver:
@@ -143,8 +142,8 @@ class Driver:
         self._sendQ = deque()
 
     @coroutine
-    def send(self, transaction: Any, sync: bool = True, **kwargs: Any) -> None:
-        """Blocking send call (hence must be "yielded" rather than called).
+    async def send(self, transaction: Any, sync: bool = True, **kwargs: Any) -> None:
+        """Blocking send call (hence must be "awaited" rather than called).
 
         Sends the transaction over the bus.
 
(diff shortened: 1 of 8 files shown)


eric-wieser:
@corna points out in https://github.com/cocotb/cocotb/pull/1248#discussion_r472918507

> Isn't it going to break existing user code? E.g., in my test I have sometimes:
> 
> ```
> result = await First(axim.write(...), ClockCycles(...))
> ```
> 
> to check for a timeout in the write, however dropping `coroutine` on `write()` before #1846 is going to break it.

Which applies to the removed `@cocob.coroutine` decorators here.

It sees either we need #1846, or a variant of it that does the same thing but with a deprecation warning.

ktbarrett:
@eric-wieser Maybe it's prudent to leave the `cocotb.coroutine` decorators on the things that have had before, rather than blocking this PR on #1846.

eric-wieser:
Agreed

ktbarrett:
@eric-wieser I edited my comment, we only need that for public interfaces, the rest can stay as it is.

cocotb/drivers/__init__.py
```diff
@@ -252,45 +249,44 @@ def __init__(self, entity: SimHandleBase, name: Optional[str], clock: SimHandleB
...
             sync: Synchronize the transfer by waiting for a rising edge.
         """
         if sync:
-            yield RisingEdge(self.clock)
+            await RisingEdge(self.clock)
         self.bus <= transaction
 
     @coroutine
-    def _wait_for_signal(self, signal):
+    async def _wait_for_signal(self, signal):
```
eric-wieser, line 264: I think this is "protected", and called by downstream possibly-third-party drivers

cocotb/drivers/__init__.py
```diff
@@ -154,9 +153,9 @@ def send(self, transaction: Any, sync: bool = True, **kwargs: Any) -> None:
             **kwargs: Additional arguments used in child class'
                 :any:`_driver_send` method.
         """
-        yield self._send(transaction, None, None, sync=sync, **kwargs)
+        await self._send(transaction, None, None, sync=sync, **kwargs)
 
-    def _driver_send(self, transaction: Any, sync: bool = True, **kwargs: Any) -> None:
+    async def _driver_send(self, transaction: Any, sync: bool = True, **kwargs: Any) -> None:
```
ktbarrett, line 158: This could also an abstractmethod.

cocotb/monitors/__init__.py
```diff
@@ -126,15 +126,15 @@ def wait_for_recv(self, timeout=None):
...
                 return None
         else:
-            yield self._wait_event.wait()
+            await self._wait_event.wait()
 
         return self._wait_event.data
 
-    @coroutine
+    # this is not `async` so that we fail in `__init__` on subclasses without it
     def _monitor_recv(self):
```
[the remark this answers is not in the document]
eric-wieser: I agree, but didn't want to tackle that in this PR.

I suspect the answer is simply that abstract methods were hard to declare when supporting both python 2 and python 3.

PR #2024, opened 2020-08-19
Make error messages more consistent

cmarqu:

### DIFF
--- a/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
+++ b/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
@@ -190,7 +190,7 @@ int VhpiArrayObjHdl::initialise(std::string &name, std::string &fq_name) {
     }
 
     if (NULL == type) {
-        LOG_ERROR("Unable to get vhpiBaseType for %s", fq_name.c_str());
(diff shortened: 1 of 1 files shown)


eric-wieser (approved):
Looks fine. Might be sensible to pass in `vhpi` in as the logger name, to make it filterable in python - but this PR is fine without that.

ktbarrett:
@cmarqu Is there a reason only some of the logs are converted?

@eric-wieser The test would then need to be aware of the interface being used. Putting everything under `cocotb.gpi` abstracts that away. I agree it seems messy.

cmarqu:
> @cmarqu Is there a reason only some of the logs are converted?

I just noticed first a typo and then inconsistencies in this file, and didn't want to go much further - after all, I could do this for VPI and FLI and all files where ``LOG_*`` is used.

Also, ``LOG_ERROR`` may be printed for users out of the blue and thus warrants some more context, while ``LOG_DEBUG`` is printed if the user has enabled it explicitly and likely knows what she is expecting, and those messages often already mention "vhpi" in some form.
I see that I have missed some ``LOG_WARN`` and a single ``LOG_INFO`` in that file though. Next time. 💯

PR #2028, opened 2020-08-20
Enable assertion rewriting with pytest

### LINKED ISSUE #1449: Add some basic assert methods ?
```
assertEqual(a, b) 
assertNotEqual(a, b) 
assertTrue(x) 
assertFalse(x)
```

It would potentially make logging easier/better.

Based on https://docs.python.org/3/library/unittest.html#unittest.TestCase.debug

### LINKED ISSUE #2027: Use the pytest assertion rewriting hook
I think something like this in our cocotb entry point might work:
```python
from _pytest.config import Config
from _pytest.assertion import install_importhook
pytest_conf = Config.fromdictargs([], {})
install_importhook(pytest_conf)
```
or perhaps
```python
from _pytest.config import get_config
from _pytest.assertion import install_importhook
pytest_conf = get_config()
pytest_conf.parse([])  # pytest command line args
install_importhook(pytest_conf)
```

cc @imphil

eric-wieser:
This means that code like `assert dut.signal.value == 3` will now show the actual signal value when the assertion fails.

I think by default this only applies to files called `test_*.py` or `*_test.py`.

Close #2027. Close #1449.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -240,6 +240,23 @@ def _initialise_testbench(argv_):
 
     dut = cocotb.handle.SimHandle(handle)
 
+    try:
+        import pytest
+    except ImportError:
+        log.warning("Pytest not found, assertion rewriting will not occur")
+    else:
+        try:
+            # Install the assertion rewriting hook, which must be done before we
+            # import the test modules.
+            from _pytest.config import Config
+            from _pytest.assertion import install_importhook
+            pytest_conf = Config.fromdictargs([], {})
+            install_importhook(pytest_conf)
+        except Exception:
+            log.exception(
+                "Configuring the assertion rewrite hook using pytest {} failed. "
+                "Please file a bug report!".format(pytest.__version__))
+
     # start Regression Manager
     global regression_manager
     regression_manager = RegressionManager.from_discovery(dut)
--- a/documentation/source/newsfragments/2028.feature.rst
+++ b/documentation/source/newsfragments/2028.feature.rst
@@ -1,0 +1,2 @@
+If ``pytest`` is installed, its assertion-rewriting framework will be used to
+produce more informative tracebacks from the :keyword:`assert` statement.
--- a/tests/test_cases/test_cocotb/test_pytest.py
+++ b/tests/test_cases/test_cocotb/test_pytest.py
@@ -1,0 +1,22 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+""" Tests relating to pytest integration """
+
+import cocotb
+
+# pytest is an option dependency
+try:
+    import pytest
+except ImportError:
+    pytest = None
+
+
+@cocotb.test(skip=pytest is None)
+async def test_assertion_rewriting(dut):
+    """ Test that assertion rewriting hooks take effect in cocotb tests """
+    try:
+        assert 1 != 42
+    except AssertionError as e:
+        assert "42" in str(e), (
+            "Assertion rewriting seems not to work, message was {}".format(e))
--- a/tests/test_cases/test_cocotb/test_pytest.py
+++ b/tests/test_cases/test_cocotb/test_pytest.py
@@ -5,7 +5,7 @@
 
 import cocotb
 
-# pytest is an option dependency
+# pytest is an optional dependency
 try:
     import pytest
 except ImportError:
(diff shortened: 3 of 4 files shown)


cocotb/__init__.py
```diff
@@ -240,6 +240,23 @@ def _initialise_testbench(argv_):
...
+    except ImportError:
+        log.warning("Pytest not found, assertion rewriting will not occur")
+    else:
+        try:
+            # Install the assertion rewriting hook, which must be done before we
+            # import the test modules.
+            from _pytest.config import Config
+            from _pytest.assertion import install_importhook
+            pytest_conf = Config.fromdictargs([], {})
+            install_importhook(pytest_conf)
```
eric-wieser, line 254: The alternative suggestion in #2027 did not work, as it loaded the coverage settings from our `pytest.ini` used for non-simulator tests, and couldn't find the coverage extension.

ktbarrett:
Could we add some tests for this to the regressions/CI?

eric-wieser:
I suppose something like
```
try:
    assert x == y
except AssertionError as e:
    assert "some pytest text" in str(e)

File: cocotb/__init__.py [hunk 2]
```diff
@@ -240,6 +240,23 @@ def _initialise_testbench(argv_):
...
+        except Exception:
+            log.exception(
+                "Configuring the assertion rewrite hook using pytest {} failed. "
+                "Please file a bug report!".format(pytest.__version__))
+
```
ktbarrett, line 259: This isn't being covered and with these new warnings codecov is posting, it's getting really annoying. My best guess is that it's because we are starting coverage in this function. Should we move the coverage start code into a stub (that is `pragma: no cover`ed out) before loading the rest of the initialization?

tests/test_cases/test_cocotb/test_pytest.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@cocotb.test(skip=pytest is None)
+async def test_assertion_rewriting(dut):
+    """ Test that assertion rewriting hooks take effect in cocotb tests """
+    try:
+        assert 1 != 42
+    except AssertionError as e:
+        assert "42" in str(e), (
+            "Assertion rewriting seems not to work, message was {}".format(e))
```
ktbarrett, line 22: Does raising `AssertionError` with `assert` not chain by default? You might not need to add the original assertion message to this one.

cocotb/__init__.py
eric-wieser: Well, we can always just ignore those warnings - it's not like we edit the initialization code that often. I wouldn't worry about it too much.

tests/test_cases/test_cocotb/test_pytest.py
eric-wieser: It does, but I figured I'd play it safe. I guess it doesn't really matter, since no one will ever see this message.

ktbarrett (approved):

cocotb/__init__.py
```diff
@@ -240,6 +240,23 @@ def _initialise_testbench(argv_):
 
     dut = cocotb.handle.SimHandle(handle)
 
+    try:
+        import pytest
+    except ImportError:
+        log.warning("Pytest not found, assertion rewriting will not occur")
```
cmarqu, line 246: Maybe we could suggest ``pip install pytest`` here.

tests/test_cases/test_cocotb/test_pytest.py
```diff
@@ -0,0 +1,22 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+""" Tests relating to pytest integration """
+
+import cocotb
+
+# pytest is an option dependency
```
cmarqu, line 8:
```suggestion
# pytest is an optional dependency
```

cmarqu (approved):

cocotb/__init__.py
ktbarrett: Well... are we going for "optional", or "optional, but recommended"? I think we may have a runtime dependency on `pytest` in 2.0, so this is probably a good thing.

eric-wieser: Recommending `pip` in an error message is likely to get users into a trap who are using `conda`. The users who don't know how to fix it from this message alone are also the ones who are most likely to screw up when installing it.

documentation/source/newsfragments/2028.feature.rst
```diff
@@ -0,0 +1,2 @@
+If ``pytest`` is installed, its assertion-rewriting framework will be used to
```
ktbarrett, line 1: Link to pytest main page?

cocotb/__init__.py
ktbarrett: Should we be warning at all if it isn't loaded?

ktbarrett:
Re-running regressions.

ktbarrett (approved):

PR #2038, opened 2020-08-27
Fix GHDL Makefile for builds with multiple libraries

corna:
Commit a3fb4b5 changed the Makefile of GHDL, replacing a
`cd $(SIM_BUILD)` with the `--workdir` argument of GHDL, which
however changes only the path of the current WORK library.

Add the `-P$(SIM_DIR)` argument to specify the path for all the other
libraries.


themperek:
From GHDL manual: `Use option -P to specify where libraries other than WORK are placed`

It may be better if this is set by the user to whatever path he wants. Can one use `EXTRA_ARGS` to set this?

ktbarrett:
@themperek It would have to be passed to `COMPILE_ARGS` and `EXTRA_ARGS` or it would break. But I'm wondering when you would ever not want build artifacts in `SIM_BUILD`.

themperek:
@ktbarrett I think if something is not strictly needed and can be set by the user better leave to the user. I am not ghdl user but if this option exists I would expect is used for something. Not a very strong option.

corna:
Note that, without this PR, [cocotb's libraries support for GHDL](https://github.com/cocotb/cocotb/blob/master/cocotb/share/makefiles/simulators/Makefile.ghdl#L68-L69) is broken.

E.g.

```
VHDL_SOURCES_UTILS = util1.vhd util2.vhd util3.vhd
VHDL_SOURCES = top.vhd
```

`util*.vhd` files are analyzed and put in the `UTILS` library (`utils-obj93.cf` under `SIM_BUILD`), however, without `-P$(SIM_BUILD)`, this directory is not added to the libraries search path.

themperek:
@corna can you use `COMPILE_ARGS` to set those options?

corna:
Yes, but either we remove the libraries support or we merge this patch, as the current behavior is inconsistent (building the libraries but not including them).

themperek (approved):
@corna Now I understand. Sorry and thank you. 

All this makefile business is not great but ...

corna:
Thanks!

Issue #1642, opened 2020-04-12
Document deprecation and removal strategy


cmarqu:
We should document our deprecation and removal strategy. 

From https://github.com/cocotb/cocotb/pull/1496#pullrequestreview-387405087

"Only deprecate things in minor releases, and remove them in the next major release. (See https://symfony.com/doc/current/setup/upgrade_major.html#make-your-code-deprecation-free for a project that follows this principle"

cmarqu:
What we could do is for every ``.. deprecated::`` line, specify the version number where support is going to be removed.
This would require that we do use predictable future version numbers.

themperek:
Last supported versions would be easier.

cmarqu:
Would it really? We would have to write something like "Last version this is supported in is 1.x.y", since we cannot know how many x'es and y's may still come.

themperek:
>Would it really? We would have to write something like "Last version this is supported in is 1.x.y", since we cannot know how many x'es and y's may still come.

This is exactly what I get with some IC place & route tools.

cmarqu:
We should have a section like that in the docs of the 1.5 release: https://docs.antora.org/antora/2.3/whats-new/#caution-coming-soon

PR #2041, opened 2020-08-29
Deprecate Scoreboard and generators

### LINKED ISSUE #2017: cocotb 1.4 issue in scoreboard with hexdiffs and hexdumps
Since release 1.4 where strings passed to `cocotb.utils.hexdump()` and `cocotb.utils.hexdiffs()` are deprecated, I found out that I experienced issues with the scoreboard results when tests fail. 
Although all my tests are updated to work with bytes instead of string data, it seems that inside the scoreboard code the `got` and `exp` data is still internally converted to strings before the `cocotb.utils.hexdump()` and `cocotb.utils.hexdiffs()` are called. It results in a hanging behaviour of the simulator. The same applies for the `log.info()` where expected and received data is logged inside this piece of code.

ktbarrett:
Fixes #2017. I don't have a test, seeing as we aren't currently testing the scoreboard, and may move it out of repo soon, I'm not sure if that's necessary. I will make one if it is desired.

@hjwmedenblik This should fix the issue you were seeing with hexdiff.
@bluecmd This should remove the DeprecationWarning you are seeing.

Could you both please test it?

### DIFF
--- a/cocotb/scoreboard.py
+++ b/cocotb/scoreboard.py
@@ -30,6 +30,7 @@
 """Common scoreboarding capability."""
 
 import logging
+import warnings
 
 from cocotb.utils import hexdump, hexdiffs
 from cocotb.log import SimLog
@@ -57,6 +58,8 @@ class Scoreboard:
         fail_immediately (bool, optional): Raise :any:`TestFailure`
             immediately when something is wrong instead of just
             recording an error. Default is ``True``.
+
+    .. deprecated:: 1.4.1
     """
 
     def __init__(self, dut, reorder_depth=0, fail_immediately=True):  # FIXME: reorder_depth needed here?
@@ -65,6 +68,12 @@ class Scoreboard:
         self.errors = 0
         self.expected = {}
         self._imm = fail_immediately
+
+        warnings.warn(
+            "This Scoreboard implementation has been deprecated and will be removed soon.\n"
+            "If this implementation works for you, copy the implementation into your project, "
+            "while following cocotb's license agreement.",
+            DeprecationWarning)
 
     @property
     def result(self):
--- a/examples/endian_swapper/tests/generators/__init__.py
+++ b/examples/endian_swapper/tests/generators/__init__.py
@@ -31,18 +31,8 @@
 import math
 import random
 import itertools
-import warnings
 
 from cocotb.decorators import public
-
-
-warnings.warn(
-    "The contents of the cocotb.generators package will soon be removed.\n"
-    "Most of the functionality can be replaced with utilities provided "
-    "by other packages or the standard library.\n Alternatively, you can "
-    "copy this package or individual functions into your project, if you "
-    "follow cocotb's license agreement.",
-    DeprecationWarning)
 
 
 @public
@@ -51,9 +41,9 @@ def repeat(obj, nrepeat=None):
 
     Args:
         obj (any): The object to yield
-        nrepeat (int, optional): The number of times to repeatedly yield *obj*
 
-    .. deprecated:: 1.4.1
+    Kwargs:
+        nrepeat (int): The number of times to repeatedly yield obj
     """
     if nrepeat is None:
         return itertools.repeat(obj)
@@ -68,8 +58,6 @@ def combine(generators):
 
     Args:
         generators (iterable): Generators to combine together
-
-    .. deprecated:: 1.4.1
     """
     return itertools.chain.from_iterable(generators)
 
@@ -82,9 +70,7 @@ def gaussian(mean, sigma):
     Args:
         mean (int/float): mean value
 
-        sigma (int/float): Standard deviation
-
-    .. deprecated:: 1.4.1
+        signma (int/float): Standard deviation
     """
     while True:
         yield random.gauss(mean, sigma)
@@ -96,14 +82,12 @@ def sine_wave(amplitude, w, offset=0):
     Generates a sine wave that repeats forever
 
     Args:
-        amplitude (int/float): peak deviation of the function from zero
+        amplitude (int/float):  peak deviation of the function from zero
 
         w (int/float): is the rate of change of the function argument
 
     Yields:
         floats that form a sine wave
-
-    .. deprecated:: 1.4.1
     """
     twoPiF_DIV_sampleRate = math.pi * 2
     while True:
@@ -112,11 +96,9 @@ def sine_wave(amplitude, w, offset=0):
 
 
 def get_generators(module):
-    """Return an iterator which yields all the generators in a module
+    """Return an iterator which yields  all the generators in a module
 
     Args:
         module (python module): The module to get the generators from
-
-    .. deprecated:: 1.4.1
     """
     return (getattr(module, gen) for gen in module.__all__)
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,7 +81,6 @@ jobs:
       after_success: *travis_linux_post
 
     - stage: SimTests
-      python: 3.7
       os: windows
       language: shell
       before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,6 @@ jobs:
         - if [[ ! -e "./iverilog/README.txt" ]]; then rm -rf iverilog; git clone https://github.com/steveicarus/iverilog.git --depth=1 --branch v10_3; fi
         - cd iverilog && autoconf && ./configure && make -j2 && sudo make install && cd ..
         - pip install --upgrade tox virtualenv codecov
-        - pyenv global system $TRAVIS_PYTHON_VERSION
       script:
         - echo 'Running cocotb tests with Python 3.7 ...' && echo -en 'travis_fold:start:cocotb_py37'
         - tox -e py37
@@ -124,7 +123,6 @@ jobs:
         - git clone https://github.com/ghdl/ghdl.git --depth=1 --branch v0.36
         - cd ghdl && ./configure && make -j2 && sudo make install && cd ..
         - pip install tox
-        - pyenv global system $TRAVIS_PYTHON_VERSION
       script:
         - tox -e py37
 
--- a/cocotb/generators/__init__.py
+++ b/cocotb/generators/__init__.py
@@ -40,9 +40,7 @@ def repeat(obj, nrepeat=None):
 
     Args:
         obj (any): The object to yield
-
-    Kwargs:
-        nrepeat (int): The number of times to repeatedly yield obj
+        nrepeat (int, optional): The number of times to repeatedly yield *obj*
     """
     if nrepeat is None:
         return itertools.repeat(obj)
@@ -69,7 +67,7 @@ def gaussian(mean, sigma):
     Args:
         mean (int/float): mean value
 
-        signma (int/float): Standard deviation
+        sigma (int/float): Standard deviation
     """
     while True:
         yield random.gauss(mean, sigma)
@@ -81,7 +79,7 @@ def sine_wave(amplitude, w, offset=0):
     Generates a sine wave that repeats forever
 
     Args:
-        amplitude (int/float):  peak deviation of the function from zero
+        amplitude (int/float): peak deviation of the function from zero
 
         w (int/float): is the rate of change of the function argument
 
@@ -95,7 +93,7 @@ def sine_wave(amplitude, w, offset=0):
 
 
 def get_generators(module):
-    """Return an iterator which yields  all the generators in a module
+    """Return an iterator which yields all the generators in a module
 
     Args:
         module (python module): The module to get the generators from
--- a/cocotb/generators/bit.py
+++ b/cocotb/generators/bit.py
@@ -32,7 +32,7 @@
     cycles onto a bus.
 
     These yield a tuple which is intended to be interpreted as a number of
-    cycles (ON,OFF)
+    cycles ``(ON,OFF)``
 """
 from cocotb.decorators import public
 from cocotb.generators import gaussian, sine_wave, repeat
@@ -43,7 +43,6 @@ def bit_toggler(gen_on, gen_off):
 
     Args:
         gen_on (generator): generator that yields number of cycles on
-
         gen_off (generator): generator that yields number of cycles off
     """
     for n_on, n_off in zip(gen_on, gen_off):
@@ -54,10 +53,9 @@ def bit_toggler(gen_on, gen_off):
 def intermittent_single_cycles(mean=10, sigma=None):
     """Generator to intermittently insert a single cycle pulse
 
-    Kwargs:
-        mean (int):     Average number of cycles in between single cycle gaps
-
-        sigma (int):    Standard deviation of gaps.  mean/4 if sigma is None
+    Args:
+        mean (int, optional): Average number of cycles in between single cycle gaps
+        sigma (int, optional): Standard deviation of gaps.  mean/4 if sigma is None
     """
     if sigma is None:
         sigma = mean / 4.0
@@ -68,10 +66,10 @@ def intermittent_single_cycles(mean=10, sigma=None):
 @public
 def random_50_percent(mean=10, sigma=None):
     """50% duty cycle with random width
-    Kwargs:
-        mean (int):     Average number of cycles on/off
 
-        sigma (int):    Standard deviation of gaps.  mean/4 if sigma is None
+    Args:
+        mean (int, optional): Average number of cycles on/off
+        sigma (int, optional): Standard deviation of gaps.  mean/4 if sigma is None
     """
     if sigma is None:
         sigma = mean / 4.0
--- a/cocotb/generators/byte.py
+++ b/cocotb/generators/byte.py
@@ -29,7 +29,7 @@
 
 
 """
-    Collection of generators for creating byte streams
+    Collection of generators for creating byte streams.
 
     Note that on Python 3, individual bytes are represented with integers.
 """
@@ -44,7 +44,7 @@ def get_bytes(nbytes: int, generator: Iterator[int]) -> bytes:
     """Get nbytes from generator
 
     .. versionchanged:: 1.4.0
-        This now returns :type:`bytes` not :type:`str`. """
+        This now returns :class:`bytes`, not :class:`str`. """
     return bytes(next(generator) for i in range(nbytes))
 
 
@@ -53,7 +53,7 @@ def random_data() -> Iterator[int]:
     r"""Random bytes
 
     .. versionchanged:: 1.4.0
-        This now returns integers not single-character :type:`str`\ s. """
+        This now returns integers, not single-character :class:`str`\ s. """
     while True:
         yield random.randrange(256)
 
@@ -63,7 +63,7 @@ def incrementing_data(increment=1) -> Iterator[int]:
     r"""Incrementing bytes
 
     .. versionchanged:: 1.4.0
-        This now returns integers not single-character :type:`str`\ s. """
+        This now returns integers, not single-character :class:`str`\ s. """
     val = 0
     while True:
         yield val
--- a/cocotb/generators/byte.py
+++ b/cocotb/generators/byte.py
@@ -41,29 +41,41 @@ from typing import Iterator
 
 @public
 def get_bytes(nbytes: int, generator: Iterator[int]) -> bytes:
-    """Get nbytes from generator
+    """
+    Get nbytes from generator
 
     .. versionchanged:: 1.4.0
-        This now returns :class:`bytes`, not :class:`str`. """
+        This now returns :class:`bytes`, not :class:`str`.
+
+    .. deprecated:: 1.4.1
+    """
     return bytes(next(generator) for i in range(nbytes))
 
 
 @public
 def random_data() -> Iterator[int]:
-    r"""Random bytes
+    r"""
+    Random bytes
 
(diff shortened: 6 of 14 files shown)


bluecmd:
Thanks!

It seems to work for the case where you are not getting something you expected, but for the case where you get something unexpected I still have the deprecation warning.

Good case:
```
 50000.00ns INFO     cocotb.fejkon_fc_debug.st_in                  avalon.py:219  in _monitor_recv                   Received a packet of 7 bytes
 50000.00ns INFO     cocotb.fejkon_fc_debug.st_in                    test.py:27   in model                           0000   00 01 02 03 04 05 06                              .......
                                                                                                                     
 60000.00ns WARNING  cocotb.scoreboard.fejkon_fc_debug         scoreboard.py:86   in result                          Still expecting 1 transactions on AvalonSTPkts(st_out)
 60000.00ns INFO     cocotb.scoreboard.fejkon_fc_debug         scoreboard.py:89   in result                          Expecting 0:
                                                                                                                     0000   00 01 02 03 04 05 06                              .......
                                                                                                                     
 61000.00ns ERROR    cocotb.regression                         regression.py:396  in _score_test                     Test Failed: test_passthrough (result was TestFailure)

```

Still broken case:
```
 50000.00ns INFO     cocotb.fejkon_fc_debug.st_in                  avalon.py:219  in _monitor_recv                   Received a packet of 7 bytes
 50000.00ns INFO     cocotb.fejkon_fc_debug.st_in                    test.py:27   in model                           0000   00 01 02 03 04 05 06                              .......
                                                                                                                     
 60000.00ns INFO     cocotb.fejkon_fc_debug.st_out                 avalon.py:219  in _monitor_recv                   Received a packet of 7 bytes
 60000.00ns ERROR    ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:240  in check_received_transaction      Received a transaction but wasn't expecting anything
/home/bluecmd/cocotb/cocotb/scoreboard.py:242: DeprecationWarning: Passing a string to hexdump is deprecated, pass bytes instead
  log.info("Got: %s" % (hexdump(str(transaction))))
 60000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:242  in check_received_transaction      Got: 0000   62 27 5C 78 30 30 5C 78 30 31 5C 78 30 32 5C 78   b'\x00\x01\x02\x
                                                                                                                     0010   30 33 5C 78 30 34 5C 78 30 35 5C 78 30 36 27      03\x04\x05\x06'
                                                                                                                     
 60000.00ns INFO     ..Pkts._monitor_recv.0x7f9b68aef860        scheduler.py:503  in unschedule                      Test stopped by this forked coroutine

```

I think the issue is here:

https://github.com/cocotb/cocotb/blob/656b8cab0392c8d8b6327a65833c5495cfe294d1/cocotb/scoreboard.py#L240-L244

ktbarrett:
@bluecmd I missed that one. It's been fixed.

eric-wieser (approved):
Looks at least better than before. Will still blow up if used for non-byte-like types (ones rejected by `b"%b".__mod__`), but who knows how this was supposed to be used anyway.

cocotb/scoreboard.py
```diff
@@ -128,19 +128,18 @@ def compare(self, got, exp, log, strict_type=True):
...
-            # Try our best to print out something useful
-            strgot, strexp = str(got), str(exp)
-
             log.error("Received transaction differed from expected output")
             if not strict_type:
-                log.info("Expected:\n" + hexdump(strexp))
+                log.info("Expected:\n" + hexdump(exp))
             else:
                 log.info("Expected:\n" + repr(exp))
             if not isinstance(exp, str):
```
bluecmd, line 145: This part here is quite weird as well, it should probably be removed. It tries to iterate over things that are not strings in "words" and printing them. This results in:

```
 90000.00ns ERROR    ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:140  in compare                         Received transaction differed from expected output
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:144  in compare                         Expected:
                                                                                                                     b'\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef'
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         222
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         173
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         190
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         239
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         222
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         173
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         190
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         239
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         222
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         173
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         190
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         239
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         222
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         173
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         190
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         239
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         222
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         173
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         190
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         239
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         222
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         173
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         190
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         239
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         222
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         173
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         190
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         239
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         222
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         173
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         190
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         239
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:154  in compare                         Received:
                                                                                                                     b'\xde\xad\xbe\xef'
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:158  in compare                         222
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:158  in compare                         173
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:158  in compare                         190
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:158  in compare                         239
```

bluecmd (changes requested):

cocotb/scoreboard.py
bluecmd:
```patch
diff --git a/cocotb/scoreboard.py b/cocotb/scoreboard.py
index f8a23d24..36d664d4 100755
--- a/cocotb/scoreboard.py
+++ b/cocotb/scoreboard.py
@@ -142,22 +142,10 @@ class Scoreboard:
                 log.info("Expected:\n" + hexdump(exp))
             else:
                 log.info("Expected:\n" + repr(exp))
-            if not isinstance(exp, str):
-                try:
-                    for word in exp:
-                        log.info(str(word))
-                except Exception:
-                    pass
             if not strict_type:
                 log.info("Received:\n" + hexdump(got))
             else:
                 log.info("Received:\n" + repr(got))
-            if not isinstance(got, str):
-                try:
-                    for word in got:
-                        log.info(str(word))
-                except Exception:
-                    pass
             log.warning("Difference:\n%s" % hexdiffs(exp, got))
             if self._imm:
                 raise TestFailure("Received transaction differed from expected "
```

That makes the output 100% nice in my use case.

```
 90000.00ns ERROR    ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:140  in compare                         Received transaction differed from expected output
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:144  in compare                         Expected:
                                                                                                                     b'\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef\xde\xad\xbe\xef'
 90000.00ns INFO     ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:148  in compare                         Received:
                                                                                                                     b'\xde\xad\xbe\xef'
 90000.00ns WARNING  ..scoreboard.fejkon_fc_debug.st_out       scoreboard.py:149  in compare                         Difference:
                                                                                                                     0000      DEADBEEFDEADBEEF DEADBEEFDEADBEEF ................
                                                                                                                          0000 DEADBEEF                          ....            
                                                                                                                     0004      DEADBEEFDEADBEEF DEADBEEFDEADBEEF ................
                                                                                                                     
 90000.00ns INFO     ..Pkts._monitor_recv.0x7f624ba9dc18        scheduler.py:503  in unschedule                      Test stopped by this forked coroutine

```

ktbarrett:
@bluecmd I doubt the scoreboard was written to work with arbitrary types. I think it was only ever meant for use with BinaryValue and bytes objects. The behavior given any other type is not consistent.

ktbarrett:
Cherry-picked a commit that should prevent Travis from failing.

cocotb/scoreboard.py
eric-wieser: I think the intent here was to give (only slightly) helpful outputs for lists, but given that blows up on the `hexdiffs` line anyway, I'd say we don't care. Your patch looks fine to me.

ktbarrett: Alternatively, we could change `isinstance(exp, str)` to `isinstance(exp, bytes)`.

bluecmd:
@ktbarrett I am giving it a bytes() type. I understand if you don't want to change too much in this PR though.

cocotb/scoreboard.py
eric-wieser: Yeah, that would work fine too - either way, not much of the scoreboard makes sense for non-bytes types anyway.

ktbarrett:
@bluecmd It is an issue of using `str` where `bytes` are actually expected. I'm not trying to change behavior of the scoreboard in this commit, just cleanup the whole bytes/str confusion.

bluecmd:
@ktbarrett Sounds good, I will open a PR for that after you merge this. I believe that part of code should just be removed, so I will propose it.

ktbarrett:
@bluecmd Does the newest change prevent the issue you were seeing?

bluecmd:
Yep, works wonderfully!

bluecmd (approved):

hjwmedenblik:
@ktbarrett  Hi Kaleb, it does not crash but I am still not happy with this piece:
https://github.com/ktbarrett/cocotb/blob/fix-scoreboard/cocotb/scoreboard.py#L141-L144

Why not do the hexdump on those received strings instead of no meaningful and hard to read string repr for humans

eric-wieser:
Frankly `strict_type=False` seems like something we never should have supported anyway.

I'm inclined to merge this as is to stop the bleeding, and then throw out the whole `Scoreboard` and start afresh later.

hjwmedenblik:
I also never understood the meaning for these lines:

https://github.com/ktbarrett/cocotb/blob/fix-scoreboard/cocotb/scoreboard.py#L155-L160

hjwmedenblik:
This was my homemade patch:
``
        # Compare the types
        if strict_type and type(got) != type(exp):
            self.errors += 1
            log.error("Received transaction type is different than expected")
            log.info("Received: %s but expected %s" %
                     (str(type(got)), str(type(exp))))
            if self._imm:
                raise TestFailure("Received transaction of wrong type. "
                                  "Set strict_type=False to avoid this.")
            return
        # Or convert to a string before comparison
        elif not strict_type:
            got, exp = str(got), str(exp)

        # Compare directly
        if got != exp:
            self.errors += 1

            # Try our best to print out something useful
            strgot, strexp = str(got), str(exp)

            log.error("Received transaction differed from expected output")
            log.info("Expected:\n" + hexdump(exp))

            log.info("Received:\n" + hexdump(got))
            log.warning("Difference:\n%s" % hexdiffs(exp, got))
            if self._imm:
                raise TestFailure("Received transaction differed from expected "
                                  "transaction")
        else:
            # Don't want to fail the test
            # if we're passed something without __len__
            try:
                log.debug("Received expected transaction %d bytes" %
                          (len(got)))
                log.debug(repr(got))
            except Exception:
                pass
``

eric-wieser:
> I also never understood the meaning for these lines:

I think they're for when the monitor returns lists instead of bytes.

hjwmedenblik:
> > I also never understood the meaning for these lines:
> 
> I think they're for when the monitor returns lists instead of bytes.

But still then what additional information does it provide the user next to expected or received which is printed out before that?
It clutters the output of the scoreboard.

ktbarrett:
> @ktbarrett Hi Kaleb, it does not crash but I am still not happy with this piece:
> 
> https://github.com/ktbarrett/cocotb/blob/fix-scoreboard/cocotb/scoreboard.py#L141-L144
> and
> https://github.com/ktbarrett/cocotb/blob/fix-scoreboard/cocotb/scoreboard.py#L151-L154
> 
> Why not do the hexdump on those received strings instead of no meaningful and hard to read string repr for humans

We can't just add in hexdump, that assumes it only works with incoming bytes while perhaps the scoreboard works for other people with other datatypes as well.

That linked code is still assuming str == bytes. If strict_type=False, it tries to do hexdiff on whatever string is returned by the cast, which is just nonsense, but is a str/bytes mixup. I'm not sure how to fix it beyond adding .encode() before hexdump.

I also decided this is getting a big fat FutureWarning about removal.

ktbarrett:
The newest changes directly confront the whole str/bytes issue by encoding strings rather than trying to work around the issue. It's not great, but it's the assumption that Python 2 makes, so making the same assumption *should* be fine.

cocotb/scoreboard.py
```diff
@@ -66,6 +67,11 @@ def __init__(self, dut, reorder_depth=0, fail_immediately=True):  # FIXME: reord
         self.expected = {}
         self._imm = fail_immediately
 
+        warnings.warn(
+            "This Scoreboard implementation has been deprecated and will be removed soon.\n"
+            "If it works for you, please move the implementation into your own project and update your imports",
+            FutureWarning)
```
eric-wieser, line 73:
```suggestion
            PendingDeprecationWarning)
```
or
```suggestion
            DeprecationWarning)
```
`FutureWarning` is

> for warnings about constructs that will change semantically in the future.

Not for constructs that will be removed.

documentation/source/newsfragments/2041.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix usage of :func:`~cocotb.util.hexdump` and :func:`~cocotb.util.hexdiff` in the :class:`~cocotb.scoreboard.Scoreboard`.
```
cmarqu, line 1: Should we add a [``.. deprecated::``](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) here already? 
Some short discussion about it in #1642.

ktbarrett:
I'm actually not so sure the newest set of changes will work. I'm kind of over worrying about this stuff.

@eric-wieser @imphil @cmarqu Since we decided to move the scoreboard, drivers, monitors, and generators into the other repo, perhaps instead of fixing this issue, we should take this opportunity to add `DeprecationWarning`s to the source and `.. deprecated::` tags to the documentation for all the moved entities. That way we give users a version cycle (~6 months) to adjust.

eric-wieser:
> Since we decided to move the scoreboard, drivers, monitors, and generators into the other repo,

Did we? I don't think the generators warrant moving, we can just deprecate them in place.

documentation/source/newsfragments/2041.bugfix.rst
eric-wieser: No, that should probably go in the `Scoreboard` class itself.

ktbarrett:
@eric-wieser We are moving the drivers and monitors, deprecating the generators in place, and didn't decide on the scoreboard. The scoreboard is used in several tests/examples, and so in the Avalon driver, so it probably should tag along with them for the existing Avalon tests.

No matter where they end up, I think we can deprecate everything but the base driver and monitor now for 1.4.1, since we will eventually be doing something with it.

documentation/source/newsfragments/2041.bugfix.rst
ktbarrett: I added it to the scoreboard and everything in cocotb.generators, but only the scoreboard is in the documentation.

ktbarrett:
Not entirely sure why the Windows CI is using the wrong pip. It updates to 20.2 [here](https://travis-ci.org/github/cocotb/cocotb/jobs/723443994#L373), but when running tests uses 19.2 [here](https://travis-ci.org/github/cocotb/cocotb/jobs/723443994#L624).

It also might be a problem because Miniconda now ships with Python 3.8 and we are specifying 3.7.

cocotb/scoreboard.py
```diff
@@ -66,6 +69,11 @@ def __init__(self, dut, reorder_depth=0, fail_immediately=True):  # FIXME: reord
         self.expected = {}
         self._imm = fail_immediately
 
+        warnings.warn(
+            "This Scoreboard implementation has been deprecated and will be removed soon.\n"
+            "If it works for you, please move the implementation into your own project and update your imports.",
```
cmarqu, line 74: Maybe add the "follow cocotb's license agreement" part that you have for generators?

documentation/source/newsfragments/2041.bugfix.rst
cmarqu: #2042 adds generators to the docs so that they can be visibly deprecated.

examples/endian_swapper/tests/generators/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+@public
+def repeat(obj, nrepeat=None):
+    """Generator to repeatedly yield the same object
+
+    Args:
+        obj (any): The object to yield
+
+    Kwargs:
```
cmarqu, line 45: I have fixed the "Kwargs:" stuff to be proper Google-style docstrings and replaced the non-existing ``:type:`` with ``:class:``  in #2042, but I guess git is not *that* smart handle that for us?

ktbarrett: I'll cherry-pick #2042 in.

cmarqu: I'm talking about the versions copied into the example here, and they still show up without the fixes from #2042 in the GitHub GUI. Not super important though, they are not fed through Sphinx currently (just someone may do that in their own project).

cmarqu (approved):
Trusting the CI result on the Travis change.

ktbarrett:
The solution of copying the generators where they were used was probably a bit heavy-handed... I apologize. The one that will go on master will just use the `catch_warnings` approach like with the scoreboard.

PR #2042, opened 2020-09-03
Document generators in Python library reference

cmarqu:
I fixed some small docstring problems but didn't polish it all the way since these will be deprecated anyway with #2041.
I also left out ``packet`` since that depends on ``scapy`` and IMHO doesn't warrant adding that dependency for doc building.


eric-wieser (approved):
Looks good.

https://cocotb--2042.org.readthedocs.build/en/2042/library_reference.html#module-cocotb.generators

ktbarrett (approved):

PR #2048, opened 2020-09-05
Add newsfragments for PRs that should have them

ktbarrett:
I looked through all the changes between the 1.4.0 release and the current master and added a newsfragment for everything I thought should have one, but didn't get merged with one. #1982, #2038, and #2022 are included. I asked @garmin-mjames to comment on whether #1805 should have one. I'm open to other suggestions, rewording, dropping, etc.

### DIFF
--- a/documentation/source/newsfragments/2022.change.rst
+++ b/documentation/source/newsfragments/2022.change.rst
@@ -1,0 +1,1 @@
+Updated :class:`~cocotb.drivers.Driver`, :class:`~cocotb.monitors.Monitor`, and all their subclasses to use the :keyword:`async`/:keyword:`await` syntax instead of the :keyword:`yield` syntax.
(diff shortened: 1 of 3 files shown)


documentation/source/newsfragments/2022.change.rst
```diff
@@ -0,0 +1 @@
+Updated :class:`~cocotb.drivers.Driver`, :class:`~cocotb.monitors.Monitor`, and all subclasses to use the :keyword:`async`/:keyword:`await` syntax, instead of the :keyword:`yield` syntax.
```
cmarqu, line 1: Maybe
```suggestion
Updated :class:`~cocotb.drivers.Driver`, :class:`~cocotb.monitors.Monitor`, and all their subclasses to use the :keyword:`async`/:keyword:`await` syntax, instead of the :keyword:`yield` syntax.
```
I would have left out the last two commas but maybe that's overcompensating the tendency in German to have lots of commas.

cmarqu (approved):

documentation/source/newsfragments/2022.change.rst
ktbarrett: Also removed the last comma. It's unnecessary.

PR #2052, opened 2020-09-07
Describe GUI option for ModelSim

bluecmd:
This option allows the ModelSim simulator to be launched with the waveform visible.

### DIFF
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -257,7 +257,7 @@ If you try to use them with :term:`FLI`, you will see a ``vsim-FLI-3155`` error:
 ModelSim DE and SE (and Questa, of course) support the :term:`FLI`.
 
 In order to start ModelSim or Questa with the graphical interface and for the simulator to remain active after the tests have completed, set :make:var:`GUI=1`.
-If you have previously launched a test without this setting you might have to delete the ``sim_build`` directory to get the correct behavior.
+If you have previously launched a test without this setting, you might have to delete the :make:var:`SIM_BUILD` directory (``sim_build`` by default) to get the correct behavior.
 
 .. _sim-modelsim-issues:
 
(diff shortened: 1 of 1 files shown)


documentation/source/simulator_support.rst
```diff
@@ -256,6 +256,9 @@ If you try to use them with :term:`FLI`, you will see a ``vsim-FLI-3155`` error:
 
 ModelSim DE and SE (and Questa, of course) support the :term:`FLI`.
 
+In order to start ModelSim or Questa with the graphical interface and for the simulator to remain active after the tests have completed, set :make:var:`GUI=1`.
+If you have previously launched a test without this setting you might have to delete the ``sim_build`` directory to get the correct behavior.
```
cmarqu, line 260:
```suggestion
If you have previously launched a test without this setting, you might have to delete the ``sim_build`` directory to get the correct behavior.
```
maybe? Native speaker required ;)

bluecmd: Sounds good to me - I think it is referred to as an Introductory Phrase which should be a good use of commas.

cmarqu (approved):

ktbarrett:
Hmmm, this makes me wonder if we should merge the Questa and Modelsim sections. It's odd the note is in one section and not the other. The Modelsim makefile simply aliases Questa's. They are essentially the same simulator with certain features disabled or behind a pay wall.

Just throwing thoughts out there. It doesn't have to be actionable unless someone else agrees.

themperek:
>Hmmm, this makes me wonder if we should merge the Questa and Modelsim sections. It's odd the note is in one section and not the other. The Modelsim makefile simply aliases Questa's. They are essentially the same simulator with certain features disabled or behind a pay wall.

@ktbarrett Makes sense, maybe separate PR.

Issue #2054, opened 2020-09-08
[Active-HDL] Cached writes in ReadWrite callback


marlonjames:
There is some weirdness around doing a cached write to a signal while in the `ReadWrite` callback.

`Scheduler._do_writes` unconditionally waits on `ReadWrite`:

https://github.com/cocotb/cocotb/blob/8744492c8af4aaa73681362f86fe62be6665bb4d/cocotb/scheduler.py#L254-L261

Additionally, `Scheduler._do_writes` is always run as the first coroutine during the `ReadWrite` callback, if it is waiting on that trigger and not the `Event`.

I've noticed that running **Active-HDL** in VHPI, the `ReadWrite` isn't called again in the same time step unless something else is scheduled to happen. So if you `await ReadWrite` before doing a cached write, it may not apply the write if nothing else is scheduled to cause an additional delta cycle.

---
Here are two testcases I added to **test_timing_triggers.py**:
```python3

async def write_in_readwrite(dut):
    dut.stream_in_valid.setimmediatevalue(0)
    dut.stream_out_ready.setimmediatevalue(0)

    async def write_delayed():
        await ReadWrite()
        dut.stream_in_valid.value = 1

    @cocotb.coroutine
    async def wait_signal():
        await RisingEdge(dut.stream_in_valid)
        dut.stream_out_ready.value = 1

    write_coro = cocotb.fork(write_delayed())

    await cocotb.triggers.with_timeout(wait_signal(), timeout_time=10, timeout_unit='ns')

    await ReadOnly()
    assert dut.stream_out_ready.value == 1


@cocotb.test()
async def test_write_in_readwrite_no_do_writes(dut):
    """Test doing a normal (delayed) write after awaiting ReadWrite trigger.

    Don't fork _do_writes() task before-hand.
    """
    await write_in_readwrite(dut)


@cocotb.test()
async def test_write_in_readwrite(dut):
    """Test doing a normal (delayed) write after awaiting ReadWrite trigger.

    Fork _do_writes() task by doing a delayed write. This ensures that _do_writes() runs
    before write_delayed() when the ReadWrite() trigger fires.
    """
    dut.stream_in_data.value = 0
    await write_in_readwrite(dut)

```
With **Active-HDL 11.1**,
Both tests pass in VPI:
```
     0.00ns INFO     *******************************************************************************************************************
                     ** TEST                                                       PASS/FAIL  SIM TIME(NS)  REAL TIME(S)  RATIO(NS/S) **
[...]

PR #2056, opened 2020-09-08
Add tests for cached writes in ReadWrite callback.

garmin-mjames:
Add tests for #2054 to see Pass/Fail on other simulators.

These fail on Active-HDL VHPI, but pass on VPI.

Possible fix is to use `Scheduler._MODE_WRITE`, and not `await self._read_write` in `Scheduler._do_writes` if in that mode.


garmin-mjames:
This is some bug specific to Active-HDL was showing up here.

PR #2064, opened 2020-09-09
Remove endian swapper from sdist and regression

### LINKED ISSUE #2060: Remove examples from sdist
The examples contained in the sdist *may* have been added to allow people to run self-checks or as templates, however there are problems with this solution.

* they are located in the package prefix, but there is no supported way to clone them to perform a self-check
* you shouldn't run the examples out of the the package as it will pollute the package tree with garbage
* it isn't documented

While the requirements are legitimate, there are better ways to solve these issues that don't require packaging the examples in the sdist.

ktbarrett:
Closes #2060 by removing the examples from the sdist. Also, removes the endian swapper from the regression so the regression can be run without additional dependencies (the hal tests fail on my workstation and anyone else running stock RHEL 6 or 7).


themperek:
Maybe better just skip hal test if we do not find swig executable?

ktbarrett:
Also if the version is too old or too new. I have swig on my RHEL7 machine, but it's out of date so the HAL test fails. I'd need to track that down, but if you think it's beneficial to keep the endian swapper in the regression, I can do that.

ktbarrett:
Newest solution segregates the test_endian_swapper_hal from the other tests and makes it not a part of the regression, while the other tests remain.

themperek:
I think @eric-wieser should comment since he spends some time to make it work as regression at some point.

ktbarrett:
Fixed a rename gprof2dot.py -> gprof2dot.

ktbarrett:
@eric-wieser @themperek Can I get a checkmark on this?

themperek:
For me is OK. But @eric-wieser need to say since he invested there time to fix it.
What I can do is try to change this to use pybind11/cython/... (will add a dependency for the tests) if you think this could help.

imphil (approved):
Thanks for this PR. I agree, let's remove those examples from the sdist, they are of no use in there in the current form.

PR #1498, opened 2020-03-13
Update VHPI callbacks

garmin-mjames:
**PR for testing simulators that support VHPI**
Closes #1061
Closes #1880 

The current use of the `vhpiCbRepEndOfProcesses` callback for `ReadWrite`
triggers can cause subtle issues in simulation. Signal values can take
multiple delta cycles to propogate through designs, and
`vhpiCbRepEndOfProcesses` is called at the end of the current delta cycle.
This means that depending on when signal values are cached when using
the `<=` operator, signal values may be set at the wrong time relative
to other signals.

Changing `ReadWrite` to `vhpiCbRepLastKnownDeltaCycle` and `ReadOnly`
to `vhpiCbRepEndOfTimeStep` brings VHPI behavior more in-line with VPI.

See the following references for further details:
IEEE Std 1076-2008 VHDL Language Ref Manual Section 14.7.5 and 21.3.6.1
IEEE Std 1364-2001 Verilog Hardware Description Language Section 27.33.2

PR #2068, opened 2020-09-10
Update and simplify Scheduler docstring.

garmin-mjames:
The VHPI callbacks were updated in #1498, but this docstring was missed.

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -156,53 +156,20 @@ class external_waiter:
 
 
 class Scheduler:
-    """The main scheduler.
+    """The main coroutine scheduler.
 
-    Here we accept callbacks from the simulator and schedule the appropriate
-    coroutines.
+    The scheduler is entered from a simulator event callback, and the appropriate
+    coroutines are run.
 
-    A callback fires, causing the :any:`react` method to be called, with the
-    trigger that caused the callback as the first argument.
-
-    We look up a list of coroutines to schedule (indexed by the trigger) and
-    schedule them in turn.
+    The simulator callback starts scheduling with a :class:`cocotb.triggers.GPITrigger`,
+    and any coroutines that are waiting on that trigger are scheduled. After all coroutines
+    are finished or waiting on a :class:`cocotb.triggers.Trigger` (including any new ones),
+    the scheduler returns control to the simulator and waits for the next event callback.
 
     .. attention::
 
        Implementors should not depend on the scheduling order!
 
-    Some additional management is required since coroutines can return a list
-    of triggers, to be scheduled when any one of the triggers fires.  To
-    ensure we don't receive spurious callbacks, we have to un-prime all the
-    other triggers when any one fires.
-
-    Due to the simulator nuances and fun with delta delays we have the
-    following modes:
-
-    Normal mode
-        - Callbacks cause coroutines to be scheduled
-        - Any pending writes are cached and do not happen immediately
-
-    ReadOnly mode
-        - Corresponds to :any:`cbReadOnlySynch` (VPI) or :any:`vhpiCbLastKnownDeltaCycle`
-          (VHPI).  In this state we are not allowed to perform writes.
-
-    Write mode
-        - Corresponds to :any:`cbReadWriteSynch` (VPI) or :c:macro:`vhpiCbEndOfProcesses` (VHPI)
-          In this mode we play back all the cached write updates.
-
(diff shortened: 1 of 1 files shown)


cocotb/scheduler.py
```diff
@@ -184,11 +184,11 @@ class Scheduler:
         - Any pending writes are cached and do not happen immediately
 
     ReadOnly mode
-        - Corresponds to :any:`cbReadOnlySynch` (VPI) or :any:`vhpiCbLastKnownDeltaCycle`
+        - Corresponds to :any:`cbReadOnlySynch` (VPI) or :any:`vhpiCbRepEndOfTimeStep`
           (VHPI).  In this state we are not allowed to perform writes.
 
     Write mode
-        - Corresponds to :any:`cbReadWriteSynch` (VPI) or :c:macro:`vhpiCbEndOfProcesses` (VHPI)
+        - Corresponds to :any:`cbReadWriteSynch` (VPI) or :c:macro:`vhpiCbRepLastKnownDeltaCycle` (VHPI)
```
ktbarrett, line 191: The other ones are linked using `:any:` and this one with `:c:macro:`, I don't think either actually link so perhaps it doesn't matter.

ktbarrett (commented):
These docstrings weren't very good to start with. I seriously doubt most users understand the VPI and VHPI callbacks so they aren't a good point of reference. It doesn't mention the FLI callback used. And if we update these again we will have to change this docstring.

I'd also be happy if you removed most of the detail here and left it a "typical" docstring that contains a small blurb about what it does and document the constructor arguments (which are none). Most of the details belong in the internal docs not in the docstring.

cocotb/scheduler.py
```diff
@@ -156,53 +156,20 @@ def thread_wait(self):
...
-    of a given timestep then we must move to a new timestep before performing
-    any writes.  The mechanism for moving to a new timestep may not be
-    consistent across simulators and therefore we provide an abstraction to
-    assist with compatibility.
-
-
-    Unless a coroutine has explicitly requested to be scheduled in ReadOnly
-    mode (for example wanting to sample the finally settled value after all
-    delta delays) then it can reasonably be expected to be scheduled during
-    "normal mode" i.e. where writes are permitted.
```
eric-wieser, line 205: I'm a little wary of completely removing this text, unless we're certain it's completely outdated. I haven't read it in detail though.

garmin-mjames: I could also add a summary of the order of `GPITrigger` callbacks. i.e.,
1. `Timer` / `NextTimeStep` (I'm not sure if VPI and VHPI are same order for these two)
2. `Edge`
3. `ReadWrite`
4. `ReadOnly`

garmin-mjames: Should these details live here in the docstring or a dedicated section of the docs, though?

ktbarrett: You could make an argument they belong with the Triggers rather than the Scheduler. Instead, I would have a separate section on timing triggers and have all of those things link to that documentation.

ktbarrett: I'm also not sure what this paragraph is saying.  Is it just saying "you can write in any mode besides ReadOnly"?

cocotb/scheduler.py
cmarqu: They actually do link (both): https://docs.cocotb.org/en/latest/library_reference.html#cocotb.scheduler.Scheduler

ktbarrett: We are documenting the VPI and VHPI in this project?

cmarqu: A tiny bit, kind of, due to Doxygen picking up the headers and breathe converting it to Sphinx. Would you prefer not to link because it's mostly useless here?

ktbarrett: That can probably be handled a little more elegantly. We don't even need the whole generate apidoc and include it, we can directly use doxygen directives. We can list out the files we actually want in the C++ documentation.

Of course there is the issue: what should be in the C++ documentation? None of it is user-face code besides maybe the GPI, which we aren't version controlling rn. It does seem like the best place to put the C++ documentation... Maybe we should unlist it in the toctree? But still be able to access it for linking to/from the internal docs? Just throwing stuff out there...

cmarqu: Something for the maintainer call. We could certainly separate the developer documentation more from the end user documentation.

PR #2069, opened 2020-09-11
Fixup documentation for global  variables.

garmin-mjames:
Documentation for globals in the cocotb module show "= None" because the
autodata directive tries to get the value by default.
A blank annotation option overrides the default to show no value.

Looks like it also removes type info: https://cocotb--2069.org.readthedocs.build/en/2069/library_reference.html#cocotb.scheduler

### DIFF
--- a/documentation/source/library_reference.rst
+++ b/documentation/source/library_reference.rst
@@ -361,18 +361,25 @@ Other Runtime Information
 -------------------------
 
 .. autodata:: cocotb.argv
+   :no-value:
 
 .. autodata:: cocotb.SIM_NAME
+   :no-value:
 
 .. autodata:: cocotb.SIM_VERSION
+   :no-value:
 
 .. autodata:: cocotb.RANDOM_SEED
+   :no-value:
 
 .. autodata:: cocotb.plusargs
+   :no-value:
 
 .. autodata:: cocotb.LANGUAGE
+   :no-value:
 
 .. autodata:: cocotb.top
+   :no-value:
 
 
 Signal Tracer for WaveDrom
@@ -397,6 +404,7 @@ The Scheduler
 .. currentmodule:: cocotb.scheduler
 
 .. autodata:: cocotb.scheduler
+   :no-value:
 
 .. autoclass:: Scheduler
     :members:
@@ -408,6 +416,7 @@ The Regression Manager
 .. currentmodule:: cocotb.regression
 
 .. autodata:: cocotb.regression_manager
+   :no-value:
 
 .. autoclass:: RegressionManager
     :members:
--- a/documentation/requirements.txt
+++ b/documentation/requirements.txt
@@ -1,6 +1,6 @@
 # Python dependencies to build the documentation
 
-Sphinx ~= 3.1
+Sphinx ~= 3.4
 breathe
 sphinx-rtd-theme
 cairosvg


ktbarrett:
If we add type comments will the type come back? We unfortunately can't use variable type annotations since we have to support 3.5 [for now](https://www.python.org/downloads/release/python-3510/#:~:text=Python%203.5%20will%20reach%20its,release%20of%20the%203.5%20series.).

garmin-mjames:
@ktbarrett The type hints comments are in `__init__.py` but they get wiped out. I could try adding them back to the `:annotation;` option but then we're dealing with dual maintenance.

ktbarrett:
Looks like at one point there was a `:novalue:` directive. https://github.com/sphinx-doc/sphinx/commit/6b9f35455ddadfe519289ed6867a9798ba40918a#diff-c31a64b012060e63a33ad6b6c3f97f01
But it was removed. Perhaps we should add it back in? Or if/when we drop Python 3.5 we can just use type annotations and not assign a value.

garmin-mjames:
I'm also fine with closing this and waiting for a better alternative.

ktbarrett:
If the only purpose is to remove "= None" from the documentation, I don't think it's huge deal to leave it in there if there isn't a better alternative.

documentation/source/library_reference.rst
```diff
@@ -358,16 +358,22 @@ Other Runtime Information
 -------------------------
 
 .. autodata:: cocotb.argv
+   :annotation:
```
eric-wieser, line 361: Does 
```suggestion
   :value:
```

by itself work?

garmin-mjames: Shows this:

`documentation\source\library_reference.rst:360: WARNING: An option to autodata is either unknown or has an invalid value: 'value'`

And the variable doesn't show up in the docs output.

garmin-mjames:
Using `:annotation: : typeHint` works, but again it's a manual process. I can open an issue in the `sphinx-doc` project.

ktbarrett:
AFAIK sphinx-build supports 3.5, so it should support this use case.

garmin-mjames:
I've got a sphinx issue https://github.com/sphinx-doc/sphinx/issues/8209 and PR https://github.com/sphinx-doc/sphinx/pull/8222 for adding a **:novalue:** option that would do what we want.

garmin-mjames:
https://github.com/sphinx-doc/sphinx/pull/8424 was merged as an alternative to the above PR.
Needs sphinx 3.4

cmarqu (approved):

ktbarrett:
Sphinx 3.4 will be release December 19, 2020. We are blocked until then.

eric-wieser:
We could always use an unreleased version if we want the fixed docs now.

ktbarrett:
Not sure it's worth it. 1.5 won't be released until after that date (I think), so it will still make it in the released docs for that version.

garmin-mjames:
Preview:

image

garmin-mjames:
Added `cocotb.top`

ktbarrett:
Sphinx 3.4 has been released! Re-running tests...

ktbarrett:
It's not working... even locally... AHHHHHHH.

garmin-mjames:
When I build docs on master, the type info is gone as well, but the `= None` value is still there.

garmin-mjames:
Git bisect shows this commit broke the type hints: https://github.com/sphinx-doc/sphinx/pull/8542/commits/55c110f60965ea6f0ea32d129b62413f1f84b3f4

ktbarrett:
@garmin-mjames I guess they figured that is fine since they are not worried about supporting 3.5 anymore. Maybe we should just give up on this and switch to variable attributes after 1.5. Or we can merge this now, since it isn't blowing up, and try to fix that change in Sphinx for 3.4.1?

garmin-mjames:
Latest docs show the type hints: https://docs.cocotb.org/en/latest/library_reference.html#other-runtime-information
But with `tox -e docs` locally, they are missing...

ktbarrett:
@garmin-mjames When I run `tox -e docs` locally, it brings in Sphinx 3.3.1, not 3.4. Some dependency is probably blocking the upgrade and conda can figure it out while pip can't?

garmin-mjames:
Locally I get:
```
docs installed: 
breathe==4.25.1,
Sphinx==3.4.2,
```
And the type info is gone:

image

The sphinx autodoc tests show the `:type:` is generated for module-level variables that use comment type hints (`# type: str`, etc), so I'm not sure how to debug this further.

ktbarrett:
That's weird. I have the same versions and I get *nothing* after the name, no type info or value.

garmin-mjames:
Sorry, that was running on master. I get nothing with this PR.

garmin-mjames:
I'm going to close this, as the Sphinx `:no-value:` option is not working. This will become moot when Python 3.6 is the minimum supported version.

Issue #497, opened 2016-11-08
Hierarchy access for generate statement


kirknodeng:
When we instantiate some block with generate statement. 
How to access that hierarchical module?

example) When I want to access "SubBlock[3].u_SubModule.a". 
genvar i;
generate 
  for(i=0;i<5;i=i+1) begin : SubBlock
    SubModule u_SubModule(
       .a(...),
       .b(...),
    );
  end
endgenerate

First, I tried to do this 
RisingEdge(dut.SubBlock[3].u_SubModule.a) -> cocotb can't find SubBlock

and, Second
RisingEdge(dut.u_SubModule.a) -> cocotb find their object. but, which module accessed? 
                                                         I instantiated five modules.

leftink:
The first way should work I believe.  What simulator are you using?

You should try running test_array.  It tests something similar.

kirknodeng:
I tried to running test_array using VCS. It's similar. but, different.

So, I want to make this case test into test_cases. Can I make it? 
It is first time to make test case and push request. Is it okay?

ktbarrett:
If you are still having trouble create a new issue.

mattvenn:
I believe there is still an issue with hierarchy access - please see this MVCE
https://github.com/mattvenn/cocotb-mvce-generate-bug/blob/master/test.py

mattvenn:
This is with icarus 10.3 compiled from source

ktbarrett:
Running with debugging I get [this](https://github.com/cocotb/cocotb/files/5255524/output.log). If I set line 9 to `True`, the test passes and I get [this](https://github.com/cocotb/cocotb/files/5255513/output.log). I'll look into this more later, but there is definitely a problem with the discovery of pseudo-region. Calling in the expert on pseudo-regions @garmin-mjames.

ktbarrett:
I see the problem and I'm astonished that this hasn't been caught before...

When attempting to access `genblk1`, the VPI implementation is unaware that it is a pseudo-region, causing `vpi_handle_by_name` to fail. When iterating over "top" beforehand, the pseudo-region is discovered, created, and cached so that following accesses succeeds. There are a lot of ways to fix this issue, I'm not sure which is best.

I'm starting to think we should be iterating over all related objects, properly creating and caching those related-objects, whenever we access an object. This approach is also the best solution for #1128.

garmin-mjames:
@ktbarrett It might be enough to call `_discover_all` before looking for sub handles.

[...]

PR #2075, opened 2020-09-22
fix suggested by ktbarret for issue #497

mattvenn:
I ran make test before and after the patch. 
I have tests that fail both before and after, but I can't see easily if there are more fails after.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -249,6 +249,10 @@ class HierarchyObject(RegionObject):
         except KeyError:
             pass
 
+        if not self._discovered:
+            self._discover_all()
+            return self.__get_sub_handle_by_name(name)
+
         # Cache to avoid a call to the simulator if we already know the name is
         # invalid. Unclear if we care, but we had this before.
         if name in self._invalid_sub_handles:


ktbarrett:
Azure is showing a legitimate failure with this change. I guess we are going to have to come up with a more elegant solution.

garmin-mjames:
Looking into `test_unicode_handle_assignment_deprecated` failure on Active-HDL 11.1 (should be similar to Riviera):

The problem is that iterating over `dut` (in this case **sample_module**) finds `sample_module.stream_in_string` as a `vpiPort`, so the handle becomes `HierarchyObject(sample_module.stream_in_string)`.

If we try to get the handle directly, `vpi_handle_by_name` returns a `vpiStringVar` so the handle becomes `StringObject(sample_module.stream_in_string)`.

Riviera and Active-HDL aren't finding `stream_in_string` using the list of iterators until `vpiPort`:
https://github.com/cocotb/cocotb/blob/fb02018e9b44100a22a43e841d0cd2293cac3627/cocotb/share/lib/vpi/VpiCbHdl.cpp#L565-L597

ktbarrett:
@garmin-mjames That blasted SINGLETON_HANDLES again... I brought this up in the issue. I really think we need to refactor how objects are discovered since they can have multiple types. SINGLETON_HANDLES isn't the real problem (well, it is a problem, but that's a completely different issue), it's the fact we let handles be created with only `vpiPort` or `vpiInterconnectNet`, etc (the net type or errata) rather than associating *all* type information with the object handle.

What I meant when I talked about a more elegant solution in the issue is when we search for subhandles we need to iterate over all related objects using `vpi_iterate`, create handles to objects we find ***with all the types associated with that object*** (net type, portness, constness, etc.), and then cache that object in SINGLETON_HANDLES. 

The big issue here is ***none*** of the FOSS simulators and only some of the commercial simulators implement `vpi_compare_objects` which we would need to implement this.

garmin-mjames:
@ktbarrett, Yeah, ports should probably not be returned as their own handle. `VhpiImpl::create_gpi_obj_from_handle` gets the base type and looks at that, so we always get the signal associated with the port. We might be able to follow the port to the actual:
```cpp
new_handle = vpi_get(vpiLowConn, port_handle);
```
But that can return any expression, so we'd have to investigate further...

garmin-mjames:
For this issue, I think we can fix the new errors by adding `vpiVariables` back into the module iterators so it gets seen before `vpiPort` iteration.

ktbarrett:
@garmin-mjames Want to try that?

@mattvenn Sorry for taking over your PR. We will try to fix your original issue.

garmin-mjames:
@ktbarret it worked on Active-HDL, i'll make a new PR.

ktbarrett:
@mattvenn Thanks for putting this up, it helped us locate an issue with the suggested fix, but we are going to take #2079.

PR #2079, opened 2020-09-24
Iterate over handle if sub-handle lookup fails

### LINKED ISSUE #497: Hierarchy access for generate statement

garmin-mjames:
Fixes #497. 

Pseudo-region lookup may fail, so iterate over sub-handles using `_discover_all` and try lookup again.

Once this was added, it uncovered another issue where iterating on a module would find the vpiPort for
a variable and create a `HierarchyObject` rather than the underlying variable itself which becomes a
`ModifiableObject` or subclass.

---
Todo:
- [x] Test
- [x] Newsfragment?

### DIFF
--- a/documentation/source/newsfragments/2079.bugfix.rst
+++ b/documentation/source/newsfragments/2079.bugfix.rst
@@ -1,0 +1,8 @@
+Generate blocks are now accessible directly via lookup without having to iterate over parent handle. (:pr:`2079`)
+
+  .. code-block:: python3
+
+      # Example pseudo-region
+      dut.genblk1       #<class 'cocotb.handle.HierarchyArrayObject'>
+
+  .. consume the towncrier issue number on this line.
--- a/tests/designs/sample_module/sample_module.vhdl
+++ b/tests/designs/sample_module/sample_module.vhdl
@@ -107,7 +107,19 @@ architecture impl of sample_module is
     type twoDimArrayType is array (natural range <>) of unsignedArrayType(31 downto 28);
     signal array_2d         : twoDimArrayType(0 to 1);
 
+    constant NUM_OF_MODULES : natural := 4;
+    signal temp             : std_logic_vector(NUM_OF_MODULES-1 downto 0);
+
 begin
+
+    genblk1: for i in NUM_OF_MODULES - 1 downto 0 generate
+    begin
+        process (clk) begin
+            if rising_edge(clk) then
+                temp(i) <= '0';
+            end if;
+        end process;
+    end generate;
 
     process (clk) begin
         if rising_edge(clk) then
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -249,6 +249,10 @@ class HierarchyObject(RegionObject):
         except KeyError:
             pass
 
+        if not self._discovered:
+            self._discover_all()
+            return self.__get_sub_handle_by_name(name)
+
         # Cache to avoid a call to the simulator if we already know the name is
         # invalid. Unclear if we care, but we had this before.
         if name in self._invalid_sub_handles:
--- a/cocotb/share/lib/vpi/VpiCbHdl.cpp
+++ b/cocotb/share/lib/vpi/VpiCbHdl.cpp
@@ -576,7 +576,7 @@ decltype(VpiIterator::iterate_over) VpiIterator::iterate_over = []{
         vpiRealNet,
         vpiStructVar,
         vpiStructNet,
-        //vpiVariables          // Aldec SEGV on plain Verilog
+        vpiVariables,
         vpiNamedEvent,
         vpiNamedEventArray,
         vpiParameter,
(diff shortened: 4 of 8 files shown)


garmin-mjames:
@ktbarrett of course, Riviera iteration tests fail now because they find more things.

garmin-mjames:
This makes Verilator hang for some reason.

ktbarrett:
> Riviera iteration tests fail now because they find more things.

That's a good thing!

> This makes Verilator hang for some reason.

Not surprised. Their VPI support has proven itself to be pretty unstable.

garmin-mjames:
@ktbarrett I'm inferring from the failed test names in the Riviera CI. I can't access the actual results, so I can't see what the updated numbers are.

`vpiVariables` wasn't added before because according to the comment:
> // Aldec SEGV on plain Verilog

I don't know if we have a test that hits that right now, or if there is even a difference between Verilog and SystemVerilog in newer versions of Riviera.

ktbarrett:
> I can't access the actual results

Yes, we will fix that soon.

> // Aldec SEGV on plain Verilog

That comment is 5 years old. It's possible newer version of Aldec simulators don't have this bug anymore.

ktbarrett:
#### test_iteration_mixedlang/test_iteration/recursive_discovery
```
cocotb.result.TestFailure: Expected 966 but found 1024
```

#### test_array/test_array/test_discover_all
```
cocotb.result.TestFailure: Expected 947 objects but found 1006
```

garmin-mjames:
#### test_iteration_mixedlang/test_iteration/recursive_discovery

The difference of 58 is discovery of both `verilog_toplevel.bus_verilog` and `verilog_toplevel.bus_vhdl` and their internal objects.

#### test_array/test_array/test_discover_all

`array_module.port_rec_out` and `array_module.sig_rec` are found and iterated over.

We have several recursive iteration tests. I'd like to refactor them to build a set of expected handle objects and check:

- path
- type (handle.py class)
- GPI_* type

and print out at the end any expected that weren't found, and any found that weren't expected. It'd give us some information rather than a number mismatch.

ktbarrett:
I agree number mismatch is a pretty bad way to do it. I absolutely agree! Limited testbenches that you can check everything would be a lot better. Could that be in another PR?

garmin-mjames:
Yeah I can just update the number here.

garmin-mjames:
**test_iteration_mixedlang/test_iteration/recursive_discovery** should probably be run with Modelsim, etc

ktbarrett:
Tested on Questa. No issues...

garmin-mjames:
Hmm, I think the newsfragment could be clearer. Do we have documentation on accessing generate blocks?

documentation/source/newsfragments/2079.bugfix.rst
```diff
@@ -0,0 +1,8 @@
+Generate blocks are now accessible directly via lookup without having to iterate over parent handle. (:pr:`2079`)
+
+  .. code-block:: python3
+
+      # Example pseudo-region
+      dut.genblk1       #<class 'cocotb.handle.HierarchyArrayObject'>
+
+  .. consume the towncrier issue number on this line.
```
ktbarrett, line 8: Is this line a pragma? Is it doing anything?

ktbarrett (commented):
[This](https://docs.cocotb.org/en/stable/quickstart.html?highlight=handle#assigning-values-to-signals) is the only documentation we have on accessing handles I think.

This looks good. I'll let it sit for a couple days in case anyone else wants a look, then merge it.

documentation/source/newsfragments/2079.bugfix.rst
cmarqu: Towncrier will always append the issue/PR number to the last non-empty line of a newsfragment, and it doesn't respect ReST markup or anything.

The comment is a small hack to neutralize this append step.

ktbarrett (approved):

tests/designs/sample_module/sample_module.vhdl
```diff
@@ -107,8 +107,20 @@ architecture impl of sample_module is
     type twoDimArrayType is array (natural range <>) of unsignedArrayType(31 downto 28);
     signal array_2d         : twoDimArrayType(0 to 1);
 
+    signal temp             : std_logic_vector(3 downto 0);
```
cmarqu, line 110: Do you have a specific reason why you don't derive the ``3``from ``NUM_OF_MODULES-1``?

garmin-mjames: No, I'll fix that.

ktbarrett:
I'm going to merge this, but I will say I don't think this is the most elegant solution, but it is a solution we can easily bring in now.

PR #1787, opened 2020-05-11
Fix icarus master on Windows

tpambor:
Using latest icarus verilog from master fails with:
```
/mingw64/bin/vvp -M /mingw64/lib/python3.8/site-packages/cocotb/libs -m libcocotbvpi_icarus   sim_build/sim.vvp
libcocotbvpi_icarus:`C:/msys64/mingw64/lib/python3.8/site-packages/cocotb/libs\libcocotbvpi_icarus.vpl' failed to open using dlopen() because:
    %1 is not a valid Win32 application.
```
Starting with [a59b183b](https://github.com/steveicarus/iverilog/commit/a59b183bd1890eff63cc59d38084dde15de281ff) icarus verilog on windows no longer exports `vpi_*` in vvp.exe. As a result, the current libcocotbvpi_icarus.vpl is unable to load because it no longer can import these symbols.

Using this PR compatibility using old and new icarus verilog versions on Windows is maintained.

New icarus verilog versions on Windows pass pointers to `vpi_*` via `vpip_set_callback`.
If functions are provided via `vpip_set_callback` use these, otherwise fallback to resolve these from vvp.exe at runtime.

This PR is implemented on top of #1783.

Closes #1786

PR #2082, opened 2020-09-27
Fix GHA regression tests

### LINKED ISSUE #2092: CI regressions on Windows
We are currently using Conda for Python installation management on Windows, mainly so we can use the msys2 packages to pull in a `gcc` and `make` to build and run cocotb. We have had issues with Windows CI in the past (#1909) with sqlite.dll not being on the path, and again recently with cocotb not being importable.

Both have the same root cause: tox using virtualenvs is not compatible with the weirdness of conda installations on Windows. Compare a tox environment with a top-level miniconda environment and you'll see significant differences. All, the tests tend to pass when *not* using tox.

There are 2 alternatives here:
1. Don't use tox on Windows. This is fine, we are booting a new VM, installing miniconda afresh. Tox is actually not offering us much in terms of isolation this instance.
2. Use `tox-conda` and move conda requirements into tox.ini. I'm not sure how this will affect regressions in non-conda environments like Linux and MacOS.

I'm in favor of 1.

ktbarrett:
Currently it passes because it's testing nothing.

We are now installing iverilog to a location we can access from powershell. Also, removed tox-gh-actions, which is not being used right now. Still seeing the exact same problem as before with cocotb magically not being importable.

Comparing to the last passing Travis run, not much has changed. https://travis-ci.org/github/cocotb/cocotb/jobs/727068949. Python version shouldn't matter, I tested going down to 3.7 and didn't see an issue. tox version is the same...

Also closes #2092.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -29,7 +29,6 @@ jobs:
             python-version: 3.8
             os: 'windows-latest'
             may_fail: true
-    continue-on-error: ${{matrix.may_fail || false}}
     env:
       SIM: ${{matrix.sim}}
       TOPLEVEL_LANG: ${{matrix.lang}}
@@ -44,7 +43,7 @@ jobs:
         python-version: ${{matrix.python-version}}
     - name: Set up cocotb build dependencies
       run: |
-        conda install -c msys2 m2-base m2-make m2-autoconf m2-flex m2-bison m2-gperf m2w64-toolchain libpython virtualenv=16
+        conda install -c msys2 m2-base m2-make m2-autoconf m2-flex m2-bison m2-gperf m2w64-toolchain libpython
     - name: Build and install Icarus
       run: |
         git clone https://github.com/steveicarus/iverilog.git -b "${{matrix.sim-version}}"
@@ -56,7 +55,21 @@ jobs:
         echo "::add-path::C:\iverilog\bin"
     - name: Install Python testing dependencies
       run: |
-        pip install tox
-    - name: Test with tox
+        pip install coverage xunitparser pytest pytest-cov
+    - name: Install cocotb
       run: |
-        tox -e py3_mingw
+        python -m pip install  --global-option build_ext --global-option --compiler=mingw32 -e .
+    - name: Test
+      continue-on-error: ${{matrix.may_fail || false}}
+      run: |
+        pytest
+        $pass = $?
+        make test
+        exit -not ($pass -and $?)
+      # inverted the pass for exit because the boolean is converted to 0/1, but an exit code of 1 is a failure
+    - name: Upload coverage
+      if: steps.Test.outcome != 'failure'
+      run: |
+        pip install codecov
+        bash -c 'find . -type f -name "\.coverage\.cocotb" | xargs coverage combine --append'
+        codecov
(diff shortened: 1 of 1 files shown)


garmin-mjames:
Could be **virtualenv**. Went from 16.7.5 to 20.x.

ktbarrett:
@garmin-mjames Wow that's a jump, didn't think to check that. I'll try pinning it and see if that helps. But  will need a better solution going forward.

ktbarrett:
So pinning virtualenv to 16 prevented the issue. There are numerous issues in virtualenv and tox about breakages when going to virtualenv==20, but none relating to what we are seeing. Seems like that transition wasn't handled so well.

I tried using tox-conda to use conda instead of virtualenv to handle the environment since using virtualenv with conda is probably not kosher. Doing makes you run into another known issue that hasn't been fixed for years in distutils: Visual Studio 14 renames msvcr140 to vcruntime140, so you end up not being able to find "-lmsvcr140" during link. I tried manually fixing up my own distuils on Windows and it at least gets past that issue.

Distutils *freaks* if Visual Studio is installed. "cl.exe exists? Why *wouldn't* you use it? Of course!" Since installing Visual Studio, I cannot install cocotb on Windows. Github Actions has Visual Studio installed on all windows instances.

ktbarrett:
Icarus master is expected to fail regressions. We currently don't have a nice way to check simulator version.

themperek:
To fix master: https://github.com/cocotb/cocotb/pull/1787

themperek (approved):

ktbarrett:
@themperek Ah yes, there is also that issue.

With that solved we also have the issue of two tests failing in the regression `test_discovery.access_single_bit` and `test_discovery.assign_single_bit` due to Icarus finally implementing the ability to reference single bits in a vector; those tests are expected to fail, but don't now. We need #1987 to solve that.

ktbarrett:
Changed how we were doing Windows regressions, no long using tox. See #2092.

.github/workflows/regression-tests.yml
```diff
@@ -36,18 +43,36 @@ jobs:
         python-version: ${{matrix.python-version}}
     - name: Set up cocotb build dependencies
       run: |
-        conda install -c msys2 m2-base m2-make m2-autoconf m2-flex m2-bison m2-gperf m2w64-toolchain libpython
+        conda install -c msys2 m2-base m2-make m2-autoconf m2-flex m2-bison m2-gperf m2w64-toolchain libpython virtualenv=16
```
garmin-mjames, line 46: We shouldn't need `virtualenv` anymore.

.github/workflows/regression-tests.yml
```diff
@@ -42,12 +49,29 @@ jobs:
...
+        python -m pip install  --global-option build_ext --global-option --compiler=mingw32 -e .
+    - name: Test
+      continue-on-error: ${{matrix.may_fail || false}}
+      run: |
+        $pass = $true
+        pytest
+        $pass = $pass -and $?
+        make test
+        $pass = $pass -and $?
+        exit -not $pass
```
ktbarrett, line 70: This might be silly. If one command fails does the step mark failure? Does it keep going and mark failure? Or is it only the exit code of the last command?

ktbarrett: Nope, it seemingly only considers the outcome of the final command.

imphil (approved):

themperek (approved):

PR #2086, opened 2020-09-27
Document envvar COCOTB_LIBRARY_COVERAGE.

### LINKED ISSUE #2083: Document COCOTB_LIBRARY_COVERAGE envvar
The environment variable ``COCOTB_LIBRARY_COVERAGE`` -
https://github.com/cocotb/cocotb/blob/b2eda5325a1e75ef6bdde4e6523bcb0ac81a3137/cocotb/__init__.py#L160
is not documented.

cmarqu:
Closes #2083

### DIFF
--- a/documentation/source/building.rst
+++ b/documentation/source/building.rst
@@ -127,17 +127,6 @@ Cocotb
 
     This needs the :mod:`cherrypy` and :mod:`dowser` Python modules installed.
 
-.. envvar:: COCOTB_PY_DIR
-
-    Path to the directory containing the cocotb Python package in the :file:`cocotb` subdirectory;
-    for cocotb-internal use.
-
-.. envvar:: COCOTB_SHARE_DIR
-
-    Path to the directory containing the cocotb Makefiles and simulator libraries in the subdirectories
-    :file:`lib`, :file:`include`, and :file:`makefiles`;
-    for cocotb-internal use.
-
 
 Regression Manager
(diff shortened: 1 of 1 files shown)


imphil:
Is that really a variable we want to document for end users? I see it more as an internal helper that might change at any time, and users should not rely on it (and shouldn't have any reason to use it, actually).

eric-wieser:
It might still be nice to have it documented somewhere for cocotb maintainers, but I agree that it probably doesn't belong next to the user environment variables.

imphil:
@cmarqu maybe you can find a more "cocotb developers-internal" section in the docs. If not, I'm fine with leaving it where it is now, but adding text like "For cocotb internal use only."

cmarqu:
We already have documentation for cocotb-internal variables, and I figured consistency/completeness would be nice here.
I'll move these internals to a new section.

documentation/source/building.rst
```diff
@@ -326,6 +315,27 @@ and
    `CXXFLAGS`, `LDLIBS` are not supported by distutils/pip
 
 
+Internal Variables
+------------------
+
+The following variables are used for cocotb-internals.
```
imphil, line 321:
```suggestion
The following variables are used for cocotb internals.
```

imphil (approved):

ktbarrett (approved):

PR #2091, opened 2020-09-30
Random fixes

ktbarrett:
Small nitpicks causing warnings.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,8 @@ from io import StringIO
 
 # Note: cocotb is not installed properly yet and is missing dependencies and binaries
 # We can still import other files next to setup.py, as long as they're in MANIFEST.in
+# The below line is necessary for PEP517 support
+sys.path.append(path.dirname(__file__))
 from cocotb_build_libs import get_ext, build_ext
 
 
--- a/documentation/source/newsfragments/2091.changes.rst
+++ b/documentation/source/newsfragments/2091.changes.rst
@@ -1,0 +1,1 @@
+The package build process is now fully PEP517 compliant.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
 [tool.towncrier]
     package = "cocotb"
     directory = "documentation/source/newsfragments"
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,6 +30,5 @@ per-file-ignores =
 [tool:pytest]
 addopts = -v --cov=cocotb --cov-branch
 testpaths = tests/pytest
-confcutdir = tests/pytest
 # log_cli = true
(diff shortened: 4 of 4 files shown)


ktbarrett:
PEP517 does an isolated build. I bet you the project root isn't in `sys.path`. I'll test later.

ktbarrett:
Add a work around for the import issue. Not sure if this is kosher, hopefully I will get a response from the pypa devs.

ktbarrett:
The modification of `sys.path` is the proper way to handle this according to [this response](https://github.com/pypa/setuptools/issues/2405#issuecomment-704875478). The PR is good for a final review.

setup.py
```diff
@@ -53,8 +53,7 @@
 from os import path, walk
 from io import StringIO
 
-# Note: cocotb is not installed properly yet and is missing dependencies and binaries
-# We can still import other files next to setup.py, as long as they're in MANIFEST.in
```
eric-wieser, line 57: I think this comment is still mostly accurate, despite the new path code.

ktbarrett: Added it back and added a note about the sys.path line.

setup.py
```diff
@@ -55,6 +55,8 @@
 
 # Note: cocotb is not installed properly yet and is missing dependencies and binaries
 # We can still import other files next to setup.py, as long as they're in MANIFEST.in
+# The below line is necessary for PEP517 support (gh-2091)
+sys.path.append(path.dirname(__file__))
```
eric-wieser, line 59: This is needs because we edited pyproject.toml, or was it already needed before?

ktbarrett: Only because we edited pyproject.toml. Basically we changed build backends.

Having a pyproject.toml enabled PEP517 support, but if you don't specify a backend it defaults to the legacy backend (behaves like `python setup.py bdist_wheel`), the new backend `setuptools.build_meta` doesn't respect MANIFEST.in since it isn't a part of the PEP517 spec, it's a part of building an sdist. Eventually the legacy backend will be removed and all projects will be required to be fully PEP517 compliant. Specifying the legacy backend via `setuptools.build_meta:__legacy__` is already deprecated.

eric-wieser: It would be good if you could capture that in the commit message. Supporting PEP517 is probably also release note worthy too.

I'd remove the PR reference from this comment, because the PR title and description doesn't help at all.

setup.py
```diff
@@ -55,6 +55,8 @@
 
 # Note: cocotb is not installed properly yet and is missing dependencies and binaries
 # We can still import other files next to setup.py, as long as they're in MANIFEST.in
```
eric-wieser, line 57: Based on your other comment, the remark about `MANIFEST.in` is now false?

ktbarrett: It's kind of both. Without MANIFEST.in it's not in the sdist, without the second line you can't build it from source or sdist.

eric-wieser (approved):

PR #2100, opened 2020-10-06
Add test for the .loads() iterator

cmarqu:
The test itself is only for completeness' sake, as it doesn't seem to be implemented in most (any?) of our supported simulators.
Add a comment about this to the docstrings.

### DIFF
--- a/tests/test_cases/test_iteration_mixedlang/test_iteration.py
+++ b/tests/test_cases/test_iteration_mixedlang/test_iteration.py
@@ -45,39 +45,33 @@ def recursive_dump(parent, log):
 
 
 @cocotb.test(expect_fail=True)
-def test_drivers(dut):
-    """
-    Try iterating over drivers of a signal.
+async def test_drivers(dut):
+    """Try iterating over drivers of a signal.
 
-    Seems that few simulators implement vpiDriver
+    Seems that few simulators implement ``vpiDriver``.
     """
     tlog = logging.getLogger("cocotb.test")
-    yield Timer(100)
-    for driver in dut.i_verilog.uart1.uart_rx_1.rx_data.drivers():
+    drivers = list(dut.i_verilog.uart1.uart_rx_1.rx_data.drivers())
+    assert drivers, "No drivers found for dut.i_verilog.uart1.uart_rx_1.rx_data"
+    for driver in drivers:
         tlog.info("Found %s" % repr(driver))
-        break
-    else:
-        raise TestFailure("No drivers found for dut.i_verilog.uart1.uart_rx_1.rx_data")
 
 
 @cocotb.test(expect_fail=True)
-def test_loads(dut):
-    """
-    Try iterating over loads of a signal.
+async def test_loads(dut):
+    """Try iterating over loads of a signal.
 
-    Seems that few simulators implement vpiLoad
+    Seems that few simulators implement ``vpiLoad``.
     """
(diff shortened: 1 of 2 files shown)


tests/test_cases/test_iteration_mixedlang/test_iteration.py
```diff
@@ -60,6 +60,22 @@ def test_drivers(dut):
         raise TestFailure("No drivers found for dut.i_verilog.uart1.uart_rx_1.rx_data")
 
 
+@cocotb.test(expect_fail=True)
+def test_loads(dut):
```
ktbarrett, line 64:
```suggestion
async def test_loads(dut):
```

tests/test_cases/test_iteration_mixedlang/test_iteration.py
```diff
@@ -60,6 +60,22 @@ def test_drivers(dut):
...
+    """
+    Try iterating over loads of a signal.
+
+    Seems that few simulators implement vpiLoad
+    """
+    tlog = logging.getLogger("cocotb.test")
+    yield Timer(100)
```
ktbarrett, line 71: Don't need this now that it's `async`.
```suggestion
```

tests/test_cases/test_iteration_mixedlang/test_iteration.py
```diff
@@ -60,6 +60,22 @@ def test_drivers(dut):
...
+    for load in dut.i_verilog.uart1.ser_in.loads():
+        tlog.info("Found %s" % repr(load))
+        break
+    else:
+        raise TestFailure("No loads found for dut.i_verilog.uart1.ser_in")
```
ktbarrett, line 76: Is this intended? Print the first result and exit? Even if it is, it looks a bit odd.

I might suggest.
```suggestion
    loads =  dut.i_verilog.uart1.ser_in.loads()
    assert loads, "No loads found for dut.i_verilog.uart1.ser_in"
    for load in loads:
        tlog.info("Found %s" % repr(load))
```

tests/test_cases/test_iteration_mixedlang/test_iteration.py
cmarqu: This is an almost exact copy of the existing test for ``vpiDriver``, and this whole file is not yet converted to ``async`` syntax, so I'd do that in one go in another PR.

tests/test_cases/test_iteration_mixedlang/test_iteration.py
cmarqu: It's indeed a strange test, maybe buggy in the path that has never been taken since no simulator supports it.

tests/test_cases/test_iteration_mixedlang/test_iteration.py
ktbarrett: I see the corresponding `test_drivers` test above it now that mostly identical. You could update that test as well in this PR with another commit? If not, I understand.

ktbarrett (commented):
Thank you for updating the tests! Changing them all to `async` (besides a select few) is a stretch goal. And it's not going to be fun updating them all at once. So piecemeal changes are great.

The updates look good. One last thing I would change is the `TestFailure` conditions over to `assert`s in `recursive_discovery` and `recursive_discovery_boundary`.

eric-wieser (approved):

ktbarrett (approved):

PR #2101, opened 2020-10-06
Print object type info in VpiSingleIterator

cmarqu:
Some small message improvement while looking at ``.drivers()``/``.loads()``.
Can we easily get the strings ``vpiLoad``/``vpiDriver`` from the raw ``vpitype`` (``91``/``93``) that is handed in?

### DIFF
--- a/cocotb/share/lib/vpi/VpiImpl.h
+++ b/cocotb/share/lib/vpi/VpiImpl.h
@@ -223,7 +223,8 @@ public:
         vpiHandle vpi_hdl = m_parent->get_handle<vpiHandle>();
         m_iterator = vpi_iterate(vpitype, vpi_hdl);
         if (NULL == m_iterator) {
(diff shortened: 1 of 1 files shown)


ktbarrett:
> Can we easily get the strings `vpiLoad`/`vpiDriver` from the raw `vpitype` (`91`/`93`) that is handed in?

Unfortunately I do not see a way besides creating and maintaining our own map. I don't think we need to do that.

garmin-mjames:
Yeah I think we'd need something like `reason_to_string` in `VpiImpl.cpp` but for types.

cmarqu:
Indeed, creating our own map for this rare occurrence is overkill.
This PR still prints a bit more useful info.

ktbarrett:
Again with the failing... I guarantee you the problem is that too many jobs are being started up simultaneously. We need to modify the private CI to only allow one job for the entire project to run at once. Or allow it a longer time to wait.

ktbarrett (approved):

Issue #1709, opened 2020-04-25
Verilator: Offer option to use FST Tracer


markusdd:
Hi,

very short proposal: Verilator includes FST-Wavedump support for some time now, enabling it would only require an option variabe and instantiating a `VerilatedFstC` instead of the vcd tracer.

FSTs are much more space efficient compared to vcd's .

cmarqu:
The main location that needs a change is probably this:
https://github.com/cocotb/cocotb/blob/66d1cad02c003d3e010993d186439394114041cb/cocotb/share/lib/verilator/verilator.cpp#L37-L43

markusdd:
Yes. For FST, it should something like this (without the switch yet):

```
#if VM_TRACE
    Verilated::traceEverOn(true);

    std::unique_ptr<VerilatedFstC> tfp(new VerilatedFstC);
    top->trace(tfp.get(), 99);
    tfp->open("dump.fst");
#endif
```

bluecmd:
+1! Another really cool feature is the support for enum values.

![image](https://user-images.githubusercontent.com/149442/92406583-a829dd80-f138-11ea-9166-4493d4b2eb34.png)

I tested the above patch and it worked (just needed an extra include as well, and needed to install libz-dev).

bluecmd:
@cmarqu Would you be open for a first time code contribution for this? It doesn't look that hard and it would benefit me so I can give it a shot.

themperek:
One could reuse `VM_TRACE` and add recognistion for `VM_TRACE=fst` or `VM_TRACE=dump.fst` ....

cmarqu:
@bluecmd Yes, sure!

markusdd:
Also, the makefile would need to change as well:

```
╰─ grep -ir Vcd .                                                                                                                                                                                           ─╯
./share/makefiles/simulators/Makefile.verilator:	@rm -f dump.vcd
```

markusdd:
Closing here as my PR was just accepted.

PR #2102, opened 2020-10-06
imlement FST support for verilator

### LINKED ISSUE #2131: Simulation fail when enabling Verilator FST trace
Verilator supports FST format tracing through the --trace-fst option.  However, when this option is added, cocotb fails to build it with the following error:

    /usr/lib/python3.8/site-packages/cocotb/share/lib/verilator/verilator.cpp: In function ‘int main(int, char**)’:
    /usr/lib/python3.8/site-packages/cocotb/share/lib/verilator/verilator.cpp:41:23: error: cannot convert ‘std::unique_ptr<VerilatedVcdC>::pointer’ {aka ‘VerilatedVcdC*’} to ‘VerilatedFstC*’
       41 |     top->trace(tfp.get(), 99);
          |                ~~~~~~~^~
          |                       |
          |                       std::unique_ptr<VerilatedVcdC>::pointer {aka VerilatedVcdC*}
    In file included from /usr/lib/python3.8/site-packages/cocotb/share/lib/verilator/verilator.cpp:5:
    ./Vtop.h:235:31: note:   initializing argument 1 of ‘void Vtop::trace(VerilatedFstC*, int, int)’
      235 |     void trace(VerilatedFstC* tfp, int levels, int options = 0);
          |                ~~~~~~~~~~~~~~~^~~
    make[2]: *** [Vtop.mk:61: verilator.o] Error 1

Using --trace builds fine, but the resulting VCD traces are many GB in size.

markusdd:
This addresses Ticket #1709 

The VM_TRACE define is set by verilator and its value cannot be modified.
Cleanest solution seems to be to add another orthogonal define via CFLAGS.

Documentation is also adapted, because verilator also needs to be called via --trace-fst instead of --trace

I have checked their docs and there is no seperate define being set when FST option is selected, so
we have to do it by ourselves for now.

Closes #2131.

### DIFF
--- a/cocotb/share/lib/verilator/verilator.cpp
+++ b/cocotb/share/lib/verilator/verilator.cpp
@@ -8,6 +8,11 @@
 
 #include <memory>
 
+#ifndef VM_TRACE_FST
+    //emulate new verilator behavior for legacy versions
+    #define VM_TRACE_FST 0
+#endif
+
 #if VM_TRACE
 #if VM_TRACE_FST
 #include <verilated_fst_c.h>
--- a/cocotb/share/lib/verilator/verilator.cpp
+++ b/cocotb/share/lib/verilator/verilator.cpp
@@ -9,7 +9,7 @@
 #include <memory>
 
 #if VM_TRACE
-#if TRACE_FST
+#if VM_TRACE_FST
 #include <verilated_fst_c.h>
 #else
 #include <verilated_vcd_c.h>
@@ -40,7 +40,7 @@ int main(int argc, char** argv) {
 
 #if VM_TRACE
     Verilated::traceEverOn(true);
-#if TRACE_FST
+#if VM_TRACE_FST
     std::unique_ptr<VerilatedFstC> tfp(new VerilatedFstC);
     top->trace(tfp.get(), 99);
     tfp->open("dump.fst");
--- a/cocotb/share/lib/verilator/verilator.cpp
+++ b/cocotb/share/lib/verilator/verilator.cpp
@@ -9,7 +9,11 @@
 #include <memory>
 
 #if VM_TRACE
-# include <verilated_vcd_c.h>
+#if TRACE_FST
+#include <verilated_fst_c.h>
+#else
+#include <verilated_vcd_c.h>
+#endif
 #endif
 
 vluint64_t main_time = 0;       // Current simulation time
@@ -36,10 +40,15 @@ int main(int argc, char** argv) {
 
 #if VM_TRACE
     Verilated::traceEverOn(true);
-
+#if TRACE_FST
+    std::unique_ptr<VerilatedFstC> tfp(new VerilatedFstC);
+    top->trace(tfp.get(), 99);
+    tfp->open("dump.fst");
+#else
     std::unique_ptr<VerilatedVcdC> tfp(new VerilatedVcdC);
     top->trace(tfp.get(), 99);
     tfp->open("dump.vcd");
+#endif
 #endif
 
     while (!Verilated::gotFinish()) {
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -129,7 +129,7 @@ For Verilator 4.102 and above, the `-CFLAGS -DVM_TRACE_FST=1` argument is no lon
 
   .. code-block:: make
 
-    EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DVM_TRACE_FST
+    EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DVM_TRACE_FST=1
 
 The resulting file will be ``dump.fst`` and can be opened by ``gtkwave dump.fst``.
 Note only the most recent versions of gtkwave and verilator support the FST format.
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -131,9 +131,8 @@ For Verilator 4.102 and above, the `-CFLAGS -DVM_TRACE_FST=1` argument is no lon
 
     EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DVM_TRACE_FST
 
-instead. The resulting file will be ``dump.fst`` and can be opened by ``gtkwave dump.fst``.
-Please note that you should use the most recent versions of verilator and gtkwave for FST to be
-properly supported with acceptable performance.
+The resulting file will be ``dump.fst`` and can be opened by ``gtkwave dump.fst``.
+Note only the most recent versions of gtkwave and verilator support the FST format.
 
 .. _sim-verilator-issues:
 
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -132,7 +132,6 @@ For Verilator 4.102 and above, the `-CFLAGS -DVM_TRACE_FST=1` argument is no lon
     EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DVM_TRACE_FST=1
 
 The resulting file will be ``dump.fst`` and can be opened by ``gtkwave dump.fst``.
-Note only the most recent versions of gtkwave and verilator support the FST format.
 
 .. _sim-verilator-issues:
 
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -125,7 +125,7 @@ If you wish to use FST-Tracing (much smaller file size), which is supported by G
 
   .. code-block:: make
 
-    EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DTRACE_FST
+    EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DVM_TRACE_FST
 
 instead. The resulting file will be ``dump.fst`` and can be opened by ``gtkwave dump.fst``.
 Please note that you should use the most recent versions of verilator and gtkwave for FST to be
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -121,7 +121,11 @@ To get waveforms in VCD format, add Verilator's trace option(s) to the
 To set the same options on the command line, use ``EXTRA_ARGS="--trace --trace-structs" make ...``.
 A VCD file named ``dump.vcd`` will be generated in the current directory.
 
-If you wish to use FST-Tracing (much smaller file size), which is supported by GTKWAVE, then use
+Verilator can produce waveform traces in the FST format, the native format of GTKWave. 
+FST traces are much smaller and more efficient to write, but require the use of GTKWave.
+
+To enable FST tracing, add `--trace-fst -CFLAGS -DVM_TRACE_FST=1` to `EXTRA_ARGS` as shown below.
+For Verilator 4.102 and above, the `-CFLAGS -DVM_TRACE_FST=1` argument is no longer necessary.
 
   .. code-block:: make
 
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -121,6 +121,16 @@ To get waveforms in VCD format, add Verilator's trace option(s) to the
 To set the same options on the command line, use ``EXTRA_ARGS="--trace --trace-structs" make ...``.
 A VCD file named ``dump.vcd`` will be generated in the current directory.
 
+If you wish to use FST-Tracing (much smaller file size), which is supported by GTKWAVE, then use
+
+  .. code-block:: make
+
+    EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DTRACE_FST
+
+instead. The resulting file will be ``dump.fst`` and can be opened by ``gtkwave dump.fst``.
+Please note that you should use the most recent versions of verilator and gtkwave for FST to be
+properly supported with acceptable performance.
+
 .. _sim-verilator-issues:
 
 Issues for this simulator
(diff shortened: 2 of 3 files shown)


ktbarrett (commented):
It's kind of a shame you have to do this. It seems like verilator could offer a generic facade and make the type of tracer testable with VM_TRACE.

markusdd:
> It's kind of a shame you have to do this. It seems like verilator could offer a generic facade and make the type of tracer testable with VM_TRACE.

True, although I know that FST can do more than VDC, so the extra class might be warranted to provide a richer interface.

But the define thing definetly should be addressed.

Should I create an upstream issue?

ktbarrett:
@markusdd since you have been working on this, can I ask if trace VCD functionality actually works for you? I can't seem to get it to actual dump anything...

markusdd:
You mean based on this PR or generally?

Have you added the --trace --trace-structs options?

(I use the Makefile style like in the debug repo I made for the other issue, with that it works for me)

ktbarrett:
@markusdd Generally. Running in your repo with a test that just runs the clock, like I have below, results in [dump.vcd.txt](https://github.com/cocotb/cocotb/files/5341165/dump.vcd.txt). Nothing is actually dumped, despite the signal actually changing (otherwise the test wouldn't run to completion). I am using Verilator master with my patches for the bug he have previously discussed. I will try with this branch dumping in FST and see if I get different results tonight.

```python
@cocotb.test()
async def test1(dut):

    async def clock():
        while True:
            dut.clk_w_i <= 1
            await Timer(10, 'ns')
            dut.clk_w_i <= 0
            await Timer(10, 'ns')
    cocotb.fork(clock())

    for i in range(100):
        await RisingEdge(dut.clk_w_i)
```

markusdd:
what does the verilated verilog look like?

Maybe you are the victim of an optimization if there is nothing in the module that would warrant keeping it.

markusdd:
Your test yields this in my environment, so it works:
grafik

ktbarrett:
Fun... is it possible you can write a self checking test for this feature? Since I won't be able to replicate it myself until I track that down.

markusdd:
Question is what such a test would look like.
It is hard to check the validity of a waveform dump without looking at it...

imphil:
> It is hard to check the validity of a waveform dump without looking at it...

There's vcddiff (https://github.com/veripool/vcddiff), used in Verilator's test suite, for example.

markusdd:
That's an interesting find.

How would this integrate with your test flow/env setup.

I'm not familiar with this, the test would probably need to be running a cocotb test wrappen in a script and then compare against an existing vcd (also: wehere would we store the vcd, we don't want this in the repo I guess, those guys are huge)

markusdd:
@ktbarrett what is your take here, how should we setup the testing.
(I do not wanna do something which doesn't fit the scheme and throw it away)

I see 4 issues:

1) How to integrate that test into the flow
2) where to store a sample vcd, these are huge
3) shoudl we rely on something external like vcd-diff
4) does a vcd test belong into this PR?

markusdd:
Ok, so good news, we now have upstream support to get the VM_TRACE_FST define, my merge request was accepted: https://github.com/verilator/verilator/pull/2592

So using the -CFLAGS way remains optional, but is a good way for users who are forced to use a verilator version that is a bit older.

ktbarrett:
@markusdd I'm reconsidering doing that. It is a lot of work and will require us to depend upon non-essential libraries to do regression testing. Since this only affects one simulator doing one type of waveform tracing, I think it's sufficient to just upload pictures of it working on your end. So nevermind 😄.

markusdd:
That is where I'm at as well.
This will be a lot of effort for little gain overall.

I can get you an FST, still I would like to understand why this doesn't work on your end.

ktbarrett:
I am building verilator from source. Is there a particular directive I need to give to `configure` to enable tracing support?

markusdd:
[dump.fst.zip](https://github.com/cocotb/cocotb/files/5373058/dump.fst.zip)

So here is an fst. (had to zip it for guthub to accept it)

markusdd:
> I am building verilator from source. Is there a particular directive I need to give to `configure` to enable tracing support?

Can you check your configure output? It will skip vcd and especially fst tracing support when there are libraries missing.

ktbarrett:
Ubuntu 20.04 install on WSL, I built master as of 1 week ago. I'll try it on Debian unstable later, building verilator takes a while on my old T420...

markusdd:
> Ubuntu 20.04 install on WSL, I built master as of 1 week ago. I'll try it on Debian unstable later, building verilator takes a while on my old T420...

OK WSL might have a few quirks there, that is possible.
I've seen it working under MacOS and Fedora 32.

documentation/source/simulator_support.rst
```diff
@@ -121,6 +121,16 @@ To get waveforms in VCD format, add Verilator's trace option(s) to the
 To set the same options on the command line, use ``EXTRA_ARGS="--trace --trace-structs" make ...``.
 A VCD file named ``dump.vcd`` will be generated in the current directory.
 
+If you wish to use FST-Tracing (much smaller file size), which is supported by GTKWAVE, then use
+
+  .. code-block:: make
+
+    EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DVM_TRACE_FST
```
themperek, line 128: Since https://github.com/verilator/verilator/pull/2592 will it work like this?
```suggestion
    EXTRA_ARGS += --trace-fst --trace-structs
```

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -9,7 +9,11 @@
 #include <memory>
 
 #if VM_TRACE
```
imphil, line 11: Please try to match the new Verilator behavior here, which unconditionally defines VM_TRACE_FST and sets it to 1 or 0. For backwards-compat, introduce a block like this:

```c
#ifndef VM_TRACE_FST
    #define VM_TRACE_FST 0
#endif
```

imphil (commented):
Thanks for your contribution @markusdd. LGTM in general with a couple of small comments. Please have a look at those. Can you also combine all your commits into a single one with a descriptive commit message (see https://github.com/cocotb/cocotb/blob/master/CONTRIBUTING.md#patch-requirements). Please let us know if you're getting lost in git commands, and I'll have a look.

documentation/source/simulator_support.rst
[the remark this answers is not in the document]
themperek: Although at the moment of cocotb realase it may be ok.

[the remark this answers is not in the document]
markusdd: We should keep the -CFLAGS hint.
The newest verilator will not be in major distributions for at least a couple of months, I would at least make mention of this possibility.

documentation/source/simulator_support.rst
```diff
@@ -121,6 +121,16 @@ To get waveforms in VCD format, add Verilator's trace option(s) to the
...
+
+instead. The resulting file will be ``dump.fst`` and can be opened by ``gtkwave dump.fst``.
+Please note that you should use the most recent versions of verilator and gtkwave for FST to be
```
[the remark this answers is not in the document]
markusdd: Yes but in fact does not apply to vcd. vcd is fine on older gtkwave and verilator.
The only real improvements happendes very recently in verialtor and gtkwave (actually gtkwave is more important in that context as it provides the fst library)

cocotb/share/lib/verilator/verilator.cpp
markusdd: does that do anything useful in that context? 

Shouldn't make a difference even when using #if vs. #ifdef?

documentation/source/simulator_support.rst
markusdd: see below.

Yes it does, but I would not move to that just yet, we need to mention you can 'workaround' a slightly older verilator version like this.

cocotb/share/lib/verilator/verilator.cpp
ktbarrett: When you enable FST tracing is `VM_TRACE` defined? I think he is saying you can lift the `if VM_TRACE_FST` out of this `if`.

```c++
#if VM_TRACE_FST
#include <verilated_fst_c.h>
#elif VM_TRACE
#include <verilated_vcd_c.h>
#endif
```

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -36,10 +40,15 @@ int main(int argc, char** argv) {
...
-
+#if VM_TRACE_FST
+    std::unique_ptr<VerilatedFstC> tfp(new VerilatedFstC);
+    top->trace(tfp.get(), 99);
+    tfp->open("dump.fst");
+#else
     std::unique_ptr<VerilatedVcdC> tfp(new VerilatedVcdC);
     top->trace(tfp.get(), 99);
     tfp->open("dump.vcd");
+#endif
```
ktbarrett, line 51: And you could do the same here.

cocotb/share/lib/verilator/verilator.cpp
markusdd: Hi, to quote myself:

> The VM_TRACE define is set by verilator and its value cannot be modified.
Cleanest solution seems to be to add another orthogonal define via CFLAGS.

That was the whole point of using an extra define because VM_TRACE always appears.

cocotb/share/lib/verilator/verilator.cpp
markusdd: see above

cocotb/share/lib/verilator/verilator.cpp
ktbarrett: I'm not sure what the problem is with this code. It looks fine to me. I'm partial to merge this anyways unless @imphil explain a bit better. I'll give it 2 days. This is certainly not the last time we will have to deal with Verilator support, it doesn't have to be perfect now, but hopefully will be before 1.5 release at the end of the year.

imphil: > does that do anything useful in that context?

Yes. The goal is to get exactly the same defines for older and newer Verilator versions.

* For newer Verilator versions, there's unconditionally a define VM_TRACE_FST set to either 0 or 1. 
* For older Verilator versions, you recommend in the docs to set `-DVM_TRACE_FST=1` iff FST tracing is wanted. This means, VM_TRACE_FST is only set to 1 if FST tracing is enabled, and undefined otherwise.

If you add the code snippet I've shown, you get identical behavior.

Now I know that this doesn't make much of a difference given the way the pre-processor treats undefined macros, but I feel it's better to be explicit here.

documentation/source/simulator_support.rst
```diff
@@ -121,6 +121,19 @@ To get waveforms in VCD format, add Verilator's trace option(s) to the
...
 
+Verilator can produce waveform traces in the FST format, the native format of GTKWave. 
+FST traces are much smaller and more efficient to write, but require the use of GTKWave.
+
+To enable FST tracing, add `--trace-fst -CFLAGS -DVM_TRACE_FST=1` to `EXTRA_ARGS` as shown below.
+For Verilator 4.102 and above, the `-CFLAGS -DVM_TRACE_FST=1` argument is no longer necessary.
+
+  .. code-block:: make
+
+    EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DVM_TRACE_FST
```
imphil, line 132:
```suggestion
    EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DVM_TRACE_FST=1
```

to match your advice above.

documentation/source/simulator_support.rst
```diff
@@ -121,6 +121,19 @@ To get waveforms in VCD format, add Verilator's trace option(s) to the
...
+
+The resulting file will be ``dump.fst`` and can be opened by ``gtkwave dump.fst``.
+Note only the most recent versions of gtkwave and verilator support the FST format.
```
imphil, line 135: GtkWave has been supporting FST for ages (ever?), and Verilator supported it also for multiple years (May 2018 actually, see https://github.com/verilator/verilator-announce/issues/17). Just remove this sentence. Earlier versions don't work with cocotb anyways.

markusdd: That is not completely accurate as I've experienced myself. There have been relatively recent updates (meaning this year) which have fixed a few issues that allowed it to be used. Before that performance was abysmal.

But I do not care, we can remove it, I just thought it might be a helpful hint.

documentation/source/simulator_support.rst
markusdd: Fine by me.

imphil (approved):
LGTM, please squash into a single commit with a descriptive commit message before merging.

markusdd:
> 
> 
> LGTM, please squash into a single commit with a descriptive commit message before merging.

will do tonight, working on the wrong machine currently.

ktbarrett:
@markusdd Squashing it something I can do when I merge the PR, I did it on your last one too =)

markusdd:
> 
> 
> @markusdd Squashing it something I can do when I merge the PR, I did it on your last one too =)

even better. Just out of curiosity, can this be done online here in Github? (I'm more of a Gitlab user in my daily job, so I do not know all ins and outs of this here.)

ktbarrett:
I don't think you can do that from the Github UI. Maintainers, when accepting a PR, get to add the change to the commit tree by either merging (creating a merge commit), rebasing all of the commits on master, or squashing all the commits into 1 and rebasing that on master. People can do it manually, but it's equally easy for the Maintainer to do it from the UI. Sometimes they do it because there was a lot of "applied suggestions from code review" commits; or they want to modify/improve a commit message.

markusdd:
> 
> 
> I don't think you can do that from the Github UI. Maintainers, when accepting a PR, get to add the change to the commit tree by either merging (creating a merge commit), rebasing all of the commits on master, or squashing all the commits into 1 and rebasing that on master. People can do it manually, but it's equally easy for the Maintainer to do it from the UI. Sometimes they do it because there was a lot of "applied suggestions from code review" commits; or they want to modify/improve a commit message.

Makes sense, thanks for the explanation.

Issue #2096, opened 2020-10-04
Simulation ends although it should wait 1000 Cycles


markusdd:
I'm currently having a weird issue with cocotb 1.4 and verilator 4.100

I'm verifying an asynchronous FIFO design and have a basic 'manual, section which works perfectly, where I test reads and writes and the empty/full flags etc. then, I tried to do this to generate some more stress:

```python
    # randomize
    cocotb.log.info(f"Test random read/write.")
    cocotb.fork(write_fifo(dut, vals))
    cocotb.fork(read_fifo(dut, vals))
    await ClockCycles(dut.clk_r_i, 1000)
```

So I am launching 2 coroutines, one of which is writing, the other one is reading and checking, both with random delays.

The coroutines look like this:

```python
async def write_fifo(dut, vals):
    w_cnt = 0
    while True:
        cocotb.log.info(f"In write proc.")
        for _ in range(random.randrange(1, 32)):
            await FallingEdge(dut.clk_w_i)
        if dut.full_o == 0:
            cocotb.log.info(f"Random write number {w_cnt}.")
            dut.wen_i <= 1
            dut.wdata_i <= vals[w_cnt]
            await RisingEdge(dut.clk_w_i)
            dut.wen_i <= 0
            w_cnt = w_cnt + 1


async def read_fifo(dut, vals):
    r_cnt = 0
    while True:
        cocotb.log.info(f"In read proc.")
        for _ in range(random.randrange(0, 12)):
            await FallingEdge(dut.clk_r_i)
        if dut.empty_o == 0:
            cocotb.log.info(f"Random read number {r_cnt}.")
            dut.ren_i <= 1
            await RisingEdge(dut.clk_r_i)
            dut.ren_i <= 0
            await FallingEdge(dut.clk_r_i)
            assert dut.rdata_o == vals[r_cnt], f"Read Data mismatch at value {r_cnt}, got {hex(int(dut.rdata_o))} but expected {hex(int(vals[r_cnt]))}"
            r_cnt = r_cnt + 1
```

The weird thing is, the simulation does not wait for the 1000 cycles of the read clock, it just ends:

```
0.00ns INFO     Running test 1/2: test_afifo
     0.00ns INFO     Starting test: "test_afifo"
                     Description:  Test that data is being passed across clock domains.
     0.00ns INFO     Reset AFIFO on both sides.
     5.00ns INFO     Check empty and not full.
     5.00ns INFO     Fill FIFO with random values.
[...]

Issue #1689, opened 2020-04-22
Conditional waiting on events : API discussion


themperek:
Based on https://github.com/cocotb/cocotb/pull/1250 (by @Fatsie) proposals and discussion in https://github.com/cocotb/cocotb/issues/1682

We have discussed a bit internally on https://github.com/cocotb/cocotb/pull/1250 and concluded is a that it is a good idea but:
- function names have to be clear
- we can gain performance if we implement this in a slightly different way 

For performance reasons the function should include as input: 
- synchronization trigger (clock and edge)
- sensitivity list (signals to observe)
- condition for trigger

Implementation of the waiting part should be somewhere along the lines (from @eric-wieser ) :
```python3
while True:
    await RisingEdge(dut.clk)
    if condition():
        break
    await First(*[Edge(hdl) for hdl in signals])
```
We should support signals, vectors and buses? Should we use ReadOnly()?

This way we minimize the number of triggers/callbacks that will improve performance and may become the recommended way to wait for the signal.
In the future, such a trigger can be moved to GPI/C++.

Considering those points we should agree on API before implementation.
Proposals are welcome. https://github.com/cocotb/cocotb/pull/1250 can be a starting point.

ktbarrett:
Just to be clear, what we are talking about is creating a function *specifically* for improving the performance of synchronizing ourselves on registered logic, correct? I think in that case we should specifically call out the use case with the word "registered" in the name. I'm thinking `RegisteredEdge`, `RegisteredRisingEdge`, and `RegisteredFallingEdge`.

I don't really see much of a point in adding the conditional to the call, that conditioning can be in the user's code. Adding the condition will just make it harder to port to C in the future, and there is no potential for optimization at that point. Instead of passing the conditional we could support both direct waits as well as decorator syntax:

```python

# trigger style
await RegisteredRisingEdge(dut.clk, dut.validOut)
result = dut.dataOut.value

#---------------------------

# callback decorator style
@RegisteredRisingEdge(dut.clk, dut.valid)
def read_dataOut():
    return dut.dataOut.value

#---------------------------

# ???
@Always
@RegisteredRisingEdge(dut.clk, dut.valid)
def dataOut_monitor():
    result = dut.dataOut.value
    do_thing(result)

```

eric-wieser:
The callback style can't work, as there's no `await` there.

eric-wieser:
> I don't really see much of a point in adding the conditional to the call, that conditioning can be in the user's code.

[...]

Issue #1682, opened 2020-04-20
Wait on event and condition in `c/c++`?


themperek:
For performance reasons, it would be probably useful to have the ability to wait on condition in `c/c++`.  Looking at monitors/drivers that could have a performance gain if used correctly.

Something like below to wait in `c` and not in `python`:
```python
while condition == x:
    await event
```

Example python syntes can be like in #1250:
```python
await RisingEdge(dut.clock).until_after(lambda: dut.ack.value)
```

I think @eric-wieser @ktbarrett have some ideas.

eric-wieser:
Right now it is impossible to have more than one C-level callback for a signal edge. The issue is that we allocate the callbacks statically per signal:

https://github.com/cocotb/cocotb/blob/c4cfcd3e3faae6ea12851c0336fa92924ffce176/cocotb/share/lib/vpi/VpiImpl.h#L188-L190

I imagine this was motivated by a reluctance to implement proper memory management of these callbacks. Allocating on the heap would be a bad idea unless we were capable of cleaning up after ourselves.

themperek:
But doing like bellow can get the same (with some limitations) :
```python
await RisingEdge(dut.signal)
await RisingEdge(dut.clk)
```

eric-wieser:
That's not correct, it gives a false positive if a pulse appears on `signal` between clock edges. You need
```python
while True:
    await RisingEdge(dut.clk)
    if dut.signal.value.binstr == "1":
        break
    await Edge(dut.signal)
```

themperek:
@eric-wieser But this is not implemented like this: https://github.com/cocotb/cocotb/blob/b5780f3f478ddaff4629cf46e8c998ebebc3b9ad/cocotb/triggers.py#L910-L921 ?

eric-wieser:
That code isn't specific to edges, but to any complex condition. I'm also starting to suspect that `ReadOnly` is wrong, and that the signal should be captured immediately at the rising edge, to match how the verilog would capture it.

eric-wieser:
I'm also starting to suspect that `ReadOnly` is wrong, and that the signal should be captured immediately at the rising edge of `clock`, to match how the verilog would capture it.

themperek:
BTW: For `ClockCycles` one can use `Timer` in simple cases.

ktbarrett:
@eric-wieser

[...]

Issue #1676, opened 2020-04-19
Document performance improvements suggestions 


themperek:
Add a documentation section that can be done for improving performance.

Some initial suggestions can be (please add more):

- Check optimization options of your simulator (default from cocotb may not be optimal for performance).
- Use `async`/`await` syntax.
- Try to minimize the number of triggers (`await`/yield`):
   - Generate clocks on the HDL side
   - Used slowest possible clock/(or use signal/strobe as a trigger) to exchange data between python and HDL
   - Exchange a larger amount of data on one trigger (parallel) than multiple triggers.
- Use profiler (`COCOTB_ENABLE_PROFILING=1 make ... `)

cmarqu:
* Reduce the amount of log messages

nndurj:
Is it possible to import verilated design as ctypes module to avoid interprocess communication? See 
https://github.com/myhdl/myhdl/pull/338. 
>No interprocess communication is used since the verilated design is imported as a ctypes module.

ktbarrett:
Instead of reducing the amount of log messages, we could use a [faster log implementation](https://github.com/brmmm3/fastlogging).

jshaker000:
I think with Xcelium / Questa, a big part of the performance hit is that we have to compile / elaborate the design access +rwc (because cocotb can in theory read / write to ANY net in the whole design). And so the compilers can make less optimizations.   
   
Unfortunately, I dont see a very good way around this. As a test I've run plain SV testbenches both with and without rwc and I see a 2-2.5x speed difference typically.  
   
I guess one solution would be specify in compiling which nets you actually want to read / write specifically and specify only those in access, but then that starts to make the cocotb flow much less appealing to maintain. Its tough ..

cmarqu:
I suppose we could have a "learning mode" where we record every path that is accessed and use this information in a second run - VCS even has such a mode built in.
It would not necessarily work so well with randomization since different, previously unrecorded, paths could be taken.

themperek:
It may need to part of the design to actually know what one is doing. We may help by documenting this and giving (easy) way to change.

PR #2105, opened 2020-10-07
Update the verilator main control loop.

### LINKED ISSUE #2008: Verilator hangs if the simulator exits abnormally
For example issue_1279 test hangs indefinitely after deadlock is invoked. test_exit_error fails after encountering the syntax error that causes cocotb init to fail. Several tests seem to hang, and then just die with the error "Killed".

garmin-mjames:
After some discussion in #2096, here's a stab at improving the verilator main control loop.

Closes #2008. 

Uses solutions for verilator/verilator#2536 and verilator/verilator#2589, so this will bump the minimum supported version of Verilator to 4.106.

- [x] Update supported version in docs
- [x] Newsfragment for better support and version bump

---

- Don't call eval() on every time step, instead skip ahead to the next
registered Timer trigger.

    - All values should be settled after the inner loop, and verilator ignores delays (#) in SystemVerilog files, so the only thing that can cause additional stimulus is a `Timer` trigger. There's no use in iterating over each time step and calling eval(), etc.

    - This should fix some of the "freeze" issues people have seen which are really just *very* long waits filled with useless calls into Verilator.

- If there are no more Timer triggers registered but we end up back in
the loop, end the simulation so we don't get stuck in an infinite loop.

    - I don't know if this is the cause of some of the lockups people like @ktbarrett have seen, but there are cases where cocotb can return to the simulator with no future `Timer` triggers.

- Move the cbAfterDelay callbacks (for Timer triggers) to the beginning of
the time step (Pre-Active region).

    - SystemVerilog standard IEEE 1800-2012 section 4.4 and 4.10 detail when `cbAfterDelay` callbacks should be called, which is before the iterative regions, in the same Pre-Active region as `cbNextSimTime`

### DIFF
--- a/cocotb/share/lib/verilator/verilator.cpp
+++ b/cocotb/share/lib/verilator/verilator.cpp
@@ -21,7 +21,7 @@
 #endif
 #endif
 
-vluint64_t main_time = 0;       // Current simulation time
+static vluint64_t main_time = 0;       // Current simulation time
 
 double sc_time_stamp() {       // Called by $time in Verilog
     return main_time;           // converts to double, to match
@@ -32,6 +32,20 @@ extern "C" {
 void vlog_startup_routines_bootstrap(void);
 }
 
+static inline bool settle_value_callbacks() {
+    bool cbs_called, again;
+
+    // Call Value Change callbacks
+    // These can modify signal values so we loop
+    // until there are no more changes
+    cbs_called = again = VerilatedVpi::callValueCbs();
+    while (again) {
+        again = VerilatedVpi::callValueCbs();
+    }
+
+    return cbs_called;
+}
+
 int main(int argc, char** argv) {
     Verilated::commandArgs(argc, argv);
 #ifdef VERILATOR_SIM_DEBUG
@@ -39,6 +53,10 @@ int main(int argc, char** argv) {
 #endif
     std::unique_ptr<Vtop> top(new Vtop(""));
     Verilated::fatalOnVpiError(false); // otherwise it will fail on systemtf
+
+#ifdef VERILATOR_SIM_DEBUG
+    Verilated::internalsDump();
+#endif
 
     vlog_startup_routines_bootstrap();
     VerilatedVpi::callCbs(cbStartOfSimulation);
@@ -57,42 +75,65 @@ int main(int argc, char** argv) {
 #endif
 
     while (!Verilated::gotFinish()) {
-        bool again = true;
+        // Call registered timed callbacks (e.g. clock timer)
+        // These are called at the beginning of the time step
+        // before the iterative regions (IEEE 1800-2012 4.4.1)
+        VerilatedVpi::callTimedCbs();
+
+        // Call Value Change callbacks triggered by Timer callbacks
+        // These can modify signal values
+        settle_value_callbacks();
 
         // We must evaluate whole design until we process all 'events'
+        bool again = true;
         while (again) {
             // Evaluate design
             top->eval();
 
-            // Call Value Change callbacks as eval()
-            // can modify signals values
-            VerilatedVpi::callValueCbs();
+            // Call Value Change callbacks triggered by eval()
+            // These can modify signal values
+            again = settle_value_callbacks();
 
-            // Call registered Read-Write callbacks
-            again = VerilatedVpi::callCbs(cbReadWriteSynch);
+            // Call registered ReadWrite callbacks
+            again |= VerilatedVpi::callCbs(cbReadWriteSynch);
 
-            // Call Value Change callbacks as cbReadWriteSynch
-            // can modify signals values
-            VerilatedVpi::callValueCbs();
+            // Call Value Change callbacks triggered by ReadWrite callbacks
+            // These can modify signal values
+            again |= settle_value_callbacks();
         }
 
         // Call ReadOnly callbacks
         VerilatedVpi::callCbs(cbReadOnlySynch);
 
-        // Call registered timed callbacks (e.g. clock timer)
-        VerilatedVpi::callTimedCbs();
-
 #if VM_TRACE
         tfp->dump(main_time);
 #endif
-        main_time++;
+        // cocotb controls the clock inputs using cbAfterDelay so
+        // skip ahead to the next registered callback
+        vluint64_t next_time = VerilatedVpi::cbNextDeadline();
+
+        // If there are no more cbAfterDelay callbacks,
+        // the next deadline is max value, so end the simulation now
+        if (next_time == static_cast<vluint64_t>(~0ULL)) {
+            vl_finish(__FILE__, __LINE__, "");
+            break;
+        } else {
+            main_time = next_time;
+        }
 
         // Call registered NextSimTime
-        // It should be called in new slot before everything else
+        // It should be called in simulation cycle before everything else
+        // but not on first cycle
         VerilatedVpi::callCbs(cbNextSimTime);
+
+        // Call Value Change callbacks triggered by NextTimeStep callbacks
+        // These can modify signal values
+        settle_value_callbacks();
     }
 
     VerilatedVpi::callCbs(cbEndOfSimulation);
+
+    top->final();
 
 #if VM_TRACE
     tfp->close();
--- a/cocotb/share/makefiles/simulators/Makefile.verilator
+++ b/cocotb/share/makefiles/simulators/Makefile.verilator
@@ -22,6 +22,13 @@ ifeq ($(shell which $(CMD) 2>/dev/null),)
 $(error Cannot find verilator.)
 endif
 
+VLT_MIN := 4.106
+VLT_VERSION := $(shell $(CMD) --version | cut -d " " -f 2)
+MIN_VERSION := $(shell printf "%s\n%s\n" "$(VLT_MIN)" "$(VLT_VERSION)" | sort -g | head -1)
+ifneq ($(MIN_VERSION),$(VLT_MIN))
+  $(error cocotb requires Verilator $(VLT_MIN) or later, but using $(VLT_VERSION))
+endif
+
 ifeq ($(VERILATOR_SIM_DEBUG), 1)
   COMPILE_ARGS += --debug
   PLUSARGS += +verilator+debug
--- a/documentation/source/newsfragments/2105.change.rst
+++ b/documentation/source/newsfragments/2105.change.rst
@@ -1,0 +1,1 @@
+Improved support and performance for :ref:`sim-verilator` (version 4.106 or later now required).
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -79,21 +79,19 @@ In order to use this simulator, set :make:var:`SIM` to ``verilator``:
 
     make SIM=verilator
 
-cocotb supports Verilator 4.020 and above.
 Verilator converts Verilog code to C++ code that is compiled.
 It does not support VHDL.
 One major limitation compared to standard Verilog simulators is that it does not support delayed assignments.
 
 To run cocotb with Verilator, you need ``verilator`` in your PATH.
 
-Finally, cocotb currently generates a Verilator toplevel C++ simulation loop which is timed at the highest precision.
-If your design's clocks vary in precision, the performance of the simulation can be improved in the same order of magnitude by adjusting the precision in the Makefile, e.g.,
+.. note::
 
-.. code-block:: makefile
-
-    COCOTB_HDL_TIMEPRECISION = 1us # Set precision to 10^-6s
+    cocotb requires Verilator 4.106 or later.
 
 .. versionadded:: 1.3
+
+.. versionchanged:: 1.5 Improved cocotb support and greatly improved performance when using a higher time precision. Verilator 4.106 or later is required.
 
 Coverage
 --------
--- a/tests/test_cases/issue_1279/Makefile
+++ b/tests/test_cases/issue_1279/Makefile
@@ -2,16 +2,6 @@
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
 
-ifeq ($(SIM),verilator)
-
-all:
-	@echo "Skipping test due to Verilator hanging when prematurely shutting down"  # gh-2008
-clean::
-
-else
-
 include ../../designs/sample_module/Makefile
 
 MODULE = issue_1279
-
-endif
--- a/tests/test_cases/test_exit_error/Makefile
+++ b/tests/test_cases/test_exit_error/Makefile
@@ -25,14 +25,6 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
-ifeq ($(SIM),verilator)
-
-all:
-	@echo "Skipping test due to Verilator hanging when prematurely shutting down"  # gh-2008
-clean::
-
-else
-
 # detecting an expected failure has to happen here, as the python code is invalid
 .PHONY: override_for_this_test
 override_for_this_test:
@@ -41,5 +33,3 @@ override_for_this_test:
 include ../../designs/sample_module/Makefile
 
 MODULE = test_exit
-
-endif


garmin-mjames:
Pinging @wallento

imphil:
CC @wsnyder does this look right to you?

ktbarrett:
> * Don't call eval() on every time step, instead skip ahead to the next
>   registered Timer trigger.

I think you are missing value change callbacks causing additional stimulus. In any time step eval() may cause value changes which require servicing. Their servicing might cause other signals to be updated by cocotb. If we don't service the value changes in the correct time step, we miss intended stimulus. I think that our current eval()-every-step behavior is unfortunately unavoidable (which means that Verilator support will always have terrible performance)

> * Move the cbAfterDelay callbacks (for Timer triggers) to the beginning of
the time step (Pre-Active region).
>   * SystemVerilog standard IEEE 1800-2012 section 4.4 and 4.10 detail when `cbAfterDelay` callbacks should be called, which is before the iterative regions, in the same Pre-Active region as `cbNextSimTime`

This is definitely correct.

garmin-mjames:
@ktbarrett I think we are missing `setimmediatevalue` calls in the first `callValueCbs()` in certain instances, where no cached writes occur. 
The fix for that is to return a bool from `callValueCbs()` and see if any callbacks were called that could have caused stimulus, and do another `eval()`.
If any cached writes occur, or any `ReadWrite` trigger is awaited/yielded, `again` is true and we do another `eval()`.

Is there some way stimulus could occur in a time step with no `cbAfterDelay` registered?

ktbarrett:
> Is there some way stimulus or value changes could occur in a time step with no `cbAfterDelay` registered?

Yes. Imagine a test of a FF where the clock is run in HDL with the following test.

```python
@cocotb.test()
async def test_clock(dut):
    dut.d.value = 12
    await RisingEdge(dut.clk)
    assert dut.q.value == 12
```

There are no `Timer`s used in this test.

garmin-mjames:
How is the clock generated in that example? My understanding is that the primary clock(s) have to be generated in the C++ loop.

With the non-modified cocotb loop, that has to be a Python clock using `Timer`s. If you change the C++ loop to generate clocks then you'd need to modify the `main_time` step anyways.

ktbarrett:
> My understanding is that the primary clock(s) have to be generated in the C++ loop.

My familiarity with Verilator is low. Is that a limitation with Verilator? Can you not generate a clock in a testbench with HDL?

> I think we are missing setimmediatevalue calls in the first callValueCbs() in certain instances, where no cached writes occur.
The fix for that is to return a bool from callValueCbs() and see if any callbacks were called that could have caused stimulus, and do another eval().
If any cached writes occur, or any ReadWrite trigger is awaited/yielded, again is true and we do another eval().

This seems correct as well.

In the current code it calls the value callbacks again immediately after servicing ReadWrite, this seems wrong without an eval() first. Should this statement just be removed?

garmin-mjames:
> My familiarity with Verilator is low. Is that a limitation with Verilator? Can you not generate a clock in a testbench with HDL?

I think it basically creates a synthesized model to run. That's part of why it's fast to execute, because it simplifies/combines logic like a synthesis tool.
https://www.veripool.org/wiki/verilator/Manual-verilator#Synthesis-Subset
https://www.veripool.org/wiki/verilator/Manual-verilator#Time

> In the current code it calls the value callbacks again immediately after servicing ReadWrite, this seems wrong without an eval() first. Should this statement just be removed?

I was thinking about this as well. It depends on if you want cocotb value change callbacks caused by other cocotb stimulus to execute before the values cascade through the entire design.

It's definitely a complicated issue in that all of the value change callbacks happen after the iterative `eval()` has settled all values in the design.

ktbarrett:
> It's definitely a complicated issue in that all of the value change callbacks happen after the iterative `eval()` has settled all values in the design.

Yeah. I don't think we are going to be able to have an implementation that is equivalent to other supported simulator's behaviors. As long as it behaves "reasonably", I'm happy. And I'm sure the afflicted users will be as well.

garmin-mjames:
We may want to add another `callValueCbs()` before the `eval()` inner loop so that any `setimmediatevalue()` calls in a `Timer` callback are seen before `eval()`.

I'm not sure if adding `callValueCbs()` everywhere will make behavior closer to other simulators or make it harder to reason about what values are and when they change.

wallento:
Hi, this is very important and key to better performance. I know there is some work in Verilator underway that goes into similar directions. I will have a look at it asap.

ktbarrett:
@garmin-mjames I think what might be beneficial is a set of minimal tests that encode our assumptions about scheduling.

For example:

```python
async def test_timer_before_rising_edge(dut):
    half_period = 5

    dut.clk.value = 0
    async def clock():
        while True:
            await Timer(half_period)
            dut.clk.value = ~int(dut.clk.value)
    cocotb.fork(clock())

    # the clock coro updates the clock value in a timer callback
    # the new value should be applied to the design immediately so clocks appear to run at the appropriate time
    await Timer(half_period)
    # coro scheduling behavior does not guarantee the clock process will run before or after this so we can't test dut.clk.value
    # but the next RisingEdge(clk) should occur at the same time (delay of 0)
    t = utils.get_sim_time()
    await RisingEdge(dut.clk)
    assert dut.clk.value == 1
    assert t == utils.get_sim_time()
```

Of course there is about a million more... They should help us understand our requirements, understand how the loop should be written, and help identify short comings and allow us to make design trade-offs. I really think these tests should be run in separate simulator instances. Perhaps we can even run them in parallel?

themperek:
Just a comment.

If you really care about performance it is also important to generate clocks on HDL/C++ side and call to python only for data exchange (see #1676). I can imagine in the future many projects could only use #1689 (or similar) for performance.

In the case of Verilator if the clock is inside C++ and one response to signals in cocotb everything seems to be fast. This requires a custom loop at the moment. One way (maybe) to make it user friendly would be to have the clock (object) somehow running inside GPI (and not via scheduler)...

ktbarrett:
@themperek #2108

garmin-mjames:
@themperek There was the start of clock creation in GPI but it wasn't finished and got removed in https://github.com/cocotb/cocotb/pull/1429.

themperek:
> @themperek There was the start of clock creation in GPI but it wasn't finished and got removed in #1429.

One would need (probably) to start with multiple callback support: https://github.com/cocotb/cocotb/issues/1682#issuecomment-616508278

garmin-mjames:
I am still working on this. Both verilator/verilator#2588 and verilator/verilator#2605 have been merged, which will help cocotb support verilator.

I'm working on a set of tests to validate simulator scheduling as suggested by @ktbarrett to give us more confidence and hopefully point to some of the limitations when using verilator.

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -22,6 +22,7 @@
 #endif
 
 vluint64_t main_time = 0;       // Current simulation time
+vluint64_t next_time = 0;       // Next time with Timer callback
```
ktbarrett, line 25:
```suggestion
```

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -22,6 +22,7 @@
 #endif
 
 vluint64_t main_time = 0;       // Current simulation time
```
ktbarrett, line 24:
```suggestion
static vluint64_t main_time = 0;       // Current simulation time
```

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -56,44 +75,68 @@ int main(int argc, char** argv) {
...
-        // Call registered timed callbacks (e.g. clock timer)
-        VerilatedVpi::callTimedCbs();
-
 #if VM_TRACE
         tfp->dump(main_time);
 #endif
-        main_time++;
+        // cocotb controls the clock inputs using cbAfterDelay so
+        // skip ahead to the next registered callback
+        next_time = VerilatedVpi::cbNextDeadline();
```
ktbarrett, line 115:
```suggestion
        vluint64_t next_time = VerilatedVpi::cbNextDeadline();
```

ktbarrett (commented):
Looks good.

ktbarrett:
`issue_1279` and deadlock are now handled properly :tada:.
Verilator is now faster than Icarus in the matrix_multiplier :tada:.
`examples/axi_lite_slave` used to run, but hangs now?

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -56,44 +75,68 @@ int main(int argc, char** argv) {
...
         while (again) {
             // Evaluate design
             top->eval();
 
-            // Call Value Change callbacks as eval()
-            // can modify signals values
-            VerilatedVpi::callValueCbs();
+            // Call Value Change callbacks triggered by eval()
+            // These can modify signal values
+            again = settle_value_callbacks();
```
ktbarrett, line 97: If there are every any value change callbacks added in any of the following statements, this will evaluate to true and you will enter an infinite loop. I used the original logic where only the readwrite callbacks were considered for running again, and the examples worked flawlessly.

The loop 
1. Evaluates the model to allow it to react to changes by the user
2. Allow the user to react to value changes by the model and cache up writes of their own.
3. ReadWrite occurs and cached writes are flushed
4. Allow the user to react to value changes by the user and potentially cache more writes
5. Restart from 1.

The process stops when everyone is done changing. There are two exit conditions: no values changed in the model (we don't know that since `eval` doesn't return anything), or the user is done making changes (we use the fact there were no `ReadWrite` callbacks for that, since cached writes are flushed in that stage). Whether the user needed to react to changes or not doesn't matter, what matters is if they changed anything. Because of `setimmediate` this is always possible, but we have to draw a line in the sand somewhere.

garmin-mjames: There's a possibility of an infinite loop if something keeps toggling/glitching back and forth, but I think that scenario can occur in most/all simulators.

When a `cbValueChange` callback is registered, verilator stores the value at that time as the previous value to compare against for a value change. After `callValueCbs` loops through the registered callbacks, those previous values are updated with the changed value. So even new value change callbacks registered during a previous value change callback on the same object should not be called unless the value changes again, between when it's registered and `callValueCbs` is called subsequently.

`axi_lite_slave` isn't getting stuck in the loop in a single time step, it keeps going forever, waiting for the end of the transaction. For some reason the `AXIML_BVALID` signal is not high during the `ReadOnly` callback when `AXI4Master.write` is looking for it. I don't know the root cause of that though.

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -56,44 +75,68 @@ int main(int argc, char** argv) {
...
 
-            // Call registered Read-Write callbacks
-            again = VerilatedVpi::callCbs(cbReadWriteSynch);
+            // Call registered ReadWrite callbacks
+            again = (again || VerilatedVpi::callCbs(cbReadWriteSynch));
```
ktbarrett, line 100: This short circuits, `ReadWrite` is never entered if `settle_value_callbacks` returns `true`.

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -32,6 +33,20 @@ extern "C" {
 void vlog_startup_routines_bootstrap(void);
 }
 
+bool settle_value_callbacks() {
```
ktbarrett, line 36:
```suggestion
static inline bool settle_value_callbacks() {
```

cocotb/share/lib/verilator/verilator.cpp
ktbarrett:
```suggestion
            again = (again | VerilatedVpi::callCbs(cbReadWriteSynch));
```

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -56,44 +75,68 @@ int main(int argc, char** argv) {
...
-            again = VerilatedVpi::callCbs(cbReadWriteSynch);
+            // Call registered ReadWrite callbacks
+            again = (again || VerilatedVpi::callCbs(cbReadWriteSynch));
 
-            // Call Value Change callbacks as cbReadWriteSynch
-            // can modify signals values
-            VerilatedVpi::callValueCbs();
+            // Call Value Change callbacks triggered by ReadWrite callbacks
+            // These can modify signal values
+            again = (again || settle_value_callbacks());
```
ktbarrett, line 104:
```suggestion
            again = (again | settle_value_callbacks());
```

ktbarrett (commented):
Just tested. This works.

ktbarrett:
A better solution would be to have some way to detect if a user wrote during a callback. You might be able to exit the loop more quickly.

cocotb/share/lib/verilator/verilator.cpp
wsnyder, line 100: BTW clangtidy would probably complain that you can simplify "again = (again | ...":

```suggestion
            again |= VerilatedVpi::callCbs(cbReadWriteSynch);
```

cocotb/share/lib/verilator/verilator.cpp
wsnyder, line 104:
```suggestion
            again |= settle_value_callbacks();
```

wsnyder (commented):
Minor nit....

ktbarrett (approved):
Testing looks good. Great job @garmin-mjames and thanks!

garmin-mjames:
@ktbarrett This change will require Verilator 4.105, which isn't released yet, let alone on the stable branch or in package managers.
We could provide dual support for a while through some mechanism.
Pinging @imphil, who noted these concerns in https://github.com/cocotb/cocotb/pull/2102#discussion_r506186389.

ktbarrett:
@garmin-mjames When was the verilator mutating list bug introduced? Presumably there is only a small span of commits/releases that are affected. If that's the case I think it's possible to make this forward compatible using a metafunction to check if `callValueCbs` returns and if it doesn't, don't consider the return value and only call it once (preserving the old behavior of teh value callbacks, but with the `NextTimeStep` and `NextDeadline` updates).

garmin-mjames:
Technically I think 4.104 works for the `bool` return, but the mutating list but isn't fixed until 4.105. The bug goes back to at least 2017, so **proper** support for cocotb is going to require 4.105 regardless.

ktbarrett:
@wsnyder I don't want to push you to release unless you are okay with it. It would be nice to have proper support available in tandem with the cocotb 1.5 release, which will be early next year. So you even have some wiggle room if you wanted to help.

@garmin-mjames Right. Support for cocotb before 2017 was probably not-existent. Should we perhaps add a warning if a version < 4.106 is detected?

garmin-mjames:
@ktbarrett a warning print out would be useful.

imphil:
What is the expected behavior here with regards to Verilator support? Looking at CI I see `VerilatedVpi::cbNextDeadline();` function-not-found errors, so 4.028 is clearly not supported.

We need to ensure that we clearly specify (and test) the supported version(s) of Verilator, and error out otherwise.

I'm fine not supporting anything older than the current Verilator master for the next cocotb release, Verilator support in cocotb was always marked as "experimental/early" so we don't need to worry about backwards-compat too much, but should optimize to get the support improved quickly. (And Verilator is super easy to update [unless you're on a system without a C++11 compiler.])

A version check for a development version (i.e. >= 4.105) is fine for cocotb master, it's almost guaranteed we'll have a released 4.106 version by the time cocotb master is released as next version.

garmin-mjames:
I'm in favor of only supporting 4.106 or later. cocotb callback issues are likely to occur in older versions.

cocotb/share/makefiles/simulators/Makefile.verilator
```diff
@@ -22,6 +22,13 @@ ifeq ($(shell which $(CMD) 2>/dev/null),)
 $(error Cannot find verilator.)
 endif
 
+VLT_MIN := 4.105
+VLT_VERSION := $(shell $(CMD) --version | cut -d " " -f 2)
+MIN_VERSION := $(shell printf "%s\n%s\n" "$(VLT_MIN)" "$(VLT_VERSION)" | sort -g | head -1)
+ifneq ($(MIN_VERSION),$(VLT_MIN))
+  $(error cocotb supports Verilator $(VLT_MIN) or later, but using $(VLT_VERSION))
```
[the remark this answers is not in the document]
garmin-mjames: ?
```suggestion
  $(error cocotb requires Verilator $(VLT_MIN) or later, but using $(VLT_VERSION))
```

wsnyder:
I pushed Verilator 4.106, so you can code for that.  I usually release near the top of the month so this worked out well.

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -57,43 +75,66 @@ int main(int argc, char** argv) {
...
 #endif
-        main_time++;
+        // cocotb controls the clock inputs using cbAfterDelay so
+        // skip ahead to the next registered callback
+        vluint64_t next_time = VerilatedVpi::cbNextDeadline();
+
+        // If there are no more cbAfterDelay callbacks,
+        // the next deadline is max value, so end the simulation now
+        if (next_time == static_cast<vluint64_t>(~0ULL)) {
+            Verilated::gotFinish(true);
```
ktbarrett, line 118: Does Verilator not support `vpi_control(vpiFinish, 0)`? Is that not desirable here? Are `cbEndOfSimulation` callbacks called in this case?

ktbarrett (approved):

cocotb/share/lib/verilator/verilator.cpp
garmin-mjames: `vl_finish()` may be more appropriate, which `vpi_control(vpiFinish, 0)` ends up calling. And we are breaking the loop here anyway.

`vpi_control()` is for controlling the simulator from a VPI application, but this context is the simulator ending the simulation because there isn't any more stimulus.

`vl_finish()` also flushes stdout/stderr and VCD/FST files, so I think that's the better option.

Either way, `cbEndOfSimulation` callbacks are explicitly called below on line 134.

ktbarrett:
@garmin-mjames Is there something else gating this?

garmin-mjames:
@ktbarrett I'll push the `vl_finish()` change shortly.

ktbarrett:
Oh, I missed that. Glad I asked 😄.

ktbarrett (approved):

ktbarrett:
@garmin-mjames What changed?

garmin-mjames:
@ktbarrett rebase on master to get #2265 in order to check Verilator CI.

Issue #2104, opened 2020-10-07
Timer does not protect against invalid input at construction


ktbarrett:
`Timer(0)` is undefined behavior. Whether it blows up or not depends entirely on your simulator and the current mode you are in when you issue it. `Timer` values < 0 are total nonsense. However, neither of these conditions are currently caught in the destructor. They are caught (or not caught at all!) when the underlying VPI, VHPI or FLI callback is registered. This doesn't lead to a Python exception the user can detect or trace back to a particular line of their code. We need to check for this during `Timer` construction.

PR #2090, opened 2020-09-30
Updates to CONTRIBUTING.md

ktbarrett:
Address some of the problems in #2057. Closes #2058. Closes #1642. What I think is still missing is a rather large section, that might be better to put into the Internal Documentation on the Wiki, on contributing to each major codebase (skills, tools, "experts"). That section will not be present in this PR.

Issue #1773, opened 2020-05-06
Consider calling logging.captureWarnings as part of the logging setup


eric-wieser:
Using [`logging.captureWarnings`](https://docs.python.org/3/library/logging.html#logging.captureWarnings) would put `DeprecationWarning`s in the same format as our other warnings.

PR #2107, opened 2020-10-07
check for sane timer value

markusdd:
Address issue #2104


cocotb/triggers.py
```diff
@@ -195,6 +195,8 @@ def __init__(self, time_ps, units=None):
         See Also:
             :func:`~cocotb.utils.get_sim_steps`
         """
+        if time_ps <= 0:
+            raise TriggerException("Timer value time_ps needs to be greater than 0")
```
eric-wieser, line 199: Should we only do this on simulators where it's known to fail?

markusdd: The question is more like: Why would there be any sane scenario where a timer is 0 or negative?

I would consider this an error in any testcase, unless I'm overlooking something very weird.

eric-wieser: Zero seems semi-reasonable to me. Negative numbers I agree are nonsensical.

ktbarrett: > Why would there be any sane scenario where a timer is 0 or negative?

There isn't. Negative `Timer`s are obviously nonsense (time only goes one way). `Timer(0)` is undefined behavior. No standard for any language defines its meaning, and the result is that it behaves differently on every simulator. In some way it is also nonsense. `cbAfterDelay` callbacks are the first callbacks serviced in a time step. A `Timer(0)` issued later than that implies it must go back in time to the start of the time step for servicing.

eric-wieser: My thinking is that allowing 0 might be useful for users wanting to explore quirks in simulators that allow it.

markusdd: But what would the Time(0) semantics look like? From what @ktbarrett listed and what we found in verilator it is more than likely that it would lead to a very unpredictable behavior across simulator implementations.

And given from what I've seen over the years working with Mentor, Viviado Sim, Synopsys VCS and Cadence Incisive and Xcelium such corner cases are usually very unstable across releases. They make new optimizations without taking such stuff into account.

But as I said, I might be overlooking a reasonable use-case, but to me it looks kind of dangerous.

cmarqu:
Could this have a noticeable performance impact in pathological cases?

markusdd:
> Could this have a noticeable performance impact in pathological cases?

I considered using an assert first, but this is indeed a more heavy operation as it definetly would call a different context.
Checking an integer or float for range though should be very very fast.

cocotb/triggers.py
garmin-mjames: The only case where it might work is in a callback in the Pre-Active region that then happens to be correctly handled by the simulator, but it's probably an edge case that's poorly handled.

There is a slightly related case for **cbAtStartOfSimTime** in the SV spec:

> Placing a callback for **cbAtStartOfSimTime** and a delay of zero during a callback for reason **cbAtStartOfSimTime** will result in another **cbAtStartOfSimTime** callback occurring during the same time slice.

That being said I can't find anything about **cbAfterDelay**

garmin-mjames:
@cmarqu one more reason to re-use `Timer` objects.

cocotb/triggers.py
ktbarrett: Since there is some opposition, lets make a concession. Raise on `time < 0`, warn on `time == 0`?

markusdd: Given the literal hell I went through at Intel with VCS and such specifics of DPI/VPI: I would not rely on anything that touches those corners of the spec 😆 
It is almost certainly incorrectly interpreted or implemented or sacrificed for the sake of optimization.

ktbarrett (commented):
You should update the docstring with a "Raises" section, and add a `.. versionadded:: 1.5.0` section at mentioning the change in behavior.

markusdd:
> .. versionadded:: 1.5.0

done

ktbarrett:
@markusdd It looks like a couple of tests in the regression suite are failing because of this. Could you fix them so the regressions pass? `Timer(1)` would work, but moving to the `async`/`await` syntax would be preferable, so we don't have to have those statements in the tests at all.

cocotb/triggers.py
cmarqu: > Since there is some opposition, lets make a concession. Raise on `time < 0`, warn on `time == 0`?

Agreed, we actually have ``Timer(0)`` in at least one of our test.

cmarqu:
> @cmarqu one more reason to re-use `Timer` objects.

Reminds me, we should have a section about performance in our docs.

markusdd:
> @markusdd It looks like a couple of tests in the regression suite are failing because of this. Could you fix them so the regressions pass? `Timer(1)` would work, but moving to the `async`/`await` syntax would be preferable, so we don't have to have those statements in the tests at all.

Ok we need to decide now. Do we only warn on Timer(0) and not change the tests or do we forbid Timer(0)?

cocotb/triggers.py
eric-wieser: @cmarqu, does "our" mean "cocotb" or "another project"?

If that latter, that's basically what my argument is - people may have found that in the simulator they use `Timer(0)` does something useful for them, and this exception will break their code for little reason. A warning will make them aware that their code will likely fail on other simulators, but will not force them to fix it.

cmarqu: I meant cocotb.

markusdd: I pushed a modified version which splits between negative and 0.
I kept the 'toplevel' if and split below that, as in the majority of the cases this check should not fire.
This saves us from having 2 toplevel 'ifs' for the sake of performance.

cocotb/triggers.py
```diff
@@ -194,7 +194,19 @@ def __init__(self, time_ps, units=None):
...
+        Raises:
+            TriggerException: If a negative value is passed for Timer setup.
+
+        .. versionadded:: 1.5
+            Raise an exception when Timer uses a negative value as it is undefined behavior.
+            Warn for 0 as this will cause erratic behavior in some simulators as well.
         """
+        if time_ps <= 0:
+            if time_ps == 0:
+                self.log.warning("Timer setup with value 0, which might exhibit undefined behavior in some simulators")
```
eric-wieser, line 207: This will fail because `self.log` doesn't exist yet.

I think I'd lean towards using `warnings.warn("...", stacklevel=2)` here instead.

markusdd: It did work here in Travis CI:
grafik

eric-wieser: Ah, you're right - I changed how this worked

The testability argument still applies though. #1773 would make my suggestion appear as a normal log line too.

markusdd: I'm not quite sure I get your point, don't we have a test right now with what I have pasted above?

Or do you have something else/additional in mind?

(feel free to make the edit or give the exact line you would like to see, I have no real preference)

eric-wieser: It's not a proper test if it can't fail :). You can use `warnings.catch_warnings` or similar to record what warnings happen, and then `assert` the expected one is present.

cocotb/triggers.py
```diff
@@ -194,7 +194,19 @@ def __init__(self, time_ps, units=None):
...
+            else:
+                raise TriggerException("Timer value time_ps must not be negative")
```
eric-wieser, line 209: On that note, it would be good to add a test for this too

cocotb/triggers.py
ktbarrett:
```suggestion
               warnings.warn(
                   message="Timer setup with value 0, which might exhibit undefined behavior in some simulators", 
                   category=RuntimeWarning,
                   stacklevel=2)
```

You will need to import `warnings` at the top.

ktbarrett: Mark, you will see the codecov bot posting coverage results. It shows certain paths not being tested. If you add a new feature, you need to add a test to cover your addition, at a minimum. See the `tests/test_cases/test_cocotb` directory. There are a couple examples of tests that catch warnings in `test_deprecated`. I suggest you use the `assert_raises` context manager defined there in your test as well. Those tests should go in `test_timing_triggers.py` in that directory.

ktbarrett: I'm sorry this isn't better spelt out yet, we are working on it (#2090).

markusdd: Yep, so for the warning we should be set.
I decided to do it directly in this helper function, so the check will be carried out whenever this is used.

cocotb/triggers.py
markusdd: Done.
grafik

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -12,7 +12,7 @@
 """
 import cocotb
 import warnings
-from cocotb.triggers import Timer, RisingEdge, ReadOnly, ReadWrite, Join, NextTimeStep
+from cocotb.triggers import Timer, RisingEdge, ReadOnly, ReadWrite, Join, NextTimeStep, TriggerException
 from cocotb.utils import get_sim_time
 from cocotb.result import TestFailure
 from cocotb.clock import Clock
```
eric-wieser, line 18:
```suggestion
from cocotb.clock import Clock

from common import assert_raises
```

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -248,3 +248,9 @@ async def test_singleton_isinstance(dut):
     assert isinstance(NextTimeStep(), NextTimeStep)
     assert isinstance(ReadOnly(), ReadOnly)
     assert isinstance(ReadWrite(), ReadWrite)
+
+
+@cocotb.test(expect_fail=True, expect_error=TriggerException)
+def test_neg_timer(dut):
+    """Test intentionally failing by requesting negative timer value"""
+    yield Timer(-42)
```
eric-wieser, line 256: Better to verify the exception is thrown rather than that the test fails:
```suggestion
@cocotb.test()
async def test_neg_timer(dut):
    """Test intentionally failing by requesting negative timer value"""
    with assert_raises(TriggerException):
        Timer(-42)  # no need to even `await`, constructing it is an error
```

markusdd: Doesn't expect_error=TriggerException do exactly this?

markusdd: Or I should say: then probably some other tests in that suite should be rewritten as well.

ktbarrett: We intending to use that feature only for exempting tests that fail on certain simulators.

>  some other tests in that suite should be rewritten as well.

Absolutely they do, but there are a freaking lot of them that need to be rewritten... That is outside the scope of this PR, so don't worry about them for right now.

markusdd: Getting your point for the expect_fail parameter, but how would that apply to expect_error.

From what I read its feature set basically exactly mirrors assert_raises and in my view it is much more elegant that way.

Because now we introduce 2 flavors of intentionally failing tests, which looks a bit inconsistent:

a) some tests tell you right away in the decorator they will fail and do so intentionally
b) some other tests look as if they did something 'useful', but truly they also fail, but you can only tell after looking at the code

Bit complicated to explain, you get my point?

markusdd: Also: I think cocotb produces a very nice error output by sying: 'failing as expected, Result was TriggerException'

In the other instance the test would just pass, almost implying you could use negative timer values.

eric-wieser: There are two simple answers for why `assert_raises` is a better method:

* It's the python way - it's how `pytest` and `unittest` do things
* It lets you test multiple errors in a single test
  ```python
      with assert_raises(TriggerException):
          Timer(-42)
      with assert_raises(TriggerException):
          Timer(-1)
  ```

eric-wieser: > some tests tell you right away in the decorator they will fail and do so intentionally

Following the example of `pytest` and `unitest`, the only reasons tests should fail intentionally are:

* You know the test doesn't yet pass, but you'd like it to pass in future
* The test started breaking and you want to put off fixing it till later.

Semantically, "the test failed" and "my test that an exception was thrown succeeded" are very different things.

markusdd: Ok, so you are saying you basically wish to move to method b) completely, unless there is some technical reason to let a test fail. Which makes complete sense to me, because then it would be consistent.

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -244,3 +250,10 @@ async def test_singleton_isinstance(dut):
     assert isinstance(NextTimeStep(), NextTimeStep)
     assert isinstance(ReadOnly(), ReadOnly)
     assert isinstance(ReadWrite(), ReadWrite)
+
+
+@cocotb.test()
+async def test_neg_timer(dut):
+    """Test intentionally failing by requesting negative timer value"""
```
eric-wieser, line 257:
```suggestion
    """Test negative timer values are forbidden"""
```

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -113,7 +116,10 @@ def do_test_afterdelay_in_readonly(dut, delay):
     global exited
     yield RisingEdge(dut.clk)
     yield ReadOnly()
-    yield Timer(delay)
+    with warnings.catch_warnings(record=True) as w:
+        yield Timer(delay)
+        if delay == 0:
+            assert "Timer setup with value 0, which might exhibit undefined behavior in some simulators" in str(w[-1].message)
```
eric-wieser, line 122: Can you move this code to `test_neg_timer` instead? The test that calls this coroutine is marked `expect_fail=...` below, so this line probably never gets executed.

markusdd: It gets executed. Question is: is that a good idea?
The test it now resides in makes the special handling for the simualtor already, the test_beg_timer doesn't, as we just discussed.

Can this cause an unintentional crash?

markusdd: I did the change...let's see how it goes.

markusdd: Ok that doesn't look good:
grafik

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -256,5 +256,5 @@ async def test_neg_timer(dut):
         Timer(-42)  # no need to even `await`, constructing it is an error
     # handle 0 special case
     with warnings.catch_warnings(record=True) as w:
-        Timer(0)
+        await Timer(0)
```
eric-wieser, line 259: There is no need to `await` here, the warning is emitted without it.

eric-wieser: (the test is failing because you are using `self.log.warn` and not `warnings.warn`)

cocotb/triggers.py
eric-wieser: Marking as unresolved because you need @ktbarrett's suggested code

tests/test_cases/test_cocotb/test_timing_triggers.py
markusdd: yes but I felt it would make more sense as construction does succeed in this case and this is how it would be used practically, but I don't care. (I'm aware this is not the failure cause)

markusdd: will revert if you wish

eric-wieser: When doing warning tests, it's best to put as little code as possible in the context manager. In principle we could do
```
with warnings.catch_warnings(record=True) as w:
    t = Timer(0)
    assert ...

await t
```
which would both check that the warning is emitted, and that the resulting trigger works. Remember though that the whole point in the warning is that on some simulators `await Timer(0)` does _not_ work, and so if we tested that our tests would fail.

cocotb/triggers.py
```diff
@@ -194,7 +195,21 @@ def __init__(self, time_ps, units=None):
 
         See Also:
             :func:`~cocotb.utils.get_sim_steps`
+
+        Raises:
+            TriggerException: If a negative value is passed for Timer setup.
+
+        .. versionadded:: 1.5
```
eric-wieser, line 202:
```suggestion
        .. versionchanged:: 1.5
```

tests/test_cases/test_cocotb/test_timing_triggers.py
markusdd: Mhh. Ok I admit this is a bit beyond my knowledge now, why does the CI fail on newer versions of Python?
It seems an Exception is generated which wasn't there before.

I'm apparently not aware of a particular change between versions here. Any hints?

ktbarrett: It's a CI thing. Any uncaught warnings are turned into exceptions. You need to fix all invocations of `Timer(0)`.

eric-wieser: Or wrap them in a warnings context manager to avoid the warning

markusdd: But why does this depend on the Python version? That's what I'm not getting.

markusdd: And what other context manager is there for this, aren't we catching the warning already and recording it?
(sorry if I sound stupid, I'm not an expert on the python logging module)

ktbarrett: It's only run on Python 3.8. I forget exactly why this is and where it is set, @eric-wieser do you remember?. It's bitten me before.

In this case you definitely want to remove all of the `Timer(0)`. They are only used to satisfy the requirement that generators need a `yield` statement in them. The offending tests should have that line removed, and should be converted to `async`/`await` syntax. Sorry, deprecations are never fun :)

markusdd: Ok, but how do we test then? I need to throw that warning at least in the new test for this PR.

And I don't blame any of you, I know how ugly this can get.
At my workplace we also maintain a large Python codebase that always needs to maintain legacy
for at leat 3-4 years, which can be painful^^

ktbarrett: [This](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONWARNINGS. ) has the answer.

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -244,3 +247,14 @@ async def test_singleton_isinstance(dut):
...
+
+@cocotb.test()
+async def test_neg_timer(dut):
+    """Test negative timer values are forbidden"""
+    with assert_raises(TriggerException):
+        Timer(-42)  # no need to even `await`, constructing it is an error
+    # handle 0 special case
+    with warnings.catch_warnings(record=True) as w:
+        Timer(0)
+        assert "Timer setup with value 0, which might exhibit undefined behavior in some simulators" in str(w[-1].message)
```
garmin-mjames, line 260:
```suggestion
        assert "Timer setup with value 0, which might exhibit undefined behavior in some simulators" in str(w[-1].message)
        assert issubclass(w[-1].category, RuntimeWarning)
```

tests/test_cases/test_cocotb/test_timing_triggers.py
markusdd: Interesting. Didn't know that.

So which approach should be used? Turn it off, demote the warning?

tests/test_cases/test_cocotb/test_timing_triggers.py
markusdd: good point!

tests/test_cases/test_cocotb/test_timing_triggers.py
ktbarrett: By design we run with `exception`, you will need to update the test code so that it isn't tripped. That is the only solution.

garmin-mjames: If that's not clear, the idea is that no other tests should raise the `RuntimeWarning` except this one, which catches it.
So the other tests need to be fixed so that they don't use `Timer(0)`.

eric-wieser: What @ktbarrett said isn't quite true. There are two possibilities in the tests:

* The test is using `Timer(0)` for no good reason. Change it to `Timer(1)`.
* The test is deliberated testing the undefined behavior of `Timer(0)`. Change it to
  ```python
  with warnings.catch_warnings():
      zero_timer = Timer(0)
  await zero_timer
  ```
  local warning settings override global ones.

markusdd: Sorry, I don't get your point.

How should I test it without firing the warning?
Scrap the test we talked about the last few hours?
You quoted a CLI documentation, which I understand turns wanrings into exceptions, which makes sense for testing in certain scenarios.
But in this case we want the warning, but do not wish to turn of the exception conversion?

Do you wish to make a code suggestion? I'm kind of lost here and did not expect this to turn into such a huge discussion, was kind of just tryin to help out here while waiting for the initial verilator issue to be fixed to continue my project...

eric-wieser: > How should I test it without firing the warning?

That isn't the test that's failing.

markusdd: > > How should I test it without firing the warning?
> 
> The test you added isn't the test that's failing.

Isn't it? Then I'm not able to read the CI output correctly:
grafik

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -244,3 +247,15 @@ async def test_singleton_isinstance(dut):
...
     assert isinstance(ReadWrite(), ReadWrite)
+
+
+@cocotb.test()
+async def test_neg_timer(dut):
+    """Test negative timer values are forbidden"""
+    with assert_raises(TriggerException):
+        Timer(-42)  # no need to even `await`, constructing it is an error
+    # handle 0 special case
+    with warnings.catch_warnings(record=True) as w:
```
eric-wieser, line 258:
```suggestion
    with warnings.catch_warnings(record=True) as w:
        warnings.simplefilter("ignore")
```

cocotb/triggers.py
```diff
@@ -194,7 +195,21 @@ def __init__(self, time_ps, units=None):
...
+            Warn for 0 as this will cause erratic behavior in some simulators as well.
         """
+        if time_ps <= 0:
+            if time_ps == 0:
+                warnings.warn("Timer setup with value 0, which might exhibit undefined behavior in some simulators",
+                              category=RuntimeWarning,
+                              stacklevel=2)
+            else:
+                raise TriggerException("Timer value time_ps must not be negative")
         GPITrigger.__init__(self)
```
eric-wieser, line 213: If you don't call `GPITrigger.__init__` first, then you get a weird exception in `Trigger.__del__`
```suggestion
        GPITrigger.__init__(self)
        if time_ps <= 0:
            if time_ps == 0:
                warnings.warn("Timer setup with value 0, which might exhibit undefined behavior in some simulators",
                              category=RuntimeWarning,
                              stacklevel=2)
            else:
                raise TriggerException("Timer value time_ps must not be negative")
```

tests/test_cases/test_cocotb/test_timing_triggers.py
eric-wieser: Sorry, you were right all along - it was your test that was failing, but it's easy to fix.

markusdd: ok, thank you!

I was kinda questioning my sanity a bit^^

cocotb/triggers.py
markusdd: ah, alright. That makes sense. (not easy to spot though for someone non-familiar)

tests/test_cases/test_cocotb/test_timing_triggers.py
ktbarrett: @eric-wieser The only occurrences are case 2. And in those cases they only exist to satisfy the definition of a generator. They are truly pointless statements in the wake of `async`. That's why I recommended removal instead of `Timer(1)`.

tests/test_cases/test_cocotb/test_timing_triggers.py
markusdd: Mhhh, still fail
grafik
Do we need to move the ignore filter after the assert check?

tests/test_cases/test_cocotb/test_timing_triggers.py
markusdd: In my view those 2 tests are identical, except for one doing something illegal, which is the Time(0).
So should we scrap 1 of the tests completely?
grafik

ktbarrett: Ahhhh there is another test. `grep -r` is not a panacea. Personal opinion is the first test should be scrapped. I don't see a reason for us to test a potentially invalid operation in the regression.

markusdd: I would agree, especially now that we emit a warning.

I've said it before: Such stuff is a rabbit-hole of infinite depth, especially with commercial simualtors. They will not look left and right for such stuff in order to maintain competitive performance.

So yes, I would agree to scrap it, but let's wait for some core maintainers.

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -244,3 +247,16 @@ async def test_singleton_isinstance(dut):
...
+        Timer(0)
+        assert "Timer setup with value 0, which might exhibit undefined behavior in some simulators" in str(w[-1].message)
+        assert issubclass(w[-1].category, RuntimeWarning)
+        warnings.simplefilter("ignore")
```
eric-wieser, line 262: This has to be first in the `with`

eric-wieser, line 262:
```suggestion
    with warnings.catch_warnings(record=True) as w:
        warnings.simplefilter("always")
        Timer(0)
        assert "Timer setup with value 0, which might exhibit undefined behavior in some simulators" in str(w[-1].message)
        assert issubclass(w[-1].category, RuntimeWarning)
```

eric-wieser: Sorry for the pain here - this is exactly why I would prefer to use `with pytest.warns` rather than rolling our own!

markusdd: :D yes, because the other version was also fail
grafik

markusdd:
This is turning into a rabbit hole: now we have 2 more fails:
https://travis-ci.org/github/cocotb/cocotb/jobs/734125358

garmin-mjames:
The other `yield Timer(0)` in tests are probably fixed by #1705 and I missed them in #1776 cleanup.

Also, test `discover_value_not_in_dut` is marked as PASS since it errored, but since it's:
```py3
@cocotb.test(expect_error=True)
def discover_value_not_in_dut(dut):
```
it doesn't check the error type. Another reason to explicitly check in the test.

markusdd:
> The other `yield Timer(0)` in tests are probably workarounds fixed by #1705 and I missed them in #1776 cleanup.
> 
> Also, test `discover_value_not_in_dut` is marked as PASS since it errored, but since it's:
> 
> ```python
> @cocotb.test(expect_error=True)
> def discover_value_not_in_dut(dut):
> ```
> 
> it doesn't check the error type. Another reason to explicitly check in the test.

OK, so will you clean this up and we rebase here once merged?

markusdd:
I had a look and I think I understood what needs to be done, let's see.

markusdd:
Ok, nice, now we have another issue.

The common.py, which contains assert_raises lives in the test_cocotb directory.
The test to be fixed though lives inside test_discovery.

If everything should eventually be using assert_raises, this probably should be restructured?

grafik

eric-wieser:
Indeed. I think in this case though, that `discover_value_not_in_dut` should probably just go in the `test_handle.py` file - indeed, I suspect such a test is already there.

eric-wieser:
Yeah, I recommend you merge `discover_value_not_in_dut` into this test:

https://github.com/cocotb/cocotb/blob/89e7bd77b1e9ceb3172b4659e267bc4c42217a0f/tests/test_cases/test_cocotb/test_handle.py#L36-L44

markusdd:
> Yeah, I recommend you merge `discover_value_not_in_dut` into this test:
> 
> https://github.com/cocotb/cocotb/blob/89e7bd77b1e9ceb3172b4659e267bc4c42217a0f/tests/test_cases/test_cocotb/test_handle.py#L36-L44

actually I think those tests are in fact identical. We can just remove the one in the discover suite.

markusdd:
ok cool. Travis passed. we also covered both cases, warning+exception and cleaned up the spurious Timer(0) occurences.

good to go?

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -34,8 +34,12 @@ def test_lessthan_raises_error(dut):
 
 
 @cocotb.test()
-def test_bad_attr(dut):
-    yield cocotb.triggers.NullTrigger()
+async def test_bad_attr(dut):
+    await cocotb.triggers.NullTrigger()
```
ktbarrett, line 38:
```suggestion
```

This only existed previously to make it a valid generator. It isn't doing anything here.

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -154,19 +157,20 @@ def test_cached_write_in_readonly(dut):
...
+# def test_afterdelay_in_readonly(dut):
+#     """Test doing invalid sim operation"""
+#     global exited
+#     exited = False
+#     clk_gen = cocotb.fork(Clock(dut.clk, 100).start())
+#     coro = cocotb.fork(do_test_afterdelay_in_readonly(dut, 0))
+#     yield [Join(coro), Timer(1000)]
+#     clk_gen.kill()
+#     if exited is not True:
+#         raise TestFailure
```
ktbarrett, line 173:
```suggestion
```

We don't let dead commented out code lie around.

ktbarrett (commented):
Just one last review.

tests/test_cases/test_cocotb/test_timing_triggers.py
markusdd: yes I just left it like this to await some further comments but to move on to get this fixed finally. we can remove now.

markusdd: accepted it, but also removed to additional empty lines, otherwise py linters will complain

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -34,8 +34,11 @@ def test_lessthan_raises_error(dut):
 
 
 @cocotb.test()
-def test_bad_attr(dut):
-    yield cocotb.triggers.NullTrigger()
+async def test_bad_attr(dut):
+
+    with assert_raises(AttributeError):
+        fake_signal = dut.fake_signal
+
```
ktbarrett, line 41: This is testing the exact same thing below. You probably could delete one or the other.

ktbarrett (commented):
I'm happy. I'll let @eric-wieser make the final determination.

tests/test_cases/test_cocotb/test_handle.py
eric-wieser: One is testing a heirarchy object, the other an attribute of a nested signal object. I think having both tests is probably a good thing.

markusdd: I was not quite sure, because the other one goes down two levels, I didn't check the RTL if this is supposed to test some record/interface access or so.

What do you prefer, keep the old or the new one? (if, I would vote for new, because apparantely assert_raises should be the new standard.)

markusdd: > One is testing a hierarchy object, the other an attribute of a nested signal object. I think having both tests is probably a good thing.

same thought same time :D

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -154,21 +157,6 @@ def test_cached_write_in_readonly(dut):
...
-def test_afterdelay_in_readonly(dut):
-    """Test doing invalid sim operation"""
-    global exited
-    exited = False
-    clk_gen = cocotb.fork(Clock(dut.clk, 100).start())
-    coro = cocotb.fork(do_test_afterdelay_in_readonly(dut, 0))
-    yield [Join(coro), Timer(1000)]
-    clk_gen.kill()
-    if exited is not True:
-        raise TestFailure
```
eric-wieser, line 169: So, my reading of this test is that it's intended to check that cocotb can recover and keep running the remaining tests if we ask the sim to do something illegal.

I'd therefore argue we should be keeping this test, and we just need to make sure the warning never happens. That would be straightforward to do with
```
with warnings.capture_warnings():
    warnings.simplefilter('ignore')
    t = Timer(0)
...
coro = cocotb.fork(do_test_afterdelay_in_readonly(dut, t))
```

markusdd: I think we do not need that particular Timer(0) scenario for that.

There are other tests that do exactly what you describe:
grafik

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -154,21 +157,6 @@ def test_cached_write_in_readonly(dut):
...
-
-
 @cocotb.test()
 def test_afterdelay_in_readonly_valid(dut):
     """Same as test_afterdelay_in_readonly but with valid delay > 0"""
```
eric-wieser, line 162: If @ktbarrett doesn't agree with https://github.com/cocotb/cocotb/pull/2107/files#r502474208 and we leave the `test_afterdelay_in_readonly ` test deleted, then we need to adjust this docstring to not talk about something which doesn't exist.

tests/test_cases/test_cocotb/test_timing_triggers.py
eric-wieser: That's a good point. Unless @ktbarrett agrees with me, this is fine to leave out.

markusdd: as he originally proposed to take it out and we agreed above I assume that is ok.

markusdd:
Ok, so Travis is happy. Azure CI I cannot judge.

tests/test_cases/test_cocotb/test_handle.py
ktbarrett: I see. Yeah it's appropriate to leave it.

ktbarrett:
@markusdd The Azure CI is busted right now, it's nothing on you.

ktbarrett (approved):

ktbarrett:
@eric-wieser Final comments?

ktbarrett:
@markusdd Unfortunately there is a conflict. After you rebase this I'll merge.

markusdd:
merged.

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -68,6 +69,32 @@ async def discover_module_values(dut):
...
+    You can now send this process into the background with "Ctrl-Z bg" and run
+        jupyter console --existing
+    or
+        jupyter qtconsole --existing
+    or
+        jupyter console --existing kernel-{}.json
+    ###############################################################################""".format(os.getpid())))
+    IPython.embed_kernel()
+
+
```
ktbarrett, line 97: These were removed in another PR, we don't want them back.
```suggestion
```

markusdd: wasn't sure. but we also need to clean up imports.

markusdd:
Finally, that was a hard one xD

PR #2115, opened 2020-10-08
Add doc section about performance considerations

cmarqu:


documentation/source/performance.rst
```diff
@@ -0,0 +1,37 @@
...
+          :caption: More efficient ``Timer`` object re-use
+
+          halfperiod = Timer(0.5, "ns")
+          while True:
+              await halfperiod
+              signal <= ~signal
+
+This is because creating an object has a small additional overhead compared to using an existing object.
+
+
```
ktbarrett, line 32:
```suggestion
Performance Impact: Minor

```

documentation/source/performance.rst
```diff
@@ -0,0 +1,37 @@
...
+Clock Generator in HDL
+======================
+
+Generate your clock with a HDL module instead of using :class:`~cocotb.clock.Clock`.
+This way, there is no constant updating the simulator with the new value of the clock.
```
ktbarrett, line 37:
```suggestion
Do so prevents calls back into Python every single clock half-period.

Performance Impact: **MAJOR**
```

cmarqu:
I'm leaning towards making this a subsection of a "Best Practices" section actually, in order to not clutter our sidebar/1st level TOC too much.

documentation/source/best_practices.rst
```diff
@@ -0,0 +1,50 @@
...
+
+In order to get the maximum performance out of a cocotb testbench,
+the following points may help you.
+
+
+Clock Generator in HDL
+----------------------
+
+Generate your clock with a HDL module instead of using :class:`~cocotb.clock.Clock`.
+Doing so prevents calls back into Python every single clock half-period.
```
ktbarrett, line 20: It might be worth expanding this. For example, what are best practices about how to implement that? I'd recommend they use a wrapper instead of adding behavior clock generation to their design. And that has implications, `TOPLEVEL` will have to change, and so the dut parameter will change as well.

Maybe that's all obvious?

documentation/source/best_practices.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Performance Impact: **Major**
+
+
+Reusing Objects
+---------------
+
+Do not create single-use objects in a loop but instead create them outside the loop and reuse them.
+Typical cases are :class:`~cocotb.triggers.Timer` or :class:`~cocotb.clock.Clock` objects as shown below:
```
ktbarrett, line 29:
```suggestion
Typical cases are :class:`~cocotb.triggers.Timer` or :class:`~cocotb.triggers.RisingEdge` of the clock signal, as shown below:
```

You shouldn't create multiple `cocotb.clock.Clock`s at all, but `RisingEdge(clk)` is very common.

documentation/source/best_practices.rst
```diff
@@ -0,0 +1,50 @@
...
+
+.. list-table::
+
+   * - .. code-block:: python
+          :caption: Inefficient ``Timer`` object re-creation
+
+          while True:
+              await Timer(0.5, "ns")
+              signal <= ~signal
```
ktbarrett, line 38:
```suggestion
              signal <= ~signal.value
```

the handle class does not have the `__invert__` overload

documentation/source/best_practices.rst
```diff
@@ -0,0 +1,50 @@
...
+
+     - .. code-block:: python
+          :caption: More efficient ``Timer`` object re-use
+
+          halfperiod = Timer(0.5, "ns")
+          while True:
+              await halfperiod
+              signal <= ~signal
```
ktbarrett, line 46:
```suggestion
              signal <= ~signal.value
```

documentation/source/best_practices.rst
cmarqu: Ah, that's a small bug also in our existing docs then.

documentation/source/best_practices.rst
cmarqu: Indeed, changed.

documentation/source/best_practices.rst
cmarqu: Done, also in ``coroutines.rst``.

documentation/source/best_practices.rst
cmarqu: Not obvious, we should actually have a whole example for this.
I'll see what I can do.

cmarqu:
Depends on #2192 which should be referenced in this section.

PR #2116, opened 2020-10-08
Adapt number of discovered objects for Cadence Xcelium

cmarqu:
Current master finds more objects.


ktbarrett:
Just to make sure, this is due to #2079 and not a simulator version bump?

cmarqu:
I explicitly tested an older simulator version, yes, and while I haven't reverted #2079 to cross-check, it's very likely due to it.

Hmm. I now see in https://github.com/cocotb/cocotb/pull/2079#issuecomment-699127430 that I need to check/update more tests; tomorrow.

ktbarrett (approved):

cmarqu:
Ran the tests; I'll work on a full pass with Xcelium at some point.

PR #2119, opened 2020-10-09
Remove skipping test_array with Icarus Verilog

cmarqu:
Not much HDL is left that Icarus actually sees but it's better than nothing - and enables us to test more with future Icarus versions.


tests/test_cases/test_array/test_array.py
```diff
@@ -400,7 +403,9 @@ async def test_direct_constant_indexing(dut):
     _check_type(tlog, dut.const_cmplx[1].b[1], ConstantObject)
 
 
-@cocotb.test()
+@cocotb.test(
+    # Icarus 10.3 doesn't support bit-selects, see https://github.com/steveicarus/iverilog/issues/323
+    expect_error=IndexError if cocotb.SIM_NAME.lower().startswith("icarus") else False)
```
ktbarrett, line 408: Icarus 11.0 has been released and supports bit selects. Should we predicate it on the version number in `cocotb.SIM_VERSION`?

ktbarrett (commented):
Looks like it can compile `sig_t2`, but it can't find it via the VPI. I'm wondering if that's now possible in 11.0, or a VPI bug we haven't been able to test?

tests/test_cases/test_array/test_array.py
cmarqu: Let's close #1987 in the next days one way or the other and use it here.

cmarqu: I fixed it up a bit now, but the `sig_t2` thing still needs to be checked.

cmarqu: Even in 12.0, `sig_t2` is not seen in Icarus, but I wonder if it is a matter of optimizing those away see also https://github.com/steveicarus/iverilog/issues/322). There is `sig_t1` too which is just a 1-dim vector wire that is also not seen.
I'll try using those wires for something in HDL and check more.

cmarqu: That was it indeed. Current problem: https://github.com/cocotb/cocotb/runs/1542442627?check_suite_focus=true#step:22:656 (ping @garmin-mjames)

garmin-mjames: Icarus flattens multi-dimensional arrays into a single dimension, so a 4x4 matrix becomes 1x16.

cmarqu: Thanks, something to be documented and a special case to be introduced.

PR #2125, opened 2020-10-14
Update issue templates

cmarqu:


ktbarrett (commented):
So what will this look like once merged? Since there is no config.yml there won't be an "issue type" drop down, right? `name` and `about` are not displayed anywhere, but simply used for identification in the settings? There will be no default title, label, or assignee, it will simply add the comment about the user environment to the issue body?

cmarqu:
> So what will this look like once merged? Since there is no config.yml there won't be an "issue type" drop down, right? `name` and `about` are not displayed anywhere, but simply used for identification in the settings? 

I don't know, and the preview they offer is not helpful there. We will just have to try it and back out again if needed.

> There will be no default title, label, or assignee, it will simply add the comment about the user environment to the issue body?

Yes, I assume so.

ktbarrett (approved):

ktbarrett:
Well lets see...

PR #2129, opened 2020-10-16
Fix VHPI cast issue

### LINKED ISSUE #2126: Fails to read negative integer values from VHDL testbenches
When reading integer values from a VHDL testbench, negative values are treated as if they are unsigned. Note that reading an integer value does not read a BinaryValue but instead a ConstantObject. I've attached a simple example to reproduce the issue.

[integer_read_fail.zip](https://github.com/cocotb/cocotb/files/5384567/integer_read_fail.zip)

ktbarrett:
Closes #2126.

`vhpiIntT` before the 2008 standard was defined as an unsigned integer in
vhpi_user.h. However, the standard states that all integers are 32-bit
integers with two's complement representation.

The issue occurs when `vhpiIntT` is defined as `uint32_t` and `long` is a
signed 64 bit integer. Negative values are represented as large unsigned
values in two's complement. When this value is returned as a long, it is
implicitly converted ("safely") to the larger signed representation.

To prevent this issue we convert the `vhpiIntT` into a 32-bit two's
complement integer type (`int32_t`) before allowing the conversion to
take place.

### DIFF
--- a/documentation/source/newsfragments/2129.bugfix.rst
+++ b/documentation/source/newsfragments/2129.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed issue on Mac OS and Linux where negative integers were returned as large positive values.
--- a/documentation/source/newsfragments/2129.bugfix.rst
+++ b/documentation/source/newsfragments/2129.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed issue on Mac OS and Linux where negative integers were returned as large positive values.
+Fixed an issue with VHPI on Mac OS and Linux where negative integers were returned as large positive values.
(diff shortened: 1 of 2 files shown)


documentation/source/newsfragments/2129.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed issue on Mac OS and Linux where negative integers were returned as large positive values.
```
cmarqu, line 1:
```suggestion
Fixed an issue with VHPI on Mac OS and Linux where negative integers were returned as large positive values.
```
?

cmarqu (approved):

PR #2132, opened 2020-10-17
Replace Travis with Github Actions

ktbarrett:
As discussed in the maintainers meeting. The runtime of the tests with Github Actions is 3x faster than with Travis. I have also increased the number of tests to cover a larger set of the matrix of OSes and simulators. And this lays the groundwork for even more tests.

I did not move the PyPI deployment over. That can be done in another PR, or in this one if I can get an API key for PyPI.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -11,65 +11,278 @@ on:
 
 jobs:
 
+  lint-whitespace:
+    runs-on: ubuntu-latest
+    name: whitespace
+    strategy:
+      fail-fast: false
+    steps:
+    - uses: actions/checkout@v2
+    - name: whitespace
+      run: |
+        ! git --no-pager grep --color --line-number --full-name --extended-regexp -I '\\s+$'
+
+  lint-flake8:
+    runs-on: ubuntu-latest
+    name: flake8
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.5, 3.9]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{matrix.python-version}}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{matrix.python-version}}
+    - name: flake8
+      run: |
+        pip install flake8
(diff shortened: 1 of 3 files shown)


cmarqu:
It's not easy to see the ``may_fail`` flag in the GHA output (AFAICS), maybe put it into the name until they improve that?

themperek (approved):
Looks OK to me.

.github/workflows/regression-tests.yml
```diff
@@ -11,65 +11,278 @@ on:
...
+        python-version: ${{matrix.python-version}}
+    - name: flake8
+      run: |
+        pip install flake8
+        flake8
+
   tests:
-    name: ${{matrix.sim}} (${{matrix.sim-version}}) | ${{matrix.os}} | Python ${{matrix.python-version}}
+
+    name: ${{matrix.extra_name}}${{matrix.sim}} (${{matrix.sim-version}}) | ${{matrix.os}} | Python ${{matrix.python-version}} ${{matrix.may_fail || '| May Fail'}}
```
themperek, line 45: Does it show `true` when `may_fail` is false?

ktbarrett: You can see for yourself below. It does not. It shows 'May Fail' at the end or nothing.

themperek: I think have extra `true`.
"Regression Tests / icarus (master) | ubuntu-20.04 | Python 3.8 **true** (pull_request)" 

It does not matter.

ktbarrett: You are right, I totally got it backwards. I'll fix it.

imphil:
Can you also resolve the warnings around add-path/set-path, e.g. 

"The `add-path` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/"

imphil (approved):
LGTM once you have it passing CI :-)

Issue #2093, opened 2020-10-01
Support for SystemVerilog Assertions (SVA)


gtaylormb:
Cocotb seems to have very limited support for SVA at this point. If I create an assertion:

```
  ERROR_condition:
    assert property (@(posedge clk) 
      condition
    );
```

The simulator would normally log the hierarchical name of the scope in which the assertion failed as well as the ``__FILE__ `` and `` __LINE__ `` to give you context. Cocotb gives you:

`# 101990.00ns ERROR    *** NO MESSAGE PROVIDED ***`

No context for the assertion failure and the test passes.

```
  ERROR_condition:
    assert property (@(posedge clk) 
      condition
      )
    else
      $error("%m in %s at line %0d", `__FILE__, `__LINE__); 
```

Cocotb gives you:

`# 101990.00ns ERROR    %m in %s at line %0d`

$error doesn't handle formatting input (as reported in #130), test still passes.

```
  ERROR_condition:
    assert property (@(posedge clk) 
      condition
      )
    else begin
      $error("Assertion failure");
      $display("%m in %s at line %0d", `__FILE__, `__LINE__);      
    end
```

Cocotb gives you:

```
# 102340.00ns ERROR    Assertion failure
# instance.ERROR_condition in path/module.sv at line 176
```
Now we've hacked our way to getting the assertion context but the test still passes.

```
  ERROR_condition:
    assert property (@(posedge clk) 
      condition
      )
    else begin
      $display("%m in %s at line %0d", `__FILE__, `__LINE__);      
      $fatal("Assertion failure");
    end
```

This crashes the whole simulator (I'm using Questa), so the test will fail in a regression suite. Unfortunately if you want to use the GUI to debug the assertion failure you're out of luck.

The main 2 issues I see are: 

- Cocotb does not log any context for errors during simulation, and 

- errors during simulation do not cause the test to fail (possibly reported in #911)

ktbarrett:
We are willing to accept any recommendations or changes to better support SVA's. There was another user interested in it (#2043).  I have no idea what the requirements are, or how assertions are instrumented. Please advise.

garmin-mjames:
IEEE 1800-2012 Clause 39 has details on assertion API, including VPI callbacks you can register.
40.5 has details on assertion coverage.

jjj11x:
> The main 2 issues I see are:
> 
> * ...
> * errors during simulation do not cause the test to fail (possibly reported in #911)

[...]

Issue #2094, opened 2020-10-01
Introduce new system tasks to end the test with pass or fail


ktbarrett:
Following the issues in #2093 and discussion #1722, I think we need to introduce two new system tasks, whose names _don't_ overload builtin system tasks and whose purpose is to end the currently running cocotb test with either a pass or fail. Example names could be `$cocotb_pass` and `$cocotb_fail`.

This can be added onto the current system of system task declarations in VpiImpl.cpp. They would issue `gpi_sim_event` with appropriate error codes and messages that would be fed up to the regression manager / scheduler.

gtaylormb:
Additionally, it would be useful if these behaved similarly to the SystemVerilog severity system tasks, from IEEE Std 1800-2012 20.20:

> Each of the severity system tasks can include optional user-defined information to be reported. The userdefined message shall use the same syntax as the $display system task (see 21.2.1) and thus can include
> any number of arguments.
> All of the severity system tasks shall print a tool-specific message, indicating the severity of the exception
> condition and specific information about the condition, which shall include the following information:
> — The file name and line number of the severity system task call. The file name and line number shall
> be the same as `__FILE__ and `__LINE__ compiler directives, respectively.
> — The hierarchical name of the scope in which the severity system task call is made.
> — For simulation tools, the simulation run time at which the severity system task is called.
> The tool-specific message shall include the user-defined message if specified.

ktbarrett:
@gtaylormb I recommended removing cocotb's severity system task overloads instead (#2095) and allowing the simulator to handle severity reporting as it normally would. The only thing the user would have to do to make the cocotb test fail is issue `$cocotb_fail` in an else statement, like you were originally doing with `$fatal` (but it won't crash the sim).

```
  ERROR_condition:
    assert property (@(posedge clk) 
      condition
      )
    else
      $cocotb_fail(); 
```

Would that work for you?

gtaylormb:
Yes I think that would work. If $cocotb_fail can report context where the error occurred in addition to failing the test, all the better. Otherwise I could do something like this:

```
  ERROR_condition:
    assert property (@(posedge clk) 
      condition
      )
    else begin
      $error();
      $cocotb_fail(); 
    end
```

which in theory would give the context from the default simulator $error implementation and then fail the test. Probably have to wrap $cocotb_fail in a `define to allow the RTL to be parsed outside of the cocotb environment.

garmin-mjames:
[...]

PR #2133, opened 2020-10-17
Remove the builtin system task overloads for VPI

### LINKED ISSUE #130: System-call implementations do not handle format values

### LINKED ISSUE #2095: Deprecate and remove the current set of system tasks
The currently registered system tasks in verilog code are `$info`, `$warning`, `$error`, and `$fatal`. These system tasks are also defined by the Verilog standard. Many simulators don't allow you to overload certain system tasks.

From Icarus [README](https://github.com/steveicarus/iverilog/blob/c4883da334344da016e5af9cc5ce58e2f02d9582/README.txt#L414-L415), emphasis mine.
>  Certain of the system functions have well-defined meanings, so can theoretically be evaluated at compile-time, instead of using runtime VPI code. Doing so means that VPI cannot override the definitions of functions handled in this manner. ... __Implementations of these system functions in VPI modules will be ignored.__

Aldec simulators don't like it (from [here](https://github.com/cocotb/cocotb/issues/1722#issuecomment-702362301))
> The Aldec simulators ignore the cocotb overloading and use the built-in ones.

And Mentor doesn't allow overloading `$fatal` (from [here](https://github.com/cocotb/cocotb/issues/2093#issue-713085580)).

> This crashes the whole simulator (I'm using Questa), so the test will fail in a regression suite.
```
  ERROR_condition:
    assert property (@(posedge clk) 
      condition
      )
    else begin
      $display("%m in %s at line %0d", `__FILE__, `__LINE__);      
      $fatal("Assertion failure");
    end
```

While `$fatal` is designed to fail the current test, the rest existed solely to overload logging... And does so poorly (#130). The overloaded implementations also get in the way of the use of the system tasks for SVA (#2093).

These system task overloads should be removed.

ktbarrett:
Closes #130. Closes #2095.

These system tasks overload tasks defined by in the Verilog LRM. This
makes many simulators very unhappy; they do not allow you to overload
some or none of the functions and either do not work, or protest
(#2095). These system tasks never properly parsed additional arguments
correctly (#130, #2093). Also they were never supported with either the
VHPI or FLI interfaces. Their main purpose was to hook HDL reporting
into the Python logger. Since the overload is global, this is fairly
dubious design decision to begin with.

If this behavior is supported and desired, we leave it to a GPI
extension to reintroduce.


ktbarrett:
Removed `test_closedown`.

garmin-mjames:
Is the idea to re-use `SIM_TEST_FAIL` in a later PR closing #2094?

ktbarrett:
@garmin-mjames That is what I had intended. It would be in a separate PR and in the system task definition I would call `gpi_sim_event` to signal test ends to the scheduler.

garmin-mjames (commented):
Without `test_closedown` there's nothing testing `$fatal`/`$finish` explicitly anymore.
The outcome should be the same as `issue_1279`, but it would be nice to have test coverage for it.
Also `close_module` isn't needed without a test.

ktbarrett:
@garmin-mjames Added a Verilog and VHDL test that explicitly test failing the simulation with a behavioral HDL statement. `$fatal(1, ...)` in Verilog, `report "..." severity FAILURE;` in VHDL.

garmin-mjames (commented):
Looks good.

ktbarrett:
Looks like Icarus 10.3 does not set the exit code to the value given to `$fatal`, but 11.0 does... Guess I need to instead ignore the error code and just read results.xml. I can probably reuse `./bin/combine_results.xml for that.

tests/test_cases/test_fatal/Makefile
```diff
@@ -0,0 +1,22 @@
...
+	VERILOG_SOURCES := fatal.sv
+else ifeq ($(TOPLEVEL_LANG),vhdl)
+	VHDL_SOURCES := fatal.vhd
+endif
+
+# squash error code from simulator and ensure the cocotb test passed
+.PHONY: override_for_this_test
+override_for_this_test:
+	-$(MAKE) all
+	../../../bin/combine_results.py &> /dev/null
```
garmin-mjames, line 20: I think this exits `0` if no `results.xml` is found.

ktbarrett: Added the call to `check_for_results_file`.

ktbarrett:
Active-HDL doesn't like being shut down in this way. 

```
# COUT:      0.01ns ERROR    cocotb.scheduler                            __init__.py:280  in _sim_event                      Failing test at simulator request before test run completion: Simulator shutdown prematurely
# COUT: Traceback (most recent call last):
# COUT:   File ".../cocotb/__init__.py", line 281, in _sim_event
# COUT:     scheduler.finish_scheduler(SimFailure(msg))
# COUT:   File ".../cocotb/scheduler.py", line 865, in finish_scheduler
# COUT:     self._test.abort(exc)
# COUT:   File ".../cocotb/decorators.py", line 337, in abort
# COUT:     cocotb.scheduler.unschedule(self)
# COUT:   File ".../cocotb/scheduler.py", line 492, in unschedule
# COUT:     self.cleanup()
# COUT:   File .../cocotb/scheduler.py", line 885, in cleanup
# COUT:     raise Exception("Cleanup() called outside of the main thread")
# COUT: Exception: Cleanup() called outside of the main thread
# COUT:      0.01ns ERROR    cocotb.gpi                                gpi_embed.cpp:348  in embed_sim_event                 Passing event to upper layer failed
```

This isn't the first time we have seen issue with this.

garmin-mjames:
Yeah, I think the Aldec simulators call the shutdown callback on a different thread.
You could skip it like `issue_1279` via #1859 until that's resolved.

ktbarrett:
Tell me if I'm wrong, but calling into a VPI cosimulation from different threads sounds pretty dangerous in general and not just our application. Thread-local storage is a thing... This should be bubbled up to Aldec.

garmin-mjames:
What happened to the Riviera CI?

garmin-mjames (approved):

PR #2134, opened 2020-10-17
Add global alias for DUT parameter

### LINKED ISSUE #1827: Global DUT handle
Right now we supply a reference to the root `SimHandle` (aka the DUT object) as the sole argument to every test. There are some disadvantages to this, the biggest being the DUT object is not available from module-level code. I noticed this in https://github.com/cocotb/cocotb/pull/1502#discussion_r425582308, where information about the DUT is used at the module-level. I myself have had a need for this in the past, but simply worked around it.

Instead, I propose we make the DUT object globally available from the cocotb module: `cocotb.dut`. We now have access to the DUT object at the module-level. This is safe because the DUT cannot change from test to test. We could eventually stop supplying the DUT as an argument to the test functions, which aligns our tests better with other testing frameworks like pytest.

We would have to be careful because we could end up with module-level code modifying the DUT. We would want all tests in a module to run before loading the next module. That means we cannot pre-discover all functions to calculate the total test count. We could still do counts on a module level and have the summary print the total number of tests run. Or we could make the decision that we are not allowed to modify the DUT at the module-level, which can be enforced.

ktbarrett:
This feature will allow users to parameterize module-level classes and functions, and arguments to `TestFactory`s. This was mentioned in https://github.com/cocotb/cocotb/pull/1502#discussion_r425582308.

Closes #1827.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -129,6 +129,16 @@ See :envvar:`RANDOM_SEED` for details on how the value is computed.
 _library_coverage = None
 """ used for cocotb library coverage """
 
+top = None  # type: cocotb.handle.SimHandleBase
+r"""
+A handle to the :envvar:`TOPLEVEL` entity/module.
+
+This is equivalent to the :term:`DUT` parameter given to cocotb tests, so it can be used wherever that variable can be used.
+It is particularly useful for extracting information about the :term:`DUT` in module-level class and function definitions;
+and in parameters to :class:`.TestFactory`\ s.
+``None`` if :mod:`cocotb` was not loaded from a simulator.
+"""
+
 
 def fork(coro: Union[RunningTask, Coroutine]) -> RunningTask:
     """ Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on its use. """
@@ -239,11 +249,12 @@ def _initialise_testbench(argv_):
     if not handle:
         raise RuntimeError("Can not find root handle ({})".format(root_name))
 
-    dut = cocotb.handle.SimHandle(handle)
+    global top
+    top = cocotb.handle.SimHandle(handle)
 
     # start Regression Manager
     global regression_manager
-    regression_manager = RegressionManager.from_discovery(dut)
+    regression_manager = RegressionManager.from_discovery(top)
     regression_manager.execute()
 
     _rlock.release()
--- a/documentation/source/newsfragments/2134.feature.rst
+++ b/documentation/source/newsfragments/2134.feature.rst
@@ -1,0 +1,1 @@
+The handle to :envvar:`TOPLEVEL`, typically seen as the first argument to a cocotb test function, is now available globally as :data:`cocotb.top`.
--- a/documentation/source/building.rst
+++ b/documentation/source/building.rst
@@ -45,6 +45,9 @@ Cocotb
     Use this to indicate the instance in the hierarchy to use as the :term:`DUT`.
     If this isn't defined then the first root instance is used.
 
+    The DUT is available in cocotb tests as a Python object at :data:`cocotb.top`;
+    and is also passed to all cocotb tests as the :ref:`first and only parameter <quickstart_creating_a_test>`.
+
 .. envvar:: RANDOM_SEED
 
     Seed the Python random module to recreate a previous test stimulus.
--- a/documentation/source/library_reference.rst
+++ b/documentation/source/library_reference.rst
@@ -369,6 +369,8 @@ Other Runtime Information
 
 .. autodata:: cocotb.LANGUAGE
 
+.. autodata:: cocotb.top
+
 
 Signal Tracer for WaveDrom
 --------------------------
--- a/documentation/source/quickstart.rst
+++ b/documentation/source/quickstart.rst
@@ -78,6 +78,8 @@ Python test script to load.
 We would then create a file called ``test_my_design.py`` containing our tests.
 
 
+.. _quickstart_creating_a_test:
+
 Creating a test
 ---------------
 


cocotb/__init__.py
```diff
@@ -128,6 +128,9 @@ def _reopen_stream_with_buffering(stream_name):
 _library_coverage = None
 """ used for cocotb library coverage """
 
+top = None  # type: SimHandleBase
+"""A handle to the :make:var:`TOP` entity. ``None`` if :mod:`cocotb` was not loaded from a simulator"""
```
cmarqu, line 132:
```suggestion
"""A handle to the :envvar:`TOPLEVEL` entity. ``None`` if :mod:`cocotb` was not loaded from a simulator"""
```

eric-wieser: It would be good if this could not only describe _what_ this variable is, but _when_ it is useful and a better choice than `dut` arguments.

ktbarrett: I added more detail about when to use it.

cmarqu: My suggested change still applies, or am I misunderstanding something?

cocotb/__init__.py
```diff
@@ -128,6 +128,14 @@ def _reopen_stream_with_buffering(stream_name):
 _library_coverage = None
 """ used for cocotb library coverage """
 
+top = None  # type: SimHandleBase
+"""
+A handle to the :make:var:`TOPLEVEL` entity. ``None`` if :mod:`cocotb` was not loaded from a simulator.
+This is equivalent to the :term:`DUT` parameter given to cocotb tests, so it can be used wherever that variable can be used.
+It is particularly useful for extracting information about the :term:`DUT` in module-level class and function definitions;
+and in parameters to :class:`~cocotb.regression.TestFactory`s.
```
garmin-mjames, line 136: `TestFactory` isn't displaying properly in the docs. This is what the examples page uses:
```suggestion
and in parameters to :class:`.TestFactory`s.
```

cocotb/__init__.py
```diff
@@ -128,6 +128,14 @@ def _reopen_stream_with_buffering(stream_name):
 _library_coverage = None
 """ used for cocotb library coverage """
 
+top = None  # type: SimHandleBase
+"""
+A handle to the :make:var:`TOPLEVEL` entity. ``None`` if :mod:`cocotb` was not loaded from a simulator.
```
garmin-mjames, line 133: I think @cmarqu is right. It's not linking in the docs, but works in the newsfragment.
```suggestion
A handle to the :envvar:`TOPLEVEL` entity. ``None`` if :mod:`cocotb` was not loaded from a simulator.
```

ktbarrett: Whoops missed that too.

cocotb/__init__.py
ktbarrett: Lets try it. I am not sure why that works. The relative path implies the current module is `cocotb.regression`, but I don't see that being set anywhere.

cocotb/__init__.py
```diff
@@ -128,6 +128,14 @@ def _reopen_stream_with_buffering(stream_name):
 _library_coverage = None
 """ used for cocotb library coverage """
 
+top = None  # type: SimHandleBase
+"""
+A handle to the :envvar:`TOPLEVEL` entity. ``None`` if :mod:`cocotb` was not loaded from a simulator.
+This is equivalent to the :term:`DUT` parameter given to cocotb tests, so it can be used wherever that variable can be used.
+It is particularly useful for extracting information about the :term:`DUT` in module-level class and function definitions;
+and in parameters to :class:`~cocotb.regression.TestFactory`s.
```
cmarqu:
```suggestion
A handle to the :envvar:`TOPLEVEL` object.

This is equivalent to the :term:`DUT` parameter given to cocotb tests, so it can be used wherever that variable can be used.
It is particularly useful for extracting information about the :term:`DUT` in module-level class and function definitions;
and in parameters to :class:`~cocotb.regression.TestFactory`s.
``None`` if :mod:`cocotb` was not loaded from a simulator.
```
Separating the first line with a blank line will make this look better in e.g. IDE tooltips.
I'm not sure if the "object" in the first line is good, but "entity" might lead someone into thinking this is for VHDL only.

garmin-mjames: Could use either "entity/module" or "entity or module"

cocotb/__init__.py
cmarqu: Maybe this works?
```
:class:`TestFactories <cocotb.regression.TestFactory>`
```

cmarqu: The problem is likely the "s" appended directly to the ReST markup which breaks the parsing.

cocotb/__init__.py
ktbarrett: Is it possible for `TOPLEVEL` to be something other than an entity or module?

cmarqu: A ``configuration`` in VHDL or ``config`` in Verilog (the latter being rather rare).

cmarqu:
Doc rendering is broken now: https://cocotb--2134.org.readthedocs.build/en/2134/library_reference.html#cocotb.top

cocotb/__init__.py
```diff
@@ -128,6 +128,16 @@ def _reopen_stream_with_buffering(stream_name):
 _library_coverage = None
 """ used for cocotb library coverage """
 
+top = None  # type: cocotb.handle.SimHandleBase
+"""
+A handle to the :envvar:`TOPLEVEL` entity/module.
+
+This is equivalent to the :term:`DUT` parameter given to cocotb tests, so it can be used wherever that variable can be used.
+It is particularly useful for extracting information about the :term:`DUT` in module-level class and function definitions;
+and in parameters to :class:`~cocotb.regression.TestFactory`s.
```
cmarqu, line 137:
```suggestion
and in parameters to :class:`TestFactories <cocotb.regression.TestFactory>`.
```

cocotb/__init__.py
eric-wieser: You can use `` :class:`.TestFactory`\ s`` to fix that

cocotb/__init__.py
```diff
@@ -128,6 +128,16 @@ def _reopen_stream_with_buffering(stream_name):
 _library_coverage = None
 """ used for cocotb library coverage """
 
+top = None  # type: cocotb.handle.SimHandleBase
+"""
+A handle to the :envvar:`TOPLEVEL` entity/module.
+
+This is equivalent to the :term:`DUT` parameter given to cocotb tests, so it can be used wherever that variable can be used.
+It is particularly useful for extracting information about the :term:`DUT` in module-level class and function definitions;
+and in parameters to :class:`TestFactories <cocotb.regression.TestFactory>`.
```
eric-wieser, line 137:
```suggestion
and in parameters to :class:`.TestFactory`\ s.
```

documentation/source/newsfragments/2134.feature.rst
```diff
@@ -0,0 +1 @@
+Made the handle to :envvar:`TOPLEVEL`, typically seen as the first argument to a cocotb test function, available globally as :data:`cocotb.top`.
```
eric-wieser, line 1:
```suggestion
The handle to :envvar:`TOPLEVEL`, typically seen as the first argument to a cocotb test function, is now available globally as :data:`cocotb.top`.
```

cocotb/__init__.py
```diff
@@ -128,6 +128,16 @@ def _reopen_stream_with_buffering(stream_name):
 _library_coverage = None
 """ used for cocotb library coverage """
 
+top = None  # type: cocotb.handle.SimHandleBase
+"""
```
ktbarrett, line 132:
```suggestion
r"""
```

cmarqu (approved):

cmarqu:
Oh, actually, what would be nice is to also have a reference to ``cocotb.top`` from the description of the ``TOPLEVEL`` envvar.

cmarqu (approved):

PR #2136, opened 2020-10-18
Prevent warnings from collection of TestFactories by unit testing frameworks

alexforencich:
Prevent pytest from issuing a PytestCollectionWarning when using TestFactory

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -654,6 +654,9 @@ class TestFactory:
     the test description) includes the name and description of each generator.
     """
 
+    # Prevent warnings from collection of TestFactories by unit testing frameworks.
+    __test__ = False
+
     def __init__(self, test_function, *args, **kwargs):
         if sys.version_info > (3, 6) and inspect.isasyncgenfunction(test_function):
             raise TypeError("Expected a coroutine function, but got the async generator '{}'. "


ktbarrett:
You are going to have to be far more descriptive than that. What does pytest have to do with cocotb? What are the details behind this bug?

alexforencich:
When I run a cocotb testbench that imports `TestFactory` using `cocotb-test`, `pytest` tries to collect `TestFactory` since it starts with the word `Test` and since it has an `__init__` method it results in a `PytestCollectionWarning`.  The solution is to add `__test__ = False` inside `TestFactory` to prevent `pytest` from attempting to collect it.

cocotb/regression.py
```diff
@@ -654,6 +654,9 @@ class TestFactory:
     the test description) includes the name and description of each generator.
     """
 
+    # Prevent PytestCollectionWarning
```
cmarqu, line 657:
```suggestion
    # Prevent PytestCollectionWarning - do not try to collect "TestFactory" despite it matching "Test*"
```

cocotb/regression.py
```diff
@@ -654,6 +654,10 @@ class TestFactory:
     the test description) includes the name and description of each generator.
     """
 
+    # Prevent PytestCollectionWarning from pytest
```
ktbarrett, line 657:
```suggestion
    # Prevent PytestCollectionWarning from cocotb-test
```

ktbarrett (approved):
I'm fine with this. We do have to place nice with extensions.

cocotb/regression.py
cmarqu: Strictly speaking, ``__test__`` also works with ``unittest`` and ``nose``.

ktbarrett: Then maybe something more general?
```suggestion
    # Prevent warnings from collection of TestFactories by unit testing frameworks.
```

alexforencich: Yeah, that's a much better description.

themperek:
@alexforencich Thank you. First commit ✨ !

cmarqu:
@alexforencich Thanks for this PR, and also all your other open source work!

PR #2138, opened 2020-10-19
Implement case-insensitive bus signal name matching

### LINKED ISSUE #2137: Case-insensitive bus signal name matching
Some coding conventions use all caps for port names, some use all lowercase for port names.  Currently, any module that uses bus (BusDriver, BusMonitor, etc.) forces the use of one convention over the other due to how the signal name matching is implemented.  Implementing case-insensitive signal name matching would therefore be quite useful as it would enable operation across both coding styles.  

A simple way to do this would be to create a case-insensitive mapping inside of `Bus.__init__` like so:

    signal_name_mapping = dict([(n.casefold(), n) for n in dir(self._entity)])

and then remapping the signal names if the computed signal name does not match:

    signame_folded = signame.casefold()
    if signame not in dir(self._entity) and signame_folded in signal_name_mapping:
        signame = signal_name_mapping[signame_folded]

alexforencich:
This PR implements case-insensitive bus signal name matching.  It adds a new parameter (case_insensitive, defaulting to True) to the `Bus` object to control this feature.  The `BusMonitor` object has also been updated to use `kwargs` for additional arguments to the `Bus` object to match `BusDriver`.  

The idea is simple: when matching signal names for buses, first try an exact match, if that fails then map everything to case-insensitive strings and try again.  Creating case-insensitive strings is done using `casefold()` which should also cover most unicode strings. 

Closes #2137.

### DIFF
--- a/cocotb/bus.py
+++ b/cocotb/bus.py
@@ -53,7 +53,7 @@ class Bus:
         Support for ``struct``/``record`` ports where signals are member names.
     """
 
-    def __init__(self, entity, name, signals, optional_signals=[], bus_separator="_", array_idx=None):
+    def __init__(self, entity, name, signals, optional_signals=[], bus_separator="_", case_insensitive=True, array_idx=None):
         """
         Args:
             entity (SimHandle): :any:`SimHandle` instance to the entity containing the bus.
@@ -70,17 +70,29 @@ class Bus:
                 See the *signals* argument above for details.
             bus_separator (str, optional): Character(s) to use as separator between bus
                 name and signal name. Defaults to '_'.
+            case_insensitive (bool, optional): Perform case-insensitive match on signal names.
+                Defaults to True.
             array_idx (int or None, optional): Optional index when signal is an array.
         """
         self._entity = entity
         self._name = name
         self._signals = {}
 
+        if case_insensitive:
+            signal_names = dir(self._entity)
+            signal_name_mapping = dict([(n.casefold(), n) for n in signal_names])
+
         for attr_name, sig_name in _build_sig_attr_dict(signals).items():
             if name:
                 signame = name + bus_separator + sig_name
             else:
                 signame = sig_name
+
+            # case insensitive remap
+            if case_insensitive:
+                signame_folded = signame.casefold()
+                if signame not in signal_names and signame_folded in signal_name_mapping:
+                    signame = signal_name_mapping[signame_folded]
 
             if array_idx is not None:
                 signame += "[{:d}]".format(array_idx)
@@ -92,6 +104,12 @@ class Bus:
                 signame = name + bus_separator + sig_name
             else:
                 signame = sig_name
+
+            # case insensitive remap
+            if case_insensitive:
+                signame_folded = signame.casefold()
+                if signame not in signal_names and signame_folded in signal_name_mapping:
+                    signame = signal_name_mapping[signame_folded]
 
             if array_idx is not None:
                 signame += "[{:d}]".format(array_idx)
--- a/cocotb/monitors/__init__.py
+++ b/cocotb/monitors/__init__.py
@@ -172,14 +172,13 @@ class BusMonitor(Monitor):
     _optional_signals = []
 
     def __init__(self, entity, name, clock, reset=None, reset_n=None,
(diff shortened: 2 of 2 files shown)


corna:
I think we should add a note in the comment that, if the base language is already case-insensitive (i.e., VHDL), then `case_insensitive=False` has no effect (or skip this logic altogether if a case-insensitive language is detected, as this should be done for us by the simulator).

ktbarrett:
I added the proper phrase to close the associated issue to your comment.

cocotb/bus.py
```diff
@@ -53,7 +53,7 @@ class Bus:
         Support for ``struct``/``record`` ports where signals are member names.
     """
 
-    def __init__(self, entity, name, signals, optional_signals=[], bus_separator="_", array_idx=None):
+    def __init__(self, entity, name, signals, optional_signals=[], bus_separator="_", case_insensitive=True, array_idx=None):
```
ktbarrett, line 56:
```suggestion
    def __init__(self, entity, name, signals, optional_signals=[], bus_separator="_", case_insensitive=False, array_idx=None):
```

`False` is the current default, there is no need to be breaking interfaces over this. It could matter in case-sensitive languages.

cmarqu:
> I think we should add a note in the comment that, if the base language is already case-insensitive (i.e., VHDL), then `case_insensitive=False` has no effect (or skip this logic altogether if a case-insensitive language is detected, as this should be done for us by the simulator).

In a mixed-language simulation, this would have to be done based on the language a certain signal is in, not on ``TOPLEVEL_LANG``, because at least a monitor might be connected to a bus lower in the hierarchy.
I'm not sure if we can get to that information easily right now.

cocotb/bus.py
```diff
@@ -70,18 +70,30 @@ def __init__(self, entity, name, signals, optional_signals=[], bus_separator="_"
                 See the *signals* argument above for details.
             bus_separator (str, optional): Character(s) to use as separator between bus
                 name and signal name. Defaults to '_'.
+            case_insensitive (bool, optional): Perform case-insensitive match on signal names.
+                Defaults to True.
```
cmarqu, line 74:
```suggestion
                Defaults to False.
```
...once Kaleb's review is in.

cmarqu:
This should also have a test.

cocotb/bus.py
cmarqu: This should also get a ``.. versionadded:: 1.5``marker (https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded).

ktbarrett:
@alexforencich Following on Colin's suggestions, see the [codecov report](https://codecov.io/gh/cocotb/cocotb/pull/2138/diff) on what's untested.

cocotb/bus.py
alexforencich: The logic does a case-sensitive check first, if that fails it falls back on case-insensitive.  So it will not break any existing tests even if case_insensitive=True by default.

alexforencich:
In a simulation with a case-insensitive language, it should probably be forced to `True` even if the user specifies `False`.  The simulator will not deal with this for us - it simply provides a list of signal names, which `Bus` then checks against each target signal.

cocotb/bus.py
ktbarrett: Currently if it fails case sensitive check, it would stop searching and warn instead of trying a case insensitive search (where it might pass without a warning in a corner case). That is a silent change in behavior. I have seen people lose an extraordinary amount of time to silent changes in behavior in corner cases, don't do it.

If this is the behavior you want I would make it tristate: explicit `True`, explicit `False`, and default. Default can be equivalent to `True` if that's what you think is best, but it must be explicit. I suggest a warning like "Found something with a case insensitive search that I otherwise wouldn't have. If you don't like that behavior, set to `False`."

ktbarrett:
The `Bus`, `Driver`, `Monitor`, and `Scoreboard` sources are being moved to [cocotb-bus](github.com/cocotb/cocotb-bus). Please reopen this PR there if you wish. An easy way to do that is:

```bash
git checkout my-pr-branch
git fetch cocotb-bus
git merge cocotb-bus/master
```

PR #2142, opened 2020-10-20
Added latency example with brief note in documentation.

mkhazraee:
This is an example for how to implement Verilog style latency in Python, discussed in Gitter chat with Colin Marquardt.

### DIFF
--- a/documentation/source/examples.rst
+++ b/documentation/source/examples.rst
@@ -34,6 +34,15 @@ to start and stop generation of input data.
 
 A :class:`.TestFactory` is used to generate the random tests.
 
+Latency
+=======
+
+The directory :file:`cocotb/examples/simple_latency/` contains a simple testbench to demonstrate how to implementat latency in python similar to Verilog, e.g. `assign #(delay) output_sig = input_sig;`.
+To do so 3 outputs from :term:`RTL` design are looped back with different delay values.
+This example uses 4 flip-flops based on the D flip-flop model in :term:`RTL`, one to register the original signal from testbench, and 3 to register the looped back signals.
+Origianl signal value is saved for the 2 previous cycles to verify the registerd signal values. Also waveform is generated for better visualization.
+
+This example does not use any :class:`.Driver`, :class:`.Monitor`, or :class:`.Scoreboard`.
 
 .. _matrix_multiplier:
 
--- a/examples/simple_latency/.gitignore
+++ b/examples/simple_latency/.gitignore
@@ -1,0 +1,46 @@
+# Python
+__pycache__
+
+# Waveforms
+*.vcd
+*.fst
+
+# Results
+results.xml
+sim_build
+
+# VCS files
+*.tab
+ucli.key
+
+# Cadence Incisive/Xcelium
+*.elog
+irun.log
+xrun.log
+irun.key
+xrun.key
+irun.history
+xrun.history
+INCA_libs
+xcelium.d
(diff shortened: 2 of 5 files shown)


cmarqu:
@mkhazraee Thanks for the effort! Though to be honest, I thought we would be talking about using https://docs.python.org/3/library/asyncio-queue.html or https://docs.python.org/3/library/queue.html in https://gitter.im/cocotb/Lobby?at=5f81204202e81701b010fed7 ?

mkhazraee:
> @mkhazraee Thanks for the effort! Though to be honest, I thought we would be talking about using https://docs.python.org/3/library/asyncio-queue.html or https://docs.python.org/3/library/queue.html in https://gitter.im/cocotb/Lobby?at=5f81204202e81701b010fed7 ?

@cmarqu Actually in the next message I said I'm not using it anymore and then you said examples.rst is good, so I thought you are interested in the delay example: https://gitter.im/cocotb/Lobby?at=5f867a68a605930a78aa6365

I can send the implementation using asyncio-queue, but it's both overkill and restricting. Previously I was looping back with delay on a source synchronous channel, so I could write and read the asyncio-queue on every clock (write on posedge, read on negedge after delay). But after putting more thought to it and finding the Edge trigger, it made more sense to make it a simple fork and delay, without requiring a clock. It was also more accurate delay value, instead of waiting until negedge. 

Still I'm not sure if the examples.rst is a good location for this, as it's more of a python side feature. I was also thinking of implementing an async fifo in python with asyncio-queue, but not sure how useful that is and also it needs more thinking to avoid python scheduler error due to empty fifo and so.

cmarqu:
Oh, I see... I apologize for the miscommunication, what was in my mind was not what I wrote.
I also did not expect that you would go all the way and set up a runnable example instead of something like a heading, two sentences and a code snippet.

I'm not sure what a simple-but-not-too-simple example showing queues would be.
A packet router is a often used to show methodologies like UVM, but if we want to do that, it would make sense to use it for more, like showcasing different possible implementations. (cocotb-coverage has one in https://github.com/mciepluc/cocotb-coverage/tree/master/examples/pkt_switch)

Maybe let's leave this open for a while and see if ideas trickle in?

mkhazraee:
No worries. I was actually thinking more of a code snippet too, but since all of the examples had a directory thought that's what should be done. I guess I should have asked too before making the example directory. 

If it's ok to put the snippet in examples.rst without any directory, that works actually better and becomes less complicated: I just mention the Verilog line and how to do it in python. Let me know if you want that and I make another pull request. 

Agreed about there should be a better example for async queue, that packet switch is really not benefiting much from it, or this delay example. Maybe in my own project I will run into such case, specially doing analog parts, and would make a simplified example.

cmarqu:
> If it's ok to put the snippet in examples.rst without any directory, that works actually better and becomes less complicated: I just mention the Verilog line and how to do it in python. Let me know if you want that and I make another pull request.

I had started https://github.com/cocotb/cocotb/wiki/Code-Examples a longer while ago with the intention of collecting enough examples to eventually transform into a proper "How to" section in Sphinx; this would be another place to put it.

> Agreed about there should be a better example for async queue, that packet switch is really not benefiting much from it, or this delay example. Maybe in my own project I will run into such case, specially doing analog parts, and would make a simplified example.

Analog parts, that is interesting, I'm doing this mostly in my day job, though unfortunately not with cocotb.

Issue #813, opened 2019-02-12
Unable to find constant in mixed lang simulation with FLI


leftink:
With a verilog top, the FLI interface fails to find a vhdl CONSTANT.  Looking at this it has to do with the fact that the interface uses the "fully qualified" name of the parent to create the name to search for.  FLI expects the root to start with "/", but VPI naming convention doesn't.

Funny enough, the FLI interface has no problem finding signals, it is just constants and variables that have this problem.

Looking at the approach across all three interfaces for get_handle_by_name, we seem to be going about it in a less than optimal way.  Providing a full path each time seems like the simulator will have to locate each element in the path before finding the final element we want.  This seems like a waste given that we have the parent handle and should be able to iterate over it or possibly restrict the search by it.

ktbarrett:
Likewise, I think this may be resolved, but if not, open a new issue.

PR #2143, opened 2020-10-20
GPI: iterate over parent if VPI gen scope array not found

garmin-mjames:
Follow-up to #2079. This brings the VPI code in-line with VHPI/FLI.

The underlying problem in #497 (as re-opened):
Icarus does not support `vpiGenScopeArray`, only `vpiGenScope`. So when trying to access a pseudo-region by name, `vpi_handle_by_name` lookup fails.

For both FLI and VHPI, the GPI code will fall-back to iterating over the parent handle to check if a generate statement can be found whose name starts with the pseudo-region name, and creates the pseudo-region if so. This matches that behavior when using VPI.

For example,
if searching for `dut.genblk` by name in Icarus, no handle will be found. However, `dut.genblk[0]` is discovered by iteration so the
pseudo-region can be inferred.

### DIFF
--- a/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -26,6 +26,7 @@
 ******************************************************************************/
 
 #include "VpiImpl.h"
+#include <cstring>
 #include <cocotb_utils.h>  // COCOTB_UNUSED
 
 extern "C" {
@@ -296,13 +297,52 @@ GpiObjHdl* VpiImpl::native_check_create(void *raw_hdl, GpiObjHdl *parent)
 GpiObjHdl* VpiImpl::native_check_create(std::string &name, GpiObjHdl *parent)
 {
     vpiHandle new_hdl;
+    const vpiHandle parent_hdl = parent->get_handle<vpiHandle>();
     std::string fq_name = parent->get_fullname() + "." + name;
-    std::vector<char> writable(fq_name.begin(), fq_name.end());
-    writable.push_back('\0');
 
-    new_hdl = vpi_handle_by_name(&writable[0], NULL);
+    new_hdl = vpi_handle_by_name(const_cast<char*>(fq_name.c_str()), NULL);
 
-    /* No need to iterate to look for generate loops as the tools will at least find vpiGenScopeArray */
+#ifdef ICARUS
+    /* Icarus does not support vpiGenScopeArray, only vpiGenScope.
+     * If handle is not found by name, look for a generate block with
+     * a matching prefix.
+     *     For Example:
+     *         genvar idx;
+     *         generate
+     *             for (idx = 0; idx < 5; idx = idx + 1) begin
+     *                 ...
+     *             end
+     *         endgenerate
+     *
+     *     genblk1      => vpiGenScopeArray (not found)
+     *     genblk1[0]   => vpiGenScope
+     *     ...
+     *     genblk1[4]   => vpiGenScope
+     *
+     *     genblk1 is not found directly, but if genblk1[n] is found,
+     *     genblk1 must exist, so create the pseudo-region object for it.
+     */
+    if (new_hdl == NULL) {
+        vpiHandle iter = vpi_iterate(vpiInternalScope, parent_hdl);
+        if (iter == NULL) {
+            goto skip_iterate;
+        }
+
+        for (auto rgn = vpi_scan(iter); rgn != NULL; rgn = vpi_scan(iter)) {
+            if (vpi_get(vpiType, rgn) == vpiGenScope) {
+                auto rgn_name = vpi_get_str(vpiName, rgn);
+                /* Check if name is a prefix of rgn_name */
+                if (rgn_name && name.length() > 0 && std::strncmp(name.c_str(), rgn_name, name.length()) == 0) {
+                    new_hdl = parent_hdl;
+                    vpi_free_object(iter);
+                    break;
+                }
+            }
+        }
+    }
+skip_iterate:
+#endif
+
     if (new_hdl == NULL) {
         LOG_DEBUG("Unable to query vpi_get_handle_by_name %s", fq_name.c_str());
         return NULL;
@@ -319,7 +359,7 @@ GpiObjHdl* VpiImpl::native_check_create(std::string &name, GpiObjHdl *parent)
     if (vpi_get(vpiType, new_hdl) == vpiGenScopeArray) {
         vpi_free_object(new_hdl);
 
-        new_hdl = parent->get_handle<vpiHandle>();
+        new_hdl = parent_hdl;
     }
 
 
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -257,10 +257,6 @@ class HierarchyObject(RegionObject):
         except KeyError:
             pass
 
-        if not self._discovered:
-            self._discover_all()
-            return self.__get_sub_handle_by_name(name)
-
         # Cache to avoid a call to the simulator if we already know the name is
         # invalid. Unclear if we care, but we had this before.
         if name in self._invalid_sub_handles:
--- a/documentation/source/newsfragments/2079.bugfix.rst
+++ b/documentation/source/newsfragments/2079.bugfix.rst
@@ -1,4 +1,4 @@
-Generate blocks are now accessible directly via lookup without having to iterate over parent handle. (:pr:`2079`)
+In :ref:`Icarus Verilog <sim-icarus>`, generate blocks are now accessible directly via lookup without having to iterate over parent handle. (:pr:`2079`, :pr:`2143`)
 
   .. code-block:: python3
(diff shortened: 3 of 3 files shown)


cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -296,16 +296,36 @@ GpiObjHdl* VpiImpl::native_check_create(void *raw_hdl, GpiObjHdl *parent)
 GpiObjHdl* VpiImpl::native_check_create(std::string &name, GpiObjHdl *parent)
 {
     vpiHandle new_hdl;
+    vpiHandle parent_hdl = parent->get_handle<vpiHandle>();
```
ktbarrett, line 299:
```suggestion
    const vpiHandle parent_hdl = parent->get_handle<vpiHandle>();
```

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -296,16 +296,36 @@ GpiObjHdl* VpiImpl::native_check_create(void *raw_hdl, GpiObjHdl *parent)
...
     std::string fq_name = parent->get_fullname() + "." + name;
     std::vector<char> writable(fq_name.begin(), fq_name.end());
     writable.push_back('\0');
 
     new_hdl = vpi_handle_by_name(&writable[0], NULL);
 
-    /* No need to iterate to look for generate loops as the tools will at least find vpiGenScopeArray */
+    /* If not found, check to see if the name of a generate loop.
```
ktbarrett, line 306:
```suggestion
    /* If not found, check to see if the name is a generate loop.
```

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -296,16 +296,36 @@ GpiObjHdl* VpiImpl::native_check_create(void *raw_hdl, GpiObjHdl *parent)
...
+                if (vpi_get(vpiType, rgn) == vpiGenScope) {
+                    std::string rgn_name = vpi_get_str(vpiName, rgn);
+                    if (rgn_name.compare(0,name.length(),name) == 0) {
+                        new_hdl = parent_hdl;
+                        vpi_free_object(iter);
+                        break;
+                    }
+                }
+            }
+        }
```
ktbarrett, line 324: I might refactor this slightly doing a `iter == NULL` check after the call to `vpi_iterate`, logging failure and returning. Then you can de-indent this loop 1 level.

ktbarrett (commented):
The solution is definitely better than what was there before. Do we have a test for this we can prove fails without it?

One possible change is: since only Icarus does not support `vpiGenScopeArray`, could we ifdef this section out if `ICARUS` isn't defined?

garmin-mjames:
**test_discovery.pseudo_region_access** should fail using icarus without this or #2079.

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -296,13 +296,46 @@ GpiObjHdl* VpiImpl::native_check_create(void *raw_hdl, GpiObjHdl *parent)
 GpiObjHdl* VpiImpl::native_check_create(std::string &name, GpiObjHdl *parent)
 {
     vpiHandle new_hdl;
+    const vpiHandle parent_hdl = parent->get_handle<vpiHandle>();
     std::string fq_name = parent->get_fullname() + "." + name;
-    std::vector<char> writable(fq_name.begin(), fq_name.end());
-    writable.push_back('\0');
 
-    new_hdl = vpi_handle_by_name(&writable[0], NULL);
+    new_hdl = vpi_handle_by_name(const_cast<char*>(fq_name.c_str()), NULL);
```
ktbarrett, line 302: One thing I've always wondered is why are we looking up with the full name form the root scope and not passing the parent object as the second argument and just using the element name? Maybe it has something to do with mixed language support?

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -296,13 +296,46 @@ GpiObjHdl* VpiImpl::native_check_create(void *raw_hdl, GpiObjHdl *parent)
...
+        vpiHandle iter = vpi_iterate(vpiInternalScope, parent_hdl);
+        if (iter == NULL) {
+            goto skip_iterate;
+        }
+
+        vpiHandle rgn;
+        for (rgn = vpi_scan(iter); rgn != NULL; rgn = vpi_scan(iter)) {
+            if (vpi_get(vpiType, rgn) == vpiGenScope) {
+                std::string rgn_name = vpi_get_str(vpiName, rgn);
+                if (rgn_name.compare(0, name.length(), name) == 0) {
```
ktbarrett, line 328: Just use `rgn_name == name`. I'd suggest you leave `rgn_name` as a `const char*` using `auto` to avoid an allocation.

```suggestion
                auto rgn_name = vpi_get_str(vpiName, rgn);
                if (rgn_name == name) {
```

cocotb/share/lib/vpi/VpiImpl.cpp
ktbarrett: Also, this should be safe since the `std::string` is non-const. Good catch.

garmin-mjames: See #813. The full name might be worse for mixed language.

Also with C++11, we get contiguous chars, which may be the original reason for using a vector.

ktbarrett: Yeah, that is why I brought it up. I'm not certain how iterating with `vpi_iterate` over a handle retrieved with the FLI would work... that's my concern. I don't think the VPI, VHPI, or FLI were designed with mixed language cosimulation in mind, so I'm sure support is all over the place.

cocotb/share/lib/vpi/VpiImpl.cpp
garmin-mjames: Will that work?
Example:
`name` is "genblk"
`rgn_name` is "genblk[0]"

We want to check that `rgn_name` *starts with* `name`. `compare()` without *pos* and *len* parameters would return >0 I think.

ktbarrett: Ah, that wasn't immediately apparent to me that's what this code was doing (maybe warrants a comment or a `startswith` function?). You can still avoid the allocation though https://en.cppreference.com/w/cpp/string/basic_string/compare overload 5.

garmin-mjames: Overload 5 `name.compare(0, std::string::npos, rgn_name)` fails as it compares to the end of the longer null-terminated `rgn_name`.
Resorted to `std::strncmp` unless you have another idea.

cocotb/share/lib/vpi/VpiImpl.cpp
ktbarrett: Perhaps we can check how the handle was retrieved? If it is the same interface as we are currently using, we can use parent. Otherwise, we build the full name (and cache it) and look up with the other registered implementations from the root scope?

cocotb/share/lib/vpi/VpiImpl.cpp
ktbarrett: My bad overload 6 would work.

cocotb/share/lib/vpi/VpiImpl.cpp
garmin-mjames: We need more thorough tests for mixed language.
Looks like before e276c34c2e65ef9545430b14f1d86573a1853aea they used the parent handle for scope.

If the main issue is path format (`:` for VHPI, `/` for FLI, `.` for VPI), we could add a path conversion function between `GpiImplInterface`s for full name lookup?

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -296,13 +297,52 @@ GpiObjHdl* VpiImpl::native_check_create(void *raw_hdl, GpiObjHdl *parent)
 GpiObjHdl* VpiImpl::native_check_create(std::string &name, GpiObjHdl *parent)
 {
     vpiHandle new_hdl;
+    const vpiHandle parent_hdl = parent->get_handle<vpiHandle>();
```
garmin-mjames, line 300: Should we protect against casting a non-VPI parent handle to `vpiHandle`?
I'm not sure how a `vpiGenScopeArray` could have a non-VPI parent.

ktbarrett: I don't think we need to worry about that in this PR. It's systemic.

cocotb/share/lib/vpi/VpiImpl.cpp
ktbarrett: It might. I'm not too familiar with the issues in supporting mixed language sims, but it seems to work with Questa right now.

ktbarrett (approved):

PR #2148, opened 2020-10-20
Improve VHPI logging output and some comments

cmarqu:

### DIFF
--- a/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
+++ b/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
@@ -39,7 +39,7 @@ extern "C" void handle_vhpi_callback(const vhpiCbDataT *cb_data);
 
 VhpiArrayObjHdl::~VhpiArrayObjHdl()
 {
-    LOG_DEBUG("Releasing VhpiArrayObjHdl handle at %p", (void *)get_handle<vhpiHandleT>());
+    LOG_DEBUG("VHPI: Releasing VhpiArrayObjHdl handle at %p", (void *)get_handle<vhpiHandleT>());
     if (vhpi_release_handle(get_handle<vhpiHandleT>()))
         check_vhpi_error();
 }
@@ -48,7 +48,7 @@ VhpiObjHdl::~VhpiObjHdl()
 {
     /* Don't release handles for pseudo-regions, as they borrow the handle of the containing region */
     if (m_type != GPI_GENARRAY) {
-        LOG_DEBUG("Releasing VhpiObjHdl handle at %p", (void *)get_handle<vhpiHandleT>());
+        LOG_DEBUG("VHPI: Releasing VhpiObjHdl handle at %p", (void *)get_handle<vhpiHandleT>());
         if (vhpi_release_handle(get_handle<vhpiHandleT>()))
             check_vhpi_error();
     }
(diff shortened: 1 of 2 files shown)


ktbarrett (commented):
I'm starting to wonder if we should be prefixing everything, or be using a different logger, say `cocotb.gpi.vhpi` that is _under_ `cocotb.gpi`.

cmarqu:
A different logger would certainly be cleaner, but OTOH, this is a low-tech way that works and is well readable in both source code and output, so maybe not worth additional brain cycles.

ktbarrett (approved):
There is nothing objectional here.

cmarqu:
I'm actually getting compile errors on Mac OS and with clang that I need to investigate... or just remove that code.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -438,13 +438,16 @@ int VhpiCbHdl::arm_callback()
...
             LOG_ERROR("VHPI: Unable to register a callback handle for VHPI type %s(%d)",
-                         m_impl->reason_to_string(cb_data.reason), cb_data.reason);
+                      m_impl->reason_to_string(cb_data.reason), cb_data.reason);
             goto error;
         }
 
         cbState = (vhpiStateT)vhpi_get(vhpiStateP, new_hdl);
-        if (vhpiEnable != cbState) {
-            LOG_ERROR("VHPI: Registered callback isn't enabled! Got %d", cbState);
+        if (vhpiUndefined == cbState) {
```
ktbarrett, line 446: I think you need to check this before you cast the result of the call to `vhpi_get` to a `vhpiStateP`. `vhpiUndefined` is a valid result of `vhpi_get`, but it isn't in the `vhpiStateP` enum, hence the warning.

cmarqu: Ah, thanks for the explanation.
The current code is not exactly nice, is there a way to improve it? Or is all that casting too much for too little gain?
What prompted this is that `vhpiUndefined` is simulator-dependent:
```
#ifndef IUS
#define vhpiUndefined -1
#else
#define vhpiUndefined 1000
#endif
```
But, oh,  I now see that I am casting `vhpiUndefined` wrong (to `unsigned`), but then, `typedef uint32_t vhpiIntT`... what a mess. I guess I'll just revert that part of the PR.

cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -438,13 +438,17 @@ int VhpiCbHdl::arm_callback()
...
+                      m_impl->reason_to_string(cb_data.reason), cb_data.reason);
             goto error;
         }
 
-        cbState = (vhpiStateT)vhpi_get(vhpiStateP, new_hdl);
-        if (vhpiEnable != cbState) {
-            LOG_ERROR("VHPI: Registered callback isn't enabled! Got %d", cbState);
+        // don't cast to vhpiStateT immediately because vhpiUndefined is not in the enum
+        vhpiIntT cbState_raw = vhpi_get(vhpiStateP, new_hdl);
+        if ((unsigned int)vhpiUndefined == cbState_raw) {
```
ktbarrett, line 447:
```suggestion
        if ((vhpiIntT)vhpiUndefined == cbState_raw) {
```

ktbarrett (approved):

Issue #1731, opened 2020-04-29
Convert tests to use `async def` syntax


eric-wieser:
Currently, the tests use [the old `yield` syntax.](https://docs.cocotb.org/en/latest/coroutines.html#generator-based-coroutines)

We should change them to use the now-recommended [`async def` / `await` syntax](https://docs.cocotb.org/en/latest/coroutines.html#async-functions).

Split from #1289.

Some care should be taken to ensure that we still have _some_ tests for the `yield` syntax.
This makes this less suitable for new contributors than gh-1730.

It's also probably less important to have these converted before the next release, so feel free to unmilestone this if we're cutting it tight on a release.

Issue #839, opened 2019-03-11
Result of a yield of a list of triggers is not always in the list


jwprice100:
This issue occurs when dealing with a series of forking coroutines which read signals driven by the design under test and cocotb. Under the right conditions the result returned by yielding a list of triggers is not in the list of triggers.  Another related issue is that under these same conditions the erroneous trigger will fire over and over. Eventually simulation resumes but not for a significant time. 

**Context:**
CentOS based system running Python 3.6
Using VHDL2008
Simulator: Riviera Pro 2018.10
Added -fgnu89-inline switch to gcc_args in order for CoCoTb to compile on my system

Please note I am new to CoCoTb and some of what I first believed to be bugs were in fact user error. This may be as well but I have not been able to determine why. I've attached a simple example that reproduces the issue.

[simple_coco_testcase.zip](https://github.com/potentialventures/cocotb/files/2950395/simple_coco_testcase.zip)

jwprice100:
Digging more into this issue, I believe I know what one of the issues I am seeing is. 

Consider the case of two coroutines A and B. A has a single trigger (say RisingEgde of clock) that it yields two indefintely. Coroutine B also a list of triggers it yields to, then at a later point yields to a different list. That second list includes the trigger that coroutine A uses. 

When a trigger fires, the scheduler considers all the other triggers a scheduled coroutine is associated with that did not fire. For a particular pending trigger it checks if the set of coroutines that will be scheduled is a superset of the routines associated with the trigger. If so, it unprimes the trigger and removes it. This makes sense as if all coroutines that are waiting on that pending trigger awoke due a different trigger they are not relevant.

However consider the case where a pending trigger which did not fire and is attached to a different coroutine that is not being scheduled to run as in my example.

The code which handles this is given below:
```python
          for pending in other_triggers:
                # every coroutine waiting on this trigger is already being woken
                if scheduling_set.issuperset(self._trigger2coros[pending]):
                    if pending.primed:
                        pending.unprime()
                    del self._trigger2coros[pending]
```

My hacky workaround is as follows:
```python
            for pending in other_triggers:
                # every coroutine waiting on this trigger is already being woken
                if scheduling_set.issuperset(self._trigger2coros[pending]):
                    if pending.primed:
                        pending.unprime()
                    del self._trigger2coros[pending]
                else:
                     #For each coroutine
                     for coro in scheduling_set:
                        #If that coroutine is associated with a pending trigger
                        if coro in self._trigger2coros[pending]:
[...]

PR #2153, opened 2020-10-22
Convert more tests to async/await syntax

cmarqu:
Refs #1731

### DIFF
--- a/tests/test_cases/test_cocotb/test_clock.py
+++ b/tests/test_cases/test_cocotb/test_clock.py
@@ -7,44 +7,36 @@ Tests relating to cocotb.clock.Clock
 import cocotb
 from cocotb.clock import Clock
 from cocotb.triggers import Timer, RisingEdge
-from cocotb.result import TestFailure
 from cocotb.utils import get_sim_time
 from math import isclose
 
 
-@cocotb.test(expect_fail=False)
-def test_clock_with_units(dut):
+@cocotb.test()
+async def test_clock_with_units(dut):
     clk_1mhz   = Clock(dut.clk, 1.0, units='us')
     clk_250mhz = Clock(dut.clk, 4.0, units='ns')
 
-    if str(clk_1mhz) != "Clock(1.0 MHz)":
-        raise TestFailure("{} != 'Clock(1.0 MHz)'".format(str(clk_1mhz)))
-    else:
-        dut._log.info('Created clock >{}<'.format(str(clk_1mhz)))
+    assert str(clk_1mhz) == "Clock(1.0 MHz)"
+    dut._log.info('Created clock >{}<'.format(str(clk_1mhz)))
 
-    if str(clk_250mhz) != "Clock(250.0 MHz)":
-        raise TestFailure("{} != 'Clock(250.0 MHz)'".format(str(clk_250mhz)))
-    else:
-        dut._log.info('Created clock >{}<'.format(str(clk_250mhz)))
+    assert str(clk_250mhz) == "Clock(250.0 MHz)"
+    dut._log.info('Created clock >{}<'.format(str(clk_250mhz)))
 
     clk_gen = cocotb.fork(clk_1mhz.start())
 
     start_time_ns = get_sim_time(units='ns')
 
-    yield Timer(1)
-
-    yield RisingEdge(dut.clk)
+    await Timer(1, "ns")
+    await RisingEdge(dut.clk)
 
     edge_time_ns = get_sim_time(units='ns')
-    if not isclose(edge_time_ns, start_time_ns + 1000.0):
-        raise TestFailure("Expected a period of 1 us")
+    assert isclose(edge_time_ns, start_time_ns + 1000.0), "Expected a period of 1 us"
 
     start_time_ns = edge_time_ns
 
-    yield RisingEdge(dut.clk)
+    await RisingEdge(dut.clk)
     edge_time_ns = get_sim_time(units='ns')
-    if not isclose(edge_time_ns, start_time_ns + 1000.0):
-        raise TestFailure("Expected a period of 1 us")
+    assert isclose(edge_time_ns, start_time_ns + 1000.0), "Expected a period of 1 us"
 
     clk_gen.kill()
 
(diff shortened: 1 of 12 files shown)


tests/test_cases/test_cocotb/test_concurrency_primitives.py
```diff
@@ -11,68 +11,63 @@
 
 
 @cocotb.test()
-def test_yield_list_stale(dut):
-    """ Test that a trigger yielded as part of a list can't cause a spurious wakeup """
+async def test_await_list_stale(dut):
+    """ Test that a trigger awaited as part of a list can't cause a spurious wakeup """
```
eric-wieser, line 15: Should probably adjust the test name now that there's no list here.

cmarqu: There is still the list of ``waiters`` which I'd assume the "list" referred to?
If not, what would the docstring now be? I also don't completely get the meaning of "stale" in the test name.

cmarqu: @garmin-mjames IIRC the docstring was from you initially, do you have a suggestion?

garmin-mjames: This test is by @eric-wieser from https://github.com/cocotb/cocotb/issues/839#issuecomment-474171169, it was just related to an issue I filed at the same time.

The test is about two coroutines waiting on a shared trigger, but one of the coroutines is waiting on a list/`First` that contains the shared trigger.
If a different trigger in the `First` fires, we want to test that the scheduler removes that coro from the other still-pending trigger(s) so that it doesn't get woken up again later. (So, `events[1].set()` shouldn't resume `wait_for_lists`)

All that said, maybe:
```suggestion
    async def test_unfired_first_triggers(dut):
    """ Test that un-fired trigger(s) in First don't later cause a spurious wakeup """
```

tests/test_cases/test_cocotb/test_concurrency_primitives.py
```diff
@@ -11,68 +11,64 @@
 
 
 @cocotb.test()
-def test_yield_list_stale(dut):
-    """ Test that a trigger yielded as part of a list can't cause a spurious wakeup """
+async def test_await_list_stale(dut):
+    async def test_unfired_first_triggers(dut):
```
garmin-mjames, line 15: Original comment didn't have the test definition line in the suggestion.
```suggestion
async def test_unfired_first_triggers(dut):
```

tests/test_cases/test_cocotb/test_clock.py
```diff
@@ -7,76 +7,64 @@
...
     clk_gen.kill()
 
 
-@cocotb.test(expect_fail=False)
-def test_anternal_clock(dut):
-    """Test ability to yield on an external non cocotb coroutine decorated
-    function"""
-    clk_gen = cocotb.fork(Clock(dut.clk, 100).start())
+@cocotb.test()
+async def test_anternal_clock(dut):
```
ktbarrett, line 63: "Anternal"? I think what was meant was "external".

ktbarrett (commented):
There are some things that can be updated from `TestComplete`s to `asserts`, but I'm guessing you are leaving that for a different PR?

tests/test_cases/test_cocotb/test_clock.py
cmarqu: Yes, I'm going to make this and the assert changes in a different PR.

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -92,36 +88,36 @@ def test_integer(dut):
...
     val = random.randint(-2**31, 2**31 - 1)
     log = logging.getLogger("cocotb.test")
-    yield Timer(1)
+    await Timer(1, "ns")
     log.info("Setting the value %i" % val)
     dut.stream_in_real <= val
-    yield Timer(1)
-    yield Timer(1)  # FIXME: Workaround for VHPI scheduling - needs investigation
+    await Timer(1, "ns")
+    await Timer(1, "ns")  # FIXME: Workaround for VHPI scheduling - needs investigation
```
eric-wieser, line 120: I feel like these two are deliberately intended to delay the absolute minimal time, and the comment here is about very short timers.

eric-wieser (approved):
I think we need to be slightly careful of changing `Timer(1)` to `Timer(1, 'ns'`), as often it's deliberate to test the smallest time step possible. Otherwise looks great!

tests/test_cases/test_cocotb/test_handle.py
cmarqu: Oh, yes, I'll go over them again.

cmarqu:
I wonder if we should have a predefined Timer for the smallest timestep to make this more explicit in code. Verilog has something like this, I just don't remember what it's called right now.

eric-wieser:
`Timer.shortest` or `Timer.one_timestep` would be an reasonable name, but if verilog has a better name matching it would be sensible.

cmarqu:
I slightly misremembered, it's the unit in SystemVerilog which can be ``step`` in addition to ``fs``, ``ps``, ...

tests/test_cases/test_cocotb/test_handle.py
cmarqu: I now also reverted specifying ``units`` in ``test_scheduler/test_kill_coroutine_waiting_on_the_same_trigger`` since it might have been deliberate there as well.

tests/test_cases/test_cocotb/test_concurrency_primitives.py
```diff
@@ -11,68 +11,63 @@
...
+    """ Test that un-fired trigger(s) in First don't later cause a spurious wakeup """
     # gh-843
     events = [Event() for i in range(3)]
 
     waiters = [e.wait() for e in events]
 
-    @cocotb.coroutine
-    def wait_for_lists():
-        ret_i = waiters.index((yield [waiters[0], waiters[1]]))
+    async def wait_for_lists():
```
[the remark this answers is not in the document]
cmarqu: Indeed, fixed, thanks.

tests/test_cases/test_cocotb/test_clock.py
cmarqu: This follow-up PR is #2170.

cmarqu:
> I slightly misremembered, it's the unit in SystemVerilog which can be `step` in addition to `fs`, `ps`, ...

This is now in #2172.

PR #2154, opened 2020-10-22
Remove superfluous coroutine decorators

cmarqu:


ktbarrett:
This is a breaking change to the interface of the drivers. I remember we decided to leave these in #2022.

cmarqu:
Oh, indeed.

PR #2162, opened 2020-10-26
Fix lexer name for doctest examples

cmarqu:
See https://pygments.org/docs/lexers/#pygments.lexers.python.PythonConsoleLexer


ktbarrett:
Looks the same to me?
https://cocotb--2162.org.readthedocs.build/en/2162/quickstart.html#reading-values-from-signals
https://docs.cocotb.org/en/stable/quickstart.html#reading-values-from-signals

cmarqu:
Yes, it renders the same, but it is still the semantically right thing to do.
(What triggered this change is https://kodare.net/2020/10/23/100_percent_coverage_docs.html where this change becomes relevant.)

ktbarrett:
Could you remove line 188 while you are here? It doesn't render the way I'd like it to.

cmarqu:
Your wish is my command.

ktbarrett (approved):

PR #2163, opened 2020-10-26
Use absolute path for VCS and compile as VHDL-93 for Xcelium

cmarqu:

### DIFF
--- a/tests/test_cases/test_fatal/Makefile
+++ b/tests/test_cases/test_fatal/Makefile
@@ -7,10 +7,15 @@ TOPLEVEL_LANG ?= verilog
 TOPLEVEL := fatal
 MODULE := test_fatal
 
+PWD=$(shell pwd)
+
 ifeq ($(TOPLEVEL_LANG),verilog)
-	VERILOG_SOURCES := fatal.sv
+    VERILOG_SOURCES := $(PWD)/fatal.sv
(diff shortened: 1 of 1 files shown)


tests/test_cases/test_fatal/Makefile
```diff
@@ -7,10 +7,15 @@ TOPLEVEL_LANG ?= verilog
...
+PWD=$(shell pwd)
+
 ifeq ($(TOPLEVEL_LANG),verilog)
-	VERILOG_SOURCES := fatal.sv
+    VERILOG_SOURCES := $(PWD)/fatal.sv
 else ifeq ($(TOPLEVEL_LANG),vhdl)
-	VHDL_SOURCES := fatal.vhd
+    VHDL_SOURCES := $(PWD)/fatal.vhd
+    ifneq ($(filter $(SIM),ius xcelium),)
+        SIM_ARGS += -v93
```
ktbarrett, line 17: Is there an issue with it running with a different version of VHDL?

cmarqu: IUS/Xcelium still defaults to VHDL-87 and thus errors out on ``end entity ...``.
(The IUS makefile just sets using VHDL-93 unconditionally, I removed that for Xcelium so that the user can also select newer standard versions.)

ktbarrett (approved):

PR #2165, opened 2020-10-26
Convert examples to use assert

cmarqu:

### DIFF
--- a/examples/axi_lite_slave/tests/test_axi_lite_slave.py
+++ b/examples/axi_lite_slave/tests/test_axi_lite_slave.py
@@ -1,7 +1,5 @@
 import os
 import cocotb
-from cocotb.result import TestFailure
-from cocotb.result import TestSuccess
 from cocotb.clock import Clock
 from cocotb.triggers import Timer
 from cocotb.drivers.amba import AXI4LiteMaster
@@ -44,11 +42,8 @@ async def write_address_0(dut):
     await Timer(CLK_PERIOD_NS * 10, units='ns')
 
     value = dut.dut.r_temp_0
-    if value != DATA:
-        # Fail
-        raise TestFailure("Register at address 0x%08X should have been: \
-                           0x%08X but was 0x%08X" % (ADDRESS, DATA, int(value)))
-
+    assert value == DATA, ("Register at address 0x%08X should have been "
+                           "0x%08X but was 0x%08X" % (ADDRESS, DATA, int(value)))
     dut.log.info("Write 0x%08X to address 0x%08X" % (int(value), ADDRESS))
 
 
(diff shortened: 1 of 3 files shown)


examples/axi_lite_slave/tests/test_axi_lite_slave.py
```diff
@@ -157,8 +148,8 @@ async def write_fail(dut):
         print("Exception: %s" % str(e))
         dut._log.info("Bus successfully raised an error")
         raise TestSuccess()
-    raise TestFailure("AXI bus should have raised an error when writing to \
-                        an invalid address")
+    raise TestFailure("AXI bus should have raised an error when writing to "
+                      "an invalid address")
```
ktbarrett, line 152: I would also remove the `raise TestSuccess()` line above. New idiomatic is `assert` for failures, falling off the end of the test for passes. I think it was stated as a stretch goal to remove the `TestFinish` exceptions, but I can't seem to find it? @eric-wieser?

```suggestion
    assert False, "AXI bus should have raised an error when writing to an invalid address"
```

examples/axi_lite_slave/tests/test_axi_lite_slave.py
```diff
@@ -190,5 +181,5 @@ async def read_fail(dut):
         print("Exception: %s" % str(e))
         dut._log.info("Bus Successfully Raised an Error")
         raise TestSuccess()
-    raise TestFailure("AXI bus should have raised an error when reading from \
-                        an invalid address")
+    raise TestFailure("AXI bus should have raised an error when reading from "
+                      "an invalid address")
```
ktbarrett, line 185: Same here.

examples/axi_lite_slave/tests/test_axi_lite_slave.py
cmarqu: The suggestion is fine, but currently we still need the ``raise TestSuccess()`` line to prevent this line from being reached..

ktbarrett: You could use an `else` clause. The entire thing could be refactored to use `assert_raises` from `test_cocotb`. If you wish to use `assert`s, I would do it all the way, even if it requires some minor refactoring.

I'm wondering if we can use pytest for that? We could start depending on pytest for test regressions. That is far less egregious than a runtime dependency. But you can ignore this off-tangent rambling for this PR.

cmarqu: Indeed, having the ``assert_raises`` available generally would be useful; something for a different PR.

ktbarrett (approved):

PR #2170, opened 2020-11-02
Convert test_cocotb to use assert

cmarqu:

### DIFF
--- a/tests/test_cases/test_cocotb/common.py
+++ b/tests/test_cases/test_cocotb/common.py
@@ -9,7 +9,6 @@ import traceback
 from contextlib import contextmanager
 
 import cocotb
-from cocotb.result import TestFailure
 from cocotb.triggers import Timer
 
 
@@ -31,17 +30,14 @@ def _check_traceback(running_coro, exc_type, pattern):
     except exc_type:
         tb_text = traceback.format_exc()
     else:
-        raise TestFailure("Exception was not raised")
+        assert False, "Exception was not raised"
 
-    if not re.match(pattern, tb_text):
-        raise TestFailure(
-            (
-                "Traceback didn't match - got:\n\n"
-                "{}\n"
-                "which did not match the pattern:\n\n"
-                "{}"
-            ).format(tb_text, pattern)
-        )
+    assert re.match(pattern, tb_text), (
+        "Traceback didn't match - got:\n\n"
+        "{}\n"
+        "which did not match the pattern:\n\n"
+        "{}"
+    ).format(tb_text, pattern)
 
 
 @contextmanager
@@ -54,4 +50,4 @@ def assert_raises(exc_type, pattern=None):
                 "Correct exception type caught, but message did not match pattern"
         pass
     else:
-        raise AssertionError("{} was not raised".format(exc_type.__name__))
+        assert False, "{} was not raised".format(exc_type.__name__)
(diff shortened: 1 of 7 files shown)


tests/test_cases/test_cocotb/test_edge_triggers.py
```diff
@@ -30,16 +29,14 @@ def do_single_edge_check(dut, level):
     """Do test for rising edge"""
     old_value = dut.clk.value.integer
     dut._log.info("Value of %s is %d" % (dut.clk._path, old_value))
-    if old_value is level:
-        raise TestError("%s not to %d start with" % (dut.clk._path, not level))
+    assert old_value != level, "%s not to %d start with" % (dut.clk._path, not level)
```
cmarqu, line 32: The ``is`` here was wrong IMO.

tests/test_cases/test_cocotb/test_edge_triggers.py
```diff
@@ -77,33 +72,27 @@ def test_either_edge(dut):
     yield Timer(1)
     dut.clk <= 1
     yield Edge(dut.clk)
-    if dut.clk.value.integer != 1:
-        raise TestError("Value should be 0")
+    assert dut.clk.value.integer == 1
```
cmarqu, line 75: Half the messages in this test were wrong, and with #2028 useless anyway.

tests/test_cases/test_cocotb/test_edge_triggers.py
```diff
@@ -30,16 +29,14 @@ def do_single_edge_check(dut, level):
...
+    assert old_value != level, "%s not to %d start with" % (dut.clk._path, not level)
     if level == 1:
         yield RisingEdge(dut.clk)
     else:
         yield FallingEdge(dut.clk)
     new_value = dut.clk.value.integer
     dut._log.info("Value of %s is %d" % (dut.clk._path, new_value))
-    if new_value is not level:
-        raise TestError("%s not %d at end" % (dut.clk._path, level))
+    assert new_value == level, "%s not %d at end" % (dut.clk._path, level)
```
cmarqu, line 39: The differentiation between ``TestError`` and ``TestFailure`` is now lost but was it ever useful here?

tests/test_cases/test_cocotb/common.py
```diff
@@ -31,17 +30,15 @@ def _check_traceback(running_coro, exc_type, pattern):
...
-                "which did not match the pattern:\n\n"
-                "{}"
-            ).format(tb_text, pattern)
-        )
+        assert False, (
+            "Traceback didn't match - got:\n\n"
+            "{}\n"
+            "which did not match the pattern:\n\n"
+            "{}"
+        ).format(tb_text, pattern)
```
ktbarrett, line 41: You could put the `re.match` in the `assert` so we can see what was matched when the assertion rewriting comes in?

tests/test_cases/test_cocotb/test_edge_triggers.py
ktbarrett: Here? Nope. Elsewhere? Yes and no, the test could be set to `expect_error`, but the user should never be throwing `TestError` so that's a bad justification. I think `TestError` can be deprecated at this point? It was only ever used for internal test handling, but that's all been refactored out IIRC.

ktbarrett (approved):

tests/test_cases/test_cocotb/common.py
cmarqu: Oh, yes, indeed.

tests/test_cases/test_cocotb/test_edge_triggers.py
cmarqu: I opened a new issue about the deprecation: #2174

PR #2172, opened 2020-11-02
Add units='step'

### LINKED ISSUE #2169: Method for "wait minimum time"
Following the discussion in https://github.com/cocotb/cocotb/pull/2153#pullrequestreview-521152671, we should have a method like ``Timer.shortest()`` to more semantically express that a minimum waiting time is wanted.

### LINKED ISSUE #2171: Time unit "step" should represent the simulator time step
Right now, ``Timer`` and ``utils.get_sim_steps`` use ``units=None`` (the default value) to mean "use the simulator time step".
Add ``units='step'`` (as exists in SystemVerilog) to mean the same.

cmarqu:
Use Timer.shortest() in cocotb_test.

Closes #2169, closes #2171

### DIFF
--- a/cocotb/clock.py
+++ b/cocotb/clock.py
@@ -28,6 +28,7 @@
 """A clock class."""
 
 import itertools
+import warnings
 
 from cocotb.log import SimLog
 from cocotb.triggers import Timer
@@ -66,10 +67,9 @@ class Clock(BaseClock):
         period (int): The clock period. Must convert to an even number of
             timesteps.
         units (str, optional): One of
-            ``None``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
-            When no *units* is given (``None``) the timestep is determined by
-            the simulator.
-
+            ``'step'``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
+            When *units* is ``'step'``,
+            the timestep is determined by the simulator (see :make:var:`COCOTB_HDL_TIMEPRECISION`).
 
     If you need more features like a phase shift and an asymmetric duty cycle,
     it is simple to create your own clock generator (that you then :func:`fork`):
@@ -106,10 +106,21 @@ class Clock(BaseClock):
         await Timer(1000, units="ns")
         high_delay = low_delay = 10  # change the clock speed
         await Timer(1000, units="ns")
+
+    .. versionchanged:: 1.5
+        Support ``'step'`` as the the *units* argument to mean "simulator time step".
+
+    .. deprecation:: 1.5
+        Using None as the the *units* argument is deprecated, use ``'step'`` instead.
     """
 
-    def __init__(self, signal, period, units=None):
+    def __init__(self, signal, period, units="step"):
         BaseClock.__init__(self, signal)
+        if units is None:
+            warnings.warn(
+                'Using units=None is deprecated, use units="step" instead.',
+                DeprecationWarning, stacklevel=2)
+            units="step"  # don't propagate deprecated value
         self.period = get_sim_steps(period, units)
         self.half_period = get_sim_steps(period / 2.0, units)
         self.frequency = 1.0 / get_time_from_sim_steps(self.period, units='us')
--- a/cocotb/utils.py
+++ b/cocotb/utils.py
@@ -102,24 +102,32 @@ def get_time_from_sim_steps(steps, units):
     return _ldexp10(steps, _get_simulator_precision() - _get_log_time_scale(units))
 
 
-def get_sim_steps(time, units=None):
+def get_sim_steps(time, units="step"):
     """Calculates the number of simulation time steps for a given amount of *time*.
 
     Args:
         time (numbers.Real or decimal.Decimal):  The value to convert to simulation time steps.
-        units (str or None, optional):  String specifying the units of the result
-            (one of ``None``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``).
-            ``None`` means time is already in simulation time steps.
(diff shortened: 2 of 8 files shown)


cmarqu:
TBD: add test in ``test_cocotb/test_timing_triggers.test_timer_with_units``.

cmarqu:
TBD: Check if docs need to be updated.

ktbarrett (commented):
I'm not sure if I like having a 'shortest' shortcut, it's not really offering anything over just writing `Timer(1, 'step')`.

cmarqu:
Ok, I could just make a reference to such a Timer with a speaking name in the tests without offering it as part of cocotb.

cmarqu:
Tbd: add versionchanged tag

alexforencich:
Maybe there is an argument for having `Timer.ns(n)`, `Timer.step(1)`, etc.

ktbarrett:
@alexforencichI agree I don't like the string arguments. I had started #1130 for making a physical type for simulation time that worked like some chrono libraries.

cmarqu:
TBDs fixed, doc best found via https://cocotb--2172.org.readthedocs.build/en/2172/release_notes.html#features

cocotb/utils.py
```diff
@@ -118,7 +118,7 @@ def get_sim_steps(time, units=None):
         :exc:`ValueError`: If given *time* cannot be represented by simulator precision.
     """
     result = time
-    if units is not None:
+    if units is not None and units != "step":
```
ktbarrett, line 121: This is the typical idiom. 
```suggestion
    if units not in (None, 'step'):
```
One could also do a default check to correct `None` to `'step'` and only do the singular check  of `'step'` thereafter.

cocotb/clock.py
```diff
@@ -66,10 +66,9 @@ class Clock(BaseClock):
         period (int): The clock period. Must convert to an even number of
             timesteps.
         units (str, optional): One of
-            ``None``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
-            When no *units* is given (``None``) the timestep is determined by
-            the simulator.
-
+            ``None``, ``'step'``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
+            When no *units* is given (``None``) or is ``'step'``,
```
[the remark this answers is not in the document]
cmarqu: Yep, thanks.

cocotb/clock.py
```diff
@@ -66,10 +66,9 @@ class Clock(BaseClock):
...
+            the timestep is determined by the simulator.
```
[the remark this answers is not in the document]
cmarqu: I did the linking. I don't feel it's bad that we don't explain details here, the user just needs to pick a unit that makes sense to them. "Not sure what your simulator's time step is? Don't rely on that then."

garmin-mjames:
Newsfragment should be 2172, not 2171.

cmarqu:
@garmin-mjames Towncrier works fine with either issue or PR, and in this case, the first item in the PR is misleading because obsolete (and editing it now feels wrong).
We have linked to issues in the past, they actually explain the problem being solved better most of the time (but there are PRs without accompanying issues, so we link to the PR then).

cocotb/utils.py
```diff
@@ -118,7 +118,7 @@ def get_sim_steps(time, units=None):
         :exc:`ValueError`: If given *time* cannot be represented by simulator precision.
     """
     result = time
-    if units is not None:
+    if units not in (None, "step"):
```
eric-wieser, line 121: Is there any point supporting `None` here? It seems odd to introduce two ways to spell the same thing, unless our plan is to deprecate one of them.

I'd be tempted to change the default to `units="step"`.

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -39,35 +39,39 @@ async def test_function_reentrant_clock(dut):
...
 
+    time_fs = get_sim_time(units='fs')
+    await Timer(1, units='step')
+    assert time_step == get_sim_time(units='fs') - time_fs, "Expected units='step' == units=None"
+
     try:
         # Await for 2.5 timesteps, should throw exception
         await Timer(2.5*time_step, units='fs')
-        raise TestFailure("Timers should throw exception if time cannot be achieved with simulator resolution")
+        assert False, "Timers should throw exception if time cannot be achieved with simulator resolution"
```
eric-wieser, line 53: If you're changing this, may as well use `assert_raises`

cocotb/utils.py
ktbarrett: This change will need to be reflected in the docstring and newsfrag, as a reminder.

cmarqu: Changed the default and deprecated None, with tests. Docstrings and newsfragments should be fine now too.

tests/test_cases/test_cocotb/test_timing_triggers.py
cmarqu: Done.

documentation/source/newsfragments/2171.feature.rst
```diff
@@ -0,0 +1,6 @@
+The ``units`` argument to :class:`cocotb.triggers.Timer`,
+:class:`cocotb.clock.Clock` and :func:`cocotb.utils.get_sim_steps`,
+and the ``timeout_unit`` argument to
+:func:`cocotb.triggers.with_timeout` and :class:`cocotb.test`
+now accepts ``'step'`` to mean the simulator time step.
+This used to be expressed using ``None`` which is now deprecated.
```
ktbarrett, line 6:
```suggestion
This used to be expressed using ``None``, which is now deprecated.
```

ktbarrett (approved):

PR #2175, opened 2020-11-04
Add support for adding sets of parameters with TestFactory.add_option

alexforencich:
Add support for adding sets of parameters with TestFactory.add_option using tuples, similar to pytest.mark.parametrize.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -627,30 +627,38 @@ class TestFactory:
     This Factory allows us to generate sets of tests based on the different
     permutations of the possible arguments to the test function.
 
-    For example if we have a module that takes backpressure and idles and
-    have some packet generation routines ``gen_a`` and ``gen_b``:
+    For example, if we have a module that takes backpressure, has two configurable
+    features where enabling ``feature_b`` requires ``feature_a`` to be active, and
+    need to test against data generation routines ``gen_a`` and ``gen_b``:
 
     >>> tf = TestFactory(test_function=run_test)
     >>> tf.add_option(name='data_in', optionlist=[gen_a, gen_b])
     >>> tf.add_option('backpressure', [None, random_backpressure])
-    >>> tf.add_option('idles', [None, random_idles])
+    >>> tf.add_option(('feature_a', 'feature_b'), [(False, False), (True, False), (True, True)])
     >>> tf.generate_tests()
 
     We would get the following tests:
 
-        * ``gen_a`` with no backpressure and no idles
-        * ``gen_a`` with no backpressure and ``random_idles``
-        * ``gen_a`` with ``random_backpressure`` and no idles
-        * ``gen_a`` with ``random_backpressure`` and ``random_idles``
-        * ``gen_b`` with no backpressure and no idles
-        * ``gen_b`` with no backpressure and ``random_idles``
-        * ``gen_b`` with ``random_backpressure`` and no idles
-        * ``gen_b`` with ``random_backpressure`` and ``random_idles``
+        * ``gen_a`` with no backpressure and both features disabled
+        * ``gen_a`` with no backpressure and only ``feature_a`` enabled
+        * ``gen_a`` with no backpressure and both features enabled
+        * ``gen_a`` with ``random_backpressure`` and both features disabled
+        * ``gen_a`` with ``random_backpressure`` and only ``feature_a`` enabled
+        * ``gen_a`` with ``random_backpressure`` and both features enabled
+        * ``gen_b`` with no backpressure and both features disabled
+        * ``gen_b`` with no backpressure and only ``feature_a`` enabled
+        * ``gen_b`` with no backpressure and both features enabled
+        * ``gen_b`` with ``random_backpressure`` and both features disabled
+        * ``gen_b`` with ``random_backpressure`` and only ``feature_a`` enabled
+        * ``gen_b`` with ``random_backpressure`` and both features enabled
 
     The tests are appended to the calling module for auto-discovery.
 
     Tests are simply named ``test_function_N``. The docstring for the test (hence
     the test description) includes the name and description of each generator.
+
+    .. versionchanged:: 1.5
+        Groups of options are now supported
     """
 
     # Prevent warnings from collection of TestFactories by unit testing frameworks.
@@ -675,11 +683,22 @@ class TestFactory:
         """Add a named option to the test.
 
         Args:
-            name (str): Name of the option. Passed to test as a keyword
-                argument.
+            name (str or iterable of str): An option name, or an iterable of
+                several option names.  Passed to test as keyword arguments.
 
             optionlist (list): A list of possible options for this test knob.
+                If N names were specified, this must be a list of N-tuples or
+                lists, where each element specifies a value for its respective
+                option.
+
+        .. versionchanged:: 1.5
+            Groups of options are now supported
         """
+        if not isinstance(name, str):
+            name = tuple(name)
+            for opt in optionlist:
+                if len(name) != len(opt):
+                    raise ValueError("Mismatch between number of options and number of option values in group")
         self.kwargs[name] = optionlist
 
     def generate_tests(self, prefix="", postfix=""):
@@ -714,7 +733,18 @@ class TestFactory:
             name = "%s%s%s_%03d" % (prefix, self.name, postfix, index + 1)
             doc = "Automatically generated test\n\n"
 
+            # preprocess testoptions to split tuples
+            testoptions_split = {}
             for optname, optvalue in testoptions.items():
+                if isinstance(optname, str):
+                    testoptions_split[optname] = optvalue
+                else:
+                    # previously checked in add_option; ensure nothing has changed
+                    assert len(optname) == len(optvalue)
+                    for n, v in zip(optname, optvalue):
+                        testoptions_split[n] = v
+
+            for optname, optvalue in testoptions_split.items():
                 if callable(optvalue):
                     if not optvalue.__doc__:
                         desc = "No docstring supplied"
@@ -729,7 +759,7 @@ class TestFactory:
                            (name, mod.__name__))
             kwargs = {}
             kwargs.update(self.kwargs_constant)
-            kwargs.update(testoptions)
+            kwargs.update(testoptions_split)
             if hasattr(mod, name):
                 self.log.error("Overwriting %s in module %s. "
                                "This causes a previously defined testcase "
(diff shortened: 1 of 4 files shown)


ktbarrett (commented):
You need to make a test and create a newsfragment.

alexforencich:
Where should I add tests?  I don't see any TestFactory tests outside of issue 857.

ktbarrett:
@alexforencich Probably a new module under `test_cocotb`.

alexforencich:
Okay, and you want tests for just my additions or for all of TestFactory?

ktbarrett:
If you are willing to do all of `TestFactory` so we can get 100% coverage that would be fantastic!

alexforencich:
I guess I can write one for TestFactory; there isn't much in there and testing my additions cover most of the code path anyway.  The newsfragment will just be `.. versionchanged:: 1.5` at the bottom of the `add_option` docstring, followed by a line or two about what changed, right?

ktbarrett:
Yes the `.. versionchange:: 1.5` line is necessary, but you also need to add a small blurb to a file named `2175.feature.rst` to `documentation/source/newsfragments/` that states the new feature in a sentence or two. It might be verbatim what you put in the `TestFactory` docstring, but it's for generating the changelog.

cocotb/regression.py
```diff
@@ -714,7 +723,19 @@ def generate_tests(self, prefix="", postfix=""):
             name = "%s%s%s_%03d" % (prefix, self.name, postfix, index + 1)
             doc = "Automatically generated test\n\n"
 
+            # preprocess testoptions to split tuples
+            testoptions_split = {}
             for optname, optvalue in testoptions.items():
+                if ',' in optname:
+                    optname = optname.split(',')
+                if isinstance(optname, tuple) or isinstance(optname, list):
+                    assert len(optname) == len(optvalue)
```
ktbarrett, line 732: This should be an `if` and `raise`, this isn't an appropriate use of `assert`. Also, this check should be moved to `add_option`. Catch errors as early as you can.

cocotb/regression.py
```diff
@@ -677,8 +677,17 @@ def add_option(self, name, optionlist):
...
+                tuple denoting option names. Passed to test as keyword
+                arguments.
 
             optionlist (list): A list of possible options for this test knob.
+                If N names were specified, this must be a list of N-tuples or
+                lists, where each element specifies a value for its respective
+                option.
+
+        .. versionchanged:: 1.5
+            Support groups of options
```
ktbarrett, line 690: I feel like this could be expanded to be a little clearer, but I'm not sure how to word it.

tests/test_cases/test_cocotb/test_testfactory.py
```diff
@@ -0,0 +1,21 @@
...
+from cocotb.regression import TestFactory
+
+
+async def run_testfactory_test(dut, arg1, arg2, arg3):
+    assert (arg1, arg2, arg3) in [
+        ("a1v1", "a2v1", "a4v1"),
+        ("a1v2", "a2v1", "a4v1"),
+        ("a1v1", "a2v2", "a4v2"),
+        ("a1v2", "a2v2", "a4v2"),
+    ]
```
ktbarrett, line 16: This proves that only the expected values are generated, but doesn't prove all the expected combinations are generated.

For a test I would have this "factoried" test here save the arguments it got into a list. Then add a test definition *after* the call to `factory.generate_test()` check the list to see if all the expected values were generated.

cocotb/regression.py
```diff
@@ -677,8 +677,17 @@ def add_option(self, name, optionlist):
         Args:
             name (str): Name of the option. Passed to test as a keyword
                 argument.
+            name (str, list, or tuple): A comma-separated string, list, or
+                tuple denoting option names. Passed to test as keyword
+                arguments.
```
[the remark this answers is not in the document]
alexforencich: Whoops, I meant to replace the original one.  I will update this.

cocotb/regression.py
```diff
@@ -714,7 +723,19 @@ def generate_tests(self, prefix="", postfix=""):
             name = "%s%s%s_%03d" % (prefix, self.name, postfix, index + 1)
             doc = "Automatically generated test\n\n"
 
+            # preprocess testoptions to split tuples
+            testoptions_split = {}
             for optname, optvalue in testoptions.items():
+                if ',' in optname:
+                    optname = optname.split(',')
```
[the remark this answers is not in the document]
alexforencich: `pytest.mark.parametrize` supports this so I added it for consistency, but maybe they support it for legacy reasons.  I will remove it.

cocotb/regression.py
alexforencich: OK, what should I raise?

cocotb/regression.py
```diff
@@ -714,7 +723,19 @@ def generate_tests(self, prefix="", postfix=""):
...
+                if isinstance(optname, tuple) or isinstance(optname, list):
+                    assert len(optname) == len(optvalue)
+                    for n, v in zip(optname, optvalue):
+                        testoptions_split[n] = v
+                else:
+                    testoptions_split[optname] = optvalue
```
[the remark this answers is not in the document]
alexforencich: Why?  If you remove that, then calling add_option with a string as a name will fail, and all existing code that uses add_option will have to be updated to wrap the name into a list or tuple.

tests/test_cases/test_cocotb/test_testfactory.py
alexforencich: Okay, is there an example of something like this somewhere?

cocotb/regression.py
```diff
@@ -714,7 +723,19 @@ def generate_tests(self, prefix="", postfix=""):
             name = "%s%s%s_%03d" % (prefix, self.name, postfix, index + 1)
             doc = "Automatically generated test\n\n"
 
+            # preprocess testoptions to split tuples
+            testoptions_split = {}
             for optname, optvalue in testoptions.items():
+                if ',' in optname:
+                    optname = optname.split(',')
+                if isinstance(optname, tuple) or isinstance(optname, list):
```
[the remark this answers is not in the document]
alexforencich: Good point, I switched this to check against string, and in `add_option` convert the name to a tuple (also has to be hashable, tuple is but list is not) and added a loop there to check that all of the option values are the correct length.

cocotb/regression.py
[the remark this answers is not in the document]
ktbarrett: Sorry that was supposed to be deleted.

cocotb/regression.py
ktbarrett: `ValueError` is kind of a catch all for bad argument values.

tests/test_cases/test_cocotb/test_testfactory.py
ktbarrett: Not that I know of. You are trailblazing :) But it's basically a module-level set that you add the arguments into and a regular cocotb test after the testfactory code that reads the set and ensures all the combos are there.

```python
args = set()

async def test_testfactory_run(dut, arg1, arg2, arg3):
    args.add((arg1, arg2, arg3))

# ... test factory stuff ...

async def test_testfactory_generates(dut):
   assert args == set(
        ("a1v1", "a2v1", "a4v1"),
        ("a1v2", "a2v1", "a4v1"),
        ("a1v1", "a2v2", "a4v2"),
        ("a1v2", "a2v2", "a4v2"),
    )

Comment (alexforencich) [reply tests/test_cases/test_cocotb/test_testfactory.py]: OK, that seems to work.  I'm a bit new to writing unit tests using a testing framework in this way, so writing tests to test that the test framework is generating tests correctly is a bit of a test (yo dawg...).

**Comment by alexforencich** [comment]
@ktbarrett what else needs to be done here?

File: cocotb/regression.py [hunk 8]
```diff
@@ -675,11 +675,22 @@ def add_option(self, name, optionlist):
...
+            name (str or iterable): An option name, or an iterable of several option
+                names.  Passed to test as keyword arguments.
 
             optionlist (list): A list of possible options for this test knob.
+                If N names were specified, this must be a list of N-tuples or
+                lists, where each element specifies a value for its respective
+                option.
+
+        .. versionchanged:: 1.5
+            Support groups of options
```
eric-wieser, line 687:
```suggestion
            Groups of options are now supported
```

tests/test_cases/test_cocotb/test_testfactory.py
```diff
@@ -0,0 +1,30 @@
...
+factory.generate_tests()
+
+
+@cocotb.test()
+async def test_testfactory_verify_args(dut):
+    assert testfactory_test_args == set([
+        ("a1v1", "a2v1", "a4v1"),
+        ("a1v2", "a2v1", "a4v1"),
+        ("a1v1", "a2v2", "a4v2"),
+        ("a1v2", "a2v2", "a4v2"),
```
eric-wieser, line 29:
```suggestion
factory.add_option(("arg2", "arg3"), [("a2v1", "a3v1"), ("a2v2", "a3v2")])
factory.generate_tests()


@cocotb.test()
async def test_testfactory_verify_args(dut):
    assert testfactory_test_args == set([
        ("a1v1", "a2v1", "a3v1"),
        ("a1v2", "a2v1", "a3v1"),
        ("a1v1", "a2v2", "a3v2"),
        ("a1v2", "a2v2", "a3v2"),
```

cocotb/regression.py
eric-wieser: Can you update the docstring at the top of this class to show how to use the new feature?

cocotb/regression.py
```diff
@@ -675,11 +675,22 @@ def add_option(self, name, optionlist):
         """Add a named option to the test.
 
         Args:
-            name (str): Name of the option. Passed to test as a keyword
-                argument.
+            name (str or iterable): An option name, or an iterable of several option
```
eric-wieser, line 678:
```suggestion
            name (str or iterable of str): An option name, or an iterable of several option
```

cocotb/regression.py
eric-wieser: So, as I understand it, this `assert` is trying to verify a precondition we already ensure in `add_option`? If so, a comment to that effect would be good.

alexforencich: I added the check in `add_option` that raises the exception, but kept the assert to ensure no funny business takes place between calling `add_option` and `generate_tests`.  But sure, adding a comment is a good idea.

cocotb/regression.py
alexforencich: Yes, I just updated the PR with that change

alexforencich:
What else needs to be done here?

ktbarrett (approved):

PR #2176, opened 2020-11-05
GHA: fix lint-whitespace

garmin-mjames:
Fix **lint-whitespace** command.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -20,7 +20,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: whitespace
       run: |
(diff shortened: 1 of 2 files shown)


garmin-mjames:
Wow that's ugly when the tests are skipped.

ktbarrett:
We aren't paying for compute credits; the developer can clearly see the result of each individual job, unlike with Travis; and this gives the developer less information. Is there some benefit I'm not seeing?

garmin-mjames:
@ktbarrett I was mostly thinking about fail sooner for feedback, but I get your point. It's annoying to have the substance of the PR stuck behind making the linter happy. I'll change it to just the whitespace fix.

eric-wieser (approved):

PR #2177, opened 2020-11-05
Cleanup TestError

### LINKED ISSUE #2174: Deprecate TestError/expect_error
With the move to using assertions instead of raising exceptions, ``TestError``/``expect_error`` could/should be deprecated. https://github.com/cocotb/cocotb/pull/2170#discussion_r516712751

ktbarrett:
`TestError` was unnecessary here. It was used to fail the test if a signal could not be mapped to a Python SimHandle type. This could be easily done with a standard exception. This cleans up the last usage of `TestError` in core code and will allow for it's deprecation.

Closes #2174.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -217,9 +217,10 @@ class RegionObject(SimHandleBase):
         self._log.debug("Discovering all on %s", self._name)
         for thing in self._handle.iterate(simulator.OBJECTS):
             name = thing.get_name_string()
+            path = self._child_path(name)
             try:
-                hdl = SimHandle(thing, self._child_path(name))
-            except TypeError as e:
+                hdl = SimHandle(thing, path)
+            except NotImplementedError as e:
                 self._log.debug("%s", e)
                 continue
 
@@ -918,7 +919,7 @@ def SimHandle(handle, path=None):
         The `SimHandle` object.
 
     Raises:
-        TypeError: If no matching object for GPI type could be found.
+        NotImplementedError: If no matching object for GPI type could be found.
     """
     _type2cls = {
         simulator.MODULE:      HierarchyObject,
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -36,7 +36,6 @@ import cocotb
 from cocotb import simulator
 from cocotb.binary import BinaryValue
 from cocotb.log import SimLog
-from cocotb.result import TestError
 
 # Only issue a warning for each deprecated attribute access
 _deprecation_warned = set()
@@ -220,7 +219,7 @@ class RegionObject(SimHandleBase):
             name = thing.get_name_string()
             try:
                 hdl = SimHandle(thing, self._child_path(name))
-            except TestError as e:
+            except TypeError as e:
                 self._log.debug("%s", e)
                 continue
 
@@ -919,7 +918,7 @@ def SimHandle(handle, path=None):
         The `SimHandle` object.
 
     Raises:
-        TestError: If no matching object for GPI type could be found.
+        TypeError: If no matching object for GPI type could be found.
     """
     _type2cls = {
         simulator.MODULE:      HierarchyObject,
@@ -956,7 +955,7 @@ def SimHandle(handle, path=None):
         return obj
 
     if t not in _type2cls:
-        raise TestError("Couldn't find a matching object for GPI type %d (path=%s)" % (t, path))
+        raise TypeError("Couldn't find a matching object for GPI type %d (path=%s)" % (t, path))
     obj = _type2cls[t](handle, path)
     _handle2obj[handle] = obj
     return obj
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -955,7 +955,7 @@ def SimHandle(handle, path=None):
         return obj
 
     if t not in _type2cls:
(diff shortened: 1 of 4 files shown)


cocotb/result.py
```diff
@@ -118,8 +116,19 @@ def __init__(self, exception):
...
+    .. deprecated:: 1.5
+        Raise a standard Python exception instead of calling this function.
+        A stacktrace will be printed by cocotb automatically if the exception is unhandled.
+    """
+
+    def __init__(self, *args, **kwargs):
+            warnings.warn(
+                "TestError is deprecated - raise a standard Exception instead",
+                DeprecationWarning, stacklevel=2)
+            super().__init__(*args, **kwargs)
```
ktbarrett, line 131:
```suggestion
        warnings.warn(
            "TestError is deprecated - raise a standard Exception instead",
            DeprecationWarning, stacklevel=2)
        super().__init__(*args, **kwargs)
````

File: cocotb/result.py [hunk 2]
```diff
@@ -118,8 +116,19 @@ def __init__(self, exception):
 
 
 class TestError(TestComplete):
-    """Exception showing that the test was completed with severity Error."""
-    pass
+    """
+    Exception showing that the test was completed with severity Error.
+
```
ktbarrett, line 121:
```suggestion

```

cocotb/handle.py
```diff
@@ -956,7 +955,7 @@ def SimHandle(handle, path=None):
         return obj
 
     if t not in _type2cls:
-        raise TestError("Couldn't find a matching object for GPI type %d (path=%s)" % (t, path))
+        raise TypeError("Couldn't find a matching object for GPI type %d (path=%s)" % (t, path))
```
eric-wieser, line 958: I would go for `RuntimeError` (this is a problem with cocotb) or `NotImplementedError` (please make a PR for this!).

TypeError is usually for "this argument is the wrong type", not "a type could not be found".

ktbarrett:
```suggestion
        raise NotImplementedError("Couldn't find a matching object for GPI type %d (path=%s)" % (t, path))
```

cocotb/handle.py
```diff
@@ -919,7 +918,7 @@ def SimHandle(handle, path=None):
         The `SimHandle` object.
 
     Raises:
-        TestError: If no matching object for GPI type could be found.
+        TypeError: If no matching object for GPI type could be found.
```
eric-wieser, line 921:
```suggestion
        NotImplementedError: If no matching object for GPI type could be found.
```

ktbarrett:
Instead of creating 2 warnings I could remove the warning from `create_error` and `raise_error` and let only the one in `TestError` be reported.

cocotb/handle.py
```diff
@@ -220,7 +219,7 @@ def _discover_all(self):
             name = thing.get_name_string()
             try:
                 hdl = SimHandle(thing, self._child_path(name))
-            except TestError as e:
+            except TypeError as e:
```
eric-wieser, line 222:
```suggestion
            except NotImplementedError as e:
```

cocotb/handle.py
```diff
@@ -220,7 +219,7 @@ def _discover_all(self):
             name = thing.get_name_string()
             try:
                 hdl = SimHandle(thing, self._child_path(name))
```
eric-wieser, line 221: This would be clearer as:
```suggestion
            path = self._child_path(name)
            try:
                hdl = SimHandle(thing, path)
```
Otherwise there's a question of which bit is failing

ktbarrett: right, will change

cocotb/result.py
```diff
@@ -118,8 +116,18 @@ def __init__(self, exception):
 
 
 class TestError(TestComplete):
-    """Exception showing that the test was completed with severity Error."""
-    pass
+    """
+    Exception showing that the test was completed with severity Error.
+    .. deprecated:: 1.5
```
eric-wieser, line 121:
```suggestion

    .. deprecated:: 1.5
```

cocotb/handle.py
```diff
@@ -956,7 +956,7 @@ def SimHandle(handle, path=None):
         return obj
 
     if t not in _type2cls:
-        raise TestError("Couldn't find a matching object for GPI type %d (path=%s)" % (t, path))
+        raise NotImplementedError("Couldn't find a matching object for GPI type %d (path=%s)" % (t, path))
```
garmin-mjames, line 959:
```suggestion
        raise NotImplementedError("Couldn't find a matching object for GPI type %s(%d) (path=%s)" % (handle.get_type_string(), t, path))
```
For debugging it would be nice to have the handle address too but I don't see an easy way to grab that.

cocotb/result.py
```diff
@@ -118,8 +116,19 @@ def __init__(self, exception):
 
 
 class TestError(TestComplete):
-    """Exception showing that the test was completed with severity Error."""
-    pass
+    """
+    Exception showing that the test was completed with severity Error.
+
+    .. deprecated:: 1.5
+        Raise a standard Python exception instead of calling this function.
```
garmin-mjames, line 123:
```suggestion
        Raise a standard Python exception instead.
```

cmarqu:
Should have a ``2177.removal.rst`` newsfragment and a "Closes #2174." in the commit message.

cmarqu (approved):

PR #2179, opened 2020-11-05
Add more mentions of pip's --user option

cmarqu:

### DIFF
--- a/documentation/source/install.rst
+++ b/documentation/source/install.rst
@@ -107,6 +107,12 @@ Installation of cocotb
 
           pip install cocotb
 
+.. note::
+
+    If your user does not have permissions to install cocotb using the instructions above,
+    try adding the ``--user`` option to ``pip``
+    (see :ref:`the pip documentation<https://pip.pypa.io/en/stable/user_guide/#user-installs>`).
+
 .. warning::
 
     ``pip`` may belong to a different Python installation to what you expect.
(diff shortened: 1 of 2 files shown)


documentation/source/install_devel.rst
```diff
@@ -18,7 +18,15 @@ The instructions for installing the development version of cocotb vary depending
 
           pip install --global-option build_ext --global-option --compiler=mingw32 https://github.com/cocotb/cocotb/archive/master.zip
 
-      See also the `pip User Guide <https://pip.pypa.io/en/stable/user_guide/#user-installs/>`_.
+      This requires administrator permissions.
```
cmarqu, line 21: I'm not sure why we didn't have this part all the time. Did I just forget it back then, or did/do we not know whether user installs work on Windows?
@themperek, do you remember?

imphil: Because it's not universal pip behavior. It could be that pip defaults to local installations, it could be that it falls back to local installations if the global path is not writable, it could be that it fails. Really depends on too many factors to give general guidance. (Documenting the installation of Python packages for inexperienced users is more like a huge FSM than a simple line.)

ktbarrett: I think a single line should suffice if it conveys that it may or may not work. 

`If your user does not have permissions to install cocotb via 'pip', try 'pip install --user'` with a link to the pip documentation. It *might* prevent some user from coming to us about what is clearly not our problem, and might prevent some novice users from installing with sudo.

cmarqu: Cleaned this up. Please take another look.

ktbarrett (approved):
I think it looks good now.

documentation/source/install.rst
```diff
@@ -107,6 +107,10 @@ Installation of cocotb
 
           pip install cocotb
 
+.. note:: If your user does not have permissions to install cocotb using the instructions above,
+          try adding the ``--user`` option to ``pip``
+          (see :ref:`the pip documentation<https://pip.pypa.io/en/stable/user_guide/#user-installs>`).
```
eric-wieser, line 112: rST nit:

```suggestion
.. note::

    If your user does not have permissions to install cocotb using the instructions above,
    try adding the ``--user`` option to ``pip``
    (see :ref:`the pip documentation<https://pip.pypa.io/en/stable/user_guide/#user-installs>`).
```

(has no effect on the rendering)

cmarqu: Done.

PR #2180, opened 2020-11-07
Add pytest-github-actions-annotate-failures plugin to GHA

cmarqu:
See https://pypi.org/project/pytest-github-actions-annotate-failures/


cmarqu:
I don't see any effect. Maybe it's not worth spending time debugging this, at least not until we run all of our tests under pytest control.

PR #2181, opened 2020-11-08
Add config for pylint

cmarqu:
Add some cocotb-specific settings to the generated template,
and ignore some of the less-useful messages.

This is *not* meant as a config with which we should eventually strive
to have no pylint messages left or use in CI, just as a starting point
for improving the code now and then.


ktbarrett (approved):
We can bikeshed later

PR #2182, opened 2020-11-08
Clean up some pylint findings

cmarqu:
See also #2181.

### DIFF
--- a/cocotb/drivers/__init__.py
+++ b/cocotb/drivers/__init__.py
@@ -51,6 +51,7 @@ class BitDriver:
         self._signal = signal
         self._clk = clk
         self._generator = generator
+        self._cr = None
 
     def start(self, generator: Iterable[Tuple[int, int]] = None) -> None:
         """Start generating data.
@@ -310,6 +311,9 @@ class ValidatedBusDriver(BusDriver):
         valid_generator: Iterable[Tuple[int, int]] = None, **kwargs: Any
     ) -> None:
         BusDriver.__init__(self, entity, name, clock, **kwargs)
+        self.on = None
+        self.off = None
+        # keep this line after the on/off attributes since it overwrites them
         self.set_valid_generator(valid_generator=valid_generator)
 
     def _next_valids(self):
@@ -364,7 +368,7 @@ async def polled_socket_attachment(driver, sock):
             else:
                 driver.log.error(repr(e))
                 raise
-        if not len(data):
+        if not len(data) > 0:
             driver.log.info("Remote end closed the connection")
             break
         driver.append(data)
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -42,23 +42,6 @@ import inspect
 from typing import Any, Union
 from collections.abc import Coroutine
 
-# Debug mode controlled by environment variables
-if "COCOTB_ENABLE_PROFILING" in os.environ:
-    import cProfile
-    import pstats
-    _profile = cProfile.Profile()
-    _profiling = True
-else:
-    _profiling = False
-
-# Sadly the Python standard logging module is very slow so it's better not to
-# make any calls by testing a boolean flag first
-if "COCOTB_SCHEDULER_DEBUG" in os.environ:
-    _debug = True
-else:
-    _debug = False
-
-
 import cocotb
 import cocotb.decorators
 from cocotb.decorators import RunningTask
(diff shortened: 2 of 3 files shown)


cocotb/drivers/__init__.py
```diff
@@ -319,24 +323,24 @@ def _next_valids(self):
         on followed by the number of cycles to be off.
         The ``on`` cycles should be non-zero, we skip invalid generator entries.
         """
-        self.on = False
+        on = False
 
         if self.valid_generator is not None:
-            while not self.on:
```
eric-wieser, line 325: I feel like the point of these `on` an `off` properties is so the caller can inspect the current state of the object

cocotb/drivers/__init__.py
```diff
@@ -29,6 +29,9 @@
 
 """Set of common driver base classes."""
 
+import socket
+import errno
+
```
eric-wieser, line 34: I don't agree with this change - there's no point increasing everyone's import time for a socket driver that probably no one uses.

cocotb/drivers/__init__.py
cmarqu: Indeed, e.g. the Avalon driver makes use of it.

cocotb/drivers/__init__.py
cmarqu: Rolled back.

cocotb/drivers/__init__.py
```diff
@@ -311,6 +312,8 @@ def __init__(
     ) -> None:
         BusDriver.__init__(self, entity, name, clock, **kwargs)
         self.set_valid_generator(valid_generator=valid_generator)
+        self.on = None
+        self.off = None
```
eric-wieser, line 316:
```suggestion
        self.on = None
        self.off = None
        self.set_valid_generator(valid_generator=valid_generator)
```
since `self.set_valid_generator` overwrites `on`

cmarqu: Will do this, with a comment.

cocotb/scheduler.py
```diff
@@ -459,7 +455,6 @@ def _event_loop(self, trigger):
                 # to try and avoid them being destroyed at a weird time (as
                 # happened in gh-957)
                 del trigger
-                del coro
```
ktbarrett, line 462: @eric-wieser You introduced this, is this critical?

eric-wieser: I think it's a bad idea to delete this line. What's the pylint error we're addressing here?

ktbarrett: `cocotb/scheduler.py:462:20: W0631: Using possibly undefined loop variable 'coro' (undefined-loop-variable)`

cmarqu: Is there a way for coro to exist at this point? If yes, I will have learned something.

ktbarrett: I'm confused as to why `del` is used here. It's not guaranteed to call `__del__` immediately or ever. There is probably a better way to do it.

ktbarrett: @cmarqu loop variables still exist after loops finish.

ktbarrett: I would just mark that `noqa` for now.

cmarqu: Not sure if noqa is even worth it, as pylint reports so much else...

ktbarrett: I've noticed... One thing that would help is configuring pylint to not report comment warnings and "not enough/too many methods/variables/attributes" warnings which I think are total bunk.

eric-wieser: Moving the `del` into the loop would fix the warning, and behave practically identically

eric-wieser: The warning is telling us that `del` could raise NameError if the loop never runs, which is true.

eric-wieser: > It's not guaranteed to call `__del__` immediately or ever

True, but in practice it _is_ predictable in CPython, as long as we avoid reference cycles.

cmarqu: Moved the ``del coro`` inside the loop.

ktbarrett (approved):

cocotb/scheduler.py
cmarqu: > One thing that would help is configuring pylint to not report comment warnings and "not enough/too many methods/variables/attributes" warnings which I think are total bunk.

The "too-many" warnings are actually already ignored in #2181. (You may have to use the ``--rcfile`` option to make pylint use it.)

cocotb/scheduler.py
```diff
@@ -67,7 +50,19 @@
 from cocotb.log import SimLog
 from cocotb.result import TestComplete
 from cocotb.utils import remove_traceback_frames
-from cocotb import _py_compat
+from cocotb import outcomes,_py_compat
```
garmin-mjames, line 53: nit:
```suggestion
from cocotb import outcomes, _py_compat
```

cmarqu: Sharp eyes :)

ktbarrett: I'm interested that flake8 didn't pick that up

garmin-mjames: @ktbarrett
https://github.com/cocotb/cocotb/blob/43adc60479c193d9dc700c3851c7db2907034e35/setup.cfg#L17

PR #2183, opened 2020-11-08
Fix tox on Windows

ktbarrett:
`tox-conda` recently fixed an issue (tox-dev/tox-conda#42) with environments on Windows and we should now be able to use it for regressions on Windows. I reverted GHA back to using tox-conda for regressions.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -39,6 +39,7 @@ conda_deps =
     windows: m2-make
     windows: m2w64-toolchain
     windows: libpython
+    
 install_command =
     windows: python -m pip install --global-option build_ext --global-option --compiler=mingw32 {opts} {packages}
     python -m pip install {opts} {packages}
--- a/tox.ini
+++ b/tox.ini
@@ -39,7 +39,6 @@ conda_deps =
     windows: m2-make
     windows: m2w64-toolchain
     windows: libpython
-    
 install_command =
     windows: python -m pip install --global-option build_ext --global-option --compiler=mingw32 {opts} {packages}
     python -m pip install {opts} {packages}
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -34,6 +34,15 @@ deps =
...
     pytest-cov
 
+conda_deps =
+    windows: m2-base
+    windows: m2-make
+    windows: m2w64-toolchain
+    windows: libpython
+
+conda_channels =
+    windows: msys2
```
garmin-mjames, line 44: The msys2 channel has been a default on Windows for some time: https://github.com/conda/conda/pull/2999

tox.ini
```diff
@@ -34,6 +34,15 @@ deps =
...
+
```
ktbarrett, line 45:
```suggestion

```

tox.ini
```diff
@@ -34,6 +34,11 @@ deps =
     pytest
     pytest-cov
 
+conda_deps =
+    windows: m2-base
+    windows: m2-make
+    windows: m2w64-toolchain
+    windows: libpython
```
ktbarrett, line 41:
```suggestion
    windows: libpython
    
```

tox.ini
```diff
@@ -34,6 +34,12 @@ deps =
...
+
```
ktbarrett, line 42:
```suggestion

```

garmin-mjames:
`tox-conda` is still broken, it doesn't add the conda environment paths correctly. See https://github.com/tox-dev/tox-conda/issues/50.
It might "work" because the toolchain is installed in the base environment:
https://github.com/cocotb/cocotb/blob/9154475c6b8087820719c9b5a34f54d864411380/.github/workflows/regression-tests.yml#L205

ktbarrett:
We can come back to this at a later point in time (if ever).

PR #2188, opened 2020-11-11
Fix deprecation warning for str(handle)

cmarqu:
The ``AttributeError`` will be thrown for ``RegionObject``s:

```
FutureWarning: `str(ConstantObject)` is deprecated, and in future will return `ConstantObject._path`. To get a string representation of the value, use `str(ConstantObject.value)`.
```

This can be seen with `test_discovery`.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -203,7 +203,7 @@ class RegionObject(SimHandleBase):
                     self._log.debug("Yielding index %d from %s (%s)", subindex, name, type(subhdl))
                     yield subhdl
             else:
-                self._log.debug("Yielding %s (%s)", name, handle)
+                self._log.debug("Yielding %s of type %s (%s)", name, type(handle), handle._path)
                 yield handle
 
     def _discover_all(self):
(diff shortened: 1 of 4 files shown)


ktbarrett:
What `DeprecationWarning` does it fix? The change doesn't make much sense to me, nothing that inherits from `RegionObject` has a `.value` attribute.

cmarqu:
The warning is https://github.com/cocotb/cocotb/blob/master/cocotb/handle.py#L743-L748

cmarqu:
It's iterating over sub-handles of ``RegionObject`` which don't need to be ``RegionObject``s.

ktbarrett (dismissed):

cocotb/handle.py
```diff
@@ -203,7 +203,10 @@ def __iter__(self):
                     self._log.debug("Yielding index %d from %s (%s)", subindex, name, type(subhdl))
                     yield subhdl
             else:
-                self._log.debug("Yielding %s (%s)", name, handle)
+                try:
+                    self._log.debug("Yielding %s of type %s (value %s)", name, type(handle), handle.value)
```
ktbarrett, line 207: I think what we want to see here is the path, and we don't necessarily care about the value of the sub handles.

cmarqu: This would be a change of behavior - no problem for debug output though.

ktbarrett: Didn't `str(handle)` print the path at one point /  for certain types of handles?

cmarqu: Could be. Let's just take `_path`  now.

ktbarrett (approved):

Issue #1872, opened 2020-05-27
RFC: Waivers/exclusions for tests


imphil:
I'm proposing to add the ability to have waive or exclude certain tests through a waiver/exclusion file which is separate from the tests itself. This proposal is concerned mainly with our own tests, but could be equally useful for the users of cocotb.

Currently, we mark tests which are known to fail or produce different results in our test suite with `skip` or `expect_*` inside `[@cocotb.test()](https://docs.cocotb.org/en/latest/library_reference.html#cocotb.test)`. This is problematic for a variety of reasons:
- It badly clutters the test files
- It makes it harder to add meta-data to exclusions
- It doesn't scale well for fine-grained exclusions (e.g. targeted exclusions for a specific version of a specific tool)

So what I'm proposing instead is a separate file to collect those exclusions/waivers.

Each waiver/exclusion entry should have:
- The name of the test to be excluded (`classname.testname`)
- How to exclude the test (expect failure, skip, etc. Effectively what we have in the cocotb.test decorator today)
- Meta-information or a comment: why is the test excluded? A link to a bug or a known simulator issue (see also #1871)

I'm not set on any format for that, but I imagine making it (close to) executable Python will make our life much easier.

I'm also not quite sure what the best way to structure waivers is, and how much of inclusion/normalization vs. duplication we should do. For example, I can imagine one waiver file per tool/version combination, which gets copied when we qualify a new tool. The less logic there are in waivers the less likely are accidental modifications of a waiver which doesn't belong to the tool currently being tested.

PR #2191, opened 2020-11-12
Add pass_total value for VCS in test_array.test_discover_all

cmarqu:

### DIFF
--- a/tests/test_cases/test_array/test_array.py
+++ b/tests/test_cases/test_array/test_array.py
@@ -351,6 +351,8 @@ async def test_discover_all(dut):
             pass_total = 1006
         else:
             pass_total = 1038
+    elif cocotb.LANGUAGE in ["verilog"] and cocotb.SIM_NAME.lower().startswith(("chronologic simulation vcs")):
+        pass_total = 606
     else:
         pass_total = 1078
 


tests/test_cases/test_array/test_array.py
```diff
@@ -351,6 +351,8 @@ async def test_discover_all(dut):
             pass_total = 1006
         else:
             pass_total = 1038
+    elif cocotb.LANGUAGE in ["verilog"] and cocotb.SIM_NAME.lower().startswith(("chronologic simulation vcs")):
```
ktbarrett, line 354: What about VHDL? Does it not run? Is it the same as the else clause?

garmin-mjames:
Wouldn't it be easier to export and use `os.getenv('SIM_LOWERCASE')` for simulator-specific behavior in regression tests?

tests/test_cases/test_array/test_array.py
cmarqu: We don't support VHDL in VCS (it has a strange VHPI implementation).

cmarqu:
> Wouldn't it be easier to export and use `os.getenv('SIM_LOWERCASE')` for simulator-specific behavior in regression tests?

That could be something for a different PR, yes.

ktbarrett (approved):

ktbarrett:
@garmin-mjames That might be of use for what we come up for #1872.

Issue #89, opened 2013-09-05
Move clock generation into C++


chiggs:
Having moved clock into python to ensure externals have a scheduling trigger we have diminished performance. We should use a c clock and create a monitor while external is in parties

garmin-mjames:
This would be easier with `CancelledError` in order to clean up the clock between tests from a coroutine, rather than have the scheduler manage yet another thing.

themperek:
This may be required/needed first #2208 ?

garmin-mjames:
If it's just a clock generator, it should only need a timed callback (or two for different duty cycle) and a handle to the signal to be toggled.

garmin-mjames:
Possible GPI interface for clocks:

```cpp
class GpiClkHdl;
typedef GpiClkHdl *gpi_clk_hdl;

/**
 * Create a clock object associated with a signal handle.
 * @param[in] gpi_hdl       Handle to clock signal
 * @param[in] high_steps    Number of timesteps clock is high per period
 * @param[in] low_steps     Number of timesteps clock is low per period
 * @returns                 Clock object handle (NULL on error)
*/
GPI_EXPORT gpi_clk_hdl gpi_clock_create (gpi_sim_hdl gpi_hdl, uint64_t high_steps, uint64_t low_steps);

/**
 * Start driving a clock.
 * @param[in] clk_hdl       Handle to clock object
 * @param[in] cycles        Number of clock periods to cycle. Negative value will cycle clock until stopped
 * @param[in] start_high    Non-0 to start clock high, 0 to start clock low
 * @returns                 0 for successful start, non-0 on error
*/
GPI_EXPORT int gpi_clock_start (gpi_clk_hdl clk_hdl, uint64_t cycles, int start_high);

/**
 * Stop driving a clock.
 * If the clock is already stopped, this has no effect.
 * @param[in] clk_hdl       Handle to clock object
 * @returns                 0 if clock was already stopped, non-0 if clock was started
*/
GPI_EXPORT int gpi_clock_stop (gpi_clk_hdl clk_hdl);

/**
 * Delete a clock object.
 * If the clock is started, this will stop the clock before deleting.
 * @param[in] clk_hdl       Handle to clock object
 * @returns                 0 if clock was already stopped, non-0 if clock was started
*/
GPI_EXPORT int gpi_clock_delete (gpi_clk_hdl clk_hdl);

```
Maybe a `gpi_clock_is_started()`, although that should be managed by the GPI user.

[...]

PR #2193, opened 2020-11-12
Add example showing a TB clock generator in HDL

### LINKED ISSUE #2192: Add example showing the use of a clock generator implemented in HDL
We should have an example where a design is driven by both a clock generator implemented in HDL and in Python to show 
(a) the difference in simulation performance and 
(b) how the HDL version is done.

Related to #2115.

cmarqu:
Closes #2192

### DIFF
--- a/documentation/source/examples.rst
+++ b/documentation/source/examples.rst
@@ -6,6 +6,44 @@ Apart from the examples covered with full tutorials in the previous sections,
 the directory :file:`cocotb/examples/` contains some more smaller modules you may want to take a look at.
 
 
+.. _clkgen_in_hdl:
+
+Clock Generator in HDL
+======================
+
+The directory :file:`cocotb/examples/clkgen_in_hdl`
+contains three HDL modules:
+a very simple counter design under test (``dut``),
+a clock generator (``clkgen``),
+and a testbench (``tb``) instantiating both the counter and the clock generator.
+
+All HDL modules are implemented in both VHDL and SystemVerilog.
+
+The example also includes a Python testbench with two testcases,
+``test_clkgen_hdl`` which uses the HDL clock generator to drive the DUT, and
+``test_clkgen_python`` which uses :class:`cocotb.clock.Clock` to drive the DUT.
+
+Note that :envvar:`TOPLEVEL` and :data:`cocotb.top` point to the testbench instead of the DUT.
+
+Here is the last part of the output of this example (with :envvar:`COCOTB_REDUCED_LOG_FMT` set),
+showing the different runtimes::
+
+    2469132.00ns INFO     ****************************************************************************************
+                          ** TEST                            PASS/FAIL  SIM TIME(NS)  REAL TIME(S)  RATIO(NS/S) **
+                          ****************************************************************************************
+                          ** test_clkgen.test_clkgen_hdl       PASS      1234566.00          0.03   41720542.34  **
+                          ** test_clkgen.test_clkgen_python    PASS      1234566.00         29.95     41219.68  **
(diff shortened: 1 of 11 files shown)


garmin-mjames:
You need to add the new example in examples/Makefile.

examples/clkgen_in_hdl/test_clkgen.py
```diff
@@ -0,0 +1,74 @@
...
+
+async def progress_reporter(tb_hdl):
+    """Report progress by printing intermediate counter values."""
+    while True:
+        # With the following Timer, we sleep exactly the right time
+        # to end up printing every 10000th counter value.
+        # We could also wake up on each counter value change and
+        # print the value if it was modulo 10000,
+        # but simulation performance would be much worse.
+        await Timer(10_000 * 10, "ns")
```
garmin-mjames, line 31:
```suggestion
        await Timer(10000 * 10, "ns")
```
Underscores were apparently added in 3.6: https://docs.python.org/3.6/whatsnew/3.6.html#whatsnew36-pep515

ktbarrett: Damn... they added a lot of QoL stuff in 3.6.

cmarqu: Damn indeed... changed.

examples/clkgen_in_hdl/test_clkgen.py
```diff
@@ -0,0 +1,74 @@
+# This file is public domain, it can be freely copied without restrictions.
+# SPDX-License-Identifier: CC0-1.0
+
+import cocotb
+from cocotb.clock import Clock
+from cocotb.triggers import Timer
+
+MAXCOUNT = 123456
```
ktbarrett, line 8: This is a good case for a plusarg or environment variable communicating to cocotb.

documentation/source/examples.rst
```diff
@@ -6,6 +6,44 @@ Apart from the examples covered with full tutorials in the previous sections,
...
+
+Here is the last part of the output of this example (with :envvar:`COCOTB_REDUCED_LOG_FMT` set),
+showing the different runtimes::
+
+    2469132.00ns INFO     ****************************************************************************************
+                          ** TEST                            PASS/FAIL  SIM TIME(NS)  REAL TIME(S)  RATIO(NS/S) **
+                          ****************************************************************************************
+                          ** test_clkgen.test_clkgen_hdl       PASS      1234566.00          0.03   41720542.34  **
+                          ** test_clkgen.test_clkgen_python    PASS      1234566.00         29.95     41219.68  **
+                          ****************************************************************************************
```
ktbarrett, line 36: Hmmmm... While it does show pretty effectively that using an HDL clock is a major improvement, I don't want to get their hopes up on a 100x performance increase across the board. Is there a way to temper their expectations?

garmin-mjames: Other things based on the clock need to happen. Right now it's either crossing into Python for every clock edge, or in the other case not at all until the end, so a sort of all-or-nothing.

Maybe change to a clock divider and use the slower clock in Python to count? That way a decent amount of "work" is happening in Python, but the speed-up will still be noticeable between the slower clock edges.

cmarqu:
This is obsoleted by #89 once merged.

cmarqu:
This will be obsoleted by the successor of #2293.

documentation/source/examples.rst
jiao-x: Is there currently a way to speed up the simulation in this case when only python is used to generate the clock

Issue #2195, opened 2020-11-15
Add functions to GPI for determining which procedural interfaces are implemented


ktbarrett:
The GPI abstracts away the knowledge of the procedural interface actually used (VPI, VHPI, or FLI), but one disadvantage is it precludes the user from adding new interface specific code. Currently you would have to add a new feature to the GPI abstracting the feature for all interfaces. This isn't always possible or desirable. For example, in the case of VPI's system tasks. Making a GPI-level abstraction for registering new system tasks is not really possible. Instead users should be able to write interface specific code, but will need someway to know whether it's safe to load those extensions/call those functions.

To remedy this limitation new functions should be added to the GPI to allow the user to determine which procedural interfaces are available. For example `gpi_vpi_available()`, `gpi_vhpi_available()` and `gpi_fli_available()`.

garmin-mjames:
So an example would be to check `gpi_vpi_available()` first, then directly interact with the VPI?

Would there be any use cases for a VPI/VHPI/FLI only feature that could be exposed out of the GPI, or should we continue to restrict the interface to the `gpi_*` functions.

ktbarrett:
> So an example would be to check gpi_vpi_available() first, then directly interact with the VPI?

That's correct. 

> Would there be any use cases for a VPI/VHPI/FLI only feature that could be exposed out of the GPI

In this case we either have the user define the system task and call the VPI functions after checking `gpi_vpi_available()`, or defining the system task and calling `gpi_vpi_register_task(....)` which would call the vpi functions or return an error code if it isn't available?

garmin-mjames:
The second part was more of a general question about how strict of an abstraction layer GPI should be over the individual interfaces.

There are some features that are only supported by a subset, like the VPI system tasks, or for VHPI/FLI there's driver creation. Does non-universal functionality belong in the GPI, and how should it be exposed to GPI users if so.

PR #2196, opened 2020-11-15
Adds $cocotb_pass_test() and $cocotb_fail_test() Verilog tasks

### LINKED ISSUE #2093: Support for SystemVerilog Assertions (SVA)

### LINKED ISSUE #2094: Introduce new system tasks to end the test with pass or fail

ktbarrett:
These two new system tasks end the currently running cocotb test with either a `TestSuccess` or a `TestFailure`, respectively. These system tasks are currently only implemented in VPI.

Closes #2094. Closes #2093.

I'm not very certain about this approach. It buries more assumption about the use of cocotb's testing framework into the GPI. That seems like a bleeding of responsibility to me. But #2195 is in the way if we wanted to move this into `gpi_embed`. In which case the registering code would be run in `gpi_embed_init` before elaboration has finished.

Suggestions welcome. Pinging @umarcor. Perhaps there should be a system task that returns the currently running test's name so the user doesn't have to make assumptions about which test is running in cocotb?

### DIFF
--- a/tests/test_cases/test_end_system_tasks/test_end_system_tasks.sv
+++ b/tests/test_cases/test_end_system_tasks/test_end_system_tasks.sv
@@ -1,0 +1,17 @@
+//-----------------------------------------------------------------------------
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//-----------------------------------------------------------------------------
+
+`timescale 1 ns / 1 ps
+
+module test_end_system_tasks;
+
+initial begin
+    #10ns $cocotb_pass_test();
+    #10ns $cocotb_fail_test();
+end
+
+endmodule
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -263,16 +263,14 @@ def _initialise_testbench(argv_):
 
 def _sim_event(level, message):
     """Function that can be called externally to signal an event."""
-    # SIM_INFO = 0
-    SIM_TEST_FAIL = 1
-    SIM_FAIL = 2
-    from cocotb.result import TestFailure, SimFailure
+    from cocotb.result import TestFailure, TestSuccess, SimFailure
+    from cocotb.simulator import TEST_FAIL_EVENT, TEST_PASS_EVENT, FAIL_EVENT
 
-    if level is SIM_TEST_FAIL:
-        scheduler.log.error("Failing test at simulator request")
-        scheduler.finish_test(TestFailure("Failure from external source: %s" %
-                              message))
-    elif level is SIM_FAIL:
+    if level == TEST_FAIL_EVENT:
+        scheduler.finish_test(TestFailure(message))
+    elif level == TEST_PASS_EVENT:
+        scheduler.finish_test(TestSuccess(message))
+    elif level == FAIL_EVENT:
         # We simply return here as the simulator will exit
         # so no cleanup is needed
         msg = ("Failing test at simulator request before test run completion: "
--- a/cocotb/share/include/gpi.h
+++ b/cocotb/share/include/gpi.h
@@ -116,6 +116,7 @@ typedef enum gpi_event_e {
     SIM_INFO = 0,
     SIM_TEST_FAIL = 1,
     SIM_FAIL = 2,
+    SIM_TEST_PASS = 3
 } gpi_event_t;
 
 // Functions for controlling/querying the simulation state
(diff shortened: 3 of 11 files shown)


cmarqu:
Runs fine with both Xcelium (20.03) and VCS (Q-2020.03-1) - the latter with its "normal" segfaults in some tests.

ktbarrett:
@cmarqu I totally forgot to add the tests, could you try again?

ktbarrett:
Tested the actual tests on Questa 10.5a, 10.7c, and 2020.1.

cmarqu:
> @cmarqu I totally forgot to add the tests, could you try again?

The new tests also run fine.

eric-wieser:
It seems slightly awkward to use these with python files containing multiple tests, especially if the user sets the `TESTCASE`. In the case of this PR, the test will fail if I set `TESTCASE=test_system_task_fails`.

I think if we're going to add this, we should at least construct a test-cases for our unit tests that does not suffer from this problem - perhaps something simple like having the python code set a `current_test_no` signal at the beginning of each test.

tests/test_cases/test_end_system_tasks/test_end_system_tasks.sv
```diff
@@ -0,0 +1,17 @@
+//-----------------------------------------------------------------------------
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
eric-wieser, line 5:
```suggestion
// SPDX-License-Identifier: BSD-3-Clause
```

ktbarrett:
@eric-wieser I mentioned in the OP that maybe we should add a system task that returns the current cocotb test name for the purpose you listed. Maybe another for the current module? I was hoping I'd get some direction from @umarcor seeing as he is interested in running HDL tests alongside cocotb. I'll consider this a WIP.

eric-wieser:
For what it's worth, at my old employer we typically had either:

* Normal cocotb tests operated almost entirely by python (exception clock generation)
* HDL tests with a cocotb file containing only a single test, a thin python wrapper that did nothing other than wait for a `test_done` signal and check the value of `test_status` and `test_message` (a string).

Even this second case doesn't really benefit in any substantial way from `cocotb_fail_test`.

Clearly the place to discuss this though is the issue, not this PR.

garmin-mjames (commented):
I agree that it would be better suited to be outside the GPI.

`SIM_FAIL` should also probably be `SIM_END` or `SIM_FINISH`. It's more descriptive and neutral.

PR #2198, opened 2020-11-15
Use setuptools_scm to automatically generate version number

### LINKED ISSUE #2197: Use setuptools_scm for generating a version number
[setuptools_scm](https://github.com/pypa/setuptools_scm/) can automatically generate version numbers based on git tags and versioning information. This will allow us to exactly identify which commit a user is reporting an issue with if it is not a released version. It will also remove one step of the release process.

ktbarrett:
Closes #2197.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools", "wheel", "setuptools_scm"]
 build-backend = "setuptools.build_meta"
 
 [tool.towncrier]
@@ -36,3 +36,4 @@ build-backend = "setuptools.build_meta"
         directory = "change"
         name = "Changes"
         showcontent = true
+
--- a/setup.py
+++ b/setup.py
@@ -73,9 +73,6 @@ def package_files(directory):
     return paths
 
 
-# this sets the __version__ variable
-exec(read_file(path.join('cocotb', '_version.py')))
-
 # store log from build_libs and display at the end in verbose mode
 # see https://github.com/pypa/pip/issues/6634
 log_stream = StringIO()
@@ -87,7 +84,11 @@ log.addHandler(handler)
 setup(
     name='cocotb',
     cmdclass={'build_ext': build_ext},
-    version=__version__,  # noqa: F821
+    use_scm_version=dict(
+        write_to='cocotb/_version.py',
+        write_to_template='__version__ = {version!r}',
+        version_scheme='release-branch-semver'
+    ),
     description='cocotb is a coroutine based cosimulation library for writing VHDL and Verilog testbenches in Python.',
     url='https://docs.cocotb.org',
     license='BSD',
@@ -96,6 +97,7 @@ setup(
     author='Chris Higgs, Stuart Hodgson',
     maintainer='cocotb contributors',
     maintainer_email='cocotb@lists.librecores.org',
+    setup_requires=['setuptools_scm'],
     install_requires=[],
     python_requires='>=3.5',
     packages=find_packages(),
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -148,6 +148,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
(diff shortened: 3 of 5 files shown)


ktbarrett:
Fixed the versioning scheme so now it's generating the correct version.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools", "wheel", "setuptools_scm"]
```
cmarqu, line 2: The example at https://pypi.org/project/setuptools-scm/ says "Note that the ``toml`` extra must be supplied.":
```
requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4"]
```
Maybe it works fine without it in most cases but maybe not all?

eric-wieser:
I've no experience with setuptools_scm, but this looks reasonable assuming it works.

setup.py
```diff
@@ -87,7 +84,11 @@ def package_files(directory):
 setup(
     name='cocotb',
     cmdclass={'build_ext': build_ext},
-    version=__version__,  # noqa: F821
+    use_scm_version={
+        'write_to': 'cocotb/_version.py',
+        'write_to_template': '__version__ = "{version}"',
```
eric-wieser, line 89: nit: 
```suggestion
        'write_to_template': '__version__ = {version!r}',
```
although it only makes a difference if the version contains quotes or special characters.

setup.py
```diff
@@ -87,7 +84,11 @@ def package_files(directory):
 setup(
     name='cocotb',
     cmdclass={'build_ext': build_ext},
-    version=__version__,  # noqa: F821
+    use_scm_version={
+        'write_to': 'cocotb/_version.py',
```
eric-wieser, line 88: Stylistically, I sometimes like using
```suggestion
    use_scm_version=dict(
        write_to='cocotb/_version.py',
```
just to make the nested arguments resemble the non-nested ones.

pyproject.toml
ktbarrett: That's if you are using pyproject.toml to configure the file.

setup.py
```diff
@@ -87,7 +84,11 @@ def package_files(directory):
...
+        'write_to_template': '__version__ = "{version}"',
+        'version_scheme': 'release-branch-semver'
+    },
```
ktbarrett, line 91:
```suggestion
    ),
```

ktbarrett:
@eric-wieser The version is `cocotb-1.5.0.dev173+gdcafa990`. `173` is the number of commits on the current branch since the last release and `dcafa990` is the commit sha. On releases it should just be `cocotb-1.5.0`. The versioning scheme detects version branches (by checking if the branch name is the same as the last version tag) and will rev the patch version rather than the minor version in those cases.
Screenshot from 2020-11-16 07-36-09

eric-wieser:
Since @imphil will likely be doing the release process, I'll leave it up to him to decide. Certainly this looks like an improvement for users installing from master, it just might add some slight friction at release time. Or it might just work.

garmin-mjames:
Why is the version showing up as `v0.1.0.dev1+g5414f47` in the regressions?
https://github.com/cocotb/cocotb/pull/2198/checks?check_run_id=1406584885#step:22:103

ktbarrett:
@garmin-mjames Shallow clone comes in without the tags. `setuptools_scm` uses git tags to determine the version number. We could probably add arguments to `action/checkout` step to have it retrieve the tags. I will make that an action.

ktbarrett:
Added `fetch_depth: 0` to workflow to ensure tags are cloned. Fixed a typo causing failure.

imphil:
Thanks for this change. I finally had time to experiment a bit with the versioning scheme (I know setuptools_scm from the past, and it works well itself). The tagging and branching detection works well, we just need to create a release branch right when we first do the RC release, otherwise we get slightly confusing version number (i.e. after tagging v1.5.0rc1, the next version number is v1.6.0.dev0 when we do everything on master). But in the worst case it's a slight "ugliness", but never a problem for actual releases.

imphil (approved):

PR #2200, opened 2020-11-16
Increase coverage in xunit_reporter.py

### LINKED ISSUE #920: cocotb.hook does not do anything unless in a module present in COCOTB_HOOKS
Marking a function with `@cocotb.hook` is insufficient for it to do anything - you also have to set the `COCOTB_HOOKS` environment variable to contain the module name where the decorated function resides.

Either we should document this, or we should just remove `cocotb.hook`. None of our tests test `cocotb.hook`. Is anyone using it?

ktbarrett:
Adding more deprecations, removals, and tests to increase code coverage. While the deprecated code is still being counted towards the total coverage it will be removed after 1.5 releases and 2.0 development begins. If desired we can add `pragma: no cover` comments to deprecated code blocks to reflect post-removal coverage numbers.

~~Deprecate `cocotb.hook` Closes #920.~~ Remove internal dead code in xunit_reporter.py. ~~Deprecate `cocotb.utils.unpack` and `cocotb.utils.pack`.~~ Add test for a test failure that escapes the regression manager.

### DIFF
--- a/cocotb/xunit_reporter.py
+++ b/cocotb/xunit_reporter.py
@@ -28,40 +28,6 @@
 from xml.etree.ElementTree import Element, SubElement
 import xml.etree.ElementTree as ET
 
-import mmap
-from io import StringIO
-
-TRUNCATE_LINES = 100
-
-
-# file from  http://stackoverflow.com/questions/136168/get-last-n-lines-of-a-file-with-python-similar-to-tail
-class File(StringIO):
-
-    def countlines(self):
-        buf = mmap.mmap(self.fileno(), 0)
-        lines = 0
-        while buf.readline():
-            lines += 1
-        return lines
-
-    def head(self, lines_2find=1):
-        self.seek(0)                            # Rewind file
-        return [self.next() for x in range(lines_2find)]
-
-    def tail(self, lines_2find=1):
-        self.seek(0, 2)                         # go to end of file
-        bytes_in_file = self.tell()
-        lines_found, total_bytes_scanned = 0, 0
-        while (lines_2find+1 > lines_found and
-               bytes_in_file > total_bytes_scanned):
-            byte_block = min(1024, bytes_in_file-total_bytes_scanned)
-            self.seek(-(byte_block+total_bytes_scanned), 2)
-            total_bytes_scanned += byte_block
-            lines_found += self.read(1024).count('\n')
-        self.seek(-total_bytes_scanned, 2)
-        line_list = list(self.readlines())
-        return line_list[-lines_2find:]
-
 
 class XUnitReporter:
 
@@ -84,30 +50,6 @@ class XUnitReporter:
             testsuite = self.last_testsuite
         self.last_property = SubElement(testsuite, "property", **kwargs)
         return self.last_property
-
-    def update_testsuite(self, testsuite=None, **kwargs):
-        if testsuite is None:
-            testsuite = self.last_testsuite
-        for k in kwargs:
-            testsuite.set(k, str(kwargs[k]))
-
-    def update_testsuites(self, **kwargs):
-        for k in kwargs:
-            self.results.set(k, str(kwargs[k]))
-
-    def add_log(self, logfile, testcase=None):
-        if testcase is None:
-            testcase = self.last_testcase
-        log = SubElement(testcase, "system-out")
-        f = File(logfile, 'r+')
-        lines = f.countlines()
-        if lines > (TRUNCATE_LINES * 2):
-            head = f.head(TRUNCATE_LINES)
-            tail = f.tail(TRUNCATE_LINES)
-            log.text = "".join(head + list("[...truncated %d lines...]\n" %
-                               ((lines - (TRUNCATE_LINES*2)))) + tail)
-        else:
-            log.text = "".join(f.readlines())
 
     def add_failure(self, testcase=None, **kwargs):
         if testcase is None:
--- a/tests/test_cases/test_failure/Makefile
+++ b/tests/test_cases/test_failure/Makefile
@@ -1,0 +1,14 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+MODULE := test_failure
+
+# ensure the test runs, squash any error code, and ensure a failing test was reported
+.PHONY: override_for_this_test
+override_for_this_test:
+	-$(MAKE) all
+	$(call check_for_results_file)
+	test $$(../../../bin/combine_results.py | grep "Failure in testsuite" | wc -l) -eq 1 && rm -f results.xml
+
+include ../../designs/sample_module/Makefile
--- a/tests/test_cases/test_failure/test_failure.py
+++ b/tests/test_cases/test_failure/test_failure.py
@@ -4,7 +4,7 @@
 """
 Tests the failure path for tests in the regression.py and xunit_reporter.py.
 
-The Makefile in the same repo is specially set up to squash any error code due
+The Makefile in this folder is specially set up to squash any error code due
 to a failing test and ensures the failing test is reported properly.
 """
 import cocotb
--- a/tests/test_cases/test_failure/test_failure.py
+++ b/tests/test_cases/test_failure/test_failure.py
@@ -1,0 +1,15 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+"""
+Tests the failure path for tests in the regression.py and xunit_reporter.py.
+
+The Makefile in the same repo is specially set up to squash any error code due
+to a failing test and ensures the failing test is reported properly.
+"""
+import cocotb
+
+
+@cocotb.test()
+async def test_fail():
+    assert False
--- a/documentation/source/newsfragments/2200.removal.rst
+++ b/documentation/source/newsfragments/2200.removal.rst
@@ -1,0 +1,1 @@
+The undocumented class ``cocotb.xunit_reporter.File`` has been removed.


tests/test_cases/test_failure/test_failure.py
```diff
@@ -0,0 +1,10 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
cmarqu, line 4:
```suggestion
# SPDX-License-Identifier: BSD-3-Clause
```

tests/test_cases/test_failure/Makefile
```diff
@@ -0,0 +1,15 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
cmarqu, line 4:
```suggestion
# SPDX-License-Identifier: BSD-3-Clause
```

bin/combine_results.py
```diff
@@ -79,14 +79,13 @@ def main():
             testcase_count += 1
             for failure in testcase.iter('failure'):
                 if args.set_rc:
-                    rc = 1
+                    rc += 1
```
eric-wieser, line 82: This seems like quite an odd thing to do, I don't know of other applications that use the return code to indicate a count.

cocotb/utils.py
```diff
@@ -174,7 +174,13 @@ def pack(ctypes_obj):
 
     Returns:
         New Python string containing the bytes from memory holding *ctypes_obj*.
+
+    .. deprecated:: 1.5
+        This function is deprecated. The functionality is a duplicate of :py:class:`struct.Struct`.
```
eric-wieser, line 179:
```suggestion
        This function is deprecated, use `bytes(ctypes_obj)` instead.
```

cocotb/utils.py
```diff
@@ -196,7 +202,13 @@ def unpack(ctypes_obj, string, bytes=None):
         :exc:`ValueError`: If length of *string* and size of *ctypes_obj*
             are not equal.
         :exc:`MemoryError`: If *bytes* is longer than size of *ctypes_obj*.
+
+    .. deprecated:: 1.5
+        This function is deprecated. The functionality is a duplicate of :py:class:`struct.Struct`.
```
eric-wieser, line 207: I don't agree with this comment. I think we're still right to deprecate it, but `ctypes.Structure` and `struct.Struct` are fairly incompatible APIs.

I think an exact replacement is:
```
memoryview(ctypes_obj).cast('B')[:bytes] = string
```
while something the user might also be looking for one of
* `ctypes.pointer(ctypes_obj).contents = type(ctypes_obj).from_buffer_copy(string)`
* `obj = MyStruct.from_buffer_copy(string)`

bin/combine_results.py
ktbarrett: It was a hack, certainly. Should I just write a local python script instead of modifying this script?

cocotb/utils.py
```diff
@@ -174,7 +174,13 @@ def pack(ctypes_obj):
 
     Returns:
         New Python string containing the bytes from memory holding *ctypes_obj*.
+
+    .. deprecated:: 1.5
+        This function is deprecated, use ``bytes(ctypes_obj)`` instead.
     """
+    warnings.warn(
+        "This function is being removed, use ``struct.Struct`` instead.",
```
ktbarrett, line 182:
```suggestion
        "This function is deprecated and will be removed, use ``bytes(ctypes_obj)`` instead.",
```

bin/combine_results.py
ktbarrett: I can grep the output. Not sure why I didn't think of that.

cocotb/decorators.py
```diff
@@ -451,15 +451,23 @@ def decorator(f):
...
+    Decorator to mark a function as a hook for cocotb.
 
     Used as ``@cocotb.hook()``.
 
     All hooks are run at the beginning of a cocotb test suite, prior to any
-    test code being run."""
+    test code being run.
+
+    .. deprecated:: 1.5
+        Hooks are deprecated, and the functionality can be replaced with normal module-level Python code.
```
eric-wieser, line 463: Is this true? I though you could run arbitrary coroutines from hooks, even if it didn't work well at all.

eric-wieser: Perhaps we should give an example somewhere of how to do hook-like things, such as
```python
def with_reset(test_func):
    @functools.wraps(test_func)
    def wrapped(dut):
        dut.rst <= 1
        await Timer(10, "ns")
        dut.rst <= 0
        await RisingEdge(dut.clk)
        await test_func(dut)
    return wrapped

@cocotb.test()
@with_reset
async def my_test(dut):
    # dut is already reset and ready to go!
```

cocotb/utils.py
```diff
@@ -196,7 +202,13 @@ def unpack(ctypes_obj, string, bytes=None):
         :exc:`ValueError`: If length of *string* and size of *ctypes_obj*
             are not equal.
         :exc:`MemoryError`: If *bytes* is longer than size of *ctypes_obj*.
+
+    .. deprecated:: 1.5
+        This function is deprecated. The functionality is a duplicate of :py:class:`struct.Struct`.
     """
+    warnings.warn(
+        "This function is being removed, use ``struct.Struct`` instead.",
```
eric-wieser, line 210:
```suggestion
        "This function is being removed, use ``memoryview(ctypes_obj)[:bytes] = string`` instead.",
```

cocotb/utils.py
eric-wieser:
```suggestion
        This function is deprecated. Use use ``memoryview(ctypes_obj)[:bytes] = string`` instead.
        You may also be looking for the :func:`~ctypes._CData.from_buffer_copy` method of :mod:`ctypes` types,
        if you are trying to convert bytes to a structure, or ``ctypes.pointer(obj).contents = new_obj`` if you are trying
        to modify a structure in place.
```

cocotb/decorators.py
ktbarrett: I guess I'm still not sure what the whole point of hooks were, so I had a hard time coming up with an alternative.

eric-wieser: Yeah, I'm not sure either. I found hooks when looking for a way to do the above, but decided that they weren't actually useful for that. I don't think we should give the above example in the docs for `hook`s, but I think it would be good tutorial content.

cocotb/utils.py
```diff
@@ -196,7 +202,16 @@ def unpack(ctypes_obj, string, bytes=None):
         :exc:`ValueError`: If length of *string* and size of *ctypes_obj*
             are not equal.
         :exc:`MemoryError`: If *bytes* is longer than size of *ctypes_obj*.
+
+    .. deprecated:: 1.5
+        This function is deprecated. Use use ``memoryview(ctypes_obj)[:bytes] = string`` instead.
+        You may also be looking for the :func:`~ctypes._CData.from_buffer_copy` method of :mod:`ctypes` types,
+        if you are trying to convert bytes to a structure, or ``ctypes.pointer(obj).contents = new_obj`` if you are trying
+        to modify a structure in place.
```
eric-wieser, line 210:
```suggestion
        This function is deprecated, as its functionality is now built into
        Python 3 as ``memoryview(ctypes_obj)[:bytes] = string``. You
        may also be looking for the :func:`~ctypes._CData.from_buffer_copy`
        method of :mod:`ctypes` types, if you are trying to convert
        :class:`bytes` to a :class:`ctypes.Structure` instance, or
        ``ctypes.pointer(obj).contents = new_obj`` if you are trying to modify
        a :class:`ctypes.Structure` in place.
```

eric-wieser: (with line wrapping, and a couple more links)

eric-wieser:
Needs (multiple?) release notes

ktbarrett:
It's still picking up the failing results.xml at the end of the regression. Will just `rm` it after the `test_failure` test succeeds.

ktbarrett:
Reduce the scope of this PR. Moving other changes into their own PR so I can have separate newsfragments. Also won't gate this obvious change on other changes.

cocotb/xunit_reporter.py
```diff
@@ -28,41 +28,9 @@
 from xml.etree.ElementTree import Element, SubElement
 import xml.etree.ElementTree as ET
 
-import mmap
-from io import StringIO
-
 TRUNCATE_LINES = 100
```
eric-wieser, line 31: This is no longer used, you removed the only code that used it.

cocotb/xunit_reporter.py
```diff
@@ -28,41 +28,9 @@
...
 
 
-# file from  http://stackoverflow.com/questions/136168/get-last-n-lines-of-a-file-with-python-similar-to-tail
-class File(StringIO):
```
eric-wieser, line 38: Probably should have a release note, "The undocumented `cocotb.xunit_report.File` class has been removed"

cocotb/xunit_reporter.py
```diff
@@ -85,30 +53,6 @@ def add_property(self, testsuite=None, **kwargs):
...
-        log = SubElement(testcase, "system-out")
-        f = File(logfile, 'r+')
-        lines = f.countlines()
-        if lines > (TRUNCATE_LINES * 2):
-            head = f.head(TRUNCATE_LINES)
-            tail = f.tail(TRUNCATE_LINES)
-            log.text = "".join(head + list("[...truncated %d lines...]\n" %
-                               ((lines - (TRUNCATE_LINES*2)))) + tail)
-        else:
-            log.text = "".join(f.readlines())
```
eric-wieser, line 110: It would be cool if the log lines actually did end up in the xunit output - but perhaps the approach used here isn't useful.

tests/test_cases/test_failure/Makefile
```diff
@@ -0,0 +1,14 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+MODULE := test_failure
+
+# squash error code from simulator and ensure the cocotb test passed
```
eric-wieser, line 7: This comment is a bit confusing, given that we're testing the xunit reporting does something sensible when CI does _not_ pass.

tests/test_cases/test_failure/test_failure.py
```diff
@@ -0,0 +1,9 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+import cocotb
```
eric-wieser, line 4: A docstring explaining the purpose of this file would be good (and mentions that the makefile is special to handle it)

cmarqu (approved):

tests/test_cases/test_failure/test_failure.py
```diff
@@ -0,0 +1,15 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+"""
+Tests the failure path for tests in the regression.py and xunit_reporter.py.
+
+The Makefile in the same repo is specially set up to squash any error code due
```
eric-wieser, line 7:
```suggestion
The Makefile in this folder is specially set up to squash any error code due
```

PR #2201, opened 2020-11-17
Deprecate cocotb.hook

### LINKED ISSUE #920: cocotb.hook does not do anything unless in a module present in COCOTB_HOOKS

ktbarrett:
Closes #920. Deprecates `cocotb.hook`.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -451,7 +451,7 @@ class _decorator_helper(type):
 
 @public
 class hook(coroutine, metaclass=_decorator_helper):
-    """
+    r"""
     Decorator to mark a function as a hook for cocotb.
 
     Used as ``@cocotb.hook()``.
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -461,16 +461,16 @@ class hook(coroutine, metaclass=_decorator_helper):
 
     .. deprecated:: 1.5
         Hooks are deprecated.
-        Their functionality can be replaced with module-level Python code;
-        higher-priority tests using the ``stage`` option on `cocotb.test()`s;
-        or function decorators;
-        depending on what your original use case entailed.
+        Their functionality can be replaced with module-level Python code,
+        higher-priority tests using the ``stage`` argument to :func:`cocotb.test`\ s,
+        or custom decorators which perform work before and after the tests
+        they decorate.
     """
 
     def __init__(self, f):
         super(hook, self).__init__(f)
         warnings.warn(
-            "Hooks have been deprecated. Module-level code can be written to replace hooks.",
+            "Hooks have been deprecated. See the documentation for suggestions on alternatives.",
             DeprecationWarning, stacklevel=2)
         self.im_hook = True
         self.name = self._func.__name__
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -451,15 +451,25 @@ class _decorator_helper(type):
 
 @public
 class hook(coroutine, metaclass=_decorator_helper):
-    """Decorator to mark a function as a hook for cocotb.
+    """
+    Decorator to mark a function as a hook for cocotb.
 
     Used as ``@cocotb.hook()``.
 
     All hooks are run at the beginning of a cocotb test suite, prior to any
-    test code being run."""
+    test code being run.
+
+    .. deprecated:: 1.5
+        Hooks are deprecated. Their functionality can be replaced with module-level Python code;
+        higher-priority tests using the ``stage`` option; or function decorators; depending on
+        what your original use case entailed.
+    """
 
     def __init__(self, f):
         super(hook, self).__init__(f)
+        warnings.warn(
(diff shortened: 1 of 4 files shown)


cocotb/decorators.py
```diff
@@ -451,15 +451,25 @@ def decorator(f):
...
     Used as ``@cocotb.hook()``.
 
     All hooks are run at the beginning of a cocotb test suite, prior to any
-    test code being run."""
+    test code being run.
+
+    .. deprecated:: 1.5
+        Hooks are deprecated. Their functionality can be replaced with module-level Python code;
+        higher-priority tests using the ``stage`` option; or function decorators; depending on
+        what your original use case entailed.
```
ktbarrett, line 465:
```suggestion
        Hooks are deprecated.
        Their functionality can be replaced with module-level Python code;
        higher-priority tests using the ``stage`` option on `cocotb.test()`s;
        or function decorators;
        depending on what your original use case entailed.
```

cocotb/decorators.py
```diff
@@ -451,15 +451,27 @@ def decorator(f):
...
     All hooks are run at the beginning of a cocotb test suite, prior to any
-    test code being run."""
+    test code being run.
+
+    .. deprecated:: 1.5
+        Hooks are deprecated.
+        Their functionality can be replaced with module-level Python code;
+        higher-priority tests using the ``stage`` option on `cocotb.test()`s;
+        or function decorators;
+        depending on what your original use case entailed.
```
ktbarrett, line 467:
```suggestion
        depending on what your original use case.
```

cocotb/decorators.py
```diff
@@ -451,15 +451,27 @@ def decorator(f):
...
+    """
 
     def __init__(self, f):
         super(hook, self).__init__(f)
+        warnings.warn(
+            "Hooks have been deprecated. Module-level code can be written to replace hooks.",
```
ktbarrett, line 473:
```suggestion
            "Hooks have been deprecated. See the documentation for suggestions on alternatives.",
```

documentation/source/newsfragments/2201.removal.rst
```diff
@@ -0,0 +1,5 @@
+Deprecated :class:`cocotb.hook` and :envvar:`COCOTB_HOOKS`.
+Write module-level Python code;
+create higher-priority tests using the ``stage`` option of :class:`cocotb.test`;
+or define your own function decorators;
+depending on how you were using hooks.
```
ktbarrett, line 5:
```suggestion
See the documentation for :class:`cocotb.hook` for suggestions on alternatives.
```

cocotb/decorators.py
eric-wieser, line 467:
```suggestion
        Hooks are deprecated.
        Their functionality can be replaced with module-level Python code,
        higher-priority tests using the ``stage`` argument to :func:`cocotb.test`s,
        or custom decorators which perform work before and after the tests
        they decorate.
```

documentation/source/newsfragments/2201.removal.rst
eric-wieser: There is a danger here that once we remove them entirely, our release notes will contain dead links.

Perhaps the way to fix this is to turn on bad-references-as-errors in sphinx (`-n`), so that when we try to remove `hook` entirely, we're forced to think about this link.

documentation/source/building.rst
```diff
@@ -168,6 +168,9 @@ Regression Manager
     A comma-separated list of modules that should be executed before the first test.
     You can also use the :class:`cocotb.hook` decorator to mark a function to be run before test code.
 
+    .. deprecated:: 1.5
+        :class:`cocotb.hook` has been deprecated and this variable is no longer used.
```
eric-wieser, line 172:
```suggestion
        :class:`cocotb.hook` is deprecated, and in future this variable will be ignored.
```

ktbarrett: Forgot a `the` so I added it.

documentation/source/newsfragments/2201.removal.rst
ktbarrett: Once we release we could update the link to point to the 1.5 version notes by creating a new intersphinx mapping.

eric-wieser (approved):

cocotb/decorators.py
```diff
@@ -451,15 +451,27 @@ def decorator(f):
 
 @public
 class hook(coroutine, metaclass=_decorator_helper):
-    """Decorator to mark a function as a hook for cocotb.
+    """
```
eric-wieser, line 454:
```suggestion
    r"""
```
to fix flake8

PR #2211, opened 2020-11-20
gitpod: Run verilator with trace, improve general usability

cmarqu:
Injecting commands into the shell history now works correctly.
Only the VNC port is offered to open a preview or browser tab with.
Configure the Theia IDE to show leading and trailing whitespace.

_To check the effect of this PR, run: https://gitpod.io/#https://github.com/cocotb/cocotb/pull/2211_

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -12,12 +12,24 @@ tasks:
       gp open ../hdl/dff.vhdl &&
       gp open Makefile &&
       gp open dff_cocotb.py &&
-      make SIM=icarus
+      make SIM=icarus &&
+      EXTRA_ARGS="-CFLAGS -DVM_TRACE_FST=1 --trace-fst --trace-structs" make SIM=verilator &&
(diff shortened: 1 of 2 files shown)


themperek (approved):
Looks OK. 
Would not be easier to dump vcd/fst with Icarus?

cmarqu:
> Looks OK.
> Would not be easier to dump vcd/fst with Icarus?

With Icarus, you would have to add the Verilog dump functions to the design, right? Just adding some command line options is nicer.

themperek:
> With Icarus, you would have to add the Verilog dump functions to the design, right? Just adding some command line options is nicer.

There are some workarounds but yes.

cmarqu:
> There are some workarounds but yes.

I seem to remember someone saying that it's possible to have two toplevels in the simulator. Then we may be able to use OOMR to refer from the dump module to the DUT. But maybe I'm misremembering.

PR #2212, opened 2020-11-20
Mention Verilog's join/join_any in triggers.py

cmarqu:

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -762,6 +762,8 @@ class Combine(_AggregateWaitable):
     Fires when all of *triggers* have fired.
 
     Like most triggers, this simply returns itself.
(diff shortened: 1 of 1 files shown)


cocotb/triggers.py
```diff
@@ -791,6 +793,8 @@ class First(_AggregateWaitable):
 
     Returns the result of the trigger that fired.
 
+    This is cocotb's equivalent to Verilog's ``join_any``.
```
eric-wieser, line 796: I might go for the weaker wording of "This is similar to Verilog's ``join_any``"

cmarqu: Changed.

ktbarrett (approved):

PR #2213, opened 2020-11-21
Add link to Gitpod with latest released version to docs

cmarqu:


cmarqu:
For some reason, what looks fine locally for me (a link with "Gitpod" as the text) doesn't render nicely in RTD.

ktbarrett:
We aren't using Gitpod anymore, so I think this can close.

Issue #2215, opened 2020-11-22
Feature Request: Support for Xs


jshaker000:
Would it be possible for cocotb to define some kind of wrapper 'Logic' class that can hold a bus with X's on it, and defines and / or / etc as per the Verilog LRM? Sometimes it would be useful to be able to read back X's into python.

It would even be helpful to just have a function like 'isunknown' to know if any values on a bus are Z or X at a given time. This might exist, but if it does I couldnt find it in the docs.

ktbarrett:
#2059

cmarqu:
@jshaker000 ``isunknown`` exists as https://docs.cocotb.org/en/stable/library_reference.html?highlight=x#cocotb.binary.BinaryValue.is_resolvable but it's indeed hard to find.

jshaker000:
Are nets read from the dut always 'BinaryValues' then (which can be implicitly casted to int), or how does that work?

ktbarrett:
@jshaker000 Arrays of logic's and subtypes of that (`sfixed`, `unsigned`, etc.) are `BinaryValue`. Integer nets and constants (`integer`, `natural`, etc.) return `int`. Floating point nets and constants (`real`) return `float`. Boolean nets and constants (`boolean`) return `bool`. String nets and constants (`string`) return `bytes`.

jshaker000:
Thanks, yea it could be clearer on thw wiki but this is enough to get me started. Thanks!

ktbarrett:
@jshaker `int` is a Python builtin, so that's not possible. However, `is_resolvable` could be a free function that would accept `int` or `BinaryValue`, etc.

PR #2219, opened 2020-11-23
Make docstring for BinaryValue.is_resolvable clearer

cmarqu:
Refs #2215.

### DIFF
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -328,7 +328,8 @@ class BinaryValue:
 
     @property
     def is_resolvable(self) -> bool:
-        """Return whether the value contains only resolvable (i.e. no "unknown") bits.
+        """
+        Return whether the value contains only resolvable (i.e. no "unknown") bits.
 
         By default the values ``X``, ``Z``, ``U`` and ``W`` are considered unresolvable.
         This can be configured with :envvar:`COCOTB_RESOLVE_X`.
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -335,7 +335,7 @@ class BinaryValue:
         This can be configured with :envvar:`COCOTB_RESOLVE_X`.
 
         This is similar to the SystemVerilog Assertion ``$isunknown`` system function
-        (with an inverted meaning).
+        or the VHDL function ``is_x`` (with an inverted meaning).
         """
         return not any(char in self._str for char in BinaryValue._resolve_to_error)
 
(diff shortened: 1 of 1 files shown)


cocotb/binary.py
```diff
@@ -327,8 +327,15 @@ def signed_integer(self, val):
     get_value_signed = signed_integer.fget
 
     @property
-    def is_resolvable(self):
-        """Does the value contain any ``X``'s?  Inquiring minds want to know."""
+    def is_resolvable(self) -> bool:
+        """Return whether the value contains only resolvable (i.e. no "unknown") bits.
```
ktbarrett, line 331:
```suggestion
        """
        Return whether the value contains only resolvable (i.e. no "unknown") bits.
```
Is there an "official" recommendation on this style? I was under the impression keeping them on the same line for a one-liner docstring and having them on different lines for multiline docstrings was the recommended style. But what do I know?

cocotb/binary.py
```diff
@@ -327,8 +327,15 @@ def signed_integer(self, val):
...
+
+        By default the values ``X``, ``Z``, ``U`` and ``W`` are considered unresolvable.
+        This can be configured with :envvar:`COCOTB_RESOLVE_X`.
+
+        This is similar to the SystemVerilog Assertion ``$isunknown`` system function
+        (with an inverted meaning).
```
ktbarrett, line 337:
```suggestion
        This is similar to the SystemVerilog Assertion ``$isunknown`` system function
        or the VHDL function ``is_x`` (with an inverted meaning).
```

ktbarrett (commented):
Honestly I think the current docstring is clearer, we should stick with that /s.

cocotb/binary.py
cmarqu: It's in https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

ktbarrett: >  The summary line may be on the same line as the opening quotes or on the next line.

Seems like it doesn't matter. I think most of the current docstrings use the next line.

cmarqu:
> Honestly I think the current docstring is clearer, we should stick with that /s.

The current docstring is actually wrong though, isn't it? Only because we know what we expect inverts its meaning in our heads.
It also doesn't tell the whole truth.

Not that it really matters after all these years and with less than one year to live, but...

ktbarrett:
`/s` implied the previous comment was sarcasm. It's a terrible docstring. Thank you for changing it!

cmarqu:
Ah, heh.

ktbarrett (approved):

PR #2221, opened 2020-11-23
Improve _compat_mapping deprecation warnings

cmarqu:


ktbarrett:
codecov is failing because there is no longer any path checking `_compat_mapping`. Could you add a test for that to `test_deprecated`?

cmarqu:
Added the test, and actually made these proper `DeprecationWarnings` with the right `stacklevel` set.

ktbarrett (approved):

PR #2223, opened 2020-11-24
Clean up and deprecate passing bools to expect_error option to cocotb.test

### LINKED ISSUE #2117: Cleanup @cocotb.test(expect_error=True)
#2107 revealed how `expect_error=True` can mask a problem, because any `Exception` raised by the test will mark it as PASS.

This test passes even though the wrong error is causing the failure:

https://github.com/cocotb/cocotb/blob/89e7bd77b1e9ceb3172b4659e267bc4c42217a0f/tests/test_cases/test_discovery/test_discovery.py#L101-L106

These tests should be cleaned up to use `assert_raises` style or provide the expected `Exception` type.

ktbarrett:
Closes #2117. Cleans up all usages of `expect_error=True` option in the test and also adds a `DeprecationWarning` for it's use.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -510,9 +510,8 @@ class test(coroutine, metaclass=_decorator_helper):
         expect_fail (bool, optional):
             Don't mark the result as a failure if the test fails.
 
-        expect_error (bool or exception type or tuple of exception types, optional):
-            If ``True``, consider this test passing if it raises *any* :class:`Exception`, and failing if it does not.
-            If given an exception type or tuple of exception types, catching *only* a listed exception type is considered passing.
+        expect_error (exception type or tuple of exception types, optional):
+            Mark the result as a pass only if one of the exception types is raised in the test.
             This is primarily for cocotb internal regression use for when a simulator error is expected.
 
             Users are encouraged to use the following idiom instead::
@@ -529,6 +528,10 @@ class test(coroutine, metaclass=_decorator_helper):
             .. versionchanged:: 1.3
                 Specific exception types can be expected
 
+            .. deprecated:: 1.5
+                Passing a :class:`bool` value is now deprecated.
+                Pass a specific :class:`Exception` or a tuple of Exceptions instead.
+
         skip (bool, optional):
             Don't execute this test as part of the regression. Test can still be run
             manually by setting :make:var:`TESTCASE`.
@@ -540,7 +543,7 @@ class test(coroutine, metaclass=_decorator_helper):
     _id_count = 0  # used by the RegressionManager to sort tests in definition order
 
     def __init__(self, f, timeout_time=None, timeout_unit="step",
-                 expect_fail=False, expect_error=False,
+                 expect_fail=False, expect_error=(),
                  skip=False, stage=None):
 
         if timeout_unit is None:
@@ -571,6 +574,11 @@ class test(coroutine, metaclass=_decorator_helper):
         self.timeout_time = timeout_time
         self.timeout_unit = timeout_unit
         self.expect_fail = expect_fail
+        if isinstance(expect_error, bool):
+            warnings.warn(
+                "Passing bool values to `except_error` option of `cocotb.test` is deprecated. "
+                "Pass a specific Exception type instead",
+                DeprecationWarning, stacklevel=2)
         if expect_error is True:
             expect_error = (Exception,)
         elif expect_error is False:
--- a/tests/test_cases/test_cocotb/test_generator_coroutines.py
+++ b/tests/test_cases/test_cocotb/test_generator_coroutines.py
@@ -6,7 +6,7 @@ Tests that specifically test generator-based coroutines
 """
 import cocotb
 from cocotb.triggers import Timer
-from common import clock_gen, _check_traceback
+from common import clock_gen, _check_traceback, assert_raises
 import textwrap
 
 
@@ -91,23 +91,24 @@ def test_yield_list(dut):
 
 
 @cocotb.coroutine
-def syntax_error():
+def erroring_coro():
     yield Timer(100)
     fail  # noqa
 
 
-@cocotb.test(expect_error=True)
-def test_coroutine_syntax_error(dut):
-    """Syntax error in a coroutine that we yield"""
+@cocotb.test()
+def test_coroutine_error(dut):
+    """Error in a coroutine that we yield"""
     yield clock_gen(dut.clk)
-    yield syntax_error()
+    with assert_raises(NameError):
+        yield erroring_coro()
 
 
-@cocotb.test(expect_error=True)
-def test_fork_syntax_error(dut):
-    """Syntax error in a coroutine that we fork"""
+@cocotb.test(expect_error=NameError)
+def test_fork_error(dut):
+    """Error in a coroutine that we fork"""
     yield clock_gen(dut.clk)
-    cocotb.fork(syntax_error())
+    cocotb.fork(erroring_coro())
     yield clock_gen(dut.clk)
 
 
--- a/tests/test_cases/test_discovery/test_discovery.py
+++ b/tests/test_cases/test_discovery/test_discovery.py
@@ -68,7 +68,7 @@ async def discover_module_values(dut):
         raise TestFailure("Expected to discover things in the DUT")
 
 
-@cocotb.test(expect_error=True)
+@cocotb.test(expect_error=AttributeError)
 async def discover_value_not_in_dut(dut):
     """Try and get a value from the DUT that is not there"""
     fake_signal = dut.fake_signal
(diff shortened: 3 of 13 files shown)


cocotb/decorators.py
```diff
@@ -525,6 +525,9 @@ async def my_test(dut):
 
             .. versionchanged:: 1.3
                 Specific exception types can be expected
+
+            .. versionchanged:: 1.5
+                Giving ``True`` as an argument has been deprecated, use a specific Python Exception instead.
```
eric-wieser, line 530:
```suggestion
            .. deprecated:: 1.5
                Passing ``True`` is now deprecated, use a specific Python Exception instead.
```

cocotb/decorators.py
```diff
@@ -567,6 +570,8 @@ async def f(*args, **kwargs):
         self.timeout_unit = timeout_unit
         self.expect_fail = expect_fail
         if expect_error is True:
+            warnings.warn(
+                "`except_error=True` option to `cocotb.test` has been deprecated, specify an expected Python Exception instead")
```
eric-wieser, line 574:
```suggestion
                "`except_error=True` option to `cocotb.test` has been deprecated, specify an expected Python Exception instead",
                DeprecationWarning, stacklevel=2)
```

tests/test_cases/test_cocotb/test_generator_coroutines.py
```diff
@@ -96,14 +96,14 @@ def syntax_error():
     fail  # noqa
 
 
-@cocotb.test(expect_error=True)
+@cocotb.test(expect_error=NameError)
 def test_coroutine_syntax_error(dut):
     """Syntax error in a coroutine that we yield"""
     yield clock_gen(dut.clk)
```
eric-wieser, line 102: This one can be tested with `assert_raises`. Might also want to replace `syntax_error` with `name_error`.

cocotb/decorators.py
```diff
@@ -567,6 +570,8 @@ async def f(*args, **kwargs):
...
             expect_error = (Exception,)
         elif expect_error is False:
             expect_error = ()
```
eric-wieser, line 577: Should we deprecate `False` too?

cocotb/decorators.py
cmarqu, line 574:
```suggestion
                "`expect_error=True` option to `cocotb.test` has been deprecated, specify an expected Python Exception instead")
```
Nobody excepts the Spanish Expection!

cocotb/decorators.py
ktbarrett: It's currently used for spelling `expect_error=Exception if cond else False`. We could deprecate `False`, and replace it with `None`? So instead of `Union[bool, Exception]` it's `Optional[Exception]`, which is a more common interface type.

eric-wieser: We already allow `Exception if cond else ()`, which is shorter than `False`. `None` would be ok, but given the choice between `Union[Exception, Tuple[Exception], None]` and `Union[Exception, Tuple[Exception]]`, I'd lean towards the shorter choice.

cocotb/decorators.py
```diff
@@ -567,6 +570,9 @@ async def f(*args, **kwargs):
         self.timeout_unit = timeout_unit
         self.expect_fail = expect_fail
         if expect_error is True:
+            warnings.warn(
+                "Passing `expect_error=True` to `cocotb.test` is deprecated, specify a specific Exception type instead",
```
eric-wieser, line 574:
```suggestion
                "Passing `expect_error=True` to `cocotb.test` is deprecated, pass a specific Exception type instead",
```
Using "specific" twice here is odd.

cocotb/decorators.py
ktbarrett: I replaced all instances of `expect_error=False` and updated the default value and docstring.

cocotb/decorators.py
```diff
@@ -507,7 +507,7 @@ class test(coroutine, metaclass=_decorator_helper):
                 Using None as the the *timeout_unit* argument is deprecated, use ``'step'`` instead.
          expect_fail (bool, optional):
             Don't mark the result as a failure if the test fails.
-        expect_error (bool or exception type or tuple of exception types, optional):
+        expect_error (exception type or tuple of exception types, optional):
             If ``True``, consider this test passing if it raises *any* :class:`Exception`, and failing if it does not.
             If given an exception type or tuple of exception types, catching *only* a listed exception type is considered passing.
             This is primarily for cocotb internal regression use for when a simulator error is expected.
```
garmin-mjames, line 513:
```suggestion
        expect_error (exception type or tuple of exception types, optional):
            Mark the result as a pass only if one of the exception types is raised in the test.
            This is primarily for cocotb internal regression use for when a simulator error is expected.
```

garmin-mjames:
`test_readwrite_in_readonly` raises `TriggerException` in Riviera Verilog. VHDL is skipped, citing #1245, but I don't understand why it should be skipped.

tests/test_cases/test_cocotb/test_generator_coroutines.py
ktbarrett: I fixed this instance, unfortunately other instances that could otherwise be fixed are abuse the test end behavior of `cocotb.fork` and can't be caught; or aren't in `test_cocotb` and don't have access to `assert_raises`.

ktbarrett:
@garmin-mjames Probably because it causes a hard crash (or did at one point). There was another simulator that did that, but I don't remember which. We should be catching before we call into the simulator. I think Riviera is actually the only one with good behavior if what you said is the case.

ktbarrett:
@garmin-mjames I see why now. It has the same issue as Questa does. The test passes when it shouldn't. Passing isn't expected, we have broken behavior. Should we mark the test to pass anyways? I will rewrite that tests once I add in support for catching `ReadWrite` in `ReadOnly`.

garmin-mjames:
@ktbarrett For VPI, it's explicitly an error:
> 38.36.2 Simulation time callbacks
> - Attempting to place a **cbReadWriteSynch** callback with a delay of zero at read-only synch time.

For VHPI, the _time_ value in the callback structure is ignored, and registering a **vhpiCbRepLastKnownDeltaCycle** callback looks to be valid. So Riviera's behavior is correct, and we should expect the test to pass with VHDL currently.

ktbarrett:
Has a deprecation newsfragment and test, updated for `readwrite_in_readonly` for our current understanding. Ready for final review.

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -67,7 +67,7 @@ async def discover_module_values(dut):
         raise TestFailure("Expected to discover things in the DUT")
 
 
-@cocotb.test(expect_error=True)
+@cocotb.test(expect_error=AttributeError)
 async def discover_value_not_in_dut(dut):
     """Try and get a value from the DUT that is not there"""
     fake_signal = dut.fake_signal
```
eric-wieser, line 73:
```suggestion
@cocotb.test()
async def discover_value_not_in_dut(dut):
    """Try and get a value from the DUT that is not there"""
    with assert_raises(AttributeError):
        fake_signal = dut.fake_signal
```

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -133,7 +133,7 @@ async def access_single_bit_erroneous(dut):
     dut.stream_in_data[bit] <= 1
```
eric-wieser, line 133:
```suggestion
    with assert_raises(IndexError):
        dut.stream_in_data[bit] <= 1
```

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -124,7 +124,7 @@ async def access_single_bit_assignment(dut):
                          dut.stream_out_data_comb.value.integer, (1 << 2)))
 
 
-@cocotb.test(expect_error=True)
+@cocotb.test(expect_error=IndexError)
```
eric-wieser, line 127:
```suggestion
@cocotb.test()
```

tests/test_cases/test_discovery/test_discovery.py
ktbarrett: `assert_raises` isn't available here. Should I copy-paste it?

ktbarrett:
@eric-wieser I am removing the previous commit.

tests/test_cases/test_discovery/test_discovery.py
eric-wieser: Yeah, I think doing that's fine.

eric-wieser (approved):
I guess we can always cleanup more `expect_error` -> `assert_raises` in another PR.

ktbarrett:
@eric-wieser Yeah I had posted the comment and commit at the same time you had posted yours. I think we should find a way to put code common to all tests in a single place anyways.

Issue #1279, opened 2020-01-03
Regressions continue if expecting a SimFailure


ktbarrett:
```python
@cocotb.test(expect_error=cocotb.result.SimFailure)
async def test_sim_failure_a(dut):
    # invoke a deadlock
    await cocotb.triggers.RisingEdge(dut.clk)


@cocotb.test()
async def test_sim_failure_b(dut):
    raise TestFailure("This test should never run")
```

Because of the logic in `handle_result`, an expected `SimFailure` will continue to run tests and blow up in a strange and wonderful way.

PR #2236, opened 2020-11-27
Update some more tests for new syntax/conventions

ktbarrett:
Updated the issue-specific tests to use `async`/`await`, `assert`, and specify `units` in `Timer`. Passing specific exceptions to `expect_error` is #2223.

### DIFF
--- a/tests/test_cases/issue_1279/issue_1279.py
+++ b/tests/test_cases/issue_1279/issue_1279.py
@@ -9,9 +9,9 @@ import cocotb
     expect_error=cocotb.result.SimFailure,
     stage=1,
 )
-def test_sim_failure_a(dut):
+async def test_sim_failure_a(dut):
     # invoke a deadlock, as nothing is driving this clock
-    yield cocotb.triggers.RisingEdge(dut.clk)
+    await cocotb.triggers.RisingEdge(dut.clk)
 
 
 @cocotb.test(
@@ -19,6 +19,5 @@ def test_sim_failure_a(dut):
     expect_error=cocotb.result.SimFailure,
     stage=2,
 )
-def test_sim_failure_b(dut):
-    yield cocotb.triggers.NullTrigger()
-    raise cocotb.result.TestFailure("This test should never run")
+async def test_sim_failure_b(dut):
+    assert False, "This test should never run"
--- a/tests/test_cases/issue_348/issue_348.py
+++ b/tests/test_cases/issue_348/issue_348.py
@@ -1,27 +1,23 @@
 import cocotb
 from cocotb.log import SimLog
 from cocotb.triggers import Timer, Edge, RisingEdge, FallingEdge
-from cocotb.result import TestFailure
 
 
-@cocotb.coroutine
-def clock_gen(signal, num):
+async def clock_gen(signal, num):
     for x in range(num):
         signal <= 0
-        yield Timer(500)
+        await Timer(5, 'ns')
         signal <= 1
-        yield Timer(500)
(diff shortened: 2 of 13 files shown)


tests/test_cases/issue_1279/issue_1279.py
```diff
@@ -9,16 +9,15 @@
...
 @cocotb.test(
     skip=cocotb.SIM_NAME.lower().startswith("riviera"),  # gh-1859
     expect_error=cocotb.result.SimFailure,
     stage=2,
 )
-def test_sim_failure_b(dut):
-    yield cocotb.triggers.NullTrigger()
-    raise cocotb.result.TestFailure("This test should never run")
+async def test_sim_failure_b(dut):
+    assert False, "This test should never run"
```
eric-wieser, line 23: Are we sure the null trigger was not an important part of this test?

tests/test_cases/issue_348/issue_348.py
```diff
@@ -1,27 +1,23 @@
...
 
 
-@cocotb.coroutine
-def signal_mon(signal, idx, edge):
-    log = SimLog("cocotb.signal_mon.%d.%s" % (idx, signal._name))
-    value = signal.value
-
+async def signal_mon(signal, idx, edge):
+    _ = SimLog("cocotb.signal_mon.%d.%s" % (idx, signal._name))
+    _ = signal.value
```
eric-wieser, line 16: Why the `_`? Should we just delete these lines?

tests/test_cases/issue_348/issue_348.py
```diff
@@ -34,43 +30,40 @@ def __init__(self, edge, signal):
...
-    @cocotb.coroutine
-    def start(self):
+    async def start(self):
         clock_edges = 10
 
         cocotb.fork(clock_gen(self.signal, clock_edges))
-        first = cocotb.fork(self.signal_mon(self.signal, 0, self.edge_type))
-        second = cocotb.fork(self.signal_mon(self.signal, 1, self.edge_type))
+        _ = cocotb.fork(self.signal_mon(self.signal, 0, self.edge_type))
+        _ = cocotb.fork(self.signal_mon(self.signal, 1, self.edge_type))
```
eric-wieser, line 43: Should we just drop the assignment here?

tests/test_cases/issue_1279/issue_1279.py
ktbarrett: #1279 is about the next test running at all. So I don't think it was.

tests/test_cases/issue_348/issue_348.py
ktbarrett: I had intended to come back to this, but forgot. Yes, it's dead code. I'll remove.
```suggestion
```

tests/test_cases/issue_348/issue_348.py
ktbarrett:
```suggestion
        cocotb.fork(self.signal_mon(self.signal, 0, self.edge_type))
        cocotb.fork(self.signal_mon(self.signal, 1, self.edge_type))
```

cmarqu:
There are some `Clock()` and `Timer()` without `units` left, did you keep them unchanged on purpose?

ktbarrett:
@cmarqu The description said I handled them, but I don't think I actually did that work when I made the changes. Whoops... I addressed that now, manual scanning is not the greatest and it doesn't warn.

cmarqu (approved):

PR #2237, opened 2020-11-28
Remove test_doctests and use 'pytest --doctest-modules' instead

### LINKED ISSUE #1988: Run doctests with pytest directly
We could replace `tests/test_cases/test_cocotb/test_doctests.py` by using `pytest --doctest-modules` instead of just `pytest` in `tox.ini`.

ktbarrett:
Closes #1988.


cmarqu (approved):

eric-wieser (approved):
I think we didn't have pytest running at the point I wrote these

PR #2238, opened 2020-11-29
Support waveform dumping in verilator via WAVES environment variable

alexforencich:
Similar to other simulators, support waveform dumping in verilator via WAVES environment variable

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.verilator
+++ b/cocotb/share/makefiles/simulators/Makefile.verilator
@@ -29,7 +29,11 @@ ifeq ($(VERILATOR_SIM_DEBUG), 1)
 endif
 
 ifeq ($(VERILATOR_TRACE),1)
-  EXTRA_ARGS += --trace --trace-structs
+  WAVES = 1
+endif
+
+ifeq ($(WAVES),1)
+  EXTRA_ARGS += --trace-fst --trace-structs
 endif
 
 ifdef COCOTB_HDL_TIMEPRECISION


alexforencich:
Actually, this one more or less duplicates `VERILATOR_TRACE`.  I will rework it so specifying either `VERILATOR_TRACE` or `WAVES` will work, and remove the redundancy.

themperek (commented):
Looks OK. Could you also change documentation: https://github.com/cocotb/cocotb/blob/master/documentation/source/building.rst#makefile-based-test-scripts ?

ktbarrett:
Rerunning regressions.

ktbarrett (commented):
Just like the other issue, I think we need to be able to support both VCD and FST formats.

cocotb/share/makefiles/simulators/Makefile.verilator
```diff
@@ -29,7 +29,11 @@ ifeq ($(VERILATOR_SIM_DEBUG), 1)
 endif
 
 ifeq ($(VERILATOR_TRACE),1)
-  EXTRA_ARGS += --trace --trace-structs
+  WAVES = 1
```
ktbarrett, line 32: One additional action here is to add `VERILATOR_TRACE` to `Makefile.deprecations` if it isn't already.

WernerFS: Could `--trace-fst` be skipped? Although it is a great compressed format, it is only supported by gtkwave for viewing and Verilator for generation, as far as I know. Sticking with VCD would make things easier I think.

ktbarrett: Surfer might support it?

WernerFS: There is also another open-source project looking to bring an FST recording backend for other simulators but seems a bit far in the future. I just don't think it is ready to be the default option. But those are my 2 cents.

imphil: Surfer supports FST, just like gtkwave. So for those people using open source tools, both the producer and the consumer side are covered well. FST is significantly better, both in terms of features (type representation for enums, for example), as well as in compression (which also means less data to write to disk and hence faster simulations). Anyone using open source tools is likely to want FST. VCD is really just a safe fallback for interop.

WernerFS:
Now `waves` is handled at runner.py specifically: [13f8545](https://github.com/cocotb/cocotb/commit/13f8545930201250e2dc7a95ec9b165571482bfc). 

This only passes --trace to Verilator probably we would want to pass --trace-fst? If that is not the case then this merge is no longer needed and [3894](https://github.com/cocotb/cocotb/issues/3894) can be closed.

ktbarrett:
Thanks for the bump. Yeah I need to clean up old PRs at some point.

ktbarrett:
So since we have to support both FST and VCD outputs and that requires passing different arguments during build (ugh) we have documentation on how to do this specially with Verilator using the Makefiles and that's as good as it's going to get. Makefiles are going to receive no new features in 2.x and will be deprecated at some point.

@WernerFS As for the runner, we should probably work on that. Maybe `waves=true` defaults to VCD for backwards compat, but we can also pass `"vcd"`, `"fst"`, or add a waves file argument and have it read the extension.

PR #2239, opened 2020-11-29
Support waveform dumping in Icarus Verilog via WAVES environment variable

alexforencich:
Similar to other simulators, support waveform dumping in Icarus Verilog via WAVES environment variable

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.icarus
+++ b/cocotb/share/makefiles/simulators/Makefile.icarus
@@ -59,9 +59,24 @@ endif
 COMPILE_ARGS += -f $(SIM_BUILD)/cmds.f -g2012 # Default to latest SystemVerilog standard
 
 # Compilation phase
+
+ifeq ($(WAVES), 1)
+    VERILOG_SOURCES += $(SIM_BUILD)/iverilog_dump.v
+    COMPILE_ARGS += -s iverilog_dump
+    PLUSARGS += -fst
+endif
+
 $(SIM_BUILD)/sim.vvp: $(VERILOG_SOURCES) $(CUSTOM_COMPILE_DEPS) | $(SIM_BUILD)
 	@echo "+timescale+$(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION)" > $(SIM_BUILD)/cmds.f
 	$(CMD) -o $(SIM_BUILD)/sim.vvp -D COCOTB_SIM=1 -s $(TOPLEVEL) $(COMPILE_ARGS) $(EXTRA_ARGS) $(VERILOG_SOURCES)
+
+$(SIM_BUILD)/iverilog_dump.v: | $(SIM_BUILD)
+	@echo 'module iverilog_dump();' > $@
+	@echo 'initial begin' >> $@
+	@echo '    $$dumpfile("$(TOPLEVEL).fst");' >> $@
+	@echo '    $$dumpvars(0, $(TOPLEVEL));' >> $@
+	@echo 'end' >> $@
+	@echo 'endmodule' >> $@
 
 # Execution phase
 


ktbarrett:
@alexforencich You have two nearly identical PRs?

themperek:
This probably needs also documentation update and news fragment.

Just a comment/question:
Using `TOPLELVEL.fst` as the dump file name makes sense. Maybe we should do the same for Verilator? Maybe even allow setting `WAVES=XYZ.fst`?

alexforencich:
That's an interesting idea.  However, I'm not sure if all of the simulators support arbitrary dump file names.  I think at least verilator can only write out "dump.vcd" or "dump.fst".  

I will update the docs and add a news fragment.  

Apparently there was also some discussion on possibly deprecating `WAVES` (#1495).  I think it is super useful to have a standard way of just turning on some form of waveform dumps in most simulators, and supporting a minimal "dump to file" scheme would be extremely convenient.  Anything more complex and simulator-specific can then be done manually if desired.  If renaming `WAVES` to something more descriptive makes sense, then we can do that.  Deprecating `COCOTB_NVC_TRACE` and `VERILATOR_TRACE` make sense as these are simulator-specific, but instead of outright removing those, the functionality should be rolled in to `WAVES`, or a replacement for `WAVES`.  Now, it might make sense to add one or two standard dump configuration options, such as file name, file type, and possibly module name (module name is useful at least for icarus verilog as it has to add a top-level module with the verilog commands to enable waveform dumping, and changing the module name may be necessary if there is a conflict).

ktbarrett:
Rerunning regressions.

cocotb/share/makefiles/simulators/Makefile.icarus
```diff
@@ -59,10 +59,25 @@ endif
...
+endif
+
 $(SIM_BUILD)/sim.vvp: $(VERILOG_SOURCES) $(CUSTOM_COMPILE_DEPS) | $(SIM_BUILD)
 	@echo "+timescale+$(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION)" > $(SIM_BUILD)/cmds.f
 	$(CMD) -o $(SIM_BUILD)/sim.vvp -D COCOTB_SIM=1 -s $(TOPLEVEL) $(COMPILE_ARGS) $(EXTRA_ARGS) $(VERILOG_SOURCES)
 
+$(SIM_BUILD)/iverilog_dump.v: | $(SIM_BUILD)
+	@echo 'module iverilog_dump();' > $@
+	@echo 'initial begin' >> $@
+	@echo '    $$dumpfile("$(TOPLEVEL).fst");' >> $@
```
ktbarrett, line 76: I think we would want to support both FST and VCD outputs. Is there currently a mechanism to select one or the other? I know Verilator has a similar problem.

ktbarrett (commented):
I agree that having a standard way to turn on dumping does help beginners greatly. I think `WAVES` should remain and be expanded to at least cover all FOSS simulators. Unfortunately I don't know anything about it, and which simulators are currently support and which aren't.

alexforencich:
TBH, I don't care if we extend WAVES or change to a different env var so long as whatever we do is consistent.  I like the idea of having multiple variables, though - one for enable/disable, one for the name, one for the type, etc.  I think this would be more flexible than simply providing a dump file name, like some people have suggested.  Especially when simulators support different combinations of dump formats.  So WAVES=0/1, WAVES_TYPE=vcd,fst,lxt,... WAVES_NAME=file_name_without_extension .  And default would be all unset/blank, and then the makefile can fill in sensible defaults (fst type, top module as name, etc.).  Or something along those lines.

ktbarrett:
Having `WAVES` as a separate variables does make turning it on and off easier. `WAVES_NAME` and `WAVES_FMT` could exist in the Makefile and even have defaults in the `Makefile.inc` and the user could issue `make WAVES=1` to enable dumping without additional configuration.

alexforencich:
Exactly.  The only question is do we keep WAVES as the prefix, or do we use something else?  Only reason I used WAVES is that a couple of other makefiles use that; I originally thought the simulator was looking at that but it turns out that it's actually handled in the makefiles.  It's also trivial to map WAVES onto a new name for deprecation purposes.  And passing this stuff through to verilator is IIRC only dependent on adding some command line argument parsing capability to the verilator C++ file, which can pass the setting along to verilator.  Yes, sure, you can also do it with a define instead of a command line arg, but if you change the value then you may need to do a make clean before it will take effect.

ktbarrett:
I think the prefix `WAVES` is fine.

cmarqu:
Renaming WAVES would need a deprecation/aliasing to the new name for the simulators currently supporting it. I'm pro renaming, but I don't mind so much since I don't use it.
And I don't use this because I would need a pretty elaborate support for it to be useful for Xcelium which would be better implemented as a variable for a "GUI setup config file" that's then loaded, but this loading I can just do with `SIM_ARGS`.

themperek:
So I see 2 options:
 - `WAVES`=`0/1`, `WAVES_TYPE`=`vcd,fst,lxt,...` `WAVES_NAME`=`file_name_without_extension`
 - `WAVES`=`file_name`.`vcd,fst,lxt,...` or `0/1`

ddribin:
This would be very useful to have! The changes here should basically still work on top of current `master`. Any chance this PR could be updated? I'm also happy to to open a new PR, too, @alexforencich, if you are okay to hand it off.

As far as VCD vs. FST and new environment variables. While that seems like a good long-term goal, what do you think about  landing this simple change to get `WAVES` working with Icarus and then following up with a separate PR/issue for those changes?

PR #2240, opened 2020-11-29
Deprecate assigning dict object to signals

ktbarrett:
This custom format `dict` value is not documented anywhere, and it's an operation that can be easily done by the user. The format was of the form `dict(values=[0, 1, 2], bits=4)`, and represents `values` as digits of bit size `bits`.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -729,6 +729,11 @@ class ModifiableObject(NonConstantObject):
             :class:`ctypes.Structure` objects are no longer accepted as values for assignment.
             Convert the struct object to a :class:`~cocotb.binary.BinaryValue` before assignment using
             ``BinaryValue(value=bytes(struct_obj), n_bits=len(signal))`` instead.
+
+        .. deprecated:: 1.5
+            :class:`dict` objects are no longer accepted as values for assignment.
+            Convert the dictionary to an integer before assignment using
+            ``sum(v << (d['bits'] * i) for i, v in enumerate(d['values']))``.
         """
         value, set_action = self._check_for_set_action(value)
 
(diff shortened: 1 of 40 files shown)


eric-wieser:
I don't find the replacement particularly compelling - I'd be inclined to leave this behavior in, and try to design a replacement as part of the new `BinaryValue`. In this case, it seems what's needed is a way to view an `m*n`-bit signal as an array of `m`-bit signals.

ktbarrett:
@eric-wieser If we plan on replacing it with something else in the next version, this interface will break, so this interface should be deprecated.

ktbarrett:
As for a proper replacement, this could be a feature of the `Vector` class, where vectors of vectors of Logic values can be flattened. This behavior doesn't really belong in the assignment operator, which should be kept as simple as possible.

eric-wieser:
Perhaps a compromise is to introduce in `cocotb.utils`:
```python
# or a better name
def pack_bit_vector(values : List[int], *, bits : int):
    """ Pack the integers in `values` into a single integer, with each entry occupying `bits` bits. """
    return sum(v << (bits * i) for i, v in enumerate(values))
```
which means:

* We can deprecate the weird dict thing now, and recommend something short but less surprising
* We can deprecate `pack_bit_vector` in 2.0 when we gut binaryvalue.

cocotb/handle.py
```diff
@@ -698,6 +698,11 @@ def _set_value(self, value, call_sim):
         Raises:
             TypeError: If target is not wide enough or has an unsupported type
                  for value assignment.
+
+        .. deprecated:: 1.5
+            :class:`dict` objects are no longer accepted as values for assignment.
+            Convert the dictionary to an integer before assignment using
+            ``sum(v << (d['bits'] * i) for i, v in enumerate(d['values']))``.
```
eric-wieser, line 705: This implementation isn't the same as the one we are deprecating, it reverses the entries!

ktbarrett: That's due to how it's building the integer. I got tricked by using `reversed` at first. The test shows it's correct. But I will do what you had mentioned above.

eric-wieser (commented):
The suggested replacement is less safe: `if len(vallist) * value["bits"] != len(self):` is present in the original codepath.

ktbarrett:
@eric-wieser So is the suggestion to document an assert in the recommended replacement, or to flip this change on it's head and document the interface since it's potentially useful? I don't particularly care. It's undocumented, so I think it can be removed with prejudice. Even if we keep it, it will live only to be deprecated once we have proper logic vector  value assignment support.

PR #2247, opened 2020-12-01
Don't try to load libpython.so dynamically on Mac

### LINKED ISSUE #2246: Cannot run on Big Sur
Cocotb cannot find libpytyon3.8.so  

I do not see it in the libs:

```
(base) raysalemi@WriteNow lib % ls libpython*
libpython3.8.a		libpython3.8.dylib	libpython3.8.nolto.a
```

Am I doing something wrong?  

* Latest version of Cocotb (just downloaded)
* MacOS 11.0.1
* icarus-verilog 11.0
* anaconda python 3.8.3
* log file:
```
MODULE=tinyalu_cocotb TESTCASE= TOPLEVEL=tinyalu TOPLEVEL_LANG=verilog \
        /usr/local/bin/vvp -M /Users/raysalemi/opt/anaconda3/lib/python3.8/site-packages/cocotb/libs -m libcocotbvpi_icarus   sim_build/sim.vvp
     -.--ns ERROR    cocotb.gpi                         ..s/cocotb_utils.cpp:69   in utils_dyn_open                  Unable to open lib libpython3.8.so: dlopen(libpython3.8.so, 9): image not found

     -.--ns ERROR    cocotb.gpi                         ..mbed/gpi_embed.cpp:127  in embed_init_python               Failed to find Python shared library

```

ktbarrett:
Closes #2246. Prevents an invalid error message. This is not needed now as far as we can tell.

Should investigate doing the same for Windows, but I will do some testing first.

Previously: https://github.com/cocotb/cocotb/runs/1477280581?check_suite_focus=true#step:22:100
Now: https://github.com/cocotb/cocotb/pull/2247/checks?check_run_id=1480568336#step:22:100

### DIFF
--- a/cocotb/share/include/cocotb_utils.h
+++ b/cocotb/share/include/cocotb_utils.h
@@ -39,15 +39,17 @@
 
 #include <gpi_logging.h>
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 #define xstr(a) str(a)
 #define str(a) #a
 
-extern COCOTBUTILS_EXPORT void* utils_dyn_open(const char* lib_name);
-extern COCOTBUTILS_EXPORT void* utils_dyn_sym(void *handle, const char* sym_name);
+COCOTBUTILS_EXPORT void* utils_dyn_open(
+    const char* lib_name,
+    enum gpi_log_levels error_log_level = GPIError);
+
+COCOTBUTILS_EXPORT void* utils_dyn_sym(
+    void *handle,
+    const char* sym_name,
+    enum gpi_log_levels error_log_level = GPIError);
 
 extern COCOTBUTILS_EXPORT int is_python_context;
 
@@ -73,9 +75,5 @@ extern COCOTBUTILS_EXPORT int is_python_context;
 } while (0)
 
 #define COCOTB_UNUSED(x) ((void)x)
-
-#ifdef __cplusplus
-}
-#endif
 
 #endif /* COCOTB_UTILS_H_ */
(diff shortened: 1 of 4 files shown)


themperek:
Do we understand this: https://github.com/cocotb/cocotb/pull/1417 ?
I guess on mac only Icarus/Verilator is used?

themperek:
@ktbarrett After some thinking maybe better do an `ifndef` for `ICARUS` that we know will work?

ktbarrett:
@themperek I just tried removing that section, like was done in #1416 and received the same error. https://github.com/ContinuumIO/anaconda-issues/issues/6401 could be the issue. When we `dlopen` libpython, we do so with the `RTLD_GLOBAL` flag. I'm not sure how libraries are loaded when they are loaded by the loader as a dependency in Linux.

ktbarrett:
We know the correct path forward is to runtime-load libpython, which is the opposite of the current change. I wonder if any commerical sims (that run on Mac) can run cocotb? The big issue is that we are getting the name wrong. On Mac, libpython ends in `.dylib`. Perhaps that should be the actual change?

themperek:
Using `dylib` extension on mac does not seem to help: https://github.com/cocotb/cocotb/runs/1485641312 (Not mac user)

Maybe we should just make it a single WARNING instead of 2 ERRORS (It is confusing now)?

ktbarrett:
We can definitely just remove the error from the calling code and leave the one in `utils_dyn_open`. And it would be nice if we could diminish the error to a warning.

ktbarrett:
Well... actually... on some platforms with some version of Python it is an error, which will become immediately apparent. We know this isn't the right answer, do we even keep trying?

ktbarrett:
Closing in favor of #2263.

PR #2249, opened 2020-12-01
Check BigSur 

themperek:


themperek:
So it seems to work under CI: https://github.com/cocotb/cocotb/pull/2249/checks?check_run_id=1481370246

(The overall status if fail but this is other/setup issues).

Issue #2253, opened 2020-12-03
Documentation warning about sphinxcontrib-makedomain


ktbarrett:
We use [`sphinxcontrib-makedomain`](https://github.com/sphinx-contrib/makedomain) for documenting our make variables. However the project hasn't been maintained in 7 years and is not up to date, so it spits out the following warnings:

```
/home/kaleb/dev/cocotb/documentation/source/building.rst:: WARNING: 4 column based index found. It might be a bug of extensions you use: [('pair', 'GUI; Make Variable', 'var-GUI', '')]
/home/kaleb/dev/cocotb/documentation/source/building.rst:: WARNING: 4 column based index found. It might be a bug of extensions you use: [('pair', 'SIM; Make Variable', 'var-SIM', '')]
/home/kaleb/dev/cocotb/documentation/source/building.rst:: WARNING: 4 column based index found. It might be a bug of extensions you use: [('pair', 'WAVES; Make Variable', 'var-WAVES', '')]
/home/kaleb/dev/cocotb/documentation/source/building.rst:: WARNING: 4 column based index found. It might be a bug of extensions you use: [('pair', 'TOPLEVEL_LANG; Make Variable', 'var-TOPLEVEL_LANG', '')]
...
```

And so on for every make variable. Per [this comment](https://github.com/sphinx-doc/sphinx/issues/2673#issuecomment-226076269), 4 column based indexes has been deprecated since forever, but since these were buried in the million other warnings they were missed.

ktbarrett:
Closed by #3352.

PR #2254, opened 2020-12-03
Documentation cleanup

ktbarrett:
Removed the call to `breath-apidoc` in `conf.py` and instead used a breath directive to pull in the generated documentation. 
Limited C/C++ documentation to only the GPI (the only thing the user should be touching). This can be expanded in the future for internal documentation purposes. Also changed several doxygen and breathe options to fix/prevent warning messages.

The only warning remaining I see is #2253. The documentation build also runs much quicker.


cocotb/scheduler.py
```diff
@@ -184,11 +184,11 @@ class Scheduler:
         - Any pending writes are cached and do not happen immediately
 
     ReadOnly mode
-        - Corresponds to :any:`cbReadOnlySynch` (VPI) or :any:`vhpiCbLastKnownDeltaCycle`
+        - Corresponds to ``cbReadOnlySynch`` (VPI) or ``vhpiCbLastKnownDeltaCycle``
```
garmin-mjames, line 187: Update these to the correct callbacks while you're here.
```suggestion
        - Corresponds to ``cbReadOnlySynch`` (VPI) or ``vhpiCbRepEndOfTimeStep``
```

cocotb/scheduler.py
```diff
@@ -184,11 +184,11 @@ class Scheduler:
...
           (VHPI).  In this state we are not allowed to perform writes.
 
     Write mode
-        - Corresponds to :any:`cbReadWriteSynch` (VPI) or :c:macro:`vhpiCbEndOfProcesses` (VHPI)
+        - Corresponds to ``cbReadWriteSynch`` (VPI) or ``vhpiCbEndOfProcesses`` (VHPI)
```
garmin-mjames, line 191:
```suggestion
        - Corresponds to ``cbReadWriteSynch`` (VPI) or ``vhpiCbRepLastKnownDeltaCycle`` (VHPI)
```

cmarqu (approved):
I like it, thanks!
It has now even become feasible to add Doxygen strings to the elements we list there before the release.

PR #2258, opened 2020-12-03
Add tests for freeze and combinations with struct ports

cmarqu:


ktbarrett:
@cmarqu This looks like a good change. Any reason to rebase and see what happens?

PR #2265, opened 2020-12-04
Remove erroneous assignment to input port

### LINKED ISSUE #2262: Riviera failing because of #2256 ?

cmarqu:
Interestingly, both Xcelium and Questa allow that
while Verilator and Riviera show an error.

Closes #2262.


garmin-mjames (approved):
Verilator and Riviera are compiling `sample_module.sv` now.

PR #2266, opened 2020-12-04
Try inout struct

cmarqu:


cmarqu:
Ok, this alone doesn't cut it.
```
  %Error: Internal Error: /home/runner/work/cocotb/cocotb/tests/test_cases/test_array_simple/../../../tests/designs/sample_module/sample_module.sv:54:49: ../V3Gate.cpp:1099: Consumer doesn't match lhs of assign
     54 |     inout  test_if                              inout_if,
        |                                                 ^~~~~~~~
                          ... See the manual and https://verilator.org for more assistance
```

Issue #1945, opened 2020-06-25
PYTHONHOME issue?


juliusbaxter:
I had cocotb running in our work environment only last week or so, but something has started to cause the following issue.

I'm on CentOS 7 and the `SIM=xcelium`

The `simple_dff` example appears to die the same way, so here's the output from that:

```
        module worklib.dff:sv
                errors: 0, warnings: 0
     -.--ns INFO     cocotb.gpi                         ..mbed/gpi_embed.cpp:74   in set_program_name_in_venv        Did not detect Python virtual environment. Using system-wide Python interpreter
Fatal Python error: Py_Initialize: Unable to get the locale encoding
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00007fa44e62ee00 (most recent call first):
ImportError: No module named site
xmelab: *F,INTERR: INTERNAL EXCEPTION
```
Some output from `cocotb-config`:
```
$ cocotb-config --version
1.4.0rc1
$ cocotb-config --python-bin
/usr/bin/python3
$ cocotb-config --prefix
/usr/local/lib64/python3.6/site-packages
```
Running the following (as per `Makefile.inc`'s way of calculating `PYTHONHOME`):
```
$ /usr/bin/python3 -c 'from distutils.sysconfig import get_config_var; print(get_config_var("prefix"))'
/usr
```

An interesting observation (I think anyway?) is if I set `PYTHONHOME` to `/usr/local` under which the `site-packages/cocotb` is really located (as per the `cocotb-config` output), I see the same crash.
```
$ PYTHONHOME=/usr /usr/bin/python3
Python 3.6.8 (default, Apr  2 2020, 13:34:55)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-39)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>

$ PYTHONHOME=/usr/local /usr/bin/python3
Fatal Python error: Py_Initialize: Unable to get the locale encoding
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00007f4ce0135740 (most recent call first):
Aborted
```

Any ideas? I'm quite certain this worked only a week ago, I just updated to the latest cocotb, but I'm not ruling out something changing on my system, too.

Thanks for any pointers, and great job with cocotb so far all!!

mtikekar:
[...]

PR #2268, opened 2020-12-05
Runtime-load libpython and make binaries relocatable

ktbarrett:
Builds on #2000. 

## Break load-time dependency on libcocotb and load libpython at runtime

libcocotb A.K.A gpi_embed calls Python functions. To support Windows and some security-hardened Linuxes, these functions must be resolved by the time libcocotb is finished loading. However, libpython is attempted to be loaded *in* libcocotb. This works on most Linuxes, but not all, and definitely not Windows. To prevent this issue, we link libpython as a dependency.

This is less than ideal since some simulators monkey with LD_LIBRARY_PATH and the wrong libpython could be loaded by the simulator on first load. Setting the RPATH is an effective work-around, but makes the binaries not relocatable. The best solution is to load libpython at runtime using the absolute path to the libpython associated with the cocotb install.

To prevent libpython being loaded as a dependency when the simulator first loads the PLI object, we must break the link dependency between the GPI and libcocotb. To accomplish that we create a small stub library to fill in for the "embed" library. This stub library runtime-loads libpython using an absolute path to the library. The stub library then runtime-loads the "implementation" embed library and proxies calls to it. This forces the correct libpython to be loaded, and all dependent libraries to load strictly after that.

## Find libpython in makefiles and load in embed.cpp

When we feed the libpython path to the embed stub library at build time, we bind the binaries to the environment permanently. This means the binaries are not relocatable, which is not good for wheel reuse or for binary distribution in general.

So instead we obtain the path to libpython when the makefiles run, and set the path in an environment variable that is used by the embed library stub. We use the find_libpython library to locate the libpython associated with the cocotb installation.

## TODO
- [x] Add `cocotb-config --libpython`
- [x] Fail more gracefully in embed if `LIBPYTHON_LOC` is not set.
- [x] Fix all uses of ANSI-encoding functions in Win API code. Use the Unicode version and translate to UTF-8 before passing to Python (for instance `LOG_*`).
- [x] Work code reviews from #2000.

## Future Work

- ~~The final step is to prevent Python distutils < 3.8 from linking all `Extensions` against libpython. This can probably be accomplished using #1739.~~ May not be necessary.
- With this merged, we can producing and distributing relocatable wheels using the `manylinux2014` spec. Ideally this can be done automatically in CI.

### DIFF
--- a/cocotb/share/include/gpi_logging.h
+++ b/cocotb/share/include/gpi_logging.h
@@ -37,39 +37,180 @@
 #define GPILOG_EXPORT COCOTB_IMPORT
 #endif
 
+#include <cstdarg>
+
 #ifdef __cplusplus
-# define EXTERN_C_START extern "C" {
-# define EXTERN_C_END }
-#else
-# define EXTERN_C_START
-# define EXTERN_C_END
+extern "C" {
 #endif
 
-EXTERN_C_START
-
+/** Named logging level
+ *
+ *  The native logger only logs level names at these log level values.
+ *  They were specifically chosen to align with the default level values in the Python logging module.
+ *  Implementers of custom loggers should emit human readable level names for these value, but may support other values
+ */
 enum gpi_log_levels {
-    GPIDebug = 10,
-    GPIInfo = 20,
-    GPIWarning = 30,
-    GPIError = 40,
-    GPICritical = 50
+    GPIDebug    = 10,   ///< Prints `DEBUG` by default. Verbose information, useful for debugging
+    GPIInfo     = 20,   ///< Prints `INFO` by default. Information about major events in the current program
+    GPIWarning  = 30,   ///< Prints `WARN` by default. Encountered a recoverable bug, or information about surprising behavior
+    GPIError    = 40,   ///< Prints `ERROR` by default. An unrecoverable error
+    GPICritical = 50    ///< Prints `CRITICAL` by default. An unrecoverable error, to be followed by immediate simulator shutdown
 };
 
+/** Logs a message at the given log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+    @param level The level at which to log the message
+ */
+#define LOG_(level, ...) \
+    gpi_log("cocotb.gpi", level,      __FILE__, __func__, __LINE__, __VA_ARGS__);
 
-#define LOG_DEBUG(...)     gpi_log("cocotb.gpi", GPIDebug,         __FILE__, __func__, __LINE__, __VA_ARGS__);
-#define LOG_INFO(...)      gpi_log("cocotb.gpi", GPIInfo,          __FILE__, __func__, __LINE__, __VA_ARGS__);
-#define LOG_WARN(...)      gpi_log("cocotb.gpi", GPIWarning,       __FILE__, __func__, __LINE__, __VA_ARGS__);
-#define LOG_ERROR(...)     gpi_log("cocotb.gpi", GPIError,         __FILE__, __func__, __LINE__, __VA_ARGS__);
-#define LOG_CRITICAL(...)  gpi_log("cocotb.gpi", GPICritical,      __FILE__, __func__, __LINE__, __VA_ARGS__);
+/** Logs a message at DEBUG log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
+#define LOG_DEBUG(...) LOG_(GPIDebug, __VA_ARGS__)
 
-GPILOG_EXPORT void set_log_handler(void *handler);
-GPILOG_EXPORT void clear_log_handler(void);
-GPILOG_EXPORT void set_log_filter(void *filter);
-GPILOG_EXPORT void clear_log_filter(void);
-GPILOG_EXPORT void set_log_level(enum gpi_log_levels new_level);
+/** Logs a message at INFO log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
+#define LOG_INFO(...) LOG_(GPIInfo, __VA_ARGS__);
 
-GPILOG_EXPORT void gpi_log(const char *name, enum gpi_log_levels level, const char *pathname, const char *funcname, long lineno, const char *msg, ...);
+/** Logs a message at WARN log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
+#define LOG_WARN(...) LOG_(GPIWarning, __VA_ARGS__);
 
-EXTERN_C_END
+/** Logs a message at ERROR log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
+#define LOG_ERROR(...) LOG_(GPIError, __VA_ARGS__);
+
+/** Logs a message at CRITICAL log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
+#define LOG_CRITICAL(...) LOG_(GPICritical, __VA_ARGS__);
+
+/** Type of a log handler function.
+    @param userdata  private implementation data registered with this function
+    @param name      Name of the logger
+    @param level     Level at which to log the message
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-sprintf-style format specifier
+    @param args      Additional arguments; formatted and inserted in message according to format specifier in msg argument
+ */
+typedef void (gpi_log_handler_type)(
+    void *userdata,
+    const char *name,
+    int level,
+    const char *pathname,
+    const char *funcname,
+    long lineno,
+    const char *msg,
+    va_list args);
+
+/** Log a message using the currently registered log handler.
+    User is expected to populate all arguments to this function.
+    @param name      Name of the logger
+    @param level     Level at which to log the message
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-sprintf-style format specifier
+    @param ...       Additional arguments; formatted and inserted in message according to format specifier in msg argument
+ */
+GPILOG_EXPORT void gpi_log(
+    const char *name,
+    int level,
+    const char *pathname,
+    const char *funcname,
+    long lineno,
+    const char *msg,
+    ...);
+
+/** Log a message using the currently registered log handler.
+    User is expected to populate all arguments to this function.
+    @param name      Name of the logger
+    @param level     Level at which to log the message
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-sprintf-style format specifier
+    @param args      Additional arguments; formatted and inserted in message according to format specifier in msg argument
+ */
+GPILOG_EXPORT void gpi_vlog(
+    const char *name,
+    int level,
+    const char *pathname,
+    const char *funcname,
+    long lineno,
+    const char *msg,
+    va_list args);
+
+/** Retrieve the current log handler.
+    @param handler  Location to return current log handler. If no custom logger is registered this will be `NULL`.
+    @param userdata Location to return log handler userdata
+ */
+GPILOG_EXPORT void gpi_get_log_handler(gpi_log_handler_type **handler, void **userdata);
+
+/** Set custom log handler
+    @param handler   Handler function to call when the GPI logs a message
+    @param userdata  Data to pass to the handler function when logging a message
+ */
+GPILOG_EXPORT void gpi_set_log_handler(gpi_log_handler_type *handler, void *userdata);
+
+/** Clear the current custom log handler and use native logger
+ */
+GPILOG_EXPORT void gpi_clear_log_handler(void);
+
+/** Log a message using the native log handler.
+    User is expected to populate all arguments to this function.
+    @param name      Name of the logger
+    @param level     Level at which to log the message
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-sprintf-style format specifier
+    @param ...       Additional arguments; formatted and inserted in message according to format specifier in msg argument
+ */
+GPILOG_EXPORT void gpi_native_logger_log(
+    const char *name,
+    int level,
+    const char *pathname,
+    const char *funcname,
+    long lineno,
+    const char *msg,
+    ...);
+
+/** Log a message using the native log handler.
+    User is expected to populate all arguments to this function.
+    @param name      Name of the logger
+    @param level     Level at which to log the message
+    @param pathname  Name of the file where the call site is located
+    @param funcname  Name of the function where the call site is located
+    @param lineno    Line number of the call site
+    @param msg       The message to log, uses C-sprintf-style format specifier
+    @param args      Additional arguments; formatted and inserted in message according to format specifier in msg argument
+ */
+GPILOG_EXPORT void gpi_native_logger_vlog(
+    const char *name,
+    int level,
+    const char *pathname,
+    const char *funcname,
+    long lineno,
+    const char *msg,
+    va_list args);
+
+/** Set minimum logging level of the native logger.
+    If a logging request occurs where the logging level is lower than the level set by this function, it is not logged.
+    Only affects the native logger.
+    @param level     Logging level
+    @return          Previous logging level
+ */
+GPILOG_EXPORT int gpi_native_logger_set_level(int level);
+
+#ifdef __cplusplus
+}
+#endif
 
 #endif /* COCOTB_GPI_LOGGING_H_ */
--- a/cocotb/share/lib/embed/embed.cpp
+++ b/cocotb/share/lib/embed/embed.cpp
@@ -1,0 +1,160 @@
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <embed.h>
+#include <cocotb_utils.h>   // xstr, utils_dyn_open, utils_dyn_sym
+#include <gpi.h>            // gpi_event_t
+#include <cstdlib>          // getenv
+#if ! defined(__linux__) && ! defined(__APPLE__)
+#include <windows.h>        // Win32 API for loading the embed impl library
+#include <string>           // string
+#endif
+
+#ifndef PYTHON_LIB
+#error "Name of Python library required"
+#else
+#define PYTHON_LIB_STR xstr(PYTHON_LIB)
+#endif
+
+#ifndef EMBED_IMPL_LIB
+#error "Name of embed implementation library required"
+#else
+#define EMBED_IMPL_LIB_STR xstr(EMBED_IMPL_LIB)
+#endif
+
+
+static void (*_embed_init_python)();
+static void (*_embed_sim_cleanup)();
+static int (*_embed_sim_init)(int argc, char const * const * argv);
+static void (*_embed_sim_event)(gpi_event_t level, const char *msg);
+
+static bool init_failed = false;
+
+
+#if ! defined(__linux__) && ! defined(__APPLE__)
+HANDLE act_ctx = NULL;
+
+BOOL WINAPI DllMain(HINSTANCE, DWORD fdwReason, LPVOID)
+{
+
+    if (fdwReason == DLL_PROCESS_ATTACH) {
+        // Save current activation context.
+        GetCurrentActCtx(&act_ctx);
+    } else if (fdwReason == DLL_PROCESS_DETACH) {
+        if (act_ctx) {
+            ReleaseActCtx(act_ctx);
+        }
+    }
+
+    return TRUE;
+}
+#endif
+
+
+extern "C" void embed_init_python(void)
+{
+    // preload python library
+    char const * libpython_path = getenv("LIBPYTHON_LOC");
+    if (!libpython_path) {
+        // default to libpythonX.X.so
+        libpython_path = PYTHON_LIB_STR;
+    }
+    auto loaded = utils_dyn_open(libpython_path);
+    if (!loaded) {
+        // LCOV_EXCL_START
+        init_failed = true;
+        return;
+        // LCOV_EXCL_STOP
+    }
+
+#if ! defined(__linux__) && ! defined(__APPLE__)
+    if (!act_ctx) {
+        // LCOV_EXCL_START
+        init_failed = true;
+        return;
+        // LCOV_EXCL_STOP
+    }
+
+    ULONG_PTR Cookie;
+    if(!ActivateActCtx(act_ctx, &Cookie)) {
+        // LCOV_EXCL_START
+        init_failed = true;
+        return;
+        // LCOV_EXCL_STOP
+    }
+#endif
+
+    // load embed implementation library and functions
+    void * embed_impl_lib_handle;
+    if (!(embed_impl_lib_handle = utils_dyn_open(EMBED_IMPL_LIB_STR))) {
+        // LCOV_EXCL_START
+        init_failed = true;
+        return;
+        // LCOV_EXCL_STOP
+    }
+    if (!(_embed_init_python = reinterpret_cast<decltype(_embed_init_python)>(utils_dyn_sym(embed_impl_lib_handle, "_embed_init_python")))) {
+        // LCOV_EXCL_START
+        init_failed = true;
+        return;
+        // LCOV_EXCL_STOP
+    }
+    if (!(_embed_sim_cleanup = reinterpret_cast<decltype(_embed_sim_cleanup)>(utils_dyn_sym(embed_impl_lib_handle, "_embed_sim_cleanup")))) {
+        // LCOV_EXCL_START
+        init_failed = true;
+        return;
+        // LCOV_EXCL_STOP
+    }
+    if (!(_embed_sim_init = reinterpret_cast<decltype(_embed_sim_init)>(utils_dyn_sym(embed_impl_lib_handle, "_embed_sim_init")))) {
+        // LCOV_EXCL_START
+        init_failed = true;
+        return;
+        // LCOV_EXCL_STOP
+    }
+    if (!(_embed_sim_event = reinterpret_cast<decltype(_embed_sim_event)>(utils_dyn_sym(embed_impl_lib_handle, "_embed_sim_event")))) {
+        // LCOV_EXCL_START
+        init_failed = true;
+        return;
+        // LCOV_EXCL_STOP
+    }
+
+#if ! defined(__linux__) && ! defined(__APPLE__)
+    if(!DeactivateActCtx(0, Cookie)) {
+        // LCOV_EXCL_START
+        init_failed = true;
+        return;
+        // LCOV_EXCL_STOP
+    }
+#endif
+
+    // call to embed library impl
+    _embed_init_python();
+}
+
+
+extern "C" void embed_sim_cleanup(void)
+{
+    if (!init_failed) {
+        _embed_sim_cleanup();
+    }
+}
+
+
+extern "C" int embed_sim_init(int argc, char const * const * argv)
+{
+    if (init_failed) {
+        // LCOV_EXCL_START
+        return -1;
+        // LCOV_EXCL_STOP
+    } else {
+        return _embed_sim_init(argc, argv);
+    }
+}
+
+
+extern "C" void embed_sim_event(gpi_event_t level, const char *msg)
+{
+    if (!init_failed) {
+        _embed_sim_event(level, msg);
+    }
+}
--- a/cocotb/share/lib/embed/gpi_embed.cpp
+++ b/cocotb/share/lib/embed/gpi_embed.cpp
@@ -31,7 +31,9 @@
 
 #include <Python.h>
 #include <cocotb_utils.h>
-#include "embed.h"
+#include <gpi_logging.h>        // LOG_* macros
+#include <py_gpi_logging.h>     // py_gpi_logger_set_level, py_gpi_logger_initialize, py_gpi_logger_finalize
+#include <gpi.h>                // gpi_event_t
 #include "locale.h"
 #include <cassert>
 
@@ -181,8 +183,7 @@ extern "C" void embed_sim_cleanup(void)
         PyGILState_Ensure();    // Don't save state as we are calling Py_Finalize
         Py_DecRef(pEventFn);
         pEventFn = NULL;
-        clear_log_handler();
-        clear_log_filter();
+        py_gpi_logger_finalize();
         Py_Finalize();
         to_simulator();
     }
@@ -228,7 +229,8 @@ extern "C" int embed_sim_init(int argc, char const * const * argv)
 
     PyObject *cocotb_module, *cocotb_init, *cocotb_retval;
     PyObject *cocotb_log_module = NULL;
-    PyObject *simlog_func;
+    PyObject *log_func;
+    PyObject *filter_func;
     PyObject *argv_list;
 
     cocotb_module = NULL;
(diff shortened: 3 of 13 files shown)


ktbarrett:
Still having issues on Windows. It can't find `libcocotb.dll` in the same directory. I was under the impression that the first search location was the directory of the library doing the call to `LoadLIbrary`?

ktbarrett:
Also checked on Arch Linux and it works. Will add an Arch CI task in the coming week.

themperek:
> Still having issues on Windows. It can't find `libcocotb.dll` in the same directory. I was under the impression that the first search location was the directory of the library doing the call to `LoadLIbrary`?

I think not. You have to add to `PATH` or we use this https://github.com/cocotb/cocotb/blob/c89e1de085f1204ee67e60fbe7e4d68192b4cf9f/cocotb_build_libs.py#L89

setup.py
```diff
@@ -96,14 +96,14 @@ def package_files(directory):
     author='Chris Higgs, Stuart Hodgson',
     maintainer='cocotb contributors',
     maintainer_email='cocotb@lists.librecores.org',
-    install_requires=[],
+    install_requires=['find_libpython'],
```
ktbarrett, line 99: We could copy the script locally to prevent a dependency. Or make this optional and if you don't have it, all you get is `libpython.so` so it resolves as it otherwise would?

ktbarrett:
@themperek libembed and libcocotb already have rcfiles being generated.

ktbarrett:
@themperek According to [this documentation](https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-desktop-applications), the directory from which the application (`libembed.dll` I presume is "the application" here) is searched first.

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    auto embed_lib_filepath = std::string(path);
+
+    // strip filename from path
+    auto embed_lib_dirpath = embed_lib_filepath.substr(0, embed_lib_filepath.rfind("\\"));
+
+    // combine the directory from the last step and the supplied filename
+    auto embed_impl_lib_path = embed_lib_dirpath + "\\" + EMBED_IMPL_LIB_STR;
+
+    return utils_dyn_open(embed_impl_lib_path.c_str());
```
ktbarrett, line 65: Implementation gleaned from [this answer](https://stackoverflow.com/a/65165593/6614127) and [this answer](https://stackoverflow.com/a/6924332/6614127).

While the use of `PathRemoveFileSpec` and `PathCombine` are recommended, it would require we link against `shwapi` library. We can expect the path returned from `GetModuleFileNameA` to be canonicalized, so spliting on `\` and appending `\libcocotb.dll` should "just work". We can combat that in the future if it becomes a problem.

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,133 @@
...
+#else
+    return utils_dyn_open(EMBED_IMPL_LIB_STR);
+#endif
+}
+
+
+extern "C" void embed_init_python(void)
+{
+    // preload python library
+    auto libpython_path = getenv("LIBPYTHON_LOC");
```
themperek, line 75: Maybe can store the path to libpython during setup and then load from c? No need for an extra environmental variables?
Use something like here: https://stackoverflow.com/questions/43409167/find-location-of-loaded-shared-library-from-in-that-shared-library to get the path?

themperek:
It seem to solve this problem https://github.com/cocotb/cocotb/issues/1945

themperek:
BTW: I think I do not see linking to libpython during setup when using python 3.6. Will check more.

cocotb/share/lib/embed/embed.cpp
ktbarrett: I'm not sure what you are proposing?

ktbarrett:
@themperek You are correct about not linking against libpython. You can see the code for distutils [here](https://github.com/python/cpython/blob/8a62887dfb4bb2835048780ad673362f7ee3c7bf/Lib/distutils/command/build_ext.py#L705). I seem to remember that fact stated in some previous thread, but I guess it wasn't the whole truth.

cocotb/share/lib/embed/embed.cpp
themperek: Not to use environmental variable but read from some file created on setup?

ktbarrett: If by setup you mean package installation, no. In fact that's what we are trying to avoid.
If you mean setup like right before we call the simulator then it's possible. It is roughly equivalent to using an environment variable, but involves writing and reading from disk, which seem unnecessary.

themperek: Not during compilation but some post install script? Should work even with binary distribution.

themperek: Maybe ...

ktbarrett: I looked into that and wheels don't support running arbitrary post-install code. We could have the behavior run once and cache in the site-packages, but I'm not sure if that kosher (?). Since this is an implementation detail the user doesn't need to be involved in, we can always change it later once we know. I'll ask on the PyPa discourse.

cocotb/share/include/gpi_logging.h
```diff
@@ -37,39 +37,183 @@
...
+    int level,
+    const char *pathname,
+    const char *funcname,
+    long lineno,
+    const char *msg,
+    va_list args);
 
-EXTERN_C_END
+/** Retrieve the current log handler.
+    @param handler  Location to return current log handler. If no custom logger is register this will be `NULL`.
```
cmarqu, line 154:
```suggestion
    @param handler  Location to return current log handler. If no custom logger is registered this will be `NULL`.
```

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+static void (*_embed_init_python)(void);
+static void (*_embed_sim_cleanup)(void);
+static int (*_embed_sim_init)(int argc, char const * const * argv);
+static void (*_embed_sim_event)(gpi_event_t level, const char *msg);
+
+static bool init_failed = false;
+
+
+#define PRINT_ERROR_WINDOZE(log_fmt) {\
```
cmarqu, line 29:
```suggestion
#define PRINT_ERROR_WINDOWS(log_fmt) {\
```
The original doesn't age well.

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -329,7 +320,7 @@ extern "C" int embed_sim_init(int argc, char const * const * argv)
     return ret;
 }
 
-extern "C" void embed_sim_event(gpi_event_t level, const char *msg)
+extern "C" COCOTB_EXPORT void _embed_sim_event(gpi_event_t level, const char *msg)
 {
     /* Indicate to the upper layer a sim event occurred */
```
cmarqu, line 325:
```suggestion
    /* Indicate to the upper layer that a sim event occurred */
```
While you are here...

cocotb/share/lib/py_gpi_log/py_gpi_logging.cpp
```diff
@@ -0,0 +1,167 @@
...
+        // check if log level is enabled
+        PyObject *filter_ret = PyObject_CallFunctionObjArgs(pLogFilter, logger_name_arg, level_arg, NULL);
+        if (filter_ret == NULL) {
+            goto error;
+        }
+
+        int is_enabled = PyObject_IsTrue(filter_ret);
+        Py_DECREF(filter_ret);
+        if (is_enabled < 0) {
+            /* A python exception occured while converting `filter_ret` to bool */
```
cmarqu, line 82:
```suggestion
            /* A Python exception occured while converting `filter_ret` to bool */
```
nit

cocotb/share/lib/py_gpi_log/py_gpi_logging.cpp
```diff
@@ -0,0 +1,167 @@
...
+            goto error;
+        }
+        Py_DECREF(handler_ret);
+    }
+
+    goto ok;
+error:
+    /* Note: don't call the LOG_ERROR macro because that might recurse */
+    gpi_native_logger_vlog(name, level, pathname, funcname, lineno, msg, argp_copy);
+    gpi_native_logger_log("cocotb.gpi", GPIError, __FILE__, __func__, __LINE__, "Error calling Python logging function from C while logging the above");
```
cmarqu, line 134:
```suggestion
    gpi_native_logger_log("cocotb.gpi", GPIError, __FILE__, __func__, __LINE__, "Error calling Python logging function from C++ while logging the above");
```

cmarqu:
The PR description should go into the wiki as documentation once merged.

cocotb/share/makefiles/Makefile.sim
```diff
@@ -100,4 +100,6 @@ ifeq ($(HAVE_SIMULATOR),0)
     $(error "Couldn't find makefile for simulator: "$(SIM_LOWERCASE)"! Available simulators: $(AVAILABLE_SIMULATORS)")
 endif
 
+export LIBPYTHON_LOC := $(shell find_libpython)
```
ktbarrett, line 103: Add error checking in case `find_libpython` does not find libpython. Could mention something along the lines of "libpython not found. You may not have the python development libraries installed, or find_libpython might not be able to find a libpython for your environment."

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+static void * load_embed_impl_lib()
+{
+#if ! defined(__linux__) && ! defined(__APPLE__)
+    HMODULE hm = NULL;
+
+    // get a handle to the module
+    if (!GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, reinterpret_cast<LPCSTR>(&embed_init_python), &hm))
+    {
+        PRINT_ERROR_WINDOZE("Unable to open embed library");
```
ktbarrett, line 47:
```suggestion
        PRINT_ERROR_WINDOWS("Unable to open embed library");
```

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    // get the path to the module
+    char path[MAX_PATH];
+    if (!GetModuleFileNameA(hm, path, sizeof(path)))
+    {
+        PRINT_ERROR_WINDOZE("Unable to get filepath of embed library");
```
ktbarrett, line 54:
```suggestion
        PRINT_ERROR_WINDOWS("Unable to get filepath of embed library");
```

cocotb/share/lib/embed/embed.cpp
ktbarrett: See [this answer](https://stackoverflow.com/a/58331693/6614127) about post-install scripts.

ktbarrett:
That's all I had intended to change. Fell free to add more comments. We will discuss how to handle `find_libpython` during the meeting on Wednesday.

cocotb/share/lib/embed/embed.cpp
themperek: In any case maybe should implement `cocotb-config --libpython-loc`?

themperek: Maybe use `std::system`? 
It would be just nice not to have to set variables in custom flows.

cocotb/share/include/gpi_logging.h
```diff
@@ -37,39 +37,183 @@
...
-# define EXTERN_C_START
-# define EXTERN_C_END
+extern "C" {
 #endif
 
-EXTERN_C_START
-
+/** Named logging level
+ *
+ *  Only logs at one of these log level will print a level name.
```
eric-wieser, line 48:
```suggestion
 *  Only logs at one of these log level will print a level name with the "native" logger
```
This statement is false for the python logger.

cocotb/share/include/gpi_logging.h
```diff
@@ -37,39 +37,183 @@
...
+    GPIDebug    = 10,   ///< Prints `DEBUG`    . Verbose information, useful for debugging
+    GPIInfo     = 20,   ///< Prints `INFO`     . Information about major events in the current program
+    GPIWarning  = 30,   ///< Prints `WARN`     . Encountered a recoverable bug, or information about surprising behavior
+    GPIError    = 40,   ///< Prints `ERROR`    . An unrecoverable error
+    GPICritical = 50    ///< Prints `CRITICAL` . An unrecoverable error, to be followed by immediate simulator shutdown
 };
 
+/** Logs a message at DEBUG log level using the current log handler.
+    Automatically populates arguments using information in the called context.
+ */
```
eric-wieser, line 60: Adding these docstrings would make a nice self-contained PR, rather that combining it with this one.

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+#ifndef EMBED_IMPL_LIB
+#error "Name of embed implementation library required"
+#else
+#define EMBED_IMPL_LIB_STR xstr(EMBED_IMPL_LIB)
+#endif
+
+
+static void (*_embed_init_python)(void);
+static void (*_embed_sim_cleanup)(void);
```
eric-wieser, line 22: This is C++ not C:
```suggestion
static void (*_embed_init_python)();
static void (*_embed_sim_cleanup)();
```

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }}
+
+
+static void * load_embed_impl_lib()
+{
+#if ! defined(__linux__) && ! defined(__APPLE__)
+    HMODULE hm = NULL;
+
+    // get a handle to the module
+    if (!GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, reinterpret_cast<LPCSTR>(&embed_init_python), &hm))
```
eric-wieser, line 45: https://docs.microsoft.com/en-us/windows/win32/intl/conventions-for-function-prototypes suggests we should be using the version without the `A` suffix:

> New Windows applications should use Unicode to avoid the inconsistencies of varied code pages and for ease of localization. They should be written with generic functions, and should define UNICODE to compile the functions into Unicode functions.

cocotb/share/lib/embed/embed.cpp
ktbarrett: We can certainly add `cocotb-config --libpython` to help out, but we can't run that from C++, since LD_LIBRARY_PATH or PATh may be modified by that point, affecting results. This must be run before the simulator is started, there isn't really a way around it.

cocotb/share/include/gpi_logging.h
ktbarrett: #2000 is half of this PR if that helps. But some changes to those files were already suggested and made here.

cocotb/share/lib/embed/embed.cpp
themperek: >LD_LIBRARY_PATH or PATh may be modified by that point, affecting results. 

This cannot influence the position of `cocotb-config`. How?

cocotb/share/lib/embed/embed.cpp
ktbarrett: The `cocotb_utils` library and the error logging macro here uses `FormatMessageA` since the `LOG_*` functions do not handle `wchar_t` strings. If the W-suffix were selected, logs will break. The MSDN documentation states that W-suffix and A-suffix functions should not be used together.

Right now, cocotb is an ANSI, English-only project. Fixing this will require a large internationalization PR and this suggestion would be a part of that.

eric-wieser: > Right now, cocotb is an ANSI, English-only project. 

Python is very much not ANSI-only, so even if we're "english only", we should try to avoid failing to install on systems with non-ascii paths.

> since the LOG_* functions do not handle wchar_t strings.

Python interprets the arguments to those functions as UTF8, not as ANSI - so the `A` apis may also not work correctly for characters not in the ASCII range.

cocotb/share/lib/embed/embed.cpp
themperek: I can imagine some crazy so `cocotb-config` will fail. Load wrong python interpreter?

cocotb/share/lib/embed/embed.cpp
ktbarrett: > we should try to avoid failing to install on systems with non-ascii paths

Perhaps this requires some experimentation since I'm fairly certain this is already the case. The issue isn't in the Python, but in the C code and the Python API code. If we get `wchar_t` and pass it to the Python logger we use the `s` format, but we need the `u` specifier to handle `wchar_t`. To support `wchar_t` in the logger we need more work that is out of scope for this PR. That being the case, we should follow Microsoft's guidance and not mix the two formats.

> so the A apis may also not work correctly for characters not in the ASCII range

"ANSI" doesn't mean anything to me, I think on Windows that means Latin-1 8 bit encoding.

cocotb/share/lib/embed/embed.cpp
ktbarrett: Yeah. Pick up the wrong libpython, wrong `PYTHONHOME` is deduced.

ktbarrett: What I can do to help is to make the `getenv` code check for `nullptr` and default to `"libpython{version}.{ext}"`, so you get standard lookup if it isn't specified. This way custom flows that trust their environment/simulator don't have to set anything extra.

themperek: 😭

cocotb/share/makefiles/Makefile.sim
```diff
@@ -100,4 +100,27 @@ ifeq ($(HAVE_SIMULATOR),0)
...
+# adapted from https://stackoverflow.com/a/24658961/6614127
+FIND_LIBPYTHON_RES := $(shell find_libpython; echo; echo $$?)
+FIND_LIBPYTHON_RC := $(lastword $(FIND_LIBPYTHON_RES))
+LIBPYTHON_LOC := $(strip $(subst $(FIND_LIBPYTHON_RC)QQQQ,,$(FIND_LIBPYTHON_RES)QQQQ))
+
+# complain if libpython isn't found, and export otherwise
+ifneq ($(FIND_LIBPYTHON_RC),0)
+    $(error $(find_libpython_errmsg))
+endif
+export LIBPYTHON_LOC
```
themperek, line 124: Maybe add and use `cocotb-config --libpython-loc` for this?

cocotb/share/lib/embed/embed.cpp
eric-wieser: >  I think on Windows that means Latin-1 8 bit encoding.

In which case, any latin1 characters will result in a `UnicodeDecodeError` from python, as the bytes in the latin1 range are used by UTF8 as a marker for a continuation byte - so the behavior is already broken on ANSI data too.

Using the `W` versions of these functions and `WideCharToMultiByte` to convert the wchar_t to UTF8 would prevent this error. We then just document that the `gpi_log` API takes utf8-encoded strings, which is already the case today.

eric-wieser: > If we get wchar_t and pass it to the Python logger we use the s format, but we need the u specifier to handle wchar_t.

I don't know what you mean by that. In `str.__mod__`, `%s` and `%u` are synonymous. `PyUnicode_FromString` is how we currently construct the log message, and that uses UTF8 input.

ktbarrett: > any latin1 characters will result in a UnicodeDecodeError from python, as the bytes in the latin1 range are used by UTF8 as a marker for a continuation byte

Right... so using `A` isn't sufficient either. So it sounds like you want me to make it an action to update to use the `W` suffix unconditionally. And translate all occurrences into UTF-8 before passing to the logger. 

> In str.__mod__, %s and %u are synonymous.

This isn't Python, in C they take two different types: `char*` and `wchar_t*`. Passing `wchar_t` to `s` specifier in Python C API code will most certainly break as it will reinterpret the byte sequence as `char`s. But that doesn't matter now since we will be manually calling out a conversion function.

eric-wieser: > Right... so using A isn't sufficient either.

Sadly not - otherwise I would probably have accepted your argument.

> Passing wchar_t to s specifier in Python C API code will most certainly break as it will reinterpret the byte sequence as chars.

Where is this `s` specifier in our code that you're referring to?

ktbarrett: I guess there is no specific use of `s` for logging in a `BuildTuple` or similar that I thought existed (it's been a while since I've looked at that code, I made these changes in June I think). But `gpi_log` and friends only work on UTF-8 and cannot accept `wchar_t*` or UTF-16 directly. If we translate it before passing, then nbd. That's what I had though you were suggesting originally.

eric-wieser: "translate before passing" sounds like the right approach to me.

cocotb/share/lib/embed/embed.cpp
themperek: > What I can do to help is to make the `getenv` code check for `nullptr` and default to `"libpython{version}.{ext}"`, so you get standard lookup if it isn't fully specified. This way custom flows that trust their environment/simulator don't have to set anything extra.

After some thinking, maybe better to fail with proper error than some nonsense.

tpambor:
`LoadLibrary` failed because the wrong [activation context](https://docs.microsoft.com/en-us/windows/win32/sbscs/activation-contexts) was used for loading `libcocotb`. I attached two patches that fix it. It is not necesarry to fiddle around manually with paths and character encoding.
[0001-Revert-windows-specific-workarounds.patch](https://github.com/cocotb/cocotb/files/5655422/0001-Revert-windows-specific-workarounds.patch.txt)
[0002-Use-the-correct-activation-context-for-loading-libco.patch](https://github.com/cocotb/cocotb/files/5655420/0002-Use-the-correct-activation-context-for-loading-libco.patch.txt)

cocotb/share/makefiles/Makefile.sim
```diff
@@ -100,4 +100,27 @@ ifeq ($(HAVE_SIMULATOR),0)
     $(error "Couldn't find makefile for simulator: "$(SIM_LOWERCASE)"! Available simulators: $(AVAILABLE_SIMULATORS)")
 endif
 
+define find_libpython_errmsg =
+
+
+find_libpython was not able to find a libpython in the current Python environment. Ensure
+the Python development packages are installed. If they are installed and find_libpython
+is not finding the path to libpython, file an upstream bug in find_libpython; then
+manually override the LIBPYTHON_LOC make variable with the aboslute path to libpython.so
```
cmarqu, line 109:
```suggestion
manually override the LIBPYTHON_LOC make variable with the absolute path to libpython.so
```

cocotb/share/lib/embed/embed.cpp
ktbarrett: Thanks to Tim's patches, we don't have to worry about this anymore for this PR. Opened #2273.

cocotb/share/makefiles/Makefile.sim
ktbarrett: Done.

cocotb/share/lib/embed/embed.cpp
ktbarrett: Done.

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+static bool init_failed = false;
+
+#if ! defined(__linux__) && ! defined(__APPLE__)
+HANDLE act_ctx = NULL;
+
+BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
+{
+    UNREFERENCED_PARAMETER(hinstDLL);
+    UNREFERENCED_PARAMETER(lpvReserved);
```
ktbarrett, line 34:
```suggestion
```

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,140 @@
...
+static void (*_embed_sim_cleanup)();
+static int (*_embed_sim_init)(int argc, char const * const * argv);
+static void (*_embed_sim_event)(gpi_event_t level, const char *msg);
+
+static bool init_failed = false;
+
+#if ! defined(__linux__) && ! defined(__APPLE__)
+HANDLE act_ctx = NULL;
+
+BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
```
ktbarrett, line 31:
```suggestion
BOOL WINAPI DllMain(HINSTANCE, DWORD fdwReason, LPVOID)
```

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,140 @@
...
+{
+    UNREFERENCED_PARAMETER(hinstDLL);
+    UNREFERENCED_PARAMETER(lpvReserved);
+
+    if (fdwReason == DLL_PROCESS_ATTACH) {
+        // Save current activation context.
+        GetCurrentActCtx(&act_ctx);
+    } else if (fdwReason == DLL_PROCESS_DETACH) {
+        if (act_ctx)
+            ReleaseActCtx(act_ctx);
```
ktbarrett, line 41:
```suggestion
        if (act_ctx) {
            ReleaseActCtx(act_ctx);
        }
```

garmin-mjames:
Is `libembed` a replacement for #1994 or more a stepping stone down that path?

ktbarrett:
@garmin-mjames Its just a stepping stone. Its a minimum code change to get the positive outcome (binary distibution) I wanted now. The rest will come later.

ktbarrett:
Vendored in find_libpython. pip installing in editable mode doesn't make find_libpython locatable. pip editable only makes a single cocotb.egg-link based on the project name. So EVERYTHING must go under the cocotb module if we want to use it in editable mode. This stuff is so jank...

ktbarrett:
Apparently if the vendored libs are included at the top-level, it has no problem finding the package. Still I think the previous approach of putting it in the cocotb namespace is better, just in case they have the actual package installed it doesn't collide.

cocotb/share/lib/embed/embed.cpp
themperek: From some simple testing, it looks `cocotb-config` can be called and is working fine.

ktbarrett:
If desired I can add gcov exemptions for uncovered code that should never be hit.

cocotb/share/lib/embed/embed.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return utils_dyn_open(EMBED_IMPL_LIB_STR);
+}
+
+
+extern "C" void embed_init_python(void)
+{
+    // preload python library
+    auto libpython_path = getenv("LIBPYTHON_LOC");
+    if (!libpython_path) {
+        LOG_ERROR("Environment variable LIBPYTHON_LOC not set to the path of your environment's libpython");
```
themperek, line 58: Sorry for going this one more time. I did some basic testing and it looks `cocotb-config` can be called even on currently broken xcelium. Do not have a strong opinion but what you think to try to call it in case  `LIBPYTHON_LOC` not set with some worning?

themperek (approved):
From a packaging point of view, this is a great improvement. 
From library changes, others should comment.

cocotb/share/lib/embed/embed.cpp
ktbarrett: Are you suggesting calling `cocotb-config --libpython` from embed.cpp? If this is desired then we should move the call unconditionally to embed.cpp. I'm just afraid that won't work in certain cases if `cocotb-config` is not on the PATH; for instance on Windows when you are not using conda, or maybe a simulator modifies PATH.

I'd rather it always work without fail and have 1 additional requirement on advanced users, than ___usually___ work and save little work. I had thought it would be a good idea to try defaulting to "libpythonX.X.so" if `LIBPYTHON_LOC` is not set and that will work in the common case without setting `LIBPYTHON_LOC`.

ktbarrett:
@cmarqu Unfortunately, gcovr doesn't understand the "new" GCOV_* prefix names or the branch exclusion pragmas like `LCOV_EXCL_BR_START` (https://github.com/gcovr/gcovr/issues/121).

cocotb/share/lib/embed/embed.cpp
ktbarrett: I added the `libpythonX.X.so` default back in.

cmarqu:
> @cmarqu Unfortunately, gcovr doesn't understand the "new" GCOV_* prefix names or the branch exclusion pragmas like `LCOV_EXCL_BR_START` ([gcovr/gcovr#121](https://github.com/gcovr/gcovr/issues/121)).

So we could use lcov instead of gcovr... I chose the latter because it's just a `pip install`away, I cannot so easily get hold of lcov.

cocotb/share/lib/embed/embed.cpp
themperek: In principle, we require `cocotb-config` to be on the `PATH` always: https://docs.cocotb.org/en/stable/install.html#installation-of-cocotb
Like now is also ok. As I said no strong opinion. I Will test this a bit in next days.

ktbarrett: Thank you for testing it.

ktbarrett:
@cmarqu I added a *bunch* of exclusions to `gpi_embed.cpp`, `embed.cpp`, and `py_gpi_logging.cpp` using just `LCOV_EXCL_START` and `LCOV_EXCL_STOP`. gcovr is probably the better option, I'm not sure how easy distribution of lcov is to platforms that aren't Linux. We might need to/want to contribute to gcovr to better support the other pragmas. As soon as we understand exactly how they are supposed to work, that is. I get < 20 results for `LCOV_EXCL_BR_START` on Google, and none are examples. It might not even work on lcov!

garmin-mjames (commented):
Only comments/doc cleanup.

Also, looks like `_get_python_lib()` and `_get_python_lib_link()` can be removed from `cocotb_build_libs.py` now.

cocotb/share/makefiles/Makefile.sim
```diff
@@ -100,4 +100,27 @@ ifeq ($(HAVE_SIMULATOR),0)
...
+FIND_LIBPYTHON_RES := $(shell cocotb-config --libpython; echo $$?)
+FIND_LIBPYTHON_RC := $(lastword $(FIND_LIBPYTHON_RES))
+LIBPYTHON_LOC := $(strip $(subst $(FIND_LIBPYTHON_RC)QQQQ,,$(FIND_LIBPYTHON_RES)QQQQ))
+
+# complain if libpython isn't found, and export otherwise
+ifneq ($(FIND_LIBPYTHON_RC),0)
+    $(error $(find_libpython_errmsg))
+endif
+export LIBPYTHON_LOC
+
```
themperek, line 125: Should not be in `Makefile.inc`? `Makefile.sim` is mostly to dispatch to the appropriate simulator.

cocotb/vendor/find_libpython/__init__.py
```diff
@@ -0,0 +1,383 @@
+"""
```
themperek, line 1: When installed from zip/sdist on centos 7 I get this:
```
  File "/usr/local/bin/cocotb-config", line 11, in <module>
    load_entry_point('cocotb==1.5.0.dev0', 'console_scripts', 'cocotb-config')()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 476, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2700, in load_entry_point
    return ep.load()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2318, in load
    return self.resolve()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2324, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/lib64/python3.6/site-packages/cocotb/config.py", line 46, in <module>
    import cocotb.vendor.find_libpython as find_libpython
ModuleNotFoundError: No module named 'cocotb.vendor'
```

ktbarrett: Added `cocotb/vendor/__init__.py`, which solves the problem.

cocotb/share/lib/py_gpi_log/py_gpi_logging.cpp
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * GILState before calling: Unknown
+ *
+ * GILState after calling: Unknown
+ *
+ * Makes one call to PyGILState_Ensure and one call to PyGILState_Release
+ *
+ * If the Python logging mechanism is not initialised, dumps to `stderr`.
+ *
+ */
```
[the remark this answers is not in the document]
ktbarrett: This is a "private" function, so I just removed it.

ktbarrett:
I'll rebase to clean up commits if all is well.

cocotb/share/lib/embed/embed.cpp
themperek: All platform and simulators I have access seem to be happy.

ktbarrett:
I will be working on a commit to add exemptions to the code coverage so we can exclude everything under `cocotb.vendor` and `cocotb.config`, etc.

ktbarrett:
Thanks again for the assistance @tpambor!

PR #2287, opened 2020-12-12
Use the lowest possible Python version for 3.5-3.7 in CI

cmarqu:

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -61,19 +61,22 @@ jobs:
           - sim: icarus
             sim-version: apt
             lang: verilog
-            python-version: 3.5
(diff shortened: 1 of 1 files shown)


cmarqu:
The .0 versions are not available (I requested some in https://github.com/actions/virtual-environments/issues/2253), so this PR takes the lowest possible versions for now.
I'm wondering whether we need those highest minor version to run in addition? In theory they could also break.

ktbarrett:
> I'm wondering whether we need those highest minor version to run in addition? In theory they could also break.

Technically every patch version could break. I don't think we need to check that on every PR, maybe we should test more during release?

ktbarrett (approved):

PR #2288, opened 2020-12-12
Remove last occurrences of legacy coroutine syntax from tests

### LINKED ISSUE #1731: Convert tests to use `async def` syntax

ktbarrett:
Closes #1731. Added justification comments to all uses of `cocotb.coroutine` and all uses of `cocotb.test` wrapping generators.

### DIFF
--- a/tests/test_cases/test_cocotb/test_generator_coroutines.py
+++ b/tests/test_cases/test_cocotb/test_generator_coroutines.py
@@ -175,7 +175,7 @@ def test_exceptions_direct(dut):
     Traceback \(most recent call last\):
       File ".*common\.py", line \d+, in _check_traceback
         await running_coro
-      File ".*\/cocotb\/decorators.py", line \d+, in __await__
+      File ".*cocotb[\\\/]decorators.py", line \d+, in __await__
         return \(yield self\)
       File ".*test_generator_coroutines\.py", line \d+, in raise_soon
         yield raise_inner\(\)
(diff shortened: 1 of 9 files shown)


tests/test_cases/test_cocotb/test_generator_coroutines.py
```diff
@@ -169,7 +174,9 @@ def raise_soon():
     expected = textwrap.dedent(r"""
     Traceback \(most recent call last\):
       File ".*common\.py", line \d+, in _check_traceback
-        yield running_coro
+        await running_coro
+      File ".*\/cocotb\/decorators.py", line \d+, in __await__
```
ktbarrett, line 178:
```suggestion
      File ".*cocotb[\\\/]decorators.py", line \d+, in __await__
```

tests/test_cases/test_cocotb/test_generator_coroutines.py
```diff
@@ -198,7 +205,9 @@ def raise_soon():
...
```
ktbarrett, line 209:
```suggestion
      File ".*cocotb[\\\/]decorators.py", line \d+, in __await__
```

cmarqu (approved):

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -419,7 +418,6 @@ async def test_start_soon_decorator(_):
     """ Tests start_soon works with RunningTasks """
     a = 0
 
-    @cocotb.coroutine
```
marlonjames, line 422: Without the decorator this testcase is a duplicate of the one above it.

PR #2289, opened 2020-12-14
Move bus and testbenching components to cocotb-bus repo

ktbarrett:
Companioned by cocotb/cocotb-bus#1.

* Move the bus and testbenching components into the cocotb-bus repo and `cocotb_bus` package.
* Adds extras_require for regressions, since we are leaving the tests and documentation of the moved components in the repo for this PR.
* Allows current imports to work by renaming the imports at runtime (also prints `DeprecationWarning`).

### TODO

- [x] Fix documentation names
- [x] Mention in README and documentation the existance of new package
- [x] Add to installation and README the use of `cocotb[bus]`
- [x] Decide what to do with `cocotb.wavedrom`

### FOLLOW ON CHANGES
* move tests into cocotb-bus
* move documentation into cocotb-bus


ktbarrett:
@eric-wieser Looks like `Wavedrom` never worked with anything other than `Bus`. The `get` method expects the object to have a `_name` and that attribute is only set when using a `Bus`. I will add an optional name to the constructor to set `_name` in cases where we aren't constructing from a `Bus`.

ktbarrett:
Switch to using reimport stubs rather than the in-flight import renaming I was using. the previous solution didn't cover `from cocotb import bus` and it slowed down the common case to support an exceptional case. 

There are issues with the current solution. If the structure of `cocotb-bus` package changes, the reimport stubs must also change. But this is preferable for the end user, and eventually the stubs will disappear anyways...

ktbarrett:
Fixed Windows CI, I had forgot to add cocotb-bus as a dep.

ktbarrett:
I guess I falsely figured that 1.5.0dev > 1.4. I opened https://github.com/cocotb/cocotb-bus/pull/18 and will have to put out a new patch release after it is merged =/.

ktbarrett:
```
> tox -e py38-linux
py38-linux recreate: /home/kaleb/dev/cocotb/.tox/py38-linux
py38-linux installdeps: coverage, xunitparser, pytest, pytest-cov, cocotb-bus
ERROR: invocation failed (exit code 1), logfile: /home/kaleb/dev/cocotb/.tox/py38-linux/log/py38-linux-1.log
========================================================================= log start ==========================================================================
Collecting cocotb-bus
  Downloading cocotb_bus-0.1.1-py3-none-any.whl (34 kB)
  Downloading cocotb_bus-0.1.0-py3-none-any.whl (34 kB)
ERROR: Cannot install cocotb-bus==0.1.0 and cocotb-bus==0.1.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    cocotb-bus 0.1.1 depends on cocotb<2.0 and >=1.5.0.dev
    cocotb-bus 0.1.0 depends on cocotb<2.0 and >1.4

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies

========================================================================== log end ===========================================================================
ERROR: could not install deps [coverage, xunitparser, pytest, pytest-cov, cocotb-bus]; v = InvocationError('/home/kaleb/dev/cocotb/.tox/py38-linux/bin/python -m pip install coverage xunitparser pytest pytest-cov cocotb-bus', 1)
__________________________________________________________________________ summary ___________________________________________________________________________
ERROR:   py38-linux: could not install deps [coverage, xunitparser, pytest, pytest-cov, cocotb-bus]; v = InvocationError('/home/kaleb/dev/cocotb/.tox/py38-linux/bin/python -m pip install coverage xunitparser pytest pytest-cov cocotb-bus', 1)
```

Forgot deps are installed first... I don't know what I should do here.

.github/workflows/regression-tests.yml
```diff
@@ -237,7 +237,7 @@ jobs:
       if: startsWith(matrix.os, 'windows')
       run: python -m pip install  --global-option build_ext --global-option --compiler=mingw32 -e .
     - name: Install Python testing dependencies (Windows)
-      run: pip install coverage xunitparser pytest pytest-cov
+      run: pip install coverage xunitparser pytest pytest-cov cocotb-bus
```
eric-wieser, line 240: I don't think this is a good idea. Any tools that depend on cocotb should be installed after cocotb itself.

eric-wieser: Oh, nevermind

tox.ini
```diff
@@ -36,6 +36,7 @@ deps =
     xunitparser
     pytest
     pytest-cov
+    cocotb-bus
```
eric-wieser, line 39: I think this line is a mistake

ktbarrett: I understand tox doesn't separate build dependencies from runtime dependencies, so it installes depends first. We have a catch-22. Should cocotb-bus it be installed via extras?

eric-wieser: Is this actually a dependency anyway? Surely its only needed for the docs?

ktbarrett: A lot of tests depend on what was moved. I just tested it, using `extras` works. I will also add `pip_pre = true` option and we can "yank" cocotb-bus v0.1.1 from Pypi and revert that commit, if desired.

ktbarrett:
Had to fix ReadTheDocs by adding the `bus` extra.

ktbarrett:
Using `--pre` instead of updating the cocotb version requirement in cocotb-bus would require that ReadTheDocs support `--pre`, which is current does not.

documentation/source/library_reference.rst
```diff
@@ -77,27 +73,34 @@ classes used within ``cocotb``.
 Testbench Structure
 ===================
```
cmarqu, line 75:
```suggestion
These are provided by the `cocotb-bus <https://github.com/cocotb/cocotb-bus>`_ package.
```

documentation/source/newsfragments/2289.removal.rst
```diff
@@ -0,0 +1,9 @@
+The bus and testbenching components in cocotb have been officially moved to the `cocotb-bus <https://github.com/cocotb/cocotb-bus>`_ package.
+This includes
+:class:`~cocotb_bus.bus.Bus`,
+:class:`~cocotb_bus.scoreboard.Scoreboard`,
+everything in :mod:`cocotb_bus.drivers <cocotb.drivers>`,
+and everything in :mod:`cocotb_bus.monitors <cocotb.monitors>`.
+Documentation will remain in the main cocotb repo for now.
+Old names will continue to exist, but their use will cause :class:`DeprecationWarning`,
```
cmarqu, line 8:
```suggestion
Old names will continue to exist, but their use will cause a :class:`DeprecationWarning`,
```
?

cocotb/wavedrom.py
```diff
@@ -1,49 +1,40 @@
...
+            try:
+                from cocotb_bus.bus import Bus
+            except ImportError:
+                raise TypeError(
+                    "Cannot use {} with {} objects".format(type(self).__qualname__, type(obj).__name__)
+                ) from None
+            else:
+                if isinstance(obj, Bus):
+                    self._hdls.update(obj._signals)
+                    self._name = obj._name
```
ktbarrett, line 37: This is off. We need the `raise TypeError` statement `if instance(obj, Bus)` check fails, as well. Should move the `Bus` check to the toplevel and default it to `None` if the import fails. Then add a `Bus is not None` clause before the `insinstance` check and deindent it.

tests/pytest/test_deprecated_imports.py
```diff
@@ -0,0 +1,15 @@
...
+        import cocotb.bus  # noqa: F401
+
+    with pytest.warns(DeprecationWarning):
+        import cocotb.drivers  # noqa: F401
+
+    with pytest.warns(DeprecationWarning):
+        import cocotb.monitors  # noqa: F401
+
+    with pytest.warns(DeprecationWarning):
+        import cocotb.scoreboard  # noqa: F401
```
ktbarrett, line 15: Add tests for all modules mocked out as well to increase coverage. Maybe use `importlib.import_module` and make this a list of names instead for flexibility and conciseness.

ktbarrett: Nvm, `import_module` is swallowing the `DeprecationWarning`s for some reason. And the warnings are only printed once.

cmarqu (approved):

imphil (approved):

PR #2290, opened 2020-12-14
Update Clock to drive all bits of connected signal

alexforencich:
Update `Clock` to drive all bits of connected signal.  This is a bit of a stop-gap fix for handling concatenated ports, but I think the functionality is sensible regardless.

### DIFF
--- a/cocotb/clock.py
+++ b/cocotb/clock.py
@@ -63,7 +63,8 @@ class Clock(BaseClock):
         cocotb.fork(c.start())
 
     Args:
-        signal: The clock pin/signal to be driven.
+        signal: The clock pin/signal to be driven.  Given a vector signal, all bits will
+            be driven.
         period (int): The clock period. Must convert to an even number of
             timesteps.
         units (str, optional): One of
@@ -109,6 +110,7 @@ class Clock(BaseClock):
 
     .. versionchanged:: 1.5
         Support ``'step'`` as the the *units* argument to mean "simulator time step".
+        Given a vector signal, set all bits in the vector, not just the rightmost bit.
 
     .. deprecated:: 1.5
         Using None as the the *units* argument is deprecated, use ``'step'`` instead.
@@ -150,17 +152,19 @@ class Clock(BaseClock):
             it = range(cycles)
 
         # branch outside for loop for performance (decision has to be taken only once)
+        high_value = 2**len(self.signal)-1
+        low_value = 0
         if start_high:
             for _ in it:
-                self.signal <= 1
+                self.signal <= high_value
                 await t
-                self.signal <= 0
+                self.signal <= low_value
                 await t
         else:
             for _ in it:
-                self.signal <= 0
+                self.signal <= low_value
                 await t
-                self.signal <= 1
+                self.signal <= high_value
                 await t
 
     def __str__(self):
--- a/documentation/source/newsfragments/2290.change.rst
+++ b/documentation/source/newsfragments/2290.change.rst
@@ -1,0 +1,1 @@
+:class:`cocotb.clock.Clock` now drives all bits of a vector signal.


cocotb/clock.py
```diff
@@ -109,6 +109,7 @@ async def custom_clock():
 
     .. versionchanged:: 1.5
         Support ``'step'`` as the the *units* argument to mean "simulator time step".
+        Drive all bits of connected signal.
```
ktbarrett, line 112:
```suggestion
        Given a vector signal, all bits in the vector will be set, not just the rightmost bit. 
```

cocotb/clock.py
```diff
@@ -121,6 +122,8 @@ def __init__(self, signal, period, units="step"):
                 'Using units=None is deprecated, use units="step" instead.',
                 DeprecationWarning, stacklevel=2)
             units="step"  # don't propagate deprecated value
+        self.high_value = 2**len(signal)-1
+        self.low_value = 0
```
ktbarrett, line 126:
```suggestion
```

cocotb/clock.py
```diff
@@ -152,15 +155,15 @@ async def start(self, cycles=None, start_high=True):
         # branch outside for loop for performance (decision has to be taken only once)
         if start_high:
```
ktbarrett, line 156:
```suggestion
        high_value = 2**len(self.signal)-1
        low_value = 0
        if start_high:
```

cocotb/clock.py
```diff
@@ -152,15 +155,15 @@ async def start(self, cycles=None, start_high=True):
...
             for _ in it:
-                self.signal <= 1
+                self.signal <= self.high_value
```
ktbarrett, line 158:
```suggestion
                self.signal <= high_value
```

cocotb/clock.py
```diff
@@ -152,15 +155,15 @@ async def start(self, cycles=None, start_high=True):
...
                 await t
-                self.signal <= 0
+                self.signal <= self.low_value
```
ktbarrett, line 160:
```suggestion
                self.signal <= low_value
```

cocotb/clock.py
```diff
@@ -152,15 +155,15 @@ async def start(self, cycles=None, start_high=True):
...
                 await t
         else:
             for _ in it:
-                self.signal <= 0
+                self.signal <= self.low_value
```
ktbarrett, line 164:
```suggestion
                self.signal <= low_value
```

cocotb/clock.py
```diff
@@ -152,15 +155,15 @@ async def start(self, cycles=None, start_high=True):
...
                 await t
-                self.signal <= 1
+                self.signal <= self.high_value
```
ktbarrett, line 166:
```suggestion
                self.signal <= high_value
```

documentation/source/newsfragments/2290.change.rst
```diff
@@ -0,0 +1 @@
+In :class:`cocotb.clock.Clock`, drive all bits of the connected signal.
```
ktbarrett, line 1:
```suggestion
In :class:`cocotb.clock.Clock`, when given a vector signal, all bits of the signal are driven.
```

ktbarrett (commented):
Update the doctsring in `Clock` to mention the `signal` argument can be a vector.

I suggest changing the implementation to avoid an unnecessary public interface and attribute lookup (for perf).

alexforencich:
Ah, I didn't realize there was a penalty to using instance variables instead of locals.  I will update that, and probably make similar changes to all of my extension modules.

ktbarrett:
@alexforencich `timeit` is your friend. Accessing an attribute is ~3x slower. Local is always better. It's a common convention in some languages, like Lua, to create local references to global or nested objects because they are provably faster.

```
> python -m timeit -s "a = 8" "a"
20000000 loops, best of 5: 12.5 nsec per loop

> python -m timeit -s "class A: pass" -s "a = A()" -s "a.a = 7" "a.a"
10000000 loops, best of 5: 37 nsec per loop
```

ktbarrett:
Reminder that the docstring needs to be updated to reflect that you can pass a vector to `Clock` and it will drive all the bits as the same clock. And the newsfragment should match the versionchanged line.

alexforencich:
Updated the docstring.  Looks like I missed adding the newsfragment change to my commit, that's fixed now.

documentation/source/newsfragments/2290.change.rst
```diff
@@ -0,0 +1 @@
+When passing a vector signal to :class:`cocotb.clock.Clock`, drive all bits of the signal, not just the right-most bit.
```
garmin-mjames, line 1:
```suggestion
:class:`cocotb.clock.Clock` now drives all bits of a vector signal.
```

cocotb/clock.py
```diff
@@ -109,6 +110,7 @@ async def custom_clock():
 
     .. versionchanged:: 1.5
         Support ``'step'`` as the the *units* argument to mean "simulator time step".
+        Given a vector signal, set all bits in the vector, not just the rightmost bit.
```
eric-wieser, line 113:
```suggestion
        Given a vector signal, set all bits in the vector, not just the least-significant bit.
```

eric-wieser:
I'm not really sure how much sense this makes, but I guess it's better than the current behavior. IMO making `Clock` just raise an error on signals that aren't one bit wide would also be fairly sensible.

ktbarrett:
I agree with Eric. I'm not sure this should be supported, but this is better than the current behavior. You *could* just instantiate multiple clocks per signal, but I think your issue is that you are having issues doing part selects in some simulators?

alexforencich:
I would be fine with not merging this if you get part-selects working for all simulators.  It should be possible to emulate this in cocotb even if the simulator does not support it - reads can trivially be done as full-width and then sliced and diced, writes can be coalesced and then applied either bit by bit (if supported by the simulator) or full-width.  This functionality may need to be implemented at the GPI to support GPI clocks on all simulators.

eric-wieser:
I guess my thought would be that driving a multi-bit signal as a "clock" is a special enough case that it a user should just write their own trivial coroutine.

alexforencich:
Well, if you want to get the performance advantage of GPI clocks, you can't just write a coroutine.  So IMO it makes sense to think about some of these edge cases so that we can make the GPI clock implementation sufficiently flexible.  

On a related note, I think it would be a good idea to consider adding duty cycle and perhaps start delay (for multiphase clocks) to the clock object as well, for the same reason (incorporation into the GPI clock API).

ktbarrett:
> I would be fine with not merging this if you get part-selects working for all simulators.

I'm not sure that's up to us, simulator support for VPI and lal the object relations is not the greatest, across the board. Maybe we can do something about it, do you have an open issue about the failure?

alexforencich:
> > I would be fine with not merging this if you get part-selects working for all simulators.
> 
> I'm not sure that's up to us, simulator support for VPI and all the described object relations is not the greatest, across the board. Maybe we can do something about it, do you have an open issue about the failure?

Right, so the idea would be to implement a layer somewhere - could be in python, but ideally it would be in GPI so it could be used by simulation models that operate at the GPI layer, including GPI clocks - that would permit this in a simulator-independent way by transparently emulating it when the simulator or API does not support it.  I will put together an issue for my primary use-case for discussion.

alexforencich:
I just opened #2302

ktbarrett:
Rather than let this sit for forever I'm going to suggest we close this. I don't think we need this feature in the base clock object. Another agreement can close the issue.

alexforencich:
If #2302 is implemented in a timely fashion, then this can be closed.  Otherwise, this should be merged.

PR #2292, opened 2020-12-15
Use Python 3.7.1 which is new on GitHub Actions

cmarqu:
See https://github.com/actions/virtual-environments/issues/2253
for details about problems going all the way down to 3.7.0 and 3.6.0.

Refs #2287.


ktbarrett:
Is there some reason they didn't add 3.7.0? Is that coming in the near future?

ktbarrett (approved):

Issue #2189, opened 2020-11-11
Agree on and add clang-format configuration


imphil:
Our C/C++ code is currently a mix of styles and not auto-formatted. I'm proposing to add a clang-format configuration to solve that.

In addition, we should agree on a C/C++ coding style. I tried to play around with clang-format to get it to match what we currently have, but our style is so inconsistent that it will be a rather large diff anyways. I'd therefore go with a full reformat with a standard coding style already supported in clang-format.

As usual, this is a matter of taste, but also doesn't matter much. I'd go with the Google C/C++ coding style (https://google.github.io/styleguide/cppguide.html), which is rather comprehensive and well supported in clang-format. (It has a couple things in there which I personally find odd, but that's expected in any style and I ultimately don't care enough.)

- [ ] Agree on a style
- [ ] Add a clang-format config file to the tree
- [ ] Reformat all C/C++ code with clang-format.
- [ ] Add clang-format lint to CI.

ktbarrett:
Need to take #1148 into consideration. It had to do with the `Python.h` include being moved around, but I'm afraid I forget the details.

ktbarrett:
Apparently that's not an issue, I just tried it. I'm fine with the Google code style. [clang-format-lint-action](https://github.com/DoozyX/clang-format-lint-action) can be used for linting.

garmin-mjames:
Google style is fine with me. We may be able to use [include-what-you-use-action](https://github.com/emilgedda/include-what-you-use-action) for include checking.

PR #2293, opened 2020-12-16
Add GPI-layer clock driver

### LINKED ISSUE #89: Move clock generation into C++

marlonjames:
Closes #89.

This adds a `cocotb.simulator.sim_clock` class as a GPI-layer clock driver.
The callback, registered with `gpi_register_timed_callback()`, doesn't interact with Python at all, it simply drives a clock signal
and registers for the next edge. It is designed to take values in simulation time steps.

This PR also introduces the ~~`cocotb.clock.SimClock`~~ `cocotb.clock.CClock` class to wrap and control the GPI driver. It is responsible for checking clock configuration values and calculating the time steps to pass to the wrapped `sim_clock` object.

Another important difference with `sim_clock` objects is that the clock value is being set in the callback (which is called along with `Timer` trigger callbacks during a time step), and not cached for writing in the `ReadWrite` callback like a `dut.clk <= 1` assignment is.

Per discussion in #89, the clocks have jitter support. Jitter is defined in terms of a clock period, rising edge to rising edge.
Jitter budget is divided between high and low parts of the clock period, scaled to the duty cycle.
Jitter is symmetrical, with bounds `[-jitter, +jitter]` for the clock period.
Jitter is generated from a `std::normal_distribution` and seeded using the same `RANDOM_SEED` as Python for determinism.
Each clock has it's own normal distribution, configured with 3-sigma (~99.7%) inside the jitter bounds.

While the jitter budget is specified per-period, the random jitter is generated per edge-to-edge, which results in a narrower distribution (higher sigma-squared) when looking at jitter from one rising edge to the next. This means ~84% period jitter is within 1-sigma vs ~68% in a standard normal distribution.

The scheduler needs to keep track of the `sim_clock` objects so that they can be stopped between tests. When something like #1349 is in, each clock can be managed by a forked coroutine that do cleanup when it's cancelled/killed.

I used the google style guide linked from #2189 for the C++ code.

Todo:
- [ ] newsfragment
---

Outstanding issues:
- [x] ~~Should `start()` instead return a coroutine to match `Clock.start()` behavior?~~ No.
- [ ] Converting `Clock` to use this will require additional signalling to match clock behavior when using `cycles` for a clock that stops itself
    - i.e., `await Join(forked_clock_coro)` to know when the clock stopped
- [ ] Change jitter calculation from per-edge to per-period?
- [ ] Immediate vs cached signal value change

I'm leaving as draft for now. Comments are very welcome.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -232,15 +232,20 @@ def _initialise_testbench(argv_):
     if RANDOM_SEED is None:
         if 'ntb_random_seed' in plusargs:
             RANDOM_SEED = eval(plusargs['ntb_random_seed'])
+            RANDOM_SOURCE = "plusarg ntb_random_seed"
         elif 'seed' in plusargs:
             RANDOM_SEED = eval(plusargs['seed'])
+            RANDOM_SOURCE = "plusarg seed"
         else:
             RANDOM_SEED = int(time.time())
-        log.info("Seeding Python random module with %d" % (RANDOM_SEED))
+            RANDOM_SOURCE = "time.time() (default)"
     else:
         RANDOM_SEED = int(RANDOM_SEED)
-        log.info("Seeding Python random module with supplied seed %d" % (RANDOM_SEED))
+        RANDOM_SOURCE = "envar RANDOM_SEED"
+
+    log.info("Seeding random generators with seed {} from {}".format(RANDOM_SEED, RANDOM_SOURCE))
     random.seed(RANDOM_SEED)
+    simulator.set_random_seed(RANDOM_SEED)
 
     # Setup DUT object
     from cocotb import simulator
--- a/cocotb/clock.py
+++ b/cocotb/clock.py
@@ -28,11 +28,16 @@
 """A clock class."""
 
 import itertools
+from typing import Union, Tuple, Optional
 import warnings
 
+import cocotb
 from cocotb.log import SimLog
 from cocotb.triggers import Timer
 from cocotb.utils import get_sim_steps, get_time_from_sim_steps, lazy_property
+from cocotb.handle import SimHandleBase
+
+_logger = SimLog(__name__)
 
 
 class BaseClock:
@@ -165,3 +170,170 @@ class Clock(BaseClock):
 
     def __str__(self):
         return type(self).__qualname__ + "(%3.1f MHz)" % self.frequency
+
+
+class CClock:
+    r"""High performance, configurable simulator clock driver.
+
+    This class is a wrapper for a C++ layer clock driver with methods for control.
+    Use :meth:`start` to start the clock driver and :meth:`stop` to stop it.
+    The :attr:`is_running` property returns whether the driver is active.
+
+    Example:
+
+    .. code-block:: python
+
+        # 0.5 duty cycle by default
+        c = CClock(dut.clk, 100, units='ns')
+        c.start()
+
+    Args:
+        signal: The clock signal to be driven.
+        period: The clock period in *units*.
+        units: One of
+            ``'step'``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
+            When *units* is ``'step'``,
+            the timestep is determined by the simulator (see :make:var:`COCOTB_HDL_TIMEPRECISION`).
+        duty_cycle: The clock duty cycle as a float in [0, 1].
+            Duty cycle is defined as the fraction of the period where the clock is high.
+            Default is 0.5.
+
+            .. note::
+                Duty cycles of 0 or 1 may cause glitches or unexpected simulator behavior.
+
+        jitter: Maximum jitter for one clock period in *units*.
+            Default is 0.
+
+    Raises:
+        ValueError: If duty cycle value is not in [0, 1].
+
+    If clock jitter is non-zero, random edge-to-edge clock jitter will be generated
+    using a normal (gaussian) distribution for the jitter values.
+    With a clock period defined as one rising edge to the next rising edge, the jitter
+    will be bounded each clock period by ``[-jitter, +jitter]``.
+
+    Jitter example:
+
+    .. code-block:: python
+
+        # 1MHz clock with 10ns of jitter
+        c = CClock(dut.clk, 1000, units='ns', jitter=10)
+        c.start()
+
+    .. versionadded:: 1.5
+    """
+    def __init__(
+        self, signal: SimHandleBase,
+        period: Union[int, float],
+        units: str = 'step',
+        duty_cycle: Optional[float] = 0.5,
+        jitter: Union[int, float] = 0
+    ):
+        self.signal = signal
+
+        if not 0.0 <= duty_cycle <= 1.0:
+            raise ValueError("Duty cycle must be in range [0, 1]")
+
+        if duty_cycle == 0.0 or duty_cycle == 1.0:
+            warnings.warn("Duty cycles of 0 or 1 may cause glitches or unexpected simulator behavior",
+                          category=RuntimeWarning)
+
+        period_steps = get_sim_steps(period, units)
+
+        high_steps = round(period_steps * duty_cycle)
+        low_steps = period_steps - high_steps
+
+        # Divide jitter budget between high and low, scaled to duty cycle
+        jitter_steps = get_sim_steps(jitter, units)
+        high_jitter = round(jitter_steps * duty_cycle)
+        low_jitter = jitter_steps - high_jitter
+
+        _logger.debug("Creating simulator.sim_clock with ({}, {}, {}, {}, {})".format(
+            self.signal._name, high_steps, low_steps, high_jitter, low_jitter))
+
+        self._sim_clk = cocotb.simulator.create_clock(
+            self.signal._handle, high_steps, low_steps,
+            high_jitter, low_jitter)
+
+        # All simulator clocks must be tracked by the scheduler for cleanup between tests
+        cocotb.scheduler._add_sim_clock(self._sim_clk)
+
+    @classmethod
+    def from_period_tuple(
+        cls, signal: SimHandleBase,
+        period: Tuple[Union[int, float], Union[int, float]],
+        units: str = 'step',
+        jitter: Union[int, float] = 0
+    ):
+        """Create CClock by using separate high and low clock times.
+
+        Example:
+
+        .. code-block:: python
+
+            c = CClock.from_period_tuple(dut.clk, (60, 40), units='ns')
+            c.start()
+
+        Args:
+            signal: The clock signal to be driven.
+            period: The clock period as a 2-tuple of (high_time, low_time) in *units*.
+            units: One of
+                ``'step'``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
+                When *units* is ``'step'``,
+                the timestep is determined by the simulator (see :make:var:`COCOTB_HDL_TIMEPRECISION`).
+            jitter: Maximum jitter for one clock period in *units*.
+                Default is 0.
+        """
+        if not isinstance(period, tuple):
+            raise TypeError("Clock period must be a 2-tuple")
+
+        if len(period) != 2:
+            raise ValueError("Clock period tuple must have 2 members")
+
+        high_steps = get_sim_steps(period[0], units)
+        low_steps = get_sim_steps(period[1], units)
+        period = high_steps + low_steps
+
+        duty_cycle = high_steps / period
+
+        return cls(signal, period, 'step', duty_cycle, jitter)
+
+    def start(self, *, periods: Optional[int] = None, posedge_first: bool = True):
+        """Start driving the clock.
+
+        Args:
+            periods: Drive the clock for this many periods,
+                or if ``None`` then drive the clock forever.
+
+                .. note::
+                    ``0`` is not the same as ``None``, as ``0`` will not start the clock.
+
+            posedge_first: If ``True``, start driving the clock with the
+                positive-going edge at the start of the high part of the period.
+                If ``False``, start driving with the negative-going edge.
+                This is not the same as inverting the clock.
+                Default is ``True``.
+
+        Raises:
+            ValueError: If *periods* is negative.
+        """
+        if periods == 0:
+            return
+
+        if periods is None:
+            toggles = 0
+        else:
+            if periods < 0:
+                raise ValueError("Periods value must be a non-negative integer")
+            toggles = periods * 2
+
+        self._sim_clk.start(toggles, posedge_first)
+
+    def stop(self):
+        """Stop driving the clock."""
+        self._sim_clk.stop()
+
+    @property
+    def is_running(self) -> bool:
+        """Active status of the clock driver."""
+        return self._sim_clk.is_running()
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -251,6 +251,8 @@ class Scheduler:
         self._write_coro_inst = None
         self._writes_pending = Event()
 
+        self._simulator_clocks = set()
+
     async def _do_writes(self):
         """ An internal coroutine that performs pending writes """
         while True:
@@ -851,6 +853,10 @@ class Scheduler:
             while self._pending_coros:
                 self.add(self._pending_coros.pop(0))
 
+    def _add_sim_clock(self, clock):
+        """Keep track of simulator clocks for cleanup between tests."""
+        self._simulator_clocks.add(clock)
+
     def finish_test(self, exc):
         self._test.abort(exc)
         self._check_termination()
@@ -882,6 +888,15 @@ class Scheduler:
                 if _debug:
                     self.log.debug("Killing %s" % str(coro))
                 coro.kill()
+
+        # Stop and delete all simulator clocks
+        # TODO: Move clock management to tasks once a task can cleanup on kill/cancel
+        while (self._simulator_clocks):
+            clock = self._simulator_clocks.pop()
+            if _debug:
+                self.log.debug("Stopping simulator clock {!r}".format(clock))
+            clock.stop()
+            del clock
 
         if self._main_thread is not threading.current_thread():
             raise Exception("Cleanup() called outside of the main thread")
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -40,6 +40,8 @@ static int releases = 0;
 static int sim_ending = 0;
 
 #include <cocotb_utils.h>     // COCOTB_UNUSED
+#include <chrono>
+#include <random>
 #include <type_traits>
 #include <Python.h>
 #include "gpi_logging.h"
@@ -61,6 +63,11 @@ struct callback_data {
     PyObject *kwargs;                   // Keyword arguments to call the function with
     gpi_sim_hdl cb_hdl;
 };
+
+// Simulator clock
+struct sim_clock_data;
+typedef struct sim_clock_data *sim_clock;
+static std::default_random_engine rand_gen{};
 
 /* define the extension types as templates */
 namespace {
@@ -148,6 +155,9 @@ namespace {
     PyTypeObject gpi_hdl_Object<gpi_iterator_hdl>::py_type;
     template<>
     PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type;
+    template<>
+    PyTypeObject gpi_hdl_Object<sim_clock>::py_type;
+
 }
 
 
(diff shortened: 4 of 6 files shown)


cocotb/__init__.py
```diff
@@ -241,6 +241,7 @@ def _initialise_testbench(argv_):
         RANDOM_SEED = int(RANDOM_SEED)
         log.info("Seeding Python random module with supplied seed %d" % (RANDOM_SEED))
```
cmarqu, line 242:
```suggestion
        log.info("Seeding cocotb's Python and GPI parts with supplied seed %d" % (RANDOM_SEED))
```
?

cocotb/clock.py
```diff
@@ -165,3 +168,134 @@ async def start(self, cycles=None, start_high=True):
...
+
+    Args:
+        signal: The clock pin/signal to be driven.
+        period (int, float, 2-tuple): The clock period. Must convert to an even number of
+            timesteps. Provide a 2-tuple as (high_time, low_time) to set the duty cycle manually.
+        units (str, optional): One of
+            ``'step'``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
+            When *units* is ``'step'`` (the default),
+            the timestep is determined by the simulator (see :make:var:`COCOTB_HDL_TIMEPRECISION`).
+        duty_cycle (float, optional): The clock duty cycle as a float in [0, 1]. Default is 0.5.
```
cmarqu, line 196: Is this a closed interval by design, i.e. `0.0` would produce a constant-0 (or, in case there is a `start_value` later, a constant-`start_value`)?

cocotb/clock.py
```diff
@@ -165,3 +168,134 @@ async def start(self, cycles=None, start_high=True):
...
+    .. code-block:: python
+
+        # defaults to 50:50 duty cycle
+        c = SimClock(dut.clk, 100, units='ns')
+        c.start()
+
+    Args:
+        signal: The clock pin/signal to be driven.
+        period (int, float, 2-tuple): The clock period. Must convert to an even number of
+            timesteps. Provide a 2-tuple as (high_time, low_time) to set the duty cycle manually.
```
cmarqu, line 191:
```suggestion
            timesteps. Provide a 2-tuple as (start_value_time, end_value_time) to set the duty cycle manually.
```
(to account for `start_high`)?

cocotb/clock.py
```diff
@@ -165,3 +168,134 @@ async def start(self, cycles=None, start_high=True):
...
+        else:
+            if duty_cycle is None:
+                duty_cycle = 0.5
+
+            if not 0.0 <= duty_cycle <= 1.0:
+                raise ValueError("Duty cycle must be in range [0, 1]")
+
+            # Don't bother creating a clock with glitches if duty cycle is 0 or 1
+            if duty_cycle == 0.0 or duty_cycle == 1.0:
+                warnings.warn(r"Duty cycles of 0% or 100% may cause glitches or unexpected behavior", category=RuntimeWarning)
```
cmarqu, line 246: Maybe this info should also go into the class docstring.

cocotb/clock.py
```diff
@@ -165,3 +168,134 @@ async def start(self, cycles=None, start_high=True):
...
+
+    def start(self, *, cycles: Optional[int] = None, start_high: bool = True):
+        r"""Start driving the clock.
+
+        Args:
+            cycles (int, optional): Cycle the clock *cycles* number of times,
+                or if ``None`` then cycle the clock forever.
+                Note: ``0`` is not the same as ``None``, as ``0`` will cycle no times.
+            start_high (bool, optional): Whether to start the clock with a ``1``
+                for the first half of the period.
```
cmarqu, line 276:
```suggestion
                for the first part of the period.
```

cocotb/clock.py
```diff
@@ -165,3 +168,134 @@ async def start(self, cycles=None, start_high=True):
...
+            high_jitter, low_jitter)
+
+        # All simulator clocks must be tracked by the scheduler for cleanup between tests
+        cocotb.scheduler._add_sim_clock(self.sim_clk)
+
+    def start(self, *, cycles: Optional[int] = None, start_high: bool = True):
+        r"""Start driving the clock.
+
+        Args:
+            cycles (int, optional): Cycle the clock *cycles* number of times,
```
cmarqu, line 272: I forget... With type annotations, can the `(int, optional)` go now maybe?

cocotb/clock.py
```diff
@@ -165,3 +168,134 @@ async def start(self, cycles=None, start_high=True):
...
+    ):
+        super().__init__(signal)
+        if isinstance(period, tuple):
+            if len(period) != 2:
+                raise ValueError("Clock period tuple must have 2 members")
+            if duty_cycle is not None:
+                warnings.warn("Duty cycle cannot be used with clock period tuple. "
+                              "Duty cycle value will be ignored", category=RuntimeWarning)
+
+            high_steps = get_sim_steps(period[0], units)
```
cmarqu, line 232:
```suggestion
            # this naming applies to the default `start_high=True`
            high_steps = get_sim_steps(period[0], units)
```
I guess internally, we can normalize the wording like here, but toward the user, we may want to speak of "first part of the period" etc.?

cocotb/clock.py
```diff
@@ -165,3 +168,134 @@ async def start(self, cycles=None, start_high=True):
...
+            duty_cycle = high_steps / (high_steps + low_steps)
+
+        else:
+            if duty_cycle is None:
+                duty_cycle = 0.5
+
+            if not 0.0 <= duty_cycle <= 1.0:
+                raise ValueError("Duty cycle must be in range [0, 1]")
+
+            # Don't bother creating a clock with glitches if duty cycle is 0 or 1
```
cmarqu, line 244: This comment doesn't seem to fit the implementation, does it?

cocotb/clock.py
```diff
@@ -165,3 +168,134 @@ async def start(self, cycles=None, start_high=True):
...
+
+        self.sim_clk = cocotb.simulator.create_clock(
+            self.signal._handle, high_steps, low_steps,
+            high_jitter, low_jitter)
+
+        # All simulator clocks must be tracked by the scheduler for cleanup between tests
+        cocotb.scheduler._add_sim_clock(self.sim_clk)
+
+    def start(self, *, cycles: Optional[int] = None, start_high: bool = True):
+        r"""Start driving the clock.
```
cmarqu, line 269: Doesn't look like the `r"""` is needed.

tests/test_cases/test_cocotb/test_clock.py
```diff
@@ -68,3 +74,277 @@ async def test_external_clock(dut):
...
+    for period in period_values:
+        assert period >= get_sim_steps(970, units='ns'), "{}".format(period)
+        assert period <= get_sim_steps(1030, units='ns'), "{}".format(period)
+
+    # Negative 3-sigma to 3-sigma
+    jitter_dist = [[] for _ in range(6)]
+
+    for jitter in jitter_values:
+        jitter_ns = get_time_from_sim_steps(jitter, 'ns')
+        for sigma in range(-3,3):
```
cmarqu, line 155:
```suggestion
        for sigma in range(-3, 3):
```

cocotb/clock.py
```diff
@@ -165,3 +168,134 @@ async def start(self, cycles=None, start_high=True):
...
+        c.start()
+
+    Raises:
+        ValueError: If tuple is used for clock period and not a 2-tuple,
+            or duty cycle value is not in [0, 1].
+
+    .. versionadded:: 1.5
+    """
+    def __init__(
+        self, signal: SimHandleBase,
```
themperek, line 219: Should we have also? :
- start_time
- posedge_first

ktbarrett (commented):
I much prefer the `.start()` `.stop()` semantics to what the current `Clock` does. I wouldn't bother inheriting from `BaseClock`, it doesn't offer anything useful (not everything needs a logger!); and it says nothing about a "Clock-Driver-like" interface.

cocotb/scheduler.py
```diff
@@ -251,6 +251,8 @@ def __init__(self):
         self._write_coro_inst = None
         self._writes_pending = Event()
 
+        self._simulator_clocks = set()
```
ktbarrett, line 254: Hmmm, maybe we need a task-like object to represent an independent external process so we don't have to special case every time we add a new feature to C. Could this be modeled with `cocotb.external`?

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -912,6 +921,24 @@ static PyObject *log_level(PyObject *self, PyObject *args)
     Py_RETURN_NONE;
 }
 
+static PyObject *set_random_seed(PyObject *self, PyObject *args)
+{
+    COCOTB_UNUSED(self);
```
ktbarrett, line 926:
```suggestion
static PyObject *set_random_seed(PyObject *, PyObject *args)
{
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1325,3 +1386,204 @@ PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
...
+
+  // Jitter support
+  std::normal_distribution<float> *gaussian;
+  int64_t high_jitter;
+  int64_t low_jitter;
+};
+
+static PyObject *create_clock(PyObject *self, PyObject *args, PyObject *keywds)
+{
+  COCOTB_UNUSED(self);
```
ktbarrett, line 1412:
```suggestion
static PyObject *create_clock(PyObject *, PyObject *args, PyObject *keywds)
{
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1325,3 +1386,204 @@ PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
...
+  if (high_jitter < 0) {
+    PyErr_SetString(PyExc_ValueError, "high_jitter value must be a non-negative integer");
+    return NULL;
+  }
+  if (low_jitter < 0) {
+    PyErr_SetString(PyExc_ValueError, "low_jitter value must be a non-negative integer");
+    return NULL;
+  }
+
+  sim_clock clock = new sim_clock_data{};
```
ktbarrett, line 1450: This leaks. Is there anyway to tie it's lifetime to the lifetime of the Python object returned?

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1325,3 +1386,204 @@ PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
...
+  if (clock->timed_callback) gpi_deregister_callback(clock->timed_callback);
+  clock->timed_callback = nullptr;
+
+  auto set_value = clock->next_value;
+  gpi_set_signal_value_long(clock->signal, clock->next_value, GPI_DEPOSIT);
+
+  if (clock->toggles == LAST_TOGGLE) {
+    clock->toggles = TOGGLE_ALWAYS;
+    return 0;
+  }
```
ktbarrett, line 1483: What is this doing?

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1325,3 +1386,204 @@ PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
...
+    clock->low_jitter = static_cast<int64_t>(low_jitter);
+  }
+
+  PyObject *new_clk = gpi_hdl_New(clock);
+
+  return new_clk;
+}
+
+static const uint64_t TOGGLE_ALWAYS = 0;
+static const uint64_t LAST_TOGGLE = 1;
```
ktbarrett, line 1468: I might even move these to be static constexpr member variables in the sim_clock_data struct.
```suggestion
constexpr uint64_t TOGGLE_ALWAYS = 0;
constexpr uint64_t LAST_TOGGLE = 1;
```

cocotb/share/lib/simulator/simulatormodule.cpp
ktbarrett: It looks like when the toggles reach a minimum they go into the TOGGLE_ALWAYS state and continue to run? Am I missing something?

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1325,3 +1386,204 @@ PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
...
+  }
+
+  clock->timed_callback = gpi_register_timed_callback((gpi_function_t)clock_callback, user_data, static_cast<uint64_t>(steps));
+
+  return 0;
+}
+
+static PyObject *sim_clock_stop(gpi_hdl_Object<sim_clock> *self, PyObject *args)
+{
+  COCOTB_UNUSED(args);
```
ktbarrett, line 1508:
```suggestion
static PyObject *sim_clock_stop(gpi_hdl_Object<sim_clock> *self, PyObject *)
{
```

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1325,3 +1386,204 @@ PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
...
+  clock_callback((void *)self->hdl);
+
+  // Set toggles here so it's not decremented in callback
+  self->hdl->toggles = static_cast<uint64_t>(toggles);
+
+  Py_RETURN_NONE;
+}
+
+static PyObject *sim_clock_is_running(gpi_hdl_Object<sim_clock> *self, PyObject *args) {
+  COCOTB_UNUSED(args);
```
ktbarrett, line 1546:
```suggestion
static PyObject *sim_clock_is_running(gpi_hdl_Object<sim_clock> *self, PyObject *) {
```

cocotb/clock.py
garmin-mjames: The tuple defines the clock period, length of high part followed by length of low part.
It's analogous to duty cycle, which defines the fraction of the period where the clock is high.

My understanding of `start_high` is that it sets which part of the clock period to start from, it doesn't change how the duty cycle is interpreted.

cocotb/clock.py
garmin-mjames: I'm not sure if we should enforce at least 1 time step for high/low parts.
Icarus and Verilator work with 0/1, where you just get a 0-time glitch every period interval. The clock still toggles, it just toggles back in the same time step. Looking at the CI, Riviera may be having issues with that.

cocotb/clock.py
garmin-mjames: See my comment above. High is always the first part of the defined period, `start_high` just decides if we start with the high part or start with the low part.

cocotb/clock.py
garmin-mjames: Correct. Originally I was going to not allow a 0 or 1 duty cycle.

cocotb/clock.py
garmin-mjames: Those are associated with `start()`. For **start_time**, just use a `Timer` trigger before starting. **posedge_first** is `start_high`.

cocotb/share/lib/simulator/simulatormodule.cpp
garmin-mjames: See `sim_clock_dealloc()`.

cocotb/clock.py
ktbarrett: One thing I do and I think is sort of common is: support only one way to construct an object, and support conveniences as named constructors using `@classmethod`.

cocotb/share/lib/simulator/simulatormodule.cpp
garmin-mjames: It returns before registering the next callback, so the clock stops. It's not actually necessary to reset toggles here, as it's set in `sim_clock_start()` every time the clock is started.

cocotb/scheduler.py
garmin-mjames: I'll look at `external`, but I'm not sure there's a way to signal it in order to have it stop the clock. We're missing something like #1349 for resource cleanup.

ktbarrett: Right... That's why I mentioned task-like. Something we can add to the scheduler that looks like a task and waits on a trigger that will never fire (`InfinityTrigger`?), that is `.kill`ed at the end of the test. Tasks don't have a well-defined interface according to how the scheduler uses them, do they?

ktbarrett: So we can support the current `Clock` interface by using `fork(SimClock(...).start())` like we do now.

Or we can do something completely different! I'm just throwing out ideas, it's your effort. I'm just here to make sure you dot your ı's, errr... i's.

cocotb/clock.py
garmin-mjames: Which one should be the default?

cmarqu: > My understanding of `start_high` is that it sets which part of the clock period to start from, it doesn't change how the duty cycle is interpreted or invert the clock.

Oh, interesting... it never occurred to me that it could be viewed like this, as that's not what I had in mind when I introduced `start_high` back then. I saw it as "invert the clock", but since there is no commonly agreed upon "normal"/"non-inverted" clock, I made it explicit in the name.

Maybe we should have a drawing in the docstring, like... (I need to change to a real computer for that)
```
         ┌─────────
─────────┘          
```

cocotb/clock.py
cmarqu: Haha, I can already see someone trying to misuse that as an equivalent to a Verilog event. :)

cocotb/clock.py
garmin-mjames: I can use @themperek's `posedge_first` instead. I'll add clarification in the docstring as well.

cocotb/scheduler.py
garmin-mjames: There's also no way currently to notify the scheduler or any task that a clock using `cycles` has finished toggling and stopped.

ktbarrett: Well it is cleaned up by the test end. Users have no way of knowing when the clock stopped either, which is the bigger issue I'd think.

garmin-mjames: What about:
```py3
clock_coro = cocotb.fork(Clock(100, units='ns').start(cycles=15))
await Join(clock_coro)
```
Someone somewhere might be using that.

ktbarrett: Hmmmmm... The trigger would have to be set in C++.

I definitely don't think it's a good idea right now. Just getting a feel for what would have to be added to make that work. I have a feeling this isn't the first of these types of changes and having that functionality there would make adding similar features much easier in the future.

garmin-mjames: Something like a `GPIEvent` trigger. It should be safe to call `Scheduler.react` (with GIL), as the C++ callback is just a timed callback like `Timer`.

That can wait for when `Clock` is moved to use this as a backend.
If we make that opt-in, we could probably get away with not supporting this at all and force users to poll `is_running` manually if they want to know when the clock stops.

cocotb/__init__.py
garmin-mjames: I changed the log message here to reflect source of the seed.

alexforencich:
A couple of things that I would like to see:

1. Phase parameter.  Yes, this is effectively just a start delay, but computed based on the requested period.  This makes it much eaiser to set up clocks that have fixed phase relationships, but with possibly variable frequency.  
2. The ability to change the clock period and possibly phase and duty cycle on the fly.  Right now, it looks like the clock has to be stopped, deleted, and recreated.  Seems like it should be possible to have an update() call that takes one or more kwargs, then this will internally fork a coroutine to wait for the next edge and restart the clock with the appropriate changes applied.  For multiphase clocks, there may be something to be said for creating a container object that can manage all of the clocks together to maintain the phase relationship across period updates.  
3. Automatic start.  Why does `start()` need to be called for a free-running clock?  Perhaps add a `start_delay` parameter to the constructor, default `0`.  If set to `None`, then let the user start the clock manually by calling `start()`, otherwise fork a coroutine to start the clock with the specified delay.  Probably need to do something like this to implement the phase parameter anyway.  

I think all of this can be done purely in Python, I don't think any changes to the GPI implementation are necessary for implementing any of this.

Additionally, with the phase parameter, perhaps 'start_high' is no longer necessary...just specify a phase of 180 degrees.  Or implement start_high by adding 180 degrees to the phase.  Or use start_high for .setimmediatevalue when creating the clock, and nothing else.

alexforencich:
Also, if given a signal that has more than one bit, I think it should drive all of the bits instead of just the LSB.

ktbarrett:
@alexforencich We really only should encapsulate the behavior that cannot be expressed separately. This PR is mainly focused on adding the basic features for running a clock from C. Phase and changing clock attributes can be handled separately; duty cycle and jitter cannot.

marlonjames:
Closing this. It was decided to move this to a GPI extension initially, after 1.5 release.

PR #2297, opened 2020-12-18
Add asyncio-style Queue, PriorityQueue, and LifoQueue

alexforencich:
This commit adds three asyncio-style queues for communication between coroutines - Queue, PriorityQueue, and LifoQueue.

### DIFF
--- a/cocotb/queue.py
+++ b/cocotb/queue.py
@@ -1,0 +1,148 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+import collections
+import heapq
+from asyncio import QueueEmpty, QueueFull
+
+import cocotb
+from cocotb.triggers import Event
+
+
+class Queue:
+    def __init__(self, maxsize=0):
+
+        self._maxsize = maxsize
+
+        self._finished = Event()
+        self._finished.set()
+
+        self._getters = collections.deque()
+        self._putters = collections.deque()
+
+        self._unfinished_tasks = 0
+
+        self._init(maxsize)
+
+    def _init(self, maxsize):
+        self._queue = collections.deque()
+
+    def _put(self, item):
+        self._queue.append(item)
+
+    def _get(self):
+        return self._queue.popleft()
+
+    def _wakeup_next(self, waiters):
+        while waiters:
+            event, task = waiters.popleft()
+            if not task._finished:
+                event.set()
+                break
+
+    def __repr__(self):
+        return '<{} at {:#x} {}>'.format(type(self).__name__, id(self), self._format())
+
+    def __str__(self):
+        return '<{} {}>'.format(type(self).__name__, self._format())
+
+    def __class_getitem__(cls, type):
+        return cls
+
+    def _format(self):
+        result = 'maxsize={}'.format(repr(self._maxsize))
+        if getattr(self, '_queue', None):
+            result += ' _queue={}'.format(repr(list(self._queue)))
+        if self._getters:
+            result += ' _getters[{}]'.format(len(self._getters))
+        if self._putters:
+            result += ' _putters[{}]'.format(len(self._putters))
+        if self._unfinished_tasks:
+            result += ' tasks={}'.format(self._unfinished_tasks)
+        return result
+
+    def qsize(self):
+        return len(self._queue)
+
+    @property
+    def maxsize(self):
+        return self._maxsize
+
+    def empty(self):
+        return not self._queue
+
+    def full(self):
+        if self._maxsize <= 0:
+            return False
+        else:
+            return self.qsize() >= self._maxsize
+
+    async def put(self, item):
+        while self.full():
+            event = Event('{} put'.format(type(self).__name__))
+            self._putters.append((event, cocotb.scheduler._current_task))
+            await event.wait()
+        self.put_nowait(item)
+
+    def put_nowait(self, item):
+        if self.full():
+            raise QueueFull()
+        self._put(item)
+        self._unfinished_tasks += 1
+        self._finished.clear()
+        self._wakeup_next(self._getters)
+
+    async def get(self):
+        while self.empty():
+            event = Event('{} get'.format(type(self).__name__))
+            self._getters.append((event, cocotb.scheduler._current_task))
+            await event.wait()
+        return self.get_nowait()
+
+    def get_nowait(self):
+        if self.empty():
+            raise QueueEmpty()
+        item = self._get()
+        self._wakeup_next(self._putters)
+        return item
+
+    def task_done(self):
+        if self._unfinished_tasks <= 0:
+            raise ValueError("task_done() called with no outstanding tasks")
+        self._unfinished_tasks -= 1
+        if self._unfinished_tasks == 0:
+            self._finished.set()
+
+    async def join(self):
+        if self._unfinished_tasks > 0:
+            await self._finished.wait()
+
+
+class PriorityQueue(Queue):
+    """A subclass of Queue; retrieves entries in priority order (smallest item first).
+
+    Entries are typically tuples of the form: (priority number, data).
+    """
+
+    def _init(self, maxsize):
+        self._queue = []
+
+    def _put(self, item):
+        heapq.heappush(self._queue, item)
+
+    def _get(self):
+        return heapq.heappop(self._queue)
+
+
+class LifoQueue(Queue):
+    """A subclass of Queue; retrieves most recently added entries first."""
+
+    def _init(self, maxsize):
+        self._queue = collections.deque()
+
+    def _put(self, item):
+        self._queue.append(item)
+
+    def _get(self):
+        return self._queue.pop()
--- a/tests/test_cases/test_cocotb/test_queues.py
+++ b/tests/test_cases/test_cocotb/test_queues.py
@@ -1,0 +1,279 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+"""
+Tests relating to cocotb.queue.Queue, cocotb.queue.LifoQueue, cocotb.queue.PriorityQueue
+"""
+import cocotb
+from cocotb.queue import Queue, PriorityQueue, LifoQueue, QueueFull, QueueEmpty
+from cocotb.regression import TestFactory
+from cocotb.triggers import Combine, Timer
+
+
+async def run_queue_nonblocking_test(dut, queue_type):
+    QUEUE_SIZE=10
+
+    q = queue_type(maxsize=QUEUE_SIZE)
+
+    # queue empty
+    assert q.maxsize == QUEUE_SIZE
+    assert q.qsize() == 0
+    assert q.empty()
+    assert not q.full()
+
+    assert str(q) == '<{} maxsize={}>'.format(type(q).__name__, q.maxsize)
+    assert isinstance(repr(q), str)
+
+    # put one item
+    q.put_nowait(0)
+
+    assert q.qsize() == 1
+    assert not q.empty()
+    assert not q.full()
+
+    # fill queue
+    if queue_type is PriorityQueue:
+        for k in range(QUEUE_SIZE-1, 0, -1):
+            q.put_nowait(k)
+    else:
+        for k in range(1, QUEUE_SIZE):
+            q.put_nowait(k)
+
+    assert q.qsize() == QUEUE_SIZE
+    assert not q.empty()
+    assert q.full()
+
+    # overflow
+    got_exception = False
+    try:
+        q.put_nowait(100)
+    except QueueFull:
+        got_exception = True
+    assert got_exception
+
+    # check queue contents
+    if queue_type is LifoQueue:
+        for k in range(QUEUE_SIZE-1, -1, -1):
+            assert q.get_nowait() == k
+    else:
+        for k in range(QUEUE_SIZE):
+            assert q.get_nowait() == k
+
+    assert q.qsize() == 0
+    assert q.empty()
+    assert not q.full()
+
+    # underflow
+    got_exception = False
+    try:
+        q.get_nowait()
+    except QueueEmpty:
+        got_exception = True
+    assert got_exception
+
+    # task counts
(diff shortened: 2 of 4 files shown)


cocotb/queue.py
```diff
@@ -0,0 +1,179 @@
...
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL POTENTIAL VENTURES LTD BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
cmarqu, line 25:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause
```
New files can get this SPDX header, we just don't touch existing files.

cocotb/queue.py
```diff
@@ -0,0 +1,179 @@
...
+
+import collections
+import heapq
+
+import cocotb
+from cocotb.triggers import Event
+
+
+class QueueEmpty(Exception):
+    """Raised when Queue.get_nowait() is called on an empty Queue."""
```
cmarqu, line 35: Should use markup that creates cross-references, but no need to do this in this early stage.
```suggestion
    """Raised when :meth:`Queue.get_nowait()` is called on an empty :class:`Queue`."""
```
(not sure if this works, maybe
```
:class:`~cocotb.queue.Queue`
```
 is needed.)

cocotb/queue.py
```diff
@@ -0,0 +1,179 @@
...
+        if self._unfinished_tasks == 0:
+            self._finished.set()
+
+    async def join(self):
+        if self._unfinished_tasks > 0:
+            await self._finished.wait()
+
+
+class PriorityQueue(Queue):
+    """A subclass of Queue; retrieves entries in priority order (lowest first).
```
cmarqu, line 154:
```suggestion
    """A subclass of Queue; retrieves entries in priority order (lowest number first).
```
(because "lowest number --> highest priority"?)

alexforencich: It wraps heapq: https://docs.python.org/3/library/heapq.html .  It seems that heapq implements a min heap, so it returns the lowest value first.  Not sure if describing it with "number" makes the most sense, maybe "value" would be better.  I think it works the same as `sort()`, so it works with strings, tuples, and other objects.

cmarqu: They say "smallest item", how about this?

alexforencich: sure, that sounds reasonable

cocotb/queue.py
```diff
@@ -0,0 +1,179 @@
...
+    """Raised when Queue.get_nowait() is called on an empty Queue."""
+    pass
+
+
+class QueueFull(Exception):
+    """Raised when the Queue.put_nowait() method is called on a full Queue."""
+    pass
+
+
+class Queue(object):
```
eric-wieser, line 44: No need for `(object)` now that Python 2 is dead
```suggestion
class Queue:
```

eric-wieser (commented):
I think this could do with a doc comment saying the API is designed to be as similar as possible to asyncio.queues.Queue

tests/test_cases/test_cocotb/test_queues.py
```diff
@@ -0,0 +1,191 @@
...
+    cr.kill()
+    cr_list.append(cocotb.fork(putter(queue, 101)))
+
+    for k in range(20):
+        cr_list.append(cocotb.fork(getter(queue, k)))
+
+    cr_list.append(cocotb.fork(getter(queue, 101)))
+
+    while cr_list:
+        await cr_list.popleft().join()
```
ktbarrett, line 121:
```suggestion
    await Combine(*cr_list)
```
You don't really need the `join()` either.

ktbarrett (commented):
What is the practical use case of `task_done()`? I don't get it.

There is a missing test for `__str__` and `__repr__`. I don't think you need to test the content, just make sure they don't blow up :) so we get coverage on them.

One important test that is missing is one for fair scheduling. The whole point of this implementation is to enforce fair scheduling when there are multiple readers or writers. There are much simpler implementations if we don't care about that, so lets test it to ensure we are guaranteeing that behavior going forward.

And you probably already know, but you should add a section to the Python library reference page on this module and the classes. It should be a reference, we can add tutorials for using it elsewhere.

tests/test_cases/test_cocotb/test_queues.py
```diff
@@ -0,0 +1,191 @@
...
+    assert not queue.empty()
+    assert queue.full()
+
+    # overflow
+    got_exception = False
+    try:
+        queue.put_nowait(100)
+    except QueueFull:
+        got_exception = True
+    assert got_exception
```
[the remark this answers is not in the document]
eric-wieser: `pytest.raises` does not throw `AssertionError`, but some other pytest object. Perhaps we don't care about the difference, but its there.

cocotb/queue.py
```diff
@@ -0,0 +1,179 @@
...
+    def __str__(self):
+        return '<{} {}>'.format(type(self).__name__, self._format())
+
+    def __class_getitem__(cls, type):
+        return cls
+
+    def _format(self):
+        result = 'maxsize={}'.format(repr(self._maxsize))
+        if getattr(self, '_queue', None):
+            result += ' _queue={}'.format(repr(list(self._queue)))
```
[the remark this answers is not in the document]
eric-wieser: This code is copied verbatim from `asyncio`, I'd be inclined to leave it.

alexforencich:
So, `task_done()` is a feature of the asyncio queues, so I implemented it for consistency.  The idea is simple: if you have one "dispatcher" coroutine, it can `put` a bunch of stuff into the queue, and then await `queue.join()`.  The worker coroutines can `get` items to process.  When each worker finishes processing its item, it calls `task_done()`.  When the number of calls to `task_done()` equal the number of calls to `put()`, `queue.join()` returns and the dispatcher coroutine starts running again.  I should add a test case for that.

cocotb/queue.py
```diff
@@ -0,0 +1,179 @@
...
+        if self._getters:
+            result += ' _getters[{}]'.format(len(self._getters))
+        if self._putters:
+            result += ' _putters[{}]'.format(len(self._putters))
+        if self._unfinished_tasks:
+            result += ' tasks={}'.format(self._unfinished_tasks)
+        return result
+
+    def qsize(self):
+        return len(self._queue)
```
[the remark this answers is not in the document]
alexforencich: Beats me.  We can always add that, but keep qsize to be consistent with asyncio.

cocotb/queue.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class QueueEmpty(Exception):
+    """Raised when Queue.get_nowait() is called on an empty Queue."""
+    pass
+
+
+class QueueFull(Exception):
+    """Raised when the Queue.put_nowait() method is called on a full Queue."""
+    pass
```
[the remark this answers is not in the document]
alexforencich: That's an interesting idea.  I will note that python already has a number of different exceptions for empty and full queues, including: https://github.com/python/cpython/blob/3.9/Lib/queue.py#L16-L25 .  But, if we're trying to make a drop-in replacement for the asyncio queue, it would make sense to try to use the same exceptions.

alexforencich:
And are my contention test cases not sufficient?  I set up a queue, maxsize 10, and start 20 putters, so 10 get blocked.  Then I start 20 getters that are each looking for a specific value.  Then I do the same thing again, but start the getters first so they all get blocked.  I only do that test for the normal queue, but the contention resolution code is the same for all of the queues, and the normal queue preserves the ordering so the verification is a lot easier.  Perhaps I can add a couple more queues to verify the order the getters and putters run in, in addition to the getters checking that they got the expected values.  

Now, the last test that compares blocking and nonblocking is maybe a bit useless, now that I look at it again.  But at least it runs on all of the queues.

tests/test_cases/test_cocotb/test_queues.py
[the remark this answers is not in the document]
alexforencich: Yeah, that is a big difference.  The simulation terminates if I use the pytest version.  This is probably something that deserves some additional thought; maybe there should be an equivalent of pytest.raises in regression, or maybe things need to be updated to handle the pytest version properly.  

Stack trace when using `pytest.raises`:

    ERROR: called callback function threw exception
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 323, in _test_completed
        cocotb.regression_manager.handle_result(test)
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/regression.py", line 299, in handle_result
        self._record_result(
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/regression.py", line 429, in _record_result
        test_pass, sim_failed = self._score_test(test, outcome)
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/regression.py", line 355, in _score_test
        outcome.get()
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/outcomes.py", line 53, in get
        raise self.error
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/regression.py", line 599, in _my_test
        await function(dut, *args, **kwargs)
      File "/home/alex/Projects/cocotb/tests/test_cases/test_cocotb/test_queues.py", line 48, in run_queue_nonblocking_test
        pass
      File "/usr/lib/python3.9/site-packages/_pytest/python_api.py", line 728, in __exit__
        fail(self.message)
      File "/usr/lib/python3.9/site-packages/_pytest/outcomes.py", line 153, in fail
        raise Failed(msg=msg, pytrace=pytrace)
    Failed: DID NOT RAISE <class 'asyncio.queues.QueueFull'>

cocotb/queue.py
alexforencich: Good point; force of habit.  I should probably strip that in a few other places as well (cocotbext-*)

cocotb/queue.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def full(self):
+        if self._maxsize <= 0:
+            return False
+        else:
+            return self.qsize() >= self._maxsize
+
+    async def put(self, item):
+        while self.full():
+            event = Event()
```
garmin-mjames, line 83: Having named Events would help with debug.
```suggestion
            event = Event('{} put'.format(type(self).__name__))
```

ktbarrett:
> So, `task_done()` is a feature of the asyncio queues, so I implemented it for consistency. The idea is simple: if you have one "dispatcher" coroutine, it can `put` a bunch of stuff into the queue, and then await `queue.join()`. The worker coroutines can `get` items to process. When each worker finishes processing its item, it calls `task_done()`. When the number of calls to `task_done()` equal the number of calls to `put()`, `queue.join()` returns and the dispatcher coroutine starts running again.

I see, so they decided to combine task pool features into their queue object, and you have to manually signal the worker task is done? Yuck. Of course, the only reason that interface exists is for consistency with `queue.Queue`. They are sticking features together assuming a particular use case which does not align with how it will be used in cocotb. There are no dynamic resources in cocotb, no one will be spinning up 100 worker tasks to do work concurrently returning results via an aggregation queue.

> I should add a test case for that.

Or just remove it. As long as it's covered.


> And are my contention test cases not sufficient? 

The issue occurs when you have multiple readers or writers that start their operation "at the same time" who put multiple values in the queue "immediately". The implementation you've ported gives the waiters receipts immediately which are placed in a queue and must be handled in that order. This prevents an unfair scheduler from giving the same task priority over the other, livelocking the rest. The simpler queue implementation I'm familiar with, which is designed only for creating a channel between two tasks, doesn't handle that case and will livelock.

The best way to test this is to spin up 2 writers sending 10 pieces of data each (`maxsize=1`). Have them contend for write with no wait between each write. Ensure the reader reads messages from each in a fair order: 1, 2, 1, 2, 1, 2, etc.

ktbarrett:
@alexforencich A test for fair scheduling.

```python
@cocotb.test()
async def fair_scheduling(dut, queue_type, NUM_PUTTERS, NUM_PUTS):

    q = Queue(maxsize=1)

    async def putter(i):
        for _ in range(NUM_PUTS):
            await q.put(i)

    # fill queue to force contention
    q.put_nowait(None)

    # create NUM_PUTTER contending putters
    putters = [cocotb.fork(putter(i)) for i in range(NUM_PUTTERS)]

    # remove value that forced contention
    q.get_nowait()

    # test fair scheduling by ensuring that each putter is serviced for it's first
    # write before the second write on any putter is serviced.
    for _ in range(NUM_PUTS):
        remaining = set(range(NUM_PUTTERS))
        for _ in range(NUM_PUTTERS):
            v = await q.get()
            assert v in remaining, "Unfair scheduling occured"
            remaining.remove(v)

    assert all(not p for p in putters), "Not all putters finished?"
```

ktbarrett (commented):
What's left? Let track it.

- [ ] Class documentation. You can probably copy [asynco.Queue](https://docs.python.org/3/library/asyncio-queue.html)'s docs verbatim. Striping out asyncio-isms like the loop parameter.
- [ ] ReadTheDocs. You can just add it to the refrence manual for now and we can work out an example and how to talk about it later on.

ktbarrett:
See #2404 for the addition of type annotations, docstrings, and RTD documentation.

ktbarrett:
Closing in favor of #2408.

PR #2298, opened 2020-12-18
Reintroduce NextTimeStep() in Avalon bus driver

marcinwoj:
The NextTimeStep() was introduced here 793fb695c7f5cf0372f7dee0c7d617b108a685a3
in order to fix the attempted writing during a
read-only phase and was removed here a61c371be6608f111c10a9550b8a274b02bb80a2
It looks like it is still needed as it's braking
the ping tup tap example.

### DIFF
--- a/cocotb/drivers/avalon.py
+++ b/cocotb/drivers/avalon.py
@@ -684,6 +684,7 @@ class AvalonSTPkts(ValidatedBusDriver):
             empty = BinaryValue(n_bits=len(self.bus.empty), bigEndian=False)
 
         # Drive some defaults since we don't know what state we're in
+        await NextTimeStep()
         if self.use_empty:
             self.bus.empty <= 0
         self.bus.startofpacket <= 0


ktbarrett:
Thank you for hunting the issue down! This change should be accompanied by changes to the cocotb test so it can run in Python 3. I'll run the test locally before merging.

garmin-mjames:
I don't think that this is the right approach. `NextTimeStep` will wake the task back up at the start of the next timestep where an event occurs.
I think the solution is to fix `AvalonST._wait_ready()`. Why is `ReadOnly()` required?

ktbarrett:
The `Bus`, `Driver`, `Monitor`, and `Scoreboard` sources are being moved to [cocotb-bus](github.com/cocotb/cocotb-bus). Please reopen this PR there if you wish. An easy way to do that is:

```bash
git checkout my-pr-branch
git fetch cocotb-bus
git merge cocotb-bus/master
```

Issue #2296, opened 2020-12-16
Fail on error?


toddstrader:
Is there a way to make tests fail if, say, error or critical messages are logged?  I haven't found anything immediately and am reading up on the Python logging module, but wondered if this had been discussed previously (haven't found it in issues yet either).

ktbarrett:
Who is creating the error messages you want to fail on?

Failing when an error-level message occurs _by default_ is surprising behavior to most people. And unnecessary when the user can use `assert` to fail the test if they want to. You seem to be implying some thing outside of your control is creating an error that you think should be a failure (there are plenty of those cases in the C++ code).

If you want this behavior you can add a handler to the cocotb logger to call `cocotb.scheduler.finish_test` with an exception (I suggest `cocotb.result.TestFailure` or `AssertionError`). The cocotb logger is a standard Python logger, you can get more details [here](https://docs.python.org/3/library/logging.html) on how to define custom log handlers and how to register them.

toddstrader:
Nope, this is for errors that our test code logs.  Understood this is not the current behavior of cocotb, nor would it be reasonable to change it.  But I'm wondering if there's a way to add this as a global option to cocotb or a way to add hooks so that we could perform this linkage outside of cocotb.

I'll mess around with `logging` Handlers and see what I can come up with.

toddstrader:
Is this supposed to make the test fail?
```
$ git diff
diff --git a/examples/simple_dff/test_dff.py b/examples/simple_dff/test_dff.py
index d108ff4b..bd51356d 100644
--- a/examples/simple_dff/test_dff.py
+++ b/examples/simple_dff/test_dff.py
@@ -5,6 +5,7 @@ import random
 import cocotb
 from cocotb.clock import Clock
 from cocotb.triggers import FallingEdge
+import cocotb.result
 
 
 @cocotb.test()
@@ -20,3 +21,4 @@ async def test_dff_simple(dut):
         dut.d <= val  # Assign the random value val to the input port d
         await FallingEdge(dut.clk)
         assert dut.q == val, "output q was incorrect on the {}th cycle".format(i)
+    cocotb.scheduler.finish_test(cocotb.result.TestFailure)
```
Because I'm running this and `test_dff_simple` still passes (both in `_log_test_summary` and in `results.xml`).  I also tried `AssertionError` but that didn't change the results.

[...]

PR #2299, opened 2020-12-18
Hooks for failing tests on errors

toddstrader:
Related to:
https://github.com/cocotb/cocotb/issues/2296

These are the changes I needed to add my own `logging.Handler` to count error / critical log messages and fail a test if it takes any.  If this is acceptable, I'm happy to work on testing for this as well.  Also, if desirable I can add my module which has the customer Handler which I'm invoking via `COCOTB_HOOKS`.

Some notes:
`_tests_cbs` cannot be a member of `RegressionManager` because the `COCOTB_HOOKS` are invoked during the construction of the `RegressionManager`.

I'm exposing `_test` in `RegressionManager` so that I can tell which test is logging an error and/or finishing.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -136,6 +136,9 @@ class RunningTask:
     def _finished(self):
         return self._outcome is not None
 
+    def fail(self, exc):
+        self._outcome = outcomes.Error(exc)
+
     def __iter__(self):
         return self
 
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -72,6 +72,9 @@ def _my_import(name: str) -> Any:
 
 _logger = SimLog(__name__)
 
+_test_cbs = []
+def register_test_cb(cb):
+    _test_cbs.append(cb)
 
 class RegressionManager:
     """Encapsulates all regression capability into a single place"""
@@ -227,6 +230,10 @@ class RegressionManager:
                 if hasattr(thing, "im_hook"):
                     yield thing
 
+    @property
+    def test(self) -> Test:
+        return self._test
+
     def _init_hook(self, hook: Hook) -> Optional[RunningTask]:
         try:
             test = hook(self._dut)
@@ -295,6 +302,9 @@ class RegressionManager:
 
         # stop capturing log output
         cocotb.log.removeHandler(test.handler)
+
+        for cb in _test_cbs:
+            cb(self._test, self._test_task)
 
         self._record_result(
             test=self._test,


eric-wieser (changes requested):
This looks a bit fragile to me. Can you get what you need with a custom decorator in your own code instead, something like:
```python
def logs_are_failures(func):
    async def wrapped(*args, **kwargs):
        handler = ...
        attach_the_handler()
        try:
            ret = await func(*args, **kwargs)
        finally:
            detach_the_handler()
        if handler.has_errors:
            raise AssertionError("test succeeded, but logs had errors")
        return ret
    return wrapped

@cocotb.test()
@logs_are_failures
async def my_test(dut):
    ...
```

toddstrader:
Yeah, I can confirm that this does work.  However, having to opt tests into being able to fail sounds like a recipe for disaster.

Full disclosure: we're working off of an ancient cocotb fork that a number of people made a bunch of changes to, so I'm currently trying to fix the impedance mismatch (and upstream or discard these changes) so that we can upgrade.  So this and other possible future questions are just me trying to figure out how to upgrade without refactoring our entire codebase.

I think the custom decorator idea works if I can:
```
cocotb.test = my_test_decorator
```
so that everything just picks up my wrapper and there's no possibility of using the stock cocotb test decorator.  I messing around with subclassing `test` to do this.  There's things in here like `_decorator_helper` that I don't fully understand yet so I may be back with more questions.

ktbarrett:
I really don't think we are interested in supporting this behavior in cocotb, even with a switch. This is supported by the fact that it is possible to accomplish what you want without modifying the core. 

The suggestion of overloading the cocotb.test decorator for your distribution is a hood one IMO, and is absolutely possible if we support a custom entry point that allows you to make site based customizations (#1225). I have a 2 prototypes up, but want to get some other changes merged before I make a PR.

eric-wieser:
At my old company, we had a `@our_framework.test()` decorator that wrapped the `cocotb.test` decorator. I think directly replacing `cocotb.test` as you do there is a bit "rude", and may confused developers looking at the cocotb docs. It should be very easy to do a find-and-replace in your code for `@cocotb.test` with `@yourlibrary.test` though, and you can easily implement that as:
```python
def test(*test_args, **test_kwargs):
    def decorator(func):
        return cocotb.test(*test_args, **test_kwargs)(logs_are_failures(func))
    return decorator
```
Note that by keeping `logs_are_failures` independent of the test framework, you can apply it to coroutines individually.

toddstrader:
Thanks for all the feedback.  I'll keep hacking at this to see if I can find a way through without modifying cocotb.

Issue #2300, opened 2020-12-22
Cocotb only supports Verilator 4.106


themperek:
cocotb: master 9b442d07870751ff656eb1ed53d24a3d806e9610
os: centos 7 64bit
simulator: Verilator  4.107 devel
Python: 3.7.6 (conda)

log:
```
     -.--ns INFO     cocotb.gpi                         ..mbed/gpi_embed.cpp:74   in set_program_name_in_venv        Did not detect Python virtual environment. Using system-wide Python interpreter
     -.--ns INFO     cocotb.gpi                         ../gpi/GpiCommon.cpp:105  in gpi_print_registered_impl       VPI registered
     -.--ns INFO     cocotb.gpi                         ..mbed/gpi_embed.cpp:244  in embed_sim_init                  Python interpreter initialized and cocotb loaded!
     0.00ns INFO     cocotb                                      __init__.py:212  in _initialise_testbench           Running on Verilator version 4.107 devel
     0.00ns INFO     cocotb                                      __init__.py:219  in _initialise_testbench           Running tests with cocotb v1.5.0.dev0 from /themperek/git/cocotb/cocotb
     0.00ns INFO     cocotb                                      __init__.py:239  in _initialise_testbench           Seeding Python random module with 1608640613
     0.00ns INFO     cocotb.regression                         regression.py:127  in __init__                        Found test test_dff.test_dff_simple
     0.00ns INFO     cocotb.regression                         regression.py:466  in _start_test                     Running test 1/1: test_dff_simple
     0.00ns INFO     ..st.test_dff_simple.0x7f08b057d650       decorators.py:313  in _advance                        Starting test: "test_dff_simple"
                                                                                                                     Description:  Test that d propagates to q 
```
[simple_dff_verilator.log](https://github.com/cocotb/cocotb/files/5734327/simple_dff_verilator.log)

themperek:
It seems that Verilator 4.106 works.

garmin-mjames:
Looking at it, there were recent changes to the Verilator VPI code relating to iterating over the Module hierarchy.
I can have a look at this soon, but setting `VERILATOR_SIM_DEBUG=1` and `COCOTB_LOG_LEVEL=DEBUG` should help figure out what's going on.

themperek:
@garmin-mjames I have added log file. It is cut it goes like this forever.

garmin-mjames:
[...]

PR #2304, opened 2020-12-24
Fix array bus driver with Verilator

phhorrein:
This is definitely a workaround for Verilator not providing "name[i]"
handles for array signals.

It is only a basis for discussion

### DIFF
--- a/cocotb/bus.py
+++ b/cocotb/bus.py
@@ -82,9 +82,7 @@ class Bus:
             else:
                 signame = sig_name
 
-            if array_idx is not None:
-                signame += "[{:d}]".format(array_idx)
-            self._add_signal(attr_name, signame)
+            self._add_signal(attr_name, signame, array_idx)
 
         # Also support a set of optional signals that don't have to be present
         for attr_name, sig_name in _build_sig_attr_dict(optional_signals).items():
@@ -93,19 +91,18 @@ class Bus:
             else:
                 signame = sig_name
 
-            if array_idx is not None:
-                signame += "[{:d}]".format(array_idx)
-
-            self._entity._log.debug("Signal name {}".format(signame))
             if hasattr(entity, signame):
-                self._add_signal(attr_name, signame)
+                self._add_signal(attr_name, signame, array_idx)
             else:
                 self._entity._log.debug("Ignoring optional missing signal "
                                         "%s on bus %s" % (sig_name, name))
 
-    def _add_signal(self, attr_name, signame):
-        self._entity._log.debug("Signal name {}".format(signame))
-        setattr(self, attr_name, getattr(self._entity, signame))
+    def _add_signal(self, attr_name, signame, array_idx=None):
+        self._entity._log.debug("Signal name {}, idx {}".format(signame, array_idx))
+        handle = getattr(self._entity, signame)
+        if array_idx is not None:
+            handle = handle[array_idx]
+        setattr(self, attr_name, handle)
         self._signals[attr_name] = getattr(self, attr_name)
 
     def drive(self, obj, strict=False):
--- a/tests/test_cases/test_array_buses/Makefile
+++ b/tests/test_cases/test_array_buses/Makefile
@@ -1,0 +1,17 @@
+TOPLEVEL_LANG ?= verilog
+PWD=$(shell pwd)
+TOPLEVEL := array_buses
+
+ifeq ($(TOPLEVEL_LANG),verilog)
+    VERILOG_SOURCES = $(PWD)/array_buses.sv
+else
+ifeq ($(TOPLEVEL_LANG), vhdl)
+    VHDL_SOURCES = $(PWD)/array_buses.vhd
+else
+    $(error "A valid value (verilog or vhdl) was not provided for TOPLEVEL_LANG=$(TOPLEVEL_LANG)")
+endif
+endif
+include $(shell cocotb-config --makefiles)/Makefile.sim
+
+
+MODULE = test_array_buses
--- a/tests/test_cases/test_array_buses/array_buses.vhd
+++ b/tests/test_cases/test_array_buses/array_buses.vhd
@@ -1,0 +1,43 @@
+library ieee;
+
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+
+package array_types is
+    type io_array is array (0 to 1) of std_logic_vector(7 downto 0);
+end package array_types;
+
+library ieee;
+
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+use work.array_types.all;
+
+entity array_buses is
+    port (
+        clk        : in  std_logic;
+        in_data    : in  io_array;
+        in_valid   : in  std_logic_vector(0 to 1);
+        out_data   : out io_array;
+        out_valid  : out std_logic_vector(0 to 1)
+    );
+end;
+
+architecture impl of array_buses is
+    signal tmp_data: io_array := (others => (others => '0'));
+    signal tmp_valid: std_logic_vector(0 to 1) := (others => '0');
+begin
+    process (clk)
+    begin
+        if (rising_edge(clk)) then
+            tmp_data <= in_data;
+            tmp_valid <= in_valid;
+        end if;
+    end process;
+
+   out_data(0) <= tmp_data(0);
+   out_data(1) <= tmp_data(1);
+   out_valid(0) <= tmp_valid(0);
+   out_valid(1) <= tmp_valid(1);
+
+end architecture;
--- a/tests/test_cases/test_array_buses/test_array_buses.py
+++ b/tests/test_cases/test_array_buses/test_array_buses.py
@@ -1,0 +1,79 @@
+import cocotb
+from cocotb.clock import Clock
+from cocotb.drivers import BusDriver
+from cocotb.monitors import BusMonitor
+from cocotb.triggers import RisingEdge
+
+
+class TestDriver(BusDriver):
+    _signals = ["data", "valid"]
+
+    def __init__(self, entity, name, clock, **kwargs):
+        BusDriver.__init__(self, entity, name, clock, **kwargs)
+        self.bus.valid <= 0
+
+    async def _driver_send(self, transaction, sync=True):
+        clkedge = RisingEdge(self.clock)
+        if sync:
+            await clkedge
+
+        self.log.info("Sending {}".format(transaction))
+        self.bus.valid <= 1
+        self.bus.data <= transaction
+
+        await clkedge
+        self.bus.valid <= 0
+
+
+class TestMonitor(BusMonitor):
+
+    _signals = ["data", "valid"]
+
+    def __init__(self, entity, name, clock, bank=0, **kwargs):
+        BusMonitor.__init__(self, entity, name, clock, **kwargs)
+        self.bank = bank
+        self.add_callback(self._get_result)
+        self.expected = []
+
+    async def _monitor_recv(self):
+        clkedge = RisingEdge(self.clock)
+        while True:
+            await clkedge
+            if self.bus.valid.value:
+                self._recv(int(self.bus.data))
+
+    def _get_result(self, transaction):
+        self.log.info("Received {} on bank {}".format(transaction, self.bank))
+        exp = self.expected.pop(0)
+        assert exp == int(transaction)
+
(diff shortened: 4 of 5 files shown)


phhorrein:
I think the correct way to do this would probably be to remove the "[x]" from the signame anyway, and work with index in the bus. But I don't know how other simulators work, and I'm pretty sure this could be breaking... I don't know what's better: an ugly fix for Verilator, or a larger modification.

phhorrein:
I updated the PR to modify the way indexed signals are managed in the bus:
* it now looks for the full signal
* if array_idx is set, it selects the index in the resulting signal (instead of looking for a modified name)
I included a test for this (it does nothing useful, except not failing). I tested with Verilator and with Icarus. It will fail with current master version of Verilator (and with v4.106) due to an issue in Verilator when processing unpacked array (PR pending also). I will test on Mentor, but I can't test on anything else. 

I'm sorry, I should have discussed this to ensure that this is a valid way to fix and test from a maintainer point of view, but it was a blocker for me. I can change the fix if needs be.

phhorrein:
This pull request can now be reviewed I think. Please let me know if anything is troublesome from your point of view.

cocotb/bus.py
```diff
@@ -93,19 +91,18 @@ def __init__(self, entity, name, signals, optional_signals=[], bus_separator="_"
...
 
-    def _add_signal(self, attr_name, signame):
-        self._entity._log.debug("Signal name {}".format(signame))
-        setattr(self, attr_name, getattr(self._entity, signame))
+    def _add_signal(self, attr_name, signame, array_idx=None):
+        self._entity._log.debug("Signal name {}, idx {}".format(signame, array_idx))
+        if array_idx is None:
+            setattr(self, attr_name, getattr(self._entity, signame))
+        else:
+            setattr(self, attr_name, getattr(self._entity, signame)[array_idx])
```
eric-wieser, line 105: Probably worth avoiding some duplication here:
```suggestion
        handle = getattr(self._entity, signame)
        if array_idx is None:
            handle = handle[array_idx]
        setattr(self, attr_name, handle)
```

tests/test_cases/test_array_buses/test_array_buses.py
```diff
@@ -0,0 +1,68 @@
...
+    @cocotb.coroutine
+    def _monitor_recv(self):
+        clkedge = RisingEdge(self.clock)
+        while True:
+            yield clkedge
+            if self.bus.valid.value:
+                self._recv(int(self.bus.data))
+
+    def _print_res(self, transaction):
+        self.log.info("Received {} on bank {}".format(transaction, self.bank))
```
eric-wieser, line 47: The way this test is currently written means it can't actually fail. Can you append `transaction` to a list like `self.transactions`, and add an assert on the value of `self.transactions` at the end of `test_array_buses`?

eric-wieser (commented):
Thanks for the fix - it looks reasonable to me

tests/test_cases/test_array_buses/test_array_buses.py
```diff
@@ -0,0 +1,79 @@
...
+        while True:
+            yield clkedge
+            if self.bus.valid.value:
+                self._recv(int(self.bus.data))
+
+    def _get_result(self, transaction):
+        self.log.info("Received {} on bank {}".format(transaction, self.bank))
+        exp = self.expected.pop(0)
+        if exp != int(transaction):
+            raise cocotb.result.TestFailure("Expected {}, got {}".format(exp, int(transaction)))
```
eric-wieser, line 51: If `_get_result` is never called (eg, the clock isn't actually working), then your test will still pass - you also need to check that `expected` is empty at the end of the main coroutine.

Also, shorter as:

```suggestion
    def _get_result(self, transaction):
        self.log.info("Received {} on bank {}".format(transaction, self.bank))
        exp = self.expected.pop(0)
        assert exp == int(transaction)
```

tests/test_cases/test_array_buses/test_array_buses.py
```diff
@@ -0,0 +1,79 @@
...
+
+    await in_data_0.send(10)
+    await clkedge
+    await in_data_1.send(20)
+    await clkedge
+    await in_data_0.send(30)
+    await clkedge
+    await in_data_1.send(40)
+    await clkedge
+    await clkedge
```
eric-wieser, line 79:
```suggestion
    await clkedge
    
    assert not in_data_0.expected
    assert not in_data_1.expected
```

phhorrein: Fixed in_data_0 to out_data_0 (and same for 1), but applied

phhorrein:
It seems I have a coverage issue (it was not there before, I don't get it...). I need to investigate...

eric-wieser:
The coverage is down simply because not all the tests have run yet

phhorrein:
OK, my bad, it took some time to update. All tests are good then.

tests/designs/array_buses/Makefile
```diff
@@ -0,0 +1,24 @@
...
+
+else
+ifeq ($(TOPLEVEL_LANG),verilog)
+   TOPLEVEL := array_buses
+
+   PWD=$(shell pwd)
+
+   COCOTB?=$(PWD)/../../..
+
+   VERILOG_SOURCES = $(COCOTB)/tests/designs/array_buses/array_buses.sv
```
ktbarrett, line 17: We no longer use `COCOTB`
```suggestion
   VERILOG_SOURCES = $(PWD)/array_buses.sv
```

tests/test_cases/test_array_buses/test_array_buses.py
```diff
@@ -0,0 +1,81 @@
...
+        clkedge = RisingEdge(self.clock)
+        if sync:
+            yield clkedge
+
+        self.log.info("Sending {}".format(transaction))
+        self.bus.valid <= 1
+        self.bus.data <= transaction
+
+        yield clkedge
+        self.bus.valid <= 0
```
ktbarrett, line 26: Please use the new syntax.
```suggestion
    async def _driver_send(self, transaction, sync=True):
        clkedge = RisingEdge(self.clock)
        if sync:
            await clkedge

        self.log.info("Sending {}".format(transaction))
        self.bus.valid <= 1
        self.bus.data <= transaction

        await clkedge
        self.bus.valid <= 0
```

tests/test_cases/test_array_buses/test_array_buses.py
```diff
@@ -0,0 +1,81 @@
...
+        self.add_callback(self._get_result)
+        self.expected = []
+
+    @cocotb.coroutine
+    def _monitor_recv(self):
+        clkedge = RisingEdge(self.clock)
+        while True:
+            yield clkedge
+            if self.bus.valid.value:
+                self._recv(int(self.bus.data))
```
ktbarrett, line 45:
```suggestion
    async def _monitor_recv(self):
        clkedge = RisingEdge(self.clock)
        while True:
            await clkedge
            if self.bus.valid.value:
                self._recv(int(self.bus.data))
```

tests/test_cases/test_array_buses/Makefile
```diff
@@ -0,0 +1,3 @@
+include ../../designs/array_buses/Makefile
+
+MODULE = test_array_buses
```
ktbarrett, line 3: Since this is the only test that will use this design. I would move the design into the test directory. No need to keep it separate.

ktbarrett (commented):
Thanks for tracking the Verilator issue down and solving that as well! The Python changes look good, trying to get `signal[index]` instead of getting each piece was not consistent with the rest of cocotb.

tests/test_cases/test_array_buses/Makefile
phhorrein: Agreed, and done.

ktbarrett (approved):

tests/test_cases/test_array_buses/Makefile
```diff
@@ -0,0 +1,24 @@
+TOPLEVEL_LANG ?= verilog
+
+ifneq ($(TOPLEVEL_LANG),verilog)
+
+all:
+	@echo "Skipping test due to TOPLEVEL_LANG=$(TOPLEVEL_LANG) not being verilog"
```
cmarqu, line 6: In principle, we could also have a VHDL implementation of the RTL and it should also work fine.

phhorrein: Yes, I have one. But it is not working with GHDL. I'll try with ModelSim and add it to the repo.

phhorrein: OK, I have no way to test (no VHDL license on my ModelSim), and ghdl does not support this (in_data is not found with the array type, and I can't access element i of a std_logic_vector like this). I can push the vhd, and disable the test in the Makefile, and create a new issue. Or, I can try and make it work (but it might take some time).

ktbarrett:
@phhorrein I'll test on Questa.

ktbarrett:
The questa test passed. @phhorrein I would enable the VHDL test by default and because GHDL will fail the test, add a `expect_error = AttributeError if cocotb.SIM_NAME.startswith('ghdl') else ()` (might have to adjust that, look at the other tests) argument to the cocotb tests.

phhorrein:
It's done. I added a note in the commit message.

ktbarrett (approved):
Excellent. If this gets no more reviews in the next couple days, or at least one more approval, it will be merged.

phhorrein:
By the way, I'm working on Verilator to get the fix for the single bit unpacked array merged. I had a solution, but it's not optimal, and I'm working on a new one (which takes more time, and I'm not full time on this). So tests on Verilator will still fail for now (and will required a master or next version, which is currently stalling... so not soon !)

ktbarrett:
@phhorrein This issue isn't the only one preventing Verilator from passing regressions AFAIK. We just recently rewrote the main Verilator loop and fixed a few bugs in Verilator's VPI implementation that require a minimum version of Verilator that is less than 1 month old. Verilator support is still very experimental in cocotb. Same for GHDL.

phhorrein:
OK, great, thanks to all for the reviews!

garmin-mjames:
Verilator master is broken currently due to #2300. I'm working on a fix.

tests/test_cases/test_array_buses/array_buses.vhd
```diff
@@ -0,0 +1,44 @@
...
+        in_data    : in  io_array;
+        in_valid   : in  std_logic_vector(0 to 1);
+        out_data   : out io_array;
+        out_valid  : out std_logic_vector(0 to 1)
+    );
+end;
+
+architecture impl of array_buses is
+    signal tmp_data: io_array := (others => (others => '0'));
+    signal tmp_valid: std_logic_vector(0 to 1) :=(others => '0');
```
cmarqu, line 28:
```suggestion
    signal tmp_valid: std_logic_vector(0 to 1) := (others => '0');
```
Ok, this is *really* minor :)

tests/test_cases/test_array_buses/Makefile
```diff
@@ -0,0 +1,17 @@
+TOPLEVEL_LANG ?= verilog
+PWD=$(shell pwd)
+TOPLEVEL := array_buses
+
+ifeq ($(TOPLEVEL_LANG),verilog)
+   VERILOG_SOURCES = $(PWD)/array_buses.sv
+else
+ifeq ($(TOPLEVEL_LANG), vhdl)
+   VHDL_SOURCES = $(PWD)/array_buses.vhd
```
cmarqu, line 9:
```suggestion
    VERILOG_SOURCES = $(PWD)/array_buses.sv
else
ifeq ($(TOPLEVEL_LANG), vhdl)
    VHDL_SOURCES = $(PWD)/array_buses.vhd
```

cmarqu (approved):
Sorry for being pedantic with the spaces, but since your code is so well-formatted otherwise...

I notice the SV code uses a three space indent which is uncommon for us, but I actually like three spaces (it's a good width - two is too short, four unnecessarily long, and multiples of three don't easily "alias" with the standard tab width...), and we are not strict in our existing SV code anyway, so it's fine IMO.

Thanks!

phhorrein:
Yes, I find the 3-space indentation for (System)Verilog clear, as well as alignment of equals. But as always for indentation, this is a very personal opinion :grin:

PR #2306, opened 2020-12-28
Use '<=' in places where 'dut.sig = val' was used

cmarqu:


ktbarrett:
Should this syntax also be deprecated? It's certainly not consistent and has the capacity to confuse a beginner. @eric-wieser thoughts?

ktbarrett:
Re-running sims, Icarus master was broken for a little bit when this was run.

ktbarrett (approved):

PR #2307, opened 2020-12-28
Implement strict type check for simulator version comparisons

cmarqu:

### DIFF
--- a/cocotb/_sim_versions.py
+++ b/cocotb/_sim_versions.py
@@ -6,39 +6,102 @@
 Classes to compare simulation versions.
 
 These are for cocotb-internal use only.
-
-.. warning::
-    These classes silently allow comparing versions of different simulators.
 """
 
 from distutils.version import LooseVersion
 
 
-class ActivehdlVersion(LooseVersion):
+class _LooseVersionTypeChecked(LooseVersion):
+    """Do a strict type check so that versions of different simulators cannot be compared.
+
+    Example:
+        >>> IcarusVersion("10.3 (stable)") == VerilatorVersion("10.3 (stable)")
+        Traceback (most recent call last):
+        ...
+        TypeError: Comparison not supported between instances of 'IcarusVersion' and 'VerilatorVersion'
+        >>> IcarusVersion("10.3 (stable)") != VerilatorVersion("10.3 (stable)")
+        Traceback (most recent call last):
+        ...
+        TypeError: Comparison not supported between instances of 'IcarusVersion' and 'VerilatorVersion'
+        >>> IcarusVersion("10.3 (stable)") > VerilatorVersion("10.3 (stable)")
+        Traceback (most recent call last):
+        ...
+        TypeError: '>' not supported between instances of 'IcarusVersion' and 'VerilatorVersion'
+        >>> IcarusVersion("10.3 (stable)") >= VerilatorVersion("10.3 (stable)")
+        Traceback (most recent call last):
+        ...
+        TypeError: '>=' not supported between instances of 'IcarusVersion' and 'VerilatorVersion'
+        >>> IcarusVersion("10.3 (stable)") < VerilatorVersion("10.3 (stable)")
+        Traceback (most recent call last):
+        ...
+        TypeError: '<' not supported between instances of 'IcarusVersion' and 'VerilatorVersion'
+        >>> IcarusVersion("10.3 (stable)") <= VerilatorVersion("10.3 (stable)")
+        Traceback (most recent call last):
+        ...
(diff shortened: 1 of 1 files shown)


cocotb/_sim_versions.py
```diff
@@ -6,39 +6,108 @@
...
+        ...
+        TypeError: Comparing versions of different simulators doesn't make sense
+        >>> IcarusVersion("10.3 (stable)") <= VerilatorVersion("10.3 (stable)")
+        Traceback (most recent call last):
+        ...
+        TypeError: Comparing versions of different simulators doesn't make sense
+    """
+
+    def __eq__(self, other):
+        if type(other) is not type(self):
```
ktbarrett, line 45: I think it may be better if is the type check isn't so strict. This allows comparisons between common subtypes.
```suggestion
        if not isinstance(self, type(other)) and not isinstance(other, type(self)):
```

cocotb/_sim_versions.py
```diff
@@ -6,39 +6,108 @@
...
+            raise TypeError(
+                "Comparing versions of different simulators doesn't make sense"
+            )
+        else:
+            return super().__eq__(other)
+
+    def __lt__(self, other):
+        if type(other) is not type(self):
+            raise TypeError(
```
eric-wieser, line 54: `return NotImplemented` might be just as good here

cmarqu: I had tried `NotImplemented` before because that would also allow me to use `@total_ordering`, but it didn't actually raise an exception in the end but somehow resolved the comparison instead.

eric-wieser: Are you sure? Note that `NotImplemented` does _not_ work for `__eq__`, only the other comparison methods.

cmarqu: Ah, ok, I did also use it for `__eq__`. I'll try later.

cocotb/_sim_versions.py
cmarqu: Done.

cocotb/_sim_versions.py
cmarqu: Ok, please take another look - though I have to say it look more consistent/uniform before.

cocotb/_sim_versions.py
```diff
@@ -6,39 +6,102 @@
...
+
+    def __eq__(self, other):
+        if not isinstance(self, type(other)) and not isinstance(other, type(self)):
+            raise TypeError(
+                "Comparison not supported between instances of '{}' and '{}'".format(
+                    type(self).__name__, type(other).__name__
+                )
+            )
+        else:
+            return super().__eq__(other)
```
eric-wieser, line 52: Are we sure we want this overload? This prevents doing semi-reasonable things like having a set of blacklisted versions across simulators and checking membership, as `set.__contains__` will try `__eq__`.

Returning `NotImplemented` as below will make `IcarusVersion("1") == AldecVersion("1")` `False`, which seems quite sensible to me.

cmarqu: Unless I'm misunderstanding you, it would also make `IcarusVersion("1") == IcarusVersion("1")` be `False`, wouldn't it?

eric-wieser: No, it would be True - I'm advocating you do exactly the same for `__eq__` as you did for the others

ktbarrett: Version sets are infinite, so I'm not sure if getting this working in a Python hash set is a requirement. Lets decide what we want the end product to look like and go from there.

eric-wieser: Well, even without doing that, there are very few python objects whose `__eq__` throws like this - comparing `icarus_ver == modelsim_ver` is far less nonsensical than `"three" == 3`, but the latter still doesn't error.

ktbarrett: I'm not against it returning `False`, but probing for a direction from anybody. Python's behavior here is inconsistent simply to better support how sets are implemented. The other ordering operations throw. I guess consistently inconsistent is preferable to inconsistently consistent.

eric-wieser: Note that in python 2, `3 < "three"` was legal albeit giving a consistent but otherwise mostly meaningless result.

cmarqu:
Maybe we should just leave the code as is and revisit this PR once we have a real need for it.

ktbarrett:
@cmarqu I agree, right now it "does the thing", and I think we should think about our requirements and what we want to the end product to look like before making more changes.

cmarqu:
Alright, closing this for now.

Issue #2312, opened 2020-12-30
Merge cocotb libraries?


ktbarrett:
#2274 will merge libembed, libgpilog, libgpi, and libcocotbutils into 1 library that forms the GPI "core" of cocotb. The remaining libraries: simulator extension module, libcocotb, and libpygpilog *may* be able to be combined into a single "Python wrapper" library around the GPI. There may be some weirdness there, but I think it should work. I will try after #2274 is in. This will leave only 2 main libraries (and many simulator interface libraries) in the repo.

PR #2315, opened 2020-12-30
Merge GPI libraries

### LINKED ISSUE #2274: Merge all utility C++ libraries into GPI
After #2268 is merged. All the reasons that libgpilog, libembed, and libcocotb_utils were separate objects will be gone and they should be merged into the GPI library, then libcocotb will gain a dependency on libgpi (which was not previously possible).

ktbarrett:
Closes #2274.

libgpilog, libgpi, libcocotbutils, and libembed have been merged into a
single libgpi library.

* other libraries header includes and linkage have been updated to point
  to libgpi
* all sources have been moved to the same directory
* embed.h, which was always private, is now located in the source
  directory
* documentation was added to what was merged into the GPI
* to_simulator, to_python, and the counter object have been moved to
  py_gpi_log since they are specific to Python/cocotb


cocotb_build_libs.py
```diff
@@ -441,8 +393,8 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
     libsim = Extension(
         os.path.join("cocotb", "simulator"),
         define_macros=_extra_defines,
-        include_dirs=[include_dir],
-        libraries=["cocotbutils", "gpilog", "gpi", "pygpilog"],
+        include_dirs=[include_dir, os.path.join(share_lib_dir, "py_gpi_log")],
```
ktbarrett, line 396:
```suggestion
        include_dirs=[include_dir],
```

cocotb_build_libs.py
```diff
@@ -404,8 +348,8 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
     libcocotb = Extension(
         os.path.join("cocotb", "libs", "libcocotb"),
         define_macros=_extra_defines,
-        include_dirs=[include_dir],
-        libraries=["gpilog", "cocotbutils", "pygpilog"],
+        include_dirs=[include_dir, os.path.join(share_lib_dir, "py_gpi_log")],
```
ktbarrett, line 351:
```suggestion
        include_dirs=[include_dir)],
```

ktbarrett (commented):
Left py_gpi_logging.h in `include`, so this isn't necessary,

ktbarrett:
The Windows run is failing because the activation context is not correct? Not sure if it's picking up the "wrong" context, or if the manifest is corrupt, or something else. I confirmed `DllMain` is running, but it can't find `libcocotb.dll`. @tpambor any thoughts?

garmin-mjames:
@ktbarrett I think this needs to be updated (`libembed` -> `libgpi`):

https://github.com/cocotb/cocotb/blob/728acfe83ef4bbcb067216b2c6758512c2df978d/cocotb_build_libs.py#L146-L147

ktbarrett:
@garmin-mjames Good catch, that fixed the issue. I hadn't seen that there before, what a mess.

I am getting issues still, there is apparently an issue with the side-by-side configuration, it tells me to use sxstrace to find the issue. I get:

```
INFO: Parsing Manifest File c:\users\kaleb\cocotb\cocotb\libs\libpygpilog.DLL.
	INFO: Manifest Definition Identity is libpygpilog,processorArchitecture="amd64",type="win32",version="1.0.0.0".
	ERROR: The file libpygpilog.dll is included by multiple assemblies.
ERROR: Activation Context generation failed.
```

tpambor:
@ktbarrett I moved the dependency to runtime with #2319. You could give it a try.

garmin-mjames (commented):
I know the goal is to reduce the number of libraries, but `to_python()`/`to_simulator()` doesn't really fit with `libpygpilog`. They are used by `simulator` and `libcocotb`, but not `libpygpilog`.
`libcocotbutils` could stay, or be renamed something like `libpycontext`?
Also, I noticed the Python logging doesn't track the context, and it can't do it in a straightforward way because `LOG_*` macros are called from both contexts.

`cocotb_utils.cpp` could be renamed `gpi_utils.cpp` or just `utils.cpp`.

garmin-mjames:
Need to remove `-lgpilog -lcocotbutils`:
https://github.com/cocotb/cocotb/blob/d235ece58bd161d671c3c26f4e6c54912f3b80a5/cocotb/share/makefiles/simulators/Makefile.verilator#L46

See https://github.com/cocotb/cocotb/pull/2315/checks?check_run_id=1640120832#step:22:126

ktbarrett:
I'm not 100% happy with this change either, but I was trying to piecemeal changes so I don't have one massive shuffling of code.

In the end I'd like to see the following:
* move `utils_dyn_*` into the GPI
* move context tracking into PyGPI (currently we don't have a unified object #2312, so this will have to be another object =/ )
* embed stub library that loads libpython and then gpi_embed
* move GPI logging into GPI

I guess that isn't too much. I'll do some more work.

ktbarrett:
After some thought, I think that the next step should be to instead break the compile-time link from GPI to the embed library. Otherwise I will have to compile the embed library and GPI into the same object if I wish to combine the gpi_log and utils_dyn* functions into the GPI. I'd like to avoid that because the embed stub *should* be a separate entity.

I'm going for a proper onion architecture. The biggest problem with that is that each layer currently is hardcoded to point 'up'. The embed library in the GPI (#1994), the cocotb entry point in the pygpi layer (#1225). At some point everything in the pygpi layer except the embed stub could be merged into a single object (#2312).
stuff

PR #2319, opened 2021-01-03
Move SxS dependency on libcocotb to runtime

tpambor:
Create the activation context at runtime before loading `libcocotb` rather than having the dependency already at link time.

A second manifest is created with the dependency which is then activated before loading `libcocotb`.

### DIFF
--- a/cocotb/share/lib/embed/embed.cpp
+++ b/cocotb/share/lib/embed/embed.cpp
@@ -33,18 +33,22 @@ static bool init_failed = false;
 
 
 #if ! defined(__linux__) && ! defined(__APPLE__)
-HANDLE act_ctx = NULL;
+static ACTCTX act_ctx = {
+    /* cbSize */ sizeof(ACTCTX),
+    /* dwFlags */ ACTCTX_FLAG_HMODULE_VALID | ACTCTX_FLAG_RESOURCE_NAME_VALID,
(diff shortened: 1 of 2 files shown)


cocotb/share/lib/embed/embed.cpp
```diff
@@ -33,18 +33,22 @@ static bool init_failed = false;
...
+    .cbSize = sizeof(ACTCTX),
+    .dwFlags = ACTCTX_FLAG_HMODULE_VALID | ACTCTX_FLAG_RESOURCE_NAME_VALID,
+    .lpSource = NULL,
+    .wProcessorArchitecture = 0,
+    .wLangId = 0,
+    .lpAssemblyDirectory = NULL,
+    .lpResourceName = MAKEINTRESOURCE(1000),
+    .lpApplicationName = NULL,
+    .hModule = 0
+};
```
ktbarrett, line 46: Designated initializers in C++ prior to C++20 are a GCC extension. Is there some other way to write this? If not, and MSVC will swallow it, so be it. Also should be `static`.

ktbarrett (approved):
I did just rebase this onto my branch from #2315 and updated `libembed` to `libgpi` and it worked. Thanks again for handling this! I'll merge your first and then rebase #2315. After potentially handling the designated initializer code above.

PR #2320, opened 2021-01-03
Make logging more customization friendly

cmarqu:
The added overhead for using function calls to format the single
log line elements is compensated by introducing LRU caching.

Add a custom logger test that adds visual brackets to sim timestamps.

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -33,6 +33,7 @@ import os
 import sys
 import logging
 import warnings
+from functools import lru_cache
 
 from cocotb.utils import (
     get_sim_time, get_time_from_sim_steps, want_color_output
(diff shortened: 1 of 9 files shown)


themperek:
I am not sure if this is the right place to comment. But out default logging, especially (multiple lines per log message) makes it harder to handle in any kind of parsing/post processing.

cmarqu:
Note that this PR doesn't change the default log format at all.

cmarqu:
I suppose it could be said that the new functions for sub-elements of the log line should be marked private so that we do not commit to them being stable until formally deprecated.

cocotb/log.py
```diff
@@ -173,86 +166,149 @@ class SimLogFormatter(logging.Formatter):
...
         super().__init__()
 
     # Justify and truncate
     @staticmethod
     def ljust(string, chars):
+        """Left-justify and truncate *string* to *chars* length."""
+
         if len(string) > chars:
-            return ".." + string[(chars - 2) * -1:]
+            return "".join(["..", string[(chars - 2) * -1:]])
```
cmarqu, line 180: The truncation should be more configurable, for instance, we could use the ellipsis … to save one character.

ktbarrett (commented):
The point of this seems to be adding a couple hooks to tweak the existing log formatter, but is incapable of adding or deleting information, or changing order of the elements of a log line. cocotb's code is MIT licensed, any user could copy-paste the formatter and change it in more ways that this change provides. I don't see the value.

gilbertoabram:
I think most of the goals in this PR have been addressed by #4900 and #5061. The others could be implemented via python logging filters by end users (e.g. the ellipsis glyph in abbreviations).

ktbarrett:
Superseded by #4900.

PR #2321, opened 2021-01-03
Add a reference card to the documentation

cmarqu:


documentation/source/refcard.rst
```diff
@@ -0,0 +1,77 @@
...
+      https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#line-blocks
+
+
+| Assign: ``dut.mysignal <= 0xFF00``
+| Assign immediately: ``dut.mysignal.setimmediatevalue(0xFF00)``
+| Assign a metavalue: ``dut.mysignal <= BinaryValue("X")``
+| Read: ``val = dut.mysignal.value``
+|       *(Note:* ``mysig = dut.mysignal`` *creates an alias/reference!)*
+| Bit slice: ``mybit = dut.mysignal[0].value``
+|            ``mybits = dut.mysignal[1:0].value``
```
eric-wieser, line 20: Or should this be
```suggestion
| Bit slice: ``mybit = dut.mysignal.value[0]``
|            ``mybits = dut.mysignal.value[1:0]``
```

cmarqu: I guess not:
https://github.com/cocotb/cocotb/blob/8ff91fa9e47ec6a49d3a0bb7f947a62f813cc184/cocotb/handle.py#L534-L550

cmarqu:
The rendering problem in the table when there are multiple lines per cell is due to 
https://github.com/readthedocs/readthedocs.org/blob/96d1170ccc3d84f3b7b7943e55352143a76a1474/media/css/readthedocs-doc-embed.css#L28-L31 which should probably not be active insde tables.

documentation/source/refcard.rst
```diff
@@ -0,0 +1,93 @@
...
++------------------------+----------------------------------------------------------------+
+| Keep value             | ``dut.mysignal <= cocotb.handle.Freeze()``                     |
++------------------------+----------------------------------------------------------------+
+| Release Force/Freeze   | ``dut.mysignal <= cocotb.handle.Release()``                    |
++------------------------+----------------------------------------------------------------+
+| *Normal assignment with* ``<=`` *is a "deposit"* (``cocotb.handle.Deposit()``)          |
++------------------------+----------------------------------------------------------------+
+|                                                                                         |
++------------------------+----------------------------------------------------------------+
+| Mark as test           | ``@cocotb.test(skip, expect_fail, timeout_time, timeout_unit`` |
```
ktbarrett, line 80:
```suggestion
| Mark as test           | ``@cocotb.test(skip, expect_fail, timeout_time, timeout_unit)``|
```

documentation/source/refcard.rst
```diff
@@ -0,0 +1,93 @@
...
++------------------------+----------------------------------------------------------------+
+| Assign immediately     | ``dut.mysignal.setimmediatevalue(0xFF00)``                     |
++------------------------+----------------------------------------------------------------+
+| Assign metavalue       | ``dut.mysignal <= BinaryValue("X")``                           |
++------------------------+----------------------------------------------------------------+
+| Read                   | | ``val = dut.mysignal.value``                                 |
+|                        | | (``mysig = dut.mysignal`` *creates an alias/reference)*      |
++------------------------+----------------------------------------------------------------+
+| Bit slice              | | ``mybit = dut.mysignal[0].value``                            |
+|                        | | ``mybits = dut.mysignal[1:0].value``                         |
```
ktbarrett, line 26: I'm not sure slicing the signal always works. You might want to move it after the `.value`

ktbarrett: Is this not an issue?

cmarqu: `dut.mysignal[0].value` should work if it is an array, so I'll use a different name to make it clearer and have `dut.mysignal.value[0]` in an extra line.
`[1:0]` may indeed not work at all.

cmarqu: Done.

marlonjames:
It would be nice to have links to the page where each part is documented in detail.

ktbarrett:
@marlonjames I think the issues is that we don't have those pages yet.

cmarqu:
I primarily want this to look nice when printed (in b/w), and, as @ktbarrett says, most of the items wouldn't have good reference targets.

documentation/source/refcard.rst
```diff
@@ -0,0 +1,93 @@
...
++------------------------+-----------------------------------------------------------------+
+| Generate clock         | ``clk = cocotb.fork(Clock(dut.clk, 12, "ns").start())``         |
++------------------------+-----------------------------------------------------------------+
+| Wait for signal edge   | | ``await cocotb.triggers.RisingEdge(dut.mysignal)``            |
+|                        | | ``await cocotb.triggers.FallingEdge(dut.mysignal)``           |
+|                        | | ``await cocotb.triggers.Edge(dut.mysignal)``                  |
++------------------------+-----------------------------------------------------------------+
+| Run coros concurrently | | ``task_0 = cocotb.fork(coro_0())``                            |
+|                        | | ``task_1 = cocotb.fork(coro_1())``                            |
+|                        | | ``task_2 = cocotb.scheduler.start_soon(coro)``                |
```
ktbarrett, line 47: I might not repeat the use of `fork`, but use a line to describe the difference between `start_soon` and `fork`?

ktbarrett (approved):

documentation/source/refcard.rst
cmarqu: I removed the 2nd for line but didn't add an explanation - we aren't explaining things anywhere else here (except for pitfalls); after all this is only meant to be a reminder of what you already know.

alexforencich:
Needs a warning that indicies can be reversed between slicing the signal and slicing the binary value

PR #2330, opened 2021-01-06
Fix libcocotbutils on Linux and Mac OS

ktbarrett:
libcocotbutils on Linux and Mac OS depends on the loader library "dl",
however this dependency was never added and cocotb works fine. This is
because libpython loaded libdl.so as a dependency, so it was resolved
transitively.

As of #2268, this dependency on libpython as a load-time dependency was
severed. This continued to not be a problem because simulators link
against libdl to load VPI library (like cocotb). However, verilator is
different, it is not a AoT compiled simulator that dynamically links in
cocotb, but compiles cocotb support into each simulation, so it doesn't
have the libdl dependency.

The libdl dependency by libcocotbutils is now explicitly added for Linux
and Mac OS. This is not a problem for Windows as all extension modules link
against kernel32.dll which provides that equivalent dependency on Windows.

### DIFF
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -343,11 +343,14 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
     ]
     if os.name == "nt":
         libcocotbutils_sources += ["libcocotbutils.rc"]
+    libcocotbutils_libraries = ["gpilog"]
+    if os.name != "nt":
+        libcocotbutils_libraries.append("dl")
     libcocotbutils = Extension(
         os.path.join("cocotb", "libs", "libcocotbutils"),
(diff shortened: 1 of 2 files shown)


cocotb_build_libs.py
```diff
@@ -343,11 +343,14 @@ def _get_common_lib_ext(include_dir, share_lib_dir):
     ]
     if os.name == "nt":
         libcocotbutils_sources += ["libcocotbutils.rc"]
+    libcocotbutils_libraries = ["gpilog"]
+    if os.name != "nt":
+        libcocotbutils_libraries.append("dl")
```
garmin-mjames, line 348: Maybe:
```suggestion
        libcocotbutils_libraries.append("dl")  # dlopen, dlerror, dlsym
```

garmin-mjames (approved):

Issue #2278, opened 2020-12-09
Segregate public and private interfaces


ktbarrett:
In Python, the use of methods and attributes starting with a single underscore (`_method()`) are used to communicate that method or attribute is a private implementation detail and should not be used by the user. Sphinx, Python linters, Python REPLs, and more respect this convention. However, the code in this repo does not. The use of non-underscore methods and attributes on what are clearly private interfaces is rampant. During the 2.0 development phase this should be cleaned up.

alexforencich:
And the other way around as well - for example `cocotb.utils._get_simulator_precision()` maybe shouldn't start with an underscore.

ktbarrett:
I am going to mark this as done as a lot of work has been done on this and public APIs are good. It may not be perfect, but we will leave that to individual issues.

PR #2331, opened 2021-01-07
Make scheduler interfaces private

ktbarrett:
Most of the scheduler methods which are currently "public" (don't start with an underscore) are not safe or reasonable for a user to call under any circumstance and were definitely intended to be private. Stubs remain in place of the original name to throw deprecation warnings. The interfaces not privatized are:

* `add`, same thing as `cocotb.fork`.
* `create_task` , is a staticmethod so it is safe to call whenever.
* `start_soon`, designed to be user facing.

It is intended that the stubs will be removed in 2.0.

xref #2278.

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -39,6 +39,7 @@ import sys
 import logging
 import threading
 import inspect
+import warnings
 from typing import Any, Union
 from collections.abc import Coroutine
 
@@ -327,6 +328,14 @@ class Scheduler:
 
     def react(self, trigger):
         """
+        .. deprecated:: 1.5
+            This function is now private.
+        """
+        warnings.warn("This function is now private.", DeprecationWarning, stacklevel=2)
+        return self._react(trigger)
+
+    def _react(self, trigger):
+        """
         Called when a trigger fires.
 
         We ensure that we only start the event loop once, rather than
@@ -440,7 +449,7 @@ class Scheduler:
                         continue
                     if _debug:
                         self.log.debug("Scheduling coroutine %s" % (coro._coro.__qualname__))
-                    self.schedule(coro, trigger=trigger)
+                    self._schedule(coro, trigger=trigger)
                     if _debug:
                         self.log.debug("Scheduled coroutine %s" % (coro._coro.__qualname__))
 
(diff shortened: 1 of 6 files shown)


cocotb/scheduler.py
```diff
@@ -390,13 +399,13 @@ def _event_loop(self, trigger):
                 trigger = self._pending_triggers.pop(0)
 
                 if not is_first and isinstance(trigger, GPITrigger):
-                    self.log.warning(
+                    self.log.warnings(
```
garmin-mjames, line 402: Inadvertent?
```suggestion
                    self.log.warning(
```

cocotb/scheduler.py
```diff
@@ -390,13 +399,13 @@ def _event_loop(self, trigger):
...
                         "A GPI trigger occurred after entering react - this "
                         "should not happen."
                     )
                     assert False
 
-                # this only exists to enable the warning above
+                # this only exists to enable the warnings above
```
garmin-mjames, line 408: ?
```suggestion
                # this only exists to enable the warning above
```

cocotb/scheduler.py
```diff
@@ -326,6 +327,14 @@ def _test_completed(self, trigger=None):
             self._check_termination()
 
     def react(self, trigger):
+        """
+        .. deprecated:: 1.5
+            This function is now private
```
garmin-mjames, line 332: Similarly for others, to be more consistent with the rest of the docs.
```suggestion
        .. deprecated:: 1.5
            This function is now private.
```

garmin-mjames: Was this missed or do you prefer no period?

ktbarrett: Just forgot to save before pushing.

ktbarrett:
Should I create a newsfragment for #2278? I don't think I should list every change because in the end there may be a lot.

garmin-mjames:
I think a summary newsfragment would be good. We can link to the individual PRs from that issue if we want to track all of the changes.

garmin-mjames (approved):

documentation/source/newsfragments/2278.removal.rst
```diff
@@ -0,0 +1,4 @@
+The access mode of many interfaces in the cocotb core libraries were re-evaluated.
```
cmarqu, line 1:
```suggestion
The access modes of many interfaces in the cocotb core libraries were re-evaluated.
```
?

Issue #2208, opened 2020-11-17
Allow more than one C-level callback for a signal edge


themperek:
Right now it is impossible to have more than one C-level callback for a signal edge. The issue is that we allocate the callbacks statically per signal:

https://github.com/cocotb/cocotb/blob/c4cfcd3e3faae6ea12851c0336fa92924ffce176/cocotb/share/lib/vpi/VpiImpl.h#L188-L190

I imagine this was motivated by a reluctance to implement proper memory management of these callbacks. Allocating on the heap would be a bad idea unless we were capable of cleaning up after ourselves.

Because there is only one callback, we can't afford to let C code replace the callback the currently wakes the scheduler, as doing so would make python miss events.

_Originally posted by @eric-wieser in https://github.com/cocotb/cocotb/issues/1682#issuecomment-616508278_

PR #2338, opened 2021-01-12
VPI: Fix callback handle leak and optimize callback registration

### LINKED ISSUE #2300: Cocotb only supports Verilator 4.106

garmin-mjames:
Closes #2300.

All callbacks except `cbValueChange` and `cbAfterDelay` were leaking callback handles
and weren't properly removing the callbacks from the simulator.

After some recent changes on Verilator master it gets stuck in an infinite loop because the `ReadWrite` callback wasn't removed, cocotb just released the handle.

VPI callback cleanup will now properly remove simulator callbacks when
appropriate. During the execution of a callback, calling cleanup_callback()
will mark the callback for removal, but won't call `vpi_remove_cb()` until after
the user callback is executed, because the callback may be re-primed during
callback execution. This saves the work of removing and then re-adding a new
simulator callback.

`cleanup_callback()` was also changed to return 0 instead of -1 when an error
occurs, as a non-zero return will delete the callback object. Some of the
callback objects are singletons, so returning a -1 would likely cause major issues or
a crash.

---

This needs testing with Modelsim, as `vpi_free_object()` is now being called on the `cbAfterDelay` callback handle for `vpiTimedCbHdl` objects.

### DIFF
--- a/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -631,7 +631,7 @@ int32_t handle_vpi_callback(p_cb_data cb_data)
 {
     int rv = 0;
 
-    VpiCbHdl *cb_hdl = (VpiCbHdl*)cb_data->user_data;
+    VpiCbHdl *cb_hdl = reinterpret_cast<VpiCbHdl*>(cb_data->user_data);
 
     if (!cb_hdl) {
         LOG_CRITICAL("VPI: Callback data corrupted: ABORTING");
@@ -648,14 +648,19 @@ int32_t handle_vpi_callback(p_cb_data cb_data)
 
         gpi_cb_state_e new_state = cb_hdl->get_call_state();
 
-        /* We have re-primed in the handler */
-        if (new_state != GPI_PRIMED)
+        if (new_state == GPI_REPRIME) {
+            cb_hdl->set_call_state(GPI_PRIMED);
+        }
+
+        // Cleanup if the callback wasn't re-primed in the handler
+        if (new_state != GPI_PRIMED && new_state != GPI_REPRIME){
             if (cb_hdl->cleanup_callback()) {
                 delete cb_hdl;
             }
+        }
 
     } else {
-        /* Issue #188: This is a work around for a modelsim */
+        // Issue #188: This is a work around for Modelsim
         if (cb_hdl->cleanup_callback()) {
             delete cb_hdl;
         }
--- a/cocotb/share/lib/gpi/gpi_priv.h
+++ b/cocotb/share/lib/gpi/gpi_priv.h
@@ -44,11 +44,11 @@
 #include <cocotb_utils.h>  // COCOTB_UNUSED
 
 typedef enum gpi_cb_state {
-    GPI_FREE = 0,
-    GPI_PRIMED = 1,
-    GPI_CALL = 2,
-    GPI_REPRIME = 3,
-    GPI_DELETE = 4,
+    GPI_FREE = 0,           ///< No sim callback is registered or sim callback is disabled
+    GPI_PRIMED = 1,         ///< Sim callback is registered and enabled
+    GPI_CALL = 2,           ///< User callback function is executing
+    GPI_REPRIME = 3,        ///< Callback is armed again while in user callback function
+    GPI_DELETE = 4,         ///< Callback is marked for cleanup and user callback will not be called
 } gpi_cb_state_e;
 
 class GpiCbHdl;
--- a/cocotb/share/lib/vpi/VpiCbHdl.cpp
+++ b/cocotb/share/lib/vpi/VpiCbHdl.cpp
@@ -54,72 +54,80 @@ VpiCbHdl::VpiCbHdl(GpiImplInterface *impl) : GpiCbHdl(impl)
  */
 int VpiCbHdl::arm_callback() {
 
-    if (m_state == GPI_PRIMED) {
-        fprintf(stderr,
-                "Attempt to prime an already primed trigger for %s!\n",
-                m_impl->reason_to_string(cb_data.reason));
-    }
-
-    // Only a problem if we have not been asked to deregister and register
-    // in the same simulation callback
-    if (m_obj_hdl != NULL && m_state != GPI_DELETE) {
-        fprintf(stderr,
-                "We seem to already be registered, deregistering %s!\n",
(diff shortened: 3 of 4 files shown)


ktbarrett:
> Some of the callback objects are singletons, so this will likely cause major issues or a crash.

That doesn't sound good. You are saying that a crash is possible with this change? This is because it would attempt to free that which is supposed to be a singleton with a lazy-static lifetime?

It seems that absolutely no attention was paid originally to the issue of object lifetimes. It's understandable, it is a hard problem. With the prospect of multiple possible GPI users I think we probably need to invest time into implementing *some* kind of garbage collection strategy and employing it in the GPI; or this is going to get painful going forward...

garmin-mjames:
@ktbarrett No, I'm saying that I fixed that issue. The `return -1`'s were added in #1863.

garmin-mjames:
Looks like there's an issue with Riviera. I'll check Active-HDL.

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -648,14 +648,19 @@ int32_t handle_vpi_callback(p_cb_data cb_data)
...
         gpi_cb_state_e new_state = cb_hdl->get_call_state();
 
-        /* We have re-primed in the handler */
-        if (new_state != GPI_PRIMED)
+        if (new_state == GPI_REPRIME) {
+            cb_hdl->set_call_state(GPI_PRIMED);
+        }
+
+        // Cleanup if the callback wasn't re-primed in the handler
+        if (new_state != GPI_PRIMED && new_state != GPI_REPRIME){
```
ktbarrett, line 656: `else if` or a `switch` like in the previous bits? I tend to like switches when testing enums, there are some useful warnings for catching gotchas.

ktbarrett (commented):
Could you add coverage exceptions (`LCOV_EXCL_START` and `LCOV_EXCL_STOP`) for code sections that we never intend/could possibly test, like handling `vpi_free_object` returning non-0?

ktbarrett (commented):
So just for posterity, the actual issue was that when arm_callback was called and the callback wasn't in the GPI_PRIMED or GPI_DELETE state, cleanup_callback is called. cleanup_callback only cleaned up the handle if it was in the GPI_PRIMED state. The other states (CALL, REPRIME) would allow the the object to be overwritten with NULL without vpi_free_object being called and be leaked.

And all this for a dubious optimization?

garmin-mjames:
@ktbarrett The problem was that `handle_vpi_callback()` sets the state to `GPI_CALL`, and so unpriming the trigger would call `cleanup_callback()` and that would release the simulator callback handle but not remove the callback.

We could take the approach of always removing the callback on cleanup, and registering a new one when arming the callback. Slightly simpler but less efficient.

VHPI allows disabling registered callbacks, which is what the VHPI implementation does rather than remove/add.

garmin-mjames:
Also, `vpi_free_object` is deprecated, replaced by `vpi_release_handle`, but I can leave that for a different PR.

ktbarrett:
> We could take the approach of always removing the callback on cleanup, and registering a new one when arming the callback. Slightly simpler but less efficient.

At this point there isn't a reason to remove it to since the optimization *now* works. Good job tracking that down, it's not an easy error to spot.

> Also, vpi_free_object is deprecated, replaced by vpi_release_handle, but I can leave that for a different PR.

Probably a good idea. Anytime we change the when, where, and what of VPI calls some simulator ends up unhappy about it. I have doubts that most simulators have even implemented `vpi_release_handle`, we should only change for the simulators that are currently complaining about it.

garmin-mjames:
Found the problem in Active-HDL / Riviera.

VSimSA is re-using callbacks for `cbValueChange`.


In `test_compare` you have this pattern:
```py3
    await RisingEdge(dut.clk)
    ...
    await FallingEdge(dut.clk)
```
So, we have a `cbValueChange` callback handle for the `dut.clk` signal, and after the `RisingEdge` trigger fires, we are inside that callback.
Then we yield `FallingEdge`, which when primed will register for a new `cbValueChange` callback on the same signal.
Riviera gives us back the same handle we already have for `RisingEdge` (but doesn't update the callback data for the new `FallingEdge` object).
We then cleanup `RisingEdge`, which calls `vpi_remove_cb` on the shared handle, sets the handle to NULL, and sets the state to `GPI_FREE`.
On the falling edge of `dut.clk`, the callback fires again but with the original `RisingEdge` callback object that is `GPI_FREE` so nothing happens, it just returns to the simulator.

I'm pretty sure this falls afoul of the VPI standard. I found this, which is in **38.36.1.1** in context of callbacks on statements:

> Multiple calls to `vpi_register_cb()` with the same data shall result in multiple callbacks.

This problem didn't happen before because the handle for `VpiValueCbHdl` objects was always removed by the scheduler with `trigger.unprime`

---

First try didn't make it through the minefield of simulator VPI support, *quelle surprise*. 😞

garmin-mjames:
We could fix the problem at the same time as #2208 by using only one callback for each signal, and keeping lists of registered user callbacks with the edge type for each one.

garmin-mjames:
See https://github.com/cocotb/cocotb/issues/2300#issuecomment-763195043

PR #2352, opened 2021-01-17
Timer param deprecation

### LINKED ISSUE #1115: Timer value argument is misleading

ktbarrett:
Closes #1115.

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -158,12 +158,14 @@ class GPITrigger(Trigger):
 class Timer(GPITrigger):
     """Fires after the specified simulation time period has elapsed."""
 
-    def __init__(self, time_ps, units="step"):
+    def __init__(self, time=None, units="step", *, time_ps=None):
         """
         Args:
-           time_ps (numbers.Real or decimal.Decimal): The time value.
+           time (numbers.Real or decimal.Decimal): The time value.
                Note that despite the name this is not actually in picoseconds
                but depends on the *units* argument.
+               .. versionchanged:: 1.5.0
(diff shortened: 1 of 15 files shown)


cocotb/triggers.py
```diff
@@ -155,13 +155,30 @@ def unprime(self):
...
     """Fires after the specified simulation time period has elapsed."""
 
-    def __init__(self, time_ps, units="step"):
+    def __init__(self, time=_REQUIRED, units="step", *, time_ps=_DEPRECATED):
         """
         Args:
-           time_ps (numbers.Real or decimal.Decimal): The time value.
+           time (numbers.Real or decimal.Decimal): The time value.
                Note that despite the name this is not actually in picoseconds
                but depends on the *units* argument.
```
eric-wieser, line 183:
```suggestion
               but depends on the *units* argument.
               
               .. versionchanged:: 1.5.0
                  Previously this argument was misleadingly called `time_ps`.
```

eric-wieser (commented):
C++ changes all look good, I'm not sure if it's really worth worrying about actually emitting deprecation warnings here.

cmarqu (approved):

cocotb/triggers.py
```diff
@@ -155,15 +155,35 @@ def unprime(self):
...
+
```
cmarqu, line 184:
```suggestion

```
Remove whitespace.

PR #2355, opened 2021-01-18
Use .value in asserts in example code

cmarqu:
This is not strictly necessary, but easier to understand for beginners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ async def test_dff_simple(dut):
         val = random.randint(0, 1)
         dut.d <= val  # Assign the random value val to the input port d
         await FallingEdge(dut.clk)
-        assert dut.q == val, "output q was incorrect on the {}th cycle".format(i)
+        assert dut.q.value == val, "output q was incorrect on the {}th cycle".format(i)
 ```
 
 A simple Makefile:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -81,7 +81,7 @@ async def test_dff_simple(dut):
         val = random.randint(0, 1)
         dut.d <= val  # Assign the random value val to the input port d
         await FallingEdge(dut.clk)
-        assert dut.q == val, "output q was incorrect on the {}th cycle".format(i)
+        assert dut.q.value == val, "output q was incorrect on the {}th cycle".format(i)
```
ktbarrett, line 84: What does pytest do here? It only does assertion message rewriting if the message was not explicitly passed, no?

ktbarrett (approved):

README.md
cmarqu: The output looks like this  in case of failure:
```
#  16000.00ns ERROR    cocotb.regression                         regression.py:399  in _score_test                     Test Failed: test_dff_simple (result was AssertionError)
#                                                                                                                      Traceback (most recent call last):
#                                                                                                                        File "/opt/tools/cocotb/examples/simple_dff/test_dff.py", line 22, in test_dff_simple
#                                                                                                                          assert dut.q == val-1, "output q was incorrect on the {}th cycle".format(i)
#                                                                                                                      AssertionError: output q was incorrect on the 0th cycle
#                                                                                                                      assert ModifiableObject(dff.q) == (0 - 1)
#                                                                                                                       +  where ModifiableObject(dff.q) = HierarchyObject(dff).q
#  16000.00ns ERROR    cocotb.regression                         regression.py:479  in _log_test_summary               Failed 1 out of 1 tests (0 skipped)
```

PR #2358, opened 2021-01-19
Move to 'src' directory layout

### LINKED ISSUE #2286: Move to a src/ layout
The benefits of putting packages in a `src/` directory in Python/pip projects has been discussed for a few years now. The advantages include running coverage collection without doing an editable install and testing what is installed and not what is local. We have to use an editable install in our tox to be able to correctly collect coverage. This has the disadvantage of not being PEP517 compliant in the future (well maybe they will add specifications for editable installs, but I doubt it), and developers have to clean all the build artifacts out every time we run `pip`/`tox` to prevent using potentially out of date C++ libraries which aren't rebuilt if they exist. And editable installs just fundamentally break isolation as provided by tox.

Preliminary testing of a move to a `src/` layout shows that the directory rewriting introduced in #2155 break paths and the build process once we move to a `src/` layout. I'm sure there is some way around this.

### LINKED ISSUE #2698: 'make help' reveals a circular dependency problem in handle.py while importing simulator from cocotb
**Environment:**
* cocotb: 1.5.2
* MinGW32 on Windows 10 64-bit
* GHDL latest MinGW32 installation with mcode backend
* Python 3.9 from MinGW

**Description:**  
After following installation instructions for cocotb (installing gcc, make, pip, ghdl, ...) and then downloading cocotb sources, it gives an error when `make help` is called.
```
Patrick Lehmann@Paebbels-Laptop MINGW32 ~/temp/cocotb
$ make help
  File "C:/msys64/mingw32/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:/msys64/mingw32/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:/msys64/mingw32/bin/cocotb-config.exe/__main__.py", line 4, in <module>
  File "C:/msys64/home/Patrick Lehmann/temp/cocotb/cocotb/__init__.py", line 43, in <module>
    import cocotb.handle
  File "C:/msys64/home/Patrick Lehmann/temp/cocotb/cocotb/handle.py", line 39, in <module>
    from cocotb import simulator
ImportError: cannot import name 'simulator' from partially initialized module 'cocotb' (most likely due to a circular import) (C:/msys64/home/Patrick Lehmann/temp/cocotb/cocotb/__init__.py)

Available simulators:

```

I think this circular dependency problem could be checked in CI with a testcase calling `make help` in CI too, right?

ktbarrett:
Closes #2286. Closes #2698.

* Move package sources into a 'src' directory

There are many advantages to this set up, the biggest being testing the
package *as installed*. https://hynek.me/articles/testing-packaging/


cmarqu:
There are paths that need `src/` added at least in the beginning of `documentation/source/building.rst`, two times in `documentation/source/library_reference_c.rst`.
https://github.com/cocotb/cocotb/blob/ef93b38336e83fb91db74fb2136bcd81d4122c95/documentation/source/conf.py#L309 maybe also needs to be `../src/`, see the warning at the end of https://cocotb--2358.org.readthedocs.build/en/2358/library_reference_c.html

ktbarrett:
The 2nd and 3rd commit can be left off if they are deemed objectionable, or if they are better split off into another PR.

ktbarrett:
C++ sources are no longer being pulled into coverage.

marlonjames:
The pip option `--no-clean` might help.

marlonjames:
The docs say:
> MANIFEST.in may still be used: anything defined there overrides the hook. This is mostly useful to exclude files tracked in your SCM from packages, although in principle it can be used to explicitly include non-tracked files too.

So we could keep it if we want to exclude files.

ktbarrett:
Right, but we don't have any exclude rules rn. Should I add back an empty MANIFEST.in?

ktbarrett:
Should include some attempt of running tests without editable install.

imphil:
In the maintainer meeting today we decided to not follow along with this change for the moment due to the churn it causes and the limited benefit. Closing for now, and will re-open if we find good reasons to do this change.

PR #2371, opened 2021-01-24
Add GCC problem matcher for GitHub Actions

cmarqu:

### DIFF
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -34,6 +34,7 @@
  * Uses GPI calls to interface to the simulator.
  */
 
+int unused;
 static int takes = 0;
 static int releases = 0;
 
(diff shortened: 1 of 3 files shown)


ktbarrett:
Could we see a commit add an example warning added to the C++ to show that it works? Then you can add a revert commit for the warning code, just so a pointer to the failing action is recorded in the PR for posterity.

cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -34,6 +34,7 @@
  * Uses GPI calls to interface to the simulator.
  */
 
+int unused;
```
ktbarrett, line 37: Doesn't seem to work?

cmarqu: Yeah, I need to find the time to work on this further.

imphil:
Let's close this for the time being, we can always reopen if someone finds more time to work on this idea.

PR #2376, opened 2021-01-25
Describe relation of HDL and Python types in docs

### LINKED ISSUE #2217: Explicitly document types of values read from HDL
@jshaker000 Arrays of logic's and subtypes of that (`sfixed`, `unsigned`, etc.) are `BinaryValue`. Integer nets and constants (`integer`, `natural`, etc.) return `int`. Floating point nets and constants (`real`) return `float`. Boolean nets and constants (`boolean`) return `bool`. String nets and constants (`string`) return `bytes`.

_Originally posted by @ktbarrett in https://github.com/cocotb/cocotb/issues/2215#issuecomment-731768459_

We a lacking a paragraph that explicitly says it like this in https://docs.cocotb.org/en/latest/quickstart.html#reading-values-from-signals (and such a section should not (only) be in a "Quickstart" chapter but in e.g. "Using cocotb").

cmarqu:
Closes #2217.

Rendered as https://cocotb--2376.org.readthedocs.build/en/2376/quickstart.html#reading-values-from-signals

### DIFF
--- a/documentation/source/quickstart.rst
+++ b/documentation/source/quickstart.rst
@@ -210,11 +210,26 @@ the various actions described in :ref:`assignment-methods` can be used.
     dut.my_signal <= Freeze()     # my_signal stays at current value until released
 
 
+.. _quickstart_reading_values:
+
 Reading values from signals
 ---------------------------
 
-Accessing the :attr:`~cocotb.handle.NonHierarchyObject.value` property of a handle object will return a :any:`BinaryValue` object.
-Any unresolved bits are preserved and can be accessed using the :attr:`~cocotb.binary.BinaryValue.binstr` attribute,
+Values in the DUT can be accessed with the :attr:`~cocotb.handle.NonHierarchyObject.value`
+property of a handle object.
+A common mistake is forgetting the ``.value`` which just gives you a reference to a handle
+(useful for defining an alias name), not the value.
+
+The Python type of a value depends on the handle's HDL type:
+
+* Arrays of ``logic`` and subtypes of that (``sfixed``, ``unsigned``, etc.) are of type :class:`~cocotb.binary.BinaryValue`.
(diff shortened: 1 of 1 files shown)


ktbarrett (approved):

documentation/source/quickstart.rst
```diff
@@ -210,11 +210,26 @@ the various actions described in :ref:`assignment-methods` can be used.
...
+A common mistake is forgetting the ``.value`` which just gives you a reference to a handle
+(useful for defining an alias name), not the value.
+
+The Python type of a value depends on the handle's HDL type:
+
+| Arrays of ``logic`` and subtypes of that (``sfixed``, ``unsigned``, etc.) are of type :class:`~cocotb.binary.BinaryValue`.
+| Integer nets and constants (``integer``, ``natural``, etc.) return :class:`int`.
+| Floating point nets and constants (``real``) return :class:`float`.
+| Boolean nets and constants (``boolean``) return :class:`bool`.
+| String nets and constants (``string``) return :class:`bytes`.
```
eric-wieser, line 229: Possibly more suited to a list:
```suggestion
* Arrays of ``logic`` and subtypes of that (``sfixed``, ``unsigned``, etc.) are of type :class:`~cocotb.binary.BinaryValue`.
* Integer nets and constants (``integer``, ``natural``, etc.) return :class:`int`.
* Floating point nets and constants (``real``) return :class:`float`.
* Boolean nets and constants (``boolean``) return :class:`bool`.
* String nets and constants (``string``) return :class:`bytes`.
```

ktbarrett: I agree. [The block](https://cocotb--2376.org.readthedocs.build/en/2376/quickstart.html#reading-values-from-signals) isn't teh prettiest.

cmarqu: Agreed, changed.

Issue #2351, opened 2021-01-16
Extend cocotb-config to return path to vpi/fli/vhpi library


themperek:
Considering that https://github.com/cocotb/cocotb/pull/1798 will make vpi/fli/vhpi different depending on platform/os we can add extra functionality to `cocotb-config` that would allow easier custom flow integration.

An example:
`cocotb-config --vpi-lib icarus` return absolut path to `libcocotbvpi_icarus.so`
`cocotb-config --vpi-lib questa` return absolut path to `libcocotbvpi_modelsim.so`
`cocotb-config --lib-dir` return path where `libcocotbvpi_icarus.so` is located

Could also add something like this:
`cocotb-config --vpi-cmd icarus` that would return `-M $(cocotb-config --prefix)/cocotb/libs -m libcocotbvpi_icarus`
`cocotb-config --fli-cmd questa` that would return `-foreign "cocotb_init $(cocotb-config --prefix)/cocotb/libs/libcocotbfli_modelsim.so`

Any support for this?
Any suggestions?

Related https://github.com/cocotb/cocotb/pull/2340

ktbarrett:
I agree that something like this would be useful. At a minimum giving the directory of the libs and the name of each particular lib will be useful. Especially if we do any renaming or combining of libraries in the future, creating this abstraction should allow those changes to not break users. It might also be useful for people making GPI extensions

I'm thinking
```
cocotb-config --includes                          # prints to include dirs for VPI, VHPI, FLI, and GPI (for use with '-I$(cocotb-config --includes)')
cocotb-config --lib-dirs                          # prints directory containing libgpi (for use with '-L$(cocotb-config --lib-dirs)')
cocotb-config --lib-path {interface} {simulator}  # prints absolute path to library the GPI implementation of a given interface for a simulator  (for use with custom flows)
cocotb-config --cmd-args {interface} {simulator}  # prints arguments to pass simulator to get it to run cocotb (for use with custom flows)
```

cmarqu:
`--vpi-cmd` or `--cmd-args` would need an additional `--format` option because not all flows will require the `$(...)` syntax for calling a command, but in general, I like this idea.

themperek:
> `--vpi-cmd` or `--cmd-args` would need an additional `--format` option because not all flows will require the `$(...)` syntax for calling a command, but in general, I like this idea.

Maybe we can return raw text without parameters (`$(...)` syntax) in this case?

eric-wieser:
[...]

PR #2387, opened 2021-02-05
Extend cocotb-config to return path to libraries

themperek:
Related #2351

An interesting complication is that Icarus accepts library name without extension only. All others seem to be ok with extension name.

- [x] news fragment
- [x] Should I change all makefiles to use it?

### DIFF
--- a/cocotb/config.py
+++ b/cocotb/config.py
@@ -37,6 +37,7 @@ This module provides information in module global variables and through a
 Global variables:
     share_dir: str, path where the cocotb data is stored
     makefiles_dir: str, path where the cocotb makefiles are installed
+    libs_dir: str, path where the cocotb interface libraries are located
 """
 import argparse
 import os
@@ -45,11 +46,16 @@ import textwrap
 import cocotb
 import cocotb._vendor.find_libpython as find_libpython
 
-__all__ = ["share_dir", "makefiles_dir"]
+__all__ = ["share_dir", "makefiles_dir", "libs_dir"]
 
 
 share_dir = os.path.join(os.path.dirname(cocotb.__file__), "share")
 makefiles_dir = os.path.join(os.path.dirname(cocotb.__file__), "share", "makefiles")
+libs_dir = os.path.join(os.path.dirname(cocotb.__file__), "libs")
+
+# On Windows use mixed mode "c:/a/b/c" as this work in all cases
+if os.name == "nt":
+    libs_dir = libs_dir.replace("\\", "/")
 
 
 def help_vars_text():
@@ -97,6 +103,59 @@ def help_vars_text():
     return helpmsg
 
 
+def lib_name(interface: str, simulator: str) -> str:
+    """
+    Returns the name of interface library for given interface (VPI/VHPI/FLI) and simulator.
+    """
+
+    interface_name = interface.lower()
+    supported_interfaces = ["vpi", "vhpi", "fli"]
+    if interface_name not in supported_interfaces:
+        raise ValueError("Wrong interface used. Supported: " + ", ".join(supported_interfaces))
+
+    simulator_name = simulator.lower()
+    supported_sims = ["icarus", "questa", "modelsim", "ius", "xcelium", "vcs", "ghdl", "riviera", "activehdl", "cvc"]
+    if simulator not in supported_sims:
+        raise ValueError("Wrong simulator name. Supported: " + ", ".join(supported_sims))
+
+    if simulator_name in ["questa", "cvc"]:
+        library_name = "modelsim"
+    elif simulator_name == "xcelium":
+        library_name = "ius"
+    elif simulator_name in ["riviera", "activehdl"]:
+        library_name = "aldec"
+    else:
+        library_name = simulator_name
+
+    if library_name == "icarus":
+        lib_ext = ""
+    elif os.name == "nt":
+        lib_ext = ".dll"
+    else:
+        lib_ext = ".so"
+
+    # check if compiled with msvc
+    if os.path.isfile(os.path.join(libs_dir, "cocotb.dll")):
+        lib_prefix = ""
+    else:
+        lib_prefix = "lib"
+
+    return lib_prefix + "cocotb" + interface_name + "_" + library_name + lib_ext
+
+
+def lib_name_path(interface, simulator):
+    """
+    Returns the absolute path of interface library for given interface (VPI/VHPI/FLI) and simulator
+    """
+    library_name_path = os.path.join(libs_dir, lib_name(interface, simulator))
+
+    # On Windows use mixed mode "c:/a/b/c" as this work in all cases
+    if os.name == "nt":
+        return library_name_path.replace("\\", "/")
+
+    return library_name_path
+
+
 class PrintAction(argparse.Action):
     def __init__(self, option_strings, dest, text=None, **kwargs):
         super(PrintAction, self).__init__(option_strings, dest, nargs=0, **kwargs)
@@ -104,6 +163,19 @@ class PrintAction(argparse.Action):
 
     def __call__(self, parser, namespace, values, option_string=None):
         print(self.text)
+        parser.exit()
+
+
+class PrintFuncAction(argparse.Action):
+    def __init__(self, option_strings, dest, function=None, **kwargs):
+        super().__init__(option_strings, dest, **kwargs)
+        self.function = function
+
+    def __call__(self, parser, args, values, option_string=None):
+        try:
+            print(self.function(*values))
+        except ValueError as e:
+            parser.error(e)
         parser.exit()
 
 
@@ -144,17 +216,39 @@ def get_parser():
         text=help_vars_text(),
     )
     parser.add_argument(
+        "--libpython",
+        help="prints the absolute path to the libpython associated with the current Python installation",
+        action=PrintAction,
+        text=find_libpython.find_libpython(),
+    )
+    parser.add_argument(
+        "--lib-dir",
+        help="Print the absolute path to the interface libraries location",
+        action=PrintAction,
+        text=libs_dir
+    )
+    parser.add_argument(
+        "--lib-name",
+        help='Print the name of interface library for given interface (VPI/VHPI/FLI) and simulator',
+        nargs=2,
+        metavar=('INTERFACE', 'SIMULATOR'),
+        action=PrintFuncAction,
+        function=lib_name
+    )
+    parser.add_argument(
+        "--lib-name-path",
+        help='Print the absolute path of interface library for given interface (VPI/VHPI/FLI) and simulator',
+        nargs=2,
+        metavar=('INTERFACE', 'SIMULATOR'),
+        action=PrintFuncAction,
+        function=lib_name_path
+    )
+    parser.add_argument(
         "-v",
         "--version",
         help="echo the version of cocotb",
         action=PrintAction,
         text=version,
-    )
-    parser.add_argument(
-        "--libpython",
-        help="prints the absolute path to the libpython associated with the current Python installation",
-        action=PrintAction,
-        text=find_libpython.find_libpython(),
     )
 
     return parser
--- a/cocotb/config.py
+++ b/cocotb/config.py
@@ -105,7 +105,7 @@ def help_vars_text():
 
 def lib_name(interface: str, simulator: str) -> str:
     """
-    Returns the name of interface library for given interface (VPI/VHPI/FLI) and simulator.
+    Return the name of interface library for given interface (VPI/VHPI/FLI) and simulator.
     """
 
     interface_name = interface.lower()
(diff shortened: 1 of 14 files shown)


cocotb/config.py
```diff
@@ -157,6 +216,29 @@ def get_parser():
         text=find_libpython.find_libpython(),
     )
 
+    parser.add_argument(
+        "--lib-dir",
+        help="prints the absolute path to the interface libraties location",
```
ktbarrett, line 221:
```suggestion
        help="prints the absolute path to the interface libraries location",
```

cocotb/config.py
```diff
@@ -107,6 +153,19 @@ def __call__(self, parser, namespace, values, option_string=None):
         parser.exit()
 
 
+class PrintFuncAction(argparse.Action):
+    def __init__(self, option_strings, dest, function=None, **kwargs):
+        super(PrintFuncAction, self).__init__(option_strings, dest, **kwargs)
```
ktbarrett, line 158: Prefer no arguments in Python 3.
```suggestion
        super().__init__(option_strings, dest, **kwargs)
```

ktbarrett (commented):
Looks good, I'd like to test it out first.

> Update all makefiles ?

Well that would act as a good self-checking test... I would say yes.

cocotb/config.py
```diff
@@ -97,6 +99,50 @@ def help_vars_text():
     return helpmsg
 
 
+def lib_name(interface, simulator):
+
+    interface_name = interface.lower()
+    supported_interfaces = ["vpi", "vhpi", "fli"]
+    if interface_name not in supported_interfaces:
+        raise ValueError("Wrong interface used. Supported :" + ", ".join(supported_interfaces))
```
cmarqu, line 107:
```suggestion
        raise ValueError("Wrong interface used. Supported:" + ", ".join(supported_interfaces))
```

cocotb/config.py
```diff
@@ -97,6 +99,50 @@ def help_vars_text():
...
+
+    simulator_name = simulator.lower()
+    supported_sim = ["icarus", "questa", "modelsim", "ius", "xcelium", "vcs", "ghdl", "riviera", "activehdl", "cvc"]
+    if simulator not in supported_sim:
+        raise ValueError("Wrong simulator name. Supported : " + ", ".join(supported_sim))
```
cmarqu, line 112:
```suggestion
        raise ValueError("Wrong simulator name. Supported: " + ", ".join(supported_sim))
```

cocotb/config.py
cmarqu:
```suggestion
        help="Print the absolute path to the interface libraries location",
```

cocotb/config.py
```diff
@@ -157,6 +216,29 @@ def get_parser():
...
 
+    parser.add_argument(
+        "--lib-dir",
+        help="prints the absolute path to the interface libraties location",
+        action=PrintAction,
+        text=libs_dir
+    )
+    parser.add_argument(
+        "--lib-name",
+        help='prints the name of interface library for given interface (VPI/VHPI/FLI) and simulator',
```
cmarqu, line 227:
```suggestion
        help='Print the name of interface library for given interface (VPI/VHPI/FLI) and simulator',
```

cocotb/config.py
```diff
@@ -157,6 +216,29 @@ def get_parser():
...
+        nargs=2,
+        metavar=('INTERFACE', 'SIMULATOR'),
+        action=PrintFuncAction,
+        function=lib_name
+    )
+    parser.add_argument(
+        "--lib-name-path",
+        help='prints the absolute path of interface library for given interface (VPI/VHPI/FLI) and simulator',
```
cmarqu, line 235:
```suggestion
        help='Print the absolute path of interface library for given interface (VPI/VHPI/FLI) and simulator',
```

cocotb/config.py
```diff
@@ -97,6 +99,50 @@ def help_vars_text():
...
 
+def lib_name(interface, simulator):
+
+    interface_name = interface.lower()
+    supported_interfaces = ["vpi", "vhpi", "fli"]
+    if interface_name not in supported_interfaces:
+        raise ValueError("Wrong interface used. Supported :" + ", ".join(supported_interfaces))
+
+    simulator_name = simulator.lower()
+    supported_sim = ["icarus", "questa", "modelsim", "ius", "xcelium", "vcs", "ghdl", "riviera", "activehdl", "cvc"]
```
cmarqu, line 110:
```suggestion
    supported_sims = ["icarus", "questa", "modelsim", "ius", "xcelium", "vcs", "ghdl", "riviera", "activehdl", "cvc"]
```
to be consistent with `supported_interfaces`.

cocotb/config.py
```diff
@@ -97,6 +99,50 @@ def help_vars_text():
     return helpmsg
 
 
+def lib_name(interface, simulator):
+
+    interface_name = interface.lower()
+    supported_interfaces = ["vpi", "vhpi", "fli"]
+    if interface_name not in supported_interfaces:
+        raise ValueError("Wrong interface used. Supported:" + ", ".join(supported_interfaces))
```
eric-wieser, line 107:
```suggestion
        raise ValueError("Wrong interface used. Supported: " + ", ".join(supported_interfaces))
```

cocotb/config.py
```diff
@@ -97,6 +99,50 @@ def help_vars_text():
...
+        simulator_name = "modelsim"
+
+    if simulator_name == "xcelium":
+        simulator_name = "ius"
+
+    if simulator_name == "cvc":
+        simulator_name = "modelsim"
+
+    if simulator_name in ["riviera", "activehdl"]:
+        simulator_name = "aldec"
```
eric-wieser, line 124: This doesn't allow `aldec` in the list of simulator names. Is this deliberate?

cocotb/config.py
```diff
@@ -97,6 +99,50 @@ def help_vars_text():
...
+        lib_ext = ".dll"
+
+    if simulator_name == "icarus":
+        lib_ext = ""
+
+    # check if compiled with msvc
+    if os.path.isfile(os.path.join(libs_dir, "cocotb.dll")):
+        lib_prefix = ""
+
+    return lib_prefix + "cocotb" + interface_name + "_" + simulator_name + lib_ext
```
eric-wieser, line 139: Slightly clearer as:
```suggestion
    if simulator_name == "icarus":
        lib_ext = ""
    elif os.name == "nt":
        lib_ext = ".dll"
    else:
        lib_ext = ".so"

    # check if compiled with msvc
    if os.path.isfile(os.path.join(libs_dir, "cocotb.dll")):
        lib_prefix = ""
    else:
        lib_prefix = "lib"

    return lib_prefix + "cocotb" + interface_name + "_" + simulator_name + lib_ext
```

cocotb/config.py
themperek: We have this:
https://github.com/cocotb/cocotb/blob/29845d6a0e7300f03a9fdc0d562d4551fbb86003/cocotb/share/makefiles/Makefile.deprecations#L9-L12

eric-wieser: Perhaps we should have separate `simulator_name` and `library_name` variables to make that distinction clearer.

cocotb/share/makefiles/simulators/Makefile.verilator
```diff
@@ -44,7 +44,7 @@ EXTRA_ARGS += --timescale $(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION)
 
 SIM_BUILD_FLAGS += -std=c++11
 
-COMPILE_ARGS += --vpi --public-flat-rw --prefix Vtop -o $(TOPLEVEL) -LDFLAGS "-Wl,-rpath,$(LIB_DIR) -L$(LIB_DIR) -lcocotbvpi_verilator"
+COMPILE_ARGS += --vpi --public-flat-rw --prefix Vtop -o $(TOPLEVEL) -LDFLAGS "-Wl,-rpath,$(shell cocotb-config --lib-dir) -L$(shell cocotb-config --lib-dir) -lcocotbvpi_verilator"
```
ktbarrett, line 47: I'm wondering if instead of `cocotb-config --lib-dir` it would be better to use `dirname $(cocotb-config --lib-name-path vpi verilator)`

cocotb/config.py
```diff
@@ -45,11 +46,16 @@
...
+__all__ = ["share_dir", "makefiles_dir", "libs_dir"]
 
 
 share_dir = os.path.join(os.path.dirname(cocotb.__file__), "share")
 makefiles_dir = os.path.join(os.path.dirname(cocotb.__file__), "share", "makefiles")
+libs_dir = os.path.join(os.path.dirname(cocotb.__file__), "libs")
+
+# On Windows use mixed mode "c:/a/b/c" as this work in all cases
+if os.name == "nt":
+    libs_dir = libs_dir.replace("\\", "/")
```
ktbarrett, line 58: Doesn't this only work in powershell? It might be worth noting (somewhere) that only bash and powershell are currently supported. I don't think that it's currently listed anywhere.

themperek: Both `\` and `/` are supposed on widows for very long time.

cocotb/share/makefiles/simulators/Makefile.verilator
themperek: It is rather to make custom flow bit easier and not our makefiles. No strong opinion.

cocotb/config.py
```diff
@@ -97,6 +103,59 @@ def help_vars_text():
     return helpmsg
 
 
+def lib_name(interface, simulator):
```
eric-wieser, line 106:
```suggestion
def lib_name(interface: str, simulator: str) : str:
```

cocotb/share/makefiles/simulators/Makefile.verilator
ktbarrett: Looks like Windows supports forward slashes at the API level, but cmd.exe doesn't; so it depends if the path goes in as a value, or if cmd.exe parses it first. I guess we shouldn't worry about this for now.

cocotb/config.py
ktbarrett: Looks like Windows supports forward slashes at the API level, but cmd.exe doesn't; so it depends if the path goes in as a value, or if cmd.exe parses it first. I guess we shouldn't worry about this for now.
https://stackoverflow.com/questions/10523708/why-does-the-cmd-exe-shell-on-windows-fail-with-paths-using-a-forward-slash

ktbarrett (approved):

cocotb/config.py
ktbarrett: Why add annotations here? This is an internal function and we aren't using mypy.

eric-wieser: Annotations are documentation too, and documentation isn't just for users :)

cocotb/config.py
```diff
@@ -143,19 +215,41 @@ def get_parser():
         action=PrintAction,
         text=help_vars_text(),
     )
+    parser.add_argument(
+        "--libpython",
+        help="prints the absolute path to the libpython associated with the current Python installation",
```
cmarqu, line 220:
```suggestion
        help="Print the absolute path to the libpython associated with the current Python installation",
```

documentation/source/newsfragments/2387.feature.rst
```diff
@@ -0,0 +1 @@
+Added ``--lib-dir``,  ``--lib-name`` and ``--lib-name-path`` to ``cocotb-config`` command.
```
cmarqu, line 1:
```suggestion
Added the ``--lib-dir``,  ``--lib-name`` and ``--lib-name-path`` options to the ``cocotb-config`` command to make cocotb integration into existing flows easier.
```

cocotb/config.py
```diff
@@ -97,6 +103,59 @@ def help_vars_text():
...
+        lib_prefix = ""
+    else:
+        lib_prefix = "lib"
+
+    return lib_prefix + "cocotb" + interface_name + "_" + library_name + lib_ext
+
+
+def lib_name_path(interface, simulator):
+    """
+    Returns the absolute path of interface library for given interface (VPI/VHPI/FLI) and simulator
```
cmarqu, line 148:
```suggestion
    Return the absolute path of interface library for given interface (VPI/VHPI/FLI) and simulator
```

cocotb/config.py
```diff
@@ -97,6 +103,59 @@ def help_vars_text():
     return helpmsg
 
 
+def lib_name(interface: str, simulator: str) -> str:
+    """
+    Returns the name of interface library for given interface (VPI/VHPI/FLI) and simulator.
```
cmarqu, line 108:
```suggestion
    Return the name of interface library for given interface (VPI/VHPI/FLI) and simulator.
```

PR #2392, opened 2021-02-09
Deploy source packages to PyPi

imphil:
Deploy to PyPi when pushing to master and stable branches, and clean up some GH Actions files while at it.


ktbarrett (commented):
Not to add to your plate, but `towncrier` should be called to do a final commit of the version's changelog.

imphil:
> towncrier should be called to do a final commit of the version's changelog.

That's something we'll need to do before pushing the v* release tag; when doing it within the release action we would create a commit after the tag. I'll think about a bit of scripting to do that in the right order, maybe we can somehow leverage some GitHub automation to call this script then (or alternatively I'll just call it locally).

Or is there another way?

imphil:
Unfortunately we can't see a test deployment until we tag some kind of release. It looks good so far, we can fix things if necessary in a tagged RC release.

ktbarrett:
You can push commits from Github Actions. You would have to tag it with some kind of "prepare release" tag, Github Actions would handle all the release stuff automatically, push the version and tag, publish to PyPi and deleted the "prepare" tag after releasing successfully. Something for next time, sounds involved.

.github/workflows/pypi-upload.yml
```diff
@@ -0,0 +1,46 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+name: Upload the cocotb package to PyPi
```
ktbarrett, line 5: Kind of a long name, gets chopped off on the Actions page.
```suggestion
name: Publish to PyPi
```

ktbarrett (approved):

themperek (approved):

PR #2399, opened 2021-02-10
 Fix coverage collection in CI

imphil:
* Properly collect and combine Python coverage on Windows.
  *On Windows, C coverage isn't collected yet as I haven't found out how
  to actually do that. Taking the same CFLAGS & Co. approach we do on
  Linux doesn't seem to do the trick.*
* Switch from the codecov-action GitHub Action to calling the codecov
  bash uploader directly. The codecov-action doesn't install the Python
  coverage package, which is required to collect our coverage data.
  Installing it before calling the action doesn't work, unfortunately,
  for not entirely clear reasons (`command -v coverage` does not return
  the path where the binary is installed; `command` is both a bash
  builtin and a PowerShell alias, and it also makes use of $PATH, so
  somewhere along those lines something is going wrong -- it's not clear
  what exactly is going on, however).
* Improve the `find` command line to avoid an additional `xargs`.

I'll open a follow-up issue to get C coverage looked into for Windows, but I'm running out of desire to do anything about that for now.

Now also includes a second commit to prevent codecov notifications from appearing before all CI jobs have reported back. That's more a workaround than a solution, but the best we can do for now.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,20 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Codecov configuration
+codecov:
+    notify:
+        # Codecov claims [1] to wait until all CI runs have completed, but that
+        # doesn't work for our CI setup, causing early reports to come in which
+        # indicate a drop in coverage. These reports are later updated as more
+        # reports come in. But by that time the first issue comment has already
+        # been made and an email has been sent. Prevent that by explicitly
+        # specifying the number of builds that need to be uploaded to codecov
+        # [2].
+        #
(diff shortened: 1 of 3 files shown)


ktbarrett:
That turned out to be more involved than I though, sorry. We do really need tox working correctly on Windows...

ktbarrett:
Still doesn't look like it's working.
https://github.com/cocotb/cocotb/runs/1872640134?check_suite_focus=true#step:26:36

imphil:
Painful as expected. I'll let you know once I have something working.

ktbarrett:
Might save time to comment out all the tests you aren't screwing with.

imphil:
OK, ready for review now, I also updated the PR description to reflect what we now have.

ktbarrett (approved):

PR #2400, opened 2021-02-10
Improve coverage

ktbarrett:
Adds collection omissions for `cocotb-config` and everything under `cocotb/_vendor/`. Also improves collection of `cocotb/__init__.py` by working around the issue of `_initialise_testbench` not being collected.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -154,7 +154,7 @@ def mem_debug(port):
     cocotb.memdebug.start(port)
 
 
-def _initialise_testbench(argv_):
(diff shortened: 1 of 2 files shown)


.coveragerc
```diff
@@ -0,0 +1,4 @@
+[run]
```
themperek, line 1: Maybe better `setup.cfg` than new file: https://coverage.readthedocs.io/en/coverage-5.0/config.html#configuration-reference ?

themperek (approved):
I imagine some will say to make separate PR for `with` but ..

PR #2402, opened 2021-02-10
Add Github Actions annotations for test failures

marlonjames:
Add annotations for failed tests so they can be seen from the summary page without diving into the failed job.

Edit: Now shows file path and line number of test function https://github.com/cocotb/cocotb/actions/runs/555861757


marlonjames:
I didn't get the full path to the file to display, but those links only work for a given PR's "Modified files" anyway.

eric-wieser:
> but those links only work for a given PR's "Files changed" anyway.

That's not true, there's a beta feature that makes it work on all files, even those not changed in the PR.

cmarqu (approved):

eric-wieser (approved):

PR #2403, opened 2021-02-10
Add codecov dependencies to CI

ktbarrett:
Python code coverage was no longer being collected after switching to the codecov-action. This is seemingly because the codecov-action does not attempt to install its dependencies. The coveragepy module is necessary for collecting Python coverage. See it complain [here](https://github.com/cocotb/cocotb/runs/1871844530?check_suite_focus=true#step:26:171) about not having that dependency.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -290,6 +290,9 @@ jobs:
         tox
 
     # codecov
+    - name: Install codecov dependencies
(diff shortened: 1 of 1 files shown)


imphil:
Thanks! I saw that as well while experimenting with the Windows coverage, let me fold that into https://github.com/cocotb/cocotb/pull/2399

PR #2404, opened 2021-02-11
asyncio-style queue implementations

ktbarrett:
Identical to and supersedes #2297. Adds type annotations and documentation. Pinging @alexforencich.

### DIFF
--- a/cocotb/queue.py
+++ b/cocotb/queue.py
@@ -1,0 +1,168 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+from typing import Generic, TypeVar
+import collections
+import heapq
+from asyncio import QueueEmpty, QueueFull
+
+import cocotb
+from cocotb.triggers import Event, _pointer_str
+
+T = TypeVar('T')
+
+
+class Queue(Generic[T]):
+    """A queue, useful for coordinating producer and consumer coroutines.
+
+    If *maxsize* is less than or equal to 0, the queue size is infinite. If it
+    is an integer greater than 0, then :meth:`put` will block when the queue
+    reaches *maxsize*, until an item is removed by :meth:`get`.
+    """
+
+    def __init__(self, maxsize: int = 0):
+
+        self._maxsize = maxsize
+
+        self._finished = Event()
+        self._finished.set()
+
+        self._getters = collections.deque()
+        self._putters = collections.deque()
+
+        self._init(maxsize)
+
+    def _init(self, maxsize):
+        self._queue = collections.deque()
+
+    def _put(self, item):
+        self._queue.append(item)
+
+    def _get(self):
+        return self._queue.popleft()
+
+    def _wakeup_next(self, waiters):
+        while waiters:
+            event, task = waiters.popleft()
+            if not task._finished:
+                event.set()
+                break
+
+    def __repr__(self):
+        return '<{} {} at {}>'.format(type(self).__name__, self._format(), _pointer_str(self))
+
+    def __str__(self):
+        return '<{} {}>'.format(type(self).__name__, self._format())
+
+    def __class_getitem__(cls, type):
+        return cls
+
+    def _format(self):
+        result = 'maxsize={}'.format(repr(self._maxsize))
+        if getattr(self, '_queue', None):
+            result += ' _queue={}'.format(repr(list(self._queue)))
+        if self._getters:
+            result += ' _getters[{}]'.format(len(self._getters))
+        if self._putters:
+            result += ' _putters[{}]'.format(len(self._putters))
+        return result
+
+    def qsize(self) -> int:
+        """Number of items in the queue."""
+        return len(self._queue)
+
+    @property
+    def maxsize(self) -> int:
+        """Number of items allowed in the queue."""
+        return self._maxsize
+
+    def empty(self) -> bool:
+        """Return ``True`` if the queue is empty, ``False`` otherwise."""
+        return not self._queue
+
+    def full(self) -> bool:
+        """Return ``True`` if there are :meth:`maxsize` items in the queue.
+
+        .. note::
+            If the Queue was initialized with ``maxsize=0`` (the default), then
+            :meth:`full` is never ``True``.
+        """
+        if self._maxsize <= 0:
+            return False
+        else:
+            return self.qsize() >= self._maxsize
+
+    async def put(self, item: T) -> None:
+        """Put an *item* into the queue.
+
+        If the queue is full, wait until a free
+        slot is available before adding the item.
+        """
+        while self.full():
+            event = Event('{} put'.format(type(self).__name__))
+            self._putters.append((event, cocotb.scheduler._current_task))
+            await event.wait()
+        self.put_nowait(item)
+
+    def put_nowait(self, item: T) -> None:
+        """Put an *item* into the queue without blocking.
+
+        If no free slot is immediately available, raise :exc:`asyncio.QueueFull`.
+        """
+        if self.full():
+            raise QueueFull()
+        self._put(item)
+        self._finished.clear()
+        self._wakeup_next(self._getters)
+
+    async def get(self) -> T:
+        """Remove and return an item from the queue.
+
+        If the queue is empty, wait until an item is available.
+        """
+        while self.empty():
+            event = Event('{} get'.format(type(self).__name__))
+            self._getters.append((event, cocotb.scheduler._current_task))
+            await event.wait()
+        return self.get_nowait()
+
+    def get_nowait(self) -> T:
+        """Remove and return an item from the queue.
+
+        Return an item if one is immediately available, else raise
+        :exc:`asyncio.QueueEmpty`.
+        """
+        if self.empty():
+            raise QueueEmpty()
+        item = self._get()
+        self._wakeup_next(self._putters)
+        return item
+
+
+class PriorityQueue(Queue):
+    r"""A subclass of :class:`Queue`; retrieves entries in priority order (smallest item first).
+
+    Entries are typically tuples of the form ``(priority number, data)``.
+    """
+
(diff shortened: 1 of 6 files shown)


tests/test_cases/test_cocotb/test_queues.py
```diff
@@ -0,0 +1,279 @@
...
+    assert not q.empty()
+    assert q.full()
+
+    # overflow
+    got_exception = False
+    try:
+        q.put_nowait(100)
+    except QueueFull:
+        got_exception = True
+    assert got_exception
```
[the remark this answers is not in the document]
ktbarrett: I'm thinking I should replace this with `pytest.raises` instead.

tests/test_cases/test_cocotb/test_queues.py
```diff
@@ -0,0 +1,279 @@
...
+    getter_list = []
+
+    # test put contention
+    for k in range(NUM_PUTTERS):
+        cr_list.append(cocotb.fork(putter(putter_list, k)))
+
+    assert q.qsize() == QUEUE_SIZE
+
+    # test killed putter
+    cr = cocotb.fork(putter(putter_list, 100))
```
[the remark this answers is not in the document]
ktbarrett: Sure

cocotb/queue.py
```diff
@@ -0,0 +1,204 @@
...
+            result += ' tasks={}'.format(self._unfinished_tasks)
+        return result
+
+    def qsize(self) -> int:
+        """Number of items in the queue."""
+        return len(self._queue)
+
+    @property
+    def maxsize(self) -> int:
+        """Number of items allowed in the queue."""
```
cmarqu, line 80: Not sure if the "infinite if <= 0" behavior should be mentioned here once more?

cocotb/queue.py
```diff
@@ -0,0 +1,204 @@
...
+            raise QueueFull()
+        self._put(item)
+        self._unfinished_tasks += 1
+        self._finished.clear()
+        self._wakeup_next(self._getters)
+
+    async def get(self) -> T:
+        """Remove and return an item from the queue.
+
+        If queue is empty, wait until an item is available.
```
cmarqu, line 126:
```suggestion
        If the queue is empty, wait until an item is available.
```

marlonjames:
Could you improve the tests to get full coverage?

tests/test_cases/test_cocotb/test_queues.py
```diff
@@ -0,0 +1,268 @@
...
+    # remove value that forced contention
+    assert q.get_nowait() is None, "Popped unexpected value"
+
+    # test fair scheduling by ensuring that each putter is serviced for its first
+    # write before the second write on any putter is serviced.
+    for _ in range(NUM_PUTS):
+        remaining = set(range(NUM_PUTTERS))
+        for _ in range(NUM_PUTTERS):
+            v = await q.get()
+            assert v in remaining, "Unfair scheduling occured"
```
cmarqu, line 179:
```suggestion
            assert v in remaining, "Unfair scheduling occurred"
```

cmarqu (approved):
Just these tiny fixes...

cocotb/queue.py
```diff
@@ -0,0 +1,204 @@
...
+        if self._maxsize <= 0:
+            return False
+        else:
+            return self.qsize() >= self._maxsize
+
+    async def put(self, item: T) -> None:
+        """Put an *item* into the queue.
+
+        If the queue is full, wait until a free
+        slot is available before adding item.
```
cmarqu, line 103:
```suggestion
        slot is available before adding the item.
```

cmarqu (commented):
Two more minor things from looking at the asyncio docs (which don't seem to be overly consistent in the markup).

ktbarrett:
@cmarqu Documentation was from the asyncio docstrings. Most of the documentation issues exist there as well. https://github.com/python/cpython/blob/master/Lib/asyncio/queues.py

cocotb/queue.py
```diff
@@ -0,0 +1,204 @@
...
+    def __str__(self):
+        return '<{} {}>'.format(type(self).__name__, self._format())
+
+    def __class_getitem__(cls, type):
+        return cls
+
+    def _format(self):
+        result = 'maxsize={}'.format(repr(self._maxsize))
+        if getattr(self, '_queue', None):
+            result += ' _queue={}'.format(repr(list(self._queue)))
```
ktbarrett, line 65: I'm not entirely sure how this line isn't being covered. The base implementation has a `_queue` attribute.

ktbarrett (commented):
Is there any objection to me removing the task pool features? `task_done`, `join`, and `_unfinished_tasks`? I mentioned how I seriously doubt they are going to be useful in a cocotb context before, and it's manual and easy to screw up as well, and there is no test for it.

alexforencich:
My thought would be to keep them so that it's directly compatible with the asyncio queue.  But I suppose if the cocotb queue implements a subset of the functionality, then there should be no problem with substituting the asyncio queue if/when the scheduler supports asyncio code.

alexforencich:
And why is "fixes to timer docstring" in here?

ktbarrett:
@alexforencich I noticed it and didn't want to create a separate PR for a 1 line fix. Now its 4 lines, but things get away from you 🤷.

cocotb/queue.py
marlonjames: I think it's because the `_queue` is empty at that point in the test, so the `if` is `False`.

marlonjames:
I would be fine with removing the task tracking functionality. The basic features (and Exceptions) are compatible with `asyncio.Queue`, and we wouldn't include the task stuff if it wasn't in `asyncio`. Make it useful for cocotb but compatible where it makes sense.

eric-wieser:
I don't really understand the argument for removing the task features - if they're useful in asyncio, why couldn't they also be useful in cocotb?

ktbarrett:
1. There isn't a practical use case in a testbench. At least I can't think of one that isn't terribly contrived.
2. Breaks SRP.
3. It's not a very good implementation (must manually call `task_done` to signal a consumer has finished and anyone can do it, consumer must correctly handle possibility of exceptions to prevent it causing deadlock, there is no mapping of a particular input to task finishing, it's easy for beginners to misuse it, I could go on...).

If it's actually useful, we can add first class support for the functionality that avoids a lot of the design issues, and not add some hacked on afterthought that was obviously designed with the old threading system in mind and has been propagated since the early 00's.

cocotb/queue.py
ktbarrett: I don't think that would have an affect on the value returned, so I'm sure it needs to be mentioned.

marlonjames:
Should `__repr__` use `cocotb.triggers._pointer_str` ?

ktbarrett:
@marlonjames It could. Do you have a recommendation?

cocotb/queue.py
```diff
@@ -0,0 +1,168 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+from typing import Generic, TypeVar
+import collections
+import heapq
+from asyncio import QueueEmpty, QueueFull
+
+import cocotb
+from cocotb.triggers import Event
```
marlonjames, line 10:
```suggestion
from cocotb.triggers import Event, _pointer_str
```

cocotb/queue.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def _wakeup_next(self, waiters):
+        while waiters:
+            event, task = waiters.popleft()
+            if not task._finished:
+                event.set()
+                break
+
+    def __repr__(self):
+        return '<{} at {:#x} {}>'.format(type(self).__name__, id(self), self._format())
```
marlonjames, line 52:
```suggestion
        return '<{} {} at {}>'.format(type(self).__name__, self._format(), _pointer_str(self))
```

tests/test_cases/test_cocotb/test_queues.py
```diff
@@ -0,0 +1,255 @@
...
+    q = Queue[int](maxsize=1)
+
+    q.put_nowait(0)
+    putter = cocotb.fork(q.put(1))
+
+    s = repr(q)
+    assert "maxsize" in s
+    assert "_queue" in s
+    assert "_putters" in s
+
```
marlonjames, line 244:
```suggestion
    assert str(q)[:-1] in s

```

tests/test_cases/test_cocotb/test_queues.py
```diff
@@ -0,0 +1,255 @@
...
+    assert q.get_nowait() == 0
+    getter = cocotb.fork(q.get())
+
+    s = repr(q)
+    assert "_putters" not in s
+    assert "_getters" in s
+
```
marlonjames, line 251:
```suggestion
    assert str(q)[:-1] in s

```

tests/test_cases/test_cocotb/test_queues.py
```diff
@@ -0,0 +1,255 @@
...
+    await getter
+
+    s = repr(q)
+    assert "_getters" not in s
```
marlonjames, line 255:
```suggestion
    assert "_getters" not in s
    assert str(q)[:-1] in s
```

marlonjames (approved):
Suggested change to match repr format of triggers, and get the last bit of coverage.

tests/test_cases/test_cocotb/test_queues.py
```diff
@@ -0,0 +1,255 @@
...
+    assert ref_q.qsize() == 0
+
+
+factory = TestFactory(run_queue_blocking_test)
+factory.add_option("queue_type", [Queue, PriorityQueue, LifoQueue])
+factory.generate_tests()
+
+
+@cocotb.test()
+async def test_repr(_):
```
ktbarrett, line 234:
```suggestion
async def test_str_and_repr(_):
```

ktbarrett:
Merging tomorrow if there are no further comments.

Issue #2384, opened 2021-02-02
Cocotb New Version Not Working with Questa VHPI


raysalemi:
Hi All,

The most recent version of Questa (2021.1)  implements enough VHPI to work with Cocotb (as a first step.)  Our tests show that the beta version of cocotb fails with Questa 2021.1 while the current version works. 

The Questa team would like to know the next steps to take to help debug this issue. Could you suggest next steps? For example, if we were to push our modifications to cocotb to a branch, would someone be able to debug it with Questa 2021.1?

Thanks!
Ray

PS. For those who don't know: I am the Aerospace and Defense Solutions Manager for Siemens EDA DVT division (I'm in marketing.) We make Questa.

ktbarrett:
@raysalemi Rick White informed me that I (Kaleb Barrett of Raytheon Technologies) would be testing 2021.1 internally before its release. I helped the team get started running the build against our regressions last week, after which they were seeing numerous errors.

The first thing I'd like to do is add the Questa VHPI to our build. Previously, I think the build team was working with the Aldec-specific VHPI library. Simulator-specific hacks in each implementation of our GPI means that what works for them, might not work for you. A recent change may have caused the break. I can have a look if I can get access to the build.

ktbarrett:
I have news that the tool is in the final stages of the approval process and I will have it to test soon.

raysalemi:
So you can use the Raytheon build to test cocotb?

Do you need us to push a branch with our mods to the build scripts?

ktbarrett:
If you could. Thank you.

themperek:
Sorry for off-topic. Maybe you can get a license so we can include Questa in CI. This can be private as is done for Aldec now.

raysalemi:
> Sorry for off-topic. Maybe you can get a license so we can include Questa in CI. This can be private as is done for Aldec now.

We've been wondering if you had a license.  Let me look into it.

ktbarrett:
Neither us personally or our care-taking organization, the [FOSSi Foundation](http://www.fossi-foundation.org/), have a Questa license for use in our CI. It would __also__ be very useful for [building binary packages for distribution that support the FLI](https://github.com/cocotb/cocotb/issues/2318#issuecomment-763793708). Pinging @imphil.

raysalemi:
Could you tell me more about the FOSSi Foundation and its relationship to
Cocotb?  I also hear talk of a "Cocotb Consortium".  Does that exist and
[...]

PR #2408, opened 2021-02-15
Add support for the Questa VHPI

ktbarrett:
Adds support to build the VHPI library for Questa, and adds support to makefiles for starting questa with the VHPI.

There are two ways IMO to support switching between VHPI and FLI at the top level.
1. Use two different `TOPLEVEL_LANG` values: `vhdlvhpi` and `vhdlfli`, both `.startwith("vhdl")` but also list the interface to use
2. Introduce a second variable `TOPLEVEL_INTERFACE` that can be `vpi`, `vhpi`, or `fli`, but will only be used for Questa (right now).

It seems to me `TOPLEVEL_LANG` is a misnomer. Users and the script shouldn't care about the language used, but the scripts care about the interface to use. I guess this was chosen because selecting a language is easier for users to understand. Of course we end up in trouble now that there are two different interface for a single language.

I chose 1 because I think `TOPLEVEL_LANG` should be an interface selector, and that we can control default interface selections for Verilog and VHDL per simulator by specifying `verilog` or `vhdl`, while also supporting exact interfaces. I could be talked into some other option. I also don't know if `startswith("vhdl")` is really important and if we should just rename them `vhpi` and `fli`.

P.S. this isn't the only change necessary to get Questa VHPI to run through regressions, many makefiles support only `verilog`/`vhdl` and they would need to be updated. This is fine because Questa 2021.1 does not support cocotb yet.

Pinging @raysalemi.

TODO
- [x] Add VHPI definitions to `modelsim.def`
- [ ] Refactor to use `TOPLEVEL_INTERFACE` variable, have the value of `TOPLEVEL_INTERFACE` inferred from `TOPLEVEL_LANG`


raysalemi:
One note.  While the 1.5 VHPI bug needs to be fixed so that Mentor can continue development, the tox tests do not pass yet with VHPI.  Questa 2021.1 currently supports all the example designs, but there is additional development needed to support the rest of CocoTB. 

The current 1.5 bug is blocking development.

themperek:
> There are two ways IMO to support switching between VHPI and FLI at the top level.
> 
> 1. Use two different `TOPLEVEL_LANG` values: `vhdlvhpi` and `vhdlfli`, both `.startwith("vhdl")` but also list the interface to use
> 2. Introduce a second variable `TOPLEVEL_INTERFACE` that can be `vpi`, `vhpi`, or `fli`, but will only be used for Questa (right now).

3. Add `VSIM_VHDL_INTERFACE` or similar. It looks like the problem will be Questa only issue and bit experimental for some time. In this case, you may avoid changing all other files.

marlonjames:
`modelsim.def` needs updating for the VHPI functions

ktbarrett:
@themperek That is basically my second suggestion, to have a separate variable to select which interface to use. I don't want something Questa-specific, GHDL may gain a VHPI interface.

@raysalemi Could you report the bug? I'm not yet sure what you mean.

raysalemi:
> @themperek That is basically my second suggestion, to have a separate variable to select which interface to use. I don't want something Questa-specific, GHDL may gain a VHPI interface.
> 
> @raysalemi Could you report the bug? I'm not yet sure what you mean.

@ktbarrett , The bug is #2384

themperek (approved):
BTW: Check on Windows and seems to be working on simple example.

ktbarrett:
Went with @themperek's suggestion of a variable to decide which interface to use with VHDL designs. This applies to both the initially loaded library as well as `GPI_EXTRA` in mixed language sims.

marlonjames (commented):
It's hardcoded to `cocotbfli_entry_point` for both `vhpi` 
and `fli`.

ktbarrett:
Shoot. Will fix.

PR #2409, opened 2021-02-15
Improve Verilator doc section

cmarqu:

### DIFF
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -335,7 +335,7 @@ In order to use this simulator, set :make:var:`SIM` to ``ghdl``:
 
 .. warning::
 
-    ghdl support is currently experimental. Some features of cocotb may not work correctly or at all.
+    GHDL support is currently experimental. Some features of cocotb may not work correctly or at all.
 
 Noteworthy is that despite GHDL being a VHDL simulator, it implements the :term:`VPI` interface.
 
--- a/documentation/source/simulator_support.rst
+++ b/documentation/source/simulator_support.rst
@@ -89,6 +89,10 @@ To run cocotb with Verilator, you need ``verilator`` in your PATH.
 
     cocotb requires Verilator 4.106 or later.
 
+.. warning::
+
+    Verilator support is currently experimental. Some features of cocotb may not work correctly or at all.
+
 .. versionadded:: 1.3
 
 .. versionchanged:: 1.5 Improved cocotb support and greatly improved performance when using a higher time precision. Verilator 4.106 or later is required.
@@ -102,7 +106,7 @@ To enable :term:`HDL` code coverage, add Verilator's coverage option(s) to the :
 
     EXTRA_ARGS += --coverage
 
-This will result in coverage data being written to ``coverage.dat``.
+This will result in coverage data being written to :file:`coverage.dat`.
 
 .. _sim-verilator-waveforms:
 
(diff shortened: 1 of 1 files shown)


documentation/source/simulator_support.rst
```diff
@@ -91,7 +91,7 @@ To run cocotb with Verilator, you need ``verilator`` in your PATH.
 
 .. versionadded:: 1.3
 
-.. versionchanged:: 1.5 Improved cocotb support and greatly improved performance when using a higher time precision. Verilator 4.106 or later is required.
```
ktbarrett, line 94: I think removing this disclaimer is not a good idea. I also think any improvement to the Verilator docs include a strong warning about Verilator support for cocotb being unstable.

documentation/source/simulator_support.rst
```diff
@@ -102,7 +102,7 @@ To enable :term:`HDL` code coverage, add Verilator's coverage option(s) to the :
 
     EXTRA_ARGS += --coverage
 
-This will result in coverage data being written to ``coverage.dat``.
+This will result in coverage data being written to :file:`coverage.dat`.
```
ktbarrett, line 105: I think this will try to link against the non-existent file coverage.dat. Best to leave it where it is.

documentation/source/simulator_support.rst
```diff
@@ -122,14 +122,13 @@ A VCD file named ``dump.vcd`` will be generated in the current directory.
...
-For Verilator 4.102 and above, the `-CFLAGS -DVM_TRACE_FST=1` argument is no longer necessary.
+To enable FST tracing, add ``--trace-fst`` to :make:var:`EXTRA_ARGS` as shown below.
 
   .. code-block:: make
 
-    EXTRA_ARGS += --trace-fst --trace-structs -CFLAGS -DVM_TRACE_FST=1
+    EXTRA_ARGS += --trace-fst --trace-structs
 
-The resulting file will be ``dump.fst`` and can be opened by ``gtkwave dump.fst``.
+The resulting file will be :file:`dump.fst` and can be opened by ``gtkwave dump.fst``.
```
ktbarrett, line 131: Same here.

ktbarrett (commented):
Removing the bit about Verilator 4.102 and FST is fine, seeing as we active prevent usage of Verilator less than 4.106.

documentation/source/simulator_support.rst
```diff
@@ -91,7 +91,7 @@ To run cocotb with Verilator, you need ``verilator`` in your PATH.
```
marlonjames, line 91: Something like this?

```suggestion

.. warning::

    Verilator support is currently experimental. Some features of cocotb may not work correctly or at all.

```

documentation/source/simulator_support.rst
marlonjames: Although it is repeated from the note above, having it in the `versionchanged` tag tracks the history of supported versions.

marlonjames: Although it is repeated from the note above, having it in the `versionchanged` tag tracks the history of supported versions. Not critical though.

documentation/source/simulator_support.rst
cmarqu: To me, the support for curly braces to support variable parts (https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-file) suggests that this role is meant for semantic markup, not necessarily to reference existing files; at least that seemed to be the initial idea.

documentation/source/simulator_support.rst
cmarqu: Agreed, especially since there is this `changes` Sphinx builder which can extract these comments.

ktbarrett:
@cmarqu While you are here could you add the same warning to GHDL?

documentation/source/simulator_support.rst
ktbarrett: I see. I am seeing warnings related to the referencing of options that weren't declared and assumed this would have a similar issue.

documentation/source/simulator_support.rst
cmarqu: Reinstated.

documentation/source/simulator_support.rst
```diff
@@ -330,7 +333,10 @@ In order to use this simulator, set :make:var:`SIM` to ``ghdl``:
 
     make SIM=ghdl
 
-Support is preliminary.
+.. warning::
+
+    ghdl support is currently experimental. Some features of cocotb may not work correctly or at all.
```
ktbarrett, line 338:
```suggestion
    GHDL support is currently experimental. Some features of cocotb may not work correctly or at all.
```

ktbarrett (approved):

PR #2411, opened 2021-02-17
Add documentation checkers, do spellchecking

cmarqu:
With the `en_US` dictionary that I have installed locally, the only spellchecker finding that may still occur is the following example which is easily ignored mentally (I wouldn't want to be so strict as to require zero findings; they are warnings):
```
cocotb/documentation/source/release_notes.rst:2: Spell check: ae: ["eye", "ea", "aw", "a", "e", "nae", "are", "ate", "ale", "ace", "age", "ape", "aye", "ave", "awe", "AXI", "AMS", "ns", "pre", "ps"]: cocotb 1.5.0.dev298+g5a6ae305.d20210217 (2021-02-17)
```


documentation/source/install_devel.rst
```diff
@@ -23,7 +23,7 @@ than the stable version:
    * pytest
 
 * Python 3.5+
-* Python-dev packages
```
eric-wieser, line 26: I think this might be referring to `apt install python-dev` or similar, but I'm not sure

documentation/source/spelling_wordlist.txt
```diff
@@ -37,55 +41,91 @@ Wiki
...
 backpressure
+bugfixes
+builtin
 callgraph
 cartesian
 cocotb
 coroutine
 coroutines
-dev
+deliminated
```
eric-wieser, line 51: `delimited` is a better word here anyway

documentation/source/install_devel.rst
cmarqu: We later list the concrete package names, so I figured this change would be understandable for readers.

ktbarrett (approved):

documentation/source/spelling_wordlist.txt
```diff
@@ -37,55 +41,90 @@ Wiki
...
+bugfixes
+builtin
 callgraph
 cartesian
 cocotb
 coroutine
 coroutines
-dev
 diff
+disrecommended
```
eric-wieser, line 52: I've never seen this word before, and I don't think its in widespread use. "discouraged" is almost certainly more idiomatic.

cmarqu: Just when I thought I could enrich the world with a new word!

PR #2423, opened 2021-02-23
Run 'pyupgrade --py36-plus' on all *.py files

cmarqu:
Depends on #2422.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -140,7 +140,7 @@ class RunningTask:
         return self
 
     def __str__(self):
-        return "<{}>".format(self.__name__)
+        return f"<{self.__name__}>"
 
     def _get_coro_stack(self):
         """Get the coroutine callstack of this Task."""
@@ -305,7 +305,7 @@ class RunningTest(RunningCoroutine):
         self.handler = RunningTest.ErrorLogHandler(self.error_messages.append)
 
     def __str__(self):
-        return "<{}>".format(self.__name__)
+        return f"<{self.__name__}>"
 
     def _advance(self, outcome):
         if not self.started:
@@ -314,7 +314,7 @@ class RunningTest(RunningCoroutine):
             self.start_time = time.time()
             self.start_sim_time = get_sim_time('ns')
             self.started = True
-        return super(RunningTest, self)._advance(outcome)
+        return super()._advance(outcome)
 
     # like RunningTask.kill(), but with a way to inject a failure
     def abort(self, exc):
@@ -333,13 +333,13 @@ class RunningTest(RunningCoroutine):
             raise InternalError("Outcome already has a value, but is being set again.")
         outcome = outcomes.Error(exc)
         if _debug:
-            self.log.debug("outcome forced to {}".format(outcome))
+            self.log.debug(f"outcome forced to {outcome}")
         self._outcome = outcome
         cocotb.scheduler._unschedule(self)
 
     def sort_name(self):
         if self.stage is None:
-            return "%s.%s" % (self.module, self.funcname)
+            return f"{self.module}.{self.funcname}"
         else:
             return "%s.%d.%s" % (self.module, self.stage, self.funcname)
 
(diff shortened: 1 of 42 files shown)


ktbarrett:
Oh no the 3.5 test is failing!

ktbarrett (commented):
Any idea why it uses `.format()` in some cases and f-strings in others?

marlonjames:
@ktbarrett Maybe it does `%s` -> `.format()` and `.format()` -> `f"{}"`, so those need a second pass?

cocotb/decorators.py
```diff
@@ -333,13 +333,13 @@ def abort(self, exc):
...
+            self.log.debug(f"outcome forced to {outcome}")
         self._outcome = outcome
         cocotb.scheduler._unschedule(self)
 
     def sort_name(self):
         if self.stage is None:
-            return "%s.%s" % (self.module, self.funcname)
+            return f"{self.module}.{self.funcname}"
         else:
             return "%s.%d.%s" % (self.module, self.stage, self.funcname)
```
marlonjames, line 344: It missed this.

cocotb/regression.py
```diff
@@ -188,7 +188,7 @@ def _discover_tests() -> Iterable[Test]:
                         test = getattr(module, test_name)
                     except AttributeError:
                         _logger.error("Requested test %s wasn't found in module %s", test_name, module_name)
```
marlonjames, line 190: Looks like it doesn't find this type of logging that uses `%s` but not the `%` operator.

ktbarrett: Well that is used by the logger so that is expected.

ktbarrett:
@cmarqu I just ran the command on your branch and it fixed even more stuff. Are you using the most up to date version?

ktbarrett:
From [their docs](https://github.com/asottile/pyupgrade#f-strings).

> note: pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are anything but simple names or dotted names (as this can decrease readability).

cocotb/decorators.py
cmarqu: This is indeed interesting.

cocotb/regression.py
cmarqu: Yes, it appears to leave loggers lazy evaluation alone.

cmarqu:
> From [their docs](https://github.com/asottile/pyupgrade#f-strings).
> 
> > note: pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are anything but simple names or dotted names (as this can decrease readability).
> 
> I think we are going to have to wipe those manually, or find another tool.

This fork here [supposedly](https://github.com/getsentry/sentry/pull/23506) is more aggressive: https://github.com/joshuarli/pyupgrade/tree/joshuarli/individual-fixers

ktbarrett:
Needs rebase.

tests/test_cases/test_array/test_array.py
```diff
@@ -62,7 +62,7 @@ async def test_read_write(dut):
         _check_int(tlog, dut.param_char, ord('p'))
         _check_str(tlog, dut.param_str , b"ARRAYMOD")
 
-        if not cocotb.SIM_NAME.lower().startswith(("riviera")):
+        if not cocotb.SIM_NAME.lower().startswith("riviera"):
```
imphil, line 65: Slightly unsure about that one -- we are trying to always pass a set here. But no strong feelings either way.

imphil (commented):
This PR contains a couple changes where we went from `func((stuff))` to `func(stuff)`. I didn't double-check them, but it feels like the current code was actually intended that way. Could you have a look?

tests/test_cases/test_array/test_array.py
ktbarrett: `(value)` is equivalent to `value`. What should have been here was `(value,)` which is a tuple literal. But it is correct for now, `startswith` can take a single string or a tuple/list of strings.

cmarqu: Indeed, I think this was a mistaken way of preparing for future extensions of that set.

ktbarrett (approved):
It's an automated process and the regression pass. I seriously doubt it go anything wrong. And it looks like it at least sort of helped remove out of date styles. It did totally miss the attribute annotations in handle.py, and a lot of the Python 2 string formatting. But we can do that in a follow on.

ktbarrett:
Feel free to merge whenever.

PR #2424, opened 2021-02-23
Convert type annotations for variables to Py3.6 style

cmarqu:
See also #2069.
Depends on #2422.

Doc rendering: https://cocotb--2424.org.readthedocs.build/en/2424/library_reference.html#other-runtime-information

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -37,7 +37,7 @@ import threading
 import random
 import time
 import warnings
-from typing import Dict, List, Union
+from typing import Dict, List, Optional, Union
 from collections.abc import Coroutine
 
 import cocotb.handle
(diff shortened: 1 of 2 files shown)


cocotb/__init__.py
```diff
@@ -96,31 +96,31 @@ def _reopen_stream_with_buffering(stream_name):
...
-argc = None  # type: int
+argc: Optional[int] = None
 """The length of :data:`cocotb.argv`"""
 
-plusargs = None  # type: Dict[str, Union[bool, str]]
+plusargs: Optional[Dict[str, Union[bool, str]]] = None
 """A dictionary of "plusargs" handed to the simulation. See :make:var:`PLUSARGS` for details."""
 
-LANGUAGE = os.getenv("TOPLEVEL_LANG")  # type: str
+LANGUAGE: str = os.getenv("TOPLEVEL_LANG")
```
ktbarrett, line 114:
```suggestion
LANGUAGE: Optional[str] = os.getenv("TOPLEVEL_LANG")
```

`getenv` can return `None`.

ktbarrett (commented):
While this is more correct in terms of type checking, the variables cannot be `None` when a cocotb test is executing. Saying it's `Optional` might be muddying the waters. @eric-wieser thoughts?

cocotb/__init__.py
cmarqu: Fixed.

eric-wieser:
`cocotb` is still importable when not in a test, right? For that case, the optional annotation is correct. But if we eventually expose mypy annotations to users, then they'd get errors in their tests when trying to use these attributes.

I think marking them optional is fine as long as we indicate in the docs that they're guaranteed to hold a value at test-time. We can worry about mypy later.

ktbarrett:
Needs rebase.

cmarqu:
> I think marking them optional is fine as long as we indicate in the docs that they're guaranteed to hold a value at test-time. 

I have added several such notes; hopefully to the right ones. Please re-review.

cocotb/__init__.py
```diff
@@ -96,31 +96,43 @@ def _reopen_stream_with_buffering(stream_name):
...
 
-SIM_VERSION = None  # type: str
+SIM_VERSION: Optional[str] = None
 """The version of the running simulator. ``None`` if :mod:`cocotb` was not loaded from a simulator"""
 
-RANDOM_SEED = None  # type: int
+RANDOM_SEED: Optional[int] = None
 """
 The value passed to the Python default random number generator.
 See :envvar:`RANDOM_SEED` for details on how the value is computed.
```
ktbarrett, line 138: There is no message here?

cocotb/__init__.py
```diff
@@ -96,31 +96,43 @@ def _reopen_stream_with_buffering(stream_name):
...
+
+argc: Optional[int] = None
+"""The length of :data:`cocotb.argv`
+
+This is guaranteed to hold a value at test time.
+"""
 
-plusargs = None  # type: Dict[str, Union[bool, str]]
+plusargs: Optional[Dict[str, Union[bool, str]]] = None
 """A dictionary of "plusargs" handed to the simulation. See :make:var:`PLUSARGS` for details."""
```
ktbarrett, line 124: Or here?

cocotb/__init__.py
cmarqu: Fixed.

cocotb/__init__.py
cmarqu: Fixed.

ktbarrett (approved):

PR #2316, opened 2021-01-02
set_signal_value_long() bugfixes and improvements

elgorwi:
Some bugfixes and improvements for setting signals using python ints: 
closes #1202 
closes #913 
closes #268 

This PR is essentially a merge of PR #1792 , PR #1742  and #1744 .
For more details refer to the linked issues and PRs.

PR #2425, opened 2021-02-24
Revert behavioral change for 1.5

ktbarrett:
Partial revert of #2316, keeps C++ level changes. Still solves #913 by doing the right thing when negative values are assigned to handles. No longer throws an `OverflowError` when a value larger than the signal is assigned to handle. Continues the tried and true silent truncation of values, but throws a `FutureWarning` stating that it won't be allowed in 2.0.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -723,7 +723,7 @@ class ModifiableObject(NonConstantObject):
             TypeError: If target has an unsupported type for value assignment.
 
         .. deprecated:: 1.5
-            Automatic truncation of values that exceed the size of the signal will result in a
+            Truncation of values that exceed the size of the signal will result in a
             :exc:`OverflowError` starting in 2.0.
 
         .. deprecated:: 1.5
@@ -735,9 +735,9 @@ class ModifiableObject(NonConstantObject):
 
         if isinstance(value, int):
             min_val, max_val = _value_limits(len(self), _Limits.VECTOR_NBIT)
-            if value < min_val or max_val < value:
+            if value < min_val or value > max_val:
                 warnings.warn(
-                    "Automatic truncation of values that exceed the size of the signal will raise an `OverflowError` starting in 2.0.",
+                    "Truncation of values that exceed the size of the signal will raise an `OverflowError` starting in 2.0.",
                     FutureWarning, stacklevel=3)
             if len(self) <= 32:
                 call_sim(self, self._handle.set_signal_val_int, set_action, value)
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -722,8 +722,9 @@ class ModifiableObject(NonConstantObject):
         Raises:
             TypeError: If target has an unsupported type for value assignment.
 
-            OverflowError: If int value is out of the range that can be represented
-                by the target. -2**(Nbits - 1) <= value <= 2**Nbits - 1
+        .. deprecated:: 1.5
+            Automatic truncation of values that exceed the size of the signal will result in a
+            :exc:`OverflowError` starting in 2.0.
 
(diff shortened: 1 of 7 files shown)


ktbarrett:
Added a test to cover assignment with both positive and negative integers.

documentation/source/release_notes.rst
```diff
@@ -102,8 +102,7 @@ Deprecations and Removals
 Changes
 -------
 
-- Assigning out-of-range Python integers to signals would previously truncate the value silently for signal widths <= 32 bits and truncate the value with a warning for signal widths > 32 bits.
-  Assigning out-of-range Python integers to signals will now raise an :exc:`OverflowError`. (:pr:`913`)
+- Assigning negative Python integers to handles does an implicit two's compliment conversion. (:pr:`913`)
```
cmarqu, line 105:
```suggestion
- Assigning negative Python integers to handles does an implicit two's complement conversion. (:pr:`913`)
```

cmarqu (approved):

ktbarrett:
Will extend test to *actually* hit all coverage points tonight.

cocotb/handle.py
```diff
@@ -722,8 +722,9 @@ def _set_value(self, value, call_sim):
         Raises:
             TypeError: If target has an unsupported type for value assignment.
 
-            OverflowError: If int value is out of the range that can be represented
-                by the target. -2**(Nbits - 1) <= value <= 2**Nbits - 1
+        .. deprecated:: 1.5
+            Automatic truncation of values that exceed the size of the signal will result in a
```
imphil, line 726:
```suggestion
            Truncation of values that exceed the size of the signal will result in a
```

cocotb/handle.py
```diff
@@ -734,19 +735,17 @@ def _set_value(self, value, call_sim):
...
-                    call_sim(self, self._handle.set_signal_val_int, set_action, value)
-                    return
-
-                if value < 0:
-                    value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.TWOS_COMPLEMENT)
-                else:
-                    value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.UNSIGNED)
+            if value < min_val or max_val < value:
+                warnings.warn(
+                    "Automatic truncation of values that exceed the size of the signal will raise an `OverflowError` starting in 2.0.",
```
imphil, line 740:
```suggestion
                    "Truncation of values that exceed the size of the signal will raise an `OverflowError` starting in 2.0.",
```

Since the stack trace doesn't probably contain that information it would be nice to add the value/min_val/max_value values to the warning.

cocotb/handle.py
```diff
@@ -734,19 +735,17 @@ def _set_value(self, value, call_sim):
...
-            if min_val <= value <= max_val:
-                if len(self) <= 32:
-                    call_sim(self, self._handle.set_signal_val_int, set_action, value)
-                    return
-
-                if value < 0:
-                    value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.TWOS_COMPLEMENT)
-                else:
-                    value = BinaryValue(value=value, n_bits=len(self), bigEndian=False, binaryRepresentation=BinaryRepresentation.UNSIGNED)
+            if value < min_val or max_val < value:
```
imphil, line 738:
```suggestion
            if value < min_val or value > max_val:
```

would be sightly easier to read IMO.

imphil (approved):

imphil:
Thanks a lot for going this extra mile for backwards-compat!

PR #2426, opened 2021-02-24
Deprecate assigning dict object to signals

ktbarrett:
This custom format `dict` value is not documented anywhere, and it's an operation that can be easily done by the user. The format was of the form `dict(values=[0, 1, 2], bits=4)`, and represents `values` as digits of bit size `bits`.

Based on #2240. Rebased on stable/1.5.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -730,6 +730,11 @@ class ModifiableObject(NonConstantObject):
             :class:`ctypes.Structure` objects are no longer accepted as values for assignment.
             Convert the struct object to a :class:`~cocotb.binary.BinaryValue` before assignment using
             ``BinaryValue(value=bytes(struct_obj), n_bits=len(signal))`` instead.
+
+        .. deprecated:: 1.5
+            :class:`dict` objects are no longer accepted as values for assignment.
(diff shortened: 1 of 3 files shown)


cmarqu (approved):

cocotb/handle.py
```diff
@@ -755,6 +760,11 @@ def _set_value(self, value, call_sim):
                 DeprecationWarning, stacklevel=3)
             value = BinaryValue(value=cocotb.utils.pack(value), n_bits=len(self))
         elif isinstance(value, dict):
+            warnings.warn(
+                "dict values are no longer accepted for value assignment. "
```
eric-wieser, line 764: Wording is a bit odd, since the dict is still accepted, just with a warning

imphil (approved):
LGTM, thanks!

cocotb/handle.py
ktbarrett: Following the wording from the `ctypes.Structure` example above. You have a suggestion? I'll probably change the other one as well.

imphil:
Did this one go into master already? If not, merge it into master first before merging it into the stable branch.

ktbarrett:
@imphil ping.

PR #2429, opened 2021-02-25
Add clang-format

### LINKED ISSUE #2189: Agree on and add clang-format configuration

marlonjames:
Closes #2189.

Add Google clang-format file with slight modification for smaller diff.

Apply to C++ files. I left the VPI/VHPI standard headers alone and exclude them from the CI linting.
It's quite a large diff. I looked over the output and only cleaned up the alignment on comments and
where string literals were broken across lines.

Add clang-format linting, and also exclude the examples directory.

The format moved the includes around, and so on MacOS `vpi_user.h` didn't have the typedef for `uint64_t`. I added `__APPLE__` to the `stdint.h` include for both VPI and VHPI headers.


ktbarrett (commented):
Lets start with those changes and go from there? That probably should be fine.

.clang-format
```diff
@@ -0,0 +1,169 @@
+---
```
imphil, line 1: Can you avoid dumping the whole clang-format file and use `BasedOnStyle: Google`?

Also, Ubuntu 18.04 comes with clang-format 6, so we need to support this version at least.

imphil:
On changes to the style: Everyone has their favorite thing where they disagree with any style guide. Let's not open that can. Take the Google style as-is and move on. It is well documented, well supported by tools, and reasonably consistent.

eric-wieser:
[At a glance](https://gist.github.com/andrewseidl/8066c18e97c40086c183), the clang "webkit" style seems to be closest to the style that we are already using. Simply for the sake of minimizing the diff, should we give that style a go?

imphil:
Good point Eric. I'm fine with any other well-defined style as well, and https://webkit.org/code-style-guidelines/ looks like it has nice docs as well.

eric-wieser:
But I don't feel too strongly - I think we should pick a builtin style + one or two overrides to make the diff as small as possible. At a glance, google + 4 spaces + `{` on its own line in functions would fit the bill too.

ktbarrett:
I think clang-format needs a new option `--style=emergent` 😏.

marlonjames:
@eric-wieser's suggestion of modified Google gave a slightly smaller diff than straight Webkit.

ktbarrett (approved):
Just recording my complaints for posterity.

cocotb/share/include/gpi.h
```diff
@@ -43,11 +43,12 @@ Implementation specific notes
 
 By amazing coincidence, VPI and VHPI are strikingly similar which is obviously
 reflected by this header file. Unfortunately, this means that proprietry,
-non-standard, less featured language interfaces (for example Mentor FLI) may have
-to resort to some hackery, or may not even be capable of implementing a GPI layer.
```
eric-wieser, line 47: Where did this comment go?

ktbarrett: Doesn't seem correct anymore, seeing as an FLI implementation was created.

.clang-format
```diff
@@ -0,0 +1,9 @@
...
+# Requires clang-format 10 or newer
+Language:        Cpp
+BasedOnStyle:  Google
+BraceWrapping:
+  AfterFunction:   true
```
eric-wieser, line 6: I assume we want `AfterMethod` or whatever the spelling is too? This is currently a weird middle ground

.clang-format
```diff
@@ -0,0 +1,9 @@
+---
+# Requires clang-format 10 or newer
+Language:        Cpp
```
imphil, line 3: You can remove these three lines.

Add a license header.

CONTRIBUTING.md
```diff
@@ -378,6 +378,7 @@ All changes which should go into the main codebase of cocotb must follow this se
   By contributing to this project you signal your agreement with these license terms.
 - All code must follow the established coding standards.
   For Python code, follow the [PEP 8](https://www.python.org/dev/peps/pep-0008/) style guide.
+  For C++ code, follow the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html).
```
imphil, line 381: Also mention the modifications.

cocotb/share/include/exports.h
```diff
@@ -5,11 +5,12 @@
 #ifndef COCOTB_EXPORTS_H_
 #define COCOTB_EXPORTS_H_
 
-// Make cocotb work correctly when the default visibility is changed to hidden
-// Changing the default visibility to hidden has the advantage of significantly reducing the code size,
-// load times as well as letting the optimizer produce better code
+// Make cocotb work correctly when the default visibility is changed to hidden.
+// Changing the default visibility to hidden has the advantage of significantly
+// reducing the code size and load times, as well as letting the optimizer
+// produce better code.
```
imphil, line 11: Please make *only* automated changes in this commit. Anything else that was done manually should please be in a separate commit.

cocotb/share/lib/gpi_log/gpi_logging.cpp
```diff
@@ -1,174 +1,125 @@
...
-    { 10,       "DEBUG"         },
-    { 20,       "INFO"          },
-    { 30,       "WARNING"       },
-    { 40,       "ERROR"         },
-    { 50,       "CRITICAL"      },
-    { 0,        NULL}
-};
+static struct _log_level_table log_level_table[] = {
+    {10, "DEBUG"}, {20, "INFO"},     {30, "WARNING"},
+    {40, "ERROR"}, {50, "CRITICAL"}, {0, NULL}};
```
imphil, line 97: You can get similar formatting as before by adding a trailing `,`.

.clang-format
imphil: Honestly I would just avoid overriding this one. It's making the diff a tiny bit larger, but avoiding the style exception is a clear win to me.

marlonjames: Webkit only had `AfterFunction` set to `true`. I don't really care either way.

cocotb/share/include/gpi.h
marlonjames: I removed it while cleaning up the mismatched line-splitting clang-format applied, since there is now an FLI implementation.

imphil (commented):
The last commit "Explicitly include stdint.h on MacOS" doesn't really fit into the topic of this PR and doesn't have any further description. Can you at least expand the commit message a bit to describe why this change is necessary?

cocotb/share/lib/vhpi/VhpiImpl.h
```diff
@@ -82,129 +83,131 @@ static inline int __check_vhpi_error(const char *file, const char *func, long li
...
-    __check_vhpi_error(__FILE__, __func__, __LINE__); \
-} while (0)
+#define check_vhpi_error()                                \
+    do {                                                  \
+        __check_vhpi_error(__FILE__, __func__, __LINE__); \
+    } while (0)
 
 class VhpiCbHdl : public virtual GpiCbHdl {
-public:
+   public:
```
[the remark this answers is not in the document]
marlonjames: My best guess is 1-less than methods/variables, as the default of 2 space indents looks like this:
```
class MoveOnly {
 public:
  MoveOnly(MoveOnly&& other) = default;
```

.clang-format
marlonjames: I mentioned this in the PR description. MacOS builds fail as a direct consequence of applying `clang-format`, and this is needed to make them succeed.

> The format moved the includes around, and so on MacOS `vpi_user.h` didn't have the typedef for `uint64_t`. I added `__APPLE__` to the `stdint.h` include for both VPI and VHPI headers.

cocotb/share/lib/gpi_log/gpi_logging.cpp
marlonjames: It moved the closing `};` bracket but kept the 3 per line formatting.

cocotb/share/lib/vhpi/VhpiImpl.h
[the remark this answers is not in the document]
ktbarrett: Perhaps it should be half, instead of -1 to preserve the spirit of the thing?

imphil (approved):
Thanks a lot! 

It's a good step forward for consistency and makes it easier to contribute to our code base by removing the discussion on C++ formatting completely. I know we all will find something we find uglier than before, but we'll also find things which are nicer now. Let's focus on those and move on to writing code -- a code style discussion rarely brings us closer to world domination :smile:

.clang-format
```diff
@@ -0,0 +1,5 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+BasedOnStyle:  Google
+IndentWidth:     4
```
eric-wieser, line 5:
```suggestion
IndentWidth:     4
AccessModifierOffset: -2
```
which is consistent with doubling the indent.

eric-wieser: Or

```suggestion
IndentWidth:     4
AccessModifierOffset: -4
```
if we want to match what we already have.

cocotb/share/include/gpi_logging.h
```diff
@@ -190,20 +182,17 @@ GPILOG_EXPORT void gpi_native_logger_log(
...
+GPILOG_EXPORT void gpi_native_logger_vlog(const char *name, int level,
+                                          const char *pathname,
+                                          const char *funcname, long lineno,
+                                          const char *msg, va_list args);
 
 /** Set minimum logging level of the native logger.
-    If a logging request occurs where the logging level is lower than the level set by this function, it is not logged.
-    Only affects the native logger.
+    If a logging request occurs where the logging level is lower than the level
+   set by this function, it is not logged. Only affects the native logger.
```
eric-wieser, line 195:
```suggestion
    set by this function, it is not logged. Only affects the native logger.
```

Issue #2433, opened 2021-03-01
Cannot select testcase from the module that is not first in the MODULE list


sebastiancieslak:
Hi All,

test names passed by the TESTCASE attribute are only searched in the first module on the MODULE list. If they are not there, an AttributeError is raised, even if the tests are in the next modules on the MODULE list.

**Example:**
_Makefile:_ MODULE = x_tests,y_tests
_x_tests.py:_ x_test
_y_tests.py:_ y_test

make sim -> 2 tests runned
make sim TESTCASE=x_test -> 1 test runned
make sim TESTCASE=y_test -> AttributeError: Test doesn't exist in x_tests

I've already fixed this (changes in _discover_tests in regression.py), but due to your Contribution Guidelines decided to start with creating this issue first. I will open a pull request with the MRFE test in a moment.

ktbarrett:
Yeah, there is [a comment in the source stating as much](https://github.com/cocotb/cocotb/blob/9fe131d93f31438d032307f70096fac797af78e0/cocotb/regression.py#L204), but I'm not sure exactly why this decision was made. Either way, I agree the current behavior is poor.

ktbarrett:
As a quick fix, I think the logic can be modified so it selects the first test that matches the `TESTCASE` as it loops through the `MODULE`s. Maybe you can think of something better.

Requirements:
1. All `MODULE`s specified should be loaded *exactly once*.
2. Only the first test that matches `TESTCASE` (in order of appearance in `MODULE`) should be run.

x74ea6f:
Just idea. Is it possible to use regular expressions for TESTCASE?

I tried.

regression.py _discover_tests()
```python
            if test_str:
                # Specific functions specified, don't auto-discover
                test_inst_list = []
                for d in dir(module):
                    item = getattr(module, d)
                    if isinstance(item, Test):
                        test_inst_list.append(item.__name__)

                test_list = []
                for t_str in test_str.rsplit(','):
                    for inst in test_inst_list:
                        m = re.fullmatch(t_str, inst)
                        if m:
                            test_name = m.group()
                            if not test_name in test_list:
                                test_list.append(test_name)

                for test_name in test_list:
                    test = getattr(module, test_name)
                    test.skip = False
                    yield test

                # only look in first module for all functions and don't complain if all functions are not found
                break

            # auto-discover
```

ktbarrett:
@x74ea6f This was already resolved in #2434.

PR #2434, opened 2021-03-01
Add the ability to select testcases from all modules

sebastiancieslak:
Adding MRFE test for the #2433 issue.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -168,6 +168,11 @@ class RegressionManager:
 
         modules = [s.strip() for s in module_str.split(',') if s.strip()]
 
+        tests = []
+        if test_str:
+            tests = [s.strip() for s in test_str.split(',') if s.strip()]
+            not_found_tests = []
+
         for module_name in modules:
             try:
                 _logger.debug("Python Path: " + ",".join(sys.path))
@@ -180,34 +185,41 @@ class RegressionManager:
                 _logger.info(traceback.format_exc())
                 raise
 
-            if test_str:
-
+            if tests:
                 # Specific functions specified, don't auto-discover
-                for test_name in test_str.rsplit(','):
-                    try:
-                        test = getattr(module, test_name)
-                    except AttributeError:
-                        _logger.error("Requested test %s wasn't found in module %s", test_name, module_name)
-                        err = AttributeError("Test %s doesn't exist in %s" % (test_name, module_name))
-                        raise err from None  # discard nested traceback
+                for test_name in tests:
+                    test = getattr(module, test_name, 0)
+                    if test:
+                        if not isinstance(test, Test):
+                            _logger.error("Requested %s from module %s isn't a cocotb.test decorated coroutine",
+                                          test_name, module_name)
+                            raise ImportError("Failed to find requested test %s" % test_name)
 
-                    if not isinstance(test, Test):
-                        _logger.error("Requested %s from module %s isn't a cocotb.test decorated coroutine",
-                                      test_name, module_name)
-                        raise ImportError("Failed to find requested test %s" % test_name)
+                        # If we request a test manually, it should be run even if skip=True is set.
+                        test.skip = False
 
-                    # If we request a test manually, it should be run even if skip=True is set.
-                    test.skip = False
+                        yield test
+                    else:
+                        not_found_tests.append(test_name)
 
-                    yield test
+                # Clear not_found_tests for the next module search
+                tests = not_found_tests.copy()
+                not_found_tests.clear()
 
-                # only look in first module for all functions and don't complain if all functions are not found
-                break
+                # Break if all tests were found
+                if not tests:
+                    break
+            else:
+                # auto-discover
+                for thing in vars(module).values():
+                    if isinstance(thing, Test):
+                        yield thing
 
-            # auto-discover
-            for thing in vars(module).values():
-                if isinstance(thing, Test):
-                    yield thing
+        # If any test were not found in any module, raise an error
+        if tests:
+            _logger.error("Requested test(s) %s wasn't found in module(s) %s", tests, modules)
+            err = AttributeError("Test(s) %s doesn't exist in %s" % (tests, modules))
+            raise err from None  # discard nested traceback
 
     @staticmethod
     def _discover_hooks() -> Iterable[Hook]:
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -192,16 +192,17 @@ class RegressionManager:
                         test = getattr(module, test_name)
                     except AttributeError:
                         not_found_tests.append(test_name)
-                    else:
-                        if not isinstance(test, Test):
-                            _logger.error("Requested %s from module %s isn't a cocotb.test decorated coroutine",
-                                          test_name, module_name)
-                            raise ImportError("Failed to find requested test %s" % test_name)
+                        continue
 
-                        # If we request a test manually, it should be run even if skip=True is set.
-                        test.skip = False
+                    if not isinstance(test, Test):
+                        _logger.error("Requested %s from module %s isn't a cocotb.test decorated coroutine",
+                                      test_name, module_name)
+                        raise ImportError("Failed to find requested test %s" % test_name)
 
-                        yield test
+                    # If we request a test manually, it should be run even if skip=True is set.
+                    test.skip = False
+
+                    yield test
 
                 # Clear not_found_tests for the next module search
                 tests = not_found_tests.copy()
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -171,7 +171,6 @@ class RegressionManager:
         tests = None
         if test_str:
             tests = [s.strip() for s in test_str.split(',') if s.strip()]
-            not_found_tests = []
 
         for module_name in modules:
             try:
@@ -186,6 +185,7 @@ class RegressionManager:
                 raise
 
             if tests is not None:
+                not_found_tests = []
                 # Specific functions specified, don't auto-discover
                 for test_name in tests:
                     try:
@@ -206,7 +206,6 @@ class RegressionManager:
 
                 # Clear not_found_tests for the next module search
                 tests = not_found_tests.copy()
-                not_found_tests.clear()
 
             else:
                 # auto-discover
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -168,7 +168,7 @@ class RegressionManager:
 
         modules = [s.strip() for s in module_str.split(',') if s.strip()]
 
-        tests = []
+        tests = None
         if test_str:
             tests = [s.strip() for s in test_str.split(',') if s.strip()]
             not_found_tests = []
@@ -185,7 +185,7 @@ class RegressionManager:
                 _logger.info(traceback.format_exc())
                 raise
 
-            if tests:
+            if tests is not None:
                 # Specific functions specified, don't auto-discover
                 for test_name in tests:
                     try:
@@ -208,9 +208,6 @@ class RegressionManager:
                 tests = not_found_tests.copy()
                 not_found_tests.clear()
 
-                # Break if all tests were found
-                if not tests:
-                    break
             else:
                 # auto-discover
                 for thing in vars(module).values():
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -188,8 +188,11 @@ class RegressionManager:
             if tests:
                 # Specific functions specified, don't auto-discover
                 for test_name in tests:
-                    test = getattr(module, test_name, 0)
-                    if test:
+                    try:
+                        test = getattr(module, test_name)
+                    except AttributeError:
+                        not_found_tests.append(test_name)
+                    else:
                         if not isinstance(test, Test):
                             _logger.error("Requested %s from module %s isn't a cocotb.test decorated coroutine",
                                           test_name, module_name)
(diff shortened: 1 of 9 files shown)


ktbarrett (commented):
The logic could be simplified if you used a `set` rather than a `list`, it would also prevent an issue I see in that if you have the same name listed multiple times in `TESTCASE`, it will have to pick up a second test of that same name. Which might be surprising for a user who accidentally or due to automation has a duplicate and is confounded by a `ERROR: Can't find 'test_that_really_exists' ...` message.

Could you also add another module to the test with the name `y_test` that fails, so we can test to ensure it is only picking up the first test of that name?

ktbarrett:
I take the first section back. I didn't realize that it was an existing behavior. A newsfragment documenting the change in behavior is also needed.

cocotb/regression.py
```diff
@@ -180,34 +185,41 @@ def _discover_tests() -> Iterable[Test]:
...
-                                      test_name, module_name)
-                        raise ImportError("Failed to find requested test %s" % test_name)
-
-                    # If we request a test manually, it should be run even if skip=True is set.
-                    test.skip = False
-
-                    yield test
+                for test_name in tests:
+                    test = getattr(module, test_name, 0)
+                    if test:
```
eric-wieser, line 192:
```suggestion
                    try:
                        test = getattr(module, test_name)
                    except AttributeError:
                        not_found_tests.append(test_name)
                    else:
```

cocotb/regression.py
```diff
@@ -180,34 +185,41 @@ def _discover_tests() -> Iterable[Test]:
...
+                            _logger.error("Requested %s from module %s isn't a cocotb.test decorated coroutine",
+                                          test_name, module_name)
+                            raise ImportError("Failed to find requested test %s" % test_name)
+
+                        # If we request a test manually, it should be run even if skip=True is set.
+                        test.skip = False
+
+                        yield test
+                    else:
+                        not_found_tests.append(test_name)
```
eric-wieser, line 203: paired with the suggestion above:
```suggestion
```

cocotb/regression.py
```diff
@@ -180,34 +185,41 @@ def _discover_tests() -> Iterable[Test]:
...
+                # auto-discover
+                for thing in vars(module).values():
+                    if isinstance(thing, Test):
+                        yield thing
+
+        # If any test were not found in any module, raise an error
+        if tests:
+            _logger.error("Requested test(s) %s wasn't found in module(s) %s", tests, modules)
+            err = AttributeError("Test(s) %s doesn't exist in %s" % (tests, modules))
+            raise err from None  # discard nested traceback
```
eric-wieser, line 222: You're not in an `except` block so there's nothing to discard any more:
```suggestion
            raise AttributeError("Test(s) %s doesn't exist in %s" % (tests, modules))
```

sebastiancieslak:
@ktbarrett 
Passing the same test name multiple times in `TESTCASE` will result in the same way as before - test will be run multiple times.

I've extended the test so it now checks if only `y_test` from the module that is earlier in the `MODULE` list is used. It also checks if using the same test name twice works.

A newsfragment added. I'm not sure if a `bugfix` tag fits this change the best, but lack of ability to run a test if all Makefile variables are set correctly looks like a bug.

cocotb/regression.py
```diff
@@ -180,34 +185,41 @@ def _discover_tests() -> Iterable[Test]:
...
-                    if not isinstance(test, Test):
-                        _logger.error("Requested %s from module %s isn't a cocotb.test decorated coroutine",
-                                      test_name, module_name)
-                        raise ImportError("Failed to find requested test %s" % test_name)
+                        not_found_tests.append(test_name)
+                    else:
+                        if not isinstance(test, Test):
+                            _logger.error("Requested %s from module %s isn't a cocotb.test decorated coroutine",
+                                          test_name, module_name)
+                            raise ImportError("Failed to find requested test %s" % test_name)
```
eric-wieser, line 199: Perhaps to minimize the diff:
```suggestion
                        continue

                    if not isinstance(test, Test):
                        _logger.error("Requested %s from module %s isn't a cocotb.test decorated coroutine",
                                      test_name, module_name)
                        raise ImportError("Failed to find requested test %s" % test_name)
```

cocotb/regression.py
```diff
@@ -180,34 +185,41 @@ def _discover_tests() -> Iterable[Test]:
...
 
-                    # If we request a test manually, it should be run even if skip=True is set.
-                    test.skip = False
+                        # If we request a test manually, it should be run even if skip=True is set.
+                        test.skip = False
```
eric-wieser, line 202:
```suggestion
                    # If we request a test manually, it should be run even if skip=True is set.
                    test.skip = False
```

cocotb/regression.py
```diff
@@ -180,34 +185,41 @@ def _discover_tests() -> Iterable[Test]:
...
 
-                    yield test
+                        yield test
```
eric-wieser, line 204:
```suggestion
                    yield test
```

cocotb/regression.py
```diff
@@ -168,6 +168,11 @@ def _discover_tests() -> Iterable[Test]:
 
         modules = [s.strip() for s in module_str.split(',') if s.strip()]
 
+        tests = []
```
eric-wieser, line 171:
```suggestion
        tests = None
```

cocotb/regression.py
```diff
@@ -180,34 +185,41 @@ def _discover_tests() -> Iterable[Test]:
                 _logger.info(traceback.format_exc())
                 raise
 
-            if test_str:
-
+            if tests:
```
eric-wieser, line 188:
```suggestion
            if tests is not None:
```

cocotb/regression.py
```diff
@@ -180,34 +185,41 @@ def _discover_tests() -> Iterable[Test]:
...
+                tests = not_found_tests.copy()
+                not_found_tests.clear()
 
-            # auto-discover
-            for thing in vars(module).values():
-                if isinstance(thing, Test):
-                    yield thing
+                # Break if all tests were found
+                if not tests:
+                    break
```
eric-wieser, line 212:
```suggestion
```
not needed with my change to `None` above

sebastiancieslak: I'm not sure if it is a good idea. After removing this `break`, we will always import all listed in the `MODULE` modules, even if all tests were found in the first one. So in cases where user has many modules and often runs some basic test that is in the first module, this can have a negative impact on performance.

ktbarrett: That is actually desired. Module-level code can have side effects. So if they are listed we should intend to load them, at a minimum.

sebastiancieslak: Ok, thanks for the explanation, I'll add this change.

eric-wieser: Another argument in favor of my suggestion is that it won't silently ignore missing modules just because it found all the tests already.

cocotb/regression.py
sebastiancieslak: Changed in 7b89f41f006ed5c7da2b03838072de5474024bbf.

cocotb/regression.py
sebastiancieslak: Changed in 7b89f41f006ed5c7da2b03838072de5474024bbf.

cocotb/regression.py
sebastiancieslak: Changed in 7b89f41f006ed5c7da2b03838072de5474024bbf.

cocotb/regression.py
sebastiancieslak: Changed in cb4ed6cd7a8e05c43ceda4fb305d6a20dbf4be1e.

cocotb/regression.py
sebastiancieslak: Changed in cb4ed6cd7a8e05c43ceda4fb305d6a20dbf4be1e.

cocotb/regression.py
sebastiancieslak: Changed in cb4ed6cd7a8e05c43ceda4fb305d6a20dbf4be1e.

cocotb/regression.py
sebastiancieslak: Changed in ccc0ce20dd809c137e93ae20666fd06190fa0f4e.

cocotb/regression.py
sebastiancieslak: Changed in ccc0ce20dd809c137e93ae20666fd06190fa0f4e.

cocotb/regression.py
sebastiancieslak: Changed in ccc0ce20dd809c137e93ae20666fd06190fa0f4e.

cocotb/regression.py
```diff
@@ -201,13 +204,20 @@ def _discover_tests() -> Iterable[Test]:
...
+            else:
+                # auto-discover
+                for thing in vars(module).values():
+                    if isinstance(thing, Test):
+                        yield thing
+
+        # If any test were not found in any module, raise an error
+        if tests:
+            _logger.error("Requested test(s) %s wasn't found in module(s) %s", tests, modules)
+            raise AttributeError("Test(s) %s doesn't exist in %s" % (tests, modules))
```
ktbarrett, line 220: If you have the time could you make another test that covers this line? You can look at [test_failure's Makefile](https://github.com/cocotb/cocotb/blob/master/tests/test_cases/test_failure/Makefile) for idea on how to let the test fail, but the regression pass.

If you don't have time that's okay, just let me know.

documentation/source/newsfragments/2434.bugfix.rst
```diff
@@ -0,0 +1 @@
+Selecting tests from a module that is not the first on the :make:var:`MODULE` list will now work correctly.
```
ktbarrett, line 1: I would think this would be a "change". And the text should read
```suggestion
Selecting tests using :make:var:`TESTCASE` will now search for the first occurrence of a test of that name in order of :make:var:`MODULE`s, and not just the first module.
```

cocotb/regression.py
sebastiancieslak: Yeah, I'll do it tomorrow, if that's not a problem.

documentation/source/newsfragments/2434.bugfix.rst
sebastiancieslak: Changed in 0e7e5da8b21482cffeca7f66299555edf20a1c00.

cocotb/regression.py
sebastiancieslak: Added in cf648d4874d0f8c846a7bf27d549ae06f6b3fb89.

cocotb/regression.py
```diff
@@ -180,16 +185,14 @@ def _discover_tests() -> Iterable[Test]:
                 _logger.info(traceback.format_exc())
                 raise
 
-            if test_str:
-
+            if tests is not None:
```
eric-wieser, line 188:
```suggestion
            if tests is not None:
                not_found_tests = []
```

cocotb/regression.py
```diff
@@ -168,6 +168,11 @@ def _discover_tests() -> Iterable[Test]:
 
         modules = [s.strip() for s in module_str.split(',') if s.strip()]
 
+        tests = None
+        if test_str:
+            tests = [s.strip() for s in test_str.split(',') if s.strip()]
+            not_found_tests = []
```
eric-wieser, line 174:
```suggestion
```

cocotb/regression.py
```diff
@@ -201,13 +204,20 @@ def _discover_tests() -> Iterable[Test]:
 
                     yield test
 
-                # only look in first module for all functions and don't complain if all functions are not found
-                break
+                # Clear not_found_tests for the next module search
+                tests = not_found_tests.copy()
+                not_found_tests.clear()
```
eric-wieser, line 209:
```suggestion
                # use not_found_tests for the next module search
                tests = not_found_tests
```

eric-wieser (commented):
The logic is a little clearer if you push `not_found_tests` into the loop

cocotb/regression.py
sebastiancieslak: Changed in 0d656990c6bc950012502e9d25ff767c6ea59468.

cocotb/regression.py
sebastiancieslak: Changed in 0d656990c6bc950012502e9d25ff767c6ea59468.

cocotb/regression.py
sebastiancieslak: Changed in 0d656990c6bc950012502e9d25ff767c6ea59468.

cocotb/regression.py
```diff
@@ -201,13 +204,19 @@ def _discover_tests() -> Iterable[Test]:
 
                     yield test
 
-                # only look in first module for all functions and don't complain if all functions are not found
-                break
+                # Clear not_found_tests for the next module search
+                tests = not_found_tests.copy()
```
eric-wieser, line 208:
```suggestion
                # Use the non-matching test names in the next module search
                tests = not_found_tests
```

ktbarrett (approved):

cocotb/regression.py
sebastiancieslak: Right, I forgot to delete it.

sebastiancieslak: Changed in 368e81fd804297276d9e9ac6a7138b256d7f1b63.

PR #2436, opened 2021-03-02
Use bright blue for info logs

### LINKED ISSUE #2435: Can ANSI avoid using BLUE_FG? Save my eyes please.
The default color:
Screenshot_2021-03-02_10-29-18

After I changed the BLUE_xx in COLOR_INFO and COLOR_TEST in ANSI.py to BRIGHT_BLUE_xx and the result is much better:
Screenshot_2021-03-02_10-29-41

By the way, thanks to the contributors of cocotb.

ktbarrett:
Closes #2435.


marlonjames:
If we're going to change Info we should change Warning through Critical as well. They shouldn't have less emphasis in the terminal.

eric-wieser:
Using "bright" seems semantically wrong here, I agree with @marlonjames. The original issue is really just a terminal config issue, but on the other hand lots of terminals are misconfigured this way by default.

Should info messages just be regular white? Then we could mute the stack information in the left columns by rendering it as "bright black"

imphil:
I agree that this misconfiguration is very common, so let's find a sensible default for cocotb that works for most of our users.

Choosing another blue might be one option, but I'd go for just making it white as well, there is no semantic need for a color here.

ktbarrett:
Actually, I like the idea of "default FG" for info and debug statements.

marlonjames:
If you change:
https://github.com/cocotb/cocotb/blob/9fe131d93f31438d032307f70096fac797af78e0/cocotb/log.py#L239
Instead of `ANSI.COLOR_INFO`, It would remove lots of unnecessary ANSI codes from the log output.

ktbarrett:
Yeah it would, but I left it there so users could change it if desired.

eric-wieser:
> Yeah it would, but I left it there so users could change it if desired.

Users could still change it with `SimColourLogFormatter.loglevel2colour` if they felt so inclined.

... which uses a different spelling of `color` to the constants we're talking about.

ktbarrett:
So no ANSI codes or to ANSI codes? If there is another way to do it I'm fine either way, not having ANSI codes in logs in nice.

marlonjames:
I would change `loglevel2colour` so that INFO matches DEBUG, simply `"%s"` with no ANSI codes.

marlonjames (approved):

PR #2438, opened 2021-03-02
Add example for an analog circuit model

cmarqu:

### DIFF
--- a/documentation/requirements.txt
+++ b/documentation/requirements.txt
@@ -13,3 +13,4 @@ sphinx-argparse
 towncrier
 IPython
 sphinx-tabs
+sphinxcontrib-details-directive
--- a/documentation/source/analog_model.rst
+++ b/documentation/source/analog_model.rst
@@ -1,0 +1,70 @@
+#########################
+Models of Analog Circuits
+#########################
+
+.. versionadded:: 2.0
+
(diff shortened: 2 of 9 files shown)


ktbarrett (commented):
Is this not ready for review? It looks really good as it stands.

I think the one thing that could change to make it more clear you are mocking HDL with a Python model is to create an "empty" HDL module and instantiate and hook it up like you would if it were implemented. Then have the Python model monitor and drive the empty module's internal signals. To me, that would make it clearer what the intent of the Python model was.

cmarqu:
Unless I'm misunderstanding you, there is no such "empty" HDL module there really, because no other HDL will come into existence later. What *will* come into existence is the real implementation of this analog front-end which is done with an analog design framework using schematics that down at the very lowest level use transistors - but that is inaccessible to cocotb, accessible only to the analog kernel of a mixed-signal simulator.

ktbarrett (approved):
Okay I see what it's doing. It's modelling the AFE which is an "external device". That makes sense now.

cmarqu:
I guess I should describe this better in the example introduction. This AFE would be an external circuit for an FPGA but usually (not necessarily) on the same chip for an ASIC.

ktbarrett (approved):

Issue #2327, opened 2021-01-05
Define and document what edge triggers do with "unusual" signals and values


cmarqu:
Currently, we don't define or document what e.g. the following code does or should do:

* `RisingEdge(multi_bit_vector)` changing from `0b0001` to `0b1110`
* `RisingEdge(int)` changing from `0x1` to `0xE`
* `RisingEdge(bit)` changing from `X` to `1`
* `RisingEdge(std_logic)` changing from `H` to `1`
* `Edge(string)` changing from `"a"` to `"b"`
* `RisingEdge(string)` (maybe changing from `""` to `"a"`, and `"a"` to `""`)

Some are fairly clear, others not so much.

ktbarrett:
`RisingEdge` or `FallingEdge` on anything but logic signals or arrays of logics is nonsense. `Edge` should fire on any signal when the value changes.

PR #2439, opened 2021-03-02
Add test for Edge(<vector>) with 0/1 bits

cmarqu:
Questa needs an extra `ReadOnly()` after `Edge()`.
Checking behavior for metavalues is for another PR.

See also #2327.

### DIFF
--- a/tests/test_cases/test_cocotb/test_edge_triggers.py
+++ b/tests/test_cases/test_cocotb/test_edge_triggers.py
@@ -10,7 +10,7 @@ Tests for edge triggers
 * ClockCycles
 """
 import cocotb
-from cocotb.triggers import RisingEdge, FallingEdge, Edge, Timer, ClockCycles, First, Combine
+from cocotb.triggers import RisingEdge, FallingEdge, Edge, Timer, ClockCycles, First, Combine, ReadOnly
 from cocotb.clock import Clock
 from cocotb.result import SimTimeoutError
 
@@ -233,3 +233,35 @@ async def test_both_edge_triggers(dut):
     falling_coro = cocotb.fork(wait_falling_edge())
     cocotb.fork(Clock(dut.clk, 10, units='ns').start())
     await Combine(rising_coro, falling_coro)
+
+
+@cocotb.test()
+async def test_edge_on_vector(dut):
+    """Test that Edge() triggers on any 0/1 change in a vector"""
+
+    cocotb.fork(Clock(dut.clk, 100, "ns").start())
+
+    edge_cnt = 0
+
+    async def wait_edge():
+        nonlocal edge_cnt
+        while True:
+            await Edge(dut.stream_out_data_registered)
+            if cocotb.SIM_NAME.lower().startswith("modelsim"):
+                await ReadOnly()  # not needed for other simulators
+            edge_cnt = edge_cnt + 1
+
+    cocotb.fork(wait_edge())
+
+    dut.stream_in_data <= 0
+    await RisingEdge(dut.clk)
+
+    for val in range(1, 2**len(dut.stream_in_data)-1):
+        # produce an edge by setting a value != 0:
+        dut.stream_in_data <= val
+        await RisingEdge(dut.clk)
+        # set back to all-0:
+        dut.stream_in_data <= 0
+        await RisingEdge(dut.clk)
+
+    assert edge_cnt == 2 * ((2**len(dut.stream_in_data)-1)-1)


ktbarrett:
I'll test on Questa tomorrow, then approve.

cmarqu:
> I assume you already tested Cadence simulators?

Yes, I tested Icarus (VPI), Xcelium (VPI. VHPI), Questa (VPI, FLI), VCS (VPI).
In addition, ghdl in CI also passes.
The result for Questa with VHPI would be interesting.

ktbarrett:
Questa with VHPI doesn't work at all right. They haven't implemented what we use for `ReadWrite`, so regular cocotb writes never happen.

tests/test_cases/test_cocotb/test_edge_triggers.py
```diff
@@ -233,3 +233,39 @@ async def wait_falling_edge():
...
+    cocotb.fork(wait_edge())
+
+    dut.stream_in_data <= 0
+    await ClockCycles(dut.clk, 1)
+
+    for val in range(1, 2**len(dut.stream_in_data)-1):
+        # produce an edge by setting a value != 0:
+        dut.stream_in_data <= val
+        await ClockCycles(dut.clk, 1)
+        edge_cnt_exp = edge_cnt_exp + 1
```
ktbarrett, line 262: Is there some reason you can't precompute what the expected count will be? I figured it would be `2 * (2**len(dut.stream_in_data))`.

tests/test_cases/test_cocotb/test_edge_triggers.py
```diff
@@ -233,3 +233,39 @@ async def wait_falling_edge():
...
+
+    cocotb.fork(wait_edge())
+
+    dut.stream_in_data <= 0
+    await ClockCycles(dut.clk, 1)
+
+    for val in range(1, 2**len(dut.stream_in_data)-1):
+        # produce an edge by setting a value != 0:
+        dut.stream_in_data <= val
+        await ClockCycles(dut.clk, 1)
```
ktbarrett, line 261: Is there a reason to use `ClockCycles(clk, 1)` over `RisingEdge(clk)`?

tests/test_cases/test_cocotb/test_edge_triggers.py
```diff
@@ -233,3 +233,39 @@ async def wait_falling_edge():
...
+        edge_cnt_exp = edge_cnt_exp + 1
+        # set back to all-0:
+        dut.stream_in_data <= 0
+        await ClockCycles(dut.clk, 1)
+        edge_cnt_exp = edge_cnt_exp + 1
+
+    if cocotb.SIM_NAME.lower().startswith(("modelsim")):
+        assert edge_cnt == 1
```
ktbarrett, line 269: Big yikes. Why is this? Maybe a comment would be good. I think it would be better to mark this as an expected fail rather than baking it into the test. It will be easier to locate issues with simulators in the future.

tests/test_cases/test_cocotb/test_edge_triggers.py
cmarqu: This is single-step driving the clock, not reacting to it.
A small side-effect is that I can test Xcelium which otherwise would fail because of #1076.

tests/test_cases/test_cocotb/test_edge_triggers.py
cmarqu: Why... no idea. At least it's consistent between VPI and FLI.

tests/test_cases/test_cocotb/test_edge_triggers.py
cmarqu: (Previous comment was BS...)
No, no reason, will change.

tests/test_cases/test_cocotb/test_edge_triggers.py
cmarqu: It's `2 * ((2**len(dut.stream_in_data)-1)-1)` (508) which takes a while to reason about.

ktbarrett: It's more about verifying the code is meets our expectations. If there is poor simulator behavior in both loops, the test may pass when it shouldn't (probably unlikely though).

tests/test_cases/test_cocotb/test_edge_triggers.py
ktbarrett: I think this is worth opening an issue about. Is it hanging on the first edge seen? What happens if you put a `ReadOnly` after every watch?

cmarqu: Cool, the extra `ReadOnly` fixed it.

tests/test_cases/test_cocotb/test_edge_triggers.py
cmarqu: Changed.

tests/test_cases/test_cocotb/test_edge_triggers.py
ktbarrett: So `Edge(signal)` immediately after `Edge(signal)` doesn't work on Modelsim.

ktbarrett: That makes no sense since there is a `RisingEdge` after `RisingEdge` in the main process...

ktbarrett (approved):

PR #2441, opened 2021-03-03
Revert "Use setuptools_scm to automatically generate version number"

imphil:
This reverts commit b7a756e76c95d36525759beee6e3022f4f6e6b43.

setuptools_scm doesn't work with releases done off a branch, see 
https://github.com/pypa/setuptools_scm/issues/523 for details.

Beyond that, setuptools_scm and the generation of release notes with
towncrier also poses interesting ordering challenge, which are solvable if
one takes care, however.

Not using setuptools_scm removes the exact git hash from installed 
development versions, which makes it harder for us and users of our 
development releases to figure out the version they're running. We have 
been OK with that for a fair while until we started using setuptools_scm,
so it's not the end of the world. (We can always look into including this
information back into the version without setuptools_scm.)

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -158,8 +158,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-      with:
-        fetch-depth: 0  # ensure we clone all the tags so we can see setuptools_scm working
 
       # Install Python
     - name: Set up Python ${{matrix.python-version}}
--- a/.gitignore
+++ b/.gitignore
@@ -105,6 +105,3 @@ examples/*/tests/compile
 # Tachyon DA CVC
 tests/test_cases/*/verilog.log
 examples/*/tests/verilog.log
-
-# setuptools_scm artifact
-/cocotb/_version.py
--- a/cocotb/_version.py
+++ b/cocotb/_version.py
@@ -5,4 +5,4 @@
 # 1) we don't load dependencies by storing it in __init__.py
 # 2) we can import it in setup.py for the same reason
 # 3) we can import it into your module
-__version__ = '1.5.0.dev0'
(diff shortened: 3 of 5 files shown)


imphil:
Two things:

- Before we can start the 2.0 cycle, we need to have a cocotb-bus version that's compatible. 
- Even though I like setuptools_scm and use it in other projects successfully, I always found it slightly fragile, and we just found another case that fully breaks it. I now trend towards less magic and just increment the version number manually. If we want the git hash in the output of `cocotb-config --version` (and I think that is a good thing form a user support perspective) we can add it back manually.

ktbarrett:
> Not using setuptools_scm removes the exact git hash from installed
> development versions, which makes it harder for us and users of our
> development releases to figure out the version they're running. We have
> been OK with that for a fair while until we started using setuptools_scm,
> so it's not the end of the world. (We can always look into including this
> information back into the version without setuptools_scm.)

It also makes user's lives harder. Many didn't realize that master.zip would no longer work and that git was necessary as a build prerequisite. It was also causing issues with using the Msys2 environment as well. It's sort of useful, but also a burden. One thing to look at, setuptools_scm automatically included all version controlled files in sdists, so you might need to look at that.

> If we want the git hash in the output of `cocotb-config --version` (and I think that is a good thing form a user support perspective) we can add it back manually.

I'm fine what that.

> * Before we can start the 2.0 cycle, we need to have a cocotb-bus version that's compatible.

I can't say that cocotb-bus *will* be compatible with 2.0, so putting out a release of cocotb-bus that says it will be is asking for trouble or shooting ourselves in the foot in regards to making breaking changes. We should move all cocotb-bus related stuff out of the repo and into cocotb-bus. It might be better for right now that the version on master stay as 1.6.0dev?

ktbarrett (approved):

PR #2443, opened 2021-03-05
Add Queue empty/full waits

marlonjames:
Adds `Queue.wait_empty()` and `Queue.wait_full()` coroutines.

I tried f-strings but felt the code was easier to read with `''.format()` style.

### DIFF
--- a/tests/test_cases/test_cocotb/test_queues.py
+++ b/tests/test_cases/test_cocotb/test_queues.py
@@ -7,7 +7,7 @@ Tests relating to cocotb.queue.Queue, cocotb.queue.LifoQueue, cocotb.queue.Prior
 import cocotb
 from cocotb.queue import Queue, PriorityQueue, LifoQueue, QueueFull, QueueEmpty
 from cocotb.regression import TestFactory
-from cocotb.triggers import Combine
+from cocotb.triggers import Combine, NullTrigger
 import pytest
 
 
@@ -256,3 +256,99 @@ async def test_str_and_repr(_):
     s = repr(q)
     assert "_getters" not in s
     assert str(q)[:-1] in s
+
+
+@cocotb.test()
+async def test_wait_full(_):
+    QUEUE_SIZE = 5
+
+    q = Queue()
+
+    with pytest.raises(RuntimeError):
+        await q.wait_full()
+
+    q = Queue(maxsize=QUEUE_SIZE)
+
+    async def full_coro():
+        await q.wait_full()
+
+    fwaiter = cocotb.fork(full_coro())
+    assert not fwaiter._finished
+
+    for i in range(QUEUE_SIZE - 1):
+        q.put_nowait(i)
+
+    assert not q.full()
+    assert not fwaiter._finished
+
+    q.put_nowait(QUEUE_SIZE - 1)
+
+    assert q.full()
+    assert not fwaiter._finished
+
+    await NullTrigger()
+
+    assert q.full()
+    assert fwaiter._finished
+
+    # wait on already full queue
+    fwaiter2 = cocotb.fork(full_coro())
+
+    await NullTrigger()
+
+    assert fwaiter2._finished
+
+    q.get_nowait()
+
+    assert not q.full()
+
+    fwaiter = cocotb.fork(full_coro())
+    assert not fwaiter._finished
+
+
+@cocotb.test()
+async def test_wait_empty(_):
+    QUEUE_SIZE = 5
+    q = Queue(maxsize=QUEUE_SIZE)
+
+    # fill queue
+    for i in range(QUEUE_SIZE):
+        q.put_nowait(i)
+
+    async def empty_coro():
+        await q.wait_empty()
+
+    ewaiter = cocotb.fork(empty_coro())
+    assert not q.empty()
+    assert not ewaiter._finished
+
+    for _ in range(QUEUE_SIZE - 1):
+        q.get_nowait()
+
+    assert not q.empty()
+    assert not ewaiter._finished
+
+    q.get_nowait()
+
+    assert q.empty()
+    assert not ewaiter._finished
+
+    await NullTrigger()
+
+    assert q.empty()
+    assert ewaiter._finished
+
+    # wait on already empty queue
+    ewaiter2 = cocotb.fork(empty_coro())
+
+    await NullTrigger()
+
+    assert ewaiter2._finished
+
+    q.put_nowait(0)
+
+    assert not q.empty()
+
+    ewaiter = cocotb.fork(empty_coro())
+    assert not ewaiter._finished
+    ewaiter.kill()
--- a/cocotb/queue.py
+++ b/cocotb/queue.py
@@ -24,8 +24,9 @@ class Queue(Generic[T]):
 
         self._maxsize = maxsize
 
-        self._finished = Event()
-        self._finished.set()
+        self._full_event = Event('{} full'.format(type(self).__name__))
+        self._empty_event = Event('{} empty'.format(type(self).__name__))
+        self._empty_event.set()
 
         self._getters = collections.deque()
         self._putters = collections.deque()
@@ -60,7 +61,7 @@ class Queue(Generic[T]):
     def _format(self):
         result = 'maxsize={}'.format(repr(self._maxsize))
         if getattr(self, '_queue', None):
-            result += ' _queue={}'.format(repr(list(self._queue)))
+            result += ' _queue={!r}'.format(list(self._queue))
         if self._getters:
             result += ' _getters[{}]'.format(len(self._getters))
         if self._putters:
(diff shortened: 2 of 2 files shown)


tests/test_cases/test_cocotb/test_queues.py
```diff
@@ -256,3 +256,88 @@ async def test_str_and_repr(_):
     s = repr(q)
     assert "_getters" not in s
     assert str(q)[:-1] in s
+
+
+@cocotb.test()
+async def test_waits(_):
```
ktbarrett, line 262: It's kind of a long tests. Maybe you can split it in 2: 1 for full, 1 for empty?

ktbarrett (approved):
Can't say I have any reservations about the implementation.

ktbarrett:
I think this deserved to be in 1.5 since we are still in rc process. We have to do at least one more anyways. Could you cherry-pick this onto the 1.5 branch?

ktbarrett:
Actually, I forgot about a newsfragment for 1.6. I don't think one is necessary for 1.5 since we introduced it in the same version.

ktbarrett (approved):

eric-wieser:
I can't help feeling this is what the `task_done` and `join` API that was part of the asyncio queue (which @ktbarrett decided to gut) was for.

marlonjames:
@eric-wieser while it seems related, to me the distinction is that `wait_empty`/`wait_full` are about the status of the Queue, whereas `task_done`/`join` were about the status of handling the objects that come out of the Queue.

eric-wieser:
`task_done` / `join` is strictly more useful than `wait_empty`, as you can build the former out of the latter by calling `task_done` immediately after `get` in the consumer.

I'm not really sure I understand the use case of `wait_full` yet.

ktbarrett:
@eric-wieser While `wait_empty` could be implemented in terms of `task_done`/`join` it doesn't have a good use outside of exactly that purpose, and it's more error prone on the user. I don't know why anyone would use it over `threading.Pool` or `asyncio.gather`, for example. If implementing something like `threading.Pool` or `asyncio.gather` is useful in the context of cocotb (it's not AFAICT), we can build that abstraction without having to maintain the whole `task_done`/`join` hack.

`wait_full` was simply for completeness, since `empty` and `full` exist, both waits should exist. I haven't had a need for `full` or `wait_full` in any of my code though.

ktbarrett (commented):
I do see one issue, `full` and `empty` don't respect `_putters` and `_getters`, so we get false positives about fullness if there are outstanding getters or emptiness if there are oustanding putters. Seems like just because you copy it from the cpython codebase it isn't high quality or correct =/ Maybe there is a good reason for the current behavior.

eric-wieser:
> it doesn't have a good use outside of exactly that purpose

I disagree - it's very common to want to wait for a queue of _tasks_ to be done, and not just a queue of _requests_ to have been sent. I can't actually think of any compelling use case for an `empty` event.

Can you come up with a simple-but-not-too-contrived way you would want to use this trigger?

eric-wieser:
> Seems like just because you copy it from the cpython codebase it isn't high quality or correct =/ Maybe there is a good reason for the current behavior.

I'd encourage you to post on bugs.python.org if you think there's a problem with the CPython implementation.

marlonjames:
> I do see one issue, `full` and `empty` don't respect `_putters` and `_getters`, so we get false positives about fullness if there are outstanding getters or emptiness if there are oustanding putters.

Only with a `maxsize` of 1, right?

ktbarrett:
@eric-wieser 
> wait for a queue of _tasks_ to be done

That's exactly what I mentioned. There are simpler and less error prone ways to implement that. *If* it is useful, we will spring to implement one of those schemes

 `empty` is useful in the case where a user wants to judge if an input queue is drained, or a scoreboard's queue is settled. It's more about checking if there are there any outstanding *values*. You and the cpython authors of the mid 00's seem to be conflating two different use cases that should be handled separately. They realized that, which is why there are other utilities for implementing task pools/queues now. With asyncio, we have Futures which beat the crap out of some hacked together task queue methods, which weren't implemented for threads.

`full` and `empty` are correct for their use internally, but it's *probably* not what users want. And at this point backwards compatibility dictates it will never change.

ktbarrett:
@marlonjames Imagine a condition where a queue is filled up by multiple writers, they give up control and a consumer drains the queue until it is "empty". It can't do anymore and must yield control, but there are still values in the queue, yet to be written. So an outside observer waits for empty and it returns then, but there are still values "in the queue". It will still work, but only in limited circumstances. Hmmmm...

marlonjames:
As it is now, `full()` and `empty()` have the same problem. They're only really useful with the `*_nowait()` methods, since there can be a "wait queue to access the queue" when using the async functions.

eric-wieser:
> where a user wants to judge if an input queue is drained

I struggle to see when this is useful - what is the consumer has just moved it to another queue (eg to an axis transaction that is still in progress)?

ktbarrett:
Presumably some kind of busy event on the AXI driver so you know you are finished with inputs.

eric-wieser:
That busy event is independent of the queue empty event thought right? Since first the queue is emptied, then some signals are waggled, and only _then_ is the driver idle. So why have the empty event at all?

ktbarrett:
A Driver can only know it is not currently trying to apply a transaction or not, an idle event would trigger often, unless it had knowledge of whether the queue that feeds it is empty or not. And we are back to the original argument. Reduction arguments are usually a waste of time. After second thought, outside observance of a queue with the current implementation was not considered and isn't sound. 1-1, 1-N, N-1, and N-N data communication works and is sound, so I'm fine to leave it at that.

@marlonjames Do you have any arguments one way or the other seeing as you did the work?

marlonjames:
Closing this, as there doesn't seem to be a strong use case for it.
It seems to me that `full()` and `empty()` are useful for the `*_nowait()` functions, but not when using the async `put()` and `get()`.

ktbarrett:
Thanks for trying though.

jwprice100:
As a user, I would think a wait empty would be wonderful. I think a wait for full is potentially useful as well. 

A couple of use cases:
1. I might slam an input transaction into a bus _and_ an input queue. Sometime later I may get an output transaction. I then pull from the input queue, predict the output transaction, then compare to the actual.  Being able to then know when all my input queues are empty allows me to know when to terminate simulation. 
2. I often have a fair amount of data and don't want to buffer it up all at once, so it's generated on the fly. A couple of times it would have been nice to know an input queue was drained so I can check to replenish it. You could imagine a producer that fill sand drains a queue. Obviously the consumer can signal the producer, but having it in the queue itself (the object they already use to communicate) is very nice. 
3. Modeling hardware - I've had designs that wait until a FIFO hits a threshold to trigger behavior. This suggests a wait_full might be best replaced with a wait for n items in the queue (maybe default to maxsize if applicable). 

There are many ways to get what I need by working around the Queue class. But that isn't natural and seems like a lot of reinventing the wheel.

alexforencich:
Couple of points:
1. Now that the task tracking stuff has been removed, @eric-wieser 's argument that the task tracking stuff satisfies this use case is invalid.
2. One of the main arguments against something like this was that it would deviate from the asyncio queue API.  However, now that the task tracking stuff has been removed, we're already not consistent with the asyncio queue API
3. If we add wait_full and wait_empty, we also should add wait_not_full and wait_not_empty.

eric-wieser:
> 1. I then pull from the input queue, predict the output transaction, then compare to the actual.  Being able to then know when all my input queues are empty allows me to know when to terminate simulation.

The correct time to terminate the simulation is after the comparison is done, not after the queue is emptied, as you might end the test before completing the comparison. Creating a third queue for comparison results would fix this, and eliminate the need for `empty`.

eric-wieser:
> 1. Now that the task tracking stuff has been removed, @eric-wieser 's argument that the task tracking stuff satisfies this use case is invalid.

My argument is that if presented with a problem that would be solved by the task tracking stuff, we should just add it back rather than inventing something different.

jwprice100:
> The correct time to terminate the simulation is after the comparison is done, not after the queue is emptied, as you might end the test before completing the comparison. Creating a third queue for comparison results would fix this, and eliminate the need for `empty`.

The way I typically do this, the queue contains data for the comparison. That is, the coroutines that do the comparison pull from this queue. When all such Queues are empty, the test is done. A common structure is I spawn off a bunch of monitor coroutines that each receive bus transactions,  then pull from the input queue and do comparisons. The main test coroutine then generates transactions and shoves them into buses and these input queues. Then waits for all the input queues to drain before ending. I think this is a perfectly valid way of viewing things and works well for me. 

But yes, your third queue also works! There are number of other ways to accomplish this, they just feel less natural.

eric-wieser:
Your approach is dangerous, because the queue empties before the last comparison is performed, which means you can end up concluding the test without performing the comparison at all.

jwprice100:
> Your approach is dangerous, because the queue empties before the last comparison is performed, which means you can end up concluding the test without performing the comparison at all.

I assume you're thinking of some race condition where a coroutine waiting on empty wake up before a coroutine that called get on the last item? That seems like it would be possible to avoid.

eric-wieser:
> That seems like it would be possible to avoid.

While we could design an API that promises a certain scheduler wakeup order, this seems fragile, and it seems easier just to inform users that the safest way is to use an additional queue.

For instance, even if the queue does awake the reader first, what happens if that reader awaits `cocotb.external` to send some data to a bit of threaded code for it to be checked? You don't want the scheduler to then resume the task that did `await queue.empty()`, because then you still have the same problem.

PR #2444, opened 2021-03-06
Backport Queue full/empty waits

marlonjames:
Backport of #2443 for 1.5


ktbarrett (approved):

imphil:
I'm sorry, we're shortly before the release and only accept bug fixes at this point, no new features. If there is a desire for users to have similar functionality with cocotb 1.5 they can rather easily provide a version themselves. The discussion in #2443 also shows that we do not have agreement on how this feature should look. There's always a next release!

PR #2451, opened 2021-03-09
Strip whitespace in TOPLEVEL name to prevent bugs

### LINKED ISSUE #2448: TOPLEVEL should be whitespace stripped
From here: https://github.com/cocotb/cocotb/blob/ecb43878a6af0605fdf08c2f1829cc401bfdeb36/cocotb/__init__.py#L219-L225

On Questa with the FLI, trailing whitespace causes the toplevel entity to not be found since it does a strict `!strcmp()`.

ktbarrett:
Closes #2448.

### DIFF
--- a/documentation/source/building.rst
+++ b/documentation/source/building.rst
@@ -47,6 +47,7 @@ Cocotb
 
     Use this to indicate the instance in the hierarchy to use as the :term:`DUT`.
     If this isn't defined then the first root instance is used.
+    Leading and trailing whitespace are automaticall discarded.
 
     The DUT is available in cocotb tests as a Python object at :data:`cocotb.top`;
     and is also passed to all cocotb tests as the :ref:`first and only parameter <quickstart_creating_a_test>`.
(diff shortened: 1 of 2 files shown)


marlonjames (approved):

cmarqu (approved):

eric-wieser:
I think a note in the docs saying whitespace is stripped might be a good idea.

eric-wieser:
To be clear, I mean a comment in the documentation for environment variables, not a release note. I probably wouldn't consider this a bugfix anyway - It's just making things easier for users who are sloppy with their makefiles (which isn't to say I'm opposed to it).

documentation/source/building.rst
```diff
@@ -47,6 +47,7 @@ Cocotb
 
     Use this to indicate the instance in the hierarchy to use as the :term:`DUT`.
     If this isn't defined then the first root instance is used.
+    Leading and trailing whitespace are automaticall discarded.
```
eric-wieser, line 50:
```suggestion
    Leading and trailing whitespace are automatically discarded.
```

eric-wieser: Could add `..versionchanged` too

PR #2459, opened 2021-03-10
Remove cocotb bus docs

ktbarrett:
Lots of changes

- removed endian_swapper and HAL tutorials since they are no longer in the project
- removed all other examples that aren't in the project: AXI Lite, Mean, others
- removed "Testbenching Tools" page since it was 90% about the bus components
- removed the bus objects from the library reference
- removed the generator objects from the library reference

Depends on #2457.


documentation/source/testbench_tools.rst
```diff
@@ -1,218 +0,0 @@
...
-
-Cocotb uses the builtin :mod:`logging` library, with some configuration described in :ref:`logging-reference-section` to provide some sensible defaults.
-Each :term:`DUT`, monitor, driver, and scoreboard (as well as any other function using the coroutine decorator) holds a :class:`logging.Logger`, and each can be set to its own logging level.
-Within a :term:`DUT`, each hierarchical object can also have individual logging levels set.
-
-When logging :term:`HDL` objects, beware that ``_log`` is the preferred way to use
-logging. This helps minimize the change of name collisions with an :term:`HDL` log
-component with the Python logging functionality.
-
-Log printing levels can also be set on a per-object basis.
```
cmarqu, line 16: I'm not sure if we have this information somewhere else still, maybe some of it could move to "Writing Testbenches"?

ktbarrett: I had thought about it, but I would need to rewrite the examples to not reference the bus objects. I took the lazy way out. I'll work on porting it tonight.

cmarqu: I wasn't thinking of the examples really (I could live without them), just this information about `logging` being used and that each object can have its own log level etc.

ktbarrett: So just move the highlight bit to "Writing Testbenches" in the logging section? I can do that.

ktbarrett: This was moved to Writing Testbenches and the wording and examples were touched up.

documentation/source/writing_testbenches.rst
```diff
@@ -217,3 +217,26 @@ thus the access will raise an :exc:`AttributeError`.
...
+and can be set to its own logging level.
+
+.. code-block:: python3
+
+    task = cocotb.fork(coro)
+    task.log.setLevel(logging.DEBUG)
+    task.log.debug("Running Task!")
+
+The :term:`DUT` and each hierarchical object can also have individual logging levels set.
+When logging :term:`HDL` objects, beware that ``_log`` is the preferred way to use
```
cmarqu, line 235: I guess we will remove the existing compat_mapping and so _log will be the only way to log.

cmarqu (approved):

PR #2461, opened 2021-03-10
gitpod: Use simple_dff example, update tool versions

cmarqu:


ktbarrett (approved):
I don't know much about gitpod, but this looks fairly straightforward.

themperek (approved):

PR #2465, opened 2021-03-12
Add VHDL library support to questa

martijnbastiaan:
For GHDL users can specify:

```
VHDL_SOURCES_a = [..]
```

To compile a bunch of VHDL files into a library called `a`. This PR adds support for this feature for modelsim/questasim.


ktbarrett:
I'll have a look on Monday when I'm near a Questa license.

cocotb/share/makefiles/Makefile.sim
```diff
@@ -57,7 +57,7 @@ SIM                       Selects which simulator Makefile to use
 WAVES                     Enable wave traces dump for Riviera-PRO and Questa
 VERILOG_SOURCES           A list of the Verilog source files to include
 VHDL_SOURCES              A list of the VHDL source files to include
-VHDL_SOURCES_<lib>        VHDL source files to include in *lib* (GHDL only)
+VHDL_SOURCES_<lib>        VHDL source files to include in *lib* (GHDL/Modelsim/Quastasim only)
```
cmarqu, line 60:
```suggestion
VHDL_SOURCES_<lib>        VHDL source files to include in *lib* (GHDL/ModelSim/Questa only)
```

documentation/source/building.rst
```diff
@@ -268,7 +268,7 @@ The following variables are makefile variables, not environment variables.
 
 .. make:var:: VHDL_SOURCES_<lib>
 
-      A list of the VHDL source files to include in the VHDL library *lib* (currently for the GHDL simulator only).
+      A list of the VHDL source files to include in the VHDL library *lib* (currently for GHDL/Modelsim/Questasim only).
```
cmarqu, line 271:
```suggestion
      A list of the VHDL source files to include in the VHDL library *lib* (currently for GHDL/ModelSim/Questa only).
```

documentation/source/building.rst
```diff
@@ -268,7 +268,7 @@ The following variables are makefile variables, not environment variables.
 
 .. make:var:: VHDL_SOURCES_<lib>
 
-      A list of the VHDL source files to include in the VHDL library *lib* (currently for GHDL/Modelsim/Questasim only).
+      A list of the VHDL source files to include in the VHDL library *lib* (currently for GHDL/ModelSim/Questa only).
```
cmarqu, line 271:
```suggestion
      A list of the VHDL source files to include in the VHDL library *lib* (currently for GHDL/ModelSim/Questa only).
      
      .. versionadded:: 2.0 Add support for ModelSim and Questa.
```

cmarqu (commented):
Sorry, I forgot this in the first review round.

cmarqu:
Please also add a newsfragment (see https://github.com/cocotb/cocotb/blob/master/documentation/source/newsfragments/README.rst).

cmarqu:
A test using this feature would also be nice.
We shipped this for GHDL all this time without a test, but GHDL support is limited while Questa is in heavy use among our users.

martijnbastiaan:
Sure thing, this would only be tested on CI with GHDL, correct?

ktbarrett:
We don't yet have a Questa license to be able to test this in CI (but one is coming soon, potentially). It would be nice to have a yes/no test for this feature for me and @cmarqu to test it manually.

martijnbastiaan:
@ktbarrett Thanks! I've added a test. You can simply run `make` in `tests/test_cases/test_vhdl_libraries`. I've tested with ModelSim up to (but not including) simulation and up to and including simulation with GHDL. 

Could you help out with CI? I'm not sure what the proper way is to prevent the test from running on Verilog simulators.

themperek:
> Could you help out with CI? I'm not sure what the proper way is to prevent the test from running on Verilog simulators.

Something like this:

https://github.com/cocotb/cocotb/blob/83476d8babcabc38a9dede451768d9fc196501e0/examples/mixed_language/tests/Makefile#L4-L16

ktbarrett:
The specific Makefile I would use is:

```makefile
SIM ?= ghdl
TOPLEVEL_LANG ?= vhdl
VHDL_SOURCES_blib := Test.b/b.vhdl
VHDL_SOURCES := Test.a/a.vhdl
TOPLEVEL := a
MODULE := test_ab

ifneq ($(shell echo $(TOPLEVEL_LANG) | tr A-Z a-z),vhdl)
all:
	@echo "Skipping test since only VHDL is supported"
clean::
else
endif
```

ktbarrett (commented):
Your newsfragment must be a pull request or issue number. I suggest the pull request: `2465.feature.rst` should be the name.

documentation/source/newsfragments/vhdl_libraries_questa.feature.rst
```diff
@@ -0,0 +1,3 @@
+VHDL libraries can now be used in combination with Modelsim/Questa
+
+GHDL have been able to use VHDL libraries by using ``VHDL_SOURCES_libName`` for a while. This is now also supported for Modelsim/Questa.
```
cmarqu, line 3:
```suggestion
ModelSim and Questa now support compilation into a named VHDL library ``lib`` using :make:var:`VHDL_SOURCES_<lib>`.
```

tests/test_cases/test_vhdl_libraries/Test.b/b.vhdl
```diff
@@ -0,0 +1,36 @@
+library IEEE;
```
cmarqu, line 1: I would reduce the complexity of the VHDL files down to the bare minimum to not distract from what you really want to show.
E.g. a hierarchy of two entities (without ports or with a single port) that just `report` their `<entity>'instance_name` or `<entity>'path_name` (I forget which one shows more info)?
If you want to show that the VHDL library being used is transparent to Python, you could have a process that does the `report`ing that you can trigger from the Python test.

cmarqu: Oh, and please use all-lowercase directory names too. And you don't actually need those subdirectories anyway, right?
Some beginner might take those as "being the VHDL libraries" which would be wrong.

Thanks for the test and the quick follow-ups to the reviews!

martijnbastiaan: > I would reduce the complexity of the VHDL files down to the bare minimum to not distract from what you really want to show.

Done!

> Oh, and please use all-lowercase directory names too. And you don't actually need those subdirectories anyway, right?

That's correct. It's just something the tool I use to generate VHDL does and I didn't think of removing it.

> Thanks for the test and the quick follow-ups to the reviews!

Np, thanks for being so helpful!

martijnbastiaan:
@cmarqu @ktbarrett I've applied all suggestions once again. Thanks for being so helpful!

Running `make` now prints a message similar to:

```
b.vhdl:9:5:@0ms:(report note): :a(structural):b@b(structural):
```

I'm unsure how to actually test for this in the cocotb test case though. Having said that, I don't think it's really necessary: if something would break library support it should simply stumble on `library blib;` in `a.vhdl`.

ktbarrett (approved):
@martijnbastiaan Thanks for the work. Looks good. The release notes docs are busted, but I don't think it has anything to do with you. I'll fix them up tonight and rerun the CI in this PR, so it could be merged tonight/tomorrow unless @cmarqu has more notes. You don't need to test the reported value, the makefiles in the regression are basically the test, seeing if the libraries can be built or not. 

I spoke too soon. The Riviera regressions are failing because this feature isn't supported there yet. I'll add a fix for the Makefile soon.

tests/test_cases/test_vhdl_libraries/Makefile
```diff
@@ -0,0 +1,14 @@
+SIM ?= ghdl
+TOPLEVEL_LANG ?= vhdl
+VHDL_SOURCES_blib := b.vhdl
+VHDL_SOURCES := a.vhdl
+TOPLEVEL := a
+MODULE := test_ab
+
+ifneq ($(shell echo $(TOPLEVEL_LANG) | tr A-Z a-z),vhdl)
```
themperek, line 8: One needs to restrict this also to ghdl/modesim/questa) since Riviera will fail. (Or add this functionality to Riviera)
Example: https://github.com/cocotb/cocotb/blob/aa956f83e119e4ca4615c5b8660eb15c4dfa8623/examples/mixed_signal/tests/Makefile#L67

martijnbastiaan: Yeah, it's only supported in GHDL/ModelSim/Questa for now. @ktbarrett mentioned adding fix, so I won't interfere for a bit.

tests/test_cases/test_vhdl_libraries/Makefile
```diff
@@ -0,0 +1,14 @@
...
+all:
+	@echo "Skipping test since only VHDL is supported"
+clean::
+else
```
ktbarrett, line 12:
```suggestion
else ifeq ($(filter ghdl questa modelsim,$(shell echo $(SIM) | tr A-Z a-z)),)
all:
	@echo "Skipping test since only GHDL and Questa/Modelsim is supported"
clean::
else
```

2465.feature.rst
```diff
@@ -0,0 +1 @@
+ModelSim and Questa now support compilation into a named VHDL library ``lib`` using :make:var:`VHDL_SOURCES_<lib>`.
```
cmarqu, line 1: This file needs to live in `documentation/source/newsfragments`.

cmarqu (approved):
Please move the newsfragment into the right directory, and then it's good to go IMO.
Thanks very much!

2465.feature.rst
ktbarrett: Ah, that's why it isn't showing up. I thought it was due to the previous PR's newsfragment getting screwed up. I'll still add that fix tonight.

martijnbastiaan: Oh huh, I swear it was in there before. Sorry about the troubles, I'll move it right now.

martijnbastiaan: Hmmyeah, earlier comments indicate it was in there. Must have botched something in the process. 

Anyway, suggestion applied, commits squashed. Should be good to go now!

cmarqu: Ah, the linking to the make variable `VHDL_SOURCES_<lib>` doesn't work unfortunately, and the `<lib>` part isn't shown. Not sure how that would need to be quoted. :/
Maybe the linking isn't that important and we can just write ` ``VHDL_SOURCES_<lib>`` `.

martijnbastiaan: Yeah. I've changed it to ``VHDL_SOURCES_<lib>`` :)

martijnbastiaan: Well with the extra ` that is, but I can't get GitHub to cooperate.

cmarqu: LGTM!

PR #2466, opened 2021-03-12
Refector Code Quality Issues

ankitdobhal:
## Description

Hey 👋 , I work at DeepSource, I ran the DeepSource static analyzer on the forked copy of this repo and found some [interesting code quality issues](https://deepsource.io/gh/ankitdobhal/cocotb/issues). This PR fixes a few of them.

## Following are the changes-

- Remove implicit `object` from the base class
- Use literal syntax to create data structure
- Remove unnecessary return statement
- Add .deepsource.toml file for continuous analysis on bug risks/performance/code-quality issues on new changes

_Note: I've fixed a few issues for now. You can find and fix other issues [here](https://deepsource.io/gh/ankitdobhal/cocotb/issues)._

## Type of change

- [x] Antipattern
- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code

### DIFF
--- a/cocotb/wavedrom.py
+++ b/cocotb/wavedrom.py
@@ -168,7 +168,6 @@ class trace:
         for sig in self._signals:
             sig.clear()
         self.disable()
-        return None
 
     def write(self, filename, **kwargs):
         with open(filename, "w") as f:
--- a/tests/test_cases/test_compare/test_compare.py
+++ b/tests/test_cases/test_compare/test_compare.py
@@ -11,7 +11,7 @@ from cocotb.handle import SimHandleBase, NonHierarchyObject
 from cocotb.triggers import RisingEdge, FallingEdge
 
 
-class Testbench(object):
+class Testbench():
 
     def __init__(self, dut):
         self.dut = dut
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,12 @@
+version = 1
+
+test_patterns = ["tests/**"]
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
+
+
--- a/examples/mixed_signal/tests/test_regulator_trim.py
+++ b/examples/mixed_signal/tests/test_regulator_trim.py
@@ -7,7 +7,7 @@ from cocotb.triggers import Timer
 import itertools
 
 
-class Regulator_TB(object):
+class Regulator_TB():
     """Class for collecting testbench objects.
 
     Args:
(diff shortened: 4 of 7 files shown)


cocotb/scheduler.py
```diff
@@ -453,24 +451,13 @@ def _event_loop(self, trigger):
                     if _debug:
                         self.log.debug("Scheduled coroutine %s" % (coro._coro.__qualname__))
 
-                    # remove our reference to the objects at the end of each loop,
-                    # to try and avoid them being destroyed at a weird time (as
-                    # happened in gh-957)
-                    del coro
-
```
eric-wieser, line 460: This `del`, and many of the others, is important.

tests/test_cases/test_compare/test_compare.py
```diff
@@ -11,7 +11,7 @@
 from cocotb.triggers import RisingEdge, FallingEdge
 
 
-class Testbench(object):
+class Testbench():
```
eric-wieser, line 14: This change is good, but I'd be inclined to drop the `()` everywhere too.

I also think that #2423 probably does this for us anyway.

imphil:
I'm sorry, but this PR feels a bit like an ad for DeepSource. We can discuss if it brings something useful for us, but not in a PR. And as @eric-wieser said, the Python2 -> Python3 issues should be solved in #2423

cocotb/scheduler.py
ankitdobhal: @eric-wieser reverted the changes.

cocotb/wavedrom.py
```diff
@@ -168,7 +168,6 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         for sig in self._signals:
             sig.clear()
         self.disable()
-        return None
```
ktbarrett, line 171: In fact. I think this is the only change not covered by #2423.

ktbarrett (commented):
Looking at the report generated there are other issues that could be addressed like importing `cocotb.simulator` multiple times in `_initialize_testbench`. The definition of `__iter__` on the decorators (which I noted before as being wrong). Lots of instances of unused variables rather than using `_`. Is there a reason you didn't include any of those changes?

cmarqu:
E.g. pylint still also has lots of things to report that may be fishy in the code.
flake8 found unused variables in the tests and examples, but I feel having them named makes the code "more didactic", so our flake8 config ignores these reports for those files.

ktbarrett:
Seeing as the current set of code changes are not really useful, I guess this PR can be more about whether we want to add DeepSource support. I can't tell from their website whether running it solely in CI is considered a single "team member" or not. If there is any restriction on our use in CI, I don't want it in the repo.

ktbarrett:
Gonna say no. If someone else agrees, go ahead an close this PR.

PR #2472, opened 2021-03-14
Add underscores to make big numbers more readable

cmarqu:
This could be seen as unnecessary churn.
Opinions?


ktbarrett (approved):
I prefer it.

PR #2475, opened 2021-03-16
Fix newsfragment for #2434

ktbarrett:
ReStructuredText syntax was broken a bit. Fixed that and slightly clarified the text.


documentation/source/newsfragments/2434.change.rst
```diff
@@ -1 +1 @@
-Selecting tests using :make:var:`TESTCASE` will now search for the first occurrence of a test of that name in order of :make:var:`MODULE`s, and not just the first module.
+Selecting tests using :make:var:`TESTCASE` will now search for the first occurrence of a test of that name in order of models in :make:var:`MODULE`\ s, and not just the first module in that list.
```
ktbarrett, line 1:
```suggestion
Selecting tests using :make:var:`TESTCASE` will now search for the first occurrence of a test of that name in order of modules listed in :make:var:`MODULE`\ s, and not just the first module in that list.
```

cmarqu (approved):

PR #2476, opened 2021-03-17
Add dependency on cocotb-bus to 1.5

ktbarrett:
What we did with cocotb-bus was perhaps a little overzealous. We added deprecation warnings, but we still broke users by requiring them to manually install cocotb-bus to resolve import errors when importing from `cocotb.bus`, etc. Adding cocotb-bus as a dependency will fix the break for 1.5. This does not need to be added to the master branch.


ktbarrett:
I screwed up. Wrong base.

PR #2477, opened 2021-03-17
Add dependency on cocotb-bus to 1.5

ktbarrett:
What we did with cocotb-bus was perhaps a little overzealous. We added deprecation warnings, but we still broke users by requiring them to manually install cocotb-bus to resolve import errors when importing from `cocotb.bus`, etc. Adding cocotb-bus as a dependency will fix the break for 1.5. This does not need to be added to the master branch.

### DIFF
--- a/documentation/source/release_notes.rst
+++ b/documentation/source/release_notes.rst
@@ -14,6 +14,7 @@ Bugfixes
 --------
 
 - Prevent pytest assertion rewriting (:pr:`2028`) from capturing stdin, which causes problems with IPython support (:pr:`1649`). (:pr:`2462`)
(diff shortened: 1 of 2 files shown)


marlonjames:
If `cocotb-bus` is a dependency, shouldn't it be removed from extras?

ktbarrett:
It could be, but I'm not sure it's necessary. We have already created that API, I don't think we should break it at this point.

marlonjames (approved):

imphil (approved):
LGTM, with the same comment as for the other PR going into 1.5: please edit the release notes directly.

PR #2481, opened 2021-03-19
Revert BinaryValue change

### LINKED ISSUE #2480: Binary representation with Cocotb 1.5
In the class BinaryValue there is the possibility to pass during initialization a parameter called binaryRepresentation
` def __init__(self, value=None, n_bits=None, bigEndian=True,
                 binaryRepresentation=BinaryRepresentation.UNSIGNED,
                 bits=None):`

This setting can later be read out over the 'binaryRepresentation' variable.
While this was editable in cocotb1.4, changing this variable in 1.5 has no influence on the conversion functions any more. 
The representation can only be set during creation of the object. 

I think this change was introduced by @marlonjames on Jan 5

I would suggest to
- make the variable binaryRepresentation private, so users don't try to change it any more. 
or
- update the conversion function when binaryRepresentation is chagned

Maybe you have other inputs?

ktbarrett:
Closes #2480. Reintroduces to the 1.5 release the ability of the user to change the `binaryRepresentation` of a `BinaryValue` object after its creation.

### DIFF
--- a/cocotb/binary.py
+++ b/cocotb/binary.py
@@ -157,12 +157,18 @@ class BinaryValue:
 
         self._n_bits = n_bits
 
-        self._convert_to = self._convert_to_map[self.binaryRepresentation].__get__(self, self.__class__)
-
-        self._convert_from = self._convert_from_map[self.binaryRepresentation].__get__(self, self.__class__)
-
         if value is not None:
             self.assign(value)
+
+    @property
+    def binaryRepresentation(self):
+        return self._binaryRepresentation
+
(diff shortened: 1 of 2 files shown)


marlonjames:
@Botnic can you try this PR with your use-case?

ktbarrett:
I can add a small test.

Botnic:
Hi ktbarrett
Yes, your solution works for my usecase. 
No modification on the tests was necessary to make them work again. 
Thank you

imphil (approved):
LGTM, thanks!

Issue #2482, opened 2021-03-19
Cant't pass plusargs with more than one "=" sign to testbench


bk388:
https://github.com/cocotb/cocotb/blob/master/cocotb/__init__.py#L347
The line above causes cocotb to crash if we attempt to give it a plusarg with more than one "=" sign in it.

PR #2483, opened 2021-03-19
Allow plusargs to have values containing `=`

bk388:
Fixes https://github.com/cocotb/cocotb/issues/2482

Ensuring plusargs are always split into at most two parts
This would enable passing plusargs with more than one "=" sign to a testbench


marlonjames:
Please also update `test_plusargs`.

marlonjames (approved):

PR #2484, opened 2021-03-20
Load custom entry point in Python from the GPI

### LINKED ISSUE #1225: Allow a custom entry point

ktbarrett:
Closes #1225. From https://github.com/cocotb/cocotb/issues/1225#issuecomment-738573856, solution 2. Sorry the diff is mostly unreadable. I found it was more work than it was worth to split the changes between DEFER and the rest.

The value of this PR is that we can easily experiment with alternative regression managers and schedulers while keeping existing code working by loading into another entry point. It can also be used by users to setup hooks on cocotb functionality before a test runs.

gpi_embed is no longer hardcoded to enter
"cocotb:_initialize_testbench", but uses an entry module and function
specified by the COCOTB_ENTRY_POINT environment variable. The variable
is of the format: "path.to.entry.module:entry_point.function". The GPI
embed also assumes the "_log_from_c", "_filter_from_c", and "_sim_event"
function are located in the entry module. This may be cleaned up in the
future.

Additionally, the body of the "_embed_sim_init" function has been
improved to remove some of the complexity regarding object lifetimes in
the function by using a "DEFER" statement, which is seen in a few other
languages.

Finally, "py_gpi_log_initialize" was changed to increment the ref on the
passed functions to simplify object lifetime concerns in
"_embed_sim_init".

## TODO

- [x] document COCOTB_ENTRY_POINT
- [x] test

### DIFF
--- a/cocotb/share/lib/embed/gpi_embed.cpp
+++ b/cocotb/share/lib/embed/gpi_embed.cpp
@@ -188,172 +188,138 @@ extern "C" COCOTB_EXPORT void _embed_sim_cleanup(void) {
     }
 }
 
-/**
- * @name    Initialization
- * @brief   Called by the simulator on initialization. Load cocotb Python module
- * @ingroup python_c_api
- *
- * GILState before calling: Not held
- *
- * GILState after calling: Not held
- *
- * Makes one call to PyGILState_Ensure and one call to PyGILState_Release
- *
- * Loads the Python module called cocotb and calls the _initialise_testbench
- * function
- */
+namespace {
 
-static int get_module_ref(const char *modname, PyObject **mod) {
-    PyObject *pModule = PyImport_ImportModule(modname);
+template <typename F>
+class Deferable {
+  public:
+    constexpr Deferable(F f) : f_(f){};
+    ~Deferable() { f_(); }
 
-    if (pModule == NULL) {
-        // LCOV_EXCL_START
-        PyErr_Print();
-        LOG_ERROR("Failed to load Python module \"%s\"", modname);
-        return -1;
-        // LCOV_EXCL_STOP
-    }
+  private:
+    F f_;
+};
 
-    *mod = pModule;
-    return 0;
+template <typename F>
+constexpr Deferable<F> make_deferable(F f) {
+    return Deferable<F>(f);
 }
+
+}  // namespace
+
+#define DEFER1(a, b) a##b
+#define DEFER0(a, b) DEFER1(a, b)
+#define DEFER(statement) \
+    auto DEFER0(_defer, __COUNTER__) = make_deferable([&]() { statement; });
 
 extern "C" COCOTB_EXPORT int _embed_sim_init(int argc,
                                              char const *const *argv) {
-    int i;
-    int ret = 0;
-
-    /* Check that we are not already initialized */
-    if (pEventFn) return ret;
-
-    PyObject *cocotb_module, *cocotb_init, *cocotb_retval;
-    PyObject *cocotb_log_module = NULL;
-    PyObject *log_func;
-    PyObject *filter_func;
-    PyObject *argv_list;
-
-    cocotb_module = NULL;
+    // Check that we are not already initialized
+    if (pEventFn) {
+        return 0;
+    }
 
     // Ensure that the current thread is ready to call the Python C API
-    PyGILState_STATE gstate = PyGILState_Ensure();
+    auto gstate = PyGILState_Ensure();
+    DEFER(PyGILState_Release(gstate));
+
     to_python();
+    DEFER(to_simulator());
 
-    if (get_module_ref("cocotb", &cocotb_module)) {
-        // LCOV_EXCL_START
-        goto cleanup;
-        // LCOV_EXCL_STOP
-    }
-
-    LOG_INFO("Python interpreter initialized and cocotb loaded!");
-
-    if (get_module_ref("cocotb.log", &cocotb_log_module)) {
-        // LCOV_EXCL_START
-        goto cleanup;
-        // LCOV_EXCL_STOP
-    }
-
-    // Obtain the function to use when logging from C code
-    log_func = PyObject_GetAttrString(cocotb_log_module,
-                                      "_log_from_c");  // New reference
-    if (log_func == NULL) {
+    auto entry_utility_module = PyImport_ImportModule("pygpi.entry");
+    if (!entry_utility_module) {
         // LCOV_EXCL_START
         PyErr_Print();
-        LOG_ERROR("Failed to get the _log_from_c function");
-        goto cleanup;
+        return -1;
         // LCOV_EXCL_STOP
     }
+    DEFER(Py_DECREF(entry_utility_module));
 
-    // Obtain the function to check whether to call log function
-    filter_func = PyObject_GetAttrString(cocotb_log_module,
-                                         "_filter_from_c");  // New reference
-    if (filter_func == NULL) {
-        // LCOV_EXCL_START
-        Py_DECREF(log_func);
-        PyErr_Print();
-        LOG_ERROR("Failed to get the _filter_from_c method");
-        goto cleanup;
-        // LCOV_EXCL_STOP
-    }
-
-    py_gpi_logger_initialize(
-        log_func, filter_func);  // Note: This function steals references to
-                                 // log_func and filter_func.
-
-    pEventFn =
-        PyObject_GetAttrString(cocotb_module, "_sim_event");  // New reference
-    if (pEventFn == NULL) {
+    auto entry_info_tuple =
+        PyObject_CallMethod(entry_utility_module, "load_entry", NULL);
+    if (!entry_info_tuple) {
         // LCOV_EXCL_START
         PyErr_Print();
-        LOG_ERROR("Failed to get the _sim_event method");
-        goto cleanup;
+        return -1;
         // LCOV_EXCL_STOP
     }
+    DEFER(Py_DECREF(entry_info_tuple));
 
-    cocotb_init = PyObject_GetAttrString(
-        cocotb_module, "_initialise_testbench");  // New reference
-    if (cocotb_init == NULL) {
+    PyObject *entry_module;
+    PyObject *entry_point;
+    if (!PyArg_ParseTuple(entry_info_tuple, "OO", &entry_module,
+                          &entry_point)) {
         // LCOV_EXCL_START
         PyErr_Print();
-        LOG_ERROR("Failed to get the _initialise_testbench method");
-        goto cleanup;
+        return -1;
         // LCOV_EXCL_STOP
     }
+    // Objects returned from ParseTuple are borrowed from tuple
+
+    auto log_func = PyObject_GetAttrString(entry_module, "_log_from_c");
+    if (!log_func) {
+        // LCOV_EXCL_START
+        PyErr_Print();
+        return -1;
+        // LCOV_EXCL_STOP
+    }
+    DEFER(Py_DECREF(log_func));
+
+    auto filter_func = PyObject_GetAttrString(entry_module, "_filter_from_c");
+    if (!filter_func) {
+        // LCOV_EXCL_START
+        PyErr_Print();
+        return -1;
+        // LCOV_EXCL_STOP
+    }
+    DEFER(Py_DECREF(filter_func));
+
+    py_gpi_logger_initialize(log_func, filter_func);
+
+    pEventFn = PyObject_GetAttrString(entry_module, "_sim_event");
+    if (!pEventFn) {
+        // LCOV_EXCL_START
+        PyErr_Print();
+        return -1;
+        // LCOV_EXCL_STOP
+    }
+    // cocotb must hold _sim_event until _embed_sim_end runs
 
     // Build argv for cocotb module
-    argv_list = PyList_New(argc);  // New reference
+    auto argv_list = PyList_New(argc);
     if (argv_list == NULL) {
         // LCOV_EXCL_START
         PyErr_Print();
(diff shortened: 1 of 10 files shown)


pygpi/entry.py
```diff
@@ -0,0 +1,18 @@
+import os
+import importlib
+from functools import reduce
+
+
+def load_entry():
+    """Gather entry point information by parsing :envar:`COCOTB_ENTRY_POINT`."""
```
cmarqu, line 7:
```suggestion
    """Gather entry point information by parsing :envvar:`COCOTB_ENTRY_POINT`."""
```

pygpi/entry.py
```diff
@@ -0,0 +1,18 @@
+import os
+import importlib
+from functools import reduce
+
+
+def load_entry():
```
marlonjames, line 6: I'm not sure how useful this actually is:
```suggestion
from types import ModuleType
from typing import Tuple, Callable


def load_entry() -> Tuple[ModuleType, Callable]:
```

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -188,172 +188,138 @@ extern "C" COCOTB_EXPORT void _embed_sim_cleanup(void) {
...
+    Py_DECREF(cocotb_retval);
 
-    goto ok;
-
-cleanup:
-    ret = -1;
-ok:
-    Py_XDECREF(cocotb_module);
-    Py_XDECREF(cocotb_log_module);
+    LOG_INFO("cocotb initialization successful");
```
marlonjames, line 320: Should this stay `DEBUG`? Or maybe removed altogether. It doesn't show up until the first yield back to the simulator, so it can look out of place. 

Example: https://github.com/cocotb/cocotb/pull/2484/checks?check_run_id=2157206436#step:25:4296

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -188,172 +188,138 @@ extern "C" COCOTB_EXPORT void _embed_sim_cleanup(void) {
...
+    pEventFn = PyObject_GetAttrString(entry_module, "_sim_event");
+    if (!pEventFn) {
         // LCOV_EXCL_START
         PyErr_Print();
-        LOG_ERROR("Failed to get the _initialise_testbench method");
-        goto cleanup;
+        return -1;
         // LCOV_EXCL_STOP
     }
+    // cocotb must hold _sim_event until _embed_sim_end runs
```
marlonjames, line 286:
```suggestion
    // cocotb must hold _sim_event until _embed_sim_cleanup runs
```

cocotb/share/lib/embed/gpi_embed.cpp
ktbarrett: I thought the same thing. I'd like to remove it.
```suggestion
```

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -188,172 +188,137 @@ extern "C" COCOTB_EXPORT void _embed_sim_cleanup(void) {
...
     }
+    Py_DECREF(cocotb_retval);
 
-    goto ok;
-
-cleanup:
-    ret = -1;
-ok:
-    Py_XDECREF(cocotb_module);
-    Py_XDECREF(cocotb_log_module);
```
ktbarrett, line 320:
```suggestion
```

ktbarrett:
Questions:
1. Should we put the entry loader code in another package (`pygpi`) or in the existing `cocotb` package? It currently is in another package. If the user intends to use the simulator, they will need to reach into the `cocotb` package anyways for `cocotb.simulator`. Or we could move the `simulator` module to the new `pygpi` package in a follow on?
2. Should th environment variable be prefixed with `COCOTB_`, or with `PYGPI_`?

documentation/source/building.rst
```diff
@@ -222,6 +222,27 @@ GPI
         and loading from libraries that `aren't` prefixed with "lib".
         Paths `should not` contain commas.
 
+PyGPI
+-----
+
+.. envvar:: COCOTB_ENTRY_POINT
+
+    The Python module and callable that starts up the Python cosimulation environment.
+    This defaults to :value:`cocotb:_initialise_testbench`, which is the cocotb standard entry point.
```
marlonjames, line 231: The `:value:` isn't rendering on the docs.

ktbarrett: Yeah it should be `:data:`. I keep getting that one wrong.

ktbarrett:
```suggestion
    This defaults to :data:`cocotb:_initialise_testbench`, which is the cocotb standard entry point.
```

documentation/source/building.rst
```diff
@@ -222,6 +222,27 @@ GPI
...
+    The variable is formatted as ``path.to.entry.module:entry_point.function``.
+    The string before the colon is the Python module to import
+    and the string following the colon is the object to call as the entry function.
+
+    The entry function must be a callable that takes a single positional argument, the command ``argv`` list.
+    The entry module must have the following additional functions defined:
+
+    * ``_sim_event(level: int)``
+    * ``_log_from_c(logger_name: str, level: int, filename: str, lineno: int, msg: str, function_name: str))``
+    * ``_filter_from_c(logger_name: str, level: int)``
```
marlonjames, line 242: `_filter_from_c` return is checked for `True`. The others have errors on `Exception`s, but maybe they should allow returning a bool for success/failure as well?
```suggestion
    * ``_sim_event(level: int) -> bool``
    * ``_log_from_c(logger_name: str, level: int, filename: str, lineno: int, msg: str, function_name: str)) -> bool``
    * ``_filter_from_c(logger_name: str, level: int) -> bool``
```

documentation/source/building.rst
```diff
@@ -222,6 +222,27 @@ GPI
...
+
+    These additional requirements on the entry module may be relaxed over time.
```
marlonjames, line 244:
```suggestion
    The entry module functions return ``True`` on success. On failure, they should return ``False`` or raise an :class:`Exception`.
    These additional requirements on the entry module may be relaxed over time.
```

cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -188,172 +188,136 @@ extern "C" COCOTB_EXPORT void _embed_sim_cleanup(void) {
...
- *
- * Loads the Python module called cocotb and calls the _initialise_testbench
- * function
- */
+namespace {
 
-static int get_module_ref(const char *modname, PyObject **mod) {
-    PyObject *pModule = PyImport_ImportModule(modname);
+template <typename F>
+class Deferable {
```
eric-wieser, line 194: This is basically `gsl::final_action`

https://github.com/microsoft/GSL/blob/84aeb59f269ee9747e85d780c8aa180bca3a9a60/include/gsl/util#L50-L89

which is basically [E.19 from the C++ code guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-finally).

I'm not saying we should depend on the gsl, but I think it would be a good idea to reuse the name `final_action`.

Any chance you could put this `Deferable` cleanup in a small standalone PR?

documentation/source/building.rst
```diff
@@ -222,6 +222,31 @@ GPI
...
+
+    * ``entry_function(argv: List[str]) -> bool``
+
+    The entry module must have the following additional functions defined:
+
+    * ``_sim_event(level: int) -> bool``
+    * ``_log_from_c(logger_name: str, level: int, filename: str, lineno: int, msg: str, function_name: str)) -> bool``
+    * ``_filter_from_c(logger_name: str, level: int) -> bool``
+
+    The entry module functions return ``True`` on success. On failure, they should return ``False`` or raise an :class:`Exception`.
```
eric-wieser, line 247: This API seems odd to me; I'd prefer the API be something like
```
def my_handle_event(level: int) -> None:
   # raise exception on error

def my_log_from_c(...) -> None:
   # raise exception on error

def my_filter_from_c(...) -> bool:
   # return bool indicating whether to filter the message, raise exception on error

def entry_function(argv: List[str]) -> bool:
    # raise exception on error
    cocotb.simulator.set_sim_event_handler(my_handle_event)
    cocotb.simulator.set_log_handler(my_log_from_c, my_filter_from_c)
```
or perhaps
```
def entry_function(argv: List[str]):
    return (my_handle_event, my_log_from_c, my_filter_from_c)
```

Returning `bool`s to indicate errors is unpythonic, and I don't think we should be doing it.

eric-wieser: Note that this approach allows different handlers to be registered based on the `argv`.

ktbarrett: The interface is odd, but I just encoded what cocotb was already doing. I'd prefer the first approach, and it's where I intend to end up; which is why I mentioned "these additional requirements will be relaxed in the future". Doing it all at once will make this diff huge since the object will need to live in some common library used by both the simulatormodule and gpi_embed. I have mentioned maybe we can combine gpi_embed and simulatormodule into a single library to avoid that unnecessary complexity.

ktbarrett: I know it's better, I'm trying to chip away bit by bit because large diffs don't get reviewed.

cocotb/share/lib/embed/gpi_embed.cpp
ktbarrett: The whole point of making it was so I didn't have to think through all the object lifetime issues in sim_embed_init and which things need to be freed at which exit points. If I split it out, I did the work I was trying to avoid and it's pointless. If not introducing it now is preferable I'll just refactor it out.

ktbarrett: It was also accepted into the C++20 standard https://en.cppreference.com/w/cpp/experimental/scope_exit.

documentation/source/building.rst
marlonjames: We can change them to return None and use Exceptions. I noticed that `_initialise_testbench` returned True. Whichever way it goes make the functions we have match the documented behavior.

marlonjames:
To your questions:
1. I think leaving it in the separate `pygpi` package is fine.
2. I like `PYGPI_*` to keep the names tied to the layers: `GPI_EXTRA`, `PYGPI_*`, `COCOTB_LOG_LEVEL`, etc.

documentation/source/building.rst
marlonjames: For now let's change them to raise on error and fixup the documentation. As Kaleb said we can iterate and improve on it in a later PR.

ktbarrett: I removed the boolean returns. But the names remain the same and they remain required-by-name members of the init module. A follow on will move the interface to instead be registered by calling functions on the `simulator` module.

ktbarrett:
Coverage is only decreasing because of a change in total LoC in the project.

marlonjames:
The new test diff is failing on mingw. LF vs CRLF?

ktbarrett:
Seems that Linux closes the file and doesn't complain, but mingw complains. https://github.com/cocotb/cocotb/pull/2484/checks?check_run_id=3366618894#step:5:1267

tests/test_cases/test_custom_entry/custom_entry.py
```diff
@@ -0,0 +1,13 @@
+from typing import List
+
+from cocotb.log import _filter_from_c, _log_from_c  # noqa: F401
+
+file = open("results.log", "w")
```
ktbarrett, line 5:
```suggestion
```

tests/test_cases/test_custom_entry/custom_entry.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def entry_func(argv: List[str]) -> None:
+    print("got entry", file=file)
```
ktbarrett, line 9:
```suggestion
    with open("results.log", "w") as file:
        print("got entry", file=file)
```

tests/test_cases/test_custom_entry/custom_entry.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def _sim_event(level: int, message: str) -> None:
+    print(f"got event level={level} message={message}", file=file)
```
ktbarrett, line 13:
```suggestion
    with open("results.log", "a") as file:
        print(f"got event level={level} message={message}", file=file)
```

ktbarrett:
Apparently it's that `diff` doesn't come standard with mingw.

tests/test_cases/test_custom_entry/Makefile
```diff
@@ -0,0 +1,9 @@
+export PYTHONPATH := .
+export PYGPI_ENTRY_POINT := custom_entry:entry_func
+
+.PHONY: override_for_this_test
+override_for_this_test:
+	-$(MAKE) all
+	diff --strip-trailing-cr results.log expected_results.log &>/dev/null
```
ktbarrett, line 7:
```suggestion
	python -c 'assert open("results.log".readlines() == open("expected_results.log").readlines()'
```

marlonjames (approved):
I'm good with this.
Thanks for doing the work!

Issue #2059, opened 2020-09-09
[tracking] Redesign BinaryValue


imphil:
BinaryValue has a fair number of issues which cannot be solved in a backwards-compatible way. We are in the process of coming up with a new design, this issue tracks the progress on it.

ktbarrett:
# Minimal Requirements
* Accurately model...
    * (System)Verilog 4-value logic type
    * VHDL std_logic and bit
* Vectors of that type (???)
    * Is this a true requirement? Icarus supports single element access now. This might be a quality of life thing.
    * When getting or setting a vector of logic, we could instead yield or take a list of logic elements.

## Quality of life
* Model...
    * (System)Verilog packed logic arrays
    * VHDL std_logic_vector
* Slicing
* Logic operations
* Arbitrary bounds
* Generic / nestable vectors to support N dimensional arrays (???)
* Convert Python integers to/from 1D vectors of logics
    * Two's complement
    * Unsigned
    * Signed magnitude (???)
    * One's complement (???)
* Convert Python strings to/from 1D vectors of logics

## Antifeatures of BinaryValue
* "endianness"
    * confusing, does it mean to/downto range or byte endianness?
    * unnecessary since the new type won't deal in bytes
* associating integer representation with the object
    * associate with the conversion operator
    * have implicit conversions (like in `handle <= 1`) infer either two's complement or unsigned
* "buff", conversion to/from bytes
    * convert to/from integer and use `int.to_bytes` and `int.from_bytes` instead
* integer operators
    * convert to an integer first and get well defined behavior...

## Backwards compatibility
* COCOTB_RESOLVE_X (???)

eric-wieser:
> integer operators
> * convert to an integer first and get well defined behavior...

One think that might be neat is if `b'010X + b'10X0 = b'11XX` - but maybe that's not what verilog does, and maybe that logic belongs elsewhere anyway

ktbarrett:
I'd have to look into it. Considering how adders work, all bits more significant than an X should be an X, but that's not how they are typically modeled are they?

And I am trying to stay away from the replacement being a modeling datatype, we can let other libraries handle that. The type just needs to be a way to express logic values and offer conversions to different representations.

eric-wieser:
> Considering how adders work, all bits more significant than an X should be an X, but that's not how they are typically modeled are they?

[...]

PR #2485, opened 2021-03-21
Add Logic and Bit datatypes

ktbarrett:
Logic (4-value) and Bit (2-value) datatypes models are added for use in system modelling and eventual use in the handle class. xref #2059.

I think doing this piecemeal is better than all at once, we can put a lot of attention into each piece.
A heterogenous `Array` type is basically done, and ` LogicArray` specialization is implemented and needs some testing.
I am thinking we will think through how we want to intregrate these types with `cocotb.handle` after we get the types in.
Since `cocotb.handle` just emits and consumes `BinaryValue`s (it doesn't do anything weird like update references in place or anything like that), any types with value semantics should be easily substitutable (and by "easily", I mean it will break a lot of crap).

The one thing I'm iffy about is the use of `cache` (unbounded cache) on functions that can return `NotImplemented` in response to *any* value put in. I feel better knowing that 99.999999 times out of 100 it will result in an exception. The only way to cause an issue with this would be through malice.

### DIFF
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -60,3 +60,13 @@ if sys.version_info[:2] >= (3, 7):
 else:
     import collections
     insertion_ordered_dict = collections.OrderedDict
+
+
+# backport of Python 3.8's functools.cache decorator
+if sys.version_info < (3, 9):
+    from functools import lru_cache
+
+    def cache(f):
+        return lru_cache(maxsize=None)(f)
+else:
+    from functools import cache  # noqa: F401
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -62,7 +62,7 @@ else:
     insertion_ordered_dict = collections.OrderedDict
 
 
-# backport of Python 3.8's functools.cache decorator
+# backport of Python 3.9's functools.cache decorator
 if sys.version_info < (3, 9):
     from functools import lru_cache
 
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -29,7 +29,6 @@ These are for internal use - users should use a third party library like `six`
 if they want to use these shims in their own code
 """
 import sys
-from functools import lru_cache
 
 
 # backport of Python 3.7's contextlib.nullcontext
@@ -61,11 +60,3 @@ if sys.version_info[:2] >= (3, 7):
 else:
     import collections
     insertion_ordered_dict = collections.OrderedDict
-
-
-# backport of Python 3.9's functools.cache decorator
-if sys.version_info < (3, 9):
-    def cache(f):
-        return lru_cache(maxsize=None)(f)
-else:
-    from functools import cache  # noqa: F401
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -29,6 +29,7 @@ These are for internal use - users should use a third party library like `six`
 if they want to use these shims in their own code
 """
 import sys
+from functools import lru_cache
 
 
 # backport of Python 3.7's contextlib.nullcontext
@@ -64,8 +65,6 @@ else:
 
 # backport of Python 3.9's functools.cache decorator
 if sys.version_info < (3, 9):
-    from functools import lru_cache
-
     def cache(f):
         return lru_cache(maxsize=None)(f)
 else:
--- a/cocotb/types/__init__.py
+++ b/cocotb/types/__init__.py
@@ -1,1 +1,4 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
 from .logic import Logic, Bit  # noqa: F401
--- a/cocotb/types/__init__.py
+++ b/cocotb/types/__init__.py
@@ -1,0 +1,1 @@
+from .logic import Logic, Bit  # noqa: F401
--- a/cocotb/types/logic.py
+++ b/cocotb/types/logic.py
@@ -1,7 +1,7 @@
 # Copyright cocotb contributors
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
-from typing import Any, Optional
+from typing import Any, Optional, Dict
 from typing import Tuple  # noqa: F401
 from functools import lru_cache
 
@@ -71,6 +71,8 @@ class Logic:
     """
     __slots__ = ("_repr",)
 
+    __singleton_cache__: Dict[int, "Logic"] = {}
+
     _repr_map = {
         # 0 and weak 0
         False: 0,
@@ -107,8 +109,11 @@ class Logic:
             raise ValueError(
                 "{!r} is not convertible to a {}".format(value, cls.__qualname__)
             ) from None
-        obj = super().__new__(cls)
-        obj._repr = _repr
+        obj = cls.__singleton_cache__.get(_repr, None)
+        if obj is None:
+            obj = super().__new__(cls)
+            obj._repr = _repr
+            cls.__singleton_cache__[_repr] = obj
         return obj
 
     def __and__(self, other: "Logic") -> "Logic":
@@ -254,6 +259,8 @@ class Bit(Logic):
         ValueError: if the value cannot be constructed into a :class:`Bit`.
     """
     __slots__ = ()
+
+    __singleton_cache__: Dict[int, "Bit"] = {}
 
     _repr_map = {
         # 0
--- a/cocotb/types/logic.py
+++ b/cocotb/types/logic.py
@@ -137,10 +137,6 @@ class Logic:
     def __invert__(self):
         return type(self)(("1", "0", "X", "X")[self._repr])
 
-    __eq__ = object.__eq__
-
-    __hash__ = object.__hash__
-
     @cache
     def __repr__(self):
         return "{}({!r})".format(type(self).__name__, str(self))
--- a/cocotb/types/logic.py
+++ b/cocotb/types/logic.py
@@ -33,7 +33,7 @@ class Logic:
     Model of a 4-value (``0``, ``1``, ``X``, ``Z``) datatype commonly seen in HDLs.
 
     Modeled after (System)Verilog's 4-value ``logic`` type.
-    VHDL's 8-value ``std_logic`` type maps to this type by treating weak values as full strength values
+    VHDL's 9-value ``std_ulogic`` type maps to this type by treating weak values as full strength values
     and "uninitialized" (``U``) and "don't care" (``-``) as "unknown" (``X``).
 
     Supports common logic operations ``&``, ``|``, ``^``, and ``~``.
--- a/cocotb/types/logic.py
+++ b/cocotb/types/logic.py
@@ -9,30 +9,64 @@ class Logic:
     r"""
     Model of a 4-value (``0``, ``1``, ``X``, ``Z``) datatype commonly seen in HDLs.
 
+    .. currentmodule:: cocotb.types
+
     This is modeled after (System)Verilog's 4-value ``logic`` type.
     VHDL's 9-value ``std_ulogic`` type maps to this type by treating weak values as full strength values
     and treating "uninitialized" (``U``) and "don't care" (``-``) as "unknown" (``X``).
 
-    :class:`Logic` can be converted from :class:`int`, :class:`str`, :class:`bool`, and :class:`~cocotb.types.Bit` using the ``Logic(value)`` syntax.
-    The list of acceptable values includes ``0``, ``1``, ``True``, ``False``, ``"0"``, ``"1"``, ``"X"``, and ``"Z"``.
+    :class:`Logic` can be converted to and from :class:`int`, :class:`str`, :class:`bool`, and :class:`Bit`
+    by using the appropriate constructor syntax.
+    The list of values convertable to :class:`Logic` includes
+    ``0``, ``1``, ``True``, ``False``, ``"0"``, ``"1"``, ``"X"``, ``"Z"``, ``Bit('0')``, and ``Bit('1')``.
     For a comprehensive list of values that can be converted into :class:`Logic` see :file:`tests/pytest/test_logic.py`.
 
-    :class:`Logic` can be converted to :class:`int`, :class:`str`, :class:`bool` using the appropriate constructor syntax.
-    For example, ``int(Logic(0)) == 0``, ``bool(Logic(1)) is True``, and ``str(Logic("X")) == "X"``.
-    The :class:`int` and :class:`bool` conversions will raise :exc:`ValueError` if the value is not ``0`` or ``1``.
+    .. code-block:: python3
 
-    The default value of ``Logic()`` is ``Logic("X")``.
+        >>> Logic("X")
+        Logic('X')
+        >>> Logic(True)
+        Logic('1')
+        >>> Logic(1)
+        Logic('1')
+        >>> Logic(Bit(0))
+        Logic('0')
+
+        >>> Logic()  # default value
+        Logic('X')
+
+        >>> str(Logic("Z"))
+        'Z'
+        >>> bool(Logic(0))
+        False
+        >>> int(Logic(1))
+        1
+        >>> Bit(Logic("1"))
+        Bit('1')
+
+    .. note::
+
+        The :class:`int` and :class:`bool` conversions will raise :exc:`ValueError` if the value is not ``0`` or ``1``.
 
     :class:`Logic` values are hashable and can be placed in :class:`set`\ s and used as keys in :class:`dict`\ s.
 
     :class:`Logic` supports the common logic operations ``&``, ``|``, ``^``, and ``~``.
 
-    .. code-block:: py3
+    .. code-block:: python3
 
-        def full_adder(a: Logic, b: Logic, c_in: Logic) -> (Logic, Logic):
-            res = a ^ b ^ c_in
-            c_out = (a & b) | (b & c_in) | (a & c_in)
-            return res, c_out
+        >>> def full_adder(a: Logic, b: Logic, carry: Logic) -> (Logic, Logic):
+        ...     res = a ^ b ^ carry
+        ...     carry_out = (a & b) | (b & carry) | (a & carry)
+        ...     return res, carry_out
+
+        >>> full_adder(a=Logic('0'), b=Logic('1'), carry=Logic('1'))
+        (Logic('0'), Logic('1'))
+
+    Args:
+        value: value to construct into a :class:`Logic`.
+
+    Raises:
+        ValueError: if the value cannot be constructed into a :class:`Logic`.
     """
     __slots__ = ("_repr",)
 
@@ -153,17 +187,49 @@ class Bit(Logic):
     r"""
     Model of a 2-value (``0``, ``1``) datatype commonly seen in HDLs.
 
+    .. currentmodule:: cocotb.types
+
     This is modeled after (System)Verilog's 2-value ``bit`` type.
     VHDL's ``bit`` type maps to this type perfectly.
 
-    :class:`Bit` can be converted from :class:`int`, :class:`str`, :class:`bool`, and :class:`~cocotb.types.Logic` using the ``Bit(value)`` syntax.
-    The list of acceptable values includes ``0``, ``1``, ``True``, ``False``, ``"0"``, ``"1"``.
+    :class:`Bit` is a proper subtype of :class:`Logic`, meaning a use of :class:`Logic` can be substituted with a :class:`Bit`.
+    Some behavior may surprise you if you do not expect it.
+
+    .. code-block:: python3
+
+        >>> Bit(0) == Logic(0)
+        True
+        >>> Bit(0) in {Logic(0)}
+        True
+
+    :class:`Bit` can be converted to and from :class:`int`, :class:`str`, :class:`bool`, and :class:`Logic`
+    by using the appropriate constructor syntax.
+    The list of values convertable to :class:`Bit` includes
+    ``0``, ``1``, ``True``, ``False``, ``"0"``, ``"1"``, ``Logic('0')``, and ``Logic('1')``.
     For a comprehensive list of values that can be converted into :class:`Bit` see :file:`tests/pytest/test_logic.py`.
 
-    :class:`Bit` can be converted to :class:`int`, :class:`str`, :class:`bool` using the appropriate constructor syntax.
-    For example, ``int(Bit(0)) == 0``, ``bool(Bit(1)) is True``, and ``str(Bit("1")) == "1"``.
+    .. code-block:: python3
 
-    The default value of ``Bit()`` is ``Bit("0")``.
+        >>> Bit("0")
+        Bit('0')
+        >>> Bit(True)
+        Bit('1')
+        >>> Bit(1)
+        Bit('1')
+        >>> Bit(Logic(0))
+        Bit('0')
+
+        >>> Bit()  # default value
+        Bit('0')
+
+        >>> str(Bit("0"))
+        '0'
+        >>> bool(Bit(False))
+        False
+        >>> int(Bit(1))
+        1
+        >>> Logic(Bit("1"))
+        Logic('1')
 
     :class:`Bit` values are hashable and can be placed in :class:`set`\ s and used as keys in :class:`dict`\ s.
 
@@ -171,8 +237,20 @@ class Bit(Logic):
 
     .. code-block:: py3
 
-        def mux(a: Bit, b: Bit, s: Bit) -> Bit
-            return (a & ~s) | (b & s)
+        >>> def mux(a: Bit, b: Bit, s: Bit) -> Bit:
+        ...     return (a & ~s) | (b & s)
+
+        >>> a = Bit(0)
+        >>> b = Bit(1)
+        >>> sel = Bit(1)  # choose second argument
+        >>> mux(a, b, sel)
+        Bit('1')
+
+    Args:
+        value: value to construct into a :class:`Bit`.
+
+    Raises:
+        ValueError: if the value cannot be constructed into a :class:`Bit`.
     """
     __slots__ = ()
 
--- a/cocotb/types/logic.py
+++ b/cocotb/types/logic.py
@@ -17,13 +17,13 @@ operator `isinstance(other, type(self)`
 strict type equality
     To use `Logic` or `Bit` in hashable collections we need to define `__hash__` and `__eq__` such that
     "Hashable objects which compare equal must have the same hash value."
-    Retrieved from https://hynek.me/articles/hashes-and-equality/ on 2021-03-20.
+    (retrieved from https://hynek.me/articles/hashes-and-equality/ on 2021-03-20).
     The best way to achieve this is to make `Logic` and `Bit` never equal, like `tuple` and `list`.
     If we instead made them equal in hash and value, they would be substitutable; which we don't want.
 
 @cache
-    Shows extreme performance improvements.
-    Even faster than precomputing the results and storing them in the type.
+    Shows extreme performance improvements,
+    even faster than precomputing the results and storing them in the type.
     Everything can be cached since there are a (hopefully) limited number of valid values in all subclasses.
 """
 
@@ -32,21 +32,21 @@ class Logic:
     r"""
     Model of a 4-value (``0``, ``1``, ``X``, ``Z``) datatype commonly seen in HDLs.
 
-    Modeled after (System)Verilog's 4-value ``logic`` type.
+    This is modeled after (System)Verilog's 4-value ``logic`` type.
     VHDL's 9-value ``std_ulogic`` type maps to this type by treating weak values as full strength values
     and treating "uninitialized" (``U``) and "don't care" (``-``) as "unknown" (``X``).
 
-    Can be converted from :class:`int`, :class:`str`, :class:`bool`, and :class:`~cocotb.types.Bit` using the ``Logic(value)`` syntax.
+    :class:`Logic` can be converted from :class:`int`, :class:`str`, :class:`bool`, and :class:`~cocotb.types.Bit` using the ``Logic(value)`` syntax.
     The list of acceptable values includes ``0``, ``1``, ``True``, ``False``, ``'0'``, ``'1'``, ``'X'``, and ``'Z'``.
     For a comprehensive list of values that can be converted into :class:`Logic` see :file:`tests/pytest/test_logic.py`.
 
-    Can be converted to :class:`int`, :class:`str`, :class:`bool` using the appropriate constructor syntax.
+    :class:`Logic` can be converted to :class:`int`, :class:`str`, :class:`bool` using the appropriate constructor syntax.
     For example, ``int(Logic(0)) == 0``, ``bool(Logic(1)) is True``, and ``str(Logic('X')) == 'X'``.
     The :class:`int` and :class:`bool` conversions will raise :exc:`ValueError` if the value is not ``0`` or ``1``.
 
     :class:`Logic` values are hashable and can be placed in :class:`set`\ s and used as keys in :class:`dict`\ s.
 
-    Supports common logic operations ``&``, ``|``, ``^``, and ``~``.
+    :class:`Logic` supports the common logic operations ``&``, ``|``, ``^``, and ``~``.
 
     .. code-block:: py3
 
@@ -182,19 +182,19 @@ class Bit(Logic):
     r"""
     Model of a 2-value (``0``, ``1``) datatype commonly seen in HDLs.
 
-    Modeled after (System)Verilog's 2-value ``bit`` type.
+    This is modeled after (System)Verilog's 2-value ``bit`` type.
     VHDL's ``bit`` type maps to this type perfectly.
 
-    Can be converted from :class:`int`, :class:`str`, :class:`bool`, and :class:`~cocotb.types.Logic` using the ``Bit(value)`` syntax.
+    :class:`Bit` can be converted from :class:`int`, :class:`str`, :class:`bool`, and :class:`~cocotb.types.Logic` using the ``Bit(value)`` syntax.
     The list of acceptable values includes ``0``, ``1``, ``True``, ``False``, ``'0'``, ``'1'``.
     For a comprehensive list of values that can be converted into :class:`Bit` see :file:`tests/pytest/test_logic.py`.
 
-    Can be converted to :class:`int`, :class:`str`, :class:`bool` using the appropriate constructor syntax.
+    :class:`Bit` can be converted to :class:`int`, :class:`str`, :class:`bool` using the appropriate constructor syntax.
     For example, ``int(Bit(0)) == 0``, ``bool(Bit(1)) is True``, and ``str(Bit('1')) == '1'``.
 
     :class:`Bit` values are hashable and can be placed in :class:`set`\ s and used as keys in :class:`dict`\ s.
 
-    Supports common logic operations ``&``, ``|``, ``^``, and ``~``.
+    :class:`Bit` supports the common logic operations ``&``, ``|``, ``^``, and ``~``.
 
     .. code-block:: py3
 
--- a/cocotb/types/logic.py
+++ b/cocotb/types/logic.py
@@ -7,26 +7,24 @@ from cocotb._py_compat import cache
 """ DESIGN NOTES
 
 __singleton_cache__
-    Used to ensure there is only ever one instance of a particular value. This is a
-    memory optimization: there will only ever be 4 objects allocated; and also a speed
-    optimization: value equality can be identity equality (as inherited from the `object`
-    base class).
+    Used to ensure there is only ever one instance of a particular value.
+    This is a memory optimization: there will only ever be 4 objects allocated;
+    and also a speed optimization: value equality can be identity equality (as inherited from the `object` base class).
 
 operator `isinstance(other, type(self)`
     Allows operation between a `Bit` and a `Logic` to return a `Logic`.
 
 strict type equality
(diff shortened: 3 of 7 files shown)


cocotb/_py_compat.py
```diff
@@ -60,3 +60,13 @@ def __exit__(self, *excinfo):
 else:
     import collections
     insertion_ordered_dict = collections.OrderedDict
+
+
+# backport of Python 3.8's functools.cache decorator
```
ktbarrett, line 65:
```suggestion
# backport of Python 3.9's functools.cache decorator
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,205 @@
...
+operator `isinstance(other, type(self)`
+    Allows operation between a `Bit` and a `Logic` to return a `Logic`.
+
+strict type equality
+    To use `Logic` or `Bit` in hashable collections we need to define `__hash__` and
+    `__eq__` such that "Hashable objects which compare equal must have the same hash
+    value." (https://hynek.me/articles/hashes-and-equality/, 3/20/21). The best way to
+    achieve this is to make `Logic` and `Bit` never equal, like `tuple` and `list`. If we
+    instead made them equal in hash and value, they would be substitutable; which we
+    don't want.
```
ktbarrett, line 21: Just a comment as to why we don't want substitutability here:
```python
d = {Logic(0)}
assert Bit(0) in d  # passes
```

While technically valid, it would be surprising.

documentation/source/library_reference.rst
```diff
@@ -64,6 +64,17 @@ Interacting with the Simulator
...
 .. autofunction:: cocotb.decorators.RunningTask.kill
 
+HDL Datatypes
+-------------
+
+These are a set of datatypes that model the behavior of common HDL datatypes.
+They can be used independently of cocotb for modeling and will replace :class:`BinaryValue`
+as the types used by cocotb's `simulator handles <#simulation-object-handles>`_.
+
+.. autoclass:: cocotb.types.Logic
```
ktbarrett, line 74: This just shows the type docstring and nothing else since everything is a dunder method. Should we turn on showing dunder methods on these classes?

cocotb/types/logic.py
```diff
@@ -0,0 +1,205 @@
+from cocotb._py_compat import cache
+
+
+""" DESIGN NOTES
+
+__singleton_cache__
+    Used to ensure there is only ever one instance of a particular value. This is a
```
cmarqu, line 7: "One sentence per line" would make for better diffs in the future.

cocotb/types/logic.py
```diff
@@ -0,0 +1,205 @@
...
+@cache
+    Shows extreme performance improvements. Even faster than precomputing the results and
+    storing them in the type. Everything can be cached since there are a (hopefully)
+    limited number of valid values in all subclasses.
+"""
+
+
+class Logic:
+    """
+    Models 4-value (0, 1, X, Z) datatype commonly seen in HDLs
```
cmarqu, line 32:
```suggestion
    Model 4-value (0, 1, X, Z) datatype commonly seen in HDLs.
```

tests/pytest/test_logic.py
```diff
@@ -0,0 +1,222 @@
+from cocotb.types import Logic, Bit
```
cmarqu, line 1: Missing the SPDX header.

cocotb/types/__init__.py
```diff
@@ -0,0 +1 @@
+from .logic import Logic, Bit  # noqa: F401
```
cmarqu, line 1: Missing the SPDX header.

cocotb/types/logic.py
```diff
@@ -0,0 +1,205 @@
+from cocotb._py_compat import cache
```
cmarqu, line 1: Missing the SPDX header.

documentation/source/library_reference.rst
```diff
@@ -64,6 +64,17 @@ Interacting with the Simulator
...
+
+.. autoclass:: cocotb.types.Bit
+
```
cmarqu, line 77:
```suggestion
.. versionadded:: 2.0

```

cocotb/types/logic.py
```diff
@@ -0,0 +1,206 @@
...
+    This is a memory optimization: there will only ever be 4 objects allocated;
+    and also a speed optimization: value equality can be identity equality (as inherited from the `object` base class).
+
+operator `isinstance(other, type(self)`
+    Allows operation between a `Bit` and a `Logic` to return a `Logic`.
+
+strict type equality
+    To use `Logic` or `Bit` in hashable collections we need to define `__hash__` and `__eq__` such that
+    "Hashable objects which compare equal must have the same hash value."
+    Retreived from https://hynek.me/articles/hashes-and-equality/ on 20/3/21.
```
cmarqu, line 20:
```suggestion
    Retrieved from https://hynek.me/articles/hashes-and-equality/ on 20/3/21.
```

documentation/source/library_reference.rst
cmarqu: Maybe the description of these types and their operators is easier to understand when written in a few consecutive paragraphs instead of each operator having its own docstring.

cocotb/types/logic.py
```diff
@@ -0,0 +1,206 @@
...
+        "H": 1,
+        "h": 1,
+        # unknown, unassigned, and weak unknown
+        None: 2,
+        "X": 2,
+        "x": 2,
+        "U": 2,
+        "u": 2,
+        "W": 2,
+        "w": 2,
```
cmarqu, line 65: Is `-` missing here?

tests/pytest/test_logic.py
```diff
@@ -0,0 +1,225 @@
...
+def test_bit_bool_conversions():
+    assert bool(Bit(1)) is True
+    assert bool(Bit(0)) is False
+
+
+def test_logic_str_conversions():
+    assert str(Logic('0')) == '0'
+    assert str(Logic('1')) == '1'
+    assert str(Logic('X')) == 'X'
+    assert str(Logic('Z')) == 'Z'
```
cmarqu, line 95:
```suggestion
    assert str(Logic('Z')) == 'Z'
    assert str(Logic('W')) == 'X'
    assert str(Logic('-')) == 'X'
    assert str(Logic('z')) == 'Z'
    assert str(Logic('x')) == 'X'
```
?

cocotb/types/logic.py
```diff
@@ -0,0 +1,206 @@
...
+@cache
+    Shows extreme performance improvements.
+    Even faster than precomputing the results and storing them in the type.
+    Everything can be cached since there are a (hopefully) limited number of valid values in all subclasses.
+"""
+
+
+class Logic:
+    """
+    Model of 4-value (0, 1, X, Z) datatype commonly seen in HDLs.
```
cmarqu, line 33:
```suggestion
    Model of a 4-value (``0``, ``1``, ``X``, ``Z``) datatype commonly seen in HDLs.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,206 @@
...
+        if self._repr < 2:
+            return self._repr
+        raise ValueError(
+            "Cannot convert non-0/1 {} to int".format(type(self).__qualname__)
+        )
+
+
+class Bit(Logic):
+    """
+    Model of 2-value (0, 1) datatype commonly seen in HDLs.
```
cmarqu, line 170:
```suggestion
    Model of a 2-value (``0``, ``1``) datatype commonly seen in HDLs.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,206 @@
...
+
+    Modeled after (System)Verilog's 2-value ``bit`` type.
+    VHDL's ``bit`` type maps to this type perfectly.
+
+    Supports common logic operations ``&``, ``|``, ``^``, and ``~``.
+    Can be converted to and from :class:`int`, :class:`str`, :class:`bool`, and :class:`cocotb.types.Logic`..
```
cmarqu, line 176:
```suggestion
    Can be converted to and from :class:`int`, :class:`str`, :class:`bool`, and :class:`cocotb.types.Logic`.
```

cocotb/types/logic.py
ktbarrett: Good catch, it is.

tests/pytest/test_logic.py
ktbarrett: I don't think we need to test that exactly if we can prove that `Logic('x') == Logic('-')`, etc. I don't have a test for that.

cocotb/types/logic.py
ktbarrett:
```suggestion
        "w": 2,
        "-": 2,
```

tests/pytest/test_logic.py
```diff
@@ -0,0 +1,225 @@
...
+def test_logic_invert():
+    assert ~Logic(0) == Logic(1)
+    assert ~Logic(1) == Logic(0)
+    assert ~Logic('X') == Logic('X')
+    assert ~Logic('Z') == Logic('X')
+
+
+def test_bit_invert():
+    assert ~Bit(0) == Bit(1)
+    assert ~Bit(1) == Bit(0)
```
ktbarrett, line 225: Missing a test for hashability (see if we can stick it in a set). This could also be a good way to test that all the different literals for a value result in the same value by checking if a set holding the values is of size 1 after adding all the literals.

documentation/source/library_reference.rst
ktbarrett: I'll got about expanding on examples in the type docstring using `..codeblock`s.

tests/pytest/test_logic.py
ktbarrett: Added tests for identity check between value constructed from different literals.

tests/pytest/test_logic.py
```diff
@@ -0,0 +1,284 @@
...
+    Logic('L')
+    Logic('1')
+    Logic('h')
+    Logic('H')
+    Logic('x')
+    Logic('X')
+    Logic('w')
+    Logic('W')
+    Logic('u')
+    Logic('U')
```
marlonjames, line 24:
```suggestion
    Logic('U')
    Logic('-')
```

tests/pytest/test_logic.py
```diff
@@ -0,0 +1,284 @@
...
+    assert Logic('h') == l
+    assert Logic('H') == l
+    assert Logic(Logic('1')) == l
+
+    l = Logic('X')
+    assert Logic('x') == l
+    assert Logic('w') == l
+    assert Logic('W') == l
+    assert Logic('u') == l
+    assert Logic('U') == l
```
marlonjames, line 100:
```suggestion
    assert Logic('U') == l
    assert Logic('-') == l
```

tests/pytest/test_logic.py
ktbarrett: It's in the name, I can't help it.

cocotb/types/logic.py
```diff
@@ -0,0 +1,207 @@
...
+    def __rxor__(self, other):
+        return self ^ other
+
+    @cache
+    def __invert__(self):
+        return type(self)(("1", "0", "X", "X")[self._repr])
+
+    __eq__ = object.__eq__
+
+    __hash__ = object.__hash__
```
eric-wieser, line 142: Why is this necessary?

cocotb/types/logic.py
```diff
@@ -0,0 +1,207 @@
...
+        raise ValueError(
+            "Cannot convert non-0/1 {} to bool".format(type(self).__qualname__)
+        )
+
+    @cache
+    def __int__(self):
+        if self._repr < 2:
+            return self._repr
+        raise ValueError(
+            "Cannot convert non-0/1 {} to int".format(type(self).__qualname__)
```
eric-wieser, line 165: Why not print whether it was `x` or `z` here?

tests/pytest/test_logic.py
```diff
@@ -0,0 +1,286 @@
...
+    assert Logic(1) != Bit(1)
+
+
+def test_logic_self_identity():
+    l = Logic('0')
+    assert Logic('l') == l
+    assert Logic('L') == l
+    assert Logic(0) == l
+    assert Logic(False) == l
+    assert Logic(Logic('0')) == l
```
eric-wieser, line 87: If you're testing `identity` you should use `is` not `==`

cocotb/types/logic.py
```diff
@@ -0,0 +1,207 @@
...
+    def __new__(cls, value=None):
+        # convert to internal representation
+        try:
+            _repr = cls._repr_map[value]
+        except KeyError:
+            raise ValueError(
+                "{!r} is not convertable to a {}".format(value, cls.__qualname__)
+            ) from None
+        # ensure only one object is made per representation
+        if _repr not in cls.__singleton_cache__:
```
eric-wieser, line 82: Does the `@cache` above mean that this `if` is always true?

cocotb/types/logic.py
ktbarrett: It isn't necessary, just wanted to be explicit. Many types override `__eq__` which causes the inherited `__hash__` to turn off.

cocotb/types/logic.py
ktbarrett: Sure. Originally this was a 9-value logic type, so non-0/1 was shorter.

tests/pytest/test_logic.py
ktbarrett: I think its just poorly named. I don't want `is` in the test, its an implementation detail. I just want to show that certain literals all result in the same value.

cocotb/types/logic.py
eric-wieser: Sure, and that's a feature. I don't see any reason to keep these lines; by having them, we _increase_ the chance of a mistake if someone modifies `__eq__` but forgets to modify `__hash__`.

tests/pytest/test_logic.py
eric-wieser: > its an implementation detail.

Are you sure we don't want to declare it part of the API? For instance, python's `Enum` declares it part of the API, https://docs.python.org/3/library/enum.html#comparisons

> Enumeration members are compared by identity

And IMO `Logic` is just a glorified enum with some bitwise-ish operators.

cocotb/types/logic.py
```diff
@@ -0,0 +1,207 @@
...
+
+operator `isinstance(other, type(self)`
+    Allows operation between a `Bit` and a `Logic` to return a `Logic`.
+
+strict type equality
+    To use `Logic` or `Bit` in hashable collections we need to define `__hash__` and `__eq__` such that
+    "Hashable objects which compare equal must have the same hash value."
+    Retrieved from https://hynek.me/articles/hashes-and-equality/ on 2021-03-20.
+    The best way to achieve this is to make `Logic` and `Bit` never equal, like `tuple` and `list`.
+    If we instead made them equal in hash and value, they would be substitutable; which we don't want.
```
eric-wieser, line 22: I don't see a test for this - can you add a test that `len({Bit(1), Logic(1)}) == 2`?

cocotb/types/logic.py
eric-wieser: Just to be clear, I meant `Cannot convert {self!r} to int` or similar, not `cannot convert X/Z to int`.

cocotb/types/logic.py
```diff
@@ -0,0 +1,207 @@
...
+
+@cache
+    Shows extreme performance improvements.
+    Even faster than precomputing the results and storing them in the type.
+    Everything can be cached since there are a (hopefully) limited number of valid values in all subclasses.
+"""
+
+
+class Logic:
```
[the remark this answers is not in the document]
ktbarrett: That is why I avoided it.

cocotb/types/logic.py
ktbarrett: No. The caches are different. This will fill the `__singleton_cache__` with the 4 values and ensure only those objects are returned, and the `@cache` will map the accepted convertible values to the value. Just using `@cache` is not sufficient. Instead of this line I could fill the cache after the class definition.

```python
class Logic:
    # ...

Logic.__singleton_cache__[0] = object.__new__(Logic)
Logic.__singleton_cache__[1] = object.__new__(Logic)
Logic.__singleton_cache__[2] = object.__new__(Logic)
Logic.__singleton_cache__[3] = object.__new__(Logic)
```

tests/pytest/test_logic.py
ktbarrett: > Are you sure we don't want to declare it part of the API?

Not sure why we would want to. I know they do, but I don't see the value. `is` would work like `==` and I think that's it. I don't want people to do that either.

> And IMO Logic is just a glorified enum with some bitwise-ish operators.

Exactly. That's how it's defined in both Verilog and VHDL. There isn't much to it. Most of the magic happens in the array of logic specialization, which is yet to come.

cocotb/types/logic.py
ktbarrett, line 142:
```suggestion
```

cocotb/types/logic.py
ktbarrett:
```suggestion
            "Cannot convert {sefl!r} to int".format(type(self).__qualname__)
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,203 @@
...
+    @cache
+    def __str__(self):
+        return ("0", "1", "X", "Z")[self._repr]
+
+    @cache
+    def __bool__(self):
+        if self._repr < 2:
+            return bool(self._repr)
+        raise ValueError(
+            "Cannot convert non-0/1 {} to bool".format(type(self).__qualname__)
```
ktbarrett, line 153:
```suggestion
            "Cannot convert {self!r} to bool".format(type(self).__qualname__)
```

cocotb/types/logic.py
eric-wieser: Ok, that makes sense to me. I don't think filling the cache after really makes any difference, and it's clearer the way you have it.

tests/pytest/test_logic.py
eric-wieser: > I know they do, but I don't see the value

IMO consistency with design decisions in Python is an argument in itself, especially in the absence of counter-arguments.
As a bonus, `is` is faster than `==`.

> Exactly. That's how it's defined in both Verilog and VHDL.

Sorry, to be clear, my point is since we agree that `Logic` is a glorified enum, it makes sense to match the bits of the Python API that don't get in the way.

cocotb/types/logic.py
```diff
@@ -0,0 +1,203 @@
...
+    @cache
+    def __str__(self):
+        return ("0", "1", "X", "Z")[self._repr]
+
+    @cache
+    def __bool__(self):
+        if self._repr < 2:
+            return bool(self._repr)
+        raise ValueError(
+            "Cannot convert {self!r} to bool".format(type(self).__qualname__)
```
ktbarrett, line 153:
```suggestion
            f"Cannot convert {self!r} to bool"
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,203 @@
...
+        )
+
+    @cache
+    def __int__(self):
+        if self._repr < 2:
+            return self._repr
+        raise ValueError(
+            "Cannot convert {sefl!r} to int".format(type(self).__qualname__)
```
ktbarrett, line 161:
```suggestion
            f"Cannot convert {sefl!r} to int"
```

cocotb/types/logic.py
```diff
@@ -150,15 +150,15 @@ def __bool__(self):
...
+            f"Cannot convert {self!r} to bool"
         )
 
     @cache
     def __int__(self):
         if self._repr < 2:
             return self._repr
         raise ValueError(
-            "Cannot convert {sefl!r} to int".format(type(self).__qualname__)
+            f"Cannot convert {sefl!r} to int"
```
eric-wieser, line 161:
```suggestion
            f"Cannot convert {self!r} to int"
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    @cache
+    def __new__(cls, value=None):
+        # convert to internal representation
+        try:
+            _repr = cls._repr_map[value]
+        except KeyError:
+            raise ValueError(
+                "{!r} is not convertable to a {}".format(value, cls.__qualname__)
```
cmarqu, line 79:
```suggestion
                "{!r} is not convertible to a {}".format(value, cls.__qualname__)
```
?

cocotb/types/logic.py
```diff
@@ -0,0 +1,232 @@
...
+    This is a memory optimization: there will only ever be 4 objects allocated;
+    and also a speed optimization: value equality can be identity equality (as inherited from the `object` base class).
+
+operator `isinstance(other, type(self)`
+    Allows operation between a `Bit` and a `Logic` to return a `Logic`.
+
+strict type equality
+    To use `Logic` or `Bit` in hashable collections we need to define `__hash__` and `__eq__` such that
+    "Hashable objects which compare equal must have the same hash value."
+    Retrieved from https://hynek.me/articles/hashes-and-equality/ on 2021-03-20.
```
cmarqu, line 20:
```suggestion
    (retrieved from https://hynek.me/articles/hashes-and-equality/ on 2021-03-20).
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,232 @@
...
+    The best way to achieve this is to make `Logic` and `Bit` never equal, like `tuple` and `list`.
+    If we instead made them equal in hash and value, they would be substitutable; which we don't want.
+
+@cache
+    Shows extreme performance improvements.
+    Even faster than precomputing the results and storing them in the type.
```
cmarqu, line 26:
```suggestion
    Shows extreme performance improvements,
    even faster than precomputing the results and storing them in the type.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,232 @@
...
+
+class Logic:
+    r"""
+    Model of a 4-value (``0``, ``1``, ``X``, ``Z``) datatype commonly seen in HDLs.
+
+    Modeled after (System)Verilog's 4-value ``logic`` type.
+    VHDL's 9-value ``std_ulogic`` type maps to this type by treating weak values as full strength values
+    and treating "uninitialized" (``U``) and "don't care" (``-``) as "unknown" (``X``).
+
+    Can be converted from :class:`int`, :class:`str`, :class:`bool`, and :class:`~cocotb.types.Bit` using the ``Logic(value)`` syntax.
```
cmarqu, line 39:
```suggestion
    :class:`Logic` can be converted from :class:`int`, :class:`str`, :class:`bool`, and :class:`~cocotb.types.Bit` using the ``Logic(value)`` syntax.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,232 @@
...
+    The list of acceptable values includes ``0``, ``1``, ``True``, ``False``, ``'0'``, ``'1'``, ``'X'``, and ``'Z'``.
+    For a comprehensive list of values that can be converted into :class:`Logic` see :file:`tests/pytest/test_logic.py`.
+
+    Can be converted to :class:`int`, :class:`str`, :class:`bool` using the appropriate constructor syntax.
```
cmarqu, line 43:
```suggestion
    :class:`Logic` can be converted to :class:`int`, :class:`str`, :class:`bool` using the appropriate constructor syntax.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,232 @@
...
+    For example, ``int(Logic(0)) == 0``, ``bool(Logic(1)) is True``, and ``str(Logic('X')) == 'X'``.
+    The :class:`int` and :class:`bool` conversions will raise :exc:`ValueError` if the value is not ``0`` or ``1``.
+
+    :class:`Logic` values are hashable and can be placed in :class:`set`\ s and used as keys in :class:`dict`\ s.
+
+    Supports common logic operations ``&``, ``|``, ``^``, and ``~``.
```
cmarqu, line 49:
```suggestion
    :class:`Logic` supports the common logic operations ``&``, ``|``, ``^``, and ``~``.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+class Bit(Logic):
+    r"""
+    Model of a 2-value (``0``, ``1``) datatype commonly seen in HDLs.
+
+    Modeled after (System)Verilog's 2-value ``bit`` type.
+    VHDL's ``bit`` type maps to this type perfectly.
+
+    Can be converted from :class:`int`, :class:`str`, :class:`bool`, and :class:`~cocotb.types.Logic` using the ``Bit(value)`` syntax.
```
cmarqu, line 188:
```suggestion
    :class:`Bit` can be converted from :class:`int`, :class:`str`, :class:`bool`, and :class:`~cocotb.types.Logic` using the ``Bit(value)`` syntax.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,232 @@
...
+    The list of acceptable values includes ``0``, ``1``, ``True``, ``False``, ``'0'``, ``'1'``.
+    For a comprehensive list of values that can be converted into :class:`Bit` see :file:`tests/pytest/test_logic.py`.
+
+    Can be converted to :class:`int`, :class:`str`, :class:`bool` using the appropriate constructor syntax.
```
cmarqu, line 192:
```suggestion
    :class:`Bit` can be converted to :class:`int`, :class:`str`, :class:`bool` using the appropriate constructor syntax.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,232 @@
...
+    For example, ``int(Bit(0)) == 0``, ``bool(Bit(1)) is True``, and ``str(Bit('1')) == '1'``.
+
+    :class:`Bit` values are hashable and can be placed in :class:`set`\ s and used as keys in :class:`dict`\ s.
+
+    Supports common logic operations ``&``, ``|``, ``^``, and ``~``.
```
cmarqu, line 197:
```suggestion
    :class:`Bit` supports the common logic operations ``&``, ``|``, ``^``, and ``~``.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,232 @@
...
+    Even faster than precomputing the results and storing them in the type.
+    Everything can be cached since there are a (hopefully) limited number of valid values in all subclasses.
+"""
+
+
+class Logic:
+    r"""
+    Model of a 4-value (``0``, ``1``, ``X``, ``Z``) datatype commonly seen in HDLs.
+
+    Modeled after (System)Verilog's 4-value ``logic`` type.
```
cmarqu, line 35:
```suggestion
    This is modeled after (System)Verilog's 4-value ``logic`` type.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,232 @@
...
+        raise ValueError(
+            f"Cannot convert {self!r} to int"
+        )
+
+
+class Bit(Logic):
+    r"""
+    Model of a 2-value (``0``, ``1``) datatype commonly seen in HDLs.
+
+    Modeled after (System)Verilog's 2-value ``bit`` type.
```
cmarqu, line 185:
```suggestion
    This is modeled after (System)Verilog's 2-value ``bit`` type.
```

cmarqu (approved):

cmarqu:
I keep thinking that maybe these should live outside of the main repo or even the cocotb repo.
(I say this because in the far past, I had to manually extract a similar type from myhdl because nothing else was available as a standalone package. This had been for some kind of cosimulation/lab work with pyvisa; cocotb didn't exist back then.)

ktbarrett:
@cmarqu Seeing at these types will be used by the handle class to replace `BinaryValue` eventually, do you think it's a good idea to put core code in another package?

tests/pytest/test_logic.py
ktbarrett: > IMO consistency with design decisions in Python is an argument in itself, especially in the absence of counter-arguments.

Many Python leaders have argued against using `is` for comparsion, and Python even throws a `SyntaxWarning` now if it can detect it. In general having a smaller interface is better than a larger interface. There is also nothing like `is` in Verilog or VHDL.

> As a bonus, `is` is faster than `==`.

Not in this case since they are equivalent. I tested.

cmarqu:
Not sure how to handle this best, it would just be nice. We also don't want extra work for unknown prospective uses of course.
Maybe we can rethink this for 3.0 or 4.0 when we split out GPI because other languages want to use it.

ktbarrett:
Right. The types can work standalone and there is a consumer relationship between cocotb and the types, but for at least the next version, I think it's best they just stay here.

marlonjames (approved):
Any reason for `'` in the docstrings but `"` in the code?

ktbarrett:
@marlonjames I prefer single quotes and default to that. `black` prefers double quotes, but it doesn't fix docstrings.

cmarqu:
We should also add `cocotb/types/logic.py` and `cocotb/types/bit.py` after https://github.com/cocotb/cocotb/blob/1536fd14dddaed797ad5d68858baad1a5f578076/setup.cfg#L35 so that the doctests are also run.

ktbarrett:
Should `Bit` be a proper subtype of `Logic`? Meaning `Bit(0) == Logic(0)` and `hash(Bit(0)) == hash(Logic(0))`. It is mathematically "safe", but I'm not sure if it's a good idea. Right now things are weird because `isinstance(Bit(), Logic) is True` and all relation laws are otherwise followed, but we prevent substitution and purposefully break equality.

An example problem is when a function that can take a `Logic` takes a `Bit` instead (because `isinstance(Bit(), Logic)`). The function expects to be able to use the value as a `Logic`, but attempting to check equality of that value with a `Logic` will not work as expected...

A good overview of this problem is [here](https://medium.com/@floyd.may/liskov-equality-oo-and-math-e1095a919992).

ktbarrett:
Even using a cached `__new__` is not offering extreme performance gains, ~36% slowdown. But it's worth keeping I think. Adding `@lru_cache(typed=True)` to the binary operators is offering <10% slowdown, so I'm not going to worry about that.

cmarqu (approved):

tests/pytest/test_logic.py
ktbarrett: This is no longer an issue, `Bit` is now a proper subtype and the singleton cache is gone, using `is` here won't work.

ktbarrett:
@eric-wieser Also fixed the `type` vs `__class__` issue in this PR. Speak now, or forever hold your peace. I'm merging on Monday.

tests/pytest/test_logic.py
eric-wieser: > Many Python leaders have argued against using is for comparsion

Only for built-in types not designed to use `is` for comparison like str and int.

What was the conclusion here - are values of `bit` singletons like `True` and `False` are?

cocotb/types/logic.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def __int__(self) -> int:
+        if self._repr < 2:
+            return self._repr
+        raise ValueError(f"Cannot convert {self!r} to int")
+
+
+class Bit(Logic):
+    r"""
+    Model of a 2-value (``0``, ``1``) datatype commonly seen in HDLs.
```
eric-wieser, line 188: A remark somewhere that explains that `Bit` is a stricter version of the builtin `bool` that does not allow arithmetic and prints as 0/1 instead of `True/False` would be a good thing.

cocotb/types/logic.py
```diff
@@ -0,0 +1,279 @@
...
+
+        The :class:`int` and :class:`bool` conversions will raise :exc:`ValueError` if the value is not ``0`` or ``1``.
+
+    :class:`Logic` values are hashable and can be placed in :class:`set`\ s and used as keys in :class:`dict`\ s.
+
+    :class:`Logic` supports the common logic operations ``&``, ``|``, ``^``, and ``~``.
+
+    .. code-block:: python3
+
+        >>> def full_adder(a: Logic, b: Logic, carry: Logic) -> (Logic, Logic):
```
eric-wieser, line 57:
```suggestion
        >>> def full_adder(a: Logic, b: Logic, carry: Logic) -> Tuple[Logic, Logic]:
```
What you have isn't how annotation are intended to be used

cocotb/types/logic.py
```diff
@@ -0,0 +1,279 @@
...
+        >>> int(Logic(1))
+        1
+        >>> Bit(Logic("1"))
+        Bit('1')
+
+    .. note::
+
+        The :class:`int` and :class:`bool` conversions will raise :exc:`ValueError` if the value is not ``0`` or ``1``.
+
+    :class:`Logic` values are hashable and can be placed in :class:`set`\ s and used as keys in :class:`dict`\ s.
```
eric-wieser, line 51:
```suggestion
    :class:`Logic` values are immutable and therefore hashable and can be placed in :class:`set`\ s and used as keys in :class:`dict`\ s.
```

cocotb/types/logic.py
```diff
@@ -0,0 +1,279 @@
...
+    .. code-block:: python3
+
+        >>> Logic("X")
+        Logic('X')
+        >>> Logic(True)
+        Logic('1')
+        >>> Logic(1)
+        Logic('1')
+        >>> Logic(Bit(0))
+        Logic('0')
```
eric-wieser, line 33: What's your thought on the canonical spelling (and `__repr__`) being `Logic.X`, Logic.Z`, `Logic._0`, and `Logic._1`?

eric-wieser:
I still think we want a singleton cache for at least `Bit`, for consistency with how `bool` works.

cocotb/types/logic.py
ktbarrett: I'm guessing this would also tie into a singleton cache and keeping the values on the type like enum does? I had that in a previous implementation. That's fine by me.

eric-wieser: > I'm guessing this would also tie into a singleton cache and keeping the values on the type like enum does?

Yes, exactly. I'm not particularly attached to `Logic._0` over `Logic(0)`, but I think `Logic.X` is a clear readability improvement over `Logic("X")`, and should be faster too.

cocotb/types/logic.py
```diff
@@ -0,0 +1,279 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+from typing import Any, Optional
```
ktbarrett, line 4:
```suggestion
from typing import Any, Optional, Tuple
```

cocotb/types/logic.py
ktbarrett: I think there is an issue with where `Bit.X` would work and would return `Logic("X")`.

ktbarrett:
> for behavioural consistency with how bool works.

Do you just mean that you can check equality with `is`? Or something else?

cocotb/types/logic.py
eric-wieser: Oh, good point. That's not too awful to fix:
```python
In [112]: class StaticOnlyProp:
     ...:     def __init__(self, fget):
     ...:         self.fget = fget
     ...:     def __set_name__(self, cls, name):
     ...:         self.__cls = cls
     ...:     def __get__(self, instance, cls):
     ...:         if cls is not self.__cls or instance is not None:
     ...:             raise AttributeError
     ...:         return self.fget()
     ...:
     ...:

In [113]: class Logic:
     ...:     @StaticOnlyProp
     ...:     def X():
     ...:         return "X"
     ...:

In [114]: class Bit(Logic): pass

In [115]: Logic.X
Out[115]: 'X'

In [116]: Logic().X  # this is nonsense
AttributeError

In [117]: Bit.X  # this is nonsense
AttributeError
```

Perhaps that's enough of an argument to leave it for a follow-up; your current spelling is likely something we want to have anyway, and changing a preferred spelling is a much easier thing to do later than removing a spelling.

ktbarrett:
I'm not so sure we care about hashability anymore, it was originally added to make `lru_cache` happy. Having to support hashability prevents defining a "loose" equality `Logic(0) == '0'` like BinaryValue supports. And since I wanted all operators to behave the same (`Logic(0) ^ '0'` shouldn't work if `Logic(0) == '0'` does not work) it follows from that decision causes supporting logic operators with convertible values to not work. 

One solution is to introduce a `match(a, b)` function, suggest using that instead of `==`, and then support implicit conversions from `(str, int, bool)` to the wider types rather than returning `NotImplemented`. It would be nice if Python had two equality operators: "equality" and "match", something like `~==` and `~!=`.

Additionally, I'm not so sure the suggested static attributes or `is` comparison are value adds.

eric-wieser:
> I'm not so sure we care about hashability anymore, it was originally added to make `lru_cache` happy. Having to support hashability prevents defining a "loose" equality `Logic(0) == '0'` like BinaryValue supports.

Does it? Exactly what types of loose equality do you want to support here? As long as your loose equality is transitive (which I'd argue it should be), hashability is still easy.

ktbarrett:
> Does it?

Yes, I see `dut.handle.value == 0` everywhere. It currently only handles comparison with `int`. The new implementation requires additional verbosity for negligible utility.

> Exactly what types of loose equality do you want to support here?

Not just equality, but implicit conversion for all binary operators. This would allow things like the following.

```python
dut.handle.value == "X"
Logic(0) ^ 1
l = LogicArray("01101")
l[:] = "XZ010"  # this is already 
```

Basically it would just try to convert the value and return `NotImplemented` if there was a failure.

```python
def __and__(self, other) -> "Logic":
    if not isinstance(other, type(self)):
        try:
            other = type(self)(other)
        except ValueError:
            return NotImplemented
    # ...
```

> As long as your loose equality is transitive (which I'd argue it should be), hashability is still easy.

It wouldn't (nor could it be) transitive as that would require convertible values to equal each other, e.g. `"1" == 1`. Part of what I was questioning was whether we needed hashability. The current `BinaryValue` isn't hashable, `Array` will not be hashable since it is mutable. etc. Determining what we want is definitely the biggest problem with this effort.

PR #2493, opened 2021-03-27
Deprecate setting handle values with __setattr__ syntax

### LINKED ISSUE #2490: Deprecate __setattr__ overload for setting a handle value
We already have two other ways to accomplish this and the [Zen of Python](https://www.python.org/dev/peps/pep-0020/) states "There should be one-- and preferably only one --obvious way to do it."

```python
handle.value = value
handle <= value
```

The `__setattr__` syntax depends on there being at least one additional level of hierarchy to be able to set a signal; and this can result in unintended behavior with newbies to Python. I have seen this happen at least 3 times.

```python
dut.handle = value  # okay
handle = value  # doh!
```
 
I think this should behavior should be deprecated.

ktbarrett:
Closes #2490. Adds deprecation for `dut.handle = value` syntax.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -315,6 +315,10 @@ class HierarchyObject(RegionObject):
         # then try handles
         sub = self.__get_sub_handle_by_name(name)
         if sub is not None:
+            warnings.warn(
+                "Setting values on handles using the ``dut.handle = value`` syntax is deprecated."
+                "Instead use the ``handle.value = value`` or ``handle <= value`` syntax",
+                DeprecationWarning, stacklevel=2)
             sub.value = value
             return
 
@@ -582,6 +586,10 @@ class NonHierarchyIndexableObject(NonHierarchyObject):
     def __setitem__(self, index, value):
         """Provide transparent assignment to indexed array handles."""
         self[index].value = value
+        warnings.warn(
+            "Setting values on handles using the ``dut.handle[i] = value`` syntax is deprecated."
+            "Instead use the ``handle[i].value = value`` or ``handle[i] <= value`` syntax",
+            DeprecationWarning, stacklevel=2)
 
     def __getitem__(self, index):
         if isinstance(index, slice):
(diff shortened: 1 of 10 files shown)


ktbarrett:
Must also deprecate `__setitem__` on arrayed objects. Need to find out why the remaining code in `__setattr__` is not being hit.

cocotb/handle.py
```diff
@@ -316,6 +316,10 @@ def __setattr__(self, name, value):
         sub = self.__get_sub_handle_by_name(name)
         if sub is not None:
             sub.value = value
+            warnings.warn(
+                "Setting values on handles using the ``dut.handle = value`` syntax is deprecated."
+                "Instead use the ``handle.value = value`` or ``handle <= value`` syntax",
+                DeprecationWarning, stacklevel=2)
```
eric-wieser, line 322: I'd swap the order of these lines, so that if the user promotes warnings to errors, the assignment never happens.

cmarqu (approved):

cocotb/handle.py
```diff
@@ -582,6 +586,10 @@ def __init__(self, handle, path):
     def __setitem__(self, index, value):
         """Provide transparent assignment to indexed array handles."""
         self[index].value = value
+        warnings.warn(
+            "Setting values on handles using the ``dut.handle[i] = value`` syntax is deprecated."
+            "Instead use the ``handle[i].value = value`` or ``handle[i] <= value`` syntax",
+            DeprecationWarning, stacklevel=2)
```
eric-wieser, line 592: This should happen before the write occurs, not after.

documentation/source/newsfragments/2490.removal.rst
```diff
@@ -0,0 +1,2 @@
+Setting values on handles using the ``dut.handle = value`` syntax is deprecated.
+Instead use the ``handle.value = value`` or ``handle <= value`` syntax.
```
eric-wieser, line 2: Can you do a follow-up to mention `handle[i] = val` syntax being deprecated too?

cocotb/handle.py
ktbarrett: Damn, I thought I fixed that...

documentation/source/newsfragments/2490.removal.rst
ktbarrett: Sure.

cocotb/handle.py
eric-wieser: You did, but only for attr not index

ktbarrett: Resolved in #2509.

PR #2501, opened 2021-03-30
Log full object name when releasing VHPI handle

marlonjames:
Add the object name when releasing the handle for better diagnostics.

`VHPI: Releasing VhpiSignalObjHdl handle for test.clock at 0979C810`
vs
`VHPI: Releasing VhpiSignalObjHdl handle at 0979C810`


ktbarrett (commented):
I find it odd that at least some of the lines weren't covered. I suppose they weren't covered before.

ktbarrett (approved):

marlonjames:
Yeah I don't understand the missing line coverage, but the second lines were also missing coverage before.

PR #2502, opened 2021-03-31
Add custom TRACE logging level below DEBUG.

### LINKED ISSUE #2141: GPI: Address DEBUG log spam to cocotb.gpi logger
There are two places that produce lots of DEBUG log statements in the GPI:

https://github.com/cocotb/cocotb/blob/a9cd7ad2625cd8cbfcde3fb5959fb7316b33b4a2/cocotb/share/lib/gpi/GpiCbHdl.cpp#L94-L100

https://github.com/cocotb/cocotb/blob/6eb637ae1ec4110fb3aa10ef04051fe8ef163108/cocotb/share/include/cocotb_utils.h#L57-L73

They are both related to tracing the program flow between GPI and Python.

I'd like to be able to filter out these messages without losing other DEBUG info on the `cocotb.gpi` logger. I would prefer to have them disabled by default with a way to enable them when debugging cocotb issues.

I don't know what the best approach is. New lower log level (#1226)? New distinct logger with a filter?

marlonjames:
Closes #2141.

cocotb produces enormous amounts of `DEBUG` log messages relating to
the execution flow between Python and GPI. These messages bloat the
resulting logs and are not useful in most cases.

Move these messages to a new lower log level `TRACE`, which can be
enabled using `COCOTB_LOG_LEVEL=TRACE`.

TODO:
- [x] newsfragment

### DIFF
--- a/documentation/source/building.rst
+++ b/documentation/source/building.rst
@@ -111,7 +111,9 @@ Cocotb
 .. envvar:: COCOTB_LOG_LEVEL
 
     The default logging level to use. This is set to ``INFO`` unless overridden.
-    Valid values are ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, ``CRITICAL``.
+    Valid values are ``TRACE``, ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, ``CRITICAL``.
+
+    ``TRACE`` is used for internal low-level logging and produces very verbose logs.
 
 .. envvar:: COCOTB_RESOLVE_X
 
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -52,6 +52,10 @@ _FILENAME_CHARS = 20  # noqa
 _LINENO_CHARS   = 4  # noqa
 _FUNCNAME_CHARS = 31  # noqa
 
+# Custom log level
+logging.TRACE = 5
+logging.addLevelName(5, "TRACE")
+
 # Default log level if not overwritten by the user.
 _COCOTB_LOG_LEVEL_DEFAULT = "INFO"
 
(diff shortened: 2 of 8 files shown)


ktbarrett (approved):
I'm fine with it. I feel like there is probably more that can be moved to the TRACE level.

eric-wieser:
Are we sure the right approach here is to create a new log level, and not to log to a different logger?

ktbarrett:
> Does anything log to "cocotb.gpi" with level debug after this change?
Yeah. All the handle iteration and discovery messages.

The way I see it, `DEBUG` is for pertinent debug info and `TRACE` include info about program execution flow. `to_python` and `to_simulator` are things that could probably moved to `TRACE`.

marlonjames:
@eric-wieser We could log to a different logger. Would we add a new envvar to set its level?

@ktbarrett This PR switches `to_python`/`to_simulator` to `LOG_TRACE`.

ktbarrett:
Ahh missed that at first. Adding another logger would likely require adding another cached local log level unless we want all `TRACE`s to go to Python... And of course then there were would `cocotb.gpi` and `cocotb.gpi_trace` or something? That's not a reasonable solution, `TRACE` is.

marlonjames:
@eric-wieser do you have any objections to this PR?

The goal is to reduce the noise in debug logs and show user-relevant information only. Also I want to avoid extra overhead for these (very) frequent log messages by going to Python for each one, and I didn't want to start implementing complex log functionality in the GPI code.

documentation/source/building.rst
```diff
@@ -111,7 +111,7 @@ Cocotb
 .. envvar:: COCOTB_LOG_LEVEL
 
     The default logging level to use. This is set to ``INFO`` unless overridden.
-    Valid values are ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, ``CRITICAL``.
+    Valid values are ``TRACE``, ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, ``CRITICAL``.
```
cmarqu, line 114:
```suggestion
    Valid values are ``TRACE`` (used by cocotb-internal verbose low-level messages), ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, ``CRITICAL``.
```

cmarqu (approved):
Just one small thing I'd add to the docs.

Issue #2503, opened 2021-03-31
Issue awaiting start_soon queued coroutine


marlonjames:
cocotb version: master (0b96ca4a8ac95f385753ebb6e5fb62c56e475747)

`fork()` and `start_soon()` may interact badly in certain circumstances.

The following test fails:

```py3
@cocotb.test()
async def test_fork_after_start_soon(_):
    async def waiter(coro):
        await coro

    async def timer_coro():
        await Timer(1, "ns")

    t_coro = cocotb.scheduler.start_soon(timer_coro())
    w_coro = cocotb.fork(waiter(t_coro))

    start_time = cocotb.utils.get_sim_time(units="ns")

    await Join(w_coro)

    assert cocotb.utils.get_sim_time(units="ns") == start_time + 1
```

The reason is that `fork()` runs the coroutine immediately, and `await coro` calls through to `_trigger_from_unstarted_coro()`, which adds the unstarted coroutine to `_pending_coros`.
But the coroutine was already added by `start_soon()`, so now it's in the queue twice, and will be scheduled twice.
The second time it's scheduled with `None` (as if it was a new coroutine starting) instead of the expected `Timer` trigger.

Changing `waiter` to use `await Join(coro)` passes, because the scheduler doesn't get to `_trigger_from_unstarted_coro`.

marlonjames:
`fork()` isn't even necessary:

```py3
@cocotb.test()
async def test_await_start_soon(_):
    async def coro():
        start_time = cocotb.utils.get_sim_time(units="ns")
        await Timer(1, "ns")
        assert cocotb.utils.get_sim_time(units="ns") == start_time + 1

    coro = cocotb.scheduler.start_soon(coro())

    await coro
```

PR #958, opened 2019-06-21
Delete local references to triggers and coroutines in the event loop

eric-wieser:
By doing this, we are able to process (and ignore) Event` objects scheduled within coroutine destructors, as the destructor now runs before the `while self._pending_events` loop terminates.

Destructor runtime is not super deterministic (and will not work on PyPI), so this might not be a sufficient fix for all cases.

Fixes gh-957

PR #2504, opened 2021-03-31
Fix scheduling issues with start_soon

### LINKED ISSUE #2489: Stack overflow issue with lots of coroutines waiting on queues
I'm running in to an issue where I have a bunch of coroutines that are blocked on queue.get() and for some reason this is causing a stack overflow in the scheduler.  Minimal example:

    async def test_cr(queue):
        item = await queue.get()

    async def run_test(dut):
        k = 1024

        queue_list = [Queue() for k in range(k)]
        cr_list = [cocotb.scheduler.start_soon(test_cr(q)) for q in queue_list]

        await Timer(10, 'ns')

If I run that, I get something like this:

                                         
                                                                                                                         
         0.00ns ERROR    ..coroutine.Task 491.0x7fd8bc463ee0        scheduler.py:523  in _unschedule                     Exception raised by this forked coroutine
    Fatal Python error: _Py_CheckRecursiveCall: Cannot recover from stack overflow.
    Python runtime state: initialized

    Current thread 0x00007fd8bea76d00 (most recent call first):
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/triggers.py", line 396 in __init__
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/triggers.py", line 449 in wait
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/queue.py", line 126 in get
      File "/home/alex/Projects/cocotb/stack_overflow_test/test.py", line 39 in test_cr
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/outcomes.py", line 36 in send
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/decorators.py", line 201 in _advance
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 857 in _schedule
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 719 in add
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 909 in _schedule
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 719 in add
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 909 in _schedule
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 719 in add
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 909 in _schedule
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 719 in add
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 909 in _schedule
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 719 in add
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 909 in _schedule
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 719 in add
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 909 in _schedule
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 719 in add
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 909 in _schedule
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 719 in add
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 909 in _schedule
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 719 in add
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 909 in _schedule
      File "/home/alex/.local/lib/python3.9/site-packages/cocotb/scheduler.py", line 719 in add

If I set k = 512, I get a different stack trace:

     0.00ns ERROR    ..coroutine.Task 491.0x7feba324e7f0        scheduler.py:523  in _unschedule                     Exception raised by this forked coroutine
     0.00ns ERROR    cocotb.regression                         regression.py:408  in _score_test                     Test Failed: run_test_001 (result was RecursionError)
                                                                                                                       File "/home/alex/Projects/cocotb/stack_overflow_test/test.py", line 39, in test_cr
                                                                                                                         item = await queue.get()
                                                                                                                       File "/home/alex/.local/lib/python3.9/site-packages/cocotb/queue.py", line 126, in get
                                                                                                                         await event.wait()
                                                                                                                       File "/home/alex/.local/lib/python3.9/site-packages/cocotb/triggers.py", line 449, in wait
                                                                                                                         return _Event(self)
                                                                                                                       File "/home/alex/.local/lib/python3.9/site-packages/cocotb/triggers.py", line 396, in __init__
                                                                                                                         PythonTrigger.__init__(self)
                                                                                                                     RecursionError: maximum recursion depth exceeded

If I set k = 256, it does not crash.  

I'm running the latest version of cocotb from git.  I ran this with icarus verilog.  I think the max recursion depth is set to 1000.

marlonjames:
Closes #2489 and #2503.

This fixes ~two~ four scheduling issues related to `start_soon()`:
- `_schedule()` was recursively calling itself for each new queued coroutine, so if enough are queued up the stack will overflow.
- `await`ing a coroutine that's been queued by `start_soon()` but hasn't started would hit `_trigger_from_unstarted_coro()`, which would add it to the queue again, causing it to be incorrectly scheduled more than once.
- If a queued coroutine is killed before it starts scheduling, it is still run.
- If the test ends with any queued coroutines, they will be scheduled to run after the test ends.

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -458,6 +458,17 @@ class Scheduler:
                     # happened in gh-957)
                     del coro
 
+                # Handle any newly queued coroutines that need to be scheduled
+                while self._pending_coros:
+                    task = self._pending_coros.pop(0)
+                    if _debug:
+                        self.log.debug("Scheduling queued coroutine %s" % (task._coro.__qualname__))
+                    self._schedule(task)
+                    if _debug:
+                        self.log.debug("Scheduled queued coroutine %s" % (task._coro.__qualname__))
+
+                    del task
+
                 # Schedule may have queued up some events so we'll burn through those
                 while self._pending_events:
                     if _debug:
@@ -487,6 +498,12 @@ class Scheduler:
 
     def _unschedule(self, coro):
         """Unschedule a coroutine.  Unprime any pending triggers"""
+        if coro in self._pending_coros:
+            assert not coro.has_started()
+            self._pending_coros.remove(coro)
+            # Close coroutine so there is no RuntimeWarning that it was never awaited
+            coro.close()
+            return
 
         # Unprime the trigger this coroutine is waiting on
         trigger = coro._trigger
@@ -580,7 +597,9 @@ class Scheduler:
 
     def _queue(self, coroutine):
         """Queue a coroutine for execution"""
-        self._pending_coros.append(coroutine)
+        # Don't queue the same coroutine more than once (gh-2503)
+        if coroutine not in self._pending_coros:
+            self._pending_coros.append(coroutine)
 
     def queue_function(self, coro):
         """
@@ -904,10 +923,6 @@ class Scheduler:
                         self._pending_threads.remove(ext)
                         self._pending_events.append(ext.event)
 
-            # Handle any newly queued coroutines that need to be scheduled
-            while self._pending_coros:
-                self.add(self._pending_coros.pop(0))
-
     def finish_test(self, exc):
         """
         .. deprecated:: 1.5
@@ -963,6 +978,10 @@ class Scheduler:
                 if _debug:
                     self.log.debug("Killing %s" % str(coro))
                 coro.kill()
+
+        # kill any queued coroutines
+        for task in self._pending_coros:
+            task.kill()
 
         if self._main_thread is not threading.current_thread():
             raise Exception("Cleanup() called outside of the main thread")
--- a/tests/test_cases/test_cocotb/test_scheduler.py
+++ b/tests/test_cases/test_cocotb/test_scheduler.py
@@ -175,6 +175,21 @@ async def test_stack_overflow(dut):
 
 
 @cocotb.test()
+async def test_stack_overflow_pending_coros(dut):
+    """
+    Test against stack overflow when queueing many pending coroutines
+    before yielding to scheduler.
+    """
+    # gh-2489
+    async def simple_coroutine():
+        await Timer(10, "step")
+
+    coros = [cocotb.scheduler.start_soon(simple_coroutine()) for _ in range(1024)]
+
+    await Combine(*coros)
+
+
+@cocotb.test()
 async def test_kill_coroutine_waiting_on_the_same_trigger(dut):
     # gh-1348
     # NOTE: this test depends on scheduling priority.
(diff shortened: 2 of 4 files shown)


ktbarrett (commented):
The change looks good and makes sense. The tests using `get_sim_time` is not a good idea seeing as none of the tests should ever consume simulation time, so it will report false positives. I might use `None` vs `not None` to ensure proper scheduling behavior.

Of course this needs a newsfragment as well. I wouldn't worry about the coverage issues right now, they will all get hit once we run a test with debugging on.

marlonjames:
> The tests using `get_sim_time` is not a good idea seeing as none of the tests should ever consume simulation time, so it will report false positives.

I'm not sure what you mean here.

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -426,3 +441,27 @@ async def example():
...
+@cocotb.test()
+async def test_start_soon_scheduling(_):
+    async def coro():
+        nonlocal start_time
+        assert cocotb.utils.get_sim_time() == start_time
+
+    start_time = cocotb.utils.get_sim_time()
+    cocotb.scheduler.start_soon(coro())
+    await NullTrigger()
+    assert cocotb.utils.get_sim_time() == start_time
```
ktbarrett, line 455: This is the test I was talking about. Since there is no passage of time anywhere, `get_sim_time` isn't a good value. If we are just testing to ensure the task hasn't been started until after we yield control, there is an easier way.
```suggestion
    async def coro():
        pass

    task = cocotb.scheduler.start_soon(coro())
    assert not task._started
    await NullTrigger()
    assert task._started
```

marlonjames: I wanted to test that `coro` started in the same time step.

ktbarrett: What I'm saying is that it's impossible not to, even if there was a bug, since there are no triggers which advance simulation time. There is no chance for incorrect behavior to occur.

ktbarrett: Also, time step might not be good enough, we don't want it to go back to the simulator at all before servicing. `get_sim_time` will be the same after a `ReadWrite`, for example.

marlonjames: Figured out something by wrapping `_react` and checking before returning to simulator.

cocotb/scheduler.py
```diff
@@ -458,6 +458,17 @@ def _event_loop(self, trigger):
...
+                # Handle any newly queued coroutines that need to be scheduled
+                while self._pending_coros:
+                    task = self._pending_coros.pop(0)
+                    if _debug:
+                        self.log.debug("Scheduling queued coroutine %s" % (task._coro.__qualname__))
+                    self._schedule(task)
+                    if _debug:
+                        self.log.debug("Scheduled queued coroutine %s" % (task._coro.__qualname__))
+
+                    del task
```
ktbarrett, line 470: Are we sure this is needed?

ktbarrett (approved):
Great work tracking this issue down and finding a fix. I like the solution wrapping `react`, that should be foolproof.

cocotb/scheduler.py
marlonjames: I'm not sure, I just followed the pattern above it from #958.

ktbarrett: Alright. It can stay.

PR #2494, opened 2021-03-27
Add support for managing coroutines in with_timeout

alexforencich:
This PR adds support to with_timeout for managing coroutines.  With this change, an `async def` coroutine passed to with_timeout will be forked and awaited.  When the coroutine terminates, the return value will be forwarded.  If the timeout expires, then the coroutine will be killed.

PR #2506, opened 2021-04-01
Change with_timeout behavior

ktbarrett:
Potential follow-on to #2494. Changes the behavior of `with_timeout` so that coroutines that are passed in unstarted are both started and killed (if timeout occurs) by `with_timeout`, only coroutines that have been previous `fork`ed will remain alive if timeout occurs. Also adds more descriptive documentation to `with_timeout`. Pinging @alexforencich.

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -890,7 +890,7 @@ class ClockCycles(Waitable):
 
 
 async def with_timeout(trigger, timeout_time, timeout_unit="step"):
-    """
+    r"""
     Waits on triggers and coroutines, throws an exception if it waits longer than the given time.
 
     When used with a coroutine, the coroutine will be forked and awaited, and the return value forwarded when the coroutine finishes.  If the timeout expires, the coroutine will be killed.
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -919,7 +919,7 @@ async def with_timeout(trigger, timeout_time, timeout_unit="step"):
     .. versionadded:: 1.3
 
     .. versionchanged:: 1.6
-        Support passing :term:`python:coroutine`\\ s.
+        Support passing :term:`python:coroutine`\ s.
 
     .. deprecated:: 1.5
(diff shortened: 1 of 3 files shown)


tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -194,6 +208,21 @@ async def test_timeout_func_pass(dut):
...
+    async def example():
+        await Timer(10, 'ns')
+        return 1
+    coro = example()
+    try:
+        await cocotb.triggers.with_timeout(coro, timeout_time=1, timeout_unit='ns')
+    except cocotb.result.SimTimeoutError:
+        pass
+    assert not coro._finished
+    assert await coro == 1
```
marlonjames, line 223: Isn't this testing that it wasn't killed?
```suggestion
    assert coro._finished
    assert await coro is None
```

cocotb/triggers.py
```diff
@@ -917,7 +937,22 @@ async def with_timeout(trigger, timeout_time, timeout_unit="step"):
...
-   """
+
+    .. versionchanged:: 1.6
+        Support passing :term:`python:coroutine`\ s.
+
+    .. versionchanged:: 2.0
+        :class:`~cocotb.coroutine`\ s that are passed unstarted are now killed if there is a timeout.
+    """
+
+    if inspect.iscoroutine(trigger) or isinstance(trigger, cocotb.decorators.coroutine) and not trigger._started:
```
marlonjames, line 948: I think we need `RunningCoroutine` here, otherwise it just skips to the normal behavior.
```suggestion
    if inspect.iscoroutine(trigger) or isinstance(trigger, cocotb.decorators.RunningCoroutine) and not trigger._started:
```

tests/test_cases/test_cocotb/test_timing_triggers.py
ktbarrett:
```suggestion
    assert coro._finished
    assert await coro == 1
```

ktbarrett:
I'll leave this change in behavior up to whomever works on the scheduler rewrite.

PR #2507, opened 2021-04-01
Remove RunningTest class

ktbarrett:
After the regression manager refactor, which made putting `test` decorator arguments on the `RunningTest` unnecessary, the only use of `RunningTest` was to hold error messages caught by the special handler. One could never access the `RunningTest` object without a private attribute access, and the error messages were never used internally, so the handler and error message are dead code. Considering that, the entire `RunningTest` class is pointless and can be removed.

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -518,11 +518,11 @@ class Scheduler:
             except (TestComplete, AssertionError) as e:
                 coro.log.info("Test stopped by this forked coroutine")
                 e = remove_traceback_frames(e, ['_unschedule', 'get'])
-                self._test.abort(e)
+                self._abort_test(e)
             except Exception as e:
                 coro.log.error("Exception raised by this forked coroutine")
                 e = remove_traceback_frames(e, ['_unschedule', 'get'])
-                self._test.abort(e)
+                self._abort_test(e)
 
     def _schedule_write(self, handle, write_func, *args):
         """ Queue `write_func` to be called on the next ReadWrite trigger. """
@@ -917,8 +917,26 @@ class Scheduler:
         return self._finish_test(exc)
 
     def _finish_test(self, exc):
-        self._test.abort(exc)
+        self._abort_test(exc)
         self._check_termination()
+
+    def _abort_test(self, exc):
+        """Force this test to end early, without executing any cleanup.
+
+        This happens when a background task fails, and is consistent with
+        how the behavior has always been. In future, we may want to behave
+        more gracefully to allow the test body to clean up.
+
+        `exc` is the exception that the test should report as its reason for
+        aborting.
+        """
+        if self._test._outcome is not None:
+            raise InternalError("Outcome already has a value, but is being set again.")
+        outcome = outcomes.Error(exc)
+        if _debug:
+            self._test.log.debug(f"outcome forced to {outcome}")
+        self._test._outcome = outcome
+        self._unschedule(self._test)
 
     def finish_scheduler(self, exc):
         """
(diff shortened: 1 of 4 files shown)


eric-wieser:
> One could never access the RunningTest object without a private attribute access, and the error messages were never used internally,

Aren't they emitted in the xunit XML?

eric-wieser:
I think it would make sense to split this PR in two - removing hooks seems completely orthogonal to removing `RunningTest` to me.

ktbarrett:
>Aren't they emitted in the xunit XML?

Unless there is some dark magic happening, I can't see how. I had thought the same thing. I guess it was unintentionally removed at some point. No one has come complaining though.

I can split them.

eric-wieser:
Ah, you removed that behavior in #1647.

eric-wieser:
It happens here:

https://github.com/cocotb/cocotb/blame/master/cocotb/regression.py#L483-L484

ktbarrett:
That adds the handler, but the handler just appends to `error_messages` on the `RunningTest` object. I don't see how it makes its way to the xunit reporter?

ktbarrett:
That adds the `RunningTest`'s handler to the logger, but that handler just appends to `error_messages` AFAICT. I don't see how it gets to the XUnit reporter?

eric-wieser:
Ah, looks like I removed it in 2a893070a901baf41a3315fff2c79ddb0c69f8f1

ktbarrett:
Split off the removal of hooks into #2508.

marlonjames:
Removing running task removes the special `__name__`/`__qualname__` for tests.
This was discussed here:
https://github.com/cocotb/cocotb/pull/1805/files#r424823502

I think it's useful to distinguish the current test from other tasks. I would override these in `RegressionManager._init_test`.

ktbarrett:
@marlonjames I put it in the `test` decorator `__call__` overload. Seems to make a little more sense there.

marlonjames (approved):

cocotb/scheduler.py
```diff
@@ -917,9 +917,27 @@ def finish_test(self, exc):
...
+        """Force this test to end early, without executing any cleanup.
+
+        This happens when a background task fails, and is consistent with
+        how the behavior has always been. In future, we may want to behave
+        more gracefully to allow the test body to clean up.
+
+        `exc` is the exception that the test should report as its reason for
+        aborting.
+        """
+        if self._test._outcome is not None:
```
ktbarrett, line 933:
```suggestion
        if self._test._outcome is not None:  # pragma: no cover
```

cocotb/scheduler.py
```diff
@@ -917,9 +917,27 @@ def finish_test(self, exc):
         return self._finish_test(exc)
 
     def _finish_test(self, exc):
-        self._test.abort(exc)
+        self._abort_test(exc)
```
ktbarrett, line 920: Since removing the system tasks there is no way to finish a test from `_sim_event`, which is the only caller here (and the only caller of this function that is valid). The line wasn't previously covered. I'm going to excuse it, I don't think it shuld be pragma'd out.

cocotb/scheduler.py
```diff
@@ -917,9 +917,27 @@ def finish_test(self, exc):
...
+        more gracefully to allow the test body to clean up.
+
+        `exc` is the exception that the test should report as its reason for
+        aborting.
+        """
+        if self._test._outcome is not None:
+            raise InternalError("Outcome already has a value, but is being set again.")
+        outcome = outcomes.Error(exc)
+        if _debug:
+            self._test.log.debug(f"outcome forced to {outcome}")
```
ktbarrett, line 937: Will be caught when scheduler debugging is turned on.

PR #2509, opened 2021-04-02
Fixup #2493

ktbarrett:
Fixes issues noted in #2493 after the merge occurred. Added a newsfragment stating that `__setitem__` syntax was similar being deprecated. And moved the warning to before the setting, so when warnings are error the value setting never occurs.


eric-wieser (approved):

ktbarrett:
Fixing broken tests and missing space in deprecation message.

PR #2510, opened 2021-04-03
Add Range and Array modeling types

ktbarrett:
Follow on to #2485. xref #2059. Adds a `Range` and `Array` type.

The `Range` type is much like Python's `range` type, but has an inclusive right bound like is seen in both Verilog and VHDL. It support's VHDL's `to` and `downto` directions, Verilog's automatic directionality by not specifying a direction, and VHDL's null ranges.

The `Array` type is a heterogenous fixed-length sequence type. It uses an initial value or a range to determine it's size at construction and it's size is fixed. It supports must of the same operations as a list that don't mutate the length. Uses a `Range` to describe the indexing scheme, which is arbitrary, like is seen in both Verilog and VHDL. Some indexing and slicing behavior had to change compared to a list to make something that behaved intuitively.

### DIFF
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -69,12 +69,3 @@ if sys.version_info < (3, 9):
         return lru_cache(maxsize=None)(f)
 else:
     from functools import cache  # noqa: F401
-
-
-# Emulates Python 3.8's functools.cached_property decorator.
-# Uses lru_cache with LRU functionality, otherwise the cache could hold references to
-# dead objects and grow infinitely.
-if sys.version_info >= (3, 8):
-    from functools import cached_property
-else:
-    cached_property = property
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -78,4 +78,4 @@ if sys.version_info >= (3, 8):
     from functools import cached_property
 else:
     def cached_property(method):
-        return property(lru_cache(method))
+        return property(lru_cache()(method))
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -77,5 +77,4 @@ else:
 if sys.version_info >= (3, 8):
     from functools import cached_property
 else:
-    def cached_property(method):
-        return property(lru_cache()(method))
+    cached_property = property
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -29,6 +29,7 @@ These are for internal use - users should use a third party library like `six`
 if they want to use these shims in their own code
 """
 import sys
+from functools import lru_cache
 
 
 # backport of Python 3.7's contextlib.nullcontext
@@ -60,3 +61,21 @@ if sys.version_info[:2] >= (3, 7):
 else:
     import collections
     insertion_ordered_dict = collections.OrderedDict
+
+
+# backport of Python 3.9's functools.cache decorator
+if sys.version_info < (3, 9):
+    def cache(f):
+        return lru_cache(maxsize=None)(f)
+else:
+    from functools import cache  # noqa: F401
+
+
+# Emulates Python 3.8's functools.cached_property decorator.
+# Uses lru_cache with LRU functionality, otherwise the cache could hold references to
+# dead objects and grow infinitely.
+if sys.version_info >= (3, 8):
+    from functools import cached_property
+else:
+    def cached_property(method):
+        return property(lru_cache(method))
--- a/cocotb/_py_compat.py
+++ b/cocotb/_py_compat.py
@@ -29,7 +29,6 @@ These are for internal use - users should use a third party library like `six`
 if they want to use these shims in their own code
 """
 import sys
-from functools import lru_cache
 
 
 # backport of Python 3.7's contextlib.nullcontext
@@ -61,11 +60,3 @@ if sys.version_info[:2] >= (3, 7):
 else:
     import collections
     insertion_ordered_dict = collections.OrderedDict
-
-
-# backport of Python 3.9's functools.cache decorator
-if sys.version_info < (3, 9):
-    def cache(f):
-        return lru_cache(maxsize=None)(f)
-else:
-    from functools import cache  # noqa: F401
--- a/cocotb/types/__init__.py
+++ b/cocotb/types/__init__.py
@@ -3,3 +3,4 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from .logic import Logic, Bit  # noqa: F401
 from .range import Range  # noqa: F401
+from .array import Array  # noqa: F401
--- a/cocotb/types/__init__.py
+++ b/cocotb/types/__init__.py
@@ -16,14 +16,27 @@ def concat(a: Array, b: Array) -> Array:
     Raises:
         TypeError: when the arguments do not support concatenation in the given order.
     """
-    if hasattr(a, '__concat__'):
-        res = a.__concat__(b)
+    a_concat = getattr(a.__class__, "__concat__", None)
+    a_rconcat = getattr(a.__class__, "__rconcat__", None)
+    b_rconcat = getattr(b.__class__, "__rconcat__", None)
+
+    if isinstance(b, a.__class__) and a_rconcat != b_rconcat:
+        # 'b' is a subclass of 'a' with a more specific implementation of 'concat(a, b)'
+        call_order = ((b, b_rconcat, a), (a, a_concat, b))
+    elif a.__class__ != b.__class__:
+        # normal call order
+        call_order = ((a, a_concat, b), (b, b_rconcat, a))
+    else:
+        # types are the same, we expect implementation of 'concat(a, b)' to be in 'a.__concat__'
+        call_order = ((a, a_concat, b),)
+
+    for lhs, method, rhs in call_order:
+        if method is None:
+            continue
+        res = method(lhs, rhs)
         if res is not NotImplemented:
             return res
-    if hasattr(b, '__rconcat__'):
-        res = b.__rconcat__(a)
-        if res is not NotImplemented:
-            return res
+
     raise TypeError(
         "cannot concatenate {!r} with {!r}".format(
             a.__class__.__qualname__, b.__class__.__qualname__
--- a/cocotb/types/__init__.py
+++ b/cocotb/types/__init__.py
@@ -2,3 +2,4 @@
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
 from .logic import Logic, Bit  # noqa: F401
+from .range import Range  # noqa: F401
--- a/cocotb/types/__init__.py
+++ b/cocotb/types/__init__.py
@@ -4,3 +4,28 @@
 from .logic import Logic, Bit  # noqa: F401
 from .range import Range  # noqa: F401
 from .array import Array  # noqa: F401
+
+
+def concat(a: Array, b: Array) -> Array:
+    """
+    Create a new array that is the concatenation of one array with another.
+
+    Uses the :meth:`__concat__` or :meth:`__rconcat__` special methods to dispatch to a particular implementation,
+    exactly like other binary operations in Python.
+
+    Raises:
+        TypeError: when *other* is an object of dissimilar type.
+    """
+    if hasattr(a, '__concat__'):
+        res = a.__concat__(b)
+        if res is not NotImplemented:
+            return res
+    if hasattr(b, '__rconcat__'):
+        res = b.__rconcat__(a)
+        if res is not NotImplemented:
+            return res
+    raise TypeError(
+        "can't concat value of type {!r} with value of type {!r}".format(
+            a.__class__.__qualname__, b.__class__.__qualname__
+        )
+    )
--- a/cocotb/types/__init__.py
+++ b/cocotb/types/__init__.py
@@ -16,14 +16,16 @@ def concat(a: Array, b: Array) -> Array:
     Raises:
         TypeError: when the arguments do not support concatenation in the given order.
     """
-    a_concat = getattr(a.__class__, "__concat__", None)
-    a_rconcat = getattr(a.__class__, "__rconcat__", None)
-    b_rconcat = getattr(b.__class__, "__rconcat__", None)
+    type_a = type(a)
+    type_b = type(b)
+    a_concat = getattr(type_a, "__concat__", None)
+    a_rconcat = getattr(type_a, "__rconcat__", None)
+    b_rconcat = getattr(type_b, "__rconcat__", None)
 
-    if isinstance(b, a.__class__) and a_rconcat != b_rconcat:
+    if type_a is not type_b and issubclass(type_b, type_a) and a_rconcat != b_rconcat:
         # 'b' is a subclass of 'a' with a more specific implementation of 'concat(a, b)'
         call_order = ((b, b_rconcat, a), (a, a_concat, b))
-    elif a.__class__ != b.__class__:
+    elif type_a is not type_b:
         # normal call order
         call_order = ((a, a_concat, b), (b, b_rconcat, a))
     else:
@@ -39,6 +41,6 @@ def concat(a: Array, b: Array) -> Array:
 
     raise TypeError(
         "cannot concatenate {!r} with {!r}".format(
-            a.__class__.__qualname__, b.__class__.__qualname__
+            type_a.__qualname__, type_b.__qualname__
         )
     )
--- a/cocotb/types/__init__.py
+++ b/cocotb/types/__init__.py
@@ -16,11 +16,12 @@ def concat(a: Array, b: Array) -> Array:
     Raises:
         TypeError: when the arguments do not support concatenation in the given order.
     """
+    MISSING = object()
     type_a = type(a)
     type_b = type(b)
-    a_concat = getattr(type_a, "__concat__", None)
-    a_rconcat = getattr(type_a, "__rconcat__", None)
-    b_rconcat = getattr(type_b, "__rconcat__", None)
+    a_concat = getattr(type_a, "__concat__", MISSING)
+    a_rconcat = getattr(type_a, "__rconcat__", MISSING)
+    b_rconcat = getattr(type_b, "__rconcat__", MISSING)
 
     if type_a is not type_b and issubclass(type_b, type_a) and a_rconcat != b_rconcat:
         # 'b' is a subclass of 'a' with a more specific implementation of 'concat(a, b)'
@@ -33,7 +34,7 @@ def concat(a: Array, b: Array) -> Array:
         call_order = ((a, a_concat, b),)
 
     for lhs, method, rhs in call_order:
-        if method is None:
+        if method is MISSING:
             continue
         res = method(lhs, rhs)
         if res is not NotImplemented:
--- a/cocotb/types/__init__.py
+++ b/cocotb/types/__init__.py
@@ -14,7 +14,7 @@ def concat(a: Array, b: Array) -> Array:
     exactly like other binary operations in Python.
 
     Raises:
-        TypeError: when *other* is an object of dissimilar type.
+        TypeError: when the arguments do not support concatenation in the given order.
     """
     if hasattr(a, '__concat__'):
         res = a.__concat__(b)
@@ -25,7 +25,7 @@ def concat(a: Array, b: Array) -> Array:
         if res is not NotImplemented:
             return res
     raise TypeError(
-        "can't concat value of type {!r} with value of type {!r}".format(
+        "cannot concatenate {!r} with {!r}".format(
             a.__class__.__qualname__, b.__class__.__qualname__
         )
     )
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -212,10 +212,16 @@ class Array(Sequence):
         return item in self._value
 
     def __eq__(self, other: Any) -> bool:
+        sentinel = object()
         try:
-            return all(a == b for a, b in zip_longest(self, other, fillvalue=object()))
+            it = zip_longest(self, other, fillvalue=sentinel)
         except TypeError:
             return NotImplemented
+        for a, b in it:
+            # sentinel check MUST come before element equality check
+            if b == sentinel or a != b:
+                return False
+        return True
 
     @overload
     def __getitem__(self, item: int) -> Any:
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -1,3 +1,6 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
 from typing import Optional, Any, Iterable, Iterator, overload
 from collections.abc import Sequence
 from .range import Range
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -1,7 +1,7 @@
 from typing import Optional, Any, Iterable, Iterator, overload
 from collections.abc import Sequence
 from .range import Range
-from functools import cached_property
+from cocotb._py_compat import cached_property
 from itertools import zip_longest
 from sys import maxsize
 
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -5,7 +5,6 @@ from typing import Optional, Any, Iterable, Iterator, overload
 from collections.abc import Sequence
 from .range import Range
 from sys import maxsize
-from itertools import chain
 
 
 class Array(Sequence):
@@ -292,7 +291,7 @@ class Array(Sequence):
                     other.__class__.__qualname__
                 )
             )
-        return self.__class__(chain(self, other))
+        return self.__class__(self._value + other._value)
 
     def index(
         self, value: Any, start: Optional[int] = None, stop: Optional[int] = None
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -204,7 +204,11 @@ class Array(Sequence):
                 raise IndexError("slice direction does not match array direction")
             value = list(value)
             if len(value) != (stop_i - start_i + 1):
-                raise ValueError("value is not the same length as the slice")
+                raise ValueError(
+                    "value  of length {!r} will not fit in slice [{}:{}]".format(
+                        len(value), start, stop
+                    )
+                )
             self._value[start_i : stop_i + 1] = value
         else:
             raise TypeError(
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -289,14 +289,14 @@ class Array(Sequence):
         return "{}({!r}, {!r})".format(type(self).__name__, self._value, self._range)
 
     def __concat__(self, other: "Array") -> "Array":
-        if not isinstance(other, self.__class__):
+        if not isinstance(other, type(self)):
             return NotImplemented
-        return self.__class__(self._value + other._value)
+        return type(self)(self._value + other._value)
 
     def __rconcat__(self, other: "Array") -> "Array":
-        if not isinstance(other, self.__class__):
+        if not isinstance(other, type(self)):
             return NotImplemented
-        return self.__class__(other._value + self._value)
+        return type(self)(other._value + self._value)
 
     def index(
         self, value: Any, start: Optional[int] = None, stop: Optional[int] = None
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -302,7 +302,12 @@ class Array(Sequence):
     def index(
         self, value: Any, start: Optional[int] = None, stop: Optional[int] = None
     ) -> int:
-        """Return index of first occurrence of *value*."""
+        """
+        Return index of first occurrence of *value*.
+
+        Raises :exc:`ValueError` if the value is not found.
+        Search only within *start* and *stop* if given.
+        """
         if start is not None:
             start = self._translate_index(start)
         else:
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -120,11 +120,15 @@ class Array(Sequence):
         self, value: Optional[Iterable[Any]] = None, range: Optional[Range] = None
     ):
         if value is not None and range is None:
-            self._value = self._construct_value(value)
-            self._range = self._construct_range(Range(0, "to", len(self._value) - 1))
+            constructor = self._construct_element
+            self._value = [constructor(v) for v in value]
+            self._range = Range(0, "to", len(self._value) - 1)
         elif value is not None and range is not None:
-            self._value = self._construct_value(value)
-            self._range = self._construct_range(range)
+            if not isinstance(range, Range):
+                raise TypeError("range argument must be of type 'Range'")
+            constructor = self._construct_element
+            self._value = [constructor(v) for v in value]
+            self._range = range
             if len(self._value) != len(self._range):
                 raise ValueError(
                     "init value of length {!r} does not fit in {!r}".format(
@@ -132,39 +136,23 @@ class Array(Sequence):
                     )
                 )
         elif value is None and range is not None:
-            self._value = self._construct_value(None for _ in range)
-            self._range = self._construct_range(range)
+            if not isinstance(range, Range):
+                raise TypeError("range argument must be of type 'Range'")
+            constructor = self._construct_element
+            self._value = [constructor(None) for _ in range]
+            self._range = range
         else:
             raise TypeError("must pass a value, range, or both")
-
-    @staticmethod
-    def _construct_value(value: Iterable[Any]) -> List[Any]:
-        """
-        Construct the value portion of the array.
-
-        For overriding by subclasses.
-        Used by constructor when values are given
-        and by __setitem__ when setting a slice to an iterable.
-
-        Args:
-            value: Any iterable of values that are constructable into elements using :meth:`_construct_elements`
-
-        Returns:
-            A :class:`list` of elements.
-
-        Raises:
-            TypeError: When the type isn't supported.
-            ValueError: When the iterable's values prevent construction into a list of elements.
-        """
-        return list(value)
 
     @staticmethod
     def _construct_element(elem: Any) -> Any:
         """
         Construct a single element of an array.
 
-        For overriding by subclasses.
-        Used by __setitem__ when setting a single element.
+        Will be fed elements of the *value* iterable if given to the constructor,
+        :data:`None` when no *value* is given to the constuctor,
+        the *value* in __setitem__ when assigning to a single index,
+        and elements of the *value* iterable when assigning to a slice.
 
         Args:
             elem: Any object that the implementation can turn into an array element.
@@ -177,29 +165,6 @@ class Array(Sequence):
             ValueError: When the value prevents construction into an element.
         """
         return elem
-
-    @staticmethod
-    def _construct_range(rng: Any) -> Range:
-        """
-        Construct the range portion of the array.
-
-        For overriding by subclasses.
-        Used by the constructor when setting the range
-        and by __setitem__ when creating a slice.
-
-        Args:
-            rng: Any object that the implementation can turn into a :class:`Range` object.
-
-        Returns:
-            The :class:`Range` object describing the array's indexing scheme.
-
-        Raises:
-            TypeError: When type isn't supported.
-            ValueError: When the value prevents construction into a range.
-        """
-        if isinstance(rng, Range):
-            return rng
-        raise TypeError("range argument must be of type 'Range'")
 
     @property
     def left(self) -> int:
@@ -301,7 +266,8 @@ class Array(Sequence):
                         start, stop, self.left, self.right
                     )
                 )
-            value = self._construct_value(value)
+            constructor = self._construct_element
+            value = [constructor(v) for v in value]
             if len(value) != (stop_i - start_i + 1):
                 raise ValueError(
                     "value of length {!r} will not fit in slice [{}:{}]".format(
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -25,19 +25,19 @@ class Array(Sequence):
     .. code-block:: python3
 
         >>> Array("1234")  # the 0-based range `(0, len(value)-1)` is inferred
-        Array(['1', '2', '3', '4'], Range(0, 3))
+        Array(['1', '2', '3', '4'], Range(0, 'to', 3))
 
         >>> Array(range=Range(0, 'downto', -3))  # the initial values are `None`
-        Array([None, None, None, None], Range(0, -3))
+        Array([None, None, None, None], Range(0, 'downto', -3))
 
-        >>> Array([1, True, object(), 'example'], Range(-2, 1))  # initial value and range lengths must be equal
-        Array([1, True, <object object at 0x7f4cff7b5570>, 'example'], Range(-2, 'to', 1))
+        >>> Array([1, True, None, 'example'], Range(-2, 1))  # initial value and range lengths must be equal
+        Array([1, True, None, 'example'], Range(-2, 'to', 1))
 
     Arrays also support "null" ranges; "null" arrays have zero length and cannot be indexed.
 
     .. code-block:: python3
 
-        >>> Array(range=Range(0, 'to', 1))
+        >>> Array(range=Range(1, 'to', 0))
         Array([], Range(1, 'to', 0))
 
     Indexing and slicing is very similar to :class:`list`\ s, but it uses the indexing scheme specified.
@@ -55,7 +55,7 @@ class Array(Sequence):
         >>> ''.join(a)
         '12ba43cd'
 
-        >>> b = Array("1234", Range(0, 'downto', -3))
+        >>> b = Array("1234", Range(0, -3))
         >>> b[-2]
         '3'
         >>> b[-1:]
@@ -144,7 +144,7 @@ class Array(Sequence):
     def __contains__(self, item: Any) -> bool:
         return item in self._value
 
-    def __eq__(self, other: object) -> bool:
+    def __eq__(self, other: Any) -> bool:
         try:
             return all(a == b for a, b in zip_longest(self, other, fillvalue=object()))
         except TypeError:
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -236,8 +236,8 @@ class Array(Sequence):
             idx = self._translate_index(item)
             return self._value[idx]
         elif isinstance(item, slice):
-            start = item.start or self.left
-            stop = item.stop or self.right
+            start = item.start if item.start is not None else self.left
+            stop = item.stop if item.stop is not None else self.right
             if item.step is not None:
                 raise IndexError("do not specify step")
             start_i = self._translate_index(start)
@@ -269,8 +269,8 @@ class Array(Sequence):
             value = self._construct_element(value)
             self._value[idx] = value
         elif isinstance(item, slice):
-            start = item.start or self.left
-            stop = item.stop or self.right
+            start = item.start if item.start is not None else self.left
+            stop = item.stop if item.stop is not None else self.right
             if item.step is not None:
                 raise IndexError("do not specify step")
             start_i = self._translate_index(start)
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -83,6 +83,18 @@ class Array(Sequence):
         >>> a == b
         True
 
+    You can change the bounds of an array by setting the :attr:`range` to a new value.
+    The new bounds must be the same length of the array.
+
+    .. code-block:: python3
+
+        >>> a = Array("1234")
+        >>> a.range
+        Range(0, 'to', 3)
+        >>> a.range = Range(3, 'downto', 0)
+        >>> a.range
+        Range(3, 'downto', 0)
+
     Arrays support the methods and semantics defined by :class:`collections.abc.Sequence`.
 
     .. code-block:: python
@@ -185,6 +197,15 @@ class Array(Sequence):
     def range(self) -> Range:
         """:class:`Range` of the indexes of the array."""
         return self._range
+
+    @range.setter
+    def range(self, new_range: Range) -> None:
+        """Sets a new indexing scheme on the array, must be the same size"""
+        if not isinstance(new_range, Range):
+            raise TypeError("range argument must be of type 'Range'")
+        if len(new_range) != len(self):
+            raise ValueError(f"{new_range!r} not the same length as old range ({self._range!r}).")
+        self._range = new_range
 
     def __len__(self) -> int:
         return len(self.range)
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -4,7 +4,6 @@
 from typing import Optional, Any, Iterable, Iterator, overload, List
 from collections.abc import Sequence
 from .range import Range
-from cocotb._py_compat import cached_property
 from itertools import zip_longest
 from sys import maxsize
 
@@ -88,7 +87,6 @@ class Array(Sequence):
     __slots__ = (
         "_value",
         "_range",
-        "__dict__",  # necessary for cached_property
     )
 
     def __init__(
@@ -176,27 +174,27 @@ class Array(Sequence):
             return rng
         raise TypeError("range argument must be of type 'Range'")
 
-    @cached_property
+    @property
     def left(self) -> int:
         """Leftmost index of the array."""
-        return self._range.left
+        return self.range.left
 
-    @cached_property
+    @property
     def direction(self) -> str:
         """``'to'`` if indexes are ascending, ``'downto'`` otherwise."""
-        return self._range.direction
+        return self.range.direction
 
-    @cached_property
+    @property
     def right(self) -> int:
         """Rightmost index of the array."""
-        return self._range.right
+        return self.range.right
 
-    @cached_property
+    @property
     def length(self):
         """Length of the array."""
-        return self._range.length
+        return self.range.length
 
-    @cached_property
+    @property
     def range(self) -> Range:
         """:class:`Range` of the indexes of the array."""
         return self._range
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -243,7 +243,11 @@ class Array(Sequence):
             start_i = self._translate_index(start)
             stop_i = self._translate_index(stop)
             if start_i > stop_i:
-                raise IndexError("slice direction does not match array direction")
+                raise IndexError(
+                    "slice [{}:{}] direction does not match array direction [{}:{}]".format(
+                        start, stop, self.left, self.right
+                    )
+                )
             value = self._value[start_i : stop_i + 1]
             range = Range(start, self.direction, stop)
             return type(self)(value=value, range=range)
@@ -272,7 +276,11 @@ class Array(Sequence):
             start_i = self._translate_index(start)
             stop_i = self._translate_index(stop)
             if start_i > stop_i:
-                raise IndexError("slice direction does not match array direction")
+                raise IndexError(
+                    "slice [{}:{}] direction does not match array direction [{}:{}]".format(
+                        start, stop, self.left, self.right
+                    )
+                )
             value = self._construct_value(value)
             if len(value) != (stop_i - start_i + 1):
                 raise ValueError(
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -65,13 +65,10 @@ class Array(Sequence):
         Array(['4', '3', '2', '1'], Range(0, 'downto', -3))
 
     .. note::
-        Slice indexes must be specified in the same direction as the array and do not support specifying a "step".
 
-    .. note::
-        When setting a slice, the new value must be an iterable of the same size as the slice.
-
-    .. note::
-        Negative indexes are *not* treated as an offset from the end of the array, but are treated literally.
+        - Slice indexes must be specified in the same direction as the array and do not support specifying a "step".
+        - When setting a slice, the new value must be an iterable of the same size as the slice.
+        - Negative indexes are *not* treated as an offset from the end of the array, but are treated literally.
 
     Arrays support the methods and semantics defined by :class:`collections.abc.Sequence` including, but not limited to:
 
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -37,8 +37,8 @@ class Array(Sequence):
         >>> Array(range=Range(0, 'to', 1))
         Array([], Range(1, 'to', 0))
 
-    Indexing and slicing is very similar to :class:`list`\ s, except you use the indexing scheme you specified.
-    Like :class:`list`\ s, you don't have to specify a start or stop, and the start or end of the array are inferred.
+    Indexing and slicing is very similar to :class:`list`\ s, but it uses the indexing scheme specified.
+    Like :class:`list`\ s, if a start or stop index is not specified, it is inferred as the start or end of the array.
     Slicing an array returns a new :class:`~cocotb.types.Array` object, whose bounds are the slice indexes.
 
     .. code-block:: python3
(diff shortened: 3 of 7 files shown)


cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+from functools import cached_property
+from itertools import zip_longest
+from sys import maxsize
+
+
+class Array(Sequence):
+    r"""
+    Fixed-size, arbitrarily-indexed, heterogenous sequence type.
+
+    Arrays are similar, but different from Python :class:`list`\ s.
```
cmarqu, line 13:
```suggestion
    Arrays are similar to, but different from Python :class:`list`\ s.
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+from collections.abc import Sequence
+from .range import Range
+from functools import cached_property
+from itertools import zip_longest
+from sys import maxsize
+
+
+class Array(Sequence):
+    r"""
+    Fixed-size, arbitrarily-indexed, heterogenous sequence type.
```
cmarqu, line 11:
```suggestion
    Fixed-size, arbitrarily-indexed, heterogeneous sequence type.
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+
+    Arrays are similar, but different from Python :class:`list`\ s.
+    An array can store values of any type or values of multiple types at a time, just like a :class:`list`.
+    Unlike :class:`list`\ s, an array's size cannot change.
+
+    The indexes of an array can start or end at any integer value, they are not limited to 0-based indexing.
+    Indexing schemes can be either ascending or descending in value.
+    An array's indexes are described using a :class:`~cocotb.types.Range` objects.
```
cmarqu, line 19:
```suggestion
    An array's indexes are described using a :class:`~cocotb.types.Range` object.
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+    Initial values are treated as iterables, which are copied into an internal buffer.
+
+    .. code-block:: python3
+
+        >>> Array("1234")  # the 0-based range `(0, len(value)-1)` is infered
```
cmarqu, line 24:
```suggestion
        >>> Array("1234")  # the 0-based range `(0, len(value)-1)` is inferred
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+        Array(['1', '2', '3', '4'], Range(0, 3))
+
+        >>> Array(range=Range(0, 'downto', -3))  # the initial values are `None`
+        Array([None, None, None, None], Range(0, -3))
+
+        >>> Array([1, True, object(), 'example'], Range(-2, 1))  # inital value and range lengths must be equal
```
cmarqu, line 30:
```suggestion
        >>> Array([1, True, object(), 'example'], Range(-2, 1))  # initial value and range lengths must be equal
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+
+    Arrays also support "null" ranges; "null" arrays have zero length and cannot be indexed.
+
+    .. code-block:: python3
+
+        >>> Array(range=Range(0, 'to', 1))
+        Array([], Range(1, 'to', 0))
+
+    Indexing and slicing is very similar to :class:`list`\ s, except you use the indexing scheme you specified.
+    Like :class:`list`\ s, you don't have to specify a start or stop, and the start or end of the array are infered.
```
cmarqu, line 41:
```suggestion
    Like :class:`list`\ s, you don't have to specify a start or stop, and the start or end of the array are inferred.
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+        Array(['2', '3', '4'], Range(-1, 'downto', -3))
+        >>> b[:] = reversed(b)
+        >>> b
+        Array(['4', '3', '2', '1'], Range(0, 'downto', -3))
+
+    .. note::
+        Slice indexes must be specified in the same direction as the array and do not support specifying a "step".
+
+    .. note::
+        When setting a slice, the new value must be an iterable the same size as the slice you are selecting.
```
cmarqu, line 68:
```suggestion
        When setting a slice, the new value must be an iterable of the same size as the slice you are selecting.
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+
+    .. note::
+        Negative indexes are *not* treated as an offset from the end of the array, but are treated literally.
+
+    Arrays support the methods and semantics defined by :class:`collections.abc.Sequence` including, but not limited to:
+
+    - ``len(array)`` for getting the length of the array,
+    - ``value in array`` to check if an array contains a value,
+    - ``array.index(value)`` to find the index of the first occurences of a value in the array.
```
cmarqu, line 77:
```suggestion
    - ``array.index(value)`` to find the index of the first occurrence of a value in the array.
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+
+    Args
```
cmarqu, line 79:
```suggestion
    Args:
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+        self, value: Optional[Iterable[Any]] = None, range: Optional[Range] = None
+    ):
+        if value is not None and range is None:
+            self._value = list(value)
+            self._range = Range(0, "to", len(self._value) - 1)
+        elif value is not None and range is not None:
+            self._value = list(value)
+            self._range = range
+            if len(self._value) != len(self._range):
+                raise ValueError("Init value does not fit in given range")
```
cmarqu, line 96: Being more verbose might help debugging later.

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+                "indexes must be ints or slices, not {}".format(
+                    type(item).__name__
+                )
+            )
+
+    def __repr__(self) -> str:
+        return "{}({!r}, {!r})".format(type(self).__name__, self._value, self._range)
+
+    def index(self, value: Any, start: Optional[int] = None, stop: Optional[int] = None) -> int:
+        """Return index of first occurence of value."""
```
cmarqu, line 213:
```suggestion
        """Return index of first occurrence of *value*."""
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+        return self._range.right
+
+    @cached_property
+    def length(self):
+        """Length of the array"""
+        return self._range.length
+
+    @cached_property
+    def range(self) -> Range:
+        """:class:`Range` of the indexes of the array"""
```
cmarqu, line 125:
```suggestion
        """:class:`Range` of the indexes of the array."""
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+        return self._range.direction
+
+    @cached_property
+    def right(self) -> int:
+        """Rightmost index of the array"""
+        return self._range.right
+
+    @cached_property
+    def length(self):
+        """Length of the array"""
```
cmarqu, line 120:
```suggestion
        """Length of the array."""
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+        return self._range.left
+
+    @cached_property
+    def direction(self) -> str:
+        """``'to'`` if indexes are ascending, ``'downto'`` otherwise"""
+        return self._range.direction
+
+    @cached_property
+    def right(self) -> int:
+        """Rightmost index of the array"""
```
cmarqu, line 115:
```suggestion
        """Rightmost index of the array."""
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+            raise TypeError("must pass a value, range, or both")
+
+    @cached_property
+    def left(self) -> int:
+        """Leftmost index of the array"""
+        return self._range.left
+
+    @cached_property
+    def direction(self) -> str:
+        """``'to'`` if indexes are ascending, ``'downto'`` otherwise"""
```
cmarqu, line 110:
```suggestion
        """``'to'`` if indexes are ascending, ``'downto'`` otherwise."""
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+                raise ValueError("Init value does not fit in given range")
+        elif value is None and range is not None:
+            self._value = [None] * len(range)
+            self._range = range
+        else:
+            raise TypeError("must pass a value, range, or both")
+
+    @cached_property
+    def left(self) -> int:
+        """Leftmost index of the array"""
```
cmarqu, line 105:
```suggestion
        """Leftmost index of the array."""
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+            start = 0
+        if stop is not None:
+            stop = self._translate_index(stop)
+        else:
+            stop = maxsize  # same default value used by Python lists
+        idx = self._value.index(value, start, stop)
+        return self._range[idx]
+
+    def count(self, value: Any) -> int:
+        """Return number of occurences of value."""
```
cmarqu, line 226:
```suggestion
        """Return number of occurrences of *value*."""
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+        return self._value.count(value)
+
+    def _translate_index(self, item: int) -> int:
+        try:
+            return self._range.index(item)
+        except ValueError:
+            raise IndexError("index {} out of range".format(item)) from None
```
cmarqu, line 233:
```suggestion
            raise IndexError(f"index {item} out of range") from None
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+from functools import cached_property, lru_cache
+
+
+class Range(Sequence):
+    r"""
+    Variant of :class:`range` with inclusive right bound.
+
+    In Python, :class:`range` and :class:`slice` have a non-inclusive right bound.
+    In both Verilog and VHDL, ranges and arrays have an inclusive right bound.
+    This type mimic's Python's :class:`range` type, but supports HDL-like inclusive right bounds.
```
cmarqu, line 12:
```suggestion
    This type mimic's Python's :class:`range` type, but implements HDL-like inclusive right bounds.
```
?

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+        (1, 'to', 0)
+        >>> r.length
+        0
+
+    .. note::
+        This is only possible when specifying the direction.
+
+    Ranges also support all the features of :class:`range` including, but not limited to:
+
+    - ``value in range`` to see if a vlaue is in the range,
```
cmarqu, line 50:
```suggestion
    - ``value in range`` to see if a value is in the range,
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+    - ``range.index(value)`` to see what position in the range the value is,
+    - ``len(range)`` which is equivalent to :attr:`length`.
+
+    The typical use case of this type is in conjunction with :class:`~cocotb.types.Array`.
+
+    Args
```
cmarqu, line 56:
```suggestion
    Args:
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+        left: leftmost bound of range
+        direction: 'to' if values are ascending or 'downto' if decending
```
cmarqu, line 58:
```suggestion
        direction: 'to' if values are ascending or 'downto' if descending
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+        if direction is not None and right is None:
+            right, direction = direction, None
+        if direction is None:
+            step = 1 if left < right else -1
+        else:
+            step = self._translate_direction(direction)
+        self._range = range(left, right + step, step)
+
+    @staticmethod
+    @lru_cache(maxsize=None)
```
cmarqu, line 86: Do you intend to switch to `@cache` from `_py_compat` of that Logic/Bit PR once it is merged?

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+        if direction == "to":
+            return 1
+        elif direction == "downto":
+            return -1
+        else:
+            raise ValueError("direction must be 'to' or 'downto'")
+
+    @classmethod
+    def from_range(cls, rng: range) -> "Range":
+        """Converts :class:`range` to :class:`Range`"""
```
cmarqu, line 98:
```suggestion
        """Converts :class:`range` to :class:`Range`."""
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+        if rng.step not in (1, -1):
+            raise ValueError("step must be 1 or -1")
+        obj = cls.__new__(cls)
+        obj._range = rng
+        return obj
+
+    def to_range(self) -> range:
+        """Converts :class:`Range` to :class:`range`"""
```
cmarqu, line 106:
```suggestion
        """Convert :class:`Range` to :class:`range`."""
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+        return self._range
+
+    @cached_property
+    def left(self) -> int:
+        """Leftmost value in a range"""
```
cmarqu, line 111:
```suggestion
        """Leftmost value in a range."""
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+        return self._range.start
+
+    @cached_property
+    def direction(self) -> str:
+        """``'to'`` if values are meant to be ascending, ``'downto'`` otherwise"""
```
cmarqu, line 116:
```suggestion
        """``'to'`` if values are meant to be ascending, ``'downto'`` otherwise."""
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+        return "to" if self._range.step == 1 else "downto"
+
+    @cached_property
+    def right(self) -> int:
+        """Rightmost value in a range"""
```
cmarqu, line 121:
```suggestion
        """Rightmost value in a range."""
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+        return self._range.stop - self._range.step
+
+    @cached_property
+    def length(self) -> int:
+        """Length of range"""
```
cmarqu, line 126:
```suggestion
        """Length of range."""
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+        >>> r.to_range()
+        range(-2, 4)
+
+    Supports "null" ranges as seen in VHDL.
+    "null" ranges occur when a left bound cannot reach a right bound with the given direction.
+    They have a length of 0, but the left, right, and direction values remain as given.
+
+    .. code-block:: python3
+
+        >>> r = Range(1, 'to', 0)  # no way to count from 1 'to' 0
```
eric-wieser, line 39: What are your thoughts on the spelling `Range(1).to(0)`, where `type(Range).__call__` is overloaded to return some dummy object?

cocotb/types/array.py
ktbarrett: Do you have a particular opinion on when to use f strings vs format?

cocotb/types/range.py
ktbarrett: I can.

cocotb/types/array.py
ktbarrett: It would, but despite that most of the built in types have concise and low detail error message. I'm guessing the thought is you should use a debugger? I'll dive more into that since I haven't seen much opinion one way or the other.

cocotb/types/array.py
cmarqu: I'd do it like pyupgrade does it, only for "simple expressions" - actually, it would be easiest to just run it on the code.

ktbarrett:
I keep forgetting that `cache` and `cached_property` aren't available until 3.8. For the current use of `cached_property`, I could just alias `property` and those using less than 3.8 will suffer from slightly lower performance?

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+from functools import cached_property, lru_cache
+
+
+class Range(Sequence):
+    r"""
+    Variant of :class:`range` with inclusive right bound.
+
+    In Python, :class:`range` and :class:`slice` have a non-inclusive right bound.
+    In both Verilog and VHDL, ranges and arrays have an inclusive right bound.
+    This type mimic's Python's :class:`range` type, but implements HDL-like inclusive right bounds.
```
cmarqu, line 12:
```suggestion
    This type mimics Python's :class:`range` type, but implements HDL-like inclusive right bounds.
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+    Also supports :attr:`left`, :attr:`right`, and :attr:`length` attributes as seen in VHDL.
```
cmarqu, line 13:
```suggestion
    This type mimic's Python's :class:`range` type, but implements HDL-like inclusive right bounds,
    and also supports :attr:`left`, :attr:`right`, and :attr:`length` attributes as seen in VHDL.
```

cocotb/types/range.py
ktbarrett: It's neat, but it makes it hard to use an existing range to feed into another range if we do that.

```python
getattr(Range(r.left - 3), r.direction)(r.right)
# vs
Range(r.left - 3, r.direction, r.right)
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+
+    - ``value in range`` to see if a value is in the range,
+    - ``range.index(value)`` to see what position in the range the value is,
+    - ``len(range)`` which is equivalent to :attr:`length`.
+
+    The typical use case of this type is in conjunction with :class:`~cocotb.types.Array`.
+
+    Args:
+        left: leftmost bound of range
+        direction: 'to' if values are ascending or 'downto' if descending
```
cmarqu, line 58:
```suggestion
        direction: ``'to'`` if values are ascending, ``'downto'`` if descending
```

cocotb/types/array.py
```diff
@@ -0,0 +1,233 @@
...
+        Array([1, True, <object object at 0x7f4cff7b5570>, 'example'], Range(-2, 'to', 1))
+
+    Arrays also support "null" ranges; "null" arrays have zero length and cannot be indexed.
+
+    .. code-block:: python3
+
+        >>> Array(range=Range(0, 'to', 1))
+        Array([], Range(1, 'to', 0))
+
+    Indexing and slicing is very similar to :class:`list`\ s, except you use the indexing scheme you specified.
```
cmarqu, line 40:
```suggestion
    Indexing and slicing is very similar to :class:`list`\ s, except you use the indexing scheme specified.
```
?

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
...
+
+    :meth:`from_range` and :meth:`to_range` can be used to convert from and to :class:`range`.
+
+    .. code-block:: python3
+
+        >>> r = Range(-2, 3)
+        >>> r.to_range()
+        range(-2, 4)
+
+    Supports "null" ranges as seen in VHDL.
```
cmarqu, line 33:
```suggestion
    :class:`Range` supports "null" ranges as seen in VHDL.
```

cocotb/types/range.py
ktbarrett: I moved these changes into this PR as well, and provided a compat layer for `cached_property`.

cocotb/types/array.py
ktbarrett: I might try to strike `you` from the documentation entirely, there are only ~3 uses.

cocotb/types/range.py
ktbarrett:
```suggestion
    This type mimics Python's :class:`range` type, but implements HDL-like inclusive right bounds,
    and also supports :attr:`left`, :attr:`right`, and :attr:`length` attributes as seen in VHDL.
```

ktbarrett: It wasn't happy about me trying to apply multiple fixups to the same line.

cocotb/types/array.py
```diff
@@ -0,0 +1,237 @@
+from typing import Optional, Any, Iterable, Iterator, overload
```
cmarqu, line 1:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause
from typing import Optional, Any, Iterable, Iterator, overload
```

cocotb/types/range.py
```diff
@@ -0,0 +1,175 @@
+from typing import Any, Iterator, overload
```
cmarqu, line 1:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause
from typing import Any, Iterator, overload
```

tests/pytest/test_array.py
```diff
@@ -0,0 +1,147 @@
+from cocotb.types import Array, Range
```
cmarqu, line 1:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause
from cocotb.types import Array, Range
```

tests/pytest/test_range.py
```diff
@@ -0,0 +1,125 @@
+from cocotb.types import Range
```
cmarqu, line 1:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause
from cocotb.types import Range
```

cocotb/types/array.py
```diff
@@ -0,0 +1,240 @@
...
+        Array(['4', '3', '2', '1'], Range(0, 'downto', -3))
+
+    .. note::
+        Slice indexes must be specified in the same direction as the array and do not support specifying a "step".
+
+    .. note::
+        When setting a slice, the new value must be an iterable of the same size as the slice.
+
+    .. note::
+        Negative indexes are *not* treated as an offset from the end of the array, but are treated literally.
```
cmarqu, line 74:
```suggestion
    .. note::
        - Slice indexes must be specified in the same direction as the array and do not support specifying a "step".
        - When setting a slice, the new value must be an iterable of the same size as the slice.
        - Negative indexes are *not* treated as an offset from the end of the array, but are treated literally.
```
Maybe that looks better? (Assuming the markup is right, which it might not be.)

ktbarrett:
Use `lru_cache` and `property` together only works if the object is hashable, which `Array`s are not since they are mutable. So I went back to the original idea of just aliasing `property` to `cached_property` if it isn't available and for now people will have to deal with the performance drop (it shouldn't be noticeable anyways).

cocotb/types/array.py
ktbarrett: It doesn't render as a bulleted list.

cmarqu:
We should also add `cocotb/types/array.py` and `cocotb/types/range.py` after https://github.com/cocotb/cocotb/blob/1536fd14dddaed797ad5d68858baad1a5f578076/setup.cfg#L35 so that the doctests are also run.

cocotb/types/array.py
ktbarrett: I googled and asked a few people and there is no clear guidance. I'll add the slice value and the length of the value to the error message. I don't the value to be printed since it could be rather large.

ktbarrett:
@cmarqu if you think there are any other exception that could improve, please comment.

I need to refactor the `__init__` on `Array` slightly so I can better check that `range` argument, if specified, is in fact a `Range` since I depend on it's implementation in the body.

ktbarrett:
Resolved issue where constructor would not check if `range` value was a `Range`. The new implementation allows easy extension of type/value checking, etc.

ktbarrett:
Not entirely sure if the last set of changes is welcome, but it should make extending `Array` for `LogicArray` simpler. I already wanted to encapsulate range construction, since it was done in more than one place and needed to do type checking. We could even extend `_construct_range` to handle an iterable that can be splatted into a `Range` constructor to be able to support `Array(range=(1, 'to', 8))` without having to explicitly construct the value into a Range first, however, that might be of dubious merit.

cocotb/types/array.py
ktbarrett: I tried that and it didn't work on my machine. I guess I'll try again later.

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+    ):
+        if value is not None and range is None:
+            self._value = self._construct_value(value)
+            self._range = self._construct_range(Range(0, "to", len(self._value) - 1))
+        elif value is not None and range is not None:
+            self._value = self._construct_value(value)
+            self._range = self._construct_range(range)
+            if len(self._value) != len(self._range):
+                raise ValueError(
+                    "init value of length {!r} not fit in {!r}".format(
```
cmarqu, line 108:
```suggestion
                    "init value of length {!r} does not fit in {!r}".format(
```

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+            return self._value[idx]
+        elif isinstance(item, slice):
+            start = item.start or self.left
+            stop = item.stop or self.right
+            if item.step is not None:
+                raise IndexError("do not specify step")
+            start_i = self._translate_index(start)
+            stop_i = self._translate_index(stop)
+            if start_i > stop_i:
+                raise IndexError("slice direction does not match array direction")
```
cmarqu, line 245: A potential candidate for printing the directions or slice and array bounds. I have miscalculated slices in the past...

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+            self._value[idx] = value
+        elif isinstance(item, slice):
+            start = item.start or self.left
+            stop = item.stop or self.right
+            if item.step is not None:
+                raise IndexError("do not specify step")
+            start_i = self._translate_index(start)
+            stop_i = self._translate_index(stop)
+            if start_i > stop_i:
+                raise IndexError("slice direction does not match array direction")
```
cmarqu, line 274: See comment for this line in `__getitem__`.

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+            value = self._construct_value(value)
+            if len(value) != (stop_i - start_i + 1):
+                raise ValueError(
+                    "value  of length {!r} will not fit in slice [{}:{}]".format(
```
cmarqu, line 278:
```suggestion
                    "value of length {!r} will not fit in slice [{}:{}]".format(
```

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+                "indexes must be ints or slices, not {}".format(type(item).__name__)
+            )
+
+    def __repr__(self) -> str:
+        return "{}({!r}, {!r})".format(type(self).__name__, self._value, self._range)
+
+    def index(
+        self, value: Any, start: Optional[int] = None, stop: Optional[int] = None
+    ) -> int:
+        """Return index of first occurrence of *value*."""
```
cmarqu, line 294:
```suggestion
        """Return index of first occurrence of *value*.
        
        Search only within *start* and *stop* if given.
        """
```
(or something like that.)

cocotb/types/range.py
```diff
@@ -0,0 +1,178 @@
...
+
+    .. code-block:: python3
+
+        >>> r = Range(-2, 3)
+        >>> r.to_range()
+        range(-2, 4)
+
+    :class:`Range` supports "null" ranges as seen in VHDL.
+    "null" ranges occur when a left bound cannot reach a right bound with the given direction.
+    They have a length of 0, but the left, right, and direction values remain as given.
```
cmarqu, line 38:
```suggestion
    They have a length of 0, but the :attr:`left`, :attr:`right`, and :attr:`direction` values remain as given.
```
?

cocotb/types/range.py
```diff
@@ -0,0 +1,178 @@
...
+        if direction == "to":
+            return 1
+        elif direction == "downto":
+            return -1
+        else:
+            raise ValueError("direction must be 'to' or 'downto'")
+
+    @classmethod
+    def from_range(cls, rng: range) -> "Range":
+        """Converts :class:`range` to :class:`Range`."""
```
cmarqu, line 101:
```suggestion
        """Convert :class:`range` to :class:`Range`."""
```

cocotb/_py_compat.py
```diff
@@ -60,3 +61,20 @@ def __exit__(self, *excinfo):
...
+if sys.version_info < (3, 9):
+    def cache(f):
+        return lru_cache(maxsize=None)(f)
+else:
+    from functools import cache  # noqa: F401
+
+
+# Emulates Python 3.8's functools.cached_property decorator.
+# Uses lru_cache with LRU functionality, otherwise the cache could hold references to
+# dead objects and grow infinitely.
```
eric-wieser, line 76: I don't understand this comment - where is `lru_cache` used?

ktbarrett: Whoops, that got removed.

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+        >>> Array("1234")  # the 0-based range `(0, len(value)-1)` is inferred
+        Array(['1', '2', '3', '4'], Range(0, 'to', 3))
+
+        >>> Array(range=Range(0, 'downto', -3))  # the initial values are `None`
+        Array([None, None, None, None], Range(0, 'downto', -3))
+
+        >>> Array([1, True, None, 'example'], Range(-2, 1))  # initial value and range lengths must be equal
+        Array([1, True, None, 'example'], Range(-2, 'to', 1))
+
+    Arrays also support "null" ranges; "null" arrays have zero length and cannot be indexed.
```
eric-wieser, line 36: Isn't indexing with `[:]` allowed?

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+        Returns:
+            The :class:`Range` object describing the array's indexing scheme.
+
+        Raises:
+            TypeError: When type isn't supported.
+            ValueError: When the value prevents construction into a range.
+        """
+        if isinstance(rng, Range):
+            return rng
+        raise TypeError("range argument must be of type 'Range'")
```
eric-wieser, line 180: I'd be tempted to leave these extensionality hooks to a future PR - it's hard to review extension points without an example of something extending them.

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+
+    @cached_property
+    def length(self):
+        """Length of the array."""
+        return self._range.length
+
+    @cached_property
+    def range(self) -> Range:
+        """:class:`Range` of the indexes of the array."""
+        return self._range
```
eric-wieser, line 205: Why not just call the attribute `self.range` in the first place? `cached_property` does exactly that anyway, but defers it to first access.

ktbarrett: Read-only access. But maybe that's not super important.

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+        return reversed(self._value)
+
+    def __contains__(self, item: Any) -> bool:
+        return item in self._value
+
+    def __eq__(self, other: Any) -> bool:
+        try:
+            return all(a == b for a, b in zip_longest(self, other, fillvalue=object()))
+        except TypeError:
+            return NotImplemented
```
eric-wieser, line 223: This feels very python2-esque - silencing the TypeError seems like a bad idea to me.

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+        elif isinstance(item, slice):
+            start = item.start or self.left
+            stop = item.stop or self.right
+            if item.step is not None:
+                raise IndexError("do not specify step")
+            start_i = self._translate_index(start)
+            stop_i = self._translate_index(stop)
+            if start_i > stop_i:
+                raise IndexError("slice direction does not match array direction")
+            value = self._value[start_i : stop_i + 1]
```
eric-wieser, line 246: I don't see any documentation describing this behavior; at the very least, this needs a big `.. warning` in the docstring saying that unlike _all builtin python objects_, the slice indices are inclusive in the upper bound.

I'm personally dubious whether this is a good idea at all, while convenient for an HDL programmer, it feels like a trap for python programmers, and for passing these objects into functions expecting well-behaved sequences.

One compromise would be to reserve `value.v[x:y]` for verilog-style inclusive indexing, and use `value[x:y]` for python-style index.

cocotb/types/range.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def index(self, item: Any) -> int:
+        return self._range.index(item)
+
+    def count(self, item: Any) -> int:
+        return self._range.count(item)
+
+    def __repr__(self) -> str:
+        return "{}({!r}, {!r}, {!r})".format(
+            type(self).__name__, self.left, self.direction, self.right
```
eric-wieser, line 177:
```suggestion
            type(self).__qualname__, self.left, self.direction, self.right
```

cocotb/types/range.py
eric-wieser: That's a fair argument.

I suppose another gimmick you could play, but probably not in this PR, is `R[3:"downto":1]` as a shorthand, but the string literal makes me uneasy for the same reason.

cocotb/types/array.py
eric-wieser: This doesn't make it readonly anyway - try it and see.

eric-wieser: `@property` instead of `@cached_property` would.

cocotb/types/range.py
ktbarrett: Perhaps `Range[3:1]` `Range[3:1:-1]` syntax can be added as a follow on.

cocotb/types/array.py
ktbarrett: Hmmmm, it worked the second time, maybe there is something up with my environment.

cocotb/types/array.py
ktbarrett: It does not. Should it?

cocotb/types/array.py
ktbarrett: Yeah. At one point it was doing some non-trivial stuff, but it's not necessary anymore. I just removed all uses of `cached_property` and replaced it with `property`. It also means the code in `_py_compat` isn't exactly correct.

cocotb/types/array.py
ktbarrett: I think this is not right anyways.

cocotb/types/array.py
ktbarrett: Well `_construct_range` itself isn't really useful for overloading, but I already wanted to split it off into another function because I call it in multiple places in the constructor. `_construct_element` and `_construct_value` are of immediate use for `LogicArray`. 

```python
@staticmethod
def _construct_element(elem: Any) -> Any:
    return Logic(elem)

@staticmethod
def _construct_value(value: Iterable[Any]) -> List[Any]:
    return [self._construct_element(v) for v in value]
```
For automatic conversion of all values into `Logic`.

ktbarrett:
Factored out the code for guessing the direction based on `(left, right)`, I realized it was different in two places, so it's best to live in just one place.

cocotb/types/range.py
ktbarrett: Looks like this is preferable. https://github.com/ericvsmith/dataclasses/issues/100

ktbarrett: Looks like this is preferable https://github.com/ericvsmith/dataclasses/issues/100.

cocotb/types/array.py
ktbarrett: Added info. It now reads.

```python
>>> from cocotb.types import Array, Range
>>> a = Array("1234")
>>> a[3:0]
  File "<stdin>", line 1, in <module>
  File "/home/kaleb/dev/cocotb/cocotb/types/array.py", line 246, in __getitem__
    raise IndexError(
IndexError: slice [3:0] direction does not match array direction [0:3]
```

cocotb/types/array.py
ktbarrett: I added a sentence stating that slicing works just like the `Range` object being fed to it, it uses an inclusive right bound. This parallels how slicing regular Python sequence types and `range` work. The whole point of `Range` was to describe index sequences with inclusive right bounds using typical VHDL parlance (or Verilog-esque automatic directionality); so slicing should behave the same way.

The first thing I tried was just using `range` and non-inclusive right bound. It's very awkward for people accustomed to HDL. Non-inclusive right bounds and non-0-based indexing don't play well together and it isn't very ergonmic. Every non-0-based indexes language I know of uses inclusive bounds. If supporting Python's semantics here is worth doing (I don't agree that it is), it should be the second-class syntax.

Nothing in `collections.abc` or any documentation I see states explicitly that higher-order functions that take `Sequence`s will assume a particular indexing scheme. The default mixins of `Sequence` do though... which is why I had to override them. I guess it's a good question if this is even a `Sequence` or not?

cocotb/types/array.py
```diff
@@ -0,0 +1,330 @@
...
+        elif value is None and range is not None:
+            self._value = self._construct_value(None for _ in range)
+            self._range = self._construct_range(range)
+        else:
+            raise TypeError("must pass a value, range, or both")
+
+    @staticmethod
+    def _construct_value(value: Iterable[Any]) -> List[Any]:
+        """
+        Constructs the value portion of the array.
```
cmarqu, line 118:
```suggestion
        Construct the value portion of the array.
```
```suggestion
        Constructs the value portion of the array.
```

cocotb/types/array.py
```diff
@@ -0,0 +1,330 @@
...
+        Raises:
+            TypeError: When the type isn't supported.
+            ValueError: When the iterable's values prevent construction into a list of elements.
+        """
+        return list(value)
+
+    @staticmethod
+    def _construct_element(elem: Any) -> Any:
+        """
+        Constructs a single element of an array.
```
cmarqu, line 139:
```suggestion
        Construct a single element of an array.
```
```suggestion
        Constructs a single element of an array.
```

cocotb/types/array.py
```diff
@@ -0,0 +1,330 @@
...
+        Raises:
+            TypeError: When the type isn't supported.
+            ValueError: When the value prevents construction into an element.
+        """
+        return elem
+
+    @staticmethod
+    def _construct_range(rng: Any) -> Range:
+        """
+        Constructs the range portion of the array.
```
cmarqu, line 159:
```suggestion
        Construct the range portion of the array.
```
```suggestion
        Constructs the range portion of the array.
```

cocotb/types/array.py
eric-wieser: > The default mixins of Sequence do though

As does the builtin `bisect` module - so we probably need to warn against using that in the docstring too.

cocotb/types/array.py
```diff
@@ -0,0 +1,330 @@
...
+        sentinel = object()
+        try:
+            it = zip_longest(self, other, fillvalue=sentinel)
+        except TypeError:
+            return NotImplemented
+        for a, b in it:
+            # sentinel check MUST come before element equality check
+            if b == sentinel or a != b:
+                return False
+        return True
```
eric-wieser, line 226: I'm a little confused by this - can you add a docstring describing exactly what you define as equality (and why?).

For instance, is `Array([1, 2, 3]) == (1, 2, 3)` true? How about `Array([1, 2, 3]) == [1, 2, 3]`? Are you sure comparison to arbitrary sequence types is something you want to support given that `list` and `tuple` don't behave that way?

ktbarrett: Ehhh, I probably shouldn't. It's easy enough to just convert it to an `Array` first.

cocotb/types/array.py
ktbarrett: Added additional warning information on differences with builtin sequence types.

ktbarrett:
@eric-wieser More code examples added.

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+from typing import Optional, Any, Iterable, Iterator, overload, List
```
ktbarrett, line 4:
```suggestion
from typing import Optional, Any, Iterable, Iterator, overload
```

ktbarrett:
Took @eric-wieser's advice and got rid of unnecessary subclass override points. Left the element constructor, since that can be used for enforcing homogenously-typed arrays.

tests/pytest/test_array.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def test_bad_construction():
+    with pytest.raises(TypeError):
+        Array()
+    with pytest.raises(TypeError):
+        Array(value=1)
+    with pytest.raises(TypeError):
+        Array(range=tuple())
+
```
marlonjames, line 40: Missing coverage:
```suggestion
    with pytest.raises(TypeError):
        Array(range=tuple())
    with pytest.raises(TypeError):
        Array("1234", range(0, 4))

```

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+        else:
+            raise TypeError("must pass a value, range, or both")
+
+    @staticmethod
+    def _construct_element(elem: Any) -> Any:
+        """
+        Construct a single element of an array.
+
+        Will be fed elements of the *value* iterable if given to the constructor,
+        :data:`None` when no *value* is given to the constuctor,
```
ktbarrett, line 153:
```suggestion
        :data:`None` when no *value* is given to the constructor,
```

cocotb/types/array.py
```diff
@@ -0,0 +1,314 @@
...
+        the *value* in __setitem__ when assigning to a single index,
```
ktbarrett, line 154:
```suggestion
        the *value* in ``__setitem__`` when assigning to a single index,
```

ktbarrett (commented):
From #2514.

cocotb/types/range.py
```diff
@@ -0,0 +1,173 @@
...
+        """Rightmost value in a range."""
+        return self._range.stop - self._range.step
+
+    @property
+    def length(self) -> int:
+        """Length of range."""
+        return len(self._range)
+
+    def __len__(self) -> int:
+        return self.length
```
eric-wieser, line 128:
```suggestion
    def __len__(self) -> int:
        """Length of range."""
        return len(self._range)
```
Let's not provide two ways to spell the same thing. Python has a canonical way to ask for the length of something, let's use it.

cocotb/types/range.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+class Range(Sequence):
+    r"""
+    Variant of :class:`range` with inclusive right bound.
+
+    In Python, :class:`range` and :class:`slice` have a non-inclusive right bound.
+    In both Verilog and VHDL, ranges and arrays have an inclusive right bound.
+    This type mimics Python's :class:`range` type, but implements HDL-like inclusive right bounds,
+    and also supports :attr:`left`, :attr:`right`, and :attr:`length` attributes as seen in VHDL.
```
eric-wieser, line 15:
```suggestion
    and uses :attr:`left` and :attr:`right` instead of ``start`` and ``stop`` to match VHDL and indicate this
    difference.
```

cocotb/types/array.py
```diff
@@ -0,0 +1,325 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+from typing import Optional, Any, Iterable, Iterator, overload
+from collections.abc import Sequence
+from .range import Range
+from sys import maxsize
+from itertools import chain
```
ktbarrett, line 8:
```suggestion
```

cocotb/types/array.py
```diff
@@ -0,0 +1,325 @@
...
+            TypeError: when *other* is an object of dissimilar type.
+        """
+        if not isinstance(other, self.__class__):
+            raise TypeError(
+                "unsupported operand types for concat() {!r} and {!r}".format(
+                    self.__class__.__qualname__,
+                    other.__class__.__qualname__
+                )
+            )
+        return self.__class__(chain(self, other))
```
ktbarrett, line 295:
```suggestion
        return self.__class__(self._value + other._value)
```

cmarqu:
If you have a PASSing mypy setup, should we check this also in CI? Maybe in a different PR.

ktbarrett:
@cmarqu I have no interest in ever using mypy seriously.

cocotb/types/__init__.py
```diff
@@ -3,3 +3,28 @@
...
+
+def concat(a: Array, b: Array) -> Array:
+    """
+    Create a new array that is the concatenation of one array with another.
+
+    Uses the :meth:`__concat__` or :meth:`__rconcat__` special methods to dispatch to a particular implementation,
+    exactly like other binary operations in Python.
+
+    Raises:
+        TypeError: when *other* is an object of dissimilar type.
```
cmarqu, line 16:
```suggestion
        TypeError: when *b* is an object of dissimilar type.
```

ktbarrett: I think it needs a rewrite.
```suggestion
        TypeError: when concatenation is not supported between the two object.
```

cocotb/types/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    if hasattr(a, '__concat__'):
+        res = a.__concat__(b)
+        if res is not NotImplemented:
+            return res
+    if hasattr(b, '__rconcat__'):
+        res = b.__rconcat__(a)
+        if res is not NotImplemented:
+            return res
+    raise TypeError(
+        "can't concat value of type {!r} with value of type {!r}".format(
```
ktbarrett, line 27:
```suggestion
        "cannot concatenate {!r} with {!r}".format(
```

cocotb/types/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+            a.__class__.__qualname__, b.__class__.__qualname__
```
ktbarrett, line 28:
```suggestion
        "can only concatenate {a!r} (not {b!r}) to {a!r}".format(
            a=a.__class__.__qualname__, b=b.__class__.__qualname__
```

cocotb/types/__init__.py
ktbarrett, line 27:
```suggestion
        "cannot concatenate {!r} with {!r}".format(
```

cocotb/types/__init__.py
```diff
@@ -0,0 +1,30 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+from .range import Range  # noqa: F401
+from .array import Array  # noqa: F401
+
+
+def concat(a: Array, b: Array) -> Array:
```
eric-wieser, line 8: To make things confusing, python already has `operator.concat`, and from the C API it is actually possible to make it mean something different from operator.add

cocotb/types/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Create a new array that is the concatenation of one array with another.
+
+    Uses the :meth:`__concat__` or :meth:`__rconcat__` special methods to dispatch to a particular implementation,
+    exactly like other binary operations in Python.
+
+    Raises:
+        TypeError: when the arguments do not support concatenation in the given order.
+    """
+    if hasattr(a, '__concat__'):
```
eric-wieser, line 18: There are situations where `Rrconcat` should be tried first: see https://github.com/brettcannon/desugar/blob/290bf659bfc054af3979a9ab29003db6f855e959/desugar/operator.py#L75

cocotb/types/__init__.py
ktbarrett: I was totally unaware that existed... But, if it doesn't work with it's own special methods in Python, that's a no go for us. I just wish Python had it's own concat operator :disappointed:

cocotb/types/__init__.py
ktbarrett: Ah, I think I understand. If the right hand operation is a subtype of the left hand and the `__rconcat__` is not the same function, it's likely something more specific. The whole `__op__`, `__rop__` business being a poor man's multiple dispatch, where in such a system the runtime would pick the most specific implementation, the second in the example being the most specific if `concat(Array, Unsigned)` were called.

```
concat(Array, Array) -> Array
concat(Array, Unsigned) -> Array  # Unsigned repr is an int here, so we have to do some extra work if we are concatenating with an Array
```

ktbarrett: Addressed in the most recent commit.

cocotb/types/__init__.py
eric-wieser: Well, you could build a custom C type that connected those hooks to `__concat__` methods if you really wanted too, but that's probably not worth the trouble.

cocotb/types/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Uses the :meth:`__concat__` or :meth:`__rconcat__` special methods to dispatch to a particular implementation,
+    exactly like other binary operations in Python.
+
+    Raises:
+        TypeError: when the arguments do not support concatenation in the given order.
+    """
+    a_concat = getattr(a.__class__, "__concat__", None)
+    a_rconcat = getattr(a.__class__, "__rconcat__", None)
+    b_rconcat = getattr(b.__class__, "__rconcat__", None)
```
eric-wieser, line 20: Strictly the semantics should be:
```suggestion
    a_concat = getattr(type(a), "__concat__", None)
    a_rconcat = getattr(type(a), "__rconcat__", None)
    b_rconcat = getattr(type(b), "__rconcat__", None)
```

cocotb/types/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+
+    if isinstance(b, a.__class__) and a_rconcat != b_rconcat:
```
eric-wieser, line 22:
```suggestion
    if issubclass(type(b), type(a)) and type(a) is not type(b) and a_rconcat is not b_rconcat:
```

But I'd recommend you just take the `desugar` version with less editorialization.

cocotb/types/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    elif a.__class__ != b.__class__:
+        # normal call order
+        call_order = ((a, a_concat, b), (b, b_rconcat, a))
+    else:
+        # types are the same, we expect implementation of 'concat(a, b)' to be in 'a.__concat__'
+        call_order = ((a, a_concat, b),)
+
+    for lhs, method, rhs in call_order:
+        if method is None:
+            continue
```
eric-wieser, line 34: THis makes `__concat__ = None` permitted, which is unusual for python magic methods.

cocotb/types/__init__.py
ktbarrett: Is there a difference between `type` and `__class__` in Python 3? I have read that they are equivalent, and `__class__` is faster. I have used `__class__` everywhere in these PRs, if there is a good reason to switch I guess I should take this as a general comment.

eric-wieser:
```
In [90]: class Foo:
    ...:     @property
    ...:     def __class__(self):
    ...:         return "oops"
    ...:

In [91]:

In [91]: type(Foo())
Out[91]: __main__.Foo

In [92]: Foo().__class__
Out[92]: 'oops'
```

eric-wieser: More realistically, this comes up if someone defines an object with a custom `__getattribute__`

cocotb/types/__init__.py
ktbarrett: I've read that using `is not` for type comparisons is not advised, I'm guessing that's because parameterized types (depending on the implementation) might not return the same object given the same parameters, type `==` can handle that.

Using `isinstance` is equivalent here and I don't get why there needs to be protection against matching the same type, seeing as the second part of the condition will not match in that case. If people write code in a non-intuitive way and *don't* put a comment as to why they did it that way, I consider it superfluous.

cocotb/types/__init__.py
ktbarrett: Alright. I'll update all the occurrences in this PR.

cocotb/types/__init__.py
eric-wieser: > I've read that using is not for type comparisons is not advised,

I think that's usually in the context of "are you sure you didn't mean `not issubclass(X, Y)`"

> If people write code in a non-intuitive way and don't put a comment as to why they did it that way, I consider it superfluous.

The comment is in the description of the package as a whole - it's written to exactly match what CPython does internally.

CPython does things the way it does for historical reasons; in particular, the `a_rconcat != b_rconcat` condition was a mistake that was too late to revert.
If we're going to add a python-like dunder hook, then we should make it python like, along with all the quirks. Otherwise, the absence of those quirks can bite users.

> parameterized types (depending on the implementation) might not return the same object given the same parameters

IMO this would be a very bad design.
At any rate, Cpython internall uses `is` not `==`, so we should match it.

ktbarrett: [CPython binary operation for Python-defined types](https://github.com/python/cpython/blob/a328d73843cfd42d2aee1434c78df1ef2845931a/Objects/typeobject.c#L6940).

ktbarrett: > the a_rconcat != b_rconcat condition was a mistake that was too late to revert.

The mistake is to check that at all, right? If the right hand side is a subclass of the left, the right `__rop__` should *always* be called first (if it exists)?

marlonjames (approved):

cocotb/types/array.py
```diff
@@ -0,0 +1,329 @@
...
+        >>> b
+        Array(['4', '3', '2', '1'], Range(0, 'downto', -3))
+
+    .. warning::
+        Arrays behave differently in certain situations than Python's builtin sequence types (:class:`list`, :class:`tuple`, etc.).
+
+        - Arrays are not necessarily 0-based and slices use inclusive right bounds,
+          so many functions that work on Python sequences by index (like :mod:`bisect`) may not work on arrays.
+        - Slice indexes must be specified in the same direction as the array and do not support specifying a "step".
+        - When setting a slice, the new value must be an iterable of the same size as the slice.
```
ktbarrett, line 73:
```suggestion
        - When setting a slice, the new value must be an Array of the same size as the slice.
```

cocotb/types/array.py
```diff
@@ -0,0 +1,329 @@
...
+        raise TypeError(
+            "indexes must be ints or slices, not {}".format(type(item).__name__)
+        )
+
+    @overload
+    def __setitem__(self, item: int, value: Any) -> None:
+        pass  # pragma: no cover
+
+    @overload
+    def __setitem__(self, item: slice, value: Iterable[Any]) -> None:
```
ktbarrett, line 255:
```suggestion
    def __setitem__(self, item: slice, value: "Array") -> None:
```

cocotb/types/array.py
marlonjames: Is this something you want to enforce?

ktbarrett: At first I thought it was inconsistent because `Logic` can only be combined with `Logic`, but slice assignment on `Array` can work with non-`Array`s? We also decided to not support implicit conversion to `Logic` in binary operations with `Logic`, but for `LogicArray` implicit casting is either required, or we only support assigning with other `LogicArray`s, which is just totally inconsistent.

ktbarrett: Basically it comes down to whether we want to support implicit casting of `Iterable[Any]` in `LogicArray` slice assignments or not. I'm choosing the "no implicit casting" solution because that is what we did for `Logic`.

ktbarrett: `Iterable[Any]` is fine because `Array`s are that anyways.

ktbarrett:
I think that's enough last minute cleanup @cocotb/maintainers. If there are no more comments in a couple days, I will merge.

PR #2514, opened 2021-04-10
Adds LogicArray type

ktbarrett:
Based on #2485 and #2510. Adds `LogicArray` specialization of `Array`. These types contain functions for converting to and from integers and logical operations.

~~One issue that needs to be resolved in this PR and #2510 is the typing annotations are not linking correctly because I'm not using the original names `cocotb.types.logic_array.LogicArray`. @eric-wieser You are familiar with Sphinx, any idea what can be done here? Changing the annotation to use the "expected" name like `"cocotb.types.LogicArray"` instead of `LogicArray` corrects the issue, but I don't think is a good solution.~~ Sphinx 4 makes this work.

~~Another question is whether `concat` should exist in the `Array` base class?~~ It now does

Starting out as a draft until #2485 and #2510 are merged.

### DIFF
--- a/cocotb/types/logic_array.py
+++ b/cocotb/types/logic_array.py
@@ -89,7 +89,7 @@ class LogicArray(Array[Logic]):
         >>> la.signed_integer   # uses two's complement representation
         -6
 
-    :class:`LogicArray` \s also support element-wise logical operations: ``&``, ``|``,
+    :class:`LogicArray`\ s also support element-wise logical operations: ``&``, ``|``,
     ``^``, and ``~``.
 
     .. code-block:: python3
--- a/cocotb/types/logic_array.py
+++ b/cocotb/types/logic_array.py
@@ -4,112 +4,15 @@
 import typing
 
 from cocotb.types.array import Array
-from cocotb.types.logic import Bit, Logic, LogicConstructibleT
+from cocotb.types.logic import Logic, LogicConstructibleT
 from cocotb.types.range import Range
 
 LogicT = typing.TypeVar("LogicT", bound=Logic)
 S = typing.TypeVar("S")
-Self = typing.TypeVar("Self", bound="_LogicArrayBase[Logic]")
+Self = typing.TypeVar("Self", bound="LogicArray")
 
 
-class _LogicArrayBase(Array[LogicT]):
-    r"""
-    Base class for :class:`Array`\ s of :class:`Logic`\ s
-    """
-
-    _element_type: typing.Type[LogicT]
-
-    def __init__(
-        self,
-        value: typing.Iterable[LogicConstructibleT],
-        range: typing.Optional[Range] = None,
-    ) -> None:
-        constructor = type(self)._element_type
-        super().__init__(
-            value=(constructor(v) for v in value),
-            range=range,
-        )
-
-    @typing.overload
-    def __setitem__(self, item: int, value: LogicConstructibleT) -> None:
-        ...
-
-    @typing.overload
-    def __setitem__(
-        self, item: slice, value: typing.Iterable[LogicConstructibleT]
-    ) -> None:
-        ...
-
-    def __setitem__(
-        self,
-        item: typing.Union[int, slice],
-        value: typing.Union[LogicConstructibleT, typing.Iterable[LogicConstructibleT]],
-    ) -> None:
-        constructor = type(self)._element_type
-        if isinstance(item, int):
-            super().__setitem__(
-                item, constructor(typing.cast(LogicConstructibleT, value))
-            )
-        elif isinstance(item, slice):
-            super().__setitem__(
-                item,
-                (
-                    constructor(v)
-                    for v in typing.cast(typing.Iterable[LogicConstructibleT], value)
-                ),
-            )
-        else:
-            raise TypeError(
-                "indexes must be ints or slices, not {}".format(type(item).__name__)
-            )
-
-    def __str__(self) -> str:
-        return "".join(str(v) for v in self)
-
-    def __repr__(self) -> str:
-        return "{}({!r}, {!r})".format(type(self).__qualname__, str(self), self.range)
-
-    def __and__(self: Self, other: Self) -> Self:
-        if isinstance(other, type(self)):
-            if len(self) != len(other):
-                raise ValueError(
-                    "cannot perform bitwise & on arrays of different length"
-                )
-            return type(self)(a & b for a, b in zip(self, other))  # type: ignore
-        return NotImplemented
-
-    def __rand__(self: Self, other: Self) -> Self:
-        return self & other
-
-    def __or__(self: Self, other: Self) -> Self:
-        if isinstance(other, type(self)):
-            if len(self) != len(other):
-                raise ValueError(
-                    "cannot perform bitwise | on arrays of different length"
-                )
-            return type(self)(a | b for a, b in zip(self, other))  # type: ignore
-        return NotImplemented
-
-    def __ror__(self: Self, other: Self) -> Self:
-        return self | other
-
-    def __xor__(self: Self, other: Self) -> Self:
-        if isinstance(other, type(self)):
-            if len(self) != len(other):
-                raise ValueError(
-                    "cannot perform bitwise ^ on arrays of different length"
-                )
-            return type(self)(a ^ b for a, b in zip(self, other))  # type: ignore
-        return NotImplemented
-
-    def __rxor__(self: Self, other: Self) -> Self:
-        return self ^ other
-
-    def __invert__(self: Self) -> Self:
-        return type(self)(~v for v in self)
-
-
-class LogicArray(_LogicArrayBase[Logic]):
+class LogicArray(Array[Logic]):
     r"""
     Fixed-sized, arbitrarily-indexed, array of :class:`cocotb.types.Logic`.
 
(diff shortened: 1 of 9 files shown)


cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+
+    def __rxor__(self, other: Any) -> "LogicArray":
+        return self ^ other
+
+    def __invert__(self) -> "LogicArray":
+        return self.__class__(~a for a in self)
+
+    def concat(self, other: Any) -> "LogicArray":
+        """
+        Creates a new array that is the concatenation one array with another.
```
cmarqu, line 116:
```suggestion
        Creates a new array that is the concatenation of one array with another.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+                "unsupported operand types for concat() {!r} and {!r}".format(
+                    self.__class__.__qualname__,
+                    other.__class__.__qualname__
+                )
+            )
+        return self.__class__(chain(self, other))
+
+    def is_resolvable(self) -> bool:
+        """
+        Returns ``True`` if the array is only comprised of ``Logic("0")`` and ``Logic("1")```.
```
cmarqu, line 132:
```suggestion
        Return ``True`` if the array is only comprised of ``Logic("0")`` and ``Logic("1")```.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+
+        Useful when you want to know if an array can be converted into a :class:`~cocotb.types.BitArray`
+        or :class:`int` without raising an exception.
+        """
+        return all(v in (Logic('0'), Logic('1')) for v in self)
+
+    @classmethod
+    def from_twos_complement(cls, value: int, range: Optional[Range] = None) -> "LogicArray":
+        """
+        Creates an array with the two's complement binary representation of the given *value*.
```
cmarqu, line 142:
```suggestion
        Create an array with the two's complement binary representation of the given *value*.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+                raise ValueError(f"signed integer {value} does not fit in {range!r}")
+        if value < 0:
+            value &= (1 << length) - 1
+        value_binstr = format(value, f"0{length}b")
+        return cls(value_binstr, range)
+
+    @classmethod
+    def from_unsigned(cls, value: int, range: Optional[Range] = None) -> "LogicArray":
+        """
+        Creates an array with the unsigned binary representation of the given :class:`int`.
```
cmarqu, line 169:
```suggestion
        Create an array with the unsigned binary representation of the given :class:`int`.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+        else:
+            length = len(range)
+            if value < _min_value_unsigned(length) or value > _max_value_unsigned(length):
+                raise ValueError(f"unsigned integer {value} does not fit in {range!r}")
+        value_binstr = format(value, f"0{length}b")
+        return cls(value_binstr, range)
+
+    def to_twos_complement(self) -> int:
+        """
+        Converts an array into an :class:`int` assuming two's complement representation.
```
cmarqu, line 191:
```suggestion
        Convert an array into an :class:`int` assuming two's complement representation.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+
+        Raises:
+            ValueError: when the array contains non resolvable values.
```
cmarqu, line 194:
```suggestion
            ValueError: when the array contains non-resolvable values.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+        unsigned = int(str(self), 2)
+        max_value = _max_value_signed(len(self))
+        if unsigned > max_value:
+            return unsigned - (2**len(self))
+        else:
+            return unsigned
+
+    def to_unsigned(self) -> int:
+        """
+        Converts an array into an :class:`int` assuming unsigned representation.
```
cmarqu, line 207:
```suggestion
        Convert an array into an :class:`int` assuming unsigned representation.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+
+        Raises:
+            ValueError: when the array contains non resolvable values.
```
cmarqu, line 210:
```suggestion
            ValueError: when the array contains non-resolvable values.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+    @classmethod
+    def from_unsigned(cls, value: int, range: Optional[Range] = None) -> "LogicArray":
+        """
+        Creates an array with the unsigned binary representation of the given :class:`int`.
+
+        If *range* is specified, the resulting array has the given range.
+        If *range* is not specified, a 0-based range is chosen, large enough to fit the value.
+
+        Raises:
+            ValueError: when the value will not fit in the given *range*
```
cmarqu, line 175:
```suggestion
            ValueError: when the value will not fit in the given *range*.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+    @classmethod
+    def from_twos_complement(cls, value: int, range: Optional[Range] = None) -> "LogicArray":
+        """
+        Creates an array with the two's complement binary representation of the given *value*.
+
+        If *range* is specified, the resulting array has the given range.
+        If *range* is not specified, a 0-based range is chosen, large enough to fit the value.
+
+        Raises:
+            ValueError: when the value will not fit in the given *range*
```
cmarqu, line 148:
```suggestion
            ValueError: when the value will not fit in the given *range*.
```

tests/pytest/test_logic_array.py
```diff
@@ -0,0 +1,119 @@
+from cocotb.types import LogicArray, Logic, Range
```
cmarqu, line 1:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause
from cocotb.types import LogicArray, Logic, Range
```

documentation/source/library_reference.rst
```diff
@@ -64,6 +64,33 @@ Interacting with the Simulator
...
 
+HDL Datatypes
+-------------
+
+These are a set of datatypes that model the behavior of common HDL datatypes.
+They can be used independently of cocotb for modeling and will replace :class:`BinaryValue`
+as the types used by cocotb's `simulator handles <#simulation-object-handles>`_.
+
+.. versionadded:: 2.0
+
```
cmarqu, line 75:
```suggestion
.. inheritance-diagram:: cocotb.types
   :parts: 1
   
```

documentation/source/newsfragments/2059.feature.rst
```diff
@@ -0,0 +1 @@
+Added :class:`~cocotb.types.Logic` and :class:`~cocotb.types.Bit` modeling datatypes.
```
cmarqu, line 1:
```suggestion
Added modeling datatypes
:class:`~cocotb.types.Logic`, :class:`~cocotb.types.Bit`,
:class:`~cocotb.types.Array`, :class:`~cocotb.types.Range`,
:class:`~cocotb.types.LogicArray` and :class:`~cocotb.types.BitArray`.
```

cmarqu:
It would be nice to also support `__lshift__` and `__rshift__`.

ktbarrett:
@cmarqu would that be make a new array of the same size and do lossy shifts in either direction? Should __rshift__ be 0 or sign extending?

eric-wieser:
Haven't looked at the diff yet, but I think (very) recent sphinx versions added a way to set "canonical" names to resolve what you're describing.

documentation/source/newsfragments/2059.feature.rst
ktbarrett: These are going to be split up into multiple PRs and newsfragments.

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+        ValueError: When argument values cannot be used to construct an array.
+        TypeError: When invalid argument types are used.
+    """
+
+    __slots__ = ()
+
+    _construct_element = Bit
+
+
+@lru_cache
```
ktbarrett, line 242:
```suggestion
@lru_cache()
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+def _min_value_signed(bitwidth: int) -> int:
+    return -(2 ** (bitwidth - 1))
+
+
+@lru_cache
```
ktbarrett, line 247:
```suggestion
@lru_cache()
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,258 @@
...
+def _max_value_signed(bitwidth: int) -> int:
+    return (2 ** (bitwidth - 1)) - 1
+
+
+def _min_value_unsigned(bitwidth: int) -> int:
+    return 0
+
+
+@lru_cache
```
ktbarrett, line 256:
```suggestion
@lru_cache()
```

ktbarrett (commented):
Prior to Python 3.8, `lru_cache` didn't support wrapping without calling.

documentation/source/library_reference.rst
```diff
@@ -64,6 +64,35 @@ Interacting with the Simulator
...
+-------------
+
+These are a set of datatypes that model the behavior of common HDL datatypes.
+They can be used independently of cocotb for modeling and will replace :class:`BinaryValue`
+as the types used by cocotb's `simulator handles <#simulation-object-handles>`_.
+
+.. versionadded:: 2.0
+.. inheritance-diagram:: cocotb.types
+   :parts: 1
+
```
ktbarrett, line 77:
```suggestion
.. inheritance-diagram:: cocotb.types
   :parts: 1

```

cmarqu:
> @cmarqu would that make a new array of the same size and do lossy shifts in either direction?

Yes, I guess so. There is separate in-place operators (https://docs.python.org/3/library/operator.html#operator.ilshift) which I haven't come across yet IIRC.

> Should `__rshift__` be 0 or sign extending?

I'd say 0-extending since a plain `LogicArray`/`BitArray` is just a bunch of bits so you cannot know whether there is a sign bit or not.

documentation/source/library_reference.rst
```diff
@@ -64,6 +64,35 @@ Interacting with the Simulator
...
 
+HDL Datatypes
+-------------
+
+These are a set of datatypes that model the behavior of common HDL datatypes.
+They can be used independently of cocotb for modeling and will replace :class:`BinaryValue`
+as the types used by cocotb's `simulator handles <#simulation-object-handles>`_.
+
+.. versionadded:: 2.0
+.. inheritance-diagram:: cocotb.types
```
cmarqu, line 75:
```suggestion

.. inheritance-diagram:: cocotb.types
```
It's not showing up, assuming it's because of the missing newline.

ktbarrett: Just seems like it isn't working. I don't see any warnings either.

ktbarrett:
@cmarqu I ask about `__rshift__` because future subtypes of `BitArray`: `Unsigned`, `Signed`, `Sfixed`, and `Ufixed` would have well defined shift and arithmetic. Defining `__rshift__` would also break subtype relations in the subtypes that don't do the same behavior.

cmarqu:
> @cmarqu I ask about `__rshift__` because future subtypes of `BitArray`: `Unsigned`, `Signed`, `Sfixed`, and `Ufixed` would have well defined shift and arithmetic. Defining `__rshift__` would also break subtype relations in the subtypes that don't have the same shifting behavior.

We don't want breakage of course, but sometimes you just want to shift raw bits. Ah well, users need to just define their own function or their own subclass with a method in that case.

documentation/source/library_reference.rst
ktbarrett: I think because we aren't defining the types in `__init__`. Sphinx ignores imported things. Maybe if I add them to an `__all__` variable it will show up.

tests/pytest/test_logic_array.py
```diff
@@ -13,6 +13,29 @@ def test_logic_array_constructor():
...
+    assert LogicArray(0).signed_integer == 0
+    assert LogicArray(0).binstr == "0"
+    assert LogicArray(10).integer == 10
+    assert LogicArray(10).signed_integer == -6
+    assert LogicArray(10).binstr == "1010"
+    assert LogicArray(-6).integer == 10
+    assert LogicArray(-6).signed_integer == -6
+    assert LogicArray(-6).binstr == "1010"
+    assert LogicArray(0).is_resolvable
+    assert not LogicArray("XX").is_resolvable
```
cmarqu, line 37:
```suggestion
    assert not LogicArray("1X1").is_resolvable
```
to make sure that not just the first or last element is taken for the resolvability check?

cmarqu:
Maybe remove `BitArray` from the PR heading and description.

ktbarrett (commented):
Should add descriptions of integer conversion, `binstr`, `integer`, and `signed_integer` to docstrings.

ktbarrett:
Crap... I have `LogicArray` instantiated from `int`s being `downto`, and those instantiated from arrays being `to`. I'm going to go out on a limb and say that the range should be `downto` in both cases, if not given.

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,247 @@
...
+        >>> l = LogicArray("1010")
+        >>> l[3] = "Z"
+        >>> l[3]
+        Logic('Z')
+
+        >>> l[2:] = ['X', True, 0]
+        >>> l
+        LogicArray('ZX10', Range(3, 'downto', 0))
+
+    LogicArrays can be converted into :class:`str`\ s or :class:`int`\s.
```
cmarqu, line 78:
```suggestion
    LogicArrays can be converted into :class:`str`\ s or :class:`int`\ s.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,247 @@
...
+        if isinstance(value, int):
+            if value < 0:
+                bitlen = int.bit_length(value + 1) + 1
+            else:
+                bitlen = max(1, int.bit_length(value))
+            if range is None:
+                self._value = [Logic(v) for v in _int_to_bitstr(value, bitlen)]
+            else:
+                if bitlen > len(range):
+                    raise ValueError(f"{value} will not fit in {range}")
```
cmarqu, line 132:
```suggestion
                    raise ValueError(f"{value} of length {bitlen} will not fit in {range}")
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,247 @@
...
+        return NotImplemented
+
+    def __rxor__(self: Self, other: Self) -> Self:
+        return self ^ other
+
+    def __invert__(self: Self) -> Self:
+        return type(self)(~v for v in self)
+
+
+def _int_to_bitstr(value: int, bit_length: int) -> str:
```
cmarqu, line 244: We have historically used `n_bits` to mean "bit length".
You are also using `bitlen` elsewhere, maybe normalize those two.

documentation/source/newsfragments/2514.feature.rst
```diff
@@ -0,0 +1 @@
+Added :class:`~cocotb.types.LogicArray` modeling datatype.
```
cmarqu, line 1:
```suggestion
Added the :class:`~cocotb.types.LogicArray` modeling datatype.
```

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,247 @@
...
+            )
+
+    def __repr__(self) -> str:
+        return "{}({!r}, {!r})".format(type(self).__qualname__, self.binstr, self.range)
+
+    def __and__(self: Self, other: Self) -> Self:
+        if isinstance(other, type(self)):
+            if len(self) != len(other):
+                raise ValueError(
+                    "cannot perform bitwise & on arrays of different length"
```
cmarqu, line 208: Maybe even print value and length of the two arguments?

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,247 @@
...
+Self = typing.TypeVar("Self", bound="LogicArray")
+
+
+class LogicArray(Array[Logic]):
+    r"""
+    Fixed-sized, arbitrarily-indexed, array of :class:`cocotb.types.Logic`.
+
+    .. currentmodule:: cocotb.types
+
+    LogicArrays can be constructed from either iterables of values constructible into
```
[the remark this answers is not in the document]
ktbarrett: It gets a little cluttered. I wondering if we should adopt a rule to only make a reference once in the same section and avoid self-references. I guess we can decide on that later.

marlonjames (approved):

cocotb/types/logic_array.py
```diff
@@ -0,0 +1,253 @@
...
+        >>> la.binstr
+        '1010'
+
+        >>> la.integer          # uses unsigned representation
+        10
+
+        >>> la.signed_integer   # uses two's complement representation
+        -6
+
+    :class:`LogicArray` \s also support element-wise logical operations: ``&``, ``|``,
```
cmarqu, line 92:
```suggestion
    :class:`LogicArray`\ s also support element-wise logical operations: ``&``, ``|``,
```

cmarqu (approved):

Issue #2522, opened 2021-04-15
Documentation update - python 3.5 is no longer supported for example, python requires libpython3xxx.so


hiviah:
# Several fixes to docs needed

SW versions:

* cocotb version 1.5.1
* Ubuntu 16.04 LTS 64-bit
* Icarus Verilog 11 64-bit
* several Python versions tried, 3.5.2 (system), 3.7.4, 3.9.4 (manually built, in virtualenv)

So multiple issues:

## Python 3.5 no longer supported because of 3.6 syntax

`pip3 install cocotb` no longer works with python 3.5, since it fails on python 3.6 syntax:

    % pip3 install --user cocotb
    Collecting cocotb
      Using cached https://files.pythonhosted.org/packages/e0/be/113b2722dfdf2a44aad58e2ec416ab0a0ae0346bddb0d2cf2e596fcb2a60/cocotb-1.5.1.tar.gz
        Complete output from command python setup.py egg_info:
        Modelsim/Questa executable (vdbg) executable not found. No FLI interface will be available.
        Traceback (most recent call last):
          File "<string>", line 1, in <module>
          File "/tmp/pip-build-k2mm149g/cocotb/setup.py", line 138, in <module>
            "bus": ["cocotb_bus"]
          File "/home/ondro/.local/lib/python3.5/site-packages/setuptools/__init__.py", line 140, in setup
            return distutils.core.setup(**attrs)
          File "/usr/lib/python3.5/distutils/core.py", line 108, in setup
            _setup_distribution = dist = klass(attrs)
          File "/home/ondro/.local/lib/python3.5/site-packages/setuptools/dist.py", line 370, in __init__
            k: v for k, v in attrs.items()
          File "/usr/lib/python3.5/distutils/dist.py", line 281, in __init__
            self.finalize_options()
          File "/home/ondro/.local/lib/python3.5/site-packages/setuptools/dist.py", line 529, in finalize_options
            ep.load()(self, ep.name, value)
          File "/home/ondro/.local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2325, in load
            return self.resolve()
          File "/home/ondro/.local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2331, in resolve
            module = __import__(self.module_name, fromlist=['__name__'], level=0)
[...]

PR #2523, opened 2021-04-15
Fix missing libpython

ktbarrett:
`find_libpython` returns a non-0 exit code when a libpython isn't found, we need to do the same since the makefiles expect that and will print and appropriate error message. The refactor to this module accidentally removed that behavior. xref #2522.

### DIFF
--- a/cocotb/config.py
+++ b/cocotb/config.py
@@ -156,7 +156,7 @@ def lib_name_path(interface, simulator):
     return library_name_path
 
 
-def findlibpython():
+def _findlibpython():
     libpython_path = find_libpython.find_libpython()
     if libpython_path is None:
         sys.exit(1)
--- a/cocotb/config.py
+++ b/cocotb/config.py
@@ -156,6 +156,13 @@ def lib_name_path(interface, simulator):
     return library_name_path
 
 
+def findlibpython():
+    libpython_path = find_libpython.find_libpython()
+    if libpython_path is None:
+        sys.exit(1)
+    return libpython_path
+
+
 class PrintAction(argparse.Action):
     def __init__(self, option_strings, dest, text=None, **kwargs):
         super().__init__(option_strings, dest, nargs=0, **kwargs)
@@ -218,8 +225,10 @@ def get_parser():
     parser.add_argument(
         "--libpython",
         help="Print the absolute path to the libpython associated with the current Python installation",
-        action=PrintAction,
-        text=find_libpython.find_libpython(),
+        nargs=0,
+        metavar=(),
+        action=PrintFuncAction,
+        function=findlibpython,
     )
     parser.add_argument(
         "--lib-dir",
--- a/cocotb/config.py
+++ b/cocotb/config.py
@@ -228,7 +228,7 @@ def get_parser():
         nargs=0,
         metavar=(),
         action=PrintFuncAction,
-        function=findlibpython,
+        function=_findlibpython,
     )
     parser.add_argument(
         "--lib-dir",
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -133,6 +133,8 @@ manually override the LIBPYTHON_LOC make variable with the absolute path to libp
 
 endef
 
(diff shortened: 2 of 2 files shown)


cocotb/config.py
```diff
@@ -156,6 +156,13 @@ def lib_name_path(interface, simulator):
     return library_name_path
 
 
+def findlibpython():
+    libpython_path = find_libpython.find_libpython()
+    if libpython_path is None:
+        sys.exit(1)
```
themperek, line 162: Would exception be better?

ktbarrett: Printing nothing and having a non-0 exit code is exactly what `find_libpython` does. Because this is just a wrapper, I don't see why we shouldn't do the same thing.

There is a return code check and error message in the makefiles. I guess it doesn't matter where the error message comes from, but I still need to catch non-0 exit and then `$(error) in the makefiles.

themperek: Was thinking about something like this 
```suggestion
        raise Exception('find_libpython was not able to find a libpython in the current Python environment')
```
but I see now it does not matter.

cocotb/share/makefiles/Makefile.inc
```diff
@@ -133,6 +133,8 @@ manually override the LIBPYTHON_LOC make variable with the absolute path to libp
 
 endef
 
+ifndef LIBPYTHON_LOC
```
themperek, line 136: Can one do?
```
LIBPYTHON_LOC ?= $(strip $(subst $(FIND_LIBPYTHON_RC)QQQQ,,$(FIND_LIBPYTHON_RES)QQQQ))
```

On the other hand, you do not want to call `cocotb-config --libpython` if not needed. 
Maybe a comment about this would be nice.

Something else:
If one rises execution (above) in `cocotb-config` maybe do not need all this makefile part/massage just?

`export LIBPYTHON_LOC ?= $(shell cocotb-config --libpython)`

Formating is not so nice but visible and makefile is cleaner but is a matter of taste.

themperek (approved):

cocotb/config.py
```diff
@@ -156,6 +156,13 @@ def lib_name_path(interface, simulator):
     return library_name_path
 
 
+def findlibpython():
```
eric-wieser, line 159:
```suggestion
def _findlibpython():
```
To make this clear this isn't public API

cocotb/config.py
```diff
@@ -218,8 +225,10 @@ def get_parser():
     parser.add_argument(
         "--libpython",
         help="Print the absolute path to the libpython associated with the current Python installation",
-        action=PrintAction,
-        text=find_libpython.find_libpython(),
+        nargs=0,
+        metavar=(),
+        action=PrintFuncAction,
+        function=findlibpython,
```
ktbarrett, line 231:
```suggestion
        function=_findlibpython,
```

cocotb/share/makefiles/Makefile.inc
ktbarrett: > `LIBPYTHON_LOC ?= $(strip $(subst $(FIND_LIBPYTHON_RC)QQQQ,,$(FIND_LIBPYTHON_RES)QQQQ))`

Well if `cocotb-config --libpython` fails the makefiles will still die here, even if the user specified a value for `LIBPYTHON_LOC`.

> `export LIBPYTHON_LOC ?= $(shell cocotb-config --libpython)`

The issue with this is that failing shell calls don't cause make to stop. It will: print the error message, `LIBPYTHON_LOC` will be empty (I think exceptions are printed on stderr), execution will continue. I would still need to check the return code in this instance to prevent execution from continuing.

We need to:
1. Allow the user override `LIBPYTHON_LOC`
2. Have the test fail iff `cocotb-config --libpython` fails and the user didn't set `LIBPYTHON_LOC`
3. (Optional) Avoid calling `cocotb-config` if `LIBPYTHON_LOC` is specified.

The current solution is the one that hits all the points with the smallest diff. I'm also backporting these commits onto 1.5.

themperek: 👍🏻

PR #2531, opened 2021-04-20
Run only test_scheduler in CI with increased logging

cmarqu:
Aldec Riviera was unhappy with the long logs introduced by #2085.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,9 +45,13 @@ do_tests::
 	$(MAKE) -k -C examples
 # increase coverage
 do_tests::
-	$(MAKE) -k -C tests/test_cases/test_cocotb/ COCOTB_LOG_LEVEL=DEBUG > test_cocotb_DEBUG.log
+	$(MAKE) -k -C tests/test_cases/test_cocotb/ \
+		MODULE=test_scheduler TESTCASE=test_coroutine_kill,test_coroutine_close_down \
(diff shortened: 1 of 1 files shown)


marlonjames:
Looks like it's still too big.
Locally I get ~50MB for **test_scheduler_SCHEDULER_DEBUG.log**.

cmarqu:
Redirecting to `/dev/null` now, but it also may be too much because of the `| tee sim_build.log` in the Makefile. Let's see.

imphil:
I'm missing a bit of context here and on the previous PR:
* Why are we running tests more than once just for increased logging?
* What do we gain except for coverage for logging statements?

ktbarrett (commented):
Why not add this to the test_cocotb Makefile? Run `test_scheduler` with `COCOTB_SCHEDULER_DEBUG=1` while the rest run normal. Use recursive calls to get the selection.

ktbarrett:
Needs rebase.

marlonjames:
Additionally, I found that running the same test multiple times will overwrite coverage data.
The fix for that is to add `data_suffix=True` to the `coverage.coverage()` object initialization, and then collect `.coverage.cocotb*` when combining the final coverage stats.

imphil:
@cmarqu is this still relevant?

cmarqu:
No, not relevant anymore.

PR #2549, opened 2021-05-20
Remove debug code

toddstrader:
Just noticed this while tracing through the code.


imphil:
Thanks! (I actually found out the other day that this code would now be just `breakpoint()` in newer versions of Python, which is rather handy!)

PR #2547, opened 2021-05-18
Add check for TOPLEVEL in Makefiles

haroonshafique-10xe:
Referring to #2364, this PR adds checks for TOPLEVEL in makefiles for icarus and verilator.

PR #2550, opened 2021-05-26
Bump version of Icarus to use with clang test

ktbarrett:
Should fix regression failures on master. Currently blocking #2547.


imphil:
What has changed? Doesn't Icarus 10.3 build (any more) with the clang shipped with Ubuntu 20.04 (or 20.10?), or does cocotb now require Icarus 11.0?

ktbarrett:
Newer clang doesn't build old Icarus. The version of icarus is irrelevant in this test.

PR #2552, opened 2021-05-28
Update README.md

ktbarrett:
Update python version in readme to 3.6+.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 The current stable version of cocotb requires:
 
-- Python 3.5+
+- Python 3.6+
 - Python development packages (Python/C API headers and embedding library)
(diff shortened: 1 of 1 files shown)


marlonjames:
Stable (v1.5.2) is 3.5+.
We could add a note for current development / master branch being 3.6+ ?

ktbarrett:
The README should reflect the current state of the repo IMO.

imphil:
The README is a bit of a tricky place since many people use it to learn about cocotb, and others use it to install the latest master version. I generally agree that the README should reflect master, but it's easy to confuse users here. How about we reduce the README to a minimum and point to the versioned documentation instead?

marlonjames:
I'm fine with having it reflect the master branch, I was mostly commenting that the line above it says
> The current stable version of cocotb requires:

PR #2561, opened 2021-06-10
Make Event an Awaitable[None]

### LINKED ISSUE #2560: Make Event an Awaitable[None]
Make `cocotb.triggers.Event` an `Awaitable[None]`. Supply an `__await__` implementation that proxies `event.wait()` and returns `None`.

ktbarrett:
Closes #2560.


eric-wieser (changes requested):
As mentioned in the issue, I don't think we should do this.

> There should be one-- and preferably only one --obvious way to do it.

Adding a second way that nothing else in the python ecosystem seems to use doesn't exactly feel obvious.

PR #2566, opened 2021-06-20
Change generated names of TestFactory tests

### LINKED ISSUE #2548: Customization of results.xml with "factory.add_option(...)"
Hello! I'd like to suggest a feature to the results.xml file that we get at the end of a simulation.

I'm running some tests permutations with a code that looks like this:
```python
for test in [test1, test2]:
        factory = TestFactory(test)
        factory.add_option("config1",            [True, False])
        factory.add_option("config2",            ["opt-a", "opt-b", "opt-c"])
        factory.add_option("config3",            [True, False])
        factory.generate_tests()
```

Then in my results.xml, I got the list of all the test permutations with this setting.
Right now it is only possible to get the name of the test, if it has passed/failed, its simulation time, its real time, and its ratio.

Could you guys add a way to add to each test in the results.xml the value of the custom options?

ktbarrett:
Closes #2548.

Previously the test name did not reflect the options passed and were
simply incrementing numbers appended to the test name. This made it hard
to determine from the results.xml what options were passed to any test:
`test_name_001`, `test_name_002`, etc.

This change formats the name like a function call with named arguments:
`test_name(arg1="arg", arg2=6)`, `test_name(arg1="", arg2=101)`, etc.

See [here](https://github.com/cocotb/cocotb/runs/2866967563?check_suite_focus=true#step:25:762).

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -639,7 +639,7 @@ class TestFactory:
         Groups of options are now supported
 
     .. versionchanged:: 2.0
-        Options are used in names instead of numbers
+        Options are listed in test names instead of a counter
     """
 
     # Prevent warnings from collection of TestFactories by unit testing frameworks.
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -637,6 +637,9 @@ class TestFactory:
 
     .. versionchanged:: 1.5
         Groups of options are now supported
+
+    .. versionchanged:: 2.0
+        Options are used in names instead of numbers
     """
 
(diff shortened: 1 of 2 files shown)


cocotb/regression.py
```diff
@@ -637,6 +637,9 @@ class TestFactory:
 
     .. versionchanged:: 1.5
         Groups of options are now supported
+
+    .. versionchanged:: 2.0
+        Options are used in names instead of numbers
```
ktbarrett, line 642:
```suggestion
        Passed options are used in test names instead of a counter
```

cmarqu (approved):

PR #2576, opened 2021-06-23
doc: add ref to Maintainers section in CONTRIBUTING.md through extlinks

umarcor:
As discussed in gitter, this adds a reference to the list of maintainers in the contributors section of the docs.

I could not test this because `tox -e docs` does not work on MSYS2, and it seems that CI for documentation is not triggered in GitHub Actions of the forks. I hope it is triggered by this PR.

### DIFF
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -35,6 +35,7 @@ extensions = [
     'sphinx.ext.imgmath',
     'sphinx.ext.viewcode',
     'sphinx.ext.napoleon',
+    'sphinx.ext.extlinks',
     'sphinx.ext.intersphinx',
     'sphinxcontrib.makedomain',
     'sphinx.ext.inheritance_diagram',
@@ -347,3 +348,11 @@ in_progress_notes = subprocess.check_output(['towncrier', '--draft', '--name', '
                                             universal_newlines=True)
 with open('master-notes.rst', 'w') as f:
     f.write(in_progress_notes)
+
+# -- External link helpers -----------------------------------------------------
+
+extlinks = {
+   'wikipedia': ('https://en.wikipedia.org/wiki/%s', None),
+   'reposharp': ('https://github.com/cocotb/cocotb/issues/%s', '#'),
+   'reposrc':   ('https://github.com/cocotb/cocotb/blob/master/%s', None)
+}
--- a/documentation/source/contributors.rst
+++ b/documentation/source/contributors.rst
@@ -9,7 +9,8 @@ Contributors
    Solarflare
 
 cocotb is developed and maintained by an active community.
-Our GitHub repository contains a list of all `contributors <https://github.com/cocotb/cocotb/graphs/contributors>`_.
+Our GitHub repository contains a list of the :reposrc:`Maintainers <CONTRIBUTING.md#maintainers>` and all
+`contributors <https://github.com/cocotb/cocotb/graphs/contributors>`_.
 
 Since mid-2018 cocotb is an independent project under the umbrella of the
 `Free and Open Source Silicon Foundation <https://www.fossi-foundation.org>`_.


umarcor:
It's green now:

- https://readthedocs.org/projects/cocotb/builds/14078976/
- https://cocotb--2576.org.readthedocs.build/en/2576/contributors.html#contributors

/cc @ktbarrett @cmarqu

ktbarrett (approved):

ktbarrett:
> I could not test this because tox -e docs does not work on MSYS2

Did you use the version of tox in the official repos? Just `pip install tox` does not work. mingw's Python environment is... "special" in such a way that many Python tools are simply broken without patching, which is doubly broken by anything using PEP517 build isolation or tools like tox where packages are fetched from PyPI and not local patched versions. That being said. I fixed `find_libpython` and we removed the (previously broken in mingw) dependency on `setuptools_scm`, so cocotb *should* work in mingw and msys2 environments now. If that is not the case, feel free to file an issue. We don't currently officially support the mingw and msys2 environments, but maybe if we can get CI in place to test in those environments with all the issues fixed, we would officially support them.

cmarqu (approved):

umarcor:
> Did you use the version of tox in the official repos? Just `pip install tox` does not work. mingw's Python environment is... "special" in such a way that many Python tools are simply broken without patching, which is doubly broken by anything using PEP517 build isolation or tools like tox where packages are fetched from PyPI and not local patched versions.

Note that I was not criticising nor expecting you to fix anything on your side. I just wanted to state why I was force-pushing several times until I got it correct.

Yeah, tox is pretty broken on MSYS2. We use tox in VUnit too, and it's so painful.
However, during the last year, several issues were fixed and last time I tried it seemed to work.
In fact, we are currently using it in VUnit's CI: https://github.com/VUnit/vunit/blob/master/.github/workflows/push.yml#L99-L142.
I think you can just copy and paste that, and do some minimal modifications:

- Might want to install iverilog and/or verilator too.
- Might want to preinstall some additional system packages.

It seems that I already tried it in this repo a year ago:

- https://github.com/umarcor/cocotb/commits/gha
- https://github.com/umarcor/cocotb/commit/a3801dda18ac2728abb7b8f07275b4303ab7aaaa

I don't recall what failed. Maybe GHDL tests were not in shape and iverilog/verilator were not available in MSYS2 yet?
@cmarqu, I vaguely recall pinging you about it in some issue of this repo. Do you remember?

> That being said. I fixed `find_libpython` and we removed the (previously broken in mingw) dependency on `setuptools_scm`, so cocotb _should_ work in mingw and msys2 environments now. If that is not the case, feel free to file an issue.

I'm not sure about this having anything to do with running cocotb. I think I've used cocotb on MSYS2 successfully before. I tried doing it with verilator and it had some problem, but I believe I achieved it with iverilog and/or GHDL.
I guess the problem here is tox.
It tried to install Pillow, and compile it from sources...
Therefore, "fixing" the issue should be a matter of finding which dependencies need to be installed through pacman and not pip or tox.

> We don't currently officially support the mingw and msys2 environments, but maybe if we can get CI in place to test in those environments with all the issues fixed, we would officially support them.

However, installation guidelines (https://docs.cocotb.org/en/stable/install.html#installation-of-prerequisites) suggest installing msys2 through conda. That is exactly the same, because the author of the msys2 packages in conda was the same person who added Python support to msys2, and his purpose was precisely to automatically expose all msys2 packages through conda (thus the `m2` prefix). That's why I believe that cocotb does already work on msys2, even though you are not explicitly testing it.
The point is that I have always tried to use it with open source simulators only. I don't know if anyone tried using cocotb on MSYS2 with vendor simulators.

BTW, there are some relevant changes coming with regard to Python and MSYS2: https://twitter.com/msys2org/status/1407073330746757128.

ktbarrett:
@umarcor 

> I just wanted to state why I was force-pushing several times until I got it correct.

Most of my patches I force-push a bunch. Why run test locally when GH will do it for me? =)

> That's why I believe that cocotb does already work on msys2, even though you are not explicitly testing it

It definitely will build in mingw and msys2, the real kicker is getting everything to work with their Python installation. The Anaconda installation with the msys2 GCC and Make works great, however. I tried once before to get mingw/msys2 to work, maybe I should finish that effort out. BTW thanks for the work on the setup-msys2 GHA, I'm using it for `find_libpython` and it works great!

>I guess the problem here is tox.

Tox also doesn't work correctly on Windows with Anaconda (`tox-conda` is also broken), so we just don't use tox in our Windows CI. Problem solved =)

PR #2583, opened 2021-06-27
Deprecate gen tests options

### LINKED ISSUE #2575: Deprecate TestFactory.generate_tests prefix and postfix
They originally existed so repeated usage of the same function with a `TestFactory` would not cause name conflicts. This was because previously the generated test name was simply the test name with a counter appended to it. Now that `TestFactory` generated tests use the option values in the test name, conflict can only occur if the same test with the same options occur in different `TestFactory` calls. This should instead cause a warning to be printed and the conflict resolved by appending a counter to the test name of the conflicting test.

ktbarrett:
Closes #2575. Deprecates the `prefix` and `postfix` options to `TestFactory.generate_tests`. Also changes how name collisions are handled since prefix and postfix are deprecated and collisions should be far less likely.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -34,6 +34,7 @@ import sys
 import os
 import traceback
 import pdb
+import warnings
 from typing import Any, Optional, Tuple, Iterable
 
 import cocotb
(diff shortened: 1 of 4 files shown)


cmarqu (approved):

ktbarrett:
Turns out the current implementation of the test name is not going to work and the solution is going to require these options to remain. I'll cherry-pick the tests for the next PR to fix this mess I made.

PR #2604, opened 2021-06-29
Add ghdl expected errors

ktbarrett:
Supersedes #2532. Pinging @umarcor. Hopefully the regression should now pass with GHDL.

Add expected errors/fails to tests that GHDL is currently failing. Also, fixed some exemptions for Icarus. Filed [several issues](https://github.com/cocotb/cocotb/issues?q=is%3Aopen+is%3Aissue+label%3Acategory%3Asimulators%3Aghdl+label%3Aupstream).

PR #2609, opened 2021-07-03
Add MINGW64 shell and packages to regression

ktbarrett:
Test cocotb using MINGW64 packages and shell to ensure it builds and runs in that environment. Pinging @umarcor. I took what you wrote in your fork, touched it up, and added the ability to fail.

Also, forgot to remove the `may_fail` tag in #2604.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -115,7 +115,6 @@ jobs:
             lang: vhdl
             python-version: 3.8
             os: ubuntu-latest
-            may_fail: true
 
             # Test Verilator on Ubuntu
 
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -325,3 +325,42 @@ jobs:
           -n "${{matrix.extra_name}}${{matrix.sim}} (${{matrix.sim-version}}) | ${{matrix.os}} | Python ${{matrix.python-version}}" \
           -e SIM,TOPLEVEL_LANG,CXX,OS,PYTHON_VERSION \
           -a "-rl"
+
+  mingw-env:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - sim: verilator
+            lang: verilog
+            pkg: verilator
+            may_fail: true
+          - sim: icarus
+            lang: verilog
+            pkg: iverilog
+          - sim: ghdl
+            lang: vhdl
+            pkg: ghdl-llvm
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+    - uses: msys2/setup-msys2@v2
+      with:
+        msystem: MINGW64
+        update: true
+        install: |
+            git
+            make
+            mingw-w64-x86_64-gcc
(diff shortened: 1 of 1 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -115,7 +115,6 @@ jobs:
             lang: vhdl
             python-version: 3.8
             os: ubuntu-latest
-            may_fail: true
```
cmarqu, line 118: I thought you left this in on purpose because `nightly` IMO makes no claim to be stable and thus would conflate ghdl behavior changes with cocotb PR changes.
But I now realize we have the same setup for icarus master already...

ktbarrett: Nightly has to pass their regressions. Icarus master does not, but there is no nightly branch there, so we just hope it does break. It hasn't broken to date...

cmarqu: Makes sense then.

cmarqu (approved):

themperek (approved):

.github/workflows/regression-tests.yml
umarcor: As @ktbarrett said, assets in GHDL are only updated after all tests are successful. Therefore, using nighly assets is as safe as 'stable releases'. For that reason, tagged releases of GHDL do not have assets anymore.

.github/workflows/regression-tests.yml
```diff
@@ -325,3 +324,42 @@ jobs:
...
+          - sim: verilator
+            lang: verilog
+            pkg: verilator
+            may_fail: true
+          - sim: icarus
+            lang: verilog
+            pkg: iverilog
+          - sim: ghdl
+            lang: vhdl
+            pkg: ghdl-llvm
```
umarcor, line 343: You might use `ghdl`, instead of `ghdl-llvm`. That should work on either MINGW32 or MINGW64.

Are you interested in testing on MINGW32 or cocotb is only expected to be used on 64 bits?

ktbarrett: > Are you interested in testing on MINGW32

Not unless I have to. Our regressions are long enough already.

> is only expected to be used on 64 bits?

We still have to deal with the 32 bit free version of Modelsim. Until that goes away for good we need to care about 32 bit support. That being said, we have 0 users using MINGW right now. The reason for adding MINGW at all is because their Python setup is weird, I put in effort to get it to work, and I'd like to track our support of that environment.

ktbarrett: Are you still interested in packaging cocotb for MINGW?

umarcor: Fair enough.

For the record, in https://github.com/hdl/MINGW-packages/actions/runs/1080755117 I'm running cocotb with GHDL and iverilog on MINGW64 and MING32. That is a canary workflow though, not something I check every day.

umarcor: > Are you still interested in packaging cocotb for MINGW?

Yes. However, I did not package any Python project yet. I need to check/learn what is specific compared to packagin Ada/C/C++.

PR #2613, opened 2021-07-05
Simplify and expand env var passthrough for tox

imphil:
* Instead of individually listing all possible vendor license daemon
  environment variables, simply pass through all environment variables
  that are typically used with a glob pattern.
* Pass through `MTI_*` environment variables, which are used by
  Quest/ModelSim, e.g. to choose between a 64 or 32 bit installation.
  Again, instead of choosing individual variables, simply pass through all
  of them to avoid playing whack-a-mole.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -25,12 +25,13 @@ setenv =
 passenv =
     SIM
     TOPLEVEL_LANG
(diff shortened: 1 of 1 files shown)


ktbarrett (approved):
Looks good to me.

Issue #644, opened 2018-08-13
ImportError with import hashlib.py


flamehj:
I get this error when running the examples:
`Traceback (most recent call last):`
` File "/home/d.baburin/work/cocotb/cocotb/__init__.py", line 37, in <module>`
`   import random`
` File "/usr/lib64/python2.7/random.py", line 49, in <module>`
`   import hashlib as _hashlib`
` File "/usr/lib64/python2.7/hashlib.py", line 116, in <module>`
`   import _hashlib`
`ImportError: /usr/lib64/libssl.so.10: symbol private_ossl_minimum_dh_bits, version libcrypto.so.10 not defined in file libcrypto.so.10 with link time reference`
`Failed to load "cocotb"`
`
I have Centos 7 and Cadence Xcelium simulator (new version of Cadence IUS). 

As far as I understand it requires earlier version for openssl. My version 1.0.2k and in /lib64/ libssl.so.10 and libcrypto.so.10 symlinks links on 1.0.2k version files. When I change simlinks on previous version - 0.9.8e, I got next error( but simulation was started):
`Traceback (most recent call last):`
` File "/usr/libexec/abrt-action-ureport", line 11, in <module>`
`   import augeas`
` File "/usr/lib/python2.7/site-packages/augeas.py", line 43, in <module>`
`   import ctypes.util`
` File "/usr/lib64/python2.7/ctypes/util.py", line 89, in <module>`
`   import re, tempfile, errno`
` File "/usr/lib64/python2.7/tempfile.py", line 35, in <module>`
`   from random import Random as _Random`
` File "/usr/lib64/python2.7/random.py", line 49, in <module>`
`   import hashlib as _hashlib`
` File "/usr/lib64/python2.7/hashlib.py", line 116, in <module>`
`   import _hashlib`
`ImportError: /usr/lib64/python2.7/lib-dynload/_hashlib.so: symbol HMAC_CTX_copy, version libcrypto.so.10 not defined in file libcrypto.so.10 with link time reference`

jrpetrus:
I upgraded to Xcelium and get this same error from `xrun` when loading my DPI-C code that links in Octave 4.0.3 routines I wrote.
```
xmsim: *W,LIBRUN: Could not load the dynamic library:  ./xcelium.d/tb_default_opt.lnx8664.18.03.d/librunpost
[...]

Issue #2016, opened 2020-08-14
universalize debug target


ktbarrett:
The `debug` target is seen in a couple simulator's makefiles (but not all), and only seems to run `gdb`. Now that we have `COCOTB_ATTACH`, running the simulator by calling gdb is unnecessary. However, running simulators with valgrind is a great way to help diagnose safety issues and you must run the simulator by calling valgrind. There may be similar debugging tools on Windows.

Instead of calling gdb or valgrind directly, I advocate that we make the process that executes the simulator a make variable `COCOTB_DEBUG_CMD`. The variable will default to gdb. The debug target will concatenate `COCOTB_DEBUG_CMD` and the rest of the simulator command. Simple concatenation should support gdb, valgrind, and similar tools directly. Other tools could be supported by creating a wrapper script and using that as the debug command.

Additionally, I would add the debug target to all simulators' makefiles.

ktbarrett:
Closed by #2615.

PR #2615, opened 2021-07-08
Add $(CMD) prefixing with SIM_CMD_PREFIX

martijnbastiaan:
Can be used to add environment variables or prefixed commands to
invocations of $(CMD). This feature has been motivated by issue #644 to
work around issues with conflicting dynamic libraries. In these
situations users can set `SIM_CMD_PREFIX := LD_PRELOAD="foo.so bar.so"` to
force a particular library.


ktbarrett (commented):
I think this is generally useful, but could be even more useful. I think this variable should appear just before the `$(CMD)`. It can be used for `LD_PRELOAD`; `LD_DEBUG`; `LD_LIBRARY_PATH`, `PYTHONPATH` and `PATH`; but can even be used for prepending `gdb`, `valgrind`, `perf`, etc. expanding on and nullifying the `debug` target. I proposed this in #2016.

Thoughts?

martijnbastiaan:
Hmm, I'm not sure. On the one hand, having `COCOTB_DEBUG_CMD` would solve my issue related to having to instert `LD_PRELOAD` in order for my Xcelium installation to properly work in combination with cocotb. On the other hand, it would feel a bit like abusing the variable as the name implies it's used for debugging.

On the position of `SIM_ENV`: I purposely moved it as far forward as possible to make sure it wouldn't override any environment variables set by cocotb (although I see I failed for the very first file listed in this PR :see_no_evil:). 

My proposal would be to implement both:

1. `SIM_ENV` as far forward as possible
2. `COCOTB_DEBUG_CMD` right before `$(CMD)`

I'm happy to take on that task. If implementing both doesn't make sense to you, let's go with (2).

themperek:
This cannot be done by adding `export  LD_PRELOAD="foo.so bar.so"` in user Makefile?

martijnbastiaan:
That gives me a slew of warnings of all other commands now running with `LD_PRELOAD`. When it reaches the Xcelium invocation it fails in a way as if it wasn't given the `LD_PRELOAD`. Ideally, I'd pass it to Xcelium and Xcelium only.

ktbarrett:
> On the other hand, it would feel a bit like abusing the variable as the name implies it's used for debugging.

Feel free to pick a different name. `CMD_PREFIX`, `PRECMD`, etc. 

>  If implementing both doesn't make sense to you, let's go with (2).

I think the makefiles are too knob-happy as it stands. If this can be one feature (and I think it can), it should be.

martijnbastiaan:
`CMD_PREFIX` sounds good to me. Will implement!

documentation/source/building.rst
```diff
@@ -292,6 +292,18 @@ The following variables are makefile variables, not environment variables.
...
+
+      Prefix added to invocations of $(CMD)
+
+      Can be used to add environment variables or prefixed commands to invocations of
+      $(CMD). This feature has been motivated by `issue 644 <https://github.com/cocotb/cocotb/issues/644>`_
+      to work work around issues with conflicting dynamic libraries. In these situations
+      users can set ``CMD_PREFIX := LD_PRELOAD="foo.so bar.so"`` to force a particular
+      library.
+
+      .. versionadded:: 1.5.3
```
martijnbastiaan, line 305: I've set it to `1.5.3`, but we might want to wait on a major release?

martijnbastiaan:
@ktbarrett I've updated the PR. It implements `CMD_PREFIX` now.

documentation/source/building.rst
ktbarrett:
```suggestion
      .. versionadded:: 1.6.0
```

The master branch is for the next feature release. If this were to go into a patch release this commit would need to be cherry-picked into the `stable/1.5` branch. Since this is a new non-breaking feature and not a bugfix, it should go in the next feature release.

documentation/source/building.rst
```diff
@@ -292,6 +292,18 @@ The following variables are makefile variables, not environment variables.
...
 
+.. make:var:: CMD_PREFIX
+
+      Prefix added to invocations of $(CMD)
+
+      Can be used to add environment variables or prefixed commands to invocations of
+      $(CMD). This feature has been motivated by `issue 644 <https://github.com/cocotb/cocotb/issues/644>`_
+      to work work around issues with conflicting dynamic libraries. In these situations
+      users can set ``CMD_PREFIX := LD_PRELOAD="foo.so bar.so"`` to force a particular
+      library.
```
ktbarrett, line 303: Convention is 1 sentence per line and I don't think you should mention motivation in reference documentation.

```suggestion
      Can be used to add environment variables or prefixed commands to invocations of $(CMD).
      For example, ``CMD_PREFIX='LD_PRELOAD="foo.so bar.so"'`` can be used to force a particular library to load.
      Or, ``CMD_PREFIX="gdb --args"`` to run the simulation with the GDB debugger.
```

cocotb/share/makefiles/simulators/Makefile.ghdl
```diff
@@ -68,15 +68,15 @@ endif
 # Compilation phase
 analyse: $(VHDL_SOURCES) $(CUSTOM_COMPILE_DEPS) | $(SIM_BUILD)
 	$(foreach SOURCES_VAR, $(filter VHDL_SOURCES_%, $(.VARIABLES)), \
-		$(CMD) -i $(GHDL_ARGS) $(COMPILE_ARGS) --workdir=$(SIM_BUILD) --work=$(SOURCES_VAR:VHDL_SOURCES_%=%) $($(SOURCES_VAR)) && ) \
-	$(CMD) -i $(GHDL_ARGS) $(COMPILE_ARGS) --workdir=$(SIM_BUILD) --work=$(RTL_LIBRARY) $(VHDL_SOURCES) && \
-	$(CMD) -m $(GHDL_ARGS) $(COMPILE_ARGS) --workdir=$(SIM_BUILD) -P$(SIM_BUILD) --work=$(RTL_LIBRARY) $(TOPLEVEL)
+		$(CMD_PREFIX) $(CMD) -i $(GHDL_ARGS) $(COMPILE_ARGS) --workdir=$(SIM_BUILD) --work=$(SOURCES_VAR:VHDL_SOURCES_%=%) $($(SOURCES_VAR)) && ) \
+	$(CMD_PREFIX) $(CMD) -i $(GHDL_ARGS) $(COMPILE_ARGS) --workdir=$(SIM_BUILD) --work=$(RTL_LIBRARY) $(VHDL_SOURCES) && \
+	$(CMD_PREFIX) $(CMD) -m $(GHDL_ARGS) $(COMPILE_ARGS) --workdir=$(SIM_BUILD) -P$(SIM_BUILD) --work=$(RTL_LIBRARY) $(TOPLEVEL)
```
ktbarrett, line 73: I'm not sure this is necessary. These are just compilation steps, cocotb is never loaded here.

cocotb/share/makefiles/simulators/Makefile.cvc
```diff
@@ -80,7 +80,7 @@ ifeq ($(CVC_ITERP),1)
     debug:  $(CUSTOM_SIM_DEPS)
 	MODULE=$(MODULE) \
         TESTCASE=$(TESTCASE) TOPLEVEL=$(TOPLEVEL) TOPLEVEL_LANG=$(TOPLEVEL_LANG) \
-        gdb --args $(CMD) $(CVC_ARGS) +acc+2 -o $(SIM_BUILD)/sim.vvp +define+COCOTB_SIM=1 +loadvpi=$(shell cocotb-config --lib-name-path vpi cvc):vlog_startup_routines_bootstrap $(COMPILE_ARGS) $(EXTRA_ARGS) $(VERILOG_SOURCES)
+        gdb --args $(CMD_PREFIX) $(CMD) $(CVC_ARGS) +acc+2 -o $(SIM_BUILD)/sim.vvp +define+COCOTB_SIM=1 +loadvpi=$(shell cocotb-config --lib-name-path vpi cvc):vlog_startup_routines_bootstrap $(COMPILE_ARGS) $(EXTRA_ARGS) $(VERILOG_SOURCES)
```
ktbarrett, line 83:
```suggestion
         $(CMD_PREFIX) gdb --args $(CMD) $(CVC_ARGS) +acc+2 -o $(SIM_BUILD)/sim.vvp +define+COCOTB_SIM=1 +loadvpi=$(shell cocotb-config --lib-name-path vpi cvc):vlog_startup_routines_bootstrap $(COMPILE_ARGS) $(EXTRA_ARGS) $(VERILOG_SOURCES)
```
I don't think GDB allows you to add environment variables before commands like this. It needs to be before the call to `gdb`.

cocotb/share/makefiles/simulators/Makefile.icarus
```diff
@@ -67,7 +67,7 @@ COMPILE_ARGS += -f $(SIM_BUILD)/cmds.f -g2012 # Default to latest SystemVerilog
 # Compilation phase
 $(SIM_BUILD)/sim.vvp: $(VERILOG_SOURCES) $(CUSTOM_COMPILE_DEPS) | $(SIM_BUILD)
 	@echo "+timescale+$(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION)" > $(SIM_BUILD)/cmds.f
-	$(CMD) -o $(SIM_BUILD)/sim.vvp -D COCOTB_SIM=1 $(TOPMODULE_ARG) $(COMPILE_ARGS) $(EXTRA_ARGS) $(VERILOG_SOURCES)
+	$(CMD_PREFIX) $(CMD) -o $(SIM_BUILD)/sim.vvp -D COCOTB_SIM=1 $(TOPMODULE_ARG) $(COMPILE_ARGS) $(EXTRA_ARGS) $(VERILOG_SOURCES)
```
ktbarrett, line 70: Icarus also has a debug target which should get the `$(CMD_PREFIX)`.

martijnbastiaan: Yeah. And this `CMD_PREFIX` was added to the compilation phase too.

This makefile is a bit weird in the sense that `CMD_PREFIX` now _doesn't_ prefix `$(CMD)`. So perhaps `CMD_PREFIX` is still a misnomer.

martijnbastiaan:
I've applied your suggestions @ktbarrett. `CMD_PREFIX` might still be a misnomer (see inline comment). Perhaps `SIM_PREFIX`?

ktbarrett:
@martijnbastiaan It's probably fine as is, names are not as important if there is documentation. But if you want to change it, maybe `SIM_CMD_PREFIX`, `SIM_PREFIX` doesn't make it clear its for the `CMD`. Thoughts?

ktbarrett:
And because of your other PR, you will need to rebase =)

martijnbastiaan:
`SIM_CMD_PREFIX` it is. I'd have to look at my own misnomer for years to come mind you ;-). I'll push the change tomorrow.

ktbarrett (approved):

ktbarrett:
I usually like to leave it so others can take a look before merging. So it will get merged in a couple days.

documentation/source/newsfragments/2615.feature.rst
```diff
@@ -0,0 +1 @@
+Added `SIM_CMD_PREFIX` to supported Makefile variables, allowing users to pass environment variables and other command prefixes to simulators.
```
cmarqu, line 1:
```suggestion
Add the :make:var:`SIM_CMD_PREFIX` to supported Makefile variables, allowing users to pass environment variables and other command prefixes to simulators.
```

documentation/source/building.rst
```diff
@@ -292,6 +292,16 @@ The following variables are makefile variables, not environment variables.
       to set the random seed value if :envvar:`RANDOM_SEED` is not set.
       If both ``+ntb_random_seed`` and ``+seed`` are set, ``+ntb_random_seed`` is used.
 
+.. make:var:: SIM_CMD_PREFIX
+
+      Prefix added to invocations of $(CMD)
+
+      Can be used to add environment variables or prefixed commands to invocations of $(CMD).
```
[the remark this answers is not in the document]
ktbarrett: Might suggest `Prefix for simulation command invocations.`

cocotb/share/makefiles/Makefile.sim
```diff
@@ -58,6 +58,7 @@ WAVES                     Enable wave traces dump for Riviera-PRO and Questa
 VERILOG_SOURCES           A list of the Verilog source files to include
 VHDL_SOURCES              A list of the VHDL source files to include
 VHDL_SOURCES_<lib>        VHDL source files to include in *lib* (GHDL/ModelSim/Questa/Xcelium only)
+SIM_CMD_PREFIX            Prefix added to invocations of $(CMD)
```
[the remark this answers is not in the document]
ktbarrett: Same here.

ktbarrett:
Not sure how coverage could have changed, so I'll ignore that.

Issue #2564, opened 2021-06-17
Improving cocotb output formatting


ktbarrett:
Following on #2436.

Right now, test outputs looks like this.

![image](https://user-images.githubusercontent.com/22162772/122325831-e686d100-cef0-11eb-92be-689cc863be52.png)

Notice the following:
* Most things aren't highlighted
* Starting the test `Running test: ...` is BG highlighted blue
* `Test Passed` is not highlighted
* `Test Failed` is FG highlighted red
* FAIL line in summary is BG highlighted red (and the coloring includes the newline =/)
* The final summary is also not aligned with the per-test summary

Maybe it's just me, but I think this could be improved. What are your opinions?

ktbarrett:
Another thing that could be improve may be startup and shutdown messages.

cmarqu:
Slightly related: #2320

cmarqu:
Maybe we could collect example output of other programs. AFL is one such example that I like: https://www.youtube.com/watch?v=uj1FTiczJSE&t=820s

ktbarrett:
It seems the failure red BG highlighting extending passed the end of the summary is an issue specific to Windows Terminal and likely the line-endings support.

I propose we merge the summary into the PASS/FAIL table. A final line of the PASS/FAIL table could report totals: Somehting like this:

```
     0.00ns INFO     ********************************************************************************
                     ** TEST                    PASS/FAIL  SIM TIME(NS)  REAL TIME(S)  RATIO(NS/S) **
                     ********************************************************************************
                     ** test_module.test_pass     PASS            0.00          0.00         1.99  **
                     ** test_module.test_fail     FAIL            0.00          0.00         1.99  **
                     **                            1/1            0.00          0.00         0.99  **
                     ********************************************************************************
```

ktbarrett:
Perhaps we can also list skipped tests as `SKIP` rather than `N/A` and count those towards the total?

ktbarrett:
I think between #2618 and #2619, this has been accomplished.

PR #2619, opened 2021-07-15
Make COCOTB_REDUCED_LOG_FMT enabled by default

ktbarrett:
xref #2564. Turns `COCOTB_REDUCED_LOG_FMT` from an existential flag into a boolean option. The option defaults to `1`, so the reduced format is the new standard format, and the old full format can be selected by setting the variable to `0`.

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -40,10 +40,10 @@ from cocotb.utils import (
 
 import cocotb.ANSI as ANSI
 
-if "COCOTB_REDUCED_LOG_FMT" in os.environ:
-    _suppress = True
-else:
-    _suppress = False
+try:
+    _suppress = int(os.environ.get("COCOTB_REDUCED_LOG_FMT", "1"))
+except ValueError:
+    _suppress = 1
 
 # Column alignment
 _LEVEL_CHARS    = len("CRITICAL")  # noqa
--- a/documentation/source/newsfragments/2564.change.rst
+++ b/documentation/source/newsfragments/2564.change.rst
@@ -1,1 +1,1 @@
-Make :envvar:`COCOTB_REDUCED_LOG_FMT` enabled by default.
+Shorter log lines (configurable with :envvar:`COCOTB_REDUCED_LOG_FMT`) are now the default. For wider log output, similar to previous cocotb releases, set the :envvar:`COCOTB_REDUCED_LOG_FMT` environment variable to ``0``.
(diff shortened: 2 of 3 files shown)


cocotb/log.py
```diff
@@ -40,10 +40,10 @@
 
 import cocotb.ANSI as ANSI
 
-if "COCOTB_REDUCED_LOG_FMT" in os.environ:
-    _suppress = True
-else:
-    _suppress = False
+try:
+    _suppress = int(os.environ.get("COCOTB_REDUCED_LOG_FMT", "1"))
```
cmarqu, line 44:
```suggestion
    _suppress = int(os.getenv("COCOTB_REDUCED_LOG_FMT", default="1"))
```
The existing codebase leans toward `getenv` I think.

documentation/source/newsfragments/2564.change.rst
```diff
@@ -0,0 +1 @@
+Make :envvar:`COCOTB_REDUCED_LOG_FMT` enabled by default.
```
cmarqu, line 1:
```suggestion
Shorter log lines (configurable with :envvar:`COCOTB_REDUCED_LOG_FMT`) are now the default.
```

cmarqu (approved):

cocotb/log.py
imphil: No strong feelings wither way, but I would also go for `os.environ.get()` :-)

documentation/source/newsfragments/2564.change.rst
imphil: Can we add a second sentence: "For wider log output, similar to previous cocotb releases, set the COCOTB_REDUCED_LOG_FMT environment variable to 0."

themperek (approved):

imphil (approved):

cocotb/log.py
ktbarrett: And this is why "There should be one &mdash; and preferably only one &mdash; obvious way to do it." =)

documentation/source/newsfragments/2564.change.rst
ktbarrett:
```suggestion
Shorter log lines (configurable with :envvar:`COCOTB_REDUCED_LOG_FMT`) are now the default. For wider log output, similar to previous cocotb releases, set the :envvar:`COCOTB_REDUCED_LOG_FMT` environment variable to ``0``.
```

PR #2620, opened 2021-07-15
Improve regression summary

ktbarrett:
Combines the two regression summaries into a single summary and make a few formatting improvements.

Before:
image

After:
image

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -529,22 +529,25 @@ class RegressionManager:
         summary += "** {a:<{a_len}}  {b:^{b_len}}  {c:>{c_len}}  {d:>{d_len}}  {e:>{e_len}} **\n".format(**header_dict)
         summary += LINE_SEP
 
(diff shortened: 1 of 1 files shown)


cmarqu (approved):

themperek (approved):

eric-wieser:
Can you update the PR description to show a before / after?

ktbarrett:
Added before and after screenshots.

Changes:
* merged the final banner into the previous table
* added total number of tests, passed, failed, and skipped and not just `ERRORS`
* color coded `PASS`, `FAIL` and `SKIP` only instead of highlighting the entire line
* removed the error log message before the summary that prints the number of failed tests (that's now redundant, you can find that info in the table)
* removed the `shutting down...` message that comes after summary, since it's kind of redundant
* lower cased the units in the header

During the meeting we determined that this is human readable info and isn't an "interface", so changes are okay.

imphil (approved):
Looks great, thanks @ktbarrett!

PR #2624, opened 2021-07-17
Add coverage exclusions 

ktbarrett:
Just for `pass` and `...` statements right now. But that could extend so we have less `# pragma: no cover` comments strewn about.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,3 +48,4 @@ omit =
 exclude_lines =
     pragma: no cover
     \.\.\.
+    raise\s+NotImplementedError
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,3 +43,9 @@ testpaths =
 omit =
     cocotb/config.py
     cocotb/_vendor/*
+
+[coverage:report]
+exclude_lines =
+    pragma: no cover
+    pass
+    \.\.\.
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,5 +47,4 @@ omit =
 [coverage:report]
 exclude_lines =
     pragma: no cover
-    pass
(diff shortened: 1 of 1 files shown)


cmarqu (approved):

eric-wieser:
Why is ignoring coverage on `pass` statements a sensible thing to do? Consider:
```
try:
    x = foo()
except:
    pass
return x + y
```
A coverage failure on the `pass` highlights a real missing test that would find a bug

eric-wieser (changes requested):
I think it's better to use the comment explicitly on a case by case basis rather than trying to guess which coverage exemptions don't matter (ie `pass`).

`...` is probably the exception to that rule since that's only for mypy, right?

ktbarrett:
Well the intent is to ignore empty function bodies for type stubs and abstract methods, but regexes aren't the greatest in that respect. By convention we could say that `...` is ignored while `pass` is not?

setup.cfg
```diff
@@ -43,3 +43,9 @@ testpaths =
 omit =
     cocotb/config.py
     cocotb/_vendor/*
+
+[coverage:report]
+exclude_lines =
+    pragma: no cover
+    pass
```
ktbarrett, line 50:
```suggestion
```

eric-wieser (approved):
Thanks - if you can add a comment explaining why ... is there (for overload), that would be great

ktbarrett:
I think I merged this just a little too early... https://github.com/nedbat/coveragepy/issues/1131#issuecomment-884982574. Apparently there is a soon-to-be-documented feature where decorators in `exclude_lines` skip the entire function that are decorated.

Issue #2628, opened 2021-07-18
asyncio.queues.QueueFull?


raysalemi:
Would it make sense to create cocotb.queues.QueueFull by extending asyncio.queues.QueueFull so that the user would not be tempted to import `asyncio`?

eric-wieser:
This seems like a good idea to me. I'm not even convinced that we should extend the asyncio exceptions at all.

@ktbarrett, do you feel strongly about extending the asyncio exceptions vs defining our own? The asyncio defs are just
```
class QueueFull(Exception):
    """Raised when the Queue.put_nowait() method is called on a full Queue."""
    pass
```
so we're not really gaining anything by extending them.

ktbarrett:
I don't really see a reason to extend it. They are available to be imported from `cocotb.queue` already, maybe we just need to document that?

eric-wieser:
The downside is that when you write `cocotb.queue.QueueEmpty` in a console, you get a `repr` containing `asyncio`, which is misleading because nothing else from asyncio will actually work from cocotb.

ktbarrett:
If that's a problem, then we to extend them. It's just a couple lines of code. @raysalemi Could you make a PR with this and the `peek` enhancement (#2626)?

```python
class QueueFull(asyncio.queues.QueueFull):
    """Raised when the Queue.put_nowait() method is called on a full Queue."""

class QueueEmpty(asyncio.queues.QueueEmpty):
    """Raised when the Queue.get_nowait() method is called on a empty Queue."""
```

eric-wieser:
I think we should do this in two PRs, simply because there's a lot of precedent against a `peek` API being pythonic, but the exceptions sound uncontroversial.

raysalemi:
Sorry. I just saw this. I have one PR for both.  Want me to resubmit?

ktbarrett:
Resolved by #2631.

Issue #2626, opened 2021-07-18
Enhancement idea


raysalemi:
Would it make sense to add `peek` and `peek_nowait` to the cocotb.queues.Queue`?

eric-wieser:
Can you give an example of an application where `peek` would be useful? I assume you're thinking of a single-consumer queue, since `peek` is useless on a multi-consumer queue?

Note some other Asyncio libraries have rejected this kind of feature request, eg https://github.com/aio-libs/janus/issues/67

eric-wieser:
It's quite straightforward to build what you're asking for as a standalone component
```python
class InspectableQueueConsumer:
    """ This object is not safe to use from multiple coroutines """
    def __init__(self, queue):
        self.queue = queue
        self._peeked = []

    async def get(self):
        try:
            return self._peeked.pop(0)
        except Indexerror:
            return await self.queue.get()

    async def peek(self):
        if not self.peeked:
            self.peeked.append(await self.queue.get())
        return self.peeked[0]
```

raysalemi:
> Can you give an example of an application where `peek` would be useful? I assume you're thinking of a single-consumer queue, since `peek` is somewhere between useless and dangerous on a multi-consumer queue?
> 
> Note some other python libraries and even python itself have rejected this kind of feature request, eg [aio-libs/janus#67](https://github.com/aio-libs/janus/issues/67) and https://groups.google.com/g/comp.lang.python/c/ujy2LIHUy3o and https://bugs.python.org/issue35118

I'm modifying `pyuvm` to use coroutines instead of threads as @ktbarrett suggested.  The UVM TLM FIFO has a peek function and I have extended Queue to provide it to myself (as you suggested).  

I must admit, I have never used peek in anger, but I'm following the spec.

raysalemi:
> It's quite straightforward to build what you're asking for as a standalone component
> 
> ```python
> class InspectableQueueConsumer:
>     """ This object is not safe to use from multiple coroutines """
>     def __init__(self, queue):
>         self.queue = queue
>         self._peeked = []
> 
>     async def get(self):
>         try:
>             return self._peeked.pop(0)
>         except Indexerror:
>             return await self.queue.get()
> 
>     async def peek(self):
>         if not self.peeked:
>             self.peeked.append(await self.queue.get())
>         return self.peeked[0]
> ```

I just modified the `get` pattern in the current code:

```python
 def _peek(self):
        return self._queue[0]

    async def peek(self):
[...]

PR #2629, opened 2021-07-18
Now with QueueFull and QueueEmpty exceptions and peek functions in Queue

raysalemi:
This pull request contains two minor enhancements to queues:
#2626 
* It adds a peek function to the queue both the coroutine `peek()` and function `peek_nowait()`
* It rolls tests for peek into existing tests

#2628 
Now extends asyncio.queues.QueueFull and asyncio.queues.QueueEmpty to create cocotb.queue versions

### DIFF
--- a/cocotb/queue.py
+++ b/cocotb/queue.py
@@ -4,10 +4,16 @@
 from typing import Generic, TypeVar
 import collections
 import heapq
-from asyncio import QueueEmpty, QueueFull
+import asyncio.queues
 
 import cocotb
 from cocotb.triggers import Event, _pointer_str
+
+class QueueFull(asyncio.queues.QueueFull):
+    """Raised when the Queue.put_nowait() method is called on a full Queue."""
+
+class QueueEmpty(asyncio.queues.QueueEmpty):
+    """Raised when the Queue.get_nowait() or Queue.peek_nowait() method is called on a empty Queue."""
 
 T = TypeVar('T')
 
@@ -35,6 +41,9 @@ class Queue(Generic[T]):
     def _init(self, maxsize):
         self._queue = collections.deque()
 
+    def _peek(self):
+        return self._queue[0]
+        
     def _put(self, item):
         self._queue.append(item)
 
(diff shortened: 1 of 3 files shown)


cocotb/queue.py
```diff
@@ -138,6 +147,28 @@ def get_nowait(self) -> T:
...
+
+    def peek_nowait(self) -> T:
+        """
+        Return the first item in the queue. If the queue is empty
+        raise QueueEmpty()
+        """
+        if self.empty():
+            raise QueueEmpty()
+        item = self._peek()
+        self._wakeup_next(self._putters)
```
eric-wieser, line 169: Why do you wakeup the other events after a peek?

cocotb/queue.py
```diff
@@ -35,6 +41,9 @@ def __init__(self, maxsize: int = 0):
     def _init(self, maxsize):
         self._queue = collections.deque()
 
+    def _peek(self):
+        return self._queue[0]
```
eric-wieser, line 45: This raises the wrong exception when the queue is empty.

eric-wieser (commented):
I remain unconvinced that adding `peek` is a good idea. Yes, UVM has it, but it is not cocotb's job to be UVM, that's what PyUVM is for. It's trivially to write your own subclass of the queue that operates more in line with UVM semantics.

cocotb/queue.py
raysalemi: Monkey see, monkey do. That shouldn't be in there.

raysalemi:
> I remain unconvinced that adding `peek` is a good idea. Yes, UVM has it, but it is not cocotb's job to be UVM, that's what PyUVM is for. It's trivially to write your own subclass of the queue that operates more in line with UVM semantics.

I've already implemented it in pyuvm. I think that you make a reasonable argument against it.

cocotb/queue.py
raysalemi: Let's reject this pull request.  I'll put in another one with only the exceptions.

cocotb/queue.py
raysalemi: I've removed this code from a new pull request I"m entering

raysalemi:
I'll enter w new pull request with only the exceptions.

PR #2634, opened 2021-07-20
Python based runner

themperek:
Python-based runner based on https://github.com/themperek/cocotb-test/blob/master/cocotb_test/simulator.py see also: [cocotb-test slides](https://github.com/cocotb/cocotb/files/3908378/cocotb-test_WOSDV2019.pdf)

This is just something we can start with.

Trying to follow: https://github.com/cocotb/cocotb/wiki/Python-Test-Runner-Proposal

Issues:
- some simulators (GHDL, Verilator, maybe also Icarus) require `top HDL module name` for build/compilations
- some simulators (Verilator, maybe Icarus) require `wave` for build/compilations

### DIFF
--- a/cocotb/runner.py
+++ b/cocotb/runner.py
@@ -2,30 +2,34 @@
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
 
-
-import subprocess
+import abc
 import os
-import sys
-import tempfile
 import re
 import shutil
-import warnings
-from xml.etree import cElementTree as ET
-import threading
-import signal
-import cocotb.config
+import subprocess
+import sys
 import sysconfig
-from distutils.spawn import find_executable
+import tempfile
+import warnings
+from typing import Dict, List, Mapping, Optional, Sequence, Type, Union
+from xml.etree import cElementTree as ET
+
+import cocotb.config
+
+PathLike = Union["os.PathLike[str]", str]
+Command = List[str]
 
 warnings.warn(
-    "Python runners and associated APIs are an experimental feature and subject to change.", UserWarning, stacklevel=2
+    "Python runners and associated APIs are an experimental feature and subject to change.",
+    UserWarning,
+    stacklevel=2,
 )
 
 _magic_re = re.compile(r"([\\{}])")
 _space_re = re.compile(r"([\s])", re.ASCII)
 
 
-def as_tcl_value(value):
+def as_tcl_value(value: str) -> str:
     # add '\' before special characters and spaces
     value = _magic_re.sub(r"\\\1", value)
     value = value.replace("\n", r"\n")
@@ -36,26 +40,15 @@ def as_tcl_value(value):
     return value
 
 
-class Simulator(object):
-    def __init__(self):
+class Simulator(abc.ABC):
+    def __init__(self) -> None:
 
         self.check_simulator()
 
-        self.env = {}
+        self.env: Dict[str, str] = {}
 
-        # Catch SIGINT and SIGTERM
-        self.old_sigint_h = signal.getsignal(signal.SIGINT)
-        self.old_sigterm_h = signal.getsignal(signal.SIGTERM)
-
-        # works only if main thread
-        if threading.current_thread() is threading.main_thread():
-            signal.signal(signal.SIGINT, self.exit_gracefully)
-            signal.signal(signal.SIGTERM, self.exit_gracefully)
-
-        self.process = None
-
-    @staticmethod
-    def check_simulator():
+    @abc.abstractmethod
+    def check_simulator(self) -> None:
         raise NotImplementedError()
 
     def init_dir(self, build_dir, work_dir):
@@ -70,7 +63,7 @@ class Simulator(object):
             if os.path.isdir(absworkdir):
                 self.work_dir = absworkdir
 
-    def set_env(self):
+    def set_env(self) -> None:
 
         for e in os.environ:
             self.env[e] = os.environ[e]
@@ -92,25 +85,27 @@ class Simulator(object):
         if not os.path.exists(self.build_dir):
             os.makedirs(self.build_dir)
 
-    def build_command(self):
+    @abc.abstractmethod
+    def build_command(self) -> Sequence[Command]:
         raise NotImplementedError()
 
-    def test_command(self):
+    @abc.abstractmethod
+    def test_command(self) -> Sequence[Command]:
         raise NotImplementedError()
 
     def build(
         self,
-        library_name="work",
-        verilog_sources=[],
-        vhdl_sources=[],
-        includes=[],
-        defines=[],
-        parameters={},
-        extra_args=[],
-        hdl_topmodule=None,
-        always=False,
-        build_dir="sim_build",
-        work_dir=None,
+        library_name: str = "work",
+        verilog_sources: Sequence[PathLike] = [],
+        vhdl_sources: Sequence[PathLike] = [],
+        includes: Sequence[PathLike] = [],
+        defines: Sequence[str] = [],
+        parameters: Mapping[str, object] = {},
+        extra_args: Sequence[str] = [],
+        hdl_topmodule: Optional[str] = None,
+        always: bool = False,
+        build_dir: PathLike = "sim_build",
+        work_dir: Optional[PathLike] = None,
     ):
         self.init_dir(build_dir, work_dir)
         os.makedirs(self.build_dir, exist_ok=True)
@@ -120,12 +115,12 @@ class Simulator(object):
         # a better docstring than using `None` as a default in the parameters
         # list.
         self.library_name = library_name
-        self.verilog_sources = self.get_abs_paths(verilog_sources)
-        self.vhdl_sources = self.get_abs_paths(vhdl_sources)
-        self.includes = self.get_abs_paths(includes)
+        self.verilog_sources = get_abs_paths(verilog_sources)
+        self.vhdl_sources = get_abs_paths(vhdl_sources)
+        self.includes = get_abs_paths(includes)
         self.defines = list(defines)
         self.parameters = dict(parameters)
-        self.compile_args = extra_args
+        self.compile_args = list(extra_args)
         self.always = always
         self.hdl_topmodule = hdl_topmodule
 
@@ -137,19 +132,19 @@ class Simulator(object):
 
     def test(
         self,
-        py_module,
-        hdl_topmodule,
-        toplevel_lang="verilog",
-        testcase=None,
-        seed=None,
-        python_search=[],
-        extra_args=[],
-        plus_args=[],
-        extra_env=None,
-        waves=None,
-        gui=None,
-        build_dir=None,
-        work_dir=None,
+        py_module: Union[str, Sequence[str]],
+        hdl_topmodule: str,
+        toplevel_lang: str = "verilog",
+        testcase: Optional[str] = None,
+        seed: Optional[Union[str, int]] = None,
+        python_search: Sequence[PathLike] = [],
+        extra_args: Sequence[str] = [],
+        plus_args: Sequence[str] = [],
+        extra_env: Mapping[str, str] = {},
+        waves: Optional[bool] = None,
+        gui: Optional[bool] = None,
+        build_dir: PathLike = "sim_build",
+        work_dir: Optional[PathLike] = None,
     ):
 
         self.init_dir(build_dir, work_dir)
@@ -166,9 +161,9 @@ class Simulator(object):
         self.python_search = list(python_search)
         self.hdl_topmodule = hdl_topmodule
         self.toplevel_lang = toplevel_lang
-        self.sim_args = extra_args
+        self.sim_args = list(extra_args)
         self.plus_args = list(plus_args)
-        self.env = dict(extra_env) if extra_env is not None else {}
+        self.env = dict(extra_env)
 
         if testcase is not None:
             self.env["TESTCASE"] = testcase
@@ -186,7 +181,9 @@ class Simulator(object):
         else:
             self.gui = bool(gui)
 
-        results_xml_file = os.getenv("COCOTB_RESULTS_FILE", os.path.join(self.build_dir, "results.xml"))
+        results_xml_file = os.getenv(
+            "COCOTB_RESULTS_FILE", os.path.join(self.build_dir, "results.xml")
+        )
 
         try:
             os.remove(results_xml_file)
@@ -197,127 +194,121 @@ class Simulator(object):
         self.set_env()
         self.execute(cmds)
 
-        self.check_results_file(results_xml_file)
+        check_results_file(results_xml_file)
 
         print("INFO: Results file: %s" % results_xml_file)
         return results_xml_file
 
-    @staticmethod
-    def check_results_file(results_xml_file):
-
-        __tracebackhide__ = True  # Hide the traceback when using PyTest.
-
-        results_file_exist = os.path.isfile(results_xml_file)
-        if not results_file_exist:
-            raise SystemExit("ERROR: Simulation terminated abnormally. Results file not found.")
-
-        failed = 0
-
-        tree = ET.parse(results_xml_file)
-        for ts in tree.iter("testsuite"):
-            for tc in ts.iter("testcase"):
-                for _ in tc.iter("failure"):
-                    failed += 1
-
-        if failed:
-            raise SystemExit("ERROR: Failed %d tests." % failed)
-
-    @staticmethod
-    def get_include_commands(includes):
+    @abc.abstractmethod
+    def get_include_commands(self, includes: Sequence[str]) -> List[str]:
         raise NotImplementedError()
 
-    @staticmethod
-    def get_define_commands(defines):
+    @abc.abstractmethod
+    def get_define_commands(self, defines: Sequence[str]) -> List[str]:
         raise NotImplementedError()
 
-    def get_parameter_commands(self, parameters):
+    @abc.abstractmethod
+    def get_parameter_commands(self, parameters: Mapping[str, object]) -> List[str]:
         raise NotImplementedError()
 
-    @staticmethod
-    def get_abs_paths(paths):
-        paths_abs = []
-        for path in paths:
-            if os.path.isabs(path):
-                paths_abs.append(os.path.abspath(path))
-            else:
-                paths_abs.append(os.path.abspath(os.path.join(os.getcwd(), path)))
-
-        return paths_abs
-
-    def execute(self, cmds):
+    def execute(self, cmds: Sequence[Command]) -> None:
 
         __tracebackhide__ = True  # Hide the traceback when using PyTest.
 
         for cmd in cmds:
-            print("INFO: Running command: " + ' "'.join(cmd) + '" in directory:"' + self.work_dir + '"')
+            print(
+                "INFO: Running command: "
+                + ' "'.join(cmd)
+                + '" in directory:"'
+                + self.work_dir
+                + '"'
+            )
 
             # TODO: create at thread to handle stderr and log as error?
             # TODO: log forwarding
 
-            self.process = subprocess.run(cmd, cwd=self.work_dir, env=self.env)
+            process = subprocess.run(cmd, cwd=self.work_dir, env=self.env)
 
-            if self.process.returncode != 0:
+            if process.returncode != 0:
                 raise SystemExit(
-                    "Process '%s' termindated with error %d" % (self.process.args[0], self.process.returncode)
+                    "Process '%s' termindated with error %d"
+                    % (process.args[0], process.returncode)
                 )
 
-            self.process = None
 
-    @staticmethod
-    def outdated(output, dependencies):
+def check_results_file(results_xml_file: PathLike) -> None:
 
-        if not os.path.isfile(output):
-            return True
+    __tracebackhide__ = True  # Hide the traceback when using PyTest.
 
-        output_mtime = os.path.getmtime(output)
+    results_file_exist = os.path.isfile(results_xml_file)
+    if not results_file_exist:
+        raise SystemExit(
+            "ERROR: Simulation terminated abnormally. Results file not found."
+        )
 
-        dep_mtime = 0
-        for file in dependencies:
-            mtime = os.path.getmtime(file)
-            if mtime > dep_mtime:
-                dep_mtime = mtime
+    failed = 0
 
-        if dep_mtime > output_mtime:
-            return True
+    tree = ET.parse(results_xml_file)
+    for ts in tree.iter("testsuite"):
+        for tc in ts.iter("testcase"):
+            for _ in tc.iter("failure"):
+                failed += 1
 
-        return False
+    if failed:
+        raise SystemExit("ERROR: Failed %d tests." % failed)
 
-    def exit_gracefully(self, signum, frame):
-        pid = None
-        if self.process is not None:
-            pid = self.process.pid
-            self.process.stdout.flush()
-            self.process.kill()
-            self.process.wait()
-        # Restore previous handlers
-        signal.signal(signal.SIGINT, self.old_sigint_h)
-        signal.signal(signal.SIGTERM, self.old_sigterm_h)
-        raise SystemExit("Exiting pid: {} with signum: {}".format(str(pid), str(signum)))
+
+def outdated(output: PathLike, dependencies: Sequence[PathLike]) -> bool:
+
+    if not os.path.isfile(output):
+        return True
+
+    output_mtime = os.path.getmtime(output)
+
+    dep_mtime = 0.0
+    for file in dependencies:
+        mtime = os.path.getmtime(file)
+        if mtime > dep_mtime:
+            dep_mtime = mtime
+
+    if dep_mtime > output_mtime:
+        return True
+
+    return False
+
+
+def get_abs_paths(paths: Sequence[PathLike]) -> List[str]:
+    paths_abs: List[str] = []
+    for path in paths:
+        if os.path.isabs(path):
+            paths_abs.append(os.path.abspath(path))
+        else:
+            paths_abs.append(os.path.abspath(os.path.join(os.getcwd(), path)))
+
+    return paths_abs
 
 
 class Icarus(Simulator):
-    def __init__(self, *argv, **kwargs):
-        super().__init__(*argv, **kwargs)
-
     @staticmethod
-    def check_simulator():
-        if find_executable("iverilog") is None:
+    def check_simulator() -> None:
+        if shutil.which("iverilog") is None:
             raise SystemExit("ERROR: iverilog exacutable not found!")
 
     @staticmethod
-    def get_include_commands(includes):
+    def get_include_commands(includes: Sequence[str]) -> List[str]:
         return ["-I" + dir for dir in includes]
 
     @staticmethod
-    def get_define_commands(defines):
+    def get_define_commands(defines: Sequence[str]) -> List[str]:
         return ["-D" + define for define in defines]
 
-    def get_parameter_commands(self, parameters):
+    def get_parameter_commands(self, parameters: Mapping[str, object]) -> List[str]:
         return [
-            "-P" + self.hdl_topmodule + "." + name + "=" + str(value) for name, value in parameters.items()
+            "-P" + self.hdl_topmodule + "." + name + "=" + str(value)
+            for name, value in parameters.items()
         ]
 
-    def test_command(self):
+    def test_command(self) -> List[Command]:
 
         return [
             [
@@ -332,7 +323,7 @@ class Icarus(Simulator):
             + self.plus_args
         ]
 
-    def build_command(self):
+    def build_command(self) -> List[Command]:
 
         if self.vhdl_sources:
             raise ValueError("This simulator does not support VHDL")
@@ -340,7 +331,7 @@ class Icarus(Simulator):
         self.sim_file = os.path.join(self.build_dir, self.library_name + ".vvp")
 
         cmd = []
-        if self.outdated(self.sim_file, self.verilog_sources) or self.always:
+        if outdated(self.sim_file, self.verilog_sources) or self.always:
 
             cmd = [
                 ["iverilog", "-o", self.sim_file, "-D", "COCOTB_SIM=1", "-g2012"]
@@ -359,22 +350,23 @@ class Icarus(Simulator):
 
 class Questa(Simulator):
     @staticmethod
-    def check_simulator():
-        if find_executable("vsim") is None:
+    def check_simulator() -> None:
+        if shutil.which("vsim") is None:
             raise SystemExit("ERROR: vsim exacutable not found!")
 
     @staticmethod
-    def get_include_commands(includes):
+    def get_include_commands(includes: Sequence[str]) -> List[str]:
         return ["+incdir+" + as_tcl_value(dir) for dir in includes]
 
     @staticmethod
-    def get_define_commands(defines):
+    def get_define_commands(defines: Sequence[str]) -> List[str]:
         return ["+define+" + as_tcl_value(define) for define in defines]
 
-    def get_parameter_commands(self, parameters):
+    @staticmethod
+    def get_parameter_commands(parameters: Mapping[str, object]) -> List[str]:
         return ["-g" + name + "=" + str(value) for name, value in parameters.items()]
 
-    def build_command(self):
+    def build_command(self) -> List[Command]:
 
         cmd = []
 
@@ -403,7 +395,7 @@ class Questa(Simulator):
 
         return cmd
 
-    def test_command(self):
+    def test_command(self) -> List[Command]:
 
         cmd = []
 
@@ -419,15 +411,25 @@ class Questa(Simulator):
                 ["vsim"]
                 + ["-gui" if self.gui else "-c"]
                 + ["-onfinish", "stop" if self.gui else "exit"]
-                + ["-foreign", "cocotb_init " + as_tcl_value(cocotb.config.lib_name_path("fli", "questa"))]
+                + [
+                    "-foreign",
+                    "cocotb_init "
+                    + as_tcl_value(cocotb.config.lib_name_path("fli", "questa")),
+                ]
                 + [as_tcl_value(v) for v in self.sim_args]
-                + [as_tcl_value(v) for v in self.get_parameter_commands(self.parameters)]
+                + [
+                    as_tcl_value(v)
+                    for v in self.get_parameter_commands(self.parameters)
+                ]
                 + [as_tcl_value(self.hdl_topmodule)]
                 + ["-do", do_script]
             )
 
             if self.verilog_sources:
-                self.env["GPI_EXTRA"] = cocotb.config.lib_name_path("vpi", "questa") + ":cocotbvpi_entry_point"
+                self.env["GPI_EXTRA"] = (
+                    cocotb.config.lib_name_path("vpi", "questa")
+                    + ":cocotbvpi_entry_point"
+                )
 
         else:
             cmd.append(
@@ -436,56 +438,63 @@ class Questa(Simulator):
                 + ["-onfinish", "stop" if self.gui else "exit"]
                 + ["-pli", as_tcl_value(cocotb.config.lib_name_path("vpi", "questa"))]
                 + [as_tcl_value(v) for v in self.sim_args]
-                + [as_tcl_value(v) for v in self.get_parameter_commands(self.parameters)]
+                + [
+                    as_tcl_value(v)
+                    for v in self.get_parameter_commands(self.parameters)
+                ]
                 + [as_tcl_value(self.hdl_topmodule)]
                 + [as_tcl_value(v) for v in self.plus_args]
                 + ["-do", do_script]
             )
 
             if self.vhdl_sources:
-                self.env["GPI_EXTRA"] = cocotb.config.lib_name_path("fli", "questa") + ":cocotbfli_entry_point"
+                self.env["GPI_EXTRA"] = (
+                    cocotb.config.lib_name_path("fli", "questa")
+                    + ":cocotbfli_entry_point"
+                )
 
         return cmd
 
 
 class Ghdl(Simulator):
     @staticmethod
-    def check_simulator():
-        if find_executable("ghdl") is None:
+    def check_simulator() -> None:
+        if shutil.which("ghdl") is None:
             raise SystemExit("ERROR: ghdl exacutable not found!")
 
     @staticmethod
-    def get_include_commands(includes):
-        include_cmd = []
-        for dir in includes:
-            include_cmd.append("-I")
-            include_cmd.append(dir)
-
-        return include_cmd
+    def get_include_commands(includes: Sequence[str]) -> List[str]:
+        return [f"-I{dir}" for dir in includes]
 
     @staticmethod
-    def get_define_commands(defines):
-        defines_cmd = []
-        for define in defines:
-            defines_cmd.append("-D")
-            defines_cmd.append(define)
+    def get_define_commands(defines: Sequence[str]) -> List[str]:
+        return [f"-D{define}" for define in defines]
 
-    def get_parameter_commands(self, parameters):
+    @staticmethod
+    def get_parameter_commands(parameters):
         return ["-g" + name + "=" + str(value) for name, value in parameters.items()]
 
-    def build_command(self):
+    def build_command(self) -> List[Command]:
 
         if self.verilog_sources:
             raise ValueError("This simulator does not support Verilog")
 
         return [
-            ["ghdl", "-i"] + ["--work=%s" % self.library_name] + self.compile_args + [source_file]
+            ["ghdl", "-i"]
+            + ["--work=%s" % self.library_name]
+            + self.compile_args
+            + [source_file]
             for source_file in self.vhdl_sources
         ]
 
-    def test_command(self):
+    def test_command(self) -> List[Command]:
 
-        cmd_elaborate = ["ghdl", "-m"] + ["--work=%s" % self.library_name] + self.compile_args + [self.hdl_topmodule]
+        cmd_elaborate = (
+            ["ghdl", "-m"]
+            + ["--work=%s" % self.library_name]
+            + self.compile_args
+            + [self.hdl_topmodule]
+        )
 
         cmd = [cmd_elaborate]
         cmd_run = (
@@ -504,42 +513,53 @@ class Ghdl(Simulator):
 
 class Riviera(Simulator):
     @staticmethod
-    def check_simulator():
-        if find_executable("vsimsa") is None:
+    def check_simulator() -> None:
+        if shutil.which("vsimsa") is None:
             raise SystemExit("ERROR: vsimsa exacutable not found!")
 
     @staticmethod
-    def get_include_commands(includes):
+    def get_include_commands(includes: Sequence[str]) -> List[str]:
         return ["+incdir+" + as_tcl_value(dir) for dir in includes]
 
     @staticmethod
-    def get_define_commands(defines):
+    def get_define_commands(defines: Sequence[str]) -> List[str]:
         return ["+define+" + as_tcl_value(define) for define in defines]
 
-    def get_parameter_commands(self, parameters):
+    @staticmethod
+    def get_parameter_commands(parameters: Mapping[str, object]) -> List[str]:
         return ["-g" + name + "=" + str(value) for name, value in parameters.items()]
 
-    def build_command(self):
+    def build_command(self) -> List[Command]:
 
         do_script = "\nonerror {\n quit -code 1 \n} \n"
 
-        out_file = os.path.join(self.build_dir, self.library_name, self.library_name + ".lib")
+        out_file = os.path.join(
+            self.build_dir, self.library_name, self.library_name + ".lib"
+        )
 
-        if self.outdated(out_file, self.verilog_sources + self.vhdl_sources) or self.always:
+        if outdated(out_file, self.verilog_sources + self.vhdl_sources) or self.always:
 
-            do_script += "alib {RTL_LIBRARY} \n".format(RTL_LIBRARY=as_tcl_value(self.library_name))
+            do_script += "alib {RTL_LIBRARY} \n".format(
+                RTL_LIBRARY=as_tcl_value(self.library_name)
+            )
 
             if self.vhdl_sources:
-                do_script += "acom -work {RTL_LIBRARY} {EXTRA_ARGS} {VHDL_SOURCES}\n".format(
-                    RTL_LIBRARY=as_tcl_value(self.library_name),
-                    VHDL_SOURCES=" ".join(as_tcl_value(v) for v in self.vhdl_sources),
-                    EXTRA_ARGS=" ".join(as_tcl_value(v) for v in self.compile_args),
+                do_script += (
+                    "acom -work {RTL_LIBRARY} {EXTRA_ARGS} {VHDL_SOURCES}\n".format(
+                        RTL_LIBRARY=as_tcl_value(self.library_name),
+                        VHDL_SOURCES=" ".join(
+                            as_tcl_value(v) for v in self.vhdl_sources
+                        ),
+                        EXTRA_ARGS=" ".join(as_tcl_value(v) for v in self.compile_args),
+                    )
                 )
 
             if self.verilog_sources:
                 do_script += "alog -work {RTL_LIBRARY} +define+COCOTB_SIM -sv {DEFINES} {INCDIR} {EXTRA_ARGS} {VERILOG_SOURCES} \n".format(
                     RTL_LIBRARY=as_tcl_value(self.library_name),
-                    VERILOG_SOURCES=" ".join(as_tcl_value(v) for v in self.verilog_sources),
+                    VERILOG_SOURCES=" ".join(
+                        as_tcl_value(v) for v in self.verilog_sources
+                    ),
                     DEFINES=" ".join(self.get_define_commands(self.defines)),
                     INCDIR=" ".join(self.get_include_commands(self.includes)),
                     EXTRA_ARGS=" ".join(as_tcl_value(v) for v in self.compile_args),
@@ -553,7 +573,7 @@ class Riviera(Simulator):
 
         return [["vsimsa"] + ["-do"] + ["do"] + [do_file.name]]
 
-    def test_command(self):
+    def test_command(self) -> List[Command]:
 
         do_script = "\nonerror {\n quit -code 1 \n} \n"
 
@@ -562,24 +582,34 @@ class Riviera(Simulator):
                 TOPLEVEL=as_tcl_value(self.hdl_topmodule),
                 EXT_NAME=as_tcl_value(cocotb.config.lib_name_path("vhpi", "riviera")),
                 EXTRA_ARGS=" ".join(
-                    as_tcl_value(v) for v in (self.sim_args + self.get_parameter_commands(self.parameters))
+                    as_tcl_value(v)
+                    for v in (
+                        self.sim_args + self.get_parameter_commands(self.parameters)
+                    )
                 ),
             )
             if self.verilog_sources:
-                self.env["GPI_EXTRA"] = cocotb.config.lib_name_path("vpi", "riviera") + "cocotbvpi_entry_point"
-        else:
-            do_script += (
-                "asim +access +w -interceptcoutput -O2 -pli {EXT_NAME} {EXTRA_ARGS} {TOPLEVEL} {PLUS_ARGS} \n".format(
-                    TOPLEVEL=as_tcl_value(self.hdl_topmodule),
-                    EXT_NAME=as_tcl_value(cocotb.config.lib_name_path("vpi", "riviera")),
-                    EXTRA_ARGS=" ".join(
-                        as_tcl_value(v) for v in (self.sim_args + self.get_parameter_commands(self.parameters))
-                    ),
-                    PLUS_ARGS=" ".join(as_tcl_value(v) for v in self.plus_args),
+                self.env["GPI_EXTRA"] = (
+                    cocotb.config.lib_name_path("vpi", "riviera")
+                    + "cocotbvpi_entry_point"
                 )
+        else:
+            do_script += "asim +access +w -interceptcoutput -O2 -pli {EXT_NAME} {EXTRA_ARGS} {TOPLEVEL} {PLUS_ARGS} \n".format(
+                TOPLEVEL=as_tcl_value(self.hdl_topmodule),
+                EXT_NAME=as_tcl_value(cocotb.config.lib_name_path("vpi", "riviera")),
+                EXTRA_ARGS=" ".join(
+                    as_tcl_value(v)
+                    for v in (
+                        self.sim_args + self.get_parameter_commands(self.parameters)
+                    )
+                ),
+                PLUS_ARGS=" ".join(as_tcl_value(v) for v in self.plus_args),
             )
             if self.vhdl_sources:
-                self.env["GPI_EXTRA"] = cocotb.config.lib_name_path("vhpi", "riviera") + ":cocotbvhpi_entry_point"
+                self.env["GPI_EXTRA"] = (
+                    cocotb.config.lib_name_path("vhpi", "riviera")
+                    + ":cocotbvhpi_entry_point"
+                )
 
         if self.waves:
             do_script += "log -recursive /*;"
@@ -594,38 +624,44 @@ class Riviera(Simulator):
 
 
 class Verilator(Simulator):
-    def __init__(self, *argv, **kwargs):
-        super().__init__(*argv, **kwargs)
-
     @staticmethod
-    def check_simulator():
-        if find_executable("verilator") is None:
+    def check_simulator() -> None:
+        if shutil.which("verilator") is None:
             raise SystemExit("ERROR: verilator exacutable not found!")
 
     @staticmethod
-    def get_include_commands(includes):
+    def get_include_commands(includes: Sequence[str]) -> List[str]:
         return ["-I" + dir for dir in includes]
 
     @staticmethod
-    def get_define_commands(defines):
+    def get_define_commands(defines: Sequence[str]) -> List[str]:
         return ["-D" + define for define in defines]
 
-    def get_parameter_commands(self, parameters):
+    @staticmethod
+    def get_parameter_commands(parameters: Mapping[str, object]) -> List[str]:
         return ["-G" + name + "=" + str(value) for name, value in parameters.items()]
 
-    def build_command(self):
+    def build_command(self) -> List[Command]:
 
         if self.vhdl_sources:
             raise ValueError("This simulator does not support VHDL")
 
         if self.hdl_topmodule is None:
-            raise ValueError("This simulator requires hdl_topmodule parameter to be specified")
+            raise ValueError(
+                "This simulator requires hdl_topmodule parameter to be specified"
+            )
 
         cmd = []
 
-        verilator_cpp = os.path.join(os.path.dirname(cocotb.__file__), "share", "lib", "verilator", "verilator.cpp")
+        verilator_cpp = os.path.join(
+            os.path.dirname(cocotb.__file__),
+            "share",
+            "lib",
+            "verilator",
+            "verilator.cpp",
+        )
 
-        verilator_exec = find_executable("verilator")
+        verilator_exec = shutil.which("verilator")
 
         cmd.append(
             [
@@ -645,7 +681,9 @@ class Verilator(Simulator):
                 "-o",
                 self.hdl_topmodule,
                 "-LDFLAGS",
-                "-Wl,-rpath,{LIB_DIR} -L{LIB_DIR} -lcocotbvpi_verilator".format(LIB_DIR=cocotb.config.libs_dir),
+                "-Wl,-rpath,{LIB_DIR} -L{LIB_DIR} -lcocotbvpi_verilator".format(
+                    LIB_DIR=cocotb.config.libs_dir
+                ),
             ]
             + self.compile_args
             + self.get_define_commands(self.defines)
@@ -659,15 +697,15 @@ class Verilator(Simulator):
 
         return cmd
 
-    def test_command(self):
+    def test_command(self) -> List[Command]:
         out_file = os.path.join(self.build_dir, self.hdl_topmodule)
         return [[out_file] + self.plus_args]
 
 
-def get_runner(simulator_name):
+def get_runner(simulator_name: str) -> Type[Simulator]:
 
     sim_name = simulator_name.lower()
-    supported_sims = {
+    supported_sims: Dict[str, Type[Simulator]] = {
         "icarus": Icarus,
         "questa": Questa,
         "ghdl": Ghdl,
@@ -677,13 +715,15 @@ def get_runner(simulator_name):
     try:
         return supported_sims[sim_name]
     except KeyError:
-        raise NotImplementedError("Set SIM variable. Supported: " + ", ".join(supported_sims)) from None
+        raise NotImplementedError(
+            "Set SIM variable. Supported: " + ", ".join(supported_sims)
+        ) from None
 
 
-def clean(recursive=False):
+def clean(recursive: bool = False) -> None:
     dir = os.getcwd()
 
-    def rm_clean():
+    def rm_clean() -> None:
         build_dir = os.path.join(dir, "sim_build")
         if os.path.isdir(build_dir):
             print("INFO: Removing:", build_dir)
--- a/cocotb/runner.py
+++ b/cocotb/runner.py
@@ -166,7 +166,7 @@ class Simulator(abc.ABC):
         # list.
         self.python_search = list(python_search)
         self.sim_toplevel = toplevel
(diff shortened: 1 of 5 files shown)


cocotb/run.py
```diff
@@ -0,0 +1,929 @@
...
+
+        if not os.path.exists(self.sim_dir):
+            os.makedirs(self.sim_dir)
+
+    def build_command(self):
+        raise NotImplementedError()
+
+    def run(self):
+
+        sys.tracebacklimit = 0  # remove not needed traceback from assert
```
eric-wieser, line 193: This is probably a bad idea - if you want to exit without a traceback, use `raise SystemExit(message)`

cocotb/run.py
```diff
@@ -0,0 +1,929 @@
...
+
+        self.sim_args = sim_args + extra_args
+
+        if plus_args is None:
+            plus_args = []
+
+        self.plus_args = plus_args
+        self.force_compile = force_compile
+        self.compile_only = compile_only
+
```
eric-wieser, line 139: This avoidance of mutable defaults creates huge additional verbosity and reduces the effectiveness of Sphinx / `help` - I'd usually recommend just making sure to copy all the argument before there's any chance of anyone mutating them:

```suggestion
    def __init__(
        self,
        toplevel,
        module,
        work_dir=None,
        python_search=[],
        toplevel_lang="verilog",
        verilog_sources=[],
        vhdl_sources=[],
        includes=[],
        defines=[],
        parameters={},
        compile_args=[],
        sim_args=[],
        extra_args=[],
        plus_args=[],
        force_compile=False,
        testcase=None,
        sim_build="sim_build",
        seed=None,
        extra_env=None,
        compile_only=False,
        waves=None,
        gui=False
    ):

        self.logger = logging.getLogger()

        self.sim_dir = os.path.abspath(sim_build)
        os.makedirs(self.sim_dir, exist_ok=True)

        self.lib_dir = os.path.join(os.path.dirname(cocotb.__file__), "libs")

        self.lib_ext = "so"
        if os.name == "nt":
            self.lib_ext = "dll"

        if isinstance(module, str):
            self.module = module
        else:
            self.module = ",".join(module)

        self.work_dir = self.sim_dir

        if work_dir is not None:
            absworkdir = os.path.abspath(work_dir)
            if os.path.isdir(absworkdir):
                self.work_dir = absworkdir

        # note: to avoid mutating argument defaults, we ensure that no value
        # is written without a copy. This is much more concise and leads to
        # a better docstring than using `None` as a default in the parameters
        # list.
        self.python_search = list(python_search)
        self.toplevel = toplevel
        self.toplevel_lang = toplevel_lang
        self.verilog_sources = self.get_abs_paths(verilog_sources)
        self.vhdl_sources = self.get_abs_paths(vhdl_sources)
        self.includes = self.get_abs_paths(includes)
        self.defines = list(defines)
        self.parameters = dict(parameters)
        self.compile_args = compile_args + extra_args
        self.sim_args = sim_args + extra_args
        self.plus_args = list(plus_args)
        self.force_compile = force_compile
        self.compile_only = compile_only

```

ktbarrett:
Should add the new files to the [coverage omit list](https://github.com/cocotb/cocotb/blob/d37ecb5d81a2315ff740d318301f814b5a970467/setup.cfg#L43)

eric-wieser:
Shouldn't we instead be measuring coverage for them?

ktbarrett:
@eric-wieser What value would that add and is it worth the effort? Do we intend on getting coverage on simulators we don't have access to via mocking, or do we plan on excluding them? `cocotb/config.py` is excluded, should it instead be tested by the same logic because it contains code used in regressions? We didn't get coverage on the makefiles, is that simply because we couldn't, or because we didn't care?

I don't particularly care about the outcome, I just want consistency and to not have to constantly evaluate our position on every change.

How about we make a discussion to discuss changing our position and follow up by recording that position in `CONTRIBUTING.md` and potentially removing `cocotb/config.py` from the exclusion list, depending on the outcome of that discussion?

cocotb/run.py
```diff
@@ -0,0 +1,929 @@
...
+import logging
+import shutil
+from xml.etree import cElementTree as ET
+import threading
+import signal
+import cocotb._vendor.find_libpython as find_libpython
+import cocotb.config
+
+from distutils.spawn import find_executable
+from distutils.sysconfig import get_config_var
```
ktbarrett, line 20: `distutils` is begin [deprecated](https://www.python.org/dev/peps/pep-0632/). `distutils.sysconfig.get_config_var` can be replace with `sysconfig.get_config_var`. `find_executable` can be replaced with `shutil.which`.

ktbarrett:
I also believe we should support running tests from the CLI and not just pytest.

cocotb/run.py
ktbarrett: `None` is really nice in that wrappers don't have to be aware of, and replicate, the default value. Really, I think this should probably be a configuration dataclass. We can replicate some of the functionality in that dataclass now, then remove it and just use `@dataclass` once we drop Python 3.6 support (it's EOL at the end of this year).

eric-wieser: True, although if they just forward `**kwargs` then they don't need to be aware of the argument at all. No objection to using data classes, although I have no experience with them either.

alexforencich:
I have some ideas for doing that in a "nice" way, I'll work on putting together a toy example in the next couple of weeks when I have some time that supports running from both pytest and from the command line

ktbarrett:
@alexforencich One way I thought of was loading a Python module as a configuration object. Declare your project's sources, toplevel, etc. in that module and specify the module on the command line. Command line args would take precedence.

alexforencich:
My plan is to create a "testbench" object that manages parameters, calling `run()`, and has entry points for `main()` that does command line argument parsing as well as an entry point from python that can be exposed to pytest with decorators, etc.

ktbarrett:
[My proposal for Python test runner requirements and design](https://github.com/cocotb/cocotb/wiki/Python-Test-Runner-Proposal).

themperek:
I have updated mostly according to https://github.com/cocotb/cocotb/wiki/Python-Test-Runner-Proposal . 

Issues/discrepancies:
- some simulators (GHDL, Verilator, maybe also Icarus) require `top HDL module name` for build/compilations
- some simulators (Verilator, maybe Icarus) require `wave` for build/compilations

No idea why Questa-VHDL is failing. It is ok with my setup.

cocotb/runner.py
```diff
@@ -0,0 +1,694 @@
...
+
+        for e in os.environ:
+            self.env[e] = os.environ[e]
+
+        self.env["LIBPYTHON_LOC"] = cocotb._vendor.find_libpython.find_libpython()
+
+        self.env["PATH"] += os.pathsep + cocotb.config.libs_dir
+
+        self.env["PYTHONPATH"] = os.pathsep.join(sys.path)
+        for path in self.python_search:
```
ktbarrett, line 74: Where is this attribute coming from? I don't see it in the `__init__`?

cocotb/runner.py
```diff
@@ -0,0 +1,694 @@
...
+            signal.signal(signal.SIGTERM, self.exit_gracefully)
+
+        self.process = None
+
+    def set_env(self):
+
+        for e in os.environ:
+            self.env[e] = os.environ[e]
+
+        self.env["LIBPYTHON_LOC"] = cocotb._vendor.find_libpython.find_libpython()
```
neelgala, line 69: I have been using cocotb-test with verilator for sometime. In my setup find_libpython did not work properly so I had set the LIBPYTHON_LOC separately as extra_env. This is what was suggested by the error when find_libpython. To get that working, I had to change line-69 above to the following. This should work in a back-compatible way I guess
```
        if 'LIBPYTHON_LOC' not in self.env or self.env['LIBPYTHON_LOC'] is None:
            self.env["LIBPYTHON_LOC"] = find_libpython.find_libpython()
```

cocotb/runner.py
```diff
@@ -0,0 +1,694 @@
...
+            ]
+            + self.compile_args
+            + self.get_define_commands(self.defines)
+            + self.get_include_commands(self.includes)
+            + self.get_parameter_commands(self.parameters)
+            + [verilator_cpp]
+            + self.verilog_sources
+        )
+
+        cmd.append(["make", "-C", self.build_dir, "-f", "Vtop.mk"])
```
neelgala, line 647: in verilator you can set a few options during this make command like: 
```
make OPT_SLOW="-O3" OPT_FAST="-O3" VM_PARALLEL_BUILDS=1 -j8 -C sim_build/ -f Vtop.mk
```

Is there a possibility of exposing these to the user ?

ktbarrett (changes requested):
Not a fan of this implementation. It's essentially using the builder pattern which makes the class fragile due to coupling between the methods. This looks more like a `Project` class and less like a `Simulator` class. Those two concepts need to be decoupled, which is why I suggested the model types in the proposal.

themperek:
> Not a fan of this implementation. It's essentially using the builder pattern which makes the class fragile due to coupling between the methods. This looks more like a `Project` class and less like a `Simulator` class. Those two concepts need to be decoupled, which is why I suggested the model types in the proposal.

@ktbarrett Can you sketch how user API could look like?

themperek:
marlonjames
>Is recompiling only files that have changed (whether in libraries or the design) on subsequent test runs outside the scope of this?

It is there for Icarus. Should be consistently applied (TODO). Questa and Xcelium have an internal mechanism for this as far as I see.

themperek:
@ktbarrett 

> I think the main interface should look like this:

Thank you. Will try to do something. 

In practice (as far as I can see) there is little difference between `Design` and `Library` for simulators that support libraries (GHDL, Questa, Xcelium, Riviera). Not sure about VCS. Maybe can keep just `Library`?

cocotb/runner.py
themperek: https://github.com/cocotb/cocotb/blob/d3b1ec8538a33f706fccced0152de063114d90fa/cocotb/runner.py#L130

Only needed for finding tests. Maybe there is a better way?

cocotb/runner.py
themperek: Makes sense. Will do.

cocotb/runner.py
themperek: For verilator, one can add an extra argument to test like `make_args`. It will break compatibility but ... 🤷🏻

ktbarrett:
@themperek designs have tops while libraries don't. libraries have names while designs don't. Maybe they can be combined and just have that option be contextual? Or should top be removed from design and associated with the test since a design can have potentially many tops?

themperek:
@ktbarrett 
>designs have tops while libraries don't

Would also assume so but in GHDL this seems not to be the case 🤷🏻‍♂️ 

>libraries have names while designs don't.

In principle yes, but all simulators that support libraries (as I see) make top design a library (`work` library). The rest seemed to "compile" everything to some kind of file/executable (can be treated as a library?).

>Or should top be removed from design and associated with the test since a design can have potentially many tops? I'm all for reduction if it makes sense.

This is very annoying. 
The top is used/associated with run/test for Questa, Riviera. 
GHDL, Verilator, Icarus require `top` during compilation 🤷🏻‍♂️ 

As stated before not sure how to deal with this in a nice way.
?

ktbarrett:
@themperek 
>In principle yes, but all simulators that support libraries (as I see) make top design a library (work library). The rest seemed to "compile" everything to some kind of file/executable (can be treated as a library?).

This is a VHDLism. Simulators like Questa and Riviera support Verilog libraries and the final build command does not have a name.

>The top is used/associated with run/test for Questa, Riviera.
GHDL, Verilator, Icarus require top during compilation 🤷🏻‍♂️

It seems that GHDL doesn't support multiple top-like entities or selecting from them. It deduces a singular top. It seems like the best thing to do is support `top` option in both `build` and `test`. Simulators can throw exceptions or ignore options and emit warnings when they don't support a feature.

ktbarrett:
@themperek The top unit in GHDL is the first argument to the elab/run/make commands.

cocotb/runner.py
ktbarrett: @neelgala Please submit an issue on [`find_libpython`](https://github.com/ktbarrett/find_libpython).

cocotb/runner.py
```diff
@@ -0,0 +1,694 @@
...
+        do_file.close()
+
+        return [["vsimsa"] + ["-do"] + ["do"] + [do_file.name]]
+
+
+class Verilator(Simulator):
+    def __init__(self, *argv, **kwargs):
+        super(Verilator, self).__init__(*argv, **kwargs)
+
+        self.env["CXXFLAGS"] = self.env.get("CXXFLAGS", "") + " -std=c++11"
```
ktbarrett, line 579: This is not necessary any longer.

cocotb/runner.py
ktbarrett: I don't see how it will break compatibility? Other simulators would just ignore or throw a warning if the argument is given.

ktbarrett (approved):
The test should change directories to the directory where simulation should take place, likewise with the build directory. Simulators may dump files in the directory they are executed from and parallel executions might not work as expected.

I would pass the build dir and sim dir to `build` and `test` instead of the constructor. I would also not do any environment building in the constructor because the environment may change by the time the commands are called. It would also be a good idea to check the simulator exists in the `__init__` rather than wait until `build` or `test` is called. The simulator is a resource you are acquiring when calling the constructor, it should do those checks then.

themperek:
> The test should change directories to the directory where simulation should take place, likewise with the build directory. Simulators may dump files in the directory they are executed from and parallel executions might not work as expected.

What does this mean in practice?

> I would pass the build dir and sim dir to `build` and `test` instead of the constructor. 

Should be ~done.

>I would also not do any environment building in the constructor because the environment may change by the time the commands are called. 

Should be ~done.

>It would also be a good idea to check the simulator exists in the `__init__` rather than wait until `build` or `test` is called. The simulator is a resource you are acquiring when calling the constructor, it should do those checks then.

Should be ~done.

umarcor:
@ktbarrett pinged me about this issue a couple of months ago. Back then, EDAA was in a very early stage. There is still much to do, but I believe it is now possible to understand the big picture.
Please, bear with me while I provide some feedback in the context of [umarcor.github.io/osvb](https://umarcor.github.io/osvb/) and [edaa-org.github.io](https://edaa-org.github.io/).

The main idea of EDAA is to make modules that EDA tool developers can plug together in order to provide management/building solutions for simulation, co-simulation, synthesis, etc. based on Python.
We don't want to modify the UX of existing projects. Instead, we want to allow tool developers to share internal modules, while each providing a different experience to their user bases.
In order to favour the communication between maintainers, we defined a set of layers, and we applied some colour (see [edaa-org.github.io/ConceptualModel](https://edaa-org.github.io/ConceptualModel.html)):



In that conceptual model, most of the discussion in this issue belongs to layer Workflows (3). You are discussing about how should users interact with the cocotb API (be it CLI, pytest, whatever.; see https://github.com/cocotb/cocotb/wiki/Python-Test-Runner-Proposal#python-api).
That is precisely the specific knowledge that only cocotb maintainers and users can understand.
By the same token, maintainers/users of Edalize, VUnit, pyIPCMI, etc. have particular preferences with regard to the granularity of the tasks and how to keep/pass data across commands/steps/tasks.
EDAA does not want to interfere with those aspects that provide identity to each of the projects.

However, after users tell what they want to achieve (through the CLI/API) and the maintainers understand what they need to execute, all of us need to interact with exactly the same tools. We need to build a project/design (put any name you want for the set of HDL sources, data, constraints, etc.) and we need to call the same CLI tools (because almost no vendor allows interacting with simulators through a shared library).
That's where we are wasting so much resources as a community, and that's what we started with in EDAA.

# Project

pyEDAA.ProjectModel, is an abstract model of HDL projects and EDA tooling. As seen in [pyEDAA.ProjectModel: FileTypes](https://edaa-org.github.io/pyEDAA.ProjectModel/FileTypes/index.html) and [pyEDAA.ProjectModel: pyEDAA.ProjectModel/pyEDAA.ProjectModel](https://edaa-org.github.io/pyEDAA.ProjectModel/pyEDAA.ProjectModel/pyEDAA.ProjectModel.html), the set of classes there is very similar to the Testbench, Design, SourceLibrary and PrecompiledLibrary in https://github.com/cocotb/cocotb/wiki/Python-Test-Runner-Proposal#models and https://github.com/cocotb/cocotb/pull/2634#issuecomment-906099723. See an [example](https://github.com/edaa-org/pyEDAA.ProjectModel#examples) of how to build a pyEDAA.ProjectModel.

*It is to be noted that pyEDAA.ProjectModel uses [pyVHDLModel](https://vhdl.github.io/pyVHDLModel/index.html) and [pySVModel](https://edaa-org.github.io/pySVModel/index.html). Currently, it uses a subset only, for representing libraries and so. However, it is feasible to actually execute a pyVHDLModel frontend (say pyGHDL.dom) to get the full internal syntax model of the VHDL sources. Find practical applications of pyVHDLModel in [OSVB: Project » pyVHDLModelUtils](https://umarcor.github.io/osvb/apis/project/pyVHDLModelUtils.html#api-project-pyvhdlmodelutils). Moreover, in the future it will be possible to use pyVHDLParser (Python only) instead of pyGHDL.dom (which requires libghdl).*

*Moreover, pyEDAA.ProjectModel can read Xilinx's Vivado project files, and OSVVM's `.pro` files.*

*See also pyEDAA.Reports, pyEDAA.UCIS (expected joint work with Aldec and @mballance) and pyEDAA.IPXACT, the "Data" part of [edaa-org.github.io: Languages and Data](https://edaa-org.github.io/SyntaxModels.html).*

*As side note, I think an interesting use case of pyVHDLModel for cocotb is to automatically generate instantiation models (say components/entities) for HDL users to use cocotb verification models as modules in ther HDL testbenches. I.e., treating cocotb as a blackbox from an HDL designer perspective.*

Other tooling solutions do have similar project abstraction classes in Python (see [edaa-org/pyEDAA.ProjectModel: README#references](https://github.com/edaa-org/pyEDAA.ProjectModel#references)). So, I would like to ask the cocotb community to evaluate wheter having a pyEDAA.ProjectModel.cocotb class would be suitable. Note that the class does not need to be located in the ProjectModel repo. It could be located here. The point is whether inheriting from ProjectModel can satisfy the requirements. The main advantage of doing so would be interoperability with other solutions at the cost of having a dependency. On the other hand, it makes sense for pyEDAA.ProjectModel to be co-maintained/co-developed as soon as it becomes a significant piece of some tools.

I did have some brief discussion about it with @ktbarrett at the begining of september. He raised concerns about the (lack of) flexibility due to having explicit classes for each file type/variant. I believe that might be because of the VHDL background and focus of @Paebbels and me. EDAA is expected to support System Verilog and mixed-language designs as well; thus, the creation of pySVModel. However, I acknowledge we are biased towards the very strong and explicit typing in the VHDL community. Nevertheless, see edaa-org/pyEDAA.ProjectModel#7. File attributes allow building functionality on top of existing classes, instead of deriving new classes from there. Such approach is also closer to @olofk's EDAM format.

@ktbarret did also raise concerns about the conception of a Project or a Design. I believe we need to discuss and clarify the terminology. On the one hand, in pyGHDL.dom we use a single Project and Design, because that's a current constraint of how GHDL works. However, ProjectModel should not be constrained by that. On the other hand, ProjectModel does currently support having multiple Designs defined within a Project instance. We need to agree on how to represent TestBenches, whether it needs to be in ProjectModel or in ProjectModel.cocotb.

Please, go ahead and raise your concerns either here, in any of the EDAA repos or in gitter room [hdl/community](https://gitter.im/hdl/community).

# Tool

You might have noted that the layers mentioned so far are all located at the right or top of the Workflow in [edaa-org.github.io/ConceptualModel](https://edaa-org.github.io/ConceptualModel.html).
That is because, taking a bunch of sources and the CLI/API input as a reference, those layers allow raising the abstraction. The concepts of Library, Project, Design, Test(bench|suite), etc. are handled in layers Language (4), Data (5) and Project(6). On the other side, we have layers EDA (2), CLI (1) and Installation (0).

See pyEDAA.CLIAbstraction, pyEDAA.CLITool and pyEDAA.OutputFilter in [edaa-org.github.io: Interacting with CLIs](https://edaa-org.github.io/CLI.html). Similarly to [edaa-org/pyEDAA.ProjectModel: README#references](https://github.com/edaa-org/pyEDAA.ProjectModel#references), there is duplication in the community in this regard. However, I did not gather a list of specific references yet.

There is an additional complexity in these layers: many projects pass all the parameters through a single class/object, which implies implicit knowledge/metadata constrainted to the framework/workflow.
For instance, @olofk is working on branch [esr3](https://github.com/olofk/edalize/commits/esr3) of Edalize, which uses signatures such as `__init__(self, edam=None, work_root=None, eda_api=None, verbose=True)` (see https://github.com/olofk/edalize/commit/21c17c9a4df01729163bba725d17b6396d0ca1f1#). In a similar fashion, @Paebbels' pyIPCMI uses `def __init__(self, host, dryRun, simulationSteps)` (see https://github.com/Paebbels/pyIPCMI/blob/master/pyIPCMI/Simulator/GHDLSimulator.py#L58). VUnit is no exception, where `def _get_command(self, config, output_path, elaborate_only, ghdl_e, wave_file)` is used, because the files are available in `self._project`. The runner proposal in this PR seems to be a more granularly implemented subset of VUnit's approach: https://github.com/cocotb/cocotb/blob/fa7a826cc855d783b7fbc81444d4e6b9edc487b9/cocotb/runner.py#L444-L499.

There is no way we can change that in the short or mid term. Apart from the particular motivations of each project to endorse their own configuration/project format, having it passed implicitly or as a single object is probably the best solution for a monolithic tool (which all of those are). However, most of them are doing something such as https://github.com/VUnit/vunit/blob/96a24b38ba1553f520beb81f26753ca85dafb707/vunit/sim_if/ghdl.py#L272-L322, i.e. creating CLI commands by concating strings and variables converted to string. Actually, pyFPGA and Edalize are slightly different: the former generates Makefiles and the latter uses pre-written Makefiles.

Hence, the purpose of pyEDAA.CLITool is to provide a Python abstraction for each EDA CLI tool, which is agnostic to any particular project/configuration format/object/structure. The purpose is not to provide any homogeneous solution for all EDA tools, but to have a Python interface for each of them. I.e. pyEDAA.CLITool.GHDL abstracts the different backends, usages (simulation, synthesis, linting, etc.). By the same token, pyEDAA.CLITool.Siemens abstracts ModelSim and QuestaSim. How to then make different vendor tools look similar belongs to the Workflows (cocotb, Edalize, VUnit, pyIPCMI, etc.). One may ask why don't vendors provide those Python interfaces themselves? The answer is they are invited to do so, as soon as they accept that the good times for TCL are done.

The case of the Docker CLI is very illustrative. One can interact with Docker from Python by building CLI commands through strings and then using e.g. `check_call` from `subprocess`: https://github.com/hdl/containers/blob/main/utils/pyHDLC/__init__.py#L301-L321. Alternatively, The Docker SDK for Python provides `client.images.build` (see https://docker-py.readthedocs.io/en/stable/images.html). Using the SDK provides little benefit for the most simple commands, but it reduces the cognitive complexity when passing args and reading outputs, while it allows multiple client types.

Therefore, the second proposal I'd like to do is to make pyEDAA.CLITool a community built "EDA SDK for Python". That is, to gather the knowledge about "how to translate from a Python API to a CLI commands, and how to interpret the output of the command". The actual execution of the command, and the management of stdin/stdout/stderr is offloaded to layer 1 CLI in the conceptual model. Actually, one type of CLI client can be "remote execution through SSH" or "execution in a container" (yes, pyEDAA.CLITool might use the Docker SDK through layer 1 and 0).

---

> I also believe we should support running tests from the CLI and not just pytest.

Similarly to the discussions we had about JSON/YAML configuration file format, I think that neither CLI nor pytest should be explicit targets of this implementation. The target should be a Python API that replaces the Makefiles. If that is done correctly, then adding a CLI, pytest, pydoit or any web API is trivial. Conversely, if pytest or a CLI library are used first, there is a risk of unwantedly making a feature dependent or constrained to that library.

As far as I understand, the particularities of cocotb from a project/build/execution point of view are the following:

- A VPI/VHPI lib/module needs to be passed as an argument to the simulation.
- The sources are not only HDL sources, but also include Python scripts.
- Environment variables need to be set for each testbench/testcase.
- Results/reports are written to some location/in some format.

Other than that, cocotb (this proposal) is no different from Edalize, VUnit, OSVVM's TCL plumbing, pyIPCMI, etc. I'd like to raise my concerns about the sense of adding yet another solution, instead of collaborating with any (or all) of those projects to better deal with the four required features:

```
                 ┌─────────┐
                 │ Project │
                 │  Model  │
                 └────┬────┘
 ┌────────┐           │          ┌────────┐
 │ pytest ├───────┐   │    ┌─────► runner │
 └────────┘       │   │    │     └────────┘
                  │   │    │
 ┌────────┐     ┌─▼───▼────┴┐    ┌────────┐
 │  CLI   ├─────►  cocotb   ├────► CLITool│
 └────────┘     │           │    └────────┘
              ┌─►  project  ├─┐
 ┌────────┐   │ └─▲─────┬─┬─┘ │  ┌────────┐
 │ falcon ├───┘   │     │ │   └──► Edalize│
 └────────┘       │     │ │      └────────┘
                  │     │ │
 ┌────────┐       │     │ │      ┌────────┐
 │  ...   ├───────┘     │ └──────► VUnit  │
 └────────┘             │        └────────┘
                        │
                        │        ┌────────┐
                        └────────►  ...   │
                                 └────────┘
```

# Runner

As commented, I find a lot of similarities between the implementation proposed in this PR and VUnit. That is probably because VUnit uses pytest intensively for dealing with multiple projects. Per previous discussions, the main stoppers for cocotb having a nice integration with VUnit are the following:

- Having environment variables defined per test. This was investigated by @jwprice100 (see VUnit/vunit#708). Although there are some technical details to polish, it seems feasible. Moreover, I found myself in the need of defining envvars per test in VUnit, for other use cases. Unfortunately, I could not prioritise that during this year.
- VUnit's runner's requirement of having a top-level generic of type string named `runner`. Cocotb uses the HDL entity of the UUT directly, because the whole testbench is implemented in Python and top-level I/O is toggled through VPI/VHPI.
  - As I prototyped in [OSVB: SFF](https://github.com/umarcor/osvb/tree/main/SFF), it is possible to autogenerate an HDL testbench which instantiates the UUT and which has a top-level runner generic. Moreover, pyVHDLModel would help in achieving that. However, that is unidiomatic from cocotb's point of view. Code generation is typically a symptom of lack of capability.
  - Recently, I created several issues in VUnit to discuss the generalisation of some features. On the one hand, compiling the HDL builtins by default is deprecated and it'll be removed in a future release (see VUnit/vunit#777 and VUnit/vunit#763). On the other hand, VUnit/vunit#772 is a discussion about documenting the syntax of the Python (runner) <-> HDL (runner) interface and make a public API from it:



The API in this case would be:

- How does Python tell the VUnit HDL runner, the OSVVM HDL runner or cocotb's Python runner, what tests to execute and which parameters to use.
- How do the runners tell Python the results. Note that "Python" in this case can be "VUnit", "Edalize", or, in the context of this PR, "pytest". Since cocotb is bootstraped by the simulator through the shared library, there is no practical difference between the cocotb runner reporting results or an HDL testbench (VUnit or OSVVM).

Note that these are 2 of the 4 particularities of cocotb that I mentioned above. By having them properly specified in a Python API, any project (VUnit, Edalize, etc.) could better integrate with cocotb without custom solutions. This is a very naive snippet:

```py
...

from cocotb.Project import cocotbProject

# Assume that HDL sources and simulator stuff are already configured in 'prj'

...

cocotbProject.fromTool(prj)
# or
cocotbProject = Tool2cocotb(prj)
# or
cocotbProject = prj.toCocotb()

...

# Interact with the cocotbProject to attach testbenches and data to the UUTs

...

# Get the runner configuration for some testbench
Tool.get_test('my_uut').set_runner_cfg(
    cocotbProject.get_runner_cfg(
        'testbench',
        ['test', 'cases'],
        any,
        other,
        param
    )
)

Tool.run()
# or
Tool.build()
Tool.test()
...

```

Naturally, you can write your own Project and your own Tool which uses that one implicitly. Nonetheless, I would strongly recommend that you keep the management of the "runner configuration" explicitly isolated (orthogonal) to the management of HDL sources and the simulators. This is because the project and the tool are subject to be replaced, while the runner configuration is specific to cocotb.

Note that the concept of "runner" is not mentioned in EDAA, but there is a section in OSVB. That is because the runner belongs to the Workflow layer in EDAA; and OSVB is just one possible workflow (based on VUnit). Hence, I acknowledge that this last section might be opinionated towards VUnit. Please, do not be mislead by that. Analyse each of the section as an standalone proposal, because they are independent despite the EDAA umbrella.

---

Overall, my perception is that the value of cocotb is on the "writing testbenches in Python" part. That is, the shared library, the common co-simulation interface, all the tweaks to deal with the VPI/VHPI implementations, the type/data conversions between languages, etc. Conversely, multiple test management, report generation and building/running were never the main focus of the project. They were afterthoughts. That is, by the way, not an exception (see OSVVM's TCL plumbing, or VUnit's cosim features). Almost a decade after those not-first-thought features were added out of need, I believe we should now take an step back and think what we want each to focus on. The complexity is far beyond what any project can achieve alone. Cocotb is undoubtely the best positioned project at the moment, but that doesn't make it immune to increased maintenace burden, which results in technical debt or rotting.

@themperek, from a volume point of view, this PR is adding ~800 lines of code. Conversely, ProjectModel is currently ~9k lines of code. Cocotb probably does not need half of the features in ProjectModel. That is a legit criticism. Nonetheless, 1k is not realistic. In the mid term, this "Python based runner" is likely to grow to several thousand lines. For instance, the GHDL class in this PR is 50 lines, while the equivalent class in VUnit is 400 lines. Therefore, I would expect a documented and mature version of this PR to get to 5-6k lines in a few years. This is not to discourage you. If you are commited to being the developer and maintainer of another solution, go ahead. In the end, that's one of beauties of open source! However, if you don't want to develop and maintain those 5-10k lines alone (or solely with cocotb contributors), I suggest to take some time to think about EDAA, or VUnit, or Edalize, or...

*As a matter of fact, several years ago (6-7?) @Paebbels discussed with VUnit maintainers about using a common project class for VUnit and pyIPCMI (PoC back then). VUnit maintainers did not want to have features in the class which were not useful for simulation. pyIPCMI's Project (the foundation of pyEDAA.ProjectModel) can handle sources for both simulation and synthesis. That is the reason for VUnit to use a different Project class. However, we are willing to evaluate replacing it now.*

ktbarrett:
> So, I would like to ask the cocotb community to evaluate whether having a pyEDAA.ProjectModel.cocotb class would be suitable.

What does cocotb have to do with projects? Projects can contain multiple designs and multiple tests of multiple types... IMO designs and tests exist at the same level. Perhaps a `CocotbTestbench` class that extends `Design` would be better.

> Overall, my perception is that the value of cocotb is on the "writing testbenches in Python" part. ... Conversely, multiple test management, report generation and building/running were never the main focus of the project. 

Correct. They are unwanted, but unavoidable, scope. We have to have *something* to give to new users and run our regressions. I'm not against adding N lines of code to support someone else handling building and running simulations if it gives users more flexibility than we can duplicate ourselves, without forcing users to use cocotb in a particular way.

> Per previous discussions, the main stoppers for cocotb having a nice integration with VUnit are the following

We have no interest in integrating with VUnit and appeasing it's assumptions at this time. That will take a non trivial amount of work to refactor our regression manager, test reporter, and scheduler. We can't just change them or throw them away at this point no matter how much I'd like to.

>The purpose is not to provide any homogeneous solution for all EDA tools, but to have a Python interface for each of them. ... How to then make different vendor tools look similar belongs to the Workflows (cocotb, Edalize, VUnit, pyIPCMI, etc.).
> Conversely, ProjectModel is currently ~9k lines of code. Cocotb probably does not need half of the features in ProjectModel. That is a legit criticism.

So we can use pyEDAA tool classes, but we are still going to have to write a "workflow" on top that understands EDAA Project Model because the tool classes don't understand how to build Designs? And EDAA is a larger interface (and not one that brings additional value IMO) than the one we are proposing.

So what exactly is the benefit here? I see barely any code reuse possible, unless we make significant changes to depend upon some other framework. At best I see things "looking the same" to the user, but they are still tied to our duplicate workflow, so that's kind of moot. Why can't EDAA be built to *not* depend on workflows for the incredibly simply tasks we are asking of it: build some HDL with this tool and pass these arguments.

themperek:
> @themperek, from a volume point of view, this PR is adding ~800 lines of code. Conversely, ProjectModel is currently ~9k lines of code. Cocotb probably does not need half of the features in ProjectModel. That is a legit criticism. Nonetheless, 1k is not realistic. In the mid term, this "Python based runner" is likely to grow to several thousand lines. For instance, the GHDL class in this PR is 50 lines, while the equivalent class in VUnit is 400 lines. Therefore, I would expect a documented and mature version of this PR to get to 5-6k lines in a few years. This is not to discourage you. If you are commited to being the developer and maintainer of another solution, go ahead. In the end, that's one of beauties of open source! However, if you don't want to develop and maintain those 5-10k lines alone (or solely with cocotb contributors), I suggest to take some time to think about EDAA, or VUnit, or Edalize, or...

As far as I understand from our internal discussions this PR serves:
- easy start without makefiles (especially for Windows users)
- so we can run cocotb tests/regresions

We are not trying to build any sophisticated solution. We want to replicate makefile functionality mainly and stop there.  Anything else has to come with extensions/other projects that I think all of us will welcome.
We still would like to make it nice/usable for ourselves so there is a bit of discussion.

umarcor:
@ktbarrett

> What does cocotb have to do with projects? Projects can contain multiple designs and multiple tests of multiple types... IMO designs and tests exist at the same level.

A Project is a set of files organised in filesets. That is used for "any purpose", be it simulation or synthesis. In fact, some sources (the UUT) are typically used for both cases, because the purpose of simulation is to verify what is to be synthesised. When the top-level is for synthesis, we say it's a Design. When it is for simulation, we say it's a Testbench. Nonetheless, both of them are primary units and their tree of dependencies (files and logical).

Hence, name it as you wish: Project, Design, Testbench... The point is that you have classes such as https://github.com/cocotb/cocotb/wiki/Python-Test-Runner-Proposal#models and those do exist in other solutions as well, such as EDAA o VUnit.

On the hight level, I want to have a project with multiple designs, each design having multiple target boards (sub-designs), multiple traditional testbenches (sub-designs), multiple OSVVM testbenches (sub-designs) and multiple cocotb testbenches (sub-designs). We do have a terminology issue that we need to discuss, because testbenches can be considered sub-designs as you said. Moreover, at the moment each target board is expected to be a Design, and not a sub-design. So, maybe a Project needs to be a design, and the Designs be sub-designs. I recall a discussion with @Paebbels about having a third level above Project or below Design, but I'm unsure about the conclusion.

> Perhaps a CocotbTestbench class that extends Design would be better.

Agree. With `pyEDAA.ProjectModel.cocotb` I meant "overloading or adding some features to the 'default' ProjectModel". Adding a CocotbTestbench class while reusing the Project abstraction as-is is a suitable solution.

As a matter of fact, Patrick did support cocotb in PoC/pyIPCMI until cocotb went dead in 2016-2018: #513. This is to say we might have explicit first-class support for cocotb testbenches, on par with "traditional" HDL testbenches, as long as the cocotb community wants to engage.

> Correct. They are unwanted, but unavoidable, scope. We have to have something to give to new users and run our regressions. I'm not against adding N lines of code to support someone else handling building and running simulations if it gives users more flexibility than we can duplicate ourselves, without forcing users to use cocotb in a particular way.

The usage of Project and Tools is transparent to the users. That is something that you (developers) can use, instead of duplicating those internal classes yourself. If users are aware of that, then we are doing something incorrectly.

I do agree that running your regression manager is something you cannot avoid at the moment, and I would not want to alienate users in that regard. However "building/compiling sources and running a simulator" are out that scope. You depend on those capabilities, but the abstraction layer is different. You are explicitly reworking that infrastructure, in order to replace/complement makefiles with a different CLI solution.

As commented, please do not be mislead by the Runner/VUnit stuff. That is independent from using the Project and Tools from EDAA.

> We have no interest in integrating with VUnit and appeasing it's assumptions at this time. That will take a non trivial amount of work to refactor our regression manager, test reporter, and scheduler. We can't just change them or throw them away at this point no matter how much I'd like to.

Precisely, I am proposing that VUnit adapts itself to the assumptions of other frameworks, such as cocotb or OSVVM. I'm willing to let cocotb users keep your regression manager, test reporter and scheduler. The issue is that cocotb's features are isolated, because they only exist in a bootstraped instance. How do you communicate those with any external "service" or "wrapper"? So, how can e.g. Edalize tell cocotb which tests to run or where to put the result files?
Forget about VUnit. Whatever works for any other wrapper is potentially reusable in VUnit.

> So we can use pyEDAA tool classes, but we are still going to have to write a "workflow" on top that understands EDAA Project Model because the tool classes don't understand how to build Designs?

That is correct. The ProjectModel and the Tool classes in EDAA are independent, by design. If that was not the case, we would be providing yet another monolithic solution, incompatible with any of the existing solutions.

There is going to be a ready-to-use workflow based on EDAA: [pyIPCMI](https://github.com/Paebbels/pyIPCMI). If you want to use almost all of the EDAA layers in a bundle which provides commands such as "simulate this project with this tool" or "synthesise this design with this other tool", then you'd better wait for pyIPCMI to be restructured. That will take a year, at least, though. The ProjectModel is ready to use, but we don't expect hard work on the CLI and Tool classes until a couple of week during xmas, and then the summer of 2022. Naturally, that can change if other people are willing to engage and accelerate those layers. I will likely write the GHDL and Yosys classes in the following weeks (because they provide the synthesis features I miss in VUnit), but those are just two tools.

> And EDAA is a larger interface (and not one that brings additional value IMO) than the one we are proposing.

The additional value of pyEDAA.ProjectModel (not the whole EDAA) is interoperability. Users with a set of HDL sources need to define multiple projects:

- `run.py` script for VUnit.
- makefile or a `runner.py` script for cocotb.
- `.core` file for FuseSoC/Edalize.
- `hdl-prj.json` file for GHDL's language server.
- `.xpr` or `.tcl` for Vivado.
- `main.py` script for PyFPGA.
- `vhdl_ls.toml` for rust_hdl.
- `.pro` files for OSVVM.
- `.txt` files for UVVM.
- ...

The huge majority of the content in those files is exactly the same. Then, each has some minor configuration stuff for the particular tool.

I cannot speak for all the projects, but I can state that VUnit, OSVVM, Vivado and GHDL, at least, will be interoperable through pyEDAA.ProjectModel. Reading Vivado's `.xpr` and OSVVM's `.pro` files is supported already. Using pyEDAA.ProjectModel to load OSVVMLibraries into VUnit is work in progress. Whether VUnit will replace the internal Project class with EDAA's is yet to be evaluated. They feel so similar because I fed back @Paebbels with my expectations as a VUnit user and maintainer, but we did not look into the details yet.

> So what exactly is the benefit here? I see barely any code reuse possible, unless we make significant changes to depend upon some other framework. At best I see things "looking the same" to the user, but they are still tied to our duplicate workflow, so that's kind of moot.

We can analyse it from the other side: how can the code in this PR be reusable in EDAA?
The `build` and `test` look nice, in the sense that most parameters are explicit, even though that implies duplication; instead of using a class/object to carry them. However, the simulator classes don't have those explicit parameters. There, it is inherited from the common class.
So, we might refactor the simulator classes (`Icarus`, `Questa`, `Ghdl`, `Riviera` and `Verilator`) for them to be usable standalone (regardless of the `Simulator` class), then rename `Simulator` to `SimulationCocotb` and have mixins of those (`IcarusCocotb`, `QuestaCocotb`, `GhdlCocotb`, `RivieraCocotb` and `VerilatorCocotb`). Then, all the non `*Cocotb` classes would be a suitable candidate for the prototype of pyEDAA.CLITool, and the `SimulationCocotb` would contain the specific knowledge about "the simple cocotb workflow".

The reason for EDAA to be based on pieces of pyIPCMI is that @Paebbels was the first and only person in the community who said "I am willing to break my monolith into pieces and invest several months of effort if that makes at least parts of it reusable by others". However, we don't need to wait until he refactors all the layers of pyIPCMI. Moreover, in the context of EDAA it is ok to have multiple implementations of the same layers during the initial phases. The end-goal is to homogenise the ecosystem, but the reality now is fragmented so we must acknowledge that.

> Why can't EDAA be built to not depend on workflows for the incredibly simply tasks we are asking of it: build some HDL with this tool and pass these arguments (all of this customization would be built into the CocotbTestbench class).

The main reason is I don't personally feel like competing with anyone for a solution that exists already. On average, I'm 10y younger than the main developers of most of the mentioned tools. I arrived late at that party.
The motivation for creating EDAA is not to repeat the same mistake of claiming that our monolithic approach is the best for arbitrary reasons. I did prototype my own solution 3 years ago (bash -> python -> [dbhi/run](https://github.com/dbhi/run)) and I understood it was far beyond the capacity of a single person, no matter how clever I am.
We need to get people from different communities/camps to collaborate. If we cannot fix that, at least we tried. The worst case scenario is an updated pyIPCMI with nicer documentation as a result of 3-5 people collaborating, which is not a bad outcome *per se*. Some pieces, such as pyEDAA.UCIS or pyEDAA.Reports will be usable standalone regardless of the fate of EDAA as a whole.

Nevertheless, it is perfectly suitable to write a CocotbTestbench class which uses pyEDAA.ProjectModel and the Simulator classes written by @themperek. So, ignore the Tool layers of EDAA for now. That CocotbTestbench class would be considered a Workflow layer in the EDAA concept model, which is just a semantic classification. In this case, we would not be targeting much code reuse, but interoperability. In fact, that would require no code removal in this PR, but just adding a couple of methods to the existing Simulator class, to take a Project object and call `build` and/or `test`.

Moreover, I understand your criticism about EDAA not providing a reference implementation of how can layers be used together. That will exist, because it is a must for testing, for documentation and to communicate what EDAA is about. However, I would consider it an example or a demo, not something to actively maintain and have end-users consume. This is again strategic: we want all the existing projects to preserve their identity and user bases. We want to facilitate people to learn about other verification approaches, by reducing the burden to jump from one entrypoint to the other; we want to allow development teams to handle sub-teams using different frameworks. By ensuring that users cannot consume EDAA directly, we "force" them to go and read the philosophical motivations of cocotb, VUnit, OSVVM, UVVM, SVUnit, etc. about how needs testing to be done from a theoretical/conceptual/ideological point of view. A very pretentious analogy on my side would be: why does Linus not provide an incredibly simple distro on top of Linux?

We might attempt to funnel all the community into an EDAA workflow (be it named pyIPCMI or something else), but building workflows is not our area of expertise. Edalize has been trying to achieve that for several years now with bittersweet results. The knowledge we are trying to gather in EDAA is primarily: which kind of files/data needs to be handled in a typical HDL/EDA project, and which tools need to be used to manipulate/generate/consume those files. To me, EDAA is mostly about "documentation as code". Not as md/rst code, but as a reference implementation with auto-generated documentation, cross-references, diagrams, etc. which is the main expertise of Patrick aside of VHDL. Other than that, we do have knowledge about the workflows we use (in my case, see [dbhi.github.io/concept/run](https://dbhi.github.io/concept/run)), but we cannot afford to develop and maintain a generalised solution to that layer.

If you want incredibly simple tasks, without waiting to EDAA Tool classes and without writing/maintaining them yourselves, I suggest you have a look at FuseSoC/Edalize, PyFPGA or Xeda. Those are as straightforward as you can get. As commented above, if the integration with one (any) of those tools is done properly, the knowledge is reusable in any other. You don't need to "buy" any of them, but specify the interaction with any external wrapper.

As we discussed in the chat some months ago, in the mid term I would expect several of those workflow focused solutions to provide atomic tasks that can be arranged in a DAG. @olofk did already start working in that direction, using EDAM as the data format from task to task. I did not look into the details yet to understand whether Edalize is the only possible orchestrator or some alternative can be plugged in.

> As far as I understand from our internal discussions this PR serves:
>
> * easy start without makefiles (especially for Windows users)
>
> We are not trying to build any sophisticated solution. We want to replicate makefile functionality mainly and stop there.

@themperek, I believe that was the initial motivation for all the projects mentioned above. It is interesting to analyse how so many initially simple and straightforward solutions evolved into the monoliths we have nowadays.

If you can keep the complexity of this solution around 1-2k lines for several years, then absolutely go ahead. No other solution will be as effective. However, as soon as it starts drifting away (coverage, checking versions/features, docstrings, etc.), please consider revisiting this dialogue.

Paebbels:
@ktbarrett , @themperek and @umarcor let's organize a Zoom meeting and discuss the needs of Cocotb.

In general, I think pyEDAA.ProjectModel needs another branch of classes for testbench/testharness/testcase besides  
`project -> design -> fileset -> file`.

ktbarrett (commented):
In general there are a few style things.
* methods that don't need self in the base class should probably be free functions
* methods that are meant to be overridden that don't use self should be staticmethods (so you don't have to instance the type to use the function independently)
* some odd style with branches (create empty list, append in branches, then return outside of branches?)

The biggest design issue I see is how clean() is implemented. It deletes "sim_build" directories, potentially recursively. But you can't change it from sim_build to something else?

The other design issue is Questa and Riviera invoking recursively. I don't see a need to ever do that.

cocotb/runner.py
```diff
@@ -0,0 +1,721 @@
...
+        include_cmd = []
+        for dir in includes:
+            include_cmd.append("-I" + dir)
+
+        return include_cmd
+
+    def get_define_commands(self, defines):
+        defines_cmd = []
+        for define in defines:
+            defines_cmd.append("-D" + define)
```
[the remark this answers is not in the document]
jonpovey: This would only return the first define? What am I missing?
Could instead do list comprehension:
`return ["-D" + define for define in defines]`

cocotb/runner.py
```diff
@@ -0,0 +1,721 @@
...
+        self.toplevel_lang = toplevel_lang
+        self.sim_args = extra_args
+        self.plus_args = list(plus_args)
+        self.env = dict(extra_env) if extra_env is not None else {}
+
+        if testcase is not None:
+            self.env["TESTCASE"] = testcase
+
+        if seed is not None:
+            self.env["RANDOM_SEED"] = str(seed)
```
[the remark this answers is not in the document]
themperek: Some are very handy like `WAVES` and `GUI` so one can debug without changing the code.

`COCOTB_RESULTS_FILE` is special. In `cocotb-test` I used it in `pytest` plugin to combine the results see: https://github.com/themperek/cocotb-test/blob/master/cocotb_test/plugin.py . Maybe there is a better way?

The rest can probably go.

cocotb/runner.py
```diff
@@ -0,0 +1,721 @@
...
+    if sim_name == "icarus":
+        runner = Icarus
+    elif sim_name == "questa":
+        runner = Questa
+    elif sim_name == "ghdl":
+        runner = Ghdl
+    elif sim_name == "riviera":
+        runner = Riviera
+    elif sim_name == "verilator":
+        runner = Verilator
```
eric-wieser, line 703:
```suggestion
    supported_sims = {
      "icarus": Icarus,
      "questa": Questa,
      "ghdl": Ghdl,
      "riviera": Riviera,
      "verilator": Verilator,
    }  # TODO: "ius", "xcelium", "vcs"
    try:
        return supported_sims[sim_name]
    except KeyError:
        raise NotImplementedError("Set SIM variable. Supported: " + ", ".join(supported_sim)) from None
```

cocotb/runner.py
```diff
@@ -0,0 +1,721 @@
...
+            if self.vhdl_sources:
+                self.env["GPI_EXTRA"] = cocotb.config.lib_name_path("fli", "questa") + ":cocotbfli_entry_point"
+
+        if self.waves:
+            do_script += "log -recursive /*;"
+
+        if not self.gui:
+            do_script += "run -all; quit"
+
+        cmd.append(["vsim"] + (["-gui"] if self.gui else ["-c"]) + ["-do"] + [do_script])
```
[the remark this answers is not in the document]
themperek: Like now is ok?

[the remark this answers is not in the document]
themperek: As some reported It is bit faster to do it in a script (start vsim once). This way is nicer.

cocotb/runner.py
```diff
@@ -0,0 +1,721 @@
...
+                    EXTRA_ARGS=" ".join(as_tcl_value(v) for v in self.compile_args),
+                )
+        else:
+            print("WARNING: Skipping compilation:" + out_file)
+
+        do_file = tempfile.NamedTemporaryFile(delete=False)
+        do_file.write(do_script.encode())
+        do_file.close()
+
+        return [["vsimsa"] + ["-do"] + ["do"] + [do_file.name]]
```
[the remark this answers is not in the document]
themperek: I did not manage to get this running even without generating a physical script file at a time. Maybe someone will figure this out.

cocotb/runner.py
```diff
@@ -0,0 +1,721 @@
...
+        except OSError:
+            pass
+
+        cmds = self.test_command()
+        self.set_env()
+        self.execute(cmds)
+
+        self.check_results_file(results_xml_file)
+
+        print("INFO: Results file: %s" % results_xml_file)
```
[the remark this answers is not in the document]
themperek: If one adds the logger in the case of pytest the login stream will be captured different than `stdout` 
- https://pytest.org/en/6.2.x/capture.html  
- https://docs.pytest.org/en/6.2.x/logging.html

In principle not an issue but ...

cocotb/runner.py
```diff
@@ -0,0 +1,721 @@
...
+        return include_cmd
+
+    def get_define_commands(self, defines):
+        defines_cmd = []
+        for define in defines:
+            defines_cmd.append("+define+" + as_tcl_value(define))
+
+        return defines_cmd
+
+    def get_parameter_commands(self, parameters):
```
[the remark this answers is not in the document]
themperek: For Icarus it will not work we need `self.toplevel` name 🤷🏻‍♂️

cocotb/runner.py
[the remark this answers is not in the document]
themperek: So we would need to forward the internal simulator/cocotb logger to the top-level logger.

cocotb/runner.py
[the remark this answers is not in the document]
ktbarrett: How about we change WAVES and GUI to COCOTB_WAVES and COCOTB_GUI? Or is there a way to pass those using pytest commandline switches like `pytest --cocotb-waves --cocotb-gui` like what pytest-coverage does?

cocotb/runner.py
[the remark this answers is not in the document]
ktbarrett: I don't see the problem? The logs for the cocotb runner are separated from the rest. That's a good thing. Am I missing something?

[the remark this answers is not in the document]
themperek: Not a problem just but anoying and different ways to display.

cocotb/runner.py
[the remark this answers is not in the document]
themperek: > Or is there a way to pass those using pytest commandline switches like `pytest --cocotb-waves --cocotb-gui` like what pytest-coverage does?

This can be done but will need those variables in any case (as I know).

ktbarrett:
@themperek What is `self.process`?

themperek:
> The biggest design issue I see is how clean() is implemented. It deletes "sim_build" directories, potentially recursively. But you can't change it from "sim_build" to something else?

I have no good idea how to do this better. My idea would be to add commend `cocotb-clean` clean `sim_build` by default. If one uses something else 🤷🏻

themperek:
> @themperek What is `self.process`? It doesn't seem to be set anywhere. I don't think that the signal handler is necessary on second inspection. It seems like subprocess.run handles forwarding SIGINT.

Sorry. Corrected.

>And I'm guessing toplevel in Icarus is supposed to be hdl_topmodule?

Fixed

cocotb/runner.py
themperek: >I don't see how it will break compatibility? Other simulators would just ignore or throw a warning if the argument is given.

We can `make_args` to all simulators  `Simulator` class or only for `Verilator` class. It will break if we do `Verilator` class.

cocotb/runner.py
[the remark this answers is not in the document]
themperek: I like `WAVES` and `GUI` more. This is just for runner not for `cocotb` itself. But no strong opinion.

ktbarrett:
@themperek I made a few changes [here](https://github.com/ktbarrett/cocotb/commit/1f647da10368a9a74d9c006f58b75e5734f05083). There are some issues with the typing, most of them regarding `hdl_topmodule`, which is Optional. If it isn't optional, most of the errors go away.

themperek:
@ktbarrett 
>I made a few changes here. 

It looks OK to me. How should we proceed? Should copy/marge it here? Do you open new PR?

>There are some issues with the typing, most of them regarding hdl_topmodule, which is Optional. If it isn't optional, most of the errors go away.

Maybe it should not be optional? We have this https://github.com/cocotb/cocotb/blob/3de64b8257d2a3974c79cb45b9c57510df0b2533/cocotb/__init__.py#L254
In this case should be optional in `build()` and `test()` or only test()?

ktbarrett:
@themperek I would have to remove the `Optional` on `hdl_topmodule` in `build` and make it a required argument. During build the only use is for setting parameters on Icarus. And that is valid for the *build* toplevel. The sim toplevel could be any entity in the design. Do you think we should maybe separate the two? hdl_topmodule in build and sim_toplevel in test?

Looks like I had permission to push, so I did so. Wanted to make sure you were okay with it first.

ktbarrett:
I'm not planning on removing `hdl_module`. It does seem like it's optional during build for some simulators. And during sim toplevel can be optional as well

* Most simulator don't need the toplevel to build the design (exception is Verilator and parameters for Icarus which must be fully qualified)
* If toplevel is given during build it doesn't need to be passed during test, it can be copied.
* If toplevel is given during both build and test, they can be different on some simulators. Questa and Riviera I know support this, and users have complained about the makefiles not supporting this in the past. Other simulators can just bomb in this case.

Separating them into two variables removes the typing problem and also makes it more flexible. We can even call them the same thing `toplevel` in both calls, they are just stored in different variables internally and used in different places.

themperek:
@ktbarrett 
> Separating them into two variables removes the typing problem and also makes it more flexible. We can even call them the same thing `toplevel` in both calls, they are just stored in different variables internally and used in different places.
> 


We will not find the ideal solution here.  I think what you propose is OK for a compromise. So `toplevel` in `test()` is required and `hdl_topmodule` in `build()` only for Verilator and Icarus in cases necessary. This means for `Verilator` will need it always 2 times but this is how it is ...

> I'll put together a commit and see what you think. See [here](https://github.com/ktbarrett/cocotb/commit/eac406875f975b1b10f6dffb888e48bfa9037b3d).

Is OK. Added one comment.

ktbarrett:
@themperek Addressed your comments with https://github.com/ktbarrett/cocotb/commit/9e3ccb9a9489b3be223608b00b920196f7560f9e.

ktbarrett:
Addressed most recent comment. https://github.com/ktbarrett/cocotb/commit/76040ecf863fe3e7e597e38377660f90f35cc280

themperek:
@ktbarrett 
> Addressed most recent comment. [ktbarrett@76040ec](https://github.com/ktbarrett/cocotb/commit/76040ecf863fe3e7e597e38377660f90f35cc280)

Looks OK

cocotb/runner.py
```diff
@@ -0,0 +1,800 @@
...
+        toplevel_lang: Optional[str] = None,
+        testcase: Optional[str] = None,
+        seed: Optional[Union[str, int]] = None,
+        python_search: Sequence[PathLike] = [],
+        extra_args: Sequence[str] = [],
+        plus_args: Sequence[str] = [],
+        extra_env: Mapping[str, str] = {},
+        waves: Optional[bool] = None,
+        gui: Optional[bool] = None,
+        build_dir: PathLike = "sim_build",
```
themperek, line 152: @ktbarrett 
Can we make `build_dir` optional here and in this case use `build_dir` from `build`?

tests/pytest/test_runner.py
```diff
@@ -0,0 +1,56 @@
...
+        vhdl_sources=vhdl_sources,
+        toplevel="runner",
+        parameters=parameters,
+        defines=["DEFINE=4"],
+        includes=[os.path.join(tests_dir, "designs", "basic_hierarchy_module")],
+        build_dir=sim_build + "/test_runner/" + "_".join(("{}={}".format(*i) for i in parameters.items())),
+    )
+
+    runner.test(
+        toplevel_lang="verilog",
```
themperek, line 51:
```suggestion
```

cocotb/runner.py
```diff
@@ -0,0 +1,800 @@
...
+
+        self.env["PATH"] += os.pathsep + cocotb.config.libs_dir
+
+        self.env["PYTHONPATH"] = os.pathsep.join(sys.path)
+        for path in self.python_search:
+            self.env["PYTHONPATH"] += os.pathsep + str(path)
+
+        prefix = sysconfig.get_config_var("prefix")
+        if prefix is not None:
+            self.env["PYTHONHOME"] = prefix
```
themperek, line 86:
```suggestion
        self.env["PYTHONHOME"] = sys.prefix
```

cocotb/runner.py
```diff
@@ -0,0 +1,798 @@
...
+            )
+
+            # TODO: create at thread to handle stderr and log as error?
+            # TODO: log forwarding
+
+            process = subprocess.run(cmd, cwd=self.work_dir, env=self.env)
+
+            if process.returncode != 0:
+                raise SystemExit(
+                    "Process '%s' termindated with error %d"
```
marlonjames, line 238:
```suggestion
                    "Process '%s' terminated with error %d"
```

ktbarrett:
Well that didn't quite go as planned. So what exactly is build_dir and what exactly is work_dir? I would presume the point of separate build step from test step is to build once and then test many, potentially even in parallel. So there needs to be a build dir and a sim_dir. Builds are done in the build dir and sims are done in the sim dir. I'll revert and push up a proposal.

themperek:
> Well that didn't quite go as planned. So what exactly is build_dir and what exactly is work_dir? I would presume the point of separate build step from test step is to build once and then test many, potentially even in parallel. So there needs to be a build dir and a sim_dir. Builds are done in the build dir and sims are done in the sim dir. I'll revert and push up a proposal ([ktbarrett@e199f2e](https://github.com/ktbarrett/cocotb/commit/e199f2e1f5abbb06f766e7fe824f89ef1ce0a2af)).

I think it is OK how you propose. The original intention was to be able to execute in a different place but maybe like now is better. 

After you add this I think will add one more test where we separate build and test and see how this works.  Something like.

```python3
@pytest.fixture(scope="module")
def runner():

    runner = get_runner(sim)()

    runner.build(
    )

    return runner

def test_runner(runner):

    runner.test(
    )
```

olofk:
Admittedly late to the party but have you considered using Edalize for interfacing the EDA tools instead? That would save you from maintaining code for that. Or if there are some reasons against that, perhaps at least use the EDAM API for your runner implementations. That would allow for easy interoperability with other EDAM-speaking tools like FuseSoC, Edalize and TerosHDL

ktbarrett:
@olofk We are avoiding depending on an outside tool in this repo. I'd rather have integration into multiple other toolchains done out of repo. There were a couple attempts at integrating cocotb support into edalize and FuseSoC, but they were deemed unsatisfactory and never completed. Using another tool's project models is an option. How does EDAM significantly differ from PyEDAA's Project Model? IIRC you and Unai were talking about converging the two?

themperek:
@olofk Few comments as far as I understand:
- We would like to have a light python runner especially for windows (issues with make etc.) 
- We would like to make use of unit testing frameworks like pytest (also for parallel runs)
- It should be rather easy to integrate cocotb in other flows: https://docs.cocotb.org/en/stable/custom_flows.html
- Can help and full support using cocotb with FuseSoC, Edalize, ...

olofk:
Fair enough. Totally understand the reluctance to depend on external projects. Hoping to make a new effort to work on cocotb integration from the Edalize side. IIUC the previous attempts have been a bit overcomplicated and it should pretty much just be to load cocotb as a VPI lib and set some env vars but I'll happily take up @themperek's offer for assistance since I'm still new to cocotb.

Re the EDAM API, it's just a data structure (could be python dicts/lists, yaml, json or whatever) that describes the design in terms of things like design files (verilog, vhdl etc), TCL scripts, parameters/generics, defines, VPI libs, toplevel(s) as well as tool-specific options. Kind of the same information that I have seen being put into cocotb makefiles. The idea is to have something mostly tool-agnostic that can then be turned into tool-specific scripts or project files (which is what Edalize does basically)

themperek:
@ktbarrett Maybe we can try to get this in as it is and then work with smaller PRs on this?

ktbarrett:
I think this can come in now.

tests/designs/runner/runner.v
```diff
@@ -0,0 +1,15 @@
+
```
cmarqu, line 1:
```suggestion
// Copyright cocotb contributors
// Licensed under the Revised BSD License, see LICENSE for details.
// SPDX-License-Identifier: BSD-3-Clause

```

tests/designs/runner/runner.vhdl
```diff
@@ -0,0 +1,18 @@
+
```
cmarqu, line 1:
```suggestion
-- Copyright cocotb contributors
-- Licensed under the Revised BSD License, see LICENSE for details.
-- SPDX-License-Identifier: BSD-3-Clause

```

tests/pytest/test_cocotb.py
```diff
@@ -0,0 +1,65 @@
+import os
```
cmarqu, line 1:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause

import os
```

tests/pytest/test_runner.py
```diff
@@ -0,0 +1,55 @@
+import pytest
```
cmarqu, line 1:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause

import pytest
```

cocotb/runner.py
```diff
@@ -0,0 +1,784 @@
...
+        self.set_env()
+        self.execute(cmds, cwd=self.sim_dir)
+
+        check_results_file(results_xml_file)
+
+        print(f"INFO: Results file: {results_xml_file}")
+        return results_xml_file
+
+    @abc.abstractmethod
+    def get_include_commands(self, includes: Sequence[str]) -> List[str]:
```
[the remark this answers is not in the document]
themperek: Yes. Probably `get_include_options` is better?. 

Will change.

cocotb/runner.py
```diff
@@ -0,0 +1,807 @@
...
+            process = subprocess.run(cmd, cwd=cwd, env=self.env)
+
+            if process.returncode != 0:
+                raise SystemExit(
+                    f"Process {process.args[0]!r} terminated with error {process.returncode}"
+                )
+
+
+def check_results_file(results_xml_file: PathLike) -> None:
+    """Checks if cocotb result file exisit and include faild tests."""
```
cmarqu, line 249:
```suggestion
    """Check whether result file exists and contains failed tests."""
```

cocotb/runner.py
```diff
@@ -0,0 +1,807 @@
...
+        for tc in ts.iter("testcase"):
+            for _ in tc.iter("failure"):
+                failed += 1
+
+    if failed:
+        raise SystemExit(f"ERROR: Failed {failed} tests.")
+
+
+def outdated(output: PathLike, dependencies: Sequence[PathLike]) -> bool:
+    """Checks if sources files have changed in reference to output."""
```
cmarqu, line 272:
```suggestion
    """Check if sources files are newer than output."""
```
?

cmarqu (approved):
Found these two, but fine otherwise for now.

cocotb/runner.py
```diff
@@ -0,0 +1,807 @@
...
+            self.gui = bool(gui)
+
+        results_xml_file = os.getenv(
+            "COCOTB_RESULTS_FILE", os.path.join(self.build_dir, "results.xml")
+        )
+
+        try:
+            os.remove(results_xml_file)
+        except OSError:
+            pass
```
cmarqu, line 195:
```suggestion
        with suppress(OSError):
            os.remove(results_xml_file)
```
(needs `from contextlib import suppress`)

cocotb/runner.py
```diff
@@ -0,0 +1,807 @@
...
+        return ["-I" + dir for dir in includes]
+
+    @staticmethod
+    def get_define_options(defines: Sequence[str]) -> List[str]:
+        return ["-D" + define for define in defines]
+
+    def get_parameter_options(self, parameters: Mapping[str, object]) -> List[str]:
+        assert self.hdl_toplevel is not None
+        return [
+            "-P" + self.hdl_toplevel + "." + name + "=" + str(value)
```
cmarqu, line 330:
```suggestion
            f"-P{self.hdl_toplevel}.{name}={value}"
```

cocotb/runner.py
```diff
@@ -0,0 +1,807 @@
...
+                )
+
+        return cmd
+
+
+class Ghdl(Simulator):
+    @staticmethod
+    def simulator_in_path() -> None:
+        if shutil.which("ghdl") is None:
+            raise SystemExit("ERROR: ghdl exacutable not found!")
```
cmarqu, line 501:
```suggestion
            raise SystemExit("ERROR: ghdl executable not found!")
```

cmarqu (commented):
Sorry, some more small things :)

PR #2640, opened 2021-07-21
cocotb.types.Array improvements

ktbarrett:
* Made Array a generic homogenous type, like list
* Split out an ABC (for use later)
* ~~replaced concat() with unused binary operator '@' (better typing, no import necessary)~~
* improved typing support
* fixed a couple documentation issues

#2514 Will use the ABC split, so this PR is necessary. Other changes could be contested.

### DIFF
--- a/cocotb/types/array.py
+++ b/cocotb/types/array.py
@@ -1,15 +1,19 @@
 # Copyright cocotb contributors
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
-from typing import Optional, Any, Iterable, Iterator, overload
-from collections.abc import Sequence
-from .range import Range
-from sys import maxsize
+import typing
+from itertools import chain
+
+from cocotb.types.range import Range
+
+T = typing.TypeVar("T")
+S = typing.TypeVar("S")
+Self = typing.TypeVar("Self", bound="Array[typing.Any]")
 
 
-class Array(Sequence):
+class Array(typing.Reversible[T], typing.Collection[T]):
     r"""
-    Fixed-size, arbitrarily-indexed, heterogeneous sequence type.
+    Fixed-size, arbitrarily-indexed, homogeneous collection type.
 
     Arrays are similar to, but different from Python :class:`list`\ s.
     An array can store values of any type or values of multiple types at a time, just like a :class:`list`.
@@ -25,9 +29,6 @@ class Array(Sequence):
         >>> Array("1234")  # the 0-based range `(0, len(value)-1)` is inferred
         Array(['1', '2', '3', '4'], Range(0, 'to', 3))
 
-        >>> Array(range=Range(0, "downto", -3))  # the initial values are `None`
-        Array([None, None, None, None], Range(0, 'downto', -3))
-
         >>> Array([1, True, None, "example"], Range(-2, 1))  # initial value and range lengths must be equal
         Array([1, True, None, 'example'], Range(-2, 'to', 1))
 
@@ -35,7 +36,7 @@ class Array(Sequence):
 
     .. code-block:: python3
 
-        >>> Array(range=Range(1, "to", 0))
+        >>> Array([], range=Range(1, "to", 0))
         Array([], Range(1, 'to', 0))
 
     Indexing and slicing is very similar to :class:`list`\ s, but it uses the indexing scheme specified.
@@ -128,30 +129,11 @@ class Array(Sequence):
         "_range",
     )
 
-    _value: Sequence
-    """
-    Private interface for subclasses to access the value as a :class:`~collections.abc.Sequence`
-
-    Subclasses that don't use a Sequence as their main representation should emulate this object,
-    or override *all* :class:`~cocotb.types.Array` methods *except*:
-        - :attr:`left`
-        - :attr:`direction`
-        - :attr:`right`
-        - :attr:`range`
-        - :attr:`__len__`
-        - :attr:`_translate_index`
-    """
-
-    def __init__(
-        self, value: Optional[Iterable[Any]] = None, range: Optional[Range] = None
-    ):
-        if value is not None and range is None:
-            self._value = list(value)
+    def __init__(self, value: typing.Iterable[T], range: typing.Optional[Range] = None):
+        self._value = list(value)
+        if range is None:
             self._range = Range(0, "to", len(self._value) - 1)
-        elif value is not None and range is not None:
-            if not isinstance(range, Range):
-                raise TypeError("range argument must be of type 'Range'")
-            self._value = list(value)
+        else:
             self._range = range
             if len(self._value) != len(self._range):
                 raise ValueError(
@@ -159,13 +141,6 @@ class Array(Sequence):
                         len(self._value), self._range
                     )
                 )
-        elif value is None and range is not None:
-            if not isinstance(range, Range):
-                raise TypeError("range argument must be of type 'Range'")
-            self._value = [None] * len(range)
-            self._range = range
-        else:
-            raise TypeError("must pass a value, range, or both")
 
     @property
     def left(self) -> int:
@@ -201,29 +176,31 @@ class Array(Sequence):
     def __len__(self) -> int:
         return len(self.range)
 
-    def __iter__(self) -> Iterator[Any]:
+    def __iter__(self) -> typing.Iterator[T]:
         return iter(self._value)
 
-    def __reversed__(self) -> Iterator[Any]:
+    def __reversed__(self) -> typing.Iterator[T]:
         return reversed(self._value)
 
-    def __contains__(self, item: Any) -> bool:
+    def __contains__(self, item: object) -> bool:
         return item in self._value
 
-    def __eq__(self, other: Any) -> bool:
-        if not isinstance(other, type(self)):
-            return NotImplemented
-        return self._value == other._value
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, type(self)):
+            return self._value == other._value
+        return NotImplemented
 
-    @overload
-    def __getitem__(self, item: int) -> Any:
-        pass  # pragma: no cover
+    @typing.overload
+    def __getitem__(self, item: int) -> T:
+        ...
 
-    @overload
-    def __getitem__(self, item: slice) -> "Array":
-        pass  # pragma: no cover
+    @typing.overload
+    def __getitem__(self, item: slice) -> "Array[T]":
+        ...
 
-    def __getitem__(self, item):
+    def __getitem__(
+        self, item: typing.Union[int, slice]
+    ) -> typing.Union[T, "Array[T]"]:
         if isinstance(item, int):
             idx = self._translate_index(item)
             return self._value[idx]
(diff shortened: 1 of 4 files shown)


ktbarrett:
Coverage will go to 100% once #2624 is in.

cocotb/types/array.py
```diff
@@ -285,42 +329,20 @@ def __setitem__(self, item, value):
...
-            stop = self._translate_index(stop)
-        else:
-            stop = maxsize  # same default value used by Python lists
-        idx = self._value.index(value, start, stop)
-        return self._range[idx]
-
-    def count(self, value: Any) -> int:
-        """Return number of occurrences of *value*."""
-        return self._value.count(value)
+    def __matmul__(self, other: ArrayABC[S]) -> "Array[typing.Union[T, S]]":
```
cmarqu, line 342: There is no doc about concatenating Arrays anymore now.

tests/pytest/test_array.py
```diff
@@ -164,30 +160,13 @@ def test_slice_correct_infered():
 def test_array_concat():
     l = Array("01ZX", Range(0, 'to', 3))
```
cmarqu, line 161:
```suggestion
    """Test array concatenation (``@``)."""
    l = Array("01ZX", Range(0, 'to', 3))
```

ktbarrett: Should I add docstrings to all the tests, or just this one?

cmarqu: Just this one, the other ones are self-explanatory IMO; I just thought that every reminder that `@` means "concatenation" is useful. (Reminders will become less important once `@` is showing up in examples, so feel free to ignore the docstring suggestion.)

eric-wieser:
Can you elaborate on the motivation for using `@` as array concatenation? I'm fairly against this idea, as it has semantically quite a different meaning in the rest of python (going against the PEP that introduces it), and has an unsuitable operator precedence.

If easy of typing is the concern, have you considered `from cocotb import c_`, `c_[a, b, c]` to concatenate `a`, `b`, and `c`? That looks more like the verilog `{a, b, c}`, and has the correct precedence for expressions like `c_[a + b, c]`, which ` a + b @ c` does not.

cocotb/types/array.py
```diff
@@ -1,15 +1,116 @@
...
+    def __setitem__(self, item: int, value: T) -> None:
+        ...
+
+    @typing.overload
+    def __setitem__(self, item: slice, value: typing.Iterable[T]) -> None:
+        ...
+
+    @abstractmethod
+    def __setitem__(self, item, value):  # type: ignore
+        ...
```
eric-wieser, line 47: I think since this is an actual method that can be called and not just a stub for typing, this should probably `raise NotImplementedError`

documentation/source/library_reference.rst
```diff
@@ -71,19 +71,18 @@ as the types used by cocotb's `simulator handles <#simulation-object-handles>`_.
...
 
-.. autoclass:: cocotb.types.Range
+.. autoclass:: cocotb.types.range.Range
     :members:
     :exclude-members: count, index
 
-.. autofunction:: cocotb.types.concat
-
-.. autoclass:: cocotb.types.Array
+.. autoclass:: cocotb.types.array.Array
```
eric-wieser, line 82: As I understand it, new versions of Sphinx will do the right thing with the old version given the contents of `types/__init__.py.`

cocotb/types/array.py
ktbarrett: I removed that change.

cocotb/types/array.py
```diff
@@ -1,15 +1,116 @@
...
 
+from cocotb.types.range import Range
 
-class Array(Sequence):
+T = typing.TypeVar("T")
+S = typing.TypeVar("S")
+
+
+class ArrayABC(typing.Reversible[T], typing.Collection[T], ABC):
+    """ """
```
eric-wieser, line 15: Can you add a summary of the type of thing this is intended to be used for?

cocotb/types/array.py
ktbarrett: Will probably add `raise NotImplementedError` to the list of exclude_lines in #2624.

documentation/source/library_reference.rst
ktbarrett: Doesn't seem to be happy locally, but I can always leave it the way it was and eventually it will start working?

eric-wieser: Not a big deal either way - if you can get it working how it was, then Sphinx will show the type name as `cocotb.types.Array` instead of `cocotb.types.array.Array`.

ktbarrett: I returned it to the way it was. The worst thing is that some type annotations aren't linking correctly, for types that are not even a page above/below.

eric-wieser: It's already released(4.x), but our rtd setup isn't finding it

ktbarrett: We are calling out `sphinx ~= 3.1` in `documentation/requirements.txt`. It switches to Sphinx 4.1.1 if I remove that, but it errors out. I'll chase that down.

ktbarrett:
The original idea behind using `@` was that
1. you don't have to import it
2. there is currently no way to type multiply-dispatched methods in Python except for builtin binary operators without creating a plugin

eric-wieser:
A plugin that monkey-patched https://github.com/python/mypy/blob/f98f78216ba9d6ab68c8e69c19e9f3c7926c5efe/mypy/operators.py would do the trick. Either way, let's leave that discussion to another PR, the rest of this looks fine (once #2641) is in

cocotb/types/array.py
```diff
@@ -128,59 +233,18 @@ class Array(Sequence):
...
             if len(self._value) != len(self._range):
                 raise ValueError(
                     "init value of length {!r} does not fit in {!r}".format(
                         len(self._value), self._range
                     )
                 )
-        elif value is None and range is not None:
-            if not isinstance(range, Range):
-                raise TypeError("range argument must be of type 'Range'")
-            self._value = [None] * len(range)
```
eric-wieser, line 165: I assume we decided this was not useful?

cocotb/types/array.py
```diff
@@ -247,18 +296,20 @@ def __getitem__(self, item):
...
         pass  # pragma: no cover
 
-    def __setitem__(self, item, value):
+    def __setitem__(
+        self, item: typing.Union[int, slice], value: typing.Union[T, typing.Iterable[T]]
+    ) -> None:
         if isinstance(item, int):
             idx = self._translate_index(item)
-            self._value[idx] = value
+            self._value[idx] = typing.cast(T, value)
```
eric-wieser, line 312: I'm surprised I've not seen these casts needed before, and that the mypy docs don't mention the need for them. At any rate, they're obviously at worst only unecessary anyway, and I'll take your word for it that mypy requires them.

cocotb/types/array.py
```diff
@@ -285,42 +336,20 @@ def __setitem__(self, item, value):
                 "indexes must be ints or slices, not {}".format(type(item).__name__)
             )
 
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, type(self)):
+            if len(self) != len(other):
+                return False
+            return all(a == b for a, b in zip(self, other))
```
eric-wieser, line 343: What's the motivation for not just doing `self == other`?

cocotb/types/array.py
```diff
@@ -285,42 +336,20 @@ def __setitem__(self, item, value):
...
-            stop = maxsize  # same default value used by Python lists
-        idx = self._value.index(value, start, stop)
-        return self._range[idx]
-
-    def count(self, value: Any) -> int:
-        """Return number of occurrences of *value*."""
-        return self._value.count(value)
+    def __concat__(self, other: ArrayABC[S]) -> "Array[typing.Union[T, S]]":
+        if isinstance(other, ArrayABC):
+            return Array(chain(self, other))
```
eric-wieser, line 351: This logic probably isn't right - if the other object is `AlternativeArray`, it's not clear to me that returning `Array` vs `AlternativeArray` is the right thing to do. Take for instance how `[] + ()` is an error in python.

cocotb/types/array.py
ktbarrett: Yeah, now that it's a homogenous type you need a default value, instead of trying to figure out how to provide that I just removed the feature.

cocotb/types/array.py
ktbarrett: In this instance it's necessary because `value` is `typing.Union[T, typing.Iterable[T]]` at this point. You need to narrow it somehow. `item` is already narrowed here to `int` from the `isinstance` checks.

cocotb/types/array.py
ktbarrett: Ah, missed that. Originally I had `__eq__` in the ABC, but after seeing a total lack of `__eq__` in the builtin ABC, I figured that was for a good reason and moved it, but didn't simplify it.

ktbarrett, line 343:
```suggestion
            return self._value == other._value
```

cocotb/types/array.py
ktbarrett: I did this on purpose so the result of concatenation is always `Array`.

This a part of my trying to find a path through the mypy minefield. You might say that if `LogicArray` was a subtype of `Array` you could do an `isinstance` check and just use `Self` types here. This requires direct inheritance because mypy does not understand virtual inheritance (`ABC.register`). That's fine for `LogicArray` because the implementation will likely be inherited, but not for subclasses like `Unsigned`, `Signed`, `Sfixed`, and `Ufixed`, etc which will likely use a different implementation.

So my options have been
1. `ArrayABC` and use virtual subclasses (which is what Python does), and forsake mypy
2. Fold `ArrayABC` back into `Array` and do direct inheritance and deal with that when implementation needs to change.
3. Only support concatenation with the current type, and likely only on `Array`, `LogicArray` and `BitArray`, requiring a cast otherwise.
4. Only support concatenation with subtypes and do direct inheritance from `Array` to `LogicArray` and `BitArray`, but leave out the integer and fixed point types, so they would need to be cast to `BitArray` first (even though they are `ArrayABC[Bit]`).

*Ideally, I'd like to see
```
def concat(Array[T], Array[S]) -> Array[T | S]
def concat(Array[T], LogicArray) -> Array[T | Logic]
def concat(Array[T], BitArray) -> Array[T | Bit]
def concat(LogicArray, LogicArray) -> LogicArray
def concat(LogicArray, BitArray) -> LogicArray
def concat(BitArray, BitArray) -> BitArray
```
In all orders and with `Unsigned`, `Signed`, `Sfixed`, `Ufixed` working in place of `BitArray`. This is fairly easy to accomplish in Python, but I am struggling trying to get this working in mypy.

eric-wieser: What does mypy do for `List[X] + List[Y]`?

ktbarrett: They have tried to change it to `List[X | Y]` [in typeshed](https://github.com/python/typeshed/pull/2404) before and it caused a [bunch of errors](https://github.com/python/mypy/issues/5492) in mypy due to how mypy handles list literals. Instead of fixing the problem, they just [reverted it](https://github.com/python/typeshed/pull/2653) =/ I [opened an issue](https://github.com/python/typeshed/issues/5756) about it.

ktbarrett: One more thing to note is that the operations I mentioned are not technically typesafe due to the invariance of mutable collection types. But I decided it was fine, practicality over purity, because I didn't feel like introducing a mutable vs immutable version of every type (imagine `FrozenArray[T_co]`, `Array[T]`, `FrozenLogicArray[LogicT_co]`, `LogicArray[LogicT]`, and corresponding `ArrayABC[T_co]` and `MutableArrayABC[T]` ABCs) since Python does not have easily mixed-in mutability/immutability via `const` or `mut`.

eric-wieser: My feeling is that probably 3 is the right choice from the list above.

eric-wieser: Mainly because it's easiest to add implicit conversions in a future release than it is to remove them - so I'd lean towards putting them off until we have the rest of the picture in place.

ktbarrett: Changed it so it only works with `Array` and not any `ArrayABC`.

marlonjames (approved):

PR #2641, opened 2021-07-22
Move to Sphinx 4 and replace cairosvgconverter

ktbarrett:
Removes the dependency restriction from Sphinx, so we can move to Sphinx 4. The implementation of `cairosvgconvert` in the repo stopped working with Sphinx 4 (due to deprecations). So instead we are adding a dependency on an contrib library that does the same thing and works with Sphinx 4.


documentation/requirements.txt
```diff
@@ -1,10 +1,11 @@
 # Python dependencies to build the documentation
 
 # != 3.5.0: https://github.com/sphinx-doc/sphinx/issues/8880
-Sphinx ~= 3.1, != 3.5.0
+Sphinx != 3.5.0
 breathe
```
eric-wieser, line 5: It's probably a good idea to pin to `~= 4.0` just so we don't end up with immediately breakage in random PRs when sphinx 5 is released.

documentation/requirements.txt
```diff
@@ -1,10 +1,11 @@
 # Python dependencies to build the documentation
 
 # != 3.5.0: https://github.com/sphinx-doc/sphinx/issues/8880
-Sphinx ~= 3.1, != 3.5.0
+Sphinx != 3.5.0
```
ktbarrett, line 4:
```suggestion
Sphinx ~= 4.0 != 3.5.0
```

documentation/requirements.txt
```diff
@@ -1,10 +1,11 @@
 # Python dependencies to build the documentation
 
 # != 3.5.0: https://github.com/sphinx-doc/sphinx/issues/8880
-Sphinx ~= 3.1, != 3.5.0
+Sphinx ~= 4.0 != 3.5.0
```
ktbarrett, line 4:
```suggestion
Sphinx ~= 4.0, != 3.5.0
```

documentation/requirements.txt
```diff
@@ -1,10 +1,11 @@
 # Python dependencies to build the documentation
 
 # != 3.5.0: https://github.com/sphinx-doc/sphinx/issues/8880
-Sphinx ~= 3.1, != 3.5.0
+Sphinx ~= 4.0, != 3.5.0
 breathe
 sphinx-rtd-theme
 cairosvg
+sphinxcontrib-svg2pdfconverter
```
imphil, line 8: `sphinxcontrib-svg2pdfconverter[CairoSVG]`

And remove the cairosvg dependency above.

documentation/requirements.txt
```diff
@@ -1,10 +1,11 @@
 # Python dependencies to build the documentation
 
 # != 3.5.0: https://github.com/sphinx-doc/sphinx/issues/8880
-Sphinx ~= 3.1, != 3.5.0
+Sphinx ~= 4.0, != 3.5.0
```
imphil, line 4: You can remove the `!= 3.5.0` part, you're already restricting to at least 4.0 with the previous clause. (The comma is an AND here.)

imphil (approved):
LGTM, thanks!

documentation/requirements.txt
eric-wieser:
```suggestion
Sphinx ~= 4.0
```

eric-wieser (approved):

documentation/requirements.txt
```diff
@@ -1,10 +1,11 @@
...
 breathe
 sphinx-rtd-theme
 cairosvg
```
eric-wieser, line 7:
```suggestion
```

documentation/requirements.txt
eric-wieser:
```suggestion
sphinxcontrib-svg2pdfconverter[CairoSVG]
```

PR #2645, opened 2021-07-25
Verilator debug improvements

ktbarrett:
Improvements to the Verilator makefile's debugging facilities.
* Fix bug where `-g` isn't passed to the build process
* Turn off optimizations for debugging
* Remove unnecessary `-std=c++11`, Verilator's makefiles automatically set an appropriate standard
* Add `SIM_CMD_PREFIX` to Verilator (also in #2644)
* make `debug` target automatically set `VERILATOR_SIM_DEBUG=1`


marlonjames (approved):
Looks reasonable. I verified the debug target with a cocotb test keeps the debug info for the Verilator VPI that gets built into the sim model.

PR #2660, opened 2021-08-12
Add start and start_soon scheduling functions

marlonjames:
Add `cocotb.start()` and `cocotb.start_soon()` for users to try now, before 2.0.

I updated the Coroutines page of the docs, but didn't change any other examples from `fork()`.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -171,6 +171,31 @@ def fork(coro: Union[RunningTask, Coroutine]) -> RunningTask:
     return scheduler.add(coro)
 
 
+def start_soon(coro: Union[RunningTask, Coroutine]) -> RunningTask:
+    """
+    Schedule a coroutine to be run concurrently.
+
+    Note that this is not an async function,
+    and the new task will not execute until the calling task yields control.
+
+    .. versionadded:: 1.6.0
+    """
+    return scheduler.start_soon(coro)
+
+
+async def start(coro: Union[RunningTask, Coroutine]) -> RunningTask:
+    """
+    Schedule a coroutine to be run concurrently, then yield control to allow pending tasks to execute.
+
+    The calling task will resume execution before control is returned to the simulator.
+
+    .. versionadded:: 1.6.0
+    """
+    task = scheduler.start_soon(coro)
+    await cocotb.triggers.NullTrigger()
+    return task
+
+
 # FIXME is this really required?
 _rlock = threading.RLock()
 
--- a/documentation/source/coroutines.rst
+++ b/documentation/source/coroutines.rst
@@ -5,18 +5,22 @@
    Async
 
 
-**********
-Coroutines
-**********
+********************
+Coroutines and Tasks
+********************
 
-Testbenches built using cocotb use coroutines. While the coroutine is executing
-the simulation is paused. The coroutine uses the :keyword:`await` keyword to
+Testbenches built using cocotb use Python coroutines.
+*Tasks* are cocotb objects that wrap coroutines
+and are used to schedule concurrent execution of the testbench coroutines.
+
+While active tasks are executing, the simulation is paused.
+The coroutine uses the :keyword:`await` keyword to
 block on another coroutine's execution or pass control of execution back to the
 simulator, allowing simulation time to advance.
 
 Typically coroutines :keyword:`await` a :class:`~cocotb.triggers.Trigger` object which
-indicates to the simulator some event which will cause the coroutine to be woken
-when it occurs.  For example:
+pauses the task, and indicates to the simulator some event which will cause the task resume execution.
+For example:
 
 .. code-block:: python3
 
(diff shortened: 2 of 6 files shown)


documentation/source/coroutines.rst
```diff
@@ -47,8 +51,25 @@ Coroutines can :keyword:`return` a value, so that they can be used by other coro
         if first == second:
             raise TestFailure("Signal did not change")
 
+Concurrent Execution
+====================
+
+Coroutines can be scheduled for concurrent execution with :func:`~cocotb.fork`, :func:`~cocotb.start`, and :func:`~cocotb.start_soon`.
+
+.. note::
+    :func:`~cocotb.start` and :func:`~cocotb.start_soon` are new scheduling functions.
```
cmarqu, line 60: The "new" here will age fast, maybe better express it with another `.. versionadded::` in the note?

documentation/source/coroutines.rst
```diff
@@ -92,8 +113,8 @@ Forked coroutines can be used in an :keyword:`await` statement to block until th
             else:
                 break
 
-Forked coroutines can be killed before they complete, forcing their completion before
-they'd naturally end.
+Tasks can be killed before they complete,
+forcing their completion before they'd naturally end.
```
cmarqu, line 117:
```suggestion
forcing their completion before they would naturally end.
```
(while we are here...)

cocotb/__init__.py
```diff
@@ -171,6 +171,31 @@ def fork(coro: Union[RunningTask, Coroutine]) -> RunningTask:
...
+
+async def start(coro: Union[RunningTask, Coroutine]) -> RunningTask:
+    """
+    Schedule a coroutine to be run concurrently, then yield control to allow pending tasks to execute.
+
+    The calling task will resume execution before control is returned to the simulator.
+
+    .. versionadded:: 1.6.0
+    """
+    task  = scheduler.start_soon(coro)
```
cmarqu, line 194:
```suggestion
    task = scheduler.start_soon(coro)
```

documentation/source/coroutines.rst
```diff
@@ -5,18 +5,22 @@
...
+Coroutines and Tasks
+********************
 
-Testbenches built using cocotb use coroutines. While the coroutine is executing
-the simulation is paused. The coroutine uses the :keyword:`await` keyword to
+Testbenches built using cocotb use Python coroutines.
+*Tasks* are cocotb objects that wrap coroutines
+and are used to schedule concurrent execution of the testbench coroutines.
+
+While active tasks are executing the simulation is paused.
```
cmarqu, line 16:
```suggestion
While active tasks are executing, the simulation is paused.
```
(?)

documentation/source/coroutines.rst
```diff
@@ -5,18 +5,22 @@
...
+The coroutine uses the :keyword:`await` keyword to
 block on another coroutine's execution or pass control of execution back to the
 simulator, allowing simulation time to advance.
 
 Typically coroutines :keyword:`await` a :class:`~cocotb.triggers.Trigger` object which
-indicates to the simulator some event which will cause the coroutine to be woken
-when it occurs.  For example:
+indicates to the simulator some event which will cause the task to be scheduled
+and the coroutine to resume execution when it occurs. For example:
```
ktbarrett, line 23: A little wordy and it doesn't explicitly state the current task pauses/blocks.
```suggestion
pauses the task, and indicates to the simulator some event which will cause the task to resume execution.
For example:
```

documentation/source/coroutines.rst
```diff
@@ -47,8 +51,25 @@ Coroutines can :keyword:`return` a value, so that they can be used by other coro
...
+
+Coroutines can be scheduled for concurrent execution with :func:`~cocotb.fork`, :func:`~cocotb.start`, and :func:`~cocotb.start_soon`.
+
+.. note::
+    :func:`~cocotb.start` and :func:`~cocotb.start_soon` are new scheduling functions.
+    They are future-compatible and are preferred over :func:`~cocotb.fork`.
+
+:func:`~cocotb.fork` schedules and executes the new coroutine immediately,
+returning control to the calling task after the new coroutine finishes or yields control.
+No other pending tasks are run.
```
ktbarrett, line 65: Though it may be beneficial to state that `fork` will be removed without a deprecation notice in the future. Or do you plan on adding a deprecation to `fork` now that the new interfaces exist?

documentation/source/coroutines.rst
```diff
@@ -47,8 +51,25 @@ Coroutines can :keyword:`return` a value, so that they can be used by other coro
...
 
-Coroutines can be used with :func:`~cocotb.fork` for concurrent operation.
+The *async* function :func:`~cocotb.start` schedules the new coroutine to be executed concurrently,
+then yields control to allow the new task (and any pending tasks) to run,
```
ktbarrett, line 68:
```suggestion
then yields control to allow the new task (and any other pending tasks) to run,
```

ktbarrett (commented):
I'm definitely liking it. I also noticed from your test we have the `create_task` primitive function, do we also want to make that a public API? So you can create the task and schedule it with `start_soon` vs `start` at some later point? Obviously not necessary in this PR, just looking for thoughts.

documentation/source/coroutines.rst
marlonjames: We didn't come to a conclusion on whether to remove `fork` or just alias it to `start_soon`.

marlonjames:
> I also noticed from your test we have the `create_task` primitive function, do we also want to make that a public API? So you can create the task and schedule it with `start_soon` vs `start` at some later point?

Yes I think that's probably a good idea. Unlike `asyncio`, it doesn't schedule the task, so we should note that in the docs.

documentation/source/coroutines.rst
ktbarrett: I happen to agree with removal. If that is the decision we need a deprecation before 1.6 is released ideally.

documentation/source/coroutines.rst
```diff
@@ -5,18 +5,22 @@
...
+
+While active tasks are executing, the simulation is paused.
+The coroutine uses the :keyword:`await` keyword to
 block on another coroutine's execution or pass control of execution back to the
 simulator, allowing simulation time to advance.
 
 Typically coroutines :keyword:`await` a :class:`~cocotb.triggers.Trigger` object which
-indicates to the simulator some event which will cause the coroutine to be woken
-when it occurs.  For example:
+pauses the task, and indicates to the simulator some event which will cause the task resume execution.
```
ktbarrett, line 22:
```suggestion
pauses the task, and indicates to the simulator some event which will cause the task to resume execution.
```
Didn't check myself...

ktbarrett (approved):

documentation/source/coroutines.rst
marlonjames: My typo

documentation/source/coroutines.rst
ktbarrett: I opened an issue about it.

PR #2667, opened 2021-08-14
Deprecate `<=` syntax for setting values on handles

### LINKED ISSUE #2658: Operator precedence of `<=` misleading when using `if-then`
In Python `<=` has higher precedence than `not`, `or`, `and`, `if-then`, and `lambda`. In cocotb `<=` is reused as an assignment operator, which leads to some highly surprising behavior when compared to the "normal" assignment operator `=`.

|   | Example              | Expected               | Actual                 | |
|---|----------------------|------------------------|------------------------|-----------------|
| 1 | `a <= b or c`        | `a <= (b or c)`        | `(a <= b) or c`        | _Edit: Mitigation implemented, see below_ |
| 2 | `a <= b and c`       | `a <= (b and c)`       | `(a <= b) and c`       | _Edit: Mitigation implemented, see below_ |
| 3 | `a <= b if c else d` | `a <= (b if c else d)` | `(a <= b) if c else d` |  |

I recently had my third encounter with this behavior. Although I had seen it twice before, it still took me some time to realize it was an operator precedence issue thanks to the pervasiveness of "normal" assignments.

# Partial solution
<details>
  <summary>Edit: this has already been implemented. With a recent version of cocotb only `if-then` is still an issue. Click to expand original text.</summary>

To prevent (1) and (2) `<=` could return a singleton `DontUse` like:

```python
class DontUse:
    def __bool__(self):
        raise ValueError("Don't use me like this dummy")

DontUse = DontUse()
```

This would make Python throw the error when trying to evaluate `or` and `and`. There's no such workaround available for `if-then` AFAIK.
</details>

# Full solution
I don't think there's much we can do. I think using `<=` should be deprecated and eventually removed entirely. There's very little benefit, but it can cause some pretty annoying debug sessions.  Assignment on simulator objects should probably replace `<=`'s current function. 

| Old               | New                |
|-------------------|--------------------|
| `dut.a <= b if c else d` | `dut.a = b if c else d`   |
| `a <= b if c else d`          | `a.value = b if c else d` |

The good news is that there's a good upgrade path:

1. Deprecation warning on `<=`
2. Hard error in `__le__`
3. Removal of `__le__` (?)

# Workarounds
1. Use parentheses:  `a <= (b if c else d)`
2. Use `.value`: `a.value = b if c else d`

### LINKED ISSUE #2681: Deprecate the ``<=`` syntax for setting values on handles
> I'm OK with this change. The `<=` notation was a neat idea, but it has three problems, IMO.
> 
> * It doesn't always work (and there's now way of fixing it). That's what triggered this PR, ultimately.
> * It's not obvious what it does for someone knowing Python. I know that I got the question multiple times what the strange syntax does, and every intro presentation of cocotb I've done has a paragraph on it.
> * It provides more than one way of doing things, which is bad for teaching cocotb.
> 
> With all that in mind, deprecating and ultimately removing the `<=` syntax makes sense to me.

Originally posted by @imphil in https://github.com/cocotb/cocotb/pull/2667#pullrequestreview-734467878

xref #2658.

ktbarrett:
Closes #2658. Closes #2681.

### DIFF
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -317,7 +317,7 @@ class HierarchyObject(RegionObject):
         if sub is not None:
             warnings.warn(
                 "Setting values on handles using the ``dut.handle = value`` syntax is deprecated. "
-                "Instead use the ``handle.value = value`` or ``handle <= value`` syntax",
+                "Instead use the ``handle.value = value`` syntax",
                 DeprecationWarning, stacklevel=2)
             sub.value = value
             return
@@ -479,6 +479,11 @@ class NonHierarchyObject(SimHandleBase):
         Example:
         >>> module.signal <= 2
         """
+        warnings.warn(
+            "Setting values on handles using the ``handle <= value`` syntax will be deprecated. "
+            "Instead use the ``handle.value = value`` syntax",
+            PendingDeprecationWarning,
+            stacklevel=2)
         self.value = value
         return _AssignmentResult(self, value)
 
@@ -587,7 +592,7 @@ class NonHierarchyIndexableObject(NonHierarchyObject):
         """Provide transparent assignment to indexed array handles."""
         warnings.warn(
             "Setting values on handles using the ``dut.handle[i] = value`` syntax is deprecated. "
-            "Instead use the ``handle[i].value = value`` or ``handle[i] <= value`` syntax",
+            "Instead use the ``handle[i].value = value`` syntax",
             DeprecationWarning, stacklevel=2)
         self[index].value = value
 
--- a/cocotb/handle.py
+++ b/cocotb/handle.py
@@ -480,9 +480,9 @@ class NonHierarchyObject(SimHandleBase):
         >>> module.signal <= 2
         """
         warnings.warn(
-            "Setting values on handles using the ``handle <= value`` syntax will be deprecated. "
+            "Setting values on handles using the ``handle <= value`` syntax is deprecated. "
             "Instead use the ``handle.value = value`` syntax",
-            PendingDeprecationWarning,
+            DeprecationWarning,
             stacklevel=2)
         self.value = value
         return _AssignmentResult(self, value)
--- a/documentation/source/newsfragments/2658.removal.rst
+++ b/documentation/source/newsfragments/2658.removal.rst
@@ -1,1 +1,0 @@
-Deprecate ``signal <= newval`` syntax for setting values on handles, due to its unreliability in corner cases which make cocotb harder to learn and teach. Use the ``signal.value = newval`` syntax instead.
--- a/documentation/source/newsfragments/2658.removal.rst
+++ b/documentation/source/newsfragments/2658.removal.rst
@@ -1,0 +1,1 @@
+Deprecate ``<=`` syntax for setting values on handles.
--- a/documentation/source/newsfragments/2658.removal.rst
+++ b/documentation/source/newsfragments/2658.removal.rst
@@ -1,1 +1,1 @@
-Deprecate ``<=`` syntax for setting values on handles.
+Deprecate ``signal <= newval`` syntax for setting values on handles, due to its unreliability in corner cases which make cocotb harder to learn and teach. Use the ``signal.value = newval`` syntax instead.
--- a/tests/test_cases/test_cocotb/test_deprecated.py
+++ b/tests/test_cases/test_cocotb/test_deprecated.py
@@ -193,7 +193,7 @@ async def test_assigning_setitem_syntax_deprecated(dut):
 
 @cocotb.test()
 async def test_assigning_less_than_syntax_deprecated(dut):
-    with assert_deprecated(PendingDeprecationWarning):
+    with assert_deprecated(DeprecationWarning):
         dut.stream_in_data <= 1
 
 
--- a/tests/test_cases/test_cocotb/test_deprecated.py
+++ b/tests/test_cases/test_cocotb/test_deprecated.py
@@ -10,6 +10,7 @@ from contextlib import contextmanager
 from common import assert_raises
 from typing import List
 from cocotb._sim_versions import IcarusVersion
+import pytest
 
 
 @contextmanager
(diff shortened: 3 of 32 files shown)


marlonjames:
The clock change breaks how `cycles` is supposed to work:
> cycles (int, optional): Cycle the clock *cycles* number of times,  or if ``None`` then cycle the clock forever.
>     Note: ``0`` is not the same as ``None``, as ``0`` will cycle no times.

cocotb/handle.py
```diff
@@ -479,6 +479,11 @@ def __le__(self, value):
         Example:
         >>> module.signal <= 2
         """
+        warnings.warn(
+            "Setting values on handles using the ``handle <= value`` syntax is deprecated. "
+            "Instead use the ``handle.value = value`` syntax",
+            DeprecationWarning,
```
eric-wieser, line 485: I wonder if we want to be more gentle about this and use `PendingDeprecationWarning` for one release first.

eric-wieser (commented):
I understand and agree with the technical motivation here, although I do worry this is going to alienate existing users. I'd also argue the readability goes down without `<=`, because it's harder to separate the signal name from the operator.

The writing seems to be on the wall that either we'll end up changing the dot notation (perhaps `dut._("module")._("signal").value = 42` or `dut.module.signal._.value = 42`) or changing the value assignment (`v_[dut.module.signal] = 42`). If we end up doing the second, we should try not to tell users to change their behavior twice; which would be introducing that syntax before this deprecation and pointing users to it.

Is it worth assembling a list of like 10 different assignment (and extraction) spellings and putting it to a user vote?

Let's keep this PR focused and not put things like the (incorrect!) clock change in at the same time.

eric-wieser (approved):
LGTM, thanks - just needs a release note.

I'd like @imphil to sign off on the idea of deprecating this (if it hasn't already happened in a meeting I didn't make).

documentation/source/newsfragments/2658.removal.rst
```diff
@@ -0,0 +1 @@
+Deprecate ``<=`` syntax for setting values on handles.
```
imphil, line 1: Please indicate the alternative and the motivation; users reading this will want to know more.

cocotb/handle.py
```diff
@@ -479,6 +479,11 @@ def __le__(self, value):
         Example:
         >>> module.signal <= 2
         """
+        warnings.warn(
+            "Setting values on handles using the ``handle <= value`` syntax will be deprecated. "
```
imphil, line 483: I know we heard otherwise as part of this review, and I'll leave it up to you to weigh the arguments, but my 2c: Either it's deprecated, or not. A "will be deprecated" is silly, IMO.

imphil (approved):
I'm OK with this change. The `<=` notation was a neat idea, but it has three problems, IMO.

- It doesn't always work (and there's now way of fixing it). That's what triggered this PR, ultimately.
- It's not obvious what it does for someone knowing Python. I know that I got the question multiple times what the strange syntax does, and every intro presentation of cocotb I've done has a paragraph on it.
- It provides more than one way of doing things, which is bad for teaching cocotb.

With all that in mind, deprecating and ultimately removing the `<=` syntax makes sense to me.

Thanks for pushing this forward @ktbarrett!

documentation/source/newsfragments/2658.removal.rst
ktbarrett: Will the link to the issue which is generated by towncrier be sufficient? #2658. Will add the alternative for sure.

imphil: #2658 feels to me like the final nail in the coffin. Maybe copy-paste the list of reasons that came to my mind in https://github.com/cocotb/cocotb/pull/2667#pullrequestreview-734467878?

cocotb/handle.py
ktbarrett: I changed this based on @eric-wieser's suggestion this should be a `PendingDeprecationWarning` instead of a `DeprecationWarning`, which to me say that it will become a `DeprecationWarning` in the future. I agree it should be full deprecation and don't really understand what `PendingDeprecationWarning` means. [Some core devs agree](https://discuss.python.org/t/pendingdeprecationwarning-is-really-useful/1038).

documentation/source/newsfragments/2658.removal.rst
ktbarrett: Multiline newsfragments are not supported by default by towncrier. If included it should be succinct, or I could change the link to a dedicated issue, or this PR?

imphil: How about something like:

" Deprecate ``signal <= newval`` syntax for setting values on handles for being unreliable in corner cases and making cocotb harder to learn and teach. Use ``signal.value <= newval`` instead."

marlonjames:
I think these 3 deprecations could use a little clean-up / harmonization.
https://cocotb--2667.org.readthedocs.build/en/2667/release_notes.html#deprecations-and-removals

cocotb/handle.py
ktbarrett: Changed wording to `is deprecated` and changed warning to `DeprecationWarning`. I don't feel like messing around with this stuff.

ktbarrett:
@marlonjames Harmonized.

tests/test_cases/test_cocotb/test_deprecated.py
```diff
@@ -188,3 +189,18 @@ async def test_assigning_setitem_syntax_deprecated(dut):
         with assert_raises(IndexError):
             # attempt to use __setitem__ syntax on signal that doesn't exist
             dut.stream_in_data[800000] = 1
+
+
+@cocotb.test()
+async def test_assigning_less_than_syntax_deprecated(dut):
+    with assert_deprecated(PendingDeprecationWarning):
```
ktbarrett, line 196:
```suggestion
    with assert_deprecated(DeprecationWarning):
```

marlonjames (approved):

PR #2668, opened 2021-08-14
Emit FutureWarning when an unwatched Task throws an exception

### LINKED ISSUE #2664: Add `FutureWarning` when a `fork`ed coroutine exception causes the test to end
I think we have decided that this behavior is not a good one, but instead we should capture the error in the Task's outcome and then throw a `RuntimeWarning` when the outcome is an exception, and it is never used before collection.

ktbarrett:
Closes #2664.


ktbarrett:
Fails on my machine...

ktbarrett:
@eric-wieser would you have any idea why the test is passing when it should be failing? Like I mentioned, it's failing for me locally.

ktbarrett:
Not entirely sure why the new line isn't covered, there is at least one test that causes a warning to fail (that is the point of `test_fork_exception`). Will investigate later.

PR #2670, opened 2021-08-15
Support using Coroutine objects and classes in coroutine APIs

### LINKED ISSUE #2647: Support `Coroutine` classes and objects wherever coroutines are
Currently we do runtime type checks in many places where coroutine functions and objects are accepted. These type checks prevent `Coroutine` classes and objects from being used.

ktbarrett:
Closes #2647. 

Adds support for:
* Using `collection.abc.Coroutine` objects in `cocotb.fork`, `cocotb.start`, `cocotb.start_soon`, and `cocotb.create_task`.
* Using a `collection.abc.Callable` which returns a `collection.abc.Coroutine` as a test function with `TestFactory` and `cocotb.test()`

Blocked by `cocotb.start` and friends coming in. While it can still come in, the docs will be broken until then.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -274,7 +274,7 @@ class RunningTest(RunningCoroutine):
     """
     The result of calling a :class:`cocotb.test` decorated object.
 
-    All this class does is change __name__ to show "Test" instead of "Task".
+    All this class does is change ``__name__`` to show "Test" instead of "Task".
     """
 
     _name: str = "Test"
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -30,6 +30,7 @@ import functools
 import inspect
 import os
 import warnings
+import collections.abc
 
 import cocotb
 from cocotb.log import SimLog
@@ -86,11 +87,12 @@ class RunningTask:
         triggers to fire.
     """
 
+    _name: str = "Task"  # class name of schedulable task
     _id_count = 0  # used by the scheduler for debug
 
     def __init__(self, inst):
 
-        if inspect.iscoroutine(inst):
+        if isinstance(inst, collections.abc.Coroutine):
             self._natively_awaitable = True
         elif inspect.isgenerator(inst):
             self._natively_awaitable = False
@@ -115,7 +117,7 @@ class RunningTask:
 
         self._task_id = self._id_count
         RunningTask._id_count += 1
-        self.__name__ = "Task %d" % self._task_id
+        self.__name__ = f"{type(self)._name} {self._task_id}"
         self.__qualname__ = self.__name__
 
     @lazy_property
@@ -268,6 +270,16 @@ class RunningCoroutine(RunningTask):
         self.funcname = parent._func.__name__
 
 
+class RunningTest(RunningCoroutine):
+    """
+    The result of calling a :class:`cocotb.test` decorated object.
+
+    All this class does is change __name__ to show "Test" instead of "Task".
+    """
+
+    _name: str = "Test"
+
+
 class coroutine:
     """Decorator class that allows us to provide common coroutine mechanisms:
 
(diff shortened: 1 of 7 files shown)


cocotb/decorators.py
```diff
@@ -115,7 +117,7 @@ def __init__(self, inst):
 
         self._task_id = self._id_count
         RunningTask._id_count += 1
-        self.__name__ = "Task %d" % self._task_id
+        self.__name__ = f"{type(self)._name} {self._task_id}"
```
ktbarrett, line 120: If we simply change the name to `type(self).__qualname__` here, we can get rid of the `RunningTest` class. However, that will change the value of `__name__` for tasks from `Task 0` to `RunningTask 0` and `cocotb.coroutine` decorated generators from `Task 0` to `RunningCoroutine 0`.

cocotb/regression.py
```diff
@@ -617,21 +618,21 @@ def _create_test(function, name, documentation, mod, *args, **kwargs):
...
-    _my_test.__module__ = mod.__name__
     return cocotb.test()(_my_test)
 
 
 class TestFactory:
     """Factory to automatically generate tests.
 
     Args:
-        test_function: The function that executes a test.
+        test_function: A callable that return the test Coroutine.
```
cmarqu, line 635:
```suggestion
        test_function: A callable that returns the test Coroutine.
```
Or maybe even an uppercase `Callable`?

cocotb/decorators.py
```diff
@@ -268,6 +270,16 @@ def __init__(self, inst, parent):
         self.funcname = parent._func.__name__
 
 
+class RunningTest(RunningCoroutine):
+    """
+    The result of calling a :class:`cocotb.test` decorated object.
+
+    All this class does is change __name__ to show "Test" instead of "Task".
```
cmarqu, line 277:
```suggestion
    All this class does is change ``__name__`` to show "Test" instead of "Task".
```

documentation/source/newsfragments/2647.feature.2.rst
```diff
@@ -0,0 +1 @@
+:class:`~cocotb.regression.TestFactory` and :class:`cocotb.test` now accepts as a test function any :class:`collections.abc.Callable` object which returns a :class:`collections.abc.Coroutine`.
```
ktbarrett, line 1:
```suggestion
:class:`~cocotb.regression.TestFactory` and :class:`cocotb.test` now accept any :class:`collections.abc.Callable` object which returns a :class:`collections.abc.Coroutine` as a test function.
```

cocotb/regression.py
```diff
@@ -617,21 +618,21 @@ def _create_test(function, name, documentation, mod, *args, **kwargs):
     Returns:
         Decorated test function
     """
+    # causes early type check, during call to generate_tests()
+    test = cocotb.scheduler.create_task(function(cocotb.top, *args, **kwargs))
```
marlonjames, line 622: I think it's better to leave the task creation as part of the Regression Manager, so we have one place that the root handle is passed in. If you want the check at this point, you could create the coroutine object here but not keep it.

ktbarrett: The issue if that the callable may have side effects. Calling it more than once may not be permissible. I'm not against simply removing the check. `cocotb.test()` does not check the type is correct until the test is collected, I don't see why `TestFactory` should have ever been different (I think I was the one who added the checks so feel free to call me inconsistent).

cocotb/regression.py
```diff
@@ -617,21 +618,19 @@ def _create_test(function, name, documentation, mod, *args, **kwargs):
     Returns:
         Decorated test function
     """
+
+    @wraps(function)
     async def _my_test(dut):
-        await function(dut, *args, **kwargs)
+        await function(cocotb.top, *args, **kwargs)
```
marlonjames, line 624: I know `cocotb.top` is equivalent (right now), but I think it's better to leave the handle passing to the Regression Manager.

```suggestion
        await function(dut, *args, **kwargs)
```

marlonjames (approved):

PR #2675, opened 2021-08-18
Add Python 3.10 to CI

cmarqu:
rc1 is out: https://docs.python.org/3.10/whatsnew/3.10.html
This is with `may_fail` for now.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -100,6 +100,13 @@ jobs:
             python-version: 3.9
             os: ubuntu-20.04
 
+          - sim: icarus
+            sim-version: apt
+            lang: verilog
+            python-version: "3.10.0-alpha - 3.10"
+            os: ubuntu-20.04
+            may_fail: true
+
             # Test Icarus dev on Ubuntu
 
           - sim: icarus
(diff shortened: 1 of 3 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -100,6 +100,12 @@ jobs:
             python-version: 3.9
             os: ubuntu-20.04
 
+          - sim: icarus
+            sim-version: apt
+            lang: verilog
+            python-version: "3.10"
```
marlonjames, line 106: Maybe this works?
```suggestion
            python-version: "3.10.0-rc - 3.10"
```

cmarqu: Ah yes, range syntax, thanks. I'm going to use https://github.com/actions/setup-python/issues/20#issuecomment-896221379 now as that is proven to work.

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 # when changing this list, adapt CONTRIBUTING.md accordingly:
```
ktbarrett, line 2: Does this not count as "supporting" 3.10? If it is you should add the new version to CONTRIBUTING.md per the comment.

ktbarrett (approved):

PR #2677, opened 2021-08-19
Make `create_task` public

### LINKED ISSUE #2665: Make `cocotb.scheduler.create_task` a public API
By creating a top-level alias `cocotb.create_task`. This function will turn a coroutine into a Task, but will not schedule it. Later the user can schedule that Task using `cocotb.start_soon` or `cocotb.start`.

ktbarrett:
Closes #2665. Blocked by #2670.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -196,6 +196,18 @@ async def start(coro: Union[RunningTask, Coroutine]) -> RunningTask:
     return task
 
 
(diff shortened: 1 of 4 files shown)


marlonjames (commented):
Add `create_task` to the docs by the other functions.
It might be worth using a note or warning in the docstring about not scheduling the new task, given the behavior is different than `asyncio`.

marlonjames (approved):

PR #2678, opened 2021-08-19
Replace deprecated distutils

### LINKED ISSUE #2676: Replace distutils
We should replace our usage of `distutils` soonish:  `DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives`

This is reported with Python 3.10, see https://github.com/cocotb/cocotb/pull/2675/checks?check_run_id=3364741779#step:25:199 for an example.

cmarqu:
distutils.version has been vendored in.

Closes #2676.

### DIFF
--- a/cocotb/_vendor/distutils_version.py
+++ b/cocotb/_vendor/distutils_version.py
@@ -1,3 +1,10 @@
+# Copyright (c) 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010,
+# 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021 Python Software Foundation;
+# All Rights Reserved
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
 #
 # distutils/version.py
 #
--- a/cocotb/_vendor/distutils_version.py
+++ b/cocotb/_vendor/distutils_version.py
@@ -1,0 +1,339 @@
+#
+# distutils/version.py
+#
+# Implements multiple version numbering conventions for the
+# Python Module Distribution Utilities.
+#
+# $Id$
+#
+
+"""Provides classes to represent module version numbers (one class for
+each style of version numbering).  There are currently two such classes
+implemented: StrictVersion and LooseVersion.
+Every version number class implements the following interface:
+  * the 'parse' method takes a string and parses it to some internal
+    representation; if the string is an invalid version number,
+    'parse' raises a ValueError exception
+  * the class constructor takes an optional string argument which,
+    if supplied, is passed to 'parse'
+  * __str__ reconstructs the string that was passed to 'parse' (or
+    an equivalent string -- ie. one that will generate an equivalent
+    version number instance)
+  * __repr__ generates Python code to recreate the version number instance
+  * _cmp compares the current instance with either another instance
+    of the same class or a string (which will be parsed to an instance
+    of the same class, thus must follow the same rules)
+"""
+
+import re
+
+class Version:
+    """Abstract base class for version numbering classes.  Just provides
+    constructor (__init__) and reproducer (__repr__), because those
+    seem to be the same for all version numbering classes; and route
+    rich comparisons to _cmp.
+    """
+
+    def __init__ (self, vstring=None):
+        if vstring:
+            self.parse(vstring)
+
+    def __repr__ (self):
+        return "%s ('%s')" % (self.__class__.__name__, str(self))
+
+    def __eq__(self, other):
+        c = self._cmp(other)
+        if c is NotImplemented:
+            return c
+        return c == 0
+
+    def __lt__(self, other):
+        c = self._cmp(other)
+        if c is NotImplemented:
+            return c
+        return c < 0
+
+    def __le__(self, other):
+        c = self._cmp(other)
+        if c is NotImplemented:
+            return c
+        return c <= 0
+
+    def __gt__(self, other):
+        c = self._cmp(other)
+        if c is NotImplemented:
+            return c
+        return c > 0
+
+    def __ge__(self, other):
+        c = self._cmp(other)
+        if c is NotImplemented:
+            return c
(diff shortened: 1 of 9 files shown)


cocotb_build_libs.py
```diff
@@ -6,17 +6,55 @@
...
 import textwrap
 
 from setuptools import Extension
-from distutils.spawn import find_executable
 from setuptools.command.build_ext import build_ext as _build_ext
-from distutils.file_util import copy_file
-from distutils.ccompiler import get_default_compiler
 from typing import List
 
+# vendored in from distutils/ccompiler.py:
```
ktbarrett, line 18: Are you sure this is necessary? I was under the impression that `setuptools` has a copy of `distutils` which is what gets imported when setup.py is run.

ktbarrett: See [this](https://github.com/pypa/setuptools/blob/0c302d232dd1a1f4ec531050532a84274edac2f4/_distutils_hack/__init__.py#L49) import hack in setuptools. I don't think we need to worry about vendoring in distutils code that is used in setup.py. We must only ensure that setuptools is imported first.

cmarqu: Ah yes, this works. Let's see if CI confirms...

ktbarrett:
Surprise! Unexpected behavior! It doesn't seem to be omitting `cocotb/_vendor/distutil_version.py` from the coverage despite the `cocotb/_vendor/*` omit rule in `setup.cfg`.

cmarqu:
> Surprise! Unexpected behavior! It doesn't seem to be omitting `cocotb/_vendor/distutil_version.py` from the coverage despite the `cocotb/_vendor/*` omit rule in `setup.cfg`.

Hard to believe that this should be a bug in Coverage.py...

ktbarrett:
My experience tell me that there is probably a "very good reason" why it works this way.

cmarqu:
Hmm, is it tox running this from a different path so that the relative location is different? Trying with a wildcard.

setup.cfg
```diff
@@ -41,8 +41,8 @@ testpaths =
 
 [coverage:run]
```
ktbarrett, line 42: I don't this this omit is in the right section.

themperek:
We do not want a dependency on https://packaging.pypa.io/ ?

ktbarrett:
@themperek We don't want a dependency and it won't work anyways since there is no `LooseVersion`. They only have classes for "clean" semantic versioning per PEP 440, which simulators tend to not follow.

setup.cfg
cmarqu: I'll try `coverage ... --debug=trace` later which should help in finding the problem, as per https://coverage.readthedocs.io/en/latest/cmd.html#diagnostics-coverage-debug

cmarqu:
Huh, strange behavior. I added trace debugging, and locally, I'm getting output like
```
Source matching against trees <TreeMatcher ['/opt/tools/cocotb/cocotb']>
Omit matching: <FnmatchMatcher ['*/cocotb/config.py', '*/cocotb/_vendor/*']>
Not tracing '/tools/anaconda/2019.10-python3.7/Linux-x86_64/lib/python3.7/threading.py': falls outside the --source spec
[...]
Not tracing '/opt/tools/cocotb/tests/pytest/test_utils.py': falls outside the --source spec
Tracing '/opt/tools/cocotb/cocotb/_sim_versions.py'
Not tracing '/opt/tools/cocotb/cocotb/_vendor/__init__.py': is inside an --omit pattern
Not tracing '/opt/tools/cocotb/cocotb/_vendor/distutils_version.py': is inside an --omit pattern
```
but not in CI (also the first two lines are not there). This is with coverage 5.5 both times.

ktbarrett:
@cmarqu Is there a reason to not bring this in for 1.6? We can figure out the coverage problem later if need be.

cmarqu:
I'm fine with merging this as is.

ktbarrett:
There is a failing test.

cmarqu:
@themperek Do you see what's wrong here?: https://github.com/cocotb/cocotb/pull/2678/checks?check_run_id=3402399790#step:25:67
It's probably https://github.com/cocotb/cocotb/pull/2678/files#diff-1e9f10ea33562835a183dcbff5c9431b590b33053c7c59463cd83bfcba7dde2eR581-R584 that is behaving slightly different now.

themperek:
> @themperek Do you see what's wrong here?: https://github.com/cocotb/cocotb/pull/2678/checks?check_run_id=3402399790#step:25:67
> It's probably https://github.com/cocotb/cocotb/pull/2678/files#diff-1e9f10ea33562835a183dcbff5c9431b590b33053c7c59463cd83bfcba7dde2eR581-R584 that is behaving slightly different now.

It looks like `config_vars` has changed. Maybe try `os.environ['LDSHARED']`? 
I will try to dig this.

themperek:
Maybe also can try `ext.extra_link_args` here https://github.com/cocotb/cocotb/blob/master/cocotb_build_libs.py#L232

cocotb/_vendor/distutils_version.py
```diff
@@ -0,0 +1,339 @@
+#
+# distutils/version.py
+#
+# Implements multiple version numbering conventions for the
+# Python Module Distribution Utilities.
+#
+# $Id$
+#
```
ktbarrett, line 8: Does PSF require use to include the license with this file? That will also transitively reclassify the entire project as I think the PSF license is stricter than MIT. @imphil Are you more familiar with license issues?

imphil: No license needs to be included, but please prepend our standard (cocotb) SPDX header and add a copyright line following https://github.com/python/cpython/blob/main/LICENSE#L86-L88.

ktbarrett: I think I got that right.

cmarqu:
Thanks for figuring all this out!

ktbarrett (approved):

ktbarrett:
@cmarqu And thank you for pushing for Python 3.10 support in 1.6!

PR #2683, opened 2021-08-24
Adding support for net/wire discovery under SV interface

avidan-efody:
Simulators such as vcs/ius return vpiInterface when they get to an SV interface. when the simulator was returning vpiInterface as the type of the underlying HDL object cocotb would give an error saying this type is not supported, and therefore the signals inside the interface could not be found/driven. Instead, map this type to the basic vpiInstance which is also shared with vpiModule. Moving forward we could move most of vpiModule members to vpiInstance, and add support for vpiProgram and other elements currently not supported.

### DIFF
--- a/tests/test_cases/test_sv_interface/Makefile
+++ b/tests/test_cases/test_sv_interface/Makefile
@@ -1,0 +1,14 @@
+# This file is public domain, it can be freely copied without restrictions.
+# SPDX-License-Identifier: CC0-1.0
+
+TOPLEVEL_LANG ?= verilog
+VERILOG_SOURCES = $(shell pwd)/top.sv
+MODULE = test_sv_if
+TOPLEVEL = top
+
+ifneq ($(shell echo $(SIM) | tr A-Z a-z),verilator)
+all::
+clean::
+else
+include $(shell cocotb-config --makefiles)/Makefile.sim
+endif
--- a/cocotb/share/lib/vpi/VpiCbHdl.cpp
+++ b/cocotb/share/lib/vpi/VpiCbHdl.cpp
@@ -545,11 +545,17 @@ VpiNextPhaseCbHdl::VpiNextPhaseCbHdl(GpiImplInterface *impl)
 
 decltype(VpiIterator::iterate_over) VpiIterator::iterate_over = [] {
     /* for reused lists */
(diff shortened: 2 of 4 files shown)


tests/test_cases/test_sv_interface/Makefile
```diff
@@ -0,0 +1,14 @@
+# This file is public domain, it can be freely copied without restrictions.
+# SPDX-License-Identifier: CC0-1.0
+
+TOPLEVEL_LANG ?= verilog
+VERILOG_SOURCES = $(shell pwd)/top.sv
+MODULE = test_sv_if
+TOPLEVEL = top
+
+ifneq ($(shell echo $(SIM) | tr A-Z a-z),verilator)
```
ktbarrett, line 9: Allow this test to run with Riviera and Questa. All verilator tests are allowed to fail since it's broken upstream rn, so this isn't being tested.

cmarqu: Also Xcelium/IUS and VCS I guess.

cmarqu:
Thanks for this contribution! \o/

avidan-efody:
> Thanks for this contribution! \o/

Thanks for reviewing. Will get those fixed!

tests/test_cases/test_sv_interface/Makefile
```diff
@@ -0,0 +1,21 @@
...
+MODULE = test_sv_if
+TOPLEVEL = top
+
+# simulator behavior:
+# icarus sees SV interfaces as vpiModule but doesn't discover signals inside them
+# verilator sees SV interfaces as vpiModule and discovers signals inside them
+# commercial simulators all see SV interfaces as vpiInterface and discover signals inside them
+
+
+ifeq ($(filter questa xcelium vcs riviera verilator icarus,$(shell echo $(SIM) | tr A-Z a-z)),)
```
ktbarrett, line 15: This test is running when `TOPLEVEL_LANG=vhdl` is set and that is what is causing the Questa VHDL test to fail. Please guard against that.

avidan-efody: @ktbarrett yep, I did...Think it is ok now right? I can't view the build logs on azure...

PR #2685, opened 2021-08-25
Add rounding modes for time operations

### LINKED ISSUE #2684: Handling imprecise time values
## Problem

A couple users (to my memory @alexforencich and @jwprice100) have lamented about how timing is handled in cocotb. When one creates a time wait with `Timer` with a value that cannot be accurately represented in the simulator, a `ValueError` is thrown. Additionally, the useful helper method `get_sim_steps` also throws a `ValueError` if a value is given which cannot be accurately represented, which is *very* limiting.

## Proposed Solutions

There are several suggestions, feel free to agree/debate each independently.

Change `get_sim_steps` to take an argument to decide how to handle imprecise values by rounding. The round modes would be:
* `error`: throw `ValueError` like is currently done (default)
* `round`: round to the nearest step using `round` builtin
* `ceil`: round up to the nearest step using `math.ceil`
* `floor`: round down to the nearest step using `math.floor`

This argument could also be passed to `Timer`.

An environment variable, a `cocotb` module global variable, or an attribute on the `get_sim_steps` object could be added to globally affect the default rounding mode.

The default of `error` could be changed in 2.0 to be something more sensible like `ceil`. If this was to be done, a `FutureWarning` about the default should be added. This change would not break existing passing tests, but may make existing failing tests pass and would be difficult to change in the future without breaking tests, so it must be considered carefully.

ktbarrett:
Closes #2684. Split into multiple commits. Can take or leave commits we don't agree with off the end.
1. Adds rounding mode argument to `utils.get_sim_steps`
2. Adds rounding mode argument to `Timer`
3. Adds ability to change globally the default rounding mode of `Timer`
4. ~~Adds `FutureWarning` about the default rounding mode changing to `ceil`.~~

I think `ceil` is probably the best rounding mode if rounding is to be preferred over erroring by default. To me, when waiting for some amount of time you are waiting *at least* that amount of time. By allowing to round down you are loosing time and it may put you before the event you intended. Additionally, allowing the possibility of rounding down may cause rounds to `0`, which is known to break certain simulators. Of course because of the argument, the user can change the mode when it makes sense (like random waits for modelling flow control).

### DIFF
--- a/cocotb/utils.py
+++ b/cocotb/utils.py
@@ -130,7 +130,7 @@ def get_sim_steps(
 
     Raises:
         ValueError: if the value cannot be represented accurately in terms of simulator
-            time steps wound *round_mode* is ``"error"``.
+            time steps when *round_mode* is ``"error"``.
 
     .. versionchanged:: 1.5
         Support ``'step'`` as the the *units* argument to mean "simulator time step".
--- a/cocotb/utils.py
+++ b/cocotb/utils.py
@@ -111,6 +111,12 @@ def get_sim_steps(
 ) -> int:
     """Calculates the number of simulation time steps for a given amount of *time*.
 
+    When *round_mode* is ``"error"``, a :exc:`ValueError` is thrown if the value cannot
+    be accurately represented in terms of simulator time steps.
+    When *round_mode* is ``"round"``, ``"ceil"``, or ``"floor"``, the corresponding
+    rounding function from the standard library will be used to round to a simulator
+    time step.
+
     Args:
         time: The value to convert to simulation time steps.
         units: String specifying the units of the result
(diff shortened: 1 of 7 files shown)


ktbarrett:
Turns out issuing multiple warnings per call really complicates the tests. I'll leave that for a follow on.

cocotb/utils.py
```diff
@@ -102,41 +104,67 @@ def get_time_from_sim_steps(steps, units):
     return _ldexp10(steps, _get_simulator_precision() - _get_log_time_scale(units))
 
 
-def get_sim_steps(time, units="step"):
+def get_sim_steps(
+    time: Union[Real, Decimal],
+    units: str = "step",
+    round_mode: str = "error"
```
eric-wieser, line 110: In my opinion this should probably be keyword-only

PR #2687, opened 2021-08-31
Allow use of make -f

suzizecat:
Following gitter discussion of the 1st of july 2021 (https://gitter.im/cocotb/Lobby?at=60dd54a0110daa37b115e09c),

This pull request allows to use a custom makefile through the use of  `make -f filename.mk`
As `filename.mk` name will be prepended to the special `MAKEFILE_LIST` variable, we select it through `firstword` and force its for the internal make call. 

The limitation might be if the user have one more level of makefile inclusion.

### DIFF
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -37,7 +37,7 @@ COCOTB_MAKEFILE_INC_INCLUDED = 1
 .PHONY: sim
 sim:
 	-@rm -f $(COCOTB_RESULTS_FILE)
-	$(MAKE) $(COCOTB_RESULTS_FILE)
+	$(MAKE) -f $(firstword $(MAKEFILE_LIST)) $(COCOTB_RESULTS_FILE)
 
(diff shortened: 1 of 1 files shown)


ktbarrett (commented):
`MAKEFILE_LIST` is in gmake 3.80 which is the standard on RHEL 7 machines, which is the oldest distro we support (now that Ubuntu 16.04 is EOL). So this seems safe to bring in.

It would be nice to see a test, but if you have already been using it, I'll take your word. I don't think this is significant enough to warrant a newsfragment.

suzizecat:
How should I present a test ?
As the `filename.mk` file is exactly as your usual `Makefile` (beside the name), you can try it with any test you already have.
If you don't specify a name, it is equivalent to using `-f Makefile` and the first name in `MAKEFILE_LIST` will indeed be `Makefile`. 

Please let me know if/how can I provide you such a test.

ktbarrett:
Tested locally. It works.

ktbarrett (approved):

cmarqu (approved):

PR #2689, opened 2021-09-02
Make round_mode on Timer and get_sim_steps keyword-only

ktbarrett:
Per https://github.com/cocotb/cocotb/pull/2685#discussion_r697994287.


eric-wieser (approved):
Thanks!

PR #2691, opened 2021-09-03
Cleanup to cocotb.utils

ktbarrett:
* Add support for `units="step"` to `cocotb.utils.get_sim_time`
* Deprecate hexdiffs and hexdump totally. They were used by Scoreboard, which we are no longer supporting in 2.0, and are available in the `scapy` package
* nitpicks

### DIFF
--- a/cocotb/utils.py
+++ b/cocotb/utils.py
@@ -99,13 +99,13 @@ def _ldexp10(frac, exp):
         return frac / (10 ** -exp)
 
 
-def get_time_from_sim_steps(steps, units):
+def get_time_from_sim_steps(steps: int, units: str) -> int:
     """Calculates simulation time in the specified *units* from the *steps* based
     on the simulator precision.
 
     Args:
-        steps (int): Number of simulation steps.
-        units (str): String specifying the units of the result
+        steps: Number of simulation steps.
+        units: String specifying the units of the result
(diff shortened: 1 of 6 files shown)


cocotb/utils.py
```diff
@@ -53,28 +53,35 @@ def _get_simulator_precision():
...
 # Simulator helper functions
-def get_sim_time(units=None):
+def get_sim_time(units: str = "step") -> int:
     """Retrieves the simulation time from the simulator.
 
     Args:
-        units (str or None, optional): String specifying the units of the result
+        units: String specifying the units of the result
             (one of ``None``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``).
             ``None`` will return the raw simulation time.
```
marlonjames, line 67:
```suggestion
            (one of ``'step'``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``).
            ``'step'`` will return the raw simulation time.
 
            .. deprecated:: 1.6.0
                Using None as the *units* argument is deprecated, use ``'step'`` instead.
```

ktbarrett: `get_sim_steps` never got this section. Guess I should add it there as well.

ktbarrett:
Added newsfragments.

cmarqu (approved):

ktbarrett:
Added the actual deprecation warnings to `hexdump` and `hexdiffs`.

PR #2692, opened 2021-09-03
Cleanup cocotb.result

ktbarrett:
* Deprecates the stdout and stderr attributes on `TestComplete` exceptions
* Deprecates `TestFailure`

The stdout and stderr functions only seemed to be used by `_raise_error`, which is deprecated and otherwise unused.

PR #2695, opened 2021-09-04
Remove `TestComplete`'s from tests

ktbarrett:
Replaces `raise TestError` and `raise TestFailure` with `assert` statements so that `TestFailure` can be deprecated.


ktbarrett:
Already reviewed in #2692.

PR #2708, opened 2021-09-14
Fix logging code in test examples

### LINKED ISSUE #2705: Broken string formatting in quickstart guide
When running the first example in the [quickstart guide](https://docs.cocotb.org/en/stable/quickstart.html), a type error is raised by the following line:
```
dut._log.info("my_signal_1 is", dut.my_signal_1.value)
```
Changing the line to this prevents the error:
```
dut._log.info("my_signal_1 is %d", dut.my_signal_1.value)
```
There are several other examples of this in the guide, ideally they should all be fixed in the same way.

cmarqu:
Closes #2705.

### DIFF
--- a/documentation/source/quickstart.rst
+++ b/documentation/source/quickstart.rst
@@ -37,6 +37,7 @@ Let's create a test file ``test_my_design.py`` containing the following:
     import cocotb
     from cocotb.triggers import Timer
 
+
     @cocotb.test()
     async def my_first_test(dut):
         """Try accessing the design."""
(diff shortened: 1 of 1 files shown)


cmarqu:
@Ben1152000 I have only found these two occurrences in the docs, and used the `%` syntax for the fix.
Does that work for you? Did you find more problematic code in the docs?

Ben1152000:
@cmarqu This seems good! I didn't find any other issues.

ktbarrett (approved):
Looks correct now.

PR #2710, opened 2021-09-15
Move code in quickstart doc to its own example

cmarqu:
This makes the code testable in CI.

### DIFF
--- a/examples/doc_examples/quickstart/Makefile
+++ b/examples/doc_examples/quickstart/Makefile
@@ -1,0 +1,20 @@
+# This file is public domain, it can be freely copied without restrictions.
+# SPDX-License-Identifier: CC0-1.0
+
+# Makefile
+
+# defaults
+SIM ?= icarus
+TOPLEVEL_LANG ?= verilog
+
+VERILOG_SOURCES += $(PWD)/my_design.sv
+# use VHDL_SOURCES for VHDL files
+
+# TOPLEVEL is the name of the toplevel module in your Verilog or VHDL file
+TOPLEVEL = my_design
+
+# MODULE is the basename of the Python test file
+MODULE = test_my_design
+
+# include cocotb's make rules to take care of the simulator setup
+include $(shell cocotb-config --makefiles)/Makefile.sim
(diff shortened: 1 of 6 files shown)


ktbarrett (commented):
You aren't guarding against running the example in VHDL regressions, which is why the VHDL tests are failing. But *how* it's failing is rather interesting... Questa blows up because of mixed language funniness, but Riviera seems to be incorrectly generating the `GPI_EXTRA` argument.

cmarqu:
The "guarding" in form of hardcoding `TOPLEVEL_LANG` is in.
CI is broken differently currently, apparently the `xunitparser` package?

ktbarrett:
Let's try one more time.

examples/doc_examples/quickstart/Makefile
```diff
@@ -0,0 +1,20 @@
+# This file is public domain, it can be freely copied without restrictions.
+# SPDX-License-Identifier: CC0-1.0
+
+# Makefile
+
+# defaults
+SIM ?= icarus
+TOPLEVEL_LANG = verilog
```
ktbarrett, line 8: Check out https://github.com/cocotb/cocotb/blob/master/tests/test_cases/test_sv_interface/Makefile. This stills runs the verilog test during the VHDL regression and can be overloaded to break using `make TOPLEVEL_LANG=vhdl`.

This only works for Questa and Riviera because they support both languages and GHDL skips tests when `TOPLEVEL_LANG` is set to `verilog`, which seems wrong to me... it should fail instead.

cmarqu: I won't be able to make that change for the next two weeks, could you take over unless it can wait until that time?

ktbarrett: Yeah I can do that.

ktbarrett: It does not seem I have permission to push to your branch.

cmarqu:
Strange, "Allow edits by maintainers" is enabled.

ktbarrett:
@cmarqu Is this good by you? I added the guard around the inclusion in `example/Makefile` so the Makefile is untouched, forgot that file was being included in the docs.

cmarqu:
> @cmarqu Is this good by you?

Yes, thanks very much!

ktbarrett (approved):

PR #2711, opened 2021-09-15
Update docs to use the `sig.value = value` syntax

cmarqu:
Refs #2667.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ async def test_dff_simple(dut):
 
     for i in range(10):
         val = random.randint(0, 1)
-        dut.d <= val  # Assign the random value val to the input port d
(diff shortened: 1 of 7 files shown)


cmarqu:
The same change in `quickstart.rst` will come in via #2710.

ktbarrett (approved):
Looks good. I think in addition to updating the quickstart we will need to scrub the docstrings as well.

cmarqu:
My grepping didn't find any docstrings to fix, just two comments and one additional occurrence in code.
Did you see a docstring that needed fixing?

ktbarrett:
Looks like I fixed the ones I'm thinking about in #2667.

PR #2713, opened 2021-09-16
Convert fork usage to start and start_soon

marlonjames:
Replace almost all instances of `cocotb.fork` with `cocotb.start` or `cocotb.start_soon`.

I'm looking for feedback on the following:
1. We probably want to keep some of the scheduler tests using fork, or add new ones.
2. I'm not sure if it's better to recommend `start` or `start_soon` in the docs. For clocks there is mixed usage that we may want to change to be consistent.

### DIFF
--- a/cocotb/clock.py
+++ b/cocotb/clock.py
@@ -51,8 +51,10 @@ class BaseClock:
 class Clock(BaseClock):
     r"""Simple 50:50 duty cycle clock driver.
 
-    Instances of this class should call its :meth:`start` method and :func:`fork` the
-    result.  This will create a clocking thread that drives the signal at the
+    Instances of this class should call its :meth:`start` method
+    and pass the coroutine object to one of the functions in :ref:`task-management`.
+
+    This will create a clocking task that drives the signal at the
     desired period/frequency.
 
     Example:
@@ -60,7 +62,7 @@ class Clock(BaseClock):
     .. code-block:: python
 
         c = Clock(dut.clk, 10, 'ns')
-        cocotb.fork(c.start())
+        await cocotb.start(c.start())
 
     Args:
         signal: The clock pin/signal to be driven.
@@ -72,7 +74,7 @@ class Clock(BaseClock):
             the timestep is determined by the simulator (see :make:var:`COCOTB_HDL_TIMEPRECISION`).
 
     If you need more features like a phase shift and an asymmetric duty cycle,
-    it is simple to create your own clock generator (that you then :func:`fork`):
+    it is simple to create your own clock generator (that you then :func:`~cocotb.start`):
 
     .. code-block:: python
 
@@ -102,7 +104,7 @@ class Clock(BaseClock):
                 await Timer(low_delay, units="ns")
 
         high_delay = low_delay = 100
-        cocotb.fork(custom_clock())
+        await cocotb.start(custom_clock())
         await Timer(1000, units="ns")
         high_delay = low_delay = 10  # change the clock speed
         await Timer(1000, units="ns")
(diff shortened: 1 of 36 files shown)


cocotb/clock.py
```diff
@@ -51,16 +51,18 @@ def log(self):
...
+    This will create a clocking task that drives the signal at the
     desired period/frequency.
 
     Example:
 
     .. code-block:: python
 
         c = Clock(dut.clk, 10, 'ns')
-        cocotb.fork(c.start())
+        cocotb.start_soon(c.start())
```
ktbarrett, line 65: You use `start_soon` here, but `start` in the other example. The task doesn't actually set an initial value on the clock until the body runs, does it? So if you use `start_soon` and then try to access the signal afterwards it may still be uninitialized. Of course the real problem is that the initialization doesn't happen in the `__init__` because `start_high` is an argument on the `start` method instead of an argument to the driver... Accepting that, we should recommend using `start` here.

ktbarrett (approved):
For the most part it seems like `start_soon` is used to replace `fork` (beside on `Clock.start`). The other exception is in test_queues, where `start` is used vs `start_soon`. Is this necessary for the test only, or are there usability issues using the queues with `start_soon`?

Do you think its safe to recommend that users replace `fork` with `start_soon` everywhere? And that where it doesn't work they can use `start` in a pinch, but they should prefer refactoring to not depend on the task starting immediately? And finally, is there any usability issues with using `fork` with `First` and `Combine` now that we have changed them to `start_soon`?

The change seems good. I'm not sure if there are other occurrences of the phrase "forked coroutine" elsewhere in the documentation or comments.

ktbarrett:
A final question is, do you think we should deprecate `fork` in 1.6, or wait until 1.7? I think it would be best to not deprecated fork for 1.6 as I dont know to what extend `start` and `start_soon` have been tested. Of course, any depreciation would be a follow on PR.

marlonjames:
`start` was necessary for the Queue tests in order to manipulate the state of the queues for `__repr__` coverage.

Remember that `start` is just `start_soon` + `await NullTrigger`.
`cocotb.start` is the closer analogue to `fork`, but it's just a convenience function for users. I think it makes sense to recommend `start_soon` generally.

I'm not aware of any issues using `fork` and `start_soon`. The ones I know about were fixed in #2504.

It's probably a good idea to either revert some tests to keep using `fork` or add some new ones.

I think we should leave `fork` as is for now, and give users a release to start trying `start`/`start_soon` before we deprecate.

marlonjames:
I reverted some of the tests in `test_scheduler` back to `cocotb.fork` to keep some test coverage.

And I'm almost certain there are still some uses of "fork" as a verb in the documentation.

ktbarrett:
> And I'm almost certain there are still some uses of "fork" as a verb in the documentation.

We can clean that up in a follow on. I think this can come in. I'm guessing we should update `fork a coroutine` with `start a task` and `forked coroutine` with `task`?

tests/test_cases/test_cocotb/test_concurrency_primitives.py
```diff
@@ -135,6 +135,19 @@ async def test_combine(dut):
     """ Test the Combine trigger. """
     # gh-852
 
+    async def do_something(delay):
+        await Timer(delay, "ns")
+
+    crs = [await cocotb.start(do_something(dly)) for dly in [10, 30, 20]]
```
cmarqu, line 141: I'd use `coro`/`coros` here

cmarqu (approved):
One minor thing...

PR #2715, opened 2021-09-18
Remove dependency on xunitparser

### LINKED ISSUE #2714: Find a replacement for xunitparser
The project has not been maintained since 2014 and is now causing [installation failures](https://github.com/cocotb/cocotb/runs/3635193264?check_suite_focus=true#step:25:106). WE should look for a new project to replace it with, or find someone to take over the project, make the fix, and upload a new version.

themperek:
It seems like we do not use xunitparser anywhere

Closes #2714

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -250,7 +250,7 @@ jobs:
       # Windows Testing
     - name: Install Python testing dependencies (Windows)
       if: startsWith(matrix.os, 'windows')
-      run: pip install coverage xunitparser pytest pytest-cov
+      run: pip install coverage pytest pytest-cov
     - name: Install cocotb build dependencies (Windows, mingw)
       if: startsWith(matrix.os, 'windows') && matrix.toolchain == 'mingw'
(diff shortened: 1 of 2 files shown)


themperek:
msvc fails. Any idea? Seems not to be related.

themperek:
@marlonjames Any idea on why msvc fails?

imphil:
>     C:\Program Files (x86)\Windows Kits\10\include\10.0.22000.0\um\winnt.h(253) : error RC2188: D:\a\cocotb\cocotb\build\temp.win-amd64-3.8\Release\cocotb\libs\libgpilog\RCa02644(15) : fatal error RC1116: RC terminating after preprocessor errors

Obscure.

themperek:
Maybe because setuptools version change in conda? https://anaconda.org/anaconda/setuptools/files

themperek:
Looks like Widows 11 SDK update https://bugs.python.org/issue45220

ktbarrett (approved):
I don't think a bug in the Windows 11 Preview SDK should get in the way of this coming in as it's breaking all the other tests.

Issue #1522, opened 2020-03-20
Consider changing base class of NonConstantObject


marlonjames:
I don't see why `NonConstantObject` should inherit from `NonHierarchyIndexableObject`. Wouldn't it make more sense for it to be another branch from `NonHierarchyObject`?

See https://cocotb.readthedocs.io/en/latest/library_reference.html#simulation-object-handles for class diagram.

eric-wieser:
It's worth thinking about the effect this would have on these lines:

https://github.com/cocotb/cocotb/blob/7c3f44df806184d1b3ccade7b8492bde826dd34e/cocotb/handle.py#L810-L821

Not to say that I've done that thinking yet.

PR #2720, opened 2021-09-24
Cleanup handle

ktbarrett:
Starting some of the cleanup mentioned in https://github.com/cocotb/cocotb/wiki/Handle-Cleanup-Proposal without significantly breaking the interface.

Things this breaks:
* importing `ConstantObject` or `NonConstantObject`: this could be mitigated by leaving them in place and just not using them.
* subclass/instance checks of `ConstantObject` or `NonConstantObject`: this could be mitigated by leaving them in place.
* `isinstance(obj, ConstantObject)` for constant object: this could be mitigated by leaving `ConstantObject` and overriding `__instancecheck__` to work on any `NonHierarchyObject` that `is_const`.
* subclass/instance checks involving many types/values and `IndexableNonHeirarchyObject`: no mitigitation, but I don't think this is a big break
* trying to instantiate base class handle types: we could just drop commit 6, but I doubt this is a big break.
* `type(logic_handle) is ModifiableObject`: no mitigation, but I doubt such strict checks are used, `isinstance(logic_handle, ModifiableObject)` still works.


tests/test_cases/test_array/test_array.py
```diff
@@ -389,26 +389,26 @@ async def test_direct_constant_indexing(dut):
...
-    _check_type(tlog, dut.const_rec.b[1], ConstantObject)
+    _check_type(tlog, dut.const_rec.b[1], ModifiableObject)
 
     _check_type(tlog, dut.const_cmplx, NonHierarchyIndexableObject)
     _check_type(tlog, dut.const_cmplx[1], HierarchyObject)
-    _check_type(tlog, dut.const_cmplx[1].a, ConstantObject)
+    _check_type(tlog, dut.const_cmplx[1].a, ModifiableObject)
     _check_type(tlog, dut.const_cmplx[1].b, NonHierarchyIndexableObject)
-    _check_type(tlog, dut.const_cmplx[1].b[1], ConstantObject)
+    _check_type(tlog, dut.const_cmplx[1].b[1], ModifiableObject)
```
ktbarrett, line 411: I think all of these should be `LogicObject`,

cocotb/handle.py
```diff
@@ -940,10 +1014,33 @@ def _set_value(self, value, call_sim):
...
+    def __getitem__(self, item):
+        warnings.warn(
+            "Attempting to index RealObject is deprecated",
+            DeprecationWarning, stacklevel=2)
+        return super().__getitem__(item)
+
+    def __setitem__(self, item):
+        warnings.warn(
+            "Attempting to set an index on a RealObject is deprecated",
+            DeprecationWarning, stacklevel=2)
```
ktbarrett, line 1035: Doesn't actually try to set here... same comment for all occurences.

ktbarrett:
The recursive discovery tests don't like me deprecating iteration over unit objects like reals and ints. I could condition the recursion to only objects which are iterable: `HierarchyObject` and `NonHierarchyIndexableObject`, or maybe we want to continue supporting iteration over clearly non-iterable objects just because the VPI/VHPI doesn't bomb out?

ktbarrett:
Seems like a lot of tests worked in GHDL with `ConstantObject`. You couldn't tell what the type was, but if you *know* what the type is you could access the value correctly. It is uncertain what would have happened if you accesses the value incorrectly. "Works by happenstance". removing `ConstantObject` will likely break some GHDL users.

Questa and Riviera are also failing because the tests were previously counting individual bits discovered by iterating over logic array or string types. I can add those back in to the recursion, but perhaps there is a better way to iterate where invalid iterations (like over reals and ints) is not allowed? Unfortunately, the test code assumes that you can iterate safely over all handles, and while that's true, it isn't sensible.

src/cocotb/handle.py
```diff
@@ -30,14 +30,17 @@
 # -*- coding: utf-8 -*-
```
cmarqu, line 31:
```suggestion
```
While we are here?

src/cocotb/handle.py
```diff
@@ -758,13 +699,13 @@ def value(self) -> BinaryValue:
         result._set_trusted_binstr(binstr)
         return result
 
-    def __int__(self):
-        return int(self.value)
-
 
 class RealObject(ModifiableObject):
     """Specific object handle for Real signals and variables."""
```
cmarqu, line 704:
```suggestion
    """Specific object handle for real-valued signals and variables."""
```
(while here)

tests/test_cases/test_discovery/test_vhdl_indexed_name.py
```diff
@@ -26,12 +26,20 @@
 import cocotb
 
 
-@cocotb.test()
+@cocotb.test(skip=True)
```
cmarqu, line 29: Skipped for a reason?

ktbarrett: It doesn't actually test anything. In fact I'll do some more refactoring

ktbarrett:
@cocotb/maintainers Review ping.

marlonjames:
This closes #1522

src/cocotb/handle.py
```diff
@@ -425,6 +410,10 @@ def _set_value(self, value, call_sim):
             )
         )
 
+    @lru_cache(maxsize=None)
+    def __len__(self) -> int:
+        return self._handle.get_num_elems()
```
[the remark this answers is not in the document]
ktbarrett: Uh. Yeah. Not sure what I was thinking here.

examples/matrix_multiplier/tests/Makefile
```diff
@@ -63,8 +63,18 @@ export RANDOM_SEED := 123456789
 TOPLEVEL    := matrix_multiplier
 MODULE      := test_matrix_multiplier
 
+ifeq ($(SIM),ghdl)
+
+all:
+	@echo "Skipping since GHDL doesn't support constants effectively"
```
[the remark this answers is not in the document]
ktbarrett: #2722 exists, but doesn't detail much. I'll add some and add the github issue here as well.

ktbarrett:
I had to update `access_integer` and `access_boolean` in `test_discovery`. Getting the length and indexing those handles are not supported anymore, so I just removed those parts of the test.

ktbarrett:
Added newsfragments for the user-pertinent removals and changes. @cmarqu could you opine? Then I'll merge this.

docs/source/newsfragments/2720.change.1.rst
```diff
@@ -0,0 +1 @@
+Testing equality of :class:`simulator value object <cocotb.handle.NonHierarchyObject>` subtypes no longer does a value equality, but a object equality. All simulator objects can now be used in dictionaries.
```
marlonjames, line 1:
```suggestion
Testing equality of :class:`simulator value object <cocotb.handle.NonHierarchyObject>` subtypes no longer does a value equality, but an object equality. All simulator objects can now be used in dictionaries.
```

docs/source/newsfragments/2730.change.3.rst
marlonjames, line 1: Should be `2720.change.3.rst`

docs/source/newsfragments/2720.change.rst
```diff
@@ -0,0 +1 @@
+``cocotb.handle.ConstantObject`` was removed. Constant value objects are now returned as their corresponding :class:`simulator value object <cocotb.handle.NonHierarchyObject>` subtype (e.g. integers in VHDL and Verilog are now :class:`~cocotb.handle.IntegerObject`\ s). You can check if the value object is constant using the :meth:`~cocotb.handle.NonHierarchyObject.is_const` method.
```
marlonjames, line 1: `NonHierarchyObject.is_const()` doesn't show up in the docs and this link doesn't work.

ktbarrett: `WARNING: Failed to get a method signature for cocotb.handle.NonHierarchyObject.is_const: <cocotb.utils.lazy_property object at 0x7f413f993880> is not a callable object`

PR #2721, opened 2021-09-24
Better support test reproducibility

### LINKED ISSUE #2672: Potential rerun/reproducibility problem when using TESTCASE with multi-test MODULE
Opening PR per @ktbarrett in discussion here: https://gitter.im/cocotb/Lobby

In short, cocotb sets random.seed at the start of running a MODULE in init.py but does not set the seed again at the start of each @cocotb.test().  If the MODULE had multiple tests in it and a user sets TESTCASE to rerun any test but the first one in MODULE, then it's possible the random behavior for the rerun test will be different than the original failure.

A few thoughts:   Resetting the seed back to RANDOM_SEED (or the randomly generated seed) before each test would not be an attractive solution because then the random behavior for each test could be identical for all MODULE tests.   Ideally, the "testcase seed" would be set to a different value for each test and derived from/based on the RANDOM_SEED (or randomly generated seed).   One way to get such a seed could be to offset the main seed based on some algorithm based on the test currently running or in other words:   Testcase Seed = Main Seed + **f(testcase)**        If this were true then:

- Each test would get a unique seed hopefully producing different random behavior than the other tests
- When rerunning a failing test in a multi-test MODULE, the user would only need to set RANDOM_SEED to the same Main Seed which was used (likely randomly generated) for the original failure.

I am new to cocotb so I could be missing details but it appears MODULE test names are essentially unique in that cocotb seems to combine all @cocotb.test() names in the MODULE which are the same into one test, no matter where they appear in the file.  If so, then one possibility for the **f(testcase)** offset would be to hash the testcase name to produce a unique number.  Granted, it's possible the same number might result from different test names but would be hopefully uncommon.

ktbarrett:
Closes #2672. Pinging @walt3k.

Tests are now reproducible if:
* the test name and module does not change
* the test body does not change
* the test is "pure", meaning it's not affected by its environment (the state of the DUT, the state of the Python environment)
* the same `RANDOM_SEED` is used


cocotb/regression.py
```diff
@@ -308,8 +310,17 @@ def _init_test(self, test: Test) -> Optional[RunningTask]:
...
             return None
 
-        test = test_init_outcome.get()
-        return test
+        running_test = test_init_outcome.get()
+
+        # seed random number generator based on test module, name, and RANDOM_SEED
+        hasher = hashlib.sha1()
+        hasher.update(test.__qualname__.encode())
+        hasher.update(test.__module__.encode())
```
walt3k, line 318: As I commented, it's not unusual where I work for people to collect failing tests from many different test files (MODULE files) into one big regression MODULE file.  Users will often call this file "The Kitchen Sink" or something creative like that. :-)  If the MODULE name is used here then users would not be able to do that  i.e. if they had a hundred failing tests they could potentially be forced to have a hundred separate MODULE files as well... rather than one.  Not a deal breaker but I thought I should mention it.

ktbarrett: `__module__` will be the name of the module they were defined in, not the module they were discovered in. So importing tests from other modules into a "kitchen sink" module will continue to work.

ktbarrett:
Needs newsfragment.

cocotb/regression.py
```diff
@@ -308,8 +310,17 @@ def _init_test(self, test: Test) -> Optional[RunningTask]:
...
+        seed = cocotb.RANDOM_SEED + int(hasher.hexdigest(), 16)
+        print(test.__qualname__, seed, flush=True)
```
cmarqu, line 320:
```suggestion
        print(f"Test {test.__qualname__} runs with seed {seed}", flush=True)
```

tests/test_cases/test_seed/test_seed.py
```diff
@@ -0,0 +1,25 @@
...
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+import random
+
+import cocotb
+
+
+@cocotb.test()
+async def test_first(_):
+    # move generator to test it doesn't affect the next test
```
cmarqu, line 11:
```suggestion
    # move generator to test that it doesn't affect the next test
```

cmarqu (approved):

cocotb/regression.py
ktbarrett:
```suggestion
```

marlonjames (approved):
Looks good for now.
After we improve GPI extension support, we should provide a central mechanism for updating the PRNG seed and being notified of the updates.

PR #2728, opened 2021-09-30
Fix release note formatting

ktbarrett:
Fix an odd formatting issue. Will add to 1.6 branch as well.


imphil:
I'll copy over all release notes from 1.6 to master once we're publishing the release. That worked well in the past, feel free to tweak the release notes on the stable branch till then.

PR #2737, opened 2021-10-14
Add doc for len()/__len__ on handles

cmarqu:


imphil:
Thanks! Merging as this PR addresses the problem raised. Having additional documentation for _range can be done in a follow-up if you want to do so.

PR #2740, opened 2021-10-16
Lock the setup-msys2 action to v2.4.2

imphil:
setup-msys2 v2.5.0 broke Python for yet unclear reasons, avoid it until
that's resolved.

See also cocotb#2739

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -351,7 +351,7 @@ jobs:
       run:
         shell: msys2 {0}
     steps:
-    - uses: msys2/setup-msys2@v2
+    - uses: msys2/setup-msys2@v2.4.2
       with:
         msystem: MINGW64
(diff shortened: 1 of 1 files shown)


Readon:
I have met the same problem when running cocotb on msys2 on my pc.
is there any suggestion on how to workaround?

ktbarrett:
@Readon Please add a comment [here](https://github.com/msys2/setup-msys2/issues/167) or [here](https://github .com/msys2-contrib/cpython-mingw/issues/51) so upstream can assess urgency.You could also try removing [this](https://github.com/cocotb/cocotb/blob/0346c035f0e1fac432c5a35ea38b95787e7d7223/cocotb/share/makefiles/Makefile.inc#L74) line from the makefiles to see if it works without that? Total shot in the dark.

Readon:
Thanks, it works.

Issue #2749, opened 2021-10-20
VHPI: `ReadOnly` callback does not allow use of `Timer` triggers


marlonjames:
When the VHPI callback reasons for `ReadWrite` and `ReadOnly` where changed (see #1061),
a different problem was introduced.

Using `vhpiCbRepEndOfTimeStep` for the `ReadOnly` callback reason no longer allows use of `Timer`.

The VHDL LRM explicitly forbids registering a callback using `vhpiCbAfterDelay`:

> **21.3.6.8 `vhpiCbEndOfTimeStep` and `vhpiCbRepEndOfTimeStep`**
> It is an error if a `vhpiCbEndOfTimeStep` or `vhpiCbRepEndOfTimeStep` callback causes activity
on a driver or a signal (see 14.7.3.1) or registers or enables a `vhpiCbAfterDelay`,
`vhpiCbRepAfterDelay`, `vhpiCbTimeout`, or `vhpiCbRepTimeOut` callback.

The purpose of `ReadOnly` is to allow seeing the "final" values of simulation objects after all processes have run, where signals cannot be changed, but `Timer` is useful for taking future action based on these values.

VHPI provides a callback reason for this: `vhpiCbRepStartOfPostponed`.

> **11.3 Process statement**
> NOTE 2— [...] When a postponed process is resumed, it does not execute until a simulation cycle occurs
in which the next simulation cycle is not a delta cycle. In this way, a postponed process accesses the values of signals
that are the “final” values at the current simulated time.

These callbacks are still able to affect future scheduling with `Timer` triggers, as `Tn` (the time of the next simulation cycle) is calculated after all registered callbacks are called.
This is a better match for `ReadOnly`.

Simulator support is an open question.

ktbarrett:
Siemens has agreed to allow the illegal behavior in Questa 2022.1, so the Questa VHPI should fully support cocotb at that time.

marlonjames:
I know we usually prefer not to have simulator-specific workarounds, but this seems like a good candidate for it.

PR #2750, opened 2021-10-20
VHPI: Change ReadOnly callback reason

marlonjames:
Change VHPI `ReadOnly` callback reason to address #2749.


marlonjames:
Questa doesn't support `vhpiCbRepStartOfPostponed` yet.
Riviera also has some issue.

imphil:
Conclusion from today's maintainer meeting: Let's `#ifdef` this change for Questa only for the moment. I opened https://github.com/cocotb/cocotb/issues/2801 to keep track of the other VHPI simulators.

imphil (commented):
LGTM with a Questa ifdef.

imphil (approved):

marlonjames:
Questa 2021.3 doesn't support `vhpiCbRepStartOfPostponed`.

imphil:
All simulators, including Questa, are currently supporting vhpiCbRepEndOfTimeStep (only). I guess that's good enough for now, no matter what the standard ultimately says. So let's leave things as they are and revisit if that's not good enough any more.

PR #2759, opened 2021-11-01
[ci] Add ecosystem compatibility tests

imphil:
Add a GitHub Actions workflow that tests released versions of code which
depends on cocotb against current master versions of cocotb. Runs once a
day and can be run on-demand by triggering it through the web UI.

For now start with testing the latest released version of cocotb-coverage.

### DIFF
--- a/.github/workflows/ecosystem-compat.yml
+++ b/.github/workflows/ecosystem-compat.yml
@@ -1,0 +1,59 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Tests to ensure that projects depending on cocotb continue to work with the
+# latest development version of cocotb.
+#
+# Generally, we test the development version of cocotb against supported,
+# released versions of the other projects. (It is expected that the projects
+# themselves test their in-development code against the released version of
+# cocotb.)
+
+name: Ecosystem compatibility tests
+
+on:
+  # Run daily at midnight (UTC).
+  schedule:
+    - cron: '0 0 * * *'
+  # Allow triggering a CI run from the web UI.
+  workflow_dispatch:
+
+jobs:
+  cocotb-coverage:
+    name: Test cocotb-coverage 1.1
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.10"
+
+      - name: Install Icarus Verilog
+        run: sudo apt install -y --no-install-recommends iverilog
+
+      - name: Checkout cocotb repository
+        uses: actions/checkout@v2
+        with:
+          path: cocotb
+
+      - name: Install the development version of cocotb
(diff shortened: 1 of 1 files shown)


themperek:
Should we rather move cocotb-coverage under cocotb organization and add tests there?

imphil:
> Should we rather move cocotb-coverage under cocotb organization

The testing is completely independent of where the code lives. I don't want to interfere with the decision of the code author where their code should live -- it's really up to them.

> and add tests there?

There are already automated tests in cocotb-coverage which run against the released version of cocotb. What seems like duplication is intended. 

* Tests in cocotb-coverage are meant to test if the development version of cocotb-coverage works with the currently released version of cocotb. That's what the authors of cocotb-coverage need to take care of.
* The tests in here are for cocotb developers: we want to ensure that changes to cocotb don't (unintentionally) break packages in our ecosystem. That's why we run the development version of cocotb against a released version of e.g. cocotb-coverage.

umarcor:
For reference, in GHDL we have this "extended-tests" repository. Within an organisation it is acceptable to have workflows cross-triggered. See for instance: https://github.com/ghdl/ghdl/blob/f43dcd402ed7783cda796675c6147ec0e68d2714/.github/workflows/Test.yml#L625-L631

ktbarrett:
I don't appreciate the extra scope. This is will also prevent making breaking changes when we are about to start doing that. Perhaps instead we should provide a setup-cocotb github action in the org for downstream's use.

imphil:
This GH action isn't blocking any merge or change (it doesn't even run then). It's only for us to know what's going on in associated projects. Before, we had to manually check before a release if e.g. cocotb-bus or cocotb-coverage are compatible. Now we know by just looking at the CI results.

ktbarrett:
So what's the process if something gets broken? Do we revert our changes or tell the maintainer in the downstream project they need to update their stuff? If the former is the case, running weekly is kind of a disservice compared to a check on PR: merge, find out Sunday night something's broke, revert. If the later is the case, shouldn't these tests be set up in their repos so they can better track it? If we are checking on PRs, it would have to be done in a way that did not prevent development of conscious breaking changes, we aren't guaranteeing that any downstream project will be compatible with 2.0, including cocotb-bus or cocotb-coverage.

imphil:
There's no requirement or policy to change anything in cocotb when a downstream project breaks. This CI here only to make sure we know and are able to separate between breakages we intended, and breakages that weren't intended. If we feel like it, we can tell the other maintainer. If not, then not. 

This kind of test is only to increase our field of vision and awareness how changes to cocotb affect others. No policy we have today changes.

> shouldn't these tests be set up in their repos so they can better track it?

Other projects can certainly test against cocotb master as well. But this CI here is for us: we want to have a quick-and-easy way to see if e.g. we unintentionally introduce backwards-incompatible changes when we didn't intend to.

themperek (approved):

ktbarrett (approved):

ktbarrett:
@imphil Is this good to merge?

PR #2760, opened 2021-11-03
Remove CoroutineComplete.

marlonjames:
CoroutineComplete isn't needed. The scheduler can examine if the task finished with `_finished` property.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -63,18 +63,6 @@ def public(f):
 public(public)  # Emulate decorating ourself
 
 
-@public
-class CoroutineComplete(Exception):
-    """To ensure that a coroutine has completed before we fire any triggers
-    that are blocked waiting for the coroutine to end, we create a sub-class
-    exception that the scheduler catches and the callbacks are attached
-    here.
-    """
-
-    def __init__(self, text=""):
-        Exception.__init__(self, text)
-
-
 class RunningTask:
     """Per instance wrapper around a running generator.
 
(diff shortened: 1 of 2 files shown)


cocotb/decorators.py
```diff
@@ -190,24 +178,18 @@ def _advance(self, outcome):
...
         try:
             self._started = True
             return outcome.send(self._coro)
         except ReturnValue as e:
             self._outcome = outcomes.Value(e.retval)
-            raise CoroutineComplete()
         except StopIteration as e:
             self._outcome = outcomes.Value(e.value)
-            raise CoroutineComplete()
         except BaseException as e:
```
marlonjames, line 191: I don't understand why this is `BaseException` and not `Exception`.

cmarqu: @eric-wieser had changed this from `Exception` in https://github.com/cocotb/cocotb/commit/b514cad8b083e07acfd68340f53c5d57f2352a08, maybe he remembers why?

eric-wieser: The idea was to catch Ctrl+C so that it is propagated to the calling coroutine.

cocotb/scheduler.py
```diff
@@ -876,31 +876,25 @@ def _schedule(self, coroutine, trigger=None):
...
 
             # Don't handle the result if we're shutting down
             if self._terminate:
                 return
 
-            if not coro_completed:
+            if not coroutine._finished:
+                if _debug:
+                    self.log.debug("Coroutine %s yielded %s (mode %d)" %
+                                   (coroutine._coro.__qualname__, str(result), self._mode))
```
ktbarrett, line 897: Are we sure that `str(result)` is always going to result in something reasonable?

marlonjames: Doubtful. The debug logging needs some improvements.

marlonjames: We can leave it for another PR?

ktbarrett: We can. I was also thinking it could be moved to after the `_trigger_from_any` call, so you know what will be printed will be a trigger. Either way works for me, you have the checkmark.

Issue #2739, opened 2021-10-16
mingw tests fail in CI


imphil:
Tests running in mingw-env in CI fail now, most likely not due to a change in our code, but due to an environment problem in the CI testing.

Good: https://github.com/cocotb/cocotb/runs/3753710579?check_suite_focus=true (30 Sep 2021 08:32)
Bad: https://github.com/cocotb/cocotb/runs/3898922153?check_suite_focus=true (14 Oct 2021 19:40)

Things I know changed:
* msys2/setup-msys2@v2 updated from 549e904fc53cb840ec72fec29f3e8fc00f579b5b (v2.4.2) to 6e532aa72709f068e400cc0677a3f49162891b0e (v2.5.0)
* The Windows base image changed from 20210920.1 to 20211011.0

Error:
```
Python path configuration:
  PYTHONHOME = 'D:/a/_temp/msys/msys64/mingw64'
  PYTHONPATH = (not set)
  program name = 'D:/a/_temp/msys64/mingw64/bin/python.exe'
  isolated = 0
  environment = 1
  user site = 1
  import site = 1
  sys._base_executable = 'D:/a/_temp/msys64/mingw64/bin/python.exe'
  sys.base_prefix = 'D:/a/_temp/msys/msys64/mingw64'
  sys.base_exec_prefix = 'D:/a/_temp/msys/msys64/mingw64'
  sys.platlibdir = 'lib'
  sys.executable = 'D:/a/_temp/msys64/mingw64/bin/python.exe'
  sys.prefix = 'D:/a/_temp/msys/msys64/mingw64'
  sys.exec_prefix = 'D:/a/_temp/msys/msys64/mingw64'
  sys.path = [
    'D:/a/_temp/msys/msys64/mingw64/lib/python39.zip',
    'D:/a/_temp/msys/msys64/mingw64/lib/python3.9',
    'D:/a/_temp/msys/msys64/mingw64/lib/python3.9/lib-dynload',
    '',
  ]
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'
```

imphil:
Looks like it's down to the setup-msys2 version: 2.4.2 works, 2.5.0 doesn't. Diff between the log output of the two versions: 
[diff.txt](https://github.com/cocotb/cocotb/files/7358316/diff.txt)

[...]

PR #2763, opened 2021-11-04
CI: Use setup-msys2 options 'location' and 'pacboy'

umarcor:
Reliance on the 'prefix' broke usage of cocotb with setup-msys2 v2.5.0, because the default installation location was changed.
As a workaround, setup-msys2 was pinned to v2.4.2 in this repo.
In release v2.6.0 of setup-msys2, two options were added:
* location: allows to specify the location of the MSYS2 installation.
* pacboy: allows simplifying the syntax for installing MINGW packages.

Having the Action pinned, prevents usage of pacboy. However, using option 'location' allows unpinning the Action, so that 'pacboy' can be used.

This PR uses both 'location' and 'pacboy'.

Refs:

- #2739
- #2740
- msys2/setup-msys2#167
- msys2-contrib/cpython-mingw#51

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -338,18 +338,19 @@ jobs:
       run:
         shell: msys2 {0}
     steps:
-    - uses: msys2/setup-msys2@v2.4.2
+    - uses: msys2/setup-msys2@v2
       with:
+        # Workaround for https://github.com/cocotb/cocotb/issues/2739
+        location: D:/a/_temp/msys
         msystem: MINGW64
         update: true
-        install: |
-            git
-            make
-            mingw-w64-x86_64-gcc
-            mingw-w64-x86_64-python-pip
-            mingw-w64-x86_64-python-pytest
-            mingw-w64-x86_64-python-wheel
-            mingw-w64-x86_64-${{ matrix.pkg }}
+        install: git make
+        pacboy: >-
+          gcc:p
+          python-pip:p
+          python-pytest:p
+          python-wheel:p
+          ${{ matrix.pkg }}:p
     - uses: actions/checkout@v2
     - name: install cocotb
       run: pip install --no-build-isolation .


ktbarrett:
Not sure I understand why "location" is necessary. This seems to be an issue with setup-msys2 and not a user issue?

imphil:
It's most likely the workaround for https://github.com/msys2/setup-msys2/issues/167? If so, @umarcor please add a comment to this line in the config file.

umarcor:
> Not sure I understand why "location" is necessary. This seems to be an issue with setup-msys2 and not a user issue?

- cocotb relies on the "hardcoded" prefix in the Python installation.
- Until v2.4.2, setup-msys2 used one location.
  - setup-msys2 is used by MSYS2 CI for building packages.
  - Python has the hardcoded prefix used when the package is built in MSYS2's CI.
- In v2.5.0, the default location used by setup-msys2 was changed.
  - However, the Python packages were not rebuilt yet.
  - Therefore, packages do still have the "old" prefix, but using a new version of setup-msys2 will install then in the "new" location.
  - That crashes cocotb's installation, and it was pinned to v2.4.2 to avoid it.
- In v2.6.0, feature "pacboy" was introduced, which simplifies installing packages for multiple environments (as it read the prefix from an envvar instead of requiring it explicitly).
  - I want to enhance the workflow using pacboy, but I cannot because the Action is pinned.
  - Unpinning the Action is possible if option "location" is used for preserving the "old" prefix.
  
Using the configuration in this PR, everything will be green until Python packages in MSYS2 are rebuilt and use the "new" prefix. Then, we can remove `location` here and close #2739.

Alternatively, the issue might be silently fixed, if msys2-contrib/cpython-mingw/issues/51 is solved. I'm subscribed to that issue.

> @umarcor please add a comment to this line in the config file.

Done.

ktbarrett:
>cocotb relies on the "hardcoded" prefix in the Python installation.

Doesn't this mean that every user is also broken since `'prefix'` config var won't be correct? After reading the linked issue, I'm fairly confident that msys2's Python installation is misconfigured.

umarcor:
> Doesn't this mean that every user is also broken since `'prefix'` config var won't be correct? After reading the linked issue, I'm fairly confident that msys2's Python installation is misconfigured.

It is correct that this is a misconfiguration in MSYS2's Python (see https://github.com/msys2-contrib/cpython-mingw/issues/51). However, not every other user is also broken. For instance, the following CI's are using latest setup-msys2 and Python without problems:

- https://github.com/ghdl/ghdl/actions/runs/1418716817
- https://github.com/VUnit/vunit/actions/runs/1423869694

ktbarrett:
Looks like users are also broken by this. https://github.com/cocotb/cocotb/pull/2740#issuecomment-962566628
The pacboy changes look good, but the location changes are going to need to be fixed upstream, so I think this should be put on hold until we have a fix. We should be testing as users will have their environment configured.

marlonjames:
#2964 removed mingw tests from CI

PR #2765, opened 2021-11-04
CI: Test on multiple MSYS2 environments 

umarcor:
This PR is based on #2763.

Thanks to `pacboy`, writing a matrix for dealing with testing on multiple MSYS2 environments is easier now. GHDL, Icarus Verilog and Verilator are available on MINGW64, MINGW32 and UCRT64. This PR enhances the matrix to test all three tools on all three environments.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -323,7 +323,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
+        sys:
+          - { icon: '⬛', msys: MINGW32 }
+          - { icon: '🟦', msys: MINGW64 }
+          - { icon: '🟨', msys: UCRT64  } # Experimental!
+        tool:
           - sim: verilator
             lang: verilog
             pkg: verilator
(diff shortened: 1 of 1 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -336,7 +336,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
+        sys:
+          - { icon: '⬛', msys: MINGW32 }
+          - { icon: '🟦', msys: MINGW64 }
+          - { icon: '🟨', msys: UCRT64  } # Experimental!
```
imphil, line 342: Is this something we want support? I think this probably needs wider discussion in an issue first. We have been trying to reduce the amount of things we support, not extend it.

umarcor: I think that UCRT64 might replace MINGW64 in the near future, since it makes it easier to use DLLs from MSVC in MSYS2 or the other way round. I believe that is particularly interesting for cocotb, since it uses Python and it builds shared libraries. Hence, I would recommend to test both MINGW64 and UCRT64 for now.

With regard to 32 bits, that's up to you. I know some people use GHDL's 32 bit version because they want the `mcode` backend. However, that should be supported on 64 bits sometime in the future. It is acceptable to deprecate 32 bits.

umarcor: See https://www.msys2.org/docs/environments/

ktbarrett:
My concern is more about adding tests for verilator and GHDL. These tests really only existed to show that cocotb could be compiled and built successfully in these environments. The PR checks already take enough time. Perhaps the Icarus tests should be per PR and the remainder should be weekly? We are introducing weekly CI runs in the other PR.

umarcor:
> My concern is more about adding tests for verilator and GHDL. These tests really only existed to show that cocotb could be compiled and built successfully in these environments. The PR checks already take enough time.

Fair enough. 

> Perhaps the Icarus tests should be per PR and the remainder should be weekly? We are introducing weekly CI runs in the other PR.

Do you mean #2759? That's daily, not weekly.
Anyway, I'm good with waiting until #2759 and #2763 are solved/merged, before deciding the strategy here.

ktbarrett:
@umarcor #2759 has been merged.

imphil:
> @ktbarrett shall I move all the MSYS2 tests to workflow ecosystem-compat.yml?

No. The ecosystem-compat tests are there to test other projects against cocotb. The test you're thinking about is testing cocotb itself. If we want "tier 2" tests for cocotb itself that run less frequently we should have another workflow for that. Feel free to create it, however!

PR #2768, opened 2021-11-07
Fix issue with msys2 environments

ktbarrett:
Removing line that is breaking in msys2 Python environments. The [recommendation that sparked this change](https://docs.python.org/3.9/using/windows.html#finding-modules) may not be necessary. Pinging @umarcor.

### DIFF
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -70,9 +70,6 @@ ifeq ($(filter $(PYTHON_ARCH),64bit 32bit),)
     $(error Unknown Python architecture: $(PYTHON_ARCH))
 endif
 
-# Set PYTHONHOME to properly populate sys.path in embedded python interpreter
-export PYTHONHOME := $(shell $(PYTHON_BIN) -c 'from sysconfig import get_config_var; print(get_config_var("prefix"))')
-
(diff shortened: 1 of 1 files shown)


ktbarrett:
Coverage changed because of some issue with codecov. Not all coverage files are in the build ([here](https://codecov.io/gh/cocotb/cocotb/commit/24d4e8185926f8dbb09bc994d31fd7861a21b0bf/build)), but looking at the Azure jobs they all successfully uploaded the coverage files. We might need to bump [this number](https://github.com/cocotb/cocotb/blob/7cc4e12b22213b031fa6e16565c1b5b60d19fd22/.codecov.yml#L20) since the inclusion of additional Azure tests.

themperek:
With this change, Questa (2021.01) on Widows fails with:
```
Fatal Python error: initfsencoding: unable to load the file system codec
ModuleNotFoundError: No module named 'encodings'
```
🤷🏻‍♂️

ktbarrett:
@themperek Which environment is that? System Python or Anaconda? Does it only fail with Questa?

themperek:
> @themperek Which environment is that? System Python or Anaconda?

Anaconda (3.6.4). 

> Does it only fail with Questa?

I maybe had the same with Aldec (not sure). Do not have a license anymore.

PR #2775, opened 2021-11-13
Implement Force/Release for FLI

ktbarrett:


cocotb/share/lib/fli/FliObjHdl.cpp
```diff
@@ -249,14 +249,41 @@ int FliEnumObjHdl::set_signal_value(const int32_t value,
...
     if (m_is_var) {
-        mti_SetVarValue(get_handle<mtiVariableIdT>(),
-                        static_cast<mtiLongT>(value));
+        switch (action) {
+            case GPI_DEPOSIT:
+                mti_SetVarValue(get_handle<mtiVariableIdT>(),
+                                static_cast<mtiLongT>(value));
+                return 0;
+            case GPI_FORCE:
+                LOG_ERROR("Cannot force a VHDL variable");
```
cmarqu, line 258: Maybe reword these messages to
```suggestion
                LOG_ERROR("Forcing VHDL variables is not supported by FLI");
```
to prevent users reading this as a message about a specific variable (might happen if they are not native speakers).

tests/test_cases/test_force_release/test_force_release.py
```diff
@@ -8,8 +8,6 @@
 # Questa FLI GPI implementation does not support Force/Release (gh-1855)
```
cmarqu, line 8:
```suggestion
```

cocotb/share/lib/fli/FliObjHdl.cpp
ktbarrett: Perhaps instead:
```suggestion
                LOG_ERROR("Cannot force VHDL variables");
```

VHDL variables can only be set, they do not support inertial writes or force/release in the language. IDK how the VHPI handles these cases, perhaps instead we should treat force as a regular variable set so it's consistent?

cocotb/share/lib/fli/FliObjHdl.cpp
```diff
@@ -357,30 +406,93 @@ int FliLogicObjHdl::set_signal_value(const int32_t value,
...
         } else {
-            mti_SetSignalValue(get_handle<mtiSignalIdT>(),
-                               (mtiLongT)m_mti_buff);
+            switch (action) {
+                case GPI_DEPOSIT:
+                    mti_SetSignalValue(get_handle<mtiSignalIdT>(),
+                                       (mtiLongT)m_mti_buff);
+                    return 0;
+                case GPI_FORCE: {
+                    std::string value_str = "2#";
```
benbr8, line 431: You could also use `10#<val>` and `-10#<val>`. Questa won't make a difference, since it will create an int from it anyway. Not sure which is better, just a note :).

benbr8 (approved):

cocotb/share/lib/fli/FliObjHdl.cpp
marlonjames: VHPI allows forcing variables. Section 22.5.2 of the LRM has details.

ktbarrett: So there is no language-level support for forcing variables, but you can do it with the VHPI? Okay. I guess the first suggestion is more correct.

cocotb/share/lib/fli/FliObjHdl.cpp
ktbarrett: Does the force command accepts negative numbers for setting std_logic_vectors?

cocotb/share/lib/fli/FliObjHdl.cpp
```diff
@@ -249,14 +249,42 @@ int FliEnumObjHdl::set_signal_value(const int32_t value,
...
     if (m_is_var) {
-        mti_SetVarValue(get_handle<mtiVariableIdT>(),
-                        static_cast<mtiLongT>(value));
+        switch (action) {
+            case GPI_DEPOSIT:
+                mti_SetVarValue(get_handle<mtiVariableIdT>(),
+                                static_cast<mtiLongT>(value));
+                return 0;
+            case GPI_FORCE:
+                LOG_ERROR("Forcing VHDL variables is not suported by the FLI");
```
cmarqu, line 258:
```suggestion
                LOG_ERROR("Forcing VHDL variables is not supported by the FLI");
```
(and in the rest of the file :) ).

cmarqu:
This feature is worth a newsfragment.

documentation/source/newsfragments/2775.feature.rst
```diff
@@ -0,0 +1 @@
+:class:`~cocotb.handle.Force`, :class:`~cocotb.handle.Freeze`, and :class:`~cocotb.handle.Release` are now supported when using the FLI.
```
cmarqu, line 1:
```suggestion
:class:`~cocotb.handle.Force`, :class:`~cocotb.handle.Freeze`, and :class:`~cocotb.handle.Release` are now supported when using the FLI, Questa's traditional method to access VHDL.
```

cmarqu (approved):

PR #2777, opened 2021-11-13
Support log messages that exceed the size of the static buffer

ktbarrett:
Pinging @walt3k.

Using a `std::vector` as a dynamic buffer to write log messages to. Initial allocation of 512, but can dynamically be resized above the previously fixed limit. Does not shrink to fit because it is likely that another log of the same size will be seen again.


ktbarrett:
Fun... clang-format-11 and clang-format-10 do different things here.

cmarqu:
The value 512 was likely considered big enough to not be limiting even in extreme cases (we now know this not to be true).
Now that we can resize, would it make sense to start with a smaller size?

ktbarrett:
512 is still pretty small and it's only allocated once, perhaps we should increase it instead?

cocotb/share/lib/gpi_log/gpi_logging.cpp
```diff
@@ -125,10 +119,23 @@ extern "C" void gpi_native_logger_vlog(const char *name, int level,
...
-
-    if (n < 0 || n >= LOG_SIZE) {
-        fprintf(stderr, "Log message construction failed\n");
+    va_list argp_copy;
+    va_copy(argp_copy, argp);
+
+    static std::vector<char> log_buff(512);
+
+    log_buff.clear();
+    size_t n = vsnprintf(log_buff.data(), log_buff.capacity(), msg, argp);
```
marlonjames, line 128:
```suggestion
    int n = vsnprintf(log_buff.data(), log_buff.capacity(), msg, argp);
```

cocotb/share/lib/gpi_log/gpi_logging.cpp
```diff
@@ -125,10 +119,23 @@ extern "C" void gpi_native_logger_vlog(const char *name, int level,
...
+    if (n < 0) {
+        fprintf(stderr, "Log message construction failed: (error code) %d\n",
+                n);
+    } else if (n >= log_buff.capacity()) {
```
marlonjames, line 132: Will probably need a cast here for signed / unsigned comparison.

cocotb/share/lib/gpi_log/gpi_logging.cpp
ktbarrett: Thanks C committee for the great interface... size out should be a size_t and error code should be a separate output.

marlonjames:
Can we add a test to get coverage on this? What causes it, a super deep hierarchy? @walt3k

walt3k:
Well, we are adding LOG_DEBUG messages to be used to help people debug cases where they've made incorrect signal specifications, indexes, or record element specifications.  Our simulator flattens complex types for performance so we have apis which allow us to print complex types, among other things.  We hit this trying to print out a relatively small VHDL record, around 750 characters.  I think printing any diagnostic string would be fine.  Thx all for fixing this so fast!

ktbarrett:
I'm not sure if I have ever turned on LOG_DEBUG, but there are logs that dump signal names and we absolutely have names that exceed 512 characters due to deep hierarchy.

ktbarrett:
py_gpi_logging needs to e updated as well.

ktbarrett:
Seems like GHDL doesn't like long names. It just crashes hard when trying to compile the module. https://github.com/cocotb/cocotb/runs/4327782594?check_suite_focus=true#step:25:3945

ktbarrett:
[This run](https://github.com/cocotb/cocotb/actions/runs/1505097521) shows the test fails.

imphil (approved):
Thanks for this improvement @ktbarrett. 

I'm unconvinced that we actually need (the performance gains from) reusing a buffer. How about something simple like this? This way we also don't have to worry about multiple calls into the logger from various threads. 

```c++
extern "C" void gpi_native_logger_vlog(const char *name, int level,
                                       const char *pathname,
                                       const char *funcname, long lineno,
                                       const char *msg, va_list argp) {
    if (level < current_native_logger_level) {
        return;
    }

    // Determine required buffer size.
    va_list argp_copy;
    va_copy(argp_copy, argp);
    int msg_size = vsnprintf(nullptr, 0, msg, argp_copy);
    if (msg_size < 0) {
        fprintf(stderr, "Failed to determine buffer size for log message: %d\n",
                msg_size);
        va_end(argp_copy);
        return;
    }

    // Expand `msg` into `log_buff`.
    std::string log_buff;
    if (msg_size > 0) {
        log_buff.resize(msg_size);
        vsnprintf(&log_buff[0], msg_size+1, msg, argp);
    }

    // Print log message to STDOUT.
    fprintf(stdout, "     -.--ns ");
    fprintf(stdout, "%-9s", log_level(level));
    fprintf(stdout, "%-35s", name);

    size_t pathlen = strlen(pathname);
    if (pathlen > 20) {
        fprintf(stdout, "..%18s:", (pathname + (pathlen - 18)));
    } else {
        fprintf(stdout, "%20s:", pathname);
    }

    fprintf(stdout, "%-4ld", lineno);
    fprintf(stdout, " in %-31s ", funcname);
    fprintf(stdout, "%s", log_buff.data());
    fprintf(stdout, "\n");
    fflush(stdout);
}
```

This is just a suggestion: I'm leaving the decision up to you. Your code looks good to me and solves the problem.

cocotb/share/lib/py_gpi_log/py_gpi_logging.cpp
```diff
@@ -71,15 +72,26 @@ static void py_gpi_log_handler(void *, const char *name, int level,
...
-        if (n < 0 || n >= (int)sizeof(log_buff)) {
+        log_buff.clear();
+        int n = vsnprintf(log_buff.data(), log_buff.capacity(), msg, argp);
+        if (n < 0) {
             // LCOV_EXCL_START
             fprintf(stderr, "Log message construction failed\n");
             // LCOV_EXCL_STOP
+        } else if ((unsigned)n >= log_buff.capacity()) {
+            log_buff.reserve((unsigned)n + 1);
+            vsnprintf(log_buff.data(), (unsigned)n + 1, msg, argp_copy);
```
ktbarrett, line 87:
```suggestion
            n = vsnprintf(log_buff.data(), (unsigned)n + 1, msg, argp_copy);
```

cocotb/share/lib/gpi_log/gpi_logging.cpp
```diff
@@ -125,10 +119,27 @@ extern "C" void gpi_native_logger_vlog(const char *name, int level,
...
+    log_buff.clear();
+    int n = vsnprintf(log_buff.data(), log_buff.capacity(), msg, argp);
+    if (n < 0) {
+        // LCOV_EXCL_START
+        fprintf(stderr, "Log message construction failed: (error code) %d\n",
+                n);
+        // LCOV_EXCL_STOP
+    } else if ((unsigned)n >= log_buff.capacity()) {
+        log_buff.reserve((unsigned)n + 1);
+        vsnprintf(log_buff.data(), (unsigned)n + 1, msg, argp_copy);
```
ktbarrett, line 136:
```suggestion
        n = vsnprintf(log_buff.data(), (unsigned)n + 1, msg, argp_copy);
```

cocotb/share/lib/py_gpi_log/py_gpi_logging.cpp
```diff
@@ -129,8 +141,8 @@ static void py_gpi_log_handler(void *, const char *name, int level,
 error:
     // LCOV_EXCL_START
     /* Note: don't call the LOG_ERROR macro because that might recurse */
-    gpi_native_logger_vlog(name, level, pathname, funcname, lineno, msg,
-                           argp_copy);
+    gpi_native_logger_log(name, level, pathname, funcname, lineno,
+                          log_buff.data());
```
marlonjames, line 145: The `goto error`s at the top of the function will skip updating `log_buff`, so it would print whatever the last log message was.

marlonjames (changes requested):

cocotb/share/lib/py_gpi_log/py_gpi_logging.cpp
ktbarrett: I can move all the Python checks to below the formatting. The message needs to be formatted and put into the buffer regardless, and if the formatting fails we don't want to try to print the message anyways. Maybe, I'll just rewrite this whole function and use the `defer` from gpi_embed.cpp to make my life a little easier. `goto error` sucks.

ktbarrett:
This change went a little off the rails.
* Moved `DEFER` into cocotb_util.h and made it a C++-only header (because it now has a template). This is fine as it's really a private header, only C++ source files use it.
* Rewrote py_gpi_log_handler in py_gpi_logging.cpp to use `DEFER` instead of `goto error`, and moved the fallback logging into a separate function for reuse instead handling that in the `error` label.

cocotb/share/lib/gpi_log/gpi_logging.cpp
```diff
@@ -125,10 +120,30 @@ extern "C" void gpi_native_logger_vlog(const char *name, int level,
...
+        // LCOV_EXCL_STOP
+    } else if ((unsigned)n >= log_buff.capacity()) {
+        log_buff.reserve((unsigned)n + 1);
+        n = vsnprintf(log_buff.data(), (unsigned)n + 1, msg, argp_copy);
+        if (n < 0) {
+            // LCOV_EXCL_START
+            fprintf(stderr,
+                    "Log message construction failed: (error code) %d\n", n);
+            // LCOV_EXCL_STOP
+            return;
```
marlonjames, line 145:
```suggestion
            return;
            // LCOV_EXCL_STOP
```

cocotb/share/lib/py_gpi_log/py_gpi_logging.cpp
```diff
@@ -24,127 +36,130 @@ static void py_gpi_log_handler(void *, const char *name, int level,
...
-    // refcount cleanup
-    PyObject *logger_name_arg = NULL, *filename_arg = NULL, *lineno_arg = NULL,
-             *msg_arg = NULL, *function_arg = NULL;
+    static std::vector<char> log_buff(512);
+
+    log_buff.clear();
+    int n = vsnprintf(log_buff.data(), log_buff.capacity(), msg, argp);
+    if (n < 0) {
+        // LCOV_EXCL_START
+        fprintf(stderr, "Log message construction failed\n");
```
marlonjames, line 51:
```suggestion
        fprintf(stderr, "Log message construction failed: (error code) %d\n",
                n);
```

marlonjames (commented):
Just to make the code consistent between gpi and py_gpi.

cocotb/share/include/cocotb_utils.h
```diff
@@ -77,8 +72,24 @@ extern COCOTBUTILS_EXPORT int is_python_context;
...
 }
-#endif
 
-#endif /* COCOTB_UTILS_H_ */
+#define DEFER1(a, b) a##b
+#define DEFER0(a, b) DEFER1(a, b)
+#define DEFER(statement) \
+    auto DEFER0(_defer, __COUNTER__) = make_deferable([&]() { statement; });
+
+#endif /* COCOTB_UTILS_HPP_ */
```
ktbarrett, line 95:
```suggestion
#endif /* COCOTB_UTILS_H_ */
```

cocotb/share/include/cocotb_utils.h
```diff
@@ -27,8 +27,8 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  ******************************************************************************/
 
-#ifndef COCOTB_UTILS_H_
-#define COCOTB_UTILS_H_
+#ifndef COCOTB_UTILS_HPP_
+#define COCOTB_UTILS_HPP_
```
ktbarrett, line 31:
```suggestion
#ifndef COCOTB_UTILS_H_
#define COCOTB_UTILS_H_
```

ktbarrett (commented):
The filename wasn't changed at this time.

marlonjames (approved):

PR #2782, opened 2021-11-22
Fix regression in TestFactory wrt unique test names

ktbarrett:
Closes #2781. BTW the last line breaks https://github.com/cocotb/cocotb/pull/2721#discussion_r717707761. We just can't win with this stuff can we?

PR #2788, opened 2021-11-25
Fix regression in TestFactory wrt unique test names

ktbarrett:
Cherry-pick of #2782 w/ test.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -633,6 +633,11 @@ def _create_test(function, name, documentation, mod, *args, **kwargs):
     async def _my_test(dut):
         await function(dut, *args, **kwargs)
 
+    _my_test.__name__ = name
+    _my_test.__qualname__ = name
+    _my_test.__doc__ = documentation
+    _my_test.__module__ = mod.__name__
+
     return cocotb.test()(_my_test)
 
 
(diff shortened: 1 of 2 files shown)


ktbarrett:
Odd that names are different on 3.6... Needs more investigation.

imphil:
I'll wait with the release until we're sure that this fix is the right one. Users needing this fix urgently can always cherry-pick it or install the 1.6 branch.

ktbarrett:
@imphil The names are correct, see [here](https://github.com/cocotb/cocotb/runs/4328028340?check_suite_focus=true#step:25:688), that is the important part that fixes the reported problem.

The test, as I wrote it, is not working. My guess is that what I'm testing didn't work in 1.5 either.

imphil:
For those of you who don't want to dig into the logs: the test failure on Python 3.6.7 is
>  AssertionError: assert {'_my_test'} == {'cYlBrun_testfactory_testvUCT_001', 'cYlBrun_testfactory_testvUCT_002', 'cYlBrun_testfactory_testvUCT_003', 'cYlBrun_testfactory_testvUCT_004'}

ktbarrett:
So the tests wouldn't pass before either. Going to have to modify the test to just check stdout. No clue why this isn't working though...

ktbarrett:
Big dumb. I was testing against the test's Task object instead of the Test function object in the regression manager.

imphil (approved):
Great, thanks! I got to the same conclusion before I saw the update here :-)

PR #2793, opened 2021-11-29
Fix write scheduling problem

### LINKED ISSUE #2792: Handle writes not called in the order they were scheduled
Based on gitter discussion here:
https://gitter.im/cocotb/Lobby?at=61a061ee8853d316401680f3
Consider the example:
```
a.b.array= [0xFFFF,0xFFFF,0xFFFF]
a.b.array[0][0]=0
await Timer(1)
print ("ROW - %d=%x"% (0,a.b.array[0].value) )
```

this should print` ROW - 0=0x7FFF` but `prints 0xFFFF`.

This appears to be because of a bug in how the following data structure is being used in scheduler.py

`self._write_calls = _py_compat.insertion_ordered_dict()`

The handle for `a.b.array[0][0]=0` is being executed before the handle for `a.b.array[0]=0xFFFF`

@ktbarrett shared a fix which appears to work for me here:  https://gist.github.com/ktbarrett/874008b482f7f3ff618cacd4836f761b

ktbarrett:
Closes #2792 

The previous solution did not guarantee that writes were applied oldest to newest. This causes problems when two handles alias the same value, such as setting a logic vector with an int, then later setting a single element of that vector. The previous solution counted on no two handles in the write cache aliasing; however this assumption is false.

The fix works by using `.popitem(last=False)` which iterates over the value in the dict in insertion order (rather than reverse insertion ordered like the previous solution). We guarantee that forward insertion order is equivalent to oldest to newest write by deleting and re-adding the scheduled write to the cache if the handle is already in the cache.

This most certainly will cause a performance regression in code that performs lots of writes, though I'm unsure how much yet.

### TODO
- [x] add test (may be possible with Icarus, Questa or Riviera?)

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -40,6 +40,7 @@ import threading
 import inspect
 import warnings
 from typing import Any, Union
+from collections import OrderedDict
 from collections.abc import Coroutine
 
 import cocotb
@@ -231,9 +232,10 @@ class Scheduler:
         # Our main state
         self._mode = Scheduler._MODE_NORMAL
 
-        # A dictionary of pending (write_func, args), keyed by handle. Only the last scheduled write
-        # in a timestep is performed, all the rest are discarded in python.
-        self._write_calls = _py_compat.insertion_ordered_dict()
+        # A dictionary of pending (write_func, args), keyed by handle.
+        # Writes are applied oldest to newest (least recently used).
+        # Only the last scheduled write to a particular handle in a timestep is performed.
+        self._write_calls = OrderedDict()
 
         self._pending_coros = []
         self._pending_triggers = []
@@ -261,7 +263,7 @@ class Scheduler:
             await self._read_write
 
             while self._write_calls:
-                handle, (func, args) = self._write_calls.popitem()
+                handle, (func, args) = self._write_calls.popitem(last=False)
                 func(*args)
             self._writes_pending.clear()
 
@@ -286,7 +288,7 @@ class Scheduler:
             self._timer1.prime(self._test_completed)
             self._trigger2coros = _py_compat.insertion_ordered_dict()
             self._terminate = False
-            self._write_calls = _py_compat.insertion_ordered_dict()
+            self._write_calls = OrderedDict()
             self._writes_pending.clear()
             self._mode = Scheduler._MODE_TERM
 
@@ -555,6 +557,8 @@ class Scheduler:
         if self._write_coro_inst is None:
             self._write_coro_inst = self.add(self._do_writes())
 
+        if handle in self._write_calls:
+            del self._write_calls[handle]
         self._write_calls[handle] = (write_func, args)
         self._writes_pending.set()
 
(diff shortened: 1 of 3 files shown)


tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -331,6 +328,20 @@ async def test_last_scheduled_write_wins_array(dut):
     assert dut.array_7_downto_4.value == [10, 2, 3, 4]
 
 
+# Most simulators do not support setting the value of a single bit of a packed array
+@cocotb.test(skip=not cocotb.SIM_NAME.lower().startswith("questa"))
```
ktbarrett, line 332:
```suggestion
@cocotb.test(skip=not cocotb.SIM_NAME.lower().startswith("modelsim"))
```

ktbarrett:
Questa w/ VPI does not blow up trying to access the bit, but it doesn't look like it behaves reasonably. It still has the 10 from the previous test, while the other indexes were updated to 0. 
```
# COUT:                          assert dut.array_7_downto_4.value == [1, 0, 0, 0]
# COUT:                      AssertionError: assert [00001011, 00000000, 00000000, 00000000] == [1, 0, 0, 0]
# COUT:                       +  where [00001011, 00000000, 00000000, 00000000] = NonHierarchyIndexableObject(sample_module.array_7_downto_4).value
# COUT:                       +    where NonHierarchyIndexableObject(sample_module.array_7_downto_4) = HierarchyObject(sample_module).array_7_downto_4
```

Questa w/ FLI works. Riviera w/ VPI does not work either, exhibiting the same bug as Questa's VPI. Riviera w/ VHPI passes. We only need one test to prove that behavior change works, so I'll try to limit the test to run Questa and Riviera with VHDL only. I'll also rebase stuff so it is clear the test fails before the change.

ktbarrett:
See the Questa and Riviera VHDL tests in the first commit failing, then passing in the next commit.

marlonjames (approved):

cocotb/scheduler.py
```diff
@@ -231,9 +232,10 @@ def __init__(self):
         # Our main state
         self._mode = Scheduler._MODE_NORMAL
 
-        # A dictionary of pending (write_func, args), keyed by handle. Only the last scheduled write
-        # in a timestep is performed, all the rest are discarded in python.
-        self._write_calls = _py_compat.insertion_ordered_dict()
+        # A dictionary of pending (write_func, args), keyed by handle.
+        # Writes are applied oldest to newest (least recently used).
+        # Only the last scheduled write to a particular handle in a timestep is performed.
+        self._write_calls = OrderedDict()
```
imphil, line 238: We only support cpython 3.6+, which guarantees insertion ordered dicts (the Python "language" guarantees it since 3.7). You can simply use `{}` here.

imphil (commented):
LGTM with one comment.

cocotb/scheduler.py
ktbarrett: Normal dicts don't support `.popitem(last=False)`.

imphil: oh indeed.

PR #2802, opened 2021-12-08
Close file descriptors in test_custom_entry Makefile

imphil:
Python 3.8 is reporting the following when running test_custom_entry:

```
python -c 'assert open("results.log").readlines() == open("expected_results.log").readlines()'
Exception ignored in: <_io.FileIO name='results.log' mode='rb' closefd=True>
ResourceWarning: unclosed file <_io.TextIOWrapper name='results.log' mode='r' encoding='UTF-8'>
Exception ignored in: <_io.FileIO name='expected_results.log' mode='rb' closefd=True>
ResourceWarning: unclosed file <_io.TextIOWrapper name='expected_results.log' mode='r' encoding='UTF-8'>
```

That's not a functional problem, but we can easily work around that by
using `filecmp` instead.

### DIFF
--- a/tests/test_cases/test_custom_entry/Makefile
+++ b/tests/test_cases/test_custom_entry/Makefile
@@ -4,6 +4,6 @@ export PYGPI_ENTRY_POINT := custom_entry:entry_func
 .PHONY: override_for_this_test
 override_for_this_test:
 	-$(MAKE) all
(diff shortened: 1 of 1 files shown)


ktbarrett (approved):
Python has a ***builtin*** module for doing file comparison? The memes are true...

```python
import solution
```

PR #2816, opened 2021-12-25
Friendly pytest warning

raysalemi:
I've changed the pytest warning to a less alarming warning:

```
0.00ns WARNING  Install pytest to enable better AssertionError messages
```

Is this a correct description?

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -325,7 +325,8 @@ def _initialise_testbench_(argv_):
     try:
         import pytest
     except ImportError:
-        log.warning("Pytest not found, assertion rewriting will not occur")
+        log.warning(
+            "Install pytest to enable better AssertionError messages")
     else:
         try:
             # Install the assertion rewriting hook, which must be done before we


cocotb/__init__.py
```diff
@@ -325,7 +325,8 @@ def _initialise_testbench_(argv_):
     try:
         import pytest
     except ImportError:
-        log.warning("Pytest not found, assertion rewriting will not occur")
+        log.warning(
```
ktbarrett, line 328:
```suggestion
        log.info(
```

cocotb/__init__.py
```diff
@@ -325,7 +325,8 @@ def _initialise_testbench_(argv_):
...
+            "Install pytest to enable better AssertionError messages")
```
eric-wieser, line 329: Perhaps clearer as
```suggestion
            "pytest not found, install it to enable better AssertionError messages")
```

raysalemi: I like it!

ktbarrett (approved):

eric-wieser (commented):
Looks fine to me

PR #2828, opened 2022-01-09
ci: generate list of tasks through a Python script

umarcor:
This PR is a draft to dynamically generate the GitHub Actions jobs/tests/tasks through a Python script. Currently, it is a plain conversion from the YAML fields into Python. However, it can be extended to read environment variables and/or context fields.

/cc @ktbarrett

### DIFF
--- a/.github/generate-tasks.py
+++ b/.github/generate-tasks.py
@@ -2,104 +2,76 @@
 
 
 tasks = [
-  # Test different Python versions with package managed Icarus on Ubuntu
-  {
-    'sim': 'icarus',
-    'sim-version': 'apt',
-    'lang': 'verilog',
-    # lowest version according to https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
-    'python-version': '3.6.7',
-    'os': 'ubuntu-20.04'
-  },
-  {
-    'sim': 'icarus',
-    'sim-version': 'apt',
-    'lang': 'verilog',
-    # lowest version according to https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
-    'python-version': '3.7.1',
-    'os': 'ubuntu-20.04'
-  },
-  {
-    'sim': 'icarus',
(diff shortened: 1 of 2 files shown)


.github/generate-tasks.py
```diff
@@ -0,0 +1,77 @@
...
+        'lang': 'verilog',
+        'sim': 'icarus', 'sim-version': 'master',    # use msvc instead of mingw
+        'os': 'windows-latest', 'python-version': '3.8',
+        'toolchain': 'msvc', 'extra_name': 'msvc | '
+    },
+    # Other
+    {
+        'lang': 'verilog',
+        'sim': 'icarus', 'sim-version': 'v11_0',    # use clang instead of gcc
+        'os': 'ubuntu-20.04', 'python-version': '3.8"',
```
ktbarrett, line 72:
```suggestion
        'os': 'ubuntu-20.04', 'python-version': '3.8',
```

.github/generate-tasks.py
```diff
@@ -0,0 +1,77 @@
...
+        'lang': 'vhdl',
+        'sim': 'ghdl', 'sim-version': 'nightly',
+        'os': 'ubuntu-latest', 'python-version': '3.8'
+    },
+    # Test Verilator on Ubuntu
+    {
+        'lang': 'verilog',
+        'sim': 'verilator', 'sim-version': 'master',
+        'os': 'ubuntu-20.04', 'python-version': '3.8',
+        'may_fail': 'true'
```
ktbarrett, line 50:
```suggestion
        'may_fail': true
```

ktbarrett:
@imphil Is this still valid?

imphil:
No. We by now do something similar, actually based on this PR. I thought I mentioned that when I did the other work and closed this PR, but apparently that's not the case.

Thanks for your work, @umarcor! I wasn't convinced in this PR, but we ultimately found the use case that requires it :)

PR #2832, opened 2022-01-21
Fix example doc reference to TestFactory

jonpovey:
TestFactory was removed from matrix multiplier example by ad981341


jonpovey:
Separately, I found this because I went looking for a `TestFactory` example, and note that there is no example that shows `TestFactory` usage. (but I can find it in tests).

ktbarrett (approved):

PR #2853, opened 2022-02-05
Separate build and test for python runner

themperek:
This allows to run the parallel example with `pytest-xdist`:
```bash
pytest -m compile
pytest -m "not compile" -n auto
```

~~Not sure if we should put all those arguments in `test()` like with `build_dir` , `library_name` or `parameters` ? @ktbarrett~~

Need to add some optional arguments to `test()`
- `build_dir`
- ~~`library_name` -> needed for Icarus executable file name could use some generic name and not touch this.~~
- `parameters` -> needed for GHDL/Questa/Riviera

### DIFF
--- a/cocotb/runner.py
+++ b/cocotb/runner.py
@@ -47,6 +47,10 @@ class Simulator(abc.ABC):
 
         self.env: Dict[str, str] = {}
 
+        # for running test() independently of build()
+        self.build_dir = "sim_build"
+        self.parameters = {}
+
     @abc.abstractmethod
     def simulator_in_path(self) -> None:
         """Check that the simulator executable exists in `PATH`."""
@@ -78,9 +82,6 @@ class Simulator(abc.ABC):
 
         self.env["TOPLEVEL"] = self.sim_toplevel
         self.env["MODULE"] = self.module
-
-        if not os.path.exists(self.build_dir):
-            os.makedirs(self.build_dir)
 
     @abc.abstractmethod
     def build_command(self) -> Sequence[Command]:
@@ -145,11 +146,19 @@ class Simulator(abc.ABC):
         extra_env: Mapping[str, str] = {},
         waves: Optional[bool] = None,
         gui: Optional[bool] = None,
+        parameters: Mapping[str, object] = None,
+        build_dir: Optional[PathLike] = None,
         sim_dir: Optional[PathLike] = None,
     ) -> PathLike:
         """Run a test."""
 
         __tracebackhide__ = True  # Hide the traceback when using pytest
+
+        if build_dir is not None:
+            self.build_dir = build_dir
+
+        if parameters is not None:
+            self.parameters = dict(parameters)
 
         if sim_dir is None:
             self.sim_dir = self.build_dir
@@ -344,6 +353,10 @@ class Icarus(Simulator):
             for name, value in parameters.items()
         ]
 
+    @property
+    def sim_file(self) -> PathLike:
+        return os.path.join(self.build_dir, "sim.vvp")
+
     def test_command(self) -> List[Command]:
 
         return [
@@ -363,8 +376,6 @@ class Icarus(Simulator):
 
         if self.vhdl_sources:
             raise ValueError("This simulator does not support VHDL")
-
-        self.sim_file = os.path.join(self.build_dir, self.library_name + ".vvp")
 
         cmd = []
         if outdated(self.sim_file, self.verilog_sources) or self.always:
@@ -457,7 +468,17 @@ class Questa(Simulator):
         if not self.gui:
             do_script += "run -all; quit"
 
+        fli_lib_path = cocotb.config.lib_name_path("fli", "questa")
+
         if self.toplevel_lang == "vhdl":
+
+            if not os.path.isfile(fli_lib_path):
+                raise SystemExit(
+                    "ERROR: cocotb was not installed with an FLI library, as the mti.h header could not be located.\n\
+                    If you installed an FLI-capable simulator after cocotb, you will need to reinstall cocotb.\n\
+                    Please check the cocotb documentation on ModelSim support."
+                )
+
             cmd.append(
                 ["vsim"]
                 + ["-gui" if self.gui else "-c"]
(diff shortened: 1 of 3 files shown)


themperek:
Maybe should probably rise exception if not provided or default?

themperek:
Maybe can add something like "self.did_build" and based on this require some extra settings or return warnings?

ktbarrett:
> Not sure if we should put all those arguments in test() like with build_dir , library_name or parameters ?

This is exactly why I didn't like the current approach where all the options were heaped into the Simulator object. `build` should instead produce a `Library`/`Build` object with the pertinent fields saved. This is also necessary for describing pre-built libraries like IP cores. `test` would take a `Library`/`Build` object. Perhaps even `build` would take `Library`/`Build` dependencies for doing separate compilation.

themperek:
> > Not sure if we should put all those arguments in test() like with build_dir , library_name or parameters ?
> 
> This is exactly why I didn't like the current approach where all the options were heaped into the Simulator object. `build` should instead produce a `Library`/`Build` object with the pertinent fields saved. This is also necessary for describing pre-built libraries like IP cores. `test` would take a `Library`/`Build` object instead of having `library_name` and `build_dir` passed in. Perhaps even `build` would take `Library`/`Build` dependencies for doing arbitrarily-nested separate compilation.

I guess everything will have its pros and cons. I removed `library_name` from `test()`

cocotb/runner.py
```diff
@@ -492,11 +501,9 @@ def test_command(self) -> List[Command]:
...
             )
 
-            if self.vhdl_sources:
-                self.env["GPI_EXTRA"] = (
-                    cocotb.config.lib_name_path("fli", "questa")
-                    + ":cocotbfli_entry_point"
-                )
+            self.env["GPI_EXTRA"] = (
+                cocotb.config.lib_name_path("fli", "questa") + ":cocotbfli_entry_point"
+            )
```
ktbarrett, line 506: You'll need to condition this on whether the FLI extension was built or not.

ktbarrett:
If we moved to `build` returning a build object, we could use pytest fixtures and not have a two step process.

```python
@pytest.fixture(scope="session")
def simulator():
    return get_runner(sim)()


@pytest.fixture(scope="session")
def build(simulator):
    return runner.build(
        always=True,
        verilog_sources=verilog_sources,
        vhdl_sources=vhdl_sources,
        toplevel=toplevel,
        build_dir=sim_build,
        extra_args=compile_args,
    )


@pytest.mark.parametrize("seed", list(range(4)))
def test_cocotb_parallel(seed, simulator, build):
    simulator.test(
        build=build,   # pass build object here
        toplevel=toplevel,
        python_search=python_search,
        py_module=module_name,
        seed=seed,
    )
```

themperek:
> If we moved to `build` returning a build object, we could use pytest fixtures and not have a two step process.
> 
> ```python
> @pytest.fixture(scope="session")
> def simulator():
>     return get_runner(sim)()
> 
> 
> @pytest.fixture(scope="session")
> def build(simulator):
>     return runner.build(
>         always=True,
>         verilog_sources=verilog_sources,
>         vhdl_sources=vhdl_sources,
>         toplevel=toplevel,
>         build_dir=sim_build,
>         extra_args=compile_args,
>     )
> 
> 
> @pytest.mark.parametrize("seed", list(range(4)))
> def test_cocotb_parallel(seed, simulator, build):
>     simulator.test(
>         build=build,   # pass build object here
>         toplevel=toplevel,
>         python_search=python_search,
>         py_module=module_name,
>         seed=seed,
>     )
> ```

If you use pytest-xdist it will compile in every thread (if you span 4 runners fixture will run 4 times). This is also a use case but ...

themperek:
> It looks like many people are interested in such a feature, but they have put no effort into implementing it, even though they have a potential viable design. [pytest-dev/pytest-xdist#271](https://github.com/pytest-dev/pytest-xdist/issues/271).
> 
> The current approach only works if you are running xdist tests locally. Builds should be done per node, but per worker may be fine, it consumes N disk space vs 1, but should take the same amount of wall time for the test to complete since it's all happening in parallel.

I have lost hope we will have the perfect solution. Just some solution. 

As it is in this PR would give full flexibility if needed. 
One can use hooks from `pytest-xdist` and do all sorts of "magic" but probably only for advanced users.

ktbarrett (approved):
We need to make sure we track the fact we need documentation for this once it's all complete and we are happy with it.

cocotb/runner.py
```diff
@@ -492,11 +511,15 @@ def test_command(self) -> List[Command]:
...
 
-            if self.vhdl_sources:
+            if os.path.isfile(fli_lib_path):
                 self.env["GPI_EXTRA"] = (
                     cocotb.config.lib_name_path("fli", "questa")
                     + ":cocotbfli_entry_point"
                 )
+            else:
+                print(
+                    "WARNING: FLI library not found. Mixed-mode simulation will not be availabe."
```
cmarqu, line 521:
```suggestion
                    "WARNING: FLI library not found. Mixed-mode simulation will not be available."
```

cocotb/runner.py
```diff
@@ -532,34 +555,37 @@ def build_command(self) -> List[Command]:
         if self.verilog_sources:
             raise ValueError("This simulator does not support Verilog")
 
-        return [
+        if self.hdl_toplevel is None:
+            raise ValueError(
+                "This simulator requires hdl_toplevel parameter to be specified"
```
cmarqu, line 560:
```suggestion
                "This simulator requires the hdl_toplevel parameter to be specified"
```

cocotb/runner.py
```diff
@@ -344,6 +353,10 @@ def get_parameter_options(self, parameters: Mapping[str, object]) -> List[str]:
             for name, value in parameters.items()
         ]
 
+    @property
+    def sim_file(self):
```
cmarqu, line 357:
```suggestion
    def sim_file(self) -> PathLike:
```
?

ktbarrett:
I probably should add a mypy check for the files that are checkable. We need to first decide if that is a goal, or if that should all be handled with cocotb-stubs.

PR #2855, opened 2022-02-09
Format with black and isort

### LINKED ISSUE #879: Use Black to have consistent formatting?
When most PRs are in should we run [Black](https://github.com/ambv/black) and continue with the consistent format?

ktbarrett:
Closes #879.

Formats the codebase (excluding cocotb/_vendor/) with isort and black. Eliminates most flake8 exceptions and fixes several lingering flake8 issues. Adds a pre-commit config file and changes the flake8 CI check to use pre-commit instead.


ktbarrett:
What a nightmare these tools are...

pre-commit by default passes the filenames of files that have changed to the tool it's supposed to run. isort, black, and flake8 all have exclude/skip/ignore file configuration (barring the fact they all are named differently and use different pattern matching), but they are set up to override that exclusion if the user passes an excluded filename explicitly on the command line. Of course, you can't differentiate the user explicitly passing the excluded filename and pre-commit passing the filenames. So it appears that the tools just ignore your exclude rules when using pre-commit.

You can configure pre-commit to *not* do it's default behavior of passing filenames and just run the tool like you normally would. This mode is probably what you want for most tools. Why this isn't the default IDK. However running flake8 normally leads to it picking up untracked files that might have formatting issues (because pre-commit doesn't stash untracked or gitignored files) or your virtualenv, which causes it to run long and error a lot. pre-commit *can't* stash untracked changes because some people run venvs out of their project directory, stashing it would break everything.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
+repos:
+- repo: "https://github.com/ambv/black"
+  rev: "21.12b0"
```
ktbarrett, line 3:
```suggestion
  rev: "22.1.0"
```

ktbarrett:
I did leave two ignores in flake8 config for things that black doesn't fix that are preferential. One in exceeding max line length, the other is trivial names.

marlonjames:
As a follow-on we may want to create one of these: https://github.com/paritytech/substrate/blob/master/.git-blame-ignore-revs
We can put the `black` and `clang-format` refactoring commits in it.

imphil (approved):
Thanks! LGTM.

Is the commit "Remove traceback tests" intentionally in here? If so, it needs a bit more explanation.

The new developer workflow needs a bit of documentation, e.g. at https://docs.cocotb.org/en/stable/install_devel.html, but we can do that in a follow-up.

tests/test_cases/test_array/test_array.py
```diff
@@ -54,93 +70,93 @@ async def test_read_write(dut):
...
-        _check_int(tlog, dut.param_int , 6)
+        _check_int(tlog, dut.param_int, 6)
         _check_real(tlog, dut.param_real, 3.14)
-        _check_int(tlog, dut.param_char, ord('p'))
-        _check_str(tlog, dut.param_str , b"ARRAYMOD")
+        _check_int(tlog, dut.param_char, ord("p"))
+        _check_str(tlog, dut.param_str, b"ARRAYMOD")
 
         if not cocotb.SIM_NAME.lower().startswith("riviera"):
-            _check_logic(tlog, dut.param_rec.a        , 0)
```
cmarqu, line 68: I suppose this file has some useful manual alignment, but I don't have strong feelings about it.

cmarqu (approved):

marlonjames (approved):
Looks good. I was also curious about the traceback tests.

ktbarrett:
@imphil @cmarqu The reason I refactored those tests was due to [this failure](https://github.com/cocotb/cocotb/runs/5120241436?check_suite_focus=true#step:25:424). The only issue was the string delimiters in the expected value which were changed by black. However I never particularly liked testing the traceback as a string since I thought that would be too fragile (turns out they were, this isn't the first time I've had to fix them), so I just removed them. But since it was noteworthy, I just put them back and fixed the expected string.

imphil:
> However I never particularly liked testing the traceback as a string since I thought that would be too fragile (turns out they were, this isn't the first time I've had to fix them), so I just removed them. But since it was noteworthy, I just put them back and fixed the expected string.

Thanks for your explanation! Sounds reasonable, but let's maybe do that in a separate PR to check what we actually gain/loose from these tests.

themperek (approved):

PR #2856, opened 2022-02-09
Codecov: Make reporting more reliable

imphil:
* Report failures only for drops >= 1%
* Update the number of expected tests with coverage


imphil:
Unfortunately, CI in this PR makes it hard to judge if the change actually works, since we have a 0.00% change in coverage. But at least it's not completely broken :-)

ktbarrett (approved):

marlonjames (approved):

imphil:
Seems like it worked: Current master shows "71.96% (-0.01%) compared to" and still passes, which wasn't the case before.

PR #2857, opened 2022-02-10
Update Python versions in CI

imphil:
* Run lint on the latest Python release
* Add a test for the upcoming Python version

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -85,6 +85,13 @@ jobs:
             python-version: "3.10"
             os: ubuntu-20.04
 
+          # A single test for the upcoming Python version.
+          - sim: icarus
+            sim-version: apt
+            lang: verilog
+            python-version: "3.11.0-alpha - 3.11.0"
+            os: ubuntu-20.04
+
             # Test Icarus dev on Ubuntu
 
           - sim: icarus
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.9]
+        python-version: ["3.10"]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{matrix.python-version}}
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -527,8 +527,9 @@ class test(coroutine, metaclass=_decorator_helper):
             Don't execute this test as part of the regression. Test can still be run
             manually by setting :make:var:`TESTCASE`.
 
-        stage (int, optional)
+        stage (int)
             Order tests logically into stages, where multiple tests can share a stage.
+            Defaults to 0.
     """
 
     _id_count = 0  # used by the RegressionManager to sort tests in definition order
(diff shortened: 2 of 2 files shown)


imphil:
I'll have a closer look tomorrow, need to get a Python 3.11 version locally first.

marlonjames:
This works on Python 3.8, but fails on 3.11:
```py3
l = [(None, 2), (None, 1)]
l.sort()
```
`stage` defaults to `None`, causing the issue in 3.11
https://github.com/cocotb/cocotb/blob/cffafa91f380158ca9ba612aec81f77a0d60ae35/cocotb/decorators.py#L443-L452
Tests are sorted here:
https://github.com/cocotb/cocotb/blob/cffafa91f380158ca9ba612aec81f77a0d60ae35/cocotb/regression.py#L139

ktbarrett:
@marlonjames Seems like we could just change the default to 0 then?

imphil:
One down, more to go:


```
   /home/runner/work/cocotb/cocotb/tests/test_cases/test_cocotb/test_scheduler.py:665: RuntimeWarning: coroutine 'test_create_task.<locals>.coro' was never awaited
    assert type(cocotb.create_task(coro())) == RunningTask
  RuntimeWarning: Enable tracemalloc to get the object allocation traceback
  /home/runner/work/cocotb/cocotb/tests/test_cases/test_cocotb/test_scheduler.py:684: RuntimeWarning: coroutine 'test_create_task.<locals>.coro' was never awaited
    assert type(cocotb.create_task(CoroType())) == RunningTask
  RuntimeWarning: Enable tracemalloc to get the object allocation traceback
  SCHEDULER ERROR: read-only sync events created RW events!
  sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name=2 mode='w' encoding='UTF-8'>
  ResourceWarning: Enable tracemalloc to get the object allocation traceback
  ```

Seems like this early testing is worthwhile :-)

ktbarrett:
The `stage` change in default to `0` and remove support for `None` fixes a bug where if you mix tests setting stage and not setting stage it breaks because comparison of `None` and and int. It's very likely that the `stage` feature could be removed and no one would miss it.

ktbarrett:
Looks like tests which test the traceback by string are failing. I will change those tests to not do that. Those tests need to go.

eric-wieser:
Regarding traceback strings, I'd lean towards:
* Adding the missing `^` lines included in new python versions to the matches
* Stripping out those lines before doing the match if the python version is too old

Sure, checking the traceback data structure would be less fragile, but it's also more work and makes it harder to read the test.

ktbarrett:
Can you justify those tests? What are they actually testing? Traceback output *is not a stable API*. Seems to me you are needlessly creating work.

ktbarrett:
The traceback tests will be fixed instead by what is mentioned in https://github.com/cocotb/cocotb/pull/2865#issuecomment-1043297331. This will continue testing the traceback, but in a less fragile way.

cmarqu (approved):

ktbarrett (approved):

imphil:
Rebased on top of master with the nox changes.

imphil:
The GHDL nightly fail is due to a missing apt-get update before installing packages in the relevant GH action, which will be resolved once they do a new release (fixed as of an hour ago: https://github.com/ghdl/setup-ghdl-ci/commit/d1153380e62073d4d426b0627fea21fe9bf8d784)

PR #2859, opened 2022-02-11
Refactorings in regression.py

imphil:
Non-functional changes in regression.py.


ktbarrett (approved):

cocotb/regression.py
```diff
@@ -127,17 +127,21 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
 
         if not self._queue:
             self.log.warning("No tests were discovered")
+            return
 
         self._queue.sort(key=lambda test: (test.stage, test._id))
 
+    @property
+    def ntests(self) -> int:
+        """Gets the number of tests in this regression."""
```
cmarqu, line 142:
```suggestion
        """Get the number of tests in this regression."""
```

imphil: Thanks for the suggestion. I kept it as-is to be in line with the rest of the file, which uses "XXXs" as well.

PR #2860, opened 2022-02-15
Introduce noxfile

ktbarrett:
As a replacement for tox (currently they exist side by side). I find tox hard to use without staring at the config documentation for longer than I'd like, all while being less flexible.

nox includes a `tox-to-nox` utility I used to convert our existing tox.ini to a noxfile.py, then I touched it up to suit us a little better. I also added a "dev" session to build a development environment for cocotb. You can use it in one of the following ways.

```sh
# build env
nox -e dev
# activate environment
. .nox/dev/bin/activate
# do your business
# deactivate the environment
deactivate
```

Or...

```sh
nox -e dev -- bash
# you are now in a nested interactive bash shell
# you are only an `exit` away from cleanup
```

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,0 +1,82 @@
+import glob
+
+import nox
+
+test_deps = ["coverage", "pytest", "pytest-cov"]
+
+dev_deps = [
+    "black",
+    "isort",
+    "mypy",
+    "pre-commit",
+    "nox",
+    "tox",
+]
+
+
+@nox.session
+def docs(session: nox.Session) -> None:
+    """invoke sphinx-build to build the HTML docs"""
+    session.install("-r", "documentation/requirements.txt")
+    session.install("-e", ".")
+    outdir = session.cache_dir / "docs_out"
+    session.run(
+        "sphinx-build", "./documentation/source", str(outdir), "--color", "-b", "html"
+    )
+    index = (outdir / "index.html").resolve().as_uri()
+    session.log(f"Documentation is available at {index}")
+
+
+@nox.session
+def docs_linkcheck(session: nox.Session) -> None:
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -0,0 +1,82 @@
+import glob
```
ktbarrett, line 1:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause

import glob
```

cmarqu (approved):
Once all these changes are settled and the tox setup gets removed, we need to update the `CONTRIBUTING.md` and `documentation/README.md`, but maybe it's a bit too early for that now.

imphil:
I'm currently looking into a similar area of code as part of the wheel build refactorings, I'll have a look at this PR in the coming days to ensure we don't go in circles.

noxfile.py
```diff
@@ -0,0 +1,85 @@
...
+
+test_deps = ["coverage", "pytest", "pytest-cov"]
+
+dev_deps = [
+    "black",
+    "isort",
+    "mypy",
+    "pre-commit",
+    "nox",
+    "tox",
```
ktbarrett, line 16:
```suggestion
    "tox",
    "flake8",
```

imphil (approved):
LGTM, this is a good starting point! I gave it a try for my wheelbuild stuff and it plays together nicely (from what I could tell) and I agree, the dev sessions are much nicer than remembering the tox invocations.

I'd suggest moving the test session to the top, this one runs when calling `nox` without further arguments.

ktbarrett:
test session moved to the top. Added `flake8` and `clang-format` to the dev session.

PR #2862, opened 2022-02-16
Fix additional dependencies

Warl0ck73:
Fix additional dependencies to the simulation target for windows

### DIFF
--- a/cocotb/share/makefiles/Makefile.inc
+++ b/cocotb/share/makefiles/Makefile.inc
@@ -93,21 +93,13 @@ COCOTB_HDL_TIMEPRECISION ?= 1ps
 
 export COCOTB_RESULTS_FILE
 
-ifeq ($(OS), Msys)
-    # Depend on all Python from the cocotb package. This triggers a
-    # recompilation of the simulation if cocotb is updated.
-    CUSTOM_SIM_DEPS += $(shell dir $(COCOTB_PY_DIR)/cocotb/*.py -b)
+# Depend on all Python from the cocotb package. This triggers a
+# recompilation of the simulation if cocotb is updated.
+CUSTOM_SIM_DEPS += $(shell $(PYTHON_BIN) -c 'import glob; print(" ".join(glob.glob("$(COCOTB_PY_DIR)/cocotb/*.py")))')
 
-    # This triggers a recompilation of the simulation if cocotb library is updated.
-    CUSTOM_SIM_DEPS += $(shell dir $(LIB_DIR)/* -b)
-else
-    # Depend on all Python from the cocotb package. This triggers a
-    # recompilation of the simulation if cocotb is updated.
-    CUSTOM_SIM_DEPS += $(shell find $(COCOTB_PY_DIR)/cocotb/ -name "*.py")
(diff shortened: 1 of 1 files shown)


cocotb/share/makefiles/Makefile.inc
```diff
@@ -93,12 +93,21 @@ COCOTB_HDL_TIMEPRECISION ?= 1ps
 
 export COCOTB_RESULTS_FILE
 
-# Depend on all Python from the cocotb package. This triggers a
-# recompilation of the simulation if cocotb is updated.
-CUSTOM_SIM_DEPS += $(shell find $(COCOTB_PY_DIR)/cocotb/ -name "*.py")
+ifeq ($(OS), Msys)
```
ktbarrett, line 96: This is the only value for OS we care about on Windows? What's the value of OS when in an Anaconda environment, or a MINGW environment?

themperek (commented):
Maybe something like this?

```bash
CUSTOM_SIM_DEPS += $(shell $(PYTHON_BIN) -c 'import glob; print(" ".join(glob.glob("$(COCOTB_PY_DIR)/cocotb/*.py")))')
CUSTOM_SIM_DEPS += $(shell $(PYTHON_BIN) -c 'import glob; print(" ".join(glob.glob("$(LIB_DIR)/*")))')
```

or maybe remove this @ktbarrett @marlonjames ?

cocotb/share/makefiles/Makefile.inc
Warl0ck73: it works on Windows and Miniconda. didnt check on MINGW

ktbarrett:
I'm not sure about removing it. I don't see the use case, but @Warl0ck73 does. But maybe using Python instead would be better.

Warl0ck73:
I use your framework by MS Visual Code + TerosHDL ext. In this case shell command evaluate by cmd. Find command has different functionality and format on Windows and Unix. So it return error.
I agree, using Python would be better.

cocotb/share/makefiles/Makefile.inc
```diff
@@ -93,12 +93,21 @@ COCOTB_HDL_TIMEPRECISION ?= 1ps
...
+    # Depend on all Python from the cocotb package. This triggers a
+    # recompilation of the simulation if cocotb is updated.
+    CUSTOM_SIM_DEPS += $(shell dir $(COCOTB_PY_DIR)/cocotb/*.py -b)
```
themperek, line 99:
```suggestion
# Depend on all Python from the cocotb package. This triggers a
# recompilation of the simulation if cocotb is updated.
CUSTOM_SIM_DEPS += $(shell $(PYTHON_BIN) -c 'import glob; print(" ".join(glob.glob("$(COCOTB_PY_DIR)/cocotb/*.py")))')

# This triggers a recompilation of the simulation if cocotb library is updated.
CUSTOM_SIM_DEPS += $(shell $(PYTHON_BIN) -c 'import glob; print(" ".join(glob.glob("$(LIB_DIR)/*")))')
```

ktbarrett (approved):

PR #2864, opened 2022-02-17
Deprecate `cocotb.fork`

### LINKED ISSUE #2663: Deprecate `cocotb.fork`
With #2660 bringing in `cocotb.start` and `cocotb.start_soon`, and these APIs replacing `cocotb.fork` we should deprecate `cocotb.fork` in favor of these APIs so it may be removed in 2.0. This is supported by the fact that a new scheduler implementation will likely not support `fork`'s semantics.

## Rationale
* `fork` causes the coroutine to be scheduled immediately, which causes recursion in the scheduler leading to bugs in corner cases that are not trivial to fix
* because of the immediate scheduling it is impossible to catch an exception that occur before the first `await`, `start_soon` does not have this problem
* Can lead to recursion exceptions in programs that recursively fork

ktbarrett:
Closes #2663. Deprecates `cocotb.fork` in place of `cocotb.start_soon` and `cocotb.start`.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -186,7 +186,19 @@ and in parameters to :class:`.TestFactory`\ s.
 
 
 def fork(coro: Union[RunningTask, Coroutine]) -> RunningTask:
-    """Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on its use."""
+    """
+    Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on its use.
+
+    .. deprecated:: 1.7.0
+    """
+    warnings.warn(
+        "cocotb.fork has been deprecated in favor of cocotb.start_soon and cocotb.start.\n"
+        "In most cases you can simply substitute cocotb.fork with cocotb.start_soon.\n"
+        "For more information about when you would want to use cocotb.start see the docs.\n"
+        "https://docs.cocotb.org/en/latest/coroutines.html#concurrent-execution",
+        DeprecationWarning,
+        stacklevel=2,
+    )
     return scheduler.add(coro)
 
 
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -190,6 +190,9 @@ def fork(coro: Union[RunningTask, Coroutine]) -> RunningTask:
     Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on its use.
 
     .. deprecated:: 1.7.0
+        This function has been deprecated in favor of :func:`cocotb.start_soon` and :func:`cocotb.start`.
+        In most cases you can simply substitute ``cocotb.fork`` with ``cocotb.start_soon``.
+        For more information on when to use ``start_soon`` vs ``start`` see :ref:`_coroutines`.
     """
     warnings.warn(
         "cocotb.fork has been deprecated in favor of cocotb.start_soon and cocotb.start.\n"
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -197,7 +197,7 @@ def fork(coro: Union[RunningTask, Coroutine]) -> RunningTask:
     warnings.warn(
         "cocotb.fork has been deprecated in favor of cocotb.start_soon and cocotb.start.\n"
         "In most cases you can simply substitute cocotb.fork with cocotb.start_soon.\n"
-        "For more information about when you would want to use cocotb.start see the docs.\n"
+        "For more information about when you would want to use cocotb.start see the docs,\n"
         "https://docs.cocotb.org/en/latest/coroutines.html#concurrent-execution",
         DeprecationWarning,
         stacklevel=2,
--- a/documentation/source/coroutines.rst
+++ b/documentation/source/coroutines.rst
@@ -55,7 +55,7 @@ Concurrent Execution
 
 Coroutines can be scheduled for concurrent execution with :func:`~cocotb.fork`, :func:`~cocotb.start`, and :func:`~cocotb.start_soon`.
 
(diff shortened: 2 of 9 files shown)


marlonjames (commented):
I'm a bit wary of removing all testing while `fork` is deprecated but still supported.

Otherwise looks good.

ktbarrett:
Should I add a couple more tests to test_deprecated that really nail down `fork` behavior?

themperek:
I would keep the test for now. Maybe make a copy?

cocotb/__init__.py
```diff
@@ -187,6 +187,11 @@ def _reopen_stream_with_buffering(stream_name):
 
 def fork(coro: Union[RunningTask, Coroutine]) -> RunningTask:
     """Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on its use."""
+    warnings.warn(
+        "fork has been deprecated in favor of cocotb.start_soon and cocotb.start",
+        DeprecationWarning,
+        stacklevel=2,
```
imphil, line 193: Deprecation warnings, IMO, should not only tell what's wrong, but also give actionable (ideally, copy/paste) advice how to change to the new behavior.

For that, we should suggest a single preferred replacement for fork, instead of giving start_soon and start as options. https://www.fossi-foundation.org/2021/10/20/cocotb-1-6-0 says: "In most cases, you can simply replace cocotb.fork() with cocotb.start_soon().". Something like that would be in order here as well.

We could also add this blog post as link to the error message to help people decide if they need start() instead. (I wanted for a while now move this content into the documentation. I'll update the link once we do that.)

imphil (commented):
I'm happy with the deprecation warnings, thanks for adding that. I think we should make the error message more actionable. 

And we really need some testing of fork to remain in cocotb. Ideally something that tests the special scheduling behavior of fork, but I guess we didn't have that before either? So at least some basic testing to ensure we don't break things.

ktbarrett:
> Ideally something that tests the special scheduling behavior of fork, but I guess we didn't have that before either?

The only test that didn't pass when changing `fork` was the `test_task_repr` because forked coroutines display "started" and start_soon display "pending". If we are going to test `fork`'s actual behavior, we will need new tests. I'll create a `test_fork` file with tests for immediate scheduling and recursive forks.

ktbarrett:
Blocked by #2865.

cocotb/__init__.py
ktbarrett: Added a longer and more descriptive message.

cocotb/__init__.py
```diff
@@ -187,6 +187,14 @@ def _reopen_stream_with_buffering(stream_name):
 
 def fork(coro: Union[RunningTask, Coroutine]) -> RunningTask:
     """Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on its use."""
+    warnings.warn(
+        "cocotb.fork has been deprecated in favor of cocotb.start_soon and cocotb.start.\n"
+        "In most cases you can simply substitute cocotb.fork with cocotb.start_soon.\n"
+        "For more information about when you would want to use cocotb.start see the docs.\n"
+        "https://docs.cocotb.org/en/latest/library_reference.html#cocotb.start",
```
marlonjames, line 194: https://docs.cocotb.org/en/latest/coroutines.html#concurrent-execution might be better?

marlonjames (commented):
With this PR or in a follow-up we should update https://github.com/cocotb/cocotb/blob/master/documentation/source/coroutines.rst to include the deprecation.

ktbarrett:
@marlonjames Should I make a section on fork like the generator-based coroutines and remove the mention of fork from the rest of the document?

marlonjames:
I think it's probably enough to do something like:
> :func:`~cocotb.fork` **(deprecated)** schedules and executes the new coroutine immediately,

And update the Note to say `fork` is deprecated and *will* be removed in the future.

marlonjames (approved):

cocotb/__init__.py
```diff
@@ -186,7 +186,19 @@ def _reopen_stream_with_buffering(stream_name):
 
 
 def fork(coro: Union[RunningTask, Coroutine]) -> RunningTask:
-    """Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on its use."""
+    """
+    Schedule a coroutine to be run concurrently. See :ref:`coroutines` for details on its use.
+
+    .. deprecated:: 1.7.0
```
ktbarrett, line 192:
```suggestion
    .. deprecated:: 1.7.0
        This function has been deprecated in favor of :func:`cocotb.start_soon` and :func:`cocotb.start`.
        In most cases you can simply substitute ``cocotb.fork`` with ``cocotb.start_soon``.
        For more information on when to use ``start_soon`` vs ``start`` see :ref:`_coroutines`.
```

cocotb/__init__.py
```diff
@@ -186,7 +186,22 @@ def _reopen_stream_with_buffering(stream_name):
...
+        This function has been deprecated in favor of :func:`cocotb.start_soon` and :func:`cocotb.start`.
+        In most cases you can simply substitute ``cocotb.fork`` with ``cocotb.start_soon``.
+        For more information on when to use ``start_soon`` vs ``start`` see :ref:`_coroutines`.
+    """
+    warnings.warn(
+        "cocotb.fork has been deprecated in favor of cocotb.start_soon and cocotb.start.\n"
+        "In most cases you can simply substitute cocotb.fork with cocotb.start_soon.\n"
+        "For more information about when you would want to use cocotb.start see the docs.\n"
```
cmarqu, line 200:
```suggestion
        "For more information about when you would want to use cocotb.start see the docs,\n"
```

cmarqu (approved):

PR #2865, opened 2022-02-17
Cleanup test_cocotb

ktbarrett:
* replaced assert_deprecated with pytest.warns
* replaced assert_raises with pytest.raises
* removed _check_traceback from tests, checking tracebacks textually is
  too fragile
* removed mostly useless clock_gen routine

Should unblock #2857.

### DIFF
--- a/tests/test_cases/test_cocotb/common.py
+++ b/tests/test_cases/test_cocotb/common.py
@@ -6,22 +6,9 @@ Common utilities shared by many tests in this directory
 """
 import re
 import traceback
-from contextlib import contextmanager
-
-from cocotb.triggers import Timer
 
 
-async def clock_gen(clock):
-    """Example clock gen for test use"""
-    for i in range(5):
-        clock.value = 0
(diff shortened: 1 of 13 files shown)


eric-wieser:
The point in the traceback checking was to verify that the traceback is actually being preserved through coroutines. There are only three of those tests, I think the benefit of keeping them outweighs the cost of maintaining them.

ktbarrett:
> The point in the traceback checking was to verify that the traceback is actually being preserved through coroutines

This could be done more easily than testing against the whole traceback. You only need to check for unique identifiers in the traceback string.

> There are only three of those tests

There only needs to be one.

ktbarrett:
I ended up leaving all three and check_traceback, but allowed the tests to test less. Basically it just tests that both `raise_inner` and `raise_soon` are in the traceback, in the correct order. Which is sufficient to test that exceptions were propagated correctly.

marlonjames (approved):

PR #2866, opened 2022-02-18
Add more pre-commit hooks

ktbarrett:
Adds clang-format, whitespace, and other pre-commit hooks, then removes those hooks from the github regression. I checked, a newline at the end of an svg document is permissible.


cmarqu:
I would prefer excluding the diagrams - I last edited them live at diagrams.net and saved/exported them to GitHub directly, without having a chance to add a final newline.

eric-wieser:
Just to check - are these "pre-commit" hooks still run by CI?

ktbarrett:
I addressed the SVGs.

ktbarrett:
@eric-wieser https://github.com/cocotb/cocotb/runs/5248845577?check_suite_focus=true#step:4:57

documentation/source/diagrams/xml/hal_cosimulation.xml
```diff
@@ -1 +1 @@
...
...47.132 Safari/537.36" type="device"><diagram>1Zhfb+I4EMA/DdLeQ1eQhLR9BArtSlepOrrq3qMvNsSSiZExBb59x8k4/8wG9gKU8gKe4InnN+OZsTv+aLF9VGQZP0vKRMfr0m3Hf+h4ntfvduHLSHaZ5Db0M8FccZqJeihYc8pWFZGWUmi+rAojmSQs0hXZTIqqsiWZM0cwjYhwpW+c6jiT3nlhIX9ifB7b1/TC++zJSu+sDspmZC30TSqCZ+bxglhdqd3+GNgoKUGN+bXYjpgwfKzpGZzJb57mi1QswYU0TwgQ9jsRa1yks+pNzDWbLklkxhtwWscfUrKKmdGBb3xnSjN04Z5FpCJcwSOTC6bVDv6CE/p3uAr0eM+GwKYA3QtRFpcg58sn6PB5rruwGX6g2fsR+OinT0UQWJObEFhMp0dwewUIbjBUGxHcng1B/woQ9PqHEUCCPBMCz42C6duPRweDkuuEos3+cB+UGRdiJIVU6QR/MhmNxiHIVzGhcoMT54pQDmAqfzSfU4C8r4G0fA7Fkp3XiqMbSq9spdtzHIeG5EU5HoHRxmgZY56qWnHE0lri+DT4+xThOA4vjLEWjt6xqc0/BcY7d1vHfAESwf9TBJV98Q0OXeKRRDHJtSLqdgwPP1/bYwSIfaBzQYxhre3wQsxdZYy2Lpcx2m6sVcl1e88B/JbJzfNzx2x+f0vA/gfFwcavmD0Dmy5zusH3S/K1LUlDS8MSOlAqpREJslrxCJiAeWr3C0RdO/jXDGDtZkgnwLVoeBh1zk01PvBCuVapZ0pe10TNmW1f0g3lclRMEA3eryhrhQQJlJCwrWYqIeLbXy4cMG2Kw0Qm8DVkW64NGRM65ncGBip+E7X/xcl2ERVOafJyOR0sx1Z4NFB8x4vksMQi61rnYUAH9TKfGYCzysfOmqI8ClCRX1eUWe0oSv2bG36cy/H2oORyxQg17p5sFGSNfY5v3hW4EayHT7ErbK9T8XZauC+xKwIH0WydRJrL5M/hnCFl2A6mDCcLoEvAcbtQmFNjAjUC1jUkgs8TwwWsgZLlD0354HCHNMAHC06pmTM8QXXJj6wNh8V9NwYnOS27PeXLTscyuQ4yh0/R5wITuDX3isAER9wvnI2MW3r/eXVPc5+BJbxcwMCwuNzNqllxC+6PPwA=</diagram></mxfile>
```
imphil, line 1: Add this file to the exclude list as well, please. (As a general rule: All auto-generated files should be excluded.)

imphil (approved):
LGTM, thanks!

I already see problems with users not being able to run clang-format, which seems to be a binary that gets downloaded on demand and built for Ubuntu 20.04, so it doesn't work on RHEL7 and similar distributions. But let's address that if it comes up.

documentation/source/diagrams/xml/hal_cosimulation.xml
ktbarrett: The other XMLs in the same directory have newlines already.

.pre-commit-config.yaml
```diff
@@ -24,10 +24,20 @@ repos:
...
+
+- repo: "https://github.com/pre-commit/pre-commit-hooks"
+  rev: "v4.1.0"
+  hooks:
+  - id: "trailing-whitespace"
+  - id: "mixed-line-ending"
+    args:
+    - "--fix=lf"
+  - id: "end-of-file-fixer"
+    exclude: "^documentation/source/diagrams/svg/"
```
ktbarrett, line 43:
```suggestion
    exclude: "^documentation/source/diagrams/(svg|xml)/"
```

ktbarrett:
> I already see problems with users not being able to run clang-format, which seems to be a binary that gets downloaded on demand and built for Ubuntu 20.04, so it doesn't work on RHEL7 and similar distributions

clang-format is a manylinux_2_17 wheel (equivalent to manylinux2014), which supports RHEL7, Ubuntu 16.04, Debian 8, and more.

imphil:
> clang-format is a manylinux_2_17 wheel (equivalent to manylinux2014), which supports RHEL7, Ubuntu 16.04, Debian 8, and more.

Ah nice, then I looked that up wrong. Great!

marlonjames (approved):

ktbarrett:
Rolled back the xml file change. After it passes I'll merge.

Issue #2843, opened 2022-01-28
Dropping support for mingw environments and compiler


ktbarrett:
This includes dropping official support and installation instructions for anaconda + mingw and the msys2 and mingw32/64 environment. 

## Rationale
These environments have been broken a lot recently. We have fed issues upstream and the issues have not been resolved. And basic functionality in that environment has always been a little broken: virtualenv support is broken which requires user add the `--no-build-isolation` flag; and their Python configuration for finding libpython was also always a little broken. Despite this, CI works fine, which is a false flag.

## Actions
With #2634 coming in, we can recommend on Windows one of two paths:
1. anaconda or python.org installer + Visual Studio as a compiler + python test runner
2. WSL

- [ ] Change the installation instructions in the documentation
- [ ] Add existing installation instructions for these environments to the Wiki for posterity and mark them as "UNSUPPORTED".
- [ ] Add a warning to the documentation we are dropping support.
- [ ] Add a warning to the build script that GCC compilations on Windows are not officially supported.
- [ ] Close any new or outstanding issues or PRs for msys2 or mingw environments.

ktbarrett:
The mingw compiler does "just work" in anaconda. The only issue is that you can't select the mingw compiler if visual studio is available, but ultimately that shouldn't matter. There is a path forward for mingw environments and those instructions are already in the "tier 2 installation instructions" section. I don't think there is really anything to be done here.

We should start recommending WSL instead of mingw on Windows. Perhaps a blurb in the documentation to that end?

imphil:
mingw-env tests are removed from CI in https://github.com/cocotb/cocotb/pull/2964.

umarcor:
VSCode + `Remote - Containers` + `gcr.io/hdl-containers/sim/osvb` is also an interesting alternative on Windows (equivalent to the Gitpod badge).

themperek:
It seems to be working (for now) https://github.com/cocotb/cocotb/pull/3706

PR #2867, opened 2022-02-18
Suggest user to use WSL in Windows instructions

ktbarrett:
Addresses some of #2843. This follows what we decided in the last meeting.

The second part of the instructions which recommend Conda + mingw compiler will be replaced with python.org installer once we have wheels and are happy with the python runner.

### DIFF
--- a/documentation/source/install.rst
+++ b/documentation/source/install.rst
@@ -34,8 +34,11 @@ The installation instructions vary depending on your operating system:
 
    .. group-tab:: Windows
 
-      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for Windows.
+      We recommend users who are comfortable with a Unix shell and who are running Windows 10 or 11 use Windows Subsytem for Linux (WSL).
+      See `here <https://docs.microsoft.com/en-us/windows/wsl/install>`_ for installation instructions of WSL,
+      then follow the Linux installation instructions for cocotb.
 
+      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for users who are more comfortable with Windows.
       Download and install `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_ from https://conda.io/.
       From an Anaconda Prompt, use the following line to install a compiler (GCC or Clang) and GNU Make:
 
--- a/documentation/source/install.rst
+++ b/documentation/source/install.rst
@@ -37,7 +37,7 @@ The installation instructions vary depending on your operating system:
       We recommend users who are running Windows and who are more comfortable with a Unix shell,
       or who have legacy Makefile-based projects,
       to use Windows Subsystem for Linux (WSL).
-      After installing WSL or a Virtual Machine and a supported Linux distribution, follow the Linux installation instructions for cocotb.
+      After installing WSL and a supported Linux distribution, follow the Linux installation instructions for cocotb.
 
       `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for users who are more comfortable with native Windows development.
       Download and install `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_ from https://conda.io/.
--- a/documentation/source/install.rst
+++ b/documentation/source/install.rst
@@ -34,8 +34,9 @@ The installation instructions vary depending on your operating system:
 
    .. group-tab:: Windows
 
-      We recommend users who are running Windows, and who are more comfortable with a Unix shell or who have legacy Makefile-based projects,
-      to use either Windows Subsystem for Linux (WSL) or a Virtual Machine, such as Oracle VirtualBox.
+      We recommend users who are running Windows and who are more comfortable with a Unix shell,
+      or who have legacy Makefile-based projects,
+      to use Windows Subsystem for Linux (WSL).
       After installing WSL or a Virtual Machine and a supported Linux distribution, follow the Linux installation instructions for cocotb.
 
       `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for users who are more comfortable with native Windows development.
--- a/documentation/source/install.rst
+++ b/documentation/source/install.rst
@@ -34,11 +34,11 @@ The installation instructions vary depending on your operating system:
 
    .. group-tab:: Windows
 
-      We recommend users who are comfortable with a Unix shell and who are running Windows 10 or 11 use Windows Subsystem for Linux (WSL).
-      See `here <https://docs.microsoft.com/en-us/windows/wsl/install>`_ for instructions on installing WSL,
-      then follow the Linux installation instructions for cocotb.
+      We recommend users who are running Windows, and who are more comfortable with a Unix shell or who have legacy Makefile-based projects,
+      to use either Windows Subsystem for Linux (WSL) or a Virtual Machine, such as Oracle VirtualBox.
+      After installing WSL or a Virtual Machine and a supported Linux distribution, follow the Linux installation instructions for cocotb.
 
-      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for users who are more comfortable with Windows.
+      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for users who are more comfortable with native Windows development.
       Download and install `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_ from https://conda.io/.
       From an Anaconda Prompt, use the following line to install a compiler (GCC or Clang) and GNU Make:
 
--- a/documentation/source/install.rst
+++ b/documentation/source/install.rst
@@ -34,8 +34,8 @@ The installation instructions vary depending on your operating system:
 
    .. group-tab:: Windows
 
-      We recommend users who are comfortable with a Unix shell and who are running Windows 10 or 11 use Windows Subsytem for Linux (WSL).
-      See `here <https://docs.microsoft.com/en-us/windows/wsl/install>`_ for installation instructions of WSL,
+      We recommend users who are comfortable with a Unix shell and who are running Windows 10 or 11 use Windows Subsystem for Linux (WSL).
+      See `here <https://docs.microsoft.com/en-us/windows/wsl/install>`_ for instructions on installing WSL,
       then follow the Linux installation instructions for cocotb.
 
       `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for users who are more comfortable with Windows.


documentation/source/install.rst
```diff
@@ -34,8 +34,11 @@ The installation instructions vary depending on your operating system:
 
    .. group-tab:: Windows
 
-      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for Windows.
+      We recommend users who are comfortable with a Unix shell and who are running Windows 10 or 11 use Windows Subsytem for Linux (WSL).
+      See `here <https://docs.microsoft.com/en-us/windows/wsl/install>`_ for installation instructions of WSL,
+      then follow the Linux installation instructions for cocotb.
```
marlonjames, line 39:
```suggestion
      We recommend users who are comfortable with a Unix shell and who are running Windows 10 or 11 use Windows Subsystem for Linux (WSL).
      See `here <https://docs.microsoft.com/en-us/windows/wsl/install>`_ instructions on installing WSL,
      then follow the Linux installation instructions for cocotb.
```

marlonjames (approved):

documentation/source/install.rst
marlonjames: hah, yeah

ktbarrett:
> What's the simulator story there?

I presume you are concerned about GUI features? There are two possibilities. Windows 11 has WSLg, which is builtin GUI support with WSL, but it's pretty bad compared to the alternative right now. Alternatives are X servers for Windows, of which there are a couple. I use VcXsrv. That works flawlessly from what I can tell.

marlonjames:
I use WSL on Windows 10 for iverilog, GHDL, and Verilator. But those are non-GUI.

I have successfully used Aldec Active-HDL with both system Python and conda on Windows 7 and 10.

ktbarrett:
The thought is on Windows to recommend one of 2 solutions.
1. Use WSL is you are comfortable with a Unix shell *instead of* mingw, msys, cygwin, etc, etc, etc.
2. Use Anaconda + mingw/msvc if you are comfortable with Windows.

Right now we have no Tier 1 supported recommendation for point 1 since most alternative Unix subsystems are broken or hokey. This PR is adding an official recommendation for point 1 to use WSL, implying that other Unix subsystems are not recommended. The recommendation for point 2 will stay the same for now. The recommendation for point 2 will change in the future to system Python installation + pre-built wheel, so no need for a compiler + Python runner, so no need for make. The anaconda solution will be relegated to Tier 2 and installation instructions moved to the Wiki.

The end result is something that Windows people will love and Linux people who are forced to use Windows will love.

imphil:
I have a hard time seeing the user base for WSL: Windows users who are familiar with Linux, don't want to use a proprietary simulator, and don't want to use a full Linux in a VirtualBox or HyperV VM?

Or in other words: Given the support matrix for proprietary simulators I don't think "being comfortable with Unix: yes/no" is the right decision criteria. It's more: do you want to use a proprietary simulator: yes/no? And then the rest falls into place.

Looking at our user survey two years ago, the majority of Windows users are using Aldec or Mentor simulators. 

Given the really limited amount of resources we have available for Windows support: How about just dropping WSL and providing a single Tier 1 solution that works on Windows together with proprietary simulators? And I agree, system python + wheels seems right now the best solution, and is very close to becoming a reality.

And if users want an easy Linux "apt-get install icarus gtkwave; pip install cocotb; make my_test; gtkwave *.fst" experience, just use a Ubuntu VM.

ktbarrett:
> Given the support matrix for proprietary simulators

I'm not sure where you are getting this from? They all work on WSL. GUI stuff requires extra effort, but it's not difficult, or really our responsibility. There are a million medium articles on setting up an X server for WSL. There are also advantages of WSL that are not available to a VM. I haven't tried it, but it is theoretically possible to program a board from WSL.

> Given the really limited amount of resources we have available for Windows support: How about just dropping WSL

WSL requires no additional support from us.

I think you are missing the point. We need to recommend _something_ for users who are forced to use Windows (common) and would prefer a Unix environment (possible) or have legacy makefile-based projects (possible) or they will pick the _something_ and inevitably come bothering us about it when it doesn't work. Recommending WSL is a solution, recommending a VM is also a solution, but a less ideal one for people doing EDA IMO.

gretzteam:
I have been using cocotb on wsl with xcelium for about 2 years. I beleive this is the best setup for windows users. Setting up the gui is trivial which is only really needed for the waveform viewer. Everything else is done through the fantastic vscode with wsl extension.
This setup allows you to use windows for day to day task (if you click on a xls file it opens it in excel, not libreoffice.😉). At the same time you have a full blown Linux dev machine.
I highly recommend it.

imphil:
> I have been using cocotb on wsl with xcelium for about 2 years.

And I assume you're using the Linux version of Xcelium installed into WSL?

gretzteam:
> > I have been using cocotb on wsl with xcelium for about 2 years.
> 
> And I assume you're using the Linux version of Xcelium installed into WSL?

Yes. I would actually do the same for any simulator as it's just easier to install in Linux. Verilator, icarus...even python. 
Then Linux stuff runs on Linux and you use windows for interacting with the computer. Where you draw the line is up to the user.

ktbarrett:
Another justification for recommending something over not recommending anything https://gitter.im/cocotb/Lobby?at=621cb0ffc61ef0178e4df216.

cmarqu (approved):

documentation/source/install.rst
```diff
@@ -34,8 +34,11 @@ The installation instructions vary depending on your operating system:
 
    .. group-tab:: Windows
 
-      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for Windows.
+      We recommend users who are comfortable with a Unix shell and who are running Windows 10 or 11 use Windows Subsystem for Linux (WSL).
+      See `here <https://docs.microsoft.com/en-us/windows/wsl/install>`_ for instructions on installing WSL,
+      then follow the Linux installation instructions for cocotb.
```
ktbarrett, line 39: @imphil Is this any better?

```suggestion
      We recommend users who are running Windows, and who are more comfortable with a Unix shell or who have legacy Makefile-based projects
      to use either Windows Subsystem for Linux (WSL) or a Virtual Machine, such as Oracle VirtualBox.
      After installing WSL or a Virtual Machine and a supported Linux distribution, follow the Linux installation instructions for cocotb.
```

documentation/source/install.rst
```diff
@@ -34,8 +34,11 @@ The installation instructions vary depending on your operating system:
...
 
+      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for users who are more comfortable with Windows.
```
ktbarrett, line 41:
```suggestion
      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for users who are more comfortable with native Windows development.
```

umarcor:
For GUIs on Windows (either on WSL or using containers) check [x11docker](https://github.com/mviereck/x11docker/) and [runx](https://github.com/mviereck/runx). See https://hdl.github.io/containers/ug/GUI.html.

Alternatively, as explained in https://github.com/ghdl/docker/blob/master/USE_CASES.md#environments-with-guis GTKWave can use GTK3's Broadway feature and be used through a browser. Container `ghdl/ext:broadway` includes the following script: https://github.com/ghdl/docker/blob/master/broadway.sh. Note that some builds of GTKWave use GTK2; this needs GTK3.

VSCode's Docker tab allows automating the options to start containers with x11docker or with predefined ports bind to the host.

umarcor:
> it is theoretically possible to program a board from WSL since you have direct access to Windows devices.

@ktbarrett that works for UART/COM and disk drives only, not for FPGAs (libusb, FTDI, etc.). In order to program FPGAs from a container (say HyperV, say WSL2), USB/IP is required. See https://hdl.github.io/containers/ug/USBIP.html.

documentation/source/install.rst
```diff
@@ -34,8 +34,11 @@ The installation instructions vary depending on your operating system:
 
    .. group-tab:: Windows
 
-      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for Windows.
+      We recommend users who are running Windows, and who are more comfortable with a Unix shell or who have legacy Makefile-based projects,
+      to use either Windows Subsystem for Linux (WSL) or a Virtual Machine, such as Oracle VirtualBox.
```
ktbarrett, line 38:
```suggestion
      We recommend users who are running Windows and who are more comfortable with a Unix shell or who have legacy Makefile-based projects to use either Windows Subsystem for Linux (WSL) or a Virtual Machine such as Oracle VirtualBox.
```
Better grammar I think? @cmarqu

ktbarrett:
@umarcor What are the differences between using runx or x11docker and running an X11 server and configuring your `DISPLAY`?

> that works for UART/COM and disk drives only

That's a shame.

ktbarrett:
Should we point to a Wiki page which contains more in-depth living documentation on setting up WSL2 for cocotb's use?

umarcor:
> What are the differences between using runx or x11docker and running an X11 server and configuring your `DISPLAY`?

x11docker is a 10kLOC bash script to automate running the server and passing ports/sockets/files to the container. It's essentially the same. However, x11docker includes multiple security and performance related features. It applies sane options to prevent the server being accessed by any other than the intended container, it supports xpra/wayland/others, window or desktop mode, etc. I strongly recommend to read the two page paper and the wiki.

Disclaimer: I triggered the author to support windows and I helped him test with MSYS2, cygwin, WSL, xming, vcxsrv... Then, runx was split from x11docker, which I also helped test.

documentation/source/install.rst
cmarqu: Uh, not sure really... based on feeling I might even add a comma before the "or who have".

documentation/source/install.rst
```diff
@@ -34,8 +34,12 @@ The installation instructions vary depending on your operating system:
 
    .. group-tab:: Windows
 
-      `Conda <https://conda.io/>`_ is an open-source package and environment management system that we recommend for Windows.
+      We recommend users who are running Windows and who are more comfortable with a Unix shell,
+      or who have legacy Makefile-based projects,
+      to use Windows Subsystem for Linux (WSL).
+      After installing WSL or a Virtual Machine and a supported Linux distribution, follow the Linux installation instructions for cocotb.
```
ktbarrett, line 40:
```suggestion
      After installing WSL and a supported Linux distribution, follow the Linux installation instructions for cocotb.
```

PR #2879, opened 2022-02-26
Remove 3.6 support

### LINKED ISSUE #2844: Drop Python 3.6 support
Python 3.6 is EOL as of the end of 2021. 1.7 should be the last release to support Python 3.6. It might be a good idea to evaluate what can be changed by dropping 3.6. One thing I know about is OrderedDict in the compat layer (still needs to stay in the scheduler).

ktbarrett:
Closes #2844.

* removes Python 3.6 from supported version list
* trusts pip to error on unsupported Python version file in setup.py
* removes 3.6 from regressions
* runs `pyupgrade --py37-plus` on all python files (besides cocotb/_vendor)
* removes _py_compat library as all components were for <3.7 support


cmarqu:
There is another mention of "3.6+" in the `README.md` (that I had forgotten to change last time).

ktbarrett:
And the private CI apparently... it's still running 3.6.

imphil (changes requested):
Let's sort if we really *need* to drop Python 3.6 in https://github.com/cocotb/cocotb/issues/2844.

PR #2882, opened 2022-02-28
Add support for Cadence Xcelium to Python runner

cmarqu:

### DIFF
--- a/cocotb/runner.py
+++ b/cocotb/runner.py
@@ -53,7 +53,7 @@ class Simulator(abc.ABC):
 
     @abc.abstractmethod
     def simulator_in_path(self) -> None:
-        """Check that the simulator executable exists in `PATH`."""
+        """Check that the simulator executable exists in :envvar:`PATH`."""
 
         raise NotImplementedError()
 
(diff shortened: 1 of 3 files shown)


cocotb/runner.py
```diff
@@ -805,6 +804,112 @@ def test_command(self) -> List[Command]:
...
+            elif self.verilog_sources and not self.vhdl_sources:
+                return "verilog"
+            else:
+                raise ValueError("Must specify a toplevel_lang in a mixed design")
+        elif toplevel_lang in ("verilog", "vhdl"):
+            return toplevel_lang
+        else:
+            raise ValueError(
+                f"Xcelium does not support {toplevel_lang!r} as a toplevel_lang"
+            )
```
themperek, line 826: This seems to be repeating maybe can put this in the `Simulator` class?

cocotb/runner.py
```diff
@@ -805,6 +804,112 @@ def test_command(self) -> List[Command]:
...
+    def build_command(self) -> List[Command]:
+
+        self.env["CDS_AUTO_64BIT"] = "all"
+        cmd = [
+            ["xrun"]
+            + ["-logfile xrun_build.log"]
+            + ["-elaborate"]
+            + [f"-xmlibdirname {self.build_dir}/xrun_snapshot"]
+            + ["-licqueue"]
+            # TODO: way to switch to these verbose messages?:
```
themperek, line 849: Can add parameters for this specific simulator something like :
```python3
    def __init__(self, *argv, verbose=False, **kwargs):
        self.verbose = verbose
        super().__init__(*argv, **kwargs)
```

themperek (approved):

cocotb/runner.py
cmarqu: Indeed, I'll do this in another PR.

cocotb/runner.py
cmarqu: I was thinking about whether this should also be controllable by an environment variable - either an existing/documented one (`COCOTB_LOG_LEVEL`) or an independent new one...

themperek: Could also add parameters via pytest plugin.

PR #2877, opened 2022-02-26
Improvements to RunningTask

ktbarrett:
Closes #2876.

* adds typing
* adds asyncio-esque interface
* deprecates old interfaces which duplicate functionality

Deciding to not deprecate `.kill()` at this point because it's used *a lot*. Wanted to make sure everyone was happy with this before going through that effort.

PR #2889, opened 2022-03-02
Move Task and friends to cocotb.task

ktbarrett:
From https://github.com/cocotb/cocotb/pull/2877#issuecomment-1054415770. I'm being purposefully non-descriptive about Task's "friends" since I think they probably should be private; they are also not documented.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -43,10 +43,11 @@ from typing import Dict, List, Optional, Union
 import cocotb.handle
 
 # Things we want in the cocotb namespace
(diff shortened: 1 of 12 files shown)


ktbarrett:
Eh, I'm not so certain this is necessary.

marlonjames:
Well, we could put everything in a single module if we wanted...
I've never been a fan of `decorators.py`

ktbarrett:
I'm not a fan either. I think we just need a plan and some goals before we go about moving stuff about. For right now, it works.

PR #2897, opened 2022-03-06
Add runner cli

cmarqu:
A "TODO" block is near the top, with a few other TODOs further below.


cocotb/runner.py
```diff
@@ -934,13 +959,201 @@ def clean(recursive: bool = False) -> None:
...
+        help="Verilog include directories",
+    )
+    parser.add_argument(
+        "--library-name",
+        default="work",
+        help="Library name to compile into (default: %(default)s)",
+    )
+    parser.add_argument(
+        "--parameters",
+        "--generics",  # TODO: discuss this alt name
```
ktbarrett, line 1032: is there any way to support both?

cocotb/runner.py
```diff
@@ -934,13 +959,201 @@ def clean(recursive: bool = False) -> None:
...
+        dest="parameters",
+        nargs="*",
+        metavar=("NAME=VALUE"),
+        action=ParseKwargs,
+        help="Verilog parameters or VHDL generics",
+    )
+    parser.add_argument(
+        "--plus-args",
+        "--plusargs",  # TODO: discuss this alt name
```
ktbarrett, line 1041: Much prefer `--plusargs`.

cocotb/runner.py
```diff
@@ -934,13 +959,201 @@ def clean(recursive: bool = False) -> None:
...
+        dest="plus_args",
+        nargs="*",
+        default=[],
+        help="'plusargs' to set for the simulator",
+    )
+    parser.add_argument(
+        "--py-module",
```
ktbarrett, line 1048: I'd much prefer `--test-module` or just `--module`.

cocotb/runner.py
cmarqu: Both options work, or do you mean we should differentiate paramters from generics down to where the simulator option names are involved?

cocotb/runner.py
cmarqu: I agree, but lets move renaming to a different PR after this one is through.

cocotb/runner.py
ktbarrett: No, just leave them the alias rather than selecting one, whereas with the other, I would remove the original name.

cocotb/runner.py
cmarqu: Done this anyway now.

cmarqu:
I pushed again with name changes (most of which we discussed in the call). @themperek wanted to have these early to be able to convert more examples and test to use the runner and to write tests for the runner itself.

It would be good to get some more eyes on these names soon.

More work is to be done on the CLI part in the dict/list "grey area" of some arguments.

documentation/source/runner.rst
```diff
@@ -0,0 +1,91 @@
...
+
+See the full list of supported runner arguments by running ``cocotb-runner -h``.
+
+``cocotb-runner`` will set the return code ``1``
+if no tests have been run at all
+or if any failures have been found in the result files.
+It will return ``0`` otherwise.
+
+
+Usage with pytest
```
themperek, line 44: Maybe add that one could also run just do `python examples/simple_dff/test_dff.py` ?

cocotb/runner.py
```diff
@@ -97,148 +163,196 @@ def test_command(self) -> Sequence[Command]:
 
     def build(
         self,
-        library_name: str = "work",
+        hdl_library: str = "work",
         verilog_sources: Sequence[PathLike] = [],
         vhdl_sources: Sequence[PathLike] = [],
         includes: Sequence[PathLike] = [],
-        defines: Sequence[str] = [],
+        defines: Mapping[str, object] = {},
```
themperek, line 170: Did we decide to use the list for defines (some bad experience from edalize)?

cocotb/runner.py
```diff
@@ -2,16 +2,26 @@
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
 
+"""Build HDL and run cocotb tests."""
+
+# TODO: --clean cleans both build and test dir (if different). Do we also need --clean-build and --clean-test?
+# TODO: maybe do globbing and expanduser/expandvars in --include, --vhdl-sources, --verilog-sources
+# TODO: create a short README and a .gitignore (content: "*") in both build_dir and test_dir? (Some other tools do this.)
+# TODO: support timescale
+
```
themperek, line 11:
```suggestion
# TODO: support  custom dependencies

```

cocotb/runner.py
```diff
@@ -97,148 +163,196 @@ def test_command(self) -> Sequence[Command]:
...
             os.remove(results_xml_file)
 
-        cmds = self.test_command()
+        cmds: Sequence[Command] = self.test_command()
         self.set_env()
-        self.execute(cmds, cwd=self.sim_dir)
+        self.execute(cmds, cwd=self.test_dir)
 
-        check_results_file(results_xml_file)
+        # only when running under pytest, raise exception with failing testcases
```
themperek, line 334: Why only for pytest?

cocotb/runner.py
```diff
@@ -921,26 +1035,278 @@ def get_runner(simulator_name: str) -> Type[Simulator]:
...
+                key, value = value.split("=")
+            except ValueError:
+                raise SystemExit(
+                    f"ERROR: Expecting option format {option_string} [NAME=VALUE [NAME=VALUE ...]]"
+                )
+
+            getattr(namespace, self.dest)[key] = value
+
+
+def get_parser() -> argparse.ArgumentParser:
```
themperek, line 1072: Would need (at some point) to tests for this.

ktbarrett (commented):
The gpi_interface should be a list of interfaces to load. The first one will be the entry (until I refactor all of that to enter into the GPI from a single point). No matter what the toplevel lang is, that order should be preserved, because it is possibly valid no matter the toplevel_lang. gpi_interface can be defaulted based on toplevel_lang if people find that easy. And a toplevel_lang of `None` should be valid which does not export the variable. cocotb.LANGUAGE is nonsense and really only useful for our regressions because it ultimately shouldn't matter to the user and with simulators that support multiple tops, it is possible to have multiple languages at the top.

cocotb/runner.py
cmarqu: Yes, this is not reworked yet.

cocotb/runner.py
cmarqu: Added some explanatory comment.

cocotb/runner.py
cmarqu: Thought about this some more.

If we have this as a list, we would have to handle the format per simulator on our own in each of the tests and examples where we are using defines.
No such problem with a dict.

In case we *do* get the argument format wrong for edge cases, the user can just not use this option  and provide the settings in the correct simulator-specific format via `build_args`.

cocotb/runner.py
```diff
@@ -97,148 +163,196 @@ def test_command(self) -> Sequence[Command]:
 
     def build(
         self,
-        library_name: str = "work",
+        hdl_library: str = "work",
```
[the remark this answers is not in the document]
cmarqu: Used `cocotb` now, but it's also confusing... is `cocotblib` any better? Other suggestions?

cmarqu:
I'm not finding much time for this these days, sorry.

cocotb/runner.py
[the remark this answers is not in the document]
ktbarrett: Perhaps `top`?

themperek:
FYI: `HDL_TOPLEVEL_LANG` is not compatible with our ci and would require duplication in many places or some translation in the nox.

cmarqu:
> Should we make most of those private https://cocotb--2897.org.readthedocs.build/en/2897/library_reference.html#api-runner ?

We could also make the whole runner private (`import _runner`) - the `cocotb-runner` command could stay the same and print a warning with each run.

cocotb/runner.py
```diff
@@ -366,99 +475,91 @@ def test_command(self) -> List[Command]:
...
 
-        cmd = []
+        cmds = []
 
         if self.vhdl_sources:
-            cmd.append(["vlib", as_tcl_value(self.library_name)])
-            cmd.append(
+            cmds.append(["vlib", as_tcl_value(self.hdl_library)])
+            cmds.append(
                 ["vcom", "-mixedsvvh"]
```
ktbarrett, line 538: Not everyone has licenses for mixedsvvh mode. Also there are VHDL 2008 features that work in normal mode that don't work in mixedsvvh mode. This flag should only be added if a library has mixed sources.

Also this shows the issues with separate VHDL_SOURCES and VERILOG_SOURES. In mixed language designs the order of VHDL vs VERILOG sources matters, so they should not be considered separately.

ktbarrett (commented):
I'm a little surprised that there aren't different subcommands for the cocotb-runner for building and running. I think this thing needs more serious work before we start recommending it.

imphil:
Removing this from the 1.7 milestone to give us more time to address remaining issues.

cocotb/runner.py
themperek: >Not everyone has licenses for mixedsvvh mode.

Addressed in https://github.com/cocotb/cocotb/pull/3103

>Also this shows the issues with separate VHDL_SOURCES and VERILOG_SOURES. In mixed language designs the order of VHDL vs VERILOG sources matters, so they should not be considered separately.

This should be possible. Multiple `build()` calls in given order. Do you have something else in mind.

cocotb/runner.py
```diff
@@ -97,148 +132,208 @@ def test_command(self) -> Sequence[Command]:
 
     def build(
         self,
-        library_name: str = "work",
+        hdl_library: str = "cocotb",
```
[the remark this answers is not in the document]
themperek: Addressed in https://github.com/cocotb/cocotb/pull/3103

imphil:
https://github.com/cocotb/cocotb/pull/3103 has been merged with some of the changes in here. 

I'm closing this PR for now, but it will stay around for when we want to get back to the CLI work.

PR #2902, opened 2022-03-09
Fix logo URLs

eric-wieser:
Looks like something went wrong with the preview, this seems like it should work.

Proof this now works: https://github.com/cocotb/cocotb/blob/eric-wieser-fix-logo/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-![cocotb logo](/../../../../cocotb-web/blob/master/assets/img/cocotb-logo.svg#gh-light-mode-only)
-![cocotb logo](/../../../../cocotb-web/blob/master/assets/img/cocotb-logo-dark.svg#gh-dark-mode-only)
+![cocotb logo](/../../../../cocotb/cocotb-web/blob/master/assets/img/cocotb-logo.svg#gh-light-mode-only)
+![cocotb logo](/../../../../cocotb/cocotb-web/blob/master/assets/img/cocotb-logo-dark.svg#gh-dark-mode-only)
 
 **cocotb** is a coroutine based cosimulation library for writing VHDL and Verilog testbenches in Python.
 


ktbarrett (approved):

README.md
```diff
@@ -1,5 +1,5 @@
-![cocotb logo](/../../../../cocotb-web/blob/master/assets/img/cocotb-logo.svg#gh-light-mode-only)
-![cocotb logo](/../../../../cocotb-web/blob/master/assets/img/cocotb-logo-dark.svg#gh-dark-mode-only)
+![cocotb logo](/../../../../cocotb/cocotb-web/blob/master/assets/img/cocotb-logo.svg#gh-light-mode-only)
+![cocotb logo](/../../../../cocotb/cocotb-web/blob/master/assets/img/cocotb-logo-dark.svg#gh-dark-mode-only)
```
imphil, line 2: How about simply using an absolute URL from cocotb.org?

ktbarrett, line 2:
```suggestion
![cocotb logo](https://www.cocotb.org/assets/img/cocotb-logo.svg#gh-light-mode-only)
![cocotb logo](https://www.cocotb.org/assets/img/cocotb-logo-dark.svg#gh-dark-mode-only)
```

eric-wieser: That doesn't work I don't think, I think they have to be relative for github to respect the hash

imphil: Hm. And relative URLs won't work on PyPi or LibreCores where we use the same README. Maybe we have to put the logo into the cocotb source tree.

eric-wieser: Oh, good point. The whole light / dark mode thing won't work too well there anyway.

eric-wieser:
Given this is strictly better than what we have in master (images that are broken everywhere), I'm going to go ahead and merge this. Feel free to revert the whole thing though if this makes things messy for PyPI / LibreCores.

README.md
eric-wieser: #2915

PR #2907, opened 2022-03-09
Configure dependabot to update GH Actions

imphil:


imphil:
There is no strict need for this right now, it's just something I had in my patch queue for a while and thought that it would be helpful.

ktbarrett (approved):
See [here](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot) for more info.

PR #2912, opened 2022-03-13
Add VHDL toplevel and a functional testcase in examples/mixed_language

maleiter:
This pull request completes the incomplete example located at examples/mixed_language. 
- Implemented the VHDL top-level entity for examples/mixed_language 
- Added a test case where the input and the output of the swapper chain are compared 

Tested with Questa IntelFPGA Starters Edition. 

closes/referes to https://github.com/cocotb/cocotb/issues/778

### DIFF
--- a/examples/mixed_language/hdl/toplevel.vhdl
+++ b/examples/mixed_language/hdl/toplevel.vhdl
@@ -1,7 +1,7 @@
 -- Example using mixed-language simulation
 --
 -- Here we have a VHDL toplevel that instantiates both SV and VHDL
---  sub entities
+-- sub entities
 library ieee;
 
 use ieee.std_logic_1164.all;
@@ -38,9 +38,12 @@ entity endian_swapper_mixed is
     );
 end;
 
-architecture impl of endian_swapper_mixed is 
+architecture impl of endian_swapper_mixed is
 
-    component endian_swapper_sv 
+    -- The SV entity is instantiated as a component because cocotb is
+    -- executing vcom before vlog. Hence the toplevel VHDL file is compiled
+    -- before the SV module. Therefore, an entity instantiation is not possible here.
+    component endian_swapper_sv
         generic (
             DATA_BYTES              : integer := 8);
         port (
@@ -69,7 +72,7 @@ architecture impl of endian_swapper_mixed is
             csr_waitrequest         : out   std_ulogic;
             csr_writedata           : in    std_ulogic_vector(31 downto 0)
         );
-  end component; 
+  end component;
 
 
     signal sv_to_vhdl_data:             std_ulogic_vector(DATA_BYTES*8-1 downto 0) ;
--- a/examples/mixed_language/hdl/toplevel.vhdl
+++ b/examples/mixed_language/hdl/toplevel.vhdl
@@ -38,9 +38,51 @@ entity endian_swapper_mixed is
     );
 end;
 
-architecture impl of endian_swapper_mixed is begin
+architecture impl of endian_swapper_mixed is 
 
-i_swapper_sv : entity work.endian_swapper_sv
+    component endian_swapper_sv 
+        generic (
+            DATA_BYTES              : integer := 8);
+        port (
+            clk                     : in    std_ulogic;
+            reset_n                 : in    std_ulogic;
+
+            stream_in_data          : in    std_ulogic_vector(DATA_BYTES*8-1 downto 0);
+            stream_in_empty         : in    std_ulogic_vector(2 downto 0);
+            stream_in_valid         : in    std_ulogic;
+            stream_in_startofpacket : in    std_ulogic;
+            stream_in_endofpacket   : in    std_ulogic;
+            stream_in_ready         : out   std_ulogic;
+
+            stream_out_data         : out   std_ulogic_vector(DATA_BYTES*8-1 downto 0);
+            stream_out_empty        : out   std_ulogic_vector(2 downto 0);
+            stream_out_valid        : out   std_ulogic;
+            stream_out_startofpacket: out   std_ulogic;
+            stream_out_endofpacket  : out   std_ulogic;
+            stream_out_ready        : in    std_ulogic;
+
+            csr_address             : in    std_ulogic_vector(1 downto 0);
+            csr_readdata            : out   std_ulogic_vector(31 downto 0);
+            csr_readdatavalid       : out   std_ulogic;
+            csr_read                : in    std_ulogic;
+            csr_write               : in    std_ulogic;
+            csr_waitrequest         : out   std_ulogic;
+            csr_writedata           : in    std_ulogic_vector(31 downto 0)
+        );
+  end component; 
+
+
+    signal sv_to_vhdl_data:             std_ulogic_vector(DATA_BYTES*8-1 downto 0) ;
+    signal sv_to_vhdl_empty:            std_ulogic_vector(2 downto 0);
+    signal sv_to_vhdl_valid:            std_ulogic;
+    signal sv_to_vhdl_startofpacket:    std_ulogic;
+    signal sv_to_vhdl_endofpacket:      std_ulogic;
+    signal sv_to_vhdl_ready:            std_ulogic;
+
+    signal csr_waitrequest_vhdl, csr_waitrequest_sv : std_ulogic;
+
+begin
+i_swapper_sv : endian_swapper_sv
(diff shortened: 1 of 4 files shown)


cmarqu (commented):
Thanks for working on this!

examples/mixed_language/tests/test_mixed_language.py
```diff
@@ -24,3 +26,64 @@ async def mixed_language_test(dut):
...
+    dut.csr_write.value = 0
+    dut.csr_writedata.value = 0
+
+    # reset cycle
+    await Timer(100, units="ns")
+    dut.reset_n.value = 1
+    await Timer(100, units="ns")
+
+    # start clock
+    cocotb.start_soon(Clock(dut.clk, 10, units='ns').start())
```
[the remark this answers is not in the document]
maleiter: I'll set up the pre-commit step and fix it in the afternoon/evening.

examples/mixed_language/hdl/toplevel.vhdl
```diff
@@ -38,9 +38,51 @@ entity endian_swapper_mixed is
     );
 end;
 
-architecture impl of endian_swapper_mixed is begin
+architecture impl of endian_swapper_mixed is 
 
-i_swapper_sv : entity work.endian_swapper_sv
+    component endian_swapper_sv
```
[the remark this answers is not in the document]
maleiter: Questa and Modelsim do not require components. This is due to the compile order of cocotb. The script _runsim.do_ first executes _vcom_ to compile VHDL then _vlog_ to compile Verilog. 
An entity instantiation in VHDL is not possible because the Verilog module is not yet in the work library. 

So currently ```i_swapper_sv : entity work.endian_swapper_sv``` gives the error

> (vcom-1195) Cannot find expanded name "work.endian_swapper_sv


If I switch the compile order of vlog and vcom in runsim.do it works fine.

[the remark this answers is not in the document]
cmarqu: Oh, right. All the more reason to add a comment.

examples/mixed_language/tests/test_mixed_language.py
```diff
@@ -24,3 +29,66 @@ async def mixed_language_test(dut):
...
+    dut.reset_n.value = 1
+    await Timer(100, units="ns")
+
+    # start clock
+    cocotb.start_soon(Clock(dut.clk, 10, units="ns").start())
+    await Timer(500, units="ns")
+
+    previous_indata = 0
+    # transmit some packages
+    for pkg in range(1, 5):
```
cmarqu, line 71:
```suggestion
    # transmit some packets
    for _ in range(1, 5):
```

examples/mixed_language/tests/test_mixed_language.py
```diff
@@ -24,3 +29,66 @@ async def mixed_language_test(dut):
...
+
+        for i in range(1, 11):
+            await RisingEdge(dut.clk)
+            previous_indata = dut.stream_in_data.value
+
+            # write stream in data
+            dut.stream_in_startofpacket.value = 1 == 1
+            # assert start of packet when i = 1
+            dut.stream_in_endofpacket.value = 1 == 10
+            # assert end of packet when i = 10
```
cmarqu, line 81:
```suggestion
            # set start of packet when i is 1
            dut.stream_in_startofpacket.value = int(i == 1)
            # set end of packet when i is 10
            dut.stream_in_endofpacket.value = int(i == 10)
```

examples/mixed_language/tests/test_mixed_language.py
```diff
@@ -24,3 +29,66 @@ async def mixed_language_test(dut):
...
+            dut.stream_in_valid.value = 0
+
+            # await stream out data
+            await RisingEdge(dut.clk)
+            await RisingEdge(dut.clk)
+
+            # compare in and out data
+            assert int(previous_indata) == int(
+                dut.stream_out_data.value
+            ), "stream in data and stream out data were different"
```
cmarqu, line 94:
```suggestion
            ), f"stream in data and stream out data were different in round {i}"
```

ktbarrett:
Questa VHDL tests are failing with `AttributeError: i_swapper_sv contains no object named reset_n`. Riviera VHDL tests are failing with `bash: -c: line 0: syntax error near unexpected token ')'`.

maleiter:
> Questa VHDL tests are failing with `AttributeError: i_swapper_sv contains no object named reset_n`.
> 
> Riviera VHDL tests are failing with
> 
> ```
> bash: -c: line 0: syntax error near unexpected token ')'
> bash: -c: line 0: 'set -o pipefail; GPI_EXTRA=.../cocotb/libs/libcocotbvpi_aldec.so):cocotbvpi_entry_point TOPLEVEL_LANG=vhdl \'
> ```

This is related to the test that was already in the file. 
It seems like there is some error with discovering the SV module with a VHDL toplevel. 

There might be a fix/workaround because when debugging the python code with debugpy, the SV module is discoverable.  After adding any SV signal to the watch window of the debugger, the test case even doesn't fail anymore. Unfortunately, I've absolutely no idea how to fix it. I definitely need some help here. 

Here are some screenshots, to make this clear: 

1. immediately after connecting the debugger, all SV attributes in the watch window are not discoverable, but they can be viewed in the variables window. 
image


2. Now after adding _verilog.clk.value_ to the watch window, all other SV attributes are visible too and the test case does not fail anymore. It does not matter which signal is added to the watch window. 
image

 
Simulator: Questa-Intel Starter FPGA Edition-64 2021.2 Revision 2021.04
Debugging with debugpy and VSCode's build in debugger


-----------------------------------------

Would it be an option to simply exclude the failing test case if the toplevel is VHDL? 
In my opinion, it'd be still an advantage to have a mixed language example with a VHDL toplevel design. Even if the SV submodules are not discoverable. The second test case still shows that the mixed language simulation is functionally working.

ktbarrett:
[Here](https://github.com/cocotb/cocotb/tree/ktbarrett-patch-1) is the fix for the Riviera issue. Since we cannot test it without this test you are writing, you should cherry-pick that into this PR.

As for the Questa issue, when using debugpy, the `_discover_all` hook is run. Something in the C++ layer's logic when using `vpi_get_handle_by_name` or whatever the FLI equivalent is fails to find the SV entity. Does this happen with a direct instantiation or only component instantiation? For now you can stick `verilog._discover_all()` in the test and see if it passes. If it does, then file an issue about it and we can merge this PR with all tests passing.

examples/mixed_language/tests/test_mixed_language.py
```diff
@@ -1,15 +1,25 @@
...
+async def mixed_language_accessing_test(dut):
     """Try accessing handles and setting values in a mixed language environment."""
     await Timer(100, units="ns")
 
     verilog = dut.i_swapper_sv
     dut._log.info("Got: %s" % repr(verilog._name))
 
+    # discover all attributes of the SV component
+    # This is a workaround since SV modules are not discovered automatically
+    # when using Modelsim/Questa and a VHLD toplevel file.
```
cmarqu, line 20:
```suggestion
    # when using Modelsim/Questa and a VHDL toplevel file.
```
Sorry, can't help it :)

maleiter:
> [Here](https://github.com/cocotb/cocotb/tree/ktbarrett-patch-1) is the fix for the Riviera issue. Since we cannot test it without this test you are writing, you should cherry-pick that into this PR.

done

> As for the Questa issue, when using debugpy, the `_discover_all` hook is run. Something in the C++ layer's logic when using `vpi_get_handle_by_name` or whatever the FLI equivalent is fails to find the SV entity. Does this happen with a direct instantiation or only component instantiation? For now you can stick `verilog._discover_all()` in the test and see if it passes. If it does, then file an issue about it and we can merge this PR with all tests passing.

When debugging an SV entity instantiation, the SV component does not show up in the debugger at all. 
The test's error message changed to 
`AttributeError: endian_swapper_mixed contains no object named i_swapper_sv`

But `verilog._discover_all()` did the job. I've created an issue.

ktbarrett (approved):
As long as the test pass we are good. The GHDL test is not passing for other reasons, no need to worry about that.

ktbarrett:
Riviera PRO fails with `AttributeError: endian_swapper_mixed contains no object named i_swapper_sv`. Just skip the test with Riviera.

cmarqu (approved):

PR #2920, opened 2022-03-17
Add more deprecations for private interfaces

ktbarrett:
Adds a `@deprecated` decorator which wraps functions. This deprecator
should work on functions, methods, properties (must be inner decorator),
and class `__init__` (to deprecate the class).

Uses this new decorator to deprecate several should-be-private interfaces.

### DIFF
--- a/cocotb/_deprecation.py
+++ b/cocotb/_deprecation.py
@@ -1,0 +1,36 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+import functools
+import warnings
+from typing import Any, Callable, Type, TypeVar
+
+AnyCallableT = TypeVar("AnyCallableT", bound=Callable[..., object])
+
+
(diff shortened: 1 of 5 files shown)


cmarqu:
Might add a note about the decorator in https://github.com/cocotb/cocotb/blob/master/CONTRIBUTING.md#deprecations-and-removals

cocotb/_deprecation.py
```diff
@@ -0,0 +1,36 @@
...
+) -> Callable[[AnyCallableT], AnyCallableT]:
+    """Emits a DeprecationWarning when the decorated function is called.
+
+    This decorator works on normal functions, methods, and properties.
+    Usage on properties requires the ``@property`` decorator to appear outside the
+    ``@deprecated`` decorator.
+    Concrete classes can be deprecated by decorating their ``__init__`` or ``__new__``
+    method.
+
+    Args
```
cmarqu, line 23:
```suggestion
    Args:
```

cmarqu (approved):

PR #2922, opened 2022-03-22
Remove unnecessary development dependencies

imphil:
The noxfile accumulated several development dependencies which are
installed in another way; installing them in the nox run isn't
necessary.

- nox itself must be installed by the time the noxfile is executed, so
  we can remove it.
- black, isort, flake8, and clang-format are installed (and versioned) through
  pre-commit, which installs those tools into a separate venv.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -8,13 +8,8 @@ import nox
 test_deps = ["coverage", "pytest", "pytest-cov"]
 
 dev_deps = [
-    "black",
-    "isort",
     "mypy",
     "pre-commit",
-    "nox",
-    "flake8",
-    "clang-format",
 ]
 
 


ktbarrett:
These tools were added so the developer could run them independent of pre-commit, I'd prefer they stay. nox must stay if `nox -e dev -- bash` is to allow recursive calls to `nox`, as nox is not installed into the virtualenv otherwise. Removing these dependencies makes the dev environment mostly useless and if that is desired, I'd say just remove the whole thing.

imphil:
Thanks for your reply @ktbarrett.

I see the point of installing nox, I'll add a comment about that.

I'm not convinced about the other ones, however. With pre-commit, we install a very specific version of those tools (and require exactly this version for CI to pass). Having two ways of doing the same checking, with likely slightly different results, is confusing, IMO. (I'm a fan of having a single way to do things that is tested and guaranteed to work, as you have probably seen :smile:.)

What are the pain points of running e.g. black through pre-commit, where running it directly works better for you?

Either way, I'm not too obsessed with this change, so a "please keep it, it makes my life so much easier" is sufficient to drop this patch completely :-)

imphil:
> My use of these was for editor plugins.

If it helps you let's keep things as they are; you also know how to deal with the fallout if the versions happen to be incompatible :smile: I only want to prevent unexperienced contributors falling into traps somewhere if we can avoid it, but for that goal it should be sufficient to only document pre-commit and nothing else.

PR #2924, opened 2022-03-23
[WIP] Build wheels

### LINKED ISSUE #2318: Distribute wheels on PyPI
We can now distribute cocotb as a wheel, except for FLI support (well maybe we can, need to talk to Mentor about that). I suggest we provide the following matrix of wheels at first:

* OS
  * Linux (`manylinux1` supports C++11)
  * Mac OS X
  * Windows-MinGW
* Arch
  * x86_64
* Python version
   * CPython 3.X `cp3`

These packages should be built and uploaded in CI on release candidate and final releases. May need to investigate distributing with Github to work around the issue of Pypi not supporting nightlies.

imphil:
(This PR is currently an experiment, I'll use this PR for CI testing and split out individual sub-parts into separate PRs. Have a look for the big picture, though.)

Closes #2318.


allexoll:
I'm not sure if this is the correct place to talk about this, but you PR could facilitate building cocotb for other archs as well. I was able to build it on my M1 Mac by adding 

```
build = "cp*-manylinux_x86_64 cp*-manylinux_i686 cp*-win_amd64 cp*-win32 cp*-macosx_x86_64 cp*-macosx_arm64"
``` 

in `pyproject.toml` 

I'm not sure how that would tie in with the publishing but it's a first step. I'll also try to test for linux-aarch64 too.

imphil:
@allexoll Good question, it's something I have to spell out better once we get closer to merging this PR :-) 

This PR intends to build wheels for all platforms we support, as listed at https://docs.cocotb.org/en/latest/platform_support.html. 

The platform support list isn't set in stone, but we do need to have a considerable user base (or growth potential) on the respective platform to add them, since we don't want to build binaries for platform we don't (or cannot) test. If you'd like to propose e.g. linux-aarch64 or m1 macs as supported platform, please file an issue and let's see where it takes us.

ktbarrett:
`nox` uses virtualenvs, there's high chance it does not work on Windows environment since virtualenv tends to not work well on most Windows environments. 

We don't "support" mingw environments anymore and this commit breaks how we were doing testing. We are already in a situation where new effort will have to be done to support testing on this environment, so feel free to remove the mingw tests in this commit.

imphil:
So, we're slowly getting there. It's looking better now, but at least one of the Windows tests is still failing. I can't see which one from the log, just yet. Another pair of eyes is appreciated.

> nox uses virtualenvs, there's high chance it does not work on Windows environment since virtualenv tends to not work well on most "alternative" environments.

I instructed nox to avoid building a virtualenv, that seems to get us further.

Do we have an issue somewhere that discusses virtualenv + windows == not good? 

> We don't "support" mingw environments anymore and this commit breaks how we were doing testing. We are already in a situation where new effort will have to be done to support testing on this environment, so feel free to remove the mingw tests in this commit.

Done in PR https://github.com/cocotb/cocotb/pull/2964

.github/workflows/build-test-release.yml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false # Keep going even if one matrix build fails.
+      matrix:
+        os:
+          - ubuntu-20.04
+          - windows-2022
+          - macos-10.15
+    steps:
+      - uses: actions/checkout@v2
```
cmarqu, line 30:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,26 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+name: Lint
```
cmarqu, line 4:
```suggestion

name: Lint
```

noxfile.py
```diff
@@ -225,6 +226,148 @@ def dev_coverage_combine(session: nox.Session) -> None:
...
+    """Build a release (sdist and bdist)."""
+    session.notify("release_build_bdist")
+    session.notify("release_build_sdist")
+
+
+@nox.session
+def release_build_bdist(session: nox.Session) -> None:
+    """Build a binary distribution (wheels) on the current operating system."""
+
+    # cibuildwheel 2.4 is broken until is broken for us until
```
cmarqu, line 269:
```suggestion
    # cibuildwheel 2.4 is broken for us until
```

.github/workflows/lint.yml
imphil: :+1: applied that to https://github.com/cocotb/cocotb/pull/2962

noxfile.py
imphil: Removed this comment since the latest cibuildwheel release should fix things for us.

ktbarrett:
I also see this.
> nox > python -m pip install pytest coverage pytest-cov
D:\a\cocotb\cocotb\noxfile.py:125: FutureWarning: Session dev_test_sim(sim='icarus', toplevel_lang='verilog', gpi_interface='vpi') does not have a virtual environment, so use of session.install() is deprecated since it would modify the global Python environment. If you're really sure that is what you want to do, use session.run("pip", "install", ...) instead.
  session.install("-e", ".")

`nox` supports multiple virtual environment backends, including venv, which probably has a higher chance of working correctly on "alternative" environments. The switch is `venv_backend="venv"` IIRC.

marlonjames:
Looks like it's `test_long_log_msg` in both Windows environments.

ktbarrett:
@marlonjames And why is `test_long_log_msg` failing? Is the log actually failing, or is it simply that the grep command isn't working as intended on Windows?

marlonjames:
It is failing: https://github.com/cocotb/cocotb/runs/6455291669?check_suite_focus=true#step:19:3143

The signal name is not printing from GPI log, but appears to be passed correctly to Python.

imphil:
Wow, what a mess. Looks like we run into a wild array of issues with various simulators and other things. I'll untangle that tomorrow. BUT: We have passing tests on Windows (!), Linux, and macOS for both sdist (dev) and wheel builds. :tada:

.github/workflows/regression-tests.yml
```diff
@@ -72,19 +78,22 @@ jobs:
...
+            sim-version: b83daa3ae36891a372655652e53c9b4eefdfcafa
             lang: verilog
             python-version: "3.8"
             os: ubuntu-20.04
 
             # Test GHDL on Ubuntu
 
           - sim: ghdl
-            sim-version: nightly
+            sim-version: v1.0.0 # Git SHA
```
ktbarrett, line 96:
```suggestion
            sim-version: v2.0.0 # Git SHA
```

ktbarrett:
Verilator v4.106 still has a ton of known issues, I'd leave it as may fail.

documentation/source/simulator_support.rst
```diff
@@ -342,6 +342,7 @@ GHDL
 
     GHDL support in cocotb is experimental.
     Some features of cocotb may not work correctly or at all.
+    At least GHDL 2.0 is required.
```
ktbarrett, line 345: Earlier versions of GHDL do work, it was just that we added exemptions to get GHDL to pass the regression sometime before 2.0 released. So this sentence is not really correct.

I'd rather not make any promises about which versions of simulators re supported outside of "the latest release".

imphil: Moved over to https://github.com/cocotb/cocotb/pull/2980/files#r878893191, this PR only contains some of the same commits for testing, but let's keep the discussion in the other PR.

imphil:
This PR is roughly ready, but will need to go in after we have sorted out our CI situation in https://github.com/cocotb/cocotb/pull/2980.

imphil:
(Updated the PR description and rebased on top of the refactored CI infrastructure. This PR should be ready for review now.)

marlonjames (changes requested):
This test step needs to be moved after testing with simulators:
`- name: Run tests without simulators`
Otherwise the later nox sessions will use the non-coverage editable install created here.

Alternatively we could always build for coverage so that we don't have an ordering dependency.

imphil:
@marlonjames I'm confused. Can you explain what needs to change?

This PR doesn't change the order or anything related to the nosim tests (at least I don't see it).

> Alternatively we could always build for coverage so that we don't have an ordering dependency.

dev builds (sdist) are always built with coverage. Release builds are never built with coverage. (That's the intent at least.)

marlonjames:
I missed that in #2980, this was added before the simulator tests.

https://github.com/cocotb/cocotb/blob/ddd33e79688097803d036be7cc51a36f40138c49/.github/workflows/regression-tests.yml#L174-L176

That results in no library code coverage:
https://github.com/cocotb/cocotb/runs/6559149795?check_suite_focus=true#step:24:4536

I found this happened with the local `dev_test` session and reported it here: https://github.com/cocotb/cocotb/pull/2968#issuecomment-1125387604

So we will need to update `noxfile.py` to either build `dev_test_nosim` with coverage, or ensure that `dev_test_sim` session is run first.

imphil:
Ah I see, thanks @marlonjames  for explaining. Let me see what solutions I can come up with.

imphil:
Btw, the coverage isn't something that's affected by this PR, or needs to be solved as part of it. Fixes to that will go into a separate PR.

imphil:
An attempt to solve the coverage issue is at https://github.com/cocotb/cocotb/pull/2989

imphil:
Review ping @cocotb/maintainers

ktbarrett (approved):
Looks good. A follow on PR might move some of the tests around to reduce the CI check latency.

PR #2925, opened 2022-03-23
Refactor development build steps in noxfile

imphil:
Split the development build steps in the noxfile in anticipation of the
release build steps.

- Split the tests run against a simulator from the pytests. The
  simulator tests need to be run once per simulator, the pytests only
  once overall (they are simulator-agnostic).
- Make nox more ergonomic to use by allowing the user to parametrize
  tests for different simulators directly from nox, without the need to
  set the cocotb-specific environment variables such as TOPLEVEL_LANG or
  SIM.
- Keep a single `dev_test` session (target) which has identical
  functionality to the existing "tests" target.
- Include "dev" in all step names to distinguish them from other types
  of builds (which will come at a later stage).
- Remove the pytest configuration which always runs the doctests as
  well; these tests cannot be run against an installed version of
  cocotb. Instead of using a config file to specify those files, pass the
  files to pytest directly when invoking it for an in-tree development
  build.
  To still use the config in `setup.cfg` we have to invoke pytest twice:
  once with the standard configuration in `setup.cfg`, and once with
  an explicit file list for tests in the source tree.

Usage examples:

Run all tests in development mode.
```
nox

 # or equivalently:
nox test_dev
```

Python coverage is now available at `.coverage`.

Run all development tests against Icarus and GHDL:

```
 # Show which sessions would be run
nox -l -k 'dev and (icarus or ghdl)'

 # Run them.
nox -k 'dev and (icarus or ghdl)'
```

Again, Python coverage is available at `.coverage`. This time it will be
combined coverage for Icarus and GHDL runs.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -2,10 +2,17 @@
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
 import glob
+from contextlib import suppress
+from pathlib import Path
+from typing import Dict, List, Optional, Tuple
 
 import nox
 
-test_deps = ["coverage", "pytest", "pytest-cov"]
+# Sessions run by default if nox is called without further arguments.
+nox.options.sessions = ["dev_test"]
+
+test_deps = ["pytest"]
+coverage_deps = ["coverage", "pytest-cov"]
 
 dev_deps = [
     "black",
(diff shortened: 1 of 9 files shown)


imphil:
This refactor makes it much easier to fit the release (wheel) pipeline in. See https://github.com/cocotb/cocotb/pull/2924 for the full picture. But even independently, this provides us some nice usability benefits when running local tests e.g. against two simulators, or two toplevel languages.

noxfile.py
```diff
@@ -17,20 +24,194 @@
...
+    env = env_vars_for_test(sim, toplevel_lang, gpi_interface)
+    config_str = stringify_dict(env)
+
+    # Remove a potentially existing coverage file from a previous run for the
+    # same test configuration.
+    coverage_file = Path(f".coverage.test.sim-{sim}-{toplevel_lang}-{gpi_interface}")
+    try:
+        coverage_file.unlink()
+    except FileNotFoundError:
+        pass
```
cmarqu, line 132:
```suggestion
    with suppress(FileNotFoundError):
        coverage_file.unlink()
```
and `from contextlib import suppress`.

noxfile.py
```diff
@@ -17,20 +24,194 @@
...
+
+
+def stringify_dict(d: Dict[str, str]) -> str:
+    return ", ".join(f"{k}={v}" for k, v in d.items())
+
+
+#
+# Development pipeline
+#
+# - Use nox to build a sdist; no separate build step is required.
```
cmarqu, line 84:
```suggestion
# - Use nox to build an sdist; no separate build step is required.
```

noxfile.py
```diff
@@ -17,20 +24,194 @@
...
+    session.log("Running pure-Python tests with pytest")
+    session.run(
+        "pytest",
+        "-v",
+        "--cov=cocotb",
+        "--cov-branch",
+    )
+
+    # Run pytest for files which can only be tested in the source tree, not in
+    # the installed binary (otherwise we get a "import file mismatch" error
```
cmarqu, line 177:
```suggestion
    # the installed binary (otherwise we get an "import file mismatch" error
```

imphil:
Thanks @cmarqu, updated.

noxfile.py
```diff
@@ -17,20 +24,193 @@
...
+    # Not using session.notify() here to be able to pass parameters to the
+    # session.
+    dev_test_pytest(session)
+    dev_test_sim(session, sim=None, toplevel_lang=None, gpi_interface=None)
+
 
 @nox.session
-def tests(session: nox.Session) -> None:
-    """run cocotb regression suite"""
+@nox.parametrize("sim,toplevel_lang,gpi_interface", simulator_support_matrix())
```
ktbarrett, line 106: Since we aren't ever running the full matrix in any environment I don't see the value here. The environment variables are more flexible and less work. gpi_interface will change in the future, so this creates future work for ourselves.

ktbarrett (commented):
The rest looks fine.

noxfile.py
imphil: This is actually a feature I have been waiting for for a while. Three points:
- If gpi_interface changes, this code has to change as well. That's fine, it's just code. (And gpi_interface doesn't change regularly.)
- Having a test depend on the environment makes it harder to reproduce, and easier to test the wrong thing. Good tests are isolated from that; that's what tox and nox are trying hard to do, and something we break by relying on those environment variables for test selection.
- I regularly get annoyed if jobs which are run in CI are hard to reproduce locally. There should simply be no magic in a CI setup, it should all be in scripting that I can run locally as well. Take two examples:
  - When preparing for a release I want to e.g. test against three simulators that I have installed locally, and let that just run overnight. With the new code I can do that with a single command line. With the existing setup, I'd have to write a wrapper bash script that sets the environment variables and then kicks of a run.
  - I want to give a single command line to an EDA vendor for them to test cocotb exhaustively against their simulator, with all interfaces they support. We can now do that.

imphil:
(Private CI is failing due to the rename of the tests, I'll fix that once we're done with the review here.)

imphil:
I now added a second commit, "Explicitly mark all pytests which need a simulator", which does what it says. I realized that some of the pytests are actually simulator-specific and need to be run within that session. Having a pytest marker allows us to query for such tests.

tests/pytest/test_cocotb.py
```diff
@@ -4,8 +4,12 @@
 
 import os
 
+import pytest
+
 from cocotb.runner import get_runner
 
+pytestmark = pytest.mark.simulator_required
```
ktbarrett, line 11: A simpler solution is just just move this into a different directory.

noxfile.py
```diff
@@ -17,20 +24,207 @@
...
+) -> None:
+    """Test a development version of cocotb against a simulator."""
+
     session.env["CFLAGS"] = "-Werror -Wno-deprecated-declarations -g --coverage"
     session.env["COCOTB_LIBRARY_COVERAGE"] = "1"
     session.env["CXXFLAGS"] = "-Werror"
     session.env["LDFLAGS"] = "--coverage"
-    session.install(*test_deps)
+
+    session.run("pip", "install", *test_deps, *coverage_deps)
```
ktbarrett, line 118: This is cute, but I'm not a fan. `--no-install` was really only ever intended as an additional speed-up for `-R` (reusing session) not as a way to conditionally skip certain installs. That should probably be handled with posargs and an argparser.

noxfile.py
ktbarrett: `parameterize` is intended for running the cross product of the values for each parameter in testing, at least that's what it's used for in pytest. The ability to run only a limited cross section is not the intended way to test in pytest. So either nox is misapplying the feature, or we are. In either case it seems out needs have already expanded past the capabilities of nox unless we resort to abusing features. 

I'm willing to ignore this to get the wheel build stuff in quickly.

imphil: Leaving the intention-guessing aside: 
* `parametrize` is public API of nox, and documented to be used in a similar way as we use it (have a look at https://nox.thea.codes/en/stable/config.html#parametrizing-the-session-python). 
* Using `parametrize` is beneficial over using argparse because we 
  * have all valid parameter combinations showing up in the help output (`nox -l`), and 
  * can use `nox -k` to select groups of combinations easily.
  We could of course implement similar or identical functionality ourselves with argparse, at the cost of duplicating functionality that already exists in nox.

noxfile.py
imphil: I removed this part.

tests/pytest/test_cocotb.py
ktbarrett: I prefer solutions that don't require looking at documentation. When it comes to complexity "ounces make pounds". If it brings actual value over the other possible solutions then we can leave it.

noxfile.py
ktbarrett: I'm not sure it's intention guessing, the default behavior runs everything, which clearly indicates the intended use case. But if `nox` thinks that's a valid way to use it, and `tox-gh-actions` did the same, then I guess it's general understanding though I'm not a fan.

Perhaps `nox` needs to implement something like click's `option` decorator. There is a clear distinction between the use cases of tasks for project automation vs CI testing. It seems to be trying to both at the same time, and not cleanly.

imphil: Maybe that's the root of our disagreement. You wrote:
> the default behavior runs everything

I would argue running tests against every supported simulator is a good default, also for our use case. A full (acceptance) test run of cocotb should run against all supported simulators. If we run against a subset of simulators, we also run a subset of a full test run.

tests/pytest/test_cocotb.py
imphil: > I prefer solutions that don't require looking at documentation.

Well, either way you have to know something: either you have to know that tests which require a simulator need to go into a specific directory, or you need to know that a marker is needed.

I initially had a comment on top of the line, but removed it, because it read "This test requires a simulator", which is exactly what the line below says. I'll create a wiki page (after merging this change) to document some useful test writing tips.

imphil:
pre-commit in CI failed due to a black problem, which got fixed in the latest black release. See https://github.com/psf/black/issues/2964 for details on the issue. I included a commit to update the version of black in pre-commit here. If we cannot get this PR merged today or tomorrow I'll create a separate PR for this commit.

ktbarrett (approved):

PR #2947, opened 2022-04-24
Unbreak GHDL nightly CI

imphil:
GHDL nightly builds for Ubuntu 20.04 are not provided for mcode any
more, but only for llvm. Let's switch to that.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -214,7 +214,7 @@ jobs:
       if: startsWith(matrix.os, 'ubuntu') && matrix.sim == 'ghdl' && matrix.sim-version == 'nightly'
       uses: ghdl/setup-ghdl-ci@c4e457b6596ee2872274c3b7d2615ad8ddacd70c
       with:
-        backend: mcode
+        backend: llvm
 
       # Install Verilator
     - name: Set up Verilator (Ubuntu - apt)


imphil:
GHDL nightly is fixed by this PR, but we see another error with Python 3.11. I'm not yet sure where the failure really comes from, but we at least have to look into 

```
2022-04-24T11:42:40.7823855Z Exception ignored in: <_io.FileIO name=2 mode='wb' closefd=True>
2022-04-24T11:42:40.7824437Z ResourceWarning: unclosed file <_io.TextIOWrapper name=2 mode='w' encoding='UTF-8'>
```

ktbarrett:
3.11 actually due to an issue with coverage package https://github.com/cocotb/cocotb/runs/6147105555?check_suite_focus=true#step:25:976.

imphil:
> 3.11 failure actually due to an issue with coverage package [cocotb/cocotb/runs/6147105555?check_suite_focus=true#step:25:976](https://github.com/cocotb/cocotb/runs/6147105555?check_suite_focus=true#step:25:976).

Ah, good catch. I missed that one when scrolling through. I filed https://github.com/cocotb/cocotb/issues/2949, as it's unrelated to what this PR is trying to accomplish.

imphil:
And I filed https://github.com/cocotb/cocotb/issues/2950 for the ResourceWarning.

ktbarrett (approved):

ktbarrett:
Is there an issue in GHDL or some announcement from them as to why they aren't providing the mcode backend in the nightly builds anymore?

imphil:
> Is there an issue in GHDL or some announcement from them as to why they aren't providing the mcode backend in the nightly builds anymore?

I haven't seen anything. But I'm slightly annoyed by the repeated breakage of the GHDL nightly regression anyways, maybe it's time to remove it -- there's little point in testing unstable code in a PR against unstable simulators.

PR #2954, opened 2022-04-29
Add Logic/LogicArray to reference card

cmarqu:
Mark BinaryValue as deprecated.

### DIFF
--- a/documentation/source/refcard.rst
+++ b/documentation/source/refcard.rst
@@ -25,7 +25,9 @@ Reference Card
 +------------------------+-----------------------------------------------------------------+
 | Assign immediately     | ``dut.mysignal.setimmediatevalue(0xFF00)``                      |
 +------------------------+-----------------------------------------------------------------+
-| Assign metavalue       | ``dut.mysignal.value = BinaryValue("X")``                       |
+| Assign metavalue       | | ``dut.mysignal.value = Logic("X")``                           |
+|                        | | ``dut.mysignal.value = LogicArray("01XZ")``                   |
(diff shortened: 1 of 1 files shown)


documentation/source/refcard.rst
```diff
@@ -25,7 +25,9 @@ Reference Card
 +------------------------+-----------------------------------------------------------------+
 | Assign immediately     | ``dut.mysignal.setimmediatevalue(0xFF00)``                      |
 +------------------------+-----------------------------------------------------------------+
-| Assign metavalue       | ``dut.mysignal.value = BinaryValue("X")``                       |
+| Assign metavalue       | | ``dut.mysignal.value = Logic("X")``                           |
+|                        | | ``dut.mysignal.value = LogicArray("HLUXZ")``                  |
```
ktbarrett, line 29: Check me on this but IIRC LogicArray is 4-value 01XZ.
```suggestion
|                        | | ``dut.mysignal.value = LogicArray("HLUXZ")``                  |
```

ktbarrett (approved):

documentation/source/refcard.rst
cmarqu: Ok, it converts from string ``"01hHlLuUwWxXzZ-"`` but itself holds only ``01XZ``.

ktbarrett (approved):
Yeah I wondering if I should remove support for `LlHhXxUuWw-` since the conversion might be surprising.

cmarqu:
Both ways (automatically converting these values or not) might have usability problems that will only show up with prolonged use in real scenarios.

Maybe ship an explicit conversion function, or provide an example of one in the docs?

PR #2959, opened 2022-05-05
Add logger name to log output

### LINKED ISSUE #2953: Include logger name in log output
The culprit is logged, but the new "short" log format doesn't include the logger name so I don't get any useful information from this. IMO the default log format should include the logger name.
https://github.com/cocotb/cocotb/blob/c69454db92388a8915c99a35ca9cba06565fe4d5/cocotb/scheduler.py#L451

ktbarrett:
Closes #2953.

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -44,11 +44,11 @@ except ValueError:
     _suppress = 1
 
 # Column alignment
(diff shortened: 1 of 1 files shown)


marlonjames:
I'm ok with this, but I wasn't bothered by the longer log messages.

marlonjames (approved):

PR #2963, opened 2022-05-07
CI: Remove sxstrace test from CI

imphil:
Make it possible to run the sxstest through nox.

Remove it from CI, since it doesn't seem to provide enough value to run
repeatedly.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -231,9 +231,6 @@ jobs:
     - name: Install cocotb runtime dependencies (Windows, msvc)
       if: startsWith(matrix.os, 'windows') && matrix.toolchain == 'msvc'
       run: conda install --yes -c msys2 m2-base m2-make
-    - name: SxS trace test (Windows)
-      if: startsWith(matrix.os, 'windows')
-      run: .\tests\sxs.ps1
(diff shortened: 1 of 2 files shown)


imphil:
@ktbarrett I'm very unsure about the value of this test (and the script in general). Please let me know how to proceed. (We can keep it in CI, but it would be additional work, that I thought I could avoid.)

ktbarrett (approved):
It's useful for debugging sxs configuration issues. It never belonged in CI. Ideally I'd like to restructure the C++ libraries so we don't need that stuff, since I don't think any of the maintainers understand it.

ktbarrett:
Must fix formatting first ofc.

imphil:
> Ideally I'd like to restructure the C++ libraries so we don't need that stuff, since I don't think any of the maintainers understand it.

I repeatedly understood it, and forgot the details again roughly half a day later :-)

imphil:
@ktbarrett would you mind having a quick look again, since I now removed the nox part, and just added the command as comment to the powershell script. (The nox session depended on functionality in the wheelbuild PR,  and I don't think it's super valuable to have this invocation in nox anyways.)

ktbarrett (approved):

PR #2967, opened 2022-05-10
Set level of gpi logger correctly

marlonjames:
The GPI logger was changed in #2875 from "cocotb.gpi" to "gpi",
but the Python "gpi" logger was not set to match `COCOTB_LOG_LEVEL`.
Set the level on startup.

Also fix setLevel for "gpi" logger to update C++ logging level.

`test_long_log_msg` was not displaying the `DEBUG` logs from GPI.

### DIFF
--- a/cocotb/log.py
+++ b/cocotb/log.py
@@ -32,6 +32,7 @@ Everything related to logging
 import logging
 import os
 import sys
+import typing
 import warnings
 
 import cocotb.ANSI as ANSI
@@ -110,7 +111,7 @@ def default_config():
 
     # Notify GPI of log level, which it uses as an optimization to avoid
     # calling into Python.
-    simulator.log_level(log.getEffectiveLevel())
+    logging.getLogger("gpi").setLevel(level)
 
 
 class SimBaseLog(logging.getLoggerClass()):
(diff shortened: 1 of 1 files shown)


cocotb/log.py
```diff
@@ -136,10 +138,10 @@ def colour(self):
         )
         return want_color_output()
 
-    def setLevel(self, level: int) -> None:
+    def setLevel(self, level: typing.Union[int, str]) -> None:
         super().setLevel(level)
         if self.name == "gpi":
-            simulator.log_level(level)
+            simulator.log_level(self.level)
```
ktbarrett, line 144: Should this be the following? `Logger.level` is not documented.
```suggestion
            simulator.log_level(self.getEffectiveLevel())
```

cocotb/log.py
```diff
@@ -110,6 +111,7 @@ def default_config():
 
     # Notify GPI of log level, which it uses as an optimization to avoid
     # calling into Python.
+    logging.getLogger("gpi").setLevel(level)
     simulator.log_level(log.getEffectiveLevel())
```
ktbarrett, line 115: This line is now redundant.
```suggestion
```

ktbarrett (approved):

PR #2968, opened 2022-05-11
Add Python and library coverage report to end of nox dev_test_* sessions

marlonjames:
Add `coverage` and `gcovr` reports at the end of test sessions.
Don't display Python coverage during test run.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -13,6 +13,10 @@ nox.options.sessions = ["dev_test"]
 
 test_deps = ["pytest"]
 coverage_deps = ["coverage", "pytest-cov"]
+# gcovr 5.1 has an issue parsing some gcov files, so pin to 5.0. See https://github.com/gcovr/gcovr/issues/596
+# When using gcovr 5.0, deprecated jinja2.Markup was removed in 3.1, so an Exception is raised during html report generation. See https://github.com/gcovr/gcovr/pull/576
+# These issues are fixed on gcovr master branch, so next release should work.
+coverage_report_deps = ["coverage", "jinja2<3.1", "gcovr==5.0"]
 
 dev_deps = [
     "black",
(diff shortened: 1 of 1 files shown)


ktbarrett:
@marlonjames btw you can run `pre-commit install` so the linting checks are run locally before you commit.

noxfile.py
```diff
@@ -13,6 +13,7 @@
 
 test_deps = ["pytest"]
 coverage_deps = ["coverage", "pytest-cov"]
+coverage_report_deps = ["coverage", "jinja2<3.1", "gcovr==5.0"]
```
marlonjames, line 16: - There is an issue with gcovr 5.1 and gcc11 coverage files, so we pin to 5.0
- Deprecated `jinja2.Markup` was removed in jinja2 3.1, and that breaks gcovr 5.0 if you try to generate html reports.

marlonjames:
Hmm, library coverage not showing up https://github.com/cocotb/cocotb/runs/6395593448?check_suite_focus=true#step:24:4756

noxfile.py
imphil: Is there an upstream issue for that that we can link to in a comment? (To know when to remove the pinning again.)

imphil (approved):
Nice one!

marlonjames:
Any ideas why the `find` command isn't picking up the coverage files on GHA and Azure runners?

marlonjames:
So i deleted my `build/` directory and tried `nox` again, and the editable install is not producing the `*.gcno` files, so I don't have library coverage locally either

marlonjames:
I figured out that once the editable install has built the libraries once, it won't rebuild them again on the next session install.

Because `dev_test_nosim` is run first, the non-coverage build is performed and used for other sessions.
https://github.com/cocotb/cocotb/blob/17d599d3b51e0b0aed4c91c7be897a59e0959a3d/noxfile.py#L96-L100

PR #2970, opened 2022-05-13
Logger name cleanup

### LINKED ISSUE #2960: Cleanup logger names
As a follow-on to #2953, there are some logger names that need some cleanup.

#2875 changed the GPI logger from `cocotb.gpi` to `gpi`, but a few places that don't use the `LOG_*` macros need to be updated.

The `Task` logger name is `cocotb.coroutine.Task <#>.<address of Task object>`.
I think it could be improved by removing `coroutine` and the object id. It may be useful to add the wrapped coroutine name.

marlonjames:
Closes #2960.

Also fixes an issue introduced by #2670 where `RunningTest.__repr__` was of the form `<Test 123>` with the `Task` id rather than the test name.

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -101,8 +101,7 @@ class Task(typing.Coroutine[typing.Any, typing.Any, T]):
             )
         else:
             raise TypeError(
-                "%s isn't a valid coroutine! Did you forget to use the yield keyword?"
-                % inst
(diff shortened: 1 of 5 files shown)


ktbarrett (approved):
Of course, after fixing the lint.

ktbarrett (approved):

PR #2971, opened 2022-05-13
Fix code coverage reporting

marlonjames:
The line numbers for Python coverage coming from the mingw CI job do not match the other environments, and appears to be messing up the coverage in certain files.

Here is an example: https://app.codecov.io/gh/cocotb/cocotb/blob/4e313310ff18eed7850b41f74765035eb2c92b57/cocotb/triggers.py

Test a PR without including the mingw coverage.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -280,7 +280,8 @@ jobs:
       # - When changing how many builds are uploaded to codecov also update
       #   `codecov.notify.after_n_builds` in `.codecov.yml`.
       if: >
-        (steps.windowstesting.outcome == 'success'
+        ((steps.windowstesting.outcome == 'success'
+          && matrix.toolchain != 'mingw')
          || steps.unixtesting.outcome == 'success')
         && !startsWith(matrix.os, 'macos')
         && matrix.cxx != 'clang++'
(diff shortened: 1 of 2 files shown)


marlonjames:
Had to fix the `fetch-depth` issue for PRs to work.

Here is a corrected coverage report: https://app.codecov.io/gh/cocotb/cocotb/blob/de913e243c9b02cae87ce701dfc4c657015d6239/cocotb/triggers.py

ktbarrett:
Increasing the fetch depth was required because otherwise there was no commit information present? Is there a reason it's 2 instead of 1? Simply asking because 1 would be intuitive.

ktbarrett (approved):

marlonjames:
It is needed because PRs create a merge commit, and a fetch depth of 1 only gets the merge commit, but not the tip of the actual branch of the PR.
So codecov associates the coverage with the merge commit and not last commit on the PR branch.
Azure figures it out correctly, so the coverage info is split between two commits on codecov.io

.github/workflows/regression-tests.yml
```diff
@@ -135,6 +135,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
```
imphil, line 139: Can you please add a comment as to why this is necessary? We had fetch-depth in there previously for setuptools_scm, and I removed it because we don't use setuptools_scm any more. Now we need it because of codecov, which isn't obvious by looking at that line.

PR #2980, opened 2022-05-18
CI: Move non-pinned tools into a separate workflow

imphil:
Split the test workflow into two:
- One workflow, which runs on every PR and main branch commit, uses pinned (preferably released) versions of tools and Python. This workflow is designed to test cocotb changes against a known-good baseline of tools.
- An new workflow, called experimental, tests new or known-broken tool versions against a known-good version of cocotb. The purpose of this workflow is to see if newer tool versions or newer Python versions are going to break cocotb, or potentially fix a problem that cocotb depends on. These tests are scheduled to run weekly and need to be manually diagnosed. Possible actions include filing an upstream issue (if a new breakage was found), enabling a test in the main regression (if a simulator bug was fixed), or (in most cases) ignoring the error until a later time.

Notes:
- Look at the individual commit messages.
- I'm not happy with the state of Icarus testing. Ideally, we'd have a single, stable release version we test against in our normal CI. That would be v11.0 -- but this version seems to be broken on Windows? Do we know exactly which versions of Icarus work on Windows (i.e., pass our test suite)?


imphil:
(The first commit, [CI: Refactor regression test workflow](https://github.com/cocotb/cocotb/pull/2980/commits/e4f0639ad5f1f3d829d96df5a3f14f7e78109f22), is taken from the wheelbuild PR.)

imphil:
Note: This PR still needs some fiddling with GH Action syntax to actually run the right tests at the right time. For now I'd appreciate reviews on the general approach.

marlonjames:
I like the idea of splitting this, to have only one moving target at a time.

documentation/source/simulator_support.rst
```diff
@@ -342,6 +342,7 @@ GHDL
 
     GHDL support in cocotb is experimental.
     Some features of cocotb may not work correctly or at all.
+    At least GHDL 2.0 is required.
```
imphil, line 345: @ktbarrett wrote in https://github.com/cocotb/cocotb/pull/2924#discussion_r877099883:

> Earlier versions of GHDL do work, it was just that we added exemptions to get GHDL to pass the regression sometime before 2.0 released. So this sentence is not really correct.


GHDL 1.0 does not pass our test suite  as of today (I tried). And passing the test suite is the cocotb definition of "works with cocotb".

My goal here was to give our users guidance and let them know what we know. Of course, there might be use cases where GHDL 1.0 is sufficient. That's something users can then figure out if they really want/have to.

> I'd rather not make any promises about which versions of simulators re supported outside of "the latest release".

I'm not trying to make a promise (i.e. a positive support statement), but rather a negative one: versions below 2.0 are not working.

imphil:
This PR is now ready for review. It's passing the tests we're expecting to pass (i.e. all "CI" tests).

I'll remove the last commit before merging to remove all experimental CI jobs from the pull request CI; only the jobs starting with "CI /" will be left in PR CIs.

This PR contains both basic infrastructure, as well as further changes. If it turns out some of those changes are controversial I'll remove them from this PR and keep only the "easy" ones to avoid stalling the whole thing for too long.

marlonjames (commented):
As @ktbarrett found in #2973, the mingw tests are not passing, and they are not even testing the current version.

cocotb fails to build. cocotb-bus pulls in cocotb from pypi (1.6.2):
https://github.com/cocotb/cocotb/runs/6544522296?check_suite_focus=true#step:17:59

And then there are failing tests but the CI run is not stopped due to https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

I would prefer moving the mingw test to experimental until those are fixed.

imphil:
About the mingw tests: I'm sorry I forgot to add a comment on those. I saw they are broken (now and before), and just kept the status quo. The test will work again as soon as we merge the wheelbuild work (#2924).

> I would prefer moving the mingw test to experimental until those are fixed.

I agree that's a good idea. Done.

marlonjames (approved):
Looks good to me. Thanks!

Issue #2982, opened 2022-05-19
Issues with SV structs in both Questa & Xcelium (some is probably a tool bug, but others are cocotb bug)


jshaker000:
I was having issues running Questa 2020.2 and 2022.2 to read or write from system verilog structs declared as wires that have arrays under them.

To that end I made a 'test_quest_vpi' test case with a simple vpi.c and test_quest_vpi.sv.

Its also possible to see these kinds of error by running a cocotb bench and modifying the nets that I named. I beleive that the simplified VPI code runs similarly to what the cocotb code is doing because I get essentially the same seg crash

``` # calling vpi_get_value(0xec2302c0, &{vpiBinStrVal, {0}}) ... ** Fatal: (SIGSEGV) Bad handle or reference.```

I already opened a ticket with Siemens about this but I wanted to let the broader community know.

I also noticed that in `VpiImpl.cpp` that there is no `case vpiPackedArrayNet:` under either `static gpi_objtype_t to_gpi_objtype` or `*VpiImpl::create_gpi_obj_from_handle` and that *might* be an additional issue in the cocotb code base itself, but my minimal example shows that there are atleast some issues with Questa in general.

[test_questa_vpi.zip](https://github.com/cocotb/cocotb/files/8734474/test_questa_vpi.zip)

**Comment by jshaker000** [comment]
This shows the `VpiImpl.cpp` issue, you can see that its dying probably because of the missing case statement before it gets to where my vpi sim breaks. In fact this attached code dies in both Xcelium and in Questa if the lines I commented are uncommented


[test_questa_vpi_cocotb.zip](https://github.com/cocotb/cocotb/files/8734580/test_questa_vpi_cocotb.zip)

**Comment by marlonjames** [comment]
If you run cocotb with environment variable `COCOTB_LOG_LEVEL=DEBUG`, you should get more verbose logging of the VPI layer.

**Comment by jshaker000** [comment]
For the test_quest_vpi_cocotb.zip, with `COCOTB_LOG_LEVEL=DEBUG`, I get 

```
#      0.00ns DEBUG    cocotb.gpi                                GpiCommon.cpp:314  in __gpi_get_handle_by_name        Checking if in_wire_struct_array is native through implementation VPI
#      0.00ns DEBUG    cocotb.gpi                                  VpiImpl.cpp:235  in create_gpi_obj_from_handle      VPI: Not able to map type vpiPackedArrayNet(693) to object.
[...]

PR #2985, opened 2022-05-20
Add support for `vpiPackedArrayNet` type

jshaker000:
Update VpiImpl.cpp for partial fix of git issue #2982 to correctly identify vpiPackedArrayNet as a valid type. A runnable example is shown in the git issue [test_questa_vpi_cocotb.zip](https://github.com/cocotb/cocotb/files/8734580/test_questa_vpi_cocotb.zip)

### DIFF
--- a/src/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/src/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -109,6 +109,7 @@ static gpi_objtype_t to_gpi_objtype(int32_t vpitype) {
 
         case vpiInterfaceArray:
         case vpiPackedArrayVar:
+        case vpiPackedArrayNet:
         case vpiRegArray:
         case vpiNetArray:
         case vpiGenScopeArray:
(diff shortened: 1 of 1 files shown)


ktbarrett:
> A runnable example is shown in the git issue

Could you adapt that example as a test in the regression?

jshaker000:
> > A runnable example is shown in the git issue
> 
> Could you adapt that example as a test in the regression?

I think I'd need a little help to understand what exactly is required to do that, especially because in my flow I currently only use Xcelium and Questa (perhaps I should've mentioned this in my pull request above) so its not clear to me if this causes issues with other simulators (or even, for example, if that test uses an SV idion not understood by some of the simulators).

Also, I still haev a seperate ticket open with Siemans to fix a VPI bug in Questa, so essentially what my patch did was it moved a cocotb assertion to suceeding in Xcelium and having a VPI crash in Questa.

marlonjames:
You could take your SV and Python files and make a new cocotb test like `tests\test_cases\test_sv_struct_array`.
I think you'll have to exclude icarus as in `test_array` because of the lack of support.

ktbarrett:
Now that we have a better set of commercial sims in CI. I'm tempted to day this is fine to come in as is as long as it doesn't break anything (it doesn't).

cmarqu:
I agree. Commercial simulator vendors that break can either help us help them, or just help themselves.

ktbarrett (approved):

PR #2989, opened 2022-05-23
Ensure editable installs are always built with coverage

imphil:
Editable development installations are not rebuilt by nox for each session.
Since any development build (any editable install of development) could be
the "first" one, we need to ensure that they are all built in the same way.
We especially need to ensure that all editable installs are built with
coverage enabled.

This issue was found and explained by Marlon at 
https://github.com/cocotb/cocotb/pull/2924#issuecomment-1134972026. Thanks!


marlonjames (approved):
Thanks!

Issue #3005, opened 2022-06-16
Can't deregister callbacks that are already queue'd to fire


ktbarrett:
We have noticed for some time now that every so often we get messages like the one below. They always occurred when using `First` on multiple `Trigger`s. 

```
     10.00ns CRITICAL cocotb.scheduler                   No coroutines waiting on trigger that fired: RisingEdge(ModifiableObject(tb.b))
     10.00ns INFO     cocotb.RisingEdge.0x7fffe2eaadb8   I'm the culprit
```

After deep diving I determined that the `Trigger`s were firing "at the same time". However, one callback must come before the other, so whatever trigger comes in first finishes the `First` and cleans up the other `Trigger`. On simulators other than Questa, the callback associated with the cleaned up Trigger is deregistered and does not fire. On Questa, the callback is not deregistered and it runs, leading to that error message.

Not sure yet if this is limited to the FLI or Questa in general.

<details><summary>MRFE</summary>

```vhdl
entity test is
end entity test;

architecture behav of test is
  signal a : bit;
  signal b : bit;
begin
  process is
  begin
    a <= '0';
    b <= '0';
    wait for 10 ns;
    a <= '1';
    b <= '1';
    wait for 10 ns;
  end process;
end architecture behav;
```

```python
import logging
import unittest

import cocotb
from cocotb.triggers import First, RisingEdge, Timer


@cocotb.test()
async def test_first_on_coincident_trigger(dut) -> None:
    with unittest.TestCase().assertLogs("cocotb.scheduler", level=logging.CRITICAL) as logs:
        await First(
            RisingEdge(dut.a),
            RisingEdge(dut.b),
        )
        await Timer(10, 'ns')
        assert "No coroutines waiting on trigger that fired" in logs.records[0].message
```

</details>

ktbarrett:
This is probably the problem.
https://github.com/cocotb/cocotb/blob/b4a2ce033fb21b10e2a2deb0dbb95d0b9e895e9c/cocotb/share/lib/fli/FliCbHdl.cpp#L43-L49

Other implementations use `m_state` and set the callback in the `GPI_FREE` state in `cleanup_callback`. `handle_fli_callback` checks to see if the callback is in `GPI_PRIMED` before doing the callback. Since deregistered callbacks are never put into `GPI_FREE`, the callback is called when it shouldn't be. Whoops...

ktbarrett:
Closed by #3010.

PR #3010, opened 2022-06-26
Issue with Questa firing triggers that were cancelled

ktbarrett:
Addresses #3005. Justing testing now, not sure how to approach fixing it.

### DIFF
--- a/cocotb/share/lib/fli/FliImpl.cpp
+++ b/cocotb/share/lib/fli/FliImpl.cpp
@@ -1084,11 +1084,16 @@ void handle_fli_callback(void *data) {
         gpi_cb_state_e new_state = cb_hdl->get_call_state();
 
         /* We have re-primed in the handler */
-        if (new_state != GPI_PRIMED)
-            if (cb_hdl->cleanup_callback()) delete cb_hdl;
+        if (new_state != GPI_PRIMED) {
+            if (cb_hdl->cleanup_callback()) {
+                delete cb_hdl;
+            }
+        }
     } else {
         /* Issue #188 seems to appear via FLI as well */
-        cb_hdl->cleanup_callback();
+        if (cb_hdl->cleanup_callback()) {
+            delete cb_hdl;
+        }
     }
 
     gpi_to_simulator();
--- a/cocotb/share/lib/fli/FliImpl.cpp
+++ b/cocotb/share/lib/fli/FliImpl.cpp
@@ -563,6 +563,7 @@ error:
 }
 
 GpiCbHdl *FliImpl::register_timed_callback(uint64_t time) {
+    // get time from cache instead of allocating
     FliTimedCbHdl *hdl = cache.get_timer(time);
 
     if (hdl->arm_callback()) {
@@ -1057,7 +1058,19 @@ FliTimedCbHdl *FliTimerCache::get_timer(uint64_t time) {
     return hdl;
 }
 
-void FliTimerCache::put_timer(FliTimedCbHdl *hdl) { free_list.push(hdl); }
+static constexpr size_t FLI_TIMER_CACHE_SIZE =
+    256;  // Arbitrary large value, it's doubtful more than 256 simultaneous
+          // Timer triggers will be active at any time
+
+void FliTimerCache::put_timer(FliTimedCbHdl *hdl) {
+    // save FLI_TIMER_CACHE_SIZE Timer objects before deleting, this should
+    // prevent "live leaking"
+    if (free_list.size() < FLI_TIMER_CACHE_SIZE) {
+        free_list.push(hdl);
+    } else {
+        delete hdl;
+    }
+}
 
 extern "C" {
 
--- a/cocotb/share/lib/fli/FliImpl.h
+++ b/cocotb/share/lib/fli/FliImpl.h
@@ -343,6 +343,11 @@ class FliStringObjHdl : public FliValueObjHdl {
     char *m_mti_buff = nullptr;
 };
 
+/** Maintains a cache of FliTimedCbHdl objects which can be reused.
+ *
+ * Apparently allocating and freeing Timer callback objects is very expensive
+ * compared to anything Python or the simulator are doing.
+ */
 class FliTimerCache {
   public:
     FliTimerCache(FliImpl *impl) : impl(impl) {}
--- a/cocotb/share/lib/fli/FliImpl.h
+++ b/cocotb/share/lib/fli/FliImpl.h
@@ -56,13 +56,12 @@ class FliSignalObjHdl;
 class FliProcessCbHdl : public virtual GpiCbHdl {
   public:
     FliProcessCbHdl(GpiImplInterface *impl)
-        : GpiCbHdl(impl), m_proc_hdl(NULL), m_sensitised(false) {}
+        : GpiCbHdl(impl), m_proc_hdl(NULL) {}
 
     int cleanup_callback() override;
 
   protected:
     mtiProcessIdT m_proc_hdl;
-    bool m_sensitised;
 };
 
 // One class of callbacks uses mti_Sensitize to react to a signal
--- a/cocotb/share/lib/gpi/gpi_priv.h
+++ b/cocotb/share/lib/gpi/gpi_priv.h
@@ -47,7 +47,6 @@ typedef enum gpi_cb_state {
     GPI_FREE = 0,
     GPI_PRIMED = 1,
     GPI_CALL = 2,
-    GPI_REPRIME = 3,
     GPI_DELETE = 4,
 } gpi_cb_state_e;
 
--- a/cocotb/share/lib/vhpi/VhpiImpl.cpp
+++ b/cocotb/share/lib/vhpi/VhpiImpl.cpp
@@ -966,9 +966,13 @@ int VhpiImpl::deregister_callback(GpiCbHdl *gpi_hdl) {
 }
 
 void VhpiImpl::sim_end() {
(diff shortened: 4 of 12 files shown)


ktbarrett:
Seems to only affect Questa FLI, which is where I was experiencing it.

tests/test_cases/test_first_on_coincident_triggers/test_first_on_coincident_triggers.py
```diff
@@ -0,0 +1,29 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
imphil, line 4: Something went wrong here, and in all the other license headers.

tests/test_cases/test_first_on_coincident_triggers/test.vhd
```diff
@@ -0,0 +1,22 @@
...
+
+entity test is
+end entity test;
+
+architecture behav of test is
+    signal a : bit;
+    signal b : bit;
+begin
+	process is
+	begin
```
imphil, line 14: Can you remove the tabs?

tests/test_cases/test_first_on_coincident_triggers/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+PWD=$(shell pwd)
+
+ifeq ($(TOPLEVEL_LANG),verilog)
+    VERILOG_SOURCES := $(PWD)/test.sv
+else ifeq ($(TOPLEVEL_LANG),vhdl)
+    VHDL_SOURCES := $(PWD)/test.vhd
+    ifneq ($(filter $(SIM),ius xcelium),)
+        SIM_ARGS += -v93
+    endif
```
imphil, line 17: Which simulators is this condition targeting?

cocotb/share/lib/gpi/gpi_priv.h
```diff
@@ -47,7 +47,7 @@ typedef enum gpi_cb_state {
     GPI_FREE = 0,
     GPI_PRIMED = 1,
     GPI_CALL = 2,
-    GPI_REPRIME = 3,
+    // GPI_REPRIME = 3,  // deprecated
```
imphil, line 50: Just remove it.

cocotb/share/lib/vhpi/VhpiImpl.cpp
```diff
@@ -966,9 +966,13 @@ int VhpiImpl::deregister_callback(GpiCbHdl *gpi_hdl) {
 }
 
 void VhpiImpl::sim_end() {
-    sim_finish_cb->set_call_state(GPI_DELETE);
-    vhpi_control(vhpiFinish, vhpiDiagTimeLoc);
-    check_vhpi_error();
+    // Some sims do not seem to be able to deregister the end of sim callback
```
imphil, line 969: Can you add an issue reference here to specify "some sims"? Otherwise we'll never know if we can remove the workaround.

cocotb/share/lib/vhpi/VhpiImpl.cpp
```diff
@@ -966,9 +966,13 @@ int VhpiImpl::deregister_callback(GpiCbHdl *gpi_hdl) {
...
+    // so we need to make sure we have tracked this and not call the handler.
+    if (GPI_DELETE != sim_finish_cb->get_call_state()) {
```
imphil, line 971: Nit: I'm not a fan of yoda-style assignments, and would turn the conditions around to make them easier to read for humans.

cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -1057,7 +1058,19 @@ FliTimedCbHdl *FliTimerCache::get_timer(uint64_t time) {
     return hdl;
 }
 
-void FliTimerCache::put_timer(FliTimedCbHdl *hdl) { free_list.push(hdl); }
+#ifndef FLI_TIMER_CACHE_SIZE
+#define FLI_TIMER_CACHE_SIZE 256
+#endif
```
imphil, line 1063: Can we make this a variable instead of a define? And can you add a comment on its sizing? (something vague as "Arbitrary size, shown to be sufficient" would do the trick, unless you've thought about it more)

cocotb/share/lib/fli/FliImpl.h
```diff
@@ -343,6 +343,11 @@ class FliStringObjHdl : public FliValueObjHdl {
     char *m_mti_buff = nullptr;
 };
 
+/** Maintains a cache of FliTimedCbHdl objects which can be reused.
+ *
+ * Apparently allocating and freeing Timer callback objects is very expensive
```
imphil, line 347: Too bad we don't know why it was added :( Shows even more why commit messages are important (https://github.com/cocotb/cocotb/commit/ff68e85d3863c492a2189785731716f4bb257d6f)

imphil (approved):
LGTM, thanks!

cocotb/share/lib/vhpi/VhpiImpl.cpp
ktbarrett: This came from the VPI. There is no listed rationale. Related commits: https://github.com/cocotb/cocotb/commit/8c435b0d3b7e855897a7c72e9b0aa8d005ac89e6 and https://github.com/cocotb/cocotb/commit/70dc1a6a39ad9d3991e5d20d9e7105bf5f59b4db.

cocotb/share/lib/vhpi/VhpiImpl.cpp
ktbarrett:
```suggestion
    if (sim_finish_cb->get_call_state() != GPI_DELETE) {
```

tests/test_cases/test_first_on_coincident_triggers/Makefile
ktbarrett: IUS and Xcelium. This same condition is seen in several other Makefiles.

cocotb/share/lib/fli/FliImpl.cpp
ktbarrett, line 1063:
```suggestion
static constexpr size_t FLI_TIMER_CACHE_SIZE = 256;  // Arbitrary large value, it's doubtful more than 256 simultaneous Timer triggers will be active at any time
```

tests/test_cases/test_first_on_coincident_triggers/Makefile
cmarqu: Yeah, the Xcelium default is still VHDL'87.

tests/test_cases/test_first_on_coincident_triggers/test.sv
```diff
@@ -0,0 +1,22 @@
+//-----------------------------------------------------------------------------
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause
+//-----------------------------------------------------------------------------
+
+`timescale 1 ps / 1 ps
```
cmarqu, line 7: Might want to use the more SV'y
```
timeunit 1ps;
timeprecision 1ps;
```

cmarqu (approved):

cmarqu:
Worth a newsfragment?

ktbarrett:
@cmarqu I added a newsfragment.

Is there any way to retry the RTD build from their site? Otherwise I'll just do the close/reopen trick later.

cmarqu:
They have an admin interface where this is likely possible, but can't try right now.

tests/test_cases/test_first_on_coincident_triggers/test.sv
```diff
@@ -0,0 +1,23 @@
+//-----------------------------------------------------------------------------
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause
+//-----------------------------------------------------------------------------
+
+timeunit 1ps;
+timeprecision 1ps;
```
ktbarrett, line 8: These time unit args conflict with the ones being passed on the command line and it causes compilation failure. `COCOTB_HDL_TIMEUNIT` and `COCOTB_HDL_TIMEPRECISION` are applied no matter what which means if people are using the makefiles they either can't use these directives or the directives and the makefile variables must match. IMO I think these variables were a bad idea.

cmarqu: The option these are applied to for Xcelium at least are for setting a timescale to fall back to when there is no other timescale in a file.

Is that not the case for other simulators?

cmarqu:
The main anaconda channel appears to be down still, I just tried a rebuild at https://readthedocs.org/projects/cocotb/builds/17853615/ (when logged in, there is a small link for that near the top right).

tests/test_cases/test_first_on_coincident_triggers/test.sv
ktbarrett: No, Questa is failing because the two don't match. The file doesn't override the command line option.

imphil:
@ktbarrett Do you think you'll get this one done in the coming days for the 1.7 release?

ktbarrett:
@imphil It's good to go. The only failing test is the Windows nox issue. If no one has anymore comments I'll merge.

imphil:
Go ahead, you got enough approving reviews :-)

PR #3011, opened 2022-06-26
Add support for assertion rewriting regardless of test module name

ktbarrett:
Adds support for assertion rewriting to occur in any test module listed in MODULE. This will not perform assertion rewriting in other modules.


cocotb/regression.py
```diff
@@ -226,6 +228,32 @@ def _discover_tests() -> Iterable[Test]:
...
+            try:
+                # Install the assertion rewriting hook, which must be done before we
+                # import the test modules.
+                from _pytest.assertion import install_importhook
+                from _pytest.config import Config
+
+                test_modules_str = " ".join(f"{name}.py" for name in test_modules)
+                pytest_conf = Config.fromdictargs(
+                    {}, ["--capture=no", "-o", f"python_files={test_modules_str!r}"]
+                )
```
eric-wieser, line 249: I would guess that this works just as well as
```suggestion
                pytest_conf = Config.fromdictargs(
                    dict(python_files=["{name}.py" for name in test_modules],
                    ["--capture=no"]
                )
```
Does it?

cocotb/regression.py
```diff
@@ -226,6 +228,32 @@ def _discover_tests() -> Iterable[Test]:
...
+    @classmethod
+    def _setup_pytest_assertion_rewriting(cls, test_modules: Iterable[str]) -> None:
+        try:
+            import pytest
+        except ImportError:
+            _logger.info(
+                "pytest not found, install it to enable better AssertionError messages"
+            )
+        else:
+            try:
```
eric-wieser, line 240:
```suggestion
        except ImportError:
            _logger.info(
                "pytest not found, install it to enable better AssertionError messages"
            )
            return
        try:
```

tests/test_cases/test_cocotb/Makefile
```diff
@@ -19,7 +19,7 @@ MODULE := "\
 	test_async_generators,\
 	test_handle,\
 	test_logging,\
-	test_pytest,\
+	pytest_assertion_rewriting,\
```
eric-wieser, line 22: Can you add a comment in this file explaining why it is deliberately named in a different way to all the others?

cocotb/regression.py
ktbarrett: Unfortunately it does not. https://github.com/pytest-dev/pytest/discussions/10052#discussioncomment-3026738

cocotb/regression.py
```diff
@@ -226,6 +228,32 @@ def _discover_tests() -> Iterable[Test]:
...
+                "pytest not found, install it to enable better AssertionError messages"
+            )
+            return
+        try:
+            # Install the assertion rewriting hook, which must be done before we
+            # import the test modules.
+            from _pytest.assertion import install_importhook
+            from _pytest.config import Config
+
+            test_modules_str = " ".join(f"{name}.py" for name in test_modules)
```
eric-wieser, line 246: A link to the GitHub discussion in a comment would be great here

ktbarrett: Added.

tests/test_cases/test_cocotb/Makefile
ktbarrett: Added the comment above, it doesn't quite work on this line.

ktbarrett:
Need to change the cocotb runner tests, that's why tests are failing.

imphil (approved):
LGTM, thanks!

PR #3021, opened 2022-07-04
Support Questa with the VHPI interface

imphil:
Fix all remaining issues to make the cocotb test suite pass with Questa/VHPI. Tested on a pre-release version of an upcoming Questa release (2022_weekly_220527).

This PR contains a mix of fixes to our test suite, which somehow got unnoticed so far, and adjustments in our test expectations to support Questa VHPI.

I'm not quite sure yet why we didn't catch the various deprecation and future warnings before with Aldec VHPI that I now fixed.

### DIFF
--- a/tests/test_cases/test_array/test_array.py
+++ b/tests/test_cases/test_array/test_array.py
@@ -139,7 +139,7 @@ async def test_read_write(dut):
         dut.sig_int.value = 5000
         dut.sig_real.value = 22.54
         dut.sig_char.value = ord("Z")
-        dut.sig_str.value = "Testing"
+        dut.sig_str.value = b"Testing"
         dut.sig_rec.a.value = 1
         dut.sig_rec.b[0].value = 0x01
         dut.sig_rec.b[1].value = 0x23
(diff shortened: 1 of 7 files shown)


tests/test_cases/test_array/test_array.py
```diff
@@ -363,6 +363,8 @@ async def test_discover_all(dut):
         pass_total = 571
     elif cocotb.SIM_NAME.lower().startswith("ghdl"):
         pass_total = 56
+    elif cocotb.LANGUAGE in ["vhdl"] and cocotb.SIM_NAME.lower().startswith("modelsim"):
```
ktbarrett, line 366: How do we differentiate between FLI and VHPI here?

ktbarrett: I see we use `VHDL_GPI_INTERFACE` below.

imphil: You're pointing in the right direction. After some more edits I forgot to run against the fli interface as well; I'll update this PR to avoid breaking those tests.

PR #3023, opened 2022-07-07
Removal of COCOTB_UNUSED 

### LINKED ISSUE #2313: Cleanup COCOTB_UNUSED
In C++ instead of using the syntax `(void)unused_param`, it is customary to simply leave the parameter unnamed. Now that all libraries are C++, we can do this universally and deprecate `COCOTB_UNUSED`.

tushar-wiz:
Removes the use of the macro COCOTB_UNUSED().
Closes #2313 .


ktbarrett (approved):
Looks good. Thanks @tushar-wiz!

PR #3031, opened 2022-07-22
Use 'X as default LogicArray value

mrv96:
In System Verilog when a logic array is declared but not initialized, its value is X. This proposal aims to mimic such behavior.

### DIFF
--- a/cocotb/types/logic_array.py
+++ b/cocotb/types/logic_array.py
@@ -28,6 +28,8 @@ class LogicArray(Array[Logic]):
     used.
     Like :class:`Array`, if no *range* argument is given, it is deduced from the length
     of the iterable or bit string used to initialize the variable.
+    If a *range* argument is given, but no value,
+    the array is filled with the default value of Logic().
 
     .. code-block:: python3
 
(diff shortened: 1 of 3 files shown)


ktbarrett (commented):
Can you add a test/example to the docstring?

ktbarrett:
Also looks like the code doesn't pass the linting checks. You should install and use pre-commit.
```sh
python -m pip install pre-commit
pre-commit install
```

This will prevent you from creating a commit if it would fail linting.

cocotb/types/logic_array.py
```diff
@@ -117,9 +117,15 @@ class LogicArray(Array[Logic]):
...
     def __init__(
         self,
-        value: typing.Union[int, typing.Iterable[LogicConstructibleT], BinaryValue],
+        value: typing.Optional[typing.Union[int, typing.Iterable[LogicConstructibleT], BinaryValue]] = None,
         range: typing.Optional[Range] = None,
     ) -> None:
+        if value is None:
+            if range is None:
+                raise ValueError(
+                    "at least one between value and range input parameters must not be None"
```
cmarqu, line 126:
```suggestion
                    "at least one of value and range input parameters must not be None"
```

cocotb/types/logic_array.py
```diff
@@ -37,12 +39,15 @@ class LogicArray(Array[Logic]):
...
 
-        >>> LogicArray(0xA)                     # picks smallest range that can fit the value
+        >>> LogicArray(0xA)  # picks smallest range that can fit the value
         LogicArray('1010', Range(3, 'downto', 0))
 
-        >>> LogicArray(-4, Range(0, "to", 3))   # will sign-extend
+        >>> LogicArray(-4, Range(0, "to", 3))  # will sign-extend
         LogicArray('1100', Range(0, 'to', 3))
 
+        >>> LogicArray(range=Range(0, "to", 3))  # defaults values
```
imphil, line 48:
```suggestion
        >>> LogicArray(range=Range(0, "to", 3))  # default values
```

cocotb/types/logic_array.py
```diff
@@ -115,12 +120,38 @@ class LogicArray(Array[Logic]):
 
     __slots__ = ()
 
+    @typing.overload
     def __init__(
         self,
         value: typing.Union[int, typing.Iterable[LogicConstructibleT], BinaryValue],
+        range: typing.Optional[Range],
```
imphil, line 127: nit for a future cleanup: It's rather common to import the used types from typing to use e.g. "Optional" instead of "typing.Optional". The type annotations are already long enough, prefixing them with "typing." isn't helping.

imphil (approved):
LGTM, please squash with a proper commit message before merging.

PR #3033, opened 2022-07-24
Use new Github theme-dependent image syntax

### LINKED ISSUE #2915: README dark / light images result in duplicate and/or broken logos on PyPI
We should see if we can get absolute URLs to work, but will also need to strip the dark mode images before publication; either via the github action mentioned at https://www.npmjs.com/package/strip-gh-theme-links, or doing something in setup.py.

Alternatively, we could just accept an unreadable logo on github dark mode and just remove the light mode image.

Milestoning as 1.7 so we don't forget before a release.

ktbarrett:
Closes #2915. Uses the new syntax described [here](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to). Check my fork to see it work. You might need to refresh the page a couple times to get the logo to change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
-![cocotb logo](/../../../../cocotb/cocotb-web/blob/master/assets/img/cocotb-logo.svg#gh-light-mode-only)
-![cocotb logo](/../../../../cocotb/cocotb-web/blob/master/assets/img/cocotb-logo-dark.svg#gh-dark-mode-only)
+<picture>
+  <source media="(prefers-color-scheme: dark)" srcset="https://www.cocotb.org/assets/img/cocotb-logo-dark.svg">
+  <source media="(prefers-color-scheme: light)" srcset="https://www.cocotb.org/assets/img/cocotb-logo.svg">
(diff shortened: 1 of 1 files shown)


imphil (approved):

eric-wieser:
I suspect this also misbehaves on PyPI, which has no dark theme.

eric-wieser:
But worst case scenario we get a poor-contrast image on PyPI for system dark mode users, which we can always try and fix in a future release (or submit a patch to PyPI...)

ktbarrett:
@eric-wieser I wouldn't expect the first 2 lines to work on PyPI, but the 3rd line should, which is why the light theme logo is used.

eric-wieser:
My concern is that the first two lines might end up working, since the syntax works in standard html (and I don't know the details of PyPI's markdown sanitization); but the easy thing to do is just try it and see.

ktbarrett:
@imphil Would it be possible to try this on the testing pypi?

eric-wieser:
I'm happy with just merging and finding out when we release, rather than spending lots of effort on trying the test pypi

PR #3048, opened 2022-08-12
CI: Switch to GitHub Actions runners for proprietary tools

imphil:
We now have self-hosted runners available which also provide access to
proprietary tools. Use those, and remove the existing setup which used
Azure Pipelines for these tests.

- Add GitHub Actions configuration for proprietary tool tests.
- Remove the Azure Pipelines trigger.
- Remove the codecov workaround to deal with two CI pipelines.
- In a separate commit: Fix a small issue in the noxfile for Aldec Riviera-PRO.


imphil:
It seems like we pick self-hosted runners at times even though we don't want to, see https://github.com/cocotb/cocotb/runs/8024112389?check_suite_focus=true

Also: we need to add ~/.local/bin to PATH by default to mimic the GH-hosted runners.

.github/workflows/regression-tests.yml
```diff
@@ -200,6 +200,11 @@ jobs:
       continue-on-error: ${{matrix.may_fail || false}}
       timeout-minutes: 15
       run: |
+        export
+        if [ ${{matrix.self-hosted}} == "true" ]; then
+          module load "${{ matrix.sim }}"
```
ktbarrett, line 205: Usually `module` is an alias or bash function. Is the script where they are defined sourced? Alternatively, you can call `modulecmd` directly.

imphil:
This PR is now ready for review. The changes on the cocotb side are minimal, the main work is (as expected) in the infrastructure that provides the self-hosted runners.

Runners are clean (ephemeral) AWS EC2 instances which are spawned on demand, and terminated after each build. Only Linux runners are available. Job spawning takes a little longer on our self-hosted runners than it does for the GitHub-provided ones (~ 1 minute), but after that all simulation jobs are run in parallel, reducing the overall CI latency to ~ 15 minutes (down from 2-3 hours in our current serialized setup).

We continue to be unable to show tool output for Questa and Riviera-PRO, but we're now showing everything else that is printed during the build and test process, which should make it at least a bit easier to debug test failures with these simulators.

.github/workflows/regression-tests.yml
```diff
@@ -200,6 +200,9 @@ jobs:
       continue-on-error: ${{matrix.may_fail || false}}
       timeout-minutes: 15
       run: |
+        if [ "${{matrix.self-hosted}}" == "true" ]; then
+          module load "${{ matrix.sim-version }}"
```
cmarqu, line 204: Are the Modulefiles themselves somewhere we could access?

cmarqu (approved):

.github/workflows/regression-tests.yml
imphil: No. (But then, they are not fancy either. Effectively just a PATH modification and potentially a small set of further environment variables needed for the tool.)

PR #3053, opened 2022-08-24
Use venv as the default nox backend in Windows regression

ktbarrett:
Fixes current CI break where `nox` now complains when using `session.install` when there is no virtual environment backend. I expect the builtin module `venv` will have better support on Windows than `virtualenv`. Will try out `conda` as a backend if that fails.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -170,7 +170,7 @@ jobs:
       # Virtual environments don't work on Windows with cocotb. Avoid using them
       # in nox testing.
       run: |
-        nox --no-venv -k "${{ inputs.nox_session_test_sim }} and ${{ matrix.sim }} and ${{ matrix.lang }}"
+        nox -k "${{ inputs.nox_session_test_sim }} and ${{ matrix.sim }} and ${{ matrix.lang }}"
 
       # Ubuntu / MacOS Testing
     - name: Install cocotb build dependencies (Ubuntu - g++)
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -170,7 +170,7 @@ jobs:
       # Virtual environments don't work on Windows with cocotb. Avoid using them
       # in nox testing.
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -170,7 +170,7 @@ jobs:
       # Virtual environments don't work on Windows with cocotb. Avoid using them
       # in nox testing.
       run: |
-        nox --no-venv -k "${{ inputs.nox_session_test_sim }} and ${{ matrix.sim }} and ${{ matrix.lang }}"
+        nox -db venv -k "${{ inputs.nox_session_test_sim }} and ${{ matrix.sim }} and ${{ matrix.lang }}"
```
ktbarrett, line 173:
```suggestion
        nox -db conda -k "${{ inputs.nox_session_test_sim }} and ${{ matrix.sim }} and ${{ matrix.lang }}"
```

ktbarrett:
Looks like none of the backends work. I'll do the recommended suggestion and do `session.run("pip", "install", ...)` instead specifically for Windows (What an OS! Props Microsoft!).

imphil:
Yeah, I came to the same conclusion -- nothing works apart from not using a venv (of any sort). The alternative would be pinning nox to an older version, which isn't viable in the long term.

I wouldn't bother making the code OS-specific, simply do session.run() in any case.

.github/workflows/regression-tests.yml
```diff
@@ -170,7 +170,7 @@ jobs:
       # Virtual environments don't work on Windows with cocotb. Avoid using them
       # in nox testing.
       run: |
-        nox --no-venv -k "${{ inputs.nox_session_test_sim }} and ${{ matrix.sim }} and ${{ matrix.lang }}"
+        nox -k "${{ inputs.nox_session_test_sim }} and ${{ matrix.sim }} and ${{ matrix.lang }}"
```
ktbarrett, line 173:
```suggestion
        nox --no-venv -k "${{ inputs.nox_session_test_sim }} and ${{ matrix.sim }} and ${{ matrix.lang }}"
```

imphil (approved):
Beautiful (TM). Thanks!

PR #3057, opened 2022-08-25
Remove support for macOS 10.15

imphil:
GitHub actions will EOL macOS 10.15 support at the end of August.
Officially remove support for cocotb as well, since we cannot test it.

Add macOS 12 instead, where we got some successful users reports
already.


imphil:
I intend to cherry-pick this change into the 1.7 branch as well, since we cannot test macOS 10 from September on.

.github/generate-envs.py
```diff
@@ -124,7 +124,7 @@
         "lang": "verilog",
         "sim": "icarus",
         "sim-version": "homebrew-stable",
-        "os": "macos-latest",
+        "os": "macos-11",
```
imphil, line 127: These two are equivalent as of today, but I'd like to get our CI a bit more predictable by removing as many "variable versions" as possible.

cmarqu (approved):

PR #3058, opened 2022-08-26
Remove logo from README

imphil:
The README gets rendered by a variety of Markdown renderers on GitHub,
on PyPi, and in various other aggregation sites. Finding a common syntax
for pictures, including one which switches between light and dark
themes, is (close to) impossible. Let's simplify things by not having a
logo in our README.


imphil:
PyPi cannot render the current syntax: https://test.pypi.org/project/cocotb/1.8.0.dev0/

image

themperek:
Keeping just one would not work?

imphil:
> Keeping just one would not work?

Then we don't get (apparently) good enough rendering in dark mode, which was the reason we iterated through different solutions already (https://github.com/cocotb/cocotb/pull/3033)

ktbarrett (approved):

PR #3065, opened 2022-09-02
Fix test failure on same trigger

### LINKED ISSUE #1347: Two coroutines that fail on the same trigger are not handled gracefully by the scheduler.

ktbarrett:
Closes #1347.

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -265,6 +265,7 @@ class Scheduler:
         self._pending_triggers = []
         self._pending_threads = []
         self._pending_events = []  # Events we need to call set on once we've unwound
+        self._scheduling = []
 
         self._terminate = False
         self._test = None
@@ -437,7 +438,7 @@ class Scheduler:
                 # Scheduled coroutines may append to our waiting list so the first
                 # thing to do is pop all entries waiting on this trigger.
                 try:
-                    scheduling = self._trigger2coros.pop(trigger)
+                    self._scheduling = self._trigger2coros.pop(trigger)
                 except KeyError:
                     # GPI triggers should only be ever pending if there is an
                     # associated coroutine waiting on that trigger, otherwise it would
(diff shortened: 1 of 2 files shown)


imphil:
Thanks @ktbarrett for looking into this long-standing issue!

@eric-wieser would you mind having a look?

Since 1.7.0 has settled nicely I'm going the cautious route and not picking this up for 1.7.0 (scheduler changes are too much risk for me at this point). I'm happy, however, to do a 1.7.1 in two weeks or so with this change included.

cocotb/scheduler.py
```diff
@@ -1072,6 +1074,12 @@ def _cleanup(self):
                     self.log.debug("Killing %s" % str(coro))
                 coro.kill()
 
+        # if there are coroutines being scheduled when the test ends, kill them (gh-1347)
```
eric-wieser, line 1077: It's not immediately clear to me that `_cleanup` should be allowed to run while in the middle of the `_event_loop`.

ktbarrett: Intuitively, you are right. Unfortunately, it doesn't work that way. Maybe one day, after a significant refactor, the scheduler will be a little more intuitive. I've already tried a few times to do this without a lot of success.

If a task fails it is `_unschedule`d. If there are no waiters on it, then it `_abort_test`s. The test is then `_unschedule`d which results in a call to `_cleanup`.

cmarqu: At some point we should probably have some PlantUML state diagrams or something like that in the docs.

marlonjames (approved):
This is the most straight-forward fix given the current scheduler design.

PR #3068, opened 2022-09-04
Don't overwrite coverage metrics in dev_test

### LINKED ISSUE #3026: Missing coverage from `dev_test_nosim` nox session
1. `dev_test_nosim` is run first. It runs `dev_coverage_combine` at the end of the session.
2. Later, `dev_test_sim` runs, followed by `dev_coverage_combine` again.
3. `dev_coverage_combine` runs `coverage combine`, which unless the `--append` option is used, will overwrite the existing `.coverage` file.

This results in losing the coverage from the `dev_test_nosim` session.

ktbarrett:
Closes #3026.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -248,7 +248,7 @@ def dev_coverage_combine(session: nox.Session) -> None:
     session.run("pip", "install", *coverage_report_deps)
 
     coverage_files = glob.glob("**/.coverage.test.*", recursive=True)
-    session.run("coverage", "combine", *coverage_files)
+    session.run("coverage", "combine", "--append", *coverage_files)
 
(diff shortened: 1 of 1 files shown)


imphil (commented):
I think we need a different approach here; we don't want to simply append to any "stray" `.coverage` file (e.g. from a previous run). 

The trivial approach might be removing the notify calls from the dev_* sessions, add it explicitly to the dev_test session, and call the dev_coverage_combine session explicitly in CI.

ktbarrett:
I think maybe we can decompose the sessions into steps like you mentioned and provide "flow" sessions that simply call the appropriate stages.

ktbarrett:
Superceded by #3112.

PR #3090, opened 2022-09-21
Build and smoke-test the sdist in CI

### LINKED ISSUE #3084: CI: Test building of the sdist
We currently produce a sdist, but don't test if it actually builds. The dev tests are installed directly from source, and the release tests are installed from wheels.

We probably don't want a full testing on the sdists, but a simple build test should be added to ensure #3071 doesn't repeat.

imphil:
Prevent packaging errors from slipping through which make the sdist unusable. No full testing is done on the sdist, we continue to do that on the wheels (otherwise we'd have the same CI pipeline twice, with very limited benefit).

Fixes #3084

### DIFF
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -43,6 +43,24 @@ jobs:
             dist/*.whl
             dist/*.tar.gz
 
+  test_sdist:
+    name: Test the source distribution (sdist)
+    needs: build_release
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
+      with:
+        python-version: "3.10"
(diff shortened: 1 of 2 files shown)


imphil:
Action run showing an (intentionally triggered) failure: https://github.com/cocotb/cocotb/actions/runs/3100416305/jobs/5027859588

imphil:
And a passing run: https://github.com/cocotb/cocotb/actions/runs/3103972632/jobs/5029335507

I'm removing the debug commits now.

marlonjames (approved):
`release_clean` nox session is just for local development? I don't see it used anywhere in CI

imphil:
Yeah, we don't need the clean session in CI since we run on fresh machines every time. But I needed it locally :)

PR #3092, opened 2022-09-21
Support Python 3.11

imphil:
Python 3.11 is in its release candidate stage, start supporting it -- they promise not to break backwards-compat any more, even for wheels.


cmarqu:
Does this need an update of https://github.com/cocotb/cocotb/blob/1ca0dfff36c04bc657ef87645362292a377c8f68/.codecov.yml#L20 ?

imphil:
@cmarqu good catch. I think it's slightly wrong right now anyways, but let's at least not make the problem worse. Updated.

cmarqu (approved):

Issue #3097, opened 2022-09-28
RHEL8/Python 3.8: cocotb 1.7.1 wheel install fails with "ModuleNotFoundError: No module named 'pygpi'"


imphil:
I've uninstalled cocotb, changed Python to 3.9 and installed cocotb again
```
pip3 uninstall cocotb
dnf module install python39
alternatives --config python3
pip3 install cocotb --user --verbose
```
Wheel install:
```
Collecting cocotb
  Created temporary directory: /tmp/pip-unpack-otqwhxbk
  Looking up "https://files.pythonhosted.org/packages/77/90/5a819a2eaa92fbdd53e5031b86ef49d610cb8283cea2efa5e22864ce7b14/cocotb-1.7.0-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl" in the cache
  No cache entry available
  Starting new HTTPS connection (1): files.pythonhosted.org:443
  https://files.pythonhosted.org:443 "GET /packages/77/90/5a819a2eaa92fbdd53e5031b86ef49d610cb8283cea2efa5e22864ce7b14/cocotb-1.7.0-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl HTTP/1.1" 200 3695683
  Downloading cocotb-1.7.0-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (3.7 MB)
     |████████████████████████████████| 3.7 MB 6.4 MB/s eta 0:00:01  Ignoring unknown cache-control directive: immutable
  Updating cache with response from "https://files.pythonhosted.org/packages/77/90/5a819a2eaa92fbdd53e5031b86ef49d610cb8283cea2efa5e22864ce7b14/cocotb-1.7.0-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl"
  Caching due to etag
     |████████████████████████████████| 3.7 MB 6.4 MB/s
  Added cocotb from https://files.pythonhosted.org/packages/77/90/5a819a2eaa92fbdd53e5031b86ef49d610cb8283cea2efa5e22864ce7b14/cocotb-1.7.0-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl#sha256=6eded67546534bb51dd7b2f95ea0e2f7ac90579a7fad72a170fc796d05ac775a to build tracker '/tmp/pip-req-tracker-1rsv85er'
[...]

PR #3100, opened 2022-10-06
Depend on find_libpython as external package

imphil:
Previously, find_libpyton, a library to (surprisingly) find libpython
matching a given 'python' executable, was vendored into cocotb. That was
done to keep the "no external dependencies" that cocotb traditionally
had.

This commit removes the vendored-in version of find_libpython, and
depends on it instead. The functionality provided by cocotb remains the
same.

The change from vendoring to a dependency is done for two reasons:

* Vendoring requires copying sources from another place, and then
  patching them to fix the imports. There are tools to help with that
  (e.g. the `vendoring` package), but in the end it's a "messy" process
  that requires custom scripting that needs to be kept up-to-date.
* The logic to find libpython is based on heuristics, which need to be
  regularly updated when we encounter another interesting way to "hide"
  libpython in the wild. Being able to do such updates without doing a
  new cocotb release as well helps our users (since we can respond
  faster), and helps cocotb by avoiding the work to do a fresh release.


imphil:
See also #3097 for context why an update of find_libpython is necessary.

Note that this change per se does not solve #3097; we'll also need to do a find_libpython release. @ktbarrett can you do that? I've tested both a vendored version (with one patch to fix the imports), as well as the vendored version we're having here.

I attempted to do the vendoring as well, and it's not overly painful. So if there's a strong feeling in the room that we should continue to use the vendoring approach I'm happy to do a PR with that as well.

ktbarrett:
@imphil I had planned on doing an [0.3 release](https://github.com/ktbarrett/find_libpython/milestone/1) soon. I can move the issues there to a 0.4 release if you need a release by this weekend (I'll be on vacation).

ktbarrett (approved):

imphil:
Let's give this a couple days on master, I'll then do a backport to stable/1.7.

PR #3102, opened 2022-10-07
Remove an unused pytest mark

imphil:
The 'compile' pytest mark is not registered, leading to warning emitted
by pytest. We don't actually seem to use this marker, so simply remove
it.

```
tests/pytest/test_parallel_cocotb.py:24
  /home/az_devops_agent/agent/_work/1/s/tests/pytest/test_parallel_cocotb.py:24: PytestUnknownMarkWarning: Unknown pytest.mark.compile - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.compile
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,7 @@ testpaths =
 # log_cli_level = DEBUG
 markers =
     simulator_required: mark tests as needing a simulator
+    compile: the compile step in runner-based tests
 
 [coverage:run]
 omit =


imphil:
It might be that someone used this marker in manual runs and it's still valuable -- let me know if that's the case!

themperek (approved):
Until this is not solved https://github.com/pytest-dev/pytest-xdist/issues/385.

The nicest/easy way I found to compile and then test is to do something like this:

```cmd
pytest -m compile test_parallel.py
pytest -m 'not compile' -n 2 test_parallel.py
```

But it can be removed for now (add later with an example)

imphil:
Oh, in that case I'm happy to leave the marker in and get rid of the warning instead. I'll update the PR.

imphil:
Updated to declare the marker instead. @themperek would you mind having another look?

themperek (approved):

PR #3103, opened 2022-10-08
Python runner improvements and documentation 

themperek:
- Based on https://github.com/cocotb/cocotb/pull/2897
- Removed CLI for the moment (get stable runner first)
- Make private methods private


documentation/source/runner.rst
```diff
@@ -0,0 +1,80 @@
...
+.. versionadded:: 1.7
+
+.. warning::
+    Python runners and associated APIs are an experimental feature and subject to change.
+
+
+The Python Test Runner (short: "runner") described here is a replacement
+for cocotb's traditional Makefile flow.
+It builds the HDL for the simulator and runs the tests.
+
```
cmarqu, line 16: Maybe we should also say here what it is not: a tool that will be able to handle any "strange situation". (Point to "Custom Flows" for such purposes.)

documentation/source/runner.rst
```diff
@@ -0,0 +1,83 @@
...
+.. note::
+    Take a look at the
+    :ref:`list of command line flags <pytest:command-line-flags>`
+    in pytest's official documentation.
+
+Direct usage
+=============
+
+You can also run the test directly.
```
cmarqu, line 67:
```suggestion
You can also run the test directly, that is, without using pytest, like so
```

documentation/source/runner.rst
```diff
@@ -0,0 +1,83 @@
...
+
+.. code-block:: bash
+
+    python examples/simple_dff/test_dff.py
+
+For this you need to define the test to run in the script. For example:
```
cmarqu, line 73:
```suggestion
This requires that you define the test to run in the testcase file itself.
For example, add the following code at the end:
```

documentation/source/runner.rst
```diff
@@ -0,0 +1,83 @@
...
+.. code-block:: bash
+
+    SIM=questa HDL_TOPLEVEL_LANG=vhdl pytest examples/simple_dff/test_dff.py
+
+Note that the environment variables ``SIM`` and ``HDL_TOPLEVEL_LANG``
+are defined in this test file to set arguments to the runner's
+:meth:`~cocotb.runner.build` and :meth:`~cocotb.runner.test`  functions;
+they are not directly handled by the runner itself.
+
+FIXME: mention testcase naming scheme ``foo_test``
```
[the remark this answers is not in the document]
themperek: Maybe link here: https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#test-discovery

documentation/source/runner.rst
```diff
@@ -45,7 +45,9 @@ are defined in this test file to set arguments to the runner's
 :meth:`~cocotb.runner.build` and :meth:`~cocotb.runner.test` functions;
 they are not directly handled by the runner itself.
 
-FIXME: mention testcase naming scheme ``foo_test``
+Test filenames and functions has to follow
+`pytest discovery <https://docs.pytest.org/explanation/goodpractices.html#test-discovery>`_
+convention in order to be automatically find out.
```
cmarqu, line 50:
```suggestion
Test filenames and functions have to follow the
`pytest discovery <https://docs.pytest.org/explanation/goodpractices.html#test-discovery>`_
convention in order to be automatically found.
```

cmarqu (approved):

cocotb/runner.py
```diff
@@ -73,187 +111,236 @@ def set_env(self) -> None:
...
 
         for cmd in cmds:
-            print(
-                "INFO: Running command: "
-                + ' "'.join(cmd)
-                + '" in directory:"'
-                + str(cwd)
-                + '"'
-            )
+            print(f"INFO: Running command {' '.join(cmd)} in directory {cwd}")
```
imphil, line 343: To get the shell quoting right, define a polyfill for shlex.join(), e.g. 

```
def shlex_join(split_command):
    """Return a shell-escaped string from *split_command*."""
    return ' '.join(quote(arg) for arg in split_command)
```

(From https://github.com/python/cpython/blob/main/Lib/shlex.py)

and then
```suggestion
            print(f"INFO: Running command {shlex_join(cmd)} in directory {cwd}")
```

imphil (approved):

cocotb/runner.py
themperek: One need also add.
```
_find_unsafe = re.compile(r'[^\w@%+=:,./-]', re.ASCII).search

def quote(s):
    """Return a shell-escaped version of the string *s*."""
    if not s:
        return "''"
    if _find_unsafe(s) is None:
        return s

    # use single quotes, and put single quotes into double quotes
    # the string $'b is then quoted as '$'"'"'b'
    return "'" + s.replace("'", "'\"'\"'") + "'"
```

Should I add this?

themperek:
If no one objects will merge tomorrow.

cocotb/runner.py
tianrui-wei: May I ask why shlex.quote/shlex.join is not used?

themperek: @tianrui-wei 
>May I ask why shlex.quote/shlex.join is not used?

If is supported in python 3.6 than probably no good reason. Can you propose a PR?

tianrui-wei: @themperek PR at https://github.com/cocotb/cocotb/pull/3150

PR #3105, opened 2022-10-12
GPI: Mark various string references as const

imphil:
Looks like we didn't apply const too generously to std::string references. Let's do that to increase performance (a bit) and make it clearer who's modifying what.


ktbarrett (approved):
Nothing controversial about adding `const`.

marlonjames (approved):

PR #3107, opened 2022-10-12
Enable the use of the PyPi index for release tests

imphil:
cocotb installs binary wheels without using the package index to ensure
we test against the version of cocotb we just built.
Now that cocotb has an external dependency we need to get that from
PyPi, otherwise the installation fails.

pip seems to have no way unfortunately to indicate "install cocotb from
the local wheels (guaranteed!) and everything else from PyPi". To work
around that we first install the dependencies from PyPi explicitly, and
then install cocotb itself.

This setup isn't pretty and duplicates the dependencies into the
noxfile, but at least it's robust.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -373,13 +373,24 @@ def release_test_sdist(session: nox.Session) -> None:
 def release_install(session: nox.Session) -> None:
     """Helper: Install cocotb from wheels and also install test dependencies."""
 
+    # We have to disable the use of the PyPi index when installing cocotb to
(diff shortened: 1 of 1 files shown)


imphil:
Updated to change the approach, looks like pip isn't providing us an easy solution here :( (insert typical rant about Python packaging here)

imphil:
The release CI run looks good, I'll remove that commit again.

ktbarrett (approved):

PR #3112, opened 2022-10-14
Fix coverage reporting

imphil:
Implement the ideas discussed in https://github.com/cocotb/cocotb/pull/3068. See the individual commits for details.


marlonjames:
Unfortunately it appears that all files that match the `coverage.*` glob are removed at the start of the pytest run if we don't use `--cov-append` when invoking pytest.

Here is the function where it happens:
https://github.com/nedbat/coveragepy/blob/64130ee5cebd07b1ca4b3acbb0a9b387ebbd7b7f/coverage/sqldata.py#L761

marlonjames (approved):

PR #3128, opened 2022-11-15
Fix incorrect string splitting for Xcelium

tianrui-wei:
Currently, the way the cmd list is passed into the execute could make shell wrongly quote parts of the command, triggering E,BDOPTQ error message for Xcelium

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>

### DIFF
--- a/cocotb/runner.py
+++ b/cocotb/runner.py
@@ -844,9 +844,11 @@ class Xcelium(Simulator):
         self.env["CDS_AUTO_64BIT"] = "all"
         cmd = [
             ["xrun"]
-            + ["-logfile xrun_build.log"]
+            + ["-logfile"]
+            + ["xrun_build.log"]
             + ["-elaborate"]
-            + [f"-xmlibdirname {self.build_dir}/xrun_snapshot"]
+            + ["-xmlibdirname"]
+            + [f"{self.build_dir}/xrun_snapshot"]
             + ["-licqueue"]
             # TODO: way to switch to these verbose messages?:
             + ["-messages"]
@@ -857,11 +859,8 @@ class Xcelium(Simulator):
             # + ["-plierr_verbose"]  # Expand handle info in PLI/VPI/VHPI messages
             # + ["-vpicompat 1800v2005"]  # <1364v1995|1364v2001|1364v2005|1800v2005> Specify the IEEE VPI
             + ["-access +rwc"]
-            + [
-                "-loadvpi "
-                + cocotb.config.lib_name_path("vpi", "xcelium")
-                + ":vlog_startup_routines_bootstrap"
-            ]
+            + ["-loadvpi"]
+            + [cocotb.config.lib_name_path("vpi", "xcelium") + ":vlog_startup_routines_bootstrap"]
             + [f"-work {self.library_name}"]
             + self.compile_args
             + ["-define COCOTB_SIM"]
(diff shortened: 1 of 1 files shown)


tianrui-wei:
The existing test_dff should serve as a testcase if there's a Xcelium CI machine running btw

imphil (commented):
Thanks for your pull request, @tianrui-wei. Much appreciated.

Can you do the same change slightly further down in the ` '-input "@probe -create {self.sim_toplevel} -all -depth all"'`  line. And then we probably need to split self.sim_args and self.plus_args as well using shlex.

Unfortunately we don't have Xcelium running in CI, so I'd appreciate if you can give this change a good testing locally.

cocotb/runner.py
```diff
@@ -883,10 +882,12 @@ def test_command(self) -> List[Command]:
...
-            + [f"-logfile xrun_{self.current_test_name}.log"]
-            + [
-                f"-xmlibdirname {self.build_dir}/xrun_snapshot -cds_implicit_tmpdir {tmpdir}"
-            ]
+            + ["-logfile"]
+            + [f"xrun_{self.current_test_name}.log"]
+            + ["-xmlibdirname"]
+            + [f"{self.build_dir}/xrun_snapshot"]
+            + ["-cds_implicit_tmpdir"]
+            + [f"{tmpdir}"]
```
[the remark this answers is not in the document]
tianrui-wei: Thanks, addressed in bc408e83397bb39af398d0610b738bfca206e8fe

cocotb/runner.py
```diff
@@ -857,11 +859,8 @@ def build_command(self) -> List[Command]:
             # + ["-plierr_verbose"]  # Expand handle info in PLI/VPI/VHPI messages
             # + ["-vpicompat 1800v2005"]  # <1364v1995|1364v2001|1364v2005|1800v2005> Specify the IEEE VPI
             + ["-access +rwc"]
```
imphil, line 861:
```suggestion
            + ["-access"]
            + ["+rwc"]
```

cocotb/runner.py
```diff
@@ -857,11 +859,8 @@ def build_command(self) -> List[Command]:
...
-            + [
-                "-loadvpi "
-                + cocotb.config.lib_name_path("vpi", "xcelium")
-                + ":vlog_startup_routines_bootstrap"
-            ]
+            + ["-loadvpi"]
+            + [cocotb.config.lib_name_path("vpi", "xcelium") + ":vlog_startup_routines_bootstrap"]
             + [f"-work {self.library_name}"]
             + self.compile_args
             + ["-define COCOTB_SIM"]
```
imphil, line 866:
```suggestion
            + ["-define"] 
            + ["COCOTB_SIM"]
```

imphil (commented):
> Could you perhaps calrify on what you means by shlex? Applying them to self.sim_args seems to cause errors.

I'm sorry, I confused you there. I was under the impression `self.sim_args` was a string with potentially multiple command-line arguments. That's not the case: `self.sim_args` is a list. That means: nothing needs splitting, and the code is correct as-is. Ignore my comment on using `shlex.split()`

I however found two more cases where we need to split the command-line arguments and left an inline comment. Can you have a look at those two cases as well? 

With that addressed I think we're good to go here.

tianrui-wei:
Hi Philipp,

Thanks for the suggestion, I've made those style changes in the new commit, as well as testing them locally. Thanks for reviewing the PR.

Best,
Tianrui

imphil (approved):
Thanks for your updates and testing, @tianrui-wei!

cocotb/runner.py
```diff
@@ -856,15 +858,17 @@ def build_command(self) -> List[Command]:
...
                 + ":vlog_startup_routines_bootstrap"
             ]
             + [f"-work {self.library_name}"]
             + self.compile_args
-            + ["-define COCOTB_SIM"]
+            + ["-define"]
+            + ["COCOTB_SIM"]
             + self.get_define_options(self.defines)
             + self.get_include_options(self.includes)
             + self.get_parameter_options(self.parameters)
```
eric-wieser, line 874: Looks to me like these three functions are also wrong.

imphil: Sorry @eric-wieser, I didn't see your comment before merging. I opened https://github.com/cocotb/cocotb/issues/3134 to track, @tianrui-wei if you have time, would you mind having a look?

tianrui-wei: Hi @imphil , @eric-wieser ,

Thanks for pointing them out. I'll conduct some more testing and fix the issue should I find one.

Best,
Tianrui

PR #3136, opened 2022-11-18
Fixing testcase reproducibility issue.

jahagirdar:
__qualname__ resolves to b'/home/vijayvithal/prj/switch/test/env1_test/cctb__run_hardware_test__fast___001', 

Where `'/home/vijayvithal/prj/` is the local folder where the design repo is checked out.
While running regressions on a remote server(github/jenkins/bitbucket etc.) the folder where the repo is checkout will be different.
This results in the same RANDOM_SEED computing to different test seed on different machines resulting in issues with reproducibility of test failures

This fix considers only the basename of __qualname__ which is same across machines e.g b'cctb__run_hardware_test__fast___001

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -505,7 +505,9 @@ class RegressionManager:
         else:
             test_pass, sim_failed = self._score_test(test, outcome)
             if not test_pass:
-                self.xunit.add_failure()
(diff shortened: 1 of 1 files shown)


jahagirdar:
Closing PR because commit history is not clean.

PR #3137, opened 2022-11-18
Removed path from qualname to eliminate test reproducibility issues

jahagirdar:
qualname contains the full path to the test folder

e.g.
`/home/vijayvithal/repos/switch/tests/env1/...` on my local machine
`home/runner/switch/tests/env1/...` on github Actions
`/var/lib/jenkins/workspace/switch_main/tests/env1/...` on jenkins

When the full qualname is used to generate the testseed it results in issues with reproducing testfailure found during regressions.

This PR uses only the basename of the qualname which will remain the same across all machines.

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -376,7 +376,7 @@ class RegressionManager:
 
         # seed random number generator based on test module, name, and RANDOM_SEED
         hasher = hashlib.sha1()
-        hasher.update(test.__qualname__.encode())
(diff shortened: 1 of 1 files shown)


ktbarrett:
This is not an issue with cocotb, but with how you are specifying your tests to run. `__qualname__` does not contain the full path to the directory. Add the root of your Python test directory structure within your project to PYTHONPATH which should be the same no matter where it is run and the `__qualname__` will be the same.

I vote to close.

PR #3150, opened 2022-11-27
feat: use shlex instead of manually implementing it

tianrui-wei:
To maintain backwards compatibility, we only implement a shim for shlex.join, which isn't available until python 3.8

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>

### DIFF
--- a/cocotb/runner.py
+++ b/cocotb/runner.py
@@ -12,6 +12,7 @@
 import abc
 import os
 import re
+import shlex
 import shutil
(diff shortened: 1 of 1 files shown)


themperek (approved):

themperek:
Thank you @tianrui-wei

tianrui-wei:
@themperek Thanks for merging the PR, my pleasure :)

PR #3151, opened 2022-11-28
fix: try to fix windows ci

tianrui-wei:

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -91,7 +91,7 @@ jobs:
     - name: Install Python testing dependencies
       run: |
         conda install --yes sqlite=3.39
-        pip install nox==2022.8.7
+        pip install nox==2022.8.7 virtualenv==20.16.7
     - name: Run tests without simulators
       run: nox -s "${{ inputs.nox_session_test_nosim }}"
       continue-on-error: ${{matrix.may_fail || false}}
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -87,7 +87,7 @@ jobs:
         auto-update-conda: true
         python-version: ${{matrix.python-version}}
         auto-activate-base: true
-        activate-environment: ""
+        activate-environment: base
 
       # Run tests that don't need a simulator.
     - name: Install Python testing dependencies
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -85,8 +85,10 @@ jobs:
       uses: conda-incubator/setup-miniconda@v2
       with:
         auto-update-conda: true
-        python-version: ${{matrix.python-version}}
-        
+        # python-version: ${{matrix.python-version}}
+        auto-activate-base: true
+        activate-environment: ""
+
       # Run tests that don't need a simulator.
     - name: Install Python testing dependencies
       run: |
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -87,7 +87,6 @@ jobs:
         auto-update-conda: true
         python-version: ${{matrix.python-version}}
         auto-activate-base: true
-        activate-environment: base
 
       # Run tests that don't need a simulator.
     - name: Install Python testing dependencies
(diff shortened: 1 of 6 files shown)


tianrui-wei:
Note that it's current still failing for reasons beyond my understanding. Feel free to take the PR from here.

.github/workflows/regression-tests.yml
```diff
@@ -165,7 +165,7 @@ jobs:
       run: conda install --yes -c msys2 m2-base m2-make m2w64-toolchain libpython
     - name: Install cocotb runtime dependencies (Windows, msvc)
       if: startsWith(matrix.os, 'windows') && matrix.toolchain == 'msvc'
-      run: conda install --yes -c msys2 m2-base m2-make
+      run: conda install --yes -c msys2 m2-base m2-make sqlite
```
themperek, line 168: Maybe this?

```suggestion
      run: conda install --yes -c msys2 m2-base m2-make sqlite=3.39
```

tianrui-wei: Appreciate the suggestion, it's still broken

themperek:
Looks like it is still instaling `sqlite-3.40.0`
https://github.com/cocotb/cocotb/actions/runs/3569923878/jobs/6000387537#step:5:436

tianrui-wei:
I suspect this is a upstream windows github runner issue. Let's wait for some time to see whether if someone else will report the problem to upstream

themperek:
@tianrui-wei I have tried few other things on you branch. I hope you do not mind?

themperek:
And obviously messed up the comment. Sorry.

tianrui-wei:
Not at all, be my guest to push directly to this branch

.github/workflows/regression-tests.yml
```diff
@@ -90,7 +90,8 @@ jobs:
       # Run tests that don't need a simulator.
     - name: Install Python testing dependencies
       run: |
-        pip install nox
+        conda install --yes sqlite=3.39
```
ktbarrett, line 93: Is this specific version necessary? If so, add a comment as to why.

.github/workflows/regression-tests.yml
```diff
@@ -90,7 +90,8 @@ jobs:
...
+        pip install nox==2022.8.7 virtualenv==20.16.7
```
ktbarrett, line 94: Are these pins necessary if sqlite is being installed? Pinning to nox 2022.8.7 means that 3.6 test will fail. Switch instead to `nox<2022.9` or similar.

cocotb/__init__.py
```diff
@@ -254,7 +254,7 @@ def _initialise_testbench_(argv_):
     # sys.path normally includes "" (the current directory), but does not appear to when python is embedded.
     # Add it back because users expect to be able to import files in their test directory.
     # TODO: move this to gpi_embed.cpp
-    sys.path.insert(0, "")
+    sys.path.append("")
```
ktbarrett, line 257: I'm positive this is not a valid change. We want local modules to be resolved before external ones. If there are collisions because of this, we need to put the power in the user's hand to solve. A better solution is to not do this at all. While "users expect" things, all of a sudden tests not working and the user not immediately knowing why or having the power to change that behavior are bad.

.github/workflows/regression-tests.yml
themperek: Just previous version used. But it is still failing. I have no idea.

cocotb/__init__.py
tianrui-wei: The intention for this change is to smoke out issues that may be related to PATH, as I suspected prepending things to PATH may be part of the issue. If the CI gets fixed I can revert this change.

.github/workflows/regression-tests.yml
tianrui-wei: As I replied to Tomasz, I suspect it's not related to conda, but a github image issue. I suppose we could wait a bit to see if there's a bug reported upstream. Alternatively, could you re-run some previous ci runs that passed?

themperek:
@tianrui-wei If one uses _base_ coda environment it seems to work 🤷🏻‍♂️ 

Would you like to clean this PR? If not I make a new one.

imphil:
@themperek since you're at it, feel free to take over this PR and update it to be merge-ready. A comment why the base version was chosen would be helpful, and remove all unrelated changes.

tianrui-wei:
@themperek Feel free to make a new one :) It's quite impressive that you've managed to make the ci work.

themperek:
Closing in favour of #3167

Issue #3166, opened 2022-12-29
Error importing _sqlite3 DLL on Windows (Conda) 


themperek:
CI fails on Windows/Conda : https://github.com/cocotb/cocotb/actions/runs/3788709749/jobs/6441765333

```
  File "D:\a\cocotb\cocotb\.nox\dev_test_nosim\lib\site-packages\pytest_cov\plugin.py", line 6, in <module>
    import coverage
  File "D:\a\cocotb\cocotb\.nox\dev_test_nosim\lib\site-packages\coverage\__init__.py", line 15, in <module>
    from coverage.control import Coverage, process_startup
  File "D:\a\cocotb\cocotb\.nox\dev_test_nosim\lib\site-packages\coverage\control.py", line 23, in <module>
    from coverage.data import CoverageData, combine_parallel_data
  File "D:\a\cocotb\cocotb\.nox\dev_test_nosim\lib\site-packages\coverage\data.py", line 19, in <module>
    from coverage.sqldata import CoverageData
  File "D:\a\cocotb\cocotb\.nox\dev_test_nosim\lib\site-packages\coverage\sqldata.py", line 16, in <module>
    import sqlite3
  File "C:\Miniconda\envs\test\lib\sqlite3\__init__.py", line 23, in <module>
    from sqlite3.dbapi2 import *
  File "C:\Miniconda\envs\test\lib\sqlite3\dbapi2.py", line 27, in <module>
    from _sqlite3 import *
ImportError: DLL load failed while importing _sqlite3: The specified module could not be found.
nox > Command pytest -v --cov=cocotb --cov-branch --cov-report= -k 'not simulator_required' failed with exit code 1
```

Seems same issues as : #1909

DLL is not properly installed/configured when using conda environment.

PR #3167, opened 2022-12-29
Fix conda missing sqlite3 dll in CI #3166

themperek:
Use _base_ conda environment.

Closes: #3166

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -85,7 +85,11 @@ jobs:
       uses: conda-incubator/setup-miniconda@v2
       with:
         auto-update-conda: true
-        python-version: ${{matrix.python-version}}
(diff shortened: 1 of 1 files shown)


imphil (approved):
Lgtm. Do you have any (part of an) explanation why this approach works?

themperek:
> Lgtm. Do you have any (part of an) explanation why this approach works?

`DLL is not properly installed/configured when using conda environment.` #3166

Why is this the case I do not know. Same was true in the past #1909

ktbarrett:
Now that you mention it I remember having to do the same thing for [`find_libpython`'s CI](https://github.com/ktbarrett/find_libpython/blob/e6f9cb4547265e9aa01147dfcc66bcce8ec2762a/.github/workflows/tests.yml#L88). Doesn't appear I wrote down why I discovered it was necessary.

Issue #3168, opened 2023-01-01
parameter with large number does not get handled correctly


catkira:
I have a parameter with a rather long number (because iverilog does not support multidimensional parameters, so I sequeeze multiple numbers into one).
The parameter reaches the hdl code correctly, but something goes wrong on the python side.

The parameter is this number:
`
1043161809724755117294630463171082084379269244686810311029607490900248788420381392348463362674117467351380280716396061994807856922595934162415011370737945306097471762891582355257266129874607407022695562401368170834331191570587525271668692425035152628033539633035671272822809848481969293316809956131050719687308142627180187719859186358020489682827020566615291426464441948895237962125901086180674120729539754342713718442926328439298090536281793869892530559822533699520885460200043828645675500690787021624293704335548279396031745742539130949779391845360041558468656930060946865658437656234172368262891111818743752245284837962427971856480068009631783627803463593158425393489613888769529055843669738695165161575481544351020262523014509861557762884136920067057263401584677476858151676420644240180037273464619824125783108857328600765813075760694118749230749032679295972977002489474784749875908710432315440784621321055372136598051978044337017931726632721237632407266511479457338798037948587694505331102427565179203351786911664035714952736332056963449894161364778566454808486588047682353303007806935225584948894648374367526802728090180560360178179648464611823616681910331503501853221338831442171912737474561170439869566397269273475997939859449
`

However in my Python model the parameter becomes `65529`

You can see the details in the test run of my workflow at https://github.com/catkira/5G_SSB_sync/actions/runs/3816910503/jobs/6492835861
Just unfold "Verify with cocotb and icarus"

The parameter gets assembled here: https://github.com/catkira/5G_SSB_sync/blob/45874427129b440ed794299b9668608070e83452/tests/test_PSS_correlator.py#LL123-L125C128
And printed inside the model here:
https://github.com/catkira/5G_SSB_sync/blob/45874427129b440ed794299b9668608070e83452/model/PSS_correlator.py#LL5-L8C48

You can see all the versions that I am using the the CI run.

catkira:
I checked a bit more where the problem lies.

`PSS_LOCAL `is of type `ModifiableObject`

The value function looks like this
```python
    @NonConstantObject.value.getter
    def value(self) -> BinaryValue:
        binstr = self._handle.get_signal_val_binstr()
        # Skip BinaryValue.assign() as we know we are using a binstr
        result = BinaryValue(n_bits=len(binstr))
[...]

PR #3169, opened 2023-01-01
Allow parameters/variables with values larger than long

catkira:
This fixes an issue where parameters/variables with long numbers get truncated, because Python's int type can contain numbers much larger than a C++ long.

A more elaborate discussion of the problem can be found in [this](https://github.com/cocotb/cocotb/issues/3168) issue.


catkira:
There seems to be a problem in some CI runs, but I cannot access the log files. Can somebody post the relevant parts of the log files here?

imphil:
CI fail:

```

2023-01-01T19:55:27.5086799Z #      1.00ns INFO     cocotb.regression                  running check_objects (2/3)
2023-01-01T19:55:27.5088280Z #                                                           Check the types of objects that are returned
2023-01-01T19:55:27.5089836Z #      1.00ns INFO     cocotb.test                        inst_axi4s_buffer(GPI_MODULE) is HierarchyObject
2023-01-01T19:55:27.5091655Z #      1.00ns INFO     cocotb.test                        gen_branch_distance(0)(GPI_MODULE) is HierarchyObject
2023-01-01T19:55:27.5093269Z #      1.00ns INFO     cocotb.test                        inst_branch_distance(GPI_MODULE) is HierarchyObject
2023-01-01T19:55:27.5095065Z #      1.00ns INFO     cocotb.test                        inbranch_tdata_low(GPI_REGISTER) is ModifiableObject
2023-01-01T19:55:27.5096801Z #      1.00ns INFO     cocotb.test                        inbranch_tdata_low(0)(GPI_REGISTER) is ModifiableObject
2023-01-01T19:55:27.5102331Z #      1.00ns INFO     cocotb.test                        aclk(GPI_REGISTER) is ModifiableObject
2023-01-01T19:55:27.5103563Z #      1.00ns INFO     cocotb.test                        s_axis_input_tdata(GPI_REGISTER) is ModifiableObject
2023-01-01T19:55:27.5104932Z #      1.00ns INFO     cocotb.test                        current_active(GPI_INTEGER) is IntegerObject
2023-01-01T19:55:27.5115130Z #      1.00ns ERROR    gpi                                VHPI: Size of m_binvalue.value.str was not large enough: req=33 have=2 for type vhpiIntVal
2023-01-01T19:55:27.5117567Z #      2.00ns INFO     cocotb.regression                  check_objects failed
2023-01-01T19:55:27.5119408Z #                                                         Traceback (most recent call last):
2023-01-01T19:55:27.5121685Z #                                                           File "/home/az_devops_agent/agent/_work/1/s/tests/test_cases/test_vhdl_access/test_vhdl_access.py", line 83, in check_objects
2023-01-01T19:55:27.5124377Z #                                                             fails += check_instance(dut.inst_axi4s_buffer.DATA_WIDTH, ConstantObject)
2023-01-01T19:55:27.5128579Z #                                                           File "/home/az_devops_agent/agent/_work/1/s/.nox/dev_test/lib/python3.6/site-packages/cocotb/handle.py", line 363, in __getattr__
2023-01-01T19:55:27.5130523Z #                                                             handle = self.__get_sub_handle_by_name(name)
2023-01-01T19:55:27.5134071Z #                                                           File "/home/az_devops_agent/agent/_work/1/s/.nox/dev_test/lib/python3.6/site-packages/cocotb/handle.py", line 321, in __get_sub_handle_by_name
2023-01-01T19:55:27.5137547Z #                                                             sub_handle = SimHandle(new_handle, self._child_path(name))
2023-01-01T19:55:27.5142898Z #                                                           File "/home/az_devops_agent/agent/_work/1/s/.nox/dev_test/lib/python3.6/site-packages/cocotb/handle.py", line 1166, in SimHandle
2023-01-01T19:55:27.5145930Z #                                                             obj = ConstantObject(handle, path, t)
2023-01-01T19:55:27.5149235Z #                                                           File "/home/az_devops_agent/agent/_work/1/s/.nox/dev_test/lib/python3.6/site-packages/cocotb/handle.py", line 570, in __init__
2023-01-01T19:55:27.5151237Z #                                                             self._value = int(self._handle.get_signal_val_binstr(), 2)
2023-01-01T19:55:27.5153471Z #                                                         ValueError: invalid literal for int() with base 2: ''
2023-01-01T19:55:27.5155068Z #      2.00ns INFO     cocotb.regression                  running port_not_hierarchy (3/3)
2023-01-01T19:55:27.5157255Z #                                                           Test for issue raised by Luke - iteration causes a toplevel port type to
2023-01-01T19:55:27.5158796Z #                                                           change from ModifiableObject to HierarchyObject
2023-01-01T19:55:27.5160304Z #      2.00ns INFO     cocotb.test                        dut.aclk is ModifiableObject
2023-01-01T19:55:27.5161865Z #      2.00ns INFO     cocotb.test                        dut.aclk is ModifiableObject
2023-01-01T19:55:27.5163311Z #      3.00ns INFO     cocotb.regression                  port_not_hierarchy passed
2023-01-01T19:55:27.5165082Z #      3.00ns INFO     cocotb.regression                  *********************************************************************************************
2023-01-01T19:55:27.5167074Z #                                                         ** TEST                                 STATUS  SIM TIME (ns)  REAL TIME (s)  RATIO (ns/s) **
2023-01-01T19:55:27.5168781Z #                                                         *********************************************************************************************
2023-01-01T19:55:27.5170482Z #                                                         ** test_vhdl_access.check_enum_object    PASS           1.00           0.00        272.43  **
2023-01-01T19:55:27.5172072Z #                                                         ** test_vhdl_access.check_objects        FAIL           1.00           0.00        282.56  **
2023-01-01T19:55:27.5173636Z #                                                         ** test_vhdl_access.port_not_hierarchy   PASS           1.00           0.01         73.40  **
2023-01-01T19:55:27.5175429Z #                                                         *********************************************************************************************
2023-01-01T19:55:27.5177290Z #                                                         ** TESTS=3 PASS=2 FAIL=1 SKIP=0                         3.00           0.38          7.82  **
2023-01-01T19:55:27.5179082Z #                                                         *********************************************************************************************
```

catkira:
I think this is the relevant error
```
2023-01-01T19:55:27.5115130Z #      1.00ns ERROR    gpi                                VHPI: Size of m_binvalue.value.str was not large enough: req=33 have=2 for type vhpiIntVal
```
Is that maybe another bug somewhere else? binstr should always work for integer type variables right?

catkira:
I am afraid I cant fix the Questa/RivieraPro problem, because I dont have licenses for these applications.
I will just leave this PR open. Maybe somebody from the team will have a chance to look at it sometime.

catkira:
It looks like I was able to solve the problem 'blindly' by adding a catch-try statement and just executing the old code in case of an exception. I think the reason why this works is that some Simulators don't support binstr properly. There is a comment in the code (see [here](https://github.com/cocotb/cocotb/blob/5f40366c448a1828d40fe6d07919788bbb71fbdd/cocotb/share/lib/vhpi/VhpiCbHdl.cpp#L748)) that also points in that direction.

I don't understand why codecov complaints, I am pretty sure that all my added lines are executed during tests. So I think its a false alarm.

catkira:
I am trying to get an educational or time limited RivieraPRO license so that I can reproduce this problem locally...

catkira:
@stuarthodgson Do You have an idea why RivieraPRO hangs when I try to read a binstr?

catkira:
I cant figure out why the RivieraPRO-VHDL does not finish. CI pipelines should not include proprietary software that other developers cannot test ^^

catkira:
CIs should pass now. I enabled long integers only for iverilog by checking `simulator.get_simulator_product()`, for all other simulators everything stays the same as before.

catkira:
finally all green, I also added a test.

marlonjames:
I don't believe this is the right approach.
There does appear to be a problem with `bufSize` for `vhpiBinStrVal` values for scalars such as integers. That can be fixed.

But isn't your example equivalent to an `UNSIGNED` generic in VHDL rather than an integer? Implementations may restrict the maximum size for integer types in VHDL.

We should look at the classification of VPI constants as `GPI_INTEGER` and determine if there is a better way to handle `logic`/`reg` types that may be larger than a 64-bit integer.

catkira:
Yes it would be best do distinguish between 'integer' and `logic` or `bit` parameter/signals.
Introducing a new type like `GPI_BITS` could help to prevent the extra overhead of converting 'normal' integers to bitstrings and back. I can look into it as far as VPI is concerned. For VHPI and FLI it should be done by somebody who has a license for Questa/RivieraPRO.

catkira:
I have added a GPI_BITS type for variables/constants which are not size constrained integers. A VPI example is already working.

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -155,6 +155,7 @@ static gpi_objtype_t const_type_to_gpi_objtype(int32_t const_type) {
     switch (const_type) {
         case vpiDecConst:
         case vpiBinaryConst:
+            return GPI_BITS;
         case vpiOctConst:
         case vpiHexConst:
```
eric-wieser, line 160: Surely you want to include these too?

catkira: Yes I thought so too, but wasn't sure because I could not think of a Verilog construct that generates any of these. Do You know how to construct an Oct or Hex constant in verilog? I guess constant means it must a `parameter` or `localparam`, but how to make them 'octal' or 'hex'?

catkira: `localparam HEX_PARAM = 16'hABCD;` does not become `vpiHexConst` and that would be my only idea how to generate one.

eric-wieser: I would guess that parameters (that are not of `integer` type) initialized with `8’hA` end up of type Hex, in some simulators?

catkira:
All green now. Should I squeeze the commits into one?

catkira:
I also tried to make something comparable work with GHDL, but it turned out that GHDL throws an error if the value given to an integer GENERIC is too large
```
Execution of /usr/bin/ghdl-mcode terminated by unhandled exception
raised CONSTRAINT_ERROR : grt-values.adb:173 overflow check failed
```
I think its not so critical to have it in VHDL, because one can just use array generics instead of super large parameters like shown [here](https://stackoverflow.com/questions/11791484/how-to-specify-an-integer-array-as-generic-in-vhdl).

In iverilog array parameters are not yet implemented, therefore parameters with large numbers are the only clean way to get many coefficients ie for a FIR filter into a module. (one could also do it via a data file and read them via `$readmemh()` like I did [here](https://github.com/catkira/DDS/blob/ec24162c51a78a0bb200047d978f6b5d85968568/hdl/dds.sv#LL50-L57C11)).

imphil:
Thanks for your work on that, @catkira. I'm trying to catch up here, so bear with my potentially already answered questions. I'm looking at Verilog only for the moment, leaving the VHDL part to @marlonjames and others:

In your test, you're assigning a number to an untyped Verilog parameter (`parameter INT_PARAM = 12,` as defined in https://github.com/cocotb/cocotb/blob/master/tests/designs/sample_module/sample_module.sv#L41). That means the simulator has to determine the type, which in this case should be a logic vector ("A parameter declaration with no type or range specification shall default to the type and range of the final value assigned to the parameter, after any value overrides have been applied. If the expression is real, the parameter is real. If the expression is integral, the parameter is a logic vector of the same size with range [size-1:0].", Section 6.20.2 Value parameters of the SV-2012 LRM). The right data type for a logic vector in GPI is `GPI_REGISTER`. Have you tried using that data type? Or in other words: Why do you need to introduce GPI_BITS?

While you're at it, I'd love to see more elaborate testing:

* Assign a negative value to the parameter on the command line.
* Assign a hex or octal value to the parameter on the command line.
* Let's do the same for real numbers well.
* What happens if you use a typed parameter, instead of an untyped one?

Having an untyped parameter that gets its type from a command-line assignment at runtime looks like something that simulators might struggle with, so the more testing we have here, the better.

marlonjames:
> The right data type for a logic vector in GPI is `GPI_REGISTER`. Have you tried using that data type? Or in other words: Why do you need to introduce `GPI_BITS`?

This was my thought as well.

catkira:
I added a testcase for a large negative number `-10^100`. It seems like iverilog cannot handle it. The assert in the test_case gives this error:
```
  File "/mnt/d/git/cocotb/tests/test_cases/test_long_parameter/test_long_parameter.py", line 11, in test_long_paramter
                                                            assert dut.INT_PARAM_2.value == -10**100
                                                        AssertionError: assert 249960115965281907...1576210832368861184 == -(10 ** 100)
                                                         +  where 249960115965281907...1576210832368861184 = ConstantObject(sample_module.INT_PARAM_2).value
                                                         +    where ConstantObject(sample_module.INT_PARAM_2) = HierarchyObject(sample_module).INT_PARAM_2
```
It looks like iverilog returns a 2s-complement number as binstr in case of a negative number. But its difficult or not possible to convert it properly to decimal because one does not know whether it is a 2s-complement number.
Does anybody have an idea how to solve this? Or should we say large negative numbers are not supported?

catkira:
> > The right data type for a logic vector in GPI is `GPI_REGISTER`. Have you tried using that data type? Or in other words: Why do you need to introduce `GPI_BITS`?
> 
> This was my thought as well.

I tried to do that, but then this error appears:
```
vvp error: get 18 not supported by vpiBinaryConst
vvp: vpi_const.cc:367: virtual int __vpiBinaryConst::vpi_get(int): Assertion `0' failed.
make[1]: *** [/mnt/d/git/cocotb/cocotb/share/makefiles/simulators/Makefile.icarus:76: results.xml] Aborted
make[1]: Leaving directory '/mnt/d/git/cocotb/tests/test_cases/test_long_parameter'
make: *** [/mnt/d/git/cocotb/cocotb/share/makefiles/Makefile.inc:40: sim] Error 2
```

imphil:
> It looks like iverilog returns a 2s-complement number as binstr in case of a negative number. But its difficult or not possible to convert it properly to decimal because one does not know whether it is a 2s-complement number.

Well, an untyped parameter is a logic vector, according to the SV standard. So the parameter should be available in cocotb as BinaryValue(), and it's then up to the user of that value to interpret it the way they want it interpreted, [using one of the properties of BinaryValue](https://docs.cocotb.org/en/stable/library_reference.html#cocotb.binary.BinaryValue).

What happens if you actually used a typed parameter, e.g. a "parameter int unsigned" or a "parameter int signed"?

catkira:
> > It looks like iverilog returns a 2s-complement number as binstr in case of a negative number. But its difficult or not possible to convert it properly to decimal because one does not know whether it is a 2s-complement number.
> 
> Well, an untyped parameter is a logic vector, according to the SV standard. So the parameter should be available in cocotb as BinaryValue(), and it's then up to the user of that value to interpret it the way they want it interpreted, [using one of the properties of BinaryValue](https://docs.cocotb.org/en/stable/library_reference.html#cocotb.binary.BinaryValue).
> 
> What happens if you actually used a typed parameter, e.g. a "parameter int unsigned" or a "parameter int signed"?

Can You describe a bit more detailed what changes in the code You propose?

Typed parameters like `parameter int unsigned` will have a size constraint and would not allow large numbers like 10^100.

marlonjames:
This needs a rebase after #3691 has been merged, and the recent handle changes may affect this as well.

ktbarrett:
@marlonjames Isn't this change superseded by the change to return `GPI_REGISTER` for all parameters? Getting those values is done via binstr and then turned into an integer by the user (rather than implicitly).

PR #3171, opened 2023-01-05
Xcelium Makefile: use filter instead of findstring to avoid false positives

jahagirdar:
findstring has false positives for inputs like `xyztimescale.v`
the change only matches timescale directive and not partial matches.. either " -timescale=..." or " -timescale ..."


jahagirdar:
[make.zip](https://github.com/cocotb/cocotb/files/10357389/make.zip)
The attached makefile illustrates the difference between the two methods.

cmarqu (approved):

PR #3180, opened 2023-01-11
CI: Switch to GitHub $GITHUB_OUTPUT

imphil:
set-output is deprecated and will be removed in June 2023. Prepare for that
by using the new method of doing things, which is, write the output to a
file $GITHUB_OUTPUT.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ 
explains it all.

### DIFF
--- a/.github/generate-envs.py
+++ b/.github/generate-envs.py
@@ -188,6 +188,11 @@ def main() -> int:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("--group")
     parser.add_argument("--output-format", choices=("gha", "json"), default="json")
+    parser.add_argument(
+        "--gha-output-file",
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test-release.yml
```diff
@@ -12,6 +12,10 @@ on:
       - "stable/**"
     tags:
       - 'v*'
+  pull_request:
```
cmarqu, line 15: Should amend the comment above (line 8).

Is this also very indirectly the reason for the coverage dropping to 0 for all C++ files? This is also part of the other PR you just opened where the coverage also drops.

imphil: Oh thats a second commit for testing only, I'll remove it before merging.

ktbarrett (approved):

PR #3191, opened 2023-01-21
add a workaround for test_sv_interface with SIM=verilator

forensicgarlic:
This simple workaround gets the test_sv_interface test to pass with Verilator.

### DIFF
--- a/documentation/source/newsfragments/3191.bugfix.rst
+++ b/documentation/source/newsfragments/3191.bugfix.rst
@@ -1,0 +1,1 @@
+Updated test_sv_interface with a workaround for Verilator so the test now passes with that simulator
--- a/tests/test_cases/test_sv_interface/top.sv
+++ b/tests/test_cases/test_sv_interface/top.sv
@@ -13,5 +13,7 @@ endinterface
 module top ();
 
 sv_if sv_if_i();
-
+`ifdef VERILATOR
+   logic d; // without this workaround for Verilator, cocotb gets a gpi error 'no root handle found'
+`endif
 endmodule


imphil:
Thanks for this PR, @forensicgarlic. Is there a Verilator issue open for this problem? I'd like to avoid working around this without getting some feedback from the Verilator devs first.

forensicgarlic:
That's fair. I can try to get a verilator test case written up, now that i believe baseline verilator can run with cocotb without any unaccepted vpi callback patches. 

I was coming from a place of wanting to get things working (even with workarounds) since sometimes getting the 'right' patch applied to the right repo is either harder, or a different skill set than finding some workarounds that can still let folks make progress.

ktbarrett:
I've seen this cause problems in past versions of Questa as well, not in this specific test, but in other "empty" designs.

documentation/source/newsfragments/3191.bugfix.rst
```diff
@@ -0,0 +1 @@
+Updated test_sv_interface with a workaround for Verilator so the test now passes with that simulator
```
ktbarrett, line 1: I'm not sure this is newsworthy.

ktbarrett:
Any word from the Verilator devs about this? I'm fine with the change if a comment is added as to why it was done.

forensicgarlic:
i'm still working on taking the test and getting it to run through the verilator `test_regress` perl test framework.

documentation/source/newsfragments/3191.bugfix.rst
forensicgarlic: makes sense. The comments in the Pull request dialog talking about steps necessary for a pull request to be accepted made it seem like newsfragments were mandatory.

ktbarrett: They aren't. That being said, thank you for the PR

forensicgarlic:
initial investigation with just verilator and c++ wrappers around the baseline top.sv file makes it look like verilator can discover the interface. I have to go and compare now what i have in my c++ wrapper against what cocotb's doing to see if there's something i'm doing that's enabling this discovery.

forensicgarlic:
The difference is vpi iteration. 

```
  vpiHandle vh1 = vpi_handle_by_name((PLI_BYTE8*)"TOP.top.sv_if_i.a", NULL);
```

works fine with or without the workaround code in this commit. 

But Cocotb's `__init__.py` looks like it runs something like 
```
  vpiHandle it = vpi_iterate(vpiModule, NULL);
```
which doesn't return an iterator without the workaround. 

This sounds related, though not identical to this similar issue reported on the Verilator project. https://github.com/verilator/verilator/issues/3919

Now that i've isolated it to what's different, this sounds like an issue on the Verilator side. While it does feel like a bug, it doesn't feel like it's directly related to this test case, can a simulator access system verilog interfaces.

forensicgarlic:
Wilson figured out the interface Scope issue pretty quick after pointing out the issue to him, so with the latest master of Verilator i confirmed that this test passes without this pull request. 

This request can be closed.

PR #3216, opened 2023-01-31
VHDL/GHDL move the make phase in build runner command

garzetti:
This PR give more separation to the build and test phase using cocotb runner with ghdl simulator.

Since there is no meaning in specifying the `hdl_toplevel` in VHDL additional libraries, when the runner `build()` method is called without the top entity to be analyzed only the `ghdl -i` command is invoked. Otherwise, when the `hdl_toplevel` is present, the `build()` method also add `ghdl -m` in commands.
It is now possible to invoke the `ghdl -m` and `ghdl -r` command using different arguments. 

This PR is related to the #3205

### DIFF
--- a/cocotb/runner.py
+++ b/cocotb/runner.py
@@ -656,18 +656,19 @@ class Ghdl(Simulator):
             + [str(source_file) for source_file in self.vhdl_sources]
         ]
 
+        if self.hdl_toplevel is not None:
+            cmds += [
+                ["ghdl", "-m"]
+                + [f"--work={self.hdl_library}"]
+                + self.build_args
+                + [self.hdl_toplevel]
+            ]
+
         return cmds
 
     def _test_command(self) -> List[Command]:
 
         cmds = [
-            ["ghdl", "-m"]
-            + [f"--work={self.hdl_toplevel_library}"]
-            + self.test_args
-            + [self.sim_hdl_toplevel]
-        ]
-
-        cmds += [
             ["ghdl", "-r"]
             + [f"--work={self.hdl_toplevel_library}"]
             + self.test_args
--- a/tests/pytest/test_vhdl_libraries_multiple.py
+++ b/tests/pytest/test_vhdl_libraries_multiple.py
@@ -40,13 +40,24 @@ def test_toplevel_library():
     if sim == "xcelium":
         compile_args = ["-v93"]
 
-    for lib in ["e", "d", "c", "b", "a"]:
+    # Build additional libraries
+    for lib in ["e", "d", "c", "b"]:
         runner.build(
             hdl_library=f"{lib}lib",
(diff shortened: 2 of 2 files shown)


cocotb/runner.py
```diff
@@ -656,18 +656,19 @@ def _build_command(self) -> List[Command]:
             + [str(source_file) for source_file in self.vhdl_sources]
         ]
 
+        if self.hdl_toplevel is not None:
+            cmds += [
+                ["ghdl", "-m"]
+                + [f"--work={self.hdl_library}"]
```
imphil, line 662: I know you only copied this code, but while you're at it, let's clean this up:

```suggestion
                + ["--work", self.hdl_library"]
```

imphil (approved):
Thanks for your pull request, @garzetti. LGTM from a high-level Point of view, @themperek please have a look as well.

>  It is now possible to invoke the ghdl -m and ghdl -r command using different arguments.

is this something we want a test for? (I'm OK if the answer is no.)

ktbarrett:
Isn't this just a revert of #3208? What makes #3208 invalid?

cocotb/runner.py
marlonjames: I don't think that's correct. It needs to be [`--work=<LIB_NAME>`](https://ghdl.github.io/ghdl/using/InvokingGHDL.html#cmdoption-ghdl-work).

imphil: Are you sure? Most command-line parsers accept the version without `=` as well, with the added benefit of getting automatic shell escaping for `self.hdl_library` as value (imagine it containing a `=` itself, which is unlikely here, of course).

marlonjames: It requires "--work=" as the prefix:
https://github.com/ghdl/ghdl/blob/a0321bd7ae81196b1b3a260224b0ee5102c32d61/src/options.adb#L187

imphil: Ugh. Looks like no argument parsing library was available :-) Ignore my comment then, sorry for the noise.

imphil:
@ktbarrett See the discussion at https://github.com/cocotb/cocotb/discussions/3205#discussioncomment-4827484

ktbarrett (approved):
Let's merge.

imphil:
Thanks for your contribution, @garzetti!

themperek:
Sorry for late comment. If I understand it this solution make the user code more complicated to only solve a problem for `ghdl` to give same arguments 2 times.  So not sure about it.

garzetti:
I do not have any preferences in keep this modification or revert it, both will work. Probably @themperek , that is more experienced than me in cocotb, is right, for this reasons I have opened the discussion before the PR. Maybe, at this point , the best solution to be chosen is the one which tries be similar with others runner simulators. If you think is better to revert I can prepare a PR, let me know 🙂

Issue #1478, opened 2020-03-04
cocotb verilator code coverage support


lavanyajagan:
Could you add verilator coverage support in the following file:

``cocotb/share/lib/verilator/verilator.cpp``

```
diff --git a/cocotb/share/lib/verilator/verilator.cpp b/cocotb/share/lib/verilator/verilator.cpp
index aff06d9..fc43374 100644
--- a/cocotb/share/lib/verilator/verilator.cpp
+++ b/cocotb/share/lib/verilator/verilator.cpp
@@ -84,5 +84,9 @@ int main(int argc, char** argv) {
     tfp->close();
 #endif
 
+#if VM_COVERAGE
+    VerilatedCov::write("coverage.dat");
+#endif
+
     return 0;
 }
```

imphil:
Thank you for your issue. Would you mind opening a pull request with the proposed change?

themperek:
We already have `COVERAGE` in Aldec but the same variable is used to collect python coverage so maybe we should use different variables.

imphil:
VM_COVERAGE is a Verilator define which is set if Verilator is instructed to collect coverage (when compiling the simulation).

This is different from a variable that users would use in a cocotb makefile to instruct Verilator to actually build in coverage (the equivalent to `COVERAGE` in Aldec)

lavanyajagan:
https://github.com/cocotb/cocotb/pull/1479 

Hope this is sufficient. Thanks

sammy17:
I noticed that even when I enable verilator code coverage with `EXTRA_ARGS= --coverage`, coverage.dat file is not getting generated even with the above fix (as mentioned in https://github.com/cocotb/cocotb/issues/2573#issuecomment-865454544). 

After a quick debug I noticed that anything after `vl_finish(__FILE__, __LINE__, "");` in `cocotb/share/lib/verilator/verilator.cpp` is not getting executed. So `VerilatedCov::write("coverage.dat");` that was added for this issue never got executed and `coverage.dat` file was not dumped.

I was able to get the coverage file to be dumped by moving the `VerilatedCov::write` call right before `vl_finish` as below.

```
        // If there are no more cbAfterDelay callbacks,
        // the next deadline is max value, so end the simulation now
        if (next_time == static_cast<vluint64_t>(~0ULL)) {
#if VM_COVERAGE
            VerilatedCov::write("coverage.dat");
#endif
            vl_finish(__FILE__, __LINE__, "");
            break;
        } else {
[...]

PR #3217, opened 2023-02-01
update verilator coverage to work if vl_finish ends the testrun

forensicgarlic:
update inspired by https://github.com/cocotb/cocotb/issues/1478#issuecomment-903345076 However, through trial and error I discovered that sometimes tests end with vl_finish, and other times not, so both write("coverage.dat") lines are needed in verilator.cpp.

### DIFF
--- a/documentation/source/newsfragments/1478.bugfix.rst
+++ b/documentation/source/newsfragments/1478.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed Verilator not writing coverage files in some cases
+Fixed Verilator not writing coverage files in some cases.
--- a/documentation/source/newsfragments/1478.bugfix.rst
+++ b/documentation/source/newsfragments/1478.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed Verilator not writing coverage files in some cases
--- a/cocotb/share/lib/verilator/verilator.cpp
+++ b/cocotb/share/lib/verilator/verilator.cpp
@@ -115,7 +115,6 @@ int main(int argc, char** argv) {
         // If there are no more cbAfterDelay callbacks,
         // the next deadline is max value, so end the simulation now
         if (next_time == static_cast<vluint64_t>(~0ULL)) {
-            vl_finish(__FILE__, __LINE__, "");
             break;
         } else {
             main_time = next_time;


marlonjames:
`vl_finish()` will stop the simulation immediately and exit if it is called twice.
Do any of your testbenches use the `$finish` system task? Because that calls `vl_finish()` in Verilator.

marlonjames:
The better approach is probably to either replace the call to `vl_finish()` with `Verilated::gotFinish(true);`, or to register a Verilator flush or exit callback to ensure the coverage is written out before exit.

On second though, because we break out of the while loop anyway, we might not need to call either function, just break.

forensicgarlic:
the tests i'm running seeing the need for writing before vl_finish is called, and outside that loop checking the time left are a mix of the things in the cocotb repo in examples/ and tests/ I'd have to go back and run them again to see which tests precisely were which. 

Happy to update for a different exit path. Do you prefer just removing the vl_finish() call?

marlonjames:
Yes, I would just remove the `vl_finish()` call.

forensicgarlic:
gave that a shot, and verified with test_1279 and adding `--coverage` to the Makefile.verilator `EXTRA_ARGS` variable, without the change, no coverage.dat was generated, but with the change it was.

documentation/source/newsfragments/1471.bugfix.rst
```diff
@@ -0,0 +1 @@
+fix verilator.cpp for writing coverage files when calls to vl_finish end tests
```
marlonjames, line 1:
```suggestion
Fixed Verilator not writing coverage files in some cases.
```

Also update this file name to `1478.bugfix.rst`

forensicgarlic: fixed in recent push.

documentation/source/newsfragments/1478.bugfix.rst
```diff
@@ -0,0 +1 @@
+fix verilator.cpp for writing coverage files when calls to vl_finish end tests
```
marlonjames, line 1:
```suggestion
Fixed Verilator not writing coverage files in some cases.
```

documentation/source/newsfragments/1478.bugfix.rst
```diff
@@ -0,0 +1 @@
+fix Verilator not writing coverage files in some cases
```
marlonjames, line 1:
```suggestion
Fixed Verilator not writing coverage files in some cases
```
Apologies for nitpicking but these end up in the release notes.

forensicgarlic: my apologies, i missed this on the first request.

documentation/source/newsfragments/1478.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed Verilator not writing coverage files in some cases
```
marlonjames, line 1:
```suggestion
Fixed Verilator not writing coverage files in some cases.
```

marlonjames (approved):

PR #3226, opened 2023-02-03
Add support for pytest.raises "Failed" exception

ktbarrett:
Closes #3196.

PR #3227, opened 2023-02-03
Catch exceptions derived from BaseException inside Tasks

### LINKED ISSUE #3196: Regression manager does not handle failed pytest.raises() result
cococtb: master (4aee0a0895a36606f9263df9370405663f496d22)
Ubuntu 20.04.2 LTS on WSL2
simulator: Verilator 5.007 devel rev v5.006-7-g4a8cfe367

When `pytest.raises()` fails, it raises a `Failed` exception.

This exception derives from `BaseException`, so the check in `RegressionManager._score_test()` fails:
https://github.com/cocotb/cocotb/blob/4aee0a0895a36606f9263df9370405663f496d22/cocotb/regression.py#L395-L400

 The raised `Failed` exception propagates up through `Scheduler._test_completed()` and into the simulator extension module.

Here is an example traceback:
```
567237.42ns INFO     cocotb.regression                  running test_integer_overflow (158/183)
                                                          Test integer overflow.
ERROR: called callback function threw exception
  File "/home/marlon/git/cocotb/.nox/dev_test_sim-sim-verilator-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.8/site-packages/cocotb/scheduler.py", line 349, in _test_completed
    self._handle_result(test)
  File "/home/marlon/git/cocotb/.nox/dev_test_sim-sim-verilator-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.8/site-packages/cocotb/regression.py", line 332, in _handle_result
    self._record_result(
  File "/home/marlon/git/cocotb/.nox/dev_test_sim-sim-verilator-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.8/site-packages/cocotb/regression.py", line 506, in _record_result
    test_pass, sim_failed = self._score_test(test, outcome)
  File "/home/marlon/git/cocotb/.nox/dev_test_sim-sim-verilator-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.8/site-packages/cocotb/regression.py", line 396, in _score_test
    outcome.get()
  File "/home/marlon/git/cocotb/.nox/dev_test_sim-sim-verilator-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.8/site-packages/cocotb/outcomes.py", line 53, in get
    raise self.error
  File "/home/marlon/git/cocotb/tests/test_cases/test_cocotb/test_handle.py", line 312, in test_integer_overflow
    await int_overflow_test(dut.stream_in_int, 32, "ovfl", limits)
  File "/home/marlon/git/cocotb/tests/test_cases/test_cocotb/test_handle.py", line 148, in int_overflow_test
    signal.value = value
  File "/home/marlon/git/cocotb/.nox/dev_test_sim-sim-verilator-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.8/site-packages/_pytest/python_api.py", line 984, in __exit__
    fail(self.message)
  File "/home/marlon/git/cocotb/.nox/dev_test_sim-sim-verilator-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.8/site-packages/_pytest/outcomes.py", line 194, in fail
    raise Failed(msg=reason, pytrace=pytrace)
Failed: DID NOT RAISE <class 'OverflowError'>
- :0: Verilog $finish
- /home/marlon/git/cocotb/.nox/dev_test_sim-sim-verilator-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.8/site-packages/cocotb/share/lib/verilator/verilator.cpp:118: Verilog $finish
- /home/marlon/git/cocotb/.nox/dev_test_sim-sim-verilator-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.8/site-packages/cocotb/share/lib/verilator/verilator.cpp:118: Second verilog $finish, exiting
```

marlonjames:
Closes #3196.
Alternative to #3226.

Exceptions that are raised inside coroutines running in Tasks are captured inside an outcomes.Error object. Previously, when this object was unwrapped in the Scheduler or RegressionManager, the exception inside was not handled correctly if the exception was derived from BaseException and not Exception.

This was the case with pytest's Failed exception that is raised when the following checks fail in a Task:
* pytest.raises
* pytest.warns
* pytest.deprecated_call

### DIFF
--- a/cocotb/regression.py
+++ b/cocotb/regression.py
@@ -391,7 +391,9 @@ class RegressionManager:
 
         try:
             outcome.get()
-        except Exception as e:
+        except (KeyboardInterrupt, SystemExit):
+            raise
+        except BaseException as e:
             result = remove_traceback_frames(e, ["_score_test", "get"])
         else:
             result = TestSuccess()
@@ -402,9 +404,6 @@ class RegressionManager:
             and not test.expect_error
         ):
             self._log_test_passed(test, None, None)
-
-        elif isinstance(result, AssertionError) and test.expect_fail:
-            self._log_test_passed(test, result, "failed as expected")
 
         elif isinstance(result, TestSuccess) and test.expect_error:
             self._log_test_failed(test, None, "passed but we expected an error")
@@ -429,6 +428,9 @@ class RegressionManager:
             else:
                 self._log_test_failed(test, result, "errored with unexpected type ")
                 result_pass = False
+
+        elif test.expect_fail:
+            self._log_test_passed(test, result, "failed as expected")
 
         else:
             self._log_test_failed(test, result, None)
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -581,7 +581,7 @@ class Scheduler:
                 coro.log.info("Test stopped by this forked coroutine")
                 e = remove_traceback_frames(e, ["_unschedule", "get"])
                 self._abort_test(e)
-            except Exception as e:
+            except BaseException as e:
                 coro.log.error("Exception raised by this forked coroutine")
                 e = remove_traceback_frames(e, ["_unschedule", "get"])
                 warnings.warn(
--- a/documentation/source/newsfragments/3196.bugfix.rst
+++ b/documentation/source/newsfragments/3196.bugfix.rst
@@ -1,0 +1,1 @@
+The :data:`Regression Manager <cocotb.regression_manager>` now correctly handles exceptions raised in tests when the exceptions inherit from `BaseException`.
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -27,7 +27,7 @@
 
 # Do not fail on DeprecationWarning caused by virtualenv, which might come from
 # the site module.
-export PYTHONWARNINGS = error,ignore::DeprecationWarning:site
+export PYTHONWARNINGS = error,ignore::DeprecationWarning:site,always::FutureWarning:cocotb.scheduler
 
 REGRESSIONS :=  $(shell ls test_cases/)
 
--- a/tests/test_cases/test_cocotb/common.py
+++ b/tests/test_cases/test_cocotb/common.py
@@ -26,3 +26,7 @@ async def _check_traceback(running_coro, exc_type, pattern, *match_args):
 
 class MyException(Exception):
     ...
+
+
+class MyBaseException(BaseException):
+    ...
--- a/tests/test_cases/test_cocotb/test_tests.py
+++ b/tests/test_cases/test_cocotb/test_tests.py
@@ -10,10 +10,12 @@ Tests of cocotb.test functionality
 """
 from collections.abc import Coroutine
 
-from common import MyException
+import pytest
+from _pytest.outcomes import Failed
+from common import MyBaseException, MyException
 
 import cocotb
-from cocotb.triggers import Timer
+from cocotb.triggers import NullTrigger, Timer
 
 
 @cocotb.test(expect_error=NameError)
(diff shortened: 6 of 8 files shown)


marlonjames:
Looks like test scoring will fail with `@cocotb.test(expect_fail=True)`.

ktbarrett:
Catching `BaseException` will also catch `OSError`, `KeyboardInterrupt`, etc. which we do not want to catch and squash. I think we should be specifically catching `Failed`. Really `Failed` should *not* inherit from `BaseException`, the pytest devs should know better.

marlonjames:
`asyncio` catches `BaseException`, but re-raises `KeyboardInterrupt` and `SystemExit`:

https://github.com/python/cpython/blob/433fb3ef08c71b97a0d08e522df56e0afaf3747a/Lib/asyncio/tasks.py#L266-L289

cocotb is also already catching `BaseException`, it just blows up later when it's unwrapped:

https://github.com/cocotb/cocotb/blob/49e46b7b7067faaed7f66680fbc785bf8e3a9d22/cocotb/decorators.py#L215-L225

marlonjames:
Also, `asyncio.CancelledError` is derived from `BaseException`, so we would have to special-case that later when cocotb throws that into the running Task.

I think it makes more sense to catch `BaseException` and add special-casing for things that should be handled differently.

ktbarrett:
They really muddied the Exception hierarchy up didn't they? I thought the idea was that that which derived from `BaseException` represents unrecoverable errors that should probably immediately exit (respecting cleanup) and that which derives from `Exception` is potentially recoverable as a part of program logic (like catching `KeyError` rather than using `.get()`). Maybe that's just me rationalizing.

I don't really like there not being general solution and having to decide on a case-by-case basis what constitutes test failure, test exception, and panic; but it seems like that's what they left us. Can get this PR to the point where all builtin and pytest exceptions are handled?

marlonjames:
Should I update the tests to use your trick for getting `Failed`?

marlonjames:
Hmm, for some reason, when running inside a nox session, the `FutureWarning` about unwatched tasks causes an error.
When I run the test directly on a bash shell it works fine.

https://github.com/cocotb/cocotb/actions/runs/4090365416/jobs/7053734418#step:22:1036

marlonjames:
The `warnings` filters are different.

`nox -s dev -- bash`:
```
warnings.filters :
[
    ('default', None, <class 'Warning'>, None, 0),
    ('default', None, <class 'DeprecationWarning'>, '__main__', 0),
    ('ignore', None, <class 'DeprecationWarning'>, None, 0),
    ('ignore', None, <class 'PendingDeprecationWarning'>, None, 0),
    ('ignore', None, <class 'ImportWarning'>, None, 0),
    ('ignore', None, <class 'ResourceWarning'>, None, 0)
]
```


`nox -s "dev_test_sim(sim='icarus', toplevel_lang='verilog', gpi_interface='vpi')"`:
```
warnings.filters :
[
    ('ignore', None, <class 'DeprecationWarning'>, re.compile('site\\Z'), 0),
    ('error', None, <class 'Warning'>, None, 0),
    ('default', None, <class 'DeprecationWarning'>, '__main__', 0),
    ('ignore', None, <class 'DeprecationWarning'>, None, 0),
    ('ignore', None, <class 'PendingDeprecationWarning'>, None, 0),
    ('ignore', None, <class 'ImportWarning'>, None, 0),
    ('ignore', None, <class 'ResourceWarning'>, None, 0)
]
```

cmarqu:
> The `warnings` filters are different.

This is because the `dev_test_sim` nox session calls `make test` and the Makefile sets `PYTHONWARNINGS`: 
https://github.com/cocotb/cocotb/blob/cf3486f7473b52b78dde2f025af9791e95055e3c/tests/Makefile#L30

cmarqu:
Maybe worth a newsfragment?

ktbarrett:
> `asyncio` catches `BaseException`, but re-raises `KeyboardInterrupt` and `SystemExit`

Are you not going to handle that? If I Ctrl-C in a sim, I want it to die, cleanly.

marlonjames:
I was going to leave that for another PR.
We will want something similar to `SimFailure`, where we can cleanly fail the current test and the rest of the tests after.
Also, even with that, we aren't protecting internal scheduler code, so if you hit Ctrl+C outside of a running coroutine it will blow up anyway.

ktbarrett:
> I was going to leave that for another PR.

I would consider cocotb to be in a broken state if it's not handled in this PR. I don't think it's an incredible amount of work to special case `KeyboardInterrupt` and `SystemExit` to simply reraise. What it does right now isn't fantastic, but it works.

ktbarrett (approved):

Issue #3194, opened 2023-01-22
[tracking] Get the cocotb test suite passing with Verilator


imphil:
This issue tracks progress towards getting the cocotb test suite passing with Verilator.

Status updates:
- [x] Timed callbacks. Fixed in Verilator 5.004 (https://github.com/verilator/verilator/commit/c2bdd06fccd4ef551948ed6dfe0eff76004fa117), so we're back to where we roughly were with Verilator 4.106.
- [ ] Once we have a passing CI suite: Update documentation to indicate which Verilator version is supported.

Various other test failures need to be diagnosed and issues have to be created.

imphil:
Test failures with Verilator master as of today:

```
Failure in testsuite: 'all' classname: 'test_iteration_es' testcase: 'recursive_discovery' with parameters 'all'
Error: Test test_iteration_es:recursive_discovery failed
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'pseudo_region_access' with parameters 'all'
Error: Test test_discovery:pseudo_region_access failed
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'access_signal' with parameters 'all'
Error: Test test_discovery:access_signal failed
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'access_single_bit' with parameters 'all'
Error: Test test_discovery:access_single_bit failed
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'access_integer' with parameters 'all'
Error: Test test_discovery:access_integer failed
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'access_const_string_verilog' with parameters 'all'
Error: Test test_discovery:access_const_string_verilog failed
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'access_var_string_verilog' with parameters 'all'
Error: Test test_discovery:access_var_string_verilog failed
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'access_internal_register_array' with parameters 'all'
Error: Test test_discovery:access_internal_register_array failed
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'access_gate' with parameters 'all'
Error: Test test_discovery:access_gate failed
Failure in testsuite: 'all' classname: 'test_discovery' testcase: 'type_check_verilog' with parameters 'all'
Error: Test test_discovery:type_check_verilog failed
[...]

Issue #1275, opened 2020-01-01
Access SV structs as HierarchyObject on verilator


nndurj:
On verilator, SV structs are shown as  `ModifiableObject`. Because of this it shows up a single dimensional packed vector, and we can't access the struct fields separately. On questasim, they are shown as `HierarchyObject`, and we can access the fields.

### Test Program

```systemverilog
module t();

struct packed{
    logic a;
    logic b;
} hdr;

endmodule
```
![image](https://user-images.githubusercontent.com/1689958/71638108-ee12d580-2c7c-11ea-87f9-bf218a0a7fd8.png)

wallento:
This is certainly a Verilator issue, I am just adding vpi support on the fly and this is probably not handled as queried by cocotb. I will trace it and create a standalone test for Verilator.

wallento:
This seems harder than I hoped as structs are not handled at all right now (https://github.com/verilator/verilator/issues/860). After that I can access it with VPI which will be another larger task. This will probably take a while..

nanduraj1:
There is a switch `--trace-structs` that shows struct/union hierarchy in vcd file. The code snippet is below. Is it possible to do something similar for vpi?  

https://github.com/verilator/verilator/blob/174fd1bf0ed5256cf62aa1cebf0774aafb4ff1cc/src/V3TraceDecl.cpp#L306-L341

imphil:
@nanduraj1 This is a missing functionality in Verilator. You need to work with the Verilator project to get this functionality added; once it is there and implemented according to the VPI standard, cocotb shouldn't be able to use it immediately.

imphil:
Closing this for now as there's nothing we can do on the cocotb side.

ktbarrett:
Support for this was removed, so this is a noop.

PR #3237, opened 2023-02-06
CI: Update regression tests for Verilator

marlonjames:
Update regression test suite for Verilator. Add expected test failures and skip some tests because of missing VPI support.

Part of #3194.

Includes alternative to #3235.

Needs #3196 fixed.

### DIFF
--- a/.github/generate-envs.py
+++ b/.github/generate-envs.py
@@ -125,22 +125,23 @@ ENVS = [
     {
         "lang": "verilog",
         "sim": "verilator",
-        "sim-version": "v4.106",
-        "os": "ubuntu-20.04",
+        "sim-version": "8ea814e4b4ce7416188130647c87dea41401a5b6",
+        # Needs 22.04 for newer GCC with C++ coroutine support used with --timing mode
+        "os": "ubuntu-22.04",
         "python-version": "3.8",
-        # Various cocotb tests are known to fail with Verilator 4.106.
-        # Newer versions of Verilator are not working at all.
-        # See also https://github.com/cocotb/cocotb/issues/2300
-        "group": "experimental",
+        # Various cocotb tests are known to fail with Verilator.
+        # See also https://github.com/cocotb/cocotb/issues/3194
+        "group": "ci",
     },
     {
         "lang": "verilog",
         "sim": "verilator",
         "sim-version": "master",
-        "os": "ubuntu-20.04",
+        # Needs 22.04 for newer GCC with C++ coroutine support used with --timing mode
+        "os": "ubuntu-22.04",
         "python-version": "3.8",
-        # Tests are currently not expected to work at all.
-        # See also https://github.com/cocotb/cocotb/issues/2300
+        # Various cocotb tests are known to fail with Verilator.
+        # See also https://github.com/cocotb/cocotb/issues/3194
         "group": "experimental",
     },
     # Test other OSes
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -171,7 +171,7 @@ jobs:
     - name: Set up Verilator (Ubunutu - source)
       if: startsWith(matrix.os, 'ubuntu') && matrix.sim == 'verilator'
       run: |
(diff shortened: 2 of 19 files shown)


marlonjames:
Do we want Verilator master in the test suite, or some fixed known good version and keep master in the experimental CI run?

imphil:
> Do we want Verilator master in the test suite, or some fixed known good version and keep master in the experimental CI run?

Only use pinned versions in CI to ensure we are actually testing the PR, and not the PR and a new tool version at the same time. So yes, let's keep Verilator master in the extended tests, and use (ideally) the latest release version of Verilator, or if it has to be, a specific development version, in CI.

imphil:
Thanks @marlonjames! 

I am slightly worried that we are hiding problems with Verilator that then bite our users. So if we add test exceptions, we should also make sure that we keep track of these deficiencies in a way that (a) the upstream project knows about them and can eventually get them fixed, and (b) our users know about them as well.

Most of the time we have been using the following approach:
* Have a cocotb issue open for a missing simulator feature, and link to that in the test exception. Ensure that the issue description is understandable to the end user (i.e., not an obscure bug report that then 10 comments down is identified as an simulator issue -- rewrite the issue description if necessary, or create a new issue).
* If it's an upstream issue, mark the issue as "upstream" and link to an upstream bug report. Keep this issue open until it's actually fixed upstream, and verified within cocotb (which then also involves removing the exception).

We are linking to all simulator-related issue in the docs (https://docs.cocotb.org/en/stable/simulator_support.html#sim-verilator-issues), which gives users a reasonably easy way to look for simulator-specific limitations. [I just realized we also show closed issues, I'll do a PR that fixes that next.]

cocotb/share/makefiles/simulators/Makefile.verilator
```diff
@@ -15,11 +15,17 @@ clean::
 
 else
 
-CMD := verilator
+CMD_BIN := verilator
 
-ifeq ($(shell which $(CMD) 2>/dev/null),)
-# Verilator is not in PATH, lets start searching for it
-$(error Cannot find verilator.)
+ifdef VERILATOR_BIN_DIR
```
imphil, line 20: Is there a specific need for this feature? We require most other simulators to be in PATH.

imphil:
Addition: Looking at https://github.com/cocotb/cocotb/issues/1275 for example I realize that we and I were rather inconsistent in our approach in the past. So let's maybe use this bunch of exceptions to find a consistent approach that we're happy with :smile:

cocotb/share/makefiles/simulators/Makefile.verilator
marlonjames: Every other simulator makefile supports a `<SIM>_BIN_DIR` environment variable. I also find it useful for my testing setup.

imphil: Good with me then, sorry for the noise!

ktbarrett:
@marlonjames Uhhh. This has been here for a while. How is this going?

marlonjames:
It is close. I ran into a few issues with Verilator that I fixed upstream. There are a few left but I am going to make sure they are documented and leave them for now in the interest of getting this done. I should have some time over the next week to wrap it up.

toddstrader:
Awesome.  FYI, `test_package` works now on HEAD of Verilator, but if we want to test against released versions we'll have to wait for the next cut for that test.

imphil:
I'm happy to test against an arbitrary git hash in this PR, and then switch to a released version once that's available.

ktbarrett:
Is there any reason not to test against HEAD?

imphil:
We test against verilator master in the experimental bucket (https://github.com/cocotb/cocotb/blob/cb683a83368af6703a6a326a62d9911603028a66/.github/generate-envs.py#L138-L139). For the CI bucket, we need to have fixed versions.

imphil (approved):
Great work @marlonjames, thanks a lot for the many issue references and the work you did upstream with Verilator as well!

Issue #2769, opened 2021-11-07
`LogicArray`: higher performance implementation


ktbarrett:
The new `LogicArray` type is not nearly as fast as it should be (in fact it's quite a bit slower than `BinaryValue`). One possibility is to use a polymorphic representation. That is, the constructor, if it takes an integer or string, will validate the value but keep the passed in value as it's representation instead of converting it to a `List[Logic]` immediately. The conversion to `List[Logic]`, or conversion to some other representation, will be done by need. The alternative representations should be invalidated when the value is mutated via `__setitem__`. This representation should allow for better performance in the common `int` -> `LogicArray` -> `int` and `str` -> `LogicArray` -> `str` paths used when values are set in the simulator, and in such cases `LogicArray` simply acts as a validator.

AKA the PIMPL pattern.

ktbarrett:
Closed by #4093.

PR #706, opened 2018-11-29
Big endian unsigned bit representation fix

mciepluc:
Resolves #608 

I find it a bit strange that big endian representation is default, but I don't want to change this.

Issue #2666, opened 2021-08-13
Remove `Bit`?


ktbarrett:
We decided that all we really need from the types right now are a minimal set of types that correspond to the what's current available to the GPI. `Bit` and `BitArray` don't map, and `BitArray` was removed from the `LogicArray` PR. Should we remove `Bit`? I can't see it being of much use, but it isn't much to maintain. But removal will open the doors for adding simpler semantics to `Logic`:

```python
handle.value[0] == "0"  # not currently possible due to how Logic.__eq__ must behave
```

ktbarrett:
We don't have to remove `Bit` to accomplish binary operations with literals.

ktbarrett:
Lets reconsider on whether to remove `Bit`, as I don't think it has any real value and I doubt it's used anywhere.

PR #3244, opened 2023-02-12
Replace BinaryValue with LogicArray 

aubindetrez:
Replace BinaryValue with LogicArray like mentioned in issue #608 and PR #706.
All internal references to BinaryValue are removed.

Please **this is a Breaking Changes - do not merge as-is**, further discussions are required.

I think BinaryValue should be reintroduced for compatibility reason but marked deprecated. (And next major release it can be definitely removed). Let me know what you think.

Status / To dos:
- [x] Remove BinaryValue and internal references to it.
- [x] Pass all tests
     - See list of tests below
- [x] Implement all missing operators
     - See list below
- [ ] Make type conversion more explicit https://github.com/cocotb/cocotb/issues/3418
    - Remove implicit conversion to unsigned integers (https://github.com/cocotb/cocotb/pull/3244#issuecomment-1722938137)
- [ ] Remove Bit
- [x] Construct LogicArray from String
- [x] Construct LogicArray from Iterable of Logic
- [ ] Construct LogicArray from integer with explicit endianness
- [x] Convert LogicArray to String (binary representation)
- [ ] Convert LogicArray to Integer and explicitly specify endianness and representation
- [x] get / set an element
- [x] get the length
- [ ] bitwise logic
- [ ] Explicit arithmetic shifts
- [ ] Explicit logical shifts
- [ ] No arithmetic operators with implicit conversion to int...
- [x] get 'slices' (`value = signal[3:0]`)
- [ ] set 'slices'  (`signal[3:0] = value`)
- [ ] Have comparable performance to BinaryValue
    - Refactor LogicArray according to https://github.com/cocotb/cocotb/issues/2769
    - Note: We may try different implementation in other branches, and merge the best one in this PR.
    - See below for the current performance compared to BinaryValue
- [ ] Proof of concept: Port one "big cocotb project" to use LogicArray instead of BinaryValue
- [ ] Add meaningful error message to help with porting big code bases
- [ ] `cocotb/cocotb/handle.py` uses `downto` by default, do we want to add an option for using `to` ?
- [ ] Update the documentation (There are many references to BinaryValue)

## `LogicArray`'s operators compared to `BinaryValue`
The list of operators have been moved to: https://gist.github.com/aubindetrez/16ffe3b172ff0edd138001d61437d4ef (to improve clarity.

## Failing tests
- [x] `pytest cocotb/tests/pytests/test_logic_array.py` depends on `BinaryValue`
- [x] Remove `pytest cocotb/tests/pytest/test_binary_value.py`
- [x] Test `test_matrix_multiplier_icarus` from `cocotb/cocotb/examples/matrix_multiplier/tests/test_matrix_multiplier.py` with error `ModuleNotFoundError: No module named 'cocotb.binary'`
- [x] Test `test_matrix_multiplier_ghdl` 
- [x] Test `issue_142_overflow_error` from `cocotb/tests/test_cases/issue_142/issue_142.py` with error `Test failed with RANDOM_SEED=1694419089`
- [x] Test `test_assigning_structure` from `cocotb/tests/test_cases/test_cocotb/test_deprecated.py` with error `Test failed with RANDOM_SEED=1694419168`
- [x] Test `test_force_release` from `cocotb/tests/test_cases/test_force_release/test_force_release.py` with error `Test failed with RANDOM_SEED=1694419237`
- [x] Test `test_my_design.my_first_test` from `cocotb/cocotb/examples/doc_examples/quickstart` with error `AssertionError: my_signal_2[0] is not 0!`
- [x] Test `test_my_design.my_second_test` from `cocotb/cocotb/examples/doc_examples/quickstart` with error `AssertionError: my_signal_2[0] is not 0!`
- [x] `issue_142_overflow_error`
- [x] `test_force_release`

*If checked, the test passed at least locally (nox + pytest) but for the CI you have to scroll down this page.

## Performance
- Performance results before any optimization: `LogicArray is 5x slower than BinaryValue for MatrixMultiplication` https://gist.github.com/aubindetrez/bb6e311bbfe8044a839ee22f98ae7b18
- [Experiment 1: 'Cache' self.integer](https://github.com/aubindetrez/cocotb/tree/logicarray_cache_integer): `LogicArray is 2.7x slower than BinaryValue for MatrixMultiplication`


aubindetrez:
@ktbarrett I would greatly appreciate your feedback on this.

cocotb/types/logic_array.py
```diff
@@ -189,6 +201,31 @@ def is_resolvable(self) -> bool:
...
+        """
+        return int(self)
+
+    def __eq__(self, other):
+        """
+        Overrides the default implementation to enable comparision
+        with integers and strings (case insensitive)
+        """
+        if isinstance(other, int):
+            return self.integer == other
```
ktbarrett, line 215: Lets not make an assumption about binary representation.
```suggestion
            return self == LogicArray(other)
```

cocotb/types/logic_array.py
```diff
@@ -189,6 +201,31 @@ def is_resolvable(self) -> bool:
...
+        if isinstance(other, str):
+            # case insensitive to make sure 'x' == 'X' and 'z' == 'Z'
+            # internally LogicArray uses upper case but use may not know about it
+            return self.binstr == other.upper()
```
ktbarrett, line 219: And in fact, this could be collapsed into the previous statement by using a tuple in the `isinstance` check.
```suggestion
            return self == LogicArray(other)
```

tests/pytest/test_logic_array.py
```diff
@@ -81,6 +81,22 @@ def test_logic_array_and():
...
+        p.integer
+
+def test_integer_reverse():
+    l = LogicArray("00XZ1")
+    p = l.reverse()
+    assert(l == LogicArray("00XZ1"))
+    assert(p == LogicArray("1ZX00"))
+
+    m = LogicArray("1010")
+    assert(m.reverse().binstr == "0101")
```
ktbarrett, line 99: Also add a check for the reversal of the range.

ktbarrett (commented):
I'm not against this change. I appreciate the work! I think this PR should come in, but maybe we should fix a few critical bugs and do a 1.8 release before this comes in. 

There is some additional development on these types that needs to occur, namely for performance (#2769). Also we need to consider removing `Bit` (#2666), as I don't think it's really useful.

cocotb/types/logic_array.py
```diff
@@ -189,6 +201,31 @@ def is_resolvable(self) -> bool:
...
+        else:
+            return self == other
+
+    # __ne__ is not required for python3 because by default it negates __eq__
+
+    def __int__(self):
+        """
+        int(LogicArray) is equivalent to LogicArray.integer
+        """
```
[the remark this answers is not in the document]
aubindetrez: Good point.
Removing it would break a few tests but I can make it work.

cocotb/types/logic_array.py
```diff
@@ -162,10 +166,11 @@ def __init__(
                 if bitlen > len(range):
                     raise ValueError(f"{value} will not fit in {range}")
                 self._value = [Logic(v) for v in _int_to_bitstr(value, len(range))]
+        elif isinstance(value, ctypes.Structure):
+            # ctypes.Structure is also typing.Iterable, but it is not supported
+            raise ValueError(f"{value} is an instance of ctypes.Structure which cannot be converted to LogicArray")
```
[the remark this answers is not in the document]
aubindetrez: The idea was to help user switch to LogicArray with a (hopefully) straightforward error.
While migrating big projects it could save even more time.

ktbarrett:
You should install pre-commit and run pre-commit on your commits. This should make sure that your commits will pass all the lint checks.

```sh
pip install pre-commit
pre-commit install
```

You can run the checks at any time using
```sh
pre-commit run -a
```

teobiton:
Hi,

Regarding the status, performance analysis is something to consider as well.
First observations by comparing profiling from the matrix multiplier with BinaryValue and with LogicArray, is that there is a huge drop in performance.

Based on current master and on the `remove_binary_value` branch:

| **Branch**                   | master                                | remove_binary_value         |
|------------------------------|---------------------------------------|-----------------------------|
| Most time consuming function | binary.py:249(_convert_from_unsigned) | logic_array.py:202(integer) |
| Total time spent (s)         | 1.294                                 | 16.51                       |
| Cumulative time spent (s)    | 1.993                                 | 31.4                        |

The source of the performance drop comes from the way the value is handled in LogicArray: each time the integer representation is needed, the array is looped over to build it.
As mentionned in #2769, it would be best to keep to python native types for better performance.

aubindetrez:
I tried to "cache" LogicArray.integer in a separate branch (look for "Experiment 1: 'Cache' self.integer" in first top of the PR). I get significant performance improvements with a simple change, but it's nowhere near BinaryValue.

I will keep experimenting with different implementations and share results and code in the first comment, at the top of the PR. If anyone else want to give it a try, you can comment below with your results/branch and I will add it to the list too.

aubindetrez:
The latest commit passes CI.

FYI: To run python3.7 tests locally, I used `nox -s dev_test-3.7` with:
```diff
--- a/noxfile.py
+++ b/noxfile.py
@@ -123,7 +123,7 @@ def dev_build(session: nox.Session) -> None:
     session.warn("No building is necessary for development sessions.")
 
 
-@nox.session
+@nox.session(python=["3.10", "3.7", "3.8"])
 def dev_test(session: nox.Session) -> None:
     """Run all development tests as configured through environment variables."""
```
Just in case someone else needs to debug python3.7 compatibility.

ktbarrett:
I think we can also remove `Bit` as it currently serves no purpose. There was interest by me originally to add more types, but I think we aren't doing that anymore.

ktbarrett:
Generally what we need is:
- construct from string
- construct from iterable of logic
- construct from integer specifying the endianness (representation can be deduced since we only support two's complement and unsigned, which is a subset)
- conversion to string
- conversion to integer specifying the endianness and representation
- sequence-like stuff: get and set elements, iteration, length, blah blah blah
- bitwise logic

What we cannot assume:
- shifts (are they arithmetic or logical)
- any arithmetic operators
- how users want hex and octal representations formatted

teobiton:
Hi all,

To expand on the performance trials, I implemented an int and string representation for LogicArray that is used as much as possible when valid, i.e not modified by __set_item__ and other arithmetic operations that are implemented for now (see [logicarray_perf_repr](https://github.com/aubindetrez/cocotb/tree/logicarray_perf_repr)).

The results are far better than the original implementation but we are nowhere near the results with BinaryValue per the construction of LogicArray inheriting from Array[Logic]. A lot of time is spent rebuilding the array even with the string and int representations (total time 7 seconds for binstr, integer has better results around 4 seconds total time).

ktbarrett:
@teobiton The next stumbling block is probably the construction of the `Range` object, which occurs if no range is given. You can test that by creating a Range object *once* and passing it in the range argument in your benchmark and see if performance improves.

ktbarrett:
@teobiton Actually, I see you are still creating the `List[Logic]` repr even when being constructing from a string. You should populate the `str` represent *instead of* the `List[Logic]` repr when given a string and create the `List[Logic]` impl on demand. 

An easy way to do this is with getter which encapsulate the creation. The only time you need to set the string and int representations is on construction and I don't know if we need to cache them after that fact, so once they are invalidated on a `__setitem__`, they are permanently invalidated. You might even get super extra and do general-purpose caching on the string and/or int representations as well.

A small mock up
```python
class LogicArray:
    @overload
    def __init__(self, value: str):
        ...

    @overload
    def __init__(self, value: int, *, byteorder: str) -> None:
        ...

    @overload
    def __init__(self, value: SizedIterable[LogicConstructible]) -> None:
        ...

    def __init__(self, value, *, byteorder: str):
        self._str_impl_valid = False
        self._list_impl_valid = False
        self._int_impl_valid = False
        if isinstance(value, str):
            self._str_impl = value
            # validate
            self._str_impl_valid = True
        elif isinstance(value, int):
            self._int_impl = value
            self._int_impl_byteorder = byteorder
            # validate
            self._int_impl_valid = True
        else:
            self._list_impl = list(value)
            # validate
            self._list_impl_valid = True

    def __str__(self) -> str:
        if self._str_impl_valid:
            return self._str_impl
        # otherwise construct from int or logic representation

    def as_int(self, *, byteorder: str, signed: bool) -> int:
        if self._int_impl_valid and self._int_impl_byteorder == byteorder:
            # might need to work out if two's complement is needed if 'signed' is False and '_int_impl' is negative
            return self._int_impl
        # otherwise construct from str, int, or logic representation

    @property
    def _list_impl(self) -> List[Logic]:
        if self._list_impl_valid:
            return self._list_impl_
        # construct from str or int representation

    def __iter__(self) -> Iterable[Logic]:
        return iter(self._list_impl)

    def __reversed__(self) -> Iterable[Logic]:
        return reversed(self._list_impl)

    def __setitem__(self, key: int, value: LogicConstructable) -> None:
        self._int_impl_valid = False
        self._str_impl_valid = False
        # validate
        return self._list_impl[index]
```

aubindetrez:
We (I) still need to change the API as discussed by Philipp and you. This will break many tests and examples.

We also need to improve performance further.

I personally don't have a lot of free time right now but I hope it should get better in a few weeks. If anyone else wants to work on it he/she is more than welcome.

teobiton:
So I picked up things where I left them and tried a new implementation based on your comments @ktbarrett, still on the same branch [logicarray_perf_repr](https://github.com/aubindetrez/cocotb/tree/logicarray_perf_repr)). It's not quite done yet.

Do we still want a `Range` being constructed and inheriting from `Array[Logic]` ? In my current trial I totally removed them from the class (breaking everything obviously), only building the Logic(_) array when required.

marlonjames:
@ktbarrett Is this PR superseded by your other handle PRs?

ktbarrett:
Superceded by #3634.

PR #3251, opened 2023-02-18
Remove public decorator

### LINKED ISSUE #1605: 'public' decorator?
It's only used by the decorators module and the generators module (at risk in #1493).

Alternatively, should we be using it more?

ktbarrett:
Closes #1605.


marlonjames (approved):

eric-wieser (commented):
This makes `from cocotb.decorators import *` do the wrong thing. If you remove the decorator, you should set `__all__` manually.

ktbarrett:
Maybe we should explicitly set the `__all__` in all the modules, and IMO that should be `__all__ = ()`, except maybe in a couple modules like `cocotb` and `cocotb.triggers`. Many of the current modules really should be private, the repo is currently pre unprincipled wrt imports. There should be one and perferably only one place an object is imported from.

ktbarrett:
Well seeing as we don't care about backwards compatibility anymore. I'm just gonna remove this.

PR #3263, opened 2023-03-09
Add `round_mode` to `with_timeout`

### LINKED ISSUE #3258: Make `round_mode` for `Timer` available on `wait_timeout`
`with_timeout` is missing the `round_mode` argument available to `Timer`. This would be very convenient to have.

ktbarrett:
Closes #3258. Depends on #3067.

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -192,6 +192,10 @@ class Timer(GPITrigger):
                When *units* is ``'step'``,
                the timestep is determined by the simulator (see :make:var:`COCOTB_HDL_TIMEPRECISION`).
 
+            round_mode (str, optional):
+                String specifying how to handle time values that sit between time steps
+                (one of ``'error'``, ``'round'``, ``'ceil'``, ``'floor'``).
+
         Examples:
 
             >>> await Timer(100, units='ps')
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -33,7 +33,7 @@ import warnings
 from collections.abc import Awaitable
 from decimal import Decimal
 from numbers import Real
-from typing import Optional, Union
+from typing import Any, Coroutine, Optional, TypeVar, Union
 
 import cocotb
 from cocotb import outcomes, simulator
@@ -46,6 +46,8 @@ from cocotb.utils import (
     lazy_property,
     remove_traceback_frames,
 )
+
+T = TypeVar("T")
 
 
 def _pointer_str(obj):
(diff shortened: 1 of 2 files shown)


cocotb/triggers.py
```diff
@@ -934,7 +940,12 @@ def __repr__(self):
         return fmt.format(type(self).__qualname__, self.signal, self.num_cycles)
 
 
-async def with_timeout(trigger, timeout_time, timeout_unit="step"):
+async def with_timeout(
+    trigger: Union[Trigger, Waitable, Coroutine[Any, Any, T]],
```
ktbarrett, line 944: The reason that `cocotb.coroutine` is not listed here is because of a circular dependency =/ We need to make a clear dependency graph.

cocotb/triggers.py
```diff
@@ -934,7 +940,12 @@ def __repr__(self):
...
+    timeout_time: float,
```
cmarqu, line 945:
```suggestion
    timeout_time: Union[numbers.Real, decimal.Decimal],
```
?

cocotb/triggers.py
```diff
@@ -978,6 +989,9 @@ async def with_timeout(trigger, timeout_time, timeout_unit="step"):
             Simulation time duration before timeout occurs.
         timeout_unit (str, optional):
             Units of timeout_time, accepts any units that :class:`~cocotb.triggers.Timer` does.
+        round_mode (str):
```
cmarqu, line 992:
```suggestion
        round_mode (str, optional):
```
?

cocotb/triggers.py
```diff
@@ -189,6 +191,10 @@ def __init__(
                When *units* is ``'step'``,
                the timestep is determined by the simulator (see :make:var:`COCOTB_HDL_TIMEPRECISION`).
 
+            round_mode (str):
```
cmarqu, line 194:
```suggestion
            round_mode (str, optional):
```
?

cocotb/triggers.py
ktbarrett: Technically this doesn't work with type checking. No type checker supports the numeric tower right now. Maybe
```suggestion
    timeout_time: Union[float, decimal.Decimal],
```

cmarqu (approved):

cocotb/triggers.py
ktbarrett: Closing #3067 would solve this.

eric-wieser: Usually you can write `cocotb.coroutine` when that happens

PR #3274, opened 2023-03-30
Update to new flows QIS and Qrun

AbdouTharwat:
Updated Makefile for Questa simulator to use the new flow QIS and qrun instead of +acc and vlog/vsim respectively
Also using Visualizer instead of the classic GUI for Live simulation and adding option for Visualizer post simulation mode 
By Default Questa runs in batch mode, to enable Live SIM mode set GUI=1, to enable Post SIM mode set GUI=2 
Added vis cmd for Visualizer beside vsim cmd for Post SIM mode, and added same checks for the cmd as vsim 
Changed way of invoking Questa from running .DO file into directly invoking Questa from bash/shell terminal which now enables user to change arguments or options and rerun without having to delete sim_build directory
VSIM_ARGS and other variables kept untouched for backward compatibility, also SCRIPT_FILE is kept for user custom DO file Added more variables for new design.bin and qwave.db file names for user custom preferences 
Changed MODELSIM_BIN_DIR variable name to QUESTA_BIN_DIR as Visualizer is now included with Questa
Updated also the documentation section for Mentor/Questa to mention the same

This is not a bug fix this is to enhance QuestaSim performance by using updated flows and options instead of legacy ones, the new Makefile.questa was tested and verified on all cocotb test examples with other tests. This request was old from Siemens EDA, but there was no bandwidth for working on this, thus we did it from our side in Siemens EDA.


imphil:
Thanks for your pull request, @AbdouTharwat! I'll have a closer look at your PR in the coming days, but a quick question before we start: we need to support older version of Questa and MentorSim in the Makefiles, either using compatible syntax, or (more likely, looking at your changes) using a version switch. Can you introduce such a switch for the minimum version of Questa and ModelSim that works with your changes?

AbdouTharwat:
Hi @imphil, thanks for your reply. Can you clarify what is the minimum version, do you mean ModelSim with least features? 
Also some of the new features will not work on older versions like 10.7 up to 2019.4 for example. Do you need the Makefile to work with old options for those old versions, and work with new ones for newer versions?

imphil:
> Also some of the new features will not work on older versions like 10.7 up to 2019.4 for example. Do you need the Makefile to work with old options for those old versions, and work with new ones for newer versions?

So the modified Makefile in this PR requires at least Questa 2019.4 or ModelSim 10.7?

AbdouTharwat:
Sorry for confusion. The modified Makefile in this PR requires at least version 2020.1 for both ModelSim or QuestaSim (they have same releases) to function with all options included. Any older versions than 2020.1 will not function as qrun was introduced in 2020.1

imphil:
> @imphil, Sorry for confusion. The modified Makefile in this PR requires at least version 2020.2 for both ModelSim or QuestaSim (they have same releases) to function with all options included. Any older versions than 2020.2 will not function

Thanks for confirming that. 

Can you introduce a version switch in the Makefile that first determines the version of Questa or ModelSim, and then calls the old (existing) code if it is a version below 2020.2, and your new code otherwise? This way we ensure that users of older versions can continue to use cocotb as they use it today, and users of newer versions get the updated qrun flow.

Let me know if you are lost somewhere, and I'm happy to take a closer look (but that might take a small while, I'm currently traveling).

AbdouTharwat:
Hi @imphil , The flowing modifications are made in Questa and ModelSim Makefiles:

1. Old flow is now in (Makefile.modelsim), while new flow is in (Makefile.questa)
2. Old Flow is chosen automatically if simulator is ModelSim, as it does not support all new options
3. If simulator is QuestaSim then we have two scenarios:
     a. Old flow is chosen if version is older than 2020.2 
     b. New flow is chosen otherwise
4. For VHDL, FLI is forced if version is older than 2022.3
     a. Modified this in old flow also, VHPI was left as option to user although it will not work
5. VHPI is left as an option starting version 2022.3
     a. FLI is left as default as it was
6. Also updated Questa section in documentation for instructions about Visualizer invoking in new flow

Now we made sure that old versions of Questa and ModelSim will work with old flow, and new versions will benefit from new options
Please let me know if you have other requests or inquiries. Thank you

themperek:
@AbdouTharwat we discussed this a bit and concluded that best idea would probably by create new Makefile `Makefile.questa_qrun` so that no back compatibility is broken. Is that ok for you @imphil ?

 We would like to include this in the next release. Would you have time to work on this @AbdouTharwat ? If not I can try to help.

imphil:
> create new Makefile Makefile.questa_qrun so that no back compatibility is broken. 

So the user would have to explicitly opt into the new behavior? I think that's non-ideal for two reasons:
* Users are unlikely to notice that they have to change something.
* The non-qrun functionality will go away at some point, so we wouldn't be providing a seamless update path for users.

Since we can relatively easily version-detect and give users the best possible approach I'd prefer that. That being said, we could of course do the implementation in different Makefiles and have the existing Makefile.questa spawn out to either Makefile.questa_qrun or Makefile.questa_compat -- but that feels much like an implementation detail that I would leave to the PR author.

imphil:
I'll have to fix #3281 before we can get more private CI testing on this, but I did a quick copy-and-paste to see how two different Makefiles would work out, with a version-detect to choose the most suitable one. https://github.com/imphil/cocotb/tree/questa-qrun-imphil has the updates. @themperek is that what roughly what you were thinking of?

AbdouTharwat:
Hi @imphil and @themperek, I thought about creating Makefile.questa_qrun but felt there will be many Questa files. Now that you are OK with that, I am fine too.
I agree with choosing best flow for user based on version detected. Users may not notice they had to.
For VHPI, I am OK with leaving option for user in all cases and errors in unsupported versions.
Regarding Visualizer, starting Questa version 2023.1 Visualizer is bundled with QuestaSim. Users of older versions may have QuestaSim without Visualizer . Problem is new flow does not support waves dumping with Classic GUI, only supports waves with Visualizer.

imphil:
Just a quick status update here: There were some issue with us getting access to the latest Questa 2023 releases. That's now resolved and I am making Questa 2023.2 available in CI for us to test the changes in this PR. I'm still short on time in the coming week, but I'll push forward as time permits.

imphil:
So, after a too long time I was finally able to finish and push the updates for this PR. Since I couldn't update this PR directly I opened a new PR at https://github.com/cocotb/cocotb/pull/3306. 

@AbdouTharwat Can you please have a look?

imphil:
Superseded by https://github.com/cocotb/cocotb/pull/3306

PR #3288, opened 2023-04-28
Fixing issue where startup callback caused mti_Desensitize to be called with a null parameter

kcuzner:
A recent change to `FliCbHdl` causes Modelsim DE 2022.4 to emit this suppressible error:

```
# ** Error (suppressible): (vsim-FLI-3132) mti_Desensitize(): NULL process ID parameter.
```

This fix performs a null check before calling `mti_Desensitize` to avoid this error being emitted. While not a functional problem it is annoying for log parsing to have errors like this contaminating the log.

### DIFF
--- a/cocotb/share/def/modelsim.def
+++ b/cocotb/share/def/modelsim.def
@@ -93,6 +93,8 @@ mti_Now
 mti_NowUpper
 mti_Quit
 mti_ReleaseSignal
+mti_RemoveLoadDoneCB
+mti_RemoveQuitCB
 mti_ScheduleWakeup
 mti_ScheduleWakeup64
 mti_Sensitize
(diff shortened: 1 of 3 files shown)


ktbarrett:
Why would `m_proc_hdl` be NULL and ever reach this code? I'm curious.

kcuzner:
I will find that out, that is a good question. It did happen and this fix eliminated the warning. Would seem like the cases tested before calling this should have eliminated that possibility.

kcuzner:
I found that the *CbHdl causing this issue was in fact the startup callback, which makes sense because it doesn't ever allocate a handle. I've changed this PR to instead provide an overridden version of `cleanup_callback` which cleans up the startup (and shutdown) callbacks appropriately.

ktbarrett:
@kcuzner Excellent work!

ktbarrett (approved):

ktbarrett:
@kcuzner Don't worry about the Windows/conda test failing; it's been doing that. We've "fixed" the same problem probably 4 times now and they keep breaking it. We are dropping support for that environment in the next release.

Issue #3270, opened 2023-03-21
Nested context managers not unscheduled correctly at the end of tests.


suzizecat:
Using cocotb 1.7.2 from pip on Ubuntu 22.04 and Python 3.10.

This bug as been seen with verilator 5 and a hacked 4.212 versions. However I suspect that this is a Python-side issue. 

When ending a test while a coroutine is started from within a nested context manager, `First()` events seems not to be unscheduled properly.
This leads to a non-empty event list upon starting the next test and blows everything up (due to #3267 )

Please note that starting the `buggy` test alone will work fine and that the actual test result isn't relevant.

It is a bit tricky to makes this appear, but the following codes demonstrates the issue.
Run with `make SHOW_BUG=1`  to display the unexpected behavior.  

For understanding purpose, the goal of `MonitorChange`  is to provide a context manager which will throw if changes are unexpected on a combination of signals. This is a reduced testcase from a much bigger bench and appears quite useless in the present bug report.

The following dummy IP is a counter triggering a pulse and managing an additional `valid`  signal for the purpose of showing the issue.
```systemverilog
module trigger_counter (
    input logic i_clk,  //!
    input logic i_rst_n,  //!
    input logic i_trg,  //!
    input logic i_valid,
    input logic [7:0] i_cnt,  //!
    output logic o_pulse,  //!
    output logic o_valid
);


  logic [7:0] cnt;

  always_ff @(posedge i_clk or negedge i_rst_n) begin : p_seq_
    if (~i_rst_n) begin
      cnt <= 0;
      o_pulse <= 0;
      o_valid <= 0;
    end else begin
      o_pulse <= 0;
      o_valid <= i_valid;
      if (i_trg) begin
        cnt <= i_cnt;
      end else if (cnt != 0) begin
        cnt <= cnt - 1;
        o_pulse <= (cnt == 1) ? 1'b1 : 1'b0;
      end
    end
  end

endmodule
```
The following python demonstrate the relevant bug.
```python
from cocotb.clock import Clock
from cocotb.triggers import *
import cocotb

import os

class MonitorChange:
    def __init__(self, signal, valid,expect_change = True) -> None:
        self.monitor_process = None
        self.signal = signal
        self.valid = valid
        self.expect_change = expect_change

    def __enter__(self) :
        self.start_monitor()
    
    def __exit__(self, exc_type, exc_val, exc_tb):
        if exc_type is None :
            self.assert_change()
            return True
        else :
            self.stop_monitor()
[...]

PR #3291, opened 2023-04-30
Fix scheduler bug

### LINKED ISSUE #3267: RecursionError in _resume_coro_upon when error happen
Using :
 - Cocotb 1.7.2
 - Ubuntu 22.04 (In WSL 2.0 for Windows 10) 
 - Verilator 5.004

When an exception occurs in the scheduler `_react` function on:
```python
        if self._pending_triggers:
            raise InternalError(
                "Expected all triggers to be handled but found {}".format(
                    self._pending_triggers
                )
            )
```
With remaining triggers, during the call to the `trigger.prime` function of the `_resume_coro_upon()` :
```python
if not trigger.primed:
    # ... 
    try:
        trigger.prime(self._react)
    except Exception as e:
        # discard the trigger we associated, it will never fire
        self._trigger2coros.pop(trigger)
        # replace it with a new trigger that throws back the exception
        self._resume_coro_upon(
            coro,
            NullTrigger(
                name="Trigger.prime() Error", outcome=outcomes.Error(e)
            ),
        )
```

The call of `self._resume_coro_upon` with the `NullTrigger` will end up throwing the same `InternalError`, hence recursively throwing until Python itself explodes.

I managed to avoid the recursion error by changing the code around `InternalError`. Adding a `self._cleanup()` force the scheduler to cleanup the remaining triggers and fail in another way, after properly reporting the `InternalError`. 

The resulting code for `_react` is:
```python
def _react(self, trigger):
    """
    Called when a trigger fires.
    We ensure that we only start the event loop once, rather than
    letting it recurse.
    """
    # ... 
    if self._pending_triggers:
        self._cleanup() # <--- Force the cleanup of remaining _pending_triggers
        raise InternalError(
            "Expected all triggers to be handled but found {}".format(
                self._pending_triggers
            )
        )
```

ktbarrett:
Fixes #3267 and #3270. The shallow copy of `_triggers2coros` and the cleanup of triggers (in the `_event_loop` loop) after coros (in `_cleanup`) cause #3270. #3267 is dealt with by making `InternalError` a `BaseException` as was discussed in that issue.

### DIFF
--- a/tests/test_cases/test_3270/Makefile
+++ b/tests/test_cases/test_3270/Makefile
@@ -1,0 +1,20 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+TOPLEVEL_LANG ?= verilog
+
+ifneq ($(TOPLEVEL_LANG),verilog)
+
+sim:
+	@echo Only Verilog is supported for this test, skipping...
+
+else
+
+MODULE := test_3270
+VERILOG_SOURCES := test_3270.v
+TOPLEVEL := trigger_counter
+
+include $(shell cocotb-config --makefiles)/Makefile.sim
+
+endif
--- a/tests/test_cases/test_3270/test_3270.v
+++ b/tests/test_cases/test_3270/test_3270.v
@@ -1,0 +1,35 @@
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause
+
+module trigger_counter (
+    input logic i_clk,
+    input logic i_rst_n,
+    input logic i_trg,
+    input logic i_valid,
+    input logic [7:0] i_cnt,
+    output logic o_pulse,
+    output logic o_valid
+);
+
+
+  logic [7:0] cnt;
+
+  always @(posedge i_clk or negedge i_rst_n) begin : p_seq_
+    if (~i_rst_n) begin
+      cnt <= 0;
+      o_pulse <= 0;
+      o_valid <= 0;
+    end else begin
+      o_pulse <= 0;
+      o_valid <= i_valid;
+      if (i_trg) begin
+        cnt <= i_cnt;
+      end else if (cnt != 0) begin
+        cnt <= cnt - 1;
+        o_pulse <= (cnt == 1) ? 1'b1 : 1'b0;
+      end
+    end
+  end
+
+endmodule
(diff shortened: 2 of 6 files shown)


tests/test_cases/test_3270/Makefile
```diff
@@ -0,0 +1,19 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+TOPLEVEL_LANG ?= verilog
+
+ifneq ($(TOPLEVEL_LANG),verilog)
+
+$(info Only Verilog is supported for this test)
```
ktbarrett, line 9:
```suggestion
all:
    @echo Only Verilog is supported for this test, skipping...
```

tests/test_cases/test_3270/Makefile
```diff
@@ -0,0 +1,20 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+TOPLEVEL_LANG ?= verilog
+
+ifneq ($(TOPLEVEL_LANG),verilog)
+
+sim:
+    @echo Only Verilog is supported for this test, skipping...
```
ktbarrett, line 10:
```suggestion
	@echo Only Verilog is supported for this test, skipping...
```

tests/test_cases/test_3270/test_3270.v
```diff
@@ -0,0 +1,35 @@
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause
+
+module trigger_counter (
+    input logic i_clk,  //!
```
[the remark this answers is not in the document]
ktbarrett: Nope, just a copy-paste of the test from #3267.

ktbarrett:
@cmarqu What should the newsfragment say?

cmarqu:
> @cmarqu What should the newsfragment say?

How about "Fix trigger cleanup that, in some cases, resulted in a `RecursionError`"?

ktbarrett:
They are separate issues. The RecursionError is caused by a *potentially large* set of corner cases that I don't understand well. And the cleanup prevents a *potentially large* set of corner cases that I don't understand well. I can't really be exact. I think the best I could do is "Fixed bugs in scheduler triggers by certain corner cases", but I'm not sure that's valuable for general consumption.

cmarqu:
My thought was that someone who encountered a `RecursionError` in the past (but didn't follow  up) might read the release notes and would give 1.8 another try based on seeing that term.
But I don't have strong opinions here...

PR #3294, opened 2023-05-05
Update simple_dff example to assert values on the rising edge. Fixes #3286

### LINKED ISSUE #3286: The `test_dff.py` example does not fail for a non-working DFF
The `test_dff.py` example in `examples/simple_dff/` does not fail for a non-working DFF. Changing `dff.sv`
to just pass through `d` without registering it still passes the test:

```verilog
module dff (
  input logic clk, d,
  output logic q
);

assign q = d;

endmodule
```

As this is obviously not a flip fop, it should fail, but it still passes:

```
     0.00ns INFO     cocotb.regression                  Found test test_dff.dff_simple_test
     0.00ns INFO     cocotb.regression                  running dff_simple_test (1/1)
                                                          Test that d propagates to q
105000.00ns INFO     cocotb.regression                  dff_simple_test passed
105000.00ns INFO     cocotb.regression                  **************************************************************************************
                                                        ** TEST                          STATUS  SIM TIME (ns)  REAL TIME (s)  RATIO (ns/s) **
                                                        **************************************************************************************
                                                        ** test_dff.dff_simple_test       PASS      105000.00           0.00   49886942.02  **
                                                        **************************************************************************************
                                                        ** TESTS=1 PASS=1 FAIL=0 SKIP=0             105000.00           0.23     455049.98  **
                                                        **************************************************************************************
```

- cocotb version: 1.7.2
- OS: macOS Monterey (12.6.5) (Intel 64-bit)
- Simulator: Icarus Verilog version 11.0
- Python version: 3.11.3, installed via [Homebrew](https://brew.sh)

ddribin:
See discussion in Issue #3286. If this looks good, I can update the example in `README.md`, too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,22 +68,36 @@ An example of a simple randomized cocotb testbench:
 # test_dff.py
 
 import random
+
 import cocotb
 from cocotb.clock import Clock
-from cocotb.triggers import FallingEdge
+from cocotb.triggers import RisingEdge
+from cocotb.types import LogicArray
 
 @cocotb.test()
-async def test_dff_simple(dut):
-    """ Test that d propagates to q """
+async def dff_simple_test(dut):
+    """Test that d propagates to q"""
+
+    # Set initial input value to prevent it from floating
+    dut.d.setimmediatevalue(1)
+    # Assert initial output is unknown
+    assert LogicArray(dut.q.value) == LogicArray("X")
 
     clock = Clock(dut.clk, 10, units="us")  # Create a 10us period clock on port clk
     cocotb.start_soon(clock.start())  # Start the clock
 
+    await RisingEdge(dut.clk)  # Synchronize with the clock
+    expected_val = 1  # Matches initial input value
     for i in range(10):
         val = random.randint(0, 1)
         dut.d.value = val  # Assign the random value val to the input port d
-        await FallingEdge(dut.clk)
-        assert dut.q.value == val, "output q was incorrect on the {}th cycle".format(i)
+        await RisingEdge(dut.clk)
+        assert dut.q.value == expected_val, f"output q was incorrect on the {i}th cycle"
+        expected_val = val
+
+    # Check the final input on the next clock
+    await RisingEdge(dut.clk)  # Synchronize with the clock
+    assert dut.q.value == expected_val, "output q was incorrect on the last cycle"
 ```
 
 A simple Makefile:
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ async def dff_simple_test(dut):
     # Assert initial output is unknown
     assert LogicArray(dut.q.value) == LogicArray("X")
     # Set initial input value to prevent it from floating
-    dut.d.setimmediatevalue(0)
+    dut.d.value = 0
 
     clock = Clock(dut.clk, 10, units="us")  # Create a 10us period clock on port clk
     # Start the clock. Start it low to avoid issues on the first RisingEdge
--- a/README.md
+++ b/README.md
@@ -78,25 +78,27 @@ from cocotb.types import LogicArray
 async def dff_simple_test(dut):
     """Test that d propagates to q"""
 
-    # Set initial input value to prevent it from floating
-    dut.d.setimmediatevalue(1)
     # Assert initial output is unknown
     assert LogicArray(dut.q.value) == LogicArray("X")
+    # Set initial input value to prevent it from floating
+    dut.d.setimmediatevalue(0)
 
     clock = Clock(dut.clk, 10, units="us")  # Create a 10us period clock on port clk
-    cocotb.start_soon(clock.start())  # Start the clock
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -68,22 +68,36 @@ An example of a simple randomized cocotb testbench:
...
 @cocotb.test()
-async def test_dff_simple(dut):
-    """ Test that d propagates to q """
+async def dff_simple_test(dut):
+    """Test that d propagates to q"""
+
+    # Set initial input value to prevent it from floating
+    dut.d.setimmediatevalue(1)
+    # Assert initial output is unknown
+    assert LogicArray(dut.q.value) == LogicArray("X")
```
ddribin, line 84: Could potentially leave this out of the `README.md` to make the example more concise?

ktbarrett: I think it's good as is.

ddribin:
I needed to install GHDL with `brew install --cask ghdl`, but I was able to reproduce locally with:

```
SIM_ARGS=--vcd=dff.vcd make TOPLEVEL_LANG=vhdl SIM=ghdl
```

Looks like it is a difference in simulation between Icarus and GHDL. In Icarus, the first `RisingEdge` registers `0`, but in GHDL, the first `RisingEdge` registers `U`. (I assume this is like `X`? I don't know VHDL.)

I am able to work around this by doing two `RisingEdge` at the start:

```python
    await RisingEdge(dut.clk)  # Synchronize with the clock
    await RisingEdge(dut.clk)
```

Here's the waveforms GHDL in this case:

vhdl-start-high

Compare this to the waveforms for Icarus:

sv-start-high

I'm guessing there's a race condition/undefined behavior between setting `d = 0`, starting the clock high, and `RisingEdge` at time 0? And the two simulators choose different (but valid) behaviors? I would have thought using `dut.d.setimmediatevalue(0)` would have set `d` "before" time 0 and the initial `RisingEdge`, but I guess not?

An alternative workaround is to only do one `RisingEdge` but start the clock low with:

```python
    cocotb.start_soon(clock.start(start_high=False))  # Start the clock
```

Then I get these waveforms for GHDL:

vhdl-start-low

And these waveforms for Icarus:

sv-start-low

@ktbarrett: Do you have a preference between two initial `RisingEdge` calls or starting the clock low?

ktbarrett:
@ddribin In VHDL `'U'` is "uninitialized", its what all `std_logic` signals start as if they aren't initialized. `std_logic` also supports `'X'` which means "unknown". Verilog's `logic` type uses `'X'` for both uninitialized and unknown.

I think the difference in behavior is how `RisingEdge` in cocotb (and the VHPI and FLI) are different than the semantics of `rising_edge` in VHDL. In Verilog, and with `RisingEdge`, the rising edge occurs when the signal changes to a "high" value, so transitions from `'X'` to `'1'` are considered clock edges. While with VHDL's `rising_edge`, a rising edge only occurs when the signal changes from a "low" value to a "high" value, so  `'X'` to `'1'` is *not* a clock edge. This is why you don't see `q` get a value at time 0 in the VHDL simulation.

I think the correct course of action is to use `start_high=False` and modifying the test to accommodate that if needed.

examples/simple_dff/dff.sv
```diff
@@ -12,4 +12,12 @@ always @(posedge clk) begin
...
 end
 
+// the "macro" to dump signals
+`ifdef COCOTB_SIM
+initial begin
+  $dumpfile ("dff.vcd");
+  $dumpvars (0, dff);
+  #1;
+end
+`endif
```
ddribin, line 22: @ktbarrett: Is it okay to leave this in here? Or do you think PR https://github.com/cocotb/cocotb/pull/2239, allowing `WAVES` to work with Icarus is going to land in the next version?

examples/simple_dff/test_dff.py
```diff
@@ -8,22 +8,36 @@
...
 
 
 @cocotb.test()
 async def dff_simple_test(dut):
     """Test that d propagates to q"""
 
+    # Assert initial output is unknown
+    assert LogicArray(dut.q.value) == LogicArray("X")
+    # Set initial input value to prevent it from floating
+    dut.d.setimmediatevalue(0)
```
ddribin, line 22: @ktbarrett: Is it worth a comment on `setimmediatevalue(0)` vs. `value = 0` ? I don't think the difference matters here anymore, now that it is set before the first `RisingEdge`, so this could use `value = 0` for simplicity.

examples/simple_dff/dff.sv
ktbarrett: Seeing as #2239 has had no activity in 2+ years, it probably won't. Can we remove the ifdef?

examples/simple_dff/test_dff.py
ktbarrett: Yeah the difference doesn't matter anymore. I might prefer using `signal.value = 0` in that case to keep it simple.

ktbarrett (approved):

Issue #3254, opened 2023-02-21
Verilator: Support `--timing` mode


marlonjames:
Verilator has support for [timing constructs](https://verilator.org/guide/latest/exe_verilator.html#cmdoption-timing) via the `--timing` argument.

When using `--timing`, the wrapper needs to call `eventsPending()` and `nextTimeSlot()` in order to not miss a time step that needs to be evaluated.
Details here: https://verilator.org/guide/latest/connecting.html#wrappers-and-model-evaluation-loop

imphil:
I wonder if we can use the auto-generated C++ wrapper with Verilator now, which would hopefully give us all these things for free (and keep the updated).

marlonjames:
I don't think that the generated wrapper has support for any of the callbacks for VPI applications.

Issue #10, opened 2013-06-18
Finer grained control over logging verbosity


chiggs:
We globally set the loglevel in cocotb.**init**.py:

> log.setLevel(logging.INFO)

We've split the messages hierarchically using the standard python logging mechanism.  It would be useful to provide and document a more convenient way to set up, for example all of the following can be different verbosity:

scheduler
drivers
monitors
scoreboard
coroutines

ktbarrett:
This can be accomplished using the standard Python logging facilities now.

For example:
```python
import logging
logging.getLogger("cocotb.regression").setLevel(logging.WARNING)
```

PR #3316, opened 2023-05-24
Patch support for --timing flag in verilator

mczyz-antmicro:
This PR adds a fix to error "Simulator shut down prematurely" when cocotb is used with Verilator and `--timing` parameter. The proposed fix is related to issue #3254

If the HDL code contains timing statements, e.g. `#10 clk = ~clk;`, then the Verilator should be called with a `--timing` flag. In current implementation Verilator shuts down prematurely, beacuse the call to `VerilatedVpi::cbNextDeadline()` returns uint64 max, meaning no more scheduled events (`cbNextDeadline()` is the deadline of the next registered VPI callback). If we want to support the Verilator's `--timing` flag, then a check must be performed to see if there are any events pending `top->eventsPending()` and then find the next time slot `top->nextTimeSlot()`. The next time step is then the smaller of time steps returned by `VerilatedVpi::cbNextDeadline()` and `top->nextTimeSlot()`

### DIFF
--- a/cocotb/share/lib/verilator/verilator.cpp
+++ b/cocotb/share/lib/verilator/verilator.cpp
@@ -111,7 +111,10 @@ int main(int argc, char** argv) {
 #endif
         // cocotb controls the clock inputs using cbAfterDelay so
         // skip ahead to the next registered callback
-        vluint64_t next_time = VerilatedVpi::cbNextDeadline();
+        vluint64_t next_time_cocotb = VerilatedVpi::cbNextDeadline();
+        vluint64_t next_time_timing =
+            top->eventsPending() ? top->nextTimeSlot() : (vluint64_t)~0ULL;
+        vluint64_t next_time = std::min(next_time_cocotb, next_time_timing);
 
         // If there are no more cbAfterDelay callbacks,
         // the next deadline is max value, so end the simulation now
--- a/cocotb/share/lib/verilator/verilator.cpp
+++ b/cocotb/share/lib/verilator/verilator.cpp
@@ -111,14 +111,15 @@ int main(int argc, char** argv) {
 #endif
         // cocotb controls the clock inputs using cbAfterDelay so
         // skip ahead to the next registered callback
+        const vluint64_t NO_TOP_EVENTS_PENDING = static_cast<vluint64_t>(~0ULL);
         vluint64_t next_time_cocotb = VerilatedVpi::cbNextDeadline();
         vluint64_t next_time_timing =
-            top->eventsPending() ? top->nextTimeSlot() : (vluint64_t)~0ULL;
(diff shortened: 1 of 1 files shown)


ktbarrett:
Yeah, they seem unrelated, which is odd since https://github.com/cocotb/cocotb/pull/3321 is passing.

ktbarrett:
So when not in `--timing` mode, what does `top->eventsPending()` and `top->nextTimeSlot()` return? What happens if cocotb "deadlocks" by waiting on an event that will never happen? We expect the simulator to detect that and exit instead of hang indefinitely. Does this solution avoid that?

ktbarrett:
I don't know enough about Verilator to say this doesn't change existing behavior. Waiting on feedback.

imphil (approved):
LGTM, thanks! Can you do the small nitpick update, @mczyz-antmicro?

> So when not in --timing mode, what does top->eventsPending() and top->nextTimeSlot() return?

* eventsPending() returns false without complaining,
* nextTimeSlot() returns 0 with a fatal error

https://github.com/verilator/verilator/blob/c2483189580e0c0b9ba87a9087d1a1032ba5c011/src/V3EmitCModel.cpp#L451-L455

mczyz-antmicro:
- @imphil Added your suggestion [in this commit](https://github.com/antmicro/cocotb/commit/a76b29655252e3821aa2a0db39046026e3431b44)
- Testcase is [on this branch](https://github.com/antmicro/cocotb/commits/mczyz/verilator-parameters)

I rebased both branches to latest master, too

cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -111,11 +111,15 @@ int main(int argc, char** argv) {
 #endif
         // cocotb controls the clock inputs using cbAfterDelay so
         // skip ahead to the next registered callback
-        vluint64_t next_time = VerilatedVpi::cbNextDeadline();
+        const vluint64_t NO_TOP_EVENTS_PENDING = static_cast<vluint64_t>(~0ULL);
```
ktbarrett, line 114: Technically this could be `constexpr`, but who cares. The compiler is going to const prop it regardless.

ktbarrett (approved):

ktbarrett:
@mczyz-antmicro The test looks good but we don't really need both the Python runner and the Makefile for the test. The only issue I see is that it would be nice to timeout and kill the hung process so the rest of the regression can complete.

imphil:
Thanks @mczyz-antmicro for your work on this PR! Thanks especially for the in-depth thoughts on testing your change in https://github.com/cocotb/cocotb/pull/3316#issuecomment-1580849641, that's exceptional and very much appreciated.

mczyz-antmicro:
I will follow-up with testcase in #3333

PR #3324, opened 2023-06-04
Update Icarus Verilog build to support WAVES environment variable, 2023 version

ddribin:
Details:
- Update Makefile.icarus to support WAVES by creating an iverilog_dump.v file that is automatically added to the build.
- Remove VCD support from examples/simple_dff/dff.sv

This is basically an updated version of PR #2239 against current top-of-tree. It generates a VCD instead of an FST. And it puts the waveform file in `sim_build/`.

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.icarus
+++ b/cocotb/share/makefiles/simulators/Makefile.icarus
@@ -69,9 +69,23 @@ ifdef VERILOG_INCLUDE_DIRS
 endif
 
 # Compilation phase
+
+ifeq ($(WAVES), 1)
+    VERILOG_SOURCES += $(SIM_BUILD)/iverilog_dump.v
+    COMPILE_ARGS += -s iverilog_dump
+endif
+
 $(SIM_BUILD)/sim.vvp: $(VERILOG_SOURCES) $(CUSTOM_COMPILE_DEPS) | $(SIM_BUILD)
 	@echo "+timescale+$(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION)" > $(SIM_BUILD)/cmds.f
 	$(CMD) -o $(SIM_BUILD)/sim.vvp -D COCOTB_SIM=1 $(TOPMODULE_ARG) $(COMPILE_ARGS) $(EXTRA_ARGS) $(VERILOG_SOURCES)
+
+$(SIM_BUILD)/iverilog_dump.v: | $(SIM_BUILD)
+	@echo 'module iverilog_dump();' > $@
+	@echo 'initial begin' >> $@
+	@echo '    $$dumpfile("$(SIM_BUILD)/$(TOPLEVEL).vcd");' >> $@
+	@echo '    $$dumpvars(0, $(TOPLEVEL));' >> $@
+	@echo 'end' >> $@
+	@echo 'endmodule' >> $@
 
 # Execution phase
 
--- a/cocotb/share/makefiles/simulators/Makefile.icarus
+++ b/cocotb/share/makefiles/simulators/Makefile.icarus
@@ -71,18 +71,19 @@ endif
 # Compilation phase
 
 ifeq ($(WAVES), 1)
-    VERILOG_SOURCES += $(SIM_BUILD)/iverilog_dump.v
-    COMPILE_ARGS += -s iverilog_dump
+    VERILOG_SOURCES += $(SIM_BUILD)/cocotb_iverilog_dump.v
+    COMPILE_ARGS += -s cocotb_iverilog_dump
+    PLUSARGS += -fst
 endif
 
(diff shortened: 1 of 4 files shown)


ktbarrett (approved):

imphil:
Are we sure we want "It generates a VCD instead of an FST."? FST can be read just fine by GtkWave, and whoever is using icarus is most likely also using a free waveform viewer. And of course it's the significantly better format, with more information being preserved and files being much smaller.

cocotb/share/makefiles/simulators/Makefile.icarus
```diff
@@ -69,10 +69,24 @@ ifdef VERILOG_INCLUDE_DIRS
...
+    VERILOG_SOURCES += $(SIM_BUILD)/iverilog_dump.v
+    COMPILE_ARGS += -s iverilog_dump
+endif
+
 $(SIM_BUILD)/sim.vvp: $(VERILOG_SOURCES) $(CUSTOM_COMPILE_DEPS) | $(SIM_BUILD)
 	@echo "+timescale+$(COCOTB_HDL_TIMEUNIT)/$(COCOTB_HDL_TIMEPRECISION)" > $(SIM_BUILD)/cmds.f
 	$(CMD) -o $(SIM_BUILD)/sim.vvp -D COCOTB_SIM=1 $(TOPMODULE_ARG) $(COMPILE_ARGS) $(EXTRA_ARGS) $(VERILOG_SOURCES)
 
+$(SIM_BUILD)/iverilog_dump.v: | $(SIM_BUILD)
+	@echo 'module iverilog_dump();' > $@
```
imphil, line 83: To reduce the risk of accidentally clashing with a user module of that name can we prefix this module with something rare? A random number, cocotb_internal_implementation_of_iverilog_dump, something of that sort?

imphil (approved):
LGTM! I'd go for FST traces unless there's a strong reason not to. And this will need a newsfragment, but I can do that after this PR goes in.

ktbarrett:
There is more tlak about adding support for FST vs VCD into WAVES at some point more globally. For now just getting the feature in there in any way is a step forward.

ddribin:
> Are we sure we want "It generates a VCD instead of an FST."? FST can be read just fine by GtkWave, and whoever is using icarus is most likely also using a free waveform viewer. And of course it's the significantly better format, with more information being preserved and files being much smaller.

Happy to do FST. I just went with VCD because it is more "standard" ? But, true, I use GTKWave, so FST would work just fine.

cocotb/share/makefiles/simulators/Makefile.icarus
ddribin: Yeah, good point. I think a `cocotb_` prefix ought to be good enough? So how about `cocotb_iverilog_dump.v` ?

ddribin:
I also should update the docs, which currently mention the need to add a code fragment to your `.v` file for Icarus.

cocotb/share/makefiles/simulators/Makefile.icarus
imphil: There might be collisions no matter what name we're choosing, so feel free to use the one you like best.

imphil:
On FST vs VCD: We have other simulators writing waveforms in their proprietary format when WAVES is set, so doing the same for Icarus seems consistent. So I'd personally go for that, but I'll leave it up to you @ddribin, it's your PR after all! Thanks for working on it!

Let me know when you're ready for a merge.

ddribin:
> On FST vs VCD: We have other simulators writing waveforms in their proprietary format when WAVES is set, so doing the same for Icarus seems consistent. So I'd personally go for that, but I'll leave it up to you @ddribin, it's your PR after all! Thanks for working on it!
> 
> Let me know when you're ready for a merge.

I was going to ask if there's precedent for `WAVES` producing formats other than VCD. Given other simulators do produce something other than VCD, I'll switch to FST, by default. We can add an override option for VCD later.

I just pushed changes for this and the documentation. Please take a look at the doc changes, but if there are no further comments, this is ready to merge.

documentation/source/simulator_support.rst
```diff
@@ -47,28 +47,15 @@ See also https://github.com/steveicarus/iverilog/issues/323.
...
-      $dumpvars (0, button_deb);
-      #1;
-    end
-    `endif
-    endmodule
+Icarus Verilog can produce waveform traces in the FST format, the native format of GTKWave.
+FST traces are much smaller and more efficient to write than VCD, but requires the use of GTKWave.
+
+To enable FST tracing, set :make:var:`WAVES` to ``1``. This can be set on the command line,
+as shown below, or as an environment variable with an ``EXPORT`` statement in your shell.
```
ktbarrett, line 54: I think this is more info than needed. 
```suggestion
To enable FST tracing, set :make:var:`WAVES` to ``1``.
```

PR #3325, opened 2023-06-06
Make tests work in non-isolated environments

imphil:
The dev_test_sim nox target installed cocotb non-editable (for coverage collection), while dev_test_nosim installed it editable. Always use a non-editable install to avoid installation issues.


imphil:
OK, with this change I now get consistent errors in private and public CI :-) Next up: figure out how not to get an error at all.

imphil:
Let's see if this passes private CI, but if so, I'd just go ahead with this stopgap and sort out the import error in a more quiet moment. These tests are not the most important one anyways.

ktbarrett (approved):

PR #3334, opened 2023-06-12
Fix importing DLLs on wine

sean-anderson-seco:
I encountered the following error in wine:

```
0034:err:module:import_dll Library cocotbutils.dll (which is needed by L"C:\\Program Files\\Python311\\Lib\\site-packages\\cocotb\\simulator.cp311-win_amd64.pyd") not found 0034:err:module:import_dll Library gpilog.dll (which is needed by L"C:\\Program Files\\Python311\\Lib\\site-packages\\cocotb\\simulator.cp311-win_amd64.pyd") not found 0034:err:module:import_dll Library gpi.dll (which is needed by L"C:\\Program Files\\Python311\\Lib\\site-packages\\cocotb\\simulator.cp311-win_amd64.pyd") not found 0034:err:module:import_dll Library pygpilog.dll (which is needed by L"C:\\Program Files\\Python311\\Lib\\site-packages\\cocotb\\simulator.cp311-win_amd64.pyd") not found Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Program Files\Python311\Scripts\cocotb-config.exe\__main__.py", line 4, in <module>
  File "C:\Program Files\Python311\Lib\site-packages\cocotb\__init__.py", line 43, in <module>
    import cocotb.handle
  File "C:\Program Files\Python311\Lib\site-packages\cocotb\handle.py", line 39, in <module>
    from cocotb import simulator
ImportError: DLL load failed while importing simulator: Module not found.
```

The fix is simple, but I'm surprised no one else has seen it.

### DIFF
--- a/cocotb/__init__.py
+++ b/cocotb/__init__.py
@@ -40,6 +40,10 @@ import warnings
 from collections.abc import Coroutine
 from typing import Dict, List, Optional, Union
 
+if os.name == "nt":
+    _libs_dir = os.path.join(os.path.dirname(__file__), "libs")
+    os.add_dll_directory(_libs_dir)
+
 import cocotb.handle
 from cocotb._deprecation import deprecated
 from cocotb.log import default_config


ktbarrett:
I'm surprised you are using wine. Most EDA tools and cocotb work better under Linux natively. Also, we have used that approach in the past, but moved away from it for reasons I don't remember off the top of my head. @themperek do you remember?

sean-anderson-seco:
I'm experimenting with Active-HDL, which is windows-only. Obviously I would prefer a native simulator, but I wanted to try it out before writing it off entirely.

---

Oddly, it seems like registered callbacks never get run. Still looking into this...

```
run -all
     -.--ns INFO     gpi                                ..\gpi\GpiCommon.cpp:101  in gpi_print_registered_impl       VHPI registered
KERNEL: Simulation has finished. There are no more test vectors to simulate.
endsim
```

imphil:
See
* https://github.com/cocotb/cocotb/commit/950f8f4aec7b3fcc18576cd14c7ce7f11acc1db3, which has a similar approach to this PR, and 
* https://github.com/cocotb/cocotb/commit/9d98a4b44f7bc5aa953fc074dbbf7918fc59a480, which then changes to something different

That being said, I wouldn't even try to get started using Wine. Either stay on Windows (e.g. in a VM if you'd like to use Active-HDL), or move to Linux. Wine gives you the worst of both worlds and you're unlikely to have fun using cocotb this way.

ktbarrett:
So the issue here is that the new solution using side-by-side assemblies is not supported by wine? Personally, I'm not a fan of the new solution: it's Microsoft black magic that requires a lot of code on our part to work correctly. I don't think I understand what was exactly wrong with the old approach (#1829) that caused us to want to change to the new approach. Without additional information, what it seems like is the MinGW Python being busted; a platform we plan on dropping support for in the next version.

imphil:
Thanks for your PR, @sean-anderson-seco! I don't think we should take this change: it provides, it seems, a workaround for something that Wine does different than native Windows. Windows support has always been painful, but we at least have zoomed in on a solution for now that has not created a ton of new user issues, aka it seems to work for the supported Windows use cases. I feel it's too risky to deviate from that status quo for the sake of supporting a rather obscure use case.

Please reopen if you feel strongly otherwise and let's discuss!

PR #3342, opened 2023-06-15
Fixed bug with Questa test_args parameters converting

vborchsh:
There is a simple bug: in case of spaces in the test_args (e.g. -L xpm) they would be converted into something like `-L\ xpm` and so on. Which is unacceptable from vsim.

There is a working example in the past cocotb_test: https://github.com/themperek/cocotb-test/blob/32b19b0813ae40cefc2b6d5bf1a1a715682fb18c/cocotb_test/simulator.py#LL615C43-L615C43

### DIFF
--- a/cocotb/runner.py
+++ b/cocotb/runner.py
@@ -608,7 +608,7 @@ class Questa(Simulator):
             + ["-gui" if self.gui else "-c"]
             + ["-onfinish", "stop" if self.gui else "exit"]
             + lib_opts
-            + [as_tcl_value(v) for v in self.test_args]
+            + self.test_args
             + [as_tcl_value(v) for v in self._get_parameter_options(self.parameters)]
             + [as_tcl_value(f"{self.hdl_toplevel_library}.{self.sim_hdl_toplevel}")]
             + [as_tcl_value(v) for v in self.plusargs]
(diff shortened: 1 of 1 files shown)


themperek:
So the intention would be `test_args` to be a list `["-L", "xpm"]`.

I agree it is a bit inconvenient and we 
- call it `split()`
- do as the PR proposes?

Any opinion?

vborchsh:
> So the intention would be test_args to be a list ["-L", "xpm"].

Btw, looks acceptable from my point. Didn't think this way. But I'd like test it first :)

themperek:
@vborchsh In case this is ok would be great to add few word into documentation about this.

vborchsh:
Ok, let's leave it as is.

ktbarrett:
@themperek Are we good?

cocotb/runner.py
```diff
@@ -226,7 +226,8 @@ def test(
                 If not set, run all testcases found in *test_module*.
                 Can be a comma-separated list.
             seed: A specific random seed to use.
-            test_args: Extra arguments for the simulator.
+            test_args: Extra arguments for the simulator. Mustn't include space symbols.
+                In case of space split command into list.
```
themperek, line 230:
```suggestion
            test_args: A list of extra arguments for the simulator.
```
Like this?

themperek (approved):

imphil:
Argh, didn't look at the commit structure and didn't do a squash :(

@vborchsh Please try to fold changes into a clean commit structure, i.e. a single commit with a proper commit message. You can use (on your local machine) git's interactive rebase for this. (No harm done this time, though, apart from a less nice git history.)

PR #3345, opened 2023-06-16
Add .readthedocs.yaml which becomes a required file by RTD

cmarqu:
See https://blog.readthedocs.com/migrate-configuration-v2/


cmarqu:
Oops, I didn't remember we already have a file `.readthedocs.yml`.

PR #3351, opened 2023-06-23
`clean` argument for runner

vborchsh:
The argument might be used for 100% pure rebuild the testbench and delete any previous results such as build artifacts, simulator libraries and so on.

Actually, simple copy-paste of [cocotb-test](https://github.com/themperek/cocotb-test/blob/32b19b0813ae40cefc2b6d5bf1a1a715682fb18c/cocotb_test/simulator.py#L1219) with minor editing, because cocotb runner knows exact path to the building directory thanks to build_dir argument.

Not sure about tests, it could be something like:

1. create `build_dir` before `runner.build()` call
2. put some file into folder
3. run `runner.build()`
4. assert file in the folder. Must be `false` for `clean` build, `true` for `non-clean` build.

Looks a bit poor, but still... Feedback is welcome :)

### DIFF
--- a/cocotb/runner.py
+++ b/cocotb/runner.py
@@ -151,6 +151,7 @@ class Simulator(abc.ABC):
         hdl_toplevel: Optional[str] = None,
         always: bool = False,
         build_dir: PathLike = "sim_build",
+        clean: bool = False,
         verbose: bool = False,
     ) -> None:
         """Build the HDL sources.
@@ -166,10 +167,14 @@ class Simulator(abc.ABC):
             hdl_toplevel: The name of the HDL toplevel module.
             always: Always run the build step.
             build_dir: Directory to run the build step in.
+            clean: Delete build_dir before building
             verbose: Enable verbose messages.
         """
 
+        self.clean: bool = clean
         self.build_dir = get_abs_path(build_dir)
+        if self.clean:
+            self.rm_build_folder(self.build_dir)
         os.makedirs(self.build_dir, exist_ok=True)
 
         # note: to avoid mutating argument defaults, we ensure that no value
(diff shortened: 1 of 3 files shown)


themperek:
I am OK with this. Maybe should also take the `COCOTB_CLEAN` environmental variable if set? 

Could we just "clean" when `always` is set?

Maybe add `runner.clean()` instead?

The question would be should we have a command like `cocotb-clean` command? Since `build_dir` can be set it will only work on the default.

vborchsh:
> I am OK with this. Maybe should also take the COCOTB_CLEAN environmental variable if set?

If I get it right, you're going to leave Makefiles flow for back compatibility only. So, this case I'd leave it as is.

> Could we just "clean" when always is set?
> Maybe add runner.clean() instead?

This is the idea I thought about. But I think, `clean` + `always` is more flexible. You can keep any libs and artifacts in the folder in case `always`. And purge _everything_ in case `clean`.

> The question would be should we have a command like cocotb-clean command? Since build_dir can be set it will only work on the default.

I used to use the command with `cocotb-test` before. And we can do the same, plus couple keys: `--recursive`, `--buildir`. In the other PR, I guess.

ktbarrett:
I think this is fine as is. Just needs a newsfragement. See instructions [here](https://github.com/cocotb/cocotb/tree/master/documentation/source/newsfragments).

vborchsh:
Are there any ideas about failed MacOS icarus test? There is a timeout in the tests but I didn't touch them... And `test_runner` is passed:

```
tests/pytest/test_runner.py::test_runner[False-parameters0] PASSED       [ 66%]
tests/pytest/test_runner.py::test_runner[False-parameters1] PASSED       [ 75%]
tests/pytest/test_runner.py::test_runner[True-parameters0] PASSED        [ 83%]
tests/pytest/test_runner.py::test_runner[True-parameters1] PASSED        [ 91%]
```

vborchsh:
Everything is merged into single commit. Thanks @imphil for advice.

ktbarrett (approved):

PR #3352, opened 2023-06-27
Use vendored fork of domaintools

### LINKED ISSUE #3314: Find a solution for sphinx domaintools bugs
Sphinx domaintools (https://github.com/sphinx-contrib/domaintools) seems to be unmaintained and has at least two issues that we stumble across regularly:

* https://github.com/sphinx-contrib/domaintools/pull/13 (patch by @eric-wieser)
* https://github.com/sphinx-contrib/domaintools/pull/7

Since these issues are rather fundamental, I'm wondering if we're close to the only user of sphinx-domaintools? What would be our alternatives?

ktbarrett:
Closes #3314. Vendors in the [sphinxcontrib-domaintools](https://github.com/sphinx-contrib/domaintools) repo with the patches mentioned in https://github.com/cocotb/cocotb/issues/3314#issue-1719488035 applied. Unfortunately I had to add the dependency to the noxfile instead of the requirements file since relative paths in requirements.txt aren't handled very well (https://github.com/pypa/pip/issues/8765).

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -513,11 +513,18 @@ def release_test_nosim(session: nox.Session) -> None:
     session.log("All tests passed!")
 
 
+def create_env_for_docs_build(session: nox.Session) -> None:
+    session.run(
+        "pip", "install", "documentation/_vendor/domaintools"
+    )  # not done in requirements.txt due to the way relative paths are handled in that file (gh-pypa/pip#8765)
+    session.run("pip", "install", "-r", "documentation/requirements.txt")
+    session.run("pip", "install", "-e", ".")
+
+
 @nox.session
 def docs(session: nox.Session) -> None:
     """invoke sphinx-build to build the HTML docs"""
-    session.run("pip", "install", "-r", "documentation/requirements.txt")
-    session.run("pip", "install", "-e", ".")
+    create_env_for_docs_build(session)
     outdir = session.cache_dir / "docs_out"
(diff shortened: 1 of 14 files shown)


noxfile.py
```diff
@@ -516,6 +516,7 @@ def release_test_nosim(session: nox.Session) -> None:
 @nox.session
 def docs(session: nox.Session) -> None:
     """invoke sphinx-build to build the HTML docs"""
+    session.run("pip", "install", "documentation/_vendor/domaintools")
```
imphil, line 519: Add a comment why this isn't part of documentation/requirements.txt below.

imphil (approved):
LGTM with comments:

* It would be nicer if you'd import an unmodified master version of domaintools in one commit and mention which SHA hash it is, and then add a second commit to patch this version to meet our needs. That makes future updates much easier.
* Add a README file somehwere in the _vendor tree to indicate what we're vendoring (including the upstream version), and what changes we made.
* I would consider stripping out all non-relevant files (leave the license, but the toxfile etc. isn't relevant). But there are good reasons to do things either way.

I tested this PR locally and it solves #3314 for me. Thanks!

imphil:
Thanks a lot for going the extra mile here, @ktbarrett, and restructuring the commits. I'm sure we'll appreciate that if we update domaintools again in the future.

PR #3355, opened 2023-06-29
cocotb-clean CLI command

vborchsh:
As mentioned [here](https://github.com/cocotb/cocotb/pull/3351#issuecomment-1604511354), I've added `cocotb-clean` CLI command. `cocotb-clean` makes possible removing build folder even recursively. There is also simple test with `subprocess` lib.

Partially taken from @themperek [cocotb-test](https://github.com/themperek/cocotb-test). So, is it enough to have default cocotb license header here?

~~PS. I'll proceed after #3351. Now, I'm just wanna run CI to check possible issues.~~

### DIFF
--- a/cocotb/clean.py
+++ b/cocotb/clean.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Clean cocotb build directories with CLI"""
+
+import argparse
+import os
+import shutil
+from pathlib import Path
+
+
+def get_parser():
+    parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument(
+        "-r",
+        "--recursive",
+        default=False,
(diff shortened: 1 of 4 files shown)


ktbarrett:
I'm not sure we need it to be its own command. Anything like this should be a subcommand of cocotb-runner.

vborchsh:
In general, is it good idea to have this functionality in `cocotb`? I found its useful.

If yes, I bring couple ideas. Main point here is CLI command to delete folders and clean workspace:

1. Move this to `cocotb-config` command. Something like `cocotb-config --clean --recursive`;
2. There is no `cocotb-runner` command, so we can do something like `cocotb-runner + <some keys>` command. With possible extension of functionality.

What do you think?

themperek:
@vborchsh Look at this at https://github.com/cocotb/cocotb/pull/2897/files The plan would be to have this in at some point. So this could ba part of it.

vborchsh:
@themperek thanks, I see now. So, I'll have a look. And probably reopen the PR.

vborchsh:
@ktbarrett anyway, I’m gonna leave it as is :) because there is a PR already, as @themperek mentioned above

PR #3360, opened 2023-07-06
Remove unnecessary objects from VPI discovery

### LINKED ISSUE #3353: Remove unnecessary VPI objects from discovery
Remove support for the discovery of initial, always, and function blocks from the VPI, and potentially other interfaces. The discovery of these blocks is of little actual value, especially seeing as variables cannot be declared in initial or always blocks and discovery of function blocks does not allow the discovery of per-instance local variables.

https://github.com/cocotb/cocotb/blob/c28a922da81996d21855568c0f9c6723d772f17c/cocotb/share/lib/vpi/VpiImpl.cpp#L138-L140

Removing these should impact exactly 0 people, but will affect the results of the discovery tests in the regressions.

ktbarrett:
Closes #3353.

Removed translation of syntactic elements like always, initial, and function blocks and modport defitions. `vpiRefObj` are aliases which require a `vpiActual` discovery to determine the underlying element. We will not translate those until we have that code in place.

Removed iteration over:
* `vpiProcess`
* `vpiModPath` (delay path)
* `vpiTchk`
* `vpiAttribute` (syntax)

Added translation of `vpiSwitch` which is a kind of primitive object like `vpiGate` and `vpiPrimTerm` for sake of completeness.

### DIFF
--- a/src/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/src/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -141,15 +141,11 @@ static gpi_objtype_t to_gpi_objtype(int32_t vpitype) {
         case vpiUnionNet:
             return GPI_STRUCTURE;
 
-        case vpiModport:
         case vpiInterface:
         case vpiModule:
-        case vpiRefObj:
         case vpiPort:
-        case vpiAlways:
-        case vpiFunction:
-        case vpiInitial:
         case vpiGate:
+        case vpiSwitch:
         case vpiPrimTerm:
         case vpiGenScope:
             return GPI_MODULE;
@@ -254,13 +250,9 @@ GpiObjHdl *VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
             break;
         case vpiModule:
         case vpiInterface:
-        case vpiModport:
-        case vpiRefObj:
         case vpiPort:
-        case vpiAlways:
-        case vpiFunction:
-        case vpiInitial:
         case vpiGate:
+        case vpiSwitch:
         case vpiPrimTerm:
         case vpiGenScope:
         case vpiGenScopeArray: {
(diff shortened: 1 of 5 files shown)


ktbarrett:
Perhaps we need to add a case where we expect the type, but throw the object away because we know it isn't valuable?

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -225,14 +228,6 @@ GpiObjHdl *VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
...
         case vpiInterface:
-        case vpiModport:
-        case vpiRefObj:
-        case vpiPort:
-        case vpiAlways:
-        case vpiFunction:
-        case vpiInitial:
-        case vpiGate:
-        case vpiPrimTerm:
         case vpiGenScope:
```
ktbarrett, line 231:
```suggestion
        case vpiGate:
        case vpiGenScope:
```

cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -246,6 +242,16 @@ GpiObjHdl *VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
             }
             break;
         }
+        case vpiModport:
+        case vpiRefObj:
+        case vpiPort:
+        case vpiAlways:
+        case vpiFunction:
+        case vpiInitial:
+        case vpiGate:
```
ktbarrett, line 251:
```suggestion
```

ktbarrett:
We can't see any signals under a gate object, so their discovery is not important. The `access_gate` test will be removed.

ktbarrett:
This is actually a bigger PITA than expected. Our "expected don't cares" are still passed back up to the GPI layer as NULLs, which get encoded as GPI_NOT_NATIVE objects, causing an error. This occurs when iterators over hierarchies are done.

ktbarrett:
The discovery tests are now not finding certain objects. Investigation is needed to find out what's now missing after removing objects from iteration.

ktbarrett:
Rerunning CI.

imphil (approved):
Very nice!

PR #3358, opened 2023-07-05
Scheduler: Fix _pending_coros cleanup

marlonjames:
Closes #3354.

Previously, `_cleanup()` would iterate over `_pending_coros` in a for loop.
Because `Task.kill()` calls `_unschedule()`, the Task was removed from `_pending_coros` while iterating over it.
Use a while loop to ensure all of the Tasks are killed and removed.

PR #3361, opened 2023-07-07
Scheduler: Allow _unschedule() to cleanup unstarted Tasks to avoid "never awaited" warnings

marlonjames:
Missed this in #3358.

For unstarted Tasks, `_unschedule()` calls `Task.close()` and removes it from `_pending_coros`. The `close()` call is necessary to avoid the warning for each coroutine object of the following form:

    RuntimeWarning: coroutine '<coroutine object>' was never awaited

### DIFF
--- a/cocotb/scheduler.py
+++ b/cocotb/scheduler.py
@@ -1095,7 +1095,6 @@ class Scheduler:
             # Get first task but leave it in the list so that _unschedule() will correctly close the unstarted coroutine object.
             task = self._pending_coros[0]
             task.kill()
-        assert not self._pending_coros
 
         if self._main_thread is not threading.current_thread():
             raise Exception("Cleanup() called outside of the main thread")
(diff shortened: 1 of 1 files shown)


cocotb/scheduler.py
```diff
@@ -1092,7 +1092,8 @@ def _cleanup(self):
         # We use a while loop because task.kill() calls _unschedule(), which will remove the task from _pending_coros.
         # If that happens a for loop will stop early and then the assert will fail.
         while self._pending_coros:
-            task = self._pending_coros.pop(0)
+            # Get first task but leave it in the list so that _unschedule() will correctly close the unstarted coroutine object.
+            task = self._pending_coros[0]
             task.kill()
         assert not self._pending_coros
```
ktbarrett, line 1098: This line is kinda pointless now.

ktbarrett (approved):

PR #3365, opened 2023-07-12
Create a new ValueChange() trigger, deprecate the Edge() trigger

imphil:
Add a new trigger, `ValueChange()`, which behaves identically to the
existing `Edge()` trigger. However, `Edge()` is confusingly named when
considering multi-bit signals (vectors): the SystemVerilog keyword
`edge` only considers the LSB of a multi-bit signal, while cocotb
considers changes to all bits.
    
Introduce a new, better-named trigger, `ValueChange()`, and deprecate
the old one. Add a note to the documentation discussing the issue

- Better document RisingEdge() and FallingEdge() triggers
- Make test_edge_on_vector() more robust
- Add ValueChange() trigger, deprecate Edge() trigger

### DIFF
--- a/src/cocotb/triggers.py
+++ b/src/cocotb/triggers.py
@@ -418,8 +418,14 @@ class FallingEdge(_EdgeBase):
         return signal
 
 
-class Edge(_EdgeBase):
-    """Fires on any value change of *signal*."""
+class ValueChange(_EdgeBase):
+    """
+    Fires on any value change of *signal*.
+
+    *signal* may have any width.
+
+    .. versionadded:: 1.9.0
+    """
 
     __slots__ = ()
     _edge_type = 3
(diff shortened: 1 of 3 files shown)


imphil:
Notes:
* This will a need a newsfragment once we're happy with the change.
* I tried hard to test the deprecation warning or place it somewhere reliable, but failed: somehow the ParametrizedSingleton is taking over in a way that I don't fully understand, and honestly I gave up on it for this small inconvenience.

imphil:
(I'll look at the remaining test failures after we have agreement on moving forward with this change, so please feel free to jump in with your thoughts!)

ktbarrett:
I'm not sure I see the value here, it's just changing a name. We aren't beholden to mimicing or avoiding the naming or behavior of any particular HDL language.

imphil:
Indeed, we don't have to follow any naming conventions: we can choose the most descriptive name we can find. And I'd claim "Edge()" doesn't meet that test.

The name `Edge()` for a trigger that fires on "any value change" is confusing for at least three reasons:
* An "edge" on a single-bit signal is a well-understood concept in logic design. An "edge" on a multi-bit signal is not.
* The name "edge" matches a SystemVerilog trigger, but it behaves differently.
* We have RisingEdge() and FallingEdge(), which operate on a 0->1 and 1->0 transition only. Edge, which looks at first sight like RisingEdge OR FallingEdge, however, does something else.

Taken together, the current naming leaves users who know SystemVerilog's `edge` keyword misinterpreting our `Edge()` trigger, and people who do not know it are left confused as to what an Edge on a multi-bit signal might be.

We can avoid all that by choosing a descriptive name like `ValueChange()`. An easy win for usability!

All of that is not theoretical: I opened this PR after concluding a longer discussion on the exact behavior of these triggers.

cocotb/triggers.py
```diff
@@ -407,26 +407,77 @@ def __repr__(self):
...
+    """Deprecated alias of :class:`cocotb.triggers.ValueChange`.
+
+    .. note::
+
+        ``Edge()`` in cocotb and ``edge`` in SystemVerilog do different things!
+
+        Cocotb's ``Edge()`` trigger considers changes to *all* bits the signal.
+        The SystemVerilog ``edge`` keyword only considers the least significant
+        bit, or in other words, ``edge`` in SystemVerilog is equivalent to
+        ``posedge`` or ``negedge``.
```
cmarqu, line 454:
```suggestion
        (``posedge`` or ``negedge``).
```

cmarqu (approved):

cocotb/triggers.py
```diff
@@ -407,26 +407,77 @@ def __repr__(self):
...
+
+        Hence, the SystemVerilog statement ``@(edge my_multibit_signal)`` is
+        equivalent to
+        ``await First(RisingEdge(my_multibit_signal), FallingEdge(my_multibit_signal))``
+        in cocotb.
+
+    .. deprecated:: 1.9.0
```
marlonjames, line 461: ?
```suggestion
    .. deprecated:: 1.8.2
```

ktbarrett: I think what @imphil meant was doing a 1.9 release off the stable/1.8 branch which is just deprecations and housekeeping.

ktbarrett (approved):
Rebase and figure out 1.8.2 vs 1.9. I don't mind either way.

ktbarrett:
I'll get to this tonight.

src/cocotb/triggers.py
```diff
@@ -423,6 +437,43 @@ def __singleton_key__(cls, signal):
...
+        reflects what this trigger does.
+    """
+
+    def __init__(self, signal):
+        # The deprecation warning in here is best-effort: If the same signal
+        # was used previously with a different _EdgeBase subclass __init__()
+        # won't be called again and no deprecation warning is shown (triggers
+        # are singletons, and all _EdgeBase triggers share the same cache).
+        warnings.warn(
+            "Event() is a deprecated alias for ValueChange(). Please update "
```
ktbarrett, line 468:
```suggestion
            "Edge() is a deprecated alias for ValueChange(). Please update "
```

ktbarrett:
Superseded by #4382.

PR #3366, opened 2023-07-12
Build a live-preview of the documentation

imphil:
Run `nox -s doc_preview` to get the documentation served on a HTTP port.
Whenever making a change to the documentation sources or the API doc,
the documentation will rebuild and automatically refresh in the browser,
no manual reload is necessary. This improves productivity a lot when
editing documentation.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -534,6 +534,28 @@ def docs(session: nox.Session) -> None:
 
 
 @nox.session
+def docs_preview(session: nox.Session) -> None:
+    """Build a live preview of the documentation"""
+    create_env_for_docs_build(session)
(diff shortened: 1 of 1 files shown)


cmarqu:
Maybe also add a note about this to the `CONTRIBUTING.md`?

imphil:
> Maybe also add a note about this to the CONTRIBUTING.md?

I'd give it a bit of baking time first to understand better where this setup fails to work (it looks all nice so far, however).

imphil:
RTD is failing due to #3367, we can ignore this failure here.

cmarqu (approved):

PR #3373, opened 2023-07-15
Update test expectations for Xcelium 23.03

imphil:
Update the test expectations for Xcelium 23.03 to make the TOPLEVEL_LANG=verilog tests pass. Issues are filed for the two new failures, both of which need upstream work most likely.

- Xcelium: Update test expectations for array access
- Xcelium: Skip mixed-language discovery tests


imphil:
`test_timing_triggers.test_readwrite_in_readonly` is the last test that's failing (I hope), I'll dig into that later.

ktbarrett:
> `test_timing_triggers.test_readwrite_in_readonly` is the last test that's failing (I hope), I'll dig into that later.

This test is essentially testing the simulator and not cocotb. We should be tracking our current state and not allowing illegal transitions.

ktbarrett (approved):

PR #3378, opened 2023-07-19
Update cocotb-config to produce unix-style paths

tj-scherer:
These changes should allow anyone to use Cygwin or Git Bash (MINGW64) to run cocotb tests. I think this adds a lot of value, because setting up conda with msys is a hurdle many of my colleagues were uninterested in tackling.

Using `$(realpath` does not produce consistent behavior between these two platforms. Since the only paths used come from `cocotb-config`, we should be able to make sure that produces a path that anyone will be happy with.

### DIFF
--- a/cocotb/config.py
+++ b/cocotb/config.py
@@ -51,9 +51,13 @@ import cocotb
 __all__ = ["share_dir", "makefiles_dir", "libs_dir"]
 
 
-share_dir = os.path.join(os.path.dirname(cocotb.__file__), "share").replace("\\", "/")
-makefiles_dir = os.path.join(os.path.dirname(cocotb.__file__), "share", "makefiles").replace("\\", "/")
-libs_dir = os.path.join(os.path.dirname(cocotb.__file__), "libs").replace("\\", "/")
+share_dir = os.path.join(os.path.dirname(cocotb.__file__), "share")
+makefiles_dir = os.path.join(os.path.dirname(cocotb.__file__), "share", "makefiles")
+libs_dir = os.path.join(os.path.dirname(cocotb.__file__), "libs")
+
+share_dir = share_dir.replace("\\", "/")
+makefiles_dir = makefiles_dir.replace("\\", "/")
+libs_dir = libs_dir.replace("\\", "/")
 
 
 def help_vars_text():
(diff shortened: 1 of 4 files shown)


ktbarrett:
Straight up replacing all backslashes with forward slashes is going to have tons of unintended consequences. You'll have to accomplish this some other way.

cocotb/config.py
```diff
@@ -43,21 +43,18 @@
...
 
-
-share_dir = os.path.join(os.path.dirname(cocotb.__file__), "share")
-makefiles_dir = os.path.join(os.path.dirname(cocotb.__file__), "share", "makefiles")
-libs_dir = os.path.join(os.path.dirname(cocotb.__file__), "libs")
-
-# On Windows use mixed mode "c:/a/b/c" as this work in all cases
-if os.name == "nt":
-    libs_dir = libs_dir.replace("\\", "/")
+base_dir = Path(cocotb.__file__).joinpath("..").resolve()
```
themperek, line 54:
```suggestion
base_dir = Path(cocotb.__file__).parent.resolve()
```

cocotb/config.py
```diff
@@ -205,9 +198,9 @@ def __call__(self, parser, args, values, option_string=None):
 
 
 def get_parser():
-    prefix_dir = os.path.dirname(os.path.dirname(cocotb.__file__))
+    prefix_dir = base_dir.joinpath("..").resolve().as_posix()
```
themperek, line 201:
```suggestion
    prefix_dir = base_dir.parent.resolve().as_posix()
```

ktbarrett:
Does this work in "normal" Windows environments? Going forward we are going to de-emphasize msys2, cygwin, and similar "Unix for Windows" environments as they all have serious issues in our experience. Does this change not break them? It still seems like there needs to be conditionals and that Cygwin, msys2, etc. need to get better at being able to distinguish them from non-Unix environments on Windows.

themperek:
It looks like this patch should help see: https://github.com/cocotb/cocotb/pull/3392

Maybe should get https://github.com/cocotb/cocotb/pull/3392 first, rebase and see?

ktbarrett:
#3392 will prove this PR works, not that this PR doesn't break non-Unix environments on Windows. We need to set one of those up.

themperek:
@ktbarrett You mean one that do not use make (only python runner)?

ktbarrett:
That's good to hear. Setting up the test shouldn't take very long and we can merge this afterwards. You are good here and we just need to do some stuff on our end and will merge when we can. Thanks for the work!

themperek:
Something for another PR (maybe). Maybe we could replace all Unix commands in the Makefile (like `tr`) with Python so there is only make needed?

ktbarrett:
If we are using Make, we can assume Unix. Unix commands will be quicker than invoking Python.

ktbarrett:
@themperek I mean does `cocotb-config --lib-dir` return a valid Windows path that people running custom flows can use?

themperek:
> I mean does cocotb-config --lib-dir and friends return a valid Windows path that people running custom flows can use?
 
How would you know this?

@ktbarrett  This PR makes things better for us (no errors in CI). Is it ok to go?

ktbarrett:
Re-running to get that new test.

PR #3344, opened 2023-06-16
Questa/ModelSim and Riviera pre_cmd test argument

vborchsh:
This feature extremely useful at least for me :) A similiar variable exist in Makefile flow: `SCRIPT_FILE`. `pre_cmd` makes possible:

**1. Save coverage file.**

```
runner.test(
        ...
        pre_cmd=["coverage save -onexit covres.ucdb;"],
        ...
    )
```

Which is impossible (AFAIK) in the cocotb + Questa/Modelsim suite in batch mode at the moment.

**2. Set up waveforms if restart!**

```
runner.test(
        ...
        pre_cmd=["do wave.do;"],
        ...
    )
```

**3. For sure, we can mix the things**

```
runner.test(
        ...
        pre_cmd=["coverage save -onexit covres.ucdb; do wave.do;"],
        ...
    )
```

I also added this command to the Riviera flow, but I have no chance to test it.

PR #3379, opened 2023-07-19
Add GUI_SCRIPT_FILE variable for Questa/Riviera

### LINKED ISSUE #3029: Modelsim wave.do SCRIPT_FILE runs before signals are defined
For Questa/Modelsim, the SCRIPT_FILE is currently run before the simulation is initialized using the vsim command

In the common case where SCRIPT_FILE=wave.do, a file that sets up traces in the GUI, these traces fail to initialize since the underlying signals don't exist yet.

A simple fix is to execute the script file after the simulation has been initialized, but I don't know if this will have side effects. Are there any cases where the SCRIPT_FILE would need to be executed before the vsim command?

timothyscherer:
Fixes #3029 

Also distinguishes the difference between the variables in the docs.

Although I personally have no use for the regular SCRIPT_FILE, I understand why we want to maintain backwards compatibility. If there is no specific application for the SCRIPT_FILE variable, maybe it should be deprecated at some point. I don't see what you would do with a script prior to loading the design (with `vsim` or `asim`).

### DIFF
--- a/cocotb/share/makefiles/simulators/Makefile.questa
+++ b/cocotb/share/makefiles/simulators/Makefile.questa
@@ -145,6 +145,9 @@ ifeq ($(WAVES),1)
 endif
 ifeq ($(GUI),1)
 	@echo "add log -r *" >> $@
+ifdef GUI_SCRIPT_FILE
+	@echo "do $(GUI_SCRIPT_FILE)" >> $@
+endif
 else
 	@echo "onbreak resume" >> $@
 	@echo "run -all" >> $@
--- a/cocotb/share/makefiles/simulators/Makefile.riviera
+++ b/cocotb/share/makefiles/simulators/Makefile.riviera
@@ -135,6 +135,9 @@ ifeq ($(WAVES),1)
 endif
 ifeq ($(GUI),1)
 	@echo "wave -rec *" >> $@
+ifdef GUI_SCRIPT_FILE
+	@echo "do $(GUI_SCRIPT_FILE)" >> $@
+endif
 else
 	@echo "run -all" >> $@
 	@echo "endsim" >> $@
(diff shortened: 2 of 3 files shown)


ktbarrett:
Not sure why we need a separate variable for GUI vs non-GUI.

tj-scherer:
The conclusion from #3029 was that a new variable is needed to accomplish this. The variable is needed to set up the wave viewer window. That is not needed when the GUI is not running.

If it were up to me, I would just use SCRIPT_FILE the same way #3030 did.

The documentation is just wrong here... this is a feature specific to Questa/Riviera, and it doesn't even work for what it describes. This pull request fixes that.

ktbarrett:
Ah. So the issue is whether this should be run before or after the call to vsim. #3344 seems to think that running before is just fine and wants to see the same supported in the Python runner. There is enough opinion about this that I'm tempted to remove SCRIPT_FILE rather than enable more quibbling.

tj-scherer:
It seems that #3344 provides a valid use case for SCRIPT_FILE (although they are not using the Makefile structure).

That use case does not include setting up waveforms, as described [here](https://docs.cocotb.org/en/stable/building.html#envvar-SCRIPT_FILE). Waveforms cannot be configured before vsim, only after, hence the addition of a variable that performs the function currently described in the documentation.

If coverage is configured prior to vsim (haven't gotten that far yet), I think this is a good approach to enabling both.

vborchsh:
> It seems that https://github.com/cocotb/cocotb/pull/3344 provides a valid use case for SCRIPT_FILE (although they are not using the Makefile structure).
> ...
> That use case does not include setting up waveforms, as described [here](https://docs.cocotb.org/en/stable building.html#envvar-SCRIPT_FILE)
> ...

Actually, it does. You can put `"do wave.do"` in `pre_cmd` list to setup waveforms from `wave.do` file.

themperek:
Thank you for the effort. 
If it is fine, I close it. If you still need it please reopen but I think we do not want to add any new "features" to Makefiles anymore.

PR #3380, opened 2023-07-19
Questa prevent removing library without sources

tj-scherer:
If Questa is not rebuilding RTL_LIBRARY (in my case, it is being built in another make target which is part of COMPILE_DEPS) there is no reason to delete it.


ktbarrett:
This is the point of custom flows. If what's there doesn't suit you, do it yourself; and feel free to copy and paste what cocotb provides, it is provided under a permissive license. We aren't going to support everyone's special requirements, especially wrt build flows and project automation; that is not in our scope.

I vote to close. @cocotb/maintainers

tj-scherer:
The Riviera-PRO makefile does not use vdel at all (although it does support it). I think this is at least inconsistent behavior that should be reviewed by someone more familiar with that tool. This makes me think it may not even be necessary for questa (or it needs to be added to riviera).

https://www.aldec.com/en/support/resources/documentation/articles/1520
https://www.microsemi.com/document-portal/doc_view/136364-modelsim-me-10-4c-command-reference-manual-for-libero-soc-v11-7 pg. 320

themperek:
>I vote to close. https://github.com/orgs/cocotb/teams/maintainers

I agree.

PR #3385, opened 2023-07-28
Add support for timescale and wave dumping in Icarus Verilog runner

rw1nkler:
This PR adds support for timescale and wave dumping in Icarus Verilog Runner. The change is proposed because the current runner implementation does not handle those options, but they exist for the make flow.

The changes included in this PR:
* `waves` argument added to `Simulator.build()` function
* `cocotb_hdl_timeunit` and `cocotb_hdl_timeprecision` arguments added to `Simulator.build()` function
* timescale and waveform dumping handled in a similar way as in the make flow (see [[LINK]](https://github.com/cocotb/cocotb/blob/8b9a50af956aaba52a12e760af7c064237c8aca7/cocotb/share/makefiles/simulators/Makefile.icarus#L83), [[LINK]](https://github.com/cocotb/cocotb/blob/8b9a50af956aaba52a12e760af7c064237c8aca7/cocotb/share/makefiles/simulators/Makefile.icarus#L83))

During working on adding this option I noticed that the runner flow uses absolute paths when referring to sources, while the make flow uses relative paths.


rw1nkler:
I believe that the current failures in the CI are not related to the introduced changes.

ktbarrett:
The icarus test timing out is odd. I'm gonna try re-running.

themperek:
@rw1nkler Something that was discussed a while ago. 

What do you think of using `waves=name.vcd` as an argument? In this case, `waves` would be the dump file name. The extension can select the format.

This will not back-compatible.

themperek (approved):

themperek:
Maybe could check if the `timescale` resolution is set correctly. Using `cocotb.utils`.

themperek:
One more thing. Maybe we should leave defaults and not overwrite them?

rw1nkler:
> @rw1nkler Something that was discussed a while ago.
> 
> What do you think of using `waves=name.vcd` as an argument? In this case, `waves` would be the dump file name. The extension can select the format.
> 
> This will not back-compatible.

I'm not sure how configurable other simulators are with regard to selecting the dumpfile format.  However, I think that it would be beneficial to have more flexibility in choosing both the name and format of the dumpfile. I have observed that crashes of Icarus Verilog can result in malformed FST files. To troubleshoot issues with my designs, I frequently have to switch to VCD to identify the problem.

rw1nkler:
> Maybe could check if the `timescale` resolution is set correctly. Using `cocotb.utils`.

I wrote two additional tests, one for checking the time precision of the simulator, and the second for verifying if the dumpfile is created.

rw1nkler:
> One more thing. Maybe we should leave defaults and not overwrite them?

I have made changes that preserve the default behavior of the Cocotb runner. However, I think that some additional changes could be useful in this scenario. Specifically, having two separate variables for time precision (`cocotb_hdl_timeprecision`) and time unit (`cocotb_hdl_timeunit`) creates a risk that the user will only set one of them. In that situation, what should be done? Should setting the time precision without setting the time unit lead to an error or selecting a default value for the time unit? If we want to preserve the default behavior of the runner, I believe that having the timescale as a tuple may be a better idea. This eliminates the risk of setting only one variable related to the timescale.

If we want to preserve the defaults, it also means that we are not fully compliant with the make flow. Because of that, I removed the test checking if the instructions generated by the Cocotb runner match with the corresponding make commands.

rw1nkler:
@ktbarrett I've rebased my changes on the latest master. Please take another look

ktbarrett (approved):

themperek:
@rw1nkler There is a error on windows can you check this: https://github.com/cocotb/cocotb/actions/runs/5864390413/job/15899362784#step:18:4236

rw1nkler:
Yes, sure

PR #3386, opened 2023-07-28
A set of Xcelium/VHPI fixes to our test suite

imphil:
Various small updates to the cocotb test suite to better support Xcelium with VHDL toplevels.

Note that many tests still fail, but at least they now fail for reasons mostly out of our control.

### DIFF
--- a/examples/adder/tests/Makefile
+++ b/examples/adder/tests/Makefile
@@ -37,6 +37,9 @@ ifeq ($(TOPLEVEL_LANG),verilog)
     VERILOG_SOURCES = $(PWD)/../hdl/adder.sv
 else ifeq ($(TOPLEVEL_LANG),vhdl)
     VHDL_SOURCES = $(PWD)/../hdl/adder.vhdl
+    ifneq ($(filter $(SIM),ius xcelium),)
+        COMPILE_ARGS += -v93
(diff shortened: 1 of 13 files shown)


marlonjames:
The VHDL 2008 LRM in section 9.2.9 NOTE 2:
> NOTE 2—The condition operator is defined for type STD_ULOGIC defined in package STD_LOGIC_1164 (see 16.7).
Conversion of a value of type STD_ULOGIC converts '1' and 'H' to TRUE and all other values to FALSE.

`matrix_multiplier` explicitly uses VHDL 2008 syntax. I would prefer to have that test fail than change it to VHDL 1993 syntax.

imphil:
On matrix_multiplier: The goal of this test is, apart from being an example, to [test the performance of BinaryValue](https://github.com/cocotb/cocotb/pull/1502). The VHDL syntax isn't part of the test's purpose, and hence whatever syntax gives us a running test should be fine, IMO.

(It's unfortunate that Xcelium doesn't support the VHDL-2018 syntax fully here, but testing language conformance isn't the job of cocotb's test suite.)

marlonjames (approved):

cmarqu (approved):

PR #3387, opened 2023-07-29
Support not using parenthesis on cocotb.test

### LINKED ISSUE #2731: Allow not using parens on cocotb.test decorator
In many (if not the majority) of cases, users don't need to pass arguments to the test decorator. In this case it would be nice if they didn't need to "call" the decorator.

```python
@cocotb.test  # no call here
async def test_stuff(dut):
    ...
```

Standard library decorator support this (like `lru_cache`), so there is precedence to support this as well.

teobiton:
Added a test case under tests/test_cases/issue_2731.

Fixes #2731

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -140,7 +140,8 @@ class external:
 class _decorator_helper(type):
     """
     Metaclass that allows a type to be constructed using decorator syntax,
-    passing the decorated function as the first argument.
+    passing the decorated function as the first argument. Supports
+    construction with or without having the type called.
 
     So:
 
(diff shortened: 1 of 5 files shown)


ktbarrett (commented):
Technically the tests don't need to do anything. These should also be integrated into the test_cocotb/test_tests.py suite. The documentation will need updating and a newsfragment (see documentation/README.md) will need to be created.

ktbarrett:
Also an easy way to set up the linting checks to run automatically whenever you make a commit is by installing pre-commit: `pip install pre-commit` and then installing it for the cocotb repo: `pre-commit install`. Now every time you commit, it will run the listing checks.

teobiton:
Hi,

I updated the test case and added the one without parenthesis to the test suite.
I also updated the documentation with a section that introduces the `@cocotb.test` decorator and created a newsfragment.

marlonjames:
You should remove the `issue_2731` test and just keep the `test_cocotb` changes.

teobiton:
Just removed the duplicated test. The linting errors should also be fixed.

ktbarrett:
For some reason, not all the tests are running. Forcing a re-run.

PR #3391, opened 2023-08-07
String as define issue

### LINKED ISSUE #4153: The Python runner is not properly escaping string parameters when used with Verilator
Cocotb 1.9.1, Ubuntu 24.04 64 but, Verilator 5.024, System python 3.12.3. 

I have SystemVerilog design with string parameter and I'm using python runner like this: 

```
    param = {
        'MEM_MACRO_NAME': 'example_mem_cell',
        'WIDTH': 16,
    }

    runner.build(
       <...>
        parameters=param,
        <...>
    )

   runner.test(<...>)
```

This launches verilator with argument `-GMEM_MACRO_NAME=example_mem_cell`, and results in error:

```
%Error: PATH_TO_FILE_REDACTED.sv: Illegal character in decimal constant: e
    8 |   parameter MEM_MACRO_NAME = "example_mem_cell",  
    
```

This is explained in [verilator documentation](https://verilator.org/guide/latest/exe_verilator.html#cmdoption-G-name):

> -G\<name\>=\<value\>
> Overwrites the given parameter of the top-level module. The value is limited to basic data literals:
> <...>
> 
> Strings
> Strings must be in double quotes (“”). 
> They must be escaped properly on the command line, e.g., as -GSTR="\\"My String\\"" or -GSTR='"My String"'.

So it seems like python runner is not passing string parameters correctly. 

I managed to get it working with     
`
        'MEM_MACRO_NAME': '\"example_mem_cell\"',
`
but python runner probably should do it automatically in this case.

vborchsh:
Closes #4153

String as define doesn't work. At least, for Questa and Icarus. I tried some tricks, but no luck. Is there a way to do it?

Thanks.

### DIFF
--- a/tests/pytest/test_runner.py
+++ b/tests/pytest/test_runner.py
@@ -10,7 +10,7 @@ import pytest
 
 import cocotb
 from cocotb.triggers import Timer
-from cocotb_tools.runner import _as_tcl_value, get_runner
+from cocotb_tools.runner import VHDL, _as_tcl_value, get_runner
 
 pytestmark = pytest.mark.simulator_required
 
@@ -31,6 +31,9 @@ def test_empty_string():
 
 def test_special_char():
     assert _as_tcl_value("Test \n end\ttest\r") == "Test\\ \\n\\ end\\\ttest\\\r"
+
+
+string_define_value = "path/to/some/(random quote)/file.wow'"
 
 
 @cocotb.test()
@@ -57,16 +60,14 @@ def test_runner(parameters, pre_cmd, clean_build):
     vhdl_gpi_interfaces = os.getenv("VHDL_GPI_INTERFACE", None)
 
     if hdl_toplevel_lang == "verilog":
-        sources = [os.path.join(tests_dir, "designs", "runner", "runner.v")]
+        sources = [os.path.join(tests_dir, "designs", "runner", "runner.sv")]
         gpi_interfaces = ["vpi"]
     else:
         sources = [os.path.join(tests_dir, "designs", "runner", "runner.vhdl")]
         gpi_interfaces = [vhdl_gpi_interfaces]
 
     runner = get_runner(sim)
-    compile_args = []
-    if sim == "xcelium":
-        compile_args = ["-v93"]
+    compile_args = [VHDL("-v93")] if sim == "xcelium" else []
 
     # Pre-make build directory and test file for clean build assertions
     build_dir = (
@@ -83,7 +84,7 @@ def test_runner(parameters, pre_cmd, clean_build):
         sources=sources,
         hdl_toplevel="runner",
         parameters=parameters,
-        defines={"DEFINE": 4},
+        defines={"DEFINE": 4, "DEFINE_STR": string_define_value},
         includes=[os.path.join(tests_dir, "designs", "basic_hierarchy_module")],
         build_args=compile_args,
         clean=clean_build,
@@ -116,7 +117,7 @@ def test_runner(parameters, pre_cmd, clean_build):
 def test_missing_libpython(monkeypatch):
     hdl_toplevel_lang = os.getenv("HDL_TOPLEVEL_LANG", "verilog")
     if hdl_toplevel_lang == "verilog":
-        hdl_sources = [os.path.join(tests_dir, "designs", "runner", "runner.v")]
+        hdl_sources = [os.path.join(tests_dir, "designs", "runner", "runner.sv")]
         gpi_interfaces = ["vpi"]
     else:
         hdl_sources = [os.path.join(tests_dir, "designs", "runner", "runner.vhdl")]
@@ -128,7 +129,7 @@ def test_missing_libpython(monkeypatch):
         WIDTH_IN="8",
         WIDTH_OUT="8",
     )
-    build_args = ["-v93"] if sim_tool == "xcelium" else []
+    build_args = [VHDL("-v93")] if sim_tool == "xcelium" else []
     build_dir = os.path.join(sim_build, "test_missing_libpython")
 
     os.makedirs(build_dir, exist_ok=True)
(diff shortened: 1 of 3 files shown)


ktbarrett:
All failing tests are due to syntax error.

vborchsh:
Ah, sorry, I was not clear. And I have missed commas :)

> All failing tests are due to syntax error.

This is true, because of wrong formatting of string define parameter for the simulators. E.g:

```
iverilog ...blabla... -DDEFINE_PATH=path/to/some/file.wow
```

This part must be like `-DDEFINE_PATH="path/to/some/file.wow"`

The same situation for Questa:

```
vlog ...blabla... +define+DEFINE_PATH=path/to/some/file.wow
```
to `+define+DEFINE_PATH="path/to/some/file.wow"`

I played a bit with combinations of lists, tuples, strings and so on. But no luck. Maybe there is a way to do it...

ktbarrett:
I think the real answer is to create a function that takes the value and serializes it appropriately and use that instead of `as_tcl_value`.

themperek (approved):

themperek:
Sorry I approve by mistake.

themperek:
Would this work `"DEFINE_PATH": "\"path/to/some/file.wow\""` ?

Lockedloop:
Removing the `as_tcl_value` completely works for me.
```
    @staticmethod
    def _get_define_options(defines: Mapping[str, object]) -> Command:
        return [
            f"+define+{name}={value}"
            for name, value in defines.items()
        ]
```
I could be missing something, but seeing that `_get_parameter_options` does the same thing in the current version, it seems to me as a quick fix. Does anyone know why was the `as_tcl_value` introduced in the first place?

ktbarrett:
@Lockedloop Good to know. `as_tcl_value` was added to serialize values in expected TCL formatting since some simulators need that, but it's obviously not appropriate here. I'll try to get around to finishing this PR off if @davekeeshan doesn't in the next couple days.

Lockedloop:
> @Lockedloop Good to know. `as_tcl_value` was added to serialize values in expected TCL formatting since some simulators need that, but it's obviously not appropriate here. I'll try to get around to finishing this PR off if @vborchsh doesn't in the next couple days.

Great, thanks a million!

vborchsh:
> @Lockedloop Good to know. `as_tcl_value` was added to serialize values in expected TCL formatting since some simulators need that, but it's obviously not appropriate here. I'll try to get around to finishing this PR off if @vborchsh doesn't in the next couple days.

Sorry, just found this thread. Completely fine for me if you finish this :)

ktbarrett:
Not sure what the Riviera and Xcelium issues are, but the icarus ones are because the one on apt is too old, I'll have to split that into a separate test and put a version ifdef on it. But I ca. See it being done soon.

cmarqu:
Xcelium appears to need more quoting. Maybe one that `as_tcl_value` has done so far...

ktbarrett:
Using `as_tcl_value` didn't help. The issue appears to be that the slashes need escaping.

ktbarrett:
\\\\\\\"But\\\\ of\\\\ course!\\\\ It\\\\ needs\\\\ to\\\\ be\\\\ `+define+DEFINE_PATH=\\\"path/to/some\\\\\\\"(random\\ quote)/file.wow\\\"`\\\\\\\"

ktbarrett:
So Xcelium doesn't seem to support spaces in defines passed on the command line no matter how many backslahes I add. This must be a limitation of their option parser. And it doesn't seem to like `$error()`? I'll just leave the problem child be.

cmarqu:
> So Xcelium doesn't seem to support spaces in defines passed on the command line no matter how many backslahes I add. This must be a limitation of their option parser. And it doesn't seem to like `$error()`? I'll just leave the problem child be.

@imphil Do you think Cadence would help us here?

imphil:
> @imphil Do you think Cadence would help us here?

Absolutely. Can you write the text you'd like me to submit as issue to Cadence? I'll then copy-paste it over.

ktbarrett:
Let's see where this is now.

ktbarrett:
I will have to split the test into two so we can run test_runner on xcelium without also running the define test as xcelium's option parser prevents us from passing strings with whitespace correctly (it will be set to expect fail).

tests/pytest/test_runner.py
```diff
@@ -57,16 +60,14 @@ def test_runner(parameters, pre_cmd, clean_build):
...
     if hdl_toplevel_lang == "verilog":
-        sources = [os.path.join(tests_dir, "designs", "runner", "runner.v")]
+        sources = [os.path.join(tests_dir, "designs", "runner", "runner.sv")]
         gpi_interfaces = ["vpi"]
     else:
         sources = [os.path.join(tests_dir, "designs", "runner", "runner.vhdl")]
         gpi_interfaces = [vhdl_gpi_interfaces]
 
     runner = get_runner(sim)
     compile_args = []
```
ktbarrett, line 70:
```suggestion
    compile_args = [VHDL("-v93")] if sim == "xcelium" else []
```

tests/pytest/test_runner.py
```diff
@@ -128,7 +129,7 @@ def test_missing_libpython(monkeypatch):
         WIDTH_IN="8",
         WIDTH_OUT="8",
     )
-    build_args = ["-v93"] if sim_tool == "xcelium" else []
+    build_args = []
```
ktbarrett, line 132:
```suggestion
    build_args = [VHDL("-v93")] if sim_tool == "xcelium" else []
```

marlonjames (approved):

vborchsh:
Well, it's became definitely more complicated, that I thought... Thanks for fix it!

ktbarrett:
@vborchsh Of course. Sorry it took so long. We finally have a good environment in place to test commercial sims arbitrarily which is what made this happen. Thanks @imphil!

Issue #3326, opened 2023-06-07
CI: MSVC/Windows/conda tests are failing with "ImportError: DLL load failed while importing _sqlite3: The specified module could not be found."


imphil:
Another version of https://github.com/cocotb/cocotb/issues/3166 and https://github.com/cocotb/cocotb/issues/3183. We somehow need to find a more sustainable solution to this mess, it's our only Windows test in CI.

```
2023-06-06T22:59:19.6260505Z nox > Running simulator-agnostic tests with pytest
2023-06-06T22:59:19.6265500Z nox > pytest -v --cov=cocotb --cov-branch --cov-report= -k 'not simulator_required'
2023-06-06T22:59:19.9500265Z Traceback (most recent call last):
2023-06-06T22:59:19.9500850Z   File "C:\Miniconda\lib\runpy.py", line 194, in _run_module_as_main
2023-06-06T22:59:19.9501647Z     return _run_code(code, main_globals, None,
2023-06-06T22:59:19.9502232Z   File "C:\Miniconda\lib\runpy.py", line 87, in _run_code
2023-06-06T22:59:19.9502615Z     exec(code, run_globals)
2023-06-06T22:59:19.9503082Z   File "D:\a\cocotb\cocotb\.nox\dev_test_nosim\Scripts\pytest.EXE\__main__.py", line 7, in <module>
2023-06-06T22:59:19.9503508Z     sys.exit(console_main())
2023-06-06T22:59:19.9504394Z   File "D:\a\cocotb\cocotb\.nox\dev_test_nosim\lib\site-packages\_pytest\config\__init__.py", line 189, in console_main
2023-06-06T22:59:19.9504859Z     code = main()
2023-06-06T22:59:19.9505426Z   File "D:\a\cocotb\cocotb\.nox\dev_test_nosim\lib\site-packages\_pytest\config\__init__.py", line 147, in main
2023-06-06T22:59:19.9505897Z     config = _prepareconfig(args, plugins)
2023-06-06T22:59:19.9506505Z   File "D:\a\cocotb\cocotb\.nox\dev_test_nosim\lib\site-packages\_pytest\config\__init__.py", line 328, in _prepareconfig
2023-06-06T22:59:19.9507034Z     config = pluginmanager.hook.pytest_cmdline_parse(
[...]

PR #3392, opened 2023-08-07
CI: Enable MSVC test

themperek:
Enable CI "windows + conda + msvc"
Change python to 3.11

There are errors `bash: C:Minicondapython.exe: command not found` that probably #3378 will solve.

Closes https://github.com/cocotb/cocotb/issues/3326


ktbarrett:
Need to update codecov config for correct number of report to expect.

ktbarrett (approved):

PR #3401, opened 2023-08-23
Alternative to #3833

ktbarrett:
Adds some fixes to #3833.

### DIFF
--- a/tests/test_cases/test_discovery/Makefile
+++ b/tests/test_cases/test_discovery/Makefile
@@ -25,6 +25,7 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
+export COCOTB_LOG_LEVEL := TRACE
 
 include ../../designs/sample_module/Makefile
 
--- a/tests/test_cases/test_discovery/Makefile
+++ b/tests/test_cases/test_discovery/Makefile
@@ -25,7 +25,6 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
-export COCOTB_LOG_LEVEL := TRACE
 
 include ../../designs/sample_module/Makefile
 
--- a/tests/test_cases/test_discovery/test_discovery.py
+++ b/tests/test_cases/test_discovery/test_discovery.py
@@ -276,7 +276,13 @@ async def test_write_single_character(dut):
         lowercase = chr(c.value)
         uppercase = lowercase.upper()
         uppercase_as_int = ord(uppercase)
-        dut._log.debug("Converting '%s' (%s) to '%s' (%s)", lowercase, c.value, uppercase, uppercase_as_int)
+        dut._log.debug(
+            "Converting '%s' (%s) to '%s' (%s)",
+            lowercase,
+            c.value,
+            uppercase,
+            uppercase_as_int,
+        )
         c.setimmediatevalue(uppercase_as_int)
     await Timer(1, "ns")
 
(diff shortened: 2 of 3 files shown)


ktbarrett:
@Forty-Bot you stil okay with these changes coming in?

tests/test_cases/test_discovery/Makefile
```diff
@@ -25,6 +25,7 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
+export COCOTB_LOG_LEVEL := TRACE
```
ktbarrett, line 28:
```suggestion
```

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -276,6 +287,13 @@ async def test_write_single_character(dut):
         lowercase = chr(c.value)
         uppercase = lowercase.upper()
         uppercase_as_int = ord(uppercase)
+        dut._log.debug(
+            "Converting '%s' (%s) to '%s' (%s)",
+            lowercase,
+            c.value,
+            uppercase,
+            uppercase_as_int,
+        )
```
ktbarrett, line 296:
```suggestion
```

ktbarrett:
Previously we were getting the value from a single character handle in a string using `vpiIntVal` as the value type and for uninitialized characters it was returning `0` or NUL. But now that it is using either `vhpiEnumVal`, `vhpiSmallEnumVal`, or `vhpiCharVal` as the type of value to get from the single character handle the value being returned is `N` (it's probably random). I'm not entirely sure why this is, but it's certainly a bug on Questa's end.

imphil:
We need to avoid the regression for Questa users, no matter who's "at fault".

> But now that it is using either vhpiEnumVal, vhpiSmallEnumVal, or vhpiCharVal as the type of value to get from the single character handle

Which one of these is it?

What does the VHPI standard say it should be?

If we want to move forward with this change (which seems needed for NVC), we need understand the failure better, and then probably keep the current code at least for Questa (ifdef it), and file an *actionable and detailed* bug that I can forward to Siemens to look into further.

nickg:
> 
> What does the VHPI standard say it should be?
> 

The relevant bit in the 2008 LRM is section 23.19 vhpi_get_value:

> The function reads the value of the object referred to by expr (see 22.3) and represents it in the format specified in the format member of the value structure (see 22.2).
>
> [..]
>
> It is an error if the format member of the value structure specifies a format that cannot be used to represent the value. It is an error if the formatted value is outside of the range of values that can be represented.

And earlier in 22.2:

> It is an error if a VHPI program uses a given type described in this clause to represent a VHDL scalar type, and there are position numbers in the scalar type that exceed the range of position numbers that can be represented in the given type.

Reading this more carefully I think the original code here is ok: it should be possible to read a character as `vhpiIntVal` (but not an integer as `vhpiCharVal`, because of the last clause above). At the moment NVC reports an error if the requested format doesn't exactly match the format expected for the type. I can fix that.

But I think @ktbarrett is right that ModelSim has a bug here: the value should be zero regardless of whether you read it as `vhpiIntVal` or `vhpiCharVal`.

ktbarrett:
So do we want to instead use `vhpiIntVal` under Questa and the others introduced by the change otherwise? The Questa VHPI interface is new(-ish) so we might just be able to get Questa fixed and sweep the broken versions of Questa under the rug.

nickg:
> 
> At the moment NVC reports an error if the requested format doesn't exactly match the format expected for the type. I can fix that.
> 

I made this change in nickg/nvc@ff6a36d so NVC can work with either now.

imphil:
Thanks, @nickg! To keep the cocotb code simple, let's avoid the ifdef and just keep the current code. Nonetheless, we should get a bug report out to Siemens. Maybe someone can write up one (that enables them to reproduce the issue ideally without much cocotb involvement), or I'll do it when I find time (which is rather limited, unfortunately).

PR #3409, opened 2023-09-14
Enable Python 3.12 builds

imphil:
Update cibuildwheel to build wheels for Python 3.12, which is now in rc2
mode and will not see any backwards-incompatible changes to its API.
Also update the CI configuration to build Python 3.12 by default for all
builds, not only the experimental ones.

### DIFF
--- a/.github/generate-envs.py
+++ b/.github/generate-envs.py
@@ -61,15 +61,24 @@ ENVS = [
         "python-version": "3.11",
         "group": "ci",
     },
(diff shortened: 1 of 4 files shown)


themperek (approved):

ktbarrett:
Should we update the version compatability list to include 3.12 (if there is one)? And maybe a changeling entry.

imphil:
Updated the PR to include a newsfragment and updated docs.

PR #3411, opened 2023-09-16
Improve typing of `cocotb.test`

ktbarrett:
This is accomplished by changing the impl of the decorator to use "normal" decorators instead of metaclasses, which often confuse type checkers.


imphil (approved):
LGTM, thanks!

PR #3416, opened 2023-09-18
Simplify includes in simulatormodule

imphil:
The LOG_* macros are used in code that lives in cocotb_utils.h, so we
don't need to import the header into simulatormodule.cpp.

### DIFF
--- a/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -39,7 +39,6 @@ static int releases = 0;
 
 #include <Python.h>
 #include <cocotb_utils.h>    // to_python to_simulator
-#include <gpi_logging.h>     // LOG_* macros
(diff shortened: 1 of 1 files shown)


ktbarrett:
Relying on one header pulling in another header is known as "transitive includes" and is generally considered bad practice. For example, if the cocotb_utils.h is refactored to not need the `LOG_*` commands, the gpi_logging.h include in cocotb_utils.h can be removed, and all downstreams that implicitly depended on gpi_logging.h are now broken.

imphil:
> Relying on one header pulling in another header is known as "transitive includes" and is generally considered bad practice. 

I get that part, but how is it related to this PR? I'm removing an include that is not used, so it's a win, isn't it?

ktbarrett:
Oh. I guess I figured we used to the LOG_ macros at least once in that file, but apparently not. Let's remove then.

ktbarrett (approved):

PR #3417, opened 2023-09-18
Remove gpi_free_handle()

imphil:
This function is only part of the interface, but not implemented. Remove
it until we actually need/implement it. (I guess we would already need
it today, but aren't overly thorough with our cleanup.)


ktbarrett (approved):

ktbarrett:
"No space left on device" errors

ktbarrett:
re-running CI

ktbarrett:
Well I figured because I didn't see those errors in my PR that it was resolved. Oh well...

imphil:
I'm taking a risk and won't rerun Private CI for that, it's rather unlikely we'll miss a build error with the existing set of tests.

PR #3438, opened 2023-10-11
Run tests without clean from nox

imphil:
In nox, run "make test" without a "make clean" run first. "make clean" for
the full test suite takes multiple minutes (mostly due to the overhead
of repeatedly calling cocotb-config) and is unnecessary on CI runners,
which always start from a clean state. Remove "make clean" from the
noxfile, developers can still call it on demand if necessary.


cmarqu (approved):

imphil:
Backported into stable/1.9.

PR #3427, opened 2023-09-29
Add NVC support by reverting removal

nickg:
This is another attempt to add NVC support by doing a more-or-less straight revert of 7f07ddac36242 which therefore shouldn't have any copyright assignment issues. On top of that is some tweaks to the Makefile to work with the current version of cocotb plus a patch to make name comparison case-insensitive in the VHPI implementation when `NVC` is defined (my own work).

PR #3439, opened 2023-10-11
Fix incorrect comparison of generate region labels

### LINKED ISSUE #2255: GPI: gpi_get_handle_by_name for generate blocks may find wrong block
Pseudo-region lookup may find the wrong generate block, as the lookup only checks if the generate block starts with the lookup name, instead of a full match of the name before the index.

As a consequence, `dut.genblk` will match against `genblk1` in the design, or `dut.gen_something` may match `gen_something_else` instead, depending on the iterator order.

This type of lookup iterates over the parent module handle, looking for generate blocks, and is the fall-back method if **vpi_handle_by_name()**/**vhpi_handle_by_name()** lookup fails (which it will if given a truncated name).

This issue is present in all 3 GPI implementations:

https://github.com/cocotb/cocotb/blob/38f199c8fbcaaa6147f3a5da815880cb8d16f36e/cocotb/share/lib/fli/FliImpl.cpp#L312

https://github.com/cocotb/cocotb/blob/38f199c8fbcaaa6147f3a5da815880cb8d16f36e/cocotb/share/lib/vhpi/VhpiImpl.cpp#L525

https://github.com/cocotb/cocotb/blob/38f199c8fbcaaa6147f3a5da815880cb8d16f36e/cocotb/share/lib/vpi/VpiImpl.cpp#L335

nickg:
In various places we compare two generate labels by checking if the second label starts with the first. But this fails if two generate blocks with different names share the same prefix - e.g. "foo" and "foobar". Instead we should strip off any index suffix like "[0]" and then compare the full strings.

The included test fails with current cocotb and `SIM=icarus` with `ValueError: Unable to match an index pattern: foobar[0]` but passes after this change. I tested the VHPI implementation with NVC by rebasing on top of #3427. I have no way to test the FLI implementation but it does build.

Fixes #2255

### DIFF
--- a/cocotb/share/lib/fli/FliImpl.cpp
+++ b/cocotb/share/lib/fli/FliImpl.cpp
@@ -349,7 +349,7 @@ GpiObjHdl *FliImpl::native_check_create(const std::string &name,
             if (acc_fetch_fulltype(rgn) == accForGenerate) {
                 std::string rgn_name =
                     mti_GetRegionName(static_cast<mtiRegionIdT>(rgn));
-                if (rgn_name.compare(0, name.length(), name) == 0) {
+                if (compare_generate_labels(rgn_name, name)) {
                     FliObj *fli_obj = dynamic_cast<FliObj *>(parent);
                     return create_gpi_obj_from_handle(
                         parent->get_handle<HANDLE>(), name, fq_name,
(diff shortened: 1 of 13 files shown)


marlonjames (commented):
Looks reasonable.
This should have a news fragment for fixing #2255.

nickg:
> This should have a news fragment for fixing #2255.

Done

imphil:
> I have no way to test the FLI implementation but it does build.

FLI gets tested in CI, so we're ok if the test passes.

marlonjames (approved):

PR #3447, opened 2023-10-18
Fix tracing for Xcelium runner

toddstrader:
Previously trying to trace with Xcelium would yield `-input '-input @database . . .` for the runner command which is not correct / makes Xcelium mad.  I see that there is Xcelium CI:
https://github.com/cocotb/cocotb/pull/3437
but am not sure how to add a test for this.  Please advise if that would be desirable.

For what it's worth, the runner now works for me with `waves` as both `True` and `False`.

PR #3451, opened 2023-10-21
Use directly coverage instead of pytest-conv

themperek:
When using `nox+pytest-conv` some files are not properly covered. https://github.com/cocotb/cocotb/pull/3447

There are some import issues.


themperek:
I have no import issues locally ?

PR #3452, opened 2023-10-23
Fix pytest coverage

toddstrader:
Fixes pytest coverage (I think).

See: #3447 
`runner.py` claims to have not coverage (except for things which would get covered by importing the module).  This is clearly not true because the runner is being exercised by `test_runner.py`.  @themperek pointed out that this is because we're missing `__init__.py`.  Adding this angers python in ways I don't understand, but with enough trial and error I found that passing `tests/pytest/` to `pytest` gets around that problem.

I'm definitely just beating on this with a wrench until it works, so please tell me if this is bad in some way.

Running `nox` locally gives me correct looking coverage for `runner.py`.  I'm going to check if codecov agrees with my local results after I post this PR.  For reference, here's a recent run showing the missing `runner.py` coverage:
https://app.codecov.io/gh/cocotb/cocotb/commit/a40673d02c3e304b66cc2a82baaa9496152f76b7/blob/cocotb/runner.py

As an aside, does anyone know if it is possible to run the cocotb CI in my personal fork?  I'd love to check this out before posing the PR but do not know how.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -181,6 +181,7 @@ def dev_test_sim(
         "--cov-report=",
         "-k",
         "simulator_required",
+        "tests/pytest/",
         env=env,
     )
     Path(".coverage").rename(".coverage.pytest")
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,3 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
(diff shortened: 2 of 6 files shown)


toddstrader:
Blech.  `nox -s "dev_test_nosim"` works for me locally.  Not yet sure what the difference is.

themperek:
Same here https://github.com/cocotb/cocotb/pull/3451

It works locally. Do not know (yet) what is wrong here.

toddstrader:
> Same here #3451
> 
> It works locally. Do not know (yet) what is wrong here.

Oh, whoops.  Didn't see this before I posted this PR.  I have no particular preference for how the problem is solved.

ktbarrett:
I looked into it last night and the usual culprits: version of packages changing, wasn't the cause. The issue seems to be that it's preferring the local package instead of the installed one. Not sure why it's suddenly acting up. I had discussed moving to a `src/` style project to alleviate this.

ktbarrett:
The lint is definitely broken and that needs looking at.

PR #3457, opened 2023-10-29
Add graphviz package to doc build config

### LINKED ISSUE #3449: Documentation: error rendering inheritance diagrams (v1.8.1)
Diagram images not rendering (https://docs.cocotb.org/en/stable/library_reference.html#simulation-object-handles)

cocotb version: v1.8.1

cmarqu:
This makes the 'dot' binary available to sphinx.ext.inheritance-diagrams
See also https://github.com/readthedocs/readthedocs.org/issues/8672

Closes #3449

In addition, remove the `graphviz_output_format = "svg"` setting since this
renders badly (much too big and without scrollbars) with the currently used
package versions (this could be https://github.com/sphinx-doc/sphinx/issues/10117).

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,6 +8,8 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  apt_packages:
+    - graphviz
 
 python:
   install:
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -342,6 +342,10 @@ spelling_ignore_wiki_words = False
 spelling_show_suggestions = True
 spelling_ignore_acronyms = True
 
+# -- Extra setup for inheritance_diagram directive which uses graphviz ---------
+
+graphviz_output_format = "svg"
+
 # -- Extra setup for towncrier -------------------------------------------------
 # see also https://towncrier.readthedocs.io/en/actual-freaking-docs/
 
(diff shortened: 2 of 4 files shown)


cmarqu:
To review, look at https://cocotb--3457.org.readthedocs.build/en/3457/library_reference.html#simulation-object-handles

marlonjames:
The image is scaled down from `768x207` to `696.4x187.7` in the docs, but the overlay map with click-links to the classes uses the full-scale measurements.
This results in the link areas being offset from the visible class rectangles.

cmarqu:
@marlonjames Seems something like https://github.com/sphinx-doc/sphinx/issues/4865 has crept up again.
I don't feel like reporting and potentially following up on a Sphinx ticket right now, given the state of the SVG problem in https://github.com/sphinx-doc/sphinx/issues/10117.
Just accept the behavior, hoping it will be fixed at some point, or just remove the whole inheritance diagram?

marlonjames:
I would be fine with either removing the `<map>` or leaving it scaled wrong.

ktbarrett:
Not having a clickable overlay is acceptable.

cmarqu:
Strictly speaking, we now don't use `sphinx.ext.inheritance_diagram` and `graphviz` anymore and could remove it from the doc building setup, but I'd still keep it in to easier re-generate the diagram if needed.

ktbarrett (approved):

PR #3465, opened 2023-10-31
enable Verilator CI

toddstrader:
I believe this should be possible now.  Most things pass when I run `nox` but there are also a number of failures.  I'm posting this separately from the test fixes as I'm thinking it may be desirable to have separate conversations for some of those tests and their implications.  But if it's preferred I can lump everything together here.  Obviously this is not land-able until all of the tests are passing / skipped / something.

### DIFF
--- a/.github/generate-envs.py
+++ b/.github/generate-envs.py
@@ -126,23 +126,10 @@ ENVS = [
     {
         "lang": "verilog",
         "sim": "verilator",
-        "sim-version": "v4.106",
-        "os": "ubuntu-20.04",
-        "python-version": "3.8",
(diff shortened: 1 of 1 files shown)


marlonjames:
I am working on updating/finishing a PR for Verilator CI at #3237.

toddstrader:
> I am working on updating/finishing a PR for Verilator CI at #3237.

Blast, missed that.  Sorry about that.

marlonjames:
It's ok, it hasn't been updating in a while but I am working on it again, including creating cocotb and upstream issues where they are missing.

PR #3466, opened 2023-10-31
expect failure for test_force_release with Verilator

toddstrader:
As noted in the comment, Verilator currently ignores the `flags` field to `vpi_put_value()`.  I was looking at adding some kind of assert to `VpiSignalObjHdl::set_signal_value()` but then I noticed that `vpiInertialDelay` is usually provided for `GPI_DEPOSIT`.

Would it be better to also assert `action == GPI_DEPOSIT` for Verilator?  And are there other issues because Verilator doesn't do `vpiInertialDelay`?  I'm unclear of the implications for cocotb.

### DIFF
--- a/tests/test_cases/test_force_release/test_force_release.py
+++ b/tests/test_cases/test_force_release/test_force_release.py
@@ -5,7 +5,8 @@ from cocotb.handle import Force, Release
 from cocotb.triggers import Timer
 
 
-@cocotb.test(expect_fail=cocotb.SIM_NAME.lower().startswith("ghdl"))
+# Verilator does not observe the vpi_put_value() flags field
+@cocotb.test(expect_fail=cocotb.SIM_NAME.lower().startswith(("ghdl", "verilator")))
 async def test_force_release(dut):
     """
     Test force and release on simulation handles


toddstrader:
Relates to #3465

ktbarrett:
> And are there other issues because Verilator doesn't do vpiInertialDelay? I'm unclear of the implications for cocotb.

Verilator already behaves differently than any other simulator, so I don't necessarily think this is a huge deal.

tests/test_cases/test_force_release/test_force_release.py
```diff
@@ -5,7 +5,8 @@
 from cocotb.triggers import Timer
 
 
-@cocotb.test(expect_fail=cocotb.SIM_NAME.lower().startswith("ghdl"))
+# Verilator does not observe the vpi_put_value() flags field
+@cocotb.test(expect_fail=cocotb.SIM_NAME.lower().startswith(("ghdl", "verilator")))
```
ktbarrett, line 9: I wish we also knew why GHDL doesn't work so we could add a github issue label now before we forget, but oh well.

ktbarrett (approved):

toddstrader:
Also redundant with #3237

Issue #3293, opened 2023-05-02
Aldec Riviera segfaults with cocotb installed by pip


abyszuk:
Hi, I think I've hit a bug similar to #3078 , but a bit more difficult to trigger.
I've got a mixed-language design with both verilog and VHDL modules. For this particular case the path looks like this:
`tb_pb_top (SV) -> pb_top (AKA dut) (VHDL) -> x_aurora_c2c (VHDL)`
and I've got symptoms very similar with what was going on in #3078 :
- cocotb installed with `pip3 install cocotb==1.7.2` segfaults with `COUT: free(): invalid pointer`
- cocotb installed with `pip3 install git+https://github.com/cocotb/cocotb@v1.7.2 --user` works

Software stack is also similar to #3078 :
- Aldec Riviera 2022.10 (previously 2021.04) in a docker container (based on CentOS 8)
- Ubuntu 22.04 host
- Python 3.6 inside the container

Here's the log collected with `COCOTB_LOG_LEVEL=debug`:
```
# COUT:      0.00ns INFO     cocotb                             Running on Riviera-PRO version 2022.10.117.8716
# COUT:      0.00ns INFO     cocotb                             Running tests with cocotb v1.7.2 from /home/adrian/.local/lib/python3.6/site-packages/cocotb
# COUT:      0.00ns INFO     cocotb                             Seeding Python random module with 1683037810
# COUT:      0.00ns DEBUG    gpi                                Looking for root handle 'tb_pb_top' over 2 implementations
# COUT:      0.00ns DEBUG    gpi                                Got a Root handle (tb_pb_top) back from VPI
# COUT:      0.00ns DEBUG    gpi                                Checking tb_pb_top exists
# COUT:      0.00ns DEBUG    cocotb.tb_pb_top                   Created
# COUT:      0.00ns DEBUG    cocotb.regression                  Python Path: ,/home/adrian/praca/cce/fgc4/common/python,/usr/lib64/python36.zip,/usr/lib64/python3.6,/usr/lib64/python3.6/lib-dynload,/home/adrian/.local/lib/python3.6/site-packages,/usr/lib64/python3.6/site-packages,/usr/lib/python3.6/site-packages
# COUT:      0.00ns DEBUG    cocotb.regression                  PWD: /home/adrian/praca/cce/fgc4/pb/tb
# COUT:      0.00ns INFO     cocotb.regression                  Found test sim_server.test
[...]

PR #3468, opened 2023-10-31
Skip test_iteration_mixedlang on Riviera

imphil:
This test crashes on Riviera-PRO, skip it.

See https://github.com/cocotb/cocotb/issues/3293

### DIFF
--- a/tests/test_cases/test_iteration_mixedlang/Makefile
+++ b/tests/test_cases/test_iteration_mixedlang/Makefile
@@ -25,5 +25,13 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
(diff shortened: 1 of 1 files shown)


imphil:
Turns out the test that was failing release CI was *not* the custom enty point one, but `test_iteration_mixedlang`, which is also discussed at https://github.com/cocotb/cocotb/issues/3293 (in there, we just didn't know we had an easy way to reproduce with our built-in test suite. And since we never ran tests against wheel releases, we didn't catch this issue in CI before either).

I don't see any hope that we'll solve https://github.com/cocotb/cocotb/issues/3293 soon, so I'm disabling this test for now to get our CI green again (which prevents us from doing releases, among other things).

themperek (approved):

imphil:
Backported into stable/1.9

PR #3470, opened 2023-11-01
Compile dev builds with -Wsuggest-override

### LINKED ISSUE #1368: Build with -Wsuggest-override in CI runs
Split from #1149. Using `-Wsuggest-override` would catch some additional errors.
Unfortunately not all compilers support this flag, so we cannot just add it to our default set of build flags.

The attempt there was to use:
```
# version-specific flags
# easiest to compare versions with python
version_less_than_equal = $(shell $(PYTHON_BIN) -c "import sys; from distutils.version import LooseVersion; print(LooseVersion(sys.argv[1]) <= LooseVersion(sys.argv[2]))" $(1) $(2))
GXX_VERSION = $(shell g++ -dumpversion)
ifeq ($(call version_less_than_equal,4.9,$(GXX_VERSION)),True)
CXX_WARNS += -Wsuggest-override
endif
```
but this assumed that the user was using GCC.

We might be able to get away with just adding to `tox.ini`, assuming all our CI compilers are new enough.

imphil:
Fixes #1368

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -104,7 +104,15 @@ def configure_env_for_dev_build(session: nox.session) -> None:
     - Enable coverage collection.
     - Build with more aggressive error checking.
     """
-    session.env["CFLAGS"] = "-Werror -Wno-deprecated-declarations -g --coverage"
+    session.env["CFLAGS"] = " ".join(
+        [
+            "-Werror",
+            "-Wno-deprecated-declarations",
+            "-Wsuggest-override",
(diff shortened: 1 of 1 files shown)


imphil:
Conveniently this new compiler check did not flag any new issues, but it's still good to have it (based on previous experience, which got @eric-wieser to file https://github.com/cocotb/cocotb/issues/1368)

ktbarrett (approved):

PR #3471, opened 2023-11-01
Update to the latest version of Riviera in CI

imphil:
- Remove test waiver for test_both_edge_triggers on recent Riviera
- Use the latest version of Riviera in CI

### DIFF
--- a/tests/test_cases/issue_330/issue_330.py
+++ b/tests/test_cases/issue_330/issue_330.py
@@ -3,14 +3,23 @@
 import logging
 
 import cocotb
+from cocotb._sim_versions import RivieraVersion
 
 SIM_NAME = cocotb.SIM_NAME.lower()
 
 
 # GHDL unable to access record signals (gh-2591)
 # Icarus doesn't support structs (gh-2592)
+# Riviera-PRO 2022.10 and newer does not discover inout_if correctly over VPI (gh-3587)
 @cocotb.test(
-    expect_error=AttributeError if SIM_NAME.startswith(("icarus", "ghdl")) else ()
+    expect_error=AttributeError
+    if SIM_NAME.startswith(("icarus", "ghdl"))
+    or (
+        SIM_NAME.startswith("riviera")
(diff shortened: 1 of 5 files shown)


imphil:
Oh, now other tests are failing, I'll take a look. *But* I'm loving the new CI setup, which makes these failures so much easier to see:

grafik

marlonjames:
Superceded by #3585

marlonjames:
Actually this has one of the tests fixed, so we should complete this one.

imphil:
Filed #3587 for the test failures, filed an upstream issue with Aldec, and waived the failures in our tests for the affected versions.

tests/test_cases/issue_330/issue_330.py
```diff
@@ -45,4 +54,13 @@ async def issue_330_iteration(dut):
         tlog.info("Found %s" % member._path)
         count += 1
 
-    assert count == 2, "There should have been two members of the structure"
+    # Riviera-PRO 2022.10 and newer does not discover inout_if correctly over VPI (gh-3587)
+    rv_2022_10_plus = RivieraVersion(cocotb.SIM_VERSION) >= RivieraVersion("2022.10")
+    if (
+        SIM_NAME.startswith("riviera")
+        and rv_2022_10_plus
```
ktbarrett, line 61: Weird that this is the only one that isn't inlined, but it also doesn't matter.

ktbarrett (approved):

tests/test_cases/issue_330/issue_330.py
imphil: Yeah, but it's a different exception than the other waiver. I could have widened the waiver to all Exception objects, but found that too course grained. That's why we have this solution.

imphil: Oh sorry, you meant the variable! The autoformatting looks really strange if I inline that, that's why I pulled it out and chose the rather strange variable name.

PR #3475, opened 2023-11-02
Update Makefile.inc

wyxdexyq:
duplicate $ lead to 
"bash: C:******miniconda3python.exe: command not found"


imphil:
Thanks for your pull request, @wyxdexyq! At this point, I suspect the problem you're having is caused by something else, and the fix here isn't actually the right one. Can you please describe a bit more what your problem is in a new issue? Please fill out all fields it asks for, especially your Windows version, etc.

PR #3476, opened 2023-11-02
CI: Use elevated permissions to push benchmark results

imphil:
Use permissions from a new GitHub App to push to the benchmark results
repository.

Also, use a fixed version of GHDL when running the performance benchmark to actually test cocotb's performance, and not the performance of GHDL over time.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -9,12 +9,6 @@ on:
     branches:
       - master
 
-permissions:
-  # deployments permission to deploy GitHub pages website
-  deployments: write
-  # contents permission to update benchmark contents in gh-pages branch
-  contents: write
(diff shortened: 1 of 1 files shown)


imphil:
Looks like results are being published: https://cocotb.github.io/cocotb-benchmark-results/dev/bench/

imphil:
@themperek, this PR fixes the permission issues we were seeing when pushing the performance benchmark. The setup we have now is not trivial (I had to create a GitHub App, assign keys to it, store the keys in GH Action secrets, install the app, and then update the workflow here). But either way, it's now working.

themperek (approved):
Thank you!

https://cocotb.github.io/cocotb-benchmark-results/dev/bench/

PR #3477, opened 2023-11-02
Fix Task repr for Coroutine objects

marlonjames:
`Task.__repr__` would raise an exception if the wrapped object does not have a `__name__` attribute.
Added fallback to `type().__name__` in that case.


ktbarrett:
Is this necessary? object attribute access defers to the type when not present AFAIK. Maybe it doesn't work in this instance?

marlonjames:
It raises `AttributeError` and blows up the scheduler when running with `COCOTB_SCHEDULER_DEBUG=1`.

ktbarrett (approved):

PR #3479, opened 2023-11-03
Random changes

### LINKED ISSUE #3464: `import` order is important and may break
If `cocotb.triggers` is not imported before `cocotb.task` it throws `ImportError: cannot import name 'Task' from partially initialized module 'cocotb.task' (most likely due to a circular import)`. It is pretty clear how `cocotb.task` and `cocotb.trigger` are circularly dependent. Maybe there is a way to avoid this by design rather than depend on the particular import order in `cocotb/__init__.py`.

ktbarrett:
Closes #3464, for now. Also fixes the typing on `cocotb.test` and makes the `cocotb.external` and `cocotb.function` into function-based decorators, which does change the interface, but simplifies the implementation and makes the resulting object more like the regular `async def` functions that cocotb has now, instead of unique types like `cocotb.coroutine`.

Next will be the removal of `cocotb.coroutine` (which requires I first support `coroutines` in `First` and `Combine`), then privatization of `Test` (since there is now no way for the user to get a handle on the object, and it might move to another module).

### DIFF
--- a/cocotb/decorators.py
+++ b/cocotb/decorators.py
@@ -166,7 +166,7 @@ def test(
     expect_error: Union[bool, Type[Exception], Sequence[Type[Exception]]] = (),
     skip: bool = False,
     stage: int = 0,
-) -> Callable[[Callable[..., None]], Test]:
+) -> Callable[[Callable[..., Coroutine[Any, Any, None]]], Test]:
     """
     Decorator to mark a Callable which returns a Coroutine as a test.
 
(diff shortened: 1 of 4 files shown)


ktbarrett:
```
.nox/dev_test_nosim/lib/python3.6/site-packages/cocotb/regression.py:43: in <module>
    import cocotb.ANSI as ANSI
E   AttributeError: module 'cocotb' has no attribute 'ANSI'
```

Do what now?

marlonjames (approved):
I'm fine with the changes, but I don't use external/function

ktbarrett:
@marlonjames The only functional change is they don't support the `log` attribute and don't support the `yield` syntax anymore. The scheduler doesn't need modification as the two functions used do all the heavy lifting, the types did nothing really. This should mirror the capabilities of undecorated coroutine functions, which should be fine as most people aren't using the decorator or old syntax anymore.

PR #3481, opened 2023-11-04
Changing `TriggerException`

### LINKED ISSUE #3459: Only use `TriggerException` in `prime` methods
There are two cases where `TriggerException` is used in user-facing code where another type would be better. It's only utility is in `expect_error` clauses for our own testing as the user will never be able to catch it.

ktbarrett:
Changes it's use in user-facing code and then makes it private. There doesn't seem to be any special handling of these in the scheduler, so I'm not entirely sure the point?

Closes #3459.

### DIFF
--- a/cocotb/triggers.py
+++ b/cocotb/triggers.py
@@ -223,7 +223,7 @@ class Timer(GPITrigger):
             :func:`~cocotb.utils.get_sim_steps`
 
         Raises:
-            TriggerException: If a negative value is passed for Timer setup.
+            ValueError: If a negative value is passed for Timer setup.
 
         .. versionchanged:: 1.5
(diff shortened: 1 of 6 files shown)


cocotb/triggers.py
```diff
@@ -250,7 +250,7 @@ def __init__(
                     stacklevel=2,
                 )
             else:
-                raise TriggerException("Timer value time_ps must not be negative")
+                raise ValueError("Timer value time_ps must not be negative")
```
marlonjames, line 253:
```suggestion
                raise ValueError("Timer argument time must not be negative")
```

marlonjames (approved):

imphil:
The performance benchmark CI fail is due to https://github.com/cocotb/cocotb/pull/3482.

PR #3484, opened 2023-11-06
Move to src/ project

### LINKED ISSUE #2286: Move to a src/ layout

### LINKED ISSUE #2698: 'make help' reveals a circular dependency problem in handle.py while importing simulator from cocotb

ktbarrett:
Reopens #2358 in a new PR. Closes #2286 and closes #2698. Supersedes #3452 and supersedes #3451. This seems to fix the coverage reporting on `cocotb/runner.py`.

All PRs will need to be rebased after this commit. I'm not sure how cleanly that will be done.


cmarqu (approved):

imphil (approved):
Thanks! It's time to get that annoyance out of the way.

Rebases should be fine, git typically handles renames rather well.

PR #3486, opened 2023-11-06
Rename documentation/ to the more common docs/

cmarqu:


imphil (approved):
I like it, documentation always sounded overly long and unusual to me. Since we're in the renaming business at the moment this one fits in nicely.

PR #3488, opened 2023-11-06
Replace flake8 with ruff; convert setup.cfg to pyproject.toml

cmarqu:


imphil (commented):
Thanks @cmarqu!

Have you tried replacing black and isort with ruff as well? (We only get the speed benefits really if we get rid of most of the tools, otherwise we're back to parsing the AST multiple times.)

If ruff works out formatting-wise, the main concern I have is that we're relying on binary components (which people cannot really build from source, a rust toolchain isn't that common these days). However, the wheels provided by the ruff project (https://pypi.org/project/ruff/#files) seem to be targeting manylinux2014/glibc 2.17, which means RHEL7 and include wheels even for s390x and ppc64lc. We seem to be loosing support for Python 3.6 and Ubuntu 18.04, which are really EOL on development machines these days. 

So I think I convinced myself that these binaries are good enough. I hope the ruff project keeps up the commitment to older versions of various tools -- and if not, we can always go back to our existing setup.

ktbarrett:
I wonder if we could get the `ruff` guys to do fix https://github.com/psf/black/issues/2156 since the black authors are so unwilling to change that awful formatting strategy.

cmarqu:
Superseded.

PR #3491, opened 2023-11-07
nox: Cleanup cocotb install for sessions

marlonjames:
Starting a `dev` session would install cocotb twice.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -527,13 +527,13 @@ def create_env_for_docs_build(session: nox.Session) -> None:
         "pip", "install", "docs/_vendor/domaintools"
     )  # not done in requirements.txt due to the way relative paths are handled in that file (gh-pypa/pip#8765)
     session.run("pip", "install", "-r", "docs/requirements.txt")
-    session.run("pip", "install", "-e", ".")
 
 
 @nox.session
 def docs(session: nox.Session) -> None:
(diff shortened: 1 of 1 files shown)


imphil (approved):
It would probably be nicer to move the cocotb install out of the  `create_env_for_docs_build()` call, but looking at the code that's annoying to do for very little gain. So your solution is nicely pragmatic. Thanks!

ktbarrett:
I would move the cocotb install out of the create_env_for_docs function. Is there a particular reason we are install cocotb in editable mode for documentation anyways? Seems like an assumption that has nothing to do with doc generation.

imphil:
>  Is there a particular reason we are install cocotb in editable mode for documentation anyways?

I *think* we get automatic rebuilds of the docs (at least in the preview docs mode) if the Python sources change, and editable installs enable that.

marlonjames:
Hmm, the clang CI run failed:
https://github.com/cocotb/cocotb/actions/runs/6788949643/job/18455067554?pr=3491#step:8:158

imphil:
The clang job ended up on our self-hosted runner, where clang is not installed. The labels should prevent that, but I've seen that previously and haven't had a chance to understand the issue yet. 

Instead of reopening the PR and rerunning all tests, you can just rerun the failing test in the actions view.

ktbarrett (approved):

PR #3497, opened 2023-11-08
Scheduler: remove context manager tracking _current_task

marlonjames:
With fork() removed, the scheduler is no longer re-entrant, and no task should be scheduled while another is already running.

There is an added benefit of performance increase. In testing, runtime reduction ranged from ~2% for matrix_multiplier, to ~15% on some workloads involving a very high number of task switches.


src/cocotb/scheduler.py
```diff
@@ -809,7 +798,10 @@ def _schedule(self, task, trigger=None):
             trigger (cocotb.triggers.Trigger): The trigger that caused this
                 task to be scheduled.
         """
-        with self._task_context(task):
+        if self._current_task is not None:  # pragma: no cover
```
imphil, line 801: I wouldn't exclude this line from coverage, it's an edge case that's clearly not tested (and could be, if we wanted to reach coverage closure). [Just to be clear: I don't think a test is needed as part of this PR.]

imphil (approved):
:partying_face: Thanks, nice cleanup and nice performance gain. It's great to see the benefits from removing `fork()` materializing.

ktbarrett (approved):

Issue #2342, opened 2021-01-13
Implement proper initialization failure handling


ktbarrett:
`embed_init_python` doesn't allow failure (no return code), requiring us to put safety checks in the other functions in embed.cpp. When loading `GPI_EXTRA` libraries, `exit()` is called if there was a failure. `exit()` is hooked on some simulators to call the shutdown callback. In Questa this callback is immediate, since `exit()` must not return, this isn't good behavior and we should avoid it. There are probably other examples of initialization not being able to fail gracefully, they should be fixed.

ktbarrett:
xref #2018.

To expand, we can't use `__del__` and `Py_Finalize` for this behavior as the cleanup order is not deterministic.

Calling `gpi_sim_end` if initialization fails is appropriate, need to investigate if that will cause `cocotb._sim_event` to not be called. It's probably a good idea for `cocotb._initialize_testbench` to be in charge of cleaning up initialization if it fails, so that `cocotb._sim_event` can assume if it is run that all global objects are initialized.

ktbarrett:
xref #2272. All other specific issues besides calling `exit` have been fixed, so I will close this and make a new issue for that. Not a fan of vague tracking issues.

Issue #2018, opened 2020-08-14
call gpi_sim_end if cocotb init function errors


ktbarrett:
`embed_sim_init` can fail for various reasons, including the Python entry point `cocotb._initialize_testbench` throwing. We don't currently call `gpi_sim_end` if we enter the error case. This causes Verilator to hang since it does not shutdown during simulator scheduling exhaustion.

PR #3508, opened 2023-11-14
Cleanup startup errors

marlonjames:
This fixes part of #2018, but the catch and cleanup is in `_initialise_testbench_()` instead of `_embed_sim_init()`.

If the error occurs in setting up the `RegressionManager`, then `cocotb._sim_event()` won't be called any longer, we instead properly shut down the simulation.

xref #2342

### DIFF
--- a/src/cocotb/__init__.py
+++ b/src/cocotb/__init__.py
@@ -295,13 +295,21 @@ def _initialise_testbench_(argv_):
     except BaseException as e:
         log.error(e)
         simulator.stop_simulator()
-        return
+        _stop_library_coverage()
+        return  # pragma: no cover
 
     global scheduler
     scheduler = Scheduler(handle_result=regression_manager._handle_result)
 
     # start Regression Manager
     regression_manager._execute()
+
+
+def _stop_library_coverage() -> None:
+    if _library_coverage is not None:
+        # TODO: move this once we have normal shutdown behavior to _sim_event
+        _library_coverage.stop()
+        _library_coverage.save()  # pragma: no cover
 
 
 def _sim_event(message):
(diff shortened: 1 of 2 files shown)


marlonjames:
Given
https://github.com/cocotb/cocotb/blob/8e189cb675a7d543b0b8cf6a10884362b2595a9b/src/cocotb/regression.py#L302-L305
maybe we should take a different approach on where to catch errors.

ktbarrett:
Honestly a lot of this functionality is smeared and could use a refactoring.

As things get initialized during startup they should flag it's deconstruction in the reverse order of it's initialization. The library coverage cleanup should be moved to the end of `_sim_event`. Also have you noticed we are now overwriting `_initialize_testbench` and so `COCOTB_LIBRARY_COVERAGE` should no longer work? There was also no saving or writing out of the report?

ktbarrett:
Ah, my eyes deceived me. There's a trailing `_`.

ktbarrett (approved):

src/cocotb/__init__.py
```diff
@@ -290,7 +290,13 @@ def _initialise_testbench_(argv_):
     top = cocotb.handle.SimHandle(handle)
 
     global regression_manager
-    regression_manager = RegressionManager.from_discovery(top)
+    try:
+        regression_manager = RegressionManager.from_discovery(top)
+    except BaseException as e:
+        log.error(e)
+        simulator.stop_simulator()
+        _stop_library_coverage()
```
ktbarrett, line 298: Is this redundant since _sim_event would be called?

ktbarrett: It is not redundant. `simulator.stop_simulator` deregisters the sim end callback (not sure why, this isn't a particularly good idea IMO), so the `_sim_event` callback won't be called otherwise.

PR #3425, opened 2023-09-26
Remove fork

ktbarrett:
It's happening...

PR #3509, opened 2023-11-14
Remove cocotb.coroutine and yield syntax

### LINKED ISSUE #1967: Deprecate cocotb.coroutine and generator-based coroutine support
I think it's a good idea to deprecate the `cocotb.coroutine` decorator and `cocotb.decorators.RunningCoroutine` before the 1.5 release, since I would like to see these removed in 2.0. Before we can deprecate them, we need to remove all internal references. This is the biggest problem with the tests (#1731).

## Rationale
`async`/`await` syntax performs the same job as the `cocotb.coroutine` decorated generators; however, `async` does it better.

- Is a stackful coroutine
  - better performance; no need to trampoline to and from the scheduler
  - stack inspection
- Built into Python
  - better performance; implemented in C
  - existing knowledge
  - can leverage a larger ecosystem of existing tools for advanced functionality

Are there any arguments against it's removal besides backwards compatibility? Is backwards-compatibility into 2.0 desired?

ktbarrett:
Closes #1967.


marlonjames:
`Task.__init__()` should remove this and let the `TypeError` be raised, right? The error message on that needs to be cleaned up as well, it still references `yield`.

https://github.com/cocotb/cocotb/blob/32b8581c8434fbadf2ea6ce35cd97e44b9fc9a45/src/cocotb/task.py#L47-L48

I think it's safe to remove the generator support from `walk_coro_stack()` with this change:

https://github.com/cocotb/cocotb/blob/32b8581c8434fbadf2ea6ce35cd97e44b9fc9a45/src/cocotb/utils.py#L288-L305

And I think a newsfragment is needed for this as well, similar to #3425

docs/source/release_notes.rst
```diff
@@ -99,8 +99,8 @@ Changes
 -------
 
 - Passing :term:`python:coroutine`\ s to :func:`~cocotb.triggers.with_timeout` is now supported. (:pr:`2494`)
-- Renamed ``RunningTask`` to :class:`~cocotb.decorators.Task`. (:pr:`2876`)
-- Made :class:`~cocotb.decorators.Task` interface more like :py:class:`asyncio.Task`'s. (:pr:`2876`)
+- Renamed ``RunningTask`` to :class:`~cocotb.task.Task`. (:pr:`2876`)
+- Made :class:`~cocotb.task.Task` interface more like :py:class:`asyncio.Task`'s. (:pr:`2876`)
```
cmarqu, line 103:
```suggestion
- Made :class:`~cocotb.task.Task` interface more like :python:class:`asyncio.Task`'s. (:pr:`2876`)
```
This is what our intersphinx mapping uses, https://github.com/cocotb/cocotb/blob/32b8581c8434fbadf2ea6ce35cd97e44b9fc9a45/docs/source/conf.py#L52C6-L52C12, but I think strictly speaking, you would not need it at all.

cmarqu (approved):

marlonjames (approved):
The only thing left is the `TypeError` message in `Task.__init__()`. Now is a good time to update it but it could be left for later cleanup.

```py3
        else:
            raise TypeError(
                f"{inst} isn't a valid coroutine! Did you forget to use the yield keyword?"
            )
```

src/cocotb/task.py
```diff
@@ -87,9 +85,8 @@ def _get_coro_stack(self) -> typing.Any:
         coro_stack = extract_coro_stack(self._coro)
 
         # Remove Trigger.__await__() from the stack, as it's not really useful
-        if self._natively_awaitable and len(coro_stack):
-            if coro_stack[-1].name == "__await__":
-                coro_stack.pop()
+        if coro_stack[-1].name == "__await__":
```
marlonjames, line 88:
```suggestion

        if len(coro_stack) and coro_stack[-1].name == "__await__":
```

PR #3510, opened 2023-11-14
Monkeytype integration

AndrewNolte:
This adds a [https://github.com/Instagram/MonkeyType](https://github.com/Instagram/MonkeyType) integration to cocotb. Users can override the MONKEYTYPE_TRACE_MODULES env var to specify modules to trace, and MT_DB_PATH to record the types to a local path. I used this to generate better type hints for cocotb, and I plan on adding those in separate PRs.

Open questions:
- what is the oldest version of python new cocotb is trying to support? this determines the style of type annotations I'll put in
- should this be a cocotb dependency? I think it shouldn't since it's not core to cocotb, and it should give a useful error message if the user tries to use it but doesn't have monkeytype installed. I can add that if requested

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -15,7 +15,7 @@ import nox
 nox.options.sessions = ["dev_test"]
 
 test_deps = ["pytest"]
-coverage_deps = ["coverage", "pytest-cov"]
+coverage_deps = ["coverage", "pytest-cov", "monkeytype"]
 # gcovr 5.1 has an issue parsing some gcov files, so pin to 5.0. See
 # https://github.com/gcovr/gcovr/issues/596
 # When using gcovr 5.0, deprecated jinja2.Markup was removed in 3.1, so an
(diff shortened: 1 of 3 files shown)


ktbarrett:
Is this for dumping the type of the DUT object?

AndrewNolte:
> Is this for dumping the type of the DUT object?

No, and I'm not sure it could be used for that. I think it relies on attrs statically being there; it's mainly for typing function args/return types. But it could be used to help type a user's normal python code

ktbarrett:
Whose function args and return types? Cocotb's internal ones or users? I don't really understand the use case here. You should explain 1. What this is intended to do *exactly* 2. How this solves the issue and 3. Why this needs to be a part of cocotb.

As for your other questions. We are still intending to support Python 3.6. This is definitely be an optional dependency and needs a check and message.

ktbarrett:
We aren't intending to bring this in soon, so we will close this PR for now. If you have rationale, feel free to reopen it.

PR #3513, opened 2023-11-16
add cocotb.parameterize() similar to nox.parameterize()

AndrewNolte:
- cocotb.parameterize() allows for more idiomatic generation of parameterized tests
- factory.generate_tests() now takes test_dec as an arg, meaning normal test() kwargs are now possible to specify on factory tests

### DIFF
--- a/docs/source/newsfragments/3513.feature.rst
+++ b/docs/source/newsfragments/3513.feature.rst
@@ -1,0 +1,1 @@
+Added :func:`cocotb.parameterize`, a decorator that serves as an alternative to TestFactory
--- a/docs/source/newsfragments/3513.feature.rst
+++ b/docs/source/newsfragments/3513.feature.rst
@@ -1,1 +1,1 @@
-Added :func:`cocotb.parameterize`, a decorator that serves as an alternative to TestFactory
+Added :func:`cocotb.parameterize`, a decorator that serves as an alternative to TestFactory.
--- a/docs/source/newsfragments/3513.feature.rst
+++ b/docs/source/newsfragments/3513.feature.rst
@@ -1,1 +1,1 @@
-Added :func:`cocotb.parameterize`, a decorator that serves as an alternative to TestFactory.
+Added :func:`cocotb.parameterize`, a decorator that serves as an alternative to :class:`TestFactory`.
--- a/src/cocotb/__init__.py
+++ b/src/cocotb/__init__.py
@@ -53,6 +53,7 @@ from cocotb.decorators import (  # isort: skip # noqa: F401
     external,
     function,
     test,
+    parameterize,
 )
 from cocotb.log import _filter_from_c, _log_from_c  # isort: skip # noqa: F401
 
--- a/src/cocotb/decorators.py
+++ b/src/cocotb/decorators.py
@@ -29,9 +29,7 @@ import functools
 from typing import Any, Callable, Coroutine, Optional, Sequence, Type, TypeVar, Union
 
 import cocotb
-from cocotb.log import SimLog
-from cocotb.task import Task, _RunningTest
-from cocotb.utils import lazy_property
+from cocotb.regression import Test, TestFactory
 
 Result = TypeVar("Result")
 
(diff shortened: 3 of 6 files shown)


src/cocotb/__init__.py
```diff
@@ -306,8 +307,9 @@ def _sim_event(message):
     # We simply return here as the simulator will exit
     # so no cleanup is needed
     msg = f"Failing test at simulator request before test run completion: {message}"
-    scheduler.log.error(msg)
-    scheduler._finish_scheduler(SimFailure(msg))
+    log.error(msg)
```
AndrewNolte, line 310: When parameterize would get errors, a subsequent error would arise here from scheduler being None

src/cocotb/decorators.py
```diff
@@ -166,7 +168,7 @@ def test(
     expect_error: Union[bool, Type[Exception], Sequence[Type[Exception]]] = (),
     skip: bool = False,
     stage: int = 0,
-) -> Callable[[Callable[..., Coroutine[Any, Any, None]]], Test]:
+) -> Callable[[Callable[..., Coroutine[Any, Any, None]]], Test | None]:
```
AndrewNolte, line 171: Are we allowing the | operator on types? I believe this raises the minimum python version to 3.8 or something like that

ktbarrett:
We haven't discussed raising the required Python version, so it's still 3.6+. I do like this change, I'll try to find some time to review tomorrow. I've been kinda busy recently.

src/cocotb/decorators.py
cmarqu: This becomes possible with `from __future__ import annotations` which is supported since Python 3.7 (https://docs.python.org/3/library/__future__.html).

src/cocotb/__init__.py
ktbarrett: This is being fixed in #3508. Lets leave this change for that PR,

ktbarrett (commented):
There is no need to move `_create_test` and `TestFactory` to decorators.py

AndrewNolte:
> There is no need to move `_create_test` and `TestFactory` to decorators.py

If it was left there would be a circular import, regression.py has ```from cocotb.decorators import Test```, and so this change would then have decorators importing from regression.

ktbarrett:
Ah yes. That crap again. There's probably a better way to go about it. If moving `Test` to regression.py works, I'd prefer that, as it's a better model of whose responsible for what.

AndrewNolte:
> Ah yes. That crap again. There's probably a better way to go about it. If moving `Test` to regression.py works, I'd prefer that, as it's a better model of whose responsible for what.

I don't think this'll work either, because the test() decorator has to check if it's being passed a TestFactory. I think it makes sense to put the Test class and TestFactory class in a separate file, maybe test_types.py? Then decorators.py and regression.py can import from that

ktbarrett:
It should be possible now that `Test` doesn't inherit from `coroutine`.

src/cocotb/decorators.py
AndrewNolte: I'm going to use optional here for 3.6

ktbarrett:
Nice. It just needs a newsfragment. There's a README in /docs/source/newsfragments/ that details how to do that. Thanks for this!

docs/source/newsfragments/3513.feature.rst
```diff
@@ -0,0 +1 @@
+Added :func:`cocotb.parameterize`, a decorator that serves as an alternative to TestFactory
```
ktbarrett, line 1:
```suggestion
Added :func:`cocotb.parameterize`, a decorator that serves as an alternative to TestFactory.
```

docs/source/newsfragments/3513.feature.rst
```diff
@@ -0,0 +1 @@
+Added :func:`cocotb.parameterize`, a decorator that serves as an alternative to TestFactory.
```
cmarqu, line 1:
```suggestion
Added :func:`cocotb.parameterize`, a decorator that serves as an alternative to :class:`TestFactory`.
```

ktbarrett (approved):

src/cocotb/regression.py
```diff
@@ -861,8 +941,8 @@ def generate_tests(self, prefix="", postfix=""):
                 else:
                     doc += "\t{}: {}\n".format(optname, repr(optvalue))
 
-            self.log.debug(
-                'Adding generated test "%s" to module "%s"' % (name, mod.__name__)
+            self.log.info(
```
cmarqu, line 944: Changing the log level is a bit of an unrelated behavioral change which I don't mind but wanted to point out.

src/cocotb/decorators.py
```diff
@@ -216,3 +181,68 @@ def wrapper(f: Callable[..., None]) -> Test:
...
+    This will generate a test for each of the Cartesian products of the parameters and their values.
+
+    .. code-block:: python3
+
+        @cocotb.test(
+            skip=False,
+        )
+        @cocotb.parameterize(
+            arg1=[0,1],
+            arg2=['a','b'],
```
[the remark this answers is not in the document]
ktbarrett: Ah yeah. `ruff` can't do that just yet https://github.com/astral-sh/ruff/issues/7146 =/

cmarqu:
I was going to suggest adding a note about this nice functionality to where we describe how to actually *use* `TestFactory`, but it turns out we don't have such a section at all.

PR #3515, opened 2023-11-19
Add NVC support to runner.py

nickg:
I missed this before because I was running only the make tests and not pytest.


cmarqu (approved):

ktbarrett:
I think this is newsfrag worthy.

ktbarrett:
Actually based on the way the newfrag for #3427 is worded, I think no additional newsfrag is necessary.

ktbarrett (approved):

PR #3517, opened 2023-11-21
fix signed representation of zero

toddstrader:
In this case `_str` is `""` so the translation through the resolve table explodes:
```
ValueError: invalid literal for int() with base 2: ''
```

### DIFF
--- a/src/cocotb/binary.py
+++ b/src/cocotb/binary.py
@@ -413,6 +413,8 @@ class BinaryValue:
     @property
     def signed_integer(self):
         """The signed integer representation of the underlying vector."""
+        if len(self._str) == 0:
+            return 0
         ival = int(self._str.translate(_resolve_table), 2)
         bits = len(self._str)
         signbit = 1 << (bits - 1)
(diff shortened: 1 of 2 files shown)


ktbarrett:
`BinaryValue` will be deleted by the next release, so I don't really see a reason to bother with this.

toddstrader:
> `BinaryValue` will be deleted by the next release, so I don't really see a reason to bother with this.

@ktbarrett is there a thread which discusses the migration plan?  Especially wrt handle access?  E.g. what will `ModifiableObject.value` return in the future and is there a way to behave that way now?

ktbarrett:
There is #3244. But unfortunately the work has paused and we realize there are even gaps in the current design of LogicArray which will require us breaking the current interface. So I don't think there is any way to future proof that I can see right now. But `ModifiableObject.value` will return a `LogicArray` or similar.

toddstrader:
OK, I'll hold onto this commit locally, but I'll reserve the right to reopen this PR if we find more or rebasing becomes challenging.  Our entire hope right now is to get off of our cocotb fork, so I'm trying to avoid creating a new one.

ktbarrett:
Right now what you can do is use LogicArray, just cast the handle value like `LogicArray(handle.value)`. That should be more correct than anything coming out of `BinaryValue` and the closest thing to interface compatable.

PR #3520, opened 2023-11-22
run ruff --fix

AndrewNolte:
This adds isort, pyupgrade, and pylint to pyproject.toml. I then ran ruff --fix, added some rule ignores, and manually fixed a few minor violations

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,12 @@ extend-exclude = [
     "_vendor",
     ".nox/",
 ]
+target-version = "py37"
+select = [
+    "I",  # isort
+    "UP", # pyupgrade
+    "PL", # pylint
+]
 ignore = [
     # ambiguous variable name (preference)
     "E741",
@@ -52,8 +58,11 @@ ignore = [
     "E501",
     # whitespace before ':' (black)
     "E203",
-    # line break before binary operator (black)
-    # "W503",  # not implemented by ruff
+    "PLR0912", # Too many branches (>12) (preference)
+    "PLR0913", # Too many arguments to function call (>5) (preference)
+    "PLR0915", # Too many statements (>50) (preference)
+    "PLR2004", # Magic value used in comparison (preference)
+    "PLW0603", # Using the global statement (preference)
 ]
 
 [tool.ruff.per-file-ignores]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ exclude: "^.*/_vendor/"
 repos:
 # ruff with --fix should run before other formatting tools
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: "v0.1.4"
+  rev: "v0.1.6"
   hooks:
     # Run the linter.
     - id: "ruff"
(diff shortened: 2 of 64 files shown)


ktbarrett:
Considering this a follow on from #3499. That PR missed enabling the isort rules which it intended, as only flake8 rules are enabled by default. In addition, pyupgrade and pylint rules are added. The version is specified (presumably needed for pyupgrade features) which was missing before.

Some things:
1. We are targeting Python 3.6. Even though the Python Foundation considers it EOL, we are supporting RHEL8 through its lifecycle which uses Python 3.6 as the standard Python version. Outside of some typing functionality (typing_extensions helps), this isn't a problem so I don't see us deciding to move away from this.
2. There seems to be some needless formatting changes in the pyproject.toml, I'm guessing that your editor is doing it automatically? `git add -p` helps.
3. We might need to bump pre-commit to get the lint to pass, `pre-commit autoupdate` works.

pyproject.toml
```diff
@@ -45,6 +45,12 @@ extend-exclude = [
     "_vendor",
     ".nox/",
 ]
+target-version = "py37"
```
cmarqu, line 48: I'd set this via `project.requires-python` even though this then could have an influence on other tools, but it would be the "right influence": https://docs.astral.sh/ruff/settings/#target-version

cmarqu:
Was the conversion to f-string also done by ruff? If not, this should maybe be a different commit/PR so that we can use https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

ktbarrett:
I agree. The fix should be a separate commit so we can blame ignore it.

ktbarrett:
Sorry @AndrewNolte for taking over. It's boring back here in rural Maryland. Happy Thanksgiving!

AndrewNolte:
> Sorry @AndrewNolte for taking over. It's boring back here in rural Maryland. Happy Thanksgiving!

Oh no worries, thanks for fixing this! Yeah the toml change was done by an extension I have, I should've turned that off. And @cmarqu for the python 3.7, that's unfortunately just the minimum version that ruff has for that setting, although it doesn't look like that broke anything here. I'm guessing if they use standard requires-python version it'll just snap to 3.7.

btw @ktbarrett it looks like it's just missing name: "cocotb" under [project] for those CI runs ^

pyproject.toml
ktbarrett: Looks like we are going to punt on this issue.

1. ruff doesn't support Python 3.6 so we can't make `target-version = "py36"`
2. We can make `project.requires-python = ">=3.6"` but ruff just internally bumps this to 3.7.
3. Specifying `[project]` requires specifying multiple fields
4. A full move to using pyproject for project metadata isn't possible yet (#3529).

ktbarrett:
@AndrewNolte btw, if you are going to be pushing from work, I suggest you add your work email to your github account so all your commits link back to you.

cmarqu:
Let's add `8607c63feb128c46ceb803778aee6cbad6b25704` (https://github.com/cocotb/cocotb/pull/3520/commits/8607c63feb128c46ceb803778aee6cbad6b25704) to 
https://github.com/cocotb/cocotb/blob/master/.git-blame-ignore-revs also.

ktbarrett:
@cmarqu That will have to be a follow on. We can't anticipate what the rebased commit will be.

PR #3521, opened 2023-11-22
Fix docs

ktbarrett:
Some small fixes to the docs.

### DIFF
--- a/docs/source/install_devel.rst
+++ b/docs/source/install_devel.rst
@@ -80,4 +80,4 @@ The development version of cocotb can be installed by running
 
 After installation, you should be able to execute ``cocotb-config``.
 If it is not found, you need to append its location to the ``PATH`` environment variable.
-This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented :ref:`here<python:inst-alt-install-user>`.
+This may happen when you use the ``--user`` option to ``pip``.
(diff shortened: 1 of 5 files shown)


docs/source/install_devel.rst
```diff
@@ -80,4 +80,4 @@ The development version of cocotb can be installed by running
 
 After installation, you should be able to execute ``cocotb-config``.
 If it is not found, you need to append its location to the ``PATH`` environment variable.
-This may happen when you use the ``--user`` option to ``pip``, in which case the location is documented :ref:`here<python:inst-alt-install-user>`.
+This may happen when you use the ``--user`` option to ``pip``.
```
ktbarrett, line 83: This section of the documentation no longer exists, it's been totally rewritten and there isn't an equivalent section to link to.

docs/source/troubleshooting.rst
```diff
@@ -73,7 +73,7 @@ Debug the test ``test_array_simple`` with Questa, using the VHDL toplevel and th
 
 2. Open a new terminal window or tab, and attach GDB to the running process.
 
-  .. code-block:: shell-session
+  .. code-block::
```
ktbarrett, line 76: The shell syntax parsing failed here since we break into a gdb shell half-way through. Highlighting isn't the most important thing, so I just removed it.

cmarqu (approved):

PR #3529, opened 2023-11-24
Use pyproject.toml for all configuration

ktbarrett:
Allows us to remove setup.cfg file. Moves all project metadata and pytest and coverage configurations to pyproject.toml. ~~Uses a slightly different mechanism to get project version.~~ EDIT: That don't work before Python 3.8.


ktbarrett:
Ugh... Python 3.6 really doesn't like this. I'm guessing it can't find a version of setuptools new enough to support this. Python 3.7 doesn't work either, but that's because `importlib.metadata` isn't available until Python 3.8.

PR #3541, opened 2023-11-28
Add  to GHDL makefile to simulate different VHDL architectures from the top level

### LINKED ISSUE #3525: GHDL specify architecture to use of VHDL entity
As far as I can tell, when simulating VHDL using GHDL using the latest version of CoCoTB, there's no way to specify an architecture of a dut to use from CoCoTB, either the python test, or the Makefile to drive the test. Is that correct? 

It looks like it may be possible to add the architecture via make variable to the Makefile.ghdl, but i wanted to double check this isn't easily solved some other way before i start messing with the Makefiles and then try to figure out how it would also work for the new pythonic test runs.

forensicgarlic:
Closes #3525.

This will make it possible with the MAKE build flow to specify the VHDL architecture to use for a specific test. Without this, the architecture would default to the last one in a VHDL file and can not be specified for COCOTB driven tests.

### DIFF
--- a/docs/source/simulator_support.rst
+++ b/docs/source/simulator_support.rst
@@ -343,6 +343,8 @@ In order to use this simulator, set :make:var:`SIM` to ``ghdl``:
 
 Noteworthy is that despite GHDL being a VHDL simulator, it implements the :term:`VPI` interface.
 
+To specify a VHDL architecture to simulate, use the `$(ARCH)` make variable
+
 .. _sim-ghdl-issues:
 
 Issues for this simulator
--- a/docs/source/simulator_support.rst
+++ b/docs/source/simulator_support.rst
@@ -343,7 +343,7 @@ In order to use this simulator, set :make:var:`SIM` to ``ghdl``:
 
 Noteworthy is that despite GHDL being a VHDL simulator, it implements the :term:`VPI` interface.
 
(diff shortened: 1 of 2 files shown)


docs/source/custom_flows.rst
```diff
@@ -170,6 +170,8 @@ GHDL
 * Extend the ``ghdl -r`` call with the option
   ``--vpi=$(cocotb-config --lib-name-path vpi ghdl)``.
 
+* VHDL architectures can be specified via the `$(ARCH)` make variable
+
```
marlonjames, line 174: This should go in `simulator_support.rst` under the GHDL section.

forensicgarlic: ok, moving it there.

docs/source/simulator_support.rst
```diff
@@ -343,6 +343,8 @@ In order to use this simulator, set :make:var:`SIM` to ``ghdl``:
 
 Noteworthy is that despite GHDL being a VHDL simulator, it implements the :term:`VPI` interface.
 
+To specify a VHDL architecture to simulate, use the `$(ARCH)` make variable
```
ktbarrett, line 346:
```suggestion
To specify a VHDL architecture to simulate, set the ``ARCH`` make variable to the architecture name.
```

ktbarrett (approved):

marlonjames (approved):

PR #3547, opened 2023-12-01
Replace lazy_property with cached_property

ktbarrett:
We don't use lazy_property's implementation before Python 3.8 because sphinx doesn't detect them as valid properties and generate documentation for them. Following on from https://github.com/cocotb/cocotb/pull/2720#discussion_r1412165833.

### DIFF
--- a/src/cocotb/_py_compat.py
+++ b/src/cocotb/_py_compat.py
@@ -64,7 +64,7 @@ else:
 
 
 # simple, but less than optimal backport of Python 3.8's cached_property
-if sys.version_info >= (3, 12):
+if sys.version_info >= (3, 8):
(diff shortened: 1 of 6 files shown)


src/cocotb/_py_compat.py
```diff
@@ -61,3 +61,13 @@ def __exit__(self, *excinfo):
     import collections
 
     insertion_ordered_dict = collections.OrderedDict
+
+
+# simple, but less than optimal backport of Python 3.8's cached_property
+if sys.version_info >= (3, 12):
```
ktbarrett, line 67:
```suggestion
if sys.version_info >= (3, 8):
```

PR #3553, opened 2023-12-02
Fix test_parallel_cocotb.py test with NVC

nickg:
NVC has two kinds of arguments: "global" ones (e.g. for setting standard version, library path, etc.) and command-specific ones (e.g. to set generics for elaboration). In `runner.py` I borrowed the `build_args` property for the global arguments but that is only set if `build()` is called on the `Runner` object before `test()` which `test_parallel_cocotb` doesn't do. 

This patch seems like a bit of a hack so I'm not sure it's the best way to resolve this. An alternative might be to add an `extra_args` property that can be set on `build()` and `test()` similar to the makefile.

### DIFF
--- a/src/cocotb/runner.py
+++ b/src/cocotb/runner.py
@@ -154,6 +154,7 @@ class Simulator(abc.ABC):
         verbose: bool = False,
         timescale: Optional[Timescale] = None,
         waves: Optional[bool] = None,
+        extra_args: Sequence[str] = [],
     ) -> None:
         """Build the HDL sources.
 
@@ -172,6 +173,7 @@ class Simulator(abc.ABC):
             verbose: Enable verbose messages.
             timescale: Tuple containing time unit and time precision for simulation
             waves: Record signal traces.
+            extra_args: Additional arguments required by some simulators.
         """
 
         self.clean: bool = clean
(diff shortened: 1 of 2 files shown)


marlonjames (approved):

ktbarrett:
@themperek What say you about the addition of `extra_args`. Personally, I think the name is terrible.

themperek:
I know this is not the nicest thing to do but I would be in favour of passing same to `test_args` and `build_args` rather than creating new one. Would this work for you?

marlonjames:
We could name it something like
```
general_args: A list of general arguments to the simulator that may be independent of simulator command.
    Examples include selecting HDL standard version and setting paths to library, project, or other support files.
```

ktbarrett:
Before we go off too far, the issue is that NVC requires two sets of args in two places. One after the `-e` (elab) option and another after the `-r` (run) option. Currently he is reusing the `build_args` from the build step to pass to the elab step. This breaks test_parallel_cocotb as the simulator object is not shared between the compile step and the multiple run steps. The only way to set build_args needed by the `test()` command is by calling the `build()` command.

marlonjames:
> The only way to set build_args needed by the `test()` command is by calling the `build()` command.

That's only because we don't have a separate args list that is shared between `test()` and `build()`, so he was re-using `build_args`.

marlonjames:
What about a separate method on `Simulator` that sets the common args?

ktbarrett:
@marlonjames They are public methods, you can just "build" the object back into what it was in the compile step by setting the public attributes. That's what I did in #3559.

ktbarrett:
We chose #3559 over this. This requires more discussion on our part.

PR #3554, opened 2023-12-03
Update NVC expected results for test_discovery

nickg:
I fixed discovering the string elements in nickg/nvc@3ce014f.

I think `SAMPLE_BLOCK.clk_inv` gets discovered twice because it's returned once when iterating over `vhpiDecls` and again when iterating `vhpiSigDecls`. I can't find anything in the LRM which specifies what `vhpiDecls` is supposed to return so I'm happy to change it to behave the same as other simulators if someone can tell me what that is.

### DIFF
--- a/tests/test_cases/test_discovery/test_discovery.py
+++ b/tests/test_cases/test_discovery/test_discovery.py
@@ -523,9 +523,8 @@ async def discover_all_in_component_vhdl(dut):
         # doesn't find elements of stream_in_data or stream_out_data_registered
         assert total_count == 7
     elif sim.startswith("nvc"):
-        # doesn't find elements of string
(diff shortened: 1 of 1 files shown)


ktbarrett:
> I fixed discovering the string elements in [nickg/nvc@3ce014f](https://github.com/nickg/nvc/commit/3ce014f).

Nice!

>I think SAMPLE_BLOCK.clk_inv gets discovered twice because it's returned once when iterating over vhpiDecls and again when iterating vhpiSigDecls. I can't find anything in the LRM which specifies what vhpiDecls is supposed to return so I'm happy to change it to behave the same as other simulators if someone can tell me what that is.

@marlonjames?

ktbarrett (approved):

PR #3559, opened 2023-12-06
Fix test_parallel_cocotb

ktbarrett:
Alternative to #3553.


marlonjames (approved):
This fixes the problem. We can continue discussion about the runner API later.

PR #3564, opened 2023-12-07
Fix broken newsfrag

ktbarrett:


marlonjames:
While you are here can you fix `2255.bugfix.rst`? It is a multi-line newsfragment that is also broken.

PR #3558, opened 2023-12-06
Fix defaultless parameters for Xcelium

toddstrader:
@ktbarrett this commit broke defaultless parameter access for Xcelium:
e15b23f7a8c0125e54b6ee6c75b4d0d2403d76c9

I'm actively investigating and will report back with what I find.

PR #3567, opened 2023-12-08
Fix test_package RTL location

toddstrader:
As discussed in #3558 the `test_package` RTL did not need to be in `tests/designs`.


ktbarrett (approved):

ktbarrett:
Well we are ignoring the MSVC test for now, so this is good to go.

PR #3568, opened 2023-12-08
Add git rev to local version identifier

### LINKED ISSUE #3562: Thoughts on adding git rev to dev version numbers?
How would everyone feel about adding `git rev-parse --short HEAD` to `__version__`?  Currently it is `2.0.0.dev0` and I'm not sure what the final `0` is for, but I'm guessing it's to manually increment as one feels like?  What if instead it were `2.0.0.dev-d379318e`?  I'd propose that we build the file on the fly in `setup.py` but only add the git rev when the non-suffixed version (stored as a variable in `setup.py`) includes `dev`.

I can take a stab at this if people are onboard.

toddstrader:
Adds the Git rev to the local version label.  There's a chicken-and-egg problem wrt PR numbers and creating news fragments.  I'll add that next.  Also, should I add a test for this?  I guess I could check that the `+` exists and there's something git rev-like if it's a `dev` version?

closes #3562


setup.py
```diff
@@ -69,8 +70,20 @@ def package_files(directory):
     return paths
 
 
-# this sets the __version__ variable
-exec(read_file(path.join("src", "cocotb", "_version.py")))
+__version__ = "2.0.0.dev0"
```
ktbarrett, line 73: While we're here, let's sort out what the proper version string should be.

marlonjames:
I wonder why Python 3.6 doesn't work.

https://github.com/cocotb/cocotb/actions/runs/7143704333/job/19455778417?pr=3568#step:23:65

toddstrader:
> I wonder why Python 3.6 doesn't work.
> 
> https://github.com/cocotb/cocotb/actions/runs/7143704333/job/19455778417?pr=3568#step:23:65

Huh, dunno yet.  But I guess that answers my question about needing a test.

setup.py
ktbarrett: Well we aren't releasing 2.0.0, we are releasing 2.0, so I'd think 2.0.dev0 or 2.dev0 would be more appropriate, the PEP would shed more light.

toddstrader:
3.6 was failing because `text` was too new for `subprocess`:
> Changed in version 3.7: Added the text parameter, as a more understandable alias of universal_newlines. Added the capture_output parameter.

I confirmed that the next test caught this issue and then resolved it.

toddstrader:
Also additional confirmation that things are working in 3.6 now:
https://github.com/cocotb/cocotb/actions/runs/7145071037/job/19459990612?pr=3568#step:23:65

setup.py
ktbarrett: It will be `2.0`. In the past the initial release has always been `MAJOR.MINOR` and patch versions only added after the fact. So let's change that to `2.0.dev0` and we can close that out.

toddstrader: Done.

ktbarrett:
Gentle reminder on the ready for review.

toddstrader:
> Gentle reminder on the ready for review.

I respond well to loud yelling.

ktbarrett:
GOOD JOB, YOU DID IT!@

ktbarrett (approved):

ktbarrett:
Reading deeper into PEP 440:

https://peps.python.org/pep-0440/#final-releases
> When comparing release segments with different numbers of components, the shorter segment is padded out with additional zeros as necessary.

I read this as meaning `2.0.0.dev0` and `2.0.dev0` are the same. If this is a point of contention, I'm willing to concede.

marlonjames:
I don't really care `2.0` vs `2.0.0`, I think it was probably to match the 3 levels we have used since `1.2.0`.

toddstrader:
OK, so what's the verdict? Back to `2.0.0`?

ktbarrett:
Yeah, sorry.

setup.py
```diff
@@ -69,8 +70,20 @@ def package_files(directory):
...
+        rev = subprocess.check_output(
+            ["git", "rev-parse", "--short", "HEAD"], universal_newlines=True
+        ).strip()
+        __version__ += f"+{rev}"
+    except Exception as e:
+        print(e, file=sys.stderr)
+with open(path.join("src", "cocotb", "_version.py"), "w") as f:
+    f.write("# Package version\n")
+    f.write("# Generated by setup.py -- do not modify directly\n\n")
+    f.write(f'__version__ = "{__version__}"')
```
marlonjames, line 85: sdist installation doesn't work, as it is installed from the `tar.gz` file, not a git repo.
You get this in the output:
```
  fatal: not a git repository (or any of the parent directories): .git
  Command '['git', 'rev-parse', '--short', 'HEAD']' returned non-zero exit status 128.
```

But, `_version.py` is included in the sdist, so use it in that case:

```suggestion
version_file_path = path.join("src", "cocotb", "_version.py")
__version__ = "2.0.0.dev0"
if "dev" in __version__:
    try:
        rev = subprocess.check_output(
            ["git", "rev-parse", "--short", "HEAD"], universal_newlines=True
        ).strip()
        __version__ += f"+{rev}"
    except Exception as e:
        # if this is not a git repository and _version.py already exists,
        # we are probably installing from an sdist, so use the existing _version.py
        if path.exists(version_file_path):
            exec(read_file(version_file_path))
        else:
            print(e, file=sys.stderr)
with open(version_file_path, "w") as f:
    f.write("# Package version\n")
    f.write("# Generated by setup.py -- do not modify directly\n\n")
    f.write(f'__version__ = "{__version__}"'
```

toddstrader: How can I reproduce this?  I would have assumed catching the exception in the case of not installing from Git would have let setup.py finish and create _version.py just without the `+hash`>

marlonjames: I edited my comment as it may have been unclear. It does install, but it is just `2.0.0.dev0` without the `+hash` suffix.
But the correct local version exists in the sdist `_version.py`, it is just overwritten by the default when it doesn't need to be.

setup.py
```diff
@@ -69,8 +70,26 @@ def package_files(directory):
...
+        # if this is not a git repository and _version.py already exists,
+        # we are probably installing from an sdist, so use the existing _version.py
+        if path.exists(version_file_path):
+            exec(read_file(version_file_path))
+        else:
+            print(e, file=sys.stderr)
+with open(version_file_path, "w") as f:
+    f.write("# Package version\n")
+    f.write("# Generated by setup.py -- do not modify directly\n\n")
+    f.write(f'__version__ = "{__version__}"'
```
marlonjames, line 91:
```suggestion
    f.write(f'__version__ = "{__version__}"')
```

imphil (approved):
Thanks for your pull request, @toddstrader.

> There's a chicken-and-egg problem wrt PR numbers and creating news fragments.

You can use the issue number instead of the PR number for the newsfragment.

PR #3578, opened 2023-12-11
Refactoring RegressionManager and test registration

### LINKED ISSUE #3461: Change how tests are registered in `RegressionManager`
This is inspired by [the issues seen with pyuvm](https://github.com/pyuvm/pyuvm/issues/89) and also with my framework which has the same issue. Currently, decorating an object with `cocotb.test` returns a different type, if it were a class, you would not be able to subclass that test to make another test. Instead, we could have the `cocotb.test` decorator simply register the test callable with `RegressionManager` and return it unmolested allowing this behavior.

ktbarrett:
This PR was initially intended to close #3461. It makes changes to the `RegressionManager`, and the `test` decorator and `TestFactory` such that tests are "registered" to the `RegressionManager` when the decorator is invoked, or `TestFactory.generate` is called. This allows the decorated test function to be returned by the `test` decorator.

There were several follow on changes:
* The way TESTCASE worked previously doesn't really work with this method, so it was changed to be a general-purpose test filter, which filters tests (whitelist) according to comma-separated regular expressions.
* The `_Test` class is now inaccessible by the user, so it was made private. Some cleanup was done on this class, especially with regards to the abuse of special attributes (`__name__`, etc.)
* The regression start time is set in better place to make it more accurate (the new `start_regression` method)
* Documentation of `test` decorator and `TestFactory` are greatly improved.
* Typing of `test` decorator and `TestFactory` are greatly improved.
* The functionality added by #3387 was accidentally broken at some point and we didn't know because the test wasn't great. So the functionality was fixed, documentation improved, and the test improved.
* The use of "registration" means we can override test names, so a `name` parameter was added to the `test` decorator and `TestFactory`.
* The `name` parameter makes the `prefix` and `postfix` parameter of `TestFactory.generate` superfluous, so they were deprecated (#2575).
* cocotb library coverage collection setup was moved into its own function and improved (there was no check if `coverage` was not installed)
* cocotb setup errors are better covered by moving the exception handling to the root-most function call in Python (this will be handled in C++ in the future).

TODO:
- [x] We need to decide if no tests being discovered is an error or not
- [x] Add newsfrags
- [x] change how skip and TESTCASE interact
- [x] remove oopsies

### DIFF
--- a/src/cocotb/__init__.py
+++ b/src/cocotb/__init__.py
@@ -202,19 +202,15 @@ def _initialise_testbench(argv_):  # pragma: no cover
     :envvar:`MODULE` and :envvar:`TESTCASE`.
     """
     with _rlock:
-        if "COCOTB_LIBRARY_COVERAGE" in os.environ:
-            import coverage
+        try:
+            _start_library_coverage()
+            _initialise_testbench_(argv_)
+        except BaseException:
+            log.exception("cocotb testbench initialization failed. Exiting.")
+            from cocotb import simulator
 
-            global _library_coverage
-            _library_coverage = coverage.coverage(
-                data_file=".coverage.cocotb",
-                config_file=False,
-                branch=True,
-                include=[f"{os.path.dirname(__file__)}/*"],
-            )
-            _library_coverage.start()
-
-        _initialise_testbench_(argv_)
+            simulator.stop_simulator()
+            _stop_library_coverage()
 
 
 def _initialise_testbench_(argv_):
@@ -283,8 +279,6 @@ def _initialise_testbench_(argv_):
     random.seed(RANDOM_SEED)
 
     # Setup DUT object
-    from cocotb import simulator
-
     handle = simulator.get_root_handle(root_name)
     if not handle:
         raise RuntimeError(f"Can not find root handle ({root_name})")
@@ -295,20 +289,44 @@ def _initialise_testbench_(argv_):
     _start_user_coverage()
 
     global regression_manager
-    try:
-        regression_manager = RegressionManager.from_discovery(top)
-    except BaseException as e:
-        log.error(e)
-        simulator.stop_simulator()
-        _stop_user_coverage()
-        _stop_library_coverage()
-        return  # pragma: no cover
+    regression_manager = RegressionManager()
+
+    # discover tests
+    module_str = os.getenv("MODULE").strip()
+    if module_str is None or len(module_str) == 0:
+        raise RuntimeError(
+            "Environment variable MODULE, which defines the module(s) to execute, is not defined or empty."
+        )
+    modules = [s.strip() for s in module_str.split(",") if s.strip()]
+    test_str = os.getenv("TESTCASE", "")
+    filters = [s.strip() for s in test_str.split(",") if s.strip()]
+    regression_manager.setup_pytest_assertion_rewriting(modules)
+    regression_manager.discover_tests(modules=modules, filters=filters)
 
     global scheduler
     scheduler = Scheduler(handle_result=regression_manager._handle_result)
 
     # start Regression Manager
-    regression_manager._execute()
+    regression_manager.start_regression()
+
+
+def _start_library_coverage() -> None:  # pragma: no cover
+    if "COCOTB_LIBRARY_COVERAGE" in os.environ:
+        try:
+            import coverage
+        except ImportError as e:
+            log.error(
+                "cocotb library coverage collection requested but coverage package not available. Install it using `pip install coverage`."
+            )
+        else:
+            global _library_coverage
+            _library_coverage = coverage.coverage(
+                data_file=".coverage.cocotb",
+                config_file=False,
+                branch=True,
+                include=[f"{os.path.dirname(__file__)}/*"],
+            )
+            _library_coverage.start()
 
 
 def _stop_library_coverage() -> None:
--- a/src/cocotb/decorators.py
+++ b/src/cocotb/decorators.py
@@ -212,10 +212,12 @@ def test(
         if isinstance(_func, TestFactory):
             if cocotb.regression_manager is not None:
                 _func.generate_tests()
+            _func.test_function.__cocotb_test__ = True
             return _func.test_function
         else:
             if cocotb.regression_manager is not None:
                 cocotb.regression_manager.register_test(func=_func)
+            _func.__cocotb_test__ = True
             return _func
 
     def wrapper(f: Union[F, TestFactory[F]]) -> F:
@@ -230,6 +232,7 @@ def test(
                     skip=skip,
                     stage=stage,
                 )
+            f.test_function.__cocotb_test__ = True
             return f.test_function
 
         else:
@@ -244,6 +247,7 @@ def test(
                     skip=skip,
                     stage=stage,
                 )
+            f.__cocotb_test__ = True
             return f
 
     return wrapper
--- a/src/cocotb/decorators.py
+++ b/src/cocotb/decorators.py
@@ -36,10 +36,11 @@ from typing import (
     Type,
     TypeVar,
     Union,
+    overload,
 )
 
 import cocotb
-from cocotb.regression import Test, TestFactory
+from cocotb.regression import TestFactory
 
 Result = TypeVar("Result")
 
@@ -83,25 +84,68 @@ def external(func: Callable[..., Result]) -> Callable[..., Coroutine[Any, Any, R
     return wrapper
 
 
+F = TypeVar("F", bound=Callable[..., Coroutine[Any, Any, None]])
+
+
+@overload
+def test(_func: Union[F, TestFactory[F]]) -> F:
+    ...
+
+
+@overload
 def test(
+    *,
     timeout_time: Optional[float] = None,
     timeout_unit: str = "step",
     expect_fail: bool = False,
-    expect_error: Union[bool, Type[Exception], Sequence[Type[Exception]]] = (),
+    expect_error: Union[Type[Exception], Sequence[Type[Exception]]] = (),
     skip: bool = False,
     stage: int = 0,
-) -> Callable[[Callable[..., Coroutine[Any, Any, None]]], Optional[Test]]:
-    """
-    Decorator to mark a Callable which returns a Coroutine as a test.
+    name: Optional[str] = None,
+) -> Callable[[Union[F, TestFactory[F]]], F]:
+    ...
 
-    The test decorator provides a test timeout, and allows us to mark tests as skipped
-    or expecting errors or failures.
+
+def test(
+    _func: Optional[Union[F, TestFactory[F]]] = None,
+    *,
+    timeout_time: Optional[float] = None,
+    timeout_unit: str = "step",
+    expect_fail: bool = False,
+    expect_error: Union[Type[Exception], Sequence[Type[Exception]]] = (),
+    skip: bool = False,
+    stage: int = 0,
+    name: Optional[str] = None,
+) -> Callable[[Union[F, TestFactory[F]]], F]:
+    """
+    Decorator to register a Callable which returns a Coroutine as a test.
+
+    The test decorator provides a test timeout, and allows us to mark tests as skipped or expecting errors or failures.
     Tests are evaluated in the order they are defined in a test module.
 
-    Used as ``@cocotb.test(...)``.
+    Usage:
+        .. code-block:: python3
+
+            @cocotb.test(timeout_time=10, timeout_unit="ms")
+            async def test_thing(dut):
+                ...
+
+    .. versionchanged:: 2.0
+        Support using decorator on test function without supplying parameters first.
+
+        Assumes all default values for the test parameters.
+
+        .. code-block:: python3
+
+            @cocotb.test
+            async def test_thing(dut):
+                ...
+
+    .. versionchanged:: 2.0
+        Decorated tests now return the decorated object.
 
     Args:
-        timeout_time (numbers.Real or decimal.Decimal, optional):
+        timeout_time:
             Simulation time duration before timeout occurs.
 
             .. versionadded:: 1.3
@@ -111,7 +155,7 @@ def test(
                 Users should use :class:`~cocotb.triggers.with_timeout` if they require a
                 more general-purpose timeout mechanism.
 
-        timeout_unit (str, optional):
+        timeout_unit:
             Units of timeout_time, accepts any units that :class:`~cocotb.triggers.Timer` does.
 
             .. versionadded:: 1.3
@@ -119,12 +163,12 @@ def test(
             .. versionchanged:: 2.0
                 Passing ``None`` as the *timeout_unit* argument was removed, use ``'step'`` instead.
 
-        expect_fail (bool, optional):
+        expect_fail:
             If ``True`` and the test fails a functional check via an ``assert`` statement, :class:`pytest.raises`,
             :class:`pytest.warns`, or :class:`pytest.deprecated_call` the test is considered to have passed.
             If ``True`` and the test passes successfully, the test is considered to have failed.
 
-        expect_error (exception type or tuple of exception types, optional):
+        expect_error:
             Mark the result as a pass only if one of the exception types is raised in the test.
             This is primarily for cocotb internal regression use for when a simulator error is expected.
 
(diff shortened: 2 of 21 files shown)


src/cocotb/regression.py
```diff
@@ -176,109 +168,78 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
-
-        if module_str is None:
-            raise ValueError(
-                "Environment variable MODULE, which defines the module(s) to execute, is not defined."
-            )
+        for module_name in modules:
+            self.log.debug("Searching for tests in module %s", module_name)
+            self.log.debug("Python Path: %s", ",".join(sys.path))
+            self.log.debug("PWD: %s", os.getcwd())
+            __import__(module_name)
```
marlonjames, line 191: This will return the top-level package if `module_name` is multi-level, such as `my_package.my_test_module`.
If we want to replace `_my_import()`, let's use [`importlib.import_module()`](https://docs.python.org/3/library/importlib.html#importlib.import_module).

We also need a testcase for that...

src/cocotb/regression.py
```diff
@@ -176,109 +168,78 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
-                        yield thing
-
-        # If any test were not found in any module, raise an error
-        if tests:
-            _logger.error(
-                "Requested test(s) %s wasn't found in module(s) %s", tests, modules
-            )
-            raise AttributeError(f"Test(s) {tests} doesn't exist in {modules}")
+        Args:
+            test: The test to register.
```
marlonjames, line 225: Incomplete?

src/cocotb/regression.py
```diff
@@ -807,28 +741,42 @@ class TestFactory:
 
     .. versionchanged:: 1.5
         Groups of options are now supported
-    """
 
-    # Prevent warnings from collection of TestFactories by unit testing frameworks.
-    __test__ = False
+    .. versionchanged:: 2.0
+        You can now specifying :func:`cocotb.test` decorator arguments when generating tests.
```
marlonjames, line 746:
```suggestion
        You can now specify :func:`cocotb.test` decorator arguments when generating tests.
```

src/cocotb/decorators.py
```diff
@@ -146,39 +190,50 @@ async def my_test(dut):
...
+        name:
+            Override the default name of the test.
+            The default test name is the ``__qualname__`` of the decorated test function.
+
+            .. versionadded:: 2.0
     """
 
-    def wrapper(f: Union[Callable[..., None], TestFactory]) -> Optional[Test]:
+    if _func is not None:
+        cocotb.regression_manager.register_test(func=_func)
```
marlonjames, line 209: Looks like when the test module is loaded outside sim environment (like when using Python runner) this fails because `cocotb.regression_manager` is `None`.

src/cocotb/regression.py
ktbarrett: We aren't using the return value anymore. Does `__import__` not execute `my_test_module.py` in that case? I'm fine with leaving `_my_import` as is if it is necessary.

src/cocotb/decorators.py
ktbarrett: I was never quite happy we allowed that. Oh well. Special case added.

ktbarrett:
`test_select_testcase` is broken now with the change in how TESTCASE works. This test should be rewritten to test the new behavior.

src/cocotb/regression.py
marlonjames: I think you are right.

src/cocotb/__init__.py
```diff
@@ -295,20 +289,44 @@ def _initialise_testbench_(argv_):
...
+    regression_manager.start_regression()
+
+
+def _start_library_coverage() -> None:  # pragma: no cover
+    if "COCOTB_LIBRARY_COVERAGE" in os.environ:
+        try:
+            import coverage
+        except ImportError as e:
+            log.error(
+                "cocotb library coverage collection requested but coverage module not available. Install it using `pip install coverage`."
```
cmarqu, line 319:
```suggestion
                "cocotb library coverage collection requested but coverage package not available. Install it using `pip install coverage`."
```

src/cocotb/decorators.py
marlonjames: No happy about allowing what? The global being `None` ?

ktbarrett: Suggesting other tools read files with cocotb code in them.

src/cocotb/decorators.py
```diff
@@ -146,39 +190,58 @@ async def my_test(dut):
...
             manually by setting :make:var:`TESTCASE`.
 
-        stage (int)
+        stage:
             Order tests logically into stages, where multiple tests can share a stage.
             Defaults to 0.
+
+        name:
+            Override the default name of the test.
+            The default test name is the ``__qualname__`` of the decorated test function.
```
cmarqu, line 203:
```suggestion
            The default test name is the :any:`__qualname__` of the decorated test function.
```
Will hopefully point to https://docs.python.org/3/library/stdtypes.html?highlight=__qualname__#definition.__qualname__

src/cocotb/decorators.py
marlonjames: Ah. I would normally put them in separate modules, but even our tests include the pytest test and cocotb test in the same file.

src/cocotb/regression.py
```diff
@@ -176,109 +168,109 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
-        ####################
-        self._queue = []
-        for test in tests:
-            self.log.info(f"Found test {test.__module__}.{test.__qualname__}")
-            self._queue.append(test)
-        self.ntests = len(self._queue)
+    def discover_tests(
+        self, modules: Iterable[str], filters: Optional[Sequence[str]] = None
+    ) -> None:
+        """Discovers tests in files automatically.
```
cmarqu, line 174:
```suggestion
        """Discover tests in files automatically.
```

src/cocotb/regression.py
```diff
@@ -176,109 +168,109 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
-        if not self._queue:
-            self.log.warning("No tests were discovered")
+        Should be called before :meth:`start_regression` is called.
+        If no filters are given, all tests are included.
+        If one or more filters are given, only those tests which match a filter are included; the rest are excluded.
 
-        self._queue.sort(key=lambda test: (test.stage, test._id))
+        Args:
+            modules: a list of module names to execute to find tests.
+
```
cmarqu, line 182:
```suggestion
```

src/cocotb/regression.py
```diff
@@ -176,109 +168,109 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
+        name: Optional[str] = None,
+        module: Optional[str] = None,
+        timeout_time: Optional[float] = None,
+        timeout_unit: str = "step",
+        expect_fail: bool = False,
+        expect_error: Union[Type[Exception], Sequence[Type[Exception]]] = (),
+        skip: bool = False,
+        stage: int = 0,
+    ) -> None:
+        """Registers a test with the RegressionManager.
```
cmarqu, line 220:
```suggestion
        """Register a test with the RegressionManager.
```

src/cocotb/regression.py
```diff
@@ -176,109 +168,109 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
-        if module_str is None:
-            raise ValueError(
-                "Environment variable MODULE, which defines the module(s) to execute, is not defined."
-            )
+            timeout_time:
+                Simulation time duration before timeout occurs.
 
-        modules = [s.strip() for s in module_str.split(",") if s.strip()]
+            timeout_unit:
+                Units of timeout_time, accepts any units that :class:`~cocotb.triggers.Timer` does.
```
cmarqu, line 240:
```suggestion
                Units of ``timeout_time``, accepts any units that :class:`~cocotb.triggers.Timer` does.
```

src/cocotb/regression.py
```diff
@@ -176,109 +168,109 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
 
-        cls._setup_pytest_assertion_rewriting(modules)
+            expect_fail:
+                If ``True`` and the test fails a functional check via an ``assert`` statement, :class:`pytest.raises`,
+                :class:`pytest.warns`, or :class:`pytest.deprecated_call` the test is considered to have passed.
```
cmarqu, line 244:
```suggestion
                :class:`pytest.warns`, or :class:`pytest.deprecated_call`, the test is considered to have passed.
```
?

src/cocotb/regression.py
```diff
@@ -176,109 +168,109 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
-                    # If we request a test manually, it should be run even if skip=True is set.
-                    test.skip = False
-
-                    yield test
-
-                # Use the non-matching test names in the next module search
-                tests = not_found_tests
+            skip:
+                Don't execute this test as part of the regression.
+                Test can still be run manually by setting :make:var:`TESTCASE`.
```
cmarqu, line 253:
```suggestion
                The test can still be run manually by setting :make:var:`TESTCASE`.
```

src/cocotb/regression.py
```diff
@@ -302,6 +294,18 @@ def _setup_pytest_assertion_rewriting(cls, test_modules: Iterable[str]) -> None:
                 "Please file a bug report!".format(pytest.__version__)
             )
 
+    def start_regression(self) -> None:
+        """Starts the regression.
```
cmarqu, line 298:
```suggestion
        """Start the regression.
```

ktbarrett:
We need to decide if no tests being discovered is an error or not.

marlonjames:
How would we mark it as a failure? `results.xml` for `test_module_var_empty` is has no testcases:
```xml
<testsuites name="results">
  <testsuite name="all" package="all">
    <property name="random_seed" value="1702489885" />
  </testsuite>
</testsuites>
```

The log message about no tests found is WARNING level.

ktbarrett:
@marlonjames I think based on what we discussed...

* Empty MODULE is an error causing an exception since that is a required argument.
* MODULE not finding tests is a warning as it's probably not intended, but the results.xml will be empty and that should become an error
* TESTCASE filtering out tests is explicit, so it should be a warning. TESTCASE is defeating the standard regression, so it leading to no tests running shouldn't be a big deal since it shouldn't happen in automated runs. I think the best way to support this is to start marking filtered out tests as "skipped" in the results.xml to prove we found the test, we just didn't run it.

src/cocotb/regression.py
```diff
@@ -176,109 +171,127 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
-        self._queue = []
-        for test in tests:
-            self.log.info(f"Found test {test.__module__}.{test.__qualname__}")
-            self._queue.append(test)
+    def discover_tests(
+        self, modules: Iterable[str], filters: Optional[Sequence[str]] = None
+    ) -> None:
+        """Discover tests in files automatically.
+
+        Should be called beforTruee :meth:`start_regression` is called.
```
cmarqu, line 179:
```suggestion
        Should be called before :meth:`start_regression` is called.
```

marlonjames:
I've seen code using cocotb that examines test files to find `Test` objects. With this change, as far as I can tell, there's no real way to find a function decorated with `@cocotb.test` if you aren't running in a simulator environment.

Do we want to provide some mechanism, like a special attribute added to the function/object returned by the decorator?

ktbarrett:
> I've seen code using cocotb that examines test files to find `Test` objects

Interesting. What was the use case?

> Do we want to provide some mechanism, like a special attribute added to the function/object returned by the decorator?

I can have the decorator try to set a `__cocotb_test__` attribute on the object?

marlonjames:
I can't remember the project I saw it in, but I think it was for auto test discovery to build `PYTHONPATH` and `MODULE` to run all of the cocotb tests in the repo without maintaining an explicit list.

marlonjames:
They can always create their own decorator that calls `cocotb.test()`, but I don't think it hurts to provide an attribute.

ktbarrett:
The only issue with the attribute is that things like parameterized tests will only mark the decorated object *once*, so it isn't sufficient for discovering all the tests actually generated. I think we can provide a better mechanism for this considering we have to explicitly turn off registration if `cocotb.regression_manager` is `None`. But lets make that a positive documented feature rather than an accidental one.

I can add the special attribute for now.

src/cocotb/regression.py
```diff
@@ -176,109 +171,127 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
+    ) -> None:
+        """Discover tests in files automatically.
+
+        Should be called before :meth:`start_regression` is called.
+        If no filters are given, all tests are included.
+        If one or more filters are given, only those tests which match a filter are included; the rest are excluded.
+
+        Args:
+            modules: a list of module names to execute to find tests.
+            filters: regex patterns for test names
```
marlonjames, line 185:
```suggestion
            filters: regex patterns for test names.
```

src/cocotb/regression.py
```diff
@@ -176,109 +171,127 @@ def __init__(self, dut: SimHandle, tests: Iterable[Test]):
...
+            self.log.debug("Searching for tests in module %s", module_name)
+            self.log.debug("Python Path: %s", ",".join(sys.path))
+            self.log.debug("PWD: %s", os.getcwd())
+            __import__(module_name)
+
         self.ntests = len(self._queue)
 
-        if not self._queue:
-            self.log.warning("No tests were discovered")
+        # warn if no tests were discovered
```
marlonjames, line 197:
```suggestion
        # error if no tests were discovered
```

marlonjames (approved):
Only small things left.

AndrewNolte:
This seems to want to promote code reuse by having tests return the same type, and not something different, however the old implementation was a Callable as well, so it worked just fine. Now on top of this though, anything that imports other test files will also register those tests, which is a major change for larger, more interconnected projects. It was possible before to explicitly add tests from other files to the namespace to get this behavior.

Also re: "They can always create their own decorator", generate_tests now has the test_dec argument removed, so this isn't possible for these unless a custom test factory is also used. Also, even if a custom decorator is used, how would that allow for the older behavior without the user touching regression.py, or adding a script at the end of each file to register tests?

AndrewNolte:
I think we should go back to the old way in regard to searching for "Test" objects so tests from imported files aren't accidentally added. We can also maybe add the cocotb.regression_manager.add_tests(other_module) to allow for the explicit adding of tests from other modules.

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -349,7 +349,6 @@ async def access_constant_boolean(dut):
     skip=cocotb.LANGUAGE in ["verilog"],
     expect_fail=cocotb.SIM_NAME.lower().startswith("ghdl"),
 )
-@cocotb.test()
```
AndrewNolte, line 352: This seems like an accidental deletion?

tests/test_cases/test_select_testcase/y_tests.py
```diff
@@ -3,4 +3,4 @@
 
 @cocotb.test()
 async def y_test(dut):
```
AndrewNolte, line 5: If there's a second test with the same prefix (y_test_dos), and TESTCASE=y_test, the filter will match both. I don't think this is the expected behavior.

src/cocotb/regression.py
```diff
@@ -903,33 +997,27 @@ def generate_tests(
...
+
+            @functools.wraps(self.test_function)
+            async def _my_test(dut, kwargs: Dict[str, Any] = kwargs):
+                await self.test_function(dut, *self.args, **kwargs)
+
+            _my_test.__doc__ = doc
+            _my_test.__name__ = name
+            _my_test.__qualname__ = name
+
+            cocotb.regression_manager.register_test(
```
AndrewNolte, line 1012: I think we can solve the over-zealous test importer by filtering the tests on the __module__ attr. The only problem here is if generate_tests() is called on a function that isn't in the current module. We would need to solve this by passing in the module here from the calling location of generate_tests

tests/test_cases/test_cocotb/test_testfactory.py
```diff
@@ -18,14 +19,16 @@
...
     testfactory_test_args.add((arg1, arg2, arg3))
 
 
 factory = TestFactory(run_testfactory_test)
 factory.add_option("arg1", ["a1v1", "a1v2"])
 factory.add_option(("arg2", "arg3"), [("a2v1", "a3v1"), ("a2v2", "a3v2")])
-factory.generate_tests(prefix=prefix, postfix=postfix)
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore")
+    factory.generate_tests(prefix=prefix, postfix=postfix)
```
AndrewNolte, line 31: No test for the new name= attribute?

tests/test_cases/test_discovery/test_discovery.py
ktbarrett: It wasn't. There are two decorators.

tests/test_cases/test_select_testcase/y_tests.py
ktbarrett: This is the new expected behavior. It more closely resembles pytest's `-k` option. If you only want to match "exactly" use `^name$`. It's regular expressions.

src/cocotb/regression.py
ktbarrett: `TestFactory.generate_tests` looking up the stack to find the module to insert tests into is one of the most poorly thought out things I've seen. I'm glad it's dead. `TestFactory`'s usage should be supplanted by `cocotb.parametrize` as it leaves less to interpretation.

ktbarrett:
The new implementation brings new functionality
* dynamic test creation without having to insert things into a module and have a unique name for each
* returning the originally decorated object
   * useful for decorating a class as a test, then inheriting from it to define another test. See pyuvm.
* more expressive test filtering using `TESTCASE`

This is a 2.0 release. Backwards compatibility is not a priority if new functionality and cleanup can be justify it. This has been a LONG time coming. This project before now had some major design and API issues. It was a prototype pressed into service as production-quality SW. 2.0 is an attempt to make it into actual production-quality SW.

AndrewNolte:
> * useful for decorating a class as a test, then inheriting from it to define another test. See pyuvm.

Inheriting from a function? Should we change tests to be classes like in pyuvm?

And I get why the TESTCASE change, name=, unique names are all useful features, I don't really see how that justifies combining them all into one PR. This is an area that isn't really tested well on the cocotb repo because all of the tests are small and independent. I think it'd be nice if there was some CD linked with a larger project that uses cocotb.

I agree the generate_tests() inspecting the stack is pretty messy. I think the module filtering in regression.py is probably the way to go here. The problem is if you import the function from another module here, you'll have to overwrite that function's module to the current one when calling generate_tests. So I think generate_tests should actually return early if it's not being called from a module in MODULE. But then again it'd be inspecting the stack...

AndrewNolte:
Has anyone looked at having cocotb use pytest to mark/discover tests?

ktbarrett:
What pyuvm and my home-grown test framework at my previous employer intended to do, but is impossible in cocotb 1.X because it constructs a `test` object.

```python

@cocotb.test()
class MyTest(Test):
    def __init__(self, dut):
        # create testbenching objects
    async def run(self):
        # do stuff

@cocotb.test()
class MyOtherTest(MyTest):  # inherit
    async def run(self):
        # do other stuff
```

The existing logic for `TESTCASE` looked through the `MODULES` for *exactly* the objects in the module matching `TESTCASE` and stopping once it found an object for all the tests listed (so not all `MODULES` were actually loaded, lol). With the refactor, there was not a clear way to accomplish this anymore, so it was changed then.

I think it would be a great idea to integrate pytest, but I haven't seen a way to do it. But then it also runs afoul of cocotb-test and the Python runner which have been presented now or in the past as being able to used in the same file as a cocotb test. Pytest outside and inside wouldn't be able to distinguish. Pytest can't be arbitrarily modified to only pick out `cocotb.test` decorated objects or similar. So then best we can do is mimic the functionality.

I agree that the refactor did make importing objects from other test modules not listed in MODULE no longer possible. We really do need to have the regression manager simply pick up objects in modules, but we need to not disturb the original object.

I think the best *new* course of action would instead of having `cocotb.test` tag objects with `__cocotb_test__` as a little convenience, instead put the test info in that and have the regression manager use that info to register tests. `TestFactory` could be changed so `generate_tests` yields an object with `__cocotb_test__` attr that when collected registers multiple tests. This would allow one to call `generate_tests` in modules other than the one where the `TestFactory` was defined.

ktbarrett:
That might be more than we're willing to commit rn, but if we make the implementation generic enough, that could be implemented pretty easily.

PR #3608, opened 2023-12-27
[Vpi] Allow using packed struct as a signal on sim module

### LINKED ISSUE #3623: [Verilator] VPI: Allow reading/writing entire structs
This is allowable through vpi, but not with the current simulator implementation. The simulator change is in #3608.

A handle.py change is needed in order to interact with structs while supporting existing subhandle access. We could have a new Structobject that supports read/write through "_value", or have it as a preference to parse structs as logic objects

AndrewNolte:
The purpose of this is to be able to set/get a struct's entire value, rather than just its individual signals, which in turn will allow for more shared user code between verilator and the other environments.

I default packed structs to the new behavior since this is more correct, and allow casting from to the old hierarchy object via asHierarchyObject().

Closes #3623.


AndrewNolte:
I realized when running this that doing
```
make -C tests/test_cases/test_struct SIM=xcelium
make -C tests/test_cases/test_struct SIM=verilator
```
in sequence causes the verilator one to not run and exit early, any idea why this might be the case? I have to make clean before I can run verilator

AndrewNolte:
Can I get a look at this @marlonjames @ktbarrett. I'm not sure why it's tripping up on vhdl runs

ktbarrett:
I'll try to do this tonight. Otherwise it might be a while.

AndrewNolte:
It looks like vhdl sims actually seg fault when trying to call get_signal_val_binstr, which is why they failed even though expect_fail was set to true. And the verilog ones rn are because of the logicarray change, I'll fix those rq

ktbarrett:
Just changing structs to arrays is not the right answer here. I'm actually really surprised that the regression passes with this change. Clearly not enough testing is being done. What's needed is a new type that supports both accessing fields and setting the whole value.

I don't know what happens when you attempt to set the full value on something that isn't a packed struct, or if most simulators even support differentiating packed vs unpacked structs.

AndrewNolte:
It's essentially just adding the signal methods to the gpi handle, whereas before it didn't have these. Changing handle.py to allow both hierarchy and signal will be the tricky part

marlonjames:
`sample_module.inout_if` happens to be a packed struct here, but unpacked structures and unpacked unions cannot be set in their entirety with `vpi_put_value()`.

This type of feature would at minimum require getting `vpi_get(vpiPacked)` to differentiate, and I agree with Kaleb on developing a new type for interfacing to it.

ktbarrett:
Actually now that I've been thinking about this. I think this might be a good idea. Indexing into packed objects, whether that's packed arrays, or packed structs, etc. is not kosher, at least conceptually. I'm now more surprised that what we currently have works.

We do need a new type for packed objects that aren't single-dimension packed arrays.

AndrewNolte:
"What's needed is a new type that supports both accessing fields and setting the whole value." - so only VpiSignal provide special extra methods, VpiObjHdl does not. Everything for accessing fields is just operating on the raw vpi handle, so our check for that is simply 'does the vpi throw an error' and relying on handle.py to call it on valid objects.

So I can make a VpiStructObjHdl that just extends VpiSignalObjHdl, but it'll add nothing. And I'll add the vpiPacked check as well. It would be nice to also have it extend VpiObjHdl to get defname and deffile, but then we'd have multiple inheritance

ktbarrett:
I'm still trying to figure out a holistic approach to this.

In my mind:
1. You can't index into packed objects.
2. Packed objects are discovered as single GPI objects.
3. This gets mapped to a `LogicObject` or similar, which cannot be indexed.
4. You should be able to query the structure of that object (`[0:2][0:2]packed_struct_t[7:0]`, etc.).
5. The Python type we return from `.value` of such objects should use the structure definition to support indexing and updated the value, but getting and setting the value on the handle only happens whole packed object at a time (read-modify-write).

I know this is going to require both C++ and Python changes, but I'm uncertain to what extent. Also I'm uncertain if point 3 is even possible.

> What's needed is a new type that supports both accessing fields and setting the whole value.

Yeah... maybe we should forget about accessing fields on the object and only worry about accessing fields on the value coming out.

AndrewNolte:
Do many simulators support the type spec? For xcelium (and seemingly other sims) it just treats them as unpacked. I think just having this for now supports the old way via an explicit type change. I was originally going to support both .value and child handle access, but with the latest rebase there were some method conflicts between LogicObject and HierarchyObject, so I thought it best to default to the proper way for Packed Structs. (__len__) especially would be confusing

And again, what cpp changes need to be made if we're not inspecting the type spec? The object hierarchy was implemented as methods on GpiObjHdl and VpiImpl, not VpiObjHdl, so it works just fine.

ktbarrett:
So uhhh... yeah I definitely want to go back to the original implementation on this. Let's just remove support for now and try to realize #3857 sooner rather than later.

ktbarrett (changes requested):
So uhhh... yeah I definitely want to go back to the original implementation on this. Let's just remove support for now and try to realize #3857 sooner rather than later.

tests/test_cases/test_struct/test_struct.py
```diff
@@ -0,0 +1,51 @@
...
+from cocotb.triggers import Timer
+
+test_dec = cocotb.test(
+    expect_error=AttributeError
+    if cocotb.SIM_NAME.lower().startswith(("icarus", "ghdl", "nvc"))
+    else (),
+    expect_fail=cocotb.SIM_NAME.lower().startswith(("modelsim", "riviera")),
+    skip=cocotb.SIM_NAME.lower().startswith(
+        "verilator"
+    ),  # verilator already treats packed structs as logic arrays
```
ktbarrett, line 17: Why would we skip this then? Shouldn't it pass?

tests/test_cases/test_iteration_mixedlang/test_iteration.py
```diff
@@ -58,7 +58,7 @@ async def recursive_discovery(dut):
         pass_total = 975
     elif cocotb.SIM_NAME.lower().startswith("xmsim"):
         # Xcelium sometimes doesn't find bits in a std_logic_vector
-        pass_total = 1257
+        pass_total = 1201
     elif cocotb.SIM_NAME.lower().startswith("modelsim"):
         pass_total = 1332
```
ktbarrett, line 63:
```suggestion
        pass_total = 1276
```
A difference of 56. Does that sound right?

ktbarrett (changes requested):

ktbarrett (approved):

tests/test_cases/test_struct/test_struct.py
```diff
@@ -1,15 +1,58 @@
...
+    expect_error=AttributeError
+    if SIM_NAME.startswith(("icarus", "ghdl", "nvc"))
+    else (),
+    expect_fail=SIM_NAME.startswith("modelsim"),
+)
+async def test_packed_struct_format(dut):
+    """Test that the correct objects are returned for a struct"""
+    assert repr(dut.my_struct) == "LogicObject(sample_module.my_struct)"
+
+    # use value or value to access signal
```
cmarqu, line 30:
```suggestion
    # use value to access signal
```
?

tests/test_cases/test_struct/test_struct.py
```diff
@@ -1,15 +1,58 @@
...
+    cocotb.log.info("dut.my_struct.value=%s", dut.my_struct.value)
```
ktbarrett, line 31:
```suggestion
```

ktbarrett:
@cocotb/maintainers Considering I plan on reintroducing the indexing in a future commit, is this good?

PR #3612, opened 2023-12-30
Fix artifact upload

imphil:
https://github.com/cocotb/cocotb/pull/3593 updated the version of
actions/upload-artifact from 3 to 4, but didn't perform the migration as
described at
https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md.
Do that to fix the broken release CI.

While at it, update all jobs where the Python version doesn't matter from Python 3.11 to Python 3.12

- CI: Run build tasks with latest Python
- CI: Fix artifact upload

### DIFF
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.12"
       - name: Install nox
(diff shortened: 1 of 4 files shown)


imphil:
The release run is now successful: https://github.com/cocotb/cocotb/actions/runs/7370839706?pr=3612

ktbarrett (approved):
What was the reason for splitting the dists by os?

imphil:
The builds are native, hence we need to build on three machines and then combine them.

PR #3613, opened 2024-01-01
Xcelium: support VHDL toplevels

imphil:
A set of commits to add support for VHDL toplevels with Xcelium 23.09.004 and newer. 

The underlying simulator issues have been resolved by Cadence and have been included in Xcelium 23.09.004 (as well as other hotfix releases of earlier branches, which I didn't test any further). Have a look at the individual commits for a detailed view into what had to change in our test suite.


cmarqu:
Great!
(Needs a "Closes #1076".)

cmarqu (approved):

ktbarrett (approved):

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -42,8 +43,20 @@ async def test_function_reentrant_clock(dut):
         await timer
 
 
-@cocotb.test()
+# Xcelium/VHDL does not correctly report the simulator precision.
+# See also https://github.com/cocotb/cocotb/issues/3419
+# NVC and GHDL do not support setting precision and always use 1 fs.
```
imphil, line 48: @nickg Is it an intentional design choice that there seems to be no way to set the time precision, or is it worth filing a feature request for it?

@tgingold The same question for you on GHDL, only that I found an issue a while back that stated that GHDL internally always uses 1 fs and is designed around that assumption. Still checking with you to be sure.

tgingold: No, GHDL can use precision other than 1 fs.

imphil: @tgingold Good to know! How does one configure that from the command line?

tgingold: There is the `--time-resolution` option, but in general it's automatic (use the least precision)

imphil: @tgingold Excellent, thanks! I'll follow up with a PR to use this switch. (We have a test where we need to force a specific resolution, hence the need for a switch.)

nickg: There's an open issue at nickg/nvc#607 but I've not got round to implementing it yet.

imphil: @tgingold FYI: https://github.com/cocotb/cocotb/pull/3622 has the update for GHDL.

imphil: @nickg Thanks! I added a link to the open issue in https://github.com/cocotb/cocotb/pull/3622.

PR #3614, opened 2024-01-01
Add a missing newsfragment for #3490

imphil:


imphil:
Looks like I didn't `git add` this file, but it's still valuable information.

cmarqu (approved):

Issue #3617, opened 2024-01-02
Validate pyproject.toml in pre-commit


imphil:
The `validate-pyproject` tool can validate the contents of the `pyproject.toml` file against a schema, which helps to catch wrong configuration options.

At this point, however, we have two non-ideal options:
* The schema built into validate-pyproject provides fast but incomplete validation (many sections are not validated)
* Using an external schema with `validate-pyproject --store=https://json.schemastore.org/pyproject.json` slows down a pre-commit run by around 8 seconds (on my machine), which is very noticeable.

I filed https://github.com/abravalheri/validate-pyproject/issues/138 to see if something can be done about the speed problem, and I'm stalling further work on this until we have a faster solution.

imphil:
Draft PR at https://github.com/cocotb/cocotb/pull/3618 (not to be merged until we have a faster solution in place).

PR #3618, opened 2024-01-02
Validate the pyproject.toml file in pre-commit

imphil:
The Python tool "validate-pyproject" validates the pyproject.toml
against a schema. The schemas are shipped in
validate-pyproject-schema-store. We want to pin the version of the
schema used to validate our pyproject.toml. We do that by using
validate-pyproject-schema-store directly (which depends on
validate-pyproject).

As a downside, the schema validation adds around 1.5 seconds to a
pre-commit run (or in relative numbers, a run now takes roughly 2.5x as
long on my machine).

Fixes https://github.com/cocotb/cocotb/issues/3617

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,3 +32,8 @@ repos:
     - "--fix=lf"
   - id: "end-of-file-fixer"
     exclude: "^docs/source/diagrams/(svg|xml)/"
+
+- repo: https://github.com/henryiii/validate-pyproject-schema-store
+  rev: "2024.01.31"
+  hooks:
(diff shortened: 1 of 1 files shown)


ktbarrett (approved):
Is it running faster now? The issue in the other repo still seems open.

imphil:
> Is it running faster now? The issue in the other repo still seems open.

It's running much faster now, taking around 1.5 seconds to validate pyproject.toml file. Which isn't super fast, but it's not as slow as it was either and doesn't bother me too much any more.

ktbarrett:
Merge away then.

PR #3622, opened 2024-01-02
GHDL: Support COCOTB_HDL_TIMEPRECISION if possible

imphil:
GHDL with the mcode backend supports some values of
COCOTB_HDL_TIMEPRECISION (fs, ps, ns, us, ms, s), and only if the
mcode backend is used. Try to pass on the value of
COCOTB_HDL_TIMEPRECISION if possible, without checking for the backend:
GHDL will error out in that case and users have to avoid setting
COCOTB_HDL_TIMEPRECISION when that happens.

See also https://github.com/cocotb/cocotb/pull/3613#discussion_r1439710866

### DIFF
--- a/src/cocotb/runner.py
+++ b/src/cocotb/runner.py
@@ -6,7 +6,7 @@
 
 # TODO: maybe do globbing and expanduser/expandvars in --include, --vhdl-sources, --verilog-sources
 # TODO: create a short README and a .gitignore (content: "*") in both build_dir and test_dir? (Some other tools do this.)
-# TODO: support timescale
+# TODO: support timescale on all simulators
 # TODO: support custom dependencies
 
 import abc
@@ -170,7 +170,7 @@ class Simulator(abc.ABC):
             build_dir: Directory to run the build step in.
             clean: Delete build_dir before building
             verbose: Enable verbose messages.
-            timescale: Tuple containing time unit and time precision for simulation
+            timescale: Tuple containing time unit and time precision for simulation.
             waves: Record signal traces.
         """
 
@@ -223,6 +223,7 @@ class Simulator(abc.ABC):
         test_dir: Optional[PathLike] = None,
         results_xml: str = "results.xml",
         verbose: bool = False,
+        timescale: Optional[Timescale] = None,
     ) -> Path:
         """Run the tests.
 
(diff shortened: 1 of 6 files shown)


src/cocotb/share/makefiles/simulators/Makefile.ghdl
```diff
@@ -63,6 +63,31 @@ ifeq ($(OS),Msys)
     export PATH := $(GHDL_BIN_DIR)/../lib:$(PATH)
 endif
 
+GHDL_RUN_ARGS ?=
+ifneq ($(COCOTB_HDL_TIMEPRECISION),)
+    # Convert the time prescision to a format string supported by GHDL, if
```
marlonjames, line 68:
```suggestion
    # Convert the time precision to a format string supported by GHDL, if
```

src/cocotb/runner.py
```diff
@@ -248,6 +249,7 @@ def test(
             results_xml: Name of xUnit XML file to store test results in.
                 When running with pytest, the testcase name is prefixed to this name.
             verbose: Enable verbose messages.
+            timescale: Tuple containing time unit and time precision for simulation
```
cmarqu, line 252:
```suggestion
            timescale: Tuple containing time unit and time precision for simulation.
```

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -45,11 +45,11 @@ async def test_function_reentrant_clock(dut):
 
 # Xcelium/VHDL does not correctly report the simulator precision.
 # See also https://github.com/cocotb/cocotb/issues/3419
-# NVC and GHDL do not support setting precision and always use 1 fs.
+# NVC does not support setting precision and always use 1 fs.
```
cmarqu, line 48:
```suggestion
# NVC does not support setting precision and always uses 1 fs.
```

cmarqu (approved):

tmeissner:
It seems that this PR breaks using GHDL with other backend than mcode as the following check

```
ifneq ($(COCOTB_HDL_TIMEPRECISION),)
...
```

is always true because `COCOTB_HDL_TIMEPRECISION` is set in `Makefile.inc`:

```
COCOTB_HDL_TIMEPRECISION ?= 1ps
```

regardless if they are set in the users Makefile.

imphil:
@tmeissner I filed https://github.com/cocotb/cocotb/issues/3629 to track this further, comments in a closed PR are hard to find.

PR #3624, opened 2024-01-03
Simplify VpiArrayObjHdl::initialise

AndrewNolte:
I had issues with range_idx being calculated incorrectly, so I wanted to see what simulators actually require this, if at all. It gets calculated incorrectly when using get_handle_by_name() with a fully qualified name (not from handle.py) with brackets in other parts of the path. I didn't see range_idx being actually used in testing on Xcelium.  It doesn't look like there were any tests when it was put in https://github.com/cocotb/cocotb/commit/54d2965dc29f226f3971ca6e5b6b3cc67149fbc1

### DIFF
--- a/src/cocotb/share/lib/vpi/VpiObj.cpp
+++ b/src/cocotb/share/lib/vpi/VpiObj.cpp
@@ -29,6 +29,7 @@
 
 #include <assert.h>
 
+#include <algorithm>
 #include <stdexcept>
 
 #include "VpiImpl.h"
@@ -47,18 +48,15 @@ int VpiArrayObjHdl::initialise(const std::string &name,
 
     /* Removing the hdl_name from the name will leave the pseudo-indices */
     if (hdl_name.length() < name.length()) {
-        std::string idx_str = name.substr(hdl_name.length());
-
-        while (idx_str.length() > 0) {
-            std::size_t found = idx_str.find_first_of("]");
-
-            if (found != std::string::npos) {
-                ++range_idx;
-                idx_str = idx_str.substr(found + 1);
-            } else {
-                break;
-            }
+        // get the last index of hdl_name in name
+        std::size_t idx_str = name.rfind(hdl_name);
+        if (idx_str == std::string::npos) {
+            LOG_ERROR("Unable to find name %s in %s", hdl_name.c_str(),
+                      name.c_str());
+            return -1;
         }
+        // count occurences of [
+        range_idx = std::count(name.begin() + idx_str, name.end(), '[');
     }
 
     /* After determining the range_idx, get the range and set the limits */
@@ -68,14 +66,12 @@ int VpiArrayObjHdl::initialise(const std::string &name,
     if (iter != NULL) {
         rangeHdl = vpi_scan(iter);
 
-#ifdef MODELSIM
         for (int i = 0; i < range_idx; ++i) {
             rangeHdl = vpi_scan(iter);
             if (rangeHdl == NULL) {
                 break;
             }
         }
-#endif
         if (rangeHdl == NULL) {
             LOG_ERROR("Unable to get range for indexable array");
             return -1;
--- a/src/cocotb/share/lib/vpi/VpiObj.cpp
+++ b/src/cocotb/share/lib/vpi/VpiObj.cpp
@@ -76,11 +76,11 @@ int VpiArrayObjHdl::initialise(const std::string &name,
             }
         }
 #endif
-        vpi_free_object(iter);  // Need to free iterator since exited early
         if (rangeHdl == NULL) {
             LOG_ERROR("Unable to get range for indexable array");
             return -1;
         }
+        vpi_free_object(iter);  // Need to free iterator since exited early
     } else if (range_idx == 0) {
         rangeHdl = hdl;
     } else {
--- a/src/cocotb/share/lib/vpi/VpiObj.cpp
+++ b/src/cocotb/share/lib/vpi/VpiObj.cpp
@@ -56,7 +56,7 @@ int VpiArrayObjHdl::initialise(const std::string &name,
             return -1;
         }
         // count occurences of [
-        range_idx = (int)std::count(name.begin() + idx_str, name.end(), '[');
+        range_idx = (int)std::count(&name.begin()[idx_str], name.end(), '[');
     }
 
     /* After determining the range_idx, get the range and set the limits */
--- a/src/cocotb/share/lib/vpi/VpiObj.cpp
+++ b/src/cocotb/share/lib/vpi/VpiObj.cpp
@@ -63,47 +63,40 @@ int VpiArrayObjHdl::initialise(const std::string &name,
 
     /* After determining the range_idx, get the range and set the limits */
     vpiHandle iter = vpi_iterate(vpiRange, hdl);
+    vpiHandle rangeHdl;
+
+    if (iter != NULL) {
+        rangeHdl = vpi_scan(iter);
+
+#ifdef MODELSIM
+        for (int i = 0; i < range_idx; ++i) {
+            rangeHdl = vpi_scan(iter);
+            if (rangeHdl == NULL) {
+                break;
+            }
+        }
+#endif
+        vpi_free_object(iter);  // Need to free iterator since exited early
+        if (rangeHdl == NULL) {
+            LOG_ERROR("Unable to get range for indexable array");
+            return -1;
+        }
+    } else if (range_idx == 0) {
+        rangeHdl = hdl;
+    } else {
+        LOG_ERROR("Unable to get range for indexable array or memory");
+        return -1;
+    }
 
     s_vpi_value val;
     val.format = vpiIntVal;
+    vpi_get_value(vpi_handle(vpiLeftRange, rangeHdl), &val);
+    check_vpi_error();
+    m_range_left = val.value.integer;
 
-    if (iter != NULL) {
-        vpiHandle rangeHdl;
-        int idx = 0;
-
-        while ((rangeHdl = vpi_scan(iter)) != NULL) {
-            if (idx == range_idx) {
-                break;
-            }
-            ++idx;
-        }
-
-        if (rangeHdl == NULL) {
-            LOG_ERROR("Unable to get range for indexable object");
-            return -1;
-        } else {
-            vpi_free_object(iter);  // Need to free iterator since exited early
-
-            vpi_get_value(vpi_handle(vpiLeftRange, rangeHdl), &val);
-            check_vpi_error();
-            m_range_left = val.value.integer;
-
-            vpi_get_value(vpi_handle(vpiRightRange, rangeHdl), &val);
-            check_vpi_error();
-            m_range_right = val.value.integer;
-        }
-    } else if (range_idx == 0) {
-        vpi_get_value(vpi_handle(vpiLeftRange, hdl), &val);
-        check_vpi_error();
-        m_range_left = val.value.integer;
-
-        vpi_get_value(vpi_handle(vpiRightRange, hdl), &val);
-        check_vpi_error();
-        m_range_right = val.value.integer;
-    } else {
-        LOG_ERROR("Unable to get range for indexable object");
-        return -1;
-    }
+    vpi_get_value(vpi_handle(vpiRightRange, rangeHdl), &val);
+    check_vpi_error();
+    m_range_right = val.value.integer;
 
     /* vpiSize will return a size that is incorrect for multi-dimensional arrays
      * so use the range to calculate the m_num_elems.
--- a/src/cocotb/share/lib/vpi/VpiObj.cpp
+++ b/src/cocotb/share/lib/vpi/VpiObj.cpp
@@ -56,7 +56,9 @@ int VpiArrayObjHdl::initialise(const std::string &name,
             return -1;
         }
         // count occurences of [
-        range_idx = (int)std::count(&name.begin()[idx_str], name.end(), '[');
+        auto start =
+            name.begin() + static_cast<std::string::difference_type>(idx_str);
+        range_idx = std::count(start, name.end(), '[');
     }
 
     /* After determining the range_idx, get the range and set the limits */
--- a/src/cocotb/share/lib/vpi/VpiObj.cpp
+++ b/src/cocotb/share/lib/vpi/VpiObj.cpp
@@ -56,7 +56,7 @@ int VpiArrayObjHdl::initialise(const std::string &name,
             return -1;
         }
         // count occurences of [
-        range_idx = std::count(name.begin() + idx_str, name.end(), '[');
+        range_idx = (int)std::count(name.begin() + idx_str, name.end(), '[');
     }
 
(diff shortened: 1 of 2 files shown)


src/cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -125,10 +125,14 @@ int VpiArrayObjHdl::initialise(const std::string &name,
 
     m_indexable = true;
 
-    int range_idx = 0;
+    vpiHandle iter = vpi_iterate(vpiRange, hdl);
+
+    s_vpi_value val;
+    val.format = vpiIntVal;
+    vpiHandle rangeHdl;
```
ktbarrett, line 132: Not sure why these were moved up, this is less readable. Variable declarations should be as close to their use as possible.

ktbarrett (commented):
Does this fix your issue simply because only Questa/ModelSim, which you aren't using, need pseudo-regions, which you've segregated into ifdef's? So the problem of supporting `get_handle_by_name` from the root on Questa, if there are multiple indexes, is still present, correct? But this commit removes this logic, which isn't necessary for the other simulators?

What tests exist in the regression which test this?

@marlonjames Do you have any concerns here?

AndrewNolte:
> s_vpi_value

Yeah we have use cases where we query objects from the root handle, but the logic for searching through the brackets causes issues. Are we okay with adding tests specifically for the simulator module? I think a good test case would be iterating through handles in a large design and asserting that subhdl == roothdl.get_handle_by_name(subhdl._path), and can be skipped on questa for now.

ktbarrett:
> Are we okay with adding tests specifically for the simulator module?

Oh absolutely! I encourage it.

> can be skipped on questa for now.

That's fine. We can maybe open an upstream ticket or at least track it in an issue if you could open that with the particulars that I'm not yet groking.

AndrewNolte:
I added it to the iteration test case, but ideally it would test on all of the designs, with multiple arrays in the path. Also, while looking at the iteration tests, I think I found that these test tags were removed on accident? https://github.com/cocotb/cocotb/commit/db1cce76e63a52bb7b4c06db39a767199c54092c#diff-1082c9445dd85ec000286580d8e757077f81782e15cd253a1279b956a3cad2cb

marlonjames:
I will try to take a look at this soon.

ktbarrett:
Unfortunately I don't think there's anything we can do until the private CI job hang is resolved. So sorry if we don't look at this sooner.

imphil:
> Unfortunately I don't think there's anything we can do until the private CI job hang is resolved. So sorry if we don't look at this sooner.

That's resolved now, sorry for the disruption.

tests/test_cases/test_iteration_verilog/test_iteration_es.py
```diff
@@ -52,25 +53,34 @@ async def recursive_discovery(dut):
...
+            yield thing
+            yield from sim_iter(thing)
+
+    total = 0
+    for thing in sim_iter(dut):
+        tlog.info("Found %s (%s)", thing._path, type(thing))
+
+        subpath = thing._path.split(".", 1)[1]
+        if not SKIP_HANDLE_ASSERT:
+            assert dut._handle.get_handle_by_name(subpath) == thing._handle
```
ktbarrett, line 80: So just to get this straight, this is testing that the object retrieved by name from the root object is the same object as the one found during iteration-based discovery? And that is not the case for Icarus and Riviera for some reasons... A comment might be worthwhile.

The above C code changes look good.

AndrewNolte: Yeah, but I might've mistaked the CI being out of commission as a test failure?

ktbarrett: Well it was just re-run and seems to be working now.

AndrewNolte: ok I'll revert and push to see if it works on those sims

ktbarrett (approved):

ktbarrett:
@marlonjames Could you have a look?

src/cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -146,51 +144,44 @@ int VpiArrayObjHdl::initialise(const std::string &name,
...
-        check_vpi_error();
-        m_range_left = val.value.integer;
-
-        vpi_get_value(vpi_handle(vpiRightRange, hdl), &val);
-        check_vpi_error();
-        m_range_right = val.value.integer;
+        rangeHdl = hdl;
     } else {
-        LOG_ERROR("Unable to get range for indexable object");
+        LOG_ERROR("Unable to get range for indexable arrray");
```
marlonjames, line 171: ?
```suggestion
        LOG_ERROR("Unable to get range for indexable array or memory");
```

src/cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -122,13 +122,11 @@ int VpiCbHdl::cleanup_callback() {
...
                                const std::string &fq_name) {
     vpiHandle hdl = GpiObjHdl::get_handle<vpiHandle>();
-
     m_indexable = true;
 
     int range_idx = 0;
 
-    /* Need to determine if this is a pseudo-handle to be able to select the
-     * correct range */
+#ifdef MODELSIM
```
marlonjames, line 129: `std::string hdl_name = vpi_get_str(vpiName, hdl);` should be pulled out of the `#ifdef`, and the non-Modelsim case should enforce that `hdl_name == name` so that we get an error if there is another VPI simulator that needs pseudo-regions for multi-dimensional arrays.

AndrewNolte: I'm don't think name != hdl_name always implies this? In Verilator when there's a generate scope, hdl_name = 'my_mod.scope' and name = 'scope', but maybe that's another issue for fix on the Verilator side...

marlonjames: `vpiName` should be the local name of the object, `vpiFullName` should be the full hierarchical name.

marlonjames: Because pseudo-regions use the handle of the parent (there is no proper VPI object with a handle), the names should mismatch.
That is how it works now. Pseudo-regions should be re-worked to preserve more context, right now the checking for and handling of pseudo-regions is scattered across the GPI codebase.

ktbarrett: Maybe this should be a follow on then? One thing at a time with a PR. Maybe open an issue for future investigation.

marlonjames: Then I'd rather special-case the simulator(s) that have a problem with the pseudo-region check, instead of the other way around.
I see in Verilator's code that for scopes, there isn't a difference between `vpiName` and `vpiFullName`, so that is likely the cause of the problem here. I might be okay with special-casing for Verilator until that is fixed, but like #3664 I think the change belongs in Verilator.

AndrewNolte: Yeah I think verilator needs a test case for just iterating through the tree and printing the handle info. Maybe we can add that to cocotb to measure simulator vpi compatibility, since the infrastructure for running different sims is already there? Or maybe that would belong on https://chipsalliance.github.io/sv-tests-results/

AndrewNolte: The original issue for me was that in create_gpi_obj_from_handle, name is assumed to be the vpiName of the object, but it's not the case if calling the get_handle_by_name from a handle that isn't the direct parent handle. Would it be possible to add some flag there to not create pseudo handles, or not create them if we're querying down further than just the children?

AndrewNolte: Hey following up on this, if the pseudo handle check here is only needed for modelsim, wouldn't we rather have it only create handles for that sim? 

Also, I made the vpi dump verilator test case, and it looks like the scopes are generated, so we may just need to add pseudo-handle code specific to verilator https://github.com/verilator/verilator/pull/4838/files

marlonjames: Ok, I think I have a better understanding now.
So your problem came from passing a hierarchy to `SimHandleBase._handle.get_handle_by_name()` rather than just the name of a direct child object.

There is no analysis of the `name` passed into `native_check_create()`, it is just concatenated with the name of the `parent` handle and looked up via `vpi_handle_by_name()`. The possibly-a-hierarchy name is then passed through to `create_gpi_obj_from_handle()`, which also has the assumption that it is the name of the object. Finally, that is passed to `initialise()` as if it is the name.

I think we need to either fix the code making the assumptions, or provide a separate function to call when looking up a handle that is not a direct child (e.g. we recommend using `_id()` with a hierarchical path for some cases where simulators have problems with generates etc).

AndrewNolte: Yeah I did some more testing and this ifdef should actually be included. The one below does seem to be verify specific to modelsim though

marlonjames:
I'll take another look at this soon. Please don't merge until then.

ktbarrett:
Re-running CI.

tests/test_cases/test_iteration_verilog/test_iteration_es.py
```diff
@@ -52,25 +52,37 @@ async def recursive_discovery(dut):
     else:
         pass_total = 265
 
+    # Icarus doesn't support array indexes like get_handle_by_name("some_path[x]")
+    SKIP_HANDLE_ASSERT = cocotb.SIM_NAME.lower().startswith(
+        ("riviera", "icarus")
+    ) or cocotb.LANGUAGE in ["vhdl"]
```
ktbarrett, line 58: `cocotb.LANGUAGE` was removed.

ktbarrett (approved):
LGTM. @marlonjames?

src/cocotb/share/lib/vpi/VpiObj.cpp
```diff
@@ -63,48 +63,41 @@ int VpiArrayObjHdl::initialise(const std::string &name,
...
+        for (int i = 0; i < range_idx; ++i) {
+            rangeHdl = vpi_scan(iter);
+            if (rangeHdl == NULL) {
                 break;
             }
-            ++idx;
         }
-
+#endif
+        vpi_free_object(iter);  // Need to free iterator since exited early
```
marlonjames, line 79: If `rangeHdl` is null, the iterator handle is no longer valid so must not be freed here.

src/cocotb/share/lib/vpi/VpiObj.cpp
```diff
@@ -63,48 +63,41 @@ int VpiArrayObjHdl::initialise(const std::string &name,
...
+    vpiHandle rangeHdl;
 
     if (iter != NULL) {
-        vpiHandle rangeHdl;
-        int idx = 0;
+        rangeHdl = vpi_scan(iter);
 
-        while ((rangeHdl = vpi_scan(iter)) != NULL) {
-            if (idx == range_idx) {
+#ifdef MODELSIM
```
marlonjames, line 71: This code isn't Modelsim specific.
The problem in `recursive_discovery` test has to do with `dut._handle.get_handle_by_name(subpath)` as detailed in #3710.

AndrewNolte: If it's not modelsim specific, why isn't there a testcase in the other sims that require this codepath? Modelsim was the only one that failed when I took this out, and taking this logic out sovled #3710 for me.

src/cocotb/share/lib/vpi/VpiObj.cpp
```diff
@@ -63,48 +63,41 @@ int VpiArrayObjHdl::initialise(const std::string &name,
...
+        for (int i = 0; i < range_idx; ++i) {
+            rangeHdl = vpi_scan(iter);
+            if (rangeHdl == NULL) {
                 break;
```
ktbarrett, line 75: I think this line is a different error. We *can* scan, but it's yielding less than the identified number of dimensions. So instead of "Unable to get range for indexable array", maybe it's "Unable to get range for all dimensions of indexable array"?

src/cocotb/share/lib/vpi/VpiObj.cpp
ktbarrett: I think the issue is that conceptually this code should *always* run. If we get a multidimensional array, such that `range_idx` is non-0, this code should always be run to get the correct range spec.

Perhaps the only reason that Questa hits this line is because Questa is the only sim that *tests* this line. Might this have to do with how pseudo-regions are handled?

ktbarrett (changes requested):
Now that I better understand the context, I see a better solution. 

We need to get the number of dimensions if the object is a pseudo-region. What we need to do is find only the trailing indexes *that aren't a part of the handle*. The problem is when objects are attempts to be created not from parents, but from higher in the hierarchy. So what needs to be done is:

```
std::string hdl_name = vpi_get(vpiName, hdl);
std::string idx_str = name.substr(name.find_last_of(hdl_name)+hdl_name.length());
std::count(idx_str.begin(), idx_str.end(), '[');
```

Then *unconditionally* do the part in the MODELSIM ifdef.

src/cocotb/share/lib/vpi/VpiObj.cpp
```diff
@@ -47,64 +48,52 @@ int VpiArrayObjHdl::initialise(const std::string &name,
...
-            }
+        // get the last index of hdl_name in name
+        std::size_t idx_str = name.rfind(hdl_name);
+        if (idx_str == std::string::npos) {
+            LOG_ERROR("Unable to find name %s in %s", hdl_name.c_str(),
+                      name.c_str());
+            return -1;
         }
+        // count occurences of [
+        range_idx = (int)std::count(name.begin() + idx_str, name.end(), '[');
```
ktbarrett, line 59:
```suggestion
        range_idx = (int)std::count(&name.begin()[idx_str], name.end(), '[');
```

tests/test_cases/test_iteration_verilog/test_iteration_es.py
```diff
@@ -45,25 +45,35 @@ async def recursive_discovery(dut):
         # everyone seems to find the byteswap() function besides verilator
         pass_total = 27
 
+    # Icarus doesn't support array indexes like get_handle_by_name("some_path[x]")
+    SKIP_HANDLE_ASSERT = cocotb.SIM_NAME.lower().startswith(("riviera", "icarus"))
+    # SKIP_HANDLE_ASSERT = cocotb.SIM_NAME.lower().startswith(("icarus")
```
ktbarrett, line 50: Commented out code.

ktbarrett (approved):
Looks good now.

ktbarrett:
The failing test is unrelated. A combination of PRs that passed their tests separately don't pass together after merging. Whoops...

PR #3631, opened 2024-01-03
CI: Show diffs in Lint action logs where possible

marlonjames:
Running `pre-commit` locally should change the files in question,
but for CI it is more useful to show what would be changed rather than just Passed/Failed for each hook.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -24,4 +24,4 @@ jobs:
     - name: pre-commit
       run: |
         pip install pre-commit
-        pre-commit run -a
+        pre-commit run -a --show-diff-on-failure
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,6 +21,7 @@ repos:
   hooks:
   - id: "clang-format"
     exclude: "^src/cocotb/share/include/(sv_vpi|vhpi|vpi)_user(_ext)?.h"
+    types_or: [c, c++]
 
 - repo: "https://github.com/pre-commit/pre-commit-hooks"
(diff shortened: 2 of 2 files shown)


marlonjames:
I found a better diff output for `clang-format`, but it is not setup for pre-commit, so I think this is the best we can do without additional work.

ktbarrett:
Should make sure this is also kept up to date with the pre-commit autoupdate action.

ktbarrett (commented):
I wish it were somehow possible to not duplicate these files... I'm fine with this, lets see what @imphil says.

imphil:
Thanks @marlonjames for looking at ways to make contributions easier! I think there's value in presenting users with the diff of the changes, this way they can manually apply them without having to setup pre-commit.

An alternative to having two pre-commit configuration files would be running pre-commit as-is and then do a `git diff` afterwards. You loose the information which tool reformatted which piece of code (which isn't overly interesting anyways), but can get away with a single unmodified config file.

marlonjames:
Good suggestion @imphil 

Output with git diff:
https://github.com/cocotb/cocotb/actions/runs/7415178186/job/20177799321?pr=3631#step:4:247

marlonjames:
I'll remove the failing commit once this has been reviewed.

cmarqu:
I just found that `pre-commit run` also has a `--show-diff-on-failure` option:
https://pre-commit.com/#pre-commit-run

imphil (approved):
LGTM, "precommit run -a --show-diff-on-failures" is probably the easier option. Whatever you prefer, @marlonjames.

PR #3632, opened 2024-01-03
Print exception before launching pdb

AndrewNolte:
When running with COCOTB_PDB_ON_EXCEPTION, the exception wasn't getting printed out

### DIFF
--- a/src/cocotb/regression.py
+++ b/src/cocotb/regression.py
@@ -460,6 +460,7 @@ class RegressionManager:
             result_pass = False
 
             if _pdb_on_exception:
+                print(result)
                 pdb.post_mortem(result.__traceback__)
 
         return result_pass, sim_failed


src/cocotb/regression.py
```diff
@@ -460,6 +460,7 @@ def _score_test(self, test: Test, outcome: Outcome) -> Tuple[bool, bool]:
             result_pass = False
 
             if _pdb_on_exception:
+                print(result)
```
ktbarrett, line 463: `traceback.print_exception` might provide more utility.

AndrewNolte: It looks like the print the same exact thing, but I can change it

AndrewNolte: Oh actually had the wrong worktree loaded, but I realize now it's printing twice because it's in the test summary. It probably wasn't printing for us because we modify the log level of cocotb's logger

PR #3633, opened 2024-01-04
Decouple implementation of Array and LogicArray

ktbarrett:
Decouples the implementation of `Array` and `LogicArray`. We create a Protocol for ArrayLike types and move a lot of common functionality into that class. We duplicate the implementation of Array into LogicArray with the anticipation that the implementation of LogicArray will change in the near future for performance reasons.

### DIFF
--- a/src/cocotb/types/__init__.py
+++ b/src/cocotb/types/__init__.py
@@ -1,15 +1,33 @@
 # Copyright cocotb contributors
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
-import typing
+from abc import ABC, abstractmethod
+from typing import Any, Generic, Iterable, Iterator, Optional, TypeVar, Union, overload
 
-from .array import Array  # noqa: F401
-from .logic import Logic  # noqa: F401
-from .logic_array import LogicArray  # noqa: F401
-from .range import Range  # noqa: F401
+try:
+    from typing import Protocol
+except ImportError:
+    try:
+        from typing_extensions import Protocol
+    except ImportError:
+        Protocol = ABC
+
+T = TypeVar("T")
+Self = TypeVar("Self")
 
 
-def concat(a: typing.Any, b: typing.Any) -> typing.Any:
+class Concatable(Protocol, Generic[T]):
+    def __concat__(self: Self, other: Self) -> Self:
+        return NotImplemented
+
+    def __rconcat__(self: Self, other: Self) -> Self:
+        return NotImplemented
+
+
+ConcatableT = TypeVar("ConcatableT", bound=Concatable[Any])
+
+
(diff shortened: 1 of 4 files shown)


src/cocotb/types/__init__.py
```diff
@@ -46,3 +64,117 @@ def concat(a: typing.Any, b: typing.Any) -> typing.Any:
...
+
+    @property
+    @abstractmethod
+    def range(self) -> Range:
+        """:class:`Range` of the indexes of the array."""
+
+    @range.setter
+    @abstractmethod
+    def range(self, new_range: Range) -> None:
+        """Sets a new indexing scheme on the array, must be the same size"""
```
cmarqu, line 96:
```suggestion
        """Set a new indexing scheme on the array. Must be the same size."""
```

src/cocotb/types/__init__.py
```diff
@@ -46,3 +64,117 @@ def concat(a: typing.Any, b: typing.Any) -> typing.Any:
...
+
+
+class ArrayLike(Concatable[T], Protocol, Generic[T]):
+    @property
+    def left(self) -> int:
+        """Leftmost index of the array."""
+        return self.range.left
+
+    @property
+    def direction(self) -> str:
```
cmarqu, line 79: Maybe this could return something like a Py3.11 `StrEnum`?

src/cocotb/types/logic_array.py
```diff
@@ -179,6 +180,40 @@ def __init__(
...
             )
 
+    @property
+    def range(self) -> Range:
+        """:class:`Range` of the indexes of the array."""
+        return self._range
+
+    @range.setter
+    def range(self, new_range: Range) -> None:
+        """Sets a new indexing scheme on the array, must be the same size"""
```
cmarqu, line 190:
```suggestion
        """Set a new indexing scheme on the array. Must be the same size."""
```

src/cocotb/types/__init__.py
ktbarrett: Something like this?

```python
class Direction(str):
    def __new__(cls: Type[Self], value: str) -> Self:
        value = value.lower()
        if value not in ("to", "downto"):
            raise ValueError(f"{value!r} not one of the possible enumerations ('to', 'downto')")
        return super().__new__(cls, value)

TO = Direction("to")
DOWNTO = Direction("downto")
```

String interning means we don't have to work about singletons and whatnot.

cmarqu: Not sure if it's worth the effort, but maybe it helps catching some problems earlier.
About identity... really no idea :)

ktbarrett: It might be, but that's definitely a follow-up PR.

marlonjames (approved):

PR #3634, opened 2024-01-04
Remove `BinaryValue`

### LINKED ISSUE #2059: [tracking] Redesign BinaryValue

ktbarrett:
Closes #2059.

Partial rework of #3244. That PR does a *lot* more than simply remove `BinaryValue`. This PR intends only to remove `BinaryValue` and update documentation and tests related to `BinaryValue` and nothing more.

To accomplish that without changing a ton of code I had to add support for comparison with `int` to `LogicArray` and `Logic`. As a consequence this means I had to remove the hashability of `Logic`, which I don't think should affect anyone. Follow-ons might work a little more on this.

I think it's unfortunate, but unfortunately necessary, to do these cross-type equalities since it's very verbose and would require a lot of changes otherwise. The way Python was designed means that `__eq__` needs to be strict equality (despite the rest of the language being "if it looks and sounds like a duck") and we are kind of breaking that unspoken idiom, which also means we can't support any of these types in hashed collections, but there isn't much we can do. Rock and a hard place...


ktbarrett:
Needs newsfrag.

cmarqu (approved):

ktbarrett:
Added a newsfragment that will eventually point to a guide on how to upgrade or maybe this will be changed in the future. It's all very TBD rn.

cmarqu (approved):

ktbarrett:
Well I guess we did agree on this change ahead of time and there is still plenty of time to go back on the handful of changes made before the 2.0 release.

PR #3637, opened 2024-01-04
Move around commandline args in Makefile.icarus

### LINKED ISSUE #3635: Unable to override icarus verilog language generation
I am unable to override icarus verilog language generation by passing the `-g` flag in `COMPILE_ARGS`. It seems that icarus only keeps the last flag value.

The following command produced by the Makefile succeeds which is not the expected behaviour if the language generation is overwritten.
`iverilog -o sim_build/sim.vvp -D COCOTB_SIM=1 -s alu -g 2001 -f sim_build/cmds.f -g2012 alu.v`

The following command fails with a syntax error which is expected if the language generation is overwritten.
`iverilog -o sim_build/sim.vvp -D COCOTB_SIM=1 -s alu -g2012 -g 2001 -f sim_build/cmds.f alu.v`

## Files to reproduce
alu.v
```verilog
enum {
    ADD
} operation;

module alu();

endmodule
```

Makefile
```makefile
SIM ?= icarus
TOPLEVEL_LANG ?= verilog

VERILOG_SOURCES += alu.v

TOPLEVEL = alu

MODULE = test_alu

COMPILE_ARGS = -g 2001

```

test_alu.py
```python
import cocotb

@cocotb.test()
async def test(dut):
    pass
```

## System information
- cocotb version: 1.8.0
- linux version: 6.6.9
- python version: 3.11.4
- icarus verilog version: 12.0

ktbarrett:
Moves default language version in iverilog commandline args earlier to allow the user to override it later with COMPILE_ARGS. Closes #3635.

### DIFF
--- a/src/cocotb/share/makefiles/simulators/Makefile.icarus
+++ b/src/cocotb/share/makefiles/simulators/Makefile.icarus
@@ -62,7 +62,7 @@ else
   TOPMODULE_ARG :=
 endif
 
-COMPILE_ARGS += -f $(SIM_BUILD)/cmds.f -g2012 # Default to latest SystemVerilog standard
+COMPILE_ARGS += -f $(SIM_BUILD)/cmds.f
 
 ifdef VERILOG_INCLUDE_DIRS
     COMPILE_ARGS += $(addprefix -I, $(VERILOG_INCLUDE_DIRS))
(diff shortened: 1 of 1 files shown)


imphil (approved):
Thanks!

Good enough. We got better over time to not add any argument that's not strictly required for cocotb to work, but the ship has sailed here.

PR #3642, opened 2024-01-05
Ensure initial happens before cocotb starts

toddstrader:
Tests were starting before initial blocks had been evaluated for Verilator sims.  This fixes the issue and adds a test to demonstrate / enforce consistent behavior.

### DIFF
--- a/docs/source/newsfragments/3642.bugfix.rst
+++ b/docs/source/newsfragments/3642.bugfix.rst
@@ -1,0 +1,1 @@
+Fix tests being started before initial blocks are evaluated under Verilator.
--- a/src/cocotb/share/lib/verilator/verilator.cpp
+++ b/src/cocotb/share/lib/verilator/verilator.cpp
@@ -58,6 +58,9 @@ int main(int argc, char** argv) {
     Verilated::internalsDump();
 #endif
 
+    // Evaluate initial blocks before starting cocotb
+    top->eval_step();
+
     vlog_startup_routines_bootstrap();
     VerilatedVpi::callCbs(cbStartOfSimulation);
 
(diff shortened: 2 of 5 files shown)


toddstrader:
So I know Xcelium does `initial` before the tests start.  I clearly didn't test Icarus, but see now that it does not.  And I'm waiting to see what the other simulators do.  Is this something that is desired?  `await ReadWrite()` at the beginning of time also seems to do the trick.  If consistent behavior is desired across all simulators and we want initial to happen before any python, then we could do that `await` somewhere in `__initial__.py`.  Otherwise I can do it before all my tests.  Also, I'm not sure if there would be other side effects from that `await` which would cause problems.

ktbarrett:
I think the current behavior is the ideal behavior. The simulation should start at time 0, not in `ReadWrite` after time 0. Properly functioning simulators shouldn't be evaluating `initial` blocks before the first cocotb test starts. `initial` blocks are not initializers! They are just non-looping processes like `always`. If initial values are desired, use initializers when declaring a wire/reg.

toddstrader:
> initial blocks are not initializers!
Agreed.  However, this was the only way I could see to enforce uniformity.  But it's true that it does remove the ability to interact with the DUT before `initial` happens.  Just an FYI that, in this case, Xcelium is not a `properly functioning simulators`.

ktbarrett:
> Xcelium is not a `properly functioning simulators`.

Well we already knew that. It's unfortunate that behavior isn't consistent, and this is certainly not the only case where it isn't, but people aren't paying for a "simulator that behaves consistently", they are paying for a "simulator." And the standard isn't always the most precise thing, so you can't even blame the companies producing "simulators."

Issue #1884, opened 2020-05-29
No access to signals below 'generate' hierarchy for Verilator


stu-e:
Signals below a point in the HDL hierarchy instantiated with a generate-if statement are 'invisible' to the Python test bench.
I placed the endian swapper module within a generate block within a wrapper module and this is the error:

`Traceback (most recent call last):
  File "/home/stuart/cocotb/examples/endian_swapper/tests/test_endian_swapper.py", line 59, in stream_out_config_setter
    if dut.MY_GEN_BLOCK.i_endian_swapper.byteswapping.value:
  File "/home/stuart/.local/lib/python3.8/site-packages/cocotb/handle.py", line 295, in __getattr__
    raise AttributeError("%s contains no object named %s" % (self._name, name))
AttributeError: endian_swapper_wrapper contains no object named MY_GEN_BLOCK`

Looking at the Verilator Vtop.h file, the path is correct

themperek:
Most lily ventilator issue.

garmin-mjames:
Could you post a log with environment variable `COCOTB_LOG_LEVEL=DEBUG`? It should show debug log messages related to traversing the hierarchy.

stu-e:
[endian_swapper_debug.txt](https://github.com/cocotb/cocotb/files/4713067/endian_swapper_debug.txt)

garmin-mjames:
Related to #1432. The problem appears to be that Verilator gives a type of `vpiScope` for `MY_GEN_BLOCK`, which cocotb does not currently support.

stu-e:
Ok thanks for checking

imphil:
@stu-e can you come up with a minimal Verilog and cocotb example that we can add to our test bench to keep track of this issue? You can either paste it here, or (if you feel comfortable) open an pull request.

I guess we'll then need to have a closer look what the next steps could be; I wouldn't be surprised if this is something that needs to be fixed in Verilator.

garmin-mjames:
I think `test_discovery.pseudo_region_access` covers this.

AndrewNolte:
I put a vpi dump test case in Verilator here https://github.com/verilator/verilator/pull/4838/files#diff-5c0aca2db885444f49208c5bbbf322d83b90b08ffb0c320426f493974f59057e The vpiNames look correct in that .out file, but they still don't show up on cocotb.

ktbarrett:
@AndrewNolte Is this issue still present?

stu-e:
[...]

PR #3664, opened 2024-01-16
Allow indexing to vpiscope objects

AndrewNolte:
This allows for accessing vpiScope objects, but they are still invisible in iteration in verilator at least. Addresses https://github.com/cocotb/cocotb/issues/1884

Questions: 

- Can someone confirm that the pseudo-handle stuff is only needed for modelsim/questa? without that ifdef the nine_int inside the scope was not found, because the scope was GPI_GEN_ARRAY
- Is a better solution to just have verilator to return vpiGenScope? What's the correct thing to do here?

tagging @marlonjames in case there's any repeated work

### DIFF
--- a/src/cocotb/share/lib/vpi/VpiCbHdl.cpp
+++ b/src/cocotb/share/lib/vpi/VpiCbHdl.cpp
@@ -596,7 +596,7 @@ decltype(VpiIterator::iterate_over) VpiIterator::iterate_over = [] {
         vpiPrimitive, vpiPrimitiveArray,
         // vpiContAssign,        // Don't care
         vpiProcess,  // Don't care
-        vpiModPath, vpiTchk, vpiAttribute, vpiPort, vpiInternalScope,
+        vpiModPath, vpiTchk, vpiAttribute, vpiPort, vpiInternalScope, vpiScope,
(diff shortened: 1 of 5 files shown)


marlonjames:
`vpiScope` is a 1-to-1 relationship, not an object type.
Instead of this, Verilator should be updated to support `vpiGenScope` and `vpiGenScopeArray` object types.

As far as I know icarus still doesn't support `vpiGenScopeArray`, only `vpiGenScope`, so we still need pseudo-region objects for that.

AndrewNolte:
Ah I see now. So simply changing that to vpiGenScope on verilator, and removing the pseudo-regision creation allows access to gen scope, conditional scopes, but not array scopes. I think vpiName for named scopes on verilator just always do parent.scope instead of just 'scope', so it tried to make those as GPI_GENARRAY incorrectly. But we need pseudo-regions for loop scopes, because those don't have a corresponding handle on Verilator.

ktbarrett:
Can this PR be completed in some new way?

marlonjames:
I think both this and #3624 will be unnecessary if Verilator updates its gen scope to return the right type and fixes anywhere that has `vpiName` match `vpiFullName` when it shouldn't.

Issue #3662, opened 2024-01-16
Vpi Package: >32 bit parameters are interpreted as Integers


AndrewNolte:
Large parameters are mapped to GPI_INTEGER, which leads to issues when trying to read from them. I think we should check the size of the parameters before assigning to GPI_INTEGER

ktbarrett:
Which simulator does this?

AndrewNolte:
I'm seeing it on verilator, will test on xcelium as well

ktbarrett:
If you write up a test you can open a PR with the test and run it against all the simulators? Even if we have to merge it with everything marked fail, that's still good info.

ktbarrett:
Closed by #3729.

PR #3665, opened 2024-01-17
Long parameter test case

AndrewNolte:
Verilator raises an error when calling vpiIntVal on long parameters in packages, however Xcelium lets this through. We should be checking the size of these parameters, as the Vpi doesn't specify the size for these types on vpiDecConst, vpiBinConst, vpiHexConst and vpiOctConst. I think we ought to set the format to vpiLongIntVal for 64 bit integers, not sure how we want to approach this.

- Adds check for size on ambiguous types
- Adds test case for parameters > 32 bits, and longint

https://github.com/cocotb/cocotb/issues/3662

### DIFF
--- a/tests/test_cases/test_package/cocotb_package_pkg.sv
+++ b/tests/test_cases/test_package/cocotb_package_pkg.sv
@@ -5,8 +5,8 @@
 package cocotb_package_pkg_1;
     parameter int five_int = 5;
     parameter logic [31:0] eight_logic = 8;
-    parameter logic [63:0] long_param = 64'hFF;
-    parameter logic [99:0] really_long_param = 100'hFF;
+    parameter longint unsigned long_param = 64'h5a89901af1;
+    parameter logic [99:0] really_long_param = 100'h5a89901af1;
 endpackage
 
 package cocotb_package_pkg_2;
--- a/tests/test_cases/test_package/cocotb_package_pkg.sv
+++ b/tests/test_cases/test_package/cocotb_package_pkg.sv
@@ -5,6 +5,8 @@
 package cocotb_package_pkg_1;
     parameter int five_int = 5;
     parameter logic [31:0] eight_logic = 8;
+    parameter logic [63:0] long_param = 64'hFF;
+    parameter logic [99:0] really_long_param = 100'hFF;
 endpackage
 
 package cocotb_package_pkg_2;
--- a/tests/test_cases/test_package/test_package.py
+++ b/tests/test_cases/test_package/test_package.py
@@ -41,7 +41,7 @@ async def test_stringification(dut):
     assert str(pkg2.eleven_int) == "IntegerObject(cocotb_package_pkg_2::eleven_int)"
 
 
-@cocotb.test(expect_fail=cocotb.SIM_NAME.lower().startswith("verilator"))
+@cocotb.test()
 async def test_long_parameter(dut):
     # On verilator:
     # 0.00ns ERROR    gpi                                VPI error
@@ -50,8 +50,17 @@ async def test_long_parameter(dut):
 
     # returns 0 on verilator for unsupported format
     pkg1 = cocotb.packages.cocotb_package_pkg_1
-    assert pkg1.long_param.value == 255
-    assert pkg1.really_long_param.value == 255
+    assert pkg1.five_int.value == 5
+    assert str(pkg1.five_int) == "IntegerObject(cocotb_package_pkg_1::five_int)"
+
+    assert pkg1.long_param.value.integer == int("5a89901af1", 16)
+    assert str(pkg1.long_param) == "LogicObject(cocotb_package_pkg_1::long_param)"
+
+    assert pkg1.really_long_param.value.integer == int("5a89901af1", 16)
+    assert (
+        str(pkg1.really_long_param)
+        == "LogicObject(cocotb_package_pkg_1::really_long_param)"
+    )
 
 
 @cocotb.test()
--- a/tests/test_cases/test_package/test_package.py
+++ b/tests/test_cases/test_package/test_package.py
@@ -41,14 +41,17 @@ async def test_stringification(dut):
     assert str(pkg2.eleven_int) == "IntegerObject(cocotb_package_pkg_2::eleven_int)"
 
 
-@cocotb.test()
+@cocotb.test(expect_fail=cocotb.SIM_NAME.lower().startswith("verilator"))
 async def test_long_parameter(dut):
-    # On verilator:
-    # 0.00ns ERROR    gpi                                VPI error
-    # 0.00ns ERROR    gpi                                vl_check_format: Unsupported format (vpiIntVal) for cocotb_package_pkg_1.long_param
-    # 0.00ns ERROR    cocotb.regression                  Failed to initialize test test_long_parameter
+    """
+    Test package parameter access
 
-    # returns 0 on verilator for unsupported format
+    # On verilator:
+    # 0.00ns ERROR    VPI error
+    # 0.00ns ERROR    vl_check_format: Unsupported format (vpiIntVal) for cocotb_package_pkg_1.long_param
+    # 0.00ns ERROR    Failed to initialize test test_long_parameter
+    """
+
     pkg1 = cocotb.packages.cocotb_package_pkg_1
     assert pkg1.five_int.value == 5
     assert str(pkg1.five_int) == "IntegerObject(cocotb_package_pkg_1::five_int)"
--- a/tests/test_cases/test_package/test_package.py
+++ b/tests/test_cases/test_package/test_package.py
@@ -41,7 +41,7 @@ async def test_stringification(dut):
     assert str(pkg2.eleven_int) == "IntegerObject(cocotb_package_pkg_2::eleven_int)"
 
 
-@cocotb.test(expect_fail=cocotb.SIM_NAME.lower().startswith("verilator"))
+@cocotb.test()
 async def test_long_parameter(dut):
     """
     Test package parameter access
@@ -56,8 +56,12 @@ async def test_long_parameter(dut):
     assert pkg1.five_int.value == 5
     assert str(pkg1.five_int) == "IntegerObject(cocotb_package_pkg_1::five_int)"
 
-    assert pkg1.long_param.value.integer == int("5a89901af1", 16)
-    assert str(pkg1.long_param) == "LogicObject(cocotb_package_pkg_1::long_param)"
+    if cocotb.SIM_NAME.lower().startswith(("verilator", "icarus")):
+        assert pkg1.long_param.value == int("5a89901af1", 16)
+        assert str(pkg1.long_param) == "IntegerObject(cocotb_package_pkg_1::long_param)"
+    else:
+        assert pkg1.long_param.value.integer == int("5a89901af1", 16)
+        assert str(pkg1.long_param) == "LogicObject(cocotb_package_pkg_1::long_param)"
 
     assert pkg1.really_long_param.value.integer == int("5a89901af1", 16)
     assert (
--- a/tests/test_cases/test_package/test_package.py
+++ b/tests/test_cases/test_package/test_package.py
@@ -60,24 +60,19 @@ async def test_long_parameter(dut):
     """
 
     pkg1 = cocotb.packages.cocotb_package_pkg_1
-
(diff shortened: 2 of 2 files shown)


ktbarrett (approved):
I expect an upstream ticket on verilator was made?

AndrewNolte:
Is the expected behavior that vpiInteger should truncate the value?

ktbarrett:
The issue is that the parameter is being mapped to `GPI_INTEGER`. I can only imagine this is due to some issue with Verilator since no other simulator does that.

AndrewNolte:
Xcelium actually also maps it to GPI_INTEGER, but returns the correct value

ktbarrett:
Do you know what `VPI` type is being returned? Is this something cocotb is doing, or are the simulators after some size of logic vector switching over to integer as the VPI type?

I think we should be adding `assert isinstance(param_handle, LogicObject)` to this test if Xcelium is also retuning an inconsistent type. This PR is primarily to get the test in, so after that is added I'll merge. But it's nice to get some debugging done so we work on next steps.

tests/test_cases/test_package/test_package.py
```diff
@@ -41,6 +41,19 @@ async def test_stringification(dut):
...
+async def test_long_parameter(dut):
+    # On verilator:
+    # 0.00ns ERROR    gpi                                VPI error
+    # 0.00ns ERROR    gpi                                vl_check_format: Unsupported format (vpiIntVal) for cocotb_package_pkg_1.long_param
+    # 0.00ns ERROR    cocotb.regression                  Failed to initialize test test_long_parameter
+
+    # returns 0 on verilator for unsupported format
+    pkg1 = cocotb.packages.cocotb_package_pkg_1
+    assert pkg1.long_param.value == 255
+    assert pkg1.really_long_param.value == 255
```
ktbarrett, line 54:
```suggestion
    assert isinstance(pkg1.long_param, LogicObject)
    assert pkg1.long_param.value == 255
    assert isinstance(pkg1.really_long_param, LogicObject)
    assert pkg1.really_long_param.value == 255
```

tests/test_cases/test_package/test_package.py
```diff
@@ -41,6 +41,19 @@ async def test_stringification(dut):
     assert str(pkg2.eleven_int) == "IntegerObject(cocotb_package_pkg_2::eleven_int)"
 
 
+@cocotb.test(expect_fail=cocotb.SIM_NAME.lower().startswith("verilator"))
+async def test_long_parameter(dut):
+    # On verilator:
+    # 0.00ns ERROR    gpi                                VPI error
+    # 0.00ns ERROR    gpi                                vl_check_format: Unsupported format (vpiIntVal) for cocotb_package_pkg_1.long_param
+    # 0.00ns ERROR    cocotb.regression                  Failed to initialize test test_long_parameter
```
ktbarrett, line 49:
```suggestion
    # 0.00ns ERROR    cocotb.regression                  Failed to initialize test test_long_parameter
    from cocotb.handle import LogicObject
```

marlonjames:
We could check `vpiSize` in `const_type_to_gpi_objtype()` and use `GPI_REG` instead of `GPI_INTEGER` if it's larger than 32 bits.

marlonjames:
> Do you know what VPI type is being returned?

It's a `vpiParameter` with `vpiConstType` of `vpiDecConst`.

AndrewNolte:
^ @marlonjames I had it return GPI_NET because that's what we had for that xcelium edge case, should I change both of those to GPI_REG?

marlonjames:
I don't think there are any differences in how `REG` vs `NET` objects are handled. They both map to `LogicObject` in Python and `VpiSignalObjHdl` in GPI code. `GPI_NET` is fine with me.

src/cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -169,6 +170,11 @@ static gpi_objtype_t const_type_to_gpi_objtype(int32_t const_type) {
         case vpiBinaryConst:
         case vpiOctConst:
         case vpiHexConst:
+            if (vpi_get(vpiSize, hdl) <= 32) {
+                return GPI_INTEGER;
+            } else {
+                return GPI_NET;
+            }
```
ktbarrett, line 177: Shouldn't this always be `GPI_NET`? Or are we making a bad assumption somewhere else?

tests/test_cases/test_package/cocotb_package_pkg.sv
```diff
@@ -5,6 +5,8 @@
 package cocotb_package_pkg_1;
     parameter int five_int = 5;
     parameter logic [31:0] eight_logic = 8;
+    parameter longint unsigned long_param = 64'h5a89901af1;
+    parameter logic [99:0] really_long_param = 100'h5a89901af1;
```
ktbarrett, line 9: I really don't know much about SV. Seems like a confusingly inconsistent language to me, but what do I know...

Is `longint unsigned` an integer type where values can be 0/1 only, or is it a subtype of a logic vector?

marlonjames: image

src/cocotb/share/lib/vpi/VpiImpl.cpp
marlonjames: I think there are probably several issues with integer nets/var/parameters.

For one, `IntegerObject` (mapped to `GPI_INTEGER`) assumes 32-bits when checking overflow:
https://github.com/cocotb/cocotb/blob/d6c1a2ec9fa3e41d071c8e5ee2e9965dd9295a7d/src/cocotb/handle.py#L941

AndrewNolte: Yeah I really don't like vpiConst types, I don't think they make any sense. The subtypes for that seem to more accurately describe the format of literals in the CST, which aren't useful to the sim. And not every parameter is a literal. Do you think we can change these back on verilator, and instead use the vpiConstantVariable property to mark them as constant? These technically aren't variables, but I think that's a much better way than this for describing params @marlonjames

ktbarrett: I had the same thought with `vpiConstantVariable`. I think we should be using that when we can, but I'm going out on a limb and assuming simulator support is non-existant. It might be worth looking to see if any simulators support it. And I wonder if there's a way to look at the parameter type name so we can map everything perfectly for the handful of types listed below by Marlon.

What we want to see is 4 value types (which includes `integer`) map to logic and 2-value types to ints.

marlonjames: I think most simulators are going to return type as `vpiParameter`, not a variable, so that property won't work for those.

marlonjames: @ktbarrett parameter objects have a `vpiTypespec` association that would give you the info, but I bet support for that is even lower than everything else discussed.

marlonjames: Thinking about it, I think the `vpiConstType` maps to the `s_vpi_value.format` for retrieving the value.
`vpiDecConst` => `vpiDecStrVal`,
`vpiBinaryConst` => `vpiBinStrVal`,
`vpiOctConst` => `vpiOctStrVal`,
`vpiHexConst` => `vpiHexStrVal`,
`vpiIntConst` => `vpiIntVal`

@ktbarrett I don't know how to map 4-state vs 2-state unless we can get the Typespec info.
I'm going to write up a small VPI program so we can see what info simulators are giving for different parameters.

ktbarrett: `vpiConstType` is probably a misleading name. I think Andrew is correct that I think it's more about the format of the *literal* used to initialized the parameter, and less about the actual type of the parameter, which is what we care about.

ktbarrett:
I think this PR has gotten off the rails since it has uncovered something we aren't sure how to proceed from. Let's ignore the C++ changes, encode what we are expected in the test, and expect_fail/skip everything that causes an issue. That might be the whole damn test, but so be it...

I expect `int` and `longint unsigned` to be an `IntegerObject`. I expect `logic[31:0]` and `logic[99:0]` to be a `LogicObject`. And of course we expect the values to be correctly retrieved.

tests/test_cases/test_package/test_package.py
```diff
@@ -41,6 +41,31 @@ async def test_stringification(dut):
...
+    # 0.00ns ERROR    vl_check_format: Unsupported format (vpiIntVal) for cocotb_package_pkg_1.long_param
+    # 0.00ns ERROR    Failed to initialize test test_long_parameter
+    """
+
+    pkg1 = cocotb.packages.cocotb_package_pkg_1
+    assert pkg1.five_int.value == 5
+    assert str(pkg1.five_int) == "IntegerObject(cocotb_package_pkg_1::five_int)"
+
+    assert pkg1.long_param.value.integer == int("5a89901af1", 16)
+    assert str(pkg1.long_param) == "LogicObject(cocotb_package_pkg_1::long_param)"
```
ktbarrett, line 60: `long_param` should be an integer (and Icarus and Verilator seem to think it is).

AndrewNolte: Could we get __int__ defined on logic array so int(pkg1.long_param.value) would work for both interpretations?

Also, I had expect_fail set for verilator, am I doing something wrong there?

AndrewNolte:
Why should we want longint unsigned to be mapped to integer object if the bounds check is for 32 bits? I'm thinking of just changing the const type mapping to only have vpiIntConst map to integer, and the others to GPI_REG

tests/test_cases/test_package/test_package.py
```diff
@@ -41,6 +41,40 @@ async def test_stringification(dut):
...
+        # most sims truncate the value to 32 bits, and interpret as signed despite the marking
+        assert str(pkg1.long_param) == "IntegerObject(cocotb_package_pkg_1::long_param)"
+        # should really be 0x5A89901AF1 (40 bits)
+        # -1987044623 = (two's comp) 0b10001001100100000001101011110001 = 0x5A89901AF1[31:0]
+        assert get_integer(pkg1.long_param) == -1987044623
+        assert (
+            str(pkg1.really_long_param)
+            == "IntegerObject(cocotb_package_pkg_1::really_long_param)"
+        )
+        assert get_integer(pkg1.really_long_param) == -1987044623
```
AndrewNolte, line 75: I think the fact that this passes with the wrong value on all sims really shows the shortcomings of this. I think it makes sense to have only have vpiIntConst map to GPI_INTEGER

marlonjames: I've tested all of the SV simulators except Xcelium, and none of them use `vpiIntConst` at all.
Only `vpiBinaryConst` and `vpiDecConst`.

AndrewNolte: So is the interpretation that vpiBinaryConst = logic, and vpiDecConst = int (and longint) ? So then the size check is what we want to go with? We can discuss here https://github.com/cocotb/cocotb/pull/3691. Could we get this tc landed for now?

marlonjames:
There is no consistency with respect to what is `vpiBinaryConst` vs `vpiDecConst`. Only Xcelium supports `vpiTypespec`.

Given that, I think the best path for consistency is to make all of these constants `LogicObject`s, and have the user convert to integer as needed.

ktbarrett:
We will also bring back the `int` cast so you can just cast to int without having to worry about whether it's an `int` or `LogicArray`.

tests/test_cases/test_package/cocotb_package_pkg.sv
marlonjames: While we are adding more parameter types, might as well cover the gamut.
You can pull from https://github.com/marlonjames/vpi-experiments/blob/ab1665e3bbaba4522220efc343fc8a64c6eb107b/int_param_check.sv if you want.

src/cocotb/share/lib/vpi/VpiImpl.cpp
marlonjames: For the edification of anybody reading this in the future, you can use the `int_param_check` VPI program from https://github.com/marlonjames/vpi-experiments to get information about `vpiConstType` for your simulator of choice.

marlonjames:
Closed in favor of #3729

PR #3666, opened 2024-01-18
Don't require Verilator binary for testing

toddstrader:
Verilator is different from other simulators in that the tool install is not required past the build stage.  Requiring the `verilator` binary to be present when testing presents challenges, especially in distributed  compute settings.

I'm not exactly sure how to test this in CI, but it's getting me past my current problem locally.

### DIFF
--- a/src/cocotb/runner.py
+++ b/src/cocotb/runner.py
@@ -946,6 +946,10 @@ class Verilator(Simulator):
     supported_gpi_interfaces = {"verilog": ["vpi"]}
 
     def _simulator_in_path(self) -> None:
+        # the verilator binary is only needed for building
+        return
+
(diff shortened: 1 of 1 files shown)


toddstrader:
> I'm not exactly sure how to test this in CI

Let me rephrase.  I can imagine a script which builds a Verilator target, removes Verilator from `$PATH` and then runs the test.  However, that sounds painful / annoying to write especially since there will be likely be differences between my setup and the GitHub CI which would make it necessary to debug via GitHub itself.

So maybe a better way to put this is that I don't believe we're currently testing negative cases for `_simulator_in_path()` and that I'm hopeful that this change is logical / straightforward enough to warrant the same approach.

themperek (approved):
Looks ok to me to be this way.

PR #3667, opened 2024-01-18
Add Verilator run-time trace switch

toddstrader:
Adds support for enabling tracing at test-time for Verilator.

This doesn't exactly fit the mold of the runner class, but there is also not a consistent story across the various simulators right now.  Currently some simulators (e.g. Icarus) will only compile in trace support if `waves` is `True`, while others (e.g. Xcelium) always compile in trace support and only observe `waves` at test time.

If the status quo is fine, I'm happy to leave things as-is.  However, if it's desirable to split `waves` into `build_for_waves` and `waves` (or whatever) I can add that here as well.

This is useful to us (and probably others) since monolithic Verilator builds are painfully slow once you reach a certain point.  I don't believe solving that problem is in scope for the cocotb runner, so we (and possibly others) will be using `test()` but not `build()` for Verilator.

Related: Wilson has some good notes on Verilator building here:
https://veripool.org/papers/Verilator_Accelerated_OSDA2020.pdf

### DIFF
--- a/docs/source/newsfragments/3667.feature.rst
+++ b/docs/source/newsfragments/3667.feature.rst
@@ -1,1 +1,1 @@
-Add `--trace` to Verilator binaries for run-time trace generation and integrate with :ref:`Python Test Runner <howto-python-runner>`
+Add `--trace` to Verilator binaries for run-time trace generation and integrate with :ref:`Python Test Runner <howto-python-runner>`.
--- a/docs/source/newsfragments/3667.feature.rst
+++ b/docs/source/newsfragments/3667.feature.rst
@@ -1,0 +1,1 @@
+Add `--trace` to Verilator binaries for run-time trace generation and integrate with :ref:`Python Test Runner <howto-python-runner>`
--- a/src/cocotb/share/lib/verilator/verilator.cpp
+++ b/src/cocotb/share/lib/verilator/verilator.cpp
@@ -85,7 +85,7 @@ int main(int argc, char** argv) {
     std::unique_ptr<VerilatedFstC> tfp(new VerilatedFstC);
     const char* traceFile = "dump.fst";
 #else
-    std::unique_ptr<VerilatedvcdC> tfp(new VerilatedVcdC);
+    std::unique_ptr<VerilatedVcdC> tfp(new VerilatedVcdC);
     const char* traceFile = "dump.vcd";
 #endif
 
--- a/src/cocotb/share/lib/verilator/verilator.cpp
+++ b/src/cocotb/share/lib/verilator/verilator.cpp
@@ -47,6 +47,25 @@ static inline bool settle_value_callbacks() {
 }
 
 int main(int argc, char** argv) {
+    bool traceOn = false;
+
+    for (int i = 1; i < argc; i++) {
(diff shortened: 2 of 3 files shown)


src/cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -62,16 +81,19 @@ int main(int argc, char** argv) {
...
     std::unique_ptr<VerilatedFstC> tfp(new VerilatedFstC);
-    top->trace(tfp.get(), 99);
-    tfp->open("dump.fst");
+    const char* traceFile = "dump.fst";
 #else
-    std::unique_ptr<VerilatedVcdC> tfp(new VerilatedVcdC);
-    top->trace(tfp.get(), 99);
-    tfp->open("dump.vcd");
+    std::unique_ptr<VerilatedvcdC> tfp(new VerilatedVcdC);
+    const char* traceFile = "dump.vcd";
```
ktbarrett, line 89: Any possibility that this can be made into an argument and the FST vs VCD also decided at runtime by looking at the extension?

I'm guessing we could also make `VERILATOR_SIM_DEBUG` a runtime flag, no?

Also I'm wondering if there is a reason not to always compile with `VM_TRACE`, `VM_COVERAGE` and use runtime switches to turn them on. Do they slow down the sim or the build considerably?

toddstrader: Unfortunately the verilated C++ can only be no trace, vcd trace or fst trace.  There's no vcd or fst trace option.

An option to set `VERILATOR_SIM_DEBUG` while building does sound like a good idea.  Seems like a separate commit, but I can add it here is desired.

I have tried benchmarking trace vs no trace models over the years and have never been impressed enough with the performance gains to bother building no trace variants.  But other runner flavors seem to support this construct, so maybe  it's best to keep it?  As I mentioned, we don't plan to use `build()` in the Verilator runner so I don't have a strong preference.

Coverage does slow down the Verilator sims so I would not suggest adding that by default.

ktbarrett: > Unfortunately the verilated C++ can only be no trace, vcd trace or fst trace. There's no vcd or fst trace option.

So Verilator simply doesn't support compiling support for both in and only using the appropriate type? Odd.

ktbarrett (commented):
Are you not going to put `tp->close();` line in a `traceOn` guard?

toddstrader:
> Are you not going to put `tp->close();` line in a `traceOn` guard?

Oh whoops, let me fix that.

ktbarrett:
Also add a newsfragment.

docs/source/newsfragments/3667.feature.rst
```diff
@@ -0,0 +1 @@
+Add `--trace` to Verilator binaries for run-time trace generation and integrate with :ref:`Python Test Runner <howto-python-runner>`
```
ktbarrett, line 1:
```suggestion
Add `--trace` to Verilator binaries for run-time trace generation and integrate with :ref:`Python Test Runner <howto-python-runner>`.
```

ktbarrett (approved):

PR #3668, opened 2024-01-18
Runner log redirection

toddstrader:
Adds an option to send runner build or test logs to a file instead of stdout / stderr.  Currently just squishes the two together, but a separate stderr file option could be added later.  Also added a test to demonstrate.

### DIFF
--- a/docs/source/newsfragments/3668.feature.rst
+++ b/docs/source/newsfragments/3668.feature.rst
@@ -1,0 +1,1 @@
+Add `log_file` argument to :ref:`Python Test Runner <howto-python-runner>` to redirect stdout and stderr to the specified file
--- a/docs/source/newsfragments/3668.feature.rst
+++ b/docs/source/newsfragments/3668.feature.rst
@@ -1,1 +1,1 @@
-Add `log_file` argument to :ref:`Python Test Runner <howto-python-runner>` to redirect stdout and stderr to the specified file
+Add `log_file` argument to :ref:`Python Test Runner <howto-python-runner>` to redirect stdout and stderr to the specified file.
--- a/src/cocotb/runner.py
+++ b/src/cocotb/runner.py
@@ -20,7 +20,7 @@ import tempfile
 import warnings
 from contextlib import suppress
 from pathlib import Path
-from typing import Dict, List, Mapping, Optional, Sequence, Tuple, Type, Union
+from typing import Dict, List, Mapping, Optional, Sequence, TextIO, Tuple, Type, Union
 from xml.etree import ElementTree as ET
 
 import find_libpython
@@ -154,6 +154,7 @@ class Simulator(abc.ABC):
         verbose: bool = False,
         timescale: Optional[Timescale] = None,
         waves: Optional[bool] = None,
+        log_file: Optional[PathLike] = None,
     ) -> None:
         """Build the HDL sources.
 
@@ -172,6 +173,7 @@ class Simulator(abc.ABC):
             verbose: Enable verbose messages.
             timescale: Tuple containing time unit and time precision for simulation.
             waves: Record signal traces.
+            log_file: File to write log
         """
 
         self.clean: bool = clean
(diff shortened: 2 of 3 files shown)


src/cocotb/runner.py
```diff
@@ -363,13 +369,27 @@ def _get_parameter_options(self, parameters: Mapping[str, object]) -> Command:
...
+    def _execute_cmds(
+        self, cmds: Sequence[Command], cwd: PathLike, stdout: Optional[TextIO] = None
+    ) -> None:
+        __tracebackhide__ = True  # Hide the traceback when using PyTest.
+
         for cmd in cmds:
             print(f"INFO: Running command {shlex_join(cmd)} in directory {cwd}")
 
             # TODO: create a thread to handle stderr and log as error?
             # TODO: log forwarding
```
toddstrader, line 387: Leaving this because I assume "log forwarding" means capturing stdout / stderr and yielding that as part of the `build()` or `test()` call, which is not what this PR is doing.

src/cocotb/runner.py
```diff
@@ -363,13 +369,27 @@ def _get_parameter_options(self, parameters: Mapping[str, object]) -> Command:
...
 
-            subprocess.run(cmd, cwd=cwd, env=self.env, check=True)
+            stderr = None if stdout is None else subprocess.PIPE
```
ktbarrett, line 389:
```suggestion
            stderr = None if stdout is None else subprocess.STDOUT
```

toddstrader: Oh, yup.

src/cocotb/runner.py
```diff
@@ -172,6 +173,7 @@ def build(
             verbose: Enable verbose messages.
             timescale: Tuple containing time unit and time precision for simulation.
             waves: Record signal traces.
+            log_file: File to write log
```
cmarqu, line 176:
```suggestion
            log_file: File to write the build log to.
```

docs/source/newsfragments/3668.feature.rst
```diff
@@ -0,0 +1 @@
+Add `log_file` argument to :ref:`Python Test Runner <howto-python-runner>` to redirect stdout and stderr to the specified file
```
cmarqu, line 1:
```suggestion
Add `log_file` argument to :ref:`Python Test Runner <howto-python-runner>` to redirect stdout and stderr to the specified file.
```

src/cocotb/runner.py
```diff
@@ -250,6 +254,7 @@ def test(
                 When running with pytest, the testcase name is prefixed to this name.
             verbose: Enable verbose messages.
             timescale: Tuple containing time unit and time precision for simulation.
+            log_file: File to write log
```
cmarqu, line 257:
```suggestion
            log_file: File to write the test log to.
```

tests/pytest/test_logs.py
```diff
@@ -0,0 +1,66 @@
+import os
```
cmarqu, line 1: Add an SPDX header please.

tests/pytest/test_logs.py
```diff
@@ -0,0 +1,66 @@
...
+    hdl_toplevel,
+    hdl_toplevel_lang,
+    sim_args,
+    sim_build,
+    tests_dir,
+    verilog_sources,
+    vhdl_sources,
+)
+
+sys.path.insert(0, os.path.join(tests_dir, "pytest"))
```
cmarqu, line 23: Would be nicer to stay with `pathlib` instead of `os.path` but that's minor.

toddstrader: This was copypastated from `test_waves.py`.  But I can take a look here.

cmarqu: We are certainly not consistent ourselves yet.

docs/source/newsfragments/3668.feature.rst
toddstrader: I'd argue that most of the news fragments aren't sentences and therefore shouldn't have periods.  But I see that most-to-all do, so I'm happy enough to go along.

ktbarrett (approved):

PR #3669, opened 2024-01-19
Allow runner absolute path results XML

toddstrader:
Allows for `results_xml` to be an absolute path, i.e. something from `tempfile`.

I don't understand what's going on with `PYTEST_CURRENT_TEST`, but I'm guessing this change won't play nice with that.  I could:

- not worry about this
-  assert that `results_xml` is not absolute when `PYTEST_CURRENT_TEST` is set
- disect `results_xml` and add `PYTEST_CURRENT_TEST` at the correct spot

I'm least fond of the last option but can do whatever.


ktbarrett (approved):

src/cocotb/runner.py
```diff
@@ -321,7 +322,10 @@ def test(
             self.current_test_name = "test"
             results_xml_name = results_xml
 
-        results_xml_file = Path(self.test_dir) / results_xml_name
+        if os.path.isabs(results_xml):
```
cmarqu, line 325:
```suggestion
        if Path(results_xml).is_absolute():
```

cmarqu (approved):

toddstrader:
> @toddstrader What do you mean by the issues with `PYTEST_CURRENT_TEST`?

Previously if you were running with `PYTEST_CURRENT_TEST` that env var would get munged into the results.xml file name.  With this change, if you are running with `PYTEST_CURRENT_TEST` and also pass an absolute path to `results_xml` the file name will not have the value of `PYTEST_CURRENT_TEST` munged into it.  I'm unclear what the consequences would be in that scenario.

ktbarrett:
Never even knew that was there. Anyways, IMO `results_xml` should only be set to `PYTEST_CURRENT_TEST` if `results.xml` isn't explicitly set. The way it's done now means even relative paths that include a directory don't work. Feel free to make sweeping changes.

src/cocotb/runner.py
```diff
@@ -247,17 +247,15 @@ def test(
             build_dir: Directory the build step has been run in.
             test_dir: Directory to run the tests in.
             results_xml: Name of xUnit XML file to store test results in.
-                When running with pytest, the testcase name is prefixed to this name.
+                If an absolute path is provided it will be used as-is,
+                :file:`{build_dir}/results.xml` otherwise.
```
ktbarrett, line 251: This link isn't valid.
```suggestion
                ``{build_dir}/results.xml`` otherwise.
```

src/cocotb/runner.py
```diff
@@ -315,17 +313,21 @@ def test(
 
         # When using pytest, use test name as result file name
         pytest_current_test = os.getenv("PYTEST_CURRENT_TEST", "")
```
ktbarrett, line 315: If I were being pedantic I would update this to not provide a default value and update the logic value below to use `None`, which should be a little more precise.

toddstrader: I'm all for pedantry.

ktbarrett (approved):

PR #3673, opened 2024-01-20
Changes to `cocotb.log`

ktbarrett:
Removes the superfluous `SimLog` function and make the `cocotb.log` module private.


cmarqu:
https://github.com/cocotb/cocotb/blob/master/docs/source/rotating_logger.rst needs adaptations as well.
Ideally we would run the code as a test (like in https://github.com/cocotb/cocotb/blob/master/docs/source/quickstart.rst?plain=1#L41-L44) and only include (parts of) the file in the docs.

src/cocotb/scheduler.py
```diff
@@ -103,7 +102,9 @@ def __init__(self):
         self.event = Event()
         self.state = external_state.INIT
         self.cond = threading.Condition()
-        self._log = SimLog(f"cocotb.external.thead.{self.thread}", id(self))
+        self._log = logging.getLogger(
+            f"cocotb.external.thead.{self.thread}.0x{id(self):x}"
```
cmarqu, line 106:
```suggestion
            f"cocotb.external.thread.{self.thread}.0x{id(self):x}"
```

ktbarrett:
Hmmm... okay. I guess we are providing the classes in `cocotb.log` as public interfaces. I guess this should stay public. Though I *really* don't like shadowing the module `cocotb.log` with the variable `cocotb.log`, so I might change the name to `cocotb.logging` instead.

ktbarrett:
Odd...

Apparently, importing an object from a module also does an import of that module. And Python also fixes up relative modules `import cocotb.logging` into relative imports `import .logging` which brings in the name `logging`. So in cocotb/__init__.py when doing `from cocotb.logging import ...` the `logging` name is overwritten which was once the standard library logging module. Very very strange behavior IMO.

cmarqu:
So the least surprising way would be to call it something like `cocolog`?

ktbarrett:
Well it only matters in the implementation of this file and specifically only on the call to `logging.getLogger` in `__init__.py`. It's not an issue for the user.

imphil (approved):

PR #3676, opened 2024-01-24
Have gpi logger read env var

AndrewNolte:
Gpi logging wasn't reading the env var, allowing prints at gpi_embed.cpp:108 and gpi/GpiCommon.cpp:101 to seep through, even if the level was set to warning.

### DIFF
--- a/src/cocotb/share/lib/embed/gpi_embed.cpp
+++ b/src/cocotb/share/lib/embed/gpi_embed.cpp
@@ -37,6 +37,9 @@
 #include <py_gpi_logging.h>  // py_gpi_logger_set_level, py_gpi_logger_initialize, py_gpi_logger_finalize
 
 #include <cassert>
+#include <cstdlib>
+#include <cstring>
+#include <map>
 #include <string>
 
 #include "locale.h"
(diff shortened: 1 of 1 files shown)


src/cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -122,6 +125,22 @@ static void set_program_name_in_venv(void) {
...
+            {"ERROR", GPIError},
+            {"WARNING", GPIWarning},
+            {"INFO", GPIInfo},
+            {"DEBUG", GPIDebug},
+            {"TRACE", GPITrace}
+        };
+        auto it = logStrToLevel.find(log_level);
+        if (it != logStrToLevel.end()) {
+            py_gpi_logger_set_level(it->second);
+        }
```
ktbarrett, line 141: Maybe add an else here with a message stating that the log level wasn't valid.

ktbarrett:
I see you keep having issues with the linting step. you know you can have your environment automatically fix linting errors by installing `pre-commit`?

```sh
pip install pre-commit
pre-commit install
```

Afterwards every time you attempt to commit, it will run the linting and prevent you from making the commit until the errors are fixed, which it also does for you automatically if it can, so you can just add them and re-commit.

src/cocotb/share/lib/embed/gpi_embed.cpp
AndrewNolte: It has a decent message in log.py: 
```  File "/usr/scratch/anolte/local/cocos/cocotb/src/cocotb/log.py", line 104, in default_config
    raise ValueError(
ValueError: Invalid log level 'INF' passed through the COCOTB_LOG_LEVEL environment variable. Valid log levels: CRITICAL, ERROR, WARNING, INFO, DEBUG, TRACE
```
Not sure if we want to duplicate this

ktbarrett (approved):

PR #3682, opened 2024-01-26
Verilator runner test_args

toddstrader:
Pipe `test_args` through to verilated binary during test phase


docs/source/newsfragments/3682.bugfix.rst
```diff
@@ -0,0 +1 @@
+Pipe `test_args` through :ref:`Python Test Runner <howto-python-runner>` for Verilator
```
ktbarrett, line 1:
```suggestion
Pass `test_args` to test command when using :ref:`Python Test Runner <howto-python-runner>` for Verilator, which was previously missing.
```
Good to mention the previous state so people can map that to a potential change in behavior.

cmarqu: Also please use ` ``test_args`` ` in RST to get monospace rendering, single backticks produce italics.

ktbarrett (approved):

PR #3681, opened 2024-01-26
runner trace building for Verilator

toddstrader:
Make the Verilator runner observe the `waves` parameter

PR #3683, opened 2024-01-26
Verilator trace file name parameter

toddstrader:
Not sure about `waves_file` since not all simulators support providing a path for this, nor do they all just write out a single file.  I was going to make Icarus observe `waves_file` too, but then I noticed that I'd need to specify it during the build phase.  So I could:

- Leave it as is
- Assert for non-Verilator simulators
- Drop `waves_file` and just pass what I need via `test_args`

Regardless, the `verilator.cpp` changes are needed and I can make the test work whichever way we want to go.

### DIFF
--- a/docs/source/newsfragments/3683.feature.rst
+++ b/docs/source/newsfragments/3683.feature.rst
@@ -1,0 +1,1 @@
+Add ``--trace-file`` to Verilator binaries which specifies the trace file name.
--- a/src/cocotb/share/lib/verilator/verilator.cpp
+++ b/src/cocotb/share/lib/verilator/verilator.cpp
@@ -48,20 +48,34 @@ static inline bool settle_value_callbacks() {
 
 int main(int argc, char** argv) {
     bool traceOn = false;
+#if VM_TRACE_FST
+    const char* traceFile = "dump.fst";
+#else
+    const char* traceFile = "dump.vcd";
+#endif
 
     for (int i = 1; i < argc; i++) {
         std::string arg = std::string(argv[i]);
         if (arg == "--trace") {
             traceOn = true;
+        } else if (arg == "--trace-file") {
+            if (++i < argc) {
+                traceFile = argv[i];
+            } else {
+                fprintf(stderr, "Error: --trace-file requires a parameter\n");
+                return -1;
+            }
         } else if (arg == "--help") {
             fprintf(stderr,
-                    "usage: %s [--trace]\n"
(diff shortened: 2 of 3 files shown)


toddstrader:
I'll add a newsfragment once we select one of the above options.

toddstrader:
Depends on #3681 (which is why `test_waves.py` is broken).  Not sure about Riviera failures.  I don't think that is related to this PR.

cmarqu:
> * Drop `waves_file` and just pass what I need via `test_args`

I would opt for this; the rabbit hole is too deep.

toddstrader:
> > * Drop `waves_file` and just pass what I need via `test_args`
> 
> I would opt for this; the rabbit hole is too deep.

That works for me.  I'll go that route once I pick this up again unless someone yells loudly.

cmarqu:
@ktbarrett Was that close accidental?:

> Regardless, the `verilator.cpp` changes are needed and I can make the test work whichever way we want to go.

ktbarrett:
Whoops. Missed that part.

toddstrader:
> > > * Drop `waves_file` and just pass what I need via `test_args`
> > 
> > 
> > I would opt for this; the rabbit hole is too deep.
> 
> That works for me. I'll go that route once I pick this up again unless someone yells loudly.

This is done.  Also added a newsfragment and even remembered to add a period this time.

ktbarrett:
[This](https://github.com/cocotb/cocotb/actions/runs/7791145095/job/21246559995?pr=3683#step:4:84) may help.

ktbarrett (approved):

toddstrader:
> [This](https://github.com/cocotb/cocotb/actions/runs/7791145095/job/21246559995?pr=3683#step:4:84) may help.

Good call.  I just added the hook.  However, for some reason I can't get `ruff` to flag the unnecessary `f` when I run it locally.  It's the same `pre-commit` version too.

ktbarrett:
Looks like it's failing.

toddstrader:
> Looks like it's failing.

Yup, this is dependent on #3681 and now also #3682 (since we removed `waves_file`).  As far as I can tell there no way to do stacked PRs on GitHub.  Once those two prerequisites land I'll rebase and this test will pass.

Probably should have marked this one as a draft until those two are merged.

ktbarrett:
Rebased it for you. Tryna close out as much stuff as possible. PRs have been stacking up.

PR #3684, opened 2024-01-26
Allow access to sim handle from python

AndrewNolte:


ktbarrett:
What's the intended use case? What about the cases when there is no implementation object (like pseudo-handles) or multiple handles (possible if we ever support enums or psychical types)?

AndrewNolte:
It's for inspecting the underlying handle, or passing it to other applications. Our use case is for cpp triggers, akin to passing the vpi handle address to Verilator.cpp. I'm not sure how it works in other sims but I'm assuming there's a benefit to accessing the underlying handle there as well. And If it's a pseudo-handle it would just return 0 right?

marlonjames:
No, pseudo regions use the parent handle.
I don't think we want to expose the raw handles cocotb is using.
You should be able to get your own handle to the object using the path and whichever interface to the simulator you are using.

PR #3691, opened 2024-01-31
[gpi] return GPI_REG for ambiguous const types

AndrewNolte:
These types don't actually say anything about the width of the parameter, so let's only return GPI_INTEGER on vpiIntConst instances.

The test_package PR should probably land first, as this may change the expected output there


src/cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -163,12 +163,13 @@ static gpi_objtype_t const_type_to_gpi_objtype(int32_t const_type) {
...
-            return GPI_NET;
+            return GPI_REGISTER;
 #endif
         case vpiDecConst:
         case vpiBinaryConst:
         case vpiOctConst:
         case vpiHexConst:
+            return GPI_REGISTER;
         case vpiIntConst:
             return GPI_INTEGER;
```
marlonjames, line 174:
```suggestion
        case vpiIntConst:
            return GPI_REGISTER;
```

src/cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -157,20 +157,26 @@ static gpi_objtype_t to_gpi_objtype(int32_t vpitype) {
...
+    // Most simulators only return vpiDecConst or vpiBinaryConst
     switch (const_type) {
 #ifdef IUS
         case vpiUndefined:
             LOG_WARN(
                 "VPI: Xcelium reports undefined parameters as vpiUndefined, "
                 "guessing this is a logic vector");
-            return GPI_NET;
+            return GPI_REGISTER;
 #endif
```
AndrewNolte, line 168: I see questa is also failing the test_discovery test that led me to this, should we maybe change the vpi_get(vpiVector) in the GPI_REGISTER code?

AndrewNolte:
^The tests I changed are verilog only, so not sure why these vhdl are failing. Also, they just say "This job failed" instead of the specific test now?

marlonjames:
Can you explain your changes to `VpiSignalObjHdl::initialise()` and why they are needed?

I don't think it is correct to always set `m_num_elems` to `vpiSize`.

IEEE 1800-2017 Section 37.16 Nets
Details
> 21 ) **vpiSize** for an array net shall return the number of nets in the array. For unpacked structures, the size returned
indicates the number of members in the structure. For an enum net, integer net, logic net, time net, packed struct
net, or packed array net, **vpiSize** shall return the size of the net in bits. For a net bit, **vpiSize** shall return 1.

IEEE 1800-2017 Section 37.17 Variables
Details
> 9 ) **vpiSize** for a variable array shall return the number of variables in the array. For variables belonging to an integer
data type (see 6.11), for enum vars, and for packed struct and union variables, **vpiSize** shall return the size of the
variable in bits. For a string var, it shall return the number of characters that the variable currently contains. For
unpacked structures and unions, the size returned indicates the number of fields in the structure or union. For a var
bit, **vpiSize** shall return 1. For all other variables, the behavior of the **vpiSize** property is not defined.


Integer types will now have `m_num_elems` set to the size in bits instead of 1.
Who knows what Xcelium returns for its extension `vpiRealNet` type.

ktbarrett:
I agree. I'm cool with everything, but don't understand the VPI changes. Unless it's something needed to get this working, I'd leave it for another PR. I'm also okay with the `__int__` cast on `LogicArray` where even though that may not be "relevant".

AndrewNolte:
The vpi changes were added to address this issue on iverilog:
```
0.00ns INFO     cocotb.regression                  running recursive_discover (2/27)
vvp error: get 18 not supported by vpiBinaryConst
vvp: vpi_const.cc:375: virtual int __vpiBinaryConst::vpi_get(int): Assertion `0' failed.
```
which is vpi_get(vpiVector).

To fix this I merged in vpiConstant and vpiParameter with the string branch that avoids this call.

Then I realized that the length of these would change now that they're all registers, and it would go from 1 to the length of the variable, even if they were actually ints. So I thought it would be best to stay consistent for int vars as well, but maybe this is better for another discussion/pr. I think it's always better to have more information here. I don't think anyone was checking the "1" value here, and this may uncover more simulator discrepencies, or allow us to detect vpiLongInts.

I'll reduce those changes to just add vpiParameter and vpiConstant to that branch

src/cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -239,14 +239,14 @@ int VpiSignalObjHdl::initialise(const std::string &name,
     } else {
         m_num_elems = vpi_get(vpiSize, GpiObjHdl::get_handle<vpiHandle>());
 
-        if (GpiObjHdl::get_type() == GPI_STRING) {
+        if (vpiConstant == type || vpiParameter == type ||
+            vpiStringVar == type) {
```
ktbarrett, line 243: No need to replace what's there.
```suggestion
        if (GpiObjHdl::get_type() == GPI_STRING || type == vpiConstant ||
            type == vpiParameter) {
```

src/cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -239,14 +239,14 @@ int VpiSignalObjHdl::initialise(const std::string &name,
...
             m_indexable = false;  // Don't want to iterate over indices
             m_range_left = 0;
             m_range_right = m_num_elems - 1;
```
ktbarrett, line 246: This is kind of unfortunate. There's no way to get the bounds on the parameter typespec, so cocotb indexes and simulator indexes will be different. Maybe we can try to get the bounds from the typespec in the future.

ktbarrett (approved):

src/cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -239,14 +239,14 @@ int VpiSignalObjHdl::initialise(const std::string &name,
     } else {
         m_num_elems = vpi_get(vpiSize, GpiObjHdl::get_handle<vpiHandle>());
 
-        if (GpiObjHdl::get_type() == GPI_STRING) {
+        if (GpiObjHdl::get_type() == GPI_STRING || type == vpiConstant ||
```
AndrewNolte, line 242: It would save us a function call if we just compare type, vpiStringVar was the only thing that mapped here other than vpiStringConst, which is covered by the other checks

src/cocotb/share/lib/vpi/VpiCbHdl.cpp
marlonjames: `parameter` objects have `vpiLeftRange` and `vpiRightRange` (which may be null if it has no explicitly defined range)

src/cocotb/share/lib/vpi/VpiCbHdl.cpp
ktbarrett: Abstraction good. It might not be the only thing one day and the compiler should be doing its job and inlining that.

src/cocotb/share/lib/vpi/VpiCbHdl.cpp
ktbarrett: I'm fine with getting this in now and filing an issue for a later fix of ranges.

marlonjames (approved):

marlonjames:
Changes look good, i'll merge after we figure out the CI problem.

PR #3704, opened 2024-02-06
[Verilog] Interface Array Testcase

AndrewNolte:
This should be a HierarchyArrayObject, not a NonHierarchyIndexableObject. I'm not sure how we'd like to approach this. I would think it should be a cast to GPI_GENARRAY. I'm not sure what this is, there's no comments in gpi.h about it except for mentioning pseudo handles with it. Do pseudo handles (GPI_GENARRAY) mean that the handle doesn't exist in the hierarchy that we query from? Or does it just mean arrays that are in the structure?

I can't map vpiInstanceArray / vpiInterfaceArray to GPI_GENARRAY because it needs vpi_iterate with vpiRange. Indexing works however with this approach. Should I make a range iterator, and use that depending on the vpitype? Or try and get it to map these special types of GPI_ARRAY to HierarhcyObjectArray?

### DIFF
--- a/tests/test_cases/test_sv_interface/test_sv_if.py
+++ b/tests/test_cases/test_sv_interface/test_sv_if.py
@@ -13,3 +13,22 @@ async def test_sv_if(dut):
     assert hasattr(dut.sv_if_i, "a")
     assert hasattr(dut.sv_if_i, "b")
     assert hasattr(dut.sv_if_i, "c")
+
+
+@cocotb.test(expect_fail=True)
+async def test_sv_if_arrays(dut):
+    """Test that interface arrays are the correct type and iterable"""
+
+    # should be a HierarchyArrayObject, currently is a NonHierarchyIndexableObject in Xcelium
+    print(dut.sv_if_arr)
+
+    assert isinstance(dut.sv_if_arr, cocotb.handle.HierarchyArrayObject)
+
+    dut.sv_if_arr._discover_all()
+
+    m = len(dut.sv_if_arr)
+    assert m == 3
+    for i in range(m):
+        assert hasattr(dut.sv_if_arr[i], "a")
+        assert hasattr(dut.sv_if_arr[i], "b")
+        assert hasattr(dut.sv_if_arr[i], "c")
--- a/tests/test_cases/test_sv_interface/test_sv_if.py
+++ b/tests/test_cases/test_sv_interface/test_sv_if.py
@@ -7,7 +7,7 @@ import cocotb
 
 @cocotb.test()
 async def test_sv_if(dut):
-    """Test that signals in an interface are discovered and iterable"""
+    """Test that signals in an interface are discovered"""
 
     dut.sv_if_i._discover_all()
     assert hasattr(dut.sv_if_i, "a")
@@ -17,7 +17,7 @@ async def test_sv_if(dut):
 
 @cocotb.test()
 async def test_sv_intf_arr_type(dut):
-    """Test that interface arrays are the correct type and iterable"""
+    """Test that interface arrays are the correct type"""
 
     print(dut.sv_if_arr)
 
@@ -30,11 +30,13 @@ async def test_sv_intf_arr_type(dut):
 
 @cocotb.test()
 async def test_sv_intf_arr_len(dut):
+    """Test that interface array length is correct"""
     assert len(dut.sv_if_arr) == 3
 
 
 @cocotb.test()
 async def test_sv_intf_arr_access(dut):
+    """Test that interface array objects can be accessed"""
     for i in range(3):
         assert hasattr(dut.sv_if_arr[i], "a")
         assert hasattr(dut.sv_if_arr[i], "b")
@@ -43,6 +45,7 @@ async def test_sv_intf_arr_access(dut):
 
 @cocotb.test()
 async def test_sv_intf_arr_iteration(dut):
+    """Test that interface arrays can be iterated"""
     count = 0
     for intf in dut.sv_if_arr:
         assert hasattr(intf, "a")
--- a/tests/test_cases/test_sv_interface/test_sv_if.py
+++ b/tests/test_cases/test_sv_interface/test_sv_if.py
@@ -36,7 +36,7 @@ async def test_sv_intf_arr_len(dut):
     assert len(dut.sv_if_arr) == 3
 
 
-@cocotb.test()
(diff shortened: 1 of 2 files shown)


tests/test_cases/test_sv_interface/test_sv_if.py
```diff
@@ -13,3 +13,22 @@ async def test_sv_if(dut):
     assert hasattr(dut.sv_if_i, "a")
     assert hasattr(dut.sv_if_i, "b")
     assert hasattr(dut.sv_if_i, "c")
+
+
+@cocotb.test(expect_fail=True)
```
ktbarrett, line 18: Also expect `AttributeError` on verilator since it can't even find the handle.

AndrewNolte:
I still don't really understand what a pseudo-region is, I think gpi.h needs a lot more clarification. My perception is it's the same as hierarchy arrays or hierarchy scopes, and sometimes the handle may be missing? However in the vpiimpl nothing in to_gpi_objtype returns GPI_GEN_ARRAY, so we only have code that treats these handles as if they're not there. Or if they are there, for convenience uses the GPI_ARRAY to avoid iterating, instead doing size/index/range discovery

ktbarrett:
> I think gpi.h needs a lot more clarification

You are absolutely correct there. I'm not certain myself. But it is an implementation detail, not necessarily an API concern like gpi.h. The entire C++ codebase needs a major facelift, but it's not public API, so it's lower on my own radar than Python rn. Sweeping C++ changes are a 3.0 issue, but until then we just need something that works, even if it's ugly as sin.

marlonjames:
See [the wiki](https://github.com/cocotb/cocotb/wiki/cocotb-Internals#pseudo-regions) for some details on pseudo-regions.

marlonjames:
For VPI, there are actual differences between an instance array and gen scope array.

We need a way to distinguish hierarchy arrays like `vpiInterfaceArray` from things like `vpiRegArray`, rather than wade into the mess of `GPI_GENARRAY` and pseudo-handles.

ktbarrett:
Is the difference in terms of implementation or capability? Still seems like it's an array of scopes, which is kind of what GPI_GENARRAY is.

marlonjames:
Conceptually `GPI_GENARRAY` is as you say, an array of scopes. But GPI code uses it only for pseudo-regions.

In fact, `vpiGenScopeArray` handles are thrown away in favor of pseudo-region objects when looking up gen scope array by name:

https://github.com/cocotb/cocotb/blob/fdf7286a3a38b703dc4a3bb7c707bfcd2263773a/src/cocotb/share/lib/vpi/VpiImpl.cpp#L367-L380

We don't iterate over `vpiGenScopeArray`, so during iteration only `vpiGenScope` shows up.

When using VPI, gen scope and gen scope array objects are not instance objects like modules, packages, and interfaces are, and need to be handled differently internally.

So I think for instance arrays, we either need to differentiate `GPI_ARRAY` between hierarchy (instance arrays) and non-hierarchy (var array, net array), or add another `GPI_*` type.

Pseudo-region implementation needs an update, and maybe at that time we could merge `GPI_GENARRAY` into whatever type we use for other hierarchy arrays.

ktbarrett:
Rerunning CI.

ktbarrett:
Since #3824 landed first, you'll need to update the test expectations. Use positive logic, not `not verilator`.

ktbarrett:
Riviera failures.
```
# COUT:   3000.00ns INFO     cocotb.regression                  running test_sv_if.test_sv_intf_arr_access (4/5)
# COUT:                                                             Test that interface array objects can be accessed
# COUT:   3000.00ns ERROR    gpi                                Invalid Index - Index 1 is not in the range of [0:0]
# COUT:   3000.00ns WARNING  gpi                                Failed to find a handle at index 1 via any registered implementation
# COUT:   4000.00ns INFO     cocotb.regression                  test_sv_if.test_sv_intf_arr_access failed
# COUT:                                                         Traceback (most recent call last):
# COUT:                                                           File "/opt/actions-runner/_work/cocotb/cocotb/tests/test_cases/test_sv_interface/test_sv_if.py", line 43, in test_sv_intf_arr_access
# COUT:                                                             assert hasattr(dut.sv_if_arr[i], "a")
# COUT:                                                           File "/opt/actions-runner/_work/cocotb/cocotb/.nox/dev_test_sim-sim-riviera-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.8/site-packages/cocotb/handle.py", line 772, in __getitem__
# COUT:                                                             raise IndexError(f"{self._path} contains no object at index {index}")
# COUT:                                                         IndexError: top.sv_if_arr contains no object at index 1
# COUT:   4000.00ns INFO     cocotb.regression                  running test_sv_if.test_sv_intf_arr_iteration (5/5)
# COUT:                                                             Test that interface arrays can be iterated
# COUT:   5000.00ns INFO     cocotb.regression                  test_sv_if.test_sv_intf_arr_iteration failed
# COUT:                                                         Traceback (most recent call last):
# COUT:                                                           File "/opt/actions-runner/_work/cocotb/cocotb/tests/test_cases/test_sv_interface/test_sv_if.py", line 58, in test_sv_intf_arr_iteration
# COUT:                                                             assert count == 3
# COUT:                                                         AssertionError: assert 1 == 3
```

tests/test_cases/test_sv_interface/test_sv_if.py
```diff
@@ -7,9 +7,50 @@
...
+        assert isinstance(dut.sv_if_arr, cocotb.handle.HierarchyArrayObject)
+
+
+@cocotb.test(expect_fail=cocotb.SIM_NAME.lower().startswith("riviera"))
+async def test_sv_intf_arr_len(dut):
+    """Test that interface array length is correct"""
+    assert len(dut.sv_if_arr) == 3
+
+
+@cocotb.test(expect_fail=cocotb.SIM_NAME.lower().startswith("riviera"))
```
ktbarrett, line 37: `expect_error=IndexError if Riviera else ()`

AndrewNolte: Ohh does expect_fail only cover asserts and not any general fail?

ktbarrett: That's correct. I'm glad you didn't know that, it's not a great system, and one we hope to replace, so users not touching them is a good thing.

PR #3717, opened 2024-02-16
Regression manager changes

### LINKED ISSUE #3689: Improve regression messages
The output for tests starting, being skipped, etc. could be better.

- [ ] Skipped tests currently don't print a docstring =(
- [ ] Only the first line (the 'brief') of the docstring should be printed to keep things short.

ktbarrett:
Following the discussion in #3578, how tests are registered with the RegressionManager was changed again. I think this solution solves everyone's issues. `cocotb.parameterize` was improved and `TestFactory` was deprecated. Closes #3689.


src/cocotb/decorators.py
```diff
@@ -298,21 +303,31 @@ async def my_test_1_b() -> None:
             arg1, arg2 = 1, "b"
             ...
 
+    Options can also be specified in much the same way that :meth:`TestFactory.add_option <cocotb.regression.TestFactory.add_option>` can
+    by supplying tuples of the parameter name to values, or a sequence of variable names and a sequence of values
+
+    .. code-block:: python 3
```
marlonjames, line 309:
```suggestion
    .. code-block:: python3
```

src/cocotb/regression.py
```diff
@@ -90,20 +91,63 @@
         assert False, "pytest.raises doesn't raise an exception when it fails"
 
 
-class _Test:
+class Test:
```
marlonjames, line 94: This should be in the docs. `register_test()` references it.

src/cocotb/regression.py
```diff
@@ -165,14 +210,10 @@ def __init__(self) -> None:
...
         self.xunit = XUnitReporter(filename=results_filename)
-
         self.xunit.add_testsuite(name=suite_name, package=package_name)
-
         self.xunit.add_property(name="random_seed", value=str(cocotb.RANDOM_SEED))
 
-    def discover_tests(
-        self, modules: Iterable[str], filters: Optional[Sequence[str]] = None
-    ) -> None:
+    def discover_tests(self, modules: Iterable[str]) -> None:
```
marlonjames, line 216: Move filter-related items out of docstring and add to `filter_tests()` below.

src/cocotb/regression.py
```diff
@@ -189,105 +230,49 @@ def discover_tests(
...
-        name: Optional[str] = None,
-        module: Optional[str] = None,
-        timeout_time: Optional[float] = None,
-        timeout_unit: str = "step",
-        expect_fail: bool = False,
-        expect_error: Union[Type[Exception], Sequence[Type[Exception]]] = (),
-        skip: bool = False,
-        stage: int = 0,
-    ) -> None:
+    def register_test(self, test: Test) -> None:
```
marlonjames, line 269: May want to add description of `test` arg for docs, right now description is empty. 
https://cocotb--3717.org.readthedocs.build/en/3717/library_reference.html#cocotb.regression.RegressionManager.register_test

src/cocotb/regression.py
```diff
@@ -134,28 +176,39 @@ async def func(*args, **kwargs):
         self.stage = stage
         self.name = self.func.__qualname__ if name is None else name
         self.module = self.func.__module__ if module is None else module
+        self.doc = self.func.__doc__ if doc is None else doc
+        if self.doc is not None:
+            self.doc = inspect.cleandoc(self.doc)
         self.fullname = f"{self.module}.{self.name}"
-        self.included: Optional[bool] = None
```
marlonjames, line 138: `Test.included` was used to explicitly run a test decorated with `skip=True` by including it in `TESTCASE`.

`test_skipped` is failing now:
https://github.com/cocotb/cocotb/actions/runs/7936272852/job/21671195158?pr=3717#step:23:3660

src/cocotb/regression.py
```diff
@@ -90,20 +89,63 @@
...
+        func:
+            The test function.
+
+        name:
+            The name of the test function.
+            Defaults to ``func.__qualname__`` if not given.
+
+        module:
+            The name of the module containing the test function.
+            Defaults to ``func.__module__`` if not given.
```
cmarqu, line 105:
```suggestion
            Defaults to ``func.__module__`` (the name of the module containing the test function) if not given.
```

src/cocotb/regression.py
```diff
@@ -90,20 +89,63 @@
...
+
+        doc:
+            The docstring for the test.
+            Defaults to ``func.__doc__`` if not given.
```
cmarqu, line 109:
```suggestion
            Defaults to ``func.__doc__`` (the docstring of the test function) if not given.
```

src/cocotb/regression.py
```diff
@@ -90,20 +89,63 @@
...
+        timeout_unit:
+            Units of ``timeout_time``, accepts any units that :class:`~cocotb.triggers.Timer` does.
+
+        expect_fail:
+            If ``True`` and the test fails a functional check via an ``assert`` statement, :class:`pytest.raises`,
+            :class:`pytest.warns`, or :class:`pytest.deprecated_call`, the test is considered to have passed.
+            If ``True`` and the test passes successfully, the test is considered to have failed.
+
+        expect_error:
+            Mark the result as a pass only if one of the exception types is raised in the test.
```
cmarqu, line 123:
```suggestion
            Mark the result as a pass only if one of the given exception types is raised in the test.
```

src/cocotb/regression.py
```diff
@@ -165,132 +218,81 @@ def __init__(self) -> None:
...
         """
         for module_name in modules:
             self.log.debug("Searching for tests in module %s", module_name)
             self.log.debug("Python Path: %s", ",".join(sys.path))
             self.log.debug("PWD: %s", os.getcwd())
-            __import__(module_name)
+            mod = import_module(module_name)
+
+            if not hasattr(mod, "__cocotb_tests__"):
+                raise RuntimeError(f"No tests were discovered in module: {module_name}")
```
cmarqu, line 242:
```suggestion
                raise RuntimeError(f"No tests were discovered in module: {module_name!r}")
```

src/cocotb/regression.py
```diff
@@ -165,132 +218,81 @@ def __init__(self) -> None:
...
+        if not self._test_queue:
             modules_str = ", ".join(repr(m) for m in modules)
-            raise RuntimeError(f"No tests were discovered in modules: {modules_str}")
+            raise RuntimeError(f"No tests were discovered in any module: {modules_str}")
 
-        # if no filters, accept all tests
-        if not filters:
-            return
+    def filter_tests(self, filters: Sequence[str]) -> None:
+        """Filters discovered tests.
```
cmarqu, line 253:
```suggestion
        """Filter discovered tests.
```

src/cocotb/regression.py
```diff
@@ -165,132 +218,81 @@ def __init__(self) -> None:
...
 
-        # else mark all tests as excluded
-        self.ntests = 0
-        for test in self._queue:
-            test.included = False
+        Only those tests which match at least one of the given filters are included;
```
cmarqu, line 255:
```suggestion

        Only those tests which match at least one of the given filters are included;
```

cmarqu (commented):
(I'm only about halfway through)

src/cocotb/regression.py
ktbarrett: `No tests were discovered in module: example` vs `No tests were discovered in module: "example"`? I'm guessing this would also apply to the similar errors below?

src/cocotb/regression.py
ktbarrett: I'm not sure adding an additional newline here is required by PEP257.

src/cocotb/regression.py
cmarqu: The error below for the multiple modules already does that, right, via the `repr()` in the `join`, or am I wrong?

src/cocotb/regression.py
ktbarrett: There is already a blank line there.

src/cocotb/regression.py
ktbarrett: Addressed in the most recent commit (which is a kinda large refactor of RegressionManager). When `TESTCASE` is used, the `TESTCASE` mode is set on the Regression Manager, which causes skipped tests to run.

src/cocotb/regression.py
```diff
@@ -350,72 +441,35 @@ def _tear_down(self) -> None:
...
 
-    def _next_test(self) -> Optional[_Test]:
-        """Get the next test to run"""
-        if not self._queue:
-            return None
-        return self._queue.pop(0)
-
     def _handle_result(self, test: Task) -> None:
-        """Handle a test completing.
+        """Callback given the the scheduler for when the current test completes.
```
marlonjames, line 445:
```suggestion
        """Callback given to the scheduler, to be called when the current test completes.
```

src/cocotb/regression.py
```diff
@@ -525,21 +608,26 @@ def _record_test_excluded(self, test: _Test) -> None:
         )
         self.xunit.add_skipped()
 
-    def _record_test_skipped(self, test: _Test) -> None:
+        # do not log anything, nor save details for the summary
+
+    def _record_test_skipped(self, test: Test) -> None:
+        """Called by :meth:`_execute` when a test initialization fails."""
```
marlonjames, line 614:
```suggestion
        """Called by :meth:`_execute` when a test is skipped."""
```

src/cocotb/regression.py
```diff
@@ -494,26 +564,39 @@ def _log_test_passed(
...
+            start_hilight,
+            stop_hilight,
+            rest,
             exc_info=result,
         )
 
-    def _record_test_excluded(self, test: _Test) -> None:
-        lineno = self._get_lineno(test)
+    def _record_test_excluded(self, test: Test) -> None:
+        """Called by :meth:`_execute` when a test initialization fails."""
```
marlonjames, line 596:
```suggestion
        """Called by :meth:`_execute` when a test is excluded by filters."""
```

src/cocotb/regression.py
```diff
@@ -755,6 +876,7 @@ def _log_test_summary(self) -> None:
 
     @staticmethod
     def _safe_divide(a: float, b: float) -> float:
+        """Used when computing time ratios to ensure no exception is raised is either time is 0."""
```
marlonjames, line 879:
```suggestion
        """Used when computing time ratios to ensure no exception is raised if either time is 0."""
```

docs/source/library_reference.rst
```diff
@@ -272,6 +272,10 @@ The Regression Manager
 
 .. autodata:: cocotb.regression_manager
 
+.. autoclass:: Test
+
+.. autoclass:: RegressionMode
```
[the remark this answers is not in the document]
ktbarrett: I'm surprised that we aren't documenting a single enum until now.

tests/test_cases/test_cocotb/test_testfactory.py
```diff
@@ -39,11 +29,20 @@ async def test_testfactory_verify_args(dut):
...
+
+
+@cocotb.test
+async def test_testfactory_verify_names(dut):
     assert testfactory_test_names == {
-        f"{prefix}run_testfactory_test{postfix}_{i:03}" for i in range(1, 5)
+        "run_testfactory_test[arg1='a1v1',arg2='a2v1',arg3='a3v1']",
+        "run_testfactory_test[arg1='a1v1',arg2='a2v2',arg3='a3v2']",
+        "run_testfactory_test[arg1='a1v2',arg2='a2v1',arg3='a3v1']",
+        "run_testfactory_test[arg1='a1v2',arg2='a2v2',arg3='a3v2']",
```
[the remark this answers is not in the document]
ktbarrett: Right... And what about strings that have a comma in them? Any ideas?

src/cocotb/regression.py
```diff
@@ -134,28 +170,71 @@ async def func(*args, **kwargs):
...
+        self._test_start_time: float
+        self._test_start_sim_time: float
         self.log = _logger
         self._regression_start_time: float
-        self.test_results = []
+        self._test_results: List[Dict[str, Any]] = []
         self.ntests = 0
+        """Total number of tests that will be run or skipped."""
         self.count = 0
+        """The current test count."""
```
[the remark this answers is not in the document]
ktbarrett: Any suggestions? I'm thinking `total_tests` and idk about the other. `count` is fine for me, so suggest away.

tests/test_cases/test_cocotb/test_testfactory.py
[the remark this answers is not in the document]
ktbarrett: Or really long strings... I might move this commit to another PR so we can hash this out.

src/cocotb/regression.py
```diff
@@ -90,24 +91,61 @@
...
+class Test:
+    """A cocotb test in a regression.
+
+    Args:
+        func:
+            The test function.
+
+        name:
+            The name of the test function.
+            Defaults to ``func.__qualname__`` (the dotted path to the test function in the module) if not given.
```
cmarqu, line 103: Actually, with "Defaults to", the "if not given" is redundant, right?

src/cocotb/regression.py
```diff
@@ -165,132 +250,73 @@ def __init__(self) -> None:
...
-                :class:`pytest.warns`, or :class:`pytest.deprecated_call`, the test is considered to have passed.
-                If ``True`` and the test passes successfully, the test is considered to have failed.
+    def set_mode(self, mode: RegressionMode) -> None:
+        """Sets the regression mode.
 
-            expect_error:
-                Mark the result as a pass only if one of the exception types is raised in the test.
-                This is primarily for cocotb internal regression use for when a simulator error is expected.
+        See :class:`RegressionMode` for more details on how each mode affect Regression Manager behavior.
+        """
```
marlonjames, line 303: For completeness:
```suggestion
        See :class:`RegressionMode` for more details on how each mode affect Regression Manager behavior.
        
        Should be called before :meth:`start_regression` is called.
        """
```

src/cocotb/regression.py
```diff
@@ -90,24 +91,61 @@
...
 
 
-class _Test:
-    _id_count = 0  # used by the RegressionManager to sort tests in definition order
+class Test:
+    """A cocotb test in a regression.
+
+    Args:
+        func:
+            The test function.
```
cmarqu, line 99:
```suggestion
            The test function object.
```
(maybe)

src/cocotb/regression.py
```diff
@@ -90,24 +91,61 @@
...
+
+        expect_error:
+            Mark the result as a pass only if one of the given exception types is raised in the test.
+
+        skip:
+            Don't execute this test as part of the regression.
+            The test can still be run manually by setting :make:var:`TESTCASE`.
+
+        stage:
+            Order tests logically into stages, where multiple tests can share a stage.
```
cmarqu, line 132: I have never intuitively gotten what this does TBH...

src/cocotb/regression.py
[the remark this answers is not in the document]
marlonjames: `count` is okay, I can't think of anything better.

src/cocotb/regression.py
```diff
@@ -134,28 +172,75 @@ async def func(*args, **kwargs):
...
+    Tests are added using :meth:`register_test` or :meth:`discover_tests`.
+    Filters for tests can be added using :meth:`add_filters`.
+    The "mode" of the regression can be controlled using :meth:`set_mode`.
+    These methods can be called in any order any number of times.
+
+    Once all the tests, filters, and regression behavior configuration is done,
+    the user starts the regression with :meth:`start_regression`.
+    This method must be called exactly once.
+
+    Until the regression is started, :attr:`ntests`, :attr:`count`, :attr:`passed`,
```
cmarqu, line 217:
```suggestion
    Until the regression is started, :attr:`total_tests`, :attr:`count`, :attr:`passed`,
```

src/cocotb/regression.py
```diff
@@ -165,132 +250,73 @@ def __init__(self) -> None:
...
+        for filter in filters:
+            compiled_filter = re.compile(filter)
+            self._filters.append(compiled_filter)
 
-            expect_fail:
-                If ``True`` and the test fails a functional check via an ``assert`` statement, :class:`pytest.raises`,
-                :class:`pytest.warns`, or :class:`pytest.deprecated_call`, the test is considered to have passed.
-                If ``True`` and the test passes successfully, the test is considered to have failed.
+    def set_mode(self, mode: RegressionMode) -> None:
+        """Sets the regression mode.
```
cmarqu, line 300:
```suggestion
        """Set the regression mode.
```

src/cocotb/regression.py
```diff
@@ -165,132 +250,73 @@ def __init__(self) -> None:
...
-            stage=stage,
-        )
         self.log.debug("Registered test %r", test.fullname)
-        self._queue.append(test)
+        self._test_queue.append(test)
 
     @classmethod
-    def setup_pytest_assertion_rewriting(cls, test_modules: Iterable[str]) -> None:
+    def setup_pytest_assertion_rewriting(cls) -> None:
+        """Configures pytest to rewrite assertions for better failure messages.
```
cmarqu, line 316:
```suggestion
        """Configure pytest to rewrite assertions for better failure messages.
```

src/cocotb/regression.py
```diff
@@ -315,29 +341,100 @@ def setup_pytest_assertion_rewriting(cls, test_modules: Iterable[str]) -> None:
...
+            )
+
+        # start test loop
         self._regression_start_time = time.time()
         self._execute()
 
+    def _execute(self, *, sim_failed: bool = False) -> None:
+        """Used by both :meth:`start_regression` and :meth:`_test_complete` to continue the main test running loop.
+
+        If ``sim_failed`` is ``True``, when a test would otherwise be run (respecting inclusion, skip, etc.),
```
cmarqu, line 375:
```suggestion
        If *sim_failed* is ``True``, when a test would otherwise be run (respecting inclusion, skip, etc.),
```

src/cocotb/regression.py
```diff
@@ -350,72 +447,31 @@ def _tear_down(self) -> None:
...
-
-        Args:
-            test: The test that completed
+        Due to the way that simulation failure is handled,
+        this function must be able to detect simulation failure and finalize the regression.
         """
-        assert test is self._test_task
 
+        # compute test completion time
         real_time = time.time() - self._test_start_time
```
cmarqu, line 458: Maybe rename this to `wall_time` while we are here?

src/cocotb/regression.py
```diff
@@ -551,19 +641,83 @@ def _record_test_skipped(self, test: _Test) -> None:
...
-        test: _Test,
+        test: Test,
         outcome: Outcome,
         wall_time_s: float,
         sim_time_ns: float,
-    ) -> None:
+    ) -> bool:
+        """Called by :meth:`_test_complete` to score and record the result of a finished test.
+
+        Because of how simulation failure is handled, this function must detect and return whether the simulator failed.
```
cmarqu, line 710:
```suggestion
        Because of how simulation failure is handled, this function must detect and return whether the simulation failed.
```

src/cocotb/utils.py
```diff
@@ -285,3 +286,35 @@ def extract_coro_stack(coro, limit=None):
...
+
+
+Self = TypeVar("Self")
+
+
+class DocEnum(Enum):
+    """Like :class:`enum.Enum`, but allows documenting enum values.
+
+    Documentation for enum members can be optionally added by setting enum values to a tuple of the intended value and the docstring.
+    This adds the provided docstring to the `__doc__` field of the enum value.
```
cmarqu, line 298:
```suggestion
    This adds the provided docstring to the ``__doc__`` field of the enum value.
```

src/cocotb/regression.py
```diff
@@ -134,28 +172,75 @@ async def func(*args, **kwargs):
...
+    )
 
 
 class RegressionManager:
-    """Encapsulates all regression capability into a single place"""
+    """Object which manages tests.
+
+    This object uses the builder pattern to build up a regression.
+    Tests are added using :meth:`register_test` or :meth:`discover_tests`.
+    Filters for tests can be added using :meth:`add_filters`.
```
cmarqu, line 209:
```suggestion
    Inclusion filters for tests can be added using :meth:`add_filters`.
```
maybe? IMO we almost can't be explicit enough with this word...

src/cocotb/regression.py
```diff
@@ -165,132 +250,73 @@ def __init__(self) -> None:
...
-                If ``True`` and the test fails a functional check via an ``assert`` statement, :class:`pytest.raises`,
-                :class:`pytest.warns`, or :class:`pytest.deprecated_call`, the test is considered to have passed.
-                If ``True`` and the test passes successfully, the test is considered to have failed.
+    def set_mode(self, mode: RegressionMode) -> None:
+        """Sets the regression mode.
 
-            expect_error:
-                Mark the result as a pass only if one of the exception types is raised in the test.
-                This is primarily for cocotb internal regression use for when a simulator error is expected.
+        See :class:`RegressionMode` for more details on how each mode affect Regression Manager behavior.
```
cmarqu, line 302:
```suggestion
        See :class:`RegressionMode` for more details on how each mode affects `class:`RegressionManager` behavior.
```

src/cocotb/regression.py
```diff
@@ -165,132 +250,73 @@ def __init__(self) -> None:
...
+        """
+        self._mode = mode
 
-            skip:
-                Don't execute this test as part of the regression.
-                The test can still be run manually by setting :make:var:`TESTCASE`.
+    def register_test(self, test: Test) -> None:
+        """Register a test with the RegressionManager.
```
cmarqu, line 307:
```suggestion
        """Register a test with the :class:`RegressionManager`.
```

src/cocotb/regression.py
ktbarrett: IMO it should be removed... but maybe someone is using it and we can't really provide an alternative outside of "lexically reorder your tests in the file", which is poopy. 

It's there to mark that some tests should run before/after other tests based on the stage number. So all stage=0 tests run, then all stage=1, then all stage=2, etc.

I'm fine with taking the opportunity to deprecate this without alternative, but will try to come up with a better docstring.

marlonjames (approved):

marlonjames:
Pinging @AndrewNolte.

src/cocotb/regression.py
```diff
@@ -90,24 +91,62 @@
...
+class Test:
+    """A cocotb test in a regression.
+
+    Args:
+        func:
+            The test function object.
+
+        name:
+            The name of the test function.
+            Defaults to ``func.__qualname__`` (the dotted path to the test function in the module)
```
cmarqu, line 103:
```suggestion
            Defaults to ``func.__qualname__`` (the dotted path to the test function in the module).
```

src/cocotb/regression.py
```diff
@@ -90,24 +91,62 @@
...
+
+        timeout_time:
+            Simulation time duration before timeout occurs.
+
+        timeout_unit:
+            Units of ``timeout_time``, accepts any units that :class:`~cocotb.triggers.Timer` does.
+
+        expect_fail:
+            If ``True`` and the test fails a functional check via an ``assert`` statement, :class:`pytest.raises`,
+            :class:`pytest.warns`, or :class:`pytest.deprecated_call`, the test is considered to have passed.
```
cmarqu, line 121:
```suggestion
            If ``True`` and the test fails a functional check via an ``assert`` statement, :pytest:class:`pytest.raises`,
            :pytest:class:`pytest.warns`, or :pytest:class:`pytest.deprecated_call`, the test is considered to have passed.
```
doesn't link otherwise, hopefully this does.

src/cocotb/regression.py
```diff
@@ -90,24 +91,62 @@
...
+            If ``True`` and the test passes successfully, the test is considered to have failed.
+
+        expect_error:
+            Mark the result as a pass only if one of the given exception types is raised in the test.
+
+        skip:
+            Don't execute this test as part of the regression.
+            The test can still be run manually by setting :make:var:`TESTCASE`.
```
cmarqu, line 129:
```suggestion
            The test can still be run manually by setting :envvar:`TESTCASE`.
```
(doesn't link otherwise, we have another such occurrence of `:make:var:` in the release notes.)

src/cocotb/regression.py
```diff
@@ -90,24 +91,62 @@
...
+        module:
+            The name of the module containing the test function.
+            Defaults to ``func.__module__`` (the name of the module containing the test function).
+
+        doc:
+            The docstring for the test.
+            Defaults to ``func.__doc__`` (the docstring of the test function).
+
+        timeout_time:
+            Simulation time duration before timeout occurs.
```
cmarqu, line 114:
```suggestion
            Simulation time duration before the timeout exception :exc:`~cocotb.result. SimTimeoutError` occurs.
```

cmarqu:
It would be nice to have a section or page in the user docs showing the different use cases.

src/cocotb/regression.py
```diff
@@ -165,132 +252,75 @@ def __init__(self) -> None:
...
 
-            skip:
-                Don't execute this test as part of the regression.
-                The test can still be run manually by setting :make:var:`TESTCASE`.
+    def register_test(self, test: Test) -> None:
+        """Register a test with the :class:`RegressionManager`.
 
-            stage:
-                Order tests logically into stages, where multiple tests can share a stage.
+        Should be called before :meth:`start_regression` is called.
```
cmarqu, line 313:
```suggestion
        Should be called before :meth:`start_regression` is called.
        
        Args:
            test: The test object to register.
```
Looks better in the generated docs, and may be helpful in some IDE.

src/cocotb/regression.py
```diff
@@ -165,132 +252,75 @@ def __init__(self) -> None:
...
-            expect_fail:
-                If ``True`` and the test fails a functional check via an ``assert`` statement, :class:`pytest.raises`,
-                :class:`pytest.warns`, or :class:`pytest.deprecated_call`, the test is considered to have passed.
-                If ``True`` and the test passes successfully, the test is considered to have failed.
+        See :class:`RegressionMode` for more details on how each mode affects :class:`RegressionManager` behavior.
 
-            expect_error:
-                Mark the result as a pass only if one of the exception types is raised in the test.
-                This is primarily for cocotb internal regression use for when a simulator error is expected.
+        Should be called before :meth:`start_regression` is called.
```
cmarqu, line 306:
```suggestion
        Should be called before :meth:`start_regression` is called.
        
        Args:
            mode: The regression mode to set.
```
Looks better in the generated docs, and may be helpful in some IDE.

AndrewNolte:
Do you mind writing down which issue's were specifically fixed? I believe I had two main ones in that dicussion, one being that an exact match TESTCASE would now add all tests that have that as a prefix, and the other being that imported tests were automatically added

src/cocotb/decorators.py
```diff
@@ -208,22 +209,24 @@ async def my_test(dut):
         The test function to which the decorator is applied.
     """
 
+    def _add_tests(*tests: Test) -> None:
+        glbs = inspect.currentframe().f_back.f_back.f_globals
```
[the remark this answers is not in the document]
ktbarrett: Yeah I think this would work. I kinda wish `__module__` was a handle to the module object and not just the name, but whatever...

ktbarrett:
@AndrewNolte

1. Importing from a test module registers all the tests in that module.

Fixed. And you can even import test implementations from other modules and use them to implement tests in the current module, without also registering them in the current module.

2. Importing a test from another module to add that test to the current module.

Not changed. Won't fix. Use `MODULE` to create multiple regressions that run different sets of tests.

3. TESTCASE now includes all tests that use the given value as a prefix

Not changed. Won't fix. Use `make TESTCASE=my_test\$` to match the test name exactly. Running all tests that match a `TestFactory`'s test name is a desirable change. Seeing as `TESTCASE` is explicitly set by the user to override a regression, there should be no silent change in behavior.

AndrewNolte:
Thanks! #1 was the main one I was concerned about, and #3 was more for concerns with regression use cases where the tests suites are split up and run separately. I suppose runner.py might need to be changed if that's still being maintained.

src/cocotb/regression.py
ktbarrett: That implies it can be caught, but it can't, so I reworded it slightly.

cmarqu (approved):

PR #3723, opened 2024-02-19
Fix docstring links to SimTimeoutError

cmarqu:


cmarqu:
Links to the exception now work: https://cocotb--3723.org.readthedocs.build/en/3723/triggers.html#cocotb.triggers.with_timeout

ktbarrett (approved):

PR #3727, opened 2024-02-20
[Vpi] Split VpiCbHdl.cpp

AndrewNolte:
I'm hoping to refactor a bunch of this code soon to better support scope arrays and packed structs, so I wanted to split up some of the code before then


ktbarrett:
I like this change, but can we do it while preserving git history? https://stackoverflow.com/questions/3887736/keep-git-history-when-splitting-a-file

src/cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -317,7 +317,7 @@ GpiObjHdl *VpiImpl::native_check_create(const std::string &name,
 
     new_hdl = vpi_handle_by_name(const_cast<char *>(fq_name.c_str()), NULL);
 
-#ifdef ICARUS
+#if defined(ICARUS) || defined(VERILATOR)
```
marlonjames, line 320: If the scope of this PR is just the split, this should not be included. Move this to a follow-on that addresses any changes for generate scope support.

AndrewNolte: oops yeah this wasn't supposed to go in, was experimenting with https://github.com/verilator/verilator/pull/4913

ktbarrett:
Superseded by #3737.

PR #3729, opened 2024-02-21
Add more integer parameters to test_package

marlonjames:
Alternative to #3665 with additional integer type parameters.
All should return `LogicObject` handles after #3691 was merged.


ktbarrett:
Pinging @AndrewNolte.

ktbarrett (approved):

PR #3733, opened 2024-02-23
More handle cleanup

### LINKED ISSUE #3548: [Tracking] cleanup handle.py
- [x] ConstantObject does not reveal with the underlying value type is (#2720)
- [x] NonConstantObject type has no purpose (#2720)
- [x] ModifiableObject is both the class for logic arrays and every other value type causing unnecessary functionality to be inherited (#2720)
- [x] NonHierarchyArrayObject is both the class for non-hierarchy arrays and a base class for ModifiableObject and all more specific typed objects causing unnecessary functionality to be inherited (#2720)
- [x] Unclear which types are abstract and which aren't (#2720)
- [x] string casts on handles is the path unless it's a value handle, then it's the string cast of the value (#2720)
- [x] equality is handle equality unless it's a value handle, then it's the equality of the value of the handle and this breaks the hashability, whoops (#2720)
- [x] unnecessary specification of `__ne__` (#2720)
- [x] unnecessary access of handle attributes on handle creation (#2720)
- [ ] unclear what is public method and what isn't
- [x] `__getattr__` syntax conflicts with attribute/method access (introduces `__getitem__` #3655)
- [x] `__iter__` on HierarchyObject and HierarchyArrayObject return child handles instead of keys, like you might expect from a pseudo-dictionary/list type (introduces `_keys`, `_values`, and `_items` #3655)
- [x] `__len__` isn't defined on HierarchyObject (#3655)
- [x] `_id` takes an "extended" argument instead of just having the user add backslashes themselves like they might expect, and extended=True by default (deprecated and replaced with `__getitem__` #3655)
- [x] `_id` is a terrible name (deprecated and replaced with `__getitem__` #3655)
- [x] `_range` isn't available on HierarchyArrayObject (#3655)
- [ ] `_range` is a crappy interface, prefer `left` and `right`
- [ ] `.value` property cannot be properly typed since type systems assume getter and setter of property get and set using the same type
- [x] clean up dead code (#2720)
- [x] clean up unnecessarily inherited attributes/methods (#2720)
- [x] tons of unnecessary logs (#2720)
- [ ] fix/improve documentation

### LINKED ISSUE #3604: Integrate `Range` and `Array` types into the handle classes
Handle classes like `HierarchyArrayObject` and `NonHierarchyIndexableObjectBase` which expose information about the array's ranges could introduce properties like `range`, `left`, and `right` which leverage `cocotb.types.Range`. And `NonHierarchyIndexableObjectBase.value` could return `Array` objects to give array objects with consistent indexing back.

### LINKED ISSUE #3678: Change `ModifiableObject.drivers` and `ModifiableObject.loads` to be Python iterables
Currently they just return `cocotb.simulator.gpi_hdl_iterator` which is not a standard Python iterator, but has it's own interface and returns raw handles instead of `SimHandleBase`s.

### LINKED ISSUE #3728: Add a `simulator.is_indexable` call
Right now, `simulator.get_range` is designed to return `None` if the type isn't indexable. This means that you can't get the range of an object that isn't indexable, but does have a range (like strings and logic arrays in some simulators).

ktbarrett:
This should be the final PR. Closes #3548.

This PR:
* renames the value handles
* removes what was `ModifiableObject` and the non-functional `drivers` and `loads` methods (closes #3678)
* adds the `set` method to set values in a type-safe way
* improves typing and documentation


ktbarrett:
So currently the issue is with `LogicObject` and how it works with multi-dimensional packed arrays and logic scalars.

# Issues with scalars

When a `logic`, `reg`, etc. in Verilog is seen, or a `std_logic` in VHDL, a `LogicObject` is used. But such an object is not indexable, doesn't have a range, and isn't iterable. However, `LogicObject` provides all that functionality, so it must be designed to work in those situations.

Currently there is an `_is_indexable` property we are checking, which should be False for scalars, which causes indexing to raise, iteration to return nothing, and range to be 1-length and whatever the defaults are (currently it's [-1, -1] from the GPI base classes).

# Issues with multi-dimensional packed arrays

When a multi-dimensional packed logic array is seen in Verilog, we can't assume we can index it, as simulators can assume that it's an indivisible object. This means that it also isn't indexable. But `LogicObject` lets you do both. In these cases `LogicObject` assumes indexing will fail, and so there is some logic in `__iter__` to convert that failure into empty iteration.

Another issue is that the length of the object is the number of total bits in the object, so a `logic[2:0][2:0]` has a length of 9. This makes sense because when getting/setting the value, you do so with the whole object at a time. But when getting the range, we only get the outer-most dimension of the range: `[2:0]`. So `len(handle) != len(handle.range)`. This means we can't set the range when getting the value appropriately.

# Solutions

IMO, we should separate scalars, single-dimension packed arrays (and `std_logic_vector` in VHDL), and multi-dimension packed arrays into separate types. ***None*** of them should be iterable or indexable.

Scalars would not have a range and getting/setting the value would use `Logic` instead of a `LogicArray`.

Single-dimension packed arrays would have a range and getting the value would return a `LogicArray` with the appropriate range set. Setting the value would take any SizedIterable of the same length (as if it were converted into `LogicArray`) or an `int`.

Multi-dimensional packed arrays would ideally return some composite range object. Getting the value would return an `Array[LogicArray]` or a custom `MultiDimensionalLogicArray` that could handle multiple-levels of indexing and also maybe setting the whole value at once. Setting would take either a flattened SizedIterable of the same length as the full object, SizedIterables of SizedIterables with the correct dimension and size (as if it were converted into `MultiDimensionalLogicArray`), or an `int`. This will require changes the GPI to return the ranges of packed objects, which seems unlikely to be supported.

Failing the above answer, multi-dimensional packed arrays could return a range of 0 to length-1, which I feel would be valid because flattened values don't necessarily have a valid range, so it can be whatever. Getting and setting the value would work like single-dimension packed arrays at that point.

The single-dimension and simple multi-dimension solution could be combined if we can ensure that at the GPI level, the range of packed objects is listed as 0 to length-1 rather than the range of the first dimension.

ktbarrett:
I decided to split this PR up. The previous comment no longer applies to this PR.

src/cocotb/handle.py
```diff
@@ -308,14 +314,14 @@ def __len__(self) -> int:
...
+    def __dir__(self) -> Iterable[str]:
         """Permits IPython tab completion and debuggers to work."""
         self._discover_all()
-        return super().__dir__() + [str(k) for k in self._keys()]
+        return set(super().__dir__()) | {str(k) for k in self._keys()}
 
 
 class HierarchyObject(HierarchyObjectBase[str]):
-    r"""Named hierarchical scope objects.
+    r"""A simulation object that is an name-indexed collection of hierarchical simulation objects.
```
cmarqu, line 324:
```suggestion
    r"""A simulation object that is a name-indexed collection of hierarchical simulation objects.
```

src/cocotb/handle.py
```diff
@@ -337,6 +343,9 @@ class HierarchyObject(HierarchyObjectBase[str]):
     Any possible name of an object is supported with the index syntax,
     but it can be more verbose.
 
+    Accessing a non-existant child with attribute syntax results in an :class:`AttributeError`,
+    and accessing a non-existant child with index syntax results in a :class:`KeyError`.
```
cmarqu, line 347:
```suggestion
    Accessing a non-existent child with attribute syntax results in an :class:`AttributeError`,
    and accessing a non-existent child with index syntax results in a :class:`KeyError`.
```

src/cocotb/handle.py
```diff
@@ -430,6 +448,8 @@ class HierarchyArrayObject(HierarchyObjectBase[int]):
         block_0 = dut.gen_pipe_stages[0]
         block_7 = dut.gen_pipe_stages[7]
 
+    Accessing a non-existant child results in an :class:`IndexError`.
```
cmarqu, line 451:
```suggestion
    Accessing a non-existent child results in an :class:`IndexError`.
```

src/cocotb/handle.py
```diff
@@ -520,69 +540,218 @@ def __iter__(self) -> Iterable[SimHandleBase]:
...
+
+class ValueObjectBase(SimHandleBase, Generic[ValuePropertyT, ValueSetT]):
+    """Base class for all simulation objects that have a value."""
 
     @property
-    def value(self):
-        """The value of this simulation object.
+    @abstractmethod
+    def value(self) -> ValuePropertyT:
+        """Gets or sets the value of the simulation object.
```
cmarqu, line 615:
```suggestion
        """Get or set the value of the simulation object.
```

src/cocotb/handle.py
```diff
@@ -520,69 +540,218 @@ def __iter__(self) -> Iterable[SimHandleBase]:
...
+    @abstractmethod
+    def _set_value(
+        self,
+        value: ValueSetT,
+        action: _GPISetAction,
+        schedule_write: Callable[
+            ["ValueObjectBase[Any, Any]", Callable[..., None], Sequence[Any]], None
+        ],
+    ) -> None:
+        """Schedules a write of the given value to a simulator object.
```
cmarqu, line 699:
```suggestion
        """Schedule a write of the given value to a simulator object.
```

src/cocotb/handle.py
```diff
@@ -520,69 +540,218 @@ def __iter__(self) -> Iterable[SimHandleBase]:
...
+    """Base class for all simulation object types that have a range and can be indexed.
+
+    These objects can be iterated over to yield child objects:
+
+    .. code-block:: python3
+
+        for child in dut.array_object:
+            print(child._path)
+
+    A particular child can be retrieved using it's index:
```
cmarqu, line 728:
```suggestion
    A particular child can be retrieved using its index:
```

src/cocotb/handle.py
```diff
@@ -939,67 +1137,130 @@ def _set_value(self, value, call_sim):
...
+class StringObject(
+    IndexableValueObjectBase[bytes, bytes, IntegerObject],
+    ValueObjectBase[bytes, bytes],
+):
+    """A string simulation object.
 
-        .. versionchanged:: 1.4
-            Takes :class:`bytes` instead of :class:`str`.
-            Users are now expected to choose an encoding when using these objects.
+    This type is used when an ``string`` (VHDL or Verilog) simulation object is seen.
```
cmarqu, line 1176:
```suggestion
    This type is used when a ``string`` (VHDL or Verilog) simulation object is seen.
```

tests/test_cases/test_compare/test_compare.py
```diff
@@ -54,18 +54,18 @@ async def test_compare_simhandlebase(dut):
 
 
 @cocotb.test()
-async def test_compare_nonhierarchy(dut):
-    """Test for NonHierarchyObject comparisons"""
+async def test_compare_valueobject(dut):
+    """Test for ValueObjectBase comparisons"""
```
cmarqu, line 58:
```suggestion
    """Test for ValueObjectBase comparisons."""
```

cmarqu (commented):
A first look, will go over it again later.

ktbarrett:
@cmarqu Take your time. This module needs *lots* of attention wrt documentation.

src/cocotb/handle.py
```diff
@@ -520,69 +540,218 @@ def __iter__(self) -> Iterable[SimHandleBase]:
...
+    FORCE = 1
+    RELEASE = 2
+
+
+#: The type of the value a :class:`Deposit` or :class:`Force` action contains.
+ValueT = TypeVar("ValueT")
+
+
+class Deposit(Generic[ValueT]):
+    """Action used for placing a value into a given handle.
```
cmarqu, line 554:
```suggestion
    """Action used for placing a value into a given handle. This is the default action.
```

src/cocotb/handle.py
```diff
@@ -622,46 +791,77 @@ def __len__(self) -> int:
...
+    """A simulation object that is an array of value-having simulation objects.
 
-    +--------------+---------------------+--------------------------------------------------------------+
-    | Verilog      | VHDL                | ``arr.value`` is equivalent to                               |
-    +==============+=====================+==============================================================+
-    | ``arr[4:7]`` | ``arr(4 to 7)``     | ``[arr[4].value, arr[5].value, arr[6].value, arr[7].value]`` |
-    +--------------+---------------------+--------------------------------------------------------------+
-    | ``arr[7:4]`` | ``arr(7 downto 4)`` | ``[arr[7].value, arr[6].value, arr[5].value, arr[4].value]`` |
-    +--------------+---------------------+--------------------------------------------------------------+
+    This object is used whenever an array, that isn't a logic array or string, is seen.
```
cmarqu, line 804:
```suggestion
    This object is used whenever an array is seen that isn't either a logic array or string.
```

src/cocotb/handle.py
```diff
@@ -520,69 +540,218 @@ def __iter__(self) -> Iterable[SimHandleBase]:
...
+        """Schedule a write of the given value to a simulator object.
+
+        Conversion from multiple Python types into a type understood by the simulator is expected.
+        This is used to implement the :attr:`value` property setter, :meth:`setimmediatevalue`, and :meth:`set`.
 
-        ``call_sim(handle, f, *args)`` should be used to schedule simulator writes,
-        rather than performing them directly as ``f(*args)``.
+        Args:
+            value: A value used to set the handle.
+            action: Whether to deposit, force, or release the value on the handle.
```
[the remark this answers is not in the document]
ktbarrett: This action can only be deposit, freeze, or release, like `_GpiSetAction`.

src/cocotb/handle.py
```diff
@@ -520,69 +540,218 @@ def __iter__(self) -> Iterable[SimHandleBase]:
...
+    :class:`Deposit`\ s, or drives from HDL processes do not cause the value to change until the handle is :class:`Release`\ d.
+    Further :class:`Force`\ s will overwrite the value.
+    """
+
+    def __init__(self, value: ValueT) -> None:
+        self.value = value
+
+
+class Freeze:
+    """Action used to make a handle keep its current value until a :class:`Release` is applied."""
```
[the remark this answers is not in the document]
ktbarrett: Please recheck what I put there. I made it more explicit.

src/cocotb/handle.py
```diff
@@ -520,69 +540,218 @@ def __iter__(self) -> Iterable[SimHandleBase]:
...
+
+    def __init__(self, value: ValueT) -> None:
+        self.value = value
+
+
+class Force(Generic[ValueT]):
+    """Action used to force a handle to a given value until a :class:`Release` is applied.
+
+    :class:`Deposit`\ s, or drives from HDL processes do not cause the value to change until the handle is :class:`Release`\ d.
+    Further :class:`Force`\ s will overwrite the value.
```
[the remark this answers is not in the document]
ktbarrett: Please recheck.

src/cocotb/handle.py
```diff
@@ -520,69 +540,226 @@ def __iter__(self) -> Iterable[SimHandleBase]:
...
+    FORCE = 1
+    RELEASE = 2
+
+
+#: The type of the value a :class:`Deposit` or :class:`Force` action contains.
+ValueT = TypeVar("ValueT")
+
+
+class Deposit(Generic[ValueT]):
+    """Action used for placing a value into a given handle.. This is the default action.
```
cmarqu, line 554:
```suggestion
    """Action used for placing a value into a given handle. This is the default action.
```

src/cocotb/handle.py
```diff
@@ -520,69 +540,226 @@ def __iter__(self) -> Iterable[SimHandleBase]:
...
+        self.value = value
+
+
+class Force(Generic[ValueT]):
+    """Action used to force a handle to a given value until a :class:`Release` is applied.
+
+    :class:`Deposit` writes from cocotb or drives from HDL processes
+    do not cause the value to change until the handle is :class:`Release`\ d.
+    Further :class:`Force`\ s will overwrite the value and leave the value frozen.
+    :class:`Freeze`\ s will  act as a no-op.
```
cmarqu, line 572:
```suggestion
    :class:`Freeze`\ s will act as a no-op.
```

src/cocotb/handle.py
```diff
@@ -939,67 +1145,130 @@ def _set_value(self, value, call_sim):
...
+            Assigns a :class:`bytes` value at the end of the current delta cycle.
+            When the value's length is less than the simulation object's,
+            the value is padded with NUL (``'\0'``) characters up to the appropriate length.
+            When the value's length is greater than the simulation object's,
+            the value is truncated without a NUL terminator to the appropriate length,
+            without warning.
+
+        Strings in both Verilog and VHDL are byte arrays without any particular encoding.
+        Encoding must be done to turn Python strings into byte arrays.
+        Because there are many encodings, this step is left up to the user.
```
cmarqu, line 1224: Maybe link to https://docs.python.org/3/library/codecs.html#standard-encodings ?

cmarqu:
The inheritance diagram at https://cocotb--3733.org.readthedocs.build/en/3733/library_reference.html#simulation-object-handles is now outdated.

It could be regenerated by locally reverting the change in https://github.com/cocotb/cocotb/pull/3457/files and overwriting the SVG with the image gotten from a browser.
Or maybe Sphinx has been fixed in the meantime?

ktbarrett:
@cmarqu It's outputting SVG, but not displaying the image, just the code. Any ideas?

cmarqu:
> @cmarqu It's outputting SVG, but not displaying the image, just the code. Any ideas?

Works for me. Do you have a file like `cocotb/.nox/.cache/docs_out/_images/inheritance-788f556e7491e4b60b5d6114abe9f8ac3c8d40f5.svg`?
If not, is it really outputting SVG, not graphviz/dot syntax (if so, you are probably missing the `dot` tool)?

Anyway, the output looks a bit "wilder" than  before, maybe we should draw it by hand anyway...

ktbarrett:
Yeah, it definitely needs to be done by hand. There is ABC and Generic and a few other base classes we really don't need to show. Ideally though, we wouldn't document the inheritance hierarchy and instead document only each concrete class. I don't expect many users understand inheritance at any level.

ktbarrett:
I'm going to leave the diagram unfinished for now, as there will be another follow on commit that changes the hierarchy (bye bye IndexableValueObjectBase).

cmarqu:
> Ideally though, we wouldn't document the inheritance hierarchy and instead document only each concrete class. I don't expect many users understand inheritance at any level.

I'm fine with just removing the diagram.

ktbarrett:
I think as long as we are documenting inheritance, it's valuable. But if we don't, obviously it isn't.

ktbarrett:
Considering I'm putting off the diagram update until the next PR, can I get a final review @cocotb/maintainers?

src/cocotb/handle.py
```diff
@@ -1104,11 +1285,10 @@ def SimHandle(
     """Factory function to create the correct type of `SimHandle` object.
```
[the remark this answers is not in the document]
cmarqu: Maybe because `SimHandleBase` is both a return type annotation and a reference in the "Returns:" part of the docstring?
It's not duplicated in the old version, https://docs.cocotb.org/en/latest/library_reference.html#cocotb.handle.SimHandle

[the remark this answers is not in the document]
ktbarrett: I ran into this in other cases as well, like on the `value` getters and setters. It's a bug in Sphinx that has something to do with the cache AFAICT. I rebuild fixed the errors that I saw locally.

marlonjames (approved):

src/cocotb/handle.py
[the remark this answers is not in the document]
marlonjames: Looks like still an issue but I'm fine with merging and seeing if it is fixed on master.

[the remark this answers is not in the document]
ktbarrett: Yeah there's something up with the cache. I fixed the missing `"` character from your last comment, but it's still missing from the documentation. Maybe master will fix itself. I wonder if there's a way to clear or not use the cache that we can set in `.readthedocs.yml`

PR #3734, opened 2024-02-23
Updates to CI

ktbarrett:
One update gets the lint and regression to run when the pre-commit is autoupdated and the other pins the version of NVC used for benchmarking so we get more consistent results.


marlonjames:
While you are pinning NVC, update `generate-envs.py` from r1.11.0 to r1.11.3

marlonjames (approved):

PR #3737, opened 2024-02-24
Split VpiCbHdl.cpp (with history)

ktbarrett:
Alternative to #3727. I left the file structure the same as in that commit. This PR would have to be "merged" in (as opposed to "rebase" or "squash") to preserve history or manually merged and pushed by using `git rebase --rebase-merges`. If we didn't care to do that, we should just go with #3727.


marlonjames (approved):
I prefer keeping the history and having a few localized merge commits.
Let's see what the others think.

cmarqu (approved):

ktbarrett:
@imphil Need your assistance. Can't merge this PR because master is protected against "merge"s.

imphil:
>@imphil Need your assistance. Can't merge this PR because master is protected against "merge"s.

I removed the protection until this PR is in. Happy merging!

imphil (approved):
I would have been fine either way, but it's nice that you gone through the trouble of doing it this way, @ktbarrett!

PR #3740, opened 2024-02-25
Docs: Update Verilator support section

### LINKED ISSUE #3194: [tracking] Get the cocotb test suite passing with Verilator

marlonjames:
Closes #3194
Add note suggesting to try the newest release if there users encounter issues.
Note the `--timing` option for delayed assignment support.

### DIFF
--- a/docs/source/simulator_support.rst
+++ b/docs/source/simulator_support.rst
@@ -26,6 +26,7 @@ In order to use this simulator, set :make:var:`SIM` to ``icarus``:
     make SIM=icarus
 
 .. note::
+
     A working installation of `Icarus Verilog <http://iverilog.icarus.com/>`_ is required.
     You can find installation instructions `in the Icarus Verilog Installation Guide <https://iverilog.fandom.com/wiki/Installation_Guide>`_.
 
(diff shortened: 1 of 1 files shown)


ktbarrett (approved):

docs/source/simulator_support.rst
```diff
@@ -71,11 +72,11 @@ Verilator
 
 .. warning::
 
-    Verilator is in the process of adding more functionality to its VPI interface, which is used by cocotb to access the design.
-    Therefore, Verilator support in cocotb is currently experimental.
-    Some features of cocotb may not work correctly or at all.
+    **cocotb supports Verilator 5.006 and later.**
```
imphil, line 75: Change that to the latest Verilator release we're actually testing, i.e., 5.022 (we're currently still using a git commit before this release in CI, but should update to this release now)

imphil (approved):

docs/source/simulator_support.rst
marlonjames: #3739 switched to the new release.

The question is what do we consider the minimum supported version. 5.006 still works, but doesn't have all of the VPI improvements from later releases.

imphil: Users want guidance on what version they should use, and what they can expect from using that version. If we say we "support" 5.006 but then users just run into problems that we know have been fixed, then we're not providing a good service to those users.

Since Verilator is super easy to update, I find it friendlier to say "cocotb supports Verilator 5.022 and later". The huge majority of users will then just get that version and be happy. And users who really cannot use that version might either reach out to us, or give earlier versions a try and use it if it works for them.

We don't even have tests running for 5.006!

docs/source/simulator_support.rst
```diff
@@ -71,11 +72,11 @@ Verilator
...
 
-    **cocotb only supports Verilator 5.006 and later.**
+    Verilator is in the process of adding more functionality to its VPI interface, which is used by cocotb to access the design.
+    Therefore, Verilator support in cocotb is currently experimental, and some features of cocotb may not work correctly or at all.
+    If you encounter an issue using Verilator with cocotb, please try the newest release and check existing issues before opening a new one.
```
cmarqu, line 79:
```suggestion
    If you encounter an issue using Verilator with cocotb, please try the newest release
    and [check existing issues](https://github.com/cocotb/cocotb/issues?q=is%3Aissue+-label%3Astatus%3Aduplicate+label%3Acategory%3Asimulators%3Averilator) before opening a new one.
```

cmarqu (approved):

PR #3748, opened 2024-02-29
Better parameterize names

### LINKED ISSUE #2373: parameterizing tests and passing arguments via the command line
There was some discussion during the hackathon regarding the need to parameterize a test and feed argument from the command line. @alexforencich brought this up, and I have needed a feature like this in the past as well. 

### Requirements
* Must support multiple tests per simulator run, like we do currently
* Should be able to pass arguments to particular tests
* Should be able to parameterize multiple tests with the same parameter
* Should have support as a make variable in the current flow

### Plusargs
Currently we have plusargs, which are also fed to the simulator. There is a limit to what can be passed in a plusargs, so this might not be the best answer. And officially they are only supported by SystemVerilog compatible simulators. GHDL, for instance, consumes them as simulator arguments. But they are supported in the make flow currently.

### Environment Variables
There is also ad-hoc argument passing via environment variables. This does work, but could be argued that it's tedium on the user that could be handled automatically. It's up to the user to decide how to handle each variable, and environment variables can be set in the current makefile flow.

### Pytest
OOTB pytest does not support this easily. According to [this](https://stackoverflow.com/questions/40880259/how-to-pass-arguments-in-pytest-by-command-line) answer it requires adding argument parser in conftest.py and then more code in the test to handle those arguments. This isn't really possible without extra work, due to inversion of control. And it requires the user to do most of the work because of the need to modify pytest's argument parser. The arguments would also need to be fed to the call to the main pytest invocation by way of environment variable. This ends up not being a very clean solution.

ktbarrett:
Follow on from #3717. Changes how `cocotb.parameterize` names tests to make them more descriptive and easier to match with `TESTCASE`. Closes #2373.

### DIFF
--- a/docs/source/newsfragments/3717.feature.1.rst
+++ b/docs/source/newsfragments/3717.feature.1.rst
@@ -1,0 +1,1 @@
+Use parameter values in generated test names of :func:`cocotb.parameterize`, which should be clearer and allow the user to better group tests using :envvar:`TESTCASE`.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,7 @@ ignore = [
     "PLR0915", # Too many statements (>50) (preference)
     "PLR2004", # Magic value used in comparison (preference)
     "PLW0603", # Using the global statement (preference)
+    "PLR0911", # Too many return statements (preference)
 ]
 
 [tool.ruff.lint.per-file-ignores]
--- a/src/cocotb/decorators.py
+++ b/src/cocotb/decorators.py
@@ -27,21 +27,28 @@
 
 import functools
 import sys
+from enum import Enum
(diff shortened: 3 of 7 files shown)


cmarqu:
I suppose the coverage holes could be filled relatively easily, maybe even using doctests?

ktbarrett:
I could also do it with a unit test. I'll work on that.

ktbarrett:
Should have 100% coverage now.

cmarqu (approved):

ktbarrett:
Well it has 100% coverage on the reprs code, not necessarily the input checking code...

ktbarrett:
Okay... it works now.

cmarqu (approved):

PR #3756, opened 2024-03-05
Vpi Structure Test and Vlt fixes

AndrewNolte:
Generate scopes, interfaces arrays, etc. are accessible on verilator after https://github.com/verilator/verilator/pull/4913 lands.  In order to make them accessible by name as well as iteration, it needs to do something similar to what icarus does with iterating through the children when querying by name


AndrewNolte:
@ktbarrett I'd like to have the first test just check the output for simplicity. I think this should be done in python instead of make, what do you think? Maybe like a test() wrapper with args for sim -> output file

tests/test_cases/test_vpi_structure/Makefile
```diff
@@ -0,0 +1,51 @@
...
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
+
+
+
+
```
cmarqu, line 30:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause
```
Newly added files should get the SPDX header.

tests/test_cases/test_vpi_structure/vpi_structure.py
```diff
@@ -0,0 +1,90 @@
...
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL POTENTIAL VENTURES LTD BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
cmarqu, line 26:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause
```

tests/test_cases/test_vpi_structure/vpi_structure.sv
cmarqu, line 1: Not sure if we can have a file with a different license in here.
I'd probably just rewrite it having just the interesting elements,
or ask Wilson if he could relicense this file (public domain?).

tests/test_cases/test_vpi_structure/vpi_structure.py
```diff
@@ -0,0 +1,90 @@
...
+@cocotb.test()
+async def test_name_matches_iter(dut):
+    """
+    Tests that
+        - all of the names accessible through iteration are also accesisble through the name,
+        - multiple instances in python don't corrupt cpp handle lengths, particularly pseudo objects.
+    """
+
+    t = cocotb.handle.HierarchyObject(dut._handle, dut._path)
+    assert id(t) != id(dut)
```
ktbarrett, line 75: ???

tests/test_cases/test_vpi_structure/vpi_structure.py
```diff
@@ -0,0 +1,90 @@
...
+    Not all of the match at the moment, but we can use this to measure progress.
+    """
+
+    # TODO store and check the outputs
+
+    for obj, depth in iter_module(dut):
+        objlen = get_len(obj)
+        objtype = type(obj).__qualname__
+        treeinfo = f"{'  ' * depth}{obj._name}: {objtype}[{objlen}]"
+        print(f"{treeinfo:50} {obj._path}")
```
ktbarrett, line 63:
```suggestion
        cocotb.log.info(f"{treeinfo:50} {obj._path}")
```

tests/test_cases/test_vpi_structure/Makefile
```diff
@@ -0,0 +1,51 @@
...
+TOPLEVEL_LANG ?= verilog
+
+TOPLEVEL := t
+
+PWD=$(shell pwd)
+
+# SIM_OUTPUT_REDIRECT:="asdf.out"
+
+COCOTB?=$(PWD)/../../..
```
cmarqu, line 40:
```suggestion
PWD = $(shell pwd)

# SIM_OUTPUT_REDIRECT := "asdf.out"

COCOTB ?= $(PWD)/../../..
```

ktbarrett:
Looks like Icarus isn't going to work. discovery between all the simulators is so inconsistent that I don't know if the kind of test you are trying to do here is reasonable.

tests/test_cases/test_vpi_structure/vpi_structure.py
```diff
@@ -0,0 +1,90 @@
...
+def get_len(obj):
+    return len(obj) if hasattr(obj, "__len__") else None
+
+
+@cocotb.test()
+async def test_structure(dut):
+    """
+    Tests that name, fullname, handle type, and length match accross simulators.
+    Not all of the match at the moment, but we can use this to measure progress.
+    """
```
cmarqu, line 55:
```suggestion
    """
    Tests that name, fullname, handle type, and length match across simulators.
    
    Not all of them match at the moment, but we can use this to measure progress.
    """
```

tests/test_cases/test_vpi_structure/vpi_structure.py
```diff
@@ -0,0 +1,90 @@
...
+        print(f"{treeinfo:50} {obj._path}")
+
+
+@cocotb.test()
+async def test_name_matches_iter(dut):
+    """
+    Tests that
+        - all of the names accessible through iteration are also accesisble through the name,
+        - multiple instances in python don't corrupt cpp handle lengths, particularly pseudo objects.
+    """
```
cmarqu, line 72:
```suggestion
    """
    Test name accessibility and handle lengths.
    
     All of the names accessible through iteration are also accessible through the name.
     Multiple instances in Python don't corrupt C++ handle lengths, particularly pseudo objects.
    """
```

tests/test_cases/test_vpi_structure/vpi_structure.sv
cmarqu: Also, if this file is originally from Verilator,
we should have a link to it for updates if needed.

marlonjames:
As a general note, please provide more descriptive commit messages.
Explain what the commit is doing, and if it is not obvious, provide a reason _why_ the change is necessary.
Also break PRs into multiple commits when doing a series of changes, so that it is easier to follow the steps taken.

src/cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -317,7 +317,7 @@ GpiObjHdl *VpiImpl::native_check_create(const std::string &name,
 
     new_hdl = vpi_handle_by_name(const_cast<char *>(fq_name.c_str()), NULL);
 
-#ifdef ICARUS
+#if defined(ICARUS) || defined(VERILATOR)
     /* Icarus does not support vpiGenScopeArray, only vpiGenScope.
```
marlonjames, line 321:
```suggestion
    /* Icarus and Verilator do not support vpiGenScopeArray, only vpiGenScope.
```

tests/test_cases/test_vpi_structure/vpi_structure.py
AndrewNolte: This ensures we're not using the already populated _sub_handles from iteration when testing name vs iteration. I noticed the ids when doing SimHandle() were actually the same bc of the logic in _handle2obj. In the future we might change this to cache __new__ instead.

tests/test_cases/test_vpi_structure/vpi_structure.py
AndrewNolte: This will add quite a bit of clutter to the output file, and I might be changing this to tee to a buffer to check against a file

tests/test_cases/test_vpi_structure/vpi_structure.py
ktbarrett: Add a comment then.

tests/test_cases/test_vpi_structure/vpi_structure.py
ktbarrett: How about instead of printing it out, we maintain the expected values in a JSON file and load into the test?

tests/test_cases/test_vpi_structure/vpi_structure.py
```diff
@@ -0,0 +1,69 @@
...
+        print(f"{treeinfo:50} {obj._path}")
+
+
+@cocotb.test()
+async def test_name_matches_iter(dut):
+    """
+    Test name accessibility and handle lengths.
+    
+     All of the names accessible through iteration are also accessible through the name.
+     Multiple instances in Python don't corrupt C++ handle lengths, particularly pseudo objects.
```
cmarqu, line 50:
```suggestion
    All of the names accessible through iteration are also accessible through the name.
    Multiple instances in Python don't corrupt C++ handle lengths, particularly pseudo objects.
```

src/cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -344,7 +344,8 @@ GpiObjHdl *VpiImpl::native_check_create(const std::string &name,
         }
 
         for (auto rgn = vpi_scan(iter); rgn != NULL; rgn = vpi_scan(iter)) {
-            if (vpi_get(vpiType, rgn) == vpiGenScope) {
+            auto rgn_type = vpi_get(vpiType, rgn);
+            if (rgn_type == vpiGenScope || rgn_type == vpiModule) {
```
AndrewNolte, line 348: This actually enabled Icarus to pick up Interface arrays by name now.  Before it would fail on test_name_matches_iter

src/cocotb/handle.py
```diff
@@ -212,7 +212,7 @@ def _items(self) -> Iterable[Tuple[KeyType, SimHandleBase]]:
         self._discover_all()
         return self._sub_handles.items()
 
-    @lru_cache(maxsize=None)
```
AndrewNolte, line 215: lru_cache keys on repr(obj), whereas cached_property attaches it to the actual object. This probably shouldn't come up in a real testbench though since everything goes through SimHandle and therefore the handle cache. But were this to change and we allowed deletion of the python handles, this would cause some issues like it did for this test

tests/test_cases/test_vpi_structure/vpi_structure.sv
AndrewNolte: Doesn't the license say you can redistribute it and/or modify it?

AndrewNolte:
> Looks like Icarus isn't going to work. discovery between all the simulators is so inconsistent that I don't know if the kind of test you are trying to do here is reasonable.

Yeah I see it more as a progress measure rather than trying to get them all to agree on one output file

AndrewNolte:
Are all of those icarus versions old? It works fine for me, and just has those parameters missing. 

Also, the verilator version will not pass until https://github.com/verilator/verilator/pull/4913 lands and the cocotb verilator version is updated. I'm worried the VpiImpl change will need to be landed in lockstep with the verilator vpi change, or have some version guard on that ifdef. So maybe the test should land first?

src/cocotb/handle.py
ktbarrett: The difference is ones a callable and the other is a property... you have a lot of code to update.

AndrewNolte: I put a small workaround in by returning a lambda. Can I revert to how it was a long time ago with having a self.is_discovered variable? I don't think it's great to have properties return nothing

ktbarrett:
We go off released versions so as long as the version of verilator is released before cocotb is released, all should be well.

And yeah, we are supporting Icarus versions as old as the one available in the package manager of our oldest supported OS (Ubuntu 20.04 uses 10.3 IIRC). Ubuntu 20.04 isn't EoL until next year.

ktbarrett:
Perhaps we should reevaluate our support for Icarus, since we only support very recently Verilators.

imphil:
@AndrewNolte, thanks for working on making cocotb and Verilator work together better! We looked at this PR, and have some suggestions to make it easier to review and merge.

1. If I'm seeing things correctly, your main goal is to make certain objects in a `generate` block accessible by name, instead of only when iterating over them. Can you come up with a single directed test for exactly this behavior? Looking at your code, this test should already pass for all simulators except for Verilator (and possibly Riviera)? You can most likely add this test to the existing discovery test file at https://github.com/cocotb/cocotb/blob/master/tests/test_cases/test_discovery/test_discovery.py. I'd suggest you do a PR with the test first, and mark it as expected to fail in Verilator. This way we can establish a baseline of what works, and what doesn't work.

1. You can then do a separate PR to do the the changes in `VpiImpl.cpp` to support direct access to elements in generate block, and remove the test-exclusion added in step (1).

In addition to the main purpose of this PR, you have included some other valuable changes, which would best go into a separate PR.

1. You have discovered that older Icarus Verilog versions have issues with compiling a Verilog design which uses a parameter in a generate block. Again, it would be helpful if you could add a directed test for this behavior, and make it conditional on the Icarus Verilog version.

2. The `test_vpi_structure` test is similar to our existing discovery tests in `test_discovery.py`. However, it does more advanced checking by comparing not only the number of discovered objects, but also including a golden reference with the actual expectations. That's nice! I have some suggestions on the implementation, and I know others have attempted to do something similar in the past. I would suggest opening an issue, or a quick draft PR (don't spend a huge amount of time on the implementation just yet!) to discuss the implementation approach in more detail.

AndrewNolte:
Thanks for the feed back!

1. The generate blocks weren't accessible in Verilator at all before. They were fixed here https://github.com/verilator/verilator/pull/4913 https://github.com/verilator/verilator/pull/4918, https://github.com/verilator/verilator/pull/4965. The improved vpi implementation is most similar to Icarus, so I went off of that and expanded support for interface arrays.
2.1 I'm not sure why this issue with Icarus is happening. It works fine for me, and I'm apparently on the same version as the CI, but perhaps it's later. I think for Verilator and Icarus we should only support the latest, especially now that we're ramping up 2.0 changes (as @ktbarrett seemed to imply)

The reason I didn't want to touch test_discovery is because I'd have to replicate the sample_module on vhdl as well. I have no idea what the analog for these structures are, or what the outputs for those sims might be. I'm not a huge fan of the back and forth with the CI outputs. Also, sample_module.sv is used everywhere, so it might break things.

I'll split it up like this:
- More test_discovery tests with generate blocks and interface arrays, skipping vhdl on those tests
- Golden structure PR with current outputs
- Vpi fixes with verilator version on CI bumped, golden outputs and test_discovery updated. (once the last PR https://github.com/verilator/verilator/pull/4965 is landed)

imphil:
Just out of curiosity:

> The improved vpi implementation is most similar to Icarus, so I went off of that and expanded support for interface arrays.

Is there a way to make the implementation in Verilator actually match all other simulators, instead of the unusual implementation in Icarus? Or the other way around: why did you choose this implementation?

imphil:
> The reason I didn't want to touch test_discovery is because I'd have to replicate the sample_module on vhdl as well. I have no idea what the analog for these structures are, or what the outputs for those sims might be. I'm not a huge fan of the back and forth with the CI outputs. Also, sample_module.sv is used everywhere, so it might break things.

I understand that. I'd be fine with just a small test in a new test file that test exactly the kind of discovery you're looking for. It doesn't need to be a huge Verilog file!

AndrewNolte:
> Just out of curiosity:
> 
> > The improved vpi implementation is most similar to Icarus, so I went off of that and expanded support for interface arrays.
> 
> Is there a way to make the implementation in Verilator actually match all other simulators, instead of the unusual implementation in Icarus? Or the other way around: why did you choose this implementation?

It was a really simple (almost) one line change. https://github.com/verilator/verilator/pull/4913/files#diff-824689d7895d43409df1624a0b2ea9402679637b419e1adb9a412cf7f5796272. Are Icarus (and now Verilator) the only reasons we ignore the vpiGenScopeArray handles and use pseudo handles?
https://github.com/verilator/verilator/pull/4913/files#diff-824689d7895d43409df1624a0b2ea9402679637b419e1adb9a412cf7f5796272

AndrewNolte:
> I don't think we are going to bump the Icarus version. 10.3 is still the standard on Ubuntu 20.04 and the only limitation it gives seems to be parameters in generate blocks, which is simple enough to just avoid. You can use just `ifdef`s for Icarus and in the expected otherwise.

My take is that if someone's on the head of cocotb, they're probably willing to upgrade their OS or their iverilog install as well. I thought for a sec maybe it would work if it was localparam, but that also fails.

ktbarrett:
Using the most recent version of Icarus requires compiling from source. There are tons of newbies and hobbyists that just want to do `sudo apt install iverilog`.

tests/test_cases/test_vpi_structure/vpi_structure.sv
```diff
@@ -0,0 +1,155 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// Copyright 2010 by Wilson Snyder. This program is free software; you can
+// redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+// Matches the corresponding verilator test file found here https://github.com/verilator/verilator/blob/master/test_regress/t/t_vpi_dump.v
```
imphil, line 9: @AndrewNolte, you seem to be the only author of the Verilator version of this file at https://github.com/verilator/verilator/commits/master/test_regress/t/t_vpi_dump.v. Would you mind relicensing it under the standard cocotb BSD license (https://github.com/cocotb/cocotb/blob/master/CONTRIBUTING.md):

```
```

The licenses are similar enough that not much changes in the end, but having a single license simplifies our life here at cocotb (e.g., when copy-pasting things between files).

AndrewNolte:
Abandoning this since it got split up into:
- https://github.com/cocotb/cocotb/pull/3704
- https://github.com/cocotb/cocotb/pull/3781
- https://github.com/cocotb/cocotb/pull/3824
- https://github.com/cocotb/cocotb/pull/3782

ian-l-kennedy:
@marlonjames @AndrewNolte @ktbarrett:
Will the addition of this PR ensure simple struct elaboration into LogicObject in python is working for each simulator? I currently use xcelium, which elaborates the fields of a simple (no nesting) SV struct into dut.\<some struct port\>.\<some member\>. It works as you would expect, and I can loop through all the attributes available in Python for a dut.\<some struct port\> and see all the members.
 
Verilator though, using cocotb latest main and verilator 5.022 is not elaborating the struct members.
 
@marlonjames is this expected? I had thought you fixed this in past PRs - in all your work in all the comments and PRs linked here: https://github.com/cocotb/cocotb/issues/1275

PR #3763, opened 2024-03-09
Don't use cache when building docs

ktbarrett:


ktbarrett:
So when `build.commands` is used, `apt_packages` is ignored. But you don't have permissions to run `apt` in `build.commands`. This feature is in beta and now I see why =/

I don't see another way to insert `-E` to the sphinx build command.

PR #3776, opened 2024-03-13
Update tool dependencies in noxfile

imphil:
Update the tools used in the noxfile to keep up with the latest and greatest.

- **Update gcov dependency in noxfile**
- **Update to cibuildwheel 2.17**

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -20,7 +20,8 @@ coverage_deps = ["coverage[toml]>=5.0", "pytest-cov"]
 # When using gcovr 5.0, deprecated jinja2.Markup was removed in 3.1, so an
 # Exception is raised during html report generation.
 # See https://github.com/gcovr/gcovr/pull/576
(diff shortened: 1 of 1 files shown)


ktbarrett:
Can we get a release build to see cibuildwheel functions?

ktbarrett (approved):
As long as it works, I'm happy.

ktbarrett:
Looks like gcovr 5.2 doesn't support Python 3.6. More and more is not supporting it. We could hold off on that change until after 2.0 (the last to support 3.6 IMO as RHEL7 is EoL in June).

imphil:
> I think it might be a good idea to add workflow_dispatch so we can manually run builds rather than do DO NOT MERGE commits.

IIRC, you cannot manually trigger commits from branches in a fork, which is where our pull requests tend to (and should) be from.

ktbarrett (approved):
LGTM. Merge away!

PR #3779, opened 2024-03-15
Improve `cocotb/__init__.py`

ktbarrett:
* Removes `cocotb.LANGUAGE` as it's a regression-only feature.
* Make `cocotb.RANDOM_SEED` private, as there's no reason to make that publicly available.
* Introduces `cocotb.is_simulation` to check if cocotb was loaded from a simulation.
* Remove `Optional` from the module-level attrs and all documentation saying "check this to see if you're in a simulation"
* Remove `cocotb.argc` as it's just `len(cocotb.argv)`.
* Some cleanup.


src/cocotb/__init__.py
```diff
@@ -432,3 +357,48 @@ def _stop_user_coverage() -> None:
...
+    if seed_envvar is None:
+        if "ntb_random_seed" in plusargs:
+            plusarg_seed = plusargs["ntb_random_seed"]
+            if not isinstance(plusarg_seed, str):
+                raise TypeError("ntb_random_seed plusarg is not a valid seed value.")
+            _random_seed = int(plusarg_seed)
+        elif "seed" in plusargs:
+            plusarg_seed = plusargs["seed"]
+            if not isinstance(plusarg_seed, str):
+                raise TypeError("ntb_random_seed plusarg is not a valid seed value.")
```
marlonjames, line 375:
```suggestion
                raise TypeError("seed plusarg is not a valid seed value.")
```

src/cocotb/__init__.py
```diff
@@ -432,3 +357,48 @@ def _stop_user_coverage() -> None:
...
+def _setup_random_seed() -> None:
+    global _random_seed
+
+    seed_envvar = os.getenv("RANDOM_SEED")
+    if seed_envvar is None:
+        if "ntb_random_seed" in plusargs:
+            plusarg_seed = plusargs["ntb_random_seed"]
+            if not isinstance(plusarg_seed, str):
+                raise TypeError("ntb_random_seed plusarg is not a valid seed value.")
+            _random_seed = int(plusarg_seed)
```
marlonjames, line 371: We don't have test coverage for these, and I don't know who uses these or what the values look like,
but using `int()` does change the behavior for some seed values.
If the seed is set to something like `0x345`, `eval()` would give an integer `837`, but `int()` raises a `ValueError`.

src/cocotb/__init__.py
```diff
@@ -432,3 +357,48 @@ def _stop_user_coverage() -> None:
...
+            root_name = None
+        elif "." in root_name:
+            # Skip any library component of the toplevel
+            root_name = root_name.split(".", 1)[1]
+
+    from cocotb import simulator
+
+    handle = simulator.get_root_handle(root_name)
+    if not handle:
+        raise RuntimeError(f"Can not find root handle ({root_name})")
```
marlonjames, line 401:
```suggestion
        raise RuntimeError(f"Can not find root handle '{root_name}'")
```

src/cocotb/__init__.py
ktbarrett: There is `ast.literal_eval` and I could add another type check afterwards for `int`. `_random_seed` must be an `int` eventually.

marlonjames: Sounds fine to me.

docs/source/newsfragments/3779.removal.rst
```diff
@@ -0,0 +1 @@
+``cocotb.LANGUAGE`` was removed.
```
cmarqu, line 1:
```suggestion
``cocotb.LANGUAGE`` was removed, use ``os.environ["TOPLEVEL_LANG"]`` if you need that information.
```

src/cocotb/__init__.py
```diff
@@ -111,47 +85,38 @@ def _setup_logging() -> None:
 .. versionadded:: 2.0
 """
 
-LANGUAGE: Optional[str] = os.getenv("TOPLEVEL_LANG")
-"""The value of :make:var:`TOPLEVEL_LANG`.
-
-This is guaranteed to hold a value at test time.
-"""
+SIM_NAME: str
+"""The running simulator product information."""
```
cmarqu, line 89:
```suggestion
"""The product information of the running simulator."""
```

src/cocotb/__init__.py
cmarqu, line 401:
```suggestion
        raise RuntimeError(f"Can not find root handle {root_name!r}")
```
Alternative to the suggestion by @marlonjames, we have that syntax elsewhere.

marlonjames (approved):

PR #3780, opened 2024-03-15
Fix pre-commit bump

ktbarrett:
More changes... this is starting to seem like more of a chore than it should be.


cmarqu:
Maybe https://pypi.org/project/pre-commit-update/#pre-commit-hook-usage-example works better?

ktbarrett:
We settled on switching to pre-commit.ci.

ktbarrett:
We no longer need the lint action as pre-commit.ci does the same thing. pre-commit.ci should also autoupdate the dependencies. I added the `autofix_prs: false` config so that the CI wouldn't push lint fix PRs up automatically. I feel that breaks "have good commits".

imphil (approved):
LGTM. Pre-commit CI seems to run on this PR, so only auto-update is still to be tested.

PR #3783, opened 2024-03-15
Small typing cleanup of Task and related

marlonjames:


src/cocotb/_outcomes.py
```diff
@@ -24,7 +24,7 @@ def send(self, gen):
         """Send or throw this outcome into a generator"""
 
     @abc.abstractmethod
-    def get(self, gen):
+    def get(self):
```
ktbarrett, line 27: Can you add typing to this?

ktbarrett (approved):
Well... I'm happy with that too.

PR #3790, opened 2024-03-19
Reintroduced casts on handles

### LINKED ISSUE #3565: Proposal: revert handle.py cast removals
This commit https://github.com/cocotb/cocotb/commit/a59fba89fe177d9d0d2532238a55f85a0a18e4f1 
breaks a lot of code

I think str(dut.myhandle) should be removed, because the cast does something that's sort of ambiguous- it seems like it might interpret the value as a string, but it actually returns the path.

int(dut.myhandle) has always returned the value as an int and turning it into int(dut.myhandle.value) doesn't make it any more readable. So I think it makes sense to add int() and float() back in. I agree in general builtin overrides aren't a great solution because of the ambiguity, but I think the cost of the amount of code these break justifies leaving them in.

ktbarrett:
Closes #3565. Reintroduced the value casts on handles, but they are deprecated. Will go through `LogicArray` and add a bunch more deprecations soon too.

### DIFF
--- a/src/cocotb/handle.py
+++ b/src/cocotb/handle.py
@@ -1002,6 +1002,18 @@ class LogicObject(
     def value(self, value: LogicArray) -> None:
         self.set(value)
 
+    @deprecated(
+        "`int(handle)` casts have been deprecated. Use `int(handle.value)` instead."
+    )
+    def __int__(self) -> int:
+        return int(self.value)
+
+    @deprecated(
+        "`str(handle)` casts have been deprecated. Use `str(handle.value)` instead."
+    )
+    def __str__(self) -> str:
+        return str(self.value)
+
 
 class RealObject(ValueObjectBase[float, float]):
     """A real/float simulation object.
@@ -1045,6 +1057,12 @@ class RealObject(ValueObjectBase[float, float]):
     @value.setter
     def value(self, value: float) -> None:
         self.set(value)
+
+    @deprecated(
+        "`float(handle)` casts have been deprecated. Use `float(handle.value)` instead."
+    )
+    def __float__(self) -> float:
+        return self.value
 
 
 class EnumObject(ValueObjectBase[int, int]):
@@ -1099,6 +1117,12 @@ class EnumObject(ValueObjectBase[int, int]):
     @value.setter
     def value(self, value: int) -> None:
         self.set(value)
+
+    @deprecated(
+        "`int(handle)` casts have been deprecated. Use `int(handle.value)` instead."
+    )
+    def __int__(self) -> int:
+        return int(self.value)
 
 
 class IntegerObject(ValueObjectBase[int, int]):
@@ -1164,6 +1188,12 @@ class IntegerObject(ValueObjectBase[int, int]):
     @value.setter
     def value(self, value: int) -> None:
         self.set(value)
+
+    @deprecated(
+        "`int(handle)` casts have been deprecated. Use `int(handle.value)` instead."
+    )
+    def __int__(self) -> int:
+        return self.value
 
 
 class StringObject(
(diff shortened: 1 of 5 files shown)


ktbarrett:
Had to refactor `test_package_access` since it was tripping the `__str__` cast on `LogicObject` by using `str` to get the `repr` =/

src/cocotb/handle.py
```diff
@@ -1002,6 +1002,18 @@ def value(self) -> LogicArray:
     def value(self, value: LogicArray) -> None:
         self.set(value)
 
+    @deprecated(
+        "`int()` casts on handles have been deprecated. Use `int(handle.value)` instead."
```
cmarqu, line 1006:
```suggestion
        "`int(handle)` casts have been deprecated. Use `int(handle.value)` instead."
```
A bit clearer IMO; but feel free to ignore.

src/cocotb/handle.py
```diff
@@ -1002,6 +1002,18 @@ def value(self) -> LogicArray:
...
+    )
+    def __int__(self) -> int:
+        return int(self.value)
+
+    @deprecated(
+        "`str()` casts on handles have been deprecated. Use `str(handle.value)` instead."
```
cmarqu, line 1012:
```suggestion
        "`str(handle)` casts have been deprecated. Use `str(handle.value)` instead."
```

src/cocotb/handle.py
```diff
@@ -1046,6 +1058,12 @@ def value(self) -> float:
     def value(self, value: float) -> None:
         self.set(value)
 
+    @deprecated(
+        "`float()` casts on handles have been deprecated. Use `str(handle.value)` instead."
```
cmarqu, line 1062:
```suggestion
        "`float(handle)` casts have been deprecated. Use `float(handle.value)` instead."
```
This one had a real error: `s/str/float/`

src/cocotb/handle.py
```diff
@@ -1100,6 +1118,12 @@ def value(self) -> int:
     def value(self, value: int) -> None:
         self.set(value)
 
+    @deprecated(
+        "`int()` casts on handles have been deprecated. Use `int(handle.value)` instead."
```
cmarqu, line 1122:
```suggestion
        "`int(handle)` casts have been deprecated. Use `int(handle.value)` instead."
```

src/cocotb/handle.py
```diff
@@ -1165,6 +1189,12 @@ def value(self) -> int:
...
```
cmarqu, line 1193:
```suggestion
        "`int(handle)` casts have been deprecated. Use `int(handle.value)` instead."
```

src/cocotb/handle.py
```diff
@@ -1235,6 +1265,12 @@ def value(self) -> bytes:
     def value(self, value: bytes) -> None:
         self.set(value)
 
+    @deprecated(
+        "`str()` casts on handles have been deprecated. Use `handle.value.decode('ascii')` instead."
```
cmarqu, line 1269:
```suggestion
        "`str(handle)` casts have been deprecated. Use `handle.value.decode('ascii')` instead."
```

src/cocotb/handle.py
```diff
@@ -1235,6 +1265,12 @@ def value(self) -> bytes:
     def value(self, value: bytes) -> None:
         self.set(value)
 
+    @deprecated(
+        "`str(handle)` casts have been deprecated. Use `handle.value.decode('ascii')` instead."
```
cmarqu, line 1269:
```suggestion
        '`str(handle)` casts have been deprecated. Use `handle.value.decode("ascii")` instead.'
```
Let's suggest code that `black` won't change for users.

cmarqu (approved):
Just another small change.

PR #3791, opened 2024-03-19
Fix help text of combine_results and integrate into cocotb_tools

rkryan:
The help text of combine_results.py incorrectly specifies the default value of --set_rc as True. Because the argument parser uses ArgumentDefaultsHelpFormatter, the fix is not as simple as updating the help string.

Rather, the issue here is that the command line argument suppress_rc and the internal variable set_rc have opposite logical meaning. Switching the set_rc variable to suppress_rc makes sure the value is consistent with the logical meaning of the argument while ensuring the command line API does not change.

---

This removes the last script from the standalone bin directory,
preferring instead to deliver this functionality as part of the
cocotb_tools package. As all usage of this is internal to the cocotb
repository, no console_scripts entry is added for this.

`python -m cocotb_tools.combine_results` is a drop-in replacement for
`combine_results.py`.


ktbarrett (approved):

ktbarrett:
I think a nice follow up to this would be to move `combine_results.py` into `cocotb_tools` and make `python -m cocotb_tools.combine_results` work. Then update the regression to use it. I don't think that should be too much work.

rkryan:
> I think a nice follow up to this would be to move `combine_results.py` into `cocotb_tools` and make `python -m cocotb_tools.combine_results` work. Then update the regression to use it. I don't think that should be too much work.

Yea, that seems pretty easy. I could put that together too.

marlonjames:
Windows environment can't find `cocotb_tools`.
https://github.com/cocotb/cocotb/actions/runs/8351427065/job/22859739370?pr=3791#step:19:3649

ktbarrett:
Well Windows is poopy, we shouldn't let it discourage us. I'm wondering if the issue is that we are calling out `python3` and that isn't the correct python executable to use. Maybe just `python` would work? Alternatively, we can add an entry point like `cocotb-config` ([see here](https://github.com/cocotb/cocotb/blob/785e0896930039023db3c2e2be47bed5375326c5/setup.py#L131)).

cmarqu:
While we are at it, should we rename the arguments to use `--suppress-rc` style instead of `--suppress_rc`?

ktbarrett:
@cmarqu The other options use the same syntax. Let's leave that for a later refactor.

ktbarrett (approved):
Looks like just `python` worked for Windows!

rkryan:
Yup, switching to `python` worked.

The one thing about additional cleanup is that this change officially exposes this script to end users. Since it's not in console_scripts it's not as obvious as if it was in bin/, but nevertheless users can call this from an install for the first time. So any cleanup affecting the API might be better sooner than later.

ktbarrett:
@rkryan Yeah, I had considered that. I anticipate doing a cleanup PR after this to work on documentation, typing, things like `--supress_rc` to `--supress-rc`, etc. It should be done before the 2.0 release.

ktbarrett:
Usually we like to get 2 approvals, but I'll merge in a couple days if we don't.

cmarqu (approved):

PR #3796, opened 2024-03-21
Introduce language-agnostic `sources` arg to runner

### LINKED ISSUE #3404: Mixed-languages multi-libraries not working as expected
# Issue description #
compilation of a design containing both Verilog and VHDL and also mixing libraries does not work. The RTL and the python script can be found here : https://github.com/oliverbm67/cocotb/tree/e63dda34cbfd64b0600cbbb3bfe4e1db7f26707c/examples/mixed_language_multi_libraries

I noticed the following :

- If hdl_toplevel is not provided in the runner.build a default top is used in the tool command
- If no hdl_library is provided, cocotb 

does not default to work as (at least in my experience) most tools do

# Logs #
The run.do script used to check compilation with questa is also attached
[cocotb.log](https://github.com/cocotb/cocotb/files/12500962/cocotb.log)
[modelsim.log](https://github.com/cocotb/cocotb/files/12500976/modelsim.log)
[run.do.txt](https://github.com/cocotb/cocotb/files/12500980/run.do.txt)

# Environment #
Python 3.11.4
cocotb 1.8.0
Questa 2021.2 ( from Intel FPGA suite)

# Notes #
I do think using Xcelium gives similar results.

ktbarrett:
Closes #3404. Introduces a language-agnostic `sources` argument to the runner. These are segregated into Verilog and VHDL sources when necessary by using the extension, or new tagging types `VHDL` and `Verilog`. The tagging types can also be applied to `build_args`.

- [x] newsfragment
- [x] documentation
- [x] tests

### DIFF
--- a/docs/source/newsfragments/3796.feature.rst
+++ b/docs/source/newsfragments/3796.feature.rst
@@ -1,0 +1,1 @@
+The ``sources`` option was added to :meth:`cocotb_tools.runner.Simulator.build` to better support building mixed-language designs.
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -60,6 +60,14 @@ def shlex_join(split_command):
     return " ".join(shlex.quote(arg) for arg in split_command)
 
 
+class VHDL(str):
+    """Tags source files and build arguments to :meth:`~cocotb_tools.runner.Simulator.build` as VHDL-specific."""
+
+
+class Verilog(str):
+    """Tags source files and build arguments to :meth:`~cocotb_tools.runner.Simulator.build` as Verilog-specific."""
+
+
 class Simulator(abc.ABC):
     supported_gpi_interfaces: Dict[str, List[str]] = {}
 
@@ -92,10 +100,17 @@ class Simulator(abc.ABC):
             ValueError: *hdl_toplevel_lang* is not supported by the simulator.
         """
         if hdl_toplevel_lang is None:
-            if self.vhdl_sources and not self.verilog_sources:
+            if self.vhdl_sources and not self.verilog_sources and not self.sources:
                 lang = "vhdl"
-            elif self.verilog_sources and not self.vhdl_sources:
+            elif self.verilog_sources and not self.vhdl_sources and not self.sources:
                 lang = "verilog"
+            elif self.sources and not self.vhdl_sources and not self.verilog_sources:
+                if is_vhdl_source(self.sources[-1]):
+                    lang = "vhdl"
+                elif is_verilog_source(self.sources[-1]):
+                    lang = "verilog"
+                else:
+                    raise UnknownFileExtension(self.sources[-1])
             else:
                 raise ValueError(
                     f"{type(self).__qualname__}: Must specify a hdl_toplevel_lang in a mixed-language design"
@@ -144,10 +159,11 @@ class Simulator(abc.ABC):
         hdl_library: str = "top",
         verilog_sources: Sequence[PathLike] = [],
         vhdl_sources: Sequence[PathLike] = [],
+        sources: Sequence[Union[PathLike, VHDL, Verilog]] = [],
         includes: Sequence[PathLike] = [],
         defines: Mapping[str, object] = {},
         parameters: Mapping[str, object] = {},
-        build_args: Sequence[str] = [],
+        build_args: Sequence[Union[str, VHDL, Verilog]] = [],
         hdl_toplevel: Optional[str] = None,
         always: bool = False,
         build_dir: PathLike = "sim_build",
@@ -159,10 +175,42 @@ class Simulator(abc.ABC):
     ) -> None:
         """Build the HDL sources.
 
+        With mixed language simulators, *sources* will be built,
+        followed by *vhdl_sources*, then *verilog_sources*.
+        With simulators that only support either VHDL or Verilog, *sources* will be built,
+        followed by *vhdl_sources* and *verilog_sources*, respectively.
+
+        If your source files use an atypical file extension,
+        use :class:`VHDL` and :class:`Verilog` to tag the path as a VHDL or Verilog source file, respectively.
+        If the filepaths aren't tagged, the extension is used to determine if they are VHDL or Verilog files.
+
+        +----------+------------------------------------+
+        | Language | File Extensions                    |
+        +==========+====================================+
+        | VHDL     | ``.vhd``, ``.vhdl``                |
+        +----------+------------------------------------+
+        | Verilog  | ``.v``, ``.sv``, ``.vh``, ``.svh`` |
+        +----------+------------------------------------+
+
+
+        .. code-block:: python3
+
+            runner.build(
+                sources=[
+                    VHDL("/my/file.is_actually_vhdl"),
+                    Verilog("/other/file.verilog"),
+                ],
+            )
+
+        The same tagging works for *build_args*.
+        Tagged *build_args* only supply that option to the compiler when building the source file for the tagged language.
+        Non-tagged *build_args* are supplied when compiling any language.
+
         Args:
             hdl_library: The library name to compile into.
             verilog_sources: Verilog source files to build.
             vhdl_sources: VHDL source files to build.
+            sources: Language-agnostic list of source files to build.
             includes: Verilog include directories.
             defines: Defines to set.
             parameters: Verilog parameters or VHDL generics.
@@ -190,6 +238,7 @@ class Simulator(abc.ABC):
         self.hdl_library: str = hdl_library
         self.verilog_sources: List[Path] = get_abs_paths(verilog_sources)
         self.vhdl_sources: List[Path] = get_abs_paths(vhdl_sources)
+        self.sources: List[Path] = get_abs_paths(sources)
         self.includes: List[Path] = get_abs_paths(includes)
         self.defines = dict(defines)
         self.parameters = dict(parameters)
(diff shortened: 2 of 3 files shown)


src/cocotb_tools/runner.py
```diff
@@ -882,53 +945,59 @@ def _get_parameter_options(parameters: Mapping[str, object]) -> Command:
...
 
+    def _build_vhdl_source(self, source: PathLike) -> str:
+        return "acom -work {RTL_LIBRARY} {EXTRA_ARGS} {VHDL_SOURCES}\n".format(
+            RTL_LIBRARY=as_tcl_value(self.hdl_library),
+            VHDL_SOURCES=" ".join(as_tcl_value(str(v)) for v in self.vhdl_sources),
+            EXTRA_ARGS=" ".join(as_tcl_value(v) for v in self.build_args),
+        )
+
+    def _build_verilog_source(self, source: PathLike) -> str:
+        return "alog -work {RTL_LIBRARY} -pli {EXT_NAME} -sv {DEFINES} {INCDIR} {EXTRA_ARGS} {VERILOG_SOURCES} \n".format(
```
ktbarrett, line 990:
```suggestion
        return "alog -work {RTL_LIBRARY} -pli {EXT_NAME} -sv {DEFINES} {INCDIR} {EXTRA_ARGS} {VERILOG_SOURCES}".format(
```

src/cocotb_tools/runner.py
```diff
@@ -882,53 +945,59 @@ def _get_parameter_options(parameters: Mapping[str, object]) -> Command:
...
 
         do_file = tempfile.NamedTemporaryFile(delete=False)
-        do_file.write(do_script.encode())
+        do_file.write("\n".join(do_script).encode())
         do_file.close()
 
         return [["vsimsa"] + ["-do"] + ["do"] + [do_file.name]]
 
+    def _build_vhdl_source(self, source: PathLike) -> str:
+        return "acom -work {RTL_LIBRARY} {EXTRA_ARGS} {VHDL_SOURCES}\n".format(
```
ktbarrett, line 983:
```suggestion
        return "acom -work {RTL_LIBRARY} {EXTRA_ARGS} {VHDL_SOURCES}".format(
```

src/cocotb_tools/runner.py
```diff
@@ -882,53 +945,59 @@ def _get_parameter_options(parameters: Mapping[str, object]) -> Command:
         return [f"-g{name}={value}" for name, value in parameters.items()]
 
     def _build_command(self) -> List[Command]:
-        do_script = "\nonerror {\n quit -code 1 \n} \n"
+        do_script: List[str] = ["onerror {\n quit -code 1 \n}"]
 
         out_file = self.build_dir / self.hdl_library / f"{self.hdl_library}.lib"
 
         if outdated(out_file, self.verilog_sources + self.vhdl_sources) or self.always:
             do_script += f"alib {as_tcl_value(self.hdl_library)} \n"
```
ktbarrett, line 953:
```suggestion
            do_script.append(f"alib {as_tcl_value(self.hdl_library)}")
```

src/cocotb_tools/runner.py
```diff
@@ -561,10 +588,16 @@ def _test_command(self) -> List[Command]:
         ]
 
     def _build_command(self) -> List[Command]:
-        if self.vhdl_sources:
-            raise ValueError(
-                f"{type(self).__qualname__}: Simulator does not support VHDL"
-            )
+        for source in self.sources:
+            if not is_verilog_source(source):
```
cmarqu, line 592: We may want to support more file types at some point, so ideally the decision logic here would already be prepared for when that time comes.

ktbarrett: That would be preferable, but I think here since we only support Verilog sources on this simulator, this is appropriate. Other places, like filtering the build args `not isinstance(arg, VHDL)` is less precise than `isinstance(arg, Verilog) or type(arg) is str`.

src/cocotb_tools/runner.py
```diff
@@ -630,30 +666,45 @@ def _get_parameter_options(parameters: Mapping[str, object]) -> Command:
...
+        cmds.append(["vlib", as_tcl_value(self.hdl_library)])
+        for source in self.sources:
+            if is_vhdl_source(source):
+                cmds.append(self._build_vhdl_command(source))
+            elif is_verilog_source(source):
+                cmds.append(self._build_verilog_command(source))
+            else:
+                raise ValueError(
+                    f"Can't determine if {source} is a VHDL or Verilog file. "
+                    "Use a standard file extension (`.vhd`, `.v`, or `.sv`) "
```
cmarqu, line 678: Pull out the supported file extensions to a common constant and report the exact list/set here?

marlonjames:
So you're protecting `build_args` in case they pass in source files in that list?

ktbarrett:
The idea is you can tag build arts to apply only to building VHDL or Verilog. This might be useful in mixed language simulators.

src/cocotb_tools/runner.py
```diff
@@ -163,6 +170,7 @@ def build(
             hdl_library: The library name to compile into.
             verilog_sources: Verilog source files to build.
             vhdl_sources: VHDL source files to build.
+            sources: Language-agnostic list of source files to build.
```
marlonjames, line 173: I think we should add some documentation about the build order for `sources`, `vhdl_sources`, and `verilog_sources`.

src/cocotb_tools/runner.py
```diff
@@ -92,9 +98,9 @@ def _check_hdl_toplevel_lang(self, hdl_toplevel_lang: Optional[str]) -> str:
             ValueError: *hdl_toplevel_lang* is not supported by the simulator.
         """
         if hdl_toplevel_lang is None:
-            if self.vhdl_sources and not self.verilog_sources:
+            if self.vhdl_sources and not self.verilog_sources and not self.sources:
```
marlonjames, line 101: If only `sources` is given, couldn't we check the last provided source file?

src/cocotb_tools/runner.py
```diff
@@ -159,10 +173,33 @@ def build(
     ) -> None:
         """Build the HDL sources.
 
+        With mixed language simulators, *sources* will be built,
+        followed by *vhdl_sources*, then *verilog_sources*.
+        With simulators that only support either VHDL or Verilog, *sources* will be built,
+        followed by *vhdl_sources* and *verilog_sources*, respectively.
+
+        If your sources files use an atypical file extension,
```
marlonjames, line 181:
```suggestion
        If your source files use an atypical file extension,
```

src/cocotb_tools/runner.py
```diff
@@ -159,10 +173,33 @@ def build(
...
+        use :class:`VHDL` and :class:`Verilog` to *tag* the path as a VHDL or Verilog source file, respectively.
```
marlonjames, line 182: Other italicized words are arguments to build. If you want emphasis you could bold it.
```suggestion
        use :class:`VHDL` and :class:`Verilog` to tag the path as a VHDL or Verilog source file, respectively.
```

docs/source/newsfragments/3796.feature.rst
```diff
@@ -0,0 +1 @@
+The ``sources`` option was added to :meth:`cocotb.runner.Simulator.build` to better support building mixed-language designs.
```
marlonjames, line 1:
```suggestion
The ``sources`` option was added to :meth:`cocotb_tools.runner.Simulator.build` to better support building mixed-language designs.
```

src/cocotb_tools/runner.py
```diff
@@ -159,10 +173,33 @@ def build(
...
+        If the filepaths aren't tagged, the extension is used to determine if they are VHDL or Verilog files.
+
```
marlonjames, line 184:
```suggestion
        If the filepaths aren't tagged, the extension is used to determine if they are VHDL or Verilog files:
        
        +----------+------------------------------------+
        | Language | File Extensions                    |
        +==========+====================================+
        | VHDL     | ``.vhd``, ``.vhdl``                |
        +----------+------------------------------------+
        | Verilog  | ``.v``, ``.sv``, ``.vh``, ``.svh`` |
        +----------+------------------------------------+

```

src/cocotb_tools/runner.py
```diff
@@ -60,6 +60,12 @@ def shlex_join(split_command):
     return " ".join(shlex.quote(arg) for arg in split_command)
 
 
+class VHDL(str): ...
+
+
+class Verilog(str): ...
+
+
```
marlonjames, line 68: Something like this? Having them blank in the documentation is a bit confusing.
```suggestion
class VHDL(str): ...
    """Wrapper to tag string arguments to :meth:`~cocotb_tools.runner.Simulator.build` as VHDL-specific."""

class Verilog(str): ...
    """Wrapper to tag string arguments to :meth:`~cocotb_tools.runner.Simulator.build` as Verilog-specific."""


```

src/cocotb_tools/runner.py
```diff
@@ -491,6 +529,40 @@ def get_abs_paths(paths: Sequence[PathLike]) -> List[Path]:
     return [get_abs_path(path) for path in paths]
 
 
+_verilog_extensions = (".v", ".sv", ".vh", ".svh")
```
cmarqu, line 532: Actually... are `.vh` and `.svh` not files that should only be included and not directly compiled? We might say that users shouldn't list them in the sources then, but it would allow them to just do a recursive glob over a toplevel directory without extra filtering.

Maybe we should expose the extensions as a user-definable option?

src/cocotb_tools/runner.py
```diff
@@ -491,6 +529,40 @@ def get_abs_paths(paths: Sequence[PathLike]) -> List[Path]:
...
+_vhdl_extensions = (".vhd", ".vhdl")
+
+_vhdl_extensions_s = ", ".join(f"`{c}`" for c in _vhdl_extensions)
+_veriog_extensions_s = ", ".join(f"`{c}`" for c in _verilog_extensions)
```
cmarqu, line 536:
```suggestion
_verilog_extensions_s = ", ".join(f"`{c}`" for c in _verilog_extensions)
```

src/cocotb_tools/runner.py
```diff
@@ -491,6 +529,40 @@ def get_abs_paths(paths: Sequence[PathLike]) -> List[Path]:
...
+
+
+class UnknownFileExtension(ValueError):
+    def __init__(self, source: PathLike) -> None:
+        super().__init__(
+            f"Can't determine if {source} is a VHDL or Verilog file. "
+            f"Use a standard file extension ({_vhdl_extensions_s} for VHDL files and {_veriog_extensions_s} for Verilog files) "
```
cmarqu, line 543:
```suggestion
            f"Use a standard file extension ({_vhdl_extensions_s} for VHDL files and {_verilog_extensions_s} for Verilog files) "
```

src/cocotb_tools/runner.py
```diff
@@ -159,10 +173,33 @@ def build(
     ) -> None:
         """Build the HDL sources.
 
+        With mixed language simulators, *sources* will be built,
+        followed by *vhdl_sources*, then *verilog_sources*.
```
cmarqu, line 177: In the ASIC world, low-level models (cell primitives from a PDK) are usually Verilog, while higher levels are VHDL or Verilog, so changing the order (or being able to change it via a switch) would be helpful in that context.

src/cocotb_tools/runner.py
ktbarrett: I'm seen some stuff you wouldn't imagine... And from major vendors at that. Sometimes header files need to be compiled as source files and not just included.

src/cocotb_tools/runner.py
ktbarrett: vhdl_sources and verilog_sources will be deprecated by a follow-on PR. The intent is to only use `sources` going forward.

cmarqu (approved):

PR #3803, opened 2024-03-22
Improve combine_results script

ktbarrett:
Follow on from #3791.

* Allow multiple directories to be provided to script to search for results.xml files.
* Removed --suppress-rc, just use the shell to ignore the return code if desired.
* Typing and public API cleanup.
* Cleanup configuration of ArgumentParser.
* Added newsfragment.
* Added documentation.

### DIFF
--- a/docs/source/newsfragments/3791.feature.rst
+++ b/docs/source/newsfragments/3791.feature.rst
@@ -1,0 +1,1 @@
+The :ref:`combine_results.py <combine-results>` script now ships with the cocotb installation.
--- a/src/cocotb_tools/combine_results.py
+++ b/src/cocotb_tools/combine_results.py
@@ -1,110 +1,98 @@
 #!/usr/bin/env python
 """
 Simple script to combine JUnit test results into a single XML file.
-
-Useful for Jenkins.
 """
 
 import argparse
 import os
+import re
 import sys
+from typing import Iterable, Pattern
 from xml.etree import ElementTree as ET
 
 
-def find_all(name, path):
-    for root, dirs, files in os.walk(path):
-        if name in files:
-            yield os.path.join(root, name)
+def _find_all(name: Pattern, path: str) -> Iterable[str]:
+    for root, _, files in os.walk(path):
+        for file in files:
+            if re.match(name, file):
+                yield os.path.join(root, file)
 
 
-def get_parser():
+def _get_parser() -> argparse.ArgumentParser:
     """Return the cmdline parser"""
     parser = argparse.ArgumentParser(
         description=__doc__, formatter_class=argparse.ArgumentDefaultsHelpFormatter
     )
-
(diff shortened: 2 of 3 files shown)


cmarqu (approved):

docs/source/newsfragments/3791.feature.rst
```diff
@@ -0,0 +1 @@
+The ``bin/combine_results.py`` script from the repo now ships with the cocotb installation. See usage by calling ``python -m cocotb_tools.combine_results -h``.
```
ktbarrett, line 1: Add a link to the documentation.

src/cocotb_tools/combine_results.py
```diff
@@ -1,110 +1,84 @@
...
-        "--directory",
-        dest="directory",
-        type=str,
-        required=False,
-        default=".",
-        help="Name of base directory to search from",
+        "directories",
+        nargs="*",
+        default=["."],
+        help="Path to base directory to search from",
```
ktbarrett, line 28:
```suggestion
        help="Directories to search for 'result.xml' files",
```

marlonjames, line 28: ?
```suggestion
        help="Paths to one or more base directory to search from",
```

marlonjames (approved):

src/cocotb_tools/combine_results.py
```diff
@@ -1,110 +1,84 @@
...
-
-    if args.debug:
-        ET.dump(result)
+                use_element = None
+                for existing in result:
+                    if existing.get("name") == ts.get("name") and existing.get(
+                        "package"
+                    ) == ts.get("package"):
+                        if args.verbose:
+                            print("Already found.")
```
ktbarrett, line 75: What is this doing exactly?

ktbarrett (commented):
Add an option to handle alternative naming of results file since we support `COCOTB_RESULTS_FILE`.

src/cocotb_tools/combine_results.py
ktbarrett: Saying whether the testsuite (whose info was just printed) was already found in the results. In that case it extends that testsuite rather than adding it "new" to the results.

ktbarrett:
I added `--input-filenames` to allow the user to specify a filename other than `results.xml` since they can do that with `COCOTB_RESULTS_FILE`. I renamed `--testsuites-name` to `--output-testsuites-name` for clarity. And I added `-i` and `-o` short aliases for `--input-filenames` and `--output-file`.

src/cocotb_tools/combine_results.py
marlonjames: If the `name` and `package` of a `testsuite` has already been found, it merges the `testcase` elements into the previously found one.

PR #3804, opened 2024-03-22
Improve cocotb-config script

ktbarrett:
* Make things private besides
   * directories
   * lib_name
   * lib_name_path
   * main
* Reword documentation
* Use mutually exclusive group for options
* Remove unused -v alias for --version
* Refactor to not use custom actions

The things remaining public are used by other parts of cocotb and are not available in any other way.

The refactor to not use custom actions leads to better generated documentation with sphinx_argparse_cli.

Currently https://github.com/tox-dev/sphinx-argparse-cli/issues/159 is in our way wrt clean documentation for `--lib-name` and `--lib-name-path`.

- [x] Consider removing the unused `--prefix` option.

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -360,5 +360,5 @@ The ``cocotb-config`` script
 
 .. sphinx_argparse_cli::
     :module: cocotb_tools.config
-    :func: get_parser
+    :func: _get_parser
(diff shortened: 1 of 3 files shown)


cmarqu (approved):

ktbarrett:
Removed `--prefix` as it was unused.

Issue #3757, opened 2024-03-05
Always `_discover_all` when finding a sub-handle


ktbarrett:
Currently we opportunistically attempt to find a handle by name without first discovering sub-handles. This often does not work. We could instead add some logic such that if it fails, it `_discover_all`s and then tries again before failing.

marlonjames:
#2079 added this to try and help with pseudo-region discovery, since the VPI iteration code was better at finding those.
The followup #2143 which added iteration fallback in child handle lookup by name for Icarus took it out. And now #3817 expanded the VPI fallback to all simulators.

I'm not opposed to it in theory, I just wonder if there are any cases where iterating using the ordered set of iterators in the GPI implementations would give a handle to a different object than when looked up by name directly.

ktbarrett:
Better to move this into the GPI.

PR #3817, opened 2024-03-29
VPI: Use iteration fallback when looking for vpiGenScope by name

marlonjames:
Like Icarus Verilog, Verilator does not support vpiGenScopeArray, but vpiGenScope support has been added.

When looking up a vpiGenScopeArray handle by name, iterate over parent handle to check if a vpiGenScope object exists with the right prefix. If it does, create the pseudo-region.

This implements just the Verilator `vpiGenScope` support from #3756.

- [x] Add newsfragment

### DIFF
--- a/docs/source/newsfragments/3817.feature.rst
+++ b/docs/source/newsfragments/3817.feature.rst
@@ -1,0 +1,1 @@
+Enable use of VPI fallback in all simulators when attempting to access generate blocks directly via lookup. This enables better support for simulators that don't support ``vpiGenScopeArray``, allowing discovery of generate blocks without having to iterate over the parent handle.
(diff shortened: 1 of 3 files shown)


ktbarrett:
Any reason to not make this unconditional?

Also, I must be telepathic.

marlonjames:
I'll try that and see if there are any problems.

Also, #2079 newsfragment was a bugfix. I don't really care what we classify this as.

marlonjames:
Interesting, that fixed Questa VPI as well.

ktbarrett (approved):
Nice... My only question is, does this run *every* time a lookup fails, or only the first time?

marlonjames:
It will run anytime `vpi_handle_by_name()` fails to find a handle.
How many handles are discovered in most tests? They are usually found at the beginning, and the parent handles in Python cache the child handles.

ktbarrett:
Well my concern is that every name lookup failure causes this code to rerun. Python only caches handles that *it* finds. So if the C++ iterates and finds it, Python doesn't know, and its gonna invoke this again.

I'm fine with this coming in because it makes it appear as if the GPI doesn't have issues. I think this also needs to be done in Python per #3757.

Issue #1024, opened 2019-07-16
VPI: Explore always using vpiNoDelay 


imphil:
Currently we're using vpiInertialDelay to set signals (except for strings). vpiNoDelay could be more appropriate, but doesn't work currently.

Explore if switching makes sense, and what roadblocks we might hit.

(Discussion started in https://github.com/potentialventures/cocotb/issues/1014 and https://github.com/potentialventures/cocotb/pull/1016)

mballance:
I echo @cmarqu's comments in #1016: using inertial delays isn't common in my experience. I worked developing CPU simulation models ~20 years ago on multiple simulators (there were quite a few Verilog simulators back then), and we never used inertial delays.
I tried using vpiNoDelay with cocotb master and a new-ish version of Questa, and it seemed to work fine. Are there specific simulators that seem to have issues?

garmin-mjames:
Icarus (as of 12.0 (devel) (s20150603-954-g359b2b65)) calls `cbValueChange` callbacks immediately when using `vpiNoDelay`: https://github.com/cocotb/cocotb/issues/1484#issuecomment-722592589

toddstrader:
Is it still desired to abandon `vpiInertialDelay` and move entirely to `vpiNoDelay`?  I'm reading through the backstory here (#1484 , #1662 , #1014 , #768 , #2178 ) and it seems like this has been a long and winding road.

Also, is it known if there are currently issues with simulators other than Icarus?

ktbarrett:
I don't think it really should matter since we already wait until the read write stage to apply writes. It does, however, make `setimmediatevalue` function weirdly. I don't really understand the original motivation to use `vpiNoDelay`. I think this was more of an investigation into how well it would work or if it were even a good idea.

> Explore if switching makes sense, and what roadblocks we might hit.

I don't anticipate this changing any time soon.

toddstrader:
Gotcha, so can you elaborate on what the desired behavior is?  `38.34 vpi_put_value()` says:
```
vpiInertialDelay All scheduled events on the object shall be removed before this event is
scheduled.
```
which I see causes a `vpiSchedEvent` to be created even if there's no other obvious (at least to me) event scheduled on the object in question:
https://edaplayground.com/x/aDri

[...]

PR #3825, opened 2024-04-03
make setimmediatevalue immediately set

toddstrader:
See: #1024 

Currently `vpiInertialDelay` prevents `setimmediatevalue()` from immediately setting a value.  This attempts to fix that, but honestly raises more questions in my mind.  Namely that the VHPI and FLI interfaces don't seem to have a notion analogous to `vpiIntertialDelay`.  So I'm once again wondering if we really want that for the VPI?  It seems odd that putting a value would be incongruous like that between the different GPI impls.

In any case, I'm positive that `setimmediatevalue()` should set immediately and `test_assign_immediate` attempts to lock that in.

### DIFF
--- a/src/cocotb/handle.py
+++ b/src/cocotb/handle.py
@@ -546,6 +546,7 @@ class _GPISetAction(enum.IntEnum):
     DEPOSIT = 0
     FORCE = 1
     RELEASE = 2
+    NO_DELAY = 3
 
 
 #: The type of the value a :class:`Deposit` or :class:`Force` action contains.
@@ -688,6 +689,8 @@ class ValueObjectBase(SimHandleBase, Generic[ValuePropertyT, ValueSetT]):
             f(*args)
 
         value_, action = _map_action_obj_to_value_action_enum_pair(self, value)
+        if action == _GPISetAction.DEPOSIT:
+            action = _GPISetAction.NO_DELAY
 
         self._set_value(value_, action, _call_now)
 
--- a/src/cocotb/share/include/gpi.h
+++ b/src/cocotb/share/include/gpi.h
@@ -171,6 +171,7 @@ typedef enum gpi_set_action_e {
     GPI_DEPOSIT = 0,
     GPI_FORCE = 1,
     GPI_RELEASE = 2,
+    GPI_NO_DELAY = 3,
 } gpi_set_action_t;
 
 // Functions for iterating over entries of a handle
--- a/src/cocotb/share/lib/fli/FliObjHdl.cpp
+++ b/src/cocotb/share/lib/fli/FliObjHdl.cpp
@@ -245,6 +245,7 @@ int FliEnumObjHdl::set_signal_value(const int32_t value,
     if (m_is_var) {
         switch (action) {
             case GPI_DEPOSIT:
+            case GPI_NO_DELAY:
                 mti_SetVarValue(get_handle<mtiVariableIdT>(),
                                 static_cast<mtiLongT>(value));
                 return 0;
--- a/src/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/src/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -41,6 +41,7 @@ static int releases = 0;
 #include <cocotb_utils.h>    // to_python to_simulator
 #include <py_gpi_logging.h>  // py_gpi_logger_set_level
 
+#include <deque>
 #include <limits>
 #include <type_traits>
 
@@ -199,6 +200,13 @@ struct sim_time {
  *
  */
 int handle_gpi_callback(void *user_data) {
+    static std::deque<void *> deferred_callbacks;
+
+    if (is_python_context) {
+        deferred_callbacks.push_back(user_data);
+        return 0;
+    }
+
     int ret = 0;
     to_python();
     callback_data *cb_data = (callback_data *)user_data;
@@ -257,6 +265,14 @@ out:
 
 err:
     to_simulator();
+    while (!deferred_callbacks.empty()) {
+        auto deferred_user_data = deferred_callbacks.front();
+        deferred_callbacks.pop_front();
+        auto deferred_ret = handle_gpi_callback(deferred_user_data);
+        if (deferred_ret) {
+            ret = deferred_ret;
+        }
+    }
     return ret;
 }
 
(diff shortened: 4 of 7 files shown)


toddstrader:
So this breaks `test_timing_triggers.test_writes_have_taken_effect_after_readwrite`.  I believe the timeline is:

- `write_manually` is started soon
- `in_data` is assigned via `.value` (this is deferred for `_do_writes` to handle later)
- `write_manually` starts up and awaits `ReadWrite`
- that causes `_do_writes` to actually put the value, but in the case of the VPI it is with `vpiInertialDelay` which means it won't actually be applied until the simulator's scheduler moves forward
- `write_manually` wakes up and assigns `in_data` with `setimmediatevalue` which with this PR is put with `vpiNoDelay`
- `in_data` is immediately updated
- we await `ReadOnly()`
- this causes the simulator's scheduler to march forward which causes the `vpiInertialDelay` put to happen
- the assert sees the `.value` value (3) instead of the `setimmediatevalue` one (2)

The test seems right to me and I think this PR makes `setimmediatevalue` via the VPI work as advertised.  This leads me back to questioning `vpiInertialDelay`.  Are we sure we want that?  It seems like we've got two stages of assignment deferring which is what's causing this test to fail now:
- `_do_writes()`
- `vpiInertialDelay`

marlonjames:
Add a test that creates a new task waiting on `Edge(signal)` and then call `signal.setimmediatevalue()` in another task, to make sure that works.
I would expect the new test to hit the icarus re-entrancy problem.

toddstrader:
Thanks @marlonjames .  Also, it's not just an icarus problem.  Xcelium also gives me `FATAL: We are calling up again`.

toddstrader:
I believe the latest commit addresses the reentrancy problem (at least it fixes my test for icarus and xcelium).  I can split that off into another PR if desired or keep it here.  Still need to dig into why `test_writes_have_taken_effect_after_readwrite` is busted . . .

toddstrader:
Scratch that ^ it's `issue_768_*` I need to dig into more (for icarus).  `test_writes_have_taken_effect_after_readwrite` is still failing, but is understood and explained above.  Still curious if people think that the failure of `test_writes_have_taken_effect_after_readwrite` demonstrates that `vpiInertialDelay` is a problem.

marlonjames:
It probably makes sense to move the re-entrancy handling to a different PR.
It shouldn't be recursive, it needs to handle a callbacks from other threads, and I think we need a little more discussion on which layer to queue the pending callbacks.

toddstrader:
> It probably makes sense to move the re-entrancy handling to a different PR.
Sounds good.  I'd like to keep them combined for now because I need the `setimmediatevalue` change to test the reentrancy change and I need GH's CI to see all the simulator results.  But I can break them up once this is all looking good.

The 768 tests are correctly applying the value to `stream_in_data` via `setimmediatevalue`, but once the `Timer` is awaited that value can no longer be observed.  I've also noticed that if I await seemingly anything at the beginning of the test (e.g. `ReadWrite`, `Timer`) I don't run into this issue.  And I only see this with Icarus.  I've tried to reproduce this in a stand-alone test to see if it's an Icarus bug but I can't get the same behavior there.  I've tried doing the `vpiNoDelay` assignment from a `cbStartOfSimulation` callback and from a systf function via `initial` but neither show this issue.

I'll see if I can figure out what the difference is, but wanted to post here to see if this sounded familiar to anyone.

marlonjames:
It may have to do with which simulation region of the timestep the `cbStartOfSimulation` callback is executed.
If that occurs before whatever sets the initial value of `stream_in_data`, it would get overridden.

ktbarrett:
We would definitely want to handle the immediate callback issue at the GPI implementation level (or even push the fix to the simulator if possible) and not in the Python wrapper. So I'm all for deferring that fix for another PR.

toddstrader:
I lied.  I can make the `768` issue happen with either `cbStartOfSimulation` or `cbAfterDelay` with a delay of zero (I believe this is what cocotb is doing for Icarus).  I'll post a reproducer for Icarus, but we'll likely need a band-aid in the meantime.

toddstrader:
> how quickly they can get a fix out
Answer: zero time.  It was already fixed.  If I re-run the `768` tests against `HEAD` of Icarus with this change, they now pass.

toddstrader:
> we shouldn't be using vpiInertialDelay at all

I ran `nox` on this branch with both Icarus (`HEAD`) and Xcelium without `vpiInertialDelay` and everything passed.  So I'm increasingly pro this idea.

imphil:
> > how quickly they can get a fix out
>
> Answer: zero time. 

We're talking about setimmediate, after all.

ktbarrett:
Putting values in VHPI and FLI are *always* inertial, so `vpiInternalDelay` is more consistent *if* the effect is visible to the user. Ideally it should not be. If we are going to jump on to no delay only, we need to test out a way of setting value immediately in the VHPI and FLI as well. I think the easiest way is to Force then immediately Release.

ktbarrett:
> I think the easiest way is to Force then immediately Release.

This doesn't work. Even `Force`s are inertial. Is there anything analagous to `vpiNoDelay` in the VHPI @marlonjames?

ktbarrett:
After further investigation I'm leaning more towards not using `vpiNoDelay` and removing `setimmediatevalue` unless there is some path forward for consistent behavior.

Currently `setimmediatevalue`'s only value is that because `_write_calls` is deferred to the beginning of `ReadWrite` and writes are inertial, `setimmediatevalue` sets are see at the beginning of `ReadWrite` while normal writes are not. I think the current `setimmediatevalue` behavior should be the default, which we can accomplish along with a performance benefit by changing when the `_write_calls` are applied as mentioned in https://github.com/cocotb/cocotb/issues/1024#issuecomment-2011289431.

ktbarrett:
I'm actually fine with this approach. It just needs to pass :)

toddstrader:
I'll revise this once #3875 lands.  Really wish GitHub allowed stacked PRs . . .

ktbarrett:
@toddstrader you could rebase onto my branch and just write "depends on #3875". We don't don't do merge commits here so, you *might* need to rebase if that other commit changes a bunch.

ktbarrett:
I'll commandeer this and get it back to working after #3861 is in.

ktbarrett:
Closing in favor of #4068.

PR #3827, opened 2024-04-05
Verilator: Use latest release in CI

marlonjames:
Also add v5.022 (minimum supported version) to the extended test suite.

### DIFF
--- a/.github/generate-envs.py
+++ b/.github/generate-envs.py
@@ -173,7 +173,7 @@ ENVS = [
     {
         "lang": "verilog",
         "sim": "verilator",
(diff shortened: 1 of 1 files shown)


marlonjames:
Hmm, `test_defaultless_parameter` can't find root handle. This looks like a regression from 5.022.
@AndrewNolte any idea on whether your VPI changes could have affected this?

ktbarrett (approved):

ktbarrett:
I don't get why, but it didn't like the totally equivalent changes that were there, so I just squashed and pushed it up.

PR #3847, opened 2024-04-18
Change how handle writes are applied 

ktbarrett:
xref #1024. This applies handle writes using inertial delays when *exiting* the current GPI callback, rather than in ReadWrite. This makes it so that cocotb writes in `Timer`/`NextSimTime` callbacks are seen in `ReadWrite` callbacks, and value change triggers triggered by writes in `Timer` are seen delta=1 rather than in delta=2 (because of the previous wait for `ReadWrite` first). This should reduce the number of times Python must be entered when writing signals.

Additionally, the slight refactoring to the scheduler was done to canonicalize how events that cause writes to be applied are entered. This seems to lead to a noticeable speed up in some workloads.

## Current Issues

- [ ] Verilator doesn't support `vpiInertialDelay`. This means that writes end up affecting the current eval step. Either it needs to be updated to add support, or we can maybe do something in verilator.cpp
- [ ] Icarus's `vpiInertialDelay` isn't inertial... this should just be fixed.

### DIFF
--- a/src/cocotb/_scheduler.py
+++ b/src/cocotb/_scheduler.py
@@ -260,25 +260,6 @@ class Scheduler:
 
         self._current_task = None
 
-        self._is_reacting = False
-
-        self._write_task = None
-        self._writes_pending = Event()
-
-    async def _do_writes(self):
-        """An internal task that performs pending writes"""
-        while True:
-            await self._writes_pending.wait()
-            if self._mode != Scheduler._MODE_NORMAL:
-                await self._next_time_step
-
-            await self._read_write
-
-            while self._write_calls:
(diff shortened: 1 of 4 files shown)


ktbarrett:
Always using VPI on xcelium is causing issues with the VHDL test now. Riviera VHDL is failing for the same reasons that Icarus is.

ktbarrett:
Opening for posterity. This doesn't work as many simulators have different ideas of when writes should be applied (`test_writes_dont_update_hdl_this_delta` fails). Maybe we can try this again after these inconsistencies are resolved, but I'm not wasting any more time on it now.

PR #3853, opened 2024-04-20
Test better error messages when Makefiles fail

ktbarrett:
We get error messages when tests fail in cocotb, but not if the Makefile fails. This should help.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,9 +40,9 @@ clean:
 
 .PHONY: do_tests
 do_tests::
-	$(MAKE) -k -C tests
+	$(MAKE) -C tests
 do_tests::
-	$(MAKE) -k -C examples
+	$(MAKE) -C examples
 
 # For Jenkins we use the exit code to detect compile errors or catastrophic
 # failures and the XML to track test results
--- a/noxfile.py
+++ b/noxfile.py
@@ -188,7 +188,7 @@ def dev_test_sim(
         coverage_file.unlink()
 
     session.log(f"Running 'make test' against a simulator {config_str}")
-    session.run("make", "test", external=True, env=env)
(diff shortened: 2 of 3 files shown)


noxfile.py
```diff
@@ -188,7 +188,7 @@ def dev_test_sim(
         coverage_file.unlink()
 
     session.log(f"Running 'make test' against a simulator {config_str}")
-    session.run("make", "test", external=True, env=env)
+    session.run("make", "-k", "test", external=True, env=env)
```
cmarqu, line 191:
```suggestion
    session.run("make", "--keep-going", "test", external=True, env=env)
```
maybe? We require GNU make in the docs, so it should be available everywhere.

Makefile
```diff
@@ -40,9 +40,9 @@ clean:
 
 .PHONY: do_tests
 do_tests::
-	$(MAKE) -k -C tests
+	$(MAKE) -C tests
```
cmarqu, line 43:
```suggestion
	$(MAKE) --directory=tests
```

Makefile
```diff
@@ -40,9 +40,9 @@ clean:
...
 do_tests::
-	$(MAKE) -k -C examples
+	$(MAKE) -C examples
```
cmarqu, line 45:
```suggestion
	$(MAKE) --directory=examples
```

ktbarrett:
[Errors](https://github.com/cocotb/cocotb/actions/runs/8767742529).

ktbarrett:
I know the fully spelled names are clearer, but the short ones are standard a pretty well understood. Also I'm going to rm the Makefiles and do everything with nox soon.

cmarqu (approved):

PR #3860, opened 2024-04-23
Conditional scope negative test cases

AndrewNolte:
I've tested this on Xcelium 23.09_004, and Xcelium segfaults on the negative branch. I've seen other behavior where it would point to another spot in the hierarchy. I'm curious to see what other simulators will do


ktbarrett (changes requested):
Unrelated (and incorrect, we don't want to ignore all dotfiles) .gitignore changes seem to be checked in.

ktbarrett (approved):

PR #3861, opened 2024-04-23
Fix Verilator v5.026 by implementing inertial writes

toddstrader:
Pending https://github.com/verilator/verilator/pull/5065 and probably a Verilator release due to the API change.

This doesn't appear to do much for `test_matrix_multiplier_verilator` but it's made noticeable performance improvements for us locally (think a monitor coro which doesn't touch the DUT).

- [x] Update docs to move minimum supported Verilator version to v5.026.

Depends upon #4080: either the fix, or changing the test.

### DIFF
--- a/.github/generate-envs.py
+++ b/.github/generate-envs.py
@@ -173,7 +173,7 @@ ENVS = [
     {
         "lang": "verilog",
         "sim": "verilator",
-        "sim-version": "v5.024",  # Latest release version.
+        "sim-version": "v5.028",  # Latest release version.
         # Needs 22.04 for newer GCC with C++ coroutine support used with --timing mode
         "os": "ubuntu-22.04",
         "python-version": "3.8",
@@ -191,7 +191,7 @@ ENVS = [
     {
         "lang": "verilog",
         "sim": "verilator",
-        "sim-version": "v5.022",  # Minimum supported version.
+        "sim-version": "v5.026",  # Minimum supported version.
         # Needs 22.04 for newer GCC with C++ coroutine support used with --timing mode
         "os": "ubuntu-22.04",
         "python-version": "3.8",
@@ -216,7 +216,7 @@ ENVS = [
         "python-version": "3.8",
         "group": "experimental",
     },
-    # Verilator macOS
+    # Verilator macOS HEAD
     {
         "lang": "verilog",
         "sim": "verilator",
@@ -229,10 +229,11 @@ ENVS = [
     {
         "lang": "verilog",
         "sim": "verilator",
-        "sim-version": "v5.024",
+        "sim-version": "v5.028",
         "os": "macos-13",
         "python-version": "3.8",
         "group": "ci",
+        "may-fail": True,  # verilator/verilator#5404
     },
     # Icarus windows from source
     {
--- a/src/cocotb/share/lib/verilator/verilator.cpp
+++ b/src/cocotb/share/lib/verilator/verilator.cpp
@@ -97,6 +97,7 @@ int main(int argc, char** argv) {
 
     vlog_startup_routines_bootstrap();
     VerilatedVpi::callCbs(cbStartOfSimulation);
+    settle_value_callbacks();
 
 #if VM_TRACE
 #if VM_TRACE_FST
@@ -113,32 +114,22 @@ int main(int argc, char** argv) {
 #endif
 
     while (!Verilated::gotFinish()) {
-        // Call registered timed callbacks (e.g. clock timer)
-        // These are called at the beginning of the time step
-        // before the iterative regions (IEEE 1800-2012 4.4.1)
-        VerilatedVpi::callTimedCbs();
+        do {
+            // We must evaluate whole design until we process all 'events' for
+            // this time step
+            do {
+                top->eval_step();
+                VerilatedVpi::clearEvalNeeded();
+                VerilatedVpi::doInertialPuts();
+                settle_value_callbacks();
+            } while (VerilatedVpi::evalNeeded());
 
-        // Call Value Change callbacks triggered by Timer callbacks
-        // These can modify signal values
-        settle_value_callbacks();
+            // Run ReadWrite callback as we are done processing this eval step
+            VerilatedVpi::callCbs(cbReadWriteSynch);
+            VerilatedVpi::doInertialPuts();
+            settle_value_callbacks();
+        } while (VerilatedVpi::evalNeeded());
 
-        // We must evaluate whole design until we process all 'events'
-        bool again = true;
-        while (again) {
-            // Evaluate design
-            top->eval_step();
-
-            // Call Value Change callbacks triggered by eval()
-            // These can modify signal values
-            again = settle_value_callbacks();
-
-            // Call registered ReadWrite callbacks
-            again |= VerilatedVpi::callCbs(cbReadWriteSynch);
-
-            // Call Value Change callbacks triggered by ReadWrite callbacks
-            // These can modify signal values
-            again |= settle_value_callbacks();
-        }
         top->eval_end_step();
 
         // Call ReadOnly callbacks
@@ -169,9 +160,12 @@ int main(int argc, char** argv) {
         // It should be called in simulation cycle before everything else
         // but not on first cycle
         VerilatedVpi::callCbs(cbNextSimTime);
+        settle_value_callbacks();
 
-        // Call Value Change callbacks triggered by NextTimeStep callbacks
-        // These can modify signal values
+        // Call registered timed callbacks (e.g. clock timer)
+        // These are called at the beginning of the time step
+        // before the iterative regions (IEEE 1800-2012 4.4.1)
+        VerilatedVpi::callTimedCbs();
         settle_value_callbacks();
     }
 
--- a/tests/test_cases/test_inertial_writes/inertial_writes_tests.py
+++ b/tests/test_cases/test_inertial_writes/inertial_writes_tests.py
@@ -10,7 +10,6 @@ from cocotb.triggers import (
     ReadOnly,
     ReadWrite,
     RisingEdge,
-    SimTimeoutError,
     Timer,
     with_timeout,
 )
@@ -26,11 +25,7 @@ simulator_test = "COCOTB_SIMULATOR_TEST" in os.environ
 riviera_vhpi_trust_inertial = SIM_NAME.startswith("riviera") and vhdl and trust_inertial
 
 
-# Verilator < v5.026 only does vpiNoDelay writes.
 @cocotb.test(
-    expect_error=SimTimeoutError
-    if (SIM_NAME.startswith("verilator") and trust_inertial)
-    else (),
     skip=riviera_vhpi_trust_inertial and not simulator_test,
 )
 async def test_writes_on_timer_seen_on_edge(dut):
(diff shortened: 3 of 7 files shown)


ktbarrett:
I have commandeered this PR.

src/cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -113,32 +114,27 @@ int main(int argc, char** argv) {
...
-            // Call Value Change callbacks triggered by eval()
-            // These can modify signal values
-            again = settle_value_callbacks();
+            // Call Value Change callbacks
+            // These can modify signal values so we loop
+            // until there are no more changes
+            bool again;
+            do {
+                VerilatedVpi::doInertialPuts();
+                again = VerilatedVpi::callValueCbs();
```
ktbarrett, line 129: Hmmmm... Swapping these actually fixes both test errors. What I don't understand is why...

ktbarrett: I see... 

1. A timer callback (the clock) wakes up before eval. Writes are queued. 
2. Eval occurs for the beginning of time step / any timing processes.
3. Writes are flushed (the clock change)
4. Value change callbacks (`RisingEdge(clk)`) are run. Writes are queued.
5. If any value change callbacks were run (they were), the writes are applied and the loop runs again (!!! without an eval !!!).

So the issue is the clock writes are applied, `RisingEdge(clk)` returns which queues writes, those writes are applied immediately.

What we need is:
```python
while True:
    doTimedCbs()
    eval()
    while True:
        doInertialPuts()
        doValueChangeCbs()
        if not evalNeeded():
            break
        eval()
    doReadWriteCbs()
```

ktbarrett: And technically the first eval is only needed if we are in `--timing` mode, because otherwise there is no time-based events in the HDL to wake up.

ktbarrett: Turns out there was not actually a problem. The inertial_writes_tests failures were due to me getting mixed up, and the test_dff failure was just due to how Verilator works after the change.

README.md
```diff
@@ -63,40 +63,46 @@ An example of a simple randomized cocotb testbench:
...
+    # Run the driver coroutine concurrently to the monitoring and checking logic below.
+    cocotb.start_soon(drive_input())
+
+    # Wait 1 clock cycle for input to propagate to output.
     await RisingEdge(dut.clk)
-    assert dut.q.value == expected_val, "output q was incorrect on the last cycle"
+
+    for expected_val in expecteds:
+        await RisingEdge(dut.clk)  # Synchronize with clock, then...
+        await ReadOnly()  # Wait for all signal changes to quiesce.
```
cmarqu, line 102:
```suggestion
        await ReadOnly()  # Wait for all signal changes to settle.
```
Use a simpler term for non-native speakers.

examples/simple_dff/test_dff.py
```diff
@@ -7,48 +7,44 @@
...
+    # Run the driver coroutine concurrently to the monitoring and checking logic below.
+    cocotb.start_soon(drive_input())
+
+    # Wait 1 clock cycle for input to propagate to output.
     await RisingEdge(dut.clk)
-    assert dut.q.value == expected_val, "output q was incorrect on the last cycle"
+
+    for expected_val in expecteds:
+        await RisingEdge(dut.clk)  # Synchronize with clock, then...
+        await ReadOnly()  # Wait for all signal changes to quiesce.
```
cmarqu, line 44:
```suggestion
        await ReadOnly()  # Wait for all signal changes to settle.
```
(same)

ktbarrett:
The Mac OS test is fixed by verilator/verilator#5300, but we won't see it until v5.028 is release (which will probably be before 2.0). I think I'll just add a "may_fail" to the CI version with an issue to update to v5.028 and remove this flag.

cmarqu (approved):

tests/test_cases/test_inertial_writes/inertial_writes_tests.py
```diff
@@ -51,8 +47,6 @@ async def test_writes_on_timer_seen_on_edge(dut):
     expect_fail = False
 elif SIM_NAME.startswith(("riviera", "modelsim")) and vhdl:
     expect_fail = False
-elif SIM_NAME.startswith("verilator"):
-    expect_fail = False
```
ktbarrett, line 55: Verilator now acts like most other simulators in this mode, where because writes are cached in Python until ReadWrite, then applied inertially, you can't read back what you just wrote.

tests/test_cases/test_inertial_writes/inertial_writes_tests.py
```diff
@@ -28,9 +27,6 @@
 
 # Verilator < v5.026 only does vpiNoDelay writes.
 @cocotb.test(
-    expect_error=SimTimeoutError
-    if (SIM_NAME.startswith("verilator") and trust_inertial)
-    else (),
```
ktbarrett, line 33: trust inertial mode now works in Verilator.

ktbarrett:
Well it seems the problem is now incorrect trigger cleanup. Reusing the `Timer` triggers in the `"py"` impl of `Clock` causes an infinite loop in test_3316_c.

ktbarrett:
Turns out `VM_TIMING` wasn't being exported as a define during compilation of `verilator.cpp`. Once that is set correctly with verilator/verilator#5383, the optimization works. But it works fine enough without it for now.

tests/test_cases/test_inertial_writes/inertial_writes_tests.py
```diff
@@ -28,9 +27,6 @@
 
 # Verilator < v5.026 only does vpiNoDelay writes.
```
imphil, line 28: Remove this comment as well.

.github/generate-envs.py
```diff
@@ -229,10 +220,11 @@
     {
         "lang": "verilog",
         "sim": "verilator",
-        "sim-version": "v5.024",
+        "sim-version": "v5.026",
         "os": "macos-13",
         "python-version": "3.8",
         "group": "ci",
+        "may-fail": True,  # TODO remove once this is bumped to v5.028
```
imphil, line 227: Can you link to a PR/add a reason?

README.md
```diff
@@ -63,40 +63,46 @@ An example of a simple randomized cocotb testbench:
...
-    expected_val = 0  # Matches initial input value
-    for i in range(10):
-        val = random.randint(0, 1)
-        dut.d.value = val  # Assign the random value val to the input port d
-        await RisingEdge(dut.clk)
-        assert dut.q.value == expected_val, f"output q was incorrect on the {i}th cycle"
-        expected_val = val # Save random value for next RisingEdge
-
-    # Check the final input on the next clock
+    # generate stimulus
```
imphil, line 83:
```suggestion
    # Generate 10 random 0/1 values as stimulus.
```

README.md
```diff
@@ -63,40 +63,46 @@ An example of a simple randomized cocotb testbench:
...
+    stimulus = [random.randint(0, 1) for _ in range(10)]
+
+    # calculated expected outputs
```
imphil, line 86:
```suggestion
    # Expected outputs: identical to the input.
```

README.md
```diff
@@ -63,40 +63,46 @@ An example of a simple randomized cocotb testbench:
...
+    expecteds = [val for val in stimulus]
```
imphil, line 87: "expecteds" is a strange word, maybe just "inputs" (instead of stimulus) "outputs_expected"?

Either way, just copy input to output: `expecteds = stimulus`

README.md
```diff
@@ -63,40 +63,46 @@ An example of a simple randomized cocotb testbench:
...
+    # Wait 1 clock cycle for input to propagate to output.
     await RisingEdge(dut.clk)
-    assert dut.q.value == expected_val, "output q was incorrect on the last cycle"
+
+    for expected_val in expecteds:
+        await RisingEdge(dut.clk)  # Synchronize with clock, then...
+        await ReadOnly()  # Wait for all signal changes to settle.
+        assert (
+            dut.q.value == expected_val
+        )  # Check the actual output against the expected.
```
imphil, line 105:
```suggestion
        # Check the actual output against the expected.
        assert dut.q.value == expected_val
```

imphil:
I'm confused ... I thought I wrote a comment about the x-initialization commit in this PR, but it seems to be lost?! Either way, that's what I wrote:

What do we need the `Support X initialization` commit for? It's a bit out of scope in this PR, and should be already supported by passing `+verilator+rand+reset` (https://verilator.org/guide/latest/exe_sim.html#cmdoption-arg-verilator-rand-reset-value).

ktbarrett:
Ah. Partially this commit is upstreaming things we have locally, which aren't always up-to-date with the newest Verilator features (being blocked by cocotb support rn). I'll drop that.

README.md
```diff
@@ -63,40 +63,45 @@ An example of a simple randomized cocotb testbench:
...
-        dut.d.value = val  # Assign the random value val to the input port d
-        await RisingEdge(dut.clk)
-        assert dut.q.value == expected_val, f"output q was incorrect on the {i}th cycle"
-        expected_val = val # Save random value for next RisingEdge
-
-    # Check the final input on the next clock
+    # Generate 10 random 0/1 values as inputs.
+    inputs = [random.randint(0, 1) for _ in range(10)]
+
+    # Expected outputs (identical to inputs)
```
imphil, line 86: Since you added a trailing `.` to most comments:
```suggestion
    # Expected outputs (identical to inputs).
```

imphil (approved):

src/cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -36,18 +36,12 @@ extern "C" {
...
 
-static inline bool settle_value_callbacks() {
-    bool cbs_called, again;
-
-    // Call Value Change callbacks
-    // These can modify signal values so we loop
-    // until there are no more changes
-    cbs_called = again = VerilatedVpi::callValueCbs();
-    while (again) {
+static void settle_value_callbacks() {
```
ktbarrett, line 39: I guess this is almost equivalent to the old code, so no need to change it

src/cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -113,32 +108,23 @@ int main(int argc, char** argv) {
...
+            do {
+                top->eval_step();
+                VerilatedVpi::clearEvalNeeded();
+                VerilatedVpi::doInertialPuts();
+                settle_value_callbacks();
+            } while (VerilatedVpi::evalNeeded());
+
+            // Run ReadWrite callback as we are done processing this eval step
+            VerilatedVpi::callCbs(cbReadWriteSynch);
+            settle_value_callbacks();
```
ktbarrett, line 123:
```suggestion
```

marlonjames (approved):

src/cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -113,32 +114,22 @@ int main(int argc, char** argv) {
...
+            // this time step
+            do {
+                top->eval_step();
+                VerilatedVpi::clearEvalNeeded();
+                VerilatedVpi::doInertialPuts();
+                settle_value_callbacks();
+            } while (VerilatedVpi::evalNeeded());
+
+            // Run ReadWrite callback as we are done processing this eval step
+            VerilatedVpi::callCbs(cbReadWriteSynch);
```
ktbarrett, line 128:
```suggestion
            VerilatedVpi::callCbs(cbReadWriteSynch);
            settle_value_callbacks();
```

ktbarrett:
MacOS CI is still failing https://github.com/verilator/verilator/issues/5404.

imphil:
:partying_face:

PR #3864, opened 2024-04-26
Small cleanup to cocotb.results

ktbarrett:
This simple set of fixes has been hanging out locally for a while now, so let's just get it merged.


marlonjames (approved):

marlonjames (commented):
Newsfragment?

ktbarrett:
Added newsfragments. We aren't providing an alternative (basically they were being made private, but also removed), so I kept the message short and sweet.

marlonjames (approved):

PR #3868, opened 2024-04-29
Throw an exception with helpful message if libpython is missing

cjmeyer:
Calls to find_libpython.find_libpython() may return an empty string or None if libpython was not properly installed. This was causing a TypeError exception to be thrown with a non-obvious error message. The value returned by find_libpython.find_libpython() is now checked and if not found a more helpful error message is provided.

### DIFF
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -133,7 +133,10 @@ class Simulator(abc.ABC):
             self.env[e] = os.environ[e]
 
         if "LIBPYTHON_LOC" not in self.env:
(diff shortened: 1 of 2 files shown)


marlonjames:
Fix the pre-commit linting failures and we'll enable the rest of the CI run.
Info here if needed https://github.com/cocotb/cocotb/blob/master/CONTRIBUTING.md#setting-up-a-development-environment

marlonjames (approved):

ktbarrett:
Should probably backport this to 1.9 as well.

PR #3872, opened 2024-04-30
update pre-commit for stable/1.9

ktbarrett:
Maybe this will make pre-commit.ci happy on this branch.


ktbarrett:
I'm guessing the issue is that the version of flake8 doesn't support the version of Python used in pre-commit.ci.

imphil (approved):

PR #3874, opened 2024-05-01
Throw `CancelledError` into cancelled tasks

### LINKED ISSUE #2679: Add `cancel()` to `RunningTask` and deprecate `kill()`
Continuing the theme of providing the new interfaces before 2.0.
For now we could keep it the same, without using `CancelledError`.

Potentially, `kill()` could be made private and used by the scheduler, whereby the task is cancelled even if it tries to suppress the exception.

ktbarrett:
Closes #2679. Implements throwing `CancelledError` into tasks when `task.cancel` is called. Like `asyncio`, cancelling a task is now a *scheduled* event, so we may need `await NullTrigger()` after cancels to see them cancelled.

* This works by using `cocotb._scheduler._resume_upon_coro` to schedule the cancellation of a task using a new `NullTrigger`.
* There may be more than one trigger that can wake a task, so task now supports multiple triggers and cleanup in the scheduler handles this case.
* Cancellation is a state of the task, so no matter which trigger causes the task to wake next, it will cancel it.
* Future might support `uncancel`, though this will not work the same as asyncio's.

I made the hard rule that after a cancellation occurs, the Task is done. In asyncio, a Task can squash a CancelledError and continue, leaving it in state where it is not `cancelled()` or `done()`, but unscheduled. awaiting on such a Task will block indefinitely because awaits on Tasks only unblock when the Task becomes `done()`. This decision avoids this problem.

The intended use case of `CancelledError` is to catch it, do clean up, then re-raise it.
```python
async def thingy():
    try:
        await Timer(10, 'ns')
    except CancelledError:
        print("Clean up, clean up, everybody do your share.")
        raise
```

Unintended things can happen:
* The code throws an exception: The task has finished, this is safe.
* The code exits normally: This *can* be safe (did they use `return` in the error handler?), but can also mean the user forgot to re-raise and the task just so happens for now to have ended before another `await`.
* The code reaches another `await`: This is suppressing the cancellation. It might be intended, it might not be, either way we don't support it so we error.

In all cases where the above issues can occur, we force finish the Task with `CancellationError`, denoting there was an issue during the process of cancellation.

### TODO
- [x] newsfrag
- [x] documentation
- [x] implement `cancel()` on test end
- [ ] deprecate `kill()`
- [x] consider only supporting 1 trigger per task and having cancellation force over the existing one (unpriming the old one). There might be other reasons for supporting more than one trigger per task, so I'm inclined to leave it be until we are sure.
- [ ] Change how test end conditions are reported. Don't force the outcome of the test task to something unrelated. Yuck. Also this causes the CancelledError to not be thrown into the test task since it's "done".
- [ ] Add way to report additional "secondary" failures during task cleanup to the regression manager
- [ ] Use secondary failures to fail tests (expect_error only applies to "primary" test end conditions)
- [ ] Add feature to test decorator to test expected errors during cleanup


src/cocotb/task.py
```diff
@@ -114,50 +121,85 @@ def __repr__(self) -> str:
         repr_string = fmt.format(
             name=self.__name__,
             coro=coro_name,
-            trigger=self._trigger,
+            triggers=self._triggers,
             outcome=self._outcome,
```
ktbarrett, line 125:
```suggestion
            outcome=self._outcome,
            cancel_exc=self.cancel_exc,
```

ktbarrett:
Also have to consider pending cancellations during cleanup.

ktbarrett:
Okay. New task list:

- [ ] Change how test end conditions are reported. Don't force the outcome of the test task to something unrelated. Yuck. Also this causes the CancelledError to not be thrown into the test task since it's "done".
- [ ] Add way to report additional "secondary" failures during task cleanup to the regression manager
- [ ] Use secondary failures to fail tests (expect_error only applies to "primary" test end conditions)
- [ ] Add feature to test decorator to test expected errors during cleanup

ktbarrett:
Also, I expect this to have a radically negative performance impact on our serial regressions =/ Every test, no matter how small, now does extra work.

src/cocotb/task.py
```diff
@@ -91,9 +95,13 @@ def __repr__(self) -> str:
...
         if cocotb._scheduler._current_task is self:
             fmt = "<{name} running coro={coro}()>"
+        elif self.cancelling():
+            fmt = "<{name} cancelling coro={coro}() cancel_exc={cancel_exc}>"
+        elif self.cancelled():
+            fmt = "<{name} cancelled coro={coro}() outcome={outcome} cancel_exc={cancel_exc}>"
         elif self.done():
             fmt = "<{name} finished coro={coro}() outcome={outcome}>"
-        elif self._trigger is not None:
+        elif self._trigger:
```
ktbarrett, line 104:
```suggestion
        elif self._trigger is not None:
```

src/cocotb/task.py
```diff
@@ -116,48 +124,79 @@ def __repr__(self) -> str:
...
+                    CancellationError(
+                        "Task was cancelled, but continued running. Did you forget to re-raise the CancelledError?",
+                        Value(None),
+                    )
+                )
 
-    def throw(self, exc: BaseException) -> Any:
-        return self._coro.throw(exc)
+    def kill(self) -> None:
+        """Stop a Task without throwing a :exc:`asyncio.CancelledError`.
```
ktbarrett, line 190:
```suggestion
        """Stop a Task immediately without throwing a :exc:`asyncio.CancelledError`.
```

ktbarrett:
Status update: I'm trying to figure out how to best clean up the scheduler's `_unschedule`, `_abort_test`, and other cleanup functions as there isn't a consistent way that tests are ended and it's making only setting the test outcome once quite difficult.

marlonjames:
I was thinking of having the test's `Task` include a private done callback to cleanup and go to the next test.
Whenever a test needs to end, collect the callbacks as you cancel each task, then run them after all of the tasks are cancelled.

ktbarrett:
This is going to be totally reimplemented when other PRs are in.

PR #3878, opened 2024-05-03
More 1.9 changes

### LINKED ISSUE #3845: Allow Logic for BinaryValue.__eq__() in cocotb 1.8
As BinaryValue is deprecated and will be removed in 2.0, but is is still the return type for handler.value

It would be practical to allow

`dut.mysignal == Logic('X')`

in cocotb 1.8 so we can already stop using BinaryValue and avoid more porting work when 2.0 arrives

ktbarrett:
Update `FutureWarning` when Tasks throw an exception. The current behavior is the opposite of what's going to happen  in 2.0. Also close #3845. Also backports #3879 and #3868.

- [x] newsfrag
- [x] update tests


ktbarrett:
Waiting on outcome of whether the python installation requirements are correct.

cocotb/binary.py
```diff
@@ -562,14 +566,16 @@ def __bool__(self):
...
+        if isinstance(other, (BinaryValue, LogicArray)):
+            return self.binstr == other.binstr
+        elif isinstance(other, int):
+            return self.integer == other
+        elif isinstance(other, str):
+            return self.binstr == other
+        elif isinstance(other, Logic):
+            return self.binstr == str(other)
+        else:
+            return False
```
ktbarrett, line 578:
```suggestion
            return NotImplemented
```

cmarqu (approved):

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -110,91 +102,36 @@ async def test_timer_with_rational_units(dut):
...
-    global exited
-    await RisingEdge(dut.clk)
-    await ReadOnly()
-    await ReadWrite()
-    exited = True
+@cocotb.test(expect_error=Exception)
+async def test_cached_write_in_readonly(dut):
+    """Test doing invalid sim operation"""
 
+    # only here because Riviera with VHPI looses its mind otherwise.
```
marlonjames, line 109: Feel free to correct these or not.
```suggestion
    # only here because Riviera with VHPI loses its mind otherwise.
```

marlonjames (approved):

PR #3890, opened 2024-05-10
Xcelium cond scope segfault fix

AndrewNolte:
This fixes the segfault issue with Xcelium, and makes it match other simulators with respect to not returning invalid scopes


AndrewNolte:
Cadence segfault ticket: https://support.cadence.com/apex/COSCaseDetailPage?CaseID=500PP00000BCxhC

ktbarrett (approved):

Issue #3896, opened 2024-05-16
Verilator v5.026+ support


ktbarrett:
The in development version of Verilator (v5.025) introduces support for `vpiInertialDelay`, however it is up to the user to decide when to apply those writes. This means we are going to have to make changes to verilator.cpp. While I see those changes coming in the near future, cocotb will be broken with verilator until some fix is released.

If 1.9 is released soon, I suggest we introduce an ifdef for Verilator in out VPI implementation to use `vpiNoDelay` instead, which was the old behavior. If 1.9 is released soon there will only be a small gap where there is no support. This will also allow us to continue to support older version of Verilator that ignore the delay flag.

For 2.0 we will need to update the verilator.cpp and revert https://github.com/cocotb/cocotb/pull/4016.

ktbarrett:
Closed by #4015. vpiInertialDelay writes are great!

PR #3897, opened 2024-05-16
Support newer Verilator in backwards compatible way

ktbarrett:
xref #3896.

### DIFF
--- a/cocotb/share/lib/vpi/VpiCbHdl.cpp
+++ b/cocotb/share/lib/vpi/VpiCbHdl.cpp
@@ -379,6 +379,14 @@ int VpiSignalObjHdl::set_signal_value(s_vpi_value value_s,
 
     switch (action) {
         case GPI_DEPOSIT:
+#ifdef VERILATOR
+            // vpiNoDelay is the existing behavior in Verilator, as until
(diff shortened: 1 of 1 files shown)


cocotb/share/lib/vpi/VpiCbHdl.cpp
```diff
@@ -379,6 +379,9 @@ int VpiSignalObjHdl::set_signal_value(s_vpi_value value_s,
 
     switch (action) {
         case GPI_DEPOSIT:
+#ifdef VERILATOR
```
marlonjames, line 382: I think a comment here would be good, explaining why it's needed, even if 2.0 will remove it...

ktbarrett:
Strange it isn't covering the new line, cocotb_build_libs.py says the define is VERILATOR?

marlonjames:
Verilator CI isn't being run. You can see it's missing from the output of the generate script.

ktbarrett:
@marlonjames Do we trust to merge this as is, or should we wait until Verilator CI changes are backported?

marlonjames (approved):
I'm okay with this, assuming we don't want to wait and bump the minimum Verilator version for 1.9.

cmarqu (approved):

Issue #1833, opened 2020-05-15
Update CoCoTB to allow test to access constants defined in packages


jwprice100:
When working with VHDL designs, it's common to have constants defined in packages for use across many source files. At the moment, the only way to access those constants is to create copies in the dut so that cocotb tests can access them. While this work around is functional, it's a bit inelegant. It would be nice to able to get a handle to other objects such as packages. I understand this is possible, but might require modifying the C++ back-end.

garmin-mjames:
A `packages` attribute could be added to `simulator` with handles to the package instantiations.

For VHPI, here's an example snippet from the language ref manual showing how to get the packages:
```c
/* if not found in the design hierarchy, search for the signal
in the instantiated packages
*/
itr = vhpi_iterator(vhpiPackInsts, NULL);
if (itr) {
  while (hdl = vhpi_scan(itr)) {
    sigHdl = vhpi_handle_by_name(sigName, hdl);
    if (sigHdl) {
      return sigHdl;
    }
  }
}
```

andrewandrepowell:
Hello,

I plan to work on this issue if no one else is doing so already. I propose the following interface:

```python

@test()
def some_test(dut, packages) -> None:
    SOME_CONSTANT = packages["package_name"].SOME_CONSTANT.value

```

The packages parameter can be an optional argument the user can add to their test to retrieve a dictionary with all the packages inside it.

eric-wieser:
I think making it a function or attribute on the simulator module is probably easier

andrewandrepowell:
I was thinking maybe during the `_initialize_testbench` the packages can all be discovered and stored, that way the user wouldn't have to directly use anything in the simulator module. Are you suggesting only have an interface to the packages from the simulator module only?

ktbarrett:
@andrewandrepowell I agree with Eric. The discovery will have to go through the simulator module anyways. As for pythonland abstraction around that, I'm not a fan of adding it to the test as an optional function. In fact I'm for the opposite (#1827). If we want to support alternative test frameworks putting less requirements on it is better.

garmin-mjames:
One more vote for a method on `simulator`. There is already `simulator.get_root_handle`, so probably `simulator.get_packages`.

I think it makes sense to have each package be a `HierarchyObject`, and we should check each GPI `registered_impls` for the mixed-language case.

I'm not sure what `get_packages` should return, though:
1. tuple of package objects
[...]

PR #3899, opened 2024-05-18
Add support for VHDL packages

nickg:
See issue #1833 and also nickg/nvc#891.

I have only been able to test this with latest NVC master branch.


src/cocotb/__init__.py
```diff
@@ -296,6 +296,19 @@ def _process_plusargs() -> None:
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def __getattr__(self, name):
+        if SIM_NAME == "nvc":
+            # NVC does not preserve the case of identifiers so make
+            # lookups in this container case-insensitive
+            return self.__getattribute__(name.upper())
+        else:
+            super().__getattr__(name)
```
ktbarrett, line 309: I don't know if I like this answer. But I don't really think it's your problem to solve.

VHDL names are case insensitive (christ why...), but in mixed language simulations it's hard to know whether package names are case insensitive or if certain packages come from VHDL or Verilog and whether the case matters. And this seems to be a larger problem for handle names during iteration.

IMO object discovery via iteration with the VHPI should return correctly cased names and this shuldn't be necessary. VHDL does have the concept of object declarations so there is a source of truth for the caseness of every object. IDK why VHPI allows case insensitive operation, it was made at least 20 years after they realized what a bad idea case insensitive languages are.

src/cocotb/share/lib/vhpi/VhpiImpl.cpp
```diff
@@ -465,7 +470,8 @@ GpiObjHdl *VhpiImpl::create_gpi_obj_from_handle(vhpiHandleT new_hdl,
               vhpi_get_str(vhpiKindStrP, query_hdl));
 
     if (gpi_type != GPI_ARRAY && gpi_type != GPI_GENARRAY &&
-        gpi_type != GPI_MODULE && gpi_type != GPI_STRUCTURE) {
+        gpi_type != GPI_MODULE && gpi_type != GPI_STRUCTURE &&
+        gpi_type != GPI_PACKAGE) {
```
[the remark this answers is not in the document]
nickg: I tried replacing it with a switch. Looks clearer to me at least.

nickg:
Doesn't seem to work with any of the commercial simulators but I don't have a license at the moment to debug.

tests/test_cases/test_package/test_package.py
```diff
@@ -79,7 +94,29 @@ async def test_package_access(_) -> None:
     assert pkg2.eleven_int.value == 11
 
 
-@cocotb.test(skip=cocotb.SIM_NAME.lower().startswith("riviera"))
+@cocotb.test(
+    expect_fail=(SIM_NAME.startswith("ghdl") or nvc_pre_1_13 or questa_fli),
```
ktbarrett, line 98:
```suggestion
    expect_fail=(SIM_NAME.startswith(("ghdl", "riviera", "xmsim")) or nvc_pre_1_13 or questa_fli),
```

tests/test_cases/test_package/test_package.py
```diff
@@ -90,3 +127,22 @@ async def test_dollar_unit(dut):
     cocotb.log.info(f"Found $unit as {unit}")
     unit_pkg = getattr(cocotb.packages, unit)
     assert unit_pkg.unit_four_int.value == 4
+
+
+@cocotb.test(
+    expect_fail=(SIM_NAME.startswith("ghdl") or nvc_pre_1_13 or questa_fli),
```
ktbarrett, line 133:
```suggestion
    expect_fail=(SIM_NAME.startswith(("ghdl", "riviera", "xmsim")) or nvc_pre_1_13 or questa_fli),
```

ktbarrett:
Questa FLI passes `test_get_root_handle`, not fails.

Questa VHPI fails `test_package_access_vhdl` and `test_get_root_handle` causes a crash so it should be skipped.

I have no idea why the discovery tests are different now in Riviera VHPI.

tests/test_cases/test_package/test_package.py
```diff
@@ -6,8 +6,27 @@
...
+SIM_NAME = cocotb.SIM_NAME.lower()
+LANGUAGE = os.environ["TOPLEVEL_LANG"].lower().strip()
+
+questa_vhpi = (
+    SIM_NAME.startswith("modelsim") and os.getenv("VHDL_GPI_INTERFACE", "fli") == "vhpi"
+)
+
+nvc_pre_1_16 = SIM_NAME.startswith("nvc") and (
+    NvcVersion(cocotb.SIM_VERSION) < NvcVersion("1.16")
+)
```
ktbarrett, line 29: @nickg If you think NVC will do a 1.15.3 release with the fix, please update this.

src/cocotb/_init.py
```diff
@@ -166,11 +166,9 @@ def _process_packages() -> None:
...
         name = handle._name
 
-        # Icarus doesn't support named access to package objects:
-        # https://github.com/steveicarus/iverilog/issues/1038
-        # so we cannot lazily create handles
-        if cocotb.SIM_NAME == "Icarus Verilog":
-            handle._discover_all()
+        # All VHPI sims have issue with discovering sub-objects,
+        # so we just always iterate to discover all children.
+        handle._discover_all()
```
nickg, line 171: What's the problem here? `vhpi_handle_by_name` doesn't work in packages? I would rather fix this in NVC than have this workaround recursively discover all objects as I build VHPI objects lazily and this would cause a lot of redundant allocations.

tests/test_cases/test_package/test_package.py
nickg: I'm not going to backport this so 1.16 is correct.

src/cocotb_tools/makefiles/simulators/Makefile.nvc
```diff
@@ -37,8 +37,8 @@ analyse: $(VHDL_SOURCES) $(SIM_BUILD) $(CUSTOM_COMPILE_DEPS)
 	$(foreach SOURCES_VAR, $(filter VHDL_SOURCES_%, $(.VARIABLES)), $(check_lib_order))
 
 	$(foreach LIB_VAR,$(VHDL_LIB_ORDER), \
-		$(CMD) $(EXTRA_ARGS) --work=$(LIB_VAR):$(SIM_BUILD)/$(LIB_VAR) -L $(SIM_BUILD) -a $(VHDL_SOURCES_$(LIB_VAR)) $(COMPILE_ARGS) && ) \
-	$(CMD) $(EXTRA_ARGS) --work=$(RTL_LIBRARY):$(SIM_BUILD)/$(RTL_LIBRARY) -L $(SIM_BUILD) -a $(VHDL_SOURCES) $(COMPILE_ARGS)
+		$(CMD) $(EXTRA_ARGS) --work=$(LIB_VAR):$(SIM_BUILD)/$(LIB_VAR) -L $(SIM_BUILD) -a $(VHDL_SOURCES_$(LIB_VAR)) --preserve-case $(COMPILE_ARGS) && ) \
+	$(CMD) $(EXTRA_ARGS) --work=$(RTL_LIBRARY):$(SIM_BUILD)/$(RTL_LIBRARY) -L $(SIM_BUILD) -a $(VHDL_SOURCES) --preserve-case $(COMPILE_ARGS)
```
nickg, line 41: I guess the Python runner needs this too?

src/cocotb/_init.py
ktbarrett: I am not entirely sure. I know for certain this is required for Icarus and Riviera. I need to poke into it more, but ran out of time before passing out last night.

ktbarrett:
Nope =/ Apparently `:show-inheritance:` still tries to list undocumented classes and it doesn't take a list to be explicit about what inheritance to show. I guess I'll just have to explicitly list what classes have members it inherit from to prevent repeating members 100 times.

src/cocotb/__init__.py
ktbarrett: NVC now supports `--preserve-case` to preserve the case of the declaration as discovered name of VHPI simulation objects.

src/cocotb/_init.py
ktbarrett: Looks like only NVC supports getting child handles by name from package objects. Weird...

I'd hate to add all those simulators to the list that needs to `_discover_all` first, and there's probably a better way to do that by delegating to discover_all if a lookup fails, but that's work I don't want to have to do before 2.0. So I guess we will just pessimize those simulators for the 2.0 release and try to fix it before 2.1.

imphil:
@nickg Would you mind splitting out the change that makes cocotb use `--preserve-case` into a separate PR? I think we can get that merged right away.

src/cocotb/_init.py
marlonjames: When I investigated RivieraPRO, I couldn't figure out any way to use `vhpi_handle_by_name` to get objects inside a package. Only iteration would work.

ktbarrett: Well I think the best way for object discovery going forward is to use iteration anyways. We need iteration to ensure we mark objects as ports or enums.

PR #3901, opened 2024-05-19
Fix instructions for extending build flow with Verilator

olofk:
Documentation has not kept up with the code changes that loads the cocotb lib at runtime.

### DIFF
--- a/docs/source/custom_flows.rst
+++ b/docs/source/custom_flows.rst
@@ -38,7 +38,7 @@ Verilator
       --vpi --public-flat-rw --prefix Vtop \
       -LDFLAGS "-Wl,-rpath,$(cocotb-config --lib-dir) \
           -L$(cocotb-config --lib-dir) \
-          -lcocotbvpi_verilator -lgpi -lcocotb -lgpilog -lcocotbutils" \
+          -lcocotbvpi_verilator" \
       $(cocotb-config --share)/lib/verilator/verilator.cpp
 
 * Run Verilator's makefile as follows: ``CPPFLAGS="-std=c++11" make -f Vtop.mk``


cmarqu (approved):

docs/source/custom_flows.rst
```diff
@@ -38,7 +38,7 @@ Verilator
       --vpi --public-flat-rw --prefix Vtop \
       -LDFLAGS "-Wl,-rpath,$(cocotb-config --lib-dir) \
           -L$(cocotb-config --lib-dir) \
-          -lcocotbvpi_verilator -lgpi -lcocotb -lgpilog -lcocotbutils" \
+          -lcocotbvpi_verilator" \
       $(cocotb-config --share)/lib/verilator/verilator.cpp
 
 * Run Verilator's makefile as follows: ``CPPFLAGS="-std=c++11" make -f Vtop.mk``
```
cmarqu, line 44: Maybe not something for this PR, but should this say `-std=c++14` these days with `--timing`?

ktbarrett: I think `timing` needs C++20 coroutines, but I'd consider that not relevant to cocotb. The rest is up to the user depending on how they user Verilator. AFAIK we aren't dependent on timing mode except for our CI.

ktbarrett (approved):

docs/source/custom_flows.rst
ktbarrett: Now that I see what line you are actually commenting on... I think this is the minimum standard to compile `verilator.cpp`. I don't know if this is disastrous recommendation or not. Does it override whatever CPPFLAGS are being used in Verilator's makefile? We can't assume support for greater than C++11 though since we still support RHEL7 (GCC 4.8) until after the 2.0 release.

ktbarrett:
CI is finished but stuck reporting it's success? =/ Gonna ignore that.

PR #3902, opened 2024-05-20
See if scheduling ReadWrite back to back keeps simulator running

ktbarrett:
I don't think this is valid. Let's see what other simulators think.


ktbarrett:
Some do, some don't. Fun.

PR #3777, opened 2024-03-13
Cleanup handle (Part ??? + 1)

ktbarrett:
Some more changes. Depends on #4038.

* Removes `IndexableValueObjectBase`, only `ArrayObject` now support indexing and iteration.

- [x] Fix discovery tests now...

PR #3905, opened 2024-05-22
Packed struct fields

AndrewNolte:
This adds in some of the functionality to access packed struct fields that was removed in https://github.com/cocotb/cocotb/pull/3608

### DIFF
--- a/src/cocotb/handle.py
+++ b/src/cocotb/handle.py
@@ -1272,6 +1272,36 @@ class StringObject(
         return self.value.decode("ascii")
 
 
+class PackedStructObject(LogicObject, HierarchyObject):
+    """
+    Incorperates the functionality of both HierarchyObject and LogicObject
+
+    For getting struct children with a name collision, use handle["name"]
+
+    Since this class lists LogicObject first for inheritance, conflicting methods will call LogicObject's method
+    """
+
+    def __init__(self, handle, path):
+        # Use HierarchyObject because _sub_handles contains SimHandleBase, which is a broader type
+        HierarchyObject.__init__(self, handle, path)
+
+    def __getattr__(self, name: str) -> Any:
+        if name in {"value", "set", "setimmediatevalue", "is_const"}:
+            return LogicObject.__getattribute__(self, name)
+        # private properties will be resolved in HierarchyObject
+        return HierarchyObject.__getattr__(self, name)
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        if name == "value":
+            return LogicObject.__setattr__(self, "value", value)
+
+        return HierarchyObject.__setattr__(self, name, value)
+
+    def __getitem__(self, index: str) -> SimHandleBase:
+        # we can't index into packed structs, so only support HierarchyObject's __getitem__
+        return HierarchyObject.__getitem__(self, index)
+
+
 _ConcreteHandleTypes = Union[
     HierarchyObject,
     HierarchyArrayObject,
@@ -1292,7 +1322,7 @@ _handle2obj: Dict[
 _type2cls: Dict[int, Type[_ConcreteHandleTypes]] = {
     simulator.MODULE: HierarchyObject,
     simulator.STRUCTURE: HierarchyObject,
-    simulator.PACKED_STRUCTURE: LogicObject,
+    simulator.PACKED_STRUCTURE: PackedStructObject,
     simulator.REG: LogicObject,
     simulator.NET: LogicObject,
     simulator.NETARRAY: ArrayObject[Any, ValueObjectBase[Any, Any]],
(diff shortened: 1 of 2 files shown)


tests/test_cases/test_struct/test_struct.py
```diff
@@ -49,6 +52,32 @@ async def test_packed_struct_setting(dut):
...
+
+    await Timer(1000, "ns")
+
+    assert str(dut.my_struct.value) == "101"
+    assert dut.my_struct.val_a.value == 1
+    assert dut.my_struct.val_b.value == 0
+    assert dut.my_struct["value"].value == 1
+
+    assert len(dut.my_struct) == 3
+
```
ktbarrett, line 80: Let's put these in a separate test and set as `expect_error=AttributeError`.

tests/test_cases/test_struct/test_struct.py
```diff
@@ -25,7 +25,10 @@
 )
 async def test_packed_struct_format(dut):
     """Test that the correct objects are returned for a struct"""
-    assert repr(dut.my_struct) == "LogicObject(sample_module.my_struct)"
+    if SIM_NAME.startswith("verilator"):
+        assert repr(dut.my_struct) == "LogicObject(sample_module.my_struct)"
```
ktbarrett, line 29: Hmmm... Can we not tell the object is a packed object, albeit with no subfields accessible?

src/cocotb/handle.py
```diff
@@ -1272,6 +1272,36 @@ def __str__(self) -> str:
...
+
+    def __init__(self, handle, path):
+        # Use HierarchyObject because _sub_handles contains SimHandleBase, which is a broader type
+        HierarchyObject.__init__(self, handle, path)
+
+    def __getattr__(self, name: str) -> Any:
+        if name in {"value", "set", "setimmediatevalue", "is_const"}:
+            return LogicObject.__getattribute__(self, name)
+        # private properties will be resolved in HierarchyObject
+        return HierarchyObject.__getattr__(self, name)
```
ktbarrett, line 1292: LogicArray is first in inheritance list, so this is implicit.

src/cocotb/handle.py
```diff
@@ -1272,6 +1272,36 @@ def __str__(self) -> str:
...
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        if name == "value":
+            return LogicObject.__setattr__(self, "value", value)
+
+        return HierarchyObject.__setattr__(self, name, value)
+
+    def __getitem__(self, index: str) -> SimHandleBase:
+        # we can't index into packed structs, so only support HierarchyObject's __getitem__
+        return HierarchyObject.__getitem__(self, index)
```
ktbarrett, line 1302: `LogicArray` doesn't support indexing, so this is implicit.

ktbarrett:
Oh wait... the last two comments only apply after #3777. Whoops...

I will probably try to finish out #3777 before merging this.

tests/test_cases/test_struct/test_struct.py
AndrewNolte: Not sure, I think we would have to indicate in https://github.com/verilator/verilator/blob/master/include/verilated_vpi.cpp, but even then I'm not sure if verilator includes info like that after verilation, so the marker might have to be put in earlier in the process

tests/test_cases/test_struct/test_struct.py
AndrewNolte: As in a separate test, as well as a separate PR to go in first?

AndrewNolte: Oh I see now, nvm I'll leave as one PR

src/cocotb/handle.py
```diff
@@ -1261,6 +1261,30 @@ def __str__(self) -> str:
...
```
ktbarrett, line 1285:
```suggestion
```

src/cocotb/handle.py
```diff
@@ -1261,6 +1261,30 @@ def __str__(self) -> str:
...
+
+    def __init__(self, handle, path):
+        # Use HierarchyObject because _sub_handles contains SimHandleBase, which is a broader type
+        HierarchyObject.__init__(self, handle, path)
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        if name == "value":
+            return LogicObject.__setattr__(self, "value", value)
+
+        return HierarchyObject.__setattr__(self, name, value)
```
ktbarrett, line 1281: I don't think this is necessary? Attempting to set `.value` should always trigger the setter property since `LogicObject` comes before `HierarchyObject` in the inheritance hierarchy.

ktbarrett:
For `test_iteration.recursive_discovery` Questa and Xcelium VPI now find 283 instead of 275. For Riviera failures: `test_array_simple.test_struct_unpacked` and `test_struct.test_struct_format` now passes.

PR #3910, opened 2024-05-29
Misc documentation fixes

teobiton:


src/cocotb/clock.py
```diff
@@ -119,15 +120,17 @@ def __init__(
         self.coro = None
         self.mcoro = None
 
-    async def start(self, cycles=None, start_high=True):
+    async def start(
+        self, cycles: Optional[int] = None, start_high: Optional[bool] = True
```
ktbarrett, line 124:
```suggestion
        self, cycles: Optional[int] = None, start_high: bool = True
```
Here optional just means that it's not required, i.e. you don't have to pass an argument, not that it's `typing.Optional` AKA `X | None`.

teobiton: Ah, I misunderstood the original comment. Fixed it!

ktbarrett (approved):

teobiton:
Oops, I referenced the wrong issue number in the commit, I'll force push again.

ktbarrett:
I don't think this is a very contentious change, so I won't wait up.

PR #3919, opened 2024-06-04
Edge count callbacks.

gilbertoabram:
Callbacks on a gpi signal change for a specific edge transition. This allows for an efficient implementation of ClockCycles that avoids going back and forth between python and simulator at each edge.

Useful for better performance when doing things such as
await ClockCycles(clk, 100)

It also provides a solution to https://github.com/cocotb/cocotb/issues/2208

### DIFF
--- a/src/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/src/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -589,7 +589,7 @@ static PyObject *register_edge_count_callback(
         return NULL;
     }
     if ((edge != GPI_RISING) && (edge != GPI_FALLING) &&
-        (edge != (GPI_RISING|GPI_FALLING))) {
+        (edge != (GPI_RISING | GPI_FALLING))) {
         PyErr_SetString(PyExc_TypeError,
                         "Attempt to register value change callback for an "
                         "invalid edge!");
@@ -615,9 +615,9 @@ static PyObject *register_edge_count_callback(
         return NULL;
     }
 
-    GpiCbHdl *hdl = gpi_register_edge_count_callback(
-        (gpi_function_t)handle_gpi_callback,
-        cb_data, sig_hdl, int(edge), count);
+    GpiCbHdl *hdl =
+        gpi_register_edge_count_callback((gpi_function_t)handle_gpi_callback,
+                                         cb_data, sig_hdl, int(edge), count);
 
     if (!hdl) {
         callback_data_delete(cb_data);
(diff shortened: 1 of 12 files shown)


ktbarrett:
I'm assuming it's possible to move `RisingEdge`, `FallingEdge`, and `Edge` to this system as well? They would just be `ClockCycles(handle, 1)`, right? (Future PR ofc).

ktbarrett:
Don't worry about Riviera for now. The CI is busted.

ktbarrett:
You might need to play with `GPI_EXPORT` and [defs](https://github.com/cocotb/cocotb/tree/master/src/cocotb/share/def) to make Windows happy.

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -538,6 +546,102 @@ static PyObject *register_value_change_callback(
...
+    PyObject *fArgs = PyTuple_GetSlice(args, 4, numargs);  // New reference
+    if (fArgs == NULL) {
+        return NULL;
+    }
+    Py_INCREF(function);
+
+    callback_data *cb_data = callback_data_new(function, fArgs, NULL);
+    if (cb_data == NULL) {
+        Py_DECREF(function);
+        Py_DECREF(fArgs);
```
ktbarrett, line 614: Rather than handle all the appropriate deallocations, just `DEFER(Py_DECREF(function))` after successfully creating the object.

marlonjames:
I'm not sold on putting this into the GPI codebase.
I think it would be better to start with solving #2208, but keep the edge counting part as an external cocotb extension for now.

src/cocotb/share/lib/simulator/simulatormodule.cpp
ktbarrett: This won't work, these objects need to be moved into the callback later. Disregard.

gilbertoabram:
Apologies, I pushed something to my branch before replying. Probably not ready to be accepted due to the comments, but I was curious if at least it fixed the CI issues.

> I'm assuming it's possible to move RisingEdge, FallingEdge, and Edge to this system as well? They would just be ClockCycles(handle, 1), right? (Future PR ofc).

Yes, ideally there would need to be some changes to have a special kind of object that does not automatically clean itself up (similarly to how it works now), but it would be possible.

> I'm not sold on putting this into the GPI codebase.
> I think it would be better to start with solving https://github.com/cocotb/cocotb/issues/2208, but keep the edge counting part as an external cocotb extension for now.

I'm not too familiar, I assumed extensions were python only, I can look into it.

While I was experimenting with it, it seemed that adding a counter did not add too much complexity. I suppose we don't have to add a new callback type and have a centralized coordinator, but doing it this way has the nice side effect that no matter how many edge callbacks there are, there is a single underlying callback driving it, as opposed to even now where if you await on all three of rising/falling/both, there will be three cbValueChange callbacks registered.

The main reason for something like this is that in our codebase we do stuff like
`await ClockCyles(clk, 100)`
which unfortunately switches to python back and forth at each edge, but it's something people prefer to use in general over using a Timer, in case the clock has jitter.

gilbertoabram:
Will look just at #2208

PR #3925, opened 2024-06-08
Prefix environment variables with `COCOTB_`

### LINKED ISSUE #3644: Prefix environment variables that are passed through to simulator with `COCOTB_`
I don't so much care about prefixing variables that are used only by the Makefiles or Python runner, but anything needed to cross into cocotb's Pythonland should be prefixed to prevent getting stomped by simulator environment variables and provide an easy way to support passthrough otherwise.

teobiton:
This change concerns environment variables MODULE, TESTCASE,
TOPLEVEL, RANDOM_SEED, COVERAGE, COVERAGE_RCFILE and PLUSARGS.
A warning is issued when these old variables are used.
The Makefile flow and runners are both affected.

Fixes #3644


src/cocotb/__init__.py
```diff
@@ -323,7 +323,16 @@ def _process_packages() -> None:
...
 def _start_user_coverage() -> None:
-    if "COVERAGE" in os.environ:
+    coverage_envvar = os.getenv("COCOTB_USER_COVERAGE")
+    if coverage_envvar is None:
+        coverage_envvar = os.getenv("COVERAGE")
+        if coverage_envvar is not None:
+            warnings.warn(
+                "COVERAGE is deprecated in favor of COCOTB_USER_COVERAGE",
+                DeprecationWarning,
+                stacklevel=2,
```
ktbarrett, line 333:
```suggestion
```
The stacklevel doesn't really apply here.

ktbarrett (commented):
There is also `PLUSARGS` and `COVERAGE_RCFILE` that are picked up by environment variable.

And in regression.py there's `RESULT_TESTSUITE` and `RESULT_PACKAGE` that should be updated, but you don't need deprecation warnings on them because they aren't documented.

Finally, I think you might need to update noxfile.py.

ktbarrett:
The failures have to do with causing warnings. We have an environment variable set in `test/Makefile` that turns warnings into errors.

src/cocotb_tools/makefiles/simulators/Makefile.icarus
```diff
@@ -70,10 +70,12 @@ endif
 
 # Compilation phase
 
+ICARUS_PLUSARGS = $(call deprecate,PLUSARGS,COCOTB_PLUSARGS)
```
ktbarrett, line 73: This warns early. Instead do the same thing as the other variables and don't use PLUSARGS to add `-fst`.

docs/source/newsfragments/3644.removal.rst
```diff
@@ -0,0 +1 @@
+Deprecated :envvar:`MODULE`, :envvar:`TOPLEVEL`, :envvar:`TESTCASE`, :envvar:`COVERAGE` and :envvar:`RANDOM_SEED`, that are respectively replaced with :envvar:`COCOTB_TEST_MODULES`, :envvar:`COCOTB_TOPLEVEL`, :envvar:`COCOTB_TESTCASE`, :envvar:`COCOTB_USER_COVERAGE` and :envvar:`COCOTB_RANDOM_SEED` to avoid issues with simulators overwriting cocotb environment variables.
```
ktbarrett, line 1: Mention PLUSARGS here.

tests/test_cases/test_seed/Makefile
```diff
@@ -5,6 +5,6 @@
 run:
 	$(RM) number
 	$(MAKE) sim MODULE=test_other,test_seed RANDOM_SEED=1234
```
ktbarrett, line 7:
```suggestion
	$(MAKE) sim MODULE=test_other,test_seed COCOTB_RANDOM_SEED=1234
```

ktbarrett (approved):

cmarqu (commented):
Thanks for the PR!
Maybe it's the GitHub app on my cell phone that is hiding things, but... is the change also reflected in the documentation where all the variables are described?

ktbarrett:
@teobiton Usually PRs are intended to self-encapsulated, but if that's going to pop up in the next couple days I'm okay with it. I'll make an issue to track it just incase things fall through.

PR #3927, opened 2024-06-09
Rename `SIM_OUTPUT_REDIRECT` to `SIM_CMD_SUFFIX` and add to all Makefiles

### LINKED ISSUE #1746: Makefiles write to sim.log inconsistently
Our ``Makefile.ghdl`` doesn't have the usual `` | tee $(SIM_BUILD)/sim.log`` at the end of the simulation command line nor any other option to that effect, so it doesn't produce a ``sim.log`` file at all.

ktbarrett:
Closes #1746. Adds `SIM_OUTPUT_REDIRECT` to all Makefiles, and then defaults it back to empty to prevent the redirection. 

From that issue:

> Since the | tee is breaking the isatty detection for color output, we should rather try to remove this.

This does it in a non-destructive way by making it opt-in.


cmarqu:
There is no description of the variable in the docs yet (apart from the release notes), right?
Maybe mention this is applied using bash syntax regardless of the user's shell.

ktbarrett:
There was, but it's now been updated.

docs/source/newsfragments/3561.feature.rst
```diff
@@ -1 +1 @@
-Added :make:var:`SIM_OUTPUT_REDIRECT`, an environment variable to control how simulator output is redirected.
+Added :make:var:`SIM_CMD_POSTFIX` to allow users to redirect simulator output or otherwise suffix the simulation command invocation.
```
cmarqu, line 1:
```suggestion
Added :make:var:`SIM_CMD_SUFFIX` to allow users to redirect simulator output or otherwise suffix the simulation command invocation.
```

src/cocotb_tools/makefiles/simulators/Makefile.icarus
```diff
@@ -94,16 +94,15 @@ $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/sim.vvp $(CUSTOM_SIM_DEPS)
...
 
 	$(call check_for_results_file)
 
 debug: $(SIM_BUILD)/sim.vvp $(CUSTOM_SIM_DEPS)
 	$(RM) -r $(COCOTB_RESULTS_FILE)
 
 	MODULE=$(MODULE) TESTCASE=$(TESTCASE) COCOTB_TEST_FILTER=$(COCOTB_TEST_FILTER) TOPLEVEL=$(TOPLEVEL) TOPLEVEL_LANG=$(TOPLEVEL_LANG) \
-        $(SIM_CMD_PREFIX) gdb --args $(ICARUS_BIN_DIR)/vvp $(shell cocotb-config --lib-dir) -m $(shell cocotb-config --lib-name vpi icarus) $(SIM_BUILD)/sim.vvp $(SIM_ARGS) $(EXTRA_ARGS) $(PLUSARGS) $(SIM_OUTPUT_REDIRECT)
-
+        $(SIM_CMD_PREFIX) gdb --args $(ICARUS_BIN_DIR)/vvp $(shell cocotb-config --lib-dir) -m $(shell cocotb-config --lib-name vpi icarus) $(SIM_BUILD)/sim.vvp $(SIM_ARGS) $(EXTRA_ARGS) $(PLUSARGS)
```
cmarqu, line 105: No suffix here? Hard to tell if that's right while on a cell phone.

cmarqu (approved):
Just two smaller things.

src/cocotb_tools/makefiles/simulators/Makefile.icarus
ktbarrett: If it has a PREFIX, it should have a suffix.

ktbarrett, line 105:
```suggestion
        $(SIM_CMD_PREFIX) gdb --args $(ICARUS_BIN_DIR)/vvp $(shell cocotb-config --lib-dir) -m $(shell cocotb-config --lib-name vpi icarus) $(SIM_BUILD)/sim.vvp $(SIM_ARGS) $(EXTRA_ARGS) $(PLUSARGS) $(SIM_CMD_SUFFIX)
```

PR #3938, opened 2024-06-12
Remove `cocotb.result.SimFailure`

### LINKED ISSUE #3788: Make `SimFailure` private
No reasonable user is going to write tests that are expected to cause the simulator to fail. I feel this is a regression-only feature. I think this will require some refactoring to list what happens when a simulator *does* fail in the results.xml.

The fact this is public provides to users a way to fail the result of the regression by throwing it themselves, or to catch it, which may not work as intended. I don't think we should be providing these implicit behaviors.

ktbarrett:
Closes #3788. Depends on #3941.

This PR ended up being more of a refactor of the RegressionManager than intended. 

This PR:
* Moves `SimFailure` to regression.py (to be removed once I refactor how tests end)
* Adds `_sim_expect_failure` flag to `cocotb.test` decorator to cover this use case.

### DIFF
--- a/docs/source/newsfragments/3938.removal.rst
+++ b/docs/source/newsfragments/3938.removal.rst
@@ -1,0 +1,1 @@
+``cocotb.result.SimFailure`` has been removed. It's replaced with the *_expect_sim_failure* argument to :func:`cocotb.test`. Users are not intended to use this functionality.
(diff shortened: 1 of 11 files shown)


ktbarrett:
Yikes. We are bumping into the awful cleanup behavior in the scheduler again. The reason that `_finish_scheduler` only called `_test_complete_cb` previously (which skips `_cleanup`), is because `test_complete` is run *after a Timer(1) callback*, which cannot happen after simulator failure.

If shutdown were *sane*, we would instead cleanup the scheduler and record test results at the time the test finishes and then *start* succeeding tests after a `Timer(1)`. I think that change will need to come in first.

ktbarrett:
This PR needs to be broken up for sure.

cmarqu (approved):

PR #3931, opened 2024-06-09
Make awaiting `Join` trigger not return `Task` result

ktbarrett:
Awaiting the Join trigger will not return the task's
result, the user can use `await task` for the same effect. This allows
the user to have some way of awaiting a task to finish *without* getting
the result, which is especially useful if the outcome is anticipated to
be an exception.

PR #3941, opened 2024-06-13
Move cleanup and test score to before wait on `Timer(1)`

ktbarrett:
Moves the cleanup and test score to before the wait on `Timer(1)` to start the next test.

~~This is a WIP. The issue is that we can't start the test with `Timer(1)` because when it attempts to start the unstarted task, it uses `Trigger._outcome` which is *usually* the trigger; but coroutines can only be started with `None`. I'm going to change `Trigger` to remove the `_outcome` and only ever send it `None`, as mentioned [here](https://github.com/cocotb/cocotb/pull/3931#issuecomment-2159465269), and have the `__await__` function return any value that's pertinent.~~ Fixed.

### DIFF
--- a/src/cocotb/regression.py
+++ b/src/cocotb/regression.py
@@ -430,7 +430,7 @@ class RegressionManager:
             else:
                 self._test_start_sim_time = get_sim_time("ns")
                 self._test_start_time = time.time()
-                return cocotb._scheduler._add_test(self._test_task)
+                return cocotb._scheduler_inst._add_test(self._test_task)
 
         return self._tear_down()
 
--- a/src/cocotb/regression.py
+++ b/src/cocotb/regression.py
@@ -58,11 +58,13 @@ from typing import (
 )
 
 import cocotb
+import cocotb._scheduler
 from cocotb import _ANSI, simulator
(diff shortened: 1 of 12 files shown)


src/cocotb/regression.py
```diff
@@ -430,10 +435,21 @@ def _execute(self, *, sim_failed: bool = False) -> None:
...
+                    self._first_test = False
+                    return self._schedule_next_test()
+                else:
+                    return self._timer1._prime(self._schedule_next_test)
 
         return self._tear_down()
 
+    def _schedule_next_test(self, trigger: Optional[Trigger] = None) -> None:
+        if trigger is not None:
+            cocotb._scheduler_inst._sim_phase = cocotb._scheduler.SimPhase.NORMAL
```
ktbarrett, line 449: I feel like I want to see `SimPhase` moved to a global and tracked by the `GPITrigger`s so we don't have to do this here.

ktbarrett:
The one failing test is due to a single test that consumes no simulation time which leads to a recursive call to `RegressionManager._execute`.

imphil (approved):

PR #3942, opened 2024-06-13
Scheduler refactor

ktbarrett:
Removed `pending_triggers` and `scheduling` queues in the scheduler. Made `pending_tasks` *the* task queue. Whenever triggers fires they now immediately add the tasks waiting upon them to the `pending_tasks` queue alongside the outcome the scheduler should inject into the task.

Also cleaned up state tracking a bit by using an Enum.

Separated react method into one for GPITriggers and the other for Python Triggers.

- [x] potential speed optimization by using a data structure other than list for the pending_tasks queue. We currently need an O(n) lookup loop in Python.

### DIFF
--- a/src/cocotb/_scheduler.py
+++ b/src/cocotb/_scheduler.py
@@ -41,7 +41,8 @@ import threading
 import warnings
 from collections import OrderedDict
 from collections.abc import Coroutine
-from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Union
+from enum import Enum, auto
+from typing import Any, Callable, Dict, List, Sequence, Tuple, Union
 
 import cocotb
 from cocotb import _outcomes, _py_compat
@@ -172,6 +173,12 @@ class external_waiter:
         return self.state
 
 
+class GPIState(Enum):
+    NORMAL = auto()
+    READ_WRITE = auto()
+    READ_ONLY = auto()
+
+
 class Scheduler:
     """The main scheduler.
 
@@ -217,16 +224,12 @@ class Scheduler:
     "normal mode" i.e. where writes are permitted.
     """
 
-    _MODE_NORMAL = 1  # noqa
-    _MODE_READONLY = 2  # noqa
-    _MODE_WRITE = 3  # noqa
-    _MODE_TERM = 4  # noqa
-
     # Singleton events, recycled to avoid spurious object creation
     _next_time_step = NextTimeStep()
     _read_write = ReadWrite()
     _read_only = ReadOnly()
     _timer1 = Timer(1)
+    _none_outcome = _outcomes.Value(None)
 
     def __init__(self, test_complete_cb: Callable[[], None]) -> None:
         self._test_complete_cb = test_complete_cb
@@ -239,10 +242,12 @@ class Scheduler:
 
         # A dictionary of pending tasks for each trigger,
         # indexed by trigger
-        self._trigger2tasks = _py_compat.insertion_ordered_dict()
+        self._trigger2tasks: Dict[Trigger, list[Task]] = (
+            _py_compat.insertion_ordered_dict()
+        )
 
         # Our main state
-        self._mode = Scheduler._MODE_NORMAL
+        self._state = GPIState.NORMAL
 
         # A dictionary of pending (write_func, args), keyed by handle.
         # Writes are applied oldest to newest (least recently used).
@@ -251,19 +256,15 @@ class Scheduler:
             SimHandleBase, Tuple[Callable[..., None], Sequence[Any]]
         ] = OrderedDict()
 
-        self._pending_tasks = []
-        self._pending_triggers = []
+        self._pending_tasks: List[Tuple[Task[Any]], _outcomes.Outcome] = []
         self._pending_threads = []
         self._pending_events = []  # Events we need to call set on once we've unwound
-        self._scheduling = []
 
         self._terminate = False
         self._test = None
         self._main_thread = threading.current_thread()
 
         self._current_task = None
-
-        self._is_reacting = False
 
         self._write_task = None
         self._writes_pending = Event()
@@ -272,8 +273,6 @@ class Scheduler:
         """An internal task that performs pending writes"""
         while True:
             await self._writes_pending.wait()
-            if self._mode != Scheduler._MODE_NORMAL:
-                await self._next_time_step
 
             await self._read_write
 
@@ -305,7 +304,6 @@ class Scheduler:
             self._terminate = False
             self._write_calls.clear()
             self._writes_pending.clear()
-            self._mode = Scheduler._MODE_TERM
 
     def _test_completed(self, trigger=None):
         """Called after a test and its cleanup have completed"""
@@ -319,13 +317,13 @@ class Scheduler:
             ctx = _py_compat.nullcontext()
 
         with ctx:
-            self._mode = Scheduler._MODE_NORMAL
             if trigger is not None:
                 trigger._unprime()
 
             # extract the current test, and clear it
             test = self._test
             self._test = None
+            self._state = GPIState.NORMAL
             if test is None:
                 raise InternalError("_test_completed called with no active test")
             if test._outcome is None:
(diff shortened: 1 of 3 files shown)


src/cocotb/_scheduler.py
```diff
@@ -251,29 +256,23 @@ def __init__(self, test_complete_cb: Callable[[], None]) -> None:
...
-
         self._write_task = None
         self._writes_pending = Event()
 
     async def _do_writes(self):
         """An internal task that performs pending writes"""
         while True:
             await self._writes_pending.wait()
-            if self._mode != Scheduler._MODE_NORMAL:
-                await self._next_time_step
```
ktbarrett, line 276: This is not possible since `_schedule_write` can't be called except in NORMAL, and set is only called there.

src/cocotb/_scheduler.py
```diff
@@ -319,13 +317,13 @@ def _test_completed(self, trigger=None):
...
 
         with ctx:
-            self._mode = Scheduler._MODE_NORMAL
             if trigger is not None:
                 trigger._unprime()
 
             # extract the current test, and clear it
             test = self._test
             self._test = None
+            self._state = GPIState.NORMAL
```
ktbarrett, line 326: I'll move this up to the `self._mode` line from above.

src/cocotb/_scheduler.py
```diff
@@ -341,170 +339,105 @@ def _test_completed(self, trigger=None):
...
-                    if task._outcome is not None:
-                        # Task was killed by another task waiting on the same trigger
-                        continue
-                    if _debug:
-                        self.log.debug(f"Scheduling task {task}")
-                    self._schedule(task, trigger=trigger)
-                    if _debug:
-                        self.log.debug(f"Scheduled task {task}")
+        # This trigger isn't needed any more
+        trigger._unprime()
```
ktbarrett, line 398: Move this up? I think we want to unprime incase the tasks reprime them.

src/cocotb/_scheduler.py
```diff
@@ -341,170 +339,105 @@ def _test_completed(self, trigger=None):
...
-                    self._pending_events.pop(0).set()
+            # remove our reference to the objects at the end of each loop,
+            # to try and avoid them being destroyed at a weird time (as
+            # happened in gh-957)
+            del task
 
-                # remove our reference to the objects at the end of each loop,
-                # to try and avoid them being destroyed at a weird time (as
-                # happened in gh-957)
-                del trigger
```
ktbarrett, line 493: No longer necessary. We could move this up to the last use of the trigger in `_react`, but it would be at the same place as `trigger._unprime` there, which is equivalent.

src/cocotb/_scheduler.py
```diff
@@ -902,27 +833,12 @@ def _cleanup(self):
...
-            if _debug:
-                self.log.debug(f"Unpriming {trigger}")
-            trigger._unprime()
-        assert not self._pending_triggers
-
         # Kill any queued coroutines.
         # We use a while loop because task.kill() calls _unschedule(), which will remove the task from _pending_tasks.
         # If that happens a for loop will stop early and then the assert will fail.
         while self._pending_tasks:
             # Get first task but leave it in the list so that _unschedule() will correctly close the unstarted coroutine object.
```
ktbarrett, line 840: Remove this comment.

src/cocotb/_scheduler.py
```diff
@@ -589,24 +522,31 @@ def _resume_task_upon(self, task, trigger):
                 raise InternalError("More than one task waiting on an unprimed trigger")
 
             try:
-                trigger._prime(self._react)
+                if isinstance(trigger, GPITrigger):
+                    trigger._prime(self._gpi_react)
+                else:
+                    trigger._prime(self._react)
```
ktbarrett, line 528: I'm thinking I'll just go back to the one `_react` call in which I'll put in a `if isinstance(trigger, GPITrigger)` guard for the GPITrigger specific stuff.

ktbarrett: Ehhh, maybe something different. Since we are using context managers for profiling `_gpi_react` wrapping `_react` is useful.

src/cocotb/_scheduler.py
ktbarrett: Unnecessary, we are only scheduling, no tasks are running.

src/cocotb/_scheduler.py
```diff
@@ -172,6 +173,12 @@ def thread_wait(self):
         return self.state
 
 
+class GPIState(Enum):
```
cmarqu, line 176: Will this be the only state of the GPI "forever", or should this be a more specific name in case of other states?

src/cocotb/_scheduler.py
```diff
@@ -341,170 +339,105 @@ def _test_completed(self, trigger=None):
...
-                        self.log.debug(f"Scheduled task {task}")
+        # This trigger isn't needed any more
+        trigger._unprime()
 
-                    # remove our reference to the objects at the end of each loop,
-                    # to try and avoid them being destroyed at a weird time (as
-                    # happened in gh-957)
-                    del task
+    def _event_loop(self) -> None:
+        """Run the main event loop
```
cmarqu, line 401:
```suggestion
        """Run the main event loop.
```

marlonjames:
`GPITrigger` objects are still being primed with `_react`, not `_gpi_react`

src/cocotb/_scheduler.py
```diff
@@ -172,6 +173,12 @@ def thread_wait(self):
         return self.state
 
 
+class GPIState(Enum):
+    NORMAL = auto()
+    READ_WRITE = auto()
+    READ_ONLY = auto()
+
+
 class Scheduler:
```
cmarqu, line 182: Maybe change the uses of "mode" to "state" in this docstring?

cmarqu (commented):
Some superficial comments FWIW.

src/cocotb/_scheduler.py
ktbarrett: This is specifically for tracking GPI states so we can do things like preventing registering ReadWrite in ReadOnly stage. If there end up being different kinds of states, they will be in a different variable. I can take an action to change the variable name to `_gpi_state`.

src/cocotb/_scheduler.py
ktbarrett: I rewrote the docstring since it could use a bit more detail.

src/cocotb/_scheduler.py
```diff
@@ -341,170 +339,110 @@ def _test_completed(self, trigger=None):
...
+            # associated task waiting on that trigger, otherwise it would
+            # have been unprimed already
+            if isinstance(trigger, GPITrigger):
+                self.log.critical(f"No tasks waiting on trigger that fired: {trigger}")
+                trigger.log.info("I'm the culprit")
+            # For Python triggers this isn't actually an error - we might do
+            # event.set() without knowing whether any tasks are actually
+            # waiting on this event, for example
+            elif _debug:
+                self.log.debug(f"No tasks waiting on trigger that fired: {trigger}")
```
imphil, line 380: Doesn't need to be in this PR, but the codecov report indicates that we could add a test for this condition (since you're already describin the scenario in the comment above).

imphil (approved):
Just some quick drive-by comments. You probably know the scheduler better than I ever will, so trusting your judgement here :smile: 

Thanks!

src/cocotb/_scheduler.py
ktbarrett: I have thought about this. I'd like to add a test to the repo which runs `test_cocotb` with the scheduler debug on to hit these debug lines, then we can more accurately judge which cases we're actually missing.

ktbarrett:
@imphil Once I'm done refactoring and documenting the scheduler, it should be much easier to understand it. I didn't understand it for a while, but that's mostly because it was doing everything it needed to do in the most unnecessarily complex way it could.

ktbarrett:
@cmarqu I'm going to need some attention to the new docstrings.

src/cocotb/_scheduler.py
```diff
@@ -172,61 +173,94 @@ def thread_wait(self):
...
+        continuing its execution from where it previously yielded control -
+        by injecting the Outcome value associated with the Task from the queue.
+        The Task's body will run until it finishes or reaches the next ``await`` statement.
+        If a Task reaches an ``await``, :meth:`_schedule` will convert the value yielded from the Task into a Trigger with :meth:`_trigger_from_any` and its friend methods.
+        Triggers are then `primed` (with :meth:`~cocotb.triggers.Trigger._prime`)
+        with a `react` function (:meth:`_sim_react` or :meth:`_react)
+        so as to wake up Tasks waiting for that Trigger to `fire` (when the event encoded by the Trigger occurs).
+        This is accomplished by :meth:`_resume_task_upon`.
+        :meth:`_resume_task_upon` also associates the Trigger with the Task waiting on it to fire by adding them to the :attr:`_triggers2tasks` map.
+        If instead, a Task finishes, :meth:`_schedule` will cause the :class:`~cocotb.triggers.Join` trigger to fire.
```
cmarqu, line 199:
```suggestion
        If, instead of reaching an ``await``, a Task finishes, :meth:`_schedule` will cause the :class:`~cocotb.triggers.Join` trigger to fire.
```

src/cocotb/_scheduler.py
```diff
@@ -172,61 +173,94 @@ def thread_wait(self):
...
+        Once a Trigger fires it calls the react function which queues all Tasks waiting for that Trigger to fire.
+        Then the process repeats.
+
+        When a Task is cancelled (:meth:`_unschedule`), it is removed from the Task queue if it is currently queued.
+        Also, the Task and Trigger are deassociated in the :attr:`_triggers2tasks` map.
+        If the cancelled Task is the last Task waiting on a Trigger, that Trigger is `unprimed` to prevent it from firing.
+
+    Simulator Phase:
```
cmarqu, line 207:
```suggestion
    Simulator Phases:
```

src/cocotb/_scheduler.py
```diff
@@ -172,61 +173,94 @@ def thread_wait(self):
...
+            and there will not be another ReadWrite phase in this time step.
+            Simulators generally handle this caveat gracefully by leaving you in the ReadWrite phase of the next time step.
+
+        ReadOnly phase
+            Corresponds to ``cbReadOnlySynch`` (VPI) or ``vhpiCbRepEndOfTimeStep`` (VHPI).
+            In this state we are not allowed to perform writes.
+            Scheduling :class:`~cocotb.triggers.ReadWrite` and :class:`~cocotb.triggers.ReadOnly` are *not* valid.
+
+    Caveats and Special Cases:
+        Tests are treated specially in the scheduler.
```
cmarqu, line 232:
```suggestion
        The scheduler treats Tests specially.
```
(to have the uppercase "Tests")

cmarqu (approved):

src/cocotb/_scheduler.py
```diff
@@ -341,170 +368,121 @@ def _test_completed(self, trigger=None):
...
-                    if _debug:
-                        self.log.debug(f"Scheduled task {task}")
+    def _event_loop(self) -> None:
+        """Run the main event loop.
 
-                    # remove our reference to the objects at the end of each loop,
-                    # to try and avoid them being destroyed at a weird time (as
-                    # happened in gh-957)
-                    del task
+        This should be triggered by only:
```
marlonjames, line 442: This phrasing sounds more natural to me.
```suggestion
        This should only be started by:
```

src/cocotb/_scheduler.py
```diff
@@ -341,170 +368,121 @@ def _test_completed(self, trigger=None):
...
+                    self.log.debug(
+                        f"Scheduling pending event {self._pending_events[0]}"
+                    )
+                self._pending_events.pop(0).set()
 
-                # remove our reference to the objects at the end of each loop,
-                # to try and avoid them being destroyed at a weird time (as
-                # happened in gh-957)
-                del trigger
+        # no more pending triggers
```
marlonjames, line 469:
```suggestion
        # no more pending tasks
```

src/cocotb/_scheduler.py
```diff
@@ -341,170 +368,121 @@ def _test_completed(self, trigger=None):
...
+            self.log.debug("All tasks scheduled, handing control back to simulator")
 
-            # no more pending triggers
-            self._check_termination()
-            if _debug:
-                self.log.debug("All tasks scheduled, handing control back to simulator")
+    def _unschedule(self, task: Task[Any]) -> None:
+        """Unschedule a task and unprime dangling pending triggers.
+
+        Also...
```
marlonjames, line 477:
```suggestion
        Also:
```

src/cocotb/_scheduler.py
```diff
@@ -589,24 +567,38 @@ def _resume_task_upon(self, task, trigger):
...
+    def _queue(
+        self, task: Task[Any], outcome: _outcomes.Outcome[Any] = _none_outcome
+    ) -> None:
+        """Queue *task* for scheduling.
 
-    def _queue(self, task):
-        """Queue a task for execution"""
+        If the task is already scheduled and you are attempting to reschedule it with a different outcome,
+        we assume that is a mistake and raise an Exception.
+        That behavior may change in the future.
```
marlonjames, line 590:
```suggestion
        This behavior may change in the future.
```

src/cocotb/_scheduler.py
```diff
@@ -785,27 +775,25 @@ def _trigger_from_any(self, result) -> Trigger:
...
+        """Schedule *task* with *outcome*.
 
         Args:
             task: The task to schedule.
-            trigger: The trigger that caused this
-                task to be scheduled.
+            outcome: The outcome to inject into the *task*.
+
+        Scheduling runs *task* until it either finishes or reaches the next ``await`` statement.
+        If the *task* completes, it is unscheduled, a Join trigger fires, and test completion is inspected.
```
marlonjames, line 786:
```suggestion
        If *task* completes, it is unscheduled, a Join trigger fires, and test completion is inspected.
```

marlonjames (commented):
I think it would be useful to have `Scheduler` in the docs somewhere, and that would check the docstrings are up-to-date and correct.
I know it was removed in #3806, but maybe it can be put in the developer section?

src/cocotb/_scheduler.py
```diff
@@ -172,61 +173,94 @@ def thread_wait(self):
...
+            Scheduling :class:`~cocotb.triggers.ReadWrite` and :class:`~cocotb.triggers.ReadOnly` are valid.
+            One caveat is that scheduling a :class:`~cocotb.triggers.ReadWrite` while in this phase may not be valid.
+            If there were no writes applied at the beginning of this phase, there will be no more events in this time step,
+            and there will not be another ReadWrite phase in this time step.
+            Simulators generally handle this caveat gracefully by leaving you in the ReadWrite phase of the next time step.
+
+        ReadOnly phase
+            Corresponds to ``cbReadOnlySynch`` (VPI) or ``vhpiCbRepEndOfTimeStep`` (VHPI).
+            In this state we are not allowed to perform writes.
+            Scheduling :class:`~cocotb.triggers.ReadWrite` and :class:`~cocotb.triggers.ReadOnly` are *not* valid.
```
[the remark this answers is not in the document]
ktbarrett: I'm not familiar with this restriction, where's this coming from?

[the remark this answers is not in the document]
marlonjames: > VHDL-2019 LRM 21.3.6.8 vhpiCbRepEndOfTimeStep
>
> It is an error if a vhpiCbEndOfTimeStep or vhpiCbRepEndOfTimeStep callback causes activity
on a driver or a signal (see 14.7.3.1) or registers or enables a vhpiCbAfterDelay,
vhpiCbRepAfterDelay, vhpiCbTimeout, or vhpiCbRepTimeOut callback.

> SystemVerilog 2017 LRM 38.36.2
>
> cbReadOnlySynch
Callback shall occur the same as for cbReadWriteSynch, except that writing
values or scheduling events before the next scheduled event is not allowed.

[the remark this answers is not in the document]
ktbarrett: > or scheduling events before the next scheduled event is not allowed.

I think this has a couple interpretations.

> registers or enables a vhpiCbAfterDelay, vhpiCbRepAfterDelay, vhpiCbTimeout, or vhpiCbRepTimeOut callback.

That is very clear cut, but I don't think any simulator has this limitation, nor do I see it as particularly reasonable. Is the issue that we are maybe using the wrong callback? Should we be using a read-only cycle before the true end of time period so we can create new time-based callbacks?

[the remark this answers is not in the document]
ktbarrett: I created an issue to investigate this. I will punt this comment until the outcome of that investigation.

src/cocotb/_scheduler.py
ktbarrett: Filed #3968.

Issue #3940, opened 2024-06-12
Fix documentation for cocotb environment variables


ktbarrett:
From https://github.com/cocotb/cocotb/pull/3925#issuecomment-2163176330.

The documentation for cocotb environment variables needs updating before the next release.

Issue #3694, opened 2024-02-01
Document which simulators support Force/Release


marlonjames:
The docs have information about how to use `Force`, `Freeze` and `Release`, but do not detail which simulators support them.

Issue #2981, opened 2022-05-19
Quickstart docs: Link to stable version of code examples


imphil:
At https://docs.cocotb.org/en/stable/quickstart.html we have this text:

> The code for the following example is available as [examples/doc_examples/quickstart](https://github.com/cocotb/cocotb/blob/master/examples/doc_examples/quickstart) in the cocotb sources.

The examples link points to Git master, not to the stable version that the text should be used with. This might have lead to confusion at #2978

imphil:
Ideally we link to the Git hash the documentation was built with.

PR #3943, opened 2024-06-14
Documentation updates

teobiton:
This PR addresses small issues of the documentation:

- #3940
- #3694 
- #2981

### DIFF
--- a/docs/source/building.rst
+++ b/docs/source/building.rst
@@ -34,7 +34,7 @@ Variables
 
 The following sections document environment variables and makefile variables according to their owner/consumer.
 
-Of the environment variables, only :envvar:`MODULE` is mandatory to be set
+Of the environment variables, only :envvar:`COCOTB_TEST_MODULES` is mandatory to be set
 (typically done in a makefile or run script), all others are optional.
 
 ..
@@ -43,7 +43,7 @@ Of the environment variables, only :envvar:`MODULE` is mandatory to be set
 Cocotb
 ------
 
-.. envvar:: TOPLEVEL
+.. envvar:: COCOTB_TOPLEVEL
 
     Use this to indicate the instance in the hierarchy to use as the :term:`DUT`.
     If this isn't defined then the first root instance is used.
@@ -54,7 +54,15 @@ Cocotb
 
     .. versionchanged:: 1.6.0 Strip leading and trailing whitespace
 
-.. envvar:: RANDOM_SEED
+    .. versionchanged:: 2.0
+
+        :envvar:`TOPLEVEL` is renamed to :envvar:`COCOTB_TOPLEVEL`.
+
+    .. deprecated:: 2.0
+
+        :envvar:`TOPLEVEL` is a deprecated alias and will be removed.
+
+.. envvar:: COCOTB_RANDOM_SEED
 
     Seed the Python random module to recreate a previous test stimulus.
     At the beginning of every test a message is displayed with the seed used for that execution:
@@ -68,9 +76,17 @@ Cocotb
 
     .. code-block:: bash
 
-       make RANDOM_SEED=1377424946
+       make COCOTB_RANDOM_SEED=1377424946
 
-    See also: :make:var:`PLUSARGS`
+    See also: :make:var:`COCOTB_PLUSARGS`
+
+    .. versionchanged:: 2.0
+
+        :envvar:`RANDOM_SEED` is renamed to :envvar:`COCOTB_RANDOM_SEED`.
+
+    .. deprecated:: 2.0
+
+        :envvar:`RANDOM_SEED` is a deprecated alias and will be removed.
 
 .. envvar:: COCOTB_ANSI_OUTPUT
 
@@ -154,19 +170,35 @@ Cocotb
 Regression Manager
 ~~~~~~~~~~~~~~~~~~
 
-.. envvar:: MODULE
+.. envvar:: COCOTB_TEST_MODULES
 
     The name of the Python module(s) to search for test functions -
-    if your tests are in a file called ``test_mydesign.py``, ``MODULE`` would be set to ``test_mydesign``.
+    if your tests are in a file called ``test_mydesign.py``, ``COCOTB_TEST_MODULES`` would be set to ``test_mydesign``.
     Multiple modules can be specified using a comma-separated list.
     All tests will be run from each specified module in order of the module's appearance in this list.
 
     The is the only environment variable that is **required** for cocotb, all others are optional.
 
-.. envvar:: TESTCASE
+    .. versionchanged:: 2.0
+
+        :envvar:`MODULE` is renamed to :envvar:`COCOTB_TEST_MODULES`.
+
+    .. deprecated:: 2.0
+
+        :envvar:`MODULE` is a deprecated alias and will be removed.
+
+.. envvar:: COCOTB_TESTCASE
 
     A comma-separated list of tests to run.
     Does an exact match on the test name.
+
+    .. versionchanged:: 2.0
+
+        :envvar:`TESTCASE` is renamed to :envvar:`COCOTB_TESTCASE`.
+
+    .. deprecated:: 2.0
+
+        :envvar:`TESTCASE` is a deprecated alias and will be removed.
 
     .. deprecated:: 2.0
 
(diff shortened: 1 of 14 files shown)


docs/source/conf.py
```diff
@@ -84,6 +84,10 @@
 # The short X.Y version.
 v_major, v_minor = LooseVersion(release).version[:2]
 version = "{}.{}".format(v_major, v_minor)
+# Cocotb commit ID
+commit_id = (
```
cmarqu, line 88: I think it would be good to also (still) support building the documentation from a tarball without git installed.

docs/source/building.rst
```diff
@@ -54,11 +54,19 @@ Cocotb
 
     .. versionchanged:: 1.6.0 Strip leading and trailing whitespace
 
+    .. deprecated:: 2.0
+
+        :envvar:`TOPLEVEL` is renamed to :envvar:`COCOTB_TOPLEVEL`.
```
cmarqu, line 59: You aren't actually introducing the new names to the list of variables yet, are you? (Or is the GH client on my cell phone fooling me?)

teobiton: No you're right I just added these messages for each that was renamed, I don't know if that's enough though. My other idea was to introduce the new names in the list and make the old ones point to their description instead.

cmarqu: I also think that using the new names and making the old ones be pointers is the best solution.

docs/source/writing_testbenches.rst
```diff
@@ -230,6 +230,15 @@ the various actions described in :ref:`assignment-methods` can be used.
...
     dut.my_signal.value = Freeze()     # my_signal stays at current value until released
 
+.. warning::
+
+    Some actions do not work on all simulators:
+
+    - Release does not work on GHDL (see :issue:`3830`).
+    - Force and Release do not work on Verilator (see :issue:`3831`).
+    - Riviera's VPI implicitly releases signal when overwriting forced signal with normal deposit (see :issue:`3832`).
+    - Questa's FLI allows overwriting forced signal with normal deposit (see :issue:`3833`).
```
ktbarrett, line 240: I like the thoroughness, but we have to remember to delete these lines when upstream fixes the issue. We either need to make this more general like "not all simulators fully support this feature, see the issues page" and link a URL that does a search on the issues that returns these PRs. ***OR*** we can just add comments to those issues to update the documentation so we don't forget.

teobiton: I'd go with the second option, that way it's explicitly written in the documentation and then the lines can be deleted when upstream issues are resolved. I can add comments on the issues once this is merged.

teobiton:
Just updated the documentation and added the list of new variables. It feels weird to have `COCOTB_TESTCASES` as a new variable but it's also deprecated at the same time.
Plus, if I change to the new names everywhere in the documentation, I also have to change it in the quick-start examples and such, is that something we want now or later ?

ktbarrett:
@teobiton Yeah you'll need to update all the existing refs. That's expected. As for the `.. deprecated::` label, I might instead use `.. versionchanged:: 2.0`. What say you @cmarqu?

cmarqu:
> @teobiton Yeah you'll need to update all the existing refs. That's expected. As for the `.. deprecated::` label, I might instead use `.. versionchanged:: 2.0`. What say you @cmarqu?

The description of https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated sounds fitting here IMO.

docs/source/writing_testbenches.rst
cmarqu: It may very well be the case that some simulators fix their behavior in newer versions even for older cocotb versions, and then our  published docs would not be accurate anymore...
We have links to issues based on labels for each supported simulator in the Simulator Support section.
Maybe it's enough to mention those links here?

teobiton:
Just updated all references to the old variables that appear in the documentation and left everything else as is.

docs/source/custom_flows.rst
```diff
@@ -16,7 +16,7 @@ this chapter shows the minimum settings to be done.
 For all simulators, the following environment variables need to be set:
 
 * Define :envvar:`LIBPYTHON_LOC` using ``$(cocotb-config --libpython)``.
-* Define :envvar:`MODULE` with the name of the Python module containing your testcases.
+* Define :envvar:`COCOTB_TEST_MODULES` with the name of the Python module containing your testcases.
```
cmarqu, line 19:
```suggestion
* Define :envvar:`COCOTB_TEST_MODULES` with the name of the Python module(s) containing your testcases.
```

docs/source/quickstart.rst
```diff
@@ -100,8 +100,8 @@ In the ``Makefile`` shown below we specify:
 * the default simulator to use (:make:var:`SIM`),
 * the default language of the toplevel module or entity (:make:var:`TOPLEVEL_LANG`, ``verilog`` in our case),
 * the design source files (:make:var:`VERILOG_SOURCES` and :make:var:`VHDL_SOURCES`),
-* the toplevel module or entity to instantiate (:envvar:`TOPLEVEL`, ``my_design`` in our case),
-* and a Python module that contains our cocotb tests (:envvar:`MODULE`.
+* the toplevel module or entity to instantiate (:envvar:`COCOTB_TOPLEVEL`, ``my_design`` in our case),
+* and a Python module that contains our cocotb tests (:envvar:`COCOTB_TEST_MODULES`.
```
cmarqu, line 104:
```suggestion
* and Python modules that contain our cocotb tests (:envvar:`COCOTB_TEST_MODULES`).
```

docs/source/writing_testbenches.rst
```diff
@@ -230,6 +230,10 @@ the various actions described in :ref:`assignment-methods` can be used.
     # Freeze action
     dut.my_signal.value = Freeze()     # my_signal stays at current value until released
 
+.. warning::
+
+    Not all simulators support these features; refer to the :ref:`simulator-support` section for details or to `issues with label upstream <https://github.com/cocotb/cocotb/issues?q=is%3Aissue+-label%3Astatus%3Aduplicate+label%3Aupstream>`_
```
cmarqu, line 235:
```suggestion
    Not all simulators support these features; refer to the :ref:`simulator-support` section for details or to `issues with label "upstream" <https://github.com/cocotb/cocotb/issues?q=is%3Aissue+-label%3Astatus%3Aduplicate+label%3Aupstream>`_
```

docs/source/building.rst
```diff
@@ -225,8 +278,18 @@ Regression Manager
...
+
+.. envvar:: COVERAGE_RCFILE
+
+    See :envvar:`COCOTB_COVERAGE_RCFILE`.
+
     .. versionadded:: 1.7
 
+    .. deprecated:: 2.0
+
+        :envvar:`COVERAGE_RCFILE` is renamed to :envvar:`COCOTB_COVERAGE_RCFILE`.
```
cmarqu, line 291: While this PR is documenting the current code correctly, I think the PR that did the renaming of this variable in the code has slightly broken the coverage functionality.
Unless I'm misunderstanding things, we did rely on `coverage.py` to read the original name `COVERAGE_RCFILE` (see https://coverage.readthedocs.io/en/latest/config.html#configuration-reference), which it won't do for `COCOTB_COVERAGE_RCFILE`.

cmarqu: I opened #3947 for this.

docs/source/building.rst
```diff
@@ -154,22 +174,36 @@ Cocotb
...
 
     A comma-separated list of tests to run.
     Does an exact match on the test name.
 
+    .. versionadded:: 2.0
+
     .. deprecated:: 2.0
 
+        :envvar:`COCOTB_TESTCASE` is renamed to :envvar:`COCOTB_TESTCASE`.
+
```
ktbarrett, line 206: This doesn't match up with the others.

ktbarrett (commented):
I don't think we should keep the old envvars around; they aren't intended to be a part of the API anymore. I think just a deprecated message mentioning the old variables name is sufficient.

docs/source/building.rst
teobiton: Oops, fixed it.

teobiton:
> I don't think we should keep the old envvars around; they aren't intended to be a part of the API anymore. I think just a deprecated message mentioning the old variables name is sufficient.

Okay but then where does the deprecation message go ? With the new variables ?

ktbarrett:
> With the new variables ?

Yeah. It's a feature related to the thing deprecated. Not all deprecated blcoks are deprecating the whole feature. Alternatively I like versionchanged. Or both? One versionchanged for the rename, another deprecation mentioning the old name is now a deprecated alias and will be removed.

ktbarrett (approved):

PR #3944, opened 2024-06-15
Add legacy `.buff` property to `LogicArray`

ktbarrett:
This turns out to be used quite a bit in some codebases. We can assume big-endian only since we only support these deprecated interfaces for the sole purpose of calling them on the result of `handle.value`, which in cocotb < 2.0 [always defaulted to](https://github.com/cocotb/cocotb/blob/05a50946ffafad7659a2f0683cd2a23ab7e12f5d/cocotb/handle.py#L909) [big endian byteorder](https://github.com/cocotb/cocotb/blob/05a50946ffafad7659a2f0683cd2a23ab7e12f5d/cocotb/binary.py#L140).

### DIFF
--- a/docs/source/newsfragments/3944.removal.rst
+++ b/docs/source/newsfragments/3944.removal.rst
@@ -1,1 +1,1 @@
-:attr:`LogicArray.buff <cocotb.types.LogicArray.buff>` has been deprecated. Use :meth:`LogicArray.to_unsigned() <cocotb.types.LogicArray.to_unsigned>` in combination with :meth:`int.to_bytes` as such ``v.to_unsigned().to_bytes(ceil(len(v) / 8), byteorder="big")`` instead.
+:attr:`LogicArray.buff <cocotb.types.LogicArray.buff>` has been deprecated. Use :meth:`LogicArray.to_unsigned() <cocotb.types.LogicArray.to_unsigned>` in combination with :meth:`int.to_bytes` instead. For example: ``v.to_unsigned().to_bytes(ceil(len(v) / 8), byteorder="big")``.
--- a/docs/source/newsfragments/3944.removal.rst
+++ b/docs/source/newsfragments/3944.removal.rst
@@ -1,0 +1,1 @@
+:attr:`LogicArray.buff <cocotb.types.LogicArray.buff>` has been deprecated. Use :meth:`LogicArray.to_unsigned() <cocotb.types.LogicArray.to_unsigned>` in combination with :meth:`int.to_bytes` as such ``v.to_unsigned().to_bytes(ceil(len(v) / 8), byteorder="big")`` instead.
--- a/src/cocotb/types/logic_array.py
+++ b/src/cocotb/types/logic_array.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import typing
 import warnings
+from math import ceil
 
 from cocotb._deprecation import deprecated
 from cocotb.types import ArrayLike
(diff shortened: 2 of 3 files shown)


ktbarrett:
Local pre-commit is not complaining. I think it's having trouble identifying what the "current" package is, so we should ignore it.

ktbarrett:
One thing to consider is adding `to_bytes` and `from_bytes` methods? They would be fairly simple to implement but may reduce verbosity if people are doing lots of bytes <-> LogicArray conversions.

```python
def to_bytes(self, *, byteorder: str) -> bytes:
    return self.to_unsigned().to_bytes(ceil(len(self) / 8), byteorder=byteorder)

@classmethod
def from_bytes(cls: Type[Self], value: bytes, *, byteorder: str) -> Self:
    return cls.from_unsigned(int.from_bytes(value, byteorder=byteorder, signed=False))
```

src/cocotb/types/logic_array.py
```diff
@@ -347,6 +348,23 @@ def signed_integer(self) -> int:
         """
         return self.to_signed()
 
+    @property
+    @deprecated(
+        '`.buff` property is deprecated. Use `v.to_unsigned().to_bytes(ceil(len(v) / 8), byteorder="big")` instead.'
+    )
+    def buff(self) -> bytes:
+        """Converts the value to :class:`bytes` by interpreting it as an unsigned integer in big-endian byteorder.
```
cmarqu, line 356:
```suggestion
        """Convert the value to :class:`bytes` by interpreting it as an unsigned integer in big-endian byteorder.
```

src/cocotb/types/logic_array.py
```diff
@@ -347,6 +348,23 @@ def signed_integer(self) -> int:
         """
         return self.to_signed()
 
+    @property
+    @deprecated(
+        '`.buff` property is deprecated. Use `v.to_unsigned().to_bytes(ceil(len(v) / 8), byteorder="big")` instead.'
+    )
+    def buff(self) -> bytes:
+        """Convert the value to :class:`bytes` by interpreting it as an unsigned integer in big-endian byteorder.
```
marlonjames, line 356:
```suggestion
        """Convert the value to :class:`bytes` by interpreting it as an unsigned integer in big-endian byte order.
```

src/cocotb/types/logic_array.py
```diff
@@ -347,6 +348,23 @@ def signed_integer(self) -> int:
...
+
+        The object is first converted to an :class:`int` as in :meth:`to_unsigned`.
+        Then the object is converted to :class:`bytes` by converting the resulting integer value as in :meth:`int.to_bytes`.
+        This assumes big-endian byteorder and the minimal number of bytes necessary to hold any value of the current object.
```
marlonjames, line 360:
```suggestion
        This assumes big-endian byte order and the minimal number of bytes necessary to hold any value of the current object.
```

docs/source/newsfragments/3944.removal.rst
```diff
@@ -0,0 +1 @@
+:attr:`LogicArray.buff <cocotb.types.LogicArray.buff>` has been deprecated. Use :meth:`LogicArray.to_unsigned() <cocotb.types.LogicArray.to_unsigned>` in combination with :meth:`int.to_bytes` as such ``v.to_unsigned().to_bytes(ceil(len(v) / 8), byteorder="big")`` instead.
```
marlonjames, line 1:
```suggestion
:attr:`LogicArray.buff <cocotb.types.LogicArray.buff>` has been deprecated. Use :meth:`LogicArray.to_unsigned() <cocotb.types.LogicArray.to_unsigned>` in combination with :meth:`int.to_bytes` instead. For example: ``v.to_unsigned().to_bytes(ceil(len(v) / 8), byteorder="big")``.
```

marlonjames (approved):

PR #3953, opened 2024-06-19
Use raw docstrings in handle.py when backslashes are present.

gilbertoabram:
This change fixes this error:

```
  File "cocotb/src/cocotb/handle.py", line 586
    """Action used to force a handle to a given value until a :class:`Release` is applied.
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: invalid escape sequence '\ '
```

which occurs in environments where `PYTHONWARNINGS=error`

### DIFF
--- a/src/cocotb/handle.py
+++ b/src/cocotb/handle.py
@@ -583,7 +583,7 @@ class Deposit(Generic[ValueT]):
 
 
 class Force(Generic[ValueT]):
-    """Action used to force a handle to a given value until a :class:`Release` is applied.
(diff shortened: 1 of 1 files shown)


ktbarrett (approved):

ktbarrett:
PEP257 recommends using raw strings over double escapes, and ruff has `D301` which enforces the usage of raw strings for docstrings when backslashes are used.

marlonjames (approved):

PR #3954, opened 2024-06-20
GpiClk: clock implemented on the simulator side.

gilbertoabram:
These changes implement #89. It passes many tests, however to fully work it needs #3874 first for proper cleanup in all cases (as noted in the original issue).

I don't fully understand what's needed from #2108, or if things have progressed in the meantime and that's no longer needed. One difference is that there the clock would run necessarily before any timed callback, whereas here it's implemented with timed callbacks, so the order relative to other CBs at the same time is not guaranteed. However, the current clocks in python also work via Timer and have the same limitation, so maybe it doesn't matter.

One notable difference with using this (and previous PRs' implementations for what I can tell) is that the toggling happens immediately, and not during ReadWrite. @ktbarrett pointed out that if using inertial writes the behavior should be identical, so one option is to use GpiClk automatically only if inertial writes are configured.

One reason in favor of still doing this in all cases (assuming tests pass) is that it's good for performance: the clock toggles before eval() is called, usually reducing the number of eval calls by one (although #3861 should fix that in most cases). And an HDL clock implementation is already different from the python clock in this regard, so if things are written in a way that is agnostic to where the clock is generated, it should also work with this.

### DIFF
--- a/src/cocotb/clock.py
+++ b/src/cocotb/clock.py
@@ -27,7 +27,6 @@
 
 """A clock class."""
 
-import itertools
 import logging
 from decimal import Decimal
 from fractions import Fraction
@@ -35,7 +34,8 @@ from logging import Logger
 from typing import Optional, Union
 
 from cocotb._py_compat import cached_property
-from cocotb.triggers import Timer
+from cocotb.simulator import clock_create
+from cocotb.triggers import Event, ReadOnly, Timer
 from cocotb.utils import get_sim_steps, get_time_from_sim_steps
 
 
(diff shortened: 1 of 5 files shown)


marlonjames:
This is not likely to be accepted. I would vote against putting this specialized code into the GPI layer, when it already provides the necessary features to implement.

See #2293 for a previous effort. As you noted, the missing piece is still the ability to stop and cleanup correctly from the Python side.

ktbarrett:
@gilbertoabram So the issue is that this code does not belong in the GPI because nothing it is doing requires it to be *in* the GPI, but could be implemented as a *user* of the GPI. i.e. it could be totally contained in simulatormodule.cpp (the PyGPI layer) or some other user library wrapped in Python.

@marlonjames Would it be fine if that were plugged into the `Clock` object rather than placed in a separate `CClock` object like in your PR?

marlonjames:
For 2.0 I think it would be fine to merge it and make a GPI-layer clock the default.
I would also probably remove the `cycles` argument when starting the clock and just make it go forever until stopped.

ktbarrett:
> I would also probably remove the cycles argument when starting the clock and just make it go forever until stopped.

Yeah that's one of many changes I would like to see before 2.0. Also, removing the dead attributes and having the driver manage it's own task.

PR #3956, opened 2024-06-20
Fix isort config

ktbarrett:
We manually list out first-party and third-party packages so that isort keeps cocotb and external packages in separate import sections. `pre-commit run -a` was run afterwards. This should address the errant lint failure mentioned [here](https://github.com/cocotb/cocotb/pull/3944#issuecomment-2169121047).


imphil (approved):
Thanks! The isort first-party/third-party detection is certainly the most fragile part of it.

PR #3960, opened 2024-06-20
Riviera-PRO: Always exit the compilation step

imphil:
Previously, we always redirected the output of Riviera-PRO in Makefiles,
which caused Riviera to unconditionally quit the compilation step. We
no longer do that, so we need to be explicit.

See also https://github.com/cocotb/cocotb/pull/3495 for the equivalent
change in the runner.


ktbarrett (commented):
I think I've heard people complain in the other direction. That in GUI mode it should *not* close immediately, since maybe you want to poke around after a failing test.

ktbarrett (approved):

cmarqu:
> nvm I read it wrong. Indentation is for nerds!

Fell into the same trap :)

cmarqu (approved):

imphil:
:smile: It took me a while as well to untangle the nested ifs.

PR #3964, opened 2024-06-21
CI: Run against Xcelium 2403

imphil:
Run against the latest version of Xcelium, 2403, in CI, and move the
2309 version to the extended tests bucket.


ktbarrett (approved):

imphil:
Boring, no new waivers or other fails :smile:

PR #3969, opened 2024-06-24
Remove outcome parameter to NullTrigger and remove handshaking between scheduler and Trigger to determine outcome

ktbarrett:
Final set of trigger updates. Depends on #3931.

- [x] Add newsfrags
- [x] Improve second commit's message

### DIFF
--- a/src/cocotb/_scheduler.py
+++ b/src/cocotb/_scheduler.py
@@ -54,7 +54,6 @@ from cocotb.triggers import (
     GPITrigger,
     Join,
     NextTimeStep,
-    NullTrigger,
     ReadOnly,
     ReadWrite,
     Timer,
@@ -805,9 +804,9 @@ class Scheduler:
                 except TypeError as exc:
                     # restart this task with an exception object telling it that
                     # it wasn't allowed to yield that
-                    result = NullTrigger(outcome=_outcomes.Error(exc))
-
-                self._resume_task_upon(task, result)
+                    self._queue(task, _outcomes.Error(exc))
+                else:
+                    self._resume_task_upon(task, result)
 
             # We do not return from here until pending threads have completed, but only
             # from the main thread, this seems like it could be problematic in cases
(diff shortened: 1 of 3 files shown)


src/cocotb/_scheduler.py
```diff
@@ -586,17 +584,18 @@ def _queue(
...
-                    f"Attempted rescheduling {task!r} with different outcome!"
+            if _debug:
+                self.log.debug(
+                    f"Rescheduling {task!r} with {outcome!r}. Previously scheduled with {self._pending_tasks[task]}"
                 )
-            elif _debug:
-                self.log.debug(f"Not rescheduling already scheduled {task!r}")
+            self._unschedule(
+                task
+            )  # this won't work currently =/ _unschedule does a lot more than it really should.
```
ktbarrett, line 597: This will work once `_unschedule` is refactored after `cancel` change comes in. We won't have to worry about scheduler reentrancy after that so we can move the test end checking stuff out of `_unschedule` into `_schedule` or something since that will be the only location that can call `_unschedule`.

src/cocotb/triggers.py
```diff
@@ -917,13 +906,7 @@ def on_done(ret: Outcome[Any]) -> None:
...
             w.kill()
 
-        # These lines are the way they are to make tracebacks readable:
-        #  - The comment helps the user understand why they are seeing the
-        #    traceback, even if it is obvious top cocotb maintainers.
-        #  - Using `NullTrigger` here instead of `result = completed[0].get()`
-        #    means we avoid inserting an `outcome.get` frame in the traceback
-        first_trigger = NullTrigger(outcome=completed[0])
-        return await first_trigger
+        return completed[0].get()
```
marlonjames, line 909: I never really understood this part.
What scenario and traceback was this concerned with?
Is it when the outcome is an `Error` and so `get()` raises?

ktbarrett: Yep, that's about it. It's just to avoid the `get()` call in the traceback.

ktbarrett: One thing to note is that in the coming future, `cocotb.start_soon` Tasks will not propagate exceptions, so this is kind of moot.

marlonjames (approved):

PR #3973, opened 2024-06-25
Improve Task's result TypeVar

marlonjames:
Adds `ResultType`. I think it makes the docs clearer.


ktbarrett (approved):

ktbarrett:
I can't wait until we can just use 3.12 type parameters, I hate having to document `TypeVar`s, but this is fine for now since the TypeVar is intended to be filled out by the user so they need to know.

PR #3974, opened 2024-06-25
Make sphinx not generate bogus undocumented return type in autoclass

ktbarrett:
This should remove the `Return type: ` statement from the description of `Join`. Apparently Sphinx thinks class definitions have return types that aren't the class type and it's losing its mind attempting to deduce it so we just turn that logic off and do it manually.


marlonjames:
`Return type:` goes from 77 hits to 16 on the library reference docs page.
In general I like the auto-generated return type info, so maybe as suggested in the sphinx issue we do whatever to override the `Join` autoclass directive causing the issue?

ktbarrett:
@marlonjames [From the issue](https://github.com/sphinx-doc/sphinx/issues/12472#issuecomment-2188255929):

> I'd say it's a bug but I don't really have time (nor the mood) to fix this one.

When we enable docstring checking in ruff, one of the checks includes enforcing the use of `Args`, `Returns`, etc. So we can compensate manually and won't have to worry about missing any since they are checked.

ktbarrett:
Seems like there's no feature for turning off the return type deduction either.

marlonjames:
Looking at the docs again, if we include
`:rtype: type` 
in the docstring we should be able to add back manually the ones we want.
Too bad we can't disable it just for classes where it fails.

marlonjames (approved):

PR #3983, opened 2024-06-26
GpiClock: C++ clock implementation using GPI.

### LINKED ISSUE #89: Move clock generation into C++

gilbertoabram:
This reimplements #3954 but in the simulator module without touching the GPI interface, and removing the `cycles` argument when starting a clock per the comments there. These notes apply here too:

> I don't fully understand what's needed from https://github.com/cocotb/cocotb/issues/2108, or if things have progressed in the meantime and that's no longer needed. One difference is that there the clock would run necessarily before any timed callback, whereas here it's implemented with timed callbacks, so the order relative to other CBs at the same time is not guaranteed. However, the current clocks in python also work via Timer and have the same limitation, so maybe it doesn't matter.
> 
> One notable difference with using this (and previous PRs' implementations for what I can tell) is that the toggling happens immediately, and not during ReadWrite. @ktbarrett pointed out that if using inertial writes the behavior should be identical, so one option is to use GpiClk automatically only if inertial writes are configured.
> 
> One reason in favor of still doing this in all cases (assuming tests pass) is that it's good for performance: the clock toggles before eval() is called, usually reducing the number of eval calls by one (although https://github.com/cocotb/cocotb/pull/3861 should fix that in most cases). And an HDL clock implementation is already different from the python clock in this regard, so if things are written in a way that is agnostic to where the clock is generated, it should also work with this.

I agree this is a better approach than #3954 for what this change is doing. My rationale for doing it in GPI was that it would be possible in a separate change to specialize GpiClk for the various underlying APIs, allowing some further performance gains, whereas doing this in the simulator means it's done as it is. I assumed that would be too much to be in a single PR, but if anyone thinks that makes sense when carried out to the full extent, I don't mind trying to implement it all the way.

That said, just removing the need to go in and out of python and for the extra eval will provide the majority of the performance gain, so in most real use cases the difference with a fully implemented GpiClk would be minimal/below measurement noise.

I thought earlier on I bumped into something in the tests that needed `CancelledError` in order to work, but it seems that the verilator tests all run to completion for me, so I'd be curious to see what in CI fails. That said, even if they all pass I still think that's required for it to work properly.

Closes #89.


ktbarrett:
My overarching comment is to not use `SimClk` or `simulator` in the name here. `simulatormodule.cpp` is a *terrible* name. This is the layer that provides an embedded Python interpreter and wraps the GPI in Python, so a better name would be PyGPI. That's the name I would use throughout.



> One notable difference with using this (and previous PRs' implementations for what I can tell) is that the toggling happens immediately, and not during ReadWrite. @ktbarrett pointed out that if using inertial writes the behavior should be identical, so one option is to use GpiClk automatically only if inertial writes are configured.

This will definitely need to be added before this comes in as well. We will have to leave the existing implementation in and only sub it out when `cocotb._conf.trust_interial` is True.

gilbertoabram:
I assumed that the existing inertial write test would cover this, but I realize now that it needs to be run explicitly. Do I need to add a test case using both impls explicitly?
Also I'm a bit surprised that the other checks succeed. Should I try to get another test in there that breaks due to lack of CancelledError (or figure out why it works without it)?

src/cocotb/clock.py
```diff
@@ -109,52 +115,61 @@ async def custom_clock():
...
-        self.signal = signal
-        self.coro = None
-        self.mcoro = None
+        self.t_high = self.period // 2
+
+        # If not using GpiClock, clkobj is None
+        if impl == "auto":
+            impl = "gpi" if cocotb._conf.trust_inertial else "py"
+        if impl == "gpi":
+            self.clkobj = clock_create(self.signal._handle)
```
ktbarrett, line 133: Namely here.

The issue is that if the `Clock` object goes out of scope, the `GpiClock` will be stopped and cleaned up as well. However, `Clock` currently allows itself to leak the actual clock driving task and this constitutes a change in behavior.

I think the easiest way to make it work like it used to is by wrapping it in a Task.

```python
self.clkobj = cocotb.start_soon(self._gpi_clock_runner())
```

```python
async def _gpi_clock_runner(self) -> None:
    gpi_clock = clock_create(self.signal._handle)
    await Event().wait() # Forever
    # when task is killed, gpi_clock will be cleaned up, stopping the clock
```

But more ideally once CancelledError works.
```python
async def _gpi_clock_runner(self) -> None:
    gpi_clock = clock_create(self.signal._handle)
    try:
        await Event().wait() # Forever
    finally:
        gpi_clock.stop()
```

ktbarrett: Nvm, you are already doing that.

src/cocotb/clock.py
```diff
@@ -109,52 +115,61 @@ async def custom_clock():
...
+            # The GPI clock will toggle for a given number of cycles, or forever,
+            # unless interrupted early (likely due to CancelledError)
+            try:
+                # The clock is meant to go on indefinitely, so if awaiting this
+                # should never complete. Await on an event that's never set.
+                e = Event()
+                await e.wait()
+            except Exception as e:
+                self.clkobj.stop()
+                raise e
```
ktbarrett, line 163: This is a bit cleaner.
```suggestion
             finally:
                self.clkobj.stop()
```

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1196,3 +1377,31 @@ PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
...
+    {NULL, NULL, 0, NULL} /* Sentinel */
+};
+
+template <>
+PyTypeObject gpi_hdl_Object<gpi_clk_hdl>::py_type = []() -> PyTypeObject {
+    auto type = fill_common_slots<gpi_clk_hdl>();
+    type.tp_name = "cocotb.simulator.GpiClock";
+    type.tp_doc = "C++ Clock using the GPI.";
+    type.tp_methods = gpi_clk_methods;
+    type.tp_dealloc = clock_dealloc;
```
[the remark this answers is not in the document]
ktbarrett: nvm.

src/cocotb/clock.py
```diff
@@ -109,52 +115,60 @@ async def custom_clock():
...
     ):
         self.signal = signal
         self.period = get_sim_steps(period, units)
-        self.half_period = get_sim_steps(period / 2, units)
         self.frequency = 1 / get_time_from_sim_steps(self.period, units="us")
-        self.hdl = None
-        self.signal = signal
-        self.coro = None
-        self.mcoro = None
+        self.t_high = self.period // 2
```
ktbarrett, line 127: Probably don't need to save this on the object.

src/cocotb/clock.py
```diff
@@ -109,52 +115,60 @@ async def custom_clock():
...
+
+        # If not using GpiClock, clkobj is None
+        if impl == "auto":
+            impl = "gpi" if cocotb._conf.trust_inertial else "py"
+        if impl == "gpi":
+            self.clkobj = clock_create(self.signal._handle)
```
ktbarrett, line 133: Probably also don't need to save this on the object, maybe just save `impl`.

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -838,6 +843,166 @@ static PyObject *log_level(PyObject *, PyObject *args) {
...
+    uint64_t t_high = 0;
+
+    int clk_val = 0;
+
+    int toggle();
+    static int toggle_cb(void *gpi_clk);
+};
+
+int GpiClock::start(uint64_t period_steps, uint64_t high_steps,
+                    uint64_t phase_steps) {
```
ktbarrett, line 872: I don't really understand this `phase_steps` argument. Why not just pass `start_high` through?

src/cocotb/clock.py
```diff
@@ -109,52 +115,60 @@ async def custom_clock():
...
                 Default is ``True``.
 
                 .. versionadded:: 1.3
         """
-        t = Timer(self.half_period)
-        if cycles is None:
-            it = itertools.count()
+        if self.clkobj:
+            phase = 0 if start_high else self.t_high
+            self.clkobj.start(self.period, self.t_high, phase)
```
ktbarrett, line 152: Check the output of the call to `start` here.

gilbertoabram: Can you add a bit more detail here? Specifically, I think a failure should always throw (added a test), and it seems that we would just let the exception propagate.

ktbarrett (approved):
This looks really good to me. 

This is being indirectly tested by all GHDL and NVC tests that use `Clock` because they are `COCOTB_TRUST_INERTIAL_WRITES` by default, so it's definitely being tested.

src/cocotb/clock.py
ktbarrett: No this is good. Other functions in this module return `None` rather than throw an exception, so I was making an incorrect assumption here. Looks good the way it is.

ktbarrett:
I have a feeling the failing tests are due to clocks not being stopped properly. I'm guessing there's a difference in how the GC works between 3.11 and 3.12 that prevents the destructor from running immediately.

gilbertoabram:
I replicated locally, and it seems that the last test I added causes the problem (artificial double start on the same GpiClock) exactly by not getting cleaned up and toggling the signal unexpectedly in subsequent tests.  I'll wait on #3874 since I suspect that is the underlying problem. I don't fully understand why it's happening in 3.12 only though (I checked that it also happens in verilator with 3.12, and it fails in the same exact way)

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -838,6 +843,165 @@ static PyObject *log_level(PyObject *, PyObject *args) {
...
+        PyErr_SetString(PyExc_TypeError, "Wrong type for clock_dealloc!");
+        return;
+        // LCOV_EXCL_STOP
+    }
+
+    GpiClock *gpi_clk = ((gpi_hdl_Object<gpi_clk_hdl> *)self)->hdl;
+
+    delete gpi_clk;
+
+    PyMem_Free(self);
```
marlonjames, line 980:
```suggestion
    Py_TYPE(self)->tp_free((PyObject *) self);
```
Or `PyObject_Del(self)` per https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_dealloc since this can't be subclassed.

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -838,6 +843,165 @@ static PyObject *log_level(PyObject *, PyObject *args) {
...
+    int start_high;
+
+    if (!PyArg_ParseTuple(args, "KKp:clk_start", &period, &t_high,
+                          &start_high)) {
+        return NULL;
+    }
+    if (self->hdl->start(period, t_high, start_high) != 0) {
+        PyErr_SetString(PyExc_RuntimeError, "Failed to start clock!\n");
+        return NULL;
+    }
```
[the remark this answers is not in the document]
gilbertoabram: I made `start()` return what error made it fail and used that to pick the right exceptions. Let me know if you prefer to check here instead.

gilbertoabram:
I realized that with the failing test not using the python Clock object, CancelledError would not fix anything, so I changed it to stop the clock explicitly with `finally`. I'm still not sure why with Python 3.12 the GpiClock object is not dealloc'd, but it is in the previous versions.

src/cocotb/clock.py
```diff
@@ -109,52 +117,62 @@ async def custom_clock():
...
-        self.half_period = get_sim_steps(period / 2, units)
         self.frequency = 1 / get_time_from_sim_steps(self.period, units="us")
-        self.hdl = None
-        self.signal = signal
-        self.coro = None
-        self.mcoro = None
+        if impl == "auto":
+            impl = "gpi" if cocotb._conf.trust_inertial else "py"
+        elif impl not in ["gpi", "py"]:
+            raise ValueError(f"Invalid clock impl '{impl}'")
```
cmarqu, line 132:
```suggestion
            raise ValueError(f"Invalid clock impl {impl!r}")
```
We might also print the supported impls here.

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1196,3 +1388,39 @@ PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
...
+               "stop() -> None\n"
+               "Stop this clock now.")},
+    {NULL, NULL, 0, NULL} /* Sentinel */
+};
+
+template <>
+PyTypeObject gpi_hdl_Object<gpi_clk_hdl>::py_type = []() -> PyTypeObject {
+    auto type = fill_common_slots<gpi_clk_hdl>();
+    type.tp_name = "cocotb.simulator.GpiClock";
+    type.tp_doc = "C++ Clock using the GPI.";
```
cmarqu, line 1422:
```suggestion
    type.tp_doc = "C++ clock using the GPI.";
```

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -992,6 +1171,14 @@ static PyMethodDef SimulatorMethods[] = {
                "--\n\n"
                "get_simulator_version() -> str\n"
                "Get the simulator's product version string.")},
+    {"clock_create", clock_create, METH_VARARGS,
+     PyDoc_STR("clock_create(signal, /)\n"
+               "--\n\n"
+               "clock_create(signal: cocotb.simulator.gpi_sim_hdl"
+               ") -> cocotb.simulator.GpiClock\n"
+               "Create a clock driver on a signal."
+               "\n"
```
cmarqu, line 1180:
```suggestion
               "\n"
               "\n"
```
Fix invalid markup (https://cocotb--3983.org.readthedocs.build/en/3983/library_reference.html#cocotb.simulator.clock_create)

src/cocotb/clock.py
```diff
@@ -63,6 +64,13 @@ class Clock:
             ``'step'``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
             When *units* is ``'step'``,
             the timestep is determined by the simulator (see :make:var:`COCOTB_HDL_TIMEPRECISION`).
+        impl: One of
+            ``'auto'``, ``'gpi'``, ``'py'``.
+            Specify whether the clock is toggled via a GpiClock (faster), or in python.
```
cmarqu, line 69:
```suggestion
            Specify whether the clock is implemented with a :class:`~cocotb.simulator.GpiClock` (faster), or with a Python coroutine.
```

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1196,3 +1388,39 @@ PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
...
+static PyMethodDef gpi_clk_methods[] = {
+    {"start", (PyCFunction)clk_start, METH_VARARGS,
+     PyDoc_STR("start($self, t_period, t_high, start_high)\n"
+               "--\n\n"
+               "start(t_period: int, t_high: int, start_high: bool) -> None\n"
+               "Start this clock now. The clock will have a period of "
+               "`t_period` steps, and out of that period it will be high "
+               "`t_high` steps. If `start_high` is True, start a the beginning "
+               "of the high state, otherwise start at the beginning of the "
+               "low state.\n"
```
cmarqu, line 1401:
```suggestion
               "low state.\n"
               "\n"
```
Hopefully fixing invalid markup (https://cocotb--3983.org.readthedocs.build/en/3983/library_reference.html#cocotb.simulator.GpiClock.start)

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -838,6 +844,170 @@ static PyObject *log_level(PyObject *, PyObject *args) {
...
+    int clk_val = 0;
+
+    int toggle();
+    static int toggle_cb(void *gpi_clk);
+};
+
+int GpiClock::start(uint64_t period_steps, uint64_t high_steps,
+                    bool start_high) {
+    if (clk_toggle_cb_hdl) {
+        LOG_ERROR("Failed to start clock: already started -- stop first");
```
marlonjames, line 877: I don't think these error log messages should be here. The Python exceptions should cover the ones in `start()` at least.

@ktbarrett thoughts?

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1196,3 +1388,40 @@ PyTypeObject gpi_hdl_Object<gpi_cb_hdl>::py_type = []() -> PyTypeObject {
...
+               "low state.\n"
+               "\n"
+               "Raises:\n"
+               "    TypeError: wrong arguments type.\n"
+               "    ValueError: t_period and t_high are such that in one "
+               "period the duration of the low or high state would be less "
+               "than one step.\n"
+               "    RuntimeError: the clock was already started, or the "
+               "underlying callback could not be registered.\n"
+               "\n")},
```
marlonjames, line 1410:
```suggestion
     PyDoc_STR("start($self, period_steps, high_steps, start_high)\n"
               "--\n\n"
               "start(period_steps: int, high_steps: int, start_high: bool) -> None\n"
               "Start this clock now.\n"
               "\n"
               "The clock will have a period of *period_steps* time steps, "
               "and out of that period it will be high for *high_steps* time steps. "
               "If *start_high* is ``True``, start at the beginning of the high state, "
               "otherwise start at the beginning of the low state.\n"
               "\n"
               "Raises:\n"
               "    TypeError: If there are an incorrect number of arguments or "
               "they are of the wrong type.\n"
               "    ValueError: If *period_steps* and *high_steps* are such that in one "
               "period the duration of the low or high state would be less "
               "than one time step, or *high_steps* is greater than *period_steps*.\n"
               "    RuntimeError: If the clock was already started, or the "
               "GPI callback could not be registered.")},
```

marlonjames (commented):
Apologies if the PR reviews feel nitpicky, but thanks for working through them.

src/cocotb/share/lib/simulator/simulatormodule.cpp
gilbertoabram: Makes sense. I left it for subsequent toggles in case it fails there, otherwise the clock would just silently die, but removed them for start

gilbertoabram:
> Apologies if the PR reviews feel nitpicky, but thanks for working through them.

No worries, thanks for reviewing!

src/cocotb/clock.py
```diff
@@ -63,6 +64,12 @@ class Clock:
...
             When *units* is ``'step'``,
             the timestep is determined by the simulator (see :make:var:`COCOTB_HDL_TIMEPRECISION`).
+        impl: One of
+            ``'auto'``, ``'gpi'``, ``'py'``.
+            Specify whether the clock is implemented with a :class:`~cocotb.simulator.GpiClock` (faster), or with a Python coroutine.
+            When ``'auto'`` is used (default), the fastest implementation that supports your environment and use case is picked.
+
+            .. versionadded:: 2.0
 
     If you need more features like a phase shift and an asymmetric duty cycle,
```
marlonjames, line 74:
```suggestion
    When *impl* is ``'auto'``, if :envvar:`COCOTB_TRUST_INERTIAL_WRITES` is defined,
    the :class:`~cocotb.simulator.GpiClock` implementation will be used.
    Otherwise, the Python coroutine implementation will be used.
    See the environment variable documentation for more information on the consequences
    of using the simulator's inertial write mechanism.

    If you need more features like a phase shift and an asymmetric duty cycle,
```

marlonjames (commented):
Needs a good newsfragment.

docs/source/newsfragments/3983.feature.rst
```diff
@@ -0,0 +1 @@
+Added :class:`GpiClock`, a C++ implementation of a simple clock, exposed as :class:`~cocotb.simulator.GpiClock`. It has the advantage of higher performance due to not having to wake up a python task on a Timer to then enqueue a write. The existing :class:`~cocotb.clock.Clock` uses it automatically when it would behave identically to the python implementation.
```
cmarqu, line 1:
```suggestion
Added :class:`cocotb.simulator.GpiClock`, a C++ clock generator implementation with higher performance due to less handshaking between Python and the GPI. :class:`~cocotb.clock.Clock` uses it automatically when it would behave identically to the Python implementation.
```

cmarqu (approved):

docs/source/newsfragments/3983.removal.rst
```diff
@@ -0,0 +1 @@
+Removed the ``cycles`` argument to :meth:`Clock.start() <cocotb.clock.Clock.start>`. Use ``kill()`` on the clock task instead, or implement manually.
```
marlonjames, line 1:
```suggestion
Removed the ``cycles`` argument to :meth:`Clock.start() <cocotb.clock.Clock.start>`. Use :meth:`~cocotb.task.Task.kill` on the clock task instead, or implement manually.
```

marlonjames (commented):
I think this is the last thing for me.

PR #3984, opened 2024-06-26
Improve trigger coverage

marlonjames:
Cleanup `Lock` to use same path for acquiring when not locked and also when released and there is a Task pending.
Add tests for missing trigger coverage.


ktbarrett (approved):

marlonjames:
Trying without `NextTimeStep()` test.

ktbarrett (approved):

PR #3992, opened 2024-06-27
CI: Use case-insensitive regex match for repr strings that contain handle paths

marlonjames:
I don't know if updating NVC version in CI caused these to fail, but let's just use case-insensitive matching here.
These tests are for coverage anyway, and everything except the path comes from cocotb.


ktbarrett (approved):
As long as we don't make it a thing.

PR #3997, opened 2024-07-01
Friendly pytest message

raysalemi:
I found this old change to an error message. I want to see whether it is still needed.


raysalemi:
This is a pointless pull request.

PR #4001, opened 2024-07-02
Instruct users to install a 1.x version by default

cmarqu:

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,10 @@ The current stable version of cocotb requires:
 [GHDL](https://docs.cocotb.org/en/stable/simulator_support.html#ghdl) or
 [other simulator](https://docs.cocotb.org/en/stable/simulator_support.html))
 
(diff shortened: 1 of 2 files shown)


cmarqu:
@marlonjames Is the update more like what you had in mind?

imphil:
Thanks for taking the initiative here, @cmarqu. It's slightly annoying business, and after some more thought, I came up with the following:

* I'd keep the README (also shown on PyPi) as-is for our master branch, but only update it in the stable branch. I filed https://github.com/cocotb/cocotb/pull/4013 for that.
* For the Sphinx documentation, we need something a bit more dynamic. I opened https://github.com/cocotb/cocotb/pull/4012 for that and will backport this change to stable/1.9 once merged into master.

PTAL at those PRs and let me know what you think!

imphil:
Went with the alternative approach, closing this one.

PR #4002, opened 2024-07-02
Use uv to build venvs, virtualenv as fallback

cmarqu:


cmarqu:
We still support Python 3.6.7 which selects the too old nox-2022.1.7 which doesn't implement the feature used.

ktbarrett:
Also, I think the builtin venv is probably a better backup than virtualenv.

PR #4012, opened 2024-07-08
Pin the version in the stable install docs

imphil:
When building documentation for a tagged release in ReadTheDocs, pin the
version we suggest users to install to a version matching the
documentation they are reading.

For example, a ReadTheDocs build for the v1.9.0 tag would suggest users
to run "pip install cocotb =~ 1.9.0", which gives them a 1.9.x version
matching the documentation they are reading.

For all other Sphinx build types (local builds, PR builds, "latest" RTD
builds, etc.), the documentation stays as is: it gives instructions how
to install the latest stable release (whatever that might be currently),
and also shows how to install a development release.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -92,6 +92,15 @@ try:
 except subprocess.CalledProcessError:
     commit_id = "master"
 
+# Is this documentation build a ReadTheDocs build for a git tag, i.e., a
+# release? Set the 'is_release_build' tag then, which can be used by the
+# '.. only::' directive.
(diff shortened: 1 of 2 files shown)


imphil:
Alternative to https://github.com/cocotb/cocotb/pull/4001 for the Sphinx part.

Screenshot of the documentation built with `READTHEDOCS=1 READTHEDOCS_VERSION_TYPE=tag nox -s docs_preview`:

cocotb-rtd-tag-bild

cmarqu (approved):

ktbarrett (approved):

PR #4014, opened 2024-07-08
CI: Run against Verilator 5.026

imphil:
Update CI to the latest release version, as usual.

### DIFF
--- a/.github/generate-envs.py
+++ b/.github/generate-envs.py
@@ -173,7 +173,7 @@ ENVS = [
     {
         "lang": "verilog",
         "sim": "verilator",
-        "sim-version": "v5.024",  # Latest release version.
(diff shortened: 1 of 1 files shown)


ktbarrett:
Welp. Maybe it doesn't work. There's also the issues with `--timing`. I don't think we can support v5.026 yet.

imphil:
It was worth a try :smile: I'll do a documentation update for stable/1.9 and leave the rest of the debugging for a potential later patch release.

ktbarrett:
Try this on 1.9. I think 1.9 *may* work. The MacOS test might fail, but that ends up being an environment issue.

ktbarrett:
Closing in favor of #3861.

PR #4017, opened 2024-07-08
stable/1.9: Update Verilator CI to latest stable release

imphil:
Use the latest stable release of Verilator in CI, and test all supported versions.

- **Verilator: Use latest release in CI, add minimum supported version to extended tests (#3827)**
- **Test Verilator 5.026 in CI**

### DIFF
--- a/.github/generate-envs.py
+++ b/.github/generate-envs.py
@@ -141,7 +141,7 @@ ENVS = [
     {
         "lang": "verilog",
         "sim": "verilator",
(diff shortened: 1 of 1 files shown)


ktbarrett:
Looks like the Verilator issue is still not resolved on MacOS, but I think it works well enough to day that we support v5.026 on 1.9.

ktbarrett (approved):

PR #4022, opened 2024-07-10
Make sim_phase a global variable

ktbarrett:
This should allow users to determine *roughly* what state they are in, and should help us prevent issues in the future when we prevent illegal timing transitions.

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -313,6 +313,8 @@ Other Runtime Information
 
 .. autodata:: cocotb.is_simulation
 
+.. autodata:: cocotb.sim_phase
+
 .. _combine-results:
 
 The ``combine_results`` script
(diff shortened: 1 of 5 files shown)


docs/source/library_reference.rst
```diff
@@ -313,6 +313,8 @@ Other Runtime Information
 
 .. autodata:: cocotb.is_simulation
 
+.. autodata:: cocotb.sim_phase
+
```
marlonjames, line 317:
```suggestion
.. autodata:: cocotb.sim_phase

.. autoenum:: cocotb.SimPhase

```

marlonjames (approved):
Fine with me.

ktbarrett:
Gonna wait until #4023 is in, so we have a cleaner history (since that PR must be merged rather than rebased).

PR #4023, opened 2024-07-11
Split utils

ktbarrett:
Split cocotb.utils into two packages: one for simulation time utility functions, and the other for implementation utilities that are meant to be private.


src/cocotb/_utils.py
```diff
@@ -28,16 +28,13 @@
 """Collection of handy functions."""
```
marlonjames, line 28: Maybe update this module docstring?

marlonjames (approved):

PR #4024, opened 2024-07-11
Remove `TriggerException`

ktbarrett:
This was not documented, but was a necessary part of the public API as if a trigger failed to register, the user saw this exception. We can use the less specific `RuntimeError` as I don't think this error is really recoverable, so there is no reason to differentiate it from other types of exceptions.

### DIFF
--- a/docs/source/newsfragments/4024.removal.rst
+++ b/docs/source/newsfragments/4024.removal.rst
@@ -1,0 +1,1 @@
+``cocotb.triggers.TriggerException`` was removed. :exc:`RuntimeError` is thrown in its place.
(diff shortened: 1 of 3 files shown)


marlonjames (approved):

tests/test_cases/test_cocotb/test_timing_triggers.py
```diff
@@ -30,7 +30,7 @@
     RisingEdge,
     Timer,
 )
-from cocotb.triggers import _TriggerException as TriggerException
+from cocotb.utils import get_sim_time
```
marlonjames, line 33:
```suggestion
from cocotb.sim_time_utils import get_sim_time
```

docs/source/newsfragments/4024.removal.rst
```diff
@@ -0,0 +1 @@
+``cocotb.triggers.TriggerException`` was removed. :exc:`RuntimeError` is thrown in its place.
```
cmarqu, line 1:
```suggestion
The undocumented ``cocotb.triggers.TriggerException``, raised when a trigger failed to register, was removed. :exc:`RuntimeError` is thrown in its place.
```

PR #4031, opened 2024-07-12
Docs: re-order Python Code Library Reference page

marlonjames:
Fix the `package_iterate()` docstring and re-order the Python reference page a bit.

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -23,39 +23,69 @@ Python Test Runner
 .. warning::
     Python runners and associated APIs are an experimental feature and subject to change.
 
-.. automodule:: cocotb_tools.runner
-    :members:
-    :member-order: bysource
+.. currentmodule:: cocotb_tools.runner
+
+.. module:: cocotb_tools.runner
     :synopsis: Build HDL and run cocotb tests.
 
+.. autofunction:: get_runner
 
-Test Results
-============
-
-The exceptions in this module can be raised at any point by any code and will terminate the test.
-
-.. automodule:: cocotb.result
+.. autoclass:: Runner
     :members:
-    :member-order: bysource
-    :synopsis: Exceptions and functions for simulation result handling.
+
+.. autoclass:: VHDL
+
+.. autoclass:: Verilog
+
+Simulator Runners
+-----------------
+
+.. autoclass:: Icarus
+
+.. autoclass:: Verilator
+
+.. autoclass:: Riviera
+
+.. autoclass:: Questa
+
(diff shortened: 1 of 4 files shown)


cmarqu:
While you are at it, could you fix the "Implementation of Runner for GHDL." where "GHDL" is used for almost all simulators, e.g. here: https://cocotb--4031.org.readthedocs.build/en/4031/library_reference.html#cocotb_tools.runner.Verilator

docs/source/library_reference.rst
```diff
@@ -23,40 +23,81 @@ Python Test Runner
...
+------------
+
+The exceptions in this module can be raised at any point by any code and will terminate the test.
+
+.. automodule:: cocotb.result
+    :members:
+    :member-order: bysource
+    :synopsis: Exceptions and functions for simulation result handling.
+
+
```
ktbarrett, line 100: All of these are being removed by 2.0.

docs/source/library_reference.rst
```diff
@@ -182,17 +223,35 @@ The following are internal classes used within ``cocotb``.
...
+Using Synchronous Functions
+---------------------------
+
+.. autofunction:: cocotb.external
+
+.. autofunction:: cocotb.function
+
+
```
ktbarrett, line 254: The plan is to keep these undocumented as we really don't want people using them. I will probably deprecate them, but leave them in cocotb until they get in the way. A replacement for `cocotb.external` that acts more like what most users want will be introduced.

docs/source/library_reference.rst
```diff
@@ -23,40 +23,70 @@ Python Test Runner
 .. warning::
     Python runners and associated APIs are an experimental feature and subject to change.
 
-.. automodule:: cocotb_tools.runner
-    :members:
-    :member-order: bysource
+.. currentmodule:: cocotb_tools.runner
```
ktbarrett, line 26: Another way I just learned about is starting the ref content with a `.`, e.g. ``:meth:`.Runner.build` ``. It may or may not help, haven't tried. I was gonna do that this morning to fix the broken links in the runner docs.

marlonjames: I can't find documentation on the dot prefix. What does it do?

ktbarrett: Yeah it's hidden...

marlonjames: That fixed it locally. Let's see.

ktbarrett: Looks like it works.

ktbarrett (approved):

PR #4033, opened 2024-07-12
Replace deprecated VCS options

AtaraxiaZ:
VCS requires users to replace `-debug` option as shown belowed.

```
Warning-[DEBUG_DEP] Option will be deprecated
  The option '-debug' will be deprecated in a future release.  Please use
  '-debug_acc+pp+f+dmptf -debug_region+cell+encrypt' instead.
```


ktbarrett:
When were the new set of options introduced? Has it been around a while? I don't want to force all users to use the newest 2024 version of the  simulator just because an option is deprecated.

ktbarrett (approved):

Issue #3832, opened 2024-04-07
Setting value on forced signal causes implicit release on Riviera


ktbarrett:
In `test_force_release.test_cocotb_writes_dont_overwrite_force_registered` when the forced signal is erroneously attempted to be overwritten using a normal deposit, the deposit does not succeed, but it does implicitly `Release` the signal. This causes the process to update the driven signal prematurely causing a test failure.

```
# COUT:    110.00ns INFO     cocotb.regression                  running test_force_release.test_cocotb_writes_dont_overwrite_force_registered (6/7)
# COUT:                                                             Test Deposits following a Force don't overwrite the value.
# COUT:    150.01ns INFO     cocotb.regression                  test_force_release.test_cocotb_writes_dont_overwrite_force_registered failed
# COUT:                                                         Traceback (most recent call last):
# COUT:                                                           File "/opt/actions-runner/_work/cocotb/cocotb/tests/test_cases/test_force_release/test_force_release.py", line 116, in test_cocotb_writes_dont_overwrite_force_registered
# COUT:                                                             assert dut.stream_out_data_registered.value == 10
# COUT:                                                         AssertionError: assert LogicArray('01001101', Range(7, 'downto', 0)) == 10
# COUT:                                                          +  where LogicArray('01001101', Range(7, 'downto', 0)) = LogicObject(sample_module.stream_out_data_registered).value
# COUT:                                                          +    where LogicObject(sample_module.stream_out_data_registered) = HierarchyObject(sample_module).stream_out_data_registered
```

## To reproduce this issue
1. Setup a cocotb development environment following https://docs.cocotb.org/en/stable/troubleshooting.html#building-cocotb-in-development-mode
2. Check out git commit `cdbf507cbb0bdbfdf19b65a6b6516e5a1fbb074a`
3. Remove the test waiver to reveal the original error:
    ```diff
	diff --git a/tests/test_cases/test_force_release/test_force_release.py b/tests/test_cases/test_force_release/test_force_release.py
	index af6991cb..bb13fb67 100644
	--- a/tests/test_cases/test_force_release/test_force_release.py
	+++ b/tests/test_cases/test_force_release/test_force_release.py
	@@ -140,7 +140,7 @@ async def test_cocotb_writes_dont_overwrite_force_combo(dut):
	 # Riviera's VPI implicitly releases signal when overwriting forced signal with normal deposit (gh-3832)
	 # Questa's FLI allows overwriting forced signal with normal deposit (gh-3833)
[...]

PR #4034, opened 2024-07-12
add bytes access to logic handles

toddstrader:
Allow for setting and getting of signals from / to `bytes` objects.  Setting is done via the existing `.value` property and getting is done via a new `value_as_bytes()` method.

I'm aware this would need documentation and a news fragment, but I'd like to make sure we're solid on the interface first before writing those up.  Please advise if the interface is acceptable and then I'll add these words and take this PR out of draft status.


toddstrader:
Mostly cleaned up now.  Looks like Questa and Riviera are having problems with one of my new tests: `test_struct:test_packed_struct_bytes`.  Perhaps they don't like VPI getting or setting a struct with `vpiVectorVal`.  Can someone provide logs so I can debug further?

ktbarrett:
Questa passes `test_packed_struct_bytes`.

Riviera gives this:
```
# VSIM: Error: Fatal error occurred during simulation.
```
I'm guessing it doesn't like setting packed structs using bytes.

toddstrader:
OK, yeah.  It wasn't clear from the logs if Questa was mad at my new test.  So if it's not that, what is it failing on?

As far as Riviera goes, I'm personally fine with skipping that test for it, making a note in the simulator support section and leaving it to future explorers.

ktbarrett:
> So if it's not that, what is it failing on?

If you expect failure or error and it doesn't fail or error, that's an error.

toddstrader:
Ah, I see.  Probably copypasta from other tests in that file:
```
expect_fail=SIM_NAME.startswith("modelsim"),
```
I'll mess around with that next.

imphil:
> As far as Riviera goes, I'm personally fine with skipping that test for it, making a note in the simulator support section and leaving it to future explorers.

You can skip the test, but please file an issue in the cocotb issue tracker with steps to reproduce this issue *for an Aldec AE*. Here's an example of how that could look: https://github.com/cocotb/cocotb/issues/3832. We are filing upstream issues for all Xcelium, Riviera-PRO and Questa whenever we discover something odd in the interaction between cocotb and the simulator.

toddstrader:
> > As far as Riviera goes, I'm personally fine with skipping that test for it, making a note in the simulator support section and leaving it to future explorers.
> 
> You can skip the test, but please file an issue in the cocotb issue tracker with steps to reproduce this issue _for an Aldec AE_. Here's an example of how that could look: #3832. We are filing upstream issues for all Xcelium, Riviera-PRO and Questa whenever we discover something odd in the interaction between cocotb and the simulator.

Hmm, I had guessed that getting or setting a packed struct with `vpiVectorValue` was what was making Riviera sad.  But it doesn't appear it's that simple:
https://edaplayground.com/x/CRSh

So yeah, I guess we'll have to rely on the cocotb test itself as the reproducer for now since I don't have access to Riviera myself.

src/cocotb/handle.py
```diff
@@ -565,6 +565,13 @@ class _GPISetAction(enum.IntEnum):
     RELEASE = 2
 
 
+class _GPIResolveX(enum.IntEnum):
```
ktbarrett, line 568: This is private, but it is something that users are expected to use? I might instead use `DocEnum` to document this, while also inheriting from `str` to mimic `StrEnum`.

toddstrader: I changed this to `DocEnum`, but did not inherit from `str` as the `int` value gets passed through to the GPI.  However, I'm not sure I'm using `DocEnum` correctly yet.  I've added this class to the docs, but the enum values aren't enumerated or described.

Also, I removed the `_` since this is user facing.

ktbarrett: Yeah, using `int` makes more sense here now that I have a better look at it.

ktbarrett:
ResolveX is not a good idea or something we want in cocotb going forward. And following on from that I don't think I have any interest or see any value in seeing a RNG and `gpi_seed` added to the GPI.

All bytes getting should be equivalent to, but faster than the equivalent operation:
```python
dut.handle.value.to_unsigned().to_bytes(ceil(len(dut.handle) / 8), byteorder=...)
```
That is: it fails if there are Xs. You should only be using this function, like you should only be using the above statement, if you *know* the data is binary.

An easy way to support this personally is to set up getting bytes to fail fast then add an exception handler for the slow path. It's fast when it works and slow, but still correct, when it doesn't work. This can be easily encapsulated into a function for reuse.
```python
 try:
    return handle.get_value_bytes()
except ValueError:
    return handle.value  # or something else
```

Finally, I think that means that `value_as_bytes` should take a `byteorder` argument. This should probably default to `"big"` since I figure that people writing `b"\x01\x02"` will expect the pattern `0000000100000010` to be written. And it's also the same byteorder that was default when using `BinaryValue.buff`.

Likewise, setting the value should be a faster version of:
```python
dut.handle.value = LogicArray.from_unsigned(int.from_bytes(value, signed=False, byteorder=...), range=Range(len(value)*8-1, 0))
```

Setting values of the incorrect size fails, and does not implicitly extend, nor truncate the value. This goes for implicit truncation of signals that are not multiples of byte width which isn't handled here. We don't want truncation, nor do we want to decide when truncated values are significant or insignificant. (Are truncated 1s worthy of error? Maybe they are just sign extension bits and are safely ignored? Maybe they aren't? We don't knowing and should refuse the temptation to guess or decide for the user.) I think this is fine since people are generally only going to be using these functions on handles that are byte-oriented, so we just need a check to ensure these function aren't being used handle that aren't byte-oriented or aren't the same size as the input data.

And of course, it would be nice to support that same `byteorder` argument defaulted to `"big"`.

PR #4038, opened 2024-07-12
Cleanup handle (part ???)

ktbarrett:
* Support assigning `str` to `LogicObject`s
* Adds support for `left`, `right`, `direction`, and `range` to `ArrayObject`, `LogicObject` and `StringObject`.

Split from #3777.


marlonjames (commented):
For the documentation, I think it would be good to show these inherited attributes on `ArrayObject`, `LogicObject`, and `StringObject`.

cmarqu (approved):

ktbarrett:
@marlonjames I will take that as a general comment to use `:inherited-members:` for all relevant inherited members rather than attempt to document the inheritance. This is the ideal I'm looking for. I'm not quite done with this module yet and I think that will be a follow on.

PR #4041, opened 2024-07-13
Add done_callback interface to `Task`s

ktbarrett:
Adds `add_done_callback` and `remove_done_callback` as seen in asyncio's Tasks with the hopes they may be used in the future for instrumenting test end behaviors rather than baking them into the scheduler, but also as a potentially useful feature for users.

### DIFF
--- a/docs/source/newsfragments/4041.feature.rst
+++ b/docs/source/newsfragments/4041.feature.rst
@@ -1,0 +1,1 @@
+Added :meth:`.Task.add_done_callback` and :meth:`.Task.remove_done_callback` to support running user-supplied callbacks when :class:`~cocotb.task.Task`\ s become "done".
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -7,7 +7,7 @@ import logging
 import os
 import warnings
 from asyncio import CancelledError, InvalidStateError
-from typing import Any, Coroutine, Generator, Generic, Optional, TypeVar
+from typing import Any, Callable, Coroutine, Generator, Generic, List, Optional, TypeVar
 
 import cocotb
 import cocotb.triggers
@@ -61,6 +61,7 @@ class Task(Generic[ResultType]):
         self._outcome: Optional[Outcome[ResultType]] = None
         self._trigger: Optional[cocotb.triggers.Trigger] = None
         self._cancelled: Optional[CancelledError] = None
+        self._done_callbacks: List[Callable[[Task[Any]], Any]] = []
 
         self._task_id = self._id_count
         type(self)._id_count += 1
@@ -139,6 +140,10 @@ class Task(Generic[ResultType]):
         except BaseException as e:
             self._outcome = Error(remove_traceback_frames(e, ["_advance", "send"]))
 
+        if self.done():
+            for callback in self._done_callbacks:
+                callback(self)
+
     def kill(self) -> None:
         """Kill a coroutine."""
         if self._outcome is not None:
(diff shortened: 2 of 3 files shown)


src/cocotb/task.py
```diff
@@ -233,6 +242,35 @@ def exception(self) -> Optional[BaseException]:
         else:
             return None
 
+    def add_done_callback(self, callback: Callable[["Task[ResultType]"], Any]) -> None:
+        """Add a callback to run once the Task becomes "done".
```
cmarqu, line 246:
```suggestion
        """Add *callback* to the list of callbacks to be run once the Task becomes "done".
```

src/cocotb/task.py
```diff
@@ -233,6 +242,35 @@ def exception(self) -> Optional[BaseException]:
...
+
+        Args:
+            callback: The callback to run on "done".
```
cmarqu, line 249:
```suggestion
            callback: The callback to add to the "run-on-done" list.
```

src/cocotb/task.py
```diff
@@ -233,6 +242,35 @@ def exception(self) -> Optional[BaseException]:
...
+        """
+        self._done_callbacks.append(callback)
+
+    def remove_done_callback(
+        self, callback: Callable[["Task[ResultType]"], Any]
+    ) -> int:
+        """Remove a callback added by a call to :meth:`add_done_callback`.
+
+        Args:
+            callback: The callback to remove.
```
cmarqu, line 259:
```suggestion
            callback: The callback to remove from the "run-on-done" list.
```

src/cocotb/task.py
```diff
@@ -233,6 +242,35 @@ def exception(self) -> Optional[BaseException]:
...
+
+        Returns:
+            The number of callbacks removed.
```
cmarqu, line 262:
```suggestion
            The number of callbacks that were removed.
```

docs/source/newsfragments/4041.feature.rst
```diff
@@ -0,0 +1 @@
+Added :meth:`.Task.add_done_callback` and :meth:`.Task.remove_done_callback`.
```
cmarqu, line 1:
```suggestion
Added :meth:`.Task.add_done_callback` and :meth:`.Task.remove_done_callback` to support callbacks that are run when the Task becomes "done".
```

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -858,3 +858,83 @@ async def noop():
...
+        callback_ran = True
+
+    async def passing_coro() -> None:
+        await Timer(10, "ns")
+
+    passing_task = cocotb.start_soon(passing_coro())
+    passing_task.add_done_callback(done_callback)
+    await Timer(1, "ns")
+    assert not callback_ran
+    passing_task.remove_done_callback(done_callback)
```
cmarqu, line 938: Check that this returns `1`.

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -858,3 +858,83 @@ async def noop():
...
+    await passing_task
+    assert not callback_ran
```
cmarqu, line 940: Add another call `passing_task.remove_done_callback(done_callback)` that should return `0`.

src/cocotb/task.py
ktbarrett: I think specifying `"run-on-done"` is maybe a bit too flowery. I used https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.add_done_callback as guidance.

src/cocotb/task.py
ktbarrett: Again [here](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future.remove_done_callback).

cmarqu (approved):

PR #4044, opened 2024-07-16
Remove support for Python 3.6 and 3.7

### LINKED ISSUE #3730: Increase minimum supported Python version to 3.8
Python 3.6 and 3.7 are EoL. The reason we are supporting 3.6 as the minimum supported version is because RHEL8's default Python version is 3.6, but RHEL 8 is EoL in May of this year (2024). Ubuntu 20.04 is still in support until May of next year (2025) and uses Python 3.8 as the minimum version. So Python 3.8 seems like a reasonable next minimum version to support for following versions.

After the release of 2.0 (and probably the 2.1 hotfix), we should increase the minimum supported version of Python to 3.8.

With 3.8 we have access to the following features:
* postponed typing annotation evaluation which should lead to faster start-up times
* typing features `Protocol`, `TypedDict`, and `Literal` (could improve typing in existing codebase)
* position-only parameters (could improve typing in existing codebase)
* PEP587 C API initialization improvements (could be used to impl #3445)
* Remove all version-compat code in `cocotb._py_compat`

ktbarrett:
Allowed us to remove a lot of version dependent code and the whole `cocotb._py_compat` module. Also we could bump the minumum version of C++ to C++14. The jump to C++14 unfortunately caused a ripple effect that made me have to bother with a lot of code's signedness.

Closes #3730.


cmarqu:
You could also run `pyupgrade --py38-plus`.

cmarqu (approved):

ktbarrett:
> pyupgrade --py38-plus

Does `ruff` do that?

imphil:
I'd like to understand the trade-offs of this change better. On the one hand, any reduction in version compatibility makes cocotb harder to use for our users. It might not be a blocker (and rarely is), but it's still annoying and has a chance to reduce the uptake of cocotb 2.0. So without being able to exactly quantify the cost, I'd like to better understand what we gain.

https://github.com/cocotb/cocotb/issues/3730 lists some bullet points:
> postponed typing annotation evaluation which should lead to faster start-up times

That's an incentive to use Python 3.8+, not a requirement we need to impose on the users?

> typing features Protocol, TypedDict, and Literal (could improve typing in existing codebase)

Seems like a minor improvement.

> position-only parameters (could improve typing in existing codebase)

Do we have a strong need for those?

> PEP587 C API initialization improvements

It's something we could ifdef, I assume?

> Remove all version-compat code in cocotb._py_compat

It's minimal effort to keep this already existing code alive.


So in this summary, it seems to me we should keep 3.6 compatibility for a while longer. Is there anything else we gain, or that's even blocking further progress for your work, @ktbarrett?




For C++14 I need to have a closer look at the compiler support matrix in various distributions we support, I'll get back on that. But already in preparation for that: what are we gaining?

ktbarrett:
The main justification for doing the change *now* was that I was running into dependency cycles (causing `ImportError`s) incurred by typing annotations in #4055. By using `from __future__ import annotations` (#4057) we postpone all type resolution. That should forever avoid these kinds of issues, of which there are already several and I see several more in the near future. The alternative is to play whack-a-mole updating type annotations to be strings (thus postponing them) whenever we run into issues. AFAIK there is no tool that  updates all type annotations to use string-based postponing (because why would anyone do that since `from __future__ import annotations` exists?).

But in addition to the above list, using `from __future__ import annotations` also allows us to use newer Python typing syntax and features. Things like `Protocol`, `ParamSpec`, `Literal`, `@final`, etc. can be used by simply guarding the imports with `if typing.TYPE_CHECKING` *because* type anontations aren't evaluated at runtime. This can be seen in #4057 where `ruff` updates tons of typing annotations to use `TypeA | TypeB` over `Union`, which wasn't introduced until Python 3.10.

`from __future__ import annotations` requires Python 3.7, but I figured if we are dropping 3.6 support there is little reason to not also drop 3.7 support and bump the C++ version. My justifications are:
* Python 3.6 and 3.7 are End of Life
* The only people who would be *significantly* affected by removing 3.6 support are: people who are only motivated to use Python because of cocotb and are using the default Python environment of their End of Life'd operating system. 
The logic being:
   * Any group that uses Python for anything other than cocotb has very likely moved to newer version of Python because compatibility with Python 3.6 was already removed from one or more other Python packages they care about
   * If they are sticking with 3.6 regardless of the above fact, they also probably have no qualms about staying behind on cocotb
   * Any group that put in the effort to craft a custom Python environment for cocotb using conda, pyenv, or manually, should have the technical wherewithal to easily move to a new version of Python
   * The likelihood that the above people made a custom Python environment and haven't already updated it for 7+ years is unlikely
   * Those without the above technical prowess are using the default Python environment of their operating system, but all operating systems that use Python 3.6 or 3.7 are End of Life
      * 3.6 had extended support by way of RHEL8, but is End of Life (May 2024)
      * 3.7 had extended support by way of Debian 10 (Buster), but is End of Life (June 2024)
      * Ubuntu 18.04 used Python 3.6, but it is also End of Life (May 2023)
      * No other operating system I could find still has a supported release that ships with Python 3.6 or 3.7
* Following on the same logic, there is no operating system that isn't End of Life that doesn't come with a version of GCC or Clang new enough to support C++14

Personally, I don't see immediate value in C++14 support and would be willing to drop that without hestitation, but I don't see a need to drop it considering the above justification.

IMO the only reason to continue supporting Python 3.6 and 3.7 and C++11 as the minimum is to support those users who are willingly using unsupported operating systems.

imphil:
I'm sorry, I'm not convinced of the need to drop Python 3.6 support *for cocotb 2.0*.

Writing some type annotations as string is established practice for a long time now and not overly painful. It's a valid, but a weak reason to drop Python 3.6.

As I said before: the "cocotb maintenance burden" is the only thing we can really judge from our side, hence my focus on that. We don't know enough about our users to really quantify their pain once we remove Python 3.6 support. Having done sysadmin work for a while I know that Python is one of the more painful ecosystems to deal with. Python 3.6 *is* supported by Red Hat, and as such not EOL for users of RHEL 8, which will continue to be *the* operating system for EDA workloads for a while longer.

My conclusion: We want people to move to cocotb 2.0 to stop maintaining cocotb 1.x. There is only a weak technical reason to drop Python 3.6. Hence, just keep the support and revisit in cocotb 2.1 or whenever we find a blocker.

PR #4046, opened 2024-07-16
Renable checks in test_array following upstream fixes to Xcelium

### LINKED ISSUE #3372: Xcelium: writes to multi-dimensional arrays are partially ignored
### Description
Some writes to multi-dimensional arrays in Xcelium 23.03-s004 never make it into the design.

The following two signals are defined in a SystemVerilog design:

https://github.com/cocotb/cocotb/blob/5d463f7ecd017efa0b15448ace8641e93aef7d0d/tests/designs/array_module/array_module.sv#L82-L85

These signals are then written in a cocotb test:

https://github.com/cocotb/cocotb/blob/5d463f7ecd017efa0b15448ace8641e93aef7d0d/tests/test_cases/test_array/test_array.py#L130-L136

And later on these signals are read.

### Expected result
All elements of the array are set.

https://github.com/cocotb/cocotb/blob/5d463f7ecd017efa0b15448ace8641e93aef7d0d/tests/test_cases/test_array/test_array.py#L162-L166

There are no simulator-specific overrides in place for this test, i.e. we get the expected outcome for all simulators (except GHDL, where the full test is disabled).

### Actual result

See waveform: It seems that only the lowest element is set, all the others remain unaffected by the write.

image

### Simulator
Xcelium 23.03-s004

### How to reproduce

```
$ git clone https://github.com/cocotb/cocotb.git
$ cd cocotb
$ git checkout v1.8.0
$ python3 -m venv cocotb-venv
$ . cocotb-venv/bin/activate
$ pip3 install cocotb==1.8.0 pytest
$ COCOTB_LOG_LEVEL=TRACE TESTCASE=test_read_write make -C tests/test_cases/test_array/ SIM=xcelium
```

ktbarrett:
Closes #3372.

### DIFF
--- a/tests/test_cases/test_array/test_array.py
+++ b/tests/test_cases/test_array/test_array.py
@@ -6,6 +6,7 @@ import logging
 import os
 
 import cocotb
+from cocotb._sim_versions import XceliumVersion
 from cocotb.clock import Clock
 from cocotb.handle import (
     ArrayObject,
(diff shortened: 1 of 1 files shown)


ktbarrett:
BTW I *hate* the way this test is written. It doesn't inform us when behavior changes because it outright skips the failing parts of the test.

cmarqu (approved):

PR #4049, opened 2024-07-17
Call immediately if done_callback is added to finished Task

ktbarrett:
If a done callback is added to a Task using `add_done_callback`, and the Task is already done, the callback is never called. asyncio on the other handle schedules the callback to run in the call to `add_done_callback`. We don't have a way to schedule callbacks in this way, so we just callback immediately.

- [x] Make this API private and delete newsfragment.

### DIFF
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -242,34 +242,18 @@ class Task(Generic[ResultType]):
         else:
             return None
 
-    def add_done_callback(self, callback: Callable[["Task[ResultType]"], Any]) -> None:
+    def _add_done_callback(self, callback: Callable[["Task[ResultType]"], Any]) -> None:
         """Add *callback* to the list of callbacks to be run once the Task becomes "done".
 
         Args:
             callback: The callback to run once "done".
+
+        .. note::
+            If the task is already done, calling this function will call the callback immediately.
         """
+        if self.done():
+            callback(self)
         self._done_callbacks.append(callback)
-
-    def remove_done_callback(
-        self, callback: Callable[["Task[ResultType]"], Any]
-    ) -> int:
-        """Remove a callback added by a call to :meth:`add_done_callback`.
-
-        Args:
-            callback: The callback to remove from the callback list.
-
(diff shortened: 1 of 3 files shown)


src/cocotb/task.py
```diff
@@ -247,7 +247,12 @@ def add_done_callback(self, callback: Callable[["Task[ResultType]"], Any]) -> No
 
         Args:
             callback: The callback to run once "done".
+
+        Raises:
+            InvalidStateError: When a done callback is added to a task that is already "done".
         """
+        if self.done():
+            raise InvalidStateError("added result to task that was already done")
```
ktbarrett, line 255:
```suggestion
            callback(self)
```

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -941,3 +941,17 @@ async def passing_coro() -> None:
...
+    async def noop() -> None:
+        pass
+
+    def done_callback(_: Task) -> None:
+        pass
+
+    task = cocotb.start_soon(noop())
+    await task
+    with pytest.raises(InvalidStateError):
+        task.add_done_callback(done_callback)
```
ktbarrett, line 957:
```suggestion
    task.add_done_callback(done_callback)
    await NullTrigger()
    assert callback_ran
```

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -941,3 +941,17 @@ async def passing_coro() -> None:
...
     assert num_removed == 0
+
+
+@cocotb.test
+async def test_task_done_callback_added_after_done(_) -> None:
+    async def noop() -> None:
+        pass
+
+    def done_callback(_: Task) -> None:
+        pass
```
ktbarrett, line 952:
```suggestion
    callback_ran = False

    def done_callback(_: Task) -> None:
        nonlocal callback_ran
        callback_ran = True
```

marlonjames (approved):

PR #4055, opened 2024-07-18
Move writes flushing into `scheduler._react` method

ktbarrett:
A single step towards and proof that moving this into a trigger-specific react method would work.

### DIFF
--- a/src/cocotb/_write_scheduler.py
+++ b/src/cocotb/_write_scheduler.py
@@ -1,0 +1,69 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+from collections import OrderedDict
+from typing import Any, Callable, Dict, Sequence, Tuple, Union
+
+import cocotb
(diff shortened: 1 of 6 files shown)


src/cocotb/_write_scheduler.py
```diff
@@ -0,0 +1,69 @@
...
+        func(*args)
+    _writes_pending.clear()
+
+
+def schedule_write(
+    handle: cocotb.handle.SimHandleBase,
+    write_func: Callable[..., None],
+    args: Sequence[Any],
+) -> None:
+    """Queue `write_func` to be called on the next ReadWrite trigger."""
```
cmarqu, line 60:
```suggestion
    """Queue *write_func* to be called on the next ``ReadWrite`` trigger."""
```
If you want to just move code without touching anything else, feel free to ignore of course.

cmarqu (approved):

src/cocotb/_write_scheduler.py
ktbarrett: I call this more than code shuffling. This is setting up a new internal API, so comment away.

PR #4062, opened 2024-07-20
pyuvm ecosystem test

raysalemi:
I've added pyuvm to the cosystem-compat test.

### DIFF
--- a/.github/workflows/ecosystem-compat.yml
+++ b/.github/workflows/ecosystem-compat.yml
@@ -82,7 +82,7 @@ jobs:
           uses: actions/checkout@v4
           with:
             repository: pyuvm/pyuvm
-            path: pyuvm
+
             
         - name: Install the release version of cocotb-coverage
           run: pip3 install pyuvm==3.0.0
--- a/.github/workflows/ecosystem-compat.yml
+++ b/.github/workflows/ecosystem-compat.yml
@@ -67,6 +67,9 @@ jobs:
           with:
             python-version: "3.11"
   
+        - name: Install GHDL
+          run: sudo apt-get install -y --no-install-recommends ghdl
+
         - name: Install Icarus Verilog
           run: sudo apt-get install -y --no-install-recommends iverilog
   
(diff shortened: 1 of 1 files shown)


cmarqu:
Just a bit of whitespace to fix in one file: https://results.pre-commit.ci/run/github/10651005/1721515566.5xw2y2mMSu2tzDC3rhI-uQ

raysalemi:
> Just a bit of whitespace to fix in one file: https://results.pre-commit.ci/run/github/10651005/1721515566.5xw2y2mMSu2tzDC3rhI-uQ

I blame CoPilot!  

It wrote that line from my comment.

.github/workflows/ecosystem-compat.yml
```diff
@@ -57,3 +57,54 @@ jobs:
...
+        - name: Install the development version of cocotb
+          run: pip3 install ./cocotb
+
+        - name: Checkout pyuvm repository
+          uses: actions/checkout@v4
+          with:
+            repository: pyuvm/pyuvm
+
+        - name: Install the release version of cocotb-coverage
+          run: pip3 install pyuvm==3.0.0
```
ktbarrett, line 90: Do we want to only test against 3.0.0? Maybe using `pyuvm~=3.0` makes sense so if patch or minor releases occur we automatically use them, but not 4.0. Or maybe it should be `pyuvm>=3`?

raysalemi: I did not know ~= existed.

Though, I think >=3 makes sense.

I'll make the change (which will restart the tests.)

raysalemi: Changed to >=3 and repushed.

ktbarrett (approved):

ktbarrett:
@raysalemi I see no issue on my end. We usually wait for two maintainers to approve/reject, but if there is no activity for a couple days then I'll just merge.

marlonjames (approved):

PR #4069, opened 2024-07-25
Fix multiple VHDL libraries in Pytest + NVC

m42uko:
Hi,

when running the `tests/pytest/test_vhdl_libraries_multiple.py` test case with NVC, it fails with the following error message: `test_vhdl_libraries_multiple.py ** Error: library ELIB not found`

This is because NVC does not search for libraries in `build_dir`.

This PR

* Activates this test case for NVC
* Fixes the issue by adding `-L <build_dir>` to the NVC arguments.

(I can add a bugfix newsfragment after review if requested.)

Thank you!
Markus

@nickg: As usual, I'd like to hear your opinion :) Thank you!

### DIFF
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -1035,7 +1035,7 @@ class Nvc(Runner):
                 )
 
         cmds = [
-            ["nvc", f"--work={self.hdl_library}"]
+            ["nvc", f"--work={self.hdl_library}", "-L", str(self.build_dir)]
             + [arg for arg in self.build_args if type(arg) in (str, VHDL)]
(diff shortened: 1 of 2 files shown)


ktbarrett:
Don't worry about the windows build failure. That thing hangs and times out all the time.

ktbarrett:
I'm guessing this only becomes relevant if build_dir is set, which it usually isn't for us, which is why we haven't run into it yet?

ktbarrett:
> Python runner with NVC and multiple libraries.

Ah. Well good.

ktbarrett (approved):

marlonjames (approved):

PR #4071, opened 2024-07-26
Remove concept of test from scheduler

ktbarrett:
* Moves `_abort_test` to regression manager. Further refactoring may make this function simpler.
* Moves logic that puts scheduler into terminating state when the test task finishes into a done callback on the test task.
* Moves logic that decides to fail the test if a task gets an exception into a done callback on all `cocotb.start_soon` tasks.
* `scheduler._add_test` is inlined into the regression manager
* Removes `scheduler._test` !!!

Further refactorings may find a way to remove the `test_complete_cb` function, but it's uncertain if that's possible rn.

### DIFF
--- a/src/cocotb/__init__.py
+++ b/src/cocotb/__init__.py
@@ -26,6 +26,7 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import ast
+import inspect
 import logging as py_logging
 import os
 import random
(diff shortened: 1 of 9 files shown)


marlonjames:
Rebase after #4087 is in.

src/cocotb/__init__.py
```diff
@@ -196,7 +219,24 @@ def create_task(
...
     .. versionadded:: 1.6.0
     """
-    return cocotb._scheduler_inst.create_task(coro)
+    if isinstance(coro, cocotb.task.Task):
+        return coro
+    elif isinstance(coro, Coroutine):
+        return cocotb.task.Task(coro)
+    elif inspect.iscoroutinefunction(coro):
+        raise TypeError(
+            f"Coroutine function {coro} should be called prior to being " "scheduled."
```
marlonjames, line 228:
```suggestion
            f"Coroutine function {coro} should be called prior to being scheduled."
```

marlonjames (commented):
We should update #2664. The `FutureWarning` has been in since 1.6, but will no longer apply.

marlonjames (approved):

Issue #3583, opened 2023-12-13
Separating symmetric and asymmetric tasks


ktbarrett:
## Background

Follow on to #2664.

As stated the original issue is that if a Task without a registered Join trigger throws an exception, the test is ended with the exception as the test exception. But if the Task has a Join trigger and an exception is thrown, that exception is propagated to the Task waiting on the Join. This is inconsistent and non-deterministic. One can intend to wait on the failing Task, but the Task can fail before the wait occurs, making it impossible to catch.

An example of the above gotcha:
```python
@cocotb.test()
async def test_fails(dut):
    async def fails_after_random_time():
        await Timer(randint(1, 100), 'ns')
        raise ValueError("pretend this was expected for valid reasons")
    task = cocotb.start_soon(fails_after_random_time())
    await Timer(50, 'ns')
    try:
        await task
    except ValueError:
        pass  # pretend we handle this
```

In the example, the `fails_after_random_time` task might throw before or after the task is await. If it throws before, the test fails. If it throws after, the exception is caught and handled. Of course, this example is contrived, but shows the inconsistent behavior.

## Currently slated solution

The current solution, from #2664, is that tasks created with `cocotb.start_soon` will *not* fail the test if they get an exception and will always store the exception for propagation. This follows how asyncio is implemented.

The issue with this approach is that there is no way currently to get the existing behavior of ending the test if the task throws an exception, which is commonly needed for verification components, such as Drivers, Monitors, Scoreboards, etc.

## Possible solution

One solution is to have two types of tasks: symmetric and asymmetric (terminology from Revisiting Coroutines, De Moura and Ierusalimschy). Symmetric tasks run at the same hierarchical level as the test and other tasks are not intended to directly wait on them. Asymmetric tasks are children of other tasks and are intended to be waited upon and have return info and exceptions propagated. In this setup, symmetric tasks getting an exception would fail the test, and asymmetric tasks getting an exception would capture the exception for propagation.

### Introducing `TaskGroup`s as asymmetric control, leave `cocotb.start_soon` as symmetric

This would do the opposite of what the `FutureWarning` in #2664 introduces: exceptions would *always* fail the test. There would need to be another interface for creating asymmetric tasks. Since we are considering adding `TaskGroup`s, which are similar to SystemVerilog's fork/join blocks, that would provide an excellent way of enforcing asymmetric Tasks.

For an example:
```python
@cocotb.test
async def test_example(dut):
    # if Clock's start task throws an exception, the test should fail
[...]

PR #4072, opened 2024-07-26
Change exception in task behavior

ktbarrett:
Depends on #4071.

Add parameter to control exception in Task behavior.
    
`fail_on_exception` can be:
* NEVER: exceptions will be stored and never cause test failure.
* ALWAYS: exceptions will always cause test failure.
* LEGACY (default): the old behavior where it only failed the test if
  there was no one awaiting the failing task.

### DIFF
--- a/docs/source/newsfragments/4072.feature.rst
+++ b/docs/source/newsfragments/4072.feature.rst
@@ -1,0 +1,1 @@
+Added *fail_on_exception* parameter to :func:`cocotb.start_soon`, :func:`cocotb.start`, and :func:`cocotb.create_task` to control the behavior when a :class:`~cocotb.task.Task` finishes with an Exception.
--- a/src/cocotb/__init__.py
+++ b/src/cocotb/__init__.py
@@ -26,6 +26,7 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import ast
+import inspect
 import logging as py_logging
 import os
 import random
@@ -42,9 +43,10 @@ import cocotb.handle
 import cocotb.task
 import cocotb.triggers
 from cocotb._scheduler import Scheduler
-from cocotb._utils import DocEnum
+from cocotb._utils import DocEnum, remove_traceback_frames
 from cocotb.logging import default_config
 from cocotb.regression import RegressionManager, RegressionMode
+from cocotb.result import TestSuccess
 
 from ._version import __version__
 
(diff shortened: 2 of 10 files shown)


marlonjames:
This should close #922 as well, should add that or similar test to this PR.
This is basically the inverse of `capture_exception` I suggested in https://github.com/cocotb/cocotb/issues/3583#issuecomment-1858618800

This has slightly different behavior though, because as I understand it, if nothing ever awaits the non-propagating task, the test won't fail.

ktbarrett:
> if nothing ever awaits the non-propagating task, the test won't fail.

Other way around: if nothing awaits a propagating task, the test won't fail.

But this shouldn't be a problem once we add structured use of this feature. The `propagate` switch mostly exists for an easy upgrade path for legacy testbenches that need some tasks to be propagating. The idea being that
1. Tests that current depend upon exception propogation will suddenly fail loudly, so which tests to upgrade will be obvious.
2. The existing test that use propagate are already probably doing the right thing and not having propagating tasks never be awaited, so slapping on `propagate=True` will bring it back to working exactly as it did before.

We could also raise `RuntimeWarning` on unawaited propagating tasks on collection like asyncio does.

New testbenches will use something akin to trio's nurseries or asyncio's TaskGroup. All associated task will be propagating and form tasks into a tree. These simply do not allow users to not await on propagating tasks, so they are easy to use right and hard to use wrong.

marlonjames:
> We could also raise `RuntimeWarning` on unawaited propagating tasks on collection like asyncio does.

This is the part I meant. Without this, the test would pass if nothing ever awaits the failed propagating task.

> `capture_exception` == `propagate`.

Yeah you're right.

As far as the name, it's describing two linked behaviors. Propagate the Exception into Tasks that are awaiting the failed Task, and also don't fail test / cancel TaskGroup. I think my confusion came from the thought of "propagating the exception up to the test/TaskGroup" vs keeping it confined to the Task, whose result can be gotten through `await task` or `task.result()`.

ktbarrett:
I'm actually considering changing this so `cocotb.start_soon` tasks behave as they did. The real problem with the old behavior is the case where you *intended* to wait on a task, but it failed before you could. We can leave the old behavior in place and just add a "dont fail the test" switch for tasks to always propagate exceptions. That should maximize backwards compatibility.

ktbarrett:
I'm going to come back to this after I get the PR which splits a `Test` object from the Scheduler/RegressionManager.

PR #4079, opened 2024-07-29
Remove the `data` field from `Event`

### LINKED ISSUE #3980: Consider removing data and name attributes of `Event`
One can set a data attribute when `set()`ing and can give the Event a name. These features aren't used anywhere in the current tests and is something the user can do with a companion variable fairly easily. Nor is this something that `asyncio` supports.

ktbarrett:
Also removes the corresponding `data` argument to `Event.set`.

Closes #3980.

### DIFF
--- a/src/cocotb/triggers.py
+++ b/src/cocotb/triggers.py
@@ -433,7 +433,7 @@ class _Event(Trigger):
     can maintain a unique mapping of triggers to tasks.
     """
 
-    def __init__(self, parent: "Event[Any]") -> None:
+    def __init__(self, parent: "Event") -> None:
         super().__init__()
         self._parent = parent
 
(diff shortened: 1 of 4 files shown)


src/cocotb/triggers.py
```diff
@@ -471,23 +471,25 @@ async def task1():
             e.set()
             await NullTrigger()  # allows task1 to execute
             # resuming!
+
+    .. versionremoved:: 2.0
+
+        Removed the undocumented *data* attribute and argument to :meth:set.
```
ktbarrett, line 477:
```suggestion
        Removed the undocumented *data* attribute and argument to :meth:`set`.
```

src/cocotb/triggers.py
```diff
@@ -471,23 +471,25 @@ async def task1():
...
 
     def _prime_trigger(
         self, trigger: _Event, callback: Callable[[Trigger], None]
     ) -> None:
         self._pending_events.append(trigger)
 
-    def set(self, data: Optional[T] = None) -> None:
-        """Unblock all coroutines blocked on this event."""
+    def set(self) -> None:
+        """Sets the Event and unblock all Tasks blocked on this Event."""
```
marlonjames, line 491:
```suggestion
        """Set the Event and unblock all Tasks blocked on this Event."""
```

marlonjames:
Partial of #3980

marlonjames (approved):

PR #4080, opened 2024-07-30
Don't force `vpiNoDelay` when writing strings

ktbarrett:
Removes special case logic for assigning strings in VPI. Previously if assigning a string with `GPI_DEPOSIT`, it was assigned with `vpiNoDelay`. The comment says

> // assigning to a vpiStringVar only seems to work with
                // vpiNoDelay

But doesn't say what simulator or version, or what error they were seeing. This is now causing a behavioral issue, so I'm investigating removing it.


cmarqu (approved):

ktbarrett:
Opened #4081 and #4082.

PR #4085, opened 2024-08-01
Guard VHDL-specific option in Xcelium Python runner.

### LINKED ISSUE #4059: #3613 breaks using xcelium older than 23.09
With cocotb 1.9.0, due to #3613, if using xcelium that's older than 23.09, xcelium errors out with 
Unkown option NEW_VHPI_PROPAGATE_DELAY

The option should only be used for xcelium versions that support it...

ktbarrett:
Closes #4059. Guards addition of `-NEW_VHPI_PROPAGATE_DELAY` on whether VHDL sources are present.

### DIFF
--- a/tests/pytest/test_parallel_cocotb.py
+++ b/tests/pytest/test_parallel_cocotb.py
@@ -48,6 +48,8 @@ def test_cocotb_parallel(seed):
 
     runner.build_args = compile_args
     runner.sources = sources
+    runner.verilog_sources = []
+    runner.vhdl_sources = []
 
     runner.test(
         seed=seed,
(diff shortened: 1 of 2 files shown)


marlonjames (approved):

ktbarrett:
Welp... test_parallel_cocotb isn't working because it's trying to access build-only attributes (`sources`/`vhdl_sources`) in the test command. Not having the runner object shared between the compile step and the test step is kinda suck. But then I guess you'd have to worry about the parallel runs stepping on the others shoes (attributes).

tests/pytest/test_parallel_cocotb.py
```diff
@@ -47,6 +47,7 @@ def test_cocotb_parallel(seed):
     runner = get_runner(sim)
 
     runner.build_args = compile_args
+    runner.sources = sources
```
ktbarrett, line 50:
```suggestion
    runner.sources = sources
    runner.verilog_sources = []
    runner.vhdl_sources = []
```

cmarqu (approved):

Issue #869, opened 2019-03-17
Mechanism for waiting for the first of multiple triggers doesn't work for Joins


eric-wieser:
Consider
```python
@cocotb.coroutinue
def task_one():
    yield Timer(10)

@cocotb.coroutinue
def task_two():
   yield Timer(2)
```

It's not possible to work out which one finishes first:
```python
@cocotb.coroutinue
def test_with_fork():
   co_one = cocotb.fork(task_one)
   co_two = cocotb.fork(task_two)
   co_which = yield [co_one, co_two]
   assert co_which in {co_one, co_two}  # fails, co_which  is None
```

`asyncio.wait` solves this by returning `Future` objects, rather than the result of the futures.

garmin-mjames:
We could add a `Future` trigger that has access to both the `Outcome` and `RunningCoroutine` objects.

Alternatively we could add a trigger that returns just the `RunningCoroutine` object, an alternate to `Join` like `Ident` or similar, so the test would explicitly yield on a trigger expecting the coroutine back.

```
co_which = yield [Ident(co_one), Ident(co_two)]
```

ktbarrett:
I think  we are going to leave `First` and `Combine` the same and introduce something newer that works better. xref #2659. xref #3583.

ktbarrett:
You can use the following as of #3931.

```python3
trigger = await First(task1.join(), task2.join())
first_task = trigger.task
# work out which one
if trigger is task1.join():
    ...
```

marlonjames:
As of #4341, use this pattern:

```py3
trigger = await First(task1.complete, task2.complete)
first_task = trigger.task
# work out which one
if first_task is task1:
    ...
```

PR #4087, opened 2024-08-06
Deprecate `Join()` and `task.join()`

ktbarrett:
This reintroduces old behavior of `Join(task)` and `task.join()`. The change was causing tests which previously failed to continue because when using `Join(task)` or `task.join()` instead of just `task` it prevented exception propagation. It also deprecates the `Join` trigger and `task.join()`.

In asyncio, which we are trying to align with, Tasks are Futures (known in cocotb as Triggers). There is no separate Future object for each Trigger firing. So Tasks are passed directly where any Future is expected. Similarly, cocotb currently supports passing Tasks directly wherever Triggers are expected. Removing Join from the public API forces users to pass Tasks directly instead of Join triggers which should align their code with the asyncio-like model we are converging on.


marlonjames (approved):
Reopens #869 
We'll have to figure out another mechanism for that.

PR #4090, opened 2024-08-12
C++ fixes

### LINKED ISSUE #2208: Allow more than one C-level callback for a signal edge

ktbarrett:
* cleanup
* don't reuse value change callback objects, allowing multiple ones to be created (closes #2208)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,3 +112,6 @@ examples/*/tests/verilog.log
 # Build artifacts
 /dist/
 src/cocotb/_version.py
+
+# clangd
+compile_commands.json
--- a/src/cocotb/share/include/cocotb_utils.h
+++ b/src/cocotb/share/include/cocotb_utils.h
@@ -90,4 +90,11 @@ constexpr Deferable<F> make_deferable(F f) {
 #define DEFER(statement) \
     auto DEFER0(_defer, __COUNTER__) = make_deferable([&]() { statement; });
 
+#if __cplusplus >= 202302L
+#include <utility>
+using unreachable = std::unreachable;
+#else
+extern "C" [[noreturn]] void unreachable();
+#endif
+
 #endif /* COCOTB_UTILS_H_ */
--- a/src/cocotb/share/include/gpi.h
+++ b/src/cocotb/share/include/gpi.h
@@ -232,8 +232,9 @@ GPI_EXPORT void gpi_set_signal_value_str(
     gpi_set_action_t action);  // String of ASCII char(s)
 
 typedef enum gpi_edge {
-    GPI_RISING = 1,
-    GPI_FALLING = 2,
+    GPI_RISING,
+    GPI_FALLING,
+    GPI_VALUE_CHANGE,
 } gpi_edge_e;
 
 // The callback registering functions
@@ -242,7 +243,7 @@ GPI_EXPORT gpi_cb_hdl gpi_register_timed_callback(int (*gpi_function)(void *),
                                                   uint64_t time);
 GPI_EXPORT gpi_cb_hdl gpi_register_value_change_callback(
     int (*gpi_function)(void *), void *gpi_cb_data, gpi_sim_hdl gpi_hdl,
-    int edge);
+    gpi_edge_e edge);
 GPI_EXPORT gpi_cb_hdl
 gpi_register_readonly_callback(int (*gpi_function)(void *), void *gpi_cb_data);
 GPI_EXPORT gpi_cb_hdl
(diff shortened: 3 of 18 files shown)


ktbarrett:
Ugh... it's always a battle.

ktbarrett:
So it turned out to be the fix for the memory leak that broke it.

<details>

```diff
diff --git a/src/cocotb/share/lib/gpi/GpiCommon.cpp b/src/cocotb/share/lib/gpi/GpiCommon.cpp
index 63c0da64..7be4792b 100644
--- a/src/cocotb/share/lib/gpi/GpiCommon.cpp
+++ b/src/cocotb/share/lib/gpi/GpiCommon.cpp
@@ -610,6 +610,10 @@ void gpi_deregister_callback(gpi_cb_hdl cb_hdl) {
     cb_hdl->m_impl->deregister_callback(cb_hdl);
 }
 
+void *gpi_get_callback_data(gpi_cb_hdl cb_hdl) {
+    return cb_hdl->get_user_data();
+}
+
 const char *GpiImplInterface::get_name_c() { return m_name.c_str(); }
 
 const string &GpiImplInterface::get_name_s() { return m_name; }
diff --git a/src/cocotb/share/lib/gpi/gpi_priv.h b/src/cocotb/share/lib/gpi/gpi_priv.h
index 2e4d24c7..119871a9 100644
--- a/src/cocotb/share/lib/gpi/gpi_priv.h
+++ b/src/cocotb/share/lib/gpi/gpi_priv.h
@@ -186,6 +186,7 @@ class GPI_EXPORT GpiCbHdl : public GpiHdl {
     gpi_cb_state_e get_call_state();
 
     int set_user_data(int (*function)(void *), void *cb_data);
+    void *get_user_data() { return m_cb_data; };
 
     virtual ~GpiCbHdl();
 
diff --git a/src/cocotb/share/lib/simulator/simulatormodule.cpp b/src/cocotb/share/lib/simulator/simulatormodule.cpp
index 12e2b66d..3ab4cdb8 100644
--- a/src/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/src/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -764,6 +764,11 @@ static PyObject *stop_simulator(PyObject *, PyObject *) {
 }
 
 static PyObject *deregister(gpi_hdl_Object<gpi_cb_hdl> *self, PyObject *) {
+    // cleanup uncalled callback
+    auto cb = static_cast<PythonCallback *>(gpi_get_callback_data(self->hdl));
+    delete cb;
+
+    // deregister from interface
     gpi_deregister_callback(self->hdl);
 
     Py_RETURN_NONE;
```
</details>

While that does prevent a memory leak of the callback data when a callback is deregistered before it fires, it also causes a double free on that callback data if it *did* fire since we delete the callback data in both `handle_gpi_callback` after the callback is run, and in `deregister_callback` when we call it via `Trigger._unprime` called in `Scheduler.react`. This is *theoretically* necessary since the Trigger object may need to do cleanup / update state, and there is no path *through* it, so the scheduler must use `Trigger._unprime` to signal it. But this function serves a dual purpose of also being the method use to prematurely cancel the Trigger.

I'll work on this in a follow on. What's here is good enough to come in and unblocks further feature work.

imphil (approved):
LGTM from a code point of view.

On "Don't attempt to reuse value change callbacks": do you have an idea why callbacks were reused before? One could imagine performance being the reason. Were you able to check the performance before and after this change?

ktbarrett:
I traced it back to https://github.com/cocotb/cocotb/commit/ef963f33c976514701abfca060dfc4c475b3c4a9 where it was a single value change callback called `value_cb`. No indication why it was done that way originally.

olagrottvik:
@ktbarrett I have traced a massive performance hit to commit f930bc77f239f4556371a1d5ea4cad8bb2349565. 

In my example the simulation runtime goes from 2.72 s to 48.92 s.

Are you already aware of this? If not, should I create an issue and try to provide a minimal example?

ktbarrett:
@ktbarrett Which simulator? There was an issue that reported a slowdown issue with Questa.

olagrottvik:
Riviera-Pro

I made an issue here: https://github.com/cocotb/cocotb/issues/4379

PR #4096, opened 2024-08-13
Interact with `libpython` through the `dlopen` handle

ipburbank:
Test per thread [HERE](https://matrix.to/#/!eqvfxHYGQmlSXRFWJt:gitter.im/$czDTYfHp1Ivp5xxHcbuigQbNQEncNPItqb7ZnQPwcZ0?via=gitter.im)


ktbarrett:
There are also plenty of Python API calls in `simulatormodule.cpp`. 

This solution basically prevents us from ever moving to something like nanobind or pybind11, which we had hoped to do at some point. I think this whole problem of loading embedding different version of Python into an executable simultaneously is something we should maybe take to the Python core devs, though I doubt they'll have a solution.

PR #4101, opened 2024-08-15
More LogicArray features

### LINKED ISSUE #4100: Add "trusted" constructor to LogicArray for performance
LogicArray includes some checks to ensure that the input values are the correct type and value. When constructing from binstr from the simulator, these checks can be skipped.

### LINKED ISSUE #4106: Ensure GPI returns binstr values always in uppercase
This is currently being handled in Python, which can be slow. Doing it in C++ might improve performance and allows us to make guarantees about values coming from the maybe-one-day public GPI.

ktbarrett:
* adds methods for converting to and from bytes
* ~~adds shorthand for specifying Ranges by just giving the length (backwards compatibility with BinaryValue)~~
* adds trusted constructor for LogicArray for use with handles (closes #4100, closes #4106)
* further documentation, testing, and coverage cleanup

### DIFF
--- a/src/cocotb/handle.py
+++ b/src/cocotb/handle.py
@@ -899,7 +899,7 @@ class LogicObject(
             [ValueObjectBase[Any, Any], Callable[..., None], Sequence[Any]], None
         ],
     ) -> None:
-        value_: LogicArray
+        value_: str
         if isinstance(value, int):
             min_val, max_val = _value_limits(len(self), _Limits.VECTOR_NBIT)
             if min_val <= value <= max_val:
@@ -909,15 +909,20 @@ class LogicObject(
                     )
                     return
 
+                # LogicArray used for checking
                 if value < 0:
-                    value_ = LogicArray.from_signed(
-                        value,
-                        Range(len(self) - 1, "downto", 0),
+                    value_ = str(
+                        LogicArray.from_signed(
+                            value,
+                            Range(len(self) - 1, "downto", 0),
+                        )
                     )
                 else:
-                    value_ = LogicArray.from_unsigned(
-                        value,
-                        Range(len(self) - 1, "downto", 0),
+                    value_ = str(
+                        LogicArray.from_unsigned(
+                            value,
+                            Range(len(self) - 1, "downto", 0),
+                        )
                     )
             else:
                 raise OverflowError(
@@ -925,28 +930,29 @@ class LogicObject(
                 )
 
         elif isinstance(value, str):
-            value_ = LogicArray(value, self.range)
+            # LogicArray used for checking
+            value_ = str(LogicArray(value, self.range))
 
         elif isinstance(value, LogicArray):
             if len(self) != len(value):
                 raise ValueError(
                     f"cannot assign value of length {len(value)} to handle of length {len(self)}"
                 )
-            value_ = value
+            value_ = str(value)
 
         elif isinstance(value, Logic):
             if len(self) != 1:
                 raise ValueError(
                     f"cannot assign value of length 1 to handle of length {len(self)}"
                 )
-            value_ = LogicArray([value])
+            value_ = str(value)
 
         else:
             raise TypeError(
                 f"Unsupported type for value assignment: {type(value)} ({value!r})"
             )
 
-        schedule_write(self, self._handle.set_signal_val_binstr, (action, str(value_)))
+        schedule_write(self, self._handle.set_signal_val_binstr, (action, value_))
 
     @property
     def value(self) -> LogicArray:
@@ -979,7 +985,7 @@ class LogicObject(
             ``sum(v << (d['bits'] * i) for i, v in enumerate(d['values']))`` instead.
         """
         binstr = self._handle.get_signal_val_binstr()
-        return LogicArray(binstr)
+        return LogicArray._from_handle(binstr)
 
     @value.setter
     def value(self, value: LogicArray) -> None:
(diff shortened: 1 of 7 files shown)


src/cocotb/handle.py
```diff
@@ -909,44 +909,50 @@ def _set_value(
...
                     )
             else:
                 raise OverflowError(
                     f"Int value ({value!r}) out of range for assignment of {len(self)!r}-bit signal ({self._name!r})"
                 )
 
         elif isinstance(value, str):
-            value_ = LogicArray(value, self.range)
+            # LogicArray used for checking
+            value_ = str(LogicArray(value, self.range))
```
ktbarrett, line 934: Interesting we aren't testing this. I'll add a small test to test_handles for this.

docs/source/newsfragments/4098.feature.1.rst
```diff
@@ -0,0 +1 @@
+:class:`~cocotb.types.LogicArray` can take :class:`int` for the ``range`` argument as shorthand for ``Range(range-1, "downto", 0)``.
```
imphil, line 1: Do we really need to overload the `range` argument here? `range=4` reads strange, more common naming would be `size=4` or `width=4`. How about keeping range as-is, and adding a new named argument (e.g., `size` or `width`) that takes an integer?

src/cocotb/types/logic_array.py
```diff
@@ -333,6 +333,15 @@ def from_bytes(
             int.from_bytes(value, byteorder=byteorder, signed=False), range=range
         )
 
+    @classmethod
+    def _from_handle(cls, value: str) -> "LogicArray":
+        self = super().__new__(cls)
+        self._value_as_array = None
+        self._value_as_int = None
+        self._value_as_str = value.upper()
```
imphil, line 341: Maybe add a TODO here to guarantee upper-case values already in GPI? Seems like this could give us a bit of performance.

src/cocotb/types/logic_array.py
```diff
@@ -273,6 +287,37 @@ def from_signed(cls, value: int, range: Range) -> "LogicArray":
             )
         return LogicArray(value, range=range)
 
+    @classmethod
+    def from_bytes(
+        cls,
+        value: Union[bytes, bytearray],
+        range: Optional[Range] = None,
+        byteorder: str = "big",
```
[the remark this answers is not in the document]
ktbarrett: Running mypy on this says this should be `Literal["big"] | Literal["little"]`. I should be able to use a `if TYPE_CHECKING` block to do the import since this is a Python 3.8+ feature.

docs/source/newsfragments/4098.feature.1.rst
ktbarrett: I thought about this. It might make the constructor slightly more complicated, making things slower, but I guess since we are using the trusted constructor for handles this is fine since only manually constructed LogicArrays would hit it.

I'll do that.

src/cocotb/types/logic_array.py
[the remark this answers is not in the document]
imphil: Yes, that's another option. Not really of highest importance, but nice to have.

src/cocotb/types/logic_array.py
```diff
@@ -3,30 +3,50 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import warnings
 from math import ceil
-from typing import Iterable, Iterator, List, Optional, Union, cast, overload
+<<<<<<< HEAD
```
cmarqu, line 6: Merge conflict.

tests/pytest/test_logic_array.py
```diff
@@ -6,46 +6,118 @@
...
+
+def test_logic_array_bytes_conversion():
+    assert LogicArray.from_bytes(b"12") == LogicArray("0011000100110010")
+
+    with pytest.raises(OverflowError):
+        LogicArray.from_bytes(b"123", Range(6, "downto", 0))
+    with pytest.raises(OverflowError):
+        LogicArray.from_bytes(b"123", 7)
+
+    # b"1" would fit in a 7 bit LogicArray, but we do not guessing about whether top bits are significant or not
```
cmarqu, line 116:
```suggestion
    # b"1" would fit in a 7 bit LogicArray, but we do no guessing about whether top bits are significant or not
```

imphil: "do not guess"? :-)

src/cocotb/types/logic_array.py
ktbarrett: Apparently the BinaryValue version of this method didn't uppercase the string, but I'm not sure if that's because it worked with mixed case or if it assumed that the values were already uppercase.

docs/source/newsfragments/4098.feature.1.rst
ktbarrett: This change actually shuffles a lot of code. I need to come up with a better scheme than I have now.

ktbarrett: I punted on that change for another PR.

src/cocotb/types/logic_array.py
```diff
@@ -336,6 +336,15 @@ def from_bytes(
             int.from_bytes(value, byteorder=byteorder, signed=False), range=range
         )
 
+    @classmethod
+    def _from_handle(cls, value: str) -> "LogicArray":
```
imphil, line 340: We cannot assert for it, but please mention in a comment that `value` is expected to be all-uppercase.

imphil (approved):

PR #4112, opened 2024-08-19
CI: Gate log dumping in backport action

imphil:
The log dumping tasks weren't gated by the label check, leading to cases
where no logs were available and the `cat` failed.

Recent example: https://github.com/cocotb/cocotb/actions/runs/10457916201/job/28958603739?pr=4107


imphil:
Looks like it could work now, skips don't see any errors: https://github.com/cocotb/cocotb/actions/runs/10460539823/job/28966921620?pr=4112

ktbarrett (approved):

PR #4114, opened 2024-08-19
Test writes in `ReadWrite`

ktbarrett:
We noticed an issue with our fork of the Verilator main loop and this reproducer seemed valuable to stick in the regression. It might catch a couple misbehaving sims as well...

PR #4115, opened 2024-08-19
Apply writes in ReadWrite phase immediately

### LINKED ISSUE #3966: Scheduling writes in `ReadWrite` may not be valid
If we schedule writes in the ReadWrite phase it waits for the next ReadWrite phase to apply them. But if no writes were applied in the current ReadWrite phase, there won't be more event, and thus there won't be another ReadWrite phase in the time step, so it would skip to the next time step or do whatever, as it's UB.

If we are in ReadWrite phase, scheduled writes should just be applied immediately, instead of scheduled for the next ReadWrite phase which may or may not exist.

First, we need a test to what the actual behavior in this case is.

ktbarrett:
Depends on #4114 and #4118. Closes #3966.

This change may increase performance by decreasing the number of cycles needed to complete a time step.
Previously, any write during the ReadWrite phase was queued up for the *next* ReadWrite phase. Waiting for the next one to apply more writes will only end up taking two more eval cycles which could be done in one.
All invariants should still hold even when collapsing what was previously two cycles into one:
* all writes during ReadWrite should read-back in the next ReadWrite or ReadOnly stage
* all writes are still inertial so you won't be able to read-back what you just wrote
* writes applied immediately in the ReadWrite stage follow application of scheduled writes applied in the ReadWrite stage (the writes maintain order)
* any HDL changes downstream of a value changed by either scheduled writes or immediate writes won't be visible until the next ReadWrite or ReadOnly

Things that will change:
* writes to the same signal in successive ReadWrites will be no longer be *seen* in HDL as a glitch as the immediate write will overwrite the applied scheduled write in 0 "time" / delta cycles.

Additionally, while all the VPI implementations (and Questa's FLI) allow the user to schedule infinitely more ReadWrite stages in a time step, every VHPI implementation does *not* allow you to do this, so writes scheduled in a ReadWrite phase where no writes were applied will invoke UB (typically apply writes on the next time step, whoops...).

Finally, in COCOTB_TRUST_INERTIAL_WRITES=1 mode, inertial writes done in ReadWrite in GHDL and NVC already behave this way and it seems to cause no issues.

- [X] newsfragment

### DIFF
--- a/docs/source/newsfragments/4115.change.rst
+++ b/docs/source/newsfragments/4115.change.rst
@@ -1,0 +1,1 @@
+Writes performed following an ``await`` on :class:`~cocotb.triggers.ReadWrite` will be applied immediately (but inertially) and not scheduled for the next ``ReadWrite``.
--- a/src/cocotb/_write_scheduler.py
+++ b/src/cocotb/_write_scheduler.py
@@ -71,12 +71,14 @@ else:
         args: Sequence[Any],
     ) -> None:
         """Queue *write_func* to be called on the next ``ReadWrite`` trigger."""
-        if cocotb.sim_phase == cocotb.SimPhase.READ_ONLY:
+        if cocotb.sim_phase == cocotb.SimPhase.READ_WRITE:
+            write_func(*args)
+        elif cocotb.sim_phase == cocotb.SimPhase.READ_ONLY:
             raise Exception(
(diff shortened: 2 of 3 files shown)


imphil (approved):
LGTM, thanks! 

If you have additional comments, @marlonjames, we can always do a follow-up.

PR #4118, opened 2024-08-20
Fix sim phase tracking bug in scheduler

ktbarrett:
Should have been an `elif` on the second condition which cause all ReadWrite phase to be overwritten with Normal.


marlonjames (approved):

marlonjames:
Is there a test that would have discovered this?

ktbarrett:
@marlonjames Added a test.

PR #4125, opened 2024-08-22
Emit TypeError if invalid type used when constructing Logic

ktbarrett:
This is more in line with idiomatic Python where `TypeError` is used if an argument is of the wrong type and `ValueError` is used if the type is correct, but the value is invalid for some reason.

### DIFF
--- a/src/cocotb/types/logic.py
+++ b/src/cocotb/types/logic.py
@@ -106,7 +106,8 @@ class Logic:
         value: value to construct into a :class:`Logic`.
 
     Raises:
-        ValueError: if the value cannot be constructed into a :class:`Logic`.
+        ValueError: If the value if of the correct type, but cannot be constructed into a :class:`Logic`.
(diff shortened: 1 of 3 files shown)


cmarqu (approved):

cmarqu (changes requested):
The docstring for the class should mention this in the `Raises:` section.

src/cocotb/types/logic.py
```diff
@@ -106,7 +106,8 @@ class Logic:
         value: value to construct into a :class:`Logic`.
 
     Raises:
-        ValueError: if the value cannot be constructed into a :class:`Logic`.
+        ValueError: If the value cannot be constructed into a :class:`Logic`.
+        TypeError: If the value is of a type that can be constructed into a :class:`Logic`, but the value is invalid.
```
cmarqu, line 110: The other way 'round, right?

ktbarrett: Right...

cmarqu (approved):

PR #4136, opened 2024-08-26
Deregister Event Triggers when a Task awaiting one is killed

### LINKED ISSUE #4135: `_Event._unprime` does not deregister the object with `Event`
All `_Event` triggers always fire, even if the Task that waits on them is killed. Either `_Event._unprime` should deregister the `_Event` with `Event`, or `Event.set` needs to check `_Event._primed` to ensure only primed callbacks are being called. Ideally the former.

ktbarrett:
Depends on #4128. Closes #4135. This should prevent Events from going through `_react` if they were unprimed.

This also removes the `_pending_unprimed` list from `Lock` as it appears to be doing nothing.

### DIFF
--- a/src/cocotb/triggers.py
+++ b/src/cocotb/triggers.py
@@ -445,9 +445,17 @@ class _Event(Trigger):
         self._parent = parent
 
     def _prime(self, callback: Callable[[Trigger], None]) -> None:
+        if self._primed:
+            return
         self._callback = callback
         self._parent._prime_trigger(self, callback)
-        super()._prime(callback)
+        return super()._prime(callback)
+
+    def _unprime(self) -> None:
+        if not self._primed:
+            return
+        self._parent._unprime_trigger(self)
+        return super()._unprime()
 
     def __repr__(self) -> str:
         return f"<{self._parent!r}.wait() at {_pointer_str(self)}>"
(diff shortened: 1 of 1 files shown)


src/cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -113,32 +114,22 @@ int main(int argc, char** argv) {
...
-        bool again = true;
-        while (again) {
-            // Evaluate design
-            top->eval_step();
-
-            // Call Value Change callbacks triggered by eval()
-            // These can modify signal values
-            again = settle_value_callbacks();
+        do {
+            // We must evaluate whole design until we process all 'events' for
```
cmarqu, line 118:
```suggestion
            // We must evaluate the whole design until we process all 'events' for
```

ktbarrett: Sorry, the rebase was weird and included already merged commits.

src/cocotb/triggers.py
```diff
@@ -445,9 +445,15 @@ def __init__(self, parent: "Event") -> None:
         self._parent = parent
 
     def _prime(self, callback: Callable[[Trigger], None]) -> None:
-        self._callback = callback
-        self._parent._prime_trigger(self, callback)
-        super()._prime(callback)
+        if not self._primed:
```
imphil, line 448: Nit: I'd return early here on `if self._primed` to reduce indentation, but either style is fine.

imphil (approved):

PR #4142, opened 2024-08-30
Logic array `int` constructor + various fixes

ktbarrett:
The feature split from #4101 for supporting passing `int` as shorthand for Range. This makes the LogicArray constructor somewhat backwards compatible with BinaryValue which took a `BinaryValue(int, int)` overload.

The intent is for int to be able to be taken as the second positional argument, or as the 'width' keyword argument, while only Range can be passed to 'range' by keyword argument. The current implementation still allows the user to write `range=4`, but is considerably less logic than enforcing otherwise. Type hints *could* help, but we can't write positional-only argument before Python 3.8.

There's also various fixes to documentation, typing, code simplification. Some overloads were removed just so there's less duplicate documentation. Performance is of little concern to methods like `index` and `count`, which will probably never get used.

### DIFF
--- a/src/cocotb/types/array.py
+++ b/src/cocotb/types/array.py
@@ -1,7 +1,7 @@
 # Copyright cocotb contributors
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
-from typing import Iterable, Iterator, Optional, TypeVar, Union, cast, overload
+from typing import Iterable, Iterator, TypeVar, Union, cast, overload
 
 from cocotb.types import ArrayLike
 from cocotb.types.range import Range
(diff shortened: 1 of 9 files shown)


src/cocotb/types/array.py
```diff
@@ -19,6 +19,7 @@ class Array(ArrayLike[T]):
     The indexes of an array can start or end at any integer value, they are not limited to 0-based indexing.
     Indexing schemes can be either ascending or descending in value.
     An array's indexes are described using a :class:`~cocotb.types.Range` object.
+    Passing an :class:`int` for the *range* parameter acts as shorthand for ``Range(range-1, "downto", 0)``.
```
imphil, line 22:
```suggestion
    Passing an :class:`int` as second positional argument acts as shorthand for ``Range(range-1, "downto", 0)``.
```

Using wording like this would make it less tepmting to write `range=17`?

src/cocotb/types/logic_array.py
```diff
@@ -34,13 +33,13 @@ class LogicArray(ArrayLike[Logic]):
 
     Like :class:`Array`, if no *range* argument is given, it is deduced from the length
     of the iterable used to initialize the variable.
-
+    Passing an :class:`int` for the *range* parameter acts as shorthand for ``Range(range-1, "downto", 0)``.
```
imphil, line 36:
```suggestion
    Passing an :class:`int` as second positional argument acts as shorthand for ``Range(range-1, "downto", 0)``.
```

as above.

imphil (approved):

PR #4143, opened 2024-08-30
Add (deprecated) `__bool__` cast for LogicArray

ktbarrett:
This was done for backwards compatibility using the same semantics as BinaryValue.

There are many possible reasonable implementations of value checking:
* all values == logical 1 `all(v == 1 for v in value)`
* any value == logical 1 `any(v == 1 for v in value)`
* `or` reduction of value == logical 1 `reduce(or_, value) == 1`

We should refuse the temptation to guess (as Verilog has done) and have users *explicitly* mention the behavior they are looking for, so this was deprecated.

Not advertising this as a feature with a newsfragment since we don't want users to necessary use it.

### DIFF
--- a/src/cocotb/types/logic_array.py
+++ b/src/cocotb/types/logic_array.py
@@ -803,6 +803,14 @@ class LogicArray(ArrayLike[Logic]):
     def __invert__(self) -> "LogicArray":
         return LogicArray(~v for v in self)
 
+    def __bool__(self) -> bool:
+        warnings.warn(
+            "The behavior of bool casts and using LogicArray in conditionals may change in the future. "
+            """Use explicit comparisons (e.g. `LogicArray() == "11"`) instead""",
+            FutureWarning,
+        )
+        return any(v in (Logic("H"), Logic("1")) for v in self)
+
 
 def _make_range(
     range: Union[Range, int, None], width: Union[int, None]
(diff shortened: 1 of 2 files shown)


p12tic:
My vote doesn't count, but this PR fixes one of more annoying problems when upgrading code to be compatible with upcoming cocotb 2.0. Constructs such as `if abc.value:` silently changed their meaning. I'm very glad to see this addressed with at least a warning.

imphil:
Thanks for your feedback, @p12tic! Of course we're not voting, but having real-world feedback is extremely valuable. Please keep it coming!

imphil (approved):
I think the semantics for bool casts that we have today (and added back in this PR) are the most natural one: it's true if it's non-zero. But I agree, guessing on behavior isn't making a testbench robust.

So:
* Let's keep the current behavior, as this PR does.
* I wouldn't spell out the behavior we might introduce in the future. Returning `True` all the time is one option, throwing an exception would be another one, making it work silently only for width 1 and throwing an exception otherewise might be another option.

ktbarrett:
> It could probably even make sense to throw an exception, as __bool__ that always returns True is probably not very useful.

This is just the default for all Python objects. `object.__bool__` always returns `True`. I think the intent was that `None` is `False`, so you can use conditionals to do null checks.

Issue #4117, opened 2024-08-20
Explore autodoc2 for Sphinx


imphil:
https://sphinx-autodoc2.readthedocs.io/en/latest/ seems to be new and looks interesting. There have been other autodoc replacements in the past, including ones which do static analysis (as opposed to runtime-inspection of the documented source). Worth having a look.

FYI @cmarqu

PR #4146, opened 2024-09-02
Experiment with sphinx-autodoc2

cmarqu:
Refs #4117

While `sphinx-autodoc2` is promising, its most important drawback for us is the lack of support for Google-style docstrings (see https://github.com/sphinx-extensions2/sphinx-autodoc2/issues/33).


cmarqu:
Commented out some objects for now that don't have `__all__` defined:
https://sphinx-autodoc2.readthedocs.io/en/stable/quickstart.html#documenting-only-the-public-api-via-all

ktbarrett:
`TypeVar`s aren't properly documented or linking like they used to.

PR #4149, opened 2024-09-04
Explicitly track Task state

### LINKED ISSUE #3975: Assess `Task.has_started`
It's used in two places:
1. In Task to know when to close the coroutine object to prevent warnings
2. In the scheduler's dispatch to `trigger_from_unstarted_coroutine`

Neither require it to be a public method. Point 2 is actually wrong, since "unstarted" also includes queued but hasn't been run yet. This causes the scheduler to unnecessarily attempt to schedule a scheduled Task again.

Considering that last case we should assess whether the current definition is *correct* as a public API, or whether it should even be a public API. It might be useful to be a public API since our system you can make Tasks that haven't been scheduled (by using `cocotb.create_task` or `Task(coro())`), but the important distinction IMO is *has been scheduled*, not *has been resumed at least once*.

ktbarrett:
* Refactors `_task_done_callback` to not raise exceptions and thus require stripping parts of the traceback.
   * Properly handles cancelled tasks.
* Refactors Task to explicitly handle the current state.
   * Introduces new "scheduled" state for when a Task is currently in the `Scheduler._pending_tasks` queue.
   * Properly handles calling `Task.kill()` and `Task.cancel` on cancelled Tasks (used to check for `_outcome`, now checks for `Task.done()`).

Closes #3975.

I wanted to get rid of `has_started`. This required replacement with another variable that is actually useful (i.e. it tracks whether it's queued or not). This led to a revamp of Task state tracking. The revamp also realized a few uncaught bugs in the Task state logic regarding cancelling already cancelled or finished tasks, and setting outcome (as if the Task had finished) when the Task was cancelled. The latter bug mentioned required changing the `_task_done_exception` as there was no `_outcome` to `get` even though the Task was finished when it was cancelled; hence the first commit.

### TODO
- [x] Document state?
- [x] Newsfragment

### DIFF
--- a/src/cocotb/_scheduler.py
+++ b/src/cocotb/_scheduler.py
@@ -387,7 +387,9 @@ class Scheduler:
 
     def _resume_task_upon(self, task: Task[Any], trigger: Trigger) -> None:
         """Schedule `task` to be resumed when `trigger` fires."""
+        # TODO Move this all into Task
         task._trigger = trigger
+        task._state = Task._State.PENDING
 
         trigger_tasks = self._trigger2tasks.setdefault(trigger, [])
         trigger_tasks.append(task)
@@ -421,6 +423,8 @@ class Scheduler:
         # Don't queue the same task more than once (gh-2503)
         if task in self._pending_tasks:
             raise InternalError("Task was queued more than once.")
+        # TODO Move state tracking into Task
+        task._state = Task._State.SCHEDULED
         self._pending_tasks[task] = outcome
 
     def _queue_function(self, task):
@@ -525,8 +529,9 @@ class Scheduler:
         if isinstance(result, Trigger):
             return result
 
+        # TODO move this into Task
         if isinstance(result, Task):
-            if not result.has_started() and result not in self._pending_tasks:
+            if result._state is Task._State.UNSTARTED:
                 return self._trigger_from_unstarted_task(result)
             else:
                 return self._trigger_from_started_task(result)
--- a/src/cocotb/_scheduler.py
+++ b/src/cocotb/_scheduler.py
@@ -228,7 +228,7 @@ class Scheduler:
             _py_compat.insertion_ordered_dict()
         )
 
-        self._pending_tasks: OrderedDict[Task[Any], _outcomes.Outcome] = OrderedDict()
+        self._scheduled_tasks: OrderedDict[Task[Any], _outcomes.Outcome] = OrderedDict()
         self._pending_threads = []
         self._pending_events = []  # Events we need to call set on once we've unwound
 
@@ -315,7 +315,7 @@ class Scheduler:
         for task in scheduling:
             # unset trigger
             task._trigger = None
-            self._queue(task)
+            self._schedule_task(task)
 
         # cleanup trigger
         trigger._cleanup()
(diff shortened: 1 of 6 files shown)


src/cocotb/task.py
```diff
@@ -40,6 +41,16 @@ class Task(Generic[ResultType]):
         Use :meth:`result`, :meth:`done`, and :meth:`done` methods instead, respectively.
     """
 
+    class _State(DocEnum):
+        """State of a Task."""
+
+        UNSTARTED = (auto(), "Task hasn't started running")
+        SCHEDULED = (auto(), "Task in Scheduler queue to run soon")
+        PENDING = (auto(), "Task waiting for Trigger to fire")
+        RUNNING = (auto(), "Task is the currently running")
```
marlonjames, line 50:
```suggestion
        RUNNING = (auto(), "Task is currently running")
```

src/cocotb/task.py
```diff
@@ -188,31 +202,36 @@ def join(self) -> "cocotb.triggers._Join[ResultType]":
         """
         return cocotb.triggers._Join(self)
 
-    def has_started(self) -> bool:
```
marlonjames, line 191: newsfrag for removal?

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -442,7 +454,7 @@ async def coroutine_timer():
     coro_task = cocotb.start_soon(coroutine_outer())
```
marlonjames, line 454: Fix comment `# created but not scheduled yet` above

src/cocotb/_scheduler.py
```diff
@@ -421,6 +423,8 @@ def _queue(
         # Don't queue the same task more than once (gh-2503)
         if task in self._pending_tasks:
             raise InternalError("Task was queued more than once.")
+        # TODO Move state tracking into Task
+        task._state = Task._State.SCHEDULED
         self._pending_tasks[task] = outcome
```
marlonjames, line 428: Rename `_pending_tasks` to reflect `SCHEDULED` state, if tasks waiting on a `Trigger` are in the `PENDING` state.

src/cocotb/task.py
```diff
@@ -40,6 +41,16 @@ class Task(Generic[ResultType]):
         Use :meth:`result`, :meth:`done`, and :meth:`done` methods instead, respectively.
     """
 
+    class _State(DocEnum):
+        """State of a Task."""
+
+        UNSTARTED = (auto(), "Task hasn't started running")
```
marlonjames, line 47: ?
```suggestion
        UNSTARTED = (auto(), "Task created but hasn't started running or been scheduled to run")
```

src/cocotb/task.py
```diff
@@ -93,13 +104,15 @@ def _get_coro_stack(self) -> Any:
...
             fmt = "<{name} finished coro={coro}() outcome={outcome}>"
-        elif self._trigger is not None:
+        elif self._state is Task._State.PENDING:
             fmt = "<{name} pending coro={coro}() trigger={trigger}>"
-        elif not self._started:
+        elif self._state is Task._State.SCHEDULED:
+            fmt = "<{name} scheduled coro={coro}()>"
+        elif self._state is Task._State.UNSTARTED:
             fmt = "<{name} created coro={coro}()>"
         else:
```
marlonjames, line 117: Need a repr for cancelled that should include the `CancelledError`'s message and probably the Task outcome.

marlonjames (approved):
typo

PR #4169, opened 2024-09-22
Documentation updates

imphil:
See the individual commits.

- ~**Document the backport process**~ (Removed here, will be a separate PR)
- **Resolve various Sphinx warnings**
- **Backport cocotb 1.9 release notes**

### DIFF
--- a/src/cocotb/regression.py
+++ b/src/cocotb/regression.py
@@ -126,8 +126,8 @@ class Test:
             Units of ``timeout_time``, accepts any units that :class:`~cocotb.triggers.Timer` does.
 
         expect_fail:
-            If ``True`` and the test fails a functional check via an ``assert`` statement, :pytest:class:`pytest.raises`,
-            :pytest:class:`pytest.warns`, or :pytest:class:`pytest.deprecated_call`, the test is considered to have passed.
+            If ``True`` and the test fails a functional check via an ``assert`` statement, :func:`pytest.raises`,
+            :func:`pytest.warns`, or :func:`pytest.deprecated_call`, the test is considered to have passed.
             If ``True`` and the test passes successfully, the test is considered to have failed.
 
         expect_error:
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -389,7 +389,7 @@ Features
   .. consume the towncrier issue number on this line. (:pr:`1507`)
 - Added support for Aldec's Active-HDL simulator. (:pr:`1601`)
 - Including ``Makefile.inc`` from user makefiles is now a no-op and deprecated. Lines like  ``include $(shell cocotb-config --makefiles)/Makefile.inc`` can be removed from user makefiles without loss in functionality. (:pr:`1629`)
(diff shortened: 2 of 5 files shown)


imphil:
Note that I didn't remove the duplicated release notes from 2.0 and 1.9, I don't think that's worth the effort and/or helping.

docs/source/maint_processes.rst
```diff
@@ -0,0 +1,41 @@
+********************
```
ktbarrett, line 1: Does this belong in the docs? WE have CONTRIBUTING.md that has a lot of the current development flow documentation.

ktbarrett:
The rest of the docs changes look good. Just need to decide where to put developer and maintainer documentation.

docs/source/maint_processes.rst
imphil: Hm, you're right. But then, `CONTRIBUTING.md` grew too big. It started as a small document and evolved into a second documentation next to the actual readthedocs documentation (and the wiki). I don't think that's good for discoverability, so I'd opt for moving some of the CONTRIBUTING.md content also into the main docs. 

If others agree, we can do that all in one go here, or separately (or not at all?)

@cmarqu thoughts?

imphil:
I removed the backport documentation for now to get the simpler changes in. I'm working on moving the contributing.md contents into our main docs, and will also include the backport process documentation there.

src/cocotb/regression.py
```diff
@@ -126,8 +126,8 @@ class Test:
             Units of ``timeout_time``, accepts any units that :class:`~cocotb.triggers.Timer` does.
 
         expect_fail:
-            If ``True`` and the test fails a functional check via an ``assert`` statement, :pytest:class:`pytest.raises`,
-            :pytest:class:`pytest.warns`, or :pytest:class:`pytest.deprecated_call`, the test is considered to have passed.
+            If ``True`` and the test fails a functional check via an ``assert`` statement, :class:`pytest.raises`,
+            :class:`pytest.warns`, or :class:`pytest.deprecated_call`, the test is considered to have passed.
```
cmarqu, line 130: The old code was supposed to use the pytest intersphinx setup from https://github.com/cocotb/cocotb/blob/e8c61232559d79787c5c4b306988608d9b742211/docs/source/conf.py#L56 which somehow doesn't produce a link (anymore?), but the change doesn't produce a link either.

docs/source/maint_processes.rst
cmarqu: We should at least leave a stub `CONTRIBUTING.md` pointing to the Sphinx doc section since that file is kinda standard and e.g. GitHub will check for it in https://github.com/cocotb/cocotb/community

ktbarrett (approved):

src/cocotb/regression.py
imphil: Hm. I thought I got it to produce a link again, but apparently not. I tried some other variants, but no luck. I'll leave it as-is for now, it at least doesn't produce a warning any more.

PR #4179, opened 2024-09-26
Throw `CancelledError` when Tasks are cancelled

ktbarrett:
Depends on #4149 and #4178. xref #2679.

Still need to change cancellation so it is scheduled and not immediate to prevent scheduler re-entrance, and so Tasks are cancelled at the end of Test and that taken into account for Test pass/fail, but this is a step in the right direction.

### DIFF
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -158,8 +158,7 @@ class Task(Generic[ResultType]):
             self._outcome = Error(remove_traceback_frames(e, ["_advance", "send"]))
             self._state = Task._State.FINISHED
 
-        if self.done():
-            self._do_done_callbacks()
+        self._do_done_callbacks()
 
     def kill(self) -> None:
         """Kill a coroutine."""
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -7,6 +7,7 @@ import logging
 import os
 import warnings
 from asyncio import CancelledError, InvalidStateError
+from enum import auto
 from typing import Any, Callable, Coroutine, Generator, Generic, List, Optional, TypeVar
 
 import cocotb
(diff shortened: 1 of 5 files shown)


src/cocotb/task.py
```diff
@@ -152,6 +151,9 @@ def _advance(self, outcome: Outcome) -> Any:
         except StopIteration as e:
             self._outcome = Value(e.value)
             self._state = Task._State.FINISHED
+        except CancelledError as e:
+            self._outcome = Error(remove_traceback_frames(e, ["_advance", "send"]))
+            self._state = Task._State.CANCELLED
```
marlonjames, line 156: Doesn't this allow the task to suppress the `CancelledError`, if it's only set as `CANCELLED` after the task raises a `CancelledError`?
We need the `CancellationError` from #3874

src/cocotb/task.py
```diff
@@ -210,21 +212,18 @@ def cancel(self, msg: Optional[str] = None) -> None:
...
-            FutureWarning,
-            stacklevel=2,
-        )
+        if msg is not None:
+            self._cancelled_error = CancelledError(msg)
+        else:
+            self._cancelled_error = CancelledError()
+
+        # throw CancelledError and unschedule
+        self._advance(Error(self._cancelled_error))
```
marlonjames, line 220: If `cancel()` is called before this finishes we'll have problems, but the followup should fix that.

src/cocotb/task.py
```diff
@@ -210,21 +212,18 @@ def cancel(self, msg: Optional[str] = None) -> None:
         if self.done():
```
marlonjames, line 211: As long as the state is updated after running the Task with `CancelledError`, calling `cancel()` again will override the already set `_cancelled_error`. The order of state update and running the Task will have to be reviewed in your scheduling followup PR.

PR #4180, opened 2024-09-26
Remove `Task.kill()`

ktbarrett:
Depends on #4179.

Removes the `Task.kill()` method which is replaced with `Task.cancel()` to prevent its scheduler re-entrancy issues. A PR will need to be also added to 1.9 to deprecate `Task.kill()` and to mention `Task.cancel()` as an alternative.


tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -45,16 +45,16 @@ async def clock_yield(task):
...
     clk_task = cocotb.start_soon(Clock(dut.clk, 100, "ns").start())
     await Timer(100, "ns")
     cocotb.start_soon(clock_yield(clk_task))
     await Timer(100, "ns")
-    clk_task.kill()
+    clk_task.cancel()
     assert not test_flag
     await Timer(1000, "ns")
-    assert test_flag
+    assert not test_flag
```
ktbarrett, line 57: Yeah... somethign was definitely broken before, lol...

PR #4204, opened 2024-10-08
Make `test_3316` run but expect a fail when `--timing` is not set

### LINKED ISSUE #4040: Fix `test_3316`
The test if for Verilator's `--timing` mode, which is useful, but this test is not run as a part of CI since it demands the user pass `EXTRA_ARG=--timing` or exits. Change this so it just runs both `--timing` and `--no-timing`.

dolphingarlic:
Closes #4040 

Not sure if this is a good way to run both tests, but I don't know how else to do it


imphil:
Looks like it works, so that's all that matters in Make :smile:  Just one small addition: Can you make the new targets `PHONY`?

dolphingarlic:
Why don't the other tests have `.PHONY: all`? The example Makefile has it for some reason

imphil:
Probably an oversight, it's unlikely that we hit the case where we have a file called "all". Makefiles are too easy to get working, but still slightly wrong :) But no need to worry about that here, thanks for your update!

I'll take another look once I'm back at my PC, but the change looks good at first sight!

ktbarrett (commented):
The design is not rebuilt on the second test. You'll need to stick them in separate SIM_BUILD folders.

dolphingarlic:
Oh oops fixed. (Also, now the `--no-timing` test fails because Verilator ignores the `#5` delay in the testbench. How should I handle this error in the test?)

ktbarrett:
Considering the other tests in the 3316 suite only run one of `--timing` or `--no-timing`, I'm guessing that the same should have been true for test_3316_a. So maybe change it so only `--timing` is run.

dolphingarlic:
But then wouldn't `test_3316_a` be the same as `test_3316_c` without the cocotb version check? Is this fine?

ktbarrett:
Ah, you are correct. Looking at the README for why this test was set up this way it looks like A was never intended to be run or always fail. Maybe the correct answer here is to just remove test_3316_a altogether. We don't need to test if Verilator fails in this case, that's Verilator's responsibility.

Sorry if the issue implied otherwise, but I knew something wasn't right and hadn't really put any effort into figuring out what should be done until now.

dolphingarlic:
Hmm I feel like we should still keep the test just for completeness. I'll just modify the test to make it expect a fail for now, but let me know if you'd rather have me remove that test entirely

ktbarrett (approved):

tests/test_cases/test_3316_a/Makefile
```diff
@@ -13,9 +13,9 @@ all:
 else ifeq ($(filter verilator,$(shell echo $(SIM) | tr A-Z a-z)),)
 all:
 	@echo "Skipping test since only Verilator is supported"
-else ifeq ($(filter $(EXTRA_ARGS),--no-timing --timing),)
-all:
-	@echo "Skipping test since Verilator requires --timing or --no-timing option."
 else
+.PHONY: override_for_this_test
+override_for_this_test:
+	if $(MAKE) all; then echo "Expected test to fail without --timing flag"; false; else echo "Failed as expected"; fi
```
cmarqu, line 19:
```suggestion
	# we cannot normally rely on the simulator return code, but this test is specific to Verilator
	if $(MAKE) all; then echo "Expected test to fail without --timing flag"; false; else echo "Failed as expected"; fi
```
(if it's true that Verilator behaves in this way)

ktbarrett:
@dolphingarlic Thank you for your patience and for getting this done.

PR #4208, opened 2024-10-09
Prevent transitions from ReadOnly to ReadWrite or ReadOnly

### LINKED ISSUE #1245: ReadWrite after ReadOnly

### LINKED ISSUE #2222: Track current time stage and prevent improper usage in Python
`test_readwrite_in_readonly` and `test_cached_write_in_readonly` exist to test invalid actions, like waiting for `ReadWrite` while in `ReadOnly`, and doing writes in `ReadOnly`. The issue is that doing these actions cause certain simulators to error or fail hard, which is why it is skipped in several simulators. Instead of letting the simulator decide what to do, we should keep track of the current time state and raise errors early.

ktbarrett:
Transitioning from ReadOnly to ReadWrite or ReadOnly are not legal, this checks prevents that from occuring by raising a `RuntimeError` when it is attempted. Closes #1245. Closes #2222.

### DIFF
--- a/docs/source/newsfragments/4208.change.rst
+++ b/docs/source/newsfragments/4208.change.rst
@@ -1,1 +1,1 @@
-Attempting to await on either the :class:`~cocotb.triggers.ReadWrite` or :class:`~cocotb.triggers.ReadOnly` trigger while in the ReadOnly state now raises a :exc:`RuntimeError`.
+Attempting to await on either the :class:`~cocotb.triggers.ReadWrite` or :class:`~cocotb.triggers.ReadOnly` trigger while in the ReadOnly phase now raises a :exc:`RuntimeError`.
--- a/docs/source/newsfragments/4208.change.rst
+++ b/docs/source/newsfragments/4208.change.rst
@@ -1,0 +1,1 @@
+Attempting to await on either the :class:`~cocotb.triggers.ReadWrite` or :class:`~cocotb.triggers.ReadOnly` trigger while in the ReadOnly state now raises a :exc:`RuntimeError`.
--- a/src/cocotb/triggers.py
+++ b/src/cocotb/triggers.py
@@ -282,7 +282,7 @@ class ReadOnly(GPITrigger, metaclass=_ParameterizedSingletonGPITriggerMetaclass)
     def _prime(self, callback: Callable[[Trigger], None]) -> None:
         if cocotb.sim_phase is cocotb.SimPhase.READ_ONLY:
             raise RuntimeError(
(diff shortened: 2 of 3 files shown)


docs/source/newsfragments/4208.change.rst
```diff
@@ -0,0 +1 @@
+Attempting to await on either the :class:`~cocotb.triggers.ReadWrite` or :class:`~cocotb.triggers.ReadOnly` trigger while in the ReadOnly state now raises a :exc:`RuntimeError`.
```
cmarqu, line 1:
```suggestion
Attempting to await on either the :class:`~cocotb.triggers.ReadWrite` or :class:`~cocotb.triggers.ReadOnly` trigger while in the ReadOnly phase now raises a :exc:`RuntimeError`.
```
We are calling this "phase" now in most places (https://docs.cocotb.org/en/latest/library_reference.html)

src/cocotb/triggers.py
```diff
@@ -280,6 +280,10 @@ def __singleton_key__(cls) -> None:
         return None
 
     def _prime(self, callback: Callable[[Trigger], None]) -> None:
+        if cocotb.sim_phase is cocotb.SimPhase.READ_ONLY:
+            raise RuntimeError(
+                "Attempted illegal transition: awaiting ReadOnly in ReadOnly state"
```
cmarqu, line 285:
```suggestion
                "Attempted illegal transition: awaiting ReadOnly in ReadOnly phase"
```

src/cocotb/triggers.py
```diff
@@ -298,6 +302,10 @@ def __singleton_key__(cls) -> None:
         return None
 
     def _prime(self, callback: Callable[[Trigger], None]) -> None:
+        if cocotb.sim_phase is cocotb.SimPhase.READ_ONLY:
+            raise RuntimeError(
+                "Attempted illegal transition: awaiting ReadWrite in ReadOnly state"
```
cmarqu, line 307:
```suggestion
                "Attempted illegal transition: awaiting ReadWrite in ReadOnly phase"
```

cmarqu (approved):

PR #4212, opened 2024-10-10
Make VHPI detect null ranges correctly

### LINKED ISSUE #3606: VHPI doesn't take null ranges into account
https://github.com/cocotb/cocotb/blob/ee652a27a32c6a0f284c6054a7c07742920fe1a5/src/cocotb/share/lib/vhpi/VhpiCbHdl.cpp#L231-L235

Computing the number of elements using left and right cannot be assumed because of null ranges. `0 downto 7` has a length of 0, but the algorithm above would compute the length as 8.

EDIT: I don't see a way to check direction, but it could be inferred by checking if the range is null using `vhpiIsNullP`. Is there a better way @marlonjames?

dolphingarlic:
Closes #3606

(Draft for now because I can't test this locally ☹️)


ktbarrett:
You can start with the public logs. I'll try to get you the private CI failures tonight or tomorrow.

dolphingarlic:
I'm so confused, why does this work for nvc and not ghdl 😕 (also sorry for the spam; I only just found out that I can test locally using nvc)

imphil:
GHDL uses VPI, NVC uses VHPI.

ktbarrett:
@imphil Could you get him the Questa logs? I'm away from my personal computer for the week.

imphil:
@dolphingarlic Can you reach out to me at philipp@fossi-foundation.org? Thanks!

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -433,3 +433,22 @@ async def watch():
...
+async def test_null_range_width(dut):
+    # Normal arrays should have the same length regardless of language
+    assert len(dut.array_7_downto_4) == 4
+    if LANGUAGE in ["vhdl"]:
+        # But in VHDL, `4 downto 7` should result in a null range
+        try:
+            dut.array_4_downto_7
+        except AttributeError:
+            return
+        assert False, "Expected null array to be deleted"
```
imphil, line 451: Drive-by (full review upcoming once you got it working): Have a look at https://docs.pytest.org/en/stable/how-to/assert.html#assertions-about-expected-exceptions for a nicer way to test an expected exception.

src/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
```diff
@@ -230,11 +241,15 @@ int VhpiArrayObjHdl::initialise(const std::string &name,
...
     }
 
-    if (m_range_left > m_range_right) {
+    if (!range_is_up) {
         m_num_elems = m_range_left - m_range_right + 1;
     } else {
         m_num_elems = m_range_right - m_range_left + 1;
     }
+    if (m_num_elems < 0) {
+        LOG_DEBUG("VHPI: Null array... Delete object")
```
ktbarrett, line 250: This is definitely not the behavior we want. We want to expose null arrays.

ktbarrett:
What I had envisioned was a new GPI method to get the direction so the Python code could use that instead of guessing based on the range. I want the null arrays exposed as often the range values are important in those instances. We won't have to worry about writes to such objects as all values will be > 0 length in size and will cause an exception.

ktbarrett:
I'll have a look at the Questa logs in a few hours, but the implementation is looking good so far. 

I would update the typing in `simulator.pyi` file and if you can add constants for the direction rather than using magic numbers (the `dir == 1`). There are other examples of adding int constants in the simulatormodule.cpp file.

And thanks for working on this!

dolphingarlic:
What types should I update in the .pyi file?

ktbarrett:
`get_range` now returns 3 values. Also I forgot there is also the documentation/annotation in simulatormodule.cpp for that function that should be updated for the new return value as well.

ktbarrett:
Yeah looks like cocotb is seeing array's be 0 length that aren't when using Questa's VHPI.

ktbarrett:
I guess for now we will have to use the old behavior when using Questa's VHPI. You'll have to add an ifdef with logic that guesses the direction for now. The define used is `MODELSIM`. Once the upstream bug is fixed we can maybe remove it.

dolphingarlic:
Ok looks like Questa VHPI is still failing `test_discovery:recursive_discover`. I think it's because it's trying to index into `array_4_downto_7` because cocotb thinks it has 4 items but it's actually a null array. Any ideas for how to get around this? Do I just tell the test to expect to fail for Questa VHPI?

dolphingarlic:
Also I'm kind of surprised that GHDL does not fail that test case...

src/cocotb/share/lib/fli/FliObjHdl.cpp
```diff
@@ -97,6 +97,7 @@ int FliValueObjHdl::initialise(const std::string &name,
     if (get_type() == GPI_ARRAY) {
         m_range_left = mti_TickLeft(m_val_type);
         m_range_right = mti_TickRight(m_val_type);
+        m_range_dir = (gpi_range_dir)mti_TickDir(m_val_type);
```
imphil, line 100:
```suggestion
        m_range_dir = static_cast<gpi_range_dir>(mti_TickDir(m_val_type));
```

(C++, please.)

dolphingarlic:
Just bumping this @ktbarrett @imphil

src/cocotb/handle.py
```diff
@@ -179,19 +179,11 @@ class RangeableObjectMixin(SimHandleBase):
...
-        if length == 0:
-            direction = "downto" if left <= right else "to"
-        else:
-            direction = "to" if left <= right else "downto"
-
-        return Range(left, direction, right)
+        left, right, direction = self._handle.get_range()
+        assert (
+            direction != simulator.RANGE_NO_DIR
+        ), "Expected range to have a direction but got none!"
```
ktbarrett, line 185:
```suggestion
        if direction == simulator.RANGE_NO_DIR:
            raise RuntimeError("Expected range to have a direction but got none!")
```

ktbarrett (approved):
Perfect!

ktbarrett:
Oh, maybe add a newsfragment, this is a useful new feature.

docs/source/newsfragments/4212.change.rst
```diff
@@ -0,0 +1 @@
+The GPI `get_range` method now returns a range's direction in addition to the left and right bounds. When using VHPI or ModelSim's FLI, the direction is set explicitly by the `to`/`downto` keywords in the range definition; otherwise, the direction is inferred by the relative values of the left and right bounds. This change also allows null range sizes to be inferred correctly in VHDL.
```
cmarqu, line 1:
```suggestion
The GPI ``get_range`` method now returns a range's direction in addition to the left and right bounds. When using VHPI or ModelSim's FLI, the direction is set explicitly by the ``to``/``downto`` keywords in the range definition; otherwise, the direction is inferred by the relative values of the left and right bounds. This change also allows null range sizes to be inferred correctly in VHDL.
```
Maybe also replace "VHPI or ModelSim's FLI" with just "VHDL", we don't need to be too technical here.

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1284,8 +1289,11 @@ static PyMethodDef gpi_sim_hdl_methods[] = {
     {"get_range", (PyCFunction)get_range, METH_NOARGS,
      PyDoc_STR("get_range($self)\n"
                "--\n\n"
-               "get_range() -> Tuple[int, int]\n"
-               "Get the range of elements (tuple) contained in the handle.")},
+               "get_range() -> Tuple[int, int, int]\n"
+               "Get the range of elements (tuple) contained in the handle. "
+               "The first two elements of the tuple specify the left and "
+               "right bounds, while the third specifies the direction (`1` "
+               "for ascending, `-1` for descending, and `0` for undefined).")},
```
cmarqu, line 1296:
```suggestion
               "right bounds, while the third specifies the direction (``1`` "
               "for ascending, ``-1`` for descending, and ``0`` for undefined).")},
```
Use monospace, not italics for ReST rendering (https://cocotb--4212.org.readthedocs.build/en/4212/library_reference.html#cocotb.simulator.gpi_sim_hdl.get_range)

cmarqu (approved):
Thanks for the polishing!

PR #4216, opened 2024-10-12
Build with Verilator in parallel

Aidan-McNay:
Minor update: use the `-j` flag in the call to Verilator's Makefile to speed up the build process (basic profiling indicated ~3x speedup for a single source file, likely scaling to high speedup for more source files)

I also use specific `CFLAGS` in the call to `verilator` in my installation to obtain a little more speedup on build; however, I'm unsure if these would be welcome, and may require more testing to make sure they're not specific to just `g++`

### DIFF
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -10,6 +10,7 @@
 # TODO: support custom dependencies
 
 import logging
+import multiprocessing
 import os
 import re
 import shlex
@@ -1300,6 +1301,8 @@ class Verilator(Runner):
         cmds.append(
             [
                 "make",
+                "-j",
+                f"{multiprocessing.cpu_count()}"
                 "-C",
                 str(self.build_dir),
                 "-f",
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -1302,7 +1302,7 @@ class Verilator(Runner):
             [
                 "make",
                 "-j",
-                f"{multiprocessing.cpu_count()}"
+                f"{multiprocessing.cpu_count()}",
                 "-C",
                 str(self.build_dir),
                 "-f",


ktbarrett (approved):

src/cocotb_tools/runner.py
```diff
@@ -1300,6 +1301,8 @@ def _build_command(self) -> List[_Command]:
         cmds.append(
             [
                 "make",
+                "-j",
+                f"{multiprocessing.cpu_count()}",
```
imphil, line 1305: Just using the CPU count is likely to be counterproductive on very large machines with many dozens of cores (both because fellow users might hate you, and because it's only producing diminishing results). The "best" value is impossible to find generally, hence it would be good to:

* Make the amount of parallelism configurable (even if only by overwriting the Verilator class in the runner)
* Cap the parallelism at 4 or 8.

ktbarrett: Environment variable might work as well. I wonder if there is a standard for that variable already?

ktbarrett: `MAKEFLAGS` can be used to pass `-j` to all `make` invocations, but it assumes we are using make and that isn't going to affect other invocations of make. All other build system use 1 or more custom environment variable for this. So maybe we just do that too. Opinions @Aidan-McNay?

Aidan-McNay: That makes sense; maybe something like this?
https://github.com/cocotb/cocotb/blob/ebbed876556bf45eec88e311d9894906c761a640/src/cocotb_tools/runner.py#L1219-L1220
I didn't see a dedicated naming convention across simulators, so was contemplating either `MAKE_JOBS` or `BUILD_JOBS`? The latter might be applicable across other simulators if needed (maybe an optional argument to `runner.build`?), but the former would show exactly what's going on?

themperek: As @ktbarrett said can you just `export MAKEFLAGS="-j$(nproc)"`?

Aidan-McNay: Yup - exporting the environment variable works. I think it runs into the same question of whether to put the burden on the user to remember to export `MAKEFLAGS` whenever they use `cocotb`, or whether we would set a default in `self.env`?

(Notably, this doesn't work if you use [Verilator's MAKEFLAGS argument](https://verilator.org/guide/latest/exe_verilator.html#cmdoption-MAKEFLAGS); this is because it only works when using [`--build`](https://verilator.org/guide/latest/exe_verilator.html#cmdoption-build). I'm not quite sure why the Verilator runner separates the verilating and building of the model into two separate commands in this case, since they're just called back-to-back; maybe another thing to change, could include in this PR or a separate one)

Aidan-McNay: That makes sense; I can modify to use that and update docs. Only had two thoughts (feel free to ignore if it's not what you're looking for):

- I'm not an expert on the other simulators, but I'd imagine they'd have similar options for parallel building. We could then have an envvar like `BUILD_PARALLEL`, which could be used across simulators where applicable?
- I'm also wondering whether we want to support the basic case of someone just wanting to use cocotb with verilator out of the box (i.e. without setting environment variables), and still get good performance. It sounds like the environment variable above wouldn't cause parallel building, not sure if we want to have some parallelism as the default when the envvar isn't present?

ktbarrett:
Really sorry, I've been distracted recently. I will look at this tonight. I agree maybe defaulting to parallelism is preferable.

src/cocotb_tools/runner.py
imphil: As said earlier: I'd be happy with a default of min($(ncpus), 4). That gives reasonable speed-up before diminishing results kick in and before overloading shared systems.

ktbarrett:
Closing in favor of #4257.

PR #4225, opened 2024-10-17
Add robustness on as_tcl_value() function

Cepadues:
Simple check on input's length before trying to access it.
As feeding a blank string would result in:

```
if value[0] == '"':
          ~~~~~^^^
IndexError: string index out of range
```

Could you please double-check the test case, as I'm not sure that I've understood the function's expected behaviour?

I didn't understand how I could get the PR ID before pushing the PR, so I didn't create a news fragment.
Sorry :(

PS : the link in [CONTRIBUTING.md](https://github.com/cocotb/cocotb/blob/master/CONTRIBUTING.md) returns 404

### DIFF
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -49,12 +49,13 @@ _space_re = re.compile(r"([\s])", re.ASCII)
 
 
 def _as_tcl_value(value: str) -> str:
-    # add '\' before special characters and spaces
-    value = _magic_re.sub(r"\\\1", value)
-    value = value.replace("\n", r"\n")
-    value = _space_re.sub(r"\\\1", value)
-    if value[0] == '"':
-        value = "\\" + value
+    if len(value):
+        # add '\' before special characters and spaces
+        value = _magic_re.sub(r"\\\1", value)
+        value = value.replace("\n", r"\n")
+        value = _space_re.sub(r"\\\1", value)
+        if value[0] == '"':
+            value = "\\" + value
 
     return value
 
--- a/tests/test_cases/test_as_tcl_value/Makefile
+++ b/tests/test_cases/test_as_tcl_value/Makefile
@@ -1,0 +1,19 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+
+TOPLEVEL_LANG ?= verilog
+
+ifneq ($(TOPLEVEL_LANG),verilog)
+
+all:
+	@echo "Skipping test due to TOPLEVEL_LANG=$(TOPLEVEL_LANG) not being verilog"
+clean::
+
+else
+
+include ../../designs/sample_module/Makefile
+COCOTB_TEST_MODULES:=test_as_tcl_value
+
+endif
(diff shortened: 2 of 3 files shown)


cmarqu:
You could probably just add the test to https://github.com/cocotb/cocotb/blob/master/tests/pytest/test_runner.py, no need for the Makefile.
You also don't need a newsfragment because the effect is not (or should not be) user-visible.

cmarqu:
Not in front of a terminal right now, but it looks like the following links would help:
* https://stackoverflow.com/a/5309528/41347
* https://github.com/python/cpython/blob/d8c864816121547338efa43c56e3f75ead98a924/Lib/tkinter/__init__.py#L63
* https://github.com/python/cpython/blob/d8c864816121547338efa43c56e3f75ead98a924/Lib/test/test_tcl.py#L326

Maybe this could work as a roundtrip test? (untested):
```python3
from cocotb.runner import as_tcl_value
import Tkinter

tclsh = Tkinter.Tcl()
orig_str = "..."
conv_str = tclsh.eval(as_tcl_value(orig_str))
assert conv_str == orig_str
```

ktbarrett:
There is also https://github.com/cocotb/cocotb/pull/3391 which does similar work. Turns out its a whole lot more work to make all the simulators happy than it should be.

ktbarrett (approved):

tests/pytest/test_runner.py
```diff
@@ -25,6 +26,14 @@
 }
 
 
+def test_empty_string():
+  assert _as_tcl_value("") == ""
+
+
+def test_special_char():
+  assert _as_tcl_value("Test \n end\ttest\r") == "Test\\ \\n\\ end\\\ttest\\\r"
+
```
ktbarrett, line 35:
```suggestion
def test_empty_string():
    assert _as_tcl_value("") == ""


def test_special_char():
    assert _as_tcl_value("Test \n end\ttest\r") == "Test\\ \\n\\ end\\\ttest\\\r"

```

tests/pytest/test_runner.py
```diff
@@ -11,6 +11,7 @@
 import cocotb
 from cocotb.triggers import Timer
 from cocotb_tools.runner import get_runner
+from cocotb_tools.runner import _as_tcl_value
```
ktbarrett, line 14:
```suggestion
from cocotb_tools.runner import _as_tcl_value, get_runner
```

src/cocotb_tools/runner.py
```diff
@@ -49,12 +49,13 @@
 
 
 def _as_tcl_value(value: str) -> str:
-    # add '\' before special characters and spaces
-    value = _magic_re.sub(r"\\\1", value)
-    value = value.replace("\n", r"\n")
-    value = _space_re.sub(r"\\\1", value)
-    if value[0] == '"':
```
eric-wieser, line 56: A simpler patch is
```suggestion
    if value[:1] == '"':
```
which avoids needing any other changes.

ktbarrett:
@Gophen Sorry about taking my time with this. Is the current set of changes good by you? They do pass your tests. Will merge in a couple days otherwise.

Gophen:
It's okay, there's no need to rush, everything looks good to me ;)

PR #4232, opened 2024-10-22
Update test_fatal test to reference correct issue

ktbarrett:
Also testing to see if Riviera still fails.


ktbarrett:
Riviera works!

themperek (approved):

PR #4235, opened 2024-10-22
Exit cleanly on HDL errors in verilator sims.

### LINKED ISSUE #4231: Verilator does not call cbEndOfSimulation if SV finishes simulation
A good example of this is test_fatal. `fatal()` is called, which ends the sim, but the callback is not called, so the simulation does not exit cleanly.

In Verilator `fatal()` just calls C `abort()` after running it's custom callback interface. The custom interface could be plumbed via the VPI.

gilbertoabram:
Use `Verilated::addExitCb' to invoke the end of sim callbacks, which shut things down orderly. I verified that this does not result in a double call.

closes #4231


src/cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -50,6 +50,8 @@ static inline bool settle_value_callbacks() {
     return cbs_called;
 }
 
+void clean_exit_cb(void*) { VerilatedVpi::callCbs(cbEndOfSimulation); }
```
ktbarrett, line 53:
```suggestion
static void clean_exit_cb(void*) { VerilatedVpi::callCbs(cbEndOfSimulation); }
```

src/cocotb_tools/makefiles/simulators/Makefile.verilator
```diff
@@ -72,7 +72,7 @@ $(SIM_BUILD)/Vtop: $(SIM_BUILD)/Vtop.mk
 $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/Vtop $(CUSTOM_SIM_DEPS)
 	$(RM) $(COCOTB_RESULTS_FILE)
 
-	COCOTB_TEST_MODULES=$(call deprecate,MODULE,COCOTB_TEST_MODULES) COCOTB_TESTCASE=$(call deprecate,TESTCASE,COCOTB_TESTCASE) COCOTB_TEST_FILTER=$(COCOTB_TEST_FILTER) COCOTB_TOPLEVEL=$(call deprecate,TOPLEVEL,COCOTB_TOPLEVEL) TOPLEVEL_LANG=$(TOPLEVEL_LANG) \
+	-COCOTB_TEST_MODULES=$(call deprecate,MODULE,COCOTB_TEST_MODULES) COCOTB_TESTCASE=$(call deprecate,TESTCASE,COCOTB_TESTCASE) COCOTB_TEST_FILTER=$(COCOTB_TEST_FILTER) COCOTB_TOPLEVEL=$(call deprecate,TOPLEVEL,COCOTB_TOPLEVEL) TOPLEVEL_LANG=$(TOPLEVEL_LANG) \
```
ktbarrett, line 75: Now that you mention it, we probably should be doing this in other simulators as well.

ktbarrett (commented):
What about the Python runner in `cocotb_tools.runner`?

src/cocotb_tools/runner.py
```diff
@@ -427,7 +427,14 @@ def test(
...
         cmds: Sequence[_Command] = self._test_command()
-        self._execute(cmds, cwd=self.test_dir)
+        try:
+            self._execute(cmds, cwd=self.test_dir)
+        except subprocess.CalledProcessError as e:
+            if isinstance(self, Verilator):
+                # Verilator aborts on $fatal, but will still generate results.
+                self.log.info("Simulation failed: %d", e.returncode)
+            else:
+                raise
```
ktbarrett, line 437: I think what we need to do here is save the return code and set the exit code correctly after checking the results file.

So:
* Catch error here, log about simulation failure
* If results file does not exist, log about file not existing, `sys.exit()` with the return code from the simulator
* If results file exists and there are test failure, log about test failures, `sys.exit(1)`
* If results file exists and there are no test failures, `sys.exit()` with the return code from the simulator

ktbarrett: And we can leave the Makefile ignoring the error altogether the default as its what other simulators do.

src/cocotb_tools/runner.py
```diff
@@ -488,13 +507,13 @@ def get_results(results_xml_file: Path) -> Tuple[int, int]:
         Tuple of number of tests and number of fails.
 
     Raises:
```
gilbertoabram, line 509: Can this also raise on parsing the file? Is that a problem?

src/cocotb_tools/runner.py
```diff
@@ -427,13 +427,32 @@ def test(
...
         if pytest_current_test:
-            check_results_file(results_xml_file)
+            try:
+                (num_tests, num_failed) = get_results(results_xml_file)
+                if num_failed:
+                    self.log.error(
+                        "ERROR: Failed %d of %d tests.", num_failed, num_tests
+                    )
+                    error_code = 1 if simulator_exit_code == 0 else simulator_exit_code
+                    sys.exit(error_code)
```
gilbertoabram, line 449: I was using it in two places at some point. I'll remove the intermediate variable.

src/cocotb_tools/runner.py
ktbarrett: It can, but it isn't an expected failure path, so we should just let it rise.

src/cocotb_tools/runner.py
ktbarrett, line 449: Usually this code is put in an `else` block. `try` blocks should be as minimal as possible.

ktbarrett (approved):

marlonjames (approved):

PR #4240, opened 2024-10-28
Vpi: remove vpiPort from iteration

AndrewNolte:
These aren't proper signals, and not handled by our python code. These split into high/low signals, rather than being signals themselves. The port signals are still accessible through the other iteration schemes.

I also reformatted these so we can view the git blame better in the future, I imagine we have some more that need to be removed.
> 
> In SystemVerilog's VPI (Verilog Procedural Interface), vpiPort is a property that refers to ports in modules or interfaces. It provides access to information about ports, such as connections and direction.
> 
> When dealing with module ports, vpiPort is used to navigate and interact with the connections. Some key properties and functions associated with vpiPort include:
> 
> vpiPortIndex: This returns the index of the port.
> vpiHighConn: This indicates the connection closer to the top module.
> vpiLowConn: This indicates the connection further from the top module.
> vpiPortType: Defines the type of the port, such as vpiInterfacePort, vpiModportPort, or vpiPort.
> For example, you can use vpiPort in combination with other VPI functions to traverse port connections in a design hierarchy, please refer to Section 37.14 of the SystemVerilog LRM on ports.
>

### DIFF
--- a/src/cocotb/share/lib/vpi/VpiIterator.cpp
+++ b/src/cocotb/share/lib/vpi/VpiIterator.cpp
@@ -35,7 +35,7 @@
 
 decltype(VpiIterator::iterate_over) VpiIterator::iterate_over = [] {
     /* for reused lists */
-
+    // clang-format off
     // vpiInstance is the base class for module, program, interface, etc.
     std::vector<int32_t> instance_options = {
         vpiNet,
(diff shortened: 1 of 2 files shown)


ktbarrett:
`test_array:test_discover_all` test expected needs to be updated to 197, as it's no longer discovering a `vpiPort` object.

ktbarrett:
Considering this is a part of #3360, I also don't see this as uncontroversial, merging.

PR #4244, opened 2024-10-29
Add `__index__` to LogicArray to support integer string casting

### LINKED ISSUE #4229: Request to add `value.hexstr`
In a similar way that `dut.o_value.value.binstr`  would it be possible to add `dut.o_value.value.hexstr` as a convenience to `hex(dut.o_value.value.integer)` ?

ktbarrett:
Closes #4229. Allows the use of LogicArray directly in indexing `memory[my_logic_value]` and in integer string casts: `hex()`, `bin()`, etc.

- [x] Add test

### DIFF
--- a/src/cocotb/types/logic_array.py
+++ b/src/cocotb/types/logic_array.py
@@ -122,6 +122,17 @@ class LogicArray(ArrayLike[Logic]):
         >>> la.to_bytes()
         b"\n"
 
+    You can also convert :class:`LogicArray`\ s to hexidecimal or binary strings using
+    the builtins :func:`hex:` and :func:`bin`, respectively.
+
+    .. code-block:: python3
+
+        >>> la = LogicArray("01111010")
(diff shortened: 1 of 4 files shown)


cmarqu:
It would be nice if this was mentioned in the docs.

marlonjames (approved):

src/cocotb/types/logic_array.py
```diff
@@ -122,6 +122,17 @@ class LogicArray(ArrayLike[Logic]):
         >>> la.to_bytes()
         b"\n"
 
+    You can also convert :class:`LogicArray`\ s to hexidecimal or binary strings using
```
cmarqu, line 125:
```suggestion
    You can also convert :class:`LogicArray`\ s to hexadecimal or binary strings using
```

cmarqu (approved):

PR #4253, opened 2024-11-07
Add back resolve x

AndrewNolte:
Adds back COCOTB_RESOLVE_X with the old values, while using low, high, random, and error for the user-passed literals.

### DIFF
--- a/src/cocotb/_py_compat.py
+++ b/src/cocotb/_py_compat.py
@@ -98,3 +98,13 @@ else:
             res = self._method(instance)
             instance.__dict__[self._method.__name__] = res
             return res
+
+
+# inheriting from (str, Enum) was broken in 3.12 and StrEnum must be used
+if sys.version_info >= (3, 12):
+    from enum import StrEnum
+else:
+    from enum import Enum
+
+    class StrEnum(str, Enum):
+        pass
--- a/src/cocotb/_py_compat.py
+++ b/src/cocotb/_py_compat.py
@@ -100,8 +100,8 @@ else:
             return res
 
 
-# inheriting from (str, Enum) was broken in 3.12 and StrEnum must be used
-if sys.version_info >= (3, 12):
+# inheriting from (str, Enum) was broken in 3.11 and StrEnum must be used
+if sys.version_info >= (3, 11):
     from enum import StrEnum
 else:
     from enum import Enum
--- a/src/cocotb/types/logic_array.py
+++ b/src/cocotb/types/logic_array.py
@@ -347,9 +347,9 @@ class LogicArray(ArrayLike[Logic]):
             # May convert list to str before converting to int.
             try:
                 self._value_as_int = int(self._get_str(), 2)
-            except ValueError as e:
+            except ValueError:
                 if resolve == "error":
-                    raise e
+                    raise
 
                 return int(
                     RE_UNRESOLVED.sub(RESOLVE_MAP[resolve], self._get_str()),
--- a/src/cocotb/types/logic_array.py
+++ b/src/cocotb/types/logic_array.py
@@ -1,6 +1,9 @@
 # Copyright cocotb contributors
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
+import os
+import random
+import re
 import warnings
 from math import ceil
 from typing import (
@@ -20,6 +23,22 @@ from cocotb.types.range import Range
 
 if TYPE_CHECKING:  # pragma: no cover
     from typing import Literal
+
+COCOTB_RESOLVE_X: "Literal['error'] | Literal['low'] | Literal['high'] | Literal['random']" = {
+    "VALUE_ERROR": "error",
+    "ZEROS": "low",
+    "ONES": "high",
+    "RANDOM": "random",
+}[os.getenv("COCOTB_RESOLVE_X", "VALUE_ERROR")]  # type: ignore
+
+
+RE_UNRESOLVED = re.compile(r"[^01]")
+
+RESOLVE_MAP = {
+    "low": "0",
+    "high": "1",
+    "random": lambda _: random.choice("01"),
+}
 
 
 class LogicArray(ArrayLike[Logic]):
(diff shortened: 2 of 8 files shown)


src/cocotb/types/logic_array.py
```diff
@@ -320,10 +339,23 @@ def _get_str(self) -> str:
...
+            try:
+                self._value_as_int = int(self._get_str(), 2)
+            except ValueError as e:
+                if resolve == "error":
+                    raise e
+
+                self._value_as_int = int(
+                    RE_UNRESOLVED.sub(RESOLVE_MAP[resolve], self._get_str()),
+                    2,  # type: ignore
+                )
```
ktbarrett, line 357: Can't save it. What happens if they get the integer value again and change the resolve method?

ktbarrett (commented):
WRT to "random" is the intent that it should be different on every cast or just once?

Also, we should consider since this changes the value if we should save the value and invalidate the other representations.

Also, should the start with integer representation in that case? Otherwise if you do:
```python
>>> a = LogicArray("01XZ")
>>> str(a)
'01XZ'
>>> int(a)
4
>>> str(a)
'0100'
```

IIRC the value was updated in the constructor if there were any Xs in BinaryValue.

src/cocotb/types/logic_array.py
```diff
@@ -320,10 +339,23 @@ def _get_str(self) -> str:
...
             # May convert list to str before converting to int.
-            self._value_as_int = int(self._get_str(), 2)
+            try:
+                self._value_as_int = int(self._get_str(), 2)
+            except ValueError:
+                if resolve == "error":
+                    raise
+
+                return int(
+                    RE_UNRESOLVED.sub(RESOLVE_MAP[resolve], self._get_str()),
```
imphil, line 355: Apart from typing, a check for valid values of `resolve` would be nice. Or just use an enum (maybe even preferred).

src/cocotb/types/logic_array.py
```diff
@@ -21,6 +24,22 @@
 if TYPE_CHECKING:  # pragma: no cover
     from typing import Literal
 
+COCOTB_RESOLVE_X: "Literal['error'] | Literal['low'] | Literal['high'] | Literal['random']" = {
+    "VALUE_ERROR": "error",
+    "ZEROS": "low",
+    "ONES": "high",
+    "RANDOM": "random",
```
imphil, line 31: That's the motivation to change the naming here? On a green field, I'd also prefer lower case and `error` over `VALUE_ERROR`, but given where we are today: is the gain big enough to justify the inconsistency between the environment variable and the function argument? I don't think so.

imphil (commented):
Thanks for adding back this feature! Before merging, can we also get a test and documentation added?

src/cocotb/types/logic_array.py
AndrewNolte: I agree- they should share the same set of values. Since we removed and then added it, and it's used a bit differently, I'm on board with renaming the environment variable (COCOTB_LOGIC_RESOLVE ?) and also using the new values. Another reason for renaming is that it's not just X that it resolves. I made them shorter and lower case since they can now be passed in as an arg, not just an environment variable. But if the maintainers want to keep the old names/values, I can switch these back.

ktbarrett: Maybe a StrEnum?

src/cocotb/types/logic_array.py
AndrewNolte: I prefer to not force people to import something extra just to pass an argument, which is why I'm of a fan of the literal type in place of enums, and not a fan of how Range() was added to LogicArray. I think StrEnum is a good combination, but is only available in 3.11. I can add a backport though.

ktbarrett:
@cmarqu This definitely needs a once over for documentation.

ktbarrett:
Need to add tests.

src/cocotb/types/logic_array.py
```diff
@@ -22,6 +25,73 @@
...
+
+    VALUE_ERROR = "error"
+    ZEROS = "zeros"
+    ONES = "ones"
+    RANDOM = "random"
+
+
+# Must add documentation after the fact because Enum member creation is weird
+ResolveX.VALUE_ERROR.__doc__ = (
+    "Throws a ValueError if the LogicArray contains non-``0``/``1`` values."
```
cmarqu, line 44:
```suggestion
    "Throws a :exc:`ValueError` if the :class:`LogicArray` contains non-``0``/``1`` values."
```

src/cocotb/types/logic_array.py
```diff
@@ -22,6 +25,73 @@
     from typing import Literal
 
 
+class ResolveX(StrEnum):
+    """Resolution behaviors supported when converting a LogicArray to an integer.
```
cmarqu, line 29:
```suggestion
    """Resolution behaviors supported when converting a :class:`LogicArray` to an integer.
```

src/cocotb/types/logic_array.py
```diff
@@ -22,6 +25,73 @@
...
+
+RESOLVE_X = ResolveX[os.getenv("COCOTB_RESOLVE_X", "VALUE_ERROR")]
+"""Global default for resolving ``X``, ``Z``, ``U``, ``W``, and ``-`` values to ``0`` or ``1``.
+
+Set using :envvar:`COCOTB_RESOLVE_X` before boot, or via this variable any time thereafter.
+Defaults to :attr:`~ResolveX.VALUE_ERROR`.
+
+.. warning::
+
+    This exists for backwards-compatability reasons.
```
cmarqu, line 61:
```suggestion
    This exists for backwards-compatibility reasons.
```

cmarqu (approved):
Small doc things only

ktbarrett:
pre-commit.ci autofix

cmarqu (approved):

src/cocotb/_py_compat.py
```diff
@@ -98,3 +98,13 @@ def __get__(
             res = self._method(instance)
             instance.__dict__[self._method.__name__] = res
             return res
+
+
+# inheriting from (str, Enum) was broken in 3.12 and StrEnum must be used
+if sys.version_info >= (3, 12):
```
ktbarrett, line 104:
```suggestion
# inheriting from (str, Enum) was broken in 3.11 and StrEnum must be used
if sys.version_info >= (3, 11):
```

PR #4267, opened 2024-11-13
Allow user to specify elaboration arguments for NVC simulator.

jacob720:
This was required to pass the appropriate options for doing coverage with the NVC simulator. Previously no elaboration arguments could be passed to NVC.

### DIFF
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -331,6 +331,7 @@ class Runner(ABC):
                 If not set, run all testcases found in *test_module*.
                 Can be a comma-separated list.
             seed: A specific random seed to use.
+            elab_args: A list of elaboration arguments for the simulator.
             test_args: A list of extra arguments for the simulator.
             plusargs: 'plusargs' to set for the simulator.
             extra_env: Extra environment variables to set.
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -302,6 +302,7 @@ class Runner(ABC):
         gpi_interfaces: Optional[List[str]] = None,
         testcase: Optional[Union[str, Sequence[str]]] = None,
         seed: Optional[Union[str, int]] = None,
+        elab_args: Sequence[str] = [],
         test_args: Sequence[str] = [],
         plusargs: Sequence[str] = [],
         extra_env: Mapping[str, str] = {},
@@ -390,6 +391,7 @@ class Runner(ABC):
 
         self.pre_cmd = pre_cmd
 
+        self.elab_args = list(elab_args)
         self.test_args = list(test_args)
         self.plusargs = list(plusargs)
         self.env = dict(extra_env)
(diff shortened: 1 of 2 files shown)


src/cocotb_tools/runner.py
```diff
@@ -302,6 +302,7 @@ def test(
         gpi_interfaces: Optional[List[str]] = None,
         testcase: Optional[Union[str, Sequence[str]]] = None,
         seed: Optional[Union[str, int]] = None,
+        elab_args: Sequence[str] = [],
```
ktbarrett, line 305: Why would this be in the test call? Isn't elaboration only done once per *build*?

ktbarrett: Maybe not since other sims allow users pass new generics which re-elabs per test.

ktbarrett, line 305: You will though need to add documentation to the docstring below.

ktbarrett (approved):
On further thought, this is a good change. Just a couple nitpicks.

We should end up using this in a couple other simulators as well.

src/cocotb_tools/runner.py
jacob720: Hi, I've added this in the last commits.

ktbarrett (approved):

nmarinos:
Is there any way to apply the --cover argument to nvc when running cocotb 1.9.2 with runners?

m-tosch:
@nmarinos yes, there is a workaround. I use it as well.

I created a simple test runner example which can generate NVC coverage for cocotb 1.9.2

The code highlighted between ### hooks into the NVC test command and adds `--cover` for the elaboration step.

`test_runner.py`
```py
import os

def test_runner():
    sim = os.getenv("SIM", "nvc")

    runner = get_runner(sim)

    runner.build(
        sources = ["foo.vhd"],
        hdl_toplevel = "foo"
    )

###
    original_test_command = runner._test_command

    def custom_test_command(self):
        cmd = original_test_command()
        idx = cmd[0].index("-e") + 2
        cmd[0].insert(idx, "--cover")
        return cmd

    runner._test_command = custom_test_command.__get__(runner, type(runner))
###

    runner.test(
        test_module = "testbench",
        hdl_toplevel = "foo",
        hdl_toplevel_lang = "vhdl",
    )

if __name__ == "__main__":
    test_runner()
```

nmarinos:
> @nmarinos yes, there is a workaround. I use it as well.
> 
> I created a simple test runner example which can generate NVC coverage for cocotb 1.9.2
> 
> The code highlighted between ### hooks into the NVC test command and adds `--cover` for the elaboration step.
> 
> `test_runner.py`
> 
> ```python
> import os
> from cocotb.runner import get_runner
> 
> def test_runner():
>     sim = os.getenv("SIM", "nvc")
> 
>     runner = get_runner(sim)
> 
>     runner.build(
>         sources = ["foo.vhd"],
>         hdl_toplevel = "foo"
>     )
> 
> ###
>     original_test_command = runner._test_command
> 
>     def custom_test_command(self):
>         cmd = original_test_command()
>         idx = cmd[0].index("-e") + 2
>         cmd[0].insert(idx, "--cover")
>         return cmd
> 
>     runner._test_command = custom_test_command.__get__(runner, type(runner))
> ###
> 
>     runner.test(
>         test_module = "testbench",
>         hdl_toplevel = "foo",
>         hdl_toplevel_lang = "vhdl",
>     )
> 
> if __name__ == "__main__":
>     test_runner()
> ```

Thank you very much! This works wonderfully!
It can be used for other elab arguments too:
```
cmd[0].insert(idx, "--cover") 
cmd[0].insert(idx, "--no-collapse") 
```

Issue #4091, opened 2024-08-12
[Tracking] Python configuration and environment issues


ktbarrett:
#3643 and #4065 show issue with our current solution to Python path configuration, which came from #3309. #4066 was a failed experiment to avoid the first of several of the points below.

## Issues
* Some virtual environment tools don't set `VIRTUAL_ENV`
* Ultimately it doesn't really matter if we are in a virtual environment or not when configuring cocotb
* Setting PYTHONHOME to the virtual environment's prefix doesn't work with `venv` or tools that use it like `pipx`
     * This is because the pyvenv.cfg they write doesn't contain base_prefix information, or it isn't setup when Python is initialized for some reason, so it results in `base_prefix` to incorrectly point to the venv rather than the base.
* Setting the program name to the virtual environment's seems to be sufficient for determine `prefix` and `base_prefix` and to setup `sys.path` as a Python interpreter would, but `Py_SetProgramName` inexplicably doesn't work on Windows.

## Proposed Solution
Based on comments in https://bugs.python.org/issue22213, this is a known issue and there is no interest in fixing it directly. The recommended solution there is to use the [Python Config initialization sequence](https://docs.python.org/3/c-api/init_config.html#python-configuration). We would set [`PyConfig.program_name`](https://docs.python.org/3/c-api/init_config.html#c.PyConfig.program_name) which would work on all platforms to the python binary gotten using `cocotb-config --python-bin`, which should return the correct venv executable name even when using isolated package installers like `pipx`.

This solution is Python 3.8+. Users of Python 3.6-3.7 will continue to experience issues.

cmarqu:
You may have seen this, but https://bugs.python.org/issue22213 is not updated anymore, https://github.com/python/cpython/issues/66409 has the current discussion.

Applepi:
Is there an actual solution to fix this on Ubuntu 24.04? I see a bunch of issues but no actual solutions to install the package. installing with pip3 and with pipx results in a broken environment and am unable to run the quick start example.

ktbarrett:
Using `pip` should work on Ubuntu 24.04 if you set up a virtualenv. If it isn't working you may need to do additional work to configure your environment or there may be something else wrong.

Applepi:
> @Applepi Using `pip` should work on Ubuntu 24.04 if you set up a virtualenv. If it isn't working you may need to do additional work to configure your environment or there may be something else wrong.

[...]

PR #4276, opened 2024-11-22
Move Python interpreter initialization to new 3.8+ pattern

ktbarrett:
This is expected to not work with Python 3.6 or 3.7. Related to #4091. Changes Python interpreter initialization to use the new Python 3.8+ initialization sequence. This allows us to set the program name to the Python executable in a way that works consistently in all operating systems and means we no longer have to set PYTHONHOME. Not setting PYTHONHOME also allows us to avoid issues with Xcelium and potentially VCS caused by setting that variable.

### DIFF
--- a/src/cocotb/share/lib/embed/gpi_embed.cpp
+++ b/src/cocotb/share/lib/embed/gpi_embed.cpp
@@ -54,76 +54,19 @@
 #else
 #include <unistd.h>
 #endif
-static PyThreadState *gtstate = NULL;
-
-static wchar_t progname[] = L"cocotb";
-static wchar_t *argv[] = {progname};
-
-#if defined(_WIN32)
-#if defined(__MINGW32__) || defined(__CYGWIN32__)
-constexpr char const *PYTHON_INTERPRETER_PATH = "/Scripts/python";
-#else
-constexpr char const *PYTHON_INTERPRETER_PATH = "\\Scripts\\python";
-#endif
-#else
-constexpr char const *PYTHON_INTERPRETER_PATH = "/bin/python";
-#endif
 
 static PyObject *pEventFn = NULL;
 
-static void set_program_name_in_venv(void) {
-    static wchar_t venv_path_w[PATH_MAX];
+static int python_init_called = 0;
 
-    const char *venv_path_home = getenv("VIRTUAL_ENV");
-    if (!venv_path_home) {
-        LOG_INFO(
-            "Did not detect Python virtual environment. "
-            "Using system-wide Python interpreter");
-        return;
-    }
-
-    std::string venv_path = venv_path_home;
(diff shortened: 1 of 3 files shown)


src/cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -139,16 +82,41 @@ extern "C" COCOTB_EXPORT void _embed_init_python(void) {
...
+    PyConfig config;
+    PyConfig_InitPythonConfig(&config);
+    DEFER(PyConfig_Clear(&config));
+
+    // Set program name to the Python interpreter at PYGPI_PYTHON_BIN.
+    // This is necessary to set up sys.path as if the PyGPI user was invoked as
+    // a Python script.
+    {
+        auto status = PyConfig_SetBytesString(&config, &config.program_name,
+                                              python_bin_path);
```
ktbarrett, line 103: This implicitly PreInitializes the Python interpreter and uses DecodeLocale under the hood, so we don't need that bit of code anymore.

src/cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -139,16 +82,41 @@ extern "C" COCOTB_EXPORT void _embed_init_python(void) {
         }
     }
 
-    to_python();
-    // must set program name to Python executable before initialization, so
-    // initialization can determine path from executable
-    set_program_name_in_venv();
-    Py_Initialize(); /* Initialize the interpreter */
-    PySys_SetArgvEx(1, argv, 0);
```
ktbarrett, line 147: AFAICT setting the argv to `["cocotb"]` does nothing. Ostensibly Python sees that it's a "script" and adds the parent path to `sys.path`, but the cocotb module's parent is already on the path since it is an installed module since cocotb 1.2 or something.

src/cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -139,16 +82,41 @@ extern "C" COCOTB_EXPORT void _embed_init_python(void) {
...
-
-    /* Swap out and return current thread state and release the GIL */
-    gtstate = PyEval_SaveThread();
```
ktbarrett, line 150: This variable was not used for anything besides preventing re-initialization, so it was swapped for a simple bool.

src/cocotb/share/lib/embed/gpi_embed.cpp
eric-wieser: I think the main benefit here is that user code can look at `sys.argv` and determine (in a suboptimal way) that they are running inside cocotb; I think what you describe doesn't happen because `updatepath` (the last argument) is `0`.

ktbarrett:
Closing in favor of #4293.

PR #3951, opened 2024-06-19
Replacing the no longer supported option '+acc+1' with '-debug_access+r+w-memcbk -debug_region+cell'

NickolayTernovoy:
In the vcs Version V-2023.12-SP1-1_Full64, I encountered the following error while building the project:

```
Error-[OBSOLETE_OPTION_PLUSACC] Obsolete option '+acc' used
The specified option '+acc+1' is no longer supported.
Please use '-debug_access+r+w-memcbk -debug_region+cell' instead.
```

```
make[1]: *** [/tmp/venv/lib64/python3.6/site-packages/cocotb/share/makefiles/simulators/Makefile.vcs:78: sim_build/simv] Error 255
```

I replaced the option `+acc+1` with the suggested option from vcs. I verified the solution on versions Version T-2022.06-1_Full64 and Version V-2023.12-SP1-1_Full64.

PR #4278, opened 2024-11-25
Vcs `-debug_access+all`

ipburbank:
`-memcbk` prevented multi-dimensional arrays, rams, etc from being dumped to waves.

image

Introduced here: https://github.com/cocotb/cocotb/pull/3951

### DIFF
--- a/src/cocotb_tools/makefiles/simulators/Makefile.vcs
+++ b/src/cocotb_tools/makefiles/simulators/Makefile.vcs
@@ -77,7 +77,7 @@ $(SIM_BUILD)/pli.tab : | $(SIM_BUILD)
 $(SIM_BUILD)/simv: $(VERILOG_SOURCES) $(SIM_BUILD)/pli.tab $(CUSTOM_COMPILE_DEPS) | $(SIM_BUILD)
 	cd $(SIM_BUILD) && \
 	COCOTB_TOPLEVEL=$(call deprecate,TOPLEVEL,COCOTB_TOPLEVEL) \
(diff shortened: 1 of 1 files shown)


ipburbank:
@NickolayTernovoy Before this merges something horrible, perhaps was there a reason you added this initially?

NickolayTernovoy:
I described the issue in this PR. I added the options and flags that were specified in the error message from VCS.

https://github.com/cocotb/cocotb/pull/3951

ipburbank:
> The specified option '+acc+1' is no longer supported.
> Please use '-debug_access+r+w-memcbk -debug_region+cell' instead.

To me, the automated message is surprising in that it contains `-memcbk`.  I feel this change is safe then, but am not an expert.

image

ipburbank:
Honestly this should be configurable from the user's makefile, but probably out of scope for this PR.

ktbarrett (approved):

ktbarrett:
@NickolayTernovoy 

> as this option is no longer supported in the latest versions of VSC

What option are you referring to? `+acc` was removed. Do you mean `+all` was removed in the most recent version?

NickolayTernovoy:
> @NickolayTernovoy
> 
> > as this option is no longer supported in the latest versions of VSC
> 
> What option are you referring to? `+acc` was removed. Do you mean `+all` was removed in the most recent version?

Exactly. We are talking about different parameters. Everything is correct, all looks good. I mixed up `all` and `acc` during the discussion – now everything seems correct.

PR #4280, opened 2024-11-26
The Refactor

ktbarrett:
This PR reorganizes all the core components of cocotb: Triggers, Tasks, the Scheduler, and Regression Manager. The intent is to reorganize the code into a more flexible configuration for follow ons to support alternative Regression Managers, CancelledError, and support multiple kinds of Tasks for implementing TaskManagers. It will also reorganize the core components in a way that is more similar to asyncio, potentially allowing us to move to asyncio in the future to replace the Scheduler, Tasks, and base Trigger objects.

TODO
- [ ] Add more context to this PR

### DIFF
--- a/src/cocotb/regression.py
+++ b/src/cocotb/regression.py
@@ -1272,3 +1272,21 @@ class TestFactory(Generic[F]):
 
             glbs["__cocotb_tests__"].append(test)
             glbs[test.name] = test
+
+
+class _RunningTest(Task[None]):
+    """
+    The result of calling a :class:`cocotb.test` decorated object.
+
+    All this class does is change ``__name__`` to show "Test" instead of "Task".
+
+    .. versionchanged:: 1.8.0
+        Moved to the ``cocotb.task`` module.
+    """
+
+    _name: str = "Test"
+
+    def __init__(self, inst: Coroutine[Any, Any, None], name: str) -> None:
+        super().__init__(inst)
+        self.__name__ = f"{type(self)._name} {name}"
+        self.__qualname__ = self.__name__
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -8,7 +8,18 @@ import os
 import warnings
 from asyncio import CancelledError, InvalidStateError
 from enum import auto
-from typing import Any, Callable, Coroutine, Generator, Generic, List, Optional, TypeVar
+from typing import (
+    Any,
+    Callable,
+    Coroutine,
+    Generator,
+    Generic,
+    List,
+    Optional,
+    Set,
+    TypeVar,
+    Union,
+)
 
 import cocotb
 import cocotb.triggers
@@ -285,31 +296,17 @@ class Task(Generic[ResultType]):
         self._done_callbacks.append(callback)
 
     def __await__(self) -> Generator[Any, Any, ResultType]:
-        # It's tempting to use `return (yield from self._coro)` here,
-        # which bypasses the scheduler. Unfortunately, this means that
-        # we can't keep track of the result or state of the coroutine,
-        # things which we expose in our public API. If you want the
-        # efficiency of bypassing the scheduler, remove the `@coroutine`
-        # decorator from your `async` functions.
-
-        # Hand the coroutine back to the scheduler trampoline.
         yield self
         return self.result()
 
 
-class _RunningTest(Task[None]):
-    """
-    The result of calling a :class:`cocotb.test` decorated object.
+_all_tasks: Set[Task[Any]] = set()
+_current_task: Union[Task[Any], None] = None
 
-    All this class does is change ``__name__`` to show "Test" instead of "Task".
 
-    .. versionchanged:: 1.8.0
-        Moved to the ``cocotb.task`` module.
-    """
+def all_tasks() -> Set[Task[Any]]:
+    return _all_tasks
 
-    _name: str = "Test"
 
-    def __init__(self, inst: Coroutine[Any, Any, None], name: str) -> None:
-        super().__init__(inst)
-        self.__name__ = f"{type(self)._name} {name}"
-        self.__qualname__ = self.__name__
+def current_task() -> Union[Task[Any], None]:
+    return _current_task
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -88,6 +88,8 @@ class Task(Generic[ResultType]):
 
         self._task_id = self._id_count
         type(self)._id_count += 1
+        _all_tasks.add(self)
+
         self.__name__ = f"{type(self)._name} {self._task_id}"
         self.__qualname__ = self.__name__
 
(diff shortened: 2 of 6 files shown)


src/cocotb/triggers.py
```diff
@@ -74,80 +76,108 @@ def _pointer_str(obj: object) -> str:
...
-        Sub-classes must override this, but should end by calling the base class
-        method.
+    @abstractmethod
+    def _prime(self) -> None:
+        """Setup the underlying trigger mechanism.
 
-        .. warning::
-            Do not call this directly within a :term:`task`. It is intended to be used
-            only by the scheduler.
+        This should set ther underlying trigger mechanism to call :meth:`_fire`.
```
cmarqu, line 110:
```suggestion
        This should set the underlying trigger mechanism to call :meth:`_fire`.
```

src/cocotb/triggers.py
```diff
@@ -74,80 +76,108 @@ def _pointer_str(obj: object) -> str:
...
         """
-        self._primed = True
 
+    @abstractmethod
     def _unprime(self) -> None:
-        """Remove the callback, and perform cleanup if necessary.
+        """Disable and cleanup the underlying trigger mechanism before it fires."""
```
cmarqu, line 115:
```suggestion
        """Disable and clean up the underlying trigger mechanism before it fires."""
```

src/cocotb/triggers.py
```diff
@@ -74,80 +76,108 @@ def _pointer_str(obj: object) -> str:
...
+
+    def _cleanup(self) -> None:
+        """Cleanup the underlying trigger mechanism after it fires."""
```
cmarqu, line 118:
```suggestion
        """Clean up the underlying trigger mechanism after it fires."""
```

src/cocotb/triggers.py
```diff
@@ -74,80 +76,108 @@ def _pointer_str(obj: object) -> str:
...
+        pass
 
-        After being un-primed, a Trigger may be re-primed again in the future.
-        Calling `_unprime` multiple times is allowed, subsequent calls should be
-        a no-op.
+    def _register(self, cb: Callable[[], None]) -> _CallbackHandle:
+        """Registers the given callback to be called when the Trigger 'fires'.
```
cmarqu, line 122:
```suggestion
        """Register the given callback to be called when the Trigger 'fires'.
```

src/cocotb/triggers.py
```diff
@@ -74,80 +76,108 @@ def _pointer_str(obj: object) -> str:
...
-        .. warning::
-            Do not call this directly within a :term:`task`. It is intended to be used
-            only by the scheduler.
+        Returns:
+            A cancellable handle to the given callback.
         """
-        self._cleanup()
+        res = _CallbackHandle(cb, self)
+        self._cb_handles[res] = None
+        # _prime must come after adding to _cb_handles incase _prime calls _react
```
cmarqu, line 131:
```suggestion
        # _prime must come after adding to _cb_handles in case _prime calls _react
```

src/cocotb/triggers.py
```diff
@@ -74,80 +76,108 @@ def _pointer_str(obj: object) -> str:
...
+        if not self._cb_handles:
+            self._prime()
+        return res
 
-    def _cleanup(self) -> None:
-        self._primed = False
+    def _deregister(self, cb_handle: _CallbackHandle) -> None:
+        """Prevents the given callback from being called once the Trigger fires.
```
cmarqu, line 137:
```suggestion
        """Prevent the given callback from being called once the Trigger fires.
```

src/cocotb/triggers.py
```diff
@@ -74,80 +76,108 @@ def _pointer_str(obj: object) -> str:
...
+
+        Args:
+            cb_handle: The Handle to the callback previously registered.
+        """
+        self._cb_handles.pop(cb_handle)
+        if not self._cb_handles:
+            self._unprime()
+
+    def _react(self) -> None:
+        """Calls all registered callbacks when the Trigger 'fires'."""
```
cmarqu, line 149:
```suggestion
        """Call all registered callbacks when the Trigger 'fires'."""
```

src/cocotb/triggers.py
```diff
@@ -74,80 +76,108 @@ def _pointer_str(obj: object) -> str:
...
-        # else:
-        self._cbhdl: Optional[simulator.gpi_cb_hdl] = None
+    def _react(self) -> None:
+        with profiling_context:
+            global _current_gpi_trigger
+            _current_gpi_trigger = self
+            Trigger._react()
+            cocotb._scheduler_inst._event_loop()
+
+    # _prime in subclasses should set up _cbhdl vaiable with GPI callback handle
```
cmarqu, line 173:
```suggestion
    # _prime in subclasses should set up _cbhdl variable with GPI callback handle
```

cmarqu (commented):
I know it's a draft but couldn't help it :)

src/cocotb/regression.py
```diff
@@ -1273,3 +1264,18 @@ async def _my_test(dut, kwargs: Dict[str, Any] = kwargs) -> None:
...
+
+class _RunningTest(Task[None]):
+    """
+    The result of calling a :class:`cocotb.test` decorated object.
+
+    All this class does is change ``__name__`` to show "Test" instead of "Task".
+
+    .. versionchanged:: 1.8.0
+        Moved to the ``cocotb.task`` module.
+    """
```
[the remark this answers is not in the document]
ktbarrett: They are private classes now so we could even just remove these comments. Developers can just use git history to figure out where things have been.

src/cocotb/task.py
```diff
@@ -41,20 +61,10 @@ class Task(Generic[ResultType]):
...
-        RUNNING = (auto(), "Task is currently running")
-        FINISHED = (auto(), "Task has finished with a value or Exception")
-        CANCELLED = (auto(), "Task was cancelled before it finished")
-
-    _name: str = "Task"  # class name of schedulable task
     _id_count = 0  # used by the scheduler for debug
 
-    def __init__(self, inst):
+    def __init__(self, inst: Coroutine[cocotb.triggers.Trigger, None, Any]) -> None:
+        self._native_coroutine = False
```
ktbarrett, line 67: Get rid of this variable, switch to a `inspect.iscoroutine` check instead. Remove the check in `_get_coro_stack`. It's private and we guard it's usage in `__repr__` so we don't have to worry about doing redundant checks.

src/cocotb/task.py
```diff
@@ -281,35 +326,27 @@ def _add_done_callback(self, callback: Callable[["Task[ResultType]"], Any]) -> N
...
-        # things which we expose in our public API. If you want the
-        # efficiency of bypassing the scheduler, remove the `@coroutine`
-        # decorator from your `async` functions.
-
-        # Hand the coroutine back to the scheduler trampoline.
-        yield self
+        if self._state is TaskState.UNSTARTED:
+            self._schedule_resume()
+        if not self.done():
+            yield self._join
```
ktbarrett, line 337: Avoids redundant checks.
```suggestion
            self._schedule_resume()
            yield self._join
        elif not self.done():
            yield self._join
```

src/cocotb/task.py
```diff
@@ -281,35 +326,27 @@ def _add_done_callback(self, callback: Callable[["Task[ResultType]"], Any]) -> N
...
-    """
-    The result of calling a :class:`cocotb.test` decorated object.
+_all_tasks: Set[Task[Any]] = set()
 
-    All this class does is change ``__name__`` to show "Test" instead of "Task".
 
-    .. versionchanged:: 1.8.0
-        Moved to the ``cocotb.task`` module.
-    """
+def all_tasks() -> Set[Task[Any]]:
```
ktbarrett, line 344: Type this so it's a read-only set.

src/cocotb/triggers.py
```diff
@@ -74,80 +77,114 @@ def _pointer_str(obj: object) -> str:
...
-        Sub-classes must override this, but should end by calling the base class
-        method.
+    @abstractmethod
+    def _prime(self) -> None:
+        """Setup the underlying trigger mechanism.
 
-        .. warning::
-            Do not call this directly within a :term:`task`. It is intended to be used
-            only by the scheduler.
+        This should set the underlying trigger mechanism to call :meth:`_fire`.
```
ktbarrett, line 116:
```suggestion
        This should set the underlying trigger mechanism to call :meth:`_react`.
```

PR #4283, opened 2024-11-30
Disable icarus waveform output when `WAVES` disabled

TrAyZeN:
When `WAVES` was disabled a waveform in the default format (vcd) was produced. This PR passes `-none` to icarus to disable waveform output.


ktbarrett:
I'm guessing this happens when first setting WAVES then later unsetting it? This seems reasonable.

TrAyZeN:
Yes, with cocotb Makefile runner, I think, this only happens by later disabling it. However, with Python runner it can also happen by passing `waves=True` to `Runner.build` and passing `waves=False` to `Runner.test`. I think the runner design could be improved on that, but I didn't come up with a better one.

Issue #3746, opened 2024-02-28
CI: mingw fails to compile Icarus Verilog and hangs after "mkdir dep"


imphil:
Currently, Windows CI fails to compile Icarus Verilog; since the job times out:

```
Wed, 28 Feb 2024 18:57:22 GMT g++ -shared -Wl,--enable-auto-image-base  -o sizer.tgt sizer.o scan_lpms.o scan_logs.o -L.. -livl
Wed, 28 Feb 2024 18:57:22 GMT make[1]: Leaving directory '/d/a/cocotb/cocotb/iverilog/tgt-sizer'
Wed, 28 Feb 2024 18:57:22 GMT make[1]: Entering directory '/d/a/cocotb/cocotb/iverilog/driver'
Wed, 28 Feb 2024 18:57:22 GMT mkdir dep
Wed, 28 Feb 2024 19:29:43 GMT Error: The operation was canceled.
```

A diff between the [last passing](https://github.com/cocotb/cocotb/actions/runs/8031851937/job/21940743798) release run and the [first failing](https://github.com/cocotb/cocotb/actions/runs/8031854068/job/21940795642) release run is here:
[passing-failing.diff.txt](https://github.com/cocotb/cocotb/files/14440210/passing-failing.diff.txt)

Observations:
* In all cases I have seen the timeout happens after one of the "mkdir dep" steps.
* From the last passing run, the next step after the failing command should be ` gcc -I. -I..  -DHAVE_CONFIG_H -Wall -Wextra -Wshadow -Wstrict-prototypes -D__USE_MINGW_ANSI_STDIO=1  -g -O2 -MD -c -DIVL_ROOT='"/c/iverilog/lib/ivl"' -DIVL_SUFFIX='""' -DIVL_INC='"/c/iverilog/include"' -DIVL_LIB='"/c/iverilog/lib"' -DDLLIB='""' -DIVL_INCLUDE_INSTALL_DIR="\"\"" ./main.c`, which looks like a totally normal gcc call.
* There are no noticeable software updates between a passing and a failing run.
* The first failing command conveniently [updated only pre-commit hooks](https://github.com/cocotb/cocotb/commit/6511d5a9e1c49937583479e7ef3ff55465dc06f7), i.e., there's no actual cocotb code change between the last passing and the first failing CI run.
* The diff shows slightly different ordering of commands, but it looks more like stdout interleaving being different than actual differences in execution (it's a parallel make, so who knows).

[...]

PR #4293, opened 2024-12-08
gpi_embed: set the program name to PYTHON_BIN

FFY00:
See #4091.

This patch removes all special virtual environment handling, and instead of setting `PYTHONHOME`, it just sets the program name. Additionally, it uses the new Python Initialization API on >=3.8.

The previous code was setting `PYTHONHOME` to `sys.prefix`, which is wrong on, it should be `sys.base_prefix`. The home value sets the initial values of `base_prefix` and `base_exec_prefix`, which are searched for the standard library, this isn't what we want to do here. We want make sure that when running in a virtual environment, it is activated, which depends only on the executable path.

If `Py_SetProgramName` is indeed broken on Windows, then we could use the workaround I mentioned in the issue.

This also fixes the incorrect virtual env detection message, which relies on `VIRTUAL_ENV` being set, which is not guaranteed.

### DIFF
--- a/docs/source/newsfragments/4293.change.rst
+++ b/docs/source/newsfragments/4293.change.rst
@@ -1,0 +1,1 @@
+Python interpreter embedding is now more robust. Setting ``PYTHONHOME`` is no longer required, so simulator usage of Python scripts should no longer fail.
--- a/docs/source/newsfragments/4293.change.rst
+++ b/docs/source/newsfragments/4293.change.rst
@@ -1,1 +1,1 @@
-Python interpreter embedding is now more robust. Setting ``PYTHONHOME`` is no longer required, so simulator usage of Python scripts should no longer fail.
+Python interpreter embedding is now more robust when using Python 3.8+. Setting ``PYTHONHOME`` is no longer required, so simulator usage of Python scripts should no longer fail.
--- a/src/cocotb/share/lib/embed/gpi_embed.cpp
+++ b/src/cocotb/share/lib/embed/gpi_embed.cpp
@@ -64,28 +64,34 @@ static PyObject *pEventFn = NULL;
 static int get_interpreter_path(wchar_t *path, size_t path_size) {
     const char *path_c = getenv("PYGPI_PYTHON_BIN");
     if (!path_c) {
+        // LCOV_EXCL_START
         LOG_ERROR(
             "PYGPI_PYTHON_BIN variable not set. Can't initialize Python "
             "interpreter!");
         return -1;
+        // LCOV_EXCL_STOP
     }
 
     auto path_temp = Py_DecodeLocale(path_c, NULL);
     if (path_temp == NULL) {
+        // LCOV_EXCL_START
         LOG_ERROR(
             "Unable to set Python Program Name. "
             "Decoding error in Python executable path.");
         LOG_INFO("Python executable path: %s", path_c);
         return -1;
+        // LCOV_EXCL_STOP
     }
     DEFER(PyMem_RawFree(path_temp));
 
     wcsncpy(path, path_temp, path_size / sizeof(wchar_t));
     if (path[(path_size / sizeof(wchar_t)) - 1]) {
+        // LCOV_EXCL_START
         LOG_ERROR(
             "Unable to set Python Program Name. Path to interpreter too long");
         LOG_INFO("Python executable path: %s", path_c);
         return -1;
+        // LCOV_EXCL_STOP
     }
 
     return 0;
@@ -94,8 +100,10 @@ static int get_interpreter_path(wchar_t *path, size_t path_size) {
 /** Initialize the Python interpreter */
 extern "C" COCOTB_EXPORT void _embed_init_python(void) {
     if (python_init_called) {
+        // LCOV_EXCL_START
         LOG_ERROR("PyGPI library initialized again!");
         return;
+        // LCOV_EXCL_STOP
     }
     python_init_called = 1;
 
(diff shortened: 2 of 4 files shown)


ktbarrett:
`PYTHON_BIN` also needs to be set in `cocotb_tools.runner.test` and `PYTHONHOME` removed.

https://github.com/cocotb/cocotb/blob/28b53df48c8b710dda66b7baa30503f1b5aacfd0/src/cocotb_tools/runner.py#L165

ktbarrett (commented):
`PYTHON_INTERPRETER_PATH` is now unused and causing linting failures.

src/cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -140,11 +130,71 @@ extern "C" COCOTB_EXPORT void _embed_init_python(void) {
...
+    }
+    LOG_INFO("Using Python interpreter at %ls", interpreter_path);
+
+#if PY_VERSION_HEX >= 0x3080000
+    /* Use the new Python Initialization Configuration from Python 3.8. */
+    PyConfig config;
+    PyStatus status;
+
+    PyConfig_InitPythonConfig(&config);
+    PyConfig_SetString(&config, &config.program_name, interpreter_path);
```
ktbarrett, line 150:
```suggestion
    DEFER(PyConfig_Clear(&config));

    PyConfig_SetString(&config, &config.program_name, interpreter_path);
```

Then you can remove all the calls below. This is prefered.

src/cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -140,11 +130,71 @@ extern "C" COCOTB_EXPORT void _embed_init_python(void) {
...
+    if (sys_executable_obj == NULL) {
+        LOG_ERROR("Failed to load sys.executable")
+    } else if (PyUnicode_AsWideChar(sys_executable_obj,
+                                    sys_executable,
+                                    sizeof(sys_executable)) == -1) {
+        LOG_ERROR("Failed to convert sys.executable to wide string");
+    } else if (wcscmp(interpreter_path, sys_executable) != 0) {
+        LOG_ERROR("Unexpected sys.executable value (expected '%ls', got '%ls')",
+                  sys_executable, interpreter_path);
+    }
```
ktbarrett, line 197: Nice...

ktbarrett (commented):
I appreciate the work here.

ktbarrett:
I'll combine the best of this and #4276 in a new PR tonight.

ktbarrett (approved):

ktbarrett:
Low coverage is low branch coverage due to our inability to use `GCOV_EXCL_BR` due to our inability to use gcovr 5 due to our support of Python <3.8.

src/cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -135,20 +117,89 @@ extern "C" COCOTB_EXPORT void _embed_init_python(void) {
...
+    } else if (PyUnicode_AsWideChar(sys_executable_obj, sys_executable,
+                                    sizeof(sys_executable)) == -1) {
+        // LCOV_EXCL_START
+        LOG_ERROR("Failed to convert sys.executable to wide string");
+        return;
+        // LCOV_EXCL_STOP
+    } else if (wcscmp(interpreter_path, sys_executable) != 0) {
+        // LCOV_EXCL_START
+        LOG_ERROR("Unexpected sys.executable value (expected '%ls', got '%ls')",
+                  sys_executable, interpreter_path);
```
FFY00, line 199: I mistakenly got this the other way round.

```suggestion
        LOG_ERROR("Unexpected sys.executable value (expected '%ls', got '%ls')",
                  interpreter_path, sys_executable);
```

src/cocotb/share/lib/embed/gpi_embed.cpp
```diff
@@ -135,20 +117,89 @@ extern "C" COCOTB_EXPORT void _embed_init_python(void) {
...
+        return;
+        // LCOV_EXCL_STOP
+    }
```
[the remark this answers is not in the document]
ktbarrett: Would you expect that a mismatch here to be indicative of an error, or it is just for debugging aid in case of a failure? Users tend to not be happy about seeing errors which aren't actually errors.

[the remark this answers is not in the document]
FFY00: Yes, I think so, but I don't think it's critical enough for us to *have* to fail. Especially considering the code this is replacing, which soft-fails when it can't set the program name on virtual environments, for eg.

cmarqu (commented):
I suppose this should have a newsfragment (https://docs.cocotb.org/en/latest/contributing.html#patch-requirements); maybe something like "Finding the right Python executable is now more robust." is already enough.

cmarqu:
The two fails on Windows are our old friend #3746.

ktbarrett:
Will merge tomorrow during maintainers meeting unless there are comments.

marlonjames (approved):

docs/source/newsfragments/4293.change.rst
```diff
@@ -0,0 +1 @@
+Python interpreter embedding is now more robust. Setting ``PYTHONHOME`` is no longer required, so simulator usage of Python scripts should no longer fail.
```
cmarqu, line 1: We should add that this only is the case for 3.8+.

ktbarrett:
Add warning for install cocotb with Windows where Python < 3.8.

PR #4296, opened 2024-12-09
Disable bool cast on handles

ktbarrett:
In 1.X some handles had __bool__ casts defined as `bool(handle.value)`. This behavior is now changed and can cause silent failures. This prevents those suprises. Users should explicitly use `handle is not None` check if they wish to check to see if a handle object is present in a variable or not; and should use `bool(handle.value)` checks for the 1.X behavior.

This was changed in 1.5.0, so it's been behaving differently for a while now. I'm hoping this might help people shake out issues they've been silently not realizing for a while now.


ktbarrett:
@cmarqu Not sure. Probably?

cmarqu (approved):

PR #4299, opened 2024-12-12
VHPI: Support accessing 9-state values

imphil:
Support, and test, accessing 9-state values (`std_ulogic`) when VHPI is
used.

Document a bit more clearly what VPI on GHDL means for the end-user.


ktbarrett (approved):
LGTM.

ktbarrett:
Another question is whether the FLI supports all 9 states.

marlonjames (approved):

imphil:
On FLI: Yes, the tests show that using FLI to access 9-value signals works.

PR #4301, opened 2024-12-14
Transition `ReadWrite`, `ReadOnly`, and `NextTimeStep` to regular Singletons

ktbarrett:
Timing Triggers like ReadWrite, ReadOnly, NextTimeStep can have full program lifetime, so there were transitioned to typical singletons. A singleton class decorator was introduced to make this easier.

### DIFF
--- a/src/cocotb/triggers.py
+++ b/src/cocotb/triggers.py
@@ -55,7 +55,11 @@ from cocotb import simulator
 from cocotb._deprecation import deprecated
 from cocotb._outcomes import Error, Outcome, Value
 from cocotb._py_compat import cached_property
-from cocotb._utils import ParameterizedSingletonMetaclass, remove_traceback_frames
+from cocotb._utils import (
+    ParameterizedSingletonMetaclass,
+    remove_traceback_frames,
+    singleton,
+)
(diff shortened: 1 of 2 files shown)


ktbarrett:
Need to add documentation for new functions. Also some comment to explain how the types work now (private constructor, `__new__` delegates to cached attribute, `__init__` must be empty to prevent double initialization).

src/cocotb/triggers.py
```diff
@@ -294,13 +283,10 @@ def __repr__(self) -> str:
         return f"{type(self).__qualname__}()"
 
 
-class ReadWrite(GPITrigger, metaclass=_ParameterizedSingletonGPITriggerMetaclass):
+@singleton
+class ReadWrite(GPITrigger):
     """Fires when the read-write portion of the simulation cycles is reached."""
```
cmarqu, line 288:
```suggestion
    """Fires when the read-write simulation phase is reached."""
```
(while we are here...)

ktbarrett:
I removed the change `RisingEdge`, `FallingEdge`, `Edge`, and `Join` as it was failing Riviera and I didn't want to wait to figure that out. Also it's less of a diff for The Refactor PR.

marlonjames (approved):

PR #4302, opened 2024-12-16
Cache calls to `get_sim_time`

ktbarrett:
This function is called on every logging call and Timer creation. It only changes value whenever control is returned to the simulator. So we cache the call and invalidate whenever a GPITrigger fires.

Also tidied up some test in test_timing_triggers because they were failing. Tests no longer assume fs precision and a couple out of date tests below are cleaned up.

### DIFF
--- a/src/cocotb/_scheduler.py
+++ b/src/cocotb/_scheduler.py
@@ -55,6 +55,7 @@ from cocotb.triggers import (
     Trigger,
     _Join,
 )
+from cocotb.utils import _get_sim_time
 
 # Sadly the Python standard logging module is very slow so it's better not to
 # make any calls by testing a boolean flag first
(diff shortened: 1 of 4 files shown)


imphil:
> Also tidied up some test in test_timing_triggers because they were failing.

What do you mean by "because they were failing"? If the caching works as advertised, the tests would need no changes at all?

imphil (approved):
LGTM, thanks! (Have you measured any performance impact of this? It looks like a good addition either way, just wondering.)

cmarqu:
Setting the "performance" label in hope it turns out that way :)
If so, a newsfragment would allow us to advertise it.

cmarqu (approved):

PR #4308, opened 2024-12-22
Improve code coverage

ktbarrett:
Improve coverage by initializing library coverage collection at the cocotb entry point.


ktbarrett:
Seems what we really need is for `COCOTB_LIBRARY_COVERAGE` to be set *before* the cocotb module is loaded. This could be achieved by creating a second GPI_ENTRY_POINT. But we can't just support that because the entry point module is also used to source `_log_from_c`, `_filter_from_c`, and `_sim_event`. So we would need to transition these to register-style interfaces first.

This depends on that depends on this other thing depends on the planets aligning.

Issue #4311, opened 2024-12-22
VCS Illegal VPI Handles 79/83


ipburbank:
Cocotb log contains dozens of these prints before testing actually starts. Tests seem to run successfully.

```
     0.00ns ERROR    gpi                                VPI error
     0.00ns ERROR    gpi                                
                                                        Error-[VPI-ITPTVOT] Illegal VPI tag
                                                        Unknown source file, unknown line number
                                                          In VPI call vpi_handle, illegal type '79' was passed in for vpi object of 
                                                          type 'vpiLogicVar'
                                                          Please fix the VPI call in your PLI program by providing the legal type.
     0.00ns ERROR    gpi                                VPI error
     0.00ns ERROR    gpi                                
                                                        Error-[VPI-ITPTVOT] Illegal VPI tag
                                                        Unknown source file, unknown line number
                                                          In VPI call vpi_handle, illegal type '83' was passed in for vpi object of 
                                                          type 'vpiLogicVar'
                                                          Please fix the VPI call in your PLI program by providing the legal type.
     0.00ns WARNING  gpi                                VPI: Cannot discover range bounds, guessing based on elements
```

## Versions
Cocotb:
5d0f4f7677858c611a3de212d6c978fdd939b925

Ubuntu:

% lsb_release -a
LSB Version:	core-11.1.0ubuntu4-noarch:security-11.1.0ubuntu4-noarch
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.5 LTS
Release:	22.04
Codename:	jammy

VCS: vcs/U-2023.03

Python: Python 3.6 (using the python bundled with VCS)

ktbarrett:
What is the type of the variable? Is is a scalar logic object, or a logic array? The former would be invalid and there are plans in cocotb to split the LogicObject into a scalar and Array type which should prevent the issue. If its complaining about logic arrays that's a VCS issue.

ipburbank:
Minimal example:
```
module my_module (
    input logic clk
)

endmodule
```


and

```
async def test_my_moduledut)
    c = Clock(dut.clk, 10, "ns")
    await cocotb.start(c.start())
```

ktbarrett:
Ah yeah. The issue is simply that it's trying to get the range of a scalar. I'll keep this around. I plan on working on the feature soon. I have a few things in flight.

ipburbank:
Confirmed locally on my project that the warnings are gone. Thanks @ktbarrett!

PR #4318, opened 2024-12-24
Separate `LogicObject` into scalar and array types

### LINKED ISSUE #4315: Split `LogicObject` into `LogicObject` and `LogicArrayObject`
These should be split up as there are several operations on scalar logics that don't make sense (such as getting the range) and can cause issues, as in #4311. Also, only single bit logic objects can support `RisingEdge` and `FallingEdge`.

ktbarrett:
xref #4311. Closes #4315.

First GPI_NET and GPI_REGISTER were combined into GPI_LOGIC. There wasn't much utility in having them be separate types. There is no functional difference to the cocotb user. 

Next GPI_LOGIC was split into GPI_LOGIC and GPI_LOGIC_ARRAY. One for scalar logic objects, the other for packed arrays of logic.

LogicObject was then split into LogicObject and LogicArrayObject. The former had its len and range removed. RisingEdge and FallingEdge were updated to only work on the scalar LogicObject type.

This should only break users who were explicitly using len and range on scalar objects (causes errors on a couple sims), or explicitly testing against LogicObject (rare).

## TODO
- [x] Actually separate the types
- [x] Update RisingEdge and FallingEdge
- [x] Update tests
- [x] Newsfragment
- [x] Update documentation

### DIFF
--- a/docs/source/newsfragments/3980.removal.rst
+++ b/docs/source/newsfragments/3980.removal.rst
@@ -1,0 +1,1 @@
+Deprecated the undocumented ``data`` attribute on :class:`~cocotb.triggers.Event` and corresponding argument to :meth:`.Event.set`.
--- a/docs/source/newsfragments/4079.removal.rst
+++ b/docs/source/newsfragments/4079.removal.rst
@@ -1,1 +1,0 @@
-Removed the undocumented *data* attribute on :class:`~cocotb.triggers.Event` and on the corresponding :meth:`.Event.set`.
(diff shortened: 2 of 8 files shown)


ktbarrett:
Icarus is currently returning `GPI_LOGIC` (scalar) for packed array elements of unpacked arrays. One possible solution is to see if the number of elements is >1.

imphil (approved):
Thanks!

PR #4330, opened 2025-01-03
Allow `KeyboardInterrupt` and `SystemExit` to bubble up

### LINKED ISSUE #3880: Don't handle `BaseException`
There are several places in the cocotb codebase that handle `BaseException`s and do stuff to them/squash them ("manage" them). This is not ideal. Instead we should be catching only `Exception` and select `BaseException`s that are safe to catch, for example: `AssertionError` and `pytest.Failed`.

Places where this unsafe handling occurs:
https://github.com/cocotb/cocotb/blob/102625f3b45e501bb498ce46289c5f5c2c61c3ae/src/cocotb/_scheduler.py#L537
https://github.com/cocotb/cocotb/blob/102625f3b45e501bb498ce46289c5f5c2c61c3ae/src/cocotb/_outcomes.py#L18
https://github.com/cocotb/cocotb/blob/102625f3b45e501bb498ce46289c5f5c2c61c3ae/src/cocotb/task.py#L137
https://github.com/cocotb/cocotb/blob/102625f3b45e501bb498ce46289c5f5c2c61c3ae/src/cocotb/triggers.py#L809

I suggest we move the `pytest.Failed` logic from `cocotb.regression` into `cocotb._conf` and setup a variable which will include all `Exception` types that should be caught in these circumstances. 

```python
managed_exceptions = (Exception, Failed, AssertionError, TestSuccess, SimFailure)
```

Closing #3787 and #3788 would help shrink that list and not create some potentially weird cyclic reference between `cocotb._conf` and `cocotb.result`.

ktbarrett:
Fixes #3880 in an alternative way. Currently these are "handled" as part of the Task and Test classes, but in both cases what's meant is "die now", not "handle me and maybe keep running tests". asyncio follows the same approach by special casing these two exception types.

- [x] Add tests
- [x] Add newsfrag

### DIFF
--- a/src/cocotb/_scheduler.py
+++ b/src/cocotb/_scheduler.py
@@ -446,6 +446,10 @@ class Scheduler:
             # This function runs in the scheduler thread
             try:
                 _outcome = _outcomes.Value(await task)
+            except (KeyboardInterrupt, SystemExit):
+                # Allow these to bubble up to the execution root to fail the sim immediately.
+                # This follows asyncio's behavior.
+                raise
             except BaseException as e:
                 _outcome = _outcomes.Error(e)
             event.outcome = _outcome
--- a/src/cocotb/regression.py
+++ b/src/cocotb/regression.py
@@ -512,8 +512,6 @@ class RegressionManager:
             outcome = self._test_task._outcome
         try:
             outcome.get()
-        except (KeyboardInterrupt, SystemExit):
-            raise
         except BaseException as e:
             result = remove_traceback_frames(e, ["_test_complete", "get"])
         else:
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -157,6 +157,10 @@ class Task(Generic[ResultType]):
         except StopIteration as e:
             self._outcome = Value(e.value)
             self._state = Task._State.FINISHED
+        except (KeyboardInterrupt, SystemExit):
+            # Allow these to bubble up to the execution root to fail the sim immediately.
+            # This follows asyncio's behavior.
+            raise
         except BaseException as e:
             self._outcome = Error(remove_traceback_frames(e, ["_advance", "send"]))
             self._state = Task._State.FINISHED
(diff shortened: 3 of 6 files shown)


src/cocotb/_scheduler.py
```diff
@@ -446,6 +446,8 @@ async def wrapper():
             # This function runs in the scheduler thread
             try:
                 _outcome = _outcomes.Value(await task)
+            except (KeyboardInterrupt, SystemExit):
```
cmarqu, line 449: Maybe add a comment to this branch that this is also what asyncio does?

src/cocotb/regression.py
```diff
@@ -512,8 +512,6 @@ def _test_complete(self) -> None:
             outcome = self._test_task._outcome
         try:
             outcome.get()
-        except (KeyboardInterrupt, SystemExit):
-            raise
```
ktbarrett, line 516: These exception types can never be an outcome now.

src/cocotb/_scheduler.py
```diff
@@ -446,6 +446,10 @@ async def wrapper():
...
+                # Allow these to bubble up to the execution root to fail the sim immediately.
+                # This follow's asyncio's behvaior.
```
cmarqu, line 451:
```suggestion
                # This follows asyncio's behvaior.
```

src/cocotb/task.py
```diff
@@ -157,6 +157,10 @@ def _advance(self, outcome: Outcome) -> Any:
         except StopIteration as e:
             self._outcome = Value(e.value)
             self._state = Task._State.FINISHED
+        except (KeyboardInterrupt, SystemExit):
+            # Allow these to bubble up to the execution root to fail the sim immediately.
+            # This follow's asyncio's behvaior.
```
cmarqu, line 162:
```suggestion
            # This follows asyncio's behvaior.
```

cmarqu (approved):

src/cocotb/task.py
ktbarrett: I surprised you missed `behvaoir`! But I'll fix this.

Issue #4333, opened 2025-01-03
combine_results.py does not raise and exit 1 if there has been a dependencay failure


davekeeshan:

ktbarrett:
Dependency failure?

davekeeshan:
> Dependency failure?

Apologies my computer posted before I had completed the report, should be in place now

davekeeshan:
OK the PR I opened has errors every where, #4334, which suggests that `results.xml` is not the best thing to use, maybe something else is appropriate. Hopefully the scope of the error is clear from the passing/failing run above.

I have made that change to my own flow, it highlighted some further errors with some of my work and helped me fix them.

https://github.com/daxzio/cocotbext-jtag/actions/runs/12600842340/job/35120736730

themperek:
> Here is an example of a passing failing test:
> 
> https://github.com/daxzio/cocotbext-jtag/actions/runs/12600538398/job/35119795921

Makefile failed (https://github.com/daxzio/cocotbext-jtag/actions/runs/12600538398/job/35119795921#step:9:79) . Should return error and fail the test. Why does  it pass? Looks like something with your setup?

davekeeshan:
This was discussed over on gitter a while back, but because python runs wholly inside the simulator, an exit event inside python is not guaranteed to be passed out of the simulator to be picked up by the CI. 

It is my understanding that is why `combine_results.py` exists, to collate and return the passes and escalate the fails. In its current incarnation it seems only sensitive to the report generated at the end and is not reporting catastrophic fails, like dependancy conflicts, which can occur on initial setup (trying to cover multiple versions of python) or cocotb migration where fundamental features change. For me at least this is a major reason to use a CI. 

My setup is visible, if you think something is missing I of course would love to get it right and fixed. My fix, at least, highlights and solves the problem I have, so in some ways I am safe to move on, however I felt this might be something the community might also be seeing.

davekeeshan:
> > Here is an example of a passing failing test:
> > https://github.com/daxzio/cocotbext-jtag/actions/runs/12600538398/job/35119795921
> 
> Makefile failed (https://github.com/daxzio/cocotbext-jtag/actions/runs/12600538398/job/35119795921#step:9:79) . Should return error and fail the test. Why does it pass? Looks like something with your setup?

OK I didn't undestand this correctly at the time (following up bugs while shopping is a no-no).

What you are saying is that an exit event is coming from cocotb, because of the failure.  However because I am doing this:

[...]

PR #4334, opened 2025-01-03
Highlight catastrophic failure on CI

davekeeshan:
CI runs that have a catastrophic failure due to a dependency conflict and do not run are not being reported as failure, #4333.

Some modules are failure due the the fundamental changes between cocotb 1.9 and 2.0

In CI, the ability to catch these types of bugs early is really handy and one of the main reasons to add this.

Adding a test for the presence of the results.xml file and just exit 1 with failure.

### DIFF
--- a/src/cocotb_tools/combine_results.py
+++ b/src/cocotb_tools/combine_results.py
@@ -59,6 +59,11 @@ def main() -> int:
     args = parser.parse_args()
     rc = 0
 
+    if not os.path.isfile("results.xml"):
+        print(f"ERROR: results.xml was not written by the simulation!")
+        rc = 1
+        return rc
+
     result = ET.Element("testsuites", name=args.output_testsuites_name)
 
     input_pattern = re.compile(args.input_filename)


ktbarrett:
The `directories` argument is intended to allow the user to pass multiple directories. Not having a `result.xml` in the current directory is not a failure. If what you really want is that there should be at least one results.xml to process, that's as simple as adding a flag like `found_results` to the loop that starts out False, is set to True whenever a `results.xml` is processed, then after the loop is checked to ensure at least one results.xml was found.

Or do you intend it to ensure that there is at least one `results.xml` per listed directory?

davekeeshan:
> The `directories` argument is intended to allow the user to pass multiple directories. Not having a `result.xml` in the current directory is not a failure. If what you really want is that there should be at least one results.xml to process, that's as simple as adding a flag like `found_results` to the loop that starts out False, is set to True whenever a results.xml is processed, then after the loop is checked to ensure at least one results.xml was found.
> 
> Or do you intend it to ensure that there is at least one `results.xml` per listed directory?

Please see the issue to understand what i am trying to do and why.  #4333

I don't care how it is done, it is the result I want, though I am too far down the chain to understand all the places this touches, so if you have a better way, then that is probably correct.

themperek:
Every test run (in case of Makefiles) check for presents of `results.xml` file. 
Example: https://github.com/cocotb/cocotb/blob/5237f9de3620aaf64e36d2e76960b9f7121bea96/src/cocotb_tools/makefiles/simulators/Makefile.icarus#L110)
Is this not enough?

Python runner also check if any test failed.

davekeeshan:
This is not the error I thought it was, there may be something to be learned from this, but it is not the fix, sorry.

PR #4337, opened 2025-01-05
Check results file for failures in Makefile flow

### LINKED ISSUE #4336: Check if any test failed in results file in Makefile flow
In Makefile flow at present only check if the results file exist.

It should be easy to add check if the any test fail.  

Same (based on) to what is done in python runner: 
https://github.com/cocotb/cocotb/blob/5237f9de3620aaf64e36d2e76960b9f7121bea96/src/cocotb_tools/runner.py#L519-L546

ktbarrett:
Closes #4336.


src/cocotb_tools/makefiles/Makefile.inc
```diff
@@ -72,6 +72,11 @@ define check_for_results_file
     @test -f $(COCOTB_RESULTS_FILE) || (echo "ERROR: $(COCOTB_RESULTS_FILE) was not written by the simulation!" >&2 && exit 1)
 endef
```
themperek, line 73: `check_for_results_file` should not be needed anymore?

themperek (approved):

src/cocotb_tools/makefiles/Makefile.inc
themperek: For compatibility reasons maybe better reuse `check_for_results_file`. In other case should probably keep `check_for_results_file` as is (it will not be covered).

ktbarrett: Yeah I was going to leave it for backwards compatibility, who knows what users have done. Perhaps I should also add a deprecation warning.

themperek: I think is ok,

PR #4338, opened 2025-01-05
Random fixes

ktbarrett:
A collection of random small fixes for issues I saw while perusing the Makefiles.

Most important of these is the change to use `$(PYTHON_BIN) -m cocotb_tools.config` over `cocotb-config`, which should reduce the number of issues users have with entry points not being on their PATH by default. There have been a couple issues opened wrt to that.

### DIFF
--- a/src/cocotb_tools/makefiles/Makefile.inc
+++ b/src/cocotb_tools/makefiles/Makefile.inc
@@ -56,13 +56,10 @@ PYTHON_BIN_RES := $(shell cocotb-config --python-bin 2>>/dev/null; echo $$?)
 PYTHON_BIN_RC := $(lastword $(PYTHON_BIN_RES))
 PYTHON_BIN := $(strip $(subst $(PYTHON_BIN_RC)QQQQ,,$(PYTHON_BIN_RES)QQQQ))
 ifneq ($(PYTHON_BIN_RC),0)
-  PYTHON_BIN := python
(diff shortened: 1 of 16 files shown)


themperek:
As far as I remember using `cocotb-config` was there to prevent using wrong python. executable. This PR solves some issue but potentially reintroduce other. No opinion what is better. https://github.com/cocotb/cocotb/pull/1574

src/cocotb_tools/makefiles/Makefile.inc
```diff
@@ -51,10 +51,13 @@ endif
 export OS
 
 # this ensures we use the same python as the one cocotb was installed into
-PYTHON_BIN ?= $(shell cocotb-config --python-bin)
-export PYGPI_PYTHON_BIN := $(PYTHON_BIN)
+PYTHON_BIN ?= python
+export PYGPI_PYTHON_BIN := $(shell $(PYTHON_BIN) -m cocotb_tools.config --python-bin)
 
-include $(shell cocotb-config --makefiles)/Makefile.deprecations
+# Directory containing the cocotb Makfiles
```
cmarqu, line 57:
```suggestion
# Directory containing the cocotb Makefiles
```

cmarqu:
IMO the Makefile change is worth a newsfragment if only to tell previously affected users that they can expect things working now.

themperek (approved):
Adding newsfragment is good idea.

ktbarrett:
Just need a check on the newsfrag.

cmarqu (approved):

PR #4339, opened 2025-01-05
Typing fixes

ktbarrett:
Fixed typing in several modules. A few more typing fixes will come in with the big refactor and cocotb should pass mypy after that.


src/cocotb/logging.py
```diff
@@ -277,6 +283,6 @@ def _log_from_c(logger_name, level, filename, lineno, msg, function_name):
     logger = logging.getLogger(logger_name)
     if logger.isEnabledFor(level):
         record = logger.makeRecord(
-            logger.name, level, filename, lineno, msg, None, None, function_name
+            logger.name, level, filename, lineno, msg, (), None, function_name
```
cmarqu, line 286: Since I just looked it up: this is `args` of https://docs.python.org/3/library/logging.html#logging.LogRecord which doesn't allow `None`.

cmarqu (approved):

marlonjames (approved):

Issue #2344, opened 2021-01-13
[Riviera-PRO] vpi_register_cb returns reused callback handle


marlonjames:
Cocotb version: master
Simulators: Riviera-PRO 2019.10+ 64-bit Linux, Active-HDL 11.1 32-bit Windows 10

VSimSA is re-using callbacks for `cbValueChange`.

If we have a `cbValueChange` callback handle for a signal such as `dut.clk`, say for a `RisingEdge` trigger.
If a `FallingEdge` trigger for the same `dut.clk` signal is primed, then it will register for a new `cbValueChange` callback on the same signal.
Riviera gives us back the same handle we already have for `RisingEdge` (but doesn't update the callback data for the new `FallingEdge` object).
When the original `RisingEdge` trigger is cleaned up, which calls `vpi_remove_cb` on the shared handle, the callback handle is set to NULL, and the state is set to `GPI_FREE`.
On the falling edge of `dut.clk`, the callback fires again but with the original `RisingEdge` callback object that is `GPI_FREE` so nothing happens, it just returns to the simulator.

I'm pretty sure this falls afoul of the VPI standard. I found this, which is in **38.36.1.1** in context of callbacks on statements:

> Multiple calls to `vpi_register_cb()` with the same data shall result in multiple callbacks.

_Based on https://github.com/cocotb/cocotb/pull/2338#issuecomment-759778275_

imphil:
Filed a support case with Aldec, case number SPT81300.

imphil:
The 2023.04 release notes mention this issue is resolved.

@marlonjames Is there something we want to do as result of that on the cocotb side?

marlonjames:
`test_both_edge_triggers` should now pass with riviera 2023.04:
https://github.com/cocotb/cocotb/blob/c34c2bc5cf2408c0c5c386d5d9c139ecf6453889/tests/test_cases/test_cocotb/test_edge_triggers.py#L232-L244

imphil:
I confirmed that this issue is fixed with Riviera 2023.10. https://github.com/cocotb/cocotb/pull/3471 removes the test exclusion.

PR #4349, opened 2025-01-08
Remove VHPI callbacks once cleaned up to prevent leaking

### LINKED ISSUE #4346: Simulation slow down with nvc
I noticed that there is a slow down in long simulations (especially with the parametrize) with nvc. After some debug I found that the simulator finished with a large number of active handles and the number increases with the duration of the simulation. I opened an issue on the nvc github https://github.com/nickg/nvc/issues/1114 and also provided to the nvc author a reduced example that put in evidence the issue. He noticed that cocotb never uses the `vhpi_remove_cb` but only the `vhpi_disable_cb`, and this is the cause of the slow down. What's your opinion about that?

ktbarrett:
Closes #4346.

### DIFF
--- a/src/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
+++ b/src/cocotb/share/lib/vhpi/VhpiCbHdl.cpp
@@ -435,19 +435,12 @@ VhpiCbHdl::VhpiCbHdl(GpiImplInterface *impl) : GpiCbHdl(impl) {
 }
 
 int VhpiCbHdl::cleanup_callback() {
-    /* For non timer callbacks we disable rather than remove */
-    int ret = 0;
-    if (m_state == GPI_FREE) return 0;
+    if (m_state == GPI_FREE) return 1;
 
-    vhpiStateT cbState =
-        (vhpiStateT)vhpi_get(vhpiStateP, get_handle<vhpiHandleT>());
-    if (vhpiEnable == cbState) {
-        ret = vhpi_disable_cb(get_handle<vhpiHandleT>());
-        m_state = GPI_FREE;
-    }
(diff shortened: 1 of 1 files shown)


ktbarrett:
Welp. This causes it to hang.

marlonjames:
I think disable is good in case it's re-armed, but maybe it needs to remove it when it's no longer needed.

ktbarrett:
> I think disable is good in case it's re-armed, but maybe it needs to remove it when it's no longer needed.

There is no way for us to re-enable once it's disabled. We would have to cache the disabled callback handle someplace to get it again and the whole point of #4090 was to stop doing that to support multiple GPI callbacks to support the C++ clock.

Ideally we would have an arrangement like the new Trigger in #4280, but that would be a *huge* refactor. We would have to introduce something like Trigger into the GPI to multiplex multiple registrations and only maintain one underlying VPI callback. Given the current architecture, making GPI and VPI/VHPI/FLI triggers one-to-one and removing them is the easiest way forward unless it simply won't work.

ktbarrett:
I confirmed this fixes the slowdown issue on Riviera and NVC.

On the dev machine running Riviera the example in #4379 went from 14.79s to 2.55s.
On my personal machine running NVC the example in #4379 went from 6.59s to 0.32s.

Also the slowdown getting worse as the simulation progresses is no longer present, every 10% of completion in the example test progresses evenly.

cmarqu (approved):

marlonjames:
FLI keeps timer callbacks around and re-uses them.
I do think it's better to have one VHPI callback registered and keep a separate list of GPI user callbacks associated with it, but as long as there aren't any problems registering multiple callbacks on the same signal like Riviera had with VPI in #2344, this is fine.

PR #4363, opened 2025-01-14
Document how to get libstdc++-static on RHEL9

maxime-lr:
Minor PR but it would have helped me if it was in the doc ;)

### DIFF
--- a/docs/source/install_devel.rst
+++ b/docs/source/install_devel.rst
@@ -47,7 +47,15 @@ The installation instructions vary depending on your operating system:
 
    .. group-tab:: Linux - Red Hat
 
-      In a terminal, run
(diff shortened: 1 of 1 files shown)


cmarqu:
Thanks for the PR!

I'd rather see this as a section in our existing "Troubleshooting" section though, ideally with an error message that you get when you try without having that repo enabled.

maxime-lr:
@cmarqu As you wish!
I put it there as it was the devel section and there were already 2 tabs explaining how to install CocoTB on Linux (either Ubuntu or RHEL).
But if you prefer it in the troubleshooting section, I can make another commit ;)

cmarqu:
Actually, right, I missed that this is the development instructions, for which it is probably fine.
Let's see if anyone else voices opinions and merge next week if there are no objections.

cmarqu (approved):

PR #4367, opened 2025-01-16
Support building docs with Python 3.12+

ktbarrett:
`sphinxcontrib.domaintools` ~~and `sphinxcontrib.details.directive` both~~ uses the old way of declaring namespace packages with pkg_resouces. pkg_resources was removed in Python 3.12 and there is no replacement for `declare_namespace`. So I removed the contents of `sphinxcontrib/__init__.py` to make it compatible with native namespace packages and simply listed the namespace package name out in the `setup.py` (since `find_namespace_packages` doesn't work well when not in a `src/` directory layout).

~~I also had to vendor `sphinxcontrib.details.directive` in as I doubt its maintained.~~


cmarqu (approved):

ktbarrett:
On hold until @cmarqu tests out sphinx-book-theme.

ktbarrett:
Was able to drop vendoring in sphinxcontrib-details-domain due to #4370.

PR #4371, opened 2025-01-18
Changes to pyuvm ecosystem test

ktbarrett:
Should fix the issue where we are accidentally using the local `pyuvm` for imports during tests rather than the installed version.

We should probably expand this to test the latest released cocotb against the latest release pyuvm, as well as cocotb master against pyuvm master.


cmarqu:
Problem description in https://github.com/pyuvm/pyuvm/issues/241#issuecomment-2598947003

cmarqu (approved):

PR #4373, opened 2025-01-18
Reword anachronistic sentence

cmarqu:


ktbarrett (approved):
Can't say I love the way the sentences are broken up over multiple lines, but whatever.

cmarqu:
> Can't say I love the way the sentences are broken up over multiple lines, but whatever.

That's to keep the diff small. We should adopt https://sembr.org/ in the long term.

PR #4374, opened 2025-01-18
Add more semantic markup

cmarqu:


docs/source/coroutines.rst
```diff
@@ -194,7 +194,7 @@ The second section in the above code using it would be ``await with_timeout(quie
 Determining Which Task Finishes First
 -------------------------------------
 
-:class:`~cocotb.triggers.First` can be used to determine which of multiple Tasks :meth:`~cocotb.task.Task.complete` first using the following idiom.
+:class:`~cocotb.triggers.First` can be used to determine which of multiple Tasks :meth:`complete <~cocotb.task.Task.complete>` first using the following idiom.
```
ktbarrett, line 197: Strange this doesn't work like classes or methods?
```suggestion
:class:`~cocotb.triggers.First` can be used to determine which of multiple Tasks :meth:`complete <cocotb.task.Task.complete>` first using the following idiom.
```

ktbarrett (approved):
I do worry that having all these things linked can make a page look busy. What do you think about Wikipedia's strategy of only linking the first occurrence of a a term?

docs/source/coroutines.rst
cmarqu: If you mean that `~` means to strip all elements from the left except the last - it does that, but it ends up with `complete()`, and this change removes the `()`.

cmarqu:
> What do you think about Wikipedia's strategy of only linking the first occurrence of a a term?

I have done that now.

PR #4375, opened 2025-01-19
Depend on Sphinx ~= 7.3 for the 'versionremoved' directive

cmarqu:
See https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionremoved


cmarqu:
Superseded by #4376

Issue #4379, opened 2025-01-20
Significant performance drop after commit f930bc7


olagrottvik:
I traced back a significant loss of performance to commit f930bc7.

Here's a minimal example to reproduce:

`deserializer_th.vhd`:
```
library ieee;
use ieee.std_logic_1164.all;
use ieee.numeric_std.all;

entity deserializer_th is
end entity deserializer_th;

architecture behavior of deserializer_th is

  -- Signals to connect to UUT
  signal clk             : std_logic := '0';


begin

end architecture behavior;
```

`Makefile`:
```
# set simulator
SIM ?= riviera

TOPLEVEL_LANG ?= vhdl

COMPILE_ARGS = "-2008"

# Generate the list of VHDL sources required for the simulation
VHDL_SOURCES += deserializer_th.vhd

# TOPLEVEL is the name of the toplevel module in your Verilog or VHDL file
TOPLEVEL = deserializer_th

# MODULE is the basename of the Python test file
MODULE = deserializer_tb

# include cocotb's make rules to take care of the simulator setup
include $(shell cocotb-config --makefiles)/Makefile.sim
```

`deserializer_tb.py`:
```
import random
import cocotb
from cocotb.clock import Clock
from cocotb.triggers import Timer, FallingEdge


# Constants
OVERSAMPLING_RATE = 4  # Must match g_oversampling_rate in VHDL
DATA_RATE_MBPS = 160  # Data rate in Mbps

async def capture_parallel_output(dut):
    while True:
        await FallingEdge(dut.clk)

async def run_test(dut, clk_mhz, num_bits=5000):

    clk_period_ns = round(1000 / clk_mhz, 3)

    # Start the sampling clock
    sampling_clock = Clock(dut.clk, clk_period_ns, units="ns")
    cocotb.start_soon(sampling_clock.start())
    await FallingEdge(dut.clk)

    # Data period for 160 Mbps
    data_period_ns = round(1000 / DATA_RATE_MBPS, 3)

    incoming_data = [random.randint(0, 1) for _ in range(num_bits)]

    cocotb.start_soon(
        capture_parallel_output(dut)
    )

    for i, bit in enumerate(incoming_data):
        if i % (num_bits // 10) == 0:
            dut._log.info(f"Transmitting data: {i / num_bits:.0%} complete")
        await Timer(data_period_ns, units="ns")


@cocotb.test(skip=False)
async def test_deserializer_nominal(dut):
[...]

PR #4382, opened 2025-01-20
Move Edge Trigger lifetimes to Handles

### LINKED ISSUE #3855: Add value change triggers as methods to handles
I'm not entirely sure why triggers like `RisingEdge`, `FallingEdge`, and `Edge` were made as standalone functions, but since they are parameterizable and only support certain objects, it makes more sense if they are associated only with the objects that support them by making them a method.

```python
await dut.handle.rising_edge()
```

This will also allow the ability to not have to check the argument type and allow bad behavior.

In lock-step we should deprecate the free-function calls.

ktbarrett:
Closes #3855. ~~May help with #4379.~~ May help performance by not deleting and remaking Trigger objects constantly due to the weakref cache. Also renames `Edge` to `ValueChange`. Supersedes #3365.

### DIFF
--- a/src/cocotb/triggers.py
+++ b/src/cocotb/triggers.py
@@ -33,7 +33,6 @@ from abc import abstractmethod
 from decimal import Decimal
 from fractions import Fraction
 from typing import (
-    TYPE_CHECKING,
     Any,
     AsyncContextManager,
     Awaitable,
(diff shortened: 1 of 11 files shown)


cmarqu (approved):

src/cocotb/triggers.py
```diff
@@ -405,38 +395,55 @@ class FallingEdge(_EdgeBase):
...
 
     Args:
         signal: The signal upon which to wait for a value change.
 
     Raises:
         TypeError: If the signal is not an object which can change value.
+
+    .. versionadded: 2.0
+
+        Use :attr:`signal.value_change <cocotb.handle.NonArrayValueObject.value_change>` instead.
```
ktbarrett, line 417:
```suggestion
```

PR #4383, opened 2025-01-20
Use Doxygen markup for GPI

cmarqu:

### DIFF
--- a/src/cocotb/share/include/gpi.h
+++ b/src/cocotb/share/include/gpi.h
@@ -101,46 +101,59 @@ typedef struct GpiIterator *gpi_iterator_hdl;
 extern "C" {
 #endif
 
-// Functions for controlling/querying the simulation state
+/**
+ * Functions for controlling/querying the simulation state
+ */
 
 /**
- * Returns 1 if there is a registered GPI implementation, 0 otherwise.
- *
+ * Return if there is a registered GPI implementation.
  * Useful for checking if a simulator is running.
+ *
+ * @return `1` if there is a registered GPI implementation, `0` otherwise.
  */
 GPI_EXPORT bool gpi_has_registered_impl(void);
 
-// Stop the simulator
+/**
+ * Stop the simulator.
+ */
 GPI_EXPORT void gpi_sim_end(void);
 
-// Returns simulation time as two uints. Units are default sim units
+/**
+ * Return simulation time as two uints. Units are default sim units.
+ */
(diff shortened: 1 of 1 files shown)


ktbarrett:
I think you can get rid of the section comments. We just need to properly document all the functions at some point.

marlonjames:
If you update the Doxyfile you can hide the extra enum typedefs:
```
TYPEDEF_HIDES_STRUCT   = YES
```

ktbarrett:
Looks like your rebase changed some of the functions back to their old return values.

cmarqu:
Oops, will fix later.

src/cocotb/share/include/gpi.h
```diff
@@ -264,13 +307,17 @@ gpi_register_nexttime_callback(int (*gpi_function)(void *), void *gpi_cb_data);
 GPI_EXPORT gpi_cb_hdl
 gpi_register_readwrite_callback(int (*gpi_function)(void *), void *gpi_cb_data);
 
-// Calling convention is that 0 = success and negative numbers a failure
-// For implementers of GPI the provided macro GPI_RET(x) is provided
+/**
+ * Calling convention is that `0` = success and negative numbers mean failure.
+ * For implementers of GPI the macro `GPI_RET(x)` is provided.
+ */
```
ktbarrett, line 313: This comment is nonsense.
```suggestion
```

src/cocotb/share/include/gpi.h
```diff
@@ -189,56 +203,83 @@ typedef enum gpi_edge_e {
...
 GPI_EXPORT gpi_range_dir gpi_get_range_dir(gpi_sim_hdl gpi_sim_hdl);
 
 // Functions for querying the properties of a handle
-// Caller responsible for freeing the returned string.
-// This is all slightly verbose but it saves having to enumerate various value
-// types We only care about a limited subset of values.
+
+/**
+ * Caller responsible for freeing the returned string.
```
ktbarrett, line 248: This is *not* true.
```suggestion
```

ktbarrett (approved):
Considering the whole point is Doxygen-ifying this file, I'm not going to be super picky. A follow on should properly document this API.

ktbarrett:
precommit.ci autofix

cmarqu:
> pre-commit.ci autofix

Ah yes, my local setup doesn't like the `clang-format` in pre-commit, so I always have to use `--no-verify` for those commits (or modify the config, which it will then complain about...).

PR #3306, opened 2023-05-14
Support Questa qrun/QIS flow

imphil:
Support the qrun/QIS flow in Questa in the Makefile flow

v2 from https://github.com/cocotb/cocotb/pull/3274:
- Default to the new flow from Questa 2023.1 on (the first version which includes Visualizer and hence is guaranteed to work with the flow)
- Give users the option to choose the old or the new flow by using SIM=questa-qrun or SIM=questa-compat.
- More documentation updates.
- Implement the Questa version detection in Python to avoid adding a dependency to awk and sed.

@AbdouTharwat I kept the original commit attributed to you, since you did the majority of the work. Thank you for it! Can you please have a look and give it a try?

Patches:
- Support the Questa qrun flow
- Questa: Remove unnecessary FLI library check
- Add minimal documentation for the custom qrun flow

PR #4388, opened 2025-01-23
Pass voptargs to Questa to improve simulation speed

### LINKED ISSUE #4364: Cocotb performance is very poor when using Questa/ModelSim
Hi Guys,

I have been an extensive user of cocotb over the last few years. I typically don't have issues with cocotb performance as most the tests I run are relatively short in terms of simulation time.
But I noticed that when using cocotb together with ModelSim/Questasim, it becomes extremely slow for longer simulation times. This occurs when we start interrupting the simulator quite often.

What I notice is that the longer the simulation time, the slower the simulation gets in terms of sim time vs real time. The curve is pretty much exponential and for long periods, Questa becomes almost unusable.

Here is my setup:

- I am currently using the latest development branch of cocotb (because I wanted to update to make sure my issue was not solved  in the most recent version). So my current version is `2.0.0.dev0+cf9d4c43`. But previously I also tested version `1.9.2`.
- my OS is archlinux, 64-bit
- Questa Prime 2024.1_1
- Python 3.12.7 (installed from system)

I made a very simple example project to illustrate the issue at hand. There is one source file called `prbs.v` and one cocotb testbench called `prbs_tb.py`.
This comes with a `Makefile` for which I have tried 2 different calls to compare Icarus Verilog with Questa:

- icarus: `make SIM=icarus`
- questa: `make SIM=questa`

I intentionally made the test duration relatively long, 2 milliseconds to illustrate the problem. The results are very surprising:

- icarus verilog; test time is below 5 seconds
- questa: test time is around 150 seconds

In the testbench `prbs_tb.py`, there is a function (`loop_checker`) that interrupts the simulator for every rising edge of the clock. This is the culprit. Although there are some optimizations to be done to avoid performing a check at every clock cycle, this should not be so slow when using Modelsim/Questa.

Here is the compressed project. Does someone know what is going wrong? Thanks for your help.

[questa_slow.zip](https://github.com/user-attachments/files/18414048/questa_slow.zip)

ktbarrett:
Closes #4364.

### DIFF
--- a/src/cocotb_tools/makefiles/simulators/Makefile.questa
+++ b/src/cocotb_tools/makefiles/simulators/Makefile.questa
@@ -61,11 +61,11 @@ endif
 # below allows for maintaining legacy syntax as well as enables using cross-simulator vars COMPILE_ARGS/SIM_ARGS
 VLOG_ARGS += $(COMPILE_ARGS)
 VCOM_ARGS += $(COMPILE_ARGS)
-VSIM_ARGS += $(SIM_ARGS)
+VSIM_ARGS += $(SIM_ARGS) -voptargs=-access=rw+/.
 
 ifeq ($(GUI),1)
     CMD += -gui
-    VSIM_ARGS += -onfinish stop
+    VSIM_ARGS += -onfinish stop -qwavedb=+signal+memory+assertion+cell+class
 else
     CMD += -c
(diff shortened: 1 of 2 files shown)


ktbarrett:
Seems like the same issue with #3306 where turning on the optimizer affects access. I'll abandon this in favor of getting that PR up and running.

PR #4392, opened 2025-01-27
Refactor GPI callbacks

ktbarrett:
I had about 0 trust in the previous GPI callback handling scheme. I figured we were leaking a lot of objects and at least in one case we absolutely were (similar issue to #4346, but in the FLI), so I decided to refactor the whole thing. I'm not perfectly happy with the way it is right now, but it does solve all my concerns wrt leaks.

## New callback interface

This was accomplished by moving to a new interface for GPI callbacks.
- `GpiCbHdl::remove()` is called when the user wants to stop a callback from being called before it fires. It is in charge of disabling the underlying implementation callback and freeing the `GpiCbHdl` object.
- `GpiCbHdl::arm()` is essentially a rename of `GpiCbHdl::arm_callback()` it's a post-constructor init that must be separate because it's a virtual function.
- `GpiCbHdl::run()` is a rename of `GpiCbHdl::run_callback()` but most importantly it's also in charge of disabling the underlying implementation callback and freeing the `GpiCbHdl` object after the event occurs.

Because `GpiCbHdl::cleanup_callback()` didn't free the object, all prematurely removed callbacks leaked.

Previously, `handle_*_callback` was in charge of freeing objects once they fired, but I had my doubts about this always occurred because of the complex logic needed to get `arm_callback` and `cleanup_callback` working if the trigger were recurring.

The gpi_state object made the interaction of any of these methods behave differently in ways that wasn't easy to guage. This thing was removed.

## Removed `GpiValueCbHdl` base class

This thing defined `GpiValueCbHdl::run` which made an assumption about how `arm_callback` and `cleanup_callback` can be called back to back to "reset" the recurring callback, or disable it if it fired. This thing was a source of much of the complexity in the previous system and offered very little in the way of utility, so this bad idea was removed.

This class also caused multiple inheritance to be needed making everything much more complicated than it needed to be.

Every interface now has it's own logic and the new logic also avoids creating a string object just to do comparison.

## Startup refactor

I noticed that the way the startup routines and startup and shutdown callbacks were written meant that we had to keep a global singleton for the GpiImpl and failures during startup couldn't effected cause a graceful exit. This was refactored so that all startup went through a single new `main()` method on the GpiImpl. There are no more global singletons.

The startup and shutdown callbacks for each GpiImpl also hardcoded their callback behavior which interacted with these global singletons. The functionality was instead move into callback functions and the startup and shutdown callbacks were set up to take arbitrary callbacks like all the other callback types. This improves future usability if anyone wants to register additional startup or shutdown callbacks.

## `FliTimerCache` refactor

Previously there was a `FliTimerCache` class that acted as a cache for `FliTimedCbHdl` objects. It wasn't clear why this was done originally. Studying the FLI interface shows there is no way to delete an MTI Process object once created, so this was the point of the cache, to prevent creating (expensive) and then leaking MTI Process objects on each callback creation. Because the MTI Process requires passing a callback data object when created, we have to cache the whole `FliCbHdl` object to prevent corrupting that object. Instead of deleting the underlying Process object when the event occurs, we simply disable and then re-enable it when we want to reuse it.

The reason this was refactored was because we actually need this cache for *every* Process-based callback object now. We can't support singletons being stored on the objects because there may be more than one callback needed for each event. But as soon as we don't have singletons we need the cache because we can't clean up.

A new `FliProcessCbHdlCache` class was made to replace this which is templated as to support all `FliProcessCbHdl` classes.

All of these caches had to be associated with the `GpiImpl` object because all `GpiCbHdl` object must store the `GpiImpl` on them, so reusing between multiple `GpiImpl` instances is not possible (not that it will ever happen, but still).

All `FliProcessCbHdl` objects define a `release()` method to return itself to the correct cache on the `FliImpl`.

The previous cache had a hard upper limit on size and then just deleted the `FliProcessCbHdl` object leaking the Process when it was exceeded, which is just dumb. 

Finally, the previous cache used a queue meaning that the oldest object in the cache was reused first, which is bad for cache locality, so this was changed to a vector with LIFO semantics instead.

## Other cleanup

- `handle_*_callback` functions were moved from `*Impl.cpp` files to `*CbHdl.cpp` files because they were only ever used in the file.
- Includes were cleaned up (mostly automatically by my IDE)
- Removed a ton of unnecessary `extern "C"`

## TODO

- [x] Fix FLI
- [x] Fix Questa VHPI sims (`nox > Session dev_test_sim(sim='questa', toplevel_lang='vhdl', gpi_interface='vhpi') aborted: No coverage files found. Something went wrong during the test execution.`)
- [x] Fix VCS
- [x] COMMENTS
- [x] newsfrag on bug fixes?
- [ ] ~~More debugging logs~~


src/cocotb/share/include/gpi.h
```diff
@@ -264,14 +261,26 @@ gpi_register_nexttime_callback(int (*gpi_function)(void *), void *gpi_cb_data);
...
-// Calling convention is that 0 = success and negative numbers a failure
-// For implementers of GPI the provided macro GPI_RET(x) is provided
-GPI_EXPORT void gpi_deregister_callback(gpi_cb_hdl gpi_hdl);
+/** Remove callback.
+ *
+ * The callback will not fire after this function is called.
+ * The argument is no longer valid if this function succeeds.
+ *
+ * @param cb_hdl The handle to the callback to remove.
+ * @returns 0 on successful removal, 1 otherwise.
```
cmarqu, line 270:
```suggestion
 * @returns `0` on successful removal, `1` otherwise.
```

src/cocotb/share/lib/gpi/gpi_priv.h
```diff
@@ -178,38 +171,55 @@ class GPI_EXPORT GpiSignalObjHdl : public GpiObjHdl {
...
+     */
+    virtual int arm() = 0;
+
+    /** Remove the callback before it fires.
+     *
+     * This function should delete the object.
+     */
+    virtual int remove() = 0;
+
+    /** Runs the callback.
```
cmarqu, line 214:
```suggestion
    /** Run the callback.
```

ktbarrett:
Interesting. Calling `vhpi_control(vhpiFinish)` in Questa's VHPI does a clean `exit()`. I don't think this is correct per the spec.

From IEEE 1076-2019 23.5, emphasis mine.
>  If command is vhpiFinish, *after control returns to the tool from the callback function from
which the vhpi_control function was invoked*, the tool enters the termination phase (see 20.10).

ktbarrett:
I followed @themperek's suggestion from https://github.com/cocotb/cocotb/pull/4410#issuecomment-2622959293 to test the tool info to determine if VCS is loaded into the simulator or not.

docs/source/newsfragments/4392.bugfix.rst
```diff
@@ -0,0 +1 @@
+Several memory and callback leaks were fixed in the GPI. Simulations may use less memory and run slightly faster.
```
cmarqu, line 1:
```suggestion
Fixed several memory and callback leaks in the GPI. Simulations may use less memory and run slightly faster.
```

cmarqu (approved):
I am not qualified to thoroughly review the changes, but if the tests are still passing, it's probably okay even though we don't have 100% coverage.

src/cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -25,42 +25,34 @@
...
-            set_call_state(GPI_DELETE);
-            break;
-        case GPI_DELETE:
-        case GPI_FREE:
-            break;
+// Main re-entry point for callbacks from simulator
+void handle_fli_callback(void* data) {
+    gpi_to_user();
+
+    // TODO Add why?
```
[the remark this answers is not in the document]
ktbarrett: Future work. I saw the `fflush` in the previous implementation and I'm not sure why it's there. This is something that deserves a comment once we figure out why it's needed (or not).

ktbarrett:
The lack of full coverage is concerning, but it was a problem before. For example, apparently we aren't testing `NextTimeStep` anywhere. Finally testing might reveal breaks and that's a can of worms I'd rather not dive into on this already big PR.

There are lots of failure branches that could be excluded since we have no way of testing those branches.

ktbarrett:
Well... I had a copy-paste error in FLI's NextTimeStep impl. We should definitely start testing that lol.

ktbarrett:
Going to add more debugging logs if this PR isn't approved first.

ktbarrett:
I might as well refactor this to remove `GpiCbHdl::arm` and refactor this into the constructors, seeing as how big the change already is. I'll also add something like `repr` to `GpiHdl` to ease debugging.

marlonjames (approved):

PR #4396, opened 2025-01-28
Clock Improvements

ktbarrett:
Upstreaming some backwards compatible improvements to the `Clock` object.

The Clock now creates a Task and stores it on itself instead of requiring the user to do so. This makes it look more like a typical Driver object. A `stop` method was also added to stop the clock.

This required some slight changes to the scheduler to allow passing already started Tasks to be passed to `cocotb.start_soon` without causing issues. This also fixes an issue where `cocotb.create_task` followed by `cocotb.start_soon` breaks. This was because the task_done_callback was registered twice causing Test outcomes to be set more than once.


src/cocotb/_scheduler.py
```diff
@@ -419,9 +419,11 @@ def _schedule_task(
 
         It is an error to attempt to queue a task that has already been queued.
         """
-        # Don't queue the same task more than once (gh-2503)
         if task in self._scheduled_tasks:
-            raise InternalError("Task was queued more than once.")
+            return
+        for tasks in self._trigger2tasks.values():
+            if task in tasks:
+                return
```
ktbarrett, line 426: Yeah this sucks and is slow, but it will go away once #4280 is in.

docs/source/newsfragments/4396.change.rst
```diff
@@ -0,0 +1 @@
+The attribute ``frequency`` on :class:`~cocotb.clock.Clock` was removed, and the attribute :attr:`~cocotb.clock.Clock.preiod` was change to be in terms of the passed in units rather than simulation steps.
```
cmarqu, line 1:
```suggestion
The attribute ``frequency`` on :class:`~cocotb.clock.Clock` was removed, and the attribute :attr:`~cocotb.clock.Clock.period` was changed to be in terms of the passed-in units rather than simulation steps.
```

src/cocotb/clock.py
```diff
@@ -152,42 +186,84 @@ async def start(self, start_high: bool = True) -> None:
...
+            Object which can be passed to :func:`cocotb.start_soon` or ignored.
+
         .. versionchanged:: 2.0
             Removed ``cycles`` arguments for toggling for a finite amount of cyles.
-            Use ``kill()`` on the clock task instead, or implement manually.
+            Use :meth:`stop` to stop a clock from running.
+
+        .. versionchanged:: 2.0
+            Previously, this method returns a :term:`coroutine` which needed to be passed to :func:`cocotb.start_soon`,
+            but now the Clock object keeps track of it's own driver Task, so this is no necessary.
```
cmarqu, line 198:
```suggestion
            but now the Clock object keeps track of its own driver Task, so this is no longer necessary.
```

src/cocotb/clock.py
```diff
@@ -152,42 +186,84 @@ async def start(self, start_high: bool = True) -> None:
 
                 .. versionadded:: 1.3
 
+        Returns:
+            Object which can be passed to :func:`cocotb.start_soon` or ignored.
+
         .. versionchanged:: 2.0
             Removed ``cycles`` arguments for toggling for a finite amount of cyles.
-            Use ``kill()`` on the clock task instead, or implement manually.
+            Use :meth:`stop` to stop a clock from running.
```
cmarqu, line 194: Maybe mention here that a stopped clock can also be restarted.

src/cocotb/clock.py
```diff
@@ -152,42 +186,84 @@ async def start(self, start_high: bool = True) -> None:
...
+
+        self._task = cocotb.start_soon(drive())
+
+        # So if a user calls `task.kill()` on the returned task the Clock object is stopped.
+        self._task._add_done_callback(lambda _: self._cleanup())
+
+        return self._task
+
+    def stop(self) -> None:
+        """Stop driving the clock signal.
```
cmarqu, line 239: Maybe mention here that a stopped clock can also be restarted.

src/cocotb/clock.py
```diff
@@ -152,42 +186,84 @@ async def start(self, start_high: bool = True) -> None:
...
+
+        .. versionadded:: 2.0
+        """
+        if self._task is None:
+            raise RuntimeError("Stopping clock that was never started.")
```
cmarqu, line 244:
```suggestion
            raise RuntimeError("Stopping a clock that was never started.")
```

src/cocotb/triggers.py
```diff
@@ -996,48 +996,108 @@ def on_done(ret: Outcome[Any]) -> None:
...
+    r"""Finishes after *num_cycles* transitions of *signal*.
 
     :keyword:`await`\ ing this Trigger returns the ClockCycle object.
 
     Args:
         signal: The signal to monitor.
         num_cycles: The number of cycles to count.
-        rising: If ``True``, count rising edges; otherwise, count falling edges.
+        rising: If ``True``, count rising edges; if ``False``, count falling edges.
+        edge: Use the Edge Trigger type passed.
```
cmarqu, line 1007:
```suggestion
        edge: Use the Edge trigger type passed.
```
?

src/cocotb/clock.py
```diff
@@ -152,42 +186,84 @@ async def start(self, start_high: bool = True) -> None:
...
+        self,
+        num_cycles: int,
+        edge_type: Union[Type[RisingEdge], Type[FallingEdge], Type[Edge]] = RisingEdge,
+    ) -> None:
+        """Wait for a number of clock cycles."""
+        # TODO Improve implementation to use a Timer to skip most of the cycles
+        await ClockCycles(self._signal, num_cycles, edge_type)
+
+    def __repr__(self) -> str:
+        return f"{type(self).__qualname__}({self._signal._path}, {self._period}, {self._units!r})"
```
cmarqu, line 263: Maybe it would be convenient to also get the frequency that results from the period here.

src/cocotb/clock.py
ktbarrett: It think this comment is more about what changed with the cycles command and that's unrelated.

src/cocotb/triggers.py
ktbarrett: We do capitalize Task and Trigger and other keywords in other places. I will link against the edge-triggers here instead though.

cmarqu:
The refcard may also need updating in the "Generate clock" row, depending on whether and how #4400 gets in first.

src/cocotb/clock.py
cmarqu: Would implementing `__str__` be helpful for users? IIRC this is meant to be more abstract.

cmarqu (approved):

src/cocotb/clock.py
ktbarrett: I don't think so. `__repr__` was definitely what was meant here. I'll go back to the original impl.

PR #4400, opened 2025-01-28
Deprecate cocotb.start

### LINKED ISSUE #4394: Deprecate `cocotb.start`
It has some utility, but will always be equivalent to `start_soon` followed by `NullTrigger`. New users seem to not understand the distinction between start and start_soon, so we should remove the confusion by removing `cocotb.start`.

ktbarrett:
Closes #4394. The difference between `cocotb.start` and `cocotb.start_soon` confuses users, so we are deprecating `cocotb.start`. If users need the `cocotb.start` behavior, they can simply call `cocotb.start_soon` followed by `await NullTrigger()`.

### DIFF
--- a/docs/source/coroutines.rst
+++ b/docs/source/coroutines.rst
@@ -53,15 +53,11 @@ Coroutines can :keyword:`return` a value, so that they can be used by other coro
 Concurrent Execution
 ====================
 
-Coroutines can be scheduled for concurrent execution with :func:`~cocotb.start` and :func:`~cocotb.start_soon`.
+Coroutines can be scheduled for concurrent execution with :func:`~cocotb.start_soon`.
 These concurrently running coroutines are called :class:`~cocotb.task.Task`\ s.
 
-The :keyword:`async` function :func:`~cocotb.start` schedules the coroutine to be executed concurrently,
-then yields control to allow the new task (and any other pending tasks) to run,
-before resuming the calling task.
-
-:func:`~cocotb.start_soon` schedules the coroutine for future execution,
-after the calling task yields control.
+:func:`~cocotb.start_soon` schedules the coroutine for *future* execution,
+some time after the current Task yields control.
 
 .. code-block:: python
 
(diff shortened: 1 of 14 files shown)


docs/source/coroutines.rst
```diff
@@ -53,15 +53,11 @@ Coroutines can :keyword:`return` a value, so that they can be used by other coro
...
 These concurrently running coroutines are called :class:`~cocotb.task.Task`\ s.
 
-The :keyword:`async` function :func:`~cocotb.start` schedules the coroutine to be executed concurrently,
-then yields control to allow the new task (and any other pending tasks) to run,
-before resuming the calling task.
-
-:func:`~cocotb.start_soon` schedules the coroutine for future execution,
-after the calling task yields control.
+:func:`~cocotb.start_soon` schedules the coroutine for *future* execution,
+som time after the current Task yields control.
```
rkryan, line 60:
```suggestion
some time after the current Task yields control.
```

cmarqu:
Please also remove it from the refcard; it's there two times.

marlonjames:
If we are going to deprecate this, I'd rather provide a better user-facing way to yield to other queued Tasks than `await NullTrigger()`.

ktbarrett:
> If we are going to deprecate this, I'd rather provide a better user-facing way to yield to other queued Tasks than `await NullTrigger()`.

Are you saying you want a Trigger that fires after the next time a Task runs?

marlonjames:
I'm more saying that `NullTrigger()` is a bad name, and we should have something else for users.

ktbarrett:
> I'm more saying that `NullTrigger()` is a bad name, and we should have something else for users.

I'll definitely agree that it's not a great name. We can discuss this elsewhere.

cmarqu:
Needs rebase.

ktbarrett:
@cmarqu Rebased.

cmarqu (approved):

Issue #4401, opened 2025-01-29
A Release() at the end of a test never occurs


rkryan:
In the following example code, the intent is to remove all applied Forces at the end of the test. If Release() is the final operation in a test, the Release never occurs, which leads to the Force() persisting to the next test. If instead the simulator is advanced before finishing the test, such as with the RisingEdge call, the Release does occur.

```
@cocotb.test()
async def test_reg_read(dut):
    await init_model(dut)

    try:
        value = random.getrandbits(64)
        dut.regs.reg_q.value = cocotb.handle.Force(value)
        read_data = await reg_read(dut, 0xabc0)
        assert read_data == value, \
            f"Actual read data 0x{int(read_data):016x} != expected read data 0x{value:016x}"
    finally:
        dut.regs.reg_q.value = cocotb.handle.Release()
        # await cocotb.triggers.RisingEdge(dut.main_clock)
```

This was initially found with a VHDL model with Xcelium.

This does also lead to the tangential general question of Forces persisting between tests if they are not Released. This is the current behavior, but can cause some unexpected behavior if care is not taken.

ktbarrett:
There are a couple issues here.

1. The write scheduler is cleaned up as soon as the test ends, not allowing the last set of writes to occur.

These lines should be moved to `_schedule_next_test`.
https://github.com/cocotb/cocotb/blob/09c268647a42087d45f95396381785cdf91253e1/src/cocotb/regression.py#L508-L509

2. If the test fails in some other Task, the `finally` block is never run. This requires implementing `CancelledError`, which should come in with #4280 some time soon.

ktbarrett:
This is a little more complicated than I thought. The issue is that `_do_writes` task must survive after the end of the test, or the task and callback will be cancelled, not just the writes. Unfortunately, because the scheduler only works on Tasks, and all Tasks must be killed at the end of the test, that means we would have to introduce a special case into the scheduler to prevent the `_do_writes` Task from being cancelled between tests. 

Seeing as I intended to totally rewrite the scheduler and test shutdown in #4280, I don't see much of a reason to fix this now. It will have to wait until that PR comes in sometime in the next couple weeks ideally.

ktbarrett:
The other part of the issue is that these writes *shouldn't be scheduled at all* (#4466), but should be applied immediately. #4479 should solve some of your issues.

ktbarrett:
[...]

PR #4402, opened 2025-01-29
Add failing testcase for release scheduling

rkryan:
A Release() occurring at the end of a test does not occur. It is scheduled to occur, but the test ends, and the scheduled writes are cleared. This testcase deliberately hits this error.

This is detailed in #4401.


cmarqu:
The test would also fit in https://github.com/cocotb/cocotb/blob/master/tests/test_cases/test_force_release/test_force_release.py, maybe with a fencing comment saying that the two tests belong together.
We try to avoid issue-specific tests (though we aren't strict about it).

ktbarrett:
I did move the test to `test_force_release` per @cmarqu's suggestion.

tests/test_cases/test_force_release/test_force_release.py
```diff
@@ -190,3 +190,27 @@ async def test_force_followed_by_release_correct_value(dut):
...
+
+@cocotb.test
+async def test_apply_force(dut) -> None:
+    """Release a Force as the last operation in a test."""
+    dut.stream_out_data_comb.value = Force(42)
+    await Timer(10, "ns")
+    dut.stream_out_data_comb.value = Release()
+
+
+@cocotb.test
```
ktbarrett, line 208:
```suggestion
@cocotb.test(expect_fail=True)
```

ktbarrett: We could probably get this in now with this change. This will start failing after we fix the issue and it starts passing. It looks like NVC and Verilator work anyway, so the expect_fail expression would need to check for those sims.

cmarqu:
"It looks like NVC and Verilator work anyway, so the expect_fail expression would need to check for those sims." from https://github.com/cocotb/cocotb/pull/4402#discussion_r1942187603 still needs to be done.

ktbarrett:
My memory is short and my reading skills poor (apparently).

tests/test_cases/test_force_release/test_force_release.py
```diff
@@ -190,3 +190,27 @@ async def test_force_followed_by_release_correct_value(dut):
...
+async def test_apply_force(dut) -> None:
+    """Release a Force as the last operation in a test."""
+    dut.stream_out_data_comb.value = Force(42)
+    await Timer(10, "ns")
+    dut.stream_out_data_comb.value = Release()
+
+
+@cocotb.test(expect_fail=True)
```
ktbarrett, line 208:
```suggestion
@cocotb.test(expect_fail=not SIM_NAME.startswith(("verilator", "nvc"))
```

cmarqu (approved):

PR #4409, opened 2025-01-29
Handle passing no triggers to Combine and First

### LINKED ISSUE #4407: Passing empty list to `Combine` or `First` hangs the simulator indefinitely
There's an InternalEvent that must be set to allow the Waitable to finish, but it's never set if there are no Triggers or Tasks running, and miraculously no other code fails if the list in empty.

There should be a check for a non-empty list in the `_AggregateWaitable` class to prevent this.

## Proposals

The real question is, should this error, or should this be equivalent to `NullTrigger`?

`First` returns the result of the first argument to first, so without at least one object, this is not possible. Conceptually "first of" requires at least one. So it make sense to me for `First()` to fail. 

`Combine` returns itself when it fires and making generic code that can wait on all of "nothing" seems reasonable, so maybe we have a split behavior here where `Combine()` returns immediately.

ktbarrett:
Closes #4407.  `First()` causes a `ValueError` and `Combine()` acts like `NullTrigger()`.

## TODO
- [x] newsfrag

### DIFF
--- a/docs/source/newsfragments/4409.bugfix.1.rst
+++ b/docs/source/newsfragments/4409.bugfix.1.rst
@@ -1,0 +1,1 @@
+Passing no triggers to :class:`~cocotb.triggers.First` used to hang the simulation indefinitely. Now it raises a :exc:`ValueError` exception.
--- a/docs/source/newsfragments/4409.bugfix.1.rst
+++ b/docs/source/newsfragments/4409.bugfix.1.rst
@@ -1,1 +1,1 @@
-Passing no triggers to :class:`~cocotb.triggers.First` used to hang the simulation indefinitely. Now it raises a :exc:`ValueError` exception.
+Passing no triggers to :class:`~cocotb.triggers.First` previously hung the simulation indefinitely. Now, doing so raises a :exc:`ValueError` exception.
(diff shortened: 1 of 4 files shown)


cmarqu (approved):

docs/source/newsfragments/4409.bugfix.1.rst
```diff
@@ -0,0 +1 @@
+Passing no triggers to :class:`~cocotb.triggers.First` used to hang the simulation indefinitely. Now it raises a :exc:`ValueError` exception.
```
ktbarrett, line 1:
```suggestion
Passing no triggers to :class:`~cocotb.triggers.First` previously hung the simulation indefinitely. Now, doing so raises a :exc:`ValueError` exception.
```

docs/source/newsfragments/4409.bugfix.rst
```diff
@@ -0,0 +1 @@
+Passing no triggers to :class:`~cocotb.triggers.Combine` used to hang the simulation indefinitely. Now it behaves equivalently to :class:`await NullTrigger() <cocotb.triggers.NullTrigger>`.
```
ktbarrett, line 1:
```suggestion
Passing no triggers to :class:`~cocotb.triggers.Combine` previously hung the simulation indefinitely. Now, doing so makes ``Combine`` behave equivalently to :class:`await NullTrigger() <cocotb.triggers.NullTrigger>`.
```

PR #4410, opened 2025-01-29
Don't load cocotb libraries when compiling with VCS

ktbarrett:
This causes erratic warnings about not supporting `vpi_register_cb` in the compilation phase. We don't register system_tfs so there's no reason for us to load it here.


themperek:
As I can see (can always we wrong) we need to `-load` there is no other way to do it. 

Why functions are called during compilation? 

Interesting is that for VHPI there seems to other method to loading during run.

ktbarrett:
Is there no way to load during simulation only?

themperek:
I do not see way. We would need to ignore `register_initial_callback` and `register_final_callback` in `vlog_startup_routines` somehow.

marlonjames:
VHPI provides a way to get which phase you are in, but I don't see an equivalent in VPI.

ktbarrett:
Being handled by https://github.com/cocotb/cocotb/pull/4392.

Issue #2811, opened 2021-12-13
The 'external' decorator has regressed in functionality since v1.0


leftink:
The issue is present as of COCOTB v1.6.1.

The issue appears to be introduced in COCOTB v1.1.

In COCOTB v1.0, the external behaves the way it was intended, i.e. a thread is created and allowed to run will a Python Event trigger is provided to the scheduler to notify when the thread is done.  The scheduler was able to proceed and allow simulation to progress and the next time the event loop was run, if the thread had completed, the coroutine that was waiting on the external would then execute.

In the current implementation of the 'external' decorator, the scheduler blocks until all external threads have completed.  It doesn't even start all the external threads and then wait for all threads to complete.  The current implementation actually buys you nothing, i.e. I would essentially get the exact same execution by not using external and just call the function that was an 'external' and invoking it directly in my coroutine.   The scheduler is blocked and not proceeding either way.

## Use Case
In my design we are using a Versal Device from Xilinx.  The CIPS (i.e. processor) is provided by Xilinx using a SystemC model, TLM ports, and a socket to send and receive commands for something external to the simulator, i.e. QEMU, etc.  In my simulation environment, we have a COCOTB test that is expected to connect to the CIPS socket and send/receive commands from the CIPS model.  Since socket calls are blocking, I wrapped them in a function that is decorated by 'external'.  I then had a 'forked' coroutine to process incoming packets and trigger events.  With the current implementation in COCOTB however the simulator hangs because it is blocked in COCOTB waiting for the my socket read() to return before proceeding which will never happen because the simulator model is responsible for sending the packet.  If 'external' behaved the way it did back in v1.0, what I have implemented would work without any issues.

ktbarrett:
Likely changed in https://github.com/cocotb/cocotb/pull/565.

The implementation of external used to be implemented entirely within the the decorator; there was no scheduler interaction. You can probably copy/paste and see if this works for you.

https://github.com/cocotb/cocotb/blob/c1e259c1ba739d244ea2cdb4f857ea54e83d17a2/cocotb/decorators.py#L296-L322

We will have to discuss `external` and `function` with the maintainers and community to see if the current behavior is valuable to them, or not. I'm going to go out on a limb and guess almost no one uses this functionality and we can just change it back to the way it worked in 1.0 without peeving too many people off, but we will see.

eric-wieser:
[...]

PR #4425, opened 2025-02-03
Add further documentation for bridge() and resume() decorators

FMaxSokol:
The `bridge()` and `resume()` (formerly `external()` and `function()`) decorators are pretty sparsly documented so far.
There used to be a tutorial for these decorators in version 1.1, however this seems to have gotten lost.

This PR aims to a little more in-depth documentation about the usage of the `bridge()` and `resume()` decorators in the `Coroutines and Tasks` section.


ktbarrett:
Thanks for the effort of writing something up on this, but unfortunately the information is not quite correct. `bridge` threads block returning to the simulator and thus no simulation time is permitted to be consumed.

docs/source/coroutines.rst
```diff
@@ -273,6 +273,81 @@ It can be used to create a special type of generator function that can be iterat
...
+        return cocotb.resume(async_read)(dut)
+
+    def sync_write(dut, value):
+        """Go back into the asynchronous context and write the register."""
+
+        return cocotb.resume(async_write)(dut, value)
+
+    @cocotb.test()
+    async def test_synchronous_register_access(dut):
+        await cocotb.start(Clock(dut.clk, 1, units='ns').start())
```
cmarqu, line 338:
```suggestion
        Clock(dut.clk, 1, units="ns").start()
```
(since very recently, #4396)
We also use formatting like black would do it (not all places are like this yet, granted).

docs/source/coroutines.rst
```diff
@@ -273,6 +273,81 @@ It can be used to create a special type of generator function that can be iterat
 More details on this type of generator can be found in :pep:`525`.
 
 
+Calling blocking code
```
cmarqu, line 276:
```suggestion
Calling Blocking Code
```
More in line with our usual Title Case

docs/source/coroutines.rst
```diff
@@ -273,6 +273,81 @@ It can be used to create a special type of generator function that can be iterat
...
+        """Read the register of the DUT."""
+
+        return dut.register_out.value
+
+    async def async_write(dut, value):
+        """Write the register of the DUT."""
+
+        dut.write_enable.value = 1
+        dut.register_in.value = value
+        await Timer(2, units="ns") # wait at least one clock period
```
cmarqu, line 323:
```suggestion
        await Timer(2, units="ns")  # wait at least one clock period
```
black would do this :)

docs/source/coroutines.rst
```diff
@@ -273,6 +273,81 @@ It can be used to create a special type of generator function that can be iterat
...
+
+        return cocotb.resume(async_write)(dut, value)
+
+    @cocotb.test()
+    async def test_synchronous_register_access(dut):
+        await cocotb.start(Clock(dut.clk, 1, units='ns').start())
+
+        write = random.randint(0, 255)
+        await cocotb.bridge(sync_write)(dut, write) # call synchronous function by using bridge()
+        await Timer(2, units="ns") # wait for the output of the register to toggle
```
cmarqu, line 342:
```suggestion
        await cocotb.bridge(sync_write)(dut, write)  # call synchronous function by using bridge()
        await Timer(2, units="ns")  # wait for the output of the register to toggle
```

docs/source/coroutines.rst
```diff
@@ -273,6 +273,81 @@ It can be used to create a special type of generator function that can be iterat
...
+invoked using the :func:`~cocotb.resume` decorator.
+This approach allows the synchronous code to block in a separate thread while permitting the simulator
+to continue advancing time.
+
+The following code snippet demonstrates in a simple example how to do register read and write
+operations within cocotb from a synchronous context:
+
+.. code-block:: python3
+
+    import cocotb
```
cmarqu, line 309:
```suggestion
    import random

    import cocotb
```
Make it self-contained.

ktbarrett:
@cmarqu As for the formatting. Would it better to move any significant code snippets in the documentation into standalone Python files we can run `ruff` on and include them like some other examples do? 

Not that that needs to be done in this PR @FMaxSokol.

cmarqu:
> @cmarqu As for the formatting. Would it better to move any significant code snippets in the documentation into standalone Python files we can run `ruff` on and include them like some other examples do?

Yeah, I was on the edge of suggesting that... In case you *do* want to do that in this PR, @FMaxSokol, an example is https://github.com/cocotb/cocotb/blob/3b9fee3ddbba491bd5b7267fe56989b3e6bcd95d/docs/source/quickstart.rst?plain=1#L41-L44

docs/source/coroutines.rst
cmarqu, line 276:
```suggestion
.. _calling_blocking_code:

Calling blocking code
```
So that you can add a reference to this page from the docstrings of `bridge()` and `resume()`,
like in https://github.com/cocotb/cocotb/blob/3b9fee3ddbba491bd5b7267fe56989b3e6bcd95d/src/cocotb/triggers.py#L925C5-L925C48

docs/source/coroutines.rst
FMaxSokol: Changed, but the code was moved to a different file. See https://github.com/cocotb/cocotb/pull/4425/commits/b1128805c06519979cde008f0ff8e9af5047b1c9#diff-3a2eb8c324e7669e329fc2d01479baf2cdec96efed25103e410480c2151aa817R40

I don't know if there are other places in the code that are not formatted like black would do it, you'd have to point me to it :)

FMaxSokol:
> Thanks for the effort of writing something up on this, but unfortunately the information is not quite correct. `bridge` threads block returning to the simulator and thus no simulation time is permitted to be consumed.

I have removed the line stating this, therefore the information should be more accurate now.
Please correct me if that's not the case.

docs/source/coroutines.rst
FMaxSokol: [Changed](https://github.com/cocotb/cocotb/pull/4425/commits/b1128805c06519979cde008f0ff8e9af5047b1c9#diff-49a484affda213546436c7069a32d050e92ab760d00e1a591cdf8e85faacd4b0R278)

docs/source/coroutines.rst
FMaxSokol: [Lines were moved to a different file with an adapted coding style](examples/doc_examples/coroutines/calling_blocking_code/test_sync_register_access.py) :)

docs/source/coroutines.rst
FMaxSokol: [Lines were moved to a different file with an adapted coding style](examples/doc_examples/coroutines/calling_blocking_code/test_sync_register_access.py) :)

docs/source/coroutines.rst
FMaxSokol: [Done](examples/doc_examples/coroutines/calling_blocking_code/test_sync_register_access.py)

FMaxSokol:
> @cmarqu As for the formatting. Would it better to move any significant code snippets in the documentation into standalone Python files we can run `ruff` on and include them like some other examples do?
> 
> Not that that needs to be done in this PR @FMaxSokol.

I like it and have now done exactly that. I have expanded the example with the neccesary files to be runnable and put it in a new folder: examples/doc_examples/coroutines/calling_blocking_code/.

Please take a look at it and provide me with feedback on what things to change :)

docs/source/newsfragments/4425.doc.rst
```diff
@@ -0,0 +1 @@
+An example of how to use the :func:`cocotb.resume` and :func:`cocotb.bridge` decorators is now available in the :ref:`Coroutines and Tasks<coroutines>` section.
```
ktbarrett, line 1: This should pull the title following the ref.
```suggestion
An example of how to use the :func:`cocotb.resume` and :func:`cocotb.bridge` decorators is now available in the :ref:`coroutines` section.
```

ktbarrett (commented):
Perhaps I don't understand the whole read/write thing you are going for in this example, but how does this require bridge/resume?

Is there a reason you can't just stick a blocking call in a coroutine?
```python
async def do_stuff():
    value = blocking_read()
    print("read", value)
```

From what I gather from https://github.com/cocotb/cocotb/issues/2811#issuecomment-1047199139 the use case is for passing a `cocotb.resume` wrapped function as a callback to SW. And because the scheduler prevents returning control to the simulator until all `cocotb.bridge` threads enter a `cocotb.resume` context, this callback must be executed without simulation time passing.

ktbarrett:
pre-commit.ci autofix

examples/doc_examples/coroutines/calling_blocking_code/test_sync_register_access.py
```diff
@@ -0,0 +1,54 @@
...
+
+@cocotb.test()
+async def test_synchronous_register_access(dut):
+    """Test synchronous register read / write"""
+
+    Clock(dut.clk, 1, units="ns").start()
+
+    write = random.randint(
+        1, 256
+    )  # write some other value than the registers initial value
```
cmarqu, line 48:
```suggestion
    )  # write some other value than the register's initial value
```
We are in the realm if nitpicking now, but why not :)

docs/source/newsfragments/4425.doc.rst
FMaxSokol: [Fixed :)](https://github.com/cocotb/cocotb/pull/4425/files#diff-e45d676ef1b1677d26b67c0eb6951bb4d20dd053b0458465735a9ce3590a58bcR1)

ktbarrett:
It's not a matter of complexity to me. It's about being very specific about the exact use case this is designed to support. Otherwise you will have people who don't really understand it overusing it (like I've seen a lot recently with `cocotb.start` vs `cocotb.start_soon`), or wasting time and becoming frustrated or bothering us because it doesn't work the way they intuitively understand it, because it's just not the right tool. Contrived examples don't help and actually hurt because now we're advertising it.

A very specific case study would be more useful. But also maybe narrowing it down to "do it in this situation and not otherwise". As far as I understand it, a quick smell test you could mention a user to apply could be: "If you aren't passing `cocotb.resume` wrapped functions to the same SW interface you eventually call into with a `cocotb.bridge` wrapped function, you probably don't need this."

Assuming even I understand how its intended to be used... I've never seen anyone use this outside of a handful of people in Github issues complaining when people have inadvertently changed the functionality in the past. It's also not very well tested. So if it's valuable to you and you don't want to end up with regressed functionality, tests would be appreciated.

Combining that all together, a less contrived example that can be included in docs, but also run in the regression like the mixed signal examples, would be perfect.

PR #4429, opened 2025-02-03
Use uv to install dependencies in RTD

cmarqu:
https://github.com/readthedocs/readthedocs.org/issues/11289#issuecomment-2103747567 and the one after that suggest quite some speed improvements.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,20 +4,40 @@
 
 version: 2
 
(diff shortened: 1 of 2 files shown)


cmarqu:
Needs work in `pyproject.toml`; I'll continue experimenting outside of the repo.

cmarqu:
While this works and appears to reduce the overall build time on RTD by about 30 secs, the required `[project]` table in `pyproject.toml` might have negative side effects, e.g. when building the cocotb package.
Closing this for now.

PR #4432, opened 2025-02-04
Coverage Improvements

### LINKED ISSUE #4422: Re-enable doctests
pytest running doctests was removed in https://github.com/cocotb/cocotb/pull/3325. These should be re-enabled if possible.

ktbarrett:
* Add optimization and reduce debugging for release
* Re-enable doctests! (closes #4422)
* Update number of CI runs for coverage reports
* Exclude all TYPE_CHECKING blocks

WRT setting the optimization and debugging flags... The flags used to build the Python executable are used to build extensions due to how distutils uses some sysconfig values. This means that it *could* be passing `-g` to the release build, or `-O2` to dev builds. So we should explicitly override these.

For release builds I added `-O2 -g1`. No optimization flags are set by cibuildwheel and `-O2` could help performance and also reduce binary size over something like `-O0`.  I'm not so worried about binary size, but `-g1` might reduce binary size over `-g` (equivalent to `-g2`) a bit while preserving function names for tracebacks. Alternatively, we could pass `-g0 -Wl,-s` to turn off debugging and strip symbols.

For debug builds we explicitly pass `-Og -g` to optimize for debuggability and improve coverage.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -6,7 +6,7 @@ import os
 import shutil
 from contextlib import suppress
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple, cast
 
 import nox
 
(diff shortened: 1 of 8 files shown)


ktbarrett:
Just to highlight the change. This PR increases coverage by 1.32% all in the C++ code.

noxfile.py
```diff
@@ -390,11 +391,23 @@ def release_build_bdist(session: nox.Session) -> None:
     # Pin a version to ensure reproducible builds.
     session.run("pip", "install", f"cibuildwheel=={cibuildwheel_version}")
 
+    flags = " ".join(
+        [
+            "-O2",  # O3 is unnecessary with the virtual-call-heavy code the GPI is and may bloat binary size
+            "-g1",  # Include only minimal debug information
```
imphil, line 396: I have rarely seen `-g1` in use. I would just keep it at `-g` for simplicity and since we (as you also said) don't really care about file size here.

imphil (approved):

noxfile.py
ktbarrett:
```suggestion
            "-g",  # Include debug information to allow users to debug
```

PR #4433, opened 2025-02-04
Add sphinx-codeautolink extension

cmarqu:
Provides automatic links from code examples (code-blocks and literalincludes) to reference docs

Example: https://cocotb--4433.org.readthedocs.build/en/4433/coroutines.html

The renaming of the Pygments lexer name `python3` to `python` is due to https://github.com/felix-hilden/sphinx-codeautolink/issues/160

- [x] Make the links less prominent (https://sphinx-codeautolink.readthedocs.io/en/latest/examples.html#custom-link-styles)
- [x] Set up https://sphinx-codeautolink.readthedocs.io/en/latest/examples.html#doctest-code-blocks
- [x] Decide if we are okay with adding all those import statements in example code snippets (I have done only a few for now) or if we want to use their "invisible imports": https://sphinx-codeautolink.readthedocs.io/en/latest/examples.html#invisible-imports
- [x] Handle missing imports ("how" depends on the previous question)


cmarqu:
An open question is if we are okay with adding all those import statements in example code snippets (I have done only a few for now) or if we want to use their "invisible imports": https://sphinx-codeautolink.readthedocs.io/en/latest/examples.html#invisible-imports

cmarqu:
I will also make the links a bit less prominent (https://sphinx-codeautolink.readthedocs.io/en/latest/examples.html#custom-link-styles)

cmarqu:
@ktbarrett ruff now wants to also re-format your aligned comments in `cocotb/types/` - are you okay with that, or should we set `fmt: off`/`fmt: on` for these?

src/cocotb/types/logic.py
```diff
@@ -90,7 +92,9 @@ class Logic:
 
     :class:`Logic` supports the common logic operations ``&``, ``|``, ``^``, and ``~``.
 
-    .. code-block:: python3
+    .. code-block:: pycon3
+
+        >>> from cocotb.types.logic import Logic
```
ktbarrett, line 97: ??? This is this type, why is this necessary?

docs/source/coroutines.rst
```diff
@@ -107,7 +119,14 @@ Other tasks can be used in an await statement to suspend the current task until
...
 
-.. code-block:: python3
+.. code-block:: python
+
+    from math import isclose
+
+    import cocotb
+    from cocotb.clock import Clock
+    from cocotb.triggers import RisingEdge, Timer
+    from cocotb.utils import get_sim_time
```
ktbarrett, line 129: Yuck. Is there any way to prevent this? Maybe if this were a standalone file?

ktbarrett:
The ruff changes look fine to me. I'm more worried about every example starting with a bunch of imports that are otherwise unnecessary to make the point this documentation is trying to make.

docs/source/coroutines.rst
ktbarrett: Nvm, it looks like codeautolink handles this with "invisible imports". Definitely do that.

cmarqu:
> I'm more worried about every example starting with a bunch of imports that are otherwise unnecessary to make the point this documentation is trying to make.

Handled via "invisible imports" now.

ktbarrett (approved):

cmarqu:
I have switched from dotted underline to solid underline for these new links because "dotted" suggests a popup, not a "real" link.

Issue #3441, opened 2023-10-13
First and Combine can leak Tasks when the waiting Task is cancelled


marlonjames:
cocotb: master 5a892705222c651a84888a316eedd8f9da03ef6e

---

Because Waitables `First` and `Combine` are not actually triggers, and they rely on an `_InternalEvent` in the scheduler, they can leak Tasks.

If a `Task` that is awaiting a `Waitable` is cancelled, only the `_InternalEvent` is unprimed, and any waiter Tasks continue until the `Trigger` they are waiting on fires, or the test ends and they are cleaned up.

Consider this test:
```py3
@cocotb.test()
async def test_cancel_task_waiting_on_first(_):
    """Test that cancelling a task waiting on First trigger correctly unprimes the trigger and cleans up waiter tasks"""

    async def waiter():
        t1 = Timer(1, 'us')
        t2 = Timer(2, 'us')
        await First(t1, t2)

    task = cocotb.start_soon(waiter())

    await Timer(100, 'ns')
    task.cancel()

    assert len(cocotb.scheduler._trigger2coros) == 2

    await Timer(10, 'us')

    assert len(cocotb.scheduler._trigger2coros) == 0
```
In this example, not only does cancelling `task` leak 2 waiter Tasks, but they **both** fire by calling `_InternalEvent.set()`, since `task` is no longer around to cancel the remaining waiter task after the first one fires.

The impact is fairly low, since no `Task` is waiting on the `_InternalEvent` when it fires.

ktbarrett:
Could this be solved by adding a `deprime` overload to the `AggregateWaitable` class?

marlonjames:
The problem is that the scheduler only has the `_InternalEvent` trigger (because Waitables are not triggers at all), and that `_InternalEvent` is local to the coroutine `_wait()` that is running in the `Task`.

I see two options:
1. Keep a list on the `AggregateWaitable` object of the tasks that are created in `_wait()`, and create `_InternalEvent.unprime()` that calls something on the parent `AggregateWaitable` to cancel those tasks.
2. Implement throwing `CancelledError` into the Task and do the clean up in `_wait()`.

ktbarrett:
I tend to prefer expressing behavior structurally in code (`CancelledError`) rather than methods.

ktbarrett:
Is there a particular reason we don't treat Waitables as Triggers though? I feel like killing the waiter tasks would be very appropriate in a deprime method.

marlonjames:
[...]

PR #4449, opened 2025-02-10
Fix leak of waiters in `Combine`

ktbarrett:
If a `Combine` ends due to an exception, the waiters aren't killed allowing them to keep running.

`First` and `Combine` were reimplemented using new Task functionality (`_add_done_callback`). `Combine` now cleans up unfinished waiter before throwing the exception. More tests are added. This also fixes a bug where `GeneratorExit`s were not being handled correctly when waiter tasks were killed in `Combine`.

#3441 is left unresolved. Once #4280  is in and `CancelledError`s are supported, we can hook that to clean up the waiters.

### DIFF
--- a/tests/test_cases/test_cocotb/test_synchronization_primitives.py
+++ b/tests/test_cases/test_cocotb/test_synchronization_primitives.py
@@ -6,12 +6,21 @@ Tests for synchronization primitives like Lock and Event
 """
 
 import re
+from typing import Any
 
 import pytest
+from common import MyException
 
 import cocotb
 from cocotb._base_triggers import _InternalEvent
-from cocotb.triggers import Combine, First, Lock, NullTrigger, Timer
+from cocotb.triggers import (
+    Combine,
+    Event,
+    First,
+    Lock,
+    NullTrigger,
+    Timer,
+)
 from cocotb.utils import get_sim_time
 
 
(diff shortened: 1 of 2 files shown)


ktbarrett:
Ooof. Testing Combine a little more on master has it breaking badly when a Combine results in an exception. This change fixes the original error, but is now not handling `GeneratorExit`s correctly. I need to spend more time understanding this to fix the bug.

ktbarrett:
@cmarqu pydocstring only considers publicly accessible objects. Functions within functions aren't publicly accessible.

src/cocotb/triggers.py
```diff
@@ -939,28 +930,45 @@ async def _wait(self) -> "Combine":
...
+
+            def on_done(
+                task: cocotb.task.Task[object],
+            ) -> None:
+                # have to check cancelled first otherwise exception() will throw
+                if task.cancelled():
+                    waiters.remove(task)
+                    if not waiters:
+                        done.set()
+                        return
```
ktbarrett, line 945: We won't be able to hit these lines until we get `CancelledError` working. I'll make an issue to come back and fix this up after `CancelledError` is working.

tests/test_cases/test_cocotb/test_synchronization_primitives.py
```diff
@@ -152,16 +162,93 @@ async def await_internalevent():
...
+    first = First(cocotb.start_soon(raises_after_1ns()), Timer(10, "ns"), e.wait())
+    start_time = get_sim_time(units="ns")
+    with pytest.raises(MyException):
+        await first
+    end_time = get_sim_time(units="ns")
+    assert end_time == (start_time + 1)
+
+
+@cocotb.test
+async def test_events_in_combine(_: Any) -> None:
```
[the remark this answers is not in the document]
ktbarrett:
```suggestion
async def test_Combine_objects_shared_by_multiple(_: Any) -> None:
    """Test waiting for the same objects in multiple concurrent Combines."""
```

src/cocotb/triggers.py
ktbarrett: I just added a comment to #3441.

ktbarrett:
Rebased.

cmarqu (approved):

PR #4458, opened 2025-02-12
Use non-repeating VHPI timing callbacks

ktbarrett:
These don't need to be repeating, especially after #4392. Let's see how well they are supported.


ktbarrett:
Looks like Questa was the only simulator to not implement this.

PR #4461, opened 2025-02-13
Upstream various changes

ktbarrett:
The first commit avoids a very egregious performance regression.

### DIFF
--- a/src/cocotb/_scheduler.py
+++ b/src/cocotb/_scheduler.py
@@ -314,7 +314,7 @@ class Scheduler:
         for task in scheduling:
             # unset trigger
             task._trigger = None
-            self._schedule_task(task)
+            self._schedule_task_internal(task)
 
         # cleanup trigger
         trigger._cleanup()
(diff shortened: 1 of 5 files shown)


src/cocotb/_scheduler.py
```diff
@@ -424,6 +422,11 @@ def _schedule_task(
         for tasks in self._trigger2tasks.values():
             if task in tasks:
                 return
```
ktbarrett, line 424: These checks right here are rather expensive on simulations of any size. What's worse is it doing these checks every time an existing Task is rescheduled after a Trigger fires. This check isn't needed in those cases. So we only do the check on user interfaces.

src/cocotb/types/logic_array.py
```diff
@@ -859,13 +859,8 @@ def __setitem__(
...
                 )
-            value_as_logics = [
-                Logic(v) for v in cast(Iterable[LogicConstructibleT], value)
-            ]
-            if len(value_as_logics) != (stop_i - start_i + 1):
-                raise ValueError(
-                    f"value of length {len(value_as_logics)!r} will not fit in slice [{start}:{stop}]"
-                )
+            value = cast(Union[str, Iterable[LogicConstructibleT], int], value)
+            value_as_logics = LogicArray(value, stop_i - start_i + 1)
```
ktbarrett, line 863: While this is less performant than the previous implementation, it is more flexible, and honestly a rare operation.

marlonjames (approved):

PR #4462, opened 2025-02-14
Cache getting values from handles

### LINKED ISSUE #4460: Cache getting values from handles
We can cache getting the values of handles to prevent going into C and then constructing a new `LogicArray` by using the assumption that values can't change until simulation time / deltas passes. 

Rather than invalidating the cache every time we return to or from the simulator via a GPI callback, we can keep count of GPI cycles, tag the value cached in the handle with the current GPI cycle on creation, and in the future check to see if the value is invalidated by comparing the current GPI cycle against the one in the value cache.

The user could modify the value coming from the handle, which can be solved by creating a `_dirty` variable which is set whenever a mutating operation is done.

Some simulators make `setimmediatevalue` actually set immediately, so this too would have to invalidate the cache by setting the cycle count back to 0.

ktbarrett:
Closes #4460.

* Cache values from the simulator on the handle.
* Invalidate the value on every GPI Trigger firing.
* Introduce a global timestamp value that increments whenever a GPI Trigger fires to aid in invalidating the cached value when GPI Trigger fires.
* Invalidate the value if the type is mutable and it becomes dirty by being mutated (only `LogicArray` suffers this)
* Invalidate the value if we write the value "immediately" as some simulators do update the value immediately.
* Refactor of handle value getter and setter to prevent repeating ourselves.

## TODO
- [x] newfragment
- [x] test LogicArray dirty
- [x] test getting handle value multiple times (I'm surprised this wasn't hit)

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -295,6 +295,8 @@ Logging
     This is added by :class:`cocotb.log.SimTimeContextFilter`.
 
 
+.. _sim-handles:
+
 Simulation Object Handles
 =========================
 
--- a/docs/source/newsfragments/4462.feature.rst
+++ b/docs/source/newsfragments/4462.feature.rst
@@ -1,0 +1,1 @@
+Improve performance by caching simulator values on :ref:`sim-handles`.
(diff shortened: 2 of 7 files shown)


ktbarrett:
Turns out people getting the value of a handle multiple times in the same time step doesn't really happen as often as I expected. `test_cocotb` and `matrix_multiplier` in this repo *never* do it. Testing it on larger codebases shows a slight degradation of performance due to the overhead of the caching mechanics. 

Another obvious performance optimization that just isn't so. I'm starting to think there really isn't a whole lot we need to do wrt performance optimizations.

PR #4463, opened 2025-02-15
Raise logging level of exceptions from tests

hexagonrecursion:
We run cocotb with COCOTB_LOG_LEVEL=WARNING because the default INFO is too chatty for us. When a test throws an exception, the only piece of information that is printed to the console is that a test failed. We would like cocotb to print which test failed, which exception was thrown and a traceback - similar to how pytest behaves when a test throws an exception.

If this is accepted, I will backport the change to stable/1.9

### DIFF
--- a/docs/source/newsfragments/4463.change.rst
+++ b/docs/source/newsfragments/4463.change.rst
@@ -1,0 +1,1 @@
+Report exceptions from tests even when COCOTB_LOG_LEVEL=WARNING or COCOTB_LOG_LEVEL=ERROR
--- a/docs/source/newsfragments/4463.change.rst
+++ b/docs/source/newsfragments/4463.change.rst
@@ -1,1 +1,1 @@
-Report exceptions from tests even when COCOTB_LOG_LEVEL=WARNING or COCOTB_LOG_LEVEL=ERROR
+Report exceptions from tests even when COCOTB_LOG_LEVEL=WARNING
--- a/docs/source/newsfragments/4463.change.rst
+++ b/docs/source/newsfragments/4463.change.rst
@@ -1,1 +1,1 @@
-Report exceptions from tests even when COCOTB_LOG_LEVEL=WARNING
+Log test failures at log level :data:`~logging.WARNING`.
--- a/src/cocotb/regression.py
+++ b/src/cocotb/regression.py
@@ -781,7 +781,8 @@ class RegressionManager:
         else:
             rest = f": {msg}"
         self.log.warning(
-            "%s %sfailed%s%s",
+            "%s%s %sfailed%s%s",
+            stop_hilight,
             self._test.fullname,
             start_hilight,
             stop_hilight,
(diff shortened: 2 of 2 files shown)


cmarqu (approved):

cmarqu:
I think this makes sense, let's hear other maintainers though.
No need to manually backport, we have automation for that since a while.

ktbarrett (changes requested):
There are people who do some funny things with logs. Even cocotb back in early 1.X would hook error logs to cause test failures. Raising it to ERROR is a no-go from me. Maybe WARNING?

But what's chatty about the cocotb logs? Right now at the default log level `INFO` it prints simulator, simulator version, cocotb version, python interpreter, and random seed at startup. Test starting and test finishing are logged to delimit logs that occur between so you know which test to attach them to. Then there's the summary at the end (Honestly we could probably just print the tests that failed instead of printing all of them). This is all necessary and sufficient information IMO.

cmarqu:
> Then there's the summary at the end (Honestly we could probably just print the tests that failed instead of printing all of them).

I suppose the case here is that users will print out more information as the test runs, and by default, they can either do that as `INFO` or `DEBUG`, with both having downsides.
At work, some of our tools have an extra level `VERBOSE` that's between `INFO` and `DEBUG` that has been useful for such additional output. Maybe we should have this in cocotb?

(FWIW, we already have an extra level `TRACE` since #2502.)

ktbarrett:
@cmarqu. We've been considering adding VERBOSE levels at my place of work as well to replace people using `verbose` flags to gate INFO logs. But users shouldn't be using loggers under `"cocotb"`. That's for cocotb internals to use. Those loggers should remain at INFO by default IMO. Considering cocotb's needs, we don't need a VERBOSE. People should make their own loggers so they can control them independently.

I know our documentation recommends using `task.log` and `handle._log` and examples show `cocotb.log`. These are all bad and it's on my long list of things to clean up.

hexagonrecursion:
> I think this makes sense, let's hear other maintainers though. No need to manually backport, we have automation for that since a while.

I doubt your automation will be able to backport this patch because the code appears to have been refactored since 1.9

> Maybe WARNING?

Ok. Works for me. Done

src/cocotb/regression.py
```diff
@@ -780,7 +780,7 @@ def _record_test_failed(
             rest = ""
         else:
             rest = f": {msg}"
-        self.log.info(
+        self.log.warning(
             "%s %sfailed%s%s",
             self._test.fullname,
```
ktbarrett, line 785: This is a little awkward looking. https://github.com/cocotb/cocotb/actions/runs/13363291841/job/37351967608#step:24:3189

```suggestion
            "%s%s %sfailed%s%s",
            stop_hilight,
            self._test.fullname,
```

docs/source/newsfragments/4463.change.rst
```diff
@@ -0,0 +1 @@
+Report exceptions from tests even when COCOTB_LOG_LEVEL=WARNING
```
ktbarrett, line 1:
```suggestion
Log test failures at log level :data:`~logging.WARNING`.
```

PR #4467, opened 2025-02-16
Improve Logging

### LINKED ISSUE #4441: Optimize logging from C
The two Python functions providing logging functionality to C call `logging.getLogger` on every call. This call could be slow and is definitely unnecessary if we simply store the "gpi" Logger object in C.

Additionally, it doesn't seem to make much sense to have both a `_filter_from_c` function as well as a locally cached log level. We should do one or neither, not both.

ktbarrett:
Closes #4441.

A series of small improvements and cleanups to loggers across the GPI and in Python.

Slight performance boost.

### Interface Changes

The `"cocotb"` log namespace is now for cocotb internals only. Some documentation that stated otherwise was removed. Loggers on cocotb objects are now all private. `cocotb.log` Logger name was changed to not be in the `"cocotb"` namespace.

`COCOTB_LOG_LEVEL` no longer defaults the `"gpi"` Loggers. There is a separate `GPI_LOG_LEVEL` for that. The `GPI_LOG_LEVEL` is also applied earlier during startup.

### TODO
- [x] newsfrags
- [x] check for coverage issues


src/cocotb/share/lib/py_gpi_log/py_gpi_logging.cpp
```diff
@@ -8,15 +8,20 @@
...
 #include <cstdio>   // fprintf, vsnprintf
+#include <map>      // std::map
 #include <vector>   // std::vector
 
 #include "cocotb_utils.h"  // DEFER
 #include "gpi_logging.h"   // all things GPI logging
-
-static PyObject *pLogHandler = nullptr;
+#include "object.h"
+#include "pytypedefs.h"
```
ktbarrett, line 17: Get rid of errant includes. `#include <Python.h>` is sufficient.

src/cocotb/share/lib/py_gpi_log/py_gpi_logging.cpp
```diff
@@ -91,15 +96,20 @@ static void py_gpi_log_handler(void *, const char *name, int level,
...
+        logger = PyObject_CallFunction(m_get_logger, "s", name);
         // LCOV_EXCL_START
-        PyErr_Print();
-        return fallback_handler(name, level, pathname, funcname, lineno,
-                                log_buff.data());
+        if (!logger) {
+            PyErr_Print();
+            return fallback_handler(name, level, pathname, funcname, lineno,
+                                    log_buff.data());
+        }
```
ktbarrett, line 108: Did we want to save the logger? lmao.

src/cocotb/share/lib/py_gpi_log/py_gpi_logging.cpp
```diff
@@ -160,15 +170,22 @@ extern "C" void py_gpi_logger_set_level(int level) {
...
 }
 
-extern "C" void py_gpi_logger_initialize(PyObject *handler) {
-    Py_INCREF(handler);
-    pLogHandler = handler;
+extern "C" void py_gpi_logger_initialize(PyObject *log_func,
+                                         PyObject *get_logger) {
+    Py_INCREF(log_func);
+    m_log_func = log_func;
+    m_get_logger = get_logger;
```
ktbarrett, line 177:
```suggestion
INCREF the get_logger

File: docs/source/building.rst [hunk 4]
```diff
@@ -128,10 +128,23 @@ Cocotb
...
+    This behaves similarly to ``INFO``.
 
-    ``TRACE`` is used for internal low-level logging and produces very verbose logs.
+    .. versionchanged:: 2.0
+        The root ``"gpi"`` logger level is no longer set when this environment variable is used.
+        Use :envvar:`GPI_LOG_LEVEL` instead.
+
+.. envvar:: GPI_LOG_LEVEL
+
+    The default log level of all ``"gpi"`` loggers, both Python and native GPI logger.
```
cmarqu, line 142: Maybe we can say someting like "(the low-level simulator interface)" next to "gpi"?

src/cocotb/__init__.py
```diff
@@ -52,7 +52,14 @@
 
 
 log: py_logging.Logger
-"""The default cocotb logger."""
+"""A logger for the user.
+
+This logger defaults to a :data:`logging.INFO` log level.
```
cmarqu, line 57:
```suggestion
This logger defaults to an :data:`logging.INFO` log level.
```

src/cocotb/__init__.py
```diff
@@ -52,7 +52,14 @@
...
+
+.. versionchanged:: 2.0
+    This used to be the case ``"cocotb"`` :class:`~logging.Logger`.
```
cmarqu, line 60:
```suggestion
    This used to be the base ``"cocotb"`` :class:`~logging.Logger`.
```

cmarqu:
We now could replace extra loggers like these, right? https://github.com/cocotb/cocotb/blob/5ccaf52e00faeecf0219c4a2901bd640e5bfcd58/tests/test_cases/test_array_simple/test_array_simple.py#L15

ktbarrett:
@cmarqu For sure. In fact, I'd say we could remove all the logging calls across the whole `test/` dir. Maybe leaving `discover_all` tests printing everything and that's occasionally useful.

ktbarrett:
The PR is simply moving around code that isn't already tested, hence the low patch code coverage due to the following defines and envvars:
* GPI_LOG_LEVEL
* COCOTB_USER_COVERAGE
* VPI_CHECKING
* VHPI_CHECKING

cmarqu (approved):

PR #4473, opened 2025-02-17
Improvements to `Lock`

### LINKED ISSUE #4281: Returning a Trigger from `Lock.acquire` results in bad behavior
```python
@cocotb.test
async def wew(dut):
    l = Lock()
    await l.acquire()
    t = l.acquire()
    # t is a Trigger than can be awaited multiple times as seen below
    async def waiter1():
        await t
    cocotb.start_soon(waiter1())
    await t
    # both this Task and the waiter1 Task will wake up when the acquire Trigger fires
    # both will think they have acquired the Lock
```

### LINKED ISSUE #4468: Guarantee fair lock acquisition
Based on the implementation it should already be fair. We should test that it is and document this guarantee. This is a fairly critical feature for a lock.

ktbarrett:
Closes #4281. Closes #4468.

- [x] newsfragments

### DIFF
--- a/tests/test_cases/test_cocotb/test_synchronization_primitives.py
+++ b/tests/test_cases/test_cocotb/test_synchronization_primitives.py
@@ -13,7 +13,7 @@ import pytest
 from common import MyException
 
 import cocotb
-from cocotb._base_triggers import _InternalEvent
+from cocotb._base_triggers import Trigger, _InternalEvent
 from cocotb.task import Task
 from cocotb.triggers import (
     Combine,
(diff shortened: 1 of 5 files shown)


tests/test_cases/test_cocotb/test_synchronization_primitives.py
```diff
@@ -254,3 +257,54 @@ async def wait_for_all_events():
...
+    # We cancel tasks randomly to ensure that doesn't effect acquisition order.
+    while not all(t.done() for t in tasks):
+        # Randomly kill a remaining waiter.
+        tasks[random.randrange(last_scheduled, len(tasks))].kill()
+        # Wait some random time until killing another.
+        await Timer(
+            (random.random() * 2 * average_killer_wakeup), "ns", round_mode="ceil"
+        )
+        # So we don't depend upon the relative scheduling order of the waiter Timers and the above Timer.
+        await ReadOnly()
```
ktbarrett, line 310: This should also test Tasks cancelling their acquisition and then trying to reacquire and see if it resets its position in queue.

Resetting position in queue is interesting... it's possible to live lock if you constantly give up acquiring and try again. Maybe maintaining position in queue is better? How would we do that?

Also this could test the behavior of reusing `_Lock` objects.

marlonjames (approved):

tests/test_cases/test_cocotb/test_synchronization_primitives.py
ktbarrett: Opened #4474 and #4475.

Issue #3995, opened 2024-06-28
`COCOTB_PDB_ON_EXCEPTION` could be more useful


ktbarrett:
Since this is only entered in the regression manager after the test ends (and `_cleanup` is called), there isn't much debugging that can be done. This should be entered *before* `_cleanup` is called. I see this as more of a debugging tool, so embedding this into the scheduler seems reasonable.

ktbarrett:
Also, it would be nice once #3880 is solved, the debugger can be entered whenever a `BaseException` (such as `KeyboardInterrupt`) is seen.

ktbarrett:
Finally, we should recommend using `breakpoint` and setting `PYTHONBREAKPOINT`.

PR #4477, opened 2025-02-18
Introduce `cocotb.pass_test` to replace `TestSuccess`

### LINKED ISSUE #1473: Add function to abort the currently running test
We should move away from raising a `TestComplete` as a way for the user to end a test immediately. Most of the hangups in #937 come from the fact that it still special cases `TestComplete`, because there is otherwise no way for a user to immediately end a test. So we should add one.

Asyncio in 3.7 introduced the `current_task` variable, we could do something similar:

```python
# __init__.py
def current_test():
    global scheduler
    return scheduler._test
```

```python
# user code
if dut.handle.value != expected_value:
    cocotb.current_test.abort(TestFailure("Actual did not match expected!"))
```

This may be relevant to #1449 as well.

### LINKED ISSUE #3787: Refactor `TestSuccess`
Asking users to throw an exception when they want to prematurely end a test successfully locks us into that implementation and signals to the user that `TestSuccess` is something that could be caught, which I don't think we want to support.

Instead, I think a better implementation of this feature uses a function, such as `cocotb.end_test()` that can take any kind of result, including `pass=True` or an `Exception`, etc. This could be implemented under the hood with an exception or whatever; we are more free in our implementation. It also prevents users from intercepting that call.

ktbarrett:
xref #3995. Closes #3787. Closes #1473.

- [x] newsfragment


ktbarrett:
The riviera break is related to #4485. Waiting until #4486 is in.

src/cocotb/__init__.py
```diff
@@ -136,7 +135,7 @@ def _task_done_callback(task: "cocotb.task.Task[Any]") -> None:
     if e is None:
         return
     # there was a failure and no one is watching, fail test
-    elif isinstance(e, (TestSuccess, AssertionError)):
+    elif isinstance(e, (_TestSuccess, _Failed, AssertionError)):
```
ktbarrett, line 138: This was a bug since before cocotb 2.0. We weren't handling `pytest.raises` and `pytest.warns` errors in Tasks. Whoops...

cmarqu (approved):

PR #4486, opened 2025-02-18
More Trigger improvements

### LINKED ISSUE #4485: `assert_takes` context manager for testing
I just ran into an issue where a test that uses the following common idiom fails because of floating point rounding errors.
Using steps would be better, but its easier to get it right once and make it reusable.

```python
start_time = get_sim_time("ns")
...
end_time = get_sim_time("ns")
assert end_time == (start_time + expected_ns)
```

ktbarrett:
Closes #4485.

I'm not sure *why* there needs to be separate `_Event`s for each Task, but making it a single shared object made the regression very unhappy so instead I made sure that, like Lock, only one Task can await on an `_Event` trigger simultaneously.

While fixing that I rewrote some broken tests that depended on that behavior. This uncovered a flaw in the implementation of `First` and `Combine` from #4449, so that was also fixed.

And that broke another test because of the issue described in #4485, so I resolved that as well.

### DIFF
--- a/tests/test_cases/test_cocotb/common.py
+++ b/tests/test_cases/test_cocotb/common.py
@@ -5,8 +5,14 @@
 Common utilities shared by many tests in this directory
 """
 
+import operator
 import re
 import traceback
+from contextlib import contextmanager
+from typing import Callable, Generator
+
+from cocotb._typing import TimeUnit
+from cocotb.utils import get_sim_steps, get_sim_time, get_time_from_sim_steps
 
 
 async def _check_traceback(running_coro, exc_type, pattern, *match_args):
@@ -29,3 +35,29 @@ class MyException(Exception): ...
 
 
 class MyBaseException(BaseException): ...
+
+
+@contextmanager
+def assert_takes(
+    time: float, unit: TimeUnit, cmp: Callable[[int, int], bool] = operator.eq
+) -> Generator[None, None, None]:
+    """Assert that the block takes a certain amount of time to finish.
+
+    The *cmp* function passes actual after first argument and expected as the second.
+    Other useful comparison functions are: :func:`math.isclose`, and the functions in the
+    :mod:`operator` module.
+
+    Args:
+        time: Time value.
+        unit: Unit of *time*.
+        cmp: Comparison function to use. Defaults to equality.
+    """
+    expected = get_sim_steps(time, unit)
+    start = get_sim_time("step")
+    yield
+    end = get_sim_time("step")
+    actual = end - start
+    actual_in_units = get_time_from_sim_steps(actual, unit)
+    assert cmp(actual, expected), (
+        f"Expected the code to take {time}{unit}, actually took {actual_in_units}{unit}"
+    )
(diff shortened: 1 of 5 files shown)


tests/test_cases/test_cocotb/common.py
```diff
@@ -29,3 +35,19 @@ class MyException(Exception): ...
 
 
 class MyBaseException(BaseException): ...
+
+
+@contextmanager
+def assert_takes(
+    time: float, unit: TimeUnit, cmp: Callable[[int, int], bool] = operator.eq
+) -> Generator[None, None, None]:
+    """Assert that the code takes at least `time` in `unit` to run"""
```
cmarqu, line 44:
```suggestion
    """Assert that the code takes at least *time* in *unit* to run."""
```

cmarqu, line 44: Docstring should say `cmp` can also be set to e.g.  `math.isclose()`.

tests/test_cases/test_cocotb/common.py
```diff
@@ -29,3 +35,19 @@ class MyException(Exception): ...
...
+    expected = get_sim_steps(time, unit)
+    start = get_sim_time("step")
+    yield
+    end = get_sim_time("step")
+    actual = end - start
+    actual_in_units = get_time_from_sim_steps(actual, unit)
+    assert cmp(actual, expected), (
+        f"Expected the code to take {time} {unit}, actually took {actual_in_units} {unit}"
```
cmarqu, line 52:
```suggestion
        f"Expected the code to take {time}{unit}, actually took {actual_in_units}{unit}"
```
(not sure how we print units in other places actually...)

tests/test_cases/test_cocotb/test_concurrency_primitives.py
```diff
@@ -8,48 +8,80 @@
...
-        assert ret_i == 1, f"Expected event 1 to fire, not {ret_i}"
 
-    async def fire_events():
-        """fire the events in order"""
-        for e in events:
-            await Timer(1, "ns")
-            e.set()
+@cocotb.test
+async def test_First_unfired_triggers_killed_on_exception(_) -> None:
+    """Test that un-fired trigger(s) in First after exception don't later cause a spurious wakeup"""
```
cmarqu, line 51:
```suggestion
    """Test that un-fired trigger(s) in First after exception don't later cause a spurious wakeup."""
```

tests/test_cases/test_cocotb/test_concurrency_primitives.py
```diff
@@ -8,48 +8,80 @@
...
+        pass
+
+    for t in triggers:
+        assert t.primed == 1
+        assert t.unprimed == 1
+
+
+@cocotb.test
+async def test_Combine_unfired_triggers_killed_on_exception(_) -> None:
+    """Test that un-fired trigger(s) in Combine after exception don't later cause a spurious wakeup"""
```
cmarqu, line 70:
```suggestion
    """Test that un-fired trigger(s) in Combine after exception don't later cause a spurious wakeup."""
```

marlonjames (approved):

PR #4492, opened 2025-02-21
Merge C++ libraries

ktbarrett:
* Random cleanup and fixing leak in handle_gpi_callback in simulator.so
* libgpilog, libgpi, libembed, and half of libcocotb_utils were merged into libgpi.
* libpygpilog, libcocotb, and simulator.so were merged into simulator.so.
* Introduced (currently) private GPI_USERS envvar which loads simulator.so dynamically rather than hardcoding it into the binary.

The GPI_USERS and entire embed code will be refactored into a more generator purpose feature for loading libraries dynamically. LIBPYTHON_LOC will be transitioned to a GPI_USERS method so there will be no more hardcoding the assumption of Python in the GPI.


ktbarrett:
Oh cool. That thing I didn't touch broke. Windows is heckin awesorino.

ktbarrett:
The failures on Windows mean that

> We can also probably drop the SXS appconfig and RPATH stuff in a follow on. All libraries are now explicitly dynamically loaded.

Will have to wait until #1994 is fixed and there is no dependency in libgpi on any unbound function in the user library.

ktbarrett:
Well... Maybe I'll just try fixing #1994 while I'm at it =/

ktbarrett:
I'm actually going to pause on this to focus on 2.0 features.

PR #4510, opened 2025-02-24
Always log interface issues at DEBUG

ktbarrett:
Adding this feature required making improvements to the GPI logging libraries. Namely, we needed a function to check the current log level so we could condition calling `vpi_chk_error` and `vhpi_check_error`.

### DIFF
--- a/docs/source/newsfragments/4510.change.rst
+++ b/docs/source/newsfragments/4510.change.rst
@@ -1,0 +1,1 @@
+``VPI_CHECKING`` and ``VHPI_CHECKING`` are removed and interface issues are now always emitted with a log level of :data:`logging.DEBUG`.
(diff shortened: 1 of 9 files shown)


docs/source/newsfragments/4510.change.rst
```diff
@@ -0,0 +1 @@
+``VPI_CHECKING`` and ``VHPI_CHECKING`` are removed and now always emitted with a log level of :data:`logging.DEBUG`.
```
ktbarrett, line 1:
```suggestion
``VPI_CHECKING`` and ``VHPI_CHECKING`` are removed and interface issues are now always emitted with a log level of :data:`logging.DEBUG`.
```

PR #4511, opened 2025-02-25
Backport #4510

ktbarrett:


ktbarrett:
Apparently this is not helpful, so saving this for 2.0 only.

PR #4512, opened 2025-02-25
`cocotb.types` API breaking cleanup

### LINKED ISSUE #4124: Cleanup exception types
- [x] ~~Is passing a negative number to `LogicArray.from_unsigned()` an OverflowError since it won't fit, or a ValueError, since the value is clearly not unsigned as required?~~ Support will actually be added.
  * The fact some values of a given type yield ValueError and others OverflowError makes handling more difficult
- [x] `Logic({})` yields `ValueError` rather than `TypeError`. (#4125)
- [x] ~~`Logic(123)`, `Logic("0101")` and similar issues should result in `OverflowError` if we really want to go down this path.~~
- [ ] Change occurences of `OverflowError` in `cocotb.handle` to be `ValueError`.

ktbarrett:
Primarily I was interested in getting rid of all the logic around constructing LogicArray with the whole 'range' vs 'width' nonsense. It's not that valuable and is slow. Same with the default values of Logic and thus LogicArray. I don't want to say the default should be `X` or `U`, and it's not valuable to us really. That helped make the LogicArray change make more sense since we will never see `LogicArray(range=5)` and might not ever see `range` passed as a keyword argument.

While I was there I cleaned up some other things and found a small bug to fix.
* Removed all uses of `OverflowError` in LogicArray and handle.py (xref #4124).
* Found and fixed a small bug with the RESOLVE_X logic.
* Added `Array._from_handle` to hopefully speed up Array construction from handles a bit.
* Touched up the documentation that looked a bit grody. 4 years is a long time ago...
* Fixed documentation of `ResolveX` which was broken.

### TODO
- [x] ~~newsfrags~~  No newsfrags needed, nothing changed from 1.9.

### DIFF
--- a/src/cocotb/types/array.py
+++ b/src/cocotb/types/array.py
@@ -1,7 +1,7 @@
 # Copyright cocotb contributors
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
-from typing import Iterable, Iterator, TypeVar, Union, cast, overload
+from typing import Iterable, Iterator, List, TypeVar, Union, cast, overload
 
 from cocotb.types import ArrayLike
 from cocotb.types.range import Range
@@ -151,6 +151,13 @@ class Array(ArrayLike[T]):
                 raise ValueError(
                     f"Value of length {len(self._value)!r} does not fit in {self._range!r}"
                 )
+
+    @classmethod
+    def _from_handle(cls, value: List[T], range: Range) -> "Array[T]":
+        self = cls.__new__(cls)
+        self._value = value
+        self._range = range
+        return self
 
     @property
     def range(self) -> Range:
--- a/src/cocotb/types/array.py
+++ b/src/cocotb/types/array.py
@@ -13,37 +13,44 @@ class Array(ArrayLike[T]):
     r"""Fixed-size, arbitrarily-indexed, homogeneous collection type.
 
     Arrays are similar to, but different from Python :class:`list`\ s.
-    An array can store values of any type or values of multiple types at a time, just like a :class:`list`.
-    Unlike :class:`list`\ s, an array's size cannot change.
-
-    The indexes of an array can start or end at any integer value, they are not limited to 0-based indexing.
-    Indexing schemes can be either ascending or descending in value.
-    An array's indexes are described using a :class:`~cocotb.types.Range` object.
-
-    Initial values are treated as iterables, which are copied into an internal buffer.
+    An array can store values of any type or values of multiple types at a time, just like a :class:`!list`.
+    Array constructor values can be any Iterable.
 
     .. code-block:: pycon3
 
-        >>> Array("1234")  # the 0-based range `(0, len(value)-1)` is inferred
-        Array(['1', '2', '3', '4'], Range(0, 'to', 3))
+        >>> Array([1, False, "example", None])
+        Array([1, False, 'example', None], Range(0, 'to', 3))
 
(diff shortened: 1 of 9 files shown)


ktbarrett:
Added a fix for the broken ResolveX documentation.

src/cocotb/types/logic_array.py
```diff
@@ -174,13 +167,27 @@ class LogicArray(ArrayLike[Logic]):
         >>> la
         LogicArray('ZX10', Range(3, 'downto', 0))
 
-    :class:`LogicArray`\ s can be converted into :class:`str`\ s, :class:`int`\ s, or :class:`bytes`\ s.
+        >>> la[:] = 0b0101
+        >>> la
+        LogicArray('0101', Range(3, 'downto', 0))
+
+    :class:`LogicArray`\ s can be converted into its :class:`str` or :class:`int` literal values using casts.
```
cmarqu, line 174:
```suggestion
    :class:`LogicArray`\ s can be converted into their :class:`str` or :class:`int` literal values using casts.
```
?

src/cocotb/types/logic_array.py
```diff
@@ -558,33 +415,33 @@ def from_bytes(
...
-            )
-        return cls.from_unsigned(
-            int.from_bytes(value, byteorder=byteorder, signed=False), range
-        )
+        else:
+            if isinstance(range, int):
+                range = Range(range - 1, "downto", 0)
+            if len(value) * 8 != len(range):
+                raise ValueError(
+                    f"Value of length {len(value)} will not fit in a LogicArray with bounds: {range!r}"
```
cmarqu, line 434:
```suggestion
                    f"Value of length {len(value)} will not fit in a LogicArray with bounds {range!r}"
```
?

src/cocotb/types/array.py
```diff
@@ -14,38 +14,43 @@ class Array(ArrayLike[T]):
...
         Array([], Range(1, 'to', 0))
 
-    Indexing and slicing is very similar to :class:`list`\ s, but it uses the indexing scheme specified.
-    Slicing, just like the :class:`~cocotb.types.Range` object uses an inclusive right bound,
-    which is commonly seen in HDLs.
-    Like :class:`list`\ s, if a start or stop index is not specified, it is inferred as the start or end of the array.
-    Slicing an array returns a new :class:`~cocotb.types.Array` object, whose bounds are the slice indexes.
+    Indexing and slicing is very similar to :class:`list`\ s, but it uses the indexing scheme specified with the *range*.
+    Unlike :class:`list`, slicing uses an inclusive right bound, which is common in HDLs.
+    But like :class:`list`, if a start or stop index is not specified in the slice, it is inferred as the start or end of the array.
```
cmarqu, line 52:
```suggestion
    Unlike :class:`!list`, slicing uses an inclusive right bound, which is common in HDLs.
    But like :class:`!list`, if a start or stop index is not specified in the slice, it is inferred as the start or end of the array.
```
Suppress excessive linking.

src/cocotb/types/array.py
```diff
@@ -14,38 +14,43 @@ class Array(ArrayLike[T]):
 
     Arrays are similar to, but different from Python :class:`list`\ s.
     An array can store values of any type or values of multiple types at a time, just like a :class:`list`.
```
cmarqu, line 16:
```suggestion
    An array can store values of any type or values of multiple types at a time, just like a :class:`!list`.
```
Suppress excessive linking.

cmarqu (approved):

Issue #4506, opened 2025-02-24
Graceful method to test for signal presence


markusdd:
Hi,

while writing a generic testbench component (a waveform log handler which forwards log messages to string signals viewable in the waveform)
I stumbled across a limitation of the SimHandle API, at least so it seems:

```python
class WaveformLogHandler(logging.Handler):
    """Log Handler which forwards messages to waveform signals."""

    def __init__(
        self,
        dut,
        log_dbg: Optional[SimHandle] = None,
        log_inf: Optional[SimHandle] = None,
        log_chk: Optional[SimHandle] = None,
        log_wrn: Optional[SimHandle] = None,
        log_err: Optional[SimHandle] = None,
        log_ech: Optional[SimHandle] = None,
        log_all: Optional[SimHandle] = None,
    ):
        super().__init__()
        self.dut = dut
        logging.getLogger().info("Installing Waveform Logger.")
        a = dut.log_dbg
        try:
            self.log_dbg = log_dbg if isinstance(log_dbg, SimHandleBase) else dut.log_dbg
        except AttributeError:
            self.log_dbg = None
...
```

We have 7 optional signals log_dbg, log_wrn, log_err etc. that if present in the system verilog testbench top should be found by the handler. Then in the emit method of the handler we intercept the logging messages and assign them to the signals.

If all are present this works beautifully.

But if in a given testbench not all of these signals exist we gracefully want to handle this and just not emit to the signal.

In the above example I try this by catching the AttributeError, which works in the sense that the test itself runs PASS.
But our log parsing then gets very upset because for each signal, we get this on the console:

```
INFO     ERROR:          VHPI       NOTOTM                                                                                                                                                                                       
INFO              Cannot iterate on type vhpiInternalRegions using a reference handle of type <unknown>.                                                                                                                         
INFO              /xxxxxxxx/tb.sv, 29:                                                                                                                                                                                                              
INFO     Simulation complete via $finish(1) at time 1 PS + 0                                   
```

Obviously we do not want to filter these messages because in other cases they could represent a real issue.

In a pythonic sense each dictionary has a 'get' method which gracefully returns None without exception.
I have found no such API for cocoTb to test for presence of a signal before trying to use it.

I get the same errors when using the _id("signal_s", extended=False) indirect method. It also thrwos an AttributeError and the above Error is logged, even if caught with try/except.

So long story short: How can we gracefully test for signal presence without causing exceptions?
(I hope this is not an embarrasing oversight while reading the docs xD)

[...]

PR #4517, opened 2025-02-28
provide get_handle_by_name_native function

markusdd:
this allows for queries that explicitly should NOT cross language barries when testing for presence
of a signal

this is to address e.g. #4506, but there have been other tickets targetting similar issues.

Some simulators misbehave badly when searching for optional signal cross-language boundary.
If one e.g. searches for optional hookups up on the toplevel to connect generic TB components, it might not be desirable that there is an automatic search in VHPI when the dut is native Verilog, or VPI when the dut is native VHDL when the signal is not found, but currently this behavior is hardcoded into the GPI.
This new function  explicitly allows to specifiy to NOT search cross-boundary and instead just return None.


src/cocotb/share/include/gpi.h
```diff
@@ -142,6 +142,8 @@ GPI_EXPORT const char *gpi_get_simulator_version(void);
 GPI_EXPORT gpi_sim_hdl gpi_get_root_handle(const char *name);
 GPI_EXPORT gpi_sim_hdl gpi_get_handle_by_name(gpi_sim_hdl parent,
                                               const char *name);
+GPI_EXPORT gpi_sim_hdl gpi_get_handle_by_name_native(gpi_sim_hdl parent,
```
ktbarrett, line 145: I think instead maybe having an argument to `gpi_get_handle_by_name` that you can specify `NATIVE` or `AUTO`, etc would be better. That way we can add additional discovery methods latter if need be.

src/cocotb/handle.py
```diff
@@ -476,8 +476,30 @@ def _child_path(self, name: str) -> str:
     def _sub_handle_key(self, name: str) -> str:
         return name
 
-    def _get_handle_by_key(self, key: str) -> Optional[simulator.gpi_sim_hdl]:
-        return self._handle.get_handle_by_name(key)
+    def _get_handle_by_key(
+        self, key: str, native=False
+    ) -> Optional[simulator.gpi_sim_hdl]:
```
ktbarrett, line 481: Something like this is fine. The additional argument would default to `AUTO` which would pass through to `gpi_get_handle_by_name`. The other discovery methods: `__getitem__`, `_id`, and `__getattr__` would just use such a function with the defaulted `AUTO`. However, I think this might be the wrong function to accomplish that, as it should hook into the sub object cache.

Maybe a new function like `_get` to parallel `dict.get`? Then refactor `__getitem__`, `_id`, and `__getattr__` to use that instead?

src/cocotb/share/include/gpi.h
markusdd: I agree this would be better, I just did'nt really dare to modify the existing signature.

As for NATIVE or AUTO: how would we specify these on the C++ and Python ? Defines? Enum?
Not that experienced in designing Python-C-intrerfaces so some guidance there would be appreciated.

src/cocotb/handle.py
markusdd: I am very much in favor of using _get to mimic dict.get, this is exactly what I had in mind originally when I opened the issue.

the refactoring should be easy to do. 

Same question as in the other comment: how do we want to handle the possible values for the "discovery=" parameter. (is that what we want to call it?)

src/cocotb/share/include/gpi.h
ktbarrett: enums in C, see some examples a few line above in gpi.h. `IntEnum`s in Python, see `_GPISetAction` in handle.py as an example. There's a way to get it all to work together, but we aren't doing that yet, so don't feel you need to be the first.

markusdd: pushed new version doing that.

src/cocotb/handle.py
markusdd: went with implementing _get

markusdd:
ok the existing regression passes, before diving into tests I think this is ready for another review round @ktbarrett

src/cocotb/simulator.pyi
```diff
@@ -50,6 +50,7 @@ class gpi_sim_hdl:
     def get_definition_name(self) -> str: ...
     def get_handle_by_index(self, index: int) -> gpi_sim_hdl | None: ...
     def get_handle_by_name(self, name: str) -> gpi_sim_hdl | None: ...
+    def get_handle_by_name_native(self, name: str) -> gpi_sim_hdl | None: ...
```
ktbarrett, line 53: I don't think this is needed anymore.

src/cocotb/share/lib/simulator/simulatormodule.cpp
```diff
@@ -1292,10 +1305,14 @@ static PyMethodDef gpi_sim_hdl_methods[] = {
...
+     PyDoc_STR("get_handle_by_name($self, name, discovery_method/)\n"
                "--\n\n"
-               "get_handle_by_name(name: str) -> cocotb.simulator.gpi_sim_hdl\n"
-               "Get a handle to a child object by name.")},
+               "get_handle_by_name(name: str, discovery_method: "
+               "cocotb.handle._GPIDiscovery) -> "
+               "cocotb.simulator.gpi_sim_hdl\n"
+               "Get a handle to a child object by name.\n"
+               "Specifiy discovery_method to determine the signal discovery\n"
+               "strategy. AUTO by default.")},
```
ktbarrett, line 1315:
```suggestion
               "Specifiy discovery_method to determine the signal discovery "
               "strategy. AUTO by default.")},
```

src/cocotb/simulator.pyi
markusdd: correct, and the signature needed updating.

markusdd:
added your suggestions and updated the function signature in .pyi to reflect the optional nature of the discovery parameter

markusdd:
added tests, marking as ready. the one failed icarus test seems like a system fail of the pipeline, but I cannot re-trigger.

Let me know if you want anything else changed/added.

markusdd:
Ok now I think we're looking better :-) 

Coverage looks decent as well.

src/cocotb/handle.py
```diff
@@ -210,6 +210,15 @@ def __len__(self) -> int:
 KeyType = TypeVar("KeyType")
 
 
+class _GPIDiscovery(enum.IntEnum):
```
ktbarrett, line 213: I think I missed this before, but `_get` is public, this should be as well.

ktbarrett: Also to get documentation to work you'll probably have to adapt `DocEnum` to an `IntEnum`. I'll link how to do that seeing as your change will probably come in before mine.

ktbarrett (commented):
Looking good. The last step will be documentation. I left a couple comments that should help. And you should add a new feature newsfragment. You can see some documentation on how to do that in `docs/source/newsfragments/README.rst`. There are plenty of examples there as well. One gotcha everyone runs into is that towncrier only supports single line newsfragments. Newlines break everything.

src/cocotb/handle.py
ktbarrett: From #4512 but adapting to `IntEnum` instead of `StrEnum`.

```python
IntEnumT = TypeVar("IntEnumT ", bound=IntEnum)


class DocIntEnum(IntEnum):
    """Like DocEnum but for IntEnum enum types."""

    def __new__(cls: Type[IntEnumT], value: str, doc: Optional[str] = None) -> IntEnumT:
        self = int.__new__(cls, value)
        self._value_ = value
        if doc is not None:
            self.__doc__ = doc
        return self
```

markusdd:
Oo ok that rework somehow bricked the discovery but I am REALLY not sure why.
The new DocIntEnum seems to behave identical to an IntEnum but apparently it does not really

grafik

markusdd:
switched now back to the DocIntEnum but I will be honest: I cannot really figure out why the pipelines fail now.
Must be some stupid little oversight

markusdd:
lesson: don't do stuff like this before the first coffee in the morning xD

forgot to adapt the enum import in the test, should work now...

src/cocotb/handle.py
```diff
@@ -296,6 +303,30 @@ def _discover_all(self) -> None:
...
 
+    def _get(
+        self, key: str, discovery_method: GPIDiscovery = GPIDiscovery.AUTO
+    ) -> Optional[simulator.gpi_sim_hdl]:
+        """Query the simulator for an object with the specified *key*.
+
+        Like Pythons native dict get-function, this returns None if the object
+        is not found instead of raising an AttributeError.
+
+        If extended identifiers are needed simply add a ``\\`` character before and after the name.
```
ktbarrett, line 314:
```suggestion
```
This is only correct for VHDL. Best to just not include it.

docs/source/newsfragments/4517.feature.rst
```diff
@@ -0,0 +1 @@
+:class:`.HierarchyObjectBase` and its inheritants now support :meth:`_get` for a dictionary-style handle retrieval (replaces deprecated :meth:`_id`), which returns :data:`None` on failure instead of raising an exception to aid with optional signal discovery.
```
ktbarrett, line 1:
```suggestion
:class:`.HierarchyObjectBase` and its inheritants now support :meth:`_get` for a dictionary-style handle retrieval, which returns :data:`None` on failure instead of raising an exception to aid with optional signal discovery.
```
Do not speak of the devil (deprecated functionality). It makes it more difficult to remove all the references later and advertises it's existence, potentially leading to new users.

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -539,3 +541,25 @@ def _discover(obj):
...
+    assert dut._get("notexistingsignal", discovery_method=GPIDiscovery.AUTO) is None
+    assert dut._get("notexistingsignal", discovery_method=GPIDiscovery.NATIVE) is None
+
+
+@cocotb.test()
+async def test_native_discovery(dut):
+    """Try accessing a signal using native strategy."""
+    assert dut._get("stream_in_data") is not None
+    assert dut._get("stream_in_data", discovery_method=GPIDiscovery.AUTO) is not None
+    assert dut._get("stream_in_data", discovery_method=GPIDiscovery.NATIVE) is not None
```
ktbarrett, line 565: I'd like to see a test where NATIVE doesn't pick up an object, but AUTO does. Like on a mixed language boundary. That will probably have to be put into `test_iteration_mixedlang`.

ktbarrett:
Looking good. This should be the last set of changes.

src/cocotb/handle.py
markusdd: should this be mentioned nowhere? I just included it because _get will kind of replace _id()

remove it completely or mention it's just for VHDL?

docs/source/newsfragments/4517.feature.rst
markusdd: haha ok

tests/test_cases/test_discovery/test_discovery.py
markusdd: good point. will do.

markusdd: just commited, let's see how this goes.

markusdd:
nice. looks like this passes.

tests/test_cases/test_discovery/test_discovery.py
```diff
@@ -30,8 +30,10 @@
 import pytest
 
 import cocotb
+import cocotb.handle
```
ktbarrett, line 33: Spurious include?

ktbarrett (approved):
One little nitpick.

Ill leave this up for a couple days to give the other maintainers some time to look at it. Will merge both PRs on Monday.

markusdd:
you are correct. very interesting that ruff did not catch this one, seems like a bug to me.
removed it.

src/cocotb/handle.py
```diff
@@ -296,6 +303,28 @@ def _discover_all(self) -> None:
 
         self._discovered = True
 
+    def _get(
+        self, key: str, discovery_method: GPIDiscovery = GPIDiscovery.AUTO
+    ) -> Optional[simulator.gpi_sim_hdl]:
+        """Query the simulator for an object with the specified *key*.
+
+        Like Pythons native dict get-function, this returns None if the object
+        is not found instead of raising an AttributeError.
```
cmarqu, line 312:
```suggestion
        Like Python's native dictionaty ``get``-function, this returns ``None`` if the object
        is not found instead of raising an :exc:`AttributeError`.
```

src/cocotb/handle.py
```diff
@@ -296,6 +303,28 @@ def _discover_all(self) -> None:
...
+
+        Generally, use the `handle[child_name]` syntax instead, unless you have to change the
+        discovery_method or want to check for optional signals.
```
cmarqu, line 315:
```suggestion
        Generally, use the ``handle[child_name]`` syntax instead, unless you have to change the
        *discovery_method* or want to check for optional signals.
```

src/cocotb/handle.py
```diff
@@ -296,6 +303,28 @@ def _discover_all(self) -> None:
...
+
+        :meta public:
+
+        Args:
+            key: The child object by name.
+            discovery_method: Optional selection of discovery strategy. ``GPIDiscovery.AUTO`` by default.
```
cmarqu, line 321:
```suggestion
            discovery_method: Optional selection of discovery strategy. :data:`~cocotb.handle.GPIDiscovery.AUTO` by default.
```

tests/test_cases/test_iteration_mixedlang/test_iteration.py
```diff
@@ -75,3 +76,13 @@ async def recursive_discovery_boundary(dut):
     actual = recursive_dump(dut.i_vhdl, tlog)
     tlog.info("Found a total of %d things", actual)
     assert actual == expected
+
+
+@cocotb.test
+async def discovery_method(dut):
+    """Verify that the different discovery methods work."""
+
+    # this is a verilog toplevel, so we should not be finding a VHDL object
```
cmarqu, line 85:
```suggestion
    # this is a Verilog toplevel, so we should not be finding a VHDL object
```
Very nitpicky, I admit :)

src/cocotb/_utils.py
```diff
@@ -210,6 +210,20 @@ def __new__(cls: Type[EnumT], value: Any, doc: Optional[str] = None) -> EnumT:
         return self
 
 
+IntEnumT = TypeVar("IntEnumT", bound=IntEnum)
+
+
+class DocIntEnum(IntEnum):
+    """Like DocEnum but for IntEnum enum types."""
```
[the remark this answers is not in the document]
markusdd: a thing that will be soon above that piece of code, see suggestions by @ktbarrett

docs/source/newsfragments/4517.feature.1.rst
```diff
@@ -0,0 +1 @@
+:meth:`_get` accepts optional parameter ``discovery_method`` that takes values from :class:`GpiGPIDiscovery` and defaults to ``AUTO``. ``NATIVE`` can be used to forbid cross-language search.
```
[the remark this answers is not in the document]
markusdd: just a hint: this is then true of many of the towncrier rst files, I only copied most of the stuff from there. very few use this full path syntax.

markusdd:
pulled these in (and found the 'add to batch' button to late) xD

cmarqu:
> pulled these in (and found the 'add to batch' button to late) xD

Oh, maybe that's new? I don't remember seeing it...

cmarqu:
Ah, the linking from the release docs primarily does not work because there is nothing to link to - `GpiGPIDiscovery` is not documented yet. Could you add it please, maybe in https://github.com/cocotb/cocotb/blob/f998c45d222e509ab11cf49e6505899e1a6397fe/docs/source/library_reference.rst?plain=1#L313 or thereabout?

markusdd:
tried it just now, but I never worked with this type of rst autodoc before so this is dumb pattern matching, I hope it works^^

docs/source/newsfragments/4517.feature.1.rst
```diff
@@ -0,0 +1 @@
+:meth:`~cocotb.handle.GpiGPIDiscovery._get` accepts optional parameter :data:`~cocotb.handle.GpiGPIDiscovery._get.discovery_method` that takes values from :class:`~cocotb.handle.GpiGPIDiscovery` and defaults to ``AUTO``. ``NATIVE`` can be used to forbid cross-language search.
```
cmarqu, line 1:
```suggestion
:meth:`~cocotb.handle.GPIDiscovery._get` accepts optional parameter :data:`~cocotb.handle.GPIDiscovery._get.discovery_method` that takes values from :class:`~cocotb.handle.GPIDiscovery` and defaults to ``AUTO``. ``NATIVE`` can be used to forbid cross-language search.
```

docs/source/newsfragments/4517.feature.rst
```diff
@@ -0,0 +1 @@
+:class:`.HierarchyObjectBase` and its inheritants now support :meth:`_get` for a dictionary-style handle retrieval, which returns ``None`` on failure instead of raising an exception to aid with optional signal discovery.
```
cmarqu, line 1:
```suggestion
:class:`.HierarchyObjectBase` and its inheritants now support :meth:`~cocotb.handle.GPIDiscovery._get` for a dictionary-style handle retrieval, which returns ``None`` on failure instead of raising an exception to aid with optional signal discovery.
```

cmarqu (commented):
More untested sugggestions :)

docs/source/newsfragments/4517.feature.1.rst
markusdd: oh yeah these are actually typos. good catch

docs/source/newsfragments/4517.feature.rst
markusdd: no this is wrong. _get is a method of HierarchyObjectBase

docs/source/newsfragments/4517.feature.1.rst
markusdd: but also from the wrong object, made a manual change.

markusdd:
this is harder than the actual code lol xD

btw...I just realized...u surely are the Colin that knows Daniel from Dresden?^^

cmarqu:
> this is harder than the actual code lol xD

It can indeed be quite fiddly at times...

> btw...I just realized...u surely are the Colin that knows Daniel from Dresden?^^

Indeed I am, please say hi! :)

markusdd:
will do haha.

as for the docs: as mentioned above this could affect other rsts as well so maybe a seperate PR for that before 1.0 would make more sense to clean this up as a lot.

btw: is there any targeted timeline for 2.0? I was headed to preapre our DB for 2.0 (which led to these PRs) but we surely wanna wait until it's declared stable before we jump onto it. But if there's more finding I'm happy to help...now that I also stuck my head into the GPI...which I never wanted xD

cmarqu:
This appears to link correctly:

```diff
diff --git i/docs/source/library_reference.rst w/docs/source/library_reference.rst
index 2540603f..e1d1a3da 100644
--- i/docs/source/library_reference.rst
+++ w/docs/source/library_reference.rst
@@ -308,11 +308,13 @@ Simulation Object Handles
     :synopsis: Classes for simulation objects.
     :exclude-members: Deposit, Force, Freeze, Release
     :special-members: __len__
-
-.. autoenum:: cocotb.handle.GPIDiscovery
 ..
    Excluding the Assignment Methods that are getting their own section below
 
+.. autofunc:: cocotb.handle.HierarchyObjectBase._get
+
+.. autoenum:: cocotb.handle.GPIDiscovery
+
 .. _assignment-methods:
 
 Assignment Methods
diff --git i/docs/source/newsfragments/4517.feature.1.rst w/docs/source/newsfragments/4517.feature.1.rst
index 2a40908f..3b0622c5 100644
--- i/docs/source/newsfragments/4517.feature.1.rst
+++ w/docs/source/newsfragments/4517.feature.1.rst
@@ -1 +1 @@
-:meth:`~cocotb.handle.HierarchyObjectBase._get` accepts optional parameter :data:`~cocotb.handle.HierarchyObjectBase._get.discovery_method` that takes values from :class:`~cocotb.handle.GPIDiscovery` and defaults to ``AUTO``. ``NATIVE`` can be used to forbid cross-language search.
+:meth:`HierarchyObjectBase._get() <cocotb.handle.HierarchyObjectBase._get>` accepts the optional parameter ``discovery_method`` that takes values from :class:`~cocotb.handle.GPIDiscovery` and defaults to :enum:`~cocotb.handle.GPIDiscovery.AUTO`. To forbid cross-language search, use :enum:`~cocotb.handle.GPIDiscovery.NATIVE`.
diff --git i/docs/source/newsfragments/4517.feature.rst w/docs/source/newsfragments/4517.feature.rst
index e892098f..e468a4a2 100644
--- i/docs/source/newsfragments/4517.feature.rst
+++ w/docs/source/newsfragments/4517.feature.rst
@@ -1 +1 @@
-:class:`.HierarchyObjectBase` and its inheritants now support :meth:`~cocotb.handle.HierarchyObjectBase._get` for a dictionary-style handle retrieval, which returns ``None`` on failure instead of raising an exception to aid with optional signal discovery.
+:class:`.HierarchyObjectBase` and its inheritants now support :meth:`~.HierarchyObjectBase._get` for a dictionary-style handle retrieval, which returns ``None`` on failure instead of raising an exception to aid with optional signal discovery.
```

markusdd:
I do not claim to understand why which syntax works when but I applied those haha.

docs/source/newsfragments/4517.feature.1.rst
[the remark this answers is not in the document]
cmarqu: > just a hint: this is then true of many of the towncrier rst files, I only copied most of the stuff from there. very few use this full path syntax.

Yeah, but we are trying to get better everywhere :)

cmarqu:
We have somehow lost the doc autobuild on PR updates, but I tested it locally.

cmarqu (approved):

cmarqu:
Thanks for the PR!

markusdd:
Thanks for accepting and the review/doc work!

This will help us greatly in dealing with ugly mixed languaage situations in the future and building more generic TB components.

PR #4520, opened 2025-02-28
Undeprecate-len-on-scalars

### LINKED ISSUE #4518: Consider not deprecating len() for Scalars
The new 2.0 API seeks to deprecate the len(function) for scalar signals, as now a split has been made beween these and Array objects.

This makes a `logic [0:0] a;` different from `logic a;`. 

I would argue this makes for a much less intuitive to use API as it becomes assymmetric.

Example:

We have a streaming source TB IP that takes a configurable amount of datalanes which can be the same width or not.

Often times streams can be accompanies by frame or valid signals which just make up one bit.

We use the len() function to get information how wide the datalane signals are to do some width checking and provide additional debug infos. Now, with the new API, if I get a datalane which is just 1 bit, I need to make cumbersome ifinstance checks just to find out if I can call `len` in each and every place.

So I would like to ask to re-dicuss/re-consider this decision? I think a more symmetrical API would be significantly easier to use, there is nothing wrong with just returning '1'.

Or: Is there a better new way to accomplish getting signal width which I have overlooked?

markusdd:
closes #4518

main rationales are:

- API unsymmetry compared to single-bit vectors
- unnecessary code complication in generic testbench code which now would reuqire isinstance calls to get a simple '1' width
- a lot of existing code breakages - we have over 500 places alone that would now require 'isinstance' checks
- keeping it does not hurt anything, it's a simple return 1

### DIFF
--- a/docs/source/newsfragments/4318.change.rst
+++ b/docs/source/newsfragments/4318.change.rst
@@ -1,1 +1,1 @@
-Scalar logic objects, such as un-arrayed ``logic`` in Verilog and ``std_logic`` in VHDL, no longer support ``handle.range`` or ``len(handle)``. Additionally, ``handle.value`` on such subject objects now returns a :class:`.Logic` rather than a :class:`.LogicArray`.
+For scalar logic objects, such as un-arrayed ``logic`` in Verilog and ``std_logic`` in VHDL, ``handle.value`` now returns a :class:`.Logic` rather than a :class:`.LogicArray`.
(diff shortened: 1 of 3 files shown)


markusdd:
sorry for re-open had to correct the branch quickly, there were some unintended commits on there

ktbarrett:
There's also 4318.change.rst that mentions this is deprecated.

markusdd:
fixed.

ktbarrett:
And the test in test_deprecated. Forgot about that =D

markusdd:
yes I saw the dreaded red bubbles coming in hahaha, will fix soon.

ktbarrett (approved):

cmarqu (approved):

PR #4524, opened 2025-03-01
Cancelled error

### LINKED ISSUE #2679: Add `cancel()` to `RunningTask` and deprecate `kill()`

ktbarrett:
Depends on #4514. Closes #2679.

### TODO
- [x] tests
- [x] newsfrags

### Follow-ons
* Deprecate `Task.kill` and update all uses in repo
* Fix `First`, `Combine`, and the GPI `Clock` to cleanup up with CancelledError
* Make tests consider `CancellationError` at test end

### DIFF
--- a/docs/source/newsfragments/4524.change.rst
+++ b/docs/source/newsfragments/4524.change.rst
@@ -1,0 +1,1 @@
+:meth:`.Task.cancel` now causes a :exc:`~asyncio.CancelledError` to be thrown into the coroutine. This behavior is *scheduled*, so Tasks will not become cancelled immediately.
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -61,7 +61,6 @@ class Task(Generic[ResultType]):
         Use :meth:`result`, :meth:`done`, and :meth:`done` methods instead, respectively.
     """
 
-    _name: str = "Task"  # class name of schedulable task
     _id_count = 0  # used by the scheduler for debug
 
     def __init__(self, inst):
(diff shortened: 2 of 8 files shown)


ktbarrett:
The C++ clock is segfaulting in Verilator. Removing the call to `vpi_remove_cb` so it always just sets `m_removed` fixes the issue. So what's probably happening is that Verilator is maintaining the value change callbacks as singletons and we have both cocotb and the GPI clock both firing at the same time. So the value change events are queued up, then we remove the cb which cleans up the data needed for the second callback.

I'm going to have to disable the C++ clock for Verilator for now. I'm not going to let an optimization get in the way of this change.

ktbarrett:
Remove GPI Clock exemption. Add may-fail to all Verilator CI.

docs/source/newsfragments/4524.change.rst
```diff
@@ -0,0 +1 @@
+:meth:`.Task.cancel` now causes a :exc:`~asyncio.CancelledError` to be throw into the coroutine. This is behavior is *scheduled*, so Tasks will not become cancelled immediately.
```
cmarqu, line 1:
```suggestion
:meth:`.Task.cancel` now causes a :exc:`~asyncio.CancelledError` to be thrown into the coroutine. This behavior is *scheduled*, so Tasks will not become cancelled immediately.
```

src/cocotb/task.py
```diff
@@ -34,6 +38,25 @@
...
+    def __init__(self, msg: str, outcome: Outcome[Any]) -> None:
+        super().__init__(msg)
+        self.outcome = outcome
+
+
+class _TaskState(DocEnum):
+    """State of a Task."""
+
+    UNSTARTED = (auto(), "Task created, but never run and not scheduled")
+    SCHEDULED = (auto(), "Task in Scheduler queue to run soon")
```
cmarqu, line 53:
```suggestion
    SCHEDULED = (auto(), "Task in Scheduler queued to run soon")
```
?

src/cocotb/task.py
```diff
@@ -107,93 +133,129 @@ def _get_coro_stack(self) -> Any:
...
+        Can optionally throw an Exception into the Task.
 
         Args:
-            outcome: The :any:`outcomes.Outcome` object to resume with.
+            exc: :exc:`BaseException` to throw into the coroutine or nothing.
 
         Returns:
-            The object yielded from the coroutine or None if coroutine finished
-
+            The object yielded from the coroutine or None if coroutine finished.
```
cmarqu, line 189:
```suggestion
            The object yielded from the coroutine or ``None`` if coroutine finished.
```

src/cocotb/task.py
ktbarrett:
```suggestion
    SCHEDULED = (auto(), "Task queued to run soon")
```

cmarqu (approved):

Issue #4530, opened 2025-03-04
Tests end with:  ERROR    gpi     VPI: Failed to end simulation


markusdd:
Hi guys.

While test driving the newest master branch after all the PRs I noticed that this also happened back when we worked on them, this seems to happen consistently.

All tests seem to end with the line in the title, I assume you already know this?

Tried to find an issue addressing this but did not find one. Is this knows and addressed somewhere already? If yes we can also just close.

ktbarrett:
IIRC you are using Xcelium, correct? We have been seeing that as well, it seems to be an upstream issue. https://github.com/cocotb/cocotb/issues/4484.

markusdd:
ok I see...more incorrect API implementations by comapnies that charge 20k/y per License xD

could we consider a switch to turn the messages off somehow? I think for development etc. they are fine, but having basically each and every test end with an error will raise eyebrowes^^

ktbarrett:
Right... Maybe this stuff should also be included in the LOG_DEBUG level so people only see it when they need to figure out why cocotb isn't working correctly. Should be as simple as making the LOG_WARN into LOG_DEBUG.

markusdd:
made a small PR

markusdd:
btw it even logs as error right now, not warning.

when you say not working correctly: is there any actual adverse effect from that that has been observed?

ktbarrett:
If it works, it works. If it doesn't work, cocotb tries to make it work by using things like `m_removed` to prevent calling up on cancelled callbacks. But it will leak GPI objects out the wazoo. And might just be leaking the simulators VPI objects as well, depending on whether it's actually failing or not.

markusdd:
I see, Lovely situation xD

ktbarrett:
For example of the worst case scenario, see #4379.

markusdd:
ok...that...isn't nice xD

One really should assume they somehow have this figured out by now but hey...constructs that have been intriduced in 2005 still do not work 20 years later so here we are I guess :)

ktbarrett:
I'm not against adding some hardcodes in there too if it turns out to be a big problem: out-of-memory or massive slowdowns. If Xcelium is just returning the wrong error code, but *actually* removing the callback handle, we can just delete the GPI callback handle. If they are failing to remove the callback, we can't delete it or the next time it fires we'll get a use-after-free and probably a segfault.

markusdd:
resolved by #4531

PR #4531, opened 2025-03-04
make vpi vhpi sim end failed logs debug level

markusdd:
move those logs to debug as per issue #4530


markusdd (commented):
should we do the same to 'unable to remove callback'?

There seems to be a lot of spam regarding that as well:
 
grafik

markusdd:
and these ones we have as well: 
grafik

ktbarrett:
That iterate message can definitely be a debug. That's not even an error, it's just expected that iterators sometimes are empty.

That got me wondering why it's even trying VPI. I forgot about this...
> \# Loading the VHPI library causes an error, so we always load the VPI library and supply
> \# GPI_EXTRA=$(shell $(PYTHON_BIN) -m cocotb_tools.config --lib-name-path vhpi xcelium) if needed.

I'm sure there's a better way. Like we could enter from VPI and then only register the VHPI? Or if the issue is registering startup and shutdown callbacks enter with the VPI, register startup and shutdown, then do discovery with VHPI only at the top? This is part of the reason I really wish the GPI libraries were set up per-simulator rather than per-interface. Maybe we'll get there in a couple years =/

markusdd:
I mean essentially it is prudent to assume that most compile flows that do actual bigger work with cocotb will usually load both interfaces anyway (we do as well). cocotb-test also defaults to this.

Question is: what do we do with this now? Demote it to debug? Go into the C and change the behavior entirely when this is being logged? difficult.

ktbarrett:
Yeah it can be a DEBUG. If that's the last of what you are seeing we can get this approved and in.

markusdd:
Yeah I guess we're going to have a lot more fun going forward as you indicated.
I've just seen this one in the CIs here as we ourselves do not use VHDL tops, but when we're at it it might be good to be consistent.

Will quickly do this in the lunch break now^^

ktbarrett (approved):

cmarqu (approved):

markusdd:
thanks guys 🥳

PR #4529, opened 2025-03-04
Update CI

ktbarrett:
See commits.

Issue #4526, opened 2025-03-01
[Verilator] calling `vpi_remove_cb` on a timing trigger that will fire in the same cycle causes segfault


ktbarrett:
From [here](https://github.com/cocotb/cocotb/actions/runs/13608459585/job/38042725933?pr=4524#step:24:2555). The following test reproduces.

```python
async def test_clocks_with_other_number_types(dut):
    clk1 = cocotb.start_soon(Clock(dut.clk, decimal.Decimal("1"), unit="ns", impl="gpi").start())
    await Timer(10, "ns")
    clk1.cancel()
```

The `Timer(10, "ns")` and a clock edge from the GPI clock will occur at the same simulation phase. This `Timer(10, "ns")` will wake up first and the Clock Task will be cancelled (equivalent to calling `Clock.stop`). This calls `GpiClk::stop`, which calls `gpi_remove_cb`, which calls `vpi_remove_cb` where the segfault occurs.

Removing the `vpi_remove_cb` and setting `m_removed` instead allows the test to work.

ktbarrett:
This isn't just GpiClk, but when any two `Timer` callbacks occur in the same cycle. https://github.com/cocotb/cocotb/pull/4532#issuecomment-2702109848. We've not run into this until after #4392 because we were just leaking GPI callback objects before then lmao.

ktbarrett:
Upstream issue closed.

PR #4532, opened 2025-03-05
Fix converting log level to string.

gilbertoabram:
Enabling this output causes the simulator to crash due to trying to access the log level as a pointer.

Is it an option to use the `format` attribute for `gpi_log_` when the compiler supports it to get warnings?

### DIFF
--- a/src/cocotb/share/lib/vhpi/VhpiImpl.h
+++ b/src/cocotb/share/lib/vhpi/VhpiImpl.h
@@ -88,8 +88,9 @@ static inline void __check_vhpi_error(const char *file, const char *func,
     }
 
     LOG_EXPLICIT("gpi", GPI_DEBUG, file, func, line,
-                 "VHPI Internal Error: %s @ %s:%d: %s", gpi_log_level(loglevel),
-                 info.file, info.line, info.message);
+                 "VHPI Internal Error: %s @ %s:%d: %s",
+                 gpi_log_level_to_str(loglevel), info.file, info.line,
+                 info.message);
 }
 
 #define check_vhpi_error()                                \
--- a/src/cocotb/share/lib/vpi/VpiImpl.h
+++ b/src/cocotb/share/lib/vpi/VpiImpl.h
@@ -82,8 +82,9 @@ static inline void __check_vpi_error(const char *file, const char *func,
     }
 
     LOG_EXPLICIT("gpi", GPI_DEBUG, file, func, line,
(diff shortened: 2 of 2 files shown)


ktbarrett (approved):
Whoops. ~~The danger of poorly named functions.~~ Apparently it's just a cast? I figured the compiler would see that and complain. Maybe we need something to get printf-style function checking to work.

ktbarrett:
Could this be https://github.com/cocotb/cocotb/issues/4526?

gilbertoabram:
> Could this be #4526?

That does sound very similar indeed. One difference is that in issue_348 the clock is implemented via Timer, so I believe it should also be reproducible with just two timers (at least with my setup and for some tests).

gilbertoabram:
Allright, so at least in my case, cocotb correctly calls `vpi_remove_cb` for the second timer, and it succeeds (returns 1), so we delete the cb obj. but later on verilator still calls the callback :(

```
-V{t1,1561}- vpi: callCbs reason=9
-V{t1,1562}- vpi: reason_callback reason=9 id=4
HANDLING: 0x555555bd6610
REMOVED: 0x55555584f540
-V{t1,1563}- vpi: vpi_remove_cb 0x555555a42b28
REMOVED ERR: 1
REMOVED: 0x555555a44d80
-V{t1,1564}- vpi: vpi_remove_cb 0x555555ad6128
REMOVED ERR: 1
   100.00ns INFO     cocotb.regression                  issue_348.issue_348_rising passed
   100.00ns INFO     cocotb.regression                  running issue_348.issue_348_falling (2/3)
                                                            Start two monitors on FallingEdge
-V{t1,1565}- vpi: vpi_register_cb reason=9 id=56 time=100001 obj=(nil)
REGISTERED: 0x555555c59e20
-V{t1,1566}- vpi: vpi_remove_cb 0x555555aef8c8
-V{t1,1567}- vpi: reason_callback reason=9 id=53
HANDLING: 0x555555a44d80
```
(I added some printouts, 0x555555a44d80 is the offending cb obj)

one fix would be to set `m_removed` as mentioned in #4526, however that would need to be done only if it was about to fire, otherwise we'll be leaking the cb objects if they never get called back.

ktbarrett:
Well this isn't breaking anything, so after #4536 comes in which adds "may-fail" to our CI, this can come in.

cmarqu:
> after #4536 comes in which adds "may-fail" to our CI, this can come in.

#4536 is closed unmerged?

ktbarrett:
#4529 is replacement. Apologize for all the shuffling.

PR #4538, opened 2025-03-07
Use NVC as default sim in CI

ktbarrett:
Uses pyenv for installing Python 3.6 and 3.7. Changes simulator used in Python version tests and clang test to use NVC. Then updates those tests to run on Ubuntu 22.04.

* The first commit was added when attempting to fix Icarus 11 in CI
* The second was to ensure, at least visually, that the Python version being run was what being requested as I ran into issues with that when adding the pyenv installer


ktbarrett:
Ugh... Questa is stack-tracing on master.

PR #4541, opened 2025-03-11
Final Handle API change

### LINKED ISSUE #4491: Improve simulation object typing
The base classes are *not* particularly useful for typing. `SimHandleBase` says nothing about the actual object and no one is going to program useful programs against it. The inheritance here is more for reuse than for creating abstract interfaces. The same could be said for `HierarchyObjectBase`, `ValueObjectBase`, and `NonIndexableValueObjectBase`.

`HierarchyObject` and `HierarchyArrayObject` also have dubious utility in typing as well as they offer no knowledge about the child handles, which is what makes those types useful at all. You could add subtypes which describe the children, but that requires downcasting the types to a non-instantiable, non-concrete subtype, which is just weird.

Users cannot, or should not, instantiate them, so their only utility is for typing, and as established, their utility for typing is pretty bad; ultimately, my argument is that these types have no utility being public.

The concrete types do have utility being public as the functionality they provide is used directly. Hierarchy objects should probably be moved to being `Protocol`s which can be extended to describe child fields. Programming against Protocols also allows for ad-hoc objects to take their place.

### Summary

* Everything that isn't a leaf ValueObject type should be made private
* ~~Possibly ValueObject types should be split into a const and non-const version. We could even support that at runtime by selecting which type a handle is constructed into by first checking the `is_const` property~~
* To fix the documentation, instead of documenting the entire object hierarchy, we should only document the concrete types and tell Sphinx to list the inherited public members
* ~~Maybe we can provide Protocol base classes for HierarchyArrayObject and HierarchyObject so users can subclass them to make new Protocol classes with the children listed, but it also contains the other attributes like `_path`, `_name`, etc.~~

ktbarrett:
Closes #4491.

Fixes the documentation by removing the out-of-date hierarchy figure and refactoring the documentation to mention less about inheritance. Makes some of the intermediate classes in the object hierarchy private since they aren't instantiable and the user will never see them.

~~Also changed the API of `_get()` to hook the function into the child object cache like I mentioned. The function now returns SimHandleBase objects instead of GPI handles, which is better for most users.~~ Split off to #4553.

## TODO
- [x] newsfragments
- [x] touch up documentation


ktbarrett:
Something is very broken here and I don't know what.

ktbarrett:
Don't include `ValueObjectBase` methods in subclasses. Also document inheritance using `:show-inheritance`. I think it only shows inheritance of documented classes, so we won't see intermediate classes.

ktbarrett:
Nope =/ Apparently `:show-inheritance:` still tries to list undocumented classes and it doesn't take a list to be explicit about what inheritance to show. I guess I'll just have to explicitly list what classes have members it inherit from to prevent repeating members 100 times.

imphil (approved):

PR #4542, opened 2025-03-13
Fix Combine and First leaking Tasks if cancelled

### LINKED ISSUE #3441: First and Combine can leak Tasks when the waiting Task is cancelled

ktbarrett:
Closes #3441.

If a Task awaiting a Combine or First is cancelled, the waiter tasks weren't cleaned up. Now they are. Also fixed a small bug in First where if a waiter task was cancelled it would continue executing the `on_done` callback causing another exception to bubble up through the scheduler.

### DIFF
--- a/docs/source/newsfragments/4542.bugfix.rst
+++ b/docs/source/newsfragments/4542.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug where :meth:`cancelling <cocotb.task.Task.cancel>` a Task that is awaiting a :class:`.First` or :class:`.Combine` doesn't cancel the underlying waiter Tasks, leading to lower performance or concealing test failures.
--- a/src/cocotb/_extended_awaitables.py
+++ b/src/cocotb/_extended_awaitables.py
@@ -129,7 +129,7 @@ class Combine(_AggregateWaitable["Combine"]):
                     completed.append(task)
                     if len(completed) == len(waiters):
                         done.set()
-                        return
+                    return
                 e = task.exception()
                 if e is not None:
                     nonlocal exception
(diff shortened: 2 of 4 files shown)


docs/source/newsfragments/4542.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix bug where :meth:`cancelling <cocotb.task.Task.cancel>` a Task awaiting on a :class:`.First` or :class:`.Combine` doesn't cancel the underlying waiters leading to reduction in performance or concealling test failures.
```
cmarqu, line 1:
```suggestion
Fix a bug where :meth:`cancelling <cocotb.task.Task.cancel>` a Task awaiting on a :class:`.First` or :class:`.Combine` doesn't cancel the underlying waiters leading to reduction in performance or concealing test failures.
```

cmarqu: Resolved but not applied, was this intentional?

ktbarrett: Yeah I had fixed it on one computer, marked it resolved, but forgot to push, switched to another computer and pushed again without the changes. One more time...

cmarqu (approved):

PR #4543, opened 2025-03-13
Use ValueError instead of OverflowError in sim objects

### LINKED ISSUE #4124: Cleanup exception types

ktbarrett:
Closes #4124. Now there is no confusion on when to use `ValueError` or `OverflowError`. It's always `ValueError`.

### DIFF
--- a/docs/source/newsfragments/4543.change.rst
+++ b/docs/source/newsfragments/4543.change.rst
@@ -1,1 +1,1 @@
-Settings values too large for :class:`~cocotb.handle.IntegerObject`, :class:`~cocotb.handle.EnumObject`, and :class:`~cocotb.handle.LogicArrayObject` now raise :exc:`ValueError` instead of :exc:`OverflowError`.
+Setting values too large for :class:`~cocotb.handle.IntegerObject`, :class:`~cocotb.handle.EnumObject`, and :class:`~cocotb.handle.LogicArrayObject` now raises :exc:`ValueError` instead of :exc:`OverflowError`.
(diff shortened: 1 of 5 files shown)


docs/source/newsfragments/4543.change.rst
```diff
@@ -0,0 +1 @@
+Settings values too large for :class:`~cocotb.handle.IntegerObject`, :class:`~cocotb.handle.EnumObject`, and :class:`~cocotb.handle.LogicArrayObject` now raise :exc:`ValueError` instead of :exc:`OverflowError`.
```
cmarqu, line 1:
```suggestion
Setting values too large for :class:`~cocotb.handle.IntegerObject`, :class:`~cocotb.handle.EnumObject`, and :class:`~cocotb.handle.LogicArrayObject` now raise :exc:`ValueError` instead of :exc:`OverflowError`.
```
?

cmarqu (approved):

PR #4547, opened 2025-03-14
Actually run these tests =/

ktbarrett:
Getting these files mypy'd would be awesome as it would catch this.


ktbarrett:
This is the third time I've seen Questa fail during test_kill_sim with a stack trace. It doesn't occur every time, only some of the time. Usually a rerun solves the issue. My instincts say it has something to do with randomization. This needs investigation.

cmarqu (approved):

PR #4548, opened 2025-03-15
Fix Unpacked arrays with Verilator

ktbarrett:
This fixes unpacked array range access with Verilator and updates the latest and minimum version up to the latest. This will be needed to be updated again after 5.036 drops to include the upstream fix for #4526.


ktbarrett:
Looks like Questa and VCS still need the weird pseudo-region access crap.

cmarqu (approved):

PR #4551, opened 2025-03-17
Improve handling of weak values in Logic/LogicArray

### LINKED ISSUE #4550: Discussion: bool cast of weak drivers
Image

currently the 2.0 style Logic implementation only allows casts to boolean for 1/0 values.

W/L/H leads to a ValueError. For W this makes sense as it is essentially a 'weak X', but L and H are perfectly valid high and low values, usually encountered in pad logic or other scenarios involving Pulls or Open Drain scenarios (think I2C)

Hence the question: has it been discussed whether these should cast to a valid boolean?

ktbarrett:
Closes #4550. @markusdd 

I made `bool(Logic)` work like ~~it does in VHDL 2019 where 1 and H are `True` and everything else is `False`~~ `int(Logic)`. I made `int(Logic)` also respect weak values since we are doing that in LogicArray with the `int` cast. And since we unconditionally resolve weak values I moved the L/H resolving up so it could cache the values in LogicArray if the value is only `01LH`.

`LogicArray.__bool__` was moved to DeprecatedWarning because instead of keeping it and changing the behavior, I'd like to remove it. It behaves as it did for 1.9 in BinaryValue, so it's a deprecated backwards-compatible interface. VHDL doesn't define this operation and what should occur when non-0/1/L/H values are seen is debatable. I will refuse the temptation to guess.

### TODO
- [x] newsfrags

### DIFF
--- a/docs/source/newsfragments/4551.change.1.rst
+++ b/docs/source/newsfragments/4551.change.1.rst
@@ -1,0 +1,1 @@
+:class:`bool(Logic) <cocotb.types.Logic>` casts now treat ``H`` as ``True`` and ``L`` as ``False`` instead of raising :exc:`ValueError`.
--- a/docs/source/newsfragments/4551.change.2.rst
+++ b/docs/source/newsfragments/4551.change.2.rst
@@ -1,0 +1,1 @@
+:meth:`.LogicArray.is_resolvable` now returns ``True`` for arrays containing ``H`` and ``L`` values.
--- a/docs/source/newsfragments/4551.change.rst
+++ b/docs/source/newsfragments/4551.change.rst
@@ -1,0 +1,1 @@
+:class:`int(Logic) <cocotb.types.Logic>` casts now treat ``H`` as ``1`` and ``L`` as ``0`` instead of raising :exc:`ValueError`.
--- a/src/cocotb/types/logic.py
+++ b/src/cocotb/types/logic.py
@@ -87,7 +87,7 @@ class Logic:
 
     .. note::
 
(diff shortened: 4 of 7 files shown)


markusdd:
is this really what we want?

I feel raising the error in case of X/Z/W/U actually was good to be honest, because these are not valid logical values in a boolean sense. (caught a bug with this just the other day because a error response line was X, not zero)

At least my initial idea was more targeted towards the actual valid values. I think including the invalid ones and just assuming False is bad from a verification perspective.

ktbarrett:
I suppose you are right. The VHDL behavior seems a bit odd to me. I feel like it was done mostly due to screeching from Verilog users. It also doesn't parallel `__int__`. `bool` is a subtype of `int`, so the behavior of `bool(logic)` should be a subset of `int(logic)`.

I'll remove it.

markusdd:
glad I'm not the only one who considers VHDL behavior a bit 'odd' at times haha

makes sense.

ktbarrett:
Apparently we *do* need both `__index__` and `__int__`, `int` calling `__index__` is a Python 3.8+ behavior...

docs/source/newsfragments/4551.change.1.rst
```diff
@@ -0,0 +1 @@
+:class:`bool(Logic) <cocotb.types.Logic>` casts now treat ``H`` as ``True`` and ``L`` ``False``, respectively, instead of raising :exc:`ValueError`.
```
cmarqu, line 1:
```suggestion
:class:`bool(Logic) <cocotb.types.Logic>` casts now treat ``H`` as ``True`` and ``L`` as ``False``, respectively, instead of raising :exc:`ValueError`.
```

cmarqu (approved):

PR #4553, opened 2025-03-17
Plumb `HierarchyObject._get` into child object caching

ktbarrett:
@markusdd 

This make `_get` first try to get the handle from the child object cache, then constructs any found object into a `SimHandle` and places it in the cache for later.

If you need the previous behavior, you can do `sim_object._handle.get_handle_by_name(key, discovery_method)`. Yes that's reaching into a private object, but the `_handle` attribute has been stable and isn't expected to change ever.


markusdd:
I think trying the cache first is fine.
Usually this will default to the "old behavior" anyway as optional signal discovery usually will anyway take place right in the beginning when setting up the testbench. (unless something triggers a complete discovery by other means).

Makes sense, we should use caching whenever possible for speed.

cmarqu (approved):

PR #4559, opened 2025-03-18
Introduce `EmptyTrigger` to replace some uses of `NullTrigger`

### LINKED ISSUE #4493: Replace `NullTrigger`
Current suggestions are `RunScheduledTasks`, `Reschedule`, or `Nothing` (may conflict with).

The current main use is to replace the old `fork` behavior of starting the Task immediately. We should instead create a `TaskStarted` Trigger/Event for that. With that taken care of, this is really an "expert" feature. One legitimate use is for user to build abstract functions which return a `Trigger` and they need a Trigger which does nothing. So names more like `NullTrigger` make more sense to me, while the other ones imply the point is to put the current task at the end of the Queue, which has a *less obvious* meaning.

ktbarrett:
Closes #4493. I think we decided not to rename NullTrigger? If we had not decided, let this PR be the battleground.
Depends on #4558.

### TODO
- [x] newsfrags

### DIFF
--- a/src/cocotb/_base_triggers.py
+++ b/src/cocotb/_base_triggers.py
@@ -143,16 +143,15 @@ class Event:
             e = Event()
 
 
-            async def task1():
+            async def coro1():
                 await e.wait()
                 print("resuming!")
 
 
-            cocotb.start_soon(task1())
-            # do stuff
+            task1 = cocotb.start_soon(coro1())
             e.set()
-            await NullTrigger()  # allows task1 to execute
-            # resuming!
+            await task1  # allows task1 to execute
+            # "resuming!"
 
     .. versionremoved:: 2.0
 
@@ -213,7 +212,7 @@ class Event:
         call :meth:`clear`.
         """
         if self._fired:
-            return NullTrigger(name=f"{str(self)}.wait()")
+            return EmptyTrigger()
         return _Event(self)
 
     def clear(self) -> None:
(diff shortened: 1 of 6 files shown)


src/cocotb/_base_triggers.py
```diff
@@ -413,10 +412,99 @@ async def __aexit__(self, *args: Any) -> None:
...
+    def __await__(self) -> Generator["Self", None, "Self"]:
+        yield from []  # Forces this to be a generator
+        return self
+
+    def __repr__(self) -> str:
+        return f"<{type(self).__qualname__} at {pointer_str(self)}>"
+
+
+class NullTrigger(Trigger):
+    """Fires after all current scheduled Tasks are resumed.
```
cmarqu, line 432:
```suggestion
    """Fires after all currently scheduled Tasks are resumed.
```
?

src/cocotb/_base_triggers.py
```diff
@@ -413,10 +412,99 @@ async def __aexit__(self, *args: Any) -> None:
...
+
+    This exact firing of this Trigger with respect to any other Trigger or Task is not deterministic,
```
cmarqu, line 434:
```suggestion
    The exact firing of this Trigger with respect to any other Trigger or Task is not deterministic,
```
?

src/cocotb/_base_triggers.py
```diff
@@ -413,10 +412,99 @@ async def __aexit__(self, *args: Any) -> None:
         self.release()
 
 
-class NullTrigger(Trigger):
+class EmptyTrigger(Trigger):
     """Fires immediately.
 
-    This is primarily for forcing the current Task to be rescheduled after all currently pending Tasks.
+    Mostly useful for building higher-order function which need to take or return Triggers.
```
cmarqu, line 418:
```suggestion
    Mostly useful for building higher-order functions which need to take or return Triggers.
```

cmarqu (approved):

PR #4561, opened 2025-03-18
Deprecate Event name

ktbarrett:
This is not supported by `asyncio.Event` and doesn't provide "functional" utility. This depends on #4559.

### TODO
- [x] newsfrag
- [x] deprecation test

### DIFF
--- a/docs/source/newsfragments/4561.removal.rst
+++ b/docs/source/newsfragments/4561.removal.rst
@@ -1,0 +1,1 @@
+Deprecated :attr:`.Event.name` and passing a *name* argument to :class:`.Event` constructors. If you need to associate a name with an Event, create an empty subclass of :class:`!.Event` with your desired name.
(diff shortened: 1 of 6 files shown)


docs/source/newsfragments/4561.removal.rst
```diff
@@ -0,0 +1 @@
+Deprecated :attr:`.Event.name` and passing a *name* argument to :class:`.Event` constructors. If you need to associate a name with an Event, create an empty subclass of :class:`!.Event` with your desired name.
```
cmarqu, line 1:
```suggestion
Deprecated :attr:`.Event.name` and passing a *name* argument to :class:`.Event` constructors. If you need to associate a name with an Event, create an empty subclass of :class:`!.Event` with your desired name attribute.
```
?

ktbarrett: Well the name is the class name, not an attribute.

cmarqu: Oh, yes, of course.

cmarqu (approved):

PR #4568, opened 2025-03-22
Support negative values with LogicArray

ktbarrett:
Rationale being that `~0`, because Python ints are arbitrary in size, is returned as a negative value assuming two's complement. We would otherwise not be able to support negated values in comparison or construction.

Also removes outdated newsfrags.

### DIFF
--- a/docs/source/newsfragments/3792.change.rst
+++ b/docs/source/newsfragments/3792.change.rst
@@ -1,1 +1,0 @@
-When constructing a :class:`~cocotb.types.LogicArray`, if the given value cannot fit in the given range, an :exc:`OverflowError` is now thrown, instead of a :exc:`ValueError`.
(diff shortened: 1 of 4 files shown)


ktbarrett:
One issue with this rationale is:
```python
LogicArray(0xFF, 8)  # okay!
LogicArray(~0xFF, 8)  # ValueError, ~0xFF is -256, not 0
```

Integers really shouldn't be used for this kind of stuff... no inherent length is a huge PITA.

We could expand the range of acceptable values to `[-2**width, 2**width)` as `-2**width` is the smallest a negated maximum value unsigned value can be. The constructor taking ints is *not* for constructing from integers really, it's for integer literals: `0b10101010`.

ktbarrett:
I *really* am not sure about this. There ends up being 2 integers that any LogicArray compares with, one positive and one negative. Also, going back to the constructor taking both positive and negative values makes the new functions pointless, so we should probably remove them. And that breaks people already integrating against 2.0. Also 1.9 `BinaryValue` only supported comparison with unsigned values, so I'm not sure what got me so worked up about this.

I'm going in a different direction so I'm going to close this.

PR #4573, opened 2025-03-22
Deprecate `Task.kill`

### LINKED ISSUE #4556: Deprecate `task.kill()`
Second half of #2679 that hasn't been done. Blocked by #4554. That is a major behavior change that will need to be handled before we can remove all internal uses of `task.kill()` and finally deprecate it.

ktbarrett:
Closes #4556. Blocked by #4574. Everything necessary to replace `Task.kill` with `Task.cancel` internally outside of doing so at test end (different PR).

### TODO
- [x] newsfrags
- [x] Fix deprecated uses of `Clock.start()` Task killing.
- [x] deprecate `Task.kill`

### DIFF
--- a/tests/test_cases/test_cocotb/test_scheduler.py
+++ b/tests/test_cases/test_cocotb/test_scheduler.py
@@ -33,25 +33,38 @@ from cocotb.triggers import (
 
 LANGUAGE = os.environ["TOPLEVEL_LANG"].lower().strip()
 
-test_flag = False
-
-
-async def clock_yield(task):
-    global test_flag
-    await task
-    test_flag = True
-
 
 @cocotb.test()
 async def test_task_kill(dut):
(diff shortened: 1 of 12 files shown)


imphil (approved):

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -34,25 +34,38 @@
...
+    async def coro():
+        await Timer(1000, "ns")
+
+    task = cocotb.start_soon(coro())
+    await Timer(1)
+    waiter_task = cocotb.start_soon(waiter(task))
+    await Timer(1)
+    task.cancel()
+
+    # task should be cancelled, but clock_yield should continue
```
ktbarrett, line 60:
```suggestion
    # task should be cancelled, but waiter_task hasn't resumed yet
```

PR #4576, opened 2025-03-24
Shorten license headers

imphil:
Use new-style license headers across the whole code base, which rely on
the SPDX identifier and a reference to the LICENSE file for the full
license text.

Keep existing Copyright lines untouched.


ktbarrett (approved):
Wooo!

PR #4578, opened 2025-03-25
Update Makefile.icarus (add -M to make debug recipe)

xwzheng1020:
I found that it can directly enter the gdb debugging console with `make debug`, but it seems that the command is missing the `-M` option


ktbarrett:
What does `-M` do and how is it different from `-m` which is already present in both the regular and debug targets?

PR #4580, opened 2025-03-25
Fix bounds check in LogicArray's from_signed()

gilbertoabram:
There are some issues with the current bounds check when converting signed integers to LogicArrays, where integers in
[-2^n,-2^(n-1)) and [2^(n-1),2^n) pass but shouldn't, and get converted to an incorrect two's complement binary number.

### DIFF
--- a/src/cocotb/types/logic_array.py
+++ b/src/cocotb/types/logic_array.py
@@ -391,6 +391,8 @@ class LogicArray(ArrayLike[Logic]):
         """
         if isinstance(range, int):
             range = Range(range - 1, "downto", 0)
+        elif not isinstance(range, Range):
+            raise TypeError("Invalid range type")
 
         limit = 1 << (len(range) - 1)
         if (value < -limit) or (value >= limit):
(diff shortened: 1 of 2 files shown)


src/cocotb/types/logic_array.py
```diff
@@ -391,15 +391,16 @@ def from_signed(
         """
         if isinstance(range, int):
             range = Range(range - 1, "downto", 0)
-        if value < 0:
-            value += 2 ** len(range)
-        # If value doesn't fit in range, it will still be negative and will blow the
-        # constructor up in a bad way.
-        if value < 0:
+        elif not isinstance(range, Range):
+            raise TypeError("Invalid range type")
```
ktbarrett, line 395: To mirror the check on `__init__`.
```suggestion
            raise TypeError(f"Expected Range or int for parameter 'range', not {type(range).__qualname__}")
```

src/cocotb/types/logic_array.py
```diff
@@ -391,15 +391,16 @@ def from_signed(
...
+
+        limit = 1 << (len(range) - 1)
+        if (value < -limit) or (value >= limit):
```
ktbarrett, line 398: Unnecessary parens and prefer `<` when doing these bounds based checks so they can be easily read left-to-right, like the inverse of `-limit <= value < limit` where the limits are on the inside because the value lies outside.
```suggestion
        if value < -limit or limit <= value:
```

ktbarrett (approved):

PR #4591, opened 2025-04-04
fix support for verilator dumpfile

keesj:
cocotb support logging to VCD and FST in verilator but I have specific requirements (and existing code) that uses the verilog $dumpfile and $dumpvars commands. 

The normal way to enable VCD logging is to use the compile and runtime flags to enable this e.g. call runner.build and runner.test with the waves=True argument. The suggestion is to build with waves support but run the test without this. This however produced an error %Error: Vtop.cpp:121: Turning on wave traces requires Verilated::traceEverOn(true) call before time 0. that this pull requests fixes.

### DIFF
--- a/tests/test_cases/test_dumpfile_verilator/Makefile
+++ b/tests/test_cases/test_dumpfile_verilator/Makefile
@@ -1,0 +1,24 @@
+# Copyright cocotb contributors
+# Copyright (c) 2015 Potential Ventures Ltd
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+TOPLEVEL_LANG ?= verilog
+SIM ?= verilator
+EXTRA_ARGS += --trace
+
+ifneq ($(SIM),verilator)
+
+all:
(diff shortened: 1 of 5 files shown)


tests/test_cases/test_dumpfile_verilator/Makefile
```diff
@@ -0,0 +1,24 @@
+# Copyright cocotb contributors
+# Copyright (c) 2015 Potential Ventures Ltd
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+TOPLEVEL_LANG ?= verilog
+SIM ?= verilator
+EXTRA_ARGS += --trace
+
+ifneq ($(TOPLEVEL_LANG),verilog)
```
ktbarrett, line 10: I'd switch this to checking if SIM is verilator. The selected `EXTRA_ARGS` only work on it.

ktbarrett:
Also check out `docs/source/newsfragments/README` and add a bug fix newsfrag.

ktbarrett:
Though it looks good overall.

ktbarrett (approved):

PR #4592, opened 2025-04-06
Change null vector support in Logic Array

ktbarrett:
Constructing a null vector using an int was never supported, so it made little sense that it could be converted to one. The warning on conversion was changed to an exception. Additionally, this means that equality between null vectors and ints are always false because ints always have value and null vectors do not. 

Plus some small unrelated changes:
* Small cleanups to LogicArray
* Removing incorrect newsfrags
* Meaningful optimization to equality between Arrays and lists and tuples

```
# Before
> python -m timeit -s "from cocotb.types import Array" -s "a = Array([1, 2, 3])" -s "b = [1, 2, 3]" "a == b"
500000 loops, best of 5: 618 nsec per loop

> python -m timeit -s "from cocotb.types import Array" -s "a = Array(range(10000))" -s "b = list(range(10000))" "a == b"
5000 loops, best of 5: 72.3 usec per loop

# After
> python -m timeit -s "from cocotb.types import Array" -s "a = Array([1, 2, 3])" -s "b = [1, 2, 3]" "a == b"
2000000 loops, best of 5: 147 nsec per loop

> python -m timeit -s "from cocotb.types import Array" -s "a = Array(range(10000))" -s "b = list(range(10000))" "a == b"
2000 loops, best of 5: 57.8 usec per loop
```

For large Arrays the difference is little, but most Arrays are going to be small, so closer to the 4x we see there.


imphil (approved):

tests/pytest/test_logic_array.py
```diff
@@ -384,20 +384,21 @@ def test_null_vector():
...
-    with pytest.warns(UserWarning):
-        assert null_vector.to_signed() == 0
-    with pytest.warns(UserWarning):
-        assert null_vector.to_unsigned() == 0
+    with pytest.raises(ValueError):
+        assert int(null_vector)
+    with pytest.raises(ValueError):
+        assert null_vector.to_signed()
+    with pytest.raises(ValueError):
+        assert null_vector.to_unsigned()
```
ktbarrett, line 392:
```suggestion
    with pytest.raises(ValueError):
        int(null_vector)
    with pytest.raises(ValueError):
        null_vector.to_signed()
    with pytest.raises(ValueError):
        null_vector.to_unsigned()
```

PR #4593, opened 2025-04-06
Fix Ctrl-C and `sys.exit()`

### LINKED ISSUE #4563: Questa hangs in CI
test_kill_sim is causing Questa to stacktrace, but also hang indefinitely. Apparently calling `sys.exit` is now causing issues.

In Questa calling `sys.exit` calls the registered shutdown callback in a re-entrant way, so this may be the source of the issues. What's unclear is why it's only hanging sometimes, and not all the times, after the stacktrace. I think this behavior changed after #4392 which is why we are seeing this now.

ktbarrett:
Closes #4563. Basically we catch `SystemExit` in C and don't print it so `exit()` is never called. Also cleans up a couple other issues seen in the test_kill_sim tests.

### TODO
- [x] ~~newsfrags~~ Not needed as this bug wasn't present in 1.9.
- [x] add a test for doing `sys.exit` and `Ctrl-C` after a GPI trigger fires to hit the missing lines in `simulatormodule.cpp`

### DIFF
--- a/src/cocotb/_scheduler.py
+++ b/src/cocotb/_scheduler.py
@@ -16,6 +16,7 @@ the ReadOnly (and this is invalid, at least in Modelsim).
 import logging
 import os
 import threading
+from bdb import BdbQuit
 from collections import OrderedDict
 from typing import Any, Dict, Union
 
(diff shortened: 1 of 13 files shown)


imphil (approved):

ktbarrett:
xref #4330.

PR #4622, opened 2025-04-18
Refactor X resolution

### LINKED ISSUE #4619: Add new docs / type checking minimum Python version
This Python version will be the minimum supported version for generating docs or doing type checking. Instead of using `if TYPE_CHECKING:` guards in many places, `if sys.version_info >= MIN_TYPING_VERSION` will be used instead. This is necessary because Sphinx does not set `TYPE_CHECKING` to `True` and so all imports and type definitions in those blocks will not resolve correctly.

~~A reasonable minimum Python version for documentation generation and type checking is probably 3.10. This adds `ParamSpec`, `TypeAlias`, and the new typing union syntax `A | B`. Features added in 3.11 aren't important to use.~~ We are currently using `Self` which is a 3.11 feature. Features added in 3.12 are nice but can't be used effectively by us in `if TYPE_CHECKING` blocks. And 3.13 is just far too new to force as a minimum requirement.

ktbarrett:
Depends upon #4592.

Rather than putting X resolving on Logic and LogicArray and tied to the
integer cast, a new `resolve()` function was added which takes a Logic
or LogicArray, maps the Xs, and returns a copy. This can be used in more
contexts than just for integer conversion.

This functionality was also used to implement a new way to support
COCOTB_RESOLVE_X that isn't strictly backwards compatible, but it
shouldn't be problematic. That being to resolve Xs as soon as the value
is gotten from the handle. This means that COCOTB_RESOLVE_X can be set
*and* X values can still be used properly with any instance of Logic or
LogicArray, when previously all instance were subject to the resolution
function.

To control handles X resolution, not only does COCOTB_RESOLVE_X still
work, but `cocotb.handle.set_resolve_x` allows the user to change it at
test time.

Additionally:
* Changes `if TYPE_CHECKING` blocks to be `sys.version_info` guarded blocks so Sphinx can correctly link functionality included in those conditional blocks (closes #4619).
* Removes some dead code from `cocotb.utils` after the refactor.
* Un-deprecates `LogicArray.__bool__` as the semantics are now self-consistent. The previous 2.0 development behavior can be achieved by using `COCOTB_RESOLVE_X=ZEROS`.

### TODO
- [x] newsfrag add `set_resolve_x()`
- [x] newsfrag `resolve()`
- [x] newsfrag change in `COCOTB_RESOLVE_X` behavior
- [x] cleanup any newsfrag of deprecated `LogicArray.__bool__`.

### DIFF
--- a/docs/source/newsfragments/4622.feature.rst
+++ b/docs/source/newsfragments/4622.feature.rst
@@ -1,0 +1,1 @@
+:class:`.LogicArray` now supports :class:`bool` casts and usage in conditionals (e.g. ``if dut.data.value: ...``).
--- a/docs/source/building.rst
+++ b/docs/source/building.rst
@@ -170,21 +170,26 @@ Cocotb
     Defines how to resolve bits with a value of ``X``, ``Z``, ``U``, ``W``, or ``-`` when being converted to integer.
     Valid settings are:
 
-    ``VALUE_ERROR``
-       Raise a :exc:`ValueError` exception.
-    ``ZEROS``
-       Resolve to ``0``.
-    ``ONES``
-       Resolve to ``1``.
-    ``RANDOM``
-       Randomly resolve to a ``0`` or a ``1``.
+    ``error``
+        Resolves nothing.
+    ``weak``
+        Resolves ``L`` to ``0`` and ``H`` to ``1``.
+    ``zeros``
+        Like ``weak``, but resolves all other non-\ ``0``\ /\ ``1`` values to ``0``.
+    ``ones``
+        Like ``weak``, but resolves all other non-\ ``0``\ /\ ``1`` values to ``1``.
+    ``random``
+        Like ``weak``, but resolves all other non-\ ``0``\ /\ ``1`` values randomly to either ``0`` or ``1``.
 
-    Set to ``VALUE_ERROR`` by default.
+    There is also a slight difference in behavior of ``bool(logic)`` when this environment variable is set.
+    When this variable is set, ``bool(logic)`` treats all non-\ ``0``\ /\ ``1`` values as equivalent to ``0``.
+    When this variable is *not* set, ``bool(logic)`` will fail on non-\ ``0``\ /\ ``1`` values.
 
     .. warning::
+        Using this feature is *not* recommended.
 
-        This exists for backwards-compatability reasons.
-        Using any value besides ``VALUE_ERROR`` is *not* recommended.
+    .. deprecated:: 2.0
+        The previously accepted values ``VALUE_ERROR``, ``ZEROS``, ``ONES``, and ``RANDOM`` are deprecated.
 
 .. envvar:: LIBPYTHON_LOC
 
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -137,11 +137,6 @@ These are a set of datatypes that model the behavior of common HDL datatypes.
     :members:
     :inherited-members:
 
-.. autoclass:: cocotb.types.logic_array.ResolveX
-
-.. autodata:: cocotb.types.logic_array.RESOLVE_X
-
-
 .. _triggers::
 
 Triggers
--- a/docs/source/newsfragments/4622.feature.1.rst
+++ b/docs/source/newsfragments/4622.feature.1.rst
@@ -1,0 +1,1 @@
+Added :meth:`.Logic.resolve` and :meth:`.LogicArray.resolve` to resolve non-``0``/``1`` values on demand.
--- a/src/cocotb/_py_compat.py
+++ b/src/cocotb/_py_compat.py
@@ -12,7 +12,11 @@ import sys
 from contextlib import AbstractContextManager
 from typing import TypeVar, Union, overload
 
-__all__ = ("StrEnum", "cached_property", "insertion_ordered_dict", "nullcontext")
+__all__ = (
+    "cached_property",
+    "insertion_ordered_dict",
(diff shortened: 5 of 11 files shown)


docs/source/install.rst
```diff
@@ -129,7 +131,7 @@ The installation instructions vary depending on your package manager:
 
       .. code-block:: bash
 
-	  guix install python-cocotb
+	    guix install python-cocotb
```
imphil, line 134: I guess the tab is the actual problem here that you tried to solve.

docs/source/newsfragments/4622.change.rst
```diff
@@ -0,0 +1 @@
+:envvar:`COCOTB_RESOLVE_X` behavior was changed so that :class:`.LogicArray` and :class:`.Logic` values are resolved as soon as they are returned from the :meth:`handle.get() <cocotb.handle.ValueObjectBase.get>` or :attr:`handle.value <cocotb.handle.ValueObjectBase.value>`, and not only when the value is converted to an :class:`int`. It also only affects values return from handles and not all :class:`!.LogicArray`\ s or :class:`!.Logic`\ s.
```
imphil, line 1:
```suggestion
:envvar:`COCOTB_RESOLVE_X` behavior was changed so that :class:`.LogicArray` and :class:`.Logic` values are resolved as soon as they are returned from the :meth:`handle.get() <cocotb.handle.ValueObjectBase.get>` or :attr:`handle.value <cocotb.handle.ValueObjectBase.value>`, and not only when the value is converted to an :class:`int`. It also only affects values returned from handles and not all :class:`!.LogicArray`\ s or :class:`!.Logic`\ s.
```

docs/source/newsfragments/4622.feature.2.rst
```diff
@@ -0,0 +1 @@
+Added :func:`cocotb.handle.set_resolve_x` for changing the non-``0``/``1`` value resolution behavior at runtime, which was previously limited to begin set with :envvar:`COCOTB_RESOLVE_X`.
```
imphil, line 1:
```suggestion
Added :func:`cocotb.handle.set_resolve_x` for changing the non-``0``/``1`` value resolution behavior at runtime (overriding :envvar:`COCOTB_RESOLVE_X` if set).
```

(Not quite sure what you wanted to say with "begin set".)

docs/source/newsfragments/4622.feature.rst
```diff
@@ -0,0 +1 @@
+:class:`.LogicArray` now supports :class:`bool` casts and usage in conditionals (e.g. ``if dut.data.value: ...``).
```
imphil, line 1: YES :partying_face:

src/cocotb/_init.py
```diff
@@ -289,3 +291,22 @@ def _setup_regression_manager() -> None:
...
+        resolver = cast(
+            Union[ResolverLiteral, None],
+            {
+                None: None,
+                "VALUE_ERROR": None,
+                "ZEROS": "zeros",
+                "ONES": "ones",
+                "RANDOM": "random",
+            }[envvar],
+        )
```
imphil, line 308: My Monkish brain dislikes the discrepancy between the all-caps keyword for the environment variable and the lowercase variant for the function argument. I guess we're not tanking performance if we lowercase the env var for backwards-compatibility (or accept the all-caps or the lowercase variant) and document the lower-case variant as the thing we want?

src/cocotb/_init.py
```diff
@@ -289,3 +291,22 @@ def _setup_regression_manager() -> None:
...
+    except KeyError:
+        raise ValueError(f"Invalid COCOTB_RESOLVE_X value {envvar!r}") from None
```
imphil, line 310:
```suggestion
        raise ValueError(f"Invalid COCOTB_RESOLVE_X value {envvar!r}. Valid options: ...") from None
```

Let's help users to find the right keyword without having to dig up the docs.

imphil (approved):
Very nice, thanks!

docs/source/install.rst
ktbarrett: Yeah it's a little busted.

src/cocotb/_init.py
ktbarrett: > I guess we're not tanking performance if we lowercase the env var for backwards-compatibility (or accept the all-caps or the lowercase variant) and document the lower-case variant as the thing we want?

I had done that as part of a previous version of this PR, so I'll do that.

ktbarrett:
Ugh... this breaks pretty bad in some codebases. I don't think I'm going to be able to get away without infecting `Logic` with the `COCOTB_RESOLVE_X` virus. There is maybe a happy medium of the existing features in this PR and backwards compatibility.

ktbarrett:
The new implementation just does the more obvious thing and adds RESOLVE_X to `Logic`. This required factoring it out into a new module shared by both `_logic.py` and `_logic_array.py`.

I made `RESOLVE_X` a global constant and removed the `set_resolve_x` from the previous implementation as it was not of much use and this allowed me to make some optimizations by switching different method implementations once at startup, rather than checking the variable all the time.

I also realize that `__bool__` works odd as soon as you have `RESOLVE_X` set. All of a sudden those non-0/1 values are becoming values that might impact the tested value. BinaryValue only ever looked for `1` or `H`. So if `COCOTB_RESOLVE_X` is used, you get the legacy behavior of `__bool__` and without, you get the new better behavior: `bool(int(self))`.

imphil (approved):

PR #4623, opened 2025-04-19
Police public API

ktbarrett:
Depends upon #4624. Several small public/private API changes.

* Made `cocotb.regression_manager` private
* Made `cocotb.random_seed` *not* private
* Made modules with `cocotb.types` types private since they are all re-exported in `cocotb.types` and there should only be one canonical public name for a thing
* Made `SimHandle` function private (typical use case is now covered by `handle["path/to/thing"]`)

### TODO
- [x] newsfrags, newsfrags, newsfrags...


imphil (approved):

docs/source/newsfragments/4623.removal.1.rst
```diff
@@ -0,0 +1 @@
+:class:`.Range`, :class:`.Logic`, :class:`.LogicArray`, and :class:`.Array` are now only available on the :mod:`cocotb.types` module and their more specific modules are now private.
```
ktbarrett, line 1:
```suggestion
:class:`.Range`, :class:`.Logic`, :class:`.LogicArray`, and :class:`.Array` are now only available on the :mod:`cocotb.types` module and their implementation modules are now private.
```

docs/source/newsfragments/4623.removal.2.rst
```diff
@@ -0,0 +1 @@
+``cocotb.handle.SimHandle`` was made private. Generally, :class:`dut.hierarchy["path/to/thing"] <cocotb.handle.HierarchyObject>` or :meth:`.HierarchyObject._get` can be used get objects by a full path, which was the typical reason for this function to be public originally.
```
ktbarrett, line 1:
```suggestion
``cocotb.handle.SimHandle`` was made private. Generally, :class:`dut.hierarchy["path/to/thing"] <cocotb.handle.HierarchyObject>` or :meth:`.HierarchyObject._get` can be used get objects by a full path, which was the typical public use for this function originally.
```

PR #4626, opened 2025-04-21
Rename `cocotb.pass_test` to `cocotb.fpass`

ktbarrett:
I think the "test" in "pass_test" is redundant and doesn't signal this is a "forced" pass. We follow pytest's naming convention where it uses "xfail" instead of "expected_fail", so we use "fpass" instead of "force_pass".

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -396,7 +396,7 @@ Miscellaneous
 Test Control
 ------------
 
-.. autofunction:: cocotb.pass_test
+.. autofunction:: cocotb.fpass
 
 Other Runtime Information
 -------------------------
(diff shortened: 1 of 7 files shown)


cmarqu:
> We follow pytest's naming convention where it uses "xfail" instead of "expected_fail", so we use "fpass" instead of "force_pass".

Pytest probably chose the short name over the long name because of the limited terminal size  in the verbose status output, and then made the decorator name match the output.
For us, `fpass` would just be the function name though, right? I would rather go with a self-explanatory name then, that is, stay with `pass_test` or use `force_pass`.

ktbarrett:
pytest also has `pytest.xfail` and `pytest.mark.xfail`, not just the status output.

cmarqu:
Just saying that the reason they chose the shorter "xfail" over "expected_fail" is likely its usage in terminal output.

PR #4628, opened 2025-04-22
Make `CancellationError` a `RuntimeError`

ktbarrett:
Making it a custom type is not particularly useful, and this reduces the total API size. If we need to make it more specific we can do so later in a backwards-compatible way by inheriting from `RuntimeError`.

### DIFF
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -41,14 +41,6 @@ ResultType = TypeVar("ResultType")
 _debug = "COCOTB_SCHEDULER_DEBUG" in os.environ
 
 
-class CancellationError(Exception):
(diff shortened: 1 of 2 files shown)


eric-wieser:
> If we need to make it more specific we can do so later in a backwards-compatible way by inheriting from `RuntimeError`.

The current version is backwards incompatible, since the old exception disappeared. Why not extend `RuntimeError` immediately?

ktbarrett:
@eric-wieser Long time no see! `CancellationError` was introduced a couple months ago in the current development cycle, so it isn't breaking API wrt releases.

imphil (approved):

PR #4629, opened 2025-04-23
Document `PYGPI_PYTHON_BIN`

### LINKED ISSUE #4609: Document new PYGPI_PYTHON_BIN environment variable
I am doing experiments with cocotb 2.0 (e4f29124ce87139db9fd5464c07ba51ad67e14dd) and it broke our build scripts. It seems we now have to set `PYGPI_PYTHON_BIN` when we did not have to set this before. If this is intentional, it seems it should be added to the [Extending existing build flows](https://docs.cocotb.org/en/development/custom_flows.html) docs.

ktbarrett:
Closes #4609.

### DIFF
--- a/docs/source/building.rst
+++ b/docs/source/building.rst
@@ -355,10 +355,13 @@ GPI
 PyGPI
 -----
 
-.. warning::
+.. envvar:: PYGPI_PYTHON_BIN
 
-    PyGPI is an experimental feature and subject to change.
-
+    The Python binary in the Python environment to use with cocotb.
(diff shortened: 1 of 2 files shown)


docs/source/building.rst
```diff
@@ -355,10 +355,13 @@ GPI
 PyGPI
 -----
 
-.. warning::
-
-    PyGPI is an experimental feature and subject to change.
+.. envvar:: PYGPI_PYTHON_BIN
 
+    The Python binary in the Python environment to use with cocotb.
+    This is set to the result of ``cocotb-config --python-bin`` in the Makefiles and Python Runner.
```
cmarqu, line 361: Might link the "Python Runner".

cmarqu (approved):

PR #4630, opened 2025-04-24
Add waves support for GHDL and NVC

oscargus:
Also changed the tests so that it is enough to set `HDL_TOPLEVEL_LANG` to vhdl to be able to run all tests. Arbitrarily selected GHDL as the default VHDL simulator.

In addition, replaced the deprecated `vhdl_sources` argument name and clarified the assertion error that one can get if the installed version is different from the latest commit in git.

### DIFF
--- a/tests/pytest/test_runner.py
+++ b/tests/pytest/test_runner.py
@@ -18,7 +18,10 @@ tests_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sim_build = os.path.join(os.path.dirname(os.path.abspath(__file__)), "sim_build")
 sys.path.insert(0, os.path.join(tests_dir, "pytest"))
 
-sim = os.getenv("SIM", "icarus")
+sim = os.getenv(
+    "SIM",
+    "icarus" if os.getenv("HDL_TOPLEVEL_LANG", "verilog") == "verilog" else "nvc",
+)
 
 pre_cmd_sims = {
     "questa",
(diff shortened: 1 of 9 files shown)


tests/pytest/test_runner.py
```diff
@@ -123,7 +126,10 @@ def test_missing_libpython(monkeypatch):
         hdl_sources = [os.path.join(tests_dir, "designs", "runner", "runner.vhdl")]
         gpi_interfaces = [os.getenv("VHDL_GPI_INTERFACE", None)]
 
-    sim_tool = os.getenv("SIM", "icarus")
+    sim_tool = os.getenv(
+        "SIM",
+        "icarus" if os.getenv("HDL_TOPLEVEL_LANG", "verilog") == "verilog" else "ghdl",
```
ktbarrett, line 131: GHDL has subpar support for cocotb. We use NVC by default in CI now, so that's probably a better default for VHDL.
```suggestion
        "icarus" if os.getenv("HDL_TOPLEVEL_LANG", "verilog") == "verilog" else "nvc",
```

oscargus: Fixed

ktbarrett (approved):

cmarqu (approved):

cmarqu:
Wondering if we should print a line with the filename that will be/has been written, or do the simulators do that themselves already?

ktbarrett:
@cmarqu I think Oscar's other PR addresses this in some capacity.

oscargus:
> Wondering if we should print a line with the filename that will be/has been written, or do the simulators do that themselves already?

I based it on the other simulators where it wasn't the case, see 

https://github.com/cocotb/cocotb/blob/e1ba051e99849fc84d36fa4046a502ed0a1ab438/tests/pytest/test_waves.py#L71-L85

However, I agree with the comment in general. Also, for most simulators you can select format, so I think it would be nice to be able to specify format and possibly filename. Not sure how to go about that though. Maybe an environment variable for the format? An option is also to create a `pytest-cocotb` extension to that one can add arguments to pytest for these things. Maybe this discussion can continue in #4635?

PR #4635, opened 2025-04-25
Add gui option for GHDL, NVC, Verilator, Icarus and Dsim

### LINKED ISSUE #4631: Support gui for open-source simulators
It would be nice to support automatically opening a waveform viewer after simulating with Verilator, NVC, and GHDL (and Icarus Verilog*?).

A preliminary design idea is to add a method to the `Runner` class that returns a filename/Path of the waveform file or None (the Runner base class returns None). Then, after running the simulation one can check if gui and the path is not None and in that case open Surfer or GTKWave, in some order if both are available on the system.

* Not sure if Icarus supports other things than shm and trn or which viewers support those...

oscargus:
Builds on #4630

Closes #4631

Opening to get feedback on design (third commit, until #4630 is merged) and if it may be possible to test it.

- [x] Add news fragment
- [x] Add documentation about the environment variable

Edit: this now allows passing WAVES and GUI using the corresponding environment variables. If this is a good idea (I think it is, they depend on the runs, not the tests), I should also:

- [x] Add documentation about this.

### DIFF
--- a/docs/source/newsfragments/4635.feature.2.rst
+++ b/docs/source/newsfragments/4635.feature.2.rst
@@ -1,0 +1,1 @@
+The ``GUI`` and ``WAVES`` environment variables can be used to override the corresponding ``gui`` and ``waves`` arguments to pytest test benches. This means that they can be provided at run-time without modifying the test bench.
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -200,6 +200,14 @@ class Runner(ABC):
     def _test_command(self) -> Sequence[_Command]:
         """Return command to run a test."""
 
+    def _use_external_viewer(self) -> bool:
+        """Return if an external viewer should be called after simulation when ``gui=True``."""
+        return False
+
+    def _waves_file(self) -> Optional[str]:
+        """Return file name of the generated waveform file for use with external viewer."""
+        return None
+
     def build(
         self,
         hdl_library: str = "top",
@@ -267,7 +275,7 @@ class Runner(ABC):
             clean: Delete *build_dir* before building.
             verbose: Enable verbose messages.
             timescale: Tuple containing time unit and time precision for simulation.
-            waves: Record signal traces.
+            waves: Record signal traces. Overridden by the ``WAVES`` environment variable.
             log_file: File to write the build log to.
 
         .. deprecated:: 2.0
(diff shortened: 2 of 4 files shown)


src/cocotb_tools/runner.py
```diff
@@ -200,6 +200,14 @@ def _build_command(self) -> Sequence[_Command]:
     def _test_command(self) -> Sequence[_Command]:
         """Return command to run a test."""
 
+    def _use_external_viewer(self) -> bool:
+        """Return if an external viewer should be called after simulation when gui=True."""
```
cmarqu, line 204:
```suggestion
        """Return if an external viewer should be called after simulation when ``gui=True``."""
```

oscargus: Sure. I was a bit confused about this. What you suggest is what I personally like, but having written lots of documentation using NumpyDoc standard, it should probably be `*gui* is True` or something, but then gui is not an argument of this method...

I have now realized that you use the Google docstring standard, so will read up on that.

src/cocotb_tools/runner.py
```diff
@@ -1234,8 +1277,8 @@ def _test_command(self) -> List[_Command]:
 class Verilator(Runner):
     """Implementation of :class:`Runner` for Verilator.
 
+    * ``waves`` argument *must* be given to :meth:`.build` if used and if ``gui`` is provided to :meth:`.test`.
```
oscargus, line 1280: This I am a bit confused about. For Verilator, `waves` is used both in `build` and in `test`. For all other simulators, it is only used in one of the methods. 

Was the first part of this sentence missing earlier (i.e., one must provide `waves` to `build` if it is used in `test`), or can I remove this sentence?

(Not a Verilator user, but not obvious why one specify it in `build` if not required? Performance?)

ktbarrett: Verilator is the only simulator where waves support is required during building. Passing `waves=True` to the build step builds the binary with tracing support which takes considerably longer to build and produces considerably larger binaries. Then deciding to run the test with waves has a considerable effect on runtime performance, so we wanted the `waves` switch in `test` to be able to turn it on/off.

src/cocotb_tools/runner.py
```diff
@@ -1234,8 +1280,8 @@ def _test_command(self) -> List[_Command]:
...
```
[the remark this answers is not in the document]
oscargus: Much better wording! Fixed.

docs/source/newsfragments/4635.feature.2.rst
```diff
@@ -0,0 +1,4 @@
+The ``GUI`` and ``WAVES`` environment variables can be used to
+override the corresponding ``gui`` and ``waves`` arguments to
+pytest test benches. This means that they can be provided at run-time
+without modifying the test bench.
```
ktbarrett, line 4: newsfrags cannot be multiline.

ktbarrett (approved):
Looks good.

Issue #4560, opened 2025-03-18
Fix warnings in docs build


ktbarrett:
Running `nox -s docs` I see ~150 Warnings. Some are innocuous, but others show things not building or linking correctly. These should be fixed before 2.0 is released.

oscargus:
Some explicit things from #4660 that I opened forgetting about this issue.
- [ ] Lots of unsupported tags for Doxygen `warning: ignoring unsupported tag`. Most likely a consequence of Doxygen development removing those tags.
- [ ] Lots of `.../docs/source/building.rst:: WARNING: 4 column based index found. It might be a bug of extensions you use: [('pair', 'GUI; Make Variable', 'var-GUI', '')]`. Some extension is using an old format of the index. Would be good to figure out which and consider if it is required. #4661 
- [ ] It seems like `setuptools` does not have an API reference documentation, so no links are working.
- [ ] It seems like `Logic` is not used in any example or is linked from anywhere.

cmarqu:
Probably an intersphinx mapping to https://setuptools.pypa.io/en/latest/ is missing.

cmarqu:
@oscargus Re "It seems like `Logic` is not used in any example or is linked from anywhere." - if you set https://sphinx-codeautolink.readthedocs.io/en/latest/reference.html#confval-codeautolink_warn_on_no_backreference in a local build, there are quite some more such occurrences if you feel like working on these :) (Thanks for all the PRs and issues recently!)

oscargus:
> Probably an intersphinx mapping to https://setuptools.pypa.io/en/latest/ is missing.

I think the problem is that there is no API documentation at https://setuptools.pypa.io/en/latest/ (I added the link in #4640 for what it is worth, but doesn't seem to change anything).

oscargus:
```
/home/docs/checkouts/readthedocs.org/user_builds/cocotb/checkouts/4640/docs/source/extensions.rst:74: WARNING: Inventory missing `setuptools` when resolving `setuptools` on line 2.
Possibly missing documentation entry entirely, or the object has been relocated from the source file. [codeautolink.missing_inventory]
/home/docs/checkouts/readthedocs.org/user_builds/cocotb/checkouts/4640/docs/source/extensions.rst:74: WARNING: Inventory missing `setuptools.setup` when resolving `setup` on line 2.
Possibly missing documentation entry entirely, or the object has been relocated from the source file. [codeautolink.missing_inventory]
[...]

PR #4640, opened 2025-04-26
Fix linking in documentation

oscargus:
Related to #4560

Most remaining warnings relate to code examples in the documentation (I do not understand how to fix those) and sloppily written news fragments.


oscargus:
Note sure why reassigning the module for `Array` etc breaks the RegressionManager...

ktbarrett:
The linking of Array, Range, etc. probably has some interplay with #4623 where the modules that define the type are made private so there is only one canonical name for those types. I think the issue also revolves around the use of `.. currentmodule::` which effects the global resolver state AFAI understand it.

A lot has changed and changed again in the 2.0 cycle and there are lots of newsfrags that are out of date. They need a cleanup pass for sure.

oscargus:
OK! Then I wait for #4623 and see what should be kept etc.

docs/source/contributing.rst
```diff
@@ -276,7 +276,7 @@ To potentially save you frustration and time, it is a good idea to get maintaine
 
 The best way to get maintainer pre-approval is to make a `Github issue <https://github.com/cocotb/cocotb/issues>`__.
 These issues can be a place for maintainers, as well as other users, to voice opinions on a proposed change before the task is worked.
-You may also propose changes on the `Gitter channel <https://gitter.im/cocotb/Lobby>`__ or by directly contacting a :ref:`maintainer`.
+You may also propose changes on the `Gitter channel <https://gitter.im/cocotb/Lobby>`__ or by directly contacting a :ref:`maintainers`.
```
cmarqu, line 279:
```suggestion
You may also propose changes on the `Gitter channel <https://gitter.im/cocotb/Lobby>`__ or by directly contacting one of the :ref:`maintainers`.
```

docs/source/contributing.rst
```diff
@@ -260,7 +260,7 @@ Follow the steps below to get your changes merged, i.e. integrated into the main
 5. Create new `pull request (PR) <https://github.com/cocotb/cocotb/pulls>`__.
 6. When you submit (or update) the pull request, a suite of regression tests will run.
    If any of them turns "red," i.e. reports a failure, you most likely need to fix your code before it can be merged.
-7. The pull request needs to be reviewed by at least one :ref:`maintainer`.
+7. The pull request needs to be reviewed by at least one :ref:`maintainers`.
```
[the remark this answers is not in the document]
oscargus: I agree. This happens when you only correct the link and do not read the text... 

I'll fix it once #4623 is in since it will not ready for merge before that anyway. Thanks for spotting! Similar below.

cmarqu (dismissed):

oscargus:
Recent push: I added `ResolveX` and `RESOLVE_X` to `cocotb.types`. As these are documented and I do not think it makes sense to document `cocotb.types._logic_array.ResolveX` etc.

oscargus:
Final(?) fix for now: Added test-stub in parametrize documentation example to suppress an error message.

cmarqu (approved):

src/cocotb/types/__init__.py
```diff
@@ -4,7 +4,22 @@
...
+    "RESOLVE_X",
+    "AbstractArray",
+)
+
+# Change module to get correct linking in Sphinx
+AbstractArray.__module__ = __name__
+Array.__module__ = __name__
+LogicArray.__module__ = __name__
+Range.__module__ = __name__
+ResolveX.__module__ = __name__
```
ktbarrett, line 25: See if this is still necessary. This is pretty ugly and not necessary in other places where we re-export things.

src/cocotb/types/__init__.py
```diff
@@ -4,7 +4,22 @@
 from ._abstract_array import AbstractArray
 from ._array import Array
 from ._logic import Logic
-from ._logic_array import LogicArray
+from ._logic_array import RESOLVE_X, LogicArray, ResolveX
```
ktbarrett, line 7: You can't import `RESOLVE_X` here, it's intended to be a global you set. But that's going to be refactored by #4622.

src/cocotb/types/__init__.py
oscargus: Still required. Problem is that the code examples otherwise try to link to `cocotb.types._range.Range` etc. which doesn't have documentation.

src/cocotb/types/__init__.py
oscargus: Fixed!

src/cocotb/types/__init__.py
ktbarrett: This is probably also why `start_soon` and a couple other things don't link. I think we need a solution (put in `cocotb.utils`) that handles re-exports properly. The `atpublic` library looks nice, but it's a pretty minor thing to add our first dependency for.

ktbarrett: Maybe pick this in? https://github.com/ktbarrett/cocotb/commit/7d6b1d9f5bef384e4c8ade9b3a8d98e229673dc0.

oscargus:
Not much remaining, but at least a bit better. :-)

ktbarrett:
@oscargus Are we good to final review and merge then?

PR #4641, opened 2025-04-26
Add codespell pre-commit hook and fix spelling

### LINKED ISSUE #4639: Add spell checker to pre-commit?
Will remove some annoying mistakes and pushing again etc.

oscargus:
Closes #4639 (didn't really plan to do this today...)

The big change in pyproject.toml is a consequence of auto-formatting in my editor. Only change is the addition of the codespell section.


pyproject.toml
```diff
@@ -3,39 +3,39 @@ requires = ["setuptools", "wheel"]
...
-    [[tool.towncrier.type]]
-        directory = "removal"
-        name = "Deprecations and Removals"
-        showcontent = true
-
-    [[tool.towncrier.type]]
-        directory = "change"
-        name = "Changes"
-        showcontent = true
+package = "cocotb"
```
ktbarrett, line 6: IIRC I've proposed removing the unnecessary indentation here before and someone getting mad about it. But that's not a problem if I'm the approver 😁

pyproject.toml
```diff
@@ -73,20 +73,11 @@ ignore = [
 
 [tool.ruff.lint.per-file-ignores]
 # necessary because of how file is included into documentation
-"examples/doc_examples/quickstart/test_my_design.py" = [
-    "E402",
-    "F811",
-]
+"examples/doc_examples/quickstart/test_my_design.py" = ["E402", "F811"]
```
ktbarrett, line 76: Now the below changes I'm not a fan of. Splitting these out onto multiple lines is nice to potentially add comments as to "why" it was set. Do you have a toml formatter in your editor?

ktbarrett (approved):
Thanks for this.

pyproject.toml
oscargus: This is the "Even better TOML"-plugin for VS Code. 2.9M installs. I guess the rationale used is that if it fits on a single line, then use a single line. But also that it wouldn't do this if there were comments per line.

I read up a bit more and did now add formatting of `pyproject.toml` with a custom formatting that kept the previous indentation of subtables. The setting is also used for the VS Code plugin.

There is no setting for always breaking arrays into separate lines, but one can see from some other places in the file, like `extend-select` that it will keep separate lines if there are comments.

pyproject.toml
oscargus: There are two settings, see https://taplo.tamasfe.dev/configuration/formatter-options.html

indent_tables: Indent subtables if they come in order.
indent_entries: Indent entries under tables. 

I activated indent_tables, but not indent_entries. This keeps parts of it, but also indents some of the other subtables. I, sort of, see the point of indenting subtables, but not really the entries. But can be discussed and I can do any setting that is agreed on. (A minor issue is that the VS Code extension seems to use four spaces rather than two, independent of the setting... So I set four spaces now just to have it consistent.)

ktbarrett:
Let's just not attempt to solve the pyproject.toml formatting this PR. I really don't like how that formatter decides to collapse multi-line expressions. No other formatter does that. black/ruff use "magic commas" and this ignores that.

oscargus:
Sure, I removed the last commit and the reformatting of pyproject.toml (so inconsistent formatting, but nothing changed, just added to it).

ktbarrett (approved):

cmarqu (approved):

PR #4646, opened 2025-04-30
Add coverage for cocotb_tools

oscargus:
I've been quite confused that the coverage was not changed when making commits towards runner.py...

One may consider adding the tests as well.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -215,7 +215,8 @@ def dev_test_sim(
         "pytest",
         "-v",
         "--doctest-modules",
-        "--cov=cocotb",
+        "--cov=src/cocotb",
+        "--cov=src/cocotb_tools/runner.py",
         "--cov-branch",
         # Don't display coverage report here
         "--cov-report=",
@@ -229,7 +230,8 @@ def dev_test_sim(
     session.run(
         "pytest",
         "-v",
-        "--cov=cocotb",
+        "--cov=src/cocotb",
+        "--cov=src/cocotb_tools/runner.py",
         "--cov-branch",
         # Don't display coverage report here
         "--cov-report=",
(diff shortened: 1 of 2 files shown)


ktbarrett:
Outside of the Runner where we can use `pytest-coverage` to collect coverage, the other things in that package can't easily have coverage collected on them. We also don't have exhaustive tests for every feature so trying to obtain high coverage in the Runner is not easy. And that goes without mentioning Runners for simulators we don't have access to yet, like DSim. So it begs the question of why collect coverage? Enabling coverage comes with the goal of actually trying to reach 100% coverage and that's a mountain of work that's low priority right now.

oscargus:
> So it begs the question of why collect coverage?

Valid point. My main reason was to confirm that my modifications were actually executed, especially for things that I didn't have the simulator set up for.

One may also ask: why *not* collect coverage. :-) 

For Dsim, I agree that it is a problem, but then it would at least be clear that Dsim is not tested (without reading the scripts etc). It would also be a constant reminder (for people checking the coverage) to possibly do something about it. But maybe I am overestimating the coverage as a driver for development (I have in many other projects started out by increasing test coverage as a way to learn the code base and project).

Admittedly, it is runner.py which is the primary interest for me. I guess that one can only collect that file in cocotb_runners, but not sure if the failures are related.

ktbarrett:
Does specifying `--cov=cocotb_tools.runner` work? That's fine by be. 

As for the failures:
* Icarus v12_0 random hangs during build randomly. It's super annoying, but we are not sure why. We can't reproduce it locally.
* The other tests looked like they timed out waiting for licenses. We only have so many licenses and all concurrent CI runs have to share them. The CI tests can sometimes time out due to license contention.

ktbarrett:
Are we sure that specifying `--cov` more than once works correctly? It seems like coverage decreases because of pytest tests coverage not being collected anymore.

oscargus:
I am quite sure: https://pytest-cov.readthedocs.io/en/latest/config.html#reference 

That is the only interpretation of "multi-allowed" that makes sense to me, but I can be wrong.

Yeah, I do not really know what to try now. I thought the pyproject changes was gonna do it, but not.

oscargus:
This is quite strange. Only using the pyproject.toml-info lead to the tests being included (maybe one should remove the nox-line?), but not runner.py.

Issue #4586, opened 2025-03-27
Add mypy checking to pre-commit and CI


ktbarrett:
`mypy` checks Python code for type correctness. Since we include type annotations, it would be nice to actually prove to ourselves and users that these annotations are at least self-consistent by checking cocotb with mypy. There is a mypy pre-commit action which should be set up to only check those files that have changed, but a CI check should run without arguments, checking every file.

imphil:
https://github.com/cocotb/cocotb/pull/4672 fixes this issue, I think @ktbarrett ?

PR #4650, opened 2025-05-05
Improve typing

oscargus:
As a preparation for #4586.

The `Optional` is a matter of taste, but to me it makes more sense to use it than `Union[foo, None]`. Both will be converted to `foo | None` in newer versions of Python though. I converted all with a single type, but not those that are like `Union[foo, bar, None]`.

I didn't get all the way with the Trigger/GPITrigger typing though, but this gives fewer errors...

### DIFF
--- a/src/cocotb/_scheduler.py
+++ b/src/cocotb/_scheduler.py
@@ -18,7 +18,7 @@ import os
 import threading
 from bdb import BdbQuit
 from collections import OrderedDict
-from typing import Any, Dict, Union
+from typing import Any, Dict, Optional
 
 import cocotb
 import cocotb.handle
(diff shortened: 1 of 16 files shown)


ktbarrett:
> but to me it makes more sense to use it than `Union[foo, None]`

Most of the core devs disagree, and I happen to agree with them (mostly). I think it makes sense for optional function arguments where the decision is pass or don't, but if you are explicitly passing or assigning `None`, `Union` makes more sense as there's an explicit list of the types that can be used.

Also after we drop Python <3.8 (after 2.0 release), we will add `from __future__ import annotations` everywhere and pyupgrade will rewrite everything, even `Optional`s to `A | None`.

src/cocotb/_scheduler.py
```diff
@@ -201,17 +201,19 @@ def __init__(self) -> None:
...
-        self._scheduled_tasks: OrderedDict[Task[Any], Union[BaseException, None]] = (
+        self._scheduled_tasks: OrderedDict[Task[Any], Optional[BaseException]] = (
             OrderedDict()
         )
-        self._pending_threads = []
-        self._pending_events = []  # Events we need to call set on once we've unwound
+        self._pending_threads: list[external_waiter] = []
+        self._pending_events: list[
+            Event
+        ] = []  # Events we need to call set on once we've unwound
```
ktbarrett, line 210: You can delete the comment, it's not useful, and the code will be a bit prettier.

src/cocotb/regression.py
```diff
@@ -303,7 +303,9 @@ def _execute(self) -> None:
 
         return self._tear_down()
 
-    def _schedule_next_test(self, trigger: Optional[Trigger] = None) -> None:
+    def _schedule_next_test(
+        self, trigger: Optional[cocotb._gpi_triggers.GPITrigger] = None
```
ktbarrett, line 307: Let's do an unqualified import for `GPITrigger` here.

src/cocotb_tools/runner.py
```diff
@@ -88,7 +89,7 @@ def _as_tcl_value(value: str) -> str:
 _sv_escape_translate_table = str.maketrans(_sv_escapes)
 
 
-def _as_sv_literal(value: str) -> str:
+def _as_sv_literal(value: Any) -> str:
```
ktbarrett, line 92: In general use `object` instead of `Any` unless you get complaints.
https://mypy.readthedocs.io/en/latest/dynamic_typing.html#any-vs-object
```suggestion
def _as_sv_literal(value: object) -> str:
```

oscargus: Didn't know that!

ktbarrett:
Superseded by #4672.

PR #4656, opened 2025-05-06
Fail if the user asks for coverage and we can't provide

ktbarrett:


oscargus:
I agree that this makes sense, but maybe add a newsfragment that the behavior has changed? I can imagine users trying to get coverage without being aware of it and then it is good to at least have a changelog entry to point at that the behavior has changed.

(With that said, I have no idea how you usually deal with these things.)

docs/source/newsfragments/4656.change.rst
```diff
@@ -0,0 +1 @@
+If the user requests coverage collection via :envvar:`COCOTB_LIBRARY_COVERAGE` or :envvar:`COCOTB_USER_COVERAGE` and the :mod:`coverage` module is not available, the regression will fail instead of continue.
```
imphil, line 1:
```suggestion
If the user requests coverage collection via :envvar:`COCOTB_LIBRARY_COVERAGE` or :envvar:`COCOTB_USER_COVERAGE` and the :mod:`coverage` module is not available, the regression will fail instead of continuing.
```

?

(or just remove the "instead of" part)

imphil (approved):

docs/source/newsfragments/4656.change.rst
ktbarrett:
```suggestion
If the user requests coverage collection via :envvar:`COCOTB_LIBRARY_COVERAGE` or :envvar:`COCOTB_USER_COVERAGE` and the :mod:`coverage` module is not available, the regression will fail.
```

ktbarrett:
Interesting... it skipped the tests... I'm wondering if we are misinterpreting `ignore-files`.

PR #4657, opened 2025-05-06
Remove pylint config

ktbarrett:
It was never integrated into CI and [ruff fulfills many of the pylint checks](https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-pylint), especially once combined with a type checker (should be done soon). We should add [`PL`](https://docs.astral.sh/ruff/rules/#pylint-pl) checks in a separate PR once we've gotten the configuration correct.

This PR also intends to test #4651.


ktbarrett:
@oscargus Looks like we should also remove the Performance Benchmark from non-source-changing builds. I wonder if there's a way to put the list in a file? Or maybe the Performance Benchmark should just be added the the regression test yaml file as a separate job?

imphil (approved):

oscargus:
> I wonder if there's a way to put the list in a file?

I do not think so. I checked a bit and didn't find any mention of it at least. Maybe our use case of having quite a bit of different files in the root directory is a bit uncommon and most people just have a few directories to skip on. (Or they are simply not that detailed with individual files?)

One can also add `.theia` and odd files in other directories as well, like `.github/PULL_REQUEST_TEMPLATE.md`, but probably people are not really doing it at that detail level (but I may).

PR #4658, opened 2025-05-06
Questa: Ensure VHDL_GPI_INTERFACE is passed to simulator

imphil:
VHDL_GPI_INTERFACE was not passed on to the simulator if the user only set it as make variable. (It was passed on if the user exported it in the shell environment.) Fix that and improve the formatting in the Makefile to better see what's going on.


ktbarrett (approved):

ktbarrett:
Catching broken tests. Nice.

Issue #1858, opened 2020-05-23
[Riviera-Pro] Figure out correct behavior in test_array.test_discover_all


imphil:
The test [`test_array.test_discover_all`](https://github.com/cocotb/cocotb/blob/master/tests/test_cases/test_array/test_array.py#L251) currently contains an interesting collection of "pass" states for different versions of Riviera-PRO. We need to work with Aldec to figure out where these issues come from, and clean up the test.

cmarqu:
There is also a table with opened issues at Aldec that may be relevant here: https://github.com/cocotb/cocotb/pull/416

garmin-mjames:
@cmarqu All of the listed Aldec **SPT*** issues are listed as resolved in the Riviera-PRO 2016.06 release notes.

PR #4664, opened 2025-05-08
Aldec: Re-enable testing for complex array types

imphil:
test_array had special handling for Aldec simulators.

* The Makefile tried to use a separate VHDL file for Aldec simulators, but the condition never matched (any more), since the simulator is called riviera, not aldec. We were running with the `array_module.vhd` for probably a long time.
* Now that we're using the full file, we can also test the associated features and they are confirmed to work with Riviera-PRO 2024.04 and 2020.04.

Part of #1858


imphil:
Unfortunately I think I have to re-introduce similar behavior for Questa with QIS/Qrun, but let's see.

ktbarrett (approved):

PR #4665, opened 2025-05-08
Cleanup newsfrags

ktbarrett:
Several clean ups to the newsfrags and a couple small unrelated changes:
* Fixed some broken links (`COCOTB_PLUSARGS`, `cocotb.__version__`, `PYTHON_BIN`)
* Removed or changed some newsfrags that were no longer accurate.
* Touched up some newsfrags wording
* When talking about arguments make using double backticks consistent vs using asterisks
* Removed newsfrags already in 1.9
* Renamed `cocotb.random_seed` to `cocotb.RANDOM_SEED` like it was in 1.9 because why change the name at this point?


docs/source/building.rst
```diff
@@ -547,6 +547,12 @@ The following variables are makefile variables, not environment variables.
     Only supported by the Aldec Riviera-PRO, Aldec Active-HDL, and Siemens EDA Questa simulators.
 
 
+.. make:var:: PYTHON_BIN
+
+    The path to the ``python`` binary.
+    Set to the result of ``cocotb-config --python-bin`` if present on the ``PATH``.
+    Otherwise defaults to ``python``.
```
imphil, line 554: It's beyond the scope of this PR, but should we not default to `python3`? (I think it was/is the recommendation to not expect `/usr/bin/python` to be Python 3, and last I checked macos still shipped Python 2 as `/usr/bin/python`

docs/source/newsfragments/3851.change.rst
```diff
@@ -0,0 +1 @@
+:meth:`.Lock.locked` is now a method instead of an attribute
```
imphil, line 1:
```suggestion
:meth:`.Lock.locked` is now a method instead of an attribute.
```

docs/source/newsfragments/3851.removal.1.rst
```diff
@@ -0,0 +1 @@
+``Event.fired`` attribute was made private. Use :meth:`.Event.is_set`
```
imphil, line 1:
```suggestion
``Event.fired`` attribute was made private. Use :meth:`.Event.is_set`.
```

imphil (approved):

ktbarrett:
But this past commit is running the tests despite the commit only touching `docs/`... The whole "ignore-paths" thing seems to be inconsistent or unintuitively applied. I guess as long as it doesn't do the wrong thing I can't complain if it doesn't always do the right thing.

PR #4668, opened 2025-05-08
Install development version of cocotb after ecosystem test package

### LINKED ISSUE #4659: pyuvm ecosystem tests do not use cocotb 2.0
Ecosystem tests such as https://github.com/cocotb/cocotb/actions/runs/14435312465/job/40475379567 are supposed to use the master version of cocotb against a released version of a tool. But since pyuvm specifies `cocotb < 2.0` it discards the master version of cocotb and uses the 1.9.2 version instead. That's not that the test is expected to do.

oscargus:
Closes #4659

(Running as part of the PR to see how it works. Will revert that part once it's been approved.)

I *think* the install works in the expected way now, but errors are triggered using cocotb 2.0.

### DIFF
--- a/.github/workflows/ecosystem-compat.yml
+++ b/.github/workflows/ecosystem-compat.yml
@@ -13,6 +13,10 @@
 name: Ecosystem compatibility tests
 
 on:
+  pull_request:
+    branches:
+      - master
+      - 'stable/**'
   # Run daily at midnight (UTC).
   schedule:
     - cron: '0 0 * * *'
@@ -38,9 +42,6 @@ jobs:
         with:
           path: cocotb
 
-      - name: Install the development version of cocotb
-        run: pip3 install ./cocotb
-
       - name: Checkout cocotb-coverage repository
         uses: actions/checkout@v4
         with:
@@ -49,6 +50,9 @@ jobs:
 
       - name: Install the release version of cocotb-coverage
         run: pip3 install cocotb-coverage==1.2
+
+      - name: Install the development version of cocotb
+        run: pip3 install ./cocotb
 
       - name: Run tests
         # Don't run tests through tox (as present in cocotb-coverage) to be able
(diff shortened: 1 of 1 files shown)


.github/workflows/ecosystem-compat.yml
```diff
@@ -47,7 +51,7 @@ jobs:
           path: cocotb-coverage
 
       - name: Install the release version of cocotb-coverage
-        run: pip3 install cocotb-coverage==1.2
+        run: pip3 install --no-deps cocotb-coverage==1.2
```
ktbarrett, line 54: These tools can have more dependencies that just cocotb. I think instead of `--no-deps`, we just move the `cocotb` install below the package-under-test install plus whatever flags are necessary to force the local version to be installed over one gathered from pypi.

It's slightly wasteful, but more manageable.

oscargus: Sure. My idea was to add those manually, when needed, but this will also work.

oscargus:
Seem to work to just change the install order:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
pyuvm 3.0.0 requires cocotb<2.0,>=1.6.0, but you have cocotb 2.0.0.dev0+c737902 which is incompatible.
Successfully installed cocotb-2.0.0.dev0+c737902
``` 

I added a second commit to remove the temporary PR execution. Not sure if you have squash activated, but if you do, please squash. Otherwise, I will squash locally when I get an OK.

(Hmm, for some reason the partial results are not visible for the first commit... But they can be seen here: https://github.com/cocotb/cocotb/actions/runs/15131935992 )

ktbarrett (approved):
At some point in the future (once they implement it), we should add `--ignore-conflicts`. Looks good!

PR #4669, opened 2025-05-08
Enable Python 3.14 test

oscargus:
Running in CI now to see that it works etc. Will change back to experimental after confirmation.

Seems to run as expected. The fail comes from a dependency not working in 3.14 (I think).

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -17,7 +17,7 @@ codecov:
         #
         # [1] https://docs.codecov.io/docs/merging-reports#how-does-codecov-know-when-to-send-notifications
         # [2] https://docs.codecov.io/docs/notifications#preventing-notifications-until-after-n-builds
-        after_n_builds: 24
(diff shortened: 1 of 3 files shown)


ktbarrett:
Looks liek this could be solved by installing `libxml` and `libxslt`?

oscargus:
I didn't read that far... Only saw the deprecation. Will figure out how to add those for 3.14 only.

ktbarrett:
Oh. One last change I forgot about. The number of jobs before reporting coverage will need to be incremented. Its in `.codecov.yml`.

oscargus:
Bumped that to 25. Is that the (now) 17 jobs in the "ci" group and the 8 in the "licensed" group? Maybe one should add a job that checks that it is up to date?

ktbarrett:
In an ideal world yeah we would write a job to check it. But it doesn't change too often.

PR #4670, opened 2025-05-09
Fix segfault

### LINKED ISSUE #4667: Segfault in VCS Slave mode exposed a bug?
Regarding #4499 

I am using VCS in a slave-mode configuration and seeing this `delete cb_data` is causing a seg fault. I suspect it's because the `delete` is happening AFTER the PyGilState_Release is called. If I move the DEFER(delete) to AFTER the DEFER(delete cb_data) then it works fine for me. 

I instrumented the Deferrable class and saw this ordering after I made the change and is working fine for me. Notice the delete happens before the GIL releaes.

```
DEFER to_simulator()
DEFER PyGILState_Release(gstate)
DEFER delete cb_data
DESTRUCTOR: delete cb_data
DESTRUCTOR: PyGILState_Release(gstate)
DESTRUCTOR: to_simulator()
```

Code change:
```
diff --git i/src/cocotb/share/include/cocotb_utils.h w/src/cocotb/share/include/cocotb_utils.h
index dc0d7fc6..54a6cea3 100644
--- i/src/cocotb/share/include/cocotb_utils.h
+++ w/src/cocotb/share/include/cocotb_utils.h
@@ -50,21 +50,22 @@ extern "C" COCOTBUTILS_EXPORT int is_python_context;
 template <typename F>
 class Deferable {
   public:
-    constexpr Deferable(F f) : f_(f) {};
-    ~Deferable() { f_(); }
+    constexpr Deferable(F f, char* statement) : f_(f), statement_(statement) {};
+    ~Deferable() { printf("DESTRUCTOR: %s\n", statement_); f_(); }
 
   private:
     F f_;
+    char* statement_;
 };
 
 template <typename F>
-constexpr Deferable<F> make_deferable(F f) {
-    return Deferable<F>(f);
+constexpr Deferable<F> make_deferable(F f, char* statement) {
+    return Deferable<F>(f, statement);
 }
 
 #define DEFER1(a, b) a##b
 #define DEFER0(a, b) DEFER1(a, b)
 #define DEFER(statement) \
-    auto DEFER0(_defer, __COUNTER__) = make_deferable([&]() { statement; });
+    printf("DEFER " # statement "\n");auto DEFER0(_defer, __COUNTER__) = make_deferable([&]() { statement; }, #statement); 
 
 #endif /* COCOTB_UTILS_H_ */
diff --git i/src/cocotb/share/lib/simulator/simulatormodule.cpp w/src/cocotb/share/lib/simulator/simulatormodule.cpp
index 799138fc..09f9a62f 100644
--- i/src/cocotb/share/lib/simulator/simulatormodule.cpp
+++ w/src/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -169,10 +169,10 @@ int handle_gpi_callback(void *user_data) {
     DEFER(to_simulator());
 
     PythonCallback *cb_data = (PythonCallback *)user_data;
-    DEFER(delete cb_data);
 
     DEFER(PyGILState_Release(gstate));
+    DEFER(delete cb_data);
 
     // Call the callback
     PyObject *pValue =

```

Reproducer here: https://github.com/awmoore-intel/cocotb-vcs-slave

ktbarrett:
Closes #4667.

`delete`ing the `PythonCallback` calls Python C API functions, which is not safe unless the GIL is acquired, so the `Ensure` and `DEFER(Release)` were moved up, so the release occurs *after* the delete of the callback when destructors are run.

### DIFF
--- a/src/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/src/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -161,11 +161,11 @@ int handle_gpi_callback(void *user_data) {
     to_python();
     DEFER(to_simulator());
 
-    PythonCallback *cb_data = (PythonCallback *)user_data;
-    DEFER(delete cb_data);
-
     PyGILState_STATE gstate = PyGILState_Ensure();
     DEFER(PyGILState_Release(gstate));
+
+    PythonCallback *cb_data = (PythonCallback *)user_data;
+    DEFER(delete cb_data);
 
     // Call the callback
     PyObject *pValue =
--- a/src/cocotb/share/lib/simulator/simulatormodule.cpp
+++ b/src/cocotb/share/lib/simulator/simulatormodule.cpp
@@ -21,11 +21,6 @@
 
 // This file defines the routines available to Python
 
-#define COCOTB_ACTIVE_ID \
-    0xC0C07B  // User data flag to indicate callback is active
-#define COCOTB_INACTIVE_ID \
-    0xDEADB175  // User data flag set when callback has been de-registered
-
 #define MODULE_NAME "simulator"
 
 // callback user data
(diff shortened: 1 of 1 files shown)


imphil (approved):

ktbarrett:
So I think fixing this bug uncovered a bug in the FLI implementation. We seem to be calling up on a callback that has already been handled.

ktbarrett:
Oh god... it's the first commit that's the issue. Removing the unused field of `PythonCallback` causes a segfault. The only thing I can think of is something is clobbering the first word of the callback object. I can't debug this right now, so I'm thinking we just leave the padding.

PR #4671, opened 2025-05-09
Change default of `PYTHON_BIN` to `python3`

ktbarrett:
Based on https://github.com/cocotb/cocotb/pull/4665#discussion_r2080095731. PEP394 states this should always be available and correct if a Python 3 is installed.


imphil (approved):
You might also want to change the docs that triggered this PR in https://github.com/cocotb/cocotb/pull/4665#discussion_r2080095731

PR #4672, opened 2025-05-11
Fix remaining type issues and add `mypy` to pre-commit

ktbarrett:
Uses version check instead of `TYPE_CHECKING` generally because Sphinx doesn't set it, so all `TypeAlias`es created in those blocks will not be seen or resolved by Sphinx.

Adds a mypy config and fixes the last of the typing issues:
* `Trigger._prime` needs to take callbacks that take `Self` to make calls to `_prime` happy.
* Added typing to `_bridge.py` and related functionality in `_scheduler.py`.
* Various other inaccuracies

We should probably try to clean up uses of `Any` at some point, especially if there are any on public APIs. Also we should extend the amount of typed Python code.

Finally added mypy to pre-commit. This will only check files that changed locally, keeping it shorter to execute. But `pre-commit.ci` runs it on all files, which is why we only explicitly include files we know we can check.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -35,6 +35,8 @@ The current stable version of cocotb requires:
 
 .. note:: In order to use a 32-bit simulator you need to use a 32-bit version of Python.
 
+.. note:: Type checking cocotb code requires Python 3.11+.
+
 The installation instructions vary depending on your operating system:
 
 .. tab-set::
@@ -127,9 +129,9 @@ The installation instructions vary depending on your package manager:
 
       First, set up the `Guix-science channel <https://codeberg.org/guix-science/guix-science>`_. Then, in a terminal, run
 
-      .. code-block:: bash
+        .. code-block:: bash
 
-	  guix install python-cocotb
+	        guix install python-cocotb
 
 Post installation
 =================
--- a/noxfile.py
+++ b/noxfile.py
@@ -2,11 +2,10 @@
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
 import glob
-import os
 import shutil
 from contextlib import suppress
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple, cast
+from typing import Dict, List, Tuple, cast
 
 import nox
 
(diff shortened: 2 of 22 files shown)


ktbarrett:
Damn... looks like before Python 3.8 there was no check on the arguments to the `Callable` object, and after 3.8 they anticipated ParamSpec, but in 3.8 the check causes it to fail, so I'm not sure we can use ParamSpec at all until we drop 3.8...

docs/source/install.rst
```diff
@@ -127,9 +129,9 @@ The installation instructions vary depending on your package manager:
 
       First, set up the `Guix-science channel <https://codeberg.org/guix-science/guix-science>`_. Then, in a terminal, run
 
-      .. code-block:: bash
+        .. code-block:: bash
 
-	  guix install python-cocotb
+	        guix install python-cocotb
```
imphil, line 134: That's unintended?

docs/source/install.rst
```diff
@@ -35,6 +35,8 @@ The current stable version of cocotb requires:
 
 .. note:: In order to use a 32-bit simulator you need to use a 32-bit version of Python.
 
+.. note:: Type checking cocotb code requires Python 3.11+.
```
imphil, line 38: Maybe add that above to the Python 3.6.2+ line, e.g. 
"At least Python 3.6.2+. We recommend using Python 3.11 or newer to enable type checking."

noxfile.py
```diff
@@ -61,32 +58,21 @@ def simulator_support_matrix() -> List[Tuple[str, str, str]]:
...
-    sim: Optional[str], toplevel_lang: Optional[str], gpi_interface: Optional[str]
+    sim: str, toplevel_lang: str, gpi_interface: str
 ) -> Dict[str, str]:
     """Prepare the environment variables controlling the test run."""
-    e = {}
-    if sim is not None:
-        e["SIM"] = sim
-
-    if os.getenv("TOPLEVEL_LANG") is not None:
-        e["HDL_TOPLEVEL_LANG"] = os.getenv("TOPLEVEL_LANG")
```
imphil, line 82: A change in behavior, but probably for the better.

imphil (approved):
Nice, thanks!

docs/source/install.rst
ktbarrett: I caught this in the other PR and when I moved that commit here this fix was still in. There was a tab and the indentation was off and IIRC it was causing issue with Sphinx.

docs/source/install.rst
ktbarrett: I removed the note. Type checking is often done with a different Python installation or by a program that can parse newer versions of Python that what is running it, so this isn't something that end users need to be concerned about. I left the note in developing.rst, because Sphinx *needs* to read those version-ifdef'd code to generate correct type stubs in the docs.

PR #4677, opened 2025-05-14
Revert test discovery

ktbarrett:
### TODO

- [ ] newsfrag


ktbarrett:
I don't love the exact implementation, but it should do any API breaks that are required for 2.0 and we can clean up in 2.X once we have a better idea of what all this stuff should exactly look like.

These part 2 changes are designed to fix the issues:
* Test objects are not reusable (breaking 1.9 compatibility) because the test state and description were stuffed together. So the `Test` class was split in two: `Test` being stateless for describing the test and `RunningTest` for all state in a running test. `Test` can be reused and can be used to generate multiple `RunningTest` objects.
* Reverted the test discovery change so that `Test` objects are found by scraping the module variables instead of looking in the special `__cocotb_tests__` variable.
* Changed how tests generated with `cocotb.parametrized` are discovered by returning the object from the test decorator and picking up the whole object in the RegressionManager, only then calling `generate_test`.

This sets us up for the future:
* Stacking multiple `cocotb.parametrize` decorators for readability.
* Stacking multiple `cocotb.test` decorators or new decorators (like a `@cocotb.skipif`) to set `Test` fields.

While also reverting a lot of changes from 1.9.

cmarqu (approved):

PR #4682, opened 2025-05-16
Skip some tests for forks

### LINKED ISSUE #4681: Do not run all tests for forks
It is a bit annoying to get a mail every day that the eco-system tests fail. It is also a waste of compute that everyone that forks cocotb should run this test.

I think something like
```
if: github.repository_owner == 'cocotb'
```
at one or more suitable locations in https://github.com/cocotb/cocotb/blob/master/.github/workflows/ecosystem-compat.yml should do the trick. For matplotlib, the line is in the script, so the job is executed, just that it doesn't do anything unless it is run from the main repo. But one can probably read the documentation a bit and see if there are other solutions (like not even starting the test).

Maybe there are other jobs that also should have something similar?

Edit:
- [ ] Nightly ecosystem
- [ ] Tests requiring licenses
- [ ] Performance Benchmark
- [ ] Extended tool versions

One should check that e.g. PRs are run as cocotb and not the user though. But for sure the scheduled tests should be changed.

oscargus:
Added a group "licensed" for the tests that require a license and do not run those on forks (which do not have access to the dedicated runners).

Added a check to only run the scheduled tests on the main repo, not on forks.

Closes #4681

(Didn't really have time to fix this today so opened an issue first, but here goes...)

### DIFF
--- a/.github/workflows/ecosystem-compat.yml
+++ b/.github/workflows/ecosystem-compat.yml
@@ -21,6 +21,7 @@ on:
 
 jobs:
   cocotb-coverage:
+    if: github.repository == 'cocotb/cocotb'
     name: Test cocotb-coverage 1.2
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/ecosystem-compat.yml
```diff
@@ -59,53 +60,54 @@ jobs:
...
-            make -C pyuvm_clone/tests/cocotb_tests/t12_tlm sim checkclean
-            make -C pyuvm_clone/tests/cocotb_tests/t13_components sim checkclean
-            make -C pyuvm_clone/tests/cocotb_tests/config_db sim checkclean
-            make -C pyuvm_clone/tests/cocotb_tests/t14_15_sequences sim checkclean
-            make -C pyuvm_clone/tests/cocotb_tests/ext_classes sim checkclean
-            make -C pyuvm_clone/examples/TinyALU sim checkclean
-            make -C pyuvm_clone/examples/TinyALU_reg sim checkclean
-            make SIM=ghdl TOPLEVEL_LANG=vhdl -C pyuvm_clone/examples/TinyALU sim checkclean
+    if: github.repository == 'cocotb/cocotb'
+    name: Test pyuvm 3+
```
oscargus, line 64: The rest of the changes are unifying indentation with the first test...

ktbarrett (approved):
This looks generally like what I laid out in my comment, so I'm happy. @imphil?

PR #4683, opened 2025-05-16
Module argument(s): Slight wording change

mferris3CS1003:
Update to wording to clarify. This may already be a clear distinction to some but i had spent a good 20 minutes giving the module argument a list of strings

### DIFF
--- a/docs/source/building.rst
+++ b/docs/source/building.rst
@@ -232,7 +232,7 @@ Regression Manager
 
     The name of the Python module(s) to search for test functions -
     if your tests are in a file called ``test_mydesign.py``, ``COCOTB_TEST_MODULES`` would be set to ``test_mydesign``.
-    Multiple modules can be specified using a comma-separated list.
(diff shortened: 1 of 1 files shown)


docs/source/building.rst
```diff
@@ -232,7 +232,7 @@ Regression Manager
 
     The name of the Python module(s) to search for test functions -
     if your tests are in a file called ``test_mydesign.py``, ``COCOTB_TEST_MODULES`` would be set to ``test_mydesign``.
-    Multiple modules can be specified using a comma-separated list.
+    Multiple modules can be specified using a comma-separated String.
```
ktbarrett, line 235:
```suggestion
    Multiple modules can be specified using a comma-separated string.
```

ktbarrett (approved):
Seems fine to me.

imphil:
Maybe adding an example (or extending the existing one) would clarify it even more.

PR #4699, opened 2025-05-26
Properly handle `__all__` and re-exports

ktbarrett:
* Added `__all__` to all public modules and populated it correctly (pylance complains about modifying or partially specifying `__all__`)
* Fixed up `__module__` on all re-exported symbols
* Deprecated importing `Join` from `cocotb.triggers` as there should be exactly one canonical location for each object
* Removed re-exporting `TaskComplete` in `cocotb.triggers`. This is a new type, so it can just live forever in `cocotb.task`

One of these days I'll write a linter that ensures:
1. Exactly one canonical import location for each object
2. That object is included in the `__all__` of the canonical location
3. The `__module__` matches the canonical location

And then prints the full API in a tree so people can easily inspect if there's something that's public that shouldn't be.

### DIFF
--- a/src/cocotb/types/_logic.py
+++ b/src/cocotb/types/_logic.py
@@ -5,7 +5,6 @@ from functools import lru_cache
 from typing import (
     Dict,
     Set,
-    Tuple,  # noqa: F401  # Used by doctests, false positive
     Type,
     Union,
 )
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -118,7 +118,7 @@ codeautolink_autodoc_inject = True  # Inject an autolink-examples table to the e
 codeautolink_global_preface = textwrap.dedent("""\
     import cocotb
     from cocotb.clock import Clock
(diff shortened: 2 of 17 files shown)


src/cocotb/types/_logic.py
```diff
@@ -97,7 +96,7 @@ class Logic:
 
     .. code-block:: pycon3
 
-        >>> def full_adder(a: Logic, b: Logic, carry: Logic) -> Tuple[Logic, Logic]:
+        >>> def full_adder(a: Logic, b: Logic, carry: Logic) -> tuple[Logic, Logic]:
```
ktbarrett, line 99: This was done because of how `--doctest-modules` tests doctests. It evaluates the module where the object is defined at before running the test to pull in any dependencies for testing. Since we are fixing up the `__module__` this is breaking which module gets evaluated so it's `cocotb/types/__init__.py` instead of `_logic.py`. 

Luckily since we are building docs with only Python > 3.11 we don't even need this import anymore.

ktbarrett: Okay so Python 3.8 doesn't like that lmao.

PR #4711, opened 2025-05-31
Fix CI issues

### LINKED ISSUE #4708: Python 3.14 release package regression failing
Pinging @oscargus. This has been failing since 3.14 support was added. If this can't be resolved, 3.14 should be removed from the upcoming 2.0 release. Or we need some way to turn off doing release builds for unreleased versions of Python. 

https://github.com/cocotb/cocotb/actions/runs/15335716488/job/43152913734

### LINKED ISSUE #4709: Release package not being tested against licensed simulators
Pinging @oscargus. Since #4682 we are no longer running the release test against licensed simulators. This is another 2.0 blocking issue.

A copy of `build-test-release.yml` for licensed simulators should be made. (Maybe update the file names as well.)

oscargus:
Closes #4708 
Closes #4709 

Move 3.14-test to experimental.

I realised that we can just run two tests in the same file, so no need to duplicate all the skip-stuff. Also this gives a better dependency chain for the deployment.

Added `cancel-in-progress: true` to stop old test runs after pushing to the same branch again (quite useful for me at least, who tend to push before I think...)

Also added so that all the changed workflows executes on pull-requests. Will remove that once it is confirmed that it works by someone with correct bits.

I do not think the failing tests are caused by this PR?


.github/workflows/experimental.yml
```diff
@@ -14,6 +14,9 @@ on:
     - cron: '0 2 * * 1'
   # Allow triggering a CI run from the web UI.
   workflow_dispatch:
+  pull_request:
+    branches:
+      - master
```
oscargus, line 19: Will be removed.

ktbarrett:
The icarus v12 build tends to hang so we have it timeout after 5 minutes. It's a common failure you just need to re-run it.
 
> Skipping simulation as VHDL is not supported on simulator=vcs.

Lmao, so we are testing that it doesn't work? I think something got missed and we didn't realize because it's in the experimental bucket. Not your problem.

.github/workflows/build-test-release.yml
```diff
@@ -4,6 +4,10 @@
 
 name: Release
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
```
ktbarrett, line 8: Is there a reason to use `${{ github.event.number }}-${{ github.event.ref }}` instead of `${{ github.ref }}` like the docs mention?

ktbarrett (commented):
Looks good.

.github/workflows/build-test-release.yml
oscargus: Not sure. I copied it from Matplotlib.

.github/workflows/build-test-release.yml
```diff
@@ -4,6 +4,10 @@
...
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
oscargus: Done.

.github/workflows/build-test-release.yml
ktbarrett: Well if it's still working for them we can leave it and see if there's any weird discrepancies in the future.

ktbarrett (approved):

imphil:
Looks like there is more fallout from https://github.com/cocotb/cocotb/pull/4682 to be fixed:

- Add the "licensed" group to the comment at the top of generate-envs.py.
- "licensed" is a bit an unfortunate name, it's not different from CI, it's a subset of the CI tasks (e.g., the extended tests also have licensed simulators in them). Make sure to at least explain that properly in the comment.

oscargus:
Should I rename them in some way? `ci-opensource` and `ci-licensed`?

ktbarrett:
If so maybe `ci-free` and `ci-licensed`. Shorter and Free is not necessarily Open Source.

oscargus:
If this works, I think everything is handled. (If not, I'll be at it tomorrow.)

.github/generate-envs.py
```diff
@@ -20,13 +20,14 @@
 - extra_name: Additional tag prepended to computed name for test. Default: <none>.
 
 What tests belong in what groups:
-- ci: The most recent stable release of a given simulator, all supported versions of Python, and all supported operating systems. Run on all PRs and master pushes.
+- ci-free: The most recent stable release of a given free simulator, all supported versions of Python, and all supported operating systems. Run on all PRs and master pushes.
+- ci-licensed: The most recent stable release of a given licensed simulator, all supported versions of Python, and all supported operating systems. Run on all PRs and master pushes in the cocotb repo, but are skipped in forks.
```
ktbarrett:
```suggestion
- ci-licensed: The most recent stable release of a given licensed simulator. Run on all PRs and master pushes in the cocotb repo, but are skipped in forks.
```

ktbarrett:
Well Questa license server is down. I'll try rerunning the tests in an hour. 

EDIT: No need, my other PR failed. It's down.

oscargus:
Update with the rewording. Also reverted a search-and-replace change in .pre-commit-config.yml that shouldn't have been there.

ktbarrett (approved):
It's good, but I want to ensure Questa checks work before merging. That may or may not be done any time soon.

imphil:
Questa tests are rerunning now.

PR #4717, opened 2025-06-02
Scheduler Refactor

ktbarrett:
Replacement for #4280. Focusing solely on Trigger, Task, and Scheduler refactors, as the other refactors mentioned in that PR have been done separately.

This PR reorganizes all the core components of cocotb: Triggers, Tasks, and the Scheduler. The intent is to provide more functionality to the user by introducing ways to register arbitrary callbacks on triggers and in the event loop rather than be solely tied to resuming Tasks. It also seeks to reduce the amount of state/sanity checking necessary when scheduling Tasks by moving all state tracking to the Task object itself, which should help reduce runtime overhead.

## Triggers

Will now keep track of all the callbacks that must be called when they fire rather than the global `trigger2tasks` dict in the scheduler. This also means that each Trigger will have it's own `_react` method rather than the global one in the scheduler. This allows each Trigger to add custom behavior like ReadWrite flushing queued writes.

It will have a new private interface for adding and removing callbacks on that trigger. `_prime` and `_unprime` will remain largely the same, but will be called more infrequently. There is a new `_register` and `_deregister` function which adds and removes callbacks on that Trigger, calling `_prime` and `_unprime` on the first add and final removal. 

Registered callbacks will be called immediately when a Trigger fires (not scheduled with the event loop).

The registration function will return a handle to the callback in the Trigger object to uniquely identify the callback in the handle.

## Scheduler (now `EventLoop`)

Most of the functionality of the scheduler has been moved to either the Triggers, Tasks, or Test object. Additionally because there are more things than Tasks that need to be queued, the scheduler has been turned into a more generic callback scheduler called `EventLoop`.

The scheduler allows users to scheduler a callback in a queue. The result will be a unique handle to the registered callback which will support `cancel()`ing before the callback is run. Then the EventLoop can be "run" until the queue is exhausted.

## Tasks

`Task._advance` (now `Task._resume`) will no longer hand Triggers back to the Scheduler, but instead be in charge of calling `Trigger._register` when Triggers are yielded. The Task will register `Task._schedule_resume` to be called when Triggers fire, which will allow them to immediately update their state, before scheduling a call to `Task._resume` with the event loop.

`cocotb.start_soon` and `await Task()` will call a new `_ensure_started()` method which will ensure the Task is not unstarted and will schedule it for the first time otherwise.

`Task.kill`, `Task.cancel` and `Task._cancel_now` are now required to call `Trigger._deregister` or `ScheduledCallback.cancel` when cancelling a Task, depending on what state they are in.

All `Task._state` state variable updates will be handled by Task and not by the Scheduler.

The `cocotb.task` module will have a `current_task()` method to support introspecting the current task. When `Task._resume` is called it will update the current task.

## `cocotb.bridge` objects

These were left as is for now. The major change was that all related code in the scheduler was moved into the `cocotb._bridge` module.

Future cleanup might refactor `external_waiter` objects into being Task-like objects that interact with the event loop in the normal way. This will prevent having to pay the cost of checking this stuff in the event loop. You'll only pay for it if you use it.

### TODO

- [x] Investigate using different data structure (OrderedDict) for Trigger callback handles
- [x] Investigate reducing the overhead of `if DEBUG` by doing wrappers or "ifdef"ing out functions (not worth it)
- [x] ~~Investigate not supporting calling `_react` in `_prime` to avoid "special" order dependent paths in Triggers and Tasks~~ (follow on)
- [x] Get bridge/resume objects working
- [x] ~~Get disabled tests working again~~ (test are over-specific)

### DIFF
--- a/tests/test_cases/test_cocotb/test_deprecated.py
+++ b/tests/test_cases/test_cocotb/test_deprecated.py
@@ -245,3 +245,64 @@ async def test_Lock_name_deprecated(_: object) -> None:
         l.name = "foobar"
     with pytest.warns(DeprecationWarning):
         assert l.name == "foobar"
+
+
+@cocotb.test
+async def test_Task_kill_unstarted(_: object) -> None:
+    async def example() -> None:
+        await Timer(10, "ns")
+
+    task = cocotb.create_task(example())
+
+    with pytest.warns(DeprecationWarning):
+        task.kill()
+
+    assert task.done()
+    assert task.result() is None
+
+
+@cocotb.test
+async def test_Task_kill_scheduled(_: object) -> None:
+    async def example() -> None:
+        await Timer(10, "ns")
+
+    task = cocotb.start_soon(example())
+
+    with pytest.warns(DeprecationWarning):
+        task.kill()
+
+    assert task.done()
+    assert task.result() is None
+
+
+@cocotb.test
+async def test_Task_kill_pending(_: object) -> None:
+    async def example() -> None:
+        await Timer(10, "ns")
+
+    task = cocotb.start_soon(example())
+    await Timer(1)
+
+    with pytest.warns(DeprecationWarning):
+        task.kill()
+
+    assert task.done()
+    assert task.result() is None
+
+
+@cocotb.test
+async def test_Task_kill_done(_: object) -> None:
+    async def example() -> int:
+        return 1
+
+    task = cocotb.start_soon(example())
+    await Timer(1)
+
+    assert task.done()
+    assert task.result() == 1
+
+    with pytest.warns(DeprecationWarning):
+        task.kill()
+
+    assert task.done()
+    assert task.result() == 1
--- a/tests/test_cases/test_cocotb/test_scheduler.py
+++ b/tests/test_cases/test_cocotb/test_scheduler.py
@@ -1056,3 +1056,29 @@ async def test_task_local_variables(_: object) -> None:
 
     task = cocotb.start_soon(coro())
     await task
+
+
+@cocotb.test(expect_error=RuntimeError)
+async def test_Task_ignored_CancelledError_return(_: object) -> None:
+    async def bad() -> None:
+        try:
+            await Timer(10, "ns")
+        except CancelledError:
+            return
+
+    task = cocotb.start_soon(bad())
+    await Timer(1)
+    task.cancel()
+
+
+@cocotb.test(expect_error=RuntimeError)
+async def test_Task_ignored_CancelledError_await(_: object) -> None:
+    async def bad() -> None:
+        try:
+            await Timer(10, "ns")
+        except CancelledError:
+            await Timer(1)
+
+    task = cocotb.start_soon(bad())
+    await Timer(1)
+    task.cancel()


src/cocotb/_base_triggers.py
```diff
@@ -21,56 +22,113 @@
...
+        self._args = args
+        self._kwargs = kwargs
+
+
 class Trigger:
     """A future event that a Task can wait upon."""
 
     def __init__(self) -> None:
-        self._primed = False
+        self._callbacks: List[TriggerCallback] = []
```
ktbarrett: Consider `OrderedDict` for fast removal.

ktbarrett: `dict` is actually sufficient. We don't need `popitem`.

ktbarrett:
Just tested against a very large codebase with minimal issues.

src/cocotb/_base_triggers.py
```diff
@@ -8,64 +8,120 @@
...
+                    cb._func,
+                    cb._args,
+                    cb._kwargs,
+                )
+            cb._func(*cb._args, **cb._kwargs)
+
+    def _react(self) -> None:
+        """Function called when a Trigger fires.
+
+        Expected to call _do_callbacks() and any cleanup of the underlying mechanism
```
cmarqu:
```suggestion
        Expected to call ``_do_callbacks()`` and any cleanup of the underlying mechanism
```

src/cocotb/_base_triggers.py
```diff
@@ -8,64 +8,120 @@
...
-        Sub-classes may override this, but should end by calling the base class
-        method.
+    @abstractmethod
+    def _prime(self) -> None:
+        """Enables the underlying mechanism for the Trigger to fire.
 
-        .. warning::
-            Do not call this directly within a :term:`task`. It is intended to be used
-            only by the scheduler.
+        The underlying mechanism should call this Trigger's _react() when it fires.
```
cmarqu:
```suggestion
        The underlying mechanism should call this Trigger's ``_react()`` when it fires.
```

src/cocotb/_base_triggers.py
```diff
@@ -8,64 +8,120 @@
...
         """
-        self._cleanup()
 
-    def _cleanup(self) -> None:
-        # Clear _primed so this Trigger can be re-primed.
-        self._primed = False
+    @abstractmethod
+    def _unprime(self) -> None:
+        """Disables the underlying mechanism for the Trigger to fire."""
```
cmarqu:
```suggestion
        """Disable the underlying mechanism for the Trigger to fire."""
```

src/cocotb/_base_triggers.py
```diff
@@ -8,64 +8,120 @@
...
-
-        After being un-primed, a Trigger may be re-primed again in the future.
-        Calling `_unprime` multiple times is allowed, subsequent calls should be
-        a no-op.
-
-        Sub-classes may override this, but should end by calling the base class
-        method.
+    @abstractmethod
+    def _prime(self) -> None:
+        """Enables the underlying mechanism for the Trigger to fire.
```
cmarqu:
```suggestion
        """Enable the underlying mechanism for the Trigger to fire.
```

src/cocotb/_base_triggers.py
```diff
@@ -289,20 +329,18 @@ def __init__(self, parent: "Lock") -> None:
...
     def _unprime(self) -> None:
-        if not self._primed:
-            return
         self._parent._unprime_lock(self)
-        return super()._unprime()
+
+    def __await__(self) -> "Generator[Self, None, Self]":
+        if self._parent._is_used(self):
+            raise RuntimeError(
+                "Lock.acquire() result can only be used by one task at a time"
```
cmarqu, line 342: Would it be helpful to print which task already has the lock?

src/cocotb/_bridge.py
```diff
@@ -184,3 +188,116 @@ def thread_wait(self) -> external_state:
                 )
 
         return self.state
+
+
+pending_threads: List[external_waiter[Any]] = []
+
+
+def queue_function(task: Coroutine[Trigger, None, Result]) -> Result:
+    """Queue a task for execution and move the containing thread
```
cmarqu:
```suggestion
    """Queue *task* for execution and move the containing thread
```

src/cocotb/_bridge.py
```diff
@@ -184,3 +188,116 @@ def thread_wait(self) -> external_state:
...
+        pending_threads.append(waiter)
+
+        await waiter.event.wait()
+
+        return waiter.result  # raises if there was an exception
+
+    return wrapper()
+
+
+def run_bridge_threads() -> None:
```
cmarqu, line 277: Missing docstring.

src/cocotb/_test.py
```diff
@@ -149,9 +150,16 @@ def start_soon(
 
     .. versionadded:: 1.6
     """
-    task = create_task(coro, name=name)
-    cocotb._scheduler_inst._schedule_task(task)
-    return task
+    if not isinstance(coro, Task):
+        coro = create_task(coro, name=name)
+    # If a task is passed in, we it to:
```
cmarqu: A verb is missing here.

src/cocotb/_event_loop.py
cmarqu, line 1: Add docstrings and add the module to `library_reference.rst`?

src/cocotb/_base_triggers.py
ktbarrett: Yeah. But we don't have this info here. Triggers are currently agnostic to Tasks.

src/cocotb/_event_loop.py
ktbarrett: I decided to make it private instead.

src/cocotb/_bridge.py
```diff
@@ -184,3 +188,117 @@ def thread_wait(self) -> external_state:
...
 
         return self.state
+
+
+pending_threads: List[external_waiter[Any]] = []
+
+
+def queue_function(task: Coroutine[Trigger, None, Result]) -> Result:
+    """Queue *task* for execution and switch back to main thread."""
+    # We should be able to find ourselves inside the _pending_threads list
```
marlonjames: nit:
```suggestion
    # We should be able to find ourselves inside the pending_threads list
```

ktbarrett:
- [x] Make `TriggerCallback` cancellable
- [x] ~~Investigate supporting only no-argument functions as callbacks in the EventLoop and Triggers~~
    - This is the common case and lambdas for binding could have lower overhead than packing and unpacking args and kwargs several times

src/cocotb/_event_loop.py
```diff
@@ -0,0 +1,86 @@
...
+        self._callbacks: Deque[ScheduledCallback] = deque()
+        if debug.debug:
+            self._cycles: int = 0
+
+    @cached_property
+    def log(self) -> logging.Logger:
+        return logging.getLogger("cocotb.event_loop")
+
+    def run(self) -> None:
+        while self._callbacks:
```
marlonjames, line 46: Shouldn't `_cycles` be set to 0 at the start of `run()`?

src/cocotb/_test.py
```diff
@@ -224,11 +232,7 @@ def create_task(
 
     .. versionadded:: 1.6
     """
-    if isinstance(coro, Task):
-        if name is not None:
-            coro.set_name(name)
-        return coro
-    elif isinstance(coro, Coroutine):
+    if isinstance(coro, Coroutine):
```
marlonjames: Why remove `Task` support? That would be a breaking change for https://docs.cocotb.org/en/stable/library_reference.html#cocotb.create_task

src/cocotb/_event_loop.py
ktbarrett:
```suggestion
        self._cycles = 0
        while self._callbacks:
```

src/cocotb/_event_loop.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def cancel(self) -> None:
+        self._cancelled = True
+
+
+class EventLoop:
+    def __init__(self) -> None:
+        self._callbacks: Deque[ScheduledCallback] = deque()
+        if debug.debug:
+            self._cycles: int = 0
```
ktbarrett:
```suggestion
        self._cycles: int = 0
```

src/cocotb/_test.py
ktbarrett: Yeah I suppose it is, IDR why I changed this, it was a while ago.

src/cocotb/_base_triggers.py
marlonjames: You could get the current task in `Lock.acquire()` and pass it in to the `_Lock`.
Edit: I guess that doesn't help, we want the task that's waiting not who created the `_Lock`.

src/cocotb/task.py
```diff
@@ -301,23 +322,40 @@ def _advance(self, exc: Union[BaseException, None]) -> Union[Trigger, None]:
...
+                    self._schedule_resume(remove_traceback_frames(e, ["_resume"]))
 
-    def _schedule_resume(self, exc: Optional[BaseException] = None) -> None:
-        cocotb._scheduler_inst._unschedule(self)
-        cocotb._scheduler_inst._schedule_task_internal(self, exc)
+        finally:
+            _current_task = None
 
     @deprecated("`task.kill()` is deprecated in favor of `task.cancel()`")
     def kill(self) -> None:
```
marlonjames, line 354: We should have some tests to cover `kill()` usage.

src/cocotb/task.py
marlonjames, line 325: Do we really not have a test for this? Coverage is missing.

marlonjames (approved):
🎉

src/cocotb/task.py
```diff
@@ -230,9 +234,15 @@ def __repr__(self) -> str:
         else:
             raise RuntimeError("Task in unknown state")
 
+    def _ensure_started(self) -> None:
+        if self._state is _TaskState.UNSTARTED:
+            self._schedule_resume()
```
ktbarrett, line 240:
```suggestion
    def _ensure_started(self) -> None:
        # start if unstarted
        if self._state is _TaskState.UNSTARTED:
            self._schedule_resume()
        # fail if it's already done
        elif self.done():
            raise RuntimeError("Cannot start a finished Task")
        # it's already running
```

src/cocotb/_test.py
```diff
@@ -149,9 +150,16 @@ def start_soon(
...
-    cocotb._scheduler_inst._schedule_task(task)
-    return task
+    if not isinstance(coro, Task):
+        coro = create_task(coro, name=name)
+    # If a task is passed in, we want it to:
+    # - start if unstarted
+    # - fail if it's already done
+    # - do nothing otherwise
+    if coro.done():
+        raise RuntimeError("Cannot start a finished Task")
```
ktbarrett:
```suggestion
```

src/cocotb/task.py
```diff
@@ -230,9 +234,20 @@ def __repr__(self) -> str:
         else:
             raise RuntimeError("Task in unknown state")
 
+    def _ensure_started(self) -> None:
+        # start if unstarted
+        if self._state is _TaskState.UNSTARTED:
+            self._schedule_resume()
+        # fail if it's already done
```
marlonjames, line 241: `await`ing the `Task` to get the result calls this and it fails 🙃

ktbarrett: Oop, I guess I forgot about the done branch to immediately return if there's a result.

PR #4721, opened 2025-06-03
iverilog wave dump - allow user to specify dump file location

### LINKED ISSUE #4720: iverilog wave dump - allow user to specify dump file location
iverilog requires the build step to create a verilog shim to dump the waves to a file. This shim can't then be modified at test time.

Currently, iverlog wave dumps are hard-coded to go to `<build_dir>/<hdl_toplevel>.fst`:
https://github.com/cocotb/cocotb/blob/master/src/cocotb_tools/runner.py#L696

This causes issues when running under a build system like bazel, where the test/run directory can't be the same as the build directory, and the build directory is read-only for the test.

The best solution I've found for this is to use plusargs, so that the user can specify the wave dump file location at run time.

This solution still means iverilog behaves differently from other simulators with regard to waves, but it doesn't change the existing behaviour, and presents a purely optional means to change the output location for those who need it.

will-keen:
iverilog requires the build step to create a verilog shim to dump the waves to a file. This shim can't then be modified at test time.

This causes issues when running under a build system like bazel, where the test/run directory can't be the same as the build directory, and the build directory is read-only for the test.

The best solution I've found for this is to use plusargs, so that the user can specify the wave dump file location at run time.

This solution still means iverilog behaves differently from other simulators with regard to waves, but it doesn't change the existing behaviour, and presents a purely optional means to change the output location for those who need it.

Closes #4720

### DIFF
--- a/docs/source/newsfragments/4721.feature.rst
+++ b/docs/source/newsfragments/4721.feature.rst
@@ -1,0 +1,1 @@
+Allow user to specify iverilog wave dump file location at run time with a plusarg.
--- a/docs/source/simulator_support.rst
+++ b/docs/source/simulator_support.rst
@@ -45,6 +45,10 @@ To enable FST tracing, set :make:var:`WAVES` to ``1``.
 
     make SIM=icarus WAVES=1
 
+By default, the wave file will be named `<hdl_toplevel>.fst`. Unlike other simulators, it will be placed in the build directory, rather than the test directory.
+
+To redirect the wave file to a different location, use the plusarg `dumpfile_path` when running the test.
+
 .. _sim-icarus-issues:
 
 Issues for this simulator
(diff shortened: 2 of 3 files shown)


ktbarrett (commented):
Check out `docs/source/newsfragments/README.rst` and add a newsfrag for your change so other people see it in the changelog.

Looks good!

will-keen:
> Check out `docs/source/newsfragments/README.rst` and add a newsfrag for your change so other people see it in the changelog.
> 
> Looks good!

Thanks! Added the newsfragment. Hope it's suitable.

will-keen:
@ktbarrett , the CI tests seem to be failing due to licence checkout issues for questa. As a non-maintainer I can't retrigger them. Is this a known issue? Should I just hang tight for now?

ktbarrett:
Yeah they've been failing for a couple days. I can't fix them either. Luckily your change doesn't affect Questa. We usually give the other maintainers a couple days to chrck the PR out for themselves. If no one else approves and merges in a couple days, I will.

themperek (approved):

will-keen:
Thanks @ktbarrett and @themperek !

PR #4723, opened 2025-06-04
Deprecate `Lock` `name` field like `Event`

### LINKED ISSUE #4715: Deprecate `Lock(name=...)`
We deprecated passing `name` to `Event` so that it matches asyncio, we should do the same for `Event`.

ktbarrett:
Closes #4715. Doing the same things as #4561. asyncio's `Lock` does not have a name. If we hope to improve interface compatibility between the two so we may someday replace the scheduler with asyncio, this will be necessary.

### DIFF
--- a/src/cocotb/_base_triggers.py
+++ b/src/cocotb/_base_triggers.py
@@ -150,22 +150,27 @@ class Event:
         self._event: _Event = _Event(self)
         self._name: Union[str, None] = None
         if name is not None:
+            warnings.warn(
+                "The 'name' argument will be removed in a future release.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
             self.name = name
         self._fired: bool = False
         self._data: Any = None
 
     @property
-    @deprecated("The name field will be removed in a future release.")
+    @deprecated("The 'name' field will be removed in a future release.")
     def name(self) -> Union[str, None]:
         """Name of the Event.
 
         .. deprecated:: 2.0
-            The name field will be removed in a future release.
+            The *name* field will be removed in a future release.
         """
         return self._name
 
     @name.setter
-    @deprecated("The name field will be removed in a future release.")
+    @deprecated("The 'name' field will be removed in a future release.")
     def name(self, new_name: Union[str, None]) -> None:
         self._name = new_name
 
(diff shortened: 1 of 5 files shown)


cmarqu:
Should this get a newsfragment?

ktbarrett:
Yeah. Guess I forgot to add one after I pushed. I wish it were easier to tell when the next ref number would be so I wouldn't have to push, add newsfrag, then push again.

src/cocotb/_base_triggers.py
```diff
@@ -150,6 +150,11 @@ def __init__(self, name: Optional[str] = None) -> None:
         self._event: _Event = _Event(self)
         self._name: Union[str, None] = None
         if name is not None:
+            warnings.warn(
+                "The name argument will be removed in a future release.",
```
imphil:
```suggestion
                "The 'name' argument will be removed in a future release.",
```

src/cocotb/_base_triggers.py
```diff
@@ -342,9 +347,31 @@ class Lock(AsyncContextManager[None]):
...
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            self._name = name
         self._locked: bool = False
 
+    @property
+    @deprecated("The name field will be removed in a future release.")
+    def name(self) -> Union[str, None]:
+        """Name of the Event.
```
ktbarrett:
```suggestion
        """Name of the Lock.
```

src/cocotb/_base_triggers.py
```diff
@@ -342,9 +347,31 @@ class Lock(AsyncContextManager[None]):
 
     def __init__(self, name: Optional[str] = None) -> None:
         self._pending_primed: List[_Lock] = []
-        self.name: Optional[str] = name
+        self._name: Union[str, None] = None
+        if name is not None:
+            warnings.warn(
+                "The name argument will be removed in a future release.",
```
ktbarrett:
```suggestion
                "The 'name' argument will be removed in a future release.",
```

src/cocotb/_base_triggers.py
ktbarrett:
```suggestion
                "The 'name' argument will be removed in a future release.",
```

PR #4730, opened 2025-06-06
Add `cocotb.task.current_task`

ktbarrett:
Split from #4717. This could come in now as it's a pretty simple change.


cmarqu (approved):

ktbarrett:
Needs to be added to the reference manual

PR #4732, opened 2025-06-09
NVC Runner fixes

ktbarrett:
Sets the logger level to INFO so we get INFO diagnostics about things like exact commands being run out which the Makefiles do. Then fixes a bug in the NVC runner where relative paths were used in commands which breaks when `test_dir` is set.

### DIFF
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -1115,7 +1115,12 @@ class Nvc(Runner):
                 )
 
         cmds = [
-            ["nvc", f"--work={self.hdl_library}", "-L", str(self.build_dir)]
+            [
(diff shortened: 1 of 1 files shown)


cmarqu (approved):

ktbarrett:
Wait, there's more! https://github.com/nickg/nvc/issues/1215

EDIT: Well I do have a path forward and I answered my question about the `--work` option separator so I think we are good here for now. I've tested this locally.

ktbarrett:
Seems like adding the path to the worklib is breaking some things. I'll need to investigate.

ktbarrett:
I just tested and this works when setting the `test_dir` as well.

PR #4734, opened 2025-06-09
Use `pathlib` instead of `os` functionality

ktbarrett:
`pathlib` seems to be the preferred way of doing things these years, and there exists `flake8-use-pathlib` and the `PTH` ruff rules to enforce this. So I added that rule and manually fixed the few uses of it in the repo.


src/cocotb/_init.py
```diff
@@ -185,7 +188,7 @@ def _start_user_coverage() -> None:
                 log.info(
                     "Collecting coverage of user code. No coverage config file supplied via COCOTB_COVERAGE_RCFILE."
                 )
-                cocotb_package_dir = os.path.dirname(__file__)
+                cocotb_package_dir = Path(__file__).parent
```
ktbarrett:
```suggestion
                cocotb_package_dir = Path(__file__).parent.absolute()
```

src/pyproject.toml
cmarqu, line 1: Should the pyproject.toml linter in pre-commit be extended to this file (or is it doing that already?)

(On a phone, can't easily check)

ktbarrett: ruff handles that automatically. When it enters a directory it sees if there's a pyproject.toml for config and if there is it uses it's configuration overriding the parent directory's. But we "override" it here with an extend.

cmarqu: Misunderstanding, I meant if our use of https://github.com/henryiii/validate-pyproject-schema-store in pre-commit is also checking this new file.

ktbarrett: Ah. Good point, I'll check.

PR #4737, opened 2025-06-09
Use ellipses instead of `pass`

ktbarrett:
It prevents confusion with "passing a test" in docs. Also deleted a couple uses in the source. `...` lines are excluded from coverage by default so we can remove the pragmas.

### DIFF
--- a/src/cocotb/_decorators.py
+++ b/src/cocotb/_decorators.py
@@ -357,7 +357,8 @@ def test(
                     @cocotb.test(**decorator_kwargs)
                     @functools.wraps(obj)
                     async def test(dut, **test_kwargs):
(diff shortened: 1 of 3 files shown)


src/cocotb/_decorators.py
```diff
@@ -357,7 +357,8 @@ def decorator(obj):
                     @cocotb.test(**decorator_kwargs)
                     @functools.wraps(obj)
                     async def test(dut, **test_kwargs):
-                        pass  # your code here
+                        # your code here
+                        ...
```
ktbarrett, line 361: Making it `...  # your code here` would cause ruff to move it to the preceding line.

PR #4741, opened 2025-06-10
Use short "a | b" syntax for Literal types in docs

cmarqu:
Improves readability, see https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-python_display_short_literal_types

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -110,6 +110,9 @@ autoclass_content = "both"
 autodoc_typehints = "description"  # show type hints in the list of parameters
 autodoc_typehints_description_target = "documented"
 
+# use short "a | b" syntax for Literal types
+python_display_short_literal_types = True
+
 # Options for automatic links from code examples to reference docs
 # (https://sphinx-codeautolink.readthedocs.io/)
 codeautolink_warn_on_missing_inventory = True


cmarqu:
Hmm.
Works for e.g. https://cocotb--4741.org.readthedocs.build/en/4741/library_reference.html#cocotb.clock.Clock.unit but not for parameters like for `unit` in https://cocotb--4741.org.readthedocs.build/en/4741/library_reference.html#cocotb.triggers.Timer

cmarqu:
Support for autodoc was supposed to be in Sphinx 8.2 via https://github.com/sphinx-doc/sphinx/pull/12003

ktbarrett:
Yeah the Timer argument needs to be fixed. Check out #4742.

cmarqu:
How do we best go forward? Merge this, then see if #4742 fixes the rest?

cmarqu:
`byteorder` in https://cocotb--4741.org.readthedocs.build/en/4741/library_reference.html#cocotb.types.LogicArray.from_bytes is also not replaced.

ktbarrett:
This could also be due to the fact that the Literals are in strings. It might be worth seeing what happens when you remove the strings (it will break Python <3.10, but just to test).

> How do we best go forward? Merge this, then see if https://github.com/cocotb/cocotb/pull/4742 fixes the rest?

I don't think it really matters, we will likely need another follow-up PR.

Also, I don't really like how the new theme doesn't put the types in function prototype.

ktbarrett (approved):

cmarqu:
> This could also be due to the fact that the Literals are in strings. It might be worth seeing what happens when you remove the strings (it will break Python <3.10, but just to test).

Removing the strings makes no difference, but this change does:
```diff
diff --git i/src/cocotb/types/_logic_array.py w/src/cocotb/types/_logic_array.py
index b4b407ab..c769d841 100644
--- i/src/cocotb/types/_logic_array.py
+++ w/src/cocotb/types/_logic_array.py
@@ -362,7 +362,7 @@ def from_bytes(
         value: Union[bytes, bytearray],
         range: Union[Range, int, None] = None,
         *,
-        byteorder: "Literal['big'] | Literal['little']",
+        byteorder: "Literal['big', 'little']",
     ) -> "LogicArray":
         """Construct a :class:`!LogicArray` from :class:`bytes`.
 
@@ -654,7 +654,7 @@ def to_signed(self) -> int:
     def to_bytes(
         self,
         *,
-        byteorder: "Literal['big'] | Literal['little']",
+        byteorder: "Literal['big', 'little']",
     ) -> bytes:
         """Convert the value to bytes.
 
```

cmarqu:
> Also, I don't really like how the new theme doesn't put the types in function prototype.

This is due to our setting `autodoc_typehints = "description"` (https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints)
When removing this config setting, some signatures become quite unreadable IMO.

PR #4743, opened 2025-06-11
Fix instances of union of literals

ktbarrett:
Addresses https://github.com/cocotb/cocotb/pull/4741#issuecomment-2963065685.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,7 @@ extend-select = [
     "PERF101",  # unnecessary list cast
     "PERF102",  # unnecessary dict.items()
     "B007",     # unused loop variable
+    "PYI030",   # unnecessary literal union
 ]
 ignore = [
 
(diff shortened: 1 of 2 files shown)


cmarqu:
Looks good now: https://cocotb--4743.org.readthedocs.build/en/4743/library_reference.html#cocotb.types.LogicArray.from_bytes

ktbarrett:
@cmarqu  I'm still seeing instances of `Literal` in the docs. I'm thinking the real issue is that it doesn't like that it's in a TypeAlias. It's resolving the value of the type alias, but not applying the style change after the fact. That's probably an autodoc bug that should be reported.

cmarqu:
> @cmarqu I'm still seeing instances of `Literal` in the docs.

I am on it. I believe there are imports for `Literal` missing even though it's only indirectly used. The problem is only showing in Napoleon-interpreted `Args:`, it works without the import of `Literal`  when autodoc is instructed to put the type hints into the function signature.

PR #4744, opened 2025-06-11
Use `setup-nvc` to run NVC CI tests with pre-built binaries

### LINKED ISSUE #4735: Install pre-built NVC binaries in CI
NVC includes ubuntu 22.04 `.deb` installers in its releases so we can skip the build step for NVC on many tests.

ktbarrett:
Closes #4735.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -176,8 +176,14 @@ jobs:
         sudo make install
 
       # Install NVC
-    - name: Set up NVC (Ubuntu)
-      if: startsWith(matrix.os, 'ubuntu') && matrix.sim == 'nvc'
+    - name: Set up NVC (Ubuntu, release)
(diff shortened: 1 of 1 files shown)


ktbarrett:
@nickg Any idea here? https://github.com/cocotb/cocotb/actions/runs/15588544309/job/43901241634?pr=4744#step:17:21

Are releases here specified without the `r` or something else?

nickg:
> Are releases here specified without the r or something else?

Yeah it should be "1.16.0" without the r. Although I've just released 1.16.2 which has some VHPI fixes.

ktbarrett:
Well since github actions doesn't provide anything like a "split" function, I think I'll have to make another variable for simulator provider to separate the two.

nickg:
It's pretty trivial to strip off any "r" prefix in the action itself, I'll do that later.

nickg:
@ktbarrett I've done that now - try rerunning?

ktbarrett:
Looks good. Thanks @nickg.

> Although I've just released 1.16.2 which has some VHPI fixes.

I'll pull those in in another PR.

PR #4745, opened 2025-06-11
Import typing.Literal to improve doc rendering

cmarqu:

### DIFF
--- a/src/cocotb/_gpi_triggers.py
+++ b/src/cocotb/_gpi_triggers.py
@@ -32,6 +32,9 @@ from cocotb.utils import get_sim_steps, get_time_from_sim_steps
 if sys.version_info >= (3, 11):
     from typing import Self
 
+if sys.version_info >= (3, 10):
+    from typing import Literal  # noqa: F401  # used in type strings
+
 
 class GPITrigger(Trigger):
     """A trigger for a simulation event."""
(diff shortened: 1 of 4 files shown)


src/cocotb/_gpi_triggers.py
```diff
@@ -32,6 +32,9 @@
 if sys.version_info >= (3, 11):
     from typing import Self
 
+if sys.version_info >= (3, 10):
+    from typing import Literal  # noqa: F401  # used in type strings
```
ktbarrett, line 36: Christ. So ruff is incorrectly decided that it wasn't used and removed it because it was in a TypeAlias?

ktbarrett (approved):

src/cocotb/_gpi_triggers.py
cmarqu: Quite a rabbit hole what was supposed to be a simple config change :)

ktbarrett: I reported the issue upstream a while ago. https://github.com/astral-sh/ruff/issues/17455. Now it's really bitten us instead of just being a nuisance.

cmarqu:
Still three occurrences of `Literal[...]` left in the rendered docs, and here only de-stringifying seems to help.

ktbarrett:
@cmarqu Is this still an issue?

cmarqu:
It's not.

PR #4746, opened 2025-06-11
Map `vpiPackedArray` objects to LogicArrayObject

ktbarrett:


imphil:
Can you have a look why our tests didn't catch that? E.g. https://github.com/cocotb/cocotb/blob/master/tests/test_cases/test_struct/test_struct.py

ktbarrett:
@imphil Xcelium returns `vpiPackedArrayNet` when the type is a packed array of packed structs. We aren't testing that anywhere. That should be in test_multi_dimensional_array, but the packed struct aspect of that test was never completed AFAICT. Probably because Icarus didn't support it.

ktbarrett:
Added tests for this. [See the original failure](https://github.com/cocotb/cocotb/actions/runs/15626784807/job/44022478628#step:27:6114).

Riviera is now failling though. It takes a complete dump trying to access packed structs.

cmarqu (approved):

PR #4750, opened 2025-06-13
Use normal dict for inertial write cache

ktbarrett:
`OrderedDict` isn't necessary since we really aren't dependent upon the exact behavior of `popitem`. This should improve performance a touch.


cmarqu:
The nvc CI fail is a Python 3.14 thing?

ktbarrett:
@cmarqu The coverage module doesn't support 3.14 yet apparently. We probably jumped the gun on that by adding it to the regression before the beta was even released.

PR #4751, opened 2025-06-13
Replace uses of `Path.as_posix` with OS-specific string

### LINKED ISSUE #4733: Replace uses of `as_posix` in Runner
There are several places in `runner.py` where paths are converted to strings using `as_posix()` instead of `str()`. This forces the string to use POSIX style paths instead of native paths, which I have a feeling makes Windows unhappy. Seeing as supporting Windows better was the whole point of the Python Runner, this seems like a necessary improvement.

### LINKED ISSUE #4752: Add option to select path output format to `cocotb-config`
The mingw64 gmake build available in the Windows conda environment expects paths to be POSIX style and does not work correctly otherwise. This is currently requiring us to hardcode POSIX paths in `cocotb_config.py` and `runner.py`. A better solution would be to use the default string repr for paths on a given system and add a commandline option to turn on POSIX outputs for environments like mingw64 via conda.

We should probably turn on this option by default in calls to `cocotb-config` in the Makefiles as we can expect that anyone running Makefiles is in a POSIX environment like MSYS or conda+mingw64.

ktbarrett:
Closes #4733. Closes #4752.

- [x] ~~Test manually in MSYS and MINGW environments~~ See https://github.com/cocotb/cocotb/pull/4751#issuecomment-2970898982
- [x] Test manually in non-posix Windows environments

### DIFF
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -89,7 +89,7 @@ for i in range(32):
 _sv_escape_translate_table = str.maketrans(_sv_escapes)
 
 
-def _as_sv_literal(value: str) -> str:
+def _as_sv_literal(value: object) -> str:
     if isinstance(value, (int, float)):
         return str(value)
     elif isinstance(value, str):
@@ -695,7 +695,7 @@ class Icarus(Runner):
             f.write("+timescale+{}/{}\n".format(*self.timescale))
 
     def _create_iverilog_dump_file(self) -> None:
-        dumpfile_path = str(Path(self.build_dir, f"{self.hdl_toplevel}.fst"))
+        dumpfile_path = _as_sv_literal(str(self.build_dir / f"{self.hdl_toplevel}.fst"))
         with open(self.iverilog_dump_file, "w") as f:
             f.write("module cocotb_iverilog_dump();\n")
             f.write("initial begin\n")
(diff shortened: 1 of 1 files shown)


ktbarrett:
~~Looks like the conda mingw environment is not happy. I'm not surprised. That environment is jank. We really should get mingw and msys environments back up and running and drop the conda/mingw environment. It's not a really cohesive environment.~~

The build is still broken for msys environment. I think we aren't doing some things correctly in `cocotb_build_libs.py`. Also `coverage` doesn't work in the mingw64 environment. I guess making this work for conda+mingw is the best we can do for now.

src/cocotb_tools/config.py
```diff
@@ -207,32 +207,44 @@ def _get_parser() -> argparse.ArgumentParser:
         help="Print the version of cocotb",
     )
 
+    parser.add_argument(
+        "--posix-paths",
+        action="store_true",
+        help="Output paths using POSIX notation, overriding system-specific notation",
```
cmarqu:
```suggestion
        help="Use POSIX notation for paths, overriding system-specific notation",
```
(works around "Output" being read as a noun by e.g. me :))

src/cocotb_tools/runner.py
```diff
@@ -893,15 +893,13 @@ def _test_command(self) -> List[_Command]:
                 "-foreign",
                 "vhpi_startup_routines_bootstrap "
                 + _as_tcl_value(
-                    cocotb_tools.config.lib_name_path("vhpi", "questa").as_posix()
+                    str(cocotb_tools.config.lib_name_path("vhpi", "questa"))
```
themperek, line 896: @ktbarrett Why this change? What was wrong?

imphil:
@ktbarrett Is there something needed from my/our side? It looks like there's still one question from @themperek outstanding before we can merge?

ktbarrett:
Closed in favor of #4836. I have a feeling this PR will be resurrected when a Windows user finally starts passing the paths coming from `cocotb-config` into tools that don't support POSIX pathing. It works currently because we expect only Powershell and newer terminal to be interpreting paths, while this will not always be the case.

PR #4754, opened 2025-06-15
Adding Forastero to ecosystem-compat.yml

Intuity:
Adding Forastero into cocotb's ecosystem-compat.yml to check compatibility as version 2 arrives as suggested by @ktbarrett in https://github.com/Intuity/forastero/issues/35

### DIFF
--- a/.github/workflows/ecosystem-compat.yml
+++ b/.github/workflows/ecosystem-compat.yml
@@ -13,6 +13,7 @@
 name: Ecosystem compatibility tests
 
 on:
+  pull_request:
   # Run daily at midnight (UTC).
   schedule:
     - cron: '0 0 * * *'
@@ -113,7 +114,7 @@ jobs:
           make SIM=ghdl TOPLEVEL_LANG=vhdl -C pyuvm_clone/examples/TinyALU sim checkclean
 
   forastero:
-    if: github.repository == 'cocotb/cocotb'
+    # if: github.repository == 'cocotb/cocotb'
     name: Test Forastero
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ecosystem-compat.yml
```diff
@@ -111,3 +111,39 @@ jobs:
...
+          path: cocotb
+
+      - name: Checkout Forastero repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
+        with:
+          path: forastero_clone
+          repository: intuity/forastero
+
+      - name: Install the release version of Forastero
+        run: pip3 install forastero>=3.0.0
```
ktbarrett: I know the other jobs do the same thing, but the issue is will the tests on master always work with the version of forastero pulled from pypi?

Intuity: Fair point, how would you suggest addressing this? I guess the options would be:

 * Pin a specific release on PyPi
 * Pin a specific hash from Forastero's git repo
 * Point at the `main` branch of Forastero's git repo

The first two are problematic because they will require periodic updates to the pin point in cocotb, the last one is problematic because Forastero's `main` could be broken

I guess a fourth option would be to instead reference a mutable `ecosystem` tag in Forastero's repo, effectively the last known stable point for cocotb's ecosystem tests to reference? This could be bumped when the stable point is moved forwards from active development.

ktbarrett: The last would be the best, but also kind of a pain for you.

I think ideally we would install the latest version from pypi, and then use the output of something like the following to get a tag to checkout from the repo. 
```
python -c "import forastero; print(forastero.__version__)"
```

It automatically would use the newest version of the library without your intervention and only requires that there be some consistency between git tags and version strings in the project (which shouldn't be controversial).

Unfortunately, GHA doesn't allow you to use environment variables in `with` statements when we would check out the forastero repo. So we would need to separate it into two jobs and feed them together via outputs which is *very annoying*.
https://stackoverflow.com/a/73305536/6614127

Intuity: Ah I understand the problem now - the alignment is between version installed from PyPI and the git repository I'm cloning to get the examples?

Provided that is the case, I've moved the `git clone` command into the "Run tests" step allowing it to use `pip freeze` to query the installed version of Forastero, then grab the matching tag from the repository - hopefully this is agreeable?

ktbarrett:
https://github.com/cocotb/cocotb/actions/runs/15715715129/job/44284873002 shows working example.

ktbarrett (approved):

PR #4758, opened 2025-06-19
Verilator: wrap up trace and coverage on abort

gilbertoabram:
This fixes truncated traces when the simulator aborts and `clean_exit_cb()` is called instead of terminating the main loop.

### DIFF
--- a/src/cocotb/share/lib/verilator/verilator.cpp
+++ b/src/cocotb/share/lib/verilator/verilator.cpp
@@ -20,9 +20,13 @@
 #if VM_TRACE
 #if VM_TRACE_FST
 #include <verilated_fst_c.h>
+using verilated_trace_t = VerilatedFstC;
 #else
 #include <verilated_vcd_c.h>
+using verilated_trace_t = VerilatedVcdC;
 #endif
+#else
+using verilated_trace_t = void*;  // arbitrary type with no tracing
 #endif
 
 static vluint64_t main_time = 0;  // Current simulation time
@@ -50,20 +54,13 @@ static inline bool settle_value_callbacks() {
     return cbs_called;
 }
 
-static void clean_exit_cb(void* _tfp) {
+void wrap_up(void* tfp) {
     VerilatedVpi::callCbs(cbEndOfSimulation);
-#if VM_TRACE
-#if VM_TRACE_FST
-    VerilatedFstC* tfp = (VerilatedFstC*)_tfp;
-#else
-    VerilatedVcdC* tfp = (VerilatedVcdC*)_tfp;
-#endif
     if (tfp) {
-        tfp->close();
-    }
-#else
-    (void)_tfp;
+#if VM_TRACE
+        reinterpret_cast<verilated_trace_t*>(tfp)->close();
 #endif
+    }
 }
 
 int main(int argc, char** argv) {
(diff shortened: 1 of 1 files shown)


gilbertoabram:
Does it make sense to also call `VerilatedCov::write()` in `clean_exit_cb()`?

src/cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -50,7 +50,21 @@ static inline bool settle_value_callbacks() {
...
 }
 
-static void clean_exit_cb(void*) { VerilatedVpi::callCbs(cbEndOfSimulation); }
+static void clean_exit_cb(void* _tfp) {
+    VerilatedVpi::callCbs(cbEndOfSimulation);
+#if VM_TRACE
+#if VM_TRACE_FST
+    VerilatedFstC* tfp = (VerilatedFstC*)_tfp;
+#else
+    VerilatedVcdC* tfp = (VerilatedVcdC*)_tfp;
```
ktbarrett: Maybe these variables should just be global. No need to pass it around and cast it a couple times.

And/or you can split this up into multiple function defs and just register multiple exitCbs. That might be cleaner.

ktbarrett: Also this is C++11! Feel free to use lambdas instead.

gilbertoabram: I tried a few things. This last should be a bit better, but I'm not super happy.

Unfortunately making tfp a global has its own issues: the pointer must be freed manually, otherwise the sim will segfault when calling `Verilated::removeFlushCb` during static destructor calls. It's easy enough to do, especially with `wrap_up()`, but at that point it's straight up misleading to have a `unique_ptr` which we must remember to manually free, and I wasn't sure if using plain pointers was fine.

Similarly for using lambdas, if it does any capture then it cannot be passed as a function pointer, so I kept it as a plain function.

I'm not exactly super clear on the various lifetimes requirements, so to be safe I made sure here that `top` is always valid when `tfp` is, although it seems to work fine to initialize the trace object beforehand (if that is fine, we can just allocate the object and do without `traceOn`.

From reading the standard, my understanding is that the end of sim should happen after the `final` blocks, which allows for reusing `wrap_up()`. Does it make sense to also move the coverage call in there, and do it on failures as well?

src/cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -103,25 +114,21 @@ int main(int argc, char** argv) {
...
     Verilated::traceEverOn(true);
     if (traceOn) {
+        tfp = std::make_unique<verilated_trace_t>();
         top->trace(tfp.get(), 99);
         tfp->open(traceFile);
     }
 #endif
 
+    vlog_startup_routines_bootstrap();
+    Verilated::addExitCb(wrap_up, tfp.get());
```
ktbarrett: Multiple gets on a unique_ptr? Might as well just make it a raw pointer.

src/cocotb/share/lib/verilator/verilator.cpp
ktbarrett: > I wasn't sure if using plain pointers was fine.

Yeah it's fine.

> Does it make sense to also move the coverage call in there, and do it on failures as well?

Makes sense to me.

ktbarrett (approved):

PR #4759, opened 2025-06-19
LogicArray: enforce _value_as_int to be an int

gilbertoabram:
When `LogicArray` is initialized with a value of type `bool` (for which `isinstance(value, int)` is true), that will become the type of `_value_as_int`. When converting the object to an integer (via `__int__()` -> `to_unsigned()` -> `_get_int()`), it triggers https://github.com/python/cpython/blob/725da50520c5cc3002f6f55bf67c56853507c604/Objects/abstract.c#L1540 .

### DIFF
--- a/src/cocotb/types/_logic_array.py
+++ b/src/cocotb/types/_logic_array.py
@@ -227,7 +227,7 @@ class LogicArray(AbstractArray[Logic]):
             else:
                 self._range = Range(len(self._value_as_str) - 1, "downto", 0)
         elif isinstance(value, int):
-            value = int(value) # force bool to int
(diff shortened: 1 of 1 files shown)


src/cocotb/types/_logic_array.py
```diff
@@ -227,6 +227,7 @@ def __init__(
             else:
                 self._range = Range(len(self._value_as_str) - 1, "downto", 0)
         elif isinstance(value, int):
+            value = int(value) # force bool to int
```
cmarqu:
```suggestion
            value = int(value)  # force bool to int
```
Fix the ruff complaint

PR #4762, opened 2025-06-20
In Runner, make detection of GPI interface if not given better

ktbarrett:
Addresses issues in #4760.


cmarqu:
Strange failure in https://github.com/cocotb/cocotb/actions/runs/15781496276/job/44500998977?pr=4762#step:23:5131

ktbarrett:
~~@cmarqu I don't understand why only this PR has the problem. It seems like it goes back to #4734.~~

Fixed in #4765. Rebased and fixed.

themperek:
Still somethng : https://github.com/cocotb/cocotb/actions/runs/15839637301/job/45261314529?pr=4762#step:27:2471

ktbarrett:
On master the `test_mixed_language` test, when run with pytest, does the wrong thing and loads into the VPI when running a VHDL test. I'm not sure why fixing that issue causes it to fail. Running the test with the Makefile behaves properly and doesn't fail in this branch or on master.

ktbarrett:
So AFAICT the issue is with how sources are built on Questa. The Makefile builds all VHDL sources at once, then the Verilog, while the runner builds them in the specified order. I guess I can fix the compilation order on Questa? I would've expected that Questa would be less dumb about this... It's obviously making bad assumptions and stripping info from the endian_swapper_sv instance during compilation.

PR #4763, opened 2025-06-22
Docs: Tweak font size of headings; small color and padding fixes

cmarqu:


ktbarrett (approved):
This is better than what's there.

ktbarrett (approved):

PR #4765, opened 2025-06-22
Various fixes

ktbarrett:


ktbarrett:
This in fact fixes the rending of Literals. But it doesn't help resolve the references in `cocotb.test`.

cmarqu (approved):

Issue #3674, opened 2024-01-20
Add guide on how to upgrade from `BinaryValue` to `LogicArray`


ktbarrett:
This could be helpful. Alternatively, we don't do this and hope the improved documentation and backwards compatibility interface (if any) is sufficient. Update newsfrag for #3634.

ktbarrett:
Enough of the list was completed that I'm happy closing this.

PR #4766, opened 2025-06-22
[2.0 upgrade docs] `BinaryValue` -> `LogicArray`

ktbarrett:
xref #3674. Adds documentation on how to upgrade from BinaryValue to LogicArray.

### TODO

- [ ] Needs proofread

### DIFF
--- a/docs/source/upgrade-2.0.rst
+++ b/docs/source/upgrade-2.0.rst
@@ -39,8 +39,8 @@ preventing scheduler re-entrancy and sidestepping an entire class of bugs and ru
 `The cocotb blog post on this change <https://fossi-foundation.org/blog/2021-10-20-cocotb-1-6-0>`_
 is very illustrative of how :func:`!cocotb.start_soon` and :func:`!cocotb.fork` are different.
 
-Caveats
-=======
+Additional Details
+==================
 
 Coroutines run immediately
 --------------------------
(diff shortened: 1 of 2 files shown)


docs/source/upgrade-2.0.rst
```diff
@@ -166,6 +166,395 @@ Rationale
 These existed to support defining coroutines in Python 2 and early versions of Python 3 before :term:`coroutine functions <coroutine function>`
 using the :keyword:`!async`\ /:keyword:`!await` syntax was added in Python 3.5.
 We no longer support versions of Python that don't support :keyword:`!async`\ /:keyword:`!await`,
-Python coroutines are noticeably faster than :deco:`cocotb.coroutine`'s implementation,
-and the behavior of :deco:`cocotb.coroutine` would have to be changed to support changes to the scheduler.
-For all those reasons the :deco:`cocotb.coroutine` decorator and generator-based coroutine support was removed.
+Python coroutines are noticeably faster than :deco:`!cocotb.coroutine`'s implementation,
+and the behavior of :deco:`!cocotb.coroutine` would have to be changed to support changes to the scheduler.
```
cmarqu:
```suggestion
and the behavior of :deco:`!cocotb.coroutine` would have had to be changed to support changes to the scheduler.
```
?

docs/source/upgrade-2.0.rst
```diff
@@ -166,6 +166,395 @@ Rationale
...
+    assert 8 * val_int == 96
+    assert val_int << 2 == 48
+    assert val_int / 6 == 2.0
+    assert -val_int == -12
+    # inplace modification
+    val[:] = val_int * 3
+    assert val == 36
+
+
+*   Change bit indexing and slicing to use the indexing provided by the ``range`` argument to the constructor.
```
cmarqu:
```suggestion
* Change bit indexing and slicing to use the indexing provided by the ``range`` argument to the constructor.
```

docs/source/upgrade-2.0.rst
```diff
@@ -166,6 +166,395 @@ Rationale
...
+
+
+How to Upgrade
+==============
+
+* Change all constructions of :class:`!BinaryValue` to :class:`!LogicArray`.
+* Replace construction from :class:`int` with :meth:`.LogicArray.from_unsigned` or :meth:`.LogicArray.from_signed`.
+* Replace construction from :class:`bytes` with :meth:`.LogicArray.from_bytes` and pass the appropriate ``byteorder`` argument.
+
+.. code-block:: python
```
cmarqu, line 194: I guess these code blocks under list items should be indented because they are belonging to their preceding list item instead of breaking up the list like a non-indented paragraph would also do?

docs/source/upgrade-2.0.rst
```diff
@@ -166,6 +166,395 @@ Rationale
...
+
+Rationale
+=========
+
+In many cases :class:`!BinaryValue` would behave in unexpected ways that were often reported as errors.
+These unexpected behaviors were either an unfortunate product of its design or done purposefully.
+They could not necessarily be "fixed" and any fix would invariably break the API.
+So rather than attempt to fix it, it was outright replaced.
+Unfortunately, a gradual change is not possible with such core functionality,
+so it was outright replaced in one step.
```
cmarqu:
```suggestion
so it was replaced in one step.
```

docs/source/upgrade-2.0.rst
```diff
@@ -166,6 +166,395 @@ Rationale
...
+
+
+Additional Details
+==================
+
+There are some behaviors of :class:`!BinaryValue` that are *not* supported anymore.
+Support for these were deliberately not added to :class:`!LogicArray` because they were unnecessary, unintuitive, and had bugs.
```
cmarqu:
```suggestion
They were deliberately not added to :class:`!LogicArray` because they were unnecessary, unintuitive, or had bugs.
```

docs/source/upgrade-2.0.rst
```diff
@@ -166,6 +166,395 @@ Rationale
...
+Previously, when modifying a :class:`!BinaryValue` in-place using :external+cocotb19:py:meth:`BinaryValue.assign <cocotb.binary.BinaryValue.assign>`
+or the :external+cocotb19:py:attr:`BinaryValue.buff <cocotb.binary.BinaryValue.buff>`,
+:external+cocotb19:py:attr:`BinaryValue.binstr <cocotb.binary.BinaryValue.binstr>`,
+:external+cocotb19:py:attr:`BinaryValue.integer <cocotb.binary.BinaryValue.signed_integer>`,
+or :external+cocotb19:py:attr:`BinaryValue.signed_integer <cocotb.binary.BinaryValue.signed_integer>` setters,
+if the provided value was smaller than the :class:`!BinaryValue`,
+the value would be zero-extended based on the endianness of :class:`!BinaryValue`.
+
+:class:`!LogicArray` has no concept of "bit-endianness" as the indexing scheme is arbitrary.
+When partially setting a :class:`!LogicArray`, you are expected to provide the slice you want to set explicitly,
```
cmarqu:
```suggestion
When partially setting a :class:`!LogicArray`, you are expected to explicitly provide the slice you want to set,
```

docs/source/upgrade-2.0.rst
```diff
@@ -166,6 +166,395 @@ Rationale
...
+
+:class:`!BinaryValue` could be constructed with a ``binaryRepresentation`` argument of the type :external+cocotb19:py:class:`~cocotb.binary.BinaryRepresentation`
+which would select how that :class:`!BinaryValue` would interpret any integer being used to set its value.
+:external+cocotb19:py:meth:`BinaryValue.assign <cocotb.binary.BinaryValue.assign>`
+and the :external+cocotb19:py:attr:`BinaryValue.integer <cocotb.binary.BinaryValue.signed_integer>`
+and :external+cocotb19:py:attr:`BinaryValue.signed_integer <cocotb.binary.BinaryValue.signed_integer>` setters all behaved the same when given an integer.
+Unlike endianness, this could not be changed after construction (setting :attr:`!BinaryValue.binaryRepresentation` has no effect).
+
+:class:`!LogicArray` does not have a concept of integer representation as a part of its value,
+it's value is just an array of :class:`!Logic`.
```
cmarqu:
```suggestion
its value is just an array of :class:`!Logic`.
```

docs/source/upgrade-2.0.rst
ktbarrett: This is necessary because it spans over multiple lines. it doesn't render correctly otherwise. RestructuredText is a weird freaking language...

docs/source/upgrade-2.0.rst
ktbarrett: Well they belong to the previous multiple list items, which makes it a bit hard. If you have some idea how you'd format the entire thing, I'm ears. I just don't really want to have two code blocks for *every* item. That'd get fairly ugly IMO.

docs/source/upgrade-2.0.rst
```diff
@@ -166,6 +166,395 @@ Rationale
...
+There are some behaviors of :class:`!BinaryValue` that are *not* supported anymore.
+Support for these were deliberately not added to :class:`!LogicArray` because they were unnecessary, unintuitive, and had bugs.
+
+
+Dynamic-sized :class:`!BinaryValue`\ s
+--------------------------------------
+
+The above examples all pass the ``n_bits`` argument to the :class:`!BinaryValue` constructor.
+However, it is possible to construct a :class:`!BinaryValue` without a set size.
+Doing so would allow the size of the :class:`!BinaryValue` to change whenever set.
```
ktbarrett:
```suggestion
Doing so would allow the size of the :class:`!BinaryValue` to change whenever the value was set.
```

docs/source/upgrade-2.0.rst
cmarqu: Oh, true.
One way would be with subheadings, but it's not easy to come up with ones that aren't yelling "we needed some heading here".

cmarqu: Another way... use several paragraphs instead of list items. Maybe also use a horizontal line after the code examples if it's not too ugly?

ktbarrett:
@cmarqu I went with using separate paragraphs for each point and adding the horizontal line. I think the line is a bit too subtle?

cmarqu:
> I think the line is a bit too subtle?

It indeed is :) I'll update the CSS in a later PR.

cmarqu (approved):

PR #4767, opened 2025-06-23
Doc Improvements

cmarqu:


src/cocotb/_base_triggers.py
```diff
@@ -335,6 +337,8 @@ class Lock(AsyncContextManager[None]):
 
         .. code-block:: python
 
+            lock = Lock()
+            ...
             async with lock:
```
ktbarrett:
```suggestion
            async with Lock():
```

ktbarrett:
@cmarqu Anything blocking this from getting merged? I'd like to use this in #4766.

cmarqu:
I don't find enough contiguous time to take another look (unlikely to change in the next days), if you can, please cleanup/merge as you see fit.

ktbarrett (approved):

PR #4769, opened 2025-06-24
Avoid using `Any`

### LINKED ISSUE #4740: Replace uses of `Any` with `object`
`object` is preferred over `Any`. See [mypy's docs on the subject](https://mypy.readthedocs.io/en/stable/dynamic_typing.html#any-vs-object).

ktbarrett:
Depends upon #4750 and #4768. Closes #4740.

This removes as many instances of `Any` as possible, replacing them with `object` in the common case. There are still remaining uses of `Any` in parameterized types because those types (`Task`, `Outcome`, etc.) are invariant so using `object` or `Never` as the type does not work as it would with covariant or contravariant types, respectively.

Generally the rules are:

> Use `object` instead of `Any`.
> Unless this is a parameter in a parameterized type, then use `object` in output types and `Any` in input types.
> Use `Any` on parameters in parameterized types internally if the type's variance makes using `object` incorrect.
> Or if the type is being downcast from a `TypeVar`.

### DIFF
--- a/src/cocotb/_outcomes.py
+++ b/src/cocotb/_outcomes.py
@@ -6,14 +6,18 @@ or :any:`asyncio.Future`, but without being tied to a particular task model.
 """
 
 import abc
-from typing import Any, Callable, Coroutine, Generic, TypeVar
+from typing import Callable, Generic, TypeVar
 
+from cocotb._py_compat import ParamSpec
 from cocotb._utils import remove_traceback_frames
 
 T = TypeVar("T")
+P = ParamSpec("P")
 
 
-def capture(fn: Callable[..., T], *args: Any, **kwargs: Any) -> "Outcome[T]":
+def capture(
+    fn: "Callable[P, T]", *args: "P.args", **kwargs: "P.kwargs"
+) -> "Outcome[T]":
     """Obtain an `Outcome` representing the result of a function call."""
     try:
         return Value(fn(*args, **kwargs))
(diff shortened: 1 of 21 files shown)


src/cocotb/_outcomes.py
```diff
@@ -22,11 +26,7 @@ def capture(fn: Callable[..., T], *args: Any, **kwargs: Any) -> "Outcome[T]":
         return Error(e)
 
 
-class Outcome(Generic[T], metaclass=abc.ABCMeta):
-    @abc.abstractmethod
-    def send(self, coro: Coroutine[Any, T, Any]) -> Any:
-        """Send or throw this outcome into a coroutine."""
```
ktbarrett, line 28: Removed `send` since we aren't using it anymore.

src/cocotb/_py_compat.py
```diff
@@ -84,30 +84,13 @@ def __exit__(self, *excinfo: object) -> None:
...
-
-        @overload
-        def __get__(
-            self, instance: None, owner: Union[Type[Any], None] = None
-        ) -> "cached_property[T_co]": ...
-
-        @overload
-        def __get__(
-            self, instance: object, owner: Union[Type[Any], None] = None
-        ) -> T_co: ...
```
ktbarrett, line 104: Removed typing because type checkers will always follow the above branch.

src/cocotb/_test.py
```diff
@@ -61,7 +61,7 @@ def _test_done_callback(self, task: Task[None]) -> None:
             self.abort(Value(task.result()))
         elif isinstance(e, TestSuccess):
             task._log.info("Test stopped early by this task")
-            self.abort(Value(e))
+            self.abort(Value(None))
```
ktbarrett, line 64: Whoops...

src/cocotb/_py_compat.py
```diff
@@ -84,30 +84,13 @@ def __exit__(self, *excinfo: object) -> None:
...
-    from typing import Any, Callable, Generic, Type, TypeVar, Union, overload
 
-    T_co = TypeVar("T_co", covariant=True)
-
-    class cached_property(Generic[T_co]):
-        def __init__(self, method: Callable[..., T_co]) -> None:
+    class cached_property:
+        def __init__(self, method):
             self._method = method
-            update_wrapper(self, method)  # type: ignore[arg-type]
```
imphil, line 94: Just for history: the diff looks strange here, the update_wrapper() call is only modified further down, it's not removed.

imphil (approved):

PR #4772, opened 2025-06-24
Add literal type RoundMode and use it

cmarqu:

### DIFF
--- a/src/cocotb/_extended_awaitables.py
+++ b/src/cocotb/_extended_awaitables.py
@@ -25,7 +25,7 @@ from typing import (
 import cocotb.handle
 from cocotb._base_triggers import NullTrigger, Trigger, _InternalEvent
 from cocotb._gpi_triggers import FallingEdge, RisingEdge, Timer, ValueChange
-from cocotb._typing import TimeUnit
+from cocotb._typing import RoundMode, TimeUnit
 from cocotb.task import Task
 
 T = TypeVar("T")
@@ -311,7 +311,7 @@ async def with_timeout(
     trigger: TriggerT,
     timeout_time: Union[float, Decimal],
     timeout_unit: TimeUnit = "step",
-    round_mode: Optional[str] = None,
+    round_mode: Optional[RoundMode] = None,
 ) -> TriggerT: ...
 
 
@@ -320,7 +320,7 @@ async def with_timeout(
     trigger: Waitable[T],
     timeout_time: Union[float, Decimal],
     timeout_unit: TimeUnit = "step",
-    round_mode: Optional[str] = None,
+    round_mode: Optional[RoundMode] = None,
 ) -> T: ...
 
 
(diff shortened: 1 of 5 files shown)


src/cocotb/_extended_awaitables.py
```diff
@@ -337,15 +337,15 @@ async def with_timeout(
...
+    round_mode: Optional[RoundMode] = None,
 ) -> T: ...
 
 
 async def with_timeout(
     trigger: Union[Trigger, Waitable[Any], Task[Any], Coroutine[Any, Any, Any]],
     timeout_time: Union[float, Decimal],
     timeout_unit: TimeUnit = "step",
-    round_mode: Optional[str] = None,
+    round_mode: Optional[RoundMode] = None,
```
cmarqu, line 349: Couldn't we use the default "error" here directly?

ktbarrett: Yeah, why not?

ktbarrett:
I would actually prefer changing the default values to `"error"` instead of having to explain the default value in the docs.

src/cocotb/_gpi_triggers.py
```diff
@@ -142,8 +141,6 @@ def __init__(
                 stacklevel=2,
             )
             unit = units
-        if round_mode is None:
-            round_mode = type(self).round_mode
```
ktbarrett: Oh, that's why... Uhhhh... Should we keep that? Seems like it could be useful. It should definitely be kept as `"error"` in `cocotb.utils`, but anything that uses `Timer`, the global seems beneficial.

ktbarrett:
@cmarqu Instead of removing `Timer.round_mode`, I tried to document it better. Feel free to modify or suggest.

src/cocotb/_extended_awaitables.py
```diff
@@ -384,7 +384,8 @@ async def with_timeout(
             Unit of timeout_time, accepts any unit that :class:`~cocotb.triggers.Timer` does.
         round_mode:
             String specifying how to handle time values that sit between time steps
-            (one of ``'error'``, ``'round'``, ``'ceil'``, ``'floor'``).
+            (one of ``'error'``, ``'round'``, ``'ceil'``, ``'floor'``, ``None``).
+            A ``None`` argument be converted to the current value of :attr:`.Timer.round_mode`.
```
cmarqu:
```suggestion
            A ``None`` argument is converted to the current value of :attr:`.Timer.round_mode`.
```
?

src/cocotb/_gpi_triggers.py
```diff
@@ -72,7 +72,8 @@ class Timer(GPITrigger):
         round_mode:
 
             String specifying how to handle time values that sit between time steps
-            (one of ``'error'``, ``'round'``, ``'ceil'``, ``'floor'``).
+            (one of ``'error'``, ``'round'``, ``'ceil'``, ``'floor'``, ``None``).
+            A ``None`` argument be converted to the current value of :attr:`.Timer.round_mode`.
```
cmarqu:
```suggestion
            A ``None`` argument is converted to the current value of :attr:`.Timer.round_mode`.
```
?

ktbarrett: Ah yeah, I had changed the wording around a few times and screwed it up.

PR #4775, opened 2025-06-25
Break out logging handler initialization

gilbertoabram:
This makes it possible to set up a custom handler early via PYGPI_USERS,
allowing the cocotb logs to conform to the rest of the application early on
(except the initial messages from C++).

### DIFF
--- a/docs/source/newsfragments/4775.removal.rst
+++ b/docs/source/newsfragments/4775.removal.rst
@@ -1,0 +1,1 @@
+Removed ``cocotb.logging.SimBaseLog``. The class no longer did anything, so simply remove it.
--- a/docs/source/newsfragments/4775.removal.rst
+++ b/docs/source/newsfragments/4775.removal.rst
@@ -1,1 +1,1 @@
-Removed ``cocotb.logging.SimBaseLog``. The class no longer did anything, so simply remove it.
+Removed ``cocotb.logging.SimBaseLog`` as the class no longer did anything.
--- a/src/cocotb/_init.py
+++ b/src/cocotb/_init.py
@@ -19,17 +19,14 @@ import cocotb._profiling
 import cocotb.handle
 import cocotb.simulator
 from cocotb._scheduler import Scheduler
-from cocotb.logging import _log_from_c, _setup_handler, _setup_loggers
 from cocotb.regression import RegressionManager, RegressionMode
 
 log: logging.Logger
 
 
 def _setup_logging() -> None:
-    _setup_loggers()
     cocotb.log = logging.getLogger("test")
     cocotb.log.setLevel(logging.INFO)
-    cocotb.simulator.initialize_logger(_log_from_c, logging.getLogger)
 
     global log
     log = logging.getLogger("cocotb")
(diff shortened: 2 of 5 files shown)


gilbertoabram:
Another option would be to break out that code to its own method and invoke it directly with the default in `load_entry()`, and the custom applications can not give it there, but it would break things for anyone using PYGPI_USERS with the default logging, which now would need to be called explicitly.

ktbarrett:
> Another option would be to break out that code to its own method and invoke it directly with the default in load_entry(), and the custom applications can not give it there

That would be my preference. I'm not sure checking the handlers in the root object will work in every case to prevent default configuration. Just not calling the method always works.

> but it would break things for anyone using PYGPI_USERS with the default logging, which now would need to be called explicitly.

It's kind of an expert feature, so breakage isn't the end of the world. Also, 2.0 isn't out yet so we can still get some breaks in!

gilbertoabram:
I kept `default_config()` as is (assuming tests/test_cases/test_cocotb/test_logging.py represents some real usage and should keep working without changes) but split it into two private functions, making the one to set up the logging handler easy to skip via `PYGPI_USERS`

ktbarrett:
@gilbertoabram What about these changes? https://github.com/ktbarrett/cocotb/tree/logging-setup

gilbertoabram:
> @gilbertoabram What about these changes? https://github.com/ktbarrett/cocotb/tree/logging-setup

It seems good, except for potentially one thing. iiuc, by moving `logging.setLoggerClass(SimBaseLog)` to where the loggers are initialized, the cocotb loggers' type cannot be overridden. Does it make sense to keep that in `default_config`?
(That would also require to call `_setup_formatter` before `_init` during initialization).

ktbarrett:
Ah. Yeah because the loggers are created below... I had originally started with a 3 step approach:
1. Override logger class.
2. Do `default_config`.
3. Setup cocotb and gpi loggers. 

That might have to be the way if we want people to override any particular part.

ktbarrett:
@gilbertoabram I pushed some other changes to that branch.

I removed the `SimBaseLog` type and accomplished the PyGPI logger optimizations by monkeypatching the `"gpi"` logger specifically, instead of overriding the logger class. That allows us to put all cocotb-required changes in `cocotb.logging._init` and put the formatting overrides in `cocotb.logging._setup_formatting`. If the user needs to override the Logger class, they can just add an entry point before `cocotb.logging._init` and all should be good.

gilbertoabram:
Looks good!

docs/source/newsfragments/4775.removal.rst
```diff
@@ -0,0 +1 @@
+Removed ``cocotb.logging.SimBaseLog``. The class no longer did anything, so simply remove it.
```
ktbarrett:
```suggestion
Removed ``cocotb.logging.SimBaseLog`` as the class no longer did anything.
```

cmarqu:
Do the docs in https://docs.cocotb.org/en/development/rotating_logger.html still work with this PR?

ktbarrett:
@cmarqu Yes, but it could be improved after this PR, but I'll do that as a follow on.

gilbertoabram:
sorry, accidental delete

PR #4777, opened 2025-06-28
Add ability to get and set Task name

### LINKED ISSUE #4776: Add `name` to `Task`s
Python 3.8 add the ability to get and set the name of a Task, as well as extended many interfaces to support that. We could do the same.

- [ ] `Task` constructor can take a name
- [ ] `Task.get_name` and `Task.set_name`
- [ ] `cocotb.create_task`, `cocotb.start`, and `cocotb.start_soon` can take a named keyword argument.

ktbarrett:
Closes #4776.

### DIFF
--- a/docs/source/newsfragments/4777.feature.1.rst
+++ b/docs/source/newsfragments/4777.feature.1.rst
@@ -1,0 +1,1 @@
+Added the ``name`` keyword argument to :func:`cocotb.start_soon`, :func:`cocotb.start`, and :func:`cocotb.create_task` to set a :class:`.Task`\ s name during creation.
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -81,7 +81,9 @@ class Task(Generic[ResultType]):
 
     _id_count = 0  # used by the scheduler for debug
 
-    def __init__(self, inst: Coroutine[Trigger, None, ResultType]) -> None:
+    def __init__(
+        self, inst: Coroutine[Trigger, None, ResultType], *, name: Optional[str] = None
+    ) -> None:
         if inspect.iscoroutinefunction(inst):
             raise TypeError(
                 f"Coroutine function {inst} should be called prior to being scheduled."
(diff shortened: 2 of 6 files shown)


docs/source/newsfragments/4777.feature.1.rst
```diff
@@ -0,0 +1 @@
+Added the ``name`` keyword argument to :func:`cocotb.start_soon`, :func:`cocotb.start`, and :func:`cocotb.create_task` to set a :class:`.Task`\ s name when constructing one.
```
ktbarrett:
```suggestion
Added the ``name`` keyword argument to :func:`cocotb.start_soon`, :func:`cocotb.start`, and :func:`cocotb.create_task` to set a :class:`.Task`\ s name during creation.
```

src/cocotb/task.py
```diff
@@ -104,7 +106,18 @@ def __init__(self, inst: Coroutine[Trigger, None, ResultType]) -> None:
 
         self._task_id = self._id_count
         type(self)._id_count += 1
-        self._name = f"Task {self._task_id}"
+        self._name = f"Task {self._task_id}" if name is None else name
+
+    def get_name(self) -> str:
+        """Return the name.
+
+        If not set a reasonable default name is given.
```
```suggestion
        If not set using :meth:`set_name` or passed during construction,
        a reasonable default name is generated.
```

src/cocotb/task.py
```diff
@@ -104,7 +106,18 @@ def __init__(self, inst: Coroutine[Trigger, None, ResultType]) -> None:
 
         self._task_id = self._id_count
         type(self)._id_count += 1
-        self._name = f"Task {self._task_id}"
+        self._name = f"Task {self._task_id}" if name is None else name
+
+    def get_name(self) -> str:
+        """Return the name.
```
```suggestion
        """Return the name of the :class:`!Task`.
```

src/cocotb/task.py
```diff
@@ -104,7 +106,18 @@ def __init__(self, inst: Coroutine[Trigger, None, ResultType]) -> None:
...
+
+        If not set a reasonable default name is given.
+        """
+        return self._name
+
+    def set_name(self, name: str) -> None:
+        """Set the name."""
```
```suggestion
        """Set the name of the :class:`!Task`."""
```

cmarqu:
Should we also mirror the argument name of https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.set_name (their `value` vs our `name`)?
Also, the docs say
> The value argument can be any object, which is then converted to a string.

which we don't support,  maybe it makes sense to mirror that too?
Or did you choose `name` explicitly to hint at it being a string? In which case a comment stating that would be good.

ktbarrett:
> > The value argument can be any object, which is then converted to a string.
> 
> which we don't support, maybe it makes sense to mirror that too?


That is weird... but okay.

cmarqu (approved):

PR #4786, opened 2025-07-03
Force normal font for 'pre' to make docs look less busy

cmarqu:
Compare these two:
* https://docs.cocotb.org/en/development/upgrade-2.0.html#id2
* https://cocotb--4786.org.readthedocs.build/en/4786/upgrade-2.0.html#id2

### DIFF
--- a/docs/source/_static/cocotb.css
+++ b/docs/source/_static/cocotb.css
@@ -244,6 +244,17 @@ strong.command {
     color: var(--color-cocotb-blue) !important;
 }
 
+/* normal font weight for .pre ... */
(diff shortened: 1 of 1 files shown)


ktbarrett (approved):

cmarqu:
More work is needed because the font-weight forcing should not happen in headings like https://cocotb--4786.org.readthedocs.build/en/4786/upgrade-2.0.html#exceptions-before-the-first-await

cmarqu:
Now it works. Ugly, but CSS doesn't allow it better.

PR #4793, opened 2025-07-08
Small Fixes

ktbarrett:
Upstreaming some small fixes.

The first commit prevents finding virtual classes (which Xcelium puts at the top-level scope) when looking for objects with `get_root_handle`.

The second commit ensures that the cocotb directory found has the libraries built which can be an issue if cocotb_tools is run from the source directory and cocotb isn't built yet or is installed elsewhere.

### DIFF
--- a/src/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/src/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -559,7 +559,8 @@ GpiObjHdl *VpiImpl::get_root_handle(const char *name) {
     for (root = vpi_scan(iterator); root != NULL; root = vpi_scan(iterator)) {
         if (to_gpi_objtype(vpi_get(vpiType, root)) != GPI_MODULE) continue;
 
-        if (name == NULL || !strcmp(name, vpi_get_str(vpiFullName, root)))
+        const char *obj_name = vpi_get_str(vpiFullName, root);
+        if ((!name && obj_name[0] != '\\') || (name && !strcmp(name, obj_name)))
             break;
     }
 
--- a/src/cocotb/share/lib/vpi/VpiImpl.cpp
+++ b/src/cocotb/share/lib/vpi/VpiImpl.cpp
@@ -559,7 +559,8 @@ GpiObjHdl *VpiImpl::get_root_handle(const char *name) {
     for (root = vpi_scan(iterator); root != NULL; root = vpi_scan(iterator)) {
         if (to_gpi_objtype(vpi_get(vpiType, root)) != GPI_MODULE) continue;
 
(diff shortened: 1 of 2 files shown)


src/cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -559,7 +559,8 @@ GpiObjHdl *VpiImpl::get_root_handle(const char *name) {
     for (root = vpi_scan(iterator); root != NULL; root = vpi_scan(iterator)) {
         if (to_gpi_objtype(vpi_get(vpiType, root)) != GPI_MODULE) continue;
 
-        if (name == NULL || !strcmp(name, vpi_get_str(vpiFullName, root)))
+        const char *obj_name = vpi_get_str(vpiFullName, root);
+        if ((!name && obj_name[0] != '\\') || (name && !strcmp(name, obj_name)))
```
imphil, line 565: Can you add a comment why this code is necessary? Finding the root handle already feels like black magic a bit :smile:

imphil (approved):
LGTM with the added comment.

src/cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -559,7 +559,8 @@ GpiObjHdl *VpiImpl::get_root_handle(const char *name) {
     for (root = vpi_scan(iterator); root != NULL; root = vpi_scan(iterator)) {
         if (to_gpi_objtype(vpi_get(vpiType, root)) != GPI_MODULE) continue;
 
-        if (name == NULL || !strcmp(name, vpi_get_str(vpiFullName, root)))
+        const char *obj_name = vpi_get_str(vpiFullName, root);
```
ktbarrett, line 564:
```suggestion
        // prevents finding virtual classes (which Xcelium puts at the top-level scope) when looking for objects with get_root_handle.
        const char *obj_name = vpi_get_str(vpiFullName, root);
```

src/cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -559,7 +559,9 @@ GpiObjHdl *VpiImpl::get_root_handle(const char *name) {
     for (root = vpi_scan(iterator); root != NULL; root = vpi_scan(iterator)) {
         if (to_gpi_objtype(vpi_get(vpiType, root)) != GPI_MODULE) continue;
 
-        if (name == NULL || !strcmp(name, vpi_get_str(vpiFullName, root)))
+        // prevents finding virtual classes (which Xcelium puts at the top-level scope) when looking for objects with get_root_handle.
```
ktbarrett:
```suggestion
        // prevents finding virtual classes (which Xcelium puts at the top-level
        // scope) when looking for objects with get_root_handle.
```

PR #4794, opened 2025-07-08
Support customizing `Clock`'s set action

ktbarrett:
This adds support for customizing the set action when setting values in Clock. This is primarily useful for globally changing the set mode to `Immediate` which can prevent dead eval cycles since the Clock is typically what creates all time steps in a simulation and nothing else will be running at the start of the time step until the Clock signal is set.


src/cocotb/clock.py
```diff
@@ -55,21 +62,28 @@ class Clock:
...
-            ``'auto'``, ``'gpi'``, ``'py'``.
+        impl:
+            One of ``'auto'``, ``'gpi'``, ``'py'``.
             Specify whether the clock is implemented with a :class:`~cocotb.simulator.GpiClock` (faster), or with a Python coroutine.
             When ``'auto'`` is used (default), the fastest implementation that supports your environment and use case is picked.
 
             .. versionadded:: 2.0
 
+        set_action:
+            One of ``Immediate``, ``Deposit``, or ``Force``.
```
ktbarrett:
```suggestion
            One of :class:`.Immediate`, :class:`.Deposit`, or :class:`.Force`.
```

src/cocotb/clock.py
```diff
@@ -126,6 +140,14 @@ async def custom_clock():
 
     _impl: Impl
 
+    default_set_action: ClassVar[Union[Type[Immediate], Type[Deposit], Type[Force]]] = (
+        Deposit
+    )
+    """The default action used to set the clock signal value.
```
ktbarrett, line 146:
```suggestion
    """The default action used to set the clock signal value.
    
    One of :class:`.Immediate`, :class:`.Deposit`, or :class:`.Force`.
```

src/cocotb/clock.py
```diff
@@ -126,6 +140,16 @@ async def custom_clock():
...
+
```
ktbarrett:
```suggestion

```

src/cocotb/clock.py
```diff
@@ -145,6 +169,9 @@ def __init__(
             )
             unit = units
         self._unit: TimeUnit = unit
+        if set_action is None:
+            set_action = type(self).default_set_action
+        self._set_action = set_action
```
ktbarrett, line 178:
```suggestion
        if set_action not in (Immediate, Deposit, Force):
            raise TypeError("Invalid value for *set_action*. *set_action* must be one of Immediate, Deposit, or Force")
        self._set_action = set_action
```

tests/test_cases/test_cocotb/test_clock.py
```diff
@@ -165,3 +168,24 @@ async def test_clock_task_cancel(dut) -> None:
...
+async def test_set_action(dut: Any) -> None:
+    c = Clock(dut.clk, 10, "ns", set_action=Immediate)
+
+    c.start(start_high=True)
+    await NullTrigger()
+    assert dut.clk.value == 1
+
+    await Timer(5, "ns")
+    await NullTrigger()
+    assert dut.clk.value == 0
```
ktbarrett, line 191:
```suggestion
    assert dut.clk.value == 0

    assert c.set_action is Immediate
```

PR #4796, opened 2025-07-09
Correct directives in clock.py

ktbarrett:

### DIFF
--- a/src/cocotb/clock.py
+++ b/src/cocotb/clock.py
@@ -113,15 +113,11 @@ class Clock:
         high_delay = low_delay = 10  # change the clock speed
         await Timer(1000, unit="ns")
 
-    .. versionchanged:: 1.5
+    .. versionadded:: 1.5
         Support ``'step'`` as the *unit* argument to mean "simulator time step".
 
-    .. versionchanged:: 2.0
+    .. versionremoved:: 2.0
         Passing ``None`` as the *unit* argument was removed, use ``'step'`` instead.
-
-    .. versionchanged:: 2.0
-        :meth:`start` now automatically calls :func:`cocotb.start_soon` and stores the Task
-        on the Clock object, so that it may later be :meth:`stop`\ ped.
     """
 
     _impl: Impl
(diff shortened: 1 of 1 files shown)


imphil (approved):

src/cocotb/clock.py
```diff
@@ -113,15 +113,11 @@ async def custom_clock():
...
+    .. versionadded:: 1.5
         Support ``'step'`` as the *unit* argument to mean "simulator time step".
 
-    .. versionchanged:: 2.0
+    .. versionremoved:: 2.0
         Passing ``None`` as the *unit* argument was removed, use ``'step'`` instead.
-
-    .. versionchanged:: 2.0
-        :meth:`start` now automatically calls :func:`cocotb.start_soon` and stores the Task
-        on the Clock object, so that it may later be :meth:`stop`\ ped.
```
cmarqu: Did you want to remove this?

ktbarrett: It's *sort of* duplicated in `start()`... maybe not.

src/cocotb/clock.py
```diff
@@ -113,15 +113,11 @@ async def custom_clock():
         high_delay = low_delay = 10  # change the clock speed
         await Timer(1000, unit="ns")
 
-    .. versionchanged:: 1.5
+    .. versionadded:: 1.5
         Support ``'step'`` as the *unit* argument to mean "simulator time step".
 
-    .. versionchanged:: 2.0
+    .. versionremoved:: 2.0
         Passing ``None`` as the *unit* argument was removed, use ``'step'`` instead.
```
ktbarrett, line 120:
```suggestion
        Passing ``None`` as the *unit* argument was removed, use ``'step'`` instead.
 
     .. versionchanged:: 2.0
        :meth:`start` now automatically calls :func:`cocotb.start_soon` and stores the Task
        on the Clock object, so that it may later be :meth:`stop`\ ped.
```

src/cocotb/clock.py
```diff
@@ -113,13 +113,13 @@ async def custom_clock():
...
-
-    .. versionchanged:: 2.0
+ 
+     .. versionchanged:: 2.0
```
ktbarrett:
```suggestion

    .. versionchanged:: 2.0
```

PR #4799, opened 2025-07-09
Update platform support

imphil:
* Remove macOS versions that are no longer supported in GitHub actions.
~* Remove Windows 10, it's EOL.~
* Remove Ubuntu 20.04 (EOL).
* Add RHEL 10.


docs/source/platform_support.rst
```diff
@@ -66,16 +67,15 @@ This version of Python can be used with cocotb unless noted otherwise.
 Supported Windows Versions
 ==========================
 
-* **Windows 10 x86_64**
```
ktbarrett: It's supported until the end of this year, but we can still remove it if we want.

imphil: Added it back.

PR #4801, opened 2025-07-10
Remove coverage exclusions as they are now default

ktbarrett:
These rules are now default in coverage>=7.1.0. Also the `\.\.\.` rule wasn't correct and would cause entire functions to be excluded if it appeared in tuple type return annotation (whoops).

`exclude_also` was added in 7.2.0 and allows extending the default list. We include `@overload` to disable coverage on overload stubs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,6 +156,11 @@ omit = [
     "*/_vendor/*",
 ]
 
+[tool.coverage.report]
+exclude_also = [
+    "@(typing\\.)?overload",
+]
+
 [tool.codespell]
 ignore-words-list = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -155,15 +155,6 @@ omit = [
     "*/cocotb_tools/*",
     "*/_vendor/*",
 ]
```
ktbarrett, line 157:
```suggestion
]

[tool.coverage.report]
exclude_also = [
    "@(typing\\.)?overload",
]
```

cmarqu (approved):

PR #4805, opened 2025-07-14
Add FutureWarning for removing support for old Python

ktbarrett:
I think if we plan on deprecating old version of Python, rather than yanking them we should add a warning at least one minor version before removal. We've already seen a lot of limitations of continuing to support Python 3.6 and 3.7 so maybe it's time to remove support (in the next version).


docs/source/upgrade-2.0.rst
```diff
@@ -574,3 +574,18 @@ Integer representation is provided when converting to and from an integer.
...
+
+
+***********************************************
+Logic scalar and vectors simulator object split
+***********************************************
+
+
+******************************************
+Array simulator object value types changed
+******************************************
```
imphil: Was that meant to be part of this PR?

imphil (approved):
LGTM.

docs/source/upgrade-2.0.rst
ktbarrett: Oops.

PR #4807, opened 2025-07-16
add generic type parameter for child object type for HierarchyArrayObject like ArrayObject

### LINKED ISSUE #4783: HierarchyArrayObject should have generic type parameter for child object type like ArrayObject
`HierarchyArrayObject` class only has a generic type parameter for the key type (`int`), but lacks a generic type parameter to specify the type of child objects it contains, while in `ArrayObject`, has both `ElemValueT` and `ChildObjectT` type parameters.

Right now it's
```py
class HierarchyArrayObject(_HierarchyObjectBase[int], _RangeableObjectMixin):
    def __getitem__(self, key: int) -> SimHandleBase:  # Returns generic SimHandleBase
    def __iter__(self) -> Iterator[SimHandleBase]:     # Returns generic SimHandleBase
```

It should rather be like
```py
class HierarchyArrayObject(_HierarchyObjectBase[int], _RangeableObjectMixin, Generic[ChildObjectT]):
    def __getitem__(self, key: int) -> ChildObjectT:    # Returns specific child type
    def __iter__(self) -> Iterator[ChildObjectT]:      # Returns specific child type
```

CheeksTheGeek:
- adds generic type parameter for child object type for HierarchyArrayObject like ArrayObject
- adds a test for it
- closes #4783

### DIFF
--- a/src/cocotb/_py_compat.py
+++ b/src/cocotb/_py_compat.py
@@ -131,4 +131,4 @@ else:
 if sys.version_info >= (3, 11):
     from typing import Self
 else:
-    Self = ""
+    from typing_extensions import Self
--- a/src/cocotb/_py_compat.py
+++ b/src/cocotb/_py_compat.py
@@ -131,4 +131,4 @@ else:
 if sys.version_info >= (3, 11):
     from typing import Self
 else:
-    from typing_extensions import Self
+    Self = ""
--- a/tests/test_cases/test_array/test_array.py
+++ b/tests/test_cases/test_array/test_array.py
@@ -191,7 +191,7 @@ async def test_hierarchy_array_generic_typing(dut):
 
     asc_gen, desc_gen = dut.asc_gen, dut.desc_gen
 
-    asc_gen_element, desc_gen_element = asc_gen[20], desc_gen[20]
+    asc_gen_element, desc_gen_element = asc_gen[4], desc_gen[4]
 
     assert isinstance(asc_gen_element, HierarchyObject)
     assert isinstance(desc_gen_element, HierarchyObject)
--- a/tests/test_cases/test_array/test_array.py
+++ b/tests/test_cases/test_array/test_array.py
@@ -191,8 +191,7 @@ async def test_hierarchy_array_generic_typing(dut):
 
     asc_gen, desc_gen = dut.asc_gen, dut.desc_gen
 
-    asc_gen_element, desc_gen_element = asc_gen[4], desc_gen[4]
-
+    asc_gen_element, desc_gen_element = asc_gen[asc_gen.left], desc_gen[desc_gen.left]
     assert isinstance(asc_gen_element, HierarchyObject)
     assert isinstance(desc_gen_element, HierarchyObject)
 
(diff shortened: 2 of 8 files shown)


src/cocotb/_py_compat.py
```diff
@@ -131,4 +131,4 @@ def args(self) -> str:
 if sys.version_info >= (3, 11):
     from typing import Self
 else:
-    Self = ""
+    from typing_extensions import Self
```
ktbarrett: We do not want to add an unnecessary dependency. Let's keep this the way it was.

tests/test_cases/test_array/test_array.py
```diff
@@ -184,6 +185,34 @@ async def test_gen_loop(dut):
...
+async def test_hierarchy_array_generic_typing(dut):
+    """Test that HierarchyArrayObject generic typing works correctly"""
+    tlog = logging.getLogger("cocotb.test")
+
+    asc_gen: HierarchyArrayObject[HierarchyObject] = cast(
+        "HierarchyArrayObject[HierarchyObject]", dut.asc_gen
+    )
+    desc_gen: HierarchyArrayObject[HierarchyObject] = cast(
+        "HierarchyArrayObject[HierarchyObject]", dut.desc_gen
+    )
```
ktbarrett: Casts are not checked at runtime and these files are skipped by mypy because of the huge number of typing violations in them, so this isn't checking anything and can be removed.

ktbarrett (commented):
Overall good.

tests/test_cases/test_array/test_array.py
CheeksTheGeek: done!

src/cocotb/_py_compat.py
CheeksTheGeek: understood

tests/test_cases/test_array/test_array.py
```diff
@@ -184,6 +184,29 @@ async def test_gen_loop(dut):
...
 
 
+@cocotb.test()
+async def test_hierarchy_array_generic_typing(dut):
+    """Test that HierarchyArrayObject generic typing works correctly"""
+    tlog = logging.getLogger("cocotb.test")
+
+    asc_gen, desc_gen = dut.asc_gen, dut.desc_gen
+
+    asc_gen_element, desc_gen_element = asc_gen[20], desc_gen[20]
```
ktbarrett: The range is `7 downto 0`, so maybe 4 would be a better index?

ktbarrett:
Uhhh, I'm not sure why it isn't happy. You'll have to do some debugging. Ask away on Gitter!

ktbarrett:
After looking a bit more into the HDL the index 20 is appropriate for asc_gen but not for desc_gen. They don't use the same indexing. Maybe refactor to know need to "know" the range statically. There are other attributes on HierarchyArrayObject that give you information about the array's range.

CheeksTheGeek:
now it should work i think

ktbarrett:
~~Instead can we add a generate loop to the VHDl version of the file?~~ This will affect `test_discovery`.

I do agree with moving this to `test_array_simple` as it supports more simulators.

ktbarrett:
Let's just delete the test so we can get this merged. None of the changes being made affect runtime so it doesn't need a test.

CheeksTheGeek:
understood, removed it

ktbarrett (approved):

PR #4824, opened 2025-07-23
Better split cocotb-required and user-overridable logging configuration

ktbarrett:
This splits setting the log level into defaults in the overridable function and overriding them with the variables `COCOTB_LOG_LEVEL` and `GPI_LOG_LEVEL` in the non-overridable function. If neither envvar is supplied it won't affect whatever the user put in the override.

Also changes the function names. Maybe they are better?

Potential part 1 of more logging configuration changes (if I can find a way that doesn't suck).

### DIFF
--- a/src/cocotb/logging.py
+++ b/src/cocotb/logging.py
@@ -61,23 +61,31 @@ def default_config() -> None:
     An example of this can be found in the section on :ref:`rotating-logger`.
 
     .. versionadded:: 1.4
-
-    .. versionchanged:: 2.0
-        No longer set the log level of the ``cocotb`` and ``gpi`` loggers.
     """
+    logging.basicConfig()
+
     hdlr = logging.StreamHandler(sys.stdout)
     hdlr.addFilter(SimTimeContextFilter())
     if want_color_output():
         hdlr.setFormatter(SimColourLogFormatter())
     else:
         hdlr.setFormatter(SimLogFormatter())
-
-    logging.basicConfig()
     logging.getLogger().handlers = [hdlr]  # overwrite default handlers
 
+    logging.getLogger("cocotb").setLevel(logging.INFO)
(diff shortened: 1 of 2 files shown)


src/cocotb/logging.py
```diff
@@ -65,14 +68,11 @@ def default_config() -> None:
...
+
     hdlr = logging.StreamHandler(sys.stdout)
     hdlr.addFilter(SimTimeContextFilter())
-    if want_color_output():
-        hdlr.setFormatter(SimColourLogFormatter())
-    else:
-        hdlr.setFormatter(SimLogFormatter())
-
-    logging.basicConfig()
+    hdlr.setFormatter(SimLogFormatter(color=want_color_output()))  # type: ignore
```
ktbarrett: xref https://github.com/python/typeshed/pull/14445

ktbarrett:
There will probably be some more changes. Thoughts:
- [x] We need one function that allows the user to totally override the formatting and another that applies cocotb required functionality that doesn't need overriding
- [x] We need to set the cocotb or gpi logger's levels if COCOTB_LOG_LEVEL or GPI_LOG_LEVEL are specified, but also default the levels as part of the default_config. This allows users to not use those envvars and customize the log level as part of the fileConfig or formatting function override

src/cocotb/logging.py
ktbarrett: The PR was rejected, will just keep this pragma in place.

ktbarrett:
There is troubles:
* `fileConfig` doesn't support passing arbitrary arguments to `Formatters` so we can't support the sim time format
* Formatter classes specified in the conf file are found by `eval`ing the field from within the `logging` module, which means we need to inject the `SimLogFormatter` into the module before configuration.

The logging module just keeps getting weirder and worse the deeper you look. We might have to do something totally bespoke.

PR #4826, opened 2025-07-27
Improvements around runner

### LINKED ISSUE #4814: Warn when unsupported options are used in Python Runner
While reading the Python Runner code to answer #4813, I noticed that we silently ignore the `waves` argument for `build` in Xcelium. 
This could be reported as a warning and is also missing in the docstring at https://github.com/cocotb/cocotb/blob/066639ec8d800942fa3a057de896dd7ca944fdde/src/cocotb_tools/runner.py#L1444-L1445

This is also a problem with other options of other simulators.
Tests for this should be added.

cmarqu:
* Adds support for timescale for Xcelium in runner
* Raise exception when waves is set in build step for Xcelium (closes #4814)
* Improve docstrings and Simulator Support section for all simulators

### DIFF
--- a/docs/source/simulator_support.rst
+++ b/docs/source/simulator_support.rst
@@ -54,8 +54,8 @@ To redirect the wave file to a different location, use the plusarg `dumpfile_pat
 
 .. _sim-icarus-issues:
 
-Issues for this simulator
--------------------------
+Reported Issues for this Simulator
+----------------------------------
 
 * `All issues with label category:simulators:icarus <https://github.com/cocotb/cocotb/issues?q=is%3Aissue+-label%3Astatus%3Aduplicate+label%3Acategory%3Asimulators%3Aicarus>`_
 
@@ -134,8 +134,8 @@ The resulting file will be :file:`dump.fst` and can be opened by ``gtkwave dump.
 
 .. _sim-verilator-issues:
 
-Issues for this simulator
--------------------------
+Reported Issues for this Simulator
+----------------------------------
 
 * `All issues with label category:simulators:verilator <https://github.com/cocotb/cocotb/issues?q=is%3Aissue+-label%3Astatus%3Aduplicate+label%3Acategory%3Asimulators%3Averilator>`_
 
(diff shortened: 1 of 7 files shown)


docs/source/simulator_support.rst
```diff
@@ -325,6 +371,13 @@ In order to use this simulator, set :make:var:`SIM` to ``xcelium``:
 
     make SIM=xcelium
 
+For simulator-specific limitations when running with the :ref:`Python Runner <howto-python-runner>` flow,
+see :class:`cocotb_tools.runner.Xcelium`.
+
+.. note::
+
+    A working installation of the simulator itself is required.
```
ktbarrett, line 390: Maybe expand on what you mean by this? Seems kinda obvious?

src/cocotb_tools/runner.py
```diff
@@ -658,10 +658,12 @@ def is_verilog_source(source: PathLike) -> bool:
 class Icarus(Runner):
     """Implementation of :class:`Runner` for Icarus Verilog.
 
-    * ``hdl_toplevel`` argument to :meth:`.build` is *required*.
-    * ``waves=True`` *must* be given to :meth:`.build` if either ``waves`` or ``gui`` are to be used during :meth:`.test`.
-    * ``timescale`` argument to :meth:`.build` must be given to support dumping the command file.
-    * Does not support the ``pre_cmd`` argument to :meth:`.test`.
+    .. admonition:: Simulator-specific Usage
```
ktbarrett, line 661: That's a new one...

docs/source/simulator_support.rst
cmarqu: We have a similar paragraph for the open source simulators and some others. Also, I have heard in the past that some people are confused that cocotb itself is not a simulator.

src/cocotb_tools/runner.py
cmarqu: Looks better in https://cocotb--4826.org.readthedocs.build/en/4826/library_reference.html#cocotb_tools.runner.Icarus IMHO

ktbarrett (approved):

PR #4835, opened 2025-07-31
Use OIDC instead of API key for PyPi release uploads

imphil:
We have configured Trusted Publishing in PyPi, which is their term for
Open ID Connect (OIDC), where GitHub and PyPi are configured to trust
each other and share short-lived keys to perform the release upload
through the PyPi API.

Remove the hardcoded API key to use OIDC instead.


imphil:
Unfortunately it's not easy to test that approach, we'll see if the next release uploads properly.

cmarqu (approved):

PR #4843, opened 2025-08-02
Better split logging configuration

ktbarrett:
This splits setting the log level into defaults in the overridable function and overriding them with the variables COCOTB_LOG_LEVEL and GPI_LOG_LEVEL in the non-overridable function. If neither envvar is supplied it won't affect whatever the user put in the override.

This spawned a larger refactor of `_init.py` which will come in a follow on.

### DIFF
--- a/src/cocotb/logging.py
+++ b/src/cocotb/logging.py
@@ -61,9 +61,6 @@ def default_config() -> None:
     An example of this can be found in the section on :ref:`rotating-logger`.
 
     .. versionadded:: 1.4
-
-    .. versionchanged:: 2.0
-        No longer set the log level of the ``cocotb`` and ``gpi`` loggers.
     """
     hdlr = logging.StreamHandler(sys.stdout)
     hdlr.addFilter(SimTimeContextFilter())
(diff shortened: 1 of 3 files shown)


src/cocotb/logging.py
```diff
@@ -75,9 +72,17 @@ def default_config() -> None:
...
+    logging.getLogger("cocotb").setLevel(logging.INFO)
+    logging.getLogger("gpi").setLevel(logging.INFO)
+
 
-def _init(_: object) -> None:
-    """Set cocotb and pygpi log levels."""
+def _init() -> None:
+    """cocotb-specific logging setup.
+
+    Initializes the GPI logger and sets up the GPI logging optimization.
```
cmarqu, line 82: "GPI logging optimization" probably needs context.

ktbarrett: There's some in the comments a couple lines below?

ktbarrett:
So I might revive the code I had in the last PR for allowing arbitrary format strings for the prefix. It was rather slow, making the full sim about 20% slower, which means logging in total was probably 100% slower, all because of a call to `eval` on every log message. Instead, I'll eval a function definition that uses that format string and calling the function should be very low overhead.

ktbarrett:
Let's just get this bit in and I can walk the rest in.

marlonjames (approved):

PR #4830, opened 2025-07-28
Add docs and features to warn about changing indexes in 2.0

ktbarrett:
Adds a new nevvar `COCOTB_INDEXING_CHANGED_WARNING` to emit a warning whenever it detects that the user is indexing or slicing an Array or LogicArray that would have changed indexes between 1.x and 2.x.

Adds a document which discusses how to use the flag, how to approach updating, and how to fix some common usages.
These sections were moved from the "Upgrading to 2.0" doc.

### TODO
- [x] Manually created Arrays are not effected, only Arrays that come from handles.
- [x] Only LogicArrays that use and int for the `range` argument (that were upgraded from BinaryValue) and LogicArrays that come from handles are affected.
- [x] Add test
- [x] Proofread

PR #4847, opened 2025-08-02
Fallback on `set_trace` when no traceback available

### LINKED ISSUE #4837: `COCOTB_PDB_ON_EXCEPTION` fails if Ctrl-C is given when simulator is running
If we have `COCOTB_PDB_ON_EXCEPTION` set and use Ctrl-C to end the simulation I expect to be dropped into a debugger, but currently it fails. The issue is that when Ctrl-C is pressed when the simulator is running, and not Python, we enter from `_sim_event` and that ends the test with an artificial exception `SimFailure` that has no stacktrace, but this is passed to `pdb.post_mortem` and it fails because there's no stacktrace to inspect.

We need to refactor so we aren't passing an artificial exception around as a test failure.

ktbarrett:
Closes #4837.

### DIFF
--- a/src/cocotb/_test.py
+++ b/src/cocotb/_test.py
@@ -89,7 +89,10 @@ class RunningTest:
 
         # Break into pdb on test end before all Tasks are killed.
         if _pdb_on_exception and isinstance(outcome, Error):
-            pdb.post_mortem(outcome.error.__traceback__)
+            try:
+                pdb.post_mortem(outcome.error.__traceback__)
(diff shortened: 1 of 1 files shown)


cmarqu:
This is hard to review because it includes the commits of #4830.

ktbarrett:
Yeah, working solely on the work laptop and my environment isn't set up quite as nicely here.

cmarqu (approved):

PR #4856, opened 2025-08-06
Misc fixes

marlonjames:
I'm a bit surprised we didn't have tests trying to restart finished Tasks.


src/cocotb/task.py
```diff
@@ -135,7 +135,13 @@ def _cancelled_error(self) -> CancelledError:
 
     @cached_property
     def _log(self) -> logging.Logger:
-        return logging.getLogger(f"cocotb.{self._name}.{self._coro.__qualname__}")
+        # abc.Coroutine subclass objects don't have __qualname__ attribute
+        coro_name: str
+        if hasattr(self._coro, "__qualname__"):
+            coro_name = self._coro.__qualname__
+        else:
+            coro_name = type(self._coro).__qualname__
```
ktbarrett: This is fine, but maybe we need to add a flag `is_native_coroutine` and use that as the condition for here and in the "get stack" function. It would be helpful when writing the debugging.

tests/test_cases/test_cocotb/test_scheduler.py
```diff
@@ -1051,3 +1054,46 @@ async def coro() -> None:
...
+
+@cocotb.test(expect_error=RuntimeError)
+async def test_start_again_finished_task(_: object) -> None:
+    async def coro():
+        await Timer(1, "ns")
+
+    a = cocotb.start_soon(coro())
+    await a
+
+    cocotb.start_soon(a)
```
ktbarrett: Can we make this more specific?
```suggestion
@cocotb.test
async def test_start_again_finished_task(_: object) -> None:
    async def coro() -> None:
        await Timer(1, "ns")

    a = cocotb.start_soon(coro())
    await a

    with pytest.raises(RuntimeError):
        cocotb.start_soon(a)
```

src/cocotb/task.py
marlonjames: This is similar to
https://github.com/cocotb/cocotb/blob/239cad406dc73aba6f1c1a8223eb8a8daf858a1d/src/cocotb/task.py#L171-L174

I think I'll defer this to later and make an issue to clean things up.

PR #4857, opened 2025-08-06
docs: Clarify test outcome with CancelledError

marlonjames:
Update docs to clarify test pass/fail/error outcome when cancellation is involved.

### DIFF
--- a/docs/source/writing_testbenches.rst
+++ b/docs/source/writing_testbenches.rst
@@ -329,69 +329,15 @@ It may appear as one or more attributes here depending on the number of compilat
 Passing and failing tests
 =========================
 
-A cocotb test is considered to have `failed` if the test coroutine or any running :class:`~cocotb.task.Task`
-fails an :keyword:`assert` statement.
-Below are examples of `failing` tests.
+When cocotb tests complete execution, they have either `passed` or `failed`.
 
-.. code-block:: python
-
-    @cocotb.test()
-    async def test(dut):
-        assert 1 > 2, "Testing the obvious"
-
-    @cocotb.test()
-    async def test(dut):
-        async def fails_test():
-            assert 1 > 2
-        cocotb.start_soon(fails_test())
-        await Timer(10, 'ns')
-
-When a test fails, a stacktrace is printed.
-If :mod:`pytest` is installed and assert statements are used,
-a more informative stacktrace is printed which includes the values that caused the assert to fail.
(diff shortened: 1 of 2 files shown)


docs/source/writing_testbenches.rst
```diff
@@ -361,7 +361,10 @@ For example, see the output for the first test from above.
 
 
 A cocotb test is considered to have `errored` if the test coroutine or any running :class:`~cocotb.task.Task`
-raises an exception that isn't considered a `failure`.
+raises an exception that isn't considered a `failure`, except for a :exc:`!CancelledError`.
```
cmarqu:
```suggestion
raises an exception that isn't considered a `failure`, except for a :exc:`CancelledError`.
```
Worth to link here (the first time in this section) IMO.

ktbarrett: We should be explicit with what exceptions are not considered a test failure (only CancelledError rn, TestSuccess is an implementation detail / deprecated)
```suggestion
raises an exception, except for a :exc:`!CancelledError`.
```

marlonjames: Well, the start of this section defines failure as failing an `assert` statement.
So any *other* exception that's not `CancelledError` is test error.

ktbarrett: I'm not sure the distinction between failure and error is meaningful. If we start to adapt more pytest-isms, it won't map. They don't separate the two.

marlonjames:
Should I be extra pedantic about Tasks being able to raise exceptions that can be caught and handled that wouldn't cause a test to error?

docs/source/writing_testbenches.rst
```diff
@@ -418,12 +421,25 @@ A passing test will print the following output.
...
 
+In summary:
+
++--------------+--------------------------------------------------+
+| Test Passed  | | Test ends without raising Exception.           |
+|              | | Test or Task calls :func:`cocotb.pass_test`.   |
+|              | | Test raises :exc:`~asyncio.CancelledError`.    |
++--------------+--------------------------------------------------+
+| Test Failed  | | Test fails an :keyword:`assert` statement.     |
+|              | | Test or Task raises :exc:`AssertionError`.     |
```
ktbarrett: Technically failing a `pytest.raises` or `pytest.warns` also is a failure.

marlonjames:
Reworked it again.

PR #4860, opened 2025-08-06
Improve handle `__setattr__` error message

### LINKED ISSUE #4858: Missing ".value" gives a misleading error message
If I attempt to do this in my Python:
```    dut.dut_0.addr_0 = 1```

I get an error of:
```     3.50ns WARNING  ..Test my_first_test.my_first_test dut_0 contains no object named addr_0```

Unfortunately, that error is quite misleading.

The problem is that I meant to do:
```    dut.dut_0.addr_0.value = 1```

which works fine.

Thanks.

ktbarrett:
Closes #4858.

### DIFF
--- a/src/cocotb/handle.py
+++ b/src/cocotb/handle.py
@@ -464,7 +464,16 @@ class HierarchyObject(_HierarchyObjectBase[str]):
         if name.startswith("_"):
             return object.__setattr__(self, name, value)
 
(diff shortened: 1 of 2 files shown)


marlonjames:
Add a test to get coverage of this?

cmarqu (commented):
If setting `.value` would work, it would be nice to suggest it in the error message.

PR #4862, opened 2025-08-07
Cleanup pragmas

ktbarrett:
Various cleanup around SV pragmas.
* Verilator supports unpacked structs in all versions we support
* Remove unnecessary dumpfile/dumpvars in SV sources

### DIFF
--- a/tests/designs/sample_module/sample_module.sv
+++ b/tests/designs/sample_module/sample_module.sv
@@ -6,7 +6,7 @@
 
 `ifndef NOTIMESCALE
 `timescale 1 ps / 1 ps
-`endif
+`endif  // `ifndef NOTIMESCALE
 
 `ifndef __ICARUS__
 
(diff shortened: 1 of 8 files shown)


tests/designs/sample_module/sample_module.sv
```diff
@@ -81,7 +76,7 @@ localparam string STRING_LOCALPARAM = "TESTING_LOCALPARAM";
 
 var   string STRING_VAR   = "TESTING_VAR";
 const string STRING_CONST = "TESTING_CONST";
-`endif
+`endif  // __ICARUS__
```
marlonjames: etc, to match the existing ones?
```suggestion
`endif  // `ifndef __ICARUS__
```

marlonjames (approved):

tests/designs/sample_module/sample_module.sv
ktbarrett: Sure.

PR #4865, opened 2025-08-07
Make 1st level Python objects stand out more against 2nd level objects in docs

cmarqu:

### DIFF
--- a/docs/source/_static/cocotb.css
+++ b/docs/source/_static/cocotb.css
@@ -555,3 +555,10 @@ body.release-notes h3 {
   top: calc(3 * 48px);
   z-index: 970;
 }
+
+
+code.literal {
+  border: none;
+  background-color: unset;
+  padding: 0;
+}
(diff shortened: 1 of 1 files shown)


marlonjames:
It's better than before.

What about something like this? Maybe it is too incongruent with the rest of the new theme.
```css
:root {
    --color-cocotb-blue-bg: rgb(245, 250, 255);
}

dl.py.class > dt.sig-object.py,
dl.py.function > dt.sig-object.py,
dl.py.data > dt.sig-object.py {
    border-top: 4px solid var(--color-cocotb-blue);
    background-color: var(--color-cocotb-blue-bg);
    padding-left: 4px;
    padding-right: 4px;
}
```
image

cmarqu:
@marlonjames I'll think about it some more over the weekend.

cmarqu:
@marlonjames What do you think of the current look, with a thin blue line between such 1st level objects?

marlonjames:
This is definitely better than nothing or just adding white-space.

marlonjames (approved):

Issue #4809, opened 2025-07-16
Debugging Improvements


ktbarrett:
Add a new module that will include cocotb debugging functionality that will be useful for debugging in `pdb` or similar.

Possible features:
* Access all living Tasks
* Print information about all those Tasks
   * Stack
   * State
   * Which trigger are they waiting on
   * Which Tasks are waiting on them
* Turn on/off internals debugging

Also `COCOTB_SCHEDULER_DEBUG` is kind of crappy right now since it randomly logs some subset of all internal debugging and is repeated in several files. Also a lot of the debugging messages aren't terribly helpful while being rather expensive to compute, but I think #4717 solves some of those issues.

cmarqu:
Yay, https://www.youtube.com/watch?v=RrsVi1P6n0w for cocotb! :)

ktbarrett:
~~If only... I remember listening to that and looking it up to be dismayed they built it specifically for asyncio. We will have to push to generalize it upstream to see the features become available for cocotb (and trio).~~

Nvm, confused it with the `pdb` changes to support asyncio Task introspection. That will require upstream changes to get it to support cocotb (and trio). https://github.com/python/cpython/issues/121468

PR #4867, opened 2025-08-10
Centralize and document debugging flag

### LINKED ISSUE #3968: Test scheduler debug statements
There is [sometimes non-trivial code](https://github.com/cocotb/cocotb/blob/2bfb730851c18883407f4a3397864fb148cc0ca9/src/cocotb/_scheduler.py#L422-L428) behind `COCOTB_SCHEDULER_DEBUG` guards that is currently not being tested or covered.

ktbarrett:
Replaces first commit of #4717. xref #4809.

Also tests the debugging logs without actually printing them. Should improve coverage a bit. Closes #3968.

### DIFF
--- a/src/cocotb/_bridge.py
+++ b/src/cocotb/_bridge.py
@@ -3,7 +3,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import functools
 import logging
-import os
 import threading
 from enum import IntEnum
 from typing import (
@@ -16,16 +15,13 @@ from typing import (
 )
 
 import cocotb
+from cocotb import debug
 from cocotb._base_triggers import Event, Trigger
 from cocotb._exceptions import InternalError
 from cocotb._py_compat import ParamSpec
 
 if TYPE_CHECKING:
     from cocotb._outcomes import Outcome
-
-# Sadly the Python standard logging module is very slow so it's better not to
-# make any calls by testing a boolean flag first
-_debug = "COCOTB_SCHEDULER_DEBUG" in os.environ
 
 P = ParamSpec("P")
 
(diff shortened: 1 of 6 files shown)


src/cocotb/_bridge.py
```diff
@@ -136,15 +132,15 @@ def result(self) -> Result:
 
     def _propagate_state(self, new_state: external_state) -> None:
         with self.cond:
-            if _debug:
+            if debug.DEBUG:
```
marlonjames, line 135: Is there any performance hit to using `debug.DEBUG` vs using `from cocotb.debug import DEBUG`?

ktbarrett: There is a functionality change. The user is allowed to change the flag at runtime. Doing the later breaks that.

ktbarrett: The idea is that the user can run the test until a particular point, turn on debugging and get more information, then turn it back off if desired.

src/cocotb/debug.py
```diff
@@ -0,0 +1,24 @@
...
+
+import os
+
+DEBUG: bool = bool(os.getenv("COCOTB_SCHEDULER_DEBUG"))
+"""Global flag to enable additional debugging functionality.
+
+Defaults to ``True`` if the :envvar:`COCOTB_SCHEDULER_DEBUG` environment variable is set,
+but can be programmatically set by the user afterwards.
+
+The ``"cocotb"`` logger should have it's logging level set to :data:`logging.DEBUG`
```
cmarqu:
```suggestion
The ``"cocotb"`` logger should have its logging level set to :data:`logging.DEBUG`
```

ktbarrett:
The project coverage has increased, but the patch coverage is low because we aren't hitting branch coverage on `not debug.DEBUG` in the debug blocks. I don't think this is valuable considering there's no code in an else branch, but whatever.

Issue #4808, opened 2025-07-16
Logging Improvements


ktbarrett:
To make logging more customizable, I've come up with the following:

* Support configuring loggers using `logging.config.fileConfig` by adding a new environment variable `COCOTB_LOG_CONFIG` which is a path to [a file config](https://docs.python.org/3/library/logging.config.html#configuration-file-format).
* Expand the `SimLogFormatter` and `SimColorLogFormatter` to support `Formatter`'s [format string style configuration](https://docs.python.org/3/library/logging.html#formatter-objects).
   * This will probably only support one "style" and be the familiar f-string style and automatically include the LogRecord's attributes as variable names.
   * Add support for customizing the format of simulated time strings.

This should allow the user to use `COCOTB_LOG_CONFIG` to point to a config file that sets up their loggers exactly as they see fit, and format them as they see fit using the new format string feature.

### Questions
* The config file can set log levels on loggers like `"cocotb"` and `"gpi"`. Should we disable that part of the config if a file is given?
   * Just do our configuration before, but let the user override it?
* How should we split up the logging initialization to allow people to reuse as much as possible?

### Possible other improvements
* Re-implement existing formats using new format strings (should be way faster than manually building strings like currently)
* Instead of using `SimTimeContextFilter` to add simulation time to only those handlers that remember to add it, instead we can use `logging.setLogRecordFactory` as described in [this section of the docs](https://docs.python.org/3/howto/logging-cookbook.html#customizing-logrecord).
   * This will add the sim time to LogRecords globally instead of only those Handlers you remember to do so.

PR #4871, opened 2025-08-11
`cocotb.logging` refactor

ktbarrett:
xref #4808. Will be the basis for in-progress addition of `COCOTB_LOG_PREFIX` which allows users to customize the logging prefix.

This is a general refactor for performance/flexibility:
* Adds ability to pass `color` and `reduced_log_fmt` to callers of `default_config` so that function isn't dependent upon private globals or environment variables
* Refactors `SimLogFormatter` and `SimColourLogFormatter` into one class, `SimColourLogFormatter` is deprecated.
* Refactors the implementation of `formatMessage` in cocotb's formatter to use f-strings and other performance-minded refactors

### DIFF
--- a/src/cocotb/logging.py
+++ b/src/cocotb/logging.py
@@ -11,65 +11,62 @@ Everything related to logging
 import logging
 import os
 import sys
+import warnings
 from functools import wraps
 from pathlib import Path
-from typing import Union
+from typing import Optional, Tuple, Union
 
 from cocotb import _ANSI, simulator
 from cocotb._deprecation import deprecated
 from cocotb._utils import want_color_output
 from cocotb.utils import get_sim_time, get_time_from_sim_steps
 
-try:
-    _suppress = int(os.environ.get("COCOTB_REDUCED_LOG_FMT", "1"))
-except ValueError:
-    _suppress = 1
-
-# Column alignment
-_LEVEL_CHARS = len("CRITICAL")
-_RECORD_CHARS = 34
-_FILENAME_CHARS = 20
-_LINENO_CHARS = 4
-_FUNCNAME_CHARS = 31
-
-# Custom log level
-logging.TRACE = 5  # type: ignore[attr-defined]  # type checkers don't like adding module attributes after the fact
-logging.addLevelName(5, "TRACE")
-
-
 __all__ = (
-    "SimColourLogFormatter",
     "SimLog",
     "SimLogFormatter",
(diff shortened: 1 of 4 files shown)


src/cocotb/logging.py
```diff
@@ -203,88 +213,81 @@ def rjust(string: str, chars: int) -> str:
...
-                + " in "
-                + self.ljust(str(record.funcName), _FUNCNAME_CHARS)
-                + " "
-            )
+        highlight_end = _ANSI.COLOR_DEFAULT if self.color else ""
+
+        prefix = f"{sim_time_str:>11} {highlight_start}{record.levelname:<8}{highlight_end} {self.ljust(record.name, 34)} "
+
+        if not self._reduced_log_fmt:
+            prefix = f"{prefix}{self.rjust(Path(record.filename).name, 20)}:{record.lineno:<4} in {self.ljust(str(record.funcName), 31)} "
```
ktbarrett, line 231: There is a slight change in behavior here for formatting lineno, but I think it's beneficial.

Instead using `self.ljust` which truncates the value and adds dots if it exceeds the limit, we just use builtin left justification that won't truncate. If somehow your test file is >10000 lines long the logs will look wonky, but preserving info is better IMO and I doubt that 10k+ test files exist.

ktbarrett: We could also make it 5 characters wide and assume ***very*** safely there aren't 100k+ line test files.

src/cocotb/logging.py
```diff
@@ -203,88 +213,81 @@ def rjust(string: str, chars: int) -> str:
...
                 record.exc_text = self.formatException(record.exc_info)
         if record.exc_text:
-            if msg[-1:] != "\n":
-                msg = msg + "\n"
-            msg = msg + record.exc_text
+            msg += record.exc_text
+        if record.stack_info:
+            if not msg.endswith("\n"):
+                msg += "\n"
+            msg += self.formatStack(record.stack_info)
```
ktbarrett, line 262: This bit is from stdlib, but was missing here.

cmarqu:
https://cocotb--4871.org.readthedocs.build/en/4871/library_reference.html#logging.LogRecord.created_sim_time has the "v1.9" label at the end of the paragraph "This is added by cocotb.log.SimTimeContextFilter", could be fixed by this PR?

cmarqu (approved):

PR #4872, opened 2025-08-11
Add `period_high` feature to Clock driver

### LINKED ISSUE #4868: time precision same as clock causes failure
If I set my time precision with either:

```timeunit 1ns/1ns;```
or
````timescale 1ns/1ns```

And subsequently declare my clock:
```    cocotb.start_soon(Clock(dut.clk, 1, unit="ns").start())```

I get an error of:
```     0.00ns WARNING  ..register_file.test_register_file Failed to start clock: invalid arguments!```

However, if i change the Clock unit to "ms", things work.  If I change the precision to 1ps, things also work.

If I specify:
````timescale 1ms/1ms```

I (properly) get: ``` 0.00ns WARNING  ..register_file.test_register_file Unable to accurately represent 1(ns) with the simulator precision of 1e-3```

So, something appears to be broken if I specify my clock with the exact unit of my timescale precision.

I'm using:
```     0.00ns INFO     cocotb                             Running on Verilator version 5.039 devel
     0.00ns INFO     cocotb                             Seeding Python random module with 1754875810
     0.00ns INFO     cocotb                             Initialized cocotb v2.0.0.dev1+239cad40 from /var/home/foo/work/.venv/lib/python3.10/site-packages/cocotb```

Thanks.

### PR DESCRIPTION
**Pull request by ktbarrett**
Allows the Clock driver to run at something other than 50% duty cycle. I figured the C++ version already supported this, so why not?

Also cleans up error messages when passing bad values and checks them sooner (in `__init__` rather than once `start`ed). Closes #4868.

### DIFF
--- a/src/cocotb/clock.py
+++ b/src/cocotb/clock.py
@@ -58,10 +58,10 @@ class Clock:
 
     Args:
         signal: The clock pin/signal to be driven.
-        period: The clock period.
+        period:
+            The clock period.
+            Must be a multiple of the time precision of the simulator.
 
-            .. note::
-                Must convert to an even number of timesteps.
         unit:
             One of ``'step'``, ``'fs'``, ``'ps'``, ``'ns'``, ``'us'``, ``'ms'``, ``'sec'``.
             When *unit* is ``'step'``,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by cmarqu** [comment]
https://cocotb--4872.org.readthedocs.build/en/4872/library_reference.html#cocotb.clock.Clock.start says "first half of the period" - reword to "first part of the period" maybe.

File: src/cocotb/clock.py [hunk 1]
```diff
@@ -196,9 +228,22 @@ def signal(self) -> LogicObject:
...
-        """The clock period (unit-less)."""
+        """The clock period.
+
+        The unit is :attr:`unit`.
+        """
         return self._period
 
+    @property
+    def period_high(self) -> Union[float, Fraction, Decimal]:
+        """The period where the clock is ``1``.
```
cmarqu:
```suggestion
        """The period of time when the clock is driven to ``1``.
```
Reuse string from above.

PR #4875, opened 2025-08-12
Optimize `Clock.cycles`

ktbarrett:
Resolves the TODO in the code. Also better check corner cases where num_cycles <= 0.

### DIFF
--- a/src/cocotb/clock.py
+++ b/src/cocotb/clock.py
@@ -31,9 +31,9 @@ from cocotb.handle import (
 from cocotb.simulator import clock_create
 from cocotb.task import Task
 from cocotb.triggers import (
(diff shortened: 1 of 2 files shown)


cmarqu (approved):
*Might* be considered a user-visible change because of better performance and thus worthy of a newsfragment.

ktbarrett:
This function was introduced in 2.0. Are we planning on having separate changelog for 2.0 beta and 2.0?

cmarqu:
> This function was introduced in 2.0. Are we planning on having separate changelog for 2.0 beta and 2.0?

Ah, no, certainly not.
(There was a similar function pre-2.0, I was confusing it with that.)

PR #4877, opened 2025-08-12
Update CI

ktbarrett:
Bumps several simulator's versions in CI.

### DIFF
--- a/.github/generate-envs.py
+++ b/.github/generate-envs.py
@@ -39,7 +39,7 @@ ENVS = [
     {
         "lang": "vhdl",
         "sim": "nvc",
-        "sim-version": "r1.16.0",
+        "sim-version": "r1.17.1",
         "os": "ubuntu-22.04",
         "python-version": "3.6",
         "group": "ci-free",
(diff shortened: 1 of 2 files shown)


.github/generate-envs.py
```diff
@@ -191,7 +191,7 @@
     {
         "lang": "verilog",
         "sim": "verilator",
-        "sim-version": "v5.036",  # Latest release version.
+        "sim-version": "v5.038",  # Latest release version.
```
imphil, line 242: I guess this is the first time in a long time that we have more than one supported Verilator version? We need the older versions in extended CI then, just like we do for Questa & friends.

ktbarrett: ~~This already exists about 10 lines below.~~ But I guess this applies to NVC.

ktbarrett:
I'm invesitgating whether NVC r1.11.0 is still supported.

ktbarrett:
The additional jobs work. https://github.com/cocotb/cocotb/actions/runs/16938073040

Issue #4866, opened 2025-08-08
I use the VCS (Version V-2023.12-SP2_Full64) but get ERROR like this, does new VCS verison unsupported?


JiangfengHong:
<pre>(cocotb) [ic@localhost /mnt/hgfs/share/xxx/AXI_Master/sim_1]$make
rm -f results.xml
&quot;make&quot; -f Makefile results.xml
make[1]: 进入目录“/mnt/hgfs/share/xxx/AXI_Master/sim_1”
mkdir -p sim_build
echo &quot;acc+=rw,wn:*&quot; &gt; sim_build/pli.tab
cd sim_build &amp;&amp; \
TOPLEVEL=axi_ram \
/home/jjt/install/synopsys/vcs/vcs/V-2023.12-SP2/bin/vcs -top axi_ram  -debug_access+r+w-memcbk -debug_region+cell +vpi -P pli.tab +define+COCOTB_SIM=1 -sverilog \
-timescale=1us/1us \
-full64 -debug -load /home/jjt/miniconda3/envs/cocotb/lib/python3.9/site-packages/cocotb/libs/libcocotbvpi_vcs.so  /mnt/hgfs/share/xxx/AXI_Master/sim_1/../rtl/axi_ram.v

Warning-[DEBUG_DEP] Option will be deprecated
  The option &apos;-debug&apos; will be deprecated in a future release.  Please use 
  &apos;-debug_acc+pp+f+dmptf -debug_region+cell+encrypt&apos; instead.

Warning-[DBGACC_DBG] Multiple debug options being used
  The debug switches &apos;-debug_access&apos; and &apos;-debug*&apos; are being used together.  
  For better performance, consider using only &apos;-debug_access&apos;.

                         Chronologic VCS (TM)
       Version V-2023.12-SP2_Full64 -- Sat Aug  9 00:16:42 2025

                    Copyright (c) 1991 - 2024 Synopsys, Inc.
   This software and the associated documentation are proprietary to Synopsys,
 Inc. This software may only be used in accordance with the terms and conditions
 of a written license agreement with Synopsys, Inc. All other use, reproduction,
   or distribution of this software is strictly prohibited.  Licensed Products
     communicate with Synopsys servers for the purpose of providing software
    updates, detecting software piracy and verifying that customers are using
    Licensed Products in conformity with the applicable License Key for such
  Licensed Products. Synopsys will use information gathered in connection with
    this process to deliver software updates and pursue software pirates and
                                   infringers.

[...]

PR #4879, opened 2025-08-12
Fix VCS Makefile GUI option

ktbarrett:
xref https://github.com/cocotb/cocotb/issues/4866#issuecomment-3170558738.


imphil (approved):
LGTM. 

We don't support the GUI mode in the runner yet, so no need to update anything there.

PR #4884, opened 2025-08-13
Support copy on modeling types

### LINKED ISSUE #4883: Logic type cannot be copied in 2.0.0b1
```python
import copy
from cocotb.types import Logic

x = Logic(1)
copy.deepcopy(x) # Fails
copy.copy(x) # Fails
```

Running this in 2.0.0b1 results in this traceback.

```
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib/python3.10/copy.py", line 265, in _reconstruct
    y = func(*args)
  File "/usr/lib/python3.10/copyreg.py", line 101, in __newobj__
    return cls.__new__(cls, *args)
TypeError: Logic.__new__() missing 1 required positional argument: 'value'
```

Running this in v1.9.2 works as expected.

This causes issues for bus monitors which want to copy bus values of type Logic and LogicArray.

Many thanks,
Jon

ktbarrett:
Closes #4883.

### DIFF
--- a/docs/source/newsfragments/4884.feature.rst
+++ b/docs/source/newsfragments/4884.feature.rst
@@ -1,0 +1,1 @@
+Add support for :func:`copy.copy` and :func:`copy.deepcopy` on :class:`.Range`, :class:`.Array`, :class:`.Logic`.
--- a/src/cocotb/types/_array.py
+++ b/src/cocotb/types/_array.py
@@ -1,8 +1,9 @@
 # Copyright cocotb contributors
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
+import copy
 import warnings
-from typing import Iterable, Iterator, List, TypeVar, Union, cast, overload
+from typing import Any, Dict, Iterable, Iterator, List, TypeVar, Union, cast, overload
 
 from cocotb.types._abstract_array import AbstractMutableArray
 from cocotb.types._indexing import IndexingChangedWarning
@@ -284,3 +285,11 @@ class Array(AbstractMutableArray[T]):
             return self._range.index(item)
         except ValueError:
             raise IndexError(f"index {item} out of range") from None
+
+    def __copy__(self) -> "Array":
+        return Array(self._value, self._range)
+
+    def __deepcopy__(self, memo: Dict[int, Any]) -> "Array":
+        return Array(
+            copy.deepcopy(self._value, memo=memo), copy.deepcopy(self._range, memo=memo)
+        )
--- a/src/cocotb/types/_array.py
+++ b/src/cocotb/types/_array.py
@@ -133,6 +133,8 @@ class Array(AbstractMutableArray[T]):
         TypeError: When invalid argument types are used.
     """
 
+    __slots__ = ("_value", "_range", "_warn_indexing")
+
     def __init__(
         self, value: Iterable[T], range: Union[Range, int, None] = None
     ) -> None:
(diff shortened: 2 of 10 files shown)


src/cocotb/types/_range.py
```diff
@@ -165,6 +166,12 @@ def __repr__(self) -> str:
 
     index = cached_method(Sequence.index)
 
+    def __copy__(self) -> "Range":
+        return Range.from_range(self._range)
+
+    def __deepcopy__(self, memo: Dict[int, Any]) -> "Range":
+        return Range.from_range(copy.deepcopy(self._range, memo=memo))
```
ktbarrett, line 173: `range` object is immutable.
```suggestion
        return Range.from_range(self._range)
```

src/cocotb/types/_logic_array.py
```diff
@@ -849,3 +860,9 @@ def resolve(self, resolver: ResolverLiteral) -> "LogicArray":
             TypeError: Unsupported *value* type.
         """
         return LogicArray(get_str_resolver(resolver)(str(self)), self.range)
+
+    def __copy__(self) -> "LogicArray":
+        return LogicArray(self._get_str(), self._range)
+
+    def __deepcopy__(self, memo: Dict[int, Any]) -> "LogicArray":
+        return LogicArray(self._get_str(), copy.deepcopy(self._range, memo=memo))
```
ktbarrett, line 868:
```suggestion
        return LogicArray(self._get_str(), self._range)
```

src/cocotb/types/_array.py
```diff
@@ -282,3 +285,11 @@ def _translate_index(self, item: int) -> int:
             return self._range.index(item)
         except ValueError:
             raise IndexError(f"index {item} out of range") from None
+
+    def __copy__(self) -> "Array":
+        return Array(self._value, self._range)
+
+    def __deepcopy__(self, memo: Dict[int, Any]) -> "Array":
+        return Array(
+            copy.deepcopy(self._value, memo=memo), copy.deepcopy(self._range, memo=memo)
```
ktbarrett, line 294:
```suggestion
            copy.deepcopy(self._value, memo=memo), self._range)
```

src/cocotb/types/_logic_array.py
```diff
@@ -849,3 +860,9 @@ def resolve(self, resolver: ResolverLiteral) -> "LogicArray":
             TypeError: Unsupported *value* type.
         """
         return LogicArray(get_str_resolver(resolver)(str(self)), self.range)
+
+    def __copy__(self) -> "LogicArray":
+        return LogicArray(self._get_str(), self._range)
```
ktbarrett: This actually violates the definition of `copy.copy`. We would have to add some flags to force only `_value_as_array` to be maintained when the value is copied. That's ugly and awful, so I'm tempted to remove support for this operation.

src/cocotb/types/_range.py
ktbarrett: I left these in, cus why not? The implementation of these types might change.

src/cocotb/types/_logic_array.py
ktbarrett: I did remove it.

docs/source/newsfragments/4884.feature.rst
```diff
@@ -0,0 +1 @@
+Add support for :func:`copy.copy` and :func:`copy.deepcopy` for :class:`.Range`, :class:`.Array`, :class:`.Logic`, and :class:`.LogicArray`.
```
marlonjames: Maybe add note that shallow copy isn't supported on `LogicArray`.

PR #4885, opened 2025-08-13
Specify minimum version of supported Icarus is 11.0

ktbarrett:


marlonjames:
NVC minimum version is a note. GHDL and DSim are warnings due to "experimental" support. Verilator minimum version is a warning but doesn't say "experimental".

This is a good PR to make sure these are all as we want them for 2.0, so I think Icarus and Verilator minimum version should both be a note to match NVC.

imphil:
Does not necessarily need to be in this PR, but GHDL is probably safe to make "non-experimental". It's not significantly worse than others. Dsim on the other hand is very experimental :-)

marlonjames (approved):

PR #4888, opened 2025-08-15
Add `Bit` back

ktbarrett:
Re-adds `Bit` class. I plan on adding more types in the near future including `BitArray` and subtypes which can use this type.

### TODO
- [x] tests

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -136,6 +136,9 @@ These are a set of datatypes that model the behavior of common HDL datatypes.
 .. autoclass:: cocotb.types.Logic
     :members:
 
+.. autoclass:: cocotb.types.Bit
+    :members:
+
 .. autoclass:: cocotb.types.Range
     :members:
     :exclude-members: count, index
--- a/src/cocotb/types/_logic.py
+++ b/src/cocotb/types/_logic.py
@@ -4,12 +4,10 @@
 from functools import lru_cache
 from typing import (
     Dict,
-    Set,
-    Type,
     Union,
 )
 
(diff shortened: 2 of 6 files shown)


marlonjames:
Add to Library Reference page of docs.

marlonjames (approved):
The note about it being a subclass of `Logic` is probably enough for clarity,
but `resolve()` is a no-op and `is_resolvable` is always `True`.

src/cocotb/types/_logic.py
```diff
@@ -91,8 +84,6 @@ class Logic:
 
         The :class:`int` and :class:`bool` conversions will raise :exc:`ValueError` if the value is not ``0``, ``1``, ``L``, or ``H``.
 
-    :class:`Logic` values are immutable and therefore hashable and can be placed in :class:`set`\ s and used as keys in :class:`dict`\ s.
```
ktbarrett, line 94: Deleting this as it is currently not true.

ktbarrett:
Maybe I should either change the docs on the inherited methods (`is_resolvable` is calling out `L` and `H` which don't apply) or just remove the inherited methods.

docs/source/library_reference.rst
```diff
@@ -136,6 +136,10 @@ These are a set of datatypes that model the behavior of common HDL datatypes.
 .. autoclass:: cocotb.types.Logic
     :members:
 
+.. autoclass:: cocotb.types.Bit
+    :members:
+    :inherited-members:
```
ktbarrett: This is quicker. We can add more appropriate docs later if needed.
```suggestion
```

PR #4900, opened 2025-08-19
Add `COCOTB_LOG_PREFIX` to customize the default log Formatter's appearance

### LINKED ISSUE #4808: Logging Improvements

ktbarrett:
Depends on #4909. Closes #4808.

In lieu of of using the half-baked `logging.fileConfig` I've added `COCOTB_LOG_PREFIX` envvar to allow users to customize their logging prefix using an f-string.

Example usage:
```bash
# no prefix
COCOTB_LOG_PREFIX=
# [COCOTB] 12.035ns My Message!
COCOTB_LOG_PREFIX="{ansi.YELLOW_FG}[COCOTB]{ansi.DEFAULT_FG {simtime.get_time_from_sim_steps(record.created_sim_time, 'ns')}ns "
# This is the non-color default log format for cocotb -ish
COCOTB_LOG_PREFIX="{simtime.get_time_from_sim_steps(record.created_sim_time, 'ns'):>9}ns {record.levelname:<8} {record.name[-34:]:<34}"
```

Pros:
* Allows customization the prefix with very little hassle (`fileConfig` is half-baked and has a few limitations and requires some additional code to run to be able to use custom formatters).
* No noticeable affect on performance compared to the default logger.

Cons:
* Must keep a curated list of variables to use in evaluation, this may mean on-going feature requests.
* Limited in evaluation logic since it isn't full Python, just f-string.

### TODO
- [x] docs
- [x] test
- [x] newsfrag
- [x] ~~factor ljust and rjust out so they can be used in expressions?~~

### DIFF
--- a/docs/source/building.rst
+++ b/docs/source/building.rst
@@ -130,19 +130,25 @@ Cocotb
     Logs will include simulation time, message type (``INFO``, ``WARNING``, ``ERROR``, ...), logger name, and the log message itself.
     If the value is set to ``0``, the filename and line number where a log function was called will be added between the logger name and the log message.
 
-.. envvar:: COCOTB_ATTACH
+.. envvar:: COCOTB_LOG_PREFIX
 
-    In order to give yourself time to attach a debugger to the simulator process before it starts to run,
-    you can set the environment variable :envvar:`COCOTB_ATTACH` to a pause time value in seconds.
-    If set, cocotb will print the process ID (PID) to attach to and wait the specified time before
-    actually letting the simulator run.
+    Customize the log message prefix.
+    The value of this variable should be in Python f-string syntax.
+    It has access to the following variables:
 
-.. envvar:: COCOTB_ENABLE_PROFILING
+    - ``record``: The :class:`~logging.LogRecord` being formatted. This includes the attribute ``created_sim_time``, which is the simulation time in steps.
+    - ``time``: The Python :mod:`time` module.
+    - ``simtime``: The cocotb :mod:`cocotb.simtime` module.
+    - ``ansi``: The cocotb :mod:`cocotb.ANSI` module, which contains ANSI escape codes for coloring the output.
 
-    Enable performance analysis of the Python portion of cocotb. When set, a file :file:`test_profile.pstat`
-    will be written which contains statistics about the cumulative time spent in the functions.
+    The following example is a color-less version of the default log prefix.
 
-    From this, a callgraph diagram can be generated with `gprof2dot <https://github.com/jrfonseca/gprof2dot>`_ and ``graphviz``.
+    .. code-block:: shell
+
+        COCOTB_LOG_PREFIX="{simtime.convert(record.created_sim_time, 'step', to='ns'):>9}ns {record.levelname:<8} {record.name[-34:]:<34} "
+
+    .. note::
+        If this variable is set, :envvar:`COCOTB_REDUCED_LOG_FMT` has no effect.
 
 .. envvar:: COCOTB_LOG_LEVEL
 
@@ -164,6 +170,20 @@ Cocotb
     This behaves similarly to ``INFO``.
 
     .. versionadded:: 2.0
+
+.. envvar:: COCOTB_ATTACH
+
+    In order to give yourself time to attach a debugger to the simulator process before it starts to run,
+    you can set the environment variable :envvar:`COCOTB_ATTACH` to a pause time value in seconds.
+    If set, cocotb will print the process ID (PID) to attach to and wait the specified time before
+    actually letting the simulator run.
+
+.. envvar:: COCOTB_ENABLE_PROFILING
+
+    Enable performance analysis of the Python portion of cocotb. When set, a file :file:`test_profile.pstat`
+    will be written which contains statistics about the cumulative time spent in the functions.
+
+    From this, a callgraph diagram can be generated with `gprof2dot <https://github.com/jrfonseca/gprof2dot>`_ and ``graphviz``.
 
 .. envvar:: COCOTB_RESOLVE_X
 
@@ -223,7 +243,6 @@ Cocotb
             make simulator_test SIM=<your simulator here> TOPLEVEL_LANG=<vhdl or verilog>
 
         If the tests pass, your simulator and version apply inertial writes as expected and you can turn on :envvar:`COCOTB_TRUST_INERTIAL_WRITES`.
-
 
 Regression Manager
 ~~~~~~~~~~~~~~~~~~
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -318,8 +318,6 @@ Logging
 .. module:: cocotb.logging
     :synopsis: Classes for logging messages from cocotb during simulation.
 
-.. autodata:: strip_ansi
-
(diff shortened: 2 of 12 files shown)


ktbarrett:
One question is, should we un-privatize `cocotb._ANSI` since this is now sort of public via `COCOTB_LOG_PREFIX`?

docs/source/building.rst
```diff
@@ -225,6 +225,27 @@ Cocotb
         If the tests pass, your simulator and version apply inertial writes as expected and you can turn on :envvar:`COCOTB_TRUST_INERTIAL_WRITES`.
 
 
+.. envvar:: COCOTB_LOG_PREFIX
+
+    Customize the log message prefix.
+
+    The value of this variable should be a Python f-string that has access to the following variables:
```
cmarqu:
```suggestion
    The value of this variable should be in Python f-string syntax. It has access to the following variables:
```

cmarqu (approved):
One suggestion.

cmarqu:
I didn't try it yet, so a theoretical question: If the user has an error in this variable, is the resulting error message understandable?
Would it be an improvement if it was mentioning the `COCOTB_LOG_PREFIX` variable?

cmarqu:
Maybe move up the documentation of the variable so that it is next to the other `COCOTBB_LOG_*` variables. (https://cocotb--4900.org.readthedocs.build/en/4900/building.html#envvar-COCOTB_LOG_PREFIX)

cmarqu:
I suppose we should add `cocotb._ANSI` to the library reference docs (and make it public?) now that we refer to it from a public variable.

ktbarrett:
> I didn't try it yet, so a theoretical question: If the user has an error in this variable, is the resulting error message understandable?

No it prints the exception to a fallback logging method (`Handler.handleError`) that makes it very obvious what the error is, but it prints it for every message that comes through. The Handler isn't necessarily in the callstack (`QueueHandler`) so it just prints the exception and continues. It's ugly, but IDK, thoughts?

And yeah I will make `cocotb.ANSI` public again. I'm not sure it's worth documenting everything in it, but I'll add a synopsys along the lines of 

> Included are ANSI escape codes for coloring output.
> Variables that end in `_FG` color the character or symbol and variables ("foreground") that end in `_BG` color the background.
> Foreground and background colors can be combined together;
> but setting a new foreground will override the previous, and the same goes for backgrounds.
> Use `DEFAULT_FG` and `DEFAULT_BG` to reset the coloring to the default colors.

ktbarrett:
@cmarqu Added the ANSI module back, need a re-review.

docs/source/building.rst
```diff
@@ -130,6 +130,27 @@ Cocotb
...
+.. envvar:: COCOTB_LOG_PREFIX
+
+    Customize the log message prefix.
+    The value of this variable should be in Python f-string syntax.
+    It has access to the following variables:
+
+    - ``record``: The :class:`~logging.LogRecord` being formatted. This includes the attribute ``created_sim_time``, which is the simulation time in steps.
+    - ``time``: The Python :mod:`time` module.
+    - ``simtime``: The cocotb :mod:`cocotb.utils` module.
+    - ``ansi``: The cocotb :mod:`!cocotb._ANSI` module, which contains ANSI escape codes for coloring the output.
```
cmarqu:
```suggestion
    - ``ansi``: The cocotb :mod:`cocotb.ANSI` module, which contains ANSI escape codes for coloring the output.
```

cmarqu:
Move the description of `COCOTB_LOG_LEVEL` and `GPI_LOG_LEVEL` next to `COCOTB_LOG_PREFIX`?

cmarqu:
> No it prints the exception to a fallback logging method (`Handler.handleError`) that makes it very obvious what the error is, but it prints it for every message that comes through. The Handler isn't necessarily in the callstack (`QueueHandler`) so it just prints the exception and continues. It's ugly, but IDK, thoughts?

How about if `COCOTB_LOG_PREFIX` is not empty, additionally print "Please check that the COCOTB_LOG_PREFIX environment variable is well-formatted."?
(I imagine that some users might not know that the variable is set in more complicated build environments where they don't control all of it.)

docs/source/library_reference.rst
```diff
@@ -336,6 +334,10 @@ Logging
     :show-inheritance:
     :no-members:
 
+.. autodata:: strip_ansi
+
+.. automodule:: cocotb.ANSI
```
cmarqu: This doesn't create anything that acts as a heading:
https://cocotb--4900.org.readthedocs.build/en/4900/library_reference.html#cocotb.logging.strip_ansi

Not sure right now how to best fix this...

cmarqu: Maybe a `:synopsis:` already helps?

ktbarrett:
@cmarqu Don't worry, I plan on organizing the "building" page a bit better in a follow-on PR.

docs/source/library_reference.rst
ktbarrett: Is the synopsis not the same thing as the module docstring?

cmarqu: `:synopsis:` is taken for the short descriptions in the index (https://cocotb--4900.org.readthedocs.build/en/4900/py-modindex.html) where nobody ever looks; but here I was just hoping it would also trigger a change in the module docs rendering. But what you have also works.

cmarqu: Actually... if we used "proper" enums or something like that in the ANSI module, this problem would solve itself and the user could see the values we have defined.

ktbarrett: If we are going to do that, now is the time. Do we replace `cocotb.ANSI` module with a `cocotb.logging.ANSI` `StrEnum`?

I would do this in a follow-on, of course.

cmarqu: IMO it would be best, yes.

PR #4901, opened 2025-08-19
Simulated time utils improvements

ktbarrett:
Moves the time precision from a function to a variable which is a tad faster and can be overridden by the user for non-simulator based testing.

Also adds a `convert` function to replace the functionality of both `get_sim_steps` and `get_time_from_sim_steps` in a intuitive package.

### TODO
- [ ] docs
- [ ] tests
- [ ] newsfrags


src/cocotb/simtime.py
```diff
@@ -0,0 +1,228 @@
...
+    to: TimeUnit,
+    round_mode: RoundMode = "error",
+) -> float:
+    """Convert times in one unit to another unit.
+
+    Args:
+        value: The time value.
+        unit: The unit of *value*.
+        to: The unit to convert *value* to.
+        round_mode:
```
cmarqu, line 67: Description missing.

src/cocotb/simtime.py
```diff
@@ -0,0 +1,228 @@
...
+        result_rounded = floor(result)
+    else:
+        raise ValueError(f"Invalid round_mode specifier: {round_mode}")
+
+    return result_rounded
+
+
+@lru_cache(maxsize=None)
+def _get_log_time_scale(unit: TimeUnitWithoutSteps) -> int:
+    """Retrieves the ``log10()`` of the scale factor for a given time unit.
```
cmarqu:
```suggestion
    """Retrieve the ``log10()`` of the scale factor for a given time unit.
```

src/cocotb/simtime.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def convert(
+    value: Union[float, Decimal, Fraction],
+    unit: TimeUnit,
+    *,
+    to: TimeUnit,
+    round_mode: RoundMode = "error",
+) -> float:
+    """Convert times in one unit to another unit.
```
cmarqu:
```suggestion
    """Convert time values in one unit to another unit.
```

src/cocotb/simtime.py
ktbarrett: Ah yeah. I was debating if I should copy the description from `get_sim_steps` wholesale since I expect this to be the newer better way and the other functions deprecated at some point.

ktbarrett: I did, more or less.

docs/source/library_reference.rst
```diff
@@ -302,11 +302,15 @@ Asynchronous Queues
 Simulation Time Utilities
 =========================
 
-.. automodule:: cocotb.utils
+.. automodule:: cocotb.simtime
     :members:
     :member-order: bysource
```
cmarqu, line 307:
```suggestion
    :member-order: bysource
    :synopsis: Tools for dealing with simulated time.
```

PR #4903, opened 2025-08-19
Lowercase `debug.DEBUG` to signal its not a constant

ktbarrett:
Was told it was weird that it was uppercase since it's intended to be modified. Fair...

### DIFF
--- a/src/cocotb/debug.py
+++ b/src/cocotb/debug.py
@@ -4,7 +4,7 @@
 
 import os
 
-DEBUG: bool = bool(os.getenv("COCOTB_SCHEDULER_DEBUG"))
+debug: bool = bool(os.getenv("COCOTB_SCHEDULER_DEBUG"))
 """Global flag to enable additional debugging functionality.
 
 Defaults to ``True`` if the :envvar:`COCOTB_SCHEDULER_DEBUG` environment variable is set,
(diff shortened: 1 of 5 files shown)


marlonjames:
Thinking about it, the feature should probably have a newsfrag, and link added to this section: https://docs.cocotb.org/en/development/building.html#envvar-COCOTB_SCHEDULER_DEBUG

src/cocotb/debug.py
```diff
@@ -4,7 +4,7 @@
 
 import os
 
-DEBUG: bool = bool(os.getenv("COCOTB_SCHEDULER_DEBUG"))
+debug: bool = bool(os.getenv("COCOTB_SCHEDULER_DEBUG"))
 """Global flag to enable additional debugging functionality.
 
 Defaults to ``True`` if the :envvar:`COCOTB_SCHEDULER_DEBUG` environment variable is set,
```
marlonjames, line 10: nit: it converts the value to bool. So if it's defined with value 0 it'll be `False`

ktbarrett: Actually it will be `True`. `bool("0")` is `True`. The logic is a bit different than some other envvars. We should probably converge the behavior of boolean flag envvars at some point.

marlonjames: TIL

marlonjames (approved):

PR #4909, opened 2025-08-22
Logger strip ANSI codes

### LINKED ISSUE #4908: Remove checking of `GUI` argument when considering colored output
Not all simulators can't interpret ANSI control codes. Instead, we should specifically default `COCOTB_ANSI_OUTPUT` when `GUI` is set in Makefiles and Runners when we know the simulator can't handle it.

ktbarrett:
Closes #4908.

* Moves `want_color_output` which always checks envvars to a global flag which the user can modify at runtime without messing with envvars.
* Fix handling of `COCOTB_ANSI_OUTPUT` which only changed the value if set to `1` (`COCOTB_ANSI_OUTPUT=0` wouldn't force no colors)
* Add functionality to strip ANSI control codes from messages and rename the color flag to "strip_ansi" to be more specific about what's happening (there are more than coloring control codes).

### TODO
- [x] tests
- [x] newsfrags (fixed `COCOTB_ANSI_OUTPUT=0`, now strip ANSI codes)

### DIFF
--- a/src/cocotb/logging.py
+++ b/src/cocotb/logging.py
@@ -35,7 +35,7 @@ logging.addLevelName(5, "TRACE")
 
 
 strip_ansi: bool = False
-"""Whether the default formatter should strip ANSI escape codes from log messages.
+"""Whether the default formatter should strip ANSI (color) escape codes from log messages.
 
 Defaults to ``True`` if ``stdout`` is not a TTY and ``False`` otherwise;
 but can be overridden with the :envvar:`!NO_COLOR` or :envvar:`COCOTB_ANSI_OUTPUT` environment variable.
@@ -65,7 +65,7 @@ def default_config(
             .. versionadded:: 2.0
 
         strip_ansi:
-            If ``True``, strip ANSI escape codes in log messages.
+            If ``True``, strip ANSI (color) escape codes in log messages.
             If ``False``, do not strip ANSI escape codes in log messages.
             If ``None``, use the value of :data:`strip_ansi`.
 
@@ -89,7 +89,7 @@ def default_config(
 def _init() -> None:
     """cocotb-specific logging setup.
 
-    - Decides whether colored output is supported or desired by checking :envvar:`!NO_COLOR` and :envvar:`COCOTB_ANSI_OUTPUT`.
+    - Decides whether ANSI escape code stripping is desired by checking :envvar:`!NO_COLOR` and :envvar:`COCOTB_ANSI_OUTPUT`.
     - Initializes the GPI logger and sets up the GPI logging optimization.
     - Sets the log level of the ``"cocotb"`` and ``"gpi"`` loggers based on
       :envvar:`COCOTB_LOG_LEVEL` and :envvar:`GPI_LOG_LEVEL`, respectively.


src/cocotb/logging.py
```diff
@@ -76,10 +89,21 @@ def default_config(reduced_log_fmt: bool = True, color: Optional[bool] = None) -
 def _init() -> None:
     """cocotb-specific logging setup.
 
-    Initializes the GPI logger and sets up the GPI logging optimization.
-    Sets the log level of the ``"cocotb"`` and ``"gpi"`` loggers based on
-    :envvar:`COCOTB_LOG_LEVEL` and :envvar:`GPI_LOG_LEVEL`, respectively.
+    - Decides whether colored output is supported or desired by checking :ennvar:`!NO_COLOR` and :envvar:`COCOTB_ANSI_OUTPUT`.
```
cmarqu:
```suggestion
    - Decides whether colored output is supported or desired by checking :envvar:`!NO_COLOR` and :envvar:`COCOTB_ANSI_OUTPUT`.
```

src/cocotb/logging.py
```diff
@@ -18,22 +19,32 @@
...
 )
 
 # Custom log level
 logging.TRACE = 5  # type: ignore[attr-defined]  # type checkers don't like adding module attributes after the fact
 logging.addLevelName(5, "TRACE")
 
 
-def default_config(reduced_log_fmt: bool = True, color: Optional[bool] = None) -> None:
+strip_ansi: bool = False
+"""Whether the default formatter should strip ANSI escape codes from log messages.
```
ktbarrett:
```suggestion
"""Whether the default formatter should strip ANSI (color) escape codes from log messages.
```

src/cocotb/logging.py
```diff
@@ -76,10 +89,21 @@ def default_config(reduced_log_fmt: bool = True, color: Optional[bool] = None) -
 def _init() -> None:
     """cocotb-specific logging setup.
 
-    Initializes the GPI logger and sets up the GPI logging optimization.
-    Sets the log level of the ``"cocotb"`` and ``"gpi"`` loggers based on
-    :envvar:`COCOTB_LOG_LEVEL` and :envvar:`GPI_LOG_LEVEL`, respectively.
+    - Decides whether colored output is supported or desired by checking :envvar:`!NO_COLOR` and :envvar:`COCOTB_ANSI_OUTPUT`.
```
ktbarrett:
```suggestion
    - Decides whether ANSI escape code stripping is desired by checking :envvar:`!NO_COLOR` and :envvar:`COCOTB_ANSI_OUTPUT`.
```

src/cocotb/logging.py
```diff
@@ -53,20 +64,22 @@ def default_config(reduced_log_fmt: bool = True, color: Optional[bool] = None) -
 
             .. versionadded:: 2.0
 
-        color:
-            If ``True``, use colored output in the log messages.
+        strip_ansi:
+            If ``True``, strip ANSI escape codes in log messages.
```
ktbarrett:
```suggestion
            If ``True``, strip ANSI (color) escape codes in log messages.
```

cmarqu:
Yes, best of both worlds.

PR #4916, opened 2025-08-26
Add verilator.cpp flag to dump the trace at every time step

gilbertoabram:
If the simulator crashes in a non-recoverable way, usually a large chunk of
the trace (which could be helpful for debugging) will be missing, because it
is heavily buffered due to compression.
Using this flag allows to inspect the trace right up to the time step prior to
the crash, at the expense of a slower execution and a much larger file.

### DIFF
--- a/src/cocotb/share/lib/verilator/verilator.cpp
+++ b/src/cocotb/share/lib/verilator/verilator.cpp
@@ -78,6 +78,7 @@ int main(int argc, char** argv) {
     const char* traceFile = "dump.vcd";
 #endif
     bool traceOn = false;
+    bool traceFlush = false;
 
     for (int i = 1; i < argc; i++) {
         std::string arg = std::string(argv[i]);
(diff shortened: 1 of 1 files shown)


ktbarrett (commented):
This could be useful for a user to know about (i.e. newsfrag), but at the same time I'm not sure we are documenting the verilated binaries as public, so maybe we don't need to.

ktbarrett (approved):

src/cocotb/share/lib/verilator/verilator.cpp
```diff
@@ -98,16 +101,18 @@ int main(int argc, char** argv) {
...
+            fprintf(
+                stderr,
+                "usage: %s [--trace] [--trace-flush] [--trace-file TRACEFILE]\n"
+                "\n"
+                "Cocotb + Verilator sim\n"
+                "\n"
+                "options:\n"
+                "  --trace       Enables tracing (VCD or FST)\n"
+                "  --trace-flush Flush trace at each time step (slow)\n"
+                "  --trace-file  Specifies the trace file name (%s by "
```
cmarqu:
```suggestion
                "  --trace       Enable tracing (VCD or FST)\n"
                "  --trace-flush Flush trace at each time step (slow)\n"
                "  --trace-file  Specify the trace file name (%s by "
```

PR #4918, opened 2025-08-27
Remove all ocurences of `handle._log` from examples

ktbarrett:
This was deprecated in 2.0. Replaced with `cocotb.log` which is the user's new "print" function.

### DIFF
--- a/examples/simple_dff/test_dff.py
+++ b/examples/simple_dff/test_dff.py
@@ -20,12 +20,14 @@ async def dff_simple_test(dut):
     # Set initial input value to prevent it from floating
     dut.d.value = 0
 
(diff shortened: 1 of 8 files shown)


examples/simple_dff/test_dff.py
```diff
@@ -20,12 +20,14 @@ async def dff_simple_test(dut):
     # Set initial input value to prevent it from floating
     dut.d.value = 0
 
-    clock = Clock(dut.clk, 10, unit="us")  # Create a 10us period clock on port clk
+    # Create a 10us period clock driver on port `clk``
```
cmarqu:
```suggestion
    # Create a 10us period clock driver on port `clk`
```

cmarqu (approved):
One tiny fix.

PR #4924, opened 2025-08-29
Use ccache on Windows in CI

ktbarrett:


ktbarrett:
>Not saving cache because no objects are cached.
https://github.com/cocotb/cocotb/actions/runs/17338888648/job/49229931404?pr=4924#step:69:20

I'm not bothering anymore with this OS.

PR #4926, opened 2025-08-29
Make `cocotb.ANSI` module an enum

ktbarrett:
Per https://github.com/cocotb/cocotb/pull/4900#discussion_r2309261697.

### DIFF
--- a/src/cocotb/regression.py
+++ b/src/cocotb/regression.py
@@ -261,7 +261,7 @@ class RegressionManager:
         Must be called before all modules containing tests are imported.
         """
         try:
-            import pytest
+            import pytest  # noqa: PLC0415
         except ImportError:
             _logger.info(
                 "pytest not found, install it to enable better AssertionError messages"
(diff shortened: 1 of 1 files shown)


src/cocotb/_ANSI.py
```diff
@@ -0,0 +1,63 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+from cocotb._py_compat import StrEnum
+
+_ESCAPE = "\033["
+
+
+class ANSI(StrEnum):
+    """ANSI escape codes for coloring output.
+
```
cmarqu, line 14:
```suggestion
    The color names supported are ``[BRIGHT_]{BLACK|RED|GREEN|YELLOW|BLUE|MAGENTA|CYAN|WHITE}{_FG|_BG}``.

```
Listing them all separately would probably be too much.

src/cocotb/_ANSI.py
```diff
@@ -0,0 +1,63 @@
...
+    Variables that end in ``_FG`` will color the character or symbol ("foreground")
+    and variables that end in ``_BG`` will color the background.
+
+    Foreground and background colors can be combined together.
```
cmarqu:
```suggestion
    Foreground and background colors can be combined together with a ``+``.
```

cmarqu (approved):
Two doc suggestions.

ktbarrett:
The fact that pre-commit is only triggering *now* on regression.py tells me that it probably isn't doing what we need it to be doing.

cmarqu (approved):

PR #4930, opened 2025-08-30
Add optimization and debugging flags to release builds

ktbarrett:
This was moved to pyproject.toml because it wasn't being applied to all runs of cibuildwheel. Also added flags for Windows builds.


ktbarrett:
Cancelled CI because we can't test this.

imphil (approved):
Let's see how far that gets us in a release build :-)

PR #4934, opened 2025-08-31
Verilator+MacOS should now pass in CI

ktbarrett:
Verilator closed verilator/verilator#5404 as there's MacOS in CI now and apparently it works. They are using MacOS 15 for their testing, but I hope we can continue to use MacOS 13 for ours by just forcing an upgrade of packages (seeing as autoconf was the true culprit).


ktbarrett:
Looks like macos-13 is being deprecated soon, so we might as well just upgrade now if 13 still doesn't work. https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down

ktbarrett:
What do you know? It doesn't work. Let's try bumping the version of macos.

ktbarrett:
Verilator v5.040 doesn't build on macos-14... fun... I guess they don't care about older versions of Mac at all. I'll consider the newer version of GCC a pre-requisite and just move to macos-15.

ktbarrett:
Looks like there is an issue with MacOS build in v5.040 (fixed in verilator/verilator#6350 right after release). The  working CI was introduced after the release. I suppose we should downgrade to v5.038 in CI.

imphil (approved):
Fun! Thanks @ktbarrett

PR #4939, opened 2025-09-02
Only run nosim tests on Python version tests

ktbarrett:
Previously the nosim tests were run on every job which is redundant. We really only need to run in all the Python version tests.


ktbarrett:
Removing ccache on macos-14 managed to fix this. I wonder why that's such a problem?

marlonjames:
Maybe add `test_nosim` to `generate-envs.py` docstring under **Optional fields**.

Issue #2930, opened 2022-03-31
Support for archs other than x86_64


allexoll:
Currently cocotb only supports x86_64. 

As Apple is turning away from x86_64 and into arm64, the current Mac user base that was previously running it either on macos-x86_64 or linux-x86_64 in docker/VM will need tools to work on their machine. 

The open source HDL tooling might be especially well setup for this since it seems that this is not on commercial tools's roadmap.

I commented that I was able to build successfully for mac_arm64 and linux_aarch64 in #2924 but this is a larger discussion about the supported platforms of cocotb.

One potential issue is that currently github action does not have arm runners, but it's still possible to have them trough [qemu + docker ](https://github.com/dbhi/qus). 

Happy to talk about it

ktbarrett:
> One potential issue is that currently github action does not have arm runners, but it's still possible to have them trough [qemu + docker ](https://github.com/dbhi/qus).

Looks promising. Can you open a PR? It works a bit differently than I expected since GHA supports using docker containers in "uses" blocks.

imphil:
Let's maybe take them separately.

* aarch64 Linux is getting more and more support with proprietary EDA tools as well (I've seen it in the release notes for VCS and Questa at least). So that's something where apparently users are, and quite in line with our platform support policy. Sounds like an easy one to add cocotb support for as well.
* M1 macs are a bit a different story that I'd like to learn more about. We have a tiny user base on macos anyways, and I'd guess it's even smaller on M1 macs. What's the simulator support story on M1 macs? Are there precompiled versions of Icarus, GHDL, or Verilator available? Or are there other simulators supporting Mac? I'm especially wondering about precompiled versions, since even if we don't officially make M1 macs a supported platform and build wheels for it, users can still compile cocotb from source.

allexoll:
linux aarch64 seems to be the easiest one to add, and the most usefull one since it could also be run trough docker on Macs anyway.

Regarding Macos arm64, icarus, ghdl, verilator are all available precompiled, trough the most used package manager on mac `brew`

```sh
brew install icarus-verilog
brew install ghdl
brew install verilator
```

both linux-aarch64 and macos-arm64 cocotb build succeed but i'm unsure what the framework is for testing it would be, especially since it has to integrate with GA.

So far the way to do it would be:
* x86_64 builds don't need to change
[...]

PR #4943, opened 2025-09-04
Build release bdists for Mac against ARM and x86

ktbarrett:
Build against both ARM64 and x86 for Mac release builds. xref #2930.

We are only testing the ARM image since we are only testing against macos-14 in the regression right now. We should also test against an x86 image (such as the large images, but they cost money, or macos-13 for this release) if we plan to do this. Alternatively we only do release builds for ARM.

### DIFF
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -34,7 +34,8 @@ jobs:
         os:
           - ubuntu-22.04
           - windows-2022
-          - macos-13
+          - macos-13  # x86_64
+          - macos-14  # ARM64
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8  # v5.0.0
       - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,7 +112,7 @@ known-third-party = [
 # - CPython on Linux i686 and x86_64 with glibc
 # - CPython on Windows i686 and x86_64
 # - CPython on macOS x86_64
-build = "cp*-manylinux_x86_64 cp*-manylinux_i686 cp*-win_amd64 cp*-win32 cp*-macosx_x86_64"
+build = "cp*-manylinux_x86_64 cp*-manylinux_i686 cp*-win_amd64 cp*-win32 cp*-macosx_x86_64 cp*-macosx_arm64"
 
 # Build with optimizations and debugging
 # -O2   Enables reasonable optimizations. -O3 is unnecessary with the virtual-call-heavy code in the GPI and may bloat binary size.


ktbarrett:
Local tests show it getting further in the process.

imphil:
All release builds pass with this PR. We currently state at https://docs.cocotb.org/en/development/platform_support.html to only support mac x86. We have an open issue requesting arm support. We can certainly do that and either drop the x86 or support both, but if we do both, we at least need to do basic smoke testing of the release artifacts.

ktbarrett:
I am going to add new groups to split up how we are selecting on these. I also added support for specifying multiple groups in each entry of `ENVS`.

### Release builds

Remove 32-bit Windows and Linux builds (sorry, not sorry). We don't currently have any way to test these images in CI.

### Tagged release testing

This should be the full cross-product of the following:
* All build triplets (python_version-OS-arch)
* All simulators (one version) supported on that OS-arch

### Master release testing

We have two options here:
1. Do the full cross-product of the tagged release testing, but maybe reduce the testing to only one run of `test_cocotb`?
2. Do mostly the CI testing.
   * Only need one environment on any given OS.
   * Only need one version of the simulator (latest).

EDIT: going with 1 since the CI testing will do 2 so together we know the build works and the tests pass.

### CI testing

This is for coverage and compatibility, but it needs to be small, so no cross-products:
* All supported Python versions (OS and simulator don't matter)
* All supported Python versions simulator-less testing
* All simulators latest supported release (Python version and OS don't matter)
* All simulators minimum supported release (Python version and OS don't matter)
* All supported OS/environment (sim and Python version don't matter)

ktbarrett:
So the best way I see to accomplish this is to generate all possible valid environments which is the cross product of
* OS/arch
* Python version
* Simulator+version
* Toplevel language

There will need to be some filters to ensure that invalid combinations aren't created:
* Some Python versions aren't supported on all OSes
* Simulators don't support all OSes
* Simulators don't support all languages

After that "extra" envs can be added like special simulator versions or environments to test (e.g. mingw vs msvc on Windows).

There will be three groups:
* `ci-free`: Limited subset for CI testing. All (FOSS simulators * lang) on Ubuntu with a single Python version and the minimum and maximum supported version to establish a support window. All Python versions on Ubuntu with a single simulator. Testing of other OSes/environments than Ubuntu.
* `ci-licensed`: All (non-FOSS simulators * lang) on Ubuntu with a single Python version and the maximum version to ensure continued support of the most recent version (no minimum version because companies don't support forever and we need to keep license use low).
* `extended`: All non-experimental envs. The cross product of all supported OSes, all simulators+versions supported by that OS, all Python versions supported by that OS, all Languages supported by that simulator. Plus all the special envs.
* `experimental`:  In-development Python versions, HEAD on simulators, things that aren't working in cocotb yet (e.g. VCS VHDL tops)

There will be two test levels:
* `smoke-test`: runs only `test_cocotb`
* `regression`: runs full regression

The PR checks will use `ci-free` and `ci-licensed` groups with `regression` tests. The extended tests will run the `extended` group and the `regression` tests. The experimental tests will run the `experimental` group and the `regression` tests. The release test will run the `extended` group and the `smoke-test` tests. Unless the release is tagged in which case it will run the `extended` group and the `regression` tests.

ktbarrett:
Also I will bring back `macos-13` for doing x86_64 Mac builds and tests, then we will have to either transition to one of the "large" CI machines and pay money to Github to continue supporting x86_64 on Mac, or we can drop it after this release.

imphil:
I'm a bit nervous about doing significant changes to the release pipeline at this time. I'm going to take this PR as-is (without the test commit, obviously). Other refactorings can come after the 2.0 release.

imphil (approved):

PR #4957, opened 2025-09-08
Hardcode parallelism in Windows CI build jobs

ktbarrett:
I have a feeling the reason this job randomly fails due to timeout is that `nproc` is returning nothing and so `make -j $(nproc)` uses the default behavior for argument-less `-j` which is maximum parallelism, one job per target. That often has the effect of blowing up any computer.

I don't feel like trying out different stuff to see what the analogue to `nproc` is on Windows.


cmarqu:
I like the "Hardcore" typo in the commit message :)

PR #4960, opened 2025-09-09
Add removal newsfrag for `TestFailure`, `TestError`, and `TestComplete`

ktbarrett:
I realized when doing the documentation nothing mentioned they were removed. So I added deprecation shims instead.


marlonjames:
Eh, `TestFailure` was deprecated in 1.6.0 (https://docs.cocotb.org/en/development/release_notes.html#id31) and `TestError` in 1.5.0 (https://docs.cocotb.org/en/development/release_notes.html#id38).
I would remove them instead and just add a newsfrag.

cmarqu (approved):

marlonjames (approved):

PR #4963, opened 2025-09-09
Make markers for links to old versions stand out less

cmarqu:


marlonjames:
I like the smaller corner radius better, but you can make the call.

PR #4975, opened 2025-09-12
Allow the deploy target write access to PyPi

imphil:


imphil:
Maybe fix the fail reported at https://github.com/cocotb/cocotb/actions/runs/17671104923.

PR #4983, opened 2025-09-15
Bump cocotb version to 2.1.dev0

ktbarrett:


setup.py
```diff
@@ -46,7 +46,7 @@ def package_files(directory):
 
 
 version_file_path = path.join("src", "cocotb", "_version.py")
-__version__ = "2.0.0"
+__version__ = "2.1.dev0"
```
marlonjames, line 49: We've used the 3 digit form up till now.
```suggestion
__version__ = "2.1.0.dev0"
```

ktbarrett: Yeah and it's kinda wrong/unnecessary. We are developing 2.1 and 2.1.0 is the first release. "devN" is always less than a release and you can add an arbitrary 0s to any version and it's just the same version.

marlonjames: I'm fine either way really. We did update the docs to use `1.9` rather than `1.9.0`, etc.

ktbarrett:
@imphil You are the packaging guy, you get final say.

setup.py
imphil: Semver is always three digits.

imphil:
I did see this PR too late and already bumped master to 2.1.0.dev0, feel free to commit 2.1 changes to master now.

PR #4949, opened 2025-09-07
Figure out why Icarus 11.0 fails to build

imphil:

PR #4986, opened 2025-09-16
Drop 32-bit builds and MacOS 13 support

ktbarrett:
Depends on #5056.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -33,8 +33,6 @@ The current stable version of cocotb requires:
 
 .. versionchanged:: 1.4 Dropped Python 2 support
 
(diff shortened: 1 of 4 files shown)


marlonjames (commented):
`install.rst` mentions 32-bit.

ktbarrett:
Was easy enough to do from a phone.

marlonjames:
Newsfrags are #4949 ... should I change them to this PR?

ktbarrett:
Please, I dont know if I can do that via web UI.

ktbarrett:
We should add a warning when building on x86 Mac targets that it isn't directly supported and may not work.

PR #4987, opened 2025-09-16
Drop support for RHEL8 and Python 3.6, 3.7, and 3.8

ktbarrett:
Depends on #4983 and #4986.


ktbarrett:
I think we will need https://github.com/cocotb/cocotb/commit/7589d322dacdccc3c0b95fc509a759bc5535a22a as well.

docs/source/platform_support.rst
```diff
@@ -44,9 +44,6 @@ We encourage especially users less experienced with Linux to use one of the supp
 Linux distributions typically ship a default version of Python called "system Python".
 This version of Python can be used with cocotb unless noted otherwise.
 
-* **Red Hat Enterprise Linux (RHEL) 8 amd64**,
-  shipping with Python 3.6 (newer versions of Python are available), pip 9, and glibc 2.28.
-  `Upstream support until May 2029 <https://access.redhat.com/support/policy/updates/errata#Life_Cycle_Dates>`_.
```
imphil, line 49: RHEL is still reasonably widely used in my experience, and also supported by simulator vendors. We can't drop it just yet. We *can*, however, drop Python 3.6, RHEL8 ships up to Python 3.12 in its appstream repos.

ktbarrett:
~~Ugh. Okay, well disabling TC001,2,3 and moving the imports out of the TYPE_CHECKING blocks is causing cyclic dependencies, so I guess I'll solve that first in another PR.~~

~~The issue is that the write scheduler in `handle.py` depends upon the `ReadWrite` trigger in `_gpi_triggers.py`, but the value change callbacks in `_gpi_triggers.py` depends upon the handle types in `handle.py`. This was previously not an issue because we were importing things in TYPE_CHECKING blocks which use type checker's deferred evaluation.~~

Will resolve later instead.

cmarqu:
Doesn't have a newsfragment yet, right?

ktbarrett:
I don't have the haziest idea how anything in this PR is causing FLI to segfault.

ktbarrett:
Well... so much for dicts retaining insertion order.

ktbarrett:
Or Questa just being Questa.

ktbarrett:
Fixed codecov build count.

docs/source/newsfragments/4987.removal.rst
```diff
@@ -0,0 +1 @@
+Support for Python 3.6, 3.7, and 3.8 has been removed.
```
cmarqu:
```suggestion
Support for Python 3.6, 3.7, and 3.8 has been removed, according to our :ref:`platform-support-policy`.
```

ktbarrett: Eh, I would think our "platform support policy" would actually recommend we keep supporting them. The real reason is it's getting too painful. It's a lie softened with corporate-y language.

PR #4988, opened 2025-09-16
Fix Python 3.6 and ensure correct version of Python in CI

### LINKED ISSUE #4984: Cocotb 2.0 failing to start on Python 3.6
Hello,

When running on Python 3.6.8, upon `import cocotb` I receive an error in _typing.py, line 6:
RoundMode: TypeAlias = Literal["error", "round", "ceil", "floor"]
TypeError: 'type' object is not subscriptable

I believe the culprit is in _py_compat.py:
```   from typing import Any

    class FakeGetItemType:
        def __class_getitem__(self, a: object) -> Any:
            return Any

    Final = FakeGetItemType
    Literal = FakeGetItemType
```

From what I can tell I think `__class_getitem__` was added in python 3.7, and is thus being ignored in python 3.6

Thanks

ktbarrett:
Closes #4984.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -110,22 +110,25 @@ jobs:
       with:
         python-version: ${{matrix.python-version}}
         allow-prereleases: true
-    - name: Set up Python ${{matrix.python-version}} (pyenv)
+    - name: Set up Python ${{matrix.python-version}} (pyenv pt.1)
       if: matrix.setup_python == 'pyenv' && (startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macos'))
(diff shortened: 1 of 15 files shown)


docs/source/release_notes.rst
```diff
@@ -7,6 +7,16 @@ Release Notes
...
+
+cocotb 2.0.0 (TBD)
+==================
+
+Bugfixes
+--------
+
+- Fix break of Python 3.6 (:pr:`4988`)
+- Fix CI to use the correct version of Python in tests (:pr:`4988`)
+
```
imphil: Please make this a normal newsfrag.

imphil (approved):
LGTM whenever it works. Please merge to master and add a backport for stable/2.0 tag.

PR #4995, opened 2025-09-20
Windows CI Improvements

marlonjames:
Try Installing Icarus through MSYS2 like their CI.
Install NVC with action.

Try running nox through MSYS2 for make.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -220,7 +220,11 @@ jobs:
       uses: nickg/setup-nvc@v1
       with:
         version: ${{matrix.sim-version}}
-
+    - name: Set up NVC (Windows, release)
+      if: startsWith(matrix.os, 'windows') && matrix.sim == 'nvc'
+      uses: nickg/setup-nvc@v1
+      with:
+        version: ${{matrix.sim-version}}
     - name: Set up NVC (Ubuntu, source)
       if: startsWith(matrix.os, 'ubuntu') && matrix.sim == 'nvc' && (!startsWith(matrix.sim-version, 'r') || matrix.cc == 'clang')
       run: |
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -105,7 +105,7 @@ jobs:
 
       # Install Python
     - name: Set up Python ${{matrix.python-version}} (setup-python)
-      if: matrix.setup_python == '' && (startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macos'))
+      if: matrix.setup_python == ''
       uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c  # v6.0.0
       with:
         python-version: ${{matrix.python-version}}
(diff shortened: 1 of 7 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -174,18 +164,32 @@ jobs:
...
+    - name: Set up Icarus (Windows - source - pt. 2)
+      if: startsWith(matrix.os, 'windows') && matrix.sim == 'icarus'
+      timeout-minutes: 10
+      shell: msys2 {0}
+      env:
+        MINGW_ARCH: MINGW64
+      run: |
+        cd iverilog
+        cd msys2
+        makepkg-mingw --noconfirm --noprogressbar -sCLf
```
ktbarrett, line 184: What's this line doing?

noxfile.py
```diff
@@ -193,8 +194,9 @@ def dev_test_sim(
     with suppress(FileNotFoundError):
         coverage_file.unlink()
 
-    session.log(f"Running 'make test' against a simulator {config_str}")
-    session.run("make", "-k", "test", external=True, env=env)
+    if "COCOTB_CI_SKIP_MAKE" not in os.environ:
```
ktbarrett, line 197: Why are we skipping make?

.github/workflows/regression-tests.yml
marlonjames: Building icarus. Based on https://github.com/steveicarus/iverilog/blob/master/.github/workflows/test.yml

src/cocotb_tools/makefiles/simulators/Makefile.nvc
```diff
@@ -16,10 +16,10 @@ ifdef NVC_BIN_DIR
 else
     # auto-detect bin dir from system path
     CMD := $(shell :; command -v $(CMD_BIN) 2>/dev/null)
-	NVC_BIN_DIR := $(shell dirname $(CMD))
+	NVC_BIN_DIR := $(shell dirname "$(CMD)")
 endif
 
-PRESERVE_CASE := $(shell $(CMD) --version | $(PYTHON_BIN) -c "from cocotb_tools.sim_versions import NvcVersion; import sys; print('--preserve-case' if NvcVersion.from_commandline(sys.stdin.read()) > NvcVersion('1.16') else '')")
+PRESERVE_CASE := $(shell "$(CMD)" --version | $(PYTHON_BIN) -c "from cocotb_tools.sim_versions import NvcVersion; import sys; print('--preserve-case' if NvcVersion.from_commandline(sys.stdin.read()) > NvcVersion('1.16') else '')")
```
ktbarrett: Why is adding quotes necessary? I understand paths can have spaces in them, but this should still work, no? Or is it that it gets interpreted as separate arguments?

Should we be wrapping all `$(SIM)` in quotes?

.github/workflows/regression-tests.yml
```diff
@@ -215,7 +219,17 @@ jobs:
...
+      if: startsWith(matrix.os, 'windows') && matrix.sim == 'nvc'
+      uses: nickg/setup-nvc@v1
+      with:
+        version: ${{matrix.sim-version}}
+    - name: Set up NVC (Windows, release - pt. 2)
+      if: startsWith(matrix.os, 'windows') && matrix.sim == 'nvc'
+      shell: msys2 {0}
+      env:
+        MINGW_ARCH: MINGW64
+      run: export PATH=/c/Program\ Files/NVC/bin:$PATH
```
ktbarrett:
```suggestion
      run: echo "/c/Program\ Files/NVC/bin" >> $GITHUB_PATH
```

.github/workflows/regression-tests.yml
```diff
@@ -174,18 +164,32 @@ jobs:
...
+        makepkg-mingw --noconfirm --noprogressbar -sCLf
+    - name: Set up Icarus (Windows - source - pt. 3)
+      if: startsWith(matrix.os, 'windows') && matrix.sim == 'icarus'
+      timeout-minutes: 10
+      shell: msys2 {0}
+      env:
+        MINGW_ARCH: MINGW64
+      run: |
+        pacman -U --noconfirm iverilog/msys2/*.zst
+        echo "$(which iverilog)"
```
ktbarrett:
```suggestion
        echo "$(which iverilog)" >> $GITHUB_PATH
```

ktbarrett:
~~That might not work at all because they are different shells, but maybe?~~

ktbarrett:
I got NVC working, but no dice on exporting the Icarus build. It also takes longer to build. I think we just drop Icarus on Windows and test with NVC instead.

This PR will also need a newsfrag for adding support for NVC on Windows.

noxfile.py
ktbarrett: Nvm I see why. No running make on Windows anymore. That's okay by me, especially after #4998 is in.

.github/workflows/regression-tests.yml
```diff
@@ -215,7 +219,17 @@ jobs:
...
+      if: startsWith(matrix.os, 'windows') && matrix.sim == 'nvc'
+      uses: nickg/setup-nvc@v1
+      with:
+        version: ${{matrix.sim-version}}
+    - name: Set up NVC (Windows, release - pt. 2)
+      if: startsWith(matrix.os, 'windows') && matrix.sim == 'nvc'
+      shell: msys2 {0}
+      env:
+        MINGW_ARCH: MINGW64
+      run: echo "/c/Program\ Files/NVC/bin" >> $GITHUB_PATH
```
ktbarrett: Shouldn't setup-nvc be doing this? Do we need to create an issue?

cmarqu (approved):

ktbarrett:
Added a newsfrag for supporting NVC on Windows.

src/cocotb_tools/makefiles/simulators/Makefile.nvc
marlonjames: Spaces in the path were causing issues, but if we don't use make on Windows, this could be reverted.

marlonjames (commented):
Needs NVC windows to generate-envs.py
```
    # NVC windows
    {
        "lang": "vhdl",
        "sim": "nvc",
        "sim-version": "r1.17.1",
        "os": "windows-latest",
        "python-version": "3.11",
        "group": "ci-free",
        "test_nosim": True,
    }
```

src/cocotb_tools/makefiles/simulators/Makefile.nvc
ktbarrett: It's probably a good idea to do it universally just in case. But since I have to modify the PR already I'll back this change out.

PR #4998, opened 2025-09-22
Convert tests to use the Python runner

cmarqu:


ktbarrett:
Are we gonna test this? Like maybe a noxfile run where it uses pytest instead of makefiles?

cmarqu:
Yes, but probably only next week.

cmarqu:
On hold, maybe we can go directly to use pytest (#5090).

PR #5000, opened 2025-09-23
Add missing "verilator" to supported_sims variable

cmarqu:
This bug prevented `cocotb-config --lib-name-path vpi verilator` from working.


ktbarrett:
Let's just ignore the Questa error for now and see what's up later.

PR #5007, opened 2025-09-26
Stackable test decorators

ktbarrett:
This PR makes decorators "stackable". This means that multiple decorators can be applied to a single test function.
Applying `cocotb.parametrize` multiple times will add all options across all invocations together. Applying `cocotb.test` multiple times will:
* merge all exceptions mentioned in all `expect_error`
* logically or the conditions of all `skip`s and all `expect_fail`s
* set the `timeout_time` and unit, `stage`, and `name` according to the last invocation (these should only ever be specified once).

Additionally, this introduces `cocotb.skipif` and `cocotb.xfail` to replace the `skip`, and `expect_error` and `expect_fail` argument to `cocotb.test`. These allow the user to specify conditions and reasons which function as documentation and may also be tracked and output (future work). They are also more readable than giant `or` chains and comments.

### DIFF
--- a/src/cocotb/_decorators.py
+++ b/src/cocotb/_decorators.py
@@ -518,3 +518,55 @@ def skipif(
         return obj
 
     return decorator
+
+
+def xfail(
+    condition: bool = True,
+    *,
+    reason: str | None = None,
+    raises: type[BaseException] | Iterable[type[BaseException]] | None = None,
+) -> Callable[[TestFuncType | TestGenerator], TestGenerator]:
+    """Marks a test as expected to fail if the condition is ``True``.
+
+    This acts as an alternative to the ``expect_fail`` option to :dec:`cocotb.test` if *raises* is not given.
+    Or as an alternative to ``expect_error`` if *raises* is given.
+
+    .. code-block:: python
+
+        @cocotb.xfail(reason="The HDL does not behave as expected.")
+        @cocotb.test
+        async def test_design(dut): ...
+
+
+        @cocotb.xfail(
+            cocotb.top.USE_MY_FEATURE.value != 0,
+            raises=IndexError,
+            reason="My design is not configured to use the new feature, so the model will eventually fail",
+        )
+        @cocotb.test
+        async def test_design_feature(dut): ...
+
+    Args:
+        condition: The condition as to whether the test is expected to fail. Defaults to ``True``.
+        reason: A string giving the reason as to why this test is expected to fail.
+
+            This argument is purely for documentation purposes.
+
+        raises: An exception, or iterable of exceptions to expect the test to fail with.
+    """
+
+    def decorator(obj: TestFuncType | TestGenerator) -> TestGenerator:
+        if not isinstance(obj, TestGenerator):
+            obj = TestGenerator(obj)
+        if condition:
+            if raises is not None:
+                if isinstance(raises, type):
+                    obj.expect_error.add(raises)
+                else:
+                    for exc in raises:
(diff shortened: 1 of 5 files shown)


src/cocotb/_decorators.py
```diff
@@ -62,59 +64,72 @@ class Test:
...
+        self.expect_error: set[type[BaseException]] = set()
+        self.skip = False
+        self.stage = 0
+        self.name = self.func.__qualname__
+        self.module = self.func.__module__
+        self.doc = self.func.__doc__
+        if self.doc is not None:
+            # cleanup docstring using `trim` function from PEP257
+            self.doc = inspect.cleandoc(self.doc)
+        self.fullname = f"{self.module}.{self.name}"
```
ktbarrett:
```suggestion
```

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -308,15 +321,17 @@ async def test_real_assign_double(dut):
...
-@cocotb.test(
-    expect_error=AttributeError
-    if SIM_NAME.startswith("icarus")
-    else AttributeError
-    if SIM_NAME.startswith("ghdl")
-    else ()
+@cocotb.xfail(
+    SIM_NAME.startswith("icarus"),
+    raises=AttributeError,
+    reason="iverilog unable to find real signals (gh-2590)",
```
ktbarrett, line 328: Without `@cocotb.test`.

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -308,15 +321,17 @@ async def test_real_assign_double(dut):
...
 )
+@cocotb.xfail(
+    SIM_NAME.startswith("ghdl"),
+    raises=AttributeError,
+    reason="GHDL unable to find real signals (gh-2589)",
+)
+@cocotb.test
```
ktbarrett, line 335: With `@cocotb.test`.

src/cocotb/_decorators.py
```diff
@@ -62,59 +64,71 @@ class Test:
             Tests from earlier stages are run before tests from later stages.
     """
 
+    # TODO Replace with dataclass in Python 3.7+
```
marlonjames, line 67: Is this still relevant?

src/cocotb/_decorators.py
```diff
@@ -492,9 +473,95 @@ async def my_test_2(arg1: int, arg2: int, arg3: int) -> None: ...
...
+        @cocotb.skipif(
+            cocotb.top.USE_MY_FEATURE.value != 1,
+            reason="The design doesn't support my feature.",
+        )
+        @cocotb.test
+        async def test_my_feature(dut) -> None: ...
+
+    Args:
+        condition: The condition as to whether the test should be skipped. Defaults to ``True``.
+        reason: A string giving the reason as to why this test was skipped.
```
marlonjames, line 506: *reason* is just here for documentation?

src/cocotb/_decorators.py
ktbarrett: Yeah, but I don't feel like doing it rn.

src/cocotb/_decorators.py
ktbarrett: Do you mean I should add on that it doesn't really do anything rn?

marlonjames: No, more just clarifying that it is intended that the value isn't used at all.

marlonjames (approved):

Issue #4642, opened 2025-04-27
pytest-cocotb?


oscargus:
With the new pytest-driven approach it would be nice to be able to pass arguments to pytest. rather than through environment variables. If nothing else, being able to pass, say, `--sim=nvc` setting the environment variable `SIM` to `nvc` for the current run. This will lead to a more consistent experience IMHO and will make it possible to list arguments as part of `pytest --help`. Also, some currently long environment variables can be accessed in a shorter way.

To me, it would make sense that this is hosted under the cocotb organization, although one can of course imagine starting somewhere else.

ktbarrett:
This sounds nice, but it should definitely be a separate package. There are many possible uses of pytest with cocotb so anything that prevents other users from using cocotb with pytest in their own way should be developed separately.

Also I'm not so sure about how stable the pytest plugin API is, but if it changes a lot, putting that package in the cocotb repo might not be a good idea as cocotb moves a lot slower than might be required to best support such a plugin (but maybe not, I don't know).

> If I understand correctly, cocotb is already registering to pytest?

I'm not sure what you mean, so I'm going to guess no.

oscargus:
> I'm not sure what you mean, so I'm going to guess no.

https://github.com/cocotb/cocotb/blob/0b47978253170dceec0e6ff4e0245daf64b19cb6/src/cocotb/regression.py#L202

sort of. That cocotb already now interacts with pytest.

I think that one can, maybe, do something similar with pytest arguments. But I get the point of having a separate package. On the other hand, passing things with environment variables may be quite tedious and it can be hard to guarantee that they are only set temporarily (see https://stackoverflow.com/questions/2059482/temporarily-modify-the-current-processs-environment for options).

ktbarrett:
The assertion writing is a bit different. We aren't registering ourselves into pytest, but registering pytest's assertion rewriting import hook. It works standalone. Same with cocotb supporting `pytest.raises`, `pytest.warns`, and in 2.1 `pytest.skip` and `pytest.xfail`. cocotb doesn't insert itself into the pytest flow in any way right now.

Based on https://docs.pytest.org/en/stable/how-to/writing_plugins.html we could keep the plugin code in the cocotb repo and require users to load the plugin via conftest.py or by adding `-p cocotb` to the commandline call or their `pytest.ini`. That makes it opt-in.

PR #5013, opened 2025-09-30
Add plugin facility to extend RegressionManager

tymonx:
This feature allows to extend or replace default implementation of `RegressionManager` using plugin facility.

New plugin can be created as standalone Python project.

With this feature, it is now possible to integrate and use any kind of existing testing framework like [pytest](https://docs.pytest.org) with cocotb test functions.

Related issues:
- https://github.com/cocotb/cocotb/issues/4642
- https://github.com/cocotb/cocotb/issues/494

Using [pytest](https://docs.pytest.org) with `cocotb` test functions will be possible via Pytest/Cocotb plugin `pytest_cocotb`:
- https://gitlab.com/tymonx/pytest-cocotb

Above plugin will allow to use **ALL** [pytest](https://docs.pytest.org) capabilities like [fixtures support](https://docs.pytest.org/en/stable/explanation/fixtures.html), [tests parametrization](https://docs.pytest.org/en/stable/example/parametrize.html) and using any kind of existing [pytest plugins](https://docs.pytest.org/en/stable/reference/plugin_list.html). 

GitLab CI job examples:
- https://gitlab.com/tymonx/pytest-cocotb/-/jobs/11543337062

Remarks:
- Almost imperceptible changes in existing cocotb codebase
- Documentation `Writing cocotb Plugins`
- Unit tests with 100% code coverage for `cocotb.factory.Factory` class
- Moving support for [pytest](https://docs.pytest.org)  to separate project (shifting responsibilities)
- Possible to integrate and use other Python testing framework with cocotb test functions

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -123,6 +123,7 @@ A test can spawn multiple coroutines, allowing for independent flows of executio
    custom_flows
    rotating_logger
    extensions
+   plugins
    upgrade-2.0
    update_indexing
 
--- a/docs/source/newsfragments/5013.feature.rst
+++ b/docs/source/newsfragments/5013.feature.rst
@@ -1,0 +1,1 @@
+Added plugin facility that will allow to extend or override the :class:`~cocotb.regression.RegressionManager` to fully support `pytest <https://docs.pytest.org>`_.
--- a/docs/source/plugins.rst
+++ b/docs/source/plugins.rst
@@ -1,0 +1,106 @@
+**********************
+Writing cocotb Plugins
+**********************
+
+This guide explains how to write a custom Python plugin to extend existing cocotb capabilities.
+
+cocotb gives its users a framework to build Python testbenches for hardware designs.
+
+For example, it comes with default implementation of :class:`cocotb.regression.RegressionManager`
+that handles execution of test functions decorated with :func:`cocotb.test`.
+
+But Python world offers a wide choice of different testing frameworks with far more
+advanced features that can help boost quality and productivity of created tests.
+Like for example `pytest`_ with
+`fixtures support <https://docs.pytest.org/en/stable/explanation/fixtures.html>`_,
+`tests parametrization <https://docs.pytest.org/en/stable/example/parametrize.html>`_ and
+`large number of existing plugins <https://docs.pytest.org/en/stable/reference/plugin_list.html>`_
+that can be used in RTL verification as well.
+
+cocotb plugin facility allows to integrate and use these existing testing frameworks
+without sacrificing any of cocotb capabilities.
+
+Creating Plugin Project
+=======================
+
+The best way to create our custom cocotb plugin is to use Python entry points available with
+`package metadata <https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-package-metadata>`_.
+
+For that, we need to create a small standalone Python project for our plugin.
+It requires only few files.
+
+Example tree structure of plugin project:
+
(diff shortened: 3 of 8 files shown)


tymonx:
@ktbarrett Thanks for feedback!

> This seems unnecessary.

Definitely it was unnecessary :sweat_smile:  Classic RTFM. I just read about [PYGPI_USERS](https://docs.cocotb.org/en/stable/library_reference.html#envvar-PYGPI_USERS) environment variable and implementation of [pygpi/entry.py:load_entry()](https://github.com/cocotb/cocotb/blob/stable/2.0/src/pygpi/entry.py#L7). Indeed, it is exactly what I wanted.

> Modifying PYGPI_USERS to point to something other than the RegressionManager should be sufficient. We don't need to make the RegressionManager API something we need to provide or require.

Indeed. I have slightly modified my pytest plugin:

```python
ENTRY_POINT: str = "pytest_cocotb.plugin:run_regression"


def run_regression(_: object) -> None:
    # using pytest as tests regression manager


def pytest_configure(config: Config) -> None:
    pygpi_users: list[str] = os.environ.get("PYGPI_USERS", ENTRY_POINT).split(",")

    if ENTRY_POINT not in pygpi_users:
        pygpi_users.append(ENTRY_POINT)

    os.environ["PYGPI_USERS"] = ",".join(pygpi_users)
```

And it worked like a charm :confetti_ball: 

> Also I was under the impression the intent was to include the pytest as RegressionManager code in cocotb_tools and not as an extension?

I can do that and prepare required Pull Request. Code could be placed in `src/cocotb_tools/pytest_plugin.py` location and mentioned `ENTRY_POINT` variable in this case will contain `cocotb_tools.pytest_plugin:run_regression`. And Cocotb `pyproject.toml` file can contain:

```toml
[project.entry-points.pytest11]
cocotb = "cocotb_tools.pytest_plugin"
```

This will automatically load `pytest_plugin` by `pytest` when Cocotb will be installed. With further possibility to disable it and use default Cocotb regression manager. Alternatively user can decide to define that entry in their `pyproject.toml` files to use pytest as regression manager in Cocotb.

I think one benefit of implementing that as separate pytest plugin (Python project) will allow to use it with existing Cocotb 2.0 and 1.x? But I can also contribute to `cocotb_tools` and make it official as part of Cocotb 2.1.

> Finally, this definitely doesn't work with cocotb.start_soon, more refactoring of cocotb is necessary for that.

Not exactly. It can even work with existing Cocotb 2.0! I was able to unroll `pytest:main` and implement similar logic like in original Cocotb RegressionManager by using callback functions to jump from start session -> test A -> test B -> ... -> end session.

tymonx:
I'm closing this Pull Request. Using only [PYGPI_USERS](https://docs.cocotb.org/en/stable/library_reference.html#envvar-PYGPI_USERS) environment variable is sufficient to what I want to achieve.

ktbarrett:
Nice! Glad that avoided some unnecessary code.

> > Finally, this definitely doesn't work with cocotb.start_soon, more refactoring of cocotb is necessary for that.
> 
> Not exactly. It can even work with existing Cocotb 2.0! I was able to unroll `pytest:main` and implement similar logic like in original Cocotb RegressionManager by using callback functions to jump from start session -> test A -> test B -> ... -> end session.

This implies its currently reaching into some internals to get stuff to work. Internals change and maintaining that is gonna suck. I'd like to get refactors in place by 2.1 so you don't have to do that.

I guess I'm surprised it's that simple.

PR #5019, opened 2025-10-01
Clarify str vs range length mismatch when initializing LogicArray

gilbertoabram:
The exception message makes it seem that it should be possible to initialize a LogicArray
with a string as long as it fits within the range, but they need to have matching lengths.

### DIFF
--- a/src/cocotb/types/_logic_array.py
+++ b/src/cocotb/types/_logic_array.py
@@ -257,7 +257,7 @@ class LogicArray(AbstractMutableArray[Logic]):
             if range is not None:
                 if len(self._value_as_array) != len(range):
                     raise ValueError(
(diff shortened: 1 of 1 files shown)


ktbarrett (approved):

src/cocotb/types/_logic_array.py
```diff
@@ -257,7 +257,7 @@ def __init__(
             if range is not None:
                 if len(self._value_as_array) != len(range):
                     raise ValueError(
-                        f"Value of length {len(self._value_as_array)} will not fit in {range!r}"
+                        f"Value of length {len(self._value_as_array)} does not match bounds: {range!r}"
```
ktbarrett:
```suggestion
                        f"Value of length {len(self._value_as_array)} does not match range: {range!r}"
```

ktbarrett (approved):

Issue #5016, opened 2025-10-01
Tagging files as Verilog or VHDL cannot work


eschmidscs:
Hi

Using version 2.0.0, python runner.
Tagging files as VHDL or Verilog in the source file list does not work.

In the runner class, build method, the absolute path is evaluated here:
https://github.com/cocotb/cocotb/blob/2cdad006e742fff7bface9f56ddbb1e4abe96056/src/cocotb_tools/runner.py#L317
The get_abs_path function converts everything into Path objects, thus losing the VHDL/Verilog information.

I'm trying to provide a PR, tooling issues in the way...

Best regards,
emanuel

PR #5023, opened 2025-10-02
Keep object type through abs_path

eschmidscs:
Fix for #5016 that works for me.

### DIFF
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -605,17 +605,25 @@ def outdated(output: Path, dependencies: Sequence[Path]) -> bool:
     return dep_mtime > output_mtime
 
 
-def get_abs_path(path: PathLike) -> Path:
+def get_abs_path(path: Union[PathLike, VHDL, Verilog]) -> Union[Path, VHDL, Verilog]:
     """Return *path* in absolute form."""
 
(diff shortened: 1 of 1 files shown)


ktbarrett (commented):
There is also `VerilatorControlFile`. And I'm not sure about this returning a mix of string-likes and Path. I think instead we should make `Verilog`, `VHDL`, and `VerilatorControlFile` inherit from `Path`.

eschmidscs:
I have missed the VerilatorControlFile because I started from v2.0.0, not master, I guess.

Not sure if I find time to go ahead from here.
Would you just inherit from Path instead of str, and that's it? Do you see if that could have problematic side effects? I'm rather new to all this.

ktbarrett:
Solving this the right way is too much effort for now. Let's just get this in and hope it tides us over.

EDIT: Eh. Actually this doesn't pass typing as is.

ktbarrett:
@eschmidscs Can you try out https://github.com/cocotb/cocotb/pull/5033?

ktbarrett:
Superseded by #5033.

eschmidscs:
> @eschmidscs Can you try out #5033?

Sorry, I was out of office this week.
Anyway, the merged solution seems to work here.
Thanks a lot!

PR #5026, opened 2025-10-03
Fix typing on get_sim_time

ktbarrett:
Found while adding typing to pyuvm.


src/cocotb/simtime.py
```diff
@@ -88,6 +88,12 @@ def convert(
         return _get_time_from_sim_steps(steps, to)
 
 
+@overload
+def get_sim_time(unit: Steps = "step", *, units: None = None) -> int: ...
+
+@overload
+def get_sim_time(unit: TimeUnitWithoutSteps = "step", *, units: None = None) -> float: ...
```
ktbarrett:
```suggestion
def get_sim_time(unit: TimeUnitWithoutSteps, *, units: None = None) -> float: ...
```

cmarqu (approved):

PR #5033, opened 2025-10-04
Overhaul sources and build_args to maintain tag and value separately

### LINKED ISSUE #5016: Tagging files as Verilog or VHDL cannot work

ktbarrett:
Closes #5016. 

This re-implements the logic for tagging across all runners so that the tag and the value are kept separately. This should avoid any destruction of the tags as was reported in #5016. I can imagine this might need to spread to `elab_args`, `analysis_args`, etc. But we will cross that bridge when we get there.

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -37,6 +37,8 @@ Python Test Runner
 
 .. autoclass:: Verilog
 
+.. autoclass:: VerilatorControlFile
+
 .. autodata:: MAX_PARALLEL_BUILD_JOBS
 
 .. envvar:: GUI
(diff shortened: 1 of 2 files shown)


cmarqu:
We aren't really documenting the `VerilatorControlFile` tag, maybe fix this in this PR as well?
https://cocotb--5033.org.readthedocs.build/en/5033/library_reference.html#cocotb_tools.runner.Runner.build

ktbarrett:
@cmarqu I added it.

cmarqu:
> @cmarqu I added it.

Thanks. Should it also be mentioned in the docstring after "If your source files use an atypical file extension...", and in the File Extension table/the code example?

PR #5048, opened 2025-10-06
Makefile.icarus: make $FST lower priority in argument order

dlmiles:
This follows the convention of trying to make framework provided settings a low precedence than user provided custom settings such as $SIM_ARS $EXTRA_ARGS which now have a chance to override the framework on command line.

Also since $WAVES=1 is the expected method to set $FST there is also a defacto standard that env-var settings have lower precedence than command line arguments.

So this is also consistent with conceptual expectations.

explain-why: The cocotb2 introduces this new value $FST but it has been added at an extremely high precedence order for a setting derived from `[["$WAVES"=="1"]]` the precedence is so high it is impossible to override with command line arguments from $SIM_ARGS or $EXTRA_ARGS

### DIFF
--- a/src/cocotb_tools/makefiles/simulators/Makefile.icarus
+++ b/src/cocotb_tools/makefiles/simulators/Makefile.icarus
@@ -65,7 +65,7 @@ $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/sim.vvp $(CUSTOM_SIM_DEPS)
 	$(RM) $(COCOTB_RESULTS_FILE)
 
 	COCOTB_TEST_MODULES=$(call deprecate,MODULE,COCOTB_TEST_MODULES) COCOTB_TESTCASE=$(call deprecate,TESTCASE,COCOTB_TESTCASE) COCOTB_TEST_FILTER=$(COCOTB_TEST_FILTER) COCOTB_TOPLEVEL=$(call deprecate,TOPLEVEL,COCOTB_TOPLEVEL) TOPLEVEL_LANG=$(TOPLEVEL_LANG) \
-        $(SIM_CMD_PREFIX) $(ICARUS_BIN_DIR)/vvp -M $(shell $(PYTHON_BIN) -m cocotb_tools.config --lib-dir) -m $(shell $(PYTHON_BIN) -m cocotb_tools.config --lib-name vpi icarus) $(SIM_ARGS) $(EXTRA_ARGS) $(SIM_BUILD)/sim.vvp $(call deprecate,PLUSARGS,COCOTB_PLUSARGS) $(FST) $(SIM_CMD_SUFFIX)
+        $(SIM_CMD_PREFIX) $(ICARUS_BIN_DIR)/vvp -M $(shell $(PYTHON_BIN) -m cocotb_tools.config --lib-dir) -m $(shell $(PYTHON_BIN) -m cocotb_tools.config --lib-name vpi icarus) $(SIM_ARGS) $(EXTRA_ARGS) $(SIM_BUILD)/sim.vvp $(FST) $(call deprecate,PLUSARGS,COCOTB_PLUSARGS) $(SIM_CMD_SUFFIX)
 
 	$(call check_results)
 
@@ -73,7 +73,7 @@ debug: $(SIM_BUILD)/sim.vvp $(CUSTOM_SIM_DEPS)
 	$(RM) -r $(COCOTB_RESULTS_FILE)
 
 	COCOTB_TEST_MODULES=$(call deprecate,MODULE,COCOTB_TEST_MODULES) COCOTB_TESTCASE=$(call deprecate,TESTCASE,COCOTB_TESTCASE) COCOTB_TEST_FILTER=$(COCOTB_TEST_FILTER) COCOTB_TOPLEVEL=$(call deprecate,TOPLEVEL,COCOTB_TOPLEVEL) TOPLEVEL_LANG=$(TOPLEVEL_LANG) \
-        $(SIM_CMD_PREFIX) gdb --args $(ICARUS_BIN_DIR)/vvp -M $(shell $(PYTHON_BIN) -m cocotb_tools.config --lib-dir) -m $(shell $(PYTHON_BIN) -m cocotb_tools.config --lib-name vpi icarus) $(SIM_BUILD)/sim.vvp $(SIM_ARGS) $(EXTRA_ARGS) $(call deprecate,PLUSARGS,COCOTB_PLUSARGS) $(FST) $(SIM_CMD_SUFFIX)
+        $(SIM_CMD_PREFIX) gdb --args $(ICARUS_BIN_DIR)/vvp -M $(shell $(PYTHON_BIN) -m cocotb_tools.config --lib-dir) -m $(shell $(PYTHON_BIN) -m cocotb_tools.config --lib-name vpi icarus) $(SIM_ARGS) $(EXTRA_ARGS) $(SIM_BUILD)/sim.vvp $(FST) $(call deprecate,PLUSARGS,COCOTB_PLUSARGS) $(SIM_CMD_SUFFIX)
 
 	$(call check_results)
 


dlmiles:
There is a discrepancy with the original Makefile over the placement with gdb and non-gdb versions which are otherwise identical:
```
$(SIM_ARGS) $(EXTRA_ARGS) $(SIM_BUILD)/sim.vvp
$(SIM_BUILD)/sim.vvp $(SIM_ARGS) $(EXTRA_ARGS)
```
So I have modified them both to match (by changing one) to the version that appeared to work locally.
```
$(SIM_ARGS) $(EXTRA_ARGS) $(SIM_BUILD)/sim.vvp $(FST)
$(SIM_BUILD)/sim.vvp $(FST) $(SIM_ARGS) $(EXTRA_ARGS)
```
Let us see if the CI agrees that is ok.

dlmiles:
Hold fire for 24h.  Lets assume yes there is a danger, will test more and research previous cocotb1.

It appears the syntax probably should be like: `harness-wrapper.exe --wrapper-args -m module simulation.obj +simulation+arg +simulation+setting=value`
However it doesn't seem the module follows this convention.  The use of `/usr/local/bin/vvp --help` hints at this style, so maybe all module/simulation arguments should be accepted in the `+` prefix form, to differenciate which side of the simulation object they target.

The point being you can take $SIM_ARGS and filter into 2 lists:
* options with a `-` dash prefix or no prefix and
* options with a plus `+` prefix

Then know if they are for the harness-wrapper.exe or the simulation.

`/usr/local/bin/vvp $(SIM_ARGS_DASH) $(EXTRA_ARGS_DASH) $(SIM_BUILD)/sim.vvp $(FST) $(SIM_ARGS_PLUS) $(EXTRA_ARGS_PLUS)`
with FST=+none or FST=+fst  since this option only works on the right side because the option is for the simulation/module not the wrapper.

ktbarrett:
I have no clue what harness-wrapper is, but options that start with `+` are something that should be reserved for SV plusargs. We do not want to support non-SV-plusargs +options. Also I'm not sure how that effects moving the args around, I was mostly concerned about moving the build object `sim.vvp` to before the build args vs after.

dlmiles:
An update on this.

I use the term `harness-wrapper` to mean whatever the simulator uses as a system process orchestrator.  This would be the superior process that loads all the DSOs and contains the simulation, which is `vvp` with icarus.

While this issue talks of the $FST priority the general use case this issue is concerned with is the major release of cocotb v2 doesn't address the 3 kinds of arguments via the two main mechanism to provide them $SIM_ARGS and $EXTRA_ARGS:
* arguments for the simulator (the infrastructure, in this case icarus, affecting the infrastructure)
* arguments for the simulation (affecting the runtime of the simulation, which has a cross over for icarus features but also verilog)
* plusargs (which are an SV feature, but also used by other simulators for the same kind of purpose to affect simulation, as `+fst` and can also be consumed by the libcocotbvpi_icarus DSO module).


Working through this problem I have come up with using (this contrived example attempts to demonstrate all the features):
```
# Contrived configuration
export FST="--cocotb-plusarg-fst"
export SIM_ARGS="--foo1 --cocotb-arg=plain --cocotb-plusarg+myoption=value1 foo2 +foo3"
export EXTRA_ARGS="--bar4 bar5 --cocotb-arg-myother=value2 +bar6"

# cocotb-config is used as a tool that Makefile can run
SIM_ARGS_SIMULATOR=$(cocotb-config --filter-args-dash-any $FST $SIM_ARGS $EXTRA_ARGS)
SIM_ARGS_SIMULATION=$(cocotb-config --filter-args-plus $FST $SIM_ARGS $EXTRA_ARGS)
echo "SIM_ARGS_SIMULATOR=$SIM_ARGS_SIMULATOR"
SIM_ARGS_SIMULATOR=--foo1 plain foo2 --bar4 bar5 -myother=value2
echo "SIM_ARGS_SIMULATION=$SIM_ARGS_SIMULATION"
SIM_ARGS_SIMULATION=-fst +myoption=value1 +foo3 +bar6

# Note the prefix "--cocotb-plusarg" and "--cocotb-arg" is stripped and naming designed so as to never conflict with a real simulator argument of the same / similar name.
# The Makefile template ends up with:
$(SIM_ARGS_SIMULATOR) $(SIM_BUILD)/sim.vvp $(SIM_ARGS_SIMULATION)
# Resulting in:
--foo1 plain foo2 --bar4 bar5 -myother=value2 $(SIM_BUILD)/sim.vvp -fst +myoption=value1 +foo3 +bar6
```

The important parts are that now the user has the proper control via configuration.  While I may disagree with the order of SIM_ARG and EXTRA_ARG (I think they should be other way around), I understand the need to keep them as-is to be compatible with cocotb v1)

However if you tell me the Makefile method are going away and the new method (whatever that is) can control the arguments in a better way that can provide this kind of control that should work in most use cases.  Then maybe I can just wait and switch to that better solution.  I would be concerned if cocotb v2 is stuck with the current inflexible scheme for the rest of the major release cycle (then in a situation that things can not be changed because it might break things).

The above contrived example is looking like it will not break too many existing things, provide the flexibility I would like to see (by default) as an additive feature.  It should also be compatible with whatever future system exists in this area.

---

Re the current commit.

I am leaning on the side that moving the arg order (the current commit as-is) *will* impact someone, the original Makefile has a discrepancy so I should edit the `gdb` line instead of the `non-gdb` line.

FWIW The use of `-fst` appears to inject some verilog code into the design, so consider it an analogue of this kind of snippet (even if this isn't the exact snippet you get):
```
initial begin
  if ($test$plusarg("-fst")) begin
    $dumpfile("dump.fst");
    $dumpvar(0, tb);
  end
end
```

ktbarrett:
I think we can move the FST higher up and change the debug to position the build object later in the call stack but before the plusargs like the standard call (and also like the Runner). If that sounds better than the current solution we can merge it and make it even better later. 

Trying to release 2.0.1 here soon.

dlmiles:
Ah maybe this is information I have not disclosed, and assume you know.

The use of `-fst` before the build object (and after the module) does not work.  Because the SIMULATOR (`vvp` the infrastructure) does not know how to handle it and it just errors.  The arguments need to be as per `vvp --help`.  Putting it after the build object works because the SIMULATION ('libcocotbvpi_icarus` the simulation runtime context) gets to process it, and does know how to handle the argument.

```
$ COCOTB_TEST_MODULES=test COCOTB_TESTCASE= COCOTB_TEST_FILTER= \
COCOTB_TOPLEVEL=tb TOPLEVEL_LANG=verilog  \
/usr/local/bin/vvp -M $HOME/projects/test/venv/lib/python3.9/site-packages/cocotb/libs \
-m libcocotbvpi_icarus -fst sim_build/rtl/sim.vvp --any-option-is-allowed-here -? --help -help
/usr/local/bin/vvp: invalid option -- 'f'
/usr/local/bin/vvp: invalid option -- 't'

# The -s is taken for "$stop right away." see below:

$ /usr/local/bin/vvp --help
/usr/local/bin/vvp: invalid option -- '-'
Usage: vvp [options] input-file [+plusargs...]
Options:
 -h             Print this help message.
 -i             Interactive mode (unbuffered stdio).
 -l file        Logfile, '-' for <stderr>
 -M path        VPI module directory
 -M -           Clear VPI module path
 -m module      Load vpi module.
 -n             Non-interactive ($stop = $finish).
 -N             Same as -n, but exit code is 1 instead of 0
 -q             Quiet mode (suppress output on MCD bit 0).
 -s             $stop right away.
 -v             Verbose progress messages.
 -V             Print the version information.

```

---

Given it is a cocotb specific option maybe any concern with reserving plusarg namespace for SV can be mitigated with namespace prefix like `+cocotb+fst`.  I just make a point of stating this, but my contrived example does not require it as it uses the syntax `--cocotb-plusarg-fst` to force the plusarg grouping even though the option starts with a dash.  Overall this is much more flexible (if this is a route to go down so it is a reduced number of envvars).

dlmiles:
This is what seems safe (for 2.0.1), to make `gdb` and `non-gdb` match.  There is a slight tweak to move before (lower precedence than) than the deprecated `$COCOTB_PLUSARGS` as that seems correct to me (framework options at a lower precedence than user configured `$COCOTB_PLUSARGS` options and the naming of this envvar suggests it is highly specific to this exact scenario, so should have higher precedence).

Use of a separate `$COCOTB_PLUSARGS` is also a possibility but I see it is deprecated.  I don't know history/reasons for this, if anything icarus did not support SV  so well in the past but is more likely to support SV in the future.  Anyhow.

My initial impression is the naming is a little confusing `SIM_ARGS` and `COCOTB_PLUSARGS` and not `SIM_PLUSARGS` kind of like looking in different places for related things.  The prefix should be same IMHO but it doesn't matter what that is.

dlmiles:
Need to reevaluate commit message wording after change.

dlmiles:
Minor tweak to comment to mention $COCOTB_PLUSARGS even if deprecated.  But otherwise the commit comment is still relevant to the purpose of the commit.

This also re-kicks CI as it appears the third-party simulators were unavailable in the last CI run.

ktbarrett:
Yeah we have maintainers meetings every so often and lots of stuff gets merged at once. Since we have limited licenses for the commercial sims they contend, timeout, and so on.

PR #5057, opened 2025-10-08
Support Python 3.14

ktbarrett:


.codecov.yml
```diff
@@ -20,7 +20,7 @@ codecov:
         #
         # [1] https://docs.codecov.io/docs/merging-reports#how-does-codecov-know-when-to-send-notifications
         # [2] https://docs.codecov.io/docs/notifications#preventing-notifications-until-after-n-builds
-        after_n_builds: 25
+        after_n_builds: 26
```
marlonjames: This will need rebase and update if #5056 comes in first.

.github/generate-envs.py
```diff
@@ -111,14 +111,13 @@
...
         "lang": "vhdl",
         "sim": "nvc",
         "sim-version": "r1.17.1",
         "os": "ubuntu-22.04",
-        "python-version": "3.14-dev",
-        "group": "experimental",
+        "python-version": "3.14",
+        "group": "ci-free",
         "test_nosim": True,
     },
```
imphil, line 95: Keep that part commented out for 3.15, we'll need it again rather soon.

ktbarrett:
Must update cibuildwheel first.

cmarqu:
Should we also support the nogil version 3.14t?

ktbarrett:
Might as well.

setup.py
```diff
@@ -89,7 +89,7 @@ def package_files(directory):
     install_requires=[
         "find_libpython",
     ],
-    python_requires=">=3.6.2",
+    python_requires=">=3.6.2,<3.15",
```
imphil: I wasn't able to quickly figure this out: do pip and alternatives honor this statement (as in: enforce it, and prevent installation on Python 3.15), or is it just a hint (as before)? 

I think the hint behavior is what we want to give users (and us) the freedom to experiment with newer Python versions. Or maybe there's a way to override a potential enforcement by pip on the command line?

ktbarrett: It is enforced, installation fails with both `pip` and `uv pip`. You can override this in `pip` by specifying `--ignore-requires-python`.

noxfile.py
```diff
@@ -28,7 +28,7 @@
 ]
 
 # Version of the cibuildwheel package used to build wheels.
-cibuildwheel_version = "2.20.0"
+cibuildwheel_version = "2.21"
```
ktbarrett:
```suggestion
cibuildwheel_version = "2.20.0"
```

ktbarrett:
Moving to version of cibuildwheel > 2.20 break macOS builds. https://github.com/cocotb/cocotb/actions/runs/18557785181/job/52900310616

ktbarrett:
I'm fairly certain the issue with newer versions of cibuildwheel is the bump in version of delocate from 0.11.0 to 0.12.0 that occurred during the cibuildwheel 0.21.0 release cycle: https://github.com/pypa/cibuildwheel/commit/b1ed7bb3bcf32f107623a93f4ef4409a16e402cc

https://github.com/matthew-brett/delocate/releases/tag/0.12.0 changed the default for macOS builds to use `--sanitize-rpaths`. Looking at the difference in the produced binaries, the 2.0.0 binary for libembed.so contain the following entry, while this PRs does not.

```
Load command 15
          cmd LC_RPATH
      cmdsize 32
         path @loader_path (offset 12)
```

Opened: https://github.com/matthew-brett/delocate/issues/255

setup.py
cmarqu: `uv pip` should actually not enforce it, as per https://docs.astral.sh/uv/pip/compatibility/#requires-python-upper-bounds

noxfile.py
```diff
@@ -406,7 +406,7 @@ def release_build_bdist(session: nox.Session) -> None:
     """Build a binary distribution (wheels) on the current operating system."""
 
     # Pin a version to ensure reproducible builds.
```
cmarqu, line 408:
```suggestion
    # Pin a "Compatible release" to ensure reproducible builds.
```

setup.py
ktbarrett: Perhaps that changed in a recent version. It's been a while now but it did fail on me when I tested locally.

imphil: I agree with the "uv" reasoning of ignoring the upper bound: the dependency solver will just backtrack to a version without this upper bound. Let's remove it.

ktbarrett: The rationale is unconvincing to me. "I typed in `pip install cocotb` without any version number and it didn't install the version I wanted." Well duh... But whatever, I'll remove it.

I recently ran into issues when working on pyuvm because I mistakenly tried to get cocotb 1.6 and Python 3.13 working together, which they very much do not. But I had to waste my time to arrive at that conclusion rather than getting a reasonable warning: `cocotb==1.6.0 requires Python>3.6,<=3.11, but your version of Python is 3.13.`

PR #5061, opened 2025-10-09
logging: performance improvements, edge cases fixes, more powerful prefix_format

gilbertoabram:
These changes come from replacing a custom formatter with the cocotb logging infrastructure.

It should not require any changes in existing code. It adds two features:
- Make some more useful stuff available to a custom prefix format string (ljust/rjust, the highlight sequences for the current log level, and the rendered sim time)
- Allow user-provided indentation by starting a message with `\n`. The idea here was to allow formatting log messages on the next line from the prefix (e.g. via a filter prepending `\n`) without indentation, being more friendly to narrow terminals. Thinking about it now, this is potentially breaking, and it could be implemented with an optional record attribute, lmk (I imagine it's unlikely that anyone would start a log message with `\n` given the way it's formatted at the moment)

On the performance side, one big cause of slowness for me was calling `Path(record.filename).name` with the full/non-reduced output. Machine dependent, but for the workloads I tested it would add about 5us per log call (It's better when called in a tight loop, ~1.5us, so I found testing it in isolation to understate its runtime). I don't think it actually does anything, since that attribute should already just contain the file name (`record.filename == Path(record.pathname).name`) 

The second biggest hit to performance came from removing ansi escape sequences, so here I avoid doing it if possible, and I added a faster heuristic (disabled unless requested explicitly) to determine if there is anything to remove before using the regex, based on whether the escape character is present.

And finally, a few little things like collecting some common code in `format()` and breaking out early in the case of a single log line also saves a little bit (most of the saving from the latter is from not having to determine a custom prefix length via stripping ansi codes, but skipping the generic multiline code also saves a little bit).

With these additions, I see log invocations in my tests going from a median/average of 30.7/40.4 us to 22.6/27.4 us per call.

I also bumped into two edge cases:
- If there is a custom color part in a message logged at a level which has some colorization, we should substitute reverting to the terminal default to instead switching back to the desired color. Unfortunately this is also complicated to do in general without sacrificing performance too much, so here I just check for `ANSI.DEFAULT`, but one could e.g. only switch and revert foreground color, and fool that logic.
- If a comment spans multiple lines and has a non-default background, the background will also be applied to the padding, so I changed it to apply it only to the text coming from the message. Not sure if that was intended, but it did not look right. But if it's not a big deal we could leave this out, simplify things a bit, and probably take slightly less time on colored multi-line output.

### DIFF
--- a/tests/test_cases/test_log_prefix/test_log_prefix.py
+++ b/tests/test_cases/test_log_prefix/test_log_prefix.py
@@ -30,19 +30,17 @@ def test_log_prefix() -> None:
         build_args = ["-v93"]
 
     build_dir = runner.build_dir
-    subdir = build_dir / "custom_prefix"
     runner.build(
         sources=[pwd / "top.sv"],
         hdl_toplevel="top",
         build_args=build_args,
-        build_dir=subdir,
+        build_dir=build_dir / "custom_prefix",
     )
     runner.test(
         test_module="log_prefix_tests",
         hdl_toplevel="top",
         hdl_toplevel_lang=LANG,
         test_args=test_args,
-        test_dir=subdir,
         test_filter="test_log_prefix_custom",
         extra_env={
             "COCOTB_LOG_PREFIX": "{ANSI.YELLOW_FG}abc{ANSI.DEFAULT_FG} {record.levelname} {record.created_sim_time} {record.name[:4]:>10} ",
@@ -50,35 +48,31 @@ def test_log_prefix() -> None:
         },
     )
 
-    subdir = build_dir / "reduced_prefix"
     runner.build(
         sources=[pwd / "top.sv"],
         hdl_toplevel="top",
         build_args=build_args,
-        build_dir=subdir,
+        build_dir=build_dir / "reduced_prefix",
     )
     runner.test(
         test_module="log_prefix_tests",
         hdl_toplevel="top",
         hdl_toplevel_lang=LANG,
         test_args=test_args,
-        test_dir=subdir,
         test_filter="test_log_prefix_default",
     )
 
-    subdir = build_dir / "full_prefix"
     runner.build(
         sources=[pwd / "top.sv"],
         hdl_toplevel="top",
         build_args=build_args,
-        build_dir=subdir,
+        build_dir=build_dir / "full_prefix",
     )
     runner.test(
         test_module="log_prefix_tests",
         hdl_toplevel="top",
         hdl_toplevel_lang=LANG,
         test_args=test_args,
-        test_dir=subdir,
         test_filter="test_log_prefix_default",
         extra_env={"COCOTB_REDUCED_LOG_FMT": "0"},
     )
--- a/tests/test_cases/test_cocotb/test_logging.py
+++ b/tests/test_cases/test_cocotb/test_logging.py
@@ -118,3 +118,28 @@ async def test_ljust_rjust(_: object) -> None:
     assert rjust("01234", 5) == "01234"
     assert ljust("012345", 5) == "..345"
     assert rjust("012345", 5) == "..345"
+
+
+@cocotb.test
+async def test_multiline_indent(_: object) -> None:
+    stripped_formatter = cocotb_logging.SimLogFormatter(strip_ansi=True)
+    with capture_logs(formatter=stripped_formatter) as logs:
+        cocotb.log.info("First line\nSecond line")
+    assert len(logs.msgs) == 1
+    lines = logs.msgs[0].splitlines()
+    assert len(lines) == 2
+    first_index = lines[0].find("First line")
+    second_index = lines[1].find("Second line")
+    assert first_index == second_index
+    assert first_index != -1
+
+    indent_formatter = cocotb_logging.SimLogFormatter(
+        strip_ansi=True, multiline_indent=first_index + 1
+    )
+    with capture_logs(formatter=indent_formatter) as logs:
+        cocotb.log.info("First line\nSecond line")
+    assert len(logs.msgs) == 1
+    lines = logs.msgs[0].splitlines()
+    assert len(lines) == 2
+    assert lines[0].find("First line") == first_index
+    assert lines[1].find("Second line") == first_index + 1


src/cocotb/logging.py
```diff
@@ -284,66 +299,48 @@ def __init__(
             """,
             re.VERBOSE,
         )
+        self._no_ansi_heuristic = not ansi_strip_heuristic
 
     def strip_ansi(self) -> bool:
         return strip_ansi if self._strip_ansi is None else self._strip_ansi
 
+    def strip_escape_patterns(self, string: str) -> str:
+        if self._no_ansi_heuristic or ("\x1b" in string):
```
ktbarrett: I don't think we need to make `_no_ansi_heuristic` a point of customization. If it's faster on average to do it, let's just always do it.

src/cocotb/logging.py
```diff
@@ -363,18 +360,56 @@ def formatExcInfo(self, record: logging.LogRecord) -> str:
...
+            msg = self.strip_escape_patterns(msg)
+        elif highlight_start:
+            # NOTE: this assumes that ANSI.DEFAULT is used to go back to normal,
+            # which is what we do but it can be not true in general.
+            msg = msg.replace(ANSI.DEFAULT, highlight_start)
+
+        # Break the message into lines
+        if highlight_start:
+            # Colorize each line separately, so that if a custom background is set
+            # it won't get applied to the padding used for alignment.
```
ktbarrett: I'm not sure this is a problem?

src/cocotb/logging.py
```diff
@@ -284,66 +299,48 @@ def __init__(
...
-            return prefix, prefix_len
+            prefix = self._prefix_func(
+                record, sim_time_str, highlight_start, highlight_end
+            )
+            return prefix, None
         else:
             prefix = f"{sim_time_str:>11} {highlight_start}{record.levelname:<8}{highlight_end} {self.ljust(record.name, 34)} "
             if not self._reduced_log_fmt:
-                prefix = f"{prefix}{self.rjust(Path(record.filename).name, 20)}:{record.lineno:<4} in {self.ljust(str(record.funcName), 31)} "
+                prefix = f"{prefix}{self.rjust(record.filename, 20)}:{record.lineno:<4} in {self.ljust(str(record.funcName), 31)} "
```
ktbarrett: This is a change as it includes the full path rather than just the basename. That's *probably* okay though since files can have the same name.

src/cocotb/logging.py
```diff
@@ -363,18 +360,56 @@ def formatExcInfo(self, record: logging.LogRecord) -> str:
...
+            lines.append("")
+
+        if len(lines) == 1:
+            # fast path for single line messages
+            return prefix + lines[0]
+
+        if not msg.startswith("\n"):
+            # add padding to each line of message
+            if prefix_len is None:
+                prefix_len = len(self.strip_escape_patterns(prefix))
```
ktbarrett: Why are we doing this rather than correctly returning `prefix_len`? This is wasteful on the common path.

ktbarrett:
Looks like test_log_prefix needs to be updated. Also you'll need to update the documentation to add `ljust`, `rjust`, `highlight_start` and `highlight_end` to the list of things available when evaluating a prefix string.

src/cocotb/logging.py
gilbertoabram: It is faster, but it will not detect all the patterns removed by the regex. It will detect sequences starting with `\x1b[' (now. was just `\x1b` in the original commit), which are used in cocotb's`ANSI` and also in the tests I'm working on, but in general it's not equivalent to just using the regex for removal every time. If that seems fine, I can remove the whole heuristic part, but at that point it makes sense to also simplify the regex to only remove those kinds of escape sequences, and I can probably re-test if that's faster and not warrant the heuristic in the first place.

src/cocotb/logging.py
gilbertoabram: Ack. It does not look right imo, but it's a pretty minor and inconsequential thing, and not doing it makes the code flow better, so I took that away.

src/cocotb/logging.py
gilbertoabram: Does this behave differently in older python versions maybe? From my tests this is just a no-op that takes a while to run. Note that there is both a `record.pathname` and a `record.filename`, with the latter effectively being `Path(record.pathname).name`, and any subsequent `.name` on it does not change it.

src/cocotb/logging.py
gilbertoabram: This is dependent on what gets logged, true, so maybe what I assumed is the common path is common only for me.
The idea was that we only need prefix_len when we need to indent/pad a multiline log line, so I was deferring computing it until needed. Which only buys us something if logging skews towards single lines more often.
If we return `prefix_len` before that part, with a custom prefix we always strip escape codes to compute the length, even if it goes unused. And for the case without custom prefix, `prefix_len` is not `None`, so just checking is very minimally wasteful.
In any case, I reorganized things with the other changes so that prefix_len is always computed when needed in both cases.

gilbertoabram:
It would be ideal to have some way to return the prefix length in a non-breaking fashion, avoiding the stripping.
Maybe there could be an optional integer argument, which if set will just be used as the prefix length, bypassing all that computation, but falling back to stripping if not set.
And since the default widths are constant, we could also have the default formatting be implemented as a specific prefix string, and just not differentiate the two cases.

src/cocotb/logging.py
ktbarrett: I think we want to actually change the escape code regex to only test for ANSI codes that start with ESC. The ones that start with characters > 0x80 are not compatible with unicode so we should ignore them. In fact the current regex will strip regular unicode text...

Well this is turning into another 2.0 bugfix.

src/cocotb/logging.py
ktbarrett: Maybe what I was missing was that `record.filename` is already the basename? If so, then yeah remove it.

src/cocotb/logging.py
ktbarrett: What are you using to render this stuff if setting foreground text on spaces is causing weird output? I'd blame that tool for being bad instead.

src/cocotb/logging.py
ktbarrett: Okay I think I understand what the problem is now. But stripping ANSI control codes isn't sufficient. If users put weird non-printable characters in the prefix, it will screw up the prefix length calculation. But I guess users don't have to use highlighting, so that also screws with the current implementation. ARGH! I wish there was some easy way to know how many columns a piece of text consumed once printed to a terminal.

What I'm saying is, okay I guess this is fine.

src/cocotb/logging.py
ktbarrett: And then after changing the regex we can use the simpler heuristic always.

tests/test_cases/test_log_prefix/log_prefix_tests.py
```diff
@@ -43,5 +43,5 @@ async def test_log_prefix(_: object) -> None:
         logger.info("Test log message")
     assert (
         logs.msgs[0]
-        == f"{ANSI.YELLOW_FG}abc{ANSI.DEFAULT_FG} INFO 0       exam Test log message{ANSI.DEFAULT}"
+        == f"{ANSI.YELLOW_FG}abc{ANSI.DEFAULT_FG} INFO 0       exam Test log message"
```
ktbarrett: This is definitely not ideal. Why is there no reset at the end of the message?

src/cocotb/logging.py
gilbertoabram: So, maybe this is something that is actually intended and it's just a matter of personal taste. It is a super useless corner case which involved experimenting with custom color schemes, but it affects critical errors with the default scheme. When you set a background, a multiline log like
`log.critical("test multiline log message\nline 2\nline 3")`
ends up looking like this in plain xterm:
image
but also in other terminal emulators that I've tested. So the idea there was to highlight each line independently, before applying the padding. In the big scheme of things it seems pretty irrelevant, and it makes the code more complicated, so I removed it, but I can put it back in, up to you.

src/cocotb/logging.py
gilbertoabram: Yup, precisely. with `record.filename` it's not necessary/a no-op (but somehow expensive), it would make sense if we only had `record.pathname`

src/cocotb/logging.py
gilbertoabram: ack. I pushed a change, but I'll need to test it. I wonder if having a regex with a fixed character in the first position will be much faster and make the heuristic entirely unneeded.

src/cocotb/logging.py
gilbertoabram: I pushed a possible approach: adding a prefix length parameter. When it defaults to none we just take the length of the stripped prefix. But it can be an integer, for fastest performance on fixed-length prefixes, or a custom function, if somebody really need it to work with some unicode craziness (e.g. combining characters) or other nonstandard things. Again, not well tested

tests/test_cases/test_log_prefix/log_prefix_tests.py
gilbertoabram: I restored that, I didn't realize it acted as a safety in case of random escapes in the log message, so I was no longer applying `highlight_end` if `highlight_start` was empty

src/cocotb/logging.py
ktbarrett: Ah, its the background text. I forgot about critical. I expected no one actually used that. Yeah we should put it back in.

src/cocotb/logging.py
gilbertoabram: The simpler regex seems to be substantially equivalent, runtime-wise, to just checking if the escape code is contained in the string, so I just removed it entirely.

tests/test_cases/test_log_prefix/log_prefix_tests.py
gilbertoabram: After some tests, what's currently in the PR is a bit different: it only appends `ANSI.DEFAULT` if it finds the escape sequence in the message. I can revert back to always doing it though if that's preferable.

src/cocotb/logging.py
```diff
@@ -254,96 +269,60 @@ def __init__(
...
-                        \x9B
-                    )
-                    [0-?]*  # Parameter bytes
-                    [ -/]*  # Intermediate bytes
-                    [@-~]   # Final byte
-                )
+                \x1B\[  # 7-bit CSI, ESC [
+                [0-?]*  # Parameter bytes
+                [ -/]*  # Intermediate bytes
+                [@-~]   # Final byte
```
ktbarrett: I think we should also handle the 2 byte case. We can just skip the middle case.

src/cocotb/logging.py
```diff
@@ -364,18 +343,52 @@ def formatExcInfo(self, record: logging.LogRecord) -> str:
...
-        prefix, prefix_len = self.formatPrefix(record)
-        msg = self.formatMessage(record)
+        msg = record.getMessage()
+
+        if self.strip_ansi():
+            highlight_start = ""
+            highlight_end = ""
+        else:
+            highlight_start = self.loglevel2colour.get(record.levelno, "")
+            if highlight_start or (cocotb._ANSI._ESCAPE in msg):
```
ktbarrett: Do we really need to scan the message? Just always adding the DEFAULT message seems safer and removes an O(n) search. We always concatenate anyways.

gilbertoabram: Reverted.

src/cocotb/logging.py
gilbertoabram: Done. It still won't work in general for some of those cases that take additional characters (e.g. DCS or OSC), but we can just deal with those if they ever become an issue. As long as the regex starts with a single possible character, it seems that there is no need for the heuristic still.

gilbertoabram:
Given the backport, the performance improvement now is not very dramatic over that: the main causes of slowness were filename/pathname, and a regex with a unique first character. So this is mostly about additional functionality, and both implementing the default logging in the same way as the custom one, as well as providing ways to make the custom logging as fast as the builtin.
Does it make sense to expose the new arguments in `default_config` and document there? Or add documentation to the formatter's `__init__()`.

src/cocotb/logging.py
```diff
@@ -246,6 +259,8 @@ class SimLogFormatter(logging.Formatter):
         "time": time,
         "simtime": cocotb.simtime,
         "ANSI": ANSI,
```
ktbarrett, line 288: Lost the other new args.

src/cocotb/logging.py
```diff
@@ -254,96 +298,86 @@ def __init__(
...
-            re.VERBOSE,
+                """,
+                re.VERBOSE,
+            )
+            self._strip_method = lambda s: ansi_escape_pattern.sub("", s)
+
+        if prefix_format is None:
+            prefix_format = "{sim_time_str:>11} {highlight_start}{record.levelname:<8}{highlight_end} {ljust(record.name, 34)} "
+            if not self._reduced_log_fmt:
+                prefix_format = "{prefix}{rjust(record.filename, 20)}:{record.lineno:<4} in {ljust(str(record.funcName), 31)} "
```
ktbarrett:
```suggestion
                prefix_format = "{prefix_format}{rjust(record.filename, 20)}:{record.lineno:<4} in {ljust(str(record.funcName), 31)} "
```

gilbertoabram: Thanks! I realized it actually just needs to be appended since prefix_format won't exist at the time of formatting. I added a test which covers non-reduced prefix formatting.

src/cocotb/logging.py
```diff
@@ -364,18 +401,49 @@ def formatExcInfo(self, record: logging.LogRecord) -> str:
...
+            level_color_start = ""
+            level_color_end = ""
+        else:
+            level_color_start = self.loglevel2colour.get(record.levelno, "")
+            level_color_end = ANSI.DEFAULT if level_color_start else ""
+
+        prefix = self.formatPrefix(record, level_color_start, level_color_end)
+
+        if self.strip_ansi():
+            output = self._strip_method(f"{prefix}{msg}")
```
ktbarrett: Hmmm, this isn't great. If a user specifies a strip function it should be run agnostically to `strip_ansi`. That should only be followed if we are using the default ANSI code stripping function.

A lot of these knobs interact with other knobs in unintuitive ways. This is one of the many reasons why knobby code is low quality.

ktbarrett (commented):
I'll need to think some more about all the stuff being added here.

There are some good things:
* supporting customized indents
* supporting more features in the prefix string eval
* fixing the `strip_ansi` stripper function
* perf improvements

There are also some questionable things:
* how the knobs interact
* `simtime_ns` seems like it could be a more general feature by leveraging [this](https://github.com/ktbarrett/coconext/blob/70092d09536d4aaef5661dfedb07201970ac57ac/src/coconext/simtime.py#L21). So the prefix string could look like `f"{simtime.ns:>11.2f}ns"`. This prevents having to support N different preformatted strings that we are computing despite potentially not using.

src/cocotb/logging.py
gilbertoabram: removed

src/cocotb/logging.py
```diff
@@ -265,6 +254,15 @@ def _rjust(string: str, chars: int) -> str:
     return string.rjust(chars)
 
 
+# Default simtime formatter
+def _simtime_fmt(record: logging.LogRecord) -> str:
```
ktbarrett:
```suggestion
def _simtime_fmt(record: logging.LogRecord, unit: TimeUnit) -> str:
```

? So users can specify things beyond ns? I like that, that seems generic and we don't have to wait to upstream `SimTime` type.

docs/source/library_reference.rst
```diff
@@ -524,13 +524,18 @@ Log Formatting
     - ``record``: The :class:`~logging.LogRecord` being formatted. This includes the attribute ``created_sim_time``, which is the simulation time in steps.
     - ``time``: The Python :mod:`time` module.
     - ``simtime``: The cocotb :mod:`cocotb.simtime` module.
+    - ``simtime_fmt``: Function which, given a ``LogRecord``, returns the formatted simulation time as ``f"{t:.2f}ns"`` if available, or as ``"-.--ns"`` if not.
```
ktbarrett: This should be updated.

tests/test_cases/test_log_prefix/test_log_prefix.py
```diff
@@ -41,25 +42,34 @@ def test_log_prefix() -> None:
...
     runner.test(
         test_module="log_prefix_tests",
         hdl_toplevel="top",
         hdl_toplevel_lang=LANG,
         test_args=test_args,
         test_filter="test_log_prefix_default",
+        test_dir=pwd,
     )
+    if coverage_file.exists():
+        coverage_file.replace(pwd / ".coverage.cocotb.reduced_prefix")
```
ktbarrett: There's definitely a better solution here, but this will work for now.

ktbarrett:
I think the issue with Questa and Riviera are because the toplevel library is set to the same as the top module. Making them the same confuses them for some reason. I'm guessing running the first test dumps a `top` directory they then think is a build directory. Maybe just changing the toplevel library in the build/run calls would work.

tests/test_cases/test_log_prefix/test_log_prefix.py
```diff
@@ -33,14 +34,46 @@ def test_indexing_warning() -> None:
         sources=[pwd / "top.sv"],
         hdl_toplevel="top",
         build_args=build_args,
+        hdl_library="test",
```
ktbarrett:
```suggestion
        hdl_library="test",
        hdl_toplevel_library="test",
```

ktbarrett:
```
QuestaSim-64 vlog 2025.2 Compiler 2025.05 May 31 2025
Start time: 18:44:27 on Oct 22,2025
vlog -incr -work top -sv "+acc" /opt/actions-runner/_work/cocotb/cocotb/tests/test_cases/test_log_prefix/tb.sv
** Note: (vlog-220) '/nas/fossif/tools/mentor/questa/2025.2/questasim/linux_x86_64/../modelsim.ini' is used as the ini file.
-- Compiling module tb

Top level modules:
        tb
End time: 18:44:27 on Oct 22,2025, Elapsed time: 0:00:00
Errors: 0, Warnings: 0
Reading pref.tcl

# 2025.2

# ** Note: (vsim-220) '/nas/fossif/tools/mentor/questa/2025.2/questasim/linux_x86_64/../modelsim.ini' is used as the ini file.
# vsim -c -onfinish exit -pli "/opt/actions-runner/_work/cocotb/cocotb/.nox/dev_test_sim-sim-questa-toplevel_lang-verilog-gpi_interface-vpi/lib/python3.9/site-packages/cocotb/libs/libcocotbvpi_modelsim.so" -t ps top.tb -do "run -all; quit"
# Start time: 18:44:28 on Oct 22,2025
# ** Error (suppressible): (vsim-19) Failed to access library 'top' at "top".
# No such file or directory. (errno = ENOENT)
# Error loading design
# End time: 18:44:28 on Oct 22,2025, Elapsed time: 0:00:00
# Errors: 1, Warnings: 0
```

ktbarrett:
Okay, we are going to need to fix the Riviera and Questa runners I think. I have no clue why they are okay with just one test, but I think adding the library name to the toplevel when the toplevel is verilog is unnecessary and incorrect. Perhaps it's only important when doing VHDL or mixed language?

This is why cocotb doesn't want to be a build system, the stuff is maddening and steals effort for stuff that actually matters.

gilbertoabram:
> Okay, we are going to need to fix the Riviera and Questa runners I think. I have no clue why they are okay with just one test, but I think adding the library name to the toplevel when the toplevel is verilog is unnecessary and incorrect. Perhaps it's only important when doing VHDL or mixed language?

I think the issue might be that the test directory must match the build directory. Unfortunately I don't have access to either tool, so this is just guesswork from the CI results (sorry for all the commits). I was able to finally get things to work by building for each test in a separate directory, which is suboptimal but hopefully not too big of an issue, as in this specific case this only adds two builds of a trivial module.

If this works well enough for this purpose, I'd prefer someone that is able to test directly with those tools to take care of it in a separate PR.

ktbarrett (approved):

PR #5071, opened 2025-10-10
Fix VPI callback memory leak by cleaning up one-shot callbacks after firing

skulis:
# Summary

This PR attempts to fix a memory leak in VPI callback handlers where one-shot callbacks were never being cleaned up after firing on non-Verilator simulators, causing unbounded memory growth proportional to simulation time.

# Problem

The base class `VpiCbHdl::run()` method only performed cleanup when compiled for Verilator (inside `#ifdef VERILATOR` guards). This meant that on Icarus, Xcelium, and potentially other VPI simulators, one-shot callbacks were never removed after firing:

  1. Callback allocated via `vpi_register_cb()`
  2. Callback fires once
  3. Callback handle left registered forever (never calling `vpi_remove_cb()` or `delete this`)

  Affected callback types: `VpiTimedCbHdl`, `VpiReadWriteCbHdl`, `VpiReadOnlyCbHdl`, `VpiNextPhaseCbHdl`.

  # Solution

Remove the `#ifdef VERILATOR` guards from `VpiCbHdl::run()` in `VpiCbHdl.cpp` to enable proper cleanup for all VPI simulators. As far as I can tell, three callback types that might need special handling (`VpiValueCbHdl`,  `VpiStartupCbHdl`, `VpiShutdownCbHdl`) already override `run()` with their own implementations.

# Testing

## Test environments

  - Cocotb version: 2.0.0
  - Simulators tested: Icarus Verilog 12.0, Xcelium 24.03.004
  - OS: Linux (Manjaro and Red Hat Enterprise Linux release 8.10)
  - Python: 3.13.7 and 3.11.13

## Test case

  ```python
  import cocotb

  @cocotb.test()
  async def memory_test(dut):
      # Start a clock
      clock_task = cocotb.start_soon(Clock(dut.clk, 25, unit="ns").start())
      for i in range(5):
          await Timer(1, unit="ms") 
          # Memory profiling showed growth here
```

## After fix
  - Memory stable throughout simulation
  - No leaked callbacks detected by heaptrack profiling

# Questions for Maintainers

The original `#ifdef VERILATOR` guard suggests there may have been a historical reason for skipping cleanup on other simulators. However, testing on Icarus and Xcelium shows they handle `vpi_remove_cb()`  correctly for one-shot callbacks. Do you know if there are any known issues with callback removal on specific simulators that this might affect?

  An alternative approach would be overriding `run()` in each affected callback class (like `VpiStartupCbHdl` does) rather than fixing the base class. This would be potentially (?) safer but creates code duplication across 4 classes. If there are compatibility concerns, I can implement the per-class override approach instead. Please advise.

### DIFF
--- a/src/cocotb/share/lib/vpi/VpiCbHdl.cpp
+++ b/src/cocotb/share/lib/vpi/VpiCbHdl.cpp
@@ -152,6 +152,10 @@ int VpiCbHdl::run() {
     else {
         delete this;
     }
+#else
+    // For other simulators: VPI spec says one-shot callbacks auto-cleanup
+    // their handle after firing. We just need to delete the C++ object.
+    delete this;
 #endif
 
     return res;


ktbarrett:
IIRC according to the VPI you shouldn't need to call this function after firing, its only for before it fires. The reason Verilator was different because these aren't one-shot callbacks in that simulator for some reason.

Let's see if the simulators are happy with this.

ktbarrett:
Questa and GHDL are pretty clearly not happy with the double-free. Are you sure this is required by the spec? Otherwise just because the test is passing doesn't mean it's not just accidentally working with a use-after-free.

skulis:
I'm sorry for the confusion, I misunderstood the VPI callback lifecycle. I've updated the fix to only delete the C++ object without calling `vpi_remove_cb()` on non-Verilator simulators. I tested successfully on both Icarus and GHDL. The memory leak is resolved and no crashes occur.

Any chance we could run the full regression again? I won't be able to test on other simulators until sometime next week.

PR #5074, opened 2025-10-13
Add support for x-epic Galaxsim simulator

chao61:
Galaxsim is a simulator by x-epic, this pr support Galaxsim in the makefile flow.

### DIFF
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -845,3 +845,18 @@ def get_ext():
         ext.append(dsim_vpi_ext)
 
     return ext
+
+    #
+    # Galaxsim
+    #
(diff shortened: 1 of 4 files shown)


imphil:
Thanks for your PR. Looks good in general, but I unfortunately have to ask you to go a bit beyond the current scope.

* Can you also add a section about the simulator to https://github.com/cocotb/cocotb/blob/master/docs/source/simulator_support.rst?
* Can you also add support to the Runner (https://github.com/cocotb/cocotb/blob/master/src/cocotb_tools/runner.py)?


How well does cocotb work with this simulator? Does it pass the test suite? Is there a way for us to run the simulator?

PR #5076, opened 2025-10-13
add separate cwd argument from build_dir for runner.build

rabidaudio:
Hello, I'm new to HDL development but I'm finding that `cocotb` is much preferable for writing testbenches than SystemVerilog, so I greatly appreciate the work!

I'm using Python test runners. Building tests for the simulator generates a lot of output files that end up in the `build_dir` (e.g. for Icarus:`*.fst` files for all modules, `cmds.f`, `cocotb_iverilog_dump.v`, xml results files, etc). Ideally these would be put in a directory separate from my HDL source code. However, setting the `build_dir` to a dedicated `my_proj/build` folder will cause the compiler to fail, as import paths (e.g. `` `include`` macros and `$readmem` paths) are no longer correct as they are relative to the project directory. This is because Cocotb's runner uses `build_dir` as both the directory that outputs should be written to and the working directory for the build step.

This is a pretty simple change that adds an additional, optional argument to `runner.build` called `cwd` which is the directory in which the build step is invoked. This defaults to `build_dir` if unspecified, retaining backwards compatibility.

### DIFF
--- a/docs/source/newsfragments/5076.feature.rst
+++ b/docs/source/newsfragments/5076.feature.rst
@@ -1,0 +1,1 @@
+Added ``cwd`` argument to :meth:`.Runner.build`.


rabidaudio:
> Seems reasonable. Can you add a newsfragment for the new feature? You'll find documentation at `docs/source/newsfragments/README.rst`.

Done!

ktbarrett:
Will merge tomorrow if there are no comments.

PR #5078, opened 2025-10-14
Make `RoundMode` and `TimeUnit` type aliases public

### LINKED ISSUE #5067: Move contents of `cocotb._typing` into public modules
These can be used by users.

ktbarrett:
Closes #5067. Making these public has utility in downstream libraries that build off the APIs that use these types in a maintenance-free way.


src/cocotb/simtime.py
```diff
@@ -17,28 +17,48 @@
...
+    "RoundMode",
+    "TimeUnit",
     "convert",
     "get_sim_time",
     "time_precision",
 )
 
+RoundMode: TypeAlias = Literal["error", "round", "ceil", "floor"]
+"""
+How to handle non-integral step values when quantizing to simulator time steps.
```
cmarqu, line 34:
```suggestion
How cocotb handles non-integral step values when quantizing to simulator time steps.
```

src/cocotb/simtime.py
```diff
@@ -17,28 +17,48 @@
...
+
+One of ``'error'``, ``'round'``, ``'ceil'``, or ``'floor'``.
+
+When *round_mode* is ``"error"``, a :exc:`ValueError` is thrown if the value cannot
```
cmarqu, line 38: Should we rewrite the description of these types so that it also works from the POV of a 3rd party library, e.g. like "cocotb throws"?

ktbarrett: Just for these, or for all documentation? IMO saying "cocotb does X" is kind of redundant when reading the docs on cocotb. Instead of "cocotb throws X..." I'd prefer impersonal perspective: "X is thrown..."

cmarqu: I was thinking about what docs/tooltip a user would get when importing the type into their own code and whether it was as clear as possible, but I can't try right now while on a phone.
Let's get this in and tweak later if useful.

PR #5091, opened 2025-10-18
Backport #5082

ktbarrett:
Backport #5082.


imphil (approved):
Not a 100% candidate for a patch release, but I'm ok with it for 2.0.1 as the "first real" 2.0 release.

ktbarrett:
I figured it'd be better to get more behavioral breaks in now as part of "the 2.0 we really wanted" release. One band-aid rip rather than many.

PR #5092, opened 2025-10-19
Merge C++ libraries

### LINKED ISSUE #2312: Merge cocotb libraries?

ktbarrett:
Closes #2312.

### DIFF
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -62,9 +62,9 @@ def create_sxs_assembly_manifest(
      - https://docs.microsoft.com/en-us/windows/win32/sbscs/using-side-by-side-assemblies
 
     Args:
-        name: The name of the assembly for which the manifest is generated, e.g. ``libcocotbutils``.
-        filename: The filename of the library, e.g. ``libcocotbutils.dll``.
-        libraries: A list of names of dependent manifests, e.g. ``["libgpilog"]``.
+        name: The name of the assembly for which the manifest is generated, e.g. ``libgpi``.
+        filename: The filename of the library, e.g. ``libgpi.dll``.
+        libraries: A list of names of dependent manifests, e.g. ``["libgpi"]``.
     """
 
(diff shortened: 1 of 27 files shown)


ktbarrett:
Seems like the names are being messed up, its expecting `lib` in the name, but they are being stripped out rn.

EDIT: Figured it out. We were specifically checking for `cocotb.dll` in `config.py` to know if we should add `lib` as a prefix when calculating extension names. This seems wrong. Maybe we should always compile with the `lib` prefix.

ktbarrett:
We need the whole RPATH like thing on Windows to point "up" to the simulator module. This isn't possible using `probing privatePath`. Indirectly loading the simulator module is the only way.

Also the whole side-by-side assembly thing is jank. Leave it to Windows to come up with unnecessarily complicated bs.

There has got to be a better way than this and RPATH on *nix. If we wanted to rename/move the extension module, we'd basically be SoL.

GPI implementations depend upon the GPI and the PyGPI depends upon the GPI. If we merge the implementations into the GPI we resolve the issue of needing to link them. If we load into the GPI, we can then load into the PyGPI indirectly. And if we load into Python and import the PyGPI, we can use something like `os.add_dll_directory` to ensure the GPI load works. That would be a bigger change than I was anticipating.

ktbarrett:
The embed library changes were merged in #5101. I'm punting on RPATH removal for now.

marlonjames:
Interesting commit message...

ktbarrett:
I rewrote the commit message.

PR #5100, opened 2025-10-23
CI: Add test for VHDL integer signal value change trigger

marlonjames:
I think I am seeing an issue with RivieraPRO not firing a `ValueChange` trigger on an integer signal, so adding a test to CI to check it.


marlonjames:
It's passing here, so i'll have to chase it down locally... Anyway i'll merge this

Issue #4585, opened 2025-03-27
Refactor GPI/PyGPI/cocotb shutdown


ktbarrett:
The GPI, PyGPI, and cocotb's shutdown is rather erratic. We are currently battling issues with Questa's VHPI and NVC and modern Python's mishandling of shutdown causing the CI to fail. This has been a long time coming...

## Premature simulator shutdown

This occurs when the simulator decides it's "done." Likely due to a call to `$finish()` or similar, or because the test has left it with no events to service.

### How it works now

The startup process registers `gpi_embed_end` to be called when a shutdown event occurs. When the test signals the end of simulation the shutdown callback is called. `gpi_embed_end` just sets a flag to prevent multiple shutdown attempts, then runs `embed_sim_event`. This is hardcoded to run PyGPI's `_embed_sim_event`.

`_embed_sim_event` calls into Python. cocotb registers `cocotb._init._sim_event` to be run when this gets called. This callback shuts down the regression by failing the remaining tests with `SimFailure` and then runs cocotb package shutdown callbacks. Among the functions called is `gpi_sim_end` which attempts to call `vpi_finish` or similar and deregister the shutdown callback.

### How it should work

While calling `gpi_embed_end` is *correct*, the cocotb callback calling `gpi_sim_end` is *not* correct. The issue is that this case and the following cases, which need calls to `gpi_sim_end`, all share the same code path.

The GPI implementation entry point should *not* register a shutdown callback to run `gpi_embed_end`, but instead allow cocotb to register a separate function for this case by calling into a PyGPI wrapper around a new `gpi_register_shutdown_callback` function. This callback should *not* call `gpi_sim_end` and should fail any outstanding tests and run the cocotb package shutdown callbacks.

## cocotb regression finishes

If all tests in a regression finish then cocotb asks the simulator to shut down for a graceful exit.

### How it works now

After a cocotb regression finishes it runs the cocotb shutdown callbacks then calls `gpi_sim_end` to request the simulator to shut down.
`gpi_sim_end` calls the GPI implementation's `sim_end` which calls VPI/VHPI functions to request simulator shutdown, but also deregisters the shutdown callback. The shutdown actions must be called immediately because there's no chance for them to be called again.

### How it should work

[...]

Issue #4584, opened 2025-03-26
cocotb libraries startup refactor


ktbarrett:
## How it works now

When the VPI or VHPI entry point are loaded, a startup and shutdown callback are registered which call `gpi_embed_init` and `gpi_embed_end`. `gpi_embed_init` calls `embed_sim_init` and not much else (besides calling `gpi_embed_end` if there's a failure and arguable that's BAD. You don't know if the user properly initialized, calling finalize may blow up). Then it calls into the GPI's entry point.

The GPI's entry point, configures the logger, loads other GPI implementations via `GPI_EXTRA` and then calls `embed_init_python`.

The "embed" library's `embed_init_python` is a hardcoded reference to load both `libpython` and to the PyGPI's `_embed_init_python` method.

The PyGPI's `_embed_init_python` loads the Python interpreter and returns.

Then in the startup callback, `gpi_embed_init` calls into `_embed_sim_init` which *then* loads into Python, initializes the cocotb module, and starts the regression immediately. This means the initialization functionality is on the stack when the first test starts.

## How it should work

The VPI and VHPI entry points should *not* register startup and shutdown callbacks. These callbacks aren't doing anything for either the GPI implementation or the GPI library that can't be otherwise handled by the implementation or the user.

The GPI implementation entry point should load into the GPI entry point which should load `GPI_EXTRA` (or ideally we only need one impl library), and then `GPI_USERS`. Loading libpython before loading the PyGPI will be implemented as a `GPI_USERS` rather than hardcoding libpython into the GPI. The PyGPI entry point will be another `GPI_USERS`.

The PyGPI entry point will take the argc and argv, load the Python interpreter, then immediately load into Python via `PYGPI_USERS`, initialize the cocotb module and start the regression. The first test will be registered to start when the startup callback fires rather than run immediately. There will need to be new GPI and PyGPI interfaces to support registering on the startup and shutdown callbacks.

marlonjames:
In general I like the new approach.

We may run into issues with `_process_packages` and `_setup_root_handle` running before the startup callback, as they interact with the hierarchy. If that happens, I think we can still split the initialization from running the first test, either by registering two startup callbacks, or just returning all the way down through the GPI and then calling the GPI-registered startup callback to run the first test before returning to the simulator.

ktbarrett:
[...]

Issue #2271, opened 2020-12-08
Supporting PyPy


ktbarrett:
To perhaps improve performance slightly within Python we could provide the ability to run with PyPy over the CPython interpreter. PyPy's [embedding process](https://doc.pypy.org/en/improve-docs/embedding.html) will require us to radically simplify our C API code. But we can expect [~4.2x increase in performance](https://www.pypy.org/) from the Python part of the testbench.

ktbarrett:
This will also require us to replace the simulator module with a CFFI or ctypes interface. This probably best belongs in an extension once we have add the capability to the GPI to load into something other than the embed library (#1994).

marlonjames:
When mature enough, HPy would allow an extension that supports both PyPy and CPython.
https://hpyproject.org/

ktbarrett:
However you would lose the speed benefits of using an FFI. `cffi` is also usable with both PyPy and CPython. If we can simplify the embedding process in `gpi_embed` we can easily support both PyPy and CPython with an `ifdef`.

PR #5101, opened 2025-10-26
Remove the `embed` library, support multiple GPI users

### LINKED ISSUE #1101: Improve scheduler error messages
This is an issue for discussion how error messages can be improved, continuation from gitter:

Doing this:
```
@cocotb.test()
def some_test(dut):
    # cocotb.fork(Clock(dut.clk, 800000, units='ns').start())
    yield RisingEdge(dut.clk)
```
gives

> Failing test at simulator request before test run completion: Simulator shutdown prematurely

Can this error message be improved? I've just spend some time debugging a tb only to find out I forgot to create one of my clocks and the message wasn't very helpful ; )

Responses so far:
@garmin-mjames 
>  In this specific case we could add something to the scheduler to list the Triggers in _trigger2coros along with the coroutines waiting on them for debugging

@eric-wieser 
> That's not a bad idea
Slightly more advanced would be to fire all remaining triggers with an exception
Which would allow tests to do cleanup like close open files

### LINKED ISSUE #1994: Support multiple GPI users

ktbarrett:
Closes #1101. Closes #1994. xref #4584. xref #4585. xref #2271. xref #5092.

Step 1 of merging the C++ libraries. To be able to merge the PyGPI into a single library we needed to be able to load the PyGPI library via an absolute path that was dynamically supplied. The reason for that is because `libcocotb` will get merged into `simulator.so` which is one directory up. We can't use Window's equivalent to RPATH here because is doesn't support "relative up" directories.

This removes a bunch of Windows-specific code and also allows us to support multiple GPI users.

A follow-on will merge the utility of LIBPYTHON_LOC into GPI_USERS.


src/cocotb/regression.py
```diff
@@ -885,7 +885,15 @@ def _log_test_summary(self) -> None:
...
-    def _fail_simulation(self, msg: str) -> None:
-        self._sim_failure = Error(SimFailure(msg))
+    def _on_sim_end(self) -> None:
+        # We assume if we get this, the simulation ended unexpectedly due to an assertion failure,
+        # or due to an end of events from the simulator.
+        self._sim_failure = Error(
+            SimFailure(
+                "Simulator shut down prematurely. "
+                "This could be due to an assertion failure or a call to an exit routine in the HDL, "
+                "or due to no events needing to be processed (is your clock running?)."
```
ktbarrett:
```suggestion
                "or due to the simulator running out of events to process (is your clock running?)."
```

docs/source/library_reference_c.rst
```diff
@@ -12,6 +12,32 @@ The interaction between cocotb's Python and GPI is via a Python extension module
...
+    Each element of the list contains a path to a library to load.
+    These paths can be full paths (e.g. ``/usr/local/lib/libstuff.so``), in which case the exact library will be loaded,
+    or the basename (e.g. ``libstuff.so``), in which case your operating system's dynamic library lookup will be used.
+
+    Optionally, after the path in each element, a function in that library to call can be specified by name
+    by suffixing the path with a ``;`` character followed by the function name.
+
+    For example:
+
+    * ``GPI_USERS=/usr/local/lib/libstuff.so;libotherstuff.so,entry_func``
```
cmarqu, line 30:
```suggestion
    * ``GPI_USERS=/usr/local/lib/libstuff.so,libotherstuff.so;entry_func``
```
according to the doc above?

src/cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -200,8 +271,22 @@ void gpi_entry_point() {
         gpi_load_libs(to_load);
     }
 
+    // preload python library
```
cmarqu:
```suggestion
    // preload Python library
```

src/cocotb/regression.py
```diff
@@ -885,7 +885,15 @@ def _log_test_summary(self) -> None:
...
         self.log.info(summary)
 
-    def _fail_simulation(self, msg: str) -> None:
-        self._sim_failure = Error(SimFailure(msg))
+    def _on_sim_end(self) -> None:
+        # We assume if we get this, the simulation ended unexpectedly due to an assertion failure,
+        # or due to an end of events from the simulator.
+        self._sim_failure = Error(
+            SimFailure(
+                "Simulator shut down prematurely. "
```
```suggestion
                "cocotb should be responsible to end the simulation, but the simulator shut down prematurely. "
```

src/cocotb_tools/config.py
```diff
@@ -93,6 +93,13 @@ def _help_vars_text() -> str:
     return helpmsg
 
 
+def pygpi_entry_point() -> str:
+    for possible_name in ("libcocotb.so", "cocotb.dll", "libcocotb.dll"):
+        if (possible_lib_path := (libs_dir / possible_name)).exists():
+            return f"{possible_lib_path},initialize"
+    raise RuntimeError("No libcocotb found")
```
cmarqu, line 100:
```suggestion
    raise RuntimeError(f'No libcocotb found - looked for either of "libcocotb.so", "cocotb.dll", "libcocotb.dll" in {libs_dir}')
```
?

ktbarrett: I don't think that adds useful information for debugging, LLM or not. Maybe suggesting reinstalling, but ultimately if anyone sees that error its going to end up a Github issue because something is seriously wrong.

docs/source/library_reference_c.rst
ktbarrett: Need to update docs instead. I changed delimiters.

src/cocotb/share/include/gpi.h
```diff
@@ -429,6 +429,33 @@ gpi_register_nexttime_callback(int (*gpi_function)(void *), void *gpi_cb_data);
 GPI_EXPORT gpi_cb_hdl
 gpi_register_readwrite_callback(int (*gpi_function)(void *), void *gpi_cb_data);
 
+/** Register a callback to run at the start of simulation time.
+ *
+ * @param cb        Callback function pointer.
+ * @param cb_data   Pointer to user data to be passed to callback function.
+ * @return          Handle to callback object.
+ */
+GPI_EXPORT int gpi_register_start_of_sim_time_callback(
```
marlonjames, line 449: `typedef`s for the callback functions would be helpful here to describe the arguments, like in `gpi_logging.h`

src/cocotb/regression.py
ktbarrett: I might choose different wording but 👍

ktbarrett:
Well the last two changes didn't seem to help Xcelium get any farther. Guess I'll split those out to ensure there is no interplay.

ktbarrett:
So Xcelium is failing because we load the VPI during the build and `_set_env` is set only during test.

I think I'm going to "fix" Xcelium before this PR comes in.

src/cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -97,27 +93,38 @@ int gpi_register_impl(GpiImplInterface *func_tbl) {
 
 bool gpi_has_registered_impl() { return registered_impls.size() > 0; }
 
-void gpi_embed_init(int argc, char const *const *argv) {
-    if (embed_sim_init(argc, argv)) {
-        gpi_embed_end();
+void gpi_start_of_sim_time(int argc, char const *const *argv) {
+    for (auto it = start_of_sim_time_cbs.rbegin();
```
marlonjames: Is there a specific reason to do these in reverse registration order?

ktbarrett: For start and end of sim, no, I suppose not. For cleanup, reverse is the safest as there might be order only dependencies between users.

ktbarrett: Actually I take that back. I think it holds for any.

ktbarrett: The idea is that you might have a user X and then a wrapper around X: Y. Y must come after X in the load list GPI_USERS. And the unload ("finalize")  must be done in reverse order.

For sim start and end, I suppose X should come before Y.

src/cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -161,6 +166,70 @@ static void gpi_load_libs(std::vector<std::string> to_load) {
...
+            lib_name = std::move(user);
+        } else {
+            lib_name = user.substr(0, split_idx);
+            func_name = user.substr(split_idx + 1, std::string::npos);
+        }
+
+        void *lib_handle = utils_dyn_open(lib_name.c_str());
+        if (!lib_handle) {
+            LOG_ERROR("Error loading library '%s'", lib_name.c_str());
+            gpi_finish();
```
marlonjames, line 202: `gpi_finalize()` won't be called to run the finalize callbacks for the `GPI_USERS` that have already executed their entry functions.
That relies on `gpi_to_simulator()` which is only called when in a callback.

ktbarrett: Yeah this stuff is all very less than ideal. I'll add it to the backlog of stuff to fix. The more you try to fix the exponentially larger the diff becomes until you've rewritten everything. Then good luck getting it reviewed.

ktbarrett:
@marlonjames I updated #4584 with follow ups.

PR #5107, opened 2025-10-28
Fix some issues with Xcelium

ktbarrett:
The test_log_prefix test is Verilog only, but the code was adding the `-v93` (VHDL-93) flag unnecessarily. I presumed it was causing issues, but it wasn't. Felt I should remove it anyways.

Additionally, the VPI is being loaded during build unnecessarily for Xcelium, the Makefile doesn't do this.


ktbarrett:
With #5160 in we could revisit this.

cmarqu (approved):

PR #5112, opened 2025-10-30
tests: Test that multiple waits on First will catch coincident triggers

marlonjames:
If two signals change value at the same time step, we should be able to get both of them if we wait on `First()` in a loop.
We wait on `First()` the second time before returning from the first callback, and should get the second one.

This probably won't work currently on some simulators,
as we register a new simulator value change callback rather than keeping a list of GPI user callbacks associated with a single simulator callback.


ktbarrett:
If you add some xfails we can get this merged and then work on supporting the functionality.

Issue #5113, opened 2025-10-31
Support SAIF trace generation in Verilator


martin-lueker:
In addition to VCD and FST Verilator (as of version 5.041, devel) now supports SAIF tracing for purposes of power estimation.
- Feature [Pull request](https://github.com/verilator/verilator/pull/5812)
- Discussion of [Workflows for Power analysis](https://github.com/antmicro/verilog-power-analysis-workflows)

To support this in cocotb, some simple changes are required in `src/cocotb/share/lib/verilator/verilator.cpp`.  As demonstrated in [this commit](https://github.com/martin-lueker/cocotb/commit/fbcd3a796fd233fbce1b983031b1db400e58d587).

martin-lueker:
Closed with PR #5114

PR #5114, opened 2025-10-31
Support SAIF tracing in Verilator

martin-lueker:
Incorporates the `VerilatedSaifC` trace type in Verilator 5.041. This is important for generating useful power estimates in synthesis.

Addresses #5113


ktbarrett:
Are you planning on adding support to the Makefiles or the Python runner for this feature? Odd numbers on Verilator are unreleased versions, so there isn't any way to test this feature yet.

martin-lueker:
> Are you planning on adding support to the Makefiles or the Python runner for this feature? Odd numbers on Verilator are unreleased versions, so there isn't any way to test this feature yet.

Fortunately 5.042 released this weekend so I'll test with that and update as necessary.

Meanwhile, I'm using the Makefile flow for testing on my side, and for that it seems that the only cocotb-side reference to the `--trace` arguments comes in a check to `VERILATOR_TRACE`: https://github.com/cocotb/cocotb/blob/3cfca15f0d5f3ba05a4de4c8b278f082a8c18e88/src/cocotb_tools/makefiles/simulators/Makefile.verilator#L44-L47

The default Makefile doesn't seem to include any mechanism for specifying the _format_ of the output.  Though the documentation recommends adding `--trace-<FORMAT>` (in this case: `--trace-saif`) to `EXTRA_ARGS`.  That said, what are your recommendations for the changes to `Makefile.verilator`, if any?

I can also take a look at the runner to see if there are any format defaults that can be changed there.  Your recommendations here would be most welcome!

ktbarrett:
The CI was updated to include Verilator v5.042, you should be able to test it now. As for Makefile and Runner support, we haven't done anything with adding support for changing formats. Maybe just add a TODO to the Makefile and Verilator Runner to add support for this.

I think what I really need is a test which add `--trace-saif` to `COMPILE_ARGS` or similar to get it working. Then check out `docs/source/newsfragments/README` on how to add a newsfragment for the new feature.

maartenboersma:
Thanks for the work. I tried the commit and it works as I expect.
I'd like to get this PR merged and wonder if I can help. @martin-lueker - do you plan to add a testcase and the newsfragment as suggested by @ktbarrett?

docs/source/simulator_support.rst
```diff
@@ -140,6 +140,17 @@ To enable FST tracing, add ``--trace-fst`` to :make:var:`EXTRA_ARGS` as shown be
 
 The resulting file will be :file:`dump.fst` and can be opened by ``gtkwave dump.fst``.
 
+In addition to waveforms, Verilator (versions 5.041 and higher) can generate SAIF activity files for use in power estimation.
```
ktbarrett: Odd versions of Verilator are just dev versions.
```suggestion
In addition to waveforms, Verilator (versions 5.042 and higher) can generate SAIF activity files for use in power estimation.
```

martin-lueker:
> Thanks for the work. I tried the commit and it works as I expect. I'd like to get this PR merged and wonder if I can help. @martin-lueker - do you plan to add a testcase and the newsfragment as suggested by @ktbarrett?

Hi @maartenboersma,
  Thanks for checking it out.  It would be helpful if an cocotb member could at least talk me though adding an approved test case here.  It's on my TO-DO list to dig into it, but I've been unfortunately been more tied up wit the actual design project that motivated this change in the first place, so I'm really sorry for the delay.  I'd be happy to jump on slack (or similar) to discuss this last step.

ktbarrett:
@martin-lueker There is the [gitter](https://matrix.to/#/#cocotb_Lobby:gitter.im).

As for how to test, I think we can add a new test directory to `tests/test_cases/` and set up a simple test (copy one of the other tests to start). The Makefile should have an ifdef to skip the test if `SIM` isn't set to `verilator`. It should also set the SAIF flag with `COMPILE_ARGS` as I mentioned.

The only thing the test needs to do is test that the SAIF file is created. You can use a recursive make call to run the simulation then do something like `test -f file.saif`. I don't want to check the contents of the file, that's more Verilator's issue. cocotb only cares we can pass the flag and the file is produced.

In the `simulator_support.rst` file add a couple sentences to mention this is supported and how to pass the option using `COMPILE_ARGS`. Then make a newsfragment as I mentioned before (there's a README on how to do that).

martin-lueker:
Thanks @ktbarrett.  I'm suspect I may need to consult with people on the gitter for passing CI.

In looking at making something  `test/test_cases` I notice that there was already a test a `test_wave_dump` test in `tests/pytest/test_waves.py`.     So I created an analogous `test_saif_dump` test in the same file, just for Verilator.

This brings up 2 questions: 
1. This test will fail for versions of verilator < 5.042.  Is there a convenient way to check the version in pytest (and skip), or enforce the correct version inside the test?
2. Verilator is not the only simulator obviously that can make SAIF files.  I contemplated putting a comment that this would be the right place to expand this test functionality for other simulators, though would it be perhaps more appropriate to make an issue to ask others to discuss the options? Or just leave this alone for now? WDYT? 



> @martin-lueker There is the [gitter](https://matrix.to/#/#cocotb_Lobby:gitter.im).
> 
> As for how to test, I think we can add a new test directory to `tests/test_cases/` and set up a simple test (copy one of the other tests to start). The Makefile should have an ifdef to skip the test if `SIM` isn't set to `verilator`. It should also set the SAIF flag with `COMPILE_ARGS` as I mentioned.
> 
> The only thing the test needs to do is test that the SAIF file is created. You can use a recursive make call to run the simulation then do something like `test -f file.saif`. I don't want to check the contents of the file, that's more Verilator's issue. cocotb only cares we can pass the flag and the file is produced.
> 
> In the `simulator_support.rst` file add a couple sentences to mention this is supported and how to pass the option using `COMPILE_ARGS`. Then make a newsfragment as I mentioned before (there's a README on how to do that).

ktbarrett:
There is `cocotb_tools.sim_versions` you can use, but the `Verilator` one doesn't yet have a `from_commandline` method to parse out `Verilator 5.037 devel rev v5.036-139-g206a0b4f` etc.

Don't worry about SAIF outputs from other simulators for now. Once we have more than one, we will try to figure out how to best support this feature across all of them.

martin-lueker:
> There is `cocotb_tools.sim_versions` you can use, but the `Verilator` one doesn't yet have a `from_commandline` method to parse out `Verilator 5.037 devel rev v5.036-139-g206a0b4f` etc.
> 
> Don't worry about SAIF outputs from other simulators for now. Once we have more than one, we will try to figure out how to best support this feature across all of them.

Great.  I added a `from_commandline` method here (one that is very similar to the one other instance), and setup the test to now be skipped unless it is for Verilator >= 5.042.

martin-lueker:
Hi @ktbarrett,
    I've now added a newsfragment also, and double checked that the updated documentation for the new features matches the existing documentation for the other trace output formats for verilator.  Is there anything else that is missing?

docs/source/newsfragments/5114.feature.rst
```diff
@@ -0,0 +1 @@
+Added support for SAIF tracing when using Verilator>=5.042 as the simulator.  This is enabled by adding ``--trace-saif`` to :make:var:`EXTRA_ARGS` (when using Makefiles) or to ``build_args`` (when using the runner)
```
cmarqu:
```suggestion
Added support for SAIF tracing when using Verilator>=5.042 as the simulator.  This is enabled by adding ``--trace-saif`` to :make:var:`EXTRA_ARGS` (when using Makefiles) or to ``build_args`` (when using the runner).
```

src/cocotb_tools/sim_versions.py
```diff
@@ -110,6 +110,13 @@ class VerilatorVersion(LooseVersion):
         True
     """
 
+    @classmethod
+    def from_commandline(cls, cmdline):
```
cmarqu, line 114: This is public API, please add a docstring.

docs/source/simulator_support.rst
```diff
@@ -140,6 +140,17 @@ To enable FST tracing, add ``--trace-fst`` to :make:var:`EXTRA_ARGS` as shown be
 
 The resulting file will be :file:`dump.fst` and can be opened by ``gtkwave dump.fst``.
 
+In addition to waveforms, Verilator (versions 5.042 and higher) can generate SAIF activity files for use in power estimation.
```
cmarqu:
```suggestion
Power Estimation
----------------

Verilator (versions 5.042 and higher) can generate SAIF activity files for use in power estimation.
```
Different enough to warrant its own subsection IMO.

martin-lueker: Hi @cmarqu, Thanks for the review.  It's good to hear from you!
I can certainly add this as a new section, however there is one concern that has to be addressed if we make this a separate subsection.

Though in the _abstract_, SAIF generation and Waveform generation serve different functions, in the context of Verilator they are closely linked in the implementation.  The two output formats are, IIRC, mutually exclusive.   They certainly are mutually exclusive in the current cocotb implementation in this PR (I'll double check if verilator can do multiple output formats to confirm.) So if we make this a separate section, I'll have to note this in documentation here as well.

src/cocotb_tools/sim_versions.py
martin-lueker: Done.

docs/source/simulator_support.rst
```diff
@@ -154,6 +154,12 @@ To enable SAIF tracing, add ``--trace-saif`` to :make:var:`EXTRA_ARGS` as shown
 
 The resulting file will be :file:`dump.saif` which can be processed by most synthesis tools.
 
+.. note::
+
+    Please note that (as of Verilator v5.044) for any given build, verilator can only output traces in one format, be it VCD, FST, or SAIF.
+    For instance if both the ``--trace-saif`` and ``--trace-fst`` flags are applied, only the latter format will be generated.
+    Thus waveform tracing and activity tracing should be done with different builds.
```
martin-lueker: @cmarqu, any thoughts on this additional note?

src/cocotb_tools/sim_versions.py
```diff
@@ -112,6 +112,22 @@ class VerilatorVersion(LooseVersion):
...
+
+        Args:
+            cmdline: The command-line output of a call to verilator --version.
+
+        Returns:
+            An instance of VerilatorVersion.
+
+        Raises:
+            AssertionError: If *cmdline* does not appear to be generated by verilator.
+        """
```
martin-lueker, line 130: @cmarqu: Here is the docstring you requested.

martin-lueker:
@ktbarrett: Can we remove the status:needsnewsfragment and status:needstest labels at this point? WDYT?

docs/source/simulator_support.rst
```diff
@@ -140,6 +140,26 @@ To enable FST tracing, add ``--trace-fst`` to :make:var:`EXTRA_ARGS` as shown be
...
+----------------
+
+Verilator (versions 5.042 and higher) can generate SAIF activity files for use in power estimation.
+These compact files do not record the state of signals over time, rather they record general switching statistics to calculate the amount of power dissipated by each signal.
+
+To enable SAIF tracing, add ``--trace-saif`` to :make:var:`EXTRA_ARGS` as shown below.
+
+  .. code-block:: make
+
+    EXTRA_ARGS += --trace --trace-saif --trace-structs
```
ktbarrett: I think we should also add an example of passing this as a `build_arg` to a Runner. Specify that the `EXTRA_ARGS` example works if you are using Makefiles.


```suggestion
To enable SAIF tracing, if using the Makefiles, add ``--trace-saif`` to :make:var:`EXTRA_ARGS` as shown below.

.. code-block:: make

    EXTRA_ARGS += --trace --trace-saif --trace-structs
    
If you are using the runner add ``--trace-saif`` to the ``build_args`` argument to :func:`.Runner.build`.

.. code-block:: python

    runner.build(
        build_args=["--trace", "--trace-saif", "--trace-structs"],
    )
```

docs/source/simulator_support.rst
cmarqu: Ah, the exclusivity wasn't obvious to me. I agree, let's keep it in one section then, unless we find they can be switched on independently.

docs/source/simulator_support.rst
cmarqu: No, perfectly fine IMO.

src/cocotb_tools/sim_versions.py
```diff
@@ -110,6 +110,29 @@ class VerilatorVersion(LooseVersion):
...
+            True
+
+        Args:
+            cmdline: The command-line output of a call to verilator --version.
+
+        Returns:
+            An instance of VerilatorVersion.
+
+        Raises:
+            AssertionError: If *cmdline* does not appear to be generated by verilator.
```
cmarqu:
```suggestion
            AssertionError: If *cmdline* does not appear to be generated by Verilator.
```
Nitpick.

tests/pytest/test_waves.py
```diff
@@ -86,6 +91,33 @@ def test_wave_dump():
...
+    ).stdout
+    sim_version = VerilatorVersion.from_commandline(sim_version_str)
+    if sim_version >= VerilatorVersion("5.042"):
+        skip_saif_dump_test = False
+
+
+@pytest.mark.simulator_required
+@pytest.mark.skipif(
+    skip_saif_dump_test,
+    reason="Skipping test because it is only for the Verilator simulator, v5.402 or higher",
```
cmarqu:
```suggestion
    reason="Skipping test because it is only for the Verilator simulator, v5.042 or higher",
```

cmarqu (approved):
@martin-lueker Sorry for the long silence, just two small things to fix before merging (plus the one Kaleb requested).

tests/pytest/test_waves.py
martin-lueker: Oh, Good Catch!  Thank you!

docs/source/simulator_support.rst
```diff
@@ -140,6 +140,26 @@ To enable FST tracing, add ``--trace-fst`` to :make:var:`EXTRA_ARGS` as shown be
...
+
+The resulting file will be :file:`dump.saif` which can be processed by most synthesis tools.
+
+.. note::
+
+    Please note that (as of Verilator v5.044) for any given build, verilator can only output traces in one format, be it VCD, FST, or SAIF.
```
martin-lueker:
```suggestion
    Please note that (as of Verilator v5.044) for any given build, Verilator can only output traces in one format, be it VCD, FST, or SAIF.
```

martin-lueker:
Alright.   Thanks @ktbarrett and @cmarqu.   I added your last changes, and squashed and rebased everything.

I think that's everything.  Please give it one more scan if you like.

docs/source/simulator_support.rst
cmarqu: FWIW, https://github.com/verilator/verilator/issues/5813 is the related enhancement ticket.

src/cocotb_tools/sim_versions.py
```diff
@@ -110,6 +110,29 @@ class VerilatorVersion(LooseVersion):
         True
     """
 
+    @classmethod
+    def from_commandline(cls, cmdline):
+        """Parses the output of verilator --version
+
+        Example:
+            >> cmdline = "Verilator 5.041 devel rev v5.040-1-g4eb030717"
+            >> VerilatorVersion.from_commandline(cmdline) >= VerilatorVersion("5.040")
```
cmarqu:
```suggestion
        """Parse the output of ``verilator --version``.

        Example:
            >>> cmdline = "Verilator 5.041 devel rev v5.040-1-g4eb030717"
            >>> VerilatorVersion.from_commandline(cmdline) >= VerilatorVersion("5.040")
```

src/cocotb_tools/sim_versions.py
```diff
@@ -110,6 +110,29 @@ class VerilatorVersion(LooseVersion):
...
+            True
+
+        Args:
+            cmdline: The command-line output of a call to verilator --version.
+
+        Returns:
+            An instance of VerilatorVersion.
```
cmarqu:
```suggestion
            cmdline: The command-line output of a call to ``verilator --version``.

        Returns:
            An instance of :class:`VerilatorVersion`.
```

martin-lueker: Looks fine to me!   I see that @ktbarrett already committed them.  I just squashed them into the final commit.

martin-lueker:
Okay, one more change from my side: added a link to verilator/verilator#5813.

docs/source/simulator_support.rst
martin-lueker: Great! I was not aware Wilson was already looking into this.  I updated this line and added this link.

docs/source/simulator_support.rst
```diff
@@ -140,6 +140,34 @@ To enable FST tracing, add ``--trace-fst`` to :make:var:`EXTRA_ARGS` as shown be
...
+        build_args=["--trace", "--trace-saif", "--trace-structs"],
+    )
+
+The resulting file will be :file:`dump.saif` which can be processed by most synthesis tools.
+
+.. note::
+
+    As noted in `verilator/verilator#5813 <https://github.com/verilator/verilator/issues/5813>`_, for any given build Verilator can only output traces in one format, be it VCD, FST, or SAIF.
+    For instance, if both the ``--trace-saif`` and ``--trace-fst`` flags are applied, only the latter format will be generated.
+    Thus until this issue is resolved, waveform tracing and activity tracing should be done with different builds.
```
ktbarrett: Linking to Github issues in documentation that lives forever is generally a bad idea.

Documentation should use a "matter of fact" tone. Also, it's not only Verilator's problem to solve, but we will need to adapt if and when they do add support for multiple tracing outputs. So avoid blame and don't trick the user into thinking that once that issue is resolved cocotb 2.1 will suddenly support it. 

The second sentence is over-specific, the user shouldn't depend on whether Verilator chooses the last or the first option, so don't offer them that info.

```suggestion
    For any given build, only one output format (VCD, FST, or SAIF) can be used.
    If both waveform tracing and activity tracing are needed, they should be done with different builds.
```

martin-lueker: Hi @ktbarrett, 
   Thanks for your point of view on this.  I sincerely apologize, I did not mean to make any implication of blame, and when I read your comment, I was more than a bit embarrassed!  My intent was to make statements of fact, which some people may see these as, but in hindsight I agree, that perspective may only apply when talking to an audience with a very narrow lens.  I do see your point that this could easily be interpreted as finger-pointing, even though that was not my intent at all!  I'm so sorry about that!

   @cmarqu, It seems I also misinterpreted your comment with the issue link, thinking that it would be helpful to the reader to have this additional context about what needs to happen next before this can be supported.   That was also my mistake.

   My first inclination was to roll back this most recent particular change (I locally keep unsquashed commit trees for all my PRs), but @ktbarrett, I like the conciseness of this change here.  I'll add it.   I'll also put back the full unsquashed version until every one agrees that this is finally really ready to be merged.

Thanks for your patience as I learn the community norms here. -M

martin-lueker:
Rebased to master, and unsquashed (until ready to merge).

docs/source/simulator_support.rst
ktbarrett: It no problem, just explaining our thought process as to what goes into documentation.

martin-lueker: I see.  Well, great.  Your changes are added, and the whole thing is squashed again.   Feel free to merge when ready.

PR #5120, opened 2025-11-05
Remove blurb about cocotb-bus from docs

ktbarrett:
We don't want people using it, so we should stop mentioning it.

### DIFF
--- a/docs/source/extensions.rst
+++ b/docs/source/extensions.rst
@@ -26,30 +26,8 @@ Assuming an extension named ``EXTNAME`` (all lower-case),
 - the package is in the ``cocotbext.EXTNAME`` namespace, and
 - the distribution (package) name is prefixed with ``cocotbext-EXTNAME``.
 
-Example:
-An SPI bus extension might be packaged as ``cocotbext-spi``, and its functionality would live in the ``cocotbext.spi`` namespace.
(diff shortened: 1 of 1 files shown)


shane-movandi:
> We don't want people using it, so we should stop mentioning it.

Why don't you want people using it?
What should be used instead?

ktbarrett:
@shane-movandi This isn't really the forum to discuss this, check out the [gitter](https://matrix.to/#/#cocotb_Lobby:gitter.im). To summarize: the code has lots of smells, isn't maintained, isn't well tested, and it doesn't do anything a junior dev couldn't put together in an afternoon (with exception of the specific bus interfaces, which we hope someone will actually copy out and maintain). There is not an alternative and it's out of scope of cocotb to provide one. Check out pyuvm, AVL, or forastero if you are looking for a more complete framework.

shane-movandi:
> @shane-movandi This isn't really the forum to discuss this, check out the [gitter](https://matrix.to/#/#cocotb_Lobby:gitter.im). To summarize: the code has lots of smells, isn't maintained, isn't well tested, and it doesn't do anything a junior dev couldn't put together in an afternoon (with exception of the specific bus interfaces, which we hope someone will actually copy out and maintain). There is not an alternative and it's out of scope of cocotb to provide one. Check out pyuvm, AVL, or forastero if you are looking for a more complete framework.

Thank you, I appreciate you taking the time to reply.

PR #5128, opened 2025-11-08
docs: Cleanup warnings and errors

marlonjames:
* Cleanup after #5033
* Fix some warnings and errors
* Skip autolink parsing in some code blocks

There are still some autolink warnings about resolving things,
but they are mostly local variables being created in the code blocks.
I'm not sure why those fail.

The autolink docs example links from the variable to the type that it is:
See ``knight`` here: https://sphinx-codeautolink.readthedocs.io/en/latest/examples.html#basic-use

### DIFF
--- a/src/cocotb/_decorators.py
+++ b/src/cocotb/_decorators.py
@@ -392,6 +392,7 @@ def parametrize(
     The call to ``parametrize`` should include the name of each test parameter and the possible values each parameter can hold.
     This will generate a test for each of the Cartesian products of the parameters and their values.
 
+    .. autolink-skip::
     .. code-block:: python
 
         @cocotb.test(
@@ -405,6 +406,7 @@ def parametrize(
 
     The above is equivalent to the following.
 
+    .. autolink-skip::
     .. code-block:: python
 
         @cocotb.test(skip=False)
@@ -434,6 +436,7 @@ def parametrize(
     either by supplying tuples of the parameter name to values,
     or a sequence of variable names and a sequence of values.
 
+    .. autolink-skip::
     .. code-block:: python
 
         @cocotb.parametrize(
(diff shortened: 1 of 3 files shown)


src/cocotb/_decorators.py
```diff
@@ -405,6 +406,7 @@ async def my_test(arg1: int, arg2: str) -> None: ...
 
     The above is equivalent to the following.
 
+    .. autolink-skip::
```
ktbarrett, line 409: Why do these two need to be skipped?

src/cocotb/task.py
```diff
@@ -121,6 +121,7 @@ def locals(self) -> SimpleNamespace:
         A modifiable namespace where any user-defined variables can be added.
         Use :func:`~cocotb.task.current_task` to access the current Task's locals.
 
+        .. autolink-skip::
         .. code-block:: python
```
ktbarrett, line 125: Same here.

ktbarrett (approved):

src/cocotb/_decorators.py
marlonjames: They warn on not being able to resolve `arg1`, `arg2`, and `arg3`, which make no sense to me, along with some of the other resolve failure warnings.
I don't really care either way, but given they are in the docs right next to `cocotb.parametrize()`, there's no big downside to not having links.

src/cocotb/task.py
marlonjames: Like the other, just a failure to resolve. In this case it doesn't know what `locals.rng` is.

ktbarrett: Oof. Another tool based on asteriod, huh? Well at least we are trending in the right direction.

ktbarrett (approved):

PR #5129, opened 2025-11-09
Make reason_to_string() static class method

marlonjames:
It's not dependent on an object, and it's only used inside each implementation.
I'm also planning on using it inside the sim callback for GPI tracing, where I might not have a valid GPI object anyway.


src/cocotb/share/lib/fli/FliImpl.cpp
```diff
@@ -435,10 +435,6 @@ GpiObjHdl *FliImpl::get_child_by_index(int32_t index, GpiObjHdl *parent) {
     }
 }
 
-const char *FliImpl::reason_to_string(int) {
-    return "Who can explain it, who can tell you why?";
```
ktbarrett, line 439: Fools give you reasons. Wise men never try.

PR #5131, opened 2025-11-09
Add GPI tracing (part 1)

marlonjames:
Add tracing to GPI and PyGPI.

This adds tracing logs for GPI startup, shutdown, and callbacks.

I don't expect much performance impact, as enable is set once at startup.
It might actually improve now that the `LOG_TRACE()` calls in `c_to_python`/`python_to_c` are inside an if statement.

---
- [x] Add newsfrag

### DIFF
--- a/docs/source/newsfragments/5131.feature.rst
+++ b/docs/source/newsfragments/5131.feature.rst
@@ -1,0 +1,1 @@
+Added :envvar:`GPI_DEBUG` and :envvar:`PYGPI_DEBUG` environment variables to enable additional debug features, including TRACE log messages.


docs/source/library_reference.rst
```diff
@@ -480,6 +480,15 @@ Logging
...
 
+.. envvar:: GPI_TRACING
+
+    Enable verbose log output tracing the code execution path through the GPI.
+    The messages are logged at GPI log level ``TRACE``,
+    so :envvar:`GPI_LOG_LEVEL` must be set to ``TRACE`` in order to see
+    tracing messages during early GPI startup.
+
+    .. versionadded:: 2.1
+
```
ktbarrett: Why do we need these? Calls aren't expensive in C and because of the C++ local log level variable it should be pretty fast.

src/cocotb/share/include/gpi_logging.h
```diff
@@ -234,9 +249,9 @@ GPILOG_EXPORT void gpi_clear_log_handler(void);
  *                    according to format specifier in msg argument
  */
 #define gpi_native_logger_log(extra_name, level, pathname, funcname, lineno, \
-                              ...)                                           \
+                              msg, ...)                                      \
     gpi_native_logger_log_(MAKE_LOG_NAME_(extra_name), level, pathname,      \
-                           funcname, lineno, __VA_ARGS__)
+                           funcname, lineno, msg, __VA_ARGS__)
```
ktbarrett: Unfortunately this isn't compliant C++11, empty `__VA_ARGS__` is a standard violation and causes warnings.

src/cocotb/share/include/py_gpi_logging.h
```diff
@@ -26,6 +28,47 @@ PYGPILOG_EXPORT void py_gpi_logger_finalize();
...
+        if (python_context_tracing_enabled) {                              \
+            if (!is_python_context) {                                      \
+                LOG_ERROR(                                                 \
+                    "FATAL: Trying Python => C but already in C context"); \
+                exit(1);                                                   \
+            }                                                              \
+            --is_python_context;                                           \
+            LOG_TRACE("Python => C");                                      \
+        }                                                                  \
+    } while (0)
```
ktbarrett, line 99: Lets put this in a source file and let LTO do its job.

tests/test_cases/test_cocotb/test_logging.py
```diff
@@ -75,7 +75,7 @@ async def test_logging_with_args(dut):
 @cocotb.test()
 async def test_custom_logging_levels(dut):
     logging.basicConfig(level=logging.NOTSET)
-    logging.addLevelName(5, "SUPER_DEBUG")
+    logging.addLevelName(7, "SUPER_DEBUG")
```
cmarqu, line 78: Some leftover?

ktbarrett: I think this was intentional. `TRACE` is level 5.

ktbarrett:
If you are worried about performance, you can move the gpi log level check out of the body of the function into the macro in the header. That would help perf at other log levels too.

tests/test_cases/test_cocotb/test_logging.py
marlonjames: Yeah intentional. It overrides the name "TRACE" in the logs with "SUPER_DEBUG". Not strictly necessary fix, but I changed it to be between `DEBUG` and `TRACE`.

docs/source/library_reference.rst
marlonjames: That's a lot of function calls through function pointers, for several discarded log messages going each direction through the GPI, for every single callback. For a feature that is ~always turned off.
Tracing is on or off, so it's perfectly aligned with a global, and wrapping everything in the (very predictable) if statement cuts off unnecessary work.
It's also a behavior switch like `COCOTB_SCHEDULER_DEBUG`, which we had discussed before.

ktbarrett: Then I think I might prefer the switch's name reflect that intent then. I had hoped to put some of the VPI and VHPI errors behind such a flag as well. (I don't love the name `COCOTB_SCHEDULER_DEBUG` either). Maybe `GPI_DEBUG` and `PYGPI_DEBUG`?

src/cocotb/share/include/py_gpi_logging.h
marlonjames: It's used in both the simulator module and libcocotb. I can move it to its own file and include in both libraries and not export it, or we can deal with it later when/if those are libraries are merged.
I don't want to force it to cross the library boundary by only building it into one lib, I would want to encourage it to be inlined by the compiler.

ktbarrett: Oh right. That'd be less of an issue after the library merge, but okay.

src/cocotb/share/lib/vpi/VpiImpl.cpp
```diff
@@ -777,19 +787,31 @@ static void vpi_main() {
...
 extern "C" {
-COCOTBVPI_EXPORT void (*vlog_startup_routines[])() = {vpi_main, nullptr};
+COCOTBVPI_EXPORT void (*vlog_startup_routines[])() = {
+    gpi_init_logging_and_tracing, vpi_main, nullptr};
 
 // For non-VPI compliant applications that cannot find vlog_startup_routines
 // symbol
-COCOTBVPI_EXPORT void vlog_startup_routines_bootstrap() { vpi_main(); }
+COCOTBVPI_EXPORT void vlog_startup_routines_bootstrap() {
+    gpi_init_logging_and_tracing();
```
ktbarrett: I do like that this is split out and done first now.

src/cocotb/share/include/gpi_logging.h
```diff
@@ -284,6 +299,16 @@ GPILOG_EXPORT bool gpi_native_logger_filtered(int level);
...
 
+extern GPILOG_EXPORT int gpi_tracing_enabled;
+
+/** Logs a message at TRACE log level if GPI tracing is enabled */
+#define GPI_TRACE_MSG(...)          \
+    do {                            \
+        if (gpi_tracing_enabled) {  \
+            LOG_TRACE(__VA_ARGS__); \
+        }                           \
+    } while (0)
```
ktbarrett: Why both this and `GPI_TRACE`?

ktbarrett:
So I think maybe just making the name a little more generic and merging that one bit of duplicated functionality is all this needs. `GPI_TRACE` always being run conditionally on the debugging flag being set I think is appropriate.

src/cocotb/share/include/gpi_logging.h
marlonjames: Yeah, that makes sense.. I saw the doxygen warning about the `msg` param missing and didn't think that through.

src/cocotb/share/include/gpi_logging.h
marlonjames: Do you mean `LOG_TRACE`? GPI users may want different criteria when trace logging. That's why I have PYGPI and GPI tracing separated. It shouldn't be all tracing on or off with a single flag.

ktbarrett: We should not be providing logging functions to users. That has them putting user log messages under the `gpi` namespace. All of these `LOG_*` functions should get moved to `gpi_priv.h` and the uses in the PyGPI updated to use `pygpi` namespace.

marlonjames:
I converted the rest of the pygpi logs to the right  namespace,
and I changed the C <=> Python error messages to critical.

src/cocotb/share/lib/fli/FliCbHdl.cpp
```diff
@@ -10,27 +10,31 @@
...
     }
     // LCOV_EXCL_STOP
 
-    if (cb_hdl->run()) {
-        // sim failed, so call shutdown
+    if (!error) {
+        GPI_TO_USER_CB(FLI);
+        error = cb_hdl->run();
+        USER_CB_TO_GPI(FLI);
+    }
```
ktbarrett, line 31: `else`?

ktbarrett (approved):

ktbarrett:
Still need newsfrag.

src/cocotb/share/lib/fli/FliCbHdl.cpp
marlonjames: The callback could return non-zero, so we check again below.

ktbarrett (approved):

PR #5139, opened 2025-11-12
Update license format in pyproject.toml to work with setuptools <v77

Ofthemasses:
Resolves dual licensing error and references license files with PEP 621 specs.

Error:
```configuration error: `project.license` must be valid exactly by one definition (2 matches found):```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,10 +6,7 @@ build-backend = "setuptools.build_meta"
 name = "cocotb"
 description = "Python-based chip (RTL) verification"
 readme = "README.md"
-license = "BSD-3-Clause"
-license-files = [
-    "LICENSE"
-]
+license = { file = "LICENSE" }
 authors = [
     {name = "Chris Higgs"},
     {name = "Stuart Hodgson"},

### CODE REVIEW

**Comment by ktbarrett** [comment]
Im fairly certain this breaks version compat with older Pythons that use older versions of setuptools, but let's let CI tell us.

**Comment by Ofthemasses** [comment]
> Im fairly certain this breaks version compat with older Pythons that use older versions of setuptools, but let's let CI tell us.

Heya, a mistake with my title, meant to be less than v77. This issue was found when using python v3.11.6, setup tools v65.5.0. setuptools v77 introduces license-files.

Here is a related issue: https://github.com/pypa/packaging-problems/issues/870

**Comment by ktbarrett** [comment]
Odd to me. I've seen it before, but only on Python <3.9. Im guessing because newer versions of Python can install newer versions of pip. So it sounds like you're using an older version of pip and a newer version of Python? 

Have you tried?
```sh
> pip install --upgrade pip
```

ktbarrett:
Sorry, the Github UI gave me a prompt and I filled it out =/ Don't Github on mobile.

Ofthemasses:
> Sorry, the Github UI gave me a prompt and I filled it out =/ Don't Github on mobile.

Wait GitHub does AI responses?

Ofthemasses:
I'm working on the oss-cad-suite project and yosys is using that particular version of python, with this repo tagged to makn. There's lots of old packages so I'm guessing an upgrade would cause issues

ktbarrett:
So new versions of setuptools are removing support for PEP621-style license field Feb 2026. So I don't think this is what we want to do. Instead we should update the build-backend field to require setuptools>=77 and leave the license fields as they are.

I'm sorry, you'll have to bug `oss-cad-suite` about updating their version of setuptools.

ktbarrett:
Superseded by #5145.

PR #5145, opened 2025-11-13
Bump required version of setuptools in build

ktbarrett:
setuptools is removing support for PEP621 style license fields in pyproject.toml in February 2026. Supersedes #5139.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools>=77", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]


imphil:
I'm missing something here: if setuptools is removing support for something, how does it help us to use a newer version of setuptools?

ktbarrett:
@imphil We currently are specifying the license fields in pyproject.toml using PEP 639 syntax which was first supported in setuptools 77. But we missed bumping that dependency. This is the issue that #5139 hoped to resolve by reverting to PEP621 syntax.

We could return to using PEP 621 syntax. If we do that, cocotb can't be built with newer versions of setuptools starting early next year, perhaps gating bug fixes to distutils. If we do revert that change we should instead add an upper bound on setuptools in the build requirements, something like `setuptools >= 62, < 82`, so that cocotb will continue to build correctly after that date by choosing an older version of setuptools. But additional bounds on packages make solving complex environments harder.

Another alternative is to move the license info back into setup.py or setup.cfg and specifying the field as `"dynamic"` in the pyproject.toml.

imphil:
Alright then, thanks for explaining!

PR #5146, opened 2025-11-14
stable/2.0: Move VCS to experimental bucket

imphil:
We don't have licenses available at the moment. Move VCS to the experimental bucket until that gets fixed to avoid blocking the release pipeline.

Backport of #5126 for stable/2.0.


imphil:
The release pipeline passes with this PR.

ktbarrett (approved):

PR #5150, opened 2025-11-17
Micro-optimizations to Task / EventLoop

ktbarrett:
Several small micro-optimizations which help performance.

### DIFF
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -208,33 +208,39 @@ class Task(Generic[ResultType]):
         else:
             coro_name = type(self._coro).__name__
 
-        if self._state is _TaskState.RUNNING:
+        state = self._state
+        if state is _TaskState.RUNNING:
             return f"<{self._name} running coro={coro_name}()>"
-        elif self._state is _TaskState.FINISHED:
+        elif state is _TaskState.FINISHED:
             return (
                 f"<{self._name} finished coro={coro_name}() outcome={self._outcome!r}>"
             )
-        elif self._state is _TaskState.ERRORED:
+        elif state is _TaskState.ERRORED:
             return f"<{self._name} error coro={coro_name}() outcome={self._outcome!r}>"
-        elif self._state is _TaskState.PENDING:
+        elif state is _TaskState.PENDING:
             return (
                 f"<{self._name} pending coro={coro_name}() trigger={self._trigger!r}>"
             )
-        elif self._state is _TaskState.SCHEDULED:
+        elif state is _TaskState.SCHEDULED:
             return f"<{self._name} scheduled coro={coro_name}()>"
-        elif self._state is _TaskState.UNSTARTED:
+        elif state is _TaskState.UNSTARTED:
             return f"<{self._name} created coro={coro_name}()>"
-        elif self._state is _TaskState.CANCELLED:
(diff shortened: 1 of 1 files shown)


ktbarrett:
~~I think I can make it so the callbacks stuff work without arguments which should help even more. Also, Im uncertain about the CancellationError stuff, specifically removing the check in _resume.~~

EDIT: resolved

src/cocotb/task.py
```diff
@@ -217,16 +217,20 @@ def __repr__(self) -> str:
 
         if self._state is _TaskState.RUNNING:
             return f"<{self._name} running coro={coro_name}()>"
-        elif self._state is _TaskState.FINISHED:
-            return f"<{self._name} finished coro={coro_name}() outcome={self._outcome}>"
+        elif self._state in (_TaskState.FINISHED, _TaskState.ERRORED):
+            return (
+                f"<{self._name} finished coro={coro_name}() outcome={self._outcome!r}>"
+            )
```
ktbarrett: I can split this into "finished" and "errored" if desired. It would make it more obvious that it ended with an exception vs returned one.

src/cocotb/task.py
```diff
@@ -278,50 +286,48 @@ def _resume(self, exc: BaseException | None = None) -> None:
...
         _current_task = self
 
         try:
-            if exc is None:
+            if self._exc is None:
                 trigger = self._coro.send(None)
             else:
-                trigger = self._coro.throw(exc)
+                trigger = self._coro.throw(self._exc)
+                self._exc = None
```
marlonjames: I haven't looked in detail, but couldn't there be an issue here where `self._exc` isn't reset to `None`? Is that a problem?

ktbarrett: I don't think so, Tasks can only be resumed by the function that sets this to None or Exception on every call. Or `_cancel_now` which also sets it (and will hopefully go away once TestManager (née RunningTask) is refactored a bit).

ktbarrett:
There must have just been a nox release, the typing fixes are in so I added to the commit to fix that.

src/cocotb/task.py
```diff
@@ -278,50 +288,48 @@ def _resume(self, exc: BaseException | None = None) -> None:
...
```
marlonjames: If we know which branch this will take before scheduling `_resume()`, is it worth splitting this into 2 functions?
We may duplicate some of the except cases but it may be a win in terms of speed?

ktbarrett: This is actually important to keep to get TaskManager working again, we need to reschedule a Task with a new exception (CancelledError). If its scheduled we can leave its position in queue and just update the exception to throw so it might change from None to exception. I suppose we could cancel the old scheduled callback and put a new one at the end of the queue, but I'm not sure if thats a great idea as there are several situations where we could deadlock by continually rescheduling to the end of the queue. (Also its more expensive).

```suggestion
```

ktbarrett: Hmmm... maybe this needs some more thought, especially wrt repeated calls to `cancel()`. I'll merge this and do some more thinking as it doesn't break anything.

ktbarrett: Opened #5153.

marlonjames (approved):

PR #5160, opened 2025-11-19
Use simulator-only registration function in xcelium

gilbertoabram:
This syntax ensures that `vlog_startup_routines_bootstrap()` is called only during simulation and not elaboration when using xcelium.

### DIFF
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -1761,12 +1761,6 @@ class Xcelium(Runner):
             + verbosity_opts
             # + ["-vpicompat 1800v2005"]  # <1364v1995|1364v2001|1364v2005|1800v2005> Specify the IEEE VPI
             + ["-access +rwc"]
-            + ["-loadvpi"]
-            # always start with VPI on Xcelium
-            + [
-                cocotb_tools.config.lib_name_path("vpi", "xcelium").as_posix()
-                + ":.vlog_startup_routines_bootstrap"
-            ]
             + vhpi_opts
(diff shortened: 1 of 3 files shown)


src/cocotb_tools/runner.py
```diff
@@ -1713,7 +1713,7 @@ def _build_command(self) -> list[_Command]:
             # always start with VPI on Xcelium
             + [
                 cocotb_tools.config.lib_name_path("vpi", "xcelium").as_posix()
-                + ":vlog_startup_routines_bootstrap"
+                + ":.vlog_startup_routines_bootstrap"
```
ktbarrett: Actually wouldn't it be more correct to remove loading the VPI here and changing the call in `_test_command` to this so it doesn't try to load during test elab?

ktbarrett: Apparently this makes Xcelium upset.

gilbertoabram:
Sorry for the wait. I checked with Cadence if it's possible in any way to use `-loadvpi` only during simulation (with or without dot) since it would make the most sense but I couldn't make it work. Sadly it is not possible at this time, but it could end up in a future release.

ktbarrett (approved):

gilbertoabram:
Turns out there is actually a flag that does what we want, so I moved it to the simulation stage per the early comment, and it seems to work. I reworked the documentation a bit: it seems that both Verilog and VHDL shared all the same instructions except one extra flag for VHDL toplevel. Not sure it makes sense, suggestions/changes are welcome.

ktbarrett (approved):

PR #5162, opened 2025-11-22
`Task.cancel` changes and Task debug logging refactor

### LINKED ISSUE #5153: Consider livelock via rescheduling
https://github.com/cocotb/cocotb/pull/5150#discussion_r2539556283

A livelock occurs when a Task is repeatedly rescheduled which bumps its position in queue to the back, never allowing it to run and progressing the program. There are a couple possible scenarios that could lead to this:
1. Repeated calls to `Task.cancel()` if it reschedules an already scheduled resume.
2. Atypical rescheduling calls (e.g. in TaskManager's soft-cancel on Task failure while in the TaskManager's context block)

We should ensure that all rescheduling "steals" the previous schedules position in queue to force the program to progress.

Also currently all scheduling is done in `_schedule_resume` which doesn't currently take into account the current state and assumes its safe to move to SCHEDULED with any other actions handled by the caller. We should for case 2 provide a function which can schedule a resume from any context and save the current function for solely internal calls (or simply inline it at all call sites).

ktbarrett:
Closes #5153. Preps the Task object for use with TaskManager.

* Refactor Task debug logs
* raise RuntimeError when cancelling currently running Task
* Hijack current schedule when cancelling scheduled Task rather than re-schedule

The logs in _schedule_resume and _set_outcome were removed so more specific logs could be added at each state change location. Also some logs which output arbitrary user values were changed to only output the type as the value might be very large making the debug logs useless.

Also changed Task.cancel() so you can't try to cancel the current Task. This wouldn't work currently since the currently running Task would continue executing enteriung the resume in an unintended state potentially causing bad state changes.

Finally, changed the implementation of Task.cancel to hijack the current schedule if a Task is scheduled when it's cancelled. This will prevent livelocks occuring if a Task is repeatedly cancelled by another Task.

### DIFF
--- a/docs/source/newsfragments/5162.change.rst
+++ b/docs/source/newsfragments/5162.change.rst
@@ -1,1 +1,1 @@
-Disallow calling :meth:`.Task.cancel` on the currently running task.
+Changed :meth:`.Task.cancel` to throw :exc:`RuntimeError` if attempting to cancel the currently running task.


ktbarrett:
Added some tests for uncovered lines. There are lines in `_cancel_now` which can never be hit since it's only run at test end (for now).

docs/source/newsfragments/5162.change.rst
```diff
@@ -0,0 +1 @@
+Disallow calling :meth:`.Task.cancel` on the currently running task.
```
ktbarrett:
```suggestion
Changed :meth:`.Task.cancel` to throw :exc:`RuntimeError` if attempting to cancel the currently running task.
```

PR #5163, opened 2025-11-22
Add `TaskManager`

### LINKED ISSUE #922: Improve exception handling in forked coroutines

### LINKED ISSUE #3583: Separating symmetric and asymmetric tasks

ktbarrett:
Closes #3583. Closes #922. Depends on #5162. This adds `exceptiongroup` as a dependency for users of Python <3.11 where `ExceptionGroup` was introduced into Python stdlib.

Adds the `TaskManager` object, which behaves almost identically to `asyncio.TaskGroup`. There are a couple differences.

`TaskManager.fork` is a decorator for an async function that is shorthand for the following:
```python
async with TaskManager() as tm:

    @tm.fork
    async def my_func_shorthand():
        ...

    async def my_func():
        ...
    tm.start_soon(tm)
```

And the second is that users cannot ignore the `CancelledError` thrown into the body of the TaskManager if a child fails. If this occurs, upon entering the TaskManager's `__aexit__` the Task will fail.

```python
async with TaskManager() as tm:
    @tm.fork
    async def my_func():
        await Timer(1)
        raise Exception()

    try:
        await my_func
    except CancelledError:
        pass  # ignoring the CancelledError here is bad

```

### TODO
- [x] newsfrag
- [x] test for ignoring CancelledError
- [x] more tests?
- [x] Run all tests directly, nested in another TaskManager block, and nested in another TaskManager's child
- [x] fix docs TODOs
- [x] Add `cancel_on_error` argument which defaults to `True`, on `False` it will allow the block and other child Tasks to continue running
- [x] Specify behavior when Task is cancelled, but TaskManager returns ExceptionGroup (if this is possible)
   - This will ignore the ExceptionGroup and propogate the CancelledError
- [x] Test the `continue_on_error` functionality

### DIFF
--- a/src/cocotb/_task_manager.py
+++ b/src/cocotb/_task_manager.py
@@ -100,33 +100,6 @@ class TaskManager:
         self._add_task(task, continue_on_error=continue_on_error)
         return task
 
-    def _add_task(
-        self,
-        task: Task[Any],
-        *,
-        continue_on_error: bool | None = None,
-    ) -> None:
-        if self._cancelled:
-            raise RuntimeError("Cannot add new Tasks to TaskManager after error")
-        elif self._finishing:
-            raise RuntimeError("Cannot add new Tasks to TaskManager after finishing")
-        elif not self._entered:
-            raise RuntimeError(
-                "Cannot add new Tasks to TaskManager before entering context"
-            )
-        if current_task() is not self._parent_task:
-            raise RuntimeError("Cannot add new Tasks to TaskManager from another Task")
-
-        # Track the Task and store per-Task continue_on_error setting
-        task._add_done_callback(self._done_callback)
-        if continue_on_error is None:
-            continue_on_error = self._default_continue_on_error
-        self._remaining_tasks[task] = continue_on_error
-        self._none_remaining.clear()
-
-        # Schedule the Task to run soon
-        task._ensure_started()
-
     @overload
     def fork(
         self,
@@ -194,6 +167,33 @@ class TaskManager:
         task = Task[T](coro_func(), name=coro_func.__name__)
         self._add_task(task, continue_on_error=continue_on_error)
         return task
+
+    def _add_task(
+        self,
+        task: Task[Any],
+        *,
+        continue_on_error: bool | None = None,
+    ) -> None:
+        if self._cancelled:
+            raise RuntimeError("Cannot add new Tasks to TaskManager after error")
+        elif self._finishing:
+            raise RuntimeError("Cannot add new Tasks to TaskManager after finishing")
+        elif not self._entered:
+            raise RuntimeError(
+                "Cannot add new Tasks to TaskManager before entering context"
+            )
+        if current_task() is not self._parent_task:
+            raise RuntimeError("Cannot add new Tasks to TaskManager from another Task")
+
+        # Track the Task and store per-Task continue_on_error setting
+        task._add_done_callback(self._done_callback)
+        if continue_on_error is None:
+            continue_on_error = self._default_continue_on_error
+        self._remaining_tasks[task] = continue_on_error
+        self._none_remaining.clear()
+
+        # Schedule the Task to run soon
+        task._ensure_started()
 
     def _done_callback(self, task: Task[Any]) -> None:
         """Callback run when a child Task finishes."""


ktbarrett:
I've decided to change the implementation to instead do a regular cancel and then once we get to `__aexit__` to uncancel.

ktbarrett:
Well I realize now why asyncio makes cancellation and uncancellation a count instead of a boolean flag: nested TaskManagers.

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,383 @@
+"""TaskManager and related code."""
```
cmarqu, line 5:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause

"""TaskManager and related code."""
```

tests/test_cases/test_cocotb/test_task_manager.py
```diff
@@ -0,0 +1,504 @@
+from __future__ import annotations
```
cmarqu, line 5:
```suggestion
# Copyright cocotb contributors
# Licensed under the Revised BSD License, see LICENSE for details.
# SPDX-License-Identifier: BSD-3-Clause

from __future__ import annotations
```

cmarqu (commented):
TaskManager is also not in the Library Reference yet.

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+
+    .. code-block:: python
+
+        async with TaskManager() as tm:
+
+            @tm.fork
+            async def drive_if1(): ...
+
+            @tm.fork
+            async def drive_if2(): ...
```
cmarqu:
```suggestion
            async def drive_interface1(): ...

            @tm.fork
            async def drive_interface2(): ...
```
Prevent users from reading it as a condition.

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+
+        # Control returns here when all drive Tasks have completed
+
+    After exiting the context block,
+    if any child :class:`!Task` fails with an exception,
+    one of two behaviors will occur:
+    if the ``continue_on_error`` parameter is ``False`` (default), all other child :class:`!Task`\ s are cancelled;
```
cmarqu:
```suggestion
    if the *continue_on_error* parameter is ``False`` (default), all other child :class:`!Task`\ s are cancelled;
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+    if the ``continue_on_error`` parameter is ``True``, other child :class:`!Task`\ s are allowed to continue running.
```
cmarqu:
```suggestion
    if the *continue_on_error* parameter is ``True``, other child :class:`!Task`\ s are allowed to continue running.
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+            @tm.fork
+            async def drive_if1(): ...
+
+            @tm.fork
+            async def drive_if2(): ...
+
+        # Control returns here when all drive Tasks have completed
+
+    After exiting the context block,
+    if any child :class:`!Task` fails with an exception,
```
cmarqu:
```suggestion
    if any child :class:`!Task` failed with an exception,
```
?

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+    one of two behaviors will occur:
+    if the ``continue_on_error`` parameter is ``False`` (default), all other child :class:`!Task`\ s are cancelled;
+    if the ``continue_on_error`` parameter is ``True``, other child :class:`!Task`\ s are allowed to continue running.
+
+    After all child :class:`!Task`\ s have finished,
+    all exceptions, besides :exc:`CancelledError`, are gathered into an :exc:`ExceptionGroup`,
+    or an :exc:`BaseExceptionGroup`, if at least one of the exceptions is a :exc:`BaseException`,
```
cmarqu:
```suggestion
    or a :exc:`BaseExceptionGroup`, if at least one of the exceptions is a :exc:`BaseException`,
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+    and raised in the enclosing scope.
+
+    You can catch the :exc:`!ExceptionGroup` to handle errors from child :class:`!Task`\ s
+    by either catching the :exc:`!ExceptionGroup` as you would typically;
+    or, if you are running Python 3.11 or later,
+    using the new ``except*`` syntax to catch specific exception types from the group.
```
cmarqu: Would be nice to link to https://docs.python.org/3/reference/compound_stmts.html#except-star

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+                cocotb.log.info(
+                    "The rest of the context block will be skipped due to task1 failing"
+                )
+                raise  # DON'T FORGET THIS
+
+            ...  # This code will be skipped
+
+
+    In addition to the :deco:`!fork` method for starting :term:`coroutine function`\ s concurrently,
+    :meth:`start_soon` is also provided for :keyword:`await` ing arbitrary :term:`awaitable`\ s concurrently.
```
cmarqu:
```suggestion
    :meth:`start_soon` is also provided for :keyword:`await`\ ing arbitrary :term:`awaitable`\ s concurrently.
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+        self._finishing: bool = False
+        self._entered: bool = False
+        # parent task will not exist if we aren't using this as a context manager
+        self._parent_task: Task[Any] | None = None
+
+        # Start with no remaining tasks
+        self._none_remaining.set()
+
+    def start_soon(self, aw: Awaitable[T], *, name: str | None = None) -> Task[T]:
+        """Awaits its argument concurrently to other calls to this function.
```
cmarqu:
```suggestion
        """Await the *aw* argument concurrently to other calls to this method.
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+
+    After exiting the context block and waiting for all child :class:`!Task`\ s to complete,
+    or :keyword:`await`\ ing the :class:`!TaskManager` instance the first time,
+    no further calls to :meth:`start_soon` or :deco:`!fork` are permitted;
+    attempting to do so will raise a :exc:`RuntimeError`.
+
+    Additionally, after a child :class:`!Task` fails and the :class:`!TaskManager` begins cancelling other child :class:`!Task`\ s,
+    no further calls to :meth:`start_soon` or :deco:`!fork` are permitted;
+    attempting to do so will raise a :exc:`RuntimeError`,
+    unless the ``continue_on_error`` parameter is ``True`` and the context block has not yet exited.
```
cmarqu:
```suggestion
    unless the *continue_on_error* parameter is ``True`` and the context block has not yet exited.
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+                    raise ValueError("An error occurred in task2")
+
+        except* ValueError as e:
+            # This will print both ValueErrors from task1 and task2
+            cocotb.log.info(f"Caught ValueError from TaskManager: {e}")
+
+    You are permitted to add any :keyword:`await` statement to the body of the context block.
+    This means that it is possible for child tasks to start running, and fail, before the context block is exited.
+    In this case, the context block will also exit with a :exc:`CancelledError` if the ``continue_on_error`` parameter is ``False`` (default),
+    or continue if ``continue_on_error`` is ``True``.
```
cmarqu:
```suggestion
    In this case, the context block will also exit with a :exc:`CancelledError` if the *continue_on_error* parameter is ``False`` (default),
    or continue if *continue_on_error* is ``True``.
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,387 @@
...
+
+T = TypeVar("T")
+
+
+async def _waiter(aw: Awaitable[T]) -> T:
+    return await aw
+
+
+class TaskManager:
+    r"""An :term:`asynchronous context manager` which enables the user to run :term:`coroutine function`\ s or :keyword:`await` :term:`awaitable`\ s concurrently and wait for them all to finish.
```
cmarqu: This docstring is almost a tutorial, wondering if its content should not be in the Library Reference but in e.g. `coroutines.html` instead...

ktbarrett:
Moved the docs to `coroutines.rst`.

docs/source/coroutines.rst
```diff
@@ -260,6 +260,180 @@ And of course, the sky is the limit when you compose the two.
...
+
+    async with TaskManager() as tm:
+        tm.start_soon(RisingEdge(cocotb.top.operation_complete))
+
+        @tm.fork
+        async def watchdog():
+            await Timer(1, "us")
+            raise TimeoutError("Operation did not complete in time")
+
+You can inspect the result of child classes by storing the :class:`!Task` objects returned by :meth:`start_soon` method.
```
cmarqu:
```suggestion
You can inspect the result of child classes by storing the :class:`!Task` objects returned by the :meth:`!start_soon` method.
```

docs/source/coroutines.rst
```diff
@@ -260,6 +260,180 @@ And of course, the sky is the limit when you compose the two.
...
+
+:class:`!TaskManager`: Structured Asymmetric Concurrency
+========================================================
+
+The :class:`~cocotb.task.TaskManager` class is another way to run multiple async routines concurrently and wait for them all to complete,
+but properly manages the lifetime of its "children" and handles exceptions and cancellations gracefully.
+This should be the preferred way to run multiple coroutines concurrently.
+
+The typical usage pattern is to create a :class:`!TaskManager` as an :term:`asynchronous context manager` using the :keyword:`async with` statement;
+then to use the :deco:`~cocotb.triggers.TaskManager.fork` decorator method to start :class:`!Task`\ s within the context block.
```
cmarqu: This somehow shows the hierarchy including the tilde in https://cocotb--5163.org.readthedocs.build/en/5163/coroutines.html#taskmanager-structured-asymmetric-concurrency

docs/source/coroutines.rst
```diff
@@ -260,6 +260,180 @@ And of course, the sky is the limit when you compose the two.
...
+        ...  # This code will be skipped
+
+.. warning::
+    Just like with :class:`~cocotb.task.Task`, cancelling a :class:`!TaskManager` instance
+    (after a child fails, or if another Task cancels the Task using the :class:`!TaskManager`)
+    and squashing the resulting :exc:`CancelledError` will cause the test to fail immediately.
+    Always remember to re-raise the :exc:`!CancelledError` if you catch it.
+
+In addition to the :deco:`!fork` method for starting :term:`coroutine function`\ s concurrently,
+:meth:`start_soon` is also provided for :keyword:`!await`\ ing arbitrary :term:`awaitable`\ s concurrently.
```
cmarqu:
```suggestion
:meth:`~cocotb.triggers.TaskManager.start_soon` is also provided for :keyword:`!await`\ ing arbitrary :term:`awaitable`\ s concurrently.
```
Hope this renders

docs/source/coroutines.rst
```diff
@@ -260,6 +260,180 @@ And of course, the sky is the limit when you compose the two.
...
+
+    # Control returns here when all drive Tasks have completed
+
+If a child :class:`!Task` raises an exception,
+one of two behaviors will occur depending on the value of the *continue_on_error* parameter passed to the :class:`!TaskManager` constructor.
+If the *continue_on_error* parameter is ``False`` (default), all other child :class:`!Task`\ s are cancelled and the :class:`!TaskManager` will begin shutting down.
+If the *continue_on_error* parameter is ``True``, the exception is captured and other child :class:`!Task`\ s are allowed to continue running.
+
+After all child :class:`!Task`\ s have finished,
+all exceptions, besides :exc:`CancelledError`, are gathered into an :exc:`ExceptionGroup`,
```
cmarqu:
```suggestion
all exceptions, besides :exc:`~asyncio.CancelledError`, are gathered into an :exc:`ExceptionGroup`,
```

docs/source/coroutines.rst
```diff
@@ -260,6 +260,180 @@ And of course, the sky is the limit when you compose the two.
     This means that Tasks passed to it are *still running*.
     You may need to cancel those Tasks with :meth:`.Task.cancel`.
 
+.. _task_manager_tutorial:
+
+:class:`!TaskManager`: Structured Asymmetric Concurrency
+========================================================
+
+The :class:`~cocotb.task.TaskManager` class is another way to run multiple async routines concurrently and wait for them all to complete,
```
cmarqu:
```suggestion
The :class:`~cocotb.triggers.TaskManager` class is another way to run multiple async routines concurrently and wait for them all to complete,
```

docs/source/coroutines.rst
```diff
@@ -260,6 +260,180 @@ And of course, the sky is the limit when you compose the two.
...
+                ...
+                raise ValueError("An error occurred in task2")
+
+    except* ValueError as e:
+        # This will print both ValueErrors from task1 and task2
+        cocotb.log.info(f"Caught ValueError from TaskManager: {e}")
+
+You are permitted to add any :keyword:`await` statement to the body of the context block.
+This means that it is possible for child tasks to start running, and end with an exception, before the context block has finished.
+In this case, a :exc:`CancelledError` will be raised from the current :keyword:`!await`
```
cmarqu:
```suggestion
In this case, a :exc:`~asyncio.CancelledError` will be raised from the current :keyword:`!await`
```

docs/source/coroutines.rst
```diff
@@ -260,6 +260,180 @@ And of course, the sky is the limit when you compose the two.
...
+This should be the preferred way to run multiple coroutines concurrently.
+
+The typical usage pattern is to create a :class:`!TaskManager` as an :term:`asynchronous context manager` using the :keyword:`async with` statement;
+then to use the :deco:`~cocotb.triggers.TaskManager.fork` decorator method to start :class:`!Task`\ s within the context block.
+When control reaches the end of the context block
+the :class:`!TaskManager` blocks the encompassing :class:`!Task` until all children :class:`!Task`\ s complete.
+
+.. code-block:: python
+
+    async with TaskManager() as tm:
```
cmarqu, line 289:
```suggestion
    from cocotb.triggers import TaskManager

    async with TaskManager() as tm:
```

docs/source/coroutines.rst
```diff
@@ -260,6 +260,180 @@ And of course, the sky is the limit when you compose the two.
...
+    assert task2.done()
+    assert task2.result() == 42
+
+After exiting the context block and waiting for all child :class:`!Task`\ s to complete,
+no further calls to :meth:`start_soon` or :deco:`!fork` are permitted.
+Attempting to do so will raise a :exc:`RuntimeError`.
+
+Additionally, after a child :class:`!Task` fails and the :class:`!TaskManager` begins cancelling other child :class:`!Task`\ s,
+no further calls to :meth:`start_soon` or :deco:`!fork` are permitted.
+Attempting to do so will raise a :exc:`RuntimeError`,
```
cmarqu:
```suggestion
Attempting to do so will raise a :exc:`!RuntimeError`,
```

docs/source/coroutines.rst
ktbarrett:
```suggestion
then to use the :deco:`@fork <cocotb.triggers.TaskManager.fork>` decorator method to start :class:`!Task`\ s within the context block.
```

docs/source/coroutines.rst
ktbarrett: I should add this to the prelude.

EDIT: Wait, it's a part of the prelude. We use `import *`. Why isn't it working?

ktbarrett: Okay, so it links correctly. Was there a reason for your suggestion?

cmarqu: Sorry for leading you down a wrong path... I was just suggesting it for the users to see where they have to get `TaskManager` from when we introduce it. (They might rightfully think `cocotb.task`.)

docs/source/coroutines.rst
```diff
@@ -260,6 +260,180 @@ And of course, the sky is the limit when you compose the two.
...
+
+:class:`!TaskManager`: Structured Asymmetric Concurrency
+========================================================
+
+The :class:`~cocotb.triggers.TaskManager` class is another way to run multiple async routines concurrently and wait for them all to complete,
+but properly manages the lifetime of its "children" and handles exceptions and cancellations gracefully.
+This should be the preferred way to run multiple coroutines concurrently.
+
+The typical usage pattern is to create a :class:`!TaskManager` as an :term:`asynchronous context manager` using the :keyword:`async with` statement;
+then to use the :deco:`@fork <cocotb.triggers.TaskManager.fork>` decorator method to start :class:`!Task`\ s within the context block.
```
cmarqu:
```suggestion
then to use the :deco:`fork <cocotb.triggers.TaskManager.fork>` decorator method to start :class:`!Task`\ s within the context block.
```
We now had a double-`@`.

ktbarrett: Interesting how some classes don't add their classes syntax, like you often have to do ```:meth:`thing() <path.to.thing>` ```

cmarqu: Yeah, it's all pretty inconsistent.

docs/source/coroutines.rst
```diff
@@ -260,6 +260,182 @@ And of course, the sky is the limit when you compose the two.
...
+                ...
+                raise ValueError("An error occurred in task2")
+
+    except* ValueError as e:
+        # This will print both ValueErrors from task1 and task2
+        cocotb.log.info(f"Caught ValueError from TaskManager: {e}")
+
+You are permitted to add any :keyword:`await` statement to the body of the context block.
+This means that it is possible for child tasks to start running, and end with an exception, before the context block has finished.
+In this case, a :exc:`~asyncio.CancelledError` will be raised from the current :keyword:`!await`
```
marlonjames:
```suggestion
In this case, a :exc:`~asyncio.CancelledError` will be raised from the current :keyword:`!await` expression
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,221 @@
...
+        .. code-block:: python
+
+            async with TaskManager() as tm:
+
+                @tm.fork
+                async def my_func():
+                    # Do stuff
+                    ...
+
+                @tb.fork
```
marlonjames:
```suggestion
                @tm.fork
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,221 @@
...
+        if self._exceptions:
+            # BaseExceptionGroup constructor will automatically return an ExceptionGroup if all elements are Exceptions.
+            raise BaseExceptionGroup(
+                "TaskManager finished with errors", tuple(self._exceptions)
+            )
+
+        # Return True to handle suppressing CancelledError if there were no exceptions
+        return True  # type: ignore[return-value]  # __aexit__ can return True
+
+    async def _finish_(self, exc: BaseException | None = None) -> None | Literal[True]:
```
marlonjames: Is this used anywhere?

tests/test_cases/test_cocotb/test_task_manager.py
```diff
@@ -0,0 +1,959 @@
...
+
+            async with TaskManager() as tm_inner:
+                tm_inner.start_soon(coro(5))
+
+                await Timer(1)
+                raise KeyboardInterrupt()
+
+
+# Can't test KeyboardInterrupt in child Task since that will take a path to shut down
+# the simulation.
```
marlonjames: Should have test cases for trying to add tasks before entering:

```py3
@cocotb.xfail(raises=RuntimeError, reason="Adding Task outside context")
@cocotb.parametrize(continue_on_error=[True, False])
async def test_start_soon_outside_context(_: object, continue_on_error: bool) -> None:
    tm = TaskManager(continue_on_error=continue_on_error)

    tm.start_soon(coro(1))


@cocotb.xfail(raises=RuntimeError, reason="Adding Task outside context")
@cocotb.parametrize(continue_on_error=[True, False])
async def test_fork_outside_context(_: object, continue_on_error: bool) -> None:
    async def coro_noargs():
        await Timer(1)

    tm = TaskManager(continue_on_error=continue_on_error)

    tm.fork(coro_noargs
```

Those don't work, but you can add tasks from outside the context once it's been entered:

```py3
@cocotb.test
@cocotb.parametrize(continue_on_error=[True, False])
async def test_add_tasks_from_another_task(_: object, continue_on_error: bool) -> None:
    async def tm_coro(tm: TaskManager, ev: Event) -> None:
        async with tm:
            await ev.wait()
        
        return 10
    
    tm = TaskManager(continue_on_error=continue_on_error)
    ev = Event()
    tm_task = cocotb.start_soon(tm_coro(tm, ev))
    
    await NullTrigger()

    tm.start_soon(coro(1))

    ev.set()

    await tm_task
    assert tm_task.result() == 10
```
Should we do a check in `TaskManager._start_soon()` to see if `current_task()` is the same as `_parent_task` to prevent that?

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,221 @@
...
+        self._cancelled: bool = False
+        self._finishing: bool = False
+        self._entered: bool = False
+        # parent task will not exist if we aren't using this as a context manager
+        self._parent_task: Task[Any] | None = None
+
+        # Start with no remaining tasks
+        self._none_remaining.set()
+
+    def start_soon(self, aw: Awaitable[T], *, name: str | None = None) -> Task[T]:
```
marlonjames: What do you think about a per-child-task `continue_on_error` argument here? So you could by default have child task failures cause a full cancel, but you could exempt certain child tasks (or inner `TaskManager`s) from that and defer handling errors until context exit? It was discussed in #3583.

ktbarrett:
@marlonjames Added per-Task `continue_on_error`. Still need to add tests. I changed the `fork` method to support this with the following. `start_soon` just take an additional keyword argument.

```python
async with TaskManager() as tm:
    @tm.fork
    async def thing(): ...

    @tm.fork(continue_on_error=True)
    async def other(): ...

    tm.start_soon(lol)
```

~~I also added support for fork to take arbitrary args and kwargs so it can be used more like trio's nursery.~~
```python
async with TaskManager() as tm:
    task = tm.fork(coro_func, 1, 2, key=3)
```
EDIT: I removed this. We can add it later if people find it valuable.

Finally, I refactored some things so that it doesn't create the coroutine object until after the checks in `_start_soon` are passed so we don't cause spurious ResourceWarnings.

src/cocotb/task.py
```diff
@@ -533,14 +533,25 @@ def _cancel_now(self, msg: str | None = None) -> bool:
...
         # Throw CancelledError now
         self._exc = cancelled_error
         self._resume()
 
         return True
 
+    def _cancelling(self) -> int:
+        """Return the number of cancellation requests if the Task is in the process of being cancelled."""
+        return self._must_cancel
+
```
```suggestion
```
No longer used.

docs/source/coroutines.rst
```diff
@@ -260,6 +260,182 @@ And of course, the sky is the limit when you compose the two.
     This means that Tasks passed to it are *still running*.
     You may need to cancel those Tasks with :meth:`.Task.cancel`.
 
+.. _task_manager_tutorial:
+
+:class:`!TaskManager`: Structured Asymmetric Concurrency
+========================================================
+
+The :class:`~cocotb.triggers.TaskManager` class is another way to run multiple async routines concurrently and wait for them all to complete,
+but properly manages the lifetime of its "children" and handles exceptions and cancellations gracefully.
```
cmarqu:
```suggestion
The :class:`~cocotb.triggers.TaskManager` class is another way to run multiple async routines concurrently and wait for them all to complete.
It properly manages the lifetime of its "children" and handles exceptions and cancellations gracefully.
```
If the "but" was to contrast this part with the section above, maybe make it more explicit.

docs/source/coroutines.rst
```diff
@@ -260,6 +260,182 @@ And of course, the sky is the limit when you compose the two.
...
+This should be the preferred way to run multiple coroutines concurrently.
```
cmarqu: Should this section be more prominent then, i.e. come earlier?

docs/source/coroutines.rst
```diff
@@ -260,6 +260,182 @@ And of course, the sky is the limit when you compose the two.
...
+
+The typical usage pattern is to create a :class:`!TaskManager` as an :term:`asynchronous context manager` using the :keyword:`async with` statement;
+then to use the :deco:`fork <cocotb.triggers.TaskManager.fork>` decorator method to start :class:`!Task`\ s within the context block.
+When control reaches the end of the context block
+the :class:`!TaskManager` blocks the encompassing :class:`!Task` until all children :class:`!Task`\ s complete.
```
cmarqu:
```suggestion
the :class:`!TaskManager` blocks the encompassing :class:`!Task` until all child :class:`!Task`\ s complete.
```
Not sure, but this is how you use it below the code example.

docs/source/coroutines.rst
```diff
@@ -260,6 +260,182 @@ And of course, the sky is the limit when you compose the two.
...
+            )
+            raise  # DON'T FORGET THIS
+
+        ...  # This code will be skipped
+
+.. warning::
+    Just like with :class:`~cocotb.task.Task`, cancelling a :class:`!TaskManager` instance
+    (after a child fails, or if another Task cancels the Task using the :class:`!TaskManager`)
+    and squashing the resulting :exc:`CancelledError` will cause the test to fail immediately.
+    Always remember to re-raise the :exc:`!CancelledError` if you catch it.
```
cmarqu:
```suggestion
    and squashing the resulting :exc:`!asyncio.CancelledError` will cause the test to fail immediately.
    Always remember to re-raise the :exc:`!asyncio.CancelledError` if you catch it.
```

docs/source/coroutines.rst
```diff
@@ -260,6 +260,182 @@ And of course, the sky is the limit when you compose the two.
...
+            @tm.fork
+            async def task2(): ...
+
+    except* ValueError as e:
+        cocotb.log.info(f"Caught ValueError from task1: {e}")
+        pass
+
+    assert task2.done()  # task2 was able to run to completion
+
+And of course, :class:`!TaskManager` can be arbitrarily nested.
```
cmarqu:
```suggestion
As shown below, :class:`!TaskManager` can be arbitrarily nested.
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,299 @@
...
+        *,
+        name: str | None = None,
+        continue_on_error: bool | None = None,
+    ) -> Task[T]:
+        """Await the *aw* argument concurrently.
+
+        Args:
+            aw: A :class:`~collections.abc.Awaitable` to :keyword:`await` concurrently.
+            name: A name to associate with the :class:`!Task` awaiting *aw*.
+            continue_on_error: override the TaskManager's continue_on_error setting for this Task only.
```
cmarqu:
```suggestion
            continue_on_error: Override the :class:`!TaskManager`\ 's ``continue_on_error`` setting for this Task only.
```

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,299 @@
...
+        coro_func: Callable[..., Coroutine[Trigger, None, T]] | None = None,
+        *,
+        continue_on_error: bool | None = None,
+    ) -> Task[T] | Callable[[Callable[[], Coroutine[Trigger, None, T]]], Task[T]]:
+        """Decorate a coroutine function to run it concurrently.
+
+        Args:
+            coro_func: A :term:`coroutine function` to run concurrently. Typically only passed as a decorator.
+            continue_on_error:
+                Override the TaskManager's ``continue_on_error`` argument for this Task only.
```
cmarqu:
```suggestion
                Override the :class:`!TaskManager`\ 's ``continue_on_error`` argument for this Task only.
```

src/cocotb/task.py
```diff
@@ -533,14 +533,21 @@ def _cancel_now(self, msg: str | None = None) -> bool:
...
+        self._must_cancel += 1
 
         # Throw CancelledError now
         self._exc = cancelled_error
         self._resume()
 
         return True
 
+    def _uncancel(self) -> None:
+        """Prevents RuntimeError from occurring when cancelled Tasks suppress CancelledError.
```
cmarqu:
```suggestion
        """Prevents :exc:`RuntimeError` from occurring when cancelled Tasks suppress :exc:`asyncio.CancelledError`.
```
Couldn't help it :)

docs/source/coroutines.rst
ktbarrett: Actually since the above section will be rewritten to use `gather` and `select` which don't have issues with lifetimes, I'll rewrite this.

docs/source/coroutines.rst
ktbarrett: No. The above section will be rewritten to use `gather` and `select` which are simpler and more common uses.

ktbarrett:
Oooo, should the constructor's `continue_on_error` also act as the per-Task `continue_on_error` for the context block (parent Task)? I think I need just a bit more work/tests.

ktbarrett:
Split constructor `continue_on_error` into `default_continue_on_error` that defaults the value for all child tasks and `context_continue_on_error` which is the continue_on_error for the context block if there's a non-external-cancellation.

@cmarqu I also did a significant refactor of the docs in coroutines.rst.

docs/source/coroutines.rst
```diff
@@ -260,6 +260,228 @@ And of course, the sky is the limit when you compose the two.
...
+    Attempting to do so will raise a :exc:`RuntimeError`.
+
+Handling Exceptions and *continue_on_error*
+-------------------------------------------
+
+:class:`!TaskManager` gracefully handles exceptions raised in child :class:`!Task`\ s or in the context block itself.
+It ensures that no child :class:`!Task` is left running unintentionally by the time the context block exits.
+
+The behavior of :class:`!TaskManager` when a child :class:`!Task` raises an exception is controlled by the *continue_on_error* parameter.
+The constructor for :class:`!TaskManager` accepts an optional parameter *continue_on_error* which is used as the default for all children Tasks;
```
cmarqu:
```suggestion
The constructor for :class:`!TaskManager` accepts an optional parameter *default_continue_on_error* which is used as the default for all children Tasks;
```

ktbarrett:
noxfile.py `release_install` needs the new dependency added.

docs/source/coroutines.rst
```diff
@@ -260,6 +260,228 @@ And of course, the sky is the limit when you compose the two.
...
     You may need to cancel those Tasks with :meth:`.Task.cancel`.
 
+.. _task_manager_tutorial:
+
+:class:`!TaskManager`
+=====================
+
+The :class:`~cocotb.triggers.TaskManager` class is another way to run multiple async routines concurrently and wait for them all to complete.
+It properly manages the lifetime of its "children" and handles exceptions and cancellations gracefully.
+Unlike :func:`gather` which takes all :term:`awaitable`\ s and :term:`coroutine`\ s at once,
```
marlonjames: `gather()` not linking in docs.

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,313 @@
...
+                f"start_soon() expected an Awaitable, got {type(aw).__name__}"
+            )
+        return self._start_soon(
+            _waiter, aw, name=name, continue_on_error=continue_on_error
+        )
+
+    def _start_soon(
+        self,
+        coro_func: Callable[..., Coroutine[Trigger, None, T]],
+        arg: Any | None = _MISSING,
```
marlonjames: Do you want to keep `arg`? It's not used.

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,313 @@
...
+        if current_task() is not self._parent_task:
+            raise RuntimeError("Cannot add new Tasks to TaskManager from another Task")
+
+        if arg is _MISSING:
+            coro = coro_func()
+        else:
+            coro = coro_func(arg)
+        try:
+            task = Task[Any](coro, name=name)
+        except Exception:
```
marlonjames: Is there an actual case where an object gets through `fork()` or `start_soon()` and fails here?

src/cocotb/_task_manager.py
ktbarrett: `arg` is used by `start_soon`.

src/cocotb/_task_manager.py
ktbarrett: Not anymore with the `inspect.iscoroutinefunction` guard in `fork`. Good catch.

ktbarrett:
So I changed the interface around a bit so that `_start_soon` now takes Tasks and each caller is in charge of cleaning up newly-started coros if we aren't in a good place to add tasks.

Some additional considerations. Once `RunningTest` (will be renamed to `TestManager`) is refactored to use `TaskManager` we will get `continue_on_error` support to `cocotb.start_soon` implicitly.

src/cocotb/_task_manager.py
```diff
@@ -0,0 +1,298 @@
...
+T = TypeVar("T")
+
+
+async def _waiter(aw: Awaitable[T]) -> T:
+    return await aw
+
+
+_MISSING = object()
+
+
```
marlonjames: Not used anymore.
```suggestion
```

PR #5195, opened 2025-12-13
Prevent pytest from picking up cocotb.test decorator as a test

ktbarrett:


imphil:
Oh that's actually neat.

PR #5197, opened 2025-12-14
Add doc section Available Triggers for Testbenches

### LINKED ISSUE #5191: Add table of available triggers in Triggers howto page
We should add a table of the available triggers to the https://docs.cocotb.org/en/stable/triggers.html page.

cmarqu:
Closes #5191

Rendered page: https://cocotb--5197.org.readthedocs.build/en/5197/triggers.html

### DIFF
--- a/docs/source/triggers.rst
+++ b/docs/source/triggers.rst
@@ -3,15 +3,62 @@ Triggers
 ********
 
 Triggers are used to indicate when the cocotb scheduler should resume coroutine execution.
+
 To use a trigger, a coroutine should :keyword:`await` it.
 This will cause execution of the current coroutine to pause.
-When the trigger fires, execution of the paused coroutine will resume:
+When the trigger fires
+(:class:`~cocotb.triggers.RisingEdge` in the following example),
+execution of the paused coroutine will resume:
 
 .. code-block:: python
+
+    from cocotb.triggers import RisingEdge
 
     async def coro():
         print("Some time before a clock edge")
         await RisingEdge(clk)
         print("Immediately after the rising clock edge")
 
-See :ref:`triggers` for a list of available triggers.
+
+Available Triggers for Testbenches
+==================================
+
+Below is a table of triggers that are useful for writing testbenches and models.
+
+For a list of *all* available triggers,
+see the :ref:`triggers` section in the :doc:`Python Code Library Reference <library_reference>`.
+
+..
+   Please keep this table aligned with the content in library_reference.rst
+
++------------------------------------------+-------------------------------------------------------------------------------------------+
+| **Edge Triggers**                                                                                                                    |
++------------------------------------------+-------------------------------------------------------------------------------------------+
+| | :class:`~cocotb.triggers.RisingEdge`   | | Resume after a *rising* edge on a single-bit signal                                     |
+| | :class:`~cocotb.triggers.FallingEdge`  | | Resume after a *falling* edge on a single-bit signal                                    |
+| | :class:`~cocotb.triggers.ValueChange`  | | Resume after *any* edge on a signal                                                     |
+| | :class:`~cocotb.triggers.ClockCycles`  | | Resume after the specified number of transitions of a signal                            |
+| | :class:`~cocotb.triggers.Edge`         | | (deprecated, see :class:`~cocotb.triggers.ValueChange`)                                 |
++------------------------------------------+-------------------------------------------------------------------------------------------+
+| **Timing Triggers**                                                                                                                  |
++------------------------------------------+-------------------------------------------------------------------------------------------+
+| | :class:`~cocotb.triggers.Timer`        | | Resume after the specified time                                                         |
+| | :class:`~cocotb.triggers.ReadOnly`     | | Resume when the simulation timestep moves to the *read-only* phase                      |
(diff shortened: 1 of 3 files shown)


docs/source/triggers.rst
```diff
@@ -3,15 +3,62 @@ Triggers
...
+| **Concurrency Triggers**                                                                                                             |
++------------------------------------------+-------------------------------------------------------------------------------------------+
+| | :func:`~cocotb.triggers.gather`        | | Resume after *all* given Tasks or Triggers                                              |
+| | :func:`~cocotb.triggers.select`        | | Resume after *any* given Tasks or Triggers                                              |
+| | :func:`~cocotb.triggers.wait`          | | Await on all given Tasks or Triggers concurrently and block until a condition is met    |
+| | :class:`~cocotb.triggers.NullTrigger`  | | Resume immediately                                                                      |
+| | :class:`~cocotb.triggers.Combine`      | | (deprecated, see :func:`~cocotb.triggers.gather`)                                       |
+| | :class:`~cocotb.triggers.First`        | | (deprecated, see :func:`~cocotb.triggers.select`)                                       |
```
ktbarrett: These are not necessarily deprecated. They still have value is producing things that look and act like triggers that are a combination of other triggers. They are deprecated only for passing Tasks.

cmarqu: Amended, but probably not in the best way. Suggestions appreciated.

Actually, should `gather` and `select` be weaved into https://docs.cocotb.org/en/development/coroutines.html#waiting-for-multiple-events-simultaneously ?

ktbarrett: Yep I opened an issue to update as a follow on. I think there should be a section on `First` and `Combine` in this document and coroutines.rst should be updated to only use `gather` and `select`.

docs/source/triggers.rst
```diff
@@ -3,15 +3,62 @@ Triggers
...
+| | :class:`~cocotb.triggers.ReadOnly`     | | Resume when the simulation timestep moves to the *read-only* phase                      |
+| | :class:`~cocotb.triggers.ReadWrite`    | | Resume when the simulation timestep moves to the *read-write* phase                     |
+| | :class:`~cocotb.triggers.NextTimeStep` | | Resume when the simulation timestep moves to the *next time step*                       |
++------------------------------------------+-------------------------------------------------------------------------------------------+
+| **Concurrency Triggers**                                                                                                             |
++------------------------------------------+-------------------------------------------------------------------------------------------+
+| | :func:`~cocotb.triggers.gather`        | | Resume after *all* given Tasks or Triggers                                              |
+| | :func:`~cocotb.triggers.select`        | | Resume after *any* given Tasks or Triggers                                              |
```
ktbarrett, line 53: Technically these aren't triggers. Maybe these shouldn't be listed here and instead used in coroutines.rst only.

docs/source/triggers.rst
```diff
@@ -3,15 +3,62 @@ Triggers
...
+| | :func:`~cocotb.triggers.wait`          | | Await on all given Tasks or Triggers concurrently and block until a condition is met    |
+| | :class:`~cocotb.triggers.NullTrigger`  | | Resume immediately                                                                      |
++------------------------------------------+-------------------------------------------------------------------------------------------+
+| **Synchronization Triggers**                                                                                                         |
++------------------------------------------+-------------------------------------------------------------------------------------------+
+| | :class:`~cocotb.triggers.Event`        | | A way to signal an event across Tasks                                                   |
+| | :class:`~cocotb.triggers.Lock`         | | A mutual exclusion lock                                                                 |
+| | :func:`~cocotb.triggers.with_timeout`  | | Resume latest at the specified timeout time                                             |
```
ktbarrett, line 63: This is also not really a trigger.

docs/source/triggers.rst
```diff
@@ -3,15 +3,62 @@ Triggers
...
+| **Concurrency Triggers**                                                                                                             |
++------------------------------------------+-------------------------------------------------------------------------------------------+
+| | :func:`~cocotb.triggers.gather`        | | Resume after *all* given Tasks or Triggers                                              |
+| | :func:`~cocotb.triggers.select`        | | Resume after *any* given Tasks or Triggers                                              |
+| | :class:`~cocotb.triggers.Combine`      | | Resume after *all* given Triggers                                                       |
+| | :class:`~cocotb.triggers.First`        | | Resume after *any* given Triggers                                                       |
+| | :func:`~cocotb.triggers.wait`          | | Await on all given Tasks or Triggers concurrently and block until a condition is met    |
+| | :class:`~cocotb.triggers.NullTrigger`  | | Resume immediately                                                                      |
```
ktbarrett, line 57:
```suggestion
| | :class:`~cocotb.triggers.NullTrigger`  | | Resume soon                                                                             |
```

cmarqu:
> The API is a bit of a mess. Not everything in `cocotb.triggers` is a Trigger. It's almost like `cocotb.triggers` is just cocotb's name for it's concurrency library.

And maybe the naming of `triggers.rst` to be about triggers is not all that useful? How about we dissolve it and rename "Coroutines and Tasks" to "Coroutines, Tasks and Triggers" or something like that?
The table of testbench-useful whatevers could be at the beginning of it, a bit like in https://docs.python.org/3/library/functions.html#built-in-functions

ktbarrett:
That sounds good to me.

cmarqu:
Superseded by #5204

PR #5205, opened 2025-12-19
Support any `Awaitable` in `with_timeout`

### LINKED ISSUE #5009: Support any Awaitable in `with_timeout`
Currently `with_timeout` is limited to Tasks, coroutines, Triggers, and Waitables. It doesn't have the same issues that `First` and `Combine` do, so it could be expanded to support any `Awaitable` with a bit of effort. We could piggyback off of the implementation of `coconext.triggers.wait`.

ktbarrett:
Closes #5009. Reimplement `with_timeout` using `select`. Allows us to support any `Awaitable` and not just `Trigger`s, `Task`s and `Coroutine`s.

### DIFF
--- a/src/cocotb/_concurrent_waiters.py
+++ b/src/cocotb/_concurrent_waiters.py
@@ -9,8 +9,8 @@ from asyncio import CancelledError
 from typing import TYPE_CHECKING, Any, Literal, TypeVar, overload
 
 import cocotb
+from cocotb._base_triggers import Event
 from cocotb.task import Task
-from cocotb.triggers import Event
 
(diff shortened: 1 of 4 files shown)


src/cocotb/_concurrent_waiters.py
```diff
@@ -9,8 +9,8 @@
 from typing import TYPE_CHECKING, Any, Literal, TypeVar, overload
 
 import cocotb
+from cocotb._base_triggers import Event
```
ktbarrett, line 12: Necessary to prevent cyclic imports.

Issue #5192, opened 2025-12-11
Document `TaskManager` and `gather`/`select` in concurrency documentation


ktbarrett:
The section on concurrency in the how-to documentation should be changed to use `gather`/`select` instead of `First`/`Combine` for tasks, and a section on `TaskManager` should be added.

PR #5206, opened 2025-12-19
Deprecate passing Tasks to `First` and `Combine`

### LINKED ISSUE #5193: Deprecate passing Tasks to `First` and `Combine`
`First` and `Combine` being triggers are thing that can be awaited multiple times. Tasks are not repeatable actions, so they don't really make much sense to use here. The only reason it was originally supported was because `First` and `Combine` were the only way to accomplish concurrency. `gather` and `select` are more appropriate alternatives as they are fundamentally "one-shot" and have a better API for concurrency (taking coroutines directly, cleaning up loose tasks, returning results).

ktbarrett:
Depends on #5205. Closes #5193. xref #5192.

First and Combine are Triggers. Triggers can be awaited multiple times
to re-prime whatever condition they represent. Tasks wrap coroutines and
can only run once, so there is an impedance mismatch. In operation,
what's happening is the First/Combine are re-awaiting the Task's
completion trigger. Repeated uses of the First and Combine will consume
no time once the Task completes.

This behavior was only supported since there needed to be some way to
wait for multiple concurrent coroutines to joined. But now that
gather/select are available, they do the job *much* better and avoid the
limitations of using Tasks in First/Combine.

### TODO
- [x] newsfragment


docs/source/newsfragments/5206.change.rst
```diff
@@ -0,0 +1 @@
+Deprecated passing :class:`~cocotb.task.Task`\ s to :class:`~cocotb.triggers.First` and :class:`!cocotb.triggers.Combine` in favor of passing coroutines directly to :class:`~cocotb.triggers.select` and :func:`~cocotb.triggers.gather`, respectively.
```
cmarqu:
```suggestion
Deprecated passing :class:`~cocotb.task.Task`\ s to :class:`~cocotb.triggers.First` and :class:`~cocotb.triggers.Combine` in favor of passing coroutines directly to :class:`~cocotb.triggers.select` and :func:`~cocotb.triggers.gather`, respectively.
```

src/cocotb/_extended_awaitables.py
```diff
@@ -129,16 +134,24 @@ def on_done(
...
+        # no _pointer_str here, since this is not a trigger, so identity
+        # doesn't matter.
+        return "{}({})".format(
+            type(self).__qualname__,
+            ", ".join(repr(t) for t in self._triggers),
+        )
 
-class First(_AggregateWaitable[object]):
+
+class First(Waitable[object]):
```
cmarqu, line 150: We should add a `.. versionchanged:: 2.1 ...` block to this docstring.

src/cocotb/_extended_awaitables.py
```diff
@@ -37,48 +39,51 @@ def __await__(self) -> Generator[Trigger, None, T]:
...
-            ", ".join(repr(t) for t in self._triggers),
-        )
-
-
 async def _wait_callback(trigger: Awaitable[T]) -> T:
     return await trigger
 
 
-class Combine(_AggregateWaitable["Combine"]):
+class Combine(Waitable["Combine"]):
```
cmarqu, line 46: We should add a `.. versionchanged:: 2.1 ...` block to this docstring.

cmarqu:
Sorry, I had the review queued but forgot to send it.

docs/source/newsfragments/5206.removal.rst
```diff
@@ -0,0 +1 @@
+Deprecated passing :class:`~cocotb.task.Task`\ s to :class:`~cocotb.triggers.First` and :class:`~cocotb.triggers.Combine` in favor of passing coroutines directly to :class:`~cocotb.triggers.select` and :func:`~cocotb.triggers.gather`, respectively.
```
marlonjames: I think this should be a `removal` newsfrag to go into the **Deprecations and Removals** section.

src/cocotb/_extended_awaitables.py
```diff
@@ -158,12 +175,48 @@ class First(_AggregateWaitable[object]):
...
-    def __init__(self, *trigger: Trigger | Waitable[Any] | Task[Any]) -> None:
-        if not trigger:
+    _task_deprecation_str = (
+        "Passing Task objects to `First` is deprecated. "
+        "Use `select` instead and pass coroutines directly instead of wrapping them in `cocotb.start_soon`."
+    )
+
+    @overload
+    def __init__(
+        self, first: Trigger | Waitable[Any], /, *trigger: Trigger | Waitable[Any]
```
marlonjames: ?
```suggestion
        self, first: Trigger | Waitable[Any], /, *triggers: Trigger | Waitable[Any]
```

docs/source/newsfragments/5206.removal.rst
marlonjames: So it shows up with `()` like gather.
```suggestion
Deprecated passing :class:`~cocotb.task.Task`\ s to :class:`~cocotb.triggers.First` and :class:`~cocotb.triggers.Combine` in favor of passing coroutines directly to :func:`~cocotb.triggers.select` and :func:`~cocotb.triggers.gather`, respectively.
```

PR #5207, opened 2025-12-20
Reimplement `LIBPYTHON_LOC` using `GPI_USERS`

ktbarrett:
Maintained backwards compatibility when using the Makefiles or Python Runner by translating LIBPYTHON_LOC into GPI_USERS. This removes all references reference to Python from the core GPI library.

### TODO
- [x] newsfrag for breaking custom flows


src/cocotb_tools/makefiles/Makefile.inc
```diff
@@ -111,9 +112,11 @@ ifneq ($(FIND_LIBPYTHON_RC),0)
     $(error $(find_libpython_errmsg))
 endif
 
+GPI_USERS := $(LIBPYTHON_LOC);
 endif
-
-export LIBPYTHON_LOC
+GPI_USERS := $(GPI_USERS)$(shell $(PYTHON_BIN) -m cocotb_tools.config --pygpi-entry-point)
+endif
+export GPI_USERS
```
marlonjames: I think remove line 115 and change 117 to
`GPI_USERS := $(LIBPYTHON_LOC)$(shell $(PYTHON_BIN) -m cocotb_tools.config --pygpi-entry-point)`.

Otherwise if `GPI_USER` isn't defined but `LIBPYTHON_LOC` is, it won't use the value of `LIBPYTHON_LOC`.

docs/source/library_reference_c.rst
```diff
@@ -38,12 +38,6 @@ Environment Variables
     You can get the default PyGPI entry point at other times by calling ``cocotb-config --pygpi-entry-point`` from the shell
```
marlonjames, line 38: Maybe update "defaults to load the PyGPI entry point." above to something like "defaults to load the Python library followed by the PyGPI entry point."

docs/source/newsfragments/5207.change.rst
```diff
@@ -0,0 +1 @@
+The ``LIBPYTHON_LOC`` environment variable was removed from the GPI. It is now a feature of the Makefiles (:make:var:`LIBPYTHON_LOC`) and Python Runner (:envvar:`LIBPYTHON_LOC`) for backwards-compatibility only.
```
marlonjames: I think a bit more explanation is warranted.
```suggestion
The ``LIBPYTHON_LOC`` environment variable was removed from the GPI. Instead, the Python library is included in ``GPI_USERS`` by default. Support remains in Makefiles (:make:var:`LIBPYTHON_LOC`) and Python Runner (:envvar:`LIBPYTHON_LOC`) for backwards-compatibility only.
```

PR #5208, opened 2025-12-20
Link pygpi against libpython, remove find_libpython dependency

ktbarrett:
Embedding libraries should link against libpython. The path to the libpython is not used to set up sys.path, so it doesn't matter if we choose the simulator's libpython over the system's or environments. As long as the Python version matches it should be good.

### TODO
- [ ] newsfrag for removing `--libpython` from `cocotb-config`
- [ ] newsfrag for mooting `LIBPYTHON_LOC`

### DIFF
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -435,42 +435,6 @@ class build_ext(_build_ext):
                 )
 
 
-def _get_python_lib_link():
-    """Get name of python library used for linking"""
-
-    if sys.platform == "darwin":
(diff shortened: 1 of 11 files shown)


ktbarrett:
Maybe I should be using something from sysconfig to computer the linked library name. Maybe one of those deleted functions does it properly.

ktbarrett:
I think user environments are enough of a nightmare that doing this isn't reasonable until we can also package libpython with cocotb or as another package with a dependency.

PR #5211, opened 2025-12-21
GPI: Split VHPI files

marlonjames:
I followed the VPI approach.


ktbarrett:
Could you follow the same approach as was done in the VPI to prevent losing git history? https://stackoverflow.com/questions/3887736/keep-git-history-when-splitting-a-file

It's just a touch laborious, but worth it IMO. I'll make sure this gets priority in merging to prevent and weird merge history.

PR #5218, opened 2025-12-25
GPI logging refactors

ktbarrett:
This is part of making the GPI a more effective and stable API so that we can reuse it in other projects. Follow-ons will:
* Simplify the API so there aren't a million functions and instead make it more like the VPI/VHPI with simple getters and enum arguments
* Refactor the API so that every function can actually report errors
* Refactoring startup and shutdown and separate cleanup and unexpected shutdown into two functions

We decided to make it so that the GPI doesn't provide logging to users, it only provides a way to allow users to hook the logging into their own system. This PR fully realizes that by moving the now private parts of the GPI logging into a source-only header and moving the public parts into the public GPI header.

This required some refactoring.
* Replaced uses of `gpi_log_filtered` with check of `gpi_debug_enabled`
* Removed dependency-injected log filter and log set level functionality and reimplemented that functionality in PyGPI. 
    * This was done because the GPI doesn't need that functionality, only users do, and we don't want to provide the log API to users anymore.
* Split `gpi_logging.h` between `gpi.h` and new private `logging.hpp` file
* Removed dependency on GPI native logger in PyGPI logger
    * Changed GPI logger default to use references to the native logger functions rather than `nullptr`
    * Split PyGPI logging configuration into two functions so pre-Python-initialization we can call into the "fallback logger"
    * This prevents us from explicitly calling out the native logger and instead getting it via the first call to `get_gpi_log_handler` we get during the call to `pygpi_logging_initialize`

Basically we are using abstraction to get the native logger rather than directly call it out, breaking the dependency and allowing us to remove all of the logging functionality from the GPI API.

The only thing left of the GPI logging API is `gpi_native_logger_set_level` which remains in case users don't register a log handler and instead just want to control verbosity.

### DIFF
--- a/src/cocotb/share/include/gpi.h
+++ b/src/cocotb/share/include/gpi.h
@@ -551,10 +551,8 @@ typedef void (*gpi_log_handler_ftype)(void *userdata, const char *name,
                                       const char *msg, va_list args);
 
 /** Retrieve the current log handler.
- * @param handler   Location to return current log handler function. If no
- *                  custom logger is registered this will be `NULL`.
- * @param userdata  Location to return log handler userdata. If no custom
- *                  logger is registered this will be `NULL`.
+ * @param handler   Location to return current log handler function.
+ * @param userdata  Location to return log handler userdata.
  */
 GPI_EXPORT void gpi_get_log_handler(gpi_log_handler_ftype *handler,
                                     void **userdata);


src/cocotb/share/include/gpi.h
```diff
@@ -502,6 +502,81 @@ GPI_EXPORT void gpi_get_cb_info(gpi_cb_hdl cb_hdl, int (**cb_func)(void *),
...
+                     ///< `INFO`.
+    GPI_TRACE = 5,   ///< Prints `TRACE` by default. Information about execution
+                     ///< of simulator callbacks and Python/simulator contexts.
+    GPI_DEBUG = 10,  ///< Prints `DEBUG` by default. Verbose information, useful
+                     ///< for debugging.
+    GPI_INFO = 20,   ///< Prints `INFO` by default. Information about major
+                     ///< events in the current program.
+    GPI_WARNING = 30,  ///< Prints `WARN` by default. Encountered a recoverable
+                       ///< bug, or information about surprising behavior.
+    GPI_ERROR = 40,    ///< Prints `ERROR` by default. An unrecoverable error
```
cmarqu:
```suggestion
    GPI_ERROR = 40,    ///< Prints `ERROR` by default. An unrecoverable error.
```

src/cocotb/share/include/gpi.h
```diff
@@ -502,6 +502,81 @@ GPI_EXPORT void gpi_get_cb_info(gpi_cb_hdl cb_hdl, int (**cb_func)(void *),
...
+    GPI_CRITICAL = 50  ///< Prints `CRITICAL` by default. An unrecoverable
+                       ///< error, to be followed by immediate simulator
+                       ///< shutdown.
+};
+
+/** Type of a logger handler function.
+ * @param userdata  private implementation data registered with this function
```
cmarqu:
```suggestion
 * @param userdata  Private implementation data registered with this function
```

src/cocotb/share/lib/gpi/logging.hpp
```diff
@@ -0,0 +1,129 @@
+// Copyright cocotb contributors
+// Copyright (c) 2013 Potential Ventures Ltd
+// Copyright (c) 2013 SolarFlare Communications Inc
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef COCOTB_GPI_LOGGING_H_
+#define COCOTB_GPI_LOGGING_H_
+
+/** @file gpi_logging.h
```
cmarqu:
```suggestion
#ifndef COCOTB_LOGGING_HPP_
#define COCOTB_LOGGING_HPP_

/** @file logging.hpp
```

src/cocotb/share/lib/gpi/logging.hpp
```diff
@@ -0,0 +1,129 @@
...
+    gpi_vlog_(MAKE_LOG_NAME_(extra_name), level, pathname, funcname, lineno, \
+              msg, args)
+
+// Don't call this function directly unless the name is "gpi" or starts with
+// "gpi."
+GPI_EXPORT void gpi_vlog_(const char *name, enum gpi_log_level level,
+                          const char *pathname, const char *funcname,
+                          long lineno, const char *msg, va_list args);
+
+#endif /* COCOTB_GPI_LOGGING_H_ */
```
cmarqu:
```suggestion
#endif /* COCOTB_LOGGING_HPP_ */
```

src/cocotb/share/lib/gpi/logging.hpp
```diff
@@ -0,0 +1,129 @@
...
+#ifndef COCOTB_GPI_LOGGING_H_
+#define COCOTB_GPI_LOGGING_H_
+
+/** @file gpi_logging.h
+
+GPI Logging
+===========
+
+This header file defines how to produce logs for GPI implementations
+as well as users of the GPI.
```
cmarqu:
```suggestion
as well as for users of the GPI.
```

cmarqu (commented):
We should also add the "Logging" group to https://github.com/cocotb/cocotb/blob/cb191c2f080d7ed752144b7e252cec6b9e1d1157/docs/source/library_reference_c.rst?plain=1#L70

cmarqu (approved):

src/cocotb/share/include/gpi.h
```diff
@@ -502,6 +502,81 @@ GPI_EXPORT void gpi_get_cb_info(gpi_cb_hdl cb_hdl, int (**cb_func)(void *),
...
+ */
+typedef void (*gpi_log_handler_ftype)(void *userdata, const char *name,
+                                      enum gpi_log_level level,
+                                      const char *pathname,
+                                      const char *funcname, long lineno,
+                                      const char *msg, va_list args);
+
+/** Retrieve the current log handler.
+ * @param handler   Location to return current log handler function. If no
+ *                  custom logger is registered this will be `NULL`.
```
marlonjames: Is this true? Looks like it'll use the GPI native logger by default.

marlonjames (approved):

PR #5221, opened 2025-12-27
CI: Update Riviera Pro to 2025.10

imphil:


imphil:
Logs show that CI is using the new version.

PR #5222, opened 2025-12-28
Added `wrap` argument to `LogicArray.from_(un)signed`

ktbarrett:
Added an wrap which automatically wraps the value passed into `from_unsigned` or `from_signed` to force it to fit in the given number of bits.

### DIFF
--- a/src/cocotb/types/_logic_array.py
+++ b/src/cocotb/types/_logic_array.py
@@ -342,15 +342,26 @@ class LogicArray(AbstractMutableArray[Logic]):
         cls,
         value: int,
         range: Range | int,
+        *,
+        on_overflow: Literal["error", "wrap"] = "error",
     ) -> LogicArray:
         """Construct a :class:`!LogicArray` from an :class:`int` with unsigned representation.
 
         The :class:`int` is treated as an arbitrary-length bit vector with unsigned representation where the left-most bit is the most significant bit.
         This bit vector is then constructed into a :class:`!LogicArray`.
 
+        The *on_overflow* parameter can be set to ``"wrap"`` to allow truncating excessive significant bits so the *value* fits into the *range*.
+
+        .. code-block:: python
+
+            >>> LogicArray.from_unsigned(0b10011101, 4, on_overflow="wrap")
+            LogicArray('1101', Range(3, 'downto', 0))
+
         Args:
             value: The integer to convert.
             range: Indexing scheme for the :class:`!LogicArray`.
+            on_overflow: If ``"wrap"``, the *value* will have excessive significant bits truncated so that it fits into the *range*.
+                If ``"error"`` (default), a :class:`ValueError` is raised when the *value* does not fit.
 
         Returns:
             A :class:`!LogicArray` equivalent to the *value*.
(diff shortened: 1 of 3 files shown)


ktbarrett:
Would *truncate* work better? Its basically just taking the bottom N bits. Just like the wrapping that occurs after integer overflow except at arbitrary bit size and not solely 32/64-bits.

imphil:
Better! While thinking about it and the question if some users would want a warning on truncation, I thought an enum argument might help us with forward-compatibility and naming at the same time: how about `overflow_handling=error|truncate` (or `on_overflow` or something like that).

src/cocotb/types/_logic_array.py
```diff
@@ -326,23 +328,68 @@ def from_unsigned(
         Args:
             value: The integer to convert.
             range: Indexing scheme for the :class:`!LogicArray`.
+            on_overflow: If ``"wrap"``, the value will be wrapped to fit in the given range.
+                If ``"error"`` (default), a :class:`ValueError` is raised when the value does not fit.
```
imphil: Thinking about it for a night I'm still not a fan of "wrap" and would prefer "truncate". "Wrap" to me means "wrap around", which is totally not what's happening. 

What do others think @cocotb/maintainers?

ktbarrett: One argument for "wrap" is that VHDL calls this operation "wrap". See "overflow_style" in fixed_pkg

cmarqu:
How about we keep the name `wrap` but mention "truncation" in the docstring, together with an example of seeing it in action?

imphil:
> How about we keep the name wrap but mention "truncation" in the docstring, together with an example of seeing it in action?

Works for me.

(Even though I've never heard of wrap being used in that way and I feel that VHDL is the odd one out here. "truncate" seems unambiguous and as such an easy way out.)

ktbarrett:
In VHDL "truncate" is used for quantization: reminders after division, float->fixed, etc. Its something we may support depending on how closely to VHDL we model the extended set of datatypes.

src/cocotb/types/_logic_array.py
```diff
@@ -317,70 +317,159 @@ def from_unsigned(
...
         range: Range | int,
+        *,
+        on_overflow: Literal["error", "wrap"] = "error",
     ) -> LogicArray:
         """Construct a :class:`!LogicArray` from an :class:`int` with unsigned representation.
 
         The :class:`int` is treated as an arbitrary-length bit vector with unsigned representation where the left-most bit is the most significant bit.
         This bit vector is then constructed into a :class:`!LogicArray`.
 
+        The *on_overflow* parameter can be set to ``"wrap"`` to allow truncating excessive significant bits so the value fits into the *range*.
```
cmarqu:
```suggestion
        The *on_overflow* parameter can be set to ``"wrap"`` to allow truncating excessive significant bits so the *value* fits into the *range*.
```

src/cocotb/types/_logic_array.py
```diff
@@ -317,70 +317,159 @@ def from_unsigned(
...
+
+        .. code-block:: python
+
+            >>> LogicArray.from_unsigned(0b10011101, 4, on_overflow="wrap")
+            LogicArray('1101', Range(3, 'downto', 0))
+
         Args:
             value: The integer to convert.
             range: Indexing scheme for the :class:`!LogicArray`.
+            on_overflow: If ``"wrap"``, the value will have excessive significant bits truncated so the value fits into the *range*.
```
cmarqu:
```suggestion
            on_overflow: If ``"wrap"``, the *value* will have excessive significant bits truncated so that it fits into the *range*.
```

src/cocotb/types/_logic_array.py
```diff
@@ -317,70 +317,159 @@ def from_unsigned(
...
+                If ``"error"`` (default), a :class:`ValueError` is raised when the value does not fit.
```
cmarqu:
```suggestion
                If ``"error"`` (default), a :class:`ValueError` is raised when the *value* does not fit.
```

src/cocotb/types/_logic_array.py
```diff
@@ -317,70 +317,159 @@ def from_unsigned(
...
         range: Range | int,
+        *,
+        on_overflow: Literal["error", "wrap"] = "error",
     ) -> LogicArray:
         """Construct a :class:`!LogicArray` from an :class:`int` with two's complement representation.
 
         The :class:`int` is treated as an arbitrary-length bit vector with two's complement representation where the left-most bit is the most significant bit.
         This bit vector is then constructed into a :class:`!LogicArray`.
 
+        The *on_overflow* parameter can be set to ``"wrap"`` to allow truncating excessive significant bits so the value fits into the *range*.
```
cmarqu:
```suggestion
        The *on_overflow* parameter can be set to ``"wrap"`` to allow truncating excessive significant bits so the *value* fits into the *range*.
```

src/cocotb/types/_logic_array.py
```diff
@@ -317,70 +317,159 @@ def from_unsigned(
...
+            LogicArray('1100111111000111', Range(15, 'downto', 0))
+
+            >>> LogicArray.from_signed(-12345, 4, on_overflow="wrap")
+            LogicArray('0111', Range(3, 'downto', 0))
+
         Args:
             value: The integer to convert.
             range: Indexing scheme for the :class:`!LogicArray`.
+            on_overflow: If ``"wrap"``, the value will have excessive significant bits truncated so the value fits into the *range*.
+                If ``"error"`` (default), a :class:`ValueError` is raised when the value does not fit.
```
cmarqu:
```suggestion
            on_overflow: If ``"wrap"``, the *value* will have excessive significant bits truncated so that it fits into the *range*.
                If ``"error"`` (default), a :class:`ValueError` is raised when the *value* does not fit.
```

cmarqu (approved):

PR #5223, opened 2025-12-28
Refactor Finalization/Initialization (part 1/N)

ktbarrett:
Step 1 of N of refactoring cocotb initialization and finalization. xref #4585, xref #4584. Breaking the dependency between RunningTest and RegressionManager will also enable refactoring RunningTest to use TaskManager.

This gets us to the point where we don't see the following part of the stacktrace when initialization raises.
```
  File "/home/csushma/anaconda3/lib/python3.12/site-packages/cocotb/_init.py", line 117, in _sim_event
    cocotb._regression_manager._fail_simulation(msg)
  File "/home/csushma/anaconda3/lib/python3.12/site-packages/cocotb/regression.py", line 895, in _fail_simulation
    self._running_test.abort(self._sim_failure)
    ^^^^^^^^^^^^^^^^^^
AttributeError: 'RegressionManager' object has no attribute '_running_test'
```

Follow-on commits will:
* Expose a Startup and Shutdown trigger
* Move Pythonland initialization to module initialization time instead of time 0.
* Split handling pre-mature shutdown to the new Shutdown trigger
* Refactor `sim_event` to be finalization which is consistently called

### DIFF
--- a/src/cocotb/_profiling.py
+++ b/src/cocotb/_profiling.py
@@ -10,6 +10,7 @@ import cProfile
 import pstats
 from contextlib import AbstractContextManager, nullcontext
 
+import cocotb._shutdown
 from cocotb_tools import _env
 
 profiling_context: AbstractContextManager[None, None]
@@ -18,16 +19,18 @@ profiling_context: AbstractContextManager[None, None]
 if _env.as_bool("COCOTB_ENABLE_PROFILING"):
     _profile: cProfile.Profile
 
-    def initialize() -> None:
+    def _init() -> None:
         global _profile
         _profile = cProfile.Profile()
 
-    def finalize() -> None:
-        ps = pstats.Stats(_profile).sort_stats("cumulative")
-        ps.dump_stats("cocotb.pstat")
+        def finalize() -> None:
+            ps = pstats.Stats(_profile).sort_stats("cumulative")
+            ps.dump_stats("cocotb.pstat")
+
+        cocotb._shutdown.register(finalize)
 
     class _profiling_context(AbstractContextManager[None, None]):
-        """Context manager that profiles its contents"""
+        """Context manager that profiles its contents."""
 
         def __enter__(self) -> None:
             _profile.enable()
@@ -39,10 +42,7 @@ if _env.as_bool("COCOTB_ENABLE_PROFILING"):
 
 else:
 
-    def initialize() -> None:
-        pass
-
-    def finalize() -> None:
+    def _init() -> None:
         pass
 
     profiling_context = nullcontext()
--- a/src/cocotb/regression.py
+++ b/src/cocotb/regression.py
@@ -15,6 +15,7 @@ import logging
 import os
 import random
 import re
+import sys
 import time
 import warnings
 from collections.abc import Coroutine
@@ -24,6 +25,8 @@ from typing import Any, Callable
 
 import cocotb
 import cocotb._event_loop
+import cocotb._shutdown as shutdown
+import cocotb._test
 from cocotb import logging as cocotb_logging
 from cocotb import simulator
 from cocotb._base_triggers import Trigger
@@ -43,6 +46,7 @@ from cocotb._xunit_reporter import XUnitReporter, bin_xml_escape
 from cocotb.logging import ANSI
 from cocotb.simtime import get_sim_time
 from cocotb.task import Task
+from cocotb_tools import _env
 
 __all__ = (
     "RegressionManager",
@@ -289,6 +293,8 @@ class RegressionManager:
     def start_regression(self) -> None:
         """Start the regression."""
 
+        self.log.info("Running tests")
+
         # sort tests into stages
         self._test_queue.sort(key=lambda test: test.stage)
 
@@ -352,6 +358,7 @@ class RegressionManager:
             except Exception:
                 self._record_test_init_failed()
                 continue
+            cocotb._test.set_current_test(self._running_test)
 
             self._log_test_start()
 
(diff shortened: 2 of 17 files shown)


src/cocotb/_profiling.py
```diff
@@ -11,19 +11,23 @@
...
-        ps = pstats.Stats(_profile).sort_stats("cumulative")
-        ps.dump_stats("cocotb.pstat")
+        def finalize() -> None:
+            ps = pstats.Stats(_profile).sort_stats("cumulative")
+            ps.dump_stats("cocotb.pstat")
+
+        cocotb._shutdown.register(finalize)
 
     class _profiling_context(AbstractContextManager[None, None]):
         """Context manager that profiles its contents"""
```
cmarqu:
```suggestion
        """Context manager that profiles its contents."""
```
While here.

src/cocotb/regression.py
```diff
@@ -897,3 +904,56 @@ def _on_sim_end(self) -> None:
...
+"""The global regression manager instance."""
+
+
+def _run_regression(_: object) -> None:
+    """Setup and run a regression."""
+
+    global _manager_inst
+    _manager_inst = RegressionManager()
+
+    # sys.path normally includes "" (the current directory), but does not appear to when python is embedded.
```
```suggestion
    # sys.path normally includes "" (the current directory), but does not appear to when Python is embedded.
```

cmarqu (approved):
Two tiny drive-by fixes, feel free to ignore.

ktbarrett:
pytest plugin needs to be updated to use the new `cocotb._test._current_test`.

ktbarrett:
@tymonx This refactor is blocked by the fact that the pytest plugin does not use `RunningTest` in `cocotb._test`. The way that the pytest runner tries to mimic the RegressionManager rather than refactoring to prevent massive coupling is getting in the way. Can you un-spaghetti the pytest RegressionManager?

ktbarrett:
I worked around the issue with the pytest RegressionManager by leaving the dependency cycle in the definition of `start_soon`. This is super ugly and will need to get fixed.

tymonx:
@ktbarrett No problem, I will do that.

ktbarrett:
@tymonx there was some refactoring I wanted to do to the RunningTest to make it a little less of a mess, so if it ends up being a chore let me know and we can try to get that in soon so you could reuse it.

tymonx:
@ktbarrett Done https://github.com/cocotb/cocotb/pull/5229 but it is failing now on some VPI/VHPI function after I merged the latest changes :melting_face: 

```plaintext
vpi_iterate returned NULL for type vpiInstance for object NULL
vpi_get: unknown property
Unable to map vpiConst type 0 onto GPI type, guessing this is a logic vector
```

tymonx:
False alarm https://github.com/cocotb/cocotb/pull/5229#issuecomment-3702381390 I need to refactor that properly.

tymonx:
@ktbarrett Done. This should help -> https://github.com/cocotb/cocotb/pull/5229

tymonx:
After merging https://github.com/cocotb/cocotb/pull/5229 you can add `cocotb._test._current_test = running_test` inside the `def _start()` method and remove assigning instance of pytest regression manager to `cocotb._regression_manager` handler :thinking: 

```patch
diff --git a/src/cocotb_tools/pytest/_init.py b/src/cocotb_tools/pytest/_init.py
index 06e97f5f..ce8d4818 100644
--- a/src/cocotb_tools/pytest/_init.py
+++ b/src/cocotb_tools/pytest/_init.py
@@ -10,7 +10,6 @@ import sys
 from logging import getLogger
 from random import seed
 from time import time
-from typing import cast
 
 import cocotb
 from cocotb import simtime, simulator
@@ -50,8 +49,7 @@ def run_regression(argv: list[str]) -> None:
         toplevel=env.as_str("COCOTB_TOPLEVEL"),
     )
 
-    cocotb._regression_manager = cast("cocotb.regression.RegressionManager", manager)
-    cocotb._regression_manager.start_regression()
+    manager.start_regression()
 
 
 def _setup_simulation_environment(argv: list[str] | None = None) -> None:
diff --git a/src/cocotb_tools/pytest/regression.py b/src/cocotb_tools/pytest/regression.py
index 785f2845..cc908ecf 100644
--- a/src/cocotb_tools/pytest/regression.py
+++ b/src/cocotb_tools/pytest/regression.py
@@ -809,6 +809,7 @@ class RegressionManager:
             running_test: Test to run.
         """
         self._running_test = running_test
+        cocotb._test._current_test = running_test
 
         if self._scheduled:
             self._timer1._register(self._running_test.start)
```

src/cocotb/regression.py
```diff
@@ -897,3 +907,55 @@ def _on_sim_end(self) -> None:
         )
         self._running_test.abort(self._sim_failure)
         cocotb._event_loop._inst.run()
+
+
+_manager_inst: RegressionManager
+"""The global regression manager instance."""
```
ktbarrett, line 913: Technically we don't need this anymore, but there is one test that inspects the current `Test` object, which is RM specific, so we'll leave it accessible somewhere.

ktbarrett:
@tymonx Why does the runner have a logger with the overloaded formatting in it?

tymonx:
> @tymonx Why does the runner have a logger with the overloaded formatting in it?

@ktbarrett Can you share with me which file/module you have in mind :thinking: I don't know if I follow.

ktbarrett:
Well crap, nvm. It was failing on my machine, but it seems to be working in CI now. I'll investigate, but it's probably my fault.

PR #5225, opened 2025-12-30
Update VPI headers

ktbarrett:
Pull the newest version of the headers from Riviera's package. From what I can tell it follows the standard in all but a few cases which were manually accounted for. One Riviera change that I kept however, was the use of stdint.h over the manual fussing about with the fixed integer size definition. I have no clue why they decided they needed to reinvent the wheel.

### DIFF
--- a/src/cocotb/_vendor/vpi/vpi_user.h
+++ b/src/cocotb/_vendor/vpi/vpi_user.h
@@ -18,6 +18,7 @@
 #define VPI_USER_H
 
 #include <stdarg.h>
+#include <stdint.h>
 
 #ifdef  __cplusplus
 extern "C" {
@@ -27,23 +28,12 @@ extern "C" {
 /*----------------------------- Portability Help -----------------------------*/
 /*----------------------------------------------------------------------------*/
 
-#if defined (_MSC_VER)
-typedef unsigned __int64 uint64_t;
-typedef unsigned __int32 uint32_t;
-typedef unsigned __int8 uint8_t;
-typedef signed __int64 int64_t;
-typedef signed __int32 int32_t;
-typedef signed __int8 int8_t;
-#elif defined(__MINGW32__) || (defined(__APPLE__) && defined(__MACH__))
-#include <stdint.h>
-#elif defined(__linux)
-#include <inttypes.h>
-#else
-#include <sys/types.h>
-#endif
+/* Define size-critical types on all OS platforms. */
+
+/* Replaced the standard's implementation of fixed-width type definitions with */
+/* the definitions from stdint.h. */
+
 /* Sized variables */
-
-
 #ifndef SVPI_TYPES
 #define SVPI_TYPES
 typedef int64_t PLI_INT64;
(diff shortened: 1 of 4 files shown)


src/cocotb/_vendor/vpi/vpi_user.h
```diff
@@ -486,7 +472,7 @@ typedef PLI_UINT32 *vpiHandle;
 #define vpiHexConst                5   /* hexadecimal integer */
 #define vpiStringConst             6   /* string literal */
 #define vpiIntConst                7   /* HDL integer constant (1364-2001) */
-#define vpiTimeConst			   8   /* time constant */
+#define vpiTimeConst		   8   /* time constant */
```
marlonjames: nit: alignment

src/cocotb/_vendor/vpi/vpi_user.h
```diff
@@ -27,23 +28,12 @@ extern "C" {
...
-#include <stdint.h>
-#elif defined(__linux)
-#include <inttypes.h>
-#else
-#include <sys/types.h>
-#endif
-/* Sized variables */
+/* Define size-critical types on all OS platforms. */
 
+/* Replaced the standards implementation of fixed-width type definitions with */
```
cmarqu:
```suggestion
/* Replaced the standard's implementation of fixed-width type definitions with */
```

src/cocotb/share/include/vpi_user_ext.h
```diff
@@ -12,11 +12,7 @@
 extern "C" {
 #endif
 
-/* used by Cadence Xcelium for packed unions */
-#define vpiUnionNet          525
-
 /* used by Cadence Xcelium for Verilog-AMS */
-#define vpiRealNet           526
```
cmarqu, line 19: This is now becoming a `vpiShortRealNet`, but the AMS example still seems to run fine in CI.

imphil:
I assume there's still no version of the VPI header in a form other than the PDF of the standard?

ktbarrett:
@imphil There was nothing in the errata section on the IEEE site, which is where it would be if anywhere. Also the PDF overlay text in the 2023 version of the standard isn't properly formatted, so getting it from Riviera was the best way.

There were also no other instances of the 2023 version of the header anywhere else I could find on Google, so we might be the first with the most complete set of 2023 version headers.

src/cocotb/share/include/vpi_user_ext.h
ktbarrett: Yeah I noticed that too. They use slightly different numbers. `vpiShortRealNet` is 530 in their version of the header (they call it `vpi_user_cds.h`). I'm not sure all the discrepencies, but considering there are some and it uses a different header and values I'm guessing there's a VPI compatibility flag we should be using to ensure it's using the 2023 standard version and not their own.

src/cocotb/_vendor/vpi/vpi_user.h
ktbarrett:
```suggestion
#define vpiTimeConst		       8   /* time constant */
```

src/cocotb/share/include/vpi_user_ext.h
ktbarrett: Opened https://github.com/cocotb/cocotb/issues/5227

src/cocotb/_vendor/vpi/vpi_user.h
imphil: That's not only alignment, that's also tabs being used.

ktbarrett:
The standard's reimplementation of fixed sized integers is more likely to be incorrect than correct, especially as time progresses, so I'll stick to using the C standard headers.

marlonjames (approved):

PR #5229, opened 2025-12-31
chore(pytest-plugin): use RunningTest class

tymonx:
- Using the `RunningTest` class
- Added more docstrings and comments
- Removed dependency to `cocotb.regression`
- Added own `SimFailure` exception instead relying on `cocotb.regression.SimFailure`
- Added `RunningTestSetup` class for pytest `setup` - `teardown`. Added sub-tasks in test setup like clock generators cannot be cancelled, they must survive test `setup` abort, test `call` abort but be aborted in test `teardown`
- Code refactoring that should simplify and help understanding pytest flow `setup` -> `call` -> `teardown`
- ~Align handling random seed in the same way like in the built-in regression manager~
  - See https://github.com/cocotb/cocotb/pull/5230)
- ~Added `error` in `PYTHONWARNINGS` environment variable was... hanging pytest after refactoring :open_mouth: Related with the latest pytest `9.0.2` :thinking:~ 

~@ktbarrett there is some regression in the latest code changes in VPI/VHPI functions.~

### DIFF
--- a/src/cocotb_tools/pytest/regression.py
+++ b/src/cocotb_tools/pytest/regression.py
@@ -17,7 +17,7 @@ from importlib import import_module
 from logging import Logger, getLogger
 from multiprocessing.connection import Client
 from pathlib import Path
-from time import sleep
+from time import sleep, time
 from typing import Any, Callable, Literal, cast
 
 from _pytest.config import default_plugins
@@ -45,8 +45,7 @@ import cocotb
 from cocotb import simulator
 from cocotb._extended_awaitables import with_timeout
 from cocotb._gpi_triggers import Timer
-from cocotb._test_functions import TestSuccess
-from cocotb.regression import SimFailure
+from cocotb._test import RunningTest
 from cocotb.simtime import TimeUnit, get_sim_time
 from cocotb.task import Task
 from cocotb_tools.pytest.fixture import (
@@ -54,11 +53,18 @@ from cocotb_tools.pytest.fixture import (
     AsyncFixtureCachedResult,
     resolve_fixture_arg,
 )
+from cocotb_tools.pytest.test import RunningTestSetup
 
 RETRIES: int = 10
 INTERVAL: float = 0.1  # seconds
 
 AsyncFunction = Callable[..., Awaitable]
+When = Literal["setup", "call", "teardown"]
+"""Test phase."""
+
+
+class SimFailure(BaseException):
+    """A Test failure due to simulator failure. Used internally."""
 
 
 def finish_on_exception(method: Callable[..., Any]) -> Callable[..., Any]:
@@ -91,6 +97,7 @@ class RegressionManager:
         keywords: Iterable[str] | None = None,
         test_modules: Iterable[str] | None = None,
         invocation_dir: Path | str | None = None,
+        seed: int | None = None,
     ) -> None:
         """Create new instance of regression manager for cocotb tests.
 
@@ -103,11 +110,25 @@ class RegressionManager:
             test_modules: List of test modules (Python modules with cocotb tests) to be loaded.
             invocation_dir: Path to directory location from where pytest was invoked.
             reporter_address: IPC address (Unix socket, Windows pipe, TCP, ...) to tests reporter.
+            seed: Initialization value for the random number generator. If not provided, use current timestamp.
         """
         self._toplevel: str = toplevel
-        self._task: Task
-        self._tasks: deque[Task] = deque[Task]()
-        self._subtasks: list[Task] = []
+        """Name of top level."""
+
+        self._running_test: RunningTest
+        """Current running test: "setup", "call" or "teardown"."""
+
+        self._setups: deque[RunningTest] = deque[RunningTest]()
+        """List of test setups that were populated from the :py:func:`pytest.hookspec.pytest_fixture_setup` hook."""
+
+        self._call: RunningTest | None = None
+        """Test call that was populated from the :py:func:`pytest.hookspec.pytest_runtest_call` hook."""
+
+        self._teardowns: deque[RunningTest] = deque[RunningTest]()
+        """List of test teardowns that were populated from registered setup finalizers via
+        :py:meth:`pytest.FixtureRequest.addfinalizer`` method.
+        """
+
         self._scheduled: bool = False
         self._index: int = 0
         self._finished: bool = False
@@ -121,6 +142,8 @@ class RegressionManager:
         self._logging_root_level: int = getLogger().level
         self._logging_level: int = 0
         self._logging_restored: bool = False
+        self._seed: int = int(time()) if seed is None else seed
+        self._random_state: Any = random.getstate()
 
         pluginmanager = PytestPluginManager()
 
@@ -205,7 +228,7 @@ class RegressionManager:
         return [
             f"Running on {cocotb.SIM_NAME} version {cocotb.SIM_VERSION}",
             f"Initialized cocotb v{cocotb.__version__} from {Path(__file__).parent.resolve()}",
-            f"Seeding Python random module with {cocotb.RANDOM_SEED}",
+            f"Seeding Python random module with {self._seed}",
             f"Top level set to {self._toplevel!r}",
         ]
 
@@ -248,8 +271,14 @@ class RegressionManager:
 
     @hookimpl(tryfirst=True)
     def pytest_runtest_protocol(self, item: Item, nextitem: Item | None) -> bool:
+        """Perform the runtest protocol for a single test item.
+
+        Args:
+            item: Test item for which the runtest protocol is performed.
+            nextitem: The scheduled-to-be-next test item (or ``None`` if this is the end my friend).
+        """
         item.ihook.pytest_runtest_logstart(nodeid=item.nodeid, location=item.location)
-        self._setup()
+        self._setup(item=item, nextitem=nextitem)
 
         return True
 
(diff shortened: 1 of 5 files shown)


tymonx:
That message was from the beginning but now with `RunningTest` is throwing the `SimFailure` exception. I will try to look into it.

tymonx:
@ktbarrett Not related. It is how pytest working. When fixture is created, like clock generator, task must be keep alive until finalizer is called (teardown). Fixture are per session, package, module, class and function. If we want to use `RunningTest` for test setup, then they should be kept alive and aborted during calling finalizer by pytest.

In other words, I need little more time to refactor that properly :slightly_smiling_face: I did that too fast.

tymonx:
@ktbarrett Major refactoring done. Tested locally with GHDL. There are some issues with other simulators. I suspect that tasks created in pytest `setup` stage don't survive to pytest `teardown` stage causing `SimFailure` (missing/cancelled clock generation). I will try to debug that with Icarus simulator. It is failing on CI.

Original implementation was keeping sub-tasks and they were cancelled by pytest finalizer functions. I will need to implement the same functionality when using `cocotb._test.RunningTest`.

tymonx:
@ktbarrett Done. Ready for review, approve and merge.

Refactoring to use the `cocotb._test.RunningTest` class exposed one flaw with original implementation. Tasks added during setup defined as normal asynchronous function (setup-only without teardown) were still running. This was a logical bug. These tasks should be cancelled. Setup defined as asynchronous generator (setup with teardown) should cancel added tasks in teardown.

```python
from collections.abc import AsyncGenerator

import pytest

@pytest.fixture(name="setup_only_without_teardown")
async def setup_only_without_teardown_fixture(dut: Any) -> None:
    Clock(dut.clk, 10, unit="ns").start(start_high=False)

    # Clock generation will end here before test call


@pytest.fixture(name="setup_with_teardown")
async def setup_with_teardown_fixture(dut: Any) -> AsyncGenerator[None, None]:
    Clock(dut.clk, 10, unit="ns").start(start_high=False)

    yield  # Calling test, clock is alive during test call

    # Clock generation will end here
```

tymonx:
CI jobs with Questa and Xcelium are still queued/pending (almost ~10 hours) :melting_face:

src/cocotb_tools/pytest/test.py
```diff
@@ -0,0 +1,29 @@
...
+    """Running test setup without cancelling added sub-tasks."""
+
+    def __init__(
+        self, test_complete_cb: Callable[[], None], main_task: Task[None]
+    ) -> None:
+        """Create new instance of running test setup."""
+        super().__init__(test_complete_cb=test_complete_cb, main_task=main_task)
+
+        self.subtasks: list[Task[Any]] = []
+        """Sub-tasks that must be keep alive during test setup, call and teardown."""
```
ktbarrett, line 25: This is a current annoyance to me as well and will hopefully be fixed by my refactor. Tasks made by TaskManager need to be tracked for cancellation at the end of the test, but they need to not end the test if they fail, since the TaskManager will handle that. These things are currently not separated, but they will be.

ktbarrett (approved):

src/cocotb_tools/pytest/regression.py
```diff
@@ -105,9 +111,20 @@ def __init__(
             reporter_address: IPC address (Unix socket, Windows pipe, TCP, ...) to tests reporter.
         """
         self._toplevel: str = toplevel
-        self._task: Task
-        self._tasks: deque[Task] = deque[Task]()
-        self._subtasks: list[Task] = []
+        """Name of top level."""
+
+        self._running_test: RunningTest
+        """Current running test: setup, call or teardown."""
```
cmarqu:
```suggestion
        """Current running test: "setup", "call" or "teardown"."""
```

src/cocotb_tools/pytest/regression.py
```diff
@@ -105,9 +111,20 @@ def __init__(
...
+
+        self._setups: deque[RunningTest] = deque[RunningTest]()
+        """List of test setups populate from the pytest_fixture_setup hook."""
```
cmarqu: Is there a word missing before "populate"?

src/cocotb_tools/pytest/regression.py
```diff
@@ -105,9 +111,20 @@ def __init__(
...
+
+        self._call: RunningTest | None = None
+        """Test call populate from the pytest_runtest_call hook."""
+
+        self._teardowns: deque[RunningTest] = deque[RunningTest]()
+        """List of test teardowns populate from setup finalizers."""
```
cmarqu: Is there a word missing before "populate"?

src/cocotb_tools/pytest/regression.py
```diff
@@ -105,9 +111,20 @@ def __init__(
...
+        """Name of top level."""
+
+        self._running_test: RunningTest
+        """Current running test: setup, call or teardown."""
+
+        self._setups: deque[RunningTest] = deque[RunningTest]()
+        """List of test setups populate from the pytest_fixture_setup hook."""
+
+        self._call: RunningTest | None = None
+        """Test call populate from the pytest_runtest_call hook."""
```
cmarqu: Is there a word missing before "populate"?

src/cocotb_tools/pytest/regression.py
```diff
@@ -248,8 +265,14 @@ def pytest_runtestloop(self, session: Session) -> bool:
 
     @hookimpl(tryfirst=True)
     def pytest_runtest_protocol(self, item: Item, nextitem: Item | None) -> bool:
+        """Perform the runtest protocol for a single test item.
+
+        Args:
+            item: Test item for which the runtest protocol is performed.
+            nextitem: The scheduled-to-be-next test item (or None if this is the end my friend).
```
cmarqu:
```suggestion
            nextitem: The scheduled-to-be-next test item (or ``None`` if this is the end my friend).
```

src/cocotb_tools/pytest/regression.py
```diff
@@ -282,10 +305,21 @@ def _collect(
...
         **kwargs: object,
-    ) -> bool:
+    ) -> TestReport:
+        """Invoke test setup, call, teardown and generate test report from it.
+
+        Args:
+            item: The pytest item.
+            when: Test setup, call or teardown.
+            func: Test function that will be called.
+            kwargs: Additional named arguments that will be passed to test function.
```
cmarqu:
```suggestion
            kwargs: Additional named arguments that will be passed to the test function.
```

src/cocotb_tools/pytest/regression.py
```diff
@@ -311,65 +345,112 @@ def _call_and_report(
...
-            # Cancel remaining tasks when exception was raised (excinfo)
-            for task in self._tasks:
-                task.cancel()
+    def _completed(self, item: Item, when: When) -> TestReport:
+        """Part of test setup, call or teardown completed callback.
 
-            self._tasks.clear()
+        Args:
+            item: The pytest item.
+            when: Test setup, call or teardown.
```
cmarqu:
```suggestion
            when: Test "setup", "call" or "teardown".
```

src/cocotb_tools/pytest/regression.py
```diff
@@ -311,65 +345,112 @@ def _call_and_report(
...
-        elif passed:
-            self._call()
+    def _setup(
+        self, item: Item, nextitem: Item | None = None, report: TestReport | None = None
+    ) -> None:
+        """Test setup.
+
+        Args:
+            item: The pytest item.
+            nextitem: The next pytest item. None if there are no more pytest items.
```
cmarqu:
```suggestion
            nextitem: The next pytest item. ``None`` if there are no more pytest items.
```

src/cocotb_tools/pytest/regression.py
```diff
@@ -311,65 +345,112 @@ def _call_and_report(
...
+    def _teardown(
+        self, item: Item, nextitem: Item | None = None, report: TestReport | None = None
+    ) -> None:
+        """Test teardown.
 
-        if self._tasks:
-            return self._execute(self._teardown, self._tasks.popleft())
+        Args:
+            item: The pytest item.
+            nextitem: The next pytest item. None if there are no more pytest items.
```
cmarqu:
```suggestion
            nextitem: The next pytest item. ``None`` if there are no more pytest items.
```

src/cocotb_tools/pytest/regression.py
tymonx: @cmarqu Fixed in this commit 2817cc159e3e12c0cdccf7cf2a550befde7e826d Now I think it should be better :thinking:

src/cocotb_tools/pytest/regression.py
tymonx: @cmarqu Fixed in this commit 2817cc159e3e12c0cdccf7cf2a550befde7e826d Now I think it should be better :thinking:

src/cocotb_tools/pytest/regression.py
tymonx: @cmarqu Fixed in this commit 2817cc159e3e12c0cdccf7cf2a550befde7e826d Now I think it should be better :thinking:

imphil:
@tymonx Just a quick note here: Please try to bundle your updates to address all review commits at the same time and avoid pushing very frequently to the same PR. Each update triggers a full CI run, some with limited licenses available.

There's also a weak preference for rebasing your changes into a series of self-contained (atomic) commits and force-pushing. Otherwise we have to remember to squash your changes before merging.

tymonx:
@imphil I will do that next time :saluting_face: Sorry for that :disappointed: I have squashed all commits to single commit for this Pull Request.

PR #5235, opened 2026-01-05
Parametrized test_cocotb to improve test coverage(Fixes #5085)

### LINKED ISSUE #5085: Test `COCOTB_REDUCED_LOG_FMT=0` in regression
We are missing some coverage (and potentially masking errors) by not testing this.

ZiaCheemaGit:
Fixes #5085
Changed file
repo/tests/pytest/test_cocotb.py
to parametrize test with REDUCED_LOG_FMT = [0, 1]
repo/src/cocotb/_init.py
repo/src/cocotb/_coverage.py
to accomodate the speculative combination.Created separate coverage tmps for each run and then combined

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -18,7 +18,7 @@ nox.options.sessions = ["dev_test"]
 
 test_deps = ["pytest>=6"]
 coverage_deps = ["coverage[toml]>=7.2", "pytest-cov"]
-coverage_report_deps = ["coverage[toml]>=7.2", "gcovr"]
+coverage_report_deps = ["coverage[toml]>=7.2", "gcovr==8.4"]
 
 dev_deps = [
     "mypy",


tymonx:
In my opinion, better would be to create standalone and isolated unit tests that will test setting `COCOTB_REDUCED_LOG_FMT` environment variable to cover https://github.com/cocotb/cocotb/issues/5085.

- It will allow to run specific test like `COCOTB_REDUCED_LOG_FMT=0` vs `COCOTB_REDUCED_LOG_FMT=1`. It is much easier to debug any code changes/improvements and test all corner cases and code paths in cocotb logging
- Parametrizing `test_cocotb` will run ALL cocotb tests without any feedback how  `COCOTB_REDUCED_LOG_FMT` affected log handlers. This will increase overall time duration of executing all tests without any gain
- You could use pytest built-in [monkeypatch](https://docs.pytest.org/en/stable/how-to/monkeypatch.html) and [caplog](https://docs.pytest.org/en/stable/how-to/logging.html) fixtures to cover all scenarios in cocotb logging in isolated unit tests
- No need to change anything how coverage is collected because coverage in this case will be recorded to the default `.coverage` database

Example of `tests/pytest/test_logging.py` file:

```python
import re
from logging import getLogger, Logger, INFO

from pytest import MonkeyPatch
from pytest import LogCaptureFixture

import cocotb.logging
import cocotb.simulator

# https://docs.pytest.org/en/stable/how-to/logging.html#caplog-fixture


def mock_get_sim_time() -> tuple[int, int]:
    return 0, int(100e6)


def test_env_cocotb_reduced_log_fmt_on(monkeypatch: MonkeyPatch, caplog: LogCaptureFixture) -> None:
    """Test :envvar:`COCOTB_REDUCED_LOG_FMT` environment variable set to ``1``."""
    monkeypatch.setenv("COCOTB_REDUCED_LOG_FMT", "1")
    monkeypatch.setattr(cocotb.simulator, "get_sim_time", mock_get_sim_time, raising=False)

    cocotb.logging._configure(())

    with caplog.at_level(INFO):
        getLogger("cocotb").info("test message")

        # 100.00ns INFO cocotb test message
        assert re.match(r'^\s*100\.00ns\s+INFO\s+cocotb\s+test message$', caplog.text)


def test_env_cocotb_reduced_log_fmt_off(monkeypatch: MonkeyPatch, caplog: LogCaptureFixture) -> None:
    """Test :envvar:`COCOTB_REDUCED_LOG_FMT` environment variable set to ``0``."""
    monkeypatch.setenv("COCOTB_REDUCED_LOG_FMT", "0")
    monkeypatch.setattr(cocotb.simulator, "get_sim_time", mock_get_sim_time, raising=False)

    cocotb.logging._configure(())

    with caplog.at_level(INFO):
        getLogger("cocotb").info("test message")

        # 100.00ns INFO cocotb test_env.py:38 in ..nv_cocotb_reduced_log_fmt_off test message
        assert re.match(r'^\s*100\.00ns\s+INFO\s+cocotb\s+test_env.py:[0-9]+\s+in\s+\.\.\w+\stest message$', caplog.text)
```

It will require to patch [src/cocotb/logging.py::default_config()](https://github.com/cocotb/cocotb/blob/v2.0.1/src/cocotb/logging.py#L51) to something like this:

```python
def default_config(...) -> None:
    logging.basicConfig()

    root: logging.Logger = logging.getLogger()

    if not root.handlers:
        root.handlers = [logging.StreamHandler(sys.stdout)]

    for handler in root.handlers:
        handler.addFilter(SimTimeContextFilter())
        handler.setFormatter(
            SimLogFormatter(
                reduced_log_fmt=reduced_log_fmt,
                strip_ansi=strip_ansi,
                prefix_format=prefix_format,
                multiline_indent=multiline_indent,
            )
        )

    logging.getLogger("cocotb").setLevel(logging.INFO)
    logging.getLogger("gpi").setLevel(logging.INFO)

    logging.captureWarnings(True)
```

Because pytest built-in [caplog](https://docs.pytest.org/en/stable/how-to/logging.html) fixture is adding own log handlers in the root logger to capture all log records from all loggers. Cocotb logging is always overriding list of handlers, discarding what pytest added. This must be patched/fixed.

Then to test specific scenario/failed regression with `COCOTB_REDUCED_LOG_FMT` I can simply do:

```shell
pytest -s -v -k test_env_cocotb_reduced_log_fmt_off tests/pytest/test_logging.py
```

src/cocotb/_init.py
```diff
@@ -174,28 +174,59 @@ def _start_user_coverage() -> None:
             ) from None
         else:
             config_filepath = os.getenv("COVERAGE_RCFILE")
+
+            tmp_data_file = f".coverage{os.getenv('COCOTB_REDUCED_LOG_FMT')}.tmp"
```
ktbarrett: Use [`NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) instead. We don't need to embed `COCOTB_REDUCED_LOG_FMT` into the name. Same in `src/cocotb_tools/_coverage.py`

tymonx:
Still parametrizing `test_cocotb` is handy because it allows to achieve [mutation testing](https://en.wikipedia.org/wiki/Mutation_testing) for HDL and cocotb. This could be useful for:

- Environment variables affecting simulators or HDL code (Verilog/SystemVerilog)
- VHDL generics and Verilog/SystemVerilog parameters
- plusargs
- Different simulator compilation/elaboration/runtime options

But this can be easily achieved by using pytest `request.node.name` as part of test directory without any changes in cocotb:

```python
import re
from pathlib import Path

import pytest
from pytest import FixtureRequest

# Used to remove invalid characters that are not POSIX compatible in paths
NORMALIZE_NAME: re.Pattern = re.compile(r'\W')


@pytest.mark.parametrize("value1", (1, 2))
def test_cocotb(request: FixtureRequest, value1: int):
    test_name: str = NORMALIZE_NAME.sub('_', request.node.name)
    test_dir: Path = Path.cwd() / "sim_build" / test_name    

    # ...

    runner.build(
        # ...
        build_dir=test_dir,
    )
    
    runner.test(
        # ...
        build_dir=test_dir,
        test_dir=test_dir,
    )
```

This will create a separate test directory per each parametrized test with own `.coverage.cocotb` database:

* `sim_build/test_cocotb_1_/.coverage.cocotb`
* `sim_build/test_cocotb_2_/.coverage.cocotb`

Then `nox` by calling [src/cocotb_tools/combine_results.py](https://github.com/cocotb/cocotb/blob/v2.0.1/src/cocotb_tools/combine_results.py) will collect and combine all generated `.coverage.cocotb` databases.

src/cocotb/_init.py
tymonx: One note that if generated coverage files will be not removed, it can lead to side effects. Consider:

- 1st pytest run will generate:
  - `sim_build/.coverage.cocotb.<random-name-1>`
- Some code changes in cocotb
- 2nd pytest run will generate:
  - `sim_build/.coverage.cocotb.<random-name-2>`

If 1st run of pytest covered something but after code changes 2nd run not, then combining coverage results in that way can lead to side effects and false positive coverage. This will require to remove previously generated coverage databases.

ktbarrett: That's fine. Old artifacts are not left around when running in CI. Maybe the noxfile should clean up those files when running tests for local changes.

Maybe we should be running `make clean` before test runs unconditionally.

src/cocotb/_init.py
```diff
@@ -174,28 +175,63 @@ def _start_user_coverage() -> None:
             ) from None
         else:
             config_filepath = os.getenv("COVERAGE_RCFILE")
+
+            tmp_data_file = tempfile.NamedTemporaryFile(
```
ktbarrett: You need to close this file in `stop_user_coverage` to get it to clean up.

src/cocotb_tools/_coverage.py
```diff
@@ -28,6 +35,26 @@ def stop_library_coverage() -> None:
...
+            if files:
+                final_data_file = ".coverage.cocotb"
+                combiner = coverage.coverage(
+                    data_file=final_data_file,
+                    config_file=False,
+                    branch=True,
+                    source=["cocotb"],
+                )
+                combiner.combine(data_paths=files, strict=True, keep=True)
+
```
ktbarrett, line 61: Same here. The tempfile needs to be closed.

src/cocotb/_init.py
```diff
@@ -231,6 +232,8 @@ def stop_user_coverage() -> None:
                             data_file=final_data_file, config_file=config_filepath
                         )
                     combiner.combine(data_paths=files, strict=True, keep=True)
+                tmp_data_file_controller.close()
+                Path(tmp_data_file).unlink()
```
ktbarrett:
```suggestion
```

src/cocotb/_init.py
```diff
@@ -176,11 +176,12 @@ def _start_user_coverage() -> None:
         else:
             config_filepath = os.getenv("COVERAGE_RCFILE")
 
-            tmp_data_file = tempfile.NamedTemporaryFile(
+            tmp_data_file_controller = tempfile.NamedTemporaryFile(
                 prefix=".coverage.cocotb.",
                 suffix=".tmp",
-                delete=True,
-            ).name
+                delete=False,
```
ktbarrett:
```suggestion
```

src/cocotb_tools/_coverage.py
```diff
@@ -18,11 +18,12 @@ def start_cocotb_library_coverage(_: object) -> None:  # pragma: no cover
             "cocotb library coverage collection requested but coverage package not available. Install it using `pip install coverage`."
         ) from None
     else:
-        tmp_data_file = tempfile.NamedTemporaryFile(
+        tmp_data_file_controller = tempfile.NamedTemporaryFile(
             prefix=".coverage.cocotb.",
             suffix=".tmp",
-            delete=True,
-        ).name
+            delete=False,
```
ktbarrett:
```suggestion
```

src/cocotb_tools/_coverage.py
```diff
@@ -54,6 +55,8 @@ def stop_library_coverage() -> None:
                     source=["cocotb"],
                 )
                 combiner.combine(data_paths=files, strict=True, keep=True)
+            tmp_data_file_controller.close()
+            Path(tmp_data_file).unlink()
```
ktbarrett:
```suggestion
```

ktbarrett (commented):
Was there a reason to set `delete=False`?

ZiaCheemaGit:
yes because by default delete=True and coverage.combine() needs this file. 
temp coverage file is closed right at top and coverage.combine() is called called at the bottom also that is why unlink is used after combine

ktbarrett:
> temp coverage file is closed right at top

This is no longer true now that you are calling `tmp_data_file_controller.close()` after the combine step.

ZiaCheemaGit:
yes you are right my intuition was that coverage.save() might not affect the tempfile behaviour I tested now by removing delete=False by this approach(https://github.com/cocotb/cocotb/issues/5085#issuecomment-3621823630) it behaves same so we can remove delete=False.

ktbarrett:
Looks like a rebase is necessary after merging some other PRs.

src/cocotb/_init.py
```diff
@@ -121,27 +123,65 @@ def _start_user_coverage() -> None:
...
+                                omit=[f"{cocotb_package_dir}/*"],
+                            )
+                        else:
+                            combiner = coverage.coverage(
+                                data_file=final_data_file, config_file=config_filepath
+                            )
+                        combiner.combine(data_paths=files, strict=True, keep=True)
+                finally:
+                    tmp_data_file_controller.close()
+                    Path(tmp_data_file).unlink()
```
ktbarrett:
```suggestion
```

src/cocotb_tools/_coverage.py
```diff
@@ -16,17 +19,46 @@ def start_cocotb_library_coverage(_: object) -> None:  # pragma: no cover
...
+                    combiner = coverage.coverage(
+                        data_file=final_data_file,
+                        config_file=False,
+                        branch=True,
+                        source=["cocotb"],
+                    )
+                    combiner.combine(data_paths=files, strict=True, keep=True)
+            finally:
+                tmp_data_file_controller.close()
+                Path(tmp_data_file).unlink()
```
ktbarrett:
```suggestion
```

src/cocotb/_init.py
```diff
@@ -7,6 +7,8 @@
 
 import logging
 import random
+import sys
```
ktbarrett:
```suggestion
```

ktbarrett:
Fixed the last pre-commit issue. After CI completes we can merge.

ZiaCheemaGit:
26 checked passed about 16 hours ago and 3 now failing checks were idle for 16 hours.
coverage report is failing for all 3. Is it related to codecov bot?
I also ran same merge in my fork(https://github.com/ZiaCheemaGit/cocotb/pull/7) and now even tests which were passing here are failing there is there any documentation for github workflows or any help material.

ktbarrett:
I think the failures are due to gcovr upgrade. Can you change the line in the noxfile.py to pin gcovr to 8.4 for now?

```diff
diff --git a/noxfile.py b/noxfile.py
index fba08edc..97247367 100644
--- a/noxfile.py
+++ b/noxfile.py
@@ -18,7 +18,7 @@ nox.options.sessions = ["dev_test"]

 test_deps = ["pytest>=6"]
 coverage_deps = ["coverage[toml]>=7.2", "pytest-cov"]
-coverage_report_deps = ["coverage[toml]>=7.2", "gcovr"]
+coverage_report_deps = ["coverage[toml]>=7.2", "gcovr==8.4"]

 dev_deps = [
     "mypy",
```

PR #5248, opened 2026-01-07
fix(logging): keep log handlers of the root logger

### LINKED ISSUE #5239: bug(logging): cocotb.logging module is forcefully overriding all added log handlers, cannot use pytest caplog fixture
Context: https://github.com/cocotb/cocotb/pull/5235#issuecomment-3709375983

Potential fix in the [src/cocotb/logging.py::default_config()](https://github.com/cocotb/cocotb/blob/v2.0.1/src/cocotb/logging.py#L51) function:

```python
def default_config(...) -> None:
    logging.basicConfig(stream=sys.stdout)

    root: logging.Logger = logging.getLogger()

    for handler in root.handlers:
        handler.addFilter(SimTimeContextFilter())
        handler.setFormatter(
            SimLogFormatter(
                reduced_log_fmt=reduced_log_fmt,
                strip_ansi=strip_ansi,
                prefix_format=prefix_format,
                multiline_indent=multiline_indent,
            )
        )

    logging.getLogger("cocotb").setLevel(logging.INFO)
    logging.getLogger("gpi").setLevel(logging.INFO)

    logging.captureWarnings(True)
```

Because pytest built-in [caplog](https://docs.pytest.org/en/stable/how-to/logging.html) fixture is adding own log handlers in the root logger to capture all log records from all loggers. Cocotb logging is always overriding list of handlers, discarding what pytest added. This must be patched/fixed.

Use cases:
- Allow to test cocotb logging module by using pytest built-in [caplog](https://docs.pytest.org/en/stable/how-to/logging.html) fixture

EDIT: Updated above example to use the `stream=sys.stdout` argument in the [logging.basicConfig()](https://docs.python.org/3/library/logging.html#logging.basicConfig) function.

tymonx:
fix(logging): keep log handlers of the root logger. Added some unit tests.
    
Fixes #5239

### DIFF
--- a/docs/source/newsfragments/5248.bugfix.rst
+++ b/docs/source/newsfragments/5248.bugfix.rst
@@ -1,0 +1,1 @@
+:func:`cocotb.logging.default_config` no longer overwrites root handlers, but configures formatting on whatever handlers are already registered.
--- a/src/cocotb/logging.py
+++ b/src/cocotb/logging.py
@@ -110,19 +110,22 @@ def default_config(
         Now captures warnings and outputs them through the logging system using
         :func:`logging.captureWarnings`.
     """
-    logging.basicConfig()
+    # Using the stream=sys.stdout argument will ensure that the root logger without any handlers
+    # will be always configured to have the stdout stream handler
+    logging.basicConfig(stream=sys.stdout)
 
-    hdlr = logging.StreamHandler(sys.stdout)
-    hdlr.addFilter(SimTimeContextFilter())
-    hdlr.setFormatter(
-        SimLogFormatter(
-            reduced_log_fmt=reduced_log_fmt,
-            strip_ansi=strip_ansi,
-            prefix_format=prefix_format,
-            multiline_indent=multiline_indent,
+    # Pytest or other frameworks can add custom log handlers, we need to ensure that log output will be consistent
+    for handler in logging.getLogger().handlers:
+        handler.addFilter(SimTimeContextFilter())
+
+        handler.setFormatter(
+            SimLogFormatter(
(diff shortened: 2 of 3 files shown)


docs/source/newsfragments/5248.bugfix.rst
```diff
@@ -0,0 +1 @@
+ The :mod:`cocotb.logging` module will no longer discard added log handlers from the root logger.
```
ktbarrett: I feel like this could be reworded to be a bit clearer.
```suggestion
 :func:`cocotb.logging.default_config` no longer overwrites root handlers, but configures formatting on whatever handlers are already registered.
```

tymonx: @ktbarrett Done. Changed as suggested.

https://github.com/cocotb/cocotb/blob/c0a9dd5d65d970624fb1d86d7aec700d1aac6548/docs/source/newsfragments/5248.bugfix.rst?plain=1#L1

ktbarrett:
Just need CI to finish.

tymonx:
There is an interesting side effect when there are two test modules with the same name:
- `tests/pytest/test_logging.py`
- `tests/test_cases/test_cocotb/test_logging.py`

When running separately, then everything is fine. With `nox`, added paths are different and `test_cocotb` is picking `tests/pytest/test_logging.py` that doesn't have any cocotb tests instead of loading  `tests/test_cases/test_cocotb/test_logging.py`.

tymonx:
I hope that renaming file will fix that:
* `tests/pytest/test_logging.py` -> `tests/pytest/test_logging_with_envs.py`

ktbarrett:
Odd... But glad it's an easy fix.

PR #5249, opened 2026-01-08
ReImplemented dev_test to avoid building cocotb twice

### LINKED ISSUE #5017: Avoid building cocotb twice in CI
When we run test that don't require simulators, we build cocotb. Then later when we run tests that require a simulator, we build cocotb again. We can probably streamline this.

ZiaCheemaGit:
Introduced var COCOTB_DEV_ALREADY_BUILT so that for
dev_test cocotb is built once and not separately each time for
dev_test_nosim and dev_test_sim.

Fixes #5017


ktbarrett:
This approach is not going to work.

See where the install command is called twice.
https://github.com/cocotb/cocotb/actions/runs/20820507517/job/59807442775?pr=5249#step:11:21
https://github.com/cocotb/cocotb/actions/runs/20820507517/job/59807442775?pr=5249#step:29:25

Perhaps the easiest way to solve this now is to fold "Run tests without simulators" step into the regular "test" step and a session to the noxfile that calls both `dev_test_sim` and `dev_test_nosim`. Then change `generate-envs` so that the `nox_session_test_sim` input is either the "with nosim" or "without nosim" session and remove the `nox_session_test_nosim` input.

Sorry, I'm not sure what I was thinking. This is definitely not a "good first issue".

ZiaCheemaGit:
@ktbarrett I still might work on this can u remove good first issue label I cant help it.

ZiaCheemaGit:
what is this issue why coverage report for all the tests is failing. I will recommit to confirm this is not a random false positive.

ZiaCheemaGit:
Now cocotb is built only once. A single nox session dev_test is run which runs both dev_test_nosim and dev_test_sim. For dev_test_sim env variables of generate_envs are directly used

ZiaCheemaGit:
@ktbarrett is it a right approach?

ZiaCheemaGit:
@ktbarrett also if this is the right approach one thing is missing. we have to add a release nox session which will run both release_test_sim and release_test_nosim because now regression_tests only take one input

ZiaCheemaGit:
@ktbarrett just a reminder for review

.github/workflows/regression-tests.yml
```diff
@@ -260,7 +251,7 @@ jobs:
       id: windowstesting
       continue-on-error: ${{matrix.may-fail || false}}
       run: |
-        nox -k "${{ inputs.nox_session_test_sim }} and ${{ matrix.sim }} and ${{ matrix.lang }}"
+        nox -s ${{ inputs.nox_session_test }}
```
ktbarrett, line 255: Lets leave these as is.

noxfile.py
```diff
@@ -150,23 +159,7 @@ def dev_build(session: nox.Session) -> None:
 def dev_test(session: nox.Session) -> None:
     """Run all development tests as configured through environment variables."""
 
-    dev_test_sim(session, sim=None, toplevel_lang=None, gpi_interface=None)
-    dev_test_nosim(session)
-    dev_coverage_combine(session)
-
-
-@nox.session
-@nox.parametrize("sim,toplevel_lang,gpi_interface", simulator_support_matrix())
```
ktbarrett, line 159: Lets leave the parameterized stuff here for now.

noxfile.py
```diff
@@ -294,11 +323,6 @@ def dev_test_sim(
 def dev_test_nosim(session: nox.Session) -> None:
     """Run the simulator-agnostic tests against a cocotb development version."""
 
-    configure_env_for_dev_test(session)
-
-    session.install(*test_deps, *coverage_deps)
-    build_cocotb_for_dev_test(session, editable=False)
```
ktbarrett, line 300: Leave this here. Just split out everything below this into a separate function and call that function here and in `dev_test`.

ktbarrett (commented):
I like simplifying the workflow to only call one function. But the approach in the noxfile is a bit off.

Just move the contents of the nosim and coverage collection sessions into separate functions (that don't contain installing the dependencies) and then from the new master `dev_test` and `release_test` sessions call those newly split out functions.

noxfile.py
ZiaCheemaGit: Are you saying that dev_test_sim should remain parametried. It already is. This confusion might be due to the git diff gui here.  If you pointing to something else I would need more info then.

noxfile.py
```diff
@@ -179,6 +172,42 @@ def dev_test_sim(
...
+        elif sim == "nvc":
+            gpi_interface = "vhpi"
+        else:
+            gpi_interface = "vhpi"
+
+    dev_test_sim(session, sim, toplevel_lang, gpi_interface)
+
+
+@nox.session
+@nox.parametrize("sim,toplevel_lang,gpi_interface", simulator_support_matrix())
```
ZiaCheemaGit, line 194: See here dev_test_sim is maintained parametrized

.github/workflows/regression-tests.yml
ZiaCheemaGit: I am sorry but I dont understand what you are saying here. Are u suggesting I should revert my change or you are saying my change is ok and leave it in that manner.

ZiaCheemaGit:
What I understood is only edits made in noxfile are to be changed. @ktbarrett yes we can just split out the nosim but for coverage combine I dont understand what is going on eg in regression tests file a command is ran already to run that session dev_coverage_combine so then maybe we dont need to call it in `dev_test` also right?
ALso For dev_coverage_combine we cannot split it like dev_test_nosim because its pre reqs are not met before this point according to my understanding Maybe they do if yes then we can split it like dev_test_nosim 
image

image

.github/workflows/regression-tests.yml
ZiaCheemaGit: nox -s is used in newest commit

noxfile.py
```diff
@@ -505,6 +557,14 @@ def release_install(session: nox.Session) -> None:
     session.install(*test_deps)
 
 
+@nox.session
+def release_test(session: nox.Session) -> None:
+    """Test a release version of cocotb with and without a simulator"""
+    # At release build tests cocotb shouldn't built twice
+    # TODO
+    pass
+
```
ZiaCheemaGit, line 566: I can work on this once other changes are compliable

noxfile.py
```diff
@@ -150,9 +159,35 @@ def dev_build(session: nox.Session) -> None:
...
+        elif sim == "questa":
+            gpi_interface = "fli"
+        elif sim == "ghdl":
+            gpi_interface = "vpi"
+        elif sim == "nvc":
+            gpi_interface = "vhpi"
+        else:
+            gpi_interface = "vhpi"
+
+    dev_test_sim_runner(session, sim, toplevel_lang, gpi_interface)
```
ZiaCheemaGit, line 191: Here call to session dev_coverage_combine is not required right? because it is already called by CI in regression tests [here](https://github.com/cocotb/cocotb/blob/522b05caf9e283becef2a94bb3ec923033a10081/.github/workflows/regression-tests.yml#L297). But that also depends on the env vars so maybe we do if thats the case then do we need [coverage_combine_runner](https://github.com/cocotb/cocotb/pull/5249/changes#diff-f7a16a65f061822bcc73b8296f4dc837353d379d8d9cc5307982cb6941442835R387) or [coverge_combine](https://github.com/cocotb/cocotb/pull/5249/changes#diff-f7a16a65f061822bcc73b8296f4dc837353d379d8d9cc5307982cb6941442835L332)

ZiaCheemaGit:
@ktbarrett If PR #5295 can resolve the problem at hand then please you can close this PR

PR #5152, opened 2025-11-18
Add signed and bit sizes to IntegerObjects

ktbarrett:
This will make it possible to correct constrain the input values and also convert any value to the appropriate signed/unsigned value in Python.

PR #5263, opened 2026-01-11
Improve settings handles with `int`

ktbarrett:
Depends on #5259. Follow-on to #5152.

* Removes support for LogicArray and str that were added in #5152.
   * Those are actually not needed, the justification for their addition was deemed invalid.
   * Enums are ints, and maybe one day we will support setting with string to support setting by member name.
* Uses the `is_signed` property to determine reasonable bounds for input integers.
   * LogicArrayObject uses `(INT_MIN, UINT_MAX)` for backwards compatibility.
   * IntegerObject and EnumObject use the signed field to determine appropriate bounds.
   * If the simulator doesn't support is_signed checking, it default to `(INT_MIN, UINT_MAX)` like LogicArray.

I'll do one more follow-on to support getting the value of integers larger than 32-bits from IntegerObject and EnumObject. And also add logic to ensure that getting values respects is_signed.

### TODO
- [x] newsfrag


ktbarrett:
So with Questa and Xcelium VHPI, writing a negative value to a `natural` or `positive` is causing an internal range constraint violation. The Questa FLI failure is something else, but not sure.

tests/test_cases/test_integers/test_integers.py
```diff
@@ -48,17 +53,56 @@ async def test_int_verilog(
     assert len(handle) == width
     assert handle.is_signed is is_signed
 
+    # For backwards compatibility, LogicArrayObjects always use (INT_MIN, UNIT_MAX) for bounds.
```
cmarqu:
```suggestion
    # For backwards compatibility, LogicArrayObjects always use (INT_MIN, UINT_MAX) for bounds.
```

tests/test_cases/test_integers/test_integers.py
```diff
@@ -68,10 +112,57 @@ async def test_int_verilog(
...
+
+    # Use bounds of the specific subtype for random testing.
+    # Otherwise we get fun range constraint violations which cause errors.
+    for _ in range(100):
+        handle.value = random.randint(min_value, max_value)
+        await Timer(1)
+
+    last_value = handle.value
+
+    # For backwards compatibility, LogicArrayObjects always use (INT_MIN, UNIT_MAX) for bounds.
```
cmarqu:
```suggestion
    # For backwards compatibility, LogicArrayObjects always use (INT_MIN, UINT_MAX) for bounds.
```

ktbarrett:
Went from using shifts to compute bounds to using `2 ** ...` because that works in the case of null vectors whereas `1 << (len(self) - 1)` break when `len(self)` is 0.

cmarqu:
> Went from using shifts to compute bounds to using `2 ** ...` because that works in the case of null vectors whereas `1 << (len(self) - 1)` break when `len(self)` is 0.

A test where https://hypothesis.readthedocs.io/ would be good for.

PR #5271, opened 2026-01-14
Add tests for getting values from `IntegerObject` and `EnumObject`

ktbarrett:
Ensures that sign and size are respected when getting values.

This didn't require any code change in `cocotb.handle` since integer types max out at 64 bits, the GPI get value function is 64-bits signed, and the 64-bit integer types in HDLs are all signed, so we don't need to fix up negative values on unsigned types getting them back from a signed interface.

This ends the integer object improvement cycle for now.

### DIFF
--- a/src/cocotb/handle.py
+++ b/src/cocotb/handle.py
@@ -1511,7 +1511,15 @@ class EnumObject(
 
         See :class:`EnumObject` for details on what :class:`int` values correspond to which enumeration values.
         """
-        return self._handle.get_signal_val_long()
+        if len(self) <= 32:
+            res = self._handle.get_signal_val_long()
+        else:
+            res = int(self._handle.get_signal_val_binstr(), 2)
+        if res > self._max_val:
+            res -= 1 << len(self)
+        elif self._handle.get_signed() == 0 and res < 0:
+            res += 1 << len(self)
+        return res
 
     def set(
         self,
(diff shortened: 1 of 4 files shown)


ktbarrett:
I love how all of the FOSS simulators work with no issue, but all of the commercial sims can't keep up.

ktbarrett:
I'm guessing the commerical sims normalize `-1` to the unsigned bit pattern truncated to the required size for storage. So for byte you write a `-1` and read back a `255`. I suppose that prevents having to normalize it on every get operation.

ktbarrett:
Those failures were likely caused by setting the value of the handle to 0 before running the failure mode tests. For the positive inputs it isn't legal to have a 0 written to it.

src/cocotb/handle.py
```diff
@@ -1277,48 +1277,35 @@ def _set_value(
...
             value_ = str(value)
 
         else:
             raise TypeError(
                 f"Unsupported type for value assignment: {type(value)} ({value!r})"
             )
 
+        if len(value_) != len(self):
+            raise ValueError(
+                f"cannot assign value of length {len(value_)} to handle of length {len(self)}"
```
cmarqu:
```suggestion
                f"Cannot assign value of length {len(value_)} to handle of length {len(self)}"
```

ktbarrett:
Ugh... 123 isn't valid either because of the `[-100, 100]` object in VHDL. Should be fine now.

ktbarrett:
There are certain code paths we can't hit because enums top out at 32 bits in our testing (technically simulators can return `vpiEnum` for enum logic vectors that are larger than 32 bits). And other paths aren't taken because all integers and enums we are testing are signed.

cmarqu (approved):

PR #5281, opened 2026-01-25
Refactor `TestManager`

ktbarrett:
Renames `RunningTest` to `TestManager` and does some initial refactoring to return `ExceptionGroup` on multiple failures, rather than just printing them.

Part 1 of N.


src/cocotb/_test_manager.py
```diff
@@ -44,60 +47,60 @@ def __init__(
...
-
-        if not isinstance(self._outcome, Error) and self._shutdown_errors:
-            return self._shutdown_errors[0]
-        return self._outcome
+    def result(self) -> BaseException | None:
+        if self._excs:
+            if len(self._excs) == 1:
+                return self._excs[0]
+            else:
+                return BaseExceptionGroup("Multiple exceptions in test", self._excs)
```
marlonjames, line 79: Are there any existing tests to hit this? codecov is showing uncovered.

marlonjames (approved):

ktbarrett:
So since we are doing `_cancel_now` we should be able to see errors. I need to write a test for ignoring a `CancelledError` during shutdown.

I think this was not done before because we had no way to test the return until we started using `ExceptionGroup`.

ktbarrett:
You know maybe we still don't. `expect_error=ExceptionGroup` isn't sufficient. I'm wondering if we need a better system for describing multi-error xfails.

ktbarrett:
Either way all of that will be handled in the follow on.

PR #5284, opened 2026-01-26
chore: pytest plugin overview and tests summary

### LINKED ISSUE #5276: pytest reports `cocotb_runner` as a test in summary
I was trying out `cocotb`'s new `pytest` integration, and I noticed that the test runner marked with `@pytest.mark.cocotb_runner` is also included in the test report summary when using `verilator`, even though it doesn't show as a test when running `pytest --collect-only`.

Here's a minimal example with a dummy module:

```py
# test_my_module.py
import os

import pytest
from cocotb_tools.pytest.hdl import HDL


@pytest.fixture(name="my_module")
def my_module_fixture(hdl: HDL) -> HDL:
    os.environ["SIM"] = "verilator"
    hdl.sources = ["my_module.sv"]
    hdl.toplevel = "my_module"
    return hdl.build()


@pytest.mark.cocotb_runner
def hdl_runner(my_module: HDL) -> None:
    my_module.test()


async def test_pass(dut):
    assert True


async def test_fail(dut):
    assert False
```

When running `pytest --collect-only`:

```
collected 2 items                                                                                                                                                                            

<Dir test_my_module>
  <Module test_my_module.py>
    <Runner hdl_runner>
      <Testbench test_my_module>
        <Function test_pass>
        <Function test_fail>
```

When actually running the tests:

```
========================== short test summary info ==========================
FAILED test_my_module.py::hdl_runner::test_fail - assert False
FAILED test_my_module.py::hdl_runner - SystemExit: 1
======================== 2 failed, 1 passed in 0.46s ========================
```

You can see that `hdl_runner` is considered as one of the tests, and I get 2 failed tests instead of 1.

I think this can be solved by implementing a custom `pytest_terminal_summary` that ignores the `hdl_runner` test status:
https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_terminal_summary

tymonx:
Fixes #5276 by explaning pytest plugin architecture where cocotb runners are treated as test functions.

It also shows cocotb runners as `<Function>` item when invoking `pytest` with the `--collect-only` option. This will fix number of collected items (tests) in collect summary info and match with number of reported tests in test summary info.

Added SVG is representing general overview of pytest plugin:

image

Added XML is exported from the [diagrams.net](https://www.drawio.com/) ([draw.io](https://www.drawio.com/)).


tymonx:
Note to myself. Never buy macOS, is not stable in CI 😛

docs/source/pytest.rst
```diff
@@ -562,6 +570,40 @@ used to configure cocotb testing environment, can be listed by invoking `pytest`
...
+* Collect all tests, including cocotb runners and cocotb tests, when invoking ``pytest`` **with** the ``--collect-only`` option.
+  This will allow to properly visualize relationship between cocotb runner, test module (testbench) and cocotb test in
+  the ``pytest`` summary report about collected tests (items).
+* Collect all non-cocotb tests, including cocotb runners, when invoking ``pytest`` **without** the ``--collect-only`` option.
+  This will allow to execute cocotb runners as test functions that will build HDL design and run simulation with cocotb tests.
+* Collect serialized test reports sent by cocotb tests via IPC (Inter-Process Communication) from running simulation process.
+
+.. note::
+
+   Cocotb runners are treated by plugin as test functions and they will be reported in pytest collect and test summary info.
```
ktbarrett:
```suggestion
   Cocotb runners are treated by the plugin as test functions and they will be reported in pytest collect and test summary info.
```

docs/source/pytest.rst
```diff
@@ -562,6 +570,40 @@ used to configure cocotb testing environment, can be listed by invoking `pytest`
...
+   This will allow to report compilation/elaboration of HDL design and simulation runtimes.
+
+The second part of the :py:mod:`cocotb_tools.pytest.plugin` is performed within simulator process. It will:
```
ktbarrett:
```suggestion
The second part of the :py:mod:`cocotb_tools.pytest.plugin` is performed within the simulator process. It will:
```

docs/source/pytest.rst
```diff
@@ -326,11 +326,19 @@ Example output::
...
           <Testbench test_sample_module>
             <Function test_dut_feature_1>
             <Function test_dut_feature_2>
 
+.. note::
+
+   Cocotb runner is treated by plugin as test function that will build HDL design and run simulation with cocotb tests.
+   Unfortunately pytest ``<Function>`` item cannot have other sub-items like test functions or test modules.
+   Plugin is using additional ``<Runner>`` node to visualize relationship between cocotb runner (``<Function>``),
+   test modules (``<Testbench>``) and cocotb tests (``<Function>``).
```
ktbarrett:
```suggestion
   Functions which build HDL designs and run tests using the cocotb Runner are treated by the plugin as test functions.
   Unfortunately, pytest ``<Function>`` items cannot have other sub-items like cocotb test functions or cocotb test modules.
   So the plugin uses the ``<Runner>`` node to visualize relationship between the cocotb Runner (``<Function>``),
   cocotb test modules (``<Testbench>``) and cocotb tests (``<Function>``).
```

ktbarrett:
@tymonx The flakey Verilator/macOS test is likely due to Verilator and the fact we are building two different designs in the same build directory. Verilator seems to not work well with that.

docs/source/pytest.rst
tymonx: @ktbarrett Applied as suggested:

https://github.com/tymonx/cocotb/blob/7d0e802c812a24dbe11782a06b6ed225be50a984/docs/source/pytest.rst?plain=1#L337-L340

docs/source/pytest.rst
tymonx: @ktbarrett Applied as suggested:

https://github.com/tymonx/cocotb/blob/7d0e802c812a24dbe11782a06b6ed225be50a984/docs/source/pytest.rst?plain=1#L599

docs/source/pytest.rst
tymonx: @ktbarrett Applied as suggested:

https://github.com/tymonx/cocotb/blob/7d0e802c812a24dbe11782a06b6ed225be50a984/docs/source/pytest.rst?plain=1#L596

ktbarrett:
Interesting the Xcelium is failing the `test_integers` test, but not your issue. Nothing you've done here would affect that.

PR #5243, opened 2026-01-06
feat(pytest): plugin hooks for external build systems

tymonx:
- Improved support for external build systems via pytest plugin hooks
- Added new `pytest_cocotb_hdl_build` hook that that can be implemented by other plugins to invoke external build system. Used to build HDL design.
- Added new `pytest_cocotb_hdl_test` hook that can be implemented by other plugins to invoke external build system. Used to test HDL design.
- Updated documentation how to use hooks with external build system
- Removed `pytest_cocotb_make_runner` hook. We don't want to promote or tighly integrate `Runner` with any plugin including also cocotb plugin
- Moved `runner.build()` to hook as default hook implementation
- Moved `runner.test()` to hook as default hook implementation
- Decoupled `Runner` from `HDL` and plugin
- Removed all arguments in `HDL.build()` and `HDL.test()` methods. Pytest hooks mostly use 1-4 arguments as object classes, where all attributes are defined in passed objects. Having 20-30 different arguments in method could easily break contract between hook definition (spefication) and hook implementation in other plugins
- Removed `hdl_session` fixture. `Runner` doesn't support building shared HDL libraries in the same build directory for different tests. External build systems may handle that better because mostly they support target dependencies
- Top level can be also automatically determined based on provided list of source files
- Simplify `HDL` class
- If the `HDL.build()` method was not called explicitly, then it will be invoked implicitly by the `HDL.test()` method

WIP:

- External `pytest-cocotb-fusesoc` plugin as golden reference how to use external build system ([fusesoc](https://github.com/olofk/fusesoc) in this case) with cocotb + pytest

Closes #5241

PR #5287, opened 2026-01-30
Separate cocotb runners in pytest summary

### LINKED ISSUE #5286: Improve pytest test summary for cocotb runners
Currently, `pytest` doesn't differentiate between `cocotb_runner` tests and actual tests when reporting the test summary at the end of the session. It would be nice to differentiate between test and runner status. 

I'm quoting a sample `conftest.py` that I've proposed in another issue: https://github.com/cocotb/cocotb/issues/5276#issuecomment-3804515182. However, it needs more work to handle different `pytest` traceback printing options:

https://docs.pytest.org/en/stable/how-to/output.html

> I thought I'd try playing around with `pytest` terminal summary, and here's what I ended up with in my `conftest.py`:
> 
> ```py
> import pytest
> 
> 
> def pytest_report_teststatus(report, config):
>     if ("cocotb_runner" in report.keywords) and (report.when == "call"):
>         if report.failed:
>             return pytest.TestShortLogReport(
>                 category="failed cocotb runner",
>                 letter="CF",
>                 word=("FAILED COCOTB RUNNER", {"red": True}),
>             )
>         else:
>             return pytest.TestShortLogReport(
>                 category="passed cocotb runner",
>                 letter=".",
>                 word=("PASSED COCOTB RUNNER", {"green": True}),
>             )
> 
> 
> @pytest.hookimpl(hookwrapper=True, tryfirst=True)
> def pytest_terminal_summary(terminalreporter, exitstatus, config):
>     # Print cocotb runner failures BEFORE test failures
>     # To print after test failures (current behaviour), just remove the `hookimpl()`
>     # decorator, and the last `yield` statement
>     reports = terminalreporter.stats.get("failed cocotb runner", [])
>     if reports:
>         terminalreporter.write_sep("=", "COCOTB RUNNER FAILURES")
>         for rep in reports:
>             if config.option.tbstyle != "no":
>                 terminalreporter.write_sep("_", rep.head_line or rep.nodeid, red=True)
>                 rep.toterminal(terminalreporter._tw)
> 
>             # Print captured stdout/stderr/log if available
>             for section in rep.sections:
>                 header, content = section
>                 terminalreporter.write_sep("-", header)
>                 terminalreporter.write_line(content)
>     yield
> ```
> 
> With this, I end up with the following report in case all tests pass:
> ```
> ========================================== test session starts ===========================================
> platform linux -- Python 3.13.2, pytest-9.0.2, pluggy-1.6.0
> rootdir: /home/foo/projects/test_cocotb_pytest/test_temp.py
> configfile: pyproject.toml
> plugins: test-cocotb-pytest-0.1.0
> collected 1 item                                                                                         
> 
> test_temp.py .                                                                                   [300%]
> 
> =============================== 2 passed, 1 passed cocotb runner in 0.45s ================================
> ```
> 
> And we get this in case there's a failure:
> ```
> ========================================== test session starts ===========================================
> platform linux -- Python 3.13.2, pytest-9.0.2, pluggy-1.6.0
> rootdir: /home/foo/projects/test_cocotb_pytest/test_temp.py
> configfile: pyproject.toml
> plugins: test-cocotb-pytest-0.1.0
> collected 1 item                                                                                         
> 
> test_temp.py CF                                                                                  [300%]
> 
> ========================================= COCOTB RUNNER FAILURES =========================================
> _______________________________________________ hdl_runner _______________________________________________
> 
> hdl = <cocotb_tools.pytest.hdl.HDL object at 0x7cef22950190>
> 
>     @pytest.mark.cocotb_runner
>     def hdl_runner(hdl: HDL) -> None:
>         os.environ["SIM"] = "verilator"
>         hdl.sources = ["my_module.sv"]
>         hdl.toplevel = "my_module"
>         hdl.build()
>       hdl.test()
> 
> test_temp.py:18: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> <TRUNCATED>
> E       SystemExit: 1
> 
> .venv/lib/python3.13/site-packages/cocotb_tools/pytest/hdl.py:416: SystemExit
> ------------------------------------------ Captured stdout call ------------------------------------------
> <TRUNCATED>
> 
> ------------------------------------------- Captured log call --------------------------------------------
> <TRUNCATED>
> 
> ================================================ FAILURES ================================================
> _______________________________________________ test_fail ________________________________________________
> 
> dut = HierarchyObject(my_module)
> 
>     async def test_fail(dut):
>       assert False
> E       assert False
> 
> /home/foo/projects/test_cocotb_pytest/test_temp.py:31: AssertionError
> ======================================== short test summary info =========================================
> FAILED test_temp.py::hdl_runner::test_fail - assert False
> ========================== 1 failed, 1 passed, 1 failed cocotb runner in 0.53s ===========================
> ```

Ehab-Ibrahim:
In this PR, I separate cocotb runners from the rest of the tests in `pytest`'s summary that is reported at the end of the session. I had to look in `pytest`'s source code to see how it handles different `--tb` and `--show-capture` options when reporting failures, and I found out that there's already a `TerminalReporter` class method that handles this for us:

https://github.com/pytest-dev/pytest/blob/3d10b5148e03eb82b3ee29181dbdc73cf82699e2/src/_pytest/terminal.py#L1166-L1168
https://github.com/pytest-dev/pytest/blob/3d10b5148e03eb82b3ee29181dbdc73cf82699e2/src/_pytest/terminal.py#L1175-L1199

I ended up reusing `summary_failures_combined()`, which made the PR fairly simple.

Closes: https://github.com/cocotb/cocotb/issues/5286

### DIFF
--- a/src/cocotb_tools/pytest/plugin.py
+++ b/src/cocotb_tools/pytest/plugin.py
@@ -13,7 +13,7 @@ from argparse import ArgumentParser
 from collections.abc import Generator, Iterable, Mapping
 from pathlib import Path
 from time import time
-from typing import Any, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from pluggy import Result
 from pytest import (
@@ -66,7 +66,9 @@ _ENTRY_POINT: str = ",".join(
 if TYPE_CHECKING:
     from typing import TypeAlias
 
-    TestStatus: TypeAlias = TestShortLogReport | tuple[str, str, str | tuple[str, Mapping[str, bool]]]
+    TestStatus: TypeAlias = (
+        TestShortLogReport | tuple[str, str, str | tuple[str, Mapping[str, bool]]]
+    )
 
 
 def _to_timescale(value: str) -> tuple[str, str]:
--- a/src/cocotb_tools/pytest/plugin.py
+++ b/src/cocotb_tools/pytest/plugin.py
@@ -10,14 +10,16 @@ import inspect
 import shlex
 import textwrap
 from argparse import ArgumentParser
-from collections.abc import Generator, Iterable
+from collections.abc import Generator, Iterable, Mapping
 from pathlib import Path
 from time import time
 from typing import Any
 
+from pluggy import Result
 from pytest import (
     Class,
     Collector,
+    CollectReport,
     Config,
     ExitCode,
     FixtureRequest,
@@ -61,16 +63,7 @@ _ENTRY_POINT: str = ",".join(
         "cocotb_tools.pytest._init:run_regression",
     )
 )
-FAILED_COCOTB_RUNNER = TestShortLogReport(
-    category="failed cocotb runner",
-    letter="CF",
-    word=("FAILED COCOTB RUNNER", {"red": True}),
-)
-PASSED_COCOTB_RUNNER = TestShortLogReport(
-    category="passed cocotb runner",
-    letter=".",
-    word=("PASSED COCOTB RUNNER", {"green": True}),
-)
+_TestStatus = TestShortLogReport | tuple[str, str, str | tuple[str, Mapping[str, bool]]]
 
 
 def _to_timescale(value: str) -> tuple[str, str]:
@@ -714,12 +707,27 @@ def _is_cocotb_test_report(item: Any) -> bool:
     return isinstance(item, TestReport) and getattr(item, "cocotb", False)
 
 
-def pytest_report_teststatus(report, config):
-    if ("cocotb_runner" in report.keywords) and (report.when == "call"):
-        if report.failed:
-            return FAILED_COCOTB_RUNNER
+@hookimpl(hookwrapper=True, trylast=True)
+def pytest_report_teststatus(
+    report: CollectReport | TestReport, config: Config
+) -> Generator[None, Result[_TestStatus], None]:
+    result: Result[_TestStatus] = yield
+    status: _TestStatus = result.get_result()
+
+    if (
+        isinstance(report, TestReport)
+        and ("cocotb_runner" in report.keywords)
+        and (report.when == "call")
+    ):
+        category, letter, word = status
+        category = f"{category} cocotb runner"
+        if isinstance(word, str):
+            word = f"{word} COCOTB RUNNER"
         else:
-            return PASSED_COCOTB_RUNNER
+            word = (f"{word[0]} COCOTB RUNNER", word[1])
+        result.force_result(
+            TestShortLogReport(category=category, letter=letter, word=word)
+        )
 
 
 @hookimpl(hookwrapper=True, tryfirst=True)
@@ -730,8 +738,8 @@ def pytest_terminal_summary(
 ) -> Generator[None]:
     # Print cocotb runner failures before test failures
     terminalreporter.summary_failures_combined(
-        which_reports=FAILED_COCOTB_RUNNER.category,
-        sep_title=FAILED_COCOTB_RUNNER.word[0] + "S",
+        which_reports="failed cocotb runner",
+        sep_title="COCOTB RUNNER FAILURES",
         style=config.option.tbstyle,
     )
     # Continue with the rest of pytest terminal summary
--- a/src/cocotb_tools/pytest/plugin.py
+++ b/src/cocotb_tools/pytest/plugin.py
@@ -10,7 +10,7 @@ import inspect
 import shlex
 import textwrap
 from argparse import ArgumentParser
-from collections.abc import Iterable
+from collections.abc import Generator, Iterable
 from pathlib import Path
 from time import time
 from typing import Any
@@ -28,6 +28,7 @@ from pytest import (
     PytestPluginManager,
     TerminalReporter,
     TestReport,
+    TestShortLogReport,
     fixture,
     hookimpl,
     mark,
@@ -59,6 +60,16 @@ _ENTRY_POINT: str = ",".join(
         "cocotb._init:init_package_from_simulation",
         "cocotb_tools.pytest._init:run_regression",
     )
+)
+FAILED_COCOTB_RUNNER = TestShortLogReport(
+    category="failed cocotb runner",
+    letter="CF",
+    word=("FAILED COCOTB RUNNER", {"red": True}),
+)
+PASSED_COCOTB_RUNNER = TestShortLogReport(
+    category="passed cocotb runner",
+    letter=".",
+    word=("PASSED COCOTB RUNNER", {"green": True}),
 )
 
 
@@ -703,12 +714,29 @@ def _is_cocotb_test_report(item: Any) -> bool:
     return isinstance(item, TestReport) and getattr(item, "cocotb", False)
 
 
-@hookimpl(tryfirst=True)
+def pytest_report_teststatus(report, config):
+    if ("cocotb_runner" in report.keywords) and (report.when == "call"):
+        if report.failed:
+            return FAILED_COCOTB_RUNNER
+        else:
+            return PASSED_COCOTB_RUNNER
+
+
+@hookimpl(hookwrapper=True, tryfirst=True)
 def pytest_terminal_summary(
     terminalreporter: TerminalReporter,
     exitstatus: ExitCode,
     config: Config,
-) -> None:
+) -> Generator[None]:
+    # Print cocotb runner failures before test failures
+    terminalreporter.summary_failures_combined(
+        which_reports=FAILED_COCOTB_RUNNER.category,
+        sep_title=FAILED_COCOTB_RUNNER.word[0] + "S",
+        style=config.option.tbstyle,
+    )
+    # Continue with the rest of pytest terminal summary
+    yield
+
     if not config.option.cocotb_summary:
         return
 
--- a/src/cocotb_tools/pytest/plugin.py
+++ b/src/cocotb_tools/pytest/plugin.py
@@ -13,7 +13,7 @@ from argparse import ArgumentParser
 from collections.abc import Generator, Iterable, Mapping
 from pathlib import Path
 from time import time
-from typing import Any
+from typing import Any, TYPE_CHECKING
 
 from pluggy import Result
 from pytest import (
@@ -63,7 +63,10 @@ _ENTRY_POINT: str = ",".join(
         "cocotb_tools.pytest._init:run_regression",
     )
 )
-_TestStatus = TestShortLogReport | tuple[str, str, str | tuple[str, Mapping[str, bool]]]
+if TYPE_CHECKING:
+    from typing import TypeAlias
+
+    TestStatus: TypeAlias = TestShortLogReport | tuple[str, str, str | tuple[str, Mapping[str, bool]]]
 
 
 def _to_timescale(value: str) -> tuple[str, str]:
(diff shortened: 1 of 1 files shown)


src/cocotb_tools/pytest/plugin.py
```diff
@@ -60,6 +61,16 @@
         "cocotb_tools.pytest._init:run_regression",
     )
 )
+FAILED_COCOTB_RUNNER = TestShortLogReport(
+    category="failed cocotb runner",
+    letter="CF",
```
Ehab-Ibrahim: If we get a failure, then it is reported as `CF`: "Cocotb Failure". Usually, test status are reported as one letter; "F" for failures and "." for passes. I'm considering just marking this as "F", because having 2 letters might give the impression that these are 2 tests: One test with status C, and another with status F

src/cocotb_tools/pytest/plugin.py
```diff
@@ -703,12 +714,29 @@ def _is_cocotb_test_report(item: Any) -> bool:
     return isinstance(item, TestReport) and getattr(item, "cocotb", False)
 
 
-@hookimpl(tryfirst=True)
+def pytest_report_teststatus(report, config):
+    if ("cocotb_runner" in report.keywords) and (report.when == "call"):
+        if report.failed:
+            return FAILED_COCOTB_RUNNER
+        else:
+            return PASSED_COCOTB_RUNNER
```
Ehab-Ibrahim: Now that I think about it, I'm only assuming "Pass" or "Fail" status. This part doesn't handle other test statuses like `skip`, `xpass`, `xfail`, and `error`. I can extend to handle those as well.

ktbarrett: Right now there is no support for other outcomes beyond PASS/FAIL. That will occur some time in the current dev cycle when I add support for `pytest.skip` and `pytest.xfail` to the existing RegressionManager (and the current pytest RM will need it as well).

tymonx: I think using pytest hook wrapper will be easier here and more agnostic to other plugins that are modifying test status as well like replacing test status letters with... emojis 😸 

```python
_TestStatus = TestShortLogReport | tuple[str, str, str | tuple[str, Mapping[str, bool]]]

@hookimpl(hookwrapper=True, trylast=True)
def pytest_report_teststatus(
    report: CollectReport | TestReport, config: Config
) -> Generator[None, _TestStatus, _TestStatus]:
    status: _TestStatus = yield

    if isinstance(report, TestReport) and "cocotb_runner" in report.keywords:
        if isinstance(status, tuple):
            status = TestShortLogReport(category=status[0], letter=status[1], word=status[2])

        if isinstance(status, TestShortLogReport):
            if status.category:
                status.category = f"{status.category} cocotb runner"

            if isinstance(status.word, tuple):
                if status.word[0]:
                    status.word = (f"{status.word[0]} COCOTB RUNNER", status.word[1])

            elif status.word:
                status.word = f"{status.word} COCOTB RUNNER"

    return status
```

src/cocotb_tools/pytest/plugin.py
tymonx: I agree about that extra `C` letter. It may be confusing. Adding a cocotb runner text to status category and status word should be enough.

src/cocotb_tools/pytest/plugin.py
```diff
@@ -703,12 +714,29 @@ def _is_cocotb_test_report(item: Any) -> bool:
...
 def pytest_terminal_summary(
     terminalreporter: TerminalReporter,
     exitstatus: ExitCode,
     config: Config,
-) -> None:
+) -> Generator[None]:
+    # Print cocotb runner failures before test failures
+    terminalreporter.summary_failures_combined(
+        which_reports=FAILED_COCOTB_RUNNER.category,
+        sep_title=FAILED_COCOTB_RUNNER.word[0] + "S",
```
tymonx: Pytest is using [FAILURES](https://github.com/pytest-dev/pytest/blob/3d10b5148e03eb82b3ee29181dbdc73cf82699e2/src/_pytest/terminal.py#L1168) word here. Maybe we should align 🤔 

```suggestion
        sep_title="COCOTB RUNNER FAILURES",
```

src/cocotb_tools/pytest/plugin.py
```diff
@@ -703,12 +714,29 @@ def _is_cocotb_test_report(item: Any) -> bool:
...
+@hookimpl(hookwrapper=True, tryfirst=True)
 def pytest_terminal_summary(
     terminalreporter: TerminalReporter,
     exitstatus: ExitCode,
     config: Config,
-) -> None:
+) -> Generator[None]:
+    # Print cocotb runner failures before test failures
+    terminalreporter.summary_failures_combined(
+        which_reports=FAILED_COCOTB_RUNNER.category,
```
tymonx: Considering https://github.com/cocotb/cocotb/pull/5287/changes#r2751144122 and https://github.com/cocotb/cocotb/pull/5287/changes#r2751157277 we could use a simple string here:

```suggestion
        which_reports="failed cocotb runner",
```

src/cocotb_tools/pytest/plugin.py
Ehab-Ibrahim: > Right now there is no support for other outcomes beyond PASS/FAIL. That will occur some time in the current dev cycle when I add support for `pytest.skip` and `pytest.xfail` to the existing RegressionManager (and the current pytest RM will need it as well).

The statuses I'm referring to right now applies to the runners, which then invokes cocotb's regression manager in another process. To my understanding, we can still mark runner tests as `skip/xpass/xfail`, right?

Ehab-Ibrahim: > I think using pytest hook wrapper will be easier here and more agnostic to other plugins that are modifying test status as well like replacing test status letters with... emojis 😸

Good point, I haven't considered other plugins modifying `pytest_report_teststatus()`. I'll apply your proposed changes in another commit.

tymonx: > Good point, I haven't considered other plugins modifying pytest_report_teststatus(). I'll apply your proposed changes in another commit.

@Ehab-Ibrahim Try with this plugin [pytest-sugar](https://pypi.org/project/pytest-sugar/). It is adding colored unicodes in the [pytest_report_teststatus](https://github.com/Teemu/pytest-sugar/blob/81335030b95741d90ae489f9216759c009854fa9/pytest_sugar.py#L215) hook like green tick `✓` for passed tests.

> To my understanding, we can still mark runner tests as skip/xpass/xfail, right?

https://github.com/cocotb/cocotb/pull/5287#discussion_r2751144122 should cover this without thinking about that :slightly_smiling_face:

There are some valid use case for `skip`. User can use `@pytest.mark.parametrize()` to generate N variants of cocotb runners and use `pytest.skip` to exclude some invalid combination. Or just use `@pytest.mark.skip()` to exclude cocotb runner because of something (specific simulator not present in current environment).

I cannot find a valid use case for `xpass/xfail`. There are few exceptions thrown from cocotb runner. But it doesn't make sense to capture them :thinking: 

This PR https://github.com/cocotb/cocotb/pull/5243 is defining custom hooks that will allow to add custom/external build systems. This will expand use cases for `xpass/xfail`. I'm currently using it for [FuseSoC](https://gitlab.com/tymonx/pytest-cocotb-fusesoc) (remark: this is more like a PoC for now) to do compilation/elaboration/run simulation where pytest is discovering and running tests.

Ehab-Ibrahim:
Thanks for the review @tymonx. I've now applied your proposed changes and tested with `pyest-sugar` and `pytest-emoji` plugins. However, with `pytest-sugar`, the short test summary behaves differently than vanilla `pytest`. In `pytest-sugar`, it does not respect `failed cocotb runner` category, and it reports it as a failing test:

```
=========================== short test summary info ============================
FAILED test_temp.py::hdl_runner::test_fail - assert False
FAILED COCOTB RUNNER test_temp.py::hdl_runner - SystemExit: 1

Results (1.80s):
       1 passed
       2 failed
         - test_temp.py:20 test_fail
         - test_temp.py:7 hdl_runner
```

For reference, vanilla pytest considers `failed cocotb runner` as an unknown type, and it reports it like this:

```
============================================== short test summary info ==============================================
FAILED test_temp.py::hdl_runner::test_fail - assert False
================================ 1 failed, 1 passed, 1 failed cocotb runner in 0.53s ================================
```

It seems like `pytest-sugar` does not handle unknown types like `pytest` does:
https://github.com/pytest-dev/pytest/blob/0d9450d24b0f15b555a824ebc8cc221acff6c813/src/_pytest/terminal.py#L1392-L1399

For now, I will not concern myself with this, since it's a `pytest-sugar` issue.

tymonx:
> For now, I will not concern myself with this, since it's a pytest-sugar issue.

@Ehab-Ibrahim You are right! I have looked quickly in their code: [summary_stats](https://github.com/Teemu/pytest-sugar/blob/54ec9991e1e6e57a6fa4d70e5094797f6dc72f11/pytest_sugar.py#L539). They don't respect test status `category`. Everything will be put in the same bucket.

Thanks for the effort! LGTM!

ktbarrett:
Python < 3.10 doesn't support `X | Y` union syntax at runtime (annotations are not computed at runtime because of `from __future__ import annotations`). You'll have to use explicit `Union` calls.

src/cocotb_tools/pytest/plugin.py
```diff
@@ -60,6 +63,7 @@
         "cocotb_tools.pytest._init:run_regression",
     )
 )
+_TestStatus = TestShortLogReport | tuple[str, str, str | tuple[str, Mapping[str, bool]]]
```
ktbarrett:
```suggestion
if TYPE_CHECKING:
    from typing import TypeAlias

    TestStatus: TypeAlias = TestShortLogReport | tuple[str, str, str | tuple[str, Mapping[str, bool]]]
```

src/cocotb_tools/pytest/plugin.py
```diff
@@ -10,14 +10,16 @@
 import shlex
 import textwrap
 from argparse import ArgumentParser
-from collections.abc import Iterable
+from collections.abc import Generator, Iterable, Mapping
 from pathlib import Path
 from time import time
 from typing import Any
```
ktbarrett:
```suggestion
from typing import Any, TYPE_CHECKING
```

src/cocotb_tools/pytest/plugin.py
```diff
@@ -703,12 +707,44 @@ def _is_cocotb_test_report(item: Any) -> bool:
     return isinstance(item, TestReport) and getattr(item, "cocotb", False)
 
 
-@hookimpl(tryfirst=True)
+@hookimpl(hookwrapper=True, trylast=True)
+def pytest_report_teststatus(
+    report: CollectReport | TestReport, config: Config
+) -> Generator[None, Result[_TestStatus], None]:
+    result: Result[_TestStatus] = yield
+    status: _TestStatus = result.get_result()
```
ktbarrett:
```suggestion
) -> Generator[None, Result[TestStatus], None]:
    result: Result[TestStatus] = yield
    status: TestStatus = result.get_result()
```

Ehab-Ibrahim:
Thanks for the feedback. I have tested with `verilator` + `python 3.9.23`, and it seems to be working

ktbarrett (approved):

Ehab-Ibrahim:
I'm not sure if CI failures are caused by this PR

ktbarrett:
It is not. AWS decided to kill our instances randomly the other day. Someone suggested that someone realized that EC2 instances were cheaper than buying their own RAM and just spun up 100k of them to train their model.

ktbarrett:
Thanks @Ehab-Ibrahim for your first commit! And @tymonx for the review.

PR #5288, opened 2026-01-30
Include a VS Code Dev Container in the repository

imphil:
Remove the Gitpod setup and add a VS Code Dev Container configuration. The Dev Container configuration can be used both locally (with VS Code and theoretically also with devcontainer cli, untested) and in GitHub Codespaces.

There's more fine-tuning to be done, especially around running tests without nox; I'll revisit that in follow-ups once discussions around how we run our tests have settled down.

I'm also not fully happy about the first run experience in Codespaces, where it's easy to re-use the existing terminal which then doesn't have the venv loaded. For now I documented this behavior until I find a good solution that's not too finicky.

To test the Codespaces setup, go to https://github.com/codespaces/new?hide_repo_select=true&ref=devcontainer&repo=10651005&skip_quickstart=true&machine=basicLinux32gb&devcontainer_path=.devcontainer%2Fdevcontainer.json&geo=EuropeWest.


.devcontainer/README.md
```diff
@@ -0,0 +1,38 @@
...
+Dev Containers combine VS Code with Docker containers and configuration within the cocotb repository.
+
+By default, Dev Containers cannot access files on your local machine;
+if you want to use proprietary simulators installed on your machine, you might be better off with a development environment on your machine itself.
+
+## What's included in the Dev Container?
+
+* Ubuntu 24.04
+* Compilers: GCC and clang
+* Simulators: GHDL, Icarus Verilog, Verilator
```
tymonx: Any chance to add also the NVC simulator as well 🤔 I can help here in a separate PR if this too much to ask.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,40 @@
+# See https://github.com/devcontainers/images/tree/main/src/base-ubuntu/history
```
tymonx, line 5: [Containerfile](https://github.com/containers/common/blob/main/docs/Containerfile.5.md) is more agnostic and should be promoted nowadays instead of `Dockerfile`.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
+// VS Code Dev Container configuration file for a cocotb development environment.
+{
+	"name": "cocotb development",
+	"build": {
+		"context": "..",
+		"dockerfile": "Dockerfile"
```
tymonx, line 6: Preferable nowadays [Containerfile](https://github.com/containers/common/blob/main/docs/Containerfile.5.md):


```suggestion
		"dockerfile": "Containerfile"
```

.devcontainer/README.md
```diff
@@ -0,0 +1,38 @@
+# Cocotb Development Environment in a Dev Container
+
+The Dev Container provides a ready-to-code development environment for cocotb on Windows, Mac, or Linux.
+Dev Containers combine VS Code with Docker containers and configuration within the cocotb repository.
```
tymonx: It can be used with [Podman](https://podman.io/) as well. [Podman](https://podman.io/) is installed nowadays by default on many Linux distributions like Fedora or RHEL. It is more secure (rootless, no daemon) and lightweight.

```suggestion
Dev Containers combine VS Code with containers and configuration within the cocotb repository.
```

I would propose to remove the `Docker` word and use containers or container technology. `Docker` is not the only available container runtime.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+        libenchant-2-dev \
+        lldb \
+        pre-commit \
+        python3-dev \
+        python3-pip \
+        python3-venv \
+        valgrind \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Install Verilator from official Verilator Docker image
```
tymonx: `Docker` is not the only available container runtime or container registry:

```suggestion
# Install Verilator from official Verilator container image
```

.vscode/tasks.json
```diff
@@ -0,0 +1,15 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
+  // for the documentation about the tasks.json format
+  "version": "2.0.0",
+  "tasks": [
```
tymonx: Have you tried 🤔 

https://krijnvanrooijen.nl/blog/devcontainers-automate-workflow-tasks/#using-vscode-tasks-in-devcontainers

Other WoW is to run `devcontainer exec` and run command directly inside container.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+# Variable expansion in --from is not supported, so we have to repeat the
+# version number here.
+COPY --from=verilator/verilator:v5.044 /usr/local/share/verilator /usr/local/share/verilator
+COPY --from=verilator/verilator:v5.044 /usr/local/bin/verilator* /usr/local/bin/
+COPY --from=verilator/verilator:v5.044 /usr/local/share/man /usr/local/share/man
+COPY --from=verilator/verilator:v5.044 /usr/local/share/pkgconfig/verilator.pc /usr/local/share/pkgconfig
```
tymonx: If you put this at the beginning of the Containerfile before `FROM mcr.microsoft.com/devcontainers/base:2-ubuntu-24.04`:

```
ARG VERILATOR_VERSION=5.044
ARG UBUNTU_VERSION=24.04

FROM verilator/verilator:v${VERILATOR_VERSION} AS verilator

FROM mcr.microsoft.com/devcontainers/base:2-ubuntu-${UBUNTU_VERSION}
```
 
Then you can do:

```suggestion
COPY --from=verilator /usr/local/share/verilator /usr/local/share/verilator
COPY --from=verilator /usr/local/bin/verilator* /usr/local/bin/
COPY --from=verilator /usr/local/share/man /usr/local/share/man
COPY --from=verilator /usr/local/share/pkgconfig/verilator.pc /usr/local/share/pkgconfig
```

tymonx:
It would be also nice to document how to build container image and how to use it without need of VS Code/devcontainer setup.

# Podman

To build a container image:

```plaintext
podman build -t cocotb -f .devcontainer/Containerfile
```

Create and enter container:

```plaintext
podman run -it --rm --userns keep-id --volume "$(pwd):$(pwd):rw,z" --workdir "$(pwd)" cocotb bash
```

# Docker

Almost the same like `podman`. Replace `podman` with `docker` command. Remove unsupported `--userns keep-id` option. Add the `,z` suffix SELinux re-labeling to the volume option if working under SELinux.

```plaintext
docker run -it --rm --user "$(id -u):$(id -g)" --volume "$(pwd):$(pwd):rw,z" --workdir "$(pwd)" cocotb bash
```

Side note: This container image could be also used in CI via [GitHub Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#about-the-container-registry) (ghcr.io) to speed-up CI and published as `cocotb:dev` :thinking:

.devcontainer/Dockerfile
tymonx: Switch to non-root user if possible:

```suggestion
COPY --from=verilator/verilator:v5.044 /usr/local/share/pkgconfig/verilator.pc /usr/local/share/pkgconfig

USER vscode
```

By default, Docker is running as daemon under `root`. It is possible to perform vector attack and escape container chroot as `root` user. That's why Podman (rootless, daemon-less) is more secure compared to Docker.

* [Why Running Docker Containers as Root is a Security Nightmare (And How to Fix It)](https://medium.com/@ayoubseddiki132/why-running-docker-containers-as-root-is-a-security-nightmare-and-how-to-fix-it-83a60509d9bf)

ktbarrett:
@tymonx This is not the place to air your opinions, especially repeatedly as it's spammy. We do have interest in moving to uv and away from nox, but this change can still be brought in regardless of that.

.devcontainer/Dockerfile
imphil: The Dev Container tooling injects and uses a non-root user.

imphil: :+1:

.vscode/tasks.json
imphil: Seems to work. I'm a bit on the fence about going the "undocumented route" here, but maybe it's worth to keep the `.vscode/tasks.json` as (git) untracked config file for the user.

.devcontainer/README.md
imphil: As said in the comment: Docker is by now far more than a company, or a product. It's become a generic term that people recognize, while OCI or "containers" are either obscure or too generic. I kept "Docker" in parentheses to at least trigger the right braincells in readers.

.devcontainer/Dockerfile
imphil: I disagree, as explained further below.

.devcontainer/devcontainer.json
imphil: As above: I disagree.

.devcontainer/README.md
imphil: No problem, I added NVC as well.

imphil:
Thanks for your feedback, @tymonx! 

I incorporated some of your suggestions where they were in the scope of this PR.

Beyond that:
- On uv: we don't recommend using uv yet in our docs. This PR isn't the place to change that. (There are ongoing discussions about uv, bit I think there are still some things that don't work properly for cocotb? I haven't followed closely.)
- Containerfile vs Dockerfile: I don't see the point of the rename (never did). All tools support the Dockerfile name, it's more way more common in the wild and widely recognized by developers. There's no reason to purge the name "Docker" from the world.
- On podman and instructions for the command-line: The purpose of the Dev Container (and this PR) is to give users *one* reliable, opinionated way of getting a working environment. If you are an experienced user who wants to do something special we have already install instructions for that.

tymonx:
> On podman and instructions for the command-line: The purpose of the Dev Container (and this PR) is to give users one reliable, opinionated way of getting a working environment.

There are plenty users that don't use VS Code as IDE or they don't want to install heavy `devcontainers` that is based on TypeScript. It will not harm to mention a few words about using `docker`/`podman` commands with `.devcontainer/Dockerfile` file to setup reliable and reproducible development environment.

> If you are an experienced user who wants to do something special we have already install instructions for that.

This is not about having something special but to have a reliable and reproducible development environment without frictions from current developer OS environment. The `.devcontainer/Dockerfile` is very good for that.

I will give an example from the Red Hat OpenShift art-tools project:

* https://github.com/openshift-eng/art-tools/tree/main/.devcontainer

.devcontainer/Dockerfile
tymonx: > The Dev Container tooling injects and uses a non-root user.

I'm not sure about that after reading this [Add a non-root user to a container](https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user) and [Files created inside container are owned by root:root instead of my user](https://github.com/microsoft/vscode-remote-release/issues/49). I'm not also a VS Code user to confirm that in 100%. But definitely this will alert security teams or XDR security scanners.

I checked that `Dockerfile`. They defined the `vscode` user but default user is set to `root`.

.devcontainer/devcontainer.json
tymonx: 👍 I will resolve this because it is not solving any technical problem here.

.devcontainer/Dockerfile
tymonx: 👍 I will resolve this because it is not solving any technical problem here.

.devcontainer/devcontainer.json
tymonx: GitHub doesn't allow me to resolve this thread 😞 It is screaming with the `Failed to resolve thread` error.

.devcontainer/Dockerfile
tymonx: GitHub doesn't allow me to resolve this thread 😞 It is screaming with the `Failed to resolve thread` error.

tymonx:
I will expend my comment. I have own `Containerfile` similar to the `.devcontainer/Dockerfile` file presented here in this PR to quickly test my changes in cocotb with multiple simulators and different Python versions (thanks to the `uv` command with the `--managed-option` option).

And the proposed `.devcontainer/Dockerfile` file in this PR could be used as standardized development environment for cocotb contributors. This will give:
- less spam to CI to test your changes against different simulators (at least Open Source)
- be on the same page with everyone (and maybe CI) if something is not working (_"it is working on my machine"_, _"this bug/issue never happen to me"_)

But not everyone are in favor of using VS Code IDE or install `devcontainers-cli` from `npm` that is only wrapping `docker`/`podman` commands.

In my opinion, it would be nice to promote `.devcontainer/Dockerfile` as consistent and reliable development environment. At least to mention that `.devcontainer/Dockerfile` can be used with `docker` or `podman`. I have tested that with vanilia `podman` and it works great.

I can also help here and document how to use/work with `.devcontainer/Dockerfile` file with `docker`/`podman` commands in separate PR.

.devcontainer/README.md
tymonx: GitHub doesn't allow me to resolve this thread 😞 It is screaming with the Failed to resolve thread error. For me, it is resolved.

.devcontainer/Dockerfile
tymonx: https://www.happihacking.com/blog/posts/2024/dev-containers-uids/

> In Microsoft's devcontainer images, the USER setting is actually root, but the images have additional devcontainer.metadata which can provide the same settings as devcontainer.json, and this contains an entry "remoteUser":"vscode", so that when started from VSCode or with devcontainer up, the user will be vscode. When started with a plain Docker command, the metadata is not parsed.

This is answering my question. I built that container image and there is an associated label:

```plaintext
podman image inspect cocotb | jq --raw-output '.[].Labels.["devcontainer.metadata"]' | jq
```

Output:

```json
[
  {
    "id": "ghcr.io/devcontainers/features/common-utils:2"
  },
  {
    "id": "ghcr.io/devcontainers/features/git:1",
    "customizations": {
      "vscode": {
        "settings": {
          "github.copilot.chat.codeGeneration.instructions": [
            {
              "text": "This dev container includes an up-to-date version of Git, built from source as needed, pre-installed and available on the `PATH`."
            }
          ]
        }
      }
    }
  },
  {
    "remoteUser": "vscode"
  }
]
```

devcontainer can use labels to inject properties. Someone could argue that this approach could create some security breach. But you are using the official container image.

For me, this is resolved as well.

tymonx:
From my side all threads are resolved :+1: If people are interested, I can document Way-of-Working with the `.devcontainer/Dockerfile` and vanilia `docker`/`podman` commands in separate PR.

tymonx:
For my own curiosity I tried `devcontainer-cli`:

* Permission issues when installing `devcontainer` globally. I needed to setup a local prefix to my home directory: `npm config set prefix ~/.local`. Then install it as user: `npm install -g @devcontainers/cli`
* `devcontainer` runs everything by using the `docker` command. Symbolic link or alias is needed to Podman: `ln -s /usr/bin/podman ~/.local/bin/docker`
* No issues when building container image: `devcontainer build`
* Permission issues when accessing `.devcontainer/post-create.sh` script file when invoking the `devcontainer up` command. This is because `devcontainer` doesn't handle environment with enabled SELinux (RHEL, Fedora, Centos, ...)
* It requires to add the `,z` option when mounting workspace [Dev container and SELinux](https://discussion.fedoraproject.org/t/dev-container-and-selinux/172257)
* In case of Podman, additional `--userns=keep-id` option `"runArgs": ["--userns=keep-id"]`
* Then the `devcontainer up` command will success:

```plaintext
{"outcome":"success","containerId":"591d8b52f820bae17b3e42bdd761c31b824cb78f9f7ad99acb29204ca0b0a2c6","remoteUser":"vscode","remoteWorkspaceFolder":"/workspaces/cocotb"}
```

* From there, user can execute `devcontainer exec <command> <arguments>...` to execute any command inside created and started container. But it is painful slow... 🐌 (I think because of the TypeScript part when `devcontainer` is trying to execute `podman exec -it <container-name> <command> <arguments>...` under the hood)
* Or enter container via `devcontainer exec bash` for better experience

At the end, it worked. The same issues with permissions denied (SELinux) will be with the VS Code as well. Missing the `,z` volume option and the `--userns=keep-id` run option (Podman only).

tymonx:
There is an issue for running devcontainer under SELinux https://github.com/microsoft/vscode-remote-release/issues/11284#issuecomment-3842807119 I have posted a possible workaround but this should be handled by the devcontainer to make it fully transparent and seamless experience. Otherwise it will require manual tweaks in the `.devcontainer/devcontainer.json` file.

Workaround could be added in `.devcontainer/devcontainer.json` file. The `relabel=shared|private` or `z|Z` option should be ignored on platform without SELinux. This is true for Linux. I cannot confirm that on Windows/macOS. But they are using Virtual Machine with Linux. In theory, it should be transparent.

ktbarrett:
Should we mention the Github Codespace in the main README? Also remove the Gitpod banner from the README.

imphil:
> Should we mention the Github Codespace in the main README? Also remove the Gitpod banner from the README.

I removed the banner from the README, but didn't add the Codespaces one yet: I'd like to see this PR merged to master and play with it for a while to remove remaining rough edges, before we advertise it more widely.

ktbarrett (approved):

imphil:
Thanks for your additional thoughts, @tymonx! For this PR, I'd like to keep things simple and focused on a single use case. If we can extend on that without loosing the simplicity we're trying to provide here we can look into a follow-up.

PR #5289, opened 2026-01-31
Refactor TestManager (part 2)

ktbarrett:
Numerous small improvements, but primarily this is intended to ensure that all tasks are finished before calling the `test_complete_cb`. Best reviewed commit-by-commit. Will squash when merging.

Depends on #5296. Depends on #5313.

### DIFF
--- a/src/cocotb/_test_manager.py
+++ b/src/cocotb/_test_manager.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import pdb
 import sys
+from asyncio import CancelledError
 from collections.abc import Coroutine
 from typing import (
     Any,
@@ -65,17 +66,17 @@ class TestManager:
         # If cancelled, end the Test without additional error. This case would only
         # occur if a child threw a CancelledError or if the Test was forced to shutdown.
         if task.cancelled():
-            self.abort()
+            self._abort()
             return
         # Handle outcome appropriately and shut down the Test.
         e = task.exception()
         if e is None:
-            self.abort()
+            self._abort()
         elif isinstance(e, TestSuccess):
             task._log.info("Test stopped early by this task")
-            self.abort()
+            self._abort()
         else:
-            self.abort(e)
+            self._abort(e)
 
     def start(self) -> None:
         # set global current test manager
@@ -95,7 +96,7 @@ class TestManager:
 
     def _on_timeout(self) -> None:
         self._timeout_cb = None
-        self.abort(SimTimeoutError())
+        self._abort(SimTimeoutError())
 
     def exception(self) -> BaseException | None:
         if self._excs:
@@ -114,8 +115,16 @@ class TestManager:
         """Return whether the test has completed."""
         return self._complete
 
-    def abort(self, exc: BaseException | None = None) -> None:
-        """Force this test to end early."""
+    def cancel(self, msg: str | None = None) -> None:
+        """End the test prematurely."""
+        if msg is not None:
+            exc = CancelledError(msg)
+        else:
+            exc = CancelledError()
+        self._abort(exc)
+
+    def _abort(self, exc: BaseException | None = None) -> None:
+        """Shutdown the test"""
 
         if exc is not None:
             self._excs.append(exc)
@@ -180,9 +189,9 @@ class TestManager:
         # there was a failure and no one is watching, fail test
         elif isinstance(e, TestSuccess):
             task._log.info("Test stopped early by this task")
-            self.abort()
+            self._abort()
         else:
-            self.abort(e)
+            self._abort(e)
 
 
 def start_soon(
--- a/src/cocotb/_test_manager.py
+++ b/src/cocotb/_test_manager.py
@@ -38,16 +38,20 @@ class TestManager:
     #  done callback with.
 
     def __init__(
-        self, test_complete_cb: Callable[[], None], main_task: Task[None]
+        self,
+        test_coro: Coroutine[Trigger, None, Any],
+        *,
+        name: str,
+        test_complete_cb: Callable[[], None],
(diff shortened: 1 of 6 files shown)


src/cocotb_tools/pytest/_fixture.py
```diff
@@ -51,6 +47,11 @@ def __getitem__(self, index: Any) -> Any:
...
 
+def mark_async_fixture(arg: Any) -> None:
+    """Mark argument as asynchronous fixture result."""
+    setattr(arg, "_cocotb_async_fixture", True)
+
+
 def resolve_fixture_arg(arg: Any) -> Any:
     """Resolve fixture argument."""
-    return arg.result() if isinstance(arg, AsyncFixture) else arg
+    return arg.result() if getattr(arg, "_cocotb_async_fixture", False) else arg
```
tymonx: @ktbarrett My bad that there are no tests for that. I have created some tests locally for asynchronous fixtures. `arg` is now a `TestManager` class with [result()](https://github.com/cocotb/cocotb/blob/2156c79de19f6dd110934730fbaabecc3d5d7ba6/src/cocotb/_test_manager.py#L82) method that is always returning `None` result. Previously it was a `Task` where `result()` returned an outcome value.

Something like this will now fail:

```python
import pytest

@pytest.fixture(name="x")
async def x_fixture() -> int:
    return 5


async def test_x(dut, x: int) -> None:
    assert x == 5
```

`x` will be `None` because [arg.result()](https://github.com/cocotb/cocotb/blob/2156c79de19f6dd110934730fbaabecc3d5d7ba6/src/cocotb/_test_manager.py#L82) is also returning `None`. I assume that it should be taken from the `_main_task` 🤔

I will prepare a Pull Request and add some tests for asynchronous fixtures.

With the `TestManager` class, I think we can simplify `if` to:

```python
return arg.result() if isinstance(arg, TestManager) else arg
```

And remove `mark_async_fixture`.

tymonx: @ktbarrett Nothing fancy https://github.com/cocotb/cocotb/pull/5296 but it should catch that situation for asynchronous fixtures.

src/cocotb_tools/pytest/_fixture.py
tymonx: Now it should be a `TestManager` type.


```suggestion
        task: TestManager = super().__getitem__(0)
```

src/cocotb_tools/pytest/_regression.py
```diff
@@ -557,16 +555,24 @@ def pytest_fixture_setup(
...
             func = self._setup_async_function(fixturedef, request)
 
-        task: Task = AsyncFixture(func(), name=f"Setup {request.fixturename}")
-        cache_key = fixturedef.cache_key(request)
-        fixturedef.cached_result = AsyncFixtureCachedResult((task, cache_key, None))
-
         if is_async_generator:
             # Test setup with teardown, added sub-tasks during setup will be cancelled by test teardown
-            setup = RunningTestSetup(self._setup_completed, task)
+            setup = RunningTestSetup(
```
tymonx, line 559: There is a chance to have a boolean option in the `TestManager` class to not cancel/remove all added sub-tasks when the main task will finish 🤔

Then we could simplify that part of code to single:

```python
            setup = TestManager(
                func(),
                name=f"Setup {request.fixturename}",
                test_complete_cb=self._setup_completed,
                keep_subtasks=is_async_generator,
            )
```

And remove the `RunningTestSetup` class.

src/cocotb_tools/pytest/_fixture.py
ktbarrett: Alright I can change this. But be aware that `TestManager` is not really appropriate for use as a setup and teardown runner. The tasks stolen from the setup object still bind to that object. So if a Task fails, the object it will report to is already dead.

We need a more holistic solution rather than hackily reusing stuff that isn't meant for it. That's not necessarily your problem to solve, part of the reason for refactoring TestManager is so I can implement setup and teardown for the current RegressionManager/TestManager. After that's implemented we can reuse it in the pytest plugin. Depending upon what I come up with, it might be a bit of a bear to map the current AsyncGenerator stuff to it, so it might necessitate an API change.

I think I mentioned that the repo wasn't really in a good state for pytest support yet. But it's in now, so we will just have to push the 2.1 release out farther until we work out all the kinks.

src/cocotb/_test_manager.py
```diff
@@ -101,19 +131,42 @@ def abort(self, exc: BaseException | None = None) -> None:
...
-        for task in self.tasks[:]:
+        for task in self._tasks[:]:
             task._cancel_now()
 
-        # TODO: Wait for Tasks to end
+        # Register function to clean up global state once all Tasks have ended.
+        if not self._done.is_set():
+            self._on_complete()
+        else:
+            self._done.wait()._register(self._on_complete)
```
marlonjames: This looks backwards to me.

src/cocotb/_test_manager.py
```diff
@@ -84,6 +110,10 @@ def result(self) -> None:
         if exc:
             raise exc
 
+    def done(self) -> bool:
+        """Return whether the test has completed."""
+        return self._complete
+
     def abort(self, exc: BaseException | None = None) -> None:
```
marlonjames: While you're refactoring, in this or another PR, this should probably be renamed. It's used for normal test ending in `_test_done_callback` as well as for actually ending the test early.

marlonjames (approved):

ktbarrett:
@marlonjames I added a commit to rename `abort()` to `_abort()` and make a public `cancel()` method which mimics the one on `Task`.

marlonjames (approved):

PR #5293, opened 2026-02-01
LogicArray support negative ints

ktbarrett:
This adds support for construction and equality with negative integers.

- [x] newsfrag

### DIFF
--- a/tests/pytest/test_logic_array.py
+++ b/tests/pytest/test_logic_array.py
@@ -68,6 +68,22 @@ def test_logic_array_int_construction():
         LogicArray(-10, Range(3, "downto", 0))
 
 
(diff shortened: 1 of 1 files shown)


tests/pytest/test_logic_array.py
```diff
@@ -284,6 +305,8 @@ def test_equality():
     assert LogicArray("0101") != "lol"
     assert LogicArray("0101") != 123
     assert LogicArray("0101") != [7, "f", dict]
+    assert LogicArray("1111") == -1
```
marlonjames, line 308:
```suggestion
    assert LogicArray("1111") == -1
    assert LogicArray("1111") == 15
```

PR #5300, opened 2026-02-02
fix(test_task_manager): RuntimeError: ignored GeneratorExit exception

### LINKED ISSUE #5297: Resolve odd `GeneratorExit` issues in `test_cocotb`
This is causing things to fail randomly: https://github.com/cocotb/cocotb/actions/runs/21559043298/job/62120114701?pr=5295#step:28:2787

tymonx:
Fixes #5297 (probably). More technical details about potential fix in my comment: https://github.com/cocotb/cocotb/issues/5297#issuecomment-3833890963

Tested aggressively locally:

```plaintext
for i in $(seq 1 20); do SIM=ghdl TOPLEVEL_LANG=vhdl make -C ./tests/test_cases/test_cocotb; done
```

### DIFF
--- a/tests/test_cases/test_cocotb/test_task_manager.py
+++ b/tests/test_cases/test_cocotb/test_task_manager.py
@@ -855,8 +855,10 @@ async def test_start_soon_after_cancel_no_continue(_: object) -> None:
             finally:
                 c = coro(2)
                 with pytest.raises(RuntimeError):
-                    tm.start_soon(c)
-                c.close()  # avoid ResourceWarning since we didn't await it.
+                    try:
+                        tm.start_soon(c)
+                    finally:
+                        c.close()
 
     except BaseExceptionGroup as e:
         my_exc, rest = e.split(MyException)
@@ -903,8 +905,10 @@ async def test_start_soon_outside_context(_: object, continue_on_error: bool) ->
 
     c = coro(1)
     with pytest.raises(RuntimeError):
-        tm.start_soon(c)
-    c.close()  # avoid ResourceWarning since we didn't await it.
+        try:
+            tm.start_soon(c)
+        finally:
+            c.close()
 
 
 @cocotb.test
@@ -934,8 +938,10 @@ async def test_add_tasks_from_another_task(_: object, continue_on_error: bool) -
 
     c = coro(1)
     with pytest.raises(RuntimeError):
-        tm.start_soon(c)
-    c.close()  # avoid ResourceWarning since we didn't await it.
+        try:
+            tm.start_soon(c)
+        finally:
+            c.close()
 
     ev.set()
 
@@ -951,8 +957,10 @@ async def test_start_soon_after_finished(_: object, continue_on_error: bool) -> 
 
     c = coro(1)
     with pytest.raises(RuntimeError):
-        tm.start_soon(c)
-    c.close()  # avoid ResourceWarning since we didn't await it.
+        try:
+            tm.start_soon(c)
+        finally:
+            c.close()
 
 
 @cocotb.test


tymonx:
Not my (GitHub) day :shrug: 

> The job was not acquired by Runner of type hosted even after multiple attempts
Internal server error. Correlation ID: bff791db-b24f-45d9-9ba5-0fb7bcc733f5

https://www.githubstatus.com/

> Update - The team continues to investigate issues causing GitHub Actions jobs on hosted runners to remain queued for extended periods, with a percentage of jobs failing. We will continue to provide updates as we make progress toward mitigation.

image

ktbarrett:
I'd be interested to see if this works. They seem equivalent to me. `c` must not be collected until after the end of the `pytest.raises` block. It never being closed would only occur if the test failed. Perhaps there's some weird lifetime stuff in there since the coroutine object gets ref'd by the traceback object?

tymonx:
> Perhaps there's some weird lifetime stuff in there since the coroutine object gets ref'd by the traceback object?

Could be related :thinking: [pytest.raises swallows warning about unawaited coroutine](https://github.com/pytest-dev/pytest/issues/10404)

ktbarrett:
So the `RuntimeWarning: coroutine 'coro' was never awaited` at the end of the test is caused by `test_task_manager.test_failure_in_block/continue_on_error=False`. 

The issue is that we cancel the `_waiter` Task when we get the exception in `__aexit__`, but that Task was never started, and the coroutine passed in to the `_waiter` is never closed because `_waiter` is never started.

I think we need to check that the object is a coroutine before wrapping it in `_waiter`. We don't need to wrap it if it's already a coroutine. And if we don't wrap it we can ensure that the cancellation will actually be thrown into the passed in coroutine.

ktbarrett:
There are still these issues. These are tests which ignore the cancellation so this is kind of expected, but obviously is causing issues.

```
Exception ignored in: <coroutine object test_external_cancel_in_block_ignored_and_await.<locals>.run_task_manager at 0x7f4e37f175a0>
  File "/usr/scratch/ktbarrett/checkouts/cocotb/tests/test_cases/test_cocotb/test_task_manager.py", line 850, in test_start_soon_after_cancel_no_continue
    async with TaskManager(default_continue_on_error=False) as tm:
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: coroutine ignored GeneratorExit
Exception ignored in: <coroutine object test_external_cancel_in_block_ignored_and_await.<locals>.run_task_manager at 0x7f4e378d0740>
  File "/usr/scratch/ktbarrett/checkouts/cocotb/tests/test_cases/test_cocotb/test_task_manager.py", line 850, in test_start_soon_after_cancel_no_continue
    async with TaskManager(default_continue_on_error=False) as tm:
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: coroutine ignored GeneratorExit
Exception ignored in: <coroutine object test_child_fails_ignore_cancel_and_await at 0x7f4e37961300>
  File "/usr/scratch/ktbarrett/checkouts/cocotb/tests/test_cases/test_cocotb/test_task_manager.py", line 850, in test_start_soon_after_cancel_no_continue
    async with TaskManager(default_continue_on_error=False) as tm:
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: coroutine ignored GeneratorExit
```

imphil:
Not ready, dropping 2.1 milestone.

PR #5309, opened 2026-02-04
Add flag to end regression after N failures

### LINKED ISSUE #5305: End regression at first test failure mode
We can support ending the regression after the first test fails. We should have a new exception called `RegressionTerminated`.

ZiaCheemaGit:
Fixes #5305

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -238,6 +238,8 @@ Discovering Tests
 
     If not set, all tests are executed regardless of failures.
 
+    .. versionadded:: 2.1
+
 .. envvar:: COCOTB_TEST_MODULES
 
     Type: :ref:`env-list`
--- a/src/cocotb_tools/config.py
+++ b/src/cocotb_tools/config.py
@@ -78,6 +78,7 @@ def _help_vars_text() -> str:
         COCOTB_RESULTS_FILE       File name for xUnit XML tests results
         COCOTB_USER_COVERAGE      Collect Python user coverage (HDL for some simulators)
         COVERAGE_RCFILE           Configuration for user code coverage
+        COCOTB_MAX_FAILURES       Maximum number of test failures before aborting the regression
 
         GPI
         ---


src/cocotb/regression.py
```diff
@@ -334,48 +340,56 @@ def _execute(self) -> None:
...
+                    self._record_test_excluded()
+                    continue
+
+                # if the test is skipped, record and continue
+                if self._test.skip and self._mode != RegressionMode.TESTCASE:
+                    self._record_test_skipped()
+                    continue
+
+                # if the test should be run, but the simulator has failed, record and continue
+                if self._sim_failure is not None:
```
ktbarrett: I think instead of throwing the exception, we could just set `self._sim_failure` to the new exception. That would require minimal code changes.

src/cocotb/regression.py
```diff
@@ -80,6 +80,10 @@ class SimFailure(BaseException):
     """
 
 
+class RegressionTerminated(BaseException):
+    """Raised to stop the regression after the first test failure."""
```
ktbarrett: Like with SimFailure we should mention this isnt intended to be raised by the user.

src/cocotb/regression.py
```diff
@@ -172,6 +176,8 @@ def __init__(self) -> None:
         self._sim_failure: SimFailure | None = None
         self._regression_seed = cocotb.RANDOM_SEED
         self._random_state: Any
+        self._stop_on_failure = _env.as_bool("COCOTB_STOP_ON_FAILURE", default=False)
```
ktbarrett: Pytest has an option call `--maxfailures` that exits after that many test failures are seen. We could easily support that here making this `COCOTB_MAX_FAILURES` as an int and incrementing on every test failure, then finally setting the `_sim_failure` field.

ZiaCheemaGit: Done

src/cocotb/regression.py
ZiaCheemaGit: Done

src/cocotb/regression.py
ZiaCheemaGit: Done

docs/source/newsfragments/5309.feature.rst
```diff
@@ -0,0 +1 @@
+Added support for terminating a regression early when the number of test failures reaches ``COCOTB_MAX_FAILURES``.
```
cmarqu:
```suggestion
Added support for terminating a regression early when the number of test failures reaches :envvar:`COCOTB_MAX_FAILURES`.
```

docs/source/library_reference.rst
```diff
@@ -227,6 +227,18 @@ Marking and Generating Tests
...
+
+    Type: :ref:`_env-boolean`
+
+    Limits the number of test failures allowed during a regression run.
+
+    If set then if the number of failed tests reaches this value, the regression
+    is terminated early and no further tests are executed.
+
+    If not set this disables early termination and causes
+    all tests to be executed regardless of failures.
```
cmarqu:
```suggestion
    If not set, all tests are executed regardless of failures.
```
No information loss with the shorter sentence, right?

src/cocotb/regression.py
```diff
@@ -80,6 +80,17 @@ class SimFailure(BaseException):
     """
 
 
+class RegressionTerminated(BaseException):
+    """Indicates the regression was terminated early.
+    The regression can be terminated early by setting ``COCOTB_MAX_FAILURES``.
```
```suggestion
    """Indicates the regression was terminated early.
    
    The regression can be terminated early by setting :envvar:`COCOTB_MAX_FAILURES`.
```

src/cocotb/regression.py
```diff
@@ -80,6 +80,17 @@ class SimFailure(BaseException):
...
+
+
+    .. caution::
+        Not intended to be raised or caught by user code.
+        Used internally by the RegressionManager.
```
```suggestion
        Used internally by the :class:`RegressionManager`.
```

docs/source/library_reference.rst
```diff
@@ -227,6 +227,17 @@ Marking and Generating Tests
 Discovering Tests
 =================
 
+.. envvar:: COCOTB_MAX_FAILURES
+
+    Type: :ref:`_env-boolean`
```
ktbarrett: Maybe it's `_env-integer`? Might need to check,
```suggestion
    Type: :ref:`_env-int`
```

ZiaCheemaGit: nice catch.

docs/source/library_reference.rst
```diff
@@ -227,6 +227,17 @@ Marking and Generating Tests
...
+
+    Type: :ref:`_env-int`
+
+    Limits the number of test failures allowed during a regression run.
+
+    If set then if the number of failed tests reaches this value, the regression
+    is terminated early and no further tests are executed.
+
+    If not set, all tests are executed regardless of failures.
+
```
marlonjames, line 240:
```suggestion

    .. versionadded:: 2.1

```

PR #5318, opened 2026-02-10
PYGPI: Simulator module cleanup

marlonjames:
Some cleanup of the simulator module.
* Change the class names to not exactly match the GPI handle types but reflect their purpose and usage.
* Align the C++ clock type name with the typing stubs.
* Cleanup the typing stubs.


ktbarrett (approved):
Looks fine to me.

PR #5320, opened 2026-02-12
Dev Container: Install uv

imphil:


.devcontainer/Dockerfile
```diff
@@ -1,6 +1,9 @@
-# Verilator image, only used to copy some binaries out of it (see below).
 ARG VERILATOR_VERSION=5.044
+ARG UV_VERSION=0.10.2
```
imphil, line 2: Dockerfile syntax requires those ARG lines to be before any FROM line, hence the slightly unusual ordering.

ktbarrett: Do we have some way to auto-bump these? Or is that not a concern?

imphil: It't not critical, but I'd love to have automation for that. However, I don't think we can get Dependabot to do that -- maybe something else? Let's see how frequent we need to act.

PR #5314, opened 2026-02-08
Replace nox with invoke

ktbarrett:
Depends on #5327.

The first commit changes the noxfile to use dependency groups for managing dependencies in tests. This is required for use to eventually use `uv` for managing these dependencies.

The second commit replaces nox with invoke. Most of these changes were done by Claude with some touching up by me.

PR #5328, opened 2026-02-16
Use `nox-uv`

ktbarrett:
~~Depends on #5327.~~ ~~Depends on #5336.~~

Adds `nox-uv` as a development dependency and convert the noxfile to use it. This allows us to use `uv sync` to install dependency groups.

As opposed to #5314 this keeps nox which means...
* Will use `nox -s docs` and `nox -k "dev_test and icarus and verilog"` etc instead of `uv run invoke docs`
* Each session still maintains it's own venv for reasons...
* But builds are cached so no more rebuilding cocotb unless the sources change
* No major rewrite

### DIFF
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -41,8 +41,10 @@ jobs:
       - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
         with:
           python-version: "3.12"
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
       - name: Install nox
-        run: python3 -m pip install nox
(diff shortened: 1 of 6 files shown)


marlonjames:
Need to install `nox-uv` in the **Install Python testing dependencies** CI step?

ktbarrett:
Right. Eventually we will just use `uv run nox ...`

ktbarrett:
Ugh... okay we are still building cocotb multiple times because `uv run nox ...` is building it and then building it again in the nox venv. There is no `--no-install-package` for `uv run`.

So I guess we install `nox` and `nox-uv` in the CI. For local development `uv run ...` is still fine if you want to do that since it will only rebuild cocotb when you issue `uv run ...` if you change the sources.

PR #5345, opened 2026-02-20
docs: Update GPI docs

marlonjames:
Re-organize `gpi.h` to move enums into their associated doxygen groups.
Update the docs page to categorize by the groups. I think it's better even though the doxygen group name is repeated below the rst section header.

Also added a handler to remove the intersphinx link to the old 1.4 documentation for `GpiObjHdl`, `GpiIterator`, and `GpiCbHdl`.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -436,3 +436,19 @@ extlinks = {
         None,
     ),
 }
+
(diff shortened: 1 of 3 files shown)


marlonjames:
Yeah right now GPI only has child sim object iterators from parent sim objects (or null for iterating packages).

PR #5350, opened 2026-02-20
Prototype: Fast-loop API for tight signal read/write loops

robtaylor:
## Summary

**This is a prototype/RFC for discussion, not a production-ready PR.**

Introduces `cocotb.fast` — a new API module that lets users write tight signal read/write loops that bypass Python's coroutine suspend/resume overhead entirely. Profiling shows ~29% of loop time in standard cocotb is spent in scheduler/coroutine machinery (`_react`, `_resume`, `send`, `_do_callbacks`, `EventLoop.run`). This API eliminates that overhead.

### Two approaches

**1. Callback API** (`fast.run_cycles`) — maximum performance:
```python
from cocotb.fast import SignalProxy, run_cycles

data = SignalProxy(dut.data_in)
out = SignalProxy(dut.data_out)

def step(cycle):
    data.set_int(cycle & 0xFF)
    val = out.get_int()
    return cycle < 1_000_000

await run_cycles(dut.clk, step)
```

**2. Async/await API** (`fast.run`) — familiar syntax, ~80% of callback speedup:
```python
from cocotb import fast

async def inner():
    data = fast.SignalProxy(dut.data_in)
    out = fast.SignalProxy(dut.data_out)
    edge = fast.RisingEdge(dut.clk)
    for i in range(1_000_000):
        data.set_int(i & 0xFF)
        await edge
        val = out.get_int()

await fast.run(inner())
```

### Key components

| Component | Purpose |
|---|---|
| `src/cocotb/fast.py` | Public API — `run()`, `run_cycles()`, `SignalProxy`, triggers |
| `src/cocotb/_fast_loop.py` | `SignalProxy`, `_FastLoopDone` trigger, `run_cycles()` |
| `src/cocotb/_fast_sched.pyx` | Cython mini-scheduler with zero-allocation `__await__` triggers |
| `src/cocotb/_fast_sched_py.py` | Pure Python fallback (no Cython required) |
| `tools/convert_to_fast.py` | AST-based converter to help migrate existing tests |
| `tests/pytest/test_fast.py` | 17 unit tests (trigger protocol, converter, imports) |
| `tests/pytest/test_fast_benchmark.py` | Simulator benchmark tests |
| `.github/workflows/fast-api.yml` | CI: unit tests, fallback tests, converter, benchmark, lint |

### How it works

- **`SignalProxy`**: Wraps a GPI handle directly — `get_int()` calls `get_signal_val_long()`, `set_int()` calls `set_signal_val_int()`. No `Logic`/`LogicArray`/`Range` construction.
- **Callback loop**: GPI value-change callback calls user's `step_fn` directly. No `Task._resume()`, no `coro.send()`, no event loop iteration per cycle.
- **Mini-scheduler**: `_FastScheduler` drives a coroutine via `coro.send(None)` → isinstance dispatch on yielded trigger → register appropriate GPI callback. Zero-allocation `__await__` protocol (triggers return `self` as iterator).
- **Return to cocotb**: `_FastLoopDone(GPITrigger)` fires via `register_readonly_callback(self._react)` to cleanly resume the awaiting cocotb Task.

### Converter tool

```bash
python tools/convert_to_fast.py my_test.py --diff    # Preview changes
python tools/convert_to_fast.py my_test.py -o fast_test.py  # Convert
```

Handles: signal reads/writes → `SignalProxy.get_int()/set_int()`, `RisingEdge`/`FallingEdge`/`ReadOnly`/`ReadWrite` → `fast.*`, wraps hot loops in `async def _fast_inner()` + `await fast.run()`. Warns about unsupported patterns (Timer, Combine, etc.).

### Questions for discussion

1. Is a separate `cocotb.fast` namespace the right approach, or should this integrate differently?
2. Should `SignalProxy` support more value types beyond int/binstr?
3. Is the Cython dependency acceptable? (Pure Python fallback is provided)
4. How should this interact with cocotb's logging/reporting infrastructure?

## Test plan

- [ ] `pytest tests/pytest/test_fast.py -v` — 17 unit tests pass (no simulator needed)
- [ ] `SIM=icarus pytest tests/pytest/test_fast_benchmark.py -v -s` — simulator benchmarks
- [ ] `python tools/convert_to_fast.py profile_bench/bench_signals.py --diff` — converter smoke test
- [ ] Run existing cocotb test suite — fast API is purely additive, no existing behavior changed

Co-developed-by: Claude Code v2.1.49 (claude-opus-4-6)

### DIFF
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -228,9 +228,7 @@ class build_ext(_build_ext):
         # With package_dir={"": "src"}, --inplace creates an extra nesting level
         # (e.g. src/cocotb/cocotb/_fast.so instead of src/cocotb/_fast.so).
         if self.inplace:
-            src_dir = os.path.join(
-                os.path.dirname(__file__), "src", "cocotb"
-            )
+            src_dir = os.path.join(os.path.dirname(__file__), "src", "cocotb")
             misplaced_dir = os.path.join(src_dir, "cocotb")
             if os.path.isdir(misplaced_dir):
                 for fname in os.listdir(misplaced_dir):
--- a/profile_bench/run_profile.py
+++ b/profile_bench/run_profile.py
@@ -1,4 +1,5 @@
 """Runner script for the profiling benchmark."""
+
 from __future__ import annotations
 
 import os
@@ -7,6 +8,7 @@ from pathlib import Path
 from cocotb_tools.runner import get_runner
 
 proj_path = Path(__file__).resolve().parent
+
 
 def run():
     sim = os.getenv("SIM", "verilator")
@@ -22,6 +24,7 @@ def run():
         test_module="bench_signals",
         test_dir=str(proj_path),
     )
+
 
 if __name__ == "__main__":
     run()
--- a/src/cocotb/fast.py
+++ b/src/cocotb/fast.py
@@ -65,26 +65,6 @@ the following behavioral differences that users should be aware of:
     ``COCOTB_TRUST_INERTIAL_WRITES=1`` is set, standard cocotb also
     writes immediately, so behavior is identical on trusted simulators.
 
-**``current_gpi_trigger()`` is updated** *(resolved)*
-    The fast scheduler sets
-    :func:`cocotb._gpi_triggers.current_gpi_trigger` to the standard
-    singleton ``ReadOnly()`` / ``ReadWrite()`` instances when entering
-    those phases.  Concurrent standard cocotb tasks see correct phase
-    awareness for write guards.
-
-**``_apply_scheduled_writes()`` called on ReadWrite** *(resolved)*
-    The fast scheduler calls ``cocotb.handle._apply_scheduled_writes()``
-    when entering the ReadWrite phase, matching standard
-    ``ReadWrite._do_callbacks()`` behavior.  Deferred writes queued by
-    ``dut.signal.value = X`` from concurrent tasks are flushed before
-    the fast coroutine resumes.
-
-**Phase-transition guards enforced** *(resolved)*
-    The fast scheduler raises :class:`RuntimeError` for illegal
-    transitions such as ``await ReadOnly()`` while already in the
-    ReadOnly phase, or ``await ReadWrite()`` from ReadOnly, matching
-    standard cocotb behavior.
-
 **Single-coroutine execution only** *(by design)*
     The fast scheduler drives exactly one coroutine.  There is no
     support for ``start_soon``, ``Combine``, ``First``, or any form


themperek:
@robtaylor In general interesting experiment.

As I can see fundamentally this is mayor show stopper to make it generally usable:
```
**No interaction with cocotb's event loop**
    While a fast loop is running, cocotb's :class:`~cocotb.EventLoop`
    is **not** pumped.  Other cocotb tasks (started via
    :func:`~cocotb.start_soon`) will not make progress until the fast
    loop completes and control returns to the standard scheduler.  Do
    not mix concurrent cocotb tasks with fast loops that expect those
    tasks to advance in lock-step.
 ```

I would need to think on this bit more but if one would provide a python wrapper direcelty around GPI maybe gets similar result.
  
What one would maybe like is to write cython directly (native) using GPI, but I am not sure how to make it easily available in "normal" python as coroutines.

robtaylor:
Thanks @themperek — good catch. I've pushed a fix: the fast scheduler now calls `cocotb._event_loop._inst.run()` after each step, which is the same pattern as `GPITrigger._react()`. This lets other cocotb tasks (Clock drivers, bus monitors, etc. started via `start_soon`) make progress while a fast loop is running.

When no other tasks are queued, `run()` just checks an empty deque and returns — so the performance cost is negligible for the isolated-loop case.

There are still some caveats documented in the module docstring (stale `current_gpi_trigger()`, no deferred write flushing on fast `ReadWrite`), but the core "other tasks are frozen" limitation is resolved.

> if one would provide a python wrapper directly around GPI maybe gets similar result

That's essentially what `SignalProxy` is — a thin wrapper around the GPI `sim_obj` handle that calls `get_signal_val_long()` / `set_signal_val_int()` directly, bypassing `Logic`/`LogicArray` construction. The scheduler layer on top is what eliminates the per-cycle coroutine overhead (Task._resume / coro.send / EventLoop dispatch).

> What one would maybe like is to write cython directly (native) using GPI

Agreed — that's the direction for Phase 2. The `_fast_sched.pyx` Cython scheduler already drives `coro.send()` + isinstance dispatch in compiled C, and `SignalProxy` could similarly be compiled. The pure-Python fallback ensures it works without Cython too.

robtaylor:
@themperek The three fixable behavioral differences you flagged are now addressed:

1. **`current_gpi_trigger()` updated** — The fast scheduler now sets `cocotb._gpi_triggers._current_gpi_trigger` to the standard `ReadOnly()`/`ReadWrite()` singletons when entering those phases, so concurrent cocotb tasks see correct phase awareness for write guards.

2. **`_apply_scheduled_writes()` called on ReadWrite** — Deferred writes queued by `dut.signal.value = X` from concurrent tasks are flushed when entering the ReadWrite phase, matching standard `ReadWrite._do_callbacks()` behavior.

3. **Phase-transition guards enforced** — The fast scheduler now raises `RuntimeError` for illegal transitions (ReadOnly→ReadOnly, ReadOnly→ReadWrite), matching standard cocotb behavior.

These are on top of the event loop pumping fix from the previous push. The behavioral differences doc in `fast.py` has been updated to mark all three as resolved.

PR #5357, opened 2026-02-23
Support multiple test failures in RegressionManager

### LINKED ISSUE #5332: RegressionManager should handle `ExceptionGroup`s
#5281 changed the TestManager to return a `BaseExceptionGroup` to the RegressionManager when there were multiple exceptions when a test ended. We should add support to the RegressionManager to handle them in a more intelligent way. The issue is that we want to be able to expect the exact shape of the ExceptionGroup (they can be very nested) and not just match every non-group exception inn the group against individually called out exceptions.

pytest solved this by introducing [`RaisesGroup`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.RaisesGroup).

Perhaps rather than reimplement the wheel, we add optional support for this to the RegressionManager for `xfail` and if the user doesn't want pytest, they just won't be able to have this funcitonality.

ktbarrett:
Closes #5332. Leverages `pytest.RaisesExc` and `pytest.RaisesGroup`.

The change to the scoring logic should increase readability. The `passed` variable was unused and removed.

Implementation may change depending upon #5363.

### TODO
- [x] newsfrag
- [x] documentation

### DIFF
--- a/src/cocotb/regression.py
+++ b/src/cocotb/regression.py
@@ -67,16 +67,16 @@ if hasattr(pytest, "RaisesGroup") and hasattr(pytest, "RaisesExc"):
             type[BaseException] | pytest.RaisesExc | pytest.RaisesGroup
         ],
     ) -> tuple[set[pytest.RaisesExc | pytest.RaisesGroup], bool]:
-    """Filter out :class:`pytest.RaisesExc` and :class:`pytest.RaisesGroup` exceptions and do checking on them.
+        """Filter out :class:`pytest.RaisesExc` and :class:`pytest.RaisesGroup` exceptions and do checking on them.
 
-    Args:
-        exc: The exception result of the test.
-        expected_error_set: The set of expected exceptions and :class:`!pytest.RaisesExc` and :class:`!pytest.RaisesGroup` objects.
+        Args:
+            exc: The exception result of the test.
+            expected_error_set: The set of expected exceptions and :class:`!pytest.RaisesExc` and :class:`!pytest.RaisesGroup` objects.
 
-    Returns:
-        A tuple of the filtered out :class:`!pytest.RaisesExc` and :class:`!pytest.RaisesGroup` objects
-        (so that the caller may remove them from the exception set)
-        and a boolean whether there was a match.
+        Returns:
+            A tuple of the filtered out :class:`!pytest.RaisesExc` and :class:`!pytest.RaisesGroup` objects
+            (so that the caller may remove them from the exception set)
+            and a boolean whether there was a match.
         """
         exception_matcher_excs = cast(
             "set[pytest.RaisesExc | pytest.RaisesGroup]",


ktbarrett:
Need to add tests for bad type to `cocotb.test(expect_error=...)` and a test where we expect an error and get a failure (`assert`). This will close coverage gaps enough. The remaining one will have to wait until we start testing `pytest` with the lowest supported version.

src/cocotb/_decorators.py
```diff
@@ -540,11 +547,25 @@ def decorator(obj: TestFuncType | TestGenerator) -> TestGenerator:
     return decorator
 
 
+_single_exception_types = (
+    type,
+    *(
+        (pytest.RaisesExc, pytest.RaisesGroup)
+        if hasattr(pytest, "RaisesExc") and hasattr(pytest, "RaisesGroup")
```
ktbarrett, line 554: I had to condition stuff on `RaisesExc` and `RaisesGroup` existing since they were introduced only in pytest 8.4, which is fairly recent.

src/cocotb/_decorators.py
```diff
@@ -581,11 +602,22 @@ def decorator(obj: TestFuncType | TestGenerator) -> TestGenerator:
...
+                        raise TypeError(
+                            "Expected error types must be an exception type or iterable of exception types"
+                        ) from None
+                    else:
+                        for exc in it:
+                            if not isinstance(exc, _single_exception_types):
+                                raise TypeError(
+                                    "Expected error types must be exception types"
+                                )
+                            obj.expect_error.add(exc)
```
ktbarrett, line 620: All of this was done was to make mypy happy since `isinstance(exc, _single_exception_types)` doesn't type narrow like it should.

But it's nice to have more informative error messages anyways.

src/cocotb/regression.py
```diff
@@ -59,6 +59,40 @@
...
 )
 
+if hasattr(pytest, "RaisesGroup") and hasattr(pytest, "RaisesExc"):
+
+    def handle_pytest_exception_matchers(
+        exc: BaseException,
+        expected_error_set: set[
+            type[BaseException] | pytest.RaisesExc | pytest.RaisesGroup
+        ],
+    ) -> tuple[set[pytest.RaisesExc | pytest.RaisesGroup], bool]:
```
ktbarrett, line 69:
```suggestion
    ) -> tuple[set[pytest.RaisesExc | pytest.RaisesGroup], bool]:
    """Filters out RaisesExc and RaisesGroup exceptions and does checking on them.

    Args:
        exc: The exception result of the test.
        expected_error_set: The set of expected exceptions and :class:`pytest.RaisesExc` and :class:`pytest.RaisesGroup` objects.

    Returns:
        A tuple of the filtered out :class:`pytest.RaisesExc` and :class:`pytest.RaisesGroup` objects
        (so that the caller may remove them from the exception set)
        and a boolean whether there was a match.
```

tests/test_cases/test_cocotb/test_tests.py
```diff
@@ -213,3 +217,56 @@ async def raise_test_success():
...
+
+
+@cocotb.test
+@cocotb.xfail(raises=pytest.RaisesGroup(pytest.RaisesGroup(ValueError)))
+async def test_xfail_with_raises_group_nested(dut: object) -> None:
+    raise ExceptionGroup("", (ExceptionGroup("", (ValueError(),)),))
+
+
+@cocotb.test
+async def test_bad_xfail(dut: object) -> None:
```
ktbarrett, line 249: Improve coverage on `cocotb.xfail`.

tests/test_cases/test_failure/test_failure.py
```diff
@@ -14,30 +14,35 @@
...
     raise Exception()
 
 
 @cocotb.test
 async def test_exception_with_nonprintable_characters(_: object) -> None:
     raise Exception("This is bad! \x00\x0b\x80")
+
+
+@cocotb.test(expect_error=TypeError)
+async def test_expect_error_get_failure(dut: object) -> None:
```
ktbarrett, line 47: Improve coverage on `_score_test` rewrite.

src/cocotb/regression.py
```diff
@@ -59,6 +59,40 @@
...
+                return exception_matcher_excs, True
+
+        return exception_matcher_excs, False
+
+else:
+
+    def handle_pytest_exception_matchers(
+        exc: BaseException,
+        expected_error_set: set[
+            type[BaseException] | pytest.RaisesExc | pytest.RaisesGroup
```
cmarqu, line 102: We determined pytest _doesn't_ have `RaisesExc` and `RaisesGroup` when in this branch, right?

src/cocotb/regression.py
```diff
@@ -450,83 +484,101 @@ def _score_test(
...
-                passed = False
+                expected_error_set = set(test.expect_error)
 
-            elif test.expect_fail:
-                self._record_test_failed(
-                    wall_time_s=wall_time_s,
-                    sim_time_ns=sim_time_ns,
-                    result=exc,
-                    msg="passed but we expected a failure",
+                # Filter out RaisesExc or RaisesGroup , which need to be handled differently.
```
```suggestion
                # Filter out RaisesExc or RaisesGroup, which need to be handled differently.
```

src/cocotb/regression.py
```diff
@@ -450,83 +484,101 @@ def _score_test(
...
+                    # Non-exception group error with unexpected type.
+                    return self._record_test_failed(
+                        wall_time_s=wall_time_s,
+                        sim_time_ns=sim_time_ns,
+                        result=exc,
+                        msg="errored with unexpected type",
+                    )
+            elif test.expect_fail:
+                if isinstance(exc, _TestFailures):
+                    # We expect a failure and got one.
```
```suggestion
                    # We expected a failure and got one.
```

src/cocotb/regression.py
```diff
@@ -450,83 +484,101 @@ def _score_test(
...
+                        result=exc,
+                        msg="failed as expected",
+                    )
+                else:
+                    # We expected a failure but got an unexpected exception type.
+                    return self._record_test_failed(
+                        wall_time_s=wall_time_s,
+                        sim_time_ns=sim_time_ns,
+                        result=exc,
+                        msg="errored but expected a failure",
```
```suggestion
                        msg="errored but we expected a failure",
```

src/cocotb/regression.py
ktbarrett: Function definitions in both sides of the branch must have the same type.

src/cocotb/regression.py
```diff
@@ -59,6 +59,50 @@
...
 
+if hasattr(pytest, "RaisesGroup") and hasattr(pytest, "RaisesExc"):
+
+    def handle_pytest_exception_matchers(
+        exc: BaseException,
+        expected_error_set: set[
+            type[BaseException] | pytest.RaisesExc | pytest.RaisesGroup
+        ],
+    ) -> tuple[set[pytest.RaisesExc | pytest.RaisesGroup], bool]:
+    """Filters out RaisesExc and RaisesGroup exceptions and does checking on them.
```
```suggestion
    """Filter out :class:`pytest.RaisesExc` and :class:`pytest.RaisesGroup` exceptions and do checking on them.
```

src/cocotb/regression.py
```diff
@@ -59,6 +59,50 @@
...
+
+    Args:
+        exc: The exception result of the test.
+        expected_error_set: The set of expected exceptions and :class:`pytest.RaisesExc` and :class:`pytest.RaisesGroup` objects.
+
+    Returns:
+        A tuple of the filtered out :class:`pytest.RaisesExc` and :class:`pytest.RaisesGroup` objects
```
```suggestion
        expected_error_set: The set of expected exceptions and :class:`!pytest.RaisesExc` and :class:`!pytest.RaisesGroup` objects.

    Returns:
        A tuple of the filtered out :class:`!pytest.RaisesExc` and :class:`!pytest.RaisesGroup` objects
```

src/cocotb/regression.py
```diff
@@ -59,6 +59,50 @@
...
+        (so that the caller may remove them from the exception set)
+        and a boolean whether there was a match.
+        exception_matcher_excs = cast(
```
cmarqu, line 81: Closing triple quote missing or is GitHub screwing with the display?
```suggestion
        """
        exception_matcher_excs = cast(
```

src/cocotb/regression.py
```diff
@@ -59,6 +59,51 @@
...
+    """Filter out :class:`pytest.RaisesExc` and :class:`pytest.RaisesGroup` exceptions and do checking on them.
+
+    Args:
+        exc: The exception result of the test.
+        expected_error_set: The set of expected exceptions and :class:`!pytest.RaisesExc` and :class:`!pytest.RaisesGroup` objects.
+
+    Returns:
+        A tuple of the filtered out :class:`!pytest.RaisesExc` and :class:`!pytest.RaisesGroup` objects
+        (so that the caller may remove them from the exception set)
+        and a boolean whether there was a match.
```
ktbarrett:
```suggestion
        """Filter out :class:`pytest.RaisesExc` and :class:`pytest.RaisesGroup` exceptions and do checking on them.

        Args:
            exc: The exception result of the test.
            expected_error_set: The set of expected exceptions and :class:`!pytest.RaisesExc` and :class:`!pytest.RaisesGroup` objects.

        Returns:
            A tuple of the filtered out :class:`!pytest.RaisesExc` and :class:`!pytest.RaisesGroup` objects
            (so that the caller may remove them from the exception set)
            and a boolean whether there was a match.
```

Issue #5073, opened 2025-10-12
I can no longer install cocotb==1.9 on my Mac (M1)


raysalemi:
I get the following clang++ error when I try to install:

  `ld: unsupported mach-o filetype (only MH_OBJECT and MH_DYLIB can be linked) in 'build/lib.macosx-26.0-x86_64-cpython-314/cocotb/libs/libgpilog.so'`

It is part of this message:

```
      building 'cocotb/libs/libgpilog' extension
      creating build/temp.macosx-26.0-x86_64-cpython-314/cocotb/libs/libgpilog/cocotb/share/lib/gpi_log
      clang++ -fno-strict-overflow -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -O3 -Wall -DGPILOG_EXPORTS= -D__STDC_FORMAT_MACROS= -Icocotb/share/include -Icocotb -I/usr/local/opt/python@3.14/Frameworks/Python.framework/Versions/3.14/include/python3.14 -c cocotb/share/lib/gpi_log/gpi_logging.cpp -o build/temp.macosx-26.0-x86_64-cpython-314/cocotb/libs/libgpilog/cocotb/share/lib/gpi_log/gpi_logging.o -std=c++11 -fvisibility=hidden -fvisibility-inlines-hidden -Wall -Wextra -Wcast-qual -Wwrite-strings -Wconversion -Wno-missing-field-initializers -Wnon-virtual-dtor -Woverloaded-virtual -flto
      creating build/lib.macosx-26.0-x86_64-cpython-314/cocotb/libs
      clang++ -fno-strict-overflow -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -O3 -Wall -bundle -undefined dynamic_lookup build/temp.macosx-26.0-x86_64-cpython-314/cocotb/libs/libgpilog/cocotb/share/lib/gpi_log/gpi_logging.o -Lbuild/lib.macosx-26.0-x86_64-cpython-314/cocotb/libs -o build/lib.macosx-26.0-x86_64-cpython-314/cocotb/libs/libgpilog.so -flto -Wl,-install_name,@rpath/libgpilog.so -Wl,-rpath,@loader_path
      building 'cocotb/libs/libpygpilog' extension
      creating build/temp.macosx-26.0-x86_64-cpython-314/cocotb/libs/libpygpilog/cocotb/share/lib/py_gpi_log
[...]

PR #5361, opened 2026-02-24
Replace deprecated distutils.sysconfig.get_config_vars() with sysconfig.get_config_vars()

aelmahmoudy:


ktbarrett:
Weird error messages... I'm not sure this is related to your change, but maybe you can look into it and confirm.

aelmahmoudy:
It might be related. I suppose that sysconfig.get_config_vars() isn't 
compatible with distutils.sysconfig.get_config_vars() on macos. I don't have access to a mac machine though.

Could this be related ?
https://stackoverflow.com/questions/71548156/how-to-get-architecture-of-running-python-interpreter-on-macos-with-apple-silico

ktbarrett:
Ah, I remember this now https://github.com/cocotb/cocotb/issues/5073. It's some weirdness with distutils on Mac building against x86 and running in an emulator rather than building and running natively.

ktbarrett:
setuptools 84 changed where distutil's configuration lives, so this should work now. Re-running CI.

ktbarrett:
Well fun. Maybe we just don't use setuptools 84 since we plan on ditching this build system right after the release.

PR #5366, opened 2026-02-25
Add test listing mode to RegressionManager

### LINKED ISSUE #5143: Add a "test listing" mode
> With cocotb 2.0’s parametrize, identifying the effective set of testcases can get complicated. For usability, it would be helpful if cocotb printed (or otherwise exposed) the list of testcases for the selected MODULE before execution—ideally even in a “list only”/dry‑run mode. While you can inspect the testcase list in results.xml, that file only exists after the entire run completes; I often want to see what will run ahead of time.

 _Originally posted by @thuako in [#5137](https://github.com/cocotb/cocotb/issues/5137#issuecomment-3526217518)_

Listing the tests is already done at the start of the simulation, do maybe a `COCOTB_DRY_RUN` would work? This would list the tests that match the filter, and emit an appropriate results.xml with all tests as skipped so that users and tools can confirm what's generated and what would be run.

ktbarrett:
Closes #5143. Adds `COCOTB_LIST_TESTS` to list tests that were collected instead of running them. They are listed in the order they would be run. There is a starting and ending log for people that want to analyse the list programmatically.


ktbarrett:
Mac's grep doesn't support `-P` or `-z` so we will have to change the newline matching. Riviera and Questa start every line output with `# ` because yeah.

ktbarrett:
I haven't the slightlest clue what's going on. Maybe the logs are using `\r\n`? Will need some time on the dev box.

ktbarrett:
Running this test manually on the dev box works. I have no clue what is causing this to be such an issue in CI, but really we only need this to work in one simulator to prove it works, so we just exempt Questa and Riviera.

docs/source/library_reference.rst
```diff
@@ -311,6 +311,14 @@ Discovering Tests
 
         Only one of :envvar:`COCOTB_TESTCASE` or :envvar:`COCOTB_TEST_FILTER` should be used.
 
+.. envvar:: COCOTB_LIST_TESTS
```
marlonjames, line 314: Should be in `cocotb-config --help-vars` too.

src/cocotb_tools/config.py
```diff
@@ -82,6 +82,7 @@ def _help_vars_text() -> str:
         COCOTB_REWRITE_ASSERTION_FILES
                                  Files to apply pytest assertion rewrites to (default *.py)
         COCOTB_MAX_FAILURES      Maximum number of test failures before aborting the regression
+        COCOTB_TEST_LIST         Prints all tests in the order they would be executed and exits
```
marlonjames:
```suggestion
        COCOTB_LIST_TESTS        Prints all tests in the order they would be executed and exits
```

marlonjames (approved):

PR #5368, opened 2026-02-25
Fix VPI error checking

### LINKED ISSUE #5367: `vpi_chk_error` returns different type in old Verilog standards
In 1364-2001 `vpi_chk_error` returns 0 for no error and non-0 for error. While in 1800-2017+ it returns 0 for no error and one of the severity enums if there was an error. We are assuming we can use the new semantics, but Xcelium uses the old semantics. This leads them to returning 1, which maps to `vpiNotice` in the new semantics. This underreports the severity of errors coming from the simulator.

Instead we should be using the `level` field of the `error_info_p` struct that is passed to the function.

ktbarrett:
Closes #5367.

### DIFF
--- a/src/cocotb/share/lib/gpi/vpi/VpiImpl.hpp
+++ b/src/cocotb/share/lib/gpi/vpi/VpiImpl.hpp
@@ -32,17 +32,16 @@ static inline void __check_vpi_error(const char *file, const char *func,
         return;
     }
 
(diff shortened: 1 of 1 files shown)


ktbarrett:
Patch coverage should incrase after #5370 is in.

marlonjames (approved):

PR #5371, opened 2026-02-26
Add cocotb.tops to expose all top-level design units

### LINKED ISSUE #5365: Support multiple tops
It's currently pretty easy to get other tops in a design using `cocotb.simulator` and `_make_sim_object`, but it'd be nice to have a top level attribute such as `cocotb.tops` that includes all the tops easily available.

ZiaCheemaGit:
partially fixes issue #5365


src/cocotb/__init__.py
```diff
@@ -106,5 +106,12 @@
 and in parameters to :class:`.TestFactory`\ s.
 """
 
+tops: dict = {}
+"""This contains all the top level modules including top i.e :envvar:`COCOTB_TOPLEVEL`
+
+If there are multiple hdl top level modules added to VERILOG_SOURCES
```
ZiaCheemaGit: Automatically infer all tops internally

src/cocotb/simulator.pyi
```diff
@@ -33,6 +33,7 @@ VALUE_CHANGE: int
 RANGE_UP: int
 RANGE_DOWN: int
 RANGE_NO_DIR: int
+tops: dict
```
ZiaCheemaGit: This is not required. Forgot to remove this

Will remove in next commit

ktbarrett:
> Does this mean vpi_iterate can be used to populate tops but only when VPI is used. For VHPI and FLI we need to figure some other way right?

This is correct. For VHPI there should be analogue to the functions I mentioned `vhpi_iterate` and maybe something like `vhpiInstance`.

As for FLI, I don't know. Maybe we can just ignore that for now. We are hoping that someday soon we can drop the FLI, as Questa now has a VHPI. So maybe just populate the `tops` dict with the one top we are already discovering.

marlonjames:
At the top level, VHPI has one root instance (`rootInst`) and 0 to many packages (`packInst`).

ktbarrett:
So I think for VHPI you only have one element to populate into the `tops` dict. But if you are doing multi-language simulation, you may find one top with VHPI and multiple with VPI, even if the TOPLEVEL_LANG is VHDL.

I think what's necessary is probably a custom iterator. This should be almost *exactly* like how `cocotb.packages` is defined.

ZiaCheemaGit:
@ktbarrett locally it builds fine but CI build fails issue seems to be the vector library in gpi.h 
i.e
-- GPI_EXPORT gpi_sim_hdl gpi_get_root_handle(const char *name);
++ GPI_EXPORT std::vector<gpi_sim_hdl> gpi_get_root_handle(const char *name);

ZiaCheemaGit:
image

I could not understand why build fails on CI but works locally

src/cocotb/share/include/gpi.h
```diff
@@ -159,7 +161,7 @@ typedef enum gpi_discovery_e {
  * @param name  Name of the root object, or `NULL`.
  * @return      Handle to simulation object or `NULL` if not found.
  */
-GPI_EXPORT gpi_sim_hdl gpi_get_root_handle(const char *name);
+GPI_EXPORT std::vector<gpi_sim_hdl> gpi_get_root_handle(const char *name);
```
marlonjames: This is not going to work, and you shouldn't try to do this.
gpi.h provides a C compatible interface.

ZiaCheemaGit:
I have mid exam from tomorrow uptill 13th, will hopefully look more into this afterwards

ZiaCheemaGit:
@marlonjames @ktbarrett  can you point towards how can I change my local python build config to mirror CI one because I already did worked around this `std::vecctor` and still got build error for some other changes in c side but locally it worked perfect

marlonjames:
You can add `-Werror` to `CFLAGS` and `CXXFLAGS` envars to ensure warnings become errors.

ZiaCheemaGit:
@marlonjames @ktbarrett 
This works but maybe I shouldn't change `get_root_handle` to `get_all_root_handles` but if an entirely new api is created then we miss the chance to get `TOPLEVEL ie cocotb.top` and  `cocotb.tops` in one api call.

Another thing I dont know why private CI is failing but for verilator --top-module is necessary, see [this](https://github.com/cocotb/cocotb/actions/runs/22817221039/job/66184077397?pr=5371#step:29:1974). Should this be dropped or is there a fix for that maybe some work around in `verilator Makefile`

And If private CI is also failing for same reason as Verilator then this needs a different approach altogether.

ktbarrett:
One thing here is that I actually wouldn't change `get_root_handle`. People use this (for better or worse) to access signals using a full path, e.g. `cocotb.simulator.get_root_handle("top.path.to.sim.object.field[0]")`. I would create a separate function.

Next returning the array and doing all the buffer management is probably not a great idea. You should follow the approach shown in https://github.com/cocotb/cocotb/pull/3899 for creating a new iterator type, an iterator over all `vhpi_iterate(vhpiRootInst, NULL)` or `vpi_iterate(vpiModule, NULL)`.

ktbarrett:
Its looking better now. Need to rebate to resolve conflicts.

ZiaCheemaGit:
@ktbarrett still verilator and all the private CI require a mandatory top module ie COCOTB_TOPLEVEL and doing so makes VPI to only expose that module as root module

tests/test_cases/test_cocotb_tops/tops.vhdl
ZiaCheemaGit, line 1: delete in next commit, VHPI doesnt support multiple roots

src/cocotb/share/lib/gpi/vpi/VpiIterator.cpp
```diff
@@ -182,6 +182,39 @@ GpiIterator::Status VpiSingleIterator::next_handle(std::string &name,
...
+    const char *obj_name = vpi_get_str(vpiFullName, root);
+    check_vpi_error();
+
+    if (!obj_name) return next_handle(*(new std::string), hdl, NULL);
+
+    std::string root_name = obj_name;
+
+    VpiImpl *vpi_impl = reinterpret_cast<VpiImpl *>(m_impl);
+
+    LOG_CRITICAL("vpi_root_iterator: root found '%s'", root_name.c_str());
```
ZiaCheemaGit: Change log level

ktbarrett:
Verilator supports specifying multiple tops, `--top` can be used multiple times if I'm reading the documentation correctly. I'm sure the commercial sims support the same. You'll need to have the Makefiles and Runner take a list of tops and translate that into multiple top specifications.

ZiaCheemaGit:
> Verilator supports specifying multiple tops, `--top` can be used multiple times if I'm reading the documentation correctly. I'm sure the commercial sims support the same. You'll need to have the Makefiles and Runner take a list of tops and translate that into multiple top specifications. 

--top-module can be used to specify multiple tops but vpi still gets to see only one top  
I checked by running verilator directly on file containing two top modules and passed both of them to verilator but generated cpp files are only for only one top, the later one. Also now `DUT` envar is not used, icarus Makefile directly uses multiple tops specified by `COCOTB_TOPLEVEL` 
image

ktbarrett:
Ahhh, right. Verilator builds each into its own Verilated model and you have to include and run both together. Well sounds like we skip Verilator then, as someone would need to overhaul `verilator.cpp` to support multiple tops regardless of this change.

How about the other simulators? This is a bit of a science project. But if we can't get them all working, if as long as one or two work, we can merge the feature and have people contribute support for their simulator at a future time.

ZiaCheemaGit:
As of now only Icarus truly supports multiple tops. 

I can open a separate PR for verilator support. 

Remaining are private CI simulators, that can be left for future

ktbarrett:
The failures are caused by `cocotb.tops[top_name]` not working with the fact that VHDL is case insensitive. The casing used in the definition is not necessarily the casing used in the map. You'll have to instead iterate over the list and compare the casefolds of the names.

ZiaCheemaGit:
still private vhdl and verilog on riviera is failing

ZiaCheemaGit:
@ktbarrett @marlonjames

src/cocotb/share/lib/gpi/vpi/VpiIterator.cpp
```diff
@@ -182,6 +182,39 @@ GpiIterator::Status VpiSingleIterator::next_handle(std::string &name,
...
+    if (root == NULL) return GpiIterator::END;
+
+    auto gpi_module = to_gpi_objtype(vpi_get(vpiType, root));
+    if (gpi_module != GPI_MODULE)
+        return next_handle(*(new std::string), hdl, NULL);
+
+    const char *obj_name = vpi_get_str(vpiFullName, root);
+    check_vpi_error();
+
+    if (!obj_name) return next_handle(*(new std::string), hdl, NULL);
```
ktbarrett: This leaks. I think in this case you just want to return `GpiIterator::END` to signal the iterator has ended.

src/cocotb/share/lib/gpi/vpi/VpiIterator.cpp
```diff
@@ -182,6 +182,39 @@ GpiIterator::Status VpiSingleIterator::next_handle(std::string &name,
...
+    if (m_iterator == NULL) return GpiIterator::END;
+
+    vpiHandle root = vpi_scan(m_iterator);
+    check_vpi_error();
+
+    if (root == NULL) return GpiIterator::END;
+
+    auto gpi_module = to_gpi_objtype(vpi_get(vpiType, root));
+    if (gpi_module != GPI_MODULE)
+        return next_handle(*(new std::string), hdl, NULL);
```
ktbarrett: This too.

ktbarrett (commented):
Overall looks good. The Riviera job is broken right now (expired license) so we will wait until that's in and if this is in a good state it should make it in 2.1.

src/cocotb_tools/makefiles/simulators/Makefile.verilator
```diff
@@ -67,7 +69,7 @@ $(SIM_BUILD)/Vtop: $(SIM_BUILD)/Vtop.mk
 $(COCOTB_RESULTS_FILE): $(SIM_BUILD)/Vtop $(CUSTOM_SIM_DEPS)
 	$(RM) $(COCOTB_RESULTS_FILE)
 
-	-$(COCOTB_ENV) \
+  -$(COCOTB_ENV) \
```
ktbarrett: This is wrong. Rule bodies must be indented with tabs.

src/cocotb/__init__.py
```diff
@@ -110,5 +110,12 @@
...
+or VHDL_SOURCES. This gets populated with all of them
+"""
```
ktbarrett:
```suggestion
"""A dictionary of top-level module name to :class:`module handle <cocotb.handles.HierarchyObject>` including :data:`cocotb.top`.

.. versionadded: 2.1
"""
```

src/cocotb/share/lib/pygpi/bind.cpp
```diff
@@ -1118,6 +1124,14 @@ static PyMethodDef SimulatorMethods[] = {
                "Get an iterator handle to loop over all HDL packages.\n"
                "\n"
                ".. versionadded:: 2.0")},
+    {"root_iterate", root_iterate, METH_NOARGS,
+     PyDoc_STR("root_iterate(/)\n"
+               "--\n\n"
+               "root_iterate() -> cocotb.simulator.sim_obj_iterator\n"
+               "Get an iterator handle to loop over all root handles for VPI "
+               "Implication.\n"
```
ktbarrett: ???

src/cocotb/share/lib/gpi/GpiCommon.cpp
```diff
@@ -503,8 +503,12 @@ gpi_iterator_hdl gpi_iterate(gpi_sim_hdl obj_hdl, gpi_iterator_sel type) {
...
         for (implIter = registered_impls.begin();
              implIter != registered_impls.end(); implIter++) {
-            GpiIterator *iter =
-                (*implIter)->iterate_handle(NULL, GPI_PACKAGE_SCOPES);
+            GpiIterator *iter = NULL;
+            if (type == GPI_ROOTS && (*implIter)->get_name_s() == "VPI") {
+                iter = (*implIter)->iterate_handle(NULL, type);
+            } else if (type == GPI_PACKAGE_SCOPES) {
+                iter = (*implIter)->iterate_handle(NULL, type);
+            }
```
ktbarrett: Looks like you can reduce this, both branches are the same. Just add null iterators to FLI and VHPI for this.

docs/source/quickstart.rst
```diff
@@ -0,0 +1,243 @@
+.. _quickstart:
```
ktbarrett: Accidentally added this back.

.github/generate-envs.py
```diff
@@ -51,8 +51,6 @@
         "sim": "nvc",
         "sim-version": "r1.20.0",
         "os": "ubuntu-22.04",
-        "python-version": "3.10",
-        "group": "ci-free",
```
ktbarrett, line 55: This needs to be added back.

src/cocotb/share/lib/gpi/vpi/VpiIterator.cpp
ktbarrett: Actually what I think is needed is just continue iterating in a loop until you get a module or exhaustion. Then you can return `GpiIterator::END`.

src/cocotb/share/lib/gpi/vpi/VpiIterator.cpp
```diff
@@ -182,6 +182,39 @@ GpiIterator::Status VpiSingleIterator::next_handle(std::string &name,
         return GpiIterator::NOT_NATIVE;
 }
 
+GpiIterator::Status VpiRootIterator::next_handle(std::string &, GpiObjHdl **hdl,
```
ktbarrett, line 185: This looks a lot better, thanks.

src/cocotb_tools/makefiles/Makefile.inc
```diff
@@ -146,6 +146,13 @@ endif
 
 export COCOTB_TRUST_INERTIAL_WRITES
 
+# Icarus supports multiple tops
+ifeq ($(SIM),icarus)
+    _COCOTB_ENV_TOPLEVEL = $(firstword $(call deprecate,TOPLEVEL,COCOTB_TOPLEVEL))
+else
+    _COCOTB_ENV_TOPLEVEL = $(call deprecate,TOPLEVEL,COCOTB_TOPLEVEL)
+endif
+
```
ktbarrett:
```suggestion
```

src/cocotb_tools/makefiles/Makefile.inc
```diff
@@ -154,7 +161,7 @@ COCOTB_ENV = \
     COCOTB_TEST_MODULES=$(call deprecate,MODULE,COCOTB_TEST_MODULES) \
     COCOTB_TESTCASE=$(call deprecate,TESTCASE,COCOTB_TESTCASE) \
     $(if $(COCOTB_TEST_FILTER),COCOTB_TEST_FILTER=$(COCOTB_TEST_FILTER)) \
-    COCOTB_TOPLEVEL=$(call deprecate,TOPLEVEL,COCOTB_TOPLEVEL) \
+    COCOTB_TOPLEVEL=$(_COCOTB_ENV_TOPLEVEL) \
```
ktbarrett: Simplify.
```suggestion
    COCOTB_TOPLEVEL=$(firstword $(call deprecate,TOPLEVEL,COCOTB_TOPLEVEL)) \
```

ktbarrett (commented):
The implementation looks good now. Shame it only works on Icarus. You need to add a newsfragment, add `cocotb.tops` to `library_reference.rst`, and open an issue into porting this to the Runners.

docs/source/newsfragments/5371.feature.rst
```diff
@@ -0,0 +1 @@
+Added ``cocotb.tops`` to expose a dictionary of all top-level design units for the Icarus simulator.
```
ktbarrett:
```suggestion
Added :data:`cocotb.tops` to expose a dictionary containing all top-level design units.
```

ktbarrett (approved):

PR #5376, opened 2026-03-03
Remove string setting hack in Questa VPI

### LINKED ISSUE #4082: Can't set string variables using `vpiInertialDelay` in Questa
We get the following when attempting to write to a string variable using `vpiInertialDelay`.

We have to use the following logic to instead set string variables using `vpiNoDelay`, which works for some reason.
https://github.com/cocotb/cocotb/blob/49a9f19e31df31963cc5ad6ccf5c365dd0d5bbd0/src/cocotb/share/lib/vpi/VpiSignal.cpp#L205-L214

ktbarrett:
Closes #4082.

### DIFF
--- a/src/cocotb/share/lib/gpi/vpi/VpiSignal.cpp
+++ b/src/cocotb/share/lib/gpi/vpi/VpiSignal.cpp
@@ -191,18 +191,7 @@ int VpiSignalObjHdl::set_signal_value(s_vpi_value value_s,
 
     switch (action) {
         case GPI_DEPOSIT:
-#if defined(MODELSIM) || defined(IUS)
-            // Xcelium and Questa do not like setting string variables using
-            // vpiInertialDelay.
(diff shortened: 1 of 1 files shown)


src/cocotb/share/lib/gpi/vpi/VpiSignal.cpp
```diff
@@ -191,18 +191,7 @@ int VpiSignalObjHdl::set_signal_value(s_vpi_value value_s,
 
     switch (action) {
         case GPI_DEPOSIT:
-#if defined(MODELSIM) || defined(IUS)
```
cmarqu, line 194: You probably wanted to only remove the `MODELSIM` part?

ktbarrett: Eh, I wanted to see if either were fixed. Nope.

PR #5391, opened 2026-03-06
Fix `cocotb.pass_test` forcing test to pass

### LINKED ISSUE #5377: Fix `cocotb.pass_test`
In 2.0 this function was a forcing test pass, i.e. it ignored `expect_error` and `expect_fail` tags on the test. This regressed in 2.1 where it is now a way to end the test early without forcing an outcome, and `expect_error` and `expect_fail` are taken into account. This should be fixed.

ktbarrett:
Closes #5377. This accidentally regressed with some of the refactoring done to TestManager. Added tests to prevent that from happening again.


tests/pytest_plugin/test_pass.py
```diff
@@ -1,16 +0,0 @@
-# Copyright cocotb contributors
```
ktbarrett, line 1: pytest has no analogue to forcing a test to pass, so this will never work. Another reason to deprecate it.

marlonjames (approved):

marlonjames:
I think you need to remove `"test_pass"` from the `mark.cocotb_runner` list in `tests/pytest_plugin/test_sample_module.py`

marlonjames (approved):

Issue #5394, opened 2026-03-07
Add reference documentation on test pass/fail


ktbarrett:
There is a section on test pass/fail conditions in writing_testbenches.rst, which seems like an inappropriate place for what is reference documentation. Additionally, this section should be updated to reference the new support for `pytest.skip`, `pytest.xfail`, and `cocotb.end_test`.

PR #5392, opened 2026-03-07
Add `cocotb.end_test` to end the test without forcing a pass

ktbarrett:
Closes #5378. This new function does what `cocotb.pass_test` did before it was fixed to return to it's 2.0 behavior in #5391: end the test immediately and allow any `expect_error`, `expect_fail` or `cocotb.xfail` predicate to judge overall test pass/fail.

This will take the place `cocotb.pass_test` in the documentation and in use in a follow on set of commits which will:
* document test end conditions and whether they are pass or fail
* deprecate cocotb.pass_test and replace any usage internally with cocotb.end_test

Documentation will come in separate PR: #5394.

PR #5395, opened 2026-03-07
Deprecate `cocotb.pass_test`

### LINKED ISSUE #5379: Deprecate `cocotb.pass_test`
This was created to force a test pass, mostly as an alternative interface for `TestSuccess`. But giving users the ability to force a test to pass isn't a great idea. Before [`pytest.skip`](#4476) was supported, it was a way to programmtically end the test early with a pass, much like would be done with `pytest.skip`. But with `pytest.skip` support it isn't needed for this. Also if #5378 is introduced, it could provide a point of confusion for users who will use `pass_test` instead of the other, or vice versa, and get unintended results. Finally, there is no analogue in pytest making support with the pytest plugin impossible.

ktbarrett:
Closes #5379. Depends on #5392.

Documentation will come in separate PR: #5394.

### TODO
- [x] newsfrag

### DIFF
--- a/docs/source/upgrade-2.0.rst
+++ b/docs/source/upgrade-2.0.rst
@@ -1276,19 +1276,20 @@ This is unexpected and generally not useful.
     assert cocotb.top.data.value == 0
 
 
-********************************************************************
-Use :func:`!cocotb.pass_test` instead of raising :exc:`!TestSuccess`
-********************************************************************
+*******************************************************************************************
+Use :func:`!cocotb.end_test` or :func:`!pytest.skip` instead of raising :exc:`!TestSuccess`
+*******************************************************************************************
 
 Change
 ======
 
-:external+cocotb19:py:exc:`~cocotb.result.TestSuccess` was deprecated and replaced with :func:`cocotb.pass_test`.
+:external+cocotb19:py:exc:`~cocotb.result.TestSuccess` was deprecated and replaced with :func:`cocotb.end_test` and :func:`pytest.skip`.
 
 How To Upgrade
 ==============
 
-Replace ``raise TestSuccess(msg)`` with a call to :func:`!cocotb.pass_test`.
+Replace ``raise TestSuccess(msg)`` with a call to :func:`!cocotb.end_test` when you need to end the test immediately.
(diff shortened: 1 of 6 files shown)


docs/source/upgrade-2.0.rst
```diff
@@ -1276,19 +1276,20 @@ This is unexpected and generally not useful.
...
 
-:external+cocotb19:py:exc:`~cocotb.result.TestSuccess` was deprecated and replaced with :func:`cocotb.pass_test`.
+:external+cocotb19:py:exc:`~cocotb.result.TestSuccess` was deprecated and replaced with :func:`cocotb.end_test` and :func:`pytest.skip`.
 
 How To Upgrade
 ==============
 
-Replace ``raise TestSuccess(msg)`` with a call to :func:`!cocotb.pass_test`.
+Replace ``raise TestSuccess(msg)`` with a call to :func:`!cocotb.end_test` when you need to end the test immediately.
+Replace ``raise TestSuccess(msg)`` with a call to :func:`!pytest.skip` when you need to end the test immediately and force the outcome to 'skip'.
```
cmarqu:
```suggestion
Replace ``raise TestSuccess(msg)`` with a call to :func:`!pytest.skip` when you need to end the test immediately and force the outcome to 'skipped'.
```

src/cocotb/_test_manager.py
```diff
@@ -303,26 +312,62 @@ def create_task(
...
+    or ``expect_error`` and ``expect_fail`` arguments to :deco:`cocotb.test` are set.
 
     Args:
         msg: The message to display when the test passes.
 
     .. versionadded:: 2.0
+
+    .. deprecated:: 2.1
+        Forcing a test to pass is considered bad practice.
+        Use :func:`cocotb.skip` if you want to end the test immediately and force the outcome to skipped.
```
cmarqu:
```suggestion
        Use :func:`cocotb.skip` if you want to end the test immediately and force the outcome to 'skipped'.
```

cmarqu (approved):
Two small doc changes.

ktbarrett:
@cmarqu Can you approve #5392 as well? This PR is based on top of that one and doesn't have all the same changes.

PR #5398, opened 2026-03-07
Fixup release notes

ktbarrett:
Applied the `:external+cocotb20` tag to all refs in the release notes for version 2.0. Also I merged the 2.0b1 and 2.0 release notes manually. I'm not sure if keeping them separate is valuable or confusing.

The tagging process was done with a simple find and replace in my editor. The pattern I used to locate all tags needing updating was:

```
(?<!:external[^:]+)(:(?:|func|deco|class|mod|ref|attr|data|meth|exc):`(?:\.[^`]+|cocotb[^`]+|[^!][^`]*cocotb[^`]+)|:make:var:`[^!][^`]+|:envvar:`[^!]?[^`]*(?:cocotb|gpi|pygpi)[^`]+)
```

This matches everything (besides `:pr:`) that doesn't already have a `:external+...` tag and either starts with a `.` (we would only use that for unique local names), does not start with a `!`, or contains the word `cocotb`. `:make:var:` is always assumed to be local. `:envvar:` matches everything that contains `cocotb`, `gpi`, or `pygpi` (which may not be everything so check twice).


cmarqu:
This needs three more lines like https://github.com/cocotb/cocotb/blob/584b52aef03f053d19973e08f0ed84d07f562ade/docs/source/_static/cocotb.css#L531-L533 for 2.0.

Also, I noticed that we have linked `cocotb.parametrize` in three different styles in the Release Notes, maybe you could fix that while here?
Oh, and `cocotb.test()` could get also the new `:deco:` role.

More still - unnecessary duplication in:
```
Added range, left, direction, and right properties to LogicArrayObject. (#3733)

Added range, left, direction, and right properties to ArrayObject, LogicArrayObject, and StringObject. (#3733)

Added range, left, direction, and right properties to StringObject. (#3733)
```

ktbarrett:
@cmarqu Fixed all those issues.

PR #5402, opened 2026-03-08
Fix release build process

ktbarrett:
I removed the special `test_sdist` step from the release flow. When the nox env `"release_test and {sim} and {lang}"` is selected, both the sdist and wheel tests will be run, each in their own venv.

I merged the `release_build_sdist` and `release_build_wheel` into a single `release_build` since building that empty venv with uv is actually pretty quick now.

And I removed support for running `release_build` in the `release_test` if builds weren't found, as I don't want the case where the download of the artifact fails/is malformed, it does a new build and tests that, but then what's deployed is the uploaded artifacts, not what was tested. This should never happen, but better safe than sorry.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -420,15 +420,27 @@ def release_build_sdist(session: nox.Session) -> None:
     session.log(f"Source distribution in release mode built into {dist_dir!r}")
 
 
+@nox_uv.session(
+    uv_no_install_project=True,
+    uv_groups=[],
+    uv_sync_locked=False,
+)
+def release_build(session: nox.Session) -> None:
+    """Build all distributions for release."""
+    session.notify("release_build_wheel")
+    session.notify("release_build_sdist")
+
+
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -468,16 +480,16 @@ def release_install_from_wheel(session: nox.Session) -> None:
...
+            f"No potential wheel found in the {dist_dir!r} directory. Run the 'release_build_wheel' session first!"
+        )
     session.log(f"Installing cocotb from wheels in {dist_dir!r}")
     session.install(
         "--force-reinstall",
         "--only-binary",
         "cocotb",
         "--no-index",
-        "--no-dependencies",
+        "--no-deps",
```
ktbarrett, line 492: This works on both `uv pip` and `pip`. Not sure why `pip` has multiple aliases for this.

imphil (approved):
Looking good!

imphil:
Don't forget to remove the NOCOMMIT change before merging, @ktbarrett

PR #5232, opened 2026-01-03
chore(env): consistent environment variables

tymonx:
- Moved `src/cocotb_tools/pytest/env.py` to `src/cocotb_tools/_env.py` (private)
- Added unit tests `tests/pytest/test_env.py`
- Code coverage for the `cocotb_tools._env` module is 100%
- Replaced `from cocotb_tools.pytest import env` with `from cocotb_tools import _env`
- Replaced `os.environ.get()` with `_env.as_*` functions in all `src/cocotb/**/*.py` and `src/cocotb_tools/**/*.py` modules
- Replaced `os.environ[]` with `_env.as_*` functions in all `src/cocotb/**/*.py` and `src/cocotb_tools/**/*.py` modules
- Replaced `os.getenv()` with `_env.as_*` functions in all `src/cocotb/**/*.py` and `src/cocotb_tools/**/*.py` modules
- Simplified handling environment variables
- Documentation about environment variables
- News fragments about environment variables
- Added types for environment variables in documentation
- Updated info about the [NO_COLOR](https://no-color.org/) environment variable. It was out-dated

Closes #5024

PR #5411, opened 2026-03-11
Restore old (pre #5232) root handle name resolution logic

gilbertoabram:
The C++ code checks against `!name`, but after changing `_setup_root_handle()` to use `_env.as_str()` we get an empty string instead of None, which gets translated to a valid pointer.

### DIFF
--- a/src/cocotb/_init.py
+++ b/src/cocotb/_init.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import logging
+import os
 import random
 import tempfile
 import time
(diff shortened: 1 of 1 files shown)


gilbertoabram:
Alternatively (or in addition to this), it could be fixed in `VpiImpl::get_root_handle(const char *name)` (and possibly in the other impls) by checking for an empty string.

ktbarrett:
Can we just remove the `_env.as_str` thing? I'm not particularly happy with that change.

ktbarrett (approved):

PR #5412, opened 2026-03-11
VPI: allow indexing into GPI_PACKED_OBJECT

gilbertoabram:


gilbertoabram:
Wrong solution.

The problem I'm trying to solve is that with verilator something like
```
entry_t [num_entries-1:0] entries;
```
has type `vpiReg`with `is_vector` true and `num_elements` greater than one.
So `to_gpi_objtype()` (in `VpiImpl.cpp`) picks packed object as its type:
```
            if (is_vector || num_elements > 1) {
                return GPI_PACKED_OBJECT;
            } else {
                return GPI_LOGIC;
            }
```
(used to be `GPI_LOGIC_ARRAY`)

PR #5414, opened 2026-03-12
Small updates to modeling types

ktbarrett:
Removes guarantee that Logic and Bit are singletons comparable with `is`. Fixes incorrect type in `Array.index`.


src/cocotb/types/_abstract_array.py
```diff
@@ -112,7 +112,7 @@ def index(
         for i in Range(start, self.direction, stop):
             if self[i] == value:
                 return i
-        raise IndexError(f"{value!r} not in array")
```
ktbarrett, line 115: Sequences raise `ValueError` here, and even the documentation mentions `ValueError`.

marlonjames (approved):

PR #5423, opened 2026-03-17
Resolved verilator hanging on $error when waves are enabled

sumkad2:
There was an issue in the `wrap_up` function defined in `src/cocotb/share/lib/verilator/verilator.cpp` which would result in cocotb hanging if verilator was built with waves enabled. The `runExitCallbacks()` function defined in verilator acquires the `VlCbStatic.s_exitCbs` lock when it is called. It then calls the `wrap_up` function which was added as an exit callback. The `wrap_up` function calls the destructor for the trace file pointer which attempts to reacquire the `VlCbStatic.s_exitCbs` lock resulting in the process hanging. This change moves the destruction handling for the trace file pointer outside of the exit callback function, which prevents this hanging.

### DIFF
--- a/docs/source/newsfragments/5423.bugfix.rst
+++ b/docs/source/newsfragments/5423.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed Verilator hanging when ``$fatal`` is triggered in simulations with waves/tracing enabled.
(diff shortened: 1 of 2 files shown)


ktbarrett:
Is there a Verilator issue about them using a re-entrant lock here? I'm fine to merge this either way as it's equivalent and not break-y.

sumkad2:
We could use a re-entrant lock, but that would involve updating a container while it is being iterated through. This seemed undesirable.

cmarqu:
A fix for a user-visible bug is worth a newsfragment (see https://docs.cocotb.org/en/development/contributing.html#patch-requirements).

ktbarrett (approved):

PR #5424, opened 2026-03-18
Let exceptions that occur during GPI logging reach the python caller

gilbertoabram:
When GPI needs to log something, both the original code that invoked a GPI method and the logging are in python, but they are separated by the GPI, which will eat up any exceptions that would be present when returning from `_log_from_c`. I'm not sure if there are other parts which involve `python -> GPI/C++ -> python`, in which case this would apply to that too, not exclusively logging.

This draft shows one possible way to pass the exceptions through, if enabled via environment variable. Just a proof of concept for discussion, as this would require adding `FORWARD_EXCEPTION` to all the methods.

The downsides are the added code for all GPI methods, and the fact that it doesn't work properly if there is a possibility that multiple exceptions get raised. The latter requires detecting if an exception occurred at every call into python from C++, and return immediately all the way back to the entry point to return `NULL`, which is doable but a lot more involved.

### DIFF
--- a/src/cocotb/share/lib/pygpi/bind.cpp
+++ b/src/cocotb/share/lib/pygpi/bind.cpp
@@ -24,6 +24,13 @@
 
 #define MODULE_NAME "simulator"
 
+#define FORWARD_EXCEPTION       \
+    do {                        \
+        if (PyErr_Occurred()) { \
+            return NULL;        \
+        }                       \
+    } while (0)
+
 // callback user data
 struct PythonCallback {
     PythonCallback(PyObject *func, PyObject *_args, PyObject *_kwargs)
@@ -521,6 +528,8 @@ static PyObject *set_signal_val_binstr(gpi_hdl_Object<gpi_sim_hdl> *self,
     }
 
     gpi_set_signal_value_binstr(self->hdl, binstr, action);
+
+    FORWARD_EXCEPTION;
     Py_RETURN_NONE;
 }
 
--- a/src/cocotb/share/lib/pygpi/bind.cpp
+++ b/src/cocotb/share/lib/pygpi/bind.cpp
@@ -521,7 +521,6 @@ static PyObject *set_signal_val_binstr(gpi_hdl_Object<gpi_sim_hdl> *self,
     }
 
     gpi_set_signal_value_binstr(self->hdl, binstr, action);
-
     Py_RETURN_NONE;
 }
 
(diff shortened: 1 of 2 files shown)


src/cocotb/share/lib/pygpi/logging.cpp
```diff
@@ -202,6 +207,9 @@ void pygpi_logging_initialize() {
     // Default to using the fallback handler until configured
     gpi_get_log_handler(&fallback_log_handler, &fallback_log_userdata);
     gpi_set_log_handler(pygpi_log_handler, nullptr);
+
+    const char *val = getenv("GPI_LOG_EXCEPTION_PASSTHROUGH");
+    exception_passthrough = val && !(strcmp(val, "0") == 0);
```
ktbarrett: I don't think this needs to be optional. Exceptions should be propagated up always. It's just a PITA to do with manually written C API code which is why none of it has been done up to now.

src/cocotb/share/lib/pygpi/bind.cpp
```diff
@@ -1102,6 +1103,20 @@ static int add_module_types(PyObject *simulator) {
...
+// that are not handled explicitly. This really only does anything for
+// functions which can call back into Python and don't handle exceptions
+// explicitly, but this is the case for most as logging does that.
+#define WRAP(method)                                    \
+    [](PyObject *obj, PyObject *args) -> PyObject * {   \
+        PyObject *ret = PyCFunction(method)(obj, args); \
+        if (ret && PyErr_Occurred()) {                  \
+            Py_DECREF(ret);                             \
+            return NULL;                                \
+        }                                               \
```
eric-wieser, line 1117: I think it's usually considered a SystemError if a function both sets an error and returns a value; do we have to opt in to get those diagnostics?

gilbertoabram: You are right about that. I got back to this after a few months and lost some cached mental state, and that comment does not make sense. I think the problem I might have been seeing was that the exception raised in the logger got "lost" in favor of the SystemError synthesized at https://github.com/python/cpython/blob/e12ee02a164c87865b78f6d5a511f6009871383b/Objects/call.c#L49.
I'm going to re-test the original issue and confirm, just in case I missed something and all I need is for PyErr_Print to not "eat up" the exceptions in the other file, and if not I'll fix the comment.

gilbertoabram: I clarified the comment a bit. The problem was indeed that: after allowing logging exceptions to propagate (in this PR) they would get wrapped in a SystemError

ktbarrett (approved):

src/cocotb/share/lib/pygpi/bind.cpp
eric-wieser: This still looks wrong to me; I think if this is a concern, then each of these functions should be performing this check internally. In general it is not safe to continue running python code once an exception is set.

ktbarrett: It's kind of a pain to do this the "right way" because of how cocotb still is.

Any log can throw a Python exception. Really the libraries should bubble failures up to the Python boundary and we can just `return NULL`. But we are not currently handling the possibility of log failures and adding manual checks (which are extremely common due to being a C library originally) are a pain. Exceptions would work but then we would be throwing across library boundaries due to how the libraries are organized still.

This works for now (based on testing), and more meaningful refactors will *eventually* happen: merging libraries, using more exceptions, handling this all automatically by leveraging nanobind instead of touching C API directly. We can't expect that from this PR.

PR #5440, opened 2026-03-23
Fix use-after-free in VpiSignalObjHdl::register_value_change_callback

### LINKED ISSUE #5439: Use-after-free in VpiSignalObjHdl::register_value_change_callback error path
## Bug Description

In `src/cocotb/share/lib/vpi/VpiSignal.cpp`, the error path in `register_value_change_callback()` has a use-after-free bug — it deletes `this` (the signal object) instead of `cb_hdl` (the failed callback):

```cpp
GpiCbHdl *VpiSignalObjHdl::register_value_change_callback(
    gpi_edge edge, int (*cb_func)(void *), void *cb_data) {
    VpiValueCbHdl *cb_hdl = new VpiValueCbHdl(this->m_impl, this, edge);
    if (cb_hdl->arm()) {
        delete this;     // BUG: should be `delete cb_hdl`
        return NULL;
    }
    cb_hdl->set_cb_info(cb_func, cb_data);
    return cb_hdl;
}
```

When `arm()` fails (i.e. `vpi_register_cb()` returns NULL), `delete this` destroys the `VpiSignalObjHdl` — but Python still holds a reference to it. When cocotb then tries to format the error message by calling `repr()` on the trigger, it invokes `gpi_get_definition_name()` on the freed signal memory, causing a segfault.

Additionally, the allocated `cb_hdl` is leaked since it is never freed.

## How to Reproduce

1. Create a SystemVerilog module with unpacked array outputs:
   ```systemverilog
   module top (output logic csn_o [12]);
     // ...
   endmodule
   ```
2. In cocotb, access an element and register an edge trigger:
   ```python
   signal = dut.csn_o[0]
   await FallingEdge(signal)
   ```
3. Run on VCS **without** `-debug_access+r` — VCS rejects the `cbValueChange` VPI callback on unpacked array elements, causing `vpi_register_cb()` to return NULL and `arm()` to fail.
4. Instead of a clean error message, the simulator segfaults.

## Expected Behaviour

A clean error message indicating the callback registration failed, without crashing.

## Proposed Fix

Line 235: change `delete this` to `delete cb_hdl`:

```diff
     if (cb_hdl->arm()) {
-        delete this;
+        delete cb_hdl;
         return NULL;
     }
```

This correctly cleans up the failed callback allocation while leaving the signal object intact so Python can still access it for error reporting.

## Environment

- cocotb version: 2.0.1
- Simulator: VCS (Synopsys VCS X-2025.06)
- OS: Rocky Linux 8
- Trigger: `cbValueChange` registration failure on unpacked SV array elements without `-debug_access+r`

## Notes

The `delete this` pattern appears intentionally in other places in `VpiCbHdl::remove()` and `VpiCbHdl::run()` where it makes sense (callback objects deleting themselves after firing). This specific instance in `register_value_change_callback` looks like a typo — `this` is the **signal** (`VpiSignalObjHdl`), not a callback.

svet-h:
## Summary

Fixes a use-after-free bug in the `cbValueChange` registration error path that causes a segfault instead of a clean error message.

In `VpiSignalObjHdl::register_value_change_callback()`, when `arm()` fails (e.g. VCS rejects `cbValueChange` on unpacked SV array elements), the error path executes `delete this` — destroying the **signal object** instead of `cb_hdl` (the failed callback). Python still holds a reference to the signal, so subsequent access (e.g. `repr()` for the error message) segfaults.

The fix: `delete this` → `delete cb_hdl`.

This also fixes a memory leak — the allocated `VpiValueCbHdl` was never freed on the error path.

Fixes #5439


ktbarrett (approved):
Is this the only place where this mistake was made?

svet-h:
> Is this the only place where this mistake was made?

the only one i could find

ktbarrett:
Have a look at `docs/source/newsfragments/README.md` to create a newsfragment for this bugfix and we should be good to merge.

svet-h:
Updated with newsfragment + unit test

tests/test_cases/test_cocotb/test_edge_triggers.py
```diff
@@ -472,3 +474,18 @@ async def issue_376_different_edges(dut):
...
+    should raise RuntimeError, not crash."""
+    trigger = dut.clk.rising_edge
+
+    with unittest.mock.patch.object(
+        simulator,
+        "register_value_change_callback",
+        return_value=None,
+    ):
+        with pytest.raises(RuntimeError, match="Unable set up .* Trigger"):
+            trigger._prime()
```
ktbarrett: These Triggers have global lifetime, just calling `_prime` may put this in an invalid state which will affect any test that follows.
```suggestion
            await trigger
```

docs/source/newsfragments/5439.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed use-after-free in ``VpiSignalObjHdl::register_value_change_callback`` where the signal object was incorrectly deleted instead of the callback handle when callback registration failed.
```
ktbarrett: Mention observable behavior. Don't mention detail someone perusing the changelog won't care about.

```suggestion
Fixed use-after-free in ``VpiSignalObjHdl::register_value_change_callback`` error handling code which caused VCS to segfault.
```

ktbarrett (approved):
Looks good. Left a couple notes.

ktbarrett:
Well Github Action keeps dying, so maybe tomorrow.

svet-h:
oh well, maybe worth another go? :D

ktbarrett:
I'm starting to think it's this change causing the CI issues. The VHDL jobs aren't failing.

svet-h:
I'll investigate tomorrow morning. Could be the extra dummy test I added

tests/test_cases/test_cocotb/test_edge_triggers.py
```diff
@@ -472,3 +474,18 @@ async def issue_376_different_edges(dut):
...
+    should raise RuntimeError, not crash."""
+    trigger = dut.clk.rising_edge
+
+    with unittest.mock.patch.object(
+        simulator,
+        "register_value_change_callback",
+        return_value=None,
+    ):
+        with pytest.raises(RuntimeError, match="Unable set up .* Trigger"):
+            await trigger
```
ktbarrett, line 493: If the problem was my recommendation to change this, change it to use `cocotb.simulator` directly.

ktbarrett:
Sorry this is turning into more than a simple fix. This codebase can be a pain sometimes. But I'm glad we added a test and can convince ourselves it works.

svet-h:
I think i got it, can we rerun CI? Apologies for the delay..

ktbarrett (approved):

ktbarrett:
Thanks!

PR #5448, opened 2026-03-26
Use setuptools-git-versioning for generating dev versions

### LINKED ISSUE #5447: Release tests broken
Caused by #5432.

The setup.py modifies the `__version__` to include the git revision. When building the sdist this version is included in the PKG-INFO and so is the `setup.py`. But when building the project when installing the sdist it's done in an isolated build environment where there is no git history and the git rev can't be included. This causes the package installer to complain since the packaged sdist version and attempted build version are not the same.

We could revert it, or we could move to a versioning system such as https://github.com/dolfinus/setuptools-git-versioning. We used setuptools-scm at one point but were unhappy with it doing more than just computing the package version and being overly opinionated about what the version actually was. Perhaps this other project will do better.

ktbarrett:
Closes #5447. This PR adds `setuptools-git-versioning` to the build process to automatically create appropriate package versions using git information.

This library is less opinionated and more configurable than setuptools-scm and *only* deals with building the release number, ignoring the whole "all checked-in items should go in the sdist" thing setuptools-scm did.

Also nice is it can detect dirty git trees and add a `.dirty` to the version number.

## How it works

There is now a VERSION file checked into the repo that describes the next released version. The pyproject.toml has configurations under `[tool.setuptools-git-versioning]`.

### Release

When we release, we will tag the commit with the same value as the `{tag}` (the value in the VERSION file). This will cause to tool to determine we are on the release commit and use the `template` template to build the package version. This defaults to just `{tag}`.

So if VERSION contains `2.1` and we tag a commit `2.1`, the generated version will be `2.1`.

After release, we bump the VERSION file to `2.2`.

### Development with full git history

The tool will detect we are not release (due to absence of the release tag) and will use the `dev_template` to build the package version. The template is `"{tag}.dev{ccount}+r{sha}"`. The `{tag}` is the value from the VERSION file (`2.1`). The `{ccount}` is the number of commits since the VERSION file was last modified. The `{sha}` is the git SHA.

To use `{ccount}` we have to have the `count_commits_from_version_file = true` config set.

The resulting version will be something like `2.1.dev123+ra47810cd`.

### Development with shallow clone

The same logic as above applies, but the number of commits since the VERSION file changed will be `0` since the current commit is when it was created.

The resulting version will be something like `2.1.dev0+ra47810cd`.

### Development with dirty repo

The same rules for development from above will apply, but the `dirty_template` will be used, which will add a `.dirty` to the git SHA local version field.

### Installing without git information

Currently this falls back to the release template, so the version will be `2.1`.

## Follow on work

For the no git info case we could improve it. I opened https://github.com/dolfinus/setuptools-git-versioning/issues/131.
The solution proposed is to add a templated file that's filled in when someone calls `git archive`, such as the `Download as ZIP` button in Github does.

Since we are trying to move to scikit-build-core, we have to consider supporting that as well. I opened https://github.com/dolfinus/setuptools-git-versioning/issues/130. This is actually quite easy to implement as scikit-build-core has a `version-provider` config option and an example on how to use it for setuptools-scm internally that's quite simple and easy to adapt to setuptools-git-versioning.


imphil (approved):
It feels a bit odd to me that `VERSION` is the upcoming (not yet released) version, (which isn't even known at this point, could end up as 3.0 instead of 2.1) instead of the current version, or explicitly a dev version. But I guess technically looking at a repository between releases will always give you something more or less undefined.

We had some chicken-and-egg problems in the past between doing final tests on a release and tagging it, I *think* we're OK with the approach in this PR, let's see with the next release!

So let's see how far that gets us. It all feels like a lot of dependencies and magic for just including the git version in a version number, but it unbreaks our release CI, so that alone is worth it!

PR #5456, opened 2026-03-28
Document how to exit the process when in post_mortem

### LINKED ISSUE #4973: quiting the `pdb` debugger started with `COCOTB_PDB_ON_EXCEPTION` keep running
We are using `pdb.post_mortem` to do post-mortems on exceptions that fail tests. One interesting behavior of `post_mortem` vs `set_trace` is that quiting the debugger with `q` does not raise `BdbQuit` like it does in `set_trace`. This causes the simulation to continue executing.

Maybe we need a separate function to request regression exit.

ktbarrett:
Closes #4973. `sys.exit` or `raise SystemExit` does not work in pbd sessions, only forceful termination. Glad they fixed this finally.

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -988,6 +988,15 @@ Debugging
     If defined, cocotb will drop into the Python debugger (:mod:`pdb`) if a test fails with an exception.
     See also the :ref:`troubleshooting-attaching-debugger-python` subsection of :ref:`troubleshooting-attaching-debugger`.
 
+    .. note::
+        Prior to Python 3.14 running the ``(q)uit`` command in the debugger would not exit the simulator process,
+        but continue running the rest of the test suite. From Python 3.14 onwards, ``(q)uit`` will exit the simulator process immediately.
+        To mimic this behavior call :func:`os._exit` from the pdb command line.
+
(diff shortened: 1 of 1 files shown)


docs/source/library_reference.rst
```diff
@@ -988,6 +988,15 @@ Debugging
     If defined, cocotb will drop into the Python debugger (:mod:`pdb`) if a test fails with an exception.
     See also the :ref:`troubleshooting-attaching-debugger-python` subsection of :ref:`troubleshooting-attaching-debugger`.
 
+    .. note::
+        Prior to Python 3.14 running the ``(q)uit`` command in the debugger would not exit the simulator process,
+        but continue running the rest of the test suite. From Python 3.14 onwards, ``(q)uit`` will exit the simulator process immediately.
```
cmarqu, line 993:
```suggestion
        but continue running for the rest of the test suite. From Python 3.14 onwards, ``(q)uit`` will exit the simulator process immediately.
```
?

ktbarrett: That reads incorrectly. Its the regression that continues running rather than shutting down the simulation, not pdb. Quitting and continue are equivalent in post_mortem up to 3.14.

PR #5473, opened 2026-04-03
docs: recommend coverage subprocess measurement as alternative

### LINKED ISSUE #5449: Recommend alternative for coverage collection in docs
The `coverage` module supports recording coverage in Python files loaded by subprocesses via a [sitecustomize.py](https://coverage.readthedocs.io/en/6.5.0/subprocess.html) which will record coverage for all packages in that site and can be turned on and off with an environment variable (which also points to a configuration file). 

This is a set-it-and-forget-it approach that will work well for users that regularly do coverage in their environments.

mvanhorn:
## Summary

Adds a section to the library reference documenting the `coverage` module's subprocess measurement via `sitecustomize.py` as an alternative to `COCOTB_USER_COVERAGE`.

## Why this matters

The `coverage` subprocess support automatically records coverage for all Python subprocesses, including those spawned by the simulator. This is useful for users who already use `coverage` across their test infrastructure and want a single configuration rather than cocotb-specific env vars.

## Changes

- `docs/source/library_reference.rst`: Added a new subsection "Alternative: using `coverage` subprocess support" after the `COVERAGE_RCFILE` entry, with setup steps and example configuration.

All flags and configuration options referenced were verified against the [coverage subprocess docs](https://coverage.readthedocs.io/en/latest/subprocess.html).

## Testing

- Verified RST syntax renders correctly
- Cross-checked `COVERAGE_PROCESS_START` and `--parallel-mode` against coverage module docs

Closes #5449

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -1077,14 +1077,33 @@ To set it up:
       branch = True
       source = my_package
 
-3. Run ``coverage run`` with the ``--parallel-mode`` flag and the
-   ``COVERAGE_PROCESS_START`` environment variable pointing to the
-   configuration file:
+3. Set the ``COVERAGE_PROCESS_START`` environment variable and run your
+   tests. The exact command depends on how you run cocotb:
+
+   **With Makefiles:**
+
+   .. code-block:: bash
+
+      export COVERAGE_PROCESS_START=.coveragerc
+      make SIM=<simulator>
+
+   **With the Python Runner directly:**
+
+   .. code-block:: bash
+
+      export COVERAGE_PROCESS_START=.coveragerc
+      python my_test_script.py
+
+   **With pytest:**
 
    .. code-block:: bash
 
       export COVERAGE_PROCESS_START=.coveragerc
       coverage run --parallel-mode -m pytest
+
+   In all cases, the ``COVERAGE_PROCESS_START`` variable causes
+   :mod:`coverage` to start measurement in simulator-spawned Python
+   subprocesses automatically via its :mod:`sitecustomize` hook.
 
 4. Combine and view the results:
 


docs/source/library_reference.rst
```diff
@@ -1052,6 +1052,51 @@ Debugging
...
+      source = my_package
+
+3. Run ``coverage run`` with the ``--parallel-mode`` flag and the
+   ``COVERAGE_PROCESS_START`` environment variable pointing to the
+   configuration file:
+
+   .. code-block:: bash
+
+      export COVERAGE_PROCESS_START=.coveragerc
+      coverage run --parallel-mode -m pytest
```
ktbarrett, line 1102: This only works for users using the pytest plugin (which we are *not* recommending at this time) or pytest with the Python Runners. Update this to provide a flow for people using the Makefiles and the Python Runners directly.

docs/source/library_reference.rst
```diff
@@ -1052,6 +1052,70 @@ Debugging
...
+
+3. Set the ``COVERAGE_PROCESS_START`` environment variable and run your
+   tests. The exact command depends on how you run cocotb:
+
+   **With Makefiles:**
+
+   .. code-block:: bash
+
+      export COVERAGE_PROCESS_START=.coveragerc
+      make SIM=<simulator>
```
ktbarrett, line 1088: Ideally they would have this set in their Makefiles.

```suggestion
      make
```

docs/source/library_reference.rst
```diff
@@ -1052,6 +1052,70 @@ Debugging
...
+   .. code-block:: ini
+
+      [run]
+      branch = True
+      source = my_package
+
+3. Set the ``COVERAGE_PROCESS_START`` environment variable and run your
+   tests. The exact command depends on how you run cocotb:
+
+   **With Makefiles:**
```
ktbarrett, line 1083: This looks good, just needs to be reorganized into `tab-set` and `tab-item`. There's an example in install.rst.

docs/source/library_reference.rst
mvanhorn: Good point. Updated in 85b61e7 to show three flows: Makefiles, Python Runner directly, and pytest. The key in all cases is just setting `COVERAGE_PROCESS_START` so the sitecustomize hook fires in simulator-spawned subprocesses.

ktbarrett:
Moved and cleaned up your docs and merged as part of #5507. Thanks!

PR #5475, opened 2026-04-03
Adjust treatment of packed objects

ktbarrett:
Support opportunistic indexing into packed arrays or packed structs.


ZiaCheemaGit:
opportunistic indexing?

ktbarrett:
@ZiaCheemaGit It's kind of a saga that you got a bit caught up in, sorry.

Basically before 2.0 we supported indexing into packed arrays and getting fields of packed structs, things that the language reference manual says you can't do. In some of those cases you could get handles but they wouldn't work correctly. So in 2.0 we decided to remove support for indexing into packed objects and told users to handle it in other ways. But that upset some people where it *did* work for them (e.g. Verilator). But worse yet, some simulators misbehave in other ways and let you index into packed objects in ways we can't prevent. It's a huge mess.

So recently we decided to revert that decision and go back to letting people index into packed objects if it works ("opportunistic support") with the ultimate goal of making packed object indexing work across all simulators. The only difference is we will be a bit more judicious about testing to prevent us from accidentally supporting things that don't actually work, like before.

Issue #5474, opened 2026-04-03
Nix: Running Tests with Verilator via Runners Is Broken Due to Direct Call to Perl


yanghuaxuan:
While using Nix for a little project I'm doing, I noticed that using Pytest runners to run simulations with Verilator mysteriously failed. 

# Steps to Replicate
To replicate this issue, I created a simple [repository](https://github.com/yanghuaxuan/cocotb_demo) with a dff in Verilog alongside a Nix shell for packaging Python and simulators. The Nix shell's nixpkgs is pinned to [6e48b28](https://github.com/NixOS/nixpkgs/commit/6e48b2865e9f5cc0c95abae887968df7881359a9).  I deliberately did not use Nix to package in Cocotb because that has its own set of [issues](https://github.com/NixOS/nixpkgs/issues/385137td) on aarch64-(linux/darwin) ☹️ 

The issue can be replicated by cloning and running the following commands:

1. Source nix-shell
```
nix-shell
```
2. Create venv + source
```
python3 -m venv .venv
source .venv/bin/activate
```
3. Install cocotb + pytest
```
pip install -r requirements.txt
```
4. Run Pytest runner with `SIM=verilator`
```
SIM=verilator pytest test_dff.py -s

```

Looking around a bit, I noticed that the issue is due to the runner directly calling `perl` in order to call Verilator, which is a problem on Nix as it'll directly all the system packaged perl rather than using the shebang to call the Nix provided perl.
https://github.com/cocotb/cocotb/blob/7d090f4fe421dc67de362e054e3992da0677c6c9/src/cocotb_tools/runner.py#L1625-L1658
This became a problem on my system:
```
============================= test session starts ==============================
platform linux -- Python 3.13.12, pytest-9.0.2, pluggy-1.6.0
rootdir: /home/hyang/projects/cocotb-verilator-replicate
collected 1 item

test_dff.py Can't locate FindBin.pm in @INC (you may need to install the FindBin module) (@INC entries checked: /usr/local/lib64/perl5/5.42 /usr/local/share/perl5/5.42 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5) at /nix/store/x0b9wzh3k048lq3if4k3c0h6d4d8892v-verilator-5.044/bin/verilator line 15.
[...]

PR #5476, opened 2026-04-03
Direct run verilator rather than call perl

yanghuaxuan:
Closes issue #5474


ktbarrett (approved):

ktbarrett:
Just in case there was some reason for this. @themperek Can you opine?

themperek (approved):

PR #5478, opened 2026-04-04
Fix ecosystem tests

ktbarrett:
Depends on #5509. These have been broken for a while. We had to update cocotb-coverage since they just added support for cocotb 2.0+ and the pyuvm one was kinda stale, I think most people have updated to 4.0 since it's deliberately 2.0 compatible.

The new flow for all tests is:
1. install tools and other testing requirements
2. clone a release version of the library under test (LUT)
3. pip install the local clone of the release version of the LUT
4. install cocotb master from Github
5. run the LUT's tests manually to allow for cocotb version override

### DIFF
--- a/.github/workflows/ecosystem-compat.yml
+++ b/.github/workflows/ecosystem-compat.yml
@@ -22,8 +22,12 @@ on:
 jobs:
   cocotb-coverage:
     if: github.repository == 'cocotb/cocotb'
-    name: Test cocotb-coverage 1.2
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


ktbarrett:
Looks like random sequence shifted and we aren't hitting coverage in the cocotb-coverage test. I wonder if it's transient. They should probably be using a consistent seed if not.

ktbarrett:
I opened https://github.com/mciepluc/cocotb-coverage/pull/94 to fix the transient cocotb-coverage failure.

PR #5480, opened 2026-04-05
Add ability to randomize test order

### LINKED ISSUE #5106: Support random test ordering
I'm thinking about adding a feature to run tests in a random order.

Currently one can get halfway there by doing something like `@cocotb.test(stage=random.randint(0, 100))`, but in case you have a `@cocotb.test` along with a `@cocotb.parametrize`, the stage gets passed down to the parametrized tests, so all the generated parametrized tests run in the same order.

My idea is to add a `randomize` field in the `Test` class. All tests will be in run in stage order. Within a stage, any tests with `randomize=True` will be shuffled and interspersed. Tests with `randomize=False` will maintain the same relative ordering.

Let me know if this sounds okay. I'm not 100% sure where to add tests for this though, so guidance would be appreciated.

chaseruskin:
Creates a new environment variable (COCOTB_RANDOM_TEST_ORDER) to enable shuffling of tests within the regression manager (closes #5106).

This change provides a new capability without breaking existing functionality by preserving the original test ordering by default. Users can set a new environment variable, `COCOTB_RANDOM_TEST_ORDER`, to enable this new capability. The randomized ordering of stages is beneficial to cocotb users because without this capability, users did not have a way to try to verify that state was not being preserved between tests, or that some state in the previous test was affecting a future test. By randomizing test order, a cocotb user can run regression tests and have a higher chance of identifying if any tests are relying on past state during a simulation.

This capability also relies on the `cocotb.RANDOM_SEED`, in which it will reproduce the same ordering if a random seed is set. This is useful to ensure reproducibility if a certain ordering yielded failed tests.

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -960,6 +960,19 @@ Other Runtime Information
 
 .. autodata:: cocotb.is_simulation
 
+.. envvar:: COCOTB_RANDOM_TEST_ORDER
+
+    Type :ref:`env-boolean`
+
+    Default: :data:`False`
+
+    Enable randomizing the order of tests within each stage. To recreate the
+    same test order, the random seed must be set.
+
+    If not set, all tests are ran in the order in which they were discovered.
+
+    .. versionadded:: 2.1
+
 Debugging
 ---------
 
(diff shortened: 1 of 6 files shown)


ktbarrett (approved):
The change looks good. Unfortunately the private CI is having issues. I've annoyed the right people and after it's fixed we will get this in.

chaseruskin:
Thank you!

ktbarrett:
Thanks!

PR #5483, opened 2026-04-05
Support Edge triggers in 1-bit LogicArray

### LINKED ISSUE #5482: Support Edge triggers in 1-bit LogicArray
I'm currently facing an issue in a specific case where I need to define all my VHDL signals as `std_logic_vector`, including 1-bit signals. The issue is that, since all the signals are `std_logic_vector`, even clocks, I can't do any `await Edge` or `ClockCycles`. Whenever I try to use them, I get:

```
TypeError: RisingEdge requires a scalar LogicObject. Got LogicArrayObject(top.clk) of type LogicArrayObject
```

 This issue started when I switched from 1.9.2 to 2.0.1, it seems it was previously supported.

I have been able to patch this with the following changes:

```diff
diff --git a/src/cocotb/_gpi_triggers.py b/src/cocotb/_gpi_triggers.py
index 46508cf2..bb94dbf0 100644
--- a/src/cocotb/_gpi_triggers.py
+++ b/src/cocotb/_gpi_triggers.py
@@ -286,7 +286,7 @@ class RisingEdge(_EdgeBase["cocotb.handle.LogicObject"]):
     _edge_type = simulator.RISING

     def __new__(cls, signal: cocotb.handle.LogicObject) -> RisingEdge:
-        if not (isinstance(signal, cocotb.handle.LogicObject)):
+        if not isinstance(signal, (cocotb.handle.LogicObject, cocotb.handle.LogicArrayObject)):
             raise TypeError(
                 f"{cls.__qualname__} requires a scalar LogicObject. Got {signal!r} of type {type(signal).__qualname__}"
             )
@@ -315,7 +315,7 @@ class FallingEdge(_EdgeBase["cocotb.handle.LogicObject"]):
     _edge_type = simulator.FALLING

     def __new__(cls, signal: cocotb.handle.LogicObject) -> FallingEdge:
-        if not (isinstance(signal, cocotb.handle.LogicObject)):
+        if not isinstance(signal, (cocotb.handle.LogicObject, cocotb.handle.LogicArrayObject)):
             raise TypeError(
                 f"{cls.__qualname__} requires a scalar LogicObject. Got {signal!r} of type {type(signal).__qualname__}"
             )
diff --git a/src/cocotb/handle.py b/src/cocotb/handle.py
index 5db38b7b..62feb454 100644
--- a/src/cocotb/handle.py
+++ b/src/cocotb/handle.py
@@ -1394,6 +1394,23 @@ class LogicArrayObject(
         # Backwards compatibility. Always wrap negative values.
         return (2 ** len(self)) - 1

+    @cached_property
+    def rising_edge(self) -> RisingEdge:
+        """A trigger which fires whenever the value changes to a ``1``."""
+        if len(self) != 1:
+            raise TypeError(f"Can't get RisingEdge on {len(self)}-bit signal")
+        if self.is_const:
+            raise TypeError("Can't get RisingEdge on immutable signal")
+        return RisingEdge._make(self)
+
+    @cached_property
+    def falling_edge(self) -> FallingEdge:
+        """A trigger which fires whenever the value changes to a ``0``."""
+        if len(self) != 1:
+            raise TypeError(f"Can't get FallingEdge on {len(self)}-bit signal")
+        if self.is_const:
+            raise TypeError("Can't get FallingEdge on immutable signal")
+        return FallingEdge._make(self)

 class PackedObject(LogicArrayObject):
     """A packed Verilog struct, union, or vector simulation object.
```

I didn't make a fork + PR because I'm not sure if this is the correct way to do it or to integrate it within the framework, but I'm happy to do it if required.

Please let me know if this would be feasible to add, or if you have any alternative suggestions to work around this issue.

Thanks in advance.

AlanVek:
This PR is aimed to fix #5482. Adding support for rising_edge/falling_edge on 1-bit LogicArrayObject so that mainly VHDL signals defined as `std_(u)logic_vector(0 downto 0)` can still be used with `await RisingEdge/FallingEdge`.

Closes #5482

### DIFF
--- a/docs/source/newsfragments/5483.feature.rst
+++ b/docs/source/newsfragments/5483.feature.rst
@@ -1,1 +1,1 @@
-Added support for :class:`~cocotb.trigger.RisingEdge` and :class:`~cocotb.triggers.FallingEdge` for instances of :class:`~cocotb.handle.LogicArrayObject` with only one bit.
+Added support for :class:`~cocotb.triggers.RisingEdge` and :class:`~cocotb.triggers.FallingEdge` for instances of :class:`~cocotb.handle.LogicArrayObject` with only one bit.


ktbarrett:
Sweet, thanks for the changes. One more thing that's needed is a newsfragment describing the change (well "feature" since we did a release that removed support). You can find directions in `docs/source/newsfragments/README.rst`.

Also, don't be alarmed if the commercial simulators fail, there's a known issue with our CI that will hopefully be resolved in the coming days.

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -458,3 +458,27 @@ async def test_handle_str_with_separators(dut: Any) -> None:
...
+    with pytest.raises(TypeError):
+        # RisingEdge on multi-bit signal
+        await cocotb.triggers.RisingEdge(dut.stream_in_data)
+    with pytest.raises(TypeError):
+        # FallingEdge on multi-bit signal
+        await cocotb.triggers.FallingEdge(dut.stream_in_data)
+
+    assert len(dut.one_bit_vector) == 1
+
+    cocotb.start_soon(cocotb.clock.Clock(dut.one_bit_vector, 10, "ns"))
```
ktbarrett:
```suggestion
    cocotb.clock.Clock(dut.one_bit_vector, 10, "ns").start()
```

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -458,3 +458,27 @@ async def test_handle_str_with_separators(dut: Any) -> None:
     dut.stream_in_data.value = "__01010011__"
     await Timer(1, "ns")
     assert dut.stream_in_data.value == LogicArray("01010011")
+
+
+@cocotb.skipif(
+    LANGUAGE != "vhdl", reason="Testing LogicArrayObject edges only for VHDL"
+)
```
ktbarrett: Why do we need to limit this to VHDL?

tests/test_cases/test_cocotb/test_handle.py
AlanVek: My bad, thanks

tests/test_cases/test_cocotb/test_handle.py
AlanVek: I wasn't able to map a Verilog type to LogicArrayObject or PackedObject for 1-bit signal.

If I do:

```verilog
reg [0:0] one_bit_vector;
```

I get a LogicObject. The test will pass, but we're not really validating the new feature. Any suggestions on how to force this to LogicArrayObject?

ktbarrett: Ohhhh, right. The VPI is kinda weird. It's okay if this is incremental.

AlanVek: Should I enable it for Verilog anyways? It wouldn't really be validating the changes, but the test still passes

ktbarrett: Sure. I guess it'd be good to ensure that things are future proof. We do intend to fix that at some point.

AlanVek: Done, thanks

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -458,3 +458,24 @@ async def test_handle_str_with_separators(dut: Any) -> None:
     dut.stream_in_data.value = "__01010011__"
     await Timer(1, "ns")
     assert dut.stream_in_data.value == LogicArray("01010011")
+
+
+@cocotb.test
+async def test_edge_on_vectors(dut: Any) -> None:
```
ktbarrett, line 469:
```suggestion
@cocotb.skipif(
    SIM_NAME.startswith("modelsim") and LANGUAGE == "verilog",
    reason="Questa VPI hangs trying to wait on the edge trigger of a vector",
)
async def test_edge_on_vectors(dut: Any) -> None:
```

AlanVek: Another solution would be to revert the changes that allow `_EdgeBase._make` to receive a LogicArrayObject, and instead changing `LogicArrayObject.rising_edge` to this:

```diff
@@ -1401,7 +1401,7 @@ class LogicArrayObject(
             raise TypeError(f"Can't get RisingEdge on {len(self)}-bit signal")
         if self.is_const:
             raise TypeError("Can't get RisingEdge on immutable signal")
-        return RisingEdge._make(self)
+        return RisingEdge._make(self[0])
```

Same for falling_edge. I think that should fix the simulation issue and be more generally compatible, right?

EDIT: Nevermind, this is failing on some simulators when signals map to PackedObject (which doesn't support indexing) instead of LogicArrayObject.

ktbarrett: Welcome to the fun!

Thinking about this makes me wonder how we can ever make this work in Questa. I'll need to think about how we can make this work before making this functionality available.

AlanVek: How was this working before (1.9.2)? Or was this always broken in Questa?

ktbarrett: It was probably always broken, we just didn't have a test for it. This is par for the course when it comes to simulator support for the VPI and VHPI. That's why our future handle discovery refactor will try to remove the simulators from the equation as much as possible.

I'm checking to see what I can do with your PR.

tests/test_cases/test_cocotb/test_handle.py
```diff
@@ -458,3 +458,28 @@ async def test_handle_str_with_separators(dut: Any) -> None:
     dut.stream_in_data.value = "__01010011__"
     await Timer(1, "ns")
     assert dut.stream_in_data.value == LogicArray("01010011")
+
+
+@cocotb.test
+@cocotb.skipif(
+    SIM_NAME.startswith("modelsim") and LANGUAGE == "verilog",
+    reason="Questa VPI hangs trying to wait on the edge trigger of a vector",
+)
```
ktbarrett, line 468: This is working for me locally. Removing to see what's happening again.
```suggestion
```

ktbarrett:
So for whatever reason this only appears when running with the Python Runners... oh... as soon as I said that out loud I figured it out. The Python Runners haven't been updated to the new qrun flow and for whatever reason that fixes a lot of issues. I'm definitely making that a priority for 2.2. 

I think we can just skip now.

ktbarrett:
Icarus v13 seems to remove undriven signals as an optimization (but Icarus 12 and HEAD don't have this issue). We could just move this to the port list.

AlanVek:
> Icarus v13 seems to remove undriven signals as an optimization (but Icarus 12 and HEAD don't have this issue). We could just move this to the port list.

@ktbarrett  I added the `initial_begin` and the `$display` like what's used in `mybits`, so as to keep the ports clean. Let me know if this is fine, otherwise I can add it to the ports as you suggested.

It's working locally with Icarus 13 (and verified that it breaks when I leave it undriven).

ktbarrett:
Oh you know that might be because of my own debugging. I thought maybe it was optimizing it to a constant since there's no HDL drivers of the signal. But if it works otherwise yeah put it back in the initial.

AlanVek:
This was already pushed before my comment, sorry if that wasn't clear. You should be able to approve the pipeline and we can see if it's working

ktbarrett:
LGTM.

docs/source/newsfragments/5483.feature.rst
```diff
@@ -0,0 +1 @@
+Added support for :class:`RisingEdge` and :class:`FallingEdge` for instances of :class:`LogicArrayObject` with only one bit.
```
ktbarrett:
```suggestion
Added support for :class:`~cocotb.trigger.RisingEdge` and :class:`~cocotb.triggers.FallingEdge` for instances of :class:`~cocotb.handle.LogicArrayObject` with only one bit.
```

docs/source/newsfragments/5483.feature.rst
```diff
@@ -0,0 +1 @@
+Added support for :class:`~cocotb.trigger.RisingEdge` and :class:`~cocotb.triggers.FallingEdge` for instances of :class:`~cocotb.handle.LogicArrayObject` with only one bit.
```
ktbarrett:
```suggestion
Added support for :class:`~cocotb.triggers.RisingEdge` and :class:`~cocotb.triggers.FallingEdge` for instances of :class:`~cocotb.handle.LogicArrayObject` with only one bit.
```

ktbarrett:
@AlanVek btw your email is misconfigured in git, so it isn't linking to your Github account.

PR #5497, opened 2026-04-08
Reduce overhead of large regressions

### LINKED ISSUE #5137: Excessive test launch time when generating many testcases (TestFactory / @cocotb.parametrize) in 2.0.0
### Summary

When a single test (for example a test created via TestFactory or decorated with @cocotb.parametrize) generates a very large number of testcases, the test launch time becomes extremely long.

### Reproduction

TestFactory:
```
from cocotb.regression import TestFactory

async def testfactory(dut, a, b, c):
    pass

factory = TestFactory(testfactory)
factory.add_option("a", range(100))
factory.add_option("b", range(100))
factory.add_option("c", range(100))
factory.generate_tests()
```

parametrize:
```
import cocotb

@cocotb.test()
@cocotb.parametrize(a=range(100), b=range(100), c=range(100))
async def parametrize(dut, a, b, c):
    pass
```

### Observed measurements

I ran only one of the tests above and measured the time. 
Comparing cocotb 1.9.2 and 2.0, I observed:

* 1,000 generated testcases: ≈ 2.2 s (1.9.2) → ≈ 2.4 s (2.0)
* 1,000,000 generated testcases: ≈ 18.7 s (1.9.2) → 432 s (2.0 TestFactory) / 503 s (2.0 parametrize)

### Question

Is there any plan to address this large launch-time overhead when a single test produces many testcases?

ktbarrett:
Closes #5137.

In the issue, generating a regression with a million tests and then selecting one regressed from taking a few seconds to practically infinite time between 1.9 and 2.0. This was because of the change to record all excluded functions in the results.xml.

We presumably want those unincluded tests recorded, so I used py-spy and speedscope to quickly identify the main culprits and address them.

Generating a million tests still takes forever, but large combinations like 125k tests are more manageable.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -54,7 +54,7 @@ jobs:
       run: |
         pip install pytest pytest-benchmark
         pip install .
-        pytest -c /dev/null tests/benchmark.py --benchmark-json output.json
+        pytest -c /dev/null tests/benchmarks --benchmark-json output.json
 
     # Pushing the benchmark requires elevated permissions to the
     # cocotb/cocotb-benchmark-results repository, which we only grant for
--- a/src/cocotb/_decorators.py
+++ b/src/cocotb/_decorators.py
@@ -96,6 +96,7 @@ class Test:
         self.expect_error = expect_error
         self.skip = skip
         self.stage = stage
+        self.included = not skip
 
     @property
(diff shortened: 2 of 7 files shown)


ktbarrett:
Not sure I love the performance test. Maybe I should make that an actual performance test.

marlonjames (approved):

PR #5499, opened 2026-04-08
Use qualified imports for cocotb.simulator

ktbarrett:
Fixing some cyclic import issues. We really need to move this module out of the cocotb namespace.

### DIFF
--- a/src/cocotb/_init.py
+++ b/src/cocotb/_init.py
@@ -76,9 +76,9 @@ def _process_plusargs() -> None:
 def _process_packages() -> None:
     pkg_dict = {}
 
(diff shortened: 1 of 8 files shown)


imphil:
Where do you see the cyclic import issues?

imphil (approved):

ktbarrett:
It depends on the exact order of import. If I do

```
import cocotb.triggers
...
```

triggers imports extended_awaitables which imports gpi_triggers which depends upon simtime which does `from cocotb import simulator`, but the import line hasn't finished so the cocotb module isn't finalized yet, leading to a cyclic import error.

PR #5504, opened 2026-04-09
Force git tags in versions to be alphanumeric

### LINKED ISSUE #5467: Version testing breaks with local tests if git rev is a number with leading 0s.
`test_version.test_version` fails when the git tag starts with 0s and the git tag is comprised entirely of numbers.

According to [PEP 440](http://peps.python.org/pep-0440/)

> If a segment consists entirely of ASCII digits then that section should be considered an integer for comparison purposes

I'm guessing this gives packaging tools the liberty to strip leading 0s since "it's just a number". I tested that `packaging.version.Version` does this striping as does `uv pip` and `pip` *after* you feed it the version number with leading 0s. It must break apart the given version value and recombine them.

Either way we need to modify the test to account for this. *Or* we need to always include an alphabet letter to prevent it.

ktbarrett:
Closes #5467.

If a git commit happens to be entirely numeric and the leading characters are 0s they will be stripped leading to odd and inconsistently sized version strings. Adding a non-hex letter forces this field to be alphanumeric and wont be treated as if it were a number.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ if "dev" in __version__:
         rev = subprocess.check_output(
             ["git", "rev-parse", "--short", "HEAD"], universal_newlines=True
         ).strip()
-        __version__ += f"+{rev}"
+        __version__ += f"+r{rev}"
     except Exception:
         pass
 
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,8 @@ if "dev" in __version__:
         rev = subprocess.check_output(
             ["git", "rev-parse", "--short", "HEAD"], universal_newlines=True
         ).strip()
+        # We add an 'r' to ensure this local version field is alphanumeric.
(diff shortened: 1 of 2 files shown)


marlonjames (approved):

setup.py
```diff
@@ -50,7 +50,7 @@ def package_files(directory):
         rev = subprocess.check_output(
             ["git", "rev-parse", "--short", "HEAD"], universal_newlines=True
         ).strip()
-        __version__ += f"+{rev}"
+        __version__ += f"+r{rev}"
```
ktbarrett, line 55:
```suggestion
        # We add an 'r' to ensure this local version field is alphanumeric.
        # Otherwise when a version is a number with leading 0s they will be stripped.
        __version__ += f"+r{rev}"
```

PR #5505, opened 2026-04-10
Add documentation on profiling

### LINKED ISSUE #15: Profiling capabilities
We want to know where all the time is going.

Initially, knowing how much time is spent in the simulator vs. how much time is spent in callback handlers would be a very useful indicator.  This should be easy to do since all callbacks go through the same handler.

ktbarrett:
Closes #15. We now enough a couple different fairly good ways to collect profiling information and now documentation on it.

### DIFF
--- a/docs/source/profiling.rst
+++ b/docs/source/profiling.rst
@@ -1,0 +1,59 @@
+************************
+Profiling Testbench Code
+************************
+
+cocotb provides two ways to profile your testbench code (and cocotb itself).
+One is by using the `built-in tracing profilers in Python <https://docs.python.org/3/library/profile.html>`_,
+and the other is by using a sampling profiler such as `py-spy <https://github.com/benfred/py-spy>`_.
+
+``cProfile`` tracing profiler
+=============================
+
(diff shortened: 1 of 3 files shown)


docs/source/profiling.rst
```diff
@@ -0,0 +1,59 @@
...
+   snakeviz cocotb.prof
+
+
+``py-spy``
+==========
+
+Using a sampling profiler such as ``py-spy`` is not as accurate as a tracing profile,
+but it can be used with much less overhead, which may be preferable for long-running simulations.
+
+To profile with py-spy, prefix the simulation command with py-spy using :envvar:`SIM_CMD_PREFIX`.
```
cmarqu, line 42:
```suggestion
To profile with py-spy, prefix the simulation command with py-spy using :make:var:`SIM_CMD_PREFIX`.
```

cmarqu (approved):

docs/source/profiling.rst
ktbarrett: See #5506.

cmarqu (approved):

PR #5508, opened 2026-04-10
Fix/icarus vpi after python finalize

davekeeshan:
## Summary
Icarus Verilog 13 can deliver VPI callbacks (sync_cb) after gpi_finalize() has called Py_Finalize(). When this happens, handle_gpi_callback() in bind.cpp calls PyGILState_Ensure() on a dead interpreter, causing a segmentation fault.

All cocotb tests pass — the segfault occurs after the regression summary is printed, during simulator shutdown. Because vvp exits non-zero, Make deletes results.xml.

## Reproducing
    cd tests/test_cases/test_icarus_python_shutdown_callback
    SIM=icarus make clean sim
    Output:

    ** TESTS=3 PASS=3 FAIL=0 SKIP=0    12000.00 ...
    make: *** [Makefile.icarus:66: results.xml] Segmentation fault (core dumped)
    make: *** Deleting file 'results.xml'

## Conditions

* Icarus Verilog 13.0 (tested on Ubuntu 24.04, Python 3.12)
* Clock period of 100 ns (100,000 sim steps at 1 ps precision); shorter periods like 10 ns change event alignment at shutdown and avoid the crash
* WAVES=0 (vvp -none)
* Multiple tests in one simulation run

## GDB backtrace
```
#0  0x00007ffff70f6ec9 in ?? () from libpython3.12.so.1.0
#1  0x00007ffff70f96ad in PyGILState_Ensure () from libpython3.12.so.1.0
#2  0x00007ffff7a9e692 in handle_gpi_callback (user_data=...) at bind.cpp:169
#3  0x00007ffff7f08b7c in VpiCbHdl::run () at VpiCbHdl.cpp:138
#4  0x00007ffff7f09459 in handle_vpi_callback_ () at VpiCbHdl.cpp:28
#5  0x00007ffff7f09a77 in handle_vpi_callback () at VpiCbHdl.cpp:58
#6  0x0000555555642cbc in sync_cb::run_run () at vpi_callback.cc:447
#7  0x000055555560ce9e in generic_event_s::run_run () at schedule.cc:511
#8  0x000055555560f59b in schedule_simulate () at schedule.cc:1264
```

## This PR
This PR adds only the failing regression test so the bug is visible in CI. A follow-up commit will add the fix (guarding handle_gpi_callback with Py_IsInitialized()).

### DIFF
--- a/docs/source/newsfragments/5440.bugfix.rst
+++ b/docs/source/newsfragments/5440.bugfix.rst
@@ -1,1 +1,0 @@
-Fixed a segmentation fault at the end of Icarus simulations when the simulator delivered VPI callbacks after the embedded interpreter was finalized (``Py_Finalize``). Added regression test ``test_icarus_python_shutdown_callback``.
--- a/docs/source/newsfragments/5440.bugfix.rst
+++ b/docs/source/newsfragments/5440.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a segmentation fault on Icarus Verilog when VPI callbacks fired after the Python interpreter was finalized during simulator shutdown.
--- a/docs/source/newsfragments/5440.bugfix.rst
+++ b/docs/source/newsfragments/5440.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a segmentation fault at the end of Icarus simulations when the simulator delivered VPI callbacks after the embedded interpreter was finalized (``Py_Finalize``). Added regression test ``test_icarus_python_shutdown_callback``.
--- a/src/cocotb/share/lib/gpi/GpiCommon.cpp
+++ b/src/cocotb/share/lib/gpi/GpiCommon.cpp
@@ -137,6 +137,8 @@ void gpi_check_cleanup(void) {
     }
 }
 
+bool gpi_is_finalizing(void) { return gpi_finalizing; }
+
 static void gpi_load_libs(std::vector<std::string> to_load) {
     std::vector<std::string>::iterator iter;
 
--- a/src/cocotb/share/lib/gpi/fli/FliCbHdl.cpp
+++ b/src/cocotb/share/lib/gpi/fli/FliCbHdl.cpp
@@ -11,6 +11,9 @@
 
 // Main re-entry point for callbacks from simulator
 void handle_fli_callback(void *data) {
+    if (gpi_is_finalizing()) {
+        return;
+    }
     SIM_TO_GPI(FLI, "callback");
 
     // TODO Add why?
--- a/src/cocotb/share/lib/gpi/gpi_priv.hpp
+++ b/src/cocotb/share/lib/gpi/gpi_priv.hpp
@@ -272,6 +272,7 @@ GPI_EXPORT void gpi_end_of_sim_time();
 
 GPI_EXPORT void gpi_entry_point();
 GPI_EXPORT void gpi_check_cleanup();
+GPI_EXPORT bool gpi_is_finalizing();
 GPI_EXPORT void gpi_init_logging_and_debug();
 
 void *utils_dyn_open(const char *lib_name);
--- a/src/cocotb/share/lib/gpi/vhpi/VhpiCbHdl.cpp
+++ b/src/cocotb/share/lib/gpi/vhpi/VhpiCbHdl.cpp
@@ -12,6 +12,9 @@
 
 // Main entry point for callbacks from simulator
 void handle_vhpi_callback(const vhpiCbDataT *cb_data) {
+    if (gpi_is_finalizing()) {
+        return;
+    }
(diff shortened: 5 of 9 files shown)


ktbarrett:
Yes, this is also visible in #5478. The main issue is that Icarus doesn't work properly when you register a callback and deregister it in the same VPI callback, it will fire regardless. It's definitely an upstream issue that needs to be reported. cocotb has the ability to allow callbacks which fail to deregister to simply be ignored when they do fire, but it's conditioned on the `vpi_remove_cb` function returning a failure, which Icarus does not do despite it not being able to deregister the callback.

I'm not sure what to do here in the mean time. Always letting cancelled events fire is bad for performance. We can't leave GPI callback objects in tact while also deregistering because if they successfully deregister and never do fire, we just leak the GPI callback objects (which is what was happening before the callback refactor in 2.0).

For specifically this issue, we could have the `handle_{intf}_callback` check the variables which denote the simulation is shutting down and ignore the callback. But I'm not sure if Icarus will continue to run considering it's calling callbacks after we've called `vpi_control(vpiFinish)` which should terminate the simulator. Icarus is all kinds of broken.

You can ignore the Questa failures, the license queue is broken right now.

ktbarrett:
I would adjust your implementation to instead check for `gpi_finalizing` in `handle_{intf}_callback` in each of the VPI, VHPI, and FLI and ignore the callback.

We clean up all triggers when cocotb is tearing down, so even getting to `handle_gpi_callback` in bind.cpp is a use-after-free, and the delete of the PythonCallback object is a double free.

davekeeshan:
Whoops, I have an over active AI agent, i was supposed to review that work before it pushed it.  However it appears to be basic enough.  See if it is going the way you want.  Writing and debugging C isn't my strong point, I think reporting and suggesting the error is probably enoughm you have a stronger understanding architecturally where or what the fix is.

ktbarrett:
No problem. I'll take a closer look and get this fix in after the Questa license server is fixed.

ktbarrett:
Crap this is actually not the issue I thought it was. I figured the fact that it was the clock and Timer in the test firing at the same time icarus was having trouble killing a callback already queued up in the "fired" queue.

What it really is is the ReadWrite callback is still alive. We never clean up the write scheduler when exiting. This would definitely be solved by TaskContexts but I can hack something together now.

ktbarrett:
Merged #5509 in favor of this one. I moved over your change to ignore callback during finalization and then added the write scheduler cleanup fix.

davekeeshan:
Cool I was just about to close it here myself.

A bit a flow issue, I wanted to show the bug really, but you can only access the CI with a PR, so I left it was irresponsible to do that without the attempt of a fix.

davekeeshan:
However in my i have upgraded my testing to the new `master` and see and error or complaint like this:

     -.--ns ERROR    gpi                                ..i/vpi/VpiCbHdl.cpp:43   in handle_vpi_callback             VPI: Callback fired during finalization.

seems related? or a new bug?

ktbarrett:
This is on your own test and not the one for this PR, correct? Its likely a different bug. Perhaps the one I posited first (that is also a know bug with Icarus AFAIK).

I should have had the error message print the VPI object pointer... but if you modify that log and enable GPI debugging, you should be able to trace it back to the culprit.

```
GPI_DEBUG=1 GPI_LOG_LEVEL=TRACE
```

ktbarrett:
Ah I see. So for VPI we don't actually remove the shutdown trigger since that makes some simulators very sad. So the shutdown trigger is going to fire. I think we should move the condition to after the "removed" shunt.

Or better yet, just remove the error logs.

davekeeshan:
I ran that just in case there is any extra information for you, but you probably have it:
```
 44400.00ns TRACE    gpi                                User Callback => [ VPI ]
 44400.00ns TRACE    gpi                                [ GPI Finalize ] => User Finalize callback
     -.--ns TRACE    gpi                                ../gpi/GpiCommon.cpp:130  in gpi_finalize                    User Finalize callback => [ GPI Finalize ]
     -.--ns TRACE    gpi                                ..i/vpi/VpiCbHdl.cpp:70   in handle_vpi_callback             [ VPI ] => Sim
     -.--ns TRACE    gpi                                ..i/vpi/VpiCbHdl.cpp:41   in handle_vpi_callback             Sim => [ VPI for cbEndOfSimulation ]
     -.--ns ERROR    gpi                                ..i/vpi/VpiCbHdl.cpp:43   in handle_vpi_callback             VPI: Callback fired during finalization.
```

ktbarrett:
I opened https://github.com/cocotb/cocotb/pull/5513 which I think solves the problem and also increases logging of callback creation and firing.

davekeeshan:
OK I have verified all the changes against my flow, looks good for me and I have tests passing on CI again. Thanks!

PR #5511, opened 2026-04-11
Add GPI logging function to check if log is enabled at a given level

### LINKED ISSUE #5510: API for checking if GPI logger is active at a given level
Right now GPI logging requires we evaluate all arguments to pass to the `gpi_log_` function. This may slightly decrease performance. We used to have a function that would check to see if the log level was active before doing a log. However when the Python logger is active there is more overhead in checking than in evaluating arguments. The middle way is the best, we should provide a function for callers to check to see if a logger is enabled at the given level, but not use it by default in the `LOG_*` macros.

ktbarrett:
Closes #5510.


src/cocotb/share/lib/gpi/logging.hpp
```diff
@@ -32,6 +32,9 @@ extern GPI_EXPORT int gpi_debug_enabled;
 /** @return The string representation of the GPI log level. */
 GPI_EXPORT const char *gpi_log_level_to_str(enum gpi_log_level level);
 
+GPI_EXPORT bool gpi_log_level_enabled(const char *logger,
```
cmarqu, line 36: Should there be a Doxygen comment here?

PR #5516, opened 2026-04-13
Remove special-casing Icarus library name

ktbarrett:
I'm not entirely sure why this was done this way. Perhaps it was a requirement of old Icarus versions?

Passing all supported Icarus versions: https://github.com/cocotb/cocotb/actions/runs/24327084698/


ktbarrett:
Looks like the new `test_integers` makes Icarus v11 sad, so I updated the test to use the Python Runner so I could do an easier version ifdef.

tests/test_cases/test_integers/Makefile
```diff
@@ -2,20 +2,9 @@
...
-VHDL_SOURCES += $(PWD)/integers.vhdl
-ifneq ($(filter $(SIM),ius xcelium),)
-	COMPILE_ARGS += -v93
-endif
-endif
-
-include $(shell cocotb-config --makefiles)/Makefile.sim
+clean:
+	rm -rf sim_build/
+	rm results.xml
```
ktbarrett:
```suggestion
	rm -f results.xml
```

ktbarrett (commented):
Also remove the first commit before merging.

docs/source/newsfragments/5516.removal.rst
```diff
@@ -0,0 +1 @@
+Removed the ``--lib-dir`` option from the :ref:`cocotb-config <cocotb-config>` script.
```
marlonjames:
```suggestion
Removed the ``--lib-name`` option from the :ref:`cocotb-config <cocotb-config>` script.
```

docs/source/newsfragments/5516.removal.1.rst
```diff
@@ -0,0 +1 @@
+Removed the :py:external+cocotb20:func:`cocotb_tools.config.lib_name` function.
```
marlonjames: This isn't linking correctly. Not sure if it is because of the `:py` prefix or not.
https://cocotb--5516.org.readthedocs.build/en/5516/release_notes.html#deprecations-and-removals

docs/source/newsfragments/5516.removal.1.rst
```diff
@@ -0,0 +1 @@
+Removed the :external+cocotb20:func:`cocotb_tools.config.lib_name` function.
```
marlonjames: I think this needs a `cocotb20` entry in the `intersphinx_mapping` dict in `conf.py`.

ktbarrett: Ah, we just aren't documenting any of the functions in that module. I'll remove the link here and add that module to the docs.

ktbarrett: https://github.com/cocotb/cocotb/pull/5526

marlonjames (approved):

Issue #5515, opened 2026-04-13
Remove RPATH and side-by-side assemblies


ktbarrett:
Currently we use RPATHs on Unix and Mac, and the equivalent side-by-side assemblies on Windows, to handle loading library dependencies since doing so via other means is prone to issues.

Currently we have the following RPATHs:
1. simulator.so: `$ORIGIN/libs`: to satisfy the simulator.so -> libgpi dep
2. libcocotb{intf}_{sim}: `$ORIGIN`: to satisfy the GPI impl -> libgpi dep

There also seems to be a now-redundant `$ORIGIN` RPATH in libgpi which is probably a remnant of the time when the util libraries were in multiple different libraries rather than just the single libgpi.

However, we can remove requirement 1 by adding `os.add_dll_directory` on Windows to `cocotb/__init__.py` before the simulator module is imported anywhere. Linux and Mac will continue to require using RPATH for this case, or we can use preloading, as there's no equivalent to `os.add_dll_directory` for Unix systems.

We can remove requirement 2 by refactoring the startup sequence to load into the libgpi first and extend `GPI_EXTRA` to also load the first implementation and not just "extras."

`GPI_EXTRA` loads impls dynamically and `GPI_USERS` loads users dynamically, so libgpi does not need RPATH set to either the impl or user. GPI impls no longer need to use RPATH to load the GPI as the GPI is loaded before any of the impls.

After this refactoring we can delete the side-by-side and RPATH stuff from the build system.

PR #5519, opened 2026-04-13
Simplify passing GPI_EXTRA

ktbarrett:
This also makes this variable only used for loading GPI implementations. Once the GPI Implementations are merged per-simulator, the value will change from being a path to a library to just an identifier (e.g. "vpi") so this creates an abstraction layer to prevent even more future changes.

### DIFF
--- a/docs/source/custom_flows.rst
+++ b/docs/source/custom_flows.rst
@@ -70,7 +70,7 @@ Aldec Riviera-PRO
       ``-loadvhpi $(cocotb-config --lib-name-path vhpi riviera):vhpi_startup_routines_bootstrap``.
 
       Set the :envvar:`GPI_EXTRA` environment variable to
-      ``$(cocotb-config --lib-name-path vpi riviera):cocotbvpi_entry_point``
+      ``$(cocotb-config --gpi-extra vpi riviera)``
       if there are also (System)Verilog modules in the design.
 
    .. tab-item:: Design with a (System)Verilog Toplevel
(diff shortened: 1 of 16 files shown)


marlonjames:
At some point, `GPI_EXTRA` should get renamed. Maybe as part of the refactor mentioned in #5515 ?

ktbarrett:
Yeah I think something like `GPI_IMPL`, which will be a list of all interfaces we want to load, including the first "main" one. So maybe we should do #5515 first. Or maybe we will just have to do it lock-step. Ultimately I think there will be some churn over the 2.2 and 2.3 cycles.

Ideally I'd like there to only be one impl per simulator that contains all the code for all supported interfaces and we don't need to specify which interfaces to use at all. It should make a reasonable choice for global callbacks things like startup and shutdown (e.g. Xcelium should always use VPI since the VHPI versions are broken) or ReadWrite, Timer, and ReadOnly. We use the appropriate value change callbacks based on the object's interface (VPI vs VHPI).

We can iterate over all impls for root objects searching for our toplevel and populating `cocotb.tops`. When we build the golden model with the VPI and VHPI, we just iterate over everything from our discovered roots, normalize paths, and stitch them together into our golden mixed language model. When we cross language boundaries with accesses, instead of doing get_handle_by_name or iterating over scopes *using the wrong interface* like we do now, we just do a lookup from the global model.

Maybe that's pie-in-the-sky thinking. Maybe not everything I mentioned is achievable. So just churning with incremental improvements gets us *somewhere*, even if it isn't an optimal route.

PR #5523, opened 2026-04-13
Fix indentation in Makefile.inc for language detection

ktbarrett:
These were accidentally tabs instead of spaces which makes make very sad.


cmarqu (approved):
make is such a PITA...

PR #5530, opened 2026-04-15
Make simulator version objects more useful

### LINKED ISSUE #5520: Make simulation version objects useful
The `cocotb_tools.sim_versions` module has implementation of each simulator's versions, but they just use `LooseVersion` without any other smarter conversion, so they ultimately don't work correctly.

This will require looking through the extended regression to see how version strings actually look in supported simulators through all the versions we support and customizing the version comparison.

Additionally, it would be nice to have a full complement of methods for constructing these versions.
One can be from `cocotb.SIM_VERSION` and the other by parsing the commandline output. In the second case, it should also call the simulator for you so it's easier to use in pytest tests:

```python
@pytest.skipif(SIM == "icarus" and IcarusVersion.from_commandline() < IcarusVersion("12.0 (devel)"))
def test_my_thing():
    ...
```

zeel2104:
Make simulator version objects more useful by replacing the current `LooseVersion`-style behavior with simulator-aware parsing and comparison.

This change adds common constructors for building version objects from `cocotb.SIM_VERSION` and from simulator command output, including a zero-argument `from_commandline()` path that invokes the simulator automatically when possible. It also updates existing call sites to use the new convenience API.

The main motivation is that several supported simulators use version strings that do not compare correctly with the current implementation, especially when they include release metadata, development tags, suffix letters, or train names. This patch makes those comparisons behave more like users expect and makes the version helpers easier to use in tests.

Changes:
- add a `SimulatorVersion` base with explicit parsing/comparison behavior
- add `from_sim_version()` and improved `from_commandline()` constructors
- implement simulator-specific parsing for Icarus, Verilator, Questa, VCS, Xcelium, GHDL, and NVC
- update existing runner/test call sites to use the convenience constructors
- add focused pytest coverage for known simulator version string formats

Testing:
- `$env:PYTHONPATH='d:\cocotb\src'; python -m pytest tests\pytest\test_sim_versions.py`

I could not exercise the zero-argument subprocess path against real simulator executables in this shell because no supported simulators were installed on `PATH`, but the parsing and comparison behavior is covered by unit tests.

Closes #5520

### DIFF
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -1257,13 +1257,7 @@ class Nvc(Runner):
     def __init__(self) -> None:
         super().__init__()
 
-        version_str = subprocess.run(
-            ["nvc", "--version"],
-            check=True,
-            text=True,
-            stdout=subprocess.PIPE,
-        ).stdout
-        version = NvcVersion.from_commandline(version_str)
+        version = NvcVersion.from_commandline()
         if version > NvcVersion("1.16"):
             self._preserve_case = ["--preserve-case"]
         else:
--- a/src/cocotb_tools/sim_versions.py
+++ b/src/cocotb_tools/sim_versions.py
@@ -2,10 +2,7 @@
 # Licensed under the Revised BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-3-Clause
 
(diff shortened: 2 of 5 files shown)


ktbarrett:
And there are pre-commit.ci failures.

Issue #5501, opened 2026-04-09
Change default build directory


ktbarrett:
Currently the build directory defaults to `sim_build`. This causes an issue if there are multiple builds in a single execution as they will all build and execute in the same directory clobbering each other and causing issues like #5311.

Instead we could default the build directory to `sim_build/{hdl_toplevel}`. This would require we pass `hdl_toplevel` to all builds.

tymonx:
What normally build systems do, they are generating a hash from the build context (list of source files, compilation arguments, ...). For example FuseSoC, Bazel, Buck2, ... This allows to have an unique build directory but still have a consistent path for incremental compilation. Most HDL simulators are supporting that. Verilator is using generated Makefiles for that. Using only the `hdl_toplevel` is not enough when using test parametrization to set HDL parameters or even different compilation/elaboration/simulation arguments for the same top level. But also when using the [pytest-xdist](https://pypi.org/project/pytest-xdist/) plugin to run everything in parallel.

I have solved that in the pytest plugin https://github.com/cocotb/cocotb/pull/5595. It hashes all build attributes [src/cocotb_tools/pytest/dut.py::Dut.id](https://github.com/tymonx/cocotb/blob/8e8451b6917d63ef86e859a799b4ea5f8c579b97/src/cocotb_tools/pytest/dut.py#L313) to generate consistent working directory that contains generated ID: `<build>/<dut.toplevel_library>/<dut.toplevel>/<dut.parameters>/<dut.id>`. This is passed to cocotb runner: [src/cocotb_tools/pytest/plugin.py::cocotb_pytest_dut_run](https://github.com/tymonx/cocotb/blob/8e8451b6917d63ef86e859a799b4ea5f8c579b97/src/cocotb_tools/pytest/plugin.py#L546).

ktbarrett:
Yeah my original solution was a bit pea brained. I had started to move it to using tempdirs on a local branch. But there were some issues and 2.1 change freeze was happening so I dropped it.

The Runners are going to get a massive refactor after 2.1 finally drops. I plan on fixing the build dir and also make build() return representations of that build that's passed to test() instead of heaping everything locally into the Runner. Finally I need to update Questa to use qrun to unblock a year+ old PR of mine.

I don't love this thing either. I'm hoping this is the last major change.

PR #5533, opened 2026-04-18
Fix flakey Verilator/MacOS test

### LINKED ISSUE #5311: Fix flakey Verilator+macOS test
https://github.com/cocotb/cocotb/actions/runs/21739514563/job/62762656047?pr=5289

When running `examples/matrix_multiplier/tests/test_matrix_multiplier.py` Verilator complains because it can't find the matrix_multiplier toplevel but finds dff. This is possibly because we are building both designs in the same directory and that is polluting the directory.

ktbarrett:
* Fixes #5311. Runs each test in separate directory to prevent builds from colliding in the same `sim_build`.
* Skip array_module tests for Icarus since it isn't supported. This wasn't seen until now because the previous step was picking up a totally different build.
* Adjust logging handling for pytest so...
   * We can handle logs using cocotb's loggers when no simulation is running. We will need this for non-simulator-context `cocotb.run` support. This is the `RuntimeError` branch of the SimTimeContext
   * We can deregister the cocotb logger after each pytest test so we aren't leaving state around after each test.
* Adds stdout printing on all CI pytest runs for better visibility on what it's doing.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -196,6 +196,7 @@ def dev_test_sim(
     ]
     session.run(
         "pytest",
+        "-s",
         "-v",
         "--doctest-modules",
         "--cov=cocotb",
@@ -211,6 +212,7 @@ def dev_test_sim(
     session.log(f"Running simulator-specific tests against a simulator {config_str}")
     session.run(
         "pytest",
+        "-s",
         "-v",
         "--cov=cocotb",
         "--cov-branch",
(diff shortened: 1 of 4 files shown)


ktbarrett:
Well... this has exposed a bunch of fun issues. We *really* need to be doing builds and tests out of unique directories. xref #5501.

ktbarrett:
Ugh... apparently the HDL fixture is run even when the test is skipped.

```
tests/pytest_plugin/test_sample_module.py::test_sample_module_parametrize[8] PASSED COCOTB RUNNER
tests/pytest_plugin/test_session.py::test_array_module SKIPPED (Icar...)
tests/pytest_plugin/test_session.py::test_sample_module /home/runner/work/cocotb/cocotb/tests/designs/array_module/array_module.sv:9: sorry: Unpacked structs not supported.
ERROR
```

I'm half tempted to just yeet that HDL thing.

ktbarrett:
I expect no coverage change. We aren't covering anything in `tests/pytest_plugin` and the lines changed in logging.py would only be hit by the exact thing we worked around with the `_restore_root_handler_filters` fixture.

marlonjames (approved):

PR #5547, opened 2026-04-24
fix: resolve memory leak in VPI callback cleanup

### LINKED ISSUE #5545: Memory leak with repeated Timer on VPI‑based simulators
## Summary

When using cocotb 2.0.1 with VPI‑based simulators, the simulator’s RSS memory grows linearly when repeatedly awaiting Timer calls. The same test case runs stably with cocotb 1.9.2 in identical environments.

## Environment

- COCOTB: 2.0.1(bad), 1.9.2(good)
- Python: 3.11
- Simulator: VCS V-2023.12-SP2, Xcelium 24.09, Xcelium 25.03
- OS: Rocky Linux 8
- Runner: Makefile, pytest 9.0

# Minimal example

```python
import os
import gc
import tracemalloc
import psutil
import cocotb

proc = psutil.Process(os.getpid())

@cocotb.test()
async def test_timer_leak(dut):
    tracemalloc.start(25)

    for i in range(1_000_000):
        await Timer(1, unit='us')

        if i % 50000 == 0:
            gc.collect()
            rss = proc.memory_info().rss / 1024 / 1024
            current, peak = tracemalloc.get_traced_memory()
            cocotb.log.warning(
                f"i={i} rss={rss:.1f}MB"
                f"i={i} py_cur={current/1024:.1f}KB"
                f"i={i} py_peak={peak/1024:.1f}KB"
            )
```

### Observed Behavior on cocotb 2.0.1:
- Simulator RSS grows linearly with iteration count.
- Python memory usage reported by tracemalloc remains stable.
- `Timer(1, unit='us')` and `Timer(10, unit='us')` show essentially the same RSS growth for the same number of iterations.
- Valgrind reports repeated memory loss records within the cocotb VPI layer.

## Root-cause hypothesis. 
The Timer trigger is implemented on VPI using `cbAfterDelay` via the `VpiTimedCbHdl` class. In cocotb 2.0.1, `VpiTimedCbHdl` inherits the default behavior of `VpiCbHdl::run()`. For non‑Verilator simulators (such as VCS and Xcelium), this default path does not call `vpi_remove_cb()` for `cbAfterDelay` callbacks. As a result, simulator‑side callback handles and associated native data are never fully released, leading to a persistent memory leak.

## Local workaround/proof. 
As an experiment, I forced the Verilator‑specific branch of `VpiCbHdl::run()` to be taken for VCS. With this modification, the simulator RSS no longer grows, and the repeated Timer reproducer becomes stable. This strongly supports the hypothesis that explicit removal of the callback is required for proper cleanup on these VPI implementations.

Dragon-Git:
Fixes #5545

This PR resolves the linear RSS memory leak when repeatedly using `Timer` triggers on Synopsys VCS and Cadence Xcelium.

### DIFF
--- a/tests/test_cases/test_trigger_memory_leak/test_trigger_memory_leak.py
+++ b/tests/test_cases/test_trigger_memory_leak/test_trigger_memory_leak.py
@@ -1,0 +1,200 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+"""
+Tests related to timing triggers
+
+* Timer
+* ReadWrite
+* ReadOnly
+* NextTimeStep
+* RisingEdge
+* FallingEdge
+* NullTrigger
+* ValueChange
+* Event
+"""
+
+from __future__ import annotations
+
+import gc
+import os
+
+import psutil
+
+import cocotb
+from cocotb.clock import Clock
+from cocotb.triggers import (
+    Event,
+    FallingEdge,
+    NextTimeStep,
+    NullTrigger,
+    ReadOnly,
+    ReadWrite,
+    RisingEdge,
+    Timer,
+    ValueChange,
+)
+
+SIM_NAME = cocotb.SIM_NAME.lower()
+proc = psutil.Process(os.getpid())
+# diff less than n * 4k for ASLR, if use THP, maybe lessthan n * 2MB
+MEMORY_LEAK_TH = 2**22 if SIM_NAME.startswith("riviera") else 2**21
+
+
+@cocotb.test
+@cocotb.skipif(
+    SIM_NAME.startswith("modelsim"),
+    reason="Questa behaves erratically with NextTimeStep",
+)
+async def test_next_time_step_leak(dut):
+    clk = Clock(dut.clk, 1, "ns")
+    cocotb.start_soon(clk.start())
+    rss_start = proc.memory_info().rss
+    await Timer(100_000, unit="ns")
+    rss_no_triger = proc.memory_info().rss
+    for _ in range(100_000):
+        await NextTimeStep()
+    gc.collect()
+    rss_triger = proc.memory_info().rss
+    diff = rss_triger - rss_no_triger - rss_no_triger + rss_start
+    assert diff <= MEMORY_LEAK_TH, "Memory leak"
+
+
+@cocotb.test()
+async def test_timer_leak(dut):
+    await Timer(1_000, unit="ns")
+    rss_start = proc.memory_info().rss
+    await Timer(100_000, unit="ns")
+    rss_no_triger = proc.memory_info().rss
+    for _ in range(100_000):
+        await Timer(1, unit="ns")
+    gc.collect()
+    rss_triger = proc.memory_info().rss
+    diff = rss_triger - rss_no_triger - rss_no_triger + rss_start
+    assert diff <= MEMORY_LEAK_TH, "Memory leak"
+
+
+@cocotb.test()
+async def test_readonly_leak(dut):
+    await Timer(1_000, unit="ns")
+    rss_start = proc.memory_info().rss
+    await Timer(100_000, unit="ns")
+    rss_no_triger = proc.memory_info().rss
+    for _ in range(100_000):
+        await ReadOnly()
+        await Timer(1, unit="ns")
+    gc.collect()
+    rss_triger = proc.memory_info().rss
+    diff = rss_triger - rss_no_triger - rss_no_triger + rss_start
+    assert diff <= MEMORY_LEAK_TH, "Memory leak"
+
+
+@cocotb.test()
+async def test_readwrite_leak(dut):
+    await Timer(1_000, unit="ns")
+    rss_start = proc.memory_info().rss
+    await Timer(100_000, unit="ns")
+    rss_no_triger = proc.memory_info().rss
+    for _ in range(100_000):
+        await Timer(1, unit="ns")
+        await ReadWrite()
+
+    gc.collect()
+    rss_triger = proc.memory_info().rss
+    diff = rss_triger - rss_no_triger - rss_no_triger + rss_start
+    assert diff <= MEMORY_LEAK_TH, "Memory leak"
+
+
+@cocotb.test()
+@cocotb.xfail(
+    SIM_NAME.startswith("modelsim"),
+    reason="Questa does not support calling vpi_remove_cb in a callback, which causes a memory leak",
+)
+async def test_rising_edge_leak(dut):
+    clk = Clock(dut.clk, 1, "ns")
+    cocotb.start_soon(clk.start())
+    rss_start = proc.memory_info().rss
+    await Timer(100_000, unit="ns")
+    rss_no_triger = proc.memory_info().rss
+    for _ in range(100_000):
+        await RisingEdge(dut.clk)
+    gc.collect()
+    rss_triger = proc.memory_info().rss
+    diff = rss_triger - rss_no_triger - rss_no_triger + rss_start
+    assert diff <= MEMORY_LEAK_TH, "Memory leak"
+
+
+@cocotb.test()
+@cocotb.xfail(
+    SIM_NAME.startswith("modelsim"),
+    reason="Questa does not support calling vpi_remove_cb in a callback, which causes a memory leak",
+)
+async def test_falling_edge_leak(dut):
+    clk = Clock(dut.clk, 1, "ns")
+    cocotb.start_soon(clk.start())
+    rss_start = proc.memory_info().rss
+    await Timer(100_000, unit="ns")
+    rss_no_triger = proc.memory_info().rss
+    for _ in range(100_000):
+        await FallingEdge(dut.clk)
+    gc.collect()
+    rss_triger = proc.memory_info().rss
+    diff = rss_triger - rss_no_triger - rss_no_triger + rss_start
+    assert diff <= MEMORY_LEAK_TH, "Memory leak"
+
+
+@cocotb.test()
+@cocotb.xfail(
+    SIM_NAME.startswith("modelsim"),
+    reason="Questa does not support calling vpi_remove_cb in a callback, which causes a memory leak",
+)
+async def test_value_change_leak(dut):
+    clk = Clock(dut.clk, 1, "ns")
+    cocotb.start_soon(clk.start())
+    rss_start = proc.memory_info().rss
+    await Timer(100_000, unit="ns")
+    rss_no_triger = proc.memory_info().rss
+    for _ in range(100_000):
+        await ValueChange(dut.clk)
+    gc.collect()
+    rss_triger = proc.memory_info().rss
+    diff = rss_triger - rss_no_triger - rss_no_triger + rss_start
+    assert diff <= MEMORY_LEAK_TH, "Memory leak"
+
+
+@cocotb.test()
+async def test_null_triger_leak(dut):
+    await Timer(1_000, unit="ns")
+    rss_start = proc.memory_info().rss
+    await Timer(100_000, unit="ns")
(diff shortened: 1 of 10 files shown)


ktbarrett:
Only thing missing is a newsfragment. See docs/source/newsfragments/README.rst for some details.

ktbarrett:
These memory leak tests take quite a long time. Can they still do a decent job if the number of iterations is cut to 200k or 500k?

ktbarrett:
The tests are so long they are blowing up the coverage tool.

Dragon-Git:
The testing time is indeed too long. I reduced the number of iterations to 100K.

Dragon-Git:
I observed a failure in the last CI run. Specifically, the `NextTimeStep` test failed when executed with Verilator 5.046 on the Ubuntu. Memory consumption reached 128KB. I tested it on the Verilator 5.046 macOS platform, and the `diff` is 0. I don't have an Ubuntu environment. Do we need to survey the environment, or do we need to feed this back upstream?

ktbarrett:
It seems like these tests are a bit flakey now that they are smaller. Is that expected? The Riviera tests weren't failing before and now they all are. Even Xcelium with the fix is failing in this run. 

I'm wondering if we keep this test at a larger size, like 200k or 500k, but run it as part of a different regression set. That isn't something you would have to do, just throwing it out there to see what you think.

Dragon-Git:
It seems that the noise is higher than I expected, but I think 100K tests are sufficient. We need to raise the threshold. Assuming that each leak is the size of the callback structure, that is, on a 32-bit system, the minimum occupancy is 28 Bytes, and 100k calls, the leak will reach 2.67 MB. It is reasonable for us to set the threshold to 2  MB.
```c
typedef struct t_cb_data
{
  PLI_INT32    reason;                        /* callback reason */
  PLI_INT32    (*cb_rtn)(struct t_cb_data *); /* call routine */
  vpiHandle    obj;                           /* trigger object */
  p_vpi_time   time;                          /* callback time */
  p_vpi_value  value;                         /* trigger object value */
  PLI_INT32    index;                         /* index of the memory word or
                                                 var select that changed */
  PLI_BYTE8   *user_data;
} s_cb_data, *p_cb_data;
```

Dragon-Git:
Could you please share the Questa and Riviera test logs for this PR?
I’d like to check the exact RSS diff values to verify whether there is a genuine memory leak or just normal memory fluctuation.

ktbarrett:
[riviera.log](https://github.com/user-attachments/files/27172364/riviera.log)

The questa failure has to do with it not consistently supporting NextTimeStep. It's kind of a finicky trigger. The issue is that the Clock and the Timer you have "finish" at the same time and so there are no active events and that causes this trigger to segfault. Playing with the Timer to not fall on the same time as the clock may fix it. Otherwise feel free to skip it.

Dragon-Git:
I skipped the test of `next_step_time` on Questa. Riviera seems that there is a memory leak. I tried to add a `vpi_free_object` to it.

tests/test_cases/test_triger_memory_leak/test_triger_memory_leak.py
```diff
@@ -0,0 +1,184 @@
...
+    ValueChange,
+)
+
+proc = psutil.Process(os.getpid())
+# diff less than n * 4k for ASLR, if use THP, maybe lessthan n * 2MB
+MEMORY_LEAK_TH = 2**21
+SIM_NAME = cocotb.SIM_NAME.lower()
+
+
+@cocotb.test(skip=(SIM_NAME.startswith("questa")))
```
ktbarrett: Yeah it's weird. "Questa" is the suite of tools and "ModelSim" is the name of the simulator.
```suggestion
@cocotb.test
@cocotb.skipif(SIM_NAME.lower().startswith("modelsim"))
```

ktbarrett:
Looks like Riviera also need to modify `VpiValueCbHdl::run`.

Dragon-Git:
`vpi_remove_cb` is now called in the `VpiValueCbHdl::run`, There should be no more memory leaks. Based on the last log, we should also consider raising the threshold for Riviera.

Dragon-Git:
Could you please share the Riviera test logs for this CI?

ktbarrett:
[log](https://github.com/user-attachments/files/27226792/wew.log) looks like Questa has the same leak of the callback, the define is `MODELSIM`.

Also, I noticed the test name is `test_triger_memory_leak`, there's a missing 'g' there.

Dragon-Git:
I noticed that for Riviera and ModelSim, both the rising edge callback and the falling edge callback consume more memory than the value change callback. I guess for these two emulators, we need to manually remove the callback or release the handle on each value change callback.

ktbarrett:
Riviera segfaults on the rising_edge test.

Questa:
```
# 802999.00ns WARNING  cocotb.regression                  test_trigger_memory_leak.test_raise_edge_leak failed
#                                                         Traceback (most recent call last):
#                                                           File "cocotb/tests/test_cases/test_trigger_memory_leak/test_trigger_memory_leak.py", line 117, in test_raise_edge_leak
#                                                             assert diff <= MEMORY_LEAK_TH, "Memory leak"
#                                                         AssertionError: Memory leak
#                                                         assert 11403264 <= 4194304
# 802999.00ns INFO     cocotb.regression                  running test_trigger_memory_leak.test_falling_edge_leak (5/8)
# 1002998.50ns WARNING  cocotb.regression                  test_trigger_memory_leak.test_falling_edge_leak failed
#                                                         Traceback (most recent call last):
#                                                           File "cocotb/tests/test_cases/test_trigger_memory_leak/test_trigger_memory_leak.py", line 132, in test_falling_edge_leak
#                                                             assert diff <= MEMORY_LEAK_TH, "Memory leak"
#                                                         AssertionError: Memory leak
#                                                         assert 12189696 <= 4194304
# 1002998.50ns INFO     cocotb.regression                  running test_trigger_memory_leak.test_value_change_leak (6/8)
# 1152998.00ns WARNING  cocotb.regression                  test_trigger_memory_leak.test_value_change_leak failed
#                                                         Traceback (most recent call last):
#                                                           File "cocotb/tests/test_cases/test_trigger_memory_leak/test_trigger_memory_leak.py", line 147, in test_value_change_leak
#                                                             assert diff <= MEMORY_LEAK_TH, "Memory leak"
#                                                         AssertionError: Memory leak
#                                                         assert 4718592 <= 4194304
```

Dragon-Git:
The attempt failed, but the above phenomenon still deserves attention. I reverted to the previous version, and skipped the value change related tests on questa. Do you have any ideas about this phenomenon?

ktbarrett:
About rising edge and falling edge using more? Yeah that's because the VPI only uses "value change" and we check to see if the value after change is 1 (rising edge) or 0 (falling edge) and if the current value change is to a different value, we keep going. If you are measuring clock cycles you'll get two edges per actual rising edge where the other is an ignored falling edge.

Dragon-Git:
Yes, I've test in Verilator, and confirmed that as long as we don't remove the callback, the callback's vpiHandle remains the same, so no memory leak occurs. However, for Questa and Riviera, I suspect that a different vpiHandle is generated each time, which causes memory leaks for half of the callbacks. So how should we handle this? Should we add dedicated handling for these two simulators when `pass == false`, by remove the callback and re-registering a new one? This may impact performance, or should we report this issue upstream instead?

Dragon-Git:
I have tested the scheme of unregistering and re-registering callbacks in Verilator. The performance impact is negligible.
 
In the test with 10M rising-edge trigger, both approaches finish in around 30 seconds.
For the 20M rising-edge trigger, the baseline scheme without unregistering/re-registering takes approximately 58 seconds, while the scheme with callback unregistration followed by re-registration takes about 60 seconds.

If the tests in Questa and Riviera pass, we can add this change, but we need to provide feedback to the upstream.

Dragon-Git:
Is the diff still consistent with before?

ktbarrett:
Sorry, been very distracted the past couple weeks. Let's try to get this merged this week. I appreciate the effort.

Dragon-Git:
Thanks for the feedback! Regarding the memory issue with Questa/Riviera-PRO during rising/falling edge tests, I've temporarily skipped the related test on Questa due to limited local debugging environment and reliance on CI for iteration. To prioritize landing this PR this week, should we:

Option A: Merge this and address the skipped test in a follow-up? (Ideal if the issue is isolated)

Option B: Debug deeper before merging? (If critical to resolve upfront)

Please advise your preference—I'm ready to adjust accordingly. Appreciate your guidance!

ktbarrett:
I want a deeper debug. I have access to the machine with the commercial tools so I can runs diagnostics more quickly.

Just to summarize where we are:

Questa is failing on these tests and we are uncertain if the issue is that is also needs the `vpi_free_object` or if it's creating a new callback on every iteration meaning we should be using `vpi_remove_cb`?

Dragon-Git:
Yes, I have tested both of these ideas in previous CIs, but they have failed. We may need to use valgrind to track memory. Riviera and Questa both have a similar phenomenon, but Questa use more memory.

ktbarrett:
It's possible it's a leak in Questa, or maybe removing a cb and adding it back in the same cycle triggers it. I wonder if it's repeating like Verilator and if so if we can just cache it and disable/re-enable it. cocotb had a system like that previously but it was broken over the years partially because it wasn't easy to understand the way it was written and totally uncommented.

ktbarrett:
Yeah this is just due to Questa not supporting `vpi_remove_cb` while the callback is executing. This will require a hefty refactor. Let's just get what we have in.

ktbarrett:
CI was broke. Lets try now.

tests/test_cases/test_trigger_memory_leak/test_trigger_memory_leak.py
```diff
@@ -0,0 +1,200 @@
...
+        await ReadWrite()
+
+    gc.collect()
+    rss_triger = proc.memory_info().rss
+    diff = rss_triger - rss_no_triger - rss_no_triger + rss_start
+    assert diff <= MEMORY_LEAK_TH, "Memory leak"
+
+
+@cocotb.test()
+@cocotb.skipif(
```
ktbarrett:
```suggestion
@cocotb.xfail(
```

tests/test_cases/test_trigger_memory_leak/test_trigger_memory_leak.py
```diff
@@ -0,0 +1,200 @@
...
+    for _ in range(100_000):
+        await RisingEdge(dut.clk)
+    gc.collect()
+    rss_triger = proc.memory_info().rss
+    diff = rss_triger - rss_no_triger - rss_no_triger + rss_start
+    assert diff <= MEMORY_LEAK_TH, "Memory leak"
+
+
+@cocotb.test()
+@cocotb.skipif(
```
ktbarrett:
```suggestion
@cocotb.xfail(
```

tests/test_cases/test_trigger_memory_leak/test_trigger_memory_leak.py
```diff
@@ -0,0 +1,200 @@
...
+    for _ in range(100_000):
+        await FallingEdge(dut.clk)
+    gc.collect()
+    rss_triger = proc.memory_info().rss
+    diff = rss_triger - rss_no_triger - rss_no_triger + rss_start
+    assert diff <= MEMORY_LEAK_TH, "Memory leak"
+
+
+@cocotb.test()
+@cocotb.skipif(
```
ktbarrett:
```suggestion
@cocotb.xfail(
```

ktbarrett:
Now the Riviera licenses are expired. Can't catch a break, can you?

Dragon-Git:
Ha, rough patch indeed 😅. Do you have an estimate on when the Riviera licenses might be available again?

ktbarrett:
Couple of days hopefully. I don't have those keys.

ktbarrett (approved):

PR #5550, opened 2026-04-30
Log test arguments and keyword arguments at test start

### LINKED ISSUE #4036: Have `cocotb.parameterize` print parameters when starting test
`TestFactory` would generate a docstring which would print all the parameters when the test was started. This is missing from `cocotb.parameterize`.

officialasishkumar:
Closes #4036.

When the regression manager begins running a test, also log the positional and keyword arguments that were passed to it. Previously only the first line of the test docstring was reported. The encoded test name (e.g. ``my_test/x=1/y=2``) uses ``_repr`` which truncates long values and silently substitutes index strings when a value is not representable, so it is not always sufficient for the user to see what parameter values a particular run is using.

For non-parametrized tests (no positional args, no keyword args) the log line is unchanged, so the existing log format is preserved for the common case. When a test does have arguments, an additional ``Parameters:`` line is appended on the same indented continuation as the docstring, so the existing one-line-per-test layout is retained.

This applies to tests generated by both :func:`cocotb.parametrize` and the legacy :class:`~cocotb.regression.TestFactory`, since both populate ``Test.args`` / ``Test.kwargs``.

Sample output for ``@cocotb.parametrize(x=[1, 2], y=[3, 4])``:

```
0.00ns INFO     running module.test/x=1/y=3 (1/4)
                Parameters: x=1, y=3
0.00ns INFO     running module.test/x=1/y=4 (2/4)
                Parameters: x=1, y=4
```

### TODO

- [x] newsfragment
- [x] tests

### DIFF
--- a/docs/source/newsfragments/4036.feature.rst
+++ b/docs/source/newsfragments/4036.feature.rst
@@ -1,0 +1,1 @@
+Test arguments and keyword arguments are now logged on test start, making the parameter values used by :func:`cocotb.parametrize` and :class:`~cocotb.regression.TestFactory` visible in the simulation log.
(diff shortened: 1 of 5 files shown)


ktbarrett:
I think this should also ensure that then test arguments are included in the XUnit report whether using pytest or the RegressionManager. Including it in the log is useful, but a half-measure.

officialasishkumar:
Addressed in f04a2d4e: the legacy RegressionManager now writes a `test_arguments` testcase property, and pytest-backed runs add the same value to JUnit user properties for parametrized tests.

PR #5551, opened 2026-04-30
Add reference docs for test pass/fail conditions

### LINKED ISSUE #5394: Add reference documentation on test pass/fail

officialasishkumar:
Closes #5394.

The narrative explanation of how a test is judged to pass, fail, skip, or expected-fail used to live exclusively in ``docs/source/writing_testbenches.rst``, even though the bulk of it was reference material rather than tutorial content. The section also still referenced ``cocotb.pass_test``, which has been deprecated since 2.1 in favor of ``cocotb.end_test``.

This PR moves the reference content into a new "Test Pass and Fail Conditions" section in ``docs/source/library_reference.rst``, where every outcome is documented and given a label (``test-pass-fail``, ``test-pass``, ``test-fail``, ``test-skip``, ``test-xfail``). The section also documents the four functions that can force a running test to end with a given outcome: ``cocotb.end_test``, ``pytest.skip``, ``pytest.xfail``, and ``pytest.fail``.

The "Passing and failing tests" section in ``writing_testbenches.rst`` is now a short tutorial overview that links into the reference, and the previously separate "Forcing a test to end with a given result" heading is folded into it. References to the deprecated ``cocotb.pass_test`` are dropped from the tutorial.

### TODO

- [x] newsfragment

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -391,24 +391,36 @@ The outcome is determined by what the test coroutine or any running :class:`~coc
 Passing tests
 -------------
 
-A test is considered to have ``passed`` if any of the following occur:
+A test is considered to have ``passed`` if it completes successfully
+and no expected-failure or expected-error condition is set.
+Successful completion occurs if any of the following happen:
 
 * The main test coroutine returns without raising an :exc:`!Exception`.
 * The test coroutine, or any running :class:`~cocotb.task.Task`, calls :func:`cocotb.end_test`.
 * The main test coroutine raises a :exc:`~asyncio.CancelledError`,
   or :keyword:`await`\ s a :class:`~cocotb.task.Task` that is cancelled and does not handle the cancellation.
 
+If an expected-failure or expected-error condition is set,
+successful completion is instead reported as ``failed`` because the expected failure or error did not occur.
+
 .. _test-fail:
 
 Failing tests
 -------------
 
-A test is considered to have ``failed`` if the main test coroutine, or any running :class:`~cocotb.task.Task`, does any of the following:
+A test is considered to have ``failed`` if it completes successfully when it was expected to fail or error,
+or if the main test coroutine, or any running :class:`~cocotb.task.Task`,
+produces an unexpected failure or error.
+Unexpected failures and errors include:
 
 * Fails an :keyword:`assert` statement (raises :exc:`AssertionError`).
 * Raises any other :exc:`!Exception` besides :exc:`!CancelledError`.
 * Fails a :func:`pytest.raises`, :func:`pytest.warns`, or :func:`pytest.deprecated_call` check.
 * Calls :func:`pytest.fail`.
+
+A matching ``expect_fail`` or ``expect_error`` argument to :deco:`cocotb.test`,
+or a matching :deco:`cocotb.xfail` decorator,
+changes the final outcome to ``expected fail`` instead.
 
 When a test fails, a stack trace is printed.
 If :mod:`pytest` is installed and :keyword:`assert` statements are used,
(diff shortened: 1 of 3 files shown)


docs/source/library_reference.rst
```diff
@@ -375,11 +375,84 @@ Discovering Tests
...
+
+.. _test-pass:
+
+Passing tests
+-------------
+
+A test is considered to have ``passed`` if any of the following occur:
+
+* The main test coroutine returns without raising an :exc:`!Exception`.
+* The test coroutine, or any running :class:`~cocotb.task.Task`, calls :func:`cocotb.end_test`.
```
ktbarrett, line 399: This is only true if no `expect_error` or `expect_fail` is set.

docs/source/library_reference.rst
```diff
@@ -375,11 +375,84 @@ Discovering Tests
...
+
+Failing tests
+-------------
+
+A test is considered to have ``failed`` if the main test coroutine, or any running :class:`~cocotb.task.Task`, does any of the following:
+
+* Fails an :keyword:`assert` statement (raises :exc:`AssertionError`).
+* Raises any other :exc:`!Exception` besides :exc:`!CancelledError`.
+* Fails a :func:`pytest.raises`, :func:`pytest.warns`, or :func:`pytest.deprecated_call` check.
+* Calls :func:`pytest.fail`.
```
ktbarrett, line 419: We do not currently support `pytest.fail`. Alternatively, you can add support. `assert False, "msg"` makes some linters upset, so supporting `pytest.fail` would be nice.

ktbarrett: Also this only causes a failure if no `xfail` decorator is used.

docs/source/library_reference.rst
```diff
@@ -375,11 +375,84 @@ Discovering Tests
...
+will end the test immediately with a given outcome.
+They can be used in places where it is awkward to apply the equivalent decorator,
+for example because the condition is only known at runtime.
+
+* :func:`cocotb.end_test` ends the test as if it returned normally.
+  Any :deco:`cocotb.xfail` decorator, or ``expect_error`` and ``expect_fail`` arguments to :deco:`cocotb.test`, are still respected,
+  so the final outcome can be ``pass``, ``fail``, or ``expected fail``.
+* :func:`pytest.skip` ends the test with a ``skipped`` outcome.
+* :func:`pytest.xfail` ends the test with an ``expected fail`` outcome.
+* :func:`pytest.fail` ends the test with a ``failed`` outcome.
```
ktbarrett: `pytest.fail` is essentially equivalent to `assert False` so decorators are still respected.

ktbarrett (commented):
Thanks for looking into this. A couple notes.

docs/source/library_reference.rst
officialasishkumar: Addressed in 89d6edc2: the pass section now distinguishes successful completion from the final outcome when `expect_fail` or `expect_error` is set.

docs/source/library_reference.rst
officialasishkumar: Addressed in 89d6edc2: the failure section now notes that matching expected-failure handling changes the final outcome to `expected fail`.

docs/source/library_reference.rst
officialasishkumar: Addressed in 89d6edc2: the `pytest.fail` entry now describes it like a failing assert and says expected-failure conditions are respected.

Issue #4948, opened 2025-09-07
Clean up extended tool version test


imphil:
We see a number of failures of various sorts in the extended tool version tests. Clean those up. 

https://github.com/cocotb/cocotb/actions/runs/17522933093

imphil:
- Icarus 11 should be fixed by https://github.com/cocotb/cocotb/pull/4949.
- The Riviera tests are a mixed bag, looking into it.

PR #5552, opened 2026-05-02
Fix extended regressions

ktbarrett:
* Bumps minimum supported version of NVC to 1.19:
   * <1.16 don't support preseve-case which is the way forward
   * 1.16-1.19 have a bug where `--work` and `--preserve-case` don't work together #1413
* Adds NVC 1.20 to regression
* Fixes a couple extended regression tests, xref #4948.
* Move commercial sim modulefile loading into noxfile

Manual run here: https://github.com/cocotb/cocotb/actions/runs/25894738596

### DIFF
--- a/.github/generate-envs.py
+++ b/.github/generate-envs.py
@@ -41,7 +41,7 @@ ENVS = [
     {
         "lang": "vhdl",
         "sim": "nvc",
-        "sim-version": "r1.19.2",
+        "sim-version": "r1.20.0",
         "os": "ubuntu-22.04",
         "python-version": "3.9",
         "group": "ci-free",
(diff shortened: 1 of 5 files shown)


ktbarrett:
Looks like Riviera is still failing with GLIBC incompatibilities. We need to build cocotb using system libraries, i.e. build before testing. But nox rebuilds cocotb when it runs =/ This limitation is biting us again... we *really* need to get on to a better tool.

For now, we can LD_PRELOAD the system glibc, presuming it's newer than any linked in tools we run.

ktbarrett:
Okay well that didn't work. Seems like the only way to make this happy is to load the modulefiles *after* cocotb is built. Will have to move the bits into nox.

imphil (approved):

PR #5555, opened 2026-05-04
add VpiRefObject resolution using vpiActual

### LINKED ISSUE #5089: Support `vpiRefObj`
Hello, when using Cocotb 2.0.0, it is not possible to access an interface which is part of a module's ports. It works fine using Cocotb 1.8.1.
Simulator Questa 2023.1
I've enhance the test test_sv_interface to demonstrate the issue.

top.sv
```
`timescale 1us/1us

interface sv_if();
  logic a;
  reg b;
  wire c;
endinterface

module child (
    sv_if sv_if
);
endmodule

module top ();

sv_if sv_if_i();

sv_if sv_if_arr[3]();

child child_inst(.sv_if(sv_if_i));

endmodule
```
test_sv_if.sv
```
@cocotb.test()
async def test_sv_if(dut):
    """Test that signals in an interface are discovered and iterable"""

    dut.sv_if_i._discover_all()
    assert hasattr(dut.sv_if_i, "a")
    assert hasattr(dut.sv_if_i, "b")
    assert hasattr(dut.sv_if_i, "c")
    assert hasattr(dut.child_inst.sv_if, "a")
    assert hasattr(dut.child_inst.sv_if, "b")
    assert hasattr(dut.child_inst.sv_if, "c")
```

With Cocotb 2.0.0, I get the following error:
```
#      0.00ns WARNING  gpi                                VPI: Not able to map type vpiRefObj(608) to object.
#      0.00ns WARNING  cocotb.Test test_sv_if.test_sv_if  top.child_inst contains no child object named sv_if
#                                                         Traceback (most recent call last):
#                                                           File "/home/fdube/sandbox/cocotb/tests/test_cases/test_sv_interface/test_sv_if.py", line 16, in test_sv_if
#                                                             assert hasattr(dut.child_inst.sv_if, "a")
#                                                           File "/home/fdube/.local/lib/python3.10/site-packages/cocotb/handle.py", line 484, in __getattr__
#                                                             raise AttributeError(f"{self._path} contains no child object named {name}")
#                                                         AttributeError: top.child_inst contains no child object named sv_if
```

felixdube:
- Add testcase for accessing an interface which is part of a module's ports.
- Fix: add VpiRefObject resolution using vpiActual
- Close #5089

### DIFF
--- a/src/cocotb/share/lib/gpi/vpi/VpiImpl.cpp
+++ b/src/cocotb/share/lib/gpi/vpi/VpiImpl.cpp
@@ -202,8 +202,8 @@ GpiObjHdl *VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
         }
         type = vpi_get(vpiType, actual_hdl);
         vpi_free_object(actual_hdl);
-        LOG_DEBUG("VPI: Resolved vpiRefObj %s to type %d",
-                  fq_name.c_str(), type);
+        LOG_DEBUG("VPI: Resolved vpiRefObj %s to type %d", fq_name.c_str(), 
+		  type);
     }
 
     /* What sort of instance is this ?*/
--- a/src/cocotb/share/lib/gpi/vpi/VpiImpl.cpp
+++ b/src/cocotb/share/lib/gpi/vpi/VpiImpl.cpp
@@ -203,7 +203,7 @@ GpiObjHdl *VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
         type = vpi_get(vpiType, actual_hdl);
         vpi_free_object(actual_hdl);
         LOG_DEBUG("VPI: Resolved vpiRefObj %s to type %d", fq_name.c_str(),
-		   type);
+                  type);
     }
 
     /* What sort of instance is this ?*/
--- a/src/cocotb/share/lib/gpi/vpi/VpiImpl.cpp
+++ b/src/cocotb/share/lib/gpi/vpi/VpiImpl.cpp
@@ -203,7 +203,7 @@ GpiObjHdl *VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
         type = vpi_get(vpiType, actual_hdl);
         vpi_free_object(actual_hdl);
         LOG_DEBUG("VPI: Resolved vpiRefObj %s to type %d", fq_name.c_str(), 
-		  type);
+		 type);
     }
 
     /* What sort of instance is this ?*/
--- a/src/cocotb/share/lib/gpi/vpi/VpiImpl.cpp
+++ b/src/cocotb/share/lib/gpi/vpi/VpiImpl.cpp
@@ -202,7 +202,7 @@ GpiObjHdl *VpiImpl::create_gpi_obj_from_handle(vpiHandle new_hdl,
         }
         type = vpi_get(vpiType, actual_hdl);
         vpi_free_object(actual_hdl);
-        LOG_DEBUG("VPI: Resolved vpiRefObj %s to type %d", fq_name.c_str(), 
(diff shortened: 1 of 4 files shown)


ktbarrett:
I'm guessing Verilator doesn't implement this?

felixdube:
That's right, Verilator does not implement this. I guess we could exclude it from that check, but I am looking into the possibility to add that feature in Verilator. I've just been caught in some other work for the past few weeks.

As for Riviera, I am not sure what's up. I don't have a Riviera license for investigating the failure.

ktbarrett:
Riviera may just be a license issue.

toddstrader:
The verilator change is here:
https://github.com/verilator/verilator/pull/8081
Hopefully that will land soon.

I'm testing the verilator change plus this PR on our codebase and can now get handles to interface references.  I can get one of our existing tests to monitor an interface reference instead of the concrete interface by making this change to the PR:
https://github.com/toddstrader/cocotb/tree/intf-ref

However, something is a bit wonky here / there are open questions.  `dut.foo.intf_ref` yields `HierarchyObject(tb.dut.intf_ref with definition SomeIntf)` if you evaluate that before `dut.intf`.  But if you do it in the other order you get `HierarchyObject(tb.intf with definition SomeIntf)`.

I think the question we need to resolve is, what happens when you ask for a handle to an interface reference?  Do you "see through" to the concrete interface and just get a handle to that?  Or does the cocotb handle know the path from which you requested the handle?  And if so, can you somehow know that two interface references or an interface reference and the concrete interface are actually the same thing?

I think not seeing through is good and being able to compare is needed.  But admittedly I haven't played around with this a lot yet.

Thoughts?

ktbarrett:
> I think not seeing through is good and being able to compare is needed.

"not seeing through" implies we treat the references as if they are the actual object, which is fine. But "being able to compare is needed" implies it's not the actual object. This seems contradictory. Is it the referred object with the referred path, or the accessor object as an accessor object? Mixing them seems like a semantic sinkhole. I'm for the first and if we later care to expose the reference as a separate object we can do so.

As for object comparison, would it intended to work like the following?

```python
dut.foo.intf_ref.compare(dut.intf)  # compare looks through refs
dut.other.intf_ref.compare(dut.other.intf_ref)  # looks through refs so both refs compare equal
dut.mod.signal.compare(cocotb.top["mod/signal"])  # same object despite different handles
```

IDR off the top of my head if this is how `vpi_compare_objects` works or not. Also `vpi_compare_objects` is generally unsupported (I tested a couple years ago), so we would have to cook this up ourselves in the GPI.

toddstrader:
In SystemVerilog the interface reference isn't a scope and therefore you can't get handles to the interface signals from it without going through `vpiActual` and getting the handle to the _actual_ interface.

I guess the other option here is that we expose the notion of references via cocotb.  We could have a `HierarchyReference` which basically just has a `resolve()` method to get the `HierarchyObject` for the actual interface.

ktbarrett:
> In SystemVerilog the interface reference isn't a scope and therefore you can't get handles to the interface signals from it without going through vpiActual and getting the handle to the actual interface.

Right, so `dut.foo.intf_ref` would just return `dut.intf`? That sounds reasonable to me. You would go through `vpiActual` to get a handle to the actual object and then get the `vpiFullname` and look that up in the handle cache and return the canonical GPI handle to that. Then you can get fields just fine.

> I guess the other option here is that we expose the notion of references via cocotb.

Unless there is other value in such a type, I don't think this is the play.

toddstrader:
Talked with @ktbarrett offline and (he can correct me if I'm misstating) we should always get a handle to the concrete interface regardless of whether you go through an interface or not.  And `._path` should always report the concrete interface.

My branch is doing the former but not the latter.  @felixdube do you want to run with this?

toddstrader:
I have a draft for what we discussed here in #5790 .

PR #5577, opened 2026-05-26
Move pytest plugin into a private module and co-locate all documentation

ktbarrett:
Walking this feature back before the release. It's going to change a lot in 2.2 as we make it more like we want.


docs/source/pytest_plugin_reference.rst
```diff
@@ -0,0 +1,88 @@
+.. _api-pytest-plugin:
```
imphil, line 1: Can we get a very loud warning on this page that this feature is subject to change and should not yet be used? Maybe also some wording on what people are actually looking for (pytest assert rewriting? pytest in cocotb runner? something like pytest for hardware == cocotb) I fear that this page will show up in Google searches as first thing users find when looking for "pytest cocotb", and we need to get people away from it again as soon as we can 😄

docs/source/pytest.rst
```diff
@@ -4,7 +4,10 @@
 Pytest Support
 **************
 
-:py:mod:`cocotb_tools.pytest.plugin` provides full `pytest`_ integration with cocotb. Including:
+.. warning::
+    The pytest plugin is under active development and the API **will** change radically over the next release or two.
```
imphil: Also here: provide alternatives to people what they might actually mean when looking for "cocotb pytest" in google "what you're most likely looking for is pytest assert rewriting, something like pytest for hardware, ..."

imphil (approved):
Thanks! Let's make the "do not use this thing" stronger for people who stumble across random pages of our docs in Google. Apart from that, ready to go!

PR #5578, opened 2026-05-26
Add option to fail nox fast

ktbarrett:
It's not always easy to see which test failed because there are many ways this can go unreported.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -172,7 +172,7 @@ def dev_test_sim(
     if "COCOTB_CI_SKIP_MAKE" not in os.environ:
         session.log(f"Running 'make test' against a simulator {config_str}")
         make_args = ["make"]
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -171,7 +171,11 @@ def dev_test_sim(
 
     if "COCOTB_CI_SKIP_MAKE" not in os.environ:
         session.log(f"Running 'make test' against a simulator {config_str}")
-        session.run("make", "-k", "test", external=True, env=env)
+        make_args = ["make"]
+        if "COCOTB_FAIL_FAST" not in os.environ:
```
marlonjames: Maybe add `CI` for clarity, otherwise it's easy to assume it affects regression manager running user tests.
```suggestion
        if "COCOTB_CI_FAIL_FAST" not in os.environ:
```

marlonjames (approved):

PR #5590, opened 2026-05-31
Add IPython support interaction test

### LINKED ISSUE #5472: Add automated testing of `ipython_support`
This has the opportunity to break since we are using a couple third-party packages and private/non-stable cocotb functionality. We need to figure out some way to script interactive IPython sessions.

ztysth:
This adds automated coverage for the interactive IPython support path requested in #5472.

The new simulator-required pytest runs the existing `sample_module` design with `cocotb_tools.ipython_support` as the test module, then drives the embedded IPython prompt with `pexpect`. It checks that the shell can access `dut`, import cocotb APIs, await a `Timer`, and exit cleanly.

The test uses `pexpect.spawn` and is skipped when that API is unavailable.

`IPython` and `pexpect` are added to the `dev_test` dependency group so this coverage is available in the development test environment.

Closes #5472.

### DIFF
--- a/tests/pytest/test_ipython_support.py
+++ b/tests/pytest/test_ipython_support.py
@@ -1,0 +1,109 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+from __future__ import annotations
+
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+import pexpect
+import pytest
+
+if sys.platform != "win32":
+    from pexpect import spawn
+
+pytestmark = pytest.mark.simulator_required
+
+tests_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+sample_module_dir = os.path.join(tests_dir, "designs", "sample_module")
+sim = os.getenv(
+    "SIM",
+    "icarus" if os.getenv("TOPLEVEL_LANG", "verilog") == "verilog" else "nvc",
+)
+
+
+def spawn_make(env: dict[str, str]):
+    return spawn(
+        "make",
+        ["-C", sample_module_dir],
+        env=env,
+        encoding="utf-8",
+        timeout=90,
+    )
+
+
+def terminate_child(child) -> None:
+    if hasattr(child, "proc"):
+        if child.proc.poll() is None:
+            child.proc.terminate()
+            try:
+                child.proc.wait(timeout=5)
+            except subprocess.TimeoutExpired:
+                child.proc.kill()
+                child.proc.wait()
+    elif child.isalive():
+        child.terminate(force=True)
+
+
+def assert_child_exited(child) -> None:
+    if hasattr(child, "proc"):
+        assert child.wait() == 0
+    else:
+        assert child.exitstatus == 0
+
+
+@pytest.mark.skipif(
+    sim not in ["icarus", "verilator", "ghdl", "nvc", "xcelium"],
+    reason="Skipping test because it requires direct access to simulator output",
+)
+@pytest.mark.skipif(
+    sys.platform == "win32",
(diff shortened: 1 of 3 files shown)


tests/pytest/test_ipython_support.py
```diff
@@ -0,0 +1,124 @@
...
+
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+pexpect = pytest.importorskip("pexpect")
+pexpect_spawn = getattr(pexpect, "spawn", None)
```
ktbarrett: This can be a hard failure. It's required by the dev environment, so if someone doesn't have it, their environment is out of sync. So worth failing.
```suggestion
from pexpect import spawn
```

tests/pytest/test_ipython_support.py
```diff
@@ -0,0 +1,124 @@
...
+
+def skip_if_prompt_unavailable(output: str) -> None:
+    if "No module named 'IPython'" in output:
+        pytest.skip("simulator Python environment does not have IPython installed")
+
+    if "IPython " in output and " supports Python " in output:
+        pytest.skip("simulator Python environment has an incompatible IPython version")
+
+    if "failed to open using dlopen()" in output and "GLIBC_" in output:
+        pytest.skip("simulator cannot load the cocotb library in this environment")
```
ktbarrett: These should all be hard failures. Having IPython available and importing correctly is something the dev environment and CI should guarantee. Same with GLIBC, the environment should handle this.

tests/pytest/test_ipython_support.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@pytest.mark.skipif(pexpect_spawn is None, reason="pexpect.spawn is not available")
+@pytest.mark.skipif(
+    sim in ["questa", "riviera"],
+    reason="Skipping test because it requires direct access to simulator output",
+)
```
ktbarrett, line 60: Better to write a positive list of simulators that do support this:

```suggestion
@pytest.mark.skipif(
    sim not in ["icarus", "verilator", "ghdl", "nvc", "xcelium"],
    reason="Skipping test because it requires direct access to simulator output",
)
```

tests/pytest/test_ipython_support.py
```diff
@@ -0,0 +1,124 @@
...
+        pytest.skip("simulator Python environment does not have IPython installed")
+
+    if "IPython " in output and " supports Python " in output:
+        pytest.skip("simulator Python environment has an incompatible IPython version")
+
+    if "failed to open using dlopen()" in output and "GLIBC_" in output:
+        pytest.skip("simulator cannot load the cocotb library in this environment")
+
+
+@pytest.mark.skipif(pexpect_spawn is None, reason="pexpect.spawn is not available")
```
ktbarrett:
```suggestion
```

ktbarrett (commented):
Looks good. Thank you so much!

Just need to remove a bunch of unnecessary skip conditions. I'd rather have hard failures we need to investigate than skipping and not realizing we aren't testing anything.

ztysth:
Thanks for the review!
I applied all suggestions locally and force-pushed the amended commit. The dependency/environment skip paths are removed, and the simulator skip now uses the positive list you suggested.
I also verified the test locally. It works.

ktbarrett:
> from pexpect import spawn
   ImportError: cannot import name 'spawn' from 'pexpect'

~~Not sure what that's about.~~ Oh, this is on Windows. Apparently pexpect doesn't work well on Windows. That's fine we can skip if on Windows. The Riviera failures are just time outs from license contention.

ztysth:
I added a Windows skip for this test since `pexpect.spawn` is not available there, while keeping the dependency/environment failures as hard failures on other platforms. I left the Riviera behavior unchanged since those failures appear to be license contention timeouts.

ktbarrett (approved):

Issue #5425, opened 2026-03-18
pytest support: API design


imphil:
I took a bit of time to go through what we currently have in the master branch as pytest support (https://docs.cocotb.org/en/development/pytest.html). My main concern with the current API is the overloading of the use "test".

In the current pytest support, there are (at least) two types of "tests". At the outer layer, a pytest test (marked with `@pytest.mark.cocotb_runner`) builds a simulation model and starts a simulator. At the inner layer, a test is a cocotb test (which does necessarily not have the well-known `@cocotb.test` marker any more).

For cocotb users, things like writing, reading and asserting signal values is a "test". Building a hardware design and running the simulator is not a test -- it's a setup step before a test can be run. Or in pytest words: a fixture.

As a rough draft, can we get to something like that?

```python3
# pseudo code, draft. Not implemented!

# A fixture that provides access to a design under test: the dut
# The only thing that's new.
@pytest.fixture
def dut() -> Dut:
    # Let's not focus on that part for the moment.
	return Design(verilog_sources=["top.sv"], toplevel="top").with_simulator("icarus").build().run()

# A cocotb test that uses the design under test. Unchanged from how you'd write that code today.
@cocotb.test
async def test_something(dut):
	dut.clk.value = 1
	dut.rst.value = 0
	assert dut.a.value == dut.b.value
```

Overall, I'd like to see an API that follows a couple of principles:
- A test actually a test: either a cocotb test, or a Python unit test (for cases where you're mixing unit tests and cocotb tests).
- Runners are an implementation detail, just like makefiles or other ways to start the simulator. Don't expose that naming.
- Keep the magic down to a minimum and the code understandable for somebody who does not know cocotb (well). That includes features like autodiscovery.
- Try to provide a single way of doing things. Users need guidance, not options. (See also "orthogonality".)

FYI @ktbarrett @tymonx

ktbarrett:
I'm down for everything mentioned. I had originally envisioned this as fixture as build which would slot into `dut` argument. The only reservation I have is that it would be nice to support those existing projects that are already using pytest with Runner tests (like old cocotb-test). While the new way would be better, the way the plugin is written now is you get all or nothing so there's no way to gradually move to the new way in old projects already using pytest if we don't support that flow at all.

[...]

PR #5595, opened 2026-06-05
feat(pytest-plugin-v2): 1/N: add plugin options, dut fixture, markers

tymonx:
Implementing a new API for pytest plugin proposed in this issue: https://github.com/cocotb/cocotb/issues/5425

I'm trying to introduce smaller and separate Pull Requests but still functional in some way.

The original plugin (v1) was moved from public module `src/cocotb_tools/pytest/` to private module `src/cocotb_tools/_pytest/`. This will allow to introduce changes in this and future Pull Requests with minimal diff but with some copy-paste from that private module. Later we can nuke that private module in separate PR.

The PR 1/N introduces the bare minimum for a very minimalistic functional plugin:

- Plugin options that can be set from the `pytest` command line interface, pytest configuration files and environment variables. They are the same like in the v1 
- The new `Dut` class that is representing the Design Under Test. It acts as a middleware between user and plugin where user is providing all necessary information required to compile and run HDL design that will be used by plugin. These information can be set by using plugin options, markers and/or `Dut` attributes from fixture functions. The `Dut` class is improved and refactored version of the `HDL` class from v1 where the `HDL` class was nuked in v2
- The dual-purpose `dut` fixture. From a non-simulation environment, it creates and returns an instance of `Dut` that will be passed to plugin. From a simulation environment, it returns the standard `cocotb.top` handler to drive and monitor signals from cocotb test. The diff from v1 is just `cocotb.top if cocotb.is_simulation else Dut.create(request)`
- `@pytest.mark.cocotb_*` markers that will allow to configure an instance of `Dut` returned from the `dut` fixture. They are the same like in the v1
- This PR already allows to use current implementation of plugin with built-in cocotb regression manager and approach with cocotb runners by invoking the `Dut.run()` method from a pytest test function. This was improved compared to v1

A working example can be seen in the `test_dut_run` test function implemented in the `tests/pytest_plugin/test_dut.py` test file.

This already implements a suggestion from this comment: https://github.com/cocotb/cocotb/issues/5425#issuecomment-4083743585

> Would it make sense to separate the plugin into two? One for adding the Runner CLI configuration options (`--cocotb-sim` and friends) that is more likely to be mature and ready soon and can be used by those old projects.

That's why the `--cocotb-regression-manager` option was added. Invoking the `pytest` with the `--cocotb-regression-manager=cocotb` option will use the built-in cocotb regression manager to run cocotb tests and still allowing to use goodies from this PR (pytest plugin options, `dut` fixture, `@pytest.mark.cocotb_*` markers).

Final version of this plugin will automatically invoke the `Dut.run()` method in separate process. And declaring a separate pytest test function with a cocotb runner will be not needed anymore.

This PR also improves few things compared to plugin v1:

- Using the pytest built-in [pytester](https://docs.pytest.org/en/stable/how-to/writing_plugins.html#testing-plugins) plugin to test plugin in isolation
- Separate tests that require simulation environment or not by using the pytest `simulator_required` marker
- 330 tests only for this PR hitting 100% code coverage
- More docstrings
- Plugin is "smart" and, if not requested explicitly, it will automatically detect and use proper HDL simulator

To run these tests without nox (they are also included in `noxfile.py`):

```plaintext
pytest ./tests/pytest_plugin/test_{dut,mark,option,plugin,simulator}.py
```

Incoming n/N PRs:

- 2/N: Collecting `Dut` fixtures, implicitly creating pytest Items from them that will invoke the `Dut.run` method. This will allow to remove need of explicitly using cocotb runners in separate test functions and even it will work with the built-in cocotb regression manager.
- Running cocotb tests via pytest (by moving Python files related with pytest as regression manager from private module `src/cocotb_tools/_pytest/` to public module)
- Documentation about pytest plugin. Removing information about usage of cocotb runners, adding more examples with `dut` fixture and markers, migration steps from cocotb runner to pytest plugin (this will be joyful because it will require only to replace existing cocotb runners with fixtures and it should just "work" :tm: with existing cocotb tests)

### DIFF
--- a/src/cocotb_tools/pytest/dut.py
+++ b/src/cocotb_tools/pytest/dut.py
@@ -1,0 +1,434 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Design Under Test (DUT) representation and configuration management."""
+
+from __future__ import annotations
+
+from argparse import Namespace
+from collections.abc import Sequence
+from hashlib import md5
+from pathlib import Path
+from typing import Any, Callable
+
+from pytest import Config, FixtureRequest, Mark, Module
+
+from cocotb_tools.pytest._simulator import (
+    detect_language,
+    detect_languages,
+    find_simulator,
+    get_supported_languages,
+)
+from cocotb_tools.runner import (
+    VHDL,
+    PathLike,
+    VerilatorControlFile,
+    Verilog,
+)
+
+#: Exclude these attributes when generating unique DUT identifier
+_EXCLUDE_FROM_ID: tuple[str, ...] = (
+    "test_modules",
+    "random_seed",
+    "build_dir",
+    "work_dir",
+    "id",
+)
+
+
+class Dut:
+    """Representation of a Design Under Test (DUT).
+
+    This class holds the configuration and state required to compile, elaborate,
+    and run an HDL module using an HDL simulator and cocotb tests.
+
+    Example:
+        To customize a DUT in a test module or fixture:
+
+        .. code-block:: python
+
+            import pytest
+            from pathlib import Path
+            from cocotb_tools.pytest.dut import Dut
+
+            DIR = Path(__file__).parent.resolve()
+
+
+            @pytest.fixture
+            def my_module(dut: Dut) -> Dut:
+                dut.sources += [DIR / "rtl" / "my_design.v"]
+                dut["WIDTH"] = 16
+
+                return dut
+    """
+
+    def __init__(self, request: FixtureRequest) -> None:
+        """Initialize a new instance of DUT (Design Under Test).
+
+        Args:
+            request: The pytest fixture request object containing configuration and context.
+        """
+        config: Config = request.config
+        option: Namespace = config.option
+
+        self._run: Callable[..., None] = request.config.hook.pytest_cocotb_dut_run
+        self._simulator: str | None
+        self.simulator = option.cocotb_simulator
+
+        # Build options
+        self._defines: dict[str, object]
+        self.defines = option.cocotb_defines
+
+        self._parameters: dict[str, object]
+        self.parameters = option.cocotb_parameters
+
+        self._timescale: tuple[str, str] | None
+        self.timescale = option.cocotb_timescale
+
+        self.library: str = option.cocotb_library
+        """The library name to compile into."""
+
+        self.sources: list[PathLike | VHDL | Verilog | VerilatorControlFile] = (
+            option.cocotb_sources.copy()
+        )
+        """Language-agnostic list of source files to build."""
+
+        self.includes: list[PathLike] = option.cocotb_includes.copy()
+        """Verilog include directories."""
+
+        self.build_args: list[str | VHDL | Verilog] = option.cocotb_build_args.copy()
+        """A list of extra build arguments for the simulator."""
+
+        self._toplevel: str | None = None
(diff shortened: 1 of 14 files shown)


tymonx:
CI job with Riviera timeouted:

> Error: The action 'Test (Ubuntu, MacOS)' has timed out after 40 minutes.

These 330 tests are only adding 10-20 seconds (tested with GHDL, Verilator and Icarus) :melting_face:

ktbarrett:
The Riviera license is expired. Phil will install a new one soon.

tymonx:
Sorry, no GitHub for you today https://www.githubstatus.com/ :melting_face: 

image

src/cocotb_tools/pytest/dut.py
```diff
@@ -0,0 +1,434 @@
...
+            setattr(self, name, bool(marker.args[0]) if marker.args else True)
+        elif marker.args:
+            setattr(self, name, marker.args[0])
+
+    def _update_dict_with_list(
+        self, items: dict[str, object], values: Sequence[str]
+    ) -> None:
+        for value in values:
+            key, _, item = value.partition("=")
+            items[key] = item
```
ktbarrett, line 434: I don't really understand this file and the simulator.py file. This looks like a 600 line duplication of effort?

src/cocotb_tools/pytest/hookspecs.py
```diff
@@ -0,0 +1,81 @@
...
+
+from __future__ import annotations
+
+from pytest import FixtureRequest, hookspec
+
+from cocotb_tools.pytest.dut import Dut
+
+
+@hookspec(firstresult=True)
+def pytest_cocotb_dut_create(request: FixtureRequest) -> Dut | None:
```
ktbarrett, line 15: The Result type should be arbitrary, it doesn't have to be `DUT`. I'm wondering if this is the cause of the massively duplicated bit of code above.

If users use a different build and simulation system, they are on their own.

This is another reason to split the plugin in two, so we have to attempt to support options for adding elab args when people are using totally different build systems.

This approach is too framework-y. Less is more.

ktbarrett (commented):
So this is another massive change that implements 100 features. It's not very reviewable and that has an effect architecturally.

I think what needs to happen here is everything that's related to runner functionality: all the markers, the `DUT` abstraction, to be *deleted* (or saved for later).

I want to see a focused single-responsibility plugin come in right now where in the fixtures you simply call `Runner.build` and `Runner.test`. This is how custom build systems will likely integrate.

Once we have that working, we can work on adding features gradually and purposefully.

tymonx:
> So this is another massive change that implements 100 features. It's not very reviewable and that has an effect architecturally.

I agree. I made this mistake that I just copied files from `src/cocotb_tools/_pytest/` to `src/cocotb_tools/pytest`. And because of that it looks like a massive PR. I will create a new PR or rebase this one but with changes on the `src/cocotb_tools/_pytest/`. Then after all reviews, we could move it to `src/cocotb_tools/pytest`.

> I want to see a focused single-responsibility plugin come in right now where in the fixtures you simply call Runner.build and Runner.test. This is how custom build systems will likely integrate.

This is something I can do. PR/2 allows to use existing cocotb runners + built-in regression manager. If I will create/rebase a PR with these changes on the `src/cocotb_tools/_pytest/` (not by copying files) then the diff itself should be relative small and reviewable.

I will prepare something in this week.

tymonx:
Closing this PR. I will create a new PR that will show using fixtures to build HDL project and run HDL simulation https://github.com/cocotb/cocotb/issues/5425#issuecomment-4968540911

PR #5597, opened 2026-06-05
Fix LogicArray equality list values

### LINKED ISSUE #5596: LogicArray equality raises TypeError for lists containing unsupported objects
Hi! When playing around with logic arrays I discovered that the LogicArray equality operator can raise a TypeError when comparing a LogicArray to a Python list containing an unsupported object type.

For example:
```
LogicArray("0101") == [object()]
```
Will output a TypeError while I believe we want to just output false. I think its of note that logic array already supports this.
 For example:
```
LogicArray("0101") == [7, "f", dict]
```
Will return False. This seems like an unaccounted for inconsistency that could probably be handled by catching both TypeError and ValueError when attempting conversion.

luddefell:
Fixes #5596

Catch `TypeError` as well as `ValueError` when normalizing list and tuple
operands in Logic Array's equality operator

Now all unsupported values compare unequal instead
of raising.

Added a regression test for the unsupported-object case.

### DIFF
--- a/src/cocotb/types/_logic_array.py
+++ b/src/cocotb/types/_logic_array.py
@@ -639,6 +639,8 @@ class LogicArray(AbstractMutableArray[Logic]):
                 other = LogicArray(other)
             except ValueError:
                 return False
+            except TypeError:
+                return NotImplemented
             return self == other
         else:
             return NotImplemented
--- a/docs/source/newsfragments/5596.bugfix.rst
+++ b/docs/source/newsfragments/5596.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed :class:`~cocotb.types.LogicArray` equality against list and tuple values containing unsupported object types.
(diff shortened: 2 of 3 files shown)


src/cocotb/types/_logic_array.py
```diff
@@ -637,7 +637,7 @@ def __eq__(
         elif isinstance(other, (list, tuple)):
             try:
                 other = LogicArray(other)
-            except ValueError:
+            except (TypeError, ValueError):
```
ktbarrett: Technically TypeError means "unsupported type" not "supported but invalid." So we should return NotImplemented in that case instead.

Ultimately that ends up being `False` but allows users some flexibility to override if they want.

luddefell: Thanks, that makes sense. I amended and pushed the PR with split handling. Let me know if there is anything else!

ktbarrett (approved):

PR #5598, opened 2026-06-07
fix: make example that uses ip

### LINKED ISSUE #2646: Make Example that uses IP
Several users have asked for an example of how to use the makefiles in projects that contain precompiled and/or encrypted IP cores or libraries. There should be a simple example. It will likely not work for them, but could act as a template.

Alternatively, documentation or an explicit API for these objects would also help.

ColdWeather-gpt:
Fixes #2646

Improved documentation.

This PR addresses issue #2646.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,10 @@ compile_commands.*.json
 
 # CI artifacts
 .ccache/
+__pycache__/
+*.egg-info/
+.venv/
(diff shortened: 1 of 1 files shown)


ktbarrett:
If the PR is still in work, please put it in draft. When you are ready hit "ready for review."

Issue #5602, opened 2026-06-09
Large number of coroutines starting in a single time step result in performance issues


jakelongo:
cocotb version: 2.0.1
python: 3.13.13

This issue came up when attempting to move from cocotb 1.9.2 -> 2.0.1. Moving all the the `cocotb.start(...)` calls to `cocotb.start_soon(...)`.

The project contains a large tree of tasks which kick off at time 0, this results in roughly 500k coroutines being scheduled at once (arguably, this shouldn't happen either but that's not a cocotb problem, that's an us problem!).

Using a combo of callgrind and py-spy, it pointed to the schedular code in two places, adding the tasks and removing the tasks (although weighted heavily towards the adding tasks).

Adding tasks:

```python
def _schedule_task(self, task):
    ...
    if task in self._scheduled_tasks:
        return
    for tasks in self._trigger2tasks.values():
        if task in tasks:
            return
    self._schedule_task_internal(task)
```

I think this makes it roughly O(n^2) loop on inserting scheduling routines which is why it's the majority of the time spent in our sim.

Worth mentioning this might already mitigated for on master as cocotb has updated this guard so I don't think it's an issue on the master branch (that said I haven't actually run it against main yet).

```python
    def _ensure_started(self) -> None:
        state = self._state
        if state is _TaskState.UNSTARTED:
            if debug.debug:
                self._log.debug("Starting %r", self)
            self._schedule_resume()
.
.
```

Removing tasks: 

```python
def _task_done_callback(self, task):
    self.tasks.remove(task)
```

`list.remove(...)` adds another O(N) penalty per task to our sim which sums to a O(N^2) over all tasks. I trialled fixing this via monkey patching the internals to store tasks as a `dict[task] = None` which makes removing entries O(1), it feels kind of hacky to do but ultimately fits the bill. I've attached a cocotb standalone reproducer that demonstrates the issue:

```bash
$ python3 repro.py              
cocotb version: 2.0.1
python:         3.13.13

Schedule issue:
         N     seconds                   ratio vs prev
      3750      0.2727                               -
      7500      1.0808                           3.96x
     15000      4.4974                           4.16x
     30000     17.1861                           3.82x

Remove issue:
         N     seconds                   ratio vs prev
      3750      0.0006                               -
      7500      0.0024                           4.03x
     15000      0.0092                           3.90x
     30000      0.0360                           3.91x

$ python3 repro.py --fix
cocotb version: 2.0.1
python:         3.13.13
[...]

PR #5603, opened 2026-06-10
Change TestManager _tasks attribute from list to dict to speed up complete callback

jakelongo:
Update the TaskManager _tasks attribute from a list to a dictionary. This speeds up the teardown of a task once it has completed and attempts to remove itself form the task list.

The performance issues encountered were discussed in this issue: #5602

In the instance where we are running many hundreds of thousands of tasks in a single time step (and a large portion of those tasks are also completing in that time step). The teardown time of those tasks starts to become noticeable. Moving to a dict means we can perform the removal based on the hash without any iteration/comparison and is O(1) time.

This could probably do with more thorough performance unit test though.....

### DIFF
--- a/tests/benchmarks/test_task_churn_perf/test_task_churn_perf.py
+++ b/tests/benchmarks/test_task_churn_perf/test_task_churn_perf.py
@@ -1,0 +1,52 @@
+# Copyright cocotb contributors
+# Licensed under the Revised BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-3-Clause
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+from cocotb_tools.runner import get_runner
+
+THIS_DIR = Path(__file__).resolve().parent
+
+
+def build_and_run(benchmark, scenario: str) -> None:
+    if str(THIS_DIR) not in sys.path:
+        sys.path.append(str(THIS_DIR))
+
+    runner = get_runner("icarus")
+
+    runner.build(
+        hdl_toplevel="task_churn_perf_top",
+        sources=[THIS_DIR / "task_churn_perf_top.sv"],
+        build_dir="sim_build",
+    )
+
+    @benchmark
+    def run_test() -> None:
+        runner.test(
+            hdl_toplevel="task_churn_perf_top",
+            test_module="task_churn_performance_tests",
+            test_filter=scenario,
+        )
+
+
+def test_task_churn_typical(benchmark) -> None:
+    build_and_run(benchmark, "typical")
+
+
+def test_task_churn_churn_random(benchmark) -> None:
+    build_and_run(benchmark, "churn_random")
+
+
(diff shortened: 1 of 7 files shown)


ktbarrett:
Can you add a newsfragment? See `docs/source/newsfragments/README.rst` for details.

jakelongo:
I'll look at adding that. Did you want me to add some kind of benchmark test for it? I've found the `benchmark.yml` which specifies the 120% threshold so I guess this will catch regressions on that going forward

ktbarrett:
We definitely need more benchmarks. If you want to make some benchmarks that stresses the scheduler in different ways that'd be great, but I won't block the PR if you don't have time for it.

Some scenarios, but feel free to do what you want and think up others.

* Typical: There are a number of tasks started at the beginning of the test and live until the end. While the test runs there are a number of tasks that start and complete in fairly short order. This should be "typical" number of tasks so a couple dozen resident and a couple dozen (at most) transient seems reasonable. We can just run the transient set a number of times until we see a difference.
* Your case / stress 1: Lots of Tasks, random completion time.
* stress 2: Lots of "resident" Tasks that are killed at the end of test. Basically just start up a half million and let the test end.

Ideally those would be separate benchmarks so we can track them separately.

jakelongo:
I've added some unit tests which should cover those cases.

Before/after tests show that it's faster in all the cases aside from the 'typical' test which is within 10% of the baseline (fractions of a second slower).

tests/benchmarks/test_task_churn_perf/test_task_churn_perf.py
```diff
@@ -0,0 +1,52 @@
...
+def build_and_run(benchmark, scenario: str) -> None:
+    if str(THIS_DIR) not in sys.path:
+        sys.path.append(str(THIS_DIR))
+
+    runner = get_runner("icarus")
+
+    runner.build(
+        hdl_toplevel="task_churn_perf_top",
+        sources=[THIS_DIR / "task_churn_perf_top.sv"],
+        build_dir="sim_build",
```
ktbarrett: You'll definitely need to run these in different directories. Icarus does builds as `sim.vvp` and then Runner checks if this file exists and then doesn't re-build. It's an awful quirk of the Runner that will be addressed as soon as 2.1 is out.

ktbarrett:
Testing on my machine showed the "typical" case was also slightly faster (probably not noticeable), but the rest of the tests much faster. I'm ready to merge once the tests pass.

tests/benchmarks/test_task_churn_perf/test_task_churn_perf.py
ktbarrett: I fixed this.

ktbarrett (approved):

ktbarrett:
@jakelongo Thanks again!

jakelongo:
Ah apologies, didn't get a chance to get to a computer over the weekend. Thank you for picking it up and updating/merging!

PR #5605, opened 2026-06-11
feat(pytest-plugin-v2): 2/N: add support for built-in cocotb regression manager, dut fixtures vs runners mode

tymonx:
Implementing a new API for pytest plugin proposed in this issue: https://github.com/cocotb/cocotb/issues/5425, part 2

PR on top of the https://github.com/cocotb/cocotb/pull/5595 that must be merged first. That's why, it is now in the `DRAFT` state.

Addressing the biggest drawbacks about using plugin that was raised by @ktbarrett in this comment https://github.com/cocotb/cocotb/issues/5425#issuecomment-4083743585

> I'm down for everything mentioned. I had originally envisioned this as fixture as build which would slot into dut argument. The only reservation I have is that it would be nice to support those existing projects that are already using pytest with Runner tests (like old cocotb-test). While the new way would be better, the way the plugin is written now is you get all or nothing so there's no way to gradually move to the new way in old projects already using pytest if we don't support that flow at all.

* Pytest plugin will work with existing cocotb tests that are using classic cocotb runners defined in separate test functions. No modifications are needed on user side in their current code base
* By default, based on information about HDL design defined in `Dut` fixtures and/or set from `@pytest.mark.cocotb_*` markers, plugin implicitly creates a test function that will build HDL design and run HDL simulation. This is delegated to the custom `pytest_cocotb_dut_run` plugin hook (see the previous PR 1/N). Default implementation is using cocotb runners. Can be easily replaced with custom build system. Implicit creation of "runners" can be disabled by using the new plugin option `--cocotb-with-user-runners` that can be set from environment variable, pytest configuration file (`pyproject.toml`, `pytest.toml`, ...) or pytest command line. Addressing @imphil suggestions from this comment https://github.com/cocotb/cocotb/issues/5425#issue-4095877078
* Plugin will also work with built-in cocotb regression manager. This can be enabled with the `--cocotb-regression-manager=cocotb` option. In this mode, plugin will normally collect and report cocotb tests. Cocotb tests results are parsed from generated JUnit XML file `<pytest-test-function-name>.result.xml` and populate back as normal pytest test reports to pytest main process. This feature is encapsulated in the `cocotb_tools.pytest._results.ResultsXML` class. Note: this mode doesn't allow to use pytest in fullness within HDL simulation process and with cocotb tests. For that, the `--cocotb-regression-manager=pytest` must be used

Added some tests that are covering all above scenarios. To run them:

```plaintext
pytest ./tests/pytest_plugin/test_manager_cocotb.py
```

The code coverage is 100%.

This PR is also removing possibility to set the `random_seed` from pytest markers or `Dut` attributes. It is only possible to set it from an environment variable, a pytest configuration file or command line. The `random_seed` cannot be part of build context otherwise it will affect the hash of the DUT fixture and name of build directory on each run. 

Next PRs:
* 3/N: Moving implementation of `src/cocotb_tools/_pytest/_regression.py` back to public module `src/cocotb_tools/pytest/`. This will allow to use the `--cocotb-regression-manager=pytest` mode that will be set to default. This mode will allow to use pytest capabilities in fullness within HDL simulations and with cocotb tests
* 4/N: Updating the documentation about the pytest plugin. Reflecting new changes and improvements, removing info about cocotb runners, in favor of `Dut` fixtures and/or `@pytest.mark.cocotb_*` markers
* 5/N: More unit tests focusing on real use cases


tymonx:
Closing in favor of https://github.com/cocotb/cocotb/pull/5595

PR #5606, opened 2026-06-11
Fix test deps

ktbarrett:
This is currently breaking regressions.

* Moved deps used by the regression to test_common
* Fixed a missing dep. release tests need all cocotb deps copied, when pytest was added this wasn't done. It works because we specify a higher version of pytest in test_common for RaisesExc and RaisesGroup testing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -316,15 +316,19 @@ release_build_sdist = [
     "build",
 ]
 test_common = [
-    "pytest>=8.4"  # need RaisesExc and RaisesGroup
+    # Needed for RaisesExc and RaisesGroup testing.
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -316,15 +316,19 @@ release_build_sdist = [
     "build",
 ]
 test_common = [
-    "pytest>=8.4"  # need RaisesExc and RaisesGroup
+    # Need RaisesExc and RaisesGroup testing.
```
cmarqu:
```suggestion
    # Needed for RaisesExc and RaisesGroup testing.
```

cmarqu (approved):

PR #5607, opened 2026-06-11
Replace `COCOTB_LIBRARY_COVERAGE` with `coverage`'s built-in subprocess patch

ktbarrett:
Depends on #5606.

Now that we have `coverage.run.patch=["subprocess"]` we don't really need `COCOTB_LIBRARY_COVERAGE` anymore. So replace it and instrument up `COVERAGE_PROCESS_START` to do the same.

Also I split out a `clean` nox session to run `make clean` and ensure coverage and results files are cleaned up.


ktbarrett:
Some collections covered branches, others didn't. That's the merge failure. We need to move the pytest-cov options into pyproject.toml.

PR #5614, opened 2026-06-15
Improve `resolve()` and `is_resolvable`

### LINKED ISSUE #5611: `resolve()` semantic improvement
What we have as resolving is not very well respected piece of code (by me). This was mostly because the automatic conversion bothers me and there were some conceptual holes in the logic. But I think we can birth a better design with some slight adjustments.

## Where we are now

We have `COCOTB_RESOLVE_X` and a global default resolution function `cocotb.types._resolve.RESOLVE_X`.

`Logic` and `LogicArray` both have `resolve()` functions which take a method, or default to the global default, and return a `LogicArray`.

They also have a `is_resolvable()` method which is hardcoded to default to the WEAK method semantics.

## Where we could be

We add a global default resolution method which has a getter and setter. `COCOTB_RESOLVE_X` is a regression-wide initializer. This should default to `WEAK` semantics.

`Logic` and `LogicArray` `resolve()` stays largely the same in that it uses the given method or the global default if no method is given. However it defaults to return a `Bit` and `BitArray` respectively. `Bit` and `BitArray` have guaranteed non-failing `int` and `bool` casts.

`is_resolvable` is modified to take a resolution method to check for resolvability, or default to the global when no argument is given. Now that the global default is WEAK, this does not change behavior. But it more better states "`resolve()` will not fail".

## What changes

Since `resolve()` should return a `Bit` or `BitArray`, `WEAK` should now *throw* on `W` instead of return `X`, since that isn't a valid `Bit`.

`is_resolvable()` with no argument behavior now changes with the global resolution function. This is an improvement across the board IMO.

## What's in the way

There is no `BitArray`... For now we can return a `LogicArray` that's simply guaranteed to be 0/1 and maybe update that to be a `BitArray` if/when we get around to implementing that type.

Risks are fairly low. We only added the WEAK resolve method in 2.0, so there's only one minor version where behavior changes. And its unlikely that behavior is currently relied upon. Who asks for `resolve()` and wants an `X` back?

ktbarrett:
These contain some breaking changes, but only to code that was added in 2.0 and in the smallest possible way.

It's designed to make resolve functionality more useful and self-consistent.

* Makes "weak" mode fail on U, X, W, Z, and - instead of mapping them
* Because of that Logic.resolve() now returns a Bit
* Added a user settable global default resolve method
* Made resolve() argument optional and use the global default resolver
* Made is_resolvable property use the global default resolver instead of being hardcoded to use "weak" semantics.
* Added ability to call is_resolvable with a resolver method
* Make COCOTB_RESOLVE_X set the initial global default resolver
* Made "weak" the default global resolve method
* made Logic and LogicArray `bool` cast respect the global default resolver instead of the quirky "use Verilog style checks only when COCOTB_RESOLVE_X is set."

Closes #5611.

- [ ] newsfrags
- [ ] Use Verilog `__bool__` semantics for RANDOM since it's nonsense that it can randomly be `True` or `False`. `ZEROS` and `ONES` should respect `resolve` as they are essentially forcing functions (`__bool__` is False for X or True for X). `WEAK` and `ERROR` fail, which is reasonable.

### DIFF
--- a/src/cocotb/types/_logic_array.py
+++ b/src/cocotb/types/_logic_array.py
@@ -20,7 +20,7 @@ from cocotb.types._abstract_array import AbstractMutableArray
 from cocotb.types._indexing import IndexingChangedWarning
 from cocotb.types._logic import Logic, LogicConstructibleT
 from cocotb.types._range import Range
-from cocotb.types._resolve import RESOLVE_X, ResolverLiteral, get_str_resolver
+from cocotb.types._resolve import IsResolvableLogicArray, ResolverLiteral, resolve
 
 if sys.version_info >= (3, 10):
     from typing import TypeAlias
@@ -337,19 +337,14 @@ class LogicArray(AbstractMutableArray[Logic]):
             # May convert list to str before converting to int.
             value_as_str = self._get_str()
 
-            # always resolve L and H to 0 and 1
-            value_as_str = value_as_str.translate(_resolve_lh_table)
-
             try:
-                self._value_as_int = int(value_as_str, 2)
+                resolved_str = resolve(value_as_str)
             except ValueError:
-                if RESOLVE_X is None:
-                    raise ValueError(
-                        f"Can't convert {type(self).__qualname__} to int: it contains non-0/1 values"
-                    ) from None
-                else:
-                    value_as_str = RESOLVE_X(value_as_str)
-                    return int(value_as_str, 2)
+                raise ValueError(
+                    f"Can't convert {type(self).__qualname__} to int: it contains non-0/1 values"
+                ) from None
+            else:
+                self._value_as_int = int(resolved_str, 2)
 
         return self._value_as_int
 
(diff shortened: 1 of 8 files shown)


src/cocotb/types/_logic_array.py
```diff
@@ -962,51 +972,34 @@ def __xor__(self, other: LogicArray) -> LogicArray:
...
+
+        .. versionchanged:: 2.1
+            The *resolver* parameter is now optional.
+            The global default resolver will be used if it is not provided.
+
+        .. versionchanged:: 2.1
+            The ``weak`` resolver now throws :exc:`ValueError` on ``W``, ``X``, ``Z``, and ``-`` values.
         """
-        return LogicArray(get_str_resolver(resolver)(str(self)), self.range)
+        return LogicArray(resolve(self._get_str(), resolver), self._range)
```
ktbarrett: to do `bool` and `int` casts here, we still have to go through the resolver *again*. So we should also compute the integer and fill that in as well so we can skip going through the resolver again.

PR #5616, opened 2026-06-15
Improve performance by not writing deselected tests to results.xml

### LINKED ISSUE #5615: XUnitReporter change massive slow down
#5219 caused a massive slow down on the parametrize perf benchmark. This was originally added resolving #5137. So I doubt this will make the reporter of that issue happy. We should see if there's a way to resolve this before the release.

ktbarrett:
Closes #5615. I confirmed that pytest does not write deselected tests to its output json file so there's no reason we need to.

### DIFF
--- a/docs/source/newsfragments/5616.change.rst
+++ b/docs/source/newsfragments/5616.change.rst
@@ -1,0 +1,1 @@
+Tests not included in a regression when using :envvar:`COCOTB_TEST_FILTER` no longer appear in ``results.xml``.
--- a/src/cocotb/_decorators.py
+++ b/src/cocotb/_decorators.py
@@ -96,7 +96,6 @@ class Test:
         self.expect_error = expect_error
         self.skip = skip
         self.stage = stage
-        self.included = not skip
 
     @property
     def fullname(self) -> str:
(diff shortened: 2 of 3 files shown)


imphil (approved):
Fine with me, but let's wait for a  day or two if there's a way to retain functionality and also improve performance, as currently discussed in https://github.com/cocotb/cocotb/issues/5615.

ktbarrett:
@imphil Looks like the other PR hasn't improved performance much more than a current stddev. This still should come in.

tymonx:
This PR will also help the pytest plugin v2 https://github.com/cocotb/cocotb/pull/5605 that is now supporting the built-in regression manager. It is using generated JUnit XML file to report cocotb tests back to pytest. This is done by the [src/cocotb_tools/pytest/_results.py::ResultsXML](https://github.com/tymonx/cocotb/blob/e084efdb00f9dabc2244f90ead80d2ef891ce465/src/cocotb_tools/pytest/_results.py) class.

ktbarrett:
@tymonx Our (kinda crappy) performance metrics if you were unaware. https://cocotb.github.io/cocotb-benchmark-results/dev/bench/.

PR #5617, opened 2026-06-15
Start compound Trigger waiter Tasks eagerly

### LINKED ISSUE #5594: Event not triggered when Event.set() , Event.wait() called at same time step. But works fine in cocotb 1.9
The following synchronization works fine in cocotb 1.9 but timesout in cocotb 2.0.1
```python
import cocotb
from cocotb.triggers import Event,Timer,First
async def adding_to_que(que,e1:Event,e2:Event,ea:Event):
    await Timer(1)
    que.append(1)
    e1.set()
    e1.clear()
    await ea.wait()
    que.append(2)
    e2.set()
    e2.clear()
``` 
        
```python
@cocotb.test(timeout_time=1, timeout_unit='us', skip=False)
async def check_events(dut): 
    que = []
    e1 = Event()
    e2 = Event()
    ea = Event()
    cocotb.start_soon(adding_to_que(que,e1,e2,ea))
    await First(e1.wait(),e2.wait())
    ea.set()
    ea.clear()
    cocotb.log.info(f"run 0 val = {que.pop(0)=}")
    await First(e1.wait(),e2.wait())
    cocotb.log.info(f"run 1 val = {que.pop(0)=}")
``` 
Result 1.9
```
     0.00ns INFO     cocotb                             Running on ModelSim for Questa-64 version 2023.2 2023.10
     0.00ns INFO     cocotb                             Running tests with cocotb v1.9.0 from #######
     0.00ns INFO     cocotb                             Seeding Python random module with 1780617421
     0.00ns INFO     cocotb.regression                  Found test fifo_basic_test.check_events
     0.00ns INFO     cocotb.regression                  running check_events (1/1)
     0.10ns INFO     cocotb                             run 0 val = que.pop(0)=1
     0.10ns INFO     cocotb                             run 1 val = que.pop(0)=2
     0.20ns INFO     cocotb.regression                  check_events passed
``` 
Result 2.01
```
     0.00ns INFO     cocotb                             Running on ModelSim for Questa-64 version 2023.2 2023.10
     0.00ns INFO     cocotb                             Seeding Python random module with 1780617440
     0.00ns INFO     cocotb                             Initialized cocotb v2.0.1 from #######
     0.00ns INFO     cocotb                             Running tests
     0.00ns INFO     cocotb.regression                  running fifo_basic_test.check_events (1/1)
     0.10ns INFO     test                               run 0 val = que.pop(0)=1
  1000.00ns WARNING  ..b.Test check_events.check_events Traceback (most recent call last):
                                                          File "#######/cocotb/regression.py", line 378, in func
                                                            await with_timeout(f(*args, **kwargs), timeout, self._test.timeout_unit)
                                                          File "#######/_extended_awaitables.py", line 417, in with_timeout
                                                            raise SimTimeoutError
                                                        cocotb.triggers.SimTimeoutError
  1000.00ns WARNING  cocotb.regression                  fifo_basic_test.check_events failed
``` 

Checking the implementation of Event in 1.9 this condition is accounted, but the same is not there in verion >= 2.0
cocotb 2.01 `src\cocotb\_base_triggers.py`
```python
    def wait(self) -> Trigger:
        """Block the current Task until the Event is set.

        If the event has already been set, the trigger will fire immediately.

        To set the Event call :meth:`set`.
        To reset the Event (and enable the use of :meth:`wait` again),
        call :meth:`clear`.
        """
        return self._event
```
cocotb 1.9 `cocotb\triggers.py`
```python
def wait(self):
        """Get a trigger which fires when another coroutine sets the event.

        If the event has already been set, the trigger will fire immediately.

        To reset the event (and enable the use of ``wait`` again),
        :meth:`clear` should be called.
        """
        if self._fired:
            return NullTrigger(name=f"{str(self)}.wait()")
        return _Event(self)
```

ktbarrett:
Depends on #5645.

The code `await First(e1.wait(), e2.wait())` has a problem: the actual Events are not waited immediately despite there being the `await` on the call. All passed in triggers are awaited on Tasks which are `start_soon`. This creates a race between already scheduled Tasks and the waiters in the First. Starting the waiter Tasks eagerly resolves the issue as each waiter gets to their first `await` before any other scheduled Task resumes. This closes #5594.

`start_eagerly` is made public and intended for users writing compound Triggers. We use `start_eagerly` for the implementation of `wait` that underlies all of the concurrent waiters in the repo. Users would use this for their own similar triggers.

### DIFF
--- a/docs/source/newsfragments/5617.bugfix.rst
+++ b/docs/source/newsfragments/5617.bugfix.rst
@@ -1,0 +1,1 @@
+:class:`~cocotb.triggers.Combine` and :class:`~cocotb.triggers.First` now schedule their waiter tasks eagerly on :keyword:`await` instead of starting "soon." This fixes a class of races between the waiter :term:`tasks <task>` and any already scheduled :term:`!tasks`, bringing the scheduler back towards 1.9's behavior.
--- a/docs/source/newsfragments/5617.feature.rst
+++ b/docs/source/newsfragments/5617.feature.rst
@@ -1,0 +1,1 @@
+Added :meth:`.Task.start_eagerly` to start a :class:`.Task` immediately (in the same stack frame as the caller).
--- a/src/cocotb/_concurrent_waiters.py
+++ b/src/cocotb/_concurrent_waiters.py
@@ -8,7 +8,6 @@ from collections.abc import Iterable
 from enum import Enum, auto
 from typing import TYPE_CHECKING, Any, Callable, Literal, TypeVar, overload
 
-import cocotb
 from cocotb._base_triggers import _InternalEvent
 from cocotb.task import Task
 
@@ -117,7 +116,7 @@ async def _wait(
 
     for task in waiters:
         task._add_done_callback(done_callback)
-        cocotb.start_soon(task)
+        task.start_eagerly()
 
     try:
         await done
--- a/src/cocotb/task.py
+++ b/src/cocotb/task.py
@@ -230,7 +230,20 @@ class Task(Generic[ResultType]):
         else:
             raise RuntimeError("Task in unknown state")
 
+    def start_eagerly(self) -> None:
+        """Starts an unstarted Task immediately.
+
+        Raises:
+            RuntimeError: If the Task has already started.
+
+        .. versionadded:: 2.1
+        """
+        if self._state is not _TaskState.UNSTARTED:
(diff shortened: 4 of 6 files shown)


ktbarrett:
- [x] Use `_InternalEvent` to get debug print tree working for `wait()` (Addressed in #5624)
- [x] Handle cancellation better in `wait()` (Addressed in #5623)

PR #5623, opened 2026-06-15
Remove `wait()`, improve `gather()` and `select()`

ktbarrett:
`asyncio.wait()` has a quirky interface which makes it unsuitable to be used to implement `asyncio.gather()`. It also has inferior semantics, so there isn't much point in providing it in cocotb as it is. The version that cocotb adds is resultantly quite a bit different. Because our version is different and `gather` and `select` provide *all* of the functionality of `wait`, we can just remove `wait`.

This commit also improves behavior when children are cancelled by other tasks (likely by them awaiting other cancelled objects). Previously, these cancellations were ignored, but now they cause shutdown and are propagated.

Some other small improvements include:
* `return_exceptions` on `select` was missing the 's'.
* `select` now actually returns the first finishing child rather then the first in the list marked as complete by the time the scheduler switches back to the parent task.
* adds tests for more scenarios.
* `return_exceptions=True` on tasks that are cancelled actually returns the `CancelledError` thrown into the task instead of a new empty instance.

This changes stuff introduced in 2.1, so no newsfrags.

### DIFF
--- a/src/cocotb/_concurrent_waiters.py
+++ b/src/cocotb/_concurrent_waiters.py
@@ -253,6 +253,7 @@ async def gather(
 
     When *return_exceptions* is ``False``, if any *awaitable* results in an exception or is cancelled,
     the remaining *awaitables* are cancelled and the exception or :exc:`~asyncio.CancelledError` is re-raised.
+    This does not cancel :class:`~cocotb.task.Task`\ s passed as arguments, only :term:`coroutines <coroutine>`,  :term:`triggers <trigger>`, or other user-defined :term:`!awaitables`.
     When *return_exceptions* is ``True``, all exceptions and cancellations are treated as successful results and returned in the resulting tuple.
     Control returns to the caller after all *awaitables* have completed and/or been cancelled.
 
--- a/src/cocotb/_concurrent_waiters.py
+++ b/src/cocotb/_concurrent_waiters.py
@@ -150,12 +150,13 @@ async def select(
 
     Regardless of the value of *return_exceptions*, after the first *awaitable* completes, whether it was cancelled, resulted in an exception, or resulted in a value,
     the remaining *awaitables* are cancelled.
+    This does not cancel :class:`~cocotb.task.Task`\ s passed as arguments, only :term:`coroutines <coroutine>`,  :term:`triggers <trigger>`, or other user-defined :term:`!awaitables`.
     Control returns to the caller after all *awaitables* have completed and/or been cancelled.
 
     Regardless of the value of *return_exceptions*, if the task awaiting a call to this function is cancelled before completion,
     all *awaitables* which have not yet completed are cancelled and :exc:`!CancelledError` is raised.
 
-    Is it possible for multiple *awaitables* to complete at the same time,
+    It is possible for multiple *awaitables* to complete at the same time,
     however due to how the cocotb scheduler works, only one will complete *first*.
     It is good practice to avoid relying on the order of completion of multiple *awaitables* which complete at the same time.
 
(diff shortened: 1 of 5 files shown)


src/cocotb/_concurrent_waiters.py
```diff
@@ -157,46 +118,50 @@ def done_callback(task: Task[Any]) -> None:
...
-    This does not cancel :class:`~cocotb.task.Task`\ s passed as arguments,
-    only the internal waiter tasks.
+    Regardless of the value of *return_exceptions*, after the first *awaitable* completes, whether it was cancelled, resulted in an exception, or resulted in a value,
+    the remaining *awaitables* are cancelled.
+    Control returns to the caller after all *awaitables* have completed and/or been cancelled.
+
+    Regardless of the value of *return_exceptions*, if the task awaiting a call to this function is cancelled before completion,
+    all *awaitables* which have not yet completed are cancelled and :exc:`!CancelledError` is raised.
+
+    Is it possible for multiple *awaitables* to complete at the same time,
```
marlonjames:
```suggestion
    It is possible for multiple *awaitables* to complete at the same time,
```

src/cocotb/_concurrent_waiters.py
```diff
@@ -157,46 +118,50 @@ def done_callback(task: Task[Any]) -> None:
...
-    *awaitables: Awaitable[T], return_exception: bool = False
+    *awaitables: Awaitable[T], return_exceptions: bool = False
 ) -> tuple[int, T | BaseException]:
     r"""Await on all given *awaitables* concurrently and return the index and result of the first to complete.
 
-    After the first *awaitable* completes, the remaining waiter tasks are cancelled.
-    This does not cancel :class:`~cocotb.task.Task`\ s passed as arguments,
-    only the internal waiter tasks.
+    Regardless of the value of *return_exceptions*, after the first *awaitable* completes, whether it was cancelled, resulted in an exception, or resulted in a value,
+    the remaining *awaitables* are cancelled.
```
marlonjames, line 152: Is this true? I think it still only cancels the waiter tasks, and having some note about not cancelling Tasks passed in is useful.

ktbarrett: It doesn't cancel the awaitables, but the waiter Tasks. There is no way to cancel awaitables. This does save from having to explain that actually there are waiter tasks in there.

Will add the note about passed in Tasks.

ktbarrett, line 152:
```suggestion
    the remaining *awaitables* are cancelled.
    This does not cancel :class:`~cocotb.task.Task`\ s passed as arguments, only :term:`coroutines <coroutine>`,  :term:`triggers <trigger>`, or other user-defined :term:`!awaitables`.
```

src/cocotb/_concurrent_waiters.py
```diff
@@ -288,15 +251,20 @@ async def gather(
 ) -> tuple[Any, ...]:
     r"""Await on all given *awaitables* concurrently and return their results once all have completed.
 
-    After the return condition, based on *return_exceptions*, is met, the remaining waiter tasks are cancelled.
-    This does not cancel :class:`~cocotb.task.Task`\ s passed as arguments,
-    only the internal waiter tasks.
+    When *return_exceptions* is ``False``, if any *awaitable* results in an exception or is cancelled,
+    the remaining *awaitables* are cancelled and the exception or :exc:`~asyncio.CancelledError` is re-raised.
```
marlonjames, line 255: For completeness.
```suggestion
    the remaining *awaitables* are cancelled and the exception or :exc:`~asyncio.CancelledError` is re-raised.
    This does not cancel :class:`~cocotb.task.Task`\ s passed as arguments, only :term:`coroutines <coroutine>`,  :term:`triggers <trigger>`, or other user-defined :term:`!awaitables`.
```

marlonjames (approved):

PR #5624, opened 2026-06-15
Implement `Combine` and `First` in terms of `gather` and `select`

ktbarrett:
This also pipes parents into `_InternalEvent` to maintain the existing debugging prints.

I had to change the interface for `_InternalEvent` to take a repr function instead of a parent object, as `gather` and `select` are not classes.

Depends on #5623.

### DIFF
--- a/tests/test_cases/test_cocotb/test_first_combine.py
+++ b/tests/test_cases/test_cocotb/test_first_combine.py
@@ -58,8 +58,9 @@ async def test_First_unfired_triggers_killed_on_exception(_) -> None:
         raise ValueError("I am a failure")
 
     with pytest.raises(ValueError), pytest.warns(DeprecationWarning):
-        await First(cocotb.start_soon(Task(fails())), *triggers)
+        await First(Task(fails()), *triggers)
 
+    # test all triggers were unprimed
     for t in triggers:
         assert t.primed == 1
         assert t.unprimed == 1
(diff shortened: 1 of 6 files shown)


tests/test_cases/test_cocotb/test_first_combine.py
```diff
@@ -429,17 +432,7 @@ async def wait_on_tasks() -> None:
...
 
-    # check other tasks haven't been cancelled and the Combine isn't finished.
-    await Timer(1, "ns")
-    assert not waiter_task.done()
-    for t in tasks[1:]:
-        assert not t.done()
-
-    # cancel remaining tasks
-    for t in tasks[1:]:
-        t.cancel()
```
ktbarrett, line 440: There is no longer a "pending" state, as the waiters start immediately.

tests/test_cases/test_cocotb/test_first_combine.py
```diff
@@ -75,8 +76,9 @@ async def fails() -> None:
         raise ValueError("I am a failure")
 
     with pytest.raises(ValueError), pytest.warns(DeprecationWarning):
-        await Combine(cocotb.start_soon(Task(fails())), *triggers)
```
ktbarrett, line 78: This only worked before by miracle of scheduling order. External Tasks are not cancelled, so this will fail because an unwatched `cocotb.start_soon` coroutine raised an error after the `Combine` or `First` "finishes".

I think this better highlights how dangerous the "if being watched" behavior is. It's very dependent on scheduling order.

marlonjames (approved):

PR #5625, opened 2026-06-15
Add `Task.start_soon`

ktbarrett:
This adds `Task.start_soon` method which allows users to schedule a Task to start after they may have created an unstarted one using `cocotb.create_task` or by directly instantiating `Task`. This is also used internally for exactly those cases.

- [x] update newsfrag


marlonjames (commented):
We should probably have at least one test that exercises this.

ktbarrett:
Reminder: squash and reword commit.

marlonjames (approved):

PR #5627, opened 2026-06-15
perf(xunit): improve XUnit Reporter performance

tymonx:
Context:
* https://github.com/cocotb/cocotb/issues/5615
* https://github.com/cocotb/cocotb/pull/5616

Benchmark test running inside container image with Icarus 13.0, Python 3.12.13 and pytest 9.0.3:

```plaintext
pytest -c /dev/null tests/benchmarks/test_parameterize_perf
```

On the `master` branch:

```plaintext
--------------------------------------------------- benchmark: 1 tests ---------------------------------------------------
Name (time in s)                      Min      Max     Mean  StdDev   Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------
test_parameterize_perf_icarus     11.4430  12.1796  11.8045  0.3143  11.7731  0.5554       2;0  0.0847       5           1
--------------------------------------------------------------------------------------------------------------------------
```

Using this PR:

```plaintext
------------------------------------------------- benchmark: 1 tests -------------------------------------------------
Name (time in s)                     Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------
test_parameterize_perf_icarus     7.5379  7.9263  7.7207  0.1867  7.6984  0.3638       1;0  0.1295       5           1
----------------------------------------------------------------------------------------------------------------------
```

What changed:
- ~Removed XML indentation~ (cannot because of Makefiles :roll_eyes:)
- Prepared the `ATTACHMENT` block text in the reporter constructor and appending it to the `system-out` only if needed
- Prepared list of file attachments as strings in the reporter constructor, normalized paths (relative)
- Prepared the `relative_to_str` path as a string to replace paths in text to relative ones
- Adding properties using a simple for-loop with `xml_element.append(Element("property", name=name, value=value))`
- Avoided type checking and handling different types if don't needed, `str` rules
- ~Removed escaping invalid characters in "trusted" variables like `classname` or `name`~
- ~Using the `_escape` function to escape invalid characters only on potentially untrusted `text`, especially from user and HDL simulators~
- Checking if Windows/macOS newline is present
- Reducing branching and unnecessary calls where possible

### DIFF
--- a/src/cocotb/_xunit_reporter.py
+++ b/src/cocotb/_xunit_reporter.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 
 import os
 import re
-from collections.abc import Generator, Iterable, Mapping
+from collections.abc import Iterable, Mapping
 from datetime import datetime, timezone
 from pathlib import Path
 from platform import node
@@ -76,6 +76,7 @@ class XUnitReporter:
         name: str = "cocotb tests",
         relative_to: Path | str | None = None,
         default_properties: Mapping[str, Any] | None = None,
+        default_attachments: Iterable[Path | str] | None = None,
     ) -> None:
         """Create new instance of xUnit reporter.
 
@@ -83,6 +84,7 @@ class XUnitReporter:
             name: Name of xUnit reporter. Used as name for the XML test suites root element.
             relative_to: If provided, all reported absolute paths will be converted to relative paths.
             default_properties: Additional common default properties that will be added to all created test cases.
+            default_attachments: Additional common default file attachments that will be added to all created test cases.
         """
         # The root of the XML document
         self._root = Element("testsuites", name=name)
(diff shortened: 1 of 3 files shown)


tymonx:
Fun fact. Cannot remove XML `indent` because Makefiles are using `grep`/`sed` to check something from generated `results.xml` files. Removing `indent` will write a very long single line to XML output file and these checks will not work, causing CI to fail.

Benchmark without removing the XML `indent`:

```plaintext
------------------------------------------------- benchmark: 1 tests -------------------------------------------------
Name (time in s)                     Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------
test_parameterize_perf_icarus     8.1435  8.4267  8.2603  0.1165  8.2809  0.1772       2;0  0.1211       5           1
----------------------------------------------------------------------------------------------------------------------
```

ktbarrett (approved):

ktbarrett:
> Fun fact. Cannot remove XML indent because Makefiles are using grep/sed to check something from generated results.xml files. Removing indent will write a very long single line to XML output file and these checks will not work, causing CI to fail.

Christ alive. Well I'll write that up as a follow-up.

PR #5631, opened 2026-06-16
chore(xunit): remove XML indent

### LINKED ISSUE #5628: Remove `indent` from XUnitReporter
This call scales linearly with the output and only serves to make the output pretty. However, the CI depends upon it by using grep to look for failures rather than using a XML parser (written in Python probably).

tymonx:
Closes https://github.com/cocotb/cocotb/issues/5628

Removing XML `indent` to reduce size of bloated JUnit XML.

One of Makefiles was using `grep` to count number of XML `<skipped/>` elements. This will not work when we remove XML `indent` because everything will be in a single line. We need to improve used regular expression to find a single match of skipped element and print matched pattern in a separate line. This can be achieved by using the `grep -o` option:

```plaintext
       -o, --only-matching
              Print only the matched (non-empty) parts of a matching line, with each such part on a separate output line.
```

### DIFF
--- a/docs/source/newsfragments/5628.change.rst
+++ b/docs/source/newsfragments/5628.change.rst
@@ -1,1 +1,1 @@
-Disabled two-space characters indentation of generated JUnit XML ``results.xml`` files to reduce their size. They are not pretty-formatted anymore. Use external libraries or tools to reformat and reindent generated JUnit XML files. For example, you can use the ``xmllint --format results.xml`` command for that.
+JUnit XML (``results.xml``) files are no longer pretty-printed to reduce file size and improve serialization performance. Use external tools like `xmllint <https://xmllint.com/>`_ if you need to format these files.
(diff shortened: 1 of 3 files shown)


ktbarrett (approved):

ktbarrett:
This deserves a newsfrag. No idea if someone is also depending on it. It should mention xmllint.

tymonx:
> This deserves a newsfrag. No idea if someone is also depending on it. It should mention xmllint.

Done. Added a news fragment.

docs/source/newsfragments/5628.change.rst
```diff
@@ -0,0 +1 @@
+Disabled two-space characters indentation of generated JUnit XML ``results.xml`` files to reduce their size. They are not pretty-formatted anymore. Use external libraries or tools to reformat and reindent generated JUnit XML files. For example, you can use the ``xmllint --format results.xml`` command for that.
```
ktbarrett:
```suggestion
JUnit XML (``results.xml``) files are no longer pretty-printed to reduce file size and improve serialization performance. Use external tools like `xmllint <https://xmllint.com/>`_ if you need to format these files.
```

PR #5637, opened 2026-06-18
Set SONAME for libgpi on Linux

### LINKED ISSUE #5636: Add SONAME to libgpi
Extension (to libgpi) libraries such as libcoconext need a way to provision the transitive libgpi dependency for the user, ideally without much or any input. Since these are distributed in Python packages, RPATH is *not* appropriate; we've learned that the hard way. Another way is to ask the package that provides the library to provision it such that when the library is dynamically loaded the environment is capable of resolving it automatically.

There are two ways to accomplish that on Linux:
* `LD_LIBRARY_PATH`: only works if you set it *before* loading the thing with the libgpi dependency.
* `dlopen("path/to/libgpi.so", RTLD_GLOBAL)`: preload the library before doing the runtime load of the library with the dependency on libgpi.

Point 2 does not work currently. If a downstream has a dependency on ligpi, preloading libgpi with `dlopen` on the full path does not work. The linker keys the library in the global table off the *full* path. So only links against the full path will resolve again.

This is exactly the problem that SONAMEs solve. libgpi (the only public SO we provide) should have this set.

ztysth:
This sets the Linux ELF SONAME for `libgpi.so` to `libgpi.so`.

Downstream extension libraries can depend on libgpi and may preload cocotb's `libgpi.so` with `dlopen(..., RTLD_GLOBAL)` before loading their own library. Without a SONAME, the dynamic linker keys the preloaded library by its full path, so later dependencies on `libgpi.so` may not resolve to the already-loaded library.

The new linker flag is Linux-only. macOS uses install names, which cocotb already sets separately, and Windows uses DLL/import-library mechanisms rather than ELF SONAMEs.

A Linux-only pytest checks the built `libgpi.so` dynamic section with `readelf` and verifies that the SONAME is present.

Fixes #5636.

### DIFF
--- a/cocotb_build_libs.py
+++ b/cocotb_build_libs.py
@@ -320,6 +320,8 @@ class build_ext(_build_ext):
                     # `GLIBCXX_3.4.29' not found (required by
                     # /path/to/libcocotbvhpi_modelsim.so)."
                     ext.extra_link_args += ["-static-libstdc++"]
+                    if lib_name == "libgpi":
+                        ext.extra_link_args += ["-Wl,-soname,libgpi.so"]
 
                 ext.extra_link_args += [f"-Wl,-rpath,{rpath}" for rpath in rpaths]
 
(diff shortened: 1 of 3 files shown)


ktbarrett (approved):

ktbarrett:
Wow! Thanks for this, it really unblocks me. Great turn around time too.

PR #5640, opened 2026-06-19
Future flags

### LINKED ISSUE #5638: Add breaking change feature flag
We should add an environment variable flag that allows cocotb switch into the "new" behavior for the next major version bump when it's feasible to support both the old and new.

This envvar should be called `COCOTB_PREVIEW`. It should turn on if set to `1` and off if set to `0`. Eventually we may turn this into a piecewise switch: `COCOTB_PREVIEW=XFAIL_IN_RESULTS,RESOLVE_ON_HANDLES,WEAK_FAILS_METAVALUES`.

ktbarrett:
Closes #5638.

This introduces `COCOTB_FUTURE` which is an environment variable which will switch on either new experimental features, or enable breaking changes. The intent is that on the next major version these future features will become standard.

Changed behavior will have a deprecation warning and this flag will allow people to opt-in to breaking changes to see what's going to change, which will also silence the deprecation warnings.

It allows us to require users to explicitly opt into experimental features so there's an established contract that such features are subject to change.

### DIFF
--- a/docs/source/library_reference.rst
+++ b/docs/source/library_reference.rst
@@ -932,6 +932,28 @@ Other Runtime Information
 
     .. versionadded:: 2.1
 
+Future Features
+---------------
+
+.. envvar:: COCOTB_FUTURE
+
+    Type: :ref:`env-list`
+
+    A comma-separated list of experimental features or breaking changes that can be enabled or disabled.
+    If this variable is not defined, all experimental features and breaking changes are disabled.
+
+    See :class:`cocotb.future.Future` for a list of available features.
+    The lowercase name of the feature is used to enable it.
+    Setting the value to ``1``, ``true``, or ``yes`` enables all features,
+    while ``0``, ``false``, or ``no`` disables all features.
+
+    .. versionadded:: 2.1
+
+.. automodule:: cocotb.future
+    :members:
+    :member-order: bysource
+    :synopsis: Experimental features or breaking changes that can be enabled or disabled.
+
 Debugging
 ---------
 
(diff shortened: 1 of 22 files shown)


docs/source/library_reference.rst
```diff
@@ -932,6 +932,14 @@ Other Runtime Information
 
     .. versionadded:: 2.1
 
+Futures
+-------
+
```
marlonjames, line 951: Can you add a `.. envvar:: COCOTB_FUTURE` section here so the `:envvar:` in release notes will link here and there's a bit of explanation.

docs/source/library_reference.rst
```diff
@@ -932,6 +932,28 @@ Other Runtime Information
 
     .. versionadded:: 2.1
 
+Future Features
+---------------
+
+.. envvar:: COCOTB_FUTURES
```
marlonjames:
```suggestion
.. envvar:: COCOTB_FUTURE
```

docs/source/library_reference.rst
```diff
@@ -932,6 +932,28 @@ Other Runtime Information
...
+
+    Type: :ref:`env-list`
+
+    A comma-separated list of experimental features or breaking changes that can be enabled or disabled.
+    If this variable is not defined, all experimental features and breaking changes are disabled.
+
+    See :class:`cocotb.future.Future` for a list of available features.
+    The lowercase name of the feature is used to enable it.
+    Setting the value to ``1```, ``true``, or ``yes`` enables all features,
```
marlonjames:
```suggestion
    Setting the value to ``1``, ``true``, or ``yes`` enables all features,
```

marlonjames (approved):

PR #5641, opened 2026-06-20
docs: document Verilog parameter and VHDL generic override syntax

seam0814:
## Summary

Adds a new "Setting Top-Level Parameters and Generics" section to
`docs/source/building.rst` that documents the per-simulator syntax for
overriding top-level Verilog `parameter` and VHDL `generic` values from
the Makefile flow.

## Motivation

The Makefile flow does not expose a simulator-agnostic variable for
parameter/generic override. The only canonical reference is the
[`matrix_multiplier` example Makefile](../tree/master/examples/matrix_multiplier/tests/Makefile),
which is easy to miss for users coming from `building.rst`. This often
leads to questions on Gitter / Stack Overflow such as "how do I
override DATA_WIDTH at simulation time?" — even though every supported
simulator already handles it, just with different syntax.

## Change

- New section in `building.rst` with two reference tables:
  - **Verilog `parameter` override**: Icarus / Questa-family / VCS / Verilator / Xcelium / IUS
  - **VHDL `generic` override**: GHDL / Questa-family / Xcelium / IUS
- A pointer at the end to the simulator-agnostic `parameters=` argument
  of `Runner.build` for users on the Python Runner flow.

The syntax table is derived from `examples/matrix_multiplier/tests/Makefile`,
so there is a single source of truth between the example and the docs.

## Test

- Reviewed the rendered RST locally; tables format correctly.
- Cross-checked each Makefile snippet against the matrix_multiplier example.
- No code changes; no simulator runs required.

## Checklist

- [x] Newsfragment added (`docs/source/newsfragments/<PR>.doc.rst`)
- [x] No code changes
- [x] No new external dependencies
- [x] Docs follow existing role conventions (`:make:var:`, `:meth:`, `:reposrc:`, `:ref:`)

### DIFF
--- a/docs/source/building.rst
+++ b/docs/source/building.rst
@@ -169,3 +169,58 @@ Make Variables
 The :envvar:`COCOTB_TOPLEVEL` variable is also often used by the Makefile-based build and runner system.
 
 The :envvar:`SIM_CMD_PREFIX` and :envvar:`SIM_CMD_SUFFIX` environment variables are also supported by the Makefile-based build and runner system.
+
+
+Setting Top-Level Parameters and Generics
+=========================================
+
+The Makefile flow does not provide a simulator-agnostic variable for overriding
+top-level Verilog ``parameter`` or VHDL ``generic`` values. Instead, the appropriate
+syntax must be appended to :make:var:`COMPILE_ARGS`, :make:var:`SIM_ARGS`, or
+:make:var:`EXTRA_ARGS` for the simulator in use.
+
+The :reposrc:`matrix_multiplier example <examples/matrix_multiplier/tests/Makefile>`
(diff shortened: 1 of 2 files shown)


ktbarrett:
Out of scope.

seam0814:
Thanks for the quick look! I'd appreciate a pointer on where this would fit — e.g. would per-simulator sections in `simulator_support.rst` be a better home for this syntax, or do you have an existing place that already covers this that I missed? Happy to revise (or close this PR) once I know which direction you'd prefer.

ktbarrett:
It doesn't belong anywhere. cocotb isn't going to document simulators. Users can look at their simulator's documentation.

Issue #4077, opened 2024-07-29
Icarus Python runner should set toplevel at end of arguments


ktbarrett:
Setting the toplevel is sensitive to where in the argument list it is placed. It should be moved to the back after all modules have been analyzed.

ktbarrett:
This doesn't work on Windows or Mac. We also have no context as to why this change is needed, so let's drop it.

If any other user is having issues with this, please comment and we will re-open and fix it.

PR #5643, opened 2026-06-20
fix(runner): place Icarus -s <toplevel> after all sources

seam0814:
## Summary

Move `-s <hdl_toplevel>` (and `-s cocotb_iverilog_dump` when waves are
enabled) from the head of the `iverilog` command built by the Icarus
Python Runner to the tail, so the toplevel selection follows every
source file. Addresses issue #4077.

## Motivation

Issue #4077 reports that the position of `-s <toplevel>` in the
`iverilog` command line is sensitive: placing it before the sources can
cause unreliable elaboration in some designs. The Icarus Python Runner
currently emits:

```
iverilog -o sim.vvp -s <TOP> -g2012 ... <sources>
```

This PR changes it to:

```
iverilog -o sim.vvp -g2012 ... <sources> -s <TOP>
```

(and analogously moves the wave-dump toplevel selection to the end when
`waves=True`).

## Change

In `src/cocotb_tools/runner.py`, inside `Icarus._build_command`:

- Collect the `-s` arguments into a `toplevel_args` list at the start.
- Append `toplevel_args` to the end of the constructed command instead
  of injecting `-s <TOP>` early and pushing `-s cocotb_iverilog_dump`
  into `build_args`.

No other call sites need to change.

## Test

- No new dedicated unit test: the Icarus build path is already exercised
  by the simulator-required integration tests in
  `tests/pytest/test_runner.py`, which fail loudly if `iverilog`
  rejects the command line.
- Verified locally by inspecting the generated command from a sample
  build that the `-s` arguments now follow the source list.

I am happy to add a focused unit test that captures the command via
`monkeypatch` on `subprocess.run` if reviewers prefer.

## Checklist

- [x] Newsfragment added (`docs/source/newsfragments/<PR>.bugfix.rst`)
- [x] Behavior change is intentional and minimal
- [x] No public API change
- [x] No external dependency change

Refs #4077

### DIFF
--- a/docs/source/newsfragments/5643.bugfix.rst
+++ b/docs/source/newsfragments/5643.bugfix.rst
@@ -1,0 +1,1 @@
+Fix the Icarus Python Runner placing the ``-s <toplevel>`` argument before the source files, which can cause unreliable elaboration. The toplevel selection now follows all sources.
--- a/src/cocotb_tools/runner.py
+++ b/src/cocotb_tools/runner.py
@@ -977,9 +977,14 @@ class Icarus(Runner):
                 )
 
         build_args = [arg.value for arg in self._build_args]
+
+        # Icarus `-s` (toplevel selection) is sensitive to its position in the
+        # argument list; placing it after all sources is the documented robust
+        # ordering. See https://github.com/cocotb/cocotb/issues/4077.
+        toplevel_args: list[str] = ["-s", self.hdl_toplevel]
         if self.waves:
             self._create_iverilog_dump_file()
-            build_args += ["-s", "cocotb_iverilog_dump"]
+            toplevel_args += ["-s", "cocotb_iverilog_dump"]
 
         if self.timescale is not None:
             self._create_cmd_file()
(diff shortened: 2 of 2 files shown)


seam0814:
The CI tells me this fix passes on Linux iverilog v13.0 but breaks the Icarus runner on Windows (mingw + msvc) and macOS (homebrew), with 21 failing tests including the non-waves cases (`test_runner[None-False-parameters0/1]`, `test_timescale[*]`, `test_plusargs::test_toplevel_library`).

I'm reading that as: placing `-s <toplevel>` after the sources is not cross-platform safe with iverilog v13 — the Linux build accepts the new order but the Windows/macOS builds reject it even when no `-s cocotb_iverilog_dump` is involved.

Since you filed #4077 originally, would you mind sharing the environment where the toplevel-position sensitivity reproduced? That would tell me whether:

- the right fix is OS-conditional (move `-s` to the end only on the platforms where the original symptom appears), or
- the original symptom was already obsoleted by a later iverilog release on your side, in which case I'm happy to close this PR.

I'd rather not push a speculative variant that burns more CI time without your input.

ktbarrett:
That was 2 years ago and I was too lazy to actually write any context about what or why. If it doesn't work, let's close the issue and see if anyone else complains.

seam0814:
Closing as discussed — leaving #4077 for the maintainers to triage. Happy to revisit if the original toplevel-position issue reproduces on a specific platform/iverilog version.

